Issue #4074, opened 2019-05-10
meta-issue: All NumPy functions with no implementation


stuartarchibald:
Below are lists of feasible but not yet implemented NumPy functions... ideal for first time contributors. If you would like to contribute a function, please first check the open "Pull Requests" to make sure someone else isn't already working on the one you chose! If the check-box next to the function is checked it means the function is already done...

TODO: Link to `@overload` guide in docs once it merged.

It should be possible to write these functions using the [`numba.extending.overload`](http://numba.pydata.org/numba-doc/latest/extending/high-level.html) functionality, they are of varying difficulty:

* [ ] `np.alen`
* [ ] `np.allclose`
* [ ] `np.alltrue`
* [ ] `np.append`
* [ ] `np.argpartition`
* [ ] `np.argwhere`
* [ ] `np.around`
* [ ] `np.array_equal`
* [ ] `np.array_split`
* [ ] `np.asanyarray`
* [ ] `np.asarray_chkfinite`
* [ ] `np.asfarray`
* [ ] `np.asscalar`
* [ ] `np.atleast_1d`
* [ ] `np.atleast_2d`
* [ ] `np.atleast_3d`
* [ ] `np.bartlett`
* [ ] `np.binary_repr`
* [ ] `np.bincount`
* [ ] `np.blackman`
* [ ] `np.compress`
* [ ] `np.count_nonzero`
* [ ] `np.cross`
* [ ] `np.cumproduct` (just needs wiring)
* [ ] `np.diag_indices`
* [ ] `np.diag_indices_from`
* [ ] `np.diagflat`
* [ ] `np.euler_gamma` (constant)
* [ ] `np.fastCopyAndTranspose`
* [ ] `np.floatnonzero`
* [ ] `np.fliplr`
* [ ] `np.flipud`
* [ ] `np.fv`
* [ ] `np.hamming`
* [ ] `np.hanning`
* [ ] `np.hsplit`
* [ ] `np.i0`
* [ ] `np.in1d`
* [ ] `np.indices`
* [ ] `np.infty` (constant)
* [ ] `np.inner`
* [ ] `np.intersect1d`
* [ ] `np.ipmt`
* [ ] `np.irr`
* [ ] `np.isclose`
* [ ] `np.iscomplex`
* [ ] `np.iscomplexobj`
* [ ] `np.isfortran`
* [ ] `np.isneginf`
* [ ] `np.isposinf`
* [ ] `np.isreal`
* [ ] `np.isrealobj`
[...]

PR #5293, opened 2020-02-21
np.isnat

guilhermeleobas:
Add support for `np.isnat`. Still a work in progress.

#4074


esc:
@guilhermeleobas thanks for opening this PR. Your effort to improve Numba are appreciated! I have labeled this WIP for now, do ping us when it's ready for review.

guilhermeleobas:
One test is failing due to big/little endian differences. As soon as I address those, this PR will be ready for a round of reviews.

guilhermeleobas:
Ready for review as soon as CI passes

esc:
@guilhermeleobas thanks, I have changed the status of this PR now.

numba/np/npyfuncs.py
```diff
@@ -1473,6 +1473,13 @@ def np_datetime_isfinite_impl(context, builder, sig, args):
     return builder.icmp_unsigned('!=', args[0], npdatetime.NAT)
 
 
+def np_isnat_impl(context, builder, sig, args):
+    _check_arity_and_homogeneity(sig, args, 1, return_type=types.boolean)
+    return builder.or_(
+        builder.icmp_unsigned('==', args[0], npdatetime.NAT),
+        builder.icmp_unsigned('==', args[0], npdatetime.NAT_BIG_ENDIAN))
+
```
stuartarchibald: Are you sure this logic is valid? I think a big endian NaT would appear as a valid little endian datetime. e.g.
```python
from numba import njit
import numpy as np

@njit
def foo(a):
    return np.isnat(a)

BENaT = np.datetime64('NaT').byteswap()
arg = np.array(BENaT.view(np.int64), np.dtype('<M8[Y]'))
print("arg", arg)

print(foo.py_func(arg))
print(foo(arg))
```
on a LE machine produces:
```
arg 2098
False
True
```
I'd anticipate that the inverse would hold on a BE machine.

stuartarchibald: Further, it's probably worth checking if any of the other `is*` functions work on BE before going down the route of attempting to support it?

guilhermeleobas: What do you mean by _BE_ and _LE_?

guilhermeleobas: Hey @stuartarchibald, you're right. My bet is that the [current NumPy implementation](https://github.com/numpy/numpy/blob/0b9b176dcef7f14f56dbd543c89e6da6c6b7b2e4/numpy/core/src/umath/loops.c.src#L1085-L1092) for `np.isnat` does a byte-swapping of the underlying data. Thus, `np.isnat(np.array(['NaT'], dtype='>datetime64[Y]'))` is actually comparing `NPY_DATETIME_NAT == arg.byteswap()`

```python
>>> np.array(['NaT'], dtype='>datetime64[Y]').view(np.int64)
array([128])

>>> np.array(['NaT'], dtype='>datetime64[Y]').byteswap().view(np.int64)
array([-9223372036854775808])

>>> np.array(['NaT'], dtype='<datetime64[Y]').view(np.int64)
array([-9223372036854775808])
```

guilhermeleobas:
Just notice that Numba already has a function to check if a value **isn't ** a NAT.
https://github.com/numba/numba/blob/13ece9b97e6f01f750e870347f231282325f60c3/numba/np/npdatetime.py#L92-L96

This will not work for the following input: `np.array(['NaT'], dtype='>datetime64[Y]')`

stuartarchibald:
Thanks, fixes. Please could you add this to the docs here https://github.com/numba/numba/blame/e20b17900ca24bada7ce51accf656b5db39f8fc7/docs/source/reference/numpysupported.rst#L745 in a new section `Datetime functions`. Thanks.

stuartarchibald:
> Thanks for the fixes. Please could you add this to the docs here https://github.com/numba/numba/blame/e20b17900ca24bada7ce51accf656b5db39f8fc7/docs/source/reference/numpysupported.rst#L745 in a new section `Datetime functions`. Thanks.

Patched in f5d250f

guilhermeleobas:
Just saw the notification for this PR. Is there anything else to be done in this PR?

stuartarchibald:
Assuming the docs build fine, all done, thanks for your efforts.

guilhermeleobas:
Glad to help

stuartarchibald:
Manually checked the doc build. Looks fine.

PR #5296, opened 2020-02-21
Refactor parfor: better exception from passes

sklam:
as titled.

### DIFF
--- a/numba/core/errors.py
+++ b/numba/core/errors.py
@@ -493,7 +493,7 @@ class UnsupportedError(NumbaError):
     pass
 
 
-class RewriteUnsupportedError(UnsupportedError):
+class UnsupportedRewriteError(UnsupportedError):
     """UnsupportedError from rewrite passes
     """
     pass
--- a/numba/core/errors.py
+++ b/numba/core/errors.py
@@ -493,6 +493,12 @@ class UnsupportedError(NumbaError):
     pass
 
 
+class RewriteUnsupportedError(UnsupportedError):
+    """UnsupportedError from rewrite passes
+    """
+    pass
+
+
 class IRError(NumbaError):
     """
     An error occurred during Numba IR generation.
--- a/numba/parfors/array_analysis.py
+++ b/numba/parfors/array_analysis.py
@@ -1825,7 +1825,7 @@ class ArrayAnalysis(object):
             shape_var = kws['shape']
         if shape_var:
             return shape_var, []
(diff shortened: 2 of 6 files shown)


numba/core/errors.py
```diff
@@ -493,6 +493,12 @@ class UnsupportedError(NumbaError):
     pass
 
 
+class RewriteUnsupportedError(UnsupportedError):
```
stuartarchibald: Wondering if `UnsupportedRewriteError` sounds better?

numba/parfors/array_analysis.py
```diff
@@ -1766,11 +1769,11 @@ def _analyze_op_call(self, scope, equiv_set, expr):
                 expr.args = args[1:]
             return result
 
-    def _analyze_op_call___builtin___len(self, scope, equiv_set, args, kws):
+    def _analyze_op_call___builtin___len(self, scope, equiv_set, loc, args, kws):
         # python 2 version of len()
```
stuartarchibald: Py 2... dead?

numba/parfors/parfor.py
```diff
@@ -1836,7 +1838,9 @@ def _numpy_to_parfor(self, equiv_set, lhs, expr):
         if call_name in ['zeros', 'ones'] or mod_name == 'numpy.random':
             return self._numpy_map_to_parfor(equiv_set, call_name, lhs, args, kws, expr)
         # return error if we couldn't handle it (avoid rewrite infinite loop)
-        raise NotImplementedError("parfor translation failed for ", expr)
+        raise errors.RewriteUnsupportedError(
+            f"parfor translation failed for {expr}", loc=exprloc,
```
stuartarchibald: Should that be `expr.loc`? If so, no build failures suggests this isn't tested?

sklam: good catch. some of the exception raise in Parfor are not covered by the testsuite

sklam: not sure how to trigger this one.

numba/core/errors.py
sklam: it does

numba/parfors/array_analysis.py
sklam: i'll go after all the `__builtin__` references

sklam: they are removed

numba/core/errors.py
sklam: renamed

PR #5283, opened 2020-02-19
Improve error location and reporting for tuple use in transformed parallel region

stuartarchibald:
As title. xref: https://github.com/numba/numba/issues/5282

PR #5299, opened 2020-02-21
Initial attempt to support tuples to parfors.

### LINKED ISSUE #5282: Numba (parallel=True) does not handle tuples in certain situations
## Reporting a bug

This issue is to declare that there is a known problem with Numba's parallel transforms in relation to the use of `tuples`. It has historically been reported as a message like:
```
LoweringError: scalar type tuple(<OF SOME TYPE>) given for non scalar argument #<SOME ARGUMENT INDEX>
```
and comes from using a `tuple` in a parallel region. For example:
```python

@njit(parallel=True)
def tuple_bug():
    t = (10,)
    z = 0
    for x in prange(10):
        z += t[0]
    return z

print(tuple_bug())
```
gives:
```
scalar type UniTuple(Literal[int](10) x 1) given for non scalar argument #2
```
The reason this happens is because the parallel regions in Numba are implemented as `guvectorize`-like functions, and as a limitation of the Generalized Universal Functions, tuples are not supported as arguments. Relating this to the above example, when Numba compiles it, it "sees" the `prange` loop as a parallelisation opportunity and converts the loop body into the kernel of a `guvectorize` function, however, the use of the variable `t` in the kernel body is essentially illegal as it is a tuple.

I'm going to propose this patch https://github.com/stuartarchibald/numba/commit/52285b9a23bbb89e0298f5d2371021051fd112bb goes into master as it will catch illegal kernel bodies much earlier in the transform process so as to provide feedback to the user about the issue.

The eventual fix should be to just rewrite the Generalized Universal Functions implementation to handle all containers.

This ticket is to act as an acknowledgement of the problem, all other tickets reporting this bug can be closed as duplicates and cross referenced here.

DrTodd13:
Partially resolves #5282 This is to get the discussion going.  This patch convert tuples of size less than 10 to individual parameters.  Are the cases we know about with really large counts UniTuples?  If so then maybe conversion to an array would be better?  

We can try to generate a better error message in this area where I check types but it is still in lowering.  I can add something to check it before lowering as well.  Thoughts?

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,5 +1,5 @@
 include MANIFEST.in
-include README.rst setup.py runtests.py versioneer.py CHANGE_LOG AUTHORS LICENSE
+include README.rst setup.py runtests.py versioneer.py CHANGE_LOG LICENSE
 
 recursive-include numba *.c *.cpp *.h *.hpp *.inc
 recursive-include docs *.ipynb *.txt *.py Makefile *.rst
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -10,7 +10,7 @@ jobs:
 - template: buildscripts/azure/azure-linux-macos.yml
   parameters:
     name: macOS
-    vmImage: xcode9-macos10.13
+    vmImage: macOS-10.14
     matrix:
       py38_np117:
         PYTHON: '3.8'
--- a/numba/core/analysis.py
+++ b/numba/core/analysis.py
@@ -347,6 +347,21 @@ def dead_branch_prune(func_ir, called_args):
         taken = do_prune(take_truebr, blk)
         return True, taken
 
+    def prune_by_predicate(branch, pred, blk):
+        try:
+            # Just to prevent accidents, whilst already guarded, ensure this
+            # is an ir.Const
+            if not isinstance(pred, (ir.Const, ir.FreeVar, ir.Global)):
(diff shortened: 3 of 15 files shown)


DrTodd13:
I did some work to try to support named tuples.  It kind of works.  The tricky part is re-creating the named tuple inside the gufunc.  I can easily create the correct syntax but that won't compile because the name of the named tuple is an unrecognized global.  My current fix is to lookup the definition of the original tuple var in the original function (that gives something like call $2load_global.0(...)) then get the definition of $2load_global.0 which gives me global(TestNamedTuple: <class 'numba.tests.test_parfors.TestNamedTuple'>).  From that, I can add the equivalent of globls["TestNamedTuple"] = exec(path_to_test_name_tuple_from_the_global) and that globls is used to exec the gufunc into existence.  This isn't very elegant.  Is there a better way?

stuartarchibald:
@DrTodd13 thanks for looking at this. It's now scheduled it for discussion at tomorrow's core dev meeting.

stuartarchibald:
From core developer meeting:
* @DrTodd13 to merge https://github.com/numba/numba/pull/5283 into this branch and edit the check/messages as needed.
* @stuartarchibald to checkout this branch and see what the impact of tuple size is on LLVM (i.e. does it break ever, what's it do to compilation time and memory etc).
* Result of the above will guide how to proceed with any restriction on tuple size.

PR #5308, opened 2020-02-24
Provide `numba.extending.is_jitted`

stuartarchibald:
Provides `numba.extending.is_jitted` as a means to query
whether a function is JIT wrapped.

### DIFF
--- a/numba/core/extending.py
+++ b/numba/core/extending.py
@@ -511,3 +511,16 @@ class BoundLiteralArgs(collections.namedtuple(
             args,
             kwargs,
         )
+
+
+def is_jitted(function):
+    """Returns True if a function is wrapped by one of the Numba @jit
+    decorators, for example: numba.jit, numba.njit
+
+    The purpose of this function is to provide a means to check if a function is
+    already JIT decorated.
+    """
+
+    # don't want to export this so import locally
+    from numba.core.dispatcher import Dispatcher
+    return isinstance(function, Dispatcher)
--- a/.flake8
+++ b/.flake8
@@ -51,7 +51,6 @@ exclude =
     numba/core/dataflow.py
     numba/runtests.py
     numba/core/pythonapi.py
-    numba/extending.py
     numba/core/decorators.py
     numba/core/typeconv/typeconv.py
     numba/core/typeconv/rules.py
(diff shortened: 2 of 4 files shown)


numba/core/extending.py
```diff
@@ -511,3 +511,16 @@ def bind(self, *args, **kwargs):
...
+    """Returns True if a function is wrapped by one of the Numba @jit
+    decorators, for example: numba.jit, numba.njit
+
+    The purpose of this function is to provide a means to check if a function is
+    already JIT decorated.
+    """
+
+    # don't want to export this so import locally
+    from numba.core.dispatcher import Dispatcher
+    return isinstance(function, Dispatcher)
```
stuartarchibald: Open questions.
1. Should this also include dynamic ufuncs, i.e.  `DUFunc` as a kwarg option like `include_dufunc=False/True`
2. This currently ignores CUDA and ROC jitted functions, do we want this?
3. Did I miss any other spelling/pattern?
4. Presume we have no way of knowing if a e.g. `guvectorize` function that materializes as a `np.ufunc` instance is actually a JIT compiled one? Would it even be useful/matter if we did know.

stuartarchibald: CC @sklam @TomAugspurger @mroeschke

TomAugspurger: I don't have strong thoughts on here but, from what I recall the current definition is sufficient for pandas.

mroeschke: Agreed this looks sufficient.

In pandas we're using `isinstance(func, numba.targets.registry.CPUDispatcher)` and I'm assuming `isinstance(function, Dispatcher)` is fully equivalent?

stuartarchibald: Great, thanks for confirming.

Should be functionally equivalent acceptance for `CPUDispatcher`, but would also allow other `Dispatcher`s through.
```pycon
In [1]: from  numba.core.registry import CPUDispatcher                                                                          

In [2]: CPUDispatcher.__mro__                                                                                                   
Out[2]: 
(numba.core.registry.CPUDispatcher,
 numba.core.dispatcher.Dispatcher,
 numba.core.dispatcher._DispatcherBase,
 _dispatcher.Dispatcher,
 object)
```

sklam: Restricting to just cpu side `@jit is enough. Ufuncs and the GPU jits are quite different.

PR #5292, opened 2020-02-21
fix argument binding with var positional and kwargs

ARF1:
Fixes and closes #5291.

PR #5309, opened 2020-02-25
support for function calls with *args and named kwargs

### LINKED ISSUE #5304: Function calls cannot combine tuple unpacking and keyword arguments
On the current master, it is not possible for function calls to use tuple unpacking and keyword arguments:

```python
from numba import njit

@njit
def foo(*args, kwarg=None):
    print(args)
    print(kwarg)

@njit
def bar(a, *args):
    foo(a, *args, kwarg='foobar')

bar(1, 2, 3)
```

Output:
```
UnsupportedError: Failed in nopython mode pipeline (step: analyzing bytecode)
CALL_FUNCTION_EX with **kwargs not supported
```

To me this looks like a bug since `**kwargs` is not used anywhere. Clearly `**kwargs` would be a problem with its undefined length at compile time, but here the number of arguments is defined at compile time.

Note: Manually unpacking the tuple with `foo(a, args[0], args[1], kwarg='foobar')` works.

ARF1:
As title.

Fixes #5304.


stuartarchibald:
Thanks for the PR.

Note for reviewers. Blocked on #5292.

ARF1:
@stuartarchibald Since #5292 is blocked on a bug of the dispatcher, would it be acceptable to limit the scope of this PR and of #5292 to nopython mode and deal with fixing the python dispatcher at some later point?

PR #5162, opened 2020-01-29
Specify synchronization semantics of CUDA Array Interface

gmarkall:
This PR adds a new key, `stream`, to the CUDA Array Interface (CAI) specification, and updates the CAI implementation in Numba to follow these changes. The semantics of this have been iterated towards through the discussion in this PR, so the original PR description suggests a different idea. However, I've left the original description in place below so that the discussion can be followed if necessary.

To save anyone generating docs locally (or trying to read them in the diff), the specification as suggested in this PR is available at: https://gmarkall.github.io/numba-rtd-theme/cuda/cuda_array_interface.html. The majority of the text added in this PR is the [Synchronization section](https://gmarkall.github.io/numba-rtd-theme/cuda/cuda_array_interface.html#synchronization).

The Numba implementation follows that described in the specification, and adds tests to ensure that it:
- Correctly populates in the `stream` field of the interface when acting as a Producer, and
- Synchronizes at the correct points when acting as a Consumer.

# Original PR description

Following on from discussions in Issues #4933 and #4886, this commit clarifies the synchronization and lifetime of `__cuda_array_interface__` objects. In summary:

- Synchronization: Producers and consumers of arrays on the CUDA array interface should operate on those arrays in the default stream, or synchronized on the default stream, in order to implicitly be in sync. In special cases (e.g. where the same stream is used across the producer / consumer boundary) the synchronization on the default stream may be elided.
- Lifetime: Consuming the CUDA array interface does not extend the lifetime of the object owning the data. The consumer must ensure that a reference to the owner is kept as long as the data is required. I think this isn't really a new requirement, but codifies something that was implicitly required in past versions.

I've avoided referring to the "legacy default stream" as that serializes everything (see e.g. https://devblogs.nvidia.com/gpu-pro-tip-cuda-7-streams-simplify-concurrency/) - I think we really do just want to synchronize with the default stream (happy to be shown why that's not the case if it isn't though :-) )

The version number is bumped - I think for these changes it's important to bump the version number as it's possible to implement v2 correctly right now, but not match the synchronization semantics specified in this PR, so it's needed to be sure both sides of the interface agree on the semantics.

Some changes are also made to Numba to support the correct implementation of these semantics by users:

- `from_cuda_array_interface` and `as_cuda_array` now have a True-by-default `sync` kwarg, which indicates Numba should bind the new device array to the default stream.
- `__cuda_array_interface__` for a device array records an event on the stream the array is bound to and inserts a wait on the event in the default stream.

[...]

PR #5323, opened 2020-02-27
Document lifetime semantics of CUDA Array Interface

### LINKED ISSUE #5164: [DOC] from_cuda_array_interface
## Feature request

It appears `from_cuda_array_interface` is available from `numba.cuda`, but currently doesn't appear in the API docs. This looks like a useful function that would be good to generally make visible to users.

gmarkall:
This documents the lifetime of objects in relation to the CUDA Array Interface - the content is taken from #5162 with some editing for clarity, as this is the uncontentious part of the PR.

The wording only clarifies and makes explicit what is already required of the user of the interface - it does not specify any semantic changes.

This also resolves #5164 (document `from_cuda_array_interface`).

### DIFF
--- a/docs/source/cuda/cuda_array_interface.rst
+++ b/docs/source/cuda/cuda_array_interface.rst
@@ -76,6 +76,43 @@ The following are optional entries:
             and will raise a ``NotImplementedError`` exception if one is passed
             to a GPU function.
 
+
+Lifetime management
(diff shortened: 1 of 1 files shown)


stuartarchibald:
Thanks for the PR!

leofang:
This one is quick!

PR #5333, opened 2020-03-02
Add: highlight Numba IR dump

### LINKED ISSUE #4657: Implement a pygments lexer for Numba IR
## Feature request

It'd be nice, especially in jupyter notebooks/ipython if Numba IR was syntax highlighted. I think most of the work to do this will involve implementing an appropriate lexer for pygments, docs: http://pygments.org/docs/lexerdevelopment/

1844144:
Closes #4657

Colors in ipython(white background) and Jupyter are ok. `Terminal256Formatter` doesn't recognize black background currently so colors a bit dark on black background.
Demo: 
image

### DIFF
--- a/numba/core/ir.py
+++ b/numba/core/ir.py
@@ -1462,8 +1462,8 @@ class FunctionIR(object):
                     raise ValueError(msg)
                 else:
                     from pygments import highlight
-                    from numba.misc.dump_style \
-                        import NumbaIRLexer as lexer, by_colorscheme
+                    from numba.misc.dump_style import NumbaIRLexer as lexer
+                    from numba.misc.dump_style import by_colorscheme
                     from pygments.formatters import Terminal256Formatter
                     print(highlight(text, lexer(), Terminal256Formatter(
                         style=by_colorscheme())))
--- a/numba/core/ir.py
+++ b/numba/core/ir.py
@@ -1462,9 +1462,11 @@ class FunctionIR(object):
                     raise ValueError(msg)
                 else:
                     from pygments import highlight
-                    from .ir_lexer import NumbaIRLexer as lexer
+                    from numba.misc.dump_style \
+                        import NumbaIRLexer as lexer, by_colorscheme
                     from pygments.formatters import Terminal256Formatter
-                    print(highlight(text, lexer(), Terminal256Formatter()))
+                    print(highlight(text, lexer(), Terminal256Formatter(
+                        style=by_colorscheme())))
             else:
                 print(text)
 
--- a/numba/core/ir.py
+++ b/numba/core/ir.py
@@ -1462,7 +1462,7 @@ class FunctionIR(object):
                     raise ValueError(msg)
                 else:
                     from pygments import highlight
-                    from pygments.lexers import DelphiLexer as lexer
+                    from .ir_lexer import NumbaIRLexer as lexer
                     from pygments.formatters import Terminal256Formatter
                     print(highlight(text, lexer(), Terminal256Formatter()))
             else:
--- a/numba/core/ir_lexer.py
+++ b/numba/core/ir_lexer.py
@@ -1,0 +1,49 @@
+# -*- coding: utf-8 -*-
+
+from pygments.lexer import RegexLexer, include, bygroups, words
+from pygments.token import Text, Name, String,  Punctuation, Keyword, Operator
+
+
+class NumbaIRLexer(RegexLexer):
+    """
+    For Numba IR code.
+    """
+    name = 'Numba_IR'
+    aliases = ['numba_ir']
+    filenames = ['*.numba_ir']
+
+    identifier = r'\$[a-zA-Z0-9._]+'
+    fun_or_var = r'([a-zA-Z_]+[a-zA-Z0-9]*)'
+
+    tokens = {
+        'root' : [
+            (r'(label)(\ [0-9]+)(:)$',
+                bygroups(Keyword, Name.Label, Punctuation)),
+
+            (r' = ', Operator),
+            include('whitespace'),
+            include('keyword'),
+
+            (identifier, Name.Variable),
+            (fun_or_var + r'(\()',
+                bygroups(Name.Function, Punctuation)),
+            (fun_or_var + r'(\=)',
+                bygroups(Name.Attribute, Punctuation)),
+            (fun_or_var, Name.Constant),
+
+            # <built-in function some>
+            (r'<.*>', String),
+
+            (r'[=<>{}\[\]()*.,!]|x\b', Punctuation)
+        ],
+
+        'keyword':[
+            (words((
+                'del', 'jump', 'call', 'branch',
+            ), suffix=' '), Keyword),
+        ],
+
+        'whitespace': [
+            (r'(\n|\s)', Text),
+        ],
+    }
--- a/numba/core/ir_lexer.py
+++ b/numba/core/ir_lexer.py
@@ -32,7 +32,7 @@ class NumbaIRLexer(RegexLexer):
             (fun_or_var, Name.Constant),
 
             # <built-in function some>
-            (r'<.*>', String),
+            (r'<[^>]*>', String),
 
             (r'[=<>{}\[\]()*.,!]|x\b', Punctuation)
         ],
--- a/numba/misc/dump_style.py
+++ b/numba/misc/dump_style.py
@@ -45,7 +45,7 @@ class NumbaIRLexer(RegexLexer):
             (r'[0-9]+', Number),
 
             # <built-in function some>
-            (r'<[^>]*>', String),
+            (r'<[^>\n]*>', String),
 
             (r'[=<>{}\[\]()*.,!\':]|x\b', Punctuation)
         ],
--- a/numba/misc/dump_style.py
+++ b/numba/misc/dump_style.py
@@ -1,6 +1,10 @@
+try:
+    from pygments.styles.default import DefaultStyle
+except ImportError:
+    msg = "Please install pygments to see highlighted dumps"
+    raise ImportError(msg)
 
 import numba.core.config
-from pygments.styles.default import DefaultStyle
 from pygments.styles.manni import ManniStyle
 from pygments.styles.monokai import MonokaiStyle
 from pygments.styles.native import NativeStyle
@@ -59,6 +63,10 @@ class NumbaIRLexer(RegexLexer):
 
 
 def by_colorscheme():
+    """
+    Get appropriate style for highlighting according to
+    NUMBA_COLOR_SCHEME setting
+    """
     styles = DefaultStyle.styles.copy()
     styles.update({
         Name.Variable:        "#888888",
--- a/numba/misc/dump_style.py
+++ b/numba/misc/dump_style.py
@@ -14,7 +14,7 @@ from pygments.style import Style
 
 class NumbaIRLexer(RegexLexer):
     """
-    For Numba IR code.
+    Pygments style lexer for Numba IR (for use with highlighting etc).
     """
     name = 'Numba_IR'
     aliases = ['numba_ir']
--- a/numba/misc/dump_style.py
+++ b/numba/misc/dump_style.py
@@ -1,14 +1,13 @@
 
 import numba.core.config
 from pygments.styles.default import DefaultStyle
-from pygments.styles.bw import BlackWhiteStyle
 from pygments.styles.manni import ManniStyle
 from pygments.styles.monokai import MonokaiStyle
 from pygments.styles.native import NativeStyle
 
 from pygments.lexer import RegexLexer, include, bygroups, words
 from pygments.token import Text, Name, String,  Punctuation, Keyword, \
-    Operator, Number, Comment, Error
+    Operator, Number
 
 from pygments.style import Style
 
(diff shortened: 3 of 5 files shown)


stuartarchibald:
@1844144 Thanks for the PR. This looks great already! Am going to try it out...!

numba/core/ir_lexer.py
```diff
@@ -0,0 +1,49 @@
+# -*- coding: utf-8 -*-
```
eric-wieser: Is this necessary in python 3?

stuartarchibald:
Just tried this out locally, looks good :)

> Closes #4657
> 
> Colors in ipython(white background) and Jupyter are ok. `Terminal256Formatter` doesn't recognize black background currently so colors a bit dark on black background.

Numba has a `color_scheme` config option:
https://github.com/numba/numba/blob/a4c4a8dceb0801672d6fda9199c36047dba16f0b/numba/core/config.py#L143
which guides the colour scheme in its error message reporting. Perhaps this could be used to influence the formatter?

stuartarchibald:
Reference for colour schemes: http://numba.pydata.org/numba-doc/latest/reference/envvars.html#envvar-NUMBA_COLOR_SCHEME

numba/core/ir_lexer.py
1844144: In case our lexer will be included in Pygments  official repo.

1844144: But you are right, not necessary

1844144:
Not sure we need to use `colorama`, because `pygments` already has enough styles for different backgrounds (so we can use/tweak it).
https://github.com/pygments/pygments/blob/827e890fe10b4bb8434ffb6350e592cece4b033c/pygments/styles/__init__.py#L17

If it is ok to not to use colorama, I will suggest a style for each case (or even we may allow to select `pygments`'s styles by name along with our predefined styles)

stuartarchibald:
> Not sure we need to use `colorama`, because `pygments` already has enough styles for different backgrounds (so we can use/tweak it).
> https://github.com/pygments/pygments/blob/827e890fe10b4bb8434ffb6350e592cece4b033c/pygments/styles/__init__.py#L17
>
> If it is ok to not to use colorama, I will suggest a style for each case (or even we may allow to select `pygments`'s styles by name along with our predefined styles)

Yes, I was thinking that the styles chosen could just follow Numba's already used names based on output type:
```
    no_color - No color added, just bold font weighting.
    dark_bg - Suitable for terminals with a dark background.
    light_bg - Suitable for terminals with a light background.
    blue_bg - Suitable for terminals with a blue background.
    jupyter_nb - Suitable for use in Jupyter Notebooks.
```
Suggesting some styles would be good thanks.

1844144:
Suggestions: 

blue_bg -> NativeStyle (not sure my blue background is that what you mean)
image

dark_bg -> MonokaiStyle
image

light_bg -> ManniStyle OR DefaultStyle
image

jupyter_nb -> DefaultStyle
image

no_color -> customized BlackWhiteStyle
image

1844144:
Few notes:
 - if developer will turn on DUMP_IR he will see black/white style as default value for COLOR_SCHEME is no_color (as for me better would be to have jupyter_nb as default)
 - there are other places, where dumps are highlighted independently from COLOR_SCHEME
https://github.com/numba/numba/blob/90654c5ac525ef7eeaa69f605c15710bc40476de/numba/core/lowering.py#L195-L197
and 
https://github.com/numba/numba/blob/90654c5ac525ef7eeaa69f605c15710bc40476de/numba/core/codegen.py#L36-L39
And code assume you will have **light** background

stuartarchibald:
Thanks for the above. These colour schemes are a good fit for the terminal/output formats we encoded for the error messages. I use dark background terminals, the `MonokaiStyle` looks good and the highlighting is really helpful!

> Few notes:
> 
>     * if developer will turn on DUMP_IR he will see black/white style as default value for COLOR_SCHEME is no_color (as for me better would be to have jupyter_nb as default)

hmmm, is there a theme that renders reasonably on both a dark and a light background? This is what put us off doing anything with colour in the error message highlighting, it wasn't clear how to reasonably accommodate both, so font weighting was applied instead.
 
>     * there are other places, where dumps are highlighted independently from COLOR_SCHEME
>       https://github.com/numba/numba/blob/90654c5ac525ef7eeaa69f605c15710bc40476de/numba/core/lowering.py#L195-L197
>       
>       and
>       https://github.com/numba/numba/blob/90654c5ac525ef7eeaa69f605c15710bc40476de/numba/core/codegen.py#L36-L39
>       
>       There are supposed that you will have **light** background

The above were thrown in during the last release cycle by me as I was working on really involved IR transforms and just wanted something more readable. If it's easy/you are motivated to patch in the `COLOR_SCHEME` stuff into these locations too it would be welcomed, else I think I understand enough about what's going on now to be able to do it :)

Thanks again, this is incredibly useful!!!

1844144:
Ok, then

1. Modified default theme used as default(no_color) scheme 
image
image

2. Please decide what style should be used for `light_bg`: ManniStyle (as now) or same style as `jupyter_nb` (see images above)

3. Color_scheme in other locations patched

stuartarchibald:
> Ok, then
> 
>     1. Customized DefaultStyle used as default(no_color) scheme

Thanks, this looks good.
 
>     2. Please decide what style should be used for `light_bg`: ManniStyle (as now) or same style as `jupyter_nb` (see images above)

I quite like `ManniStyle`, I guess it's trivial to change later if there are any objections.

>     3. Color_scheme in other locations patched

Brilliant, thanks.

numba/misc/dump_style.py
```diff
@@ -0,0 +1,76 @@
...
+            ), suffix=' '), Keyword),
+        ],
+
+        'whitespace': [
+            (r'(\n|\s)', Text),
+        ],
+    }
+
+
+def by_colorscheme():
```
stuartarchibald: Please could you add a docstring, thanks.

numba/misc/dump_style.py
```diff
@@ -0,0 +1,76 @@
...
+from pygments.styles.default import DefaultStyle
+from pygments.styles.manni import ManniStyle
+from pygments.styles.monokai import MonokaiStyle
+from pygments.styles.native import NativeStyle
+
+from pygments.lexer import RegexLexer, include, bygroups, words
+from pygments.token import Text, Name, String,  Punctuation, Keyword, \
+    Operator, Number
+
+from pygments.style import Style
```
stuartarchibald: I think it'd be good to have a `try..except ImportError` around this to catch the case where Numba extension writers might try and import this module for use in their code but not have `pygments` installed (as it's only an optional requirement for Numba).  Thinking something like:
https://github.com/numba/numba/pull/5333/files#diff-5de309739e66ef02bd05362a66f0caa1R191-R201

numba/core/ir.py
```diff
@@ -1462,9 +1462,11 @@ def dump(self, file=None):
                     raise ValueError(msg)
                 else:
                     from pygments import highlight
-                    from pygments.lexers import DelphiLexer as lexer
+                    from numba.misc.dump_style \
+                        import NumbaIRLexer as lexer, by_colorscheme
```
stuartarchibald: Perhaps split this as:
```python
from numba.misc.dump_style import NumbaIRLexer as lexer
from numba.misc.dump_style import by_colorscheme
```
the code base layout was recently refactored and it was found that refactoring tools seemed to struggle with more complex spellings like in the above.

numba/misc/dump_style.py
```diff
@@ -0,0 +1,76 @@
+
```
stuartarchibald: Empty line? delete?

numba/misc/dump_style.py
```diff
@@ -0,0 +1,76 @@
...
+from pygments.lexer import RegexLexer, include, bygroups, words
+from pygments.token import Text, Name, String,  Punctuation, Keyword, \
+    Operator, Number
+
+from pygments.style import Style
+
+
+class NumbaIRLexer(RegexLexer):
+    """
+    For Numba IR code.
```
stuartarchibald:
```suggestion
    Pygments style lexer for Numba IR (for use with highlighting etc).
```

1844144:
Done

stuartarchibald:
Thanks for the fixes. I'm happy to merge this as-is as it's a massive improvement on what's in the code base at present! Congratulations on your first contribution to Numba!

In manual testing I did notice that there are some cases which "break", for example:
```python
from numba import config
config.DEBUG=1
from numba import njit
import numpy as np

@njit
def foo(x):
    y = np.sin(x)
    return x + y[0]

foo(np.ones(5))
```
produces a load of output, including an IR dump for `_broadcast_onto`, which highlights incorrectly:

highlighting_issue1
perhaps something to fix in a follow up PR? I also wonder if we can make the Numba IR output easier to parse?!

I also noticed that the LLVM Lexer doesn't "understand" our slightly unusual LLVM IR all the time either, but that's an issue for another time!

stuartarchibald:
@gmarkall wonder if a PTX (and maybe later SASS) lexer would be useful for CUDA dumps?

stuartarchibald:
Note to self. This PR would be good to highlight in the release notebook for 0.49 as it's a great improvement over the existing output and something that should be demonstrated and publicly advertised, it's also obviously extremely helpful for users working with Numba internals.

1844144:
Thanks for testing, would be good to have most complicated function to test lexer on.

If you will explain your changes(or put a link to the source) to LLVM syntax, then will be easier to adjust it

stuartarchibald:
> Thanks for testing, 

No problem.

> would be good to have most complicated function to test lexer on.

This one is pretty involved, it's designed for testing our IR-diff tool: https://github.com/numba/numba/blob/868b8e3e8d034dac0440b75ca31595e07f632d27/numba/tests/test_ir.py#L349-L385

I've extracted it here:
```python
from numba.core import compiler
import numpy as np

_GLOBAL = 1234
def gen():
    _FREEVAR = 0xCAFE

        f = a + b
        a += _FREEVAR
        h = f + g
        i = 1j / d
        if np.abs(i) > 0:
            k = h / i
            l = np.arange(1, c + 1)
            with objmode():
                print(e, k)
            m = np.sqrt(l - g)
            if np.abs(m[0]) < 1:
                n = 0
                for o in range(a):
                    n += 0
                    if np.abs(n) < 3:
                        break
                n += m[2]
            p = g / l
            q = []
            for r in range(len(p)):
                q.append(p[r])
                if r > 4 + 1:
                    with objmode(s='intp', t='complex128'):
                        s = 123
                        t = 5
                    if s > 122:
                        t += s
                t += q[0] + _GLOBAL

        return f + o + r + t + r + a + n

    return foo

ir = compiler.run_frontend(gen())
print(ir.dump())
```
and it looks good to me. This one runs a load of the more recently supported features, including a lot of transformation/code generation:
```python
from numba import config
config.DEBUG_PRINT_AFTER='ir_legalization'
from numba import literal_unroll, njit, gdb
import numpy as np

import operator
from functools import reduce
from numba.typed import List

@njit(['int64()',])
def newish_features():

    def mul_n(x, multiplier):
        return x * multiplier

    V = 20
    def greater_than_V(x):
        return x > V

    reduce_lambda = lambda x, y: (x * 2) + y

    gdb()

    a = [x ** 2 for x in range(10)]
    n = len(a)
    try:
        return a[3]
    except Exception:
        print("caught exception")
        return -1

    tup = (1, 'a', 2j)
    for i in literal_unroll(tup):
        print(i)

    return reduce(reduce_lambda, filter(greater_than_V, map(mul_n, a, range(n))))
```
and again seems reasonable.

> If you will explain your changes(or put a link to the source) to LLVM syntax, then will be easier to adjust it

Here's an example:
```python
from numba import config
config.DUMP_OPTIMIZED=1
from numba import njit
import numpy as np

@njit
def foo(x):
    y = np.sin(x)
    return x + y[0]

foo(np.ones(5))
```
produces something like:
highlighting_issue2

the LLVM Lexer in use is the one here: https://github.com/numba/numba/blob/868b8e3e8d034dac0440b75ca31595e07f632d27/numba/core/lowering.py#L193

LLVM language reference is here http://llvm.org/docs/LangRef.html, I wonder if part of what is observed is possibly due to the Lexer not having a specific LLVM IR version, and this being a mis-match.

1844144:
Looks like you are using Gas lexer for optimized dump (which for Gas (AT&T) assembly code) is it ok?
https://github.com/numba/numba/blob/868b8e3e8d034dac0440b75ca31595e07f632d27/numba/core/codegen.py#L34-L36

stuartarchibald:
> Looks like you are using Gas lexer for optimized dump (which for Gas (AT&T) assembly code) is it ok?
> 
> https://github.com/numba/numba/blob/868b8e3e8d034dac0440b75ca31595e07f632d27/numba/core/codegen.py#L34-L36
> 
> Changing to LlvmLexer gives desired result. Is there a cases when you need Gas lexer?

Oops, that'd explain it, needs two code paths! RE GAS Lexer, yes, we often want the LLVM IR and the assembly next to each other, setting the env var `NUMBA_DEBUG=1` is common when debugging and it dumps the Numba IR, LLVM IR and assembly.

gmarkall:
>  @gmarkall wonder if a PTX (and maybe later SASS) lexer would be useful for CUDA dumps?

It would be - I find that GNU Assembler syntax already works reasonably well for PTX in vim, so probably the `GasLexer` would be a good starting point: https://pygments.org/docs/lexers/#pygments.lexers.asm.GasLexer

PR #5334, opened 2020-03-02
Fix incorrect condition in tuple unboxing

eric-wieser:
As written, this would emit an `if` with no contents, and the cleanup would be unconditional, running even if the objects to be cleaned haven't been initialized.

After this change, the `if` is now emitted inside the cleanup block.

### DIFF
--- a/numba/core/boxing.py
+++ b/numba/core/boxing.py
@@ -547,8 +547,8 @@ def unbox_tuple(typ, obj, c):
     c.builder.store(value, value_ptr)
 
     if cleanups:
-        with c.builder.if_then(size_matches, likely=True):
-            def cleanup():
+        def cleanup():
+            with c.builder.if_then(size_matches, likely=True):
                 for func in reversed(cleanups):
                     func()
     else:


eric-wieser:
Based on discussion in #5336, I no longer think this change is correct.

numba/core/boxing.py
```diff
@@ -547,8 +547,8 @@ def unbox_tuple(typ, obj, c):
     c.builder.store(value, value_ptr)
 
     if cleanups:
-        with c.builder.if_then(size_matches, likely=True):
-            def cleanup():
+        def cleanup():
+            with c.builder.if_then(size_matches, likely=True):
                 for func in reversed(cleanups):
                     func()
```
eric-wieser: This change would remove the nonsense, but leave getting the cleanup right to a future PR:
```suggestion
        def cleanup():
            for func in reversed(cleanups):
                func()
```

sklam:
@eric-wieser, 

> Based on discussion in #5336, I no longer think this change is correct.

Are you abandoning this PR?

sklam:
I took a deeper look. I don't think the testsuite is no longer exercising the path that `size_matches` is `False`.  The error would only occur if the function type signature is invalid; for instance:

```python
from numba import njit


@njit
def foo(a):
    return len(a)


foo((1, 2))

[cr] = foo.overloads.values()
print(cr.entry_point((1, 2, 3)))  # the signature is for a 2-tuple but we are giving it a 3-tuple.
```

sklam:
And, I can confirm that if `size_matches` is `False`, the `cleanup` will not be invoked.

eric-wieser:
Closing in favor of gh-5417, which I worked on without seeing this discussion.

Issue #5335, opened 2020-03-02
Memory leak in intrinsic with temporary list object 


ARF1:
On the current master. Not sure whether this is a bug or whether I am doing something wrong.

Working with temporary lists in an intrinsic leads to a memory leak I cannot track down.

This is obviously some pointless piece of code. It is intended only as a minimal reproducer:

```python
import unittest

from numba import njit
from numba.tests.support import MemoryLeakMixin, TestCase
from numba.extending import intrinsic
from numba.core import typing, types


@intrinsic
def foo(typingctx):
    dtype = types.int64
    rettype = dtype
    sig = rettype()

    def make_list():
        out = []
        out.append(1)
        out.append(2)
        return out

    def get_list_element(a, i):
        return a[i]

    def codegen(context, builder, signature, args):
        # Generate a temporary list
        inner_argtypes = []
        inner_rettype = types.List(dtype)
        inner_sig = typing.signature(inner_rettype, *inner_argtypes)
        inner_args = []
        tmp_list = context.compile_internal(
            builder, make_list, inner_sig, inner_args)

        # get a single element from the list
        list_idx = context.get_constant(types.intp, 0)
        data = context.compile_internal(builder, get_list_element,
                                        dtype(inner_rettype, types.intp),
                                        [tmp_list, list_idx])

        # decref anything that can possibly be freed
        if context.enable_nrt:
            context.nrt.decref(builder, inner_rettype, tmp_list)
            context.nrt.decref(builder, dtype, data)

        return data

    return sig, codegen


class TestLeak(MemoryLeakMixin, TestCase):
    def test_leak(self):
        @njit
        def bar():
            return foo()
        results = bar()


if __name__ == '__main__':
    unittest.main()
```

Output:
```
    self.assertEqual(total_alloc, total_free)
AssertionError: 2 != 0

----------------------------------------------------------------------
Ran 1 test in 0.339s

FAILED (failures=1)
```

sklam:
Likely a bug with the old reflected list implementation. I would recommend using the new typed-list, instead. We have planned to remove the reflected-list.

sklam:
Hmm... there's still 1 not decref'ed reference with `typed.List`.

ARF1:
@sklam Thanks for confirming that this is not because of something I am doing wrong...

sklam:
This is my current reproducer. 

```python
import unittest

from numba import njit
from numba.tests.support import MemoryLeakMixin, TestCase
[...]

PR #2352, opened 2017-04-26
Add refcount pruning pass for function that does not need refcount

sklam:
This adds a new optimization pass at the post-lowering stage to use high-level type info to determine if a function needs refcount.  If determined to be unneeded, all refcount ops are removed from the function being compiled.

Functions that does not need refcount:

* do not take any refcounted object as arguments, except Array;
* do not return any refcounted object;
* do not call any function that return refcounted object;

Issue #2350, opened 2017-04-25
refcount op in branching code affects performance


sklam:
The refcount ops emitted by this code is splitted across basicblocks.  Numba refcount pruning pass is not able to prune them.

```python
import numba
import numpy as np
from timeit import default_timer as timer


def bench(nrt):
    print("set _nrt=%s" % nrt)

    jit = numba.jit(nopython=True, nogil=True, _nrt=nrt)

    @jit
    def _append_non_na(x, y, agg, field):
        if not np.isnan(field):
            agg[y, x] += 1

    @jit
    def _append(x, y, agg, field):
        if not np.isnan(field):
            if np.isnan(agg[y, x]):
                agg[y, x] = field
            else:
                agg[y, x] += field

    @jit
    def append(x, y, agg, field):
        _append_non_na(x, y, agg, field)
        _append(x, y, agg, field)


    @jit
    def extend(arr, field):
        for i in range(arr.shape[0]):
            for j in range(arr.shape[1]):
                append(j, i, arr, field)


    arr = np.ones((10, 100))

    for i in range(10):
        ts = timer()
        extend(arr, 10)
        te = timer()
        print("%.5fms" % ((te - ts) * 1000))

    # extend.inspect_cfg(extend.signatures[0]).display(view=True)


bench(nrt=True)
bench(nrt=False)
```

Output
```
set _nrt=True
224.90708ms
0.02676ms
0.02478ms
0.02456ms
0.02460ms
0.02464ms
0.02724ms
0.02477ms
0.02457ms
0.02448ms
set _nrt=False
157.52096ms
0.00259ms
0.00072ms
0.00065ms
0.00062ms
0.00062ms
0.00065ms
0.00062ms
0.00065ms
0.00066ms
```

sklam:
With the patch #2352, the new numbers are:

```
set _nrt=True
259.69325ms
0.00284ms
0.00110ms
0.00089ms
0.00083ms
0.00080ms
0.00082ms
0.00080ms
0.00083ms
0.00081ms
set _nrt=False
156.33411ms
0.00239ms
0.00079ms
0.00066ms
0.00065ms
0.00063ms
0.00067ms
0.00061ms
0.00065ms
0.00061ms
```

PR #5338, opened 2020-03-02
Remove refcount pruning pass

sklam:
because it is no longer valid. The pass was introduced in https://github.com/numba/numba/pull/2352 to avoid NRT-ops if the function doesn't need it. However, it's no longer possible to detect; at least with the current way we emit the LLVM-IR.

This fixes https://github.com/numba/numba/issues/5335, which provides a use-case where the `@intrinsic` added a hidden use of NRT allocation not observable from the pruning pass.


sklam:
Note, the consequences of this PR is that https://github.com/numba/numba/issues/2350 will become a problem again

sklam:
Some tests are failling. I think the pass was hiding refcount errors in simple array tests.

stuartarchibald:
Performance vs leaks, I'd rather stop it leaking and dedicate time to an LLVM pass?

sklam:
https://github.com/numba/numba/pull/5338#issuecomment-593966746

agree

stuartarchibald:
Assuming this passes CI, this looks good.

stuartarchibald:
> [#5338 (comment)](https://github.com/numba/numba/pull/5338#issuecomment-593966746)
> 
> agree

Right, let's proceed with acknowledgement that we do correctness over performance, and we know how to fix the performance. Schedule that for Q2.

PR #5340, opened 2020-03-03
[WIP] Package serialization functions into a pickleable class

eric-wieser:
All the callers of these helpers seem to behave in exactly the same way, and this simplifies things a little.

Also, I think we'll want to use this class for `@jit` when `NUMBA_DISABLE_JIT=1`.


numba/core/extending.py
```diff
@@ -309,25 +310,16 @@ def __repr__(self):
...
     @classmethod
-    def _rebuild(cls, uuid, name, defn_reduced):
+    def _rebuild(cls, uuid, name, defn):
         from numba.core import serialize
 
         try:
             return cls._memo[uuid]
         except KeyError:
-            defn = serialize._rebuild_function(*defn_reduced)
-
```
eric-wieser: Note that this means that `rebuild_function` is no longer evaluated lazily, which may or may not matter.

PR #5343, opened 2020-03-03
Fix cuda spoof

c200chromebook:
Far better fidelity simulation of numpy records in cudasim mode. Essentially, just flips everything to __getitem__ if in kernels, unless a __getitem__ exists. A few little whitespace fixes too.

### DIFF
--- a/numba/cuda/simulator/cudadrv/devicearray.py
+++ b/numba/cuda/simulator/cudadrv/devicearray.py
@@ -72,7 +72,7 @@ class FakeCUDAArray(object):
     __cuda_ndarray__ = True  # There must be gpu_data attribute
 
     def __init__(self, ary, stream=0):
-        if ary.ndim == 0:
+        if ary.ndim == 0:  # needed for TestCudaNDArray.test_device_array_interface
             self._ary = ary.reshape(1)
             self._ary_access = self._ary[0]
         else:
--- a/numba/cuda/simulator/cudadrv/devicearray.py
+++ b/numba/cuda/simulator/cudadrv/devicearray.py
@@ -47,7 +47,7 @@ class FakeKernelCUDAArray(object):
 
     def __getattr__(self, attrname):
         if attrname in dir(self.__dict__['_item']._ary):  # For, eg, array size.
-            return getattr(self.__dict__['_item']._ary, attrname)
+            return self.__wrap_if_fake(getattr(self.__dict__['_item']._ary, attrname))
         else:
             return self.__wrap_if_fake(self._item.__getitem__(attrname))
 
--- a/numba/cuda/simulator/cudadrv/devicearray.py
+++ b/numba/cuda/simulator/cudadrv/devicearray.py
@@ -46,8 +46,8 @@ class FakeKernelCUDAArray(object):
         return FakeKernelCUDAArray(item) if isinstance(item, FakeCUDAArray) else item
 
     def __getattr__(self, attrname):
-        if attrname in dir(self.__dict__['_item']._ary):  # For, eg, array size.
-            return self.__wrap_if_fake(getattr(self.__dict__['_item']._ary, attrname))
+        if attrname in dir(self._item._ary):  # For, eg, array size.
+            return self.__wrap_if_fake(getattr(self._item._ary, attrname))
         else:
             return self.__wrap_if_fake(self._item.__getitem__(attrname))
 
@@ -61,7 +61,7 @@ class FakeKernelCUDAArray(object):
         self._item.__setitem__(idx, val)
 
     def __len__(self):
-        return len(self.__dict__['_item'])
+        return len(self._item)
 
 class FakeCUDAArray(object):
     '''
--- a/numba/cuda/simulator/cudadrv/devicearray.py
+++ b/numba/cuda/simulator/cudadrv/devicearray.py
@@ -47,14 +47,14 @@ class FakeKernelCUDAArray(object):
     def __getattr__(self, attrname):
         return self.__wrap_if_fake(self._item.__getitem__(attrname))
 
+    def __setattr__(self, nm, val):
+        self._item.__setitem__(nm, val)
+
     def __getitem__(self, idx):
         return self.__wrap_if_fake(self._item.__getitem__(idx))
 
     def __setitem__(self, idx, val):
         self._item.__setitem__(idx, val)
-
-    def __setattr__(self, nm, val):
-        self.__setattr__(nm, val)
 
 
 
--- a/numba/cuda/simulator/cudadrv/devicearray.py
+++ b/numba/cuda/simulator/cudadrv/devicearray.py
@@ -36,6 +36,29 @@ class FakeShape(tuple):
         return super(FakeShape, self).__getitem__(k)
 
 
+class FakeKernelCUDAArray(object):
+    def __init__(self, item):
+        assert isinstance(item, FakeCUDAArray)
+        self.__dict__['_item'] = item
+
+    def __wrap_if_fake(self,item):
+        return FakeKernelCUDAArray(item) if isinstance(item,FakeCUDAArray) else item
+
+    def __getattr__(self, attrname):
+        return self.__wrap_if_fake(self._item.__getitem__(attrname))
+
+    def __getitem__(self, idx):
+        return self.__wrap_if_fake(self._item.__getitem__(idx))
+
+    def __setitem__(self, idx, val):
+        self._item.__setitem__(idx, val)
+
+    def __setattr__(self, nm, val):
+        self.__setattr__(nm, val)
+
+
+
+
 class FakeCUDAArray(object):
     '''
     Implements the interface of a DeviceArray/DeviceRecord, but mostly just
@@ -63,10 +86,7 @@ class FakeCUDAArray(object):
             attr = getattr(self._ary, attrname)
             return attr
         except AttributeError as e:
-            try:
-                return self.__getitem__(attrname)
-            except AttributeError as e:
-                six.raise_from(AttributeError("Wrapped array has no attribute '%s'" % attrname), e)
+            six.raise_from(AttributeError("Wrapped array has no attribute '%s'" % attrname), e)
 
     def bind(self, stream=0):
         return FakeCUDAArray(self._ary, stream)
@@ -79,7 +99,11 @@ class FakeCUDAArray(object):
         return FakeCUDAArray(np.transpose(self._ary, axes=axes))
 
     def __getitem__(self, idx):
-        return FakeCUDAArray(self._ary.__getitem__(idx), stream=self.stream)
+        ret = self._ary.__getitem__(idx)
+        if np.issubdtype(type(ret), np.number) or np.issubdtype(type(ret),np.bool_):
+            return ret
+        else:
+            return FakeCUDAArray(self._ary.__getitem__(idx), stream=self.stream)
 
     def __setitem__(self, idx, val):
         return self._ary.__setitem__(idx, val)
--- a/numba/cuda/simulator/cudadrv/devicearray.py
+++ b/numba/cuda/simulator/cudadrv/devicearray.py
@@ -109,7 +109,7 @@ class FakeCUDAArray(object):
 
     def __getitem__(self, idx):
         ret = self._ary_access.__getitem__(idx)
-        if np.issubdtype(type(ret), np.number) or np.issubdtype(type(ret), np.bool_):
+        if np.issubdtype(type(ret), np.number) or np.issubdtype(type(ret), np.bool_) or np.issubdtype(type(ret), np.record):
             return ret
         else:
             return FakeCUDAArray(ret, stream=self.stream)
--- a/numba/cuda/simulator/cudadrv/devicearray.py
+++ b/numba/cuda/simulator/cudadrv/devicearray.py
@@ -39,9 +39,10 @@ class FakeShape(tuple):
 
 class FakeWithinKernelCUDAArray(object):
     '''
-    Created to emulate the behavior of arrays within kernels, where either array.item or array['item'] is valid.
-    This behaviour does not follow the semantics of Python and NumPy with
-    non-jitted code, and will be deprecated and removed.
+    Created to emulate the behavior of arrays within kernels, where either
+    array.item or array['item'] is valid. This behaviour does not follow
+    the semantics of Python and NumPy with non-jitted code, and will be
+    deprecated and removed.
     '''
 
     def __init__(self, item):
@@ -49,7 +50,10 @@ class FakeWithinKernelCUDAArray(object):
         self.__dict__['_item'] = item
 
     def __wrap_if_fake(self, item):
-        return FakeWithinKernelCUDAArray(item) if isinstance(item, FakeCUDAArray) else item
+        if isinstance(item, FakeCUDAArray):
+            return FakeWithinKernelCUDAArray(item)
+        else:
+            return item
 
     def __getattr__(self, attrname):
         if attrname in dir(self._item._ary):  # For e.g. array size.
@@ -69,6 +73,7 @@ class FakeWithinKernelCUDAArray(object):
     def __len__(self):
         return len(self._item)
 
+
 class FakeCUDAArray(object):
     '''
     Implements the interface of a DeviceArray/DeviceRecord, but mostly just
@@ -78,7 +83,8 @@ class FakeCUDAArray(object):
     __cuda_ndarray__ = True  # There must be gpu_data attribute
 
     def __init__(self, ary, stream=0):
-        if ary.ndim == 0:  # needed for TestCudaNDArray.test_device_array_interface
+        # ary/ary_access needed for TestCudaNDArray.test_device_array_interface
+        if ary.ndim == 0:
             self._ary = ary.reshape(1)
             self._ary_access = self._ary[0]
         else:
--- a/numba/cuda/simulator/cudadrv/devicearray.py
+++ b/numba/cuda/simulator/cudadrv/devicearray.py
@@ -109,7 +109,7 @@ class FakeCUDAArray(object):
 
     def __getitem__(self, idx):
         ret = self._ary_access.__getitem__(idx)
-        if np.issubdtype(type(ret), np.number) or np.issubdtype(type(ret), np.bool_) or np.issubdtype(type(ret), np.record):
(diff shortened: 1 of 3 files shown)


c200chromebook:
:(

c200chromebook:
Well this is fun. Like whack-a-mole.

stuartarchibald:
Some notes on Numba CI...

The Numba test suite causes CI systems a lot of grief:

1. It's huge, 9000+ tests.
2. In part because of 1. and that compilers are pretty involved, the test suite takes a long time to run.
3. There's sections of the test suite that are deliberately designed to stress systems almost to the point of failure (tests which concurrently compile and execute with threads and fork processes etc).
4. The combination of things that Numba has to test well exceeds the capacity of any public CI system, (Python versions x NumPy versions x Operating systems x Architectures x feature libraries (e.g. SVML) x threading backends (e.g. OpenMP, TBB)) and then there's CUDA and ROCm too and all their version variants.

As a result of the above, public CI is implemented as follows:
1. The combination of OS x Python x NumPy x Various Features in the testing matrix is designed to give a good indicative result for whether "this PR is probably ok".
2. When CI runs it:
    1. Looks for files that contain tests that have been altered by the proposed change and runs these on the whole testing matrix.
    2. Runs a subset of the test suite on each part of the testing matrix. i.e. slice the test suite up by the number of combinations in the testing matrix and each combination runs one chunk. This is done for speed, because public CI cannot cope with the load else.

The Numba project also has its own private build and test farm which actually exercises all the combinations on real hardware. If a PR is changing CUDA or ROCm code (which cannot be tested on Public CI as there's no hardware) or it is making changes to something that the core developers consider risky, then it will also be run on the Numba farm just to make sure!

c200chromebook:
In my estimation there's no risk to the core CUDA code as the only bits that are changing are tests and simulator.

leofang:
Thank you, @stuartarchibald, for sharing how your CI works! It's very impressive :)

numba/cuda/simulator/cudadrv/devicearray.py
```diff
@@ -30,23 +30,54 @@ class FakeShape(tuple):
...
+class FakeKernelCUDAArray(object):
+    def __init__(self, item):
+        assert isinstance(item, FakeCUDAArray)
+        self.__dict__['_item'] = item
+
+    def __wrap_if_fake(self, item):
+        return FakeKernelCUDAArray(item) if isinstance(item, FakeCUDAArray) else item
+
+    def __getattr__(self, attrname):
+        if attrname in dir(self.__dict__['_item']._ary):  # For, eg, array size.
```
eric-wieser: You can use `self._item` here, no need to go through `__dict__`.

numba/cuda/simulator/cudadrv/devicearray.py
```diff
@@ -30,23 +30,54 @@ class FakeShape(tuple):
...
+            return self.__wrap_if_fake(getattr(self.__dict__['_item']._ary, attrname))
```
eric-wieser: And here

numba/cuda/simulator/cudadrv/devicearray.py
```diff
@@ -30,23 +30,54 @@ class FakeShape(tuple):
...
+        self._item.__setitem__(nm, val)
+
+    def __getitem__(self, idx):
+        return self.__wrap_if_fake(self._item.__getitem__(idx))
+
+    def __setitem__(self, idx, val):
+        self._item.__setitem__(idx, val)
+
+    def __len__(self):
+        return len(self.__dict__['_item'])
```
eric-wieser: And here

numba/cuda/simulator/cudadrv/devicearray.py
```diff
@@ -30,23 +30,54 @@ class FakeShape(tuple):
...
     wraps a NumPy array.
     '''
 
-    __cuda_ndarray__ = True     # There must be gpu_data attribute
-
+    __cuda_ndarray__ = True  # There must be gpu_data attribute
 
     def __init__(self, ary, stream=0):
-        self._ary = ary.reshape(1) if ary.ndim == 0 else ary
+        if ary.ndim == 0:
```
eric-wieser: Can you add a comment explaining why this case is needed?

numba/cuda/simulator/cudadrv/devicearray.py
c200chromebook: Can you though? Because this is in self's __getattr__ function and if you ask self for '_item', it'll go right back to where it is.

eric-wieser: getattr is not called for entries present in `__dict__` (unlike setitem which is)

numba/cuda/simulator/cudadrv/devicearray.py
c200chromebook: Done.

numba/cuda/simulator/cudadrv/devicearray.py
c200chromebook: done

numba/cuda/simulator/cudadrv/devicearray.py
```diff
@@ -30,23 +30,54 @@ class FakeShape(tuple):
...
     negative indexing)
     '''
+
     def __getitem__(self, k):
         if isinstance(k, int) and k < 0:
             raise IndexError('tuple index out of range')
         return super(FakeShape, self).__getitem__(k)
 
 
+class FakeKernelCUDAArray(object):
```
gmarkall: This class could do with a docstring explaining why it does what it does, similar to how `FakeCUDAArray` has one.

numba/cuda/simulator/cudadrv/devicearray.py
```diff
@@ -30,23 +30,54 @@ class FakeShape(tuple):
...
+    def __init__(self, item):
+        assert isinstance(item, FakeCUDAArray)
+        self.__dict__['_item'] = item
+
+    def __wrap_if_fake(self, item):
```
gmarkall: A minor point, but it seems strange that this method begins with a double underscore but isn't a dunder method.

eric-wieser: This is probably deliberate - it results in a mangled name of `_FakeKernelCUDAArray__wrap_if_fake`, which ensures it does not conflict with any attributes of `self._item`

c200chromebook:
Completely agree wrt burninating this weird behavior. However, think it should be supported for now, as is very handy for development on kernels of unusual size.

numba/cuda/simulator/cudadrv/devicearray.py
c200chromebook: yup

numba/cuda/simulator/cudadrv/devicearray.py
c200chromebook: added

numba/cuda/simulator/cudadrv/devicearray.py
```diff
@@ -30,23 +30,59 @@ class FakeShape(tuple):
...
+
     def __getitem__(self, k):
         if isinstance(k, int) and k < 0:
             raise IndexError('tuple index out of range')
         return super(FakeShape, self).__getitem__(k)
 
 
+class FakeWithinKernelCUDAArray(object):
+    '''
+    Created to emulate the behavior of arrays within kernels, where either array.item or array['item'] is valid.
```
gmarkall: As far as I can tell from the code, this is to emulate the behaviour of structured types or records in kernels - if an array is passed in to `FakeCUDAArray.__init__`, then a `FakeWithinKernelCUDAArray` is not created, but the original array is used for its `_ary_access` attribute.

Am I misunderstanding what's happening here?

numba/cuda/simulator/cudadrv/devicearray.py
```diff
@@ -30,23 +30,59 @@ class FakeShape(tuple):
...
+    This is weird behavior and should eventually be removed.
```
gmarkall:
```suggestion
    This behaviour does not follow the semantics of Python and NumPy with
    non-jitted code, and will be deprecated and removed.
```

numba/cuda/simulator/cudadrv/devicearray.py
c200chromebook: I think so. Note that a FakeCUDAArray becomes a FakeWithinKernelCUDAArray in a FakeCUDAKernel. This emulates the behavior that you can only say `array.item` or `array['item']` within a kernel, and not outside.

numba/cuda/simulator/kernel.py
```diff
@@ -86,11 +86,16 @@ def fake_arg(arg):
...
                 elif isinstance(arg, ArgHint):
-                    return arg.to_device(retr)
+                    ret = arg.to_device(retr)
+                elif isinstance(arg, np.void):
+                    ret = FakeCUDAArray(arg)  # In case a np record comes in.
                 else:
-                    return arg
+                    ret = arg
+                if isinstance(ret,FakeCUDAArray):
+                    return FakeWithinKernelCUDAArray(ret)
```
c200chromebook: This is where FakeWithinKernelCUDAArrays get made - when you ship arguments to a kernel.

gmarkall: Got it, I somehow glossed over this multiple times - thanks!

gmarkall:
This looks good to me.

numba/cuda/simulator/cudadrv/devicearray.py
stuartarchibald: This line is too long, please wrap to 80 chars.

numba/cuda/simulator/cudadrv/devicearray.py
```diff
@@ -30,23 +30,60 @@ class FakeShape(tuple):
...
         if isinstance(k, int) and k < 0:
             raise IndexError('tuple index out of range')
         return super(FakeShape, self).__getitem__(k)
 
 
+class FakeWithinKernelCUDAArray(object):
+    '''
+    Created to emulate the behavior of arrays within kernels, where either array.item or array['item'] is valid.
+    This behaviour does not follow the semantics of Python and NumPy with
+    non-jitted code, and will be deprecated and removed.
```
stuartarchibald: Who is tracking and managing this deprecation?

numba/cuda/simulator/kernel.py
```diff
@@ -86,11 +86,16 @@ def fake_arg(arg):
...
+                    ret = wrap_arg(arg).to_device(retr)
                 elif isinstance(arg, ArgHint):
-                    return arg.to_device(retr)
+                    ret = arg.to_device(retr)
+                elif isinstance(arg, np.void):
+                    ret = FakeCUDAArray(arg)  # In case a np record comes in.
                 else:
-                    return arg
+                    ret = arg
+                if isinstance(ret,FakeCUDAArray):
```
stuartarchibald:
```suggestion
                if isinstance(ret, FakeCUDAArray):
```

numba/cuda/tests/cudasim/test_cudasim_issues.py
```diff
@@ -1,14 +1,49 @@
 import threading
 
 import numpy as np
+from numpy import float64
 
-from numba import cuda
+from numba import cuda, numba
```
stuartarchibald: Am I reading this right, implies `from numba import numba`?

numba/cuda/tests/cudasim/test_cudasim_issues.py
```diff
@@ -1,14 +1,49 @@
 import threading
 
 import numpy as np
+from numpy import float64
```
stuartarchibald: Convention is to use `float64` from the `np` module directly in code e.g. `np.array([1,], dtype=np.float64)`, Numba also has a `numba.float64`.

numba/cuda/tests/cudasim/test_cudasim_issues.py
```diff
@@ -1,14 +1,49 @@
...
+    BACKYARD_TYPE = [('statue', float64),
+                     ('newspaper', float64, (6, ))]
+
+    GOOSE_TYPE = [('garden', float64, (12,)),
+                  ('town', float64, (42,)),
+                  ('backyard', BACKYARD_TYPE)]
+
+    GOOSE_NP_DTYPE = np.dtype(GOOSE_TYPE, align=True)
+
+    GOOSE_NB_DTYPE = numba.from_dtype(GOOSE_NP_DTYPE)
```
stuartarchibald: Seems like these are only used in https://github.com/numba/numba/pull/5343/files#diff-7dbb9cebfbec424ce906ec5ad437908aR24 and so should go in the scope of that function?

numba/cuda/simulator/kernel.py
```diff
@@ -6,7 +6,7 @@
 import numpy as np
 
 from numba.core.utils import reraise
-from .cudadrv.devicearray import to_device, auto_device
+from .cudadrv.devicearray import to_device, auto_device, FakeCUDAArray, FakeWithinKernelCUDAArray
```
stuartarchibald: Please wrap line to 80 chars

numba/cuda/simulator/cudadrv/devicearray.py
```diff
@@ -30,23 +30,60 @@ class FakeShape(tuple):
...
+
+    def __init__(self, item):
+        assert isinstance(item, FakeCUDAArray)
+        self.__dict__['_item'] = item
+
+    def __wrap_if_fake(self, item):
+        return FakeWithinKernelCUDAArray(item) if isinstance(item, FakeCUDAArray) else item
+
+    def __getattr__(self, attrname):
+        if attrname in dir(self._item._ary):  # For, eg, array size.
```
stuartarchibald:
```suggestion
        if attrname in dir(self._item._ary):  # For e.g. array size.
```

numba/cuda/simulator/cudadrv/devicearray.py
```diff
@@ -30,23 +30,60 @@ class FakeShape(tuple):
...
+    This behaviour does not follow the semantics of Python and NumPy with
+    non-jitted code, and will be deprecated and removed.
+    '''
+
+    def __init__(self, item):
+        assert isinstance(item, FakeCUDAArray)
+        self.__dict__['_item'] = item
+
+    def __wrap_if_fake(self, item):
+        return FakeWithinKernelCUDAArray(item) if isinstance(item, FakeCUDAArray) else item
```
stuartarchibald: Line wrap to 80 chars please.

numba/cuda/simulator/cudadrv/devicearray.py
gmarkall: I'm planning to submit a PR deprecating it, but not for 0.49

stuartarchibald: Thanks for confirming. Is there an issue tracking it? If not please could one be created, I'll add it to the 0.50 milestone.

gmarkall: (It was me that suggested that text in the docstring originally, btw)

numba/cuda/tests/cudasim/test_cudasim_issues.py
c200chromebook: Will swap to something better.

numba/cuda/simulator/cudadrv/devicearray.py
gmarkall: https://github.com/numba/numba/issues/5375

numba/cuda/simulator/cudadrv/devicearray.py
```diff
@@ -30,23 +30,66 @@ class FakeShape(tuple):
...
     def __getitem__(self, k):
         if isinstance(k, int) and k < 0:
             raise IndexError('tuple index out of range')
         return super(FakeShape, self).__getitem__(k)
 
 
+class FakeWithinKernelCUDAArray(object):
+    '''
+    Created to emulate the behavior of arrays within kernels, where either
+    array.item or array['item'] is valid. This behaviour does not follow
```
eric-wieser: Can we add a 
```suggestion
    Created to emulate the behavior of arrays within kernels, where either
    array.item or array['item'] is valid (that is, give all structured
    arrays `numpy.recarray`-like semantics). This behaviour does not follow
```

numba/cuda/simulator/cudadrv/devicearray.py
```diff
@@ -30,23 +30,66 @@ class FakeShape(tuple):
...
 
     def __init__(self, ary, stream=0):
-        self._ary = ary.reshape(1) if ary.ndim == 0 else ary
+        # ary/ary_access needed for TestCudaNDArray.test_device_array_interface
+        if ary.ndim == 0:
+            self._ary = ary.reshape(1)
+            self._ary_access = self._ary[0]
+        else:
+            self._ary = ary
+            self._ary_access = self._ary
```
eric-wieser: Note that if #5359 goes in first then this will need changing, or vice versa.

numba/cuda/simulator/cudadrv/devicearray.py
c200chromebook: isn't that implicit? What's the other case?

numba/cuda/simulator/cudadrv/devicearray.py
c200chromebook: Honestly looking forward to the deprecation of all this weirdness, a lot of this is accidental complexity.

numba/cuda/simulator/cudadrv/devicearray.py
gmarkall: It's implicit, but it's perhaps not obvious if you're not familiar in detail with structured arrays and recarrays - I think it's helpful to make the distinction for clarity.

c200chromebook: works for me

eric-wieser: The point I'd like to see is that the semantics aren't absent from numpy, they're just usually opt-in.

stuartarchibald:
Thanks for the fixes. Looks good to me. Other reviewers happy?

Buildfarm ID: `numba_smoketest_cuda_37`.

stuartarchibald:
Buildfarm passed.

gmarkall:
I'm happy :-)

stuartarchibald:
Thanks for confirming @gmarkall, marking for merge.

PR #5345, opened 2020-03-04
Disable test failing due to removed pass.

stuartarchibald:
The test "test_refct_pruning_with_branches" now fails as #5338
removed the pass that made it work as it was discovered the
pass itself was buggy.

This patch adds a skip on the failing test but does not remove
it as it is anticipated that in the future the functionality
required for the test to pass will be implemented.


stuartarchibald:
xref: #2350

PR #5346, opened 2020-03-04
Docs: use sphinx_rtd_theme

gmarkall:
There was some discussion about trying another theme for the docs in the Numba dev meeting a few weeks ago - this PR is an experiment with the [RTD theme](https://sphinx-rtd-theme.readthedocs.io/en/stable/index.html) to see what people think. An example of the generated documentation is at:

https://gmarkall.github.io/numba-rtd-theme/

If this is liked / preferred, the changes in this PR are complete / ready for review - there are no outstanding changes I'd make to get this ready to go.

A summary of the changes:

- Modify build scripts to conda install sphinx_rtd_theme instead of pip installing sphinx_bootstrap_theme.
- Add SVG version of logo - remove PNG version. This enables the theme to scale the logo better.
- Also use SVG icon as favicon.
- Remove custom CSS - no longer needed.
- Update copyright notice in Sphinx configuration.
- Modify Sphinx config to use sphinx_rtd_theme and add relevant options.


seibert:
This theme looks pretty reasonable to me.  I'd be in favor of it.

leofang:
I like this style! One minor thing: this rendering looks weird
螢幕快照 2020-03-04 下午2 14 25

esc:
I think it looks good too! One stylistic change I would suggest, if possible, is to use a white Numba logo on a Numba blue background. Right now, we have a Numba blue logo on a default theme blue background and it looks a bit... off..
Screen Shot 2020-03-04 at 20 49 33

gmarkall:
Thanks for the comments!

> I like this style! One minor thing: this rendering looks weird

I think there's an error in the source (a missing empty line between the nesting levels of the list) - will push a fix shortly.

> One stylistic change I would suggest, if possible, is to use a white Numba logo on a Numba blue background.

Will try this out too and push a new version to the example Github Pages.

stuartarchibald:
I find that when each page loads I get the Numba logo flashing up taking up 50% of the screen, and then it gets squashed into the corner and the rest of the page loads.

gmarkall:
@esc Thanks for the suggestion - I've now implemented white-on-blue.
@leofang The list should look better now.
@stuartarchibald I see this too - looking into a good fix for this issue. This may be related: https://github.com/readthedocs/sphinx_rtd_theme/pull/848

Changes should be visible at https://gmarkall.github.io/numba-rtd-theme/index.html

esc:
@gmarkall thanks, it looks very neat!

gmarkall:
@stuartarchibald I think the flickering up of the Numba logo in Firefox is now fixed - does it also look better for you now?

stuartarchibald:
@gmarkall thanks, I can confirm the reflow problem is gone AFAICT.

stuartarchibald:
I've just been checking a bunch of pages, one notable disadvantage of RTD for numba docs is due to the amount of text boxes which are rather wide by virtue of them containing things generated from numba internals. e.g. IR dumps from various languages/terminal dumps. The result is just more boxes with horizontal scroll bars. Is there perhaps some option to make the RTD sidebar fold up?

gmarkall:
@stuartarchibald I couldn't find a way to fold it, but a bit of CSS can override the max width of the content - I've tried it with 1200px... What do you think of how it looks now?

docs/source/conf.py
```diff
@@ -118,20 +117,28 @@
...
-  'bootswatch_theme': "paper",
+    'canonical_url': 'http://numba.pydata.org/numba-doc/latest/index.html',
+    # 'logo_only': False,
+    # 'display_version': True,
+    # 'prev_next_buttons_location': 'bottom',
+    'style_external_links': True,
+    # 'vcs_pageview_mode': '',
+    'style_nav_header_background': '#00A3E0',
+    # Toc options
+    # 'collapse_navigation': True,
```
asmeurer: I would consider making this False. That enables digging into the TOC in the sidebar. According to the note at https://sphinx-rtd-theme.readthedocs.io/en/latest/configuring.html, if your TOC has a lot of entries it can create very large files, so you might check that first.

asmeurer:
Nice. The See Also sections here are readable now! https://gmarkall.github.io/numba-rtd-theme/user/threading-layer.html

docs/source/conf.py
gmarkall: Done - the gh-pages seem to take a few minutes to update, so the changes might not immediately be visible, but should be there soon.

gmarkall: They seem to have updated now - load time and performance does not appear to be adversely affected.

stuartarchibald:
Core devs: any objections to moving to this doc template for 0.49? It seems to fix a lot of long standing annoyances. Whilst unlikely, if there is some catastrophe we can always switch back and rebuild. CC @sklam @seibert @esc @ehsantn @DrTodd13 .

stuartarchibald:
> @stuartarchibald I couldn't find a way to fold it, but a bit of CSS can override the max width of the content - I've tried it with 1200px... What do you think of how it looks now?

In general looks good, thanks. There's still the odd minor thing, for example, highlighting doesn't extend to full text width, see last source box in https://gmarkall.github.io/numba-rtd-theme/developer/inlining.html#example-using-numba-extending-overload but this has nothing like the readability/usability issues that the current template has, so I think this is a large net gain.

I'm of the view that we should switch to this.

luk-f-a:
my two cents: the 4-levels-deep numbered index is too heavy visually, eg `1.1.7.1 GPU Targets`. I would get rid of the numbers completely (like dask : https://docs.dask.org/) or limit them to two levels, ie 1. User Manual, 1.1 5 minutes guide, and everything below goes unnumbered.

sklam:
I like the new template. I am in favor of switching. We can continue to work on the minor styling issues once it's merged.

esc:
Yes, let's merge now and then iterate further.

stuartarchibald:
Thanks all, marking for merge.

Issue #5137, opened 2020-01-23
Support per-thread default streams in CUDA


gmarkall:
Following the merge of https://github.com/numba/numba/pull/5136, it might be useful to have support for the per-thread default-stream option (reference: https://devblogs.nvidia.com/gpu-pro-tip-cuda-7-streams-simplify-concurrency/)

I think this wouldn't be a large effort to implement - opening this issue to note that it could be implemented and to gauge interest.

Are there any use cases this would unblock for anyone?

leofang:
This is interesting. I checked the documentation and saw [this](https://docs.nvidia.com/cuda/cuda-runtime-api/stream-sync-behavior.html#stream-sync-behavior):
> The per-thread default stream is not a non-blocking stream and will synchronize with the legacy default stream if both are used in a program.

So I suppose the interplay between two libraries built with different `--default-stream` options, say CuPy (on `legacy`) and Numba (on `per-thread` with this proposal), is safe? Am I right that there will be a synchronization before a CuPy array on the `legacy` default stream being passed to a Numba kernel on the `per-thread` default stream?

gmarkall:
> So I suppose the interplay between two libraries built with different `--default-stream` options, say CuPy (on `legacy`) and Numba (on `per-thread` with this proposal), is safe? Am I right that there will be a synchronization before a CuPy array on the `legacy` default stream being passed to a Numba kernel on the `per-thread` default stream?

When you say "a CuPy array on the `legacy` default stream", are you referring to a scenario like:

1. A `cuMemcpy<something>Async` with the array as the destination on the legacy default stream, then
2. The array immediately being passed to a kernel launched on a per-thread default stream?

If so, my understanding is that the interplay between them would be safe (though I would specifically be testing this scenario as part of implementing per-thread default stream support in Numba).

If I've misunderstood - could you explain a little more about the scenario you're thinking of please?

leofang:
Yes, this is more or less what I had in mind, except that in CuPy a GPU array typically does not need memcpy to initialize, so it can be as simple as
```python
import cupy as cp
from numba import cuda

@cuda.jit
def some_func(x):
   # ...

a = cp.arange(0, 100).astype(cp.float64) # <- lives on "legacy" default stream
some_func[(1,), (100,)](a)               # <- may live on "per-thread" default stream
```
In this case, CuPy internally launches a ufunc kernel to initialize the array.

On the testing:
> though I would specifically be testing this scenario as part of implementing per-thread default stream support in Numba

[...]

Issue #4797, opened 2019-11-06
CUDA API interoperability


leofang:
Context: https://gitter.im/numba/numba?at=5da0d43a89acff6ff5db56eb

A question was raised recently for how to use a stream from CuPy in Numba's jit kernel. I also happened to run into the same question myself, and this was the workaround I had:
https://gist.github.com/leofang/4a043e5d94b4702d04fde2b9e7dcebbd
The `stream_cupy_to_numba()` function there allows the following simple usage:
```python
from numba import cuda
import cupy as cp

a = cp.random.random(100)
cp_stream = cp.cuda.stream.Stream()
nb_stream = stream_cupy_to_numba(cp_stream)
# ...define a @cuda.jit kernel...
kernel[grid, block, nb_stream](a)
```

However, it might be good for Numba to define a common interface to interoperate with other GPU libraries. First, we need a list of CUDA objects to be supported, and I think this list can be simple as this:
- device
- stream
- event

Second, in Numba one could initialize the CUDA object by accepting a 3rd party counterpart. Taking streams as example, this could be made valid I suppose:
```python
# passing the raw stream pointer to initialize a Numba stream  
# also need to tie the lifetime to the foreign object
nb_stream = cuda.stream(cp_stream.ptr, cp_stream)
```

(The title may not convey the essence of this task, so please feel free to reword it.)

seibert:
This connects to a long standing idea / request I've had, which is that I would love to see an NVIDIA-managed simple Python wrapper around the CUDA Driver API.   (I've mentioned this to anyone who would listen.  Ping @kkraus14.)  Once we start talking about devices and streams (where I think different projects have no need for innovation), I think interfaces might be overkill, and dodge the problem that we have redundant implementations of these things everywhere.

A `cuda` Python package would create a standard set of classes to represent things like CUDA Devices, Streams, Events, etc and wrappers around the standard CUDA driver functions that operate on these things.  Then all the different CUDA Python projects could depend on this package, and have interoperable CUDA entities.  As a bonus, Numba could delete its wrappers around these things.

Note that this package ideally needs to be fairly small and basic: Smaller than Numba for sure, and even smaller than PyCUDA.  Wrappers based on ctypes would even allow the package to be pure Python, although Cython wrappers might be preferable for other reasons.

leofang:
[...]

PR #5347, opened 2020-03-04
CUDA: Provide more stream constructors

gmarkall:
This commit adds constructors for:

- The legacy default stream
- The per-thread default stream
- Streams allocated externally (discussed in issue #4797)

The first two of these enable an explicit choice of the type of default stream - this is in contrast to `default_stream()` which uses the legacy or per-thread default stream depending on which API is in use. At the moment Numba always uses the legacy default stream APIs, but this may change in the future with the addition of an option of which API to use (see Issue #5137).

Documentation and tests are also added. On my machine (Tesla V100), CUDA tests are all passing:

```
$ python -m numba.runtests numba.cuda.tests
<lots of noise elided>
....
Ran 567 tests in 137.158s
```


docs/source/cuda-reference/host.rst
```diff
@@ -149,8 +154,14 @@ To get the default stream:
...
-.. autoclass:: numba.cuda.cudadrv.driver.Stream
-   :members: synchronize, auto_synchronize
+.. autofunction:: numba.cuda.legacy_default_stream
+
+.. autofunction:: numba.cuda.per_thread_default_stream
+
+To construct a Numba ``Stream`` object using a stream allocated elsewhere, the
+``external_stream`` function is provided:
 
+.. autofunction:: numba.cuda.external_stream
```
leofang: I think it'd be nice to explicitly state that it's users' responsibility to manage the lifetime of any external stream. 

ref: cupy/cupy#3141

leofang:
cc: @emcastillo, this is a parallel work of your cupy/cupy#3141

docs/source/cuda-reference/host.rst
gmarkall: Good point - I had overlooked this.

gmarkall:
> I also have one quick question, not sure if it's relevant here: does it matter in Numba if the external stream is created with the non-blocking flag (`CU_STREAM_NON_BLOCKING` for driver and `cudaStreamNonBlocking` for runtime API)?

As far as I can tell it doesn't matter directly in Numba, but the user will have to be aware of the semantics that it implies (i.e. not synchronizing with the default stream) and be aware of any concurrency issues that it could create.

docs/source/cuda-reference/host.rst
gmarkall: @leofang I've now updated the documentation.

leofang: Thanks, Graham!

stuartarchibald:
@gmarkall is this ready for a run through the build farm?

gmarkall:
@stuartarchibald I believe so, yes.

stuartarchibald:
> @stuartarchibald I believe so, yes.

Build ID: `numba_smoketest_cuda_32`.

stuartarchibald:
Build passed on 534b89b.

gmarkall:
Many thanks - is it still "Pending Buildfarm" for anything else?

numba/cuda/cudadrv/driver.py
```diff
@@ -894,12 +894,28 @@ def unload_module(self, module):
...
         driver.cuStreamCreate(byref(handle), 0)
         return Stream(weakref.proxy(self), handle,
                       _stream_finalizer(self.deallocations, handle))
 
+    def create_external_stream(self, ptr):
+        if not isinstance(ptr, int):
+            raise TypeError("ptr for external stream must be an int")
+        handle = drvapi.cu_stream(ptr)
+        return Stream(weakref.proxy(self), handle,
+                      _stream_finalizer(self.deallocations, handle),
```
gmarkall: Just realised this line is wrong - external streams should not have a Numba stream finalizer, because it will try to destroy them even though it doesn't own them. Will fix...

leofang: I thought this line means when the CUDA context is about to be destroyed, the stream will be destroyed? I suppose this should be OK since the external library will share the same CUDA context with Numba?

gmarkall: The external library (e.g. CuPy) will probably want to call `cuStreamDestroy` on it - with the `_stream_finalizer`, Numba will also call `cuStreamDestroy`, which results in the `CUDA_ERROR_INVALID_HANDLE` error.

gmarkall: In other words - the user will be managing the lifetime of the stream, and Numba shouldn't be.

leofang: Ah I see...Don't know what I was thinking 😅 Good catch, Graham!

gmarkall: Thanks... Though I was originally made aware of the issue because I pointed someone to https://gist.github.com/leofang/4a043e5d94b4702d04fde2b9e7dcebbd who used `stream_cupy_to_numba` in their code and found that the double-deletion of streams occurring. They removed the finalizer ( https://gist.github.com/leofang/4a043e5d94b4702d04fde2b9e7dcebbd#file-stream_cupy_to_numba-py-L28 ) and then the error was resolved.

leofang: Ahhh sorry for the buggy gist 🤦 I am scratching my head and wondering why I didn't encounter this bug so far...

gmarkall: One possibility is that the Numba finalizer could have only been called at exit time in your scenario, which would result in a traceback being printed, but wouldn't cause abnormal program termination - could that be what's happening?

numba/cuda/cudadrv/driver.py
```diff
@@ -1735,10 +1751,18 @@ def __int__(self):
...
+            2: "<Per-thread default CUDA stream on %s>",
+        }
+        ptr = self.handle.value or 0
+        if ptr in default_streams:
+            return default_streams[ptr] % self.context
+        elif self.external:
+            return "<External CUDA stream %d on %s>" % (ptr, self.context)
         else:
-            return "<Default CUDA stream on %s>" % self.context
+            return "<CUDA stream %d on %s>" % (ptr, self.context)
```
stuartarchibald, line 1765: hows this reached?

numba/cuda/cudadrv/driver.py
```diff
@@ -1735,10 +1751,18 @@ def __int__(self):
         return self.handle.value or 0
 
     def __repr__(self):
-        if self.handle.value:
-            return "<CUDA stream %d on %s>" % (self.handle.value, self.context)
+        default_streams = {
+            0: "<Default CUDA stream on %s>",
+            1: "<Legacy default CUDA stream on %s>",
+            2: "<Per-thread default CUDA stream on %s>",
+        }
```
stuartarchibald, line 1758: Should this use this:  https://github.com/numba/numba/pull/5347/files#diff-511a1676aeb16a8663dbcddcd2d59949R22-R24

numba/cuda/cudadrv/driver.py
gmarkall: By creating a non-default internal stream - e.g.:

```
In [3]: cuda.stream()                                                                                                                                        
Out[3]: <CUDA stream 94332750074080 on <CUDA context c_void_p(94332741571184) of device 0>>
```

numba/cuda/cudadrv/driver.py
gmarkall: It should, and it would also be good sense for me to add `CU_STREAM_LEGACY` for consistency too.

gmarkall: Added now.

numba/cuda/cudadrv/driver.py
stuartarchibald: ah yes, thanks.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_54`.

stuartarchibald:
Buildfarm passed.

Issue #5214, opened 2020-02-07
Collecting things that are being used as "public" API


stuartarchibald:
This issue serves the sole purpose as a place to collect the bits and pieces of Numba that other projects are using as pseudo-public APIs. The reason for doing this is that Numba internals are undergoing a massive refactoring and things are most likely to have been moved.

Project/code owners depending on Numba, this is a helper script for exposing this information (based on [this](https://github.com/scikit-hep/awkward-1.0/blob/2f376e38dbc4599062dc8908b40e5cebb263e9c5/docs/demos/2020-01-22-numba-demo-EVALUATED.ipynb) by @jpivarski (with thanks!)): https://gist.github.com/stuartarchibald/b5d91472b19d7d1f80e55b1dcd2c275f

Dumping the output of that along with a reference to your project would be great, thanks :)

If that doesn't work or your project is private please could you add any use of Numba functions that **are not**:
1. A JIT compilation decorator (e.g. `jit`, `generated_jit`) 
2. `numba.types` or something from `numba.types` 
3. Something from `numba.special`, like `literal_unroll` or `gdb_init` etc.

Will leave this ticket open for a couple of weeks, then make decisions about how to expose commonly used parts of Numba's internals. Thanks all!

stuartarchibald:
CC:
Awkward-array: @jpivarski
clifford: @eric-wieser @hugohadfield
umap: @lmcinnes
pydata/sparse: @hameerabbasi @eric-wieser

stuartarchibald:
Pydata/sparse output:
```
decorators
---------------------------------------------------------
   15 numba.jit
    4 numba.njit

classes
---------------------------------------------------------
    6 numba.typed.List

functions
---------------------------------------------------------
   10 numba.typed.List.empty_list
    5 numba.from_dtype
    2 numba.numpy_support.as_dtype
    1 numba.numpy_support.from_dtype

builder.*
---------------------------------------------------------
   24 builder.store
    5 builder.load
    1 builder.if_else

context.*
---------------------------------------------------------
    5 context.get_constant_generic
    2 context.get_value_type
    2 context.get_constant_null

pyapi.*
---------------------------------------------------------
   18 c.pyapi.decref
    1 c.pyapi.object_getattr_string
    1 c.pyapi.pyobj
    1 c.pyapi.get_null_object
    1 c.pyapi.unserialize
    1 c.pyapi.serialize_object
    1 c.pyapi.tuple_pack
    1 c.pyapi.dict_pack
[...]

PR #5352, opened 2020-03-05
Add shim to accommodate refactoring.

stuartarchibald:
As title.

### DIFF
--- a/numba/analysis.py
+++ b/numba/analysis.py
@@ -1,0 +1,12 @@
+import numba.core.errors as _errors
+from numba.core.utils import PYVERSION as _PYVERSION
+
+_moved_mod = "numba.core.analysis"
+
+if _PYVERSION >= (3, 7):
+    __getattr__ = _errors.deprecate_moved_module_getattr(__name__, _moved_mod)
+else:
+    from numba.core.analysis import *  # noqa: F403, F401
+    from numba.core.analysis import _use_defs_result  # noqa: F401
+
+_errors.deprecate_moved_module(__name__, _moved_mod)
--- a/numba/core/errors.py
+++ b/numba/core/errors.py
@@ -14,6 +14,7 @@ from collections import defaultdict
 from numba.core.utils import add_metaclass, reraise, chain_exception
 from functools import wraps
 from abc import abstractmethod
+from importlib import import_module
 
 # Filled at the end
 __all__ = []
@@ -390,6 +391,48 @@ def deprecated(arg):
         return decorator(arg)
     else:
         return decorator
+
+
+_moved_msg1 = ("An import was requested from a module that has moved location."
+               "\nImport requested from: '{}', please update to use "
+               "'{}' or pin to Numba version 0.48.0. This alias will not be "
+               "present in Numba version 0.50.0.")
+
+
+_moved_msg2 = ("An import was requested from a module that has moved location"
+               ".\nImport of '{}' requested from: '{}', please update to use "
+               "'{}' or pin to Numba version 0.48.0. This alias will not be "
+               "present in Numba version 0.50.0.")
+
+
+_moved_no_replacement = ("No direct replacement available. Visit "
+                         "https://gitter.im/numba/numba-dev to request help. "
+                         "Thanks!")
+
+
+def deprecate_moved_module(old_module, new_module, stacklevel=3):
+    """Warn about a module level location move of some part of Numba's
+    internals. stacklevel is 3 by default as most warning locations are
+    from `numba.XYZ` shims.
+    """
+    if new_module is None:
+        msg = _moved_no_replacement
+    else:
+        msg = _moved_msg1.format(old_module, new_module)
+    warnings.warn(msg, category=NumbaDeprecationWarning, stacklevel=stacklevel)
+
+
+def deprecate_moved_module_getattr(old_module, new_module):
+    """For replacing module level __getattr__ to warn users above modules moving
+    locations
+    """
+    def wrapper(attr):
+        mod = import_module(new_module)
+        ret_attr = getattr(mod, attr)
+        msg = _moved_msg2.format(attr, old_module, new_module)
+        warnings.warn(msg, category=NumbaDeprecationWarning, stacklevel=2)
+        return ret_attr
+    return wrapper
 
 
 class WarningsFixer(object):
--- a/numba/typeconv.py
+++ b/numba/typeconv.py
@@ -1,0 +1,11 @@
+import numba.core.errors as _errors
+from numba.core.utils import PYVERSION as _PYVERSION
+
+_moved_mod = "numba.core.typeconv"
+
+if _PYVERSION >= (3, 7):
+    __getattr__ = _errors.deprecate_moved_module_getattr(__name__, _moved_mod)
+else:
+    from numba.core.typeconv import *  # noqa: F403, F401
+
+_errors.deprecate_moved_module(__name__, _moved_mod)
--- a/buildscripts/incremental/test.sh
+++ b/buildscripts/incremental/test.sh
@@ -9,7 +9,7 @@ set -v -e
 if [ "$BUILD_DOC" == "yes" ]; then rstcheck README.rst; fi
 # Ensure that the documentation builds without warnings
 pushd docs
-if [ "$BUILD_DOC" == "yes" ]; then make SPHINXOPTS=-W clean html; fi
+if [ "$BUILD_DOC" == "yes" ]; then make clean html; fi
 popd
 # Run system info tool
 pushd bin
--- a/numba/array_analysis.py
+++ b/numba/array_analysis.py
@@ -1,0 +1,11 @@
+import numba.core.errors as _errors
+from numba.core.utils import PYVERSION as _PYVERSION
+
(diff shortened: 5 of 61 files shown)


stuartarchibald:
This is to create a bit of a time buffer for users of Numba internals (as discovered here https://github.com/numba/numba/issues/5214) to make the move to the new Numba code base layout. Anything that was on the summary list here https://github.com/numba/numba/issues/5214#issuecomment-593611360 has been re-exported irrespective of whether it's a good idea or something that will be supported in the future.

Discussion items:
1. What sort of warning should be given to users importing from legacy locations? Some deprecation warning with messaging along the lines of "This module (X) has moved to location (Y), please update your code. Numba 0.50 will not support import from the currently attempted location".
2. Should 1. be switch-off-able via `numba.config`/environment variable? Customers/downstream users and projects of libraries consuming Numba internals probably won't want to see this warning/it has no benefit to them.
3. General discussion about what is reasonable to assume as public/stable. Perhaps for another time/PR.
4. How to test that the shim is ok?! Ask users/use numba-integration-testing?

stuartarchibald:
Marking ready for review, discussion items need resolving as part of the process.

eric-wieser:
Is it worth trying something like using module `__getattr__`s to `DeprecationWarning` about accessing through the shim?

esc:
1.  yes
2. probably
3. yeah, that is a whole 'nother discussion altogether.
4. yes, i'll create a test branch to test against this PR now to help with that.

stuartarchibald:
Note to self. Update deprecation docs.

esc:
Preliminary shim testing is up here: https://github.com/numba/numba-integration-testing/pull/32

stuartarchibald:
Thanks @esc.

esc:
Quick report from the first round of testing: `umap` continues to be unaffected and is good and `pydata-sparse` will probably be O.K. with the shim.

esc:
`fastparquet` and `librosa` will probably be fine also and `pygbm` is back to segfaulting (no more import errors).

asmeurer:
Won't `__getattr__` only work if people import the module and then access names as attributes? It would probably be better to just add the deprecation to the top level of the file so it always happens when you import it.

eric-wieser:
@asmeurer: it's also triggered by a direct import of those names. But yes, perhaps in this case a module-level warning is perfectly adequate.

stuartarchibald:
I think that doing both may be necessary, given removal of this patch in 0.50 will break people's non-updated code when removed, I'm inclined for it to be noisy.

esc:
It looks like awkward is still failing over (https://github.com/numba/numba-integration-testing/pull/34):

```
==================================== ERRORS ====================================
_____________ ERROR collecting tests/test_0118-numba-cpointers.py ______________
tests/test_0118-numba-cpointers.py:15: in <module>
    awkward1_numba_arrayview = pytest.importorskip("awkward1._numba.arrayview")
awkward1/_numba/arrayview.py:77: in <module>
    class LookupModel(numba.datamodel.models.StructModel):
E   AttributeError: module 'numba' has no attribute 'datamodel'
```

eric-wieser:
Seems downstream projects sometimes assume these modules are preimported

stuartarchibald:
I propose adding something like:  https://github.com/stuartarchibald/numba/commit/f5744a2c95fe20c0247f1b2e1fb1601574fa7b08

With that enabled, this:
```python
from numba import analysis
from numba.analysis import ir_extension_usedefs
```
produces:
```
$ python test_shim1.py 
test_shim1.py:1: NumbaDeprecationWarning: An import of was requested from a module that has moved location.
Import requested from: 'numba.analysis', please update to use 'numba.core.analysis' or pin to Numba version 0.48.0.
  from numba import analysis
test_shim1.py:2: NumbaDeprecationWarning: An import of was requested from a module that has moved location.
Import requested from: 'numba.analysis', please update to use 'numba.core.analysis' or pin to Numba version 0.48.0.
  from numba.analysis import ir_extension_usedefs
```

eric-wieser:
That change isn't quite correct, but I think the idea is sound. If you ping me when you turn that into a PR I'll review it.

stuartarchibald:
> That change isn't quite correct, but I think the idea is sound. If you ping me when you turn that into a PR I'll review it.

Thanks for looking, I'm going to pull that onto this PR, so please just comment on the patch and I'll make changes. I really want this all to go in as one patch so it's trivial to revert in 0.50. Thanks!

numba/analysis.py
```diff
@@ -1,2 +1,12 @@
-from numba.core.analysis import * # noqa: F403, F401
-from numba.core.analysis import _use_defs_result # noqa: F401
+from numba.core.errors import (deprecate_moved_module,
+                               deprecate_moved_module_getattr)
```
eric-wieser: Perhaps wise not to expose these properties in auto-complete:
```suggestion
import numba.core.errors as _errors
```

numba/analysis.py
```diff
@@ -1,2 +1,12 @@
...
+from numba.core.utils import PYVERSION
+
+if PYVERSION > (3, 6):
+    __getattr__ = deprecate_moved_module_getattr(__name__,
+                                                 'numba.core.analysis')
+else:
+    from numba.core.analysis import * # noqa: F403, F401
+    from numba.core.analysis import _use_defs_result # noqa: F401
+
+deprecate_moved_module(__name__, 'numba.core.analysis')
```
eric-wieser:
```suggestion
_errors.deprecate_moved_module(__name__, 'numba.core.analysis')
```

numba/analysis.py
```diff
@@ -1,2 +1,12 @@
-from numba.core.analysis import * # noqa: F403, F401
-from numba.core.analysis import _use_defs_result # noqa: F401
+from numba.core.errors import (deprecate_moved_module,
+                               deprecate_moved_module_getattr)
+from numba.core.utils import PYVERSION
+
+if PYVERSION > (3, 6):
```
eric-wieser: Or just:
```suggestion
import sys

if sys.version_info >= (3, 7):
```

numba/analysis.py
```diff
@@ -1,2 +1,12 @@
...
+    __getattr__ = deprecate_moved_module_getattr(__name__,
+                                                 'numba.core.analysis')
+else:
+    from numba.core.analysis import * # noqa: F403, F401
+    from numba.core.analysis import _use_defs_result # noqa: F401
```
eric-wieser: PEP8 wants two spaces
```suggestion
    from numba.core.analysis import *  # noqa: F403, F401
    from numba.core.analysis import _use_defs_result  # noqa: F401
```

numba/analysis.py
eric-wieser: `>=` is probably better here even if you keep using `PYVERSION`

numba/typeconv.py
```diff
@@ -0,0 +1 @@
+from numba.core.typeconv import * # noqa: F403, F401
```
eric-wieser: All the files below this line need deprecations too, right?

stuartarchibald: yes, I pushed up this patch just to see what CI thought :) not proposing for one instant that it is actually used as it's way off being finished.

numba/analysis.py
stuartarchibald: thanks, fixed

numba/analysis.py
stuartarchibald: Agree, fixed. Also, for future, Numba uses `PYVERSION`.

eric-wieser: Any reason you have `PYVERSION == sys.version_info[:2]` and not just `PYVERSION = sys.version_info`?

stuartarchibald: Historical, don't think there's any particular need.

numba/core/errors.py
```diff
@@ -392,6 +393,38 @@ def wrapper(*args, **kwargs):
...
+
+def deprecate_moved_module_getattr(old_module, new_module):
+    """For replacing module level __getattr__ to warn users above modules moving
+    locations
+    """
+    def wrapper(attr):
+        mod = import_module(new_module)
+        ret_attr = getattr(mod, attr)
+        warnings.warn(_moved_msg2.format(attr, old_module, new_module),
+                      category=NumbaDeprecationWarning, stacklevel=2)
```
eric-wieser:
```suggestion
                      category=NumbaDeprecationWarning, stacklevel=3)
```

numba/core/errors.py
```diff
@@ -392,6 +393,45 @@ def wrapper(*args, **kwargs):
         return decorator
 
 
+_moved_msg1 = ("An import was requested from a module that has moved location."
+               "\nImport requested from: '{}', please update to use "
+               "'{}' or pin to Numba version 0.48.0.")
```
stuartarchibald: Add `This alias will be removed in Numba 0.50.0.`

stuartarchibald:
https://github.com/stuartarchibald/numba/tree/wip/refactor_shim_1_backup is the same as this branch, am going to flatten all this into one patch now. 9858302 will get force pushed to a single patch.

numba/core/errors.py
stuartarchibald: I went with "This alias will not be present in Numba version 0.50.0.".

PR #5357, opened 2020-03-05
Small fix to have llvm and asm highlighted properly

1844144:
Related to #5333 
@stuartarchibald  mentioned patches https://github.com/numba/numba/pull/5333#issuecomment-594825303 for happiness of dump readers

### DIFF
--- a/numba/core/codegen.py
+++ b/numba/core/codegen.py
@@ -23,7 +23,7 @@ def _is_x86(triple):
     return arch in _x86arch
 
 
-def dump(header, body, lang=None):
+def dump(header, body, lang):
     if config.HIGHLIGHT_DUMPS:
         try:
             import pygments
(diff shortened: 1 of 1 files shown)


numba/core/codegen.py
```diff
@@ -32,9 +32,11 @@ def dump(header, body):
             raise ValueError(msg)
         else:
             from pygments import highlight
-            from pygments.lexers import GasLexer as lexer
+            from pygments.lexers import GasLexer as gas_lexer
+            from pygments.lexers import LlvmLexer as llvm_lexer
             from pygments.formatters import Terminal256Formatter
             from numba.misc.dump_style import by_colorscheme
+            lexer = llvm_lexer if lang is None else gas_lexer
```
stuartarchibald: Given the mistake that's motivating this change, I think that it would be a good idea to require a string to specify the lexer and then select based on that, this to force thinking about the target language.

stuartarchibald:
Thanks for the PR. One minor thing to resolve, then I'll give it a manual testing.

numba/core/codegen.py
1844144: My goal was to provide minimum changes to current implementation. ok, we can make 3 arg required

1844144:
hmmm
image

stuartarchibald:
@1844144 thanks for the update. RE: https://github.com/numba/numba/pull/5357#issuecomment-595793039 take a look in `numba.core.rewrites.registry` around line 70, the output that's not highlighted is coming from `block.dump()`. This makes me wonder if it'd be a good idea to just make `numba.core.ir.Block.dump()` and `numba.core.ir.FunctionIR.dump()` invoke the highlighting as determined by the config?

1844144:
Thank you. 

As for me more generic approach is to make standalone `dump(Header, Body, Lang)` function that 
 - will take into consideration HIGHLIGHT_DUMPS, and COLOR_SCHEME options 
 - will be used in many places where in fact there are duplicate of the same dump code

From another point there are more interesting/important issues.

PR #5366, opened 2020-03-09
Remove reference to Python 2.7 in install check output

stuartarchibald:
As title.


esc:
Looks good, thanks for the patch!

PR #5369, opened 2020-03-09
 Make test ordering more consistent across feature availability

### LINKED ISSUE #5368: sort test order so that generated tests go at the end
Generated tests relying on feature/package/library presence at generation time are making the test ordering inconsistent across the many platforms/combinations in CI. This has a knock-on effect in test slicing in that the slices are often inconsistent and this sometimes results in less coverage that would be good. Whilst the current slicing approach is known to be imperfect, fixing this issue would make it slightly less imperfect. The way to fix is to supply a key to the sorter for the tests that are sliced, this key pushes the generated tests to the back of the list, hence ordering is more consistent and slices provide better coverage.

stuartarchibald:
As title.

Closes #5368

### DIFF
--- a/buildscripts/azure/azure-windows.yml
+++ b/buildscripts/azure/azure-windows.yml
@@ -51,6 +51,12 @@ jobs:
 
     - script: |
         call activate %CONDA_ENV%
+        python -m numba.runtests -l
+      displayName: 'List discovered tests'
+
+    - script: |
+        call activate %CONDA_ENV%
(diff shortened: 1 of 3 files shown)


stuartarchibald:
Note to reviewers:
12664c0 is the patch that does the required.
8c833da makes the CI system print out the test order it "sees" so manual confirmation that the ordering is "right" can be made.

Once this is reviewed, I'll issue a revert on 8c833da as its purpose will have been served.

PR #5376, opened 2020-03-10
Improve the error message when _StructProxy.__setattr__ fails

eric-wieser:
Previously only the field index was emitted, now the name is too.


eric-wieser:
I think this is ready for review, unless you want a test of the error message

PR #5378, opened 2020-03-11
Improvements to tuple_setitem

ARF1:
Improvements to `tuple_setitem`:
- Permit setting items of heterogeneous tuples
  (llvm requires constant index)
- sanity-checking inputs (boundscheck, type checks)

### DIFF
--- a/numba/cpython/unsafe/tuple.py
+++ b/numba/cpython/unsafe/tuple.py
@@ -2,9 +2,82 @@
 This file provides internal compiler utilities that support certain special
 operations with tuple and workarounds for limitations enforced in userland.
 """
-
-from numba.core.cgutils import alloca_once
+from numba.core import types, typing
+from numba.core.errors import RequireLiteralValue
 from numba.core.extending import intrinsic
+
+
+# QUESTION: Is there a good place for this function somewhere else?
+#           I keep writing boiler-plate type checks and having a function that
+#           does this for most applications would get rid of much of the 
+#           boiler-plate.
+def validate_arg_type(arg, arg_name, valid_types, none_allowed=False,
+                      msg_prefix=None):
+    """Checks that a parameter of an overload has the correct type and prints
+    appropriate error message if not.
+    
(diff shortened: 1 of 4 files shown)


esc:
@ARF1 thanks for submitting this PR to the Numba issue tracker. I have added it to the queue for review. Be advised however, that we are currently in a PR freeze as we prepare the release of 0.49.0 next week. So, we will begin reviewing PRs again the week starting 23 March and we thank you in advance for your patience!

ARF1:
@esc Thanks for the heads up.

numba/cpython/unsafe/tuple.py
```diff
@@ -2,11 +2,88 @@
...
+    # if a type instance is defined, it's name takes precendence
+    valid_names = [getattr(type_, 'name', None) if name is None
+                   else name
+                   for type_, name in zip(valid_types, valid_names)]
+    # fallback to type name
+    valid_names = [getattr(type_, '__name__', None) if name is None
+                   else name
+                   for type_, name in zip(valid_types, valid_names)]
+    for x in valid_names:
+        assert x is not None, f"Cannot find name for type {x}"
```
eric-wieser: Possibly clearer as:
```suggestion
    def name_for_type(type_):
        # any defined name mapping takes precedence
        name = name_mapping.get(type_)
        if name is not None:
            return name
        # if a type instance is defined, it's name takes precendence
        name = getattr(type_, 'name', None)
        if name is not None:
            return name
        # fallback to type name
        name = getattr(type_, '__name__', None)
        if name is not None:
            return name
        raise ValueError(f"Cannot find name for type {x}")

    valid_names = [name_for_type(type_)  for type_ in valid_types]
```

ARF1: @eric-wieser Thanks for the suggestion.

esc:
We discussed this during a backlog refinement today and decided that setting an item on a heterogenous tuple isn't something that is needed in practice, hence closing this issue now.

PR #5382, opened 2020-03-12
Add support for the `order` argument to np.ones, np.zeros, and np.empty

eric-wieser:
Needs tests - can a reviewer tell me where the best place to put them would be?

### DIFF
--- a/numba/core/typing/npydecl.py
+++ b/numba/core/typing/npydecl.py
@@ -450,6 +450,10 @@ def _parse_dtype(dtype):
     if isinstance(dtype, types.DTypeSpec):
         return dtype.dtype
 
+def _parse_order(order):
+    if isinstance(order, types.StringLiteral):
+        return order.literal_value
+
 def _parse_nested_sequence(context, typ):
     """
     Parse a (possibly 0d) nested sequence type.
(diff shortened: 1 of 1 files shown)


ARF1:
@eric-wieser Thanks for this. I have been missing the order argument. Any chance you could also do `np.full`?

stuartarchibald:
Thanks for the PR. Would suggest adding tests to https://github.com/numba/numba/blob/51c098dce8409d2248aade80a17fc4c1ee58589b/numba/tests/test_dyn_array.py#L581

suspect that for this patch to work, lowering also needs wiring up, https://github.com/numba/numba/blob/51c098dce8409d2248aade80a17fc4c1ee58589b/numba/np/arrayobj.py#L3516-L3522 so that a 3 arg signature, last arg being a string literal is accepted.

stuartarchibald:

eric-wieser:
I'm a little confused by the type specs in that file. Is there a reason for the inconsistency between
```python
@lower_builtin(np.zeros, types.Any, types.Any)
@lower_builtin(np.ones, types.Any, types.DTypeSpec)
```
Should I be precise about the types, or does it not matter? That is, would you prefer me to add
```
@lower_builtin(np.zeros, types.Any, types.Any, types.Any)
```python
or
```python
@lower_builtin(np.zeros, types.Any, types.DTypeSpec, types.StringLiteral)
```

stuartarchibald:
RE: https://github.com/numba/numba/pull/5382#issuecomment-598132744

I would hazard a guess at this being a historical artifact and nothing more. Ideally, precision is preferred, the typers should guard against anything illegal being passed to lowerers and the lowerers should be specific and match what the typers permit. The challenge in matching these up along with keeping them in sync is why `@overload` was developed, however, that also has shortcomings around forward declaring type specifications and so is also due an update!

PR #5388, opened 2020-03-13
CUDA: Fix OOB write in test_round{f4,f8}

gmarkall:
In these tests, the typing of the first parameter passed to the kernel (`int64[:]`)  vs. the type of the array that was passed in (`np.int32`) resulted in 8 bytes being written to a 4-byte allocation.

This was picked up with cuda-memcheck whilst looking into #4954 (but does not fix it - it is unrelated).


eric-wieser:
Shouldn't this be a type error?

gmarkall:
Good point - it probably should have! I've made a note on the issue regarding multiple signatures in CUDA (because that's an area that's due an overhaul anyway).

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_46`.

PR #5399, opened 2020-03-17
[WIP] Add: iterate empty tuple

### LINKED ISSUE #5372: literal_unroll crashes with empty tuple
On the current master, `literal_unroll` crashes if the tuple to be unrolled is empty:

```python
from numba import njit, literal_unroll

indices = tuple()

@njit
def testnumba():
    for i in literal_unroll(indices):
        print(i)


testnumba()
```

Output:
```
Invalid use of getiter with parameters (Tuple())
 * parameterized
[1] During: typing of intrinsic-call at c:\Users\My User\Documents\numba\unroll.py (8)

File "unroll.py", line 8:
def testnumba():
    for i in literal_unroll(indices):
    ^
```

1844144:
Closes #5372. Solves reproducers.  Perhaps it will have sense in some cases.

### DIFF
--- a/numba/core/types/containers.py
+++ b/numba/core/types/containers.py
@@ -136,11 +136,15 @@ class BaseTuple(ConstSized, Hashable):
                     return NamedTuple(tys, pyclass)
         else:
             # non-named tuple
-            homogeneous = is_homogeneous(*tys)
-            if homogeneous:
-                return UniTuple(tys[0], len(tys))
+            len_tys = len(tys)
+            if len_tys:
+                homogeneous = is_homogeneous(*tys)
+                if homogeneous:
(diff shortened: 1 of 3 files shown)


eric-wieser:
I thought I saw an issue or PR somewhere to support iterating over heterogenous tuples.

Perhaps it's better to wait for that change rather than to build this hack,

1844144:
Thank you.

Please do not consider above commits as any kind of "solution". It is more "playing with test system/numba internals".

stuartarchibald:
@1844144 thanks for opening this patch and taking a look at this issue. PR #7750 is attempting to provide a more general solution for iterating over empty containers and is under review, so am going to close this PR. Thanks for your understanding.

PR #5400, opened 2020-03-18
Fix #4954, and some other small CUDA testsuite fixes

### LINKED ISSUE #4954: Running only CUDA tests results in strange crashes / failures
## Reporting a bug

Running the CUDA testsuite only seems to result in some surprising results. For example, on my system with a V100, running:

```
python -m numba.runtests numba.cuda.tests -v
```

eventually dies with:

```
test_broadcast (numba.cuda.tests.cudadrv.test_cuda_array_slicing.CudaArraySetting) ... 
Fatal Python error: Segmentation fault

Current thread 0x00007fb059d49700 (most recent call first):
  File "/home/nfs/gmarkall/numbadev/numba/numba/cuda/cudadrv/driver.py", line 293 in safe_cuda_api_call
  File "/home/nfs/gmarkall/numbadev/numba/numba/cuda/cudadrv/driver.py", line 1622 in launch_kernel
  File "/home/nfs/gmarkall/numbadev/numba/numba/cuda/cudadrv/driver.py", line 1578 in __call__
  File "/home/nfs/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 614 in _kernel_call
  File "/home/nfs/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 540 in __call__
  File "/home/nfs/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 278 in __call__
  File "/home/nfs/gmarkall/numbadev/numba/numba/cuda/cudadrv/devicearray.py", line 581 in _do_setitem
  File "/home/nfs/gmarkall/numbadev/numba/numba/cuda/cudadrv/devicearray.py", line 530 in __setitem__
  File "/home/nfs/gmarkall/numbadev/numba/numba/cuda/cudadrv/devices.py", line 225 in _require_cuda_context
  File "/home/nfs/gmarkall/numbadev/numba/numba/cuda/tests/cudadrv/test_cuda_array_slicing.py", line 209 in test_broadcast
  File "/home/nfs/gmarkall/miniconda3/envs/numbaenv/lib/python3.7/unittest/case.py", line 628 in run
  File "/home/nfs/gmarkall/miniconda3/envs/numbaenv/lib/python3.7/unittest/case.py", line 676 in __call__
  File "/home/nfs/gmarkall/miniconda3/envs/numbaenv/lib/python3.7/unittest/suite.py", line 122 in run
  File "/home/nfs/gmarkall/miniconda3/envs/numbaenv/lib/python3.7/unittest/suite.py", line 84 in __call__
  File "/home/nfs/gmarkall/miniconda3/envs/numbaenv/lib/python3.7/unittest/runner.py", line 176 in run
  File "/home/nfs/gmarkall/numbadev/numba/numba/testing/main.py", line 123 in run
  File "/home/nfs/gmarkall/miniconda3/envs/numbaenv/lib/python3.7/unittest/main.py", line 271 in runTests
  File "/home/nfs/gmarkall/numbadev/numba/numba/testing/main.py", line 354 in run_tests_real
  File "/home/nfs/gmarkall/numbadev/numba/numba/testing/main.py", line 369 in runTests
  File "/home/nfs/gmarkall/miniconda3/envs/numbaenv/lib/python3.7/unittest/main.py", line 101 in __init__
  File "/home/nfs/gmarkall/numbadev/numba/numba/testing/main.py", line 163 in __init__
  File "/home/nfs/gmarkall/numbadev/numba/numba/testing/__init__.py", line 75 in run_tests
  File "/home/nfs/gmarkall/numbadev/numba/numba/_runtests.py", line 28 in _main
  File "/home/nfs/gmarkall/numbadev/numba/numba/runtests.py", line 9 in <module>
  File "/home/nfs/gmarkall/miniconda3/envs/numbaenv/lib/python3.7/runpy.py", line 85 in _run_code
  File "/home/nfs/gmarkall/miniconda3/envs/numbaenv/lib/python3.7/runpy.py", line 193 in _run_module_as_main
Segmentation fault
```

The same on @stuartarchibald's machine with a GTX 750 Ti doesn't segfault, but has the following failure:

```
numba/cuda/tests/cudadrv/test_linker.py", line 93, in test_set_registers_57 self.assertEquals(57, compiled._func.get().attrs.regs) AssertionError: 57 != 56
```

However, the testsuite as a whole has no failures when I run it like:

```
python -m numba.runtests -m
```

The test that segfaults for me is getting discovered in this case - it shows up when listing tests with `-l` - so I don't think it's being omitted when running the whole testsuite.

Some initial investigation suggests that some CUDA API calls fail undetected, leading to launch failures later - in one case, computing the threads per block somehow results in a thread count of 0, which a kernel is then configured with and subsequently fails to launch. However, I cannot reliably reproduce this - I'm now getting failures with nonzero threads per block.

I plan to continue investigating this some more. Some other thoughts / hypotheses:

* Perhaps something related to the `CUDATestCase` class, which is used for a few of the tests, and resets the CUDA context on teardown may be involved - perhaps either resetting the context at an inappropriate time, or perhaps there's a test class that needs to be a `CUDATestCase` and is presently only a `unittest.TestCase` + `SerialMixin`. I'm not sure of the exact rationale behind / rules for using `CUDATestCase` as opposed to just the `SerialMixin` for CUDA tests - can anyone shed any light on this?
* Some memory corruption may be involved - I'm presently looking for a way to run valgrind and also hit the error in a way that isn't too painful to do.

gmarkall:
This PR aims to fix #4954. Factors that come together to cause the issue are:

- The interleaving of CUDA test methods from different classes during execution.
- Some of the CUDA test classes having teardowns that reset the context after every method.
- Other CUDA test classes expect various CUDA objects (allocations, modules, streams, etc.) to persist between methods

So, some methods for class A would run, class B would then have a few methods run and reset the context, then another method from class A runs and causes unexpected behaviour (various CUDA errors, a segfault, etc.) because objects it expected to persist no longer exist in the context.

The main fix of this PR is to ensure that CUDA test classes have all their test methods executed concurrently, and not interleaved with other classes. However, there are also some other fixes / tidy-ups to the CUDA tests:

- Rename `CUDATestCase` to `ContextResettingTestCase`, so that it's clear what the class is for and what the subclasses will do.
- Put all tests that use a CUDA device under a common superclass, `CUDATestCase`, so that they can be identified easily (e.g. for the purpose of sorting correctly, as in this PR) or their behaviour modified as a set without editing lots of individual test classes (for possible future changes, e.g. parallelising the CUDA test suite). This is a subclass of `numba.testing.support.TestCase` because some of the CUDA test cases need `self.assertPreciseEqual`, etc.
- Fix the `TestVectorizeDecor` CUDA test - there's no way this could have worked before, and in previous test runs it doesn't appear in the results. I think some set of circumstances conspired to hide the errors from the results completely - now that the class is fixed and seems to be getting picked up, I get 588 CUDA tests run instead of 584 prior to this PR.

### DIFF
--- a/numba/cuda/testing.py
+++ b/numba/cuda/testing.py
@@ -8,10 +8,20 @@ from numba.tests.support import (
 )
 from numba.cuda.cuda_paths import get_conda_ctk
 from numba.core import config
+from numba.tests.support import TestCase
 import unittest
 
 
-class ContextResettingTestCase(SerialMixin, unittest.TestCase):
+class CUDATestCase(SerialMixin, TestCase):
+    """
+    For tests that use a CUDA device. Test methods in a CUDATest case must not
+    be run out of module order, because the ContextResettingTestCase may reset
+    the context and destroy resources used by a normal CUDATestCase if any of
+    its tests are run between tests from a CUDATestCase.
+    """
+
+
+class ContextResettingTestCase(CUDATestCase):
     """
     For tests where the context needs to be reset after each test. Typically
     these inspect or modify parts of the context that would usually be expected
--- a/numba/cuda/testing.py
+++ b/numba/cuda/testing.py
@@ -14,7 +14,7 @@ import unittest
 
 class CUDATestCase(SerialMixin, TestCase):
     """
-    For tests that use a CUDA device. Test methods in a CUDATest case must not
+    For tests that use a CUDA device. Test methods in a CUDATestCase must not
     be run out of module order, because the ContextResettingTestCase may reset
     the context and destroy resources used by a normal CUDATestCase if any of
     its tests are run between tests from a CUDATestCase.
--- a/numba/cuda/testing.py
+++ b/numba/cuda/testing.py
@@ -11,10 +11,16 @@ from numba.core import config
 import unittest
 
 
-class CUDATestCase(SerialMixin, unittest.TestCase):
+class ContextResettingTestCase(SerialMixin, unittest.TestCase):
+    """
+    For tests where the context needs to be reset after each test. Typically
+    these inspect or modify parts of the context that would usually be expected
+    to be internal implementation details (such as the state of allocations and
+    deallocations, etc.).
+    """
+
     def tearDown(self):
         from numba.cuda.cudadrv.devices import reset
-
         reset()
 
 
--- a/numba/cuda/tests/cudadrv/test_array_attr.py
+++ b/numba/cuda/tests/cudadrv/test_array_attr.py
@@ -1,10 +1,9 @@
 import numpy as np
 from numba import cuda
-from numba.cuda.testing import unittest, SerialMixin
-from numba.tests.support import TestCase
+from numba.cuda.testing import unittest, CUDATestCase
 
 
-class TestArrayAttr(SerialMixin, TestCase):
+class TestArrayAttr(CUDATestCase):
     def test_contigous_2d(self):
         ary = np.arange(10)
         cary = ary.reshape(2, 5)
--- a/numba/cuda/tests/cudadrv/test_context_stack.py
+++ b/numba/cuda/tests/cudadrv/test_context_stack.py
@@ -3,11 +3,11 @@ from ctypes import byref
 import weakref
 
 from numba import cuda
-from numba.cuda.testing import unittest, SerialMixin, skip_on_cudasim
+from numba.cuda.testing import unittest, CUDATestCase, skip_on_cudasim
 from numba.cuda.cudadrv import driver
 
 
-class TestContextStack(SerialMixin, unittest.TestCase):
+class TestContextStack(CUDATestCase):
     def setUp(self):
         # Reset before testing
         cuda.close()
@@ -25,7 +25,7 @@ class TestContextStack(SerialMixin, unittest.TestCase):
         self.assertGreater(len(gpulist), 0)
 
 
-class TestContextAPI(SerialMixin, unittest.TestCase):
+class TestContextAPI(CUDATestCase):
 
     def tearDown(self):
         cuda.close()
@@ -68,7 +68,7 @@ class TestContextAPI(SerialMixin, unittest.TestCase):
 
 
 @skip_on_cudasim('CUDA HW required')
-class Test3rdPartyContext(SerialMixin, unittest.TestCase):
+class Test3rdPartyContext(CUDATestCase):
     def tearDown(self):
         cuda.close()
 
--- a/numba/cuda/tests/cudadrv/test_cuda_array_slicing.py
+++ b/numba/cuda/tests/cudadrv/test_cuda_array_slicing.py
@@ -3,10 +3,10 @@ from itertools import product
 import numpy as np
 
 from numba import cuda
-from numba.cuda.testing import unittest, SerialMixin
+from numba.cuda.testing import unittest, CUDATestCase
 
 
-class CudaArrayIndexing(SerialMixin, unittest.TestCase):
+class CudaArrayIndexing(CUDATestCase):
     def test_index_1d(self):
         arr = np.arange(10)
         darr = cuda.to_device(arr)
@@ -56,7 +56,7 @@ class CudaArrayIndexing(SerialMixin, unittest.TestCase):
             darr[0, 0, z]
 
 
-class CudaArrayStridedSlice(SerialMixin, unittest.TestCase):
+class CudaArrayStridedSlice(CUDATestCase):
 
     def test_strided_index_1d(self):
         arr = np.arange(10)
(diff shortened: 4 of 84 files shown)


stuartarchibald:
@gmarkall please ping when you want test running on the farm! Thanks.

stuartarchibald:
If this is stable and a confirmed fix ahead of 0.49RC, I'm inclined to include it. Without such a fix the CUDA test suite is of no use under certain conditions as it segfaults!

gmarkall:
@stuartarchibald If it's not too much trouble, could you do a run on the build farm of this in its present form please?

And also, does it stop the test suite segfaulting for you on you machine with a GTX 750 Ti?

I've found that even after the merge of #5369 the tests can still get put out-of-order, and therefore segfault - so far with removing the sort on mtime, I haven't got a segfault yet...

stuartarchibald:
> @stuartarchibald If it's not too much trouble, could you do a run on the build farm of this in its present form please?

Done. Buildfarm ID: `numba_smoketest_cuda_40`.

> And also, does it stop the test suite segfaulting for you on you machine with a GTX 750 Ti?

Initial testing suggests so, I think the card lacks the capacity to make it all the way through the suite, but without this patch, only a handful of tests would ever run before there was a segfault.

> I've found that even after the merge of #5369 the tests can still get put out-of-order, and therefore segfault - so far with removing the sort on mtime, I haven't got a segfault yet...

I see, is this perhaps because within the same module there's some shuffling still? #5369 was to ensure the listed test order was stable for slicing across many machines, once sliced (or if no slicing was performed) the execution order would be `mtime` based I think?

stuartarchibald:
Buildfarm testing passed for 6279ac6.

gmarkall:
Many thanks! Also with #5388 applied to this 6279ac6 and a couple of tests disabled because they lock up cuda-memcheck (IPC tests, multithreaded compiling tests and those that find libraries), the CUDA test run is clean: 

```
$ cuda-memcheck python -m numba.runtests numba.cuda.tests -v
========= CUDA-MEMCHECK
...

Ran 577 tests in 79.114s

OK (skipped=4)
========= ERROR SUMMARY: 0 errors
```

gmarkall:
> > And also, does it stop the test suite segfaulting for you on you machine with a GTX 750 Ti?
> 
> Initial testing suggests so, I think the card lacks the capacity to make it all the way through the suite, but without this patch, only a handful of tests would ever run before there was a segfault.

Thanks for confirming!

> > I've found that even after the merge of #5369 the tests can still get put out-of-order, and therefore segfault - so far with removing the sort on mtime, I haven't got a segfault yet...
> 
> I see, is this perhaps because within the same module there's some shuffling still? #5369 was to ensure the listed test order was stable for slicing across many machines, once sliced (or if no slicing was performed) the execution order would be `mtime` based I think?

I agree with your analysis - I've just restored the sorting by mtime, with the exception of CUDA tests, which sorts by module and class name first before mtime, to keep the ordering requirement for CUDA.

My experimenting with this so far indicates that:

- Touching non-CUDA test files results in those tests getting run in different orders from run to run, and even tests from different classes get interleaved.
- Touching CUDA tests does not affect the order, which is always consistent for the CUDA tests now.
- Running CUDA tests is much faster now (it used to take ~180 seconds for me on a DGX-1, now it takes about ~100 seconds - I guess because it's not struggling with all sorts of things resulting from freed memory/objects being used).

I think as soon as the CI passes (fingers crossed!) this will be ready for review.

gmarkall:
@stuartarchibald I think this is ready for review now.

> If this is stable and a confirmed fix ahead of 0.49RC, I'm inclined to include it. Without such a fix the CUDA test suite is of no use under certain conditions as it segfaults!

Thanks for this - not having it for 0.49 would mean that a released version of Numba wouldn't be able to get through its own testsuite without segfaulting, which sounds a little less than ideal.

stuartarchibald:
> @stuartarchibald I think this is ready for review now.

Great, thanks.

> > If this is stable and a confirmed fix ahead of 0.49RC, I'm inclined to include it. Without such a fix the CUDA test suite is of no use under certain conditions as it segfaults!
> 
> Thanks for this - not having it for 0.49 would mean that a released version of Numba wouldn't be able to get through its own testsuite without segfaulting, which sounds a little less than ideal.

Quite! I took a quick skim over the changeset and didn't see anything of enough concern to prevent it going into 0.49 so will schedule it for that.

Thanks for fixing it!

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_41` is testing 2b1e6c5

stuartarchibald:
`numba_smoketest_cuda_41` passed, 2b1e6c5 is good.

stuartarchibald:
Manually testing on the GTX 750 Ti box, this doesn't segfault after most of the tests whereas it used to segfault really quickly, after about 10. The entire CUDA test suite still doesn't run, but I imagine that's because the card runs out of memory or similar. @gmarkall if you cannot reproduce anywhere either then I think we'll try this patch as an assumed fix. Thanks.

gmarkall:
@stuartarchibald Many thanks for confirming - I can't reproduce #4954 at all with this PR.

stuartarchibald:
Excellent, thanks. Marked for merge.

Issue #5336, opened 2020-03-02
NativeValue.cleanup doesn't seem to be used consistently


eric-wieser:
This is a speculative report, but while trying to implement boxing and unboxing of my own compound type, I ran into some confusion about the nature of `NativeValue.cleanup`.

---

## Case A

Here's what seems to be the motivating case for `cleanup`:

https://github.com/numba/numba/blob/a4c4a8dceb0801672d6fda9199c36047dba16f0b/numba/core/boxing.py#L192-L203

As I read it, this returns either:

* A pointer to the buffer, no error set, and a cleanup function which emits code to release the buffer
* A null pointer, an error set, and a cleanup function which emits code that would segfault if called.

From this, the semantics seem to be:

* If you succeed, set `is_error` to `False`, and set `cleanup` to something that emits cleanup code
* If you fail, set `is_error` to `True`. The caller promises never to call `cleanup`.

This is sort of consistent with the python model of either returning an object implementing `tp_del`, or returning `NULL` and performing all cleanup as the callee.

## Case B

Now let's look at tuple unboxing:

https://github.com/numba/numba/blob/a4c4a8dceb0801672d6fda9199c36047dba16f0b/numba/core/boxing.py#L528-L553

Here, we set `is_error` if the size does not match, yet continue unboxing anyway and generating cleanup tasks. If one of those unboxings fail, then we keep going anyway. There are at least two problems here:
* The attempt at providing a conditional clean up based on `size_matches` fails to achieve its intent (gh-5334), as wrapping a function definition in `c.builder.if_then` has no effect
* If is_error is set, then our cleanup function seemingly should not be invoked by the caller anyway (as we deduced in case A). If this happens, then none of the cleanup actually gets done, and we leak objects.

---

Now let's look at `_lower_call_ObjModeDispatcher`:

https://github.com/numba/numba/blob/a4c4a8dceb0801672d6fda9199c36047dba16f0b/numba/core/lowering.py#L831-L854

Here, we:

* Call `cleanup` whether the error occured or not. This contradicts Case A, and as a result can cause a segfault.
* Call `cleanup` before we're done with the result, meaning that `output` is left dangling here.

---

Are the intended semantics of `cleanup` clearly defined anywhere? In the absence of a clear intent, I'd argue that everything should assume the model used by case A.

cc @stuartarchibald, whom via gitter I subjected to a barrage of questions about this.

sklam:
[...]

PR #5417, opened 2020-03-24
Ensure that tuple unboxing cleans up early items if later ones fail

eric-wieser:
Implemented based on discussion in gh-5336, and on the approach used in `callwrapper._ArgManager`.

Replaces gh-5334.

### DIFF
--- a/numba/core/boxing.py
+++ b/numba/core/boxing.py
@@ -521,6 +521,9 @@ def unbox_tuple(typ, obj, c):
     is_error_ptr = cgutils.alloca_once_value(c.builder, cgutils.false_bit)
     value_ptr = cgutils.alloca_once(c.builder, lty)
 
+    end_blk = c.builder.append_basic_block("end")
+    fail_blk = end_blk
+
     # Issue #1638: need to check the tuple size
     actual_size = c.pyapi.tuple_size(obj)
     size_matches = c.builder.icmp_unsigned('==', actual_size,
@@ -531,6 +534,7 @@ def unbox_tuple(typ, obj, c):
             "size mismatch for tuple, expected %d element(s) but got %%zd" % (n,),
             actual_size)
         c.builder.store(cgutils.true_bit, is_error_ptr)
+        c.builder.branch(fail_blk)
 
     # We unbox the items even if not `size_matches`, to avoid issues with
     # the generated IR (instruction doesn't dominate all uses)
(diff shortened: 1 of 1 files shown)


numba/core/boxing.py
```diff
@@ -531,29 +534,43 @@ def unbox_tuple(typ, obj, c):
...
+
         with c.builder.if_then(native.is_error, likely=False):
             c.builder.store(cgutils.true_bit, is_error_ptr)
+            c.builder.branch(fail_blk)
         if native.cleanup is not None:
             cleanups.append(native.cleanup)
 
+            # replace the failure block with one that cleans up this item, and
+            # then jumps to the previous failure block
+            cleanup_blk = c.builder.append_basic_block('cleanup')
```
eric-wieser: Could generate a unique name for each of these, but it seems llvm doesn't care

eric-wieser:
Fails [here](https://dev.azure.com/numba/numba/_build/results?buildId=4831&view=logs&j=a506c6d0-8428-5bb5-7d36-c242a5467a36&t=fbf6f884-d589-58cc-173a-ae041ad78914&l=9007) with
```
RuntimeError: Failed in nopython mode pipeline (step: nopython mode backend)
Instruction does not dominate all uses!
  %.241 = call i8* @PyTuple_GetItem(i8* %.20, i64 1)
  call void @numba_reset_pyobject_private_data(i8* %.241)
Instruction does not dominate all uses!
  %.33 = call i8* @PyTuple_GetItem(i8* %.20, i64 0)
  call void @numba_reset_pyobject_private_data(i8* %.33)
```
@sklam, is there an easy way to get hold of more complete IR for the failure?

stuartarchibald:
@eric-wieser http://numba.pydata.org/numba-doc/latest/reference/envvars.html#envvar-NUMBA_DUMP_LLVM and similar may be of use when reproducing this locally.

eric-wieser:
I think I concluded that the problem is that LLVM can't handle:
```
ok = ...

if ok:
    x = Foo
...
if ok:
    y = x  # error: the assembler does not know if x has been assigned without performing value-based analysis
```
I could work around it by initializing `x` to garbage, but I prefer the approach in #5422

sklam:
Regarding https://github.com/numba/numba/pull/5417#issuecomment-604423109, the second `if` block would need a phi node for `x`. The optimizer would simplify the CFG and figure out that `x` is never uninitialized.

eric-wieser:
@sklam, do I need the phi node to have a dummy value for all possible entry points, or not?

PR #5418, opened 2020-03-25
Add np.asfarray impl

guilhermeleobas:
As title. I've updated `types.float_` to be `types.float64` instead of `types.float32`. On NumPy, at least on my machine, `np.float_` is `np.float64`.


guilhermeleobas:
Ready for a round of reviews

stuartarchibald:
Thanks for the patch.

> As title. I've updated `types.float_` to be `types.float64` instead of `types.float32`. On NumPy, at least on my machine, `np.float_` is `np.float64`.

What does this mean?! I can't see related changes to the fundamental type in the patchset.

guilhermeleobas:
> Thanks for the patch.
> 
> > As title. I've updated `types.float_` to be `types.float64` instead of `types.float32`. On NumPy, at least on my machine, `np.float_` is `np.float64`.
> 
> What does this mean?! I can't see related changes to the fundamental type in the patchset.

I've undone those changes because it was breaking some tests.

guilhermeleobas:
PR is ready for a review

numba/np/arraymath.py
```diff
@@ -3899,6 +3899,20 @@ def impl(a, dtype=None):
     return impl
 
 
+@overload(np.asfarray)
+def np_asfarray(a, dtype=types.float64):
+    if is_nonelike(dtype):
+        dt = types.float64
+    else:
+        dt = dtype.dtype
+        if not isinstance(dt, types.Float):
```
luk-f-a, line 3908: numpy allows complex to be the output dtype of `asfarray` but it seems that this line would produce a float output, even when the user passes a complex dtype.

numba/tests/test_np_functions.py
```diff
@@ -3148,6 +3152,25 @@ def make_unicode_list():
         test_reject(make_nested_list_with_dict())
         test_reject(make_unicode_list())
 
+    def test_asfarray(self):
+        def inputs():
+            yield np.array([1, 2, 3]), None
+            yield np.array([2, 3], dtype=np.float32), np.float32
+            yield np.array([2, 3], dtype=np.int8), np.int8
```
luk-f-a, line 3159: I would add a test for complex64 or complex128

guilhermeleobas:
Thanks for the review @luk-f-a. I've fixed the code to support complex types.

luk-f-a:
@guilhermeleobas thanks for the changes, it looks good to me.

guilhermeleobas:
cc @hameerabbasi

guilhermeleobas:
Hi, @stuartarchibald 

When you have some cycles to spare, can you take a look in this PR?

numba/np/arraymath.py
```diff
@@ -3899,6 +3899,19 @@ def impl(a, dtype=None):
     return impl
 
 
+@overload(np.asfarray)
+def np_asfarray(a, dtype=types.float64):
```
stuartarchibald, line 3903: think this ought to be `dtype=np.float64`?

stuartarchibald: and then something check that the thing passed in is actually a dtype or can be accepted and converted to one?

guilhermeleobas: Is the check necessary? if `dtype` is not a `subdtype` of `np.inexact`, then, `dtype` will be assigned to `types.float64`.

numba/tests/test_np_functions.py
```diff
@@ -245,6 +245,10 @@ def asarray_kws(a, dtype):
     return np.asarray(a, dtype=dtype)
 
 
+def asfarray(a, dtype=np.float_):
```
stuartarchibald, line 248:
```suggestion
def asfarray(a, dtype=np.float64):
```

numba/np/arraymath.py
stuartarchibald: ok, I think the check can be skipped as the `as_dtype` will pick up nonsense types.

numba/tests/test_np_functions.py
guilhermeleobas: Done!

stuartarchibald (approved):
Thanks for the fixes, looks good!

stuartarchibald:
Note: CI fail is #5973

guilhermeleobas:
Thanks for the review @stuartarchibald

Issue #5236, opened 2020-02-11
Unboxer/boxer caching for speed


stuartarchibald:
## Reporting a bug

- [x] I am using the latest released version of Numba (most recent is visible in
 the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [x I have included below a minimal working reproducer (if you are unsure how
 to write one see http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports).

As noted here https://github.com/pygae/clifford/pull/229#issuecomment-584630861 unboxing (and boxing) impacts compile time performance. It is exacerbated by the {un,}boxers generating the same LLVM IR each time a type is encountered, irrespective of whether the type has been "seen" before and the resulting {un,}boxing LLVM IR code already existing. {un,}boxing code is particularly taxing for the LLVM optimisers because it often contains deeply nested if statements to deal with error handling.

Potential fixes:
1. Create functions containing the LLVM IR for {un,}boxers and cache these against type, then just call them if there's a match for the type in the cache.
2. As 1. but instead just cache the entry block to the {un,}boxer and emit a jump if there's a match for the type in the cache.

CC @eric-wieser @sklam

eric-wieser:
> 2\. but instead just cache the entry block

I don't see how this can work, since you need a way to return back to the calling jit'd function

PR #5422, opened 2020-03-25
Replace NativeValue.is_error with NativeValue.error_blk

eric-wieser:
Done as an experiment to fix problems in #5417. Likely has typos for now.

The idea here is to use an exception-like model for propagating errors from unboxing, rather than using an error-code model. This means that for `native = unbox`, `native.error_blk` now points to an unterminated block that will be jumped to if an error occurs (or `None` if the unboxing cannot fail).

As a bonus, I suspect this now makes it a compilation error to not attempt to handle unboxing errors, as it would result in `native.error_blk` being unterminated.

### DIFF
--- a/numba/core/boxing.py
+++ b/numba/core/boxing.py
@@ -21,10 +21,13 @@ def box_bool(typ, val, c):
 
 @unbox(types.Boolean)
 def unbox_boolean(typ, obj, c):
+    fail_blk = c.builder.append_basic_block('unbox_boolean.fail')
     istrue = c.pyapi.object_istrue(obj)
     zero = ir.Constant(istrue.type, 0)
     val = c.builder.icmp_signed('!=', istrue, zero)
-    return NativeValue(val, is_error=c.pyapi.c_api_error())
+    with c.builder.if_then(c.pyapi.c_api_error(), likely=False):
+        c.builder.branch(fail_blk)
+    return NativeValue(val, error_blk=fail_blk)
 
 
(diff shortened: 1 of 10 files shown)


sklam:
I am afraid this approach would mean that the unboxer logic can never be cached/reused. If we are making a major change to the unboxer, I'd prefer to move towards a solution that can solve the compile time issue https://github.com/numba/numba/issues/5236.

sklam:
This makes me think if we should write the (un)boxer code in high-level code; i.e. compiled python code in object-mode. That way the exception propagation can be more transparent to the writer of the (un)boxer.

eric-wieser:
> I am afraid this approach would mean that the unboxer logic can never be cached/reused. I

Is that true? Couldn't you wrap it as something like:
```C
// using GCC label syntax, but this should be easy to model in LLVM
void *cached_unboxer(void *obj, void *callback_label) {
    return do_unboxing(obj, &&fail);
fail:
    goto callback_label;
}
```

sklam:
I have never tried non-local jumps in LLVM, but I don't think this is easy. LLVM assumes static controlflow. It's `indirectbr` instruction (https://llvm.org/docs/LangRef.html#indirectbr-instruction) needs a list all potential destination.  Even if we can model it, I don't think the LLVM optimization will do a good job.

eric-wieser:
> I don't think the LLVM optimization will do a good job.

The optimizer probably doesn't need to do a good job for the error path :)

PR #5423, opened 2020-03-25
Fix #5421: Add notes on printing in CUDA kernels

### LINKED ISSUE #5421: CUDA printing is unreliable
## Reporting a bug

Reproducer:

```python
from numba import cuda

@cuda.jit
def f():
    print("Hello from f")

print("Launch f")
f[1, 1]()

@cuda.jit
def g():
    print(1)

print("Launch g")
g[1, 1]()
```

Expected:

```
Launch f
Hello from f
Launch g
1
```

Actual:

```
Launch f
Launch g
Hello from f
```

gmarkall:
As I may not be the only one who gets caught out by #5421, this PR attempts to improve the documentation surrounding it :-)

It may not be obvious that printing in CUDA kernels is an asynchronous operation, so this commit adds some notes to the manual describing the behaviour and the requirement for synchronization in order to avoid print output from a kernel only appearing during later driver operations.

### DIFF
--- a/docs/source/cuda/cudapysupported.rst
+++ b/docs/source/cuda/cudapysupported.rst
@@ -36,6 +36,14 @@ The following Python constructs are not supported:
 The ``raise`` and ``assert`` statements are supported.
 See :ref:`nopython language support <pysupported-language>`.
 
+Printing of strings and integers is supported (see :ref:`nopython builtin
+support <pysupported-builtin-functions>`), but printing is an asynchronous
+operation - in order to ensure that all output is printed after a kernel
+launch, it is necessary to call :func:`numba.cuda.synchronize`. Eliding the
+call to ``synchronize`` is acceptable, but output from a kernel may appear
+during other later driver operations (e.g. subsequent kernel launches, memory
+transfers, etc.), or fail to appear before the program execution completes.
+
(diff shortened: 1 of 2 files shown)


docs/source/cuda/cudapysupported.rst
```diff
@@ -36,6 +36,14 @@ The following Python constructs are not supported:
 The ``raise`` and ``assert`` statements are supported.
 See :ref:`nopython language support <pysupported-language>`.
 
+Printing of strings and integers is supported (see :ref:`nopython builtin
```
stuartarchibald: I thought floats worked too? e.g.
```python
from numba import cuda
import numpy as np

@cuda.jit
def g(x):
    print(x[2])

z = np.arange(5.)
g[1, 1](z)

```

gmarkall: Yes, they do - I should have put "numbers" rather than "integers" - now fixed. Many thanks for spotting!

stuartarchibald: I'm not sure it can do complex domain?

gmarkall: Doh! Another fix incoming...

PR #4769, opened 2019-10-31
objmode complex type spelling

sklam:
Fixes https://github.com/numba/numba/issues/4401
Fixes https://github.com/numba/numba/issues/3372
Fixes https://github.com/numba/numba/issues/3356

- [x] ~allow the use of global name for typename~ (removed; prefer using non-str reference to global and cell vars)
- [x] allow the use of cell vars
- [x] allow the use of global variable

PR #5425, opened 2020-03-25
Improve missing type annotation errmsg

amosbird:
The old message may confuse users with Python's type annotation, which had me tried this instead:

```
with objmode():
    start: float64 = time.time()
```

### DIFF
--- a/numba/core/withcontexts.py
+++ b/numba/core/withcontexts.py
@@ -225,7 +225,10 @@ class _ObjModeContextType(WithContext):
         # Verify that all outputs are annotated
         not_annotated = set(stripped_outs) - set(typeanns)
         if not_annotated:
-            msg = 'missing type annotation on outgoing variables: {}'
+            msg = (
+                'missing type annotation on outgoing variables: {0}'
+                'Example code: with objmode({0}=<add_type_as_str_here>):'
+            )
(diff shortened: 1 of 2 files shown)


stuartarchibald:

numba/core/withcontexts.py
```diff
@@ -225,7 +225,10 @@ def strip_var_ver(x):
         # Verify that all outputs are annotated
         not_annotated = set(stripped_outs) - set(typeanns)
         if not_annotated:
-            msg = 'missing type annotation on outgoing variables: {}'
+            msg = (
+                'missing type annotation on outgoing variables: {0}'
+                'Example code: with objmode({0}=<add_type_here>):'
```
stuartarchibald, line 230: I think at present the `<add_type_here>` is required to be a string spelling of the type, e.g. for an out-going `y` variable of type 1D intp array, `y='intp[:]'`.

eric-wieser: This seems like a weird restriction

stuartarchibald: In hindsight it is a bit weird but made some other things easier at the time of writing, this feature is/was pretty involved. This limitation is being fixed in https://github.com/numba/numba/pull/4769.

PR #5426, opened 2020-03-25
Fix a SystemError if unboxing fails in a ufunc

eric-wieser:
Previously the conversion function could be called while an error was in flight, which isn't legal.

Output on master:
```
In [8]: f(arr, out)
RuntimeError                              Traceback (most recent call last)
<ipython-input-6-bcca0651d5c8> in __int__(self)
      2     def __int__(self):
----> 3         raise RuntimeError()
      4

RuntimeError:

The above exception was the direct cause of the following exception:

SystemError                               Traceback (most recent call last)
<ipython-input-8-ec85c84fdad7> in <module>
----> 1 f(arr, out)

<ipython-input-6-bcca0651d5c8> in __int__(self)
      1 class BadType:
      2     def __int__(self):
----> 3         raise RuntimeError()
      4
      5

SystemError: <class 'RuntimeError'> returned a result with an error set
```


stuartarchibald:

eric-wieser:
Any thoughts on this one?

PR #5431, opened 2020-03-26
Add tolerance to numba extension module entrypoints.

stuartarchibald:
This makes it so that if there's a problem with an extension module
and loading it fails, that it doesn't prevent Numba from working.

Warnings are now issued. Tests added.

### DIFF
--- a/numba/core/entrypoints.py
+++ b/numba/core/entrypoints.py
@@ -1,4 +1,5 @@
 import logging
+import warnings
 
 from pkg_resources import iter_entry_points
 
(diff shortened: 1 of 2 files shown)


stuartarchibald:
CC @seibert

numba/core/entrypoints.py
```diff
@@ -20,5 +21,11 @@ def init_all():
...
-        func = entry_point.load()
-        func()
+        try:
+            func = entry_point.load()
+            func()
+        except Exception as e:
+            msg = "Numba extension module '{}' failed to load due to '{}({})'."
+            warnings.warn(msg.format(entry_point.module_name, type(e).__name__,
+                                     str(e)), stacklevel=2)
+            logger.debug('Extension loading failed for: %s', entry_point)
```
eric-wieser:
```suggestion
            logger.debug('Extension loading failed for: %s', entry_point, exc_info=True)
```

stuartarchibald: Thanks for the suggestion, however this PR is merged, please feel free to open a new PR with the suggested change. Thanks!

PR #5436, opened 2020-03-26
Fix #5408: test_set_registers_57 fails on Maxwell

### LINKED ISSUE #5408: test_set_registers_57 fails (possibly specific to Maxwell)
## Reporting a bug

Running the testsuite on a GTX 960 yields this failure:

```
FAIL: test_set_registers_57 (numba.cuda.tests.cudadrv.test_linker.TestLinker)
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudadrv/devices.py", line 224, in _require_cuda_context
    return fn(*args, **kws)
  File "/home/gmarkall/numbadev/numba/numba/cuda/tests/cudadrv/test_linker.py", line 92, in test_set_registers_57
    self.assertEquals(57, compiled._func.get().attrs.regs)
AssertionError: 57 != 56
```

This was also reported by @stuartarchibald when discussing #4954:

> The same on @stuartarchibald's machine with a GTX 750 Ti doesn't segfault, but has the following failure:
>
> ```
> numba/cuda/tests/cudadrv/test_linker.py", line 93, in test_set_registers_57 self.assertEquals(57, compiled._func.get().attrs.regs) AssertionError: 57 != 56
> ```

So this may be an issue specific to Maxwell-architecture GPUs.

gmarkall:
The test_set_registers_* tests were a little too strict in what they checked - the max_registers option should limit the number of registers used, but does not guarantee that as many registers as the limit are used. When linking for some devices, fewer than the maximum are used.

To remedy this, the tests now check that there are less than or equal to the max registers used. An additional test is added, to ensure that the maximum register count would otherwise have been exceeded.

### DIFF
--- a/numba/cuda/tests/cudadrv/test_linker.py
+++ b/numba/cuda/tests/cudadrv/test_linker.py
@@ -86,17 +86,26 @@ class TestLinker(CUDATestCase):
         self.assertTrue(A[0] == 123 + 2 * 321)
 
     @require_context
+    def test_set_registers_no_max(self):
+        """Ensure that the jitted kernel used in the test_set_registers_* tests
+        uses more than 57 registers - this ensures that test_set_registers_*
+        are really checking that they reduced the number of registers used from
+        something greater than the maximum."""
+        compiled = cuda.jit(function_with_lots_of_registers)
+        compiled = compiled.specialize(np.empty(32), *range(6))
+        self.assertGreater(compiled._func.get().attrs.regs, 57)
+
+    @require_context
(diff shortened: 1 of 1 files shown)


stuartarchibald:
Also, manually tested this on a GTX 750 Ti, CUDA 10.2, worked fine.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_43`.

stuartarchibald:
Buildfarm passed.

PR #5437, opened 2020-03-26
Fix #5429: Exception using `.get_ipc_handle(...)` on array from `as_cuda_array(...)

### LINKED ISSUE #5429: Exception using `.get_ipc_handle(...)` on array from `as_cuda_array(...)
## Reporting a bug

<hr>

Am trying to access `.get_ipc_handle(...)` on a Numba `DeviceNDArray` created through `as_cuda_array(...)`. Here's some code using CuPy that demonstrates this use case:

```python
import cupy
import numba.cuda

a = cupy.arange(5)
a2 = numba.cuda.as_cuda_array(a)

a2.get_ipc_handle()
```

However am getting the following exception when calling `.get_ipc_handle(...)`:

```python
  File "script.py", line 7, in <module>
    a2.get_ipc_handle()
  File "/datasets/jkirkham/miniconda/envs/rapids13dev/lib/python3.7/site-packages/numba/cuda/cudadrv/devicearray.py", line 298, in get_ipc_handle
    ipch = devices.get_context().get_ipc_handle(self.gpu_data)
  File "/datasets/jkirkham/miniconda/envs/rapids13dev/lib/python3.7/site-packages/numba/cuda/cudadrv/driver.py", line 847, in get_ipc_handle
    memory.owner.handle,
AttributeError: 'cupy.core.core.ndarray' object has no attribute 'handle'
```

It appears this code that is trying to compute the `offset` causes the issue

https://github.com/numba/numba/blob/cc7e7c7cfa6389b54d3b5c2c95751c97eb531a96/numba/cuda/cudadrv/driver.py#L850

cc @Iroy30 @kkraus14

gmarkall:
In the implementation of `get_ipc_handle`, the owner of a MemoryPointer is assumed to be another MemoryPointer instance (and therefore has a `handle` attribute which points to the start of the allocation in which the MemoryPointer's value points to). This was always true when MemoryPointers were only created by slicing DeviceNDArrays - however, MemoryPointers can also now have owners that are objects coming in through the CUDA array interface (e.g. via `cuda.as_cuda_array()`).

This commit removes the assumption that the owner of a MemoryPointer is another MemoryPointer, and instead gets the start of the allocation with `cuMemGetAddressRange` (via the already-extant `device_extents` function).

`get_devptr_for_active_ctx()` also returned the wrong type for passing to any driver functions - it is now fixed to return a `cu_device_ptr`.

Tests are added to test_ipc that emulate an array from another library being used via the CUDA Array Interface. The mechanism for this test is common to test_ipc and test_cuda_array_interface, so the `MyArray` class from test_cuda_array_interface is moved to `numba.cuda.testing` as `ForeignArray`.

A couple of superfluous print calls are also removed from test_ipc, in the interests of tidiness.

With this commit, the original reproducer:

```python
import cupy
import numba.cuda

a = cupy.arange(5)
a2 = numba.cuda.as_cuda_array(a)

a2.get_ipc_handle()
```

runs to completion without error.

cc @jakirkham for comment / testing :-)

### DIFF
--- a/numba/cuda/cudadrv/driver.py
+++ b/numba/cuda/cudadrv/driver.py
@@ -834,13 +834,11 @@ class NumbaCUDAMemoryManager(HostOnlyCUDAMemoryManager):
         return MemoryInfo(free=free.value, total=total.value)
 
     def get_ipc_handle(self, memory):
+        base, end = device_extents(memory)
         ipchandle = drvapi.cu_ipc_mem_handle()
-        driver.cuIpcGetMemHandle(
-            byref(ipchandle),
-            memory.owner.handle,
-        )
+        driver.cuIpcGetMemHandle(byref(ipchandle), base)
         source_info = self.context.device.get_device_identity()
-        offset = memory.handle.value - memory.owner.handle.value
+        offset = memory.handle.value - base
 
         return IpcHandle(memory, ipchandle, memory.size, source_info,
                          offset=offset)
@@ -2075,7 +2073,7 @@ def get_devptr_for_active_ctx(ptr):
     """Query the device pointer usable in the current context from an arbitrary
     pointer.
     """
-    devptr = c_void_p(0)
+    devptr = drvapi.cu_device_ptr()
     if ptr != 0:
         attr = enums.CU_POINTER_ATTRIBUTE_DEVICE_POINTER
         driver.cuPointerGetAttribute(byref(devptr), attr, ptr)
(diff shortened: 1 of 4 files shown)


jakirkham:
Thanks for working on this @gmarkall! 😄

Yeah this makes a lot of sense to me. It's similar to [CuPy's `UnownedMemory`]( https://docs-cupy.chainer.org/en/stable/reference/generated/cupy.cuda.UnownedMemory.html ), which fills a similar roll.

Iroy30:
Hi. I checked out the proposed changes and it successfully fixes the "AttributeError: 'Buffer' object has no attribute 'handle' " error when calling get_ipc_handle() described in https://github.com/rapidsai/cugraph/issues/790. Thank you!

stuartarchibald:
@jakirkham is this something you'd be able to take a look at review wise? Many thanks.

PR #5438, opened 2020-03-26
Update package deps for doc building.

stuartarchibald:
Swap bootstrap for RTD.


esc:
This appears correct, now that we have changed the theme.

PR #5287, opened 2020-02-20
First-class function type

pearu:
Closes #3405 

This PR (as a continuation of PR #4967) implements:
- [x] Major revision on how Dispatcher and FunctionType instances interact.
- [x] Full support for jit-decorated functions as first-class type functions
    - The jit-decorated functions are considered first-class type functions when using `firstclass=True` option. For instance:
        ```
        @numba.njit
        def a(x):  # `a` is first-class type function, that is,
                   # it can be used as an object in other jit-decorated functions
            return x + 1

        @numba.jit()
        def foo(f, x):
            return f(x)

        assert foo(a, 3) == 4
        ```

PR #5441, opened 2020-03-26
Additional changes to first class function

sklam:
base on #5287

Addresses items in https://github.com/numba/numba/pull/5287#issuecomment-604464489


numba/core/utils.py
```diff
@@ -495,6 +493,10 @@ def unified_function_type(numba_types):
     If any of the Numba function types is a Dispatcher type, the
     unified function type will be UndefinedFunctionType instance.
     """
+    from numba.core.errors import NumbaExperimentalFeatureWarning
```
stuartarchibald: Move import to the top? Or is there circular import problems.

sklam:
cuda smoke test at `numba_smoketest_cuda_42`

numba/core/utils.py
sklam: it's a circular import problem.

stuartarchibald:
Seems like the warning triggers a lot more often than expected?!

sklam:
The warning needs to be moved under the guard. The `unified_function_type` function is used when it doesn't know the inputs contain function types.

sklam:
cuda smoke test passed.

docs/source/reference/types.rst
```diff
@@ -100,6 +100,122 @@ the beginning or the end of the index specification::
...
+========================   ==============   ============    ===========
+Context                    cfunc compiled   jit compiled    WAP objects
+========================   ==============   ============    ===========
+Can be used as arguments   yes              yes             yes
+Namespace scoping          yes              yes             no
+Can be called              yes              yes             yes
+Can be used as items       yes              yes             yes
+Automatic overload         no               yes             no
+Can be returned            yes              yes             yes
+========================   ==============   ============    ===========
```
eric-wieser: nit: consider sorting the `no`s together for easier reading
```suggestion
========================   ============   ==============   ===========
Context                    jit compiled   cfunc compiled   WAP objects
========================   ============   ==============   ===========
Can be used as arguments   yes            yes              yes
Can be called              yes            yes              yes
Can be used as items       yes            yes              yes
Can be returned            yes            yes              yes
Namespace scoping          yes            yes              no
Automatic overload         yes            no               no
========================   ============   ==============   ===========
```
(swapped rows and columns)

pearu: Good idea.

The table needs the following update as well:
`Namespace scoping` for `WAP` is `yes`.

docs/source/reference/types.rst
```diff
@@ -100,6 +100,122 @@ the beginning or the end of the index specification::
...
+Wrapper Address Protocol (WAP, see below) with the following restrictions:
+
+========================   ============   ==============   ===========
+Context                    jit compiled   cfunc compiled   WAP objects
+========================   ============   ==============   ===========
+Can be used as arguments   yes            yes              yes
+Can be called              yes            yes              yes
+Can be used as items       yes            yes              yes
+Can be returned            yes            yes              yes
+Namespace scoping          yes            yes              no
```
sklam:
```suggestion
Namespace scoping          yes            yes              yes
```

PR #5446, opened 2020-03-27
Update CHANGE_LOG for 0.49

stuartarchibald:
As title.

### DIFF
--- a/CHANGE_LOG
+++ b/CHANGE_LOG
@@ -1,11 +1,5 @@
-Version 0.50.0dev
------------------
-
-In development
-
-
-Version 0.49.0 (April 9, 2020)
--------------------------------
+Version 0.49.0 (Apr 9, 2020)
+----------------------------
 
 This release is very large in terms of code changes. Large scale removal of
 unsupported Python and NumPy versions has taken place along with a significant
--- a/CHANGE_LOG
+++ b/CHANGE_LOG
@@ -1,7 +1,253 @@
-Version 0.49.0dev
+Version 0.50.0dev
 -----------------
 
 In development
+
+
+Version 0.49.0 (April 9, 2020)
+-------------------------------
+
+This release is very large in terms of code changes. Large scale removal of
+unsupported Python and NumPy versions has taken place along with a significant
+amount of refactoring to simplify the Numba code base to make it easier for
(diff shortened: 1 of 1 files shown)


CHANGE_LOG
```diff
@@ -1,9 +1,255 @@
-Version 0.49.0dev
+Version 0.50.0dev
```
sklam: I think we should defer mentioning 0.50.0dev. We should do this when we tag 0.50.0dev on the master branch.

sklam: also, so that the release branch will have 0.49.0 as the first entry in the change log

CHANGE_LOG
```diff
@@ -1,9 +1,255 @@
...
+important changes to solve a number of long standing issues. In addition some
+new features have been added and a large number of bugs have been fixed!
+
+Highlights of core feature changes include:
+
+* Removal of all Python 2 related code and also updating the minimum supported
+  Python version to 3.6, the minimum supported NumPy version to 1.15 and the
+  minimum supported SciPy version to 1.0. (Stuart Archibald).
+* Refactoring of the Numba code base. The code is now organised into submodules
+  by functionality, this cleaning Numba's top level namespace
```
sklam: > this cleaning Numba's top level namespace

extra? runon?

```suggestion
  by functionality. This cleans up Numba's top level namespace.
```

CHANGE_LOG
stuartarchibald: fair point, fixed!

CHANGE_LOG
```diff
@@ -1,9 +1,255 @@
...
+  (Pearu Peterson).
+
+Enhancements from user contributed PRs (with thanks!):
+
+* Aaron Meurer added support for thread masking at runtime in #4615.
+* Andreas Sodeur fixed a long standing bug that was preventing ``cProfile`` from
+  working with Numba JIT compiled functions in #4476.
+* Arik Funke fixed error messages in ``test_array_reductions`` (#5278) and fixed
+  an issue with test discovery (#5239).
+* Antonio Russo Fixed a SyntaxWarning in #5252.
```
sklam:
```suggestion
* Antonio Russo fixed a SyntaxWarning in #5252.
```

PR #5447, opened 2020-03-27
Handle idx and size unification in wrap_index manually.

### LINKED ISSUE #5442: prange working with numpy?
## Reporting a bug

- [x ] I am using the latest released version of Numba (most recent is visible in

```python

@njit(parallel=True)
def f(arr):
    out = 0.0
    for i in prange(100):
        out += np.sum(arr[i:(i + 2)])
        # but following line is ok!
        out += 1


if __name__ == "__main__":
    import numpy as np
    arr = np.random.randn(10000)
    f(arr)

```

```
LLVM IR parsing error
<string>:499:18: error: invalid cast opcode for cast from 'i64' to 'double'
  %".408" = sext i64 %".406" to double
                 ^


File "test.py", line 6:
def f(arr):
    <source elided>
    out = 0.0
    for i in prange(100):
    ^

[1] During: lowering "id=1[LoopNest(index_variable = parfor_index.40, range = (0, $const10.1, 1))]{18: <ir.Block at test.py (6)>}Var(parfor_index.40, test.py:6)" at test.py (6)
```

```bash
>>> numba.__version__
'0.49.0dev0+995.gd0c7923c5'
```

DrTodd13:
Resolves #5442

wrap_index has to compare idx and size and to do that they need to be the same bitwidth.  If one is signed and the other unsigned then the old code to unify to the same bitwidth (based on numpy rules) would also switch the type to double.  In this PR, we keep the incoming integer types but just extend the bitwidth to be the max of the incoming types.  We handle signedness extension correctly as well.


xuanqing94:
Thank you!

stuartarchibald:
Thanks for the PR @DrTodd13, any chance the originally failing minimal working reproducer(s) could be added as test(s) please to demonstrate the fix. Thanks again!

stuartarchibald:
Ping @DrTodd13 RE https://github.com/numba/numba/pull/5447#issuecomment-607824209, thanks!

numba/parfors/array_analysis.py
```diff
@@ -85,17 +85,45 @@ def wrap_index(typingctx, idx, size):
     (idx < 0 ? idx + size : idx) because we may have situations
     where idx > size due to the way indices are calculated
     during slice/range analysis.
+
+    Both idx and size have to be Integer types.
+    size should be from the array size vars that array_analysis
+    adds and the bidwidth should match the platform maximum.
```
stuartarchibald, line 91: If the bitwidth should match platform max, could this code just `{z,s}ext` to `intp.bitwidth` and or in the signedness on the chosen type?

DrTodd13: Are you just suggesting changing "unified_bitwidth = max(idx.bitwidth, size.bitwidth)" to "unified_bitwidth = types.intp.bitwidth"?

DrTodd13: I simplified it a bit.  Appreciate your input.

esc:
@DrTodd13 I think there might be a git issue lurking here somewhere?

PR #5450, opened 2020-03-27
Docs: remove numbers and add titles to TOC

luk-f-a:
This PR removes the numbers in the table of contents, e.g. "1.2 Overview" becomes simple "Overview".
Additionally, it splits the main index in three sections: "For all users", "For GPU users", "For advanced users and developers". The main effect is to add coloured titles in the TOC, splitting it in three parts and making it easier to read.

I would also split the user manual in two parts, since it has two many subitems. However, I haven't been able to think of a meaningful way to split it into clearly defined parts.


docs/source/index.rst
```diff
@@ -11,14 +11,26 @@ with Numba, we suggest you start with the :doc:`User manual <user/index>`.
...
+   :caption: For GPU users
+   :maxdepth: 2
+
    cuda/index.rst
    cuda-reference/index.rst
    roc/index.rst
+
+
+.. toctree::
+   :caption: For advanced users and developers
```
stuartarchibald: Locally, this ends up too wide for the sidebar, I can just see `For advanced users and developer` (and half of the `s`), any ideas?

docs/source/index.rst
```diff
@@ -11,14 +11,26 @@ with Numba, we suggest you start with the :doc:`User manual <user/index>`.
...
-   :numbered:
+   :caption: For all users
    :maxdepth: 2
 
    user/index.rst
    reference/index.rst
+
+
+.. toctree::
+   :caption: For GPU users
```
stuartarchibald: Perhaps this should differentiate between CUDA and ROCm GPU users. There's the cuda guide and reference that could go under cuda.

docs/source/index.rst
```diff
@@ -11,14 +11,26 @@ with Numba, we suggest you start with the :doc:`User manual <user/index>`.
 
 
 .. toctree::
-   :numbered:
```
stuartarchibald: I don't think this will be missed, having built the docs locally, I cannot think of a time when I've needed a section number apart from to convey information to another person about something that's broken when debugging doc builds. Also having enumerations like `2.1.3.4` doesn't seem particularly useful.

stuartarchibald:
@seibert @sklam @esc @gmarkall, you all commented on the original doc theme change. Any views on the above? I think it's a further improvement and perhaps a candidate for the 0.49.0 release branch.

esc:
Sounds good, I am 👍 to backport.

docs/source/index.rst
luk-f-a: If we tried to differentiate, then ROCm would end up with only one line below the caption. Since there are only 3 items below this caption, personally I wouldn't split them further.

docs/source/index.rst
luk-f-a: I struggled with this one (hadn't noticed the `s` was missing). `For advanced users / developers`, `For advanced users & developers`?
Alternative I could remove the `For` in all captions.

esc:
I'll add it to the 0.49 milestone, we should take this along for the ride.

docs/source/index.rst
gmarkall: Either of / or & look good to me. I think keeping the "For" would be nice.

gmarkall:
Looks good to me - I was a little surprised by how much more pleasant it is to read the docs without tons of numbers everywhere. On my system "For advanced users and developers" did fit in the sidebar... just - but I see the need to shorten it a little.

docs/source/index.rst
stuartarchibald: Agree with @gmarkall

docs/source/index.rst
stuartarchibald: I find it strange to lump different technologies together based on target hardware kind. I personally am ok with one line for ROCm and two for CUDA and have built the docs with this in and it seems ok. Anyone else have a view?

esc:
I built it locally and it looks awesome! I have one small request, I just noticed, that the docs at:

http://numba.pydata.org/numba-doc/latest/user/installing.html#dependency-list

still reference `sphin-bootstap` but what is needed now is `sphinx_rtd_theme`. Could you sneak that change into this PR?

luk-f-a:
@esc I don't think sphinx-bootstap here: http://numba.pydata.org/numba-doc/dev/user/installing.html#dependency-list

I think this change was taken care of in the original PR.

esc:
> @esc I don't think sphinx-bootstap here: http://numba.pydata.org/numba-doc/dev/user/installing.html#dependency-list
> 
> I think this change was taken care of in the original PR.

Oh, nevermind the noise then.

esc:
It looks good from my side also. @stuartarchibald what next, wait to merge this and then backport shortly before 0.49.0? I've added a needs-backport label to track these.

stuartarchibald:
> It looks good from my side also. @stuartarchibald what next, wait to merge this and then backport shortly before 0.49.0? I've added a needs-backport label to track these.

We'll just cherry-pick everything on the 0.49.0 milestone into the 0.49.0 release branch, don't think there's a backport need as we've not released a final 0.49.0, these are just fixes against an RC?

esc:
o.k. just want to make sure we don't miss anything. Closed issues in the Milestone will suffice.

stuartarchibald:
No comments made otherwise so I think this is good to merge. Thanks all!

PR #5451, opened 2020-03-28
instance cache for UndefinedFunction

luk-f-a:
The current implementation of `UndefinedFunction` (#5287) has a quirk:

```python
from numba import njit

@njit
def foo1(x):
    return x + 1

print(typeof((foo1,))== typeof((foo1,))) #False
```
which leads to unexpected behaviour like

```python
@njit
def foo1(x):
    return x + 1

@njit
def bar(fcs):
    fc = fcs[0]
    return fc(1)

print(bar((foo1,)))
bar.disable_compile()
print(bar((foo1,)))
# TypeError: No matching definition for argument type(s) UniTuple(UndefinedFunctionType[undefined(undefined)7] x 1)
```
This PR attempts to make the instance unique for each set of Dispatcher instances.

UPDATE:  This PR does not attempt to solve wider problems regarding the re-use of higher-order functions. This PR only addresses the particular problem of `typeof((foo1,))!= typeof((foo1,))`. However, I believe this small problem has to be solved before bigger problems can be tackled.


pearu:
I agree that
```
typeof((foo1,))== typeof((foo1,))
```
ought to return True but only because of the dispatcher instances `foo1` in both sides of the equality are the same.

However, when considering the case with two functions:
```
@njit
def foo1(x):
    return x + 1

@njit
def foo2(x):
    return x + 1
```
the result of
```
typeof((foo1,))== typeof((foo2,))
```
ought to be False even when the Python implementations of the `foo1` and `foo2` are identical.

Notice that the following example does not work either:
```
from numba import njit
@njit
def foo1(x):
    return x + 1

@njit
def bar(fcs, y):
    fc = fcs[0]
    return fc(y)

print(bar((foo1,), 2))
bar.disable_compile()
foo1.disable_compile()
print(bar((foo1,), 2))  # TypeError
```
although `typeof((foo1,))== typeof((foo1,))` returns `True`. So, the proposed PR may not be sufficient to fix the underlying problem: the second call of `bar((foo1,), 2)` after disabling compile likely works but `bar((foo2,), 2)` will not, for instance.

Depending on the application, consider the following example:
```
from numba import njit, int64
@njit
def foo1(x):
    return x + 1
@njit
def foo2(x):
    return x + 1

@njit
def bar(fcs, y):
    fc = fcs[0]
    return fc(y)

foo1.compile(int64(int64))
foo1.disable_compile()  # make typeof((foo1, )) precise
print(bar((foo1,), 2))
bar.disable_compile()
print(bar((foo1,), 2))  # works
foo2.compile(int64(int64))
foo2.disable_compile()  # make typeof((foo2, )) precise
print(bar((foo2,), 2))   # works
```

It is possible to eliminate the need to compile `foo2` after `bar` compilation is disabled because the required function type can be derived from `bar` compile result signature.

luk-f-a:
hi @pearu, thanks for looking at this. Right now, this is just a draft, I was testing some ideas and wanted to see if they would pass CI. I remember that you told me that making UndefinedFunction a singleton would not work, and I wanted to test an alternative idea: returning a consistent instance when the set of Dispatchers is the same. Unfortunately CI is not working, so I don't know if this works yet.

I agree that this does not fix the underlying problem, which probably requires bigger changes than a few lines as I did here. I will add a note to my description above to be clear about that. However, this PR, or something similar, is a step in the right direction. I don't think a long-term solution can be applied as long as `typeof(a)!=typeof(a)`.


Some replies to your comments:

> `typeof((foo1,))== typeof((foo1,))` ought to return True but only because of the dispatcher instances foo1 in both sides of the equality are the same.

Completely agree, this is what this PR attempts to ensure.

> the result of `typeof((foo1,))== typeof((foo2,))` ought to be False even when the Python implementations of the foo1 and foo2 are identical.

Complete agree too, and this PR does not attempt to do that.

> It is possible to eliminate the need to compile foo2 after bar compilation is disabled because the required function type can be derived from bar compile result signature.

Complete agree again, and much more than this PR is needed for that to happen.

 In my view, the reason why that should be possible is that both `foo1`and `foo2` are intersection types, and both are subtypes of `int64->int64`. To solve the underlying problem, we need to away to automatically detect or to let users declare subtyping relationships between Dispatcher types. I'm preparing a proposal to add a limited form of subtyping in numba, which would take us one step closer to managing the general case of being able to re-use higher-order functions. I'm still working on the proposal, and in any case will not release it before 0.49 is out, since all core-devs are busy with the release.

stuartarchibald:
@luk-f-a any chance you could please merge in master and resolve conflicts? Thanks.

stuartarchibald:
Please could the conflicts be resolved by merging master into this branch. Thanks!

pearu:
Re failure:
```
  File "/home/vsts/work/1/s/numba/tests/test_function_type.py", line 1154, in test_type_consistency_set
    self.assertTrue(typeof((foo1, foo2)) == typeof((foo2, foo1)))
```
Since (i) `foo1` and `foo2` are dispatcher instances that have their own type and (ii) the `typeof` does not trigger the function type unification, the result of `typeof((foo1, foo2))` and `typeof((foo2, foo1))` are different:
```
>>> typeof((foo1, foo2))
Tuple(type(CPUDispatcher(<function foo1 at 0x7fc6b2e4bee0>)), type(CPUDispatcher(<function foo2 at 0x7fc6b2e4bf70>)))
>>> typeof((foo2, foo1))
Tuple(type(CPUDispatcher(<function foo2 at 0x7fc6b2e4bf70>)), type(CPUDispatcher(<function foo1 at 0x7fc6b2e4bee0>)))
```
So the assert error is expected.

Note that with the current master
```
>>> typeof((foo1,)) == typeof((foo1,))
True
```
so that the grounds of this issue is already resolved.

luk-f-a:
@pearu  thanks for the analysis, you're right, `typeof((foo1, foo2))==typeof((foo2, foo1))` is something that made sense in 0.49rc1, but not in 0.49.

While it's true that `typeof((foo1,)) == typeof((foo1,))` under current master, wouldn't it fail in #5729?

pearu:
> While it's true that `typeof((foo1,)) == typeof((foo1,))` under current master, wouldn't it fail in #5729?

Yes, it will result in `False` because the default purpose of `typeof` input is `argument(=1)`. However, when switching the purpose to `parameter(=2)`, we have
```
>>> typeof((foo1,), 2) == typeof((foo1,), 2)
True
```
I think within #5729, the `UndefinedFunction` feature may require a slight revision. For instance, the real meaning of `UndefinedFunction` is `To-Be-Defined-Function` so that the life-time of `UndefinedFunction` must be defined better.

numba/core/types/function_type.py
```diff
@@ -110,6 +117,12 @@ def __init__(self, nargs, dispatchers):
         # make the undefined function type instance unique
         type(self)._counter += 1
         self._key += str(type(self)._counter)
+        self.instance_cache[frozenset(dispatchers)] = self
```
sklam, line 120: The `instance_cache` is holding Dispatcher objects alive and may lead to memory leak for long running process. At the same time, it seems tricky to make it release memory automatically using a container from `weakref`.

luk-f-a: @sklam, I've been thinking about how can we re-use type instances without necessarily holding live references to a lot of objects. It's very similar problem as this one https://github.com/numba/numba/pull/5579#issuecomment-617371499. ie, . In there you mention using `.key` to identify repeating objects.
Alternatively,  my solution in https://github.com/luk-f-a/numba/blob/222f0fd33082bd39b18a9adec97d4a18469a2599/numba/core/types/abstract.py#L48-L74 was to use the hash (instead of the object) as key in the dictionary. 

I think either way could work here. If you tell me which one you prefer, I'll make the change.

stuartarchibald:
Following OOB discussion on IRC, moving this to 0.51.x.

numba/core/types/function_type.py
sklam: Using `.key` is better. The `Type.__hash__` is by default `hash(type_instance.key)`. The hash can't guarantee unique result for different `.key`

sklam: I see why you need to use hash. Some `.key` are containing type instances.

luk-f-a: What about changing `key` to be weak references or creating a new `weak_key` property holding weak references or a `string_key` which is only text?

PR #5452, opened 2020-03-29
Add version information

TejasAvinashShetty:
Add version information for scipy, numpy, python.
Most of Numba users use at least numpy and python.
And in the future, 90% of them would also use scipy.
This is for issue template. So no need for unit tests

### DIFF
--- a/.github/ISSUE_TEMPLATE/Bug_report.md
+++ b/.github/ISSUE_TEMPLATE/Bug_report.md
@@ -30,6 +30,14 @@ Before submitting a bug report please ensure that you can check off these boxes:
 - [ ] I have included below a minimal working reproducer (if you are unsure how
  to write one see http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports).
 
+
+#### Scipy/Numpy/Python version information:
+<!-- You can simply run the following and paste the result in a code block
+```
+import sys, scipy, numpy; print(scipy.__version__, numpy.__version__, sys.version_info)
+```
+-->
+
 <!--
 
 Please include details of the bug here, including, if applicable, what you


TejasAvinashShetty:
Inspired by the scipy/scipy bug report template https://github.com/scipy/scipy/issues/new/choose

stuartarchibald:
Thanks for the patch. I'm not sure that frequency at which this information is required merits the cost of asking for it constantly. Typically, if additional information is needed it's usually Numba versions or information about the installation/OS, which `numba -s` provides, but again, this isn't often required. Also, CI testing tends to pick up problems with NumPy, SciPy and Python itself. Anyone else have a different view?

esc:
@stuartarchibald I agree, this might raise the bar too high and we need to be very deliberate when asking for information for bug reports or else we may discourage users from reporting. If we would be asking this question constantly, then such a change would be warranted.

@TejasAvinashShetty thanks for submitting this, we do really appreciate your efforts in contributing to Numba. I am sorry already that we may have to reject your first PR and I hope this will not discourage you from contributing in future. If you are not sure what to do, there is a label "good first issue" label in the issue tracker: https://github.com/numba/numba/labels/good%20first%20issue or come chat with us on Gitter (https://gitter.im/numba/numba)!

TejasAvinashShetty:
Thanks @esc and @stuartarchibald for replying back. Regarding the PR slightly disappointed.
Could you please at least let me know as to why the checks are failing? I really could not wrap my head around it.

esc:
@TejasAvinashShetty the checks are failing because `master` is/was broken and have nothing to do with your PR. There is a fix at #5455 which should remedy the issues.

stuartarchibald:
Yes, sorry to have to let you down, it's very good of you to suggest a patch, so thank you, but as noted, it's not applicable in practice to the user base at present. As mentioned by @esc, please do try a "good first issue" if you'd like to contribute a patch or come talk to us on gitter. Many thanks!

PR #5457, opened 2020-03-30
implement repr for untyped lists

esc:
This implemes a proper `__repr__` for untyped numba.typed.List and
tests.

### DIFF
--- a/numba/typed/typedlist.py
+++ b/numba/typed/typedlist.py
@@ -397,7 +397,7 @@ class List(MutableSequence):
 
     def __repr__(self):
         body = str(self)
-        prefix = str(self._list_type) if self._typed else "ListType[UNTYPED]"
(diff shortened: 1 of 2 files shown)


numba/typed/typedlist.py
```diff
@@ -397,7 +397,7 @@ def __str__(self):
 
     def __repr__(self):
         body = str(self)
-        prefix = str(self._list_type)
+        prefix = str(self._list_type) if self._typed else "ListType[UNTYPED]"
```
stuartarchibald, line 400: Perhaps a `ListType[Undefined]` ? As in, we know it's a `ListType` but the item type is `Undefined`, I think this makes it more explicit and consistent with other things (which tend to refer to "undefined")?

esc: Fixed

esc:
Thanks for merge.

PR #4610, opened 2019-09-25
Implement np.is* functions

guilhermeleobas:
Still a working in progress but any feedback is appreciated. The goal is to implement all `np.is*` functions listed [here](https://github.com/numba/numba/issues/4074). 

At the moment, only `np.iscomplexobj` and `np.isrealobj` were implemented.

Implementation and tests are based on NumPy counterparts.

- [x] np.isclose - moved to PR https://github.com/numba/numba/pull/7067
- [x] np.iscomplex
- [x] np.iscomplexobj
- [x] np.isneginf
- [x] np.isposinf
- [x] np.isreal
- [x] np.isrealobj
- [x] np.isscalar

--------
- [ ] np.isfortran ([obsolete](https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.isfortran.html)!!!)

PR #5463, opened 2020-03-31
[WIP] Add np.asarray(string) and str() impl

guilhermeleobas:
Following up on #4610, this PR implements `np.asarray('string')` and the `str` builtin for integers.

This still needs some polishing.

### DIFF
--- a/numba/cpython/unicode.py
+++ b/numba/cpython/unicode.py
@@ -2341,6 +2341,22 @@ def ol_chr(i):
             return _PyUnicode_FromOrdinal(i)
         return impl
 
+
+@overload(str)
+def str_overload(n):
+    impl = None
+    if isinstance(n, types.Integer):
+        def impl(n):
+            s = ''
+            while n > 0:
+                c = chr(ord('0') + (n % 10))
+                n = n // 10
+                s = c + s
+            return s
+    if isinstance(n, (types.UnicodeType, types.UnicodeCharSeq, types.CharSeq)):
+        def impl(n):
+            return n
+    return impl
 # ------------------------------------------------------------------------------
 # iteration
 # ------------------------------------------------------------------------------
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -3895,6 +3895,10 @@ def np_asarray(a, dtype=None):
             for i, v in enumerate(a):
                 ret[i] = v
             return ret
+    elif isinstance(a, types.UnicodeType):
+        def impl(a, dtype=None):
+            dt = 'U' + str(len(a))
+            return np.array(a, dtype=np.dtype('U10'))
 
     return impl
 
--- a/numba/np/numpy_support.py
+++ b/numba/np/numpy_support.py
@@ -95,7 +95,6 @@ def from_dtype(dtype):
         return FROM_DTYPE[dtype]
     except KeyError:
         char = dtype.char
-
         if char in 'SU':
             return _from_str_dtype(dtype)
         if char in 'mM':
@@ -112,6 +111,7 @@ _as_dtype_letters = {
     types.NPTimedelta: 'm8',
     types.CharSeq: 'S',
     types.UnicodeCharSeq: 'U',
+    types.UnicodeType: 'U',
 }
 
 
@@ -134,6 +134,11 @@ def as_dtype(nbtype):
     if isinstance(nbtype, (types.CharSeq, types.UnicodeCharSeq)):
         letter = _as_dtype_letters[type(nbtype)]
         return np.dtype('%s%d' % (letter, nbtype.count))
+    if isinstance(nbtype, types.UnicodeType):
+        # XXX: fix this! The dtype should be 'U#len(str)'
+        # print('nbtype', nbtype)
+        letter = _as_dtype_letters[type(nbtype)]
+        return np.dtype('%s%d' % (letter, 0))
     if isinstance(nbtype, types.Record):
         return as_struct_dtype(nbtype)
     if isinstance(nbtype, types.EnumMember):
@@ -600,7 +605,7 @@ def type_can_asarray(arr):
     implementation, False otherwise.
     """
 
-    ok = (types.Array, types.Sequence, types.Tuple,
+    ok = (types.Array, types.Sequence, types.Tuple, types.UnicodeType,
           types.Number, types.Boolean, types.containers.ListType)
 
     return isinstance(arr, ok)


esc:
@guilhermeleobas thanks for submitting this improvement to the Numba issue tracker. You'll want to rebase this onto master once #5455 is merged. I'll mark it as 'In Progress' for now.

guilhermeleobas:
Thanks @esc.

guilhermeleobas:
After [discussing](https://gitter.im/numba/numba?at=5e86116001c85847b2cfcc1f) about UnicodeTypes on gitter, I don't think it is possible to support `np.asarray(UnicodeType)` at the moment. Thus, I am removing the attempt of implementing it from this PR and will focus only on the builtin `str(number)`.

guilhermeleobas:
Ready for a round of reviews

esc:
@guilhermeleobas thanks, I have updated the label. As for the documentation, maybe here:

http://numba.pydata.org/numba-doc/latest/reference/pysupported.html#str

numba/np/numpy_support.py
```diff
@@ -600,7 +599,7 @@ def type_can_asarray(arr):
     implementation, False otherwise.
     """
 
-    ok = (types.Array, types.Sequence, types.Tuple,
+    ok = (types.Array, types.Sequence, types.Tuple, types.Literal,
```
guilhermeleobas, line 602: This shouldn't be here

numba/tests/test_builtins.py
```diff
@@ -505,6 +508,24 @@ def test_hex_npm(self):
...
+
+        cr = compile_isolated(pyfunc, (types.int64,), flags=flags)
+        cfunc = cr.entry_point
+
+        args = [
+            123456789,
+            2**32-1,
+            0,
+            33,
+            -33,
```
stuartarchibald, line 522: Probably ought to test other things that are integers, e.g. `np.uint64` /

guilhermeleobas: Done!

guilhermeleobas:
I think the test failure is not related to any changes introduced in this PR.

stuartarchibald:
Did anyone note down what the failures were? Is there potentially a problem on master?

guilhermeleobas:
If I remember correctly, an issue involving `test_undefined_var` defined on file `numba/numba/tests/test_ssa.py`. The issue was with an import not importing from `numba.core`.

stuartarchibald:
> If I remember correctly, the issue involved `test_undefined_var` defined on file `numba/numba/tests/test_ssa.py`. It was related to an import using the old version rather than importing from `numba.core`.
> 
> Edit: Error only happened in one of the Linux builds.

ok, thanks, sounds like it's not a problem on master!

guilhermeleobas:
PR is ready for a review

numba/tests/test_builtins.py
stuartarchibald: The change made isn't actually testing the types presented, the compile result is for a `(types.int64,)` only, all inputs will just end up being cast.

numba/np/numpy_support.py
```diff
@@ -95,7 +95,6 @@ def from_dtype(dtype):
         return FROM_DTYPE[dtype]
     except KeyError:
         char = dtype.char
-
```
stuartarchibald, line 98: Change not in scope of this PR.

numba/np/arraymath.py
```diff
@@ -3895,7 +3895,6 @@ def impl(a, dtype=None):
             for i, v in enumerate(a):
                 ret[i] = v
             return ret
-
```
stuartarchibald, line 3898: Change not in scope of this PR.

numba/tests/test_builtins.py
guilhermeleobas: Is the expected behaviour a cast to float64 in the operation `n % 10`, where `n` is an integer?

guilhermeleobas:
I had to modify a bit the tests and removed the test using `2**32-1` as input due to a failure on NumPy cast:

```
  File "/home/vsts/work/1/s/numba/tests/test_builtins.py", line 550, in test_str
    self.assertPreciseEqual(cfunc(typ(v)), pyfunc(typ(v)))
  File "/home/vsts/work/1/s/numba/core/types/abstract.py", line 176, in __call__
    return self.cast_python_value(args[0])
  File "/home/vsts/work/1/s/numba/core/types/scalars.py", line 47, in cast_python_value
    return getattr(np, self.name)(value)
OverflowError: Python int too large to convert to C long
```

AndrewEckart:
Rather than overloading `str()` for individual types, would it be better to implement this in `types.Integer.__str__` and then rely on a global `str()` overload? Described in #5536 and I drafted an implementation for the global overload in #5866.

guilhermeleobas:
Hi @AndrewEckart, I will change the implementation to overload `types.Integer.__str__`

guilhermeleobas:
cc @hameerabbasi

numba/cpython/builtins.py
```diff
@@ -595,3 +595,17 @@ def impl(func, iterable):
...
     return impl
+
+# ------------------------------------------------------------------------------
+# str overload
+
+@overload(str)
+def ol_str(s):
+    def impl(s):
+        return s.__str__()
+    return impl
```
hameerabbasi, line 606: This would probably be better off as a lambda too.

numba/cpython/unicode.py
```diff
@@ -2341,6 +2341,27 @@ def impl(i):
...
+        if n == 0:
+            return '0'
+        while n > 0:
+            # there is an implicit conversion of n to float here
+            c = chr(int(ord('0') + (n % 10)))
+            n = n // 10
+            s = c + s
+        if flag:
+            s = '-' + s
+        return s
```
hameerabbasi, line 2363: Can we append to a `list` and then do `''.join(l)`. Your solution is `O(n^2)` I think, where `n` is the number of digits.

guilhermeleobas: You're right. I forgot this detail.

numba/tests/test_builtins.py
hameerabbasi: No, this is a bug:
```python
>>> type(5 % 10)
<class 'int'>
>>> import numpy as np
>>> type(np.int64(5) % np.int64(10))
<class 'numpy.int64'>
```

hameerabbasi: Well, it doesn't cast to `float64`:
```python
>>> import numba as nb
>>> @nb.njit
... def mod(a, b):
...     return a % b
...
>>> type(mod(5, 10))
<class 'int'>
```

numba/cpython/unicode.py
guilhermeleobas: A couple of notes: 
1) List is not supported inside `overload_method`. 
2) Since the largest value will be `max(uint64_t) = 2^64 = 18446744073709551616 => 20 digits`, a `O(n^2)` solution will have almost the same performance as a `O(n)` for `n=20`. If the current solution is not ideal, I can try to scratch something using typed list.

hameerabbasi: In that case I'm okay with this solution.

stuartarchibald: list is supported inside `overload_method`, what's not working?

stuartarchibald: Am I missing something or is the suggestion here to do this:
```python
@overload_method(types.Integer, '__str__')
@overload_method(types.Integer, '__repr__')
def integer_str(n):
    def impl(n):
        s = []
        flag = False
        if n < 0:
            n = -n
            flag = True
        if n == 0:
            return '0'
        while n > 0:
            # there is an implicit conversion of n to float here
            c = chr(int(ord('0') + (n % 10)))
            n = n // 10
            s.append(c)
        if flag:
            s.append('-')
        return ''.join(s)
    return impl
```
or similar?

stuartarchibald: oh it probably needs a reverse in the join, but the essence is there.

guilhermeleobas: You're right. I might have typed something wrong and the code didn't compile.

numba/cpython/unicode.py
```diff
@@ -2347,21 +2347,22 @@ def impl(i):
...
             # there is an implicit conversion of n to float here
             c = chr(int(ord('0') + (n % 10)))
             n = n // 10
-            s = c + s
+            l.append(c)
+        l = l[::-1]
         if flag:
-            s = '-' + s
-        return s
+            '-'.join(l)
```
hameerabbasi, line 2364:
```suggestion
            l.append('-')
```

numba/cpython/builtins.py
AndrewEckart: Just wanted to mention that in the long term (outside the scope of this PR), the global overload for the built-in `str` should be expanded to call `__repr__` if `__str__` is not available (haven't figured out how to do this yet since the `hasattr` and `getattr` built-ins aren't yet implemented.

guilhermeleobas:
CI failure is not related to any changes introduced in this PR. See: https://dev.azure.com/numba/numba/_build/results?buildId=5931&view=logs&j=a506c6d0-8428-5bb5-7d36-c242a5467a36&t=fbf6f884-d589-58cc-173a-ae041ad78914&l=9276

numba/cpython/builtins.py
stuartarchibald: I think we should defer this part of the change set to https://github.com/numba/numba/pull/5866 and for this case just overload `str`, guard against any type other than `Integer`, and implement as already presented herein. Alternative is to just wait until that is done, but I'd rather get this in for 0.51.0.

@AndrewEckart RE how to do the thing you are asking, I'll comment on your PR.

Thanks all.

stuartarchibald:
> CI failure is not related to any changes introduced in this PR. See: https://dev.azure.com/numba/numba/_build/results?buildId=5931&view=logs&j=a506c6d0-8428-5bb5-7d36-c242a5467a36&t=fbf6f884-d589-58cc-173a-ae041ad78914&l=9276

Most likely not, but never say never :) I've recorded it here: https://github.com/numba/numba/issues/5973 please proceed with the view that it's not related to this PR. In the balance of likelihood I think that what's probably happened is that this PR has added some tests at some point that's shifted the CI test slices into an order that hits a bug, this isn't common but does happen some times.

numba/tests/test_builtins.py
stuartarchibald: Is there a minimal working reproducer for what is being discussed here in relation to cast to float64, if so I can take a look? Thanks.

numba/cpython/builtins.py
guilhermeleobas: Ok, I will undo those changes

numba/tests/test_builtins.py
guilhermeleobas: Try this:
```python
from numba import njit, types
import numpy as np

@njit
def foo(n):
    l = []
    while n > 0:
        c = chr(ord('0') + (n % 10))
        n = n // 10
        l.append(c)
    return ''.join(l[::-1])

x = foo(types.ulonglong(123))
print(x)

```

stuartarchibald: Thanks, the issue is that `n = n // 10` is doing an unsigned/signed module which results in a float64. It's following NumPy semantics, e.g.:
```
In [45]: np.uint64(11)//np.int64(10)                                                                            
Out[45]: 1.0

In [46]: type(np.uint64(11)//np.int64(10))                                                                                      
Out[46]: numpy.float64
```

guilhermeleobas:
@stuartarchibald, I've addressed your comments. Thanks for the review.

numba/cpython/unicode.py
```diff
@@ -2342,6 +2342,28 @@ def impl(i):
...
+            flag = False
+            if n < 0:
+                n = -n
+                flag = True
+            if n == 0:
+                return '0'
+            l = []
+            while n > 0:
+                # there is an implicit conversion of n to float here
+                c = chr(int(ord('0') + (n % 10)))
```
stuartarchibald, line 2359: Don't think the cast won't work for large values, e.g. `np.uint64(~0x0)` ?

stuartarchibald: Think this may end up requiring something like:
```python
@overload(str)
def integer_str(n):
    ten = n(10)
    def impl(n):
        flag = False
        if n < 0:
            n = -n
            flag = True
        if n == 0:
            return '0'
        l = []
        while n > 0:
            chr_arg = ord('0') + (n % ten)
            c = chr(chr_arg)
            n = n // ten
            l.append(c)
        if flag:
            l.append('-')
        return ''.join(l[::-1])
    return impl
```

numba/tests/test_builtins.py
```diff
@@ -505,6 +508,50 @@ def test_hex_npm(self):
...
+            types.int64,
+            types.uint,
+            types.uint8,
+            types.uint16,
+            types.uint32,
+            types.uint64,
+            types.uintp,
+            types.ushort,
+            types.ulong,
+            types.ulonglong,
```
stuartarchibald, line 540: These are just aliases for the unsigned ints above?

numba/tests/test_builtins.py
stuartarchibald: https://github.com/numba/numba/pull/5463#discussion_r452976222 should address this.

numba/tests/test_builtins.py
guilhermeleobas: Hum...probably.

numba/cpython/unicode.py
guilhermeleobas: Thanks, I think this last commit solves this problem

DannyWeitekamp:
Hey a while back I did a bit of work of work on a str <-> float conversion implementation. Just going to drop a link in case it is helpful: https://github.com/numba/numba/issues/5723. I gave up on this then because I couldn't quite match the native python speed, but it was close so there might be some nuggets of useful code in there.

numba/tests/test_builtins.py
stuartarchibald: perhaps removing them is a good idea else it's just testing things that are already tested? `{u,}int{8,16,32,64}` would probably cover everything?

numba/tests/test_builtins.py
```diff
@@ -505,6 +508,51 @@ def test_hex_npm(self):
...
+        for typ in typs:
+            cr = compile_isolated(pyfunc, (typ,), flags=flags)
+            cfunc = cr.entry_point
+            for v in args:
+                if v in large_inputs and typ.bitwidth < 32:
+                    continue
+
+                self.assertPreciseEqual(cfunc(typ(v)), pyfunc(typ(v)))
+
+                if typ in [types.int8, types.int16, types.int32, types.int64]:
```
stuartarchibald, line 553:
```suggestion
                if typ.signed:
```

numba/tests/test_builtins.py
```diff
@@ -505,6 +508,51 @@ def test_hex_npm(self):
...
+            types.ulong,
+            types.ulonglong,
+        ]
+
+        for typ in typs:
+            cr = compile_isolated(pyfunc, (typ,), flags=flags)
+            cfunc = cr.entry_point
+            for v in args:
+                if v in large_inputs and typ.bitwidth < 32:
+                    continue
```
stuartarchibald, line 549: Can probably remove this? If it overflows it still ought to work?

stuartarchibald:
> Hey a while back I did a bit of work of work on a str <-> float conversion implementation. Just going to drop a link in case it is helpful: #5723. I gave up on this then because I couldn't quite match the native python speed, but it was close so there might be some nuggets of useful code in there.

@DannyWeitekamp thanks for cross referencing this.

numba/tests/test_builtins.py
guilhermeleobas: Yes, you're right.

stuartarchibald:
Two things:
1. Segfault on CI is not related.
2. Is the `str`ctor documented anywhere as supported?

numba/tests/test_builtins.py
guilhermeleobas: Done!

guilhermeleobas:
> Is the strctor documented anywhere as supported?

In my last commit I added documentation saying that `str(int)` is supported.

stuartarchibald:
> > Is the strctor documented anywhere as supported?
> 
> In my last commit I added documentation saying that `str(int)` is supported.

Can't seem to find that?

stuartarchibald:
Thanks, 0f4daba  looks good

guilhermeleobas:
Thanks everyone for the reviews.

stuartarchibald:
Testing with a merge of current master fixed the segfault, must be test ordering related as suspected.

Issue #2947, opened 2018-05-04
Attribute annotations in jitclass-wrapped classes cause errors


aldanor:
Here's the minimal example:

```python
import numba

@numba.jitclass([('x', nb.int32)])
class Foo:
    x: int  # doesn't do anything; just an annotation so e.g. code completion works in IDEs
```

which fails with

```python
TypeError: class members are not yet supported: __annotations__
```

The `__annotations__` class attribute should probably be just left alone (or transferred to the wrapper).

smcveigh-phunware:
Besides type annotations, anything that inherits from ABC


```python
    import numba
    import pytest
    import numpy as np
    from typing import Callable
    from collections import OrderedDict
    
    spec = OrderedDict()
    for k in list("abc"):
        spec[k] = numba.float64
    
    
    @numba.jitclass(spec)
    class Quadratic:
    
        def __init__(self, a, b, c):
            self.a = a
            self.b = b
            self.c = c
    
        def call(self, x):
            return self.c + x * (self.b + x * self.a)
    
    @pytest.mark.parametrize("x", [
        pytest.param(np.random.uniform(-5, 5), id="value"),
        pytest.param(np.linspace(-5, 5), id="vector")
    ])
    def test_quadratic_not_callable(x, a, b, c):
        a, b, c = tuple(np.random.uniform(-2, 2, 3))
        f = Quadratic(a, b, c)
        y = f.call(x)
    
Has no trouble passing tests but:
    
    @numba.jitclass(spec)
    class CallableQuadratic(Callable):
    
        def __init__(self, a, b, c):
            self.a = a
            self.b = b
            self.c = c
    
        def __call__(self, x):
            return self.c + x * (self.b + x * self.a)
    
    @pytest.mark.parametrize("x", [
        pytest.param(np.random.uniform(-5, 5), id="value", mark=pytest.mark.xfail),
        pytest.param(np.linspace(-5, 5), id="vector", mark=pytest.mark.xfail)
    ])
    def test_quadratic_callable(x, a, b, c):
        a, b, c = tuple(np.random.uniform(-2, 2, 3))
        f = Quadratic(a, b, c)
        y = f(x)
```

can't even get through test setup:

```
collected 0 items / 1 errors

==================================== ERRORS ====================================
_____________ ERROR collecting tests/unit/test_numba_with_class.py _____________
tests/unit/test_numba_with_class.py:29: in <module>
    class CallableQuadratic(Callable):
[...]

PR #5464, opened 2020-03-31
Jitclass drops annotations to avoid error

EPronovost:
Partially addresses https://github.com/numba/numba/issues/2947

Support things like
```
@jitclass([("x", intp)])
class Foo:
     class_type: Any

     def __init__(self, x: int):
         self.x = x
```
So that later code that uses `Foo.class_type` passes mypy.


esc:
@EPronovost thanks for submitting this to the Numba issue tracker. Unfortunately our public CI is currently a bit broken as we wait for #5455 to be merged. My suggestion would be to defer starting the review until after master is green again. Then, this branch can be rebased and the review can be started. As a result of this, I will mark this as "In Progress" for the time being.

EPronovost:
I think this is ready for review now.

esc:
@EPronovost thanks for the ping, I have updated the label accordingly.

PR #5465, opened 2020-03-31
Remove macro expansion and replace uses with FE typing + BE lowering

gmarkall:
As title. WIP PR to test on CI, and for testing the change to the ROC backend (I don't have a device to test). This will need a bit of tidying up and refactoring before I remove the WIP tag.

Also includes the fix to #5408, for testing convenience.

@stuartarchibald are you able to try this on a ROC device please?


gmarkall:
Also, it looks like there's no use for the `ir.Intrinsic` node anymore - https://github.com/numba/numba/blob/master/numba/core/ir.py#L1025 - it seemed to only be used in macro expansion. There is `numba.core.extending._Intrinsic` ( https://github.com/numba/numba/blob/master/numba/core/extending.py#L258 ) but I think it is separate / different.

Any thoughts on the removal of `ir.Intrinsic` along with the macro expansion?

esc:
@gmarkall thanks for submitting this, I have marked it as in-progress for now.

gmarkall:
@esc Thanks - I'm going to split it in two as Siu suggested so that the CUDA changes can go in without needing to be blocked on testing the one change to the ROC target. I'll keep this PR for the removal of all macro expansion code, and make a new one with just CUDA.

gmarkall:
A CUDA-only version of this PR is in #5481, which is ready for review - once that is approved and/or merged, I will amend this PR accordingly.

gmarkall:
Now that #5481 is merged, I've merged master into this branch and updated this PR, and I believe it is now ready for review. It now consists of:

* Removal of the macro expansion pass and associated classes.
* Removal of the `Intrinsic` IR node, which is no longer used. "Intrinsic"  is still a bit of an overloaded term - there's the IR intrinsic nodes, "intrinsic" calls like `getitem` in typing (see e.g. uses of `numba.core.typing.TypeInferer.typeof_intrinsic_call`), and `numba.core.extending._Intrinsic`.
* Some clarifications to comments of CUDA tests, which got missed from #5481.
* Modification of the ROC/HSA target to swap the use of a macro for typing and lowering. This is done blind without testing - if someone with hardware access could test this and let me know if it works, that would be much appreciated! I do have one question about my changes, which I'll add to the diff.

numba/roc/hsaimpl.py
```diff
@@ -234,9 +235,10 @@ def hsail_atomic_add_tuple(context, builder, sig, args):
     return builder.atomic_rmw("add", ptr, val, ordering='monotonic')
 
 
-@lower('hsail.smem.alloc', types.UniTuple, types.Any)
+@lower(roc.shared.array, types.UniTuple, types.Any)
```
gmarkall, line 238: I can't work out why there was never a lowering for `hsail.smem.alloc` with an integer argument. I've swapped like-for-like here, but I can't tell whether this is going to work. Does another lowering:

```python
@lower(roc.shared.array, types.IntegerLiteral, types.Any)
```

need adding, or is there something I've missed that always converts these into `UniTuple`s?

Secondly, if there are two integer literals of different value, I suspect the argument type might be `types.Tuple` - does this case need handling?

sklam: I can't tell why this works neither. We will need to check this backend on real HW.

sklam: And for `Tuple` vs `UniTuple`, you are right:

```
In [3]: types.UniTuple.from_types([types.IntegerLiteral(1), types.IntegerLiteral(2)])
Out[3]: Tuple(Literal[int](1), Literal[int](2))

In [4]: types.UniTuple.from_types([types.IntegerLiteral(1), types.IntegerLiteral(1)])
Out[4]: UniTuple(Literal[int](1) x 2)
```

will need a version for `Tuple`

stuartarchibald: I'll try and get some hardware sorted locally to test this on.

gmarkall:
Per recent discussions (in the dev meeting, IIRC) it's not urgent or likely to get this into 0.51, so I've bumped the milestone to 0.52.

stuartarchibald:
> Per recent discussions (in the dev meeting, IIRC) it's not urgent or likely to get this into 0.51, so I've bumped the milestone to 0.52.

Thanks @gmarkall

numba/roc/hsaimpl.py
```diff
@@ -234,9 +235,20 @@ def hsail_atomic_add_tuple(context, builder, sig, args):
     return builder.atomic_rmw("add", ptr, val, ordering='monotonic')
 
 
-@lower('hsail.smem.alloc', types.UniTuple, types.Any)
-def hsail_smem_alloc_array(context, builder, sig, args):
-    shape, dtype = args
+@lower(roc.shared.array, types.IntegerLiteral, types.Any)
```
stuartarchibald, line 238: This needs `from numba import roc` at the top.

numba/roc/hsaimpl.py
gmarkall: Hopefully this is now resolved.

numba/roc/hsaimpl.py
gmarkall: Added.

numba/cuda/tests/cudapy/test_lang.py
```diff
@@ -40,11 +40,14 @@ def foo(a):
 
     def test_issue_872(self):
         '''
-        Ensure that macro expansion works for more than one block (issue #872)
+        Ensure that typing and lowering of CUDA kernel API primitives works in
+        more than one block. Was originally to nsure that macro expansion works
```
stuartarchibald, line 44:
```suggestion
        more than one block. Was originally to ensure that macro expansion works
```

stuartarchibald:
For 5daa538 the things I'd expect to work ok on ROCm in light of these changes still work, so I think it's ok on that hardware target. The patch itself looks good to me, @sklam did you want to take another look? If not, this can go through the build farm and be merged if it passes. Thanks again for the large refactoring effort, very pleased to see `Macro` removed!

sklam:
It will just need to be tested on real ROC hw i the farm to pick up anything else we missed by eyeballing.

stuartarchibald:
> It will just need to be tested on real ROC hw in the farm to pick up anything else we missed by eyeballing.

I've run this on real ROC hw locally, the things I'd expect to work ok are working.

stuartarchibald (approved):
Thanks for working on this and for all the fixes etc, especially guessing for non-local hardware! Looks good!

esc:
BFID: numba_smoketest_cuda_98

esc:
Buildfarm passed.

PR #5478, opened 2020-04-01
Fix #5471. Issue with omitted type not recognized as literal value.

### LINKED ISSUE #5471: Default value is not literal
## Reporting a bug

```python
import numba

def inner(a):
    pass

@numba.extending.overload(inner)
def inner_overload(a):
    if not isinstance(a, numba.types.Literal):
        return None

    def impl(a):
        print(a)

    return impl

@numba.extending.register_jitable
def my_func(a='a'):
    inner(a)

@numba.njit()
def f():
    my_func()

f()
```

Output:
```
  File "./test.py", line 24, in <module>
    f()
  File ".../numba/core/dispatcher.py", line 401, in _compile_for_args
    error_rewrite(e, 'typing')
  File ".../numba/core/dispatcher.py", line 344, in error_rewrite
    reraise(type(e), e, None)
  File ".../numba/core/utils.py", line 79, in reraise
Invalid use of Function(<function my_func at 0x7ff222b79f70>) with argument(s) of type(s): ()
 * parameterized
In definition 0:
Invalid use of Function(<function inner at 0x7ff23f0ab700>) with argument(s) of type(s): (unicode_type)
 * parameterized
In definition 0:
In definition 1:
```

Version:
```python
@numba.njit()
def f_1():
    my_func('a')
```

Works fine.

Numba version is 0a8b8118987944e82603bfa3c90d87b80ffe5d99

sklam:
Closes #5471.

### DIFF
--- a/numba/core/typing/context.py
+++ b/numba/core/typing/context.py
@@ -385,7 +385,7 @@ class BaseContext(object):
 
         raise typeof_exc
 
-    def resolve_value_type_as_literal(self, value):
+    def resolve_value_type_prefer_literal(self, value):
         """Resolve value type and prefer Literal types whenever possible.
         """
         lit = types.maybe_literal(value)
--- a/numba/core/typing/context.py
+++ b/numba/core/typing/context.py
@@ -385,6 +385,15 @@ class BaseContext(object):
 
         raise typeof_exc
 
+    def resolve_value_type_as_literal(self, value):
+        """Resolve value type and prefer Literal types whenever possible.
+        """
+        lit = types.maybe_literal(value)
+        if lit is None:
+            return self.resolve_value_type(value)
+        else:
+            return lit
+
     def _get_global_type(self, gv):
         ty = self._lookup_global(gv)
         if ty is not None:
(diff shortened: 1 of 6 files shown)


numba/core/typing/context.py
```diff
@@ -385,6 +385,15 @@ def resolve_value_type(self, val):
 
         raise typeof_exc
 
+    def resolve_value_type_as_literal(self, value):
```
stuartarchibald, line 388:
```suggestion
    def resolve_value_type_prefer_literal(self, value):
```
I'm wondering if this ought to be renamed as this, as the type will still resolve if possible, it's just that literals are preferred opposed to required.

numba/tests/test_types.py
```diff
@@ -785,5 +785,60 @@ def f(a):
...
+        # type(val) where val is a literal should not give a literal type.
+        def inner(a):
+            pass
+
+        @overload(inner)
+        def inner_overload(a):
+            if not isinstance(a, types.Literal):
+                return
+            self.assertIsInstance(a, types.Literal)
+            return lambda a: type(a)(a + 1)
```
stuartarchibald, line 825: As this is a type cast, would it be better to do `(a + 1.23)` or similar, to demonstrate the cast has an effect, else the changes to `type` impl could just be a `nop`?

PR #5481, opened 2020-04-02
CUDA: Replace macros with typing and lowering implementations

gmarkall:
This PR contains a few preparations in the main (non-CUDA) part of Numba (see individual commit messages) along with the changes to the CUDA target to implement all operations without using macros, and switching to typing everything "properly" in the fronted and lowering the typed IR as normal. Testing indicates no significant changes to the resulting PTX - there are one or two cases where there isa slightly different ordering of instructions, but no other changes are observed.

All CUDA tests pass on my local systems (Tesla V100 in DGX-1, and a GTX 960).

This required a few related changes:

- Addition of a new type, `dim3`, for variables like `threadIdx`, etc.
- Creation of `numba.cuda.types` and `numba.cuda.models` for keeping types and datamodels for CUDA-specific types. It might seem a little frivolous to add them for a single type in this PR, but more types will be added later, and it seems prudent to start with things organised logically.
- Rename `test_macro` to `test_sm_creation` - the tests in it are still of value - they're no longer testing anything to do with macro expansion, but they do test that various possible user errors correctly fail.
- Moving the lowering of creating constants from `cudaimpl` to the CUDA target context - from the comment that was in `CUDATargetContext.make_constant_array`, it looks like this was on somebody's to-do list already. This makes the actual constructor for constant arrays (`cuda.local.array`) practically a no-op, which just returns its argument.
- Some refactoring of stubs, and the creation of a `stub_function` wrapper. This makes autocomplete and documentation from within IPython (and possibly other IDEs) work a little more smoothly.


gmarkall:
Removed the WIP tag as I believe this is ready to review now - CI is re-running because I wrote the commit message for the last patch while CI was running, and now force-pushed - however, the code is unchanged since the last CI run.

esc:
@gmarkall thanks, I have marked it as RFR.

numba/core/typing/npydecl.py
```diff
@@ -439,7 +439,10 @@ def register_number_classes(register_global):
 # -----------------------------------------------------------------------------
 # Numpy array constructors
 
-def _parse_shape(shape):
+def parse_shape(shape):
```
stuartarchibald, line 442: API change to public, seems safe WRT https://github.com/numba/numba/issues/5214

numba/cuda/cudaimpl.py
```diff
@@ -7,135 +7,117 @@
...
-# Dynamic create all special register
-for sreg in nvvmutils.SREG_MAPPING.keys():
-    lower(sreg)(ptx_sreg_template(sreg))
+@lower(cuda.gridsize, types.int32)
+def cuda_gridsize(context, builder, sig, args):
+    restype = sig.return_type
 
+    ntidx = nvvmutils.call_sreg(builder, "ntid.x")
+    nctaidx = nvvmutils.call_sreg(builder, "nctaid.x")
+    nx = builder.mul(ntidx, nctaidx)
```
stuartarchibald, line 93: can probably make this into a little function and call for x, y, z? not essential

numba/cuda/cudaimpl.py
```diff
@@ -151,36 +133,42 @@ def _get_unique_smem_id(name):
     return "{0}_{1}".format(name, _unique_smem_id)
 
 
-@lower('ptx.smem.alloc', types.intp, types.Any)
-def ptx_smem_alloc_intp(context, builder, sig, args):
-    length, dtype = args
+@lower(cuda.shared.array, types.Integer, types.Any)
```
stuartarchibald, line 136: Does this not require a `LiteralInteger`? (same question throughout where `.literal_value` attr in use).

numba/cuda/stubs.py
```diff
@@ -1,17 +1,19 @@
 """
 This scripts specifies all PTX special objects.
 """
-import operator
-import numpy
+import functools
 import llvmlite.llvmpy.core as lc
+import numpy
```
stuartarchibald, line 6: I know this is just forwarding the original code, but for consistency in the code base this should be `import numpy as np`

numba/cuda/target.py
```diff
@@ -85,6 +85,15 @@ def codegen(self):
     def target_data(self):
         return self._target_data
 
+    @cached_property
+    def nonconst_module_attrs(self):
```
stuartarchibald, line 89: Docstring?

numba/cuda/tests/cudadrv/test_linker.py
```diff
@@ -85,18 +85,27 @@ def foo(x, y):
...
-        self.assertEquals(57, compiled._func.get().attrs.regs)
+        self.assertLessEqual(compiled._func.get().attrs.regs, 57)
 
     @require_context
     def test_set_registers_38(self):
         compiled = cuda.jit(max_registers=38)(function_with_lots_of_registers)
         compiled = compiled.specialize(np.empty(32), *range(6))
-        self.assertEquals(38, compiled._func.get().attrs.regs)
-
+        self.assertLessEqual(compiled._func.get().attrs.regs, 38)
```
stuartarchibald, line 108: think this was perhaps covered a bit by another PR, so there might be a minor conflict.

numba/cuda/types.py
```diff
@@ -0,0 +1,9 @@
+from numba.core import types
+
+
+class Dim3(types.Type):
```
stuartarchibald, line 4: Docstring?

numba/core/base.py
```diff
@@ -335,6 +335,10 @@ def get_data_packer(self, fe_types):
     def target_data(self):
         raise NotImplementedError
 
+    @utils.cached_property
+    def nonconst_module_attrs(self):
```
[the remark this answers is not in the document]
gmarkall: Added.

numba/core/typing/npydecl.py
gmarkall: there is no change suggested here, right?

numba/cuda/cudaimpl.py
gmarkall: done. That function will probably be handy for the implementation of future functions, so might as well have it now.

numba/cuda/cudaimpl.py
gmarkall: I've changed it to use `ImtegerLiteral`. For the variants that expect a `UniTuple` I couldn't see a way to enforce that it is a tuple of literals - is there a way to do that?

numba/cuda/stubs.py
gmarkall: it turned out not to be required at all so I removed it.

numba/cuda/target.py
gmarkall: added

numba/cuda/tests/cudadrv/test_linker.py
gmarkall: a merge made it disappear.

numba/cuda/types.py
gmarkall: added

numba/core/typing/npydecl.py
stuartarchibald: Correct, no change. Was just noting that it was being changed and that a check was made for immediate impact.

numba/cuda/cudaimpl.py
stuartarchibald: > I've changed it to use `ImtegerLiteral`.

Thanks.

> For the variants that expect a `UniTuple` I couldn't see a way to enforce that it is a tuple of literals - is there a way to do that?

I don't think there's a way to "spell" that for `@lower` at present, I think it'd have to accept a `UniTuple` and then scan the contents in the function body via the signature. This should really be picked up at type inference time ahead of here, i.e. `numba.cuda.cudadecl.Cuda_shared_array` ought to do a strong validation first.

gmarkall: I've added some stronger validation to `cudadecl` and a test that using non-literals results in a `TypeError` now.

stuartarchibald:
Thanks for the fixes, code looks good. Will just push it through the farm.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_53`.

stuartarchibald:
Buildfarm passed.

PR #5489, opened 2020-04-02
docs: fix rendering of nested bulleted list

ARF1:
As title.


stuartarchibald:
Thanks for the patch. Manually built the docs with this patch, it sorts out the list so it renders correctly.

PR #5492, opened 2020-04-02
WIP: do not link unreferenced environments

ARF1:


ARF1:
@stuartarchibald This is ready for a first review.

sklam:
@ARF1 was there a reason/use-case for this change? It doesn't seem the memory used by the extra Environment object would be problematic.

PR #5485, opened 2020-04-02
Show the offending module in "no direct replacement" error message

eric-wieser:

PR #5494, opened 2020-04-02
Fix/5484

### LINKED ISSUE #5484: Compatibility shims raise cryptic error
An issue with the 0.49 shims:
```
In [11]: import numba.targets

In [12]: numba.targets.test
<ipython-input-12-88852dcaa72f> in <module>
----> 1 numba.targets.test

c:\users\wiese\repos\ga\numba\numba\core\errors.py in wrapper(attr)
    434     """
    435     def wrapper(attr):
--> 436         mod = import_module(new_module)
    437         ret_attr = getattr(mod, attr)
    438         msg = _moved_msg2.format(attr, old_module, new_module)

~\AppData\Local\Programs\Python\Python38\lib\importlib\__init__.py in import_module(name, package)
    116     """
    117     level = 0
--> 118     if name.startswith('.'):
    119         if not package:
    120             msg = ("the 'package' argument is required to perform a relative "

```

I think you need a check for `new_module is None`, to emit a more useful error message.

stuartarchibald:
Fixes #5484 
Includes PR #5485 as it touches the same code.

### DIFF
--- a/numba/core/errors.py
+++ b/numba/core/errors.py
@@ -411,7 +411,7 @@ _moved_msg2 = ("An import was requested from a module that has moved location"
                "present in Numba version 0.50.0.")
 
 
-_moved_no_replacement = ("No direct replacement available. Visit "
+_moved_no_replacement = ("No direct replacement for '{}' available. Visit "
                          "https://gitter.im/numba/numba-dev to request help. "
                          "Thanks!")
 
@@ -422,7 +422,7 @@ def deprecate_moved_module(old_module, new_module, stacklevel=3):
     from `numba.XYZ` shims.
     """
     if new_module is None:
-        msg = _moved_no_replacement
+        msg = _moved_no_replacement.format(old_module)
     else:
         msg = _moved_msg1.format(old_module, new_module)
     warnings.warn(msg, category=NumbaDeprecationWarning, stacklevel=stacklevel)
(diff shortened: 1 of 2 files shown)


numba/core/errors.py
```diff
@@ -433,11 +433,17 @@ def deprecate_moved_module_getattr(old_module, new_module):
     locations
     """
     def wrapper(attr):
-        mod = import_module(new_module)
-        ret_attr = getattr(mod, attr)
-        msg = _moved_msg2.format(attr, old_module, new_module)
-        warnings.warn(msg, category=NumbaDeprecationWarning, stacklevel=2)
-        return ret_attr
+        if new_module is None:
+            msg = _moved_no_replacement
```
eric-wieser, line 437: Is it worth attempting to import `attr` here first?

eric-wieser: I'm pointing out that `import numba.target; x = numba.target.imputils` fails in .49 but works in 0.48.

eric-wieser: But could be made to work if, before giving up and raising, you attempt to import the submodule with the provided name.

eric-wieser: I've expanded on that in https://github.com/numba/numba/pull/5497

stuartarchibald:
Closing as https://github.com/numba/numba/pull/5498 does a better job!

Issue #5466, opened 2020-03-31
0.49 RC: AttributeError: module 'numba' has no attribute 'numpy_support'


hameerabbasi:
The commit https://github.com/pydata/sparse/commit/b642e115d5a66db209abf812637fad752c13ed19 of PyData/Sparse passes in Numba 0.48.

In 0.49, it fails with the message in the title.

hameerabbasi:
Further digging reveals that `numpy_support` is available, but only once it's imported:

```python
>>> import numba
>>> numba.numpy_support
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'numba' has no attribute 'numpy_support'
>>> import numba.numpy_support
<stdin>:1: NumbaDeprecationWarning: An import was requested from a module that has moved location.
Import requested from: 'numba.numpy_support', please update to use 'numba.np.numpy_support' or pin to Numba version 0.48.0. This alias will not be present in Numba version 0.50.0.
>>> numba.numpy_support
<module 'numba.numpy_support' from '/Users/hameerabbasi/miniforge3/envs/sparse/lib/python3.8/site-packages/numba/numpy_support.py'>
```

stuartarchibald:
Thanks for the report. I think its existence may well have been a side effect of import chains.

Prior to 0.49: `numba.__init__` imported `numba.special` imported `numba.typing` imported `numba.typing.context` imported `numba.typing.typeof` imported `numba.numpy_support`.

The 0.48 release doesn't have it in `__init__` https://github.com/numba/numba/blob/release0.48/numba/__init__.py

thoughts ?

hameerabbasi:
Well, there's code out there I'd like to not break on day 0, so it'd be nice if it was added to `__init__.py` for 0.49 and dropped in 0.50 if necessary.

stuartarchibald:
If I can find a reasonable way to ensure that a warning notice about the deprecation is emitted, and only for the case above, then I think that this can be patched in for 0.49.

PR #5497, opened 2020-04-03
Made `numba.numpy_support` available without an import

eric-wieser:
Closes https://github.com/numba/numba/issues/5466

### DIFF
--- a/numba/__init__.py
+++ b/numba/__init__.py
@@ -58,6 +58,31 @@ import numba.cpython.charseq
 test = runtests.main
 
 
+# needed for compatibility until 0.50.0. Note that accessing members of
+# numpy_support will raise DeprecationWarning
+_auto_import_submodules = {
+    'numba.numpy_support': 'numba.core.numpy_support'
+}
+if sys.version_info < (3, 7):
+    import importlib
+    for _old_mod in _auto_import_submodules.keys():
+        importlib.import_module(_old_mod)
+else:
+    def __getattr__(attr):
+        submodule_name = 'numba.{}'.format(attr)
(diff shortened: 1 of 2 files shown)


numba/__init__.py
```diff
@@ -58,6 +58,31 @@
...
+            pass
+        else:
+            errors.deprecate_moved_module(submodule_name, new_name)
+            return importlib.import_module(submodule_name)
+
+        # produce the normal error message
+        from types import ModuleType
+        ModuleType.__getattribute__(sys.modules[__name__], attr)
+
+
```
stuartarchibald, line 85: Should this be rolled into https://github.com/numba/numba/pull/5498 ?

eric-wieser: I'll rebase it onto #5498 once that gets merged.

stuartarchibald: Sounds good, thanks.

numba/__init__.py
```diff
@@ -58,6 +58,31 @@
 test = runtests.main
 
 
+# needed for compatibility until 0.50.0. Note that accessing members of
+# numpy_support will raise DeprecationWarning
+_auto_import_submodules = {
+    'numba.numpy_support': 'numba.core.numpy_support'
```
eric-wieser, line 64:
```suggestion
    'numba.numpy_support': 'numba.core.numpy_support'
```
```suggestion
    'numba.targets': None
```

sklam:
I have extended this PR to cover all other `numba.<mod>` in https://github.com/numba/numba/pull/5532

PR #5498, opened 2020-04-03
 Make the shim deprecation warnings work on python 3.6 too

eric-wieser:

### DIFF
--- a/numba/core/errors.py
+++ b/numba/core/errors.py
@@ -15,6 +15,7 @@ from numba.core.utils import add_metaclass, reraise, chain_exception
 from functools import wraps
 from abc import abstractmethod
 from importlib import import_module
+from types import ModuleType
 
 # Filled at the end
 __all__ = []
@@ -416,7 +417,7 @@ _moved_no_replacement = ("No direct replacement for '{}' available. Visit "
                          "Thanks!")
 
 
-def deprecate_moved_module(old_module, new_module, stacklevel=3):
+def deprecate_moved_module(old_module, new_module, stacklevel=2):
     """Warn about a module level location move of some part of Numba's
     internals. stacklevel is 3 by default as most warning locations are
     from `numba.XYZ` shims.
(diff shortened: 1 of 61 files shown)


stuartarchibald:
Thanks for the patch. Think something might be missing, `numba.core.errors._MovedModule ` isn't defined?!

eric-wieser:
Crap, working with large find/replace results is hard

eric-wieser:
Will fix flake8 after any other review comments, seems pointless running CI an additional time

gmarkall:
Thanks! Marking as ready for review.

stuartarchibald:
@sklam this patch makes the "shim" work for all Python 3s. It uses a technique described here https://mail.python.org/pipermail/python-ideas/2012-May/014969.html by Guido and it "... works because the import machinery is actively enabling this hack". Whilst it's messing with `sys.modules` at import time, it appears to be an approved way to do so! Any concerns?

numba/core/errors.py
```diff
@@ -425,20 +426,29 @@ def deprecate_moved_module(old_module, new_module, stacklevel=3):
...
+
+    def __getattr__(self, attr):
+        """ warn users above modules moving locations """
+        mod = import_module(self.__new_module)
         ret_attr = getattr(mod, attr)
-        msg = _moved_msg2.format(attr, old_module, new_module)
+        msg = _moved_msg2.format(attr, self.__name__, self.__new_module)
         warnings.warn(msg, category=NumbaDeprecationWarning, stacklevel=2)
         return ret_attr
-    return wrapper
```
eric-wieser, line 452: If it's useful, I could add

```suggestion
    def __repr__(self):
        return ModuleType.__repr__(self).replace('>', ', redirecting to {}>'.format(self.__new_module))
```
Didn't seem particularly worthwhile.

eric-wieser:
Will rebase and fix flake8 today, in the hope that silence means there are no other concerns

stuartarchibald:
> Will rebase and fix flake8 today, in the hope that silence means there are no other concerns

Thanks @eric-wieser, we discussed this yesterday during the Numba weekly triage meeting, the view in general was this method has been blessed by CPython devs and so should be fine (tests also seem to confirm that to be the case)!

I did note that WRT your comments here: https://github.com/numba/numba/pull/5494#discussion_r402581760

This patch does:
```
$ python -c 'import numba.targets; x = numba.targets.imputils'
<string>:1: NumbaDeprecationWarning: No direct replacement available. Visit https://gitter.im/numba/numba-dev to request help. Thanks!
  File "<string>", line 1, in <module>
  File "<path>/numba/numba/core/errors.py", line 447, in __getattr__
    mod = import_module(self.__new_module)
  File "<env>/lib/python3.8/importlib/__init__.py", line 118, in import_module
    if name.startswith('.'):
```
which does issue the warning but also maybe has a similar problem RE None?

eric-wieser:
Fixed the `None` issue, and also added a second commit to attempt to fix the `numba.targets.imputils` issue.

eric-wieser:
Think I'm finally through the flake8 errors

numba/core/errors.py
```diff
@@ -425,20 +426,52 @@ def deprecate_moved_module(old_module, new_module, stacklevel=3):
...
+                raise
+
+            # perhaps this is a submodule name that was previous importer, but
+            # is no longer
+            try:
+                return import_module("." + attr, package=self.__name__)
+            except ModuleNotFoundError:
+                raise AttributeError(
+                    "Moved module {!r} has no attribute or submodule {!r}"
+                    .format(self.__name__, attr))
```
stuartarchibald, line 474: No test appears to hit this branch? I think something like this would exercise it?:
```
python -c "import numba.targets; numba.targets.does_not_exist"
```

numba/testing/main.py
```diff
@@ -579,7 +579,6 @@ def add_results(self, result):
         self.expectedFailures.extend(result.expectedFailures)
         self.unexpectedSuccesses.extend(result.unexpectedSuccesses)
 
-
```
stuartarchibald, line 582: This is deleted line should be put back!?

numba/core/errors.py
eric-wieser: This branch is hit only sometimes, depending on which order the tests run in.

Starting a new interpreter would be the best way to ensure the test is hit, yes. Where would such a test go?

numba/testing/main.py
eric-wieser: Good catch

numba/core/errors.py
stuartarchibald: Yes, and also perhaps capture the output and ensure that it contains the expected? This shim is intended to be short-lived in the code base, so don't worry about code coverage. I think it'd be fine to add the test to the `numba/tests/test_refactor_moves.py` given it's testing part of the same functionality. Thanks for doing this.

eric-wieser: Apologies, I didn't read this message properly, adding a test is straightforward

stuartarchibald:
Going to run this through the buildfarm just to make sure.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cpu_30`.

stuartarchibald:
Build farm passed apart from w32 which got stuck on one build due to issues with git-bash. Not worth holding up for as there's other w32 issues for 0.49.0rc2.

PR #5510, opened 2020-04-05
[WIP] Creation of record array from list of lists/tuples

luk-f-a:
Adds support for creating a record array from a list of tuples

```python
a_b_type = np.dtype([('a', 'i8'), ('b', 'f8')], align=True)

@njit
def create():
    new = np.array([(1., 2.), (3., 4.)], a_b_type)
    return new
```

### DIFF
--- a/numba/core/typing/npydecl.py
+++ b/numba/core/typing/npydecl.py
@@ -495,20 +495,23 @@ class NpArray(CallableTemplate):
     def generic(self):
         def typer(object, dtype=None):
             ndim, seq_dtype = _parse_nested_sequence(self.context, object)
-            if isinstance(dtype.dtype, types.Record):
-                # a nested sequence should be interpreted as ndim=1 array,
-                # with the other dimension being part of the Record
-                if ndim == 2:
-                    ndim = 1
-                else:
-                    raise ValueError("Only two levels of nested sequences "
-                                     "allowed for Record arrays")
             if dtype is None:
                 dtype = seq_dtype
             else:
                 dtype = _parse_dtype(dtype)
                 if dtype is None:
                     return
+            if isinstance(dtype, types.Record):
+                # a nested sequence should be interpreted as ndim=1 array,
+                # with the other dimension being part of the Record
+                if ndim == 1:
+                    raise ValueError("Creation of Record array requires "
+                                     "a nested sequence")
+                elif ndim == 2:
+                    ndim = 1
+                else:
+                    raise ValueError("Only two levels of nested sequences "
+                                     "allowed for Record arrays")
             return types.Array(dtype, ndim, 'C')
 
         return typer
--- a/numba/core/typing/npydecl.py
+++ b/numba/core/typing/npydecl.py
@@ -495,6 +495,14 @@ class NpArray(CallableTemplate):
     def generic(self):
         def typer(object, dtype=None):
             ndim, seq_dtype = _parse_nested_sequence(self.context, object)
+            if isinstance(dtype.dtype, types.Record):
+                # a nested sequence should be interpreted as ndim=1 array,
+                # with the other dimension being part of the Record
+                if ndim == 2:
+                    ndim = 1
+                else:
+                    raise ValueError("Only two levels of nested sequences "
+                                     "allowed for Record arrays")
             if dtype is None:
                 dtype = seq_dtype
             else:
--- a/numba/tests/test_recarray_usecases.py
+++ b/numba/tests/test_recarray_usecases.py
@@ -8,6 +8,7 @@ from numba.core.compiler import compile_isolated
 from numba.tests.support import captured_stdout, tag, TestCase
 import unittest
 from numba.np import numpy_support
+from numba.core.errors import TypingError
 
 
 def usecase1(arr1, arr2):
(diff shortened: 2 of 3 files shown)


gmarkall:
Thanks for the PR - marking as "in progress" whilst the tests are red - I've had a quick skim through the failures and I'm assuming it's due to the code and not a testing error, but let me know if you need a re-run.

numba/core/typing/npydecl.py
```diff
@@ -495,6 +495,14 @@ class NpArray(CallableTemplate):
     def generic(self):
         def typer(object, dtype=None):
             ndim, seq_dtype = _parse_nested_sequence(self.context, object)
+            if isinstance(dtype.dtype, types.Record):
+                # a nested sequence should be interpreted as ndim=1 array,
+                # with the other dimension being part of the Record
+                if ndim == 2:
+                    ndim = 1
```
eric-wieser, line 502: Only if the nested sequence is of type tuple, I think.

numba/tests/test_recarray_usecases.py
```diff
@@ -144,5 +145,18 @@ def test_usecase5(self):
         self._test_usecase2to5(usecase5, self.aligned_dtype)
 
 
+class TestRecArrayCreation(unittest.TestCase):
+    def test_creation_from_tuple_list(self):
+        a_b_type = np.dtype([('a', 'i8'), ('b', 'f8')], align=True)
+        x = np.array([(1, 2.), (3, 4.)], a_b_type)
```
eric-wieser, line 151: Can you also add a test for
```
        a_b_type = np.dtype([('a', 'i8'), ('b', [('c', 'f8')]), align=True)
        x = np.array([(1, (2.,)), (3, (4.,))], a_b_type)
```

numba/core/typing/npydecl.py
luk-f-a: a list of lists is allowed, while a tuple of lists is not. This is line with Numpy behaviour. I've added tests to cover both combinations that should work: list of tuples and list of lists.
I also added a test to cover two wrong inputs: flat list and 3-nested list.

numba/tests/test_recarray_usecases.py
luk-f-a: sorry, I don't know what dtype that is. There seems to be a mismatch of square and round brackets, but even after fixing that, I don't think that's a valid dtype.

luk-f-a:
@eric-wieser , thanks for the review. It's only my second time extending compiler passes, so all feedback is very welcome.
I addressed your comments, and added new tests.

numba/tests/test_recarray_usecases.py
eric-wieser: Added the missing brackets and parens, whoops

numba/core/typing/npydecl.py
eric-wieser: I was under the impression that numpy only allowed a list of tuples.

numba/tests/test_recarray_usecases.py
```diff
@@ -144,5 +146,48 @@ def test_usecase5(self):
...
+        def create():
+            new = np.array([(1., 2.), (3., 4.)], a_b_type)
+            return new
+
+        with self.subTest("creation from list of tuple"):
+            self.assertTrue(np.array_equal(create(), x))
+
+        @njit
+        def create():
+            new = np.array([[1., 2.], [3., 4.]], a_b_type)
```
eric-wieser, line 164: I don't think this is legal in numpy, or at least does not mean the same thing.

luk-f-a: you're right! I didn't expect that difference. I don't think numpy's behaviour in the list-of-list case is worth supporting. It looks like a bad side-effect, not something useful. I'll restrict the PR to list of tuples.

numba/core/typing/npydecl.py
```diff
@@ -501,6 +501,17 @@ def typer(object, dtype=None):
                 dtype = _parse_dtype(dtype)
                 if dtype is None:
                     return
+            if isinstance(dtype, types.Record):
+                # a nested sequence should be interpreted as ndim=1 array,
+                # with the other dimension being part of the Record
+                if ndim == 1:
```
eric-wieser, line 507: I recommend you add a `stop_at_tuple=True` argument to `_parse_nested_sequence`, rather than forbidding 2d record arrays

numba/tests/test_recarray_usecases.py
```diff
@@ -144,5 +146,48 @@ def test_usecase5(self):
...
+            return new
+
+        with self.subTest("creation from nested list"):
+            self.assertTrue(np.array_equal(create(), x))
```
eric-wieser, line 168: These tests might be better as:
```suggestion
            self.assertTrue(np.array_equal(create(), jit(create)()))
```
after removing the `@jit` decorator above.

luk-f-a:
@eric-wieser , I see your point about nested record arrays (above and in gitter). It would be nice to cover them. However, that would require a quite some change, because I would have to build a recursive function to deal with repeatedly nested fields, plus some way to deal with fields which are arrays.
I realized half-way through that this constructor is not so necessary because `np.empty` works and the user can fill the structure manually. 
I think I'll park this PR for now, close the other more important ones that I have open and then return to this one and try to cover all cases.

PR #5512, opened 2020-04-06
Fix #5502

### LINKED ISSUE #5502: ZeroDivisionError if an empty array is passed to np.mean()
## Reporting a bug

The following code raises ZeroDivisionError in nopython mode.

```python
from numba import njit
import numpy as np

@njit
def mean(arr):
    return np.mean(arr)

mean(np.asarray([]))
```
Without njit decorator, the function returns NaN

### Traceback

ZeroDivisionError                         Traceback (most recent call last)
<ipython-input-11-60946af0372d> in <module>
----> 1 mean(np.asarray([]))

ZeroDivisionError: division by zero

santhisenan:
Fixes the issue where ZeroDivisionError will get thrown while trying to compute the mean of an empty array.

### DIFF
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -402,9 +402,18 @@ def array_cumprod(context, builder, sig, args):
 @lower_builtin(np.mean, types.Array)
 @lower_builtin("array.mean", types.Array)
 def array_mean(context, builder, sig, args):
+    ty = sig.args[0].dtype
     zero = sig.return_type(0)
 
+    mean_empty_array = np.nan
+    if isinstance(ty, (types.NPDatetime, types.NPTimedelta)):
+        mean_empty_array = ty('NaT')
+
     def array_mean_impl(arr):
+        # Handle empty arrays
+        if arr.size == 0:
+            return mean_empty_array
+
         # Can't use the naive `arr.sum() / arr.size`, as it would return
         # a wrong result on integer sum overflow.
         c = zero
--- a/numba/tests/test_array_reductions.py
+++ b/numba/tests/test_array_reductions.py
@@ -212,6 +212,9 @@ class TestArrayReductions(MemoryLeakMixin, TestCase):
         # Only NaNs
         arr = np.float64(['nan', 'nan'])
         check(arr)
+        # Empty array
+        arr = np.float64([])
+        check(arr)
 
     def test_all_basic(self, pyfunc=array_all):
         cfunc = jit(nopython=True)(pyfunc)
--- a/numba/tests/test_array_reductions.py
+++ b/numba/tests/test_array_reductions.py
@@ -212,9 +212,10 @@ class TestArrayReductions(MemoryLeakMixin, TestCase):
         # Only NaNs
         arr = np.float64(['nan', 'nan'])
         check(arr)
-        # Empty array
-        arr = np.float64([])
-        check(arr)
+
+        if pyfunc == array_mean:
+            arr = np.float64([])
+            check(arr)
 
     def test_all_basic(self, pyfunc=array_all):
         cfunc = jit(nopython=True)(pyfunc)


esc:
@santhisenan thanks for opening this. I think it makes sense to add a test-case to ensure the patch does indeed address the issue at hand.

santhisenan:
@esc Please see if [this](https://github.com/numba/numba/blob/917f582433170893f32ae51e5ca8fd8a8a736914/numba/tests/test_array_reductions.py#L215) is where  I should add the test.

Please see the diff below.
```
diff --git a/numba/tests/test_array_reductions.py b/numba/tests/test_array_reductions.py
index 05822eaaf..1a5abba69 100644
--- a/numba/tests/test_array_reductions.py
+++ b/numba/tests/test_array_reductions.py
@@ -212,6 +212,10 @@ class TestArrayReductions(MemoryLeakMixin, TestCase):
         # Only NaNs
         arr = np.float64(['nan', 'nan'])
         check(arr)
+        # Empty array
+        arr = np.float64([])
+        check(arr)
+

     def test_all_basic(self, pyfunc=array_all):
```

This leads to failures and errors in a number os test cases.

esc:
@santhisenan yeah, it is probably worth making that change so we can view the test case failures using our CI systems to decide on how to proceed.

esc:
@santhisenan another suggestion might be to use `np.divide()` instead of using regular division (`/`). Could you give that a try and see if it helps. It might just return `nan` as it respects the Numpy error model.

santhisenan:
@esc When using np.divide, test_mean_npdatetime test throws this error:

`No conversion from timedelta64[] to timedelta64[s] for '$46return_value.6', defined at None
`

ERROR: test_mean_npdatetime (numba.tests.test_array_reductions.TestArrayReductions)
  File "/Users/santhisenan/Documents/projects/numba/numba/tests/test_array_reductions.py", line 709, in test_mean_npdatetime
    self.check_nptimedelta(array_mean)
  File "/Users/santhisenan/Documents/projects/numba/numba/tests/test_array_reductions.py", line 681, in check_nptimedelta
    self._do_check_nptimedelta(pyfunc, arr)
  File "/Users/santhisenan/Documents/projects/numba/numba/tests/test_array_reductions.py", line 660, in _do_check_nptimedelta
    self.assertPreciseEqual(cfunc(arr), pyfunc(arr))
  File "/Users/santhisenan/Documents/projects/numba/numba/core/dispatcher.py", line 401, in _compile_for_args
    error_rewrite(e, 'typing')
  File "/Users/santhisenan/Documents/projects/numba/numba/core/dispatcher.py", line 344, in error_rewrite
    reraise(type(e), e, None)
  File "/Users/santhisenan/Documents/projects/numba/numba/core/utils.py", line 79, in reraise
numba.core.errors.TypingError: Failed in nopython mode pipeline (step: nopython mode backend)
No conversion from timedelta64[] to timedelta64[s] for '$46return_value.6', defined at None

File "numba/np/arraymath.py", line 422:
    def array_mean_impl(arr):
        <source elided>
            c += v.item()
        return np.divide(c, arr.size)
        ^

[1] During: typing of assignment at /Users/santhisenan/Documents/projects/numba/numba/np/arraymath.py (422)

File "numba/np/arraymath.py", line 422:
    def array_mean_impl(arr):
        <source elided>
            c += v.item()
        return np.divide(c, arr.size)
```

santhisenan:
@esc When I add an empty array to list of test cases, 6 test cases fail. The failing test cases can be classified as under:
* Test cases that raise a ValueError: argmax_basic, argmin_basic, max_basic, min_basic, nanmax_basic, nanmin_basic. These test cases map the numpy API, as numpy functions also raise the ValueError.
* Test cases that don't replicate the numpy API: std_basic, var_basic. The problematic numba.mean is called within these functions.

Maybe [this](https://github.com/numba/numba/blob/1ea770564cb3c0c6cb9d8ab92e7faf23cd4c4c19/numba/tests/test_array_reductions.py#L215) is not where I add an empty list to test for the current scenario, as some methods raise an error and those cases are tested separately. I am not able to figure out where to add the test for the empty array mean case. which don't raise an error without disturbing other test cases.

esc:
I am not sure what to recommend in this case. Perhaps the issue with using `np.divide` with the time-deltas needs to be fixed? Maybe all the functions that fail to work correctly on an empty array must be fixed? @santhisenan thanks for all of your work on this so far. I think you may have hit a can of worms (https://en.wiktionary.org/wiki/can_of_worms). Perhaps it will make sense to discuss this issue during the core-developer meeting next week, I'll keep thinking about this and  let you know once we come up with a reasonable course of action.

esc:
@santhisenan we spoke about this during the core developer meeting. It turns out that a new idea has surfaced to allow easier implementation using `overload`: https://github.com/numba/numba/issues/5552

The consensus is that we will probably wait for the above feature to be completed and then (I, you, someone else) rewrite `mean` and the other failing Numpy functions, which should make it easier to fix these kinds of failures systematically.

I have set the status of this issue to 'Blocked awaiting long term feature' for now.

Issue #5503, opened 2020-04-03
Code generation for np.einsum (including path optimization)


max9111:
## Aim of this Function

Numba is very good in compiling loops, but doesn't analyze (at least that I know) array sizes at compile time, even if this could be possible (JIT-compilation). I guess it will be quite hard to implement an overload, which optimizes the reduction path as np.einsum does if optimization is turned on.
This implementation isn't an overload, but simply translates a einsum function to Numba compile-able code.

## Implementation

```python
import numpy as np
import numba as nb

indend= " " *4

def parse_einsum_str(einsum_str):
    lhs,rhs=einsum_str.split("->")
    lhs=lhs.split(",")
    
    charset=set()
    for lhs_ in lhs:
        for s in lhs_:
            charset.add(s)
    return charset, lhs,rhs

def write_header(s,arr_names):
    s+="def einsum("
    for i in range(len(arr_names)-1):
        s+=arr_names[i]+","
    s+=arr_names[-1]+"):\n"
    return s

def write_size_assert(s,ind_level,sizes,asserts):
    ind_level+=1
    for i in range(len(sizes)):
        s+= indend *ind_level + sizes[i]+"\n"
    
    s+= indend *ind_level +"\n"
    for i in range(len(asserts)):
        s+= indend *ind_level + asserts[i]+"\n"
    
    s+= " " * ind_level*4 +"\n"
    return s,ind_level

def get_size(char_list_lhs,arr_names,arr):
    sizes=[]
    asserts=[]
    dict_sizes={}
    
    charset=set()
    for i ,chars in enumerate(char_list_lhs):
        for j,char in enumerate(chars):
            if char not in charset:
                sizes.append(char+"_s = " + arr_names[i] + ".shape["+str(j)+"]")
                dict_sizes[char] = arr[i].shape[j]
                charset.add(char)
            else:
                asserts.append("assert " + char+"_s == " + arr_names[i] + ".shape["+str(j)+"]")
    
    return sizes, asserts ,dict_sizes

def get_loop_order(lhs,rhs,charset,dict_sizes,idx_rm):
    cost = dict.fromkeys(charset, 0)
    
    for i ,lhs_ in enumerate(lhs):
        #C-order assumed
        stride=1
        for j in range(len(lhs_)-2,-1,-1):
            stride*=dict_sizes[lhs_[j+1]]
            cost[lhs_[j]]+=stride
    
    if len(rhs)>0:
        stride=1
[...]

PR #5514, opened 2020-04-06
Code generation for np.einsum

max9111:
## Aim of this Function

Numba is very good in compiling loops, but doesn't analyze (at least that I know) array sizes at compile time, even if this could be possible (JIT-compilation). I guess it will be quite hard to implement an overload, which optimizes the reduction path as np.einsum does if optimization is turned on.
This implementation isn't an overload, but simply translates a einsum function to Numba compile-able code.

[https://github.com/numba/numba/issues/5503#issue-593540730](url)

### DIFF
--- a/contrib/einsum_gen.py
+++ b/contrib/einsum_gen.py
@@ -1,0 +1,384 @@
+import numpy as np
+import numba as nb
+
+indend= " " *4
+
+def parse_einsum_str(einsum_str):
+    lhs,rhs=einsum_str.split("->")
+    lhs=lhs.split(",")
+    
+    charset=set()
+    for lhs_ in lhs:
+        for s in lhs_:
+            charset.add(s)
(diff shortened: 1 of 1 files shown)


gmarkall:

sklam:
A few of the core developers have concern that this patch would add significant maintenance burden due to the complexity of the implementation. We will need more time to understand and get comfortable with the code. Here's a some questions and tasks:

- Compare against numpy's implementation https://github.com/numpy/numpy/blob/v1.17.0/numpy/core/einsumfunc.py#L1004-L1432
- Understand it's performance characteristics
- Is there a simpler alternative that doesn't require python-source generation? The heavy reliance on source-generation is usual in numba.

stuartarchibald:
xref: https://numba.discourse.group/t/addons-contrib-repo/29 as a potential location?

PR #5526, opened 2020-04-08
Impl. np.asarray(literal)

guilhermeleobas:
As the title suggests. This pull request implements `np.asarray(Literal)`.


gmarkall:

numba/np/arraymath.py
```diff
@@ -3895,6 +3895,12 @@ def impl(a, dtype=None):
             for i, v in enumerate(a):
                 ret[i] = v
             return ret
+    elif isinstance(a, types.Literal):
```
stuartarchibald, line 3898: I think it would be a good idea to be specific here, this is for instances of `types.StringLiteral`s only?

numba/np/numpy_support.py
```diff
@@ -600,7 +608,7 @@ def type_can_asarray(arr):
     implementation, False otherwise.
     """
 
-    ok = (types.Array, types.Sequence, types.Tuple,
+    ok = (types.Array, types.Sequence, types.Tuple, types.Literal,
```
stuartarchibald, line 611: Same comment as previously RE being specific about `StringLiteral`.

numba/np/arraymath.py
```diff
@@ -3895,6 +3895,12 @@ def impl(a, dtype=None):
...
+        val = a.literal_value
+        dt = 'U' + str(len(val))
```
stuartarchibald, line 3900: Is this correct for empty strings? They'd have length 0.

```python
from numba import njit
import numpy as np

@njit
def foo():
    return np.asarray("")

z=foo()
print(z)
print(z.dtype)

z=foo.py_func()
print(z)
print(z.dtype)

```
gives:
```

<U0

<U1
```

numba/tests/test_np_functions.py
```diff
@@ -3106,6 +3106,23 @@ def check_pass_through(jitted, expect_same, params):
...
+            s = "hello world"
+            return np.asarray(s)
+
+        funcs = [case1, case2]
+
+        for pyfunc in funcs:
+            cfunc = jit(nopython=True)(pyfunc)
+            expected = pyfunc()
+            got = cfunc()
+            self.assertEqual(expected, got)
```
stuartarchibald, line 3124: I think this should `assertPreciseEqual` for a more rigorous check.

numba/cpython/charseq.py
```diff
@@ -249,6 +249,36 @@ def codegen(context, builder, signature, args):
...
+    dstint_t = ir.IntType(8 * unicode_byte_width)
+    dst_ptr = cgutils.alloca_once(builder, lty, size=length)
+    dst = builder.bitcast(dst_ptr, ir.IntType(32).as_pointer())
+
+    with cgutils.for_range(builder, size) as loop:
+        in_ptr = builder.gep(src, [loop.index])
+        in_val = builder.zext(builder.load(in_ptr), dstint_t)
+        builder.store(in_val, builder.gep(dst, [loop.index]))
+
+    return builder.load(dst_ptr)
```
stuartarchibald, line 279: NOTE: I've not checked this yet, more testing needed and algorithm needs checking ahead of validating this section.

numba/tests/test_np_functions.py
```diff
@@ -3106,6 +3106,23 @@ def check_pass_through(jitted, expect_same, params):
...
                     check_pass_through(cfunc, True, params)
 
+    def test_asarray_literal(self):
+
+        def case1():
+            return np.asarray("hello world")
+
+        def case2():
+            s = "hello world"
+            return np.asarray(s)
```
stuartarchibald, line 3116: Perhaps try a load of different strings and unicode encodings?

numba/cpython/charseq.py
guilhermeleobas: Had to change the code a bit to accommodate other kinds of unicodes

guilhermeleobas:
It seems that `np.asarray(大处着眼，小处着手。🐍⚡)` is being interpreted as having dtype `<U13` on Windows rather than `<U12`.

seibert:
This is being caused by the snake emoji, which is the only 4-byte character in that string.  NumPy has a UCS-4 item size for unicode (as it has to in order to have fixed-size encoding), but on Windows it computes the dtype length assuming UTF-16 (?) encoding.  Looks like an obscure bug in NumPy (it isn't wrong, just slightly memory inefficient), so it would be fair to relax the test strictness.

guilhermeleobas:
This PR is ready for a round of reviews

guilhermeleobas:
cc @hameerabbasi

hameerabbasi:
I don't think I'm qualified to review this one. ☹️

numba/cpython/charseq.py
```diff
@@ -249,6 +249,59 @@ def codegen(context, builder, signature, args):
...
+            in_ptr = builder.gep(src2, [loop.index])
+            in_val = builder.zext(builder.load(in_ptr), dstint_t)
+            builder.store(in_val, builder.gep(dst, [loop.index]))
+
+    with builder.if_then(kind4):
+        with cgutils.for_range(builder, size) as loop:
+            in_ptr = builder.gep(src4, [loop.index])
+            in_val = builder.zext(builder.load(in_ptr), dstint_t)
+            builder.store(in_val, builder.gep(dst, [loop.index]))
+
```
stuartarchibald, line 301: What do you think of perhaps folding all this into something like:
```python
    bits = 8
    i8 = ir.IntType(bits)
    i32 = ir.IntType(32)

    # sizes
    lty = context.get_value_type(toty)
    size = i8(len(literal_value))

    # returned string
    dstint_t = ir.IntType(8 * unicode_byte_width)
    dst_ptr = cgutils.alloca_once(builder, lty, size=length)
    dst = builder.bitcast(dst_ptr, i32.as_pointer())

    # const
    gv = context.insert_const_bytes(mod, databytes)

    for p in range(3):
        bw = 1 << p
        ibw = ir.IntType(bits * bw)
        src = builder.bitcast(gv, ibw.as_pointer())
        switching_kind = builder.icmp_unsigned('==', i8(kind), i8(bw))

        with builder.if_then(switching_kind):
            with cgutils.for_range(builder, size) as loop:
                in_ptr = builder.gep(src, [loop.index])
                in_val = builder.zext(builder.load(in_ptr), dstint_t)
                builder.store(in_val, builder.gep(dst, [loop.index]))
```
?

stuartarchibald: Also wonder if there's benefit in just doing:
```python
@overload(np.asarray)
def ol_asarray(string):
    if not isinstance(string, types.StringLiteral):
        return None
    arr = np.array(string.literal_value)
    def impl(string):
        return arr.copy()
    return impl
```
or similar?

guilhermeleobas: That's way better. Less code to maintain. I will replace my current implementation by this one.

guilhermeleobas:
Thanks for the review @stuartarchibald. I've replaced the current implementation by a simpler one as you suggested.

guilhermeleobas:
Thanks for the review @stuartarchibald

PR #5532, opened 2020-04-08
Make `numba.<mod>` available without an import

### LINKED ISSUE #5466: 0.49 RC: AttributeError: module 'numba' has no attribute 'numpy_support'

### LINKED ISSUE #5484: Compatibility shims raise cryptic error

### LINKED ISSUE #5528: Numba 0.49.0rc1 does not provide aliases for some use cases
## Reporting a bug

When I use Numba 0.49.0rc1 I see many warnings like this:
```
Import of 'from_dtype' requested from: 'numba.numpy_support', please update to use 'numba.np.numpy_support' or pin to Numba version 0.48.0. This alias will not be present in Numba version 0.50.0.`
```
It is OK and aliases work fine for the most of my cases but does not work for some cases I have:
1. `from numba.targets import quicksort`
1. `from numba.special import literally` and `from numba.special import prange`
1. `from numba import postproc`
1. `import numba` and `numba.analysis.ir_extension_usedefs[Join] = join_usedefs`
1. `from numba import consts`
1. `import numba` and `lower_builtin('getiter', string_array_type)(numba.targets.arrayobj.getiter_array)`
1. `from numba.config import IS_32BITS`
1. `import numba` and `numba.datamodel.register_default(...)`

Please, give me note whether it is possible to be fixed before 0.49.0 release.

sklam:
Based on https://github.com/numba/numba/pull/5497

Make all `mod` from `numba.<mod>` in 0.48.0 available without an import.

Note for py3.6, this disables warnings for `import numba.<mod>` because the `numba/__init__.py` has already imported the modules to make the top-level aliases. This doesn't affect python > 3.6.

Closes #5528 
Closes #5484
Closes #5466


numba/__init__.py
```diff
@@ -58,6 +59,90 @@
 test = runtests.main
 
 
+# needed for compatibility until 0.50.0. Note that accessing members of
+# numpy_support will raise DeprecationWarning
```
stuartarchibald, line 63: Does this note need to be specific to `numpy_support`?

numba/__init__.py
```diff
@@ -58,6 +59,90 @@
...
+}
+
+if sys.version_info < (3, 7):
+    with warnings.catch_warnings():
+        # Ignore warnings from making top-level alias.
+        warnings.simplefilter(
+            "ignore", category=errors.NumbaDeprecationWarning,
+        )
+        for _old_mod, _new_mod in _auto_import_submodules.items():
+            importlib.import_module(_old_mod)
```
stuartarchibald, line 131: should this just iterate `.keys()`?

numba/__init__.py
```diff
@@ -58,6 +59,90 @@
...
+else:
+    def __getattr__(attr):
+        submodule_name = 'numba.{}'.format(attr)
+        try:
+            new_name = _auto_import_submodules[submodule_name]
+        except KeyError:
+            raise AttributeError(
+                f"module {__name__!r} has no attribute {attr!r}",
+            )
```
eric-wieser, line 140:
```suggestion
            ) from None
```

numba/__init__.py
```diff
@@ -58,6 +59,90 @@
...
+        else:
+            errors.deprecate_moved_module(submodule_name, new_name)
+            return importlib.import_module(new_name)
```
eric-wieser, line 143: Another potentially scary option would be:
```python
class _NumbaModule(types.ModuleType):
    def __getattr__(self, attr):
        ...  # as here

...

# as the very last step
_new_module = _NumbaModule(__name__)
_new_module.__dict__.update(globals()
sys.modules[__name__] = _new_module
```

This would make the warning appear on python 3.6 too

numba/__init__.py
eric-wieser, line 63:
```suggestion
# any of these modules, or the moduels themselves, will raise DeprecationWarning
```

numba/targets/__init__.py
```diff
@@ -1,3 +1,7 @@
 import sys
 from numba.core.errors import _MovedModule
-sys.modules[__name__] = _MovedModule(locals(), None)
+
+from numba.misc import quicksort
+
+sys.modules[__name__] = _MovedModule(locals(), None,
+                                     extra_alias={'quicksort': quicksort})
```
eric-wieser, line 7:
```suggestion
sys.modules[__name__] = _MovedModule(locals(), None)
sys.modules[__name__].quicksort = quicksort
```
then you can skip `extra_alias`.

numba/__init__.py
eric-wieser: Which would remove the need for the test changes here, I think

numba/targets/__init__.py
sklam: Yes, that's better. I'll change this and remove the changes to `_MovedModule`

numba/__init__.py
sklam: Yes. It was leftover code

numba/__init__.py
sklam: It does seem very scary to patch the toplevel module

eric-wieser:
From https://github.com/numba/numba/pull/5532#discussion_r406145465, which is erroneously marked as outdated.

> It does seem very scary to patch the toplevel module

Maybe, but your extensive CI would probably catch things going wrong. Possibly worth a try, if only to find out.

numba/__init__.py
```diff
@@ -60,7 +60,7 @@
 
 
 # needed for compatibility until 0.50.0. Note that accessing members of
-# numpy_support will raise DeprecationWarning
+# any of these modules, or the moduels themselves, will raise DeprecationWarning
```
stuartarchibald, line 63:
```suggestion
# any of these modules, or the modules themselves, will raise DeprecationWarning
```

sklam:
RE https://github.com/numba/numba/pull/5532#issuecomment-611534417

There are still unknown risks in patching the top-level numba module. The testsuite covers all the static import uses but none of dynamic manipulation/inspection of the module. I rather we are not showing enough warnings for py3.6 then risking breaking the top-level modules in some cases.

sklam:
We also need to push out a RC2 as soon as possible because of scheduled maintenance we have in our buildfarm over the weekend. Usually takes us 24hrs to build + test.

eric-wieser:
Both good reasons

PR #5543, opened 2020-04-10
np.trim_zeros

sungraek:
Add support for `np.trim_zeros`. It currently does not handle nested list or tuple input well.
#4074

### DIFF
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -4255,6 +4255,8 @@ def cross2d(a, b):
 
 @overload(np.trim_zeros)
 def np_trim_zeros(a, trim='fb'):
+    if not type_can_asarray(a):
+        raise TypingError("Input must be array-like.")
     if hasattr(a, 'ndim') and a.ndim > 1:
         raise TypingError('array must be 1D')
 
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -4261,21 +4261,21 @@ def np_trim_zeros(a, trim='fb'):
         raise TypingError('array must be 1D')
 
     def impl(a, trim='fb'):
-        a = np.ravel(np.asarray(a))
+        a_ = np.asarray(a)
         first = 0
         if 'f' in trim:
-            for i in a:
+            for i in a_:
                 if i != 0:
                     break
                 else:
                     first = first + 1
         last = len(a)
         if 'b' in trim:
-            for i in a[::-1]:
+            for i in a_[::-1]:
                 if i != 0:
                     break
                 else:
                     last = last - 1
-        return np.ravel(a[first:last])
+        return a_[first:last]
 
     return impl
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -4259,6 +4259,8 @@ def np_trim_zeros(a, trim='fb'):
         raise TypeError('The first argument must be an array-like')
     if hasattr(a, 'ndim') and a.ndim > 1:
         raise TypingError('array must be 1D')
+    if not isinstance(trim, (str, types.UnicodeType, types.StringLiteral)):
+        raise TypeError('The second argument must be a string')
 
     def impl(a, trim='fb'):
         a_ = np.asarray(a)
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -4263,6 +4263,7 @@ def np_trim_zeros(a, trim='fb'):
     def impl(a, trim='fb'):
         a_ = np.asarray(a)
         first = 0
+        trim = trim.lower()
         if 'f' in trim:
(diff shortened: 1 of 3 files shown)


gmarkall:

numba/np/arraymath.py
```diff
@@ -4251,3 +4251,31 @@ def impl(a, b):
...
+def np_trim_zeros(a, trim='fb'):
+    if not isinstance(a, (types.Array, types.Sequence, types.Tuple)):
+        raise TypeError('The first argument must be an array-like')
+    if hasattr(a, 'ndim') and a.ndim > 1:
+        raise TypingError('array must be 1D')
+
+    def impl(a, trim='fb'):
+        a_ = np.asarray(a)
+        first = 0
+        if 'f' in trim:
```
EPronovost, line 4266: From [this line](https://github.com/numpy/numpy/blob/v1.18.1/numpy/lib/function_base.py#L1608) it seems the numpy implementation works with upper and lower case letters for `trim`

```
>>> np.trim_zeros(np.array([0, 1, 2, 3, 0, 0]), trim="FB")
array([1, 2, 3])
```

numba/tests/test_np_functions.py
```diff
@@ -3621,6 +3625,49 @@ def test_cross2d_exceptions(self):
...
 
+    def test_trim_zeros(self):
+
+        def arrays():
+            yield np.array([])
+            yield np.zeros(5)
+            yield np.zeros(1)
+            yield np.array([1, 2, 3])
+            yield np.array([0, 1, 2, 3])
+            yield np.array([0, 1, 2, 0, 0])
```
EPronovost, line 3636: Could we test arrays that have non-zero values that are not ints (e.g. `np.array([0., 1., 2., np.nan, 0.])`)?
We could also test other data types that work with the numpy version (e.g. `np.array(["Hello", "world"])`).

numba/tests/test_np_functions.py
```diff
@@ -3621,6 +3625,49 @@ def test_cross2d_exceptions(self):
...
+
+        pyfunc = np_trim_zeros
+        cfunc = jit(nopython=True)(pyfunc)
```
EPronovost, line 3639: Can we also test when `trim != "fb"`?

EPronovost:
Nice work! You'll probably want to update [the numpy supported features documentation](https://github.com/numba/numba/blob/master/docs/source/reference/numpysupported.rst) as part of your PR as well.

sungraek:
I added support for both lower and upper case arguments and some new test cases as you mentioned.

numba/np/arraymath.py
```diff
@@ -4251,3 +4251,32 @@ def impl(a, b):
...
 
     return impl
+
+
+@overload(np.trim_zeros)
+def np_trim_zeros(a, trim='fb'):
+    if not isinstance(a, (types.Array, types.Sequence, types.Tuple)):
+        raise TypeError('The first argument must be an array-like')
+    if hasattr(a, 'ndim') and a.ndim > 1:
+        raise TypingError('array must be 1D')
```
EPronovost, line 4261: Should we check that `trim` is a string or string literal?

sungraek: It seems like `np.trim_zeros(arr)` `np.trim_zeros(arr, 'f')` and `np.trim_zeros(arr, trim)` all produces expected output. 
Would we still need to add a check?

EPronovost: It would improve the error message if someone does something like `np.trim_zeros(arr, 1)`.

sungraek: Oh, I see. I will add the check

ionikha:
Hi, i'm working on a project that uses trim_zeros and numba, do you know why this wasn't added?
It looks good to go?

gmarkall:
@ionikha Thanks for the ping, I think it just fell by the wayside and never got a review from a maintainer (even though @EPronovost very kindly gave it a review). I've added this to the agenda for the dev meeting tomorrow to see if we can kickstart it.

gmarkall:
Assignment from triage meeting: @gmarkall to merge master and address any review comments, @guilhermeleobas to review subsequently.

guilhermeleobas:
@gmarkall, if needed, we can switch and I can merge `main` and address any comments. Just let me know

esc:
@gmarkall @guilhermeleobas -- can I ask for an update here? What is the status of this?

gmarkall:
I'll put this on my near-term todo list so that it can get done.

guilhermeleobas:
@gmarkall, I've created https://github.com/numba/numba/pull/9074 which is built on top of this PR with git conflicts fixed.

esc:
Is replaced by #9074

PR #5544, opened 2020-04-10
Add support for np.union1d

shangbol:
Pull request to add support for np.union1d from issue #4074

### DIFF
--- a/docs/source/reference/numpysupported.rst
+++ b/docs/source/reference/numpysupported.rst
@@ -437,6 +437,7 @@ The following top-level functions are supported:
 * :func:`numpy.triu` (second argument ``k`` must be an integer)
 * :func:`numpy.triu_indices` (all arguments must be integer)
 * :func:`numpy.triu_indices_from` (second argument ``k`` must be an integer)
+* :func:`numpy.union1d` (only supports arrays of the same dtype)
 * :func:`numpy.unique` (only the first argument)
 * :func:`numpy.vander`
 * :func:`numpy.vstack`
--- a/docs/source/reference/numpysupported.rst
+++ b/docs/source/reference/numpysupported.rst
@@ -437,7 +437,7 @@ The following top-level functions are supported:
 * :func:`numpy.triu` (second argument ``k`` must be an integer)
 * :func:`numpy.triu_indices` (all arguments must be integer)
 * :func:`numpy.triu_indices_from` (second argument ``k`` must be an integer)
-* :func:`numpy.union1d` (only supports arrays of the same dtype)
+* :func:`numpy.union1d` (For unicode arrays, only supports arrays of the same dtype)
 * :func:`numpy.unique` (only the first argument)
 * :func:`numpy.vander`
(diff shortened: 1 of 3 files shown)


gmarkall:

shangbol:
> This looks pretty good to me, though I have one query - the documentation states that it only supports arrays of the same dtype. However, if I try the following:
> 
> ```python
> from numba import njit
> import numpy as np
> 
> 
> @njit
> def test_union(arr1, arr2):
>     return np.union1d(arr1, arr2)
> 
> x = np.array([3, 7, 6])
> y = np.array([7, 3, 4]).astype(np.float32)
> 
> print(x.dtype)
> print(y.dtype)
> print(np.union1d(x, y))
> print(test_union(x, y))
> ```
> 
> I get:
> 
> ```
> int64
> float32
> [3. 4. 6. 7.]
> [3. 4. 6. 7.]
> ```
> 
> which suggests that it appears to work, with behaviour that seems to match NumPy. Have I misunderstood the documentation, or should this be a supported / tested case?

Yeah, I wasn't sure exactly how to word it for the documentation. Basically, the functionality depends on the type conversion in np.concatentate; it does work in most instances, like type promotion from int to float, or bool to int etc. (I think the unit tests do cover this functionality). 

The reason I wanted to add the note in the documentation was because the type conversion doesn't work well between unicode characters of different types, such as U<12 and U<1. It seems to be caused by the type conversion in np.concatentate, where using numpy's concatenate will work on different unicode dtypes but the current Numba implementation does not.

I wasn't sure how exactly to deal with this and so welcome any recommendations or advice.

Thanks!

gmarkall:
What about stating the caveat in the documentation as "For unicode character arrays, only characters of the exact same type are supported", and adding a check to `np_union1d` that raises a `TypingError` when one array is a unicode character type and the two arrays have different dtypes?

shangbol:
> What about stating the caveat in the documentation as "For unicode character arrays, only characters of the exact same type are supported", and adding a check to `np_union1d` that raises a `TypingError` when one array is a unicode character type and the two arrays have different dtypes?

Sounds good. Should be fine now.
Thanks!

esc:
@shangbol thanks for updating the issue. I have set the label accordingly.

gmarkall (dismissed):
Looks good to me!

gmarkall:
I'll leave the label as "Waiting on reviewer", since a core maintainer might want to look before this is declared ready to merge.

gmarkall:
Sorry, accidentally hit the close button.

TermeHansen:
Why was this never merged?

gmarkall:
> Why was this never merged?

Because it never got a look from a core maintainer when I suggested that one look over it. At the time, I was not a core maintainer. I'm not sure we use that term now, but I have pretty much the status that that implied now, so I'll take another look at this soon.

gmarkall:
I think this is still good and I've merged `main` to resolve conflicts. Just waiting on CI.

sklam (approved):

PR #5548, opened 2020-04-11
Fix #5537 Removed reference to `platform.linux_distribution`

### LINKED ISSUE #5537: `platform.linux_distribution` is deprecated
I was looking into the output of `numba -s` and found that the code is using `platform.linux_distribution` which is now [deprecated](https://bugs.python.org/issue1322).

The issue is located in [`numba_entry.py` line 98](https://github.com/numba/numba/blob/2b613798cf548e247e81b54b2cf0930d9865040b/numba/misc/numba_entry.py#L98)

A suggested alternative is to use the [distro package](https://pypi.org/project/distro/) or use some form of custom heuristic. 

Another way around it would be to simply ignore this information!

jagaudin:
Removed reference to `platform.linux_distribution` in `numba/misc/numba_entry.py`.

### DIFF
--- a/numba/misc/numba_entry.py
+++ b/numba/misc/numba_entry.py
@@ -95,7 +95,7 @@ def get_sys_info():
         print(fmt % ("Version", platform.version()))
         try:
             if system_name == 'Linux':
-                info = platform.linux_distribution()
+                info = ()
             elif system_name == 'Windows':
                 info = platform.win32_ver()
             elif system_name == 'Darwin':


stuartarchibald:
Looks good, thanks for the fix.

stuartarchibald:
@jagaudin are you happy to move this out of draft status? Would be great to get it merged. Thanks.

jagaudin:
@stuartarchibald I think I've changed the status now. Let me know if there's anything else needed.

stuartarchibald:
@jagaudin congratulations on your first contribution to Numba!

PR #5549, opened 2020-04-12
Check type getitem

EPronovost:
Check that calls to `Type.__getitem__` are slices with no start or stop (e.g. `intp[:]`).
Raises a KeyError if not.
Previously, `list(intp)` would infinite loop.


stuartarchibald:
Thanks for the PR, much appreciated. Looks like there's a couple of test fails on CI at present, marked as "In progress".

EPronovost:
Yeah there are several places that depend on the old behavior that I need to fix still.  I wanted to see what you thought about it, or if there were any valid use cases of indexing by something other than an open slice that I'm not aware of.

EPronovost:
@stuartarchibald I think the CI errors are fixed.

stuartarchibald:
Thanks for fixing, queued for review.

numba/stencils/stencil.py
```diff
@@ -257,7 +257,10 @@ def add_indices_to_kernel(self, kernel, index_names, ndim,
                                                        const_index_vars[dim], loc)
                             new_body.append(ir.Assign(getitemcall, getitemvar, loc))
                             # Get the type of this particular part of the index tuple.
-                            one_index_typ = stmt_index_var_typ[dim]
+                            if isinstance(stmt_index_var_typ, types.ConstSized):
```
sklam, line 260: Why is this change needed?

sklam: nvm, I found the affected test `test_basic81 (numba.tests.test_stencils.TestManyStencils)`

PR #979, opened 2015-02-04
Add cuda.atomic.max()

gmarkall:
(based on the atomic_double_add branch)

cuda.atomic.max(array, offset, val) sets:

```python
array[offset] = max(array[offset], val)
```

atomically.

PR #5556, opened 2020-04-14
CUDA: Make atomic semantics match Python / NumPy, and fix #5458

### LINKED ISSUE #5458: CUDA simulator atomic operation differs to documentation
According to [docs](https://numba.pydata.org/numba-doc/dev/cuda/intrinsics.html), the CUDA atomic `add` and `max` operations return the old value at the index location, but the fake [implementation](https://github.com/numba/numba/blob/master/numba/cuda/simulator/kernelapi.py#L106) in the CUDA simulator does not return anything.

Is that intentional in the simulator? Does the actual CUDA operation have a return value?
Thank you.

gmarkall:
The main aim for this PR is to fix #5458 by ensuring that the simulator's atomic implementations return the old value. It also includes a change in the semantics of `cuda.atomic.min` and `cuda.atomic.max` such that they now match the semantics of `min` and `max` in Python and NumPy.

There doesn't appear to be a good reason for the semantics of `cuda.atomic.min` and `max` deviating from NumPy - the original PR which added it (#979) makes no comment about it. Treating NaN values as missing data also makes it more difficult to fix Issue #5458 in a manner that is consistent between the simulator and the hardware target.

The first commit in this PR changes the CUDA atomic `min` and `max` functions to match the semantics of Python and NumPy - that is, `{min,max}(a, b)` always returns `a` if either is a NaN. In practice, for:

```python
cuda.atomic.{min,max}(ary, idx, val)
````

the stored result is always `ary[idx]` when either of `ary[idx]` or `val` is a NaN.

The second commit then fixes Issue #5458.

### DIFF
--- a/numba/cuda/cudadrv/nvvm.py
+++ b/numba/cuda/cudadrv/nvvm.py
@@ -388,18 +388,19 @@ done:
 }
 """
 
+
 ir_numba_atomic_max = """
 define internal {T} @___numba_atomic_{T}_max({T}* %ptr, {T} %val) alwaysinline {{
 entry:
     %ptrval = load volatile {T}, {T}* %ptr
-    ; Check if val is a NaN and return *ptr early if so
-    %valnan = fcmp uno {T} %val, %val
+    ; Check if either value is a NaN and return *ptr early if so
+    %valnan = fcmp uno {T} %val, %ptrval
     br i1 %valnan, label %done, label %lt_check
 
 lt_check:
     %dold = phi {T} [ %ptrval, %entry ], [ %dcas, %attempt ]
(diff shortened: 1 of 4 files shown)


sklam:
smoking at `numba_smoketest_cuda_55`

stuartarchibald:
Buildfarm passed.

uchytilc:
I saw this change in the change log and was wondering if there would be a way to revert the change via an option or something to allow atomic min/max to return the non-Nan value always. I have found this a very useful feature. For example it prevents the need to initialize different values when trying to find a max/min reduction (i.e. -inf vs +inf).

gmarkall:

Recorded in #5820.

PR #5557, opened 2020-04-14
First-class function call cannot use keyword arguments

### LINKED ISSUE #5540: named parameters not recognized with first class functions
- [x ] I am using the latest ~~released~~ **master** version of Numba (most recent is visible in
- [x ] I have included below a minimal working reproducer (if you are unsure how

There is a problem recognizing named parameters. The following fails

```python
from numba import njit, types

@njit(types.int64(types.int64))
def foo(x):
    return x+1

@njit
def loop(foos):
    f = foos[0]
    return f(x=1)

loop((foo,))

Internal error at <numba.core.typeinfer.CallConstraint object at 0x7f67f39a1a30>.
mismatch of arguments number: 0 vs 1
[1] During: resolving callee type: UndefinedFunctionType[undefined(undefined)20]
```

while this works

```python
@njit
def loop(foos):
    f = foos[0]
    return f(1)

loop((foo,))
```

pearu:
The PR implements a proper error message as in the description.

Fixes #5540

### DIFF
--- a/numba/core/types/function_type.py
+++ b/numba/core/types/function_type.py
@@ -4,7 +4,7 @@ __all__ = ['FunctionType', 'UndefinedFunctionType', 'FunctionPrototype',
 
 from abc import ABC, abstractmethod
 from .abstract import Type
-from .. import types
+from .. import types, errors
 
 
 class FunctionType(Type):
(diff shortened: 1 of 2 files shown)


sklam:
current test failure is from issues on master.

sklam:
The fix looks good. Just waiting for master to stabilize the re-run the CI.

Issue #4574, opened 2019-09-16
Default Compilation mode for CUDA Kernels


Nischay-Pro:
By default does Numba compile in separate compilation mode.

stuartarchibald:
Thanks for the query. Please could you expand on what it is you mean by this?!


> By default does Numba compile in separate compilation mode.

Nischay-Pro:
I would like to know the behavior of Numba when an application has multiple device kernels.

Does each kernel when compiled have it's own separate ptx code? Does it generate Relocatable Device Code?

This article might explain what I'm trying to say.
https://devblogs.nvidia.com/separate-compilation-linking-cuda-device-code/

malays:
@stuartarchibald ,
Specifically, does numba set the "-dc" flag of NVCC when compiling?

seibert:
Numba uses libNVVM to compile NVVM IR directly from a string in memory.  It never invokes NVCC (in fact, NVCC doesn't need to be present on the system), so I'm not quite sure what the equivalent to `-rdc` is.

Looking at the compile function: https://docs.nvidia.com/cuda/libnvvm-api/group__compilation.html#group__compilation_1g76ac1e23f5d0e2240e78be0e63450346
I'm not sure if the option we need is there.

malays:
Thanks @seibert !
I dug some more. It seems it indeed is not possible to perform separate compilation with libnvvm.
I also verified that the code me and @Nischay-Pro were looking at was not separate-compiled. I'm very new to numba, so took some time to figure out the flow.

Thanks for the help and information!

leofang:
Hi guys,

Look at the following lines from #4551 
https://github.com/numba/numba/blob/9a6921365e11a3442944c4daff8a7bb0d174fae0/numba/cuda/tests/cudapy/test_cooperative_groups.py#L22-L26
and scroll a few lines down to see how `LINK` is passed to `cuda.jit` to link to the CUDA device runtime library. The only task left is to find a way to pass in `-rdc`. I know nothing about how libNVVM works though...

This issue is directly related to my comment on dynamical parallelism: https://github.com/numba/numba/pull/4551#issuecomment-530187364.

leofang:
I have a naive question: is LLVM also used as part of the kernel compilation in Numba, or only libNVVM? I noticed that there are some effort on the LLVM side to support relocatable device code/separate compilation with the option `-fcuda-rdc`, see, e.g., https://reviews.llvm.org/D42921 and the subsequent PRs.

[...]

PR #5561, opened 2020-04-14
CUDA: Add function for compiling to PTX (+ other small fixes)

gmarkall:
The main purpose of this PR is to add a function that enables compilation of a Python function to PTX for a given set of argument types, without invoking any of the CUDA driver. This can be useful for use cases where the compiled function is not to be invoked by Numba, but instead inlined into other PTX code from another source (e.g. from outside the Numba or Python ecosystem). This PR is currently in a WIP state to solicit feedback on the interface from potential users (cc @kkraus14 @hummingtree @galipremsagar).

The proposed interface is documented at: https://gmarkall.github.io/numba-rtd-theme/cuda-reference/host.html#compilation - once a sensible / useful API is agreed on, I will add some tests and potentially expand on the documentation.

Along the way of creating this function, I came across a couple of other bits that could have done with some tidy up, that are also in this PR at the moment, but not directly connected:

* Addition of supported CCs in CUDA 9 and 10.
* Fix flake8 errors for `numba.cuda.compiler`, and remove if from the excludes in `.flake8`.


gmarkall:
I've added an empty `compile_ptx` stub to enable the simulator to continue to work for this PR (to fix the CI error with a5f2713), but I hope for a "camera-ready" implementation that it will be possible to compile to PTX even with the simulator (or rather, without actual hardware available, so that the compiler can still be used even if the runtime isn't).

hummingtree:
It would be great to have this functionality in `numba`. In `RAPIDS/cuDF` the following is a code snippet of what we have right now to compile udf into a CUDA PTX device function
```
def compile_udf(udf, type_signature):
    decorated_udf = cuda.jit(udf, device=True)
    compiled = decorated_udf.compile(type_signature)
    ptx_code = decorated_udf.inspect_ptx(type_signature).decode("utf-8")
    output_type = numpy_support.as_dtype(compiled.signature.return_type)
    return (ptx_code, output_type.type)
```
(See https://github.com/rapidsai/cudf/blob/branch-0.14/python/cudf/cudf/utils/cudautils.py#L259)

Currently we tell the compiler the input types, and let the compiler decide the output type. This is certainly not optimal, since, for example, for an input type of `int32`, the compiler almost always prefers an output type of `int64`, even in cases where `int32` is fairly reasonable as an output type. As an example the python lambda `lambda x: x + 1` with input `int32` is compiled as

```
.visible .func  (.param .b32 func_retval0) _ZN8__main__16$3clambda$3e$241Ei(
        .param .b64 _ZN8__main__16$3clambda$3e$241Ei_param_0,
        .param .b32 _ZN8__main__16$3clambda$3e$241Ei_param_1
)
{
        .reg .b32       %r<2>;
        .reg .b64       %rd<4>;


        ld.param.u64    %rd1, [_ZN8__main__16$3clambda$3e$241Ei_param_0];
        ld.param.s32    %rd2, [_ZN8__main__16$3clambda$3e$241Ei_param_1];
        add.s64         %rd3, %rd2, 1;
        st.u64  [%rd1], %rd3;
        mov.u32         %r1, 0;
        st.param.b32    [func_retval0+0], %r1;
        ret;
}

```
and the output type is `int64`.

I believe telling the python udf ->PTX compiler to have a _specific_ output type is difficult/impossible. It would be nice to have an option to **hint** the compiler whether we want to have an output type with higher or lower precision. The expected behavior is that the compiler tends to use higher precision output types (`int64` or `float64`) if it is hinted with the higher precision option, and vice versa.

gmarkall:
@hummingtree Thanks for the feedback - I'll take another go at producing an iteration of the proposed API to try and meet those requirements (along with also being able to get the PTX return type back from the function, that you mentioned to me out-of-band)

numba/cuda/compiler.py
```diff
@@ -74,6 +71,50 @@ def compile_kernel(pyfunc, args, link, debug=False, inline=False,
...
+
+    :param pyfunc: The Python function to compile.
+    :param args: A tuple of argument types to compile for.
+    :param debug: Whether to include debug info in the generated PTX.
+    :type debug: bool
+    :param fastmath: Whether to enable fast math flags (ftz=1, prec_sqrt=0,
+                     prec_div=, and fma=1)
+    :type fastmath: bool
+    :param cc: Compute capability to compile for, as a tuple ``(MAJOR, MINOR)``.
+               Defaults to ``(5, 0)``.
```
leofang, line 86: Could you default to the current device's CC if `cc=None`?

numba/cuda/compiler.py
```diff
@@ -74,6 +71,50 @@ def compile_kernel(pyfunc, args, link, debug=False, inline=False,
...
+    :type cc: tuple
+    :param opt: Enable (``3``) or disable (``0``) optimizations. Defaults to
+                ``3``.
+    :type opt: int
+    :return: PTX code generated by NVVM.
```
leofang, line 91: Perhaps an obvious question 😅: what's the encoding of the returned PTX code?

numba/cuda/compiler.py
gmarkall: I'm trying to avoid `compile_ptx` having anything to do with devices - I picked 5.0 because I thought it was the lowest non-deprecated compute capability... However upon double-checking the release notes for 10.2 (https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#deprecated-features) I see that is deprecated too, so I should probably use 5.2 instead.

numba/cuda/compiler.py
gmarkall: I've generally been decoding it as `utf-8`, but I don't know for certain - will see if I can work out the definitive answer.

gmarkall:
I've pushed a few changes now (details on the commit message).

@leofang I've changed the default to 5.2 - did you have any further thoughts on my motivation for not defaulting to the CC of the current device?

@hummingtree I've now changed the interface so that you can compile a kernel or a device function, which should then provide a drop-in replacement for the implementation in `compile_udf` - unfortunately the suggested return type isn't that easy to wire up to the internals of the compiler, so I need to have a look at whether this is possible. For now `compile_ptx` accepts the argument types and returns the inferred return type along with the PTX.

numba/cuda/compiler.py
leofang: Thanks for the pointer, Graham. Wasn't aware of the deprecation. Yeah I did not realize device independency is done here, so it makes sense to pick one.

But, I find the release note rather confusing -- a few lines below it says 5.2 is also deprecated:
> Support for Kepler and Maxwell architectures (compute capability sm_35 through sm_52) is deprecated.

My impression from working with half-precision is that `sm_53` is the minimum requirement -- can you confirm?

numba/cuda/compiler.py
```diff
@@ -74,6 +71,56 @@ def compile_kernel(pyfunc, args, link, debug=False, inline=False,
...
+    :param opt: Enable optimizations. Defaults to ``True``.
+    :type opt: bool
+    :return: PTX code generated by NVVM.
+    :rtype: bytes
+    """
+    cres = compile_cuda(pyfunc, None, args, debug=debug)
+    resty = cres.signature.return_type
+    if device:
+        llvm_module = cres.library._final_module
+        nvvm.fix_data_layout(llvm_module)
```
leofang, line 98: 1. The `device` argument is not documented in the docstring
2. Is this the option for generating relocatable device code (like `-rdc` for nvcc and nvrtc)?

leofang: Sorry, misread the discussions above 😅 So `device` is for generating device functions (`__device__`). My questions above still stand, though.

gmarkall: Thanks - I've added some documentation, and it is indeed for compiling a `device` function. As a result of your question, I did do some experimentation with the `-rdc` flag in https://github.com/gmarkall/ptx-rdc-comparison, which also relates to https://github.com/numba/numba/issues/4574.

numba/cuda/compiler.py
gmarkall: I think the release notes are a bit confusing, but from my reading of it, up to `sm_50` is deprecated in all of CUDA. However, for the CUDA libraries (cuBLAS, etc.) `sm_52` is also deprecated. I don't know anything about half-precision specifically.

As PTX is not specific to CUDA libraries, but (probably) falls under "General CUDA", I'm inclined to leave the default at `(5, 2)` - it seems that 5.3 is only implemented in Jetson, Tegra, and DRIVE, which are probably not the majority of use cases for Numba - if I were to pick a higher default than `(5, 2)` it would feel more appropriate to use `(6, 0)` instead.

gmarkall:
@leofang I've added another function, `compile_ptx_for_current_device` - does this address the use case you were thinking of when you asked about making the default the current device's compute capability?

numba/cuda/compiler.py
leofang: I see, yeah l think it's a good point. Thanks for clarifying.

gmarkall:
@hummingtree I've tested cuDF branch-0.14 with the following patch and this branch:

```diff
--- a/python/cudf/cudf/utils/cudautils.py
+++ b/python/cudf/cudf/utils/cudautils.py
@@ -258,14 +258,14 @@ def window_sizes_from_offset(arr, offset):
 
 @lru_cache(maxsize=32)
 def compile_udf(udf, type_signature):
-    """Copmile ``udf`` with `numba`
+    """Compile ``udf`` with `numba`
 
     Compile a python callable function ``udf`` with
-    `numba.cuda.jit(device=True)` using ``type_signature`` into CUDA PTX
-    together with the generated output type.
+    `numba.cuda.compile_ptx_for_current_device(device=True)` using
+    ``type_signature`` into CUDA PTX together with the generated output type.
 
     The output is expected to be passed to the PTX parser in `libcudf`
-    to generate a CUDA device funtion to be inlined into CUDA kernels,
+    to generate a CUDA device function to be inlined into CUDA kernels,
     compiled at runtime and launched.
 
     Parameters
@@ -287,8 +287,7 @@ def compile_udf(udf, type_signature):
       An numpy type
 
     """
-    decorated_udf = cuda.jit(udf, device=True)
-    compiled = decorated_udf.compile(type_signature)
-    ptx_code = decorated_udf.inspect_ptx(type_signature).decode("utf-8")
-    output_type = numpy_support.as_dtype(compiled.signature.return_type)
+    ptx_code, return_type = \
+        cuda.compile_ptx_for_current_device(udf, type_signature, device=True)
+    output_type = numpy_support.as_dtype(return_type)
     return (ptx_code, output_type.type)
```

From my quick testing, it makes a bit of difference to the performance of the test suite. Without the patch I get (running with 8 processes on an i7-6770K machine):

```
20660 passed, 923 skipped, 280 xfailed, 23 xpassed, 2548 warnings in 316.52s (0:05:16) 
```

with the patch, I get:

```
20660 passed, 923 skipped, 280 xfailed, 23 xpassed, 2548 warnings in 308.77s (0:05:08)
```

I know that using a test suite is often a poor benchmark, but I think this suggests that using `compile_ptx_for_current_device` does remove some of the overhead of compiling UDFs, because Numba is now not needlessly doing some extra work inserting the compiled functions into its typing contexts, and other housekeeping.

numba/cuda/compiler.py
leofang: Thanks, Graham! 🙏 Will take a look later and move the discussion to #4574.

leofang:
> I've added another function, `compile_ptx_for_current_device` - does this address the use case you were thinking of when you asked about making the default the current device's compute capability?

Yes, I think so! It will be a handy API. Thanks 🙂

docs/source/cuda-reference/host.rst
```diff
@@ -91,6 +91,30 @@ the functionality of the selected device:
...
+
+Numba provides an entry point for compiling a Python function to PTX without
+invoking any of the driver API. This can be useful for:
+
+- Generating PTX that is to be inlined into other PTX code (e.g. from outside
+  the Numba / Python ecosystem).
+- Generating code when there is no device present.
+- Generating code prior to a fork without initializing CUDA.
+
+.. note:: It is the user's responsiblity to manage any ABI issues arising from
```
sklam, line 106:
```suggestion
.. note:: It is the user's responsibility to manage any ABI issues arising from
```

numba/cuda/compiler.py
```diff
@@ -74,6 +71,69 @@ def compile_kernel(pyfunc, args, link, debug=False, inline=False,
...
+        lib, kernel = tgt.prepare_cuda_kernel(cres.library, fname,
+                                              cres.signature.args, debug=debug)
+        llvm_module = lib._final_module
+
+    options = {
+        'debug': debug,
+    }
+
+    if fastmath:
+        options.update(dict(ftz=True,
```
sklam, line 114: We should try not to duplicate this

numba/cuda/compiler.py
```diff
@@ -74,6 +71,69 @@ def compile_kernel(pyfunc, args, link, debug=False, inline=False,
...
+                            prec_sqrt=False,
+                            prec_div=False,
+                            fma=True))
+
+    cc = cc or (5, 2)
```
sklam, line 119: It should be a global variable to avoid duplicating a magic constant.

numba/cuda/compiler.py
```diff
@@ -74,6 +71,69 @@ def compile_kernel(pyfunc, args, link, debug=False, inline=False,
...
+    :type device: bool
+    :param fastmath: Whether to enable fast math flags (ftz=1, prec_sqrt=0,
+                     prec_div=, and fma=1)
+    :type fastmath: bool
+    :param cc: Compute capability to compile for, as a tuple ``(MAJOR, MINOR)``.
+               Defaults to ``(5, 2)``.
+    :type cc: tuple
+    :param opt: Enable optimizations. Defaults to ``True``.
+    :type opt: bool
+    :return: PTX code generated by NVVM.
```
sklam, line 94: It's returning a 2-tuple `(ptx, return_type)`

hummingtree:
> @hummingtree I've tested cuDF branch-0.14 with the following patch and this branch:
> 
> ```diff
> --- a/python/cudf/cudf/utils/cudautils.py
> +++ b/python/cudf/cudf/utils/cudautils.py
> @@ -258,14 +258,14 @@ def window_sizes_from_offset(arr, offset):
>  
>  @lru_cache(maxsize=32)
>  def compile_udf(udf, type_signature):
> -    """Copmile ``udf`` with `numba`
> +    """Compile ``udf`` with `numba`
>  
>      Compile a python callable function ``udf`` with
> -    `numba.cuda.jit(device=True)` using ``type_signature`` into CUDA PTX
> -    together with the generated output type.
> +    `numba.cuda.compile_ptx_for_current_device(device=True)` using
> +    ``type_signature`` into CUDA PTX together with the generated output type.
>  
>      The output is expected to be passed to the PTX parser in `libcudf`
> -    to generate a CUDA device funtion to be inlined into CUDA kernels,
> +    to generate a CUDA device function to be inlined into CUDA kernels,
>      compiled at runtime and launched.
>  
>      Parameters
> @@ -287,8 +287,7 @@ def compile_udf(udf, type_signature):
>        An numpy type
>  
>      """
> -    decorated_udf = cuda.jit(udf, device=True)
> -    compiled = decorated_udf.compile(type_signature)
> -    ptx_code = decorated_udf.inspect_ptx(type_signature).decode("utf-8")
> -    output_type = numpy_support.as_dtype(compiled.signature.return_type)
> +    ptx_code, return_type = \
> +        cuda.compile_ptx_for_current_device(udf, type_signature, device=True)
> +    output_type = numpy_support.as_dtype(return_type)
>      return (ptx_code, output_type.type)
> ```
> 
> From my quick testing, it makes a bit of difference to the performance of the test suite. Without the patch I get (running with 8 processes on an i7-6770K machine):
> 
> ```
> 20660 passed, 923 skipped, 280 xfailed, 23 xpassed, 2548 warnings in 316.52s (0:05:16) 
> ```
> 
> with the patch, I get:
> 
> ```
> 20660 passed, 923 skipped, 280 xfailed, 23 xpassed, 2548 warnings in 308.77s (0:05:08)
> ```
> 
> I know that using a test suite is often a poor benchmark, but I think this suggests that using `compile_ptx_for_current_device` does remove some of the overhead of compiling UDFs, because Numba is now not needlessly doing some extra work inserting the compiled functions into its typing contexts, and other housekeeping.

This is great! Thanks @gmarkall .

leofang:
Sorry one more stupid question: How do I specify the argument types with `args`?

gmarkall:
@leofang It's a tuple of numba types, e.g. `from numba import float32; args = (float32, float32)`.

numba/cuda/compiler.py
gmarkall: I couldn't find an answer, but I've put decoding it inside the `compile_ptx` function, so it doesn't need to concern the user, and if that ever turns out not to work, we can treat it as a bug and fix it appropriately.

gmarkall:
I've added some tests and addressed the feedback received so far - I think this is ready for (further) review now - i.e. not WIP anymore.

gmarkall:
Forgot to mention - I've decided not to make PTX compilation available when the simulator is enabled, as that would be a very niche use case, and can be added later if it turns out to actually seem like it would be useful for anyone.

numba/cuda/compiler.py
```diff
@@ -74,6 +71,69 @@ def compile_kernel(pyfunc, args, link, debug=False, inline=False,
...
+        lib, kernel = tgt.prepare_cuda_kernel(cres.library, fname,
+                                              cres.signature.args, debug=debug)
+        llvm_module = lib._final_module
+
+    options = {
+        'debug': debug,
+        'fastmath': fastmath,
+    }
+
+    opt = opt and 3 or 0
```
sklam, line 120: I think a `3 if opt else 0` is easier to read.

numba/cuda/compiler.py
```diff
@@ -74,6 +71,69 @@ def compile_kernel(pyfunc, args, link, debug=False, inline=False,
     return cukern
 
 
+# The default compute capability to target when compiling to PTX. This is
+# chosen as the lowest non-deprecated compute capability in the most recent
+# version of the CUDA toolkit supported (10.2 at present).
+DEFAULT_PTX_CC = (5, 2)
```
sklam, line 77: Should this be in `config`?

numba/cuda/compiler.py
gmarkall: Updated to match your suggestion.

numba/cuda/compiler.py
gmarkall: Moved to config and added appropriate documentation.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_57`.

stuartarchibald:
This is failing on CUDA 9 on all support platforms.
```
ERROR: test_nvvm_support (numba.cuda.tests.cudadrv.test_nvvm_driver.TestNvvmDriver)
Test supported CC by NVVM.
< trace back is roughly >
File "numba/cuda.cudadrc/nvvm.py", line 506 in llvm_to_ptx
ptx = cu.compile(**opts)
...stuff

libnvvm: error: -arch=compute_72 is an unsupported option
NVVM_ERROR_INVALID_OPTION
```

gmarkall:
Thanks for the pointer - this should be resolved now.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_58`.

stuartarchibald:
Buildfarm passed.

docs/source/cuda-reference/host.rst
```diff
@@ -109,12 +109,16 @@ invoking any of the driver API. This can be useful for:
 .. autofunction:: numba.cuda.compile_ptx
 
 
-If PTX for the compute capability of the current device is required, the
-``compile_ptx_for_current_device`` function can be used:
+The configuration variable ``CUDA_DEFAULT_PTX_CC`` can be set to control the
```
stuartarchibald, line 112: Should that be `NUMBA_CUDA_DEFAULT_PTX_CC` ?

stuartarchibald:
@sklam are you happy to approve this? Buildfarm is happy.

docs/source/cuda-reference/host.rst
gmarkall: The environment variable is `NUMBA_CUDA_DEFAULT_PTX_CC`, but the variable in `numba.config` is called `CUDA_DEFAULT_PTX_CC`, which I refer to as a "configuration variable" - is this terminology OK / understandable?

stuartarchibald: It's clear to me, however, my concern is, as a new user, would it be clear from this how to set it? `numba.config` use isn't particularly well defined and the semantics of use are nuanced, whereas setting an env var is well known?

gmarkall: OK - changed to refer to the environment variable.

stuartarchibald: Great, thanks!

gmarkall:
Thanks for the merge! Glad this made it for 0.50 :-)

PR #5573, opened 2020-04-16
CUDA: Skip tests under cuda-memcheck that hang it

gmarkall:
There are a few tests that hang cuda-memcheck, involving spawning threads or processes. It is useful to occasionally run the testsuite with cuda-memcheck, but each time I do it I have to manually skip these tests. This PR skips the problematic tests automatically when run under cuda-memcheck (which sets the `CUDA_MEMCHECK` environment variable).


stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_51`

stuartarchibald:
Buildfarm passed.

sklam:
Needs to resolve conflict against master.

Issue #5545, opened 2020-04-10
Implement math.modf for CUDA target


stuartarchibald:
## Feature request

As title. Request reference:  https://gitter.im/numba/numba?at=5e8f92d83a85536c431d086f

I've put together a starter patch that does the basic implementation here: https://github.com/stuartarchibald/numba/commit/e17796e04ca5d7e207ffe5532ab3f1f60ae31727
needs some tests adding, generally checking over and a doc update, but hopefully the patch is enough to help someone get a PR going.

CC @gabri94

@gmarkall am marking as a "good second issue" for the CUDA target, pinging you so you can mentor :) thanks!

gmarkall:
Thanks for CCing - I'd be keen / happy to mentor! 

@Gabri94 - if you're interested in working on this, I can provide some pointers if you're not sure where to get started - let me know, either here or on gitter.

gabri94:
Sure, i'll ping you on gitter

gmarkall:
@gabri94 Thanks for having a look into this! It looks like @stuartarchibald's commit already adds the implementation, so to complete it ready for a PR, the following would need adding:

- Tests: These should be added to [test_math.py](https://github.com/numba/numba/blob/master/numba/cuda/tests/cudapy/test_math.py) - there are some templates for adding tests for new functions quickly there, but since `modf` is not quite the same as all the other functions (instead of being a unary or binary function that returns a single value, it is a unary function that returns two values), you will have to do something not using those templates. For some appropriate values to test with, you could check out the test of `modf` in CPython: https://github.com/python/cpython/blob/master/Lib/test/test_math.py#L1114 - you should test with both `float32` and `float64` types to ensure the implementation works with both underlying functions. The CUDA programming guide for these functions suggests that the maximum ULP error for these functions is 0, so the output of the CUDA function should probably match the output from CPython (see https://docs.nvidia.com/cuda/libdevice-users-guide/__nv_modf.html#__nv_modf and https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#mathematical-functions-appendix).
- Documentation: there is a list of supported functions in the CUDA target - `modf` should be added to the list here: http://numba.pydata.org/numba-doc/dev/cuda/cudapysupported.html#math - the source for this document is https://github.com/numba/numba/blob/master/docs/source/cuda/cudapysupported.rst

With @stuartarchibald's commit, these two additions (assuming the tests you add pass!) should make the implementation ready for a PR.

[...]

PR #5578, opened 2020-04-17
Implement math.modf for CUDA target 

gabri94:
As title. Request reference: https://gitter.im/numba/numba?at=5e8f92d83a85536c431d086f

After the basic implementation by @stuartarchibald, I implemented the tests for float32 and float64 and edited the docs. 
The tests cover a range of 50 floats between 0 and 10, infinity, minus infinity and nan.

see #5545

### DIFF
--- a/docs/source/cuda/cudapysupported.rst
+++ b/docs/source/cuda/cudapysupported.rst
@@ -149,6 +149,7 @@ The following functions from the :mod:`math` module are supported:
 * :func:`math.floor`
 * :func:`math.copysign`
 * :func:`math.fmod`
+* :func:`math.modf`
 * :func:`math.isnan`
 * :func:`math.isinf`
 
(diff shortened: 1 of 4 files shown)


stuartarchibald:
Many thanks for the PR!

@gmarkall please could you review this? Thanks!

gmarkall:
@gabri94 Looks like your tests have some problems running on the simulator - you can try and reproduce them locally with the environment variable `NUMBA_ENABLE_CUDASIM=1` set.

gabri94:
It was due to the bug I fixed in the last commit. I think it is fine now.

gabri94:
I don't like to have the specific template for `modf` in the generic `TestCudaMath`. I think it should stay inside its scope. We could further abstract it into a generic Unary template with two returns, but if there are no more functions with this template I think it could be over-complex.

In the end I preferred to keep the two templates inside `test_math_modf` and split the test into 6 different subtests.

stuartarchibald:
Patch looks good to me, thanks for doing the work @gabri94, thanks for reviewing/providing guidance @gmarkall. I'll push this through the farm now.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_47`.

stuartarchibald:
Buildfarm passed.

@gabri94 congratulations on your first contribution to Numba!

PR #5579, opened 2020-04-17
Function subtyping

luk-f-a:
First attempt at getting function subtyping, similar to #5560 for records.

Among other things, this allows using first-class function types without declaring the signature of the individual functions, only the signature of the higher-order function. For example,

```python
@njit
def foo1(x):
    return x + 1


@njit
def foo2(x):
    return x + 2


tup = (foo1, foo2)
int_int_fc = types.FunctionType(types.int64(types.int64, ))


@njit(types.int64(types.UniTuple(int_int_fc, 2)))
def bar(fcs):
    x = 0
    for i in range(2):
        x += fcs[i](a)
    return x
```
The functions `foo1` and `foo2` are **not** locked to the `int->int` signature, and could be used later in the code to operate on floats, for example.

For more examples and use cases, please look at https://gist.github.com/luk-f-a/77fb61b25cadbc497337ab9b53b1d0c9


luk-f-a:
@sklam ,  the basic tests are passing, ie a dispatcher can be passed to a function expecting a `FunctionType` if the dispatcher is able to compile under that signature.

However, when testing a tuple of Dispatchers, `test_basic4` in `tests/test_function_type.py` is failing with error 

`TypeError: No matching definition for argument type(s) Tuple(type(CPUDispatcher(<function TestBasicSubtyping.test_basic4.<locals>.foo1 at 0x7fcc9f967488>)), type(CPUDispatcher(<function TestBasicSubtyping.test_basic4.<locals>.foo2 at 0x7fcc9f967620>)))`.

After a lot of `printf` debugging I narrowed it down to a problem of type codes. The tuple of dispatchers from the error message is loaded by `TypeManager.set_compatible` with code 1591. However, `Dispatcher_call` calls `dispatcher_resolve` with `sig[0] =1589`. This mismatch creates the issue. I couldn't find where this mismatch comes from.
Would you have some time to look at it? Or point me in the right direction? I cannot trace where `._code` is coming from. What is `_code` computed on? Python types? LLVM signature?

luk-f-a:
In the latest commit I have added two print statements to highlight the issue. cc: @stuartarchibald 

`Type of tuple in test code Tuple(type(CPUDispatcher(<function TestBasicSubtyping.test_basic4.<locals>.foo1 at 0x7ff5762d1400>)), type(CPUDispatcher(<function TestBasicSubtyping.test_basic4.<locals>.foo2 at 0x7ff5762d1488>)))  code:  1566`

`loading dispatcher->FunctionType in set_compatible. tuple type:  Tuple(type(CPUDispatcher(<function TestBasicSubtyping.test_basic4.<locals>.foo1 at 0x7ff5762d1400>)), type(CPUDispatcher(<function TestBasicSubtyping.test_basic4.<locals>.foo2 at 0x7ff5762d1488>)))  code:  1570`

luk-f-a:
Update: further debugging led me to `_typecache` in `abstract.py`. The type (Tuple(etc, etc)) is being registered, but only a weak reference is held. Somehow the weak reference is getting orphaned (even though the object has not disappeared), leading to the creation of a new type with a new code in https://github.com/numba/numba/blob/1f9d82fc892678242216b4e2e4f724b4ff3db973/numba/core/types/abstract.py#L47.

sklam:
@luk-f-a, your debug code helped a lot in understanding the problem. Typecodes are simply in-process ID to simplified the dispatcher in C. Everytime a unique `Type` instance is created, it is assigned with a fresh typecode, which are never going to repeat in the same process. This interacts with another property of `Type` instances that they are disposed as soon as nothing is retaining a reference to it. The combined effect is that each time compilation is triggered, a fresh argument type for the `Tuple[CPUDispatcher]` is created with a fresh typecode and then discarded when compilation is done. 

I have some hacks to make it work but I will need to find a proper way. My current hack is to recycle typecode for fresh types that have matching `.key`. I also need to review why we stopped making type immortal.

luk-f-a:
@sklam , thanks for looking at it. If it helps, I found where the weakref was introduced: https://github.com/numba/numba/pull/674

luk-f-a:
@sklam I have added a lightweight solution to the typecode instability problem. For function subtyping to work, it's not necessary that typecode are long-term stable, only short-term stable, from the moment the user calls the function until it returns the result.
So TL;DR a full solution to #5608 would solve this problem too, but it's not necessary for subtyping to work. Subtyping only requires stability during the period of time between the function call, through the compilation, until the function return.

At the moment typecodes are unstable even in this short period of time, because when a function is called with tuple of dispatchers for the first time, `Dispatcher_call` ends up (indirectly) creating the type of that tuple twice. 
The first time happens when it calls `typeof_typecode` when trying to get an exact match in https://github.com/numba/numba/blob/e043ee705d65abd3a066500be4d62b54bdf3586c/numba/_dispatcher.c#L535. Tuples of dispatchers are not typed in C, but the fall to the python path in `typeof_pyval`. This happens when a fingerprint cannot be computed in here https://github.com/numba/numba/blob/e043ee705d65abd3a066500be4d62b54bdf3586c/numba/_typeof.c#L535. The call to `typeof_pyval` is intentionally done without caching as explained in the comments. Calling this function results in the types being instantiated and assigned a typecode. But without caching or keeping a reference like it's done in `typecode_fallback_keep_ref`, the created typed will be GC'd and the weak reference in the typecache is lost, losing the typecode.

Without an exact match, new conversions are searched and, when successful (with the help of function subtyping), `compile_and_invoke` is called, which in turn calls `_compile_for_args` (back on pure python side). This function will call `typeof_pyval` again, which will create a second instance of the tuple of dispatchers. Because the weak reference was lost, a new instance, with a new typecode is created. 

To keep the same typecode in the timespan of the function call and compilation, I added `self._types_active_call` as a member of the Dispatcher. After `typeof_pyval` is called the first time, it'll keep a live reference in there, which will ensure that when `typeof_pyval` is called a second time, the typecache will be able to retrieve it and reuse the instance with the old typecode.
`_types_active_call` will be set to `None` before returning the function result, making it a short-live bridge without changing the current memory footprint, ie no additional accumulation of Dispatcher objects in memory.

hgrecco:
I have looked at this PR, tested and it seems to cover all the needs described [here](https://numba.discourse.group/t/avoid-recompilation-of-functions-that-take-other-functions-as-arguments/344/4). 

I would really like to see this as a part of numba soon!. Together with caching dispatchers (https://github.com/numba/numba/pull/6373 and https://github.com/numba/numba/pull/6284) it will remove the performance penalty in libraries like, [numbakit-ode](https://github.com/hgrecco/numbakit-ode), that are generic over user provided functions.

I think that the opt-in API is the way to go. Functions pointers might prevent optimization and therefore should be a choice of the developer.

luk-f-a:
@stuartarchibald, would it be possible to schedule for 0.53? If I could, I would move it to numba-extras, but since it affects type inference it must remain in core numba.

sklam:
I understand the temporary fix with `_type_active_call`. I just need to vet it through to see if there's any unexpected sideeffect.

numba/core/dispatcher.py
```diff
@@ -432,6 +436,10 @@ def error_rewrite(e, issue_type):
             # ignore the FULL_TRACEBACKS config, this needs reporting!
             raise e
 
+        self._types_active_call = None
```
sklam, line 439: Shouldn't this be inside a `finally` so that the reset happens even when an error is raised?

sklam (changes requested):
Thanks for the patch. I've verified that the `_types_active_call` is a safe temporary workaround for the typecode issue. That's one suggestion regarding the resetting of `_types_active_call`.

numba/core/dispatcher.py
```diff
@@ -679,6 +687,7 @@ def typeof_pyval(self, val):
         else:
             if tp is None:
                 tp = types.pyobject
+        self._types_active_call = tp
```
sklam, line 690: I think this will have to be a `self._types_active_call.append(tp)`.

numba/tests/test_function_type.py
```diff
@@ -1123,3 +1123,94 @@ def bar(fcs):
...
+        tup = (foo1, foo2)
+        int_int_fc = types.FunctionType(types.int64(types.int64,))
+
+        @njit(types.int64(types.UniTuple(int_int_fc, 2)))
+        def bar(fcs):
+            x = 0
+            for i in range(2):
+                x += fcs[i](a)
+            return x
+        self.assertEqual(bar(tup), foo1(a) + foo2(a))
```
sklam, line 1216: Suggested test to exercise multiple `FunctionType` argument:


```python

    def test_basic5(self):
        a = 1

        @njit
        def foo1(x):
            return x + 1

        @njit
        def foo2(x):
            return x + 2

        @njit
        def bar1(x):
            return x / 10

        @njit
        def bar2(x):
            return x / 1000

        tup = (foo1, foo2)
        tup_bar = (bar1, bar2)
        int_int_fc = types.FunctionType(types.int64(types.int64,))

        flt_flt_fc = types.FunctionType(types.float64(types.float64,))

        @njit((types.UniTuple(int_int_fc, 2), types.UniTuple(flt_flt_fc, 2)))
        def bar(fcs, ffs):
            x = 0
            for i in range(2):
                x += fcs[i](a)
            for fn in ffs:
                x += fn(a)
            return x

        got = bar(tup, tup_bar)
        expected = foo1(a) + foo2(a) + bar1(a) + bar2(a)
        self.assertEqual(got, expected)
```

sklam (changes requested):
I just found a case that has multiple `FunctionType` arguments and it will fail. `_types_active_call` would need to be a `list` to keep all the argument types alive.

numba/core/dispatcher.py
luk-f-a: you're right, fixed!

numba/core/dispatcher.py
luk-f-a: yes, fixed.

numba/tests/test_function_type.py
luk-f-a: good idea, added!

sklam (approved):
Looks good now!

PR #5591, opened 2020-04-20
Fix #5525 Add figure for total memory to `numba -s` output.

### LINKED ISSUE #5525: Add figure for total memory to `numba -s` output.
## Feature request

It'd be useful to have the amount of memory available on a machine added to the `numba -s` output. This is useful to know for cases where e.g. swapping might be occurring.

This is a good first issue, work involves:

1. Find a way to get the total available memory using Python standard libraries or NumPy.
2. Print the information in 1. in this section: https://github.com/numba/numba/blob/master/numba/misc/numba_entry.py#L34 such that `numba -s` reports the amount of available memory.

jagaudin:
Added memory information
Refactored output into a `dict` and display with template
Added a timer
If conda not present, list packages with pip
Added error and warning logs


stuartarchibald:
Thanks for the PR. Great to see this change being made.

Please could you merge in/rebase on current master to resolve conflicts? Thanks.

jagaudin:
@stuartarchibald All done.

jagaudin:
@stuartarchibald I've rebased and added a few features as discussed: psutil, testing and export as json. I think this is ready for review now.

jagaudin:
Strange the tests were passing on my machine, will have a look.

jagaudin:
Sorry for the remote debugging session. I encountered issues with the use of `setUpClass` in my tests. Strangely enough the tests were passing on my machine before pushing. Will investigate.

stuartarchibald:
@jagaudin Thanks for doing this, the proposed sounds good. I'm not sure the rebase has gone cleanly though because there's a load of other people's commits in the history for this branch. Any chance you could please remove them and try to rebase again, once done I'll mark as ready to review. If you need a hand doing this please shout, thanks!

jagaudin:
@stuartarchibald I've cleaned the unrelated commits and it seems fine now.

esc:
@jagaudin thanks for fixing this up, I have added it to the review queue.

esc:
@jagaudin thanks for submitting this info, I have a question about the figures. I just tried this on my machine and saw:

```
Memory Total (MB)                             : 16384
Memory Available (MB)                         : 485
```

How do I interpret the "Memory Available" value? I.e. what does it mean?

esc:
It might be worth adding some CI logic to test the new functionality with and w/o `psutil`.

jagaudin:
@esc Thanks for the feedback, much appreciated. Sorry I didn't spot that `cpu_affinity` wasn't available on mac. I fixed the script to reflect this and updated the tests.
As it stands, the tests are using a `Mock` object to simulate a few functions of `psutil`. I've fixed this representation so that `psutil.Process()` returns `None` and doesn't have a `cpu_affinity` method unless on Linux and Windows.
Regarding the mem available value, this is the memory that can be given instantly to processes without using swap. But for Mac without `psutil`, this value is the amount of free memory, i.e. not used at all, zeros. The value of available memory in this situation needs to be calculated from other values given by the os. I haven't gone to the extent of doing this but to be more explicit, the script output in this case should now say "Free memory" instead of "Memory available".

esc:
@jagaudin thanks for updating this, I check with and without `psutil` and it seems to work fine. I also tried the `pip` functionality by replacing the `conda` executable with a simple shell script that does `exit 1` and it reports that packages correctly too. It now looks good to me.

esc:
@stuartarchibald did you want to get a second set of eyes on this?

numba/misc/numba_sysinfo.py
```diff
@@ -0,0 +1,639 @@
...
+        'Linux': {
+            'cmd': (
+                CmdBufferOut(('tail', '-vn', '+1',
+                              '/sys/fs/cgroup/cpuacct/cpu.cfs_quota_us')),
+                CmdBufferOut(('tail', '-vn', '+1',
+                              '/sys/fs/cgroup/cpuacct/cpu.cfs_period_us')),
+            ),
+            'cmd_optional': (
+                ('cat', '/proc/meminfo'),
+                ('cat', '/proc/self/status'),
```
stuartarchibald, line 116: Whilst appreciably convenient, I'd rather see the use of direct file reading from Python opposed to spawning processes to run commands in a shell, particularly on OS which use files for config.

numba/tests/test_sysinfo.py
```diff
@@ -0,0 +1,183 @@
...
+        keys = chain(*self.contents.values())
+        for k in keys:
+            with self.subTest(k=k):
+                self.assertIn(k, self.info.keys())
+
+    def test_content_type(self):
+        for t, keys in self.contents.items():
+            for k in keys:
+                with self.subTest(k=k):
+                    self.assertIsInstance(self.info[k], t)
```
stuartarchibald, line 183: Should probably add a test for the `-j` flag to `numba/tests/test_cli.py `

numba/misc/numba_sysinfo.py
jagaudin: Fixed with 833cafc. Was it a security concern?

numba/tests/test_sysinfo.py
jagaudin: There i snow a test, see 75028ba.

numba/misc/numba_sysinfo.py
```diff
@@ -0,0 +1,639 @@
...
+    }
+
+    # CPU features
+    try:
+        feature_map = llvmbind.get_host_cpu_features()
+    except RuntimeError as e:
+        _error_log.append(f'Error (CPU features): {e}')
+    else:
+        features = sorted([key for key, value in feature_map.items() if value])
+        sys_info[_cpu_features] = textwrap.fill(' '.join(features), 80)
```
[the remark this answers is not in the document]
jagaudin: This one is strangely quite difficult. One my machine, I have 3 lines of features (...is it normal?). Anyway, that wouldn't fit on the same format as the other values. How about indenting? Let me know if you have something specific in mind.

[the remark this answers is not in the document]
jagaudin: I also had a second thought about the start time after changing it from utc to local. I thought it would be better but then I saw that the script is running on the CI. Should we have both local and UTC?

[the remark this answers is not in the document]
stuartarchibald: RE times, both local and UTC would probably be good.

RE CPU features, don't worry about it if it's a load of hassle!

numba/misc/numba_sysinfo.py
stuartarchibald: Thanks. My concern was use of subprocess to read a text file, less to go wrong with a direct read.

numba/tests/test_sysinfo.py
stuartarchibald: thanks

stuartarchibald:
Just need to resolve https://github.com/numba/numba/pull/5591#discussion_r427270712 and then I think this is good to go. Thanks again!

esc:
@jagaudin I just had another look at this too and it worked on my machine (OSX) with and w/o `psutil`.

numba/misc/numba_sysinfo.py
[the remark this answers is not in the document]
jagaudin: I've reinstated UTC time in 138c60a and fixed the CPU-features display to match the other lines. It looks ok with a longish output, I thought it would be worse.

numba/tests/test_sysinfo.py
```diff
@@ -0,0 +1,184 @@
...
```
stuartarchibald, line 184: Just spotted that this file is missing:
https://github.com/numba/numba/blob/807689cccaed38272d38df9b0b0cff16c1df59f8/numba/tests/test_alignment.py#L39-L40
at the bottom.

numba/tests/test_sysinfo.py
```diff
@@ -0,0 +1,184 @@
+import platform
```
stuartarchibald, line 1:
```suggestion
import platform
import unittest
```

numba/tests/test_sysinfo.py
stuartarchibald, line 184:
```suggestion
                    self.assertIsInstance(self.info[k], t)


if __name__ == '__main__':
    unittest.main()

```

stuartarchibald:
The updates you did looked good, so thanks for those :) Let's hope CI passes and get this is in!

numba/tests/test_sysinfo.py
jagaudin: Good spot thanks, I added it in e75937f

stuartarchibald: thanks, think you'll need to import `unittest`

numba/tests/test_sysinfo.py
```diff
@@ -0,0 +1,188 @@
...
+                self.assertIn(k, self.info.keys())
+
+    def test_content_type(self):
+        for t, keys in self.contents.items():
+            for k in keys:
+                with self.subTest(k=k):
+                    self.assertIsInstance(self.info[k], t)
+
+ if __name__ == '__main__':
+     unittest.main()
```
stuartarchibald, line 188:
```suggestion
if __name__ == '__main__':
    unittest.main()
```

jagaudin: Sorry it took me 3 commits to sort this, too keen to see this finished!

stuartarchibald: No worries, I think it might still have a leading space on these lines?

jagaudin: Gosh that's terrible, I thought I would do this quickly but it was not a good idea. should be ok now.

stuartarchibald: Don't worry about it, these things happen, thanks for sorting it out. Should CI pass I'll approve it for merge!

stuartarchibald:
As a small request, in future, please could you avoid doing rebases or force-pushes on reviewed branches because GitHub [cannot track these sorts of updates very well](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests#about-pull-requests) which means review comments don't get closed automatically and the reviewer ends up having to re-review the whole diff again to check fixes. Whilst I guess this was just squashing out some history, this PR was approved for merge and by altering history I unfortunately now have to revoke that.

jagaudin:
@stuartarchibald No problems it was a pleasure. I squashed the earlier commits in one to clear the mess. Will try to find some time to work on another issue.

jagaudin:
Sorry I didn't see the approval before force pushing. That was a squash indeed. Will keep that in mind in the future. sorry for the slightly erratic way of doing the last fixes.

esc:
@jagaudin thanks again for submitting this. Looks like it will make it in time for the `0.50.0RC1` release which is coming very soon!

jagaudin:
@esc Thanks for reviewing again. Sorry for the unfortunate squash that caused more work.

PR #5602, opened 2020-04-21
Fix segfault caused by pop from numba.typed.List

### LINKED ISSUE #5264: Heap corruption of Lists of numpy arrays
Lists of numpy arrays (at least of float 1D type) cannot be brought to length zero without heap corruption. Best guess at this point: refcount issue and double free of allocated memory.

Also, Lists of numpy scalars for some reason do not suffer this issue.

Minimal reproducer:

```python
import numpy as np
from numba import  float64
from numba.typed import List

float_array = float64[:]

l = List.empty_list(float_array)
l.append(np.ones(3,dtype=np.float64))
l.pop()
````

### PR DESCRIPTION
**Pull request by esc**
Fixes #5264

### DIFF
--- a/numba/cext/listobject.c
+++ b/numba/cext/listobject.c
@@ -407,12 +407,40 @@ numba_list_resize(NB_List *lp, Py_ssize_t newsize) {
  * */
 int
 numba_list_delitem(NB_List *lp, Py_ssize_t index) {
+    // FIXME the following line is the lazy implementation
+    //result = numba_list_delete_slice(lp, index, index+1, 1);
+    int result;
+    char *loc, *new_loc;
+    Py_ssize_t leftover_bytes;
+    // check for mutability
+    if (!lp->is_mutable) {
+        return LIST_ERR_IMMUTABLE;
+    }
     // check index is valid
     // FIXME: this can be (and probably is) checked at the compiler level
     if (!valid_index(index, lp->size)) {
         return LIST_ERR_INDEX;
     }
-    return numba_list_delete_slice(lp, index, index+1, 1);
+    // obtain item and decref if needed
+    loc = lp->items + lp->item_size * index;
+    list_decref_item(lp, loc);
+    if (index != lp->size - 1) {
+        // delitem from somewhere other than the end, incur the dreaded memory copy
+        leftover_bytes = (lp->size - 1 - index) * lp->item_size;
+        new_loc = lp->items + (lp->item_size * (index + 1));
+        // use memmove instead of memcpy since we may be dealing with
+        // overlapping regions of memory and the behaviour of memcpy is
+        // undefined in such situation (C99).
+        memmove(loc, new_loc, leftover_bytes);
+    }
+    // finally, shrink list by one
+    result = numba_list_resize(lp, lp->size - 1);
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by esc** [comment]
May also fix #4960

File: numba/cext/listobject.c [hunk 1]
```diff
@@ -441,6 +399,50 @@ numba_list_resize(NB_List *lp, Py_ssize_t newsize) {
...
+/* Delete a single item.
+ *
+ * lp: a list
+ * index: the index of the item to delete
+ *
+ * */
+int
+numba_list_delitem(NB_List *lp, Py_ssize_t index) {
+    // FIXME the following line is the lazy implementation
+    //result = numba_list_delete_slice(lp, index, index+1, 1);
```
stuartarchibald, line 411: Remove?

numba/cext/listobject.c
```diff
@@ -441,6 +399,50 @@ numba_list_resize(NB_List *lp, Py_ssize_t newsize) {
...
+    // check index is valid
+    // FIXME: this can be (and probably is) checked at the compiler level
+    if (!valid_index(index, lp->size)) {
+        return LIST_ERR_INDEX;
+    }
+    // obtain item and decref if needed
+    loc = lp->items + lp->item_size * index;
+    list_decref_item(lp, loc);
+    if (index != lp->size - 1) {
+        // delitem from somewhere other than the end, incur the dreaded memory copy
```
stuartarchibald, line 428:
```suggestion
        // delitem from somewhere other than the end, incur the memory copy
```

numba/cext/listobject.c
```diff
@@ -441,6 +399,50 @@ numba_list_resize(NB_List *lp, Py_ssize_t newsize) {
...
 }
 
+/* Delete a single item.
+ *
+ * lp: a list
+ * index: the index of the item to delete
+ *
+ * */
+int
+numba_list_delitem(NB_List *lp, Py_ssize_t index) {
```
stuartarchibald, line 409: The index is restricted to C semantics here, i.e. non-negative? If so perhaps add to docstring.

numba/cext/listobject.c
esc: done

numba/cext/listobject.c
esc: The `index` appears in many of the functions, so for consistency it should probably be set everywhere. How about something like:

```
 * index: the index of the item to delete (must be in range 0 <= index < len(list))
```

numba/cext/listobject.c
esc: done

esc:
@stuartarchibald I have addressed all comments except one and added the reproducers. Let me know how to proceed.

numba/cext/listobject.c
stuartarchibald: Sounds good.

stuartarchibald:
> @stuartarchibald I have addressed all comments except one and added the reproducers. Let me know how to proceed.

Thanks, have responded to the remaining item.

numba/cext/listobject.c
esc: updated!

PR #5606, opened 2020-04-21
Support None as an out argument in ufuncs

eric-wieser:
Still to do: work out how to implement the ufunc, now the the type dispatcher is happy.


eric-wieser:
Tried to rebase this on #5619, but haven't bothered testing it.

PR #5607, opened 2020-04-21
Add StackArray example

guilhermeleobas:
This PR adds an StackArray to the list of example as a complement on how to use some Numba APIs.

### DIFF
--- a/examples/stackarray.py
+++ b/examples/stackarray.py
@@ -1,0 +1,132 @@
+from numba.core.extending import type_callable, \
+    models, register_model, make_attribute_wrapper, lower_builtin, \
+    box, overload
+from numba.core import cgutils
+from numba import types, njit
+import operator
+
+
+# Array
+class StackArray(object):
+
+    def __init__(self, size, ptr):
+        self.size = size
+        self.ptr = ptr
+
+    def __str__(self):
+        return "StackArray(%d)" % self.size
+
+    def __repr__(self):
+        return self.__str__()
+
+
+# Typing layer
+## Creating a new Numba type
+class StackArrayType(types.Type):
+    def __init__(self, dtype):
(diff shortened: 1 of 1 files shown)


examples/stackarray.py
```diff
@@ -0,0 +1,132 @@
...
+@overload(operator.getitem)
+def stackarray_getitem(a, i):
+    if isinstance(a, StackArrayType):
+        def impl(a, i):
+            return a.ptr[i]
+        return impl
+
+
+@overload(len)
+def omnisci_array_len(x):
```
seibert, line 93: Did you mean to name this `stackarray_len`?  :)

rgommers:
Nice @guilhermeleobas. Maybe would be useful to add a docstring with some motivation/explanation? E.g. "if Python overhead of an array is a bottleneck in your code, such as when you have an array of indices that you need to use in a tight loop, then `StackArray` shows a way of pre-allocating an array on the stack to avoid that overhead. ..... `ArrayStack`, `ArrayStackType` and `ArrayStackModel` are <add what they do / how they relate - I don't fully understand that part from a quick read through>)".

esc:
@guilhermeleobas what was the conclusion on this one? Will it be submitted to `numba-examples` instead?

stuartarchibald:
> @guilhermeleobas what was the conclusion on this one? Will it be submitted to `numba-examples` instead?

From core dev meeting, yes. `numba-examples` needs a `mkdir` doing for `user_contributed` or similar as a location to put all these.

guilhermeleobas:
Yeah, I will open a new PR in the `numba-examples` with this.

guilhermeleobas:
Is the `numba-examples` more related to performance? In the PR, should I adapt the example to look more like a benchmark?

Issue #5218, opened 2020-02-08
Use __init__ type hints for jitclass spec?


StanczakDominik:
I really like how jitclass is coming along! I think I've got something that could clean its usage up a little and might not have been considered.

Here's the current usage, with the `spec` argument:

```python
@numba.jitclass({"numbers": numba.int64[:]})
class NumbaArraySummer:
    def __init__(self, numbers):
        self.numbers = numbers
        
    def sum(self):
        return np.sum(self.numbers)
```

But I think another possibility, cleaner and with less typing, would be to use type hints as follows:

```python
@numba.jitclass
class NumbaArraySummerAnnotated:
    def __init__(self, numbers: numba.int64[:]):
        self.numbers = numbers
        
    def sum(self):
        return np.sum(self.numbers)
```

In fact, looking at the implementation of `jitclass`, I think I may be able to contribute this as a PR. Would this be of interest?

StanczakDominik:
... I just looked on master and `jitclass` is gone.

Huh.

Feel free to close this, I guess, since this no longer seems necessary...? :sweat_smile:

stuartarchibald:
> ... I just looked on master and `jitclass` is gone.
> 
> Huh.
> 
> Feel free to close this, I guess, since this no longer seems necessary...? sweat_smile
> 
> Oh right, it's in experimental now!

Yes, note in the docs `This is a early version of jitclass support. Not all compiling features are exposed or implemented, yet.` http://numba.pydata.org/numba-doc/latest/user/jitclass.html and it turns out there's numerous issues with it that need considerable design/thinking time. Hence whilst refactoring the opportunity was taken to put it into `experimental`.

StanczakDominik:
Right, thanks for the heads up. Two questions:

1. Is there some way to follow the rethinking?
2. Do you think this kind of thing is better abandoned for now, until the re-pondering is pondered out, or should I try my hand at a proposal PR for this idea?

luk-f-a:
@StanczakDominik , how would you deal with instance members which are not inputs to `__init__`? For example `value` below:

```python
@numba.jitclass({"numbers": numba.int64[:], "value": numba.float64})
class NumbaArraySummer:
    def __init__(self, numbers):
        self.numbers = numbers
        
    def sum(self):
        return np.sum(self.numbers)

    def set_value(self):
        self.value = self.sum()   
```

StanczakDominik:
With great difficulty.

[...]

PR #5609, opened 2020-04-22
Jitclass Infer Spec from Type Annotations

EPronovost:
Extend a jitclass spec from type annotations on the class.
To enable this, we cleaned up `typeof` and extended it to support `typing` generics.  We also made the `jitclass` function more flexible for various use cases.  We can now do things like

```
@jitclass
class Counter:
    x: int
    
    def __init__(self):
        self.x = 0

    @property
    def count(self) -> int:
        return self.x

    def inc(self) -> int:
        self.x += 1
        return self.x
```

Related to issues https://github.com/numba/numba/issues/5218 and https://github.com/numba/numba/issues/2947.


examples/jitclass.py
```diff
@@ -5,36 +5,38 @@
...
 import numpy as np
-from numba import int32, float32    # import the types
+from numba import float32
 from numba.experimental import jitclass
 
-spec = [
-    ('value', int32),               # a simple scalar field
-    ('array', float32[:]),          # an array field
-]
-
```
StanczakDominik, line 16: I wonder if we shouldn't have both ways of doing things in this example, with type annotations and the old style spec? I do like the new style, though, so if the old one were to be phased out, I would shed no tears.

EPronovost:
Hi @stuartarchibald , I saw this got moved to milestone 0.51.  Is there anything I can do, or will we just need to wait a few more months on this?  I know from the meeting minutes a while ago there was interest in this PR and a note that it may require discussion.  When should we expect that discussion to happen?

stuartarchibald:
> Hi @stuartarchibald , I saw this got moved to milestone 0.51.

Yes, sorry, the view was that this is sufficiently involved and hasn't been discussed/explored enough.

> Is there anything I can do, or will we just need to wait a few more months on this?

For 0.50 we're trialing this: https://groups.google.com/a/continuum.io/forum/#!topic/numba-users/DUJhuX8uD_I helping out would free up more time to deal with involved features, though I obviously can't guarantee anything!

> I know from the meeting minutes a while ago there was interest in this PR and a note that it may require discussion. When should we expect that discussion to happen?

All PRs are discussed first in weekly triage, and then those which are proposing significant change are discussed at core developer meetings. I can propose this for discussion again at the next meeting to prompt others to form views in the next few weeks. Discussion of new features/changes often takes multiple iterations and unfortunately there's a bit of a backlog, for example the proposed https://github.com/numba/numba/pull/5560 and https://github.com/numba/numba/pull/5579 have been through multiple iterations of discussion.

numba/tests/test_typeof.py
```diff
@@ -301,6 +308,77 @@ def test_omitted_args(self):
         self.assertEqual(len({ty0, ty1, ty2}), 3)
         self.assertEqual(ty3, ty2)
 
+    def test_builtin_types(self):
+        self.assertEqual(typeof(int), typeof(0))
+        self.assertEqual(typeof(float), typeof(0.0))
+        self.assertEqual(typeof(complex), typeof(complex(0)))
+        self.assertEqual(typeof(str), typeof("numba"))
```
eric-wieser, line 315: It's not clear to me that `typeof(x) == typeof(type(x))` makes any sense.
I'd expect these to return `numba.types.TypeRef`, if not the `ValueError` they give today.

Would it make sense to introduce a new `as_numba_type` function?

docs/source/user/jitclass.rst
```diff
@@ -61,6 +61,26 @@ initializing each defined fields.  Uninitialized fields contains garbage data.
...
+        z: T
+
+        def __init__(self, x, y, z):
+            ...
+
+then the full spec used for ``Foo`` will be:
+
+* ``"x": int32`` (specified in the original `spec` passed to `jitclass`)
+* ``"y": float64`` (added from type annotation)
+* ``"z": numba.typeof(T)`` (added from type annotation)
```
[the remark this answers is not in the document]
EPronovost: I like that better as well.  Done.

examples/jitclass.py
EPronovost: Done.

numba/tests/test_typeof.py
eric-wieser: In fact, this change is incompatible with the behavior today:
```
In [22]: numba.typeof(np.uint8(1))
Out[22]: uint8

In [23]: numba.typeof(type(np.uint8(1)))
Out[23]: class(uint8)
```

EPronovost: I had a reason for making it part of `typeof` originally, but can't remember it.  Making a new `as_numba_type` gives a cleaner API and offers some cool opportunities for user extensions.

EPronovost:
Per @eric-wieser 's suggestion, I moved the type inference logic to a new `as_numba_type`.  This allowed me to make a new registry, which has interesting opportunities for user extensions (e.g. for other `jitclass`es or to modify the default behavior for `float`).  Currently there's some similar code between `typeof` and `as_numba_type`.  We could possibly do something like 
```
ty = as_numba_type(type(val))
if ty is not None:
    return ty
```
to replace all of the `typeof_impl` implementations of the form
```
@typeof_impl.register(py_type)
def impl(_, _):
    return nb_type
```

esc:
@EPronovost thanks for updating this, I have labeled it as "waiting on reviewer input now".

docs/source/extending/interval-example.rst
```diff
@@ -75,6 +75,19 @@ Function arguments and global values will thusly be recognized as belonging
...
+-------------------------------------
+
+While ``typeof`` is used to infer the Numba type of Python objects,
+``as_numba_type`` is used to infer the Numba type of Python types.  For simple
+cases, we can simply register that the Python type ``Interval`` corresponds with
+the Numba type ``interval_type``::
+
+   from numba.extending import as_numba_type
+
+   as_numba_type.register(Interval, interval_type)
```
eric-wieser, line 88: This would be simpler as `Interval._numba_type_ = interval_type`, which already informs `typeof` today.

However, this is only really possible in contrived cases. In all the extension types I've used so far, I've needed something like `IntervalType(self.dtype)` instead of just a global `IntervalType()`.

The mapping between python types and numba types is frequently one-to-many - so while `as_numba_type` is useful for the trivial cases you have being used here, it doesn't scale to all use-cases.

EPronovost: I agree that the current way Numba requires explicit types up-front does not address all possible use cases in Python.  The `as_numba_type` registry is meant to work statically on type annotations at compile type, not dynamically on instances at runtime.  I believe the combination of a [generic type annotation](https://docs.python.org/3/library/typing.html#generics) and the `@as_numba_type.register` decorator does allow for something like the `LoggedVar` example at the link above.  That being said, I agree that there's still a lot of room to improve the flexibility of `jitclass`.  There are several open issues about so-called "generic" `jitclass`es.  That was something I was interested in once this PR is concluded.

EPronovost:
Cross-posting this follow-on PR that uses the `as_numba_type` registry to implement a typed version of named tuples. https://github.com/numba/numba/pull/5840

esc:
hey, everyone, just checking on the status of this one as it is part of the 0.51.0 milestone. Where are we at?

EPronovost:
@esc I'm not clear what the review process needs to include.  Is there something more I need to do?

esc:
@EPronovost I just had another look, I believe this is awaiting a review from the core-dev team.

numba/experimental/jitclass/decorators.py
```diff
@@ -4,30 +4,83 @@
...
-        With a dictionary, use collections.OrderedDict for stable ordering.
-        With a sequence, it must contain 2-tuples of (fieldname, fieldtype).
+    If specified, ``spec`` gives the types of class fields.
+    It must be a dictionary or sequence.
+    With a dictionary, use collections.OrderedDict for stable ordering.
+    With a sequence, it must contain 2-tuples of (fieldname, fieldtype).
 
-    **returns**:
+    Any class annotations for field names not listed in spec will be added.
+    For class annotation `x: T` we will append `("x", numba.typeof(T))` to
```
sklam, line 20: isn't it `as_numba_type(T)` not `typeof(T)`?

numba/experimental/jitclass/base.py
```diff
@@ -167,8 +167,17 @@ def register_class_type(cls, spec, class_ctor, builder):
...
     # Normalize spec
-    if isinstance(spec, Sequence):
+    if spec is None:
+        spec = OrderedDict()
+    elif isinstance(spec, Sequence):
         spec = OrderedDict(spec)
+
+    # Extend spec with class annotations.
+    if hasattr(cls, "__annotations__"):
+        for attr, py_type in cls.__annotations__.items():
```
sklam, line 177: Should we consider all `__annotations__` from the `__mro__`?

Consider this:

```python

In [8]: class A:
   ...:     x : int
   ...:

In [9]: A.__annotations__
Out[9]: {'x': int}

In [10]: class B(A):
    ...:     pass
    ...:

In [11]: B.__annotations__   # B didn't add new __annotations__ and carries A.__annotations__
Out[11]: {'x': int}

In [12]: class C(A):
    ...:     y : int
    ...:

In [13]: C.__annotations__   # C has new __annotations__ which should be consider incremental
Out[13]: {'y': int}
```

docs/source/user/jitclass.rst
```diff
@@ -40,11 +40,49 @@ initializing each defined fields.  Uninitialized fields contains garbage data.
 Methods and properties (getters and setters only) can be defined.  They will be
 automatically compiled.
 
-Specifying ``numba.typed`` containers as class members
-======================================================
-It is often desirable to use a ``numba.typed.Dict`` or a ``numba.typed.List`` as
-a class member in a ``jitclass``. Methods for using these types and various
-common patterns are presented in the following:
+
+Inferred class member types with ``as_numba_type``
```
sklam, line 44: The heading should also mention "type annotations", which is a more familiar term to look for.

docs/source/user/jitclass.rst
```diff
@@ -40,11 +40,49 @@ initializing each defined fields.  Uninitialized fields contains garbage data.
...
+
+.. literalinclude:: ../../../numba/tests/doc_examples/test_jitclass.py
+   :language: python
+   :start-after: magictoken.ex_jitclass_type_hints.begin
+   :end-before: magictoken.ex_jitclass_type_hints.end
+   :dedent: 8
+
+Any type annotations on the class will be used to extend the spec if that field
+is not already present.  The Numba type corresponding to given Python type is
+inferred using ``as_numba_type``.  For example, if we have the class
```
sklam, line 57: Perhaps mention that type annotations on method arguments and return type are ignored since they are used in the example.

docs/source/extending/interval-example.rst
sklam: The comment:

> The as_numba_type registry is meant to work statically on type annotations at compile type, not dynamically on instances at runtime. 

needs to be documented.

docs/source/user/jitclass.rst
```diff
@@ -40,11 +40,49 @@ initializing each defined fields.  Uninitialized fields contains garbage data.
...
+        x: int
+        y: float
+        z: SomeOtherType
+
+        def __init__(self, x: int, y: float, z: SomeOtherType):
+            ...
+
+then the full spec used for ``Foo`` will be:
+
+* ``"x": int32`` (specified in the original ``spec`` passed to ``jitclass``)
```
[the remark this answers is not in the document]
EPronovost: > it's possible to make having type annotations bans the use of the spec

I'm not sure if this would work for numpy types.  Any array would get a type annotation `np.ndarray`, but numba requires more information.  In a future where numpy has generic type annotations for array this could be solved, but I don't think we're there yet.  This situation is why I thought to allow overwriting annotations with the `spec`.

If you remove the `@jitclass` decorator from the above example, that's still not valid Python code (also the type annotations for `Counter` seem incorrect).  What are you suggesting to change because of this example?  Add a check that every field is initialized in `__init__`?

numba/__init__.py
```diff
@@ -18,7 +18,7 @@
 # Re-export typeof
 from numba.misc.special import (
     typeof, prange, pndindex, gdb, gdb_breakpoint, gdb_init,
-    literally, literal_unroll
+    literally, literal_unroll, as_numba_type,
```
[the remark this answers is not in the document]
EPronovost: Good point.  If we move it to `numba.experimental`, could we also just not create a deprecated endpoint with a warning in `numba/__init__.py`?

docs/source/user/jitclass.rst
EPronovost: Good call.  Done.

docs/source/user/jitclass.rst
EPronovost: Done.

numba/experimental/jitclass/base.py
EPronovost: Yup! TIL that `__annotations__` doesn't include fields from parent classes.  Updated and added a test for this.

numba/experimental/jitclass/decorators.py
EPronovost: Yup. Updated.

numba/__init__.py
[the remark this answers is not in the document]
EPronovost: Removed from the top-level namespace.

docs/source/extending/interval-example.rst
```diff
@@ -75,6 +75,19 @@ Function arguments and global values will thusly be recognized as belonging
 to ``interval_type`` whenever they are instances of ``Interval``.
 
 
+Type inference for Python annotations
+-------------------------------------
+
+While ``typeof`` is used to infer the Numba type of Python objects,
+``as_numba_type`` is used to infer the Numba type of Python types.  For simple
+cases, we can simply register that the Python type ``Interval`` corresponds with
+the Numba type ``interval_type``::
```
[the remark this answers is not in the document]
EPronovost: For sure.  I added a section to the `types` doc as well.

docs/source/extending/interval-example.rst
EPronovost: Done.

EPronovost:
@sklam thanks for the review! I've updated the PR.

docs/source/user/jitclass.rst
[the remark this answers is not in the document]
sklam: You convinced me. We need to have both spec and annotations. 

> Add a check that every field is initialized in __init__?

We'll need to address the segfault in a different PR. Prevent the segfault or check that all fields are initialized, or both.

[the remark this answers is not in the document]
EPronovost: Ok so is there any action item for this comment thread (beyond the documentation updates)?

docs/source/reference/types.rst
```diff
@@ -329,3 +329,44 @@ Optional types
...
+      >>> numba.extending.as_numba_type(typing.List[float])
+      ListType[float64]
+      >>> numba.extending.as_numba_type(numba.int32)
+      int32
+
+   ``as_numba_type`` is automatically updated to include any ``@jitclass``.
+
+      >>> @jitclass
+      ... class Counter:
+      ...     x: int
```
eric-wieser, line 359: Perhaps worth allowing (or requiring) the spelling
```suggestion
      ...     x: typing.Annotated[int, numba.int64]
```

docs/source/user/jitclass.rst
```diff
@@ -40,11 +40,59 @@ initializing each defined fields.  Uninitialized fields contains garbage data.
...
+is not already present.  The Numba type corresponding to the given Python type
+is inferred using ``as_numba_type``.  For example, if we have the class
+
+.. code-block:: python
+
+    @jitclass([("w", int32), ("y", float64[:])])
+    class Foo:
+        w: int
+        x: float
+        y: np.ndarray
```
eric-wieser, line 65: Again, this could be
```suggestion
        y: Annotated[np.ndarray, float64[:]]
```

docs/source/reference/types.rst
EPronovost: I didn't realize PEP 593 had been back-ported in `typing_extensions` already.  Awesome!  Added support for it.  See the updated docs and examples.

docs/source/user/jitclass.rst
EPronovost: Yup! See comment above

docs/source/reference/types.rst
```diff
@@ -366,7 +369,30 @@ Type annotations
...
 
+   For cases where Numba requires more type information than what Python type
+   annotations support (e.g. Numpy arrays), or where a different Numba type is
+   desired for the same Python type (e.g. int32 vs int64),
+   ``typing_extensions.Annotated`` can be used.
+   Valid patterns are:
+
+      * ``Annotated[py_type, nb_type]``: ``as_numba_type`` will return ``as_numba_type(nb_type)``
+      * ``Annotated[np.ndarray, dtype, ndim]``: ``as_numba_type`` will return ``Array(dtype=as_numba_type(dtype), ndim=ndim, layout="A")``
+      * ``Annotated[np.ndarray, dtype, ndim, layout]``: ``as_numba_type`` will return ``Array(dtype=as_numba_type(dtype), ndim=ndim, layout=layout)``
```
eric-wieser, line 383: I don't think this is a good idea - why not just require the user to use the normal syntax for array types rather than inventing a new one?

EPronovost:
@eric-wieser suggested supporting `typing_extensions.Annotated`, which I think is a great idea (I didn't realize it was in `typing_extensions` already).  This requires adding the `typing_extensions` package as a dependency, which I've already done in https://github.com/numba/numba/pull/6222.  My hope would be to merge that PR (which should hopefully be quick to review) and then update and merge this one.  I tested locally under Python 3.6, 7, and 8 (since a lot of `typing` changed in 3.7).

docs/source/reference/types.rst
EPronovost: With these options, the user doesn't need to use any numba types for some cases.  For instance you can do
```
Annotated[np.ndarray, float, 2]
```
I think this lowers the barrier for entry with new users... they don't even have to read anything about numba types and what `__getitem__` does.  By just using `Annotated[np.ndarray, {float, int, complex, bool}, N]` they can get pretty far.

numba/core/typing/asnumbatype.py
```diff
@@ -0,0 +1,193 @@
...
+
+        if annotated_type_check(py_type):
+            inner_py_type = py_type.__args__[0]
+            annotations = _py_version_switch(
+                (3, 7),
+                lambda: py_type.__metadata__,
+                lambda: py_type.__args__[1]
+            )()
+
+            if len(annotations) == 1:
```
eric-wieser, line 80: The spec for annotated says you should ignore any arguments you don't recognize, as they may be for other tools. So I'd recommend you filter the annotations to those where `ininstance(arg, numba.Type)` is true, and assert that is of length 1.

EPronovost: The [PEP](https://www.python.org/dev/peps/pep-0593/#consuming-annotations) also says
> Multiple annotations: It's up to the tool consuming the annotations to decide whether the client is allowed to have several annotations on one type and how to merge those annotations.

Checking `isinstance(arg, numba.Type)` would disallow the current recursive call where `Annotated[py_type, x] => as_numba_type(x)`.

I'll leave it up to the core devs.  Either we only support explicit numba types, in which case we can do such a check, or we also support numpy and builtin types (e.g. `Annotated[np.ndarray, int, 1]`), in which case I think it's simpler to say that our tool does not support multiple annotations. 

FWIW I wouldn't expect numba to "just work" with other python tools, so supporting their annotations may be misleading.  I don't have much experience with what other tools would use `Annotated` for.

docs/source/reference/types.rst
eric-wieser: I don't think it's numba's place to advocate for a way to statically type numpy arrays. There are ongoing discussions in numpy about how to annotate arrays with shape and dtype information, inventing a new and different protocol to whatever is standardised there will IMO just cause confusion. `Annotated[np.ndarray, float, 2]` is barely shorter than `Annotated[np.ndarray, numba.Array(float, 2)]`, and the later makes it very clear to the reader that the syntax is numba-specific.

sklam: I'd also prefer to not invent new ways to annotate ndarrays. `Annotated[np.ndarray, numba.Array(float, 2)]` looks more understandable than `Annotated[np.ndarray, float, 2]`.

numba/core/typing/asnumbatype.py
sklam: I don't think I am familiar with the type annotation standard or convention to make the call.

sklam:
@EPronovost, we are near our 0.52.0rc due date. Besides the merge conflict, https://github.com/numba/numba/pull/5609/commits/7a8c080a715e62333922c3b966da8c0649a8c57e seems to be controversial and I am not well-versed in type annotation to make any decision. At the same time, I don't want this PR to drag out too long. Is it possible to revert that commit and merge what we have knowing that this feature (along with jitclass) is experimental.

sklam (approved):
Thanks for the contribution. It's looking good.

PR #5616, opened 2020-04-23
Simplify the ufunc kernel registration

eric-wieser:
`register_unary_ufunc_kernel` and `register_binary_ufunc_kernel` were trivial to merge, and as a result trinary ufuncs (like np.core.multiarray.clip) are now possible to support.

### DIFF
--- a/numba/np/npyimpl.py
+++ b/numba/np/npyimpl.py
@@ -441,38 +441,21 @@ def _ufunc_db_function(ufunc):
 
 _kernels = {} # Temporary map from ufunc's to their kernel implementation class
 
-def register_unary_ufunc_kernel(ufunc, kernel):
-    def unary_ufunc(context, builder, sig, args):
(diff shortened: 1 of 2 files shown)


stuartarchibald:
Thanks for the patch, this looks good. Would be nice to see the noted `np.core.multiarray.clip` implemented as a demonstration of the improved mechanism, but not essential.

eric-wieser:
> Would be nice to see the noted

I didn't want to deal with the can of worms that is `np.clip` vs `np.core.multiarray.clip`, so decided not to go that far. Hopefully this machinery will be exercised more in #5619

stuartarchibald:
> > Would be nice to see the noted
> 
> I didn't want to deal with the can of worms that is `np.clip` vs `np.core.multiarray.clip`, so decided not to go that far. Hopefully this machinery will be exercised more in #5619

This is a fair point :)

PR #5619, opened 2020-04-23
Add support for multi-output ufuncs

eric-wieser:
This adds a simple implementation of `divmod` for the integer types to prove the multi-output feature works.

Adding support for remaining multi-output ufuncs is out of scope for this patch.

### DIFF
--- a/docs/source/reference/numpysupported.rst
+++ b/docs/source/reference/numpysupported.rst
@@ -608,6 +608,7 @@ Math operations
  remainder           Yes          Yes
  mod                 Yes          Yes
  fmod                Yes          Yes
+ divmod (*)          ???          Yes
  abs                 Yes          Yes
  absolute            Yes          Yes
  fabs                Yes          Yes
@@ -629,6 +630,7 @@ Math operations
  lcm                 Yes          Yes
 ==============  =============  ===============
 
+(\*) not supported on timedelta types
 
 Trigonometric functions
 -----------------------
--- a/docs/source/reference/numpysupported.rst
+++ b/docs/source/reference/numpysupported.rst
@@ -608,7 +608,7 @@ Math operations
  remainder           Yes          Yes
  mod                 Yes          Yes
  fmod                Yes          Yes
- divmod (*)          ???          Yes
+ divmod (*)          Yes          Yes
  abs                 Yes          Yes
  absolute            Yes          Yes
  fabs                Yes          Yes
--- a/numba/np/npyimpl.py
+++ b/numba/np/npyimpl.py
@@ -479,7 +479,7 @@ def register_ufunc_kernel(ufunc, kernel):
     _any = types.Any
     in_args = (_any,) * ufunc.nin
 
-    # Add an overload for each out argument being missing.
+    # Add a lowering for each out argument that is missing.
     for n_explicit_out in range(ufunc.nout + 1):
         out_args = (types.Array,) * n_explicit_out
         lower(ufunc, *in_args, *out_args)(do_ufunc)
--- a/numba/np/npyimpl.py
+++ b/numba/np/npyimpl.py
@@ -16,7 +16,9 @@ import operator
 from numba.np import arrayobj, ufunc_db, numpy_support
 from numba.core.imputils import Registry, impl_ret_new_ref, force_error_model
 from numba.core import typing, types, utils, cgutils, callconv
-from numba.np.numpy_support import ufunc_find_matching_loop, select_array_wrapper, from_dtype, _ufunc_loop_sig
+from numba.np.numpy_support import (
+    ufunc_find_matching_loop, select_array_wrapper, from_dtype, _ufunc_loop_sig
+)
 from numba.core.typing import npydecl
 from numba.core.extending import overload, intrinsic
 
@@ -374,7 +376,7 @@ def numpy_ufunc_kernel(context, builder, sig, args, ufunc, kernel_class):
 
     indices = [inp.create_iter_indices() for inp in inputs]
 
-    # assume outputs are all the same size
+    # assume outputs are all the same size, which numpy requires
 
     loopshape = outputs[0].shape
     with cgutils.loop_nest(builder, loopshape, intp=intpty) as loop_indices:
--- a/numba/np/npyimpl.py
+++ b/numba/np/npyimpl.py
@@ -16,7 +16,7 @@ import operator
 from numba.np import arrayobj, ufunc_db, numpy_support
 from numba.core.imputils import Registry, impl_ret_new_ref, force_error_model
 from numba.core import typing, types, utils, cgutils, callconv
-from numba.np.numpy_support import ufunc_find_matching_loop, select_array_wrapper, from_dtype
+from numba.np.numpy_support import ufunc_find_matching_loop, select_array_wrapper, from_dtype, _ufunc_loop_sig
 from numba.core.typing import npydecl
 from numba.core.extending import overload, intrinsic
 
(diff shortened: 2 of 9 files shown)


numba/tests/test_ufuncs.py
```diff
@@ -1613,6 +1614,8 @@ class TestLoopTypesDatetimeNoPython(_LoopTypesTester):
     _compile_flags = no_pyobj_flags
     _ufuncs = supported_ufuncs[:]
 
+    _ufuncs.remove(np.divmod)  # not implemented yet
```
eric-wieser, line 1617: I didn't fancy dealing with this, the main purpose of the patch was to make it possible to write multi-output ufuncs, not to fully implement all of them.

numba/np/ufunc/array_exprs.py
```diff
@@ -392,6 +393,10 @@ def generate(self, *args):
...
                              self.outer_sig.return_type)
 
+    # create a fake ufunc object which is enough to trick numpy_ufunc_kernel
+    ufunc = SimpleNamespace(nin=len(expr_args), nout=1, __name__=expr_name)
+    ufunc.nargs = ufunc.nin + ufunc.nout
+
     args = [lowerer.loadvar(name) for name in expr_args]
     return npyimpl.numpy_ufunc_kernel(
-        context, builder, outer_sig, args, ExprKernel, explicit_output=False)
+        context, builder, outer_sig, args, ufunc, ExprKernel)
```
eric-wieser, line 402: Mildly hacky, but easier to reason about than trying to infer `ufunc.nin` and `ufunc.nout` from `sig`, `args`, and `explicit_outputs`.

eric-wieser:
I will rebase this one #5783 and #5862 are merged.

eric-wieser:
Those PRs are merged, this is rebased, and CI is passing - this should be ready for review

stuartarchibald:
Thanks @eric-wieser

numba/np/npyimpl.py
```diff
@@ -315,47 +336,58 @@ def numpy_ufunc_kernel(context, builder, sig, args, kernel_class,
...
+        [a.base_type for a in outputs],
+        [a.base_type for a in inputs]
+    )
     kernel = kernel_class(context, builder, outer_sig)
     intpty = context.get_value_type(types.intp)
 
     indices = [inp.create_iter_indices() for inp in inputs]
 
-    loopshape = output.shape
+    # assume outputs are all the same size
```
stuartarchibald, line 377: Perhaps add something to catch if they aren't?

numba/np/npyimpl.py
```diff
@@ -443,45 +474,38 @@ def generate(self, *args):
...
+        return numpy_ufunc_kernel(context, builder, sig, args, ufunc, kernel)
 
     _any = types.Any
     in_args = (_any,) * ufunc.nin
 
-    # (array or scalar, out=array)
-    lower(ufunc, *in_args, types.Array)(do_ufunc)
-    # (array or scalar)
-    lower(ufunc, *in_args)(do_ufunc_no_explicit_output)
+    # Add an overload for each out argument being missing.
```
stuartarchibald, line 482:
```suggestion
    # Add a lowering for each out argument that is missing.
```

numba/np/npyimpl.py
```diff
@@ -16,7 +16,7 @@
 from numba.np import arrayobj, ufunc_db, numpy_support
 from numba.core.imputils import Registry, impl_ret_new_ref, force_error_model
 from numba.core import typing, types, utils, cgutils, callconv
-from numba.np.numpy_support import ufunc_find_matching_loop, select_array_wrapper, from_dtype
+from numba.np.numpy_support import ufunc_find_matching_loop, select_array_wrapper, from_dtype, _ufunc_loop_sig
```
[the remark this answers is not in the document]
eric-wieser: As follows?
```suggestion
from numba.np.numpy_support import (
    ufunc_find_matching_loop, select_array_wrapper, from_dtype, _ufunc_loop_sig
)
```

docs/source/reference/numpysupported.rst
```diff
@@ -608,6 +608,7 @@ Math operations
  remainder           Yes          Yes
  mod                 Yes          Yes
  fmod                Yes          Yes
+ divmod (*)          ???          Yes
```
eric-wieser, line 611: Is `object mode` automatic?

numba/np/npyimpl.py
eric-wieser: Numba is already broken when it comes to checking the shape of output arrays, so I'd prefer to declare this out of scope. Without this PR:
```
In [11]: @numba.njit
    ...: def neg(a, b):
    ...:     return np.negative(a, b)
    ...:

In [16]: neg(np.ones(2, int), np.zeros((2, 4), int))
Out[16]:
array([[         -1,          -1, -1409738722,  2013264680],
       [         -1,          -1, -1409738722,  2013264680]])
```

eric-wieser: At any rate, numpy does not allow them to be different sizes.

numba/np/npyimpl.py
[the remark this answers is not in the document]
stuartarchibald: That looks good thanks, essentially anything flake8 is happy with.

numba/np/npyimpl.py
stuartarchibald: ok, fair point, something to address another time.

docs/source/reference/numpysupported.rst
stuartarchibald: it should be!

stuartarchibald:
Just gave this a quick test, this:
```python
from numba import njit
import numpy as np

@njit(error_model='numpy')
def foo():
    return np.divmod(np.arange(0, 100., 10) + 1, np.arange(10.))

print(foo())
print(foo.py_func())
```
produces:
```
$ python test_5619.py 
(array([inf, 11., 10., 10., 10., 10., 10., 10., 10., 10.]), array([nan,  0.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.]))
test_5619.py:6: RuntimeWarning: invalid value encountered in divmod
  return np.divmod(np.arange(0, 100., 10) + 1, np.arange(10.))
(array([nan, 11., 10., 10., 10., 10., 10., 10., 10., 10.]), array([nan,  0.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.]))
```
note inf vs. nan. Not sure that this is something that should be fixed here.

eric-wieser:
Seems out of scope to me, that's a bug in `floor_divide`:
```
In [2]: from numba import njit
   ...: import numpy as np
   ...:
   ...: @njit(error_model='numpy')
   ...: def foo():
   ...:     return np.floor_divide(np.arange(0, 100., 10) + 1, np.arange(10.))
   ...:
   ...: print(foo())
   ...: print(foo.py_func())
[inf 11. 10. 10. 10. 10. 10. 10. 10. 10.]
<ipython-input-2-1151708d7a3f>:6: RuntimeWarning: invalid value encountered in floor_divide
  return np.floor_divide(np.arange(0, 100., 10) + 1, np.arange(10.))
[nan 11. 10. 10. 10. 10. 10. 10. 10. 10.]
```

docs/source/reference/numpysupported.rst
eric-wieser: So I can, without any extra testing, set this to
```suggestion
 divmod (*)          Yes          Yes
```

stuartarchibald: Yes, IIRC if it's working in `nopython` mode it'll work in `objmode`. Think the restrictions in this list are because of multiple return values not being supported until now (e.g. `frexp`, `modf`).

stuartarchibald:
> Seems out of scope to me, that's a bug in `floor_divide`:

Agree. #5985 tracks.

numba/np/npyimpl.py
eric-wieser, line 377:
```suggestion
    # assume outputs are all the same size, which numpy requires
```

PR #5624, opened 2020-04-24
Fix erroneous rewrite of predicate to bit const on prune.

### LINKED ISSUE #5618: Jit produces wrong code with version 0.49.0
[Our library's](https://github.com/pymatting/pymatting/issues/15) tests [fail on all platforms](https://travis-ci.org/github/pymatting/pymatting/builds/678580669) when upgrading from numba version `0.48.0` to `0.49.0`.

The smallest test case that I could come up with that still reproduces the issue is the following:

```Python
import numpy as np
from numba import jit

def foo():
    values = np.zeros(1)
    tmp = 666
    if tmp:
        values[0] = tmp
        print(values)

print("Without jit")
foo()
print("With jit")
jit(foo)()
```

It prints `666.0` with Python and `1.0` when jitted.

- [ x ] I am using the latest released version of Numba (most recent is visible in
- [ x ] I have included below a minimal working reproducer (if you are unsure how

stuartarchibald:
As title.

Fixes #5618

### DIFF
--- a/numba/core/analysis.py
+++ b/numba/core/analysis.py
@@ -283,6 +283,10 @@ def _fix_loop_exit(cfg, loop):
         return loop
 
 
+# Used to describe a nullified condition in dead branch pruning
+nullified = namedtuple('nullified', 'condition, taken_br, rewrite_stmt')
+
+
 # Functions to manipulate IR
 def dead_branch_prune(func_ir, called_args):
     """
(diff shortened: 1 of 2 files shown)


numba/core/analysis.py
```diff
@@ -397,6 +397,7 @@ def resolve_input_arg_const(input_arg_idx):
     # at least one an arg of the condition is a const
     # if the condition is met it will replace the branch with a jump
     branch_info = find_branches(func_ir)
+    nullified = namedtuple('nullified', 'condition, taken_br, rewrite_stmt')
```
sklam, line 400: I'd prefer to move the `namedtuple()` out to global as it's quite an expensive operation.

stuartarchibald: Good point, done.

Issue #5615, opened 2020-04-23
First class function type breaks literal_unroll


stuartarchibald:
## Reporting a bug

- [x] I am using the latest released version of Numba (most recent is visible in
 the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [x] I have included below a minimal working reproducer (if you are unsure how
 to write one see http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports).

This worked on 0.48:
```python
from numba import literal_unroll, njit
import numba
import numpy as np

@njit
def foo1(x):
    return x + 1


@njit
def foo2(x):
    return x + 2


tup = ((foo1, foo2), (foo2, foo1))


@njit
def bar(fcs):
    x = 0
    a = 10
    for t in literal_unroll(fcs):
        i, j = t
        x += i(j(a))
    return x

print(numba.__version__)
print(bar(tup))
```
e.g.
```
$ python foo.py                                                                                                               
0.48.0                                                                                                                          
26 
```
but in 0.49 it does this:
```
0.50.0dev0+128.g82c7d37ad
<path>/numba/core/utils.py:508: NumbaExperimentalFeatureWarning: First-class function type feature is experimental
  warnings.warn("First-class function type feature is experimental",
gi196.py:22: NumbaExperimentalFeatureWarning: First-class function type feature is experimental
  for t in literal_unroll(fcs):
gi196.py:22: NumbaExperimentalFeatureWarning: First-class function type feature is experimental
  for t in literal_unroll(fcs):
gi196.py:23: NumbaExperimentalFeatureWarning: First-class function type feature is experimental
  i, j = t
<path>/numba/core/utils.py:508: NumbaExperimentalFeatureWarning: First-class function type feature is experimental
  warnings.warn("First-class function type feature is experimental",
Traceback (most recent call last):
  File "<path>/numba/core/errors.py", line 726, in new_error_context
    yield
  File "<path>/numba/core/lowering.py", line 267, in lower_block
    self.lower_inst(inst)
  File "<path>/numba/core/lowering.py", line 364, in lower_inst
    val = self.lower_assign(ty, inst)
  File "<path>/numba/core/lowering.py", line 538, in lower_assign
    return self.lower_expr(ty, value)
  File "<path>/numba/core/lowering.py", line 1193, in lower_expr
    return impl(self.builder, (self.loadvar(expr.value.name),
  File "<path>/numba/core/base.py", line 1155, in __call__
[...]

PR #5626, opened 2020-04-24
Fix unifying undefined first class function types issue.

pearu:
Tackles #5615 .

### DIFF
--- a/numba/core/utils.py
+++ b/numba/core/utils.py
@@ -532,8 +532,6 @@ def unified_function_type(numba_types, require_precise=True):
             if isinstance(t, types.UndefinedFunctionType):
                 if undefined_function is None:
                     undefined_function = t
-                elif 0:
-                    assert undefined_function == t
                 dispatchers.update(t.dispatchers)
             else:
                 if function is None:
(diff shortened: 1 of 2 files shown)


numba/tests/test_function_type.py
```diff
@@ -1072,3 +1072,28 @@ def bar_good(foos):
...
+    def test_issue_5615(self):
+        @njit
+        def foo1(x):
+            return x + 1
+
+        @njit
+        def foo2(x):
+            return x + 2
+
+        from numba import literal_unroll
```
stuartarchibald, line 1085: Think this needs to be a global (there's a bug, don't recall ticket, which makes `literal_unroll` as a freevar not work).

stuartarchibald: Fixed the bug https://github.com/numba/numba/pull/5638

pearu: Thanks @stuartarchibald, I will update this PR accordingly.

pearu: Done.

numba/core/utils.py
```diff
@@ -532,8 +532,6 @@ def unified_function_type(numba_types, require_precise=True):
             if isinstance(t, types.UndefinedFunctionType):
                 if undefined_function is None:
                     undefined_function = t
-                else:
-                    assert undefined_function == t
```
sklam, line 536: I'm trying to understand this patch. I see that when this condition is reached, this function will return None. Couldn't it return `None` at this point?

stuartarchibald:
Indirectly merged by https://github.com/numba/numba/pull/5658 thanks for the patch.

PR #5638, opened 2020-04-28
Make literal_unroll function work as a freevar.

stuartarchibald:
As title. For context see comment on #5626.


stuartarchibald:
This should probably go in 0.49.1 as it helps with #5615

Issue #5611, opened 2020-04-22
Control flow graph algorithm error when using many ternary operations


lespaul:
## Reporting a bug

- [x] I am using the latest released version of Numba (most recent is visible in
 the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [x] I have included below a minimal working reproducer (if you are unsure how
 to write one see http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports).

```python
from numba import njit
import numpy as np


def generate_code(count):
    base_code = """
    dependency_{index} = 1 if True else 0
    value_{index} = x if dependendency_{index} else 0"""

    code = """
@njit
def template(x):"""
    for i in range(count):
        code += base_code.format(**{'index': i})
    return code

if __name__ == '__main__':
    import sys
    import time

    count = int(sys.argv[1])

    exec(generate_code(count), globals(), locals())
    start = time.time()
    template(32)
    end = time.time()
    print(end - start, 'elapsed')
```
When running the above code with count > 200, the growth in compilation time appears to grow exponentially based on the control flow graph search.  @stuartarchibald suggested that the recursive algorithm would likely need to be replaced with an iterative one as "the amount of time it takes to analyse the CFG would be to some degree proportional to the number of [if] blocks."  When run with sufficiently large count > 10000, the following error is produced:

```python
  File "/***/.local/lib/python3.8/site-packages/numba/core/dispatcher.py", line 420, in _compile_for_args
    raise e
  File "/***/.local/lib/python3.8/site-packages/numba/core/dispatcher.py", line 353, in _compile_for_args
    return self.compile(tuple(argtypes))
  File "/***/.local/lib/python3.8/site-packages/numba/core/compiler_lock.py", line 32, in _acquire_compile_lock
    return func(*args, **kwargs)
  File "/***/.local/lib/python3.8/site-packages/numba/core/dispatcher.py", line 794, in compile
    cres = self._compiler.compile(args, return_type)
  File "/***/.local/lib/python3.8/site-packages/numba/core/dispatcher.py", line 77, in compile
    status, retval = self._compile_cached(args, return_type)
  File "/***/.local/lib/python3.8/site-packages/numba/core/dispatcher.py", line 91, in _compile_cached
    retval = self._compile_core(args, return_type)
  File "/***/.local/lib/python3.8/site-packages/numba/core/dispatcher.py", line 104, in _compile_core
    cres = compiler.compile_extra(self.targetdescr.typing_context,
[...]

PR #5645, opened 2020-04-28
Fix SSA redundant CFG computation

sklam:
Addresses SSA reconstruction performance problem seen in #5611.

### DIFF
--- a/numba/core/ssa.py
+++ b/numba/core/ssa.py
@@ -48,6 +48,7 @@ def _run_ssa(blocks):
         return {}
     # Run CFG on the blocks
     cfg = compute_cfg_from_blocks(blocks)
+    df_plus = _iterated_domfronts(cfg)
     # Find SSA violators
     violators = _find_defs_violators(blocks)
     # Process one SSA-violating variable at a time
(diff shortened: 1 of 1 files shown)


stuartarchibald:
This patch looks good, the CFG and derived properties like dominance frontiers are SSA transform invariant so compute once and cache makes sense.

numba/core/ssa.py
```diff
@@ -60,19 +62,19 @@ def _run_ssa(blocks):
         _logger.debug("Replaced assignments: %s", pformat(defmap))
         # Fix up the RHS
         # Re-associate the variable uses with the reaching definition
-        blocks = _fix_ssa_vars(blocks, varname, defmap)
+        blocks = _fix_ssa_vars(blocks, varname, defmap, cfg, df_plus)
```
stuartarchibald, line 65: Is it worth asserting that the CFG hasn't changed on exit? CFGs are comparable with `__eq__`

stuartarchibald:
One minor thing to resolve else looks good, thanks for the patch.

numba/core/ssa.py
sklam: That would require recomputing the cfg and would be too expensive given that SSA is already too costly in our compilation pipeline.  We have to rely on making sure the SSA algorithm are not mutating the CFG.

numba/core/ssa.py
```diff
@@ -63,6 +63,12 @@ def _run_ssa(blocks):
         # Fix up the RHS
         # Re-associate the variable uses with the reaching definition
         blocks = _fix_ssa_vars(blocks, varname, defmap, cfg, df_plus)
+
+    # Post-condition checks.
+    # CFG invariant
+    cfg_post = compute_cfg_from_blocks(blocks)
+    if cfg_post != cfg:
+        raise AssertionError("CFG mutated in SSA pass")
```
stuartarchibald, line 71: Should this be a `errors.CompilerError` perhaps?

sklam: yes indeed

numba/core/ssa.py
stuartarchibald: out of band, cost of a single CFG recomputation on loop exit was considered worth it

PR #5648, opened 2020-04-29
Unset the memory manager after EMM Plugin tests

gmarkall:
Setting the EMM plugin to the internal memory manager is incorrect when the tests are being run with an external memory manager, e.g. with:

```
NUMBA_CUDA_MEMORY_MANAGER=rmm python -m numba.runtests numba.cuda.tests
```

After the EMM plugin tests execute, testing would continue with the internal memory manager instead of the specified external memory manager. The correct behaviour is to un-set it, so that when the new context is created, it is set to the chosen memory manager (either the internal one or an external one).

### DIFF
--- a/numba/cuda/tests/cudadrv/test_emm_plugins.py
+++ b/numba/cuda/tests/cudadrv/test_emm_plugins.py
@@ -110,10 +110,9 @@ class TestDeviceOnlyEMMPlugin(CUDATestCase):
         cuda.set_memory_manager(DeviceOnlyEMMPlugin)
 
     def tearDown(self):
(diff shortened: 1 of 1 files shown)


stuartarchibald:
Buildfarm ID `numba_smoketest_cuda_48`.

esc:
@gmarkall @stuartarchibald will this be included in 0.49.1?

gmarkall:
There's no hurry for it - I would not burden 0.49.1 with it.

esc:
@gmarkall thanks for the update, I have set the milestone accordingly.

PR #5651, opened 2020-04-29
Fix some SSA issues

sklam:
as titled.


sklam:
There's a unrelated doc failure in CI.

PR #5652, opened 2020-04-29
Pin to sphinx=2.4.4 to avoid problem with C declaration

sklam:
This unbreaks master.

### DIFF
--- a/buildscripts/incremental/setup_conda_environment.sh
+++ b/buildscripts/incremental/setup_conda_environment.sh
@@ -78,7 +78,7 @@ fi
 $CONDA_INSTALL -c numba llvmlite
 
 # Install dependencies for building the documentation
-if [ "$BUILD_DOC" == "yes" ]; then $CONDA_INSTALL sphinx sphinx_rtd_theme pygments numpydoc; fi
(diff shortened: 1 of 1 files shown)


esc:
I tried this locally with the correct sphinx version and can confirm that the warnings don't appear with it. Also, judging from public CI the version change was made in the correct place as the `doc` build now passes. 👍

buildscripts/incremental/setup_conda_environment.sh
```diff
@@ -78,7 +78,7 @@ fi
 $CONDA_INSTALL -c numba llvmlite
 
 # Install dependencies for building the documentation
-if [ "$BUILD_DOC" == "yes" ]; then $CONDA_INSTALL sphinx sphinx_rtd_theme pygments numpydoc; fi
+if [ "$BUILD_DOC" == "yes" ]; then $CONDA_INSTALL sphinx=2.4.4 sphinx_rtd_theme pygments numpydoc; fi
```
stuartarchibald, line 81: This change needs making on windows `setup_conda_environment.cmd` too?

stuartarchibald:
@esc I've updated this patch series with 347b6b5 for windows, please could you re-review? Thanks.

stuartarchibald:
Abandon all that, forgot that the windows script isn't actually in use as it doesn't work on Azure. Going with original patch in 8586057, once re-tested will merge.

PR #5653, opened 2020-04-30
Update doc with respect to try-finally case

guilhermeleobas:
It seems that `try .. finally` without an `exception` is supported:

```python
@njit
def qux():
	try:
		print('try block')
	finally:
		print('finally block')
                 return "works!"

print('\n-- Test qux --')
l3 = List.empty_list(types.int64)
try:
	print('qux:', qux(l3))
except:
	print('qux: UnsupportedError')
```

Prints:
```
-- Test qux --
try block
finally block
qux: works!
```


esc:
@guilhermeleobas thanks for submitting this to the Numba issue tracker. I tried to run your code locally, and I believe that you may have made a copy and paste error. I think the snippet should perhaps be:

```
from numba import njit


@njit
def qux():
    try:
        print('try block')
    finally:
        print('finally block')
        return "works!"


print('\n-- Test qux --')
print('qux:', qux())
```

None the less, the conclusion is correct and the change to docs is also fine. I built the docs locally to double check and it looks good.

guilhermeleobas:
Hi @esc. Yes, there was a small mistake in the snipped of code I posted in the other message. Thanks for pointing it out.

guilhermeleobas:
@esc, I did another modification in the docs after you put the tag `5 - ready to merge`

stuartarchibald:
@sklam is this an oversight or just working by accident? If not sure from memory I'll take a look at the code.

a.py
```diff
@@ -0,0 +1,105 @@
+import numpy as np
```
esc, line 1: I am not sure this file is supposed to be included?

docs/source/reference/pysupported.rst
```diff
@@ -190,11 +190,32 @@ use-case would look like:
...
+    ...     except Exception:
+    ...         print('handler block')
+    ...     else:
+    ...         print('else block')
+    ...     finally:
+    ...         print('final block')
+    ...
+    >>> foo()
+    main block
+    handler block
```
esc, line 214: Should this not be `else block`.  I don't see  any exception being raised that would trigger execution of the `handler block`?

esc:
@guilhermeleobas I've taken this for another spin and left some comments.

a.py
guilhermeleobas: My mistake

docs/source/reference/pysupported.rst
guilhermeleobas: you're right

sklam:
RE: https://github.com/numba/numba/pull/5653#issuecomment-622330213

Yes, it's an oversight that I didn't put docs on try..finally. These constructs are tested.

stuartarchibald:
> The PR looks good. Although we should reach a consensus as to the status of this feature: "works by accident?!" vs. "definitely intended!!".

@esc @sklam resolved here: https://github.com/numba/numba/pull/5653#issuecomment-623574778 ?

esc:
@stuartarchibald good, then, let's merge.

PR #5659, opened 2020-04-30
Add support for parfors creating 'F'ortran layout Numpy arrays.

### LINKED ISSUE #5625: function with parallel forloops + njit raises error (inconsistently)
## Reporting a bug

I have the following function : 
``` 
@njit(parallel=True)
def cosine_similarity_loop_numba(out, x_lens, x, xx, y, yy):
    for x_idx in prange(x.shape[2]):
        for x_t in range(x_lens[x_idx]):
            for y_t in range(y.shape[1]):
                out[x_t, y_t, x_idx] = (x[:, x_t, x_idx] * y[:, y_t]).sum() / ((xx[x_t, x_idx] ** .5) * (yy[y_t] ** .5))
    return out
```
[reproduce_parfor_bug.zip](https://github.com/numba/numba/files/4528927/reproduce_parfor_bug.zip)
I have placed an example of inputs that raise the issue. It's in numpy savez format - so use 
```
obj = np.load(path_to_file)
cosine_similarity_loop_numba(obj["out"], obj["x_lens"], obj["x"], obj["xx"], obj["y"], obj["yy"])
```

The error I get is: 
```
numba.errors.UnsupportedError: Failed in nopython mode pipeline (step: nopython mode backend)
Unsupported op-code encountered: arrayexpr(expr=(<built-in function mul>, [Var($58.10, d blah.py:66), Var($58.18,  blah.py:66)]), ty=array(float64, 1d, F))
File " /blah/blah..py", line 66:
def cosine_similarity_loop_numba(out, x_lens, x, xx, y, yy):
    <source elided>
            for y_t in range(y.shape[1]):
                out[x_t, y_t, x_idx] = (x[:, x_t, x_idx] * y[:, y_t]).sum() / ((xx[x_t, x_idx] ** .5) * (yy[y_t] ** .5))
                ^
[1] During: lowering "id=1[LoopNest(index_variable = parfor_index.65, range = (0, $4.5, 1))]{36: <ir.Block at  /blah/blah..py (64)>, 38: <ir.Block at  /blah/blah..py (64)>, 20: <ir.Block at  /blah/blah..py (63)>, 183: <ir.Block at  /blah/blah..py (66)>, 56: <ir.Block at  /blah/blah.py (65)>, 185: <ir.Block at  /blah/blah.py (66)>, 58: <ir.Block at  /blah/blah..py (65)>}Var(parfor_index.65,  blah.py:63)" at /blah/blah.py (63)

```
( I've replaced the actual paths and file names with /blah/blah...

DrTodd13:
Resolves #5625 

If the target of an arrayexpr is a Fortran layout array then we allocate an empty C layout array in the parfor init block and then call numpy.asfortranarray to convert that empty array to Fortran layout.

Also, please note that this is a correctness-only fix.  Multi-dimensional access to Fortran arrays would still use the C loop structure and would therefore be inefficient from a cache perspective.  If all the arrays are F then we could invert the loop structure but if the arrays are mixed then more thought would need to go into optimizing performance.

### DIFF
--- a/numba/core/ir_utils.py
+++ b/numba/core/ir_utils.py
@@ -57,7 +57,7 @@ def next_label():
     return _max_label
 
 
-def mk_alloc(typemap, calltypes, lhs, size_var, dtype, scope, loc):
+def mk_alloc(typemap, calltypes, lhs, size_var, dtype, scope, loc, lhs_typ):
     """generate an array allocation with np.empty() and return list of nodes.
     size_var can be an int variable or tuple of int variables.
     """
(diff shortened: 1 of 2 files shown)


DrTodd13:
@stuartarchibald What would you like to do about testing this new functionality?  I don't know how your modified test case would fit in the testing regime.  So many of the existing tests could be duplicated to take 'F' arrays, so much so that we could consider automating the dual testing of C arrays versus F arrays but combinations are also potentially important to test.

stuartarchibald:
Thanks for the patch. I think perhaps adding an option for `F` order arrays to existing tests, and then based on that perhaps we could do some sort of cycle or shuffle around `C/F` similar to 
https://github.com/numba/numba/blob/master/numba/tests/test_linalg.py#L1376-L1378?

DrTodd13:
@stuartarchibald I'm not familiar with the code in this suggestion.  Would it be possible for you or Siu to do this part?

sklam:
Added tests in https://github.com/numba/numba/pull/6664

PR #5660, opened 2020-05-01
Make coverage.py work with compiled code.

### LINKED ISSUE #4268: coverage.py support for Numba
It would be extremely helpful to be able to capture line coverage in Numba-compiled functions the same as in Python functions with coverage.py.

Doing this would most likely require something like the following:

* Numba compiler pipeline detects when coverage.py is active and injects a bit of code after each line to register execution of "line X in file Y" in a data structure.  We might need to switch on compiler debug mode to avoid code relocation that would confuse the measurement.
* We could make the data structure hang off the [Environment](https://github.com/numba/numba/blob/master/numba/_dynfunc.c#L17-L24) object that is passed to every function call.  It is append-only, but needs to be thread-safe.
* When transitioning back to the interpreter, we should run down the log of executed lines, populate the relevant data structure in coverage.py, and clear the log.

@cdeil noted on the mailing list that Cython has done this already.  Relevant links:
* http://blog.behnel.de/posts/coverage-analysis-for-cython-modules.html
* https://cython.readthedocs.io/en/latest/src/tutorial/profiling_tutorial.html#enabling-coverage-analysis
* https://github.com/cython/cython/blob/master/Cython/Coverage.py

sklam:
Closes #4268

### DIFF
--- a/numba/_dispatcher.c
+++ b/numba/_dispatcher.c
@@ -67,7 +67,7 @@ call_trace_protected(Py_tracefunc func, PyObject *obj,
  * exception state is preserved correctly).
  */
 #define C_TRACE(x, call)                                        \
-if (call_trace(tstate->c_profilefunc, tstate->c_profileobj,     \
+if (call_trace(tstate->c_tracefunc, tstate->c_traceobj,     \
                tstate, tstate->frame, PyTrace_CALL, cfunc))	\
     x = NULL;                                                   \
 else                                                            \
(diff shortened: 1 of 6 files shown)


sklam:
Modified how PyFrames are injected. `cProfile` is also working and enabled by default on the latest commit. https://github.com/numba/numba/pull/5660/commits/9d6d5bb354f89091285443c103855a74cee255a5

maxnoe:
We would be very interested in this feature, since we have a codebase that is increasingly relying on numba to speed up hot parts. 

Any update on this? I fear this is too deep into numba's internals for me to contribute.

esc:
@maxnoe it's a tough one and I haven't heard about any updates just yet. If you need to check test coverage you can use `NUMBA_DISABLE_JIT=1` to allow coverage to work, but that has caveats too and isn't perfect.

namoshizun:
Any news on this PR? This seems like a quite important update...

esc:
> Any news on this PR? This seems like a quite important update...

It became stale and never made it due to lack of time. Looking for a new champion to pick up this work. Pull-requests are very welcome and you can use the code in this PR as a starter patch! 🙌☄️

PR #5676, opened 2020-05-06
Restore `numba.types` as public API

sklam:
as titled.

### DIFF
--- a/numba/tests/test_moved_modules.py
+++ b/numba/tests/test_moved_modules.py
@@ -1,0 +1,21 @@
+"""Tests for moved modules and their redirection from old path
+"""
+from numba.tests.support import TestCase
+
+
+class TestMovedModule(TestCase):
+    """Testing moved modules in Q1 2020 but were decided to kept as public API
+    """
+    def tests_numba_types(self):
+        import numba.types
+        import numba.core.types as types
+        # The old module is the new module
+        self.assertIs(numba.types, types)
+        # Attribute access are there
+        self.assertIs(numba.types.intp, types.intp)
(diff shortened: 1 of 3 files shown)


numba/tests/test_moved_modules.py
```diff
@@ -0,0 +1,21 @@
+"""Tests for moved modules and their redirection from old path
+"""
+from numba.tests.support import TestCase
+
+
+class TestMovedModule(TestCase):
+    """Testing moved modules in Q1 2020 but were decided to kept as public API
+    """
+    def tests_numba_types(self):
+        import numba.types
```
stuartarchibald, line 10: Should e.g. `from numba.types import containers` be explicitly tested too?

sklam: i can add a test. In general, I am only testing a few things because I am already checking for object identity. I don't think that's any reason for it to not work.

sklam:
That error on macos is odd.

sklam:
This is unexpected:

```python

In [1]: import numba.types.misc

In [2]: numba.types.misc.StringLiteral
Out[2]: numba.types.misc.StringLiteral

In [3]: from numba.core import types

In [4]: types.StringLiteral
Out[4]: numba.core.types.misc.StringLiteral
```

numba/tests/test_moved_modules.py
sklam: test added in https://github.com/numba/numba/pull/5676/commits/2998661e6f33d8b437aadd70f0b0f08302524b0e

Issue #5678, opened 2020-05-07
bug in analysis NumbaParallelSafetyWarning


Harveyt47:
## Reporting a bug

- [x] I am using the latest released version of Numba (most recent is visible in
 the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [x] I have included below a minimal working reproducer (if you are unsure how
 to write one see http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports).

When I run 
```
@njit(parallel=True)
def while_for(n, max_iter=10):
    a = np.empty((n,n))
    i=0
    while i <= max_iter:
        for j in prange(len(a)):
            for k in prange(len(a)):
                a[j,k] = np.random.rand()
        i+=1
    return a
while_for(4)
```
I get the following warning 
```
Variable i.3 used in parallel loop may be written to simultaneously by multiple workers and may result in non-deterministic or unintended results. 
File "<ipython-input-44-a9dc46b8042f>", line 6: 
def while_for(n, max_iter=10): 
<source elided> 
while i <= max_iter: 
for j in prange(len(a)): ^ warnings.warn(NumbaParallelSafetyWarning(msg, loc))
```
When I replace `@njit(parallel=True)` with `@njit` it goes away.

stuartarchibald:
Thanks for the report. As a guess, seems like that this is because `i` is a parfors var but unused in the loop nest. @DrTodd13 please could you take a look? Thanks.

DrTodd13:
@sklam @stuartarchibald The parfor code detects races in a parallel region by looking for assignments to variables that are live-out of the region that are not reductions.  I think that SSA is doing the following with the variable i.

```
def while_for(n, max_iter=1):
    a = np.empty((n,n))
    i=0
    #i2 = i
    while i <= max_iter:
        #i3 = i2
        for j in prange(len(a)):
            #i4 = i3
            for k in range(len(a)):
                #i4 = i4
                a[j,k] = np.random.rand()
            #i3 = i4
        i+=1
        #i1 = i3 + 1
        #i2 = i1
    return a
```

So, inside the parallel region we have i.3 = i.4 and i.3 is live-out.  It doesn't seem like i.3 and i.4 are necessary for the SSA transformation.  Maybe they were just added for consistency of the SSA code.

[...]

PR #5686, opened 2020-05-08
Fix issue with SSA not minimal

sklam:
Fixes https://github.com/numba/numba/issues/5678


numba/core/ssa.py
```diff
@@ -382,32 +363,31 @@ def _find_def_from_top(self, states, label, loc):
...
-                freshvar = scope.redefine(states['varname'], loc=loc)
-                # insert phi
-                phinode = ir.Assign(
-                    target=freshvar,
-                    value=ir.Expr.phi(loc=loc),
-                    loc=loc,
+        phi_locations = states['phi_locations']
+
+        if label in phi_locations:
+            scope = states['scope']
```
sklam, line 369: code in this `if` block is unchanged, just unindent

stuartarchibald:
Thanks for the patch. I've manually checked this, it looks correct. Did you want to consider any of those places in parfors where we had to do `unversioned_name` mangling or get this in and then take a look?

stuartarchibald:
Please could you resolve the conflicts by merging master into this branch. Thanks!

sklam:
RE: https://github.com/numba/numba/pull/5686#issuecomment-629194367

Let's get this in. We will look later.

Issue #2102, opened 2016-09-06
Array stacking and concatenation


seibert:
This is a meta-issue to track the various NumPy stacking and concatenation methods.

#2069 covers hstack, vstack, dstack, column_stack.

Still need:
* tile()
* split()
* fliplr()
* flipud()

Is there a method to physically transpose an array in memory (rather than return a view)?

PR #5699, opened 2020-05-13
Implement np.split and np.array_split

ivirshup:
This PR implements `np.split` and `np.array_split`, and would help towards closing #2102. I think this is ready for review.

Questions: 

*  For testing, I'm not sure what assertion method I should use in place of `np.testing.assert_equal`, since the result is a list of arrays. Any suggestions?
* For implementations of numpy functions, what naming convention should be used for the arguments? I've tried to follow other functions in the file, but I was wondering if I should use the same arguments names as numpy.

### DIFF
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -5007,7 +5007,7 @@ def np_split(a, indices, axis=0):
             l, rem = divmod(a.shape[axis], indices)
             if rem != 0:
                 raise ValueError("array split does not result in an equal division")
-            return np.split(a, list(range(l, len(a), l)), axis=axis)
+            return np.split(a, np.arange(l, len(a), l), axis=axis)
 
     else:
 
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -5006,7 +5006,7 @@ def np_split(a, indices, axis=0):
         def impl(a, indices, axis=0):
             l, rem = divmod(a.shape[axis], indices)
             if rem != 0:
-                raise ValueError()
+                raise ValueError("array split does not result in an equal division")
             return np.split(a, list(range(l, len(a), l)), axis=axis)
 
     else:
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -31,7 +31,7 @@ from numba.core.extending import (register_jitable, overload, overload_method,
                                   intrinsic)
 from numba.misc import quicksort, mergesort
 from numba.cpython import slicing
-from numba.cpython.unsafe.tuple import tuple_setitem
+from numba.cpython.unsafe.tuple import tuple_setitem, build_full_slice_tuple
 
 
 def set_range_metadata(builder, load, lower_bound, upper_bound):
@@ -4964,33 +4964,6 @@ def np_flip(a):
     return impl
 
 
-@intrinsic
-def _build_full_slice_tuple(tyctx, sz):
-    """Creates a sz-tuple of full slices"""
-    size = int(sz.literal_value)
-    tuple_type = types.UniTuple(dtype=types.slice2_type, count=size)
-    sig = tuple_type(sz)
-
-    def codegen(context, builder, signature, args):
-        def impl(length, empty_tuple):
-            out = empty_tuple
-            for i in range(length):
-                out = tuple_setitem(out, i, slice(None, None))
-            return out
-
-        inner_argtypes = [types.intp, tuple_type]
-        inner_sig = typing.signature(tuple_type, *inner_argtypes)
-        ll_idx_type = context.get_value_type(types.intp)
-        # Allocate an empty tuple
-        empty_tuple = context.get_constant_undef(tuple_type)
-        inner_args = [ll_idx_type(size), empty_tuple]
-
-        res = context.compile_internal(builder, impl, inner_sig, inner_args)
-        return res
-
-    return sig, codegen
-
-
 @overload(np.array_split)
 def np_array_split(ary, indices_or_sections, axis=0):
     # If this statement is put at the top of the file, numba fails to import
@@ -5012,7 +4985,7 @@ def np_array_split(ary, indices_or_sections, axis=0):
     else:
 
         def impl(ary, indices_or_sections, axis=0):
-            slice_tup = _build_full_slice_tuple(ary.ndim)
+            slice_tup = build_full_slice_tuple(ary.ndim)
             out = List.empty_list(a_type, len(indices_or_sections) + 1)
             prev = 0
             for cur in indices_or_sections:
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -4966,6 +4966,16 @@ def np_flip(a):
 
 @overload(np.array_split)
 def np_array_split(ary, indices_or_sections, axis=0):
+    if isinstance(ary, (types.UniTuple, types.ListType, types.List)):
+        def impl(ary, indices_or_sections, axis=0):
+            return np.array_split(
+                np.asarray(ary),
+                indices_or_sections,
+                axis=axis
+            )
+
+        return impl
+
     if isinstance(indices_or_sections, types.Integer):
         def impl(ary, indices_or_sections, axis=0):
             l, rem = divmod(ary.shape[axis], indices_or_sections)
@@ -5019,6 +5029,12 @@ def np_array_split(ary, indices_or_sections, axis=0):
 def np_split(ary, indices_or_sections, axis=0):
     # This is just a wrapper of array_split, but with an extra error if
     # indices is an int.
+    if isinstance(ary, (types.UniTuple, types.ListType, types.List)):
+        def impl(ary, indices_or_sections, axis=0):
+            return np.split(np.asarray(ary), indices_or_sections, axis=axis)
+
+        return impl
+
     if isinstance(indices_or_sections, types.Integer):
         def impl(ary, indices_or_sections, axis=0):
             l, rem = divmod(ary.shape[axis], indices_or_sections)
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -29,7 +29,6 @@ from numba.core.imputils import (lower_builtin, lower_getattr,
 from numba.core.typing import signature
 from numba.core.extending import (register_jitable, overload, overload_method,
                                   intrinsic)
-from numba.typed import List
 from numba.misc import quicksort, mergesort
 from numba.cpython import slicing
 from numba.cpython.unsafe.tuple import tuple_setitem
@@ -4993,19 +4992,25 @@ def _build_full_slice_tuple(tyctx, sz):
 
 
 @overload(np.split)
-def split(a, indices, axis=0):
+def np_split(a, indices, axis=0):
+    # If this statement is put at the top of the file, numba fails to import
+    from numba.typed import List
+
     if a.ndim > 1:
         a_type = types.Array(a.dtype, a.ndim, "A")
     else:
         a_type = typeof(a)
 
     if isinstance(indices, types.Integer):
+
         def impl(a, indices, axis=0):
             l, rem = divmod(a.shape[axis], indices)
             if rem != 0:
                 raise ValueError()
             return np.split(a, list(range(l, len(a), l)), axis=axis)
+
     else:
+
         def impl(a, indices, axis=0):
             slice_tup = _build_full_slice_tuple(a.ndim)
             out = List.empty_list(a_type)
@@ -5016,6 +5021,7 @@ def split(a, indices, axis=0):
                 prev = cur
             out.append(a[tuple_setitem(slice_tup, axis, slice(cur, None))])
             return out
+
     return impl
 
 
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -4991,8 +4991,8 @@ def _build_full_slice_tuple(tyctx, sz):
     return sig, codegen
 
 
-@overload(np.split)
-def np_split(a, indices, axis=0):
+@overload(np.array_split)
+def np_array_split(a, indices, axis=0):
     # If this statement is put at the top of the file, numba fails to import
     from numba.typed import List
 
@@ -5005,9 +5005,7 @@ def np_split(a, indices, axis=0):
 
         def impl(a, indices, axis=0):
             l, rem = divmod(a.shape[axis], indices)
-            if rem != 0:
-                raise ValueError("array split does not result in an equal division")
-            return np.split(a, np.arange(l, len(a), l), axis=axis)
+            return np.array_split(a, np.arange(l + rem, len(a), l), axis=axis)
 
     else:
 
@@ -5022,6 +5020,21 @@ def np_split(a, indices, axis=0):
             out.append(a[tuple_setitem(slice_tup, axis, slice(cur, None))])
             return out
 
+    return impl
+
+
+@overload(np.split)
+def np_split(a, indices, axis=0):
+    # This is just a wrapper of array_split, but with an extra error if indices is an int.
+    if isinstance(indices, types.Integer):
+        def impl(a, indices, axis=0):
+            l, rem = divmod(a.shape[axis], indices)
+            if rem != 0:
+                raise ValueError("array split does not result in an equal division")
+            return np.array_split(a, np.arange(l, len(a), l), axis=axis)
+    else:
+        def impl(a, indices, axis=0):
+            return np.array_split(a, indices, axis=axis)
     return impl
 
 
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -5037,17 +5037,13 @@ def np_split(ary, indices_or_sections, axis=0):
 
     if isinstance(indices_or_sections, types.Integer):
         def impl(ary, indices_or_sections, axis=0):
-            l, rem = divmod(ary.shape[axis], indices_or_sections)
+            _, rem = divmod(ary.shape[axis], indices_or_sections)
             if rem != 0:
                 raise ValueError(
                     "array split does not result in an equal division"
                 )
-            indices = np.cumsum(np.array(
-                [l + 1] * rem +
-                [l] * (indices_or_sections - rem - 1)
-            ))
             return np.array_split(
-                ary, indices, axis=axis
+                ary, indices_or_sections, axis=axis
             )
 
         return impl
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -14,7 +14,7 @@ from llvmlite.llvmpy.core import Constant
 
 import numpy as np
 
-from numba import pndindex
+from numba import pndindex, literal_unroll
 from numba.core import types, utils, typing, errors, cgutils, extending
 from numba.np.numpy_support import (as_dtype, carray, farray, is_contiguous,
                                     is_fortran)
@@ -4966,17 +4966,19 @@ def np_flip(a):
 
 @overload(np.array_split)
 def np_array_split(ary, indices_or_sections, axis=0):
-    # If this statement is put at the top of the file, numba fails to import
     if isinstance(indices_or_sections, types.Integer):
-
         def impl(ary, indices_or_sections, axis=0):
             l, rem = divmod(ary.shape[axis], indices_or_sections)
             return np.array_split(
                 ary, np.arange(l + rem, ary.shape[axis], l), axis=axis
             )
 
-    else:
+        return impl
 
+    elif (
+        isinstance(indices_or_sections, types.IterableType)
+        and isinstance(indices_or_sections.iterator_type.yield_type, types.Integer)
+    ):
         def impl(ary, indices_or_sections, axis=0):
             slice_tup = build_full_slice_tuple(ary.ndim)
             out = list()
@@ -4988,7 +4990,24 @@ def np_array_split(ary, indices_or_sections, axis=0):
             out.append(ary[tuple_setitem(slice_tup, axis, slice(cur, None))])
             return out
 
-    return impl
+        return impl
+
+    elif  (
+        isinstance(indices_or_sections, types.Tuple)
+        and all(isinstance(t, types.Integer) for t in indices_or_sections.types)
+    ):
+        def impl(ary, indices_or_sections, axis=0):
+            slice_tup = build_full_slice_tuple(ary.ndim)
+            out = list()
+            prev = 0
+            for cur in literal_unroll(indices_or_sections):
+                idx = tuple_setitem(slice_tup, axis, slice(prev, cur))
+                out.append(ary[idx])
+                prev = cur
+            out.append(ary[tuple_setitem(slice_tup, axis, slice(cur, None))])
+            return out
+
+        return impl
 
 
 @overload(np.split)
@@ -5005,10 +5024,12 @@ def np_split(ary, indices_or_sections, axis=0):
             return np.array_split(
                 ary, np.arange(l, ary.shape[axis], l), axis=axis
             )
+
+        return impl
+
     else:
-        def impl(ary, indices_or_sections, axis=0):
-            return np.array_split(ary, indices_or_sections, axis=axis)
-    return impl
+        return np_array_split(ary, indices_or_sections, axis=axis)
+
 
 
 # -----------------------------------------------------------------------------
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -5025,12 +5025,15 @@ def np_array_split(a, indices, axis=0):
 
 @overload(np.split)
 def np_split(a, indices, axis=0):
-    # This is just a wrapper of array_split, but with an extra error if indices is an int.
+    # This is just a wrapper of array_split, but with an extra error if
+    # indices is an int.
     if isinstance(indices, types.Integer):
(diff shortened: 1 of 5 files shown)


gmarkall:
Thanks for the PR!

> For testing, I'm not sure what assertion method I should use in place of `np.testing.assert_equal`, since the result is a list of arrays. Any suggestions?

Does it make sense to assert that the lengths of the results of `pyfunc` and `cfunc` are equal, then zip them together and call `np.testing.assert_equal` on each pair of arrays?

> For implementations of numpy functions, what naming convention should be used for the arguments? I've tried to follow other functions in the file, but I was wondering if I should use the same arguments names as numpy.

I don't have a strong preference, but generally I'd probably try and follow the NumPy names so it's easier to correlate argument names with the NumPy documentation.

ivirshup:
> Does it make sense to assert that the lengths of the results of pyfunc and cfunc are equal, then zip them together and call np.testing.assert_equal on each pair of arrays?

Oh, `np.testing.assert_equal` works fine. In a previous PR I was asked to make all assertions through the `TestCase.assert*` methods. I'm not sure if there's one of those methods that will give a useful traceback here and hopefully won't require me to futz with the results first.

>  I'd probably try and follow the NumPy names

👍, did this.

---------------------------------

I also ran some benchmarks, which look like I'd expected since the numpy function is mostly python. As the number of splits increases, this version does better. For small splits, it does worse, but that's might have to do with the overhead of calling a numba function.

This table has array shape on the rows, number of splits as the columns, and a ratio of average time for numba / numpy:

| array_shape \ n_splits      |       1 |       10 |      100 |
|------------------|---------|----------|----------|
| (1000,)          | 3.12551 | 0.98823  | 0.155848 |
| (1000, 10, 10)   | 3.03712 | 0.98269  | 0.159963 |
| (1000, 100, 100) | 3.01084 | 0.929655 | 0.161575 |
| (1000000,)       | 3.00602 | 1.03465  | 0.157308 |

gmarkall:
Thanks for the updates - I'll mark this as Ready for Review.

numba/np/arrayobj.py
```diff
@@ -4958,12 +4958,93 @@ def np_flip(a):
...
+        inner_sig = typing.signature(tuple_type, *inner_argtypes)
+        ll_idx_type = context.get_value_type(types.intp)
+        # Allocate an empty tuple
+        empty_tuple = context.get_constant_undef(tuple_type)
+        inner_args = [ll_idx_type(size), empty_tuple]
+
+        res = context.compile_internal(builder, impl, inner_sig, inner_args)
+        return res
+
+    return sig, codegen
```
stuartarchibald, line 4991: Should this perhaps be documented and go in `numba/cpython/unsafe/tuple.py` ?

numba/np/arrayobj.py
```diff
@@ -4958,12 +4958,93 @@ def np_flip(a):
...
+
+
+@overload(np.array_split)
+def np_array_split(ary, indices_or_sections, axis=0):
+    # If this statement is put at the top of the file, numba fails to import
+    from numba.typed import List
```
stuartarchibald, line 4997: I think that this should just use the `reflected list` even though its deprecated. At some point there'll be a global switch over in Numba's type inference from reflected to typed list, mixing what gets returned from numba functions now may cause confusion/is inconsistent.

numba/np/arrayobj.py
```diff
@@ -4958,12 +4958,93 @@ def np_flip(a):
...
+
+    if ary.ndim > 1:
+        a_type = types.Array(ary.dtype, ary.ndim, "A")
+    else:
+        a_type = typeof(ary)
```
stuartarchibald, line 5002: it'd probably be a good idea to employ `type_can_asarray` to assert that `ary` can be interpreted as an array, and then `ary = np.asarray(ary)` on the first line of the implementation.

numba/np/arrayobj.py
```diff
@@ -4958,12 +4958,93 @@ def np_flip(a):
...
+
+    if isinstance(indices_or_sections, types.Integer):
+
+        def impl(ary, indices_or_sections, axis=0):
+            l, rem = divmod(ary.shape[axis], indices_or_sections)
+            return np.array_split(
+                ary, np.arange(l + rem, ary.shape[axis], l), axis=axis
+            )
+
+    else:
```
stuartarchibald, line 5012: I think this ought to strongly restrict what types other than `Integer` are permissible to prevent strange errors from unhandled types being used.

numba/np/arrayobj.py
```diff
@@ -4958,12 +4958,93 @@ def np_flip(a):
...
+            if rem != 0:
+                raise ValueError(
+                    "array split does not result in an equal division"
+                )
+            return np.array_split(
+                ary, np.arange(l, ary.shape[axis], l), axis=axis
+            )
+    else:
+        def impl(ary, indices_or_sections, axis=0):
+            return np.array_split(ary, indices_or_sections, axis=axis)
```
stuartarchibald, line 5044: same comments as above RE np.asarray and strongly restricting types.

numba/np/arrayobj.py
ivirshup: To me, this should just be an iterable (preferably of known size, though this isn't a strict requirement) of integers. How could I type that?

numba/np/arrayobj.py
ivirshup: Would that just be switching `out = List.empty_list(...)` to `out = list()`?

numba/np/arrayobj.py
ivirshup: I'd like to avoid the `np.asarray` at the beginning if possible, since it could lead to making full copies of array-likes when that's not needed. Also, numpy's `np.split` throws an error for some of the types that pass `type_can_asarray`.

ivirshup:
I've moved the tuple builder to `unsafe.tuple` and switched from the typed list constructor to `list`. I'll wait for feedback before taking a look at the type restrictions. 

`axis` can be a non-literal int. This was a little surprising to me too, but I think this makes sense given the return type of `tuple_setitem` must be the same as the input tuple.

numba/np/arrayobj.py
stuartarchibald: I'm not sure what you mean by:
>  it could lead to making full copies of array-likes when that's not needed

IIRC `np.asarray` doesn't make a copy if the argument is already an array. 

> Also, numpy's np.split throws an error for some of the types that pass type_can_asarray

I see, thanks for clarifying, in which case explicit guards for the types that are accepted ought to be added to prevent spurious typing problems emanating?

numba/np/arrayobj.py
stuartarchibald: Would `isinstance(<arg>, numba.types.IterableType)` work?

numba/np/arrayobj.py
ivirshup: > IIRC np.asarray doesn't make a copy if the argument is already an array.

Yes, but this function should work fine on "array likes". For example, `np.array_split` will return xarray `DataArray`s if you call it on a `DataArray` and `DataFrame`s if called on a `DataFrame`. If this called `np.asarray` on something similar it would return `np.ndarrays`.

> explicit guards for the types that are accepted

Actually, looking at this again, it seems like the cases where I was getting errors might be bugs in numpy. For example:

```python
np.split((1, 2, 3, 4), 2)  # This throws an AttributeError
```

But these should be equivalent, and work (returning lists of arrays)

```python
np.split((1, 2, 3, 4), [2])
np.array_split((1, 2, 3, 4), 2)
```

I think the logic for type conversions is whatever `np.swapaxes` returns.

I guess I would be fine with coercing things to arrays then. I would like it if there was a way for "array likes" to not be converted.

numba/np/arrayobj.py
ivirshup: I'm not completely sure. It looks like it might, but I'm not sure how these instance checks work. Would anything that passed that check have a `yield_type` attribute?

I'm thinking this could be the check:

```python
    elif (
        isinstance(indices_or_sections, types.IterableType) 
        and isinstance(indices_or_sections.iterator_type.yield_type, types.Integer)
    ):
```

ivirshup: Hmm. `UniTuple` works, but `Tuple` does not. Not sure if this would be an issue, but it does disallow cases like: `np.split(a, (np.int32(1), np.int64(4)))`

numba/np/arrayobj.py
ivirshup: Can I know from the type information what attributes/ properties an instance will have?

numba/np/arrayobj.py
stuartarchibald: I suspect that this is because iteration requires the induction variable to be the same type across iterations, else the loop body has to be versioned, see [`literal_unroll`](http://numba.pydata.org/numba-doc/latest/reference/pysupported.html?highlight=literal_unroll#heterogeneous-tuples).

numba/np/arrayobj.py
```diff
@@ -4958,12 +4958,59 @@ def np_flip(a):
...
-        sl = _build_slice_tuple(a.ndim)
+        sl = _build_flip_slice_tuple(a.ndim)
         return a[sl]
 
     return impl
 
 
+@overload(np.array_split)
+def np_array_split(ary, indices_or_sections, axis=0):
+    # If this statement is put at the top of the file, numba fails to import
```
stuartarchibald, line 4969: What does this mean?

stuartarchibald:
@ivirshup Thanks for making those changes, they look good. Is there anything you are waiting on from me or shall I re-review? Thanks again!

ivirshup:
I've update the typing for indices_or_sections, but I'm not happy with any solution I can come up with for `ary`.

The logic I would like is:

* if it looks like an array (i.e. has `.shape`) use the code as it stands now
* If it can be converted to an array, add a `np.asarray` to the top of the implementation body

I'm not sure how to check if something looks like an array. I'd also like to use some external source of truth for figuring out which collections can be converted to an array. I think `type_can_asarray` mostly does this, but is there a consistent way to check if an object is atomic?

numba/np/arrayobj.py
ivirshup: This was leftover from some import that's no longer needed.

stuartarchibald:
> I've update the typing for indices_or_sections, but I'm not happy with any solution I can come up with for `ary`.

Thanks for the update.

> The logic I would like is:
> 
>     * if it looks like an array (i.e. has `.shape`) use the code as it stands now
> 
>     * If it can be converted to an array, add a `np.asarray` to the top of the implementation body
> 
> 
> I'm not sure how to check if something looks like an array. I'd also like to use some external source of truth for figuring out which collections can be converted to an array. I think `type_can_asarray` mostly does this, but is there a consistent way to check if an object is atomic?

What's the concern about just adding `np.asarray` to the array input value? If it's already an array it's a no-op else it converts the argument into an array, the supported types can just be restricted at typing time if there's more restrictions than normal for this function. Also, if NumPy is not hugely consistent it might be applicable to just support e.g. Arrays.

It also seems like there might be a logic bug somewhere, I started testing the code and this:
```python
from numba import njit
import numpy as np

@njit
def foo(ary, spl, axis=0):
    return np.array_split(ary, spl, axis=axis)

x = np.arange(8)
y = 3
print(foo(x, y))
print(foo.py_func(x, y))
```
does this:
```
[array([0, 1, 2, 3]), array([4, 5]), array([6, 7])]
[array([0, 1, 2]), array([3, 4, 5]), array([6, 7])]
```

thanks again for working on this!

ivirshup:
> What's the concern about just adding np.asarray to the array input value?

This is kind of related to the `scipy.sparse` issue on numba-scipy. This function should just work for sparse arrays, but it'll do the wrong thing if there's an `asarray` causing explicit conversion to dense array. This function should just work on anything that looks like an array, and slices like an array. I think the ecosystem becomes easier to extend if the restrictions are kept minimal.

To me, this is one of the most frustrating issues with the python numeric stack (mostly due to the inflexibility of numpy arrays). I'd like to avoid contributing to the assumption of "`np.ndarray` is the only array" as much as possible. I think this also comes from working with Julia, where it's encouraged to use less typing – leading to more extendable code.

> It also seems like there might be a logic bug somewhere

Thanks for catching that! It's been fixed and has tests now.

stuartarchibald:
> > What's the concern about just adding np.asarray to the array input value?
> 
> This is kind of related to the `scipy.sparse` issue on numba-scipy. This function should just work for sparse arrays, but it'll do the wrong thing if there's an `asarray` causing explicit conversion to dense array. This function should just work on anything that looks like an array, and slices like an array. I think the ecosystem becomes easier to extend if the restrictions are kept minimal.
> To me, this is one of the most frustrating issues with the python numeric stack (mostly due to the inflexibility of numpy arrays). I'd like to avoid contributing to the assumption of "`np.ndarray` is the only array" as much as possible. I think this also comes from working with Julia, where it's encouraged to use less typing – leading to more extendable code.

I don't disagree with this, however, it is Numba's job to be able to compile NumPy code and replicate the behaviour of NumPy that exists at present, typically deviations from this only occur when it's impossible to deal with in static typing (e.g. input value dependent output types, for example `np.linalg.eig`) or there's a large performance penalty (e.g. `np.dot()` on integer matrices). At present this works:
```
In [34]: np.array_split((0,1,2,3,4,5,6,7,8), 3)                                                                                 
Out[34]: [array([0, 1, 2]), array([3, 4, 5]), array([6, 7, 8])]
```
and so really Numba should support the same because it is something it can support.

I think perhaps it'd be good to direct your thoughts on the inflexibility and assumptions about `ndarray` being the only array towards the NumPy team? If it gets changed upstream then Numba will reciprocate.

> > It also seems like there might be a logic bug somewhere
> 
> Thanks for catching that! It's been fixed and has tests now.

No problem, thanks for fixing.

ivirshup:
Your point about matching what numpy currently does is fair. I will make it so tuples and lists work like numpy does. However, this is one of the places where numpy does allow extensibility since these are both "array functions", so I think they shouldn't cast arbitrary objects to arrays. The way I've implemented it allows overloading the function for more types, but does not provide a fallback base implementation.

------------------------

As a minor deviation from numpy, `np.split([1,2,3,4], 2)` and `np.split((1,2,3,4), 2)` fail like:

```pytb
AttributeError: 'list' object has no attribute 'shape'
AttributeError: 'tuple' object has no attribute 'shape'
```

Would it be alright if this worked instead?

jaistech-sp:
Hi,
Congrats on your job with array_split! I am using numba '0.50.1' but the function is not available. When are you planing to add it to Numba official repo?

esc:
@jaistech-sp thanks for asking about the availability of this function. The pull-request has not been merged yet, which means it is not yet part of the git main-line (a.k.a. `master` branch). As soon as it is merged it will become available as part of a Numba development release (e.g. `0.52.0dev0`)  which can be installed from the `numba` channel from anaconda.org with:

`conda install -c numba numba`

Once the changes are merged, they will eventually become part of the next release, `0.52.0`, which will be released later on during the year.

jaistech-sp:
@esc Thanks a lot for your quick reply I will be expecting for Numba 0.52.0 (I am updating an entire script to work with njit)
Best and congrats for your job!

ivirshup:
On that topic, @stuartarchibald I believe I've addressed your last review. How is this looking now?

jaistech-sp:
Hi! my apologies for insisting again, but this function (array_split) is critical for a processing of mine with numba (@jit) implemented on top of it. Will array split be ready for the Numba's version 0.52?
Thanks a lot guys!
Best wishes!

esc:
@jaistech-sp thank you for asking about this. At this stage, the answer to that question is 'maybe'. We can not guarantee that a sufficient number of resources will be available to complete the review of this PR. If you urgently need this feature, I would recommend obtaining a copy of the changes and then building Numba locally with the changes applied so as to suite your needs.

jaistech-sp:
> @jaistech-sp thank you for asking about this. At this stage, the answer to that question is 'maybe'. We can not guarantee that a sufficient number of resources will be available to complete the review of this PR. If you urgently need this feature, I would recommend obtaining a copy of the changes and then building Numba locally with the changes applied so as to suite your needs.

Thanks for your reply!
I'll do so, but anyway I will keep fingers crossed in case there could be array_split in Numba 0.52

stuartarchibald:
@ivirshup apologies for the delay, I entirely missed this.

> Your point about matching what numpy currently does is fair. I will make it so tuples and lists work like numpy does. However, this is one of the places where numpy does allow extensibility since these are both "array functions", so I think they shouldn't cast arbitrary objects to arrays. The way I've implemented it allows overloading the function for more types, but does not provide a fallback base implementation.

I've taken a look at the code and I think that what is present is suitable to replicate NumPy, but agree with your point RE object casting, it probably ought to be more strict.

> As a minor deviation from numpy, `np.split([1,2,3,4], 2)` and `np.split((1,2,3,4), 2)` fail like:
> 
> ```
> AttributeError: 'list' object has no attribute 'shape'
> AttributeError: 'tuple' object has no attribute 'shape'
> ```
> 
> Would it be alright if this worked instead?

I think that unfortunately this would not be ok because it means code that is written with the JIT decorator would fail if e.g. the env var `NUMBA_DISABLE_JIT` were set. It is presenting a behavioural difference and this is something that Numba tries really hard to avoid as it make debugging more challenging and breaks things like users switching off the JIT to get coverage metrics.

ivirshup:
No problem!

> It is presenting a behavioural difference and this is something that Numba tries really hard to avoid as it make debugging more challenging and breaks things like users switching off the JIT to get coverage metrics.

My instinct is that this would be "undefined behaviour" for numpy, which might change at any time. Since numba already differs from numpy is some cases like these (e.g. array ordering), would you still want it to throw an error? I've opened an issue at numpy to see what their take is on this (https://github.com/numpy/numpy/issues/17463).

Side question, when is the next numba release? I would like this to get in, so I'm wondering if I should just implement the error or wait to get a response from the numpy devs.

stuartarchibald:
> No problem!

Thanks for your understanding :)

> > It is presenting a behavioural difference and this is something that Numba tries really hard to avoid as it make debugging more challenging and breaks things like users switching off the JIT to get coverage metrics.
> 
> My instinct is that this would be "undefined behaviour" for numpy, which might change at any time. Since numba already differs from numpy is some cases like these (e.g. array ordering), would you still want it to throw an error? I've opened an issue at numpy to see what their take is on this ([numpy/numpy#17463](https://github.com/numpy/numpy/issues/17463)).

Thanks for opening [numpy/numpy#17463](https://github.com/numpy/numpy/issues/17463), you've convinced me that this is anomalous and I think in this case we can accept the analogue to `array_split` at least until that ticket declares otherwise.

> Side question, when is the next numba release? I would like this to get in, so I'm wondering if I should just implement the error or wait to get a response from the numpy devs.

We're aiming to close the [milestone](https://github.com/numba/numba/milestone/40) on 7th October 2020, the cutoff will be late in the day US central time. It'd be great to get this PR in!

numba/tests/test_np_functions.py
```diff
@@ -2175,6 +2183,80 @@ def a_variations():
...
+            yield a, [1, 3], 2
+
+            a = np.arange(100).reshape(2, -1)
+            yield a, 1
+            yield a, 1, 0
+            yield a, [1], 0
+            yield a, 50, 1
+            yield a, np.arange(10, 50, 10), 1
+            yield a, (1,)
+            yield a, (np.int32(4), 10)
```
stuartarchibald, line 2220: I added these locally:
```
            a = np.array([])
            yield a, 1
            yield a, 2
            yield a, (2, 3), 0
            yield a, 1, 0

            a = np.array([[]])
            yield a, 1
            yield a, (2, 3), 1
            yield a, 1, 0
            yield a, 1, 1
```
a few fail. Having looked at the NumPy code, I am wondering if the behaviour is a side effect of the implementation opposed to a specific design. Am considering whether this sort of thing is intended:
```
In [51]: np.array_split(np.array([[[]]]), 3)
Out[51]: 
[array([], shape=(1, 1, 0), dtype=float64),
 array([], shape=(0, 1, 0), dtype=float64),
 array([], shape=(0, 1, 0), dtype=float64)]
```

numba/np/arrayobj.py
```diff
@@ -4958,12 +4958,100 @@ def np_flip(a):
...
 
     def impl(a):
-        sl = _build_slice_tuple(a.ndim)
+        sl = _build_flip_slice_tuple(a.ndim)
         return a[sl]
 
     return impl
 
 
+@overload(np.array_split)
```
stuartarchibald, line 4967: `np.array_split` and `np.split` should be added to the supported functions list in the docs here: https://github.com/numba/numba/blame/master/docs/source/reference/numpysupported.rst#L330 thanks

numba/tests/test_np_functions.py
ivirshup: The shapes of those arrays are pretty weird though.

Which fail for you? I'm only seeing `split` tests failing (not `array_split`), and it looks like the problem there is that either version would error.

stuartarchibald: Seems like these fail:
```
(array([], dtype=float64), 1)
NB Maximum allowed size exceeded
(array([], dtype=float64), 2)
NB Maximum allowed size exceeded
(array([], dtype=float64), 1, 0)
NB Maximum allowed size exceeded
(array([], shape=(1, 0), dtype=float64), 1, 1)
NB Maximum allowed size exceeded
```
under `np.split`, I think the issue is here:
https://github.com/numba/numba/pull/5699/files#diff-ef28bfb9dfcf3597915208b3329a6ac6R5046
specifically:
```
np.arange(l, ary.shape[axis], l),
```
if `l` is zero then it leads to a zero step size in the `arange` which is where that error message is coming from.

> The shapes of those arrays are pretty weird though.

yeah, this is why I'm thinking it might be a side effect opposed to intended? What do you think?

numba/np/arrayobj.py
stuartarchibald: thanks

numba/tests/test_np_functions.py
ivirshup: Ah, found it. I was doing something wrong with `split` and have added these test cases. It's possible the implementation could be cleaner if I just passed the input integer after the check (leading to an additional `divmod` call). 

Is the compiler smart enough to skip the additional `divmod` call if that approach were taken?

stuartarchibald: > Ah, found it. I was doing something wrong with `split` and have added these test cases. It's possible the implementation could be cleaner if I just passed the input integer after the check (leading to an additional `divmod` call).

Great, thanks.

> Is the compiler smart enough to skip the additional `divmod` call if that approach were taken?

Can you show me what you mean please? Where would the `divmod` end up? I can probably make a guess from there :)

numba/np/arrayobj.py
```diff
@@ -5042,8 +5042,12 @@ def impl(ary, indices_or_sections, axis=0):
                 raise ValueError(
                     "array split does not result in an equal division"
                 )
+            indices = np.cumsum(np.array(
+                [l + 1] * rem +
+                [l] * (indices_or_sections - rem - 1)
+            ))
             return np.array_split(
-                ary, np.arange(l, ary.shape[axis], l), axis=axis
+                ary, indices, axis=axis
```
ivirshup: This is the same code as in `np.array_split` when an integer is passed for indices or sections. If I were to pass the `indices_or_sections` to `np.array_split` instead, would the compiler know it could reuse the results of the divmod call on 5040? This is with the assumption everything gets inlined.

numba/tests/test_np_functions.py
ivirshup: > Can you show me what you mean please? Where would the divmod end up? I can probably make a guess from there :)

Ah, yes, I commented before I realized the tests hadn't finished running. Here's what I'm talking about: https://github.com/numba/numba/pull/5699/commits/93a9218f01c79ca00ea8f0b906f795bb3e2887d2#r500160142

> yeah, this is why I'm thinking it might be a side effect opposed to intended?

I think this might be like passing indices that aren't monotonic. I can't imagine a case where not throwing an error is what someone wants, but it is the behavior numpy has. I've got that behavior replicating now though.

numba/np/arrayobj.py
stuartarchibald: Quickly looking at the assembly suggests it's struggling to inline `array_split` and so doesn't spot the `idiv` on the same vars.

stuartarchibald:
Ignore the flake8 fail (the CI run with `_cov` on the end), mainline was broken due to a bad merge interaction, https://github.com/numba/numba/pull/6306 fixed just now.

numba/np/arrayobj.py
stuartarchibald: Looking at this some more, I think passing through `indices_or_sections` is probably a good thing on two fronts, 1. cleans the code up a bit, 2. eventually the compiler might inline this to avoid the `divmod` but more importantly, passing it through avoids creating heap objects like arrays and lists when forming up `indices`. Are you happy to just pass it?

ivirshup: Sure! Done now.

numba/tests/test_np_functions.py
stuartarchibald: Great, thank you.

ivirshup:
Great, thanks!

jaistech-sp:
Thanks a lot for pushing it!! Go numba!

PR #3437, opened 2018-10-23
Changes to accommodate LLVM 7.0.x

stuartarchibald:
As title.

PR #5702, opened 2020-05-13
Fixes for LLVM 9

sklam:


numba/core/codegen.py
```diff
@@ -829,6 +829,8 @@ def _customize_tm_options(self, options):
         arch = ll.Target.from_default_triple().name
         if arch.startswith('x86'): # one of x86 or x86_64
             reloc_model = 'static'
+        elif arch.startswith('ppc'):
+            reloc_model = 'pic'
```
stuartarchibald, line 833: Perhaps add a note here about how this is entirely different to that in LLVM's `getEffectiveRelocModel` etc itself, but seems to work. I had another vague recollection about this yesterday regarding PIC leading to giant amounts of mapped memory on PPC for LLVM 7/8, worth taking a look?

numba/core/runtime/nrtopt.py
```diff
@@ -9,7 +9,17 @@
...
-    r'([\'"]?[-a-zA-Z$._][-a-zA-Z$._0-9]*[\'"]?:)|^define|^;\s*<label>')
+    r'|'.join([
+        # unamed BB is just a plain number
+        r'[0-9]+:',
+        # with a proper identifer (see llvm langref)
+        r'[\'"]?[-a-zA-Z$._0-9][-a-zA-Z$._0-9]*[\'"]?:',
+        # is a start of a function definition
+        r'^define',
+        # no name
+        r'^;\s*<label>',
```
stuartarchibald, line 20: thanks for adding these comments

sklam:
Patch up CI to use `.label`

numba/core/codegen.py
sklam: I ran the full testsuite on PPC and it passes basically everything. Is that a good sign?

sklam:
linux & mac has passed CI
windows llvmlite build is stuck.
not yet have aarch64 llvmdev
ppc64le have passed manual testing of numba

sklam:
```Invalid PPC CTR loop!
[2020-05-29 15:25:33,845] {docker_operator.py:262} INFO - UNREACHABLE executed at /opt/conda/conda-bld/llvmdev_1583256883192/work/lib/Target/PowerPC/PPCCTRLoops.cpp:214!
[2020-05-29 15:25:33,845] {docker_operator.py:262} INFO - Fatal Python error: Aborted

numba/core/codegen.py
stuartarchibald: Yes!

sklam:
Reverted changes that made this PR test against LLVM 9.

sklam:
This PR needs to be merged as soon as https://github.com/numba/llvmlite/pull/548 is merged

stuartarchibald:
Think this needs a change similar to https://github.com/numba/numba/pull/3437/files#diff-caa68793dddeb30ff0cf094862c99d17R61-R62 making?

PR #5704, opened 2020-05-14
CUDA Eager compilation: Fix max_registers kwarg

gmarkall:
The `max_registers` kwarg to `@cuda.jit` was getting dropped for eager compilation. This PR fixes it by passing the `max_registers` kwarg, if present, to `compile_kernel` when eagerly compiling.

### DIFF
--- a/numba/cuda/decorators.py
+++ b/numba/cuda/decorators.py
@@ -91,9 +91,12 @@ def jit(func_or_sig=None, argtypes=None, device=False, inline=False, bind=True,
         if restype and not device and restype != types.void:
             raise TypeError("CUDA kernel must have void return type.")
 
+        max_registers = kws.get('max_registers', None)
+
         def kernel_jit(func):
             kernel = compile_kernel(func, argtypes, link=link, debug=debug,
-                                    inline=inline, fastmath=fastmath)
+                                    inline=inline, fastmath=fastmath,
+                                    max_registers=max_registers)
 
             # Force compilation for the current context
             if bind:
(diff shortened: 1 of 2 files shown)


stuartarchibald:
Thanks for the patch. Whilst looking at this part of the code, are there any other kwargs that are not being forwarded correctly?

gmarkall:
From inspection, it looks to me as if all documented parameters are getting passed through now.

It looks as though extensions might not be supported for eagerly-compiled kernels (`CUDAKernel` has no `extensions` property, but `AutoJitCUDAKernel` does, which refers to the `targetoptions` passed through from the `jit` decorator). I think that's a bit orthogonal to this PR, because it would need support for extensions adding to `CUDAKernel` as well.

stuartarchibald:
> From inspection, it looks to me as if all documented parameters are getting passed through now.

Thanks for looking.

> It looks as though extensions might not be supported for eagerly-compiled kernels (`CUDAKernel` has no `extensions` property, but `AutoJitCUDAKernel` does, which refers to the `targetoptions` passed through from the `jit` decorator). I think that's a bit orthogonal to this PR, because it would need support for extensions adding to `CUDAKernel` as well.

Agree, out of scope for this PR.

stuartarchibald:
Buildfarm ID: `numba_smokestest_cuda_50`.

gmarkall:
I did later notice that the `link` kwarg is dropped when you don't provide a signature - this whole area could do with a tidy up, so I'm working to try and clean up the interaction between `cuda.jit` and `CUDAKernel` etc., in the hope of making it a bit easier to avoid missing these things in future - may follow up with another PR for that.

stuartarchibald:
> I did later notice that the `link` kwarg is dropped when you don't provide a signature - this whole area could do with a tidy up, so I'm working to try and clean up the interaction between `cuda.jit` and `CUDAKernel` etc., in the hope of making it a bit easier to avoid missing these things in future - may follow up with another PR for that.

I see, a clean up sounds good!

stuartarchibald:
Buildfarm passed.

PR #5709, opened 2020-05-15
CUDA: Refactoring of cuda.jit and kernel / dispatcher abstractions

gmarkall:
This PR is the first towards enabling the CUDA target to accept multiple signatures in the `@cuda.jit` decorator - the aim is for the CUDA and CPU dispatchers to share common common infrastructure and base classes. This PR changes the structure of the CUDA target to more closely align with the CPU target, but does not factor the two together as yet - following this PR, the next step is to modify the C-level dispatcher object in `numba/_dispatcher.c` so that it can be used as a common base for the CPU and CUDA dispatchers - the logic it uses to find compatible definitions when the types of supplied arguments don't exactly match any specialization will be used in support of multiple signatures.

The changes in this PR are broadly:

- Move the `AutoJitCUDAKernel` to `Dispatcher` since it contains mostly dispatcher-like functionality. `CUDAKernel` is renamed to `_Kernel`.
- `Dispatcher` / `AutoJitCUDAKernel` and `CUDAKernel` / `_Kernel` no longer share a common base class - I always found it a bit strange that they shared some infrastructure. Now the `Dispatcher` object mainly takes responsibility for handling signatures and compilation, and the `_Kernel` object is responsible for launches, via the `Dispatcher`.
- The launch configuration is moved out of both of these - configuring the `Dispatcher` gives a new `_KernelConfiguration` object instead - baking the configuration into a dispatcher object seemed/seems to conflate different concerns.
- A little extra new logic in the `Dispatcher` so that specializing it always returns another `Dispatcher` instance that has been specialized for the given arguments, rather than a kernel object - this ensures that the user always deals with `Dispatcher` objects, rather than getting a mix of `Dispatcher` and `_Kernel` objects as would have otherwise occurred.
- Changes to the `Dispatcher` so that all the existing APIs work consistently with previous versions - in particular, functions like `inspect_llvm` had slightly different behaviour between `CUDAKernel` and `AutoJitCUDAKernel`, so these functions in the `Dispatcher` class need to behave like the old `CUDAKernel` when specialized, and like the `AutoJitCUDAKernel` when not. This does make things a little bit messy and complicated, unfortunately.
- Deprecation of the `bind` and `argtypes` kwargs for the `@cuda.jit` decorator - `bind` is likely to be always true in practice for eager compilation, and doesn't make sense as far as I can tell for lazy compilation. `argtypes` is already deprecated in the CPU target - the way forward will be for a signature or list of signatures to be accepted as in the CPU target.
- Removal of the `CUDADispatcher` class, and use of the `Dispatcher` object in its place. The `Dispatcher` class seemed to subsume everything that the `CUDADispatcher` did.
- Documentation updates to mirror the proposed changes.


gmarkall:
Updated title and description now that this is ready for review.

sklam:
There's conflict with master. 

The PR looks good. It's nice that the API did not change and the tests all work as is.

gmarkall:
Thanks @sklam! I've now merged master and resolved the conflicts.

esc:
BFID: `numba_smoketest_cuda_80`

PR #5716, opened 2020-05-18
Docs: Add example usage of List.empty_list

gmarkall:
Also some minor edits to examples/typed_list_usage.py to fix flake8 errors, that prevent committing when a pre-commit hook is used.

This came up in https://gitter.im/numba/numba?at=5ec146b50e8a3131e4c8722e, and I noticed that `empty_list` doesn't seem to be documented anywhere.

### DIFF
--- a/docs/source/reference/pysupported.rst
+++ b/docs/source/reference/pysupported.rst
@@ -574,13 +574,24 @@ function:
    :dedent: 4
    :linenos:
 
-Finally, here's an example of using a nested `List()`:
+Here's an example of using a nested ``List()``:
 
 .. literalinclude:: ../../../examples/typed_list_usage.py
    :language: python
    :caption: from ``ex_nested_list`` of ``examples/typed_list_usage.py``
    :start-after: magictoken.ex_nested_list.begin
    :end-before: magictoken.ex_nested_list.end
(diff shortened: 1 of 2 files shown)


esc:
@gmarkall thanks for submitting this. We concluded, that `empty_list` is somewhat deprecated and was only a crutch of sorts to get the `numba.typed.List` up and running. The proposed new API will include the type in the `List()` constructor for a more pythonic approach. The constructor would be available then with the following signatures:

```
l = List()                 # empty list with undefined type
l = List(type=int64)       # empty list with defined type
l = List(iter)             # list initialized from iter
l = List(iter, type=int64) # list initialized from iter and with items cast to given type
```

This appears to be feasible to implement  with the recent fixes to the `List()` constructor. Does that make sense?

gmarkall:
It makes sense. Do you want the PR closing? I'm not too clear about whether it's more appropriate to document the only way to do this now, or to hide it and not bless construction of empty sublists until the new API is implemented.

esc:
@gmarkall Ideally and if possible/agreeable I'd prefer to not document this now and set this PR to abandoned. I'd be swayed though, if you feel strongly that this has value.

PR #5717, opened 2020-05-18
Initialise Numba extensions ahead of any compilation starting.

### LINKED ISSUE #5284: Extension types loaded via entry points are registered too late
If an extension project uses [entry points](https://numba.pydata.org/numba-doc/latest/extending/entrypoints.html) and the first function to be JIT-compiled takes that project's custom types as arguments, Numba will not recognize the custom types (error message is "non-precise type pyobject").

If the custom types are used in the second function to be compiled (i.e. start by compiling a dummy, `numba.njit(lambda x: x + 1)(123)`), then the custom types are correctly recognized (no "non-precise type pyobject" error).

If the custom types are registered with

```python
@numba.extending.typeof_impl.register(custom_python_type)
def typeof(val, c):
    return custom_numba_type()
```

before the entry point gets triggered (i.e. before compiling a dummy), then the custom types are correctly recognized.

I see that the extension project used as an entry point example, [scipy-numba](https://github.com/numba/numba-scipy), doesn't define any custom types (only functions), so it hasn't encountered this issue.

I can still put most of my type and lowering definitions behind the entry point, since `numba.extending.typeof_impl.register` only has to reference the custom types inside a function definition, which is delayed. However, the register decorator itself requires Numba to be loaded, so I can't escape the Numba load-time in processes where Numba is not actually used. (I'm avoiding any time involved in constructing my custom types, but it's more the LLVM-load that I was trying to avoid.)

Since this involves setting up entry points, I don't see how to make a minimal reproducer that's really "minimal," but I'm willing to be a guinea pig for proposed fixes.

stuartarchibald:
As title. This patch forces Numba extensions that are registering
via the Numba entry point mechanism to be discovered and loaded
ahead of any part of the compilation chain being invoked. This
means custom types get registered sufficiently early to now be used
as part of type inference.

Fixes #5284

### DIFF
--- a/numba/tests/test_entrypoints.py
+++ b/numba/tests/test_entrypoints.py
@@ -1,10 +1,25 @@
 import sys
 import types
 import warnings
-
+import unittest
+import os
+import subprocess
+import threading
 import pkg_resources
 
+from numba import njit
 from numba.tests.support import TestCase
+from numba.testing.main import _TIMEOUT as _RUNNER_TIMEOUT
+
+_TEST_TIMEOUT = _RUNNER_TIMEOUT - 60.
+
+
+class _DummyClass(object):
+    def __init__(self, value):
+        self.value = value
+
+    def __repr__(self):
+        return '_DummyClass(%f, %f)' % self.value
 
 
 class TestEntrypoints(TestCase):
@@ -35,9 +50,9 @@ class TestEntrypoints(TestCase):
             dist = "numba"
             entrypoints = pkg_resources.get_entry_map(dist)
             my_entrypoint = pkg_resources.EntryPoint(
-                "init", # name of entry point
-                mod.__name__, # module with entry point object
-                attrs=['init_func'], # name of entry point object
+                "init",  # name of entry point
+                mod.__name__,  # module with entry point object
+                attrs=['init_func'],  # name of entry point object
                 dist=pkg_resources.get_distribution(dist)
             )
             entrypoints.setdefault('numba_extensions',
@@ -84,9 +99,9 @@ class TestEntrypoints(TestCase):
             dist = "numba"
             entrypoints = pkg_resources.get_entry_map(dist)
             my_entrypoint = pkg_resources.EntryPoint(
-                "init", # name of entry point
-                mod.__name__, # module with entry point object
-                attrs=['init_func'], # name of entry point object
+                "init",  # name of entry point
+                mod.__name__,  # module with entry point object
+                attrs=['init_func'],  # name of entry point object
                 dist=pkg_resources.get_distribution(dist)
             )
             entrypoints.setdefault('numba_extensions',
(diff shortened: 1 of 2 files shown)


jpivarski:
I'm setting up for a test. I have a mostly empty conda environment with essentials like numpy, pytest, compilers, and llvmite, but no numba. I checked out branch fix/5284 from stuartarchibald/numba and installed it into this conda environment with

```bash
pip install .
```

The compilation was surprisingly fast (less than 10 seconds, I think) and the version number is surprisingly low: `0.35.0.dev0+7221.g16f0b30f0`. I'm unsure that I have the right one. I'm going ahead with the test anyway. (If it's really 0.35, it won't work at all because I strictly depend on the changes that went into 0.49.)

stuartarchibald:
> the version number is surprisingly low: 0.35.0.dev0+7221.g16f0b30f0

7k commits after 0.35.... I think I might need to sync up tags on my fork! Have done that now!

jpivarski:
It is the right version because (commenting out a check for Numba version number), all of my tests pass when I register the ak.Array type with Numba at startup, as I always have been. So the content of this PR is definitely not pre-0.49.

This PR does fix the issue. **Without it,** Numba does not recognize Awkward types on the first try, but does on the second:

```
% python -i -c 'import awkward1, numba; f = numba.njit(lambda x: x[2]); print(f(awkward1.Array([[1.1, 2.2, 3.3], [], [3.3, 4.4]])))'
  File "<string>", line 1, in <module>
  File "/home/pivarski/miniconda3/lib/python3.7/site-packages/numba/core/dispatcher.py", line 401, in _compile_for_args
    error_rewrite(e, 'typing')
  File "/home/pivarski/miniconda3/lib/python3.7/site-packages/numba/core/dispatcher.py", line 344, in error_rewrite
    reraise(type(e), e, None)
  File "/home/pivarski/miniconda3/lib/python3.7/site-packages/numba/core/utils.py", line 79, in reraise
non-precise type pyobject
[1] During: typing of argument at <string> (1)

File "<string>", line 1:

>>> print(f(awkward1.Array([[1.1, 2.2, 3.3], [], [3.3, 4.4]])))
[3.3, 4.4]
```

**With this PR,** Numba recognizes the Awkward types on the first try (and the second):

```
% python -i -c 'import awkward1, numba; f = numba.njit(lambda x: x[2]); print(f(awkward1.Array([[1.1, 2.2, 3.3], [], [3.3, 4.4]])))'
[3.3, 4.4]
>>> print(f(awkward1.Array([[1.1, 2.2, 3.3], [], [3.3, 4.4]])))
[3.3, 4.4]
```

In these tests, I was careful to uninstall, reinstall, and work with the installed version of Awkward, because it's the pip-install process that puts the entry point where it needs to be and I had to change the Awkward entry point to register the Awkward types only when triggered by it (instead of at Awkward import time).

But it's my conclusion that it works, and scikit-hep/awkward-1.0#270 is eagerly awaiting Numba 0.50. Thanks!

stuartarchibald:
Brilliant, thanks so much for your efforts in testing this and confirming that it does indeed fix the problem. Will try and work out if there's anything that can be done to test this to make sure it doesn't accidentally get broken. I've marked this for 0.50, I can't imagine there will be any objections!

jpivarski:
I think the dummy-numba-entry-points.zip on #5717 could be turned into a standalone test to prevent regressions. That module encapsulates the whole problem. As far as I see it, checking on Awkward was a sanity-check.

stuartarchibald:
3fb93d4 does some somewhat horrible things to make it possible to test this, I tried on master and the same tests fail, but with this patch they seem ok.

sklam:
Trying this on the farm: `numba_smoketest_cpu_38`

numba/tests/test_entrypoints.py
```diff
@@ -113,3 +128,119 @@ def init_function():
             # remove fake module
             if mod.__name__ in sys.modules:
                 del sys.modules[mod.__name__]
+
+    _EP_MAGIC_TOKEN = 'RUN_ENTRY'
+
+    @unittest.skipIf(os.environ.get('_EP_MAGIC_TOKEN', None) != _EP_MAGIC_TOKEN,
```
sklam, line 134: When/where should we turn this on?

sklam: oh nevermind

sklam:
The new tests failed on buildfarm.

sklam:
retry `numba_smoketest_cpu_40`

sklam:
buildfarm failed on windows for the following tests:

```
FAIL: test_entrypoint_tolerance (numba.tests.test_entrypoints.TestEntrypoints)
FAIL: test_init_entrypoint (numba.tests.test_entrypoints.TestEntrypoints)
FAIL: test_entrypoint_extension_sequence (numba.tests.test_entrypoints.TestEntrypoints)
ERROR: test_entrypoint_handles_type_extensions (numba.tests.test_entrypoints.TestEntrypoints)
```

stuartarchibald:
Thanks for testing. Strange it's just windows?! Will take a look in the logs.

stuartarchibald:
> buildfarm failed on windows for the following tests:
> 
> ```
> FAIL: test_entrypoint_tolerance (numba.tests.test_entrypoints.TestEntrypoints)
> FAIL: test_init_entrypoint (numba.tests.test_entrypoints.TestEntrypoints)
> FAIL: test_entrypoint_extension_sequence (numba.tests.test_entrypoints.TestEntrypoints)
> ERROR: test_entrypoint_handles_type_extensions (numba.tests.test_entrypoints.TestEntrypoints)
> ```

Buried in the logs is:
```
UserWarning: Numba extension module '_test_numba_extension' failed to load due to 'VersionConflict((llvmlite 0.33.0dev0+17.gc4f5a08.dirty (<path>), Requirement.parse('llvmlite<=0.33.0dev0,>=0.31.0dev0'))
```
I think this is happening because the windows builds aren't containerized so they are picking up llvmlite builds from the local conda build cache.

jpivarski:
Just a quick note for posterity: this fix works in the newly released Numba 0.50:

```python
>>> import sys
>>> "numba" in sys.modules.keys()
False
>>> import awkward1
>>> "numba" in sys.modules.keys()       # importing awkward1 does not import numba
False
>>> import numba
>>> f = numba.njit(lambda x: x[2])      # but the first JIT'ed function works
>>> f(awkward1.Array([[1.1, 2.2, 3.3], [], [3.3, 4.4]]))
<Array [3.3, 4.4] type='2 * float64'>
```

I'm making Numba 0.50 the minimum version, starting with [Awkward 0.2.24](https://github.com/scikit-hep/awkward-1.0/releases/tag/0.2.24).

stuartarchibald:
Brilliant, thanks for confirming @jpivarski!

PR #5718, opened 2020-05-18
CUDA lib path tests: unset CUDA_PATH when CUDA_HOME unset

gmarkall:
The library lookup mechanism in `numba.cuda.cuda_paths.get_cuda_home()` tries `CUDA_PATH` if `CUDA_HOME` is unset. This causes a problem for the test_library_lookup tests if `CUDA_PATH` is set, because it unsets `CUDA_HOME` then expects not to find the CUDA toolkit by the `CUDA_PATH` method.

The `CUDA_PATH` variable can sometimes be set on Windows, or on other platforms where the conda-forge CuPy package is installed - it adds an activation script to the conda environment that sets `CUDA_PATH` to point to the conda environment.

This PR rectifies the issue by changing the tests so that they also unset `CUDA_PATH` when `CUDA_HOME` is unset.

### DIFF
--- a/numba/cuda/tests/nocuda/test_library_lookup.py
+++ b/numba/cuda/tests/nocuda/test_library_lookup.py
@@ -129,6 +129,7 @@ class TestLibDeviceLookUp(LibraryLookupBase):
     @staticmethod
     def do_clear_envs():
         remove_env('CUDA_HOME')
+        remove_env('CUDA_PATH')
         remove_env('NUMBAPRO_CUDALIB')
         remove_env('NUMBAPRO_LIBDEVICE')
         return True, _get_libdevice_path_decision()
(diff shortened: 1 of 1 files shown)


leofang:
> The `CUDA_PATH` variable can sometimes be set on Windows, or on other platforms where the conda-forge CuPy package is installed - it adds an activation script to the conda environment that sets `CUDA_PATH` to point to the conda environment.

This was done in this PR: https://github.com/conda-forge/cupy-feedstock/pull/49. Thanks for addressing this, @gmarkall!

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_56`.

stuartarchibald:
Buildfarm passed.

PR #5721, opened 2020-05-18
Fix array iterator layout.

### LINKED ISSUE #5692: issue with >1d array iteration
numba 0.49.0
numpy 1.18.2
Python 3.8.2 (default, May  7 2020, 20:00:49) [GCC 7.3.0] on linux

```python
import numba as nb
import numpy as np


@nb.njit()
def outer_product_sum(x: np.ndarray, y: np.ndarray) -> np.ndarray:
    assert x.ndim == 2, "dimensions of both arrays must be 2"
    assert y.ndim == 2, "dimensions of both arrays must be 2"
    assert x.shape[1] == y.shape[1], "The second dimension of both arrays must be equal"

    o = np.zeros((x.shape[0], y.shape[0]))
    for v1, v2 in zip(x.T, y.T):
        o += np.outer(v1, v2)
    return o

x = np.array([[1, 8, 9, 1], [4, 64, 27, 1]])
y = np.array([[1, 1/2, 1/3, 1],  [1/2, 1/4, 1/9, 1], [1/2, 1/4, 1/9, 1]])
expected = np.outer(x[:,0], y[:,0]) + np.outer(x[:,1], y[:,1]) + np.outer(x[:,2], y[:,2]) + np.outer(x[:,3], y[:,3])
actual = outer_product_sum(x, y)
assert np.isclose(expected, actual).all()
```

Without the decorator, this test passes, but with the decorator, it fails. On inspection, it is clear that the issue is not with the sum, but with the inner product. In this context, `np.outer` is not producing correct results. `v1`, `v2` are as expected, but `np.outer` no longer performs the outer product.

stuartarchibald:
As title. Fixes #5692

### DIFF
--- a/numba/core/types/iterators.py
+++ b/numba/core/types/iterators.py
@@ -103,6 +103,6 @@ class ArrayIterator(SimpleIteratorType):
         elif nd == 1:
             yield_type = array_type.dtype
         else:
-            # iteration semantics leads to C order layout
-            yield_type = array_type.copy(ndim=array_type.ndim - 1, layout='C')
+            # iteration semantics leads to A order layout
+            yield_type = array_type.copy(ndim=array_type.ndim - 1, layout='A')
         super(ArrayIterator, self).__init__(name, yield_type)
--- a/numba/core/types/iterators.py
+++ b/numba/core/types/iterators.py
@@ -103,5 +103,6 @@ class ArrayIterator(SimpleIteratorType):
         elif nd == 1:
             yield_type = array_type.dtype
         else:
-            yield_type = array_type.copy(ndim=array_type.ndim - 1)
+            # iteration semantics leads to C order layout
+            yield_type = array_type.copy(ndim=array_type.ndim - 1, layout='C')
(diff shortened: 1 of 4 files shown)


sklam:
Note: are there more implementation issues? like what if inner axis are strided?

stuartarchibald:
> Note: are there more implementation issues? like what if inner axis are strided?

Added some tests in e4984d2 for outer, middle and inner stride.

numba/core/types/iterators.py
```diff
@@ -103,5 +103,6 @@ def __init__(self, array_type):
         elif nd == 1:
             yield_type = array_type.dtype
         else:
-            yield_type = array_type.copy(ndim=array_type.ndim - 1)
+            # iteration semantics leads to C order layout
+            yield_type = array_type.copy(ndim=array_type.ndim - 1, layout='C')
```
sklam, line 107: setting layout to `'A'` is probably the safest fix.

stuartarchibald: Done, also added the example as a test.

Issue #4348, opened 2019-07-20
Global locks cause semaphore leaks.


PetrochukM:
#### Description

Importing `numba` causes semaphores to leak.

#### Steps/Code to Reproduce

```python
import multiprocessing as mp

import numba


def w():
    pass


def main():
    ps = [mp.Process(target=w) for _ in range(4)]
    [p.start() for p in ps]
    [p.join() for p in ps]


if __name__ == '__main__':
    p = mp.get_context('spawn').Process(target=main)
    p.start()
    p.join()
```

#### Actual Results

```bash
$ python3 blah.py
/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/semaphore_tracker.py:144: UserWarning: semaphore_tracker: There appear to be 1 leaked semaphores to clean up at shutdown
  len(cache))
/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/semaphore_tracker.py:144: UserWarning: semaphore_tracker: There appear to be 1 leaked semaphores to clean up at shutdown
  len(cache))
/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/semaphore_tracker.py:144: UserWarning: semaphore_tracker: There appear to be 1 leaked semaphores to clean up at shutdown
  len(cache))
/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/semaphore_tracker.py:144: UserWarning: semaphore_tracker: There appear to be 1 leaked semaphores to clean up at shutdown
  len(cache))
$
```

#### Expected Results

```bash
$ python3 blah.py
$
```

PetrochukM:
This issue is causing a number of downstream issues:
- https://github.com/pytorch/pytorch/issues/23117
- https://github.com/librosa/librosa/issues/924

Also, this issue came up with `tqdm`: https://github.com/tqdm/tqdm/issues/611

stuartarchibald:
Thanks for the report, I can reproduce.

The reason for the lock here:
https://github.com/numba/numba/blob/37ab1420fa87d02b635991b32f0faf5c65e8fc4d/numba/npyufunc/parallel.py#L265
[...]

PR #5724, opened 2020-05-19
Initialize process lock lazily to prevent multiprocessing issue

belltailjp:
## Background and about this PR

Due to the side-effect made by `multiprocessing.get_context` which is called at numba import time [here](https://github.com/numba/numba/blob/f3ca34da03b03c62a27daacfbddf63f7d72e7cee/numba/np/ufunc/parallel.py#L296), some of operations like `multiprocessing.set_start_method` raise the following error.
```python
>>> import numba
>>> import multiprocessing
>>> multiprocessing.set_start_method('forkserver')
  File "/home/xxxx/.pyenv/versions/miniconda3-4.7.12/lib/python3.7/multiprocessing/context.py", line 242, in set_start_method
    raise RuntimeError('context has already been set')
RuntimeError: context has already been set
```
This PR is the fix for this issue.

### Reason

There are 2 facts in background.
* (1) `multiprocessing` module requires to set process start method by `set_start_method` at most once during the master process lifetime.
* (2) Calling `multiprocessing.get_context` with the default argument globally sets the default one (`"fork"`) implicitly, if nothing has been set until then.
    * https://github.com/python/cpython/blob/3.7/Lib/multiprocessing/context.py#L232-L238

In the current Numba this runs at import time, which means that once we import numba the process start method is already set implicitly (Fact (2)). Therefore users can no longer change the process launch method of multiprocessing module anymore (Fact (1)), regardless of what mode the user actually wants to specify (either "fork", "forkserver" or "spawn").

Here is the minimum simulation of what is happening inside.
```python
>>> import multiprocessing
>>> cxt = multiprocessing.get_context()
>>> multiprocessing.set_start_method('forkserver')
...
RuntimeError: context has already been set

>>> multiprocessing.set_start_method('fork')
...
RuntimeError: context has already been set
```

### Workaround

We can prevent this error by making sure to call `set_start_method` _before_ importing numba.
This means that we always have to be very careful of import order (and possible violation of pep8/flake8) if there is a specific need of changing process start mode.

In this PR I tried to solve this by eliminating the side-effect in import time, by initializing `_backend_init_process_lock` lazily when it's necessary.

### Effect & Confirmation of this PR

You can reproduce and confirm the effect of this PR by checking if `set_start_method` raises the error.

```python
>>> import numba
>>> import multiprocessing
>>> multiprocessing.set_start_method('forkserver')
(No error will occur after this patch is applied)
```

## Testing concerns

I'm quite new to Numba so I am unfamiliar with where and how this should be tested.
I thought that this PR intends to exactly keep the current behavior so confirmation of existing test cases would work, but if there is some suggestions, I'm happy to make any further efforts.

### DIFF
--- a/numba/tests/test_parallel_backend.py
+++ b/numba/tests/test_parallel_backend.py
@@ -862,33 +862,33 @@ class TestInitSafetyIssues(TestCase):
 
     _DEBUG = False
 
+    def run_cmd(self, cmdline):
+        popen = subprocess.Popen(cmdline,
+                                    stdout=subprocess.PIPE,
+                                    stderr=subprocess.PIPE,)
+        # finish in _TEST_TIMEOUT seconds or kill it
+        timeout = threading.Timer(_TEST_TIMEOUT, popen.kill)
+        try:
+            timeout.start()
+            out, err = popen.communicate()
+            if popen.returncode != 0:
+                raise AssertionError(
+                    "process failed with code %s: stderr follows\n%s\n" %
+                    (popen.returncode, err.decode()))
+        finally:
+            timeout.cancel()
+        return out.decode(), err.decode()
+
+
     @linux_only # only linux can leak semaphores
     def test_orphaned_semaphore(self):
         # sys path injection and separate usecase module to make sure everything
         # is importable by children of multiprocessing
 
-        def run_cmd(cmdline):
-            popen = subprocess.Popen(cmdline,
-                                     stdout=subprocess.PIPE,
-                                     stderr=subprocess.PIPE,)
-            # finish in _TEST_TIMEOUT seconds or kill it
-            timeout = threading.Timer(_TEST_TIMEOUT, popen.kill)
-            try:
-                timeout.start()
-                out, err = popen.communicate()
-                if popen.returncode != 0:
-                    raise AssertionError(
-                        "process failed with code %s: stderr follows\n%s\n" %
-                        (popen.returncode, err.decode()))
-            finally:
-                timeout.cancel()
-            return out.decode(), err.decode()
-
         test_file = os.path.join(os.path.dirname(__file__),
                                  "orphaned_semaphore_usecase.py")
-
         cmdline = [sys.executable, test_file]
-        out, err = run_cmd(cmdline)
+        out, err = self.run_cmd(cmdline)
 
         # assert no semaphore leaks reported on stderr
         self.assertNotIn("leaked semaphore", err)
@@ -896,6 +896,27 @@ class TestInitSafetyIssues(TestCase):
         if self._DEBUG:
             print("OUT:", out)
             print("ERR:", err)
+
+    def test_lazy_lock_init(self):
+        # checks based on https://github.com/numba/numba/pull/5724
+        # looking for "lazy" process lock initialisation so as to avoid setting
+        # a multiprocessing context as part of import.
+        for meth in ('fork', 'spawn', 'forkserver'):
+            # if a context is available on the host check it can be set as the
+            # start method in a separate process
+            try:
+                multiprocessing.get_context(meth)
+            except ValueError:
+                continue
+            cmd = ("import numba; import multiprocessing;"
+                   "multiprocessing.set_start_method('{}');"
+                   "print(multiprocessing.get_context().get_start_method())")
+            cmdline = [sys.executable, "-c", cmd.format(meth)]
+            out, err = self.run_cmd(cmdline)
+            self.assertIn(meth, out)
+            if self._DEBUG:
+                print("OUT:", out)
+                print("ERR:", err)
 
 
 @skip_parfors_unsupported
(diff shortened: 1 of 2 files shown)


stuartarchibald:
Thanks for the PR, just on a quick look, this seems to fix the described. Now CI is passing I'll queue it for review.

stuartarchibald:
xref : https://github.com/numba/numba/issues/4348#issuecomment-513760791

stuartarchibald:
@belltailjp apologies for the delay in getting to this. Thanks for fixing this issue, much appreciated. I've taken a look at the patch and I think it will work, but I'd like to push it through the build farm to make sure. As to testing, I've been using this patch https://github.com/stuartarchibald/numba/commit/844ec25693ceb0183a7e270be5f4913138efcfe2 if you think it's suitable please could you add it to your branch so that this PR can be tested further. Many thanks.

esc:
BFID:`numba_smoketest_cpu_57`

esc:
@belltailjp thanks again for submitting this. I have scheduled for a test run on our internal build farm, hence the comment above. It seems like public CI is complaining about some flake8 errors that will  need to be addressed:

```
numba/tests/test_parallel_backend.py:867:37: E127 continuation line over-indented for visual indent
numba/tests/test_parallel_backend.py:868:37: E127 continuation line over-indented for visual indent
numba/tests/test_parallel_backend.py:883:5: E303 too many blank lines (2)
````

**Comment by belltailjp** [comment]
Oops! Thanks for pointing it out, I forgot to care the CI result. Now it's all-green.

**Comment by esc** [comment]
The buildfarm run was fine, except for the same `aarch64` bug as #5980 and #6035 -- it'll probably be fine though.

**Comment by stuartarchibald** [comment]
Thanks for fixing `flake8` @belltailjp.

**Comment by stuartarchibald** [comment]
@esc Thanks for running through the farm, I think this patch is good but it now contains my test code, could you please check that part and approve if you see fit? Thanks.

File: numba/tests/test_parallel_backend.py [hunk 1]
```diff
@@ -862,33 +862,32 @@ class TestInitSafetyIssues(TestCase):
 
     _DEBUG = False
 
+    def run_cmd(self, cmdline):
```
esc, line 865: I just looked at the documentation for `subprocess` for Python 3.6 (the minimal version supported by Numba) and it turns out that the `run` function is equipped with a `timeout` argument. I believe the construct below using `Popen` was written for an older version of Python and the `subprocess` module. Perhaps `run_cmd` can now be replaced with a more modern variant using the `run` function and the `timeout` argument? I think it would shorten the code significantly.

numba/tests/test_parallel_backend.py
```diff
@@ -897,6 +896,27 @@ def run_cmd(cmdline):
...
+                multiprocessing.get_context(meth)
+            except ValueError:
+                continue
+            cmd = ("import numba; import multiprocessing;"
+                   "multiprocessing.set_start_method('{}');"
+                   "print(multiprocessing.get_context().get_start_method())")
+            cmdline = [sys.executable, "-c", cmd.format(meth)]
+            out, err = self.run_cmd(cmdline)
+            self.assertIn(meth, out)
+            if self._DEBUG:
```
esc, line 916: Perhaps the following print statements should be placed before the call to `assertIn`? This way, the values will be printed also if the test fails?

numba/tests/test_parallel_backend.py
stuartarchibald: Good point, am going to defer to https://github.com/numba/numba/issues/5989 though if that's ok as we know this pattern works for our, often strange, use cases.

numba/tests/test_parallel_backend.py
```diff
@@ -897,6 +896,27 @@ def run_cmd(cmdline):
...
+            try:
+                multiprocessing.get_context(meth)
+            except ValueError:
+                continue
+            cmd = ("import numba; import multiprocessing;"
+                   "multiprocessing.set_start_method('{}');"
+                   "print(multiprocessing.get_context().get_start_method())")
+            cmdline = [sys.executable, "-c", cmd.format(meth)]
+            out, err = self.run_cmd(cmdline)
+            self.assertIn(meth, out)
```
stuartarchibald, line 915:
```suggestion
```

numba/tests/test_parallel_backend.py
```diff
@@ -897,6 +896,27 @@ def run_cmd(cmdline):
...
+            if self._DEBUG:
+                print("OUT:", out)
+                print("ERR:", err)
+
```
stuartarchibald, line 919:
```suggestion
            self.assertIn(meth, out)

```

numba/tests/test_parallel_backend.py
stuartarchibald: Good point, done in 
2f8c737

esc: Great!

numba/tests/test_parallel_backend.py
esc: Sounds good!

esc:
@stuartarchibald thanks for making those changes, I think this is ready to merge now.

PR #5732, opened 2020-05-20
CUDA Docs: document `forall` method of kernels

gmarkall:
I noticed there seems not to be any mention of `forall` in the docs - as this is intended to be a public API, this PR adds them to the documentation.


gmarkall:
@esc Thanks for taking a look at this! The docs for `forall` actually seemed to be a bit generally lacking so I've added documentation of the parameters as well as making the original content a bit more readable.

esc:
@gmarkall sweet!

Issue #5679, opened 2020-05-07
printf that can be called from Numba jit-decorated function


pearu:
## Feature request

Having the following implementation
```
from numba import extending, cgutils, types
@extending.intrinsic
def printf(typingctx, format_type, *args):
    """printf that can be called from Numba jit-decorated functions.
    """
    if isinstance(format_type, types.StringLiteral):
        sig = nb_types.void(format_type, types.BaseTuple.from_types(args))
        def codegen(context, builder, signature, args):
            cgutils.printf(builder, format_type.literal_value, *args[1:]) 
        return sig, codegen
```
one call `printf` from a jit-decorated function. For instance,
```
from numba import njit
@njit
def foo(a, b):
    c = 1.3
    printf("in foo, a=%p, b=%i, c=%e\n", a, b, c)
```
and then
```
>>> foo(1, 2)
in foo, a=0x1, b=2, c=1.300000e+00
```

I have found it useful for debugging. If there is interest, where such a codelet should land in numba?

stuartarchibald:
Thanks for the suggestion. I think this is a good thing to have, but should probably be `@overload`'d or `@register_jitable`'d and made to work "kind of the same" when JIT is disabled (i.e. might have to replace the format specifiers to make it work in CPython). As to location, how about `numba.misc.utils` or similar? There's `numba.misc.special` already but it's not like this is some special token, it's a useful piece of additional functionality.

QuantFreedom1022:
@pearu  ... does this still work ... i copied what you did and nothing was returned and it also seems that they moved cgutils

pearu:
@QuantFreedom1022 not likely, for instance, cgutils now lives under numba.core. However, it should be possible to fix the feature for newer numba versions.

nelson2005:
Did this work with string, like ```printf("%s %s", "hello", "world")```?  I ended up with a core dump when more than one string value is supplied.

quantfreedom:
@nelson2005 i am not sure ... i kind of gave up on numba because there are so many problems with debugging and printing and different things. Also i am kind of giving up on python because it is too slow so now i am switching my project over to c# because c# is just stupid fast. I think running c# in general is like running numba all the time. 

But anyway good luck on figuring it out.

PR #5734, opened 2020-05-20
Add printf and fflush

guilhermeleobas:
Following up on https://github.com/numba/numba/issues/5679. This PR implements `printf` and `fflush` for jitted functions.

To-Do
- [x] Fix `%p` specifier when JIT is disabled.
- [x] Fix char/string print when JIT is enabled.
- [x] Add tests
- [x] Add docs

Obs: `fflush` function type is compiler/machine dependent since it depends on the `FILE*` stuct. But since the implementation calls it as `fflush(NULL)`, one can bypass the LLVM typing system by just declaring it as a `int32 fflush(void*)`


guilhermeleobas:
Pull request is ready for a review. Not sure how to test it since I couldn't capture the output. 

ps: flake8 is failing in files not related to this PR.

pearu:
> Pull request is ready for a review. Not sure how to test it since I couldn't capture the output.

For capturing the output, try using the method from https://stackoverflow.com/questions/24277488/in-python-how-to-capture-the-stdout-from-a-c-shared-library-to-a-variable

guilhermeleobas:
Thanks for the help @pearu

guilhermeleobas:
The test that I wrote hangs on CI. So, I will just comment it for now.

docs/source/reference/utils.rst
```diff
@@ -3,6 +3,23 @@
 Utilities
 =========
 
+Printing
+========
+
+
+.. function:: numba.misc.utils.printf(format, ...)
+
+   Writes the C string pointed by `format` to the standart output (stdout).
```
luk-f-a, line 12: as a user, the reference to "C string" makes me wonder if I need to do anything special before being able to use this function. I would add two examples (re-using the test code) to show how to pass an integer and a string to `printf`. I would also explain the difference between using `print` and `printf`? does the first one use the python machinery, and the second not? Is `printf` faster, more robust? When should it be used instead of `print`?

numba/tests/test_utils.py
```diff
@@ -0,0 +1,87 @@
...
+# 	while True:
+# 		data = os.read(stdout_pipe[0], 1024)
+# 		if not data:
+# 			break
+# 		captured_stdout += data
+
+
+# def printf():
+# 	d = i = 392
+# 	utils.printf("d=%d, i=%i\n", d, i)
```
luk-f-a, line 28: there are no tests for %d and %c. is it because they are covered by %i and %s?

docs/source/reference/utils.rst
```diff
@@ -3,6 +3,23 @@
...
+   If `format` includes specifiers, the additional arguments following format
+   are inserted in the resulting string. The following formatters are supported:
+   `%d`, `%i`, `%u`, `%x`, `%X`, `%f`, `%F`, `%e`, `%E`, `%g`, `%G`, `%c`, and `%s`
```
luk-f-a, line 15: no reference to %o or %p even though there tests for them.

numba/tests/test_utils.py
```diff
@@ -0,0 +1,87 @@
...
+# 		os.dup2(stdout_save, stdout_fileno)
+# 		os.close(stdout_save)
+
+# 		expected = 'd=392, i=392\nu=7325\no=1142\nx=7fa\nX=7FA\n' + \
+# 				'f=392.651\nF=392.6512\ne=3.926400e+02\n' + \
+# 				'E=3.926400E+02\ng=392.65\nG=392.65\n' + \
+# 				'c=a\ns=sample\np=0xa\n'
+
+# 		output = captured_stdout.decode('utf-8')
+# 		self.assertEqual(expected, output)
```
luk-f-a, line 87: I think you need 
```
if __name__ == '__main__':
    unittest.main()
```

luk-f-a:
@guilhermeleobas I reviewed the documentation from a user point of view, and also looked at the tests. Unfortunately, I know next to nothing about `@intrinsic` so I cannot tell if your implementation is correct. sorry!

docs/source/reference/utils.rst
pearu: I can provide a few answers here:

> if I need to do anything special before being able to use this function

No

> I would also explain the difference between using print and printf? does the first one use the python machinery, and the second not?

I agree. Yes. The `printf` used here is a C implementation of https://en.wikipedia.org/wiki/Printf_format_string

> Is printf faster, more robust?

It is faster for sure. It is as robust as is C `printf`. Since it only reads memory, it should be hard to crash the program: when the program crashes on the `printf` call, it is usually the argument expressions that cause the crash, not `printf` function.

> When should it be used instead of print?

Never :), `printf` is not a replacement of `print`. `printf` is useful for debugging LLVM IR as it allows printing the actual memory values of low-level variables.

luk-f-a: thanks @pearu . @guilhermeleobas I think the above is worth writing the documentation.

numba/tests/test_utils.py
```diff
@@ -0,0 +1,87 @@
...
+# def drain_pipe():
+# 	global captured_stdout
+# 	while True:
+# 		data = os.read(stdout_pipe[0], 1024)
+# 		if not data:
+# 			break
+# 		captured_stdout += data
+
+
+# def printf():
```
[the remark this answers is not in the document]
guilhermeleobas: Test hang on CI. That's why I comment them.

numba/tests/test_utils.py
guilhermeleobas: There are tests for %d and %c. Former is in the text where you highlighted and %c is below.

docs/source/reference/utils.rst
guilhermeleobas: Fixed

numba/tests/test_utils.py
guilhermeleobas: Done

numba/tests/test_utils.py
luk-f-a: sorry, I didn't see it!

guilhermeleobas:
Thanks for the review @luk-f-a

numba/misc/utils.py
```diff
@@ -0,0 +1,65 @@
...
+            # Replace %c formatter by %s, since it doesn't work here
+            ft = format_type.literal_value.replace('%c', '%s')
+            cgutils.printf(builder, ft, value)
+        return sig, codegen
+
+
+@register_jitable
+def printf(format_type, *args):
+    if config.DISABLE_JIT:
+        if '%p' in format_type:
```
pearu, line 33: Could this exception be avoided by (untested code follows):
```
new_args = []
def process_format(m):
    orig = m.string[m.start():m.end()]
    if orig == '%p':
        new_args.append(id(args[len(new_args)]))
        return '%#x'
    else:
        new_args.append(args[len(new_args)])
        return orig
format_type = re.sub(r'%[diuxXfeEgGopcs]', process_format, format_type)
args = tuple(new_args)
```
?

numba/tests/test_utils.py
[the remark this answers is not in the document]
pearu: Could you enable these tests for the local run?
Is the hang platform-dependent?

numba/misc/utils.py
guilhermeleobas: Unfortunately no. Numba cannot compile this function.

guilhermeleobas:
@pearu 

> Could you enable these tests for the local run?
> Is the hang platform-dependent?

I think the test crashes only in specific machines on CI. See:
image

I will try to enable this test only for macOS and Windows.

numba/misc/utils.py
pearu: The suggestion should live under the `if config.DISABLE_JIT:..` block. Is this still a problem?

guilhermeleobas:
Not a single test executed. 
image

Every single test stuck on test discovery phrase. I think this is related to changing the file descriptor at the beginning of the test file. I will try to address this in a near future.

numba/misc/utils.py
guilhermeleobas: From what I understand, a `register_jitable` function needs to be compilable in nopython mode. Thus, code inside `if config.DISABLE_JIT` needs to be compilable, even though it will not be executed in nopython mode.

guilhermeleobas:
No, it never made to Numba main branch. It was hard to get the tests working on CI.

nelson2005:
That's a shame, it was a great addition to numba.

Thanks for the effort you put into it!

PR #5738, opened 2020-05-21
Unbreak master on buildfarm

sklam:
buildfarm is broken because of checks on llvmlite version is failing during loading of entrypoints. 

for example, buildfarm is creating llvmlite version:

```python
>>> import llvmlite
>>> llvmlite.__version__
'0.33.0dev0+17.gc4f5a08'
```

The version spec comparison thinks the above is greater than `0.33.0dev0` and fails.

To replicate the problem, get a llvmlite from buildfarm and just run 

```
$ python -m numba.runtests numba.tests.test_entrypoints.TestEntrypoints.test_entrypoint_tolerance
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,7 +12,7 @@ min_python_version = "3.6"
 min_numpy_build_version = "1.11"
 min_numpy_run_version = "1.15"
 min_llvmlite_version = "0.31.0.dev0"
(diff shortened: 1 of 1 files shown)


sklam:
smoking at `numba_smoketest_cpu_39`

PR #5751, opened 2020-05-26
[DISCUSSION] Allow CUDA array interface to be a callable

pwuertz:
Spin-off discussion from #5162, concerning the topic of allowing `__cuda_array_interface__` to be a callable (possible realization included in PR).

**Current state**
Arguably, `__cuda_array_interface__` is always going to be a function call for gathering the required description information from an underlying array implementation. With CAI being specified as a `dict` attribute this effectively requires CAI to be a python `property`, which is a quite problematic abstraction for its purpose.

**Issues with `property`**
Providing a device array descriptor is a potentially non-trivial operation. An implementation may choose to (or be required to) execute additional steps for creating a valid descriptor (e.g. internal syncs, initialization, queries, etc.).

The `property` mechanic is almost a guarantee for making more calls to the CAI function than intended or desired:
```python
# CAI code invoked multiple times by what appears to be a simple attribute lookup
other_shape = obj.__cuda_array_interface__["shape"]
other_dtype = obj.__cuda_array_interface__["dtype"]

# CAI code invoked implicitly by `hasattr` check
has_cai = hasattr(obj, "__cuda_array_interface__")

# CAI code invoked as a side effect if the object is being inspected
members = inspect.getmembers(obj)
```

**Proposed solution**
Modify CAI specification for `__cuda_array_interface__` to be a `dict` or a `Callable[[], dict]`. Consumers can easily be adapted to work with both types in a backward compatible manner. CAI implementations are free to switch to the callable interface as they see fit.

Bonus: I can finally finish my CAI implementation for OpenCV, which doesn't support wrapping member functions as python properties ^^

### DIFF
--- a/docs/source/cuda/cuda_array_interface.rst
+++ b/docs/source/cuda/cuda_array_interface.rst
@@ -20,8 +20,11 @@ Python Interface Specification
 
 .. note:: Experimental feature.  Specification may change.
 
-The ``__cuda_array_interface__`` attribute returns a dictionary (``dict``)
-that must contain the following entries:
+For providing a CUDA array interface, an object must contain either a
+``__cuda_array_interface__`` dictionary (``dict``) attribute or a method
+named ``__cuda_array_interface__`` that takes no arguments and returns
+a dictionary.
+The dictionary must contain the following entries:
 
 - **shape**: ``(integer, ...)``
 
--- a/numba/cuda/api.py
+++ b/numba/cuda/api.py
@@ -29,6 +29,9 @@ def from_cuda_array_interface(desc, owner=None):
     The *owner* is the owner of the underlying memory.
     The resulting DeviceNDArray will acquire a reference from it.
     """
+    # Argument may be a description or callable description provider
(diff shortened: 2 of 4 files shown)


leofang:
I would like to settle the discussion on handling streams/events/etc in #5162 before moving on to a concrete change. 

cc: @gmarkall (I'm sure you will see this shortly though 😇)

gmarkall:
Thanks for the PR, and for moving things forward!

I think it's good to have this discussion - I do also think @leofang's judgement in suggesting that we should settle the discussion on handling streams/events etc. before making any concrete changes is sound.

If I've read this correctly, the proposal here is a breaking change - a consumer that presently does `obj.__cuda_array_interface__['shape']` for example will start to fail to work with an `obj` that implements `__cuda_array_interface__` as a callable - even bumping the interface version doesn't necessarily help here, because you don't get to know the interface version until you've already called `obj.__cuda_array_interface__`. If this is the case, it seems like it will be some work to align all participating libraries on the change - is my estimation of the situation correct here?

PR #5757, opened 2020-05-27
Force LLVM to use ZMM registers for vectorization.

PokhodenkoSA:
This PR fixes tests for vectorization with LLVM 10. 
Tests search for using `zmm` registers in assembly code but LLVM 10 does not use 512-bit registers and prefer 256-bit registers by default (https://reviews.llvm.org/D67259). 
This PR uses `NUMBA_CPU_FEATURES` to disable preferring of 256-bit registers.

### DIFF
--- a/numba/tests/test_parfors.py
+++ b/numba/tests/test_parfors.py
@@ -2549,7 +2549,7 @@ class TestParforsVectorizer(TestPrangeBase):
         assertions = kwargs.pop('assertions', True)
 
         env_opts = {'NUMBA_CPU_NAME': cpu_name,
-                    'NUMBA_CPU_FEATURES': '',
+                    'NUMBA_CPU_FEATURES': '-prefer-256-bit',
                     }
 
         overrides = []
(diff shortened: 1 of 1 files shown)


stuartarchibald:
Thanks for the patch, I've queued it for review.

numba/tests/test_parfors.py
```diff
@@ -2549,7 +2549,7 @@ def get_gufunc_asm(self, func, schedule_type, *args, **kwargs):
         assertions = kwargs.pop('assertions', True)
 
         env_opts = {'NUMBA_CPU_NAME': cpu_name,
-                    'NUMBA_CPU_FEATURES': '',
+                    'NUMBA_CPU_FEATURES': '-prefer-256-bit',
```
stuartarchibald, line 2552: I think it might be a good idea to provide this in a similar fashion to the other options such that this function remains more general:
https://github.com/numba/numba/pull/5757/files#diff-44914098195178c859436d0f2377912bR2547-R2549

e.g.
```
cpu_features = kwargs.pop('cpu_features', '-prefer-256-bit')
```
and then
```
'NUMBA_CPU_FEATURES': cpu_features,
```
what do you think?

stuartarchibald: Also think it'd be worthwhile adding a 1 line comment about why this is needed and link to the LLVM merge request?

PR #5761, opened 2020-05-27
Add implmentation for int types for isnan and isinf for CUDA

### LINKED ISSUE #3789: isnan(int) isinf(int) should work
## Reporting a bug

`math.isnan(123)` and `math.isinf(123)` should not cause jit'ed function to raise TypingError.

niteya-shah:
fixes #3789 
I have currently added a small change that fixes the issue. However as mentioned in the issue, it may be possible to return False whenever insnan and isinf. I would like some guidance as to how to add that.

### DIFF
--- a/numba/cuda/cudaimpl.py
+++ b/numba/cuda/cudaimpl.py
@@ -468,6 +468,12 @@ def ptx_round(context, builder, sig, args):
     ])
 
 
+@lower(math.isinf, types.Integer)
+@lower(math.isnan, types.Integer)
+def math_isinf_isnan_int(context, builder, sig, args):
+    return lc.Constant.int(lc.Type.int(1), 0)
+
+
 def gen_deg_rad(const):
     def impl(context, builder, sig, args):
         argty, = sig.args
--- a/numba/cuda/libdevice.py
+++ b/numba/cuda/libdevice.py
@@ -105,14 +105,10 @@ binarys += [('__nv_hypot', '__nv_hypotf', math.hypot)]
 
 
 for name64, name32, key in booleans:
-    implf64 = bool_implement(name64, types.float64)
-    lower(key, types.float64)(implf64)
-    implf32 = bool_implement(name32, types.float32)
-    lower(key, types.float32)(implf32)
-    impli64 = bool_implement(name64, types.int64)
-    lower(key, types.int64)(impli64)
-    impli32 = bool_implement(name32, types.int32)
-    lower(key, types.int32)(impli32)
+    impl64 = bool_implement(name64, types.float64)
+    lower(key, types.float64)(impl64)
+    impl32 = bool_implement(name32, types.float32)
+    lower(key, types.float32)(impl32)
 
 
 for name64, name32, key in unarys:
--- a/numba/cuda/tests/cudapy/test_math.py
+++ b/numba/cuda/tests/cudapy/test_math.py
@@ -1,7 +1,7 @@
 import sys
 import numpy as np
 from numba.cuda.testing import unittest, CUDATestCase
-from numba import cuda, float32, float64, int32
+from numba import cuda, float32, float64, int32, int64
 import math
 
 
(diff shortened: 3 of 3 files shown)


esc:
@niteya-shah thanks for submitting this I have added it to the queue for review. 
Please be advised however that we are currently in a PR freeze as we are preparing to ship a release candidate (RC). We will begin reviewing PRs again once the RC is out and we thank you in advance for your patience! Also note that the CI errors you are seeing are probably not related to your changes but the result of the `flake8` package having become stricter.

gmarkall:
Thanks for the PR, and your interest in contributing to Numba! I'll provide some details of what needs doing below. I hope we can work through to completing this fix and getting it merged - do let me know if you have any more questions or need some more suggestions.

I'm not sure why there was a suggestion in the issue #3789 to use an implementation from libdevice, because `isinf` and `isnan` are `False` for all integers. The `LoweringError` that we presently get for these functions:

```
No definition for lowering <built-in function isnan>(int64,) -> bool
```

suggested that the typing is working OK, but there is no implementation to lower them from Numba IR to LLVM IR in the CUDA target. In the CUDA target, the lowering implementations are kept in https://github.com/numba/numba/blob/master/numba/cuda/cudaimpl.py

To add a lowering implementation for a function with argument types, you need to add a function decorated with `lower`:

```python
@lower(function, *argtypes)
```

where `function` is the Python function to lower - e.g. `math.isinf`. The `argtypes` are the types of the arguments as determined by Numba - from `numba.core.types`. Some documentation of the types is at: http://numba.pydata.org/numba-doc/latest/reference/types.html#basic-types.

So, the decoration for an implementation of `math.isinf` for integer arguments would look like:

```python
@lower(math.isinf, types.Integer)
```

There are specific integer types, like `types.int32`, etc., but using `types.Integer` catches them all.

Next, the definition of a lowering function accepts four arguments:

```python
def func(context, builder, sig, args):
```

The arguments passed to the lowering function are:

- `context`: The target context holds the implementations of functions and provides various utility functions for constructing correct LLVM IR values for the target (though the builder is used a lot for constructing IR).
- `builder`: This is a `Builder` object from llvmlite, used to construct most of the LLVM IR.
- `sig`: The signature for the function as typed by Numba.
- `args` a list of LLVM IR nodes previously constructed by lowering functions earlier in the IR. These make up the arguments to the function being lowered.

In the body of the lowering function we need to construct appropriate IR that implements the function. Since we only need to return false here regardless of the arguments, we can do this by constructing a 1-bit integer with a value of zero:

```python
    return lc.Constant.int(lc.Type.int(1), 0)
```

Putting this all together, the complete definition for `isinf` for integer types will be:

```python
@lower(math.isinf, types.Integer)
def math_isinf_int(context, builder, sig, args):
    return lc.Constant.int(lc.Type.int(1), 0)
```

After adding this, a similar lowering is needed for `math.isnan`. Note that you can chain `lower` decorators, to re-use the same implementation for multiple functions / types, e.g.:

```python
@lower(math.isinf, types.Integer)
@lower(math.isnan, types.Integer)
def f(...):
    ...
```

Once the lowerings for these two functions have passed, this snippet should execute successfully:

```python
from numba import cuda
import math

import numpy as np


@cuda.jit
def foo(r, x, y):
    r[0] = math.isnan(x)
    r[1] = math.isinf(x)


r = np.ones(2, dtype=np.int32)

foo[1, 1](r, 1, 1)

assert r[0] == 0
assert r[1] == 0
```

Something along these lines should be converted into a test case, that should be added in https://github.com/numba/numba/blob/master/numba/cuda/tests/cudapy/test_math.py.

Once you have added the lowering and tests, I think this would then be ready for review.

niteya-shah:
Thank you so much @gmarkall for the very detailed explanation. I have made the necessary changes.

leofang:
@gmarkall It is super detailed and useful! Have you considered compiling a list of links to your such replies? 🙂

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_60`.

gmarkall:
@esc Could this have a run on the buildfarm please?

esc:
`numba_smoketest_cuda_89`

esc:
BF passed.

PR #5764, opened 2020-05-28
fix flake8 errors

esc:
As of flake8 version 3.8.2 (mccabe: 0.6.1, pycodestyle: 2.6.0, pyflakes:
2.2.0) -- some additional errors are being reported. This PR fixes those
errors.

### DIFF
--- a/numba/core/cgutils.py
+++ b/numba/core/cgutils.py
@@ -678,7 +678,7 @@ def do_boundscheck(context, builder, ind, dimlen, axis=None):
                        "for axis {} with size %d\n".format(axis), ind, dimlen)
             else:
                 printf(builder, "debug: IndexError: index %d is out of bounds "
-                       "for axis %d with size %d\n".format(axis), ind, axis,
+                       "for axis %d with size %d\n", ind, axis,
                        dimlen)
         else:
             printf(builder,
--- a/numba/core/types/function_type.py
+++ b/numba/core/types/function_type.py
@@ -52,7 +52,7 @@ class FunctionType(Type):
             # first-class function may not use the same argument names
             # that the caller assumes. [numba/issues/5540].
             raise errors.UnsupportedError(
-                f'first-class function call cannot use keyword arguments')
+                'first-class function call cannot use keyword arguments')
 
         if len(args) != self.nargs:
             raise ValueError(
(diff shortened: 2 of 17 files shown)


numba/core/cgutils.py
```diff
@@ -678,7 +678,7 @@ def _dbg():
                        "for axis {} with size %d\n".format(axis), ind, dimlen)
             else:
                 printf(builder, "debug: IndexError: index %d is out of bounds "
-                       "for axis %d with size %d\n".format(axis), ind, axis,
+                       "for axis %d with size %d\n", ind, axis,
```
esc, line 681: Yeah, this one I am not so sure about...

stuartarchibald: Think it's a copy paste bug. There's nothing to format and the rest of the args are for the `printf`. It's like doing `"hello".format(123)`.

esc: Yeah, I think you are correct.

numba/core/types/function_type.py
```diff
@@ -52,7 +52,7 @@ def get_call_type(self, context, args, kws):
             # first-class function may not use the same argument names
             # that the caller assumes. [numba/issues/5540].
             raise errors.UnsupportedError(
-                f'first-class function call cannot use keyword arguments')
+                'first-class function call cannot use keyword arguments')
```
stuartarchibald, line 55: Is flake8 complaining about a `f` string with no format expression?

esc: indeed

esc:
We could probably have ignored more errors, but since there weren't too many new ones it seemed reasonable to fix. IIRC we still ignore quite a large part of the codebase during flake8 runs:

https://github.com/numba/numba/blob/master/.flake8

stuartarchibald:
> We could probably have ignored more errors, but since there weren't too many new ones it seemed reasonable to fix. IIRC we still ignore quite a large part of the codebase during flake8 runs:
> 
> https://github.com/numba/numba/blob/master/.flake8

Yes, it's slowly improving. Numba's code base is too large to just fix manually in a single sweep (like I did for `llvmlite`, that was a couple of hours) so we'd either have to use tools, which means we'd need to agree with said tool, or do it incrementally. I'm ok with incremental for now, weird formatting is often useful in the code base because of the weird things often required, slow adjustment permits judgement on such things.

Issue #5755, opened 2020-05-27
@vectorize-decorated functions are not found by sphinx


HPLegion:
- [x] I am using the latest released version of Numba (most recent is visible in
 the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [ ] I have included below a minimal working reproducer (if you are unsure how
 to write one see http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports).

I use `sphinx` with the `apidoc` extension to have a reference documentation generated automatically.
For functions decorated with `njit` this works without any complications, for `vectorize` however, the function disappears from the docs, indicating that apidoc either cannot find it or decides not to include it. (Sorry for the lack of better info).

Since this is sphinx related I am not sure how to include a minimal example, but if it would help and I find some time I can try to set up a mini-repository.

esc:
@HPLegion thanks for submitting this. I'm afraid, w/o a reproducer this will be difficult to triage. Perhaps you could make a patch against the Numba source tree (which also uses sphinx for documentation) to expose the issue? Or if you have this in an open-source project, perhaps link to it? Perhaps you could try messing with different versions of sphinx/apidoc and see if the issue vanishes? Sorry I don't have any better ideas here.

HPLegion:
I have gotten around to construct a small example #5765. In that process I also realised that the problem does not lie with sphinx `apidoc` (which just creates the skeleton) but with `autodoc`, which generates the actual reference documentation.

HPLegion:
Okay, after stepping through sphinx I think I narrowed it down to the following:
`njit` functions have the `__wrapped__` attribute which I suppose contains the original python function. Sphinx can resolve this.
This field does not exist for `vectorize`, hence autodoc thinks that it cannot document the function and skips it (it is not rejected based on any filters as far as I can tell, sphinx just autodoc just gives up)

HPLegion:
Some more digging has revealed that `njit` functions get documented by autodoc because 
- they are a `methoddescriptor` 
- & have the python function in `__wrapped__` 
- & have the correct `__module__` string.

```python
>>> from numba import njit
>>> import inspect
>>> @njit
... def f(x):
...     return x
... 
>>> f(1)
1
>>> inspect.isfunction(f)
False
>>> inspect.ismethoddescriptor(f)
True
>>> inspect.isroutine(f)
True
>>> hasattr(f, "__wrapped__")
True
>>> f.__module__
'__main__'
>>> f.__qualname__
[...]

PR #5765, opened 2020-05-28
[DON'T MERGE] MWE for sphinx not finding vectorized funcs

HPLegion:
This is a MWE for #5755 

Folder SPHINXBUG contains a tiny demo package with 3 functions and a small sphinx config

```python
import numba

def f_python(x):
    """
    I am a python function
    """
    return x

@numba.njit()
def f_njit(x):
    """
    I am an njit function
    """
    return x

@numba.vectorize([numba.float64(numba.float64)], nopython=True)
def f_vectorize(x):
    """
    I am a vectorize function
    """
    return x
```

When building the docs using _sphinx_ with the _autodoc_ extension, the `@vectorize` function is not added to the documentation.

Easiest way to build docs is to run `make text` in the docs directory
This creates the following output file
```
Welcome to mockpackage's documentation!
***************************************


Module contents
===============

mockpackage.f_njit(x)

   I am an njit function

mockpackage.f_python(x)

   I am a python function
```


HPLegion:
Also, please let me know if there is a better way to "design" a pull request of this kind for future purposes ;-)

HPLegion:
One more observation: By explicitly adding `.. autofunction:: f_vectorize` to the docs, it does show up in the final product, but with an altered signature.
```
[...]
mockpackage.f_vectorize(*args, **kwargs)

   I am a vectorize function
```

PR #5768, opened 2020-05-28
Interval example: fix import

EQt:
`cgutils` must be imported from `numba.core`

### DIFF
--- a/docs/source/extending/interval-example.rst
+++ b/docs/source/extending/interval-example.rst
@@ -191,7 +191,7 @@ Implementing the constructor
 Now we want to implement the two-argument ``Interval`` constructor::
 
    from numba.extending import lower_builtin
-   from numba import cgutils
+   from numba.core import cgutils
 
    @lower_builtin(Interval, types.Float, types.Float)
    def impl_interval(context, builder, sig, args):


esc:
@EQt thanks for catching this one! It looks good to me so I am marking it as ready to merge.

esc:
For future reference: if you wanted to take this further, we recently established a way to include example snippets in such a way that they are tested.

Here is an example from:

https://github.com/numba/numba/blob/master/docs/source/reference/pysupported.rst

It goes like:

```
Here's an example using ``List()`` to create ``numba.typed.List`` inside a
jit-compiled function and letting the compiler infer the item type:

.. literalinclude:: ../../../numba/tests/doc_examples/test_typed_list_usage.py
   :language: python
   :caption: from ``ex_inferred_list_jit`` of ``numba/tests/doc_examples/test_typed_list_usage.py``
   :start-after: magictoken.ex_inferred_list_jit.begin
   :end-before: magictoken.ex_inferred_list_jit.end
   :dedent: 12
   :linenos:
```

Basically the snippet is in `numba/tests/doc-examples` so it will be automatically be picked up during testing. We then use a combination of `literalinclude` to select the file and `magictoken` (basically a marker in a python comment)  to cut out the correct part of the file to include in the docs.

PR #5773, opened 2020-05-29
fix exception during locale detection

esc:
Under certain circumstances for example if `$LANG` is unset in the
environment, `getdefaultlocale` can return the tuple `(None, None)`.
This in turn would cause an `Exception` to be raised by `join` at least
in Python 3.6, as it expects the iterable to contain `str`s. The result
was the following error message:

```
Error (locale): sequence item 0: expected str instance, NoneType found
```

Unfortunately the tests for the system info detection would then crash
making the test-suite fail. To mitigate this, we now cast everything
returned by `getdefaultlocale` to `str` to make sure that the call to
`join` will succeed. The output will then be `None.None` which is fine
as it indicates that no default locale was detected.

### DIFF
--- a/numba/misc/numba_sysinfo.py
+++ b/numba/misc/numba_sysinfo.py
@@ -306,7 +306,10 @@ def get_sysinfo():
     # Python locale
     # On MacOSX, getdefaultlocale can raise. Check again if Py > 3.7.5
     try:
-        sys_info[_python_locale] = '.'.join(locale.getdefaultlocale())
+        # If $LANG is unset, getdefaultlocale() can return (None, None), make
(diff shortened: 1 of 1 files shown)


esc:
This is a fix for changes introduced in #5591. Additional bugs were detected by the internal build farm and are fixed by this.

esc:
Internal build-farm ID: `numba_175`

stuartarchibald:
Pulled this into https://github.com/numba/numba/pull/5791 too.

stuartarchibald:
Closing this as it got merged as part of #5791. Thanks for fixing @esc.

esc:
🍰

PR #5779, opened 2020-05-30
Return zero array from np.dot when the arguments are empty.

### LINKED ISSUE #5539: Invalid np.dot() output for vector/matrix multiply of empty arrays
## Reporting a bug

Looks like `np.dot` implementation doesn't handle empty input properly:

```python
In [161]: X = np.array([]).reshape(0, 2)

In [162]: Y = np.array([])

In [163]: numba.njit(lambda X, Y: np.dot(Y, X))(X, Y)
Out[163]: array([-1.36311572e+57, -1.36311572e+57])

In [164]: np.dot(Y, X)
Out[164]: array([0., 0.])
```

AndrewEckart:
When np.dot(a, b) is called with empty argments, the result can be non-empty if the zero dimension is eliminated (e.g. a is m x 0 and b is 0 x n). In these cases, we should return an array of zeros. Previously, the result was initialized with np.empty, so the it contained garbage values.

Fixes #5539

### DIFF
--- a/numba/np/linalg.py
+++ b/numba/np/linalg.py
@@ -392,8 +392,6 @@ def call_xxgemv(context, builder, do_trans,
                             builder.bitcast(v_data, ll_void_p),
                             builder.bitcast(beta, ll_void_p),
                             builder.bitcast(out_data, ll_void_p)))
-    # if lda == 0:
-    #     cgutils.memset(builder, res.data, builder.mul(res.itemsize, res.nitems), 0)
     check_blas_return(context, builder, res)
 
 
@@ -444,8 +442,6 @@ def call_xxgemm(context, builder,
                             builder.bitcast(alpha, ll_void_p), data_a, lda,
                             data_b, ldb, builder.bitcast(beta, ll_void_p),
                             data_c, ldc))
-    # if k == 0:
-    #     cgutils.memset(builder, res.data, builder.mul(res.itemsize, res.nitems), 0)
     check_blas_return(context, builder, res)
 
 
@@ -685,7 +681,7 @@ def dot_3_mm(context, builder, sig, args):
             cgutils.memset(builder, out.data,
                            builder.mul(out.itemsize, out.nitems), 0)
         with nonempty:
-            # Check whether any of the operands is really a 1-d vector represented
+            # Check if any of the operands is really a 1-d vector represented
             # as a (1, k) or (k, 1) 2-d array.  In those cases, it is pessimal
             # to call the generic matrix * matrix product BLAS function.
             one = ir.Constant(intp_t, 1)
--- a/numba/np/linalg.py
+++ b/numba/np/linalg.py
@@ -599,6 +599,7 @@ def dot_3_vm(context, builder, sig, args):
         mty = yty
         m_shapes = y_shapes
         v_shape = x_shapes[0]
+        lda = m_shapes[1]
         do_trans = yty.layout == 'F'
         m_data, v_data = y.data, x.data
         check_args = dot_3_vm_check_args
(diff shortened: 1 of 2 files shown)


stuartarchibald:
Thanks for the patch. The fix seems good. Please could you add/update appropriate unit tests in here: https://github.com/numba/numba/blob/master/numba/tests/test_linalg.py#L30 to ensure that the change is correct. Many thanks.

AndrewEckart:
Seems my fix is working for `dot_2`, but `dot_3` also needs to be fixed, and I'm not quite familiar enough with the lowered API to get it working. I was trying to do something along the lines of the `numpy_zeros_nd` implementation with `_zero_fill_array`.

AndrewEckart:
Okay, I got it working for both `dot_2` and `dot_3`. I'm very much a beginner with the IR Builder, so please let me know if there's a cleaner way to write things.

numba/np/linalg.py
```diff
@@ -611,8 +618,15 @@ def dot_3_vm(context, builder, sig, args):
     for val in m_shapes:
         check_c_int(context, builder, val)
 
-    call_xxgemv(context, builder, do_trans, mty, m_shapes, m_data,
-                v_data, out.data)
+    zero = ir.Constant(intp_t, 0)
```
stuartarchibald, line 621: There's a preference in the code base for using the `context` where possible, e.g.
`zero = context.get_constant(types.intp, 0)`
(same applies throughout).

numba/tests/test_linalg.py
```diff
@@ -169,16 +168,21 @@ def samples(m, n):
         cfunc2 = jit(nopython=True)(pyfunc2)
         if pyfunc3 is not None:
             cfunc3 = jit(nopython=True)(pyfunc3)
-        for a, b in samples(m, n):
-            self.check_func(pyfunc2, cfunc2, (a, b))
-            self.check_func(pyfunc2, cfunc2, (b, a.T))
-        if pyfunc3 is not None:
+
+        for m, n in [(2, 3),
+                     (3, 0)
```
stuartarchibald, line 173: What about e.g. `(0, 3)` ?

numba/tests/test_linalg.py
```diff
@@ -229,10 +233,12 @@ def samples(m, n, k):
 
         # Test generic matrix * matrix as well as "degenerate" cases
         # where one of the outer dimensions is 1 (i.e. really represents
-        # a vector, which may select a different implementation)
+        # a vector, which may select a different implementation),
+        # or both matrices are empty but their product is not.
         for m, n, k in [(2, 3, 4),  # Generic matrix * matrix
                         (1, 3, 4),  # 2d vector * matrix
                         (1, 1, 4),  # 2d vector * 2d vector
+                        (3, 2, 0)   # Empty matrix * empty matrix
```
stuartarchibald, line 241: What about e.g. `(0, 0, 3)` ?

AndrewEckart:
> With the new use cases, I think they will probably need explicitly catching as per the current "inner dimension is zero" cases, at present, these new cases are manifesting as an error in the BLAS call, `XERBLA` is complaining!

Updated to deal with cases where one or more arguments is empty because an outer dimension is zero and added test cases to cover each case.

It's a little weird because in these cases, the output is empty, but I'm still calling `cgutils.memset` with `size=0` (which does nothing). I could add more conditional logic to avoid that if it would be preferable.

stuartarchibald:
> > With the new use cases, I think they will probably need explicitly catching as per the current "inner dimension is zero" cases, at present, these new cases are manifesting as an error in the BLAS call, `XERBLA` is complaining!
> 
> Updated to deal with cases where one or more arguments is empty because an outer dimension is zero and added test cases to cover each case.
> 
> It's a little weird because in these cases, the output is empty, but I'm still calling `cgutils.memset` with `size=0` (which does nothing). I could add more conditional logic to avoid that if it would be preferable.

Thanks for patching that too. Until demonstrated as a bottleneck, I think just leaving a `memset` of zero bytes is fine, it's a function call vs branch, and then obviously complexity that goes into handling it.

PR #5781, opened 2020-05-30
Moving record array examples to a test module

luk-f-a:
Fixes both record array examples in `docs/source/reference/numpysupported.rst`, and moves the content to a test so it gets executed.

cc @stuartarchibald

### DIFF
--- a/docs/source/reference/numpysupported.rst
+++ b/docs/source/reference/numpysupported.rst
@@ -99,40 +99,22 @@ Structured scalars support attribute getting and setting, as well as
 member lookup using constant strings. Strings stored in a local or global tuple
 are considered constant strings and can be used for member lookup.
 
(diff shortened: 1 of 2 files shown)


stuartarchibald:
Thanks for the patch. Is there an issue associated with this or was this straight from gitter chat?

luk-f-a:
> Thanks for the patch. Is there an issue associated with this or was this straight from gitter chat?

no issue that I know of, I picked it up from the gitter discussion

stuartarchibald:
> > Thanks for the patch. Is there an issue associated with this or was this straight from gitter chat?
> 
> no issue that I know of, I picked it up from the gitter discussion

Thanks for confirming, just wanted to make sure a corresponding issue was closed if there was one!

PR #5783, opened 2020-05-30
Make np.divide and np.remainder code more similar

eric-wieser:
Split out of #5619. Some simple cleanup to make adding `divrem` easier.

### DIFF
--- a/numba/np/npyfuncs.py
+++ b/numba/np/npyfuncs.py
@@ -122,13 +122,14 @@ def _dispatch_func_by_name_type(context, builder, sig, args, table, user_name):
 
 def np_int_sdiv_impl(context, builder, sig, args):
     # based on the actual code in NumPy loops.c.src for signed integer types
-    num, den = args
-    lltype = num.type
-    assert all(i.type==lltype for i in args), "must have homogeneous types"
+    _check_arity_and_homogeneity(sig, args, 2)
 
-    ZERO = lc.Constant.int(lltype, 0)
-    MINUS_ONE = lc.Constant.int(lltype, -1)
-    MIN_INT = lc.Constant.int(lltype, 1 << (den.type.width-1))
+    num, den = args
+    ty = sig.args[0]  # any arg type will do, homogeneous
+
+    ZERO = context.get_constant(ty, 0)
+    MINUS_ONE = context.get_constant(ty, -1)
+    MIN_INT = context.get_constant(ty, 1 << (den.type.width-1))
     den_is_zero = builder.icmp(lc.ICMP_EQ, ZERO, den)
     den_is_minus_one = builder.icmp(lc.ICMP_EQ, MINUS_ONE, den)
     num_is_min_int = builder.icmp(lc.ICMP_EQ, MIN_INT, num)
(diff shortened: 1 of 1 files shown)


stuartarchibald:
Thanks for the patch. I've queued it for review.

numba/np/npyfuncs.py
```diff
@@ -161,10 +162,9 @@ def np_int_srem_impl(context, builder, sig, args):
     _check_arity_and_homogeneity(sig, args, 2)
 
     num, den = args
-    ty = sig.args[0] # any arg type will do, homogeneous
-    lty = num.type
+    lltype = num.type
 
-    ZERO = context.get_constant(ty, 0)
+    ZERO = lc.Constant.int(lltype, 0)
```
stuartarchibald, line 167: There's a general preference in Numba to use the `context` opposed to `llvmpy`.

eric-wieser: So you'd want me to make the change in the reverse direction?

eric-wieser: (by changeing line 194 below instead?)

eric-wieser: Also, does numba provide any helper to convert between the numba and llvm types?

stuartarchibald: Yes please, the reverse direction change is appropriate. The context has loads of helper functions on it, to find the llvm equivalent of a numba type use `context.get_value_type`. e.g. some bits and pieces from breaking in the `ii->i` loop which hits `np_int_srem_impl`:
```
(Pdb) sig.args
(int32, int32)
(Pdb) argty = context.get_value_type(sig.args[0])
(Pdb) p argty
<<class 'llvmlite.ir.types.IntType'> i32>
(Pdb) ZERO = context.get_constant_generic(builder, sig.args[0], 0)
(Pdb) p ZERO
<ir.Constant type='i32' value=0>
```

stuartarchibald: hmmm, this might not work here, just need to test something.

stuartarchibald: checked, it's ok.

eric-wieser: To be clear, I'm asking if the type can be obtained in the other direction

stuartarchibald: Don't think so, the map would be one llvm type -> many Numba types?
```
(Pdb) context.get_value_type(types.int8)
<<class 'llvmlite.ir.types.IntType'> i8>
(Pdb) context.get_value_type(types.uint8)
<<class 'llvmlite.ir.types.IntType'> i8>
(Pdb) context.get_value_type(types.char)
<<class 'llvmlite.ir.types.IntType'> i8>
```

eric-wieser: Good point. Sounds like I need to modify #5619 to construct the signature too then.

eric-wieser: And by that, I'm referring to these lines: https://github.com/numba/numba/pull/5619/files#diff-6fb12019cea9375115bc5f3b7a71fe3dR277

PR #5785, opened 2020-05-31
[WIP] Implement np.tile

### LINKED ISSUE #2327: Please add np.tile implementation in nopython mode
Right now this is crucial, How can I circumvent the fact that Numba does not support it?

### LINKED ISSUE #3456: Support `numpy.tile`
## Feature request

Add support for `numpy.tile`

AndrewEckart:
Add basic support for np.tile, limited to when the first input (the tiles) have at most 1 dimension. Includes documentation and tests.

I'd like to support higher-dimensional inputs, but I spent a while trying to figure out how to dynamically allocate a new tuple at runtime for the output shape with little success (the number of dimensions needs to be equal to max(len(reps), a.ndim). It seems like the only option is `context.make_tuple` using the `lower_builtin` API, but that didn't seem like a great idea. A possible workaround is using a list/array for the shape, but unfortunately that's not yet supported by `np.reshape` or `array.reshape`.

Fixes #2327
Fixes #3456

### DIFF
--- a/docs/source/reference/numpysupported.rst
+++ b/docs/source/reference/numpysupported.rst
@@ -428,6 +428,7 @@ The following top-level functions are supported:
 * :func:`numpy.sort` (no optional arguments)
 * :func:`numpy.stack`
 * :func:`numpy.take` (only the 2 first arguments)
+* :func:`numpy.tile` (first argument must by a 1-D array or scalar)
 * :func:`numpy.transpose`
 * :func:`numpy.trapz` (only the 3 first arguments)
 * :func:`numpy.tri` (only the 3 first arguments; third argument ``k`` must be an integer)
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -6,7 +6,7 @@ Implementation of math operations on Array objects.
 import math
 from collections import namedtuple
 from enum import IntEnum
-from functools import partial
+from functools import partial, reduce
 
 import numpy as np
 
@@ -1568,7 +1568,7 @@ def np_triu_indices_from(arr, k=0):
     return np_triu_indices_from_impl
 
 
-def _prepare_array(arr):
+def _prepare_array(arr) -> np.ndarray:
     pass
 
 
@@ -1808,6 +1808,52 @@ def np_roll(a, shift):
         return lambda a, shift: np.asarray(a)
     else:
         return np_roll_impl
+
+
+@overload(np.tile)
+def np_tile(a, reps):
+    def np_tile_integer_impl(a, reps):
+        if reps < 0:
+            raise ValueError("tile(): second argument must be non-negative")
+        arr = np.asarray(a)
+        sz = arr.size
+        out = np.empty(reps * sz, dtype=arr.dtype)
+        for i in range(reps):
+            start = i * sz
+            end = start + sz
+            out[start:end] = arr.flatten()
+        return out
+
+    def np_tile_impl(a, reps):
+        for r in reps:
+            if r < 0:
+                raise ValueError("tile(): second argument must be "
+                                 "non-negative")
+
+        arr = np.asarray(a)
+        sz = arr.size
+        # assumes arr is 1-dimensional
+        out_shape = reps[:-1] + (reps[-1] * sz,)
+        num_copies = reduce(lambda x, y: x * y, reps, 1)
+        out = np.empty(shape=num_copies * sz, dtype=arr.dtype)
+        for i in range(num_copies):
+            start = i * sz
+            end = start + sz
+            out[start:end] = arr.flatten()
+        return np.reshape(out, out_shape)
+
+    # Type checks
+    if not type_can_asarray(a):
+        raise TypingError("First argument must be array-like")
+    if isinstance(reps, types.Integer):
+        return np_tile_integer_impl
+    elif isinstance(reps, (types.Tuple, types.Array, types.Sequence)):
+        if isinstance(reps, types.Array) and reps.ndim != 1:
+            raise TypingError("Second argument must be 1-dimensional")
+        if isinstance(reps.dtype, types.Integer):
+            return np_tile_impl
+    raise TypingError("Second argument must be an integer "
+                      "or an array/sequence of integers")
 
 
 #----------------------------------------------------------------------------
--- a/numba/tests/test_np_functions.py
+++ b/numba/tests/test_np_functions.py
@@ -4,6 +4,7 @@ import itertools
 import math
 import platform
 from functools import partial
+from itertools import product
 
 import numpy as np
 
@@ -311,6 +312,10 @@ def flip_lr(a):
 
 def flip_ud(a):
     return np.flipud(a)
+
+
+def tile(a, reps):
+    return np.tile(a, reps)
 
 
 class TestNPFunctions(MemoryLeakMixin, TestCase):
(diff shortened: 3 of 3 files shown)


stuartarchibald:
Thanks for the patch. There's a few tricks to do tuple generation. Do you want to perhaps get this restricted version in first and then work on expanding it?

AndrewEckart:
@stuartarchibald Sure, if you think that would be best. What are the tricks?

docs/source/reference/numpysupported.rst
```diff
@@ -428,6 +428,7 @@ The following top-level functions are supported:
 * :func:`numpy.sort` (no optional arguments)
 * :func:`numpy.stack`
 * :func:`numpy.take` (only the 2 first arguments)
+* :func:`numpy.tile` (first argument must by a 1-D array or scalar)
```
mishana, line 431: by->be

numba/np/arraymath.py
```diff
@@ -1810,6 +1810,52 @@ def np_roll_impl(a, shift):
         return np_roll_impl
 
 
+@overload(np.tile)
+def np_tile(a, reps):
+    def np_tile_integer_impl(a, reps):
+        if reps < 0:
+            raise ValueError("tile(): second argument must be non-negative")
```
mishana, line 1817: "second argument" -> reps

numba/np/arraymath.py
```diff
@@ -1810,6 +1810,52 @@ def np_roll_impl(a, shift):
...
+        arr = np.asarray(a)
+        sz = arr.size
+        out = np.empty(reps * sz, dtype=arr.dtype)
+        for i in range(reps):
+            start = i * sz
+            end = start + sz
+            out[start:end] = arr.flatten()
```
mishana, line 1824: I think it might be better to call `.flatten()` only once before the loop.

numba/np/arraymath.py
```diff
@@ -1810,6 +1810,52 @@ def np_roll_impl(a, shift):
...
+        arr = np.asarray(a)
+        sz = arr.size
+        # assumes arr is 1-dimensional
+        out_shape = reps[:-1] + (reps[-1] * sz,)
+        num_copies = reduce(lambda x, y: x * y, reps, 1)
+        out = np.empty(shape=num_copies * sz, dtype=arr.dtype)
+        for i in range(num_copies):
+            start = i * sz
+            end = start + sz
+            out[start:end] = arr.flatten()
```
mishana, line 1842: Again, better call `flatten()` once before the loop. Also, if you assume `arr` is 1D then why the use of `flatten()` at all?

AndrewEckart: Good point. I was using `flatten()` with an eye towards supporting arrays of arbitrary dimension, which is held up because I haven't been able to figure out how to dynamically allocate a tuple of length `max(len(reps), arr.ndim)` for `out_shape`.

numba/np/arraymath.py
AndrewEckart: Looks like in vanilla NumPy, `tile` uses `repeat` under the hood, so passing a negative in `reps` results in `ValueError: negative dimensions are not allowed`. Should we match that?

mishana: Yes, I think so

mishana:
Btw, shouldn't this be implemented in `numba.np.arrayobj` instead? (`np.repeat` is implemented there)

AndrewEckart:
> Btw, shouldn't this be implemented in `numba.np.arrayobj` instead? (`np.repeat` is implemented there)

Yeah, probably. I think my rationale was that `repeat` is available as a method on `ndarray` and `arraymath` had this "Building matrices" section with functions that also weren't really computational. But I think `tile` is more similar to `hstack`, `vstack`, etc. as well as `repeat` so you're right that it probably fits better in `arrayobj`.

numba/np/arrayobj.py
```diff
@@ -1993,6 +1993,53 @@ def array_repeat_impl(a, repeats):
...
+        arr = np.asarray(a)
+        sz = arr.size
+        # assumes arr is 1-dimensional
+        out_shape = reps[:-1] + (reps[-1] * sz,)
+        num_copies = reduce(lambda x, y: x * y, reps, 1)
+        out = np.empty(shape=num_copies * sz, dtype=arr.dtype)
+        for i in range(num_copies):
+            start = i * sz
+            end = start + sz
+            out[start:end] = arr.flatten()
```
mishana, line 2026: call `arr.flatten()` once before the loop

AndrewEckart: Oops, meant to change both implementations but only changed one.

numba/tests/test_np_functions.py
```diff
@@ -3621,6 +3626,71 @@ def test_cross2d_exceptions(self):
...
+        cfunc = njit(pyfunc)
+        for a, reps in product(tiles(), shapes()):
+            print(a, reps)
+            expected = pyfunc(a, reps)
+            got = cfunc(a, reps)
+            self.assertPreciseEqual(expected, got)
+
+    def test_tile_exceptions(self):
+        pyfunc = tile
+        cfunc = njit(pyfunc)
```
mishana, line 3655: Is the naming `cfunc` for jitted functions in the tests a convention throughout Numba? Seems a little misleading in my opinion.

numba/tests/test_np_functions.py
```diff
@@ -3621,6 +3626,71 @@ def test_cross2d_exceptions(self):
...
+            yield 10
+            yield 0
+            yield (6,)
+            yield (3, 3)
+            yield (2, 1, 4)
+
+        pyfunc = tile
+        cfunc = njit(pyfunc)
+        for a, reps in product(tiles(), shapes()):
+            print(a, reps)
```
mishana, line 3648: I don't see the need for a print inside of a unit test

numba/tests/test_np_functions.py
mishana: Ok I see it is a convention. I guess it stands for "compiled" func or something. It is ok then.

jdtuck:
what is the status on merging this in, this feature would be great

ngodber:
I'd also like to join the chorus of requestors. Tile is such a useful utility function it can entail quite a bit of refactoring to replace it!

gmarkall:
After merging master into this branch there is a small conflict which is trivial to resole - @AndrewEckart would you like to merge master into your branch and I can add this to the review queue?

numba/np/arraymath.py
```diff
@@ -1712,7 +1712,7 @@ def np_triu_indices_from_impl(arr, k=0):
     return np_triu_indices_from_impl
 
 
-def _prepare_array(arr):
+def _prepare_array(arr) -> np.ndarray:
```
AndrewEckart, line 1715: Not sure how this type annotation got in here - my guess is that it was removed some time in the last several months after I originally branched, and then git put it back when I merged master into the feature branch.

gmarkall:
Thanks for the update! I've added this to the review queue, and put it in the 0.54 milestone to make sure it doesn't get passed over or omitted from consideration for the next release without proper consideration.

ngodber:
Excited to use this in anger! One less impediment to decorating a function and being off to the races. Thanks a lot guys, very much appreciated!

raghavduddala:
It would be great to have this functionality. Is this functionality going to be merged sometime soon? Seems like this PR was added for a review for 0.54 release. Is anyone still working on this or this PR stale?

AlessandroFazio:
Hello, it would be useful to have a built-in implementation of np.tile. Hope it will be merge in the future.

moqi112358:
Is anyone still working on this or this PR?

esc:
> Is anyone still working on this or this PR?

probably not, it's been stale for over a year now.

ngodber:
I was involved when this first 'came back alive'. I didn't catch that it was never infact merged. Thats a bummer. @gmarkall  would it possible to get this reviewed again and pushed to be added to the next release? Do we know why it dropped off the radar?

PR #5789, opened 2020-06-01
[WIP] support for np.putmask

arthurfeeney:
This is a small PR that adds support for np.putmask. I believe it reproduces numpy's behavior. [WIP] is mostly for Travis CI and a question since this is my first PR: I put it in ``numba/np/arraymath.py`` with the element-wise operations (It felt similar to np.where). Is there a better spot for it? Thanks

### DIFF
--- a/numba/tests/test_np_functions.py
+++ b/numba/tests/test_np_functions.py
@@ -45,6 +45,10 @@ def append(arr, values, axis):
     return np.append(arr, values, axis=axis)
 
 
+def putmask(a, mask, values):
+    return np.putmask(a, mask, values)
+
+
 def count_nonzero(arr, axis):
     return np.count_nonzero(arr, axis=axis)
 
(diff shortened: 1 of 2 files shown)


stuartarchibald:
Thanks for the patch. Suggested location seems good. I've queued it for review.

FirefoxMetzger:
@stuartarchibald This is related to my request for `np.putmask` in https://github.com/numba/numba/issues/7374#issuecomment-91362858. Since it is already in-flight; would it help to move things forward if I leave a review?

Since this seems to be already in-flight

esc:
> @stuartarchibald This is related to my request for `np.putmask` in [#7374 (comment)](https://github.com/numba/numba/issues/7374#issuecomment-91362858). Since it is already in-flight; would it help to move things forward if I leave a review?
> 
> Since this seems to be already in-flight

Yes, I think an initial review might make this move forward quicker. Thank you!

kc611 (approved):
Needs a rebase otherwise functionally LGTM! Suitable tests are also present and the code works on the current main locally.

numba/tests/test_np_functions.py
```diff
@@ -571,6 +575,107 @@ def test_np_append_exceptions(self):
...
+
+        a = np.arange(3)
+        with self.assertRaises(ValueError) as raises:
+            cfunc(a, (True, False, True), np.inf)
+        self.assertIn(
+            'cannot convert non-finite value to integral',
+            str(raises.exception)
+        )
+
+        self.disable_leak_check()
```
kc611, line 668: `self.disable_leak_check()` needs to be invoked only once, hence the second call can be removed.

kc611 (changes requested):
Requesting a rebase.

PR #5791, opened 2020-06-01
Fix up no cgroups problem

stuartarchibald:
numba -s fails if a file read fails for whatever reason, this
patches up the reading region so that it logs the error and
continues.

### DIFF
--- a/numba/misc/numba_sysinfo.py
+++ b/numba/misc/numba_sysinfo.py
@@ -309,7 +309,10 @@ def get_sysinfo():
     # Python locale
     # On MacOSX, getdefaultlocale can raise. Check again if Py > 3.7.5
     try:
-        sys_info[_python_locale] = '.'.join(locale.getdefaultlocale())
+        # If $LANG is unset, getdefaultlocale() can return (None, None), make
+        # sure we can encode this as strings by casting explicitly.
+        sys_info[_python_locale] = '.'.join([str(i) for i in
+                                             locale.getdefaultlocale()])
     except Exception as e:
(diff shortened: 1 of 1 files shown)


stuartarchibald:
Buildfarm ID: `numba_smoketest_cpu_43`.

stuartarchibald:
Build `numba_smoketest_cpu_43` seems ok, but is failing on the same as #5773 so I've merge that in here and am rerunning as: Buildfarm ID: `numba_smoketest_cpu_44`.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cpu_45` is running against 7443b62. AARCH64 hasn't run yet but everything else has passed.

numba/misc/numba_sysinfo.py
```diff
@@ -186,12 +186,19 @@ class CmdReadFile(tuple):
...
-                    output.extend(out)
-            continue
+            if os.path.exists(cmd[0]):
+                try:
+                    with open(cmd[0], 'r') as f:
+                        out = f.readlines()
+                        if out:
+                            out[0] = ' '.join((cmd[0], out[0]))
+                            output.extend(out)
+                except Exception as e:
```
sklam, line 196: Should we restrict the exception class to `OSError` (which includes IOError) so that we don't silently ignore other programmatic errors?

stuartarchibald: Will do.

sklam:
This is good to merge as soon as azure ci passes

PR #5796, opened 2020-06-01
Add implementation for np.positive

### LINKED ISSUE #5786: np.positive is not implemented
Interestingly, the comment here says it doesn't exist!

https://github.com/numba/numba/blob/579db91677b7d82ef81599905c3afed55668cf17/numba/core/typing/npydecl.py#L146-L153

It was added in numpy 1.13, https://numpy.org/doc/1.18/reference/generated/numpy.positive.html

### LINKED ISSUE #5848: operator.pos fails on strided arrays
## Reporting a bug

```python
@numba.njit
def pos(x):
    return +x
pos(np.arange(6)[::2])
```
gives
```
Don't know how to allocate array with layout 'A'.

File "<ipython-input-12-72e97bb40740>", line 3:
def pos(x):
    return +x
    ^
```

niteya-shah:
closes #5786 
I have added the lowering and tests and have unecessary class. Please do tell if any changes are required. 

Fixes #5848

### DIFF
--- a/numba/np/ufunc_db.py
+++ b/numba/np/ufunc_db.py
@@ -75,6 +75,24 @@ def _fill_ufunc_db(ufunc_db):
         'D->D': numbers.complex_negate_impl,
     }
 
+    ufunc_db[np.positive] = {
+        '?->?': numbers.int_positive_impl,
+        'b->b': numbers.int_positive_impl,
+        'B->B': numbers.int_positive_impl,
+        'h->h': numbers.int_positive_impl,
+        'H->H': numbers.int_positive_impl,
+        'i->i': numbers.int_positive_impl,
+        'I->I': numbers.int_positive_impl,
+        'l->l': numbers.int_positive_impl,
+        'L->L': numbers.int_positive_impl,
+        'q->q': numbers.int_positive_impl,
+        'Q->Q': numbers.int_positive_impl,
+        'f->f': numbers.real_positive_impl,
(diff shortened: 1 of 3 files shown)


gmarkall:
Thanks for the PR! Once CI passes, I'll mark it ready for review.

gmarkall:

numba/np/ufunc_db.py
```diff
@@ -998,6 +1016,9 @@ def _fill_ufunc_db(ufunc_db):
     ufunc_db[np.negative].update({
         'm->m': npdatetime.timedelta_neg_impl,
     })
+    ufunc_db[np.positive].update({
+        'm->m': npdatetime.timedelta_pos_impl,
```
stuartarchibald, line 1020: Not here, but sometime it'd be good to wire in `M->M`, will probably need to patch the loop selection in `numpy_support` for that.

eric-wieser: That loop is deprecated:
```
In [6]: +np.datetime64('now')
<ipython-input-6-906fd83a7cf9>:1: DeprecationWarning: Applying '+' to a non-numerical array is ill-defined. Returning a copy, but in the future this will error.
  +np.datetime64('now')
Out[6]: numpy.datetime64('2020-06-11T09:26:02')
```

stuartarchibald: ah, I had np 1.15 in my env when I checked, fair enough, this was the right choice.

PR #5807, opened 2020-06-03
Skip failing test on POWER8 due to PPC CTR Loop problem.

stuartarchibald:
As title.


stuartarchibald:
Buildfarm ID: `numba_smoketest_cpu_46`.

stuartarchibald:
POWER8 builds are now successful.

PR #5808, opened 2020-06-03
Fix 5665 Block jit(nopython=True, forceobj=True) and suppress njit(forceobj=True)

### LINKED ISSUE #5665: No error for jit(nopython=True, forceobj=True)

When using the `jit` decorator with both `nopython=True` and `forceobj=True` no error is reported, despite these flags probably being mutually exclusive.
```python
import numba

@numba.jit(nopython=True, forceobj=True)
def f():
    pass

f()
```
The prior example runs without complaints. If the `njit` decorator is used instead, appropriate warnings are thrown.

HPLegion:
Fixes #5665

Currently it is possible to decorate a function with `jit(nopython=True, forceobj=True)`, this fix will raise an error if this happens. Further, `njit` claims to ignore `forceobj=True`, but currently passes it on to `jit`. This fix deletes `forceobj` from the passed `njit` kwargs.

I am happy to include some small test cases, if somebody can point me as to which file I should best put them into. :)

### DIFF
--- a/numba/core/decorators.py
+++ b/numba/core/decorators.py
@@ -144,7 +144,7 @@ def jit(signature_or_function=None, locals={}, target='cpu', cache=False,
     if 'restype' in options:
         raise DeprecationError(_msg_deprecated_signature_arg.format('restype'))
     if options.get('nopython', False) and options.get('forceobj', False):
-        raise ValueError("Only either 'nopython' or 'forcobj' can be True.")
+        raise ValueError("Only one of 'nopython' or 'forceobj' can be True.")
 
     options['boundscheck'] = boundscheck
 
--- a/numba/core/decorators.py
+++ b/numba/core/decorators.py
@@ -143,6 +143,8 @@ def jit(signature_or_function=None, locals={}, target='cpu', cache=False,
         raise DeprecationError(_msg_deprecated_signature_arg.format('argtypes'))
     if 'restype' in options:
         raise DeprecationError(_msg_deprecated_signature_arg.format('restype'))
+    if options.get('nopython', False) and options.get('forceobj', False):
+        raise ValueError("Only either 'nopython' or 'forcobj' can be True.")
 
     options['boundscheck'] = boundscheck
 
@@ -232,6 +234,7 @@ def njit(*args, **kws):
         warnings.warn('nopython is set for njit and is ignored', RuntimeWarning)
     if 'forceobj' in kws:
         warnings.warn('forceobj is set for njit and is ignored', RuntimeWarning)
+        del kws['forceobj']
     kws.update({'nopython': True})
     return jit(*args, **kws)
 
--- a/numba/tests/test_api.py
+++ b/numba/tests/test_api.py
@@ -42,11 +42,13 @@ class TestJitDecorator(TestCase):
 
         jit_func = jit(nopython=True)(py_func)
         jit_func(1)
-        self.assertTrue(jit_func.nopython_signatures)
+        # Check length of nopython_signatures to check
+        # which mode the function was compiled in
+        self.assertTrue(len(jit_func.nopython_signatures) == 1)
 
         jit_func = jit(forceobj=True)(py_func)
         jit_func(1)
-        self.assertFalse(jit_func.nopython_signatures)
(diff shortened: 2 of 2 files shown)


gmarkall:
Thanks for the PR! I think either of `numba.tests.test_jitmethod` or `numba.tests.test_api` would be appropriate places for the tests, probably in a new class.

numba/tests/test_api.py
```diff
@@ -42,11 +42,13 @@ def py_func(x):
 
         jit_func = jit(nopython=True)(py_func)
         jit_func(1)
-        self.assertTrue(jit_func.nopython_signatures)
+        # Check length of nopython_signatures to check
+        # which mode the function was compiled in
+        self.assertTrue(len(jit_func.nopython_signatures) == 1)
```
gmarkall, line 47:
```suggestion
        self.assertEqual(len(jit_func.nopython_signatures), 1)
```

numba/tests/test_api.py
```diff
@@ -42,11 +42,13 @@ def py_func(x):
...
 
         jit_func = jit(forceobj=True)(py_func)
         jit_func(1)
-        self.assertFalse(jit_func.nopython_signatures)
+        self.assertTrue(len(jit_func.nopython_signatures) == 0)
```
gmarkall, line 51:
```suggestion
        self.assertEqual(len(jit_func.nopython_signatures), 0)
```

numba/tests/test_api.py
```diff
@@ -60,11 +62,12 @@ def py_func(x):
 
         jit_func = njit(nopython=True)(py_func)
         jit_func(1)
-        self.assertTrue(jit_func.nopython_signatures)
+        self.assertTrue(len(jit_func.nopython_signatures) == 1)
```
gmarkall, line 65:
```suggestion
        self.assertEqual(len(jit_func.nopython_signatures), 1)
```

numba/tests/test_api.py
```diff
@@ -60,11 +62,12 @@ def py_func(x):
...
 
         jit_func = njit(forceobj=True)(py_func)
         jit_func(1)
-        self.assertTrue(jit_func.nopython_signatures)
+        # Since forceobj is ignored this has to compile in nopython mode
+        self.assertTrue(len(jit_func.nopython_signatures) == 1)
```
gmarkall, line 70:
```suggestion
        self.assertEqual(len(jit_func.nopython_signatures), 1)
```

HPLegion:
Thanks for your suggestions :)

HPLegion:
Sounds good :)

numba/core/decorators.py
```diff
@@ -143,6 +143,8 @@ def bar(x, y):
         raise DeprecationError(_msg_deprecated_signature_arg.format('argtypes'))
     if 'restype' in options:
         raise DeprecationError(_msg_deprecated_signature_arg.format('restype'))
+    if options.get('nopython', False) and options.get('forceobj', False):
+        raise ValueError("Only either 'nopython' or 'forcobj' can be True.")
```
stuartarchibald, line 147:
```suggestion
        raise ValueError("Only one of 'nopython' or 'forceobj' can be True.")
```

numba/tests/test_api.py
```diff
@@ -28,5 +29,46 @@ def test_numba_module(self):
...
+        jit_func = jit(forceobj=True)(py_func)
+        jit_func(1)
+        self.assertEqual(len(jit_func.nopython_signatures), 0)
+
+    def test_njit_nopython_forceobj(self):
+        with self.assertWarns(RuntimeWarning):
+            njit(forceobj=True)
+
+        with self.assertWarns(RuntimeWarning):
+            njit(nopython=True)
```
stuartarchibald, line 58: There's a preference in the Numba tests to be explicit where possible. I think that something along the lines of:
```python
with warnings.catch_warnings(record=True) as w:
  warnings.simplefilter('always', RuntimeWarning)
  <thing to trigger warning>
self.assertEqual(len(w), 1)
self.assertIn(<expected message>, str(w[0].message))
```
could help with checking the expected warning messages are raised?

HPLegion:
Hej Stuart,

thanks for your remarks, tried to implement your suggestions - let' hope that the CI is happy :)

PR #5811, opened 2020-06-03
Eliminate all dangerous default values found by pylint

### LINKED ISSUE #5809: Dangerous default arguments in numba
While reading the numba source code I noticed that in several places, dangerous default arguments are used.

When defining a function like
```python
def f(arg={}):
    pass
```
the default argument is created during the definition of the function, not during the call.

That means that all calls which fall to the default argument share the same instance of that dictionary (list, etc..)

This can easily have "funny" side effects, as shown in the following (somewhat contrived) example
```python
from numba import njit, int32

@njit
def f(x):
    c = 0.5
    return c+x
print(f(1))

f.locals["c"] = int32

@njit
def g(x):
    c = 0.5
    return c+x
print(g(1))
print(g.locals)
```
returns
```
1.5
1
{'c': int32}
```
This happens because `f` and `g` end up sharing the same instance of `locals` (the jit decorator is defined as `def jit([...], locals={})`).

HPLegion:
Proposal to fix #5809

Eliminates dangerous default values in numba (as identified by pylint)

Some of these cases where not actually dangerous in the way they were used, but I simply tried to get rid of all occurrences I found to be consistent. If there are good arguments in favour of keeping them, this can of course be arranged. :)

### DIFF
--- a/numba/core/base.py
+++ b/numba/core/base.py
@@ -826,7 +826,7 @@ class BaseContext(object):
 
         Note this context's flags are not inherited.
         """
-        locals = locals if locals else {}
+        locals = locals if locals is not None else {}
         # Compile
         from numba.core import compiler
 
@@ -857,7 +857,7 @@ class BaseContext(object):
         If *caching* evaluates True, the function keeps the compiled function
         for reuse in *.cached_internal_func*.
         """
-        locals = locals if locals else {}
+        locals = locals if locals is not None else {}
         cache_key = (impl.__code__, sig, type(self.error_model))
         if not caching:
             cached = None
@@ -883,7 +883,7 @@ class BaseContext(object):
         Like compile_subroutine(), but also call the function with the given
         *args*.
         """
-        locals = locals if locals else {}
+        locals = locals if locals is not None else {}
         cres = self.compile_subroutine(builder, impl, sig, locals)
         return self.call_internal(builder, cres.fndesc, sig, args)
 
--- a/numba/core/base.py
+++ b/numba/core/base.py
@@ -817,7 +817,7 @@ class BaseContext(object):
     def get_dummy_type(self):
         return GENERIC_POINTER
 
-    def _compile_subroutine_no_cache(self, builder, impl, sig, locals={},
+    def _compile_subroutine_no_cache(self, builder, impl, sig, locals=None,
                                      flags=None):
         """
         Invoke the compiler to compile a function to be used inside a
@@ -826,6 +826,7 @@ class BaseContext(object):
 
         Note this context's flags are not inherited.
         """
+        locals = locals if locals else {}
         # Compile
         from numba.core import compiler
 
@@ -847,7 +848,7 @@ class BaseContext(object):
             self.active_code_library.add_linking_library(cres.library)
             return cres
 
-    def compile_subroutine(self, builder, impl, sig, locals={}, flags=None,
+    def compile_subroutine(self, builder, impl, sig, locals=None, flags=None,
                            caching=True):
         """
         Compile the function *impl* for the given *sig* (in nopython mode).
@@ -856,6 +857,7 @@ class BaseContext(object):
         If *caching* evaluates True, the function keeps the compiled function
         for reuse in *.cached_internal_func*.
         """
+        locals = locals if locals else {}
         cache_key = (impl.__code__, sig, type(self.error_model))
         if not caching:
             cached = None
@@ -876,11 +878,12 @@ class BaseContext(object):
         self.active_code_library.add_linking_library(cres.library)
         return cres
 
-    def compile_internal(self, builder, impl, sig, args, locals={}):
+    def compile_internal(self, builder, impl, sig, args, locals=None):
         """
         Like compile_subroutine(), but also call the function with the given
         *args*.
         """
+        locals = locals if locals else {}
         cres = self.compile_subroutine(builder, impl, sig, locals)
         return self.call_internal(builder, cres.fndesc, sig, args)
 
--- a/numba/core/base.py
+++ b/numba/core/base.py
@@ -1,4 +1,5 @@
 from collections import namedtuple, defaultdict
+from types import MappingProxyType
 import copy
 import os
 import sys
@@ -826,7 +827,7 @@ class BaseContext(object):
 
         Note this context's flags are not inherited.
         """
-        locals = locals if locals is not None else {}
+        locals = MappingProxyType({} if locals is None else locals)
         # Compile
         from numba.core import compiler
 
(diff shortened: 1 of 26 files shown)


gmarkall:
Thanks for the PR!

Out of interest:

- Does pylint show up other kinds of errors?
- Is running pylint the sort of thing that would make sense to do as part of PR testing? (i.e. is it fast and lightweight, and not too many false positives?)

HPLegion:
Pylint with standard settings tends to be very trigger happy, I can do a warnings/error only run and see what comes back and how long it takes :)

HPLegion:
Aw man, so close :-D Will go and scratch my head about what went wrong here :)

gmarkall:
I just had a quick skim and I don't know the parfors parts of Numba well, but I wonder if there was some reliance on the "dangerous" behaviour here: https://github.com/numba/numba/pull/5811/files#diff-054901b0779b9ed9f103e763d25f872fL1353

HPLegion:
I will experiment locally and see if rolling back those changes makes the test pass :)

HPLegion:
Let's wait for the CI but locally reverting that change seems to make the test pass. 

I obviously went through all these occurences without thinking to much about how every function acts within numba. But I think it still raises the question of whether this should be the intended behaviour. If there is the intention to share that dictionary instance, maybe it should be made clearer?

gmarkall:
> I obviously went through all these occurences without thinking to much about how every function acts within numba. But I think it still raises the question of whether this should be the intended behaviour. If there is the intention to share that dictionary instance, maybe it should be made clearer?

Yes, it should do - I'm not familiar with that area of the code, but that should be made clearer and more explicit if sharing is relied on.

HPLegion:
Ah I think I figured it out:
I took the lazy route of checking `if swapped` instead of `if swapped is not None`.
If an empty dict (list etc.) is passed, it will actually be replaced by another empty dict. So if the caller relies on holding a reference to an empty container this screws things up.

gmarkall:
Good find! I'll mark the PR as ready for review.

gmarkall:
(when the commit changing it is pushed and passed CI, that is)

eric-wieser:
There's another way you could handle this which would add far fewer lines:
* Use `types.MappingProxyType({})` in place of `{}`
* Use `()` in place of `[]`

HPLegion:
That would probably be a neat solution on a case by case basis, agreed :)

But then one should make sure that those default values have a very limited life-cycle. As soon as they get passed on to other functions or objects it becomes difficult to predict how they should behave, right? Maybe in some of those cases the containers have to be mutable, it was just not intended to share the default instance among several consumers.

eric-wieser:
> Maybe in some of those cases the containers have to be mutable, it was just not intended to share the default instance among several consumers.

Those functions ought to make very clear in their docstrings that they mutate the input in-place. So I'd recommend:

* Applying the change I suggest above everywhere
* Noting the failures, updating the docstring to clearly indicate that the argument is mutated, and then reverting those cases (if any) to your approach.

HPLegion:
Maybe it is a good idea to get the opinion of @stuartarchibald or another core maintainer here, since this is mostly a (long term) maintenance question.

In principle I like Eric's ideas but I see some issues with this:
1) By accident I found at least one example (the parfor test that failed above) where the passed argument is mutated intentionally and the caller relies on this. The argument that this should probably be documented is 100% valid.
2) In some cases the default argument is passed on to a long living object (like in the original issue I reported). I don't think it is a good idea to pass e.g. a dictionary (supplied by the user) to some jitted functions and a MappingProxyType (default) to others; this is inconsistent and could lead to confusion down the line.

That said, do you prefer a "one fits all" solution to this, or is it better to decide on the fix on a case by case basis?

sklam:
I like @eric-wieser's suggestion (https://github.com/numba/numba/pull/5811#issuecomment-642526144). Changing to immutable objects is preferred for long term maintenance, given that we already know there are too many unwanted mutations inside the compiler. 

I agree with @HPLegion on:

> I don't think it is a good idea to pass e.g. a dictionary (supplied by the user) to some jitted functions and a MappingProxyType (default) to others; this is inconsistent and could lead to confusion down the line.

Maybe we can do it like:

```python
def function(arg=None):
    d = MappingProxyType({} if arg is None else arg)
    ...
```

Also, I think we should use more `MappingProxyType` internally.

For now, it's good to know what API is mutating dictionaries that are passed in.

eric-wieser:
> I like @eric-wieser's suggestion ([#5811 (comment)](https://github.com/numba/numba/pull/5811#issuecomment-642526144)). Changing to immutable objects is preferred for long term maintenance, given that we already know there are too many unwanted mutations inside the compiler.
> 
> I agree with @HPLegion on:
> 
> > I don't think it is a good idea to pass e.g. a dictionary (supplied by the user) to some jitted functions and a MappingProxyType (default) to others; this is inconsistent and could lead to confusion down the line.
> 
> Maybe we can do it like:
> 
> ```python
> def function(arg=None):
>     d = MappingProxyType({} if arg is None else arg)
>     ...
> ```

Or even, in the functions where the intent is immutability:
```python
def function(arg={}):
    arg = MappingProxyType(arg)
```

HPLegion:
I like that last approach suggested by Eric :)
I suppose one could even contemplate using a similar approach for lists by casting them to tuples.

I'll have a shot at this and then we can see if and where things start smoking

numba/core/base.py
```diff
@@ -817,7 +818,7 @@ def get_dummy_value(self):
     def get_dummy_type(self):
         return GENERIC_POINTER
 
-    def _compile_subroutine_no_cache(self, builder, impl, sig, locals={},
+    def _compile_subroutine_no_cache(self, builder, impl, sig, locals=None,
```
eric-wieser, line 821: No need to change this any more

eric-wieser: (same throughout)

numba/core/base.py
```diff
@@ -826,6 +827,7 @@ def _compile_subroutine_no_cache(self, builder, impl, sig, locals={},
 
         Note this context's flags are not inherited.
         """
+        locals = MappingProxyType({} if locals is None else locals)
```
eric-wieser, line 830: Which makes this
```suggestion
        locals = MappingProxyType(locals)  # avoid mutation
```

HPLegion:
Thanks for your remarks Eric, you are of course right :)

The bigger problem seems to be that "locals" are being mutated somewhere (if I interpret the tracebacks on Azure correctly). Probably won't have time to look at this again before the weekend though...

HPLegion:
/cc @eric-wieser @sklam 

I honestly think that I won't get around to working on this diligently any time soon, so if there is a wish to finish this up for the sake of general immutability (rather than just the default value problem) someone else is welcome to pick this up. Otherwise I am happy to return to this at some point but I don't really know when that will be (if only dissertations would write themselves :-P).

Should we close the issue for the time being to not clutter the issue page?

sklam:
Closing this for now.

Issue #5792, opened 2020-06-01
Not work  jit_options={'error_model': 'numpy'}


Rubtsowa:
import numba
from numba import cgutils, types
from numba.extending import (intrinsic, make_attribute_wrapper, models,
                             overload_method, overload, register_model)
from numba.typing import signature

def divide(a, b):
    pass

@overload(divide, jit_options={'error_model': 'numpy'})
def divide_overload(a, b):
    def _impl(a, b):
        return a / b
    return _impl

@numba.njit(nopython=True)
def test_divide(a, b):
    return divide(a, b)

print("TEST1: ", test_divide(1, 0))

class Dummy:
    def __init__(self, data):
        self.data = data

    def div(self, other):
        return data / other.data

class DummyType(types.Type):
    def __init__(self, data):
        self.data = data
        super().__init__(name=f'Dummy({self.data})')

@register_model(DummyType)
class DummyTypeModel(models.StructModel):
    def __init__(self, dmm, fe_type):
        members = [
            ('data', fe_type.data),
        ]
        super().__init__(dmm, fe_type, members)

make_attribute_wrapper(DummyType, 'data', '_data')

@intrinsic
def init_dummy(typingctx, data):
    def codegen(context, builder, sig, args):
        typ = sig.return_type
        data, = args
        dummy = cgutils.create_struct_proxy(typ)(context, builder)
        dummy.data = data

        if context.enable_nrt:
            context.nrt.incref(builder, sig.args[0], data)

        return dummy._getvalue()

    ret_typ = DummyType(data)
    sig = signature(ret_typ, data)

    return sig, codegen

@overload(Dummy, inline='always')
def dummy_overload(data):
    def ctor(data):
        return init_dummy(data)

    return ctor

@overload_method(DummyType, 'div', inline='always', jit_options={'error_model': 'numpy'})
def div_overload(self, other):
    def impl(self, other):
        return self._data / other._data

    return impl

@numba.njit(parallel=True)
def test_impl(data, other_data):
    dummy = Dummy(data)
    other = Dummy(other_data)

    return dummy.div(other)

data = 1
other_data = 0
res = test_impl(data, other_data)
print("TEST2: ", res)

**_Output:_**
TEST1:  inf

Traceback (most recent call last):
  File "repr.py", line 98, in <module>
    res = test_impl(data, other_data)
ZeroDivisionError: division by zero

stuartarchibald:
Thanks for the report, I can reproduce.

[...]

PR #5812, opened 2020-06-03
Fix side issue from #5792, @overload inliner cached IR being mutated.

stuartarchibald:
As title. This prevents dangerous mutation of the IR cache associated
with a given overload.

### DIFF
--- a/numba/core/inline_closurecall.py
+++ b/numba/core/inline_closurecall.py
@@ -1,4 +1,5 @@
 import types as pytypes  # avoid confusion with numba.types
+import copy
 import ctypes
 import numba.core.analysis
 from numba.core import utils, types, typing, errors, ir, rewrites, config, ir_utils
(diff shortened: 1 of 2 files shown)


numba/core/typed_passes.py
```diff
@@ -675,10 +675,16 @@ def _run_inliner(
             arg_typs = template._inline_overloads[arg_typs]['folded_args']
             iinfo = template._inline_overloads[arg_typs]['iinfo']
             freevars = iinfo.func_ir.func_id.func.__code__.co_freevars
+            # operate on the closest thing we have to a copy
+            # the IR is cached on the overload template
+            # the inliner working parts are notorious for mutating IR directly
+            blocks_copy = {k: v.copy() for k, v in iinfo.func_ir.blocks.items()}
+            func_ir_copy = iinfo.func_ir.copy()
+            func_ir_copy.blocks = blocks_copy
```
sklam, line 683: I don't think the `blocks` need to be copied explicitly. `func_ir.copy()` has already done that. 

```suggestion
            func_ir_copy = iinfo.func_ir.copy()
```

the test passes with the change.

stuartarchibald: Good point, will patch.

sklam:
`numba_smoketest_cpu_47`

numba/core/typed_passes.py
stuartarchibald: I ended up having to do a much more involved copy.

sklam:
`numba_smoketest_cpu_47` passed (only failed in entry-points tests, known buildfarm problem).

stuartarchibald:
I've since force pushed to remove the partial copy in favour of a full copy, so we better rerun the buildfarm.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cpu_48` is for d5edcdf

stuartarchibald:
`numba_smoketest_cpu_48` passed (only expected entry-point test fails), @ehsantn also confirmed that [d5edcdf](https://github.com/numba/numba/commit/d5edcdfb9af7ab70de02c606a1c1baa77de671bd) worked for Bodo here https://github.com/numba/numba/issues/5824#issuecomment-641682583.

PR #5819, opened 2020-06-04
Add support for CUDA 11 and Ampere / CC 8.0

### LINKED ISSUE #5804: test_const_record_align failing with CUDA 10.1/10.2/11.0
Code generation is slightly different with CUDA 10.1 onwards, so the expected output for `numba.cuda.tests.cudapy.test_constmem.TestCudaConstantMemorytest_const_record_align` differs from expected with it:

```
FAIL: test_const_record_align (numba.cuda.tests.cudapy.test_constmem.TestCudaConstantMemory)
  File "/home/gmarkall/numbadev/numba/numba/cuda/tests/cudapy/test_constmem.py", line 182, in test_const_record_align
    self.assertIn(
AssertionError: 'ld.const.v4.u8' not found in '//\n// Generated by NVIDIA NVVM Compiler...
<lots more output>
```

## Reporting a bug

- [ ] I am using the latest released version of Numba (most recent is visible in

gmarkall:
With CUDA 11, we need to switch to determining the supported compute capabilities using the runtime version, because:

- CC 8.0 is supported by CUDA 11.0, and not 10.2.
- The NVVM version and its IR version, as well as the soversion, are all  constant between CUDA 10.2 and 11.0.

Therefore, the existing mechanism that looks at the IR version was not sufficient to determine if CC 8.0 is supported (and also 3.0 and 3.2 unsupported, as in CUDA 11.0). It is expected that the cudart version will always be consistent with the NVVM version in the environment in which Numba is used, so the runtime version is queried instead.

This PR adds a small wrapper over the Runtime API to allow the version to be queried. It also changes the determination of the supported compute capabilities to be based on the Runtime API version rather than the NVVM IR version.

Having the Runtime API version query makes it easy to fix #5804 in a way that is precise and works for all supported toolkits, so this PR also includes a fix for that issue.

A little more detail can be found in the individual commit messages.

### DIFF
--- a/numba/cuda/cudadrv/nvvm.py
+++ b/numba/cuda/cudadrv/nvvm.py
@@ -274,21 +274,27 @@ default_data_layout = data_layout[tuple.__itemsize__ * 8]
 
 
 try:
-    NVVM_VERSION = NVVM().get_version()
+    from numba.cuda.cudadrv.runtime import runtime
+    cudart_version_major = runtime.get_version()[0]
 except:
-    # the CUDA driver may not be present
-    NVVM_VERSION = (0, 0)
+    # The CUDA Runtime may not be present
+    cudart_version_major = 0
 
 # List of supported compute capability in sorted order
-if NVVM_VERSION < (1, 4):
-    # CUDA 8.0
+if cudart_version_major == 0:
+    SUPPORTED_CC = (),
+elif cudart_version_major < 9:
+    # CUDA 8.x
     SUPPORTED_CC = (2, 0), (2, 1), (3, 0), (3, 5), (5, 0), (5, 2), (5, 3), (6, 0), (6, 1), (6, 2)
-elif NVVM_VERSION < (1, 5):
-    # CUDA 9.0 and later
+elif cudart_version_major < 10:
+    # CUDA 9.x
     SUPPORTED_CC = (3, 0), (3, 5), (5, 0), (5, 2), (5, 3), (6, 0), (6, 1), (6, 2), (7, 0)
-else:
-    # CUDA 10.0 and later
+elif cudart_version_major < 11:
+    # CUDA 10.x
     SUPPORTED_CC = (3, 0), (3, 5), (5, 0), (5, 2), (5, 3), (6, 0), (6, 1), (6, 2), (7, 0), (7, 2), (7, 5)
+else:
+    # CUDA 11.0 and later
+    SUPPORTED_CC = (3, 5), (5, 0), (5, 2), (5, 3), (6, 0), (6, 1), (6, 2), (7, 0), (7, 2), (7, 5), (8, 0)
 
 
 def find_closest_arch(mycc):
--- a/numba/cuda/cudadrv/runtime.py
+++ b/numba/cuda/cudadrv/runtime.py
@@ -1,0 +1,123 @@
+"""
+CUDA Runtime wrapper.
+
+This provides a very minimal set of bindings, since the Runtime API is not
+really used in Numba except for querying the Runtime version.
+"""
+
+import ctypes
+import functools
+
+from numba.core import config
+from numba.cuda.cudadrv.driver import ERROR_MAP, make_logger
+from numba.cuda.cudadrv.error import CudaSupportError, CudaRuntimeError
+from numba.cuda.cudadrv.libs import open_cudalib
+from numba.cuda.cudadrv.rtapi import API_PROTOTYPES
+from numba.cuda.cudadrv import enums
+
+
+class CudaRuntimeAPIError(CudaRuntimeError):
+    """
+    Raised when there is an error accessing a C API from the CUDA Runtime.
+    """
+    def __init__(self, code, msg):
+        self.code = code
+        self.msg = msg
+        super().__init__(code, msg)
+
+    def __str__(self):
+        return "[%s] %s" % (self.code, self.msg)
+
+
+class Runtime:
+    """
+    Runtime object that lazily binds runtime API functions.
+    """
+
+    def __init__(self):
+        self.is_initialized = False
+        try:
+            if config.DISABLE_CUDA:
+                msg = ("CUDA is disabled due to setting NUMBA_DISABLE_CUDA=1 "
+                       "in the environment, or because CUDA is unsupported on "
+                       "32-bit systems.")
+                raise CudaSupportError(msg)
+            self.lib = open_cudalib('cudart')
+            self.load_error = None
+        except CudaSupportError as e:
+            self.load_error = e
+
+    def _initialize(self):
+        # lazily initialize logger
+        global _logger
+        _logger = make_logger()
+        self.is_initialized = True
+
+    def __getattr__(self, fname):
+        # First request of a runtime API function
+        try:
+            proto = API_PROTOTYPES[fname]
+        except KeyError:
+            raise AttributeError(fname)
+        restype = proto[0]
+        argtypes = proto[1:]
+
+        if not self.is_initialized:
+            self._initialize()
+
+        if self.load_error is not None:
+            raise CudaSupportError("Error at runtime load: \n%s:" %
+                                   self.load_error)
+
+        # Find function in runtime library
+        libfn = self._find_api(fname)
+        libfn.restype = restype
+        libfn.argtypes = argtypes
+
+        safe_call = self._wrap_api_call(fname, libfn)
+        setattr(self, fname, safe_call)
+        return safe_call
+
+    def _wrap_api_call(self, fname, libfn):
+        @functools.wraps(libfn)
+        def safe_cuda_api_call(*args):
+            _logger.debug('call runtime api: %s', libfn.__name__)
+            retcode = libfn(*args)
+            self._check_error(fname, retcode)
+        return safe_cuda_api_call
+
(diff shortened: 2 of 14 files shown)


stuartarchibald:
Thanks for the patch. Note: probably need to wait for a `cudatoolkit` package from the Anaconda distro team to test and integrate testing into the build farm.

gmarkall:
Can this be merged with a passing run on the existing buildfarm? As far as I am aware, the buildfarm doesn't have CUDA 10.2 either.

numba/cuda/cudadrv/runtime.py
```diff
@@ -0,0 +1,117 @@
...
+        """
+        rtver = ctypes.c_int()
+        self.cudaRuntimeGetVersion(ctypes.byref(rtver))
+        # The version is encoded as (1000 * major) + (10 * minor)
+        major = rtver.value // 1000
+        minor = (rtver.value - (major * 1000)) // 10
+        return (major, minor)
+
+
+runtime = Runtime()
```
stuartarchibald, line 117: Maybe then `del Runtime` ? Presume only one can exist?

numba/cuda/cudadrv/nvvm.py
```diff
@@ -274,21 +274,27 @@ def get_log(self):
...
 
 try:
-    NVVM_VERSION = NVVM().get_version()
+    from numba.cuda.cudadrv.runtime import runtime
+    cudart_version_major = runtime.get_version()[0]
 except:
-    # the CUDA driver may not be present
-    NVVM_VERSION = (0, 0)
+    # The CUDA Runtime may not be present
+    cudart_version_major = 0
```
stuartarchibald, line 281: Wondering if it makes sense to continue without a RT?

numba/cuda/cudadrv/nvvm.py
```diff
@@ -274,21 +274,27 @@ def get_log(self):
...
+    # CUDA 9.x
     SUPPORTED_CC = (3, 0), (3, 5), (5, 0), (5, 2), (5, 3), (6, 0), (6, 1), (6, 2), (7, 0)
-else:
-    # CUDA 10.0 and later
+elif cudart_version_major < 11:
+    # CUDA 10.x
     SUPPORTED_CC = (3, 0), (3, 5), (5, 0), (5, 2), (5, 3), (6, 0), (6, 1), (6, 2), (7, 0), (7, 2), (7, 5)
+else:
+    # CUDA 11.0 and later
+    SUPPORTED_CC = (3, 5), (5, 0), (5, 2), (5, 3), (6, 0), (6, 1), (6, 2), (7, 0), (7, 2), (7, 5), (8, 0)
```
stuartarchibald, line 297: wondering if it'd be an improvement to just list all of these CC's and then slice the list? it might not be, guess the  disadvantage would be that it might be harder to just look up the supported CCs for a given CTK version.

numba/cuda/cudadrv/nvvm.py
gmarkall: If there's no runtime you can't use NVVM, but how does "not continuing" differ from "continuing" here?

numba/cuda/cudadrv/nvvm.py
gmarkall: I do prefer to see explicitly what Numba thinks different toolkits support - I've referred to this a few times in the past, and mentally evaluating a slice would be harder work.

numba/cuda/cudadrv/runtime.py
gmarkall: Technically more than one could exist - I can only see debugging as a reason for doing so though. This follows the pattern in `driver.py` where classes are instantiated and then not deleted. Is it a common pattern to delete things from module-level scope if you don't expect more than one of them to be constructed?

gmarkall:
Thanks for the review! I've added responses to the comments.

numba/cuda/cudadrv/nvvm.py
stuartarchibald: agree, happy to leave as-is.

numba/cuda/cudadrv/runtime.py
stuartarchibald: There's a vague propensity but nothing more, it's fine to leave RE debugging.

numba/cuda/cudadrv/nvvm.py
stuartarchibald: I suppose the question is more, from where does the "this isn't going to work" exception get raised, any reason for it not to be here?

stuartarchibald:
I've built a CUDA 11 RC toolkit from https://github.com/AnacondaRecipes/cudatoolkit-feedstock/pull/7 and installed the CUDA 11 RC drivers from https://developer.nvidia.com/cuda-downloads, then tested this patch with that, seems fine. FTR: `| NVIDIA-SMI 450.36.06    Driver Version: 450.36.06    CUDA Version: 11.0     |
`

numba/cuda/cudadrv/nvvm.py
gmarkall: This gets run when Numba is imported, so I don't think we want any exceptions raising here. Usually there'd be an exception on an attempt to compile a kernel, because the list of supported compute capabilities will be empty (as set in the following lines).

stuartarchibald: I see, thanks for clarifying, it's a bit strange in that no CC would not be the real problem but a side effect? However, I think fixing/altering is out of scope here. Happy to leave as-is.

stuartarchibald:
Random other thought... are there any docs about version support that need updating?

gmarkall:
I don't think so - the docs suggest CUDA 8 or later: https://numba.pydata.org/numba-doc/latest/cuda/overview.html#software

The documentation for CUDA could do with a re-working, but I think in the context of this PR they were and remain correct.

stuartarchibald:
Great, thanks for confirming!

numba/cuda/cudadrv/runtime.py
```diff
@@ -0,0 +1,117 @@
...
+
+        setattr(self, fname, absent_function)
+        return absent_function
+
+    def get_version(self):
+        """
+        Returns the CUDA Runtime version as a tuple (major, minor).
+        """
+        rtver = ctypes.c_int()
+        self.cudaRuntimeGetVersion(ctypes.byref(rtver))
```
sklam, line 110: Can we use the CUDA driver api for this?

sklam: i.e. `cuDriverGetVersion`

leofang: The driver version and the runtime version can be different. You can have a driver supporting 11.0, but have CUDA Toolkit 10.2 installed.

esc:
BFID: `numba_smoketest_cuda_75`

esc:
BF passed, marking as RTM.

jakirkham:
If this gets merged, would nightlies be produced? We are interested in trying them out on a machine with CUDA 11 support 🙂

cc @kkraus14 @mike-wendt @quasiben

quasiben:
My understanding from NVIDIA is that if hardware/cuda 11 access is necessary they can provide a GCP instance with Ampere support

esc:
@jakirkham I'll run the new `master` through CI now, if all goes well, nightly should appear by the morning 🤞

jakirkham:
Great, thanks Valentin! 😄

quasiben:
Thank you @esc

esc:
FYI: the nightlies (if there are any) will arrive here: https://anaconda.org/numba/numba/files

So you can grab them using:

```
conda install -c numba/label/dev numba
```

esc:
Unfortunately the nightlies were produced but not uploaded due to a user-error. Hopefully the next attempt tonight will be more fruitful.

esc:
Happy to report though that `master` passed all tests on all platforms, so that's at least something.

jakirkham:
Thanks for the update. Do they happen to live on a machine where an upload could be retried? Or do they get cleaned up after the build?

esc:
@jakirkham in principle yes, since we use anaconda.org as an intermediary build artifact storage right now. So it's just a matter of finding the correct packages and propagating the to the `numba/label/dev` channel.

esc:
@jakirkham if you are feeling adventurous you could try pointing your `conda install` at https://anaconda.org/seibert/numba/files -- but I'm not exactly sure what you would end up getting..

esc:
The hash of last nights master  is `bbb10ebfaabcf0c691ee822ef2c861c92ff3585a` and you can see the relevant files in the following screenshot:

Screen Shot 2020-07-29 at 11 12 21

Perhaps, for the time being, you can just manually get the exact package you need from there?  If possible I'd like to avoid manually propagating packages since it can be very error prone and I'd like to avoid doing any accidental damage.

jakirkham:
Thanks Valentin! 😄

Yep that makes sense. I think this is a unique situation 😉

esc:
I actually ran a nightly during the day and it started producing results and uploading builds to:

https://anaconda.org/numba/numba/files

So, there is now a good chance that the OSs you are interested in can be obtained from there.

jakirkham:
That's great! Thanks Valentin! Will take a look 😄

PR #5821, opened 2020-06-05
setup.py: add option to disable tbb at compile time

epsilon-0:
adds an option to ignore the tbb libraries if present
this is needed when an older version of tbb is present
which gets "frosty thomson'd" - stuartarchibald

Signed-off-by: Aisha Tammy <gentoo@aisha.cc>

### DIFF
--- a/docs/source/user/installing.rst
+++ b/docs/source/user/installing.rst
@@ -185,18 +185,28 @@ vary with target operating system and hardware. The following lists them all
   * ``numpy``
   * ``llvmlite``
   * Compiler toolchain mentioned above
-  * OpenMP C headers and runtime libraries compatible with the compiler
-    toolchain mentioned above and accessible to the compiler via standard flags
-    (Linux, Windows).
 
-* Optional build time:
-
-  * ``llvm-openmp`` (OSX) - provides headers for compiling OpenMP support into
+* Build time toggles:
+  * ``NUMBA_NO_OPENMP=1``
+    DEFAULT: undefined
+    set the environment variable to ``1`` when building using ``setup.py``
+    to disable compilations of OpenMP threading backend.
+    If enabled you need OpenMP C headers and runtime libraries compatible
+    with the compiler toolchain mentioned above and accessible to the
+    compiler via standard flags (Linux, Windows)
+    * ``llvm-openmp`` (OSX) - provides headers for compiling OpenMP support into
     Numba's threading backend
-  * ``intel-openmp`` (OSX) - provides OpenMP library support for Numba's
+    * ``intel-openmp`` (OSX) - provides OpenMP library support for Numba's
     threading backend.
-  * ``tbb-devel`` - provides TBB headers/libraries for compiling TBB support
-    into Numba's threading backend
+  * ``NUMBA_NO_TBB=1 TBBROOT=/usr``
+    DEFAULT: both undefined
+    set the environment variable ``NUMBA_NO_TBB`` to ``1`` when building using
+    ``setup.py`` to disable the compilation of the TBB threading backend.
+    If not setting it to ``1`` you can also add the location where the
+    TBB headers are going to be searched for, which is generally in ``/usr``,
+    as they are installed in ``/usr/include/tbb/tbb.h``.
+    If enabled ``tbb-devel`` is a requirement. It provides TBB headers/libraries
+    for compiling TBB support into Numba's threading backend
 
 * Required run time:
 
@@ -297,4 +307,3 @@ further information.
                                   pci bus id: 1
 
 (output truncated due to length)
-
--- a/docs/source/user/installing.rst
+++ b/docs/source/user/installing.rst
@@ -188,7 +188,7 @@ vary with target operating system and hardware. The following lists them all
 
 * Build time environment variables:
   * ``NUMBA_NO_OPENMP``
-    Set this environment variable to ``1`` when building using ``setup.py``
+    Set this environment variable to ``1`` when building
     to disable compilation of the OpenMP threading backend.
     If enabled you need OpenMP C headers and runtime libraries compatible
     with the compiler toolchain mentioned above and accessible to the
@@ -198,11 +198,13 @@ vary with target operating system and hardware. The following lists them all
     * ``intel-openmp`` (OSX) - provides OpenMP library support for Numba's
     threading backend.
   * ``NUMBA_NO_TBB``
-    Set this environment variable to ``1`` when building using
-    ``setup.py`` to disable the compilation of the TBB threading backend.
+    Set this environment variable to ``1`` when building
+    to disable the compilation of the TBB threading backend.
     If not setting it to ``1`` you can also add the location where the
     TBB headers are going to be searched for, which is generally in ``/usr``,
-    as they are installed in ``/usr/include/tbb/tbb.h``.
+    as they are installed in ``/usr/include/tbb/tbb.h`` for linux, by setting 
+    ``TBBROOT=/usr`` or as required based on your platform: `Intel documentation
+    <https://software.intel.com/content/www/us/en/develop/documentation/advisor-user-guide/top/appendix/adding-parallelism-to-your-program/adding-the-parallel-framework-to-your-build-environment/defining-the-tbbroot-environment-variable.html>`_  
     If enabled ``tbb-devel`` is a requirement. It provides TBB headers/libraries
     for compiling TBB support into Numba's threading backend
 
--- a/docs/source/user/installing.rst
+++ b/docs/source/user/installing.rst
@@ -186,11 +186,10 @@ vary with target operating system and hardware. The following lists them all
   * ``llvmlite``
   * Compiler toolchain mentioned above
 
-* Build time toggles:
-  * ``NUMBA_NO_OPENMP=1``
-    DEFAULT: undefined
-    set the environment variable to ``1`` when building using ``setup.py``
-    to disable compilations of OpenMP threading backend.
+* Build time environment variables:
+  * ``NUMBA_NO_OPENMP``
+    Set this environment variable to ``1`` when building using ``setup.py``
+    to disable compilation of the OpenMP threading backend.
     If enabled you need OpenMP C headers and runtime libraries compatible
     with the compiler toolchain mentioned above and accessible to the
     compiler via standard flags (Linux, Windows)
@@ -198,9 +197,8 @@ vary with target operating system and hardware. The following lists them all
     Numba's threading backend
     * ``intel-openmp`` (OSX) - provides OpenMP library support for Numba's
     threading backend.
-  * ``NUMBA_NO_TBB=1 TBBROOT=/usr``
-    DEFAULT: both undefined
-    set the environment variable ``NUMBA_NO_TBB`` to ``1`` when building using
+  * ``NUMBA_NO_TBB``
+    Set this environment variable to ``1`` when building using
     ``setup.py`` to disable the compilation of the TBB threading backend.
     If not setting it to ``1`` you can also add the location where the
     TBB headers are going to be searched for, which is generally in ``/usr``,
--- a/setup.py
+++ b/setup.py
@@ -1,10 +1,11 @@
-from setuptools import setup, Extension, find_packages
-from distutils.command import build
-from distutils.spawn import spawn
-from distutils import sysconfig
-import sys
 import os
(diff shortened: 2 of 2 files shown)


epsilon-0:
@stuartarchibald thanks for all your help today :smile_cat:

gmarkall:
Thanks for the PR! I'll queue it for review once the CI has passed.

stuartarchibald:
xref: https://gitter.im/numba/numba?at=5eda49f2f0b8a2053ad6b86f

Just to be clear, am guessing at the problem from the description given over IRC (essentially testing hangs if TBB present), the "Frosty Thompson" problem is also independent of TBB version. The minimum version of 2019.5 is to ensure certain TBB features in use in Numba are present for all OS and this has no bearings on the chances of certain tests hanging. I'd still advise compiling in TBB support, and then [configuring/disabling it at runtime](http://numba.pydata.org/numba-doc/dev/user/threading-layer.html#setting-the-threading-layer) via one of the [many configuration methods Numba provides](http://numba.pydata.org/numba-doc/dev/reference/envvars.html#environment-variables) (including a config file, environment variables and the `numba.config` module).

setup.py
```diff
@@ -213,34 +209,43 @@ def check_file_at_path(path2file):
...
-                os.path.join(tbb_root, 'lib', 'intel64', 'gcc4.4'),
-                # for MacOS
-                os.path.join(tbb_root, 'lib'),
-                # for Windows
-                os.path.join(tbb_root, 'lib', 'intel64', 'vc_mt'),
-            ],
-        )
-        ext_np_ufunc_backends.append(ext_np_ufunc_tbb_backend)
+    # Disable tbb if forced by user with NUMBA_NO_TBB=1
+    if os.getenv("NUMBA_NO_TBB") == "1":
```
stuartarchibald, line 213: Need it be "1", won't anything True-ish work ok?

setup.py
```diff
@@ -213,34 +209,43 @@ def check_file_at_path(path2file):
...
+                ],
+            )
+            ext_np_ufunc_backends.append(ext_np_ufunc_tbb_backend)
+        else:
+            print("TBB not found")
 
     # Disable OpenMP if we are building a wheel or
     # forced by user with NUMBA_NO_OPENMP=1
-    if is_building_wheel() or os.getenv('NUMBA_NO_OPENMP'):
+    if is_building_wheel() or os.getenv('NUMBA_NO_OPENMP') == "1":
```
stuartarchibald, line 248: Same comment as above RE "1".

stuartarchibald:
NOTE: Checked manually:
```
$ rm -rf build > /dev/null 2>&1; find numba -name "*.so"|xargs rm > /dev/null 2>&1; python setup.py build_ext --inplace > /dev/null 2>&1; python setup.py develop > /dev/null 2>&1; python -m numba -s|grep TBB
TBB Threading Layer Available                 : True
+-->TBB imported successfully.

$ export NUMBA_NO_TBB=1

$ rm -rf build > /dev/null 2>&1; find numba -name "*.so"|xargs rm > /dev/null 2>&1; python setup.py build_ext --inplace > /dev/null 2>&1; python setup.py develop > /dev/null 2>&1; python -m numba -s|grep TBB
TBB Threading Layer Available                 : False
NUMBA_NO_TBB                                  : 1

$ git rev-parse HEAD
1e5ef9c35b4688995bcd36b2cb791055ad72c5ce
```

stuartarchibald:
Also think that it'd be good if `NUMBA_NO_OPENMP` and `NUMBA_NO_TBB`  were documented here https://github.com/numba/numba/blame/53862931abcc4a00674a1e9c3c3730d132e33e43/docs/source/user/installing.rst#L166 as an addendum to the "Installing from source" section.

setup.py
epsilon-0: It is a string so any non empty string would work. 
I think keeping it to `"1"` makes it easy to keep track of.
Stylistic choice, any change is fine.

epsilon-0:
> python -m numba -s|grep TBB

so this is a good thing right ?
it seems to be working as intended ?

epsilon-0:
done

docs/source/user/installing.rst
```diff
@@ -185,18 +185,28 @@ vary with target operating system and hardware. The following lists them all
...
   * ``llvmlite``
   * Compiler toolchain mentioned above
-  * OpenMP C headers and runtime libraries compatible with the compiler
-    toolchain mentioned above and accessible to the compiler via standard flags
-    (Linux, Windows).
 
-* Optional build time:
-
-  * ``llvm-openmp`` (OSX) - provides headers for compiling OpenMP support into
+* Build time toggles:
```
stuartarchibald, line 189:
```suggestion
* Build time environment variables:
```

docs/source/user/installing.rst
```diff
@@ -185,18 +185,28 @@ vary with target operating system and hardware. The following lists them all
...
+  * ``NUMBA_NO_OPENMP=1``
```
stuartarchibald, line 190:
```suggestion
  * ``NUMBA_NO_OPENMP``
```

docs/source/user/installing.rst
```diff
@@ -185,18 +185,28 @@ vary with target operating system and hardware. The following lists them all
...
+    DEFAULT: undefined
```
stuartarchibald, line 191:
```suggestion
```

docs/source/user/installing.rst
```diff
@@ -185,18 +185,28 @@ vary with target operating system and hardware. The following lists them all
...
+    set the environment variable to ``1`` when building using ``setup.py``
+    to disable compilations of OpenMP threading backend.
```
stuartarchibald, line 193:
```suggestion
    to disable compilation of the OpenMP threading backend.
```

docs/source/user/installing.rst
```diff
@@ -185,18 +185,28 @@ vary with target operating system and hardware. The following lists them all
...
+    with the compiler toolchain mentioned above and accessible to the
+    compiler via standard flags (Linux, Windows)
+    * ``llvm-openmp`` (OSX) - provides headers for compiling OpenMP support into
     Numba's threading backend
-  * ``intel-openmp`` (OSX) - provides OpenMP library support for Numba's
+    * ``intel-openmp`` (OSX) - provides OpenMP library support for Numba's
     threading backend.
-  * ``tbb-devel`` - provides TBB headers/libraries for compiling TBB support
-    into Numba's threading backend
+  * ``NUMBA_NO_TBB=1 TBBROOT=/usr``
```
stuartarchibald, line 201:
```suggestion
  * ``NUMBA_NO_TBB``
```
would rather document these separately as they perform separate functions.

docs/source/user/installing.rst
```diff
@@ -185,18 +185,28 @@ vary with target operating system and hardware. The following lists them all
...
+    DEFAULT: both undefined
```
stuartarchibald, line 202:
```suggestion
```

docs/source/user/installing.rst
```diff
@@ -185,18 +185,28 @@ vary with target operating system and hardware. The following lists them all
...
+    set the environment variable ``NUMBA_NO_TBB`` to ``1`` when building using
+    ``setup.py`` to disable the compilation of the TBB threading backend.
+    If not setting it to ``1`` you can also add the location where the
+    TBB headers are going to be searched for, which is generally in ``/usr``,
+    as they are installed in ``/usr/include/tbb/tbb.h``.
```
stuartarchibald, line 207: This is very linux specific and also assumes it's a system package manager installed package. I think it might be appropriate to just mention that `TBBROOT` can be set if needed and to just link to the Intel documentation on it, https://software.intel.com/content/www/us/en/develop/documentation/advisor-user-guide/top/appendix/adding-parallelism-to-your-program/adding-the-parallel-framework-to-your-build-environment/defining-the-tbbroot-environment-variable.html ? What do you think?

docs/source/user/installing.rst
```diff
@@ -185,18 +185,28 @@ vary with target operating system and hardware. The following lists them all
...
-    toolchain mentioned above and accessible to the compiler via standard flags
-    (Linux, Windows).
 
-* Optional build time:
-
-  * ``llvm-openmp`` (OSX) - provides headers for compiling OpenMP support into
+* Build time toggles:
+  * ``NUMBA_NO_OPENMP=1``
+    DEFAULT: undefined
+    set the environment variable to ``1`` when building using ``setup.py``
```
stuartarchibald, line 192:
```suggestion
    Set this environment variable to ``1`` when building using ``setup.py``
```

docs/source/user/installing.rst
```diff
@@ -185,18 +185,28 @@ vary with target operating system and hardware. The following lists them all
...
+    * ``llvm-openmp`` (OSX) - provides headers for compiling OpenMP support into
     Numba's threading backend
-  * ``intel-openmp`` (OSX) - provides OpenMP library support for Numba's
+    * ``intel-openmp`` (OSX) - provides OpenMP library support for Numba's
     threading backend.
-  * ``tbb-devel`` - provides TBB headers/libraries for compiling TBB support
-    into Numba's threading backend
+  * ``NUMBA_NO_TBB=1 TBBROOT=/usr``
+    DEFAULT: both undefined
+    set the environment variable ``NUMBA_NO_TBB`` to ``1`` when building using
```
stuartarchibald, line 203:
```suggestion
    Set this environment variable to ``1`` when building using
```

docs/source/user/installing.rst
```diff
@@ -185,18 +185,28 @@ vary with target operating system and hardware. The following lists them all
...
+    ``setup.py`` to disable the compilation of the TBB threading backend.
```
stuartarchibald, line 204: I'm unsure as to what other ways there are to build other than with `setup.py` being invoked at some point?

docs/source/user/installing.rst
stuartarchibald: I'm unsure as to what other ways there are to build other than with setup.py being invoked at some point?

docs/source/user/installing.rst
epsilon-0: This is the location in gentoo, arch and ubuntu, afaik. So writing it out explicitly is a good thing, in addition to linking to the intel documentation.

Will add the intel link.

stuartarchibald: It may well be the location on a few linux distros, but what about windows and OSX users and people who aren't using linux installation defaults? I think the OS invariant link is more suitable.

epsilon-0: OK, I've added both the things for now, would that be acceptable?

docs/source/user/installing.rst
epsilon-0: removed this wording.

setup.py
esc: I would prefer to keep it as a non-empty string. This is the interface we have been using for `NUMBA_NO_OPENMP` and constraining it to be explicitly `1` would alter that interface, violate an existing contract we have and potentially break other peoples builds. So, in that light, I think it would be a good idea to remain consistent as to how we handle environment variables including the newly introduced `NUMBA_NO_TBB`.

docs/source/user/installing.rst
```diff
@@ -185,18 +185,28 @@ vary with target operating system and hardware. The following lists them all
...
-  * OpenMP C headers and runtime libraries compatible with the compiler
-    toolchain mentioned above and accessible to the compiler via standard flags
-    (Linux, Windows).
 
-* Optional build time:
-
-  * ``llvm-openmp`` (OSX) - provides headers for compiling OpenMP support into
+* Build time environment variables:
+  * ``NUMBA_NO_OPENMP``
+    Set this environment variable to ``1`` when building
```
esc, line 191: CI is complaining about unexpected indentation on this line. Would suggest to run `sphinx` locally to debug this:

```
Warning, treated as error:
/home/vsts/work/1/s/docs/source/user/installing.rst:191:Unexpected indentation.
make: *** [html] Error 2
```

stuartarchibald:
PR https://github.com/numba/numba/pull/6031 continues this with suggested fixes.

esc:
I'll close this for now, I have already reviewed #6031 . 🙏 thanks!

esc:
Frosty Thompson: https://twitter.com/esc___/status/1288565199763537921 🧟‍♂️🧟‍♂️🧟‍♂️

PR #5833, opened 2020-06-09
Fixes the source location appearing incorrectly in error messages.

stuartarchibald:
In the case of overloads, the error messages were using the source
from the target function opposed to the overload implementation.

This fixes it.

### DIFF
--- a/numba/core/types/functions.py
+++ b/numba/core/types/functions.py
@@ -145,6 +145,8 @@ class _ResolutionFailures(object):
             elif (is_external_fn_ptr and
                   isinstance(source_fn, numba.core.typing.templates.Signature)):
                 source_fn = template.__class__
+            elif hasattr(template, '_overload_func'):
+                source_fn = template._overload_func
             source_file, source_line = self._get_source_info(source_fn)
             largstr = argstr if err.literal else nolitargstr
             msgbuf.append(_termcolor.errmsg(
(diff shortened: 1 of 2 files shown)


numba/tests/test_errorhandling.py
```diff
@@ -265,6 +266,23 @@ def call_intrin():
...
+        def foo():
+            np.linalg.svd("chars")
+
+        with self.assertRaises(errors.TypingError) as raises:
+            foo()
+
+        excstr = str(raises.exception)
+        self.assertIn(error_reasons['specific_error'].splitlines()[0], excstr)
+        expected_file = os.path.join("numba", "np", "linalg.py")
+        self.assertIn(expected_file, excstr)
```
sklam, line 284: The test passes without the patch because the filename is in the traceback. Perhaps, check that `Overload in function 'svd_impl': ` in part of the line

numba/core/types/functions.py
```diff
@@ -145,6 +145,8 @@ def format(self):
             elif (is_external_fn_ptr and
                   isinstance(source_fn, numba.core.typing.templates.Signature)):
                 source_fn = template.__class__
+            elif hasattr(template, '_overload_func'):
+                source_fn = template._overload_func
```
sklam, line 149: perhaps we should add a method to `Template` to get the source function

stuartarchibald: I think that would be the ideal thing to do, but am reluctant to patch that in here with view of minimising changes between 0.50.0rc1 and 0.50.0.

PR #3640, opened 2019-01-02
PassThruContainer

asodeur:
This is an attempt at #3282. `PassThruContainer` is added to `numba.jitclass.passthru`. Usage looks like
```python
from numba import PassThruContainer

obj = dict(a=1)
container = PassThruContainer(obj)

@jit(nopython=True)
def identity(c):
    with objmode(a='int64'):
        c.obj['b'] = 2
        a = c.obj['a']

    return c, a

res, a = identity(container)

assert c is res
assert a == 1
assert res.obj['b'] == 2
```

The first version of  `PassThruContainer` was implemented as subclass of `jitclass._box.Box` but has since been rewritten to use `NRT_meminfo_new_from_pyobject` based on @sklam's comments below.

PR #5834, opened 2020-06-09
Fix the is operator on Ellipsis

eric-wieser:
This implements `a is b` on all opaque pointer types as `ptr_of(a) == ptr_of(b)`, which is exactly the meaning used by python anyway.

This implementation is copied from #3640.

### DIFF
--- a/numba/tests/test_operators.py
+++ b/numba/tests/test_operators.py
@@ -504,6 +504,19 @@ class TestOperators(TestCase):
         cfunc = cres.entry_point
         self.assertTrue(cfunc(Ellipsis, Ellipsis))
 
+    def test_is_void_ptr(self):
+        # can't call this directly from python, as void cannot be unboxed
+        cfunc_void = jit(
+            (types.voidptr, types.voidptr), nopython=True
+        )(self.op.is_usecase)
+
+        # this wrapper performs the casts from int to voidptr for us
+        @jit(nopython=True)
+        def cfunc(x, y):
+            return cfunc_void(x, y)
+
+        self.assertTrue(cfunc(1, 1))
+        self.assertFalse(cfunc(1, 2))
 
     #
     # Arithmetic operators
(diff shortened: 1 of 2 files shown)


eric-wieser:
@stuartarchibald, should be ready for review

stuartarchibald:
@eric-wieser thanks, have moved it into the queue.

eric-wieser:
@sklam: How can I write that test using the `compile_isolated` style used by the rest of this file?

stuartarchibald:
I don't think it's possible to use the `compile_isolated` style with something that cannot be unboxed.

eric-wieser:
@stuartarchibald, where would you like me to add the suggested test?

stuartarchibald:
> @stuartarchibald, where would you like me to add the suggested test?

Perhaps here: https://github.com/numba/numba/pull/5834/files#diff-fdfb3908ac2aefe7c395a313bf860a94R506 ?

eric-wieser:
Added the test

numba/tests/test_operators.py
```diff
@@ -491,6 +499,22 @@ def test_ne_scalar(self, flags=force_pyobj_flags):
...
         self.test_ne_scalar(flags=Noflags)
 
+    def test_is_ellipsis(self):
+        cres = compile_isolated(self.op.is_usecase, (types.ellipsis, types.ellipsis))
+        cfunc = cres.entry_point
+        self.assertTrue(cfunc(Ellipsis, Ellipsis))
+
+    def test_is_void_ptr(self):
+        # can't call this directly from python, as void cannot be unboxed
+        cfunc_void = njit((types.voidptr, types.voidptr))(self.op.is_usecase)
```
stuartarchibald, line 509: Think `njit` needs importing or this needs to be `@jit(nopyhon=True)` will appropriate line wrap.

eric-wieser: Any preference for which?

stuartarchibald: Whichever you like :)

PR #5838, opened 2020-06-09
Ensure `Dispatcher.__eq__` always returns a bool

eric-wieser:
Previously `typeof(some_jit_func) == "nonsense"` would return `None` instead of `False`

### DIFF
--- a/numba/tests/test_dispatcher.py
+++ b/numba/tests/test_dispatcher.py
@@ -153,6 +153,22 @@ skip_bad_access = unittest.skipUnless(_access_preventable, _access_msg)
 
 
 class TestDispatcher(BaseTest):
+  
+    def test_equality(self):
+        @jit
+        def foo(x):
+            return x
+        
+        @jit
+        def bar(x):
+            return x
+        
+        # Written this way to verify `==` returns a bool (gh-5838). Using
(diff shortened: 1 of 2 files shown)


eric-wieser:
Test added

numba/tests/test_dispatcher.py
```diff
@@ -153,6 +153,22 @@ def check_access_is_preventable():
...
+        @jit
+        def bar(x):
+            return x
+        
+        # Written this way to verify `==` returns a bool (gh-5838). Using
+        # `assertTrue(foo == foo)` or `assertEqual(foo, foo)` would defeat the
+        # purpose of this test.
+        self.assertEqual(foo == foo, True)
+        self.assertEqual(foo == bar, False)
+        self.assertEqual(foo == None, False)
```
eric-wieser, line 171:
```suggestion

    def test_equality(self):
        @jit
        def foo(x):
            return x

        @jit
        def bar(x):
            return x

        # Written this way to verify `==` returns a bool (gh-5838). Using
        # `assertTrue(foo == foo)` or `assertEqual(foo, foo)` would defeat the
        # purpose of this test.
        self.assertEqual(foo == foo, True)
        self.assertEqual(foo == bar, False)
        self.assertEqual(foo == None, False)  # noqa: E711
```

PR #5840, opened 2020-06-10
Typed Tuple

EPronovost:
This PR adds a decorator `@typed_tuple` that can be used to enforce type constraints on a `namedtuple` class.  Doing this allows us to statically define the numba type, so we do not need to escape to python during dispatch when such a tuple is passed as an argument to a jitted function.  This gives an order of magnitude speedup in such cases (see docstring of `typed_tuple` for examples).

This PR is not complete; I clearly need to add documentation and test coverage.  It also depends on the typing work from https://github.com/numba/numba/pull/5609.  The actual commit for typed tuples is [here](https://github.com/numba/numba/commit/386d833c12a705ba04d965affb837fbb006ba67c).  I wanted to get feedback on the idea, and also provide another example of how inferred numba specs from type hints could work.  If we decide to go forward with this, I would rebase and add tests and docs once https://github.com/numba/numba/pull/5609 merges.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -334,7 +334,6 @@ exclude =
     numba/core/typing/ctypes_utils.py
     numba/core/typing/enumdecl.py
     numba/core/typing/cffi_utils.py
-    numba/core/typing/typeof.py
     numba/core/typing/npdatetime.py
     numba/core/annotations/type_annotations.py
     numba/roc/mathdecl.py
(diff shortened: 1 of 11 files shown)


EPronovost:
@stuartarchibald I think this PR is "Ready for Review" now.

esc:
@EPronovost thanks for the ping, I have updated the label accordingly.

PR #5846, opened 2020-06-10
CUDA: Allow disabling NVVM optimizations, and fix debug issues

### LINKED ISSUE #5835: CUDA debug info is invalid - compile units have an empty list of subprograms
The `NvvmDIBuilder` has this method for constructing compile units:

```python
    def _di_compile_unit(self):
        filepair = self._filepair()
        empty = self.module.add_metadata([self._const_int(0)])
        return self.module.add_metadata([
            self._const_int(self.DI_Compile_unit),         # tag
            filepair,                   # source directory and file pair
            self._const_int(self.DW_LANG_Python),  # language
            'Numba',                     # producer
            self._const_bool(True),      # optimized
            "",                          # flags??
            self._const_int(0),          # runtime version
            empty,                       # enums types
            empty,                       # retained types
            self.module.add_metadata(self.subprograms),  # subprograms
            empty,                       # global variables
            empty,                       # imported entities
            "",                          # split debug filename
        ])
```

This is used to create a compile unit as soon as the `NvvmDIBuilder` is constructed. At this point, `self.subprograms` is empty, so the compile unit has an empty list of subprograms. If subprograms are added later, then the subprograms metadata referenced by the compile unit remains empty.

This can cause segfaults in NVVM, or in certain cases, a failed assertion:

```
Assertion # 1 failed!
Function:   beginFunction
Expression: TheCU && "Unable to find compile unit!"
```

The list of subprograms is used to build a reverse mapping from subprograms to compile units - because the list is empty, NVVM can't find the compile unit for a given subprogram at this point. Issues with debuginfo generation like this have also affected other users of LLVM 3.4 - e.g. https://github.com/ispc/ispc/issues/848.

I think the right thing to do is to modify the debug info generation so that the list of subprograms belonging to the compile unit can be updated each time a new subprogram is added, but I think this will take some work because the existing implementation is set up for everything to be immutable after creation.

### LINKED ISSUE #5836: CUDA: Passing opt=0 to NVVM doesn't work
`numba.cuda.cudadrv.nvvm.CompilationUnit` handles the `opt` (optimization) flag with:

```python
        if options.get('opt'):
            opts.append('-opt=%d' % options.pop('opt'))
```

at https://github.com/numba/numba/blob/master/numba/cuda/cudadrv/nvvm.py#L205-L206. This condition is never true, because `opt=0` makes `options.get('opt')` return `0`, which is False-ish.

At present Numba always passes `opt=3` so this is never hit, but it needs fixing first for adding a public-facing way of changing the optimization level.

gmarkall:
This PR adds an additional kwarg to the `@cuda.jit` decorator, `opt`, which can be used to control NVVM optimization. Previously there was no way to control this - the documented `NUMBA_OPT` environment variable was ignored by the CUDA target. I haven't added support for the environment variable as I think it's likely to be of little use in typical CUDA applications - disabling optimization can easily bring a kernel over the total registers per SM with launch configurations that would otherwise have been usable. It also allows the rest of the program to run much more quickly if optimization is only disabled for kernels that specifically need it.

Disabling optimization would always result in an NVVM segfault due to invalid debuginfo metadata, so the fixes for this issue is also contained within this PR.

See individual commit messages for more details of fixes / changes.

Fixes #5835
Fixes #5836

### DIFF
--- a/numba/cuda/tests/cudapy/test_optimization.py
+++ b/numba/cuda/tests/cudapy/test_optimization.py
@@ -1,0 +1,82 @@
+import numpy as np
+
+from numba.cuda.testing import skip_on_cudasim, CUDATestCase
+from numba import cuda, float64
+import unittest
+
+
+def kernel_func(x):
+    x[0] = 1
+
+
+def device_func(x, y, z):
+    return x * y + z
+
+
+# Fragments of code that are removed from kernel_func's PTX when optimization
+# is on
+removed_by_opt = ( '__local_depot0', 'call.uni', 'st.param.b64')
+
+
+@skip_on_cudasim('Simulator does not optimize code')
+class TestOptimization(CUDATestCase):
+    def test_eager_opt(self):
+        # Optimization should occur by default
+        kernel = cuda.jit((float64[::1],))(kernel_func)
+        ptx = kernel.inspect_asm()
+
+        for fragment in removed_by_opt:
+            with self.subTest(fragment=fragment):
+                self.assertNotIn(fragment, ptx)
+
+    def test_eager_noopt(self):
+        # Optimization disabled
+        kernel = cuda.jit((float64[::1],), opt=False)(kernel_func)
+        ptx = kernel.inspect_asm()
+
+        for fragment in removed_by_opt:
+            with self.subTest(fragment=fragment):
+                self.assertIn(fragment, ptx)
+
+    def test_lazy_opt(self):
+        # Optimization should occur by default
+        kernel = cuda.jit(opt=True)(kernel_func)
+        x = np.zeros(1, dtype=np.float64)
+        kernel[1, 1](x)
+
+        # Grab the PTX for the one definition that has just been jitted
+        ptx = next(iter(kernel.inspect_asm()))[1]
+
+        for fragment in removed_by_opt:
+            with self.subTest(fragment=fragment):
+                self.assertNotIn(fragment, ptx)
+
+    def test_lazy_noopt(self):
+        # Optimization disabled
+        kernel = cuda.jit(opt=False)(kernel_func)
+        x = np.zeros(1, dtype=np.float64)
+        kernel[1, 1](x)
+
+        # Grab the PTX for the one definition that has just been jitted
+        ptx = next(iter(kernel.inspect_asm().items()))[1]
+
+        for fragment in removed_by_opt:
+            with self.subTest(fragment=fragment):
+                self.assertIn(fragment, ptx)
+
+    def test_device_opt(self):
+        # Optimization should occur by default
+        device = cuda.jit(device=True)(device_func)
+        ptx = device.inspect_ptx((float64, float64, float64)).decode('utf-8')
+        self.assertIn('fma.rn.f64', ptx)
+
+    def test_device_noopt(self):
+        # Optimization disabled
+        device = cuda.jit(device=True, opt=False)(device_func)
+        ptx = device.inspect_ptx((float64, float64, float64)).decode('utf-8')
+        # Fused-multiply adds should be disabled when not optimizing
+        self.assertNotIn('fma.rn.f64', ptx)
+
+
+if __name__ == '__main__':
+    unittest.main()
--- a/numba/cuda/tests/cudapy/test_optimization.py
+++ b/numba/cuda/tests/cudapy/test_optimization.py
@@ -40,7 +40,7 @@ class TestOptimization(CUDATestCase):
 
     def test_lazy_opt(self):
         # Optimization should occur by default
-        kernel = cuda.jit(opt=True)(kernel_func)
+        kernel = cuda.jit(kernel_func)
         x = np.zeros(1, dtype=np.float64)
         kernel[1, 1](x)
 
--- a/numba/core/debuginfo.py
+++ b/numba/core/debuginfo.py
@@ -32,6 +32,12 @@ class AbstractDIBuilder(object):
         pass
 
     @abc.abstractmethod
+    def initialize(self):
+        """Initialize the debug info. An opportunity for the debuginfo to
+        prepare any necessary data structures.
+        """
+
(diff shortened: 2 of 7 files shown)


numba/cuda/tests/cudapy/test_optimization.py
```diff
@@ -0,0 +1,82 @@
...
+@skip_on_cudasim('Simulator does not optimize code')
+class TestOptimization(CUDATestCase):
+    def test_eager_opt(self):
+        # Optimization should occur by default
+        kernel = cuda.jit((float64[::1],))(kernel_func)
+        ptx = kernel.inspect_asm()
+
+        for fragment in removed_by_opt:
+            with self.subTest(fragment=fragment):
+                self.assertNotIn(fragment, ptx)
```
stuartarchibald, line 30: Do these need to be subtests, it's quite heavy to just check for 3 strings in some disassembly and were it to fail the assert output would contain the failing string?

numba/cuda/tests/cudapy/test_optimization.py
```diff
@@ -0,0 +1,82 @@
...
+        # Optimization disabled
+        kernel = cuda.jit((float64[::1],), opt=False)(kernel_func)
+        ptx = kernel.inspect_asm()
+
+        for fragment in removed_by_opt:
+            with self.subTest(fragment=fragment):
+                self.assertIn(fragment, ptx)
+
+    def test_lazy_opt(self):
+        # Optimization should occur by default
```
stuartarchibald, line 42: Isn't it being spelled out in the decorator on the next line?

numba/cuda/tests/cudapy/test_optimization.py
```diff
@@ -0,0 +1,82 @@
...
+
+        # Grab the PTX for the one definition that has just been jitted
+        ptx = next(iter(kernel.inspect_asm().items()))[1]
+
+        for fragment in removed_by_opt:
+            with self.subTest(fragment=fragment):
+                self.assertIn(fragment, ptx)
+
+    def test_device_opt(self):
+        # Optimization should occur by default
```
stuartarchibald, line 68: Isn't it being spelled out in the decorator on the next line?

numba/cuda/tests/cudapy/test_optimization.py
```diff
@@ -0,0 +1,82 @@
...
+        device = cuda.jit(device=True)(device_func)
+        ptx = device.inspect_ptx((float64, float64, float64)).decode('utf-8')
```
stuartarchibald, line 70: Not for here, but should the decode happen inside `inspect_ptx`?

numba/cuda/tests/cudapy/test_optimization.py
gmarkall: I'm not sure what you mean by "heavy"... I thought it would be quicker to resolve a future fail that occurs due to changed code generation if the test output spelled out exactly which fragments were and were not found, instead of stopping at the first instance of a fragment not being found.

numba/cuda/tests/cudapy/test_optimization.py
gmarkall: That would have been better imo, yes - changing it will break any code that currently expects it to be a `bytes` if/when we do change it though.

numba/cuda/tests/cudapy/test_optimization.py
gmarkall: Yes, I should not have the `opt` kwarg in there.

numba/cuda/tests/cudapy/test_optimization.py
gmarkall: I don't see that here. Am I missing something?

numba/cuda/tests/cudapy/test_optimization.py
gmarkall: I've now removed the `opt` kwarg on the next line.

gmarkall:
Thanks for the review - I've updated the PR, and made some other responses to the other comments.

numba/cuda/tests/cudapy/test_optimization.py
stuartarchibald: I just mean that they are adding extra things into the testing framework where 3 is the size of the search space, balancing weight against statistical likelihood of occurrence and inconvenience in event of occurrence. I appreciate your point and am happy to leave given you prefer this :)

numba/cuda/tests/cudapy/test_optimization.py
stuartarchibald: ok, is it worth filing a ticket for this?

numba/cuda/tests/cudapy/test_optimization.py
stuartarchibald: No :) I'm missing something! Probably read `True` and not to what it applies and assumed it was a copy paste from the above.

numba/cuda/tests/cudapy/test_optimization.py
stuartarchibald: thanks.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_64`.

gmarkall:
Did this pass the buildfarm run?

stuartarchibald:
> Did this pass the buildfarm run?

Undetermined, I'm pretty sure it would have passed if the windows Py3.6 machines hadn't picked up an unpatched-for-`atomicrmw`-upgrade llvmlite from somewhere. Everything else was fine.

esc:
BFID: `numba_smoketest_cuda_78`

gmarkall:
@esc conflicts are now resolved. Thanks for the buildfarm run and pointing this out!

esc:
BFID: `numba_smoketest_cuda_83`

esc:
BF passed again.

gmarkall:
Many thanks!

PR #5851, opened 2020-06-11
CUDA EMM enhancements - add default get_ipc_handle implementation, skip a test conditionally

gmarkall:
In some cases there need not be a specialized implementation of `get_ipc_handle` for an EMM plugin - for example, RMM no longer requires one: see https://github.com/rapidsai/rmm/pull/401#issuecomment-642506112.

This PR adds a mixin that provides Numba's implementation of `get_ipc_handle` for an EMM plugin to use. It has been tested with the Numba testsuite locally with both the default memory manager and using a modified version of RMM that uses the mixin.

One test fails with an EMM plugin (`TestContextAPI.test_context_memory`) if the plugin does not implement `get_memory_info`, so a small fix for this - skipping the test as appropriate - is rolled into this PR.

### DIFF
--- a/docs/source/cuda/external-memory.rst
+++ b/docs/source/cuda/external-memory.rst
@@ -166,6 +166,19 @@ follows:
    :member-order: bysource
 
 
+The IPC Handle Mixin
+--------------------
+
+An implementation of the ``get_ipc_handle()`` function is is provided in the
+``GetIpcHandleMixin`` class. This uses the driver API to determine the base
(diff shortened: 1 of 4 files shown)


jakirkham:
@shwina, would you be able to take a look at this? 🙂

shwina:
Thanks @gmarkall and @jakirkham - looks good to me :)

esc:
BFID: `numba_smoketest_cuda_77`

esc:
BF passed.

PR #5856, opened 2020-06-11
Add support for conversion of inplace_binop to parfor.

DrTodd13:
If we have a inplace_binop with both sides being numpy arrays then we can convert that to a parfor and enable more fusion.

### DIFF
--- a/numba/tests/test_parfors.py
+++ b/numba/tests/test_parfors.py
@@ -569,7 +569,7 @@ class TestParfors(TestParforsBase):
     def test_logistic_regression(self):
         args = (numba.float64[:], numba.float64[:,:], numba.float64[:],
                 numba.int64)
-        self.assertTrue(countParfors(lr_impl, args) == 1)
+        self.assertTrue(countParfors(lr_impl, args) == 2)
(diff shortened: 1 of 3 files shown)


esc:
Question: does this address: https://github.com/numba/numba/issues/5942 ?

esc:
Or maybe this one: https://github.com/numba/numba/issues/5923 ?

DrTodd13:
@esc No, it doesn't address either of those.

numba/tests/test_parfors.py
```diff
@@ -1622,6 +1622,18 @@ def test_impl(a):
...
         self.check(test_impl, x)
 
+    @skip_parfors_unsupported
+    def test_inplace_binop(self):
+        def test_impl(a, b):
+            b += a
+            return b
+
+        X = np.ones(10)
+        Y = np.zeros(10)
```
sklam, line 1632: The test cannot detect if the transformed code is producing the correct result because the addition is adding zeros. 

perhaps change to arrays with non-zeros
```suggestion
        X = np.arange(10) + 10
        Y = np.arange(10) + 100
```

PR #5857, opened 2020-06-12
CUDA docs: Add notes on resetting the EMM plugin

gmarkall:
There are some legitimate uses for re-setting the EMM plugin. Examples:

* In the Numba testsuite: https://github.com/numba/numba/blob/master/numba/cuda/tests/cudadrv/test_emm_plugins.py#L107-L115
* Dask-cuda tests: https://github.com/rapidsai/dask-cuda/issues/313 / https://github.com/rapidsai/dask-cuda/pull/315

This PR adds some notes explaining how to re-set the memory manager, and explains the side-effects / pitfalls of doing so.

### DIFF
--- a/docs/source/cuda/external-memory.rst
+++ b/docs/source/cuda/external-memory.rst
@@ -278,7 +278,7 @@ any contexts, as EMM Plugin instances are instantiated along with contexts.
 .. autofunction:: numba.cuda.set_memory_manager
 
 
-Re-setting the memory manager
+Resetting the memory manager
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 It is recommended that the memory manager is set once prior to using any CUDA
@@ -305,4 +305,3 @@ to set the memory manager multiple times, noting the following:
           ``@cuda.jit`` prior to context destruction will need to be
           re-defined, as the code underlying them will also have been unloaded
           from the GPU.
-
--- a/docs/source/cuda/external-memory.rst
+++ b/docs/source/cuda/external-memory.rst
@@ -272,10 +272,37 @@ Function
 --------
 
 The :func:`~numba.cuda.set_memory_manager` function can be used to set the
-memory manager at runtime. This must be called prior to the initialization of
-any contexts, and EMM Plugin instances are instantiated along with contexts.
-
-It is recommended that the memory manager is set once prior to using any CUDA
-functionality, and left unchanged for the remainder of execution.
+memory manager at runtime. This should be called prior to the initialization of
+any contexts, as EMM Plugin instances are instantiated along with contexts.
 
 .. autofunction:: numba.cuda.set_memory_manager
+
+
+Re-setting the memory manager
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+It is recommended that the memory manager is set once prior to using any CUDA
+functionality, and left unchanged for the remainder of execution. It is possible
+to set the memory manager multiple times, noting the following:
+
+* At the time of their creation, contexts are bound to an instance of a memory
+  manager for their lifetime.
+* Changing the memory manager will have no effect on existing contexts - only
+  contexts created after the memory manager was updated will use instances of
+  the new memory manager.
+* :func:`numba.cuda.close` can be used to destroy contexts after setting the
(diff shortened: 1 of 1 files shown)


docs/source/cuda/external-memory.rst
```diff
@@ -272,10 +272,37 @@ Function
...
 The :func:`~numba.cuda.set_memory_manager` function can be used to set the
-memory manager at runtime. This must be called prior to the initialization of
-any contexts, and EMM Plugin instances are instantiated along with contexts.
+memory manager at runtime. This should be called prior to the initialization of
+any contexts, as EMM Plugin instances are instantiated along with contexts.
+
+.. autofunction:: numba.cuda.set_memory_manager
+
+
+Re-setting the memory manager
```
stuartarchibald, line 281:
```suggestion
Resetting the memory manager
```

docs/source/cuda/external-memory.rst
```diff
@@ -272,10 +272,37 @@ Function
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
stuartarchibald, line 282:
```suggestion
~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
(assuming the above is accepted)

docs/source/cuda/external-memory.rst
```diff
@@ -272,10 +272,37 @@ Function
...
+  - Attempting to use invalid arrays, streams, or events will likely fail with
+    an exception being raised due to a ``CUDA_ERROR_INVALID_CONTEXT`` or
+    ``CUDA_ERROR_CONTEXT_IS_DESTROYED`` return code from a Driver API function.
+  - Attempting to use an invalid module will result in similar, or in some
+    cases a segmentation fault / access violation.
+
+.. note:: The invalidation of modules means that all functions compiled with
+          ``@cuda.jit`` prior to context destruction will need to be
+          re-defined, as the code underlying them will also have been unloaded
+          from the GPU.
```
stuartarchibald, line 307: Practically, what does "redefined" imply?

docs/source/cuda/external-memory.rst
```diff
@@ -272,10 +272,37 @@ Function
...
+    context.
+  - Attempting to use invalid arrays, streams, or events will likely fail with
+    an exception being raised due to a ``CUDA_ERROR_INVALID_CONTEXT`` or
+    ``CUDA_ERROR_CONTEXT_IS_DESTROYED`` return code from a Driver API function.
+  - Attempting to use an invalid module will result in similar, or in some
+    cases a segmentation fault / access violation.
+
+.. note:: The invalidation of modules means that all functions compiled with
+          ``@cuda.jit`` prior to context destruction will need to be
+          re-defined, as the code underlying them will also have been unloaded
```
stuartarchibald, line 306:
```suggestion
          redefined, as the code underlying them will also have been unloaded
```

docs/source/cuda/external-memory.rst
gmarkall: I'm finding it quite hard to define both correctly and succinctly - everything's a bit fiddly because you're not really supposed to do this. I was reluctant to try and enumerate exactly what would need to be done in different scenarios, because I think it will be hard to get right, and here I'd opted for "recommend not doing this, and leave the user to figure out what to do for their particular scenario if they want to go against the recommendation". To try and answer your question though, I can give one example of a redefinition when it's necessary:

```python
from numba import cuda, void, float32

def f(x):
    x[0] = 1

f_1 = cuda.jit(void(float32))(f) # f_1 is defined here and valid
cuda.close()
# f_1 now invalid - attempting to launch results in a segfault,
# CUDA_ERROR_INVALID_CONTEXT, or CUDA_ERROR_CONTEXT_IS_DESTROYED
# unpredictably

# Redefine f_1:
f_1 = cuda.jit(void(float32))(f) # f_1 is defined here and valid

# Now if f_1 is called, a context will be created if necessary, and it will be jitted in the current context.
```
```

stuartarchibald: Ah I see, thanks. I agree, I think leaving it loosely defined is probably the safer thing to do as it's likely scenario dependent and will require thought over the design/necessity. The text provides sufficient hint to get started.

PR #5859, opened 2020-06-12
CUDA: Fix reduce docs and style improvements

gmarkall:
The reduce docs were a little inaccurante - there are no circumstances under which the reduce kernel overwrites the input array (since the implementation was rewritten in 151753b2e346dfadfcc84de46eb25de25b8044e7), and the name of a parameter was incorrect. This PR fixes these issues, and makes some stylistic chances to improve readability.


gmarkall:
Thanks - typo fixed.

PR #5861, opened 2020-06-12
Added except for possible Windows get_terminal_size exception

peppingdore:
I'm trying to import numba in the embedded Python environment with piped stdout. And get_terminal_size() raises ValueError.

### DIFF
--- a/numba/core/types/functions.py
+++ b/numba/core/types/functions.py
@@ -4,7 +4,8 @@ import inspect
 import itertools
 import logging
 import textwrap
-from os import path, get_terminal_size
+from os import path
+from shutil import get_terminal_size
 
 from .abstract import Callable, DTypeSpec, Dummy, Literal, Type, weakref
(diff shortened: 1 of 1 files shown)


stuartarchibald:
Thanks for the patch. What does the `ValueError` error message contain?

stuartarchibald:
Am thinking that `get_terminal_size` from `shutil` might be the better way to go as it has a fallback in the case of the `os` level call failing.

peppingdore:
ValueError contains: "bad file descriptor". Yeah, shutil.get_terminal_size() seems to be better solution.

stuartarchibald:
> ValueError contains: "bad file descriptor". Yeah, shutil.get_terminal_size() seems to be better solution.

Thanks, I'd have thought that'd go into an `OSError` but obviously not! Would you be ok to patch in and verify a change to `shutil` please? Thanks.

PR #5862, opened 2020-06-12
Do not leak loop iteration variables into the numba.np.npyimpl namespace

eric-wieser:
Previously this exposed `numba.np.npyimpl.ufunc` as `np.bitwise_xor`, which is just bizarre.

Split from #5619 for ease of review. This should inherit the milestone from that PR, probably.


stuartarchibald:
> Previously this exposed `numba.np.npyimpl.ufunc` as `np.bitwise_xor`, which is just bizarre.

Agree!

PR #5866, opened 2020-06-13
[WIP] Implement str and repr builtins

### LINKED ISSUE #5536: Implement `str` correctly
The code base at present implements `str()` directly for each type via an overload, these are here:
```
numba/cpython/charseq.py:600:@overload(str)
numba/cpython/unicode.py:436:@overload(str)
```
it would be better if `str()` was implemented as per cpython https://docs.python.org/3/library/stdtypes.html#str whereby there's a single overload for `str()` which calls the argument's `.__str__` and if that's missing calls `repr(argument)`, `repr` also needs implementing in a similar way!

AndrewEckart:
### Changes
- Overload `str()` and `repr()` to call underlying `__str__` and `__repr__ dunder` methods
- Replace current `str()` overloads for `UnicodeCharSeq` and `UnicodeType` with equivalent overloads of `__str__`
- Add `__repr__` overloads for `UnicodeCharSeq` and `UnicodeType`
- Add `__str__` and `__repr__` overloads for `CharSeq` and `Bytes` types
- Add tests for new overloaded builtins
- Removed unused definition from `test_builtins.py`
- Minor correction to comments
This allows for some new functionality, like:
```
>>> f = njit(lambda x: str(x))
>>> f(b'foo')
b'foo
```
This can also extend to other Numba types if `__str__` implementations are defined for them (e.g. `f(3.14159)` only fails because `__str__` is not yet defined on `types.float64`).

### Known issues
- The `str` overload should fallback to `__repr__` when `__str__` is not defined. Seems like this might require lowering since `hasattr` and `getattr` are not implemented.
- Having trouble coming up with a test case for `UnicodeCharSeq`
- The new builtins actually work on a `jitclass` which defines the appropriate method:
```
>>> @jitclass([])
... class A_str:
...     def __init__(self):
...         pass
... 
...     def __str__(self):
...         return "A.__str__"
>>> a = A()
>>> f(a)
A.__str__
```
~~but test cases will fail because the vanilla Python `str(a)` cannot resolve to `.__str__` on the jitted class.~~ Fixed in #5877 per @EPronovost 
- `njit` fails with `TypeError: bad argument type for built-in operation` for the `CharSeq.__str__` overload. Quick repro:
```
def test(val):
    print(str(val))

x = np.array([b'Zoot!'])[0]
njit(test)(x)
```
Seems like this might be a bug in `jit`, but I'm not sure.

### Not implemented
- `str` with multiple arguments (to decode bytes-like objects). This could be done by modifying the `_to_str` overloads for `CharSeq` and `Bytes` to accept an encoding argument.

Should fix #5536

### DIFF
--- a/numba/cpython/builtins.py
+++ b/numba/cpython/builtins.py
@@ -595,3 +595,18 @@ def ol_filter(func, iterable):
                 if func(x):
                     yield x
     return impl
+
+
+# ------------------------------------------------------------------------------
+# str, repr
+
+@overload(str)
+def ol_str(s):
+    def impl(s):
+        return s.__str__()
+    return impl
+
+
+@overload(repr)
+def ol_repr(s):
+    return lambda s: s.__repr__()
--- a/numba/cpython/charseq.py
+++ b/numba/cpython/charseq.py
@@ -10,10 +10,10 @@ from numba.core.cgutils import is_nonelike
 from numba.cpython import unicode
 
 # bytes and str arrays items are of type CharSeq and UnicodeCharSeq,
-# respectively.  See numpy/types/npytypes.py for CharSeq,
+# respectively.  See core/types/npytypes.py for CharSeq,
 # UnicodeCharSeq definitions.  The corresponding data models are
-# defined in numpy/datamodel/models.py. Boxing/unboxing of item types
-# are defined in numpy/targets/boxing.py, see box_unicodecharseq,
+# defined in core/datamodel/models.py. Boxing/unboxing of item types
+# are defined in core/boxing.py, see box_unicodecharseq,
 # unbox_unicodecharseq, box_charseq, unbox_charseq.
 
 s1_dtype = np.dtype('S1')
@@ -597,21 +597,29 @@ def charseq_to_str_mth(s):
     return tostr_impl
 
 
-@overload(str)
-def charseq_str(s):
-    if isinstance(s, types.UnicodeCharSeq):
-        get_code = _get_code_impl(s)
+@overload_method(types.CharSeq, '__str__')
+@overload_method(types.Bytes, '__str__')
+@overload_method(types.CharSeq, '__repr__')
+@overload_method(types.Bytes, '__repr__')
+def bytes_str(s):
+    def str_impl(s):
+        return "b'" + s._to_str() + "'"
+    return str_impl
 
-        def str_impl(s):
-            n = len(s)
-            kind = s._get_kind()
-            is_ascii = kind == 1 and s.isascii()
-            result = unicode._empty_string(kind, n, is_ascii)
-            for i in range(n):
-                code = get_code(s, i)
-                unicode._set_code_point(result, i, code)
-            return result
-        return str_impl
+
+@overload_method(types.UnicodeCharSeq, '__str__')
+@overload_method(types.UnicodeCharSeq, '__repr__')
+def charseq_str(s):
+    def str_impl(s):
+        n = len(s)
+        kind = s._get_kind()
+        is_ascii = kind == 1 and s.isascii()
+        result = unicode._empty_string(kind, n, is_ascii)
+        for i in range(n):
+            code = unicode_charseq_get_code(s, i)
+            unicode._set_code_point(result, i, code)
+        return result
+    return str_impl
 
 
 @overload(bytes)
--- a/numba/cpython/unicode.py
+++ b/numba/cpython/unicode.py
@@ -433,10 +433,27 @@ def _codepoint_is_ascii(ch):
 # PUBLIC API
 
 
-@overload(str)
+@overload_method(types.UnicodeType, '__str__')
 def unicode_str(s):
-    if isinstance(s, types.UnicodeType):
-        return lambda s: s
+    def impl(s):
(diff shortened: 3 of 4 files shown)


EPronovost:
This is related to this PR https://github.com/numba/numba/pull/5877.  I extended jitclass boxing to support built-in methods from Python (which should fix the issue you mention above for `str(a)`).

stuartarchibald:
@AndrewEckart thanks for creating a draft for this. The way to do:

> The str overload should fallback to __repr__ when __str__ is not defined. Seems like this might require lowering since hasattr and getattr are not implemented.

may involve a pattern like (untested and probably needs some messing about with the signatures/args to get the references to the object right):
```python
from numba import njit, types
from numba.core import errors
from numba.extending import overload, intrinsic

@intrinsic
def dispatcher_thing(tyctx, typ):
    call_attr = tyctx.resolve_getattr(typ, '__str__') # try and resolve a `__str__` method on the type
    if call_attr is None: # it failed so try again but this time try and resolve `__repr__` 
        call_attr = tyctx.resolve_getattr(typ, '__repr__')
    if call_attr is None: # it failed again, there's no __str__ or __repr__ for this type so abort
        msg = "%s has no __str__ or __repr__ implemented" % typ
        raise errors.TypingError(msg)
    sig = call_attr.get_call_type(tyctx, (), {})
    def codegen(cgctx, builder, sig, args):
        impl = cgctx.get_function(call_attr, sig)
        return impl(builder, *args)
    return sig, codegen

@overload(str)
def ol_str(s):
    def impl(s):
        return dispatcher_thing(s)
    return impl

@njit
def call_str():
    print(str(<whatever>))
```

AndrewEckart:
> @AndrewEckart thanks for creating a draft for this. The way to do:
> 
> > The str overload should fallback to **repr** when **str** is not defined. Seems like this might require lowering since hasattr and getattr are not implemented.
> 
> may involve a pattern like (untested and probably needs some messing about with the signatures/args to get the references to the object right):
> 
> ```python
> from numba import njit, types
> from numba.core import errors
> from numba.extending import overload, intrinsic
> 
> @intrinsic
> def dispatcher_thing(tyctx, typ):
>     call_attr = tyctx.resolve_getattr(typ, '__str__') # try and resolve a `__str__` method on the type
>     if call_attr is None: # it failed so try again but this time try and resolve `__repr__` 
>         call_attr = tyctx.resolve_getattr(typ, '__repr__')
>     if call_attr is None: # it failed again, there's no __str__ or __repr__ for this type so abort
>         msg = "%s has no __str__ or __repr__ implemented" % typ
>         raise errors.TypingError(msg)
>     sig = call_attr.get_call_type(tyctx, (), {})
>     def codegen(cgctx, builder, sig, args):
>         impl = cgctx.get_function(call_attr, sig)
>         return impl(builder, *args)
>     return sig, codegen
> 
> @overload(str)
> def ol_str(s):
>     def impl(s):
>         return dispatcher_thing(s)
>     return impl
> 
> @njit
> def call_str():
>     print(str(<whatever>))
> ```

I gave it a shot, but I'm mystified. Everything through `sig = call_attr.get_call_type(tyctx, (), {})` seems to work just fine, but compilation is failing with 
```
tuple index out of range
```
I got stuck debugging this vague message.

VladKochetov007 (approved):

stuartarchibald:
> @AndrewEckart thanks for creating a draft for this. The way to do:
> 
> > The str overload should fallback to **repr** when **str** is not defined. Seems like this might require lowering since hasattr and getattr are not implemented.
> 
> may involve a pattern like (untested and probably needs some messing about with the signatures/args to get the references to the object right):
> 
> ```python
> from numba import njit, types
> from numba.core import errors
> from numba.extending import overload, intrinsic
> 
> @intrinsic
> def dispatcher_thing(tyctx, typ):
>     call_attr = tyctx.resolve_getattr(typ, '__str__') # try and resolve a `__str__` method on the type
>     if call_attr is None: # it failed so try again but this time try and resolve `__repr__` 
>         call_attr = tyctx.resolve_getattr(typ, '__repr__')
>     if call_attr is None: # it failed again, there's no __str__ or __repr__ for this type so abort
>         msg = "%s has no __str__ or __repr__ implemented" % typ
>         raise errors.TypingError(msg)
>     sig = call_attr.get_call_type(tyctx, (), {})
>     def codegen(cgctx, builder, sig, args):
>         impl = cgctx.get_function(call_attr, sig)
>         return impl(builder, *args)
>     return sig, codegen
> 
> @overload(str)
> def ol_str(s):
>     def impl(s):
>         return dispatcher_thing(s)
>     return impl
> 
> @njit
> def call_str():
>     print(str(<whatever>))
> ```

Update... the `getattr` builtin turned out to be quite involved, it's finally implemented in #7884. The `hasattr` builtin is implemented in the pending #8442, `str` and `repr` were written at the same time as this, as well as updates to the correctness of `hash`.

stuartarchibald:
#8442 has been merged, which means `str` and `repr` now work as expected by Python i.e. via `__str__` and `__repr__`. Thanks for starting the work on this @AndrewEckart, am going to close this PR as effectively completed!

Issue #5690, opened 2020-05-09
Support __len__ in jitclass


tony-fav:
## Feature request

jitclass currently seems to not see or do anything with `__len__` method. I am not sure about other magic methods, but improving this aspect would be appreciated. Example:

```
from numba import jitclass


@jitclass([])
class LenTest(object):
    def __init__(self):
        pass

    def __len__(self):
        return 1


test = LenTest()
print(len(test))
# print(test.__len__())
```
returns
`TypeError: object of type LenTest has no len()`

GuillaumeLeclerc:
This would be great. I cam here for the same reason

stuartarchibald:
Thanks for the request. Think this is related to https://github.com/numba/numba/issues/4002.

Issue #1606, opened 2016-01-07
Jitclass: support for __call__ etc


oyamad:
Despite https://github.com/numba/numba/pull/1571#issuecomment-165205412, would it be possible for `jitclass` to support special methods such as
1. `__call__`
2. `__getitem__`, `__setitem__`
3. `__repr__`, `__str__`
4. `__iter__`

and others?

`__call__` would be especially useful since it will allow the instance to look like a jitted function that can be passed to another jitted function as an argument. See also https://groups.google.com/a/continuum.io/forum/#!topic/numba-users/YE12tE2dIwY.

datnamer:
Also multiple inheritance would be great.

moble:
@datnamer Multiple inheritance seems to work, at least from my simple first test.  The only wrinkle is that I calling the super `__init__` functions is a bit tricky, even for single inheritance.  Is there a different problem that simple tests don't turn up?

moble:
I would also add the operators `__add__`, `__radd__`, `__mul__`, etc.

datnamer:
@moble ok thanks.

I was also thinking of ways for easy code reuse, how about abstract base classes and or user defined protocols/magic methods? (similar to traits).

j-towns:
I want to re-iterate that overloading of math operators like `__add__`, `__mul__` etc is essential for [Autograd](https://github.com/HIPS/autograd) to be compatible with Numba (which would be awesome).

How difficult would this be to implement? I'm happy to help if I can.

AriKatz:
I think a JIT class, which is heap allocated and mutable would be too slow.

AriKatz:
Does that mean JIT class support will stay as is? If not, what are the long term plans?

sklam:
I have plans to expand the JIT class support.  Few months ago, I was working on various jitclass features but I was slowed down by many tricky details.  Some of these details, like the operating overloading, can be implemented but with great cost in code complexity.  I think it is partially due to the technical debts in the internals of numba.  So, I am taking a pause in pushing new features for jitclasses and trying to clean things up a bit.

moble:
Is this closed by #1851?

moble:
I guess not, as my tests all fail.  #1865 implies that special cases are needed, and that was closed in favor of (the still-open) #2187.

aldanor:
Wondering if there's any plans on reviving this?

There's a few tricky magic methods, but would be very nice to at least have the 'easy' ones like `__str__` / `__repr__`.

[...]

Issue #4926, opened 2019-12-04
Numba heapq doesn't work with numba typed lists


ma-sadeghi:
- [x] I am using the latest released version of Numba
- [x] I have included below a minimal working reproducer

# Description
`heapq` doesn't seem to like anything other than lists, even `numba` typed lists don't work.

# How to reproduce
```
import heapq as hq
from numba.typed import List

l = List()
l.append(5)
l.append(3)

e = hq.heappop(l)
```
```
TypeError: heap argument must be a list
```

stuartarchibald:
Thanks for the report. The issue is that CPython requires a CPython list:
https://github.com/python/cpython/blob/664fe3996f7e05ae351526f02b21504bb065bcf8/Modules/_heapqmodule.c#L139-L142 pretty sure there's nothing that can easily be done about this from Numba.

Numba has it's own [`heapq` implementation](https://github.com/numba/numba/blob/7b2b0a1966b1d503e709a84da8aa64e72f35c9dc/numba/targets/heapq.py) for use in `jit` decorated functions, we should fix that to accept `typed.List`.

ma-sadeghi:
@stuartarchibald Thank you for your quick reply. How can I use Numba's `heapq` implementation?

stuartarchibald:
@ma-sadeghi there's a short example here: https://nbviewer.jupyter.org/github/numba/numba-examples/blob/master/notebooks/Numba_043_release_demo.ipynb#Support-for-heapq- essentially `import heapq` then use it in a jit decorated function, it behaves pretty much the same as CPython `heapq` (IIRC the implementations are very similar).

gmarkall:
Bumped into this attempting to help @truenicoco with an implementation of a priority queue on Gitter. With the following PQ implementation:

```python
import itertools

import numba as nb
from numba.experimental import jitclass
from typing import List, Tuple, Dict
from heapq import heappush, heappop


# @jitclass
class PurePythonPriorityQueue:
    def __init__(self):
        self.pq = []  # list of entries arranged in a heap
        self.entry_finder = {}  # mapping of indices to entries
        self.REMOVED = -1  # placeholder for a removed item
        self.counter = itertools.count()  # unique sequence count

    def put(self, item: Tuple[int, int], priority: float = 0.0):
        """Add a new item or update the priority of an existing item"""
        if item in self.entry_finder:
            self.remove_item(item)
        count = next(self.counter)
        entry = [priority, count, item]
        self.entry_finder[item] = entry
[...]

PR #4715, opened 2019-10-16
Add `bool` overload for several types

EPronovost:
Now `bool("Hello, world")` works.

I would like to do the same thing for `UnicodeCharSeq` so that `np.nonzero(np.array(["Hi", ""]))` works.  However, this didn't "Just Work".  Going to look into it further.

PR #5877, opened 2020-06-17
Jitclass builtin methods

EPronovost:
Add support for a limited set of python object builtin methods.
We can now do things like
```
@jitclass({"x": intp})
class MyInt:
    def __init__(self, x):
        self.x = x

    def __add__(self, other):
        return MyInt(self.x + other.x)

    ...
```
There are several built-in methods (e.g. `__repr__`, `__call__`) that need additional work and are not covered by this PR.

Fixes https://github.com/numba/numba/issues/1606.
Fixes https://github.com/numba/numba/issues/5690.

(*note*: description edited to link the above issues so they will close upon merge of this PR)


EPronovost:
The documentation is failing because of `register_simple_overload` (commenting out these fixes it).  I'm not sure how to work around this.  My guess is that it doesn't like the templated functions.

numba/experimental/jitclass/overloads.py
```diff
@@ -0,0 +1,169 @@
...
+    if not isinstance(x, ClassInstanceType):
+        return
+
+    if "__bool__" in x.jit_methods:
+        return lambda x: x.__bool__()
+
+    if "__len__" in x.jit_methods:
+        return lambda x: x.__len__() != 0
+
+    return lambda x: True
```
eric-wieser, line 63: Is there some way to provide this as a default implementation of `bool` for all objects?

numba/experimental/jitclass/overloads.py
```diff
@@ -0,0 +1,169 @@
...
+def _register_overload(func, template, glbls=None):
+    """
+    Register overload handler for func calling class attribute attr.
+    """
+    namespace = dict(ClassInstanceType=ClassInstanceType)
+    if glbls:
+        namespace.update(glbls)
+
+    exec(template, namespace)
+    handler = namespace["handler"]
```
eric-wieser, line 43: Is this necessary, vs the simpler:
```suggestion
    dunder = f"__{attr}__"
    def handler(self, *args):
        if not isinstance(self, ClassInstanceType):
            return
        if dunder in jit.methods:
            return lambda self, *args: getattr(self, dunder)(*args)
```

EPronovost: Unfortunately `getattr` is not currently supported, so that would not work.

numba/experimental/jitclass/overloads.py
EPronovost: I believe that
```
@overload(bool)
def fallback_bool(x):
    return lambda x: True
```
would not be a good idea (if I'm mistaken then cool).  We can add overloads on a type by type basis (e.g. https://github.com/numba/numba/pull/4715)

rpopovici:
Hey @EPronovost, is there a reason why this PR wasn't merged already?

gmarkall:
> Hey @EPronovost, is there a reason why this PR wasn't merged already?

Presently there are test fails in this PR, but the tests ran so long ago it looks like the logs are no longer available. According to @EPronovost's earlier comment it may just be because of docs: https://github.com/numba/numba/pull/5877#issuecomment-650497134

Once the tests are passing, this PR can be queued for review.

EPronovost:
> The documentation is failing because of register_simple_overload (commenting out these fixes it). I'm not sure how to work around this. My guess is that it doesn't like the templated functions. cc: @sklam

@rpopovici as stated above the scripts to generate numba documentation were failing to parse [these lines](https://github.com/numba/numba/pull/5877/files#diff-a659338b412f21c8bb81e8fcbd8c58d78be23c8c80f4346ccff94d2f61e9ff1fR132-R169).  I was hoping a numba core developer would offer a suggestion for ways to resolve this.

rpopovici:
@EPronovost thanks for answering. I was under the impression that this PR is abandonware..
I don't know what numba uses for documentation: sphinx, pdoc? but from the looks of it register_simple_overload has no doc string. Adding one might fix the problem
```
def register_simple_overload(func, *attrs, nargs=1):
"""This is a docstring."""
```
Second, I saw that "eq" and "ne" operators are commented. Is there a reason why they are not supported? or is just a code review mistake

EPronovost:
@rpopovici thanks for the suggestion; docstrings are not the issue here.  Given that someone else is interested in this I looked into it some more and found a workaround.

EPronovost:
Also @stuartarchibald or @sklam I think this PR is ready for review.

rpopovici:
@EPronovost thanks for looking into it. There is one more issue you might one to address. The commented eq and ne comparison operators should be accompanied by an explanation why they are commented in the first place. I am assuming that it has something to do with the missing __hash__ overload..

EPronovost:
@rpopovici Yes previously we had issues trying to overwrite the `hash` method, but I found a way around it.

rpopovici:
Brilliant work @EPronovost! Now there is only one big feature missing from jitclasses. Caching..

EPronovost:
@sklam @stuartarchibald could this PR be marked ready for review?

rpopovici:
@gmarkall @sklam can anyone review this PR? It would be nice to have it in 0.53

shaunc:
I'd bet there aren't more people jumping up and down for this because what comes up first in google are other issues which are dead ends. ... But... 🥳 ... I'm certainly looking forward to it. :)

cosmoni:
reverse methods don't seem to be working with this branch. those should be included before being merged.

gmarkall:
Moving this into 0.55 - it would be nice to give it a push, as it enables the use of lists of jitclasses with `heapq`, as mentioned in this comment: https://github.com/numba/numba/issues/4926#issuecomment-889779403

(cc @truenicoco).

gmarkall:
(Also, the merge conflict is trivial to fix - it's just some additional imports on both sides in `boxing.py` - if you get a minute, @EPronovost , could you merge + resolve conflicts please?)

LunarLanding:
It probably would need its own issue, but would it be possible to have builtin methods work on StructRef's?

esc:
> > I believe the PR is currently stalled
> 
> Is there anything I can do to help? I would really like to see this reach the 0.55 release.

This is an open-source project. You are welcome to do whatever you like (mostly). If you would like to see this in `master` I would suggest trying to contact the original author (they have been away/radio-silent for quite some time) to see what their plans are and also (at the same time) "adopt" the PR. This means, create a new PR with all the commits that have been made so far and then drive it to completion by responding to the reviewer comments until it is merged.

gmarkall:
I'm going to try and adopt this PR - I've just pushed a merge of master to it with resolved conflicts, and will have a look over the reviews shortly. (@EPronovost please let me know if you have any concerns about this)

gmarkall:
As far as I can tell from above, the only outstanding comment was that this PR doesn't presently work for reverse methods. So the path forward here consists of:

- Fix the now-failing tests that result from the merge of master,
- Determine whether support for reverse needs adding, or if it should be omitted from this PR and potentially documented as a limitation,
- A review of the code.

gmarkall:
Oddly, the tests seem to be passing locally for me but all failing in CI.

gmarkall (dismissed):
On the whole I thought this looked good but I've pushed some small changes:

- Fixing up the type error raised in complex typing to be a new style error.
- Some tidy-ups to the test suite.
- Addition of eq / ne testing for the float and int wrappers - these were present in the list of operations to test but were commented out.
- Remove `ipos` and `ineg` from the list of supported dunders, as I don't think these exist.
- List the supported dunders in the docs, to save users having to look at the Numba source code to see the list.

Although I think this is good, it may be worth someone with a little more jitclass experience checking it, e.g. @sklam / @stuartarchibald ?

gmarkall:
> Determine whether support for reverse needs adding, or if it should be omitted from this PR and potentially documented as a limitation,

In the interests of getting this merged, I decided to leave this out of this PR and list the supported dunders - I think it needn't also be explicitly pointed out that this won't be supported.

gmarkall:
Can we try and get this into 0.56? I am aware of active uses of this branch that are waiting on it being merged into Numba.

gmarkall:
Following OOB discussion, @stuartarchibald or @sklam may be able to review this.

numba/tests/test_jitclasses.py
```diff
@@ -1112,5 +1114,593 @@ def __init__(self):
...
+        class IndexClass:
+            def __init__(self):
+                pass
+
+            def __index__(self):
+                return 1
+
+        obj = IndexClass()
+
+        if sys.version[:3] >= "3.8":
```
sklam, line 1416: use `numba.core.utils.PYVERSION` instead

numba/experimental/jitclass/overloads.py
```diff
@@ -0,0 +1,205 @@
...
+        options.append(lambda x: float(x.__index__()))
+
+    return take_first(*options)
+
+
+@class_instance_overload(int)
+def class_int(x):
+    options = [try_call_method(x, "__int__")]
+
+    if (sys.version_info.major, sys.version_info.minor) >= (3, 8):
```
sklam, line 135: use `numba.core.utils.PYVERSION` instead

numba/experimental/jitclass/overloads.py
```diff
@@ -0,0 +1,205 @@
...
+
+register_simple_overload(abs, "abs")
+register_simple_overload(len, "len")
+
+# Comparison operators.
+register_simple_overload(hash, "hash")
+register_simple_overload(operator.ge, "ge", n_args=2)
+register_simple_overload(operator.gt, "gt", n_args=2)
+register_simple_overload(operator.le, "le", n_args=2)
+register_simple_overload(operator.lt, "lt", n_args=2)
```
sklam, line 171: no reflection implemented 

see <https://docs.python.org/3/reference/datamodel.html#object.__ge__>

sklam (changes requested):
Review summary:
- needs to be more testing on the `__bool__` fallback 
- rich-comparison operator lack reflection support. If decided to not support, the limitation should be noted in the docs.

Additional comment:

Too bad the lack of serialization support for jitclass type means we cannot return a Numba typed container containing a jitclass instance.

numba/experimental/jitclass/overloads.py
```diff
@@ -0,0 +1,205 @@
...
+    for o in options:
+        if o is not None:
+            return o
+
+
+@class_instance_overload(bool)
+def class_bool(x):
+    return take_first(
+        try_call_method(x, "__bool__"),
+        try_call_method(x, "__len__"),
```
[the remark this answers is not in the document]
gmarkall: This should now be fixed.

gmarkall:
@sklam Summary of changes since your initial review:

- Minor fixups fixed up
- Bool fallback fixed and tested
- Operator reflection implemented and tests
- I also noticed that an implicit `__hash__` member set when `__eq__` is defined but `__hash__` is not caused a problem because it's a class member, which is unsupported. So I also added this case to the ignored attrs, and added a specific test for the issue (this is the last commit on the branch)

I think this should be ready for another look now - many thanks in advance!

sklam (approved):
Thanks for the fixes.

PR #5879, opened 2020-06-17
Fix erroneous input mutation in linalg routines

### LINKED ISSUE #3368: lstsq sometimes overwrites its inputs
Hi,

we've noticed that `np.linalg.lstsq` overwrites its input matrix in some edge cases. I'd expect the input matrix to never change.

To reproduce:

```python
import numpy as np
import numba as nb

X = np.array([[1., 7.54, 6.52],
              [1., 2.70, 4.00],
              [1., 2.50, 3.80],
              [1., 1.15, 5.64],
              [1., 4.22, 3.27],
              [1., 1.41, 5.70],
            ], order='F')

X_orig = X.copy()

y = np.array([1., 2., 3., 4., 5., 6.])

@nb.jit(nopython=True, nogil=True)
def f1(X, y):
    return np.linalg.lstsq(X, y)

f1(X, y)
(X == X_orig).all()
# True, all fine until here

@nb.jit(nopython=True, nogil=True)
def f2(X, y, test):
    if test:
        # never executed, but necessary to trigger the bug
        X = X[1:2, :]
    return np.linalg.lstsq(X, y)

f2(X, y, False)
(X == X_orig).all()
# False, X got corrupted
```

From the values in X it looks like it's likely being used as the work matrix of `lstsq`:
```python
array([[-2.44948974, 14.24766238,  0.53161088],
       [ 0.        ,  4.23875445,  2.46241227],
       [ 0.        ,  0.30113438,  3.52520967],
       [ 0.28989795,  0.47206594, -0.33747568],
       [ 0.28989795,  0.03894287,  0.52109546],
       [ 0.28989795,  0.43538451, -0.33387113]])
```

To trigger the issue it seems to be necessary for `X` to be in fortran order, and then be modified inside the jitted function.

I'm not entirely sure what causes this exactly, but looking at https://github.com/numba/numba/blob/master/numba/targets/linalg.py#L1619 it seems like it might be confused what the order of the X matrix is? And then skips doing the copy?

Tested with numba 0.39 as that's the latest available in anaconda, apologies if this is already fixed in 0.40.

### LINKED ISSUE #5870: np.linalg.pinv Unintended mutation of input
## Reporting a bug

Hello all, long time!

The following script (apologies - rather contrived) is an example of a function which does not error when compiled but manages to mutate one of its inputs (not intentionally).

It you call the compiled function with the same inputs many times, you find the mutated input switches from one set of values to another; the 'other' values are very close to the original input values.

Per the note, simply copying the input makes all the assertions pass, as you'd expect.

Just wasn't sure why the input gets modified / why the copy is needed - this might be more of a question than an issue, but it took a while to track down so thought it might be worth posting in case other users find it of interest.

Cheers

Rob

```python
import numpy as np
import pandas as pd
from numba import njit


@njit
def some_fn(alp, v):
    rows, cols = alp.shape
    p = np.empty_like(alp)

    for i in range(rows):
        A_ = v[i]
        b_ = alp[i]

        res = np.linalg.pinv(A_) @ b_

        p[i] = res[:cols]

    return p


if __name__ == '__main__':

    v_data = pd.DataFrame(
        [
            [0.55, -0.21, 0, 0.07],
            [-0.21, 0.48, 0, 0.10],
            [0, 0, 0.31, -0.15],
            [0.07, 0.10, -0.15, 0.51]
        ]
    )

    v_data_original = v_data.copy()

    a_data = pd.DataFrame([[0.1, 0.3, -0.1, 0.05]])

    v = v_data.to_numpy().reshape((1, 4, 4))  # <- using this, the cfunc assertions all fail

    # v = v_data.to_numpy(copy=True).reshape((1, 4, 4))  # <- using this, the cfunc assertions all pass

    a = a_data.values

    # -------- pyfunc

    some_fn.py_func(a, v)
    pd.testing.assert_frame_equal(v_data_original, v_data)  # pass

    some_fn.py_func(a, v)
    pd.testing.assert_frame_equal(v_data_original, v_data)  # pass

    # -------- cfunc

    some_fn(a, v)
    pd.testing.assert_frame_equal(v_data_original, v_data)  # fails (case 1)

    some_fn(a, v)
    pd.testing.assert_frame_equal(v_data_original, v_data)  # fails (case 2)

    some_fn(a, v)
    pd.testing.assert_frame_equal(v_data_original, v_data)  # fails (case 1)

    some_fn(a, v)
    pd.testing.assert_frame_equal(v_data_original, v_data)  # fails (case 2)
```

* case 1 values are quite different to the original input
* case 2 values are quite similar to the original input

stuartarchibald:
This fixes an issue where inputs were not correctly being copied on the way into linear algebra routines that destroy data during computation. The issue stemmed from `np.asfortranarray` being called on `'A'` ordered arrays, and, in the case that they were already `'F'` ordered at run time, no copy would be made. The following implements a new helper function to ensure that `'A'` ordered arrays have an appropriate copy made at runtime, fixes up the implementations that were incorrect to now use this, and adds tests for functions that would hit these code paths.

Fixes #5870 
Fixes #3368 
Fixes https://github.com/numba/numba/issues/5933

### DIFF
--- a/numba/np/linalg.py
+++ b/numba/np/linalg.py
@@ -841,8 +841,6 @@ def inv_impl(a):
 
     kind = ord(get_blas_kind(a.dtype, "inv"))
 
-    F_layout = a.layout == 'F'
-
     def inv_impl(a):
         n = a.shape[-1]
         if a.shape[-2] != n:
@@ -851,10 +849,7 @@ def inv_impl(a):
 
         _check_finite_matrix(a)
 
-        if F_layout:
-            acpy = np.copy(a)
-        else:
-            acpy = np.asfortranarray(a)
+        acpy = _copy_to_fortran_order(a)
 
         if n == 0:
             return acpy
--- a/numba/np/linalg.py
+++ b/numba/np/linalg.py
@@ -1280,8 +1280,6 @@ def svd_impl(a, full_matrices=1):
 
     _check_linalg_matrix(a, "svd")
 
-    F_layout = a.layout == 'F'
-
     # convert typing floats to numpy floats for use in the impl
     s_type = getattr(a.dtype, "underlying_float", a.dtype)
     s_dtype = np_support.as_dtype(s_type)
@@ -1302,10 +1300,7 @@ def svd_impl(a, full_matrices=1):
 
         _check_finite_matrix(a)
 
-        if F_layout:
-            acpy = np.copy(a)
-        else:
-            acpy = np.asfortranarray(a)
+        acpy = _copy_to_fortran_order(a)
 
         ldu = m
         minmn = min(m, n)
--- a/numba/np/linalg.py
+++ b/numba/np/linalg.py
@@ -2057,8 +2057,6 @@ def _compute_singular_values_impl(a):
     u = np.empty((1, 1), dtype=np_dtype)
     vt = np.empty((1, 1), dtype=np_dtype)
 
-    F_layout = a.layout == 'F'
-
     def sv_function(a):
         """
         Computes singular values.
@@ -2080,10 +2078,7 @@ def _compute_singular_values_impl(a):
         ucol = 1
         ldvt = 1
 
-        if F_layout:
-            acpy = np.copy(a)
-        else:
-            acpy = np.asfortranarray(a)
+        acpy = _copy_to_fortran_order(a)
 
         # u and vt are not referenced however need to be
         # allocated (as done above) for MKL as it
--- a/numba/np/linalg.py
+++ b/numba/np/linalg.py
@@ -1966,8 +1966,6 @@ def slogdet_impl(a):
 
     kind = ord(get_blas_kind(a.dtype, "slogdet"))
 
-    F_layout = a.layout == 'F'
-
     diag_walker = _get_slogdet_diag_walker(a)
 
     ONE = a.dtype(1)
@@ -1984,10 +1982,7 @@ def slogdet_impl(a):
 
         _check_finite_matrix(a)
 
-        if F_layout:
-            acpy = np.copy(a)
-        else:
-            acpy = np.asfortranarray(a)
+        acpy = _copy_to_fortran_order(a)
 
         ipiv = np.empty(n, dtype=F_INT_nptype)
 
--- a/numba/np/linalg.py
+++ b/numba/np/linalg.py
@@ -800,8 +800,9 @@ def ol_copy_to_fortran_order(a):
             # decide based on runtime value
             flag_f = a.flags.f_contiguous
             if flag_f:
-                # it's already F ordered, so copy
-                acpy = np.copy(a)
+                # it's already F ordered, so copy but in a round about way to
+                # ensure that the copy is also F ordered
+                acpy = np.copy(a.T).T
             else:
                 # it's something else ordered, so let asfortranarray deal with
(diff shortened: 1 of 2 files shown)


sklam:
Can you add testing to `_copy_to_fortran_order`?

for example:
```diff
diff --git a/numba/tests/test_linalg.py b/numba/tests/test_linalg.py
index 235ff8ad9..61cc768e3 100644
--- a/numba/tests/test_linalg.py
+++ b/numba/tests/test_linalg.py
@@ -8,7 +8,7 @@ import platform
 
 import numpy as np
 
-from numba import jit
+from numba import jit, njit, typeof
 from numba.core import errors
 from numba.tests.support import (TestCase, tag, needs_lapack, needs_blas,
                                  _is_armv7l, skip_ppc64le_issue4026)
@@ -1577,8 +1577,8 @@ class TestLinalgLstsq(TestLinalgSystems):
         empties = [
         [(0, 1), (1,)], # empty A, valid b
         [(1, 0), (1,)], # empty A, valid b
-        [(1, 1), (0,)], # valid A, empty 1D b 
-        [(1, 1), (1, 0)],  # valid A, empty 2D b 
+        [(1, 1), (0,)], # valid A, empty 1D b
+        [(1, 1), (1, 0)],  # valid A, empty 2D b
         ]
 
         for A, b in empties:
@@ -2646,5 +2646,46 @@ class TestBasics(TestLinalgSystems):  # TestLinalgSystems for 1d test
         self.assert_error(cfunc, args, msg, err=errors.TypingError)
 
 
+class TestHelpers(TestCase):
+    def test_copy_to_fortran_order(self):
+        from numba.np.linalg import _copy_to_fortran_order
+
+        def check(udt, expectfn, shapes, dtypes, orders):
+            for shape, dtype, order in product(shapes, dtypes, orders):
+                a = np.arange(np.prod(shape)).reshape(shape, order=order)
+
+                r = udt(a)
+                # check correct operation
+                self.assertPreciseEqual(expectfn(a), r)
+                # check new copy has made
+                self.assertNotEqual(a.ctypes.data, r.ctypes.data)
+
+        @njit
+        def direct_call(a):
+            return _copy_to_fortran_order(a)
+
+        shapes = [(3, 4), (3, 2, 5)]
+        dtypes = [np.intp]
+        orders = ['C', 'F']
+        check(direct_call, np.asfortranarray, shapes, dtypes, orders)
+
+
+        @njit
+        def slice_to_any(a):
+            # make a 'any' layout slice
+            sliced = a[::2][0]
+            return _copy_to_fortran_order(sliced)
+
+        shapes = [(3, 3, 4), (3, 3, 2, 5)]
+        dtypes = [np.intp]
+        orders = ['C', 'F']
+
+        def expected_slice_to_any(a):
+            # make a 'any' layout slice
+            sliced = a[::2][0]
+            return np.asfortranarray(sliced)
+
+        check(slice_to_any, expected_slice_to_any, shapes, dtypes, orders)
+
 if __name__ == '__main__':
     unittest.main()

```

sklam:
Everything else looks good.

stuartarchibald:
Thanks for the review, patch suggested here https://github.com/numba/numba/pull/5879#issuecomment-653629164 is applied as https://github.com/numba/numba/pull/5879/commits/afa20e0af8ed9c3b2ca652a6b1fc6c0108b7e36d.

Issue #5885, opened 2020-06-18
Add support for overloading `__call__` on custom types with `extending.overload`


eric-wieser:
## Feature request

It would be great if I could write:
```pass
@numba.extending.overload('call')
def my_type_call(self):
    if isinstance(self, MyNumbaExtensionType):
        return lambda self: 42
```

This is likely a prerequisite for #1606

eric-wieser:
My naive attempt to do this was:
```python
class MultiVectorType(types.Callable):
    ...

    @property
    def _call_overload(self):
        return next(v for k, v in builtin_registry.globals if k == MultiVector.__call__)

    # needed to overload call
    def get_call_type(self, context, args, kws):
        return self._call_overload.get_call_type(context, args, kws)

    def get_call_signatures(self):
        return self._call_overload.get_call_signatures()

@numba.extending.overload(MultiVector.__call__)
def call(self):
    def impl(self):
        return 42
    return impl
```

Unfortunately, this fails at type inference with `some_multi_vector()`, and fails to find an implementation to lower with `some_multi_vector(1)`.

eric-wieser:
I was able to hack this together by:

* Using `@numba.extending.overload_method` as above
* Changing `MultiVectorType` to subclass `types.Function`, and having it deliberately bypass `Function.__init__`
* Adding to my extension class the method:
  ```python
      def get_impl_key(self, sig):
          return (type(self), '__call__')
  ```

ZzEeKkAa:
Any plans to add out of box support for this?

I've just faced same problem, however `_WithCallOverload` hack does not work. I'm getting error `AttributeError: 'MultiVectorType' object has no attribute 'typing_key'`. It appears that there was some logic added to Numba if type is instance of `Function` that is relying on the Function initialization that we skipped.

ZzEeKkAa:
Motivate by #5886 I've came up with a workaround. In addition to 
```
@numba.extending.overload_method(MultiVectorType, '__call__')
...
```

you need to add proper lowering function, so Numba knows how to lower it:

```python
# __call__ overload is not supported by numba, however adding this overload
# kind of activates proper behaviour and works like magic.
# This is a reimplementation of _OverloadMethodTemplate.method_impl
# Issue reference: https://github.com/numba/numba/issues/5885
# TODO: remove once supported
@lower_builtin(MultiVectorType, MultiVectorType, types.VarArg(types.Any))
def method_impl(context, builder, sig, args):
    typing_context = context.typing_context
[...]

PR #5886, opened 2020-06-18
Add support for overloading the __call__ operator

eric-wieser:
With this in place, `@numba.extending.overload_method(ExtensionType, '__call__')` appears to work.

Possibly a fix for #5885.

### DIFF
--- a/numba/core/base.py
+++ b/numba/core/base.py
@@ -563,15 +563,6 @@ class BaseContext(object):
                 # Raise exception for the type instance, for a better error message
                 pass
 
-            # It's a type instance => try to find a definition for the type class
-            # TODO: this is weird, why are both static and non-static calls in
-            # the same place.
-            try:
-                return self.get_function(type(fn), sig)
-            except NotImplementedError:
-                # Raise exception for the type instance, for a better error message
-                pass
-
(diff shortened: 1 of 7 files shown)


numba/core/base.py
```diff
@@ -556,7 +556,16 @@ def get_function(self, fn, sig, _firstcall=True):
...
+            except NotImplementedError:
+                # Raise exception for the type instance, for a better error message
+                pass
+
             # It's a type instance => try to find a definition for the type class
+            # TODO: this is weird, why are both static and non-static calls in
+            # the same place.
             try:
                 return self.get_function(type(fn), sig)
             except NotImplementedError:
```
eric-wieser, line 571: I don't really understand where this code path is hit. If I remove it, then I get errors like `No definition for lowering class(uint8)(Literal[int](1),) -> uint8`

eric-wieser:
CI failures are because `_ResolutionFailures.format` assumes that `Template.key` is a function, but with this change it is sometimes a tuple:

https://github.com/numba/numba/blob/92df8df57dd31a3da51d3b0cf94a4ec2b591f71a/numba/core/types/functions.py#L142-L149

Any suggestion for what `source_fn` should be for an overloaded call operator?

PR #5892, opened 2020-06-19
Better handling of tuple arguments whose fields are modified.

### LINKED ISSUE #5872: cannot write to numpy array in namedtuple with parallel=True
## Reporting a bug

- [X ] I am using the latest released version of Numba (most recent is visible in
- [X ] I have included below a minimal working reproducer (if you are unsure how

With `parallel=True` it does not seem as if anything is wirtten to a numpy array in a namedtuple.

Running this code:

```
import numpy as np

sol_tuple = namedtuple(f'sol',['N','x','y'])
N = 10
sol = sol_tuple(N=N,x=np.zeros(N),y=np.zeros(N))

@njit
def test(sol):
    sol.y[:] = 5
    for i in prange(sol.N):
        if i < 5:
            sol.x[i] = i**2
        else:
            sol.x[i] = i

test(sol)
print(sol.x)
print(sol.y)
```

Results in both arrays being zero:

```
[0. 0. 0. 0. 0. 0. 0. 0. 0. 0.]
[0. 0. 0. 0. 0. 0. 0. 0. 0. 0.]
```

When removing `parallel=True`  the code correctly results in:

```
[ 0.  1.  4.  9. 16.  5.  6.  7.  8.  9.]
[5. 5. 5. 5. 5. 5. 5. 5. 5. 5.]
```

DrTodd13:
Resolves #5872.  Don't treat tuple as an immutable type since fields inside of it can be mutable.  This will keep alive a modification a field inside of a tuple to propagate outside the function.  Added ability in testing code to give a custom compare function to be able to compare how the argument passed in to the various versions of testing function was changed by that function.  Also, with this change, we need separate copies of args so that we can see those effects and so we need to deepcopy tuples since fields inside those can change.


numba/tests/test_parfors.py
```diff
@@ -133,26 +135,34 @@ def copy_args(*args):
...
+        parfor_args = copy_args(*args)
+        parfor_output = cpfunc.entry_point(*parfor_args)
 
-        self.assertEqual(type(njit_output), type(parfor_output))
+        if check_args_for_equality is None:
+            np.testing.assert_almost_equal(njit_output, py_expected, **kwargs)
+            np.testing.assert_almost_equal(parfor_output, py_expected, **kwargs)
+            self.assertEqual(type(njit_output), type(parfor_output))
+        else:
+            for pyarg, njitarg, parforarg, argcomp in zip(
```
stuartarchibald, line 162: I think it'd be good to put a check in here to make sure that `check_args_for_equality` is the same size as e.g. `pyarg` so that the iteration doesn't stop short because insufficient checking functions were supplied. Might also be a good idea to have a "single checking function" option?

DrTodd13: @stuartarchibald I intentionally did it this way.  That way, if you have multiple arguments but only need to check the first one then you only need to give one comparison function.  When writing tests, you can then put the ones you need to compare at the beginning but it isn't a disaster if you don't because you can provide a comparator that always returns True.  So, we need to come to agreement on the right approach.  If mine is good then I think the PR is fine, maybe some more documentation about the feature?  If yours is selected then add that check and documentation.  Let me know what you think.

stuartarchibald: I see, thanks for explaining. As tests are so complicated I'm inclined to go with explicit size match with check and document that as such opposed to relying on test authors recalling/reading the docs and reviewers catching systematic errors. Perhaps a null comparator can be added to the module so it is trivial to just add as needed? Thoughts?

stuartarchibald:
Thanks for the fixes @DrTodd13, unfortunately it seems like CI is now failing in quite a few places, one problem looks like an unbound local, not sure about the other.

numba/tests/test_parfors.py
DrTodd13: @stuartarchibald I implemented your suggestion but in the process of pushing the commit it rejected it due to rejecting fast-forwards.  So, I did a git pull and now I'm seeing some testing error about the number of parfors in some test.  This code was passing CI before the merge?  Do you think there's some interaction causing this or did some error creep in somehow in the merge?

numba/tests/test_parfors.py
```diff
@@ -133,26 +157,35 @@ def copy_args(*args):
...
+        parfor_args = copy_args(*args)
+        parfor_output = cpfunc.entry_point(*parfor_args)
 
-        self.assertEqual(type(njit_output), type(parfor_output))
+        if check_args_for_equality is None:
+            np.testing.assert_almost_equal(njit_output, py_expected, **kwargs)
+            np.testing.assert_almost_equal(parfor_output, py_expected, **kwargs)
+            self.assertEqual(type(njit_output), type(parfor_output))
+        else:
+            assert(len(pyarg) == len(check_args_for_equality))
```
stuartarchibald, line 184:
```suggestion
            assert(len(py_args) == len(check_args_for_equality))
```
this should fix the unbound local problem

numba/tests/test_parfors.py
stuartarchibald: This PR will be merged against current master as part of the CI test strategy, I guess there's some interaction in that that is causing the problem.
```
$ git rev-parse HEAD # THIS IS THE HEAD OF THIS BRANCH
d9b60690ecf7c3aa078f13dd2cf1b31ce9649655

$ ./runtests.py numba.tests.test_parfors.TestParfors.test_logistic_regression
.
Ran 1 test in 0.675s

OK

$ git merge master
Auto-merging numba/tests/test_parfors.py
Auto-merging numba/core/ir_utils.py
Merge made by the 'recursive' strategy.
 docs/source/conf.py                       |   2 +-
 docs/source/reference/envvars.rst         |   4 ++
 docs/source/user/installing.rst           |  56 +++++++++++++----
 numba/_helperlib.c                        | 266 +++++++++++++++++++++++++++++++++++++-----------------------------------------
 numba/core/codegen.py                     |   2 +-
 numba/core/errors.py                      |  12 ++--
 numba/core/inline_closurecall.py          |   2 +
 numba/core/ir_utils.py                    |   4 +-
 numba/core/object_mode_passes.py          |   2 +-
 numba/core/typeconv/typeconv.py           |   2 +-
 numba/core/typed_passes.py                |   2 +-
 numba/core/typeinfer.py                   |   2 +-
 numba/core/types/containers.py            |   5 +-
 numba/core/types/functions.py             |  21 ++++++-
 numba/core/typing/arraydecl.py            |   2 +
 numba/core/typing/npydecl.py              |   4 ++
 numba/core/typing/templates.py            |  18 ++++++
 numba/cpython/tupleobj.py                 |  18 ++++++
 numba/cuda/envvars.py                     |   2 +-
 numba/experimental/jitclass/decorators.py |   2 +-
 numba/misc/gdb_hook.py                    |   2 +-
 numba/misc/literal.py                     |   3 +
 numba/np/ufunc/parallel.py                |  49 +++++++++------
 numba/parfors/array_analysis.py           |   2 +
 numba/parfors/parfor.py                   |  61 +++++++++++-------
 numba/tests/test_comprehension.py         |  10 +++
 numba/tests/test_deprecations.py          |   4 +-
 numba/tests/test_parallel_backend.py      |  58 +++++++++++------
 numba/tests/test_parfors.py               |  40 ++++++++++++
 numba/tests/test_tuples.py                |  13 +++-
 setup.py                                  |  77 ++++++++++++-----------
 31 files changed, 474 insertions(+), 273 deletions(-)

$ ./runtests.py numba.tests.test_parfors.TestParfors.test_logistic_regression
F
FAIL: test_logistic_regression (numba.tests.test_parfors.TestParfors)
  File "<path>/numba/numba/tests/test_parfors.py", line 605, in test_logistic_regression
    self.assertTrue(countParfors(lr_impl, args) == 2)

Ran 1 test in 0.457s

```

DrTodd13:
> Thanks for the fixes @DrTodd13, unfortunately it seems like CI is now failing in quite a few places, one problem looks like an unbound local, not sure about the other.

Not sure what is going on.  After your fix, I can run the entire parfor test suite successfully but that parfor count issue still appears in the CI testing.  Thoughts @stuartarchibald ?

stuartarchibald:
> > Thanks for the fixes @DrTodd13, unfortunately it seems like CI is now failing in quite a few places, one problem looks like an unbound local, not sure about the other.
> 
> Not sure what is going on. After your fix, I can run the entire parfor test suite successfully but that parfor count issue still appears in the CI testing. Thoughts @stuartarchibald ?

It's likely this: https://github.com/numba/numba/pull/5892#discussion_r465346497 ? This branch works fine by itself, but when current master is merged into it, as per CI runs, it breaks. So I think there's something on current master that has a side effect when merged with this branch, the side effect being test failures.

DrTodd13:
> > > Thanks for the fixes @DrTodd13, unfortunately it seems like CI is now failing in quite a few places, one problem looks like an unbound local, not sure about the other.
> > 
> > 
> > Not sure what is going on. After your fix, I can run the entire parfor test suite successfully but that parfor count issue still appears in the CI testing. Thoughts @stuartarchibald ?
> 
> It's likely this: [#5892 (comment)](https://github.com/numba/numba/pull/5892#discussion_r465346497) ? This branch works fine by itself, but when current master is merged into it, as per CI runs, it breaks. So I think there's something on current master that has a side effect when merged with this branch, the side effect being test failures.

So we should merge in master and find the issue?  Best way to do that merge?

stuartarchibald:
> > > > Thanks for the fixes @DrTodd13, unfortunately it seems like CI is now failing in quite a few places, one problem looks like an unbound local, not sure about the other.
> > > 
> > > 
> > > Not sure what is going on. After your fix, I can run the entire parfor test suite successfully but that parfor count issue still appears in the CI testing. Thoughts @stuartarchibald ?
> > 
> > 
> > It's likely this: [#5892 (comment)](https://github.com/numba/numba/pull/5892#discussion_r465346497) ? This branch works fine by itself, but when current master is merged into it, as per CI runs, it breaks. So I think there's something on current master that has a side effect when merged with this branch, the side effect being test failures.
> 
> So we should merge in master and find the issue? Best way to do that merge?

Yes, and I think just fetching current master and merging it in will work, it's what I did here https://github.com/numba/numba/pull/5892#discussion_r465346497

DrTodd13:
@stuartarchibald Ultimately, I think we'll eventually have to make these aliasing decisions based on more than the type.  For now though, I got both logistic_regression and namedtuple3 both working by saying that a tuple of ints we don't need to worry about aliasing but a tuple where some part is mutable then we do.  Please review.

JeppeDruedahl:
This look great. Is there any schedule for when this will be included in numba? (just reproduced the problem in 0.51.2)

JeppeDruedahl:
This looks great. Is there any schedule for when this will be included in numba? (just reproduced the problem in 0.51.2)

seibert:
Once this is reviewed and merged, it will appear in dev builds of numba (available with `conda install -c numba numba`) within a day or so.  As 0.51.2 just came out, I suspect the next non-patch release of Numba will be a couple months away.

stuartarchibald:
Closing this as https://github.com/numba/numba/pull/6318 continues it with a cleaner git history.

Issue #4362, opened 2019-07-24
[FEA] Enable CUDA UVM (Unified/Universal) Memory Support from Numba


awthomp:
## Feature request
NVIDIA's embedded GPU line (TX2, Xavier, Nano, to name a few) feature a shared memory space between CPU and GPU. Typically handled in CUDA with unified memory, data access between host and device involves a zero-copy.

On these embedded platforms (I'm using the TX2), I'm seeing different pointer addresses when analyzing the pointer from \_\_array_interface\_\_ and \_\_cuda_array_interface\_\_.

```
>>> import numpy as np
>>> import cupy as cp
>>> from numba import cuda
>>> 
>>> arr = np.arange(100)
>>>
>>> arr.__array_interface__['data']
(94169420899168, False)
>>> 
>>> grr = cuda.to_device(arr)
>>> 
>>> grr.__cuda_array_interface__['data']
(140351520112640, False)
```

`numba.cuda.to_device()` doesn't seem to have awareness that the memory allocated via `np.arange()`, in this case, is also accessible to the GPU. 

Instead, we see a performance overhead due to a physical memory copy and different pointer address.

## Test Setup
NVIDIA Jetson TX2
Numba 0.45 via [conda4aarch64](https://github.com/jjhelmus/conda4aarch64/releases)

seibert:
Seems like we could look at the device properties to determine whether a copy is needed.  This will also affect the CUDA Python kernels, as Numba automatically calls to_device() when host arrays are passed to the GPU for processing.

seibert:
It would be nice if the property we use to detect this situation also worked when the future support for automatic paging of host memory to the device appears in Linux x86_64.

seibert:
Also, do you know if zero-copy works on Linux POWER8/9 when the GPU is connected to the CPU over NVLink?  There may be multiple platforms affected by this, now that I think about it.

awthomp:
I don't *think* POWER CPU/GPU is a shared memory space via NVLINK, but don't quote me on that one. What you do get is massively increased BW over a PCIe Gen3 data movement. Each NVLINK lane is 50GB/s bi-directional.

Happy to see some great engagement on this feature request, @seibert!

stuartarchibald:
CC @andih, you may be interested.

seibert:
Did some hunting, and I think I was remembering this talk from GTC:

[...]

PR #5900, opened 2020-06-21
Enable CUDA Unified Memory

maxpkatz:
CUDA [Unified Memory](https://devblogs.nvidia.com/unified-memory-cuda-beginners/), also known as managed memory, is added. Managed memory can be accessed on both the CPU and the GPU. For sufficiently recent GPUs (and when running on Linux), managed memory operates in a paged manner, where individual pages in an allocation can migrate back and forth from CPU to GPU. For older GPUs (NVIDIA Kepler and Maxwell generation; or, when running on Windows) managed memory allocations can migrate to either the CPU or the GPU but cannot be simultaneously accessed by both. This addresses the enhancement request in #4362 (that request specifically was to enable functionality on the NVIDIA Jetson platform; this allows [duplicate memory allocations to be avoided](https://docs.nvidia.com/cuda/cuda-for-tegra-appnote/index.html), without having to resort to workarounds like pinned memory).

At the driver level, the CUDA driver API call [`cuMemAllocManaged`](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MEM.html#group__CUDA__MEM_1gb347ded34dc326af404aa02af5388a32) is exposed through `memallocmanaged`. (For those more familiar with the CUDA runtime API, this corresponds to [`cudaMallocManaged`](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1gd228014f19cc0975ebe3e0dd2af6dd1b).) 

At the higher CUDA level, `managed_array` is exposed. If desired, this can be used as an argument to a device kernel; an example is shown in `numba/cuda/tests/cudadrv/test_managed_alloc.py`:

```
        ary = cuda.managed_array(100, dtype=np.double)

        @cuda.jit('void(double[:])')
        def kernel(x):
            i = cuda.grid(1)
            if i < x.shape[0]:
                x[i] = 1.0

        kernel[10, 10](ary)
```

No explicit copy will be made by numba in order for this kernel to execute; the array will be transferred automatically by the CUDA driver to the device when accessed in the kernel, and then would be transferred back to the host when accessed there later.

The test cases added run both the lower-level `memallocmanaged` API as well as the `cuda.managed_array` interface. If running on a supported GPU (Pascal or newer) on Linux, `test_managed_alloc_driver_oversubscribe` will intentionally oversubscribe the GPU (that is, operate on an array larger than GPU memory; this will be a slow test, so it is not part of the default unit tests, but is important functionality to verify and can be run manually). `test_managed_alloc_driver_undersubscribe` tests the less expensive case where the data fits within GPU memory.

Per [NBEP 7](https://numba.pydata.org/numba-doc/dev/proposals/external-memory-management.html), this is not exposed in the External Memory Management interface. However, that proposal was written prior to managed memory support existing in Numba, so this could be revisited at a later date.


maxpkatz:
@awthomp, can you comment on whether these interfaces would solve the use case you had in mind in #4362?

numba/cuda/cudadrv/enums.py
```diff
@@ -115,6 +115,15 @@
 CU_MEMHOSTREGISTER_DEVICEMAP = 0x02
 
 
+# If set, managed memory is accessible from all streams on
+# all devices.
+CU_MEM_ATTACH_GLOBAL = 0x01
+
+# If set, managed memory is not accessible from all streams on
```
gmarkall, line 122: Can the comment explain to what it is accessible from if this is set?

numba/cuda/tests/cudadrv/test_managed_alloc.py
```diff
@@ -0,0 +1,79 @@
...
+        # ensure that this test works correctly on all supported GPUs,
+        # we'll select the size of our memory such that we only
+        # oversubscribe the GPU memory if we're on a Pascal or newer GPU
+        # (compute capability at least 6.0).
+
+        compute_capability_major = ctx.device.compute_capability[0]
+
+        # This test is unsupported on GPUs prior to the Kepler generation.
+
+        self.assertGreaterEqual(compute_capability_major, 3)
```
gmarkall, line 33: This will cause the test to fail if run on a device with less than CC 3.0 - what we want to do instead is skip the test if the device's CC is less than 3.0. To do this you can use something like `_safe_cc_check` in `test_warp_ops`: https://github.com/numba/numba/blob/master/numba/cuda/tests/cudapy/test_warp_ops.py#L101

I'd suggest moving `_safe_cc_check` into `numba.cuda.testing`, since it will be used by more than one test file, and then decorating this test with it to skip the test if the CC is not 3.0 or above. See https://github.com/numba/numba/blob/master/numba/cuda/tests/cudapy/test_warp_ops.py#L234 for an example of using it to skip the test conditionally on the CC.

numba/cuda/tests/cudadrv/test_managed_alloc.py
```diff
@@ -0,0 +1,79 @@
...
+
+        total_mem_size = ctx.get_memory_info().total
+
+        if compute_capability_major >= 6 and platform.system() == 'Linux':
```
gmarkall, line 37: I'd prefer if we had two variants of this test - one with and one without oversubscription, with the one requiring CC 6 and Linux skipped if the system doesn't meet the requirements.

gmarkall:
@maxpkatz Following our OOB conversation, I've added some commits on top of those in this PR on a branch in my repo: https://github.com/gmarkall/numba/tree/cuda-managed-memory

# Changes relating to comments on the diff

I tried to address all the comments I made on the diff earlier:

- Merges in the current master and resolves conflicts + fixes a small flake8 issue.
- Corrects docstrings for `MappedMemory` and `ManagedMemory` to state that they refer to, not own, the data.
- Renames `ManagedPointer` to `ManagedOwnedPointer`.
- Clarifies the comment for `CU_MEM_ATTACH_HOST`, and also adds a definition of `CU_MEM_ATTACH_SINGLE` in `enums.py`.
- Refactors `test_managed_alloc_driver` into two tests, with and without oversubscription. The one with oversubscription is skipped in general, as I think it will cause issues for many people with a GPU with lots of RAM (e.g. RTX 8000 with 48GB of RAM, probably needs at least 192GB system RAM for an oversubscription test).
- Fixes the managed alloc tests so they skip, rather than failing, on systems with unsupported devices. In the end I didn't use `_safe_skip` from `test_warp_ops.py` as I suggested earlier, as I felt they were a bit too different to what was needed here for it to make sense to factorise them together.

# Changes relating to avoiding copies when launching kernels on managed memory

I've added a class, `ManagedArray`, that subclasses `DeviceNDArray` and `np.ndarray` - this is based on the existing `MappedArray` class, and elides the need for an allocation and copy to the GPU for managed arrays. Running the `test_managed_array` test with `NUMBA_CUDA_LOG_LEVEL=DEBUG` without, and with, the subclass then comparing the logs results in the following difference:

```diff
$ diff -u without_subclass.txt with_subclass.txt 
--- without_subclass.txt	2020-09-10 15:25:53.608782780 +0100
+++ with_subclass.txt	2020-09-10 15:25:35.029921537 +0100
@@ -36,19 +36,9 @@
 == CUDA [] DEBUG -- call driver api: cuCtxGetDevice
 == CUDA [] DEBUG -- call driver api: cuCtxGetCurrent
 == CUDA [] DEBUG -- call driver api: cuCtxGetDevice
-== CUDA [] DEBUG -- call driver api: cuCtxGetCurrent
-== CUDA [] DEBUG -- call driver api: cuCtxGetDevice
-== CUDA [] DEBUG -- call driver api: cuMemAlloc_v2
-== CUDA [] DEBUG -- call driver api: cuCtxGetCurrent
-== CUDA [] DEBUG -- call driver api: cuCtxGetDevice
-== CUDA [] DEBUG -- call driver api: cuMemcpyHtoD_v2
 == CUDA [] DEBUG -- call driver api: cuLaunchKernel
 == CUDA [] DEBUG -- call driver api: cuCtxGetCurrent
 == CUDA [] DEBUG -- call driver api: cuCtxGetDevice
-== CUDA [] DEBUG -- call driver api: cuMemcpyDtoH_v2
-== CUDA []  INFO -- add pending dealloc: cuMemFree_v2 800 bytes
-== CUDA [] DEBUG -- call driver api: cuCtxGetCurrent
-== CUDA [] DEBUG -- call driver api: cuCtxGetDevice
 == CUDA [] DEBUG -- call driver api: cuCtxSynchronize
 == CUDA []  INFO -- add pending dealloc: cuMemFree_v2 800 bytes
 == CUDA [] DEBUG -- call driver api: cuCtxGetCurrent
@@ -58,8 +48,6 @@
 == CUDA []  INFO -- reset context of device 0
 == CUDA []  INFO -- dealloc: cuMemFree_v2 800 bytes
 == CUDA [] DEBUG -- call driver api: cuMemFree_v2
-== CUDA []  INFO -- dealloc: cuMemFree_v2 800 bytes
-== CUDA [] DEBUG -- call driver api: cuMemFree_v2
 == CUDA []  INFO -- add pending dealloc: module_unload ? bytes
 == CUDA []  INFO -- dealloc: module_unload ? bytes
 == CUDA [] DEBUG -- call driver api: cuModuleUnload
```

i.e. the allocation, host to device transfer, device to host transfer, and deallocation no longer occur.

# Other changes / observations

- I moved `memallocmanaged` up into `HostOnlyCUDAMemoryManager`, which I think is the right place for it - anyone writing an EMM plugin based on the `HostOnlyCUDAMemoryManager` is probably not going to want to have to implement managed allocations themselves, 
- I notice that there is nothing using / testing the use of `memallocmanaged` with the `attach_global` flag set to `False` - I wonder if this should be exposed in the `managed_array` function? And if so, whether it should be a choice between global attachment (accessible from all streams on all GPUs), host attachment (inaccessible from all streams on all GPUs) and single attachment (only accessible from a single stream on the associated device)?
- I've also noticed that mapped arrays (and therefore managed arrays, for which I copied the implementation for the class) accept a stream parameter but do nothing with it - I've filed an issue for this: https://github.com/numba/numba/issues/6233 - we may be able to resolve this as part of this PR.

numba/cuda/tests/cudadrv/test_managed_alloc.py
maxpkatz: This has been addressed; the oversubscription test is skipped in unit testing by default, but can be run manually if the platform matches.

numba/cuda/cudadrv/driver.py
```diff
@@ -1738,6 +1796,10 @@ class MappedOwnedPointer(OwnedPointer, mviewbuf.MemAlloc):
     pass
 
 
+class ManagedPointer(OwnedPointer, mviewbuf.MemAlloc):
```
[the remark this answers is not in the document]
maxpkatz: This has been addressed.

numba/cuda/cudadrv/enums.py
maxpkatz: I have clarified this comment; the original text was from the CUDA driver API documentation, which was admittedly not helpful. For those curious, more information about this can be found in the API documentation for [cuStreamMemAttachAsync](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__STREAM.html#group__CUDA__STREAM_1g6e468d680e263e7eba02a56643c50533) and [cuMemAllocManaged](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MEM.html#group__CUDA__MEM_1gb347ded34dc326af404aa02af5388a32).

maxpkatz:
@gmarkall Thank you for the detailed review! I have applied all of your changes from  https://github.com/gmarkall/numba/tree/cuda-managed-memory as cherry-picked commits into this PR.

I am very grateful that you figured out how to resolve the issue of managed arrays avoiding explicit copies when passed into kernels -- the trick of treating it the same way we use pinned memory works well.

gmarkall (approved):
Many thanks for pulling in the changes! I think this is now looking good!

 To add to one or two of the questions above for completeness on this PR:

> I notice that there is nothing using / testing the use of memallocmanaged with the attach_global flag set to False - I wonder if this should be exposed in the managed_array function? And if so, whether it should be a choice between global attachment (accessible from all streams on all GPUs), host attachment (inaccessible from all streams on all GPUs) and single attachment (only accessible from a single stream on the associated device)?

This is now exposed by the `managed_array` function. When global attachment is used, the managed memory is accessible for all streams on all devices. When it is not used, managed memory is only accessible by the host and streams that it is explicitly attached to. (My other questions above stemmed from a misunderstanding of the flag).

@stuartarchibald Could this have a buildfarm run please?

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_108`.

stuartarchibald:
The test run failed on windows 64 , py38, CUDA 10. Fail was:
```
Fail: test_spawn_concurrent_compilation (numba.cuda.tests.cudapy.test_multithreads.TestMultiThreadCompiling)
cuda\tests\cudapy\test_multithreads.py, line 73, in test_spawn_concurrent_compilation

self.assertEqual(p.exitcode, 0, 'test failed in child process')

AssertionError: 3221225477 != 0 : test failed in child process.
```

gmarkall:
> The test run failed on windows 64 , py38, CUDA 10. Fail was:

This is a bit surprising - I'll reboot into Windows shortly and give it a try to see if I can reproduce. Was that CUDA 10.0?

stuartarchibald:
> > The test run failed on windows 64 , py38, CUDA 10. Fail was:
> 
> This is a bit surprising - I'll reboot into Windows shortly and give it a try to see if I can reproduce. Was that CUDA 10.0?

I think so.

gmarkall:
When I run on Windows with CUDA 10.0, an error I get is:

```
ERROR: test_managed_alloc_driver_host_attach (numba.cuda.tests.cudadrv.test_managed_alloc.TestManagedAlloc)
  File "C:\work\numbadev\numba\numba\cuda\tests\cudadrv\test_managed_alloc.py", line 69, in test_managed_alloc_driver_host_attach
    self._test_managed_alloc_driver(0.01, attach_global=False)
  File "C:\work\numbadev\numba\numba\cuda\tests\cudadrv\test_managed_alloc.py", line 86, in _test_managed_alloc_driver
    device_memset(mem, magic, n_bytes)
  File "C:\work\numbadev\numba\numba\cuda\cudadrv\driver.py", line 2464, in device_memset
    fn(device_pointer(dst), val, size, *varargs)
  File "C:\work\numbadev\numba\numba\cuda\cudadrv\driver.py", line 301, in safe_cuda_api_call
  File "C:\work\numbadev\numba\numba\cuda\cudadrv\driver.py", line 336, in _check_error
numba.cuda.cudadrv.driver.CudaAPIError: [1] Call to cuMemsetD8 results in CUDA_ERROR_INVALID_VALUE

Ran 690 tests in 155.619s

FAILED (errors=1, skipped=14)
```

gmarkall:
Perhaps this test should not be run on Windows. From https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#um-requirements -

>  Applications running on Windows (whether in TCC or WDDM mode) will use the basic Unified Memory model as on pre-6.x architectures even when they are running on hardware with compute capability 6.x or higher.

gmarkall:
Another note: according to https://gitter.im/numba/numba-dev?at=5f5a5145a857200e6d506bce, the buildfarm is running this test with a Tesla K40c and a GTX 750 Ti.

gmarkall:
I am getting the error with both WDDM and TCC mode:

```
(numbapy38cuda10) PS C:\work\numbadev\numba> & 'C:\Program Files\NVIDIA Corporation\NVSMI\nvidia-smi.exe'
Tue Oct 06 20:45:44 2020
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 451.82       Driver Version: 451.82       CUDA Version: 11.0     |
|-------------------------------+----------------------+----------------------+
| GPU  Name            TCC/WDDM | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|===============================+======================+======================|
|   0  Quadro RTX 8000     TCC  | 00000000:15:00.0 Off |                  Off |
| 33%   32C    P8    25W / 260W |      1MiB / 48767MiB |      0%      Default |
+-------------------------------+----------------------+----------------------+
|   1  Quadro P2200       WDDM  | 00000000:2D:00.0  On |                  N/A |
| 48%   37C    P8     7W /  75W |    590MiB /  5120MiB |      7%      Default |
+-------------------------------+----------------------+----------------------+

+-----------------------------------------------------------------------------+
| Processes:                                                                  |
|  GPU   GI   CI        PID   Type   Process name                  GPU Memory |
|        ID   ID                                                   Usage      |
|=============================================================================|
|    1   N/A  N/A       660    C+G   ...zilla Firefox\firefox.exe    N/A      |
|    1   N/A  N/A       708    C+G   ...zilla Firefox\firefox.exe    N/A      |
|    1   N/A  N/A      1272    C+G   Insufficient Permissions        N/A      |
|    1   N/A  N/A      1684    C+G   ...5n1h2txyewy\SearchApp.exe    N/A      |
|    1   N/A  N/A      2084    C+G   C:\Windows\explorer.exe         N/A      |
|    1   N/A  N/A      2384    C+G   ...y\ShellExperienceHost.exe    N/A      |
|    1   N/A  N/A      3604    C+G   ...bbwe\Microsoft.Photos.exe    N/A      |
|    1   N/A  N/A      5608    C+G   ...5n1h2txyewy\SearchApp.exe    N/A      |
|    1   N/A  N/A      6412    C+G   ...zilla Firefox\firefox.exe    N/A      |
|    1   N/A  N/A      7788    C+G   ...cw5n1h2txyewy\LockApp.exe    N/A      |
|    1   N/A  N/A      7848    C+G   ...artMenuExperienceHost.exe    N/A      |
|    1   N/A  N/A      8988    C+G   ...nputApp\TextInputHost.exe    N/A      |
|    1   N/A  N/A      9756    C+G   ...3d8bbwe\MicrosoftEdge.exe    N/A      |
|    1   N/A  N/A     10084    C+G   ...lPanel\SystemSettings.exe    N/A      |
+-----------------------------------------------------------------------------+
```

gmarkall:
(From OOB conversation) the test is invalid because it uses `cuMemsetD8` to set host-attached memory, which is invalid. Though it is surprising that this error isn't seen on Linux.

gmarkall:
@stuartarchibald Now the test that shouldn't be running is disabled on Windows, could we have another buildfarm run please? (the test is a different one to the one you saw an error in, but I'm guessing that the test that shouldn't have run was messing up state somehow)

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_111`.

stuartarchibald:
> (From OOB conversation) the test is invalid because it uses `cuMemsetD8` to set host-attached memory, which is invalid. Though it is surprising that this error isn't seen on Linux.

If this is making it invalid, it is just working by luck on linux or is this a behavioural difference between windows and linux?

gmarkall:
> If this is making it invalid, it is just working by luck on linux or is this a behavioural difference between windows and linux?

It's a behavioural difference - Linux is able to page the host-attached memory into the GPU memory space, but Windows is not.

stuartarchibald:
> > If this is making it invalid, it is just working by luck on linux or is this a behavioural difference between windows and linux?
> 
> It's a behavioural difference - Linux is able to page the host-attached memory into the GPU memory space, but Windows is not.

Thanks for confirming.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_111`.

passed.

stuartarchibald (approved):
Thanks for the patch, looks good. Thanks for reviewing @gmarkall.

stuartarchibald:
@maxpkatz Congratulations on your first contribution to Numba!

PR #5912, opened 2020-06-23
Fix bad attr access on certain typing templates breaking exceptions.

### LINKED ISSUE #5844: Refactor source location info into Template
For code like: https://github.com/numba/numba/pull/5833/files/831da5efbbf4dbc547b216e9c61eca18931b8170#diff-758157deb494d50a6d6c93447b370a05

Reference:
https://github.com/numba/numba/pull/5833#discussion_r437701066

### LINKED ISSUE #5860: Typing errors in device functions aren't properly reported

With current master, the following:

```python
from numba import cuda

@cuda.jit(device=True)
def dev_func(x):
    return floor(x)


@cuda.jit
def kernel_func():
    dev_func(1.5)

kernel_func[1, 1]()
```

where `floor` is untyped because it isn't imported, results in the message:

```
Internal error at <numba.core.typeinfer.CallConstraint object at 0x7f9e808a3dc0>.
'DeviceFunctionTemplate' object has no attribute '__name__'
During: resolving callee type: Function(<numba.cuda.compiler.DeviceFunctionTemplate object at 0x7f9e8b7fdca0>)
During: typing of call at bug.py (10)

File "bug.py", line 10:
def kernel_func():
    dev_func(1.5)
    ^
```

The error message used to say:

```
numba.errors.TypingError: Failed in nopython mode pipeline (step: nopython frontend)
Invalid use of Function(<numba.cuda.compiler.DeviceFunctionTemplate object at 0x7faffda40e90>) with argument(s) of type(s): (float64)
 * parameterized
In definition 0:
Untyped global name 'floor': cannot determine Numba type of <class 'numba.ir.UndefinedType'>

File "bug.py", line 5:
def dev_func(x):
    return floor(x)
    ^

    raised from /home/gmarkall/miniconda3/envs/numba048/lib/python3.7/site-packages/numba/typeinfer.py:1366
In definition 1:
Untyped global name 'floor': cannot determine Numba type of <class 'numba.ir.UndefinedType'>

File "bug.py", line 5:
def dev_func(x):
    return floor(x)
    ^

    raised from /home/gmarkall/miniconda3/envs/numba048/lib/python3.7/site-packages/numba/typeinfer.py:1366
[1] During: resolving callee type: Function(<numba.cuda.compiler.DeviceFunctionTemplate object at 0x7faffda40e90>)
[2] During: typing of call at bug.py (10)


File "bug.py", line 10:
def kernel_func():
    dev_func(1.5)
```

I think this is probably related to recent changes to error messages.

stuartarchibald:
As title.

Fixes #5860
Closes #5844

### DIFF
--- a/numba/core/typing/templates.py
+++ b/numba/core/typing/templates.py
@@ -246,7 +246,24 @@ def fold_arguments(pysig, args, kws, normal_handler, default_handler,
     return args
 
 
-class FunctionTemplate(object):
+class _Template(object):
+
+    def get_template_info(self):
+        """
+        Override to specialise the template information that will govern display
+        in error messages that appear to users"""
+        info = {
+            'kind': "unknown",
+            'name': "unknown",
+            'sig': "unknown",
+            'filename': "unknown",
+            'lines': ("unknown", "unknown"),
+            'docstring': "unknown"
+        }
+        return info
+
+
+class FunctionTemplate(_Template):
     # Set to true to disable unsafe cast.
     # subclass overide-able
     unsafe_casting = True
@@ -310,6 +327,22 @@ class AbstractTemplate(FunctionTemplate):
 
         return sig
 
+    def get_template_info(self):
+        impl = getattr(self, "generic")
+        basepath = os.path.dirname(os.path.dirname(numba.__file__))
+        code, firstlineno = inspect.getsourcelines(impl)
+        path = inspect.getsourcefile(impl)
+        sig = str(utils.pysignature(impl))
+        info = {
+            'kind': "overload",
+            'name': getattr(impl, '__qualname__', impl.__name__),
+            'sig': sig,
+            'filename': os.path.relpath(path, start=basepath),
+            'lines': (firstlineno, firstlineno + len(code) - 1),
+            'docstring': impl.__doc__
+        }
+        return info
+
 
 class CallableTemplate(FunctionTemplate):
     """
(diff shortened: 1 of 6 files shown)


numba/core/typing/templates.py
```diff
@@ -246,7 +246,24 @@ def fold_arguments(pysig, args, kws, normal_handler, default_handler,
     return args
 
 
-class FunctionTemplate(object):
+class _Template(object):
+
+    def get_template_info(self):
```
sklam, line 251: Why can't this be in `FunctionTemplate`? I don't see other class inheriting from `_Template`.

numba/core/typing/templates.py
```diff
@@ -246,7 +246,24 @@ def fold_arguments(pysig, args, kws, normal_handler, default_handler,
...
+        in error messages that appear to users"""
+        info = {
+            'kind': "unknown",
+            'name': "unknown",
+            'sig': "unknown",
+            'filename': "unknown",
+            'lines': ("unknown", "unknown"),
+            'docstring': "unknown"
+        }
+        return info
```
sklam, line 263: This implementation seems unneeded. All subclasses are overriding this. Maybe make this a `abstractmethod` to force subclass to implement it.

numba/core/typing/templates.py
stuartarchibald: It can be I think in hindsight, at first, wasn't so sure as there appeared to be some strange inheritance.

numba/core/typing/templates.py
stuartarchibald: Can do, I had that at one point and then removed it as the inheritance was less clear, I think now having patched this it'd probably be fine/the right thing to do.

numba/core/typing/templates.py
stuartarchibald: 3ecedbb

numba/core/typing/templates.py
stuartarchibald: 3ecedbb

PR #5918, opened 2020-06-24
Fix cuda test due to #5876

sklam:
Error rewrite needs to check freevars

fixes failure in `numba.cuda.tests.cudapy.test_device_func.TestDeviceFunc.test_cpu_dispatcher_invalid`


sklam:
smoking `numba_smoketest_cuda_62`

PR #5920, opened 2020-06-24
Update master changelog for latest change in 0.50.1

sklam:
as titled


stuartarchibald:
Missed 05dd73f, well caught, it's still approved.

PR #5921, opened 2020-06-24
Don't try to analyze the broadcast if the shape of some part is not known.

### LINKED ISSUE #5735: failure to use Prange with @njit(parallel=True) on loop which can be parallelised
I am using a function func1 which is as follows :-

DrTodd13:
Resolves #5735

### DIFF
--- a/numba/parfors/array_analysis.py
+++ b/numba/parfors/array_analysis.py
@@ -2811,9 +2811,12 @@ class ArrayAnalysis(object):
                 arrs[0],
                 self._call_assert_equiv(scope, loc, equiv_set, arrs),
             )
-        return self._broadcast_assert_shapes(
-            scope, equiv_set, loc, shapes, names
-        )
+        if None not in shapes:
+            return self._broadcast_assert_shapes(
+                scope, equiv_set, loc, shapes, names
+            )
+        else:
+            return None, []
 
     def _broadcast_assert_shapes(self, scope, equiv_set, loc, shapes, names):
         """Produce assert_equiv for sizes in each dimension, taking into


DrTodd13:
@sklam I believe I fixed the obvious cause of the issue but now there's an error with type inference that I'd appreciate it if you looked at.
```import numpy as np
from numba import jit, njit, prange

spk1=np.arange(10)

#@njit(cache=True)
@njit(cache=True, parallel=True)
def cisi_numba_para(spk1):
    s=len(spk1)
    chunks=np.expand_dims(spk1, 0)
    n=chunks.shape[0]
    ttt = 0.0
    for i in prange(n):
        chunk=chunks[i]
        if i==n-1:
            chunk=chunk[:3]

        chunk2=spk1

        a1 = chunk*np.ones((chunk2.shape[0], chunk.shape[0]))
        ttt += a1[0][0]

    return ttt

cisi=cisi_numba_para(spk1)
```

sklam:
I'm looking at the error:

```pytb
numba.core.errors.TypingError: Failed in nopython mode pipeline (step: nopython mode backend)
Type of variable 'ttt_2.1' cannot be determined, operation: $ttt.2.176, location: pr5921.py (18)

File "pr5921.py", line 18:
def cisi_numba_para(spk1):
    <source elided>

        chunk2 = spk1
        ^

During: lowering "id=2[LoopNest(index_variable = parfor_index.43, range = (0, $n.51, 1))]{48: <ir.Block at pr5921.py (13)>, 70: <ir.Block at pr5921.py (16)>, 82: <ir.Block at pr5921.py (18)>}Var(parfor_index.43, pr5921.py:13)" at pr5921.py (13)
```

sklam:
The type variables in the highlighted part are undecided. 
numba_ir__numba_parfor_gufunc_0x11b3c8390$2

sklam:
@DrTodd13, I think the highlighted part from the above CFG should be part of the first loopnest (prange). For example, in this variation where I commented out the second loopnest (for the `np.ones)`, the first loopnest (prange) includes the `ttt += a1[0]`.

```python
from numba import jit, njit, prange
import numpy as np

spk1 = np.arange(10)

@njit(cache=True, parallel=True)
def cisi_numba_para(spk1):
    s = len(spk1)
    chunks = np.expand_dims(spk1, 0)
    n = chunks.shape[0]
    ttt = 0.0
    for i in prange(n):
        chunk = chunks[i]
        if i == n - 1:
            chunk = chunk[:3]

        chunk2 = spk1

        # Problem seems to be related to the array allocator turned parfors
        a1 = chunk #* np.ones((chunk2.shape[0], chunk.shape[0]))
        ttt += a1[0]#, 0]

    return ttt


cisi = cisi_numba_para(spk1)

```

DrTodd13:
@sklam I'm working through some of these issues.  First off, there's an arrayexpr in there where the the size of one of the vars (chunk) cannot be statically analyzed due to its assignment in a conditional.  ArrayAnalysis creates size variables for the output of this arrayexpr and then during parfor creation, the parfor code asks for the shape of the output to use for the loop nests and it gets the variables that are only created after the arrayexpr is complete.  So, we can either make this arrayexpr not a parfor or my preferred solution is to pre-compute the sizes for use by the parfor.  I am trying to add this latter approach in those cases where we can't statically analyze the shape.

I want ArrayAnalysis to insert a call to before the arrayexpr to calculate sizes so I'm trying to do add the runtime_broadcast_assert_shapes call as below:

bvar1 = runtime_broadcast_assert_shapes(a,b)
c = arrayexpr(mul, [a,b])

I was trying to define this as a register_jitable as follows:

```
@register_jitable
def runtime_broadcast_assert_shapes(arg0, arg1):
    new_shape = []
    shapes = [arg0.shape, arg1.shape]
    max_dim = max([len(shape) for shape in shapes])
    for i in range(max_dim):
        sizes = []
        for shape in shapes:
            if i < len(shape):
                size = shape[len(shape) - 1 - i]
                if size != 1:
                    sizes.append(size)  # non-1 size to front
        if sizes == []:
            sizes.append(1)
        new_shape.append(sizes[0])

    rev = new_shape.reverse()
    return rev
```

With this approach, I'm running into a lot of typing errors.  At the moment, I get:

```
No implementation of function Function(<function runtime_broadcast_assert_shapes at 0x7f606797f510>) found for signature:

 >>> runtime_broadcast_assert_shapes(array(int64, 1d, C), array(float64, 2d, C))

 Overload in function 'register_jitable.<locals>.wrap.<locals>.ov_wrap': File: numba/core/extending.py: Line 142.
   With argument(s): '(array(int64, 1d, C), array(float64, 2d, C))':
  Type of variable 'i' cannot be determined, operation: $phi48.1, location: /mnt/home/taanders/numba/intel_numba/numba/numba/parfors/array_analysis.py (1057)

  File "numba/parfors/array_analysis.py", line 1057:
  def runtime_broadcast_assert_shapes(arg0, arg1):
      <source elided>
      max_dim = max([len(shape) for shape in shapes])
      for i in range(max_dim):
      ^

  raised from /mnt/home/taanders/numba/intel_numba/numba/numba/core/typeinfer.py:1097
```

There's been several more errors I've been able to change the code to get rid of but I don't know how to get rid of this one.  Any suggestions?

sklam:
```python
from numba import njit, typed
import numpy as np

@njit
def foo(arg0, arg1):
    new_shape = []
    shapes = []
    shapes.append(list(arg0.shape))
    shapes.append(list(arg1.shape))
    max_dim = max([len(shape) for shape in shapes])
    for i in range(max_dim):
        sizes = []
        for shape in shapes:
            if i < len(shape):
                size = shape[len(shape) - 1 - i]
                if size != 1:
                    sizes.append(size)  # non-1 size to front
        if len(sizes) == 0:
            sizes.append(1)
        new_shape.append(sizes[0])

    rev = new_shape.reverse()
    return rev


r = foo(np.arange(10), np.arange(20).reshape(2, 10))
print(r)
```

DrTodd13:
@sklam There may be a couple issues.  Actually, this function should return a tuple.  If instead of "rev = new_shape.reverse(); return rev", I say as an experiment "return tuple(new_shape)" then I get the following:
```
  No implementation of function Function(<class 'tuple'>) found for signature:

   >>> tuple(list(int64))

     Overload of function 'tuple': File: numba/core/typing/builtins.py: Line 574.
       With argument(s): '(list(int64))':
      No match.
```

Also, I suspect that maybe new_shape.reverse() is not being typed correctly because when I do context.reoslve_function_type(...) it thinks this function returns something of none type.

Please advise.

sklam:
In broadcasting, ndim is known to be the greatest ndim of the inputs. If you can get the ndim as a constant, it can be used with https://github.com/numba/numba/blob/ad119885ba52c50b117368479db7e1cc502b2197/numba/np/unsafe/ndarray.py#L42 to make a tuple

DrTodd13:
@sklam Yes, I had already got past that point.  Up till now, we've forced parfors to operate on arrays of equal dimensions.  Thus, after my change to create the tuple we now get an assertion that the arrays are not of equal dimension, because one is 1D and the other is 2D.   This would normally be caught earlier if the static analysis could determine the dimension mismatch but in this case one of the arrays didn't have a static analysis.  This is a somewhat easier problem than generalized broadcasting and I'm still thinking through the issues on supporting this because it feels like a big change.

DrTodd13:
Superseded by some other change.

PR #5922, opened 2020-06-24
Reduce objmode and unpickling overhead

sklam:
Depends on https://github.com/numba/numba/pull/6017
Closes https://github.com/numba/numba/pull/6024

Proposed patch to reduce objmode function overhead and unpickling overhead.

- https://github.com/numba/numba/commit/831d5bf62dfa20006d68c05a3673cf0b4e6c754e Fixes objmode function unnecessarily unpickling the type object of `OmittedArg`, which is only needed for function that has default arguments.
- https://github.com/numba/numba/commit/2a6ac69f61058f5d0671679ff39bdb83a55ebf69 keeps a memo (a `dict`) to cache unpickled object by C `numba_unpickle()` to reduce overhead at unpickling at runtime. Keeping a memo allows us to clear the cached objects to reduce memory consumption.

**Background of the runtime unpickling**

Numba pickle python objects and reconstructs them at runtime for certain operations (like boxing, unboxing) that needs to reference live python objects. But a compiled function cannot hold references to live objects directly; otherwise, it can't be cached. Pickling let's us get around that problem by serializing live objects into a byte string that we can reconstruct at runtime.

### DIFF
--- a/numba/core/callconv.py
+++ b/numba/core/callconv.py
@@ -316,7 +316,7 @@ class _MinimalCallHelper(object):
             msg = "unknown error %d in native function" % exc_id
             return SystemError, (msg,)
 
-# excinfo_t is the same LLVM type created by serialize_uncache()
+# The structure type constructed by PythonAPI.serialize_uncached()
 # i.e a {i8* pickle_buf, i32 pickle_bufsz, i8* hash_buf}
 excinfo_t = ir.LiteralStructType([GENERIC_POINTER, int32_t, GENERIC_POINTER])
 excinfo_ptr_t = ir.PointerType(excinfo_t)
--- a/numba/core/callconv.py
+++ b/numba/core/callconv.py
@@ -316,8 +316,9 @@ class _MinimalCallHelper(object):
             msg = "unknown error %d in native function" % exc_id
             return SystemError, (msg,)
 
-
-excinfo_t = ir.LiteralStructType([GENERIC_POINTER, int32_t])
+# excinfo_t is the same LLVM type created by serialize_uncache()
+# i.e a {i8* pickle_buf, i32 pickle_bufsz, i8* hash_buf}
+excinfo_t = ir.LiteralStructType([GENERIC_POINTER, int32_t, GENERIC_POINTER])
 excinfo_ptr_t = ir.PointerType(excinfo_t)
 
 
--- a/numba/core/pylowering.py
+++ b/numba/core/pylowering.py
@@ -169,7 +169,7 @@ class PyLower(BaseLower):
 
     @utils.cached_property
     def _omitted_typobj(self):
-        """Return a `OmittedArg` type instance in LLVM value suitable for
+        """Return a `OmittedArg` type instance as a LLVM value suitable for
         testing at runtime.
         """
         from numba.core.dispatcher import OmittedArg
--- a/numba/core/pylowering.py
+++ b/numba/core/pylowering.py
@@ -5,6 +5,7 @@ Lowering implementation for object mode.
 
 import builtins
 import operator
+import inspect
 
 from llvmlite.llvmpy.core import Type, Constant
 import llvmlite.llvmpy.core as lc
@@ -75,10 +76,6 @@ class PyLower(BaseLower):
     def pre_lower(self):
         super(PyLower, self).pre_lower()
         self.init_pyapi()
-        # Pre-computed for later use
-        from numba.core.dispatcher import OmittedArg
-        self.omitted_typobj = self.pyapi.unserialize(
-            self.pyapi.serialize_object(OmittedArg))
 
     def post_lower(self):
         pass
@@ -170,6 +167,15 @@ class PyLower(BaseLower):
         else:
             raise NotImplementedError(type(inst), inst)
 
+    @utils.cached_property
+    def _omitted_typobj(self):
+        """Return a `OmittedArg` type instance in LLVM value suitable for
+        testing at runtime.
+        """
+        from numba.core.dispatcher import OmittedArg
+        return self.pyapi.unserialize(
+            self.pyapi.serialize_object(OmittedArg))
+
     def lower_assign(self, inst):
         """
         The returned object must have a new reference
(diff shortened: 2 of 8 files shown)


sklam:
There are some scary bugs & segfault

sklam:
There's a reproducible segfault with ` numba.tests.test_lists.TestLists.test_index3`

sklam:
Segfault is happening at https://github.com/python/cpython/blob/4bfcffe16e9742c154f54ae96b5b36903500abaa/Objects/typeobject.c#L3306

```
Program received signal SIGSEGV, Segmentation fault.
0x00005555556e0bdd in type_dealloc () at /tmp/build/80754af9/python_1565725737370/work/Objects/typeobject.c:3306
3306	/tmp/build/80754af9/python_1565725737370/work/Objects/typeobject.c: No such file or directory.
(gdb) bt
#0  0x00005555556e0bdd in type_dealloc () at /tmp/build/80754af9/python_1565725737370/work/Objects/typeobject.c:3306
#1  0x0000555555644f45 in _PyEval_EvalFrameDefault (f=<optimized out>, throwflag=<optimized out>)
    at /tmp/build/80754af9/python_1565725737370/work/Python/ceval.c:1842
#2  0x00005555556b7ccb in function_code_fastcall (globals=<optimized out>, nargs=4, args=<optimized out>, 
    co=<optimized out>) at /tmp/build/80754af9/python_1565725737370/work/Objects/call.c:283
#3  _PyFunction_FastCallKeywords () at /tmp/build/80754af9/python_1565725737370/work/Objects/call.c:408
#4  0x000055555571fa93 in call_function (kwnames=0x0, oparg=<optimized out>, pp_stack=<synthetic pointer>)
    at /tmp/build/80754af9/python_1565725737370/work/Python/ceval.c:4616
#5  _PyEval_EvalFrameDefault () at /tmp/build/80754af9/python_1565725737370/work/Python/ceval.c:3110
#6  0x00005555556b7ccb in function_code_fastcall (globals=<optimized out>, nargs=1, args=<optimized out>, 
    co=<optimized out>) at /tmp/build/80754af9/python_1565725737370/work/Objects/call.c:283
#7  _PyFunction_FastCallKeywords () at /tmp/build/80754af9/python_1565725737370/work/Objects/call.c:408
#8  0x000055555571f806 in call_function (kwnames=0x0, oparg=<optimized out>, pp_stack=<synthetic pointer>)
    at /tmp/build/80754af9/python_1565725737370/work/Python/ceval.c:4616
#9  _PyEval_EvalFrameDefault () at /tmp/build/80754af9/python_1565725737370/work/Python/ceval.c:3124
#10 0x0000555555668539 in _PyEval_EvalCodeWithName ()
    at /tmp/build/80754af9/python_1565725737370/work/Python/ceval.c:3930
```

sklam:
the problem is likely to be in `numba_do_raise`

sklam:
`numba_do_raise` is too complicated. i think the logical thing to do is to refactor it to simplify the controlflow and the refcount management. Error in refcount is the cause for the segfault; i.e. dropping Py_DECREF in that function "fixes" the segfault.

sklam:
attempting and testing fix at https://github.com/numba/numba/pull/6017

sklam:
https://github.com/numba/numba/pull/6024 tested this PR combined with https://github.com/numba/numba/pull/6017 which fixes the segfault seen here.

stuartarchibald:
> #6024 tested this PR combined with #6017 which fixes the segfault seen here.

Manually tested, it certainly doesn't segfault when combined.

numba/core/callconv.py
```diff
@@ -316,8 +316,9 @@ def get_exception(self, exc_id):
             msg = "unknown error %d in native function" % exc_id
             return SystemError, (msg,)
 
-
-excinfo_t = ir.LiteralStructType([GENERIC_POINTER, int32_t])
+# The structure type constructed by PythonAPI.serialize_uncached()
+# i.e a {i8* pickle_buf, i32 pickle_bufsz, i8* hash_buf}
```
stuartarchibald, line 320: The docstring for `serialize_uncached` needs updating, it still refers to the variant without the hash. https://github.com/numba/numba/pull/5922/files#diff-e3116b002a61bdf2a13aff3dda7ed96fR1313

numba/core/pylowering.py
```diff
@@ -170,6 +167,15 @@ def lower_inst(self, inst):
         else:
             raise NotImplementedError(type(inst), inst)
 
+    @utils.cached_property
+    def _omitted_typobj(self):
+        """Return a `OmittedArg` type instance in LLVM value suitable for
```
stuartarchibald, line 172:
```suggestion
        """Return a `OmittedArg` type instance as a LLVM value suitable for
```

numba/core/serialize.py
```diff
@@ -133,6 +133,32 @@ def _rebuild_code(marshal_version, bytecode_magic, marshalled):
     return marshal.loads(marshalled)
 
 
+# Keep unpickled object via `numba_unpickle` alive.
+_unpickled_memo = {}
+
+
+def _numba_unpickle(address, bytedata, hashed):
```
stuartarchibald, line 140: Is there merit in adding a couple of tests just to make sure memoization is working as expected?

numba/core/callconv.py
sklam: updated in https://github.com/numba/numba/pull/5922/commits/ebbae83720e6c29ef6fbb756298211d24869279f

numba/core/pylowering.py
sklam: updated in https://github.com/numba/numba/pull/5922/commits/ebbae83720e6c29ef6fbb756298211d24869279f

numba/core/serialize.py
sklam: added a simple test in https://github.com/numba/numba/pull/5922/commits/17e72af86d927aa5d01bd7f1fbb80f8b036b0797 for the memoization of its output

numba/core/pylowering.py
```diff
@@ -170,6 +167,15 @@ def lower_inst(self, inst):
         else:
             raise NotImplementedError(type(inst), inst)
 
+    @utils.cached_property
+    def _omitted_typobj(self):
+        """Return a `OmittedArg` type instance as a LLVM value suitable for
+        testing at runtime.
+        """
+        from numba.core.dispatcher import OmittedArg
```
stuartarchibald, line 175: this here due to circular import problems?

numba/core/pylowering.py
```diff
@@ -188,21 +194,29 @@ def lower_assign(self, inst):
         elif isinstance(value, ir.Yield):
             return self.lower_yield(value)
         elif isinstance(value, ir.Arg):
+            # TODO: check why value.name may not exist
```
stuartarchibald, line 197: This TODO apply still? What's the context that triggers it?

numba/core/pylowering.py
sklam: yes, confirmed. can't move it to module scope

numba/core/pylowering.py
sklam: It no longer applies. To be removed.

sklam: removed in https://github.com/numba/numba/pull/5922/commits/d328ac507d7c05fa731e856972f22344e7151b34

PR #5938, opened 2020-06-30
Initial support for dynamic gufuncs

guilhermeleobas:
As title. This PR enables the following:

```python
@guvectorize([(int64[:], int64, int64[:])], '(n),()->(n)')
def g(x, y, res):
    for i in range(x.shape[0]):
        res[i] = x[i] + y

@guvectorize('(n),()->(n)')
def f(x, y, res):
    for i in range(x.shape[0]):
        res[i] = x[i] + y


x = np.arange(5, dtype=np.int32)
y = nb_types.int32(10)
res = np.zeros(5, dtype=np.int32)
g(x, y, res)
f(x, y, res)
```

### DIFF
--- a/docs/source/user/vectorize.rst
+++ b/docs/source/user/vectorize.rst
@@ -356,7 +356,7 @@ on dynamic compilation.
 Dynamic generalized universal functions
 =======================================
 
-Similar to a dynamic universal function, if you do not specif any types to
+Similar to a dynamic universal function, if you do not specify any types to
 the :func:`~numba.guvectorize` decorator, your Python function will be used
 to build a dynamic generalized universal function, or :class:`~numba.GUFunc`.
 For example::
--- a/docs/source/user/vectorize.rst
+++ b/docs/source/user/vectorize.rst
@@ -418,6 +418,6 @@ One can also verify that Numpy ufunc casting rules are working as expected::
    >>> g(x, y, res)
    >>> res
 
-If you need support for varios type signatures, you should not rely on dynamic
+If you need precise support for various type signatures, you should not rely on dynamic
 compilation and instead, specify the types them as first
 argument in the :func:`~numba.guvectorize` decorator.
--- a/docs/source/user/vectorize.rst
+++ b/docs/source/user/vectorize.rst
@@ -157,6 +157,11 @@ argument, which must be filled in by the function.  This is because the
 array is actually allocated by NumPy's dispatch mechanism, which calls into
 the Numba-generated code.
 
+Similar to :func:`~numba.vectorize` decorator, :func:`~numba.guvectorize`
+also has two modes of operation: Eager, or decoration-time compilation and
+lazy, or call-time compilation.
+
+
 Here is a very simple example::
 
    @guvectorize([(int64[:], int64, int64[:])], '(n),()->(n)')
@@ -347,3 +352,72 @@ floating-point values.  For example::
 If you require precise support for various type signatures, you should
 specify them in the :func:`~numba.vectorize` decorator, and not rely
 on dynamic compilation.
+
+Dynamic generalized universal functions
+=======================================
+
+Similar to a dynamic universal function, if you do not specif any types to
+the :func:`~numba.guvectorize` decorator, your Python function will be used
+to build a dynamic generalized universal function, or :class:`~numba.GUFunc`.
+For example::
+
+   from numba import guvectorize
+
+   @guvectorize('(n),()->(n)')
+   def g(x, y, res):
+       for i in range(x.shape[0]):
+           res[i] = x[i] + y
+
+We can verify the resulting function :func:`g` is a :class:`~numba.GUFunc`
+instance that starts with no supported input types. For instance::
+
+   >>> g
+   <numba._GUFunc 'g'>
+   >>> g.ufunc
+   <ufunc 'g'>
+   >>> g.ufunc.types
+   []
+
+Similar to a :class:`~numba.DUFunc`, as one make calls to :func:`g()`,
+numba generates new kernels for previously unsupported input types. The
+following set of interpreter interactions will illustrate how dynamic
+compilation works for a :class:`~numba.GUFunc`::
+
+   >>> x = np.arange(5, dtype=np.int64)
+   >>> y = 10
+   >>> res = np.zeros_like(x)
+   >>> g(x, y, res)
+   >>> res
+   array([5, 6, 7, 8, 9])
+   >>> g.types
+   ['ll->l']
+
+If this was a normal :func:`guvectorize` function, we would have seen an
+exception complaining that the ufunc could not handle the given input types.
+When we call :func:`g()` with the input arguments, numba creates a new loop
+for the input types.
+
+We can add additional loops by calling :func:`g` with new arguments::
+
+   >>> x = np.arange(5, dtype=np.double)
+   >>> y = 2.2
+   >>> res = np.zeros_like(x)
+   >>> g(x, y, res)
+
+We can now verify that Numba added a second loop for dealing with
+floating-point inputs, :code:`"dd->d"`.
+
+   >>> g.types  # shorthand for g.ufunc.types
+   ['ll->l', 'dd->d']
+
+One can also verify that Numpy ufunc casting rules are working as expected::
+
+   >>> x = np.arange(5, dtype=np.int64)
+   >>> y = 2.2
+   >>> res = np.zeros_like(x)
+   >>> g(x, y, res)
+   >>> res
+
+If you need support for varios type signatures, you should not rely on dynamic
+compilation and instead, specify the types them as first
+argument in the :func:`~numba.guvectorize` decorator.
--- a/numba/np/ufunc/gufunc.py
+++ b/numba/np/ufunc/gufunc.py
@@ -117,12 +117,6 @@ class GUFunc(object):
         return types.none(*l)
 
     def __call__(self, *args):
-
-        # a gufunc is dynamic if no calls to `.add()` were made prior to this point  # noqa: E501
-        # if no calls were made, then, self.gufunc_builder._cres is empty
-        if len(self.gufunc_builder._cres) == 0:
-            self._is_dynamic = True
-
         # If compilation is disabled OR it is NOT a dynamic gufunc, call the underlying gufunc  # noqa: E501
         if self._frozen or not self.is_dynamic:
             return self.ufunc(*args)
--- a/numba/np/ufunc/gufunc.py
+++ b/numba/np/ufunc/gufunc.py
@@ -14,10 +14,10 @@ class GUFunc(object):
     """
 
     def __init__(self, py_func, signature, identity=None, cache=None,
-                 targetoptions={}):
+                 is_dynamic=False, targetoptions={}):
         self.ufunc = None
         self._frozen = False
-        self._is_dynamic = False
+        self._is_dynamic = is_dynamic
 
         # GUFunc cannot inherit from GUFuncBuilder because "identity"
         # is a property of GUFunc. Thus, we hold a reference to a GUFuncBuilder
@@ -43,10 +43,6 @@ class GUFunc(object):
     @property
     def is_dynamic(self):
         return self._is_dynamic
-
-    @is_dynamic.setter
-    def is_dynamic(self, value):
-        self._is_dynamic = value
 
     @property
     def __doc__(self):
--- a/numba/np/ufunc/gufunc.py
+++ b/numba/np/ufunc/gufunc.py
@@ -39,7 +39,6 @@ class GUFunc(object):
     @property
     def is_dynamic(self):
         return self._is_dynamic
-    
 
     @property
     def __doc__(self):
@@ -111,18 +110,18 @@ class GUFunc(object):
 
     def __call__(self, *args):
 
-        # a gufunc is dynamic if no calls to `.add()` were made prior to this point
+        # a gufunc is dynamic if no calls to `.add()` were made prior to this point  # noqa: E501
         # if no calls were made, then, self.gufunc_builder._cres is empty
         if len(self.gufunc_builder._cres) == 0:
             self._is_dynamic = True
 
-        # If compilation is disabled OR it is NOT a dynamic gufunc, call the underlying gufunc
+        # If compilation is disabled OR it is NOT a dynamic gufunc, call the underlying gufunc  # noqa: E501
         if self._frozen or not self.is_dynamic:
             return self.ufunc(*args)
 
-        if self._num_args_match(*args) == False:
-            # It is not allowed to call a dynamic gufunc without providing all the arguments
-            # see: https://github.com/numba/numba/pull/5938#discussion_r506429392
+        if self._num_args_match(*args) is False:
+            # It is not allowed to call a dynamic gufunc without providing all the arguments  # noqa: E501
+            # see: https://github.com/numba/numba/pull/5938#discussion_r506429392  # noqa: E501
             msg = (
                 f"Too few arguments for function '{self.__name__}'. "
                 "Note that the pattern `out = gufunc(Arg1, Arg2, ..., ArgN)` "
--- a/numba/np/ufunc/gufunc.py
+++ b/numba/np/ufunc/gufunc.py
@@ -17,6 +17,7 @@ class GUFunc(object):
                  targetoptions={}):
         self.ufunc = None
         self._frozen = False
+        self._is_dynamic = False
 
         # GUFunc cannot inherit from GUFuncBuilder because "identity"
         # is a property of GUFunc. Thus, we hold a reference to a GUFuncBuilder
@@ -36,12 +37,17 @@ class GUFunc(object):
         self._frozen = True
 
     @property
+    def is_dynamic(self):
+        return self._is_dynamic
+    
+
+    @property
     def __doc__(self):
         return self.ufunc.__doc__
 
     @property
     def __name__(self):
-        return self.ufunc.__name__
+        return self.gufunc_builder.py_func.__name__
 
     @property
     def nin(self):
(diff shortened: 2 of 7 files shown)


guilhermeleobas:
It seems that removing these lines in [this](https://github.com/numba/numba/pull/5938/commits/321df611ad8ace91f856f60f2f2e4d43f43503d5) commit, breaks test `TestDUFuncPickling` ([link to azure](https://dev.azure.com/numba/numba/_build/results?buildId=5868&view=logs&j=cf997ec4-3dce-5934-ed71-0c9a02b56c1b&t=114f249c-e275-599a-d9c6-cfec6d211bc9))

guilhermeleobas:
Currently, in Numba (and in Numpy), it is possible to call a guvectorized function as res = g(x, y) rather than g(x, y, res). Having this pattern when types are optional is a bit problematic, since we would have to guess the output type and shape based on signature ((n),()->(n)) and input types (x, y). Should this pattern be forbidden in Numba when types were not specified? cc @sklam

sklam:
Summary from out-of-band disscusion:

> Should this pattern be forbidden in Numba when types were not specified?

Yes, dynamic ufuncs can ban the usage without explicit output argument. It will be difficult to make it work.

guilhermeleobas:
PR is ready for review!

numba/np/ufunc/gufunc.py
```diff
@@ -0,0 +1,116 @@
...
+        # In the latter case, one would have to guess the type
+        # and format (scalar or array). This behavior will be
+        # forbidden for now.
+        # See: https://github.com/numba/numba/pull/5938#issuecomment-661978921
+
+        # self.signature[1] has always length 1
+        if len(dtypes) < len(self.signature[0]) + 1:
+            msg = '`out = gufunc(Arg1, Arg2, ..., ArgN)` is not allowed.'
+            '\nUse `gufunc(Arg1, Arg2, ..., ArgN, out)` instead'
+            raise TypeError(msg)
```
asmeurer, line 90: Is it possible for the number of args to be wrong for other reasons (like forgetting an argument). Hints like this can be nice but they can lead the user astray when they don't actually match what happened. Maybe at least prefix the message with "too few arguments" or something like that.

numba/tests/npyufunc/test_gufunc.py
```diff
@@ -68,6 +68,72 @@ class TestGUFuncParallel(TestGUFunc):
...
 
+class TestDynamicGUFunc(TestCase):
+    target = 'cpu'
+
+    def test_dynamic_matmul(self):
+
+        def check_matmul_gufunc(gufunc, A, B, C):
+            gufunc(A, B, C)
+            Gold = ut.matrix_multiply(A, B)
+            np.testing.assert_allclose(C, Gold, rtol=1e-5, atol=1e-8)
```
sklam, line 79: it should also check that new compilations were made and that they were used.

sklam:
I will do a more in-depth review later this week.

numba/np/ufunc/gufunc.py
```diff
@@ -0,0 +1,116 @@
...
+        return types.none(*l)
+
+    def __call__(self, *args, **kwargs):
+        try:
+            return self.ufunc(*args, **kwargs)
+        except TypeError:
+            sig = self._get_element_wise_signature(*args)
+            self.add(sig)
+            self.build_ufunc()
+            return self.ufunc(*args, **kwargs)
```
guilhermeleobas, line 116: Is there a better way to detect if a gufunc was already compiled for the given types?

numba/np/ufunc/gufunc.py
```diff
@@ -0,0 +1,116 @@
...
+        # 2. args doesn't contains the output array:
+        #      C = gufunc(A, B)
+        #
+        # In the latter case, one would have to guess the type
+        # and format (scalar or array). This behavior will be
+        # forbidden for now.
+        # See: https://github.com/numba/numba/pull/5938#issuecomment-661978921
+
+        # self.signature[1] has always length 1
+        if len(dtypes) < len(self.signature[0]) + 1:
```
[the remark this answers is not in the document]
guilhermeleobas: No, it shouldn't. Well, should work if you do something like the following:

```python
import numpy as np

@guvectorize('(n),()->()')
def f(x, y, res):
    c = 0
    for i in range(x.shape[0]):
        c += x[i] + y
    res[0] = c


x = np.arange(5, dtype=np.int32)
y = np.int32(10)
z = np.arange(5, dtype, np.int32)

f(x, y, z) # trigger compilation

out = f(x, y)
print(out)
print(f.types) # one ufunc loop

f(a.astype(np.float64).reshape(1, 5), y.astype(np.float32), z.astype(np.float64).reshape(1, 5)) # trigger compilation
out = f(x.astype(np.float64).reshape(1, 5), y.astype(np.float32))
print(out)
print(f.types) # two ufunc loop

```

numba/tests/npyufunc/test_gufunc.py
```diff
@@ -68,6 +68,72 @@ class TestGUFuncParallel(TestGUFunc):
...
+
+        gufunc = GUVectorize(matmulcore, '(m,n),(n,p)->(m,p)',
+                             target=self.target)
+
+        matrix_ct = 10
+        A = np.arange(matrix_ct * 2 * 4, dtype=np.float32).reshape(matrix_ct, 2, 4)
```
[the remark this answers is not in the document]
guilhermeleobas: If I change the order of the tests (run the integral type first), then one ensures that two versions will be generated, right?

numba/np/ufunc/gufunc.py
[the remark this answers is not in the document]
guilhermeleobas: > should be using parsed_sig instead.
self.signature is a str

Fixed!

numba/np/ufunc/gufunc.py
guilhermeleobas: Is it ok to check if a gufunc was already compiled in the way I did? Or should I follow the same approach as dufunc?

numba/tests/npyufunc/test_gufunc.py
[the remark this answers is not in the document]
sklam: Looks like it is not casting because of the unsafe cast from int64 to float32. The `TypeError` is:

```
ufunc 'matmulcore' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

If the gufunc starts with a float64 outputs and then with int64 output, the `TypeError` will be:

```
ufunc 'matmulcore' output (typecode 'd') could not be coerced to provided output parameter (typecode 'l') according to the casting rule ''same_kind''
```

numba/np/ufunc/gufunc.py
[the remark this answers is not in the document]
sklam: fix confirmed

numba/np/ufunc/gufunc.py
sklam: I originally thought it was ok. However, after looking into the message in `TypeError`, the exception is probably too generic. First, we can't tell if the error is `TypeError("'NoneType' object is not callable")` because `self.ufunc` is `None`. Secondly, we can't tell if the error is indeed caused by problem with type coercion. Can you try to use the `ufunc_find_matching_loop` so that it can be more precise about what is happening?

guilhermeleobas:
Hi Siu, thanks for the review. CI is currently failing on a [test](https://github.com/numba/numba/blob/a4413366a946459c8f0929e1556ee9ee765fc7c2/numba/tests/npyufunc/test_ufuncbuilding.py#L57-L62) that uses object mode. Can I just remove this test since object mode is schedule for deprecation?

numba/tests/npyufunc/test_errors.py
```diff
@@ -61,7 +61,7 @@ def check_gufunc_raise(self, **vectorize_args):
         arr = np.array([1, 2, -3, 4], dtype=np.int32)
         out = np.zeros_like(arr)
         with self.assertRaises(ValueError) as cm:
-            f(arr, 2, out)
+            f(arr, int32(2), out)
         # The gufunc bailed out after the error
         self.assertEqual(list(out), [2, 4, 0, 0])
```
guilhermeleobas, line 66: Hi @sklam, thanks for the review. I've made an attempt to use `ufunc_find_matching_loop`. By using this function rather than calling the ufunc directly, Numba fails in gufuncs compiled for `int32` with `int64` as input, where numpy works just fine. This change will probably crash any code being used nowadays by numba users. Is `ufunc_find_matching_loop` missing something or is there any way to avoid this?

guilhermeleobas (commented):
Hi Siu, I've address your comments and appreciate your feedback when you have some free time to spare.

numba/tests/npyufunc/test_errors.py
guilhermeleobas: Something that I can do is to revert the code to the old approach (call the ufunc directly) and check the TypeError message:

```python
def __call__(self, *args):
    cant_cast_msg = "not supported for the input types, and the inputs" \
                    " could not be safely coerced to any supported types" \
                    " according to the casting rule"
    nonetype_not_callable = "'NoneType' object is not callable"

    try:
        return self.ufunc(*args)
    except TypeError as e:
        if cant_cast_msg in str(e) or nonetype_not_callable in str(e):
            sig = self._get_signature(*args)
            self.add(sig)
            self.build_ufunc()
            return self.ufunc(*args)
        else:
            raise e
```

guilhermeleobas:
Hi @sklam, CI is all green

sklam:
@guilhermeleobas 

Something wrong with the force push it seems. Previously addressed items were gone. e.g. https://github.com/numba/numba/pull/5938#pullrequestreview-487008549

For future reference, please do not force-push once the PR has been reviewed. It looses the review history.

guilhermeleobas:
Hi @sklam, sorry about that. I made some mistake while rebasing the code.

numba/tests/npyufunc/test_gufunc.py
```diff
@@ -68,6 +68,74 @@ class TestGUFuncParallel(TestGUFunc):
...
+        # The outter (leftmost) dimension must match or numpy broadcasting is performed.
+
+        inp = np.arange(30000, dtype=np.int32).reshape(10000, 3)
+        out = np.zeros(10000, dtype=np.int32)
+        sum_row(inp, out)
+
+        # verify result
+        for i in range(inp.shape[0]):
+            self.assertEqual(out[i], inp[i].sum())
+
```
sklam, line 137: i'd suggest to add a test on the missing output error:

```python
        with self.assertRaisesRegex(TypeError, "Too few arguments for function 'sum_row'."):
            sum_row(inp)
```
Also note that this is failing because the `Gufunc.__call__` doesn't run `_get_signature()` when there's a matching loop due to previously compiled signature. Better to be consistent and always raise if the output argument is missing.

sklam (changes requested):
@guilhermeleobas, can you add the suggested check on the error?

numba/tests/npyufunc/test_gufunc.py
sklam: @guilhermeleobas, I was not clear. If you do `sum_row(inp)` after a `sum_row(inp, out)`, the error about missing output argument is not raised.

numba/np/ufunc/gufunc.py
```diff
@@ -0,0 +1,130 @@
...
+
+        return types.none(*l)
+
+    def __call__(self, *args):
+
+        if self._frozen:
+            return self.ufunc(*args)
+
+        ewise = self._get_ewise_dtypes(args)
+        if not (self.ufunc and ufunc_find_matching_loop(self.ufunc, ewise)):
```
sklam, line 126: missing check for lack of output argument if this branch is `ufunc_find_matching_loop()` returns True.

sklam: ^ this comment corresponds to https://github.com/numba/numba/pull/5938#discussion_r503534505

guilhermeleobas: Hi Siu, If we check for missing lack of output argument when `ufunc_find_matching_loop()` returns True, it will crash tests `test_gufunc.py::TestGUFunc.(test_ufunc_like/test_gufunc)`. This test manually adds signatures by calling `.add` and never calls `disable_compile`.

sklam: In that case, we need to add the logic to tell apart if it is a dynamic gufunc or not. Otherwise, the behavior will be confusing. 

We need the test in https://github.com/numba/numba/pull/5938/commits/9236b15e1e0a38a1fe23aa3f69d44b359d18abe3 and more tests around the error.

guilhermeleobas: Hi @sklam,

Ok, I can do that but do you think this is necessary? For instance, in the example you gave at https://github.com/numba/numba/commit/9236b15e1e0a38a1fe23aa3f69d44b359d18abe3, it is completely ok to omit the `out` argument: `out = sum_row(inp)`. The problem appears only in the first call to it.

sklam: The error behavior should not depend on the ordering of the calls. It will be confusing to users if their gufunc works sometimes by accident. So, I think it is best to guarantee an error for missing output argument for all dynamic gufuncs.

guilhermeleobas: Okay, done! I've moved the typing check part to before any computation is done. This way, numba will check if the number of arguments match in a dynamic gufunc before attempting to compile or call it. Let me know if this approach is what you're looking for.

numba/tests/npyufunc/test_gufunc.py
```diff
@@ -68,6 +68,78 @@ class TestGUFuncParallel(TestGUFunc):
...
+        a pointer to the output location.
+        """
+
+        @guvectorize('(n)->()', target=self.target, nopython=True)
+        def sum_row(inp, out):
+            tmp = 0.
+            for i in range(inp.shape[0]):
+                tmp += inp[i]
+            out[()] = tmp
+
```
sklam, line 125: one would expect the following:
```suggestion
        self.assertTrue(sum_row.is_dynamic)
```
but it is failing.

sklam (changes requested):
The behavior of the `_is_dynamic` is odd.

numba/tests/npyufunc/test_gufunc.py
guilhermeleobas: Hi @sklam, I've improved `is_dynamic` behavior in my last commit. One thing that I notice is that some tests uses `GUVectorize` class directly instead of the `@guvectorize` decorator:

https://github.com/numba/numba/blob/ed4dac1ff741341c14b6e75ad1d97b981d999301/numba/tests/npyufunc/test_gufunc.py#L38-L43

https://github.com/numba/numba/blob/ed4dac1ff741341c14b6e75ad1d97b981d999301/numba/tests/npyufunc/test_gufunc.py#L81-L88

And in those cases it is harder to guess if types will be added later via `.add()` method or if the gufunc will be dynamic.

sklam:
I have two suggestions and I commited to this branch: https://github.com/sklam/numba/commits/pr/5938suggestion
- https://github.com/sklam/numba/commit/e1d514a0121a12e76e33a36f5133302604127419 avoids setting is_dynamic in `__call__` and move the `is_dynamic=True` to right after the constructor.
- https://github.com/sklam/numba/commit/404745e5971a97d2423c5e7d7d992b83904a2941 goes one step further to remove the setter on `is_dynamic` and make the constructor set it.

guilhermeleobas:
Hi @sklam, thanks for the tip. If that's ok, I've cherry-pick those commits into this branch

guilhermeleobas:
Hi @sklam, can you take another look at this PR when you have some cycles to spare?

cristipurdel:
Sorry for asking, but is this supposed to work on 'cuda' target? 
Been waiting for this feature for 2 years to decrease the needed gpu memory.

guilhermeleobas:
Hi @cristipurdel, this feature was designed to work with `target=cpu`. Some more work needs to be done in order to support the cuda target.

numba/np/ufunc/gufunc.py
```diff
@@ -0,0 +1,136 @@
...
+
+        # add return type to signature
+        retty = ewise_types[-1]
+        ret_ndim = len(parsed_sig[-1][0]) or 1  # small hack to return scalar
+        l.append(types.Array(retty, ret_ndim, 'A'))
+
+        return types.none(*l)
+
+    def __call__(self, *args):
+        # If compilation is disabled OR it is NOT a dynamic gufunc, call the underlying gufunc  # noqa: E501
```
sklam, line 116: Better to break this comment into multiple lines.

sklam (commented):
Just some minor comment on the use of `noqa 501`.

sklam:
@guilhermeleobas 

> Edit: Let me know when the code is in a "mergeable" state so that I can write/update the docs.

Yes, it's ready for doc update.

guilhermeleobas:
Hi @sklam, thanks for the review. I've resolved the issues pointed by you in your last review and add a first draft for the documentation.

docs/source/user/vectorize.rst
```diff
@@ -347,3 +352,72 @@ floating-point values.  For example::
 If you require precise support for various type signatures, you should
 specify them in the :func:`~numba.vectorize` decorator, and not rely
 on dynamic compilation.
+
+Dynamic generalized universal functions
+=======================================
+
+Similar to a dynamic universal function, if you do not specif any types to
```
sklam, line 359:
```suggestion
Similar to a dynamic universal function, if you do not specify any types to
```

docs/source/user/vectorize.rst
```diff
@@ -347,3 +352,72 @@ floating-point values.  For example::
...
+
+One can also verify that Numpy ufunc casting rules are working as expected::
+
+   >>> x = np.arange(5, dtype=np.int64)
+   >>> y = 2.2
+   >>> res = np.zeros_like(x)
+   >>> g(x, y, res)
+   >>> res
+
+If you need support for varios type signatures, you should not rely on dynamic
```
sklam, line 421:
```suggestion
If you need precise support for various type signatures, you should not rely on dynamic
```

I'm suggesting adding "precise" as in the doc in dufunc:

> If you require precise support for various type signatures

because dynamic-gufunc can support multiple type signatures but it will compile new one if there are no exact match.

sklam (commented):
I have some minor comments.

sklam (approved):

guilhermeleobas:
Thanks for all the feedback @sklam

PR #5941, opened 2020-06-30
Fix/cuda atomic nanminmax

### LINKED ISSUE #5820: Add back support for atomics that always return the non-NaN values
## Feature request

From @uchytilc in #5556:

stuartarchibald:
Fixes #5820

### DIFF
--- a/numba/cuda/tests/cudapy/test_atomics.py
+++ b/numba/cuda/tests/cudapy/test_atomics.py
@@ -1,5 +1,6 @@
 import random
 import numpy as np
+from textwrap import dedent
 
 from numba import cuda, uint32, uint64, float32, float64
 from numba.cuda.testing import unittest, CUDATestCase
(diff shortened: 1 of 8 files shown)


numba/cuda/tests/cudapy/test_atomics.py
```diff
@@ -530,6 +569,111 @@ def kernel(x):
...
+        vals = np.random.randint(0, 128, size=32).astype(np.float64)
+        vals[1::2] = np.nan
+        res = np.array([128], np.float64)
+        cuda_func = cuda.jit('void(float64[:], float64[:])')(
+            atomic_min_double_oneindex)
+        cuda_func[1, 32](res, vals)
+
+        gold = np.nanmin(vals)
+        np.testing.assert_equal(res, gold)
+
```
stuartarchibald, line 676: Wonder if tests similar to https://github.com/numba/numba/pull/5941/files#diff-265d0d9cde3f18a334cc9633801aa6f1R509-R568 should be added?

sklam:
smoking at `numba_smoketest_cuda_65`

numba/cuda/tests/cudapy/test_atomics.py
sklam: I think we should test the return value just in case the code drift.

numba/cuda/tests/cudapy/test_atomics.py
```diff
@@ -530,6 +569,111 @@ def kernel(x):
...
+        self.check_atomic_nanmax(dtype=np.float32, lo=-65535, hi=65535)
+
+    def test_atomic_nanmax_double(self):
+        self.check_atomic_nanmax(dtype=np.float64, lo=-65535, hi=65535)
+
+    def test_atomic_nanmax_double_shared(self):
+        vals = np.random.randint(0, 32, size=32).astype(np.float64)
+        vals[1::2] = np.nan
+        res = np.array([0], dtype=vals.dtype)
+        cuda_func = cuda.jit('void(float64[:], float64[:])')(atomic_nanmax_double_shared)
```
sklam, line 608: long lines. maybe good to un-ban this file from flake8

numba/cuda/tests/cudapy/test_atomics.py
```diff
@@ -530,6 +569,111 @@ def kernel(x):
...
+        self.check_atomic_nanmin(dtype=np.float32, lo=-65535, hi=65535)
+
+    def test_atomic_nanmin_double(self):
+        self.check_atomic_nanmin(dtype=np.float64, lo=-65535, hi=65535)
+
+    def test_atomic_nanmin_double_shared(self):
+        vals = np.random.randint(0, 32, size=32).astype(np.float64)
+        vals[1::2] = np.nan
+        res = np.array([32], dtype=vals.dtype)
+        cuda_func = cuda.jit('void(float64[:], float64[:])')(atomic_nanmin_double_shared)
```
sklam, line 660: another long lines

numba/cuda/tests/cudapy/test_atomics.py
stuartarchibald: unbanned and fixed in 1b51ebc

gmarkall:
There may be some changes required - the `nanmax` and `nanmin` versions fail to return the old value when it is a NaN (I have some tests written locally and I'm still looking into fixing it). Do we want the old value to always be returned by `nanmin` and `nanmax`? Arguments for / against:

* For: this would give consistency with `min` and `max`, which always return the old value.
* Against: Returning the new value if the old value is a NaN is consistent with treating NaNs as missing data.

I'm inclined towards them not returning the old value when the old value is a NaN, and returning the new value instead, because it makes their semantics more consistent with themselves - thoughts?

stuartarchibald:
@gmarkall feel free to just push to my branch, can ask someone else to help review if it contains both our code:)

gmarkall:
@stuartarchibald I'm not sure I can do that:

```
$ git push stuartarchibald grm-pr-5941:fix/cuda_atomic_nanminmax
ERROR: Permission to stuartarchibald/numba.git denied to gmarkall.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

Unless I'm doing it wrong?

stuartarchibald:
> @stuartarchibald I'm not sure I can do that:
> 
> ```
> $ git push stuartarchibald grm-pr-5941:fix/cuda_atomic_nanminmax
> ERROR: Permission to stuartarchibald/numba.git denied to gmarkall.
> fatal: Could not read from remote repository.
> 
> Please make sure you have the correct access rights
> and the repository exists.
> ```
> 
> Unless I'm doing it wrong?

ah you might need elevated rights on the repo, oh well, I'll just pull https://github.com/gmarkall/numba/commit/9ee813236e8cdaf440ec71484385c0492e613642 in.

stuartarchibald:
Closing, https://github.com/numba/numba/pull/6020 continues this, thanks @gmarkall

PR #5945, opened 2020-07-01
Fixes the jit decorator with target='cuda'

### LINKED ISSUE #5868: TypeError: compile_kernel() got an unexpected keyword argument 'boundscheck'
Hi, I am quite new to `numba` and was trying to play with it. Thus, I was trying to run the following simple example:

```
from numba import jit, cuda 
import numpy as np 
import os

# function optimized to run on gpu
@jit(target ='cuda')
def func2(a):
    for i in range(10000000):
        a[i]+= 1

if __name__ == '__main__':
    n = 10000000
    a = np.ones(n, dtype=np.float64)
    b = np.ones(n, dtype=np.float32)

    func2(a)
    os._exit(0)
```

but for some reason I am getting the `TypeError`:

`TypeError: compile_kernel() got an unexpected keyword argument 'boundscheck'`

I checked on the web for similar issues but I did not really found anything relevant. Any idea what might be wrong.

Thanks.

System info:
Python 3.8
Windows 10
Numba 0.50

hameerabbasi:
Fixes #5868. This PR fixes the `numba.jit` decorator with the `target='cuda'` keyword when the `boundscheck` argument isn't supplied.

### DIFF
--- a/numba/cuda/decorators.py
+++ b/numba/cuda/decorators.py
@@ -51,7 +51,7 @@ def jit(func_or_sig=None, argtypes=None, device=False, inline=False, bind=True,
     if link and config.ENABLE_CUDASIM:
         raise NotImplementedError('Cannot link PTX in the simulator')
 
-    if kws.get('boundscheck') == True:
+    if kws.pop('boundscheck', False):
         raise NotImplementedError("bounds checking is not supported for CUDA")
 
     fastmath = kws.get('fastmath', False)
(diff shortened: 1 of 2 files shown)


hameerabbasi:
I think this is ready for review.

inboxsgk:
I tried this fix on my project. I get the following error 
`numba.cuda.cudadrv.error.NvvmSupportError: libNVVM cannot be found. Do `conda install cudatoolkit`:
[WinError 3] The system cannot find the path specified: 'F:\\Python\\Vocabulary\\venv\\DLLs'`

Numba Version 0.50.1

hameerabbasi:
Hi. The Numba team deprecated the `target` kwarg in `numba.jit`.

You should use `numba.cuda.jit` instead.

gmarkall:
If you also have the issue without the patch applied, please do open an issue or post about the problem you're having on the Numba discourse: https://numba.discourse.group

PR #5946, opened 2020-07-01
Implement literal dictionaries and lists.

stuartarchibald:
As title.
Permits compilation of things like:
```python
@njit
def foo():
  d = {'a': 1,
       'b': 2j,
       'c': 'STRING',
       'd': np.zeros(5),
       'e': ['list', 1, 'of', 2, 'mixed', 3, 'types']}
```

WIP, opened for CI.


sklam:
From out-of-band discussion:

* This feature is pushing the limit of type inference
* We speculate that could be problem with over-specialization due to the capturing of initial-value as part of the type.
* Planned actions:
   * benchmark for problems after merging
   * fix the overspecialization in a new PR

numba/core/types/containers.py
```diff
@@ -642,6 +673,45 @@ def unify(self, typingctx, other):
...
+
+    def unify(self, typingctx, other):
+        """
+        Unify this with the *other* one.
+        """
+        if isinstance(other, LiteralStrKeyDict):
+            for (k1, v1), (k2, v2) in zip(self.literal_value.items(),
+                                          other.literal_value.items()):
+                if k1 != k2: # keys must be same
+                    break
```
sklam, line 705: Branch not covered.

numba/core/types/containers.py
```diff
@@ -642,6 +673,45 @@ def unify(self, typingctx, other):
...
+                if unliteral(v1) != unliteral(v2): # values must be same type
+                    break
```
sklam, line 707: Branch not covered.

numba/core/types/containers.py
```diff
@@ -433,6 +434,34 @@ def __getitem__(self, args):
...
+        _HeterogeneousTuple.is_types_iterable(literal_value)
+        self._literal_init(list(literal_value))
+        tys = [unliteral(x) for x in literal_value]
+        self.types = tuple(literal_value)
+        self.count = len(self.types)
+        self.name = 'LiteralList({})'.format(literal_value)
+        literal_vals = [getattr(x, 'literal_value', None) for x in literal_value]
+
+    def __unliteral__(self):
+        return self
```
sklam, line 451: Couldn't this degrade into a list with initial_value?

numba/core/types/containers.py
```diff
@@ -642,6 +673,45 @@ def unify(self, typingctx, other):
...
+        strkeys = [x.literal_value for x in literal_value.keys()]
+        self.tuple_ty = namedtuple('_ntclazz', ' '.join(strkeys))
+        self.tuple_inst = self.tuple_ty(*literal_value.values())
+        tys = [unliteral(x) for x in literal_value.values()]
+        NamedTuple.__init__(self, tys, self.tuple_ty)
+        self.name = 'LiteralStrKey[Dict]({})'.format(literal_value)
+        literal_vals = [getattr(x, 'literal_value', None) for x in literal_value.values()]
+
+    def __unliteral__(self):
+        return self
```
sklam, line 695: Couldn't this degrade into a dict with initial_value?

numba/cpython/tupleobj.py
```diff
@@ -341,6 +342,12 @@ def static_getitem_tuple(context, builder, sig, args):
     elif isinstance(idx, slice):
         items = cgutils.unpack_tuple(builder, tup)[idx]
         res = context.make_tuple(builder, sig.return_type, items)
+    elif isinstance(tupty, types.LiteralStrKeyDict):
+        # pretend to be a dictionary
+        if isinstance(idxty, types.Literal):
```
sklam, line 347: Looks like the `else` branch is not handled.

numba/core/typeinfer.py
```diff
@@ -277,26 +307,91 @@ class BuildSetConstraint(_BuildContainerConstraint):
...
+                    init_value = self.special_value if literalvty else None
+                    key_type, value_type = tsets[0]
+                    typeinfer.add_type(self.target,
+                                       types.DictType(key_type,
+                                                      value_type,
+                                                      init_value),
+                                       loc=self.loc)
+
+
+class BuildLiteralHomogeneousMapConstraint(object):
```
sklam, line 371: dead class?

stuartarchibald: think so, it's left over from an attempt at generic homogeneous maps.

numba/core/types/containers.py
stuartarchibald: Possibly, but only if the types are coercible? e.g. `['a', 1, 'b', 2]` couldn't degrade.

numba/core/types/containers.py
stuartarchibald: May be, same issue as https://github.com/numba/numba/pull/5946/files#r450826105

numba/core/types/containers.py
stuartarchibald: likely, unify is a bit of a mess because the `__unliteral__`s are not behaving.

numba/core/types/containers.py
stuartarchibald: hmmmm, I'm wondering now how this is reachable, if the types are coercible on construction, a `List` with an `initial_value` will be constructed, if they are not a `LiteralList` is constructed. If the latter, there's no way it can degrade to a `List` as it's not type coercible in the first place. The only situation I think this could occur in is if something is deliberately spoofing a signature for some purpose.

numba/cpython/tupleobj.py
stuartarchibald: will remove, `LiteralStrKeyDict` inherits from `Literal` so it is unreachable, think it's left over from trying to coerce a LiteralDict into a TypedDict based on type inference spotting mutation.

stuartarchibald:
479b425 is failing with:
```
FAIL: test_dict_unify (numba.tests.test_dictobject.TestLiteralStrKeyDict)
  File "/Users/runner/work/1/s/numba/tests/test_dictobject.py", line 1983, in test_dict_unify
    self.assertEqual(foo(100), 'd')
AssertionError: 'CAT' != 'd'
- CAT
+ d
```
think literal dictionary unification is probably wrong

numba/core/types/containers.py
stuartarchibald: Resolved by Poison type, it's essentially invalid.

numba/core/types/containers.py
stuartarchibald: Same as https://github.com/numba/numba/pull/5946/files/3e32c6e662f01ad9cae73c36ead947dd82c1000b#r451469788

numba/core/types/containers.py
stuartarchibald: Think this is fixed.

numba/core/types/containers.py
stuartarchibald: Think this is fixed.

stuartarchibald:
https://github.com/numba/numba/pull/5946#issuecomment-655455731 is fixed by b17f841 on wards.

stuartarchibald:
@sklam this still has over-specialisation issues but is probably in a position where it's close enough to working to be reviewable! Thanks.

sklam:
The new changes are looking good. Feature-wise, it will address the use-cases for https://gist.github.com/sklam/4f08128860485485c7b3dbd2c9149e07. 

I did notice that it is difficult to search for the added feature in the current doc. We should probably address this separately.

I will want to have another look at the specific details in the new `overload_method`.

numba/cpython/listobj.py
```diff
@@ -1119,3 +1120,85 @@ def list_to_list(context, builder, fromty, toty, val):
...
+    # heterogeneous lists are a trick, they are really tuples
+    raise errors.TypingError("Cannot mutate a literal list")
+
+_index_end = types.intp.maxval
+@overload_method(types.LiteralList, 'index')
+def literal_list_index(lst, x, start=0, end=_index_end):
+    # TODO: To make this work, need consts as slice for start/end so as to
+    # be able to statically analyse the bounds, then its a just loop body
+    # versioning based iteration along with enumerate to find the item
+    if isinstance(lst, types.LiteralList):
```
sklam, line 1147: Untested. 

Propose adding the following to `TestLiteralLists`
```python
    def test_index(self):
        @njit
        def foo():
            l = ['a', 1]
            l.index('a')

        with self.assertRaises(errors.TypingError) as raises:
            foo()
        expect = "list.index is unsupported for literal lists"
        self.assertIn(expect, str(raises.exception))
```

numba/cpython/listobj.py
```diff
@@ -1119,3 +1120,85 @@ def list_to_list(context, builder, fromty, toty, val):
...
+        msg = "list.index is unsupported for literal lists"
+        raise errors.TypingError(msg)
+
+@overload_method(types.LiteralList, 'count')
+def literal_list_count(lst, x):
+    if isinstance(lst, types.LiteralList):
```
sklam, line 1153: Untested.

Propose adding the following to `TestLiteralLists`
```python
    def test_copy(self):
        @njit
        def foo():
            l = ['a', 1].copy()
            return l[0], l[1]

        self.assertEqual(foo(), foo.py_func())
```

numba/cpython/listobj.py
```diff
@@ -1119,3 +1120,85 @@ def list_to_list(context, builder, fromty, toty, val):
...
+        return
+    msg = ("Cannot __getitem__ on a literal list, return type cannot be "
+           "statically determined.")
+    raise errors.TypingError(msg)
+
+@overload(len)
+def literal_list_len(lst):
+    if not isinstance(lst, types.LiteralList):
+        return
+    llen = len(lst.types)
```
sklam, line 1191: never used. because [constsized_len](https://github.com/numba/numba/blob/e043ee705d65abd3a066500be4d62b54bdf3586c/numba/cpython/builtins.py#L367) takes precedence.

numba/typed/dictobject.py
```diff
@@ -1055,24 +1057,252 @@ def impl_iterator_iternext(context, builder, sig, args, result):
...
+        olditems = cgutils.unpack_tuple(builder, val, len(fromty))
+        items = [context.cast(builder, v, f, t)
+                 for v, f, t in zip(olditems, fromty, toty)]
+        return context.make_tuple(builder, toty, items)
+
+
+@lower_cast(types.DictType, types.DictType)
+def cast_DictType_DictType(context, builder, fromty, toty, val):
+    # should have been picked up by typing
+    return val
```
sklam, line 1308: untested

numba/typed/dictobject.py
```diff
@@ -1055,24 +1057,252 @@ def impl_iterator_iternext(context, builder, sig, args, result):
...
+    tu, tv = sig.args
+    u, v = args
+    pred = tu.literal_value == tv.literal_value
+    res = context.get_constant(types.boolean, pred)
+    return impl_ret_untracked(context, builder, sig.return_type, res)
+
+
+@lower_builtin(operator.ne, types.BaseTuple, types.BaseTuple)
+def tuple_ne(context, builder, sig, args):
+    res = builder.not_(literalstrkeydict_impl_equals(context, builder, sig,
```
sklam, line 1180: this is never used? the other [tuple_ne](https://github.com/numba/numba/blob/e043ee705d65abd3a066500be4d62b54bdf3586c/numba/cpython/tupleobj.py#L76) is taking precedence.

numba/typed/dictobject.py
sklam: how do we need this?

numba/typed/dictobject.py
```diff
@@ -1055,24 +1057,252 @@ def impl_iterator_iternext(context, builder, sig, args, result):
...
+        nbty = types.NamedTuple(unliteral_tys,
+                                dict_type.tuple_ty)
+        values = [x[1] for x in items]
+        # replace with make_tuple call?
+        tup = context.get_constant_undef(nbty)
+        literal_tys = [x for x in dict_type.literal_value.values()]
+
+        # this is to deal with repeated keys
+        value_index = dict_type.value_index
+        if value_index is None:
```
sklam, line 1073: `value_index` is never `None`

numba/core/interpreter.py
```diff
@@ -17,6 +17,13 @@
 from numba.core.byteflow import Flow, AdaptDFA, AdaptCFA
 from numba.core.unsafe import eh
 
+class _UNKNOWN_VALUE(object):
```
sklam, line 20: This class needs a docstring.

Am I right that this class has no other use than showing up for debugging?

numba/core/interpreter.py
```diff
@@ -17,6 +17,13 @@
...
+
+    def __init__(self, varname):
+        self.varname = varname
```
sklam, line 23: perhaps make `varname` a `_varname` so nothing is messing with it.

sklam:
smoking at `numba_smoketest_cpu_49`

numba/typed/dictobject.py
stuartarchibald: think it's left over from trying to make literal dictionary types gracefully degrade, removed in 120ffce5f

numba/typed/dictobject.py
stuartarchibald: removed in 4f508cd53

numba/typed/dictobject.py
stuartarchibald: It's `None` by default on the type ctor, so if you want to do e.g.
```pycon
In [5]: from numba import types                                                                                                 

In [6]: literal = types.literal                                                                                                 

In [7]: ld = types.LiteralStrKeyDict({literal(k):literal(v) for k, v in {'a':1, 'b':2}.items()})                                

In [8]: ld                                                                                                                      
Out[8]: LiteralStrKey[Dict]({Literal[str](a): Literal[int](1), Literal[str](b): Literal[int](2)})

In [9]: ld.value_index 
```
that's fine, it just puts in the 1:1 map.

numba/core/interpreter.py
stuartarchibald: Done in 8602e774b

numba/core/interpreter.py
stuartarchibald: Done in 8602e774b

numba/cpython/listobj.py
```diff
@@ -1119,3 +1120,85 @@ def list_to_list(context, builder, fromty, toty, val):
...
+# Things that mutate literal lists are banned
+@overload_method(types.LiteralList, 'append')
+@overload_method(types.LiteralList, 'extend')
+@overload_method(types.LiteralList, 'insert')
+@overload_method(types.LiteralList, 'remove')
+@overload_method(types.LiteralList, 'pop')
+@overload_method(types.LiteralList, 'clear')
+@overload_method(types.LiteralList, 'sort')
+@overload_method(types.LiteralList, 'reverse')
+def banned_mutators(*args):
```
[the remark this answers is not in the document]
stuartarchibald: good point, I'll try fix them now.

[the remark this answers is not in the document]
stuartarchibald: done in 6f9ee274d

numba/typed/dictobject.py
stuartarchibald: Turns out it is needed: https://dev.azure.com/numba/numba/_build/results?buildId=6009&view=logs&j=12ecc1d6-2964-5850-b419-89f950957853&t=abce8f77-76e0-5291-fb7e-5ca9f390cd04&l=9914 IIRC `typed.Dict` uses the list dtype and not its initial value, but the initial value is part of the type, so casts such as the one failing won't work without it.

stuartarchibald: Reverted 120ffce in ad7343f

numba/cpython/listobj.py
sklam: oops, yes for copy

numba/typed/dictobject.py
sklam: `Cannot cast DictType[int32,int32]<iv={4: 2}> to DictType[int32,int32]<iv={1: 3}>:`

why is that allowed? should it be unified to one without initial_value?

numba/typed/dictobject.py
sklam: oh ok

sklam:
`numba_smoketest_cpu_49` is failing but due to entry-point test (probably a llvmlite version pinning problem again).
I don't see other failures from my random sample

numba/typed/dictobject.py
stuartarchibald: Think this stems from two problems:
1. No attempt at unification in the cast impl https://github.com/numba/numba/blob/e043ee705d65abd3a066500be4d62b54bdf3586c/numba/typed/typedobjectutils.py#L23-L35
2. `DictType` will only unify with another `DictType` if the other is "imprecise" https://github.com/numba/numba/blob/e043ee705d65abd3a066500be4d62b54bdf3586c/numba/core/types/containers.py#L636-L643

PR #5952, opened 2020-07-02
Fix "Internal error ... local variable 'errstr' referenced before assignment during BoundFunction(...)"

jalsti:
In cc673c1 a change had been introduced in `numba/core/types/functions.py:399` which includes two problems that can occur (only) when `config.DEVELOPER_MODE` is active:
* it uses the `errstr` variable before assignment
* it does not use the real function call to `_bt_as_lines(bt)`

This pull request fixes both problems.

### DIFF
--- a/numba/core/types/functions.py
+++ b/numba/core/types/functions.py
@@ -401,8 +401,8 @@ class BoundFunction(Callable, Opaque):
                     else:
                         bt = [""]
                     nd2indent = '\n{}'.format(2 * indent)
-                    errstr += _termcolor.reset(nd2indent +
-                                               nd2indent.join(bt_as_lines))
+                    errstr = _termcolor.reset(nd2indent +
+                                               nd2indent.join(_bt_as_lines(bt)))
                     return _termcolor.reset(errstr)
             else:
                 add_bt = lambda X: ''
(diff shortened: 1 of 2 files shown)


esc:
@jalsti thanks for submitting this, I have queued it for review.

esc:
@jalsti could you potentially add the code and configuration you used to trigger it? Maybe that can form the basis of a test?

jalsti:
Minimal example would be to set developer mode, start ipython:
```
$ export NUMBA_DEVELOPER_MODE=1
$ ipython3
```

and use following snippet, which triggers the problem here (with Numba `0.50.0` and `0.50.1`)

```
import numba 
@numba.jit("int64[:,:,::1](int64[:,:,::1])") 
def trigger_5952(rgb): 
    foo = rgb.max(-1)

```
(Normally the parameter going into the function was here an RGB image as Numpy array, but the above seems enough to trigger the problem.)

stuartarchibald:
Thanks for the example. I've made this into a test here: https://github.com/stuartarchibald/numba/commit/3523ea5a5e698f4280bf8be6bcc6c15b339b796a, I think with this applied, the above patch can be accepted as it demonstrably fixes the issue. Many thanks for reporting this and for supplying a fix!

jalsti:
Thank you very much.

stuartarchibald:
@jalsti no problem, thank you for helping with this! I've applied 3523ea5 to this branch. @esc please could you verify and approve as you see fit, thanks.

stuartarchibald:
Thanks @esc

PR #5958, opened 2020-07-03
Making typed.List a typing Generic

luk-f-a:
This PR adds typing information to the typed List, so in can be used as a Generic in variable annotations, e.g.

```python
from numba.typed import List

my_list:List[int] = List()
my_list.append(1)
my_list.append('a') # would be rejected by mypy

for i in my_list:
    a: str = i # would be rejected by mypy
    print(a)
```

While some of these would also be runtime errors, this PR allows List to be fully checked as a generic by mypy.

This is not related to checking Numba itself with mypy, this just allows users to check their typed.List code with mypy.

Besides setting List as a Generic, it also annotates the main methods (append, getitem, setitem), and the methods for protocols: Iterable[T], Sized, Container[T], and Collection[T].


esc:
@luk-f-a thanks for reporting this, I have added it to the review queue for now.

luk-f-a:
@EPronovost would you have time to review this PR?

numba/typed/typedlist.py
```diff
@@ -317,41 +321,41 @@ def __gt__(self, other):
     def __ge__(self, other):
         return _ge(self, other)
 
-    def append(self, item):
+    def append(self, item: T):
```
EPronovost, line 324: To be fully annotated these should have a `-> None` as well 
https://github.com/python/typeshed/blob/master/stdlib/3/builtins.pyi#L694-L738

numba/typed/typedlist.py
```diff
@@ -317,41 +321,41 @@ def __gt__(self, other):
...
         if not self._typed:
             self._initialise_list(item)
         _insert(self, i, item)
 
-    def count(self, item):
+    def count(self, item: T):
         return _count(self, item)
 
-    def pop(self, i=-1):
+    def pop(self, i=-1) -> T:
```
EPronovost, line 358: Type annotation for `i`?

numba/typed/typedlist.py
```diff
@@ -377,7 +381,7 @@ def reverse(self):
     def copy(self):
         return _copy(self)
 
-    def index(self, item, start=None, stop=None):
+    def index(self, item: T, start=None, stop=None):
```
EPronovost, line 384: Type annotations for `start` and `stop`?

numba/typed/typedlist.py
```diff
@@ -26,6 +26,7 @@
 from numba.typed import listobject
 from numba.core.errors import TypingError, LoweringError
 from numba.core.typing.templates import Signature
+from typing import TypeVar, Generic, Iterator
```
EPronovost, line 29: To stick with style elsewhere should we `import typing as pt` instead?

EPronovost (changes requested):

numba/typed/typedlist.py
luk-f-a: you're right!

numba/typed/typedlist.py
luk-f-a: fixed

numba/typed/typedlist.py
luk-f-a: fixed

numba/typed/typedlist.py
luk-f-a: yes! I created this PR before we decided that, and forgot about it. thanks!

luk-f-a:
@EPronovost thanks for the review. I fixed what you found, plus other similar things, eg other methods returning None which I hadn't annotated. I also completed those annotations that were incomplete, those where I had only annotated one input variable, like getitem, insert, index.

luk-f-a:
Interesting finding from typeguard after I annotated `typed.List`

```
ERROR: test_extend_empty_unrefined (numba.tests.test_typedlist.TestExtend)
  File "/home/vsts/work/1/s/numba/tests/test_typedlist.py", line 779, in test_extend_empty_unrefined
    l.extend(tuple())
  File "/home/vsts/miniconda3/envs/travisci/lib/python3.8/site-packages/typeguard/__init__.py", line 841, in wrapper
    check_return_type(retval, memo)
  File "/home/vsts/miniconda3/envs/travisci/lib/python3.8/site-packages/typeguard/__init__.py", line 665, in check_return_type
    raise exc from None
  File "/home/vsts/miniconda3/envs/travisci/lib/python3.8/site-packages/typeguard/__init__.py", line 663, in check_return_type
    check_type('the return value', retval, memo.type_hints['return'], memo)
  File "/home/vsts/miniconda3/envs/travisci/lib/python3.8/site-packages/typeguard/__init__.py", line 620, in check_type
    raise TypeError(
TypeError: type of the return value must be NoneType; got numba.typed.typedlist.List instead
```

Reproducer (it should always return None but in some cases it's returning self)

```python
from numba.typed import List
  ...: my_list = List([1,2])
  ...: print(my_list.extend(List([1,2])))
  ...: print(my_list.extend(tuple()))
None
[1, 2, 1, 2]
```

numba/typed/typedlist.py
```diff
@@ -377,7 +389,7 @@ def reverse(self):
     def copy(self):
         return _copy(self)
 
-    def index(self, item, start=None, stop=None):
+    def index(self, item: T, start: int = None, stop: int = None) -> int:
```
EPronovost, line 392: Under strict mode inferred optionals are disabled, so these should be `Optional[int]`

numba/typed/typedlist.py
```diff
@@ -317,44 +321,52 @@ def __gt__(self, other):
...
             self._initialise_list(item)
         _setitem(self, i, item)
 
-    def __getitem__(self, i):
+    # noqa F811 comments required due to github.com/PyCQA/pyflakes/issues/592
+    # noqa E704 required to follow overload style of using ... in the same line
+    @pt.overload
+    def __getitem__(self, i: int) -> T: ... # noqa: F811, E704
+    @pt.overload
+    def __getitem__(self, i: slice) -> 'List': ... # noqa: F811, E704
```
EPronovost, line 339: Should this be `List[T]` as it is here? https://github.com/python/typeshed/blob/master/stdlib/3/builtins.pyi#L720

EPronovost (changes requested):
I wonder if we can/should add test coverage for this beyond just using `typeguard` checking.  For a full test of annotations, I think you need both positive and negative examples.  I've recently done similar work for `typeguard` [here](https://github.com/agronholm/typeguard/tree/master/tests/mypy).  To keep it smaller scope we could also just have positive tests; i.e. a test file that uses `List` with annotations and gets analyzed by mypy as part of CI.
I'm curious to hear what the core devs think about this additional testing for type annotations intended to be used by users. cc: @sklam @esc @stuartarchibald 
Note that if we want mypy to consider these type annotations in users' projects we need to add a `py.typed` file https://www.python.org/dev/peps/pep-0561/#packaging-type-information

numba/typed/typedlist.py
luk-f-a: thanks, I didn't realize that strict optionals was the default, I thought that inferred optionals were the default.

luk-f-a: I wonder if we shouldn't consider switching to `--no-strict-optional` to avoid the added verbosity. For example, `index` could fit in one line (easier to read) without losing much in terms of typing (I think `:int = None` is clear enough). what do you think?

```python
    def index(self, item: T, start: pt.Optional[int] = None,
              stop: pt.Optional[int] = None) -> int:
```
vs
```python
def index(self, item: T, start: int = None, stop: int = None) -> int:
```

numba/typed/typedlist.py
luk-f-a: yes, somehow I hadn't thought of it when writing it as text, but it makes sense.

luk-f-a:
@EPronovost I made some fixes following your latest comments, and also improved other things:
- I had only partially annotated `__setitem__`, now it has full annotations with overloads.
- in order to improve legibility by keeping signatures to one line, I created two type aliases.
- I ran mypy on this file and added a few fixes to remove fails
    - added  `# type: ignore[override]`. Without this mypy complains that the method conflicts with the supertype MutableSequence (a problem with the contravariance required by Liskov's substitution principle).
    - in the annotation to `extend` I had to change the input from `Iterable` to `Sequence` because the method calls `len(iterable)` and `iterable[0]`. 

Related to this last point (running mypy), the file now passes basic tests and it could included in "level 3" files. However, I didn't actually change `mypy.ini` because I'd like to do it on your improved version of `mypy.ini` from #6222. What's the best way to do it? Wait for it to be merged or just copy and paste your version here?

numba/typed/typedlist.py
EPronovost: For internal annotations that's something we can decide on.  I think for external facing annotations we should be as precise as possible to avoid incompatibilities for users who don't use `--no-strict-optional`.

luk-f-a:
> Note that if we want mypy to consider these type annotations in users' projects we need to add a py.typed file https://www.python.org/dev/peps/pep-0561/#packaging-type-information

I don't really understand this PEP. Does it mean that without a `py.typed` file someone with a project that imports numba would not be able to make use of the `typed.List`annotation when running mypy on their project?

EPronovost:
Sorry for the delay. Yes to your above question.  There's relevant code in the mypy module finder here: https://github.com/python/mypy/blob/master/mypy/modulefinder.py#L180-L194

luk-f-a:
> Sorry for the delay. Yes to your above question. There's relevant code in the mypy module finder here: https://github.com/python/mypy/blob/master/mypy/modulefinder.py#L180-L194

thanks. I've added the `py.typed` file in the latest commit. I'm still thinking how I could test that it's working correctly. Before I added `py.typed` I ran the following

```python
from numba.typed import List


l:List[int] = List([1, 2])
a=l[0]
l.append(2)
reveal_locals()
reveal_type(l)
```
and the result was correct (even without `py.typed`). So I don't know how to check whether `py.typed` is making a difference now.

numba/typed/typedlist.py
luk-f-a: you're right, I changed it to the explicit `Optional`.

numba/typed/typedlist.py
```diff
@@ -317,44 +324,58 @@ def __gt__(self, other):
...
+    def __iter__(self) -> pt.Iterator[T]:
         for i in range(len(self)):
             yield self[i]
 
-    def __contains__(self, item):
+    def __contains__(self, item: T) -> bool:  # type: ignore[override]
         return _contains(self, item)
 
-    def __delitem__(self, i):
+    def __delitem__(self, i: pt.Union[int, slice]) -> None:
```
sklam, line 364:
```suggestion
    def __delitem__(self, i: Int_or_Slice) -> None:
```

numba/typed/typedlist.py
```diff
@@ -317,44 +324,58 @@ def __gt__(self, other):
...
         if not self._typed:
             self._initialise_list(item)
         _insert(self, i, item)
 
-    def count(self, item):
+    def count(self, item: T) -> int:
         return _count(self, item)
 
-    def pop(self, i=-1):
+    def pop(self, i: int = -1) -> T:
```
sklam, line 375:
```suggestion
    def pop(self, i: "pt.SupportsIndex" = -1) -> T:
```

sklam (commented):
`typeguard` is failing for `numba.tests.test_typedlist`. On CI, the test slices didn't run any of the failing tests by "luck".

numba/typed/typedlist.py
```diff
@@ -26,6 +26,9 @@
 from numba.typed import listobject
 from numba.core.errors import TypingError, LoweringError
 from numba.core.typing.templates import Signature
+import typing as pt
+
+Int_or_Slice = pt.Union[int, slice]
```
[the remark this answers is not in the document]
luk-f-a: Should the same be used in getitem and setitem? From your explanation it sounds like it should

sklam:
Test is failing because `Protocol` is also new on py3.8.

luk-f-a:
> `typeguard` is failing for `numba.tests.test_typedlist`. On CI, the test slices didn't run any of the failing tests by "luck".

what's the solution? Did you run the full suite locally? Is it possible to expand typeguard to run on more slices?

luk-f-a:
> Test is failing because `Protocol` is also new on py3.8.

I tried a fix but made it worse. I'll build something better tomorrow.

sklam:
> what's the solution? Did you run the full suite locally? Is it possible to expand typeguard to run on more slices?

Right, we will need to make more configs use typeguard.

sklam:
If we guard the protocol type for py3.8 only and use a string to refer to it, we can avoid changing the implementation of extend.

something like
```python
if python_version >= (3, 8):
    class _Sequence(pt.Protocol[T_co]): ...

class List:
    def extend(self, iterable: "_Sequence"): ...
```

sklam:
I think it's ok to do type-checks on only py>=3.8. I'm seeing that `Protocol` is essential.

luk-f-a:
Thanks @sklam, I'll try what you suggested above. Funny you found a 5th option, I was considering the pros/cons of:
- keeping extend untyped until 3.8 is the minimum supported version
- changing extend such that it does not require len and getitem (I tried this yesterday and failed) so it can be types as Iterable
- requiring typing_extensions as a dependency (cons: not worth adding a dependency for this)
- typing with _Sequence but not with _Sequence[T], and _Sequence is Any for <3.8

sklam (approved):

PR #5960, opened 2020-07-04
Setting mypy on CI

luk-f-a:
This PR sets a CI run with mypy. To avoid a persistent failure (over 500 errors if left to run freely), the run is limited a few sub-packages from `numba.core` (types, datamodel, rewrites and unsafe). This list can be extended as more modules are annotated.

Since many names in the `typing` package would conflict with Numba's (including `typing` itself), I use aliases prefixed with `mp`, e.g. `mpDict`, `mpType`.

A `mypy_typing` folder is created in `numba.core` for any type definitions that could be needed. I suggest that each sub-package (core, numpy, etc) manages its own type definitions within its folder.

Additionally, two errors from `types` package are solved:

Before
```bash
~/numba$ mypy -p numba.core.types

numba/core/types/abstract.py:24: error: Need type annotation for '_typecache' (hint: "_typecache: Dict[<type>, <type>] = ...")
numba/core/types/abstract.py:408: error: Need type annotation for 'ctor_map' (hint: "ctor_map: Dict[<type>, <type>] = ...")
Found 2 errors in 1 file (checked 10 source files)
```
after 

```bash
~/numba$ mypy -p numba.core.types
Success: no issues found in 10 source files
```

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,11 +1,50 @@
 # Global options:
 
 [mypy]
-warn_return_any = False
 warn_unused_configs = True
-ignore_missing_imports = True
 follow_imports = skip
+show_error_context = True
+files = **/numba/core/types/*.py, **/numba/core/datamodel/*.py, **/numba/core/rewrites/*.py, **/numba/core/unsafe/*.py
 
 # Per-module options:
-[mypy-mycode.core.typing]
+# Level 1 - modules checked on the strictest settings.
+;[mypy-]
+;warn_return_any = True
+;disallow_any_expr = True
+;disallow_any_explicit = True
+;disallow_any_generics = True
+;disallow_subclassing_any = True
+;disallow_untyped_calls = True
+;disallow_untyped_defs = True
+;disallow_incomplete_defs = True
+;check_untyped_defs = True
+;disallow_untyped_decorators = True
+;warn_unused_ignores = True
+;follow_imports = normal
+;warn_unreachable = True
+;strict_equality = True
+
+# Level 2 - module that pass reasonably strict settings.
+#           No untyped functions allowed. Imports must be typed or explicitly ignored.
+;[mypy-]
+;warn_return_any = True
+;disallow_untyped_defs = True
+;disallow_incomplete_defs = True
+;follow_imports = normal
+
+# Level 3 - modules that pass mypy default settings
+[mypy-numba.core.typing.*, numba.core.datamodel.*, numba.core.rewrites.*]
 warn_return_any = True
+
+# Level 4 - modules that do not pass mypy check: they are excluded from "files" setting in global section
+
+# External packages that lack annotations
+[mypy-llvmlite.*]
+ignore_missing_imports = True
+
+[mypy-numpy.*]
+ignore_missing_imports = True
+
+[mypy-winreg.*]
+# this can be removed after Mypy 0.78 is out with the latest typeshed
+ignore_missing_imports = True
--- a/mypy.ini
+++ b/mypy.ini
@@ -7,6 +7,7 @@ show_error_context = True
 files = **/numba/core/types/*.py, **/numba/core/datamodel/*.py, **/numba/core/rewrites/*.py, **/numba/core/unsafe/*.py
 
 # Per-module options:
+# To classify a given module as Level 1, 2 or 3 it must be added both in files (variable above) and in the lists below.
 # Level 1 - modules checked on the strictest settings.
 ;[mypy-]
 ;warn_return_any = True
--- a/mypy.ini
+++ b/mypy.ini
@@ -2,7 +2,7 @@
 
 [mypy]
 warn_unused_configs = True
-follow_imports = skip
+follow_imports = silent
 show_error_context = True
 files = **/numba/core/types/*.py, **/numba/core/datamodel/*.py, **/numba/core/rewrites/*.py, **/numba/core/unsafe/*.py
 
@@ -32,8 +32,9 @@ files = **/numba/core/types/*.py, **/numba/core/datamodel/*.py, **/numba/core/re
 ;disallow_incomplete_defs = True
 ;follow_imports = normal
 
-# Level 3 - modules that pass mypy default settings
-[mypy-numba.core.typing.*, numba.core.datamodel.*, numba.core.rewrites.*]
+# Level 3 - modules that pass mypy default settings (only those in `files` global setting and not in previous levels)
+#           Function/variables are annotated to avoid mypy erros, but annotations are not complete.
+[mypy-numba.core.*]
 warn_return_any = True
 
 # Level 4 - modules that do not pass mypy check: they are excluded from "files" setting in global section
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,11 @@
+# Global options:
+
+[mypy]
+warn_return_any = False
+warn_unused_configs = True
+ignore_missing_imports = True
+follow_imports = skip
+
+# Per-module options:
+[mypy-mycode.core.typing]
+warn_return_any = True
--- a/numba/core/types/abstract.py
+++ b/numba/core/types/abstract.py
@@ -1,4 +1,5 @@
 from abc import ABCMeta, abstractmethod, abstractproperty
+from typing import Dict as mpDict, Type as mpType
 import itertools
 import weakref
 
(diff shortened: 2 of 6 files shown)


mypy.ini
```diff
@@ -0,0 +1,50 @@
+# Global options:
+
+[mypy]
+warn_unused_configs = True
+follow_imports = skip
```
EPronovost, line 5: What are your thoughts between `skip` and `silent` here?  In my work with partially-typed codebases, I've used the later.  Not sure what would be best for levels 3 and 4.
[Link for context](https://mypy.readthedocs.io/en/stable/running_mypy.html#following-imports)

luk-f-a: I think you're right, it's better to change it to silent.

numba/core/types/abstract.py
```diff
@@ -1,4 +1,5 @@
 from abc import ABCMeta, abstractmethod, abstractproperty
+from typing import Dict as mpDict, Type as mpType
```
eric-wieser, line 2: `typing` isn't `mypy`, nor is `mypy` the only package that understands `typing`. `tDict` and `tType` would probably be better names.

luk-f-a: Thanks for looking at the PR @eric-wieser 
you're right, it's a choice I wasn't completely happy with and I was expecting that someone could mention it. Normally I would have used the word "typing", since as you say `mypy` is not `typing` nor the only static checker. The problem is that `typing` already has a meaning within Numba.
Your suggestion of using `t` as a prefix instead of `mp` makes sense, but I wonder if it could be confused with `typed.Dict`. 
I'll think about it a bit more, wait to see if anyone has another idea, and if nothing better comes up, I'll switch to `tDict` and `tType`.

eric-wieser: Could use `b` for builtin or `p` for python also.

Or perhaps `import typing as pt`, then use `pt.List` etc.

luk-f-a: I think `pt` would work well, thanks for the suggestion. I used it as a prefix though, what do you think? ie `ptDict` and `ptType` (latest commits has the updated names).
I'm also adding a note in the developer manual (`contributing` section), explaining this convention.

sklam:
This PR is a good start for typing annotation uses in Numba. TBH, I have a lot to learn to understand how to use type annotations successfully. 

So to tryout, I turned on type checking on lowering.py to see what it can do. I don't understand why it is complaining about:

```bash
$ mypy  numba/core/lowering.py
numba/core/lowering.py:9: error: Module 'numba' has no attribute '_dynfunc'
Found 1 error in 1 file (checked 1 source file)
```

Secondly, I find `lower_extensions : pt.Dict[type, pt.Callable[[pt.Type["Lower"], ir.Inst], None]] = {}` very bad on readability.

Any thoughts?

<details>
<summary>full diff</summary>

```diff
diff --git a/mypy.ini b/mypy.ini
index 0b2ffdfa3..20ea81c85 100644
--- a/mypy.ini
+++ b/mypy.ini
@@ -4,7 +4,12 @@
 warn_unused_configs = True
 follow_imports = silent
 show_error_context = True
-files = **/numba/core/types/*.py, **/numba/core/datamodel/*.py, **/numba/core/rewrites/*.py, **/numba/core/unsafe/*.py
+files =
+    **/numba/core/types/*.py,
+    **/numba/core/datamodel/*.py,
+    **/numba/core/rewrites/*.py,
+    **/numba/core/unsafe/*.py,
+    **/numba/core/lowering.py,
 
 # Per-module options:
 # To classify a given module as Level 1, 2 or 3 it must be added both in files (variable above) and in the lists below.
@@ -50,3 +55,6 @@ ignore_missing_imports = True
 [mypy-winreg.*]
 # this can be removed after Mypy 0.78 is out with the latest typeshed
 ignore_missing_imports = True
+
+[mypy-pygments.*]
+ignore_missing_imports = True
\ No newline at end of file
diff --git a/numba/core/lowering.py b/numba/core/lowering.py
index a8da51e71..342ec2513 100644
--- a/numba/core/lowering.py
+++ b/numba/core/lowering.py
@@ -1,4 +1,5 @@
 import weakref
+import typing as pt
 from collections import namedtuple
 import operator
 from functools import partial
@@ -21,7 +22,7 @@ class Environment(_dynfunc.Environment):
     __slots__ = ('env_name', '__weakref__')
     # A weak-value dictionary to store live environment with env_name as the
     # key.
-    _memo = weakref.WeakValueDictionary()
+    _memo : weakref.WeakValueDictionary[str, Environment] = weakref.WeakValueDictionary()
 
     @classmethod
     def from_fndesc(cls, fndesc):
@@ -311,7 +312,7 @@ class BaseLower(object):
 
 
 # Dictionary mapping instruction class to its lowering function.
-lower_extensions = {}
+lower_extensions : pt.Dict[type, pt.Callable[[pt.Type["Lower"], ir.Inst], None]] = {}
 
 
 class Lower(BaseLower):
```
</details>

EPronovost:
> So to tryout, I turned on type checking on lowering.py to see what it can do. I don't understand why it is complaining about:
> 
> ```shell
> $ mypy  numba/core/lowering.py
> numba/core/lowering.py:9: error: Module 'numba' has no attribute '_dynfunc'
> Found 1 error in 1 file (checked 1 source file)
> ```

@sklam My guess would be that mypy doesn't support importing from C.  I think you could either add a `# type: ignore[import]` on that line or add a [stub file](https://mypy.readthedocs.io/en/stable/stubs.html) for the C module.

> Secondly, I find lower_extensions : pt.Dict[type, pt.Callable[[pt.Type["Lower"], ir.Inst], None]] = {} very bad on readability.

If the value type of that dict is a common pattern, you can use a [type alias](https://docs.python.org/3/library/typing.html#type-aliases) for it.

luk-f-a:
@sklam cool that you took it for a spin.

Regarding the readability issue, I've found that when the annotation gets hard to read, it's usually a good time to further define the internal API with a type alias (as Ethan mentioned). The combination of the type alias and a one liner explaining it is a great help to anyone trying to understand the code.
In this case, the Callable could be given an alias

```python
# LoweringFunctions take a Lower instance and an instruction and do XXXXX
LoweringFunction = pt.Callable[["Lower", ir.Inst], None]

# Dictionary mapping instruction class to its lowering function.
lower_extensions: pt.Dict[type, LoweringFunction] = {}
```

I used `"Lower"` instead of `pt.Type["Lower"]`, because if I'm reading this code right
```python
            for _class, func in lower_extensions.items():
                if isinstance(inst, _class):
                    func(self, inst)
```

then the first argument of the function is an instance of `Lower`, and not a subclass.

eric-wieser:
Isn't the type there `Dict[Type[T], Callable[[T, Instr], None]`?

eric-wieser:
Actually even that isn't right, you need a dependently typed dict where the type of each value depends on it's key.

luk-f-a:
I was thinking the same, and almost wrote it like that. But I don't think there's a way to express that, ie a TypeVar whose value differs in every entry (as opposed to being the same for the whole dictionary).

EPronovost:
Should we add `mypy` to the pre-commit hooks as well? https://github.com/pre-commit/mirrors-mypy

luk-f-a:
> Should we add mypy to the pre-commit hooks as well?

I don't use pre-commit hooks, so I don't have an opinion about it.

sklam:
We can worry about pre-commit hooks in a different PR. I think this PR is ready.

luk-f-a:
@eric-wieser @sklam 

I think I found a way to type the `lower_extensions`  dictionary accurately. It would not be very readable, it would require a comment to explain what's going on, but if it was important to get mypy to check it accurately, this works

```python
from typing import Type, TypeVar, Callable

class A:
    pass

class B(A):
    pass

T = TypeVar('T')

class SpecialDict(dict):
    def __new__(cls, *args, **kwargs):
        return dict(*args, **kwargs)

    def __setitem__(self, key: Type[T], value: Callable[[T, int], None]):
        return super().__setitem__(key, value)

foo = SpecialDict()

def myfuncA(x: A, y: int):
    pass

def myfuncB(x: B, y: int):
    pass

foo[A] = myfuncB # mypy error: error: Incompatible types in assignment (expression has type "Callable[[B, int], Any]", target has type "Callable[[A, int], None]")
foo[A] = myfuncA # OK
foo[B] = myfuncA # OK
foo[B] = myfuncB # OK

```

Issue #5397, opened 2020-03-17
How to change structure of a custom object in place?


densmirn:
- [ ] I am using the latest released version of Numba (most recent is visible in
 the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [X] I have included below a minimal working reproducer (if you are unsure how
 to write one see http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports).

Example:
```python
import numba
from numba import cgutils, types
from numba.extending import (box, intrinsic, make_attribute_wrapper, models,
                             overload_method, overload, register_model)
from numba.targets.boxing import box_list
from numba.typing import signature


class Dummy:
    def __init__(self, data):
        self.data = data

    def insert(self, pos, value):
        self.data = self.data[:pos] + [value] + self.data[pos:]


class DummyType(types.Type):
    def __init__(self, data):
        self.data = data
        super().__init__(name=f'Dummy({self.data})')


@register_model(DummyType)
class DummyTypeModel(models.StructModel):
    def __init__(self, dmm, fe_type):
        members = [
            ('data', fe_type.data),
        ]
        super().__init__(dmm, fe_type, members)


make_attribute_wrapper(DummyType, 'data', '_data')


@intrinsic
def init_dummy(typingctx, data):
    def codegen(context, builder, sig, args):
        typ = sig.return_type
        data, = args
        dummy = cgutils.create_struct_proxy(typ)(context, builder)
        dummy.data = data

        return dummy._getvalue()

    ret_typ = DummyType(data)
    sig = signature(ret_typ, data)

    return sig, codegen


@overload(Dummy)
def dummy_overload(data):
    def ctor(data):
        return init_dummy(data)

    return ctor


@box(DummyType)
def box_interval(typ, val, c):
    dummy = cgutils.create_struct_proxy(typ)(c.context, c.builder, value=val)
    data_obj = box_list(typ.data, dummy.data, c)
    class_obj = c.pyapi.unserialize(c.pyapi.serialize_object(Dummy))
    res = c.pyapi.call_function_objargs(class_obj, (data_obj,))
    c.pyapi.decref(data_obj)
    c.pyapi.decref(class_obj)

    return res


@overload_method(DummyType, 'insert')
def insert_overload(self, pos, value):
    def impl(self, pos, value):
        self = Dummy(self._data[:pos] + [value] + self._data[pos:])
        return self

    return impl


def test_impl(data, pos, value):
    d = Dummy(data)
[...]

PR #5978, opened 2020-07-10
Add structref as a mutable struct that is pass-by-ref

sklam:
Closes https://github.com/numba/numba/issues/5397

Add `structref` to support mutable structure that can be extended like typical custom types (i.e. with `@overload` and friends). Also, support some nice default to minimizing work; i.e. default boxer/unboxer, default getter/setter, default constructor.

TODOs:

- add docs. to be included in a different PR.

### DIFF
--- a/docs/source/developer/repomap.rst
+++ b/docs/source/developer/repomap.rst
@@ -82,7 +82,7 @@ These define aspects of the public Numba interface.
   regular functions on the CPU
 - :ghfile:`numba/core/extending.py` - Public decorators for extending Numba
   (``overload``, ``intrinsic``, etc)
-  - :ghfile:`numba/core/structref.py` - Public API for defining a mutable struct
+  - :ghfile:`numba/experimental/structref.py` - Public API for defining a mutable struct
 - :ghfile:`numba/core/ccallback.py` - ``@cfunc`` decorator for compiling
   functions to a fixed C signature.  Used to make callbacks.
 - :ghfile:`numba/np/ufunc/decorators.py` - ufunc/gufunc compilation
--- a/docs/source/developer/repomap.rst
+++ b/docs/source/developer/repomap.rst
@@ -82,6 +82,7 @@ These define aspects of the public Numba interface.
   regular functions on the CPU
 - :ghfile:`numba/core/extending.py` - Public decorators for extending Numba
   (``overload``, ``intrinsic``, etc)
+  - :ghfile:`numba/core/structref.py` - Public API for defining mutable struct
 - :ghfile:`numba/core/ccallback.py` - ``@cfunc`` decorator for compiling
   functions to a fixed C signature.  Used to make callbacks.
 - :ghfile:`numba/np/ufunc/decorators.py` - ufunc/gufunc compilation
--- a/docs/source/developer/repomap.rst
+++ b/docs/source/developer/repomap.rst
@@ -82,7 +82,7 @@ These define aspects of the public Numba interface.
   regular functions on the CPU
 - :ghfile:`numba/core/extending.py` - Public decorators for extending Numba
   (``overload``, ``intrinsic``, etc)
-  - :ghfile:`numba/core/structref.py` - Public API for defining mutable struct
+  - :ghfile:`numba/core/structref.py` - Public API for defining a mutable struct
 - :ghfile:`numba/core/ccallback.py` - ``@cfunc`` decorator for compiling
   functions to a fixed C signature.  Used to make callbacks.
 - :ghfile:`numba/np/ufunc/decorators.py` - ufunc/gufunc compilation
--- a/numba/core/structref.py
+++ b/numba/core/structref.py
@@ -95,14 +95,17 @@ def define_attributes(struct_typeclass):
         [instance, val] = args
         utils = _Utils(context, builder, inst_type)
         dataval = utils.get_data_struct(instance)
+        # cast val to the correct type
+        field_type = inst_type.field_dict[attr]
+        casted = context.cast(builder, val, val_type, field_type)
         # read old
-        old_value = getattr(dataval, attr, val)
+        old_value = getattr(dataval, attr)
         # incref new value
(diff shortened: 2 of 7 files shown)


numba/tests/test_struct_ref.py
```diff
@@ -0,0 +1,273 @@
...
+
+
+# Call to define_proxy is needed to register the use of `MyStruct` as a
+# PyObject proxy for creating a Numba-allocated structref.
+# The `MyStruct` class and then be used in both jit-code and interpreted-code.
+structref.define_proxy(
+    MyStruct,
+    MyStructType,
+    ['values', 'counter'],
+)
```
stuartarchibald, line 72: Is there benefit in making this into a factory? Also wonder if the getters and setters like those here https://github.com/numba/numba/pull/5978/files#diff-9b2fbc1beb945aef47f6c71220bcd54dR101-R113 could be autogenerated in such a case, perhaps into a base class from which the other's can inherit and later be proxied? Perhaps not for now :)!

numba/tests/test_struct_ref.py
```diff
@@ -0,0 +1,273 @@
...
+        self.assertEqual(td['a'].values, 1)
+        self.assertEqual(td['a'].counter, 2.3)
+        # overwrite
+        td['a'] = MyStruct(2, 3.3)
+        self.assertEqual(td['a'].values, 2)
+        self.assertEqual(td['a'].counter, 3.3)
+        # mutate
+        td['a'].values += 10
+        self.assertEqual(td['a'].values, 12)    # changed
+        self.assertEqual(td['a'].counter, 3.3)  # unchanged
```
stuartarchibald, line 185: Perhaps also check a `td['b'] = MyStruct(4, 5.6)` works, as I think it should? Makes me wonder, should a `MyStruct(int8(4), 5.6)` be castable, I'd guess not at the typing layer?

numba/tests/test_struct_ref.py
```diff
@@ -0,0 +1,273 @@
...
+
+@overload_method(MyStructType, "testme")
+def _(self, arg):
+    def impl(self, arg):
+        return self.values * arg + self.counter
+    return impl
+
+
+class TestStructRefExtending(MemoryLeakMixin, TestCase):
+    def test_overload_method(self):
```
stuartarchibald, line 211: RE overloads, is attribute overloading likely to ever be a thing that makes sense in the `structref` context?

docs/source/developer/repomap.rst
```diff
@@ -82,6 +82,7 @@ These define aspects of the public Numba interface.
   regular functions on the CPU
 - :ghfile:`numba/core/extending.py` - Public decorators for extending Numba
   (``overload``, ``intrinsic``, etc)
+  - :ghfile:`numba/core/structref.py` - Public API for defining mutable struct
```
stuartarchibald, line 85:
```suggestion
  - :ghfile:`numba/core/structref.py` - Public API for defining a mutable struct
```

numba/core/structref.py
```diff
@@ -0,0 +1,365 @@
...
+        field_type = inst_type.field_dict[attr]
+        casted = context.cast(builder, val, val_type, field_type)
+        # read old
+        old_value = getattr(dataval, attr)
+        # incref new value
+        context.nrt.incref(builder, val_type, casted)
+        # decref old value (must be last in case new value is old value)
+        context.nrt.decref(builder, val_type, old_value)
+        # write new
+        setattr(dataval, attr, casted)
```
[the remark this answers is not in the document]
sklam: I thought that would happen too. But, it turns out that the typer for `setattr` is forcing the RHS to match the attribute type.

numba/core/structref.py
```diff
@@ -0,0 +1,365 @@
+"""Utilities for defining a mutable struct.
```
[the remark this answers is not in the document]
sklam: Good point. Better be in experimental!

numba/tests/test_struct_ref.py
sklam: I was trying to autogenerate the getters/setters via a base class. Just too much work for now.

numba/tests/test_struct_ref.py
sklam: I guess to emulate `@property` without an actual field backing it. I can add a test quickly

numba/tests/test_struct_ref.py
sklam: It's not casting the fields at the moment. That would be another potential enhancement

docs/source/developer/repomap.rst
```diff
@@ -82,7 +82,7 @@ These define aspects of the public Numba interface.
   regular functions on the CPU
 - :ghfile:`numba/core/extending.py` - Public decorators for extending Numba
   (``overload``, ``intrinsic``, etc)
-  - :ghfile:`numba/core/structref.py` - Public API for defining mutable struct
+  - :ghfile:`numba/core/structref.py` - Public API for defining a mutable struct
```
stuartarchibald, line 85:
```suggestion
  - :ghfile:`numba/experimental/structref.py` - Public API for defining a mutable struct
```
should fix the doc build.

numba/core/types/containers.py
```diff
@@ -722,18 +722,28 @@ def __init__(self, fields):
...
             name, and `type` is a numba type for the field type.
         """
-        self._fields = tuple(map(tuple, fields))
+        def check_field_pair(fieldpair):
+            name, typ = fieldpair
+            if not isinstance(name, str):
+                msg = "expecting a str for field name"
+                raise ValueError(msg)
+            if not isinstance(typ, Type):
+                msg = f"expecting a Numba Type for field type"
```
stuartarchibald, line 735:
```suggestion
                msg = "expecting a Numba Type for field type"
```

stuartarchibald:
Thanks for the fixes, I think the only remaining this is the `f-string` problem here https://github.com/numba/numba/pull/5978#discussion_r455201605 else this looks good to go.

PR #5981, opened 2020-07-13
Document bounds-checking behavior in python deviations page

### LINKED ISSUE #5970: Add note to docs about boundschecking being off by default and how to switch it on
## Reporting a bug

```python
import numba as nb
import numpy as np

def foo():
    return np.zeros(5)[6]

foo()  # raises IndexError
nb.jit(foo)()   # returns something
nb.njit(foo)()  # returns something

```

I would have expected compiled code to also raise an exception? 
If I am missing some subtlety in numba behavior, perhaps needs to be added [here](https://numba.pydata.org/numba-doc/dev/reference/pysemantics.html#) ?

Took me a while to figure out an off-by-one error.

ludaavics:
closes #5970

### DIFF
--- a/docs/source/reference/pysemantics.rst
+++ b/docs/source/reference/pysemantics.rst
@@ -3,6 +3,18 @@
 Deviations from Python Semantics
 ================================
 
+Bounds Checking
+---------------
+
+By default, instead of causing an :class:`IndexError`, accessing an
+out-of-bound index of an array in a Numba-compiled function will return
+garbage results or cause a segfault. Bounds checking can be enabled on a
+specific function via the :ref:`boundscheck <jit-decorator-boundscheck>`
+option of the jit decorator. Additionally, the `NUMBA_BOUNDSCHECK`` can be set
+to 0 or 1 to globally override this flag.
+
+Bounds checking will slow down typical functions so it is recommended to only
+use this flag for debugging.
 
 Exceptions and Memory Allocation
 --------------------------------


stuartarchibald:
Thanks for the PR, I've queued it for review.

docs/source/reference/pysemantics.rst
```diff
@@ -3,6 +3,18 @@
 Deviations from Python Semantics
 ================================
 
+Bounds Checking
+---------------
+
+By default, instead of causing an :class:`IndexError`, accessing an
+out-of-bound index of an array in a Numba-compiled function will return
+garbage results or cause a segfault. Bounds checking can be enabled on a
```
stuartarchibald, line 11:
```suggestion
out-of-bound index of an array in a Numba-compiled function will return
invalid values or lead to an access violation error (it's reading from invalid memory locations). Bounds checking can be enabled on a
```
(with line wrap)

docs/source/reference/pysemantics.rst
```diff
@@ -3,6 +3,18 @@
...
+specific function via the :ref:`boundscheck <jit-decorator-boundscheck>`
+option of the jit decorator. Additionally, the `NUMBA_BOUNDSCHECK` can be set
+to 0 or 1 to globally override this flag.
```
stuartarchibald, line 14:
```suggestion
option of the jit decorator. Additionally, the :envvar:`NUMBA_BOUNDSCHECK` can be set
to 0 or 1 to globally override this flag.
```

docs/source/reference/pysemantics.rst
```diff
@@ -3,6 +3,18 @@
...
+
+Bounds checking will slow down typical functions so it is recommended to only
+use this flag for debugging.
```
stuartarchibald, line 17:
```suggestion
.. note:: Bounds checking will slow down typical functions so it is recommended to only
          use this flag for debugging purposes.
```

ludaavics:
Numba has saved me so much time I only wish I had the skill-set to do more:)
I've manually added your suggestions.

stuartarchibald:
> Numba has saved me so much time I only wish I had the skill-set to do more:)
> I've manually added your suggestions.

I'm glad to hear Numba is useful to you. Thanks for making this patch and putting in those suggestions. Congratulations on your first contribution to Numba!

If you'd like to work on other things in Numba there's a load of [`good first issues`](https://github.com/numba/numba/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22) labelled in the issue tracker. Thanks again!

PR #5984, opened 2020-07-14
Handle idx and size unification in wrap_index manually.

### LINKED ISSUE #5442: prange working with numpy?

DrTodd13:
Resolves #5442

wrap_index has to compare idx and size and to do that they need to be the same bitwidth. If one is signed and the other unsigned then the old code to unify to the same bitwidth (based on numpy rules) would also switch the type to double. In this PR, we keep the incoming integer types but just extend the bitwidth to be the max of the incoming types. We handle signedness extension correctly as well.


DrTodd13:
@esc Recreation of previous PR with some git history issues.

stuartarchibald:
@DrTodd13 from OOB discussion, am I correct in recalling that this PR is the replacement for https://github.com/numba/numba/pull/6779?

DrTodd13:
> @DrTodd13 from OOB discussion, am I correct in recalling that this PR is the replacement for #6779?



> @DrTodd13 from OOB discussion, am I correct in recalling that this PR is the replacement for #6779?

@stuartarchibald Yes, that is right.

numba/parfors/array_analysis.py
```diff
@@ -89,22 +89,34 @@ def wrap_index(typingctx, idx, size):
     (idx < 0 ? idx + size : idx) because we may have situations
     where idx > size due to the way indices are calculated
     during slice/range analysis.
+
+    Both idx and size have to be Integer types.
+    size should be from the array size vars that array_analysis
+    adds and the bidwidth should match the platform maximum.
```
stuartarchibald, line 95:
```suggestion
    adds and the bitwidth should match the platform maximum.
```

numba/parfors/array_analysis.py
```diff
@@ -89,22 +89,34 @@ def wrap_index(typingctx, idx, size):
...
+            idx = builder.zext(args[0], ll_idx_unified_ty)
+        if unified_ty.signed:
+            size = builder.sext(args[1], ll_unified_ty)
+        else:
+            size = builder.zext(args[1], ll_unified_ty)
         neg_size = builder.neg(size)
         zero = llvmlite.ir.Constant(ll_unified_ty, 0)
+        # If idx is unsigned then these signed comparisons will fail in those
+        # cases where the idx has the highest bit set, namely more than 2**31
+        # on 32-bit platforms and 2**63 on 64-bit platforms.
```
stuartarchibald, line 119: I thought parfors was 64 bit only?

stuartarchibald (commented):

numba/parfors/array_analysis.py
DrTodd13: Yes, I think so.  Do you want the comment changed?

numba/parfors/array_analysis.py
```diff
@@ -89,22 +89,34 @@ def wrap_index(typingctx, idx, size):
...
-        idx = builder.sext(args[0], ll_unified_ty)
-        size = builder.sext(args[1], ll_unified_ty)
+        if idx_unified.signed:
+            idx = builder.sext(args[0], ll_idx_unified_ty)
+        else:
+            idx = builder.zext(args[0], ll_idx_unified_ty)
+        if unified_ty.signed:
+            size = builder.sext(args[1], ll_unified_ty)
+        else:
+            size = builder.zext(args[1], ll_unified_ty)
```
stuartarchibald, line 114: This branch is not hit in testing. Can it be hit?

DrTodd13: One would think it would be more likely for a size to be unsigned that it is to be signed so if 112 is hit instead of 114 then that is a bit surprising.  I would think that both could happen but I'd have to trace all the callsites in the code to see where the parameters passed as size are coming from.  I don't know off the top of my head how to write a test to force this to happen.

stuartarchibald: hmmm, the test suite suggests the converse, size is always signed (and `int64`)? From inspection this branch is correct, so I'm happy to leave it, but I do wonder if it's actually possible to hit it. Any objections to merging as-is?

DrTodd13: Yes, let's just leave it.  Thanks.

stuartarchibald (approved):
Thanks for the patch!

PR #5986, opened 2020-07-14
flake8:   numba/core/annotations/*.py

esc:
As title.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -44,7 +44,6 @@ exclude =
     numba/core/interpreter.py
     numba/core/caching.py
     numba/core/debuginfo.py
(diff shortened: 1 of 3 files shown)


esc:
There are still some `501` errors left, but they are in a template file and I'm not sure if wrapping lines inside the template is the best idea. There doesn't seem to be a way with `flake8` to ignore besides annotating the line with a comment, which isn't an option inside a triple quoted template either. Not sure what the best practice would be here, perhaps move the offending templates to their own file?

sklam:
I think moving the offending tripled-quoted strings into a new files which flake8 ignores can work.

esc:
I will close this now, I think it's a lost cause.

PR #5996, opened 2020-07-16
Set lower bound for llvmlite to 0.33

stuartarchibald:
0.33 and 0.34 are intended to be binary compatible, the difference
being LLVM versions, 0.33 with LLVM 9, 0.34 with LLVM 10 support.


sklam:
Looks good. `pip install -e .` works fine locally against a llvmlite 0.33.0.

PR #6004, opened 2020-07-17
Fix problem in branch pruning with LiteralStrKeyDict

stuartarchibald:
Branch pruning tries to evaluate literal types via (eventually)
`typeof(literal_value)`, LiteralStrKeyDict's `literal_value`
is a python dict and has no associated `typeof` translation.
This patch mitigates such a query breaking the branch pruning
pass.

### DIFF
--- a/numba/core/analysis.py
+++ b/numba/core/analysis.py
@@ -396,7 +396,12 @@ def dead_branch_prune(func_ir, called_args):
         # literal type, return the type itself so comparisons like `x == None`
         # still work as e.g. x = types.int64 will never be None/NoneType so
         # the branch can still be pruned
-        return getattr(input_arg_ty, 'literal_type', Unknown())
+        try:
+            return getattr(input_arg_ty, 'literal_type', Unknown())
+        except ValueError:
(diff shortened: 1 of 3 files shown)


numba/core/analysis.py
```diff
@@ -396,7 +396,12 @@ def resolve_input_arg_const(input_arg_idx):
         # literal type, return the type itself so comparisons like `x == None`
         # still work as e.g. x = types.int64 will never be None/NoneType so
         # the branch can still be pruned
-        return getattr(input_arg_ty, 'literal_type', Unknown())
+        try:
+            return getattr(input_arg_ty, 'literal_type', Unknown())
+        except ValueError:
```
sklam, line 401: I think we should outside-in this fix and put the try..except in the property for `literal_type` to raise `AttributeError`

stuartarchibald: Good idea. I also have a patch somewhere that makes `typeof(dict)` at least mostly behave, but I've a feeling that opens up a lot of complications for dicts in globals.

PR #6006, opened 2020-07-18
Resolves incorrectly hoisted list in parfor.

### LINKED ISSUE #5979: Incorrect results from (deterministic) prange loop
## Reporting a bug

This seems sufficiently different from https://github.com/numba/numba/issues/5942 that I'm bringing it up as a separate issue.

First, a case where the array doesn't seem to be read correctly:
```
import numba as nb
@nb.njit(parallel=True)
def f():
    for k in nb.prange(5):
        X = [0]
        X[0] = 1
        print(k, X, X[0])
```

Actual:
```
>>> f()
4 [1] 0
0 [1] 0
3 [1] 0
2 [1] 0
1 [1] 0
```
Expected: lines of the form `k [1] 1`

Next a case where the array doesn't seem to be written to correctly.
```
import numba as nb
@nb.njit(parallel=True)
def f(n):
    for k in nb.prange(n):
        X = [0]
        print(k, X)
        for j in range(1):
            X[j] = 1
```
It seems that in each thread the array X doesn't get re-initialized to [0].
<details>

```
>>> f(2*8+1)  # my machine has 8 threads.
0 [0]
2 [0]
3 [1]
8 [0]
9 [1]
14 [0]
4 [0]
1 [1]
12 [0]
15 [1]
6 [0]
5 [1]
10 [0]
13 [1]
7 [1]
11 [1]
16 [1]
>>> 
```
The order of the lines will change but the values do not.
</details>

DrTodd13:
Resolves #5979

Treat setitems in a parfor body as if they were definitions so that a list initialization in a parfor body will not be incorrectly hoisted.

### DIFF
--- a/numba/parfors/parfor_lowering.py
+++ b/numba/parfors/parfor_lowering.py
@@ -738,6 +738,18 @@ def find_setitems_body(setitems, itemsset, loop_body, typemap):
     for label, block in loop_body.items():
         find_setitems_block(setitems, itemsset, block, typemap)
 
+def empty_container_allocator_hoist(inst, dep_on_param, call_table, hoisted,
+                                    not_hoisted, typemap, stored_arrays):
+    if (isinstance(inst, ir.Assign) and
+        isinstance(inst.value, ir.Expr) and
+        inst.value.op == 'call' and
+        inst.value.func.name in call_table):
+        call_list = call_table[inst.value.func.name]
+        if call_list == ['empty', np]:
+            return _hoist_internal(inst, dep_on_param, call_table, hoisted,
+                                   not_hoisted, typemap, stored_arrays)
+    return False
+
 def hoist(parfor_params, loop_body, typemap, wrapped_blocks):
     dep_on_param = copy.copy(parfor_params)
     hoisted = []
@@ -759,7 +771,10 @@ def hoist(parfor_params, loop_body, typemap, wrapped_blocks):
     for label, block in loop_body.items():
         new_block = []
         for inst in block.body:
-            if isinstance(inst, ir.Assign) and inst.target.name in def_once:
+            if empty_container_allocator_hoist(inst, dep_on_param, call_table,
+                                   hoisted, not_hoisted, typemap, itemsset):
+                continue
+            elif isinstance(inst, ir.Assign) and inst.target.name in def_once:
                 if _hoist_internal(inst, dep_on_param, call_table,
                                    hoisted, not_hoisted, typemap, itemsset):
                     # don't add this instruction to the block since it is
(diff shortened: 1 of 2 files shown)


numba/tests/test_parfors.py
```diff
@@ -3668,7 +3683,7 @@ def test_impl():
         self.check(test_impl,)
         cpfunc = self.compile_parallel(test_impl, ())
         diagnostics = cpfunc.metadata['parfor_diagnostics']
-        self.assert_diagnostics(diagnostics, hoisted_allocations=1)
+        self.assert_diagnostics(diagnostics, hoisted_allocations=0)
```
stuartarchibald, line 3686: Why is this now not hoisted? I can guess from the proposed changes to analysis why it's the case but should it be the case?

DrTodd13: Fixed, I hope.

numba/parfors/parfor_lowering.py
```diff
@@ -738,13 +738,25 @@ def find_setitems_body(setitems, itemsset, loop_body, typemap):
...
+                                    not_hoisted, typemap, stored_arrays):
+    if (isinstance(inst, ir.Assign) and
+        isinstance(inst.value, ir.Expr) and
+        inst.value.op == 'call' and
+        inst.value.func.name in call_table):
+        call_list = call_table[inst.value.func.name]
+        if call_list == ['empty', np]:
+            return _hoist_internal(inst, dep_on_param, call_table, hoisted,
+                                   not_hoisted, typemap, stored_arrays)
+    return False
```
stuartarchibald, line 751: Could this use this: https://github.com/numba/numba/blob/master/numba/core/ir_utils.py#L703-L726 ?

DrTodd13: No, I don't think it would be correct to use "is_pure" here because this here is specifically about the allocation of empty containers.

stuartarchibald: ok, I see, thanks, whilst the code is very similar the purpose is different.

numba/tests/test_parfors.py
```diff
@@ -2540,6 +2540,21 @@ def test_impl(image, a, b):
         image = np.zeros((3, 3), dtype=np.int32)
         self.prange_tester(test_impl, image, 0, 0)
 
+    @skip_parfors_unsupported
+    def test_list_setitem_hoisting(self):
+        # issue5979
+        # Don't hoist list initialization if list item set.
```
stuartarchibald, line 2546: Is this comment accurate? It seems like in the above the code is restricting hoisting to `np.empty` as the allocator?

DrTodd13: The code I added to stop list initialization hoisting when inappropriate (line 768 above) had the side-effect of stopping np.empty from being hoisted.  So, I added something specifically back in for just that function to make the existing hoisting tests pass.  So, this comment is accurate and relates to the core change of this PR on line 768.  Everything besides 768-769 is just in support of that change.

DrTodd13: Yes, it is accurate.  This was the main change of the PR which had the side-effect of stopping normal np.empty hoisting.  I had to add something back in specifically for that then.

stuartarchibald (approved):
Thanks for the patch @DrTodd13

numba/tests/test_parfors.py
stuartarchibald: ah, great, thanks for the explanation.

Issue #5994, opened 2020-07-16
Literal sets missing implementations


EPronovost:
## Reporting a bug

It seems using literal sets in jit decorated functions causes errors.

```
@numba.njit
def get_union():
    return {4} | {1}

get_union()
```
Yields
```
TypingError: Failed in nopython mode pipeline (step: nopython frontend)
No implementation of function Function(<built-in function or_>) found for signature:
 
 >>> or_(set(Literal[int](4)), set(Literal[int](1)))
 
There are 6 candidate implementations:
  - Of which 4 did not match due to:
  Overload of function 'or_': File: <numerous>: Line N/A.
    With argument(s): '(set(Literal[int](4)), set(Literal[int](1)))':
   No match.
  - Of which 2 did not match due to:
  Operator Overload in function 'or_': File: unknown: Line unknown.
    With argument(s): '(set(Literal[int](4)), set(Literal[int](1)))':
   No match for registered cases:
    * (bool, bool) -> bool
    * (int64, int64) -> int64
    * (int64, uint64) -> int64
    * (uint64, int64) -> int64
    * (uint64, uint64) -> uint64

During: typing of intrinsic-call at <ipython-input-5-3354ebb3e034> (3)

File "<ipython-input-5-3354ebb3e034>", line 3:
def get_union():
    return {4} | {1}
    ^
```

I tried to find where the candidate implementations are evaluated for this example (in `numba/core/base.py`) without luck.  There's a [`@lower_builtin` here](https://github.com/numba/numba/blob/master/numba/cpython/setobj.py#L1376-L1389) that I would have thought handles this.  One thought I had is that it looks like we're typing these as a set of a literal int, instead of a literal set (i.e. the `dtype`s of the two arguments are `Literal[int](4)` and `Literal[int](1)`, which don't match).

- [X] I have tried using the latest released version of Numba (most recent is
 visible in the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [X] I have included below a minimal working reproducer (if you are unsure how
 to write one see http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports).

stuartarchibald:
Thanks for the report. The first thing Numba has to do is type inference for the function, what you are seeing is type inference rejecting the operation as something that Numba doesn't support. The bug trail starts here with the declaration for supported behaviours for `operator.or_`:
[...]

PR #6010, opened 2020-07-18
Set Literals

EPronovost:
Working to address https://github.com/numba/numba/issues/5994.  This PR adds support for things like

```
@njit 
def foo(a):
    return set(a) | {1, 2}
```

It also reformats the touched files to be flake8 compliant.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -64,7 +64,6 @@ exclude =
     numba/core/types/iterators.py
     numba/core/types/scalars.py
     numba/core/fastmathpass.py
-    numba/cpython/setobj.py
     numba/core/options.py
     numba/cpython/printimpl.py
     numba/cpython/cmathimpl.py
@@ -304,7 +303,6 @@ exclude =
     numba/tests/test_chained_assign.py
     numba/tests/test_withlifting.py
     numba/tests/test_parfors.py
-    numba/tests/test_sets.py
     numba/tests/test_dyn_array.py
     numba/tests/test_objects.py
     numba/tests/test_random.py
@@ -322,7 +320,6 @@ exclude =
     numba/core/typing/listdecl.py
     numba/core/typing/builtins.py
     numba/core/typing/randomdecl.py
-    numba/core/typing/setdecl.py
     numba/core/typing/npydecl.py
     numba/core/typing/arraydecl.py
     numba/core/typing/collections.py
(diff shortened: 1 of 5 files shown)


sklam:
I think the fix can be simplified to making `Set` type always use non-literal type unless there is a use-case that we need to carry literal information in the type.

To force non-literal type in `Set`:
```diff
diff --git a/numba/core/types/containers.py b/numba/core/types/containers.py
index 37a2573ac..a8c363e02 100644
--- a/numba/core/types/containers.py
+++ b/numba/core/types/containers.py
@@ -506,7 +506,7 @@ class Set(Container):

     def __init__(self, dtype, reflected=False):
         assert isinstance(dtype, (Hashable, Undefined))
-        self.dtype = dtype
+        self.dtype = unliteral(dtype)
         self.reflected = reflected
         cls_name = "reflected set" if reflected else "set"
         name = "%s(%s)" % (cls_name, self.dtype)
```

stuartarchibald:
I think as a result of switching type inference to preferring unliteral types and with the addition of  the patch in https://github.com/numba/numba/pull/6010#issuecomment-661192184 both the originally reported #5994 and the example above in https://github.com/numba/numba/pull/6010#issue-452120062 will work.

stuartarchibald:
> I think the fix can be simplified to making `Set` type always use non-literal type unless there is a use-case that we need to carry literal information in the type.
> 
> To force non-literal type in `Set`:
> 
> ```diff
> diff --git a/numba/core/types/containers.py b/numba/core/types/containers.py
> index 37a2573ac..a8c363e02 100644
> --- a/numba/core/types/containers.py
> +++ b/numba/core/types/containers.py
> @@ -506,7 +506,7 @@ class Set(Container):
> 
>      def __init__(self, dtype, reflected=False):
>          assert isinstance(dtype, (Hashable, Undefined))
> -        self.dtype = dtype
> +        self.dtype = unliteral(dtype)
>          self.reflected = reflected
>          cls_name = "reflected set" if reflected else "set"
>          name = "%s(%s)" % (cls_name, self.dtype)
> ```

I can't think of a use case where it would be necessary to carry the literal information in the type, I think it goes down the route of over-specialisation again? Suggest we add this patch and corresponding bug reports/examples as tests.

PR #6015, opened 2020-07-20
Adding BooleanLiteral type

kozlov-alexey:
This adds new type for representing literal Boolean values supporting
common operators between literal and non-literal arguments.

### DIFF
--- a/numba/core/typing/builtins.py
+++ b/numba/core/typing/builtins.py
@@ -445,12 +445,8 @@ class ConstOpEq(AbstractTemplate):
 
 
 @infer_global(operator.ne)
-class ConstOpNotEq(AbstractTemplate):
-    def generic(self, args, kws):
-        assert not kws
-        (arg1, arg2) = args
-        if isinstance(arg1, types.Literal) and isinstance(arg2, types.Literal):
-            return signature(types.boolean, arg1, arg2)
+class ConstOpNotEq(ConstOpEq):
+    pass
 
 
 @infer_global(operator.eq)
--- a/numba/core/typing/builtins.py
+++ b/numba/core/typing/builtins.py
@@ -434,11 +434,7 @@ class CmpOpGe(OrderedCmpOp):
     pass
 
 
-@infer_global(operator.eq)
-class CmpOpEq(UnorderedCmpOp):
-    pass
-
-
+# more specific overloads should be registered first
 @infer_global(operator.eq)
 class ConstOpEq(AbstractTemplate):
     def generic(self, args, kws):
@@ -449,7 +445,16 @@ class ConstOpEq(AbstractTemplate):
 
 
 @infer_global(operator.ne)
-class ConstOpNotEq(ConstOpEq):
+class ConstOpNotEq(AbstractTemplate):
+    def generic(self, args, kws):
+        assert not kws
+        (arg1, arg2) = args
+        if isinstance(arg1, types.Literal) and isinstance(arg2, types.Literal):
+            return signature(types.boolean, arg1, arg2)
+
+
+@infer_global(operator.eq)
+class CmpOpEq(UnorderedCmpOp):
     pass
 
 
--- a/numba/cpython/builtins.py
+++ b/numba/cpython/builtins.py
@@ -103,6 +103,7 @@ def const_eq_impl(context, builder, sig, args):
     return impl_ret_untracked(context, builder, sig.return_type, res)
 
 
+# keep types.IntegerLiteral, as otherwise there's ambiguity between this and int_ne_impl
 @lower_builtin(operator.ne, types.Literal, types.Literal)
 @lower_builtin(operator.ne, types.IntegerLiteral, types.IntegerLiteral)
 def const_ne_impl(context, builder, sig, args):
--- a/numba/cpython/builtins.py
+++ b/numba/cpython/builtins.py
@@ -75,6 +75,11 @@ def opaque_is(context, builder, sig, args):
 
 @lower_builtin(operator.is_, types.Boolean, types.Boolean)
 def bool_is_impl(context, builder, sig, args):
+    """
+    Implementation for `x is y` for types derived from types.Boolean
+    (e.g. BooleanLiteral), and cross-checks between literal and non-literal
+    booleans, to satisfy Python's behavior preserving identity for bools.
+    """
     arg1, arg2 = args
     arg1_type, arg2_type = sig.args
     _arg1 = context.cast(builder, arg1, arg1_type, types.boolean)
--- a/numba/cpython/builtins.py
+++ b/numba/cpython/builtins.py
@@ -73,6 +73,20 @@ def opaque_is(context, builder, sig, args):
         return cgutils.false_bit
 
 
+@lower_builtin(operator.is_, types.Boolean, types.Boolean)
+def bool_is_impl(context, builder, sig, args):
+    arg1, arg2 = args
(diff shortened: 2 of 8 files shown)


stuartarchibald:
Thanks for the patch @kozlov-alexey, I've queued it for review.

numba/core/typing/builtins.py
```diff
@@ -449,7 +445,16 @@ def generic(self, args, kws):
 
 
 @infer_global(operator.ne)
-class ConstOpNotEq(ConstOpEq):
+class ConstOpNotEq(AbstractTemplate):
+    def generic(self, args, kws):
+        assert not kws
+        (arg1, arg2) = args
+        if isinstance(arg1, types.Literal) and isinstance(arg2, types.Literal):
+            return signature(types.boolean, arg1, arg2)
```
eric-wieser, line 453: Why doesn't subclassing `ConstOpEq` as was done before work? Seems silly to repeat this function body at any rate.

numba/cpython/builtins.py
```diff
@@ -73,6 +73,20 @@ def opaque_is(context, builder, sig, args):
...
+def bool_is_impl(context, builder, sig, args):
+    arg1, arg2 = args
+    arg1_type, arg2_type = sig.args
+    _arg1 = context.cast(builder, arg1, arg1_type, types.boolean)
+    _arg2 = context.cast(builder, arg2, arg2_type, types.boolean)
+    eq_impl = context.get_function(
+        operator.eq,
+        typing.signature(types.boolean, types.boolean, types.boolean)
+    )
+    return eq_impl(builder, (_arg1, _arg2))
```
eric-wieser, line 86: Why is this necessary? Doesn't `generic_is` above handle it?

kozlov-alexey: @eric-wieser This is to support cases when one arg is BooleanLiteral and other is types.boolean. In this case `generic_is` always returns False since `lhs_type == rhs_type` check is not met.

numba/core/typing/builtins.py
kozlov-alexey: @eric-wieser Yep, thanks. I'll revert this change.

numba/cpython/builtins.py
eric-wieser: Two points would be good to include in comments here:
* This covers both Boolean and BooleanLiteral types
* This is needed for books because bools in python are guaranteed to preserve identity. It's not needed for any other literal types.

numba/parfors/parfor_lowering.py
```diff
@@ -1621,16 +1621,17 @@ def load_potential_tuple_var(x):
                     array_strides.append(strides[j])
                 builder.store(builder.bitcast(ary.data, byte_ptr_t), dst)
         else:
+            aty_is_bool = aty is types.boolean or isinstance(aty, types.BooleanLiteral)
```
eric-wieser, line 1624: Shorter and almost as fast as:
```suggestion
            aty_is_bool = isinstance(aty, types.Boolean)
```

kozlov-alexey: @eric-wieser Yes, I thought about this, but that way it would be less restrictive (any other type that user will subclass from types.Boolean will match and that type might actually use different data model) and with original version we restrict to types actually using BooleanModel that defines 1b width.

eric-wieser: If you're worried about that, then you should worry about literal subclasses too:

```suggestion
            # do not allow any other subclasses
            aty_is_bool = type(aty) in (types.Boolean, types.BooleanLiteral)
```

numba/cpython/builtins.py
```diff
@@ -84,7 +103,8 @@ def const_eq_impl(context, builder, sig, args):
     return impl_ret_untracked(context, builder, sig.return_type, res)
 
 
-@lower_builtin(operator.ne, types.StringLiteral, types.StringLiteral)
```
stuartarchibald, line 87: Why is this deleted?

numba/parfors/parfor_lowering.py
```diff
@@ -1621,16 +1621,18 @@ def load_potential_tuple_var(x):
                     array_strides.append(strides[j])
                 builder.store(builder.bitcast(ary.data, byte_ptr_t), dst)
         else:
+            # do not allow any other subclasses
+            aty_is_bool = type(aty) in (types.Boolean, types.BooleanLiteral)
```
stuartarchibald, line 1625: `types.BooleanLiteral` inherits from `types.Boolean`, could this just be an `isinstance` check like `aty_is_bool = isinstance(aty, types.Boolean)` ?

stuartarchibald: I note this is discussed here https://github.com/numba/numba/pull/6015#discussion_r460274711, the current restrictive check does not follow what has been used elsewhere in Numba (or parfors), are there known use cases where this is a problem? Also Numba types have an `is_internal` property to determine if a type is part of the Numba type system or a user defined type, would suggest using that if restrictive behaviour is required.

CC @DrTodd13 any view on the above?

numba/cpython/builtins.py
kozlov-alexey: This is only to have the same specifications for `eq `and `ne`. Note it's not removed but changed to:
`@lower_builtin(operator.ne, types.Literal, types.Literal)`
which will work for StringLiteral types as well.

stuartarchibald: I see, thanks, guess this https://github.com/numba/numba/pull/6015/files#diff-87f8c6b1645b5d35540368237bcc89baR107 is due to https://github.com/numba/numba/pull/6015/files#diff-87f8c6b1645b5d35540368237bcc89baR94? Should probably add a similar note?

numba/parfors/parfor_lowering.py
kozlov-alexey: @stuartarchibald Yes, thank you, using the following definition works too.
`aty_is_bool = isinstance(aty, types.Boolean) and aty.is_internal`

I'm not aware of any issues related to user sub-classing from types.Boolean and affecting parfor lowering, and I agree that changing it in one place only is probably not a good idea. So I can leave it as simple  `isinstance(aty, types.Boolean)` or use above definition. Just need somebody to tell which one to use.

numba/cpython/builtins.py
kozlov-alexey: Yes, just added it.

numba/parfors/parfor_lowering.py
stuartarchibald: @kozlov-alexey,  @DrTodd13 and I just had an extensive out of band discussion about this and have concluded that a less restrictive `isinstance(aty, types.Boolean)` type check would be good for now. The implications of sub-classing from core Numba types in a breaking manner are not considered elsewhere in the code base and were such a case to occur there's a few options for resolving this, but such a solutions are inevitably quite wide reaching and would impact more than `parfors`. Hows that sound? Thanks!

numba/cpython/builtins.py
stuartarchibald: Thanks!

numba/parfors/parfor_lowering.py
stuartarchibald: @kozlov-alexey think it's just ^ to resolve then this is good to merge. Thanks.

stuartarchibald (approved):
Thanks for the patch and fixes, looks good!

PR #6016, opened 2020-07-21
Fixes change of list spelling in a cuda test.

stuartarchibald:
As title

PR #6020, opened 2020-07-22
CUDA: Fix #5820, adding atomic nanmin / nanmax

### LINKED ISSUE #5820: Add back support for atomics that always return the non-NaN values

gmarkall:
This continues on from PR #5941, to fix #5820 by adding implementations of atomic `nanmin` and `nanmax` for CUDA.

It turned out that the `max` and `min` atomic functions also failed to return the old value when it was replaced, so a fix for this is included.

### DIFF
--- a/numba/cuda/cudadrv/nvvm.py
+++ b/numba/cuda/cudadrv/nvvm.py
@@ -508,16 +508,16 @@ def llvm_to_ptx(llvmir, **opts):
                                     PTR_OR_VAL='ptr')),
         ('declare float @___numba_atomic_float_nanmax(float*, float)',
          ir_numba_atomic_max.format(T='float', Ti='i32', NAN='nan', OP='ult',
-                                    PTR_OR_VAL='ptr')),
+                                    PTR_OR_VAL='')),
         ('declare double @___numba_atomic_double_nanmax(double*, double)',
          ir_numba_atomic_max.format(T='double', Ti='i64', NAN='nan', OP='ult',
-                                    PTR_OR_VAL='ptr')),
+                                    PTR_OR_VAL='')),
         ('declare float @___numba_atomic_float_nanmin(float*, float)',
          ir_numba_atomic_min.format(T='float', Ti='i32', NAN='nan', OP='ugt',
-                                    PTR_OR_VAL='ptr')),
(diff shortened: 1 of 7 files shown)


stuartarchibald:
Thanks for carrying on with this @gmarkall. Following on from @sklam's review, https://github.com/numba/numba/pull/5941#discussion_r447589060 needs checking/doing if not done.

numba/cuda/cudadrv/nvvm.py
```diff
@@ -423,25 +423,23 @@ def get(self):
...
     ; Check if either value is a NaN and return *ptr early if so
-    %valnan = fcmp uno {T} %val, %ptrval
+    %valnan = fcmp uno {T} %val, %{PTR_OR_VAL}val
     br i1 %valnan, label %done, label %gt_check
 
 gt_check:
     %dold = phi {T} [ %ptrval, %entry ], [ %dcas, %attempt ]
     ; Continue attempts if dold > val
-    %lt = fcmp nnan ogt {T} %dold, %val
+    %lt = fcmp {OP} {T} %dold, %val
```
stuartarchibald, line 442:
```suggestion
    %gt = fcmp {OP} {T} %dold, %val
```
should probably be called `gt`? But moot/ call it `cmp`? if the templates are unified.

numba/cuda/cudadrv/nvvm.py
```diff
@@ -400,17 +400,17 @@ def get(self):
 
 
 ir_numba_atomic_max = """
-define internal {T} @___numba_atomic_{T}_max({T}* %ptr, {T} %val) alwaysinline {{
+define internal {T} @___numba_atomic_{T}_{NAN}max({T}* %ptr, {T} %val) alwaysinline {{
```
stuartarchibald, line 403: I think the two templates for max/min can be unified following the latest changes?

esc:
BFID: `numba_smoketest_cuda_71`

sklam:
Re-run buildfarm since https://github.com/numba/numba/pull/6016 merged to addresses the failed test (https://github.com/numba/numba/pull/6020#pullrequestreview-456003100)

BFID: numba_smoketest_cuda_72

PR #6021, opened 2020-07-22
[WIP] Does the c parameter to typeof_impl serve a purpose?

gmarkall:
(No pun intended) - just opening this PR to see what happens on CI.


gmarkall:
Answer: the purpose is to enable typing differently depending on whether the value is being typed as an argument or a constant. It is not used within Numba, but could be useful for someone extending Numba.

PR #6027, opened 2020-07-23
Support operators inlining in InlineOverloads

### LINKED ISSUE #5064: overload inlining fails for builtin operators
## Reporting a bug

`overload` with `inline='always'` fails for builtin operators:
```python
import operator
import numba
from numba.extending import overload, register_model, models, unbox, NativeValue, typeof_impl

class Foo:
    pass

class FooType(numba.types.Type):
    def __init__(self):
        super(FooType, self).__init__(name='Foo')


foo_type = FooType()
register_model(FooType)(models.OpaqueModel)


@typeof_impl.register(Foo)
def _typ_foo(val, c):
  return FooType()


@unbox(FooType)
def unbox_foo(typ, obj, c):
  return NativeValue(c.context.get_dummy_value())


@overload(operator.eq, inline='always')
def foo_eq(a, b):
    if a == FooType():
        return lambda a, b: True


@numba.njit
def f(a):
    return a == "A"


print(f(Foo()))

```
```
  File "/opt/anaconda3/envs/TEST3/lib/python3.7/site-packages/numba/typed_passes.py", line 349, in run_pass
    lower.lower()
  File "/opt/anaconda3/envs/TEST3/lib/python3.7/site-packages/numba/lowering.py", line 195, in lower
  File "/opt/anaconda3/envs/TEST3/lib/python3.7/site-packages/numba/lowering.py", line 248, in lower_normal_function
  File "/opt/anaconda3/envs/TEST3/lib/python3.7/site-packages/numba/lowering.py", line 273, in lower_function_body
    self.lower_block(block)
  File "/opt/anaconda3/envs/TEST3/lib/python3.7/site-packages/numba/lowering.py", line 288, in lower_block
    self.lower_inst(inst)
  File "/opt/anaconda3/envs/TEST3/lib/python3.7/contextlib.py", line 130, in __exit__
    self.gen.throw(type, value, traceback)
  File "/opt/anaconda3/envs/TEST3/lib/python3.7/site-packages/numba/errors.py", line 725, in new_error_context
    six.reraise(type(newerr), newerr, tb)
  File "/opt/anaconda3/envs/TEST3/lib/python3.7/site-packages/numba/six.py", line 669, in reraise
    raise value
numba.errors.LoweringError: Failed in nopython mode pipeline (step: nopython mode backend)
No definition for lowering _EmptyImplementationEntry('always inlined')(Foo, Literal[str](A)) -> bool

File "test_builtins_inline.py", line 35:
def f(a):
    return a == "A"
    ^

[1] During: lowering "$6compare_op.2 = a == $const4.1" at test_builtins_inline.py (35)
```

kozlov-alexey:
This commit adds support for inlining of `unary`, `binop`, `inplace_binop `and `getitem `expressions
in InlineOverload pass. Fix #5064.


esc:
@kozlov-alexey thanks for submitting this, I have added it to the review queue.

sklam:
@kozlov-alexey, the title says "WIP". Are you still working on it?

kozlov-alexey:
@sklam Yes, I wanted to add some more changes for solution to work with all operators (not just binops). But it seems my new changes broke some tests (which I didn't notice, since they are skipped on python 3.8). I'll fix that and remove the tag once it's done.

kozlov-alexey:
@sklam I finished making changes and tests now pass. Can you please add it to review queue once again?

kozlov-alexey:
@sklam, @stuartarchibald Is this really on review? Can you maybe take a look this week an provide initial feedback?

numba/core/ir_utils.py
```diff
@@ -1737,6 +1737,12 @@ def dump_blocks(blocks):
         for stmt in block.body:
             print("    ", stmt)
 
+def is_operator(expr):
```
stuartarchibald, line 1740: This name seems a bit strange as `*getitem` is in there? How about `is_operator_or_getitem`?

numba/core/typed_passes.py
```diff
@@ -648,6 +651,42 @@ def _do_work_call(self, state, work_list, block, i, expr, inline_worker):
...
 
+    def _do_work_ops(self, state, work_list, block, i, expr, inline_worker):
+
+        # search for this binop overloads with this "inline"
+        sig = state.type_annotation.calltypes[expr]
+        if sig is None:
+            return False
+
+        arg_typs = sig.args
+        templates = state.typingctx._functions.get(expr.fn, [])
```
stuartarchibald, line 662: This should probably be something like `state.typingctx.resolve_value_type(expr.fn)` which will resolve to the `types.Function` for the `expr.fn`. The templates are in the `.templates` attr on the resolved function, so something like https://github.com/numba/numba/pull/6027/files#diff-936f29e49d280cc5a6e7b5405222612dR617 should then work.

numba/core/typed_passes.py
```diff
@@ -648,6 +651,42 @@ def _do_work_call(self, state, work_list, block, i, expr, inline_worker):
...
+            if not inline_type.is_never_inline:
+                try:
+                    impl = template._overload_func(*arg_typs)
+                    if impl is None:
+                        raise Exception  # abort for this template
+                    break
+                except Exception:
+                    continue
+        else:
+            return False
```
stuartarchibald, line 682: This is a copy of the code above in `_do_work_call`? Perhaps it's time for a refactor?

numba/core/typing/context.py
```diff
@@ -307,7 +307,7 @@ def resolve_setattr(self, target, attr, value):
     def resolve_static_getitem(self, value, index):
         assert not isinstance(index, types.Type), index
         args = value, index
-        kws = ()
+        kws = {}
```
stuartarchibald, line 310: Is there a measurable impact from altering this?

numba/tests/test_ir_inlining.py
```diff
@@ -526,6 +529,48 @@ def impl():
...
+                return lambda a, idx: None
+
+        def impl_eq(x):
+            return x == 1
+
+        def impl_getitem(x):
+            return x[1]
+
+        self.check(impl_eq, Dummy(), inline_expect={'eq': True})
+        self.check(impl_getitem, Dummy(), inline_expect={'getitem': True})
```
stuartarchibald, line 572: Given the scope of what `is_operator` is covering, some more tests would be good. Would suggest covering at least one each of: `['unary', 'binop', 'inplace_binop', 'getitem', 'static_getitem']`.

stuartarchibald (commented):
Thanks for the PR. I've given it an initial review,  the implementation itself seems sound, comments are mainly around refactoring and testing. Thanks again!

numba/core/typed_passes.py
kozlov-alexey: @stuartarchibald Yes, thanks for pointing the right way. I changed it to using `resolve_value_type`, but I didn't quite understand the next part regarding rating/sorting the candidates like in `resolve_overload`. It was so that the first template, matching arg types was selected anyway. Also it turned out that the list of templates obtained with `state.typingctx.resolve_value_type(expr.fn).templates` is larger than the one with `state.typingctx._functions.get(expr.fn, [])`.   It contains certain entries looking like not yet evaluated (they have empty `_get_template_info`), but it works since they are filtered with `inline_type = getattr(template, '_inline', None)`.

numba/core/typed_passes.py
kozlov-alexey: @stuartarchibald Sure, but after that please don't say it was better the old way. 😄 
I made some refactoring keeping the logic the same. Any comments and remarks are very welcome.
If that looks ugly in your opinion we can probably revert most of it and use only `select_template`  function to actually find the inlinee template in all three branches.

numba/core/typing/context.py
kozlov-alexey: Probably no, I thought it was the reason of some test fails, but seems not, so I've reverted it.

kozlov-alexey:
@sklam, @stuartarchibald, Can you please take a look one more time and provide feedback on recent changes? Also do you think we should deal with inlining `operator.setitem` in scope of this or another PR?

numba/tests/test_ir_inlining.py
```diff
@@ -126,12 +126,34 @@ def check(self, test_impl, *args, **kwargs):
...
                     found |= func_defn.name == k
-                elif getattr(expr, 'op', False) and ir_utils.is_operator(expr):
+                elif ir_utils.is_operator_or_getitem(expr):
                     found |= expr.fn.__name__ == k
             self.assertFalse(found == v)
 
         return fir  # for use in further analysis
 
+    def make_dummy_type(self):
+        # Use test_id to makesure no collision is possible.
```
stuartarchibald, line 136:
```suggestion
        # Use test_id to make sure no collision is possible.
```

numba/tests/test_ir_inlining.py
```diff
@@ -529,47 +557,138 @@ def impl():
...
-        def overload_dummy_getitem(a, idx):
-            if a == dummy_type:
-                return lambda a, idx: None
+        def overload_dummy_getitem(obj, idx):
+            if isinstance(obj, self.DummyType):
+                return dummy_getitem_impl
 
-        def impl_eq(x):
-            return x == 1
+        # noth getitem and static_getitem Exprs refer to opertor.getitem
```
stuartarchibald, line 663:
```suggestion
        # note getitem and static_getitem Exprs refer to operator.getitem
```

numba/tests/test_ir_inlining.py
```diff
@@ -529,47 +557,138 @@ def impl():
...
+        # hence they are checked using the same expect key
```
stuartarchibald, line 664:
```suggestion
        # hence they are checked using the same expected key
```

stuartarchibald (commented):
Thanks for doing the refactor, hopefully that'll make it a bit easier to maintain/update should needs be. Recent commits look good, just 3 typos else good to go! Thanks again!

stuartarchibald:
> <snip> Also do you think we should deal with inlining `operator.setitem` in scope of this or another PR?

Would suggest getting this in and handling `setitem` in a subsequent PR.

stuartarchibald:
14cd412 fixes the 3 typos.

esc:
@stuartarchibald I checked 14cd412 and it is fine.

stuartarchibald:
> @stuartarchibald I checked [14cd412](https://github.com/numba/numba/commit/14cd41274e647fce7b74ca041e86e756aa83e629) and it is fine.

thanks @esc

PR #6031, opened 2020-07-24
Continue 5821

stuartarchibald:
Continues PR #5821 addressing feedback from review.

Closes #5821

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -201,8 +201,8 @@ def get_ext_modules():
         else:
             omplinkflags = ['-fopenmp']
 
-    # Disable tbb if forced by user with NUMBA_NO_TBB=1
-    if os.getenv("NUMBA_NO_TBB") == "1":
+    # Disable tbb if forced by user with NUMBA_DISABLE_TBB=1
+    if os.getenv("NUMBA_DISABLE_TBB"):
         print("TBB disabled")
     else:
         # Search for Intel TBB, first check env var TBBROOT then conda locations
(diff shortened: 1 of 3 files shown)


stuartarchibald:
@esc please could you add this to your review queue, thanks.

setup.py
```diff
@@ -205,33 +201,42 @@ def check_file_at_path(path2file):
...
-                os.path.join(tbb_root, 'lib', 'intel64', 'gcc4.4'),
-                # for MacOS
-                os.path.join(tbb_root, 'lib'),
-                # for Windows
-                os.path.join(tbb_root, 'lib', 'intel64', 'vc_mt'),
-            ],
-        )
-        ext_np_ufunc_backends.append(ext_np_ufunc_tbb_backend)
+    # Disable tbb if forced by user with NUMBA_DISABLE_TBB=1
+    if os.getenv("NUMBA_DISABLE_TBB"):
```
esc, line 205: The problem with this is that a Falsy value like `0` will become a string with length 1 --- which is a truthy. I'm not sure what the best solution here is, maybe it's best to state in the documentation that the variable must be set to a non-empty string? I think this is the most  accurate description.

setup.py
```diff
@@ -205,33 +201,42 @@ def check_file_at_path(path2file):
...
+                ],
+            )
+            ext_np_ufunc_backends.append(ext_np_ufunc_tbb_backend)
+        else:
+            print("TBB not found")
 
-    # Disable OpenMP if forced by user with NUMBA_NO_OPENMP=1
-    if os.getenv('NUMBA_NO_OPENMP'):
+    # Disable OpenMP if forced by user with NUMBA_DISABLE_OPENMP=1
+    if os.getenv('NUMBA_DISABLE_OPENMP'):
```
esc, line 239: As above

setup.py
stuartarchibald: Good point, done in a02bdea

PR #6036, opened 2020-07-25
Update deprecation doc from numba.errors to numba.core.errors

rht:
Currently, doing `from numba.errors import NumbaDeprecationWarning, NumbaPendingDeprecationWarning` would result in a warning message of
```
NumbaDeprecationWarning: An import was requested from a module that has moved location.
Import requested from: 'numba.errors', please update to use 'numba.core.errors' or pin to Numba version 0.48.0. This alias will not be present in Numba version 0.50.0.
```
This PR removes this annoying warning message.


esc:
@rht good catch! Thanks for submitting this, I have added it to the queue for review.

Issue #3003, opened 2018-05-30
Reduce CUDA launch overhead


seibert:
As noted on the Numba mailing list, the overhead for launching a CUDA kernel is rather high.   Some benchmarking of different size add kernels indicates that:

Numba launch overhead: 200 us
CuPy launch overhead: 25 us

(Test system is Intel(R) Core(TM) i7-4820K CPU @ 3.70GHz, Ubuntu 16.04, CUDA 9.0, Numba 0.39dev)

We should profile the kernel launch process to see where the time is going.

seibert:
Generally speaking, it looks like we are spending a long time calling ctypes constructors.  As annoying as it will be, I think we might need to do more of this critical path in C or Cython.

sjoerdos92:
@seibert I am writing CUDA functions in Python and I have many of the same issues described above. The 'resolve_argument_type' time @stuartarchibald reported on can be completely removed by typing the inputs of the CUDA kernel so numba doesn't have to determine them itself. 

One other major contribution I found was the _prepare_args function you also found in your profile. Since I execute the same kernel many times with the same arguments, my idea was to cache the results of the _prepare_args function. This reduced the overhead by this function significantly (from 6 seconds to 1 second according to my profiler).

I agree that moving parts of the critical path to C or Cython should help a lot. However, perhaps an easier and less annoying solution would be to cache the output of functions that will be called over and over again.

Please look through the ugliness of the solution I quickly attempted. I saved the kernelargs in the class above _kernel_call and retrieved them if the hash of the arguments and the argument_types is identical.

My edit:
`        prepare_args_again = True
        hash_val = hash(self.argument_types) + hash(args)
        if hasattr(self.__class__, 'cache_kernelargs'):
            if hash_val in self.__class__.cache_kernelargs:
                kernelargs = self.__class__.cache_kernelargs[hash_val]
                prepare_args_again = False
                
        if prepare_args_again:
            kernelargs = []
            for t, v in zip(self.argument_types, args):
                self._prepare_args(t, v, stream, retr, kernelargs)
            if hasattr(self, 'cache_kernelargs'):
                self.__class__.cache_kernelargs[hash_val] = kernelargs
            else:
                self.__class__.cache_kernelargs = {hash_val : kernelargs}`

Complete modified method _kernel_call in compiler.py
`    def _kernel_call(self, args, griddim, blockdim, stream=0, sharedmem=0):
        # Prepare kernel
        cufunc = self._func.get()

        if self.debug:
[...]

PR #6042, opened 2020-07-28
CUDA: Lower launch overhead by launching kernel directly

gmarkall:
Instead of configuring a `driver.Function`, which creates a copy, we can directly launch with the configuration - it seems that we don't really do anything with the configured copy other than launch with it then immediately forget it anyway. Using the code from Issue #3003 (See
https://github.com/numba/numba/issues/3003#issuecomment-627872661)., timings on my system (i7-6700K, RTX 8000) before:

```
$ ipython repro.ipy
31.3 µs ± 71.5 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
46.1 µs ± 87.7 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
56 µs ± 110 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
64.8 µs ± 173 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
73.2 µs ± 91.3 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

After:

```
$ ipython repro.ipy
26.4 µs ± 78 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
41.4 µs ± 185 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
51.1 µs ± 511 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
59.5 µs ± 193 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
68 µs ± 87 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

i.e. a saving of about 5µs per kernel launch. I plan to make further changes in subsequent PRs to eat away at other sources of overhead (`_prepare_args`, etc.) at some point in the future.


seibert:
Very excited to see this!

seibert:
Looking forward to the dream where we can launch CUDA kernels directly from nopython mode on the CPU.  :)

gmarkall:
> Looking forward to the dream where we can launch CUDA kernels directly from nopython mode on the CPU. :)

I made this PR because it seemed like a quick win, but it's really part of an experiment where I'm converting the kernel launches and argument preparation to be done by a generated wrapper function instead of using the Python driver API wrappers - I was doing this because I wanted boxing/unboxing extensions to work for CUDA, but as you point out, it's also a part of the road to being able to launch kernels from nopython mode.

gmarkall:
To add another data point: this PR seems to cut the runtime of GPU-STUMP down by 59%: https://github.com/TDAmeritrade/stumpy/issues/245

stuartarchibald:
Should this close https://github.com/numba/numba/issues/3003 ?

gmarkall:
I don't think so, there's plenty more opportunity after this PR.

stuartarchibald:
I've just taken a look at the patch and think this is a good change, however, it is altering behaviour, question is... does it warrant a deprecation cycle as it's advertised [public API](https://numba.readthedocs.io/en/stable/cuda-reference/kernel.html#dispatcher-objects)? Could something like:
```python
from numba import cuda

@cuda.jit
def foo():
    pass

foo_1_1 = foo.configure(1, 1)
foo_3_3 = foo[3, 3]

foo_1_1()
foo_3_3()
```
exist in the wild?

gmarkall:
> I've just taken a look at the patch and think this is a good change, however, it is altering behaviour, question is... does it warrant a deprecation cycle as it's advertised [public API](https://numba.readthedocs.io/en/stable/cuda-reference/kernel.html#dispatcher-objects)? Could something like:
> 
> ```python
> from numba import cuda
> 
> @cuda.jit
> def foo():
>     pass
> 
> foo_1_1 = foo.configure(1, 1)
> foo_3_3 = foo[3, 3]
> 
> foo_1_1()
> foo_3_3()
> ```
> 
> exist in the wild?

This isn't changing a public API - the `configure` that is removed here is `numba.cuda.cudadrv.Function.configure`, not `numba.cuda.compiler.Dispatcher.configure` - your above example still works with the change in this PR.

I've just merged master and resolved conflicts - is this now ready for a buildfarm run?

stuartarchibald:
> > I've just taken a look at the patch and think this is a good change, however, it is altering behaviour, question is... does it warrant a deprecation cycle as it's advertised [public API](https://numba.readthedocs.io/en/stable/cuda-reference/kernel.html#dispatcher-objects)? Could something like:
> > ```python
> > from numba import cuda
> > 
> > @cuda.jit
> > def foo():
> >     pass
> > 
> > foo_1_1 = foo.configure(1, 1)
> > foo_3_3 = foo[3, 3]
> > 
> > foo_1_1()
> > foo_3_3()
> > ```
> > 
> > 
> > exist in the wild?
> 
> This isn't changing a public API - the `configure` that is removed here is `numba.cuda.cudadrv.Function.configure`, not `numba.cuda.compiler.Dispatcher.configure` - your above example still works with the change in this PR.

Apologies, I evidently misread this and thought it was referring to the dispatcher config.

> I've just merged master and resolved conflicts - is this now ready for a buildfarm run?

I think so. Will schedule it.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_107`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_107`.

Passed.

stuartarchibald (approved):
Patch looks good, thanks @gmarkall

PR #6043, opened 2020-07-28
[WIP] Add no_unliteral flag to overloads

### LINKED ISSUE #5411: Avoid calling typers/overloads without literal types
## Feature request

Type inference currently calls every typing template twice: once using literal types if available, and again without literal types. I believe the original purpose was backwards compatibility when literal types were introduced.

However, this can be confusing for developers. For example, if an overload expects a literal type but fails due to another reason, Numba tries again without literal types. The developer doesn't see the original error, and may be confused why the overload is called without literal types (our engineers have experienced these issues *many* times).

Ideally, all templates should handle literal types so Numba wouldn't need to have this behavior. In the short term, there could be an option for overloads (e.g. "literal_only=True").

ehsantn:
This flag avoids calling overloads again without literal types if the first call with literals was unsuccessful.
Closes #5411.

I think having the opposite option of just calling without literals is also helpful to avoid overspecialization. So this flag can be renamed/expanded to support both.

### DIFF
--- a/numba/core/extending.py
+++ b/numba/core/extending.py
@@ -55,7 +55,7 @@ def type_callable(func):
 _overload_default_jit_options = {'no_cpython_wrapper': True}
 
 
-def overload(func, jit_options={}, strict=True, inline='never'):
+def overload(func, jit_options={}, strict=True, inline="never", no_unliteral=False):
     """
     A decorator marking the decorated function as typing and implementing
     *func* in nopython mode.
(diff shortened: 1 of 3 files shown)


ehsantn:
@stuartarchibald This is the no_unliteral flag I mentioned last week. We have been using this significantly for a while now, but this may need to be expanded to support non-literal only case as well. Feel free to take code from this branch or just modify it.

PR #6052, opened 2020-07-30
Fix dtype for atomic_add_double tests

gmarkall:
On Windows the default NumPy `int` is `int32`, but we compile the kernel for `int64` arguments, resulting in data being interpreted incorrectly in the kernel.

This is fixed by explicitly making the index arrays `int64` arrays.


sklam:
BFID: `numba_smoketest_cuda_90`

sklam:
smoketest passed

Issue #5770, opened 2020-05-28
List populated during multiple loops does not convert to array correctly. 


exe0cdc:
I am not sure if this is intended behaviour or if there is some obvious pitfall that I am missing but when I create a function where values are appended to a list in more than one loop and this list is converted to an array, only the values appended  during the last loop get returned. Here is what I expect: 

```python
def foo():
    result = []
    for i in range(5):
        result.append(i)
    for j in range(5,10):
        result.append(j)
    return np.array(result)
```
```python
>>>foo()
array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
```
but when I define the function with `@jit(cache=True, nopython=True)` I get this result:
```python
>>>foo()
array([5, 6, 7, 8, 9])
```
In the case where I return the list without converting to an array the full list of values is returned as expected. Strangely, if I print the list before converting to an array, the full array is also returned. 

I am on numba 0.49.1, numpy 1.18.1 and Python 3.8.1.

esc:
@exe0cdc thanks for reporting this! I can confirm that I am able to reproduce this. It does look like a bug indeed.

For reference here is the reproducer I used:

```python
from numba import njit
import numpy as np


@njit
def foo_reflected():
    result = []
    for i in range(5):
        result.append(i)
    for j in range(5,10):
        result.append(j)
    return np.array(result)


print(foo_reflected())
print(foo_reflected.py_func())
```

esc:
FWIW: here is a workaround that uses the `numba.types.List` data-structure:

```python
from numba import njit
from numba.typed import List
import numpy as np

@njit
def foo_typed():
    result = List()
    for i in range(5):
        result.append(i)
    for j in range(5,10):
        result.append(j)
    return np.asarray(result)


print(foo_typed())
print(foo_typed.py_func())
```

This yields:

```
[0 1 2 3 4 5 6 7 8 9]
[0 1 2 3 4 5 6 7 8 9]
```

Note the use the `List` and `asarray`. The plain `array` constructor can not yet handle `numba.typed.List`.

stuartarchibald:
Think this will be a problem with the array inliner.

exe0cdc:
Thanks @esc, I was going mad debugging this.

stuartarchibald:
This doesn't reproduce on 3.7. Would suggest it's a bad interaction between array inlining and the way loops end up being represented in 3.8.

This:
```python
import sys
from numba import njit
import numpy as np

@njit
def foo():
[...]

PR #6056, opened 2020-07-31
Fix issue on invalid inlining of non-empty build_list by inline_arraycall

### LINKED ISSUE #6053: Numba disregards the first element of the output list, using simple python range
This is my first try of using numba. Strange that it has such profound bugs in parsing simple python code.

## Reporting a bug

```python
@numba.njit()
def compute(n=10):
    res = [0.]
    for i in range(n):
        res.append(i)
    return np.array(res, dtype=np.float32)
```

With the `numba.njit` decorator the output is `[0. 1. 2. 3. 4. 5. 6. 7. 8. 9.]`.
Without the decorator the (correct) output is `[0. 0. 1. 2. 3. 4. 5. 6. 7. 8. 9.]`

numba ~v0.51.0~ 0.50.1, installed with conda.

sklam:
fixes #6053

The issue is only affecting python3.8 due to different bytecode.

### DIFF
--- a/numba/core/inline_closurecall.py
+++ b/numba/core/inline_closurecall.py
@@ -939,6 +939,8 @@ def _inline_arraycall(func_ir, cfg, visited, loop, swapped, enable_prange=False,
         list_var_def = get_definition(func_ir, list_var_def.value)
     # Check if the definition is a build_list
     require(isinstance(list_var_def, ir.Expr) and list_var_def.op ==  'build_list')
+    # The build_list must be empty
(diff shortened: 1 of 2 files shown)


esc:
BFID: numba_smoketest_cpu_62

esc:
smoketest was fine.

esc:
Marking as RTM

stuartarchibald:
Note this does not close the related https://github.com/numba/numba/issues/5770

PR #6057, opened 2020-07-31
fix aarch64/python_3.8 failure on master

esc:
As part of the smoke-test strategy, we discovered the following test
failures on master for the `aarch64` target and Python 3.8, around 29
Jul 2020:

```
INFO - ======================================================================
INFO - ERROR: test_basic96 (numba.tests.test_stencils.TestManyStencils)
INFO - 1D slice.
INFO - ----------------------------------------------------------------------
INFO - Traceback (most recent call last):
INFO - File "/opt/conda/envs/testenv_17dc0518-6796-4cd9-8247-44db76254a67/lib/python3.8/site-packages/numba/tests/test_stencils.py", line 2759, in test_basic96
INFO - self.check(kernel, a, options={'neighborhood': ((-1, 1),)})
INFO - File "/opt/conda/envs/testenv_17dc0518-6796-4cd9-8247-44db76254a67/lib/python3.8/site-packages/numba/tests/test_stencils.py", line 1668, in check
INFO - raise RuntimeError(str1 + str2)
INFO - RuntimeError: The following implementations should not have raised an exception but did:
INFO - ['pyStencil']
INFO - Errors were:
INFO -
INFO - pyStencil: Message: <class 'ValueError'>: Name node can't be used with 'None' constant
```

This was reproduced using the `aarch64` emulation provided by qemu and
Docker on Mac OSX and shown to be present in the most recent release
0.50.1.

Further investigation lead to some potentially related clues at:

https://github.com/beetbox/beets/pull/3621

.. and eventual trial and error led to this fix.

It is unclear, exactly where the error came from and why it  only showed
up in Python  3.8 on `aarch64`.

It should be noted, that this was only noticed as the smoke tests
executes the tests  with Python 3.8 on this target. The regular tests,
only executed the tests with Pythons 3.6 and 3.7. So, the regular tests
should also include Python 3.8 on the farm to prevent any regressions.

### DIFF
--- a/numba/tests/test_stencils.py
+++ b/numba/tests/test_stencils.py
@@ -1260,7 +1260,7 @@ class pyStencilGenerator:
             ast.copy_location(returner, node)
 
             add_kwarg = [ast.arg('neighborhood', None)]
-            defaults = [ast.Name(id='None', ctx=ast.Load())]
+            defaults = [ast.Name(id='Name', ctx=ast.Load())]
 
             newargs = ast.arguments(
                 args=node.args.args +
--- a/numba/tests/test_stencils.py
+++ b/numba/tests/test_stencils.py
@@ -1260,7 +1260,7 @@ class pyStencilGenerator:
             ast.copy_location(returner, node)
 
             add_kwarg = [ast.arg('neighborhood', None)]
-            defaults = [ast.Name(id='Name', ctx=ast.Load())]
+            defaults = []
 
             newargs = ast.arguments(
                 args=node.args.args +


esc:
BFID: `numba_smoketest_cpu_61`.

sklam:
I can replicate it on OSX using `python=3.8.5` from `conda-forge`

esc:
> I can replicate it on OSX using `python=3.8.5` from `conda-forge`

Ohhh, interesting!

numba/tests/test_stencils.py
```diff
@@ -1260,7 +1260,7 @@ def computebound(mins, maxs):
             ast.copy_location(returner, node)
 
             add_kwarg = [ast.arg('neighborhood', None)]
-            defaults = [ast.Name(id='None', ctx=ast.Load())]
+            defaults = [ast.Name(id='Name', ctx=ast.Load())]
```
sklam, line 1263:
```suggestion
            defaults = []
```
this work as well

sklam:
For future reference, I did this to find out what's going on:

```diff
diff --git a/numba/tests/test_stencils.py b/numba/tests/test_stencils.py
index e28dad9bc..908ade450 100644
--- a/numba/tests/test_stencils.py
+++ b/numba/tests/test_stencils.py
@@ -1260,7 +1260,7 @@ class pyStencilGenerator:
             ast.copy_location(returner, node)

             add_kwarg = [ast.arg('neighborhood', None)]
-            defaults = [ast.Name(id='Name', ctx=ast.Load())]
+            defaults = [ast.Constant(None)]

             newargs = ast.arguments(
                 args=node.args.args +
@@ -1428,6 +1428,8 @@ class pyStencilGenerator:
                 standard_indexing,
                 neighborhood)
             fixer.visit(tree)
+            import astor
+            print(astor.to_source(tree))

             # fix up the call sites so they work better with astor
             callFixer = self.FixCalls()
@@ -1535,6 +1537,7 @@ class TestManyStencils(TestStencilBase):
                     if not found:
                         raise
                 else:
+                    raise
                     should_not_fail.append(
                         (usecase, "%s: %s" %
                          (type(e), str(e))))
```

With `astor.to_source`, we can see the modified AST printed as python source:

```python
def __kernel(a, neighborhood=None):

    def check_stencil_arrays(*args, **kwargs):
        neighborhood = kwargs.get('neighborhood')
        init_shape = args[0].shape
        if neighborhood is not None:
            if len(init_shape) != len(neighborhood):
                raise ValueError('Invalid neighborhood supplied')
        for x in args[1:]:
            if hasattr(x, 'shape'):
                if init_shape != x.shape:
                    raise ValueError('Input stencil arrays do not commute')
    check_stencil_arrays(a, neighborhood=neighborhood)

    def kernel(a):
        return np.median(a[-1:2])
    __retdtype = kernel(a)
    __b0 = np.full(a.shape, 0, dtype=type(__retdtype))
    for __an in range(1, a.shape[0] - 1):
        __b0[__an,] = np.median(a[__an + -1:__an + 2])
    return __b0
```

The `neighborhood=None` in `def __kernel(a, neighborhood=None):` is what the original code intended but newer Python require the use of `ast.Constant(None)`. However, this is causing another failure at:

```pytb
  File "/path/to/numba/tests/test_stencils.py", line 2762, in test_basic96
    self.check(kernel, a, options={'neighborhood': ((-1, 1),)})
  File "/path/to/numba/tests/test_stencils.py", line 1574, in check
    raise ex
  File "/path/to/numba/tests/test_stencils.py", line 1566, in check
    ast_impl = pyStencil(func_or_mode=pyfunc, **options)
  File "/path/to/numba/tests/test_stencils.py", line 1488, in pyStencil
    full_func = pytypes.FunctionType(func_code, globals())
TypeError: function() argument 'code' must be code, not None
```

because the new `None` is occuping the `.co_consts[0]` in which the code expected to be the code object.

esc:
@sklam thanks for the updates and merge!

PR #6070, opened 2020-08-04
Fix overspecialized containers

stuartarchibald:

### DIFF
--- a/numba/core/analysis.py
+++ b/numba/core/analysis.py
@@ -608,6 +608,13 @@ def find_literally_calls(func_ir, argtypes):
                     first_loc.setdefault(argindex, assign.loc)
     # Signal the dispatcher to force literal typing
     for pos in marked_args:
-        if not isinstance(argtypes[pos], (types.Literal, types.InitialValue)):
+        query_arg = argtypes[pos]
+        do_raise = (isinstance(query_arg, types.InitialValue) and
+                    query_arg.initial_value is None)
+        if do_raise:
+            loc = first_loc[pos]
+            raise errors.ForceLiteralArg(marked_args, loc=loc)
+
+        if not isinstance(query_arg, (types.Literal, types.InitialValue)):
             loc = first_loc[pos]
             raise errors.ForceLiteralArg(marked_args, loc=loc)
--- a/numba/core/types/containers.py
+++ b/numba/core/types/containers.py
@@ -455,6 +455,10 @@ class List(MutableSequence, InitialValue):
         """
         return self.dtype
 
+    def __unliteral__(self):
+        return List(self.dtype, reflected=self.reflected,
+                    initial_value=None)
+
 
 class LiteralList(Literal, _HeterogeneousTuple):
     """A heterogeneous immutable list (basically a tuple with list semantics).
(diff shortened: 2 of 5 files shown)


numba/core/types/containers.py
```diff
@@ -698,11 +710,27 @@ def unify(self, typingctx, other):
...
+            else:
+                ukey_type = self.key_type == other.key_type
+                uvalue_type = self.value_type == other.value_type
+                if ukey_type and uvalue_type:
+                    siv = self.initial_value
+                    oiv = other.initial_value
+                    siv_none = siv is None
+                    oiv_none = oiv is None
+                    if siv_none and oiv_none:
+                        return DictType(ukey_type, uvalue_type)
```
sklam, line 722: `ukey_type`, `uvalue_type` are `bool`
and i think this branch should execute if `not siv_none and not oiv_none` and the initial value doesn't match

sklam:
this example contains unify that would fail:

```python
from numba import njit
import math


@njit
def bar(x):
    o = {1: 2}
    if x:
        o = {2: 3}
    return o

@njit
def foo(x):
    if x:
        d = {3: 4}
    else:
        d = bar(x)
    return d


foo(1)

```

numba/core/types/containers.py
stuartarchibald: agree, this a) has bugs and b) is left over from when I was trying to make it do something else! 965a19be8 fixes.

numba/core/analysis.py
```diff
@@ -608,6 +608,13 @@ def find_literally_calls(func_ir, argtypes):
                     first_loc.setdefault(argindex, assign.loc)
     # Signal the dispatcher to force literal typing
     for pos in marked_args:
-        if not isinstance(argtypes[pos], (types.Literal, types.InitialValue)):
+        query_arg = argtypes[pos]
+        do_raise = (isinstance(query_arg, types.InitialValue) and
+                    query_arg.initial_value is None)
+        if do_raise:
+            loc = first_loc[pos]
+            raise errors.ForceLiteralArg(marked_args, loc=loc)
```
sklam, line 616: Are there tests exercising this branch? I can't seem to find it

sklam: my test here would trigger it but I am not sure if this is useful to raise ForceLiteralArg:

```python
from numba import njit, literally

@njit
def bar(d):
    return literally(d)

@njit
def inner(d):
    return bar(d)

@njit
def foo(x):
    d = {2: x}
    return inner(d)


r = foo(123)
print(r)
```
The literal request will never be satisfied it seems.

stuartarchibald: try e.g. `./runtests.py numba.tests.test_dictobject.TestTypedDictInitialValues.test_homogeneous_and_literal`
 ?

sklam: That test would pass for me even if the branch is commented out, but it did fail `numba.tests.test_dictobject.TestTypedDictInitialValues.test_mutation_not_carried_single_function`.
But it seems that it doesn't need to be `ForceLiteralArg` exception.

sklam: The `ForceLiteralArg` would likely be useful if `prefer_literal=True`

PR #6072, opened 2020-08-04
Fix for #6005

### LINKED ISSUE #6005: Error hidden when source is not available
## Reporting a bug

Example:

```python
import numpy as np

from numba import njit, typed, typeof
from numba.core import types


@structref.register
class ParticleType(types.StructRef):
    pass


class Particle(structref.StructRefProxy):
    def __new__(cls, pos, mass):
        return structref.StructRefProxy.__new__(cls, pos) # <- didn't provide the required mass argument


structref.define_proxy(Particle, ParticleType, ["pos", "mass"])

Particle(pos=1, mass=2)
```
Because `StructRef`is building the source code dynamically, the `inspect.getsourcelines` will fail in https://github.com/numba/numba/blob/a7c121a6631ba576561bcc6e72a4bfe97151fb4d/numba/core/typing/templates.py#L752

It will be nice to fix this to make `StructRef` error be meaningful

stuartarchibald:
Adds logic to handle missing source/source file.

Closes #6005

### DIFF
--- a/numba/core/typing/templates.py
+++ b/numba/core/typing/templates.py
@@ -285,6 +285,28 @@ class FunctionTemplate(ABC):
             key = key.im_func
         return key
 
+    @classmethod
+    def get_source_code_info(cls, impl):
+        """
+        Gets the source information about function impl.
+        Returns:
+
+        code - # str: source code as a string
+        firstlineno - # int: the first line number of the function impl
+        path - #str: the path to file containing impl
+
+        if any of the above are not available something generic is returned
+        """
+        try:
+            code, firstlineno = inspect.getsourcelines(impl)
+        except OSError: # missing source, probably a string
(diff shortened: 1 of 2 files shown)


numba/core/typing/templates.py
```diff
@@ -285,6 +285,28 @@ def get_impl_key(self, sig):
...
 
+    @classmethod
+    def get_source_code_info(cls, impl):
+        """
+        Gets the source information about function impl.
+        Returns:
+
+        code - # str: source code as a string
+        firstlineno - # int: the first line number of the function impl
+        path - #str: the path to file containing impl
```
sklam, line 296: the `#` looks odd here
```suggestion
        code - str: source code as a string
        firstlineno - int: the first line number of the function impl
        path - str: the path to file containing impl
```

PR #6075, opened 2020-08-05
add np.float_power and np.cbrt

guilhermeleobas:
As title.

### DIFF
--- a/numba/_helperlib.c
+++ b/numba/_helperlib.c
@@ -196,19 +196,6 @@ numba_signbit(npy_double a)
     return npy_signbit(a);
 }
 
-NUMBA_EXPORT_FUNC(float)
-numba_cbrtf(float x)
-{
-    return npy_cbrtf(x);
-}
-
-NUMBA_EXPORT_FUNC(double)
-numba_cbrt(double x)
-{
-    return npy_cbrt(x);
-}
-
-
 /* Unpack any Python complex-like object into a Py_complex structure */
 NUMBA_EXPORT_FUNC(int)
 numba_complex_adaptor(PyObject* obj, Py_complex *out) {
--- a/numba/_helperlib.c
+++ b/numba/_helperlib.c
@@ -196,6 +196,18 @@ numba_signbit(npy_double a)
     return npy_signbit(a);
 }
 
+NUMBA_EXPORT_FUNC(float)
+numba_cbrtf(float x)
+{
+    return npy_cbrtf(x);
+}
+
+NUMBA_EXPORT_FUNC(double)
+numba_cbrt(double x)
+{
+    return npy_cbrt(x);
+}
+
 
 /* Unpack any Python complex-like object into a Py_complex structure */
 NUMBA_EXPORT_FUNC(int)
--- a/numba/np/npyfuncs.py
+++ b/numba/np/npyfuncs.py
@@ -773,14 +773,17 @@ def np_complex_square_impl(context, builder, sig, args):
 
 def np_real_cbrt_impl(context, builder, sig, args):
     _check_arity_and_homogeneity(sig, args, 1)
+    import numpy as np
 
-    dispatch_table = {
-        types.float32: 'numba_cbrtf',
-        types.float64: 'numba_cbrt',
-    }
+    def cbrt(x):
+        if np.isnan(x):
+            return np.nan
+        elif x < 0:
+            return -np.power(-x, 1.0 / 3.0)
+        else:
+            return np.power(x, 1.0 / 3.0)
 
-    return _dispatch_func_by_name_type(context, builder, sig, args,
-                                       dispatch_table, 'cbrt')
+    return context.compile_internal(builder, cbrt, sig, args)
 
 
 ########################################################################
--- a/numba/np/npyfuncs.py
+++ b/numba/np/npyfuncs.py
@@ -775,8 +775,8 @@ def np_real_cbrt_impl(context, builder, sig, args):
     _check_arity_and_homogeneity(sig, args, 1)
 
     dispatch_table = {
-        types.float32: 'npy_cbrtf',
-        types.float64: 'npy_cbrt',
+        types.float32: 'numba_cbrtf',
+        types.float64: 'numba_cbrt',
     }
 
     return _dispatch_func_by_name_type(context, builder, sig, args,
--- a/numba/np/npyfuncs.py
+++ b/numba/np/npyfuncs.py
@@ -775,8 +775,8 @@ def np_real_cbrt_impl(context, builder, sig, args):
     _check_arity_and_homogeneity(sig, args, 1)
 
     dispatch_table = {
-        types.float32: 'cbrtf',
-        types.float64: 'cbrt',
+        types.float32: 'npy_cbrtf',
+        types.float64: 'npy_cbrt',
     }
 
     return _dispatch_func_by_name_type(context, builder, sig, args,
--- a/numba/np/npyfuncs.py
+++ b/numba/np/npyfuncs.py
@@ -481,6 +481,21 @@ def np_complex_power_impl(context, builder, sig, args):
 
 
 ########################################################################
+# numpy float power funcs
+
+def real_float_power_impl(context, builder, sig, args):
+    _check_arity_and_homogeneity(sig, args, 2)
+
+    return numbers.real_power_impl(context, builder, sig, args)
+
+
+def np_complex_float_power_impl(context, builder, sig, args):
+    _check_arity_and_homogeneity(sig, args, 2)
+
+    return numbers.complex_power_impl(context, builder, sig, args)
+
+
+########################################################################
 # numpy greatest common denominator
 
 def np_gcd_impl(context, builder, sig, args):
@@ -751,6 +766,25 @@ def np_complex_square_impl(context, builder, sig, args):
     binary_sig = typing.signature(*[sig.return_type]*3)
     return numbers.complex_mul_impl(context, builder, binary_sig,
                                      [args[0], args[0]])
+
+
+########################################################################
+# NumPy cbrt
+
+def np_real_cbrt_impl(context, builder, sig, args):
+    import numpy as np
+
+    _check_arity_and_homogeneity(sig, args, 1)
+
+    def float_cbrt(x):
+        if np.isnan(x):
+            return x
+        if x < 0:
+            return -np.power(-x, 1.0 / 3.0)
+        else:
+            return np.power(x, 1.0 / 3.0)
+    res = context.compile_internal(builder, float_cbrt, sig, args)
+    return impl_ret_untracked(context, builder, sig.return_type, res)
 
 
 ########################################################################
--- a/numba/np/npyfuncs.py
+++ b/numba/np/npyfuncs.py
@@ -779,7 +779,7 @@ def np_real_cbrt_impl(context, builder, sig, args):
     def float_cbrt(x):
         if np.isnan(x):
             return x
-        if x < 0:
+        if np.signbit(x): 
             return -np.power(-x, 1.0 / 3.0)
         else:
             return np.power(x, 1.0 / 3.0)
--- a/numba/np/npyfuncs.py
+++ b/numba/np/npyfuncs.py
@@ -772,19 +772,15 @@ def np_complex_square_impl(context, builder, sig, args):
 # NumPy cbrt
 
 def np_real_cbrt_impl(context, builder, sig, args):
-    import numpy as np
-
     _check_arity_and_homogeneity(sig, args, 1)
 
-    def float_cbrt(x):
-        if np.isnan(x):
-            return x
-        if np.signbit(x): 
-            return -np.power(-x, 1.0 / 3.0)
-        else:
-            return np.power(x, 1.0 / 3.0)
-    res = context.compile_internal(builder, float_cbrt, sig, args)
-    return impl_ret_untracked(context, builder, sig.return_type, res)
+    dispatch_table = {
+        types.float32: 'cbrtf',
+        types.float64: 'cbrt',
+    }
+
+    return _dispatch_func_by_name_type(context, builder, sig, args,
+                                       dispatch_table, 'cbrt')
 
 
 ########################################################################
--- a/numba/np/npyfuncs.py
+++ b/numba/np/npyfuncs.py
@@ -11,6 +11,7 @@ from llvmlite.llvmpy import core as lc
 
 from numba.core.imputils import impl_ret_untracked
 from numba.core import typing, types, errors, lowering, cgutils
+from numba.core.extending import register_jitable
 from numba.np import npdatetime
 from numba.cpython import cmathimpl, mathimpl, numbers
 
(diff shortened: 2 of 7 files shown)


stuartarchibald:
Thanks for opening this @guilhermeleobas , looks like CI is failing with near to floating point limit issues.

guilhermeleobas:
Hi @stuartarchibald, can we do something in this case? The `np.power` function used by `cbrt` is implemented as a call to `llvm.powi`.

guilhermeleobas:
CI is all green

numba/np/npyfuncs.py
```diff
@@ -753,6 +768,25 @@ def np_complex_square_impl(context, builder, sig, args):
...
+
+    _check_arity_and_homogeneity(sig, args, 1)
+
+    def float_cbrt(x):
+        if np.isnan(x):
+            return x
+        if x < 0:
+            return -np.power(-x, 1.0 / 3.0)
+        else:
+            return np.power(x, 1.0 / 3.0)
```
stuartarchibald, line 785: This is going to be "interesting" in that in NumPy, if a system has a `cbrt` from C99 then it gets used, else something like what you've written is used. The impact being that e.g. signed zero is incorrect for me locally as I've got C99 `cbrt`!
```python
from numba import njit
import numpy as np

@njit
def foo():
    return np.cbrt(-0.)

print(foo.py_func())
print(foo())
```
gives:
```
-0.0
0.0
```

LLVM can be coerced into generating `cbrt` with appropriate hints.

guilhermeleobas: > LLVM can be coerced into generating `cbrt` with appropriate hints.

`np.signbit` can be used here to distinguish `0.0` from its negative counterpart. I didn't want to use it since the numba implementation uses the [numpy one](https://github.com/numpy/numpy/blob/master/numpy/core/src/npymath/_signbit.c) and I'd like to make this function available in RBC.  

One could mimic `npy_signbit` [in pure LLVM](https://godbolt.org/z/nd7hj9) but that would require knowledge about the system (`sizeof(int)` and endianess).

guilhermeleobas: Replaced comparison by `np.signbit` in my last commit. I think this should fix signed zero comparison.

stuartarchibald: The issues I see with this are:

* LLVM will generate a call to `cbrt` under certain conditions and NumPy is using `libm`'s `cbrt` if it's available, so this would match exactly. Generating the call correctly would also permit e.g. use of SVML's `cbrt`. If you need a hand working out how to do this then shout! I'd recommend looking at the impact of `fastmath` flags as a starting point.
* If RBC needs this then it should implement it or make Numba generic to the point that RBC can consume it?

guilhermeleobas: > If RBC needs this then it should implement it or make Numba generic to the point that RBC can consume it?

I think it is consumable in the way it is today. We already ship our own version of `np.signbit`. By the way, I just remember that the libm version of cbrt works on RBC. So, no reason for not using it.

> LLVM will generate a call to cbrt under certain conditions and NumPy is using libm's cbrt if it's available, so this would match exactly. Generating the call correctly would also permit e.g. use of SVML's cbrt. If you need a hand working out how to do this then shout! I'd recommend looking at the impact of fastmath flags as a starting point.

I will replace the code by a call to libm cbrt.

guilhermeleobas: @stuartarchibald, I notice that a few functions calls the numpy implementation (i.e. `npy_logaddexp`). Is there any reason for not calling libm directly?

stuartarchibald: > > LLVM will generate a call to cbrt under certain conditions and NumPy is using libm's cbrt if it's available, so this would match exactly. Generating the call correctly would also permit e.g. use of SVML's cbrt. If you need a hand working out how to do this then shout! I'd recommend looking at the impact of fastmath flags as a starting point.
> 
> I will replace the code by a call to libm cbrt.

What guarantees that `cbrt` is in a `libm` ?

guilhermeleobas: Isn't `cbrt` part of `libm`?

stuartarchibald: If you have C99. NumPy does this:
https://github.com/numpy/numpy/blob/31ffdecf07d18ed4dbb66b171cb0f998d4b190fa/numpy/core/src/npymath/npy_math_internal.h.src#L508-L529

guilhermeleobas: Makes sense! I will update the PR

a.py
```diff
@@ -0,0 +1,14 @@
+import numpy as np
```
stuartarchibald, line 1: Pretty sure this shouldn't be here? Would recommend erasing it from git history.

guilhermeleobas: oops, my bad

numba/np/npyfuncs.py
guilhermeleobas: Do you have any clue of why CI is failing with `npy_cbrt`?

```C
Thread 1 "python" received signal SIGSEGV, Segmentation fault.
0x0000000000000000 in ?? ()
(gdb) bt
#0  0x0000000000000000 in ?? ()
#1  0x00007ffff7dfc051 in cpython::__main__::foo$241(long long) ()
#2  0x00007ffff3b1cb3c in call_cfunc (self=0x7fffeb670c10, cfunc=0x7fffd24648b0, args=0x7fffd24c2f10, kws=0x0, locals=0x0) at numba/_dispatcher.c:353
#3  0x00007ffff3b1d6f5 in Dispatcher_call (self=0x7fffeb670c10, args=0x7fffd24c2f10, kws=<optimized out>) at numba/_dispatcher.c:576
#4  0x000055555568395f in _PyObject_MakeTpCall () at /home/conda/feedstock_root/build_artifacts/python_1596159888721/work/Objects/call.c:159
#5  0x0000555555735300 in _PyObject_Vectorcall (kwnames=0x0, nargsf=<optimized out>, args=0x7ffff7eeb5b8, callable=0x7fffeb670c10)
    at /home/conda/feedstock_root/build_artifacts/python_1596159888721/work/Include/cpython/abstract.h:125
#6  call_function (kwnames=0x0, oparg=<optimized out>, pp_stack=<synthetic pointer>, tstate=0x5555558efa90)
    at /home/conda/feedstock_root/build_artifacts/python_1596159888721/work/Python/ceval.c:4963
#7  _PyEval_EvalFrameDefault () at /home/conda/feedstock_root/build_artifacts/python_1596159888721/work/Python/ceval.c:3500
#8  0x000055555571c190 in PyEval_EvalFrameEx (throwflag=0, f=0x7ffff7eeb440)
    at /home/conda/feedstock_root/build_artifacts/python_1596159888721/work/Python/ceval.c:741
#9  _PyEval_EvalCodeWithName () at /home/conda/feedstock_root/build_artifacts/python_1596159888721/work/Python/ceval.c:4298
#10 0x000055555571d9f3 in PyEval_EvalCodeEx (closure=0x0, kwdefs=0x0, defcount=0, defs=0x0, kwcount=0, kws=0x0, argcount=0, args=0x0,
    locals=<optimized out>, globals=<optimized out>, _co=<optimized out>)
    at /home/conda/feedstock_root/build_artifacts/python_1596159888721/work/Python/ceval.c:4327
#11 PyEval_EvalCode (co=<optimized out>, globals=<optimized out>, locals=<optimized out>)
    at /home/conda/feedstock_root/build_artifacts/python_1596159888721/work/Python/ceval.c:718
#12 0x00005555557908b2 in run_eval_code_obj () at /home/conda/feedstock_root/build_artifacts/python_1596159888721/work/Python/pythonrun.c:1125
#13 0x00005555557a3492 in run_mod () at /home/conda/feedstock_root/build_artifacts/python_1596159888721/work/Python/pythonrun.c:1147
#14 0x00005555557a657e in PyRun_FileExFlags () at /home/conda/feedstock_root/build_artifacts/python_1596159888721/work/Python/pythonrun.c:1063
#15 0x00005555557a6769 in PyRun_SimpleFileExFlags () at /home/conda/feedstock_root/build_artifacts/python_1596159888721/work/Python/pythonrun.c:428
#16 0x00005555557a6c1e in pymain_run_file (cf=0x7fffffffc0c8, config=0x5555558eec20)
    at /home/conda/feedstock_root/build_artifacts/python_1596159888721/work/Modules/main.c:387
#17 pymain_run_python (exitcode=0x7fffffffc0c0) at /home/conda/feedstock_root/build_artifacts/python_1596159888721/work/Modules/main.c:612
#18 Py_RunMain () at /home/conda/feedstock_root/build_artifacts/python_1596159888721/work/Modules/main.c:691
#19 0x00005555557a6e19 in Py_BytesMain () at /home/conda/feedstock_root/build_artifacts/python_1596159888721/work/Modules/main.c:1123
#20 0x00007ffff77e6b97 in __libc_start_main (main=0x55555565a7f0 <main>, argc=2, argv=0x7fffffffc2b8, init=<optimized out>, fini=<optimized out>,
    rtld_fini=<optimized out>, stack_end=0x7fffffffc2a8) at ../csu/libc-start.c:310
#21 0x000055555574aba9 in _start ()
(gdb)
```

guilhermeleobas:
Hi @stuartarchibald, when you have some free cycles to spare, can you take a look at this PR?

numba/_helperlib.c
```diff
@@ -196,6 +196,18 @@ numba_signbit(npy_double a)
...
+numba_cbrtf(float x)
+{
+    return npy_cbrtf(x);
+}
+
+NUMBA_EXPORT_FUNC(double)
+numba_cbrt(double x)
+{
+    return npy_cbrt(x);
+}
```
stuartarchibald, line 209: I'm struggling to understand why this is needed?

This:
```python
from numba import njit
import numpy as np

@njit(fastmath={'nnan', 'nsz', 'ninf', 'afn'})
def cbrt(x):
    third = 1.0 / 3.0
    return np.power(x, third)

@njit
def foo(x):
    return cbrt(x)

x = np.zeros(100)
foo(x)
print(cbrt.inspect_llvm(cbrt.signatures[0]))

print(cbrt.inspect_asm(cbrt.signatures[0]))
```
generates a call to the `cbrt` function, LLVM "knows" that `power(x, 1/3)` is `cbrt` under the flags set. Perhaps this can be used to ensure that a `cbrt` call is generated along with adding the inf/nan handling as per NumPy?

guilhermeleobas: > I'm struggling to understand why this is needed?
Before adding this, one of the tests were failing on CI. See:
https://github.com/numba/numba/pull/6075#discussion_r482623254

> generates a call to the cbrt function, LLVM "knows" that power(x, 1/3) is cbrt under the flags set.

For me, this produces a call to `@llvm.pow.f64(x, 1/3)`:
```llvm
; Function Attrs: nofree nounwind writeonly
define i32 @"_ZN8__main__8cbrt$242Ex"(double* noalias nocapture %retptr, { i8*, i32, i8* }** noalias nocapture readnone %excinfo, i64 %arg.x) local_unnamed_addr #0 {
entry:
  %.37.le = sitofp i64 %arg.x to double
  %.38.le = tail call nnan ninf nsz afn double @llvm.pow.f64(double %.37.le, double 0x3FD5555555555555)
  store double %.38.le, double* %retptr, align 8
  ret i32 0
}
```

stuartarchibald: Yes, but that then ends up as a call to `cbrt` once compiled?

guilhermeleobas: Oh yes, you right. Nevertheless, If I remove the code you quote, some builds on CI starts to fail. Do you want me to give a try and implement `cbrt` in numba?

guilhermeleobas:
CI is all green! Can you take a look when you have some free cycles to spare, @stuartarchibald

numba/_helperlib.c
guilhermeleobas: If one enables fast-math, you'll see a call to cbrt.

```python
from numba import njit
import numpy as np

@njit('float32(float32)', fastmath=True, no_cfunc_wrapper=True, no_cpython_wrapper=True)
def foo(x):
    return np.cbrt(x)

print(foo.inspect_asm(foo.signatures[0]))
```

```asm
	.section	__TEXT,__text,regular,pure_instructions
	.macosx_version_min 10, 15
	.section	__TEXT,__literal4,4byte_literals
	.p2align	2
LCPI0_0:
	.long	2147483648
	.section	__TEXT,__text,regular,pure_instructions
	.globl	__ZN8__main__7foo$241Ef
	.p2align	4, 0x90
__ZN8__main__7foo$241Ef:
	pushq	%r14
	pushq	%rbx
	subq	$56, %rsp
	vmovaps	%xmm0, (%rsp)
	movq	%rdi, %rbx
	vcvtss2sd	%xmm0, %xmm0, %xmm0
	movabsq	$_cbrt, %r14
	callq	*%r14
	vcvtsd2ss	%xmm0, %xmm0, %xmm0
	vmovaps	%xmm0, 32(%rsp)
	movabsq	$LCPI0_0, %rax
	vbroadcastss	(%rax), %xmm0
	vmovaps	%xmm0, 16(%rsp)
	vxorps	(%rsp), %xmm0, %xmm0
	vcvtss2sd	%xmm0, %xmm0, %xmm0
	callq	*%r14
	vcvtsd2ss	%xmm0, %xmm0, %xmm0
	vxorps	16(%rsp), %xmm0, %xmm0
	vxorps	%xmm1, %xmm1, %xmm1
	vmovaps	(%rsp), %xmm2
	vcmpltss	%xmm1, %xmm2, %xmm1
	vmovaps	32(%rsp), %xmm2
	vblendvps	%xmm1, %xmm0, %xmm2, %xmm0
	vmovss	%xmm0, (%rbx)
	xorl	%eax, %eax
	addq	$56, %rsp
	popq	%rbx
	popq	%r14
	retq

	.comm	__ZN08NumbaEnv8__main__7foo$241Ef,8,3
	.comm	__ZN08NumbaEnv5numba2np8npyfuncs17np_real_cbrt_impl12$3clocals$3e8cbrt$242Ef,8,3
.subsections_via_symbols
```

numba/np/npyfuncs.py
```diff
@@ -753,6 +769,28 @@ def np_complex_square_impl(context, builder, sig, args):
...
+    def cbrt(x):
+        if x < 0:
+            return -np.power(-x, 1.0 / 3.0)
+        else:
+            return np.power(x, 1.0 / 3.0)
+
+    def _cbrt(x):
+        if np.isnan(x):
+            return np.nan
+        return cbrt(x)
```
stuartarchibald, line 789: Please could you add a quick comment explaining why this is like it is, thanks!

numba/np/npyfuncs.py
```diff
@@ -753,6 +769,30 @@ def np_complex_square_impl(context, builder, sig, args):
                                      [args[0], args[0]])
 
 
+########################################################################
+# NumPy cbrt
+
+def np_real_cbrt_impl(context, builder, sig, args):
+    _check_arity_and_homogeneity(sig, args, 1)
+    import numpy as np
```
stuartarchibald, line 777: Imports go at the top please (I assume there's no circular reference problem?).

stuartarchibald (approved):

PR #6078, opened 2020-08-05
Duplicate NumPy's PyArray_DescrCheck macro

### LINKED ISSUE #6041: Passing in dtype to array fails on NumPy master in pip-installed version
## Reporting a bug

**Please note that this doesn't happen on the conda-installed version, only pip-installed**

```python
import numba
import numpy as np

dt1 = np.dtype("float64")
dt2 = np.dtype("float64")

dtr = np.result_type(dt1, dt2)

@numba.jit(
    nopython=True,
    nogil=True,
    locals={"data_curr": numba.np.numpy_support.from_dtype(dtr)},
)
def _dot_coo_coo(coords1, data1, coords2, data2):  # pragma: no cover
    return np.empty((0,), dtype=dtr)

coords1 = np.array([], dtype=np.int64).reshape((2, 0))
coords2 = coords1

data1 = np.array([], dtype=dt1)
data2 = data1

_dot_coo_coo(coords1, data1, coords2, data2)
```

Output
```
  File "sandbox.py", line 23, in <module>
    _dot_coo_coo(coords1, data1, coords2, data2)
TypeError: expected dtype object, got 'numpy.dtype[float64]'
```

Original issue by @TomAugspurger in pydata/sparse#383

```
Package    Version
---------- -------------------
llvmlite   0.33.0
numba      0.50.1
numpy      1.20.0.dev0+1ce5457
pip        20.1.1
scipy      1.5.2
setuptools 49.2.0
sparse     0.10.0
wheel      0.34.2
```

stuartarchibald:
As title, workaround for NumPy ABI compatibility issue.

Closes #6041

### DIFF
--- a/numba/_numba_common.h
+++ b/numba/_numba_common.h
@@ -31,6 +31,8 @@
  *
  * This macro definition is copied from:
  * https://github.com/numpy/numpy/commit/6108b5d1e138d07e3c9f2a4e3b1933749ad0e698#diff-ad2213da23136c5fc5883d9eb2d88666R26
+ *
+ * NOTE: This is the NumPy 1.18 and above version of the macro.
  */
 #define NUMBA_PyArray_DescrCheck(op) PyObject_TypeCheck(op, &PyArrayDescr_Type)
 
--- a/numba/_numba_common.h
+++ b/numba/_numba_common.h
@@ -17,4 +17,21 @@
 #define VISIBILITY_HIDDEN
 #endif
 
+/*
+ * Numba's version of the PyArray_DescrCheck macro from NumPy, use it as a
+ * direct replacement of NumPy's PyArray_DescrCheck to ensure binary
+ * compatibility.
+ *
+ * Details of why this is needed:
+ * NumPy 1.18 changed the definition of the PyArray_DescrCheck macro here:
+ * https://github.com/numpy/numpy/commit/6108b5d1e138d07e3c9f2a4e3b1933749ad0e698
+ * the result of this being that building against NumPy <1.18 would prevent
+ * Numba running against NumPy >= 1.20 as noted here:
+ * https://github.com/numba/numba/issues/6041#issuecomment-665132199
+ *
+ * This macro definition is copied from:
+ * https://github.com/numpy/numpy/commit/6108b5d1e138d07e3c9f2a4e3b1933749ad0e698#diff-ad2213da23136c5fc5883d9eb2d88666R26
+ */
+#define NUMBA_PyArray_DescrCheck(op) PyObject_TypeCheck(op, &PyArrayDescr_Type)
+
 #endif /* NUMBA_COMMON_H_ */
--- a/numba/_typeof.c
+++ b/numba/_typeof.c
@@ -4,6 +4,7 @@
 #include <time.h>
 #include <assert.h>
 
+#include "_numba_common.h"
 #include "_typeof.h"
 #include "_hashtable.h"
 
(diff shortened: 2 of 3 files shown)


numba/_numba_common.h
```diff
@@ -17,4 +17,21 @@
...
+ * NumPy 1.18 changed the definition of the PyArray_DescrCheck macro here:
+ * https://github.com/numpy/numpy/commit/6108b5d1e138d07e3c9f2a4e3b1933749ad0e698
+ * the result of this being that building against NumPy <1.18 would prevent
+ * Numba running against NumPy >= 1.20 as noted here:
+ * https://github.com/numba/numba/issues/6041#issuecomment-665132199
+ *
+ * This macro definition is copied from:
+ * https://github.com/numpy/numpy/commit/6108b5d1e138d07e3c9f2a4e3b1933749ad0e698#diff-ad2213da23136c5fc5883d9eb2d88666R26
+ */
+#define NUMBA_PyArray_DescrCheck(op) PyObject_TypeCheck(op, &PyArrayDescr_Type)
```
eric-wieser, line 35: Possibly worth emphasizing that this is the _new_ definition of the macro. Perhaps better would be to do something like
```
#if numpy_is_too_old
    #undef PyArray_DescrCheck
    #define PyArray_DescrCheck ...
#endif
```
@seberg, is there an easy way to do the `numpy_is_too_old` check?

seberg: Hmmmpf, we do have `C_API_VERSION`, but I am not sure it incremented in 1.18 specifically. Is the actual numpy version available in the header? I can't find it quickly, can't say I ever looked into those things.

The new definition is valid also for old numpy versions though.

stuartarchibald: hmmm, there's `NPY_<X>_<Y>_API_VERSION`, `NPY_ABI_VERSION` and `NPY_API_VERSION`, but if they haven't been moved I'm not sure it's something to pursue?

Perhaps the best thing is to update the comment as @eric-wieser suggested and say that this is the new definition and it's backwards compatible?

seberg: The `NPY_<X>_<Y>_API_VERSION` macro seems strange, we do have a C-API version, independend of the numpy version, which may even have incremented in 1.18.x, but that macro seems to have never changed, so I am unsure what it actually represents.

stuartarchibald: @seberg any idea where the C-API version is/what it's called?

sklam: looks like the old definition is https://github.com/numpy/numpy/blob/maintenance/1.15.x/numpy/core/include/numpy/ndarrayobject.h#L32
the new code would be compatible with the old definition. I don't see the need to have it conditional on API_VERSION, which we ain't sure.

eric-wieser: > I don't see the need to have it conditional on API_VERSION, which we ain't sure.

My suggestion was more a case of guarding against it changing again - but maybe that's unlikely and/or not relevant.
The approach in the patch is fine too, with a suitable comment.

seberg: I think the current definition will always stay correct.  We might be able to define some faster method/path, but I doubt that is relevant or even likely.  And even if we define a faster way, we would have to wait until nobody uses pre 1.20...

Quite honestly, I think you could also just not use a macro and write out the TypeCheck...

stuartarchibald: Note about it being 1.18 and above version of the macro added in https://github.com/numba/numba/pull/6078/commits/2d49e4d878f1eab44eceaa7accc3134843c5188e

stuartarchibald: Thanks @eric-wieser @seberg, given the thoughts above, and with the note made in https://github.com/numba/numba/pull/6078/commits/2d49e4d878f1eab44eceaa7accc3134843c5188e, I'm inclined to leave this as-is until something proves it's not sufficient, @sklam what do you think?

PR #5969, opened 2020-07-08
test Numba on the buildfarm with LLVM 10 enabled llvmlite

esc:
DO NOT MERGE

Issue #599, opened 2014-07-08
Type unification is non-deterministic


pitrou:
It depends on dict iteration order. Take the following code and save it as a Python script:

```
import numpy as np
from numba import jit

@jit(nopython=True)
def f(a):
    res = 0.0
    for i in range(len(a)):
        res += a[i]
    return res

print(f(np.array([1.0j])))
```

Depending on the value of PYTHONHASHSEED, it can either succeed or raise an exception:

```
$ PYTHONHASHSEED=2 python ndh.py 
1j
$ PYTHONHASHSEED=1 python ndh.py 
Traceback (most recent call last):
  File "ndh.py", line 12, in <module>
    print(f(np.array([1.0j])))
  File "/home/antoine/numba/numba/dispatcher.py", line 141, in _compile_and_call
    self.jit(sig)
  File "/home/antoine/numba/numba/dispatcher.py", line 136, in jit
    return self.compile(sig, **kws)
  File "/home/antoine/numba/numba/dispatcher.py", line 124, in compile
    flags=flags, locals=locs)
  File "/home/antoine/numba/numba/compiler.py", line 94, in compile_extra
    return_type, flags, locals)
  File "/home/antoine/numba/numba/compiler.py", line 120, in compile_bytecode
    locals)
  File "/home/antoine/numba/numba/compiler.py", line 284, in type_inference_stage
    typemap, restype, calltypes = infer.unify()
  File "/home/antoine/numba/numba/typeinfer.py", line 394, in unify
    raise TypingError("Var '%s' unified to object: %s" % (var, tv))
numba.typeinfer.TypingError: Var 'res' unified to object: res := {complex128, float64}
```

PR #6080, opened 2020-08-05
CUDA: Prevent auto-upgrade of atomic intrinsics

gmarkall:
LLVM 9 onwards auto-upgrade the atomic intrinsics into an atomicrmw form unsupported by NVVM. This commit prevents the auto-upgrade by changing the "nvvm" component of the intrinsic to "numba_nvvm" until after it has been parsed and dumped by LLVM.

With thanks to Stuart Archibald for the suggestion and initial patch, and Val Haenel for catching the issue in the buildfarm prior to release.

This should fix the issue described in https://github.com/numba/llvmlite/pull/606#issuecomment-669119556

### DIFF
--- a/numba/cuda/cudadrv/nvvm.py
+++ b/numba/cuda/cudadrv/nvvm.py
@@ -500,6 +500,12 @@ def llvm_to_ptx(llvmir, **opts):
     for decl, fn in replacements:
         llvmir = llvmir.replace(decl, fn)
 
+    # llvm.numba_nvvm.atomic is used to prevent LLVM 9 onwards auto-upgrading
+    # these intrinsics into atomicrmw instructions, which are not recognized by
+    # NVVM. We can now replace them with the real intrinsic names, ready to
+    # pass to NVVM.
+    llvmir = llvmir.replace('llvm.numba_nvvm.atomic', 'llvm.nvvm.atomic')
+
     llvmir = llvm39_to_34_ir(llvmir)
     cu.add_module(llvmir.encode('utf8'))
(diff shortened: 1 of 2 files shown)


stuartarchibald:
@esc please could you push this through the farm against the LLVM 10 stack. Thanks.

stuartarchibald:
Manually tested this on linux with LLVM 10 from https://github.com/numba/llvmlite/pull/599 locally, worked fine.

esc:
🍒 picked onto #599 as a373edde8e

esc:
CUDA smoketest passed  on #5969 with this 🍒 picked

gmarkall:
Did you intend to close this PR?

gmarkall:
> Think it got scooped into the LLVM 10 patch series in #5969 as that toolchain is the only way we have to test it's working right now (anything LLVM 9 based would still have a patched LLVM)?

Ah - I saw it there, but the big "DO NOT MERGE" message in the description led me to believe it wouldn't be merged :-)

esc:
yeah, my bad, this needs to go into `master` of course

esc:
The reason I was confused was: there are two LLVM 10 related branches, one in the Numba repo, the other in the llvmlite one. For llvmlite (https://github.com/numba/llvmlite/pull/599) this branch will be merged before the release for Numba, #5969 it is only for testing and will be closed w/o merge after release. Numba master with a new llvmlite package. So, as a result in llvmlite the PRs will be 🍒 picked and closed but in Numba they need to be 🍒  picked, tested and then merged.

PR #5319, opened 2020-02-26
Make user facing error string use abs path not rel.

stuartarchibald:
This fixes a bug that appeared on windows where the installation
and CWD were on different drives and as a result `os.path.relpath`
raises a ValueError. `os.path.abspath` will produce a longer but
more OS situation tolerant string.

PR #6081, opened 2020-08-05
Fix issue with cross drive use and relpath.

stuartarchibald:
As title. On windows relpath does not work across drives as they
technically do not share the same root. A utility function is
added to provide a "safe" relpath like function for use in future.

Previous fix was https://github.com/numba/numba/pull/5319.


esc:
Running reduced DAG of only Win-32 tests at BFID: `numba_199`

sklam:
The changes make sense. I think this is just waiting on BF to check for missing unpatched relpath.

PR #6087, opened 2020-08-06
remove invalid sanity check from randrange tests

esc:
The sanity check assumes that all the integers generated by `randrange`
are unique implicity by converting to a `set`, however `randrange` can
not make such guarantees. The reason we are removing this is because it
makes the test unstable and causes radom failures every so often.

The resulting error looks like:

```
FAIL: test_random_randrange (numba.tests.test_random.TestRandom)
File "/opt/conda/envs/testenv_d03ff656-b9e2-4389-a4e7-02f1435193ed/lib/python3.7/site-packages/numba/tests/test_random.py", line 430, in test_random_randrange
max_width, False)
File "/opt/conda/envs/testenv_d03ff656-b9e2-4389-a4e7-02f1435193ed/lib/python3.7/site-packages/numba/tests/test_random.py", line 397, in _check_randrange
self.assertEqual(len(ints), len(set(ints)), ints)
AssertionError: 30 != 29 : [0, 524311, 3170417, 4751491, 6340830, 10297652, 2, 2637917, 6340832, 4751489, 6340832, 25953158, 18891276, 19415587, 31507523, 23642767, 16777803, 25953155, 18891278, 21529290, 34677746, 23642765, 13, 3956864, 4227225, 4751504, 9511229, 524312, 2113607, 3956861]
```

Where in this case the number `6340832` is the duplicate.

### DIFF
--- a/numba/tests/test_random.py
+++ b/numba/tests/test_random.py
@@ -394,7 +394,6 @@ class TestRandom(BaseTest):
             ints.append(func2(5, 500000000))
             if func3 is not None:
                 ints.append(func3(5, 500000000, 3))
-        self.assertEqual(len(ints), len(set(ints)), ints)
         if is_numpy:
             rr = self._follow_numpy(ptr).randint
         else:


numba/tests/test_random.py
```diff
@@ -394,7 +394,6 @@ def _check_randrange(self, func1, func2, func3, ptr, max_width, is_numpy, tp=Non
             ints.append(func2(5, 500000000))
             if func3 is not None:
                 ints.append(func3(5, 500000000, 3))
-        self.assertEqual(len(ints), len(set(ints)), ints)
```
stuartarchibald, line 397: I'm not surprised that this is unstable as it seems like the seed isn't set anywhere, agree with removing it.

PR #6089, opened 2020-08-06
Fix invalid reference to TypingError

### LINKED ISSUE #6088: NameError from _unlit_non_poison
## Reporting a bug

On master and release0.51 branch, this:
```
from numba import njit

@njit
def foo():
    x = [1, 'a']
    x.append(2j)

foo()
```
produces:
```
<snip>
No implementation of function Function(<function literal_list_banned_append at 0x7f95974c5050>) found for signature:

 >>> literal_list_banned_append(LiteralList((Literal[int](1), Literal[str](a))), complex128)

  - Of which 1 did not match due to:
  Overload in function 'literal_list_banned_append': File: numba/cpython/listobj.py: Line 1132.
    With argument(s): '(Poison<LiteralList((Literal[int](1), Literal[str](a)))>, complex128)':
     NameError: name 'TypingError' is not defined
<snip>
```
comes from here:
https://github.com/numba/numba/blob/833c5907c77d09150bc6ae2a1c0284e3b827af02/numba/core/types/functions.py#L233
should be `errors.TypingError`.

stuartarchibald:
As title.

Closes #6088

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -56,7 +56,6 @@ exclude =
     numba/core/rewrites/static_binop.py
     numba/core/rewrites/ir_print.py
     numba/core/types/abstract.py
-    numba/core/types/functions.py
     numba/core/types/misc.py
     numba/core/types/npytypes.py
     numba/core/types/common.py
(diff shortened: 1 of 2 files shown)


sklam:
Perhaps, remove the file from `.flake8` ban list

stuartarchibald:
> Perhaps, remove the file from `.flake8` ban list

Done in da3e3f3

PR #6096, opened 2020-08-06
[WIP] remove deprecated tbb::task_scheduler_init, use new api

Hardcode84:
`task_scheduler_init` was removed in TBB_INTERFACE_VERSION == 12002 and new replacement api was added in 12003.

Also, 12002 lib is no longer binary compatible with previous versions so library name was changed (not properly handled yet).
Fork safety selection logic also wasn't updated.

### DIFF
--- a/numba/np/ufunc/parallel.py
+++ b/numba/np/ufunc/parallel.py
@@ -348,7 +348,7 @@ def _check_tbb_version_compatible():
         elif _IS_OSX:
             libtbb_name = 'libtbb.dylib'
         elif _IS_LINUX:
-            libtbb_name = 'libtbb.so.2'
+            libtbb_name = 'libtbb.so.12'
         else:
             raise ValueError("Unknown operating system")
         libtbb = CDLL(libtbb_name)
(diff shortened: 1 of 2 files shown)


stuartarchibald:
Thanks for opening this and starting work for the new TBB interfaces @Hardcode84

Hardcode84:
12002 binaries are available here: https://github.com/oneapi-src/oneTBB/releases/tag/v2021.1-beta08

jschueller:
hello, I have trouble understanding what task_scheduler_handle does compared to task_scheduler_init,
as it has no methods outside of get/finalize, so I cant see where you can pass a number of threads etc.

Hardcode84:
Setting global thread count is still possible via `global_control` but discouraged (especially if you have other libraries which can use `tbb` internally, lke `mkl` ).

Set desired concurrency as `task_arena` ctor parameter

henryiii:
Guessing this limitation is still needed for Numba 0.53, leaving the pin in the conda-forge recipe for now.

stuartarchibald:
@Hardcode84 am hoping to get this in for the next (0.54) release, any chance you could please rebase this on current mainline? Thanks.

Hardcode84:
Rebased

PR #6097, opened 2020-08-06
Add function code and closure bytes into cache key

stuartarchibald:
As title.

### DIFF
--- a/numba/core/caching.py
+++ b/numba/core/caching.py
@@ -697,14 +697,17 @@ class Cache(_Cache):
     def _index_key(self, sig, codegen):
         """
         Compute index key for the given signature and codegen.
-        It includes a description of the OS and target architecture.
+        It includes a description of the OS, target architecture and hashes of
+        the bytecode for the function and, if the function has a __closure__,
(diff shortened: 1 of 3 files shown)


stuartarchibald:
Opened for CI, see if this works. Trying to cache with a unique but reproducible key for closures.

numba/core/caching.py
```diff
@@ -697,7 +699,15 @@ def _index_key(self, sig, codegen):
         Compute index key for the given signature and codegen.
         It includes a description of the OS and target architecture.
         """
-        return (sig, codegen.magic_tuple())
+        codebytes = self._py_func.__code__.co_code
+        if self._py_func.__closure__ is not None:
+            cvars = tuple([x.cell_contents for x in self._py_func.__closure__])
+        else:
+            cvars = b''
```
sklam, line 706: why changing the type of `cvars`?

sklam: is that intended for `cvarbytes` instead?

stuartarchibald: fixed.

PR #6099, opened 2020-08-07
Restrict upper limit of TBB version due to ABI changes.

stuartarchibald:
As title.

xref https://github.com/numba/numba/pull/6096


stuartarchibald:
Buildfarm ID: `numba_smoketest_cpu_65`

stuartarchibald:
This passed on the farm with the exception of `aarch64` as the builder was having issues.

Issue #6100, opened 2020-08-07
Incorrect results on skylake with AVX512 and icc_rt=2018.0.2


stuartarchibald:
The upgrade to LLVM 10 for Numba 0.51.0rc1 and llvmlite 0.34.0rc1 appears to trigger a bug in older SVML libraries, specifically 2018.0.2 (and may be before then, later versions seem fine). This seems to impact chips with `AVX512` support only.

Reproducer:
```python
from numba import njit
import numpy as np
import math

@njit
def foo(a):
    ret = np.empty_like(a)
    for i in range(len(a)):
        z = a[i]
        x = z.real
        y = z.imag
        c = math.cos(y)
        s = math.sin(y)
        r = math.exp(x)
        ret[i] = complex(r * c, r * s)
    return ret


a = np.arange(4).astype('D')
print(foo.py_func(a))
print(foo(a))
```
Running in the following environments:

This is 0.51.0rc1 with icc_rt=2018.0.2
```
conda create -n _051rc1_icc201802 -c numba numba=0.51.0rc1 python=3.7 icc_rt=2018.0.2
```
gives:
```
$ python svml_bug.py
[ 1.        +0.j  2.71828183+0.j  7.3890561 +0.j 20.08553692+0.j]
[0.16666667+0.16666667j 0.16666667+0.16666667j 0.16666667+0.16666667j
 0.16666667+0.16666667j]
```
which is the bug, however if you do:
```
$ NUMBA_CPU_FEATURES='-prefer-256-bit' python svml_bug.py 
[ 1.        +0.j  2.71828183+0.j  7.3890561 +0.j 20.08553692+0.j]
[ 1.        +0.j  2.71828183+0.j  7.3890561 +0.j 20.08553692+0.j]
```
the bug isn’t present, so it’s almost certainly to do with the 256 bit intrinsics.

This is 0.51.0rc1 with icc_rt=2020.2
```
conda create -n _051rc1_icc20202 -c numba numba=0.51.0rc1 python=3.7 icc_rt=2020.2
```
gives:
```
$ python svml_bug.py 
[ 1.        +0.j  2.71828183+0.j  7.3890561 +0.j 20.08553692+0.j]
[ 1.        +0.j  2.71828183+0.j  7.3890561 +0.j 20.08553692+0.j]
```
which is correct.

Finally, this is 0.50.1 (the last official release) with icc_rt=2018.0.2 (same icc_rt as for the bug above)
```
[...]

PR #6101, opened 2020-08-07
Restrict lower limit of icc_rt version due to assumed SVML bug.

stuartarchibald:
Works around https://github.com/numba/numba/issues/6100


sklam:
This is ready as soon as buildfarm passes.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cpu_66`

stuartarchibald:
This passed on the farm with the exception of `aarch64` as the builder was having issues.

PR #6107, opened 2020-08-10
Fix and test #6095

### LINKED ISSUE #6095: numpy max for arrays of several dimensions not implemented for parallelized code
Hi, 
First of all: GREAT JOB!

I am running into what I think is a bug. This can be reproduced with the following code:

**import numpy as np
import numba as nb

@nb.jit(nopython=True, parallel=True)
def find_maxima_3D_jit():
    package = np.zeros((10, 10, 10)).astype(np.int16)
    for index in range(0, 10):
        z_stack = package[index, :, :]
        B = np.max(z_stack)
    return

find_maxima_3D_jit()**

The program does not crash when any of the following is true:
1- the option parallel=True is removed
2- the function np.max is applied on an array of one dimension: B = np.max(z_stack[0])
3- when the variable z_stack is defined before the loop (for example as a numpy array of zeros)

Thank you for your support!

sklam:
Fixes #6095 

The issue is caused by Parfor analysis being turned off because it was never tested. This patch simply re-enables it and adds test.

### DIFF
--- a/numba/parfors/parfor.py
+++ b/numba/parfors/parfor.py
@@ -2679,8 +2679,6 @@ def _find_mask(typemap, func_ir, arr_def):
                 mask_indices.append(None)
             # Handle integer index
             elif isinstance(index_typ, types.Integer):
-                # The follow is never tested
-                raise AssertionError('unreachable')
                 count_consts += 1
                 mask_indices.append(ind)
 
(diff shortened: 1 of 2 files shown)


sklam:
we should consider adding this to 0.51.0 given that it simply re-enables a path.

Micro-Sandworms:
Great, thank you. I appreciate it.

stuartarchibald:
> we should consider adding this to 0.51.0 given that it simply re-enables a path.

agree, added to milestone

PR #6116, opened 2020-08-11
Fix #6102. Problem with non-unique label.

### LINKED ISSUE #6102: parallel function compiles with v0.50.1 and fails with v0.51.0rc1
## Reporting a bug

I have a bit of parallelized code that compiled fine with numba v0.50.1, but fails with a LoweringError on v0.51.0rc1. Here is a chopped down bit of code that produces the error:

```python
import numpy as np


@njit(parallel=True)
def f(r):
    for ir in prange(r.shape[0]):
        dist = np.inf
        tr = np.array([0, 0, 0], dtype=np.float32)
        for i in [1, 0, -1]:
            dist_t = np.linalg.norm(r[ir, :] + i)
            if dist_t < dist:
                dist = dist_t
                tr = np.array([i, i, i], dtype=np.float32)
        r[ir, :] += tr
    return r


r = np.array([[0., 0., 0.], [0., 0., 1.]])
f(r)
```

I've tried to chop the code down further. Setting tr to a float instead of an array allows the code to compile. Removing the inner for loop and replacing it with `i = 0` allows the code to compile. Removing the if statement allows the code to compile. Replacing `prange` with `range` allows the code to compile.

Here are the outputs of `numba -s` for each of the environments (pass = v0.50.1, fail = v0.51.0rc1):
[fail.log](https://github.com/numba/numba/files/5044487/fail.log)
[pass.log](https://github.com/numba/numba/files/5044488/pass.log)

Lastly, here is the traceback that is produced on my machine:
[error.log](https://github.com/numba/numba/files/5044495/error.log)

sklam:
Fixes  #6102.

Fixed by updating the _max_label.

### DIFF
--- a/numba/parfors/parfor.py
+++ b/numba/parfors/parfor.py
@@ -2700,6 +2700,9 @@ class ParforPass(ParforPassStates):
     def _pre_run(self):
         # run array analysis, a pre-requisite for parfor translation
         self.array_analysis.run(self.func_ir.blocks)
+        # NOTE: Prepare _max_label. See #6102
+        ir_utils._max_label = max(ir_utils._max_label,
+                                  ir_utils.find_max_label(self.func_ir.blocks))
 
     def run(self):
         """run parfor conversion pass: replace Numpy calls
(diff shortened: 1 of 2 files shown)


mturiansky:
Thanks for the quick fix! Numba is awesome, and you all are awesome!

esc:
@mturiansky thanks for the compliments, we appreciate it! 🙏

stuartarchibald:
Thanks for your support @mturiansky :)

Issue #5234, opened 2020-02-11
Numba `DeviceNDArray` doesn't handle `shape=()` correctly


jakirkham:
## Reporting a bug

- [x] I am using the latest released version of Numba (most recent is visible in
 the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [x] I have included below a minimal working reproducer (if you are unsure how
 to write one see http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports).

<hr>

When trying to create Numba `DeviceNDArray`s with `shape=()`, it turns the shape into `(1,)` instead. This differs from behavior seen in NumPy.

```python
In [1]: import numpy                                                            

In [2]: import numba.cuda                                                       

In [3]: numpy.ndarray(()).shape                                                 
Out[3]: ()

In [4]: numba.cuda.device_array(()).shape                                       
Out[4]: (1,)

In [5]: numba.cuda.device_array_like(numpy.ndarray(())).shape                   
Out[5]: (1,)
```

stuartarchibald:
Thanks for the report, I can reproduce. Problem comes from here: https://github.com/numba/numba/blob/f3ca34da03b03c62a27daacfbddf63f7d72e7cee/numba/cuda/api.py#L208-L209

Suggested fix might be:
```
diff --git a/numba/cuda/api.py b/numba/cuda/api.py
index 39da6632f..492a84af3 100644
--- a/numba/cuda/api.py
+++ b/numba/cuda/api.py
@@ -205,14 +205,14 @@ def _prepare_shape_strides_dtype(shape, strides, dtype, order):
     if isinstance(strides, (int, long)):
         strides = (strides,)
     else:
-        if shape == ():
-            shape = (1,)
         strides = strides or _fill_stride_by_order(shape, dtype, order)
     return shape, strides, dtype
 
 
 def _fill_stride_by_order(shape, dtype, order):
     nd = len(shape)
+    if nd == 0:
+        return ()
     strides = [0] * nd
     if order == 'C':
         strides[-1] = dtype.itemsize
```

eric-wieser:
I think this was fixed in #5359

gmarkall:
I'm not sure this is fixed - with:

```python
from numba import cuda
import numpy as np

print(np.ndarray(()).shape)

print(cuda.device_array(()).shape)

print(cuda.device_array_like(np.ndarray(())).shape)
```

I get:

```
()
(1,)
(1,)
```

on the latest master.

[...]

PR #6122, opened 2020-08-11
Make `strides is None` C-contiguous

jakirkham:
If the user provides `strides` that is `None`, replace this with C-contiguous striding. This behavior matches both NumPy and CuPy.

### DIFF
--- a/numba/cuda/cudadrv/devicearray.py
+++ b/numba/cuda/cudadrv/devicearray.py
@@ -87,7 +87,12 @@ class DeviceNDArrayBase(object):
             strides = (strides,)
         dtype = np.dtype(dtype)
         self.ndim = len(shape)
-        if len(strides) != self.ndim:
+        if strides is None:
+            strides = [dtype.itemsize]
+            for s in shape[:0:-1]:
+                strides.insert(0, s * strides[0])
+            strides = tuple(strides)
+        elif len(strides) != self.ndim:
             raise ValueError('strides not match ndim')
         self._dummy = dummyarray.Array.from_desc(0, shape, strides,
                                                  dtype.itemsize)


gmarkall:
Many thanks for the PR! If I were constructing a device array without providing strides, I'd normally do it like this:

```python
from numba import cuda
arr = cuda.device_array((3, 4, 5))
print(arr.is_c_contiguous())  # "True"
```

Prior to this PR, I don't think passing `None` for strides directly to a `DeviceArray` constructor would even be allowed:

```python
from numba import cuda
cuda.cudadrv.devicearray.DeviceNDArray((3, 4, 5), None, None)
# TypeError: object of type 'NoneType' has no len()
```

Can you outline the use case that this PR fixes please?

stuartarchibald:
This could also do with a test or two adding please before it goes in, thanks!

jakirkham:
Sure Graham, in Dask we serialize objects [using metadata from `__cuda_array_interface__`]( https://github.com/dask/distributed/blob/5e7285cb200d461034a439274d2a5db87c3df9e7/distributed/protocol/numba.py#L24 ). One of the fields we grab is `strides`, which is often set to `None` if the array is C-contiguous. This can cause us issues in deserialization if the array constructor (in the Numba case, `DeviceNDArray`) cannot [take `strides=None`]( https://github.com/dask/distributed/blob/5e7285cb200d461034a439274d2a5db87c3df9e7/distributed/protocol/numba.py#L44 ). Currently we workaround this by [overwriting the `strides`]( https://github.com/dask/distributed/blob/5e7285cb200d461034a439274d2a5db87c3df9e7/distributed/protocol/numba.py#L25 ) field with the `strides` of the array (and need to do this for all other serialized types as well for interop with Numba). However it would be nice to ensure Numba can handle this case in `DeviceNDArray` to avoid the need for these workarounds.

jakirkham:
Does that answer your question @gmarkall? If so, can you please advise where we should add a test?

gmarkall:
> Does that answer your question @gmarkall? If so, can you please advise where we should add a test?

It does answer my question, but I'm thinking that this fix is not quite the one I'd want, as it duplicates a subset of the logic in `numba/cuda/api.py` in `_prepare_shape_strides_dtype` and `_fill_strides_by_order`. Up until now, I had always expected that end users would construct device arrays by calling:

```python
arr = cuda.device_array(...)
```

or similar, and I would have discouraged the direct use of the `DeviceNDArray` array constructor. However, I looked at the serialization code that you linked to, and I do see that it would be odd to serialize a `DeviceNDArray` and then deserialize by calling `cuda.device_array()` instead of the constructor for `DeviceNDArray`, and this also wouldn't allow `gpu_data` to be passed in.

So, I think the right way to go about fixing these APIs is to hoist the logic from `cuda.api.device_array` into `DeviceNDArrayBase.__init__`, in particular:

- Move `_prepare_shape_strides_dtype` and `_fill_strides_by_order` somewhere common and remove the leading underscore (because they'll be used by `DeviceNDArrayBase` and `cuda.api.pinned_array` / `mapped_array`.
- Change `DeviceNDArrayBase.__init__` so that it also accepts the `order` kwarg, and give both `strides` and `dtype` default values (`None` and `np.float64`, respectively)
- Call `prepare_shape_strides_dtype` in `DeviceNDArrayBase` instead of in `cuda.api.device_array`

This way, `cuda.api.device_array` will only create a `DeviceNDArray` and return it, but will need leaving in for compatibility with existing code.

For testing, an appropriate place is probably `numba/cuda/tests/cudadrv/test_cuda_ndarray.py`.

Note also that there is presently a bug in `_prepare_shape_strides_dtype` and `_fill_strides_by_order` when `shape=()` - you might also want to change the logic according to @stuartarchibald's suggested fix (https://github.com/numba/numba/issues/5234#issuecomment-586961127) at the same time.

Does this make sense? Let me know if anything needs further clarification.

PR #6123, opened 2020-08-12
Fix #6121

### LINKED ISSUE #6121: CUDA: Recent commit causes infinite loop (or a lot of recompiling)
Looking into an issue using 0.51.0rc1 with cuDF, it seems that this commit: https://github.com/numba/numba/commit/e203c9d8c00590882fb50a9f0e70d09f255bb4ab causes an infinite loop, or a lot of recompilation. I'm investigating this now and will post further details here as I progress.

gmarkall:
Opening this PR to fix #6121 and iterate on it. The problem is that `forall` always returns a new `Dispatcher`, which immediately compiles the specialization. Prior to #5709 being merged, the cache of compilations was used and any previously-compiled kernel could be returned by `specialize`. The fix in this branch introduces a cache of specialized `Dispatcher` objects to work around the issue.

### DIFF
--- a/numba/cuda/compiler.py
+++ b/numba/cuda/compiler.py
@@ -366,7 +366,10 @@ class ForAll(object):
         if self.ntasks == 0:
             return
 
-        kernel = self.kernel.specialize(*args)
+        if self.kernel.specialized:
+            kernel = self.kernel
+        else:
+            kernel = self.kernel.specialize(*args)
         blockdim = self._compute_thread_per_block(kernel)
         griddim = (self.ntasks + blockdim - 1) // blockdim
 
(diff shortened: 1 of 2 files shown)


stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_92` for 86241da

stuartarchibald:
`numba_smoketest_cuda_92` passed, will try 4f17d85

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_93` for 4f17d85

stuartarchibald:
`numba_smoketest_cuda_93` passed.

gmarkall:
@stuartarchibald As far as I can tell from testing cuDF and other libraries, the issue appears to be resolved and I can't see that this introduces any new issues.

stuartarchibald:
> @stuartarchibald As far as I can tell from testing cuDF and other libraries, the issue appears to be resolved and I can't see that this introduces any new issues.

Thanks for confirming. This will make 0.51.0.

Issue #6125, opened 2020-08-12
IndexError in _version_info with build from master 


rpopovici:
I get this error with custom build from latest master(today's commits)

Steps to reproduce:
1. uninstall numba and llvmlite
2. download numba-master.zip and unzip to folder of your choosing
3. run: python setup.py install
4. install numba master + llvm 0.33 or 0.34 rc
5. load any script, e.g:
```
import numba as nb
print('test')
```
6. IndexError..

rpopovici:
Error report:
```
IndexError                                Traceback (most recent call last)
 in 
----> 1 import numba as nb

~\AppData\Local\Programs\Python\Python37\lib\site-packages\numba-0+unknown-py3.7-win-amd64.egg\numba\__init__.py in 
     33     return _version_info(major, minor, patch, short, full, string, tup,
     34                          git_revision)
---> 35 version_info = _gen_v_info()
     36 del _gen_v_info
     37 

~\AppData\Local\Programs\Python\Python37\lib\site-packages\numba-0+unknown-py3.7-win-amd64.egg\numba\__init__.py in _gen_v_info()
     24             return None
     25     major = try_int(parts[0])
---> 26     minor = try_int(parts[1])
     27     patch = try_int(parts[2])
     28     short = (major, minor)

IndexError: list index out of range
```

esc:
@rpopovici thanks for reporting this, would you be able to attach a debugger and print the value of `parts`?

gmarkall:
Looks like the version is `0+unknown`, which will be causing the issue - it's assumed in `_gen_v_info()` that the version can be split on dots into at least three parts. (IIRC this version number can occur if versioneer couldn't get the info using git).

esc:
May I ask, how was Numba installed?

rpopovici:
@esc downloaded the zip from github, unzip, run the install script.. I didn't use git clone if that's what you are asking

rpopovici:
@gmarkall the version is indeed 0+unknown

esc:
@rpopovici thank you! I can reproduce this on a git checkout, after removing all tags with:

`git tag | xargs git tag -d``

It is correct that `versioneer.py` will traverse the git history to look for the most recent tag and use that to determine the version number (uses `git describe` internally). 

The fix will be to check if the version number can be split and act accordingly if it can't.

esc:
I think in this case, a `version_info` of:

[...]

Issue #5973, opened 2020-07-09
Segfault in TBB, maybe parfors related.


stuartarchibald:
## Reporting a bug

There's a problem in this branch https://github.com/stuartarchibald/numba/tree/pr_5463_segfault, head is c84bfbd, it's a clone of PR #5463 which is segfaulting on CI. It's reproducible locally with:
```
./runtests.py -m 1 -v -b -j "2,None,18" --exclude-tags='long_running' numba.tests
```
and in fact any `-m <X>` will segfault. However, this:
```
./runtests.py -v -b -j "2,None,18" --exclude-tags='long_running' numba.tests
```
will not. The difference here being that the use of the multiprocessing test runner (or not, as in the latter case).

Backtrace is:
```
test_issue4963_globals (numba.tests.test_parfors.TestParfors) ... 
Program received signal SIGSEGV, Segmentation fault.
tbb::internal::generic_scheduler::allocate_task (this=this@entry=0x7fffd60b2400, number_of_bytes=number_of_bytes@entry=104, 
    parent=parent@entry=0x0, context=0x7fffffff8b80) at ../../src/tbb/scheduler.cpp:357
357     ../../src/tbb/scheduler.cpp: No such file or directory.
(gdb) bt
#0  tbb::internal::generic_scheduler::allocate_task (this=this@entry=0x7fffd60b2400, 
    number_of_bytes=number_of_bytes@entry=104, parent=parent@entry=0x0, context=0x7fffffff8b80)
    at ../../src/tbb/scheduler.cpp:357
#1  0x00007fffdbe6363b in tbb::internal::allocate_root_with_context_proxy::allocate (this=0x7fffffff8b78, size=104)
    at ../../src/tbb/task.cpp:64
#2  0x00007fffdbe3cd08 in tbb::interface7::internal::delegated_function<parallel_for(void*, char**, unsigned long*, unsigned long*, void*, unsigned long, unsigned long, int)::{lambda()#1} const, void>::operator()() const ()
   from <numba_path>/numba/numba/np/ufunc/tbbpool.cpython-37m-x86_64-linux-gnu.so
#3  0x00007fffdbe69250 in tbb::interface7::internal::task_arena_base::internal_execute (this=0x7fffffff8f50, d=...)
    at ../../src/tbb/arena.cpp:1054
#4  0x00007fffdbe3d60e in parallel_for(void*, char**, unsigned long*, unsigned long*, void*, unsigned long, unsigned long, int)
[...]

PR #6126, opened 2020-08-12
fix version_info if version can not be determined

esc:
The current implementation assumes that the version info can be split
into at least three parts using the period as delimiter. In certain
situations, such as running from a dowloaded release archive, or when
building from a git checkout that does not have any tags, this
assumption does not hold.

Fixes: #6125


esc:
I considered some tests for this function, but then was discouraged, as `_gen_v_info` is deleted again after having been defined and used and because the function is in `numba/__init__.py` I am not sure it is worth the effort to mock the `del` in that module during testing. I was unsure what the intent of this is, @stuartarchibald can you clarify? I think I would prefer to keep the function as a utility and additionally add something like `myversion` as a parameter so that it can operate on arbitrary versions rather than always getting the `__version__` from `numba`. That would also significantly ease testing as nothing would need to be mocked.

stuartarchibald:
> I considered some tests for this function, but then was discouraged, as `_gen_v_info` is deleted again after having been defined and used and because the function is in `numba/__init__.py` I am not sure it is worth the effort to mock the `del` in that module during testing. I was unsure what the intent of this is, @stuartarchibald can you clarify?

Didn't want to pollute the top level namespace.

> I think I would prefer to keep the function as a utility and additionally add something like `myversion` as a parameter so that it can operate on arbitrary versions rather than always getting the `__version__` from `numba`. That would also significantly ease testing as nothing would need to be mocked.

Could the function be moved into `numba.misc.init` or something, imported in `__init__`, run and results stored, then deleted. This keeps the top level cleaner and also gives opportunity to test? It'd also present an opportunity to get a load of the other functions out of `__init__` and somewhere testable.

esc:
This has now been updated and is ready for the first round of reviews.

numba/__init__.py
```diff
@@ -6,34 +6,15 @@
 import re
 import sys
 import warnings
-from collections import namedtuple
 
 from ._version import get_versions
+from .misc.init_utils import generate_version_info
```
stuartarchibald, line 11:
```suggestion
from numba.misc.init_utils import generate_version_info
```

numba/misc/init_utils.py
```diff
@@ -0,0 +1,25 @@
+from collections import namedtuple
```
stuartarchibald, line 1: Module docstring?

numba/misc/init_utils.py
```diff
@@ -0,0 +1,25 @@
...
+
+version_info = namedtuple('version_info',
+                          ('major minor patch short full '
+                           'string tuple git_revision'))
+
+
+def generate_version_info(version):
```
stuartarchibald, line 8: Docstring?

numba/tests/test_inti_utils.py
```diff
@@ -0,0 +1,42 @@
+import unittest
```
stuartarchibald, line 1: This test module has a typo. s/inti/init/

stuartarchibald (commented):
Thanks for the patch, just a few minor things to resolve else good to go.

stuartarchibald:
https://github.com/numba/numba/issues/6167 is probably the cause of CI failure.

numba/misc/init_utils.py
esc: Sure

numba/misc/init_utils.py
esc: sure

numba/tests/test_inti_utils.py
esc: indeed

esc:
Changes have been made and the PR is ready for the next round of reviews.

numba/misc/init_utils.py
```diff
@@ -6,6 +8,23 @@
 
 
 def generate_version_info(version):
+    """Process a version string into a structured version_info object.
+
+    Parameters
+    ----------
+    version: str
+        a string describing
```
stuartarchibald, line 16:
```suggestion
        a string describing the version
```

numba/misc/init_utils.py
```diff
@@ -6,6 +8,23 @@
...
+
+    Returns
+    -------
+    version_info: tuple
+        structured version information
+
+    See also
+    --------
+    Look at the definition of 'version_info' in this file for details.
```
stuartarchibald, line 25:
```suggestion
    Look at the definition of 'version_info' in this module for details.
```

stuartarchibald (approved):
Thanks for the fixes.

stuartarchibald:
CI fail is unrelated: https://github.com/numba/numba/issues/5973

PR #6127, opened 2020-08-12
Fix race in reduction kernels on Volta, require CUDA 9, add syncwarp with default mask

### LINKED ISSUE #2589: Eliminate warp synchronous code from CUDA kernels
Warp synchronous code has been discouraged for several CUDA releases, and now with CUDA 9 and Volta, threads in a warp are no longer guaranteed to execute together.  I found one place in Numba where kernels are generated that assume warp synchronicity:

https://github.com/numba/numba/blob/1a4cf39e982dd565be08ec5eae63989da3ce3aa1/numba/cuda/kernels/reduction.py

I don't know if there are others.  We need to change this kernel to work on Volta.

gmarkall:
This PR contains the fixes and improvements from #6112 without actually making indices unsigned. The main aim of this PR is therefore to fix the race conditions in the reduction kernels, but there are several changes required to do this:

- Synchronization of warps in reductions. The reduction kernels make an assumption that writes to shared memory by threads in a warp are automatically visible to other threads in the warp. On Volta and beyond, this is no longer true. This is fixed by adding `syncwarp()` at appropriate points in the reduction kernel. Originally I used a partial mask containing only the threads currently participating in the reduction, but this is only supported on CC 7.0 and above, so I've left it as a potential future optimization to be applied only on CC >= 7.0 devices.
- Addition of the `syncwarp()` function with no arguments, which assumes a full mask. This was not essential, but provides a nicer interface than always having to write `syncwarp(0xFFFFFFFF)`. Some tests for `syncwarp()` are also added.
- Require CUDA 9.0. `syncwarp()` etc. were only introduced in CUDA 9.0, and dropping CUDA 8.0 seems reasonable at this point in time.
- Small fixes to the code in `nvvm.py` that checks the supported compute capabilities, and addition of an error when one attempts to use an unspported NVVM version, alongside a warning at import time for `numba.cuda` that fires if an unsupported version is detected.

Fixes #2589.


gmarkall:
Following discussion in the triage meeting, I'll add some wording to the docs about the policy for supporting older CUDA releases.

gmarkall:
> Following discussion in the triage meeting, I'll add some wording to the docs about the policy for supporting older CUDA releases.

And also make a post about intentions on Discourse.

gmarkall:
> And also make a post about intentions on Discourse.

Discourse post: https://numba.discourse.group/t/plans-to-drop-support-for-cuda-8-in-0-52-and-general-support-policy/205

gmarkall:
> Following discussion in the triage meeting, I'll add some wording to the docs about the policy for supporting older CUDA releases.

A note is now added to the documentation.

numba/cuda/cudaimpl.py
```diff
@@ -245,8 +245,15 @@ def ptx_threadfence_device(context, builder, sig, args):
...
-@lower(stubs.syncwarp, types.i4)
-def ptx_warp_sync(context, builder, sig, args):
+@lower(stubs.syncwarp)
+def ptx_syncwarp(context, builder, sig, args):
+    mask = context.get_constant(types.uint32, 0xFFFFFFFF)
+    mask_sig = signature(types.none, types.uint32)
+    return ptx_syncwarp_mask(context, builder, mask_sig, [mask])
+
+
+@lower(stubs.syncwarp, types.u4)
```
stuartarchibald, line 255: Why the removal of the signed mask?

numba/cuda/cudadrv/nvvm.py
```diff
@@ -295,12 +296,15 @@ def get_supported_ccs():
...
         _supported_cc = ()
     elif cudart_version_major < 9:
-        # CUDA 8.x
-        _supported_cc = (2, 0), (2, 1), (3, 0), (3, 5), (5, 0), (5, 2), (5, 3), (6, 0), (6, 1), (6, 2) # noqa: E501
-    elif cudart_version_major < 10:
+        _supported_cc = ()
+        ctk_ver = f"{cudart_version_major}.{cudart_version_minor}"
+        msg = f"CUDA Toolkit {ctk_ver} is unsupported by Numba - 9.0 is the " \
+              + "minimum required version."
+        warnings.warn(msg)
```
stuartarchibald, line 303: Is this on the import path for `numba.cuda`? If so perhaps don't use `fstrings` as the error message gets lost for users on python's with no fstring support (e.g. someone accidentally using the unsupported Python 2 or <=3.5).

stuartarchibald:
This probably needs an update?
https://github.com/numba/numba/blob/1949c62e36a685b834f9b18da48dfa9a8f48f930/numba/cuda/tests/cudadrv/test_runtime.py#L25-L26

numba/cuda/cudadrv/nvvm.py
gmarkall: It's not on the import path - if it we're we'd have the much bigger problem of #6149. (fixed in #6147)

numba/cuda/cudaimpl.py
gmarkall: Because its argument is a bitmask, I felt that it's more natural for the argument to be unsigned, and that passing a signed value might be indicative of a logic error in the user's code.

This also makes Numba more consistent with CUDA C/C++, where the argument to `__syncwarp` is unsigned: https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#synchronization-functions

gmarkall:
> This probably needs an update?
> 
> https://github.com/numba/numba/blob/1949c62e36a685b834f9b18da48dfa9a8f48f930/numba/cuda/tests/cudadrv/test_runtime.py#L25-L26

It does - updated now.

numba/cuda/cudaimpl.py
stuartarchibald: I entirely agree that bitmasks are naturally unsigned, just wonder if such a change might brake in prod user code that assumes signed?

stuartarchibald:
> > This probably needs an update?
> > https://github.com/numba/numba/blob/1949c62e36a685b834f9b18da48dfa9a8f48f930/numba/cuda/tests/cudadrv/test_runtime.py#L25-L26
> 
> It does - updated now.

Thanks!

numba/cuda/cudaimpl.py
gmarkall: I haven't been able to work out a scenario where it would, but to be conservative I've changed it back to signed. If the masks are to change to unsigned (which I would like at some point in the future) then it makes more sense to change them all at once, and not just for one intrinsic, along with some careful testing.

gmarkall:
@stuartarchibald is this just waiting for a buildfarm run now?

gmarkall:
Conflicts now resolved. Could this have a buildfarm run please @stuartarchibald ?

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_112`.

numba/cuda/cudadrv/nvvm.py
```diff
@@ -319,6 +323,8 @@ def find_closest_arch(mycc):
     :return: Closest supported CC as a tuple ``(MAJOR, MINOR)``
     """
     supported_cc = get_supported_ccs()
+    if not supported_cc:
+        raise NvvmSupportError("No supported CUDA Toolkit found")
```
stuartarchibald, line 327: This is a duplicate of the condition 2 lines down?

stuartarchibald:
Discussed out of band, the build farm is failing. Turns out there's some issue to do with running the tests in the `multiprocessing` based test runner as invoked with e.g:
```
./runtests.py -m 4 numba.cuda.tests  -f
```

numba/cuda/cudadrv/nvvm.py
gmarkall: Yeah, it got messed up in a merge - I did something very similar here to #6227 here. Will fix.

gmarkall:
> Discussed out of band, the build farm is failing. Turns out there's some issue to do with running the tests in the `multiprocessing` based test runner as invoked with e.g:
> 
> ```
> ./runtests.py -m 4 numba.cuda.tests  -f
> ```

This should now be fixed - see the commit message on 9c29bc6.

I've also removed the duplicate exception raise, which came from a messed-up merge somewhere.

I think this is ready for another buildfarm run - many thanks in advance!

stuartarchibald:
> > Discussed out of band, the build farm is failing. Turns out there's some issue to do with running the tests in the `multiprocessing` based test runner as invoked with e.g:
> > ```
> > ./runtests.py -m 4 numba.cuda.tests  -f
> > ```
> 
> This should now be fixed - see the commit message on [9c29bc6](https://github.com/numba/numba/commit/9c29bc632e79bb12ac09eb657b4fd0f419525c7e).

Well found!

> I've also removed the duplicate exception raise, which came from a messed-up merge somewhere.

Thanks.

> I think this is ready for another buildfarm run - many thanks in advance!

Buildfarm ID: `numba_smoketest_cuda_114`

stuartarchibald (approved):
Thanks for making this version migration and fixing up. The new default `syncwarp` is a good addition too. This patch is approved pending CI pass.

gmarkall:
Per OOB discussion, another commit is added here to seed the RNG for `test_round_to_{f4,f8}` to make them stable.

stuartarchibald (approved):
Fix for RNG stability is also approved. Thanks.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_115`

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_115`

Passed all builds except CUDA 8 which is expected given this PR enforces CUDA 9 as a minimum.

PR #6139, opened 2020-08-17
Use direct syntax instead of the `add_metaclass` decorator

eric-wieser:


esc:
@eric-wieser thanks for submitting this, I have added it to the review queue.

stuartarchibald (approved):
Thanks for the patch!

PR #6140, opened 2020-08-17
Replace calls to utils.iteritems(d) with d.items()

eric-wieser:
This looks like a holdover from Python 2 support.


stuartarchibald (approved):
Thanks for the patch!

PR #6141, opened 2020-08-17
Fix #6130 objmode cache segfault

### LINKED ISSUE #6130: Cache does not work when passing numpy arrays to objmode inside overload
## Reporting a bug

Hello, I found this bug with `cache=True` when using overloads. Replicator is below:

```python
import numpy as np
import numba


def print_this(a, b):
    print(a)
    print(b)


def do_something(a, b):
    print(a)
    print(b)


@overload(do_something)
def overload_do_something(a, b):
    def _do_something_impl(a, b):
        with numba.objmode:
            print_this(a, b)

    return _do_something_impl


@numba.njit(cache=True)
def test_caching():
    a = np.arange(20)
    b = np.arange(50)
    #a = 1  # works if pass this instead of numpy array
    #b = 3
    do_something(a, b)


test_caching()

```

Thanks!

sklam:
Fixes #6130 

Thanks @juanjgalvez and @stuartarchibald for finding out the source of the problem on the lost `env`.

The patch makes sure the `CompileResult` acquires a reference to any referenced `Environment` objects. Changes to `numba/core/compiler.py` is the actual fix. The other non-test changes are for better error handling when an NULL pointer is encountered.

### DIFF
--- a/numba/_helperlib.c
+++ b/numba/_helperlib.c
@@ -276,8 +276,8 @@ numba_recreate_record(void *pdata, int size, PyObject *dtype) {
     PyArray_Descr *descr = NULL;
 
     if (dtype == NULL) {
-        PyErr_Format(PyExc_ValueError,
-            "In 'numba_recreate_record', 'descr' is NULL");
+        PyErr_Format(PyExc_Runtime,
+            "In 'numba_recreate_record', 'dtype' is NULL");
         return NULL;
     }
 
--- a/numba/_helperlib.c
+++ b/numba/_helperlib.c
@@ -276,7 +276,7 @@ numba_recreate_record(void *pdata, int size, PyObject *dtype) {
     PyArray_Descr *descr = NULL;
 
     if (dtype == NULL) {
-        PyErr_Format(PyExc_Runtime,
+        PyErr_Format(PyExc_RuntimeError,
             "In 'numba_recreate_record', 'dtype' is NULL");
         return NULL;
     }
--- a/numba/_helperlib.c
+++ b/numba/_helperlib.c
@@ -275,6 +275,12 @@ numba_recreate_record(void *pdata, int size, PyObject *dtype) {
     PyObject *record = NULL;
     PyArray_Descr *descr = NULL;
 
+    if (dtype == NULL) {
+        PyErr_Format(PyExc_ValueError,
+            "In 'numba_recreate_record', 'descr' is NULL");
+        return NULL;
+    }
+
     numpy = PyImport_ImportModuleNoBlock("numpy");
     if (!numpy) goto CLEANUP;
 
--- a/numba/core/pythonapi.py
+++ b/numba/core/pythonapi.py
@@ -146,6 +146,7 @@ class EnvironmentManager(object):
                 getitem = self.pyapi.list_getitem(consts, index)
                 builder.store(getitem, ret)
             with br_null:
+                # This can happen when the Environment is accidentally released
                 self.pyapi.err_set_string(
                     "PyExc_RuntimeError",
                     "`env.consts` is NULL in `read_const`",
--- a/numba/core/pythonapi.py
+++ b/numba/core/pythonapi.py
@@ -147,6 +147,7 @@ class EnvironmentManager(object):
                 builder.store(getitem, ret)
             with br_null:
                 # This can happen when the Environment is accidentally released
+                # and has subsequently been garbage collected.
                 self.pyapi.err_set_string(
                     "PyExc_RuntimeError",
                     "`env.consts` is NULL in `read_const`",
--- a/numba/core/pythonapi.py
+++ b/numba/core/pythonapi.py
@@ -134,7 +134,13 @@ class EnvironmentManager(object):
         """
         assert index < len(self.env.consts)
 
-        return self.pyapi.list_getitem(self.env_body.consts, index)
+        builder = self.pyapi.builder
+        consts = self.env_body.consts
+        ret = cgutils.alloca_once(builder, self.pyapi.pyobj, zfill=True)
+        with builder.if_then(cgutils.is_not_null(builder, consts)):
+            getitem = self.pyapi.list_getitem(consts, index)
+            builder.store(getitem, ret)
+        return builder.load(ret)
 
 
 _IteratorLoop = namedtuple('_IteratorLoop', ('value', 'do_break'))
--- a/numba/core/pythonapi.py
+++ b/numba/core/pythonapi.py
@@ -131,6 +131,9 @@ class EnvironmentManager(object):
         """
         Look up constant number *index* inside the environment body.
         A borrowed reference is returned.
+
+        The returned LLVM value may have NULL value at runtime which indicates
+        an error at runtime.
         """
         assert index < len(self.env.consts)
 
--- a/numba/core/pythonapi.py
+++ b/numba/core/pythonapi.py
@@ -140,9 +140,16 @@ class EnvironmentManager(object):
         builder = self.pyapi.builder
         consts = self.env_body.consts
         ret = cgutils.alloca_once(builder, self.pyapi.pyobj, zfill=True)
-        with builder.if_then(cgutils.is_not_null(builder, consts)):
-            getitem = self.pyapi.list_getitem(consts, index)
-            builder.store(getitem, ret)
+        with builder.if_else(cgutils.is_not_null(builder, consts)) as \
+                (br_not_null, br_null):
+            with br_not_null:
+                getitem = self.pyapi.list_getitem(consts, index)
+                builder.store(getitem, ret)
+            with br_null:
+                self.pyapi.err_set_string(
+                    "PyExc_RuntimeError",
+                    "`env.consts` is NULL in `read_const`",
+                )
         return builder.load(ret)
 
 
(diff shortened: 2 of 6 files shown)


numba/tests/support.py
```diff
@@ -19,6 +19,7 @@
 import ctypes
 import multiprocessing as mp
 import warnings
+import traceback
```
stuartarchibald, line 22: unused import.

numba/core/runtime/_nrt_python.c
```diff
@@ -295,6 +295,11 @@ NRT_adapt_ndarray_to_python(arystruct_t* arystruct, int ndim,
     npy_intp *shape, *strides;
     int flags = 0;
 
+    if (descr == NULL) {
+        PyErr_Format(PyExc_TypeError, "descr == NULL");
```
stuartarchibald, line 299:
```suggestion
        PyErr_Format(PyExc_ValueError, "In 'NRT_adapt_ndarray_to_python', 'descr' is NULL");
```

numba/core/pythonapi.py
```diff
@@ -134,7 +134,13 @@ def read_const(self, index):
         """
         assert index < len(self.env.consts)
 
-        return self.pyapi.list_getitem(self.env_body.consts, index)
+        builder = self.pyapi.builder
+        consts = self.env_body.consts
+        ret = cgutils.alloca_once(builder, self.pyapi.pyobj, zfill=True)
+        with builder.if_then(cgutils.is_not_null(builder, consts)):
+            getitem = self.pyapi.list_getitem(consts, index)
```
stuartarchibald, line 141: this is unchecked, whilst unlikely, could raise e.g. IndexError?

numba/core/pythonapi.py
```diff
@@ -134,7 +134,13 @@ def read_const(self, index):
         """
         assert index < len(self.env.consts)
 
-        return self.pyapi.list_getitem(self.env_body.consts, index)
+        builder = self.pyapi.builder
+        consts = self.env_body.consts
+        ret = cgutils.alloca_once(builder, self.pyapi.pyobj, zfill=True)
+        with builder.if_then(cgutils.is_not_null(builder, consts)):
```
stuartarchibald, line 140: If `consts` is NULL, what does this imply? Has something gone wrong if something is trying to index a nullptr? If so it ought be caught?

numba/tests/support.py
```diff
@@ -779,6 +780,26 @@ def run_in_new_process_caching(func, cache_dir_prefix=__name__, verbose=True):
...
 
+    Returns
+    -------
+    ret : dict
+        exitcode: 0 for success. 1 for exception-raised.
+        stdout: str
+        stderr: str
+    """
+    cache_dir = temp_directory(cache_dir_prefix)
+    return run_in_new_process_in_cache_dir(func, cache_dir, verbose=True)
```
stuartarchibald, line 791:
```suggestion
    return run_in_new_process_in_cache_dir(func, cache_dir, verbose=verbose)
```

numba/tests/test_extending.py
```diff
@@ -1748,12 +1749,56 @@ def testcase(x):
...
+            a = np.arange(20)
+            b = np.arange(20)
+            return do_something(a, b)
+
+        got = test_caching()
+        expect = test_caching.py_func()
+
+        # Check result
+        if got != expect:
+            raise AssertionError("incorrect result")
```
stuartarchibald, line 1779:
```suggestion
        self.assertPreciseEqual(got, expected)
```

numba/tests/test_extending.py
```diff
@@ -1748,12 +1749,56 @@ def testcase(x):
...
+        return test_caching
+
+    @classmethod
+    def check_objmode_cache_ndarray_check_cache(cls):
+        disp = cls.check_objmode_cache_ndarray()
+        if len(disp.stats.cache_misses) != 0:
+            raise AssertionError('unexpected cache miss')
+        if len(disp.stats.cache_hits) <= 0:
+            raise AssertionError("unexpected missing cache hit")
```
stuartarchibald, line 1788: Maybe use `self.assert*` here?

numba/tests/test_extending.py
```diff
@@ -1748,12 +1749,56 @@ def testcase(x):
...
+
+    def test_check_objmode_cache_ndarray(self):
+        # See issue #6130.
+        # Env is missing after cache load.
+        cache_dir = temp_directory(self.__class__.__name__)
```
stuartarchibald, line 1793: Whilst unlikely to get a tmpdir collision, should this be run as a serial test to ensure it can't happen?

stuartarchibald (commented):
Thanks for the patch, great to see this fixed.

numba/tests/support.py
sklam: It was a missing import that I added for
https://github.com/numba/numba/blob/0d8266613e568b344c13a8d2ee3937e9510d1b49/numba/tests/support.py#L841

numba/core/pythonapi.py
sklam: In that case `getitem` would be `NULL` and would be returned to the caller. The caller can then check for the NULL return. There isn't much this function can do.

Caller are handling it like:
- https://github.com/sklam/numba/blob/0d8266613e568b344c13a8d2ee3937e9510d1b49/numba/core/pylowering.py#L425-L426
- https://github.com/numba/numba/pull/6141/files/0d8266613e568b344c13a8d2ee3937e9510d1b49#diff-1e3a0be1e38fdb57e5bbff6f233dd1da
- I will make sure `numba_recreate_record` is checking the `dtype` argument

numba/core/pythonapi.py
sklam: If it is NULL, it is probably `env` got removed. I can make it set an exception.

numba/tests/test_extending.py
sklam: It's a function that cannot be bound to the instance of `TestCase` because `TestCase` instances cannot be pickled, which is needed for the IPC used later.

numba/tests/test_extending.py
sklam: same reason as above.

numba/tests/test_extending.py
sklam: `temp_directory()` is enough to make it safe from path collisions

numba/core/runtime/_nrt_python.c
```diff
@@ -295,6 +295,11 @@ NRT_adapt_ndarray_to_python(arystruct_t* arystruct, int ndim,
     npy_intp *shape, *strides;
     int flags = 0;
 
+    if (descr == NULL) {
+        PyErr_Format(PyExc_ValueError, "In 'NRT_adapt_ndarray_to_python', 'descr' is NULL");
```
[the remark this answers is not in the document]
sklam: changed in https://github.com/numba/numba/pull/6141/commits/efff4b13d0ac4eaef9b6d004fb79393fa41dc8b7#diff-1e3a0be1e38fdb57e5bbff6f233dd1daR299-R300

numba/_helperlib.c
```diff
@@ -275,6 +275,12 @@ numba_recreate_record(void *pdata, int size, PyObject *dtype) {
     PyObject *record = NULL;
     PyArray_Descr *descr = NULL;
 
+    if (dtype == NULL) {
+        PyErr_Format(PyExc_ValueError,
+            "In 'numba_recreate_record', 'descr' is NULL");
+        return NULL;
+    }
```
stuartarchibald, line 282: Should this be a RuntimeError (as settled previously elsewhere) also, it refers to `descr` but is checking `dtype`.

numba/core/pythonapi.py
```diff
@@ -131,10 +131,27 @@ def read_const(self, index):
...
+        builder = self.pyapi.builder
+        consts = self.env_body.consts
+        ret = cgutils.alloca_once(builder, self.pyapi.pyobj, zfill=True)
+        with builder.if_else(cgutils.is_not_null(builder, consts)) as \
+                (br_not_null, br_null):
+            with br_not_null:
+                getitem = self.pyapi.list_getitem(consts, index)
+                builder.store(getitem, ret)
+            with br_null:
+                # This can happen when the Environment is accidentally released
```
stuartarchibald, line 149:
```suggestion
                # This can happen when the Environment is accidentally released and has subsequently been garbage collected.
```

sklam: done in https://github.com/numba/numba/pull/6141/commits/62d2a8b08942146d6dfe84efec2e36f8712527fa

numba/_helperlib.c
```diff
@@ -275,6 +275,12 @@ numba_recreate_record(void *pdata, int size, PyObject *dtype) {
     PyObject *record = NULL;
     PyArray_Descr *descr = NULL;
 
+    if (dtype == NULL) {
+        PyErr_Format(PyExc_Runtime,
```
stuartarchibald, line 279:
```suggestion
        PyErr_Format(PyExc_RuntimeError,
```

sklam:
failing due to: https://github.com/numba/numba/pull/6168

stuartarchibald (approved):
Thanks for the fixes!

numba/core/pythonapi.py
stuartarchibald: > I will make sure numba_recreate_record is checking the dtype argument

Did ^ get checked?

sklam: yes

stuartarchibald: thanks !

sklam:
CI passed, marked for merge

PR #6145, opened 2020-08-18
Patch to fingerprint namedtuples.

stuartarchibald:
As title. Should recover dispatch speed to nearer that of tuple.

### DIFF
--- a/numba/_typeof.c
+++ b/numba/_typeof.c
@@ -290,18 +290,22 @@ compute_fingerprint(string_writer_t *w, PyObject *val)
                 return -1;
 
             name = PyObject_GetAttrString(clazz, "__name__");
+            Py_DECREF(clazz);
             if (name == NULL)
                 return -1;
 
             ascii_str = PyUnicode_AsEncodedString(name, "ascii", "ignore");
+            Py_DECREF(name);
             if (ascii_str == NULL)
                 return -1;
             ret = PyBytes_AsStringAndSize(ascii_str, &buf, &flen);
+
             if (ret == -1)
                 return -1;
             for(j = 0; j < flen; j++) {
                 TRY(string_writer_put_char, w, buf[j]);
             }
+            Py_DECREF(ascii_str);
 
             if (_fields == NULL)
                 return -1;
@@ -324,6 +328,7 @@ compute_fingerprint(string_writer_t *w, PyObject *val)
                 for(j = 0; j < flen; j++) {
                     TRY(string_writer_put_char, w, buf[j]);
                 }
+                Py_DECREF(ascii_str);
                 TRY(compute_fingerprint, w, PyTuple_GET_ITEM(val, i));
             }
             TRY(string_writer_put_char, w, OP_END_TUPLE);
--- a/numba/_typeof.c
+++ b/numba/_typeof.c
@@ -257,14 +257,79 @@ compute_fingerprint(string_writer_t *w, PyObject *val)
         return string_writer_put_char(w, OP_FLOAT);
     if (PyComplex_CheckExact(val))
         return string_writer_put_char(w, OP_COMPLEX);
-    if (PyTuple_CheckExact(val)) {
-        Py_ssize_t i, n;
-        n = PyTuple_GET_SIZE(val);
-        TRY(string_writer_put_char, w, OP_START_TUPLE);
-        for (i = 0; i < n; i++)
-            TRY(compute_fingerprint, w, PyTuple_GET_ITEM(val, i));
-        TRY(string_writer_put_char, w, OP_END_TUPLE);
-        return 0;
+    if (PyTuple_Check(val)) {
+        if(PyTuple_CheckExact(val)) {
+            Py_ssize_t i, n;
+            n = PyTuple_GET_SIZE(val);
+            TRY(string_writer_put_char, w, OP_START_TUPLE);
+            for (i = 0; i < n; i++)
+                TRY(compute_fingerprint, w, PyTuple_GET_ITEM(val, i));
+            TRY(string_writer_put_char, w, OP_END_TUPLE);
+            return 0;
+        }
+        /* as per typeof.py, check "_asdict" for namedtuple. */
+        else if(PyObject_HasAttrString(val, "_asdict"))
+        {
+            /*
+             * This encodes the class name and field names of a namedtuple into
+             * the fingerprint on the condition that the number of fields is
+             * small (<10) and that the class name and field names are encodable
+             * as ASCII.
+             */
+            PyObject * clazz = NULL;
+            PyObject * name = NULL;
+            PyObject * _fields =  PyObject_GetAttrString(val, "_fields");
+            PyObject * field = NULL;
+            PyObject * ascii_str = NULL;
+            Py_ssize_t i, n, j, flen;
+            char * buf = NULL;
+            int ret;
+
+            clazz = PyObject_GetAttrString(val, "__class__");
+            if (clazz == NULL)
+                return -1;
+
+            name = PyObject_GetAttrString(clazz, "__name__");
+            if (name == NULL)
+                return -1;
+
+            ascii_str = PyUnicode_AsEncodedString(name, "ascii", "ignore");
+            if (ascii_str == NULL)
+                return -1;
+            ret = PyBytes_AsStringAndSize(ascii_str, &buf, &flen);
+            if (ret == -1)
+                return -1;
+            for(j = 0; j < flen; j++) {
+                TRY(string_writer_put_char, w, buf[j]);
+            }
+
+            if (_fields == NULL)
+                return -1;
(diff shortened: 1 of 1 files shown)


c200chromebook:
Under the hood is this how numpy structured data is represented? Would those of us using numpy dtypes get benefit?

stuartarchibald:
> Under the hood is this how numpy structured data is represented? Would those of us using numpy dtypes get benefit?

This patch solely addresses instances of things inheriting from `tuple`, explicitly `namedtuple`, it won't impact anything else. There's fast path typecode lookups etc for common dtypes in the dispatcher already but not for structured dtypes.

numba/_typeof.c
```diff
@@ -257,14 +257,79 @@ compute_fingerprint(string_writer_t *w, PyObject *val)
...
         return string_writer_put_char(w, OP_COMPLEX);
-    if (PyTuple_CheckExact(val)) {
-        Py_ssize_t i, n;
-        n = PyTuple_GET_SIZE(val);
-        TRY(string_writer_put_char, w, OP_START_TUPLE);
-        for (i = 0; i < n; i++)
-            TRY(compute_fingerprint, w, PyTuple_GET_ITEM(val, i));
-        TRY(string_writer_put_char, w, OP_END_TUPLE);
-        return 0;
+    if (PyTuple_Check(val)) {
```
stuartarchibald, line 260: Whole diff needs reference count looking at.

stuartarchibald:
TODO: benchmark this.

EPronovost:
Nice work! I'd be curious how much of an improvement a previous PR of mine yields once this merges. https://github.com/numba/numba/pull/5840

One namedtuple-related question I've had for a while: Currently numba does not support subclasses of `pt.NamedTuple` which define methods.  For example,

```
class Point(typing.NamedTuple):
    x: int
    y: int
        
        
    def mag(self):
        return np.sqrt(self.x**2 + self.y**2)
```

Do you think it would be feasible to support these (maybe not in this PR)?

numba/_typeof.c
```diff
@@ -257,14 +257,79 @@ compute_fingerprint(string_writer_t *w, PyObject *val)
...
+                return -1;
+            for(j = 0; j < flen; j++) {
+                TRY(string_writer_put_char, w, buf[j]);
+            }
+
+            if (_fields == NULL)
+                return -1;
+
+            n = PyTuple_GET_SIZE(val);
+            if (n > 10)
```
sklam, line 310: May need to remove this limitation.

sklam:
To check if this is actually fixing the problem.

sklam:
Needs decref: https://github.com/sklam/numba/commit/e984eb9406e37d71b8243562d0df849ac92087ac

numba/_typeof.c
```diff
@@ -257,14 +257,79 @@ compute_fingerprint(string_writer_t *w, PyObject *val)
...
+                return -1;
```
sklam, line 311:
```suggestion
                goto _unrecognized;
```

sklam:
Here's a microbenchmark:

```python
from numba import njit

@njit
def foo(x):
    return



for i in range(1, 8):
    n = 2**i
    print(f'n = {n}')
    tupcls = namedtuple("My", [f'f{x}' for x in range(n)])
    tup = tupcls(*range(n))
    foo(tup) #warmup
    %timeit foo(tup)
```
So, testing essentially the dispatch speed.

Results:

before PR:
```
n = 2
15.9 µs ± 540 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
n = 4
21.1 µs ± 1.11 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
n = 8
30.6 µs ± 637 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
n = 16
48.7 µs ± 1.51 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
n = 32
87.7 µs ± 2.31 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
n = 64
164 µs ± 3.49 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
n = 128
312 µs ± 8.48 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

PR limit to namedtuple size 10
```
n = 2
865 ns ± 24 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
n = 4
1.08 µs ± 25.8 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
n = 8
1.34 µs ± 32.7 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
n = 16
51.9 µs ± 510 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
n = 32
94.6 µs ± 3.23 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
n = 64
178 µs ± 5.43 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
n = 128
373 µs ± 19.2 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

PR unlimited nametuple size
```
n = 2
841 ns ± 27.5 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
n = 4
986 ns ± 7.32 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
n = 8
1.35 µs ± 60 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
n = 16
2.06 µs ± 81.7 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
n = 32
3.05 µs ± 132 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
n = 64
5.41 µs ± 206 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
n = 128
9.83 µs ± 413 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

I think it is worth making it unlimited.

stuartarchibald:
Thanks for the review @sklam, using the benchmark above I see this on mainline @ 9381206
```
n = 2
40.8 µs ± 1.13 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
n = 4
49.3 µs ± 38.9 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
n = 8
66.9 µs ± 1.36 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
n = 16
99.3 µs ± 830 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
n = 32
165 µs ± 224 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
n = 64
296 µs ± 203 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each)
n = 128
567 µs ± 9.38 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
and this with this PR @ 32ae0f6
```
n = 2
1.32 µs ± 18.2 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
n = 4
1.56 µs ± 3.08 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
n = 8
2.02 µs ± 3.95 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
n = 16
3.08 µs ± 10.9 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
n = 32
4.9 µs ± 3.75 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
n = 64
8.8 µs ± 26.8 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
n = 128
16.2 µs ± 23.3 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```
so this PR makes dispatch for `namedtuple` over an order of magnitude faster.

stuartarchibald:
@EPronovost 

> One namedtuple-related question I've had for a while: Currently numba does not support subclasses of `pt.NamedTuple` which define methods. For example,
> 
> ```
> class Point(typing.NamedTuple):
>     x: int
>     y: int
>         
>         
>     def mag(self):
>         return np.sqrt(self.x**2 + self.y**2)
> ```
> 
> Do you think it would be feasible to support these (maybe not in this PR)?

I think it may be possible to support this at some point, but is out of scope for this PR, perhaps open a ticket with it as a feature request?

numba/_typeof.c
stuartarchibald: done

numba/_typeof.c
stuartarchibald: resolved via removal of condition

stuartarchibald:
> Needs decrefs. See patch [sklam@e984eb9](https://github.com/sklam/numba/commit/e984eb9406e37d71b8243562d0df849ac92087ac)

Thanks, merged in.

sklam (approved):

Issue #6133, opened 2020-08-14
Compile times are nontrivally higher in 0.50 vs 0.46


c200chromebook:
## Feature request

Hi there! We've noted nontrivial slowdown in `@cuda.jit` compilation. It's not the easiest to replicate, as it seems to only show up in large kernels- the below should do it, on my sys >2x slower. Profiling, it seems to be related to this commit: https://github.com/numba/numba/pull/5351/commits/76b606ddfde801b8ea833120fc6c3d6691bda0e6

Any thoughts/advice? I know the example below is somewhat contrived, but it demos the issue pretty well.

```
from numba import cuda
import time
import numpy as np
from numpy import float64, int32, uint64, bool_
import numba


PDTYPE = [('pvar67', float64, (30,)), ('pvar72', float64, (210,))]
P_NP_DTYPE = np.dtype(PDTYPE, align=True)
P_NB_DTYPE = numba.from_dtype(P_NP_DTYPE)


INT_64_TYPE = [("uint64_1", uint64, (121, 47,)),
               ("uint64_2", uint64, (16, 31,))]
SV_TYPE = [("svvar1", float64, (121, 31, )),  ("svvar2", float64, (30, 31, )), ("svvar6", float64, (210, 31, )), ("svvar9", float64, (210, 31,))]


STYPE = [("svar9", float64, (161, 41, )), ("svar10", SV_TYPE)]
S_NP_TYPE = np.dtype(STYPE, align=True)
S_NB_TYPE = numba.from_dtype(S_NP_TYPE)


COTPYE13 = [('co13var1', bool_),
            ('co13var2', bool_),
            ('co13var3', bool_),
            ('co13var4', int32)]


COTYPE = [('covar13', COTPYE13)]
CO_NP_TYPE = np.dtype(COTYPE, align=True)
CO_NB_TYPE = numba.from_dtype(CO_NP_TYPE)



def load_kernel_2():
    @cuda.jit(argtypes=[P_NB_DTYPE[::1], CO_NB_TYPE[::1], S_NB_TYPE[::1]])
    def kernel_2(p_group, c_group, s_group):
        lane = cuda.blockDim.x * cuda.blockIdx.x + cuda.threadIdx.x
        pn = 1
        s_float = s_group[32].svar9
        s_vec = s_group[25].svar10
        c_s = c_group[0].covar13

        s_vec.svvar1[:, lane] = 0.0
### Copy/Paste Below ~100x
        for fi in range(0, 10):
[...]

PR #6146, opened 2020-08-18
Fix compilation slowdown due to controlflow analysis

sklam:


sklam:
My current reproducer for issue #6133

```python
from pprint import pprint
import numpy as np
import numba

body = """
    for fi in range(0, 10):
        data[fi] += 1
        for index in range(0, 7):
            idx = index
            if idx % 2 == 0:
                data[idx] += fi
            else:
                data[idx] += 1 if data[idx] < 3 else 2
                data[idx] += 1
            data[idx] += 1
"""

wrapper = """
def kernel(data):
    {}
"""

source = wrapper.format(body * 100) # 75
glbs = {}
exec(source, glbs)

arg = np.zeros(100)
argtypes = (numba.typeof(arg),)

kernel = numba.njit(argtypes)(glbs['kernel'])

# 0.48.0 takes 24sec
# 0.51.0 takes 1min13sec

metadata = kernel.get_metadata(kernel.signatures[0])
timings = metadata['pipeline_times']['nopython']
sorttimings = sorted(timings.items(), key=lambda x: sum(x[1]))
for k, v in reversed(sorttimings):
    print(f"{k:40} {sum(v):.3}")
```
This prints the compiler passes in descending runtime.

With the patch:
```log
13_reconstruct_ssa                       25.3
20_nopython_backend                      13.2
18_nopython_rewrites                     2.36
```
Without the patch:
```log
13_reconstruct_ssa                       40.1
20_nopython_backend                      13.5
18_nopython_rewrites                     3.23
```

sklam:
https://github.com/numba/numba/pull/6146/commits/264665ea5d4389bf69cbd036ebfc158c8cecf6c7 replaces the backedge algorithm with a simple but faster DFS version. This further speed up the SSA reconstruction:

```log
20_nopython_backend                      18.0
13_reconstruct_ssa                       14.1
18_nopython_rewrites                     2.48
```

stuartarchibald:
Thanks for the patch. I tested this manually, locally, 10.4->4.73s change in SSA rebuild.

PR #6148, opened 2020-08-18
Initial support for Numpy subclasses.

DrTodd13:
1. Provide way to say that you don't want Numpy subclass treated exactly like ndarray.
2. Augment typing to keep the ndarray subclass type during compilation.
3. Provide way to define what type you should get if you mix Numpy subclasses.
4. Provide overloaded array allocator.


DrTodd13:
@sklam Please review and provide initial feedback.  Thanks.

numba/core/ir_utils.py
```diff
@@ -108,11 +110,13 @@ def mk_alloc(typemap, calltypes, lhs, size_var, dtype, scope, loc):
...
-            typing.Context(), [size_typ, types.functions.NumberClass(dtype)], {})
-    # signature(
-    #    types.npytypes.Array(dtype, ndims, 'C'), size_typ,
-    #    types.functions.NumberClass(dtype))
+        cac = typemap[attr_var.name].get_call_type(
+             typing.Context(), [size_typ, types.functions.NumberClass(dtype)], {})
+        # By default, all calls to "empty" are typed as returning a standard
+        # Numpy ndarray.  If we are allocating a ndarray subclass here then
+        # just change the return type to be that of the subclass.
+        cac._return_type = arr_typ
```
sklam, line 118: This monkeypatch can be avoided by using `Signature.replace`.

numba/core/types/npytypes.py
```diff
@@ -394,8 +395,9 @@ class Array(Buffer):
     Type class for Numpy arrays.
     """
 
-    def __init__(self, dtype, ndim, layout, readonly=False, name=None,
-                 aligned=True):
+    def __init__(self, dtype, ndim, layout, PyType=np.ndarray, readonly=False,
+                 name=None, aligned=True):
+        self.PyType = PyType
```
sklam, line 400: PEP8 nitpick: lowercase + underscore preferred over CamelCase.

numba/core/typing/npydecl.py
```diff
@@ -126,7 +126,21 @@ def generic(self, args, kws):
...
+                    if hasattr(a, "__array_ufunc__"):
+                        array_ufunc_type = a
+                        break
+                output_type = types.Array
+                if array_ufunc_type is not None:
+                    output_type = array_ufunc_type.__array_ufunc__(ufunc, "__call__", *args, **kws)
+                    # Eventually better error handling!  FIX ME!
+                    assert(output_type is not None)
+
+                ret_tys = [output_type(dtype=ret_ty, ndim=ndims, layout=layout)
```
sklam, line 143: This is a new behavior. I would need to see some examples.

numba/core/pythonapi.py
```diff
@@ -45,10 +45,13 @@ def lookup(self, typeclass, default=None):
 _boxers = _Registry()
 _unboxers = _Registry()
 _reflectors = _Registry()
+# Registry of special allocators for types.
+_allocators = _Registry()
 
 box = _boxers.register
 unbox = _unboxers.register
 reflect = _reflectors.register
+allocator = _allocators.register
```
sklam, line 54: Unused?

numba/_typeof.c
```diff
@@ -794,8 +795,18 @@ typeof_typecode(PyObject *dispatcher, PyObject *val)
...
+    /* Subtypes of Array handling */
     else if (PyType_IsSubtype(tyobj, &PyArray_Type)) {
-        return typecode_ndarray(dispatcher, (PyArrayObject*)val);
+        /* If the class has an attribute named __numba_no_subtype_ndarray then
+           don't treat it as a normal variant of a Numpy ndarray but as it's own
+           separate type. */
+        no_subtype_attr = PyObject_HasAttrString(val, "__numba_no_subtype_ndarray__");
+        if (!no_subtype_attr) {
+            return typecode_ndarray(dispatcher, (PyArrayObject*)val);
+        }
```
sklam, line 809: Hmm... do we really have no other way but to add the `__numba_no_subtype_ndarray__`? For now, it does seem like the only way.

Also, we will need test and example for this.

numba/np/arrayobj.py
```diff
@@ -3399,8 +3400,13 @@ def _empty_nd_impl(context, builder, arrtype, shapes):
...
     align = context.get_preferred_array_alignment(arrtype.dtype)
-    meminfo = context.nrt.meminfo_alloc_aligned(builder, size=allocsize,
-                                                align=align)
+    def alloc_unsupported(context, builder, size, align):
+        return context.nrt.meminfo_alloc_aligned(builder, size, align)
+
+    # See if the type has a special allocator, if not use the default
+    # alloc_unsuppported allocator above.
+    allocator_impl = _allocators.lookup(arrtype.__class__, alloc_unsupported)
+    meminfo = allocator_impl(context, builder, size=allocsize, align=align)
```
sklam, line 3410: I wonder if we can refactor the core compiler to better support custom allocators

numba/core/ir_utils.py
DrTodd13: Done.

numba/core/types/npytypes.py
DrTodd13: Done

numba/core/pythonapi.py
DrTodd13: Used in user code.  Will post an example.

numba/np/arrayobj.py
DrTodd13: I am the least sure on this part.  This spot is the right spot to intercept allocation for a type because the allocation type doesn't propagate further into the allocators so you can't really intercept it where you'd like to only use a different allocator.  Memory management and allocation are different and right now we're forced to overload mem-mgmt when we'd only like to do the allocator.  I'm definitely open to all your suggestions if you want to do a more significant rearchitecting.

DrTodd13:
@sklam Below is the user code (slightly simplified for illustration).

`import numpy as np
from inspect import getmembers, isfunction, isclass
from numbers import Number
import numba
from numba import types
from numba.extending import typeof_impl, register_model, type_callable, lower_builtin
from numba.np import numpy_support
from numba.core.pythonapi import box, allocator
from llvmlite import ir
from numba.core import types, cgutils
import builtins

class ndarray(np.ndarray):
    # Tell Numba to not seamlessly treat this type as a regular ndarray.
    __numba_no_subtype_ndarray__ = True

    # Interoperate with Numpy outside of Numba.
    def __array__(self):
        print("ToddArray:ndarray:__array__")
        return self

    # Interoperate with Numpy outside of Numba.
    def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
        #print("Toddarray:__array_ufunc__", ufunc, type(ufunc), method, *inputs, **kwargs)
        if method == '__call__':
            N = None
            scalars = []
            for inp in inputs:
#                print("inp:", inp, type(inp))
                if isinstance(inp, Number):
                    scalars.append(inp)
                elif isinstance(inp, (self.__class__, np.ndarray)):
                    if isinstance(inp, self.__class__):
                        scalars.append(np.ndarray(inp.shape, inp.dtype, inp))
                    else:
                        scalars.append(inp)
                    if N is not None:
                        if N != inp.shape:
                            raise TypeError("inconsistent sizes")
                    else:
                        N = inp.shape
                else:
                    return NotImplemented
            ret = ufunc(*scalars, **kwargs)
#            print("ret:", ret, type(ret))
            return self.__class__(ret.shape, ret.dtype, ret)
        else:
            return NotImplemented

class ToddArrayType(types.Array):
    def __init__(self, dtype, ndim, layout, readonly=False, name=None,
                 aligned=True, addrspace=None):
        #print("ToddArrayType:", dtype, ndim, layout)
        name = "ToddArray:ndarray(%s, %sd, %s)" % (dtype, ndim, layout)
        super(ToddArrayType, self).__init__(dtype, ndim, layout,
                                            PyType=ndarray,
                                            readonly=readonly,
                                            name=name,
                                            addrspace=addrspace)

    # Tell Numba typing how to combine ToddArrayType with other ndarray types.
    def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
        #print("ToddArrayType:__array_ufunc__", ufunc, type(ufunc), method, *inputs, **kwargs)
        if method == '__call__':
            for inp in inputs:
                if not isinstance(inp, (ToddArrayType, types.Array, types.Number)):
                    return None

            return ToddArrayType
        else:
            return None

@typeof_impl.register(ndarray)
def typeof_ta_ndarray(val, c):
    #print("typeof_ta_ndarray:", val, type(val))
    try:
        dtype = numpy_support.from_dtype(val.dtype)
    except NotImplementedError:
        raise ValueError("Unsupported array dtype: %s" % (val.dtype,))
    layout = numpy_support.map_layout(val)
    readonly = not val.flags.writeable
    return ToddArrayType(dtype, val.ndim, layout, readonly=readonly)

register_model(ToddArrayType)(numba.core.datamodel.models.ArrayModel)

@type_callable(ndarray)
def type_ndarray(context):
    def typer(shape, ndim, buf):
        #print("typer:", shape, type(shape), ndim, type(ndim), buf, type(buf))
        return ToddArrayType(buf.dtype, buf.ndim, buf.layout)
    return typer

@lower_builtin(ndarray, types.UniTuple, types.DType, types.Array)
def impl_ndarray(context, builder, sig, args):
    shape, ndim, buf = args
    return buf

@box(ToddArrayType)
def box_array(typ, val, c):
    #print("ToddArrayType:box_array", typ, val, c)
    nativearycls = c.context.make_array(typ)
    nativeary = nativearycls(c.context, c.builder, value=val)
    if c.context.enable_nrt:
        np_dtype = numpy_support.as_dtype(typ.dtype)
        dtypeptr = c.env_manager.read_const(c.env_manager.add_const(np_dtype))
        # Steals NRT ref
        newary = c.pyapi.nrt_adapt_ndarray_to_python(typ, val, dtypeptr)
        return newary
    else:
        parent = nativeary.parent
        c.pyapi.incref(parent)
        return parent

@allocator(ToddArrayType)
def allocator_ToddArray(context, builder, size, align):
    # Implementation is just Numba's allocator at the moment for illustration.
    # Haven't done the integration to test our real new allocator yet.
    context.nrt._require_nrt()

    mod = builder.module
    u32 = ir.IntType(32)
    fnty = ir.FunctionType(cgutils.voidptr_t, [cgutils.intp_t, u32])
    fn = mod.get_or_insert_function(fnty,
                                    name="NRT_MemInfo_alloc_safe_aligned")
    fn.return_value.add_attribute("noalias")
    if isinstance(align, builtins.int):
        align = context.get_constant(types.uint32, align)
    else:
        assert align.type == u32, "align must be a uint32"
    return builder.call(fn, [size, align])
`

numba/_typeof.c
DrTodd13: For a lot of these places, there were questions in my mind about the right and most consistent way to do it with respect to the rest of the codebase.   See my user code example.

numba/core/typing/npydecl.py
DrTodd13: See my big user example.  This is a mirror of the approach that Numpy takes for supporting the combination of ndarray subtypes.  In the Numpy case, it actually does the operation but in our case we only need it for typing.  Maybe the more Numba way of doing it would be to use a decorator?

DrTodd13:
I realize now that there's a big chunk missing here and that is deallocation.  If something requires a special allocator then it most likely needs a special deallocator.  I haven't looked yet but I suspect that the deallocation code is currently type agnostic and if we are to know when to use the special deallocator then we need access to the type or need to embed the deallocator in the object itself.  Thoughts @sklam ?

sklam:
@DrTodd13, if you need a different deallocator, you might need to copy `NRT_MemInfo_alloc_safe_aligned` and change the `nrt_internal_dtor_safe` in https://github.com/numba/numba/blob/780f0831cba3442af821ad48fbe8dd6cfb28fb36/numba/core/runtime/nrt.c#L288

DrTodd13:
> @DrTodd13, if you need a different deallocator, you might need to copy `NRT_MemInfo_alloc_safe_aligned` and change the `nrt_internal_dtor_safe` in
> 
> https://github.com/numba/numba/blob/780f0831cba3442af821ad48fbe8dd6cfb28fb36/numba/core/runtime/nrt.c#L288
> 
> With that you can assign the deallocator in the definition for allocator.

I think we are just going to use our own meminfo.  It seems like we were already doing that in a related project but called it a different name.

gmarkall:
Note: I've added "waiting on author" in addition to "waiting on reviewer" whilst the tests are failing (I presume the fails are not spurious).

jpivarski:
My cross-references automatically showed up here, but I'm also eagerly awaiting the completion of this PR.

The [Vector](https://github.com/scikit-hep/vector) library features arrays of 2D, 3D, and Lorentz vectors for physics. I just finished an implementation of Lorentz vector _objects_ in Numba (scikit-hep/vector#39), and an array of vectors is a subclass of a NumPy structured array with the appropriate fields. All I'd need now is to have a lowered `__getitem__` on my `VectorNumpy*D` classes return lowered `VectorObject*D` objects to be able to iterate over arrays of vectors in JITted code. I was happily surprised that subclassed arrays pass through Numba, but then I didn't see any way to access the subclass from the `numba.types.Array` object. After that, I found this PR and saw that it's still a work in progress.

So I'll just say that I'm another user looking forward to this feature!

sklam:
@DrTodd13, @stuartarchibald 

I am noticing a few things.

#### 1. NRT can be specialized by the target-context subclass.

The base context defines the `.nrt` attribute at: https://github.com/numba/numba/blob/20d12618e590cf93e06907ae951f118f7e67824d/numba/core/base.py#L346-L348. Subclass can override to provide a different [NRTContext](https://github.com/numba/numba/blob/20d12618e590cf93e06907ae951f118f7e67824d/numba/core/runtime/context.py#L6). With that, we don't need to have a new allocator registry (yet?).

#### 2. Alternative to allocator registry 

Once @stuartarchibald's [WIP target-specific overriding](https://github.com/numba/numba/pull/6898/files) is merged, we can make the `allocate()` an intrinsic like in https://github.com/sklam/numba/commit/c33bb0c3a64e9f3ff7542abee064fb3a95f3360b and specialize the intrinsic with:

```python
@intrinsic(hardware="OtherProcessor")
def allocate(): ...
```

#### 3.  changes to `nrt.c`is necessary

I don't think we can avoid the changes to the `nrt.c` file given that the `Meminfo` object needs to remember what the are the memory-management functions so that a matching `free()` or `realloc()` is always used (from jit code and from C-API). 


Thoughts?

stuartarchibald:
@sklam thanks for looking at this.

> I am noticing a few things.
> #### 1. NRT can be specialized by the target-context subclass.
> 
> The base context defines the `.nrt` attribute at:
> 
> https://github.com/numba/numba/blob/20d12618e590cf93e06907ae951f118f7e67824d/numba/core/base.py#L346-L348
> . Subclass can override to provide a different [NRTContext](https://github.com/numba/numba/blob/20d12618e590cf93e06907ae951f118f7e67824d/numba/core/runtime/context.py#L6). With that, we don't need to have a new allocator registry (yet?).

Agree, seems like a new sub-target can just override for now via standard means.

> #### 2. Alternative to allocator registry
> 
> Once @stuartarchibald's [WIP target-specific overriding](https://github.com/numba/numba/pull/6898/files) is merged, we can make the `allocate()` an intrinsic like in [sklam@c33bb0c](https://github.com/sklam/numba/commit/c33bb0c3a64e9f3ff7542abee064fb3a95f3360b) and specialize the intrinsic with:
> 
> ```python
> @intrinsic(hardware="OtherProcessor")
> def allocate(): ...
> ```

I think this would provide a consistency of approach that may be desirable long term.

> #### 3. changes to `nrt.c`is necessary
> 
> I don't think we can avoid the changes to the `nrt.c` file given that the `Meminfo` object needs to remember what the are the memory-management functions so that a matching `free()` or `realloc()` is always used (from jit code and from C-API).

This seems like it's unavoidable. Suggest that changes are made and heavily tested given it's such a fundamental part of the code base?

sklam:
> Suggest that changes are made and heavily tested given it's such a fundamental part of the code base?

Yes, the C-API changes in NRT need tests.

DrTodd13:
> > Suggest that changes are made and heavily tested given it's such a fundamental part of the code base?
> 
> Yes, the C-API changes in NRT need tests.

Are there isolated tests for these already and we just need to add more?

sklam:
> > > Suggest that changes are made and heavily tested given it's such a fundamental part of the code base?
> > 
> > 
> > Yes, the C-API changes in NRT need tests.
> 
> Are there isolated tests for these already and we just need to add more?

yes, at https://github.com/numba/numba/blob/20d12618e590cf93e06907ae951f118f7e67824d/numba/tests/test_nrt.py#L76

DrTodd13:
@sklam @stuartarchibald 

> @sklam thanks for looking at this.
> 
> > I am noticing a few things.
> > #### 1. NRT can be specialized by the target-context subclass.
> > The base context defines the `.nrt` attribute at:
> > https://github.com/numba/numba/blob/20d12618e590cf93e06907ae951f118f7e67824d/numba/core/base.py#L346-L348
> > 
> > . Subclass can override to provide a different [NRTContext](https://github.com/numba/numba/blob/20d12618e590cf93e06907ae951f118f7e67824d/numba/core/runtime/context.py#L6). With that, we don't need to have a new allocator registry (yet?).
> 
> Agree, seems like a new sub-target can just override for now via standard means.

I'm a little confused here.  Maybe there is a reason to have multiple NRTs, one per context but would that somehow solve the current problem in a more elegant way?  The nature of our custom allocations is that the memory is accessible on CPU and GPU and so that is just the default context.  You might have allocations in one function with one target that require different allocators so I don't understand how a target NRT context would help.

> 
> > #### 2. Alternative to allocator registry
> > Once @stuartarchibald's [WIP target-specific overriding](https://github.com/numba/numba/pull/6898/files) is merged, we can make the `allocate()` an intrinsic like in [sklam@c33bb0c](https://github.com/sklam/numba/commit/c33bb0c3a64e9f3ff7542abee064fb3a95f3360b) and specialize the intrinsic with:
> > ```python
> > @intrinsic(hardware="OtherProcessor")
> > def allocate(): ...
> > ```
> 
> I think this would provide a consistency of approach that may be desirable long term.

The work you are doing to extend to other hardware is great but I don't think there is this 1:1 correspondence between HW and allocator.  Just take USM memory in its three flavors: "host", shared and device.  You might say that the device version fits with some HW specific allocator but "shared" doesn't fit in this model, unless you define the model as a system composed of two kinds of HW.  Moreover, for "host" mode, the same "system as two HWs" is used as in the shared mode but the behavior of the memory is different ("host" is paged on demand whereas shared is transferred in bulk).  Maybe it makes sense to be able to easily define an allocator for different HW if you don't have this 1:1 model but I don't think the system should enforce 1:1 and so allocator needs to be independent of HW.

sklam:
@DrTodd13 

With a single NRT, what if there are multiple libraries that are registering non-compatible allocators for the same type? That is why I think there needs to be some form of dependency on the target-context. Either directly on the context, or some target options. 

There will also be problem for cached machine code. The changes to the allocator should be recognized as an ABI change for the function about it is not being tracked.

DrTodd13:
> @DrTodd13
> 
> With a single NRT, what if there are multiple libraries that are registering non-compatible allocators for the same type? That is why I think there needs to be some form of dependency on the target-context. Either directly on the context, or some target options.
> 
> There will also be problem for cached machine code. The changes to the allocator should be recognized as an ABI change for the function about it is not being tracked.

@sklam
I don't think this is an issue that we can do anything about so we shouldn't worry about it.  It seems to me it is the same issue you already have.  For example, what if two libraries both register overloads, boxing/unboxing, etc. for the same third party library's data structures and functions?  I think we have as a requirement that two variables right next to each other in the same function running in the same context can select different allocators.  Yes, we could invent an artificial context just to encapsulate allocations but then I'd have to artificially break apart my functions to get those different allocators.  Right now it is tied to a type so all things of that typed get a certain allocator but if I could I would have made the allocator even independent of type but I don't know how to do that.

Help me understand the caching issue better.  If on the original run, a type had an allocator overload but on the second run it didn't then maybe that's an issue but I don't think that is what you are referring to.  Is it that a pointer to the allocator is embedded in the machine code and that address may change between runs?

sklam:
@DrTodd13 I adapted your example to use the `overload_classmethod`  (https://github.com/sklam/numba/commit/7078301888442e8eaa1f46d74c78c2ae64bc613d) at https://gist.github.com/sklam/04e722fdf5cb215ff909ae4908712da7. The revisions on the gist shows the changes from the [original code](https://github.com/numba/numba/pull/6148#issuecomment-676541700).

sklam (commented):
I have additional comments for the changes here. Mainly it is to better exercise and document the code changes. If we are to include these changes as experimental features, we should make it more clear that they are unstable.

I can propose some changes and also modify this PR to use the `overload_classmethod` in #7061

numba/_typeof.c
```diff
@@ -964,9 +965,19 @@ typeof_typecode(PyObject *dispatcher, PyObject *val)
...
-    else if (PyType_IsSubtype(tyobj, &PyArray_Type)) {
+    else if (tyobj == &PyArray_Type) {
         return typecode_ndarray(dispatcher, (PyArrayObject*)val);
     }
+    /* Subtypes of Array handling */
+    else if (PyType_IsSubtype(tyobj, &PyArray_Type)) {
+        /* If the class has an attribute named __numba_no_subtype_ndarray then
+           don't treat it as a normal variant of a Numpy ndarray but as it's own
+           separate type. */
+        no_subtype_attr = PyObject_HasAttrString(val, "__numba_no_subtype_ndarray__");
```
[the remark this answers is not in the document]
DrTodd13: I have renamed it to __numba_array_subtype_dispatch__.

numba/core/runtime/nrt.c
```diff
@@ -271,26 +286,48 @@ void *nrt_allocate_meminfo_and_data_align(size_t size, unsigned align,
...
+    /* Only fill up a couple cachelines with debug markers, to minimize
+       overhead. */
+    memset(data, 0xCB, MIN(size, 256));
+    NRT_Debug(nrt_debug_print("NRT_MemInfo_alloc_safe_aligned %p %zu\n",
+                              data, size));
+    NRT_MemInfo_init(mi, data, size, nrt_internal_dtor_safe, (void*)size, allocator);
+    return mi;
+}
+
+void NRT_dealloc(NRT_MemInfo *mi) {
```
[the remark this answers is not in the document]
DrTodd13: NRT_Free takes a void * as the ptr parameter.  There are places NRT_Free is called where a void * is passed to it as well.  So, I don't think we can assume that "ptr" is really a NRT_MemInfo * which is what you'd have to do to switch on mi->external_allocator.  We could move NRT_dealloc into NRT_MemInfo_destroy because that seems to be the only place it is called.

numba/core/types/npytypes.py
```diff
@@ -415,8 +416,9 @@ class Array(Buffer):
     Type class for Numpy arrays.
     """
 
-    def __init__(self, dtype, ndim, layout, readonly=False, name=None,
-                 aligned=True):
+    def __init__(self, dtype, ndim, layout, py_type=np.ndarray, readonly=False,
+                 name=None, aligned=True):
+        self.py_type = py_type
```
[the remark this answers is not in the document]
DrTodd13: @sklam You said you were going to do this one if I remember correctly.  Is that right?

numba/core/typing/npydecl.py
```diff
@@ -127,7 +127,21 @@ def generic(self, args, kws):
...
+                # __array_ufunc__ method then invoke the first such one to
+                # determine the output type of the ufunc.
+                array_ufunc_type = None
+                for a in args:
+                    if hasattr(a, "__array_ufunc__"):
+                        array_ufunc_type = a
+                        break
+                output_type = types.Array
+                if array_ufunc_type is not None:
+                    output_type = array_ufunc_type.__array_ufunc__(ufunc, "__call__", *args, **kws)
```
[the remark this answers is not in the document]
DrTodd13: @sklam Same as above....you said you were going to do this right?

sklam:
I'm adding tests to this PR in #7121.

esc:
#7121 has supplanted this one.

PR #6152, opened 2020-08-19
CUDA: Provide wrappers for all libdevice functions, and fix typing of math function (#4618)

### LINKED ISSUE #4618: CUDA math.* typing incorrect
## Reporting a bug

This fails to compile:
```python
from numba import cuda
import math
import numpy as np

@cuda.jit
def foo(x, y):
    y[0] = math.cos(x[0])


a = np.zeros(10, dtype=np.int64)
foo(a, a)
```
as there's no implementation to lower a `math.cos(int64)` but such an implementation was declared here:
https://github.com/numba/numba/blob/3a6f134eae121e50eb008f202d72e9b27bc84626/numba/cuda/cudamath.py#L41-L42

I suspect that all `math.*` functions suffer a similar fate as libdevice bindings won't exist for the case where a int->float cast is required.

gmarkall:
Whilst looking into fixing #4618 I also thought it would be good to add wrappers for all libdevice functions, rather than just a subset needed for the currently-implemented math functions. This PR adds wrappers for all libdevice functions. The implementation is based around a data structure containing all relevant details of each libdevice function in `numba/cuda/libdevicefuncs.py`, from which the stubs, typing, lowering, and documentation are generated.

The changes in this PR are broadly:

- Addition of the definitions of libdevice functions in `libdevicefuncs.py`
- Generation of libdevice stubs using the definitions and function in `libdevicefuncs.py`, which are in `libdevice.py`
- Add declarations and implementations of stub functions based on the definitions in `libdevicefuncs.py`.
  - The typing and lowering of functions where only a single output is produced through the return value are straightforward.
  - The typing and lowering of functions with multiple return values are a little more complex, because the Python function signature needs to be generated such that the return value is a tuple of all output values, and the arguments are all input values. Output values can be distinguished from input values because outputs are passed as pointers (or are the returned value). To generate the implementation, a temporary value needs to be allocated for each output passed as a pointer, and after the call to the libdevice function the return values need to be packed into a struct or array (depending on whether the return type is a `Tuple` or `UniTuple`).
- Addition of the typing and lowering registries to the CUDA typing and target contexts.
- Move all the `math` module implementations into `numba.cuda.mathimpl` (instead of being in `libdevice`). These new implementations make calls to the libdevice stub functions rather than directly generating calls to libdevice functions.
- Move the implementation of lowering `math.isinf` and `math.isnan` for integer arguments from `cudaimpl` into `mathimpl`, where it is more appropriate.
- Update tests in `test_math`. This consists of:
  - Adds tests of passing integer arguments to all math functions that are typed as accepting integers, and
  - Using `assert_allclose` to provide better feedback when a test fails.
- Add tests for libdevice functions. The tests consist of:
  - One compile test per wrapped function generated programatically, and
  - One test of each different return type that compares the returned values against reference values computed with NumPy.
  It isn't practical at present to numerically test every wrapped function - the aim is that these  combinations of tests demonstrate sufficient confidence in the typing and lowering that at least the stub functions, typing, and lowering are correctly passing values to and returning values from the libdevice functions. Note also that the `test_math` tests are numerically testing the libdevice wrappers for all the functions that `math` implementations wrap, since the lowering of the `math` functions now uses the libdevice wrappers.

Inspection shows that these changes make no difference to the generated PTX, with the exception of additional variants being generated for the new tests that are added with this PR.

Fixes #4618.


gmarkall:
The last commit I pushed here was an accident - it should have been on another PR. I will revert it when back to a desktop machine (or leave it in if you like, it does correct a small documentation error).

gmarkall:
Moving this into 0.52RC as it fixes #4618, which is in 0.52RC. Will attend to the comments and conflicts tomorrow.

numba/cuda/tests/cudapy/test_libdevice.py
```diff
@@ -0,0 +1,186 @@
...
+        arr = np.linspace(start=1.0, stop=10.0, num=100, dtype=np.float64)
+        fracres = np.zeros_like(arr)
+        expres = np.zeros(shape=arr.shape, dtype=np.int32)
+
+        cufunc = cuda.jit(use_frexp)
+        cufunc[4, 32](fracres, expres, arr)
+
+        frac_expect, exp_expect = np.frexp(arr)
+
+        np.testing.assert_allclose(frac_expect, fracres)
```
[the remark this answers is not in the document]
gmarkall: Good point - thankyou! I've updated the test to use `assert_array_equal`.

gmarkall:
@stuartarchibald Could this please have a buildfarm run prior to review, so if there are any issues I can fix them up before review?

esc:
BFID: `numba_smoketest_cuda_116`

esc:
The Linux tests that use CUDA 8 are failing on this one. Is it to be expected?

esc:
The Windows builds have all glitched and wouldn't start, so I am giving this another spin at: `numba_smoketest_cuda_117`.

gmarkall:
> The Linux tests that use CUDA 8 are failing on this one. Is it to be expected?

The merge of #6127 dropped support for CUDA 8 - I would guess it should be removed from the farm now.

esc:
@gmarkall CUDA 8.0 has been dropped

.flake8
```diff
@@ -299,3 +299,8 @@ per-file-ignores =
     # errors in device_init because its purpose is to bring together a lot of
     # the public API to be star-imported in numba.cuda.__init__
     numba/cuda/device_init.py:F401,F403,F405
+    # Libdevice is an autogenerated file containing stubs for all the device
```
stuartarchibald, line 302:
```suggestion
    # libdevice.py is an autogenerated file containing stubs for all the device
```

docs/source/cuda-reference/libdevice.rst
```diff
@@ -0,0 +1,16 @@
+Libdevice functions
+===================
+
+All wrapped libdevice functions are listed in this section. All functions in
+libdevice are wrapped, with the exception of ``__nv_nan`` and ``__nv_nanf``.
+These functions return a representation of a quiet NaN, but the argument they
+take (a pointer to an object specifying the representation) is undocumented, and
+follows an unusual form compared to the rest of libdevice - it is not an output
+like every other pointer argument. If a NaN is required, one can be obtained in
+CUDA Python by other means, e.g. ``math.nan``.
```
stuartarchibald, line 10: Want to also mention inf ?

numba/cuda/mathimpl.py
```diff
@@ -0,0 +1,246 @@
...
+from numba.cuda import libdevice
+
+registry = Registry()
+lower = registry.lower
+
+
+booleans = []
+booleans += [('isnand', 'isnanf', math.isnan)]
+booleans += [('isinfd', 'isinff', math.isinf)]
+booleans += [('isfinited', 'finitef', math.isfinite)]
```
stuartarchibald, line 16: I've checked this is correct, `isfinited` for double and `finitef` for float, imagine it's a historical artifact.

numba/cuda/mathimpl.py
```diff
@@ -0,0 +1,246 @@
...
+
+    lower(key, ty)(lower_unary_impl)
+
+
+def impl_unary_int(key, ty, libfunc):
+    def lower_unary_int_impl(context, builder, sig, args):
+        if sig.args[0] == int64:
+            convert = builder.sitofp
+        else:
+            convert = builder.uitofp
```
stuartarchibald, line 87: What if it's accidentally an `int32` if someone tries to reuse this for something? Would it help typing/lowering drift to only accept 64 bit ints?

numba/cuda/mathimpl.py
```diff
@@ -0,0 +1,246 @@
...
+                                            typing.signature(ty, ty, ty))
+        return libfunc_impl(builder, args)
+
+    lower(key, ty, ty)(lower_binary_impl)
+
+
+def impl_binary_int(key, ty, libfunc):
+    def lower_binary_int_impl(context, builder, sig, args):
+        if sig.args[0] == int64:
+            convert = builder.sitofp
```
stuartarchibald, line 108: Same question as above RE int32.

numba/cuda/mathimpl.py
```diff
@@ -0,0 +1,246 @@
...
+    impl_binary_int(key, uint64, impl64)
+
+
+def impl_pow(ty, libfunc):
+    def lower_pow_impl(context, builder, sig, args):
+        pow_sig = typing.signature(ty, ty, ty)
+        libfunc_impl = context.get_function(libfunc, pow_sig)
+        return libfunc_impl(builder, args)
+
+    lower(math.pow, ty, types.int32)(lower_pow_impl)
```
stuartarchibald, line 150:
```suggestion
    lower(math.pow, ty, ty)(lower_pow_impl)
```
should it be that ^?

numba/cuda/mathimpl.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+def impl_pow_int(ty, libfunc):
+    def lower_pow_impl_int(context, builder, sig, args):
+        powi_sig = typing.signature(ty, ty, types.int32)
+        libfunc_impl = context.get_function(libfunc, powi_sig)
+        return libfunc_impl(builder, args)
+
+    lower(math.pow, ty, types.int32)(lower_pow_impl_int)
```
stuartarchibald, line 159: NOTE: device functions accept i32 only.

numba/cuda/mathimpl.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+impl_pow(types.float32, libdevice.powf)
+impl_pow(types.float64, libdevice.pow)
+impl_pow_int(types.float32, libdevice.powif)
+impl_pow_int(types.float64, libdevice.powi)
+
+
+def impl_modf(ty, libfunc):
+    retty = types.Tuple((ty, ty))
```
stuartarchibald, line 169: Unituple? Homogeneous in dtype?

numba/cuda/tests/cudapy/test_libdevice.py
```diff
@@ -0,0 +1,186 @@
...
+
+def make_test_call(libname):
+    """
+    Generates a test function for each libdevice function.
+    """
+
+    def _test_call_functions(self):
+        # Strip off '__nv_' from libdevice name to get Python name
+        apiname = libname[5:]
+        apifunc = getattr(libdevice, libname[5:])
```
stuartarchibald, line 105:
```suggestion
        apifunc = getattr(libdevice, apiname)
```

numba/cuda/tests/cudapy/test_libdevice.py
```diff
@@ -0,0 +1,186 @@
...
+    Class for holding all tests of compiling calls to libdevice functions. We
+    generate the actual tests in this class (as opposed to using subTest and
+    one test within this class) because there are a lot of tests, and it makes
+    the test suite appear frozen to test them all as subTests in one test.
+    """
+
+
+for libname in functions:
+    setattr(TestLibdeviceCompilation, 'test_%s' % libname,
+            make_test_call(libname))
```
stuartarchibald, line 182: As this is auto-generated, should probably add it to this, largely so it's not forgotten. 
https://github.com/numba/numba/blob/c5788551c582711255dc6ab737e4c76a6e794155/numba/testing/main.py#L337

I don't think it'll have any impact but want to avoid having to debug the class of problem that occurs if injects tests aren't acknowledged.

numba/cuda/tests/cudapy/test_libdevice.py
```diff
@@ -0,0 +1,186 @@
...
+            # Scalar return is a more straightforward case
+            pyargs = "r0, " + funcargs
+            retvars = "r0[0]"
+
+        # Create the string containing the function to compile
+        d = { 'func': apiname,
+              'pyargs': pyargs,
+              'funcargs': funcargs,
+              'retvars': retvars }
+        code = function_template % d
```
stuartarchibald, line 140: NOTE: I spot checked a few of these. Code gen looks fine.

stuartarchibald (commented):
Thanks for the effort on this, it's great to see the hardware math functions neatly wrapped and wired up. I've made a few comments in line to resolve, but in general looks good.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_120` for 2e3ed5d. It'll potentially need another run post review fixes but thought I'd get it started.

docs/source/cuda-reference/libdevice.rst
gmarkall: I didn't omit a libdevice function for generating the representation of an inf, so it seems odd to do so in this context.

numba/cuda/mathimpl.py
gmarkall: Yes, it is odd!

numba/cuda/mathimpl.py
gmarkall: The implementation was redundant, covered by `lower_binary_impl` from the list of `binarys` above: https://github.com/numba/numba/pull/6152/files#diff-3211256888268b006cb02721c73fd0807f610c0ef182d4a9aaf32cfcebd84f3fR48

numba/cuda/mathimpl.py
gmarkall: I've now made it a `UniTuple`

numba/cuda/tests/cudapy/test_libdevice.py
gmarkall: Change applied.

numba/cuda/tests/cudapy/test_libdevice.py
gmarkall: I've now added it to `_GENERATED`.

numba/cuda/tests/cudapy/test_libdevice.py
gmarkall: Many thanks!

numba/cuda/mathimpl.py
gmarkall: I've now changed it to only accept 64-bit integers.

numba/cuda/mathimpl.py
gmarkall: Changed as above.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_121` for a752f01 .

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_121` for [a752f01](https://github.com/numba/numba/commit/a752f01c9f7766085f47f59ebfb4715cd3a9739e) .

This passed.

sklam (approved):
Thanks for the patch!

PR #6156, opened 2020-08-20
Remove callers of `reraise` in favor of using `with_traceback` directly

eric-wieser:
More Python 2 &rarr; Python 3 cleanup

### DIFF
--- a/numba/cuda/simulator/kernel.py
+++ b/numba/cuda/simulator/kernel.py
@@ -5,7 +5,6 @@ import threading
 
 import numpy as np
 
-from numba.core.utils import reraise
 from .cudadrv.devicearray import to_device, auto_device, \
     FakeCUDAArray, FakeWithinKernelCUDAArray
 from .kernelapi import Dim3, FakeCUDAModule, swapped_cuda_module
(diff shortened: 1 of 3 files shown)


stuartarchibald:
Thanks for the PR. Please could you resolve the conflicts and then someone can review, many thanks!

eric-wieser:
Rebased

eric-wieser:
Once this goes in, I'll put up a PR to gut `numba.utils`

numba/cuda/simulator/kernel.py
```diff
@@ -176,7 +175,9 @@ def run(self):
             else:
                 msg = '%s %s: %s' % (tid, ctaid, e)
             tb = sys.exc_info()[2]
-            self.exception = (type(e), type(e)(msg), tb)
+            # Using `with_traceback` here would cause it to be mutated by
+            # future raise statements, which may or may not matter.
+            self.exception = (type(e)(msg), tb)
```
stuartarchibald, line 180: @gmarkall do you have a view here? I think what's present should be fine.

eric-wieser: To elaborate on that commentL using
```
self.exception = e.with_traceback(tb)
```
followed by
```
raise self.exception
```
results in self.exception.__traceback__ being modified in-place, making a second `raise self.exception` give a longer traceback.
I don't know if that can ever happen, but since the code before avoided this issue, I figured I should leave the semantics identical.

stuartarchibald: Thanks for expanding. I can't see a need to change this behaviour.

stuartarchibald (approved):

PR #6157, opened 2020-08-20
Add pipeline_class argument to @cfunc as supported by @jit.

arthurp:
This allows custom compiler pipelines to be used for cfuncs and makes the APIs consistent.


stuartarchibald (approved):
Thanks for the patch, looks good!

stuartarchibald:
@arthurp Congratulations on your first contribution to Numba!

PR #6162, opened 2020-08-24
Move charseq support out of init

stuartarchibald:
As title.

### DIFF
--- a/numba/__init__.py
+++ b/numba/__init__.py
@@ -75,9 +75,6 @@ from numba.experimental.jitclass.decorators import _warning_jitclass as jitclass
 import numba.core.withcontexts
 from numba.core.withcontexts import objmode_context as objmode
 
-# Bytes/unicode array support
-import numba.cpython.charseq
-
 # Keep this for backward compatibility.
 test = runtests.main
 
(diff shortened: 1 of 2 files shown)


esc:
I just reviewed this it seems fine. I also looked at the failing checks and they only appear to have failed in the github interface, once I follow the link to Azure, the checks appear to have run fine. Also, not sure how to test this best, what to look for. Lastly, one question: why was the `listimpl` import needed?

stuartarchibald:
> I just reviewed this it seems fine. I also looked at the failing checks and they only appear to have failed in the github interface, once I follow the link to Azure, the checks appear to have run fine.

Thanks.

> Also, not sure how to test this best, what to look for.

It's moving the init to where it probably ought to be, importing the package doesn't need to trigger importing charseq details, the compiler can do that if it's used. As to testing, didn't break existing tests is probably the test.

> Lastly, one question: why was the `listimpl` import needed?

Because I remembered wrong, there's no `listimpl` as there's already list lowering registered, it just looked like it was missing.

esc:
@stuartarchibald thanks for following up, marking as ready to merge.

PR #6166, opened 2020-08-25
Remove Python 2 compatibility from numba.core.utils

eric-wieser:

### DIFF
--- a/numba/core/utils.py
+++ b/numba/core/utils.py
@@ -24,7 +24,6 @@ from numba.core.config import (PYVERSION, MACHINE_BITS, # noqa: F401
                                DEVELOPER_MODE) # noqa: F401
 from numba.core import types
 
-INT_TYPES = (int,)
 longint = int
 get_ident = threading.get_ident
 intern = sys.intern
@@ -367,7 +366,7 @@ def bit_length(intval):
     """
     Return the number of bits necessary to represent integer `intval`.
     """
-    assert isinstance(intval, INT_TYPES)
+    assert isinstance(intval, int)
     if intval >= 0:
         return len(bin(intval)) - 2
     else:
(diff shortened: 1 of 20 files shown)


stuartarchibald:
note to self, check `LICENSE.*` for dropping six ref.

eric-wieser:
Remove `six` and `funcsigs` from the license, good thinking

stuartarchibald:
hmmm, segfault on the TBB build, this may be https://github.com/numba/numba/issues/5973, will take a look, it's unlikely to be changes in this PR but you never know!

stuartarchibald:
https://github.com/numba/numba/pull/6171 is also failing and that's a changelog update, so the issue is not related to the code in here.

numba/core/utils.py
```diff
@@ -146,7 +75,8 @@ def safe_relpath(path, start=os.curdir):
     'is': operator.is_,
     'is not': operator.is_not,
     # This one has its args reversed!
-    'in': operator.contains
+    'in': operator.contains,
+    '@': operator.matmul,
```
EPronovost, line 79: Are you planning to add support for the `@` operator on numpy arrays (probably not in this PR)?

eric-wieser: No, I just moved this from a line further down this file.

stuartarchibald: `@` operator already works on NumPy arrays IIRC.
```pycon
In [1]: from numba import njit

In [2]: import numpy as np

In [3]: a = np.ones((3, 3))

In [4]: @njit
   ...: def foo(a):
   ...:     return a@a
   ...: 

In [5]: foo(a)
Out[5]: 
array([[3., 3., 3.],
       [3., 3., 3.],
       [3., 3., 3.]])
```

gmarkall (approved):
Looks like this is good but just needs CI re-running. @stuartarchibald Can you re-run it please?

sklam (approved):
Verified the changes regarding the license file and that corresponding parts have been removed

Issue #6167, opened 2020-08-25
#6147 broke CUDASIM


stuartarchibald:
Simulator failing on current master (ea5655232) with:
```
$ NUMBA_ENABLE_CUDASIM=1 ./runtests.py numba.tests.test_runtests.TestCase -v -f
test_all (numba.tests.test_runtests.TestCase) ... ok
test_check_slice (numba.tests.test_runtests.TestCase) ... ok
test_check_slicing_equivalent (numba.tests.test_runtests.TestCase) ... ok
test_cuda (numba.tests.test_runtests.TestCase) ... ok
test_cuda_submodules (numba.tests.test_runtests.TestCase) ... FAIL

======================================================================
FAIL: test_cuda_submodules (numba.tests.test_runtests.TestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "<path>/numba/numba/tests/test_runtests.py", line 62, in test_cuda_submodules
    self.check_listing_prefix('numba.cuda.tests.cudadrv')
  File "<path>/numba/numba/tests/test_runtests.py", line 23, in check_listing_prefix
    self.assertTrue(ln.startswith(prefix), msg=errmsg)
AssertionError: False is not true : 'unittest.loader._FailedTest.test_nvvm_driver' not startswith 'numba.cuda.tests.cudadrv'

----------------------------------------------------------------------
Ran 5 tests in 22.324s

FAILED (failures=1)
```
likely due to https://github.com/numba/numba/pull/6147. Needs fixing for 0.51.1.

sklam:
Closed by #6168

PR #6147, opened 2020-08-18
[WIPDon't make a runtime call on import

gmarkall:
Calling `runtime.get_version()` on import initializes the CUDA driver, even though it creates no context. This is an issue for libraries that import Numba then set `CUDA_VISIBLE_DEVICES`, because the call at import time already initialized the driver and fixed the set of visible devices. This is an issue for, but not limited to, the Dask Local CUDA Cluster.

This PR fixes this by deferring the call to `runtime.get_version()` until the list of supported CCs is actually needed.

cc @kkraus14 @pentschev @quasiben

PR #6168, opened 2020-08-25
Fix Issue #6167: Failure in test_cuda_submodules

gmarkall:
PR #6147 broke the test_cuda_submodules test because the test_nvvm_driver test could no longer be imported - the change from `SUPPORTED_CC` to `get_supported_ccs` in nvvm.py was not reflected in the CUDA simulator.

This PR fixes the issue by changing `SUPPORTED_CC` to `get_supported_ccs` in the simulator.


stuartarchibald (approved):
Thanks for fixing this so quickly!

Issue #6163, opened 2020-08-24
Better error message in `with ... as ...`


sklam:
Probably we can catch the error message and give you a better message along the lines of: "Numba is currently unable to compile your code."

_Originally posted by @esc in https://github.com/numba/numba/issues/6142#issuecomment-679160903_

We can catch the NotDefinedError can report a better message that says the compiler failed to analyze the bytecode.

abhinavjonnada82:
Can I get this?

esc:
@abhinavjonnada82 feel free to start working on it and submit a PR!

stuartarchibald:
This was implemented in https://github.com/numba/numba/pull/6185. Closing.

@abhinavjonnada82 thanks for asking, as evident in #6185, this ticket should be closed, apologies for any confusion. If you'd like to contribute to Numba there's a [good first issue](https://github.com/numba/numba/labels/good%20first%20issue) label which contains other issues which might be good to try? Thanks.

Issue #6142, opened 2020-08-18
Variable '$88setup_with.1' is not defined.


OfekShochat:
I'm making a chess engine in python (I would do it in c but this is for fun) and it is pretty good but **really** slow. so I wanted to make it faster with numba njit. but it gets an error. full error code: 
```
Traceback (most recent call last):
  File "C:\Users\o\AppData\Local\Programs\Python\Python38-32\lib\site-packages\numba\core\ir.py", line 267, in get
    return self._con[name]
KeyError: '$88setup_with.1'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\o\AppData\Local\Programs\Python\Python38-32\lib\site-packages\numba\core\ir.py", line 1119, in get_exact
    return self.localvars.get(name)
  File "C:\Users\o\AppData\Local\Programs\Python\Python38-32\lib\site-packages\numba\core\ir.py", line 269, in get
    raise NotDefinedError(name)
numba.core.errors.NotDefinedError: Variable '$88setup_with.1' is not defined.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\o\AppData\Local\Programs\Python\Python38-32\lib\site-packages\numba\core\ir.py", line 267, in get
    return self._con[name]
KeyError: '$88setup_with.1'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "E:\lichess-bot\engines\engine2\main.pyx", line 320, in <module>
    main2()
  File "E:\lichess-bot\engines\engine2\main.pyx", line 317, in main2
    return negaa(board, depth, -inf, 10000, color)
  File "C:\Users\o\AppData\Local\Programs\Python\Python38-32\lib\site-packages\numba\core\dispatcher.py", line 423, in _compile_for_args
    error_rewrite(e, 'interpreter')
  File "C:\Users\o\AppData\Local\Programs\Python\Python38-32\lib\site-packages\numba\core\dispatcher.py", line 358, in error_rewrite
    reraise(type(e), e, None)
  File "C:\Users\o\AppData\Local\Programs\Python\Python38-32\lib\site-packages\numba\core\utils.py", line 80, in reraise
    raise value.with_traceback(tb)
numba.core.errors.NotDefinedError: Failed in nopython mode pipeline (step: analyzing bytecode)
Variable '$88setup_with.1' is not defined.
``` 
and the jited function:
```
    st = time()
    try:
        move = opening_move(board, book_path)
[...]

PR #6185, opened 2020-08-28
Better error message on NotDefinedError

Luiz6ustav0:
I started working on this because of #6163 , which mentions #6142 .

The only change I made is in the message that is passed to the constructor in the erros.py file.
```python
msg = "The compiler failed to analyze the bytecode"
```
If I understood it correctly that's where it should be but this is my first contribution here, if it's incorrect  maybe someone can give a hint on where I should do this work because I didn't find the exact place where I'd catch this error specifically with the "with...as" statements.

Thanks in advance

### DIFF
--- a/numba/core/errors.py
+++ b/numba/core/errors.py
@@ -544,8 +544,8 @@ class NotDefinedError(IRError):
 
     def __init__(self, name, loc=None):
         self.name = name
-        msg = "The compiler failed to analyze the bytecode\n \
-               Variable '%s' is not defined." % name
+        msg = ("The compiler failed to analyze the bytecode. "
+               "Variable '%s' is not defined." % name)
         super(NotDefinedError, self).__init__(msg, loc=loc)
 
 
--- a/numba/core/errors.py
+++ b/numba/core/errors.py
@@ -544,7 +544,8 @@ class NotDefinedError(IRError):
 
     def __init__(self, name, loc=None):
         self.name = name
-        msg = "The compiler failed to analyze the bytecode"
+        msg = "The compiler failed to analyze the bytecode\n \
+               Variable '%s' is not defined." % name
         super(NotDefinedError, self).__init__(msg, loc=loc)
 
 
--- a/numba/core/errors.py
+++ b/numba/core/errors.py
@@ -544,7 +544,7 @@ class NotDefinedError(IRError):
 
     def __init__(self, name, loc=None):
         self.name = name
-        msg = "Variable '%s' is not defined." % name
+        msg = "The compiler failed to analyze the bytecode"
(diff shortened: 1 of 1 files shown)


gmarkall:
Thanks for the PR! The CI failure seems to be due to the ongoing issue #5973, so I'll mark this as ready for review.

numba/core/errors.py
```diff
@@ -544,7 +544,7 @@ class NotDefinedError(IRError):
 
     def __init__(self, name, loc=None):
         self.name = name
-        msg = "Variable '%s' is not defined." % name
+        msg = "The compiler failed to analyze the bytecode"
```
stuartarchibald, line 547: Perhaps combine the two messages? It's important to tell the user that the compiler failed to analyse the bytecode, but also useful to say why.

stuartarchibald (commented):
Thanks for the patch, minor suggestion to resolve else looks good.

numba/core/errors.py
Luiz6ustav0: That does make sense. I just committed the change you suggested, where I merged the two messages into one.

Luiz6ustav0:
I think this is ready for review again, I'm not sure if I need to comment on this but just to make sure.

numba/core/errors.py
```diff
@@ -544,7 +544,8 @@ class NotDefinedError(IRError):
 
     def __init__(self, name, loc=None):
         self.name = name
-        msg = "Variable '%s' is not defined." % name
+        msg = "The compiler failed to analyze the bytecode\n \
+               Variable '%s' is not defined." % name
```
stuartarchibald, line 548:
```suggestion
        msg = ("The compiler failed to analyze the bytecode. "
               "Variable '%s' is not defined." % name)
```

The existing patch produces this:
```
numba.core.errors.NotDefinedError: Failed in nopython mode pipeline (step: analyzing bytecode)
The compiler failed to analyze the bytecode
                Variable '$4setup_with.1' is not defined.
```
the suggestion produces this:
```
numba.core.errors.NotDefinedError: Failed in nopython mode pipeline (step: analyzing bytecode)
The compiler failed to analyze the bytecode. Variable '$4setup_with.1' is not defined.
```

Luiz6ustav0: Thank you for the suggestion, I've just committed it. Quite a coincidence when I tested it with my terminal window and it didn't look like that.

stuartarchibald (approved):
Thanks for the fix.

stuartarchibald:
@Luiz6ustav0 Congratulations on your first contribution to Numba!

Luiz6ustav0:
Thanks! Many more to come :)

PR #6194, opened 2020-08-31
Remove recursion from traverse_types

### LINKED ISSUE #6180: Type inference recursion error for self referencing types
I don't know if this should be labeled as a bug or missing feature, but is worth exposing it. Basicaly, here I am trying to self reference(forward declaration) a structref without using the undocumented DeferredType. 
<details><summary>example code</summary>

```
import numba as nb
from numba import jit, njit, types, typed, float64, int64, typeof
from numba.core.extending import overload_method
from numba.experimental.structref import new, register

@structref.register
class PolygonStructType(types.StructRef):

    def preprocess_fields(self, fields):
        self.name = f"numba.PolygonStructType#{id(self)}" # temp name to allow Optional instantiation, this workaround can be replaced by extending StructRef or making StructRef to assign itself a name before preprocess_fields is called
        fields = tuple([
            ('value', types.Optional(int64)),
            ('parent', types.Optional(self)),
            ])

        return fields

polygon_struct_type = PolygonStructType(fields=(('value', types.Any), ('parent', types.Any)))

class PolygonStruct(structref.StructRefProxy):
    def __new__(cls, value, parent):
        return structref.StructRefProxy.__new__(cls, value, parent)

    @property
    def value(self):
        return PolygonStruct_get_value(self)

    @property
    def parent(self):
        return PolygonStruct_get_parent(self)

@njit
def PolygonStruct_get_value(self):
    return self.value

@njit
def PolygonStruct_get_parent(self):
    return self.parent

structref.define_proxy(PolygonStruct, PolygonStructType, ["value", "parent"])

@overload_method(PolygonStructType, "flip")
def PolygonStruct_flip(self):
    def impl(self):
        if self.value is not None:
            self.value = -self.value
    return impl

@njit
def test():
    # p = PolygonStruct(0, None)
    p = new(polygon_struct_type)
    p2 = new(polygon_struct_type)
    p2.value = 1
    p.parent = p2
    p.value = 3
    
    p.flip()
    return p.value

y = test()
print(y)
```
</details>

IMO, the usage of recursive algorithms in general is not recommended for production code.
And forward declaration for types is an essential feature in strong typed languages. There is no real workaround for it. I don't understand why numba does not have it already..

rpopovici:
this is intended to fix #6180


numba/core/datamodel/models.py
```diff
@@ -103,11 +103,23 @@ def traverse_types(self):
         """
         Recursively list all frontend types involved in this model.
         """
-        return [self._fe_type] + self.inner_types()
+        types = [self._fe_type]
+        queue = [self]
+        while len(queue) > 0:
+            dm = queue[0]
+            queue.pop(0)
```
sklam, line 110: `list.pop(0)` is O(n). Use `collections.deque` instead of `list` for `queue` to be more efficient.

rpopovici:
@sklam I did what you asked. There is one issue thus. Boxing/unboxing does not work with object instances referencing objects of the same type, but I guess that is a different problem than this one. I will open another ticket for that when this is merged

sklam:
Thanks for the patches. The CI is currently failing on flake8 problems.

sklam (approved):

PR #6187, opened 2020-08-29
[WIP] Don't visit already checked successors

gmarkall:
Just making the PR to try a (very tentative) fix for #6186 on CI.

Issue #6186, opened 2020-08-29
DFS-based backedge algorithm can be slow for pathological cases


gmarkall:
cc @sklam @quasiben @philippjfr @AjayThorve

See https://github.com/holoviz/datashader/issues/956

No Numba-only reproducer as yet, but the DFS-based algorithm increases the runtime of the reproducer in the Datashader issue from 10 seconds to 34 minutes. Going back to the old algorithm with:

```diff
diff --git a/numba/core/controlflow.py b/numba/core/controlflow.py
index f68572031..f37e9ae3d 100644
--- a/numba/core/controlflow.py
+++ b/numba/core/controlflow.py
@@ -572,40 +572,14 @@ class CFGraph(object):
         Find back edges.  An edge (src, dest) is a back edge if and
         only if *dest* dominates *src*.
         """
-
-        # Uses a simple DFS to find back-edges.
-        # The new algorithm is faster than the the previous dominator based
-        # algorithm.
         back_edges = set()
-        # stack: keeps track of the traversal path
-        stack = []
-        # succs_state: keep track of unvisited successors of a node
-        succs_state = {}
-        entry_point = self.entry_point()
-
-        def push_state(node):
-            stack.append(node)
-            succs_state[node] = [dest for dest in self._succs[node]]
-
-        push_state(entry_point)

-        while stack:
-            tos = stack[-1]
-            tos_succs = succs_state[tos]
-            # Are there successors not checked?
-            if tos_succs:
-                # Check the next successor
-                cur_node = tos_succs.pop()
-                # Is it in our traversal path?
-                if cur_node in stack:
-                    # Yes, it's a backedge
-                    back_edges.add((tos, cur_node))
-                else:
-                    # Push
-                    push_state(cur_node)
-            else:
-                # Checked all successors. Pop
-                stack.pop()
+        for src, succs in self._succs.items():
+            back = self._doms[src] & succs
+            # In CPython bytecode, at most one back edge can flow from a
+            # given block.
+            assert len(back) <= 1
+            back_edges.update((src, dest) for dest in back)

         return back_edges

```

Brings the runtime back down again:

```
$ time python repro.py
x_range: (1285891200000000000, 1291891140000000000) y_range: (-104.8094774985483, 100.09816883332905)

real    0m9.732s
user    0m9.601s
sys     0m0.113s
```

gmarkall:
I'm suspicious of the implementation of `push_state`:

```python
        def push_state(node):
            stack.append(node)
[...]

PR #6195, opened 2020-08-31
PR 6187 Continue. Don't visit already checked successors

sklam:
- Build on #6187 
- Add test with the reproducer in #6186 and check the iteration count, which shouldn't change unless the algorithm is changed.
- Improved the fix in #6187. Use `set` and reduce iteration count in the backedge finding algorithm.


numba/tests/test_flow_control.py
```diff
@@ -950,6 +950,79 @@ def test_loops(self):
...
+            758: {778},
+            778: {722},
+            816: {818},
+            818: set(),
+        })
+        g.set_entry_point(0)
+        g.process()
+        stats = {}
+        # Compute backedges and store the iteration count for testing
+        g._find_back_edges(stats=stats)
```
gmarkall, line 1023: Would it be a good idea to also check the set of back edges?

```suggestion
        back_edges = g._find_back_edges(stats=stats)
        self.assertEqual(back_edges, {(666, 610), (778, 722)})
```

gmarkall (commented):
The code changes look good to me. Just a question about the tests on the diff.

numba/tests/test_flow_control.py
sklam: Good idea.

sklam:
I have found a python source that will trigger this behavior:

```python
from numba import njit


@njit
def foo(a):
    x = 0
    if a < 1 or a < 2 or a < 3:
        x += 1
    else:
        x += 1

    if a < 1 or a < 2:
        x += 1

    if a < 1:

        if a < 1 or a < 2 or a < 3:
            x += 1
        else:
            x += 1

    else:

        if a < 1 or a < 2:
            x += 1

    if a < 1:
        x += 1

    if a < 1:
        x += 1

    if a < 1:
        x += 1

    if a < 1:
        x += 1

    if a < 1:
        x += 1

    for i in range(a):

        if a < 1 or a < 2 or a < 3:
            x += 1
        else:
            x += 1

        if a < 1 or a < 2:
            x += 1

        if a < 1:

            if a < 1 or a < 2 or a < 3:
                x += 1
            else:
                x += 1

    else:

        if a < 1 or a < 2 or a < 3:
            x += 1
        else:
            x += 1

        if a < 1 or a < 2:
            x += 1

        if a < 1:

            if a < 1 or a < 2 or a < 3:
                x += 1
            else:
                x += 1

        else:

            if a < 1 or a < 2:
                x += 1
    return x


foo(10)
```

numba/core/controlflow.py
```diff
@@ -583,13 +588,17 @@ def _find_back_edges(self):
...
+
         def push_state(node):
             stack.append(node)
             succs_state[node] = [dest for dest in self._succs[node]]
 
         push_state(entry_point)
 
         while stack:
+            if stats is not None:
+                stats['iteration_count'] += 1
```
sklam, line 601: TODO: lift this branch out of the loop and just do a simple counter increment in the loop

gmarkall (approved):
This looks good to me!

stuartarchibald (approved):
Thanks for fixing this @gmarkall and @sklam. I've tested the backedge alg on mainline against the "pathological" case, takes about 11s locally, with this PR it takes 0.001s. Code looks good.

stuartarchibald:
Tests are all passing on the checks page: https://github.com/numba/numba/pull/6195/checks it's reported incorrectly above.

PR #6201, opened 2020-09-01
Update CHANGE_LOG for 0.51.2

stuartarchibald:
As title.

### DIFF
--- a/CHANGE_LOG
+++ b/CHANGE_LOG
@@ -4,6 +4,21 @@ Version 0.52.0
 In development
 
 
+Version 0.51.2 (September 2, 2020)
+----------------------------------
+
+This is a bugfix release for 0.51.1, it fixes a critical performance bug in the
+CFG back edge computation algorithm that lead to exponential time complexity
+arising in compilation for use cases with certain pathological properties.
+
+* PR #6195: PR 6187 Continue. Don't visit already checked successors
+
+Authors:
+
+* Graham Markall
+* Siu Kwan Lam (core dev)
+
+
 Version 0.51.1 (August 26, 2020)
 --------------------------------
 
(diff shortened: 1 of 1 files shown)


gmarkall (approved):
Looks good to me.

esc:
I think the description is a bit long and unwieldy, how about:

```
This is a bugfix release for 0.51.1. It fixes a critical performance bug in the
control flow graph (CFG) back edge computation algorithm, which leads to
exponential time complexity when compiling use cases with certain
pathological properties.
```

ehsantn:
I'd avoid using the word "pathological" :) something like "corner case" is much better.

gmarkall:
> I'd avoid using the word "pathological" :) something like "corner case" is much better.

I've tried googling, checking definitions, and looking at the etymology to see why it would be better, but I can't find anything that suggests there's something wrong with using that word. Can you explain why it would be better to talk about a "corner case"?

(to me, "corner case" also appears less precise, which is why I'd picked the word "pathological" instead of "corner case" originally)

ehsantn:
Not a big deal. Just a minor suggestion.

sklam:
I'm preferring "pathological" over "corner case" because to me the latter sounds uncommon but I am afraid that CFG pattern is not uncommon.

CHANGE_LOG
```diff
@@ -4,6 +4,21 @@ Version 0.52.0
 In development
 
 
+Version 0.51.2 (September 2, 2020)
+----------------------------------
+
+This is a bugfix release for 0.51.1, it fixes a critical performance bug in the
```
sklam, line 10:
```suggestion
This is a bugfix release for 0.51.1. It fixes a critical performance bug in the
```

CHANGE_LOG
```diff
@@ -4,6 +4,21 @@ Version 0.52.0
...
+CFG back edge computation algorithm that lead to exponential time complexity
```
sklam, line 11:
```suggestion
CFG back edge computation algorithm that leads to exponential time complexity
```

PR #6202, opened 2020-09-01
Speed up str(int)

### LINKED ISSUE #6189: Integer to string conversion is extremely slow
## Reporting a bug

Integer to string conversion (`str(int)`) is extremely slow in Numba (~10x slower than Python and ~100x slower than C). The implementation ([here](https://github.com/numba/numba/blob/0db523ea8e0e5ad95d8a4262b4c9c44b65ecbc25/numba/cpython/unicode.py#L2353)) looks very inefficient since it allocates every single produced character. My guess is that someone copied it from CPython, but unlike Numba, CPython has string interning. The integer to string implementation can be much more efficient. Also, SSO (#3965) can avoid these allocations for many string programs in general.

```python
In [7]: def f(n):
   ...:     c = 0
   ...:     for i in range(n):
   ...:         c += len(str(n))
   ...:     return c
   ...:

In [8]: %timeit f(1000_000)
194 ms ± 3.17 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [10]: g = numba.njit(f)

In [11]: g(1000_000)
Out[11]: 7000000

In [12]: %timeit g(1000_000)
1.97 s ± 68.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

stuartarchibald:
As title.

Fixes #6189

### DIFF
--- a/numba/cpython/unicode.py
+++ b/numba/cpython/unicode.py
@@ -2352,6 +2352,7 @@ def ol_chr(i):
 @overload(str)
 def integer_str(n):
     if isinstance(n, types.Integer):
+
         ten = n(10)
 
         def impl(n):
@@ -2361,15 +2362,22 @@ def integer_str(n):
                 flag = True
             if n == 0:
                 return '0'
-            l = []
-            while n > 0:
-                c = chr(ord('0') + (n % ten))
-                n = n // ten
-                l.append(c)
+            length = flag + 1 + int(np.floor(np.log10(n)))
+            kind = PY_UNICODE_1BYTE_KIND
+            char_width = _kind_to_byte_width(kind)
+            s = _malloc_string(kind, char_width, length, True)
+            const = ord('0')
             if flag:
-                l.append('-')
-            return ''.join(l[::-1])
+                _set_code_point(s, 0, ord('-'))
+            idx = length - 1
+            while n > 0:
+                c = const + (n % ten)
+                n = n // ten
+                _set_code_point(s, idx, c)
+                idx -= 1
+            return s
         return impl
+
 
 # ------------------------------------------------------------------------------
 # iteration
--- a/numba/cpython/unicode.py
+++ b/numba/cpython/unicode.py
@@ -2366,13 +2366,12 @@ def integer_str(n):
             kind = PY_UNICODE_1BYTE_KIND
             char_width = _kind_to_byte_width(kind)
             s = _malloc_string(kind, char_width, length, True)
-            const = ord('0')
             if flag:
                 _set_code_point(s, 0, ord('-'))
             idx = length - 1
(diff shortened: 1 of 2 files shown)


stuartarchibald:
Tested speed with:
```python
from numba import njit
import time

@njit
def foo(n):
    c = 0
    for i in range(n):
        c += len(str(n))
    return c

def test(f):
    N = 10
    f(1)
    ts = time.time()
    for i in range(N):
        f(1000_000 + i)
    te = time.time()
    print("Elapsed = %s" % (te - ts))

test(foo.py_func)
test(foo)
```
This implementation seems at least twice as fast as cpython.

numba/cpython/unicode.py
```diff
@@ -2361,16 +2362,23 @@ def impl(n):
...
+            length = flag + 1 + int(np.floor(np.log10(n)))
+            kind = PY_UNICODE_1BYTE_KIND
+            char_width = _kind_to_byte_width(kind)
+            s = _malloc_string(kind, char_width, length, True)
+            const = ord('0')
+            if flag:
+                _set_code_point(s, 0, ord('-'))
+            idx = length - 1
             while n > 0:
-                c = chr(ord('0') + (n % ten))
```
eric-wieser, line 2366: Is llvm not able to infer that `ord('0')` is a constant?

numba/cpython/unicode.py
```diff
@@ -2361,16 +2362,23 @@ def impl(n):
...
+                c = const + (n % ten)
                 n = n // ten
```
eric-wieser, line 2375: Probably doesn't make any difference, but fewer operators as:
```suggestion
                n, digit = divmod(n, ten)
                c = const + digit
```

numba/cpython/unicode.py
stuartarchibald: expect it can, will just use it directly.

numba/cpython/unicode.py
stuartarchibald: heh, yes, thanks, also, pretty sure Numba internally will just use divmod for doing `%` and `//` and LLVM will sort this out. Will make this change though as it's neater.

esc (approved):
I tried this locally and it does accelerate things quite a bit:

Before
```
 💣 zsh» python issue6202.py
Elapsed = 1.975172758102417
Elapsed = 20.156193017959595
```

After:

```
 💣 zsh» python issue6202.py
Elapsed = 2.0004079341888428
Elapsed = 0.9646546840667725
```

stuartarchibald:
@esc is this ready to merge?

esc:
@stuartarchibald it sure is! I have updated the label accordingly.

PR #6203, opened 2020-09-02
Make find_callname only lookup functions that are likely part of NumPy.

### LINKED ISSUE #6175: ir_utils causes a DeprecationWarning in numpy 1.20
These lines:

https://github.com/numba/numba/blob/c99f3dea5f78f8268b361f06fa7800d1b656bc20/numba/core/ir_utils.py#L1525-L1527

result in the following behavior in numpy 1.20 due to numpy/numpy#14882
```python
In [8]: @numba.njit
   ...: def foo():
   ...:     return bool(1)
   ...:

In [9]: foo()
c:\users\wiese\repos\ga\numba\numba\core\ir_utils.py:1525: DeprecationWarning: `np.bool` is a deprecated alias for the builtin `bool`. Use `bool` by itself, which is identical in behavior, to silence this warning. If you specifically wanted the numpy scalar type, use `np.bool_` here.
  if (hasattr(numpy, value)
c:\users\wiese\repos\ga\numba\numba\core\ir_utils.py:1526: DeprecationWarning: `np.bool` is a deprecated alias for the builtin `bool`. Use `bool` by itself, which is identical in behavior, to silence this warning. If you specifically wanted the numpy scalar type, use `np.bool_` here.
  and def_val == getattr(numpy, value)):
c:\users\wiese\repos\ga\numba\numba\core\ir_utils.py:1525: DeprecationWarning: `np.bool` is a deprecated alias for the builtin `bool`. Use `bool` by itself, which is identical in behavior, to silence this warning. If you specifically wanted the numpy scalar type, use `np.bool_` here.
  if (hasattr(numpy, value)
c:\users\wiese\repos\ga\numba\numba\core\ir_utils.py:1526: DeprecationWarning: `np.bool` is a deprecated alias for the builtin `bool`. Use `bool` by itself, which is identical in behavior, to silence this warning. If you specifically wanted the numpy scalar type, use `np.bool_` here.
  and def_val == getattr(numpy, value)):
Out[9]: True
```

This doesn't make any sense to me - why is `numba` even looking at numpy when I use a builtin function?

stuartarchibald:
As title.

Closes #6175

### DIFF
--- a/numba/core/ir_utils.py
+++ b/numba/core/ir_utils.py
@@ -1527,7 +1527,8 @@ def find_callname(func_ir, expr, typemap=None, definition_finder=get_definition)
                 # other common names. This prevents deprecation warnings on
                 # e.g. getattr(numpy, 'bool') were a bool the the target.
                 # For context see #6175, impacts NumPy>=1.20.
-                numpy_toplevel = mod_name.startswith('numpy')
+                mod_not_none = mod_name is not None
+                numpy_toplevel = mod_not_none and mod_name.startswith('numpy')
                 # it might be a numpy function imported directly
                 if (numpy_toplevel and hasattr(numpy, value)
                         and def_val == getattr(numpy, value)):
@@ -1536,7 +1537,7 @@ def find_callname(func_ir, expr, typemap=None, definition_finder=get_definition)
                 elif (hasattr(numpy.random, value)
                         and def_val == getattr(numpy.random, value)):
                     attrs += ['random', 'numpy']
-                elif mod_name is not None:
+                elif mod_not_none:
                     attrs.append(mod_name)
             else:
                 class_name = def_val.__class__.__name__
--- a/numba/core/ir_utils.py
+++ b/numba/core/ir_utils.py
@@ -1527,7 +1527,7 @@ def find_callname(func_ir, expr, typemap=None, definition_finder=get_definition)
                 # other common names. This prevents deprecation warnings on
                 # e.g. getattr(numpy, 'bool') were a bool the the target.
                 # For context see #6175, impacts NumPy>=1.20.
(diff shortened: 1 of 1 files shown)


stuartarchibald:
CC @eric-wieser

numba/core/ir_utils.py
```diff
@@ -1521,15 +1521,23 @@ def find_callname(func_ir, expr, typemap=None, definition_finder=get_definition)
                 def_val = def_val._defn
             if hasattr(def_val, '__module__'):
                 mod_name = def_val.__module__
+                # The reason for first checking if the function is in NumPy's
+                # top level name space by module is that some functions are
+                # deprecated in NumPy but the functions' names are aliased with
+                # other common names. This prevents deprecation warnings on
+                # e.g. getattr(numpy, 'bool') were a bool the the target.
```
eric-wieser, line 1528:
```suggestion
                # e.g. getattr(numpy, 'bool') were a bool the target.
```

numba/core/ir_utils.py
```diff
@@ -1521,15 +1521,23 @@ def find_callname(func_ir, expr, typemap=None, definition_finder=get_definition)
...
+                # For context see #6175, impacts NumPy>=1.20.
+                mod_not_none = mod_name is not None
+                numpy_toplevel = mod_not_none and mod_name.startswith('numpy')
```
eric-wieser, line 1531:
```suggestion
                numpy_toplevel = mod_not_none and (mod_name == 'numpy' or mod_name.startswith('numpy.'))
```
so as not to cause a warnings for `numpy_custom_module.bool`.

stuartarchibald: Good point, done in c3bd15bc32cc2571267454b9d8b422663909de5d

sklam (approved):

PR #6204, opened 2020-09-02
Fix unicode kind selection for getitem.

### LINKED ISSUE #6135: Unicode characters as dict keys don't work reliably due to character width ambiguity
The following code breaks with a KeyError:

```
@njit
def test():
    s = 'a\u1234'
    d = {s[0] : 1}
    print(d['a'])

test()
```

The presence of a wider than 1 byte character in the string "s" triggers the problem. Quote from the documentation:

> String slices also use the same character width as the original string, even if the slice could be represented with a narrower character width. (These details are invisible to the user, of course.)

I suspect that the different character width results in a different hash, which breaks dictionary lookups using such strings as keys. I found this workaround for my application:

```
@njit
def fixuni(c):
    u = c.upper()
    return u if u == c else c.lower()

@njit
def test():
    s = 'a\u1234'
    d = {fixuni(s[0]) : 1}
    print(d['a'])

test()
```

stuartarchibald:
As title.

Fixes #6135

### DIFF
--- a/numba/cpython/unicode.py
+++ b/numba/cpython/unicode.py
@@ -1644,8 +1644,11 @@ def unicode_getitem(s, idx):
         if isinstance(idx, types.Integer):
             def getitem_char(s, idx):
                 idx = normalize_str_idx(idx, len(s))
-                ret = _empty_string(s._kind, 1, s._is_ascii)
-                _set_code_point(ret, 0, _get_code_point(s, idx))
+                cp = _get_code_point(s, idx)
+                kind = _codepoint_to_kind(cp)
+                is_ascii = _codepoint_is_ascii(cp)
+                ret = _empty_string(kind, 1, is_ascii)
+                _set_code_point(ret, 0, cp)
                 return ret
             return getitem_char
         elif isinstance(idx, types.SliceType):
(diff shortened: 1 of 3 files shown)


numba/cpython/unicode.py
```diff
@@ -1644,24 +1644,49 @@ def unicode_getitem(s, idx):
...
+                # Check slice to see if it's homogeneous in kind
+                for i in range(slice_idx.start + slice_idx.step,
+                               slice_idx.stop, slice_idx.step):
+                    cp = _get_code_point(s, i)
+                    is_ascii |= _codepoint_is_ascii(cp)
+                    new_kind = _codepoint_to_kind(cp)
+                    if kind != new_kind:
+                        kind = _pick_kind(kind, new_kind)
+                    # TODO: it might be possible to break here if the kind
+                    # is PY_UNICODE_4BYTE_KIND but there's potentially
```
esc, line 1672: Perhaps it would be good to enhance this description with an additional sentence as to what the symptoms of breaking here might be?

numba/tests/test_unicode.py
```diff
@@ -1173,7 +1177,19 @@ def test_getitem(self):
...
         for s in UNICODE_EXAMPLES:
-            for i in range(-len(s)):
+            for i in range(-len(s), len(s)):
+                self.assertEqual(pyfunc(s, i),
+                                 cfunc(s, i),
+                                 "'%s'[%d]?" % (s, i))
+
+    def test_getitem_scalar_kind(self):
+        # See issue #6135, make sure that getitem returns a char of the minimal
+        # kind required to represent the "got" item.
```
esc, line 1187: You may want to indicate how condition is ensured here. Also, for the ones below.

esc (changes requested):
Some things to consider regarding the comments, looks good otherwise. I also tested it on the original issue #6135 and can confirm it will resolve that.

numba/cpython/unicode.py
stuartarchibald: I'm not sure what this means? If the kind is already the widest kind then break, as looping to look for a bigger kind is redundant. Doing this would require that the strings are of the minimal kind required to hold their code point, but as noted in the comment, I suspect other internal functions aren't necessarily doing the correct thing.

numba/tests/test_unicode.py
stuartarchibald: hash does it? is that what you are implying?

stuartarchibald:
Thanks for the review!

numba/tests/test_unicode.py
esc: Yes, this is what I assumed, but maybe write "use ` hash` to ensure .... as it does X`?

numba/cpython/unicode.py
esc: ah, so some internal stings have a kind larger than required to hold their code point?

stuartarchibald: > ah, so some internal stings have a kind larger than required to hold their code point?

I suspect yes. This is on the basis that it hasn't been enforced.

numba/tests/test_unicode.py
stuartarchibald: Done in bab2a17

esc: 👌

numba/cpython/unicode.py
```diff
@@ -1644,24 +1644,49 @@ def unicode_getitem(s, idx):
...
+                    # strings coming from other internal functions that are
+                    # this wide and also actually ASCII, so it's necessary
```
esc, line 1674:
```suggestion
                    # this wide and also actually ASCII (i.e. the kind is larger than required), so it's necessary
```

numba/cpython/unicode.py
esc: I think it would be worth adding a note about this insight. I left a suggestion below on how to make this comment a little clearer. Feel free to adopt that or suggest a better alternative.

numba/cpython/unicode.py
stuartarchibald: Done similar in abd8430 .

esc (approved):
Thank you for addressing all the comments is good to merge now.

stuartarchibald:
> Thank you for addressing all the comments is good to merge now.

Great, thanks for the review.

PR #6206, opened 2020-09-02
Build all extension modules with -g -Wall -Werror on Linux x86, provide -O0 flag option

gmarkall:
The main part of this PR is building extension modules with `-Werror` and `-Wall` on Linux x86 platforms. I considered trying to make it build with these flags on all platforms, but I suspect it will introduce too many maintenance problems. Building with `-Wall` or `-Werror` can be enabled on a platform that doesn't have it by default by adding the `--wall` and `--werror` flags to `build_ext`.

The PR also provides a `--noopt` flag, that builds with `-O0` (likely to work on Linux and OS X) to make debugging compiled extensions easier. By default, extensions are built with debug symbols on OS X and Linux - this was the case before for a couple of extensions due to the unconditional addition of the `-g` flag, which took effect on OS X and Linux, but on Windows it was ignored.

Some changes were required to make the build succeed with `-Wall` and `-Werror`:

- Add some missing initialization of struct members, for `PyMemberDef`, `PyTypeObject` (`_dispatcher.c`, `_dynfunc.c`, and `_pymodule.h`).
- Remove some unused parameters from functions (`closure_new` in `_dynfunc.c`, `fingerprint_unrecognized` in `_typeof.c`)
- Removal of Python 2/Python<3.6-related  defines in `_pymodule.h`. There's some further cleanup that can be done here to make other code less surprising - e.g. removing the definitions of `PyString_*` as `PyUnicode_*` and replacing their uses in the rest of the source, but I've left that for now.
- Guarding shifts greater than 32 bits with a define instead of conditional in `string_writer_put_intp` in `_typeof.c` - on 32 bit systems the shift greater than 32 bits is undefined, though note it could not be executed (`N` was always 4 on 32 bit systems).
- Removal of support for the old buffer protocol, whose functions are deprecated (`mviewbuf.c`).
- Use of correct length modifiers for formatting `size_t` and `ptrdiff_t` in strings (`omppool.cpp` and `workqueue.c`).

### DIFF
--- a/buildscripts/condarecipe.local/bld.bat
+++ b/buildscripts/condarecipe.local/bld.bat
@@ -1,3 +1,3 @@
-%PYTHON% setup.py build --wall install --single-version-externally-managed --record=record.txt
+%PYTHON% setup.py build install --single-version-externally-managed --record=record.txt
 
 exit /b %errorlevel%
--- a/buildscripts/condarecipe.local/bld.bat
+++ b/buildscripts/condarecipe.local/bld.bat
@@ -1,3 +1,3 @@
-%PYTHON% setup.py build install --single-version-externally-managed --record=record.txt
+%PYTHON% setup.py build --wall install --single-version-externally-managed --record=record.txt
 
 exit /b %errorlevel%
--- a/buildscripts/condarecipe.local/build.sh
+++ b/buildscripts/condarecipe.local/build.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-MACOSX_DEPLOYMENT_TARGET=10.10 $PYTHON setup.py build install --single-version-externally-managed --record=record.txt
+MACOSX_DEPLOYMENT_TARGET=10.10 $PYTHON setup.py build_ext --inplace --werror --wall build install --single-version-externally-managed --record=record.txt
--- a/buildscripts/condarecipe.local/build.sh
+++ b/buildscripts/condarecipe.local/build.sh
@@ -6,4 +6,4 @@ else
     EXTRA_BUILD_EXT_FLAGS=""
 fi
 
-MACOSX_DEPLOYMENT_TARGET=10.10 $PYTHON setup.py build_ext --inplace $EXTRA_BUILD_EXT_FLAGS build install --single-version-externally-managed --record=record.txt
+MACOSX_DEPLOYMENT_TARGET=10.10 $PYTHON setup.py build_ext $EXTRA_BUILD_EXT_FLAGS build install --single-version-externally-managed --record=record.txt
--- a/buildscripts/condarecipe.local/build.sh
+++ b/buildscripts/condarecipe.local/build.sh
@@ -1,3 +1,9 @@
 #!/bin/bash
 
-MACOSX_DEPLOYMENT_TARGET=10.10 $PYTHON setup.py build_ext --inplace --werror --wall build install --single-version-externally-managed --record=record.txt
+if [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
+    EXTRA_BUILD_EXT_FLAGS="--werror --wall"
+else
+    EXTRA_BUILD_EXT_FLAGS=""
+fi
+
+MACOSX_DEPLOYMENT_TARGET=10.10 $PYTHON setup.py build_ext --inplace $EXTRA_BUILD_EXT_FLAGS build install --single-version-externally-managed --record=record.txt
--- a/buildscripts/condarecipe.local/build.sh
+++ b/buildscripts/condarecipe.local/build.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-if [ "$(expr substr $(uname -s) 1 5)" == "Linux" ] && [[ "$(uname -p)" == *"86"* ]]; then
+if [[ "$(uname -s)" == *"Linux"* ]] && [[ "$(uname -p)" == *"86"* ]]; then
     EXTRA_BUILD_EXT_FLAGS="--werror --wall"
 else
     EXTRA_BUILD_EXT_FLAGS=""
--- a/buildscripts/condarecipe.local/build.sh
+++ b/buildscripts/condarecipe.local/build.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-if [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
+if [ "$(expr substr $(uname -s) 1 5)" == "Linux" ] && [[ "$(uname -p)" == *"86"* ]]; then
     EXTRA_BUILD_EXT_FLAGS="--werror --wall"
 else
     EXTRA_BUILD_EXT_FLAGS=""
--- a/buildscripts/incremental/build.sh
+++ b/buildscripts/incremental/build.sh
@@ -6,7 +6,7 @@ source activate $CONDA_ENV
 set -v -e
 
 # Build numba extensions without silencing compile errors
-if [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
+if [[ "$(uname -s)" == *"Linux"* ]] && [[ "$(uname -p)" == *"86"* ]]; then
     EXTRA_BUILD_EXT_FLAGS="--werror --wall"
 else
     EXTRA_BUILD_EXT_FLAGS=""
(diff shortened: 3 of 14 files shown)


gmarkall:
A buildfarm run of this would be really handy prior to review, in case it turns up problems on platforms I don't have access to - many thanks! (cc @esc @sklam )

esc:
BFID: `numba_smoketest_cuda_99`

esc:
The build farm failed on this PR due to some network glitches, I have just retriggered it.

esc:
The build farm has passed.

gmarkall:
> The build farm has passed.

Many thanks!

stuartarchibald:
a68652f  was from an OOB discussion, because it was confusing!

setup.py
```diff
@@ -36,18 +37,56 @@ def run(self):
...
+        # platforms.
+        is_linux_x86 = (sys.platform == 'linux' and platform.machine()
+                        in ('i386', 'x86_64'))
+        self.werror = is_linux_x86
+        self.wall = is_linux_x86
+        # Building with debug symbols should generally work on Linux and OS X.
+        self.debug = sys.platform in ('linux', 'darwin')
+        # Optional build without optimization, to make debugging extensions
+        # easier.
+        self.noopt = 0
```
stuartarchibald, line 71:
```suggestion
        self.noopt = False
```

gmarkall: I would have done that, but I was following the convention in distutils command classes.

gmarkall:
Following some more OOB discussion, I'm also going to:

- Document `--noopt`
- Make `-Werror` optional, add a flag for it (`--werror`), and document it.
- Add the `--werror` flag to the CI setup and conda builds.

This way we get to test with `-Werror` and build recipes with it, ensuring that we test with the flag in general, but avoid issues with `-Werror` for users who build themselves with a newer GCC / Clang than we test with that triggers additional warnings.

stuartarchibald:
Notes from OOB conversation.
1. It's worth adding a short note to the documentation about these options now they are implemented and consistent.
2. `-Werror` will be off by default as it may well cause problems for people with newer compilers and/or on untested compiler/hardware combinations.
3. Conda package builds and CI will have `-Werror` enabled.

setup.py
stuartarchibald: ah ok, I guess leave as is :)

stuartarchibald:
OOB conversation with other core devs. Going to see what impact `-g` has on binary size. Likely remove for default builds (conda, pip) but keep in the CI runs so that tracebacks on remote/inaccessible systems provide hints.

gmarkall:
An update on the changes:

- I've added documentation for the `--noopt`, `--werror`, and `--wall` options.
- `-Werror` and `-Wall` are now both optional, but enabled in CI and the Conda recipe for Linux.
- Disabling building with debug by default on Linux made no difference to the size of the binaries, because the build was already passing `-g`. I've removed the default setting of `-g`, because it is either superfluous (Anaconda compilers on Linux, probably most other Python distributions on Linux / macOS) or unwanted (Windows).
- I've modified the `--noopt` option so that it passes `-Od` to disable optimizations on Windows.

stuartarchibald (approved):
Thanks for the patch and the fixes, looks good.

stuartarchibald:
Buildfarm ID: `numba_smokestest_cpu_67`.

stuartarchibald:
Windows conda builds are failing at compile time with `error: option -wall not recognised` and then giving the usual `usage setup.py ....` response.

buildscripts/condarecipe.local/bld.bat
```diff
@@ -1,3 +1,3 @@
-%PYTHON% setup.py build install --single-version-externally-managed --record=record.txt
+%PYTHON% setup.py build --wall install --single-version-externally-managed --record=record.txt
```
gmarkall, line 1:
```suggestion
%PYTHON% setup.py build install --single-version-externally-managed --record=record.txt
```

gmarkall:
> Windows conda builds are failing at compile time with `error: option -wall not recognised` and then giving the usual `usage setup.py ....` response.

Should be fixed - somehow I put `--wall` in the Windows conda recipe by mistake.

stuartarchibald:
Buildfarm ID: `numba_smokestest_cpu_68` is for df23855

stuartarchibald:
> > Windows conda builds are failing at compile time with `error: option -wall not recognised` and then giving the usual `usage setup.py ....` response.
> 
> Should be fixed - somehow I put `--wall` in the Windows conda recipe by mistake.

thanks

buildscripts/condarecipe.local/build.sh
```diff
@@ -1,3 +1,9 @@
 #!/bin/bash
 
-MACOSX_DEPLOYMENT_TARGET=10.10 $PYTHON setup.py build install --single-version-externally-managed --record=record.txt
+if [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
```
gmarkall, line 3:
```suggestion
if [ "$(expr substr $(uname -s) 1 5)" == "Linux" ] && [[ "$(uname -p)" == *"86"* ]]; then
```

buildscripts/condarecipe.local/build.sh
```diff
@@ -1,3 +1,9 @@
 #!/bin/bash
 
-MACOSX_DEPLOYMENT_TARGET=10.10 $PYTHON setup.py build install --single-version-externally-managed --record=record.txt
+if [ "$(expr substr $(uname -s) 1 5)" == "Linux" ] && [[ "$(uname -p)" == *"86"* ]]; then
```
gmarkall, line 3:
```suggestion
if [[ "$(uname -s)" == *"Linux"* ]] && [[ "$(uname -p)" == *"86"* ]]; then
```

buildscripts/incremental/build.sh
```diff
@@ -6,7 +6,12 @@ source activate $CONDA_ENV
 set -v -e
 
 # Build numba extensions without silencing compile errors
-python setup.py build_ext -q --inplace
+if [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
```
gmarkall, line 9:
```suggestion
if [[ "$(uname -s)" == *"Linux"* ]] && [[ "$(uname -p)" == *"86"* ]]; then
```

buildscripts/condarecipe.local/build.sh
```diff
@@ -1,3 +1,9 @@
 #!/bin/bash
 
-MACOSX_DEPLOYMENT_TARGET=10.10 $PYTHON setup.py build install --single-version-externally-managed --record=record.txt
+if [[ "$(uname -s)" == *"Linux"* ]] && [[ "$(uname -p)" == *"86"* ]]; then
+    EXTRA_BUILD_EXT_FLAGS="--werror --wall"
+else
+    EXTRA_BUILD_EXT_FLAGS=""
+fi
+
+MACOSX_DEPLOYMENT_TARGET=10.10 $PYTHON setup.py build_ext --inplace $EXTRA_BUILD_EXT_FLAGS build install --single-version-externally-managed --record=record.txt
```
gmarkall, line 9:
```suggestion
MACOSX_DEPLOYMENT_TARGET=10.10 $PYTHON setup.py build_ext $EXTRA_BUILD_EXT_FLAGS build install --single-version-externally-managed --record=record.txt
```

stuartarchibald:
Buildfarm ID: `numba_smokestest_cpu_69` is for 5cf70cf

stuartarchibald:
> Buildfarm ID: `numba_smokestest_cpu_69` is for [5cf70cf](https://github.com/numba/numba/commit/5cf70cf340491398ca9959a4e6fbf9c8905c2661)

This passed.

stuartarchibald:
@sklam wondering if you wanted to take a look at this too? @gmarkall and I have done a few iterations.

stuartarchibald:
Note also, the use of `0xffffffff` in https://github.com/numba/numba/blob/master/numba/cext/dictobject.c results in signed/unsigned comparison on 32bit arch, a test on the build farm of some commit where `-wall` was enabled for all linux builds picked it up on armv7l.

stuartarchibald:
> @sklam wondering if you wanted to take a look at this too? @gmarkall and I have done a few iterations.

Confirmed ok out of band. Marking for merge.

PR #6209, opened 2020-09-02
Fix deferred type for structref

sklam:
Attempt to fix deferred type

### DIFF
--- a/numba/core/datamodel/models.py
+++ b/numba/core/datamodel/models.py
@@ -1274,10 +1274,14 @@ class DeferredStructModel(CompositeModel):
         self.actual_fe_type = fe_type.get()
 
     def get_value_type(self):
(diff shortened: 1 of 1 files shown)


sklam:
Need to add testcase for https://gist.github.com/sklam/a0203c4afbe78989662d783566855766

rpopovici:
Interesting. @sklam with this PR I get the same boxing error I get with the PR I made #6194, when I call the constructor for python mode. Why?

```
structref.py", line 376, in __new__
    return ctor(*args)
TypeError: _numba_box_() missing 2 required positional arguments: 'ty' and 'mi'
```

sklam:
@rpopovici, not sure where the `_numba_box_` error is coming from yet. When I look at your PR, I thought it's a problem with the self-reference but apparently not that it happened here with deferredtype.

PR #6212, opened 2020-09-03
Fix for objmode recompilation issue

kozlov-alexey:
This PR fixes recompilations of objmode lifted code due to:
   1. different handling of function argument types at compilation and at call time,
   2. no_rewrites=True flag value causing recompilation with and without rewrites.


sklam:
> with first compiling with rewrites and only if that doesn't succeed compiling with no_rewrites=True

@kozlov-alexey, the reason that we need to run with no rewrite and then with rewrites on again is because some rewrites is breaking the IR by mutating it. By first compiling with rewrites, the retry with no rewrites may fail.

>  Also switches off rewrites for ObjModeLiftedWith code.

This makes sense. The objmode doesn't need the rewrites. Would this change alone be enough to address the costly compilation time for you?

AlexanderKalistratov:
@sklam 

>@kozlov-alexey, the reason that we need to run with no rewrite and then with rewrites on again is because some rewrites is breaking the IR by mutating it. By first compiling with rewrites, the retry with no rewrites may fail.

Can we avoid mutation of IR by rewrites? By providing copy of IR to them for example? Or make a copy of IR before starting it's compilation?

Compiling lifted cod twice significantly hits overall compilation time, so something should be done anyway.

sklam:
@AlexanderKalistratov 

> Can we avoid mutation of IR by rewrites? By providing copy of IR to them for example? Or make a copy of IR before starting it's compilation?

I can take a look at these directions.

stuartarchibald:
As a workaround for this, could the performance be regained by putting the contents of the `objmode` block into a python function and then just have the block call the function. e.g.:
```python
@njit
def foo():
   with objmode():
      <large source>
```
replaced with:
```python
def call_source(*args):
    <large source>

@njit
def foo():
   with objmode():
      call_source(*args)
```
that way the compiler has very little to do as it only sees a function call?

kozlov-alexey:
@stuartarchibald Actually, the function using the objmode already looks that way, i.e.
```
def csv_reader_py(
        filepath_or_buffer,
        sep=',',
        ...
    ):
        with objmode(df="<df type repr>"):
        df = pandas_read_csv(filepath_or_buffer,
            names=names,
            skiprows=skiprows,
            parse_dates=[],
            dtype={'a': np.float64, 'b': np.int64, 'c': str, 'd': np.float64},
            usecols=usecols,
            sep=sep,
            delimiter=delimiter,
        )
    return df
```
So it exactly calls one function, but it also has to return obtained value from the objmode, and it also builds a columns dtype dict, which seems to add a lot to the lowering time. And we are going to have this dict captured as global variable, which shows much better times, but still it seems strange recompiling something multiple times when there's no need to?

stuartarchibald:
> @stuartarchibald Actually, the function using the objmode already looks that way, i.e.
> 
> ```
> def csv_reader_py(
>         filepath_or_buffer,
>         sep=',',
>         ...
>     ):
>         with objmode(df="<df type repr>"):
>         df = pandas_read_csv(filepath_or_buffer,
>             names=names,
>             skiprows=skiprows,
>             parse_dates=[],
>             dtype={'a': np.float64, 'b': np.int64, 'c': str, 'd': np.float64},
>             usecols=usecols,
>             sep=sep,
>             delimiter=delimiter,
>         )
>     return df
> ```
> 
> So it exactly calls one function, but it also has to return obtained value from the objmode, and it also builds a columns dtype dict, which seems to add a lot to the lowering time. And we are going to have this dict captured as global variable, which shows much better times, but still it seems strange recompiling something multiple times when there's no need to?

@kozlov-alexey I totally agree with not compiling multiple times when there's no need :) Was just wondering how this was actually causing a problem given the work can be pushed out of the IR. If the columns dtype dict is a constant would partial application work so as to keep most of the pandas interface but avoid having the `build_map` in the bytecode? Also slightly puzzled as to how this is adding much compilation time, seems quick locally, is it simply that it's called a lot?

numba/core/compiler.py
```diff
@@ -646,49 +652,35 @@ def compile_ir(typingctx, targetctx, func_ir, args, return_type, flags,
         # cannot compile it. As a result the following approach is taken:
         # 1. Create some new flags that copy the original ones but switch
         #    off rewrites.
-        # 2. Compile with 1. to get a compile result
-        # 3. Try and compile another compile result but this time with the
-        #    original flags (and IR being rewritten).
-        # 4. If 3 was successful, use the result, else use 2.
+        # 2. If rewrites are on in original flags try to compile with a copy of
+        #    func_ir and original flags (with possibly IR being rewritten)
+        #    ignoring compile exceptions
```
stuartarchibald, line 657:
```suggestion
        #    func_ir and original flags (with IR possibly being rewritten)
        #    ignoring exceptions during compilation.
```

numba/core/compiler.py
```diff
@@ -646,49 +652,35 @@ def compile_ir(typingctx, targetctx, func_ir, args, return_type, flags,
...
+        # 3. If above succeeded and we have compile result return it
+        # 4. Othwerise try and compile another compile result with rewrites off
```
stuartarchibald, line 659:
```suggestion
        # 4. Otherwise try and compile another compile result with rewrites off
```

numba/core/compiler.py
```diff
@@ -646,49 +652,35 @@ def compile_ir(typingctx, targetctx, func_ir, args, return_type, flags,
...
-        else:
-            cres = norw_cres
-        return cres
+            return rw_cres
+
+        # Suppress warnings in compilation retry
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", errors.NumbaWarning)
+            norw_cres = compile_local(func_ir.copy(), norw_flags)
+
```
stuartarchibald, line 679: I think this logic is now backwards compared to the original? The original compiled a version without rewrites and then subsequently compiled one with rewrites https://github.com/numba/numba/blob/master/numba/core/compiler.py#L664-L685 this was because the rewrites could mutate the IR irreparably (even if `deepcopy`'d).

numba/core/dispatcher.py
```diff
@@ -983,6 +983,12 @@ def compile(self, sig):
             flags = self.flags
             args, return_type = sigutils.normalize_signature(sig)
 
+            # for objmode we should have ffi_forced_objects as args
+            # also there is probably no benefit in rewrites passes, so turn them off
```
stuartarchibald, line 987:
```suggestion
            # also there are no applicable rewrite passes, so turn them off
```
there's no rewrite pass triggered in the `objmode` pipeline
https://github.com/numba/numba/blob/58cd2ee50950bed13b129a07c83a7606340064fc/numba/core/compiler.py#L560-L588

stuartarchibald (commented):
Thanks for the PR. A few minor things to resolve else looks good.

numba/core/compiler.py
kozlov-alexey: > If the columns dtype dict is a constant would partial application work so as to keep most of the pandas interface but avoid having the `build_map` in the bytecode? Also slightly puzzled as to how this is adding much compilation time...

@stuartarchibald Yes, exactly removing the `build_map` from the code resolves it. Answering your question, it probably adds to compile time because the IR size itself grows in quadratic fashion of the dict size, e.g. following times were observed for DF with 64 columns:

id | pipeline | pass_name | target_func | time
-- | -- | -- | -- | --
50 | nopython | nopython_type_inference | csv_reader_py$4 | 18.9416706
895 | object | object_mode_back_end | csv_reader_py$3 | 13.9494324
885 | object | object_mode_back_end | csv_reader_py$3 | 10.2176549
39 | object | object_mode_back_end | csv_reader_py$3 | 9.8354137
49 | object | object_mode_back_end | csv_reader_py$3 | 9.0511516

And tracing pylowering itself confirmed it makes most of `object_mode_back_end` pass times for csv_reader_py$3:
> 	Line 82: _py_lowering_stage: func=csv_reader_py$3 took: 9.83552408
> 	Line 104: _py_lowering_stage: func=csv_reader_py$3 took: 9.05098248
> 	Line 1814: _py_lowering_stage: func=csv_reader_py$3 took: 10.21686435
> 	Line 1836: _py_lowering_stage: func=csv_reader_py$3 took: 13.94962978

So if we look at llvm IR of the csv_reader_py$3 we could find there blocks similar to these:
```
### the following block checks if the variable was loaded and is not null
B12.endif.endif.endif.end...endif.8.endif:
  %".1071" = load i8*, i8** %"$40build_const_key_map.15"
  %".1072" = icmp eq i8* null, %".1071"                             
  br i1 %".1072", label %"B12.endif.endif.endif.end...endif.8.endif.if", label %"B12.endif.endif.endif.end...endif.8.endif.endif", !prof !0
  
### the block below is executed at failure (when in the above %".1071" is null)
B12.endif.endif.endif.end...endif.8.endif.if:       
  call void @"PyErr_SetString"(i8* @"PyExc_NameError", i8* bitcast ([48 x i8]* @".const.name '$40build_const_key_map.15' is not defined" to i8*))
  %".1075" = load i8*, i8** %"names"
  call void @"Py_DecRef"(i8* %".1075")
  %".1077" = load i8*, i8** %"$22build_list.6"
  call void @"Py_DecRef"(i8* %".1077")
  %".1079" = load i8*, i8** %"$30load_attr.10"
  call void @"Py_DecRef"(i8* %".1079")
  %".1081" = load i8*, i8** %"$40build_const_key_map.15"    ### since we raise exception, we call 
  call void @"Py_DecRef"(i8* %".1081")            ### PyLower::cleanup_vars() and decref all live variables
  %".1083" = load i8*, i8** %"skiprows"
  call void @"Py_DecRef"(i8* %".1083")
...
  %".1097" = load i8*, i8** %"$40build_const_key_map.18"    
  call void @"Py_DecRef"(i8* %".1097")
...
  %".1103" = load i8*, i8** %"$40build_const_key_map.16"
  call void @"Py_DecRef"(i8* %".1103")
  %".1105" = load i8*, i8** %"$26load_attr.8"
  call void @"Py_DecRef"(i8* %".1105")
  ret i32 -1
```

stuartarchibald: > I think this logic is now backwards compared to the original? The original compiled a version without rewrites and then subsequently compiled one with rewrites https://github.com/numba/numba/blob/master/numba/core/compiler.py#L664-L685 this was because the rewrites could mutate the IR irreparably (even if `deepcopy`'d).

@kozlov-alexey was wondering if you had a view on the above? Thanks.

juanjgalvez:
I'm wondering if this is going to be in 0.52 RC. I have observed similar issues with objmode. If compiling the objmode blocks takes a long time, this stretches the execution time of the function by quite a bit.

numba/core/compiler.py
kozlov-alexey: @stuartarchibald Yes, I think we should skip the changes in `compile_local` for now, since it requires immutable IR anyway.  So if it's OK with you we can stick to the PR name and just fix issue locally in objmode dispatcher. That will remove objmode recompilations keeping the rest of compile logic the same.

numba/core/dispatcher.py
```diff
@@ -983,6 +983,13 @@ def compile(self, sig):
             flags = self.flags
             args, return_type = sigutils.normalize_signature(sig)
 
+            # for objmode we should have ffi_forced_objects as args
+            # and no_rewrites flag can be set to avoid recompilation with and
+            # without rewrites in compile_ir
+            if isinstance(self, ObjModeLiftedWith):
+                args = [types.ffi_forced_object] * len(args)
+                flags.no_rewrites = True
+
```
stuartarchibald, line 992: Would it work to put this logic into an override of the `compile` method in the `ObjModeLiftedWith` subclass, and have this method then call `super().compile()`, this keeping subclass details out of the base class? The flags can be set in the initializer too perhaps. Was thinking of something like:
```diff
diff --git a/numba/core/dispatcher.py b/numba/core/dispatcher.py
index 9b5b66b61..659964077 100644
--- a/numba/core/dispatcher.py
+++ b/numba/core/dispatcher.py
@@ -1063,6 +1063,8 @@ class ObjModeLiftedWith(LiftedWith):
             raise ValueError("expecting `flags.force_pyobject`")
         if self.output_types is None:
             raise TypeError('`output_types` must be provided')
+        # switch off rewrites, they have no effect
+        self.flags.no_rewrites = True
 
     @property
     def _numba_type_(self):
@@ -1107,6 +1109,12 @@ class ObjModeLiftedWith(LiftedWith):
                 )
                 raise errors.TypingError(msg.format(i))
 
+    @global_compiler_lock
+    def compile(self, sig):
+        args, _ = sigutils.normalize_signature(sig)
+        sig = (types.ffi_forced_object,) * len(args)
+        return super().compile(sig)
+
 ```
might this work?

kozlov-alexey: @stuartarchibald Yes, it certainly works and it's also much better from code organization perspective. Thanks!

sklam (approved):
Looks good

PR #6213, opened 2020-09-03
Fix #6177. Remove AOT dependency on the Numba package

### LINKED ISSUE #6177: Importing AOT compiled function in 0.51.0 still requires numba installed
I've got a CI/CD system that uses the latest numba.
I use pyinstaller, and AOT all numba functions, then exclude numba from pyinstaller. This reduces build by around 50MB.
Unfortunately, in numba 0.51.0, importing an AOT function fails, citing that the AOT function requires importing numba.
This is a regression from 0.50.1, which behaved correctly.

## Reporting a bug

sklam:
Fix #6177 

- Re-add old `numba_unpickle()` implementation that does not depend on the `numba` package.
- Use `NUMBA_COMPILING_AOT` macro to control which implementation to use.
- existing AOT tests are modified to disable the `numba` package.

### DIFF
--- a/numba/_helperlib.c
+++ b/numba/_helperlib.c
@@ -978,7 +978,7 @@ numba_do_raise(PyObject *exc_packed)
     return status;
 }
 
-#ifdef NUMBA_COMPILING_AOT
+#ifdef PYCC_COMPILING
 /* AOT avoid the use of `numba.core.serialize` */
 NUMBA_EXPORT_FUNC(PyObject *)
 numba_unpickle(const char *data, int n, const char *hashed)
--- a/numba/_helperlib.c
+++ b/numba/_helperlib.c
@@ -978,6 +978,35 @@ numba_do_raise(PyObject *exc_packed)
     return status;
 }
 
+#ifdef NUMBA_COMPILING_AOT
+/* AOT avoid the use of `numba.core.serialize` */
+NUMBA_EXPORT_FUNC(PyObject *)
+numba_unpickle(const char *data, int n, const char *hashed)
+{
+    PyObject *buf, *obj;
+    static PyObject *loads;
+
+    /* Caching the pickle.loads function shaves a couple µs here. */
+    if (loads == NULL) {
+        PyObject *picklemod;
+        picklemod = PyImport_ImportModule("pickle");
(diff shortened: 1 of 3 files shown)


numba/_helperlib.c
```diff
@@ -978,6 +978,35 @@ numba_do_raise(PyObject *exc_packed)
...
+    if (buf == NULL)
+        return NULL;
+    obj = PyObject_CallFunctionObjArgs(loads, buf, NULL);
+    Py_DECREF(buf);
+    return obj;
+}
+
+#else
 
 NUMBA_EXPORT_FUNC(PyObject *)
```
stuartarchibald, line 1011: Does this imply that AOT compiled functions do not have the same level of e.g. closure compilation support as those which are jitted and cached?

sklam: The only difference here is the lack of in-memory cache for the unpickled reuslts. The `_numba_unpickle()` method also relies on the same `pickle.loads()`. 

If we want to do better for AOT unpickling performance, we can implement `_numba_unpickle()` in C.

numba/pycc/modulemixin.c
```diff
@@ -13,6 +13,8 @@
 #define NUMBA_EXPORT_FUNC(_rettype) VISIBILITY_HIDDEN _rettype
 #define NUMBA_EXPORT_DATA(_vartype) VISIBILITY_HIDDEN _vartype
 
+#define NUMBA_COMPILING_AOT
```
stuartarchibald, line 16: I think elsewhere the `PYCC` prefix is in use, would `PYCC_COMPILING` fit better?

stuartarchibald (commented):
Thanks for the patch. I've manually tested this by using https://github.com/alex-ong/numba_aot/blob/master/numba-compiler.py, creating an extension library DSO and copying it to /tmp and then trying to import it with system python. Current mainline has the issue described in #6177 (i.e. depends on numba [`AttributeError: module 'numba.core.serialize' has no attribute '_numba_unpickle'`]) with this patch the module imports fine. One minor thing to look at else good to go.

stuartarchibald (approved):

PR #6215, opened 2020-09-04
Attempt immutable ir

sklam:

### DIFF
--- a/numba/core/ir.py
+++ b/numba/core/ir.py
@@ -1,4 +1,4 @@
-from collections import defaultdict
+from collections import defaultdict, namedtuple
 import copy
 import itertools
 import os
@@ -230,6 +230,7 @@ class SlotEqualityCheckMixin(object):
 @total_ordering
 class EqualityCheckMixin(object):
     """ Mixin for basic equality checking """
+    __slots__ = ()
 
     def __eq__(self, other):
         if type(self) is type(other):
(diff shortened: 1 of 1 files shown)


numba/core/ir.py
```diff
@@ -214,6 +232,8 @@ class SlotEqualityCheckMixin(object):
     def __eq__(self, other):
         if type(self) is type(other):
             for name in self.__slots__:
+                if name in {'_loc', '_scope'}:
```
gmarkall, line 235: If I've understood correctly, we now treat IR nodes as equal even if they differ in location and scope - presumably this is OK for now, but is there a danger of this being a gotcha for the future? For example, could two nodes with different scopes that are otherwise equal have different effects, and therefore lead to some erroneous analysis?

gmarkall (commented):
Based on my limited understanding of the details of the IR, this looks OK to me. I do have a question about equality on the diff.

I found it quite hard to understand the changes around `_lower_array_expr`, but I think that's due to my lack of familiarity with that area - it seems like the transformation takes place in parallel on the AST function representation as well as the IR, and these should remain consistent with each other - is that correct? The issue with the old code being that `Var` names were mutated during the transformation?

As a side comment, It does feel a bit like type annotations would have helped me parse what was going on in `lower_array_expr` and `_legalize_parameter_names` a bit more easily.

PR #6222, opened 2020-09-07
[WIP] Add type annotations to parts of numba/core

EPronovost:


luk-f-a:
hi @EPronovost , I've been working on typing types/abstract too. The work is almost done but I got distracted and didn't open a PR. How can we merge the work? Should I add comments here?

EPronovost:
Haha @luk-f-a I was going to tag you in this PR once I made a bit more progress.  I'm happy to work with you on adding type annotations.  I feel like it will be good to stay more closely in sync than the current pace of PRs to `master` allow.  How would you want to approach it?  One possibility for how to iterate more quickly together is on a personal fork of `numba`.

luk-f-a:
yes, a fork might work better. I won't be able to do anything for a few days. I'll clone your branch and see if I can create a PR against it.

EPronovost:
@stuartarchibald I think this may be a good checkpoint to have as a PR.  I'll probably branch off of this for continued work while this PR is under review.  Could we update the label for the PR?

luk-f-a:
@EPronovost I cannot create a PR from my branch to yours. I will add comments on this PR if I think my version has something to contribute. And I will create a WIP PR into numba/numba so you can see all my changes. 
If anything is left over after this PR is merged, I will activate my PR to add any remaining changes.

numba/core/types/abstract.py
```diff
@@ -89,19 +91,22 @@ class Type(metaclass=_TypeMetaclass):
...
+    def key(self) -> pt.Any:
         """
         A property used for __eq__, __ne__ and __hash__.  Can be overridden
         in subclasses.
         """
         return self.name
 
     @property
-    def mangling_args(self):
+    def mangling_args(self) -> pt.Tuple[str, pt.Tuple[pt.Any, ...]]:
```
luk-f-a, line 109: the docstring says that a sequence is returned so I'd write it as `def mangling_args(self) -> pt.Tuple[str, pt.Sequence[pt.Any]]:`
In fact, there's one case where a list is returned (not a string).

numba/core/types/abstract.py
```diff
@@ -22,12 +22,14 @@ def _autoincr():
...
 
-_typecache: ptDict[weakref.ref, weakref.ref] = {}
+_typecache: pt.Dict[weakref.ref, weakref.ref] = {}
 
 def _on_type_disposal(wr, _pop=_typecache.pop):
     _pop(wr, None)
 
 
+_TypeClassType = pt.TypeVar("_TypeClassType")
+
```
luk-f-a, line 32: based on the discussion on August 25 on the Numba-dev gitter channel, it would be useful to distinguish between Numba type classes and instances. I would add in this section (or above)

```
NumbaTypeClass = pt.Type["Type"]
NumbaTypeInst = pt.Union["Type"]
```

numba/core/types/abstract.py
```diff
@@ -57,14 +59,14 @@ def _intern(cls, inst):
             _typecache[wr] = wr
             return inst
 
-    def __call__(cls, *args, **kwargs):
+    def __call__(cls, *args: pt.Any, **kwargs: pt.Any) -> "Type":
```
luk-f-a, line 62: if you accept the suggestion about `NumbaTypeInst` then it would apply here and everywhere else where it says Type.

numba/core/types/abstract.py
```diff
@@ -111,16 +116,16 @@ def mangling_args(self):
...
-    def __repr__(self):
+    def __repr__(self) -> str:
         return self.name
 
-    def __hash__(self):
+    def __hash__(self) -> int:
         return hash(self.key)
 
-    def __eq__(self, other):
+    def __eq__(self, other: pt.Any) -> bool:
```
luk-f-a, line 125: since the code attempts to get the `.key` property, only `NumbaTypeInst` are allowed. Would it be valid and more precise to use `NumbaTypeInst` instead of `pt.Any`?

numba/core/types/abstract.py
EPronovost: Do you mean "(not a tuple)"?  I updated the annotation based on that interpretation.

numba/core/types/abstract.py
EPronovost: I like that idea.  Done.  I also updated the section about type annotations in the contributing doc.

numba/core/types/abstract.py
EPronovost: Done.  I think I caught all uses of `Type`.

numba/core/types/abstract.py
EPronovost: Done.

EPronovost:
@stuartarchibald I think this is "Ready for Review".

numba/core/types/abstract.py
luk-f-a: sorry, I meant "not a tuple".

EPronovost:
Flake8 is now failing repeatedly from `numba/core/types/scalars.py:120:13: E123 closing bracket does not match indentation of opening bracket's line`

However, Line 120 from `types/scalars.py` is empty https://github.com/EPronovost/numba/blob/epronovost/type-annotations/numba/core/types/scalars.py#L120

This also didn't reproduce locally.  I'm not sure what's going on with the CI.

stuartarchibald:
> Flake8 is now failing repeatedly from `numba/core/types/scalars.py:120:13: E123 closing bracket does not match indentation of opening bracket's line`
> 
> However, Line 120 from `types/scalars.py` is empty https://github.com/EPronovost/numba/blob/epronovost/type-annotations/numba/core/types/scalars.py#L120
> 
> This also didn't reproduce locally. I'm not sure what's going on with the CI.

Probably worth checking `flake8` versions, seem to recall seeing this before.

stuartarchibald:
> @stuartarchibald I think this is "Ready for Review".

Great, thanks!

EPronovost:
@stuartarchibald can you clarify what you mean about `flake8`?  Locally I'm running flake8 3.8.3 which is the same as CI.  I'm not clear what the version would do for the fact that the error message seems to be coming from a line that doesn't exist.  It's entirely possible I messed something up when adding `typing_extensions`, and would appreciate someone reviewing those changes.

stuartarchibald:
> @stuartarchibald can you clarify what you mean about `flake8`? Locally I'm running flake8 3.8.3 which is the same as CI. I'm not clear what the version would do for the fact that the error message seems to be coming from a line that doesn't exist. It's entirely possible I messed something up when adding `typing_extensions`, and would appreciate someone reviewing those changes.

@EPronovost ah ok, this was my first guess as IIRC we've seen issues before with different flake8 versions producing different complaints. I've just checked this further, the problem is that because CI is doing a merge-strategy for testing (i.e. it merges this PR branch with mainline and then tests the merge) that the `flake8` violation is coming from the merge opposed to this code. I suggest syncing this PR with mainline via a merge locally to see the problem. A PR that got merged the other day adding `BooleanLiteral` support is causing the violation. Thanks!

EPronovost:
@stuartarchibald do you know when it would be possible for this PR to be reviewed?

mypy.ini
```diff
@@ -4,49 +4,59 @@
 warn_unused_configs = True
 follow_imports = silent
 show_error_context = True
-files = **/numba/core/types/*.py, **/numba/core/datamodel/*.py, **/numba/core/rewrites/*.py, **/numba/core/unsafe/*.py
+show_error_codes = True
+files = **/numba/core/types/*.py, **/numba/core/typeconv/*.py, **/numba/core/datamodel/*.py, **/numba/core/rewrites/*.py, **/numba/core/unsafe/*.py
+ignore_errors = True
```
sklam, line 9: Why is this needed? And, is it ignoring all errors?

EPronovost: Previously we used the global level `files` setting to control which files to check with `mypy`.  This required specifying files in two places (here and below in the respective level).  This approach also doesn't seem to work with `pre-commit`.  My guess is that it's because `pre-commit` specifies the files to run on when invoking `mypy`, which [supercedes top-level config file options](https://mypy.readthedocs.io/en/stable/config_file.html#config-file-format).

This PR changes how we filter files.  Now, we always have `mypy` run on all of `numba/*`.  Levels 1-3 specify `ignore_errors = False`, so we do not ignore all errors.  For any file that does not fit one of the more specific patterns, we fall back to Level 4, which does have `ignore_errors = True`.  This way a file doesn't need to be specified twice in `mypy.ini` to get type checking, and it also works with `pre-commit` or when running `mypy` manually on specific files or directories.

EPronovost:
@sklam will this PR be reviewed for the 0.52 release?

EPronovost:
@sklam to clarify from the meeting, would you like me to explore adding a test time typeguard in this PR?  I was thinking of keeping PRs more modular and addressing it in a subsequent PR.  And I'll fix the merge conflicts later this week.

docs/source/developer/contributing.rst
```diff
@@ -270,6 +270,9 @@ There is potential for confusion between the Numba module ``typing`` and Python
 hints, as well as between Numba types---such as ``Dict`` or ``Literal``---and ``typing`` types of the same name.
 To mitigate the risk of confusion we use a naming convention by which objects of the built-in ``typing`` module are
 imported with an ``pt`` prefix. For example, ``typing.Dict`` is imported as ``from typing import Dict as ptDict``.
+Similarly, the Python built-in ``typing`` module is imported as ``import typing as pt``.
```
luk-f-a, line 273: I think we should move everything to the `pt.` naming, instead of of the `pt` prefix I used. what do you think?

luk-f-a:
@EPronovost, there seems one `.pyi` file in the PR, what should be the rule here? Wouldn't it be better to write all annotations on the main file? Or is that case special in some way?

EPronovost:
@luk-f-a the `.pyi` file is for [a C++ file](https://github.com/numba/numba/blob/master/numba/core/typeconv/_typeconv.cpp).  I agree we should prefer inline annotations instead of stubs for python files.  We can use stubs to define the Python APIs of C++ code.

docs/source/developer/contributing.rst
EPronovost: Agreed

luk-f-a:
you're right, I didn't realise that the pyi file was typing a c++ file.

EPronovost:
It's failing from typeguard because of the forward reference resolution.  I created an issue about it here https://github.com/agronholm/typeguard/issues/163
Worst case we can just replace forward references with `Any`, but I want to see what Alex says first.

srilman:
Is there any update to this PR? I'd like to add some type annotations to some other modules in `numba.core` and was wondering if I could build on top of this PR.

esc:
@srilman as far as I can tell this has been abandoned.

srilman:
Sounds good. I'll open a separate PR then

esc:
> Sounds good. I'll open a separate PR then

OK, but do be warned that any PR that modifies the Numba code-base at large is very unlikely to be reviewed. This has been attempted multiple times in the past for both code linting and such and has never ended too well for both the maintainers and the contributors. Your best bet is to make as small as possible incremental changes. As a guideline, perhaps only modify at most one file and only make a handful of changes in that file for a given PR. You can make as many PRs as you like of course over time, but try to not have more than one PR in flight at any given time.

So, select well which parts you would like to have annotated and take it one step at a time, that'll give you the highest chance of finding a reviewer and actually getting your work merged. Hope that helps.

srilman:
Yes, thanks @esc!

esc:
This PR has been abandoned. With the amount of conflicts etc.. this is not worth reviving.

PR #6225, opened 2020-09-08
adding stale bot config file

esc:
As documented in https://github.com/apps/stale.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,5 +1,5 @@
 # Number of days of inactivity before an issue becomes stale
-daysUntilStale: 30
+daysUntilStale: 14
 # Number of days of inactivity before a stale issue is closed
 daysUntilClose: 7
 # Issues with these labels will never be considered stale
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -18,6 +18,8 @@ exemptLabels:
   - feature_request
   - enhancement
   - task
+# Limit to only `issues` or `pulls`
+only: issues
 # Label to use when marking an issue as stale
 staleLabel: stale
 # Comment to post when marking an issue as stale. Set to `false` to disable
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,24 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 30
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - bug
+  - bug - failure to compile
+  - bug - ice
+  - bug - incorrect behaviour
+  - bug - memory leak
+  - bug - miscompile
+  - bug - regression
+  - bug - segfault
+  - feature_request
+# Label to use when marking an issue as stale
+staleLabel: wontfix
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed in 7 days if no further activity occurs.
+# Comment to post when closing a stale issue. Set to `false` to disable
+closeComment: >
+  This issue has been automatically closed as it was stale
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -12,7 +12,12 @@ exemptLabels:
   - bug - miscompile
   - bug - regression
   - bug - segfault
+  - performance - compile time
+  - performance - run time
+  - performance
   - feature_request
+  - enhancement
+  - task
 # Label to use when marking an issue as stale
 staleLabel: stale
 # Comment to post when marking an issue as stale. Set to `false` to disable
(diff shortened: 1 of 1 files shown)


esc:
The exclusion labels and maybe the 'stale' label need a review. 'wontfix' does exist as a label in the Numba repository, but I think it has gone out of use somewhat.

.github/stale.yml
```diff
@@ -0,0 +1,24 @@
...
+  - bug - failure to compile
+  - bug - ice
+  - bug - incorrect behaviour
+  - bug - memory leak
+  - bug - miscompile
+  - bug - regression
+  - bug - segfault
+  - feature_request
+# Label to use when marking an issue as stale
+staleLabel: wontfix
```
stuartarchibald, line 17: Agree with comments in the PR proper, think a `stale` label might be more appropriate? It's not that we `won't fix`, more that just something is probably lacking input/information to permit a classification of the problem?

.github/stale.yml
```diff
@@ -0,0 +1,24 @@
...
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed in 7 days if no further activity occurs.
+# Comment to post when closing a stale issue. Set to `false` to disable
+closeComment: >
+  This issue has been automatically closed as it was stale
```
stuartarchibald, line 24:
```suggestion
  This issue has been automatically closed as it was stale.
```

.github/stale.yml
```diff
@@ -0,0 +1,24 @@
...
+exemptLabels:
+  - bug
+  - bug - failure to compile
+  - bug - ice
+  - bug - incorrect behaviour
+  - bug - memory leak
+  - bug - miscompile
+  - bug - regression
+  - bug - segfault
+  - feature_request
```
stuartarchibald, line 15: should probably add the `performance*`, `enhancement` and `task` labels too? I think the `stale`ness will most likely apply to issue which are not successful in triage due to lack of information.

Any thoughts on the `question` label? Am wondering if this bot could close questions after e.g. a week of inactivity and also mention that `discourse` as the place to keep the Q&A corpus?

.github/stale.yml
```diff
@@ -0,0 +1,24 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 30
```
stuartarchibald, line 2: Perhaps make this 2 weeks and see how that goes? I've not often seen issues that are not yet triaged and also being resolved after a couple of weeks, perhaps the bot commenting earlier about the staleness might prompt a response on e.g. missing information?

stuartarchibald (commented):
Thanks for adding this patch to help with the issue tracker. Suggestions/questions are inline.

.github/stale.yml
esc: right, so let us make a new label `stale` and then use that. I am not sure this label will see much use anyway?

stuartarchibald: Done! https://github.com/numba/numba/labels/stale

esc: Oh, we have a `stale` label?  Does anyone know if it has been there all along or did someone make it recently?

esc: It looks like the `stale` label is not used in any issues, so let's use that.

stuartarchibald: Made it for the purpose of this PR!

.github/stale.yml
esc: I'll add the labels yes.

But I don't think this bot supports these custom kinds of messages yet.

esc:
@stuartarchibald I have addressed the comments, maybe @sklam or @seibert would like to take another look.

.github/stale.yml
```diff
@@ -0,0 +1,29 @@
...
+  - bug - memory leak
+  - bug - miscompile
+  - bug - regression
+  - bug - segfault
+  - performance - compile time
+  - performance - run time
+  - performance
+  - feature_request
+  - enhancement
+  - task
```
stuartarchibald, line 20: Am wondering also if this needs to be focused on issues only, else PRs that take a while get closed? e.g. https://github.com/probot/stale/blame/master/README.md#L68-L78 ?

esc: yes, issues only,I'll go  and  fix it up.

esc: added and pushed

sklam:
I'm wondering if we should use `onlyLabels` or we need to do an label cleanup before enabling the bot. There are labels like "Numba1.0" that should not be stale.

stuartarchibald:
@esc Should we close this for now?

esc:
@stuartarchibald we might as well. I think StaleBot was a nice try, we learnt that the GitHub search and filter bar has bugs, we also learnt that Github never got back to us about about the filter and search being buggy. The positive outcome is that we manually pruned, cleaned and labeled our backlog once, which is good.

PR #6231, opened 2020-09-09
Remove compatibility mode

gmarkall:
This PR removes compatibility mode (referred to as interpreter mode in some places) - it appears to me that it has served its purpose, and removing it simplifies a few things in the dispatcher and other places. Some thoughts on the removal:

- The docs advised using it only for migrating code from Numba 0.11 and prior - Numba 0.11 was released in September 2013.
- Anyone who still needs the flag to migrate to a newer Numba could use Numba 0.51.2 with the flag to migrate, then once their migration is complete, upgrade to Numba 0.52 or beyond.
- There were some tests that tested compatibility mode, which test things that I think are unsupported in object or nopython mode - the `del` operator, list comprehensions, and returning closures and generator expressions from jitted functions. I have accordingly removed these tests.

Are there reasons to prefer keeping compatibility mode that I missed?


stuartarchibald (approved):
Thanks for the patch, the removal diff looks clean.

Issue #4314, opened 2019-07-12
@guvectorized functions over dask distributed


crusaderky:
### UPDATE 2024
Statically defined gufuncs don't work with distributed anymore, either. Read: https://github.com/numba/numba/issues/4314#issuecomment-1992103833

### Original post

Dynamically defined functions - namely, anything from a Jupyter notebook - can be executed over a dask distributed grid thanks to cloudpickle.
Functions decorated with @guvectorize can be pickled, but only as long as they are defined in a static module.

Type in a jupyter notebook:

```python
import distributed
from numba import guvectorize, i8

def f(x):
    return x * 2


@guvectorize([(i8, i8[:])], '()->()')
def g(x, out):
    out[0] = x * 2

    
client = distributed.Client()

print(f(2))
print(g(2))
print(client.submit(f, 2).result())
print(client.submit(g, 2).result())
```
Output:
```
4
4
4

AttributeError: module '__main__' has no attribute 'g'
```

Note that the above is not the same as just writing ``cloudpickle.loads(cloudpickle.dumps(g))``, which works (but isn't terribly useful) because loads and dumps are running in the same interpreter.

Workaround:
```python
def cp_guvectorize(*args, **kwargs):
    """Same as :func:`numba.guvectorize`, but can be used to decorate dynamically
    defined function and then pickle them with
    `cloudpickle <https://pypi.org/project/cloudpickle/>`_.
    On the other hand, it can't be called from another jit-compiled function.
    """

    def decorator(func):
        return _PickleableGUVectorized(func, args, kwargs)

    return decorator


class _PickleableGUVectorized:
    def __init__(self, func, guvectorize_args, guvectorize_kwargs):
        self.args = func, guvectorize_args, guvectorize_kwargs
        decorator = numba.guvectorize(*guvectorize_args, **guvectorize_kwargs)
        self.ufunc = decorator(func)

    def __reduce__(self):
        return _PickleableGUVectorized, self.args

    def __call__(self, *args, **kwargs):
        return self.ufunc(*args, **kwargs)

    
@cp_guvectorize([(i8, i8[:])], '()->()')
def h(x, out):
    out[0] = x * 2

print(client.submit(h, 2).result())
```
Output:
```
4
```

mrocklin:
I would also love to see this.  This might also be fixed by adding `__setstate__` and `__getstate__` methods.

mrocklin:
[...]

Issue #6182, opened 2020-08-28
NumPy 1.20 support


stuartarchibald:
## Feature request

Add support for NumPy 1.20.

Current known issues:

* [x] https://github.com/numba/numba/issues/6175
* [x] https://github.com/numba/numba/pull/6176
* [x] https://github.com/numba/numba/issues/6965
* [x] https://github.com/numba/numba/issues/6812
* [x] https://github.com/numba/numba/issues/7065 (CUDA)
* [ ] https://github.com/numba/numba/issues/6864

gmarkall:
I'm going to close this issue as we support NumPy 1.20 and 1.21 now, but note that #6864 is still unresolved - I have added it to the 0.56RC milestone so we don't lose it in the near future.

PR #6234, opened 2020-09-11
[WIP] Ensure DUfunc.ufunc objects are picklable

TomAugspurger:
This PR ensures that `DUFunc.ufunc` objects (which are instances of NumPy ufuncs) are picklable. It relies on @seberg's PR to NumPy at https://github.com/numpy/numpy/pull/17289. It fixes https://github.com/dask/distributed/issues/3450.

I've marked the PR as WIP, until we have a NumPy version we can compare against to skip the test.


stuartarchibald:
Thanks @TomAugspurger @seberg, such changes to increase ease of interoperability/consumption of are Numba gratefully received! The code around ufunc generation is some of the oldest in Numba and is rarely updated, is there anything else that the ufunc generation can do to help with this issue? Thanks again!

crusaderky:
To clarify, this does not close #4314 right? e.g. gufuncs defined in jupyter notebook are still unpickleable

stuartarchibald:
> To clarify, this does not close #4314 right? e.g. gufuncs defined in jupyter notebook are still unpickleable

Correct. But I think the fix for #4314 is likey along a similar path.

seberg:
This is a try to go in the right direction, but it certainly is a bit minimal. As I said in the numpy PR/issue, we could allow setting `__qualname__` and `__module__` to make pickling a bit saner, but maybe that is not even enough for you.
The other possibility might either hooking into a new `__recuce__` directly or subclassing.
If you need things like that, I think we can make that happen, I am just not sure what the right approach would be here.

TomAugspurger:
I think deserializing the numba-generated ufunc defined in `__main__` will necessarily involve re-compiling it in the other processes. The cleanest way of doing that is likely some kind of `__reduce__` that's settable at creation time, or through subclassing `np.ufunc`.

stuartarchibald:
> I think deserializing the numba-generated ufunc defined in `__main__` will necessarily involve re-compiling it in the other processes. The cleanest way of doing that is likely some kind of `__reduce__` that's settable at creation time, or through subclassing `np.ufunc`.

Given this is IPC I think the hardware target would be the same so the pickled numba-generated ufunc can just contain the ELF binary as bytes to avoid recompilation? Numba's on-disk caching works along these lines.

mrocklin:
I'm very excited about the work here.  Thanks @TomAugspurger @seberg and others

stuartarchibald:
> This is a try to go in the right direction, but it certainly is a bit minimal. As I said in the numpy PR/issue, we could allow setting `__qualname__` and `__module__` to make pickling a bit saner, but maybe that is not even enough for you.
> The other possibility might either hooking into a new `__recuce__` directly or subclassing.
> If you need things like that, I think we can make that happen, I am just not sure what the right approach would be here.

From a Numba perspective, sticking extra things into generated code etc is likely just a question of wrangling the right bit of codegen. `__qualname__` and `__module__` can probably be added relatively easily, the issue with these sort of things has always been what to do when there's nothing reasonable to inject or there's an aliasing problem.

@sklam, having just fixed up a load of pickling for Numba JIT functions, any thoughts on the above, can `numba.core.serialize` help here perhaps?

sklam:
Customizing the `__reduce__` will be the most flexible. To do that, we will need to shove in all the states needed to re-compile a new ufunc into the ufunc instance.

seberg:
So, is it correct to say that Numba has no desire to subclass ufuncs?  We could add a C-API call for it, but I am not even sure it matters. Using a "private" property or python function is probably just as well, the only downside seems that it is easier to misuse.
At some point (hopefully soon), we have to create a new public API for creating ufuncs, (using a `PyType_FromSpec`-like API), which would allow adding this more easily.
But until then, hashing this out now – even if it is temporary and will be deprecated – is the easier path and necessary in any case.

So my question would be: Lets say we create a ufunc property/method to allow setting a hook for `__reduce__`, how should that look like? I tried to see if there is prior art, but the only thing I could think of for now are python functions, and they don't have any functionality for this?

TomAugspurger:
>  But until then, hashing this out now – even if it is temporary and will be deprecated – is the easier path and necessary in any case.

As a user of both libraries, I'd be happy to see a temporary solution to this while we work out something more robust :)

numba/np/ufunc/_internal.c
```diff
@@ -210,7 +210,7 @@ dufunc_init(PyDUFuncObject *self, PyObject *args, PyObject *kws)
     /* Construct the UFunc. */
     tmp = PyObject_GetAttrString(py_func_obj, "__name__");
     if (tmp) {
-        name = PyString_AsString(tmp);
+        name = PyString_AsString(PyString_FromFormat("%s.ufunc", PyString_AsString(tmp)));
```
seberg, line 213: We put in the change in numpy master.  The tricky part with the quick fix here is the memory management (assuming a dufunc can be deleted).  The string object currently created can never be dereferenced.

There is definitely much to be cleaned up, and it has to start in NumPy :/.  NumPy solves this by allocating the name as part of the single `ufunc->ptr`.  I expect numba already does enough dark magic that it could do that as well, but it may be tricky.  The other thing that looks a bit like it is a work-around is the `keepalive` list here?

TomAugspurger: >  The string object currently created can never be dereferenced.

Hmm, are you saying that my PR makes a string that is never dereferenced? Or it was currently like that on numba? Or that the change to NumPy means that the string never gets dereferenced?

seberg: Sorry, the first: this PR creates a string that is never dereferenced, even if the ufunc is deleted. But, the code is tricky, so I am not sure that it was even correct before.  And solving it seems not super straight forward.

TomAugspurger:
Sorry for the delay here. I updated the tests to skip for NumPy<1.2.0, which I think is the first version with the new ufunc pickling in place. Hopefully CI passes now.

TomAugspurger:
Might want to wait for NumPy 1.20 (https://github.com/numba/numba/issues/6182) before merging, so that this is actually tested :) I haven't tried it locally today.

mrocklin:
> Might want to wait for NumPy 1.20 (#6182) before merging, so that this is actually tested :) I haven't tried it locally today.

Do we have a sense for when this might occur?  (Does Numpy release on a predictable schedule?  Every six months if I recall correctly?)

If it's going to be a while then Dask might also consider looking for Numba-generated gufuncs and special-casing them.  @crusaderky had a workaround in the referenced issue.

jrbourbeau:
For reference, it appears the known issues in Numba for adding NumPy 1.20 support (at least the ones mentioned in https://github.com/numba/numba/issues/6182) seem to have been resolved -- though there may be others not linked to in that issue

stuartarchibald:
> For reference, it appears the known issues in Numba for adding NumPy 1.20 support (at least the ones mentioned in #6182) seem to have been resolved -- though there may be others not linked to in that issue

All issues except https://github.com/numba/numba/issues/6864 in the NumPy 1.20 support ticket (https://github.com/numba/numba/issues/6182) have been resolved. The upcoming 0.54 release of Numba will support NumPy 1.20. 

Numba public CI is also now running with NumPy 1.20, @TomAugspurger please could you resolve merge conflicts and that'll trigger a build. Many thanks.

PR #6241, opened 2020-09-13
 [Don't merge] Typing types/abstract.py

luk-f-a:
as described, don't merge. PR used for collaboration.

### DIFF
--- a/numba/core/types/abstract.py
+++ b/numba/core/types/abstract.py
@@ -1,5 +1,6 @@
 from abc import ABCMeta, abstractmethod, abstractproperty
-from typing import Dict as ptDict, Type as ptType
+import typing as pt
+
 import itertools
 import weakref
 
@@ -26,7 +27,9 @@ def _autoincr():
     assert n < 2 ** 32, "Limited to 4 billion types"
     return n
 
-_typecache: ptDict[weakref.ref, weakref.ref] = {}
+
+_typecache: pt.Dict[weakref.ref, weakref.ref] = {}
+
 
 def _on_type_disposal(wr, _pop=_typecache.pop):
     _pop(wr, None)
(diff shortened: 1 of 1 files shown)


numba/core/types/abstract.py
```diff
@@ -170,22 +191,34 @@ def augment(self, other):
...
+
+    def __call__(self, *args: pt.Any) -> pt.Any:
         from numba.core.typing import signature
         if len(args) == 1 and not isinstance(args[0], Type):
             return self.cast_python_value(args[0])
         return signature(self, # return_type
                          *args)
 
-    def __getitem__(self, args):
+    def __getitem__(self, args: pt.Union[slice,
```
EPronovost, line 207: Is this the type of formatting we'd like to use? I'm personally a fan of black style formatting.

numba/core/types/abstract.py
```diff
@@ -1,12 +1,23 @@
...
 
 import numpy as np
-
+from numba.core.typeconv import Conversion
+if pt.TYPE_CHECKING:
+    # these imports are only needed for python typing and would otherwise
+    # cause circular imports
+    from numba.core.typing.templates import Signature
+    from numba.core.types import Array
+    from numba.core.base import BaseContext
```
EPronovost, line 14: Just to confirm, currently `Signature` and `BaseContext` are un-annotated (i.e. aliases for `pt.Any`)?  I had started annotating `Signature` on another branch.

luk-f-a: are they really aliases for `pt.Any` if they are not themselves annotated? 
In this example, mypy complains about `bar` but not about `baz`. Even if `Foo` is not annotated mypy correctly checks inputs annotated as `Foo`

```
from typing import Any
class Foo:
    def __init__(self, a):
        self.a = a

def bar(x: Foo) -> Foo:
    return x

def baz(x: Any) -> Any:
    return x

foo = Foo(1)
bar(1)
baz(1)
```

EPronovost: I think I was incorrect for this case.  I have experience working with larger packages with more complex ways of defining public endpoints (e.g. Tensorflow).  I think for this case we do get some checking.  From the [mypy docs](https://mypy.readthedocs.io/en/stable/running_mypy.html):

> Even if mypy is unable to perfectly type check a file, it can still glean some useful information by parsing it (for example, understanding what methods a given object has).

Sorry for the confusion.

Issue #6184, opened 2020-08-28
Improve error message for NumPy alias type used as dtype in ArrayNdCtors


stuartarchibald:
## Reporting a bug

- [x] I have tried using the latest released version of Numba (most recent is
 visible in the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [x] I have included below a minimal working reproducer (if you are unsure how
 to write one see http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports).

As per #3993 This
```python
from numba import njit
import numpy as np

@njit
def foo():
    return np.zeros((2, 2), dtype=np.float)

foo()
```
produces:
```
numba.core.errors.TypingError: Failed in nopython mode pipeline (step: nopython frontend)
No implementation of function Function(<built-in function zeros>) found for signature:
 
 >>> zeros(UniTuple(Literal[int](2) x 2), dtype=Function(<class 'float'>))
 
There are 2 candidate implementations:
  - Of which 2 did not match due to:
  Overload of function 'zeros': File: numba/core/typing/npydecl.py: Line 504.
    With argument(s): '(UniTuple(int64 x 2), dtype=Function(<class 'float'>))':
   No match.

During: resolving callee type: Function(<built-in function zeros>)
During: typing of call at issue3993.py (25)


File "issue3993.py", line 25:
def foo():
    return np.zeros((2, 2), dtype=np.float)
    ^


```
which provides correct information given what Numba can infer generically, but this is a common issue and so it'd probably be a good idea to provide a specific message.

Here's starter patch to provide a better error message:
```diff
diff --git a/numba/core/typing/npydecl.py b/numba/core/typing/npydecl.py
index 2dbbed39b..3483fced2 100644
--- a/numba/core/typing/npydecl.py
+++ b/numba/core/typing/npydecl.py
@@ -508,6 +508,25 @@ class NdConstructor(CallableTemplate):
             else:
                 nb_dtype = parse_dtype(dtype)
 
+            if nb_dtype is None:
+                if isinstance(dtype, types.Function):
+                    # user is probably doing `np.float` or `np.int` or similar.
+                    try:
+                        nm = dtype.key[0].__name__
+                        if hasattr(np, nm):
+                            # user probably ought to use a specific type or the
+                            # one with an underscore
+                            if hasattr(np, '%s_' % nm):
+                                msg = ("dtype 'np.%s' is not supported, "
+                                       "perhaps use 'np.%s_' or a more "
[...]

Issue #3993, opened 2019-04-18
[Not a Bug, but a hazard] use of lists for shape args in empty(), zero(), Ones()


cems2:
Noting an inconsistency between numba and numpy  in the type of arguments allowed for empty(), zero() and ones() and probably others.

This is not a bug per se, just a case were generally expected behaviors don't match between numba and numpy.

In numpy the documentation states that the shape argument of numpy.empty is a tuple of dimensions.

**However although not documented, numpy will also take a list in place of a tuple.

Numba Will only take a tuple.  It will not take a list.**

_This is both good a bad._ 

 It's bad because it breaks a lot of existing code that, albeit sloppy, depends on a list being accepted in numpy.

I think is it probably good because Cuda types will work better if Tuples are used I believe.  

The real problem is the error output is a bit cryptic and the result unexpected by experienced numpy users.

**CODE TO REPRODUCE**
```
def testempty_List(i):
    moo = np.empty([i,i],dtype=np.float64)
    return moo

def testempty_Tuple(i):
    moo = np.empty((i,i),dtype=np.float64)
    return moo

# pure numpy
testempty_Tuple(3)   
testempty_List(3)                # this will not give an error
        
nb.njit(testempty_Tuple)(3)
nb.njit(testempty_List)(3)     # this will give an error
```

stuartarchibald:
Thanks for the report. The issue here is that Numba has to determine the types of everything to be able to compile the code. Numba "knows" that a type is a list, but has no idea how many entries are in the list at compile time, and, as a result, if used in an array constructor, it has no idea what shape the array should be. I'd recommend reading the discussion here: https://github.com/numba/numba/issues/2771 which discusses implementing the `tuple()` constructor which suffers from the same problem. Any fix for #2771 is likely to also be applicable to the `ndarray` allocation routines mentioned.

stuartarchibald:
As to:
> The real problem is the error output is a bit cryptic and the result unexpected by experienced numpy users.

I get this:

```
numba.errors.TypingError: Failed in nopython mode pipeline (step: nopython frontend)
Invalid use of Function(<built-in function empty>) with argument(s) of type(s): (list(int64))
 * parameterized
In definition 0:
    All templates rejected with literals.
In definition 1:
    All templates rejected without literals.
This error is usually caused by passing an argument of a type that is unsupported by the named function.
[1] During: resolving callee type: Function(<built-in function empty>)
[2] During: typing of call at issue3993.py (6)


[...]

PR #6243, opened 2020-09-13
Improve error message for NumPy alias type used as dtype in ArrayNdCtors

Luiz6ustav0:
##### I had some minor issues in the beginning so ignore the first 6 commits, please.
This is a PR to implement what was suggested on #6184 

Changes basically consist of checks to provide a more specific message to this common issue - #3993 - users face.

- Do I need to add some unit tests for this change or are the current ones enough?

Thanks in advance.


stuartarchibald:
> ##### I had some minor issues in the beginning so ignore the first 6 commits, please.

No problem, feel free to just squash the history to a single patch and force push. No review has taken place so it's fine to do that up until review time.

> This is a PR to implement what was suggested on #6184
> 
> Changes basically consist of checks to provide a more specific message to this common issue - #3993 - users face.
> 
>     * Do I need to add some unit tests for this change or are the current ones enough?

Great, thanks for making the PR. As to testing, yes please, new logic definitely needs testing. Would suggest adding it to `numba.tests.test_dyn_array` and following this pattern to ensure that the raised exceptions are those expected (message strings match!):

https://github.com/numba/numba/blob/c448052c875d7a84e8d067f8483a1a1caa1dfe9b/numba/tests/test_dyn_array.py#L545-L547

Thanks!

numba/tests/test_dyn_array.py
```diff
@@ -509,6 +509,20 @@ def pyfunc(x):
...
 
+    def test_check_zeros_prototype(self):
+        def func():
+            return np.zeros((2, 2), dtype=np.float)
+        cfunc = nrtjit(func)
+        expected = np.zeros((2, 2), dtype=np.float)
+        result = cfunc()
+        np.testing.assert_equal(result, expected)
+        # errors
+        with self.assertRaises(TypingError) as cm:
```
Luiz6ustav0, line 520: I'm getting another error raised from a different place here

numba/core/errors.py
```diff
@@ -544,7 +544,8 @@ class NotDefinedError(IRError):
 
     def __init__(self, name, loc=None):
         self.name = name
-        msg = "Variable '%s' is not defined." % name
+        msg = ("The compiler failed to analyze the bytecode. "
+               "Variable '%s' is not defined." % name)
```
stuartarchibald, line 548: Is this the result of a merge that's gone awry? Fairly sure I saw a patch with this in recently.

numba/tests/test_dyn_array.py
stuartarchibald: > I'm getting another error raised from a different place here.

The Numba compiler catches all errors arising from compilation and rewrites them to include contextual information (it prints the source code, points at the line with the problem, marks up the output, trims the backtraces etc). I've just pulled down this patch and triggered the error, what is present is correct.

> I'm also having some trouble with finding the correct place to put the tests,

Putting this test here seems good as it'll get inherited by the tests for `zeros`, `ones` and `empty`.

> I'd really appreciate some help here

Hopefully the above gets you started.

> sorry for some of the weird commits again because of trouble pushing/pulling to/from remote.

No problem, we can fix up the patches if necessary.

numba/tests/test_dyn_array.py
```diff
@@ -509,6 +509,20 @@ def pyfunc(x):
 
 class ConstructorBaseTest(NrtRefCtTest):
 
+    def test_check_zeros_prototype(self):
```
stuartarchibald, line 512:
```suggestion
    def test_invalid_dtype_spelling_raises(self):
```

numba/tests/test_dyn_array.py
```diff
@@ -509,6 +509,20 @@ def pyfunc(x):
...
+        def func():
+            return np.zeros((2, 2), dtype=np.float)
+        cfunc = nrtjit(func)
+        expected = np.zeros((2, 2), dtype=np.float)
+        result = cfunc()
+        np.testing.assert_equal(result, expected)
+        # errors
+        with self.assertRaises(TypingError) as cm:
+            cfunc(-1)
+        self.assertEqual(str(cm.exception), "TypingError: dtype 'np.float' is "
```
stuartarchibald, line 522:
```suggestion
        self.assertEqual(str(cm.exception), "dtype 'np.float' is "
```
suggest just checking the message contents, the type is asserted in the `self.assertRaises(TypingError)`.

numba/core/typing/npydecl.py
```diff
@@ -508,6 +508,21 @@ def typer(shape, dtype=None):
...
+                if isinstance(dtype, types.Function):
+                    try:
+                        nm = dtype.key[0].__name__
+                        if hasattr (np, nm):
+                            if hasattr(np, '%s_' %nm):
+                                msg = ("dtype 'np.%s' is not supported. "
+                                       "Try using 'np.%s_' or a more "
+                                       "specific dtype." % (nm, nm))
+                            else:
+                                msg = "dtype 'np.%s' is not supported" % nm
```
stuartarchibald, line 521: Need to figure out a way of triggering this branch.

numba/core/typing/npydecl.py
```diff
@@ -508,6 +508,21 @@ def typer(shape, dtype=None):
...
+                            raise TypingError(msg)                            
+                    except AttributeError:
+                        pass
+                raise TypingError("dtype '%s' is unsupported." % dtype)
```
stuartarchibald, line 525: Need to figure out a way of triggering this branch.

stuartarchibald:
@Luiz6ustav0 thanks for your work on this so far, I've answered your questions and left a few comments inline. It'd also be good to fix the git history if possible please... given the size of the patch, whilst against recommended behaviour, I think a history rewrite would be permissible in this instance if that is most convenient. Many thanks.

numba/core/errors.py
Luiz6ustav0: It is indeed... I wrote that patch you mentioned and I ended up merging it on my local master too and this happened

Luiz6ustav0:
@stuartarchibald ok, I took a look at your review and that's what I was needing! I'll work on your suggested fixes/improvements in the near future and hopefully get this ready for review and merge again.

PR #6244, opened 2020-09-15
CUDA Docs: Make workflow using simulator more explicit

### LINKED ISSUE #2485: Code passing CUDA simulation mode does not work
The problem is if one uses multi dimensional local or shared arrays on cuda devices and writes array expressions.
1D array expressions seem to be working but while the following code works if
NUMBA_ENABLE_CUDASIM is set it won't actually work CUDA mode.

```
import numpy as np

import numba
from numba import cuda

@cuda.jit
def generate_kernel(mem_array):
    i = cuda.grid(1)
    buffer = cuda.local.array((1, 4), numba.float64)

    if i < mem_array.shape[0]:
        buffer[0] = mem_array[0]
        #for k in range(mem_array.shape[1]):
        #    buffer[k] = mem_array[0, k]


n_init_conds = 32 * 4
mem_array = np.zeros((n_init_conds, 4))
mem_array[0] = np.array([0.1, 0.2, 0.3, 0.4])

threadsperblock = 32
blockspergrid = (n_init_conds + (threadsperblock - 1)) // threadsperblock
generate_kernel[threadsperblock, blockspergrid](mem_array)
```

I could solve it be explicitly indexing everything, like in the commented code above, but simulation mode should complain about it.

gmarkall:
Fixes #2485.

### DIFF
--- a/docs/source/cuda/simulator.rst
+++ b/docs/source/cuda/simulator.rst
@@ -11,6 +11,11 @@ be used to debug CUDA Python code, either by adding print statements to your
 code, or by using the debugger to step through the execution of an individual
 thread.
 
+The simulator to deliberately allows non-CUDA code like starting a debugger and
+printing arbitrary expressions to run for debugging purposes. Therefore, it is
+best to start from code that compiles for the CUDA target, and then move over to
+the simulator to investigate issues.
+
 Execution of kernels is performed by the simulator one block at a time. One
 thread is spawned for each thread in the block, and scheduling of the execution
 of these threads is left up to the operating system.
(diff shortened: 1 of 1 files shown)


gmarkall:
The fail is a conda HTTP timeout - could someone re-run tests please?

docs/source/cuda/simulator.rst
```diff
@@ -11,6 +11,11 @@ be used to debug CUDA Python code, either by adding print statements to your
 code, or by using the debugger to step through the execution of an individual
 thread.
 
+The simulator to deliberately allows non-CUDA code like starting a debugger and
+printing arbitrary expressions to run for debugging purposes. Therefore, it is
```
stuartarchibald, line 15:
```suggestion
The simulator deliberately allows running non-CUDA code like starting a debugger 
and printing arbitrary expressions for debugging purposes. Therefore, it is
```

stuartarchibald (commented):
Thanks for the PR. Minor wording change suggested then good to go.

stuartarchibald (approved):
Thanks for the fixes.

PR #6248, opened 2020-09-16
Add support for CUDA atomic subtract operations

testhound:
This pull request adds support for CUDA atomic subtract.

1. Adds support for atomic subtract
2. Adds units tests for integer, float32 and float64
3. Adds documentation for atomic subtract.

### DIFF
--- a/docs/source/cuda-reference/kernel.rst
+++ b/docs/source/cuda-reference/kernel.rst
@@ -177,10 +177,10 @@ Synchronization and Atomic Operations
 
 .. function:: numba.cuda.atomic.sub(array, idx, value)
 
-    Perform ``array[idx] -= value``. Support int32, int64, float32 and
+    Perform ``array[idx] -= value``. Supports int32, int64, float32 and
     float64 only. The ``idx`` argument can be an integer or a tuple of integer
-    indices for indexing into multiple dimensional arrays. The number of element
-    in ``idx`` must match the number of dimension of ``array``.
+    indices for indexing into multi-dimensional arrays. The number of elements
+    in ``idx`` must match the number of dimensions of ``array``.
 
     Returns the value of ``array[idx]`` before the storing the new value.
     Behaves like an atomic load.
--- a/docs/source/cuda-reference/kernel.rst
+++ b/docs/source/cuda-reference/kernel.rst
@@ -175,6 +175,16 @@ Synchronization and Atomic Operations
     Returns the value of ``array[idx]`` before the storing the new value.
     Behaves like an atomic load.
 
+.. function:: numba.cuda.atomic.sub(array, idx, value)
+
+    Perform ``array[idx] -= value``. Support int32, int64, float32 and
+    float64 only. The ``idx`` argument can be an integer or a tuple of integer
+    indices for indexing into multiple dimensional arrays. The number of element
+    in ``idx`` must match the number of dimension of ``array``.
+
+    Returns the value of ``array[idx]`` before the storing the new value.
+    Behaves like an atomic load.
+
 .. function:: numba.cuda.atomic.max(array, idx, value)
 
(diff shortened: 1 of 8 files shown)


gmarkall:
@testhound Many thanks for the PR! The tests are presently failing because the [CUDA Simulator](https://numba.readthedocs.io/en/stable/cuda/simulator.html) doesn't implement atomic subtract - to implement support for it, you can add a `sub` method to [`FakeCUDAAtomic`](https://github.com/numba/numba/blob/master/numba/cuda/simulator/kernelapi.py#L105) following a similar pattern to that implemented for `add`, `max`, etc.

A couple of general notes on testing: The public CI tests CUDA only with the simulator (mainly to ensure that the simulator functions correctly, but also there is no GPU CI for Numba). You can test locally with the simulator by setting the environment variable `NUMBA_ENABLE_CUDASIM=1`.

gmarkall (changes requested):
On the whole the PR is looking great! There are some small suggestions on the diff, which are all minor things just aimed at making the code and docstrings a bit clearer / more comprehensible for the future - I know that the majority of this is derived from the existing implementations of atomic add, etc., so the things I'm pointing out are inherited from them, but I hope that for new code we can improve it a bit as we go along as well.

Between these small fixes, and the implementation of atomic subtract in the simulator, that should get this PR ready to go.

gmarkall (approved):
Many thanks for the changes! This is looking good to me, and passes all tests with the simulator and the device target.

stuartarchibald:
Thanks for the PR. There's more scope in here for generating and sharing code across `add` and `sub` particularly in the testing, but I think in this case it's better to get the functionality in and refactor in a subsequent PR. Will run through the farm now.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_102`.

stuartarchibald:
Buildfarm passed.

stuartarchibald (approved):

stuartarchibald:
Thanks for reviewing @gmarkall 

Congratulations on your first contribution to Numba @testhound !

testhound:
> Thanks for the PR. There's more scope in here for generating and sharing code across `add` and `sub` particularly in the testing, but I think in this case it's better to get the functionality in and refactor in a subsequent PR. Will run through the farm now.

@stuartarchibald can you provide some of your thoughts about additional code sharing? I am beginning work on the remaining atomics and would like to keep our ideas in mind.

stuartarchibald:
> > Thanks for the PR. There's more scope in here for generating and sharing code across `add` and `sub` particularly in the testing, but I think in this case it's better to get the functionality in and refactor in a subsequent PR. Will run through the farm now.
> 
> @stuartarchibald can you provide some of your thoughts about additional code sharing? I am beginning work on the remaining atomics and would like to keep our ideas in mind.

I think in cases where there's a lot of repetition and the variations are just on a function from e.g. the `operator` module it's possible to make use of code generation and/or to write the invariant part and loop over the variations. Example of code gen: https://github.com/numba/numba/commit/9e54c7d78a2d35379724e9e8dfb1b225fd887d96

In the case of the unit tests, all the add/sub tests are identical apart from the operation, I would consider just iterating over the functions (add/sub) with the same test code in the loop body. This would make it so additional tests written would automatically get picked up for both `add` and `sub` and it present one test series to maintain instead of two.

@gmarkall are you happy with the above suggestion? There's a fine line between obfuscation and maintainability to tread but I think in cases like this shrinking the literal quantity of code is beneficial.

gmarkall:
> @gmarkall are you happy with the above suggestion? There's a fine line between obfuscation and maintainability to tread but I think in cases like this shrinking the literal quantity of code is beneficial.

I'm happy with both suggestions - I was thinking that the duplication in `cudadecl` was going to get a little bit much (though that didn't worry me so much) but the atomics tests will get pretty long if a few more operations are added.

PR #6252, opened 2020-09-16
Show channel URLs

sklam:
for inspecting when free packages we are using


sklam:
There's only one instances of `pkg/free` usage for `jedi               pkgs/free/linux-64::jedi-0.10.2-py36_2`. Perhaps we can stop using on `restore_free_channel`.

stuartarchibald (approved):

Issue #5179, opened 2020-01-31
TupleTestHashing unit tests fails on 32-bit Windows in Python 3.8


jjhelmus:
- [x] I am using the latest released version of Numba (most recent is visible in
 the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [x] I have included below a minimal working reproducer (if you are unsure how
 to write one see http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports).

The `numba.tests.test_hashing.TestTupleHasing` tests fail when run on a Windows with a 32-bit Python 3.8 interpreter.

```
(base) C:\Users\jhelmus>set CONDA_SUBDIR=win-32

(base) C:\Users\jhelmus>conda create -n test_numba python=3.8 numba=0.48.0 numpy=1.18
Collecting package metadata (current_repodata.json): done
Solving environment: done
...
(base) C:\Users\jhelmus>conda activate test_numba
(test_numba) C:\Users\jhelmus>python -m numba.runtests -k "*TupleHash*" -v
skipped CUDA tests
skipped CUDA tests
skipped HSA tests
skipped HSA tests
test_heterogeneous_tuples (numba.tests.test_hashing.TestTupleHashing) ... val, nb_hash, hash(val)
(0, 0.0) 207041508 -168982784
abs(val), hashing._PyHASH_MODULUS - 1
ERROR
test_homogeneous_tuples (numba.tests.test_hashing.TestTupleHashing) ... val, nb_hash, hash(val)
(0, 0) 207041508 -168982784
abs(val), hashing._PyHASH_MODULUS - 1
ERROR
test_py27 (numba.tests.test_hashing.TestTupleHashing) ... skipped 'Python 2 only test'

======================================================================
ERROR: test_heterogeneous_tuples (numba.tests.test_hashing.TestTupleHashing)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\jhelmus\mc\envs\test_numba\lib\site-packages\numba\tests\test_hashing.py", line 52, in check_hash_values
    self.assertEqual(nb_hash, hash(val))
  File "C:\Users\jhelmus\mc\envs\test_numba\lib\unittest\case.py", line 912, in assertEqual
    assertion_func(first, second, msg=msg)
  File "C:\Users\jhelmus\mc\envs\test_numba\lib\unittest\case.py", line 905, in _baseAssertEqual
    raise self.failureException(msg)
AssertionError: 207041508 != -168982784

[...]

PR #6250, opened 2020-09-16
Skip some failing-on-32-bit tests (may be windows only)

mingwandroid:

PR #6254, opened 2020-09-17
Fix win-32 hashing bug (from Stuart Archibald)

mingwandroid:

### DIFF
--- a/numba/cpython/hashing.py
+++ b/numba/cpython/hashing.py
@@ -264,7 +264,7 @@ if _py38_or_later:
         @register_jitable(locals={'x': types.uint64})
         def _PyHASH_XXROTATE(x):
             # Rotate left 13 bits
-            return ((x << types.uint64(13)) | (x >> types.uint64(16)))
+            return ((x << types.uint64(13)) | (x >> types.uint64(19)))
 
     # Python 3.7+ has literal_unroll, this means any homogeneous and
     # heterogeneous tuples can use the same alg and just be unrolled.


mingwandroid:
With this change, the Anaconda Distro CI passes green on all builds. CC @gmarkall @stuartarchibald

stuartarchibald:
xref: Closes https://github.com/numba/numba/issues/5179

@mingwandroid Thanks for following up on the issue, patching and testing! Much appreciated.

xref: https://github.com/numba/numba/pull/6250#issuecomment-693529389 my guess at the patch!

mingwandroid:
Should be good to go now. Thanks all.

stuartarchibald (approved):
Thanks again @mingwandroid!

numba/cpython/hashing.py
```diff
@@ -264,7 +264,7 @@ def _PyHASH_XXROTATE(x):
         @register_jitable(locals={'x': types.uint64})
         def _PyHASH_XXROTATE(x):
             # Rotate left 13 bits
-            return ((x << types.uint64(13)) | (x >> types.uint64(16)))
+            return ((x << types.uint64(13)) | (x >> types.uint64(19)))
```
stuartarchibald, line 267: For reference, this line https://github.com/python/cpython/blob/b738237d6792acba85b1f6e6c8993a812c7fd815/Objects/tupleobject.c#L362 is the source of these constants.

PR #6261, opened 2020-09-21
Add np.ndarray.ptp() support.

### LINKED ISSUE #6257: .ptp() not @njit compilable anymore?
 

I cant pass numpy arrays like the following to an `njit` compiled function that contains` numpy.ptp()`.
Why is that ( I think here it says, that .ptp() is supported: https://numba.pydata.org/numba-doc/dev/reference/numpysupported.html)
 ```
a = [0.         ,0.         ,0.         ,0.         ,0.         ,0.
 ,0.         ,0.         ,0.00224828 ,0.00317533 ,0.00830591 ,0.00934349  
 ,0.80603907 ,0.80603907 ,0.80603907 ,0.80603907 ,0.80603907 ,0.94597275,
 1.,        1.,        1.,        1.,        1.,        1., 
 1.,       ]

import math
from numba import  prange
@nb.njit(nopython=True,parallel=True)
def anynan(array): 
    for i in prange(array.size):
        if math.isnan(array[i]):
            return True
    return False
@nb.njit
def scale_zero2one(x):
      
    x -= x.min()  
    x /= x.ptp()   
    if anynan(x):  
        x[np.isnan(x)] = 0               
    return x 

a = np.asarray(a,dtype=np.float32) 
print(scale_zero2one(a))
```

stuartarchibald:
As title, just wiring.

Fixes #6257

### DIFF
--- a/docs/source/reference/numpysupported.rst
+++ b/docs/source/reference/numpysupported.rst
@@ -224,6 +224,7 @@ The following methods of Numpy arrays are supported:
 * :meth:`~numpy.ndarray.flatten` (no order argument; 'C' order only)
 * :meth:`~numpy.ndarray.item` (without arguments)
 * :meth:`~numpy.ndarray.itemset` (only the 1-argument form)
+* :meth:`~numpy.ndarray.ptp` (without arguments)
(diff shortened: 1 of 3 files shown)


esc (approved):
LGTM

Issue #2201, opened 2016-11-15
dtypes passes as strings don't work in array creation methods


seibert:
This kind of function does not work:
```python
@numba.jit(nopython=True)
def test_func():
    x = numpy.empty(100, dtype='int64')
    x[0] = 10
    return x
```
although `int64` is interpreted as a valid dtype by NumPy.

Could this be solved by creating a casting rule that allows strings to be cast to dtype objects?

sklam:
> Could this be solved by creating a casting rule that allows strings to be cast to dtype objects?

better make an overload to take `dtype=str'` instead of new casting rule.

seibert:
yeah, probably, though any function that takes a dtype object should probably do the same conversion

stuartarchibald:
Also, with reference to #2205 

```python
from numba import cuda
@cuda.jit('void(float64[:])')
def kern(arr):
    cuda.local.array(10, 'float64')

kern[1,1](np.zeros(1))
```
gives:
```
numba.errors.LoweringError: Failed at nopython (nopython mode backend)
cannot create weak reference to 'str' object
File "issue2205.py", line 4
[1] During: lowering "$0.6 = call ptx.lmem.alloc($const0.4, $const0.5)" at issue2205.py (4)
```

esc:
Since #6262 has now been merged, can this be closed? It seems like #6262 only partially addressed this issue, perhaps we could update here what is still missing, if anything.

stuartarchibald:
> Since #6262 has now been merged, can this be closed? It seems like #6262 only partially addressed this issue, perhaps we could update here what is still missing, if anything.

Technically I think yes https://github.com/numba/numba/pull/6262#pullrequestreview-493577596

stuartarchibald:
Updated https://github.com/numba/numba/issues/2205 as feature request. Suggest opening new tickets for any remaining issues.

PR #6262, opened 2020-09-21
Support dtype from str literal.

stuartarchibald:
As title.

### DIFF
--- a/numba/tests/test_array_methods.py
+++ b/numba/tests/test_array_methods.py
@@ -151,6 +151,9 @@ def np_frombuffer(b):
 def np_frombuffer_dtype(b):
     return np.frombuffer(b, dtype=np.complex64)
 
+def np_frombuffer_dtype_str(b):
+    return np.frombuffer(b, dtype='complex64')
+
 def np_frombuffer_allocated(shape):
     """
     np.frombuffer() on a Numba-allocated buffer.
@@ -484,6 +487,14 @@ class TestArrayMethods(MemoryLeakMixin, TestCase):
             expected = pyfunc(buf)
             self.memory_leak_setup()
             got = run(buf)
+            # TODO: why is run() producing a frame that holds reference if the
+            # dtype is a str literal?
+            if 'dtype_str' in self.id():
+                import gc
+                frames = gc.get_referrers(buf)
+                if frames:
+                    del frames[0]
+                gc.collect()
             self.assertPreciseEqual(got, expected)
             del expected
             self.assertEqual(sys.getrefcount(buf), old_refcnt + 1)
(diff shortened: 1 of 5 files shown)


stuartarchibald:
This adds most of the necessary for fixing #2201. `np.ndarray.{view,astype}` still don't work as the literal values are being lost somewhere, for maintainability/efficiency, probably best to just refactor to the extension API than try and work out what's going wrong.

stuartarchibald:
Need to make tests integer size invariant.

numba/tests/test_array_methods.py
```diff
@@ -484,6 +487,14 @@ def check(buf):
...
             self.memory_leak_setup()
             got = run(buf)
+            # TODO: why is run() producing a frame that holds reference if the
+            # dtype is a str literal?
+            if 'dtype_str' in self.id():
+                import gc
+                frames = gc.get_referrers(buf)
+                if frames:
+                    del frames[0]
+                gc.collect()
```
stuartarchibald, line 497: @sklam this is the weird frame holding ref problem.

sklam: the frame is from `run`? 
```python
<frame at 0x614314410, file 'numba/tests/test_array_methods.py', line 484, code run>
```

stuartarchibald: yes, I saw the same, didn't debug much further

sklam: Just tried to separate the compiling and running. It's certainly from the compiling; i.e. `self.ccache.compile(pyfunc, (typeof(buf),))`

sklam: it's the try...except in `support.py:126`

sklam: so with this patch:

```diff
diff --git a/numba/tests/support.py b/numba/tests/support.py
index ae2430f1a..f9c05a7fe 100644
--- a/numba/tests/support.py
+++ b/numba/tests/support.py
@@ -123,9 +123,9 @@ class CompilationCache(object):
         from numba.core.registry import cpu_target

         cache_key = (func, args, return_type, flags)
-        try:
+        if cache_key in self.cr_cache:
             cr = self.cr_cache[cache_key]
-        except KeyError:
+        else:
             # Register the contexts in case for nested @jit or @overload calls
             # (same as compile_isolated())
             with cpu_target.nested_context(self.typingctx, self.targetctx):
```

stuartarchibald: ah, that would explain it, thanks!

stuartarchibald: 3be2086 implements

sklam:
This is making me wonder if we should let `StringLiteral` to cast to `DTypeSpec`

sklam:
RE to myself: https://github.com/numba/numba/pull/6262#issuecomment-696781829

It probably will be simpler once we switch to `@overload` base

stuartarchibald:
> RE to myself: [#6262 (comment)](https://github.com/numba/numba/pull/6262#issuecomment-696781829)
> 
> It probably will be simpler once we switch to `@overload` base

I think it'll be easier once everything is `@overload`. This PR is to implement a long standing request, but is limited in that anything too tricky isn't done as it'll be easier to handle once written as `@overload`s.

sklam (commented):
This patch adds string dtype spelling for:

- `empty_like`
- `zeros_like`
- `full`
- `full_like`
- `ones`
- `ones_like`
- `identity`
- `from_buffer`
- `array`

sklam (approved):

PR #6265, opened 2020-09-21
Fix #6260

### LINKED ISSUE #6260: Type cannot be determined when using 'if a not in b' in list construction
- [x] I have tried using the latest released version of Numba

In trying to make a replacement for numpy's setxor1d I came across the following error (minimal example below):
```
@nb.njit()
def test():
    included = np.array([1, 2, 6, 8])
    not_included = [i for i in range(10) if i not in list(included)]
    return not_included
```
Which results in:

`Type of variable '$18compare_op.7.7054' cannot be determined, operation: unary(fn=<built-in function not_>, value=$18compare_op.7.7054), location: <ipython-input-138-e85438f982c0> (4)`

Which does not occur when the 'not' statement (in if i not in list(included)]) is not there
I could get around this by instead doing:
```
@nb.njit()
def test():
    included = np.array([1, 2, 6, 8])
    not_included = [i for i in range(10) if (i in list(included)) is False]
    return not_included
```
Which is not very pretty.

stuartarchibald:
get_ir_of_code doesn't necessarily manage to produce SSA such
that variables can be multiply defined. This leads to problems
including but not limited to nonsense IR with nodes referring to
their own label and DCE eliminating variables with the same name
that shouldn't have had the same name.

Closes #6260.

### DIFF
--- a/numba/core/ir_utils.py
+++ b/numba/core/ir_utils.py
@@ -1669,12 +1669,30 @@ def get_ir_of_code(glbls, fcode):
             self.state.typemap = None
             self.state.return_type = None
             self.state.calltypes = None
-    rewrites.rewrite_registry.apply('before-inference', DummyPipeline(ir).state)
+    state = DummyPipeline(ir).state
+    rewrites.rewrite_registry.apply('before-inference', state)
     # call inline pass to handle cases like stencils and comprehensions
     swapped = {} # TODO: get this from diagnostics store
     inline_pass = numba.core.inline_closurecall.InlineClosureCallPass(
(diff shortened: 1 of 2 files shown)


esc (approved):
I tested this locally, and it passes the unit test. I also looked at the previously malformed IR from a simpler variant of the reproducer in #6260 and found that previously missing `x in y` line is now present before the unary not:

```
    $14compare_op.5 = i in $const12.4        ['$14compare_op.5', '$const12.4', 'i']
    $14compare_op.5.1 = unary(fn=<built-in function not_>, value=$14compare_op.5) ['$14compare_op.5', '$14compare_op.5.1']
```

PR #6271, opened 2020-09-22
Support `ndarray` contains

### LINKED ISSUE #6259: Add support for `ndarray` 'contains'
- [x] I have tried using the latest released version of Numba

In trying to replace numpy's setxor1d (which is unfortunately not supported by Numba), I came across an error that does occur for (integer) ndarrays but not for lists.
In my case the setxor1d can be replaced by [i for i in a if i not in b], but for the sake of testing, here is a different minimal working example:

```
@nb.njit()
def test():
    included = np.array([1, 2, 6, 8])
    return 6 in included
```
This will give the error:

```
No implementation of function Function(<built-in function contains>) found for signature:

 >>> contains(array(int64, 1d, C), Literal[int](6))
```

While the funtion:
```
@nb.njit()
def test():
    included = np.array([1, 2, 6, 8])
    return 6 in list(included)
```

Works fine. 
I'm assuming it is supposed to work for the array as well.
I came accross this previous issue that looks similar at first glance: [github.com/numba/numba/issues/4150](https://github.com/numba/numba/issues/4150)
But there it concerns an array of floats, so it is evident that 'int in array(float)' is not going to work.

mugoh:
Add support for `ndarray` contains

resolves #6259

### DIFF
--- a/docs/source/reference/numpysupported.rst
+++ b/docs/source/reference/numpysupported.rst
@@ -251,6 +251,17 @@ The following methods of Numpy arrays are supported:
 
 * :meth:`~numpy.ndarray.transpose`
 * :meth:`~numpy.ndarray.view` (only the 1-argument form)
+* :meth:`~numpy.ndarray.__contains__` (only the 1-argument form ``key``)
+
+.. code:: pycon
+    >>> import numpy as np
+    >>> from numba import njit
+    >>> @njit
+    ... def nd_contains():
+    ...     arr = np.array([0, 3, 7, 5])
+    ...     print(5 in arr) 
+    >>> nd_contains()
+    True
 
 
 .. warning::
--- a/docs/source/reference/numpysupported.rst
+++ b/docs/source/reference/numpysupported.rst
@@ -253,16 +253,6 @@ The following methods of Numpy arrays are supported:
 * :meth:`~numpy.ndarray.view` (only the 1-argument form)
 * :meth:`~numpy.ndarray.__contains__` (only the 1-argument form ``key``)
 
-.. code:: pycon
-    >>> import numpy as np
-    >>> from numba import njit
-    >>> @njit
-    ... def nd_contains():
-    ...     arr = np.array([0, 3, 7, 5])
-    ...     print(5 in arr) 
-    >>> nd_contains()
-    True
-
 
 .. warning::
    Sorting may be slightly slower than Numpy's implementation.
--- a/docs/source/reference/numpysupported.rst
+++ b/docs/source/reference/numpysupported.rst
@@ -251,7 +251,7 @@ The following methods of Numpy arrays are supported:
 
 * :meth:`~numpy.ndarray.transpose`
 * :meth:`~numpy.ndarray.view` (only the 1-argument form)
-* :meth:`~numpy.ndarray.__contains__` (only the 1-argument form ``key``)
+* :meth:`~numpy.ndarray.__contains__` 
 
 
 .. warning::
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -7,6 +7,7 @@ import math
 from collections import namedtuple
 from enum import IntEnum
 from functools import partial
+import operator
 
 import numpy as np
 
@@ -154,7 +155,7 @@ def _gen_index_tuple(tyctx, shape_tuple, value, axis):
     return function_sig, codegen
 
 
-#----------------------------------------------------------------------------
+# ----------------------------------------------------------------------------
 # Basic stats and aggregates
 
 @lower_builtin(np.sum, types.Array)
@@ -1134,7 +1135,7 @@ def np_ptp(a):
     return np_ptp_impl
 
 
-#----------------------------------------------------------------------------
+# ----------------------------------------------------------------------------
 # Median and partitioning
 
 @register_jitable
@@ -1541,7 +1542,7 @@ def np_partition(a, kth):
     return np_partition_impl
 
 
-#----------------------------------------------------------------------------
+# ----------------------------------------------------------------------------
 # Building matrices
 
 @register_jitable
@@ -1727,7 +1728,7 @@ def _dtype_of_compound(inobj):
         if isinstance(obj, (types.Number, types.Boolean)):
             return as_dtype(obj)
         l = getattr(obj, '__len__', None)
-        if l is not None and l() == 0: # empty tuple or similar
+        if l is not None and l() == 0:  # empty tuple or similar
             return np.float64
         dt = getattr(obj, 'dtype', None)
         if dt is None:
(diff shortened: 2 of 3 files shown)


mugoh:
Tests fail with
```
No implementation of function Function(<built-in function contains>) found for signature:
 
 >>> contains(array(int64, 1d, C), int64)
 
There are 18 candidate implementations:
  - Of which 18 did not match due to:
  Overload of function 'contains': File: <numerous>: Line N/A.
    With argument(s): '(array(int64, 1d, C), int64)':
   No match.

During: typing of intrinsic-call at /Users/runner/work/1/s/numba/tests/test_np_functions.py (324)

File "numba/tests/test_np_functions.py", line 324:
def __contains__(a, key):
    return key in a
    ^
```

It has to do with how I'm trying to override `contains` [at this part of the code](https://github.com/numba/numba/blob/a14c5f85ec3b17d8b466ef9bb2f6d5b6a7b503e9/numba/np/arraymath.py#L4299), but I'm having trouble figuring out what should fix it.

esc:
@mugoh thank you for submitting this. Unfortunately we can not review your submission in it's current state. If we look at the  changes introduced:

https://github.com/numba/numba/pull/6271/files

It seems like there are a large number of cosmetic reformatting of the code. My guess would be, that you editor is configured to automatically reformat files that you edit. This makes it impossible for a reviewer to understand what changes actually pertain to the feature being implemented as overall 'diff'  is huge. My advice would be to go back and alter the git history to remove all of the cosmetic changes being made so that only changes that are part of the `contains` implementation remain.

We do really appreciate your efforts on this front and I am sorry we can't review the PR in it's current state.

mugoh:
Looking into the large-diff issue @esc

numba/np/arraymath.py
```diff
@@ -3303,6 +3303,18 @@ def np_imag_impl(a):
...
+@overload_method(types.Array, "__contains__")
+def np_contains(arr, key):
+
+    if not isinstance(arr, types.Array):
+        return
+
+    def np_contains_impl(arr, key):
+        return key in arr
+
+    return np_contains_impl
```
stuartarchibald, line 3315: I think the reason all the CI jobs are failing is because there's no logic implemented for dispatching to `__contains__` or to do the actual work for `key in arr`.

stuartarchibald:
Working on top of this PR, this demonstrates the sort of thing that should help:
```python
import operator
from numba import njit, types
import numpy as np

@overload(operator.contains)
def ol_contains_array(a, b):
        return
    def impl(a, b):
        for x in np.nditer(a):
            if x == b:
                return True
        return False
    return impl

@njit
def foo(a, b):
    return b in a

print(foo(np.ones(4), 1))
print(foo(np.ones(4), 0))
```

esc:
@mugoh thanks for submitting this. It seems like there are still a number of cosmetic changes in the diff. Also, there is a build failure for the docs on Azure pipelines that will need to be looked at:

```
Warning, treated as error:
/home/vsts/work/1/s/docs/source/reference/numpysupported.rst:257:Error in "code" directive:
maximum 1 argument(s) allowed, 30 supplied.

.. code:: pycon
    >>> import numpy as np
    >>> from numba import njit
    >>> @njit
    ... def nd_contains():
    ...     arr = np.array([0, 3, 7, 5])
    ...     print(5 in arr)
    >>> nd_contains()
    True
```

esc (changes requested):
@mugoh thanks for following up on this. The implementation looks good so far! As for the tests, I think it would be good to include a few different dtypes (float,int, complex... maybe) and to mix them during lookup. For example, looking for a float in an int based array?

But also at least the following edge cases ought to be tested:

* Empty array
* Single value array
* Numpy scalar

Keep up the good work! 👍

mugoh:
Sure!

esc:
@mugoh thanks for looking in this some more. It seems you may have inadvertently introduced some flake8 violations as discovered by our CI systems:

```
numba/tests/test_np_functions.py:469:18: E271 multiple spaces after keyword
numba/tests/test_np_functions.py:472:13: E303 too many blank lines (2)
numba/tests/test_np_functions.py:484:13: E303 too many blank lines (2)
numba/tests/test_np_functions.py:490:9: E303 too many blank lines (3)
```

numba/tests/test_np_functions.py
```diff
@@ -436,6 +440,59 @@ def check(x_types, x_values, **kwargs):
...
+            empty_arr = np.asarray([])
+            yield empty_arr, k_out
+
+            # np scalars
+
+            bool_arr = np.array([True, False])
+            yield bool_arr, True
+            yield bool_arr, k_0
+
+            float_arr = np.random.rand(10)
```
esc, line 471: In order to make the test stable, you probably ought to seed the random number generator.

numba/tests/test_np_functions.py
```diff
@@ -436,6 +440,59 @@ def check(x_types, x_values, **kwargs):
...
+            present_k = float_arr[0]
+
+            yield float_arr, rand_k
+            yield float_arr, present_k
+
+            complx_arr = float_arr.view(np.complex128)
+            yield complx_arr, complx_arr[0]
+            yield complx_arr, rand_k
+
+            uint_arr = np.random.randint(10, size=15, dtype=np.uint8)
```
esc, line 482: In order to make the test stable, you probably ought to seed the random number generator.

esc (changes requested):
Some seeding of the RNG still needed, but probably ready to merge after that.

esc (approved):

esc:
@mugoh thanks for following up on this! I have reviewed all the changes you added and it LGTM!

docs/source/reference/numpysupported.rst
```diff
@@ -251,6 +251,7 @@ The following methods of Numpy arrays are supported:
 
 * :meth:`~numpy.ndarray.transpose`
 * :meth:`~numpy.ndarray.view` (only the 1-argument form)
+* :meth:`~numpy.ndarray.__contains__` (only the 1-argument form ``key``)
```
gmarkall, line 254: What does "(only the 1-argument form ``key``)" mean? Should this just be

```suggestion
* :meth:`~numpy.ndarray.__contains__`
```

?

gmarkall (commented):
The code and tests look good to me, but I'm confused by the docs - see the comment on the diff. In case I'm missing something, I'll mark this as a comment rather than a change request.

docs/source/reference/numpysupported.rst
mugoh: Here, I was trying to frame `__contains__` as a method that takes a single argument `key`, for use as `key in some_array`. It does sound more understandable without  "(only the 1-argument form `key`)" though.

esc:
@gmarkall thanks for the review!

@mugoh the changes look, I think this can be marked for merge now.

PR #6275, opened 2020-09-23
Switch on SLP vectorize.

stuartarchibald:
As title.


stuartarchibald:
Seeing what impact this has on testing run times.

stuartarchibald:
Command: `time ./runtests.py --random=0.1`

Mainline does this:

```
Ran 995 tests in 636.697s

OK (skipped=19, expected failures=1)

real    10m42.166s
user    11m33.653s
sys     0m7.492s
```

This branch does this:
```
Ran 995 tests in 632.633s

OK (skipped=19, expected failures=1)

real    10m37.993s
user    11m38.904s
sys     0m7.392s
```
seems like the impact on the test suite run time is potentially small/zero.

stuartarchibald:
Mainline:
```pycon
In [1]: %cpaste                                                                                                                 
Pasting code; enter '--' alone on the line to stop or use Ctrl-D.
:from numba import njit
import numpy as np

@njit
def foo(a1, a2, b1, b2, A):
    for x in range(0, len(A), 4):
        Z = A[x:x + 4]
        Z[0] = a1*(a1 + b1)
        Z[1] = a2*(a2 + b2)
        Z[2] = a1*(a1 + b1)
        Z[3] = a2*(a2 + b2)

foo(1.2, 3.4, 5.6, 7.8, np.empty(4000))::::::::::::
:<EOF>

In [2]: G = np.empty(40000)                                                                                                     

In [3]: %timeit foo(1.2, 3.4, 5.6, 7.8, G)                                                                                      
12.9 µs ± 2.75 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

This branch:
```pycon
In [1]: %cpaste                                                                                                                 
Pasting code; enter '--' alone on the line to stop or use Ctrl-D.
:from numba import njit
import numpy as np

@njit
def foo(a1, a2, b1, b2, A):
    for x in range(0, len(A), 4):
        Z = A[x:x + 4]
        Z[0] = a1*(a1 + b1)
        Z[1] = a2*(a2 + b2)
        Z[2] = a1*(a1 + b1)
        Z[3] = a2*(a2 + b2)

foo(1.2, 3.4, 5.6, 7.8, np.empty(4000))::::::::::::
:<EOF>

In [2]: G = np.empty(40000)                                                                                                     

In [3]: %timeit foo(1.2, 3.4, 5.6, 7.8, G)                                                                                      
10.3 µs ± 22.8 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

```

numba/core/codegen.py
```diff
@@ -714,7 +714,8 @@ def _pass_manager_builder(self):
         missed...
         """
         pmb = lp.create_pass_manager_builder(
-            opt=config.OPT, loop_vectorize=config.LOOP_VECTORIZE)
+            opt=config.OPT, loop_vectorize=config.LOOP_VECTORIZE,
+            slp_vectorize=True)
```
sklam, line 718: Shouldn't this also get a `config` flag? e.g. `config.SLP_VECTORIZE`

stuartarchibald: Can do, but I really want to move away from specific config flags for each feature in the optimiser, I think it should be in the hands of the user but also it shouldn't be spelled one at a time like this. For 0.52 shall we do a flag so it's easy to debug/turn off but this is the last one like it!?!

sklam: Ok, make it the last flag.

sklam (approved):

PR #6277, opened 2020-09-23
fix: cross2d wrong doc. reference (issue #6276)

### LINKED ISSUE #6276: numba advices to use cross2d but specified path is wrong
Hello everyone,

# The issue (for which I propose a fix below)

While using numba, I got an error when trying to compile this code:
```python
@numba.njit(boolean(float64[:, :]))
def is_convex(polygon):
    vertices = get_vertices(polygon)
    m, _ = vertices.shape
    vectors = np.empty((m, 2), dtype=np.float64)
    for i, j in vertices:
        vectors[i] = polygon[j, :] - polygon[i, :]
        
    for i, j in vertices:
        if np.cross(vectors[i,:], vectors[j, :]) < 0:
            return False
    return True
```

The error I get:
```python
is_convex(polygon)
>>> ---------------------------------------------------------------------------
>>> ValueError                                Traceback (most recent call last)
>>> <ipython-input-18-dc4c8f71d7e1> in <module>
>>> ----> 1 is_convex(polygon)
>>> 
>>> ~/.local/lib/python3.6/site-packages/numba/np/arraymath.py in impl()
>>>    4380         else:
>>>    4381             raise ValueError((
>>> -> 4382                 "Dimensions for both inputs is 2.\n"
>>>    4383                 "Please replace your numpy.cross(a, b) call with "
>>>    4384                 "numba.numpy_extensions.cross2d(a, b)."
>>> 
>>> ValueError: Dimensions for both inputs is 2.
>>> Please replace your numpy.cross(a, b) call with numba.numpy_extensions.cross2d(a, b).
```

The problem is that `numba.numpy_extensions.cross2d` does not exist:
```python
numba.numpy_extensions.cross2d
>>> ---------------------------------------------------------------------------
>>> AttributeError                            Traceback (most recent call last)
>>> <ipython-input-19-a634a239d236> in <module>
>>> ----> 1 numba.numpy_extensions.cross2d
>>> 
>>> AttributeError: module 'numba' has no attribute 'numpy_extensions'
```

In the documentation, the reference is `numba.np.extensions.cross2d(a, b) (c.f. https://numba.pydata.org/numba-doc/dev/reference/numpysupported.html#other-functions). But it does not exist neither:

```python
numba.np.extensions.cross2d
>>> ---------------------------------------------------------------------------
>>> AttributeError                            Traceback (most recent call last)
>>> <ipython-input-20-3ddb1ccb7c32> in <module>
>>> ----> 1 numba.np.extensions.cross2d
>>> 
>>> AttributeError: module 'numba.np' has no attribute 'extensions'
```

The version I'm using:
```python
import numba
numba.__version__
>>> 0.51.2
```

Before posting this issue, I quickly read the release notes and it does not seem to be fixed in the 0.52.0 version.
https://numba.pydata.org/numba-doc/latest/release-notes.html

I hope my issue is not a duplicate.

# A fix suggestion

Digging in the `numba.np` module, I found that the correct path to the `cross2d` function is actually:
```python
numba.np.arraymath._cross2d_operation(a, b)
```

If it is a documentation error, I can easily make a PR with the fix as this will be a good way for me to start contributing to this nice project !

jeertmans:
See issue #6276 for more details

Closes #6276

### DIFF
--- a/CHANGE_LOG
+++ b/CHANGE_LOG
@@ -1199,7 +1199,7 @@ Enhancements from user contributed PRs (with thanks!):
 * Lucio Fernandez-Arjona extended Numba's ``np.sum`` support to now accept the
   ``dtype`` kwarg in #4472.
 * Pedro A. Morales Maries added support for ``np.cross`` in #4128 and also added
-  the necessary extension ``numba.numpy_.extensions.cross2d`` in #4595.
+  the necessary extension ``numba.numpy_extensions.cross2d`` in #4595.
 * David Hoese, Eric Firing, Joshua Adelman, and Juan Nunez-Iglesias all made
   documentation fixes in #4565, #4482, #4455, #4375 respectively.
 * Vyacheslav Smirnov and Rujal Desai enabled support for ``count()`` on unicode
--- a/CHANGE_LOG
+++ b/CHANGE_LOG
@@ -1199,7 +1199,7 @@ Enhancements from user contributed PRs (with thanks!):
 * Lucio Fernandez-Arjona extended Numba's ``np.sum`` support to now accept the
   ``dtype`` kwarg in #4472.
 * Pedro A. Morales Maries added support for ``np.cross`` in #4128 and also added
-  the necessary extension ``numba.numpy_extensions.cross2d`` in #4595.
+  the necessary extension ``numba.np.extensions.cross2d`` in #4595.
 * David Hoese, Eric Firing, Joshua Adelman, and Juan Nunez-Iglesias all made
   documentation fixes in #4565, #4482, #4455, #4375 respectively.
 * Vyacheslav Smirnov and Rujal Desai enabled support for ``count()`` on unicode
--- a/CHANGE_LOG
+++ b/CHANGE_LOG
@@ -1199,7 +1199,7 @@ Enhancements from user contributed PRs (with thanks!):
 * Lucio Fernandez-Arjona extended Numba's ``np.sum`` support to now accept the
   ``dtype`` kwarg in #4472.
 * Pedro A. Morales Maries added support for ``np.cross`` in #4128 and also added
-  the necessary extension ``numba.np.extensions.cross2d`` in #4595.
+  the necessary extension ``numba.numpy_.extensions.cross2d`` in #4595.
 * David Hoese, Eric Firing, Joshua Adelman, and Juan Nunez-Iglesias all made
   documentation fixes in #4565, #4482, #4455, #4375 respectively.
 * Vyacheslav Smirnov and Rujal Desai enabled support for ``count()`` on unicode
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -4395,7 +4395,7 @@ def np_cross(a, b):
             raise ValueError((
                 "Dimensions for both inputs is 2.\n"
                 "Please replace your numpy.cross(a, b) call with "
-                "numba.numpy_extensions.cross2d(a, b)."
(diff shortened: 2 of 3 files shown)


CHANGE_LOG
```diff
@@ -1199,7 +1199,7 @@ Enhancements from user contributed PRs (with thanks!):
 * Lucio Fernandez-Arjona extended Numba's ``np.sum`` support to now accept the
   ``dtype`` kwarg in #4472.
 * Pedro A. Morales Maries added support for ``np.cross`` in #4128 and also added
-  the necessary extension ``numba.numpy_extensions.cross2d`` in #4595.
+  the necessary extension ``numba.np.extensions.cross2d`` in #4595.
```
stuartarchibald, line 1202: As the change log is a reflection of history I don't think that this should be changed, irrespective of this being in the wrong place for current Numba.

numba/np/arraymath.py
```diff
@@ -4395,7 +4395,7 @@ def impl(a, b):
             raise ValueError((
                 "Dimensions for both inputs is 2.\n"
                 "Please replace your numpy.cross(a, b) call with "
-                "numba.numpy_extensions.cross2d(a, b)."
+                "numba.np.extensions.cross2d(a, b)."
```
stuartarchibald, line 4398:
```suggestion
                "a call to `cross2d(a, b)` from `numba.np.extensions`."
```

stuartarchibald (commented):
Thanks for the patch. Think it'd be a good idea to update the unit tests too to make sure it's less likely to happen again. Suggest updating the import here to reflect the public API location:
https://github.com/numba/numba/blob/db74d8ff90e3f7a06ba30575a2c5d7d521b1cf4d/numba/tests/test_np_functions.py#L18
and then checking the correct path is in the error string here:
https://github.com/numba/numba/blob/db74d8ff90e3f7a06ba30575a2c5d7d521b1cf4d/numba/tests/test_np_functions.py#L3589-L3592

Thanks!!

jeertmans:
Yes indeed, thank you for noticing @stuartarchibald . I did update the code.

I ran the test and it seems ok:

```
(numbaenv) jerome@jerome-N501VW:~/Documents/programmation/numba$ python -m numba.runtests numba.tests.test_np_functions.TestNPFunctions.test_cross_exceptions
>>> .
>>> ----------------------------------------------------------------------
>>> Ran 1 test in 2.907s
>>> 
>>> OK
```

CHANGE_LOG
jeertmans: Ok my bad, I was doubtful about changing this but thank you for the information

esc:
@jeertmans thanks for addressing the issue comments. From what I understand, @stuartarchibald suggested to add a check for the correct error string to the unit test. Something like:

```
 self.assertIn( 
     "a call to `cross2d(a, b)` from `numba.np.extensions`", 
     str(raises.exception) 
 ) 
```

To double check that the path listed in the error message is indeed the correct one.

jeertmans:
Thanks for the explanation @esc ! I did commit an additional check for that purpose. I hope it follows your standards.

esc:
@jeertmans excellent, thank you very much! I'll mark it as Ready to Merge under the assumption that @stuartarchibald will agree.

stuartarchibald (approved):

stuartarchibald:
@jeertmans Congratulations on your first contribution to Numba!

jeertmans:
Thank you both @stuartarchibald and @esc , I appreciate your help and I hope to be able to contribute more :)

PR #6282, opened 2020-09-24
Remove docs on Python 2(.7) EOL.

stuartarchibald:
As title. They are also incorrect with respect to reality, there
does not seem to be any value in keeping them!


esc (approved):
LGTM! The only bit of information that is perhaps relevant, is that 0.47 is the last Numba version to support Python 2.7 - perhaps we can add this as a footnote somewhere.

Issue #6264, opened 2020-09-21
Caching: functions captured in closure


sk1p:
## Feature request

I would like to capture functions in closures and be able to cache the compilation.  Currently, running the following code results in re-compilation on each run (that is, when starting in a new Python interpreter):

```python
import numpy as np
import numba


@numba.njit
def f1(in_arr):
    return in_arr * 2


def compose(fn):
    @numba.njit(cache=True)
    def _inner_fn(arr):
        return fn(arr)
    return _inner_fn


composed = compose(f1)
composed(np.ones((128, 128)))
```

Each time, a new version of `composed` is saved in `__pycache__`. This is caused by the UUID that is part of the `Dispatcher` - when looking at [`cvarbytes`](https://github.com/numba/numba/blob/3b3eab8922c9ca199fc3b2c758886628a2ab7494/numba/core/caching.py#L704), only the UUID differs between runs.

As a proof of concept, the following hack for `Cache._index_key` makes the `compose` consistently cache-able:

```
diff --git a/numba/core/caching.py b/numba/core/caching.py
index 45f82f314..f4f766790 100644
--- a/numba/core/caching.py
+++ b/numba/core/caching.py
@@ -701,11 +701,17 @@ class Cache(_Cache):
         codebytes = self._py_func.__code__.co_code
         if self._py_func.__closure__ is not None:
             cvars = tuple([x.cell_contents for x in self._py_func.__closure__])
+            cvars = tuple(var.__code__.co_code if hasattr(var, '__code__') else var
+                          for var in cvars)
             cvarbytes = dumps(cvars)
         else:
             cvarbytes = b''
 
         hasher = lambda x: hashlib.sha256(x).hexdigest()
         return (sig, codegen.magic_tuple(), (hasher(codebytes),
                                              hasher(cvarbytes),))
```

Implementing this as-is doesn't invalidate the cache in all cases, for example if there are multiple levels of closures. Maybe there is an easier and/or better way? Thanks!

sklam:
Thank you for your suggestion.

I think numba will also need to recursively cache the referenced jit function by specialing on the `Dispatcher` type as well.

cmdupuis3:
I am running into exactly this issue, just adding my +1.

PR #6284, opened 2020-09-24
Prevent cache busting by the UUID of Dispatcher

sk1p:
First steps towards fixing #6264.

This PR changes the cache key to explicitly include the bytecode of python functions captured in closures, and also their transitive dependencies. Before, the whole `Dispatcher` object was pickled and included in the key, which changes for each "run" (i.e. each time a new python interpreter is started) because of the UUID.

I'm not sure if this is enough, or if other attributes of the `Dispatcher` objects also need to be included in the key.

As @sklam noted in the issue, the referenced functions may also need to be cached (except if they are inlined?), but I'm not sure what the best approach would be, or if the user should be required to explicitly mark them to be cached.

Please let me know if this goes into the right direction, thanks!

### DIFF
--- a/numba/core/caching.py
+++ b/numba/core/caching.py
@@ -691,6 +691,25 @@ class Cache(_Cache):
             # No such conditions under non-Windows OSes
             yield
 
+    def _get_dependencies(self, cvar):
+        deps = [cvar]
+        if hasattr(cvar, 'py_func'):
+            # TODO: does the cache key need to depend on any other
+            # attributes of the Dispatcher?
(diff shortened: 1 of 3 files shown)


esc:
@sk1p thank you for submitting this enhancement to the Numba issue tracker! We appreciate your efforts to improve Numba and I have now added this PR to the review queue.

stuartarchibald:
Thanks for the patch, I think this is going to need similar https://github.com/numba/numba/pull/6406

PR #5719, opened 2020-05-18
[WIP] Fix #5713: cache typed dict and list functions

gmarkall:
Making this PR to test on CI.

PR #6287, opened 2020-09-28
Fix #5713: cache typed dict and list functions

### LINKED ISSUE #5713: List() and Dict() always recompile _getitem,_setitem, _length, etc. every time; maybe should cache?
- [ x] I am using the latest released version of Numba: 0.49.1
- [ x] I have included below a minimal working reproducer 

Something that has been bothering me is that for the typed List() and Dict() types every time I use a new type outside of a jitted function it will recompile all of the primitive functions I use with it. For example this simple example:

```python
import numba
from numba.types import f8,u8,i8
from numba.typed import Dict,List
import time

start = time.time()
for t1 in [f8,u8,i8]:
	l = List.empty_list(t1)
	l.append(0)
	v = l[0]
	for t2 in [f8,u8,i8]:
		d = Dict.empty(t1,t2)
		d[0] = 1
		v = d[0]
		for x in d:
			pass
end = time.time()
print("Time elapsed: %.2f" % (end - start))
```
Takes about 5 secs to run. Note that that is a ton of time since its only doing like 9 total iterations.

But when I just change all of the @njits in typedlist.py and typeddict.py to @njit(cache=True) this same script takes just 0.12 seconds, which is at least a reasonable amount of time to wait for my script to start up.

This bug is pretty frustrating because I'm trying to speed up my code with numba, but there is this annoying sunk cost whenever i want to fill a List which gets worse for every additional type I use. And, yes I am aware that with these particular types at least for the Lists() I could just instead use an np.array() instead. Unfortunately my use case requires a lot of strings, and arrays aren't particularly amenable to strings at the moment.

I'm also wondering if this might have something to do with #4584

stuartarchibald:
Same as #5719 but rebased on mainline, opened to see what the failures are like.


stuartarchibald:
Think https://github.com/numba/numba/issues/6251#issuecomment-693550324 will fix.

eric-wieser:
I assume you'll make a PR with the suggested change? It would be helpful for me for it to exist one way or the other, assuming it works.

stuartarchibald:
> I assume you'll make a PR with the suggested change? It would be helpful for me for it to exist one way or the other, assuming it works.

Any chance you could open the PR please to preserve authorship etc. I can pull the patch in here once submitted. Many thanks.

stuartarchibald:
This needs revisiting some time in the future.

PR #6290, opened 2020-09-30
CUDA: Add support for complex power

### LINKED ISSUE #2495: cmath linking issues in CUDA
From the mailing list:
https://groups.google.com/a/continuum.io/forum/#!topic/numba-users/oBcaFcsD-x4
This:
```python
from numba import cuda
import cmath
import numpy as np

@cuda.jit(device = True)
def quarticRoots(a, b, c, d, e):

	del_0 = c**2 - 3.*b*d + 12.*a*e
	del_1 = 2.*c**3 - 9.*b*c*d + 27*e*b**2 + 27*a*d**2 - 72*a*c*e

	p = (8.*a*c - 3.*b**2)/(8.*a**2)
	q = (b**3 - 4.*a*b*c + 8.*d*a**2)/(8.*a**3)

	Q = ((del_1 + cmath.sqrt(del_1**2 - 4.*del_0**3))/2.)**(1./3.)#**(1./3.) is the problem.
	S = 0.5*cmath.sqrt(-2./3.*p + 1/(3.*a)*(Q + del_0/Q))

	x1 = -b/(4.*a) - S + 0.5*cmath.sqrt(-4.*S**2 - 2.*p + q/S)
	x2 = -b/(4.*a) - S - 0.5*cmath.sqrt(-4.*S**2 - 2.*p + q/S)
	x3 = -b/(4.*a) + S + 0.5*cmath.sqrt(-4.*S**2 - 2.*p - q/S)
	x4 = -b/(4.*a) + S - 0.5*cmath.sqrt(-4.*S**2 - 2.*p - q/S)

	return (x1, x2, x3, x4)

@cuda.jit
def launcher(d_array):
	output = quarticRoots(2.,4.,2.,1.,1.)

	d_array[0] = output[0]
	d_array[1] = output[1]
	d_array[2] = output[2]
	d_array[3] = output[3]

def main():
	d_array = cuda.device_array(4, dtype = np.complex64)
	launcher[1,1](d_array)
	output = d_array.copy_to_host()
	print output

main()
```
yields:
```python
  File "<path>numba/numba/cuda/cudadrv/driver.py", line 1432, in complete
    driver.cuLinkComplete(self.handle, byref(cubin), byref(size))
  File "<path>numba/numba/cuda/cudadrv/driver.py", line 288, in safe_cuda_api_call
  File "<path>numba/numba/cuda/cudadrv/driver.py", line 323, in _check_error
numba.cuda.cudadrv.driver.CudaAPIError: [999] Call to cuLinkComplete results in CUDA_ERROR_UNKNOWN

  File "issue_13.py", line 39, in <module>
    main()
  File "issue_13.py", line 35, in main
    launcher[1,1](d_array)
  File "<path>numba/numba/cuda/compiler.py", line 701, in __call__
    kernel = self.specialize(*args)
  File "<path>numba/numba/cuda/compiler.py", line 712, in specialize
    kernel = self.compile(argtypes)
  File "<path>numba/numba/cuda/compiler.py", line 730, in compile
    kernel.bind()
  File "<path>numba/numba/cuda/compiler.py", line 489, in bind
    self._func.get()
  File "<path>numba/numba/cuda/compiler.py", line 377, in get
    cubin, _size = linker.complete()
  File "<path>numba/numba/cuda/cudadrv/driver.py", line 1434, in complete
numba.cuda.cudadrv.driver.LinkerError: [999] Call to cuLinkComplete results in CUDA_ERROR_UNKNOWN
error   : Undefined reference to 'numba_cpow' in '<cudapy-ptx>'
```
the `cmath.sqrt()` creates a complex domain result with is then raised to the power `1/3`, this ends up as a call to `numba_cpow` which is fine on the CPU but doesn't have an impl in CUDA, as a result there's a linking error.

Two issues:
1. Can the additional cmath function impls for the CPU be done for CUDA.
2. If not can this problem be caught earlier in the complication chain and a better message produced such that is doesn't appear as a link error.

### LINKED ISSUE #3386: Complex power on CUDA
## Feature request

I am using numba do some simulation,however when i try the complex power in cuda,it report error,and i try a very simple coda as below, the code and the error numba suggests are as below, it there something wrong with my code or is just the numba still not support it:

CudaAPIError                              Traceback (most recent call last)
/mnt/F/ubuntu/anaconda/lib/python3.6/site-packages/numba/cuda/cudadrv/driver.py in complete(self)
   1616         try:
-> 1617             driver.cuLinkComplete(self.handle, byref(cubin), byref(size))
   1618         except CudaAPIError as e:

/mnt/F/ubuntu/anaconda/lib/python3.6/site-packages/numba/cuda/cudadrv/driver.py in safe_cuda_api_call(*args)
    289             retcode = libfn(*args)
--> 290             self._check_error(fname, retcode)
    291         return safe_cuda_api_call

/mnt/F/ubuntu/anaconda/lib/python3.6/site-packages/numba/cuda/cudadrv/driver.py in _check_error(self, fname, retcode)
    324                     raise CudaDriverError("CUDA initialized before forking")
--> 325             raise CudaAPIError(retcode, msg)
    326 

CudaAPIError: [999] Call to cuLinkComplete results in CUDA_ERROR_UNKNOWN

LinkerError                               Traceback (most recent call last)
<ipython-input-68-41e1c3856f7c> in <module>()
      1 array1 = np.array([1,2,3,4,5],dtype = np.complex)
----> 2 power_complex[4,4](array1)

/mnt/F/ubuntu/anaconda/lib/python3.6/site-packages/numba/cuda/compiler.py in __call__(self, *args)
    751         Specialize and invoke this kernel with *args*.
    752         '''
--> 753         kernel = self.specialize(*args)
    754         cfg = kernel[self.griddim, self.blockdim, self.stream, self.sharedmem]
    755         cfg(*args)

/mnt/F/ubuntu/anaconda/lib/python3.6/site-packages/numba/cuda/compiler.py in specialize(self, *args)
    762         argtypes = tuple(
    763             [self.typingctx.resolve_argument_type(a) for a in args])
--> 764         kernel = self.compile(argtypes)
    765         return kernel
    766 

/mnt/F/ubuntu/anaconda/lib/python3.6/site-packages/numba/cuda/compiler.py in compile(self, sig)
    781             self.definitions[(cc, argtypes)] = kernel
    782             if self.bind:
--> 783                 kernel.bind()
    784         return kernel
    785 

/mnt/F/ubuntu/anaconda/lib/python3.6/site-packages/numba/cuda/compiler.py in bind(self)
    498         Force binding to current CUDA context
    499         """
--> 500         self._func.get()
    501 
    502     @property

/mnt/F/ubuntu/anaconda/lib/python3.6/site-packages/numba/cuda/compiler.py in get(self)
    383             for path in self.linking:
    384                 linker.add_file_guess_ext(path)
--> 385             cubin, _size = linker.complete()
    386             compile_info = linker.info_log
    387             module = cuctx.create_module_image(cubin)

/mnt/F/ubuntu/anaconda/lib/python3.6/site-packages/numba/cuda/cudadrv/driver.py in complete(self)
   1617             driver.cuLinkComplete(self.handle, byref(cubin), byref(size))
   1618         except CudaAPIError as e:
-> 1619             raise LinkerError("%s\n%s" % (e, self.error_log))
   1620 
   1621         size = size.value

LinkerError: [999] Call to cuLinkComplete results in CUDA_ERROR_UNKNOWN
error   : Undefined reference to 'numba_cpow' in '<cudapy-ptx>'

### LINKED ISSUE #5727: complex power fails with cuda
## Reporting a bug

When I run the following code, I expect to get the output `20+2j`:

```python
@guvectorize([(complex128[:], complex128[:])], '()->()', target='cuda')
def f(val, out):
    out = val[0] ** val[0]

f(4+2j)
```

Instead, a `numba.cuda.cudadrv.driver.LinkerError` is raised with the message `Undefined reference to 'numba_cpow' in '<cudapy-ptx>'`.

gmarkall:
This commit adds support for complex power using the `**`and `**`
inplace operators, and via the builtin `pow` function. The
implementation in `libdevice.py` is a translation of the logic from
`_Py_c_pow` in CPython 3.7.5.

Although `libdevice.py` seems a strange place to put this, it is more
that `libdevice.py` is strangely named - it is where the implementations
of most of the CUDA maths functions are. Note that PR #6152 moves all
these into a more appropriateley named `mathimpl.py`.

Lowering is added for `complex64` and `complex128` separately - the
`complex64` version carefully casts everything to ensure that single
precision arithmetic is used.

There is only a minor documentation update here, because the
documentation seemed to imply already that this was already supported -
however, support for the builtin function `pow` was not listed.

New tests are added that test with finite values. I did experiment with
testing with infs, nans, etc., but could find very few cases that
wouldn't raise a `RuntimeWarning` about invalid values when executing
the Python versions of the functions for comparison.

Fixes #5727.
Fixes #3386.
Fixes #2495.


gmarkall:
The failure is just a conda HTTP error - @stuartarchibald could you re-run the tests please?

stuartarchibald:
> The failure is just a conda HTTP error - @stuartarchibald could you re-run the tests please?

done!

numba/cuda/libdevice.py
```diff
@@ -165,3 +167,58 @@ def core(context, builder, sig, args):
 for (ty, intrin) in ((types.float64, '__nv_modf',),
                      (types.float32, '__nv_modff',)):
     lower(math.modf, ty)(modf_implement(intrin, ty))
+
+
+# Complex power implementations - translations of _Py_c_pow from CPython
+# 3.7.5.
```
stuartarchibald, line 173: Please add an exact reference with SHA to the implementation. Example:
https://github.com/numba/numba/blob/626b40eddc9a29a1e89bfb6b8706e4f390cfcc4a/numba/cpython/hashing.py#L30-L31

numba/cuda/libdevice.py
```diff
@@ -165,3 +167,58 @@ def core(context, builder, sig, args):
...
+        len = math.pow(vabs, b.real)
+        at = math.atan2(a.imag, a.real)
+        phase = at * b.real
+        if b.imag != 0.0:
+            len /= math.exp(at * b.imag)
+            phase += b.imag * math.log(vabs)
+
+        return len * (math.cos(phase) + math.sin(phase) * 1.0j)
+
+    return context.compile_internal(builder, cuda_cpow, sig, args)
```
stuartarchibald, line 224: Is it worth generating the impls by closing over the key types, as they are the same apart from the casts?

numba/cuda/tests/cudapy/test_powi.py
```diff
@@ -24,6 +25,31 @@ def cu_mat_power_binop(A, power, power_A):
...
+
+def vec_pow_inplace_binop(r, x):
+    i = cuda.grid(1)
+
+    if i < len(r):
+        r[i] **= x[i]
+
+
+def random_complex(N):
+    return (np.random.random(1) + np.random.random(1) * 1j)
```
stuartarchibald, line 50: For reproducibility, probably a good idea to set the RNG seed?

numba/cuda/tests/cudapy/test_powi.py
```diff
@@ -45,6 +71,54 @@ def test_powi_binop(self):
         kernel[1, A.shape](A, power, Aout)
         self.assertTrue(np.allclose(Aout, A ** power))
 
+    # Relative tolerance kwarg is provided because 1.0e-7 (the default for
+    # assert_allclose) is a bit tight for single precision.
```
stuartarchibald, line 75: Does this end up with 64bit floating point, by default, using a far slacker tolerance than would normally be expected of it?

stuartarchibald (commented):
Thanks for the patch, few minor things to look at else looks good. Great to see this bug fixed!

numba/cuda/libdevice.py
gmarkall: Yes, that makes sense.

numba/cuda/tests/cudapy/test_powi.py
gmarkall: The default here is the same as the default for `assert_allclose`, which is `1.0e-7` - it's only for single precision that it's overridden with `3.0e-7`. So if I've understood your question correctly, I think the answer is "no".

numba/cuda/tests/cudapy/test_powi.py
gmarkall: Done!

gmarkall:
I've merged in master - I think everything should be resolved now.

numba/cuda/tests/cudapy/test_powi.py
stuartarchibald: Thanks for the explanation. Numba's CPU impl tends to use `self.assertPreciseEqual` for testing fundamental maths functions, should the CUDA target do similar?

stuartarchibald:
> I've merged in master - I think everything should be resolved now.

Thanks, was also wondering about https://github.com/numba/numba/pull/6290#discussion_r500299691 ?

stuartarchibald (approved):

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_109`.

numba/cuda/tests/cudapy/test_powi.py
gmarkall: Maybe. At the moment most maths functions in the CUDA tests are tested to the rather loose relative tolerance of 1.0e-5. I think it would take some time going through them and determining to what extent there are differences between it and the CPU target / libc implementations to fix up in general, but I'll check if `assertPreciseEqual` works here.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_109`.

passed

joejoseph007:
Has this been actually solved?
I'm trying to run this 
`def get_grid_cuda(Z):
    i, j = cuda.grid(2)
    if i >=0 and i < Z.shape[0] and j >=0 and j < Z.shape[1]:     
        for i in range(100):
            fx=Z[i,j]**3+Z[i,j]**2+Z[i,j]+1
            dfx=3*Z[i,j]**2+2*Z[i,j]+1
            div=fx/dfx
            Z[i,j]=Z[i,j]-div
`
**Z** is just a grid on the complex plane

gmarkall:
@joejoseph007 Could you please open an issue or start a post on discourse (https://numba.discourse.group) with a runnable version of the code (something that can be copy/pasted and executed to reproduce the issue) and a description of the problem?

joejoseph007:
Sure

PR #6292, opened 2020-09-30
Document support for structured array attribute access

### LINKED ISSUE #878: Numba semantics for structured arrays differs to Numpy semantics for structured arrays
Numba allows access to fields of individual elements in structured arrays by attribute, whereas Numpy does not. Consider the following as an example:

``` python
from numba import jit, numpy_support
import numpy as np

record_type = np.dtype([('ival', np.int32),
                        ('fval', np.float64)],
                        align=True)

nb_record_type = numpy_support.from_dtype(record_type)

def f(rec):
    value = 2.5
    rec[0].ival = int(value)
    rec[0].fval = value
    return rec

arr = np.ones(1, dtype=record_type)

cfunc = jit(nb_record_type[:](nb_record_type[:]), nopython=True)(f)

# Works
print(cfunc(arr))

# Does not work
print(f(arr))
```

The call to `f(arr)` produces the following traceback:

```
$ python semantics.py 
  File "semantics.py", line 24, in <module>
    print(f(arr))
  File "semantics.py", line 12, in f
    rec[0].ival = int(value)
AttributeError: 'numpy.void' object has no attribute 'ival'
```

It may be that Numpy does not support this because it would slow down field accesses, as it does with `recarray` instances, mentioned in the first section of this document: http://wiki.scipy.org/Cookbook/Recarray

However, we have no performance penalty for access in this way, since the `getattr` operation is compiled down to accessing the array element plus some offset.

The question is, do we want to continue to support this, when Numpy does not support it? in my opinion, it is convenient for a user so it would be nice to support it.

It is also worth noting that the dispatcher presently takes two different routes for finding the typecodes of recarrays and structured arrays - structured arrays can use `typecode_ndarray` which has caching of types, whereas recarrays always have to go through the slower `typecode_fallback`at the moment (though it may be possible to cache recarray types on the C side to speed things up - I have not yet looked into this).

### LINKED ISSUE #5375: Deprecate record access by attr in CUDA target
In the CUDA target it is possible to access record items as attrs, for example:

```python
point_type = [('x', np.float64), ('y', np.float64)]
points = np.recarray(1, dtype=point_type)

@cuda.jit
def move_to_origin(point):
    point.x = 0
    point.y = 0

move_to_origin[1, 1](points[0])
```

This is not legal in normal Python / NumPy - removing the `@cuda.jit` would require the following syntactic change in `move_to_origin`:

```python
def move_to_origin(point):
    point['x'] = 0
    point['y'] = 0
```

The former syntax should be deprecated, for consistency with non-cuda-jitted Python code. Any code impacted by the deprecation can be fixed up by substituting the former syntax for the latter.

gmarkall:
The support for attribute access of structured arrays can be surprising,
as it is not supported by NumPy. This commit documents this behavior, as
an alternative to removing support for attribute access, which is
presently considered non-trivial (see discussion on Issue #5375).

Fixes #878.
Does not fix #5375


stuartarchibald (approved):
Thanks for the patch, good to have this mismatch documented!

stuartarchibald:
I've updated the original comment, I don't think this should close #5375? It can act as a reminder that this behavioural mismatch still exists.

gmarkall:
> I've updated the original comment, I don't think this should close #5375? It can act as a reminder that this behavioural mismatch still exists.

Yeah, I agree that it's better not to close it, since the docs do state it will eventually be deprecated.

gmarkall:
I've just merged in the fix for the flake issue introduced by another merge. The other fail was a conda HTTP error.

stuartarchibald (approved):
Thanks for resolving conflicts etc. Change is still approved.

luk-f-a:
@gmarkall what about `numpy.recarray`? Numba has only one type for both structured array and recarray, and the attribute access makes more sense in that context.

This is your example adjusted to use `recarray`

```python
from numba import njit
import numpy as np
record_type = np.dtype([("ival", np.int32), ("fval", np.float64)], align=True)
def f(rec):
   value = 2.5
   rec[0].ival = int(value)
   rec[0].fval = value
   return rec
arr = np.recarray((1,), dtype=record_type)
cfunc = njit(f)
# Works
print(cfunc(arr))
# works too
print(f(arr))
```

gmarkall:
IIUC, I think the behaviour is consistent between Numba and NumPy for recarrays. Is that what you are referring to?

luk-f-a:
yes. the documentation now says that the behaviour will be deprecated, which would replace one numba--numpy mismatch with another. Unless the types are split and attribute array kept for recarrays but not for struct arrays.
do you think the current behaviour causes problems? Otherwise I'd suggest keeping it.

gmarkall:
As far as I can see the documentation only talks about structured arrays, not recarrays (and to me implies that it will be deprecated for structured arrays only) - can you show me what I'm missing please?

luk-f-a:
I might be inferring too much, in that case I apologize. Where I'm coming from is that at the moment there's one type and one documentation for both objects. Deprecating something for one of them would do it for the other.
Are you saying that when the time comes to remove attribute access, a separate type (and documentation) would be added?

gmarkall:
Yes, I would add a separate type and documentation at deprecation time - because it's a bit more complicated to type them differently than I had time for, I wrote this documentation instead,

luk-f-a:
thanks for the clarification @gmarkall

PR #6294, opened 2020-10-01
Make return type propagate to all return variables

### LINKED ISSUE #6293: Typer does not propagate return type to all return variables

When there are multiple return statements in a function the typer does not propagate the unified return type to all return variables. This often works because return values are implicitly casted to the return type [here](https://github.com/numba/numba/blob/626b40eddc9a29a1e89bfb6b8706e4f390cfcc4a/numba/core/lowering.py#L357). However, it does make a difference for memory managed types (the implicit cast does not neccessarily get assigned hence a borrowed reference can escape).

In the example below the first return variable is of type `int64`, the second is `float64`. IMHO both should have been explicitly casted to `float64`.

```python
from numba import jit, types
from numba.core import ir


@jit(nopython=True)
def confuse_typer(x):
    # --- LINE 16 ---
    # label 0
    #   x = arg(0, name=x)  :: float64
    #   $6compare_op.2 = x == x  :: bool
    #   bool8 = global(bool: <class 'bool'>)  :: Function(<class 'bool'>)
    #   $8pred = call bool8($6compare_op.2, func=bool8, args=(Var($6compare_op.2, cast_quantity_to_adf_crash.py:5),), kws=(), vararg=None)  :: (bool,) -> bool
    #   del bool8
    #   del $6compare_op.2
    #   branch $8pred, 10, 18
    if x == x:
        # --- LINE 27 ---
        # label 10
        #   del $8pred
        #   $10load_global.0 = global(int: <class 'int'>)  :: Function(<class 'int'>)
        #   $14call_function.2 = call $10load_global.0(x, func=$10load_global.0, args=[Var(x, cast_quantity_to_adf_crash.py:5)], kws=(), vararg=None)  :: (float64,) -> int64
        #   del x
        #   del $10load_global.0
        #   $16return_value.3 = cast(value=$14call_function.2)  :: int64    <- cast to float64 should happen here
        #   del $14call_function.2
        #   return $16return_value.3    <- returns int64
        return int(x)
    # --- LINE 29 ---
    else:
        # --- LINE 36 ---
        # label 18
        #   del $8pred
        #   $20return_value.1 = cast(value=x)  :: float64
        #   del x
        #   return $20return_value.1    <- returns float64
        return x

confuse_typer.compile((types.float64,))
cres = confuse_typer.overloads[(types.float64,)]
typemap = cres.type_annotation.typemap
return_vars = {}


for block in cres.type_annotation.blocks.values():
    for inst in block.body:
        if isinstance(inst, ir.Return):
            varname = inst.value.name
            return_vars[varname] = typemap[varname]

# fails b/c return_vars = {'$16return_value.3': types.int64, '$20return_value.1': types.float64}
assert all(vt == types.float64 for vt in return_vars.values())
```

asodeur:
fixes #6293 

- [x] add test
- [x] propagate return type in `typeinfer.TypeInferer.unify
- [x] check if implicit casting for return values can now be removed from `lowering.py`

The last commit (b13ea4a) is removing the current implicit casting behaviour from `lowering.Lower` as this path should no longer be used (and the testsuite seems to prove this). A less risky approach might be to issue a deprecation warning instead.


stuartarchibald:
Thanks for the PR. It's queued for review.

numba/core/typeinfer.py
```diff
@@ -1207,6 +1207,8 @@ def check_var(name):
 
         try:
             retty = self.get_return_type(typdict)
+            typdict = utils.UniqueDict(
+                typdict, **{v.name: retty for v in self._get_return_vars()})
```
sklam, line 1211: Don't think this is needed in the try...except. Can you move this out in the `else` part for the `try...except...else`?

sklam (commented):
Thanks for fixing this bug. I have some suggestions:

stuartarchibald (approved):
Thanks for the fixes, looks like all the comments were addressed, patch looks good.

PR #6296, opened 2020-10-02
Fix flake8 violations in numba.cuda module

gmarkall:


stuartarchibald (approved):
Thanks for the patch.

stuartarchibald:
See https://github.com/numba/numba/pull/6303 for buildfarm pass.

PR #6305, opened 2020-10-05
Fixes #5871. Segfault when calling pyapi.call() passing args as None

### LINKED ISSUE #5871: Bug on pyapi.call when args is None
tldr: Calling `pyapi.call(obj, args=None, kws)` will result in a segfault.

Code to reproduce:

```python
from numba import njit
from numba.core import types, cgutils, extending
from numba.core.datamodel import registry, models
from llvmlite import ir

class Dense():
    def __init__(self, *, N):
        self.N = N

class DenseType(types.Type):
    def __init__(self, *, N_type):
        self.N_type: int = N_type
        name = f"Dense<{N_type}>"
        super().__init__(name)

    # Type is mutable
    mutable = True

    @property
    def key(self):
        return self.N_type


@registry.register_default(DenseType)
class DenseModel(models.StructModel):
    def __init__(self, dmm, fe_type: DenseType):
        members = [
            ("N", fe_type.N_type),
        ]
        super().__init__(dmm, fe_type, members)


@extending.type_callable(Dense)
def type_dense(context):
    def typer(N):
        return DenseType(N_type=N)
    return typer


@extending.lower_builtin(Dense, types.Any)
def sparse_dense_constructor(context, builder, sig, args):
    [N] = args
    dense = cgutils.create_struct_proxy(sig.return_type)(context, builder)
    dense.N = N
    return dense._getvalue()

@extending.box(DenseType)
def box_dense(typ: DenseType, val, c):
    """
    Convert a native dense structure to a Dense object.
    """
    i1 = ir.IntType(1)

    dense = cgutils.create_struct_proxy(typ)(c.context, c.builder, value=val)

    N_obj = c.pyapi.long_from_long(dense.N)
    class_obj = c.pyapi.unserialize(c.pyapi.serialize_object(Dense))

    kwds = c.pyapi.dict_pack({"N": N_obj}.items())
    res = c.pyapi.call(class_obj, None, kwds)  # segfault

    # empty_tuple = c.pyapi.tuple_new(0)
    # res = c.pyapi.call(class_obj, empty_tuple, kwds)  # works
    # c.pyapi.decref(empty_tuple)

    c.pyapi.decref(N_obj)
    c.pyapi.decref(class_obj)
    c.pyapi.decref(kwds)
    return res

@njit
def foo():
    d = Dense(N=3)
    return d

print(foo())
```

The interesting part happens inside `extending.box(DenseType)` in which uses `c.pyapi.call(class_obj, None, kwds)` to call the dense constructor. Note that the constructor is a keyword-only function. Inside the `c.pyapi.call`, if args is `None`, args is assigned to the return value of `PythonApi.get_null_object()` and lately calls `PyObject_call` function from CPython. The problem with this is that the CPython docs explicit says that `args` must not be `NULL`:

> Call a callable Python object callable, with arguments given by the tuple args, and named arguments given by the dictionary kwargs.
>
>args must not be NULL, use an empty tuple if no arguments are needed. If no named arguments are needed, kwargs can be NULL.

If this is really an issue, I think is just a matter of replacing `args = self.get_null_object()` by `self.tuple_new(0)` to solve this problem.

marceloFA:
As stated in the discussion on issue #5871 the fix was just a simple matter of replacing ``args = self.get_null_object()`` for ``args = self.tuple_new(0)`` at ``numba/core/pythonapi.py``.

If you will, could you add ``hacktoberfest-accepted`` to the labels of this PR so it can count for me as a valid contribution to the festival?

### DIFF
--- a/numba/core/pythonapi.py
+++ b/numba/core/pythonapi.py
@@ -942,7 +942,7 @@ class PythonAPI(object):
 
     def call(self, callee, args=None, kws=None):
         if args is None:
-            args = self.get_null_object()
+            args = self.tuple_new(0)
         if kws is None:
             kws = self.get_null_object()
         fnty = Type.function(self.pyobj, [self.pyobj] * 3)


stuartarchibald:
Thanks for the patch. Given this is fixing a segfault I think the reproducer should be included as a test, probably in a new test module e.g. `numba.tests.test_pyapi.py`.

eric-wieser (changes requested):
This needs to decref the resulting tuple after the call, doesn't it?

stuartarchibald:
> This needs to decref the resulting tuple after the call, doesn't it?

Yes, I think so.

PR #6306, opened 2020-10-06
Fix flake8 in cuda atomic test from merge.

stuartarchibald:
As title.


esc (approved):
I checked this locally with `flake8` -- `master` is broken and this fixes it!

stuartarchibald:
> I checked this locally with `flake8` -- `master` is broken and this fixes it!

Thanks @esc merging.

PR #6307, opened 2020-10-06
Remove restrictions on SVML version due to bug in LLVM SVML CC

stuartarchibald:
As title. For context see https://github.com/numba/numba/issues/6100

### DIFF
--- a/buildscripts/condarecipe.local/meta.yaml
+++ b/buildscripts/condarecipe.local/meta.yaml
@@ -52,8 +52,7 @@ requirements:
     - cudatoolkit >=8.0
     # scipy 1.0 or later
     - scipy >=1.0
-    # icc_rt > 2018.0.2, see https://github.com/numba/numba/issues/6100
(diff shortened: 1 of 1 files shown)


stuartarchibald:
xref https://github.com/numba/llvmlite/pull/636 https://github.com/numba/llvmlite/pull/626

esc:
I presume this will need to be tested, once an llvmlite "version with new SVML patch" becomes available?

stuartarchibald:
> I presume this will need to be tested, once an llvmlite "version with new SVML patch" becomes available?

Yes. Ideally we'd sort out an LLVM IR reproducer for the SVML patch too.

esc:
`numba/llvmlite#626` was merged - I don't think anything else is needed for this. Assuming that llvmlite 0.35.0 will ship with the correct  SVML patch we can merge this and then Numba 0.52.0 will no longer have this restriction.

PR #6319, opened 2020-10-07
Permit switching off boundschecking when debug is on.

stuartarchibald:
Dispatcher kwarg `debug=True` switches on both debuginfo and
boundschecking, its reasonable to just want the debuginfo part.
Also setting `debug=True` meant the value of the `boundscheck`
kwarg was just ignored. This fixes it and fixes the gdb docs to
note this.

### DIFF
--- a/numba/core/decorators.py
+++ b/numba/core/decorators.py
@@ -24,7 +24,7 @@ _msg_deprecated_signature_arg = ("Deprecated keyword argument `{0}`. "
                                  "positional argument.")
 
 def jit(signature_or_function=None, locals={}, cache=False,
-        pipeline_class=None, boundscheck=False, **options):
+        pipeline_class=None, boundscheck=None, **options):
     """
     This decorator is used to compile a Python function into native code.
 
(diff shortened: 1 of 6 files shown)


stuartarchibald:
xref https://stackoverflow.com/questions/64227615/debugging-with-numba/

numba/core/options.py
```diff
@@ -43,16 +43,19 @@ def set_flags(self, flags):
...
             flags.set("nrt")
 
         if kws.pop('debug', config.DEBUGINFO_DEFAULT):
             flags.set("debuginfo")
             flags.set("boundscheck")
 
+        # This is checked for logically after 'debug' as it's possible to want
+        # to have debuginfo on from the kwarg debug=True but to have boundscheck
+        # off.
+        if 'boundscheck' in kws:
```
sklam, line 56: `boundscheck` is defaulted to `False`. 
https://github.com/numba/numba/blob/9381206a03a3e7231d417f147474577745f6a720/numba/core/decorators.py#L27
This will always disable `boundscheck`

numba/core/options.py
```diff
@@ -43,16 +43,23 @@ def set_flags(self, flags):
...
 
-        if kws.pop('boundscheck', False):
-            flags.set("boundscheck")
-
         if kws.pop('_nrt', True):
             flags.set("nrt")
 
-        if kws.pop('debug', config.DEBUGINFO_DEFAULT):
+        debug_mode = kws.pop('debug', config.DEBUGINFO_DEFAULT)
+        bounds_check_explicit = 'boundscheck' in kws
```
sklam, line 50: `boundscheck` is always set in https://github.com/numba/numba/blob/0ac11479142ecbf74bbf2ca99d89be9e1d9a1f36/numba/core/decorators.py#L26-L27

sklam: consider these tests:

https://gist.github.com/sklam/23303f2413266e2e5238cacaf5669f99

esc:
Yes! I have been waiting for this one! It'll improve the usability of the enhance cfg patch.

numba/core/decorators.py
```diff
@@ -24,7 +24,7 @@
                                  "positional argument.")
 
 def jit(signature_or_function=None, locals={}, cache=False,
-        pipeline_class=None, boundscheck=False, **options):
+        pipeline_class=None, boundscheck=None, **options):
```
stuartarchibald, line 27: This has ended up being not particularly neat. The value for "boundscheck" is transmitted via `targetoptions` and actually originates here, and is actually captured in this function:
https://github.com/numba/numba/blob/1041fa6ee8430471da99b54b3428a673033e7e44/numba/core/decorators.py#L154
it then propagates.

To permit changing behaviour when `debug` is set a 3-state is needed, `True/False` as potentially set by a user and `None` for "not set, let it be influenced by `debug`".

sklam (approved):
The changes look good and I have manually verified it has the expected behavior.

stuartarchibald:
@sklam had to do a033497 to fix the build, some CUDA sim problems.

sklam (approved):
Latest fix looks good. Azure agrees as well.

PR #6325, opened 2020-10-08
Refactor code for atomic operations

testhound:
Used code generation suggested by @stuartarchibald here: 

https://github.com/numba/numba/commit/9e54c7d78a2d35379724e9e8dfb1b225fd887d96

this pull request refactors many of the atomic operations and reduces code duplication

### DIFF
--- a/numba/cuda/cudadecl.py
+++ b/numba/cuda/cudadecl.py
@@ -307,6 +307,7 @@ Cuda_atomic_min = _gen(cuda.atomic.min, all_numba_types)
 Cuda_atomic_nanmax = _gen(cuda.atomic.nanmax, all_numba_types)
 Cuda_atomic_nanmin = _gen(cuda.atomic.nanmin, all_numba_types)
 
+
 @register
 class Cuda_atomic_compare_and_swap(AbstractTemplate):
     key = cuda.atomic.compare_and_swap
--- a/numba/cuda/cudadecl.py
+++ b/numba/cuda/cudadecl.py
@@ -276,72 +276,36 @@ class Cuda_selp(AbstractTemplate):
         return signature(a, test, a, a)
 
 
-@register
-class Cuda_atomic_add(AbstractTemplate):
-    key = cuda.atomic.add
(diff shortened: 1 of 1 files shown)


numba/cuda/cudadecl.py
```diff
@@ -276,71 +276,36 @@ def generic(self, args, kws):
...
+            if ary.ndim == 1:
+                return signature(ary.dtype, ary, types.intp, ary.dtype)
+            elif ary.ndim > 1:
+                return signature(ary.dtype, ary, idx, ary.dtype)
+    return Cuda_atomic
+
+
+all_numba_types = (types.float64, types.float32,
+                   types.int32, types.uint32,
+                   types.int64, types.uint64)
```
stuartarchibald, line 301: This is applying additional type restriction to the types accepted by `atomic.add` and `atomic.sub`, I think this is probably a good thing as e.g.:
```python
from numba import cuda
import numpy as np

@cuda.jit
def foo(result, values):
    tid = cuda.threadIdx.x
    bid = cuda.blockIdx.x
    bdim = cuda.blockDim.x
    i = (bid * bdim) + tid
    cuda.atomic.add(result, 0, values[i])


arr = np.ones(1, dtype=np.int8)
result = np.ones(1, dtype=np.int8)

foo[1, 1](result, arr)
print(result[0])
```
on mainline results in:
```

(0) Error: unsupported operation

NVVM_ERROR_COMPILATION
```
this also matches up with the lowering available.

CC @gmarkall

stuartarchibald (commented):
Thanks for the patch, refactor looks good, one query, CC'd @gmarkall for input.

numba/cuda/cudadecl.py
gmarkall: I agree, this is a good thing.

gmarkall (approved):
Looks good to me, and testsuite passes locally for me.

stuartarchibald (approved):
Thanks for checking @gmarkall, patch approved!

Issue #6330, opened 2020-10-08
CUDA Simulator not working


nicolapivaro:
Hi guys! I have some issue with the simulator.
I'm trying to use the example you show in your documentation:
https://numba.pydata.org/numba-doc/dev/cuda/simulator.html

`import numpy as np

    @cuda.jit(debug=True)
    def vec_add(A, B, out):
        x = cuda.threadIdx.x
        bx = cuda.blockIdx.x
        bdx = cuda.blockDim.x
        if x == 1 and bx == 3:
            from pdb import set_trace;
            set_trace()
        i = bx * bdx + x
        out[i] = A[i] + B[i]

    os.environ["NUMBA_ENABLE_CUDASIM"] = "1"
    A = np.ones((4, 4))
    B = np.ones((4, 4))
    C = np.zeros((4, 4))
    vec_add(A,B,C)`

This the code I'm running and it tells me I can not import inside the function, like if it is still compiling.
"Use of unsupported opcode (IMPORT_NAME) found"
Am I doing something wrong or is not working?
Thank you
Nicola

gmarkall:
You need to set the environment variable `NUMBA_ENABLE_CUDASIM` before you import Numba.

gmarkall:
There are a couple of other issues with your code:

- The inputs `A`, `B`, and `out` are 2D, so the indexing in the last line `out[i] = A[i] + B[i]` adds entire rows - only scalar operations are supported on the CUDA target, so you need to index them like `out[i, j] = A[i, j] + B[i, j]`. To generate the `j` index, you could use a for loop or a 2D grid. Or you could pass in 1D inputs instead.
- There is no kernel launch configuration in `vec_add(A,B,C)` - you need to specify the grid and block dimensions like `vec_add[griddim, blockdim](A,B,C)`.

I've modified the example so that it runs:

```python
import numpy as np
import os
os.environ["NUMBA_ENABLE_CUDASIM"] = "1"

from numba import cuda  # noqa: E402


@cuda.jit(debug=True)
def vec_add(A, B, out):
    x = cuda.threadIdx.x
    bx = cuda.blockIdx.x
    bdx = cuda.blockDim.x
    if x == 1 and bx == 3:
        from pdb import set_trace

        set_trace()
    i = bx * bdx + x
    out[i] = A[i] + B[i]


A = np.ones((16))
B = np.ones((16))
C = np.zeros((16))
vec_add[4, 4](A, B, C)
```

[...]

PR #6331, opened 2020-10-08
Explicitly state that NUMBA_ENABLE_CUDASIM needs to be set before import

gmarkall:
See issue #6330 for the motivation for this change - it is not obvious to all users that environment variables need to be set before Numba is imported.


stuartarchibald (approved):
Agree, makes sense to be explicit about when this has to happen as it's stateful. Thanks for the patch.

PR #6333, opened 2020-10-09
Remove dead _Kernel.__call__

gmarkall:
This is a fairly minor removal of dead code, but I keep coming across `_Kernel.__call__` when tracing through from a call to a kernel launch and getting confused, so I'm motivated to remove it.


gmarkall:
> when tracing through from a call to a kernel launch

For clarification - this is when I'm manually tracing what happens though the source code, this never appears in a backtrace because it's dead code.

stuartarchibald (approved):
Thanks for the patch `_kernel_call` doesn't exist so this path must be dead.

PR #6334, opened 2020-10-09
Support attribute access from other modules

farahhariri:
Accessing an attribute from another module currently fails if it has a complex data type like string. This PR passes the builder object to the `get_getattr` function and uses `get_constant_generic` instead of `get_constant` to retrieve attributes. A test case that shows an example that would usually fail and this PR fixes is added


esc:
@farahhariri thanks for submitting this! Perhaps you could provide a little more context as to why this is needed or what it fixes? Thanks!

farahhariri:
> @farahhariri thanks for submitting this! Perhaps you could provide a little more context as to why this is needed or what it fixes? Thanks!

Hi @esc This is needed for when someone accesses an attribute (like a constant tuple) in another module. I will add a test case to it to show what it addresses

esc:
@farahhariri excellent, test cases are always welcome.

numba/tests/test_objects.py
```diff
@@ -30,6 +29,11 @@ def delattr_usecase(o):
 
 
 class TestAttributes(TestCase):
+    def test_getattr(self, flags=enable_pyobj_flags):
+        import numba.tests.test_analysis as test_analysis
+        def f():
+            return test_analysis._GLOBAL
```
ehsantn, line 35: I think the global needs to be a type like string that would need the builder for lowering.

ehsantn (approved):
Looks good to me. Thanks.

sklam (approved):
Thanks for the contribution.

stuartarchibald:
@farahhariri Congratulations on your first contribution to Numba!

farahhariri:
> @farahhariri Congratulations on your first contribution to Numba!

Thanks @stuartarchibald :)

PR #6337, opened 2020-10-09
Implements `key` on `types.TypeRef`

### LINKED ISSUE #6336: `TypeRef` should implement `key`

`types.abstract.TypeRef` should implement `key`, otherwise it can happen that `aty != bty` but `TypeRef(aty) == TypeRef(bty)` when `aty.name == bty.name and aty.key != bty.key`.

Self contained example:
```python
from numba import types

class NoUniqueNameType(types.Dummy):
    def __init__(self, param):
        super(NoUniqueNameType, self).__init__('NoUniqueNameType')
        self.param = param

    @property
    def key(self):
        return self.param

no_unique_name_type_1 = NoUniqueNameType(1)
no_unique_name_type_2 = NoUniqueNameType(2)

assert no_unique_name_type_1  != no_unique_name_type_2
assert types.TypeRef(no_unique_name_type_1  ) != types.TypeRef(no_unique_name_type_2) 
```

asodeur:
fixes #6336

### DIFF
--- a/numba/core/types/abstract.py
+++ b/numba/core/types/abstract.py
@@ -464,6 +464,10 @@ class TypeRef(Dummy):
         self.instance_type = instance_type
         super(TypeRef, self).__init__('typeref[{}]'.format(self.instance_type))
 
+    @property
+    def key(self):
+        return self.instance_type
+
 
 class InitialValue(object):
     """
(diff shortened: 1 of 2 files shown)


stuartarchibald (approved):
Thanks for the patch and for fixing this. Looks good.

PR #6342, opened 2020-10-09
Fix docs on literally usage.

### LINKED ISSUE #6238: literally does not behave like in the doc example (and the test)
I'm trying to replicate the example from https://www.github.com/numba/numba/blob/master/numba/tests/test_literal_dispatch.py .

On my computer (numba 0.51.2), it doesn't print the desired output, meaning that function specializations on literal types are not computed.  (incidentally, the test does not check for correct behaviour: might be better if specialized functions returned different output)

In the changelog, I read for version 0.50: "Numba has internally switched to prefer non-literal types over literal ones so  as to reduce function over-specialization, this with view of speeding up compile times (Siu Kwan Lam)." Indeed, with version 0.50, it works as expected.

Has the functionality completely disappeared or changed ? Is there a way to reenable it ?

### LINKED ISSUE #6274: Changes to literal behaviour between <=0.49.0 and >=0.50.0
## Reporting a bug

## Reproducer
```python
from numba.extending import overload, register_jitable
from numba import jit, literally


@jit(nopython=True, fastmath=True, parallel=False, cache=True, nogil=True)
def f(mode):
    return _f(literally(mode))


def _f(mode):
    return


@overload(_f, inline='always')
def _f_impl(mode):

    if mode.literal_value == "a":
        return impl_a
    elif mode.literal_value == "b":
        return impl_b
    else:
        raise ValueError("Unsupported mode!")


@register_jitable
def impl_a(mode):
    print("Using impl_a!")
    return sub_func(literally(mode))


@register_jitable
def impl_b(mode):
    print("Using impl_b!")
    return sub_func(literally(mode))


def sub_func(mode):
    return


@overload(sub_func, inline='always')
def _sub_func(mode):
    if mode.literal_value == "a":
        return lambda mode: 0
    else:
        return lambda mode: 1


if __name__ == "__main__":

    print(f("a"))
    print(f("b"))
```

## Problem
The above fails when using `numba>=0.50.0` with `'NoneType' object is not callable`. This example runs without issue with `numba<=0.49.0`. I believe the issue stems from literally'ing an already literally'ed value - removing the literally calls from the `sub_func` invocations makes this run with `numba>=0.50.0`.

## Question
Will this be the expected behaviour going forward and is the above pattern acceptable?

stuartarchibald:
As title.

Fixes #6238
Closes #6274

### DIFF
--- a/numba/tests/doc_examples/test_literally_usage.py
+++ b/numba/tests/doc_examples/test_literally_usage.py
@@ -7,7 +7,7 @@ from numba.tests.support import captured_stdout
 class DocsLiterallyUsageTest(unittest.TestCase):
 
     def test_literally_usage(self):
-        with captured_stdout():
+        with captured_stdout() as stdout:
             # magictoken.ex_literally_usage.begin
             import numba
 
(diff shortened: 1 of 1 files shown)


sklam (approved):
LGTM

PR #6343, opened 2020-10-09
CUDA: Add support for passing tuples and namedtuples to kernels

### LINKED ISSUE #4405: CUDA functions can be compiled but not called with Tuple and UniTuple arguments
As the title suggests, functions with `Tuple` or  `UniTuple` arguments can be compiled just fine, but are unable to actually be called without raising an exception. Here's an example using `UniTuple`:

```python
import numpy as np
from numba import cuda, types

n_array = 5
@cuda.jit(types.void(types.UniTuple(types.float32[:], n_array), types.float32[:]))
def test(arrays, out):
	i = cuda.threadIdx.x
	tmp = 0
	for arr in arrays:
		tmp += arr[i]
	out[i] = tmp

def main():
	n_elements = 3
	arrays_in = [k * np.ones(n_elements, np.float32) for k in range(n_array)]
	array_out = np.empty(n_elements, np.float32)
	test[1, n_elements](arrays_in, array_out)
	print("output:", array_out)

if __name__=="__main__": main()
```
I have no real knowledge about the internals of Numba,
but I found adding the following lines to `_prepare_args` of the `CUDAKernel` class:

```python
...
elif isinstance(ty, types.UniTuple) or isinstance(ty, types.Tuple):
    assert len(ty)==len(val)
    for t, v in zip(ty, val):
       self._prepare_args(t, v, stream, retr, kernelargs)
...
```

seems to fix the minimal example I provided.
I am using the latest version of Numba.

gmarkall:
Fixes #4405.

This adds support for tuples as per @heyer2's suggestion, and namedtuples with @atait's suggestion. Tests for passing each, and an empty tuple are added.

This passes tests for me locally with the hardware target and cudasim.

### DIFF
--- a/numba/cuda/compiler.py
+++ b/numba/cuda/compiler.py
@@ -756,6 +756,12 @@ class _Kernel(serialize.ReduceMixin):
             devrec = wrap_arg(val).to_device(retr, stream)
             kernelargs.append(devrec)
 
+        elif isinstance(ty, (types.UniTuple, types.Tuple, types.NamedUniTuple,
+                             types.NamedTuple)):
+            assert len(ty) == len(val)
+            for t, v in zip(ty, val):
+                self._prepare_args(t, v, stream, retr, kernelargs)
+
         else:
             raise NotImplementedError(ty, val)
 
--- a/numba/cuda/compiler.py
+++ b/numba/cuda/compiler.py
@@ -756,8 +756,7 @@ class _Kernel(serialize.ReduceMixin):
             devrec = wrap_arg(val).to_device(retr, stream)
             kernelargs.append(devrec)
 
-        elif isinstance(ty, (types.UniTuple, types.Tuple, types.NamedUniTuple,
-                             types.NamedTuple)):
+        elif isinstance(ty, types.BaseTuple):
             assert len(ty) == len(val)
             for t, v in zip(ty, val):
                 self._prepare_args(t, v, stream, retr, kernelargs)
--- a/numba/cuda/tests/cudapy/test_array_args.py
+++ b/numba/cuda/tests/cudapy/test_array_args.py
@@ -1,4 +1,5 @@
 import numpy as np
+from collections import namedtuple
 
 from numba import cuda
 from numba.cuda.testing import unittest, CUDATestCase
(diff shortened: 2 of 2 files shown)


gmarkall:
This could probably do with some extra testing - e.g. passing tuples of tuples, tuples of arrays.

numba/cuda/compiler.py
```diff
@@ -756,6 +756,12 @@ def _prepare_args(self, ty, val, stream, retr, kernelargs):
             devrec = wrap_arg(val).to_device(retr, stream)
             kernelargs.append(devrec)
 
+        elif isinstance(ty, (types.UniTuple, types.Tuple, types.NamedUniTuple,
+                             types.NamedTuple)):
```
stuartarchibald, line 760: Can this just check against `types.BaseTuple`?

stuartarchibald (commented):
Thanks for the patch, looks good, just one minor suggestion. I also tested against more involved stuff like:

```python
from numba import cuda
import numpy as np

@cuda.jit
def f(r, x):
    r[0] = x[0][0]
    r[1] = x[1]
    r[2] = x[2][0]

z = np.ones(1, dtype=np.int64)
x = (1 * z, 10, (4, 3))
r = np.zeros(len(x), dtype=np.int64)
f[1, 1](r, x)
print(r)
```
it works fine, largely by delegation to `self._prepare_args`.

stuartarchibald:
> This could probably do with some extra testing - e.g. passing tuples of tuples, tuples of arrays.

Adding a few spot check cases with these would be a good idea I think.

numba/cuda/compiler.py
gmarkall: Looks like it can :-)

gmarkall:
> > This could probably do with some extra testing - e.g. passing tuples of tuples, tuples of arrays.
> 
> Adding a few spot check cases with these would be a good idea I think.

Now added!

rdemaria:
Would this work on numba.roc as well?

stuartarchibald:
> Would this work on numba.roc as well?

I'm afraid not, CUDA and ROCm have different compilers.

stuartarchibald (approved):
Thanks for the patch, changes look good.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_139`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_139`.

Passed.

PR #6349, opened 2020-10-12
Refactor Dispatcher to remove unnecessary indirection

gmarkall:
This PR refactors the dispatcher so that the split between the C and C++ implementations no longer exists - the entire implementation is now in C++. The consensus in recent discussions was that this split is probably an artefact of the constraints of various C and C++ compilers of years past that no longer apply. This refactoring is mostly to remove the indirection the split caused, with one or two more modern idioms applied.

I've left in my commit history whilst making the changes so that the working is apparent. In summary, the changes required:

- Removal of `_dispatcher.h` and `_dispatcherimpl.cpp`.
- Renaming `_dispatcher.c` to `_dispatcher.cpp`
- Moving the methods and members of the `Dispatcher` object from `_dispatcherimpl.cpp` into the `DispatcherObject` in `_dispatcher.cpp` - note that this was then renamed to `Dispatcher`.
- Where the C proxy methods were used (e.g. `dispatcher_resolve`) they are replaced with calls to the `Dispatcher` methods instead (e.g. `self->resolve`).
- Replacement of C-style includes (`string.h` etc.) with C++ ones (`cstring` etc).
- Replacement of `malloc` / `free` with `new` / `delete` operators
- Some other minor changes to enable compilation of the dispatcher code with a C++ (e.g. casting string literals to `(char*)` as necessary).
- Now that `_typeof.c` is included by a C++ source, it has an `extern "C"` block added as appropriate.


numba/_dispatcher.cpp
```diff
@@ -97,11 +98,12 @@ else                                                            \
     }                                                           \
 }
 
+typedef std::vector<Type> TypeTable;
+typedef std::vector<PyObject*> Functions;
 
-typedef struct DispatcherObject{
+class Dispatcher {
```
stuartarchibald, line 104: Let's take the opportunity to document this code? Doc for class and methods?

numba/_dispatcher.cpp
```diff
@@ -97,11 +98,12 @@ else                                                            \
...
+public:
     PyObject_HEAD
-    /* Holds borrowed references to PyCFunction objects */
-    dispatcher_t *dispatcher;
     char can_compile;        /* Can auto compile */
     char can_fallback;       /* Can fallback */
     char exact_match_required;
```
stuartarchibald, line 109:
```suggestion
    char exact_match_required; /* Whether an exact match is required */
```

numba/_dispatcher.cpp
```diff
@@ -115,28 +117,76 @@ typedef struct DispatcherObject{
...
+        }
+        return NULL;
+    }
+
+    void clear() {
+        functions.clear();
+        overloads.clear();
+    }
+
+};
```
stuartarchibald, line 168: Note that this is just lifted from the current `_dispatcherimpl.cpp`. https://github.com/numba/numba/blob/3b1e4abd1b82ab5dc60f912ad43faa6e26cd87ba/numba/_dispatcherimpl.cpp#L14-L52

numba/_dispatcher.cpp
```diff
@@ -115,28 +117,76 @@ typedef struct DispatcherObject{
...
-} DispatcherObject;
+    /* Number of arguments to function */
+    int argct;
+    /* Used for selecting overloaded function implementations */
+    TypeManager *tm;
+    /* An array of overloads */
+    Functions functions;
+    /* A flattened array of argument types to all overloads
+     * (invariant: sizeof(overloads) == argct * sizeof(functions)) */
+    TypeTable overloads;
```
stuartarchibald, line 128: Is it intended for these to now be public?

numba/_dispatcher.cpp
```diff
@@ -530,14 +586,10 @@ Dispatcher_call(DispatcherObject *self, PyObject *args, PyObject *kws)
...
 
-    /* If compilation is enabled, ensure that an exact match is found and if
-     * not compile one */
-    int exact_match_required = self->can_compile ? 1 : self->exact_match_required;
-
     /* We only allow unsafe conversions if compilation of new specializations
        has been disabled. */
-    cfunc = dispatcher_resolve(self->dispatcher, tys, &matches,
-                               !self->can_compile, exact_match_required);
+    cfunc = self->resolve(tys, matches, !self->can_compile,
```
stuartarchibald, line 591: Perhaps leave a comment about `matches` being mutated? It's not obvious from the call site any more.

stuartarchibald (commented):
Thanks for undertaking this refactor, it's remove a lot of indirection that was previously required due to Python 2.7/General C++ cross compatibility issues. The changes look good in general with just  a few minor comments to resolve. Once done I'll push it through the build farm. Thanks again!

numba/_dispatcher.cpp
gmarkall: The have to be, because mixing public/private members results in a non-standard-layout class, so `offsetof` won't work with it then (used for implementing the members at https://github.com/numba/numba/pull/6349/files#diff-4374dcde8d4424f3951a6810c497d65972bfe81587b15ec6f91015a8b9dcd751R652).

numba/_dispatcher.cpp
```diff
@@ -115,28 +117,76 @@ typedef struct DispatcherObject{
...
+        matches = 0;
+        if (0 == ovct) {
+            // No overloads registered
+            return NULL;
+        }
+        if (argct == 0) {
+            // Nullary function: trivial match on first overload
+            matches = 1;
+            selected = 0;
+        }
```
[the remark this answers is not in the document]
gmarkall: How could there be multiple definitions with no arguments?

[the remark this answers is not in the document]
gmarkall: (Note that this was just moved from https://github.com/numba/numba/pull/6349/files#diff-56cb1358ce78e8b6ba8ea05ac6a688da59d236c53f039245b01cb24a15364161L31-L35)

numba/_dispatcher.cpp
gmarkall: Yes, good point.

gmarkall: Added.

numba/_dispatcher.cpp
gmarkall: Documentation added.

numba/_dispatcher.cpp
gmarkall: I've added a slightly different description.

gmarkall:
@stuartarchibald Many thanks for the review - I believe all comments are now addressed.

numba/_dispatcher.cpp
stuartarchibald: ahhh of course, `PyObject_HEAD` etc, and this is class not struct, could make it struct?

numba/_dispatcher.cpp
```diff
@@ -135,6 +155,19 @@ class Dispatcher {
...
+       Returns the best match, as well as the number of matches found.
+
+       Parameters:
+
+       - sig: an array of Type objects, one for each parameter.
+       - matches: the number of matches found (mutated by this function).
+       - allow_unsafe: whether to match overloads that would require an unsafe
+                       cast.
+       - exact_match_required: Whether all arguments types must match the
+                               overload's types exactly. When not true,
```
stuartarchibald, line 168:
```suggestion
                               overload's types exactly. When false,
```

numba/_dispatcher.cpp
stuartarchibald: thanks

numba/_dispatcher.cpp
stuartarchibald: Yours is better :) Thanks!

numba/_dispatcher.cpp
stuartarchibald: Thanks for doing that, looks good.

numba/_dispatcher.cpp
gmarkall: It's possible to, I think it's just a stylistic thing. I prefer "class with public members" to "struct" in this case because I tend to expect classes to have methods and structs not to, but could change it if there's a compelling reason to go the other way.

stuartarchibald: I'm ok with it either way, it's just the `PyObject_HEAD` making it more struct-like for `offsetof`, but agree RE methods. Will go with leaving as-is until something sufficiently compelling arrives to warrant change.

stuartarchibald (approved):
Thanks for the fix-ups, looks good.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_137`.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cpu_86`.

numba/_dispatcher.cpp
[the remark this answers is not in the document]
sklam: ah ok.

btw, there shouldn't be multiple definitions for no arguments. It will have to be a mistake.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cpu_86`.

passed

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_137`.

passed.

stuartarchibald:
@sklam given the inherent risk involved when this part of the code is changed, did you want to take a look at this too? I've reviewed and the changes seem good, build farm is happy.

sklam (approved):
The changes look good.

I also did a quick benchmark on the dispatcher overhead with the following ipython script:

```python
from numba import njit


@njit
def arity_0():
    return


@njit
def arity_1(x):
    return

@njit
def arity_2(x, y):
    return


@njit
def arity_n(*args):
    return

%timeit arity_0()
%timeit arity_1(1)
%timeit arity_2(1, 2)
%timeit arity_n(1, 2, 3, 4, 5, 6, 7, 8, 9, 10)
```

On master 3d18f1449fdb58f3e81d8b3295541cd1147700e0, i'm getting:

```
62.5 ns ± 0.768 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
135 ns ± 3.04 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
150 ns ± 2.72 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
407 ns ± 10.2 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

With this patch, I get:

```
59 ns ± 1.49 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
121 ns ± 2.79 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
149 ns ± 2.42 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
387 ns ± 1.49 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

There's a clear reduction of overhead; esp, for 0-arity function.

PR #6278, opened 2020-09-23
Improve typed list performance.

stuartarchibald:
As title.

PR #6351, opened 2020-10-12
Dev/6335+6278

esc:
#6335 & #6278 -- for testing purposes


esc:
now that #6335 and #6278 have been merged this PR is no longer needed.

PR #6353, opened 2020-10-12
[WIP] Add new impl cache entry for overloads with omitted args

### LINKED ISSUE #6253: Recompilation of bound methods templates due to omitted args in signature

Hi, we recently found that compilation time of some SDC methods can be decreased by eliminating recompilation of function templates that have default arguments in signature. It seems that during the first compilation (which occurs during type-inference) the cache_key uses args that do not refer to default values (i.e. `args=(Dummy, int64, int64)`), and during the second compilation (which occurs during lowering of 'test_impl$1' func) args are different and include omitted types (i.e. `args=(Dummy, int64, int64, omitted(default=None))`) so cache miss occurs in 
https://github.com/numba/numba/blob/c448052c875d7a84e8d067f8483a1a1caa1dfe9b/numba/core/typing/templates.py#L675-L680

Here's the reproducer:
```
import numba
from numba.core import types, cgutils
from numba.extending import (
    models,
    lower_builtin,
    register_model,
    make_attribute_wrapper,
    type_callable,
    typeof_impl,
    overload_method
)

def make_dummy_type():
    class Dummy(object):
        def __init__(self, value):
            self.value = value

    class DummyType(types.Type):
        def __init__(self):
            super(DummyType, self).__init__(name="Dummy")

    dummy_type = DummyType()

    @register_model(DummyType)
            members = [("value", types.intp)]

    make_attribute_wrapper(DummyType, "value", "value")

    @type_callable(Dummy)
    def type_dummy(context):
        def typer(value):
            return dummy_type

        return typer

    @lower_builtin(Dummy, types.intp)
        typ = sig.return_type
        [value] = args
        dummy.value = value
        return dummy._getvalue()

    @typeof_impl.register(Dummy)
    def typeof_dummy(val, c):
        return DummyType()

    # Store attributes
    return Dummy, DummyType


@overload_method(DummyType, 'foo')
def dummy_foo_ovld(self, a, b, c=None):
    def dummy_foo_impl(self, a, b, c=None):
        if c is None:
            return 42
        else:
            return a + b

    return dummy_foo_impl

@numba.njit
def test_impl():
    d = Dummy(1)
    return d.foo(1, 2)

print(test_impl())
```

The issue can be verified by two entries in LLVM dump for dummy_foo:
> 	Line 7: @"_ZN08NumbaEnv8__main__14dummy_foo_ovld12$3clocals$3e18dummy_foo_impl$242E5Dummyxx27omitted$28default$3dNone$29" = common global i8* null
> 	Line 92: @"_ZN08NumbaEnv8__main__14dummy_foo_ovld12$3clocals$3e18dummy_foo_impl$243E5Dummyxx27omitted$28default$3dNone$29" = common global i8* null

or via following traces in `_get_impl()`:
> DEBUG: cache miss for <OverloadTemplate_dummy_foo_ovld test_issue.1.py:58>, args=(Dummy, int64, int64)
> DEBUG: cache miss for <OverloadTemplate_dummy_foo_ovld test_issue.1.py:58>, args=(Dummy, int64, int64, omitted(default=None))

Debugging that further I found that `_build_impl()` actually does not use the signature it got from the first compilation (and this signature contains proper folded args), here: 
https://github.com/numba/numba/blob/c448052c875d7a84e8d067f8483a1a1caa1dfe9b/numba/core/typing/templates.py#L740-L742

So while the new signature was returned by `get_call_type` the cache_key used to add a new dispatcher still refers to the old args. If new cache_key is defined and both entries are added to impl_cache, so that we have:
>>> self._impl_cache
> {(<numba.core.typing.context.Context object at 0x00000209A2ED3AF0>, (Dummy, int64, int64), ()): (CPUDispatcher(<function dummy_foo_ovld.<locals>.dummy_foo_impl at 0x00000209A4889670>), (Dummy, int64, int64)),
>  (<numba.core.typing.context.Context object at 0x00000209A2ED3AF0>, (Dummy, int64, int64, omitted(default=None)), ()): (CPUDispatcher(<function dummy_foo_ovld.<locals>.dummy_foo_impl at 0x00000209A4889670>), (Dummy, int64, int64))}
> 

the issue with recompilation seems to be solved, but I'm not sure that's a proper fix for it.

BR,
Alexey.

kozlov-alexey:
Addresses the issue when compilation of overloads with default args in
signature is repeated due to difference between compiled signature and
a tuple of argument types in cache_key. Fixes #6253

### DIFF
--- a/numba/core/typing/templates.py
+++ b/numba/core/typing/templates.py
@@ -737,9 +737,21 @@ class _OverloadFunctionTemplate(AbstractTemplate):
         disp = jitdecor(pyfunc)
         # Make sure that the implementation can be fully compiled
         disp_type = types.Dispatcher(disp)
-        disp_type.get_call_type(self.context, args, kws)
+        sig = disp_type.get_call_type(self.context, args, kws)
+
+        if sig is None:
+            return None, None
+
         if cache_key is not None:
             self._impl_cache[cache_key] = disp, args
+
+            # in some cases (e.g. when omitted args are used) compiled sig
+            # may differ from args in the cache_key, so add another cache entry
+            context, args, kwargs = cache_key
+            if sig.args != args:
+                upd_cache_key = (context, sig.args, kwargs)
+                self._impl_cache[upd_cache_key] = disp, args
+
         return disp, args
 
     def get_impl_key(self, sig):


stuartarchibald:
@kozlov-alexey thanks for the patch. This is failing testing in a couple of places, the failures look legitimate, once fixed up I'll schedule this for review. Thanks again!

PR #6120, opened 2020-08-11
bump llvmlite to 0.34

sklam:
as titled

similar to https://github.com/numba/numba/pull/5702/commits/c2ae57a4a3ed1687f843b724a1a160188c1db2ef

PR #6354, opened 2020-10-13
Bump llvmlite to 0.35. series.

stuartarchibald:
As title.

Inverse of https://github.com/numba/numba/pull/6120/files combined with a version bump!

### DIFF
--- a/buildscripts/condarecipe.local/meta.yaml
+++ b/buildscripts/condarecipe.local/meta.yaml
@@ -32,7 +32,7 @@ requirements:
     - numpy
     - setuptools
     # On channel https://anaconda.org/numba/
-    - llvmlite >=0.34.0dev0,<0.35
+    - llvmlite >=0.35.0dev0,<0.35
     # TBB devel version is to match TBB libs
     - tbb-devel >=2019.5       # [not (armv6l or armv7l or aarch64 or linux32)]
   run:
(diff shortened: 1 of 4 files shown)


buildscripts/condarecipe.local/meta.yaml
```diff
@@ -32,15 +32,15 @@ requirements:
     - numpy
     - setuptools
     # On channel https://anaconda.org/numba/
-    - llvmlite >=0.33.0dev0,<0.34
+    - llvmlite >=0.34.0dev0,<0.35
```
stuartarchibald, line 35: Whilst similar previous patches, I'm not convinced by this, I think it's saying "anything from tag 0.34.0dev0" i.e. when work started for 0.34, "through to 0.35 final". I think it might need to say `>=0.35.0dev0` so that it included only development work from the 0.35 series.

sklam: Right, i think `>=0.35.0dev0` will be needed

setup.py
```diff
@@ -13,7 +13,7 @@
 min_python_version = "3.6"
 min_numpy_build_version = "1.11"
 min_numpy_run_version = "1.15"
-min_llvmlite_version = "0.33"
+min_llvmlite_version = "0.34.0.dev0"
 max_llvmlite_version = "0.35"
```
sklam, line 17: I think we will need to bump this as well. because of `'llvmlite >={},<{}'.format(min_llvmlite_version, max_llvmlite_version),`

sklam:
should we bump assuming 0.35 is coming out?

stuartarchibald:
> should we bump assuming 0.35 is coming out?

I've moved all the software parts, its just the README which says about 0.34.

sklam (approved):

PR #6357, opened 2020-10-13
Fix enumerate invalid decref

sklam:
as titled


sklam:
Manually checked failing test against https://github.com/numba/numba/pull/6335

stuartarchibald (approved):
Thanks for the patch. The fix correlates with my debug of this so I think it's good.

stuartarchibald:
Also, manually tested it, segfault is gone.

PR #6359, opened 2020-10-13
Fixes typed list indexing on 32bit

stuartarchibald:
As title.

### DIFF
--- a/numba/typed/listobject.py
+++ b/numba/typed/listobject.py
@@ -631,9 +631,18 @@ def fix_index(tyctx, list_ty, index_ty):
         fast_len_sig, length_fn = _list_length._defn(context.typing_context,
                                                      list_ty)
         length = length_fn(context, builder, fast_len_sig, (ll_list,))
-        zextd_idx = builder.zext(ll_idx, length.type)
-        wrapped_index = builder.add(zextd_idx, length)
-        return builder.select(is_negative, wrapped_index, zextd_idx)
+        # length is an intp
+        # index can be any sort of int
+        # indexing in general is done with a ssize_t which correlates to an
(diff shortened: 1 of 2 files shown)


stuartarchibald:
Try this...

numba/typed/listobject.py
```diff
@@ -631,9 +631,18 @@ def codegen(context, builder, sig, args):
         fast_len_sig, length_fn = _list_length._defn(context.typing_context,
                                                      list_ty)
         length = length_fn(context, builder, fast_len_sig, (ll_list,))
-        zextd_idx = builder.zext(ll_idx, length.type)
-        wrapped_index = builder.add(zextd_idx, length)
-        return builder.select(is_negative, wrapped_index, zextd_idx)
+        # length is an intp
+        # index can be any sort of int
+        # indexing in general is done with a ssize_t which correlates to an
+        # intp. In llvmlite sext and trunc are guarded to return the value
```
esc, line 637: Looking at: http://llvm.org/docs/LangRef.html#conversion-operations

it states that `Equal sized types are not allowed.` for `trunc` and `The bit size of the value must be smaller than the bit size of the destination type, ty2.` for `sext`.

I know that the documentation for LLVM is notoriously unreliable, but this may warrant a double check.

esc: I can confirm this myself, the code is in `llvmlite`, specifically, it is this line:

https://github.com/numba/llvmlite/blob/master/llvmlite/ir/builder.py#L150

esc (approved):
With all of CI passing, this looks good.

sklam (approved):
The patch looks good

PR #6361, opened 2020-10-14
relax the llvmlite requirements for ASV and testing purposes.

esc:


esc:
is no longer needed and can be closed.

PR #6363, opened 2020-10-14
Note that reference counted types are not permitted in set().

stuartarchibald:
As title.


sklam (approved):
LGTM

sklam:
Ignoring the server error in test.

PR #6368, opened 2020-10-15
Testhound/cuda atomic and

testhound:
This pull request adds support for cuda atomic "and". Note that I was force to name the operation "cuda.atomic.atomic_and" instead of "cuda.atomic.and" because "and" is a reserved word.

### DIFF
--- a/numba/cuda/cudaimpl.py
+++ b/numba/cuda/cudaimpl.py
@@ -604,6 +604,16 @@ def ptx_atomic_sub(context, builder, dtype, ptr, val):
     else:
         return builder.atomic_rmw('sub', ptr, val, 'monotonic')
 
+@lower(stubs.atomic.atomic_and, types.Array, types.intp, types.Any)
+@lower(stubs.atomic.atomic_and, types.Array, types.UniTuple, types.Any)
+@lower(stubs.atomic.atomic_and, types.Array, types.Tuple, types.Any)
+@_atomic_dispatcher
+def ptx_atomic_and(context, builder, dtype, ptr, val):
+    if dtype in (types.int32, types.int64, types.uint32, types.uint64):
+        return builder.atomic_rmw('and', ptr, val, 'monotonic')
+    else:
+        raise TypeError('Unimplemented atomic and with %s array' % dtype)
+
 
 @lower(stubs.atomic.max, types.Array, types.intp, types.Any)
 @lower(stubs.atomic.max, types.Array, types.Tuple, types.Any)
--- a/numba/cuda/cudaimpl.py
+++ b/numba/cuda/cudaimpl.py
@@ -604,6 +604,7 @@ def ptx_atomic_sub(context, builder, dtype, ptr, val):
     else:
         return builder.atomic_rmw('sub', ptr, val, 'monotonic')
 
+
 @lower(stubs.atomic.atomic_and, types.Array, types.intp, types.Any)
 @lower(stubs.atomic.atomic_and, types.Array, types.UniTuple, types.Any)
 @lower(stubs.atomic.atomic_and, types.Array, types.Tuple, types.Any)
--- a/numba/cuda/tests/cudapy/test_atomics.py
+++ b/numba/cuda/tests/cudapy/test_atomics.py
@@ -5,6 +5,10 @@ from textwrap import dedent
 from numba import cuda, uint32, uint64, float32, float64
 from numba.cuda.testing import unittest, CUDATestCase
 from numba.core import config
+from random import seed
+from random import randint
+
+seed()
 
 
 def cc_X_or_above(major, minor):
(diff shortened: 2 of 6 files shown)


gmarkall:
The fail is a bit odd:

```
FAIL: test_vectorizer_fastmath_asm (numba.tests.test_parfors.TestParforsVectorizer)
This checks that if fastmath is set and the underlying hardware
  File "/home/vsts/work/1/s/numba/tests/test_parfors.py", line 2731, in test_vectorizer_fastmath_asm
    self.assertTrue('zmm' in v)
```

Is this a known unstable parfors test? (cc @stuartarchibald).

stuartarchibald:
> The fail is a bit odd:
> 
> ```
> FAIL: test_vectorizer_fastmath_asm (numba.tests.test_parfors.TestParforsVectorizer)
> This checks that if fastmath is set and the underlying hardware
>   File "/home/vsts/work/1/s/numba/tests/test_parfors.py", line 2731, in test_vectorizer_fastmath_asm
>     self.assertTrue('zmm' in v)
> AssertionError: False is not true
> ```
> 
> Is this a known unstable parfors test? (cc @stuartarchibald).

Stability depends on the optimiser behaving consistently as the code base changes, and I could imagine recent changes to improve performance may have altered this. I suggest raising an issue.

gmarkall (changes requested):
This looks good to me in general, and tests pass for me locally with a Quadro RTX 8000 and also with the simulator.

There are some suggested changes on the diff to address - in addition to these, rather than naming the function `cuda.atomic.atomic_and`, which carries a little redundancy, I think it would be more idiomatic to name it `cuda.atomic.and_` - for example, llvmlite uses this convention for the `and` IR builder: https://llvmlite.readthedocs.io/en/latest/user-guide/ir/ir-builder.html#llvmlite.ir.IRBuilder.and_

numba/cuda/tests/cudapy/test_atomics.py
```diff
@@ -5,6 +5,10 @@
 from numba import cuda, uint32, uint64, float32, float64
 from numba.cuda.testing import unittest, CUDATestCase
 from numba.core import config
+from random import seed
+from random import randint
+
+seed()
```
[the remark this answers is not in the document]
gmarkall: Issue #6370 created to note the lack of RNG seeding.

stuartarchibald:
> > The fail is a bit odd:
> > ```
> > FAIL: test_vectorizer_fastmath_asm (numba.tests.test_parfors.TestParforsVectorizer)
> > This checks that if fastmath is set and the underlying hardware
> > ----------------------------------------------------------------------
> > Traceback (most recent call last):
> >   File "/home/vsts/work/1/s/numba/tests/test_parfors.py", line 2731, in test_vectorizer_fastmath_asm
> >     self.assertTrue('zmm' in v)
> > AssertionError: False is not true
> > ```
> > 
> > 
> > Is this a known unstable parfors test? (cc @stuartarchibald).
> 
> Stability depends on the optimiser behaving consistently as the code base changes, and I could imagine recent changes to improve performance may have altered this. I suggest raising an issue.

See https://github.com/numba/numba/issues/6371, I think I've worked it out, don't worry about creating an issue. Suspect it's the recent patch to LLVM to ensure SVML CC's are correct has meant something that used to optimise in one way now does so differently. The reason the fail seems intermittent is related to how Numba CI works: https://numba.readthedocs.io/en/latest/developer/contributing.html#continuous-integration-testing, it just so happens the tests added here have shoved the slice with the SVML failing tests onto the build job with SVML enabled.

gmarkall (approved):
Thanks for the revisions! This now looks good to me, and tests fine locally with a Quadro RTX 8000 and also with the simulator.

The CI fail is that mentioned in #6371, unrelated to the content of this PR.

gmarkall:
Could this have a buildfarm run please?

esc:
Buildfarm: "numba_smoketest_cuda_128"

numba/cuda/tests/cudapy/test_atomics.py
```diff
@@ -553,7 +553,7 @@ def test_atomic_and_global(self):
 
         gold = np.zeros(32, dtype=np.int32)
         for i in range(idx.size):
-            gold[idx[i]] &= 1
+            gold[idx[i]] &= rand_const
```
gmarkall, line 556: Why did this work before?

testhound: @gmarkall I don't know why it worked before; I racked my brain for several hours yesterday trying to figure this out. It seems to be something particular to the "&" operator. It fails on "|" for example.

seibert: Does this test even make sense?  (Maybe I'm missing something obvious...)  If `gold` and `ary` are initialized to zeros, won't the result of anding the values with anything result in a zero array at the end?

testhound: > Does this test even make sense? (Maybe I'm missing something obvious...) If `gold` and `ary` are initialized to zeros, won't the result of anding the values with anything result in a zero array at the end?

@seibert yes that's it; I was staring at this so long I missed the obvious. Updating the tests to non-zero initialized arrays.

gmarkall (approved):
This looks good to me, and tests fine locally with hardware and the simulator.

The CI fail is still that mentioned in #6371, unrelated to the content of this PR.

@esc could this have a buildfarm run please?

stuartarchibald:
> @esc could this have a buildfarm run please?

Buildfarm ID: `numba_smoketest_cuda_129`

stuartarchibald (commented):
Thanks for the patch, one minor comment to look at else looks good! @gmarkall thanks for reviewing.

stuartarchibald:
> The CI fail is still that mentioned in #6371, unrelated to the content of this PR.

Fixed this on mainline, CI builds should report correctly now.

stuartarchibald:
> > @esc could this have a buildfarm run please?
> 
> Buildfarm ID: `numba_smoketest_cuda_129`

Passed.

numba/cuda/cudaimpl.py
```diff
@@ -605,6 +605,17 @@ def ptx_atomic_sub(context, builder, dtype, ptr, val):
         return builder.atomic_rmw('sub', ptr, val, 'monotonic')
 
 
+@lower(stubs.atomic.and_, types.Array, types.intp, types.Any)
+@lower(stubs.atomic.and_, types.Array, types.UniTuple, types.Any)
+@lower(stubs.atomic.and_, types.Array, types.Tuple, types.Any)
+@_atomic_dispatcher
+def ptx_atomic_and(context, builder, dtype, ptr, val):
+    if dtype in (types.int32, types.int64, types.uint32, types.uint64):
```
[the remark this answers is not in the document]
testhound: @stuartarchibald great catch, I just made that change.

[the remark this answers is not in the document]
stuartarchibald: Great, thanks!

stuartarchibald:
Given 7d206b3 is a 1 line import change, I checked this out and ran it locally, atomic tests pass fine, unless there's any objections I don't think there's huge merit in this going through the farm again? I propose marking for merge.

gmarkall:
> Given [7d206b3](https://github.com/numba/numba/commit/7d206b36d8f3bd9f788c832be9954a520adf78a6) is a 1 line import change, I checked this out and ran it locally, atomic tests pass fine, unless there's any objections I don't think there's huge merit in this going through the farm again? I propose marking for merge.

Sounds good to me!

PR #6372, opened 2020-10-15
Skips failing tests if SVML is present.

stuartarchibald:
Seems like with SVML enabled the 512bit instruction generation is
being ignored. Skip these tests that explicitly check for `zmm*`
use if SVML is present.


stuartarchibald:
I think this patch is redundant, it's the `opt=1` vs `opt=0` on the cheap module pass manager that triggers the problem. When `opt=1` it manages to generate IR that defeats the vectorizer when the aggressive optimisation is run later.

stuartarchibald:
This patch is not needed #6395 proves.

PR #6374, opened 2020-10-15
Use Python 3 style OSError handling

eric-wieser:
In Python 3 there is no need to check `errno` any more for most use-cases, as there are specific exception subclasses.

IOError and EnvironmentError are now aliases of OSError, so those are changed here too.

I've left versioneer alone since that's vendored code.

### DIFF
--- a/numba/tests/test_dispatcher.py
+++ b/numba/tests/test_dispatcher.py
@@ -1,4 +1,3 @@
-import errno
 import multiprocessing
 import os
 import platform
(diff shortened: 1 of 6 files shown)


esc:
@eric-wieser thanks for submitting this, I've queued it for review.

numba/tests/test_dispatcher.py
```diff
@@ -137,19 +134,18 @@ def check_access_is_preventable():
     try:
         with open(os.path.join(test_dir, 'write_forbidden'), 'wt') as f:
             f.write('check2')
-    except (OSError, IOError) as e:
+        # access prevention is not possible
+        return False
+    except PermissionError as e:
```
eric-wieser, line 139:
```suggestion
    except PermissionError:
```

stuartarchibald (approved):
Thanks for the patch. Great to see this cleaned up!

Issue #6371, opened 2020-10-15
0.52.0RC1 issues


stuartarchibald:
Things broken on the farm for 0.52.0RC1:

1. CUDA 9 on windows 64bit, python 3.8, np1.18: `numba.cuda.tests.cudapy.test_multithreads.TheMultiThreadCompiling`, `self.assertEqual(p.exitcode, 0)` fails as `p.exitcode` is `0xC0000005`. CC @gmarkall 
2. PPC64LE and ARM7vl: `numba.tests.test_vectorization.TestVectorization.test_slp`, seems like it can't vectorize what's present to a `2 x double`.
3. OSX, all builds: Bizarrely, `numba.tests.npufunc.test_errors.TestFloatingPointExceptions.test_remainder_float` is failing with an assertion error, the length of the warnings/error messages doesn't match (essentially, none were found when one was expected).
4. Windows 32bit, all builds: Fails tests `numba.tests.test_vectorization.TestVectorization.*`, there's no `vector.body` in the LLVM IR. Also fails numerous tests in the family `numba.tests.test_dyn_array.TestNd*` with an integer size matching problem in the error message.
5. Linux x86_64, test environments containing SVML: Fails tests `numba.tests.tests_parfors.TestParforsVectorizer.{test_unsigned_refusal_to_vectorize, test_vectorizer_fastmath_asm}` errors are related to instruction generation being incorrect.
6. AArch64: `libm` has versioned symbol `GLIBC_2.29` missing. Testing fails.

stuartarchibald:
https://github.com/numba/numba/pull/6372 sorts out 5.

stuartarchibald:
@sklam investigated 3. Looks to be the split-optimiser with the cheap pass at `O1` being too strong, moving to `O0` seems to fix and could well increase compilation speed.

stuartarchibald:
https://github.com/numba/numba/pull/6378 hopefully fixes 2 and half of 4.

gmarkall:
>     1. CUDA 9 on windows 64bit, python 3.8, np1.18: `numba.cuda.tests.cudapy.test_multithreads.TheMultiThreadCompiling`, `self.assertEqual(p.exitcode, 0)` fails as `p.exitcode` is `0xC0000005`. CC @gmarkall

Another potential data point on this issue: https://github.com/numba/numba/pull/6367#issuecomment-712166266

stuartarchibald:
Closing, all issue resolved.

GillesC:
[...]

PR #6375, opened 2020-10-15
CUDA tests: reduce managed memory allocation to 0.1 * GPU RAM

gmarkall:
0.5 * GPU RAM can allocate a sizeable proportion of system memory, which can cause stability issues. This change reduces the proportion of GPU memory allocated to 0.1 to make testing more resilient.

This may be involved in Issue #6371.

@stuartarchibald Could you try this on the buildfarm to see if any issues occur with this change please?

### DIFF
--- a/numba/cuda/tests/cudadrv/test_managed_alloc.py
+++ b/numba/cuda/tests/cudadrv/test_managed_alloc.py
@@ -43,7 +43,7 @@ class TestManagedAlloc(ContextResettingTestCase):
     def test_managed_alloc_driver_undersubscribe(self):
         msg = "Managed memory unsupported prior to CC 3.0"
         self.skip_if_cc_major_lt(3, msg)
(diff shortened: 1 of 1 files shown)


gmarkall:
Note: The fail is just a Conda HTTP 500 error.

stuartarchibald:
Thanks for the patch.

Buildfarm ID: `numba_smoketest_cuda_123`

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_123`

This failed. Segfault. Windows 64bit, Python 3.8, NumPy 1.17, CUDA 9.0. `numba.cuda.tests.cudapy.test_multithreads.TheMultiThreadCompiling, self.assertEqual(p.exitcode, 0) fails as p.exitcode is 0xC0000005.`

Same as 1. on https://github.com/numba/numba/issues/6371 but with different packages involved.

Issue #6370, opened 2020-10-15
CUDA tests: There is no RNG seeding in test_atomics.py


gmarkall:
See: https://github.com/numba/numba/pull/6368#discussion_r505447275

The values tested will be random for each invocation of the tests, because there is no RNG seeding. The RNG should be seeded for each test, so that values are stable.

gmarkall:
Closing as resolved in #6376.

PR #6376, opened 2020-10-15
Added a fixed seed in test_atomics.py for issue #6370

John-Ted:
As per issue #6370, I added a fixed seed to the random modules of Python and Numpy in test_atomics.py.


gmarkall (changes requested):
Thanks for the PR!

These changes seed the RNG at module import time, not at test time, so the state of the RNG will have changed by the time any of the tests run. See [this commit](https://github.com/numba/numba/commit/32d4886eee13694ef23a91d5bb4a383d0ffc4fa3) for an example of seeding the RNG within a test - the seeding will need to be done for `test_atomics` in each test that uses the RNG.

John-Ted:
I added the seeding in every test that uses the RNG.

John-Ted:
This is the final version of this PR. Apologies for the previous commits that failed the checks, I am new to open-source.

gmarkall (changes requested):
Many thanks for making the changes - my observations and suggestions on the current revision are:

- Calling `random.seed` is not necessary if a test doesn't use anything from the `random` module (none of the tests in this file except for `test_atomic_compare_and_swap` use it).
- The `random` module is used exactly once in the atomics tests, for `random.shuffle`. Given that its use in these tests is functionally equivalent to `np.random.shuffle`, it would be better to remove the import of `random`, and replace `random.shuffle` with `np.random.shuffle`. Then there is only one RNG to seed, and no need to call `random.seed` anywhere.
- Almost every test uses `np.random`, so rather than adding `np.random.seed` to most of the tests, it would be a lot simpler to add a `setUp` method that calls `np.random.seed` rather than doing it in almost every test. (See the [`setUp` method docs](https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUp) if you need to understand how it works).

Thanks again for your contribution, and do let me know if you have any questions about the above, or if it would be helpful to have some more explicit/step-by-step guidance.

gmarkall:
@John-Ted Another recently-merged PR added the seeding in the `setUp` method - could you please resolve the conflicts on this branch? This should leave the switch from `random.shuffle` to `np.random.shuffle` in this PR.

John-Ted:
Done.

gmarkall (approved):
Many thanks for the update! This looks good to me.

gmarkall:
@stuartarchibald Do you want a buildfarm run for this?

stuartarchibald:
> @stuartarchibald Do you want a buildfarm run for this?

Buildfarm ID: `numba_smoketest_cuda_135`.

stuartarchibald (approved):
Thanks for the patch.

PR #6377, opened 2020-10-15
CUDA: Fix various issues in test suite

gmarkall:
This PR fixes various issues that have been discovered whilst working on the CUDA dispatcher. Since each of the changes are minor, they are all rolled into one PR. Changes:

- Replace deprecated `argtypes` / `restypes` use with signatures:
  - `test_blackscholes.py`
  - `test_idiv.py`
  - `test_matmul.py`
  - `test_nondet.py`
  - `test_powi.py`
  - `test_py2_div_issue.py`
- Correct mismatches between signatures and arguments:
  - `test_linker.py`
  - `test_constmem.py`
  - `test_intrinsics.py`
- Correct logic errors:
  - `test_optimization.py`: This test didn't correctly get the PTX, and
     was instead asserting that the second part of the key from
     `inspect_asm()` (the signature) didn't contain the removed
     fragments, instead of actually checking the PTX code.

### DIFF
--- a/numba/cuda/tests/cudadrv/test_linker.py
+++ b/numba/cuda/tests/cudadrv/test_linker.py
@@ -78,8 +78,8 @@ class TestLinker(CUDATestCase):
             i = cuda.grid(1)
             x[i] += bar(y[i])
 
-        A = np.array([123])
-        B = np.array([321])
+        A = np.array([123]).astype(np.int32)
+        B = np.array([321]).astype(np.int32)
 
         foo[1, 1](A, B)
 
--- a/numba/cuda/tests/cudadrv/test_linker.py
+++ b/numba/cuda/tests/cudadrv/test_linker.py
@@ -78,8 +78,8 @@ class TestLinker(CUDATestCase):
             i = cuda.grid(1)
             x[i] += bar(y[i])
 
-        A = np.array([123]).astype(np.int32)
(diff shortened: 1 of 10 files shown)


numba/cuda/tests/cudadrv/test_linker.py
```diff
@@ -78,8 +78,8 @@ def foo(x, y):
             i = cuda.grid(1)
             x[i] += bar(y[i])
 
-        A = np.array([123])
-        B = np.array([321])
+        A = np.array([123]).astype(np.int32)
+        B = np.array([321]).astype(np.int32)
```
stuartarchibald, line 82: perhaps use `np.array(<arg>, dtype=np.int32)` so as to not create a temporary?

stuartarchibald (commented):
Thanks for the patch, one minor suggestion else looks good.

numba/cuda/tests/cudadrv/test_linker.py
gmarkall: Done.

stuartarchibald (approved):
Thanks for the fixes.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_133`

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_133`

Passed.

PR #6378, opened 2020-10-15
Fix incorrect CPU override in vectorization test.

stuartarchibald:
As title.


stuartarchibald:
Buildfarm ID: `numba_smoketest_cpu_78`

sklam:
looks good. just pending buildfarm result.

PR #6379, opened 2020-10-15
Use O0 to enable inline and not affect loop-vectorization by later O3…

sklam:
… passes

- this seems to fix ufunc not being vectorized

Related to #6371: 

> OSX, all builds: Bizarrely, numba.tests.npufunc.test_errors.TestFloatingPointExceptions.test_remainder_float is failing with an assertion error, the length of the warnings/error messages doesn't match (essentially, none were found when one was expected).


sklam:
Marking as RTM because #6395 passed CI

PR #6384, opened 2020-10-16
Fix failing tests to match on platform invariant int spelling.

stuartarchibald:
As title.


stuartarchibald:
Fixes the bit about int sizes in item 4. of https://github.com/numba/numba/issues/6371

sklam:
I'm good with these changes If buildfarm passes.

stuartarchibald:
> I'm good with these changes If buildfarm passes.

It's part of https://github.com/numba/numba/pull/6395, to save having to test all the patches separately.

PR #6390, opened 2020-10-19
Updates inspect_cfg

### LINKED ISSUE #6389: graphviz/dot internal buffer truncation causes issues for `inspect_cfg`.
- [x] I have tried using the 0.52.0rc1 ~released~ version of Numba (most recent is

This
```python
from numba import njit, types
from numba.experimental import jitclass
import numpy as np
import string

@jitclass({x:types.int64 for x in string.ascii_letters})
class Foo():
    def __init__(self):
        pass
@njit
def bar():
    a = Foo()

bar()

bar.inspect_cfg(bar.signatures[0]).display(view=True)
```
creates a massive mangled name for `Foo` which then causes graphviz/dot problems later when trying to render the CFG.
```
graphviz.backend.CalledProcessError: Command '['dot', '-Tpdf', '-O', '_ZN8__main__7bar$241E.gv']' returned non-zero exit status 1. [stderr: b'Error: not well-formed (invalid token) in line 1 \n... ... :int64,O:int64,P:int64,Q:int64,R:int64,S:int64,T:int64,U:int64,V:int64,W:int64,X:int64,Y:int64,Z:int64&#92;&gt;&quot</td> ...\nin label of node Node0x555722175a90\n']
```
if this is truncated, it can also cause problems when trying to render the CFG as the filename is based on the graph name, which is massive, and results in an `OSError` as maximum filename length is exceeded.

stuartarchibald:
This:

1. Fixes too long string names being passed to graphviz/dot (#6389).
2. Fixes too long filenames being generated in the case of long mangled
   function names (#6389).
3. Adds highlighting for LLVM intrinsic calls and function calls
   in general.
4. Now uses the Numba refop pruner metadata to spot raises.
5. Adds a format option to the `.display()` method.

Fixes #6389

### DIFF
--- a/numba/core/codegen.py
+++ b/numba/core/codegen.py
@@ -95,9 +95,10 @@ class _CFG(object):
         self.dot = ll.get_function_cfg(fn)
         self.kwargs = kwargs
 
-    def pretty_printer(self, filename=None, view=None, highlight=True,
-                      interleave=False, strip_ir=False, show_key=True,
-                      fontsize=10):
+    def pretty_printer(self, filename=None, view=None, render_format=None,
+                       highlight=True,
+                       interleave=False, strip_ir=False, show_key=True,
+                       fontsize=10):
         """
         "Pretty" prints the DOT graph of the CFG.
         For explanation of the parameters see the docstring for
@@ -118,7 +119,9 @@ class _CFG(object):
                                     returns=_default,
                                     raises=_default,
                                     meminfo=_default,
-                                    branches=_default)
+                                    branches=_default,
+                                    llvm_intrin_calls=_default,
+                                    function_calls=_default,)
         _interleave = SimpleNamespace(python=_default, lineinfo=_default)
 
         def parse_config(_config, kwarg):
@@ -163,6 +166,8 @@ class _CFG(object):
         cs['raise'] = 'lightpink'
         cs['meminfo'] = 'lightseagreen'
         cs['return'] = 'purple'
+        cs['llvm_intrin_calls'] = 'rosybrown'
+        cs['function_calls'] = 'tomato'
 
         # Get the raw dot format information from LLVM and the LLVM IR
         fn = self.cres.get_function(self.name)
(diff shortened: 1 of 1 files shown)


stuartarchibald:
This might be useful for manual testing:
```python
from numba import njit, types
from numba.experimental import jitclass
import numpy as np
import string

@jitclass({x:types.int64 for x in string.ascii_letters[:52]})
class Foo():
    def __init__(self):
        pass

@njit
def baz(s, x):
    b = Foo()
    return hash(s), np.linalg.svd(x)

baz('abc', np.ones((3, 4)))

baz.inspect_cfg(baz.signatures[0]).display(view=True, format='png')
```

esc (approved):
I have reviewed the changes as follows:

> 1. Fixes too long string names being passed to graphviz/dot (#6389).

I tried the reproducer on master and saw it fail. I tried it with this PR and it was successful
 
> 2. Fixes too long filenames being generated in the case of long mangled
function names (#6389).

I tried this with my function who's length is 66. I manually clamped the limit to 60 and watched the warning output appear on my screen and the name of the file being truncated.

> 3. Adds highlighting for LLVM intrinsic calls and function calls in general.

Works!

Screen Shot 2020-10-19 at 15 28 08

Screen Shot 2020-10-19 at 15 28 19


> 4. Now uses the Numba refop pruner metadata to spot raises.

Yes!

Screen Shot 2020-10-19 at 15 29 07


> 5. Adds a format option to the .display() method.

This works too.

Screen Shot 2020-10-19 at 15 30 11

numba/core/codegen.py
```diff
@@ -217,8 +231,10 @@ def init_digraph(name, fname, fontsize):
         nrt_incref = re.compile(r"@NRT_incref\b")
         nrt_decref = re.compile(r"@NRT_decref\b")
         nrt_meminfo = re.compile("@NRT_MemInfo")
-        ll_raise = re.compile("ret i32 1,")
-        ll_return = re.compile("ret i32 [^1],")
+        ll_intrin_calls = re.compile(r".*call.*@llvm\..*")
+        ll_function_call = re.compile(r".*call.*@.*")
+        ll_raise = re.compile("ret i32.*\!ret_is_raise.*")
+        ll_return = re.compile("ret i32 [^1],?.*")
```
sklam, line 237: manually verified that these are working

sklam (approved):
Looks good

PR #6400, opened 2020-10-21
Testhound/cuda atomic or

testhound:
This pull request adds support for cuda atomic "or". Note the operation is name "cuda.atomic.or_" due to "or" being a reserved word in python. Additionally this branch is a sub-branch of testhound/cuda_atomic_and.

### DIFF
--- a/numba/cuda/stubs.py
+++ b/numba/cuda/stubs.py
@@ -399,7 +399,7 @@ class atomic(Stub):
         atomically.
         """
 
-    class atomic_and(Stub):
+    class and_(Stub):
         """atomic_and(ary, idx, val)
 
         Perform atomic ary[idx] &= val. Supported on int32, int64, uint32 and
--- a/numba/cuda/stubs.py
+++ b/numba/cuda/stubs.py
@@ -399,6 +399,16 @@ class atomic(Stub):
         atomically.
         """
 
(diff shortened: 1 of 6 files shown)


gmarkall (changes requested):
In general this looks good to me, and tests fine with hardware and the simulator locally.

There is one suggestion on the diff to fix the CI fail with the docs build.

gmarkall:
Note: this is based on #6368.

numba/cuda/stubs.py
```diff
@@ -399,6 +399,26 @@ class sub(Stub):
...
+
+    class or_(Stub):
+        """or_(ary, idx, val)
+
+        Perform atomic ary[idx] \|= val. Supported on int32, int64, uint32 and
+        uint64 operands only.
+
+        Returns the old value at the index location as if it is loaded
+        atomically.
+        """
```
gmarkall, line 420: I hadn't noticed that escaping the pipe symbol for Sphinx would cause a flake8 error - I can't think of another way to escape this though, so perhaps the best thing to do here is to suppress that particular warning for this docstring with:

```suggestion
        """  # noqa: W605
```

gmarkall (approved):
This looks good to me, and tests fine locally with hardware and the CUDA simulator.

PR #6402, opened 2020-10-21
Fix `typed.Dict` and `typed.List` crashing on parametrized types

### LINKED ISSUE #6401: Crash involving `typed.Dict`, caching, and parametrized type

Likely needs #6336 fixed to start, ie requires `numba>0.51`.

Sorry, this is still a little vague but since I only have Windows boxes debugging crashes is not easy. The script below crashes when run more than once, ie when data is loaded from cache. It no longer crashes when caching is turned off (`@jit(..., cache=False)` on line 64), unique type names are used (line 14), or `doit(x, y)` is called with arguments of the same type (`typeof(x) == typeof(y)` on line 65).

```python
from numba import generated_jit, jit, typed, types
from numba.core.datamodel.models import UniTupleModel
from numba.extending import box, register_model, typeof_impl, unbox


class Parametrized(tuple):
    def __init__(self, tup):
        assert all(isinstance(v, str) for v in tup)  # the actual work is already done by __new__


class ParametrisedType(types.Type):
    """this is essentially UniTuple(unicode_type, n) BUT type name is the same for all n"""
    def __init__(self, value):
        super(ParametrisedType, self).__init__('ParametrisedType')  # <- using a unique name (like f'ParametrisedType({value})') here will not crash
        self.dtype = types.unicode_type
        self.n = len(value)

    @property
    def key(self):
        return self.n

    def __len__(self):
        return self.n


register_model(ParametrisedType)(UniTupleModel)


@typeof_impl.register(Parametrized)
def typeof_unit(val, c):
    return ParametrisedType(val)


@unbox(ParametrisedType)
def unbox_parametrized(typ, obj, context):
    return context.unbox(types.UniTuple(typ.dtype, len(typ)), obj)

# reproducer currently does not need this
# @box(ParametrisedType)
# def box_parametrized(typ, val, context):
#     tup = context.box(types.UniTuple(typ.dtype, len(typ)), val)
#     cls = context.pyapi.unserialize(context.pyapi.serialize_object(Parametrized))
#     obj = context.pyapi.call_function_objargs(cls, (tup,))
#     context.pyapi.decref(tup)
#     return obj


@generated_jit
def dict_vs_cache_vs_parametrized(v):
    typ = v

    def dict_vs_cache_vs_parametrized_impl(v):
        # typed.List shows same behaviour after fix for #6397
        d = typed.Dict.empty(types.unicode_type, typ)
        d['data'] = v

    return dict_vs_cache_vs_parametrized_impl


def test_dict_vs_cache_vs_parametrized():
    # crashes when run more than once (ie when compiled function is loaded from cache)
    x, y = Parametrized(('horst', 'hanz')), Parametrized(('horst',))

    @jit(nopython=True, cache=True)
    def doit(x, y):  # <- does not crash when typeof(x) == typeof(y)
        dict_vs_cache_vs_parametrized(x)
        dict_vs_cache_vs_parametrized(y)

    for ii in range(50):  # <- somtimes works a few times
        assert doit(x, y) is None


if __name__ == '__main__':
    test_dict_vs_cache_vs_parametrized()
```

I have a hunch caching is not really to blame here. I'd suspect the non-unique type name is derailing code generation somehow, just have not found a reproducer w/o caching, yet.

asodeur:
Fixes #6401 as described [here](https://github.com/numba/numba/issues/6401#issuecomment-713474356)

### DIFF
--- a/numba/tests/test_dictimpl.py
+++ b/numba/tests/test_dictimpl.py
@@ -648,4 +648,4 @@ class TestDictImpl(TestCase):
         set_parametrized_data.targetctx.init()
 
         for ii in range(50):  # <- somtimes works a few times
-            assert set_parametrized_data(x, y) is None
+            self.assertIsNone(set_parametrized_data(x, y))
--- a/numba/tests/test_dictimpl.py
+++ b/numba/tests/test_dictimpl.py
@@ -3,11 +3,16 @@ Testing C implementation of the numba dictionary
 """
 
 import ctypes
+import os
 import random
+import subprocess
+import sys
 
 from numba.tests.support import TestCase
-from numba import _helperlib
+from numba import generated_jit, _helperlib, jit, typed, types
 from numba.core.config import IS_32BITS
+from numba.core.datamodel.models import UniTupleModel
+from numba.extending import register_model, typeof_impl, unbox
 
 
 DKIX_EMPTY = -1
@@ -179,6 +184,79 @@ class DictIter(object):
             return k.decode(), v.decode()
 
     next = __next__    # needed for py2 only
+
+
+############################################################
+# Supporting code for TestDictImpl.test_parametrized_types
+############################################################
+class Parametrized(tuple):
+    def __init__(self, tup):
+        assert all(isinstance(v, str) for v in tup)
+
+
+class ParametrisedType(types.Type):
+    """this is essentially UniTuple(unicode_type, n)
+    BUT type name is the same for all n"""
+    def __init__(self, value):
+        super(ParametrisedType, self).__init__('ParametrisedType')
+        self.dtype = types.unicode_type
+        self.n = len(value)
+
+    @property
+    def key(self):
+        return self.n
+
+    def __len__(self):
+        return self.n
+
+
+register_model(ParametrisedType)(UniTupleModel)
+
+
+@typeof_impl.register(Parametrized)
+def typeof_unit(val, c):
+    return ParametrisedType(val)
+
+
+@unbox(ParametrisedType)
+def unbox_parametrized(typ, obj, context):
+    return context.unbox(types.UniTuple(typ.dtype, len(typ)), obj)
+
+
+@generated_jit
+def dict_vs_cache_vs_parametrized(v):
+    typ = v
+    def objmode_vs_cache_vs_parametrized_impl(v):
+        # typed.List shows same behaviour after fix for #6397
+        d = typed.Dict.empty(types.unicode_type, typ)
+        d['data'] = v
+
+    return objmode_vs_cache_vs_parametrized_impl
+
+
+def dict_vs_cache_vs_parametrized_worker():
+    """crashes when run more than once
+    (ie when compiled function is loaded from cache)"""
+    @jit(nopython=True, cache=True)
+    def get_unit_system_data(x, y):
+        dict_vs_cache_vs_parametrized(x)
+        dict_vs_cache_vs_parametrized(y)
+
+    x, y = Parametrized(('horst', 'hanz')), Parametrized(('horst',))
+
+    for ii in range(50):  # <- somtimes works a few times
+        assert get_unit_system_data(x, y) is None
+
+
+def run_test_parametrized_types_in_subprocess():
+    env = os.environ.copy()
+    subproc = subprocess.Popen(
+        [sys.executable, '-c',
+         'import numba.tests.test_dictimpl as test_mod\n' +
+         'test_mod.dict_vs_cache_vs_parametrized_worker()'],
+        env=env)
+    subproc.wait()
+    return subproc.returncode
 
 
 class TestDictImpl(TestCase):
@@ -577,3 +655,14 @@ class TestDictImpl(TestCase):
         # Check different sizes of the key & value.
         for i in range(1, 8):
             self.check_sizing(key_size=i, val_size=i, nmax=2**i)
+
+    def test_parametrized_types(self):
+        """https://github.com/numba/numba/issues/6401"""
+        self.assertEqual(
+            run_test_parametrized_types_in_subprocess(),
+            0,
+            'Child process died')
+        self.assertEqual(
+            run_test_parametrized_types_in_subprocess(),
+            0,
+            'Child process died')
--- a/numba/tests/test_dictimpl.py
+++ b/numba/tests/test_dictimpl.py
@@ -226,6 +226,7 @@ def unbox_parametrized(typ, obj, context):
 @generated_jit
 def dict_vs_cache_vs_parametrized(v):
     typ = v
+
     def objmode_vs_cache_vs_parametrized_impl(v):
         # typed.List shows same behaviour after fix for #6397
         d = typed.Dict.empty(types.unicode_type, typ)
--- a/numba/tests/test_dictimpl.py
+++ b/numba/tests/test_dictimpl.py
@@ -186,19 +186,19 @@ class DictIter(object):
     next = __next__    # needed for py2 only
 
 
-############################################################
-# Supporting code for TestDictImpl.test_parametrized_types
-############################################################
 class Parametrized(tuple):
+    """supporting type for TestDictImpl.test_parametrized_types
+    needs to be global to be cacheable"""
     def __init__(self, tup):
         assert all(isinstance(v, str) for v in tup)
 
 
-class ParametrisedType(types.Type):
+class ParametrizedType(types.Type):
     """this is essentially UniTuple(unicode_type, n)
     BUT type name is the same for all n"""
+
     def __init__(self, value):
-        super(ParametrisedType, self).__init__('ParametrisedType')
+        super(ParametrizedType, self).__init__('ParametrizedType')
         self.dtype = types.unicode_type
         self.n = len(value)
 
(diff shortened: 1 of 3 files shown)


gmarkall:
Thanks for the PR! Once it's no longer [WIP], we can queue it up for review.

gmarkall:

numba/tests/test_dictimpl.py
```diff
@@ -181,6 +186,80 @@ def __next__(self):
...
 
+############################################################
+# Supporting code for TestDictImpl.test_parametrized_types
+############################################################
+class Parametrized(tuple):
+    def __init__(self, tup):
+        assert all(isinstance(v, str) for v in tup)
+
+
+class ParametrisedType(types.Type):
```
stuartarchibald, line 197: Parameterize/Parameterised, am unconcerned about which spelling, but would be good to be consistent :)

numba/tests/test_dictimpl.py
```diff
@@ -181,6 +186,80 @@ def __next__(self):
...
+        # typed.List shows same behaviour after fix for #6397
+        d = typed.Dict.empty(types.unicode_type, typ)
+        d['data'] = v
+
+    return objmode_vs_cache_vs_parametrized_impl
+
+
+def dict_vs_cache_vs_parametrized_worker():
+    """crashes when run more than once
+    (ie when compiled function is loaded from cache)"""
```
stuartarchibald, line 240:
```suggestion
    (i.e. when compiled function is loaded from cache)"""
```

numba/tests/test_dictimpl.py
```diff
@@ -181,6 +186,80 @@ def __next__(self):
...
+    def objmode_vs_cache_vs_parametrized_impl(v):
+        # typed.List shows same behaviour after fix for #6397
+        d = typed.Dict.empty(types.unicode_type, typ)
+        d['data'] = v
+
+    return objmode_vs_cache_vs_parametrized_impl
+
+
+def dict_vs_cache_vs_parametrized_worker():
+    """crashes when run more than once
```
stuartarchibald, line 239:
```suggestion
    """Has had tendency to segfault when run more than once
```

numba/tests/test_dictimpl.py
```diff
@@ -181,6 +186,80 @@ def __next__(self):
...
+    (ie when compiled function is loaded from cache)"""
+    @jit(nopython=True, cache=True)
+    def get_unit_system_data(x, y):
+        dict_vs_cache_vs_parametrized(x)
+        dict_vs_cache_vs_parametrized(y)
+
+    x, y = Parametrized(('horst', 'hanz')), Parametrized(('horst',))
```
stuartarchibald, line 246:
```suggestion
    x, y = Parametrized(('a', 'b')), Parametrized(('a',))
```

numba/tests/test_dictimpl.py
```diff
@@ -181,6 +186,80 @@ def __next__(self):
...
+
+def run_test_parametrized_types_in_subprocess():
+    env = os.environ.copy()
+    subproc = subprocess.Popen(
+        [sys.executable, '-c',
+         'import numba.tests.test_dictimpl as test_mod\n' +
+         'test_mod.dict_vs_cache_vs_parametrized_worker()'],
+        env=env)
+    subproc.wait()
+    return subproc.returncode
```
stuartarchibald, line 260: I'm trying to understand why this is in a subprocess, is it with view of isolating a potential segfault? If the PR is fixing something that has a tendency to segfault, I'd rather it segfaults the test suite as a demonstration that it's not working than survive via isolation and report it, else we'd end up with all tests in subprocesses just-in-case?

numba/typed/typedobjectutils.py
```diff
@@ -120,15 +120,16 @@ def _container_get_meminfo(context, builder, container_ty, c):
...
+def _get_incref_decref(context, module, datamodel, container_element_type):
     assert datamodel.contains_nrt_meminfo()
 
     fe_type = datamodel.fe_type
     data_ptr_ty = datamodel.get_data_type().as_pointer()
     refct_fnty = ir.FunctionType(ir.VoidType(), [data_ptr_ty])
     incref_fn = module.get_or_insert_function(
         refct_fnty,
-        name='.numba_{}_incref${}'.format(container_type, fe_type),
+        name='{}.{}_incref'.format(
```
stuartarchibald, line 131:
```suggestion
        name='.numba_{}.{}_incref'.format(
```
it's helpful to keep the `numba` name in there for debugging.

numba/typed/typedobjectutils.py
```diff
@@ -140,7 +141,8 @@ def _get_incref_decref(context, module, datamodel, container_type):
 
     decref_fn = module.get_or_insert_function(
         refct_fnty,
-        name='.numba_{}_decref${}'.format(container_type, fe_type),
+        name='{}.{}_decref'.format(
+            context.fndesc.mangled_name, container_element_type),
```
stuartarchibald, line 145: As above RE name.

numba/typed/typedobjectutils.py
```diff
@@ -177,14 +179,16 @@ def build_wrapper(fn):
 
     wrapfn = module.get_or_insert_function(
         wrapfnty,
-        name='.numba_{}_item_equal.wrap${}'.format(container_type, fe_type)
+        name='{}.{}_equal.wrap'.format(
+            context.fndesc.mangled_name, container_element_type)
```
stuartarchibald, line 183: As above RE name.

numba/typed/typedobjectutils.py
```diff
@@ -177,14 +179,16 @@ def build_wrapper(fn):
...
     )
     build_wrapper(wrapfn)
 
     equal_fnty = ir.FunctionType(ir.IntType(32), [data_ptr_ty, data_ptr_ty])
     equal_fn = module.get_or_insert_function(
         equal_fnty,
-        name='.numba_{}_item_equal${}'.format(container_type, fe_type),
+        name='{}.{}_equal'.format(
+            context.fndesc.mangled_name, container_element_type),
```
stuartarchibald, line 191: As above RE name.

numba/tests/test_dictimpl.py
```diff
@@ -181,6 +186,80 @@ def __next__(self):
     next = __next__    # needed for py2 only
 
 
+############################################################
+# Supporting code for TestDictImpl.test_parametrized_types
+############################################################
```
stuartarchibald, line 191: I suggest putting all this support code in the actual test.

stuartarchibald (commented):
Thanks for the patch and for debugging and fixing this! The suggested fix seems to work around the issue described. I've reviewed the changes, there's just a few things to take a look at, most of which are in relation to the way Numba does things opposed to anything fundamental. Thanks again!

numba/tests/test_dictimpl.py
asodeur: I have not found a reproducer w/o subprocess, yet. I've tried and `Dispatcher._make_finalizer(()(); Dispatcher._reset_overloads(); jit(...)` to force re-loading from cache (which seems to do the job) but it still does not crash unless the cached function is used in another process. Is there anything more I can try and reset/delete in a single process?

asodeur: After a little more experimentation I came-up with the below. Does the way I am resetting the dispatchers and the target context look resonable?
```
    def test_parametrized_types(self):
        """https://github.com/numba/numba/issues/6401"""

        register_model(ParametrizedType)(UniTupleModel)

        @typeof_impl.register(Parametrized)
        def typeof_unit(val, c):
            return ParametrizedType(val)

        @unbox(ParametrizedType)
        def unbox_parametrized(typ, obj, context):
            return context.unbox(types.UniTuple(typ.dtype, len(typ)), obj)

        @generated_jit
        def dict_vs_cache_vs_parametrized(v):
            typ = v

            def objmode_vs_cache_vs_parametrized_impl(v):
                # typed.List shows same behaviour after fix for #6397
                d = typed.Dict.empty(types.unicode_type, typ)
                d['data'] = v

            return objmode_vs_cache_vs_parametrized_impl

        @jit(nopython=True, cache=True)
        def set_parametrized_data(x, y):
            # Has had a tendency to segfault when the compiled function
            # was loaded from cache in a different process than the one
            # it was originally compiled in.
            # The new process is simulated below by resetting the dispatchers
            # and the target context
            dict_vs_cache_vs_parametrized(x)
            dict_vs_cache_vs_parametrized(y)

        x, y = Parametrized(('a', 'b')), Parametrized(('a',))
        set_parametrized_data(x, y)

        # reset dispatchers and targetctx to force re-load from cache as a new process would jit
        set_parametrized_data._make_finalizer()()
        set_parametrized_data._reset_overloads()
        set_parametrized_data.targetctx.init()

        for ii in range(50):  # <- somtimes works a few times
            assert set_parametrized_data(x, y) is None
```

numba/tests/test_dictimpl.py
asodeur: I'll try to stick with American english then ...

numba/tests/test_dictimpl.py
asodeur: Moved everything into the actual test, except for the extension types which need to be module globals to be pickleable/cacheable.

numba/tests/test_dictimpl.py
```diff
@@ -577,3 +603,49 @@ def test_sizing(self):
...
+        set_parametrized_data(x, y)
+
+        # reset dispatchers and targetctx to force re-load from cache as if a
+        # new process would jit the function
+        set_parametrized_data._make_finalizer()()
+        set_parametrized_data._reset_overloads()
+        set_parametrized_data.targetctx.init()
+
+        for ii in range(50):  # <- somtimes works a few times
+            assert set_parametrized_data(x, y) is None
```
sklam, line 651:
```suggestion
            self.assertIsNone(set_parametrized_data(x, y))
```
for better error message under unittest

sklam (approved):

PR #6403, opened 2020-10-21
Add `types.ListType.key`

### LINKED ISSUE #6397: `ListType` needs to implement `key`

Currently `ListType` cannot cope with parametrized types that do not have unique names because `ListType.key` is not implemented. The example below fails with `TypingError` when it tries to reuse the implementation of `l1.append` for `l2.append`:
```python
from numba import jit, types
from numba.core.datamodel.models import IntegerModel
from numba.extending import register_model
from numba.typed import List

class MyParametrizedType(types.Type):
    def __init__(self, n):
        super(MyParametrizedType, self).__init__('MyParametrizedType')  # <- type name does not contain n
        self.n = n
        self.bitwidth = 64

    @property
    def key(self):
        return self.n

register_model(MyParametrizedType)(IntegerModel)

xty, yty = MyParametrizedType(1), MyParametrizedType(2)

@jit(types.void(xty, yty), nopython=True, no_cpython_wrapper=True)  # <- fails with TypingError
def get_unit_system_data(x, y):
    l1 = List.empty_list(xty)
    l2 = List.empty_list(yty)

    l1.append(x)
    l2.append(y)
```
This likely affects the corresponding iterable and iterator types as well. `ListType.key` should probably look like:
```python
@property
def key(self):
    return self.item_type
```

asodeur:
as title plus test, fixes #6397

### DIFF
--- a/numba/core/types/containers.py
+++ b/numba/core/types/containers.py
@@ -627,6 +627,10 @@ class ListType(IterableType):
         name = "{}[{}]".format(self.__class__.__name__, itemty,)
         super(ListType, self).__init__(name)
 
+    @property
+    def key(self):
+        return self.item_type
+
     def is_precise(self):
         return not isinstance(self.item_type, Undefined)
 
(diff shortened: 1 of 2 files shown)


gmarkall:

stuartarchibald (approved):

stuartarchibald:
As an aside, I also wonder if `.key` should be made abstract so as to force implementation.

PR #6406, opened 2020-10-21
Enable caching functions arising from exec of strings.

stuartarchibald:
As title.


stuartarchibald:
This needs more thought about uniquely capturing a `FunctionType` instance.
 * globals
 * closures
 * anything else in `.__code__` that is captured and can change.

The file path is also a bit strange and needs looking at.

It also needs tests.

numba/core/caching.py
```diff
@@ -318,6 +318,42 @@ def from_function(cls, py_func, py_file):
...
+        # protection with strings (no module or source to check).
+        return self._identifier
+
+    @classmethod
+    def _hash(cls, py_func):
+        buf = []
+        for x in py_func.__code__.co_consts:
+            buf.append(str(x))
+        const_bytes = (''.join(buf)).encode('UTF-8')
+        data = py_func.__code__.co_code + const_bytes
```
sklam, line 341: This won't pick up changes in function name, function arguments, etc. Maybe consider all states in https://github.com/python/cpython/blob/66d3b589c44fcbcf9afe1e442d9beac3bd8bcd34/Objects/codeobject.c#L989-L1012

Might also need the function type signature as this does not store the value of default args.

numba/core/caching.py
```diff
@@ -318,6 +318,42 @@ def from_function(cls, py_func, py_file):
...
+        # have to be quite fussy about this as there's very little aliasing
+        # protection with strings (no module or source to check).
+        return self._identifier
+
+    @classmethod
+    def _hash(cls, py_func):
+        buf = []
+        for x in py_func.__code__.co_consts:
+            buf.append(str(x))
+        const_bytes = (''.join(buf)).encode('UTF-8')
```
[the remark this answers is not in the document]
stuartarchibald: WIP :)

numba/core/caching.py
stuartarchibald: thanks, I've been considering https://github.com/numba/numba/pull/6406#issuecomment-713861806 and generally poking at the contents of `codeobject` to see what can/cannot be easily worked out.

numba/core/caching.py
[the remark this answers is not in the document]
stuartarchibald: From a vague memory, I think I put this in when messing about locally as it was an easy thing to trigger as part of the identifier.

numba/core/caching.py
stuartarchibald: My main concern here is freevars and globals, and wiring through enough data to achieve accurate and correct behaviour.

sklam: I am just not sure if we can handle freevars and globals as is. I am open to the idea of making cacheable functions handle freevars and globals differently than non-cacheable functions. I'm thinking making them linked at load time.

numba/tests/test_caching.py
```diff
@@ -1072,5 +1072,163 @@ def test_caching(self):
...
+
+            f = mod.str_closure1
+            self.assertPreciseEqual(f(3), 6) # 3 + 3 = 6
+            f = mod.str_closure2
+            self.assertPreciseEqual(f(3), 8) # 3 + 5 = 8
+            f = mod.str_closure3
+            self.assertPreciseEqual(f(3), 10) # 3 + 7 = 8
+            f = mod.str_closure4
+            self.assertPreciseEqual(f(3), 12) # 3 + 9 = 12
+            self.check_pycache(5) # 1 nbi, 4 nbc
```
stuartarchibald, line 1182: Normally when the environment is rebuilt as part of cache replay the module for the function is loaded so as to provide globals etc. These tests are failing because there's no module associated with a str exec function and the function refers to a variable that is in the globals supplied to exec.

stuartarchibald:
I think most of the test failures in the above are because the cache file presence check is looking in a `__pycache__` based on the imported module containing the `exec(str)` functions, whereas the `_StringSrcCacheLocator` inherits from `_UserWideCacheLocator` which caches in the user-wide cache directory.

gmarkall:
Reopening as there's still interest in this work.

dlee992:
I would like to catch up with this abandoned PR. If I can find a workable solution based on it, will open a PR to continue.

I feel this draft has already worked? Just need to merge with main, and fix some corner-cases?

PR #6410, opened 2020-10-23
Fixes issue #6386

### LINKED ISSUE #6386: Can't AOT String Keyed Dictionaries:  undefined symbol: _numba_hashsecret_siphash_k1
- [x] I have tried using the latest released version of Numba (0.51.2).

Running this
```python
from numba.pycc import CC
from numba import i8
from numba.core.types import DictType, unicode_type
cc = CC('my_module')

nb_nt = DictType(unicode_type,i8)

@cc.export('foo',(DictType(unicode_type,i8),unicode_type))
def foo(d,x):
   if(x not in d):
      d[x] = len(d)

cc.compile()

import my_module
```
Produces this
```bash
  File "numba_error.py", line 15, in <module>
    import my_module
ImportError: /home/danny/my_module.cpython-38-x86_64-linux-gnu.so: undefined symbol: _numba_hashsecret_siphash_k1
```
I don't really know what's going on, but my best guess is this stuff:
https://github.com/numba/numba/blob/master/numba/cpython/hashing.py#L437
Isn't making its way into the .so

DannyWeitekamp:
There is a GlobalVariable that holds the hashsecret info read in from the ctypes pythonapi. This pull request changes this variable so that it's linkage is set to "linkonce" instead of the default  "external". This fixes #6386, where an 'undefined symbol' error is thrown when importing an AOT compiled numba module that internally hashes strings (which needs the hashsecret value at runtime). 

I'm not really sure what the best linkage choice is but 'linkonce' seems like an okay choice based on the [llvm documentation](https://llvm.org/docs/LangRef.html#linkage-types), and it fixes the issue. Unittest included.

It's unclear to me based on the existing code how the extension module can know how to find python's hashsecret, but it looks like hashes computed from the compiled functions come out the same as python, so it seems fine.


gmarkall:
Many thanks for the PR! There are a couple of issues causing the CI fails at the moment - one is flake8 failing, and the other seems to be fails in `numba.tests.test_import_pycc.TestImportStringHashing.test_import_str_dicts` - do post back if you need any  further guidance looking into the issue. 

> It's unclear to me based on the existing code how the extension module can know how to find python's hashsecret, but it looks like hashes computed from the compiled functions come out the same as python, so it seems fine.

It's not a part of the code I'm familiar with, but it looks like it might be by emitting this: https://github.com/numba/numba/blob/master/numba/cpython/hashing.py#L639

DannyWeitekamp:
@gmarkall Seems to be passing now, I guess "extern_weak" was the right choice not "linkonce". Luckily there are some recent nitty gritty tests on unicode stuff that broke it when I picked the wrong one. Thanks for for pointing out the flake8 issues, I need to start being more consistent about that in my own projects. I changed the test so that it is simplified and integrated with the existing pycc tests (which do a better job of cleaning up the build pycc modules), and confirmed that it replicates the issue when the fix is not there.

I was worried that since _build_hashsecret() was in the body of hashing.py that it wouldn't get emitted, but maybe it works fine because it's not pulling out the literal hashsecret, but linking to a symbol that python exposes? Still not sure, but this fixes the issue and doesn't break anything, so I'm going to call that a win for now. 

Let me know if there are any other issues. 
Cheers,
 -Danny

gmarkall:
> I was worried that since _build_hashsecret() was in the body of hashing.py that it wouldn't get emitted, but maybe it works fine because it's not pulling out the literal hashsecret, but linking to a symbol that python exposes?

I think that's what's happening, via: https://github.com/numba/numba/blob/master/numba/cpython/hashing.py#L437

> Still not sure, but this fixes the issue and doesn't break anything, so I'm going to call that a win for now.

I'll queue the PR as ready for review then :-) Note that as we're in a release candidate / release phase right now, it might be a short while before this gets reviewed.

DannyWeitekamp:
Hmm once I tried using this on some more complex examples "extern_weak" produced some segfaults, but "linkonce" worked fine, so maybe it's worth looking into the tests that failed when it was set to "linkonce". Seems almost like these tests don't give consistent results, and maybe I just got lucky? I guess we can discuss when ready for review.

gmarkall:
Could you share your more complex examples that cause the segfaults?

gmarkall:
Following some discussion of this issue in the Numba triage meeting today - it may be that the correct fix for the issue is not to adjust the linkage of `_numba_hashsecret_siphash_k1`, but to instead define it in the code that is included the the AOT-compiled module (i.e. `modulemixin.c`), or ensure its visibility if it is already present.

gmarkall:
> Following some discussion of this issue in the Numba triage meeting today - it may be that the correct fix for the issue is not to adjust the linkage of `_numba_hashsecret_siphash_k1`, but to instead define it in the code that is included the the AOT-compiled module (i.e. `modulemixin.c`), or ensure its visibility if it is already present.

E.g. a quick hack which seems to fix the reproducer in #6386 is:

```diff
diff --git a/numba/pycc/modulemixin.c b/numba/pycc/modulemixin.c
index 30a2aff22..4146714f2 100644
--- a/numba/pycc/modulemixin.c
+++ b/numba/pycc/modulemixin.c
@@ -23,6 +23,9 @@
 #include "../core/runtime/nrt.h"
 #endif
 
+int64_t _numba_hashsecret_siphash_k0;
+int64_t _numba_hashsecret_siphash_k1;
+int64_t _numba_hashsecret_djbx33a_suffix;
 
 /* NOTE: import_array() is macro, not a function.  It returns NULL on
    failure */
```

DannyWeitekamp:
Awesome! Here's another a reproducer for the segfault I was seeing.

```python
import numba
from numba.pycc import CC
from numba.typed import Dict
from numba.types import DictType, unicode_type, i8

cc = CC('my_module')

@cc.export('declare',(DictType(unicode_type,i8),unicode_type))
def declare(d, x):
   if(x not in d):
      d[x] = len(d)

cc.compile()

from my_module import declare

d = Dict.empty(unicode_type,i8)
for x in [0,1,2,3]:
	declare(d,str(x))
```
 I imagine what was causing it was something along the lines of without linkonce it was trying to relink the hash secret. But I imagine it should work more robustly if the symbol is explicitly defined like you do above.

DannyWeitekamp:
Seems to work now. I guess the previous commit was fine, there was just some bug on Azure's end.

DannyWeitekamp:
Want to link a related issue here: https://github.com/numba/numba/issues/6386#issuecomment-724043361. Seems this PR only fixes part of the issue, but this time I'm really lost.

numba/pycc/modulemixin.c
```diff
@@ -112,6 +116,10 @@ PYCC(pycc_init_) (PyObject *module, PyMethodDef *defs,
                                     env_def_t *envs,
                                     env_gv_t *envgvs)
 {
+    _numba_hashsecret_siphash_k0 = _Py_HashSecret.siphash.k0;
+    _numba_hashsecret_siphash_k1 = _Py_HashSecret.siphash.k1;
+    _numba_hashsecret_djbx33a_suffix = _Py_HashSecret.djbx33a.suffix;
+
```
stuartarchibald, line 122: What I think this is doing is a) providing a symbol, which is what the original issue was and b) making the values for the hash secret correct for the current python process. If you immediately import the AOT compiled module as in https://github.com/numba/numba/issues/6386#issuecomment-724043361 these end up being valid as a result of doing this assignment. If you import from another python process then they get over written with its values when the symbol map injection occurs via:
https://github.com/numba/numba/blob/f82ed9ff82f31a1ab87dc1511a442cfc2d4a22fa/numba/cpython/hashing.py#L667-L671

numba/tests/compile_with_pycc.py
```diff
@@ -3,6 +3,7 @@
 import numpy as np
 
 from numba import float32
+from numba.types import unicode_type,i8
```
stuartarchibald, line 6:
```suggestion
from numba.types import unicode_type, i8
```

numba/tests/compile_with_pycc.py
```diff
@@ -115,3 +116,11 @@ def dict_usecase(arr):
...
         out[k] = k * v
     return out
+
+# checks for issue #6386
+@cc_nrt.export('internal_str_dict', i8(unicode_type))
+def internal_str_dict(x):
+    d = typed.Dict.empty(unicode_type,i8)
+    if(x not in d):
+        d[x] = len(d)
+    return len(d)
```
stuartarchibald, line 126: Would be good to add another test in here where something like this https://github.com/numba/numba/issues/6386#issuecomment-724043361 is AOT compiled i.e. a function hashing a string or bytes. Then this AOT module is imported in the same process and the hash is checked to be the same as the hash of the same string/bytes as called from cpython. Then, to do the same again, but without compilation, in a subprocess to check that it is getting the hash secret from a new process correctly. If this isn't clear/you need a hand with this then please shout :)

stuartarchibald (commented):
Thanks for the patch, and good job debugging and fixing this. The only thing I think this needs is more involved testing and then a run through the Numba buildfarm to make sure it works ok on all OS. Thanks again!

numba/pycc/modulemixin.c
gmarkall: I think it's probably worth adding a description along these lines to comments in the code around here and/or where `_numba_hashsecret_siphash_k0` etc. are declared - it will be kind to a future reader to give them the explanation with the code, rather than making them dig through the original issue to try and determine what this solves.

numba/tests/compile_with_pycc.py
DannyWeitekamp: fixed here: 982e225

numba/pycc/modulemixin.c
DannyWeitekamp: added comment here: 982e225

numba/tests/compile_with_pycc.py
DannyWeitekamp: I think I got it: f9293cb.

DannyWeitekamp:
Believe I've addressed all the comments. Ready for the build farm i suppose whenever you're ready.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cpu_87`.

numba/pycc/modulemixin.c
```diff
@@ -112,6 +116,14 @@ PYCC(pycc_init_) (PyObject *module, PyMethodDef *defs,
                                     env_def_t *envs,
                                     env_gv_t *envgvs)
 {
+    /* Aligns hashsecret with values in current python process so that
+     * hashes computed inside pycc module are the same as in the importing
+     * python process.
+     */
```
stuartarchibald, line 122:
```suggestion
    /* Aligns hashsecret with values in current python process so that
     * hashes computed inside the pycc module are correct if imported
     * by the current process. Subsequent imports get the right hash
     * secret through `numba.cpython.hashing._load_hashsecret`.
     */
```

stuartarchibald, line 122:
```suggestion
    /* Aligns hashsecret with values in current python process so that
     * hashes computed inside the pycc module are correct if imported
     * by the current process. Imports in a new process get the right
     * hash secret through:
     * `numba.cpython.hashing._load_hashsecret`.
     */
```

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cpu_87`.

This passed.

Once the change suggested in https://github.com/numba/numba/pull/6410#discussion_r536041742 is resolved I think this can be merged!

stuartarchibald (approved):
Thanks for the fixes.

DannyWeitekamp:
Awesome!

stuartarchibald:
@DannyWeitekamp Congratulations on your first contribution to Numba!

DannyWeitekamp:
@stuartarchibald Thank you! I'm absolutely loving numba! It probably won't be my last!

PR #6422, opened 2020-10-27
CUDA: Remove deprecated items, expect CUDA 11.1

gmarkall:
Based on #6377 as it removes the use of deprecated options in the test suite.

Remove items from the CUDA target that are deprecated:

- `Device.COMPUTE_CAPABILITY`: Deprecated since 2014 when NumbaPro was open-sourced.
- `DeviceNDArrayBase.to_host`: Likewise.
- `argtypes`, `restype`, and `bind` kwargs to the `cuda.jit` decorator: Deprecated in 0.51.0.

There is some associated code that is also removed - e.g. that influenced by the `bind` kwarg, and the `__writeback` member of `DeviceNDArrayBase`.

This PR also simplifies `declare_device`, as it is only called with a signature now.

Since CUDA 11.1 is now out, `test_runtime` is modified to expect it as a valid version as well.

Whilst working on this PR, I noticed that the comment about the deprecation of the underlying CUDA driver function in `add_callback` was inaccurate, so I've changed the text to reflect that it is not presently deprecated.

### DIFF
--- a/numba/cuda/decorators.py
+++ b/numba/cuda/decorators.py
@@ -1,6 +1,5 @@
-from warnings import warn
 from numba.core import types, config, sigutils
-from numba.core.errors import NumbaDeprecationWarning
+from numba.core.errors import DeprecationError
 from .compiler import (compile_device, declare_device_function, Dispatcher,
                        compile_device_template)
 from .simulator.kernel import FakeCUDAKernel
@@ -20,8 +19,8 @@ def jitdevice(func, link=[], debug=None, inline=False, opt=True):
     return compile_device_template(func, debug=debug, inline=inline, opt=opt)
 
 
-def jit(func_or_sig=None, argtypes=None, device=False, inline=False,
-        link=[], debug=None, opt=True, **kws):
+def jit(func_or_sig=None, device=False, inline=False, link=[], debug=None,
+        opt=True, **kws):
     """
     JIT compile a python function conforming to the CUDA Python specification.
     If a signature is supplied, then a function is returned that takes a
@@ -37,8 +36,6 @@ def jit(func_or_sig=None, argtypes=None, device=False, inline=False,
        .. note:: A kernel cannot have any return value.
     :param device: Indicates whether this is a device function.
     :type device: bool
-    :param bind: (Deprecated) Force binding to CUDA context immediately
-    :type bind: bool
     :param link: A list of files containing PTX source to link with the function
     :type link: list
     :param debug: If True, check for exceptions thrown when executing the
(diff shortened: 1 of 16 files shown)


stuartarchibald:
Thanks for the patch.

RE:
>     * `argtypes`, `restype`, and `bind` kwargs to the `cuda.jit` decorator: Deprecated in 0.51.0.
> 

I don't recall if a specific deprecation cycle was agreed upon or if this part of the API was essentially implementation detail that is/was accidentally exposed and served no purpose? Considering this with respect to users (if any?!) having had long enough to make a change?

gmarkall:
> I don't recall if a specific deprecation cycle was agreed upon or if this part of the API was essentially implementation detail that is/was accidentally exposed and served no purpose? Considering this with respect to users (if any?!) having had long enough to make a change?

I don't think there was a specific deprecation cycle agreed upon, but I can make the following remarks:

- `bind` and `restype` had no effect - if anyone is passing them, they can simply remove them for to become compatible with Numba 0.53 (and observe no change with Numba <= 0.51)
- `argtypes` is a hangover from a time before Numba could infer argument types. Anyone using it can instead pass its argument as the first positional argument to the `cuda.jit` decorator, e.g. https://github.com/numba/numba/pull/6422/files#diff-54fbcfa3b598ec46cab585ba1b7efe1f1349a022b1e455115a62c53f117b8a39L76-R77 - again this will be a change compatible with many older versions of Numba.

I think it will be two releases from deprecation to removal, so given the trivial (and semantically equivalent) nature of the changes needed to update any old code, I feel that this is acceptable to remove now.

numba/cuda/decorators.py
```diff
@@ -145,14 +137,6 @@ def device_jit(func):
             return kernel_jit
 
 
-def declare_device(name, restype=None, argtypes=None):
-    restype, argtypes = convert_types(restype, argtypes)
+def declare_device(name, sig):
+    argtypes, restype = sigutils.normalize_signature(sig)
```
stuartarchibald, line 141: How was this working?!? https://github.com/numba/numba/blob/3b1e4abd1b82ab5dc60f912ad43faa6e26cd87ba/numba/cuda/tests/cudadrv/test_linker.py#L72

stuartarchibald (commented):
Thanks for the patch. Code changes look good. A typo and a pseudo-question, then a build farm run and I propose merge! Thanks again!

stuartarchibald:
> > I don't recall if a specific deprecation cycle was agreed upon or if this part of the API was essentially implementation detail that is/was accidentally exposed and served no purpose? Considering this with respect to users (if any?!) having had long enough to make a change?
> 
> I don't think there was a specific deprecation cycle agreed upon, but I can make the following remarks:
> 
>     * `bind` and `restype` had no effect - if anyone is passing them, they can simply remove them to become compatible with Numba 0.53 (and observe no change with Numba <= 0.51)
> 
>     * `argtypes` is a hangover from a time before Numba could infer argument types. Anyone using it can instead pass its argument as the first positional argument to the `cuda.jit` decorator, e.g. https://github.com/numba/numba/pull/6422/files#diff-54fbcfa3b598ec46cab585ba1b7efe1f1349a022b1e455115a62c53f117b8a39L76-R77 - again this will be a change compatible with many older versions of Numba.
> 
> 
> I think it will be two releases from deprecation to removal, so given the trivial (and semantically equivalent) nature of the changes needed to update any old code, I feel that this is acceptable to remove now.

Given the above, the likelihood of use of this API/trivial changes to fix, and that 0.51..0.53 is August 2020->Q1 2021, near 6 months, then this is change is reasonable.

numba/cuda/decorators.py
gmarkall: `restype` was misleading - it could have been a signature as well, because of what `convert_types` did with it (see just below).

I have spent a fair bit of confused time looking at this already! :-)

gmarkall:
> Thanks for the patch. Code changes look good. A typo and a pseudo-question, then a build farm run and I propose merge! Thanks again!

Typo addressed, pseudo-question pseudo answered :-)

numba/cuda/decorators.py
stuartarchibald: Thanks :) was also confused!

stuartarchibald (approved):
Thanks for the fixes.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_138`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_138`.

Passed.

PR #6430, opened 2020-10-28
CUDA docs: Add RNG example with 3D grid and strided loops

gmarkall:
Based on the question asked at https://numba.discourse.group/t/random-number-generation-on-the-gpu/309, I think the answer makes a good example for the documentation, so this PR adds an example using the RNG with a 3D grid and strided loops.


docs/source/cuda/random.rst
```diff
@@ -64,3 +64,59 @@ Here is a sample program that uses the random number generator::
...
+
+   # Total number of threads
+   nthreads = bx * by * bz * gx * gy * gz
+
+   # Initialize a state for each thread
+   rng_states = create_xoroshiro128p_states(nthreads, seed=1)
+
+   # Generate random numbers
+   arr = cuda.device_array((X, Y, Z), dtype=np.float32)
+   random_3d[(gx, gy, gz), (bx, by, bz)](arr, rng_states)
```
stuartarchibald, line 122: I think that this could benefit from being written in the "unit test examples" style where by a unit test is added to https://github.com/numba/numba/tree/d774cdd4e89c1929dda38fa6c7285e21a0fc129b/numba/tests/doc_examples  and then referred to by the documentation, for example: 
https://github.com/numba/numba/blame/d774cdd4e89c1929dda38fa6c7285e21a0fc129b/docs/source/reference/pysupported.rst#L597-L603
we've found this a good way to ensure non-trivial documented examples remain working under the current code base.

gmarkall: This is now reworked as a unit test example - I added a new place in the CUDA target tests for CUDA doc examples to go, so that the regular doc examples don't get mixed up with CUDA code.

docs/source/cuda/random.rst
```diff
@@ -64,3 +64,27 @@ Here is a sample program that uses the random number generator::
 
     compute_pi[blocks, threads_per_block](rng_states, 10000, out)
     print('pi:', out.mean())
+
+An example managing RNG state size and using a 3D grid
+''''''''''''''''''''''''''''''''''''''''''''''''''''''
```
stuartarchibald, line 69:
```suggestion
An example of managing the RNG state size and using a 3D grid
'''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
```

docs/source/cuda/random.rst
```diff
@@ -64,3 +64,27 @@ Here is a sample program that uses the random number generator::
...
+
+The size of the RNG states scales with the number of threads using the RNG, so
```
stuartarchibald, line 71: I'm not sure about this. Should it be `RNG state`, the RNG holds a state-slot per thread but the state-slots themselves don't scale with thread count, they are a property of the RNG itself?

docs/source/cuda/random.rst
```diff
@@ -64,3 +64,27 @@ Here is a sample program that uses the random number generator::
...
+it is often better to use strided loops in conjunction with the RNG in order to
+keep the state size manageable.
+
+In the following example that initializes a large 3D array with random numbers,
```
stuartarchibald, line 75:
```suggestion
In the following example, that initializes a large 3D array with random numbers,
```

docs/source/cuda/random.rst
```diff
@@ -64,3 +64,27 @@ Here is a sample program that uses the random number generator::
...
+using one thread per output element would result in 453,617,100 RNG states.
+This would take a long time to initialize and poorly utilize the GPU. Instead,
+it uses a fixed size 3D grid with a total of 2,097,152 (``(16 ** 3) * (8 **
```
stuartarchibald, line 78: I think it'd be good to note the relationship between thread count and the number of RNG states in use via this method.

numba/cuda/tests/doc_examples/test_random.py
```diff
@@ -0,0 +1,59 @@
...
+
+        # Generate random numbers
+        arr = cuda.device_array((X, Y, Z), dtype=np.float32)
+        random_3d[(gx, gy, gz), (bx, by, bz)](arr, rng_states)
+        # magictoken.ex_3d_grid.end
+
+        # Some basic tests of the randomly-generated numbers
+        host_arr = arr.copy_to_host()
+        self.assertGreater(np.mean(host_arr), 0.49)
+        self.assertLess(np.mean(host_arr), 0.51)
```
stuartarchibald, line 53: note: this is ok as the seed is set and everything is hardcoded in terms of sizes, so the sequences should always be the same and if it succeeds in CI once it will always succeed.

stuartarchibald (commented):
Thanks for adding the code as a unit test. I've reviewed the patch, few minor things to resolve else looks good. Thanks again!

docs/source/cuda/random.rst
gmarkall: I'm finding it hard to understand the question (in particular "the state-slots themselves don't scale with thread count, they are a property of the RNG itself?") , but I'll try to answer: I think of it as "RNG states" because as you describe, the RNG holds one state slot per thread.

Perhaps this would be better as:

```suggestion
The number of RNG states scales with the number of threads using the RNG, so
```

stuartarchibald: I think the suggestion is clearer and nods towards the direct proportionality with thread count.

RE: "the state-slots themselves don't scale with thread count, they are a property of the RNG itself?"

The there's one slot per thread and each slot contains a single RNG state. The size of this slot is the size of a single RNG state and is a fixed property of the RNG in use, e.g. Mersenne Twister has a different size to some `xoroshiro` variant? I think we're probably talking about the same things but with varying degrees of overloading on "state" and "RNG".

docs/source/cuda/random.rst
gmarkall: This suggestion feels like it's in the grammatical no-man's land between ", which", and "that" - c.f. the examples on slide 11 of https://cims.nyu.edu/~donev/Teaching/WrittenOral/Lecture-Language.pdf:

- “Consider the Pei matrix, which is positive definite.”
- “Consider a Pei matrix that is positive definite.”

Perhaps

```suggestion
In the following example, which initializes a large 3D array with random numbers,
```

is closer to being clear? (though, I still prefer my original "that").

stuartarchibald: I prefer the latter, but am not going to quibble :)

gmarkall: by "the latter", do you mean ", which"?

docs/source/cuda/random.rst
gmarkall: I've added an explicit statement of the number of RNG states to make the connection to the thread count clearer.

docs/source/cuda/random.rst
gmarkall: Edited as per suggestion.

gmarkall:
Edits pushed, but the RTD build seems to have failed due to llvmlite version requirements.

docs/source/cuda/random.rst
stuartarchibald: yes :) but really am not going to quibble.

stuartarchibald (approved):
Changes look good, thanks for the fix-ups.

gmarkall:
Thanks - should this be RTM now?

stuartarchibald:
Would like to get https://github.com/numba/numba/pull/6516 in and do a rebuild, in part because this is the first PR with a doc change since RTD was enabled and I'd like to verify it's working as expected. But yes, this is RTM.

gmarkall:
> Would like to get #6516 in and do a rebuild, in part because this is the first PR with a doc change since RTD was enabled and I'd like to verify it's working as expected. But yes, this is RTM.

Understood - many thanks!

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_140`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_140`.

Passed.

PR #6432, opened 2020-10-29
CUDA: Use _dispatcher.Dispatcher as base Dispatcher class 

gmarkall:
This PR is based on several others: #6349, #6377, #6422, #6333. For reviewing the actual changes in this PR prior to their merge, look only at the last commit in this PR: https://github.com/numba/numba/commit/459cb4f0b1d3e954869ae875700bb4de5ad31b43

This is the another step in more closely aligning the CPU and CUDA targets. There is no functional change visible to the user at this point, but this should enable the addition of support for multiple signatures in the CUDA target, and use of user-supplied boxing and unboxing functions for passing custom data types to kernels in future work.

In this commit, the CUDA Dispatcher uses `_dispatcher.Dispatcher` as its base class - future changes will make `dispatcher.DispatcherBase` the base class of the CUDA Dispatcher, further unifying the targets.

There were various supporting changes needed to make this work; an overview of the changes and some discussion of them follows:

- A new C extension with the base class for CUDA device arrays. This is implemented in `numba/_devicearray.cpp`. The immediate motivation in this PR for its implementation is to provide a way to test if an object is a device array in `typeof_typecode`.

  It is expected that more of the device array functionality can be moved into the C base class - for example, the flags and `ndim` attribute could be moved into the C layer so that the implementation of `typecode_devicendarray` can be made more efficient - it presently uses Python C API calls to extract the relevant information.

  Originally I tried putting this extension in `numba.cuda.cudadrv`, but it resulted in circular import issues when importing it from the Dispatcher C extension that seemed tricky to resolve, so it is presently in the `numba` package, at the same level as the Dispatcher.

  Because the Dispatcher needs access to some of the C APIs provided by the Device Array extension (at present only the `DeviceArrayType`, but more in future), a mechanism for sharing these C APIs between the two modules was needed - the `DeviceArray_API` abstraction introduced here follows the implementation used by NumPy for its C API. Broadly, a `PyCapsule` is used to store a pointer to an array of pointers to C API members and functions. When the Dispatcher is imported, it imports the Device Array module, retrieves the pointer to the API array, and stores the pointer accordingly.

- Addition of a `typecode_devicendarray` function analagous to the `typecode_ndarray` function. As mentioned above, there is scope for making this more efficient by replacing calls to the Python C API with lower-level access to members of a C device array. Additionally, it would benefit from a Device Array equivalent of `PyArray_DESCR` for implementing a typecode cache when the fast path cannot be used.

- Addition of a `_cuda_call` method to the Dispatcher class. This is needed because the overloads of CUDA Dispatchers are not compiled functions that can be directly called by C code, but are `_Kernel` objects in Python. The standard `__call__` / `Dispatcher_call` methods could not be used because they attempt to call the overload. The `_cuda_call` method is therefore close to identical, but only compiles the kernel if necessary, and returns the correct overload, instead of invoking the overload and returning its result.

  In future it would be desirable for CUDA `_Kernel` objects to be replaced by compiled functions that prepare arguments and invoke the kernel via the driver API - this would improve dispatch performance for unspecialized kernels, and could also remove the need for a separate `_cuda_call` method. This is left for future work.

- The majority of other substantial changes are in `numba/cuda/compiler.py` - these adapt the `Dispatcher` and `_Kernel` objects to deal with setting up the `_dispatcher.Dispatcher` parent class up, as well as replacing `definitions` (specific to the CUDA target) with `overloads` (now in common with the CPU target). Various supporting methods (`_compile_for_args`, `_search_new_conversions`, `typeof_pyval`, and `nopython_signatures`) that are called by the C `Dispatcher` class are added - it is expected that these will be subsumed in future when the `DispatcherBase` class becomes the base class of the CUDA `Dispatcher`.

- The changes to `decorators.py` accommodate the minor changes to the `Dispatcher` interface, and refactor it slightly to simplify some logic.

- Additional `Dispatcher` tests are added - these are based on the CPU `Dispatcher` tests, which required some modification mainly because the CPU tests expect values to be returned, which is not possible with CUDA kernel launches.

- Some changes to `test_inspect` are required because the overloads are only keyed by argument type, and not compute capability, unlike the old `definitions` dict.


numba/_devicearray.cpp
```diff
@@ -0,0 +1,117 @@
...
+};
+
+MOD_INIT(_devicearray) {
+    PyObject *m, *d;
+    MOD_DEF(m, "_devicearray", "No docs", NULL)
+    if (m == NULL)
+        return MOD_ERROR_VAL;
+
+    PyObject *c_api;
+    c_api = PyCapsule_New((void *)_DeviceArray_API, NULL, NULL);
```
gmarkall, line 93: This capsule should be given a name, e.g. `"c_api". (According to https://docs.python.org/3/extending/extending.html#using-capsules it is important to properly name capsules)

sklam:
All PR dependencies are merged. But there's now a minor conflict.

gmarkall:
A quick note: now that all the PRs this depends on are merged, this PR can be reviewed by looking at the diff (rather than any individual commit) without changes from other PRs mixed in.

gmarkall:
> All PR dependencies are merged. But there's now a minor conflict.

Didn't notice this comment before - conflicts are resolved.

stuartarchibald:
From out of band chat with @gmarkall. This PR is a stepping stone to making the CUDA target use the `numba.core.dispatcher._DispatcherBase` python class as a base class for CUDA dispatcher objects. The issue is that the CUDA `_Kernel` get's in the way somewhat along with a load of other complex parts and this makes the leap from the current code base to `_DispatcherBase` very large, hence this patch producing a workable intermediate state.

Numba also needs to provide a registration hook in `_typeof.c` so that extension writers (and indeed the CUDA target) can register their type into the `typeof` switch table and also register a finger printing (C?) function for their type. This will make it possible to push the CUDA specific additions to `_typeof.c` in this patch into the CUDA submodule and keep the `_typeof` code as-is.

numba/_typeof.c
```diff
@@ -831,6 +834,52 @@ int typecode_arrayscalar(PyObject *dispatcher, PyObject* aryscalar) {
...
+{
+    int typecode;
+    int dtype;
+    int ndim;
+    int layout = 0;
+
+    PyObject* flags = PyObject_GetAttrString(ary, "flags");
+    if (PyDict_GetItemString(flags, "C_CONTIGUOUS") == Py_True) {
+        layout = 1;
+    } else if (PyDict_GetItemString(flags, "F_CONTIGUOUS") == Py_True) {
```
stuartarchibald, line 848: unchecked calls.

numba/_typeof.c
```diff
@@ -831,6 +834,52 @@ int typecode_arrayscalar(PyObject *dispatcher, PyObject* aryscalar) {
...
 
+static
+int typecode_devicendarray(PyObject *dispatcher, PyObject *ary)
+{
+    int typecode;
+    int dtype;
+    int ndim;
+    int layout = 0;
+
+    PyObject* flags = PyObject_GetAttrString(ary, "flags");
```
stuartarchibald, line 845: unchecked call, also leaks if there's an issue later?

numba/_typeof.c
```diff
@@ -831,6 +834,52 @@ int typecode_arrayscalar(PyObject *dispatcher, PyObject* aryscalar) {
...
+        layout = 1;
+    } else if (PyDict_GetItemString(flags, "F_CONTIGUOUS") == Py_True) {
+        layout = 2;
+    }
+
+    ndim = PyLong_AsLong(PyObject_GetAttrString(ary, "ndim"));
+
+    if (ndim <= 0 || ndim > N_NDIM) goto FALLBACK;
+
+    PyObject* dtype_obj = PyObject_GetAttrString(ary, "dtype");
```
stuartarchibald, line 856: unchecked, potential leak?

numba/_typeof.c
```diff
@@ -831,6 +834,52 @@ int typecode_arrayscalar(PyObject *dispatcher, PyObject* aryscalar) {
...
+    int dtype_num = PyLong_AsLong(PyObject_GetAttrString(dtype_obj, "num"));
+    dtype = dtype_num_to_typecode(dtype_num);
+    if (dtype == -1) goto FALLBACK;
+
+    /* Fast path, using direct table lookup */
+    assert(layout < N_LAYOUT);
+    assert(ndim <= N_NDIM);
```
stuartarchibald, line 863: whilst not strictly necessary as I'm not sure how you'd created one in CUDA, perhaps also add this just to be on the safe side
```suggestion
    assert(ndim <= N_NDIM);
    assert(dtype < N_DTYPES);
```

numba/cuda/decorators.py
```diff
@@ -71,8 +70,37 @@ def jit(func_or_sig=None, device=False, inline=False, link=[], debug=None,
         msg = _msg_deprecated_signature_arg.format('bind')
         raise DeprecationError(msg)
 
+    debug = config.CUDA_DEBUGINFO_DEFAULT if debug is None else debug
     fastmath = kws.get('fastmath', False)
-    if not sigutils.is_signature(func_or_sig):
+
+    if sigutils.is_signature(func_or_sig):
```
stuartarchibald, line 76: note: think this is just code motion, switching the if/else branches

numba/cuda/tests/cudapy/test_dispatcher.py
```diff
@@ -93,6 +104,116 @@ def f(x, y):
...
+
+        with self.assertRaises(TypeError):
+            c_add[1, 1](r, 12.3, 45.6j)
+
+    def test_ambiguous_new_version(self):
+        """Test compiling new version in an ambiguous case
+        """
+        @cuda.jit
+        def foo(r, a, b):
+            r[0] = a + b
```
stuartarchibald, line 166: reuse `add_kernel` from globals?

numba/cuda/tests/cudapy/test_dispatcher.py
```diff
@@ -93,6 +104,116 @@ def f(x, y):
         self.assertEqual(len(f.specializations), 2)
         self.assertIs(f_f32c_f32c, f_f32c_f32c_2)
 
+    # The following tests are based on those in numba.tests.test_dispatcher
```
stuartarchibald, line 107: Note: I've checked against the originals and these new translated tests are a faithful representation of those tests in `numba.tests.test_dispatcher`.

numba/cuda/tests/cudapy/test_inspect.py
```diff
@@ -55,23 +55,23 @@ def foo(x, y):
         # Signature in LLVM dict
         llvmirs = foo.inspect_llvm()
         self.assertEqual(2, len(llvmirs), )
-        self.assertIn((self.cc, (intp, intp)), llvmirs)
-        self.assertIn((self.cc, (float64, float64)), llvmirs)
+        self.assertIn((intp, intp), llvmirs)
```
stuartarchibald, line 58: Should the removal of `cc` be noted somewhere? Docs? IIRC OOB we discussed a deprecation notice to be added.

setup.py
```diff
@@ -344,9 +350,10 @@ def check_file_at_path(path2file):
                                 depends=['numba/_pymodule.h'],
                                 include_dirs=["numba"])
 
-    ext_modules = [ext_dynfunc, ext_dispatcher, ext_helperlib, ext_typeconv,
-                   ext_np_ufunc, ext_npyufunc_num_threads, ext_mviewbuf,
-                   ext_nrt_python, ext_jitclass_box, ext_cuda_extras]
+    ext_modules = [ext_devicearray, ext_dynfunc, ext_dispatcher, ext_helperlib,
+                   ext_typeconv, ext_np_ufunc, ext_npyufunc_num_threads,
+                   ext_mviewbuf, ext_nrt_python, ext_jitclass_box,
+                   ext_cuda_extras]
```
stuartarchibald, line 356:
```suggestion
    ext_modules = [ext_dynfunc, ext_dispatcher, ext_helperlib,
                   ext_typeconv, ext_np_ufunc, ext_npyufunc_num_threads,
                   ext_mviewbuf, ext_nrt_python, ext_jitclass_box,
                   ext_cuda_extras, ext_devicearray,]
```
I'd prefer to keep the existing order so that build logs are directly comparable.

numba/cuda/compiler.py
```diff
@@ -404,13 +394,15 @@ def __init__(self, name, llvmir, options):
         self.cache = {}
         self._extra_options = options.copy()
 
-    def get(self):
+    def get(self, cc=None):
```
stuartarchibald, line 397: I've reviewed this out of order, below is a remark about adding a deprecation notice for CC.

numba/cuda/compiler.py
```diff
@@ -506,38 +498,59 @@ class _Kernel(serialize.ReduceMixin):
...
+
+        self.py_func = py_func
+        self.argtypes = argtypes
+        self.debug = debug
+        self.extensions = extensions or []
+
+        cres = compile_cuda(self.py_func, types.void, self.argtypes,
+                            debug=self.debug,
+                            inline=inline)
+        fname = cres.fndesc.llvm_func_name
```
stuartarchibald, line 515: NOTE: from the original `compile_kernel`.

numba/cuda/compiler.py
```diff
@@ -506,38 +498,59 @@ class _Kernel(serialize.ReduceMixin):
...
         options = {
-            'debug': debug,
+            'debug': self.debug,
             'fastmath': fastmath,
             'opt': 3 if opt else 0
         }
 
+        llvm_module = str(lib._final_module)
+        pretty_name = cres.fndesc.qualname
+        ptx = CachedPTX(pretty_name, llvm_module, options=options)
```
stuartarchibald, line 530: this originally was `str(llvm_module)`, should it still be?

numba/cuda/compiler.py
```diff
@@ -831,6 +851,29 @@ def __init__(self, func, sigs, targetoptions):
...
+            has_stararg = False
+        else:
+            has_stararg = lastarg.kind == lastarg.VAR_POSITIONAL
+
+        exact_match_required = False
+
+        _dispatcher.Dispatcher.__init__(self, self._tm.get_pointer(),
+                                        arg_count, self._fold_args, argnames,
+                                        defargs, can_fallback, has_stararg,
+                                        exact_match_required)
```
stuartarchibald, line 875: Whilst some of this is just setting things to the default values in the `_DispatcherBase` its probably a good idea to keep this as-is such that all these items are considered when doing the next refactoring step.

numba/cuda/compiler.py
```diff
@@ -897,13 +940,45 @@ def call(self, args, griddim, blockdim, stream, sharedmem):
...
         else:
-            argtypes = tuple([self.typingctx.resolve_argument_type(a)
-                              for a in args])
-            kernel = self.compile(argtypes)
+            kernel = _dispatcher.Dispatcher._cuda_call(self, *args)
+
         kernel.launch(args, griddim, blockdim, stream, sharedmem)
 
+    def _compile_for_args(self, *args, **kws):
+        # Based on _DispatcherBase._compile_for_args.
```
stuartarchibald, line 950: Eventually the `self._compilation_chain_init_hook()` should go here.

numba/cuda/compiler.py
```diff
@@ -938,22 +1013,27 @@ def specialized(self):
 
     @property
     def definition(self):
+        warn('Use overloads instead of definition',
+             category=NumbaDeprecationWarning)
```
stuartarchibald, line 1017: Would be good to add a test to check these warnings fire correctly (numerous new deprecation warnings are being added, will note with `test:deprecation`).

numba/cuda/compiler.py
```diff
@@ -938,22 +1013,27 @@ def specialized(self):
...
     @property
-    def _func(self, signature=None, compute_capability=None):
-        cc = compute_capability or get_current_device().compute_capability
-        if signature is not None:
-            return self.definitions[(cc, signature)]._func
-        elif self.specialized:
-            return self.definition._func
+    def definitions(self):
+        warn('Use overloads instead of definitions',
+             category=NumbaDeprecationWarning)
```
stuartarchibald, line 1028: `test:deprecation`

numba/cuda/compiler.py
```diff
@@ -985,14 +1069,18 @@ def inspect_llvm(self, signature=None, compute_capability=None):
         dispatcher is specialized, the IR for the single specialization is
         returned.
         '''
-        cc = compute_capability or get_current_device().compute_capability
+        if compute_capability is not None:
+            warn('passing compute_capability has no effect on the LLVM IR',
+                 category=NumbaDeprecationWarning)
```
stuartarchibald, line 1074: `test:deprecation`

numba/cuda/compiler.py
```diff
@@ -985,14 +1069,18 @@ def inspect_llvm(self, signature=None, compute_capability=None):
...
         if signature is not None:
-            return self.definitions[(cc, signature)].inspect_llvm()
+            return self.overloads[signature].inspect_llvm()
         elif self.specialized:
-            return self.definition.inspect_llvm()
+            warn('inspect_llvm will always return a dict in future',
+                 category=NumbaDeprecationWarning)
```
stuartarchibald, line 1079: `test:deprecation`

numba/cuda/compiler.py
```diff
@@ -1003,29 +1091,37 @@ def inspect_asm(self, signature=None, compute_capability=None):
         '''
         cc = compute_capability or get_current_device().compute_capability
         if signature is not None:
-            return self.definitions[(cc, signature)].inspect_asm()
+            return self.overloads[signature].inspect_asm(cc)
         elif self.specialized:
-            return self.definition.inspect_asm()
+            warn('inspect_asm will always return a dict in future',
+                 category=NumbaDeprecationWarning)
```
stuartarchibald, line 1097: `test:deprecation`

numba/cuda/compiler.py
```diff
@@ -1003,29 +1091,37 @@ def inspect_asm(self, signature=None, compute_capability=None):
...
         if given.
 
+        SASS for the device in the current context is returned.
+
         Requires nvdisasm to be available on the PATH.
         '''
-        cc = compute_capability or get_current_device().compute_capability
+        if compute_capability is not None:
+            warn('passing compute_capability has no effect on the SASS code',
+                 category=NumbaDeprecationWarning)
```
stuartarchibald, line 1115: `test:deprecation`

numba/cuda/compiler.py
```diff
@@ -1003,29 +1091,37 @@ def inspect_asm(self, signature=None, compute_capability=None):
...
         if signature is not None:
-            return self.definitions[(cc, signature)].inspect_sass()
+            return self.overloads[signature].inspect_sass()
         elif self.specialized:
-            return self.definition.inspect_sass()
+            warn('inspect_sass will always return a dict in future',
+                 category=NumbaDeprecationWarning)
```
stuartarchibald, line 1120: `test:deprecation`

numba/cuda/compiler.py
```diff
@@ -1036,22 +1132,21 @@ def inspect_types(self, file=None):
...
-                defn.inspect_types(file=file)
+        for _, defn in self.overloads.items():
+            defn.inspect_types(file=file)
 
     @property
     def ptx(self):
         if self.specialized:
-            return self.definition.ptx
+            warn('ptx will always return a dict in future',
+                 category=NumbaDeprecationWarning)
```
stuartarchibald, line 1142: `test:deprecation`

stuartarchibald (commented):
@gmarkall thanks for putting in the huge amount of effort necessary to make a change such as this. This is a good step on the way to that noted here https://github.com/numba/numba/pull/6432#issuecomment-734328508 (essentially CUDA using Numba's dispatcher without workarounds!). As with most large patches there's a few more things to resolve than normal, but in general this seems the approach seems sound and should make the next steps easier. The review I've done is a "first pass" in an attempt to familiarise myself with the changes and provide an initial round of feedback. Thanks again for working on this, it's a very exciting change!

numba/_devicearray.cpp
```diff
@@ -0,0 +1,117 @@
...
+    }
+    Py_INCREF(&DeviceArrayType);
+    PyModule_AddObject(m, "DeviceArray", (PyObject*)(&DeviceArrayType));
+
+    d = PyModule_GetDict(m);
+    if (d == NULL) {
+        return MOD_ERROR_VAL;
+    }
+
+    PyDict_SetItemString(d, "_DEVICEARRAY_API", c_api);
```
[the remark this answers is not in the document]
gmarkall: The checking is implicit because of the following `if (pyErr_Occurred)` - however, it would probably be better to make the link between what happens in `PyDict_SetItemString` and the return of the error value, so this block would probably be better as:

```C
    int error = PyDict_SetItemString(d, "_DEVICEARRAY_API", c_api);
    Py_DECREF(c_api);
    if (error) {
        return MOD_ERROR_VAL;
```

numba/_devicearray.cpp
```diff
@@ -0,0 +1,117 @@
+#include "_pymodule.h"
+
+/* Include _devicearray., but make sure we don't get the definitions intended
+ * for consumers of the Device Array API.
+ */
+#define NUMBA_IN_DEVICEARRAY_CPP_
```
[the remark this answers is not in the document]
gmarkall: I don't think I'd add `#ifndef NUMBA_IN_DEVICEARRAY_CPP_`, because that implies someone is including it somewhere, and I don't think this file should be included in anything else.

I think for similar reasons, I'd probably not undef it at the end of the file either.

numba/_devicearray.cpp
```diff
@@ -0,0 +1,117 @@
+#include "_pymodule.h"
+
```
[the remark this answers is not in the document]
gmarkall: Added.

numba/_devicearray.cpp
```diff
@@ -0,0 +1,117 @@
...
+    return 0;
+}
+
+static int
+DeviceArray_clear(DeviceArray *self)
+{
+    return 0;
+}
+
+PyTypeObject DeviceArrayType = {
```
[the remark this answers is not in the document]
gmarkall: Added.

numba/_devicearray.cpp
```diff
@@ -0,0 +1,117 @@
...
+};
+
+MOD_INIT(_devicearray) {
+    PyObject *m, *d;
+    MOD_DEF(m, "_devicearray", "No docs", NULL)
+    if (m == NULL)
+        return MOD_ERROR_VAL;
+
+    PyObject *c_api;
+    c_api = PyCapsule_New((void *)_DeviceArray_API,"_devicearray.c_api", NULL);
```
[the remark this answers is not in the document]
gmarkall: Syntax adjusted.

numba/_devicearray.cpp
```diff
@@ -0,0 +1,117 @@
...
+    if (c_api == NULL) {
+        return MOD_ERROR_VAL;
+    }
+
+    DeviceArrayType.tp_new = PyType_GenericNew;
+    if (PyType_Ready(&DeviceArrayType) < 0) {
+        return MOD_ERROR_VAL;
```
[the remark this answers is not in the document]
gmarkall: Added cleanup section at the end.

numba/_devicearray.cpp
```diff
@@ -0,0 +1,117 @@
...
+    }
+    Py_INCREF(&DeviceArrayType);
+    PyModule_AddObject(m, "DeviceArray", (PyObject*)(&DeviceArrayType));
```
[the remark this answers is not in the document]
gmarkall: Now checked.

numba/_devicearray.cpp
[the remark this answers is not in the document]
gmarkall: Something similar to the above (but with a goto to the cleanup section) now added.

numba/_devicearray.h
```diff
@@ -0,0 +1,24 @@
...
+#endif
+
+/* These definitions should only be used by consumers of the Device Array API,
+ * that access the API through the opaque pointer stored in
+ * _devicearray._DEVICEARRAY_API.  We don't want them in _devicearray.cpp itself
+ * because they would conflict with the actual definitions there.
+ */
+#ifndef NUMBA_IN_DEVICEARRAY_CPP_
+
+    extern void **DeviceArray_API;
```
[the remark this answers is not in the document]
gmarkall: As a global it will be initialized to 0 (null) anyway, but I'll make this more explicit.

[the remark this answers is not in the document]
gmarkall: I'm not sure what you mean by "a second import can catch this and return immediately?" though.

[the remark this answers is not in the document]
gmarkall: I just noticed this is extern, so it shouldn't have a value assigned here.

numba/_dispatcher.cpp
```diff
@@ -261,19 +262,28 @@ Dispatcher_clear(Dispatcher *self, PyObject *args)
...
     PyObject *sigtup, *cfunc;
     int i, sigsz;
     int *sig;
     int objectmode = 0;
+    int cuda = 0;
 
-    if (!PyArg_ParseTuple(args, "OO|i", &sigtup,
-                          &cfunc, &objectmode)) {
+    if (!PyArg_ParseTupleAndKeywords(args, kwds, "OO|ip", keywords, &sigtup,
+                                     &cfunc, &objectmode, &cuda)) {
```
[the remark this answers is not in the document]
gmarkall: Possibly because of the C89 heritage of this code?

numba/_dispatcher.cpp
```diff
@@ -678,10 +703,163 @@ Dispatcher_call(Dispatcher *self, PyObject *args, PyObject *kws)
...
+            cfunc = self->resolve(tys, matches, !self->can_compile,
+                                  exact_match_required);
+        }
+    }
+
+    if (matches == 1) {
+        /* Definition is found */
+        retval = cfunc;
+        Py_INCREF(retval);
+        goto CLEANUP;
```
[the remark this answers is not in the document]
gmarkall: It is not.

numba/_dispatcher.cpp
```diff
@@ -678,10 +703,163 @@ Dispatcher_call(Dispatcher *self, PyObject *args, PyObject *kws)
...
+    if (!PyCapsule_CheckExact(c_api)) {
+        PyErr_SetString(PyExc_RuntimeError, "_DEVICEARRAY_API is not PyCapsule object");
+        Py_DECREF(c_api);
+        return -1;
+    }
+
+    DeviceArray_API = (void**)PyCapsule_GetPointer(c_api, "_devicearray.c_api");
+    Py_DECREF(c_api);
+    if (DeviceArray_API == NULL) {
+        PyErr_SetString(PyExc_RuntimeError, "_DEVICEARRAY_API is NULL pointer");
```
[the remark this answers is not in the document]
gmarkall: Yes - NumPy does the same - the reason is that the traceback presented to the user is more informative with this override.

numba/_dispatcher.cpp
```diff
@@ -765,6 +943,12 @@ static PyMethodDef ext_methods[] = {
 
 
 MOD_INIT(_dispatcher) {
+    if (import_devicearray() < 0) {
+      PyErr_Print();
+      PyErr_SetString(PyExc_ImportError, "numba._devicearray failed to import");
```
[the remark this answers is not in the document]
gmarkall: It doesn't seem very convenient to use in this situation - what's the reasoning behind the suggestion?

numba/_dispatcher.cpp
```diff
@@ -678,10 +703,163 @@ Dispatcher_call(Dispatcher *self, PyObject *args, PyObject *kws)
...
+        retval = cfunc;
+        Py_INCREF(retval);
+        goto CLEANUP;
+    } else if (matches == 0) {
+        /* No matching definition */
+        if (self->can_compile) {
+            retval = compile_only(self, args, kws, locals);
+            if (retval) {
+              Py_INCREF(retval);
+            }
```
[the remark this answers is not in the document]
gmarkall: Probably not, I think this is a hangover from when I was doing things differently. Removing the increfs of the returned value from `cuda_compile_only` (as I've renamed it) seems to pass tests, but I'll do a valgrind run of the cudapy tests to check for any use-after-frees this might cause if I'm wrong here (this may take a while).

numba/_dispatcher.cpp
```diff
@@ -261,19 +262,28 @@ Dispatcher_clear(Dispatcher *self, PyObject *args)
...
 static
 PyObject*
-Dispatcher_Insert(Dispatcher *self, PyObject *args)
+Dispatcher_Insert(Dispatcher *self, PyObject *args, PyObject *kwds)
 {
+    static char *keywords[] = {
+        (char*)"sig",
+        (char*)"func",
+        (char*)"objectmode",
+        (char*)"cuda",
```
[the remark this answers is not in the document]
gmarkall: I've added a comment nearby in the source that also notes this, so that it doesn't get lost / forgotten in future.

numba/_dispatcher.cpp
```diff
@@ -455,6 +464,22 @@ compile_and_invoke(Dispatcher *self, PyObject *args, PyObject *kws, PyObject *lo
     return retval;
 }
 
+static
+PyObject*
+compile_only(Dispatcher *self, PyObject *args, PyObject *kws, PyObject *locals)
```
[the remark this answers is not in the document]
gmarkall: Comment added.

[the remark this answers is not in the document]
gmarkall: Done.

numba/_devicearray.h
```diff
@@ -0,0 +1,24 @@
+#ifndef NUMBA_DEVICEARRAY_H_
+#define NUMBA_DEVICEARRAY_H_
+
+#ifdef __cplusplus
+    extern "C" {
+#endif
+
+/* These definitions should only be used by consumers of the Device Array API,
```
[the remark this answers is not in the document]
gmarkall: I've edited the comment a little further than this.

numba/_dispatcher.cpp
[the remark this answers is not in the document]
gmarkall: Removed.

numba/_dispatcher.cpp
```diff
@@ -678,10 +703,163 @@ Dispatcher_call(Dispatcher *self, PyObject *args, PyObject *kws)
...
+            retval = call_cfunc(self, self->fallbackdef, args, kws, locals);
+        } else {
+            /* Raise TypeError */
+            explain_matching_error((PyObject *) self, args, kws);
+            retval = NULL;
+        }
+    } else if (self->can_compile) {
+        /* Ambiguous, but are allowed to compile */
+        retval = compile_only(self, args, kws, locals);
+        Py_INCREF(retval);
```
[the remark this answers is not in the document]
gmarkall: incref removed as per above.

numba/_typeof.c
gmarkall: These return null without setting an exception if the flag was not present, in which case the comparison fails, which would be the desired behaviour.

numba/_typeof.c
gmarkall: Added a check and a corresponding decref.

numba/_typeof.c
gmarkall: Now checked, and a decref added.

numba/_typeof.c
gmarkall: Added.

numba/cuda/compiler.py
gmarkall: The `str()` is now in `llvm_module = str(lib._final_module)` just above.

gmarkall: Maybe the name is now confusing - perhaps it should be `llvm_ir` instead of `llvm_module`? Would that be clearer as to what's going into `CachedPTX`?

numba/cuda/compiler.py
gmarkall: As `nu,mba.cuda.Dispatcher` is not yet inheriting from `_DispatcherBase`, I need to set these things here. (Am I understanding your remark correctly?)

numba/cuda/decorators.py
gmarkall: It also simplified the logic a little bit to turn it upside down - a couple of now-redundant checks are removed, but the remaining code is just moved jue to the inverted condition.

numba/cuda/tests/cudapy/test_dispatcher.py
gmarkall: Now re-used.

setup.py
gmarkall: Now preserved.

numba/cuda/compiler.py
gmarkall: Test added in test_inspect.py.

numba/cuda/compiler.py
gmarkall: Test added in test_inspect.py.

numba/cuda/compiler.py
gmarkall: Test added in test_inspect.py.

numba/cuda/compiler.py
gmarkall: Test added in test_inspect.py.

numba/cuda/compiler.py
gmarkall: Test added in test_inspect.py.

numba/cuda/compiler.py
gmarkall: Test added in test_inspect.py.

numba/cuda/tests/cudapy/test_inspect.py
gmarkall: Further to OOB conversation, I've added back support for `(cc, argtypes)` indexing (and tests below) as well as a deprecation notice in https://github.com/numba/numba/pull/6432/commits/bd4a188bcee7be2e1aafd6d2d05a249039315243

numba/_dispatcher.cpp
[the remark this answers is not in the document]
gmarkall: Valgrind tests seem to be OK with the code after addressing all feedback.

numba/_typeof.c
```diff
@@ -831,6 +834,52 @@ int typecode_arrayscalar(PyObject *dispatcher, PyObject* aryscalar) {
...
+    } else if (PyDict_GetItemString(flags, "F_CONTIGUOUS") == Py_True) {
+        layout = 2;
+    }
+
+    ndim = PyLong_AsLong(PyObject_GetAttrString(ary, "ndim"));
+
+    if (ndim <= 0 || ndim > N_NDIM) goto FALLBACK;
+
+    PyObject* dtype_obj = PyObject_GetAttrString(ary, "dtype");
+    int dtype_num = PyLong_AsLong(PyObject_GetAttrString(dtype_obj, "num"));
```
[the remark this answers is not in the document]
gmarkall: Now checked, leak addressed.

numba/_typeof.c
```diff
@@ -831,6 +834,52 @@ int typecode_arrayscalar(PyObject *dispatcher, PyObject* aryscalar) {
...
+    int layout = 0;
+
+    PyObject* flags = PyObject_GetAttrString(ary, "flags");
+    if (PyDict_GetItemString(flags, "C_CONTIGUOUS") == Py_True) {
+        layout = 1;
+    } else if (PyDict_GetItemString(flags, "F_CONTIGUOUS") == Py_True) {
+        layout = 2;
+    }
+
+    ndim = PyLong_AsLong(PyObject_GetAttrString(ary, "ndim"));
```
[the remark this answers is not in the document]
gmarkall: Now checked, leak addressed.

numba/cuda/compiler.py
gmarkall: Deprecation notices /warnings now added

numba/cuda/compiler.py
gmarkall: After some thought, I've decided that it is clearer as `llvm_ir` anyway, so I've made this change.

numba/cuda/compiler.py
gmarkall: Test added in https://github.com/numba/numba/pull/6432/commits/d9534677c20a7ab9b13e1a08df26024ed9438f8c

numba/cuda/compiler.py
gmarkall: Test added in https://github.com/numba/numba/pull/6432/commits/d9534677c20a7ab9b13e1a08df26024ed9438f8c

gmarkall:
@stuartarchibald many thanks for the great review - this has improved the PR a lot! I believe I've addressed all comments above. 

There were a couple of failing tests from recent merges from master that have been fixed too. One (https://github.com/numba/numba/pull/6432/commits/b9a0e5ad528d24b7e8d8bfe21d2524f5f2fbe4c4) changes the interface of `get_regs_per_thread` slightly in that it doesn't return a dict keyed with the compute capability, but as this hadn't made it into a release yet I don't think the change is breaking / needs deprecation.

numba/_dispatcher.cpp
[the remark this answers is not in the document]
sklam: `p` was added in py3.3. The code here is probably from py2 era. We should fix this in a separate issue.

numba/_dispatcher.cpp
```diff
@@ -678,10 +710,158 @@ Dispatcher_call(Dispatcher *self, PyObject *args, PyObject *kws)
...
+        Py_DECREF(c_api);
+        return -1;
+    }
+
+    DeviceArray_API = (void**)PyCapsule_GetPointer(c_api, "_devicearray.c_api");
+    Py_DECREF(c_api);
+    if (DeviceArray_API == NULL) {
+        PyErr_SetString(PyExc_RuntimeError, "_DEVICEARRAY_API is NULL pointer");
+        return -1;
+    }
```
sklam, line 854: Much of this code can be reduced if `PyImport_ImportModule` is used:
```suggestion
    if (devicearray == NULL) {
        return -1;
    }
    Py_DECREF(devicearray);

    DeviceArray_API = (void**)PyCapsule_Import("numba._devicearray._DEVICEARRAY_API", 0);
    if (DeviceArray_API == NULL) {
        return -1;
    }
```
```

File: numba/_devicearray.cpp [hunk 44]
```diff
@@ -0,0 +1,140 @@
...
+    PyObject *m = nullptr;
+    PyObject *d = nullptr;
+    PyObject *c_api = nullptr;
+    int error = 0;
+
+    MOD_DEF(m, "_devicearray", "No docs", NULL)
+    if (m == NULL)
+        goto error_occurred;
+
+    c_api = PyCapsule_New((void *)_DeviceArray_API, "_devicearray.c_api", NULL);
```
sklam, line 109: To enable `PyCapsule_Import`:
```suggestion
    c_api = PyCapsule_New((void *)_DeviceArray_API, "numba._devicearray._DEVICEARRAY_API", NULL);
```

sklam (commented):
A minor suggestion to reduce the amount of code needed for `PyCapsule`. If the `PyCapsule` object is to be a module attribute, it should have name that matches the `modname.attrname` pattern. That would also allow the use of `PyCapsule_Import` which cuts a lot of the attribute lookup and checking code.

The rest of it looks fine. I didn't spend as much time thinking about DECREF given @stuartarchibald has already done that.

gmarkall:
@sklam many thanks for the review and for showing me about `PyCapsule_Import` - I've now committed your suggestions, and merged in the latest master.

gmarkall:
@sklam A quick ping on this - any thoughts on the changes?

gmarkall:
@sklam @stuartarchibald I've also merged in master - it tests out OK locally for me with CUDA hardware.

gmarkall:
I will look into the CI failures first thing tomorrow.

numba/_dispatcher.cpp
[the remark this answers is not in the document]
stuartarchibald: https://github.com/numba/numba/issues/6662 tracks

numba/_dispatcher.cpp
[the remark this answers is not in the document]
stuartarchibald: Just that the import file name can be added so we know where to look. Am happy to leave, it's not a likely scenario.

numba/_typeof.c
stuartarchibald: thanks

numba/_typeof.c
stuartarchibald: thanks, agree.

numba/_typeof.c
```diff
@@ -831,6 +835,100 @@ int typecode_arrayscalar(PyObject *dispatcher, PyObject* aryscalar) {
...
+        /* No dtype: try the fallback. */
+        goto FALLBACK;
+    }
+
+    PyObject* num_obj = PyObject_GetAttrString(dtype_obj, "num");
+    Py_DECREF(dtype_obj);
+
+    if (num_obj == NULL) {
+        /* This strange dtype has no num - try the fallback. */
+        goto FALLBACK;
```
stuartarchibald, line 890: errors need clearing?

numba/_typeof.c
```diff
@@ -831,6 +835,100 @@ int typecode_arrayscalar(PyObject *dispatcher, PyObject* aryscalar) {
...
+int typecode_devicendarray(PyObject *dispatcher, PyObject *ary)
+{
+    int typecode;
+    int dtype;
+    int ndim;
+    int layout = 0;
+
+    PyObject* flags = PyObject_GetAttrString(ary, "flags");
+    if (flags == NULL)
+        goto FALLBACK;
```
stuartarchibald, line 848: Clear error?

numba/_typeof.c
```diff
@@ -831,6 +835,100 @@ int typecode_arrayscalar(PyObject *dispatcher, PyObject* aryscalar) {
...
+    }
+
+    if (ndim <= 0 || ndim > N_NDIM)
+        goto FALLBACK;
+
+    PyObject* dtype_obj = PyObject_GetAttrString(ary, "dtype");
+    if (dtype_obj == NULL) {
+        /* No dtype: try the fallback. */
+        goto FALLBACK;
```
stuartarchibald, line 882: Clear error?

numba/cuda/compiler.py
stuartarchibald: yes, this is fine.

docs/source/reference/deprecation.rst
```diff
@@ -241,3 +241,67 @@ This feature will be moved with respect to this schedule:
...
+variant directly for specialized Dispatchers, and a dict for un-specialized
+Dispatchers. It too will always return a dict in future.
+
+Recommendations
+---------------
+
+Update calls to these methods such that:
+
+- They are always called when device for which their output is required is in
+  the current CUDA context.
```
stuartarchibald, line 284:
```suggestion
- They are always called when the device for which their output is required is
  in the current CUDA context.
```

numba/cuda/compiler.py
```diff
@@ -822,17 +845,24 @@ class Dispatcher(serialize.ReduceMixin):
     Dispatcher objects are not to be constructed by the user, but instead are
     created using the :func:`numba.cuda.jit` decorator.
     '''
-    def __init__(self, func, sigs, targetoptions):
-        super().__init__()
-        self.py_func = func
+
+    # Whether to fold named arguments and default values. default values are
```
stuartarchibald, line 849:
```suggestion
    # Whether to fold named arguments and default values. Default values are
```

esc:
BFID: `numba_smoketest_cuda_yaml_2`

stuartarchibald (commented):
Thanks for doing all the changes and responding to comments. The patch looks good, just a few very minor things to resolve and then I think this is good to go.

esc:
The run on the BF failed across all Pythons on Linux with the following errors:

```
[2021-01-26 11:46:12,005] {docker_operator.py:265} INFO - ======================================================================
[2021-01-26 11:46:12,005] {docker_operator.py:265} INFO - ERROR: test_inspect_sass_deprecations (numba.cuda.tests.cudapy.test_inspect.TestInspect)
[2021-01-26 11:46:12,006] {docker_operator.py:265} INFO - ----------------------------------------------------------------------
[2021-01-26 11:46:12,006] {docker_operator.py:265} INFO - Traceback (most recent call last):
[2021-01-26 11:46:12,006] {docker_operator.py:265} INFO - File "/opt/conda/envs/testenv_a7c3961d-9241-4d99-b310-bb0ca22205eb/lib/python3.6/site-packages/numba/cuda/compiler.py", line 141, in disassemble_cubin
[2021-01-26 11:46:12,006] {docker_operator.py:265} INFO - stderr=subprocess.PIPE)
[2021-01-26 11:46:12,006] {docker_operator.py:265} INFO - File "/opt/conda/envs/testenv_a7c3961d-9241-4d99-b310-bb0ca22205eb/lib/python3.6/subprocess.py", line 423, in run
[2021-01-26 11:46:12,007] {docker_operator.py:265} INFO - with Popen(*popenargs, **kwargs) as process:
[2021-01-26 11:46:12,007] {docker_operator.py:265} INFO - File "/opt/conda/envs/testenv_a7c3961d-9241-4d99-b310-bb0ca22205eb/lib/python3.6/subprocess.py", line 729, in __init__
[2021-01-26 11:46:12,007] {docker_operator.py:265} INFO - restore_signals, start_new_session)
[2021-01-26 11:46:12,007] {docker_operator.py:265} INFO - File "/opt/conda/envs/testenv_a7c3961d-9241-4d99-b310-bb0ca22205eb/lib/python3.6/subprocess.py", line 1364, in _execute_child
[2021-01-26 11:46:12,007] {docker_operator.py:265} INFO - raise child_exception_type(errno_num, err_msg, err_filename)
[2021-01-26 11:46:12,008] {docker_operator.py:265} INFO - FileNotFoundError: [Errno 2] No such file or directory: 'nvdisasm': 'nvdisasm'
[2021-01-26 11:46:12,008] {docker_operator.py:265} INFO - 
[2021-01-26 11:46:12,008] {docker_operator.py:265} INFO - During handling of the above exception, another exception occurred:
[2021-01-26 11:46:12,008] {docker_operator.py:265} INFO - 
[2021-01-26 11:46:12,009] {docker_operator.py:265} INFO - Traceback (most recent call last):
[2021-01-26 11:46:12,009] {docker_operator.py:265} INFO - File "/opt/conda/envs/testenv_a7c3961d-9241-4d99-b310-bb0ca22205eb/lib/python3.6/site-packages/numba/cuda/tests/cudapy/test_inspect.py", line 184, in test_inspect_sass_deprecations
[2021-01-26 11:46:12,009] {docker_operator.py:265} INFO - f.inspect_sass(compute_capability=self.cc)
[2021-01-26 11:46:12,009] {docker_operator.py:265} INFO - File "/opt/conda/envs/testenv_a7c3961d-9241-4d99-b310-bb0ca22205eb/lib/python3.6/site-packages/numba/cuda/compiler.py", line 1188, in inspect_sass
[2021-01-26 11:46:12,009] {docker_operator.py:265} INFO - return next(iter(self.overloads.values())).inspect_sass()
[2021-01-26 11:46:12,010] {docker_operator.py:265} INFO - File "/opt/conda/envs/testenv_a7c3961d-9241-4d99-b310-bb0ca22205eb/lib/python3.6/site-packages/numba/cuda/compiler.py", line 634, in inspect_sass
[2021-01-26 11:46:12,010] {docker_operator.py:265} INFO - return self._func.get_sass()
[2021-01-26 11:46:12,010] {docker_operator.py:265} INFO - File "/opt/conda/envs/testenv_a7c3961d-9241-4d99-b310-bb0ca22205eb/lib/python3.6/site-packages/numba/cuda/compiler.py", line 466, in get_sass
[2021-01-26 11:46:12,010] {docker_operator.py:265} INFO - return disassemble_cubin(self.cubins[device.id])
[2021-01-26 11:46:12,010] {docker_operator.py:265} INFO - File "/opt/conda/envs/testenv_a7c3961d-9241-4d99-b310-bb0ca22205eb/lib/python3.6/site-packages/numba/cuda/compiler.py", line 148, in disassemble_cubin
[2021-01-26 11:46:12,011] {docker_operator.py:265} INFO - raise RuntimeError(msg)
[2021-01-26 11:46:12,011] {docker_operator.py:265} INFO - RuntimeError: nvdisasm is required for SASS inspection, and has not been found.
[2021-01-26 11:46:12,011] {docker_operator.py:265} INFO - 
[2021-01-26 11:46:12,011] {docker_operator.py:265} INFO - You may need to install the CUDA toolkit and ensure that it is available on your PATH.
[2021-01-26 11:46:12,012] {docker_operator.py:265} INFO - 
[2021-01-26 11:46:12,012] {docker_operator.py:265} INFO - 
[2021-01-26 11:46:12,012] {docker_operator.py:265} INFO - ----------------------------------------------------------------------
[2021-01-26 11:46:12,012] {docker_operator.py:265} INFO - Ran 1114 tests in 209.488s
[2021-01-26 11:46:12,012] {docker_operator.py:265} INFO - 
[2021-01-26 11:46:12,013] {docker_operator.py:265} INFO - FAILED (errors=1, skipped=16, expected failures=1)
```

Is this perhaps a package that will need to installed on the host?

esc:
Windows was fine,  in case you are wondering.

gmarkall:
@esc the nvdisasm checks should have been skipped if nvdisasm isn't available, so I guess I need to work out why.

esc:
@gmarkall is that something I can add on Linux, a package perhaps?

esc:
This one perhaps:

```
cuda-nvdisasm-11-2/unknown,unknown 11.2.67-1 amd64
  CUDA disassembler
```

gmarkall:
@esc perhaps... however, since you've caught this issue, I suspect it would be better to leave it not installed, so that we catch this kind of mistake in future.

esc:
@gmarkall sounds good! Do let me know when this PR is ready for  a re-run.

gmarkall:
@stuartarchibald Many thanks for spotting those errors and omissions - they should now be resolved.

stuartarchibald (approved):
@gmarkall Many thanks for working on this really challenging change. It's great to see the CUDA dispatcher take a step towards the CPU dispatcher. Thanks also for your persistence in fixing and meticulously correcting all the items raised. From a patch point of view, I'm happy with this change. This needs the build farm to verify and @sklam may also wish to take a look.

esc:
@gmarkall thanks for fixing this up, I've re-triggered the linux portion of the build farm run again with the latest changes.

gmarkall:
@esc many thanks - this doesn't need 11.2, so it should also be OK on Windows.

esc:
Screen Shot 2021-01-26 at 16 06 39

All tests green! Including CUDA 11.0 and Python 3.9 combinations.

stuartarchibald:
> Screen Shot 2021-01-26 at 16 06 39
> 
> All tests green! Including CUDA 11.0 and Python 3.9 combinations.

Thanks @esc !

sklam (approved):

PR #6437, opened 2020-10-30
Don't pass reduction variables known in an outer parfor to inner parfors when analyzing reductions.

### LINKED ISSUE #6414: After updating the Numba from 0.43 to 0.51.2, the parallel code is displaying the error
I have recently updated Numba in Anaconda at my machine from 0.43 to 0.51.2. The code which was running successfully with parallel=True is not running with prange. When I have replaced prange with range, the code executes but sequentially. Also, the sequential execution is taking much greater time with parallel=True in comparison to parallel=False.

The error message that I am getting in correspondence to my my program using parallel=True and prange is
```
TypingError: Internal error at <numba.core.typeinfer.CallConstraint object at 0x000001F463923C18>.
Failed in nopython mode pipeline (step: convert to parfors)
maximum recursion depth exceeded while calling a Python object
During: resolving callee type: type(CPUDispatcher(<function case1_flagi1_j1 at 0x000001F462EBC620>))

```

DrTodd13:
Resolves #6414

### DIFF
--- a/numba/tests/test_parfors.py
+++ b/numba/tests/test_parfors.py
@@ -2343,6 +2343,33 @@ class TestPrange(TestPrangeBase):
                'operators.')
         self.assertIn(msg, str(raises.exception))
 
+    @skip_parfors_unsupported
+    def test_prange_two_conditional_reductions(self):
+        # issue6414
+        def test_impl():
+            A = B = 0
+            for k in range(1):
+                if k == 2:
+                    A += 1
+                else:
+                    x = np.zeros((1, 1))
+                    if x[0, 0]:
+                        B += 1
+            return A, B
+        self.prange_tester(test_impl)
+
+    @skip_parfors_unsupported
+    def test_prange_nested_reduction1(self):
+        # issue6414
+        def test_impl():
+            A = 0
+            for k in range(1):
+                for i in range(1):
+                    if i == 0:
+                        A += 1
+            return A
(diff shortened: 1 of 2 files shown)


numba/tests/test_parfors.py
```diff
@@ -2343,6 +2343,33 @@ def test_impl(n):
...
+    def test_prange_nested_reduction1(self):
+        # issue6414
+        def test_impl():
+            A = 0
+            for k in range(1):
+                for i in range(1):
+                    if i == 0:
+                        A += 1
+            return A
+        self.prange_tester(test_impl)
```
stuartarchibald, line 2371: Does this reproduce this issue? I've tried:
```python
import numpy as np

@njit(parallel=True)
def test_impl():
    A = 0
    for k in prange(1):
        for i in prange(1):
            if i == 0:
                A += 1
    return A

test_impl()
```
locally on mainline and it doesn't seem to trip it?

DrTodd13: This was something else that I thought should be tested even if it wasn't triggered by the original error.  I should remove the issue number from that test.

stuartarchibald: ah ok, it worked on mainline and it also worked with this patch, but reduction in nested parfors seems like a good thing to have a test for, happy to leave in with the issue number removed, thanks.

stuartarchibald (approved):

PR #6441, opened 2020-11-02
Fix ASCII flag in Unicode slicing (0.52.0rc2 regression)

ehsantn:
All characters should be ASCII for a string to be ASCII, so the flag computation should use 'and' instead of 'or'.

### DIFF
--- a/numba/tests/test_unicode.py
+++ b/numba/tests/test_unicode.py
@@ -1302,10 +1302,10 @@ class TestUnicode(BaseTest):
         """
         @njit
         def f(s):
-            return s[::2]._is_ascii
+            return s[::2]._is_ascii, s[1::2]._is_ascii
 
         s = "¿abc¡Y tú, quién te cre\t\tes?"
-        self.assertEqual(f(s), 0)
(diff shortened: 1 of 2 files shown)


numba/tests/test_unicode.py
```diff
@@ -1295,6 +1295,18 @@ def test_slice3_error(self):
...
                                          "'%s'[%d:%d:%d]?" % (s, i, j, k))
 
+    def test_slice_ascii_flag(self):
+        """
+        Make sure ascii flag is False when ascii and non-ascii characters are
+        mixed in output of Unicode slicing.
+        """
+        @njit
+        def f(s):
+            return s[::2]._is_ascii
```
stuartarchibald, line 1305:
```suggestion
            return s[::2]._is_ascii, s[1::2]._is_ascii
```
Suggest adding and testing the converse too.

stuartarchibald (commented):
Thanks for the patch, one minor testing thing to resolve else looks good.

ehsantn:
Thanks. Done.

stuartarchibald (approved):
Thanks for the patch and fix up.

PR #6408, opened 2020-10-22
Don't do array analysis for tuples that contain arrays.

DrTodd13:
Resolves #6399

PR #6442, opened 2020-11-02
Fix array analysis regression in 0.52 RC2 for tuple of 1D arrays

ehsantn:
Enable array analysis for 1D arrays in tuples (the check in #6408 was too restrictive).


DrTodd13 (approved):
Looks good for a short-term fix to accommodate 1D arrays.

stuartarchibald:
Thanks for the patch @ehsantn thanks for reviewing @DrTodd13.

PR #6445, opened 2020-11-02
option nogil=True not required for ufuncs and gufuncs

### LINKED ISSUE #1317: Support nogil=True for ufuncs and gufuncs
```
In [3]: @numba.vectorize(nogil=True)
   ...: def f(a, b):
   ...:     return a + b
   ...:

In [4]: f(1, 2)
KeyError: "Does not support option: 'nogil'"
```

This would be extremely useful for integrating numba ufuncs with dask.array.

pradkrish:
Closes #1317  
nogil=True option is not required for ufuncs and gufuncs. Print a UserWarning if it is provided.

### DIFF
--- a/numba/np/ufunc/ufuncbuilder.py
+++ b/numba/np/ufunc/ufuncbuilder.py
@@ -2,6 +2,7 @@
 
 import inspect
 from contextlib import contextmanager
+import warnings
 
 from numba.core.decorators import jit
 from numba.core.descriptors import TargetDescriptor
@@ -82,6 +83,10 @@ class UFuncDispatcher(serialize.ReduceMixin):
         topt.update(targetoptions)
 
         flags = compiler.Flags()
+        if 'nogil' in topt:
+            warnings.warn("nogil option is not required, ufuncs and gufuncs "
+                          "always releases GIL")
+            topt.pop('nogil')
         self.targetdescr.options.parse_as_flags(flags, topt)
 
         flags.set("no_cpython_wrapper")
--- a/numba/np/ufunc/ufuncbuilder.py
+++ b/numba/np/ufunc/ufuncbuilder.py
@@ -85,7 +85,7 @@ class UFuncDispatcher(serialize.ReduceMixin):
         flags = compiler.Flags()
         if 'nogil' in topt:
             warnings.warn("nogil option is not required, ufuncs and gufuncs "
-                          "always releases GIL")
+                          "always releases GIL for CPU target")
             topt.pop('nogil')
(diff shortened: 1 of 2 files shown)


gmarkall (changes requested):
CUDA ufuncs don't release the GIL, for various reasons (some Python code needed around to setup arguments, kernel launch asynchrony, etc.). When trying to use `nogil=True` an error is produced. For example:

```python
@vectorize(nogil=True, target='cuda')
def f(x, y):
    return x + y
```

Gives:

```
KeyError: "cuda vectorize target does not support option: 'nogil'"
```

This is all fine with the functionality in this PR: However, I think the message in the warning is a little misleading - perhaps "ufuncs and gufuncs always release the GIL with the CPU target" would be more accurate?

pradkrish:
> This is all fine with the functionality in this PR: However, I think the message in the warning is a little misleading - perhaps "ufuncs and gufuncs always release the GIL with the CPU target" would be more accurate?

Thanks for the review, I have modified the warning.

numba/np/ufunc/ufuncbuilder.py
```diff
@@ -82,6 +83,10 @@ def compile(self, sig, locals={}, **targetoptions):
         topt.update(targetoptions)
 
         flags = compiler.Flags()
+        if 'nogil' in topt:
+            warnings.warn("nogil option is not required, ufuncs and gufuncs "
+                          "always releases GIL for CPU target")
```
stuartarchibald, line 88: Suggest using `numba.errors.NumbaWarning` as the warning category such that it's easier for users to filter out things from Numba.

stuartarchibald, line 88:
```suggestion
            warnings.warn("The 'nogil' option is not required, ufuncs and gufuncs "
                          "always release the GIL on the CPU target")
```

numba/tests/test_warnings.py
```diff
@@ -204,6 +205,23 @@ def test_disable_performance_warnings(self):
         out, err = popen.communicate()
         self.assertEqual(popen.returncode, not_found_ret_code)
 
+    def test_warning_vectorize_with_nogil(self):
+        @vectorize(nogil=True)
```
stuartarchibald, line 209: Would be good to check `guvectorize` as well given the error message mentions it?

stuartarchibald (commented):
Thanks for the PR, just a couple of minor things to resolve else good to go! Thanks again!

esc:
@pradkrish just enquiring about the state of this PR, where is it at?

PR #6447, opened 2020-11-03
CUDA: Add get_regs_per_thread method to Dispatcher

### LINKED ISSUE #3795: Cuda.jit does not seem to respect max_registers keyword
numba=0.42.0=py37h962f231_0

I have been running into problems with the number of registers in the kernel limiting the number of threads so I have tried using max_registers to clamp the number of registers but it seems to have no effect.

Example of how I'm using it in a decorator:
```
@cuda.jit('void(signature-omitted-for-clarity)' , max_registers=40)   
def foo( code omitted here for clarity):  
```
**expected outcome**
I expect that if I set max_registers=40 then the kernel should not use more than 40 registers.

**actual outcome**
I noticed the register problem because at run time I will get an out-of-resources error if the number of threads exceeds some value less than the SM thread capacity of 1024, and this value decreases as the kernel code gets more complicated, indicating a register count problem. But  Changing the max_registers setting has no effect on the out-of-resources error.  So I dug a little deeper...

**regression**
I inspected the ptax info to see how many registers were in use as follows:

```
with open('/tmp/info.ptx','w') as out:
    print(foo._func.get_info(),file=out)
```

the output seems to tell me that 64 registers were used despite the max_registers keyword used in the decorator.
```
cat /tmp/info.ptx 
ptxas info    : 28 bytes gmem
ptxas info    : Compiling entry function '_ZN6cudapy8__main__35cu_u_driver_global_funcs_col_py$242E5ArrayIfLi2E1A7mutable7alignedE5ArrayIfLi2E1A7mutable7alignedE5ArrayIfLi2E1A7mutable7alignedE5ArrayIfLi2E1A7mutable7alignedE5ArrayIfLi2E1A7mutable7alignedEff5ArrayIfLi2E1A7mutable7alignedE' for 'sm_75'
ptxas info    : Function properties for _ZN6cudapy8__main__35cu_u_driver_global_funcs_col_py$242E5ArrayIfLi2E1A7mutable7alignedE5ArrayIfLi2E1A7mutable7alignedE5ArrayIfLi2E1A7mutable7alignedE5ArrayIfLi2E1A7mutable7alignedE5ArrayIfLi2E1A7mutable7alignedEff5ArrayIfLi2E1A7mutable7alignedE
ptxas         .     0 bytes stack frame, 0 bytes spill stores, 0 bytes spill loads
ptxas info    : Used 64 registers, 792 bytes cmem[0]
info    : 40 bytes gmem
info    : Function properties for '_ZN6cudapy8__main__35cu_u_driver_global_funcs_col_py$242E5ArrayIfLi2E1A7mutable7alignedE5ArrayIfLi2E1A7mutable7alignedE5ArrayIfLi2E1A7mutable7alignedE5ArrayIfLi2E1A7mutable7alignedE5ArrayIfLi2E1A7mutable7aligned
```

**Code to reproduce**
```
import numpy as np
from numba import cuda

@cuda.jit('void(float32[:,:], float32[:,:], float32[:,:],float32[:,:],float32[:,:],float32,float32,float32[:,:])' , max_registers=40)        
def  foo(h, n, u, v, f, dx, dy, out_u):
    sj,si  = cuda.grid(2)  # cuda.x cuda.y skow,fast
    
    # the following dummy code is just here to do anything that is
    # going to create a a kernel using a bunch of registers
    # the code itself is irrelevant I believe.
    
    
    gj = sj
    gi = 31 #out_u.shape[0]-2
    if  (gj < out_u.shape[1]-2) and gj>0 :
        #for gi in range(1, out_u.shape[0]-2):    
        while gi>1:
                 gi=gi-1
                 
              
                 tmp=np.float32(0.0)
                 j = gj
                 i = gi
                 p5 = np.float32(0.5)
                 last_v_n_i = (v[2,j]+v[2,5])*p5
                 v_n = (v[i,j]+v[i,5])*p5  # stradles n[i,j]
                 coriolis_u = (f[99,j]*last_v_n_i+f[i,j]*v_n)*p5 # coriolis force F is n-centered
                 tmp+= coriolis_u
                 out_u[gi,gj] = tmp #funcs(si,sj,h, n, u, v, f, dx, dy)


with open('/tmp/info.ptx','w') as out:
    print(foo._func.get_info(),file=out)
```
To see output look in the file it writes in /tmp/info.ptx

### LINKED ISSUE #3798: Simplify and expose maximum register use in code in cuda.kernel
**Problem being addressed**

- invoking a kernel for which number_registers*number_threads > 65K (or whatever the cuda capacity is for the device) will always error with a cryptic "out of resources" cuda error.

This can be handled more gracefully since the inevitable train wreck is easily detected when one calls the kernel function at runtime.  Just check the above inequality and then give a warning that the register count requires a lower thread count to run.

**Going beyond this**
There are three further steps one, small and one large that would greatly improve refactoring cuda code to get better occupancy by reducing register use.

First, document the command _func.get().attr which is a way for the user to get the register count and thus understand where the resources are going.

Second, it appears to me that the kernel keyword max_registers does not actually work to limit the max_registers used.  I filed a bug report on this.  But if I'm mistaken on it's proper use then better documents would help that too.

**The Big one that would be really nice**
Third, Is there some way to tell what lines of code are requiring registers?  Code seems to use far more registers than one can easily guess by looking at it.  for example:
https://gist.github.com/sklam/0e750e0dea7571c68e94d99006ae8533

I would suggest that a  really really nice interface for tracking registers would be to build it into the .inspect_types output.    In addition to listing a type one could also add the tag  ":: K registers"  where K is the number of registers the line of code is causing to be used.

gmarkall:
Obtaining the register usage per thread shouldn't require the user to dive behind a private interface (`_Kernel._func`). This commit adds `Dispatcher.get_regs_per_thread()` to provide the information from a public API.

There are other attributes that might be nice to provide - shared memory per block, local memory per thread, const memory usage, maximum block size. I will create a "good first issue" to note the possibility to add these too.

Fixes #3798.
Fixes #3795.


gmarkall:
I've just added another commit with a note about the `max_registers` kwarg, as it is closely related to the change already in this PR.

stuartarchibald:
Q. Is there benefit in just exposing the `FuncAttr` named tuple on the `numba.cuda.cudadrv.driver.Function` instance to cover all the properties in there?

gmarkall:
> Q. Is there benefit in just exposing the `FuncAttr` named tuple on the `numba.cuda.cudadrv.driver.Function` instance to cover all the properties in there?

It could be done. I went this route because I felt it gives the opportunity to present a nicer API through functions of the `Dispatcher`.

gmarkall:
(The fail is just a Conda HTTP gateway timeout)

numba/cuda/compiler.py
```diff
@@ -579,6 +579,10 @@ def device(self):
         """
         return get_current_device()
 
+    @property
+    def regs_per_thread(self):
```
stuartarchibald, line 583: Docstr?

numba/cuda/compiler.py
```diff
@@ -919,6 +923,25 @@ def _func(self, signature=None, compute_capability=None):
         else:
             return {sig: defn._func for sig, defn in self.definitions.items()}
 
+    def get_regs_per_thread(self, signature=None):
+        '''
+        Returns the number of registers used by each thread in this kernel for
+        the device in the current context.
+
+        :param sig: The signature of the compiled kernel to get register usage
```
stuartarchibald, line 931:
```suggestion
        :param signature: The signature of the compiled kernel to get register usage
```

numba/cuda/compiler.py
```diff
@@ -919,6 +923,25 @@ def _func(self, signature=None, compute_capability=None):
...
+                 kernel for the given signature and current device.
+        '''
+        cc = get_current_device().compute_capability
+        if signature is not None:
+            return self.definitions[(cc, signature.args)].regs_per_thread
+        if self.specialized:
+            return self.definition.regs_per_thread
+        else:
+            return {sig: defn.regs_per_thread
+                    for sig, defn in self.definitions.items()}
```
stuartarchibald, line 943: Is this branch tested?

stuartarchibald (commented):
Thanks for the patch, changes look good, just few minor things to resolve.

numba/cuda/compiler.py
gmarkall: Added.

numba/cuda/compiler.py
gmarkall: Fixed.

numba/cuda/compiler.py
gmarkall: I've added a test of this branch - it wasn't tested previously.

gmarkall:
@stuartarchibald Many thanks for the review - comments should be addressed now.

stuartarchibald (approved):
Thanks for the fixes.

stuartarchibald:
Not sure what the CI failure is about. Typeguard is unhappy with something, I guess the test shuffling just so happens to have made it such that something that is either wrong or annotated wrong has been hit for this PR. I'm ignoring it as it's `jitclass` related and this PR has nothing to do with that.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_141`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_141`.

Passed.

PR #6454, opened 2020-11-05
FIX: Fixes for literals

larsoner:
Takes care of some warnings of the form:
```
../numba/numba/core/ir.py:29
  /home/larsoner/python/numba/numba/core/ir.py:29: DeprecationWarning: invalid escape sequence \s
    _defmatcher = re.compile('def\s+(\w+)\(.*')
```


stuartarchibald (approved):
Thanks for the patch and for fixing this!

PR #6455, opened 2020-11-05
Fix refprune on obfuscated refs and stabilize optimisation WRT wrappers.

stuartarchibald:
Also makes more things vectorize. As title.

### DIFF
--- a/numba/core/callconv.py
+++ b/numba/core/callconv.py
@@ -2,7 +2,7 @@
 Calling conventions for Numba-compiled functions.
 """
 
-from collections import namedtuple
+from collections import namedtuple, Iterable
 import itertools
 
 from llvmlite import ir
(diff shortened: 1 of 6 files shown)


stuartarchibald:
This needs an updated llvmlite.

stuartarchibald:
Depends on https://github.com/numba/llvmlite/pull/647

numba/core/callconv.py
```diff
@@ -561,7 +568,18 @@ def call_function(self, builder, callee, resty, argtys, args):
...
         realargs = [retvaltmp, excinfoptr] + args
-        code = builder.call(callee, realargs)
+        # deal with attrs, it's find to specify a load in a string like
+        # "noinline fast" as per LLVM or equally as an iterable of individual
+        # attributes.
+        if attrs is None:
+            _attrs = ()
+        elif isinstance(attrs, str):
+            _attrs = tuple(attrs.split(' '))
+        elif isinstance(attrs, Iterable):
```
sklam, line 578: This is never tested.

stuartarchibald: I can remove, though it just feels a bit strange API wise writing:
```
call_function(self, builder, callee, resty, argtys, args, attrs="noinline nounwind"):
```
vs.
```
call_function(self, builder, callee, resty, argtys, args, attrs=["noinline", "nounwind"]):
```
but I think testing the latter might be hard.

sklam: Agree that the latter is nicer. And, I'd rather we have only one way of spelling it, even if we have to say `attrs=['noinline']`

sklam:
Need to bump llvmlite upper limit

sklam:
Some tests are not vectorising again

numba/core/callconv.py
stuartarchibald: I've made a change along these lines, see what you think.

sklam:
Looks like whack-a-mole is done

sklam (approved):
Great that we will get back the RC1 speed.

Issue #6459, opened 2020-11-08
ByteArray doesn't support boxing or slicing


njriasan:
When working with a byte array type it seems that unboxing is supported, but boxing is not supported. Here is an example reproducer.

```python
import numba

arr = bytearray([1, 2, 3, 4])
print(numba.njit(lambda x: x)(arr))
```

with output:
```python
Traceback (most recent call last):
  File "test_bytearray.py", line 4, in <module>
    print(numba.njit(lambda x: x)(arr))
TypeError: cannot convert native bytearray(uint8, 1d, C) to Python object
```

In addition ByteArrays doesn't seems to support getitem with slices. Here is the reproducer:
```python
import numba

arr = bytearray([1, 2, 3, 4])
print(numba.njit(lambda x: x)(arr))
```

with output:
```python
Traceback (most recent call last):
  File "test_bytearray.py", line 4, in <module>
    print(numba.njit(lambda x: x[1:3])(arr))
  File "/Users/nicholasriasanovsky/Documents/bodo/numba/numba/core/dispatcher.py", line 415, in _compile_for_args
    error_rewrite(e, 'typing')
  File "/Users/nicholasriasanovsky/Documents/bodo/numba/numba/core/dispatcher.py", line 358, in error_rewrite
    reraise(type(e), e, None)
  File "/Users/nicholasriasanovsky/Documents/bodo/numba/numba/core/utils.py", line 80, in reraise
    raise value.with_traceback(tb)
numba.core.errors.TypingError: Failed in nopython mode pipeline (step: nopython frontend)
No implementation of function Function(<built-in function getitem>) found for signature:
 
 >>> getitem(bytearray(uint8, 1d, C), slice<a:b>)
 
There are 22 candidate implementations:
  - Of which 22 did not match due to:
  Overload of function 'getitem': File: <numerous>: Line N/A.
    With argument(s): '(bytearray(uint8, 1d, C), slice<a:b>)':
   No match.

During: typing of intrinsic-call at test_bytearray.py (4)
During: typing of static-get-item at test_bytearray.py (4)

File "test_bytearray.py", line 4:
arr = bytearray([1, 2, 3, 4])
print(numba.njit(lambda x: x[1:3])(arr))
```

This appears to be because the getitem overload for buffer types works with borrowed references and ByteArray has `slice_is_copy` set to True. As a result `getitem_arraynd_intp` is never entered, so I assume this requires a separate overload that makes a fresh copy.

sklam:
[...]

PR #6460, opened 2020-11-08
Added Bytearray boxing and fixed memory leak in Bytes boxing

njriasan:
Adds support for boxing of byte arrays. Addresses unboxing from #6459, but doesn't include slicing. I can add support for that if I have some feedback on what will need to be implemented.

I'm leaving this as WIP because I am unsure where the unittest belongs for this. I found a file dedicated to buffer protocols, but those didn't seem to include boxing.

### DIFF
--- a/numba/core/boxing.py
+++ b/numba/core/boxing.py
@@ -262,13 +262,17 @@ def unbox_unicodecharseq(typ, obj, c):
 @box(types.Bytes)
 def box_bytes(typ, val, c):
     obj = c.context.make_helper(c.builder, typ, val)
-    return c.pyapi.bytes_from_string_and_size(obj.data, obj.nitems)
+    ret = c.pyapi.bytes_from_string_and_size(obj.data, obj.nitems)
+    c.context.nrt.decref(c.builder, typ, val)
+    return ret
 
 
 @box(types.ByteArray)
 def box_bytearray(typ, val, c):
     obj = c.context.make_helper(c.builder, typ, val)
-    return c.pyapi.bytearray_from_string_and_size(obj.data, obj.nitems)
+    ret = c.pyapi.bytearray_from_string_and_size(obj.data, obj.nitems)
+    c.context.nrt.decref(c.builder, typ, val)
+    return ret
 
 
 @box(types.CharSeq)
--- a/numba/core/boxing.py
+++ b/numba/core/boxing.py
@@ -265,6 +265,12 @@ def box_bytes(typ, val, c):
     return c.pyapi.bytes_from_string_and_size(obj.data, obj.nitems)
 
 
+@box(types.ByteArray)
+def box_bytearray(typ, val, c):
+    obj = c.context.make_helper(c.builder, typ, val)
+    return c.pyapi.bytearray_from_string_and_size(obj.data, obj.nitems)
+
+
 @box(types.CharSeq)
 def box_charseq(typ, val, c):
     rawptr = cgutils.alloca_once_value(c.builder, value=val)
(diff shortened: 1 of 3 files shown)


stuartarchibald:
> Adds support for boxing of byte arrays. Addresses unboxing from #6459, but doesn't include slicing. I can add support for that if I have some feedback on what will need to be implemented.

Thanks for the patch. I suggest getting the boxing working, tested, and into mainline and then opening another PR for slicing.

> I'm leaving this as WIP because I am unsure where the unittest belongs for this. I found a file dedicated to buffer protocols, but those didn't seem to include boxing.

Suggest adding it there, boxing tests tend to be in the test for the type. There's not a huge number of them as round tripping tests this to some degree.

njriasan:
Thanks for your help @stuartarchibald. I added a test, but for both Bytes (which I didn't modify) and ByteArray there appears to be a memory leak resulting from the process of boxing/unboxing. I assume this is either a missing decref or an error with my test, but I don't have any experience working with memory leaks in Numba. Could you give me some advice on where to look or how to debug this issue. Thanks :)

stuartarchibald:
@njriasan thanks for the update. Docs on debugging the Numba runtime are here: https://numba.readthedocs.io/en/stable/developer/numba-runtime.html#debugging-leaks hope these help.

numba/core/boxing.py
```diff
@@ -265,6 +265,12 @@ def box_bytes(typ, val, c):
     return c.pyapi.bytes_from_string_and_size(obj.data, obj.nitems)
 
 
+@box(types.ByteArray)
```
stuartarchibald, line 268: Would guess that this and `@box(types.Bytes)` need to look something like:
```python
@box(<bytes-like-type>)
def box_bytes(typ, val, c):
    obj = c.context.make_helper(c.builder, typ, val)
    ret = c.pyapi.<bytes-like ctor call>(obj.data, obj.nitems)
    return ret
```

njriasan: This was 100% the issue. Thank you so much.

njriasan:
Ok this should be fixed now. Thank you for all of your help.

njriasan:
Hi @stuartarchibald is there any update on this? I understand this is probably low priority, but this PR seems simple and I suspect won't take more than 5 minutes to review.

esc (approved):
Thank you for this PR! I had a look at it today and it seems fine. I can confirm the tests run and test that bytearray can be boxed. I also checked that the two `decref` statements are necessary by commenting them out locally and checking that the `MemoryLeakMixin` does indeed detect memory leaks. Is good to merge from my side.

esc:
@stuartarchibald I marked it as ready to merge, hope that is OK? If you find anything that I missed, please feel free to change the label again, thanks!

numba/core/boxing.py
```diff
@@ -262,7 +262,17 @@ def unbox_unicodecharseq(typ, obj, c):
...
+    c.context.nrt.decref(c.builder, typ, val)
+    return ret
+
+
+@box(types.ByteArray)
+def box_bytearray(typ, val, c):
+    obj = c.context.make_helper(c.builder, typ, val)
+    ret = c.pyapi.bytearray_from_string_and_size(obj.data, obj.nitems)
+    c.context.nrt.decref(c.builder, typ, val)
+    return ret
```
stuartarchibald, line 275: Need to consider how to handle "reflection".
```
from numba import njit
import numpy as np

def foo(x):
    x[0] = 0xff
    return x

for f in foo, njit(foo):
    tmp = bytearray(b"abc")
    print(tmp)
    ret = f(tmp)
    print(tmp)
    print(ret is tmp)
```

esc:
@stuartarchibald that is a good find. Since `bytearray` is mutable, any modifications that happen to the unboxed variant must be "refelected" back when unboxing.

esc:
@njriasan -- may I kindly request an update on this PR?

njriasan:
@esc Sure I'll make an update this weekend. Thank you for the ping.

PR #6466, opened 2020-11-12
Remove the misspelling of finalize_dynamic_globals

PokhodenkoSA:
This commit fixes misspelling of _finalize_dynamic_globals().
This function is used only in the same file.


esc:
@PokhodenkoSA looks good, thanks!

PR #6467, opened 2020-11-12
Access `ufunc_db` via context

PokhodenkoSA:
`ufunc_db` is global variable from `numba.np`.
It is necessary to have possibility to modify `ufunc_db`
depending on context.

This PR is a proposal for starting a discussion.

### DIFF
--- a/numba/np/npyimpl.py
+++ b/numba/np/npyimpl.py
@@ -448,7 +448,12 @@ def _ufunc_db_function(ufunc):
             super(_KernelImpl, self).__init__(context, builder, outer_sig)
             loop = ufunc_find_matching_loop(
                 ufunc, outer_sig.args + tuple(_unpack_output_types(ufunc, outer_sig)))
-            self.fn = ufunc_db.get_ufunc_info(ufunc).get(loop.ufunc_sig)
+
+            if hasattr(context, 'ufunc_db'):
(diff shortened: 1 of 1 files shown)


sklam:
We are working on allowing multiple `lower_builtin` registries, one per target. The `ufunc_db` is currently a dictionary but can also be re-fitted into a registry. So that `ufunc_db` and other `lower_builtin` can share the same code for multi-target.

reazulhoque:
Hi @sklam @stuartarchibald, this PR seems good to go. Can you please review it and let me know if that is the case?

gmarkall:
@reazulhoque Given the discussion above, it's not clear that this is ready to go - it looks like the work in https://github.com/numba/numba/pull/6598 is aimed towards supporting the use-case outlined in the PR description.

sklam:
Given dppy backend find that it is useful for the target context to export the ufunc_db, I think other similar backend will find it useful in the future to be able to customize the ufunc implementation. https://github.com/numba/numba/pull/6598 will not be completely replacing the need for this patch in the near term. I think we should even consider moving `ufunc_db` into target context so we don't need a branch and that internal numba code doesn't need to refer to a random global state. 

A quick grep through the repo reveals that there are only a few places `ufunc_db` are used. I think we can easily update how it is used.

PR #6468, opened 2020-11-12
Access `replace_functions_map` via PreParforPass instance

PokhodenkoSA:
`replace_functions_map` is global variable from `numba.parfors.parfor`.
It is necessary to have possibility to modify `replace_functions_map`
depending on PreParforPass instance and access to it via the instance.

This PR is a proposal for starting a discussion.

### DIFF
--- a/numba/parfors/parfor.py
+++ b/numba/parfors/parfor.py
@@ -1350,7 +1350,7 @@ class PreParforPass(object):
     implementations of numpy functions if available.
     """
     def __init__(self, func_ir, typemap, calltypes, typingctx, options,
-                 swapped={}):
+                 swapped={}, replace_functions_map=replace_functions_map):
         self.func_ir = func_ir
         self.typemap = typemap
         self.calltypes = calltypes
@@ -1358,6 +1358,7 @@ class PreParforPass(object):
         self.options = options
         # diagnostics
         self.swapped = swapped
+        self.replace_functions_map = replace_functions_map
         self.stats = {
             'replaced_func': 0,
             'replaced_dtype': 0,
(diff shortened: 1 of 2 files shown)


sklam:
This change will very likely be needed. I cannot think of any better way to specialize the parfor passes.

stuartarchibald:
From out of band discussion, something along the lines of this patch is needed and moving information out of globals is a good idea. Once tests are added this PR can be reviewed. Many thanks.

numba/parfors/parfor.py
```diff
@@ -1350,14 +1350,15 @@ class PreParforPass(object):
     implementations of numpy functions if available.
     """
     def __init__(self, func_ir, typemap, calltypes, typingctx, options,
-                 swapped={}):
+                 swapped={}, replace_functions_map=replace_functions_map):
```
sklam, line 1353: I'd suggest letting it defaulting to `None` and setting to the `replace_functions_map` global dictionary in the function body. With that, the test do not need to know what should be the default dict.

numba/tests/test_parfors_passes.py
```diff
@@ -73,13 +73,16 @@ def _run_parfor(cls, test_func, args):
 
             diagnostics = numba.parfors.parfor.ParforDiagnostics()
 
+            if swap_map is None:
+                swap_map = numba.parfors.parfor.replace_functions_map
```
sklam, line 77: the test should not need to know the default value. see previous comment.

sklam (commented):
Thanks for the patch. I have a few suggestion.

sklam (approved):
LGTM. I restarted CI since it was failing due to internet traffic jam. Can merge once CI reports ok.

sklam:
The 404 error in CI failure did not recover and it is unrelated to this PR. Marking this for merge.

PR #6469, opened 2020-11-12
Add address space in pointer type

PokhodenkoSA:
It is necessary for representing pointers related to different
execution devices (i.e. GPU).

This PR is for starting a discussion.

### DIFF
--- a/numba/core/types/misc.py
+++ b/numba/core/types/misc.py
@@ -144,14 +144,15 @@ class CPointer(Type):
     """
     mutable = True
 
-    def __init__(self, dtype):
+    def __init__(self, dtype, addrspace=None):
         self.dtype = dtype
+        self.addrspace = addrspace
         name = "%s*" % dtype
         super(CPointer, self).__init__(name)
 
     @property
     def key(self):
-        return self.dtype
+        return self.dtype, self.addrspace
 
 
 class EphemeralPointer(CPointer):
--- a/numba/core/types/misc.py
+++ b/numba/core/types/misc.py
@@ -151,7 +151,7 @@ class CPointer(Type):
     def __init__(self, dtype, addrspace=None):
         self.dtype = dtype
         self.addrspace = addrspace
-        if addrspace:
+        if addrspace != None:
             name = "%s_%s*" % (dtype, addrspace)
         else:
             name = "%s*" % dtype
--- a/numba/core/types/misc.py
+++ b/numba/core/types/misc.py
@@ -145,7 +145,7 @@ class CPointer(Type):
     Attributes
     ----------
         dtype : The pointee type
-        addrspace : [int]
+        addrspace : int
             The address space pointee belongs to.
     """
     mutable = True
--- a/numba/core/types/misc.py
+++ b/numba/core/types/misc.py
@@ -142,16 +142,18 @@ class CPointer(Type):
     """
     Type class for pointers to other types.
 
-    Attributes:
-        dtype : The type current pointer represents.
-        addrspace (int): The address space current pointer belongs to.
+    Attributes
+    ----------
+        dtype : The pointee type
+        addrspace : [int]
+            The address space pointee belongs to.
     """
     mutable = True
 
     def __init__(self, dtype, addrspace=None):
         self.dtype = dtype
         self.addrspace = addrspace
-        if addrspace != None:
+        if addrspace is not None:
             name = "%s_%s*" % (dtype, addrspace)
         else:
             name = "%s*" % dtype
--- a/numba/core/types/misc.py
+++ b/numba/core/types/misc.py
@@ -141,13 +141,20 @@ class MemInfoPointer(Type):
 class CPointer(Type):
     """
     Type class for pointers to other types.
+
+    Attributes:
+        dtype : The type current pointer represents.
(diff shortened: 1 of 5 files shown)


sklam:
I think changing `CPointer` make sense. But for `Buffer` and `Array`, can they be extended via subclass. Would it enough equal amount of work to update all code to emit for the correct addrspace vs for using the correct subclass.

reazulhoque:
Hi @sklam I updated the patch according to your suggestion. Please take a look and let me know what changes are required.

numba/core/types/misc.py
```diff
@@ -144,14 +144,15 @@ class CPointer(Type):
     """
     mutable = True
 
-    def __init__(self, dtype):
+    def __init__(self, dtype, addrspace=None):
         self.dtype = dtype
+        self.addrspace = addrspace
         name = "%s*" % dtype
         super(CPointer, self).__init__(name)
```
sklam, line 151: Can you include the addrspace as part of the name? The name is used for symbol mangling. Maybe include the addrspace in name when `addrspace is not None`.

numba/core/types/misc.py
```diff
@@ -144,14 +144,15 @@ class CPointer(Type):
     """
     mutable = True
 
-    def __init__(self, dtype):
+    def __init__(self, dtype, addrspace=None):
```
sklam, line 147: What is the expected type for `addrspace`?

numba/core/types/misc.py
reazulhoque: I will.

numba/core/types/misc.py
reazulhoque: The expected type of `addrspace` is int.

reazulhoque: I added the docstring according to how I know best. I did not find any docstring detailing the other argument `dtype` in the same class, so my addition might not be correct. Please advise if the format or the convention is incorrect.

numba/core/types/misc.py
reazulhoque: Added.

numba/core/types/misc.py
```diff
@@ -141,17 +141,25 @@ def key(self):
...
+        addrspace (int): The address space current pointer belongs to.
     """
     mutable = True
 
-    def __init__(self, dtype):
+    def __init__(self, dtype, addrspace=None):
         self.dtype = dtype
-        name = "%s*" % dtype
+        self.addrspace = addrspace
+        if addrspace != None:
```
sklam, line 154:
```suggestion
        if addrspace is not None:
```

numba/core/types/misc.py
```diff
@@ -141,17 +141,25 @@ def key(self):
...
+            name = "%s_%s*" % (dtype, addrspace)
+        else:
+            name = "%s*" % dtype
```
sklam, line 157: Can you add a simple test so that both branches are exercised?
It can go in https://github.com/numba/numba/blob/5f7bcce048b9c0a7fd02376ca8a2e3574625c99f/numba/tests/test_types.py#L47

numba/core/types/misc.py
sklam: Format-wise, we haven't defined a styleguide but I prefer the numpydoc style like: https://github.com/numba/numba/blob/5f7bcce048b9c0a7fd02376ca8a2e3574625c99f/numba/core/errors.py#L665-L668
For reference: https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring

numba/core/types/misc.py
```diff
@@ -141,17 +141,25 @@ def key(self):
 class CPointer(Type):
     """
     Type class for pointers to other types.
+
+    Attributes:
+        dtype : The type current pointer represents.
```
sklam, line 146:
```suggestion
        dtype : The pointee type.
```
maybe a better description.

numba/core/types/misc.py
reazulhoque: Changed to fit numpydoc style.

numba/core/types/misc.py
reazulhoque: Test added in the placed in the suggested location.

stuartarchibald:
@reazulhoque thanks for the updates, CI is currently failing due to some long lines tripping up the `flake8` checks. Ref: https://dev.azure.com/numba/numba/_build/results?buildId=8259&view=logs&j=9b1c419d-97bd-5aaf-47ca-b83be1ff9bc1&t=d1ed00e6-cc06-521a-e00c-4f5e4a0f55c5&l=84

numba/core/types/misc.py
```diff
@@ -141,17 +141,27 @@ def key(self):
 class CPointer(Type):
     """
     Type class for pointers to other types.
+
+    Attributes
+    ----------
+        dtype : The pointee type
+        addrspace : [int]
```
sklam, line 148:
```suggestion
        addrspace : int
```
the square bracket from the example (`frozenset[int]`) is for generics only

reazulhoque: Fixed.

sklam (approved):
Thanks for the patch!

PR #6480, opened 2020-11-13
Update changelog for 0.52.0rc3

sklam:
as titled

### DIFF
--- a/CHANGE_LOG
+++ b/CHANGE_LOG
@@ -91,6 +91,8 @@ Performance Enhancements:
 * PR #6275: Switch on SLP vectorize. (Stuart Archibald)
 * PR #6278: Improve typed list performance. (Stuart Archibald)
 * PR #6335: Split optimisation passes. (Stuart Archibald)
+* PR #6455: Fix refprune on obfuscated refs and stabilize optimisation WRT
+  wrappers. (Stuart Archibald)
 
 Fixes:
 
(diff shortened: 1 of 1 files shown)


CHANGE_LOG
```diff
@@ -147,6 +149,13 @@ Fixes:
 * PR #6396: Remove hard dependency on tbb package. (Stuart Archibald)
 * PR #6408: Don't do array analysis for tuples that contain arrays. (Todd A.
   Anderson)
+* PR #6441: Fix ASCII flag in Unicode slicing (0.52.0rc2 regression) (Ehsan
+  Totoni)
```
stuartarchibald, line 153: Ehsan needs to go into Authors.

CHANGE_LOG
```diff
@@ -213,7 +222,8 @@ CI/Infrastructure Updates:
 * PR #6252: Show channel URLs (Siu Kwan Lam)
 * PR #6338: Direct user questions to Discourse instead of the Google Group.
   (Stan Seibert)
-
+* PR #6474: Add skip on PPC64LE for tests causing SIGABRT in LLVM. (Stuart
+  Archibald)
```
stuartarchibald, line 226: Think this perhaps ought to in ``Fixes``?

stuartarchibald (commented):
Thanks for the patch, couple of minor things to address else looks good.

CHANGE_LOG
sklam: oh, i missed that

CHANGE_LOG
sklam: Doesn't really fixes anything though. It's just a change is what is tested. That's why I think it is CI related

CHANGE_LOG
sklam: done

CHANGE_LOG
stuartarchibald: I'm not sure not consistent with the categorisation of similar items, but probably not worth worrying about!

CHANGE_LOG
stuartarchibald: thanks

stuartarchibald (approved):
Thanks for the fix ups, looks good.

PR #6490, opened 2020-11-17
Stencil output allocated with np.empty now and new code to initialize the borders.

### LINKED ISSUE #6456: stencil kernel performance not scaling in njit(parallel=True)
```python
import numba as nb
import numpy as np

@nb.stencil
def kernel1(a):
    return 0.25 * (a[0, 1] + a[1, 0] + a[0, -1] + a[-1, 0])

@nb.njit
def calc1(in_arr):
    out_arr = kernel1(in_arr)
    return out_arr

@nb.njit(parallel= True) 
def calc2(in_arr):
    out_arr = kernel1(in_arr)
    return out_arr

@nb.njit
def calc3(in_arr):
    out_arr = np.zeros(in_arr.shape)
    for i in range(1, in_arr.shape[0]-1):
        for j in range(1, in_arr.shape[1]-1):
            out_arr[i,j] = 0.25 * (in_arr[i, j+1] + in_arr[i+1, j] + in_arr[i, j-1] + in_arr[i-1, j])
    return out_arr

@nb.njit(parallel=True)
def calc4(in_arr):
    out_arr = np.zeros(in_arr.shape)
    for i in range(1, in_arr.shape[0]-1):
        for j in range(1, in_arr.shape[1]-1):
            out_arr[i,j] = 0.25 * (in_arr[i, j+1] + in_arr[i+1, j] + in_arr[i, j-1] + in_arr[i-1, j])
    return out_arr


input_arr = np.linspace(1, 50000, 500000000).reshape((50000, 10000))
```

For a large matrix (`input_arr`), I am experiencing the following run times (avg. 5 runs):
`calc1(input_arr)` [njit + stencil] - `2.09s`
`calc2(input_arr)` [Parallel njit + stencil] - `2.35s`
`calc3(input_arr)` [njit + Loops] - `2.10s`
`calc4(input_arr)` [Parallel njit + Loops] - `1.08s`

As per the above timings, `njit + stencil` is performing similar to `njit + Loops`, as by-default the stencil kernels are not executed in parallel.
I am trying to understand why `Parallel njit + stencil` is not performing similar to `Parallel njit + Loops`?
How can I parallelize the stencil execution?
I have 16 threads available on my machine.

Any solution will be highly appreciated. Thank you!

The first time this issue was mentioned in Github - https://github.com/numba/numba/issues/2982

DrTodd13:
We now use np.empty to initialize stencil output and dynamically generate code to initialize the borders of each dimension.

Resolves #6456


stuartarchibald:
Thanks for the patch @DrTodd13. The CI failures seem genuine, related to stencil tests failing with parfors.

stuartarchibald:
Thanks for the patch. Am wondering if this is going to create a performance disparity between stencil operations in parfors vs njit, as the latter will still use `full`?

DrTodd13:
> Thanks for the patch. Am wondering if this is going to create a performance disparity between stencil operations in parfors vs njit, as the latter will still use `full`?

I did the same for the non-parfor side now.

numba/stencils/stencilparfor.py
```diff
@@ -260,13 +266,142 @@ def _mk_stencil_parfor(self, label, in_args, out_arr, stencil_ir,
...
+                tuple_assign = ir.Assign(tuple_call, start_ind_var, loc)
+                stmts.append(tuple_assign)
+
+                setitem_call = ir.SetItem(out_arr, start_ind_var, zero_var, loc)
+                self.calltypes[setitem_call] = signature(
+                                                types.none, self.typemap[out_arr.name],
+                                                self.typemap[start_ind_var.name],
+                                                self.typemap[out_arr.name].dtype
+                                                )
+                stmts.append(setitem_call)
```
stuartarchibald, line 359: This block seems to be almost identical to the one below, apart from `start_inds` vs `last_inds` and `zero_index_var` vs `si_var`, perhaps make it into a local function and call it twice?

numba/stencils/stencilparfor.py
```diff
@@ -260,13 +266,142 @@ def _mk_stencil_parfor(self, label, in_args, out_arr, stencil_ir,
...
+            # For each dimension, add setitem to set border values.
+            for dim in range(in_arr_typ.ndim):
+                # First, fill all entries with ":".
+                start_tuple_items = [slice_var] * in_arr_typ.ndim
+                last_tuple_items = [slice_var] * in_arr_typ.ndim
+
+                # Handle the border at the start of the index range.
+                # ---- Generate call to slice func.
+                sig = self.typingctx.resolve_function_type(slice_fn_ty,
+                                                           (types.intp,types.intp),
```
stuartarchibald, line 321:
```suggestion
                                                           (types.intp,) * 2,
```

numba/stencils/stencil.py
```diff
@@ -577,17 +577,24 @@ def _stencil_wrapper(self, result, sigret, return_type, typemap, calltypes, *arg
...
             else:
-                out_init ="{} = np.zeros({}, dtype=np.{})\n".format(
-                            out_name, shape_name, return_type_name)
+                 cval = 0
             func_text += "    " + out_init
+            for dim in range(the_array.ndim):
+                start_items = [":"] * the_array.ndim
+                end_items = [":"] * the_array.ndim
+                start_items[dim] = ":-{}".format(self.neighborhood[dim][0])
+                end_items[dim] = "-{}:".format(self.neighborhood[dim][1])
```
stuartarchibald, line 595: this ends up with double negatives but has the desired effect.

numba/stencils/stencilparfor.py
```diff
@@ -260,13 +266,142 @@ def _mk_stencil_parfor(self, label, in_args, out_arr, stencil_ir,
...
-                                       np.full,
+            stmts = ir_utils.gen_np_call("empty",
+                                       np.empty,
                                        out_arr,
-                                       [shape_var, zero_var, dtype_attr_var],
+                                       [shape_var, dtype_attr_var],
                                        self.typingctx,
                                        self.typemap,
                                        self.calltypes)
+            # ------------------
```
stuartarchibald, line 276: I guess all this logic can be reused in the case where a use supplies an output too (not for this PR though)?

stuartarchibald (commented):
Thanks for the patch, changes look good, I've a suggestion to refactor the repeated block into a function but otherwise this is good to go.

animator:
Hi @DrTodd13 and @stuartarchibald, 
Are there any updates on this pull request.
Eagerly waiting for the fix.
Thank you for your time.

DrTodd13:
@stuartarchibald @sklam In stencil.py for this PR around line 610, we check if cval is a precise match for the array's dtype but in the else below that on line 627 we do a looser test to say that a cval is fine if it is convertible to the array's dtype.  I suggest that these tests should be the same, and I prefer the more loose version.  I can make that change if you agree which means that test_basic35 in test_stencils.py will have to go from expecting it to generate an error to not.

numba/stencils/stencil.py
DrTodd13: Yes, double negatives but innocuous.

numba/stencils/stencilparfor.py
DrTodd13: Yes.

DrTodd13:
> @stuartarchibald @sklam In stencil.py for this PR around line 610, we check if cval is a precise match for the array's dtype but in the else below that on line 627 we do a looser test to say that a cval is fine if it is convertible to the array's dtype. I suggest that these tests should be the same, and I prefer the more loose version. I can make that change if you agree which means that test_basic35 in test_stencils.py will have to go from expecting it to generate an error to not.

I went ahead and made that change when called from numba and from python.  I also fixed Stuart's request above about making a function out of the common code.

numba/stencils/stencilparfor.py
DrTodd13: Done.

numba/stencils/stencilparfor.py
```diff
@@ -263,13 +270,133 @@ def _mk_stencil_parfor(self, label, in_args, out_arr, stencil_ir,
...
+                    tuple_assign = ir.Assign(tuple_call, border_ind_var, loc)
+                    stmts.append(tuple_assign)
+
+                    setitem_call = ir.SetItem(out_arr, border_ind_var, zero_var, loc)
+                    self.calltypes[setitem_call] = signature(
+                                                    types.none, self.typemap[out_arr.name],
+                                                    self.typemap[border_ind_var.name],
+                                                    self.typemap[out_arr.name].dtype
+                                                    )
+                    stmts.append(setitem_call)
```
stuartarchibald, line 375: Does this need to be defined inside the loop? If not, suggest moving it out.

stuartarchibald (commented):
Thanks for the update @DrTodd13. I agree with your comment in https://github.com/numba/numba/pull/6490#issuecomment-983003889, seems reasonable to accept convertible types for `cval`. I've one additional query (inline) else looks good.

esc:
`Build numba_smoketest_cpu_yaml_109 has started`

esc:
> `Build numba_smoketest_cpu_yaml_109 has started`

This was 🟢

PR #6492, opened 2020-11-17
Adapt docs: full example does no longer exist

EQt:
I searched for this example notebook. Does anyone know, were they can be found, now?

Future Work:
Maybe extend the documentation on how to create in instance of `types.Record` that can be passed to a `cfunc`

### DIFF
--- a/docs/source/user/cfunc.rst
+++ b/docs/source/user/cfunc.rst
@@ -194,11 +194,6 @@ Due to ABI limitations, structures should be passed as pointers
 using ``types.CPointer(my_struct)`` as the argument type.  Inside the ``cfunc``
 body, the ``my_struct*`` can be accessed with ``carray``.
 
-Full example
-------------
-
-See full example in ``examples/notebooks/Accessing C Struct Data.ipynb``.
-
 
 Signature specification
 =======================


gmarkall:
It is here: https://github.com/numba/numba-examples/blob/master/legacy/notebooks/Accessing%20C%20Struct%20Data.ipynb

gmarkall:
If that notebook still works correctly, it may be worth moving to somewhere else in numba-examples (or fixing up first so that it does work correctly).

EQt:
... and then adapt the link in the `numba/docs` accordingly

EQt:
> It is here: https://github.com/numba/numba-examples/blob/master/legacy/notebooks/Accessing%20C%20Struct%20Data.ipynb

Is there any other way to create an instance apart from `cffi.FFI.new` as shown in `In[2]`?

gmarkall:
> Is there any other way to create an instance apart from `cffi.FFI.new` as shown in `In [2]:` of the above notebook?

You might receive a pointer to a struct from a C function - did you have something else in mind?

EQt:
I had the situation in mind where you create in Python a new struct to be passed to a C function, similarly to
```python
mydata = ffi.new('my_struct')
addr = int(ffi.cast('size_t', ptr))
print("address of data:", hex(addr))
```
using `cffi` in the notebook above. So far I found a way using `numpy.recarrays`. Following the example above that would be
```python
import numpy as np
DType = [
    ('i1', np.int32),
    ('f2', np.float32),
    ('d3', np.float64),
    ('af4', (np.float32, (7,))),
]
mydata = np.zeros(1, dtype=DType)[0]
addr = mydata.base.ctypes.data
```

Other ways might use Python's `struct` module

docs/source/user/cfunc.rst
```diff
@@ -194,11 +194,6 @@ Due to ABI limitations, structures should be passed as pointers
 using ``types.CPointer(my_struct)`` as the argument type.  Inside the ``cfunc``
 body, the ``my_struct*`` can be accessed with ``carray``.
 
-Full example
-------------
-
-See full example in ``examples/notebooks/Accessing C Struct Data.ipynb``.
```
stuartarchibald, line 200: Should this link just be updated to https://github.com/numba/numba-examples/blob/master/legacy/notebooks/Accessing%20C%20Struct%20Data.ipynb ?

EQt: I think it is best to include the documentation directly, at least in the long run. Yet a temporary solution could be just to update the link.

PR #6499, opened 2020-11-18
Testhound/cuda atomic

testhound:
This pull request adds support for atomic increment, decrement, exchange and compare and swap (extended for 64-bit integers).


numba/cuda/tests/cudapy/test_atomics.py
```diff
@@ -715,6 +795,168 @@ def test_atomic_xor_global_2(self):
...
+    def test_atomic_exch(self):
+        rand_const = np.random.randint(50, 100, dtype=np.uint32)
+        ary = np.random.randint(0, 32, size=32).astype(np.uint32)
+        idx = np.arange(32, dtype=np.uint32)
+        gold = ary.copy()
+        gold.fill(rand_const)
+        cuda_func = cuda.jit('void(uint32[:], uint32[:], uint32)')(atomic_exch)
+        cuda_func[1, 32](ary, idx, rand_const)
+
+        np.testing.assert_equal(ary, gold)
```
gmarkall, line 926: Is there a reason not to just check that `ary` is equal to `rand_const`, rather than creating a copy of `ary` and filling it with `rand_const`? (Also applies to the other `atomic_exch` tests)

numba/cuda/tests/cudapy/test_atomics.py
```diff
@@ -876,25 +1118,108 @@ def test_atomic_min_double_shared(self):
 
     def test_atomic_compare_and_swap(self):
         n = 100
-        res = [-99] * (n // 2) + [-1] * (n // 2)
+        fill = -99
+        unfill = -1
+        res = [fill] * (n // 2) + [unfill] * (n // 2)
         random.shuffle(res)
```
gmarkall, line 1124: I think this should be `np.random.shuffle`, as we seed the NumPy RNG in the `setUp` method (this may also be the source of the merge conflict).

numba/cuda/tests/cudapy/test_atomics.py
```diff
@@ -876,25 +1118,108 @@ def test_atomic_min_double_shared(self):
...
+
+        np.testing.assert_array_equal(expect_res, res)
+        np.testing.assert_array_equal(expect_out, out)
+
+    def test_atomic_compare_and_swap2(self):
+        n = 100
+        fill = -45
+        unfill = -1
+        res = [fill] * (n // 2) + [-1] * (n // 2)
+        random.shuffle(res)
```
gmarkall, line 1151:
```suggestion
        np.random.shuffle(res)
```

numba/cuda/tests/cudapy/test_atomics.py
```diff
@@ -876,25 +1118,108 @@ def test_atomic_min_double_shared(self):
...
+
+        np.testing.assert_array_equal(expect_res, res)
+        np.testing.assert_array_equal(expect_out, out)
+
+    def test_atomic_compare_and_swap3(self):
+        n = 100
+        fill = np.random.randint(50, 500, dtype=np.uint32)
+        unfill = np.random.randint(1, 25, dtype=np.uint32)
+        res = [fill] * (n // 2) + [unfill] * (n // 2)
+        random.shuffle(res)
```
gmarkall, line 1178:
```suggestion
        np.random.shuffle(res)
```

numba/cuda/tests/cudapy/test_atomics.py
```diff
@@ -876,25 +1118,108 @@ def test_atomic_min_double_shared(self):
...
+
+        np.testing.assert_array_equal(expect_res, res)
+        np.testing.assert_array_equal(expect_out, out)
+
+    def test_atomic_compare_and_swap4(self):
+        n = 100
+        fill = np.random.randint(50, 500, dtype=np.uint64)
+        unfill = np.random.randint(1, 25, dtype=np.uint64)
+        res = [fill] * (n // 2) + [unfill] * (n // 2)
+        random.shuffle(res)
```
gmarkall, line 1205:
```suggestion
        np.random.shuffle(res)
```

gmarkall (changes requested):
Many thanks for the PR - this looks good in general, and tests pass locally for me with hardware and the simulator.

There are a few comments / questions / suggestions - in addition to those on the diff, some thoughts are:

* The tests of atomic inc and atomic dec seem only to be testing with 32-bit values. Can tests for 64-bit values be added?
* Are the tests of atomic exchange all the same apart from `fill`, `unfill`, and the `dtype`? If so, could some of the logic be factorized there? If not, perhaps a short comment at the start of each explaining what it's testing could help make it clearer what each one tests.
* There are intrinsics for 32-bit increment and decrement (see page 32 of https://docs.nvidia.com/cuda/archive/9.0/pdf/NVVM_IR_Specification.pdf) - using these would be more efficient that implementing atomic inc and dec in IR. I've pointed to a couple of locations on the diff where using them would change - the implementation would be similar to that for the atomic add using an intrinsic.
* There is a merge conflict that needs to be resolved - probably from the recent PR that consistently applies use of `np.random.shuffle` in preference to `random.shuffle`.

numba/cuda/tests/cudapy/test_atomics.py
```diff
@@ -715,6 +795,168 @@ def test_atomic_xor_global_2(self):
...
+    def test_atomic_inc(self):
+        rand_const = np.random.randint(32)
+        ary = np.random.randint(0, 32, size=32).astype(np.uint32)
+        idx = np.arange(32, dtype=np.uint32)
+        orig = ary.copy()
+        cuda_func = cuda.jit('void(uint32[:], uint32[:], uint32)')(atomic_inc)
+        cuda_func[1, 32](ary, idx, rand_const)
+
+        self.assertTrue(np.all(ary == np.where(orig >= rand_const, 0,
+                                               orig + 1)))
```
gmarkall, line 807: All these `assertTrue` instances should be replaced to use `np.testing.assert_equal` instead, because it produces better diagnostics when the test fails. E.g.:

```suggestion
        np.testing.assert_equal(ary, np.where(orig >= rand_const, 0, orig + 1))
```

numba/cuda/nvvmutils.py
```diff
@@ -53,7 +53,8 @@ def declare_atomic_sub_float64(lmod):
 
 
 def declare_atomic_inc_int32(lmod):
-    fname = '___numba_atomic_i32_inc'
+    #name = '___numba_atomic_i32_inc'
```
gmarkall, line 56: This comment can now be removed.

numba/cuda/cudadrv/nvvm.py
```diff
@@ -429,6 +429,46 @@ def get(self):
 }}
 """ # noqa: E501
 
+ir_numba_atomic_inc = """
```
gmarkall, line 432: I think `ir_numba_atomic_inc` and `ir_numba_atomic_dec` need no longer be templated now they're only used for `i64`, but I'm also happy with them as they are if you'd prefer to leave it this way.

gmarkall (commented):
Thanks for the updates so far - a couple of minor comments on the diff at the moment.

numba/cuda/cudadrv/nvvm.py
gmarkall: From out-of-band conversation, this will be left as-is.

numba/cuda/cudaimpl.py
```diff
@@ -711,9 +756,14 @@ def ptx_atomic_cas_tuple(context, builder, sig, args):
     lary = context.make_array(aryty)(context, builder, ary)
     zero = context.get_constant(types.intp, 0)
     ptr = cgutils.get_item_pointer(context, builder, aryty, lary, (zero,))
-    if aryty.dtype == types.int32:
+
+    if aryty.dtype == types.int32 or aryty.dtype == types.uint32:
```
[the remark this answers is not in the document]
gmarkall: @testhound Any thoughts on this and the two following comments?

gmarkall (changes requested):
Many thanks for the update - the changes resolve some of the issues mentioned in the first review - there are a few thing still to resolve here:

- What do you think about the change suggested in https://github.com/numba/numba/pull/6499/files#r526149175?
- There is still a merge conflict that needs resolving.
- A couple of small comments / questions on the diff, partly around conventions that make it a bit easier to skim through the code and understand what to expect based on other usage in Numba.

numba/cuda/tests/cudapy/test_atomics.py
```diff
@@ -715,6 +824,209 @@ def test_atomic_xor_global_2(self):
...
+
+    def test_atomic_inc_32(self):
+        rand_const, ary, idx = self.inc_dec_1dim_setup(atype=np.uint32)
+        sig = 'void(uint32[:], uint32[:], uint32)'
+        self.check_inc_index(ary, idx, rand_const, sig, 1, 32, atomic_inc32)
+
+    def test_atomic_inc_64(self):
+        #Cannot utilize 64-bit indexes when simulating
+
+        if config.ENABLE_CUDASIM:
```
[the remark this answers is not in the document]
testhound: This error occurs when 64-bit indices are used under the simulator:

IndexError: tid=[3, 0, 0] ctaid=[0, 0, 0]: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices

gmarkall (changes requested):
Thanks for the changes! This all looks good, with the exception of two final things to resolve:

- The merge conflict - you will need to fetch the master branch from this repo and merge it into your `testhound/cuda_atomic` branch, then push it to your repo on Github.
- The tests of inc and dec with 64-bit arrays should not need to be skipped - this functionality should work.

The reason that some of the inc and dec tests don't work on the simulator is down to an edge case in NumPy's typing rules - if you index into a `uint64` array and then compute the modulo of that with respect to another integer, the result is a float64:

```python
import numpy as np

idx = np.zeros(1, dtype=np.uint64)
result = idx[0] % 2
print(result)
print(result.dtype)
```

prints:

```
0.0
float64
```

When you compute [`bin = cast_func(idx[tid] % ary_nelements)`](https://github.com/numba/numba/pull/6499/files#diff-256e605b5a313089074ed9a8d7ff60b822ca7a42936b999cbbca514b6970e6fdR60), `bin` ends up being a `float64`, which then ends up indexing `array` at https://github.com/numba/numba/pull/6499/files#diff-55fad09253581d4a7abd90da1f70bf3ebe5b9514e011880f5618ac51de42bcbfR145. The result is that NumPy complains that only integers etc. (not floats) can be used as indices.

The following change enables all the atomics tests to run under the simulator:

```patch
diff --git a/numba/cuda/tests/cudapy/test_atomics.py b/numba/cuda/tests/cudapy/test_atomics.py
index 2de20ee24..3f06cf58c 100644
--- a/numba/cuda/tests/cudapy/test_atomics.py
+++ b/numba/cuda/tests/cudapy/test_atomics.py
@@ -86,7 +86,7 @@ def atomic_binary_2dim_global(ary, op2, binop_func, y_cast_func):
 @cuda.jit(device=True)
 def atomic_binary_1dim_global(ary, idx, ary_nelements, op2, binop_func):
     tid = cuda.threadIdx.x
-    bin = idx[tid] % ary_nelements
+    bin = int(idx[tid] % ary_nelements)
     binop_func(ary, bin, op2)
 
 
@@ -284,7 +284,7 @@ def atomic_inc32(ary, idx, op2):
 
 def atomic_inc64(ary, idx, op2):
     atomic_binary_1dim_shared2(ary, idx, op2, uint64, 32,
-                               cuda.atomic.inc, atomic_cast_none)
+                               cuda.atomic.inc, atomic_cast_to_int)
 
 
 def atomic_inc2_32(ary, op2):
@@ -318,7 +318,7 @@ def atomic_dec32(ary, idx, op2):
 
 def atomic_dec64(ary, idx, op2):
     atomic_binary_1dim_shared2(ary, idx, op2, uint64, 32,
-                               cuda.atomic.dec, atomic_cast_none)
+                               cuda.atomic.dec, atomic_cast_to_int)
 
 
 def atomic_dec2_32(ary, op2):
@@ -848,7 +848,6 @@ class TestCudaAtomics(CUDATestCase):
         sig = 'void(uint32[:], uint32[:], uint32)'
         self.check_inc_index(ary, idx, rand_const, sig, 1, 32, atomic_inc32)
 
-    @skip_on_cudasim('Cannot utilize 64-bit indexes when simulating')
     def test_atomic_inc_64(self):
         rand_const, ary, idx = self.inc_dec_1dim_setup(dtype=np.uint64)
         sig = 'void(uint64[:], uint64[:], uint64)'
@@ -875,7 +874,6 @@ class TestCudaAtomics(CUDATestCase):
         self.check_inc_index2(ary, idx, rand_const, sig, 1, 32,
                               atomic_inc_global)
 
-    @skip_on_cudasim('Cannot utilize 64-bit indexes when simulating')
     def test_atomic_inc_global_64(self):
         rand_const, ary, idx = self.inc_dec_1dim_setup(dtype=np.uint64)
         sig = 'void(uint64[:], uint64[:], uint64)'
@@ -924,7 +922,6 @@ class TestCudaAtomics(CUDATestCase):
         sig = 'void(uint32[:], uint32[:], uint32)'
         self.check_dec_index(ary, idx, rand_const, sig, 1, 32, atomic_dec32)
 
-    @skip_on_cudasim('Cannot utilize 64-bit indexes when simulating')
     def test_atomic_dec_64(self):
         rand_const, ary, idx = self.inc_dec_1dim_setup(dtype=np.uint64)
         sig = 'void(uint64[:], uint64[:], uint64)'
@@ -951,7 +948,6 @@ class TestCudaAtomics(CUDATestCase):
         self.check_dec_index2(ary, idx, rand_const, sig, 1, 32,
                               atomic_dec_global)
 
-    @skip_on_cudasim('Cannot utilize 64-bit indexes when simulating')
     def test_atomic_dec_global_64(self):
 
         rand_const, ary, idx = self.inc_dec_1dim_setup(dtype=np.uint64)
```

The workaround is essentially to ensure that only integers get passed as indices to the atomic functions. What do you think about applying the above?

gmarkall (changes requested):
Thanks for the changes! The code is now looking great! Just a couple of things to get the tests passing again:

- Another merge of master is needed to pick up the readthedocs build fix that was merged yesterday.
- A change to the docstring for `atomic.dec` to fix the sphinx warning. The following change seems to fix it:

```patch
diff --git a/numba/cuda/stubs.py b/numba/cuda/stubs.py
index b94f70bf8..74c959f91 100644
--- a/numba/cuda/stubs.py
+++ b/numba/cuda/stubs.py
@@ -465,6 +465,7 @@ class atomic(Stub):
 
         Perform ary[idx] = (value if (array[idx] == 0) or
                             (array[idx] > value) else array[idx] - 1).
+
         Supported on uint32, and uint64 operands only.
 
         Returns the old value at the index location as if it is loaded
```

This is a little odd because the warning indicates that the line beginning with `Perform ary[idx] = ` is interpreted as a definition list - however, adding the blank line doesn't result in those lines getting generated as a definition list, so I think that it's OK to just add the blank line to fix the issue.

gmarkall (changes requested):
Apologies for failing to spot this before, but there's a test fail:

```
ERROR: test_issue_5835 (numba.cuda.tests.cudapy.test_debuginfo.TestCudaDebugInfo)
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudadrv/driver.py", line 2194, in complete
    driver.cuLinkComplete(self.handle, byref(cubin), byref(size))
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudadrv/driver.py", line 300, in safe_cuda_api_call
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudadrv/driver.py", line 335, in _check_error
numba.cuda.cudadrv.driver.CudaAPIError: [999] Call to cuLinkComplete results in CUDA_ERROR_UNKNOWN

  File "/home/gmarkall/numbadev/numba/numba/cuda/tests/cudapy/test_debuginfo.py", line 58, in test_issue_5835
    def f(x):
  File "/home/gmarkall/numbadev/numba/numba/cuda/decorators.py", line 130, in kernel_jit
    return Dispatcher(func, sigs, targetoptions=targetoptions)
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 849, in __init__
    self.compile(sigs[0])
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 1009, in compile
    kernel.bind()
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 580, in bind
    self._func.get()
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 455, in get
    cubin, size = linker.complete()
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudadrv/driver.py", line 2196, in complete
numba.cuda.cudadrv.driver.LinkerError: [999] Call to cuLinkComplete results in CUDA_ERROR_UNKNOWN
error   : Undefined reference to '___numba_cas_hack' in '<cudapy-ptx>'
```

This is because raising exceptions from kernels uses an atomic CAS to record whether an exception occurred in the current thread. The fix is:

```patch
diff --git a/numba/cuda/target.py b/numba/cuda/target.py
index c42275e5e..73f9417e2 100644
--- a/numba/cuda/target.py
+++ b/numba/cuda/target.py
@@ -180,8 +180,8 @@ class CUDATargetContext(BaseContext):
                 casfnty = lc.Type.function(old.type, [gv_exc.type, old.type,
                                                       old.type])
 
-                casfn = wrapper_module.add_function(casfnty,
-                                                    name="___numba_cas_hack")
+                cas_hack = "___numba_atomic_i32_cas_hack"
+                casfn = wrapper_module.add_function(casfnty, name=cas_hack)
                 xchg = builder.call(casfn, [gv_exc, old, status.code])
                 changed = builder.icmp(ICMP_EQ, xchg, old)
 
```

numba/cuda/cudadrv/nvvm.py
```diff
@@ -501,6 +539,14 @@ def llvm_to_ptx(llvmir, **opts):
         ('declare double @___numba_atomic_double_sub(double*, double)',
          ir_numba_atomic_binary.format(T='double', Ti='i64', OP='fsub',
                                        FUNC='sub')),
+        ('declare i64 @___numba_atomic_i64_inc(i64*, i64)',
+         ir_numba_atomic_inc.format(T='i64')),
+        ('declare i64 @___numba_atomic_i64_dec(i64*, i64)',
+         ir_numba_atomic_dec.format(T='i64')),
```
stuartarchibald, line 545: Should these be called `___numba_atomic_u64_{inc, dec}`?

numba/cuda/cudaimpl.py
```diff
@@ -625,6 +625,40 @@ def ptx_atomic_sub(context, builder, dtype, ptr, val):
...
+@_atomic_dispatcher
+def ptx_atomic_inc(context, builder, dtype, ptr, val):
+    if dtype == types.uint32:
+        lmod = builder.module
+        return builder.call(nvvmutils.declare_atomic_inc_int32(lmod),
+                            (ptr, val))
+    elif dtype == types.uint64:
+        lmod = builder.module
+        return builder.call(nvvmutils.declare_atomic_inc_int64(lmod),
+                            (ptr, val))
```
stuartarchibald, line 640: there's no need to change this, but you can write this sort of thing like:
```python
    if dtype in unsigned_int_numba_types:
        bw = dtype.bitwidth
        lmod = builder.module
        fn = getattr(nvvmutils, f'declare_atomic_inc_int{bw}')
        return builder.call(fn(lmod), (ptr, val))
    else:
        raise TypeError(f'Unimplemented atomic inc with {dtype} array')
```

numba/cuda/cudaimpl.py
```diff
@@ -625,6 +625,40 @@ def ptx_atomic_sub(context, builder, dtype, ptr, val):
         return builder.atomic_rmw('sub', ptr, val, 'monotonic')
 
 
+@lower(stubs.atomic.inc, types.Array, types.intp, types.Any)
+@lower(stubs.atomic.inc, types.Array, types.UniTuple, types.Any)
+@lower(stubs.atomic.inc, types.Array, types.Tuple, types.Any)
+@_atomic_dispatcher
```
stuartarchibald, line 631: Whilst the type restriction is documented, I think a load of these atomic functions could be typed more specifically to catch issues at type inference time. Example:

This is fine
```python
from numba import cuda
from numba import uint32
from numba import njit
import numpy as np

@cuda.jit('void(uint32[:], uint32, int8)')
def atomic_inc32(ary, idx, val):
    tid = cuda.threadIdx.x
    cuda.atomic.inc(ary, tid % 2, val)

ary = np.zeros(32, dtype=np.uint32)
idx = 0
rand_const = 100

atomic_inc32[1, 64](ary, idx, rand_const)
print(ary)
```
as `int8` is coercible to `uint32`. Whereas:
```python
from numba import cuda
from numba import uint32
from numba import njit
import numpy as np

@cuda.jit('void(uint32[:], uint32, complex64)')
def atomic_inc32(ary, idx, val):
    tid = cuda.threadIdx.x
    cuda.atomic.inc(ary, tid % 2, val)

ary = np.zeros(32, dtype=np.uint32)
idx = 0
rand_const = 100

atomic_inc32[1, 64](ary, idx, rand_const)
print(ary)
```
is not coercible and results in a lowering error. I don't think that this should be fixed in this PR, but perhaps this is something to look at in a follow up PR?

stuartarchibald (commented):
Thanks for the PR, it looks good, no changes required (apart from the fix in https://github.com/numba/numba/pull/6499#pullrequestreview-539820575) but I've noted a few things in the review for consideration at a later time. Thanks for reviewing @gmarkall.

Once the fix for the CUDA CAS function name change is made I'll push this through the farm.

numba/cuda/cudadrv/nvvm.py
testhound: Change made.

numba/cuda/cudaimpl.py
testhound: I made this change anyway.

numba/cuda/cudaimpl.py
testhound: Agree with catching issues at type inference time and a separate PR.

gmarkall (approved):
Looks good to me!

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_145`.

stuartarchibald (approved):
Thanks for the fixes @testhound, thanks for reviewing @gmarkall.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_145`.

Passed.

sklam:
Renamed PR title so that it is easier to when we auto-generate the release note

PR #6500, opened 2020-11-18
Update bug and feature request templates.

stuartarchibald:
Wording changes to try and push general questions onto discourse and
keep github issues/PRs for development. This makes bug and feature
request text point to "below" which has a General Question category
as present on discourse. Also changes the bug report template to ask
for a self contained reproducer with a simple definition of what that
is.

### DIFF
--- a/.github/ISSUE_TEMPLATE/Bug_report.md
+++ b/.github/ISSUE_TEMPLATE/Bug_report.md
@@ -1,6 +1,6 @@
 ---
 name: Bug Report
-about: Create a report to help us improve Numba
+about: Report a bug. Not for asking general questions - see below.
 
 ---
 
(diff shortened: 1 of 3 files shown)


stuartarchibald:
Thoughts on this change?

* The Numba discourse instance is a great place for long form discussion and since its inception the issue tracker has become more issue orientated. This change is with view of directing general questions towards discourse and keeping the issue tracker for bugs and development.
* This patch also changes the requested code sample for bug reports, I'm of the view that having a self contained piece of code that is not minimal is preferable to one that is more minimal but doesn't actually reproduce/is not self contained.

CC @numba/czars

esc:
I have reviewed this and it looks good. I also think that the suggestion to submit something that can be `python bug.py` is a good one.

esc (approved):
LGTM

gmarkall (approved):
Looks good to me!

gmarkall:
>     * This patch also changes the requested code sample for bug reports, I'm of the view that having a self contained piece of code that is not minimal is preferable to one that is more minimal but doesn't actually reproduce/is not self contained.

I think the change also explains succinctly what to include, without the reporter having to go and read a post elsewhere (which I suspect increased the effort required to work out what's expected beyond the level that a lot of people would go to).

PR #6506, opened 2020-11-19
Fix requirements.txt pinning

sklam:
Fix the malform requirements.txt

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 setuptools
 numpy>=1.10
-llvmlite 0.35.0rc3
+llvmlite>=0.35.0rc3,<0.36.0
 argparse


sklam:
The file is not included in wheels nor the tarball. It will only affect those building the release branch directly. The correct requirements are still encoded in setup.py

requirements.txt
```diff
@@ -1,4 +1,4 @@
 setuptools
 numpy>=1.10
-llvmlite 0.35.0rc3
+llvmlite>=0.35.0rc3,<0.36.0
```
stuartarchibald, line 3: This looks right, https://github.com/numba/numba/blob/release0.51/requirements.txt from 0.51 has `llvmlite>=0.34.0dev0,<0.35` but in this case `0.35.0rc3` has an API change to necessarily accommodate attributes on call sites so has to be the lower limit.

stuartarchibald (approved):

PR #6367, opened 2020-10-14
Support views on pinned/mapped arrays

c200chromebook:
Adds support for views on pinned/mapped arrays, these required both stream and is_c_contiguous, which is moved to the device array base class.

PR #6510, opened 2020-11-20
CUDA: Make device array assignment synchronous where necessary

### LINKED ISSUE #6505: CUDA: assignment to device array elements is asynchronous, and shouldn't be
Running the following example on Windows with a Quadro RTX 8000:

```python
from numba import cuda
import numpy as np

ary = cuda.mapped_array(2, dtype=np.int32)
ary[:] = 0
print(ary)

ary_v = ary.view('u1')
ary_v[1] = 1
ary_v[5] = 1
# cuda.synchronize()
print(ary)
assert sum(ary) == 512
```

produces the following output for me:

```
[0 0]
[256   0]
  File ".\repro.py", line 13, in <module>
    assert sum(ary) == 512
```

Uncommenting the call to `cuda.synchronize()` yields the following output instead:

```
[0 0]
[256 256]
```

This is because `__setitem__` for a device array uses a kernel to set the element: https://github.com/numba/numba/blob/master/numba/cuda/cudadrv/devicearray.py#L629

gmarkall:
Fixes #6505. This is done by using the default stream and synchronizing on it after the assignment, when the array has no default stream and no stream for the operation is supplied.

Also fixes the `DeviceNDArray.setitem` method, which had a typo and could not work.

A test case containing the reproducer from Issue #6505 is also added, but it is presently skipped because it depends on the functionality in #6367. I've tried this branch locally merged with #6367, and this test case passes with it.


numba/cuda/tests/cudadrv/test_cuda_array_slicing.py
```diff
@@ -308,6 +309,67 @@ def test_incompatible_shape(self):
...
+        with patch.object(cuda.cudadrv.driver.Stream, 'synchronize',
+                          return_value=None) as mock_sync:
+            darr[0] = 10
+
+        mock_sync.assert_called_once()
+
+    @skip_on_cudasim('cudasim does not use streams and operates synchronously')
+    def test_no_sync_default_stream(self):
+        # There should not be a synchronization when a the array has a default
+        # stream, whether it is the default stream, the legacy default stream,
```
stuartarchibald, line 326:
```suggestion
        # There should not be a synchronization when the array has a stream
        # provided, whether it is the default stream, the legacy default stream,
```

numba/cuda/tests/cudadrv/test_cuda_array_slicing.py
```diff
@@ -308,6 +309,67 @@ def test_incompatible_shape(self):
...
+        # assertion, due to the assignment being done with a kernel launch that
+        # returns asynchronously - there should now be a sync after the kernel
+        # launch to ensure that the writes are always visible.
+        ary = cuda.mapped_array(2, dtype=np.int32)
+        ary[:] = 0
+
+        ary_v = ary.view('u1')
+        ary_v[1] = 1
+        ary_v[5] = 1
+        assert sum(ary) == 512
```
stuartarchibald, line 371:
```suggestion
        self.assertEqual(sum(ary), 512)
```

numba/cuda/tests/cudadrv/test_cuda_array_slicing.py
```diff
@@ -308,6 +309,67 @@ def test_incompatible_shape(self):
...
+        for stream in streams:
+            darr = cuda.to_device(np.arange(5))
+
+            with patch.object(cuda.cudadrv.driver.Stream, 'synchronize',
+                              return_value=None) as mock_sync:
+                darr.setitem(0, 10, stream=stream)
+
+            mock_sync.assert_not_called()
+
+    @unittest.skip('Requires PR #6367')
```
stuartarchibald, line 359: I've merged in 6367 and manually checked that this works ok.

stuartarchibald (commented):
Thanks for the patch, changes look good, just couple of minor things to resolve.

numba/cuda/tests/cudadrv/test_cuda_array_slicing.py
gmarkall: I agree with the removal of "a" (that is an error), but saying "has a stream provided" seems less clear than "has a default stream", given that the default stream of a device array is a concept in Numba. However, this does highlight a deficiency in the docs for the stream constructors ([example](https://numba.readthedocs.io/en/stable/cuda-reference/memory.html#numba.cuda.device_array)), in that they don't explain that the stream parameter becomes the default stream for the array.

numba/cuda/tests/cudadrv/test_cuda_array_slicing.py
gmarkall: Many thanks!

numba/cuda/tests/cudadrv/test_cuda_array_slicing.py
gmarkall: (I'd like to follow this PR up today with another one that improves the docs for `device_array`, et al. so that they give a clear explanation of the "default stream".)

gmarkall:
@stuartarchibald Many thanks for the review - all comments should be addressed now.

numba/cuda/tests/cudadrv/test_cuda_array_slicing.py
stuartarchibald: Thanks for adding some more explanation, I agree that default stream needs to stay, but could do with some docs as noted.

stuartarchibald (approved):
Thanks for the fixes!

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_142.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_142.

Passed.

PR #6514, opened 2020-11-24
Catch user code with compound type use and offer advice.

stuartarchibald:
As title.

### DIFF
--- a/numba/core/typeinfer.py
+++ b/numba/core/typeinfer.py
@@ -1459,7 +1459,37 @@ https://numba.pydata.org/numba-doc/latest/user/troubleshoot.html#my-code-has-an-
         try:
             return self.context.resolve_value_type(val)
         except ValueError as e:
-            msg = str(e)
+            if issubclass(val, types.Type):
+                from numba.core.errors import termcolor
+                code = termcolor().indicate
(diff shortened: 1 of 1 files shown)


stuartarchibald:
This is WIP. It's here to catch the common case of use code with Numba container types in use. e.g.
```python
from numba import njit, typed, types
from numba.types import Tuple, float64

@njit
def foo():
    x = typed.List.empty_list(Tuple((float64,)))

foo()
```
now produces an error message like:
```
Untyped global name 'Tuple': This happens if you use a Numba container type in a JIT compiled function. To get this to work, assign the Numba type in use to a variable at module level and then use that variable in the JIT code.

Example...
Change this:
from numba import njit, typed
from numba.types import Tuple, float64

@njit
def foo():
    x = typed.List.empty_list(Tuple((float64,)))

To this:
from numba import njit, typed
from numba.types import Tuple, float64

my_type = Tuple((float64,))

@njit
def foo():
    x = typed.List.empty_list(my_type)



File "script.py", line 6:
def foo():
    x = typed.List.empty_list(Tuple((float64,)))
```

stuartarchibald:
Closing this for now, not sure of the merits of involved error messages given this problem is potentially fixable.

Issue #2630, opened 2017-12-02
Leveraging Numba as a Standalone Compiler like Nuitka without Intrusive Annotations


sirinath:
Having to add annotations to functiona is very intrusive. Sometimes you might be calling functions written by other parties and in libraries. It would be helpful to have Numba compile the code without annotations either AOT or using JIT.

dhirschfeld:
Since a decorator is just a function which returns a function AFAICS you can easily  jit any function in your own code:
```python
from numba import jit
from thirdparty import slow_func

fast_func = jit(slow_func)
```

It's likely also possible to monkeypatch `thirdparty` to replace `slow_func` with the jitted version so that all internal functions will benefit.

*Disclaimer:* Just thinking out loud - I haven't tried it, YMMV.

sirinath:
If a function calls another function which calls another functions ... These kind of usage becomes very cumbersome. Since most of the work is already done we can have an alternative python compiler which does AOT and JIT using Numba.

dscole:
I completely agree with @sirinath. It seems silly that to make this work:
```
def some_func(var):
    return var + 5

@jit 
def my_func(var):
    return some_func(var)
```
 
you just need to add `@jit` to some_func, but numba can't provide that functionality for you. For our usage it would also be very helpful if while compiling a function and numba encounters an undecorated function it just tries to jit right there and then.

seibert:
While I agree with the sentiment that Numba should try compiling dependent functions, Numba is sufficiently limited in what kinds of code it can compile that we prefer users to opt into compiling all needed functions for now.  It isn't clear how often compiling arbitrary functions in third-party code would actually work given the current limitations of the compiler.  (Maybe we've turned a corner on this that I'm not aware of...)

I think once object mode is a little more sophisticated (right now the only benefit is loop lifting) this is something we could investigate.  We would need to see if there is a way to mark functions that Numba previously failed to compile as such, so that we don't waste time attempting to compile them over and over.

dscole:
[...]

PR #6515, opened 2020-11-24
CUDA: Add support for the jit_module function

gmarkall:
This PR adds the `jit_module` decorator to CUDA as `numba.cuda.jit_module`. It is slightly different to the `numba.jit_module` function, in that it compiles functions as device functions by default, but can be made to compile functions as kernels by passing `device=False` to it. This is handy for jitting some functions as device functions and others as kernels by calling it twice in a module, and the example added to the documentation reflects this.

Whilst writing tests I noticed that `FakeCUDAKernel` from the simulator stores the Python function as `fn` instead of `py_func`, which is a discrepancy between it and the real dispatcher - to make this more consistent I've moved `fn` to `py_func` and provided a property `fn` with a deprecation warning to support the old name - though, it's unlikely this was ever used by anyone as it's not advertised and specific to the simulator only.


gmarkall:
A question related to this PR: how would others feel about the `jit_module` decorator allowing jitting of other modules, to support the UDF use case described by @dscole in https://github.com/numba/numba/issues/2630#issuecomment-349064212? I am also aware of other users that are thinking of potential UDF usecases too. I imagine implementing this with something like the following change (as a sketch - some error checking would make sense, e.g. to check a module is passed):

```patch
diff --git a/numba/cuda/decorators.py b/numba/cuda/decorators.py
index 9923bf08c..267f02951 100644
--- a/numba/cuda/decorators.py
+++ b/numba/cuda/decorators.py
@@ -163,7 +163,7 @@ def convert_types(restype, argtypes):
     return restype, argtypes
 
 
-def jit_module(**kwargs):
+def jit_module(module=None, **kwargs):
     """ Automatically ``jit``-wraps functions defined in a Python module. By
     default, wrapped functions are treated as device functions rather than
     kernels - pass ``device=False`` to treat functions as kernels.
@@ -185,9 +185,10 @@ def jit_module(**kwargs):
     if 'device' not in kwargs:
         kwargs['device'] = True
 
-    # Get the module jit_module is being called from
-    frame = inspect.stack()[1]
-    module = inspect.getmodule(frame[0])
+    if module is None:
+        # Get the module jit_module is being called from
+        frame = inspect.stack()[1]
+        module = inspect.getmodule(frame[0])
     # Replace functions in module with jit-wrapped versions
     for name, obj in module.__dict__.items():
         if inspect.isfunction(obj) and inspect.getmodule(obj) == module:
```

With this, one could do:

```python
from numba import cuda
import user_functions

cuda.jit_module(user_functions)

@cuda.jit
def wrapper(x):
    i = cuda.grid(1)
    if i < len(x):
        x = user_functions.computation(x)
```

where `user_functions` contains something like:

```python
import math

def computation(x):
    # Some arbitrary mathematical operation
    return math.cos(x) + x ** x
```

Here the user code itself needn't be aware of Numba at all, whereas presently (in this PR and in the CPU target `jit_module` function) the user is required to call `jit_module` in their module.

numba/cuda/simulator/compiler.py
```diff
@@ -7,3 +7,10 @@
 
 compile_ptx = None
 compile_ptx_for_current_device = None
+
+
+class DeviceFunctionTemplate:
```
apmasell, line 12: Please add a comment explaining why it's an empty marker class.

apmasell (approved):

gmarkall:
Still interested in this, just haven't got it to the top of my pile in a long while :-)

gmarkall:
I still intend to address this one day when it bubbles to the top of my queue.

PR #6508, opened 2020-11-19
Defer requirements to setup.py

sklam:
as suggested by https://caremad.io/posts/2013/07/setup-vs-requirement/
which is linked from https://pip.pypa.io/en/stable/user_guide/

PR #6516, opened 2020-11-25
Fix RTD build by using conda.

stuartarchibald:
As title.

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -1,0 +1,12 @@
+# This environment is used by the RTD config for PR builds. RTD uses this as the
+# base environment and then adds in the sphinx etc tools on top.
+# See: https://docs.readthedocs.io/en/stable/guides/conda.html
+name: rtd
+channels:
+  - numba/label/dev
+dependencies:
+  - python=3.7
+  - llvmlite
+  - numpy
+  - numpydoc
+  - setuptools
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -12,11 +12,13 @@ sphinx:
 formats:
   - pdf
 
+# use conda for the environment
+conda:
+  environment: docs/environment.yml
+
 # Optionally set the version of Python and requirements required to build your docs
 python:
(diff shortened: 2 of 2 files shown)


docs/environment.yml
```diff
@@ -0,0 +1,12 @@
+# This environment is used by the RTD config for PR builds. RTD uses this as the
+# base environment and then adds in the sphinx etc tools on top.
+# See: https://docs.readthedocs.io/en/stable/guides/conda.html
+name: rtd
+channels:
+  - numba/label/dev
+dependencies:
```
gmarkall, line 7: The [developer docs](https://numba.readthedocs.io/en/stable/developer/contributing.html#getting-set-up) also recommend installing jinja2 and cffi (and scipy as well, but I know there's no dependency on that in Numba) - is there any chance not having these present might result in some part of the docs getting omitted?

stuartarchibald: Good question. I don't think so on the basis that the `requirements.txt` which the docs were building against, prior to them being broken by lack of wheels, do not contain these.
https://github.com/numba/numba/blob/92861d01a5f331f4262d5c7d1798b3f7517086b7/requirements.txt
https://github.com/numba/numba/blob/92861d01a5f331f4262d5c7d1798b3f7517086b7/docs/requirements.txt

`argparse` is in `<numba root>/requirements.txt` but not mentioned in the docs.

gmarkall: I think `argparse` was only needed for Python < 3.3 (or thereabouts), and it's now part of the standard library.

gmarkall (approved):
LGTM!

docs/environment.yml
stuartarchibald: ah yes, thanks. https://docs.python.org/3/library/argparse.html 
> _New in version 3.2_

 I think https://github.com/numba/numba/pull/6508 sorts this out anyway by just deferring to `setup.py`.

sklam (approved):

PR #6517, opened 2020-11-25
CUDA: Add NVVM test of all 8-bit characters

gmarkall:
This test case is adapted from https://github.com/numba/llvmlite/pull/53 and is used to ensure that the encoding used by llvmlite does not cause issues with NVVM.

This is as suggested in https://github.com/numba/llvmlite/pull/655#issuecomment-732255325

I've tested this with llvmlite using both the `latin1` and `utf-8` encodings, and it passes in both cases.

### DIFF
--- a/numba/cuda/tests/nocuda/test_nvvm.py
+++ b/numba/cuda/tests/nocuda/test_nvvm.py
@@ -2,6 +2,10 @@ from numba.cuda.compiler import compile_kernel
 from numba.cuda.cudadrv import nvvm
 from numba.cuda.testing import skip_on_cudasim, SerialMixin
 from numba.core import types, utils
+
+from llvmlite import ir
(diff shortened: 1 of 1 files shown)


stuartarchibald:
Bumped for CI.

stuartarchibald:
might need merge in from mainline to pick up the changes that fix the doc builds

gmarkall:
Merge made - hopefully this next build passes.

stuartarchibald (approved):
Thanks for the patch, good to have this test in Numba.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_144`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_144`.

Passed.

PR #6519, opened 2020-11-25
Avoid unnecessary ir.Del generation and removal

ehsantn:
Removes unnecessary ir.Del generation and removal in Parfor code, which used to be necessary. This simplifies the code (and may slightly improve compilation time).

### DIFF
--- a/numba/core/ir_utils.py
+++ b/numba/core/ir_utils.py
@@ -1412,7 +1412,6 @@ def restore_copy_var_names(blocks, save_copies, typemap):
     replace_var_names(blocks, rename_dict)
 
 def simplify(func_ir, typemap, calltypes):
-    remove_dels(func_ir.blocks)
     # get copies in to blocks and out from blocks
     in_cps, out_cps = copy_propagate(func_ir.blocks, typemap)
     # table mapping variable names to ir.Var objects to help replacement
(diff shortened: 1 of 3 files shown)


ehsantn:
I assume the docs failure is not relevant.

DrTodd13 (approved):
@ehsantn If it passes the regression tests I'm fine with it but do you remember why we got rid of del's originally?  About my only concern is that the code movement code is able to move dels around to enable fusion but if the regressions pass then I assume this is working since we test that.

ehsantn:
Thank you @DrTodd13. I think we removed dels since our transformations would make them invalid (e.g. some variable is removed). Much easier to regenerate them than to consider them in every change.

stuartarchibald:
Thanks for the patch @ehsantn, good to see less postproc use. The IR is `Del` free for a lot of the pipeline but some things unfortunately still need it.

stuartarchibald:
Thanks for the review @DrTodd13 !

stuartarchibald:
@ehsantn any chance in future you could please use your fork of Numba for creating patches, we try really hard to keep the `numba/numba` repo with just release branches and a few others kept for historical reasons. Many thanks for your cooperation.

ehsantn:
Sorry, just a mistake. Will do.

stuartarchibald:
> Sorry, just a mistake. Will do.

No worries, and thanks :)

PR #6524, opened 2020-11-27
Fix typed.List return value

### LINKED ISSUE #6523: typed.List.extend returns self in some cases
```python
from numba.typed import List
  ...: my_list = List([1,2])
  ...: print(my_list.extend(List([1,2])))
  ...: print(my_list.extend(tuple()))
None
[1, 2, 1, 2]
```

It should always return None but in some cases it's returning self. Below is python list's behaviour.

```python
  ...: my_list = list([1,2])
  ...: print(my_list.extend(list([1,2])))
  ...: print(my_list.extend(tuple()))
None
None
```
_Originally posted by @luk-f-a in https://github.com/numba/numba/issues/5958#issuecomment-734750145_

luk-f-a:
Fixes #6523

### DIFF
--- a/numba/tests/test_typedlist.py
+++ b/numba/tests/test_typedlist.py
@@ -776,7 +776,8 @@ class TestExtend(MemoryLeakMixin, TestCase):
         # Extending an unrefined list with an empty iterable doesn't work in a
         # jit compiled function as the list remains untyped.
         l = List()
-        l.extend(tuple())
+        ret = l.extend(tuple())
+        self.assertIsNone(ret)
         self.assertEqual(len(l), 0)
         self.assertFalse(l._typed)
 
--- a/numba/typed/typedlist.py
+++ b/numba/typed/typedlist.py
@@ -357,7 +357,7 @@ class List(MutableSequence):
     def extend(self, iterable):
         # Empty iterable, do nothing
         if len(iterable) == 0:
-            return self
+            return None
         if not self._typed:
             # Need to get the first element of the iterable to initialise the
             # type of the list. FIXME: this may be a problem if the iterable


stuartarchibald:
Thanks for the patch. This change could do with a test adding please, ideally from the reproducer in #6523. It'd also be good to look at the related items in https://github.com/numba/numba/issues/6523#issuecomment-734771361 if time is available. Thanks again!

CC @esc please could you coordinate on this/review, thanks.

luk-f-a:
@stuartarchibald , from your comment, do you mean that instead of `return _extend(self, iterable)` it should do `_extend(self, iterable)` without return?

stuartarchibald:
> @stuartarchibald , from your comment, do you mean that instead of `return _extend(self, iterable)` it should do `_extend(self, iterable)` without return?

Yes. I'm in two minds about it. Perhaps its best if all the `extend` functions return the result of the extend, which should be `None`, such that if there is erroneous behaviour at least it propagates and can be caught. Thoughts?

luk-f-a:
Personally I would return the result of the function as it's currently doing, but I don't know the advantages of the other approach, so I cannot compare.
Usually, the only `return` that I avoid is a `return None` as last line.
In interpreted python both approaches are the same, does it make a difference in jitted code?

esc:
I've assigned this to myself and will get to it on Monday.

esc:
This looks fine so-far, the reminaing TODOs on this appear to be:

* Figure out which other `extend` type functions return self
* Decide on how to handle internal `return`s?

@stuartarchibald  @luk-f-a Does this sound right?

esc:
I just took a look at the source-code. If we decide to return from `_extend` we should adopt this coherently for other in-place modification functions that also return `None` such as, for example, `insert`.

esc:
Hmmm, many of the functions seem to do this already: `insert`, `clear`, `reverse` for example.

luk-f-a:
hi @esc, while I opened this PR to take care of `extend` I can also take care of the first point "Figure out which other extend type functions return self" in this PR. Typeguard didn't detect any other cases, but the code is not long, I can look manually.

The second point is up to you and Stuart to decide, I can make the changes in this PR if you'd like.

esc:
Probably a good idea to separate concerns here and fix the immediate issue first, i.e. making sure nothing returns `self`. Then , deal with the question of "should functions that return None, return the result of internal calls to make sure `None` is always returned at" in a second PR? Does that sound like a good plan of attack?

luk-f-a:
I checked manually and using Pycharm that nothing else is returning self. It seems that nothing else is.

esc (approved):
LGTM

esc:
@luk-f-a thanks for looking into this. I hav reviewed this and it is ready to merge now.

PR #6530, opened 2020-11-30
Update llvmlite dependency

sklam:
use llvmlite 0.35.*


stuartarchibald (approved):
Looks good, thanks for the patch.

sklam:
Azure CI verified the dependency in conda and I have manually verified it for the wheels.

PR #6531, opened 2020-12-01
Update canonical URL.

### LINKED ISSUE #6479: Update canonical_url configuration in docs
I think we need to fix the canonical_url in our generated docs, especially now that we have moved documentation to RTD.  This should help search engines direct users to the latest documentation even after indexing older versions.    According to the sphinx RTD theme, the URL should be the base path:

https://sphinx-rtd-theme.readthedocs.io/en/stable/configuring.html#confval-canonical_url

I think we did this incorrectly before (linking to the index file, rather than the base URL), so we will need to fix the path as well as relocate to the new Numba docs domain:

https://github.com/numba/numba/blob/master/docs/source/conf.py#L125

stuartarchibald:
As title.

Fixes #6479

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -122,7 +122,7 @@ html_theme = 'sphinx_rtd_theme'
 # All sphinx_rtd_theme options. Default values commented out; uncomment to
 # change.
 html_theme_options = {
-    'canonical_url': 'https://numba.pydata.org/numba-doc/latest/index.html',
+    'canonical_url': 'https://numba.readthedocs.io/en/stable/',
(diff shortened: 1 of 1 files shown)


seibert:
I can confirm from the RTD build that this change is inserting the proper element in the HTML header.  For example, the CUDA docs page contains this now:
```
<link rel="canonical" href="https://numba.readthedocs.io/en/stable/cuda/index.html">
```
Which links to the current stable docs page.

seibert (approved):

stuartarchibald:
Thanks for the review @seibert

PR #6545, opened 2020-12-04
Refactoring ParforDiagnostics

1e-to:
We need to be able to get "all_lines" to reuse it in our ExtendedParforDiagnostics to avoid code duplication.
Also now "dump" function is too complicated so this refactoring was made.

### DIFF
--- a/numba/parfors/parfor.py
+++ b/numba/parfors/parfor.py
@@ -889,7 +889,7 @@ class ParforDiagnostics(object):
             if print_loop_search:
                 print_wrapped(fmt % (pf.id, loc, r_pattern))
             parfors_simple[pf.id] = (pf, loc, r_pattern)
-            return parfors_simple
+        return parfors_simple
 
     def get_all_lines(self, parfors_simple):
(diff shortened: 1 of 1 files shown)


stuartarchibald:
@1e-to thanks for the patch. Before review, just want to check that the intention here is largely code motion with view of putting the diagnostics stuff all in the same place? There should be no functional change?

1e-to:
> @1e-to thanks for the patch. Before review, just want to check that the intention here is largely code motion with view of putting the diagnostics stuff all in the same place? There should be no functional change?

Yes, some functionality has been moved from the dump to separate functions so that we can reuse it.

stuartarchibald (approved):
Thanks for the patch. The diagnostic reporting isn't heavily tested in the Numba test suite, however this patch is pretty much just code motion with a small amount of clean up of dead code so I think the risk is minimal. The refactoring presented is a good improvement over having the diagnostics inlined. I've checked this locally and the reports seem identical. Happy to merge as-is. Thanks again!

Issue #4706, opened 2019-10-15
np.heaviside in `no_python` mode


paugier:
Hi,

`np.heaviside(x1, x2)` (https://docs.scipy.org/doc/numpy/reference/generated/numpy.heaviside.html) does not seem to be supported in `no_python` mode.

I would be very interested to see how it could be implemented in Numba! Nearly only modification in the pure-Python reimplementation of Numpy? Can I look at examples of similar functions in Numba code?

esc:
@paugier thanks for asking about this. We do have a fairly neat API that allows us to implement Numpy stuff easily:

https://numba.pydata.org/numba-doc/dev/extending/overloading-guide.html

The document also has some links to existing implementation which you may be able to use as a template.

david-zwicker:
I needed an implementation of `np.heaviside` for a recent project and came up with the following:

```python

@overload(np.heaviside)
def np_heaviside(x1, x2):
    @register_jitable
    def heaviside_impl(x1, x2):
        if x1 < 0:
            return 0.0
        elif x1 > 0:
            return 1.0
        else:
            return x2

    return heaviside_impl
```
This seems to do the job in my case, but I'm not sure if this is general enough and where it would need to go within the `numba` package.

stuartarchibald:
@david-zwicker what you have would work (doesn't need the `@register_jitable` on the inner function though, Numba just compiles that anyway). If you'd like to contribute this function then...

The use of Numba's extension API `@overload` decorator is strongly recommended for this task. A guide to using `@overload` is [here](http://numba.pydata.org/numba-doc/dev/extending/overloading-guide.html?highlight=overload) and API documentation is [here](http://numba.pydata.org/numba-doc/latest/extending/high-level.html#implementing-functions).

It may be useful to take a look at examples of pull requests adding support for a NumPy function that have already been done if guidance is needed... here's a couple:

* `np.roll`: https://github.com/numba/numba/pull/3563/files
* `np.roots`: https://github.com/numba/numba/pull/2135/files

The `np.heaviside` function implemenation should probably go in `numba/np/arraymath.py` and tests in `numba/tests/test_np_functions.py`. Thanks!

PR #6552, opened 2020-12-06
Added numba support for np.heaviside

david-zwicker:
Implements Numba support for `np.heaviside`.

Closes https://github.com/numba/numba/issues/4706

### DIFF
--- a/docs/source/reference/numpysupported.rst
+++ b/docs/source/reference/numpysupported.rst
@@ -467,7 +467,7 @@ The following top-level functions are supported:
 * :func:`numpy.full_like` (only the 3 first arguments)
 * :func:`numpy.hamming`
 * :func:`numpy.hanning`
-* :func:`numpy.heaviside`
+* :func:`numpy.heaviside` (only the 2 first arguments)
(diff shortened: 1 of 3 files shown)


david-zwicker:
It seems as if some checks concerning the documentation fail for this pull request. However, I only added a single line to the documentation and I have no idea how this could trigger some error. Unfortunately, the errors are incomprehensible to me. Does anyone with more insight into the package have any idea what I would need to fix?

stuartarchibald:
Thanks for the PR @david-zwicker, the log looks like something is upset by the use of `@vectorize`, I'll take a look.[

stuartarchibald:
Also, makes me wonder, is `heaviside` a true `ufunc`? If so might need some adjustment to get the `ufunc` mechanics working.

david-zwicker:
As far as I understand `np.heaviside` is a proper ufunc (see [documentation](https://numpy.org/doc/stable/reference/generated/numpy.heaviside.html)). I tried to implement it as a ufunc in Numba and thus used the `vectorize` decorator [as per the documentation](https://numba.pydata.org/numba-doc/dev/user/vectorize.html).

PR #6573, opened 2020-12-13
Improve __str__ for typed.List when invoked from IPython shell

### LINKED ISSUE #6434: Typed List causes IPython console to hang when gets printed
## Reporting a bug

All elements of `numba`'s typed `List` get printed to `IPython` console, causing it to crash if the list is long. This doesn't happen with Python's native lists, as they get shortened when printed.

```python
In [1]: x = [1] * 1_000_000
In [2]: y = numba.typed.List(x)
In [3]: y
```

ma-sadeghi:
Closes #6434 

This PR improves `typed.List`'s `__str__` method when invoked from IPython shell. Basically, it only prints the first 1,000 elements of the typed List.

Question from numba devs: I wrote a unit test for this, but I don't know how to fake an IPython shell so we can test this. For now, I put it inside a `try`/`except` clause. One idea is to do the trimming for the Python shell as well (similar to what numpy does).

### DIFF
--- a/numba/tests/test_typedlist.py
+++ b/numba/tests/test_typedlist.py
@@ -542,14 +542,15 @@ class TestTypedList(MemoryLeakMixin, TestCase):
             subprocess.check_output(base_cmd + ["--version"])
         except subprocess.CalledProcessError as e:
             self.skipTest("ipython not found: return code %d" % e.returncode)
-        cmd = base_cmd + " ".join(
+        repr_cmd = [" ".join(
             [
                 "import sys;",
                 "from numba.typed import List;",
                 "res = repr(List(range(1005)));",
-                "import sys; sys.stderr.write(res);"
+                "sys.stderr.write(res);"
             ]
-        )
+        )]
+        cmd = base_cmd + repr_cmd
         p = subprocess.Popen(
             cmd,
             stdout=subprocess.PIPE,
--- a/numba/tests/test_typedlist.py
+++ b/numba/tests/test_typedlist.py
@@ -14,7 +14,6 @@ from numba.typed import List, Dict
 from numba.core.errors import TypingError
 from numba.tests.support import (TestCase, MemoryLeakMixin, override_config,
                                  forbid_codegen, skip_parfors_unsupported)
-
 from numba.core.unsafe.refcount import get_refcount
 from numba.experimental import jitclass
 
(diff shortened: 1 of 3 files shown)


caljrobe:
@ma-sadeghi I was not expecting you to pull given how long it had been. As a result, yesterday I went ahead and implemented the feature and created a test. I spun up a `suprocess.run` and used the `stdin` argument to pass commands, I found this to simulate an interactive shell. If you would like to implement feel free. Alternatively, I could make a pull request with the work I did. No hard feelings either way.

ma-sadeghi:
@caljrobe I think the trimming is only supposed to be done just for IPython shells, not any interactive shell (probably because large pure Python lists are also only trimmed when invoked from IPython shells). I might also be mistaken though. @stuartarchibald could you please clarify?

esc:
@caljrobe thank you for implementing this. I have marked it as ready for review.

stuartarchibald:
> @caljrobe I think the trimming is only supposed to be done just for IPython shells, not any interactive shell (probably because large pure Python lists are also only trimmed when invoked from IPython shells). I might also be mistaken though. @stuartarchibald could you please clarify?

Think it's just for `ipython` shells (and their derivatives). The `python` "shell" just prints what it's asked and doesn't trim like the `ipython` shells.

numba/tests/test_typedlist.py
```diff
@@ -530,6 +530,16 @@ def test_repr(self):
         expected = "ListType[int32]([1, 2, 3])"
         self.assertEqual(expected, repr(l))
 
+    def test_repr_for_ipython_shells(self):
+        try:
+            get_ipython
```
esc, line 535: Forgive me if I missed something here, but why is this line needed?

numba/typed/typedlist.py
```diff
@@ -394,7 +394,12 @@ def __str__(self):
         buf = []
         for x in self:
             buf.append("{}".format(x))
-        return '[{0}]'.format(', '.join(buf))
+        # Check whether the code was invoked from IPython shell
+        try:
+            get_ipython
```
esc, line 399: Same comment as above.

esc (changes requested):
Left some comments to be addressed.

numba/tests/test_typedlist.py
ma-sadeghi: Thanks for your review. From what I understood, we wanted to trim the `__str__` for large typed Lists _only_ when invoked from an `ipython` shell. `get_ipython` is only available (I think) when invoked from an `ipython` shell, otherwise a `NameError` is raised, which is an indicator that we should fall back to the original `__str__`, i.e. printing the entire typed List no matter the length. Does that make sense?

esc: I see, that is interesting. I wonder why this isn't being flagged by flake8?

esc (changes requested):
I tried this manually from my ipython shell and it seems to work  well! For test, though, I think the test only runs when executed within ipython? Would it be possible to mock that magic string in such a way as to make sure the ipython representation is as expected, but without actually running it within ipython? I'd like the test to run as part of the regular suite too, if it's not too much hassle.

stuartarchibald:
@esc there's a test-time dependency on `ipython` in the conda recipe and some other tests requiring `ipython` in `test_dispatcher.py`, so I think it'd be ok to just do similar here?

esc:
@stuartarchibald probably, yes, but then I would remove the silent error exception to ensure it does actually run? Or maybe check if ipython is available and skip the test explicitly if not?

stuartarchibald:
> @stuartarchibald probably, yes, but then I would remove the silent error exception to ensure it does actually run? Or maybe check if ipython is available and skip the test explicitly if not?

I'd do the latter, cf. tests in `test_dispatcher.py`.

ma-sadeghi:
@stuartarchibald @esc Following @stuartarchibald's advice, I added a unit test for `repr` of long typed Lists similar to `test_ipython` in `test_dispatcher.py`. I put the unit test under the same file. I wonder if you want it under `test_typedlist.py`.

stuartarchibald:
> @stuartarchibald @esc Following @stuartarchibald's advice, I added a unit test for `repr` of long typed Lists similar to `test_ipython` in `test_dispatcher.py`. I put the unit test under the same file. I wonder if you want it under `test_typedlist.py`.

Thanks @ma-sadeghi! As you suggest, I'd go with putting it in the `test_typedlist.py` file to keep all the `typed.List` things together. Thanks again!

stuartarchibald:
There's a couple of failures in CI, one is `flake8` compliance, the other is weird:
https://dev.azure.com/numba/numba/_build/results?buildId=7545&view=logs&j=12ecc1d6-2964-5850-b419-89f950957853&t=48c87e6f-7efc-5267-a2f7-8be9a2382c49&l=9486
I'm taking a look.

stuartarchibald:
Saving the failure:
```
# Now run the Numba test suite with slicing
# Note that coverage is run from the checkout dir to match the "source"
# directive in .coveragerc
echo "INFO: Running slice of discovered tests: ($TEST_START_INDEX,None,$TEST_COUNT)"
if [ "$RUN_COVERAGE" == "yes" ]; then
    export PYTHONPATH=.
    coverage erase
    $SEGVCATCH coverage run runtests.py -b -j "$TEST_START_INDEX,None,$TEST_COUNT" --exclude-tags='long_running' -m $TEST_NPROCS -- numba.tests
elif [ "$RUN_TYPEGUARD" == "yes" ]; then
    echo "INFO: Running with typeguard"
    NUMBA_USE_TYPEGUARD=1 NUMBA_ENABLE_CUDASIM=1 $SEGVCATCH python runtests.py -b -j "$TEST_START_INDEX,None,$TEST_COUNT" --exclude-tags='long_running' -m $TEST_NPROCS -- numba.tests
else
    NUMBA_ENABLE_CUDASIM=1 $SEGVCATCH python -m numba.runtests -b -j "$TEST_START_INDEX,None,$TEST_COUNT" --exclude-tags='long_running' -m $TEST_NPROCS -- numba.tests
fi
INFO: Running slice of discovered tests: (9,None,20)
cc1plus: warning: command line option '-Wstrict-prototypes' is valid for C/ObjC but not for C++
skipped HSA tests
skipped HSA tests
Parallel: 434. Serial: 70
................................................................................................s.............................F.....................................................s............s...s..ssssssssssssssssss........................................................................................................................................................................................................................ssss...sss..........s.s.s..s..ss.s...ss.....s...ssssssssss..ssssssssss
FAIL: test_inline_asm (numba.tests.test_nrt.TestRefCtPruning)
The InlineAsm class from llvmlite.ir has no 'name' attr the refcount
  File "/home/vsts/work/1/s/numba/tests/test_nrt.py", line 550, in test_inline_asm
    self.assertEqual(foo(10), 22) # expect (10 + 1) * 2 = 22
AssertionError: 4 != 22

Ran 504 tests in 181.278s

FAILED (failures=1, skipped=59)
```
can't get this to reproduce locally. Will re-run.

ma-sadeghi:
@stuartarchibald Done

ma-sadeghi:
@stuartarchibald It seems that the tests are finally passing.

numba/tests/test_typedlist.py
```diff
@@ -530,6 +533,60 @@ def test_repr(self):
...
 
+    def test_repr_long_list_ipython(self):
+        # Test repr of long typed Lists in an IPython session
+        base_cmd = [sys.executable, '-m', 'IPython']
+        base_cmd += ['--quiet', '--quick', '--no-banner', '--colors=NoColor']
+        try:
+            ver = subprocess.check_output(base_cmd + ['--version'])
+        except subprocess.CalledProcessError as e:
+            self.skipTest("ipython not available: return code %d"
+                          % e.returncode)
```
esc, line 544: I see this is very similar to the code in : https://github.com/numba/numba/blob/master/numba/tests/test_dispatcher.py#L1418-L1425 I wonder if this can be turned into something re-usable. Perhaps to be added to: https://github.com/numba/numba/blob/master/numba/tests/support.py

esc (changes requested):
It's looking! Only one comment I left to address.

numba/tests/test_typedlist.py
esc: To clarify: the question is if this is a good idea or even possible? If it turns out, that attempting a re-use unnecessarily complicates matters that insight will have value in itself and the code can stay the way it is. Hope that is clear?

numba/tests/test_typedlist.py
```diff
@@ -530,6 +533,60 @@ def test_repr(self):
...
+        ver = ver.strip().decode()
+        print("ipython version:", ver)
```
stuartarchibald, line 546: Please delete this, it adds noise to test output.

numba/tests/test_typedlist.py
```diff
@@ -530,6 +533,60 @@ def test_repr(self):
...
+            # Feed the test input as stdin, to execute it in REPL context
+            with open(inputfn, "rb") as stdin:
+                p = subprocess.Popen(base_cmd, stdin=stdin,
+                                     stdout=subprocess.PIPE,
+                                     stderr=subprocess.PIPE,
+                                     universal_newlines=True)
+                out, err = p.communicate()
+            return err
+
+        out = execute_repr_long_list_in_ipython()
```
stuartarchibald, line 582: Curious about the need for a file (and it should really be file from `tempfile`?), this works on linux, is there something strange going on with stdout/stderr etc elsewhere?:
```python
cmd = [sys.executable, *'-m IPython --quiet --quick --no-banner --colors=NoColor -c'.split(), "from numba.typed import List; repr(List(range(1005)))"]
CPres = subprocess.run(cmd, capture_output=True, check=True)
print(CPres.stdout.decode('utf-8'))
```

numba/tests/test_typedlist.py
```diff
@@ -530,6 +533,60 @@ def test_repr(self):
...
+
+        l_long = List(range(1005))
+        l_long_str = [str(item) for item in l_long]
```
stuartarchibald, line 585: Wondering if this need to go via a typed list, is it just used for building a string repr of the range as a list?

stuartarchibald (commented):
This looks good, just a few minor comments. Thanks for the patch.

numba/tests/test_typedlist.py
ma-sadeghi: Thanks for your review. Just to clarify, you mean lines 545-546, right?

esc: I think @stuartarchibald means the `print` statement.

numba/tests/test_typedlist.py
ma-sadeghi: It's because of the dtype showing at the beginig of `repr`, which for Python lists is "reflected list" or something like that.

numba/tests/test_typedlist.py
ma-sadeghi: Thanks for your review. Indeed, it works! and it's much cleaner now. I initially tried to mimic the other unit test where IPython is also being used. However, this is much cleaner. Thanks!

numba/tests/test_typedlist.py
ma-sadeghi: @esc Thanks, refactored the unit test and removed the `print` statement.

numba/tests/test_typedlist.py
ma-sadeghi: Thank you for your review. We could wrap this up as a separate method, like `check_ipython_available`, but I wonder if it's worth the trouble. After all, we're just using it in only two places.

ma-sadeghi:
@stuartarchibald I tested the patch locally and the unit tests passed. For some reason Azure is struggling tonight.

esc:
@ma-sadeghi thank you for following up on this. I had another look at this PR right now and it appears as though you may have accidentally committed commenting out large parts of the test code? I presume this may have been done to ease local testing? In any case, the `flake8` style checker now reports the following:


```
numba/tests/test_typedlist.py:4:1: F401 'textwrap.dedent' imported but unused
numba/tests/test_typedlist.py:8:1: F401 'numba.config' imported but unused
numba/tests/test_typedlist.py:13:1: F401 'numba.typed.Dict' imported but unused
numba/tests/test_typedlist.py:17:1: F401 'numba.core.unsafe.refcount.get_refcount' imported but unused
numba/tests/test_typedlist.py:18:1: F401 'numba.experimental.jitclass' imported but unused
```

I presume this is because the code in question that was using the imports is now commented out. I would suggest adding another commit to remove the comment characters from the file in question and then we can run Azure again. Thank you!

ma-sadeghi:
> @ma-sadeghi thank you for following up on this. I had another look at this PR right now and it appears as though you may have accidentally committed commenting out large parts of the test code? I presume this may have been done to ease local testing? In any case, the `flake8` style checker now reports the following:
> 
> ```
> numba/tests/test_typedlist.py:4:1: F401 'textwrap.dedent' imported but unused
> numba/tests/test_typedlist.py:8:1: F401 'numba.config' imported but unused
> numba/tests/test_typedlist.py:13:1: F401 'numba.typed.Dict' imported but unused
> numba/tests/test_typedlist.py:17:1: F401 'numba.core.unsafe.refcount.get_refcount' imported but unused
> numba/tests/test_typedlist.py:18:1: F401 'numba.experimental.jitclass' imported but unused
> ```
> 
> I presume this is because the code in question that was using the imports is now commented out. I would suggest adding another commit to remove the comment characters from the file in question and then we can run Azure again. Thank you!

Oops, sorry my bad! I just uncommented those lines.

numba/tests/test_typedlist.py
ma-sadeghi: @esc Can I mark this one as resolved?

ma-sadeghi:
@stuartarchibald It seems that finally all tests are passing.

numba/tests/test_typedlist.py
esc: OK, fair enough, maybe next time someone touches this.

esc:
@ma-sadeghi thank you again for fixing this up, I have added it to the queue for review once again.

esc (approved):
LGTM! 👍

esc:
@stuartarchibald you happy too?

stuartarchibald:
@esc patch fixes reported issue, let's merge. Thanks for implementing @ma-sadeghi .

stuartarchibald:
@ma-sadeghi Congratulations on your first contribution to Numba!

esc:
🎉

esc:
@ma-sadeghi to clarify on the process here, the PR has now been merged as ready-to-merge, this means it will be merged during the next merge window (probably later on today).

ma-sadeghi:
> @ma-sadeghi Congratulations on your first contribution to Numba!

Thanks! I'm happy I could be of help 😄

> @ma-sadeghi to clarify on the process here, the PR has now been merged as ready-to-merge, this means it will be merged during the next merge window (probably later on today).

Great, thanks for letting me know 👍

PR #6574, opened 2020-12-13
Run parfor fusion if 2 or more parfors

ehsantn:
Run parfor fusion only if there are 2 or more parfors in the function. Also, eliminates an unnecessary `simplify` call. Manual testing shows this makes compilation time spent in parfor pass shorter.

### DIFF
--- a/numba/parfors/parfor.py
+++ b/numba/parfors/parfor.py
@@ -2741,7 +2741,7 @@ class ParforPass(ParforPassStates):
 
         # simplify CFG of parfor body loops since nested parfors with extra
         # jumps can be created with prange conversion
-        simplify_parfor_body_CFG(self.func_ir.blocks)
+        n_parfors = simplify_parfor_body_CFG(self.func_ir.blocks)
         # simplify before fusion
         simplify(self.func_ir, self.typemap, self.calltypes, self.metadata["parfors"])
         # need two rounds of copy propagation to enable fusion of long sequences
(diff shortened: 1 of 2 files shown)


DrTodd13 (dismissed):

stuartarchibald:
This is breaking: `numba.tests.test_array_analysis.TestArrayAnalysisInterface.test_array_analysis_extensions`
https://dev.azure.com/numba/numba/_build/results?buildId=7485&view=logs&j=4f8e7448-6bbb-53dd-1446-49d51b72f33b&t=22d62e29-1b6b-5406-981d-1aa65fc1883c&l=9468
think it's because the test is expecting array analysis to run and now it won't as there's only a single parfor in the test source.

ehsantn:
Thanks @stuartarchibald . Should be fixed now.

stuartarchibald (approved):
Thanks for the patch and fixes. @DrTodd13 are you happy with this change?

DrTodd13:
> Thanks for the patch and fixes. @DrTodd13 are you happy with this change?

Yes, I already approved it.

stuartarchibald:
> > Thanks for the patch and fixes. @DrTodd13 are you happy with this change?
> 
> Yes, I already approved it.

Thanks for confirming @DrTodd13. Due to recent changes, approvals are automatically revoked if new patches are pushed. In the history it seems like revoked approvals just appear as a "review", I'll watch out for this pattern again!

Issue #6543, opened 2020-12-03
Operation "element in set" never returns


icezyclon:
Bug: After a series of add and remove operations on a set in nopython mode the operation "in set" does not return and gets stuck at 100% CPU load.

Isolated code example:
```python
import numba as nb

@ nb.njit()
def problem():

    # init
    openSet = set()

    openSet.add((0, 0))
    openSet.add((1, 0))
    openSet.add((2, 0))
    openSet.add((3, 0))
    openSet.add((4, 0))
    openSet.add((5, 0))
    openSet.add((6, 0))
    openSet.add((7, 0))
    openSet.add((8, 0))

    openSet.remove((2, 0))
    openSet.add((2, 1))
    openSet.remove((2, 1))
    openSet.remove((4, 0))
    openSet.add((5, 1))
    openSet.add((4, 1))
    openSet.remove((7, 0))
    openSet.add((8, 1))
    openSet.add((7, 1))
    openSet.add((6, 1))
    openSet.remove((8, 0))
    openSet.remove((1, 0))
    openSet.remove((0, 0))
    openSet.remove((5, 0))
    openSet.remove((6, 0))
    openSet.remove((3, 0))
    openSet.remove((7, 1))
    openSet.add((8, 2))
    openSet.add((7, 2))
    openSet.add((6, 2))
    openSet.remove((8, 1))
    openSet.remove((6, 1))
    openSet.add((5, 2))
    openSet.remove((4, 1))
    openSet.add((4, 2))
    openSet.remove((5, 1))
    openSet.remove((6, 2))
    openSet.add((7, 3))
    openSet.add((6, 3))
    openSet.add((5, 3))
    openSet.remove((7, 2))
    openSet.add((8, 3))
    openSet.remove((8, 2))
    openSet.remove((4, 2))
    openSet.add((4, 3))
    openSet.add((3, 3))
    openSet.remove((5, 2))
    openSet.remove((7, 3))
    openSet.add((8, 4))
    openSet.add((7, 4))
    print('About to check (6,4) in openSet')
    return (6, 4) in openSet


if __name__ == "__main__":
    print(problem())
    print('DONE')
```

The code gets stuck with numba 0.52.0 with Python 3.8.5 on Ubuntu 20.04.1 LTS x86_64.

However, @esc on gitter.im also confirmed that the issue persists on OSX.

Isolated code example:
https://github.com/icezyclon/astart-pathfinding/blob/main/isolated.py
Original code example (from A* search):
[...]

PR #6578, opened 2020-12-15
Add support for numpy intersect1d and basic test cases

caljrobe:
Add support for [np.intersect1d](https://numpy.org/doc/stable/reference/generated/numpy.intersect1d.html)
Add basic test cases
Add documentation in numpysupported.rst

If merged, will complete one of the listed tasks in #4074

### DIFF
--- a/docs/source/reference/numpysupported.rst
+++ b/docs/source/reference/numpysupported.rst
@@ -472,6 +472,7 @@ The following top-level functions are supported:
 * :func:`numpy.identity`
 * :func:`numpy.kaiser`
 * :func:`numpy.interp` (only the 3 first arguments; requires NumPy >= 1.10)
+* :func:`numpy.intersect1d` (only first 2 arguments, arr1 and arr1)
 * :func:`numpy.linspace` (only the 3-argument form)
 * :class:`numpy.ndenumerate`
 * :class:`numpy.ndindex`
--- a/docs/source/reference/numpysupported.rst
+++ b/docs/source/reference/numpysupported.rst
@@ -472,7 +472,7 @@ The following top-level functions are supported:
 * :func:`numpy.identity`
 * :func:`numpy.kaiser`
 * :func:`numpy.interp` (only the 3 first arguments; requires NumPy >= 1.10)
-* :func:`numpy.intersect1d` (only first 2 arguments, arr1 and arr1)
+* :func:`numpy.intersect1d` (only first 2 arguments, ar1 and ar1)
 * :func:`numpy.linspace` (only the 3-argument form)
 * :class:`numpy.ndenumerate`
 * :class:`numpy.ndindex`
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -3451,12 +3451,12 @@ def np_array_equal(a, b):
 @overload(np.intersect1d)
 def jit_np_intersect1d(ar1, ar2):
     # Not implemented to support assume_unique or return_indices
-    # return numpy array of sorted, unique values in both input arrays
+    # https://github.com/numpy/numpy/blob/v1.19.0/numpy/lib
+    # /arraysetops.py#L347-L441
     if not (type_can_asarray(ar1) or type_can_asarray(ar2)):
         raise TypingError('intersect1d: first two args must be array-like')
 
     def np_intersects1d_impl(ar1, ar2):
-        # Mirrored from: https://github.com/numpy/numpy/blob/v1.19.0/numpy/lib/arraysetops.py#L347-L441
         ar1 = np.asarray(ar1)
         ar2 = np.asarray(ar2)
 
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -3448,6 +3448,31 @@ def np_array_equal(a, b):
     return impl
 
 
+@overload(np.intersect1d)
+def jit_np_intersect1d(a, b):
+    # Not implemented to support assume_unique or return_indices
+    # return sorted, unique values in both arrays
+    if not (type_can_asarray(a) or type_can_asarray(b)):
+        raise TypingError('intersect1d: both inputs must be array-like')
+
+    def np_intersects1d_impl(a, b):
+        a = np.asarray(a)
+        b = np.asarray(b)
+        intersect = []
+        a_map = {}
+        for v in a:
+            if v not in a_map:
+                a_map[v] = ''
+        for v in b:
+            if v in a_map:
+                if a_map[v] != '.':
+                    # Don't want duplicates
+                    a_map[v] = '.'
+                    intersect.append(v)
+        return sorted(intersect)
+    return np_intersects1d_impl
+
+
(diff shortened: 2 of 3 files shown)


esc:
@caljrobe welcome to Numba and thank you for submitting this PR. I have added it to the queue for review.

numba/np/arraymath.py
```diff
@@ -3448,6 +3448,31 @@ def impl(a, b):
...
+        for v in a:
+            if v not in a_map:
+                a_map[v] = ''
+        for v in b:
+            if v in a_map:
+                if a_map[v] != '.':
+                    # Don't want duplicates
+                    a_map[v] = '.'
+                    intersect.append(v)
+        return sorted(intersect)
```
esc, line 3472: IIRC sorted will sort the list out of place, maybe if this were a Numpy array, the values could be sorted in place?

numba/np/arraymath.py
```diff
@@ -3448,6 +3448,31 @@ def impl(a, b):
...
+def jit_np_intersect1d(a, b):
+    # Not implemented to support assume_unique or return_indices
+    # return sorted, unique values in both arrays
+    if not (type_can_asarray(a) or type_can_asarray(b)):
+        raise TypingError('intersect1d: both inputs must be array-like')
+
+    def np_intersects1d_impl(a, b):
+        a = np.asarray(a)
+        b = np.asarray(b)
+        intersect = []
```
esc, line 3461: Please do check what NumPy returns. From what I can tell, the NumPy implementation returns a NumPy array, not a list. It might also be worth using the typing section to figure out what the type of the returned array should be?

esc (changes requested):
I have reviewed this PR and it will need some more work. Regarding the core implementation, it might be worth considering the following:

Does it make sense to use the `set` datatype under the hood? Maybe not now, because https://github.com/numba/numba/issues/6543 -- but if `set` were working correctly perhaps it is the best approach?

Also, I would strongly encourage you to look at the NumPy implementation at:
https://github.com/numpy/numpy/blob/v1.19.0/numpy/lib/arraysetops.py#L347-L441
and use that as a guide.

I left some comments also about some of the issues in the current implementation.

caljrobe:
@esc I appreciate your thorough feedback and apologize for all the mistakes. I will continue to work on the implementation, and in the future make sure that it is more properly developed and correct before I make a pull request for review.

esc:
@caljrobe no worries, no need to apologize for mistakes, we are all always learning, Numba is a beast and we appreciate all contributions!

numba/tests/test_np_functions.py
```diff
@@ -588,6 +592,18 @@ def test_array_equal_exception(self):
             str(raises.exception)
         )
 
+    def test_intersect1d(self):
+        pyfunc = intersect1d
+        cfunc = jit(nopython=True)(pyfunc)
+
+        def arrays():
+            yield [1], [1]
```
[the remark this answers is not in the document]
caljrobe: @esc When yielding empty lists, `ValueError: cannot compute fingerprint of empty list` is raised. I found that yielding NumPy arrays resolves this. But does this mean there is a issue in the implementation? Surely Python arrays are array-like and should be supported. I pushed the most recent commit that shares the same potential issue.

caljrobe:
@esc After a speed test, the cfunc is taking a long time when handling empty arrays. I am concerned as to why this is.

docs/source/reference/numpysupported.rst
```diff
@@ -472,6 +472,7 @@ The following top-level functions are supported:
 * :func:`numpy.identity`
 * :func:`numpy.kaiser`
 * :func:`numpy.interp` (only the 3 first arguments; requires NumPy >= 1.10)
+* :func:`numpy.intersect1d` (only first 2 arguments, ar1 and ar1)
```
esc, line 475:
```suggestion
* :func:`numpy.intersect1d` (only first 2 arguments, ar1 and ar2)
```

esc:
@caljrobe 

> @esc When yielding empty lists, ValueError: cannot compute fingerprint of empty list is raised. I found that yielding NumPy arrays resolves this. But does this mean there is a issue in the implementation? Surely Python arrays are array-like and should be supported. I pushed the most recent commit that shares the same potential issue.

The problem is that empty lists without a datatype can not be compiled since the compiler does not know what to do with imprecise types.

numba/tests/test_np_functions.py
```diff
@@ -588,6 +592,27 @@ def test_array_equal_exception(self):
...
+            yield np.array([1, 2]), np.array([1])
+            yield np.array([1, 2, 2]), np.array([2, 2])
+            yield np.array([1, 2]), np.array([2, 1])
+            yield np.array([1, 2, 3]), np.array([1, 2, 3])
+
+        pyfunc = intersect1d
+        cfunc = jit(nopython=True)(pyfunc)
+
+        for a, b in arrays():
+            expected = pyfunc(a, b)
```
esc, line 612: Maybe you could simplify this test by casting to Numpy array here, rather than above in the test case generator?

esc (changes requested):
Looking good, still some changes to be made. Regarding the slow-down, what sort of slow-down are you seeing?

caljrobe:
@esc For the first test case, a=[] and b=[], the pyfunc runs in 4^-5 second, however the cfunc takes 3.73 seconds. The second test case, a=[1] and b=[] runs in similar time. But, when running the same test case again, the cfunc performs faster than pyfunc as expected. Is this just optimizer overhead?

numba/tests/test_np_functions.py
caljrobe: That is definitely better, I have simplified the test. Thank you

esc:
@caljrobe yes, the first time a function is run, it will be compiled by Numba. So, when benchmarking Numba compiled functions, the first run (the compilation run) is to be disregarded. You can use caching to save the compiled function on disk if need be to largely circumvent this limitation.

esc (changes requested):
The code looks good so far but there is a flake8 failure reported by CI:

numba/np/arraymath.py:3459:81: E501 line too long (103 > 80 characters)


Please do fix and this should be ready to go.

esc:
@caljrobe thanks for the fixing, the PR is now waiting on public CI to complete.

caljrobe:
> @caljrobe thanks for the fixing, the PR is now waiting on public CI to complete.

@esc As you can probably tell, I am a new developer. Thank you for your patience!

esc (approved):

esc:
@caljrobe no worries at all. Practice makes perfect and patience is a virtue. The PR now passed all of public CI and has now been marked as ready for merge. It will be merged in the next few days.

esc:
@caljrobe I was able to merge this during the merge-window just now. Thank you again for your submission, your efforts to improve Numba are appreciated.

Aaaaaand, congrats on your first contribution! 🎉

caljrobe:
@esc Awesome, and thank you again!

Issue #6534, opened 2020-12-01
bytecode analysis failure with `BUILD_TUPLE_UNPACK`.


stuartarchibald:
## Reporting a bug

- [x] I have tried using the latest released version of Numba (most recent is
 visible in the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [x] I have included a self contained code sample to reproduce the problem.
  i.e. it's possible to run as 'python bug.py'.

```python
from numba import njit

@njit
def foo(x):
    return (*x,)

foo((1,2))
```
does this:
```
<path>/numba/core/byteflow.py in _build_tuple_unpack(self, state, inst)
    846         state.append(inst, tuples=tuples, temps=temps)
    847         # The result is in the last temp var
--> 848         state.push(temps[-1])
    849 
    850     def op_BUILD_TUPLE_UNPACK_WITH_CALL(self, state, inst):

IndexError: Failed in nopython mode pipeline (step: analyzing bytecode)
list index out of range
```

sklam:
bug confirmed on py3.7

stuartarchibald:
yeah, it's been around a while, just hit it whilst trying to work out whether to peephole the python 3.9 bytecode back to using this or workout how to make lists into tuples.

PR #6579, opened 2020-12-15
Python 3.9 support.

stuartarchibald:
Adds support for Python 3.9, and fixes a few bugs in the process.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -123,7 +123,7 @@ jobs:
         TEST_START_INDEX: 17
       py39_np115:
         PYTHON: '3.9'
-        NUMPY: '1.15'
+        NUMPY: '1.16'
         CONDA_ENV: travisci
         TEST_START_INDEX: 18
       py39_np119:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -3,7 +3,7 @@ trigger:
 
 variables:
   # Change the following along with adding new TEST_START_INDEX.
-  TEST_COUNT: 20
+  TEST_COUNT: 22
 
 jobs:
 # Mac and Linux use the same template with different matrixes
@@ -121,6 +121,16 @@ jobs:
         NUMPY: '1.19'
         CONDA_ENV: travisci
         TEST_START_INDEX: 17
+      py39_np115:
+        PYTHON: '3.9'
+        NUMPY: '1.15'
+        CONDA_ENV: travisci
+        TEST_START_INDEX: 18
+      py39_np119:
+        PYTHON: '3.9'
+        NUMPY: '1.19'
+        CONDA_ENV: travisci
+        TEST_START_INDEX: 19
 
 - template: buildscripts/azure/azure-windows.yml
   parameters:
--- a/numba/core/interpreter.py
+++ b/numba/core/interpreter.py
@@ -1654,7 +1654,17 @@ class Interpreter(object):
         # build_list, drop the extend, and wire up the target as the result from
         # the build_tuple that's been rewritten.
 
-        # is last emitted statment a build_tuple?
+        # See if this is the first statement in a block, if so its probably from
+        # control flow in a tuple unpack like:
+        # `(*(1, (2,) if predicate else (3,)))`
+        # this cannot be handled as present so raise
+        if not self.current_block.body:
+            msg = ("An unsupported bytecode sequence has been encountered: "
+                   "op_LIST_EXTEND at the start of a block.\n\nThis could be "
+                   "due to the use of a branch in a tuple unpacking statement.")
+            raise errors.UnsupportedError(msg)
+
+        # is last emitted statement a build_tuple?
         stmt = self.current_block.body[-1]
         ok = isinstance(stmt.value, ir.Expr) and stmt.value.op == "build_tuple"
         # check statements from self.current_block.body[-1] through to target,
--- a/numba/core/interpreter.py
+++ b/numba/core/interpreter.py
@@ -197,7 +197,8 @@ def peep_hole_list_to_tuple(func_ir):
                         if isinstance(x.value, ir.Expr):
                             expr = x.value
                             if expr.op == 'getattr':
-                                if expr.attr in ('extend', 'append'):
+                                if (x.target.name in extends or
+                                        x.target.name in appends):
                                     # drop definition, it's being wholesale
                                     # replaced.
                                     func_ir._definitions.pop(x.target.name)
@@ -228,7 +229,8 @@ def peep_hole_list_to_tuple(func_ir):
                                     asgn = ir.Assign(new, x.target, expr.loc)
                                     append_and_fix(asgn)
                                     acc = asgn.target
-                            elif expr.op == 'build_list':
+                            elif (expr.op == 'build_list' and
+                                    x.target.name == const_list):
                                 new = ir.Expr.build_tuple(expr.items, expr.loc)
                                 asgn = ir.Assign(new, x.target, expr.loc)
                                 # Not a temporary any more
--- a/numba/core/interpreter.py
+++ b/numba/core/interpreter.py
@@ -160,6 +160,7 @@ def peep_hole_list_to_tuple(func_ir):
                                     assert 0
                 # go back through the peep hole build new IR based on it.
                 new_hole = []
+
                 def append_and_fix(x):
                     """ Adds to the new_hole and fixes up definitions"""
                     new_hole.append(x)
--- a/numba/core/interpreter.py
+++ b/numba/core/interpreter.py
@@ -205,6 +205,10 @@ def peep_hole_list_to_tuple(func_ir):
                                     # replaced.
                                     func_ir._definitions.pop(x.target.name)
                                     continue
+                                else:
+                                    # a getattr on something we're not
+                                    # interested in
+                                    new_hole.append(x)
                             elif expr.op == 'call':
                                 fname = expr.func.name
                                 if fname in extends or fname in appends:
@@ -231,6 +235,10 @@ def peep_hole_list_to_tuple(func_ir):
                                     asgn = ir.Assign(new, x.target, expr.loc)
                                     append_and_fix(asgn)
                                     acc = asgn.target
+                                else:
+                                    # there could be a call in the unpack, like
+                                    # *(a, x.append(y))
+                                    new_hole.append(x)
                             elif (expr.op == 'build_list' and
                                     x.target.name == const_list):
                                 new = ir.Expr.build_tuple(expr.items, expr.loc)
--- a/numba/core/interpreter.py
+++ b/numba/core/interpreter.py
@@ -1482,12 +1482,9 @@ class Interpreter(object):
             self._binop(op, lhs, rhs, res)
 
     def op_IS_OP(self, inst, lhs, rhs, res):
-        self._binop('is', lhs, rhs, res)
         # invert if op case is 1
-        if inst.arg == 1:
-            tmp = self.get(res)
-            out = ir.Expr.unary('not', value=tmp, loc=self.loc)
-            self.store(out, res)
+        op = 'is not' if inst.arg == 1 else 'is'
+        self._binop(op, lhs, rhs, res)
 
     def op_CONTAINS_OP(self, inst, lhs, rhs, res):
         lhs, rhs = rhs, lhs
--- a/numba/core/interpreter.py
+++ b/numba/core/interpreter.py
@@ -133,14 +133,11 @@ class Interpreter(object):
         self.syntax_blocks = []
         self.dfainfo = None
 
-        firstblk = min(self.cfa.blocks.keys())
         self.scopes.append(ir.Scope(parent=self.current_scope, loc=self.loc))
         # Interpret loop
         for inst, kws in self._iter_inst():
             self._dispatch(inst, kws)
-
         self._legalize_exception_vars()
-
         # Prepare FunctionIR
         fir = ir.FunctionIR(self.blocks, self.is_generator, self.func_id,
                              self.first_loc, self.definitions,
@@ -736,6 +733,14 @@ class Interpreter(object):
                 target = self.store(val_const, name=nm, redefine=True)
                 st.append(target)
             const = ir.Expr.build_tuple(st, loc=self.loc)
+        elif isinstance(value, frozenset):
+            st = []
+            for x in value:
+                nm = '$const_%s' % str(x)
+                val_const = ir.Const(x, loc=self.loc)
+                target = self.store(val_const, name=nm, redefine=True)
+                st.append(target)
+            const = ir.Expr.build_set(st, loc=self.loc)
         else:
             const = ir.Const(value, loc=self.loc)
         self.store(const, res)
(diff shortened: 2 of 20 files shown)


numba/core/ir.py
```diff
@@ -266,6 +266,8 @@ def get(self, name):
         try:
             return self._con[name]
         except KeyError:
+            #import pdb; pdb.set_trace()
```
stuartarchibald, line 269: remove this

numba/cpython/listobj.py
```diff
@@ -673,6 +673,14 @@ def sequence_bool_impl(seq):
     return context.compile_internal(builder, sequence_bool_impl, sig, args)
 
 
+@overload(operator.truth)
+def sequence_truth(seq):
+    if isinstance(seq, types.Sequence):
+        def impl(seq):
+            return len(seq) != 0
+        return impl
```
stuartarchibald, line 681: Don't remember why I added this, remove and see if anything is broken?

numba/core/interpreter.py
```diff
@@ -892,6 +897,114 @@ def op_BUILD_TUPLE_UNPACK_WITH_CALL(self, inst, tuples, temps):
...
+                # stick everything else in as-is
+                local_store(x)
+        # write the result back into the original build list as everything
+        # refers to it.
+        local_store(ir.Assign(acc, the_build_list, the_build_list.loc))
+
+        cpy = self.current_block.body[:]
+        tmp = cpy[:len(cpy) -count -2]
+        tmp.extend(new_hole)
+        self.current_block.body = tmp
```
sklam, line 1006: I wanted to move the peephole into a post-processing immediately after the bytecode->IR. Doing it here makes it hard to debug and the IR mutation here is not ideal.

numba/core/interpreter.py
```diff
@@ -1283,6 +1423,29 @@ def op_JUMP_IF_FALSE_OR_POP(self, inst, pred):
...
 
+    def op_JUMP_IF_NOT_EXC_MATCH(self, inst, pred, tos, tos1):
+        truebr = inst.next
+        falsebr = inst.get_jump_target()
+        gv_fn = ir.Global(
+            "exception_match", eh.exception_match, loc=self.loc,
+        )
+        exc_match_name = '$exc_match'
+        self.store(value=gv_fn, name=exc_match_name, redefine=True)
+        # not wholly convinced by this as tos is seemingly None?!
```
sklam, line 1434: How are you seeing that TOS is None?

sklam: When an exception is raised and the `except` handler is entered, (IIRC) it injects 6 stack values.

azure-pipelines.yml
```diff
@@ -121,6 +121,16 @@ jobs:
         NUMPY: '1.19'
         CONDA_ENV: travisci
         TEST_START_INDEX: 17
+      py39_np115:
+        PYTHON: '3.9'
+        NUMPY: '1.15'
```
seibert, line 126: There is no Python 3.9 build of NumPy 1.15.  Will need to bump this to 1.16.

seibert:
The pip problem (at least with np119) seems to be this line:

https://github.com/numba/numba/blob/master/buildscripts/incremental/build.sh#L20

Does not work anymore with the newer pip that is installed with Python 3.9.  I'm curious if we could switch that to:
```
python -m pip install --no-deps -e .
```
Which seems to be the more direct way of achieving what the command is trying to do?

azure-pipelines.yml
stuartarchibald: done, thanks

stuartarchibald:
> The pip problem (at least with np119) seems to be this line:
> 
> https://github.com/numba/numba/blob/master/buildscripts/incremental/build.sh#L20
> 
> Does not work anymore with the newer pip that is installed with Python 3.9. I'm curious if we could switch that to:
> 
> ```
> python -m pip install --no-deps -e .
> ```
> 
> Which seems to be the more direct way of achieving what the command is trying to do?

Thanks, trying in  d180309

numba/core/interpreter.py
stuartarchibald: Agree, now we know this works, will delay peephole rewrites to the point where there's actual tools available to help with doing it!

sklam:
`numba.tests.test_unicode.TestUnicode.test_strip` is failing and the minimal reproducer is:

```python
from numba import njit
@njit
def foo(x):
    return x.rstrip()
foo("asic")
```

stuartarchibald:
> `numba.tests.test_unicode.TestUnicode.test_strip` is failing and the minimal reproducer is:
> 
> ```python
> from numba import njit
> @njit
> def foo(x):
>     return x.rstrip()
> foo("asic")
> ```

Thanks. Raw MWR:
```python
from numba import njit

@njit
def foo(x, y=None):
    if y is not None:
        x in y

foo("a")
```
looks like `IS` handling has a bug or branch pruning needs an update. The branch should be killed as the predicate can be evaluated at compile time as false.

stuartarchibald:
2434b65 is a bit scrappy at present, but passing CI.

stuartarchibald:
Ran whole test suite locally, one failure:
`numba.tests.test_parfors.TestParfors.test_no_hoisting_with_member_function_call`
with
```
llvmlite/ir/instructions.py", line 85, in __init__
    raise TypeError(msg)
Type of #3 arg mismatch: double != i64

File "numba/tests/test_parfors.py", line 1464:
        def test_impl(X):
            <source elided>
            for i in prange(n):
                R = {1, 2, 3}
                ^

During: lowering "$32set_update.5 = call $32set_update.4($const30.3, func=$32set_update.4, args=(Var($const30.3, test_parfors.py:1464),), kws=(), vararg=None)" 
```

stuartarchibald:
Failure is due to:
```python
from numba import njit, prange, types
import numpy as np

@njit(parallel=True)
def test_impl():
    for i in prange(1):
        R = {1, 2, 3} # int64 set
        R.add(i) # add uint64 induction var, coerces `R` to float64, parfors fails to handle this
        # R.add(types.int64(i)) # this works as the type of the induction var matches the set type.
```
Python 3.9 bytecode:
```
  6           0 LOAD_GLOBAL              0 (prange)
              2 LOAD_CONST               1 (1)
              4 CALL_FUNCTION            1
              6 GET_ITER
        >>    8 FOR_ITER                28 (to 38)
             10 STORE_FAST               0 (i)

  7          12 BUILD_SET                0
             14 LOAD_CONST               2 (frozenset({1, 2, 3}))
             16 SET_UPDATE               1
             18 STORE_FAST               1 (R)

  8          20 LOAD_FAST                1 (R)
             22 LOAD_METHOD              1 (add)
             24 LOAD_GLOBAL              2 (types)
             26 LOAD_METHOD              3 (int64)
             28 LOAD_FAST                0 (i)
             30 CALL_METHOD              1
             32 CALL_METHOD              1
             34 POP_TOP
             36 JUMP_ABSOLUTE            8
        >>   38 LOAD_CONST               0 (None)
             40 RETURN_VALUE
```
Python 3.8 bytecode:
```
  6           0 LOAD_GLOBAL              0 (prange)
              2 LOAD_CONST               1 (1)
              4 CALL_FUNCTION            1
              6 GET_ITER
        >>    8 FOR_ITER                30 (to 40)
             10 STORE_FAST               0 (i)

  7          12 LOAD_CONST               1 (1)
             14 LOAD_CONST               2 (2)
             16 LOAD_CONST               3 (3)
             18 BUILD_SET                3
             20 STORE_FAST               1 (R)

  8          22 LOAD_FAST                1 (R)
             24 LOAD_METHOD              1 (add)
             26 LOAD_GLOBAL              2 (types)
             28 LOAD_METHOD              3 (int64)
             30 LOAD_FAST                0 (i)
             32 CALL_METHOD              1
             34 CALL_METHOD              1
             36 POP_TOP
             38 JUMP_ABSOLUTE            8
        >>   40 LOAD_CONST               0 (None)
             42 RETURN_VALUE
```

stuartarchibald:
This failure

> `numba.tests.test_parfors.TestParfors.test_no_hoisting_with_member_function_call`

Is due to https://github.com/numba/numba/issues/6621, this is not a new problem, just more evident as a result of the way Python 3.9 `SET_*` opcodes are currently translated.

stuartarchibald:
TODO:
* Decide how to fix https://github.com/numba/numba/pull/6579#issuecomment-755212243
* Add test for unpack operation with predicate and make sure failure is graceful.
* Decide what to do with https://github.com/numba/numba/issues/6534
* Clean up the patch.
* ...

numba/core/interpreter.py
```diff
@@ -80,6 +78,229 @@ def get_assignment_source(self, destname):
...
+    # For all blocks
+    for offset, blk in func_ir.blocks.items():
+        target_list = None
+        # keep doing the peephole rewrite until nothing is left that matches
+        while True:
+            found = False
+            region = None
+            # first try and find a matching region
+            # i.e. BUILD_LIST...<stuff>...LIST_TO_TUPLE
+            for idx, stmt in enumerate(reversed(blk.body)):
```
sklam, line 124: Kind of a brainteaser that idx is the negative index and later it needs to be `-region[1][0] - 1`. It'll be clearer as positive indices. My suggestion: https://github.com/sklam/numba/commit/492222d0e324e8bacd0c3fa1e08263b38db2eaae, (I also moved the code into its a subroutine in the patch)

numba/core/interpreter.py
```diff
@@ -80,6 +78,229 @@ def get_assignment_source(self, destname):
...
+                # It also switches out the build_list for a build_tuple and then
+                # ensures everything is wired up and defined ok.
+                t2l_agn = region[0][1]
+                acc = the_build_list
+                for x in peep_hole:
+                    if isinstance(x, ir.Assign):
+                        if isinstance(x.value, ir.Expr):
+                            expr = x.value
+                            if expr.op == 'getattr':
+                                if expr.attr in ('extend', 'append'):
```
sklam, line 200: This didn't check that the expr is related to the the const list. My suggestion: https://github.com/sklam/numba/commit/960bc6783881d7971b1a1362d3ecfe6dc50a1a50

numba/core/interpreter.py
```diff
@@ -80,6 +78,229 @@ def get_assignment_source(self, destname):
...
+
+                                    # this needs to be a binary add
+                                    new = ir.Expr.binop(fn=operator.add,
+                                                        lhs=acc,
+                                                        rhs=arg,
+                                                        loc=x.loc)
+                                    asgn = ir.Assign(new, x.target, expr.loc)
+                                    append_and_fix(asgn)
+                                    acc = asgn.target
+                            elif expr.op == 'build_list':
```
sklam, line 231: Same as above

numba/core/interpreter.py
```diff
@@ -1283,6 +1554,28 @@ def op_JUMP_IF_FALSE_OR_POP(self, inst, pred):
...
+            self.get(exc_match_name), args=(lhs, rhs), kws=(), loc=self.loc,
+        )
+        predicate = self.store(exc, pred)
+        bra = ir.Branch(cond=predicate, truebr=truebr, falsebr=falsebr,
+                        loc=self.loc)
+        self.current_block.append(bra)
+
+    def op_RERAISE(self, inst, exc):
+        # Do nothing, Numba can't handle this case and it's caught else where
+        pass
```
sklam, line 1577: Better to put in a `StaticRaise` here.
```suggestion
        self.current_block.append(ir.StaticRaise(None, None, loc=self.loc))
```

sklam: It's really to guard against getting a RERAISE unexpectedly. The `StaticRaise` will terminate the execution, even thought it is not raising the correct exception.

numba/core/interpreter.py
stuartarchibald: :) this is because I originally wrote the algorithm to work against the incoming stack, so everything is/was backwards from the last emitted statement which would have been a `LIST_TO_TUPLE`. Agree with changes to positive indices, will check and, assuming fine, apply the suggested.

numba/core/interpreter.py
stuartarchibald: Should we perhaps also note that it's not doing the right thing as part of the error message in the `StaticRaise` ?

numba/core/interpreter.py
stuartarchibald: This also made me think about a case like:
```python
@njit
def foo(a):
    z = [1, 2]
    return (*a, z.append(3), z.extend(a), np.ones(3)), z

x = (4, 5.6)
foo(x)
```
In this case there's other calls and append/extend in the unpack that do not relate to the unpacked tuple target . I'm not sure how common this sort of thing is but it's legal so I've patched up to make it work.

stuartarchibald:
More issues... Consider the contrived:
```python
@njit
def foo(a):
    return (*a, *(*a, a), *(a, (*(a, (1, 2), *(3,), *a), (a, 1, (2, 3), *a, 1), (1,))))
x = (4, 5)
foo(x)
```
Numba currently produces:
```
(4, 5, 4, 5, (4, 5), (4, 5), (3, 4, 5, ((4, 5), 1, (2, 3), 4, 5, 1), (1,)))
```
which is wrong, will investigate. Fairly sure this wasn't a problem when the code was operating directly on the stack.

numba/core/interpreter.py
stuartarchibald: https://github.com/sklam/numba/commit/492222d0e324e8bacd0c3fa1e08263b38db2eaae is applied to this branch.

stuartarchibald:
> More issues... Consider the contrived:
> 
> ```python
> @njit
> def foo(a):
>     return (*a, *(*a, a), *(a, (*(a, (1, 2), *(3,), *a), (a, 1, (2, 3), *a, 1), (1,))))
> x = (4, 5)
> foo(x)
> ```
> 
> Numba currently produces:
> 
> ```
> (4, 5, 4, 5, (4, 5), (4, 5), (3, 4, 5, ((4, 5), 1, (2, 3), 4, 5, 1), (1,)))
> ```
> 
> which is wrong, will investigate. Fairly sure this wasn't a problem when the code was operating directly on the stack.

This is caused by the rewrite trying to push const variables ahead of a build list so as to preserve the initial value information.

stuartarchibald:
https://github.com/numba/numba/pull/6579#issuecomment-756657459 is fixed by 2f03276

numba/core/interpreter.py
stuartarchibald: Any idea if this is actually reachable during execution? This sort of thing:
```
@njit
def foo():
    try:
        raise Exception
    except Exception:
        raise ValueError("ERROR")
```
will create a `RERAISE` op code that actually gets interpreted, but it's not reachable at runtime. Most things I've tried follow the pattern where the branch with the `RERAISE` on it can only be reached via a predicate from matching the pending exception against the exception class listed in the `except` clause, and as Numba only supports bare `except` or `except Exception` it renders the `RERAISE` unreachable.

sklam: It is not reachable at runtime. My intention is to guard against unexpected execution of the RERAISE. Also, we should terminate the BB as byteflow is saying this opcode is a terminator. So, I think raising an Assertion error in `StaticRaise` with a message about the impossible situation makes sense.

sklam:
https://github.com/numba/numba/pull/6579/commits/2f03276946b5e8e68b1249a0bc480ddfaf7cd904 looks good

numba/core/interpreter.py
stuartarchibald: Done in e94d583, I've also added the standard error reporting text into the `AssertionError` message string. Tested by checking for the python bytecode being present and then checking that the Numba IR has the `StaticRaise` with the right message if so.

sklam (approved):
Thanks for working on the py3.9 support. This has turned out to be more involved then anticipated with some CPython bytecode changes that were not well documented.

stuartarchibald:
@sklam Great, thanks for reviewing and lending a hand with the CFG altering op codes (I checked that part, looks good).

seibert:
Passed CPU smoketest (build 90) on all platforms except for Python 3.9, linux-ppc64le, NumPy 1.19:
```
2021-01-08 19:59:30,399] {docker_operator.py:262} INFO - ======================================================================
[2021-01-08 19:59:30,399] {docker_operator.py:262} INFO - FAIL: test_threading_concurrent_mix_use_masks_omp (numba.tests.test_parallel_backend.TestSpecificBackend)
[2021-01-08 19:59:30,400] {docker_operator.py:262} INFO - ----------------------------------------------------------------------
[2021-01-08 19:59:30,400] {docker_operator.py:262} INFO - Traceback (most recent call last):
[2021-01-08 19:59:30,400] {docker_operator.py:262} INFO - File "/opt/conda/envs/testenv_ff41bfba-b5fb-4503-9765-f740965aeba4/lib/python3.9/site-packages/numba/tests/test_parallel_backend.py", line 394, in test_template
[2021-01-08 19:59:30,400] {docker_operator.py:262} INFO - o, e = self.run_test_in_separate_process(injected_method, backend)
[2021-01-08 19:59:30,400] {docker_operator.py:262} INFO - File "/opt/conda/envs/testenv_ff41bfba-b5fb-4503-9765-f740965aeba4/lib/python3.9/site-packages/numba/tests/test_parallel_backend.py", line 372, in run_test_in_separate_process
[2021-01-08 19:59:30,400] {docker_operator.py:262} INFO - return self.run_cmd(cmdline, env_copy)
[2021-01-08 19:59:30,400] {docker_operator.py:262} INFO - File "/opt/conda/envs/testenv_ff41bfba-b5fb-4503-9765-f740965aeba4/lib/python3.9/site-packages/numba/tests/test_parallel_backend.py", line 360, in run_cmd
[2021-01-08 19:59:30,400] {docker_operator.py:262} INFO - raise AssertionError(
[2021-01-08 19:59:30,401] {docker_operator.py:262} INFO - AssertionError: process failed with code 1: stderr follows
[2021-01-08 19:59:30,401] {docker_operator.py:262} INFO - E
[2021-01-08 19:59:30,401] {docker_operator.py:262} INFO - ======================================================================
[2021-01-08 19:59:30,401] {docker_operator.py:262} INFO - ERROR: test_threading_concurrent_mix_use_masks (numba.tests.test_parallel_backend.TestParallelBackend)
[2021-01-08 19:59:30,401] {docker_operator.py:262} INFO - ----------------------------------------------------------------------
[2021-01-08 19:59:30,401] {docker_operator.py:262} INFO - Traceback (most recent call last):
[2021-01-08 19:59:30,401] {docker_operator.py:262} INFO - File "/opt/conda/envs/testenv_ff41bfba-b5fb-4503-9765-f740965aeba4/lib/python3.9/site-packages/numba/tests/test_parallel_backend.py", line 334, in test_method
[2021-01-08 19:59:30,402] {docker_operator.py:262} INFO - self.run_compile(impl, parallelism=p)
[2021-01-08 19:59:30,402] {docker_operator.py:262} INFO - File "/opt/conda/envs/testenv_ff41bfba-b5fb-4503-9765-f740965aeba4/lib/python3.9/site-packages/numba/tests/test_parallel_backend.py", line 281, in run_compile
[2021-01-08 19:59:30,402] {docker_operator.py:262} INFO - thread_impl(fnlist)
[2021-01-08 19:59:30,402] {docker_operator.py:262} INFO - File "/opt/conda/envs/testenv_ff41bfba-b5fb-4503-9765-f740965aeba4/lib/python3.9/site-packages/numba/tests/test_parallel_backend.py", line 178, in run_compile
[2021-01-08 19:59:30,402] {docker_operator.py:262} INFO - raise RuntimeError(_msg % '\n'.join([repr(x) for x in errors]))
[2021-01-08 19:59:30,402] {docker_operator.py:262} INFO - RuntimeError: Error(s) occurred in delegated runner:
[2021-01-08 19:59:30,402] {docker_operator.py:262} INFO - ImportError("cannot import name 'RLock' from partially initialized module 'multiprocessing.synchronize' (most likely due to a circular import) (/opt/conda/envs/testenv_ff41bfba-b5fb-4503-9765-f740965aeba4/lib/python3.9/multiprocessing/synchronize.py)")
[2021-01-08 19:59:30,402] {docker_operator.py:262} INFO - ImportError("cannot import name 'RLock' from partially initialized module 'multiprocessing.synchronize' (most likely due to a circular import) (/opt/conda/envs/testenv_ff41bfba-b5fb-4503-9765-f740965aeba4/lib/python3.9/multiprocessing/synchronize.py)")
[2021-01-08 19:59:30,403] {docker_operator.py:262} INFO - 
[2021-01-08 19:59:30,403] {docker_operator.py:262} INFO - ----------------------------------------------------------------------
```
Not sure why the RLock import error is showing up just for this platform.

esc:
Full DAG with available NumPy versions (1.16 and 1.19) triggered as `numba_243`.

stuartarchibald:
RLock problem reproduces locally:
```
$ ./runtests.py numba.tests.test_parallel_backend.TestSpecificBackend.test_threading_concurrent_mix_use_masks_omp
<path>/numba/np/ufunc/parallel.py:363:
F
FAIL: test_threading_concurrent_mix_use_masks_omp (numba.tests.test_parallel_backend.TestSpecificBackend)
  File "<path>/numba/tests/test_parallel_backend.py", line 394, in test_template
    o, e = self.run_test_in_separate_process(injected_method, backend)
  File "<path>/numba/tests/test_parallel_backend.py", line 372, in run_test_in_separate_process
    return self.run_cmd(cmdline, env_copy)
  File "<path>/numba/tests/test_parallel_backend.py", line 360, in run_cmd
    raise AssertionError(
AssertionError: process failed with code 1: stderr follows
E
ERROR: test_threading_concurrent_mix_use_masks (numba.tests.test_parallel_backend.TestParallelBackend)
  File "<path>/numba/tests/test_parallel_backend.py", line 334, in test_method
    self.run_compile(impl, parallelism=p)
  File "<path>/numba/tests/test_parallel_backend.py", line 281, in run_compile
    thread_impl(fnlist)
  File "<path>/numba/tests/test_parallel_backend.py", line 178, in run_compile
    raise RuntimeError(_msg % '\n'.join([repr(x) for x in errors]))
RuntimeError: Error(s) occurred in delegated runner:
ImportError("cannot import name 'RLock' from partially initialized module 'multiprocessing.synchronize' (most likely due to a circular import) (<pypath>/lib/python3.9/multiprocessing/synchronize.py)")
ImportError("cannot import name 'RLock' from partially initialized module 'multiprocessing.synchronize' (most likely due to a circular import) (<pypath>/lib/python3.9/multiprocessing/synchronize.py)")
```

stuartarchibald:
Buildfarm ID: `numba_244` passed on a7c18d9

sklam (approved):
Latest change looks ok

stuartarchibald:
Thanks for reviewing @sklam

PR #6580, opened 2020-12-15
Store partial typing errors in compiler state

ehsantn:
As title. Helps with usability of partial type inference significantly. The test infrastructure for this is useful by itself as well.

### DIFF
--- a/numba/core/typed_passes.py
+++ b/numba/core/typed_passes.py
@@ -1,5 +1,5 @@
 from contextlib import contextmanager
-from collections import defaultdict
+from collections import defaultdict, namedtuple
 from copy import copy
 import warnings
 
@@ -21,6 +21,15 @@ from numba.core.ir_utils import (raise_on_unsupported_feature, warn_deprecated,
                                  is_operator_or_getitem)
 from numba.core import postproc
 from llvmlite import binding as llvm
+
+
+# Outputs of type inference pass
+_TypingResults = namedtuple("_TypingResults", [
+    "typemap",
+    "return_type",
+    "calltypes",
+    "typing_errors",
+])
 
 
 @contextmanager
@@ -76,7 +85,7 @@ def type_inference_stage(typingctx, interp, args, return_type, locals={},
     # Output all Numba warnings
     warnings.flush()
 
-    return typemap, restype, calltypes, errs
+    return _TypingResults(typemap, restype, calltypes, errs)
 
 
 class BaseTypeInference(FunctionPass):
--- a/numba/core/typed_passes.py
+++ b/numba/core/typed_passes.py
@@ -69,13 +69,14 @@ def type_inference_stage(typingctx, interp, args, return_type, locals={},
             infer.seed_type(k, v)
 
         infer.build_constraint()
-        infer.propagate(raise_errors=raise_errors)
+        # return errors in case of partial typing
+        errs = infer.propagate(raise_errors=raise_errors)
         typemap, restype, calltypes = infer.unify(raise_errors=raise_errors)
 
     # Output all Numba warnings
     warnings.flush()
 
-    return typemap, restype, calltypes
+    return typemap, restype, calltypes, errs
 
 
 class BaseTypeInference(FunctionPass):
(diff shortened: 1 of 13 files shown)


sklam:
What is the intended use for the new `.typing_errors`? It is for developers to debug or is it something to provide feedback to the users?

ehsantn:
It is for knowing whether partial typing was successful or not. Currently, the only way to know is to check for "unknown" types in `typemap`, but there could be other errors. For example, type unification errors are lost.

ehsantn:
@sklam any potential concerns with this PR? Would appreciate some feedback. Thanks.

numba/core/typed_passes.py
```diff
@@ -91,14 +92,16 @@ def run_pass(self, state):
...
+            typemap, return_type, calltypes, errs = type_inference_stage(
                 state.typingctx,
                 state.func_ir,
                 state.args,
                 state.return_type,
                 state.locals,
                 raise_errors=self._raise_errors)
             state.typemap = typemap
+            # save errors in case of partial typing
+            state.typing_errors = errs
```
sklam, line 104: Instead of saving only the last typing error, should it save all errors in a dictionary that is keyed by the `pass_id`?

ehsantn: We access typing errors within the same pass (after calling `run_pass` of parent), so having `pass_id` is not needed. I think the added complexity is not worth it, especially because `typemap` and `calltypes` are unique as well.

numba/tests/test_typeinfer.py
```diff
@@ -759,5 +763,75 @@ def test_fold_arguments_list_inputs(self):
...
+        def impl(flag):
+            if flag:
+                a = 1
+            else:
+                a = np.ones(3)
+            return a
+
+        typing_errs = get_func_typing_errs(impl, (types.bool_,))
+        self.assert_(isinstance(typing_errs, list) and len(typing_errs) == 1)
+        self.assert_(isinstance(typing_errs[0], errors.TypingError)
```
sklam, line 832: `assert_` is a deprecated alias. Use `assertTrue` instead.

numba/core/typed_passes.py
```diff
@@ -69,13 +69,14 @@ def type_inference_stage(typingctx, interp, args, return_type, locals={},
...
-        infer.propagate(raise_errors=raise_errors)
+        # return errors in case of partial typing
+        errs = infer.propagate(raise_errors=raise_errors)
         typemap, restype, calltypes = infer.unify(raise_errors=raise_errors)
 
     # Output all Numba warnings
     warnings.flush()
 
-    return typemap, restype, calltypes
+    return typemap, restype, calltypes, errs
```
sklam, line 79: With 4 elements returning, i think it'd be clearer (esp for the caller) if this return a namedtuple instead to avoid a unnamed unpacking and the `_, _, _`.

numba/core/typed_passes.py
sklam: ok, i agree the extra complexity is not worth it for now.

sklam:
@ehsantn, thanks for the patch. There are just two minor comments. One regarding use of deprecated API and the other will make the code clearer.

numba/tests/test_typeinfer.py
ehsantn: Done.

numba/core/typed_passes.py
ehsantn: I agree. Done.

ehsantn:
Thank you @sklam. I applied the changes. Let me know if anything else is needed.

sklam (approved):
Thanks for the patch. It looks good.

numba/core/typed_passes.py
stuartarchibald: Should `type_inference_stage` also return the same? It's use is more prevalent and also leads to the same `x, _, _, _` pattern.

ehsantn: @stuartarchibald `type_inference_stage` is returning a namedtuple in this PR now. What do you mean?

PR #6582, opened 2020-12-16
Fix typed dict error with uninitialized padding bytes 

### LINKED ISSUE #6570: Dicts with tuple keys not working in 0.52.0
I’m using a dict with a tuple as the key which worked fine in 0.51.2, but does not work in 0.52.0:

```
import numpy as np
import numba

@numba.njit
def testdict(tuptype):
    tdict = numba.typed.Dict.empty(tuptype, numba.core.types.float32)
    t1 = np.array([3], dtype=np.uint64)
    t2 = np.array([5.67], dtype=np.float32)
    v1 = np.array([10.23], dtype=np.float32)
    tdict[(t1[0], t2[0])] = v1[0]
    # 0.51.2 prints True as expected; 0.52.0 prints False
    print((t1[0], t2[0]) in tdict)

testdict(
  numba.core.types.Tuple((numba.core.types.uint64, numba.core.types.float32)))
```
When I try to check membership of an element that I just stored, 0.52.0 returns False.

sklam:
Fix #6570 

The issue was that some data type requires padding bytes to be properly addressable. The `alloca` in Numba does not fill the initialize the padding bytes and the typed dict implementation key comparison is a simple byte comparison. Since the padding bytes could have garbage data, the key comparison was failing unexpectedly and non-deterministically depending on the values that happen to be in the paddings.

### DIFF
--- a/numba/tests/test_dictobject.py
+++ b/numba/tests/test_dictobject.py
@@ -931,7 +931,7 @@ class TestDictObject(MemoryLeakMixin, TestCase):
         self.assertEqual(foo(), foo.py_func())
 
     def test_issue6570_alignment_padding(self):
-        # Create a key type that is 12-byte long on a 8-byte aligned system
+        # Create a key type that is 12-bytes long on a 8-byte aligned system
         # so that the a 4-byte padding is needed.
         # If the 4-byte padding is not zero-filled, it will have garbage data
         # that affects key matching in the lookup.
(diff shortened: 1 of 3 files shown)


stuartarchibald:
Something in here, guessing the `memset`, has broken various optimisations.

sklam:
CI failing because of https://github.com/numba/numba/issues/6590

numba/tests/test_dictobject.py
```diff
@@ -930,6 +930,24 @@ def foo():
 
         self.assertEqual(foo(), foo.py_func())
 
+    def test_issue6570_alignment_padding(self):
+        # Create a key type that is 12-byte long on a 8-byte aligned system
```
stuartarchibald, line 934:
```suggestion
        # Create a key type that is 12-bytes long on a 8-byte aligned system
```

stuartarchibald (approved):
Thanks for the patch and debugging this, it was evidently challenging! Patch looks good, I fixed a 1-char typo in abf5d78. Approved for merge.

PR #6584, opened 2020-12-17
Remove jitclass from __init__ __all__.

### LINKED ISSUE #6583: from numba import * fails due to jitclass still being in __all__
5f7ecf3bbfdf25b55f663f5278e4e31ad6fdde4c removes `jitclass` from the `numba` namespace, but `__init__.py`'s `__all__` still contains `jitclass`, causing `from numba import *` to fail.

stuartarchibald:
As title.

Fixes #6583

### DIFF
--- a/numba/__init__.py
+++ b/numba/__init__.py
@@ -66,7 +66,6 @@ __all__ = """
     njit
     stencil
     jit_module
-    jitclass
     typeof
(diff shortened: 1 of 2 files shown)


esc (approved):
Tried the fix locally, works fine.

PR #6591, opened 2020-12-21
Fix inlineasm test failure.

sklam:
Fix 6590

Use 32-bit register and make sure "early clobber" flag is set (the "&")


stuartarchibald (approved):
Thanks for the fix. Think this was working by luck due to the lack of the early clobber.

PR #6598, opened 2020-12-22
Extending the extension API for hardware targets

stuartarchibald:
This PR is a draft, it contains some exploratory patches with ideas about:

1. How to make it so that it is easier to extend Numba to new hardware targets, in particular through a new kwarg `hardware=` on commonly used extension APIs e.g. `@overload`
2. Expressing a hardware hierarchy such that implementations of functions are only as specific as they need to be for a given piece of hardware. For example `hash()` is trivially implemented as `lambda x: x.__hash__()` no matter what the hardware target, therefore `hardware='generic'` would be an appropriate target. This is to help with code reuse.
3. Making it possible to "offload" selected functions from one target to another via compiler passes comprising part of a custom compilation pipeline.
4. Permitting context based management of the "current" target for a function that's being jitted.

Some of these concepts are demonstrated here for a new "dummy" target, the "DPU" (Dummy Processing Unit): https://gist.github.com/stuartarchibald/7051191e0f0cb09ba9c2d42c7c21ced6
this is probably the place to start looking at the impact of this change.

This branch breaks a load of things, contains numerous (and some quite horrible) hacks to get things to "work", and has a lot of "undesigned" parts. However, it's a starting point for making this possible and lays down some of the ground work for more abstract constraints in overloads such as declarative typing and selecting overload implementations based on metadata.

Do not merge!

### DIFF
--- a/numba/core/base.py
+++ b/numba/core/base.py
@@ -230,11 +230,14 @@ class BaseContext(object):
     # the function descriptor
     fndesc = None
 
-    def __init__(self, typing_context):
+    def __init__(self, typing_context, target):
         _load_global_helpers()
 
         self.address_size = utils.MACHINE_BITS
         self.typing_context = typing_context
+        from numba.core.extending_hardware import hardware_registry
+        self.target_name = target
+        self.target = hardware_registry[target]
 
         # A mapping of installed registries to their loaders
         self._registries = {}
--- a/numba/core/compiler.py
+++ b/numba/core/compiler.py
@@ -435,7 +435,7 @@ class Compiler(CompilerBase):
             pms.append(
                 DefaultPassBuilder.define_objectmode_pipeline(self.state)
             )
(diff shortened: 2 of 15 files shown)


sklam:
I'm adding a new example that is executing in a remote process using 0MQ: https://gist.github.com/sklam/b4187b0ee659d3e824cd3fe8a15b7b3d

stuartarchibald:
> I'm adding a new example that is executing in a remote process using 0MQ: https://gist.github.com/sklam/b4187b0ee659d3e824cd3fe8a15b7b3d

Thanks @sklam I've tested this locally and it works, good to see the abstraction level is about right.

stuartarchibald:
Closing this as I've ported the patch to post 0.53 mainline and got it mostly working in https://github.com/numba/numba/pull/6898.

PR #6602, opened 2020-12-24
Never add multiply defined scalars to shape data structures.

### LINKED ISSUE #6597: Parallel loop prange gives wrong outcomes
Hi, I've encountered wrong outcomes when prange is used. The issue can be reproduced with the code below.

```
import numba
import numpy as np

def foo(a, b):
    n_rows = b.shape[0]
    for x in numba.prange(1):
        i_row = 0
        for _ in range(n_rows):
            bi = b[i_row]
            print(i_row, a[i_row, :bi])
            i_row += 1
            if i_row >= n_rows:
                i_row -= n_rows


a = np.arange(10).reshape((10, 1))
b = np.ones(10)

# This prints the expected result, i.e. prints 0,[0] \n 1,[1] \n 2,[2], etc
numba.njit(foo, parallel=False)(a,b)

# This produces a wrong result, i.e. prints 0,[0] \n 0,[0] \n 0,[0], etc
numba.njit(foo, parallel=True)(a,b)

```

This started to happen with numba 0.47, until 0.46 was running as expected (though with numba 0.46 and 0.47 I had to use llvmlite 0.32 since more recent versions give an error at compile time).

DrTodd13:
Resolves #6597. 

Previously, we would add scalars and detect multiple definitions along the way and then we'd remove them at the end.  The problem with that approach is that intermediate steps consult the data structures and may think that a variable is a constant at some point (and then use the constant in some build_tuple expr) when in reality the variable is in a loop and may be updated by the later part of the loop in a previous iteration.  Now, we exclude non-array variables with multiple definitions from ever going in the data structures.

### DIFF
--- a/numba/parfors/array_analysis.py
+++ b/numba/parfors/array_analysis.py
@@ -1096,6 +1096,11 @@ class ArrayAnalysis(object):
             blocks = self.func_ir.blocks
 
         self.func_ir._definitions = build_definitions(self.func_ir.blocks)
+        # Calculate the non-arrays defined more than once and exclude them
+        # from ever entering the shape data structures.
+        self.multi_def = [k for k,v in self.func_ir._definitions.items()
+            if not isinstance(self.typemap[k], types.ArrayCompatible) and
+            len(v) > 1]
 
         if equiv_set is None:
             init_equiv_set = SymbolicEquivSet(self.typemap)
@@ -1334,10 +1339,12 @@ class ArrayAnalysis(object):
                 no invalidation is needed and we don't call define().
             """
             needs_define = True
-            if shape is not None:
-                needs_define = equiv_set.insert_equiv(lhs, shape)
-            if needs_define:
-                equiv_set.define(lhs, redefined, self.func_ir, typ)
+            # Don't add this variable if it is multiply defined.
+            if lhs.name not in self.multi_def:
+                if shape is not None:
+                    needs_define = equiv_set.insert_equiv(lhs, shape)
+                if needs_define:
+                    equiv_set.define(lhs, redefined, self.func_ir, typ)
         elif isinstance(inst, ir.StaticSetItem) or isinstance(
             inst, ir.SetItem
         ):
@@ -2869,8 +2876,9 @@ class ArrayAnalysis(object):
             sizes = []
             size_names = []
             for name, shape in zip(names, shapes):
-                if i < len(shape):
-                    size = shape[len(shape) - 1 - i]
+                slen = len(shape)
+                if i < slen:
+                    size = shape[slen - 1 - i]
                     const_size = equiv_set.get_equiv_const(size)
                     if const_size == 1:
                         const_size_one = size
--- a/numba/tests/test_parfors.py
+++ b/numba/tests/test_parfors.py
@@ -2694,7 +2694,8 @@ class TestPrange(TestPrangeBase):
             for x in range(1):
                 i_row = 0
                 for _ in range(n_rows):
-                    c[0] = a[i_row, 0]
+                    bi = b[i_row]
+                    c[:] = a[i_row, :bi]
                     i_row += 1
                     if i_row >= n_rows:
                         i_row -= n_rows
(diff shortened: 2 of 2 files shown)


sklam:
From out-of-band discussion:

* Need to check if multiply defined variable of array type will be problematic; e.g. an array variable changes its shape in a loop.

numba/tests/test_parfors.py
```diff
@@ -2686,6 +2707,26 @@ def test_impl():
...
+                    bi = b[i_row]
+                    c[:] = a[i_row, :bi]
+                    i_row += 1
+                    if i_row >= n_rows:
+                        i_row -= n_rows
+            return c
+
+        a = np.arange(10).reshape((10, 1))
+        b = np.ones(10, dtype=np.int)
+        c = np.ones(1, dtype=np.int)
```
stuartarchibald, line 2727: Suspect this will fail on windows as `np.int` is aliased to pythons `int`, think this will end up as a 32bit int array which is probably the root cause of the typing failures on windows CI. Suggest using `dtype=np.int64`.

numba/tests/test_parfors.py
```diff
@@ -1733,6 +1733,27 @@ def test_impl(a):
...
+                d = d * 2
+                if j < 1:
+                    if e[0] > 0:
+                        e = b
+                    else:
+                        e = a
+            return e
+
+        a = np.ones(7, dtype=np.int)
+        b = np.ones(10, dtype=np.int)
```
stuartarchibald, line 1754:
```suggestion
        a = np.ones(7, dtype=np.int64)
        b = np.ones(10, dtype=np.int64)
```
This probably needs the same `int`->`int64` change

DrTodd13: Committed.

stuartarchibald:
Thanks for the fixes @DrTodd13. I think this needs conflicts resolving, and then CI can run against it again. I think it might be down to just one failing test... this one:
https://dev.azure.com/numba/numba/_build/results?buildId=7646&view=logs&j=4f8e7448-6bbb-53dd-1446-49d51b72f33b&t=22d62e29-1b6b-5406-981d-1aa65fc1883c&l=9706

DrTodd13:
> Thanks for the fixes @DrTodd13. I think this needs conflicts resolving, and then CI can run against it again. I think it might be down to just one failing test... this one:
> https://dev.azure.com/numba/numba/_build/results?buildId=7646&view=logs&j=4f8e7448-6bbb-53dd-1446-49d51b72f33b&t=22d62e29-1b6b-5406-981d-1aa65fc1883c&l=9706

Locally, I have one test failing.  I had "fixed" this issue before but that caused a problem in another test so I need to solve this issue in a different way.

ehsantn (commented):
Thanks for the improvements in several places. I'm just wondering how some of these multiple definitions cases may occur and if it's possible to simplify the code. See comment.

numba/parfors/array_analysis.py
```diff
@@ -1334,10 +1371,34 @@ def _analyze_inst(self, label, scope, equiv_set, inst, redefined):
...
+                        if equiv_set.is_equiv(lhs, shape):
+                            needs_define = equiv_set.insert_equiv(lhs, shape)
+                        else:
+                            if config.DEBUG_ARRAY_OPT >= 2:
+                                print("Adding", lhs, "to multi_def")
+                            # This is a multiply defined var with
+                            # non-equivalent definitions so add to
+                            # multiply defined set.
+                            self.multi_def.append(lhs.name)
+                            self.remove_redefineds([lhs.name])
```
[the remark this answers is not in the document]
DrTodd13: I'm not sure the best implementation here.  Right now, I treat arrays different from scalars.  If I have multiple definitions of a scalar then I immediately exclude them.  If I have multiple definitions of some array variable then I look to see if the shapes are equivalent at each definition point so I only later may exclude them.  I could treat the scalars the same way and the code would be more consistent but the earlier we can exclude the less computation there is.

The test case test_multi_defined_array that I added for this PR illustrates this issue.

```
        def test_impl(a, b):
            d = a * 2
            e = a
            for j in range(2):
                e = e * 3
                d = d * 2
                if j < 1:
                    if e[0] > 0:
                        e = b
                    else:
                        e = a
            return e
```
The "e = e * 3" parfor should not fuse with the "d = d * 2" parfor but if you are just looking sequentially then the first time you see these you will believe they are the same sizes.

Issue #6183, opened 2020-08-28
CUDA fastmath: use fast math trig / exp / log / fdivide functions


gmarkall:
(cc @mnicely who I think will have an interest in the resolution of this)

Compiling the following with nvcc and fast math flags:

```cuda
#include <math.h>

__global__ void f(float* r, float x)
{
  r[0] = cos(x);
}
```

(using `nvcc --std=c++11 --generate-code arch=compute_75,code=sm_75 --use_fast_math test.cu --ptx -o test_fast.ptx`)

results in the following PTX:

```assembly
.visible .entry _Z1fPff(
	.param .u64 _Z1fPff_param_0,
	.param .f32 _Z1fPff_param_1
)
{
	.reg .f32 	%f<3>;
	.reg .b64 	%rd<3>;


	ld.param.u64 	%rd1, [_Z1fPff_param_0];
	ld.param.f32 	%f1, [_Z1fPff_param_1];
	cvta.to.global.u64 	%rd2, %rd1;
	cos.approx.ftz.f32 	%f2, %f1;
	st.global.f32 	[%rd2], %f2;
	ret;
}
```

However, the following CUDA Python:

```python
from numba import config, cuda, float32, void
from math import cos

config.DUMP_ASSEMBLY = True


@cuda.jit(void(float32[::1], float32), fastmath=True)
def f(r, x):
    r[0] = cos(x)
```

produces a lot of code:

```assembly
.visible .global .align 4 .u32 _ZN6cudapy8__main__5f$241E5ArrayIfLi1E1C7mutable7alignedEf__errcode__;
.visible .global .align 4 .u32 _ZN6cudapy8__main__5f$241E5ArrayIfLi1E1C7mutable7alignedEf__tidx__;
.visible .global .align 4 .u32 _ZN6cudapy8__main__5f$241E5ArrayIfLi1E1C7mutable7alignedEf__ctaidx__;
.visible .global .align 4 .u32 _ZN6cudapy8__main__5f$241E5ArrayIfLi1E1C7mutable7alignedEf__tidy__;
.visible .global .align 4 .u32 _ZN6cudapy8__main__5f$241E5ArrayIfLi1E1C7mutable7alignedEf__ctaidy__;
.visible .global .align 4 .u32 _ZN6cudapy8__main__5f$241E5ArrayIfLi1E1C7mutable7alignedEf__tidz__;
[...]

PR #6619, opened 2021-01-05
Improved fastmath code generation for trig, log, and exp/pow.

testhound:
This pull request adds fastmath code generation support for the following functions:

sin, cos, tan, log, log2, log10,  exp, and pow when the fastmath option is used with 32-bit types.

This pull request address poor code generation identified in: https://github.com/numba/numba/issues/6183


esc:
@testhound thank you for submitting this! I have added it to the queue for review.

numba/cuda/mathimpl.py
```diff
@@ -76,8 +88,20 @@ def lower_boolean_impl(context, builder, sig, args):
 
 def impl_unary(key, ty, libfunc):
     def lower_unary_impl(context, builder, sig, args):
-        libfunc_impl = context.get_function(libfunc, typing.signature(ty, ty))
-        return libfunc_impl(builder, args)
+        if ty == float32 and context.fastmath is True:
```
stuartarchibald, line 91:
```suggestion
        if ty == float32 and context.fastmath:
```

numba/cuda/mathimpl.py
```diff
@@ -76,8 +88,20 @@ def lower_boolean_impl(context, builder, sig, args):
...
+                libfunc_impl = context.get_function(libfunc,
+                                                    typing.signature(ty, ty))
+            else:
+                new_libfunc = getattr(libdevice, fast_replacement)
+                libfunc_impl = context.get_function(new_libfunc,
+                                                    typing.signature(ty, ty))
+            return libfunc_impl(builder, args)
+        else:
+            libfunc_impl = context.get_function(libfunc,
+                                                typing.signature(ty, ty))
```
stuartarchibald, line 103: Think the 3x `libfunc_impl = context.get_function...` can be pulled out as all paths use it, then the branch is just on 32bit float with fastmath and switching the library `libfunc = getattr(libdevice, fast_replacement)`, something like the (entirely untested!) code below perhaps, what do you think?
```python
if ty == float32 and context.fastmath:
    fast_replacement = unarys_fastmath.get(libfunc.__name__)
    if fast_replacement is not None:
        libfunc = getattr(libdevice, fast_replacement)

libfunc_impl = context.get_function(libfunc,
                                    typing.signature(ty, ty))
```

numba/cuda/mathimpl.py
```diff
@@ -102,9 +126,22 @@ def lower_unary_int_impl(context, builder, sig, args):
...
+            fast_replacement = binarys_fastmath.get(libfunc.__name__)
+            if fast_replacement is None:
+                libfunc_impl = context.get_function(libfunc,
+                                                    typing.signature(ty,
+                                                                     ty, ty))
+            else:
+                new_libfunc = getattr(libdevice, fast_replacement)
+                libfunc_impl = context.get_function(new_libfunc,
+                                                    typing.signature(ty,
+                                                                     ty, ty))
```
stuartarchibald, line 139: Same comment/refactor suggestion as above.

stuartarchibald (commented):
Thanks for the patch, great to see these fast math functions added. I've made a couple of suggestions, in general looks good. I think this probably needs documenting, both in the CUDA docs and also the `cuda.jit` doc strings. Thanks again!

numba/cuda/mathimpl.py
testhound: @stuartarchibald thanks for the review, I have made the code changes locally. Can you point me to the documentation files that need to be changed?

stuartarchibald:
RE https://github.com/numba/numba/pull/6619#discussion_r555549102: I've taken a look through the current CUDA docs and I can't seem to find a section talking about the various options available to the `cuda.jit` decorator and the impact they have (might be a nice in future to have an FAQ entry https://numba.readthedocs.io/en/latest/cuda/faq.html or something to add as `kernel declaration options` here https://numba.readthedocs.io/en/latest/cuda/kernels.html#kernel-declaration). For now, I think it's just the `fastmath` part of this section needs an update: https://numba.readthedocs.io/en/latest/cuda-reference/kernel.html#kernel-declaration which I think comes from here:
https://github.com/numba/numba/blob/8a4099a70c69efc82dc3154bfa48073c46870cb2/numba/cuda/decorators.py#L45-L48

Having reviewed the Numba documentation, I've also looked at the `nvcc` docs for ["fast math"](https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#options-for-steering-gpu-code-generation-use_fast_math) and think that what this PR is providing is a deviation from this behaviour. As a result, one concern I now have with this PR is over whether it would cause regressions in existing code bases due to a behavioural change in an existing configuration i.e. `fastmath=True`. It makes me wonder if low precision/fast math intrinsic functions ought to be activated via a different option, or that `fastmath` should be a dictionary like:
```
{
'ftz': <bool>,
'prec-div': <bool>,
'prec-sqrt': <bool>,
'fmad': <bool>,
'fast_libm': <bool>, # this being the thing that switches on the fast math intrinsic functions etc as wired in in this PR
}
```
On the CPU target, `fastmath` is a similar to this, it can be `True/False` but can also be specifically configured.
https://github.com/numba/numba/blob/1976c664afeb3251bc1246957c4735b1694f84f4/numba/core/cpu_options.py#L6-L40

What do you think?

testhound:
> RE [#6619 (comment)](https://github.com/numba/numba/pull/6619#discussion_r555549102): I've taken a look through the current CUDA docs and I can't seem to find a section talking about the various options available to the `cuda.jit` decorator and the impact they have (might be a nice in future to have an FAQ entry https://numba.readthedocs.io/en/latest/cuda/faq.html or something to add as `kernel declaration options` here https://numba.readthedocs.io/en/latest/cuda/kernels.html#kernel-declaration). For now, I think it's just the `fastmath` part of this section needs an update: https://numba.readthedocs.io/en/latest/cuda-reference/kernel.html#kernel-declaration which I think comes from here:
> 
> https://github.com/numba/numba/blob/8a4099a70c69efc82dc3154bfa48073c46870cb2/numba/cuda/decorators.py#L45-L48
> 
> Having reviewed the Numba documentation, I've also looked at the `nvcc` docs for ["fast math"](https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#options-for-steering-gpu-code-generation-use_fast_math) and think that what this PR is providing is a deviation from this behaviour. As a result, one concern I now have with this PR is over whether it would cause regressions in existing code bases due to a behavioural change in an existing configuration i.e. `fastmath=True`. It makes me wonder if low precision/fast math intrinsic functions ought to be activated via a different option, or that `fastmath` should be a dictionary like:
> 
> ```
> {
> 'ftz': <bool>,
> 'prec-div': <bool>,
> 'prec-sqrt': <bool>,
> 'fmad': <bool>,
> 'fast_libm': <bool>, # this being the thing that switches on the fast math intrinsic functions etc as wired in in this PR
> }
> ```
> 
> On the CPU target, `fastmath` is a similar to this, it can be `True/False` but can also be specifically configured.
> 
> https://github.com/numba/numba/blob/1976c664afeb3251bc1246957c4735b1694f84f4/numba/core/cpu_options.py#L6-L40
> 
> What do you think?

@stuartarchibald interesting suggestion. Let me digest and respond later.

gmarkall:
> Having reviewed the Numba documentation, I've also looked at the nvcc docs for "fast math" and think that what this PR is providing is a deviation from this behaviour.

As discussed out-of-band earlier, I think this is bringing Numba's behaviour into line with NVCC. An example of NVCC using a less precise cos implementation with the fast math flag can be seen in: https://github.com/gmarkall/nvcc-fastmath

> As a result, one concern I now have with this PR is over whether it would cause regressions in existing code bases due to a behavioural change in an existing configuration i.e. fastmath=True.

I think any code that was passing `fastmath=True` should not be relying on the usual precision guarantees - so if it does break some existing code, then that code was probably not safe to use with the `fastmath` flag.

> It makes me wonder if low precision/fast math intrinsic functions ought to be activated via a different option, or that fastmath should be a dictionary like:

In light of the above, would you agree that the fastmath dictionary suggestion wouldn't be required now?

numba/cuda/tests/cudapy/test_fastmath.py
```diff
@@ -17,6 +18,93 @@ def foo(arr, val):
...
+        self.assertNotIn('sin.approx.ftz.f32 ', slowver.ptx)
+
+    @skip_on_cudasim('fast exp not available in CUDASIM')
+    def test_expf(self):
+        def f4(r, x):
+            r[0] = exp(x)
+
+        fastver = cuda.jit("void(float32[::1], float32)", fastmath=True)(f4)
+        slowver = cuda.jit("void(float32[::1], float32)")(f4)
+        self.assertNotIn('fma.rn.f32 ', fastver.ptx)
```
gmarkall, line 60: Noting here that this checks that the fast version doesn't have any FMA instructions in it. I noticed that the fast version makes use of `ex2.approx.ftz.f32`, which could be checked for instead, but I also think the test as-is is sufficient, because there shouldn't be any FMA instructions in the fast version.

numba/cuda/tests/cudapy/test_fastmath.py
```diff
@@ -17,6 +18,93 @@ def foo(arr, val):
...
+        self.assertIn('fma.rn.f32 ', slowver.ptx)
+
+    @skip_on_cudasim('fast log not available in CUDASIM')
+    def test_logf(self):
+        def f5(r, x):
+            r[0] = log(x)
+
+        fastver = cuda.jit("void(float32[::1], float32)", fastmath=True)(f5)
+        slowver = cuda.jit("void(float32[::1], float32)")(f5)
+        self.assertIn('lg2.approx.ftz.f32 ', fastver.ptx)
```
gmarkall, line 70: (I guess this test is taking the approach that could have been used for `exp`)

gmarkall (changes requested):
This is looking great so far!

- The code changes look good to me.
- Tests run fine for me locally, after merging in master, with both NVVM70 (CUDA >= 11.2) and NVVM34 (CUDA < 11.2)
- For the documentation, I'd suggest two changes:


   1. Adding a section on "fast math" underneath the "Numba for CUDA GPUs" section: https://numba.readthedocs.io/en/stable/cuda/index.html - this section should refer to the CPU fastmath section (https://numba.readthedocs.io/en/stable/user/performance-tips.html#fastmath) and note the following differences for the CUDA target:
      - The flag can only be `True` or `False` for CUDA
      - The flag also enables some fast approximate trigonometric functions, and provide the list of them.
   2. Updating the `jit` decorator documentation to explain that `fastmath` also substitutes the fast versions of some functions: https://numba.readthedocs.io/en/stable/cuda-reference/kernel.html#numba.cuda.jit, and refer to the new section on "fast math" for CUDA GPUs.

numba/cuda/tests/cudapy/test_fastmath.py
testhound: @gmarkall I have made and verified the first of the two changes you requested. I am unclear how to make the change to numba/cuda/decorators.py that will update the documentation; specifically I updated 'decorators.py' but after generating the documentation with 'make html', I do not get a updated local page for 'CUDA Kernel API. Is there another command to update this portion of the documentation or did I update the wrong file?

gmarkall: I think sphinx isn't very good at detecting when docstrings change. Every time I change something documentation related, I always run `make clean html`, which is a bit inconvenient because it takes a while to build from scratch, but I haven't found a better way.

If you do this, does your change now show up?

gmarkall (changes requested):
Many thanks for the documentation additions - I have some further suggestions for the docs, which I've pushed in the commit https://github.com/gmarkall/numba/commit/f6e735f7196414d4ea5a2640cdab630e0aa32f0e. To summarise the suggestions, they are:

- Fixing the docs build by moving the `cuda-fast-math` label ahead of the section title,
- Add links to documentation for the underlying libdevice functions and NVVM optimizations,
- Explicitly state the `math` module functions affected by the transformation,
- Refer to the docs in the docstring of the `jit` decorator for `fastmath` - it's got to the point where it does a bit much to easily summarise in the docstring.

Feel free to pull in any / all of the changes in the linked commit, or do let me know what you think of the suggestions.

Following resolution of the documentation suggestions, I think this will be all looking good!

gmarkall (dismissed):
Many thanks for the update! This looks good to me!

@esc could this have a buildfarm run please? (the typeguard fail is a general issue not related to this branch)

esc:
Running on Farm as: `numba_smoketest_cuda_yaml_13`

esc:
Build farm was fine: `numba_smoketest_cuda_yaml_13`.

gmarkall:
@stuartarchibald @sklam Are you happy with this going RTM?

docs/source/cuda/fastmath.rst
```diff
@@ -0,0 +1,36 @@
...
+  library.
+
+  - :func:`math.cos`: Implemented using `__nv_fast_cosf <https://docs.nvidia.com/cuda/libdevice-users-guide/__nv_fast_cosf.html>`_.
+  - :func:`math.sin`: Implemented using `__nv_fast_sinf <https://docs.nvidia.com/cuda/libdevice-users-guide/__nv_fast_sinf.html>`_.
+  - :func:`math.tan`: Implemented using `__nv_fast_tanf <https://docs.nvidia.com/cuda/libdevice-users-guide/__nv_fast_tanf.html>`_.
+  - :func:`math.exp`: Implemented using `__nv_fast_expf <https://docs.nvidia.com/cuda/libdevice-users-guide/__nv_fast_expf.html>`_.
+  - :func:`math.log2`: Implemented using `__nv_fast_log2f <https://docs.nvidia.com/cuda/libdevice-users-guide/__nv_fast_log2f.html>`_.
+  - :func:`math.log10`: Implemented using `__nv_fast_log10f <https://docs.nvidia.com/cuda/libdevice-users-guide/__nv_fast_log10f.html>`_.
+  - :func:`math.log`: Implemented using `__nv_fast_logf <https://docs.nvidia.com/cuda/libdevice-users-guide/__nv_fast_logf.html>`_.
+  - :func:`math.pow`: Implemented using `__nv_fast_powf <https://docs.nvidia.com/cuda/libdevice-users-guide/__nv_fast_powf.html>`_.
```
stuartarchibald, line 36: most of the ```:func:``math.FOO`` ``` link to the `developer/autogen_math_listing` references, and a couple to the intended python doc. Not sure what can be done apart for explicit linking. See rendered doc: https://numba--6619.org.readthedocs.build/en/6619/cuda/fastmath.html

stuartarchibald:
> @stuartarchibald @sklam Are you happy with this going RTM?

Once https://github.com/numba/numba/pull/6619#discussion_r578554321 is resolved, patch looks good. Thanks for working on this @testhound, thanks for reviewing @gmarkall.

docs/source/cuda/fastmath.rst
stuartarchibald: This is a Numba doc problem, not for fixing in this PR. Deferring to ticket https://github.com/numba/numba/issues/6737.

stuartarchibald (approved):

stuartarchibald:
@testhound please could you resolve the conflicts when you have a moment? Many Thanks.

testhound:
> @testhound please could you resolve the conflicts when you have a moment? Many Thanks.

@stuartarchibald I just resolved the conflicts.

stuartarchibald (approved):
Thanks for resolving conflicts, looks good.

PR #6622, opened 2021-01-06
Fix 6534, handle unpack of assign-like tuples.

### LINKED ISSUE #6534: bytecode analysis failure with `BUILD_TUPLE_UNPACK`.

stuartarchibald:
As title, also removes some Python 2 and Python <3.6 restrictions.

Fixes #6534

### DIFF
--- a/numba/core/interpreter.py
+++ b/numba/core/interpreter.py
@@ -887,7 +887,8 @@ class Interpreter(object):
             gv_name = "unpack_single_tuple"
             gv_fn = ir.Global(gv_name, unpack_single_tuple, loc=self.loc,)
             self.store(value=gv_fn, name=gv_name, redefine=True)
-            exc = ir.Expr.call(self.get(gv_name), args=(first,), kws=(), loc=self.loc,)
+            exc = ir.Expr.call(self.get(gv_name), args=(first,), kws=(),
+                               loc=self.loc,)
             self.store(exc, temps[0])
         else:
             for other, tmp in zip(map(self.get, tuples[1:]), temps):
--- a/numba/core/interpreter.py
+++ b/numba/core/interpreter.py
@@ -877,20 +877,23 @@ class Interpreter(object):
             expr = ir.Expr.call(func, [], [], loc=self.loc, vararg=vararg)
             self.store(expr, res)
 
-    def _build_tuple_unpack(self, inst, tuples, temps):
+    def _build_tuple_unpack(self, inst, tuples, temps, is_assign):
         first = self.get(tuples[0])
-        for other, tmp in zip(map(self.get, tuples[1:]), temps):
-            out = ir.Expr.binop(fn=operator.add, lhs=first, rhs=other,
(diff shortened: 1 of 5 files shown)


numba/core/interpreter.py
```diff
@@ -877,20 +877,23 @@ def op_CALL_FUNCTION_EX(self, inst, func, vararg, res):
...
-    def _build_tuple_unpack(self, inst, tuples, temps):
+    def _build_tuple_unpack(self, inst, tuples, temps, is_assign):
         first = self.get(tuples[0])
-        for other, tmp in zip(map(self.get, tuples[1:]), temps):
-            out = ir.Expr.binop(fn=operator.add, lhs=first, rhs=other,
-                                loc=self.loc)
-            self.store(out, tmp)
-            first = self.get(tmp)
+        if is_assign: # it's assign-like, just store through
+            self.store(first, temps[0])
```
sklam, line 883: There needs to be a convert to tuple.

```

In [5]: @njit
   ...: def foo(x):
   ...:     y = (*x,)
   ...:     return y
   ...:

In [6]: foo([1, 2])

Out[6]: [1, 2]
```

stuartarchibald: Is there actually enough information available to handle this?
```python
In [1]: import dis

In [2]: def foo(x):
   ...:     return (*x,)
   ...: 

In [3]: foo((1,2))
Out[3]: (1, 2)

In [4]: foo([1,2])
Out[4]: (1, 2)

In [5]: dis.dis(foo)
  2           0 LOAD_FAST                0 (x)
              2 BUILD_TUPLE_UNPACK       1
              4 RETURN_VALUE

```
the return type ends up depending on the list size.

sklam: Should we raise UnsupportedError/TypeError instead when the `x` in `(*x,)` is not a `tuple`?

stuartarchibald: I think it has to be an error if it's not a `tuple`. From OOB conversation, going try with emitting an intrinsic to handle this, can't do anything from the bytecode interpreter as there's nothing unique about the tuple vs. anything else situation.

stuartarchibald: Managed to get this to work in 28bdbca.

sklam:
There are some py3.9 specific errors

sklam (approved):
Looks good. Thanks for the patch.

PR #6623, opened 2021-01-06
Fix #6556 extend variable lifetime to block end.

### LINKED ISSUE #6556: Error in Numba-compiled DPOSV call

I have a Numba-compiled function that wraps the BLAS `dposv` function. This function was working correctly up through Numba 0.51. With Numba 0.52, it fails with a DPOSV error. The following code demonstrates the problem.

- Under Numba 0.51, it runs correctly
- Under Numba 0.52 (on Windows, installed from conda-forge), the assertion fails (indicating a DPOSV argument error) and the BLAS implementation (tested with both MKL and OpenBLAS) complains about an invalid argument

As near as I can tell, Numba 0.52 changed something about how this function compiles so it no longer functions. This function is a part of [LensKit](https://github.com/lenskit/lkpy), and it can be run with LensKit's tests as well:

    python -m pytest -k dposv

```python
import numpy as np
import numba as n
import cffi
from numba.extending import get_cython_function_address

__ffi = cffi.FFI()

__uplo_U = np.array([ord('U')], dtype=np.int8)
__uplo_L = np.array([ord('L')], dtype=np.int8)
__dposv = __ffi.cast("void (*) (char*, int*, int*, double*, int*, double*, int*, int*)",
                     get_cython_function_address("scipy.linalg.cython_lapack", "dposv"))


@n.njit(n.intc(n.float64[:, ::1], n.float64[::1], n.boolean))
def _dposv(A, b, lower):
    if A.shape[0] != A.shape[1]:
        return -11
    if A.shape[0] != b.shape[0]:
        return -12

    # dposv uses Fortran-layout arrays. Because we use C-layout arrays, we will
    # invert the meaning of 'lower' and 'trans', and the function will work fine.
    # We also need to swap index orders
    uplo = __uplo_U if lower else __uplo_L
    n_p = __ffi.from_buffer(np.array([A.shape[0]], dtype=np.intc))
    nrhs_p = __ffi.from_buffer(np.ones(1, dtype=np.intc))
    info = np.zeros(1, dtype=np.intc)
    info_p = __ffi.from_buffer(info)

    __dposv(__ffi.from_buffer(uplo), n_p, nrhs_p,
            __ffi.from_buffer(A), n_p,
            __ffi.from_buffer(b), n_p,
            info_p)

    return info[0]

A = np.array([[106014.2749504 ,  53086.26042087],
       [ 53086.26042087, 252421.30232019]])
b = np.array([ 4875.45574154, -4608.45667769])
rv = _dposv(A, b, True)
assert rv >= 0
```

stuartarchibald:
Potential fix for #6556, this delays issuing `ir.Del` nodes for
dead variables until the end of the block in which they become
dead. The result is that variable lifetime is slightly extended but
it prevents use after free issues from too early decref when e.g.
`.ctypes` or FFI buffers are used with arrays.

Fixes #6556

### DIFF
--- a/numba/core/postproc.py
+++ b/numba/core/postproc.py
@@ -197,6 +197,8 @@ class PostProcessor(object):
             # rewrite body and insert dels
             body = []
             lastloc = ir_block.loc
+            # for storing dels
+            del_store = []
             for stmt, delete_set in reversed(delete_pts):
                 lastloc = stmt.loc
                 # Ignore dels (assuming no user inserted deletes)
@@ -205,7 +207,11 @@ class PostProcessor(object):
                 # note: the reverse sort is not necessary for correctness
                 #       it is just to minimize changes to test for now
                 for var_name in sorted(delete_set, reverse=True):
-                    body.append(ir.Del(var_name, loc=lastloc))
+                    del_store.append(ir.Del(var_name, loc=lastloc))
+            # add all the dels at the end of the block, this prevents lifetime
+            # issues occurring in cases where e.g. C function calls are made on
+            # e.g. `array_inst.ctypes.data`.
+            body.extend(del_store)
             body.append(ir_block.body[-1])  # terminator
             ir_block.body = body
 


ehsantn:
I think the right solution is to carry references whenever necessary (just like `array.ctypes` does). This workaround doesn't work in all cases (e.g. there is control flow before use) and has unintended consequences such as potentially massive memory footprint increase in some cases.

stuartarchibald:
> I think the right solution is to carry references whenever necessary (just like `array.ctypes` does). This workaround doesn't work in all cases (e.g. there is control flow before use) and has unintended consequences such as potentially massive memory footprint increase in some cases.

Yeah, just exploring this as it was easiest. RE the memory footprint, think it's still more aggressive than what'd happen in cpython.

stuartarchibald:
Fails for this are seemingly limited to:
* tests which are checking something specific about lifetime behaviour.
* tests which are using loop comprehensions (as the transform for this relies on `ir.Del` location in analysis).
* a couple of tests which look a bit suspicious (some generator, some typed.List).

PokhodenkoSA:
Should this change have some influence on printing annotated code with `NUMBA_DUMP_ANNOTATION=1` ?

stuartarchibald:
Closing as https://github.com/numba/numba/pull/7421 implements this in a way that also lets the current optimisation passes which use lifetimes for analysis work. The fix for #6556 is likely along the lines of what @ehsantn notes RE carrying refs cf. `ctypes`.

PR #6626, opened 2021-01-07
A simple event system to broadcast compiler events

sklam:
as titled.

Here's an example that records all compilation events for a given function: https://nbviewer.jupyter.org/gist/sklam/e6aaa5f9edaf3d092fe1fa4d5079f974


numba/core/event.py
```diff
@@ -0,0 +1,357 @@
...
+        -------
+        res : bool
+        """
+        return self._exc_details[0] is None
+
+    def __str__(self):
+        data = (f"{type(self.data).__qualname__}"
+                if self.data is not None else "None")
+        return f"Event({self._kind}, {self._status}, data: {data})"
+
```
esc, line 128: It may make sense to add `__repr__` here and have it simply call `__str__`. Also, perhaps a simple `__hash__` implementation could be added that would allow using the `Event` class in dictionaries.

numba/core/event.py
```diff
@@ -0,0 +1,357 @@
...
+        self._depth -= 1
+        if self._depth == 0:
+            self._duration = timer() - self._ts
+
+    @property
+    def done(self):
+        """Returns a ``bool`` indicating a measurement is made.
+
+        When this returns ``False``, the matching event has never fired.
+        Iff this returns ``True``, ``.duration`` can be read without error.
```
esc, line 227: I presume `Iff` here means 'if and only if'.?

numba/core/event.py
```diff
@@ -0,0 +1,357 @@
...
+
+
+@contextmanager
+def install_listener(kind, listener):
+    """Install a listener temporarily within the duration of the context.
+
+    Returns
+    -------
+    res : Listener
+        The *listener* provided.
```
esc, line 264: It may be worthwhile adding a simple example of how to use this?

esc: Maybe also for the other context managers below?

numba/core/event.py
```diff
@@ -0,0 +1,357 @@
...
+    kind : str
+        Event kind.
+    data : any; optional
+        Extra event data.
+    """
+    evt = Event(
+        kind=kind, status=EventStatus.END, data=data, exc_details=exc_details,
+    )
+    broadcast(evt)
+
```
esc, line 337: In addition to the plain `start_event` and `end_event` it may be useful to wrap the builtin kinds here for API convenience. like `start_compiler_lock` and `end_compiler_lock` as well as `start_compile` and `end_compile`. What do you think.

esc (commented):
I reviewed this by looking at the API from a programmer perspective. I left a few comments about documentation and some additional convenience. 

The pattern of having a global singleton to register and broadcast events reminds me very much of the `logging` package which comes with certain issues resulting from it's design as globally accessible structure and the usual issues associated with globals. In it's current form the registration and broadcast facilities are very much bound to the `numba/core/event` module and I am wondering if we could abstract this some more.

For example, how about a `Broadcaster` class that contains the list of listeners and facilities to register/unregister them? The module would then contain an instantiated singleton and all the module level functions (`broadcast`, `register` etc...) would redirect to that singleton. This might make for a slightly cleaner design and support re-usability in future (e.g. llvmlite)?

sklam:
@esc, I'm not sure what the singleton pattern would buy us here. The `event` module is acting as the singleton object already. This event system is not expected to be reused elsewhere. The `logging` module is more complex and customizable. This event system is probably more like the `atexit` module in stdlib.

My plan for llvmlite is to add a simple callback registration system around the ctypes wrapper, so that custom code can be injected before and after every call into the LLVM C-API. Numba would then inject a function to signal the start of llvmlite-lock and a function to signal the end.

numba/core/event.py
sklam: I don't think the extra wrappers are saving that much. I also don't want too many ways to do the same thing. Most cases can use the `mark_event()` context manager, which is only one call instead of two. I'm deliberately keep it as simple and minimal as possible.

numba/core/event.py
sklam: Added `__repr__` in https://github.com/numba/numba/pull/6626/commits/1b814b86df9c9b272c32643999a88ba75b66245b.
But `__hash__` will be tricky because `._data` and `._exc_details` may not be hashable. Hashing on just `._kind` and and `._status` is not that helpful. I don't know what behavior it should have in a `dict` or `set`. Furthermore, I think `Event` should only be used in a sequence for replaying the events. Users should even avoid storing events for a long period.

numba/core/event.py
sklam: Expanded to the full spelling.

numba/core/event.py
sklam: Docs added in https://github.com/numba/numba/pull/6626/commits/1b814b86df9c9b272c32643999a88ba75b66245b

numba/core/event.py
esc: sounds good.

numba/core/event.py
esc: Great!

numba/core/event.py
esc: OK, makes sense.

esc:
@sklam makes sense, in that case probably best to just leave it like this.

numba/core/event.py
esc: Thanks!

numba/core/dispatcher.py
```diff
@@ -1014,38 +1040,55 @@ def _pre_compile(self, args, return_type, flags):
...
+                                            lifted=(),
+                                            lifted_from=self.lifted_from,
+                                            is_lifted_loop=True,)
+
+                    # Check typing error if object mode is used
+                    if (cres.typing_error is not None and
+                            not flags.enable_pyobject):
+                        raise cres.typing_error
+                    self.add_overload(cres)
+                return cres.entry_point
```
stuartarchibald, line 1091: NOTE: Code motion.

docs/source/developer/event_api.rst
```diff
@@ -0,0 +1,5 @@
+Event API
+=========
+
+.. automodule:: numba.core.event
+    :members:
```
stuartarchibald, line 5: I like this approach

numba/core/compiler_lock.py
```diff
@@ -1,17 +1,21 @@
 import threading
 import functools
+import numba.core.event as ev
 
 
 # Lock for the preventing multiple compiler execution
 class _CompilerLock(object):
     def __init__(self):
         self._lock = threading.RLock()
+        self._duration = 0
```
stuartarchibald, line 10: Ought this be a float?

numba/core/dispatcher.py
```diff
@@ -681,6 +683,15 @@ def typeof_pyval(self, val):
...
         return tp
 
+    def _callback_add_compiler_timer(self, duration, cres):
+        md = cres.metadata
+        # md can be None when code is loaded from cache
+        if md is not None:
+            timers = md.setdefault("timers", {})
+            if "compiler_lock" not in timers:
+                # Only write if the metadata does not exist
+                timers["compiler_lock"] = duration
```
stuartarchibald, line 693: Is it an error if it does exist?

stuartarchibald (commented):
Thanks for the patch, great to see this long requested feature implemented. I've given it an initial review, the code looks good, there's a lot of comments but 90% of them are typos/for the pattern `"<python type> : <argument name>"` in docstrings which seems to render incorrectly. I'm now going to try and use this/write some tests and will report back. Thanks again!

numba/core/event.py
```diff
@@ -0,0 +1,407 @@
...
+    *data*.
+
+    Parameters
+    ----------
+    kind : str
+        Event kind.
+    data : any; optional
+        Extra event data.
+    """
+    with ExitStack() as scope:
```
stuartarchibald, line 401: Would `try/finally` suffice?

numba/core/compiler_lock.py
sklam: i'm just relying on Python duck typing it.

docs/source/developer/event_api.rst
sklam: We should do this more. Less duplication

numba/core/compiler_lock.py
sklam: actually, that's a leftover dead code i missed

numba/core/event.py
```diff
@@ -0,0 +1,407 @@
...
+    """Register a listener for a given event kind.
+
+    Parameters
+    ----------
+    kind : str
+    listener : Listener
+    """
+    assert isinstance(listener, Listener)
+    kind = _guard_kind(kind)
+    _registered[kind].append(listener)
```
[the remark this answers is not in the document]
sklam: There can be any number of listener on any event kind. Just that `_guard_kind` is preventing users from adding new builtin event kind---anything that starts with "numba:"

numba/core/event.py
```diff
@@ -0,0 +1,407 @@
...
+
+        Parameters
+        ----------
+        event : Event
+        """
+        pass
+
+    def notify(self, event):
+        """Notify this Listener with the given Event.
+        """
```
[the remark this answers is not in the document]
sklam: will add

numba/core/event.py
sklam: was using `try/finally` until I needed extra code to fish out the exception. The `ExitStack` let's me push a function like a `__exit__()` to capture the 3-tuple exception infos.

numba/core/dispatcher.py
sklam: right. will make it an error.

numba/core/event.py
```diff
@@ -0,0 +1,407 @@
...
+    """
+    if kind.startswith("numba:") and kind not in _builtin_kinds:
+        raise ValueError(f"{kind} is not a valid event kind")
+    return kind
+
+
+class Event:
+    """An event.
+    """
+    def __init__(self, kind, status, data=None, exc_details=None):
```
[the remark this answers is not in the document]
sklam: Should be fixed now

sklam:
Thanks for all the reviewing. I have addressed all the comments.

stuartarchibald:
> as titled.
> 
> Here's an example that records all compilation events for a given function: https://nbviewer.jupyter.org/gist/sklam/e6aaa5f9edaf3d092fe1fa4d5079f974

I've run this demo, can confirm it works, output looks great.

numba/core/event.py
stuartarchibald: ah, I see, seems like this is a good choice then.

numba/core/dispatcher.py
stuartarchibald: thanks

docs/source/developer/event_api.rst
stuartarchibald: Agree

numba/core/event.py
```diff
@@ -0,0 +1,420 @@
...
+class Event:
+    """An event.
+
+    Parameters
+    ----------
+    kind : str
+    status : EventStatus
+    data : any; optional
+        Additional data for the event.
+    exc_details : 3-tuple; optional
```
stuartarchibald, line 77:
```suggestion
    kind: str
    status: EventStatus
    data: any; optional
        Additional data for the event.
    exc_details: 3-tuple; optional
```

numba/core/event.py
```diff
@@ -0,0 +1,420 @@
...
+        event: Event
+        """
+        pass
+
+    def notify(self, event):
+        """Notify this Listener with the given Event.
+
+        Parameters
+        ----------
+        event : Event
```
stuartarchibald, line 228:
```suggestion
        event: Event
```

numba/core/event.py
```diff
@@ -0,0 +1,420 @@
...
+def end_event(kind, data=None, exc_details=None):
+    """Trigger the end of an event of *kind*, *exc_details*.
+
+    Parameters
+    ----------
+    kind: str
+        Event kind.
+    data: any; optional
+        Extra event data.
+    exc_details : 3-tuple; optional
```
stuartarchibald, line 392:
```suggestion
    exc_details: 3-tuple; optional
```

numba/core/event.py
```diff
@@ -0,0 +1,420 @@
...
+    if tl.done:
+        callback(tl.duration)
+
+
+@contextmanager
+def install_recorder(kind):
+    """Install a RecordingListener temporarily to record all events.
+
+    Once the context is closed, users can use ``RecordingListener.buffer``
+    to acccess the recorded events.
```
stuartarchibald, line 350:
```suggestion
    to access the recorded events.
```

stuartarchibald (commented):
Thanks very much for the fixes, just a few typos left to resolve else the code looks good. I've also trialed and manually tested the API and it is easy to use and works as expected. Feedback from production/use should feed into further changes if/as needed. Happy to merge once the typo fixes are accepted. Thanks again!

numba/core/event.py
sklam: numpydoc rendering issue should be fixed.

docs/environment.yml
```diff
@@ -10,3 +10,5 @@ dependencies:
   - numpy
   - numpydoc
   - setuptools
+  # The following is needed to fix RTD.
+  - conda
```
stuartarchibald, line 14:
```suggestion
  - conda

```
New line at EOF?

stuartarchibald (approved):
Thanks for all the fixes (especially for sorting out RTD rendering!). Patch looks good.

dlee992:
Wow, this is really nice for use. Maybe we can put this demo https://nbviewer.jupyter.org/gist/sklam/e6aaa5f9edaf3d092fe1fa4d5079f974 into numba official doc, then more people can easily find this feature and use it well.

esc:
> Wow, this is really nice for use. Maybe we can put this demo https://nbviewer.jupyter.org/gist/sklam/e6aaa5f9edaf3d092fe1fa4d5079f974 into numba official doc, then more people can easily find this feature and use it well.

@dlee992 that's probably a good idea, next step would be to create an issue asking for this as a followup.

PR #6647, opened 2021-01-17
Support prefer_literal option for intrinsic decorator

### LINKED ISSUE #6533: Support prefer_literal option for intrinsic decorator

Hi, 

I noticed that `prefer_literal` option that was added recently for overload decorators is not supported for `intrinsic`.
It seems that `make_intrinsic_template` does not have prefer_literal argument (unlike for example `make_overload_template`) and moreover objects of `_Intrinsic` class do not store any options at all. At the same time intrinsic templates are created from base `FunctionTemplate`, which has `prefer_literal` hardcoded to False:
https://github.com/numba/numba/blob/4f985a5d90c4202c5f77bbd05fcff894200752de/numba/core/typing/templates.py#L261

This results in multiple calls to `get_call_type` with non-literal arguments during type-inference and may cause subtle issues if intrinsic code does not have correct input type checks. For instance the output of the following script

```
import numba
from numba import types
from numba.extending import overload, intrinsic
from numba.core.typing.templates import signature
import numpy as np

@intrinsic
# @intrinsic(prefer_literal=True)
def tuple_take(typingctx, data, idxs):

    print("DEBUG: tuple_take: data", data)
    print("DEBUG: tuple_take: idxs", idxs)
    if not isinstance(data, types.Tuple):
        return None
    if not (isinstance(idxs, types.Tuple)
            and all(isinstance(x, types.IntegerLiteral) for x in idxs)):
        return None

    old_ncols = len(data)
    unlit_keep_idxs = [a.literal_value for a in idxs]
    keep_col_idxs, new_typs = zip(*((i, t) for i, t in enumerate(data) if i in unlit_keep_idxs))
    def codegen(context, builder, signature, args):
        old_data = args[0]
        new_data = [
            builder.extract_value(
                old_data,
                types.intp(i)
            ) for i in range(old_ncols) if i in keep_col_idxs
        ]

        res = context.make_tuple(builder, ret_typ, new_data)
        return res

    ret_typ = types.Tuple(new_typs)
    sig = signature(ret_typ, data, idxs)
    return sig, codegen

def foo(data, idxs):
    pass

@overload(foo, prefer_literal=True)
def foo_ovld(data, idxs):
    print(f"DEBUG: foo: data={data}")
    print(f"DEBUG: foo: idxs={idxs}")

    def foo_impl(data, idxs):
        return tuple_take(data, idxs)
    return foo_impl

@numba.njit
def test_func():
    data = ('a', [2, 3], np.nan, 600)
    keep_idxs = (1, 3)
    return foo(data, keep_idxs)

res = test_func()
print(f"DEBUG: res={res}")
```
shows that we entered `tuple_take` code twice with non-literal arguments
> DEBUG: foo: data=Tuple(Literal[str](a), list(int64)<iv=[2, 3]>, float64, Literal[int](600))
> DEBUG: foo: idxs=Tuple(Literal[int](1), Literal[int](3))
> DEBUG: tuple_take: data Tuple(unicode_type, list(int64)<iv=None>, float64, int64)
> DEBUG: tuple_take: idxs UniTuple(int64 x 2)
> DEBUG: tuple_take: data Tuple(Literal[str](a), list(int64)<iv=[2, 3]>, float64, Literal[int](600))
> DEBUG: tuple_take: idxs Tuple(Literal[int](1), Literal[int](3))
> DEBUG: tuple_take: data Tuple(unicode_type, list(int64)<iv=None>, float64, int64)
> DEBUG: tuple_take: idxs UniTuple(int64 x 2)
> DEBUG: res=([2, 3], 600)

and in the ideal case it can be this way:
> DEBUG: foo: data=Tuple(Literal[str](a), list(int64)<iv=[2, 3]>, float64, Literal[int](600))
> DEBUG: foo: idxs=Tuple(Literal[int](1), Literal[int](3))
> DEBUG: tuple_take: data Tuple(Literal[str](a), list(int64)<iv=[2, 3]>, float64, Literal[int](600))
> DEBUG: tuple_take: idxs Tuple(Literal[int](1), Literal[int](3))
> DEBUG: res=([2, 3], 600)

Can this be changed in the future and prefer_literal added to intrinsic too?

BR,
Alexey.

ashutoshvarma:
Fixes #6533 

As titled, add support for prefer_literal option in intrinsic decorator.

PS: I am not so sure about the test I made, any suggestions would help.


ashutoshvarma:
```python
class TestIntrinsicPreferLiteral(TestCase):
    def test_intrinsic(self):
        int64_100 = ir.Constant(ir.IntType(64), 100)
        int64_cafe = ir.Constant(ir.IntType(64), 0xcafe)
        
        def intrin(context, x):
            sig = signature(types.int64, x)
            
            if isinstance(x, types.IntegerLiteral):
                # With prefer_literal=False, this branch will not be reached.
                if x.literal_value == 1:
                    def codegen(context, builder, signature, args):
                        return int64_cafe
                    return sig, codegen
                else:
                    raise errors.TypingError('literal value')
            else:
                def codegen(context, builder, signature, args):
                    return builder.mul(args[0], int64_100)
                return sig, codegen

        prefer_lit =  intrinsic(prefer_literal=True)(intrin)
        non_lit = intrinsic(prefer_literal=False)(intrin)

        @njit
        def check_prefer_lit(x):
            return prefer_lit(1), prefer_lit(2), prefer_lit(x)

        a, b, c = check_prefer_lit(3)
        self.assertEqual(a, 0xcafe)
        self.assertEqual(b, 200)
        self.assertEqual(c, 300)

        @njit
        def check_non_lit(x):
            return non_lit(1), non_lit(2), non_lit(x)

        a, b, c = check_non_lit(3)
        self.assertEqual(a, 100)
        self.assertEqual(b, 200)
        self.assertEqual(c, 300)
```
(This is based on `TestOverloadPreferLiteral` testcase from `test_extending.py`)

I think this would more appropriate than the one I wrote previously.

numba/tests/test_extending.py
```diff
@@ -1910,5 +1911,47 @@ def check_non_lit(dummy, x):
         self.assertEqual(c, 300)
 
 
+class TestIntrinsicPreferLiteral(TestCase):
+    def test_intrinsic(self):
+        int64_100 = ir.Constant(ir.IntType(64), 100)
+        int64_cafe = ir.Constant(ir.IntType(64), 0xcafe)
+
+        def intrin(context, x):
+            sig = signature(types.int64, x)
```
ashutoshvarma, line 1920: The test is failing for 32bit because of explicit int64 types.

stuartarchibald:

Also, having just skimmed the PR, perhaps try using `context.resolve_value_type` on a Numba type to get a LLVM type (and perhaps use `numba.types.intp` for your `int`s?). It also might be useful to call `self.assert*` from within the `@overload` to check that the types are literal and the amount of "trials" of resolving the overload is 1 in the case of `prefer_literal=True`.

ashutoshvarma:
Thanks for your suggestion @stuartarchibald ! Changing signature's return type to `intp`  seems to do the trick for 32 bit.


> It also might be useful to call self.assert* from within the @overload to check that the types are literal and the amount of "trials" of resolving the overload is 1 in the case of prefer_literal=True

I am assuming you mean `@intrinsic` instead of `@overload`.
If I get you right, using `self.assert*` inside `njit` would result in compie error, wouldn't it?
> We will begin reviewing PRs again once the RC is out and we thank you in advance for your patience!

No problem ! Take your time 👍

sklam (dismissed):
I'm approving for all changes not made by me. We'll need a second reviewer for my changes

docs/upcoming_changes/6647.change.rst
```diff
@@ -0,0 +1,6 @@
+Support for ``@intrinsic(prefer_literal=True)``
+===============================================
+
+This PR adds ``prefer_literal`` option for the ``intrinsic`` decorator to 
+prioritize the use of literal types when available. This has the same 
+behavior as in the ``prefer_literal`` option in the ``overload`` decorator.
```
stuartarchibald, line 6:
```suggestion
In the high level extension API, the ``prefer_literal`` option is added to the
``numba.extending.intrinsic`` decorator to prioritize the use of literal types
when available. This has the same behavior as in the ``prefer_literal``
option in the ``numba.extending.overload`` decorator.
```
suggest this change, was hoping to clarify that this is an update to the high level extension API. What do you think?

docs/upcoming_changes/6647.change.rst
```diff
@@ -0,0 +1,6 @@
+Support for ``@intrinsic(prefer_literal=True)``
+===============================================
```
stuartarchibald, line 2:
```suggestion
Support for ``@numba.extending.intrinsic(prefer_literal=True)``
===============================================================
```

numba/core/typing/templates.py
```diff
@@ -999,15 +999,17 @@ def get_template_info(self):
         return info
 
 
-def make_intrinsic_template(handle, defn, name, kwargs):
+def make_intrinsic_template(handle, defn, name, *, prefer_literal=False,
+                            kwargs={}):
```
stuartarchibald, line 1003:
```suggestion
                            kwargs=None):
   kwargs = MappingProxyType({} if kwargs is None else kwargs)
```
should this be like this to avoid adding a mutable kwarg default?

stuartarchibald (commented):
Thanks for the updates @sklam, I've reviewed 943ff13 c1ca681 and c92a576 as you approved the rest, couple of minor comments else looks good.

numba/core/typing/templates.py
sklam: Good point

numba/core/typing/templates.py
```diff
@@ -999,15 +999,18 @@ def get_template_info(self):
         return info
 
 
-def make_intrinsic_template(handle, defn, name, kwargs):
+def make_intrinsic_template(handle, defn, name, *, prefer_literal=False,
+                            kwargs=None):
+   kwargs = MappingProxyType({} if kwargs is None else kwargs)
```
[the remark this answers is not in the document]
sklam: fixed up at https://github.com/numba/numba/pull/6647/commits/0e12ac5d8d6325263e18659d969d2571082dcc6a
I also notice that the line went before the docstring and that a import is missing.

stuartarchibald (approved):
Thanks for the fixes @sklam, I'm approving afa4a2d and 0e12ac5, which I think completes this review.

PR #6652, opened 2021-01-20
Simplify PR-6334

sklam:
Based on #6334.

It looks like the same can be achieved without adding a `builder` argument to `get_getattr`.


numba/tests/test_objects.py
```diff
@@ -30,6 +30,12 @@ def delattr_usecase(o):
 
 
 class TestAttributes(TestCase):
+    def test_getattr(self, flags=enable_pyobj_flags):
+        import numba.tests.usecases as uc
+        @njit
+        def f():
+            return uc._GLOBAL_STR
+        self.assertEquals(f(), "abc")
```
stuartarchibald, line 38: Puzzled as to why this test is here? This module seems to be related to "objects" but this is compiled under `nopython` mode?

numba/tests/test_objects.py
```diff
@@ -30,6 +30,12 @@ def delattr_usecase(o):
 
 
 class TestAttributes(TestCase):
+    def test_getattr(self, flags=enable_pyobj_flags):
```
stuartarchibald, line 33: `flags` unused ?

numba/tests/usecases.py
```diff
@@ -2,6 +2,8 @@
 import numpy as np
 from numba import jit
 
+_GLOBAL_STR = "abc"
```
stuartarchibald, line 5: As a total aside, this file itself seems a little strange with respect to the rest of testing, is it perhaps a legacy item?

numba/tests/test_objects.py
sklam: Good point. I didn't pay attention to the test file when I reviewed the original code.

sklam: i think `test_auto_constants.py` is the closest test file.

stuartarchibald: Looks good.

numba/tests/usecases.py
sklam: This file along with `test_usecases.py` are probably the oldest surviving test files and they are exercising the essentials of the core compiler.

numba/tests/test_objects.py
sklam: removed

sklam:
All comments are addressed.

numba/tests/test_auto_constants.py
```diff
@@ -27,6 +29,12 @@ def pyfunc():
 
         self.assertEqual(pyfunc(), cfunc())
 
+    def test_module_string_constant(self):
+        @njit
+        def f():
+            return uc._GLOBAL_STR
+        self.assertEquals(f(), "abc")
```
stuartarchibald, line 36:
```suggestion
        self.assertEquals(f(), f.py_func())
```
saves keeping in sync

numba/tests/test_auto_constants.py
```diff
@@ -27,7 +29,12 @@ def pyfunc():
 
         self.assertEqual(pyfunc(), cfunc())
 
+    def test_module_string_constant(self):
+        @njit
+        def f():
+            return uc._GLOBAL_STR
+        self.assertEquals(f(), f.py_func())
```
stuartarchibald, line 36: Also, shouldn't this be `assertEqual`? `assertEquals` is a deprecated alias https://docs.python.org/3/library/unittest.html#deprecated-aliases

stuartarchibald (approved):

PR #6653, opened 2021-01-20
Fix get_numba_envvar

### LINKED ISSUE #6601: get_numba_envvar doesn't seem to return the right value
If the env var value is found, Instead of returning the value, the function returns the default value.

https://github.com/numba/numba/blob/42050e4a306e53f332fbaecf45248c54fae752e5/numba/cuda/envvars.py#L32

sklam:
Fixes #6601

### DIFF
--- a/numba/cuda/tests/nocuda/test_library_lookup.py
+++ b/numba/cuda/tests/nocuda/test_library_lookup.py
@@ -305,3 +305,7 @@ class TestMisc(SerialMixin, unittest.TestCase):
         with override_env_config(f"NUMBA_{env}", "my value"):
             got = envvars.get_numba_envvar(env, default=default)
             self.assertEqual(got, "my value")
+
+
+if __name__ == '__main__':
+    unittest.main()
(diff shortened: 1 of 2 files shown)


numba/cuda/tests/nocuda/test_library_lookup.py
```diff
@@ -282,3 +284,24 @@ def _fake_non_conda_env():
...
+        default = "the default"
+
+        # Check that we are getting the default
+        got = envvars.get_numba_envvar(env, default=default)
+        self.assertIs(got, default)
+
+        # Check that we are reading the env-var correctly
+        with override_env_config(f"NUMBA_{env}", "my value"):
+            got = envvars.get_numba_envvar(env, default=default)
+            self.assertEqual(got, "my value")
```
stuartarchibald, line 307: This seems to be missing:
```
if __name__ == '__main__':
    unittest.main()
```

stuartarchibald:
Patch looks good, one comment about missing `unittest.main()`, would you like to resolve that whilst this file is being changed?

numba/cuda/tests/nocuda/test_library_lookup.py
sklam: added

stuartarchibald (approved):

PR #6030, opened 2020-07-24
CUDA: Don't optimize IR before sending it to NVVM

gmarkall:
There are various issues created by optimizing the IR before sending it to NVVM, in particular:

- Issue #5576: Optimizing the IR before sending it to NVVM results in code that makes out-of-bounds accesses.
- Issue #6022: Optimization generates memset intrinsics incompatible with LLVM 3.4 (NVVM).
- IR auto-upgrade breaks the IR for LLVM 3.4 due to changes in atomic instructions. This requires a patch to the LLVM version used for llvmlite, which is problematic because many distributions won't carry this patch. (ref: https://github.com/numba/llvmlite/pull/593)

Optimizing IR prior to sending it to NVVM was originally added to work around an NVVM bug documented as Issue #1341. The cause of issue #1341 is now resolved in NVVM, and has been since CUDA 7.5 - the test script at https://gist.github.com/sklam/f62f1f48bb0be78f9ceb executes in a fraction of a second and consumes around 100MB of RAM - the bug manifested itself as a long compilation time with gigabytes of RAM consumed.

There are some side-effects to not optimizing before sending to NVVM, which are reflected in the tests:

- `TestCudaConstantMemory.test_const_array_3d`: Instead of reading the complex value as a vector of u32s, there are two individual loads of the real and imaginary parts as f32s. This is not expected to have a negative impact on performance, since coalesced reading of f32s will use memory bandwidth effectively.
- `TestCudaConstantMemory.test_const_record`: The generated code here appears less optimal than it previously was - instead of loading three 8-bit members in a single 32-bit read, the reads are now individual 8-bit reads.
- `TestCudaConstantMemory.test_const_record_align`: A similar issue to test_const_record.
- `TestNvvmWithoutCuda.test_nvvm_memset_fixup`: now that IR is not optimized prior to sending to NVVM, memsets no longer appear in the IR. It is desirable to keep the memset fixup functionality and test so that Numba and its extensions will be able to use memsets (e.g. from `cgutils.memset`), so this test is modified to use strings of IR rather than relying on memsets naturally being produced as part of the compilation process. It is possible to generate a memset without an alignment attribute on the destination with `cgutils.memset`, so a check for this error and a test of the check are also added.

For the issues with less optimal code being generated in `TestCudaConstantMemory`, the tests for the optimal code have been moved into new test cases that are marked as expected fails - it would be better to keep these tests and note the sub-optimal behaviour, with the expectation that a future version of NVVM may improve on the code it currently generates.

PR #6661, opened 2021-01-25
CUDA: Support NVVM70 / CUDA 11.2

### LINKED ISSUE #6607: <unnamed> (81, 19): parse expected comma after load's type
### Reporting a bug

Im getting an error like pasted below while trying to run the most basic example taken from official WWW:
https://developer.nvidia.com/blog/numba-python-cuda-acceleration/

```
import numpy as np
from numba import vectorize

@vectorize(['float32(float32, float32)'], target='cuda')
def Add(a, b):
  return a + b

N = 100000
A = np.ones(N, dtype=np.float32)
B = np.ones(A.shape, dtype=A.dtype)
C = np.empty_like(A, dtype=A.dtype)

C = Add(A, B)
```
Ubuntu 18.04.5, Numpy 1.19.4, Numba 0.52.0, kernel 5.4.58-050458-generic, CUDA 11.2.0, NVIDIA driver 460.27.04, GeForce GTX 1660

>   File "GPUtest.py", line 13, in <module>
>     C = Add(A, B)
>   File "/usr/local/lib/python3.6/dist-packages/numba/cuda/dispatcher.py", line 37, in __call__
>     return CUDAUFuncMechanism.call(self.functions, args, kws)
>   File "/usr/local/lib/python3.6/dist-packages/numba/np/ufunc/deviceufunc.py", line 300, in call
>     cr.launch(func, shape[0], stream, devarys)
>   File "/usr/local/lib/python3.6/dist-packages/numba/cuda/dispatcher.py", line 148, in launch
>     func.forall(count, stream=stream)(*args)
>   File "/usr/local/lib/python3.6/dist-packages/numba/cuda/compiler.py", line 372, in __call__
>     kernel = self.kernel.specialize(*args)
>   File "/usr/local/lib/python3.6/dist-packages/numba/cuda/compiler.py", line 899, in specialize
>     self._bind, targetoptions)
>   File "/usr/local/lib/python3.6/dist-packages/numba/cuda/compiler.py", line 811, in __init__
>     self.compile(sigs[0])
>   File "/usr/local/lib/python3.6/dist-packages/numba/cuda/compiler.py", line 952, in compile
>     kernel.bind()
>   File "/usr/local/lib/python3.6/dist-packages/numba/cuda/compiler.py", line 576, in bind
>     self._func.get()
>   File "/usr/local/lib/python3.6/dist-packages/numba/cuda/compiler.py", line 446, in get
>     ptx = self.ptx.get()
>   File "/usr/local/lib/python3.6/dist-packages/numba/cuda/compiler.py", line 416, in get
>     **self._extra_options)
>   File "/usr/local/lib/python3.6/dist-packages/numba/cuda/cudadrv/nvvm.py", line 548, in llvm_to_ptx
>     ptx = cu.compile(**opts)
>   File "/usr/local/lib/python3.6/dist-packages/numba/cuda/cudadrv/nvvm.py", line 236, in compile
>     self._try_error(err, 'Failed to compile\n')
>   File "/usr/local/lib/python3.6/dist-packages/numba/cuda/cudadrv/nvvm.py", line 254, in _try_error
>     self.driver.check_error(err, "%s\n%s" % (msg, self.get_log()))
>   File "/usr/local/lib/python3.6/dist-packages/numba/cuda/cudadrv/nvvm.py", line 144, in check_error
>     raise exc
> numba.cuda.cudadrv.error.NvvmError: Failed to compile
> 
> <unnamed> (81, 19): parse expected comma after load's type
> NVVM_ERROR_COMPILATION

Is some default behavior changed and broke the decorator?

gmarkall:
Starting with CUDA 11.2, a new version of NVVM is provided that is basedon LLVM 7.0. This requires a number of changes to support, which must be maintained in parallel with the existing support for NVVM based on LLVM 3.4. This PR adds these changes, which consist of:

- Addition of a function to query the NVVM IR version, and a property indicating whether the NVVM in use is based on LLVM 3.4 or 7.0 (`is_nvvm70`).
- The CAS hack (inserting a text-based implementation of `cmpxchg` with pre-LLVM 3.5 semantics in a function) is only needed with NVVM 3.4 - on NVVM 7.0, llvmlite is used to build `cmpxchg` instructions directly
  instead.
- Templates for other atomics (inc, dec, min, max) have the right form of the `cmpxchg` instruction inserted depending on the NVVM version.
- The datalayout shorthand is now only replaced for NVVM 3.4.
- There are now two variants of the functions to rewrite the IR - `llvm100_to_70_ir` and `llvm100_to_34_ir`. `llvm100_to_34_ir` is the old `llvm_39_to_34_ir` with a name reflecting what it currently does.
- `llvm100_to_70_ir` removes the `willreturn` attribute from functions, as it is not supported by LLVM 7.0. It also converts DISPFlags to main subprogram DIFlags.  For example, `spflags: DISPFlagDefinition | DISPFlagOptimized` is rewritten as `isDefinition: true, isOptimized: true`.
- For NVVM 7.0, the `DIBuilder` also used for the CPU target can be used, instead of the `NvvmDIBuilder` that was needed to support NVVM 3.4.
- Some tests are updated to support modified function names, and also to expect a CUDA version of 11.2.
- `test_nvvm_driver` is updated to include appropriate IR for both NVVM 3.4 and 7.0. Some refactoring also makes its code clearer (e.g. renaming `get_ptx()` to `get_nvvimir()`, because it returns NVVM IR and not PTX).
- Some optimizations in LLVM 7.0 result in different code generation in `test_constmem`, so alternative expected results are added for when NVVM 7.0 is used. Note that this recovers some optimizations that were lost when IR optimization using llvmlite was switched off (PR #6030, "Don't optimize IR before sending it to NVVM").
- `test_debuginfo` is updated to match the format of the debuginfo section produced by both NVVM 3.4 and 7.0 (there is some variation in whitespace between these versions).

Fixes #6607.

### DIFF
--- a/numba/cuda/cudadrv/nvvm.py
+++ b/numba/cuda/cudadrv/nvvm.py
@@ -141,7 +141,7 @@ class NVVM(object):
 
     @property
     def is_nvvm70(self):
-        # NVVM70 uses IR version 1.6. See the documentation for
+        # NVVM70 uses NVVM IR version 1.6. See the documentation for
         # nvvmAddModuleToProgram in
         # https://docs.nvidia.com/cuda/libnvvm-api/group__compilation.html
         return (self._majorIR, self._minorIR) >= (1, 6)
@@ -907,10 +907,10 @@ def set_cuda_kernel(lfunc):
     # Set NVVM IR version
     i32 = ir.IntType(32)
     if NVVM().is_nvvm70:
-        # NNVM IR 1.6, DWARF 3.0
+        # NVVM IR 1.6, DWARF 3.0
         ir_versions = [i32(1), i32(6), i32(3), i32(0)]
     else:
-        # NNVM IR 1.1, DWARF 2.0
+        # NVVM IR 1.1, DWARF 2.0
         ir_versions = [i32(1), i32(2), i32(2), i32(0)]
 
     md_ver = m.add_metadata(ir_versions)
--- a/numba/cuda/cudadrv/nvvm.py
+++ b/numba/cuda/cudadrv/nvvm.py
@@ -101,6 +101,11 @@ class NVVM(object):
 
         # nvvmResult nvvmGetProgramLog(nvvmProgram cu, char *buffer)
         'nvvmGetProgramLog': (nvvm_result, nvvm_program, c_char_p),
+
+        # nvvmResult nvvmIRVersion (int* majorIR, int* minorIR, int* majorDbg,
+        #                           int* minorDbg )
+        'nvvmIRVersion': (nvvm_result, POINTER(c_int), POINTER(c_int),
+                          POINTER(c_int), POINTER(c_int)),
     }
 
     # Singleton reference
@@ -127,12 +132,33 @@ class NVVM(object):
 
         return cls.__INSTANCE
 
+    def __init__(self):
+        ir_versions = self.get_ir_version()
+        self._majorIR = ir_versions[0]
+        self._minorIR = ir_versions[1]
+        self._majorDbg = ir_versions[2]
+        self._minorDbg = ir_versions[3]
+
+    @property
+    def is_nvvm70(self):
+        return (self._majorIR, self._minorIR) >= (1, 6)
+
     def get_version(self):
         major = c_int()
         minor = c_int()
         err = self.nvvmVersion(byref(major), byref(minor))
         self.check_error(err, 'Failed to get version.')
         return major.value, minor.value
+
+    def get_ir_version(self):
+        majorIR = c_int()
+        minorIR = c_int()
+        majorDbg = c_int()
+        minorDbg = c_int()
+        err = self.nvvmIRVersion(byref(majorIR), byref(minorIR),
+                                 byref(majorDbg), byref(minorDbg))
+        self.check_error(err, 'Failed to get IR version.')
+        return majorIR.value, minorIR.value, majorDbg.value, minorDbg.value
 
     def check_error(self, error, msg, exit=False):
         if error:
@@ -399,15 +425,26 @@ class LibDevice(object):
         return self.bc
 
 
-ir_numba_atomic_cas = """
+ir_numba_cas_hack = """
 define internal {T} @___numba_atomic_{T}_cas_hack({T}* %ptr, {T} %cmp, {T} %val) alwaysinline {{
     %out = cmpxchg volatile {T}* %ptr, {T} %cmp, {T} %val monotonic
     ret {T} %out
 }}
 """ # noqa: E501
 
+cas_nvvm70 = """
+    %cas_success = cmpxchg volatile {Ti}* %iptr, {Ti} %old, {Ti} %new monotonic monotonic
+    %cas = extractvalue {{ {Ti}, i1 }} %cas_success, 0
+""" # noqa: E501
+
+
+cas_nvvm34 = """
+    %cas = cmpxchg volatile {Ti}* %iptr, {Ti} %old, {Ti} %new monotonic
+""" # noqa: E501
+
+
 # Translation of code from CUDA Programming Guide v6.5, section B.12
-ir_numba_atomic_binary = """
+ir_numba_atomic_binary_template = """
 define internal {T} @___numba_atomic_{T}_{FUNC}({T}* %ptr, {T} %val) alwaysinline {{
 entry:
     %iptr = bitcast {T}* %ptr to {Ti}*
@@ -419,7 +456,7 @@ attempt:
     %dold = bitcast {Ti} %old to {T}
     %dnew = {OP} {T} %dold, %val
     %new = bitcast {T} %dnew to {Ti}
-    %cas = cmpxchg volatile {Ti}* %iptr, {Ti} %old, {Ti} %new monotonic
+    {CAS}
     %repeat = icmp ne {Ti} %cas, %old
     br i1 %repeat, label %attempt, label %done
 
@@ -429,10 +466,10 @@ done:
 }}
 """ # noqa: E501
 
-ir_numba_atomic_inc = """
-define internal {T} @___numba_atomic_{Tu}_inc({T}* %ptr, {T} %val) alwaysinline {{
+ir_numba_atomic_inc_template = """
+define internal {T} @___numba_atomic_{Tu}_inc({T}* %iptr, {T} %val) alwaysinline {{
 entry:
-    %old2 = load volatile {T}, {T}* %ptr
+    %old2 = load volatile {T}, {T}* %iptr
     br label %attempt
 
 attempt:
@@ -440,7 +477,7 @@ attempt:
     %bndchk = icmp ult {T} %old, %val
     %inc = add {T} %old, 1
     %new = select i1 %bndchk, {T} %inc, {T} 0
-    %cas = cmpxchg volatile {T}* %ptr, {T} %old, {T} %new monotonic
+    {CAS}
     %repeat = icmp ne {T} %cas, %old
     br i1 %repeat, label %attempt, label %done
 
@@ -449,10 +486,10 @@ done:
 }}
 """ # noqa: E501
 
-ir_numba_atomic_dec = """
-define internal {T} @___numba_atomic_{Tu}_dec({T}* %ptr, {T} %val) alwaysinline {{
+ir_numba_atomic_dec_template = """
+define internal {T} @___numba_atomic_{Tu}_dec({T}* %iptr, {T} %val) alwaysinline {{
 entry:
-    %old2 = load volatile {T}, {T}* %ptr
+    %old2 = load volatile {T}, {T}* %iptr
     br label %attempt
 
 attempt:
@@ -460,7 +497,7 @@ attempt:
     %dec = add {T} %old, -1
     %bndchk = icmp ult {T} %dec, %val
     %new = select i1 %bndchk, {T} %dec, {T} %val
-    %cas = cmpxchg volatile {T}* %ptr, {T} %old, {T} %new monotonic
+    {CAS}
     %repeat = icmp ne {T} %cas, %old
     br i1 %repeat, label %attempt, label %done
 
@@ -469,7 +506,7 @@ done:
 }}
 """ # noqa: E501
 
-ir_numba_atomic_minmax = """
+ir_numba_atomic_minmax_template = """
 define internal {T} @___numba_atomic_{T}_{NAN}{FUNC}({T}* %ptr, {T} %val) alwaysinline {{
 entry:
     %ptrval = load volatile {T}, {T}* %ptr
@@ -488,10 +525,10 @@ lt_check:
 
 attempt:
     ; Attempt to swap in the value
-    %iold = bitcast {T} %dold to {Ti}
+    %old = bitcast {T} %dold to {Ti}
     %iptr = bitcast {T}* %ptr to {Ti}*
-    %ival = bitcast {T} %val to {Ti}
-    %cas = cmpxchg volatile {Ti}* %iptr, {Ti} %iold, {Ti} %ival monotonic
+    %new = bitcast {T} %val to {Ti}
+    {CAS}
     %dcas = bitcast {Ti} %cas to {T}
     br label %lt_check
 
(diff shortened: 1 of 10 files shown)


gmarkall:
I will look into the CI failures first thing tomorrow.

gmarkall:
Build now fixed - I'd accidentally introduced an import-time dependency on libnvvm.so, which is a problem for systems without the CUDA toolkit installed.

esc:
BFID: `numba_smoketest_cuda_yaml_3`

numba/cuda/cudadrv/nvvm.py
```diff
@@ -127,13 +132,34 @@ def __new__(cls):
...
+    def __init__(self):
+        ir_versions = self.get_ir_version()
+        self._majorIR = ir_versions[0]
+        self._minorIR = ir_versions[1]
+        self._majorDbg = ir_versions[2]
+        self._minorDbg = ir_versions[3]
+
+    @property
+    def is_nvvm70(self):
+        return (self._majorIR, self._minorIR) >= (1, 6)
```
stuartarchibald, line 144: Note: https://docs.nvidia.com/cuda/libnvvm-api/group__compilation.html#group__compilation_1g0c22d2b9be033c165bc37b16f3ed75c6
Says that a buffer should contain NVVM IR version 1.6 in a LLVM 7.x representation (text or bitcode), hence where this comes from.

numba/cuda/cudadrv/nvvm.py
```diff
@@ -515,60 +579,57 @@ def _replace_datalayout(llvmir):
...
         ('declare float @___numba_atomic_float_nanmin(float*, float)',
-         ir_numba_atomic_minmax.format(T='float', Ti='i32', NAN='nan',
-                                       OP='ugt', PTR_OR_VAL='', FUNC='min')),
+         ir_numba_atomic_minmax(T='float', Ti='i32', NAN='nan', OP='ugt',
+                                PTR_OR_VAL='', FUNC='min')),
         ('declare double @___numba_atomic_double_nanmin(double*, double)',
-         ir_numba_atomic_minmax.format(T='double', Ti='i64', NAN='nan',
-                                       OP='ugt', PTR_OR_VAL='', FUNC='min')),
+         ir_numba_atomic_minmax(T='double', Ti='i64', NAN='nan', OP='ugt',
+                                PTR_OR_VAL='', FUNC='min')),
```
stuartarchibald, line 616: this is just removing `.format` calls and dropping CAS, it's dealt with later.

esc:
This worked fine, but lack of a `condatoolkit` package version 11.2 meant I had to use 11.0. But, from what I understand that is insufficient because something about LLVM changed between 11.1 and 11.2. Currently the nvidia channel only has 11.2 for linux-64.

numba/cuda/cudadrv/nvvm.py
```diff
@@ -650,10 +714,51 @@ def patch_ptx_debug_pubnames(ptx):
...
+            #
+            # becomes:
+            #
+            #     isDefinition: true, isOptimized: true
+            m = re_spflags.search(line)
+            flags = m.group(1).split(' | ')
+            new_flags = ", ".join([ '%s: true' % spflagmap[f] for f in flags ])
+            start_of_line = line[:m.span()[0]]
+            end_of_line = line[m.span()[1] - 1:]
+            line = start_of_line + new_flags + end_of_line
```
stuartarchibald, line 752: Does this cover all cases of `DISPFlags` that might appear?

numba/cuda/cudadrv/nvvm.py
```diff
@@ -798,7 +903,12 @@ def set_cuda_kernel(lfunc):
 
     # set nvvm ir version
     i32 = ir.IntType(32)
-    md_ver = m.add_metadata([i32(1), i32(2), i32(2), i32(0)])
+    if NVVM().is_nvvm70:
+        ir_versions = [i32(1), i32(6), i32(3), i32(0)]
```
stuartarchibald, line 907: Note: this is NVVM IR 1.6 and Dwarf? 3.0

numba/cuda/tests/cudapy/test_constmem.py
```diff
@@ -134,7 +134,7 @@ def test_const_array_3d(self):
         self.assertTrue(np.all(A == CONST3D))
 
         if not ENABLE_CUDASIM:
-            if cuda.runtime.get_version() in ((8, 0), (9, 0), (9, 1)):
+            if cuda.runtime.get_version() in ((8, 0), (9, 0), (9, 1), (11, 2)):
```
stuartarchibald, line 137: What about the versions in between, e.g. 10? Are they all covered by `else`? Please remind me why :) (and then put in a note) thanks!

gmarkall:
> This worked fine, but lack of a `condatoolkit` package version 11.2 meant I had to use 11.0. But, from what I understand that is insufficient because something about LLVM changed between 11.1 and 11.2. Currently the nvidia channel only has 11.2 for linux-64.

This ought to work with both 11.0 and 11.2 (and everything between 9.0 and 11.1), so at least we've demonstrated that it is still OK with the old NVVM - will discuss a cudatoolkit 11.2 package out-of-band.

stuartarchibald (commented):
Thanks for the patch, I've taken a look through the code, just a few minor things to look at else good. Am hoping CI will catch any significant problems. Thanks again!

numba/cuda/tests/cudapy/test_constmem.py
gmarkall: CUDA 9.2 - 11.1 uses two `f32` loads to load the complex, and the other versions use a load of one `2x f32` vector to load the complex. I don't know why though!

numba/cuda/cudadrv/nvvm.py
gmarkall: Will add a comment, it would certainly help understandability.

numba/cuda/tests/cudapy/test_constmem.py
stuartarchibald: ah yes, that was it :) thanks! Perhaps this could be put in as a note?

gmarkall: (will add a comment)

numba/cuda/cudadrv/nvvm.py
gmarkall: These are all that I've encountered during testing. Since there doesn't appear to be any explicit enumeration / explanation of these flags in the documentation, and it's not necessarily clear what the mapping for all the ones I haven't seen should be, I'm inclined to leave this as-is and add support for others if/when they are encountered (e.g. when changing the LLVM version used by llvmlite, or when enhancing debugging support in Numba / llvmlite).

numba/cuda/cudadrv/nvvm.py
gmarkall: Further noting that the `.format` calls are dropped because these were previously strings, and are now functions, that accept the same parameters as `.format` did.

numba/cuda/cudadrv/nvvm.py
```diff
@@ -127,13 +132,34 @@ def __new__(cls):
...
+        return (self._majorIR, self._minorIR) >= (1, 6)
+
     def get_version(self):
         major = c_int()
         minor = c_int()
         err = self.nvvmVersion(byref(major), byref(minor))
         self.check_error(err, 'Failed to get version.')
         return major.value, minor.value
 
+    def get_ir_version(self):
```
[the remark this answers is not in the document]
gmarkall: If I'd started writing the `NVVM` class from scratch I'd probably do that, but I went for consistency with `get_version()` instead.

numba/cuda/cudadrv/nvvm.py
gmarkall: Will add a note in the code.

gmarkall:
@stuartarchibald Many thanks for the review - I've added comments to the code in the places suggested.

numba/cuda/cudadrv/nvvm.py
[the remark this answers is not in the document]
stuartarchibald: ok, let's just leave it.

numba/cuda/cudadrv/nvvm.py
stuartarchibald: thanks

numba/cuda/tests/cudapy/test_constmem.py
stuartarchibald: thanks!

numba/cuda/cudadrv/nvvm.py
stuartarchibald: thanks!

numba/cuda/cudadrv/nvvm.py
stuartarchibald: ok, this seems reasonable, will deal with on a case by case basis as they occur as we can't easily preempt it.

numba/cuda/cudadrv/nvvm.py
```diff
@@ -901,11 +904,13 @@ def set_cuda_kernel(lfunc):
     nmd = m.get_or_insert_named_metadata('nvvm.annotations')
     nmd.add(md)
 
-    # set nvvm ir version
+    # Set NVVM IR version
     i32 = ir.IntType(32)
     if NVVM().is_nvvm70:
+        # NNVM IR 1.6, DWARF 3.0
```
stuartarchibald, line 910:
```suggestion
        # NVVM IR 1.6, DWARF 3.0
```

numba/cuda/cudadrv/nvvm.py
```diff
@@ -901,11 +904,13 @@ def set_cuda_kernel(lfunc):
...
         ir_versions = [i32(1), i32(6), i32(3), i32(0)]
     else:
+        # NNVM IR 1.1, DWARF 2.0
```
stuartarchibald, line 913:
```suggestion
        # NVVM IR 1.1, DWARF 2.0
```

numba/cuda/cudadrv/nvvm.py
```diff
@@ -141,6 +141,9 @@ def __init__(self):
 
     @property
     def is_nvvm70(self):
+        # NVVM70 uses IR version 1.6. See the documentation for
```
stuartarchibald, line 144:
```suggestion
        # NVVM70 uses NVVM IR version 1.6. See the documentation for
```

gmarkall:
@stuartarchibald Suggestions applied! :-)

stuartarchibald:
> @stuartarchibald Suggestions applied! :-)

thanks!

stuartarchibald (approved):
Thanks for the patch. @sklam did you want to take a look at this too? Also, this needs a buildfarm run with CUDA 11.2.

sklam (dismissed):
Looks good. I've also tested it locally on my mac.

numba/cuda/tests/cudapy/test_constmem.py
```diff
@@ -186,9 +186,15 @@ def test_const_record_optimization(self):
...
-                          'load record fields as vector of 2x u64')
+            if windows:
+                # Two ld.const.u32 as above, but using a bit-field insert to
+                # combine them
+                self.assertIn('ld.const.u32', jcuconst.ptx,
+                              'load record fields as u32')
+            else:
+                # Load of the x and y fields fused into a single instruction
+                self.assertIn('ld.const.v2.u64', jcuconst.ptx,
+                              'load record fields as vector of 2x u64')
```
stuartarchibald, line 197: Are the `old_runtime` and `nvvm70_runtime` branches now effectively the same?

gmarkall: Looks like it - the generated code is different between the old and nvvm70 runtimes, but the check we use for both is the same (because we don't then look for the shift/or vs bfi instructions). I could merge these two branches?

gmarkall: Whoops, they're not quite the same - on non-Windows, `ld.const.v2.f64` is used for the old runtime, whereas `ld.const.v2.u64` is used with the NVVM70 one (`f64` vs `u64`).

esc:
`numba_smoketest_cuda_yaml_6` passed on `568a48b3772055bd2572d516b39bfe8c24b7a93e` -- let me know if it needs a re-run.

numba/cuda/tests/cudapy/test_constmem.py
stuartarchibald: ah, thanks, missed the f64 vs u64

stuartarchibald (approved):
Thanks for the fixes.

stuartarchibald:
> `numba_smoketest_cuda_yaml_6` passed on `568a48b3772055bd2572d516b39bfe8c24b7a93e` -- let me know if it needs a re-run.

Thanks for running this @esc, patch 568a48b looks good.

numba/cuda/tests/cudapy/test_constmem.py
gmarkall: It's easy to miss! I only saw it when I collapsed the two branches and saw it start to fail :-)

PR #6663, opened 2021-01-26
Fix error caused by missing "-static" libraries defined for some platforms

sklam:
The problem exist on mainline and it affect OSX (and probably BSD) given that no "-static" key is defined in the `DLLNAMEMAP`

### DIFF
--- a/numba/misc/findlib.py
+++ b/numba/misc/findlib.py
@@ -35,6 +35,10 @@ def find_lib(libname, libdir=None, platform=None, static=False):
     platform = 'bsd' if 'bsd' in platform else platform
     if static:
         platform = f"{platform}-static"
+    if platform not in DLLNAMEMAP:
+        # Return empty list if platform name is undefined.
+        # Not all platforms define their static library paths.
+        return []
     pat = DLLNAMEMAP[platform] % {"name": libname, "ver": RE_VER}
     regex = re.compile(pat)
     return find_file(regex, libdir)


gmarkall (approved):
This looks good to me - tests run fine for me locally, after merging in master, with both NVVM70 (CUDA >= 11.2) and NVVM34 (CUDA < 11.2)

I can't test on OSX, but I imagine you have done :-) - it's also worth noting that https://github.com/numba/numba/pull/6666 deprecates support for CUDA on OS X, as we're not regularly testing it and don't expect to in future.

> The problem exist on mainline and it affect OSX (and probably BSD)

Although it might affect BSD, I think it's only CUDA that uses the `-static` suffix, and CUDA is not supported on BSD, so I guess this should not be too much of an issue there.

sklam:
Yes, I have tested this on OSX. The main impact of bug for me is that sometimes the CPU testsuite would fail due to the missing "static". The deprecation of OSX is okay as long as it doesn't affect using numba on OSX with a CUDA card.

gmarkall:
>  The deprecation of OSX is okay as long as it doesn't affect using numba on OSX with a CUDA card.

Should that be "without"?

esc:
`numba_smoketest_cuda_yaml_8` passed.

stuartarchibald:
> `numba_smoketest_cuda_yaml_8` passed.

Thanks @esc

stuartarchibald (approved):
Thanks for the patch.

PR #6666, opened 2021-01-27
CUDA: Add a function to query whether the runtime version is supported.

### LINKED ISSUE #6645: Retrieve The Oldest and Newest Supported CUDA Versions
## Feature request

The STUMPY Python package depends on `numba` and, when `cuda` is available, it imports submodules that rely on cuda. It checks this by doing something like:

```
from numba import cuda

if cuda.is_available():  # Only import if CUDA is available
    from stumpy import gpu_stump
```

However, `numba` does not currently support CUDA 11.2 and if CUDA 11.2 is installed then the above `if` condition still returns `True`. It would be nice to have function that checks if the available CUDA version is compatible with `numba` like:

```
if cuda.is_available() and cuda.version_is_supported():
    from stumpy import gpu_stump
```

Or, as discussed with @stuartarchibald, maybe it makes more sense to add max/min supported versions enumerated and then make that part of the `cuda.is_available()` check.

gmarkall:
`numba.cuda.is_supported_version()` returns `True` if the CUDA Runtime is a supported version (as suggested in #6645). The list of supported versions matches those with which the current version of Numba is tested.

Unsupported versions (e.g. newer versions than those known to Numba) may still work. If the current version is unsupported, the caller can decide how to act. Options include:

- Continuing silently,
- Emitting a warning,
- Generating an error or otherwise preventing the use of CUDA.

Notes:

- Based on #6661, which adds support for CUDA 11.2
- Also removes a reference to an unsupported version (8.0) in `test_constmem`.
- Fixes #6645.

### DIFF
--- a/numba/cuda/cudadrv/runtime.py
+++ b/numba/cuda/cudadrv/runtime.py
@@ -118,10 +118,15 @@ class Runtime:
         Returns True if the CUDA Runtime is a supported version.
         """
 
-        supported_versions = ((9, 0), (9, 1), (9, 2),
-                              (10, 0), (10, 1), (10, 2),
-                              (11, 0), (11, 1), (11, 2))
-        return self.get_version() in supported_versions
+        return self.get_version() in self.supported_versions
+
+    @property
+    def supported_versions(self):
+        """A tuple of supported CUDA toolkit versions. Versions are given in
+        the form ``(major_version, minor_version)``."""
+        return ((9, 0), (9, 1), (9, 2),
+                (10, 0), (10, 1), (10, 2),
+                (11, 0), (11, 1), (11, 2))
 
 
 runtime = Runtime()
--- a/numba/cuda/cudadrv/runtime.py
+++ b/numba/cuda/cudadrv/runtime.py
@@ -7,6 +7,7 @@ really used in Numba except for querying the Runtime version.
 
 import ctypes
 import functools
+import sys
 
 from numba.core import config
 from numba.cuda.cudadrv.driver import ERROR_MAP, make_logger
@@ -124,6 +125,9 @@ class Runtime:
     def supported_versions(self):
         """A tuple of supported CUDA toolkit versions. Versions are given in
         the form ``(major_version, minor_version)``."""
+        if sys.platform not in ('linux', 'win32') or config.MACHINE_BITS != 64:
+            # Only 64-bit Linux and Windows are supported
+            return ()
         return ((9, 0), (9, 1), (9, 2),
                 (10, 0), (10, 1), (10, 2),
                 (11, 0), (11, 1), (11, 2))
--- a/numba/cuda/cudadrv/runtime.py
+++ b/numba/cuda/cudadrv/runtime.py
@@ -123,7 +123,7 @@ class Runtime:
 
     @property
     def supported_versions(self):
-        """A tuple of supported CUDA toolkit versions. Versions are given in
+        """A tuple of all supported CUDA toolkit versions. Versions are given in
         the form ``(major_version, minor_version)``."""
         if sys.platform not in ('linux', 'win32') or config.MACHINE_BITS != 64:
             # Only 64-bit Linux and Windows are supported
--- a/numba/cuda/cudadrv/runtime.py
+++ b/numba/cuda/cudadrv/runtime.py
@@ -113,5 +113,15 @@ class Runtime:
         minor = (rtver.value - (major * 1000)) // 10
         return (major, minor)
 
+    def is_supported_version(self):
+        """
+        Returns True if the CUDA Runtime is a supported version.
+        """
+
+        supported_versions = ((9, 0), (9, 1), (9, 2),
+                              (10, 0), (10, 1), (10, 2),
+                              (11, 0), (11, 1), (11, 2))
+        return self.get_version() in supported_versions
+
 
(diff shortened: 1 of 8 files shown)


esc:
@gmarkall thanks for submitting this. Do you happen to have any hints for the reviewers or thoughts on how to test this?

gmarkall:
@esc The tests added in https://github.com/numba/numba/pull/6666/commits/0dc25af223267fd7df211b92d4eb92f9c9ef122e#diff-987a733bf5dd1957203966a284af116015fa06c061df79faebeaa100cc6c03c3 test it by mocking various supported and unsupported versions, so I think running the testsuite is sufficient.

gmarkall:
Following on from the comment in https://github.com/numba/numba/issues/6645#issuecomment-768474962, I've added a `supported_versions` property to the `Runtime` as well.

gmarkall:
@stuartarchibald Master merged in, and the diff now looks a bit more sane.

gmarkall:
@stuartarchibald Following our discussion I've now made it explicit that CUDA is only officially supported on 64-bit Linux and Windows. There's also a small fix in the docs (somewhere it said CC 2.0 GPUs were supported), and on unsupported platforms the list of supported versions is now empty (so as to not give the impression of support).

I decided against producing a warning on unsupported platforms, given that we're close to the 0.53RC and ensuring that it only appears once and in the right circumstances could be a little fiddly, and that there are likely to be very few CUDA OS X users, and of those that do exist the expectation is likely to be that OS X support will not continue forever, because new CUDA toolkit versions don't support it.

This will probably want an entry in the release notes stating that CUDA support on OS X is deprecated.

numba/cuda/device_init.py
```diff
@@ -51,6 +51,23 @@ def is_available():
     return driver_is_available and nvvm.is_available()
 
 
+def is_supported_version():
+    """Returns true if the CUDA Runtime is a supported version.
```
stuartarchibald, line 55:
```suggestion
    """Returns True if the CUDA Runtime is a supported version.
```

numba/cuda/tests/cudadrv/test_runtime.py
```diff
@@ -17,14 +18,33 @@ def set_visible_devices_and_check(q):
...
+    def test_is_supported_version_true(self):
+        for v in SUPPORTED_VERSIONS:
+            with patch.object(runtime, 'get_version', return_value=v):
+                self.assertTrue(runtime.is_supported_version())
+
+    @skip_on_cudasim('The simulator always simulates a supported runtime')
+    def test_is_supported_version_false(self):
+        # Check with an old unsupported version and some potential future
+        # versions
+        for v in ((8, 0), (11, 3), (12, 0)):
```
stuartarchibald, line 42: this is useful to have in a test as it'll pick up newer CTK releases.

numba/cuda/tests/cudadrv/test_runtime.py
```diff
@@ -17,14 +18,33 @@ def set_visible_devices_and_check(q):
         q.put(-1)
 
 
+if config.ENABLE_CUDASIM:
+    SUPPORTED_VERSIONS = (-1, -1),
+else:
+    SUPPORTED_VERSIONS = ((9, 0), (9, 1), (9, 2),
+                          (10, 0), (10, 1), (10, 2),
+                          (11, 0), (11, 1), (11, 2))
+
```
stuartarchibald, line 27: Is it perhaps worth testing the `runtime.supported_versions` property exactly matches this?

numba/cuda/cudadrv/runtime.py
```diff
@@ -113,5 +114,23 @@ def get_version(self):
...
+    def is_supported_version(self):
+        """
+        Returns True if the CUDA Runtime is a supported version.
+        """
+
+        return self.get_version() in self.supported_versions
+
+    @property
+    def supported_versions(self):
+        """A tuple of supported CUDA toolkit versions. Versions are given in
```
stuartarchibald, line 126:
```suggestion
        """A tuple of all supported CUDA toolkit versions. Versions are given in
```
added `all` to help imply that numerous versions are going to be returned.

stuartarchibald (commented):
Thanks for the patch, great to see this implemented, should help out those with CPU/CUDA versions of their libraries in deciding whether to load the CUDA version.

numba/cuda/tests/cudadrv/test_runtime.py
gmarkall: I felt it wasn't, given what the other tests do.

gmarkall:
@stuartarchibald Many thanks for the review - suggestions applied, and a response to one comment.

numba/cuda/tests/cudadrv/test_runtime.py
stuartarchibald: I suppose the other tests are checking both for and against existence in this set of values so it's probably fine :)

stuartarchibald (approved):
Thanks for the fixes, patch looks good.

sklam:
Manually verified on macbook w/ cuda ✔️

esc:
`numba_smoketest_cuda_yaml_7` passed on build farm.

stuartarchibald:
Thanks for running it @esc, thanks for checking on OSX @sklam, think this is good to merge.

PR #6671, opened 2021-01-29
Update environment creation example to python 3.8

luk-f-a:
as titled. motivated by discussion that correct python typing annotations require py 3.8


stuartarchibald (approved):
Thanks for updating this.

PR #6676, opened 2021-02-02
Update hardware and software versions in various docs.

stuartarchibald:
Updates:
 * Hardware to note M1/Arm64 is known to work.
 * Python 3.9 now supported.
 * Unofficial BSD support.
 * Armv7l and Aarch64 not experimental any more.


esc (approved):
Awesome, thanks!

Issue #6668, opened 2021-01-28
Returning complex types from objmode fails in some cases


kozlov-alexey:
- [x] I have tried using the latest released version of Numba (most recent is
 visible in the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [x] I have included a self contained code sample to reproduce the problem.
  i.e. it's possible to run as 'python bug.py'.

Hi,

I tested #4769 and found some issues with it. Please take a look at below reproducers.

Use-case 1: works OK with ret type provided as global variable, but doesn't work with concrete types
```
import numba
from numba import objmode, typeof, njit
from numba.core import types

'''
    raise errors.CompilerError(
    numba.core.errors.CompilerError: Failed in nopython mode pipeline (step: Handle with contexts)
    objectmode context requires constants string for type annotation
'''
@njit
def test1():
    with objmode(res=types.int32):      # using concrete type is not permitted
        res = 32                        # raised error still refers to old requirement
    return res
 
result = test1()
print(result)
```

Use-case 2: returning python list from objmode.
```
'''
    Gives:
    ret_type=reflected list(int64)<iv=None>
    [<NULL>, 2, 3, 4, 5]
'''
ret_type = typeof([1, 2, 3, 4, 5])      # NOK (garbage returned)
print(f"ret_type={ret_type}")
@njit
def test2():
    with objmode(res=ret_type):
        res = [1, 2, 3, 4, 5]
    return res
 
result = test2()
print(result)
```

Use-case 3: returning python set from objmode.
```
'''
    Gives:
    File "C:\Users\akozlov\AppData\Local\Continuum\anaconda3\envs\NUMBA_MASTER\lib\site-packages\llvmlite\binding\module.py", line 25, in parse_assembly
        raise RuntimeError("LLVM IR parsing error\n{0}".format(errmsg))
    RuntimeError: Failed in nopython mode pipeline (step: Handle with contexts)
    Failed in nopython mode pipeline (step: nopython mode backend)
    LLVM IR parsing error
    <string>:410:7: error: value doesn't match function result type 'i32'
      ret i8* null
          ^
'''
ret_type = typeof({1, 2, 3, 4, 5})      # NOK (LLVM parsing error)
print(f"ret_type={ret_type}")
@njit
def test3():
    with objmode(res=ret_type):
        res = {1, 2, 3, 4, 5}
    return res

result = test3()
print(result)
```

Use-case 4: returning python dict from objmode.
```
[...]

PR #6680, opened 2021-02-02
Allow getattr to work in objmode output type spec

sklam:
Related to https://github.com/numba/numba/issues/6668

### DIFF
--- a/numba/core/withcontexts.py
+++ b/numba/core/withcontexts.py
@@ -203,6 +203,13 @@ class _ObjModeContextType(WithContext):
                 "objectmode context doesn't take any positional arguments",
                 )
         typeanns = {}
+
+        def report_error(varname, msg, loc):
+            raise errors.CompilerError(
+                    f"Error handling objmode argument {varname!r}. {msg}",
+                    loc=loc,
+                )
+
         for k, v in kwargs.items():
             if isinstance(v, ir.Const) and isinstance(v.value, str):
                 typeanns[k] = sigutils._parse_signature_string(v.value)
@@ -210,16 +217,20 @@ class _ObjModeContextType(WithContext):
                 try:
                     v = func_closures[v.name]
                 except KeyError:
-                    raise errors.CompilerError(
-                        f"Freevar {v.name!r} is not defined"
+                    report_error(
+                        varname=k,
+                        msg=f"Freevar {v.name!r} is not defined.",
+                        loc=loc,
                     )
                 typeanns[k] = v
             elif isinstance(v, ir.Global):
                 try:
                     v = func_globals[v.name]
                 except KeyError:
-                    raise errors.CompilerError(
-                        f"Global {v.name!r} is not defined"
+                    report_error(
+                        varname=k,
+                        msg=f"Global {v.name!r} is not defined.",
+                        loc=loc,
                     )
                 typeanns[k] = v
             elif isinstance(v, ir.Expr) and v.op == "getattr":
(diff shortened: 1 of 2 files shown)


numba/tests/test_withlifting.py
```diff
@@ -825,6 +825,37 @@ def global_var():
...
+    def test_objmode_gv_mod_attr_error(self):
+        @njit
+        def moderror():
+            with objmode(val=types.THIS_DOES_NOT_EXIST):
+                val = 12.3
+            return val
+        with self.assertRaises(errors.CompilerError) as raises:
+            moderror()
+        self.assertIn("Getattr cannot be resolved at compile-time",
+                      str(raises.exception))
```
gmarkall, line 857: Is it worth checking that the reported getattr that can't be resolved is as expected, e.g.:

```suggestion
                      str(raises.exception))
        self.assertIn("types.THIS_DOES_NOT_EXIST):\n            ^",
                      str(raises.exception))
```

gmarkall (commented):
The changes look good to me - there is one small suggestion for the error message check on the diff.

I presume this relates only to use case 1 in #6668?

numba/tests/test_withlifting.py
sklam: The suggested check depends on the error reporting options; e.g. i'm seeing colorama added codes and the suggested check is failing. I don't think it's worth writing all the regex just to match it correctly given that the current error message can only be coming from the code I added.

sklam:
@gmarkall, thx for reviewing. And, yes, this is the use-case 1 for #6668.

numba/tests/test_withlifting.py
gmarkall: OK, makes sense.

gmarkall (approved):
Looks good to me.

gmarkall:
Not sure if @esc @stuartarchibald also want to look at this?

gmarkall:
Hit the wrong button, sorry.

numba/tests/test_withlifting.py
stuartarchibald: I bumped into this whilst testing:
```python
from numba import njit, objmode
import numpy as np

@njit
def moderror(pred):
    if pred:
        x = 'intp'
    else:
        x = 'float32'
    with objmode(val=x):
        val = 12.3
    return val

moderror(False)
```
not necessarily to be fixed here, but https://github.com/numba/numba/blob/d89d34105cc0e1dc9402b14bbc0e4a7abf3c0a3e/numba/core/transforms.py#L382 should guard for multiply defined variables.

numba/core/withcontexts.py
```diff
@@ -220,12 +222,22 @@ def _legalize_args(self, args, kwargs, loc, func_globals, func_closures):
                         f"Global {v.name!r} is not defined"
                     )
                 typeanns[k] = v
+            elif isinstance(v, ir.Expr) and v.op == "getattr":
+                try:
+                    base_obj = func_ir.infer_constant(v.value)
+                    typ = getattr(base_obj, v.attr)
+                except (errors.ConstantInferenceError, AttributeError):
+                    raise errors.CompilerError(
+                        f"Getattr cannot be resolved at compile-time",
```
stuartarchibald, line 231: Should this include information from `base_obj` and `v.value`? From e.g. this:
```python
from numba import njit, objmode, types
import numpy as np

@njit
def moderror():
    with objmode(v1=types.NOT_A_TYPE, v2=types.intp, v3=types.float32):
        v1 = 12.3
        v2 = 12.3
        v3 = 12.3
    return v1, v2, v3

moderror()
```
gives:
```
numba.core.errors.CompilerError: Failed in nopython mode pipeline (step: Handle with contexts)
Getattr cannot be resolved at compile-time

File "test_6680.py", line 6:
def moderror():
    with objmode(v1=types.NOT_A_TYPE, v2=types.intp, v3=types.float32):
    ^
```
whilst entirely obvious in this contrived case, in general would it be clear which of the `v{1,2,3}` is the issue?

sklam: I've made the msg better. The above will now show:

```
CompilerError: Failed in nopython mode pipeline (step: Handle with contexts)
Error handling objmode argument 'v1'. Getattr cannot be resolved at compile-time.

File "<ipython-input-1-a794355d32dc>", line 6:
def moderror():
    with objmode(v1=types.NOT_A_TYPE, v2=types.intp, v3=types.float32):
    ^
```

stuartarchibald: Thanks for changing that, the latest patch makes this case much clearer.

stuartarchibald (approved):

PR #6681, opened 2021-02-03
Explicitly catch `with..as` and raise error.

### LINKED ISSUE #6679: with.. as construct breaks bytecode analysis and leads to an uncontrolled failure
numba.core.errors.UnsupportedError: Failed in nopython mode pipeline (step: analyzing bytecode)
'try' block not supported until python3.7 or later

stuartarchibald:
As title. Makes the failure reason very explicit opposed to just
a general "bytecode cannot be analysed".

Closes #6679

### DIFF
--- a/numba/core/controlflow.py
+++ b/numba/core/controlflow.py
@@ -847,6 +847,7 @@ class ControlFlowAnalysis(object):
     def _iter_inst(self):
         for inst in self.bytecode:
             if self._use_new_block(inst):
+                self._guard_with_as(inst)
                 self._start_new_block(inst)
             self._curblock.body.append(inst.offset)
             yield inst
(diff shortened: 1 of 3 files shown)


numba/core/controlflow.py
```diff
@@ -847,6 +847,7 @@ def jump(self, target, pops=0):
     def _iter_inst(self):
         for inst in self.bytecode:
             if self._use_new_block(inst):
+                self._guard_with_as(inst)
```
gmarkall, line 850: The included test still passes if this line and the implementation of `_guard_with_as` are removed - is this catching a different, untested case? (I note the code here is also exactly the same as in byteflow.py).

stuartarchibald: There's two versions of the "interpreter", https://github.com/numba/numba/commit/d36ae2af1200ab56b36e57589843974727e73196 failed https://github.com/numba/numba/runs/1821883328 as Python 3.6 goes a different route to later pythons.

gmarkall: Ah, I wasn't aware - thanks for clarifying!

gmarkall (approved):
Looks good to me!

stuartarchibald:
Thanks for reviewing @gmarkall. Please could one of you take a look @sklam @esc. Thanks.

sklam (approved):
LGTM

PR #6682, opened 2021-02-03
Update deprecation notices for 0.53

stuartarchibald:
As title.

### DIFF
--- a/docs/source/reference/deprecation.rst
+++ b/docs/source/reference/deprecation.rst
@@ -100,8 +100,8 @@ Schedule
 This feature will be removed with respect to this schedule:
 
 * Pending-deprecation warnings will be issued in version 0.44.0
-* Deprecation warnings and replacements will be issued in version 0.54.0
-* Support will be removed in version 0.55.0
+* Prominent notice will be given for a minimum of two releases prior to full
+  removal.
 
 Recommendations
 ---------------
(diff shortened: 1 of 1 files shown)


docs/source/reference/deprecation.rst
```diff
@@ -100,8 +100,8 @@ Schedule
 This feature will be removed with respect to this schedule:
 
 * Pending-deprecation warnings will be issued in version 0.44.0
-* Deprecation warnings and replacements will be issued in version 0.53.0
-* Support will be removed in version 0.54.0
+* Deprecation warnings and replacements will be issued in version 0.54.0
+* Support will be removed in version 0.55.0
```
esc, line 104: Do we really believe this will happen. Or should we perhaps formulate this differently to reflect the fact that we actually don't know when we can deprecate it?

esc (changes requested):
Just one comment regarding expectation management.

docs/source/reference/deprecation.rst
stuartarchibald: I think it's reasonable to bump it or to express "some time in the future with X releases notice", perhaps I'll add the latter (same for objmode fall back) as removal depends on confidence of stability and having the resources to do it etc.

esc: I agree adding something to indicate that this is still a lot of work would be good.

sklam (approved):
LGTM

esc:
LGTMT

esc (approved):

PR #6685, opened 2021-02-04
Add Python 3.9 to classifiers.

stuartarchibald:
As title.


sklam (approved):
Thanks for the patch!

PR #6694, opened 2021-02-05
WIP Fix #6690 'Bad token in signature' with @guvectorize where bad token is NUMBER

### LINKED ISSUE #6690: 'Bad token in signature' with @guvectorize
- [x] I have tried using the version 0.51.2 of Numba
- [x] I have included a self contained code sample to reproduce the problem

Hi,
I am sorry, I am aware it is more a 'usage' question or perhaps a 'feature request'.
I have posted the question on [SO here](https://stackoverflow.com/questions/66052723/bad-token-in-signature-with-numba-guvectorize), but it does not meet much success.

Please, can we setup a constant number of columns for a result array in function signature?
I have tried this way, but get here below error message.

```python
import numpy as np
from numba import guvectorize

@guvectorize('void(float64[:], float64, uint32[:,2])', '(m),()->(m,2)', nopython=True)
def array_copy(data, delta, result):
    for i, val in np.ndenumerate(data):
        # Open i.
        i, = i
        result[i,:] = [i, int(val)+delta]

data = np.arange(3, dtype='float64')
res = np.zeros((data.shape[0], 2), dtype='uint32')
array_copy(data, 3, res)
```

```python

  File "<ipython-input-17-63f8983bbf61>", line 5, in <module>
    def array_copy(data, delta, result):

  File "/home/pierre/anaconda3/lib/python3.8/site-packages/numba/np/ufunc/decorators.py", line 177, in wrap
    guvec = GUVectorize(func, signature, **kws)

  File "/home/pierre/anaconda3/lib/python3.8/site-packages/numba/np/ufunc/decorators.py", line 49, in __new__
    return imp(func, signature, identity=identity, cache=cache,

  File "/home/pierre/anaconda3/lib/python3.8/site-packages/numba/np/ufunc/ufuncbuilder.py", line 298, in __init__
    self.sin, self.sout = parse_signature(signature)

  File "/home/pierre/anaconda3/lib/python3.8/site-packages/numba/np/ufunc/sigparse.py", line 49, in parse_signature
    outputs = list(parse(outs))

  File "/home/pierre/anaconda3/lib/python3.8/site-packages/numba/np/ufunc/sigparse.py", line 35, in parse
    raise ValueError('bad token in signature "%s"' % tok[1])

ValueError: bad token in signature "2"
```

HPLegion:
Opening for initial test on AzurePipelines

### DIFF
--- a/numba/np/ufunc/sigparse.py
+++ b/numba/np/ufunc/sigparse.py
@@ -27,7 +27,7 @@ def parse_signature(sig):
                     tok = next(tokgen)
                     if tok[1] == ')':
                         break
-                    elif tok[0] == tokenize.NAME:
+                    elif tok[0] == tokenize.NAME or tok[0] == tokenize.NUMBER:
                         symbols.append(tok[1])
                     elif tok[1] == ',':
                         continue


HPLegion:
Thanks for labelling @esc 

This is titled WIP since I am not certain whether the PR is sufficient in its current rudimentary form or if you would expect a more elaborate fix.
I would appreciate some comments on the current state from either you or @stuartarchibald or @sklam .

Cheers! :)

sklam:
It looks like this is sufficient for CPU target where we are letting NumPy handle the looping. For other targets where we are implementing ufunc-like operation ourselves, we will need to implement the fixed dim logic or disable it for those targets.

HPLegion:
Hi :)
Sorry for being quiet on this for a while.
As far as I can tell it works on CPU single threaded and parallel targets (also covered by the tests).

Besides that `vectorize` seems to support the CUDA but not the ROC target.
I doubt that I would be able to implement the CUDA side myself without help, I had a look at it last week and was a bit overwhelmed :-P

So the best I could offer to do by myself at the moment is to block these signatures if the CUDA target is chosen, but that might of course introduce an annoying divergence between CPU and GPU usage.

yohplala:
Hello, I believe this PR is still worth it.

gmarkall:
gpuci run tests

gmarkall:
I've just merged `main` and resolved conflicts so this can have a test on gpuCI, to see if it breaks anything CUDA-related. The ROC target no longer matters, as it's been removed.

gmarkall:
Since CUDA seems to be OK with this, I think it might as well be thought of as ready for review.

Issue #6496, opened 2020-11-17
Wrap around negative indexing for cuda atomic operations


RendersJens:
## Feature request

The CUDA atomic operations currently do not support negative indices as offset from the end of the array. One would expect that

`cuda.atomic.add(ary, idx, val)`
performs
`ary[idx] += val`
Like the docs state, independent of whether idx is negative or positive. This can lead to anoying bugs to track down if you translate sequential code to parallel code.

gmarkall:
I think the documentation is further misleading - for example, might one expect:

```
cuda.atomic.add(ary, slice(2,4), val)
```

to add `val` to elements 2 and 3 of `ary`?

I think this is both a documentation bug and a feature request, so I'll label as such.

ashutoshvarma:
Adding support for negative indices seems pretty easy to me, as `cgutils.get_item_pointer` already support wraparound.
https://github.com/numba/numba/blob/00ad12751725391e8423d67ef845ba56110c55df/numba/cuda/cudaimpl.py#L610
https://github.com/numba/numba/blob/00ad12751725391e8423d67ef845ba56110c55df/numba/core/cgutils.py#L667-L668

gmarkall:
This is now implemented.

PR #6695, opened 2021-02-06
Enable negative indexing for cuda atomic operations

ashutoshvarma:
Ref #6496 

As titled, enables wraparound in CUDA atomic operations.

Instead of adding new tests, I modified existing add atomics tests `test_atomic_add_*` to prevent unnecessary redundancy in test code. 
If the separate test is a must then I can add that too.

### DIFF
--- a/numba/cuda/cudaimpl.py
+++ b/numba/cuda/cudaimpl.py
@@ -607,7 +607,8 @@ def _atomic_dispatcher(dispatch_fn):
                             (aryty.ndim, len(indty)))
 
         lary = context.make_array(aryty)(context, builder, ary)
-        ptr = cgutils.get_item_pointer(context, builder, aryty, lary, indices)
+        ptr = cgutils.get_item_pointer(context, builder, aryty, lary, indices,
+                                       wraparound=True)
         # dispatcher to implementation base on dtype
         return dispatch_fn(context, builder, dtype, ptr, val)
     return imp
--- a/numba/cuda/tests/cudapy/test_atomics.py
+++ b/numba/cuda/tests/cudapy/test_atomics.py
@@ -24,12 +24,15 @@ def atomic_cast_none(num):
 
 @cuda.jit(device=True)
 def atomic_binary_1dim_shared(ary, idx, op2, ary_dtype, ary_nelements,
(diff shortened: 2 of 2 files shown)


gmarkall (changes requested):
Many thanks for the PR! I think this is looking great in general, with just one suggestion / question:

- Tests pass for me locally with hardware (Quadro RTX 8000), the simulator, and under cuda-memcheck with no errors.
- The testing additions look good - I see that testing with negative indexing is only added for a subset of test cases, which I think covers the new functionality sufficiently without bloating the set of tests inordinately (as opposed to adding negative indexing to every single test).
- I'm happy with the additions being made to existing test cases, as it enables reuse of their code instead of either duplicating the rest of the test body or adding another level of indirection in the testing.
- The style of the changes looks good / seems to be in keeping with the surrounding code.


One thing I'd like to see - either I've missed it or it's not present - is testing of wraparound on arrays in global memory. In the test changes, I can only see tests for negative indexing on shared memory - if negative indexing on global memory is tested here, can you indicate where that testing is please? If not, would it be possible to add testing on negative indexing on global memory in a similar fashion to the testing on shared memory?

Many thanks again!

ashutoshvarma:
My bad, seems I missed global memory tests. I will update them in a minute.

gmarkall:
Many thanks for the update!

The test fail is just a conda error - could this have a re-run on Azure Pipelines please? (@esc / @stuartarchibald)

gmarkall (approved):
Many thanks for adding the tests with global memory - they run fine for me locally with hardware, the simulator and cuda-memcheck. This now looks good to me!

@esc could this have a CUDA buildfarm run please?

esc:
@gmarkall seems like AZP fine this time around.

esc:
Running BF: `numba_smoketest_cuda_yaml_11`

esc:
BF passed on this.

gmarkall:
@stuartarchibald @sklam would you like to take a look at this before it goes RTM?

stuartarchibald (approved):
Thanks for the patch @ashutoshvarma looks good, thanks for reviewing @gmarkall

Issue #6563, opened 2020-12-10
CUDA: Add support for passing tuples to kernels via vectorize


cristipurdel:
## Feature request

numba version used: 0.53.0.dev0+357.g4d3d2673c.dirty

I would like to pass tuples to CUDA kernels by using vectorize similar to #6343. Here is the code and subsequent error

```
from numba import cuda, vectorize
import numpy as np
import pkg_resources
# working based on https://github.com/numba/numba/pull/6343

print('numba version used: ' + str(pkg_resources.get_distribution("numba").version))



@cuda.jit
def f(r, x):
    r[0] = x[0][0]
    r[1] = x[1]
    r[2] = x[2][0]

z = np.ones(1, dtype=np.int64)
x = (1 * z, 10, (4, 3))
r = np.zeros(len(x), dtype=np.int64)
# Use the syntax: kernel_function[blockspergrid, threadsperblock](arg0, arg1, ..., argn)
f[1, 1](r, x)
print(r)



def f2(r2, x2): 
    out = 0
    # if the next two lines are commented out, the error is TypeError: argument #2 has invalid type of <class 'tuple'>
    for i in range(len(x2)):
        out += x2[i]
    return out

z2 = np.arange(1024, dtype=np.int64)
list_tuple = []
for n in range(1024):
    list_tuple.append((n, n+1, n+2))
x2 = (1 * z2, list_tuple)
r2 = np.zeros((len(z2),len(x2)), dtype=np.int64)

v_cuda = vectorize(['int64(int64, int64)'], target = 'cuda')(f2)  
out = v_cuda(r2, x2)
```

Error is 

```
  File "D:\Documents\Cristi\Tools\Python\GitHub\numba\6343.py", line 39, in <module>
    v_cuda = vectorize(['int64(int64, int64)'], target = 'cuda')(f2)

  File "D:\Kit\PortableApps\Miniconda_numba_dev\lib\site-packages\numba\np\ufunc\decorators.py", line 120, in wrap
    vec.add(sig)

  File "D:\Kit\PortableApps\Miniconda_numba_dev\lib\site-packages\numba\np\ufunc\deviceufunc.py", line 399, in add
    corefn, return_type = self._compile_core(devfnsig)

  File "D:\Kit\PortableApps\Miniconda_numba_dev\lib\site-packages\numba\cuda\vectorizers.py", line 15, in _compile_core
    cudevfn = cuda.jit(sig, device=True, inline=True)(self.pyfunc)

[...]

PR #6698, opened 2021-02-07
Fix #6697: Wrong dtype when using np.asarray on DeviceNDArray

### LINKED ISSUE #6697: Wrong dtype when using np.asarray on DeviceNDArray
## Reporting a bug

#### Reproduce Code

```python
>>> from numba import cuda
>>> import numpy as np
>>> arr = np.arange(10).astype(np.int16)
>>> arr.dtype
dtype('int16')

>>> darr = cuda.to_device(arr)
>>> darr.dtype
dtype('int16')

>>> arr = np.asarray(darr)
>>> arr.dtype
dtype('float64')
```

_Originally posted by @ashutoshvarma in https://github.com/numba/numba/issues/6563#issuecomment-774035027_

The `test_auto_transfer` is failing due the following change :-
Without it all tests pass without problem
```diff
@@ -107,7 +108,7 @@ class UFuncMechanism(object):
         """
         for i, ary in enumerate(self.arrays):
             if ary is not None:
-                self.argtypes[i] = ary.dtype
+                self.argtypes[i] = np.asarray(ary).dtype
```
Since `ary` would always be either `np.ndarray` or `DeviceNDArray` type,  keeping `ary.dtype` should be fine? (correct me if I missed something )

Ideally, the above change should not break anything, but for some reason, if `ary` is of DeviceNDArray type then `np.asarray(ary).dtype` gives the wrong dtype.
For DeviceNDArray ary, `np.asarray(ary)` will always be a `float64` ndarray irrespective of original dtype

```python
>>> x = np.arange(n, dtype=np.int32)
>>> dx = cuda.to_device(x)
>>> print(dx.dtype)
>>> print(np.asarray(dx).dtype)
int32
float64
```

After digging more, looks like problem might be in `DeviceNDArray.__array__()` method
https://github.com/numba/numba/blob/00ad12751725391e8423d67ef845ba56110c55df/numba/cuda/cudadrv/devicearray.py#L482-L486

In above, passing even `None` (dtype) to `__array__` of ndarray will return new array with default dtype (`float64`)
`ndarray.__array__` docs - https://numpy.org/doc/stable/reference/generated/numpy.ndarray.__array__.html
```python
>>> import numpy as np
>>> a = np.arange(10, dtype=np.int16)   
>>> b = a.__array__(np.int16)          # no copy
>>> c = a.__array__(None)              # copy
>>> d = a.__array__()                  # no copy
>>> print(f"{a.dtype=}")
>>> print(f"{b.dtype=}")
>>> print(f"{c.dtype=}")
>>> print(f"{d.dtype=}")
a.dtype=dtype('int16')
b.dtype=dtype('int16')
c.dtype=dtype('float64')
d.dtype=dtype('int16')
```

Changing `__array__(dtype) ` to either `__array__(self.dtype) ` or `__array__() ` should fix this.

ashutoshvarma:
Closes #6697

I was not sure about where tests for this should go, for now, the test is in `numba/cuda/tests/cudadrv/test_array_attr.py`.

### DIFF
--- a/numba/cuda/cudadrv/devicearray.py
+++ b/numba/cuda/cudadrv/devicearray.py
@@ -483,7 +483,10 @@ class DeviceNDArray(DeviceNDArrayBase):
         """
         :return: an `numpy.ndarray`, so copies to the host.
         """
-        return self.copy_to_host().__array__(dtype)
+        if dtype:
+            return self.copy_to_host().__array__(dtype)
+        else:
+            return self.copy_to_host().__array__()
 
     def __len__(self):
         return self.shape[0]
--- a/numba/cuda/tests/cudadrv/test_array_attr.py
+++ b/numba/cuda/tests/cudadrv/test_array_attr.py
@@ -87,6 +87,12 @@ class TestArrayAttr(CUDATestCase):
         got = dary_reshaped.copy_to_host()
         self.assertPreciseEqual(expect, got)
 
+    def test_bug6697(self):
+        ary = np.arange(10, dtype=np.int16)
+        dary = cuda.to_device(ary)
+        got = np.asarray(dary)
+        self.assertEqual(got.dtype, dary.dtype)
+
 
 if __name__ == '__main__':
     unittest.main()
--- a/numba/cuda/tests/cudadrv/test_array_attr.py
+++ b/numba/cuda/tests/cudadrv/test_array_attr.py
@@ -87,12 +87,6 @@ class TestArrayAttr(CUDATestCase):
         got = dary_reshaped.copy_to_host()
         self.assertPreciseEqual(expect, got)
 
-    def test_bug6697(self):
-        ary = np.arange(10, dtype=np.int16)
-        dary = cuda.to_device(ary)
-        got = np.asarray(dary)
-        self.assertEqual(got.dtype, dary.dtype)
-
 
 if __name__ == '__main__':
     unittest.main()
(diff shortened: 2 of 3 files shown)


gmarkall (changes requested):
Many thanks for tracking down this issue and making the PR! I think it is looking great in general:

- Tests pass for me locally with hardware and the simulator.
- The reproducer given in #6697 behaves correctly with this PR (it prints out `int16` three times instead of `int16` twice then `float64`).
- The starter patch from #6563 applied on top of this PR passes all tests, and also allows the example from https://github.com/numba/numba/issues/6563#issuecomment-769511140 to run correctly.

So, this PR really moves things in the right direction for both the issue #6697 and the feature request #6563.

Regarding the location of the test - the majority of tests that test device arrays, transferring them between the host and the device, and checking their dtypes and other properties are in `numba/cuda/tests/cudadrv/test_cuda_ndarray.py` - would you be happy to move the test there?

Many thanks again!

ashutoshvarma:
I moved the test to numba/cuda/tests/cudadrv/test_cuda_ndarray.py (`TestCudaNDArray`) as you requested.

gmarkall (approved):
Many thanks for the update - this is looking good to me!

@esc Could this have a CUDA buildfarm run please?

esc:
@gmarkall sure `numba_smoketest_cuda_yaml_10`

esc:
Screen Shot 2021-02-09 at 18 19 06

BF passed.

gmarkall:
@stuartarchibald @sklam Do you want to look at this before it goes to RTM?

stuartarchibald (approved):
Looks good, thanks for the patch. Thanks for reviewing @gmarkall .

gmarkall:
Many thanks for the contribution @ashutoshvarma !

PR #6709, opened 2021-02-09
Block matplotlib in test examples

gmarkall:
The mandelbrot doc test example plots the mandelbrot set. This is nice for the purposes of the example, but results in a plot popping up and blocking further tests when the suite is run on a machine with matplotlib installed.

This PR prevents plotting when running the test, by blocking the import of matplotlib during the examples tests.

### DIFF
--- a/numba/tests/doc_examples/test_examples.py
+++ b/numba/tests/doc_examples/test_examples.py
@@ -1,11 +1,32 @@
 # Contents in this file are referenced from the sphinx-generated docs.
 # "magictoken" is used for markers as beginning and ending of example text.
 
+import sys
 import unittest
 from numba.tests.support import captured_stdout
 
 
+class MatplotlibBlocker:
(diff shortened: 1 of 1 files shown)


stuartarchibald (approved):
Thanks for the patch, manually verified.

leofang:
I thought you could just set the backend to `Agg` before importing `matplotlib.pyplot`:
```python
>>> import matplotlib
>>> matplotlib.use('Agg')
>>> from matplotlib import pyplot as plt
```
Then,  a warning is raised when `plt.show()` or alike is called.

gmarkall:
> I thought you could just set the backend to Agg before importing matplotlib.pyplot:

Ah, I wasn't aware of the Agg backend - many thanks for drawing attention to it... Hopefully the merged PR will be OK, but if any issues crop up with this slightly unorthodox setup, I'll try switching it to the Agg backend

PR #6710, opened 2021-02-09
remove hard-coded branch

abebeos:
This allows to use the test-scripts more universal, e.g. within a usual ci setup.

Fixes a blocking bug in a typical travis-ci setup.

### DIFF
--- a/numba/testing/main.py
+++ b/numba/testing/main.py
@@ -386,7 +386,7 @@ def _choose_gitdiff_tests(tests):
         raise ValueError("gitpython needed for git functionality")
     repo = Repo('.')
     path = os.path.join('numba', 'tests')
-    target = 'origin/master..HEAD'
+    target = 'HEAD'
     gdiff_paths = repo.git.diff(target, path, name_only=True).split()
     # normalise the paths as they are unix style from repo.git.diff
     gdiff_paths = [os.path.normpath(x) for x in gdiff_paths]


esc:
@abebeos thank you for submitting this. Unfortunately, I don't think we can merge this patch in this form. The code in question is related to finding a so called `gitdiff_test`. That is, the tests that test the code that was modified since "last time". From what I understand, the expression `origin/master..HEAD` determines which files were modified between `origin/master` and `HEAD`. Simply doing a `git diff HEAD` will have no effect on a clean working directory, so I believe that adding this patch will break existing assumptions about  how tests are run in such a way that no `gitdiff_tests` will be run at all. Does this make sense?

abebeos:
> Does this make sense?

To be honest, don't know exactly. I should have possibly filed an issue first, with the problem at hand:

The scripts fail on a standard travis-ci setup, and the problem is in this location (because "master" is not available when using feature-branches, travis checks out only the feature branch. Happens on other CI's, too, e.g. saw the bug on bitbucket ci)

esc:
@abebeos ok, that does sound a little strange. Is the travis-build public? Could I check it out somewhere?

abebeos:
> @abebeos ok, that does sound a little strange. Is the travis-build public? Could I check it out somewhere?

I made it public:

https://github.com/abebeos/numba-pvt/commit/5399b8f4885f194895cace3fd6e65e6c853c978e

if you reverse the "HEAD" change, then you'll get the error. Note that you cannot use this repo, as the travis-ci is "out of credits".

(Not that I'm not active with this task anymore, forgot nearly everything.)

gmarkall:
Marking this as "in progress" for now whilst we triage the related issue.

esc:
@abebeos do you happen to have an error or stack trace that you could post? I'm not sure how to run your repo on travis.

abebeos:
> @abebeos do you happen to have an error or stack trace that you could post? I'm not sure how to run your repo on travis.

I'll file an issue with a reproduction, as this is more complex than I initially assumed. Just give me a few days.

abebeos:
@esc, here is the bug:

=> took latest numba master, cloned into branch "travis", added travis.yml which ueses the build-scripts to run the tests:

Failure:

https://travis-ci.com/github/abebeos/numba/builds/216686907

(note that I'm most possibly not available the next days, but you should have all info you need now)

abebeos:
> From what I understand, the expression `origin/master..HEAD` determines which files were modified between `origin/master` and `HEAD`.

@esc: between `origin/master HEAD" and "current branch current commit"

fixed for now via workaround found within https://github.com/travis-ci/travis-ci/issues/6069#issuecomment-319710346, thus closing.

PR #6716, opened 2021-02-11
Consider assignment lhs live if used in rhs (Fixes #6715)

### LINKED ISSUE #6715: Regression in 0.53.0rc1 due to issue in temp variable assignment removal
This fails type inference on 0.53.0rc1, worked fine on 0.52.0:
```python
from numba import njit
import numpy as np

@njit
def foo():
    increment = 0.0
    foo_sum = 0.0
    i = 2

    while foo_sum < 100:
        a = [np.random.normal(1) for _ in range(i)]
        a = np.array(a)
        increment = np.sum(a)
        increment = np.abs(increment)
        foo_sum += increment
        i += 1

    return (foo_sum, i)

foo()
```
Debug log:
```
$ cat di58.py
from numba import njit
import numpy as np

@njit
def foo():
    increment = 0.0
    foo_sum = 0.0
    i = 2

    while foo_sum < 100:
        a = [np.random.normal(1) for _ in range(i)]
        a = np.array(a)
        increment = np.sum(a)
        increment = np.abs(increment)
        foo_sum += increment
        i += 1

    return (foo_sum, i)


$ HEAD=`git rev-parse HEAD`

$ git show-ref --tags -d|grep $HEAD
90e7cc74e1c3775727289162eade941c49a0db4f refs/tags/0.53.0rc1.post1^{}

$ python di58.py
  File "di58.py", line 20, in <module>
    foo()
  <snip>
  File "/numba/core/typeinfer.py", line 1174, in check_var
    raise TypingError(msg % (var, val, loc), loc)
Type of variable 'a' cannot be determined, operation: call $42load_method.9(a, func=$42load_method.9, args=[Var(a, di58.py:11)], kws=(), vararg=None), location: di58.py (12)

File "di58.py", line 12:
def foo():
    <source elided>
        a = [np.random.normal(1) for _ in range(i)]
        a = np.array(a)
        ^

$ git revert -m 1 d938288
[release0.53 9ddebc9] Revert "Merge pull request #6575 from ehsantn/ehsan/avoid_tmp_vars"
 11 files changed, 46 insertions(+), 179 deletions(-)

$ python di58.py # works fine
```

d938288 i.e. merge of https://github.com/numba/numba/pull/6575 seems to be the cause.

CC @ehsantn

ehsantn:
As title. Numba 0.53 exposes this bug since it avoids extra temporary assignments.
Closes #6715.

### DIFF
--- a/numba/core/ir_utils.py
+++ b/numba/core/ir_utils.py
@@ -664,7 +664,12 @@ def remove_dead_block(block, lives, call_table, arg_aliases, alias_map,
         else:
             lives |= {v.name for v in stmt.list_vars()}
             if isinstance(stmt, ir.Assign):
-                lives.remove(lhs.name)
+                # make sure lhs is not used in rhs, e.g. a = g(a)
+                rhs_vars = set()
+                if isinstance(stmt.value, ir.Expr):
+                    rhs_vars = {v.name for v in stmt.value.list_vars()}
+                if lhs.name not in rhs_vars:
+                    lives.remove(lhs.name)
 
         new_body.append(stmt)
     new_body.reverse()
--- a/numba/core/ir_utils.py
+++ b/numba/core/ir_utils.py
@@ -665,10 +665,11 @@ def remove_dead_block(block, lives, call_table, arg_aliases, alias_map,
             lives |= {v.name for v in stmt.list_vars()}
             if isinstance(stmt, ir.Assign):
                 # make sure lhs is not used in rhs, e.g. a = g(a)
-                rhs_vars = set()
                 if isinstance(stmt.value, ir.Expr):
                     rhs_vars = {v.name for v in stmt.value.list_vars()}
-                if lhs.name not in rhs_vars:
+                    if lhs.name not in rhs_vars:
+                        lives.remove(lhs.name)
+                else:
                     lives.remove(lhs.name)
 
         new_body.append(stmt)
--- a/numba/tests/test_remove_dead.py
+++ b/numba/tests/test_remove_dead.py
@@ -218,6 +218,18 @@ class TestRemoveDead(unittest.TestCase):
             # recover global state
             ir_utils.alias_func_extensions = old_ext_handlers
 
+    def test_rm_dead_rhs_vars(self):
+        """make sure lhs variable of assignment is considered live if used in
+        rhs (test for #6715).
+        """
+        @numba.njit
+        def func():
+            for i in range(3):
+                a = (lambda j: j)(i)
+                a = np.array(a)
+
+        func()
+
     @skip_parfors_unsupported
     def test_alias_parfor_extension(self):
         """Make sure aliases are considered in remove dead extension for
--- a/numba/tests/test_remove_dead.py
+++ b/numba/tests/test_remove_dead.py
@@ -222,13 +222,13 @@ class TestRemoveDead(unittest.TestCase):
         """make sure lhs variable of assignment is considered live if used in
         rhs (test for #6715).
         """
-        @numba.njit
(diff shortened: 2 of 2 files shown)


numba/core/ir_utils.py
```diff
@@ -664,7 +664,12 @@ def remove_dead_block(block, lives, call_table, arg_aliases, alias_map,
         else:
             lives |= {v.name for v in stmt.list_vars()}
             if isinstance(stmt, ir.Assign):
-                lives.remove(lhs.name)
+                # make sure lhs is not used in rhs, e.g. a = g(a)
+                rhs_vars = set()
+                if isinstance(stmt.value, ir.Expr):
+                    rhs_vars = {v.name for v in stmt.value.list_vars()}
+                if lhs.name not in rhs_vars:
+                    lives.remove(lhs.name)
```
stuartarchibald, line 672:
```suggestion
                if isinstance(stmt.value, ir.Expr):
                    rhs_vars = {v.name for v in stmt.value.list_vars()}
                    if lhs.name not in rhs_vars:
                        lives.remove(lhs.name)
```
Would this work? The test condition exists solely if the assign value is an `Expr`?

numba/tests/test_remove_dead.py
```diff
@@ -218,6 +218,18 @@ def func(A, i):
...
+        """make sure lhs variable of assignment is considered live if used in
+        rhs (test for #6715).
+        """
+        @numba.njit
+        def func():
+            for i in range(3):
+                a = (lambda j: j)(i)
+                a = np.array(a)
+
+        func()
```
stuartarchibald, line 231: Perhaps return `a` and assert its value? Else this just confirms that it compiles opposed to it having compiled correctly?

stuartarchibald (commented):
Thanks for the patch, https://github.com/numba/numba/pull/6575 highlights this bug which I don't think would have been reachable prior to #6575 without synthetic IR generation, the explanation and fix seems valid on inspection. Couple of minor comments else looks good.

numba/tests/test_remove_dead.py
ehsantn: Done.

numba/core/ir_utils.py
ehsantn: This needs `lives.remove(lhs.name)` for non-Expr nodes. Updated the code since previous structure wasn't clear..

ehsantn:
Thanks for the review @stuartarchibald . I applied the suggestions. Let me know if anything else is needed.

numba/core/ir_utils.py
stuartarchibald: Would something like `a = print(a)` also potentially end up as a similar bug?

numba/tests/test_remove_dead.py
stuartarchibald: thanks

numba/core/ir_utils.py
ehsantn: The `Print` node doesn't have lhs so doesn't go through this. Also, `Print` has side effect and is never removed. 
```
    a = arg(0, name=a)                       ['a']
    $2load_global.0 = global(print: <built-in function print>) ['$2load_global.0']
    print(a)                                 ['a']
    a.1 = const(NoneType, None)              ['a.1']
    $12return_value.4 = cast(value=a.1)      ['$12return_value.4', 'a.1']
    return $12return_value.4                 ['$12return_value.4']
```

stuartarchibald: Ah yes, thanks for checking.

stuartarchibald (approved):

PR #6720, opened 2021-02-12
Faster compile

sklam:
FIxing several things to get faster compile time.

- https://github.com/numba/numba/pull/6720/commits/a852412f163b73b085869b5c4dddbfe491db15e1 avoids recursion in SSA. Almostly to allow the compile deep CFG.
- https://github.com/numba/numba/pull/6720/commits/6e9edbc4fa6c283e3cb0e9531a0fc7294b9d7596 caches `.list_vars()` in SSA.
- https://github.com/numba/numba/pull/6720/commits/3a2b7a424f58894f78c7e2558c2ce5ccf525326b stops withlifting pass from wasting cycles on functions without any `with` at all.
- https://github.com/numba/numba/pull/6720/commits/15e60449440ec93f2c8af3824913f6349aeae478 and onwards are reducing the number of `alloca` needed to cut time spent in SROA pass in LLVM.


sklam:
3a2b7a4 passes everything except flake8

sklam:
Latest commit takes 2sec off of 5sec of SROA in some cases

numba/core/lowering.py
```diff
@@ -1239,9 +1277,16 @@ def _alloca_var(self, name, fetype):
...
-        if name not in self.varmap:
+        if name in self.varmap:
+            # quit early
+            return
+
+        llty = self.context.get_value_type(fetype)
+        # find if name is used in multiple blocks
+        if name in self._singlely_assigned_vars:
+            pass
+        else:
```
stuartarchibald, line 1288:
```suggestion
        # If the name is used in multiple blocks...
        if name not in self._singly_assigned_vars:
```

numba/core/lowering.py
```diff
@@ -1239,9 +1277,16 @@ def _alloca_var(self, name, fetype):
         """
         Ensure the given variable has an allocated stack slot.
```
stuartarchibald, line 1278:
```suggestion
        Ensure the given variable has an allocated stack slot (if needed).
```

numba/core/ssa.py
```diff
@@ -65,16 +80,16 @@ def _run_ssa(blocks):
...
 
-def _fix_ssa_vars(blocks, varname, defmap, cfg, df_plus):
+def _fix_ssa_vars(blocks, varname, defmap, cfg, df_plus, cache_list_vars):
     """Rewrite all uses to ``varname`` given the definition map
     """
     states = _make_states(blocks)
     states['varname'] = varname
     states['defmap'] = defmap
     states['phimap'] = phimap = defaultdict(list)
-    states['cfg'] = cfg = compute_cfg_from_blocks(blocks)
```
stuartarchibald, line 75: NOTE: I think the logic change here is to use the CFG from the top of `_run_ssa` (and `df+` computed from that) as there's nothing that should alter the CFG or these relations prior to computing the phi's? This should save some CFG building.

stuartarchibald (commented):
Thanks for the patch @sklam, code changes in general look good, just a few minor things to resolve. Do you recall the issue(s) related to this change, I'd like to check the performance gains. Many Thanks!

numba/core/ssa.py
sklam: Yes, nothing is changing the CFG. we don't have to recompute this per variable

stuartarchibald: Great, thanks for confirming that my understanding was correct.

stuartarchibald (approved):

PR #6725, opened 2021-02-15
CUDA: Fix compile to PTX with debug for CUDA 11.2

### LINKED ISSUE #6719: Compiling to PTX for a device function with debug fails on CUDA 11.2
```python
from numba import cuda

def f():
    pass

cuda.compile_ptx_for_current_device(f, [], device=True, debug=True)
```

fails with CUDA 11.2 with:

```
  File "repro.py", line 6, in <module>
    cuda.compile_ptx_for_current_device(f, [], device=True, debug=True)
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 124, in compile_ptx_for_current_device
    return compile_ptx(pyfunc, args, debug=-debug, device=device,
    return func(*args, **kwargs)
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 114, in compile_ptx
    ptx = nvvm.llvm_to_ptx(llvmir, opt=opt, arch=arch, **options)
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudadrv/nvvm.py", line 669, in llvm_to_ptx
    ptx = cu.compile(**opts)
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudadrv/nvvm.py", line 265, in compile
    self._try_error(err, 'Failed to compile\n')
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudadrv/nvvm.py", line 283, in _try_error
    self.driver.check_error(err, "%s\n%s" % (msg, self.get_log()))
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudadrv/nvvm.py", line 173, in check_error
    raise exc

DBG version 1.0 incompatible with current version 3.0
<unnamed>: error: incompatible IR detected. Possible mix of compiler/IR from different releases.
NVVM_ERROR_IR_VERSION_MISMATCH
```

This is because the NVVM IR version metadata is not added when compiling a device function, and NVVM assumes version 1 debug metadata if the version metadata is not present. The IR in this case is:

```
; ModuleID = 'f'
source_filename = "<string>"
target datalayout = "e-p:64:64:64-i1:8:8-i8:8:8-i16:16:16-i32:32:32-i64:64:64-f32:32:32-f64:64:64-v16:16:16-v32:32:32-v64:64:64-v128:128:128-n16:32:64"
target triple = "nvptx64-nvidia-cuda"

@"_ZN08NumbaEnv8__main__5f$241E" = common global i8* null

; Function Attrs: noinline
define i32 @"_ZN8__main__5f$241E"(i8** %.ret) #0 !dbg !4 {
entry:
  %"$const2.0" = alloca i8*, !dbg !7
  store i8* null, i8** %"$const2.0", !dbg !7
  %"$4return_value.1" = alloca i8*, !dbg !7
  store i8* null, i8** %"$4return_value.1", !dbg !7
  br label %B0, !dbg !7

B0:                                               ; preds = %entry
  %.4 = load i8*, i8** %"$const2.0", !dbg !7
  store i8* null, i8** %"$const2.0", !dbg !7
  %.6 = load i8*, i8** %"$const2.0", !dbg !7
  %.8 = load i8*, i8** %"$4return_value.1", !dbg !7
  store i8* %.6, i8** %"$4return_value.1", !dbg !7
  %.10 = load i8*, i8** %"$const2.0", !dbg !7
  store i8* null, i8** %"$const2.0", !dbg !7
  %.12 = load i8*, i8** %"$4return_value.1", !dbg !7
  store i8* %.12, i8** %.ret, !dbg !7
  ret i32 0, !dbg !7
}

attributes #0 = { noinline }

!llvm.dbg.cu = !{!0}
!llvm.module.flags = !{!2, !3}

!0 = distinct !DICompileUnit(language: DW_LANG_Python, file: !1, producer: "Numba", isOptimized: true, runtimeVersion: 0, emissionKind: FullDebug)
!1 = !DIFile(filename: "repro.py", directory: "/home/gmarkall/numbadev/issues/compile-ptx-debug")
!2 = !{i32 2, !"Dwarf Version", i32 4}
!3 = !{i32 2, !"Debug Info Version", i32 3}
!4 = distinct !DISubprogram(name: "f", linkageName: "_ZN8__main__5f$241E", scope: !1, file: !1, line: 3, type: !5, scopeLine: 3, isDefinition: true, isOptimized: true, unit: !0)
!5 = !DISubroutineType(types: !6)
!6 = !{}
!7 = !DILocation(line: 4, column: 1, scope: !4)
```

Version metadata is added when the kernel wrapper is created because of the [call to `set_cuda_kernel`](https://github.com/numba/numba/blob/master/numba/cuda/target.py#L212) which adds it [here](https://github.com/numba/numba/blob/master/numba/cuda/cudadrv/nvvm.py#L907).

To resolve this issue, version metadata should also be added by `compile_ptx` when compiling a device function.

gmarkall:
The NVVM IR version metadata needs to be present in all modules passed to NVVM. The IR version was only set when kernel wrapper functions were generated, so device functions never had the IR version added.

This commit remedies this by adding the IR version to all modules in the CUDA target, rather than relying on the kernel wrapper generation function to do it.

It is hard to get at the IR at the right time before it is turned into a ModuleRef from the llvmlite binding layer - in order to work around this, a callback facility is added to the lowering that allows the metadata to be added in just after the module creation.

Fixes #6719.

### DIFF
--- a/numba/cuda/codegen.py
+++ b/numba/cuda/codegen.py
@@ -50,6 +50,7 @@ class JITCUDACodegen(BaseCPUCodegen):
         ir_module.triple = CUDA_TRIPLE[utils.MACHINE_BITS]
         if self._data_layout:
             ir_module.data_layout = self._data_layout
+        nvvm.add_ir_version(ir_module)
         return ir_module
 
     def _module_pass_manager(self):
--- a/numba/cuda/compiler.py
+++ b/numba/cuda/compiler.py
@@ -30,12 +30,10 @@ from .args import wrap_arg
 
 @global_compiler_lock
 def compile_cuda(pyfunc, return_type, args, debug=False, inline=False):
-    # First compilation will trigger the initialization of the CUDA backend.
-    from .descriptor import CUDATargetDesc
+    from .descriptor import cuda_target
+    typingctx = cuda_target.typingctx
+    targetctx = cuda_target.targetctx
 
-    typingctx = CUDATargetDesc.typingctx
-    targetctx = CUDATargetDesc.targetctx
-    # TODO handle debug flag
     flags = compiler.Flags()
     # Do not compile (generate native code), just lower (to LLVM)
     flags.set('no_compile')
@@ -249,7 +247,7 @@ def compile_device_template(pyfunc, debug=False, inline=False, opt=True):
     """Create a DeviceFunctionTemplate object and register the object to
     the CUDA typing context.
     """
-    from .descriptor import CUDATargetDesc
+    from .descriptor import cuda_target
 
     dft = DeviceFunctionTemplate(pyfunc, debug=debug, inline=inline, opt=opt)
 
@@ -275,7 +273,7 @@ def compile_device_template(pyfunc, debug=False, inline=False, opt=True):
             }
             return info
 
-    typingctx = CUDATargetDesc.typingctx
+    typingctx = cuda_target.typingctx
     typingctx.insert_user_function(dft, device_function_template)
     return dft
 
@@ -285,10 +283,9 @@ def compile_device(pyfunc, return_type, args, inline=True, debug=False):
 
 
 def declare_device_function(name, restype, argtypes):
-    from .descriptor import CUDATargetDesc
-
-    typingctx = CUDATargetDesc.typingctx
-    targetctx = CUDATargetDesc.targetctx
+    from .descriptor import cuda_target
+    typingctx = cuda_target.typingctx
+    targetctx = cuda_target.targetctx
     sig = typing.signature(restype, *argtypes)
     extfn = ExternFunction(name, sig)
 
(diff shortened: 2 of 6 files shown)


gmarkall:
This breaks the docs build when the CUDA toolkit is not installed, so I'll have to look at that shortly.

gmarkall:
The issue breaking the docs build (an attempt to load libnvvm.so at import time) is now resolved - see the commit message for more details.

sklam (approved):
The changes look great. The detailed description in the commit messages helped a lot.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_16`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_16`.

This passed.

PR #6727, opened 2021-02-15
Avoid errors in array analysis for global tuples with non-int

### LINKED ISSUE #6726: 0.53 regression in array analysis for tuple globals
## Reporting a bug

- [x] I have tried using the latest released (0.53.0rc1) version of Numba (most recent is

I'm guessing this regression is due to the newly added global support in array analysis:
```python
import numba

gval = (1.2,)

@numba.njit(parallel=True)
def f():
    d = gval[0]

f()
```
```
  File "../arr_global_tuple.py", line 10, in <module>
    f()
  File "/Users/ehsan/tmp/numba/numba/core/dispatcher.py", line 439, in _compile_for_args
    raise e
  File "/Users/ehsan/tmp/numba/numba/core/dispatcher.py", line 372, in _compile_for_args
    return_val = self.compile(tuple(argtypes))
  File "/Users/ehsan/tmp/numba/numba/core/dispatcher.py", line 909, in compile
    cres = self._compiler.compile(args, return_type)
  File "/Users/ehsan/tmp/numba/numba/core/dispatcher.py", line 79, in compile
    status, retval = self._compile_cached(args, return_type)
  File "/Users/ehsan/tmp/numba/numba/core/dispatcher.py", line 93, in _compile_cached
    retval = self._compile_core(args, return_type)
  File "/Users/ehsan/tmp/numba/numba/core/dispatcher.py", line 106, in _compile_core
    cres = compiler.compile_extra(self.targetdescr.typing_context,
  File "/Users/ehsan/tmp/numba/numba/core/compiler.py", line 606, in compile_extra
  File "/Users/ehsan/tmp/numba/numba/core/compiler.py", line 353, in compile_extra
  File "/Users/ehsan/tmp/numba/numba/core/compiler.py", line 415, in _compile_bytecode
    return self._compile_core()
  File "/Users/ehsan/tmp/numba/numba/core/compiler.py", line 395, in _compile_core
    raise e
  File "/Users/ehsan/tmp/numba/numba/core/compiler.py", line 386, in _compile_core
    pm.run(self.state)
  File "/Users/ehsan/tmp/numba/numba/core/compiler_machinery.py", line 339, in run
    raise patched_exception
  File "/Users/ehsan/tmp/numba/numba/core/compiler_machinery.py", line 330, in run
  File "/Users/ehsan/tmp/numba/numba/core/compiler_lock.py", line 35, in _acquire_compile_lock
    return func(*args, **kwargs)
  File "/Users/ehsan/tmp/numba/numba/core/compiler_machinery.py", line 289, in _runPass
  File "/Users/ehsan/tmp/numba/numba/core/compiler_machinery.py", line 262, in check
  File "/Users/ehsan/tmp/numba/numba/core/typed_passes.py", line 303, in run_pass
    parfor_pass.run()
  File "/Users/ehsan/tmp/numba/numba/parfors/parfor.py", line 2777, in run
    self._pre_run()
  File "/Users/ehsan/tmp/numba/numba/parfors/parfor.py", line 2769, in _pre_run
    self.array_analysis.run(self.func_ir.blocks)
  File "/Users/ehsan/tmp/numba/numba/parfors/array_analysis.py", line 1155, in run
    self._run_on_blocks(topo_order, blocks, cfg, init_equiv_set)
  File "/Users/ehsan/tmp/numba/numba/parfors/array_analysis.py", line 1175, in _run_on_blocks
    pending_transforms = self._determine_transform(
  File "/Users/ehsan/tmp/numba/numba/parfors/array_analysis.py", line 1256, in _determine_transform
    pre, post = self._analyze_inst(
  File "/Users/ehsan/tmp/numba/numba/parfors/array_analysis.py", line 1384, in _analyze_inst
    needs_define = equiv_set.insert_equiv(lhs, shape)
  File "/Users/ehsan/tmp/numba/numba/parfors/array_analysis.py", line 524, in insert_equiv
    obj_names = [self._get_names(x) for x in objs]
  File "/Users/ehsan/tmp/numba/numba/parfors/array_analysis.py", line 524, in <listcomp>
    obj_names = [self._get_names(x) for x in objs]
  File "/Users/ehsan/tmp/numba/numba/parfors/array_analysis.py", line 433, in _get_names
    return tuple(self._get_names(x)[0] for x in obj)
  File "/Users/ehsan/tmp/numba/numba/parfors/array_analysis.py", line 433, in <genexpr>
    return tuple(self._get_names(x)[0] for x in obj)
  File "/Users/ehsan/tmp/numba/numba/parfors/array_analysis.py", line 437, in _get_names
    raise NotImplementedError(
NotImplementedError: Failed in nopython mode pipeline (step: convert to parfors)
ShapeEquivSet does not support [1, 2]
```

ehsantn:
Closes #6726.

### DIFF
--- a/numba/tests/test_array_analysis.py
+++ b/numba/tests/test_array_analysis.py
@@ -49,7 +49,6 @@ GVAL = (1.2,)
 GVAL2 = (3, 4)
 
 
-
 class TestEquivSet(TestCase):
 
     """
@@ -1035,7 +1034,7 @@ class TestArrayAnalysisParallelRequired(TestCase):
 
     @skip_unsupported
     def test_global_tuple(self):
-        """make sure a gobal value with non-integers does not cause errors
+        """make sure a global tuple with non-integer values does not cause errors
         (test for #6726).
         """
 
--- a/numba/tests/test_array_analysis.py
+++ b/numba/tests/test_array_analysis.py
@@ -44,6 +44,11 @@ class ExampleClass3700(object):
         self.T = n + 1
 
 
+# test value for test_global_tuple
+GVAL = (1.2,)
+
+
+
 class TestEquivSet(TestCase):
 
     """
@@ -1016,6 +1021,18 @@ class TestArrayAnalysisParallelRequired(TestCase):
 
         data = np.arange(10.)
         np.testing.assert_array_equal(test_impl(data), test_impl.py_func(data))
+
+    @skip_unsupported
+    def test_global_tuple(self):
+        """make sure a gobal value with non-integers does not cause errors
+        (test for #6726).
+        """
+
+        def test_impl():
+            d = GVAL[0]
+            return d
+
+        self.assertEqual(njit(test_impl, parallel=True)(), test_impl())
 
 
 class TestArrayAnalysisInterface(TestCase):
--- a/numba/tests/test_array_analysis.py
+++ b/numba/tests/test_array_analysis.py
@@ -46,6 +46,7 @@ class ExampleClass3700(object):
 
 # test value for test_global_tuple
 GVAL = (1.2,)
+GVAL2 = (3, 4)
 
 
 
@@ -934,6 +935,16 @@ class TestArrayAnalysis(TestCase):
         self._compile_and_test(test_broadcast, (types.intp, types.intp),
                                equivs=[self.with_equiv('a', 'c', 'e')],
                                asserts=None)
+
+        # make sure shape of a global tuple of ints is handled properly
+        def test_global_tuple():
+            a = np.ones(GVAL2)
+            b = np.ones(GVAL2)
+
+        self._compile_and_test(test_global_tuple, (),
+                               equivs=[self.with_equiv('a', 'b')],
+                               asserts=None)
+
 
 class TestArrayAnalysisParallelRequired(TestCase):
     """This is to just split out tests that need the parallel backend and
--- a/numba/parfors/array_analysis.py
+++ b/numba/parfors/array_analysis.py
@@ -1306,7 +1306,11 @@ class ArrayAnalysis(object):
                 shape = inst.value
             elif isinstance(inst.value, ir.Global):
                 gvalue = inst.value.value
-                if isinstance(gvalue, tuple):
+                # only integer values can be part of shape
+                # TODO: support cases with some but not all integer values or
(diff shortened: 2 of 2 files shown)


DrTodd13 (changes requested):
Add a test that should work with a global tuple of ints?

Now that I think about it, if we compile a method and the global is a tuple of ints and after the function has executed we change the global to be something else entirely, then doesn't that invalidate the compilation?  Is Numba dispatcher keyed on globals that are used and their type?  Even if that were the case then doesn't this code bake in the value of the global at compilation time so if you change it to a different tuple of ints than the original then compilation wouldn't be triggered again and the baked in array analysis would be off.  So, unless you dispatched on not only the type but the value of the global then don't we have a potential problem here?

ehsantn:
I just assumed there is a test for the case that the new code supports, but I'll create some tests to be sure.

This is a known Numba limitation in general: https://numba.readthedocs.io/en/stable/reference/pysemantics.html#global-and-closure-variables

DrTodd13:
@ehsantn As long as it is documented that way it is fine then.  Thanks for the patch.

ehsantn:
@DrTodd13 added a test for a global tuple of ints.

numba/tests/test_array_analysis.py
```diff
@@ -44,6 +44,12 @@ def __init__(self, n):
         self.T = n + 1
 
 
+# test value for test_global_tuple
+GVAL = (1.2,)
+GVAL2 = (3, 4)
+
+
+
```
stuartarchibald, line 52:
```suggestion
```

numba/tests/test_array_analysis.py
```diff
@@ -1017,6 +1033,18 @@ def test_impl(data):
         data = np.arange(10.)
         np.testing.assert_array_equal(test_impl(data), test_impl.py_func(data))
 
+    @skip_unsupported
+    def test_global_tuple(self):
+        """make sure a gobal value with non-integers does not cause errors
```
stuartarchibald, line 1038:
```suggestion
        """make sure a global tuple with non-integer values does not cause errors
```

stuartarchibald (commented):
Thanks for the patch, couple of typos else looks good.

stuartarchibald (approved):

PR #6728, opened 2021-02-16
[WIP] Implement IR-only pipline for CUDA  (was "Pull NativeLowering pass out into pipeline")

gmarkall:


numba/cuda/codegen.py
```diff
@@ -11,6 +11,12 @@
 
 
 class CUDACodeLibrary(CodeLibrary):
```
gmarkall, line 13: Note to self: needs some commentary on the differences between it and other CodeLibrary implementations, in particular how it only handles IR from llvmlite.ir, not entities from llvmlite.binding.

gmarkall:
Closing this now that the work in it has gone into #6731 and #6735.

PR #6730, opened 2021-02-18
Fix Typeguard error

gmarkall:
Typeguard reports:

```
TypeError: can only assign string to GUFunc.__name__, not 'property'
```

It seems that setting `__name__` in `__init__` should achieve the same effect, and avoid the Typeguard error.


sklam (approved):
LGTM

PR #6731, opened 2021-02-18
Add CUDA-specific pipeline

gmarkall:
In preparation for customizing the CUDA compilation pipeline, this PR adds a CUDA pipeline separate from the Nopython pipeline it currently uses. This initial CUDA pipeline is identical to the Nopython pipeline, with two exceptions:

- It uses the `CUDABackend` stage instead of the `NoPythonBackend` stage. The `CUDABackend` stage populates a `compile_result` like the `NoPythonBackend` stage, but doesn't populate it with things that make no sense in CUDA (e.g. `objectmode`, `lifted`, etc.).
- It omits the `DumpParforDiagnostics` stage.

This needed the `NativeLowering` stage to be moved out into the pipeline definition instead of being run from within `NoPythonBackend`.

### DIFF
--- a/numba/core/compiler.py
+++ b/numba/core/compiler.py
@@ -28,7 +28,8 @@ from numba.core.typed_passes import (NopythonTypeInference, AnnotateTypes,
                                      NopythonRewrites, PreParforPass,
                                      ParforPass, DumpParforDiagnostics,
                                      IRLegalization, NoPythonBackend,
-                                     InlineOverloads, PreLowerStripPhis)
+                                     InlineOverloads, PreLowerStripPhis,
+                                     NativeLowering)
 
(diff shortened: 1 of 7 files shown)


gmarkall:
The fail was only the typeguard issue, so I'm marking this as RfR.

stuartarchibald (approved):
Thanks for the patch, and for fixing my TODO's :) Nice to see CUDA getting its own pipeline!

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_14`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_14`.

Passed.

PR #6733, opened 2021-02-18
Fix segfault and errors in #6668

sklam:
by guarding against invalid use and unsupported types.

This patch fixes usecases:

- 2, 3: reflected containers are not supported because their boxing/unboxing logic is assuming that the original PyObject to continue to exist after unboxing and through boxing; i.e. assumptions on arguments existing throughout the lifetime of a function call. However, this assumption does not hold in the case of objmode. The unboxed object is dead after unbox. Users should instead use `typed.List` instead.
- 4: it is a user error to use python `dict` when the objmode context is expecting a `typed.Dict`. The patch fixes the unboxer to check for the type of the incoming value and raises `TypeError` appropriately.

### DIFF
--- a/numba/core/pythonapi.py
+++ b/numba/core/pythonapi.py
@@ -950,6 +950,11 @@ class PythonAPI(object):
         fn = self._get_function(fnty, name="PyObject_Call")
         return self.builder.call(fn, (callee, args, kws))
 
+    def object_type(self, obj):
+        fnty = Type.function(self.pyobj, [self.pyobj] * 1)
+        fn = self._get_function(fnty, name="PyObject_Type")
+        return self.builder.call(fn, (obj,))
+
     def object_istrue(self, obj):
         fnty = Type.function(Type.int(), [self.pyobj])
         fn = self._get_function(fnty, name="PyObject_IsTrue")
--- a/numba/core/pythonapi.py
+++ b/numba/core/pythonapi.py
@@ -951,6 +951,8 @@ class PythonAPI(object):
         return self.builder.call(fn, (callee, args, kws))
 
     def object_type(self, obj):
+        """Emit a call to ``PyObject_Type(obj)`` to get the type of ``obj``.
+        """
         fnty = Type.function(self.pyobj, [self.pyobj] * 1)
         fn = self._get_function(fnty, name="PyObject_Type")
         return self.builder.call(fn, (obj,))
--- a/numba/core/pythonapi.py
+++ b/numba/core/pythonapi.py
@@ -953,7 +953,7 @@ class PythonAPI(object):
     def object_type(self, obj):
         """Emit a call to ``PyObject_Type(obj)`` to get the type of ``obj``.
         """
-        fnty = Type.function(self.pyobj, [self.pyobj] * 1)
+        fnty = Type.function(self.pyobj, [self.pyobj])
         fn = self._get_function(fnty, name="PyObject_Type")
         return self.builder.call(fn, (obj,))
 
--- a/numba/core/withcontexts.py
+++ b/numba/core/withcontexts.py
@@ -247,8 +247,9 @@ class _ObjModeContextType(WithContext):
         if getattr(typ, "reflected", False):
             msgbuf = [
                 "Objmode context failed.",
-                f"Argument {name!r} is using unsupported type: {typ}.",
-                f"Reflected types is not supported.",
+                f"Argument {name!r} is declared as "
+                f"an unsupported type: {typ}.",
+                f"Reflected types are not supported.",
             ]
             raise errors.CompilerError(" ".join(msgbuf), loc=loc)
 
--- a/numba/core/withcontexts.py
+++ b/numba/core/withcontexts.py
@@ -226,7 +226,23 @@ class _ObjModeContextType(WithContext):
                     "type annotation",
                 )
 
+        # Legalize the types for objmode
+        for name, typ in typeanns.items():
+            self._legalize_arg_type(name, typ, loc)
+
         return typeanns
+
+    def _legalize_arg_type(self, name, typ, loc):
+        def report_error(msg):
+            msgbuf = [
+                "Objmode context failed.",
+                f"Argument {name!r} is using unsupported type.",
+                msg,
+            ]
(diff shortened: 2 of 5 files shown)


numba/typed/typeddict.py
```diff
@@ -260,27 +260,59 @@ def box_dicttype(typ, val, c):
...
-    c.pyapi.decref(miptr)
-    return NativeValue(dctobj, is_error=is_error)
+            sig = signature(typ, *argtypes)
+            nil_typeref = context.get_constant_null(argtypes[1])
+            args = (mi, nil_typeref)
+            is_error, dctobj = c.pyapi.call_jit_code(convert , sig, args)
+            # decref here because we are stealing a reference.
+            c.context.nrt.decref(c.builder, typ, dctobj)
+
+            c.pyapi.decref(miptr)
```
sklam, line 289: All I changed here is increasing the indent so these lines go inside the `if-then` branch.

numba/tests/test_withlifting.py
```diff
@@ -945,6 +945,56 @@ def case_objmode_cache(x):
...
+        def test2():
+            with objmode(out=ret_type):
+                out = [1, 2, 3, 4, 5]
+            return out
+
+        with self.assertRaises(errors.CompilerError) as raises:
+            test2()
+        self.assertIn(
+            ("Objmode context failed. "
+             "Argument 'out' is using unsupported type. "
```
stuartarchibald, line 960: Should the exact unsupported Numba type also be reported to the user?

numba/typed/typeddict.py
```diff
@@ -260,27 +260,59 @@ def box_dicttype(typ, val, c):
 def unbox_dicttype(typ, val, c):
     context = c.context
 
-    miptr = c.pyapi.object_getattr_string(val, '_opaque')
+    # Check that `type(val) is Dict`
+    dict_type = c.pyapi.unserialize(c.pyapi.serialize_object(Dict))
+    valtype = c.pyapi.object_type(val)
```
stuartarchibald, line 265: Note to self: references are cleaned prior to `NativeValue` return.

numba/typed/typeddict.py
stuartarchibald: Thanks. Confirmed.

numba/tests/test_withlifting.py
```diff
@@ -945,6 +945,56 @@ def case_objmode_cache(x):
...
+        def test4():
+            with objmode(res=ret_type):
+                res = {'A': 1, 'B': 2}
+            return res
+
+        with self.assertRaises(TypeError) as raises:
+            test4()
+        self.assertIn(
+            ("can't unbox <class 'numba.typed.typeddict.Dict'> "
+             "from <class 'dict'>"),
```
stuartarchibald, line 994: Am I right in thinking that the issue here is that a CPython dictionary cannot be unboxed to a `typed.Dict`, and more generally that the returned object will not unbox as the declared type? If so, would something like `"Cannot unbox a <class 'dict'> as a <class 'numba.typed.typeddict.Dict'>"` be more clear?

numba/typed/typeddict.py
```diff
@@ -260,27 +260,59 @@ def box_dicttype(typ, val, c):
 def unbox_dicttype(typ, val, c):
     context = c.context
 
-    miptr = c.pyapi.object_getattr_string(val, '_opaque')
+    # Check that `type(val) is Dict`
```
stuartarchibald, line 263: I think this same type-check based fix is also needed in `typed.List` ?

stuartarchibald (commented):
Thanks for fixing this, patch in general looks good, there's a few minor things to resolve, most notable is whether to apply the same fix to `typed.List` whilst doing this fix. Thanks again!

numba/tests/test_withlifting.py
sklam: Yes, i'll change that.

numba/typed/typeddict.py
sklam: right, i'll add that

numba/core/pythonapi.py
```diff
@@ -950,6 +950,13 @@ def call(self, callee, args=None, kws=None):
         fn = self._get_function(fnty, name="PyObject_Call")
         return self.builder.call(fn, (callee, args, kws))
 
+    def object_type(self, obj):
+        """Emit a call to ``PyObject_Type(obj)`` to get the type of ``obj``.
+        """
+        fnty = Type.function(self.pyobj, [self.pyobj] * 1)
```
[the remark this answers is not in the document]
sklam: removed in https://github.com/numba/numba/pull/6733/commits/c2433b6959c0af1623f508ce131dd74007cf2791

numba/core/withcontexts.py
```diff
@@ -226,8 +226,32 @@ def _legalize_args(self, args, kwargs, loc, func_globals, func_closures):
...
+            argument name.
+        typ: numba.core.types.Type
+            argument type.
+        loc: numba.core.ir.Loc
+            source location for error reporting.
+        """
+        if getattr(typ, "reflected", False):
+            msgbuf = [
+                "Objmode context failed.",
+                f"Argument {name!r} is using unsupported type.",
```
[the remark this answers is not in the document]
sklam: changed in https://github.com/numba/numba/pull/6733/commits/efd58758559173b92de4adce80d7ac3a427a4cd3

numba/core/withcontexts.py
```diff
@@ -226,8 +226,32 @@ def _legalize_args(self, args, kwargs, loc, func_globals, func_closures):
...
+                f"Reflected types is not supported.",
```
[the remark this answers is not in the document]
sklam: changed in https://github.com/numba/numba/pull/6733/commits/efd58758559173b92de4adce80d7ac3a427a4cd3

numba/tests/test_withlifting.py
sklam: yes, done in https://github.com/numba/numba/pull/6733/commits/c2433b6959c0af1623f508ce131dd74007cf2791

stuartarchibald (approved):

sklam:
CI fail related to https://github.com/numba/numba/pull/6751

Issue #5135, opened 2020-01-23
Support for dictionary comprehensions.


phyukhaing7:
Using TensorFlow backend.
Traceback (most recent call last):
  File "D:/Research/code/Python/code/main.py", line 111, in <module>
    main()
  File "D:\Anaconda3\lib\site-packages\numba\dispatcher.py", line 404, in _compile_for_args
    error_rewrite(e, 'unsupported_error')
  File "D:\Anaconda3\lib\site-packages\numba\dispatcher.py", line 344, in error_rewrite
    reraise(type(e), e, None)
  File "D:\Anaconda3\lib\site-packages\numba\six.py", line 668, in reraise
    raise value.with_traceback(tb)
numba.errors.UnsupportedError: Failed in object mode pipeline (step: inline calls to locally defined closures)
Use of unsupported opcode (MAP_ADD) found

File "main.py", line 81:
```
def main():
    <source elided>
    random.shuffle(ids_train)
    X2train_shuffled = {_id: X2train[_id] for _id in ids_train}
    ^

```

esc:
@phyukhaing7 thannks for asking about this on the Numba issue tracker. This type of dictionary comprehension (not 100% about the name) isn't supported by Numba at this time. A minimal reproducer follows:


```python
from numba import njit


@njit
def foo():
    return {i: i for i in range(10)}


print(foo.py_func())
print(foo())
```

stuartarchibald:
The issue to here to fix first is to make the closure inliner give a better error message... then to work on implementing this.

gioxc88:
any update on this?
it seems like a basic feature.

Many thanks

stuartarchibald:
> any update on this?
> it seems like a basic feature.
> 
> Many thanks

@gioxc88 https://github.com/numba/numba/pull/6736 implements.

PR #6736, opened 2021-02-19
Support for dict comprehension

stuartarchibald:
As title.

Closes: #5135

### DIFF
--- a/docs/source/reference/pysupported.rst
+++ b/docs/source/reference/pysupported.rst
@@ -796,6 +796,23 @@ range of possible failures. However, the dictionary can be safely read from
 multiple threads as long as the contents of the dictionary do not
 change during the parallel access.
 
+Dictionary comprehension
+''''''''''''''''''''''''
+
+Numba supports dictionary comprehension under the assumption that a
+``numba.typed.Dict`` instance can be created from the comprehension.  For
+example::
+
+  In [1]: from numba import njit
+
+  In [2]: @njit
+  ...: def foo(n):
+  ...:     return {i: i**2 for i in range(n)}
+  ...:
+
+  In [3]: foo(3)
+  Out[3]: DictType[int64,int64]<iv=None>({0: 0, 1: 1, 2: 4})
+
 .. _feature-dict-initial-value:
 
 Initial Values
--- a/docs/source/reference/pysupported.rst
+++ b/docs/source/reference/pysupported.rst
@@ -806,9 +806,9 @@ example::
   In [1]: from numba import njit
 
   In [2]: @njit
-  ...: def foo(n):
-  ...:     return {i: i**2 for i in range(n)}
-  ...:
+     ...: def foo(n):
+     ...:     return {i: i**2 for i in range(n)}
+     ...:
 
   In [3]: foo(3)
   Out[3]: DictType[int64,int64]<iv=None>({0: 0, 1: 1, 2: 4})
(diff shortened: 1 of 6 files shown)


docs/source/reference/pysupported.rst
```diff
@@ -796,6 +796,23 @@ range of possible failures. However, the dictionary can be safely read from
...
+Numba supports dictionary comprehension under the assumption that a
+``numba.typed.Dict`` instance can be created from the comprehension.  For
+example::
+
+  In [1]: from numba import njit
+
+  In [2]: @njit
+  ...: def foo(n):
+  ...:     return {i: i**2 for i in range(n)}
+  ...:
```
esc, line 811:
```suggestion
In [2]: @njit
   ...: def foo(n):
   ...:     return {i: i**2 for i in range(n)}
   ...:
```

numba/core/dataflow.py
```diff
@@ -202,6 +202,16 @@ def op_BUILD_MAP(self, info, inst):
         info.append(inst, items=items[::-1], size=count, res=dct)
         info.push(dct)
 
+    def op_MAP_ADD(self, info, inst):
```
esc, line 205: I tried to check if this code was ever exercised, by using the patch:

```
Encountered the use of a type that is scheduled for deprecation: type 'reflected list' found for argument 'vals' of function 'TestDictObject.test_dict_values.<locals>.diff --git i/numba/core/byteflow.py w/numba/core/byteflow.py
index 5c612f3a08..bf7f48cf84 100644
--- i/numba/core/byteflow.py
+++ w/numba/core/byteflow.py
@@ -920,6 +920,8 @@ class TraceRunner(object):
         # NOTE: https://docs.python.org/3/library/dis.html#opcode-MAP_ADD
         # Python >= 3.8: TOS and TOS1 are value and key respectively
         # Python < 3.8: TOS and TOS1 are key and value respectively
+        import ipdb
+        ipdb.set_trace()
         TOS = state.pop()
         TOS1 = state.pop()
         key, value = (TOS, TOS1) if PYVERSION < (3, 8) else (TOS1, TOS)
diff --git i/numba/core/dataflow.py w/numba/core/dataflow.py
index c351086a50..646103e682 100644
--- i/numba/core/dataflow.py
+++ w/numba/core/dataflow.py
@@ -203,6 +203,8 @@ class DataFlowAnalysis(object):
         info.push(dct)

     def op_MAP_ADD(self, info, inst):
+        import ipdb
+        ipdb.set_trace()
         key = info.pop()
         value = info.pop()
         index = inst.arg
```

But it never stopped in `datalflow.py` when executing the tests in this PR.

esc (changes requested):
One suggestion to fix doc formatting and a question about test coverage.

numba/core/dataflow.py
esc: So, `dataflow.py` is for older Pythons.

docs/source/reference/pysupported.rst
stuartarchibald: well spotted, colons are lined up in https://github.com/numba/numba/pull/6736/commits/7145e6cdf6f9d53a93281d89c7efe66f871b3702

esc: excellent, ready to smoketest then?

stuartarchibald: Think so!

esc:
Build farm ID: `numba_smoketest_cpu_91`

numba/core/dataflow.py
sklam: Right, for python<3.7. I can't wait to drop py3.6.

esc (approved):

esc:
Smoketest was green, this is good to go in.

PR #6741, opened 2021-02-21
Enable SSA in IR inliner

### LINKED ISSUE #6721: Regression in 0.53.0rc1 for boolean conversion
See reproducer below. Reverting the temporary assignment change (d938288) solves the issue, but the underlying bug is most probably something else. This only happens with inlining so maybe inliner needs a closer look.

```python
import numba
import numpy as np
import pandas as pd

val_one = True

def f(S):
    pass

@numba.extending.overload(f, inline="always")
def f_overload(A):

    def impl(A):
        for i in range(len(A)):
            val = val_one
            if A[i]:
                val = A[i]
        return val
    return impl

@numba.njit
def f2(A):
    return f(A)

print(f2(np.array([True, False, True])))
```
```
  File "../bool_cast_err2.py", line 25, in <module>
    print(f2(np.array([True, False, True])))
  File "/Users/ehsan/tmp/numba/numba/core/dispatcher.py", line 420, in _compile_for_args
    error_rewrite(e, 'typing')
  File "/Users/ehsan/tmp/numba/numba/core/dispatcher.py", line 361, in error_rewrite
    raise e.with_traceback(None)
No implementation of function Function(<function f at 0x7fc275ff3f70>) found for signature:

 >>> f(array(bool, 1d, C))

  Overload in function 'f_overload': File: ../bool_cast_err2.py: Line 10.
    With argument(s): '(array(bool, 1d, C))':
     TypingError: No conversion from bool to Literal[bool](True) for 'val', defined at ../bool_cast_err2.py (15)

   File "../bool_cast_err2.py", line 17:
       def impl(A):
           <source elided>
               if A[i]:
                   val = A[i]
                   ^

   During: typing of intrinsic-call at ../bool_cast_err2.py (17)

   File "../bool_cast_err2.py", line 17:
       def impl(A):
           <source elided>
               if A[i]:
                   val = A[i]
                   ^

  raised from /Users/ehsan/tmp/numba/numba/core/typeinfer.py:1071

During: resolving callee type: Function(<function f at 0x7fc275ff3f70>)
During: typing of call at ../bool_cast_err2.py (23)


File "../bool_cast_err2.py", line 23:
def f2(A):
    return f(A)
    ^
```

ehsantn:
This adds SSA in IR inlining since some functions to be inlined may require SSA for proper typing.
Closes #6721.

### DIFF
--- a/numba/core/inline_closurecall.py
+++ b/numba/core/inline_closurecall.py
@@ -413,6 +413,8 @@ class InlineWorker(object):
             if arg_typs is None:
                 raise TypeError('arg_typs should have a value not None')
             self.update_type_and_call_maps(callee_ir, arg_typs)
+            # update_type_and_call_maps replaces blocks
+            callee_blocks = callee_ir.blocks
 
         self.debug_print("After arguments rename: ")
         _debug_dump(callee_ir)
@@ -462,10 +464,13 @@ class InlineWorker(object):
         return self.inline_ir(caller_ir, block, i, callee_ir, freevars,
                               arg_typs=arg_typs)
 
-    def run_untyped_passes(self, func):
+    def run_untyped_passes(self, func, enable_ssa=False):
         """
         Run the compiler frontend's untyped passes over the given Python
         function, and return the function's canonical Numba IR.
+
+        Disable SSA transformation by default, since the call site won't be in SSA
+        form and self.inline_ir depends on this being the case.
         """
         from numba.core.compiler import StateDict, _CompileStatus
         from numba.core.untyped_passes import ExtractByteCode, WithLifting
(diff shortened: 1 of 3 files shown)


stuartarchibald:
Thanks for the patch @ehsantn.

sklam:
@ehsantn, am I right that getting rid of multiple assignment to variable of the same name is the main fix? So applying SSA in problematic area does the magic.

ehsantn:
Yes, I believe that's correct.

stuartarchibald:
@ehsantn In reviewing this and trying to understand the original bug and how this fixes it, am wondering if something like this https://github.com/numba/numba/pull/6748/files would produce an equivalent effect to that contained in this PR?

ehsantn:
I don't think #6748 solves the problem completely, even though it happens to fix this particular reproducer. I just tried, and it doesn't fix the actual use case. To be safe, all code going through type inference should be SSA I think.

ehsantn:
Here is an example that doesn't work with #6748 :
```python
import numba
import numpy as np

dummy_one = True

def f(S):
    pass

@numba.extending.overload(f, inline="always")
def f_overload(A):

    def impl(A):
        s = dummy_one
        for i in range(len(A)):
            dummy = dummy_one
            if A[i]:
                dummy = A[i]
            s *= dummy
        return s
    return impl

@numba.njit
def f2(A):
    return f(A)

print(f2(np.array([True, False, True])))
```

sklam (approved):
Thanks for the fix.

We will need to revisit making SSA enabled everywhere later.

Issue #6737, opened 2021-02-19
Docs: redefinition of stdlib functions causing invalid cross references.


stuartarchibald:
## Reporting a bug

- [x] I have tried using the latest released version of Numba (most recent is
 visible in the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [ ] N/A ~I have included a self contained code sample to reproduce the problem.
  i.e. it's possible to run as 'python bug.py'.~

Following on from https://github.com/numba/numba/pull/6619#discussion_r578554321 the doc pages that come from "autogeneration" via `numba.misc.help.inspector` redefine functions from python stdlib modules, e.g. `math.sin`, as a result referring to `` :func:`math.sin` `` in docs leads to a link being generated to the Numba redefinition opposed to the Python docs.

gmarkall:
Would marking them `:noindex:` help?

stuartarchibald:
> Would marking them `:noindex:` help?

Good idea, thanks, https://github.com/numba/numba/pull/6742 fixes.

PR #6742, opened 2021-02-22
Do not add overload function definitions to index.

stuartarchibald:
As title. Prevents autodoc links being broken by a local overload
definition.

Fixes: #6737


sklam (approved):
Thanks for the patch!

I've checked that the stdlib references are now referring to the correct Python documentation.

PR #6751, opened 2021-02-23
Suppress typeguard warnings that affect testing.

sklam:
New typeguard 2.11.1 is released and there're new warnings raised for everything time it finds something that is not annotated. We need to suppress these warnings for tests that checks on warnings to pass.

### DIFF
--- a/buildscripts/incremental/test.sh
+++ b/buildscripts/incremental/test.sh
@@ -79,7 +79,7 @@ if [ "$RUN_COVERAGE" == "yes" ]; then
     $SEGVCATCH coverage run runtests.py -b -j "$TEST_START_INDEX,None,$TEST_COUNT" --exclude-tags='long_running' -m $TEST_NPROCS -- numba.tests
 elif [ "$RUN_TYPEGUARD" == "yes" ]; then
     echo "INFO: Running with typeguard"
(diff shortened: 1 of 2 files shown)


esc:
Seems like there is only a single issue with the test suite now:

```
FAIL: test_loop_lift_warn (numba.tests.test_warnings.TestBuiltins)
  File "/home/vsts/work/1/s/numba/tests/test_warnings.py", line 116, in test_loop_lift_warn
    self.assertEqual(len(w), 4, msg=msg)
AssertionError: 6 != 4 : ----------

Compilation is falling back to object mode WITH looplifting enabled because Function "do_loop" failed type inference due to: Cannot infer the type of variable 'a', have imprecise type: DictType[undefined,undefined]<iv={}>. 

File "numba/tests/test_warnings.py", line 103:
        def do_loop(x):
            a = {} # noqa dead
            ^

----------

Compilation is falling back to object mode WITHOUT looplifting enabled because Function "do_loop" failed type inference due to: Cannot determine Numba type of <class 'numba.core.dispatcher.LiftedLoop'>

File "numba/tests/test_warnings.py", line 104:
        def do_loop(x):
            <source elided>
            a = {} # noqa dead
            for i in range(x.shape[0]):
            ^

----------
Function "do_loop" was compiled in object mode without forceobj=True, but has lifted loops.
```

esc (approved):
LGTM -- it passed!

PR #6753, opened 2021-02-24
The check for internal types in RewriteArrayExprs

### LINKED ISSUE #5157: RewriteArrayExprs pass incorrectly rewrites all Array-returning operations
## Reporting a bug

Reproducer:
```python
from numba.extending import (
    overload,
    models,
    lower_builtin,
    register_model,
    make_attribute_wrapper,
    type_callable,
    typeof_impl,
)
from numba import types, cgutils
from numba import njit
import operator
import numpy as np


def make_dummy_type():
    class Dummy(object):
        def __init__(self, value):
            self.value = value

    class DummyType(types.Type):
        def __init__(self):
            super(DummyType, self).__init__(name="Dummy")

    dummy_type = DummyType()

    @register_model(DummyType)
            members = [("value", types.intp)]

    make_attribute_wrapper(DummyType, "value", "value")

    @type_callable(Dummy)
    def type_dummy(context):
        def typer(value):
            return dummy_type

        return typer

    @lower_builtin(Dummy, types.intp)
        typ = sig.return_type
        [value] = args
        dummy.value = value
        return dummy._getvalue()

    @typeof_impl.register(Dummy)
    def typeof_dummy(val, c):
        return DummyType()

    # Store attributes
    return Dummy, DummyType


@overload(operator.add)
def ov_add_dummy(lhs, rhs):
    if isinstance(lhs, DummyType) and isinstance(rhs, DummyType):
        print("===", lhs, rhs)

        def imp(lhs, rhs):
            return np.array([lhs.value, rhs.value])

        return imp


@njit
def foo(x, y):
    return Dummy(x) + Dummy(y)


r = foo(1, 2)
print(r)
```

The above raises:
```pytb
numba.errors.TypingError: Failed in nopython mode pipeline (step: nopython mode backend)
No conversion from array(int64, 1d, C) to int64 for '$8return_value.3', defined at None

File "array_expr_issue.py", line 68:
def foo(x, y):
    return Dummy(x) + Dummy(y)
    ^

[1] During: typing of assignment at /Users/siu/dev/numba-git/array_expr_issue.py (68)

File "array_expr_issue.py", line 68:
def foo(x, y):
    return Dummy(x) + Dummy(y)
    ^

[1] During: lowering "$14binary_add.6 = arrayexpr(expr=(<built-in function add>, [Var($6call_function.2, array_expr_issue.py:68), Var($12call_function.5, array_expr_issue.py:68)]), ty=array(int64, 1d, C))" at /Users/siu/dev/numba-git/array_expr_issue.py (68)
```

Expected Behavior:
- `RewriteArrayExprs` pass should not make `arrayexpr` node for the `Dummy(x) + Dummy(y)` expression.
- Disabling the pass will allow the code run correctly.

Alexander-Makaryev:
Fixes #5157 

Please look at issue for more details.

The implementation is based on comment from @stuartarchibald :
https://github.com/numba/numba/issues/5157#issuecomment-699958032

### DIFF
--- a/numba/np/ufunc/array_exprs.py
+++ b/numba/np/ufunc/array_exprs.py
@@ -78,12 +78,8 @@ class RewriteArrayExprs(rewrites.Rewrite):
                 expr.fn in npydecl.supported_array_operators)):
             # It is an array operator that maps to a ufunc.
             # check that all args have internal types
-            all_internal = True
-            for var in expr.list_vars():
-                if not self.typemap[var.name].is_internal:
-                    all_internal = False
-                    break
-            if all_internal:
+            if all(self.typemap[var.name].is_internal
+                   for var in expr.list_vars()):
                 array_assigns[target_name] = instr
 
         elif ((expr_op == 'call') and (expr.func.name in self.typemap)):
--- a/numba/np/ufunc/array_exprs.py
+++ b/numba/np/ufunc/array_exprs.py
@@ -77,7 +77,14 @@ class RewriteArrayExprs(rewrites.Rewrite):
         if ((expr_op in ('unary', 'binop')) and (
                 expr.fn in npydecl.supported_array_operators)):
             # It is an array operator that maps to a ufunc.
-            array_assigns[target_name] = instr
+            # check that all args have internal types
+            all_internal = True
+            for var in expr.list_vars():
+                if not self.typemap[var.name].is_internal:
+                    all_internal = False
+                    break
+            if all_internal:
+                array_assigns[target_name] = instr
 
         elif ((expr_op == 'call') and (expr.func.name in self.typemap)):
             # It could be a match for a known ufunc call.
--- a/numba/tests/test_array_exprs.py
+++ b/numba/tests/test_array_exprs.py
@@ -24,6 +24,7 @@ import shutil
 import uuid
 import os
 import sys
+import textwrap
 
 import unittest
 
@@ -648,13 +649,13 @@ class TestExternalTypes(MemoryLeakMixin, unittest.TestCase):
     """ Tests RewriteArrayExprs with external (user defined) types,
     see #5157"""
 
-    source_lines = """
-from numba.core import types
+    source_lines = textwrap.dedent("""
+        from numba.core import types
 
-class FooType(types.Type):
-    def __init__(self):
-        super(FooType, self).__init__(name='Foo')
-"""
+        class FooType(types.Type):
+            def __init__(self):
+                super(FooType, self).__init__(name='Foo')
+        """)
 
     # copied from test_types
     @contextlib.contextmanager
--- a/numba/tests/test_array_exprs.py
+++ b/numba/tests/test_array_exprs.py
@@ -5,9 +5,26 @@ import numpy as np
 
 from numba import njit, vectorize
 from numba import typeof
-from numba.core import utils, types, typing, ir, compiler, cpu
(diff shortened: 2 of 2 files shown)


numba/tests/test_array_exprs.py
```diff
@@ -627,5 +644,127 @@ def do_call(x, y):
         self.assertTrue(isinstance(oty.type.dtype, types.Float))
 
 
+class TestExternalTypes(MemoryLeakMixin, unittest.TestCase):
+    """ Tests RewriteArrayExprs with external (user defined) types,
+    see #5157"""
+
+    source_lines = """
```
esc, line 651: would suggest to use `textwrap.dedent` for better code alignment here: https://docs.python.org/3/library/textwrap.html#textwrap.dedent

Alexander-Makaryev: done
thanks

Alexander-Makaryev:
@stuartarchibald 
Looks like that we got some failure in CI tests that are not related to this PR.
Previous run of CI was green (https://dev.azure.com/numba/numba/_build/results?buildId=8051&view=results) and there are no significant changes in commits since that time.
What should I do? Do I need to report it somehow or restart tests in this PR?

stuartarchibald:
@Alexander-Makaryev I've restarted the build, and am working on a fix, it's not this PR at fault. Once a fix for it makes it into mainline the CI builders will pick it up and it shouldn't happen again (at least for this particular issue).

numba/np/ufunc/array_exprs.py
```diff
@@ -77,7 +77,14 @@ def _match_array_expr(self, instr, expr, target_name):
...
                 expr.fn in npydecl.supported_array_operators)):
             # It is an array operator that maps to a ufunc.
-            array_assigns[target_name] = instr
+            # check that all args have internal types
+            all_internal = True
+            for var in expr.list_vars():
+                if not self.typemap[var.name].is_internal:
+                    all_internal = False
+                    break
+            if all_internal:
```
stuartarchibald, line 86:
```suggestion
            # check that all args have internal types
            if all(self.typemap[var.name].is_internal
                   for var in expr.list_vars()):
```
think this is equivalent?

numba/tests/test_array_exprs.py
```diff
@@ -627,5 +645,127 @@ def do_call(x, y):
...
+                f.write(lines)
+            # Add test_module to sys.path so it can be imported
+            sys.path.insert(0, tempdir)
+            test_module = importlib.import_module(temp_module_name)
+            yield test_module
+        finally:
+            sys.modules.pop(temp_module_name, None)
+            sys.path.remove(tempdir)
+            shutil.rmtree(tempdir)
+
```
stuartarchibald, line 684: It'd be nice to refactor this into `numba.tests.support`, but perhaps not for this PR.

stuartarchibald (commented):
Thanks for the patch and for fixing this issue, just one minor comment to resolve else looks good.

numba/np/ufunc/array_exprs.py
Alexander-Makaryev: @stuartarchibald 
done
thanks

numba/tests/test_array_exprs.py
stuartarchibald: https://github.com/numba/numba/issues/6778 tracks

stuartarchibald (approved):
Thanks for the fix @Alexander-Makaryev, looks good.

PR #6755, opened 2021-02-24
install llvmlite from numba/label/dev

esc:
The `numba/label/dev` channel is the output target of development
(nightly) builds originating from the Anaconda internal build farm. Hence this is
where any CI should be looking to get the latest version of llvmlite.


sklam (approved):
Checked in #6756

PR #6750, opened 2021-02-23
Bump to llvmlite 0.37 series

sklam:
similar to https://github.com/numba/numba/pull/6463

PR #6756, opened 2021-02-24
Cross 6750 6755

esc:
#6750 + #6755


esc:
The typeguard fail is fixed by #6751 the osx test is failing due to a missing build from the Anaconda internal build farm.

esc:
#6750 and #6755 are both merged, closing.

PR #6760, opened 2021-02-25
Fix scheduler bug where it rounds to 0 divisions for a chunk.

### LINKED ISSUE #6749: Parallel accelerator yields incorrect results for high dimensional arrays
Using the parallel accelerator on high-dimensional arrays produces incorrect results. Here is a minimal example, but the same issues occurs for other functions.
```
import numpy as np
from numba import njit

@njit(parallel=True)
def numba_fun(x):
    return x * 1.0

x = np.ones([2, 2, 2, 2, 2, 15])
y = numba_fun(x)
incorrect = np.abs(x - y) > 1e-6

print(incorrect.sum())
print(y.min())
print(y.max())
print(y[incorrect])
```
In my environment (numba 0.51.2, numpy 1.19.2, python 3.8.5) I get 384 incorrect elements of `y`, ranging from `-1.36e+57` to `4.09e+179`.

The problem is fixed if I either (i) disable the parallel accelerator, or (ii) reduce the dimensions of `x` (e.g, setting `x = np.ones([2, 2, 2, 2, 15])`), or (iii) flatten `x` (e.g, setting `x = np.ones([2, 2, 2, 2, 2, 15]).ravel()`).

On the other hand, reducing the number of elements in the last dimensions completely freezes the execution (`x = np.ones([2, 2, 2, 2, 2, 10])`)

Edit: just updated to numba 0.53.0rc1 and can confirm that I continue to see the same error

DrTodd13:
Resolves #6749

### DIFF
--- a/numba/np/ufunc/gufunc_scheduler.cpp
+++ b/numba/np/ufunc/gufunc_scheduler.cpp
@@ -202,9 +202,9 @@ void divide_work(const RangeActual &full_iteration_space,
         } else {
             // We allocate the remaining threads proportionally to the ratio of the current dimension length to the total.
             divisions_for_this_dim = intp(guround(num_threads * ((float)dims[index].length / total_len)));
-			if (divisions_for_this_dim < 1) {
-				divisions_for_this_dim = 1;
-			}
+            if (divisions_for_this_dim < 1) {
+                        divisions_for_this_dim = 1;
+            }
         }
 
         // These are used to divide the iteration space.
--- a/numba/np/ufunc/gufunc_scheduler.cpp
+++ b/numba/np/ufunc/gufunc_scheduler.cpp
@@ -202,6 +202,9 @@ void divide_work(const RangeActual &full_iteration_space,
         } else {
             // We allocate the remaining threads proportionally to the ratio of the current dimension length to the total.
             divisions_for_this_dim = intp(guround(num_threads * ((float)dims[index].length / total_len)));
+			if (divisions_for_this_dim < 1) {
+				divisions_for_this_dim = 1;
+			}
         }
 
         // These are used to divide the iteration space.
--- a/numba/tests/test_parfors.py
+++ b/numba/tests/test_parfors.py
@@ -1843,7 +1843,7 @@ class TestParfors(TestParforsBase):
     def test_high_dimension1(self):
         # issue6749
         def test_impl(x):
-            return x * 1.0
+            return x * 5.0
         x = np.ones((2, 2, 2, 2, 2, 15))
         self.check(test_impl, x)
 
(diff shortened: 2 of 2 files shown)


numba/np/ufunc/gufunc_scheduler.cpp
```diff
@@ -202,6 +202,9 @@ void divide_work(const RangeActual &full_iteration_space,
         } else {
             // We allocate the remaining threads proportionally to the ratio of the current dimension length to the total.
             divisions_for_this_dim = intp(guround(num_threads * ((float)dims[index].length / total_len)));
+			if (divisions_for_this_dim < 1) {
+				divisions_for_this_dim = 1;
+			}
```
stuartarchibald, line 207:
```suggestion
            if (divisions_for_this_dim < 1) {
                        divisions_for_this_dim = 1;
            }
```

numba/tests/test_parfors.py
```diff
@@ -1839,6 +1839,14 @@ def test_impl(a):
         x = np.ones((3,3))
         self.check(test_impl, x)
 
+    @skip_parfors_unsupported
+    def test_high_dimension1(self):
+        # issue6749
+        def test_impl(x):
+            return x * 1.0
```
stuartarchibald, line 1846:
```suggestion
            return x * 5.0
```
Suggest altering the multiplier given the input is all `1.0` else there's no differentiation between it working and a NOP?

stuartarchibald:
Manually testing against the sample code in #6749... mainline does this:
```
num_dim = 6
ranges = ([0, 1], [0, 1], [0, 1], [0, 1], [0, 1], [0, 14], )
num_threads = 4
sched[0] = 1 0 1 0 1 0 1 0 -787390336 271947308 -463249520 271765979 
sched[1] = -769903936 -167047680 307827184 -463126000 0 316800560 -332696768 -787390336 317704688 272245250 0 317704688 
sched[2] = 0 0 -332320592 -770377632 -787390336 316800560 307827184 318603472 -787390336 272247143 -787390336 -463126000 
sched[3] = -770377632 318604000 -812855408 272305680 307819792 315285840 307819864 318603832 0 271783384 32 1241058768 
```
this patch does:
```
num_dim = 6
ranges = ([0, 1], [0, 1], [0, 1], [0, 1], [0, 1], [0, 14], )
num_threads = 4
sched[0] = 0 0 0 0 0 0 1 1 1 1 0 6
sched[1] = 0 0 0 0 1 0 1 1 1 1 1 6
sched[2] = 0 0 0 0 0 7 1 1 1 1 0 14
sched[3] = 0 0 0 0 1 7 1 1 1 1 1 14
```
which I think confirms it fixes the issue.

stuartarchibald (approved):
Thanks for the patch!

PR #6762, opened 2021-02-25
Glue wrappers to create @overload from split typing and lowering.

stuartarchibald:
As title.

### DIFF
--- a/numba/core/overload_glue.py
+++ b/numba/core/overload_glue.py
@@ -101,97 +101,57 @@ class OverloadWrapper(object):
 
         @overload(self._function, strict=False)
         def ol_generated(*ol_args, **ol_kwargs):
-            if ol_kwargs:
-                # TODO: Join these branchess
-                def body(tyctx):
-                    msg = f"No typer registered for {self._function}"
-                    assert self._TYPER is not None, msg
-                    typing = self._TYPER(tyctx)
-                    sig = typing.apply(ol_args, ol_kwargs)
-                    if sig is None:
-                        from numba.core import errors
-                        # TODO: Something about this, it might be possible to
-                        # fish out the actual arguments. Needs kwargs plugging
-                        # in to the error message too.
-                        err = ("No match. No implementation of %s found for "
-                               "argument type(s) %s" % (self._function,
-                                                        str(ol_args)))
-                        raise errors.TypingError(err)
-                    if self._selector is None:
-                        self._assemble()
-                    lowering = self._selector.find(sig.args)
-                    msg = (f"Could not find implementation to lower {sig} for ",
-                           f"{self._function}")
-                    assert lowering is not None, msg
-                    return sig, lowering
 
-                stub = stub_generator(len(ol_args), {'body': body}, ol_kwargs)
+            def body(tyctx):
+                msg = f"No typer registered for {self._function}"
+                assert self._TYPER is not None, msg
+                typing = self._TYPER(tyctx)
+                sig = typing.apply(ol_args, ol_kwargs)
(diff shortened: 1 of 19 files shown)


numba/core/overload_glue.py
```diff
@@ -0,0 +1,215 @@
...
+            wrapper = OverloadWrapper(func)
+            self._registered[func] = wrapper
+            return wrapper
+
+
+overload_glue = Gluer()
+del Gluer
+
+
+def glue_wrap(x):
```
DrTodd13, line 214: It seems there's a lot of boilerplate use of glue_wrap.  I didn't see any exceptions to being able to do the following:

```
def glue_wrap(*args):
    return overload_glue(args[0]).wrap_typing(*args))
```

Am I missing something?

DrTodd13 (commented):
Also, if we want to modify the wrapper for a given target, that capability is not in this PR right?  We'd have to do a follow-up PR to enable that?  Are there any infrastructure issues that would stop that from being easily bolted on to this PR?

numba/core/overload_glue.py
stuartarchibald: Thanks for looking at this. At present the API isn't in any way designed or finished :) I wrote it in bits, mostly as functions, to stick the various pieces together as needed and then stuck them into a class to do the assembling. Now that this is closer to actually working correctly I'll go back through it and try and make it less like something evolved from pieces. There's also that `@glue_wrap(overload_glue(np.array).wrap_typing(np.array))` says `np.array` twice, which is not ideal.

Some of the weirdness is also because I wrote it all with Python 3.9 which has https://www.python.org/dev/peps/pep-0614/ and then had to do a workaround.

As to `@glue_wrap`, it needs to work for both typing and lowering so there'd need to be two versions, but if it's possible to get to a place where the change can be done with `sed` then I think that'd be good!

stuartarchibald: In https://github.com/numba/numba/pull/6762/commits/f03432863868ccba32e160bb411ef53c23fd22a6 I've made the API look like this (using `np.array` as an example):

This is how to wrap the typing component, `@glue_typing` is a text-only replacement for `@infer_global`.
```python
@glue_typing(np.array)
class NpArray(CallableTemplate):
    ... #details
```
and this is how to wrap the lowering component, `@glue_lowering` is a text-only replacement for `@lower_builtin`.
```python
@glue_lowering(np.array, types.Any)
@glue_lowering(np.array, types.Any, types.DTypeSpec)
@glue_lowering(np.array, types.Any, types.StringLiteral)
def np_array(context, builder, sig, args):
    ... # details
```

stuartarchibald:
Buildfarm ID: `numba_smoketest_cpu_yaml_11`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cpu_yaml_11`.

This passed.

numba/core/overload_glue.py
```diff
@@ -0,0 +1,184 @@
...
+import types as pytypes
+import textwrap
+
+
+class _OverloadWrapper(object):
+    """This class does all the work of assembling and registering wrapped split
+    implementations.
+    """
+
+    def __init__(self, function=None):
```
sklam, line 14: why default to None when the next line is forbidding it?

stuartarchibald: I don't think it needs to be a kwarg any more.

sklam (approved):
Thanks for the patch!

This will allow us to move to the new style (overload based) implementation more quickly and allow target extensions to specialize on more things.

PR #6761, opened 2021-02-25
Fix llvmlite dep

esc:
The build farm is reporting: 

```
[2021-02-25 00:29:11,859] {docker_operator.py:265} INFO - /opt/conda/envs/testenv_5bfbfbf2-26a7-4afe-907e-39fa2477747c/lib/python3.8/site-packages/numba/tests/test_entrypoints.py:65: UserWarning: Numba extension module '_test_numba_extension' failed to load due to 'VersionConflict((llvmlite 0.37.0.dev0 (/opt/conda/envs/testenv_5bfbfbf2-26a7-4afe-907e-39fa2477747c/lib/python3.8/site-packages), Requirement.parse('llvmlite<0.37,>=0.36.0rc1')))'.
[2021-02-25 00:29:11,859] {docker_operator.py:265} INFO - entrypoints.init_all()
```

This is an experimental PR to try to fix that.

PR #6763, opened 2021-02-25
use an alternative constraint for the conda packages

esc:
To exclude `0.37.0dev0`.


esc:
Second attempt for #6761

esc:
Looks good, the resulting conda package will install `llvmlite-0.36.0rc1`, even if `llvmlite-0.37.0dev` is avail.

PR #6769, opened 2021-02-26
CUDA: Replace `CachedPTX` and `CachedCUFunction` with `CUDACodeLibrary` functionality

gmarkall:
Note: This PR is based on #6735 - the latest four commits are unique to this PR and can each be examined individually to understand the changes here.

This commit removes `CachedPTX` and `CachedCUFunction`, replacing them with additions to the `CUDACodeLibrary`. This aligns the CUDA target with more of the infrastructure in the core of Numba, simplifies the organization of code generation and caching, and provides some other benefits - generally it lays the foundation for implementations of on-disk caching of compiled kernels, ahead-of-time compilation, and improving debugging support for CUDA kernels. This description is expanded on in the final commit's message.

This PR also includes some other small changes and fixes - see individual commit messages for details.

### DIFF
--- a/numba/cuda/codegen.py
+++ b/numba/cuda/codegen.py
@@ -44,7 +44,7 @@ def disassemble_cubin(cubin):
             os.unlink(fname)
 
 
-class CUDACodeLibrary(CodeLibrary, serialize.ReduceMixin):
+class CUDACodeLibrary(serialize.ReduceMixin, CodeLibrary):
     """
     The CUDACodeLibrary generates PTX, SASS, cubins for multiple different
     compute capabilities. It also loads cubins to multiple devices (via
--- a/numba/cuda/codegen.py
+++ b/numba/cuda/codegen.py
@@ -1,27 +1,186 @@
 from llvmlite import binding as ll
 from llvmlite.llvmpy import core as lc
 
+from numba.core import config, serialize
 from numba.core.codegen import Codegen, CodeLibrary
-from .cudadrv import nvvm
+from .cudadrv import devices, driver, nvvm
+
+import ctypes
+import numpy as np
+import os
+import subprocess
+import tempfile
 
 
 CUDA_TRIPLE = 'nvptx64-nvidia-cuda'
 
 
-class CUDACodeLibrary(CodeLibrary):
+def disassemble_cubin(cubin):
+    # nvdisasm only accepts input from a file, so we need to write out to a
+    # temp file and clean up afterwards.
+    fd = None
+    fname = None
+    try:
+        fd, fname = tempfile.mkstemp()
+        with open(fname, 'wb') as f:
+            f.write(cubin)
 
-    def __init__(self, codegen, name):
+        try:
+            cp = subprocess.run(['nvdisasm', fname], check=True,
+                                stdout=subprocess.PIPE,
+                                stderr=subprocess.PIPE)
+        except FileNotFoundError as e:
+            if e.filename == 'nvdisasm':
+                msg = ("nvdisasm is required for SASS inspection, and has not "
+                       "been found.\n\nYou may need to install the CUDA "
+                       "toolkit and ensure that it is available on your "
+                       "PATH.\n")
+                raise RuntimeError(msg)
+        return cp.stdout.decode('utf-8')
+    finally:
+        if fd is not None:
+            os.close(fd)
+        if fname is not None:
+            os.unlink(fname)
+
+
+class CUDACodeLibrary(CodeLibrary, serialize.ReduceMixin):
+    """
+    The CUDACodeLibrary generates PTX, SASS, cubins for multiple different
+    compute capabilities. It also loads cubins to multiple devices (via
+    get_cufunc), which may be of different compute capabilities.
+    """
+
+    def __init__(self, codegen, name, entry_name=None, max_registers=None,
+                 nvvm_options=None):
+        """
+        codegen:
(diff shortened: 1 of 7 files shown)


numba/cuda/codegen.py
```diff
@@ -1,63 +1,308 @@
...
+                       "been found.\n\nYou may need to install the CUDA "
+                       "toolkit and ensure that it is available on your "
+                       "PATH.\n")
+                raise RuntimeError(msg)
+        return cp.stdout.decode('utf-8')
+    finally:
+        if fd is not None:
+            os.close(fd)
+        if fname is not None:
+            os.unlink(fname)
```
stuartarchibald, line 44: NOTE: this is a direct move of the function from `numba.cuda.compiler`.

numba/cuda/codegen.py
```diff
@@ -1,63 +1,308 @@
...
+                if not fn.is_declaration and fn.linkage != 'external':
+                    fn.linkage = 'linkonce_odr'
+
+        self._finalized = True
+
+    def _reduce_states(self):
+        """
+        Reduce the instance for serialization. We retain the PTX and cubins,
+        but loaded functions are discarded. They are recreated when needed
+        after unserialization.
```
stuartarchibald, line 244:
```suggestion
        after deserialization.
```

numba/cuda/compiler.py
```diff
@@ -293,13 +291,15 @@ def inspect_ptx(self, args, nvvm_options={}):
...
-        nvvm_options : dict; optional
-            See `CompilationUnit.compile` in `numba/cuda/cudadrv/nvvm.py`.
 
         Returns
         -------
         ptx : bytes
         """
+        if nvvm_options:
+            msg = 'nvvm_options kwarg for inspect_ptx is deprecated'
+            warn(msg, category=NumbaDeprecationWarning)
```
stuartarchibald, line 301: Want to add this to the deprecation notice doc?

numba/cuda/compiler.py
```diff
@@ -677,27 +568,27 @@ def regs_per_thread(self):
...
         '''
-        return self._func.get().attrs.regs
+        return self._codelibrary.get_cufunc().attrs.regs
 
     def inspect_llvm(self):
         '''
         Returns the LLVM IR for this kernel.
         '''
-        return str(self._func.ptx.llvmir)
+        return self._codelibrary.get_asm_str()
```
stuartarchibald, line 577: Is this right? Should it be `get_llvm_str`?

stuartarchibald (commented):
Thanks for doing this refactoring, it's made the CUDA target much clearer and helped tidy up the CPU side. On inspection it looks largely good, just a few minor comments to resolve. I think this should have both a CPU and CUDA test run in the build farm just to make sure there's no surprises.

@sklam you might want to take a look at this too?

Thanks again!

stuartarchibald:
Also, conflicts need resolving, thanks!

numba/cuda/compiler.py
gmarkall: I decided that deprecating it was a mistake (from the commit message un-deprecating it):

```
The idea behind deprecating this was to prevent a mismatch between
compiled code that is executed and code produced by the `inspect_ptx()`
function - however, as device functions aren't materialized standalone
(they are always part of a kernel), there is no risk of this mismatch.
Instead, `inspect_ptx()` stands as a useful mechanism for obtaining PTX
code for a device function if required - noting that `compile_ptx()`
does provide a better interface, so this whole function may be
deprecated / removed in future.
```

numba/cuda/compiler.py
gmarkall: It's wrong - it should be as you suggest - now fixed.

gmarkall:
@stuartarchibald Many thanks for the review - conflicts resolved and comments now addressed.

stuartarchibald (dismissed):

stuartarchibald:
Leaving the waiting-on-reviewer label present as @sklam might want to take a look too.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_26`.

numba/cuda/codegen.py
```diff
@@ -1,27 +1,186 @@
...
+        return cp.stdout.decode('utf-8')
+    finally:
+        if fd is not None:
+            os.close(fd)
+        if fname is not None:
+            os.unlink(fname)
+
 
-    def __init__(self, codegen, name):
+class CUDACodeLibrary(CodeLibrary, serialize.ReduceMixin):
```
sklam, line 47: Although it won't affect the code here, mixins should go to the left. The MRO goes left to right. If `CodeLibrary` defines `__reduce__`, it will take priority over the one in the `ReduceMixin`.

sklam (commented):
Just one nitpick on the baseclass ordering for mixins

numba/cuda/codegen.py
gmarkall: Many thanks - I've swapped these round now.

sklam (approved):
Thanks for the patch!

stuartarchibald:
Think this is just waiting for a build farm run now.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_30`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_30`.


Passed.

PR #6770, opened 2021-02-26
Support changing dispatcher options via a with context

sklam:
Example usage is showed in https://gist.github.com/sklam/b0f145584107f5fbd346eb1a805ec29a.

### DIFF
--- a/numba/_dispatcher.cpp
+++ b/numba/_dispatcher.cpp
@@ -601,6 +601,35 @@ find_named_args(Dispatcher *self, PyObject **pargs, PyObject **pkws)
     return 0;
 }
 
+
+/*
+ * Management of thread-local random state.
+ */
+
+#ifdef _MSC_VER
+#define THREAD_LOCAL(ty) __declspec(thread) ty
+#else
+/* Non-standard C99 extension that's understood by gcc and clang */
+#define THREAD_LOCAL(ty) __thread ty
+#endif
+
+static THREAD_LOCAL(bool) use_tls_target_stack;
+
+
+struct raii_use_tls_target_stack {
+    bool old_setting;
+
+    raii_use_tls_target_stack(bool new_setting)
+        : old_setting(use_tls_target_stack)
+    {
+        use_tls_target_stack = new_setting;
+    }
+
+    ~raii_use_tls_target_stack() {
+        use_tls_target_stack = old_setting;
+    }
+};
+
 static PyObject*
 Dispatcher_call(Dispatcher *self, PyObject *args, PyObject *kws)
 {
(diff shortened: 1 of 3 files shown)


numba/_dispatcher.cpp
```diff
@@ -601,6 +601,35 @@ find_named_args(Dispatcher *self, PyObject **pargs, PyObject **pkws)
...
+ */
+
+#ifdef _MSC_VER
+#define THREAD_LOCAL(ty) __declspec(thread) ty
+#else
+/* Non-standard C99 extension that's understood by gcc and clang */
+#define THREAD_LOCAL(ty) __thread ty
+#endif
+
+static THREAD_LOCAL(bool) use_tls_target_stack;
```
sklam, line 616: This is copied from random.c. TODO: deduplicate

sklam:
I'm also considering an alternative design that allows more flexible retargeting at https://github.com/sklam/numba/tree/wip/dispstack_alt1, which will allow for example usage like: https://gist.github.com/sklam/1eb89424dda59c8c4a7cb1f1ed2385be. The example can retarget a CPU dispatcher to a CUDA dispatcher with a specific launch configuration.

sklam:
@diptorupd, @PokhodenkoSA, here's an example showing a way to tell apart unspecified target: https://gist.github.com/sklam/e473b3731759bd360ad17710f388643a. It's based on a patch at https://github.com/sklam/numba/commit/98d916f0670e5cf753c50a85f9dd1fec47171d3e

numba/core/dispatcher.py
```diff
@@ -23,6 +25,31 @@
...
+        self._stack = tls_stack
+
+    def push(self, state):
+        self._stack.append(state)
+
+    def pop(self):
+        return self._stack.pop()
+
+    def get(self):
+        return self._stack[0]
```
PokhodenkoSA, line 47: push/pop works with the tail but get works with head. Is it correct?

numba/_dispatcher.cpp
```diff
@@ -614,6 +643,20 @@ Dispatcher_call(Dispatcher *self, PyObject *args, PyObject *kws)
     PyThreadState *ts = PyThreadState_Get();
     PyObject *locals = NULL;
 
+    // Check TLS target stack
+    if (use_tls_target_stack) {
+        raii_use_tls_target_stack turn_off(false);
```
PokhodenkoSA, line 648: Why `use_tls_target_stack` is turned off? Does it mean that nested functions will not use target stack?

numba/core/dispatcher.py
sklam: that's a mistake. it should be getting from the tail.

numba/_dispatcher.cpp
sklam: Nested function will not go through this part of the code. The dispatcher is the bridge connecting the Python interpreter code into the compiled code. The change of the target will only need to happen once. 

In this branch, the dispatcher will go into `_call_tls_target` where a new dispatcher will be invoked to compile and execute for the specific target. By turning off the`use_tls_target_stack` here, we stop the new dispatcher from calling `_call_tls_target` unnecessarily.

PR #6776, opened 2021-03-01
Fix issue with TBB interface causing warnings and parfors counting them

stuartarchibald:
As title.

### DIFF
--- a/numba/tests/test_parfors.py
+++ b/numba/tests/test_parfors.py
@@ -3499,6 +3499,11 @@ class TestParforsMisc(TestParforsBase):
 
         with warnings.catch_warnings(record=True) as raised_warnings:
             warnings.simplefilter('always')
+            # Filter out warnings about TBB interface mismatch
+            warnings.filterwarnings(action='ignore',
+                                    message=r".*TBB_INTERFACE_VERSION.*",
+                                    category=numba.errors.NumbaWarning,
+                                    module=r'numba\.np\.ufunc\.parallel.*')
             cfunc()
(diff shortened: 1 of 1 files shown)


numba/tests/test_parfors.py
```diff
@@ -3499,6 +3499,11 @@ def pyfunc():
 
         with warnings.catch_warnings(record=True) as raised_warnings:
             warnings.simplefilter('always')
```
esc, line 3501: My "guess" would be, that this reset's all warnings, so they are always shown (even if we asked for something else via an environment variable or on the command line). So my question is, can this line not just be deleted?

esc: Otherwise, this seems sane. It means parfors won't fall over because too many tbb interface version warnings are raised.

stuartarchibald: IIRC `'always'` means keep on printing them even if you've seen them before, this PR is just adding a new rule to ignore a particular class of warning.

sklam: I can confirm that the following `filterwarnings` is appending a filter that will suppress the TBB warning. I did the same in the typeguard PR.

sklam: ref: https://github.com/numba/numba/pull/6751/commits/a4ff69b49e5675b94ed883068775b59427653548

esc (approved):
Merge and see

sklam (approved):
LGTM!

PR #6780, opened 2021-03-01
Wip/pr/6767

sklam:


stuartarchibald:
am I right in reading this as "if you print a line out it works, remove the print and it doesn't"? suspect there's something else going on...

PR #6786, opened 2021-03-02
Fix gufunc kwargs support

### LINKED ISSUE #6773: Regression in 0.53rc2: axis= parameter in gufuncs
This broke in 0.53rc2; it was working in 0.52:

```python
import numba
import numpy

@numba.guvectorize(["f8[:],f8[:]"], "(n)->(n)")
def my_cumsum(x, res):
    acc = 0
    for i in range(x.shape[0]):
        acc += x[i]
        res[i] = acc


x = numpy.ones((20, 30))
y = my_cumsum(x, axis=0)
```
Output:
```
TypeError: __call__() got an unexpected keyword argument 'axis'
```

sklam:
Fix #6773

Fixes kwargs support for gufunc.

### DIFF
--- a/numba/np/ufunc/gufunc.py
+++ b/numba/np/ufunc/gufunc.py
@@ -113,11 +113,14 @@ class GUFunc(object):
 
         return types.none(*l)
 
-    def __call__(self, *args):
+    def __call__(self, *args, **kwargs):
         # If compilation is disabled OR it is NOT a dynamic gufunc
         # call the underlying gufunc
         if self._frozen or not self.is_dynamic:
-            return self.ufunc(*args)
+            return self.ufunc(*args, **kwargs)
+        elif "out" in kwargs:
+            # If "out" argument is supplied
+            args += (kwargs.pop("out"),)
 
         if self._num_args_match(*args) is False:
             # It is not allowed to call a dynamic gufunc without
@@ -137,4 +140,4 @@ class GUFunc(object):
             sig = self._get_signature(*args)
             self.add(sig)
             self.build_ufunc()
-        return self.ufunc(*args)
+        return self.ufunc(*args, **kwargs)
--- a/numba/np/ufunc/gufunc.py
+++ b/numba/np/ufunc/gufunc.py
@@ -136,7 +136,6 @@ class GUFunc(object):
         # at this point we know the gufunc is a dynamic one
         ewise = self._get_ewise_dtypes(args)
         if not (self.ufunc and ufunc_find_matching_loop(self.ufunc, ewise)):
-            self._is_dynamic = True
             sig = self._get_signature(*args)
             self.add(sig)
             self.build_ufunc()
--- a/numba/tests/npyufunc/test_gufunc.py
+++ b/numba/tests/npyufunc/test_gufunc.py
@@ -63,6 +63,7 @@ class TestGUFunc(TestCase):
         np.testing.assert_equal(out, x * x + x)
 
     def test_axis(self):
+        # issue https://github.com/numba/numba/issues/6773
         @guvectorize(["f8[:],f8[:]"], "(n)->(n)")
         def my_cumsum(x, res):
             acc = 0
@@ -163,6 +164,7 @@ class TestDynamicGUFunc(TestCase):
             sum_row(inp)
 
     def test_axis(self):
+        # issue https://github.com/numba/numba/issues/6773
         @guvectorize("(n)->(n)")
         def my_cumsum(x, res):
             acc = 0
--- a/numba/tests/npyufunc/test_gufunc.py
+++ b/numba/tests/npyufunc/test_gufunc.py
@@ -62,6 +62,24 @@ class TestGUFunc(TestCase):
 
         np.testing.assert_equal(out, x * x + x)
 
+    def test_axis(self):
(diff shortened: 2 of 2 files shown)


esc:
Buildfarm ID: `numba_smoketest_cpu_yaml_9`

numba/tests/npyufunc/test_gufunc.py
```diff
@@ -62,6 +62,24 @@ def test_ufunc_like(self):
 
         np.testing.assert_equal(out, x * x + x)
 
+    def test_axis(self):
+        @guvectorize(["f8[:],f8[:]"], "(n)->(n)")
```
esc, line 66: Maybe include a comment here pointing to: https://github.com/numba/numba/issues/6773 -- IIRC this is done quite often for tests that expose specific bugs in the Numba code-base.

numba/np/ufunc/gufunc.py
```diff
@@ -133,8 +136,7 @@ def __call__(self, *args):
         # at this point we know the gufunc is a dynamic one
         ewise = self._get_ewise_dtypes(args)
         if not (self.ufunc and ufunc_find_matching_loop(self.ufunc, ewise)):
-            self._is_dynamic = True
```
esc, line 136: This is odd. Is this fixing an oversight unrelated to kwarg processing?

numba/tests/npyufunc/test_gufunc.py
```diff
@@ -140,6 +162,24 @@ def sum_row(inp, out):
         with self.assertRaisesRegex(TypeError, msg):
             sum_row(inp)
 
+    def test_axis(self):
+        @guvectorize("(n)->(n)")
+        def my_cumsum(x, res):
+            acc = 0
```
esc, line 168: Same comment as above.

esc (changes requested):
One question, one suggestion.

esc:
So, this fails on the farm with: `[2021-03-03 08:19:54,958] {docker_operator.py:262} INFO - conda_build.exceptions.DependencyNeedsBuildingError: Unsatisfiable dependencies for platform linux-ppc64le: {"llvmlite[version='>=0.37.0dev0,<0.37']"}` -- I presume this is because there is no correct `llvmlite` on the `numba/label/main` channel, but only on the `numba/label/dev`. I'll have to check.

esc:
My assumption turns out to be correct. `conda build -c numba/label/main --no-test --python=3.7 --numpy=1.14 buildscripts/condarecipe.local` is the build command and it doesn't include `numba/label/main` as channel.

esc:
OK, after adding `numba/label/dev` to the `conda build` command, I still get:

`[2021-03-03 03:12:26,372] {bash_operator.py:137} INFO - conda_build.exceptions.DependencyNeedsBuildingError: Unsatisfiable dependencies for platform win-32: {"llvmlite[version='>=0.37.0dev0,<0.37']"}`

esc:
it's strange, because the 0.37.0dev0 does seem to be available on anaconda.org:

https://anaconda.org/numba/llvmlite/files?version=0.37.0dev0

esc:
OK, it may be fixed, linux, windows and OSX have their own config for the channels to use during build.

esc:
Happy to report this passed on the Anaconda internal build farm.

numba/np/ufunc/gufunc.py
sklam: Correct

numba/tests/npyufunc/test_gufunc.py
sklam: will do

sklam: done in https://github.com/numba/numba/pull/6786/commits/20c272d5d387c477197cd3bee18abc6adc445a32

numba/tests/npyufunc/test_gufunc.py
sklam: done in https://github.com/numba/numba/pull/6786/commits/20c272d5d387c477197cd3bee18abc6adc445a32

esc (approved):

numba/tests/npyufunc/test_gufunc.py
```diff
@@ -140,6 +163,25 @@ def sum_row(inp, out):
         with self.assertRaisesRegex(TypeError, msg):
             sum_row(inp)
 
+    def test_axis(self):
+        # issue https://github.com/numba/numba/issues/6773
+        @guvectorize("(n)->(n)")
```
stuartarchibald, line 168: This isn't the exact reproducer from the ticket as there's no type signature provided?

sklam: This is the dynamic gufunc version of it. The exact one is at https://github.com/numba/numba/pull/6786/files#diff-53626a767921908cedfa774172dffeebb669f76bb7ee26af3bc51c37987d5bfdR67, but both version are doing through the same `GUFunc.__call__`. So this is the variant to make sure that the dynamic gufunc is handling the kwargs correctly as well.

stuartarchibald: ah thanks, missed that.

stuartarchibald (approved):

PR #6788, opened 2021-03-03
Add support for fastmath 32-bit floating point divide

testhound:
This pull request adds support for generating more efficient code for 32-bit floating point divides in the presence of fastmath. This should be the final piece to: https://github.com/numba/numba/issues/6183.

### DIFF
--- a/numba/cuda/tests/cudapy/test_fastmath.py
+++ b/numba/cuda/tests/cudapy/test_fastmath.py
@@ -1,5 +1,6 @@
 from numba import cuda, float32
 from math import cos, sin, tan, exp, log, log10, log2, pow
+import numpy as np
 from numba.cuda.testing import CUDATestCase, skip_on_cudasim
 import unittest
 
@@ -7,7 +8,7 @@ import unittest
 class TestFastMathOption(CUDATestCase):
     @skip_on_cudasim('fast divide not available in CUDASIM')
     def test_kernel(self):
-
+        # Test the cast of an int being used in fastmath divide
         def foo(arr, val):
             i = cuda.grid(1)
             if i < arr.size:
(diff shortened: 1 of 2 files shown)


numba/cuda/tests/cudapy/test_fastmath.py
```diff
@@ -5,6 +5,7 @@
 
 
 class TestFastMathOption(CUDATestCase):
+    @skip_on_cudasim('fast divide not available in CUDASIM')
     def test_kernel(self):
```
gmarkall, line 9: For a while I was wondering what this is checking that differs from what `test_divf` checks for below... Eventually I concluded that it tests that casting an integer to a `float32` then using it in a division results in a `div.approx` instruction being emitted - given how finicky casts and the type system can be, I think it's good to keep this check alongside `test_divf` - Could you add a comment here explaining that this is testing the cast of an int being used in the divide, to make it a bit easier for future readers to see the difference?

Alternatively - do you have a different view on the difference between this and `test_divf`? I reached my conclusion by trying to derive meaning from the code (thinking backwards a bit) so maybe there's something else going on here that I haven't followed.

gmarkall (changes requested):
I think the general approach here looks good - I have a few comments about the implementation and testing on the diff. I think with my suggested changes to the `maybe_fast_truediv` function, the following should be true:

- A division by zero involving `float32` operands in a kernel jitted with `debug=True` will raise a `ZeroDivisionError` following the launch.
- No exception should be raised following the launch of a kernel jitted with `fastmath=True, debug=True`.

If this is the case, could you add a test that tests both these conditions hold please?

gmarkall (approved):
Many thanks for the changes - they look good to me. Tests pass locally with hardware and the simulator.

I think the doc failure is unrelated to the contents of this PR - it is either a transient HTTP failure or an issue that needs resolving with the Numba docs.

@esc Could this have a buildfarm run please?

esc:
Buildfarm ID: `numba_smoketest_cuda_yaml_20`

gmarkall:
@esc How did it go? :-)

stuartarchibald:
> @esc How did it go? :-)

Passed.

stuartarchibald (approved):
Thanks for the patch.

stuartarchibald:
Not to resolve in this PR, but this raises an interesting question about flag compatibility, should fastmath and error model be tested to ensure compatibility?

Issue #6797, opened 2021-03-05
Issue when using 'not in' in numba


Hbodor:
## Reporting a bug

- [x] I have tried using the latest released version of Numba (most recent is
 visible in the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [x] I have included a self contained code sample to reproduce the problem.

Hi,

When trying to convert my code to be compatible with numba, I remarked that 'not in' is not working properly in numba. 

Here is an example :

```python3
@njit
def test4():
    l = [ [2 for i in range(0)] for i in range(5)]
    for i in range(5):
      for j in range(i,10):
        j = j%5
        if j not in l[i]:
          l[i].append(j)
    return l
test4()

``` 

this code produces the following error 

```

TypingError: Failed in nopython mode pipeline (step: nopython frontend)
Type of variable '$66compare_op.10' cannot be determined, operation: unary(fn=<built-in function not_>, value=$66compare_op.10), location: <ipython-input-50-a39b79080121> (7)

File "<ipython-input-50-a39b79080121>", line 7:
def test4():
    <source elided>
        j = j%5
        if j not in l[i]:
        ^

```

To solve it I did : 

```python3
@njit
def test4():
    l = [ [2 for i in range(0)] for i in range(5)]
    for i in range(5):
      for j in range(i,10):
        j = j%5
        if j in l[i]:
          pass
        else:
          l[i].append(j)
    return l
test4()
```

which gives the desired output

gmarkall:
I can't reproduce this with Numba master - running:

```python
from numba import njit


@njit
def test4():
    l = [[2 for i in range(0)] for i in range(5)]
    for i in range(5):
        for j in range(i, 10):
            j = j % 5
            if j not in l[i]:
                l[i].append(j)
    return l


test4()
```

results in no output:

```
0 gmarkall@blixen 12:50:06 (main) ~/numbadev/issues/6797
$ python repro.py 
(numba) 
0 gmarkall@blixen 12:50:09 (main) ~/numbadev/issues/6797
$
```

Can you provide the output of `numba -s` please?

Hbodor:
Here is my output for ``` numba -s ```


```

System info:
--------------------------------------------------------------------------------
__Time Stamp__
2021-03-05 13:51:25.447553

__Hardware Information__
Machine                                       : x86_64
CPU Name                                      : haswell
Number of accessible CPU cores                : 2
Listed accessible CPUs cores                  : 0-1
CFS restrictions                              : None
[...]

PR #6799, opened 2021-03-05
Print Numba and llvmlite versions in sysinfo

gmarkall:
This PR adds the version numbers of Numba and llvmlite to the sysinfo tool. There are a couple of reasons I think this is useful:

1. If it's run in an environment that isn't a conda environment, the version number isn't given anywhere - see #6797 for an example of this.
2. Even if Conda is available and reports given versions (e.g. 0.52.0 / 0.35.0) we can't be certain that's what gets loaded on the import of Numba, for example if some pip-based installation has clobbered the installation - this might help make it easier to spot that situation when helping users fix their setups.

Example output on my machine:

```
System info:
__Time Stamp__
Report started (local time)                   : 2021-03-05 14:43:35.224480
UTC start time                                : 2021-03-05 14:43:35.224488
Running time (s)                              : 1.479506

__Hardware Information__
Machine                                       : x86_64
CPU Name                                      : skylake-avx512
CPU Count                                     : 12
Number of accessible CPUs                     : 12
List of accessible CPUs cores                 : 0 1 2 3 4 5 6 7 8 9 10 11
CFS Restrictions (CPUs worth of runtime)      : None

CPU Features                                  : 64bit adx aes avx avx2 avx512bw
                                                avx512cd avx512dq avx512f avx512vl
                                                bmi bmi2 clflushopt clwb cmov cx16
                                                cx8 f16c fma fsgsbase fxsr invpcid
                                                lzcnt mmx movbe pclmul pku popcnt
                                                prfchw rdrnd rdseed rtm sahf sse
                                                sse2 sse3 sse4.1 sse4.2 ssse3
                                                xsave xsavec xsaveopt xsaves

Memory Total (MB)                             : 95270
Memory Available (MB)                         : 90690

__OS Information__
Platform Name                                 : Linux-5.4.0-66-generic-x86_64-with-glibc2.10
Platform Release                              : 5.4.0-66-generic
OS Name                                       : Linux
OS Version                                    : #74~18.04.2-Ubuntu SMP Fri Feb 5 11:17:31 UTC 2021
OS Specific Version                           : ?
Libc Version                                  : glibc 2.27

__Python Information__
Python Compiler                               : GCC 7.3.0
Python Implementation                         : CPython
Python Version                                : 3.8.5
Python Locale                                 : en_GB.UTF-8

__Numba Information__
Numba Version                                 : 0.54.0dev0+96.gff93323e6.dirty

__LLVM Information__
llvmlite Version                              : 0.37.0dev0-18-g78fd512
LLVM Version                                  : 10.0.1
< ... snip ... >
```


sklam:
Good suggestion.

numba/misc/numba_sysinfo.py
```diff
@@ -585,7 +590,11 @@ class DisplaySeqMaps(tuple):
         ("__Python Information__",),
         DisplayMap({k: v for k, v in info.items() if k.startswith('Python')}),
         ("",),
+        ("__Numba Information__",),
+        ("Numba Version", info.get(_numba_version, '?')),
+        ("",),
         ("__LLVM Information__",),
+        ("llvmlite Version", info.get(_llvmlite_version, '?')),
```
stuartarchibald, line 597: Am wondering if Numba, llvmlite and LLVM information ought to go under a single header of e.g. `__Numba Toolchain Versions__` or something? I'm not sure  that `llvmlite` belongs under `LLVM` as is?

stuartarchibald (commented):
Thanks for the patch, it's definitely a good idea to include this separately, one minor comment to resolve else looks good.

numba/misc/numba_sysinfo.py
gmarkall: Yeah, that would be better.

stuartarchibald (approved):

stuartarchibald:
NOTE: tested this manually, output is in the form presented here https://github.com/numba/numba/pull/6799#discussion_r591638807

PR #6801, opened 2021-03-05
ParallelAccelerator no long will convert StaticSetItem to SetItem because record arrays require StaticSetItems.

### LINKED ISSUE #6704: Cannot lower setitem on a record in a parallel loop
## Reporting a bug

When running this code:

```python
import numba
import numpy as np

print(numba.__version__)
# => 0.52.0

state_dtype = np.dtype([
    ('var', np.int32)
 ])

@numba.njit(inline=True)
def inner_function(state):
    state['var'] = 1

@numba.njit(parallel=True)
def outer_function(states):
    for i in numba.prange(1):
        inner_function(states[i])

outer_function(np.ones(shape=1, dtype=state_dtype))
```
numba yields 
```No definition for lowering <built-in function setitem>(Record(var[type=int32;offset=0];4;False), Literal[str](var), Literal[int](1)) -> none```, which seems to be the correct signature and I'm not so sure why there is no lowering available for that situation.

When disabling parallel or inlining the code compiles and produces correct results.

### PR DESCRIPTION
**Pull request by DrTodd13**
Resolves #6704.

### DIFF
--- a/numba/parfors/parfor.py
+++ b/numba/parfors/parfor.py
@@ -1689,6 +1689,9 @@ class ConvertInplaceBinop:
         return self.pass_states.typingctx.resolve_function_type(fnty, tuple(args), {})
 
 
+def get_index_var(x):
+    return x.index if isinstance(x, ir.SetItem) else x.index_var
+
 class ConvertSetItemPass:
     """Parfor subpass to convert setitem on Arrays
     """
@@ -1713,11 +1716,10 @@ class ConvertSetItemPass:
             new_body = []
             equiv_set = pass_states.array_analysis.get_equiv_set(label)
             for instr in block.body:
-                if isinstance(instr, ir.StaticSetItem) or isinstance(instr, ir.SetItem):
+                if isinstance(instr, (ir.StaticSetItem, ir.SetItem)):
                     loc = instr.loc
                     target = instr.target
-                    index = (instr.index if isinstance(instr, ir.SetItem)
-                             else instr.index_var)
+                    index = get_index_var(instr)
                     value = instr.value
                     target_typ = pass_states.typemap[target.name]
                     index_typ = pass_states.typemap[index.name]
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by sklam — CHANGES_REQUESTED** [review]
I've suggested additional test to improve coverage.

File: numba/tests/test_parfors.py [hunk 1]
```diff
@@ -2800,6 +2800,22 @@ def test_impl():
...
+            for i in range(1):
+                states[i]['var'] = 1
+
+        def comparer(a, b):
+            assert(a[0]['var'] == b[0]['var'])
+
+        self.prange_tester(test_impl,
+                           np.empty(shape=1, dtype=state_dtype),
+                           check_arg_equality=[comparer])
+
```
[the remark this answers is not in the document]
DrTodd13: Done.

stuartarchibald:
Open/Close was to try and re-start the RTD docs build.

numba/tests/test_parfors.py
```diff
@@ -2800,6 +2800,41 @@ def test_impl():
...
+        def test_impl(states):
+            for i in range(1):
+                states[i]['var'] = 1
+
+        def comparer(a, b):
+            assert(a[0]['var'] == b[0]['var'])
+
+        self.prange_tester(test_impl,
+                           np.empty(shape=1, dtype=state_dtype),
```
stuartarchibald, line 2816:
```suggestion
                           np.zeros(shape=1, dtype=state_dtype),
```
just in case `empty` contains a `1`!

numba/parfors/parfor.py
```diff
@@ -1689,6 +1689,9 @@ def _type_getitem(self, args):
         return self.pass_states.typingctx.resolve_function_type(fnty, tuple(args), {})
 
 
+def get_index_var(x):
+    return x.index if isinstance(x, ir.SetItem) else x.index_var
+
```
stuartarchibald, line 1694:
```suggestion


```

stuartarchibald (commented):
Thanks for the patch, couple of minor things else looks good.

stuartarchibald (approved):
Thanks for the patch and fixes. @sklam's unit test is added so I think everything is resolved.

sklam (approved):

PR #6802, opened 2021-03-06
Add lineinfo flag to PTX and SASS compilation

maxpkatz:
A line mapping between CUDA assembly code (PTX, SASS) and the original Python code is useful for many developer tools. The -lineinfo/-generate-line-info flag in CUDA generates this line mapping as annotations to the assembly (without changing optimization level). So it provides some amount of lightweight debugging information.

This change adds options at the CUDA compiler level for the relevant flags to the PTX generation (and JIT compilation to SASS) to generate this line mapping. It sets up a future change to enable the line mapping directly from the @cuda.jit decorator.

A test is added to the PTX generation which verifies the line mapping is present in the PTX after compilation from LLVM to PTX. No test of the new functionality is added to the PTX->SASS JIT compilation because there isn't already a simple test of this code path to base that on and adding one would be a fair amount of effort. However, we will be able to indirectly test the new functionality later when the cuda.jit decorator is added. I have verified locally with a prototype of the cuda.jit decorator change that the generated SASS looks good.


maxpkatz:
Note: it is already possible today to get this information with the `debug=True` decorator to @cuda.jit. The reason why a separate lineinfo option is useful is that lineinfo mode is intended to only generate the line mapping without any other debugging information generated (and the consequent loss of optimization that might result in).

gmarkall:
I've not reviewed the code yet, but have experimented with this branch. Some observations:

* It only seems to work with CUDA 11.2,  which is based on NVVM70. I suspect debug info generation for NVVM34 has various issues (there are some in the issue tracker) and I think it might have bitrotted a bit. For NVVM70 the same `DIBuilder` used by the CPU target can be used, which has probably had much more practical use and testing than the `NvvmDIBuilder` (which is only used for NVVM34). I don't think this is a problem - I think we should only support lineinfo for NVVM70-based toolkits, instead of trying to go back and fix the `NvvmDIBuilder`.
* SASS seems to correlate back to SASS as expected (see attached image).
* The `lineinfo` kwarg to the `@cuda.jit` decorator also seems to be working - I used it in my experiments, because I didn't read the PR description properly before I started testing :-)

image

gmarkall (commented):
Many thanks for the PR - this is something I've wanted to get sorted for a long time, so I'm really happy to see this! In general the PR is looking good. Some comments:

- Tests are passing locally for me with the simulator and hardware target.
- As this only works with 11.2 onwards due to the broken NVVM 3.4 debug info generation in Numba, I think it would be good to document somewhere that this is only supported for 11.2 onwards (I don't think fixing the old debuginfo generation is a worthwhile endeavor).
- As mentioned in the previous comment, the `lineinfo` kwarg to the jit decorator is working, because it propagates any unknown kwargs. Adding information about lineinfo to the `cuda.jit` docstring and a test of lineinfo for the jit decorator would be sufficient to consider lineinfo for the jit decorator "complete" for this PR. Do you want to add it to this PR?
- After #6769 goes in, this PR will need some modification because `CachedCUFunction` will be removed, and the `CUDACodeLibrary` class will need the equivalent modifications. I'm happy to provide a patch with this fixup after #6769 is merged, unless you'd prefer to make the change?

numba/cuda/compiler.py
```diff
@@ -104,7 +104,10 @@ def compile_cuda(pyfunc, return_type, args, debug=False, inline=False,
     flags.set('no_compile')
     flags.set('no_cpython_wrapper')
     flags.set('no_cfunc_wrapper')
-    if debug:
+    if debug or lineinfo:
+        # Note both debug and lineinfo turn on debug information in the
+        # compiled code, but we keep them separate arguments in case we
+        # later want to overload some other behavior on the debug flag.
```
gmarkall, line 110: In particular, `-opt=3` is not supported with `-g`, so it's a good thing to keep them separate: https://docs.nvidia.com/cuda/libnvvm-api/group__compilation.html#group__compilation_1g76ac1e23f5d0e2240e78be0e63450346

maxpkatz: I've added a comment to this effect.

numba/cuda/compiler.py
```diff
@@ -181,14 +192,14 @@ def compile_ptx(pyfunc, args, debug=False, device=False, fastmath=False,
...
-                                   fastmath=False, opt=True):
+def compile_ptx_for_current_device(pyfunc, args, debug=False, lineinfo=False,
+                                   device=False, fastmath=False, opt=True):
     """Compile a Python function to PTX for a given set of argument types for
     the current device's compute capabilility. This calls :func:`compile_ptx`
     with an appropriate ``cc`` value for the current device."""
     cc = get_current_device().compute_capability
-    return compile_ptx(pyfunc, args, debug=-debug, device=device,
-                       fastmath=fastmath, cc=cc, opt=True)
+    return compile_ptx(pyfunc, args, debug=-debug, lineinfo=lineinfo,
```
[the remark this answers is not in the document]
maxpkatz: Agreed, fixed.

gmarkall (changes requested):
Many thanks for the updates! Some notes:

* The tests pass locally for me with CUDA 11.3
* Nsight compute reports the line info in the source view as shown in my earlier comment https://github.com/numba/numba/pull/6802#issuecomment-797675343

Based on previous discussion, I think this needs the docstring for the `@cuda.jit` decorator updating, and tests of the lineinfo generation with the `@jit` decorator. I'll have another look at this tomorrow with fresh eyes too (in case there are other things I missed this evening)

gmarkall:
@stuartarchibald I moved this back to 0.54 RC as I'm hoping we can get it over the line this week now it's updated.

maxpkatz:
@gmarkall I've updated the docstring for @cuda.jit and added a test which verifies the presence of the linemapping for this decorator.

maxpkatz:
Also worth noting: your previous comment indicated that I would need to make my changes from CachedCUFunction in CUDACodeLibrary but that appears not to have been necessary, because it is already implicitly passed through as part of the `nvvm_options`.

numba/cuda/decorators.py
```diff
@@ -55,6 +55,11 @@ def jit(func_or_sig=None, device=False, inline=False, link=[], debug=None,
                 enabled. When ``True``, ``-opt=3`` is passed to NVVM. When
                 ``False``, ``-opt=0`` is passed to NVVM. Defaults to ``True``.
     :type opt: bool
+
+    In addition to the above parameters the keyword ``lineinfo=True`` may
```
gmarkall, line 59: Could you follow the format of the other documented kwargs for this (i.e. `:param lineinfo:` and `:type lineinfo:`) please? I think this will make it clearer in the documentation, and this format is not checked against the list of kwargs in the function signature

gmarkall:
> Also worth noting: your previous comment indicated that I would need to make my changes from CachedCUFunction in CUDACodeLibrary but that appears not to have been necessary, because it is already implicitly passed through as part of the `nvvm_options`.

I think you're right here - I hadn't realised it would already work for the reason you outline above.

gmarkall:
(A couple of comments added, but I've flipped the status to "Waiting on reviewer") as i should still look at this more thoroughly.

numba/cuda/decorators.py
maxpkatz: Certainly can. I wasn't sure if I should since it's not explicitly in the interface.

gmarkall (approved):
I've had another look over this and this is looking good to me - I've built the documentation to check its formatting, and tested locally with hardware.

@stuartarchibald / @esc could this have a buildfarm run please?

numba/cuda/compiler.py
```diff
@@ -134,8 +134,8 @@ def define_cuda_lowering_pipeline(self, state):
 
 
 @global_compiler_lock
-def compile_cuda(pyfunc, return_type, args, debug=False, inline=False,
-                 fastmath=False, nvvm_options=None):
+def compile_cuda(pyfunc, return_type, args, debug=False, lineinfo=False,
+                 inline=False, fastmath=False, nvvm_options=None):
```
stuartarchibald, line 138: This is adjusting the position of keyword arguments in a function that doesn't prescribe keyword-only argument behaviour. As discussed OOB with @gmarkall, this does present a risk for those calling the function signatures changed in this PR with positional arguments, however this risk appears minimal due to the functions changed herein being a somewhat non-public/less used API. Deprecation notices will be added to the Numba 0.54 docs to state that the functions will adopt keyword only behaviour in Numba 0.55.

numba/cuda/cudadrv/nvvm.py
```diff
@@ -264,6 +264,10 @@ def compile(self, **options):
             if options.pop('debug'):
                 opts.append('-g')
 
+        if 'lineinfo' in options:
+            if options.pop('lineinfo'):
+                opts.append('-generate-line-info')
```
stuartarchibald, line 269:
```suggestion
        if options.pop('lineinfo', False):
            opts.append('-generate-line-info')
```

stuartarchibald (approved):
Thanks for the patch @maxpkatz.

Thanks for reviewing @gmarkall.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_69`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_69`.

Passed.

Issue #6040, opened 2020-07-28
CUDA: Add runtime version to `numba -s`


sklam:
## Feature request

With https://github.com/numba/numba/pull/5819 highlighting that Runtime API version is needed in additional to the Driver API version, i think we should add the Runtime version in `numba -s` along with https://github.com/numba/numba/blob/b3c5bcbaca955c544d06778ec778f8c62f89b60c/numba/misc/numba_sysinfo.py#L346-L368

PhanatosZou:
@gmarkall @sklam Hi, I would like to work on this feature, but I'm new here and would like to request some help in guiding to start working on this feature. Thanks in advance!

stuartarchibald:
@PhanatosZou thanks for wanting to work on this, Numba's contributing guide is here https://numba.readthedocs.io/en/stable/developer/contributing.html and the file to be edited is that noted in the above.

PhanatosZou:
@stuartarchibald Thank you so much for the information. I'll check it out.

rawwar:
Since this is open, I want to take it up. What should be the output format? Cuda runtime is returned as `1000 major + 10 minor`. So, should the output be `11.2` for `11020` for Cuda runtime 11.2?

gmarkall:
@rawwar It can - thanks for the reminder! :-)

PR #6804, opened 2021-03-07
added runtime version to `numba -s`

rawwar:
Imported `runtime` object from numba.cuda.cudadrv.runtime and called  `cudaRuntimeGetVersion` on the `runtime` object passing a c_int variable by reference to get the cuda runtime information.

### DIFF
--- a/numba/misc/numba_sysinfo.py
+++ b/numba/misc/numba_sysinfo.py
@@ -597,7 +597,7 @@ def display_sysinfo(info=None, sep_pos=45):
         ("__CUDA Information__",),
         ("CUDA Device Initialized", info.get(_cu_dev_init, '?')),
         ("CUDA Driver Version", info.get(_cu_drv_ver, '?')),
-        ("CUDA Runtime Version",info.get(_cu_rt_ver,'?')),
+        ("CUDA Runtime Version",info.get(_cu_rt_ver, '?')),
         ("CUDA Detect Output:",),
         (info.get(_cu_detect_out, "None"),),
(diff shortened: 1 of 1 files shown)


numba/misc/numba_sysinfo.py
```diff
@@ -591,6 +597,7 @@ class DisplaySeqMaps(tuple):
         ("__CUDA Information__",),
         ("CUDA Device Initialized", info.get(_cu_dev_init, '?')),
         ("CUDA Driver Version", info.get(_cu_drv_ver, '?')),
+        ("CUDA Runtime Version",info.get(_cu_rt_ver,'?')),
```
gmarkall, line 600:
```suggestion
        ("CUDA Runtime Version",info.get(_cu_rt_ver, '?')),
```

gmarkall (changes requested):
This looks good to me, with one minor syntactic edit marked on the diff.

Per the question on #6040:

> What should be the output format?

The existing output format of `1000 * major + 10 * minor` is good - this matches the format of the driver version presently given by the sysinfo tool.

Example output on my machine is:

```
__CUDA Information__
CUDA Device Initialized                       : True
CUDA Driver Version                           : 11020
CUDA Runtime Version                          : 11000
```

gmarkall (approved):
Many thanks for the update - looks good to me!

gmarkall:
(Waiting on review in case anyone else wants to check / approve first)

stuartarchibald (approved):
Thanks for the patch. Tested this locally and it works as expected.

PR #3468, opened 2018-11-02
Add support for `np.clip` and `ndarray.clip`.

arvoelke:
As title.

Fixes #3465

Issue #3489, opened 2018-11-13
kwarg support in extending API `@overload_method` broken


stuartarchibald:
## Reporting a bug

- [X] I am using the latest released version of Numba (most recent is visible in
 the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [X] I have included below a minimal working reproducer (if you are unsure how
 to write one see http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports).

This:
```python
import numpy as np
from numba import njit, types
from numba.extending import overload_method

@overload_method(types.Array, 'foo')
def fooimpl(arr, a_kwarg=10):
    def impl(arr, a_kwarg=10):
        return a_kwarg
    return impl

@njit
def bar(A):
    print(A.foo())

Z = np.arange(5)

bar(Z)
```
fails to work. The error that appears is a `TypeError`, with traceback as follows:
```
Traceback (most recent call last):
  File "<path>/numba/numba/errors.py", line 601, in new_error_context
    yield
  File "<path>/numba/numba/lowering.py", line 257, in lower_block
    self.lower_inst(inst)
  File "<path>/numba/numba/lowering.py", line 306, in lower_inst
    val = self.lower_assign(ty, inst)
  File "<path>/numba/numba/lowering.py", line 452, in lower_assign
    return self.lower_expr(ty, value)
  File "<path>/numba/numba/lowering.py", line 875, in lower_expr
    res = self.lower_call(resty, expr)
  File "<path>/numba/numba/lowering.py", line 841, in lower_call
    res = impl(self.builder, argvals, self.loc)
  File "<path>/numba/numba/targets/base.py", line 1116, in __call__
    return self._imp(self._context, builder, self._sig, args)
  File "<path>/numba/numba/typing/templates.py", line 532, in method_impl
    return call(builder, args)
  File "<path>/numba/numba/targets/base.py", line 1116, in __call__
    return self._imp(self._context, builder, self._sig, args)
  File "<path>/numba/numba/targets/imputils.py", line 192, in imp
    builder, func, fndesc.restype, fndesc.argtypes, args)
  File "<path>/numba/numba/targets/callconv.py", line 479, in call_function
    args = list(arginfo.as_arguments(builder, args))
  File "<path>/numba/numba/datamodel/packer.py", line 94, in as_arguments
    % (self._nargs, len(values)))
TypeError: invalid number of args: expected 2, got 1

During handling of the above exception, another exception occurred:

[...]

PR #6808, opened 2021-03-08
#3468 continued: Add support for `np.clip`

gmarkall:
This PR contains the code from #3468, rebased on master and current issues resolved. After rebasing, the following two issues were present:

1. NumPy behaviour for clip where min > max changed, making the check `np.testing.assert_equal(pyfunc(a, 8, 5), cfunc(a, 8, 5))` fail. The fix to this is to implement `np_clip_impl` similarly to `_NPY_CLIP`, which is defined as `_NPY_<type>_MIN(_NPY_<type>_MAX((x), (min)), (max))`.
2. The `lower_clip_result` function in the test should call the jitted function, not the Python one.

cc @arvoelke

### DIFF
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -1800,40 +1800,26 @@ def np_clip(a, a_min, a_max, out=None):
     a_min_is_none = a_min is None or isinstance(a_min, types.NoneType)
     a_max_is_none = a_max is None or isinstance(a_max, types.NoneType)
 
-    if a_min_is_none and a_max_is_none:
-        def np_clip_impl(a, a_min, a_max, out=None):
-            # implementation for no bounds
+    def np_clip_impl(a, a_min, a_max, out=None):
+        if a_min_is_none and a_max_is_none:
             raise ValueError("array_clip: must set either max or min")
 
-    elif a_min_is_none:
-        def np_clip_impl(a, a_min, a_max, out=None):
-            # implementation for upper-bound only
-            ret = _alloc_out(a, out)
-            for index, val in np.ndenumerate(a):
+        ret = _alloc_out(a, out)
+
+        for index, val in np.ndenumerate(a):
+            if a_min_is_none:
                 if val > a_max:
                     ret[index] = a_max
                 else:
                     ret[index] = val
-            return ret
-
-    elif a_max_is_none:
-        def np_clip_impl(a, a_min, a_max, out=None):
-            # implementation for lower-bound only
-            ret = _alloc_out(a, out)
-            for index, val in np.ndenumerate(a):
+            elif a_max_is_none:
                 if val < a_min:
                     ret[index] = a_min
                 else:
                     ret[index] = val
-            return ret
-
-    else:
-        def np_clip_impl(a, a_min, a_max, out=None):
-            # implementation when both bounds defined
-            ret = _alloc_out(a, out)
-            for index, val in np.ndenumerate(a):
+            else:
                 ret[index] = min(max(val, a_min), a_max)
-            return ret
+        return ret
 
     return np_clip_impl
 
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -1776,6 +1776,19 @@ def array_flatten(context, builder, sig, args):
 
 @overload(np.clip)
 def np_clip(a, a_min, a_max, out=None):
+    if not type_can_asarray(a):
+        raise errors.TypingError('The argument "a" must be array-like')
+
+    if not isinstance(a_min, (types.NoneType, types.Number)):
+        raise errors.TypingError('The argument "a_min" must be a number')
+
+    if not isinstance(a_max, (types.NoneType, types.Number)):
+        raise errors.TypingError('The argument "a_max" must be a number')
+
+    if not isinstance(out, (types.NoneType, types.Array)):
+        msg = 'The argument "out" must be an array if it is provided'
+        raise errors.TypingError(msg)
+
     # TODO: support scalar a (issue #3469)
     a_min_is_none = a_min is None or isinstance(a_min, types.NoneType)
     a_max_is_none = a_max is None or isinstance(a_max, types.NoneType)
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -1774,6 +1774,82 @@ def array_flatten(context, builder, sig, args):
     return res
 
 
+def _alloc_out(a, X):
+    # dummy function that takes the arg we need to resolved (X)
+    # and the information needed to do so (a)
+    pass
+
+
+@overload(_alloc_out)
+def _alloc_out_impl(a, X):
+    # overload the dummy function to return empty_like(a) if X is None else X
+    # with type-specific implementations, in order to fully determine type of X
+    # https://github.com/numba/numba/pull/3468#issuecomment-437974147
+    if X is None or isinstance(X, types.NoneType):
+        def impl(a, X):
+            return np.empty_like(a)
+    else:
+        def impl(a, X):
+            return X
+    return impl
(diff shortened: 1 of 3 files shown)


gmarkall:
@stuartarchibald As we discussed out-of-band, I've simplified the implementation of `np.clip` to use a single function for all cases of the presence of `a_min` and `a_max`. Assuming tests are green, this should be ready for review now.

numba/np/arrayobj.py
```diff
@@ -1774,6 +1774,63 @@ def imp(ary):
...
+        def impl(a, X):
+            return np.empty_like(a)
+    else:
+        def impl(a, X):
+            return X
+    return impl
+
+
+@overload(np.clip)
+def np_clip(a, a_min, a_max, out=None):
```
stuartarchibald, line 1798: The types of the input arguments need validation and then tests adding to check.

numba/np/arrayobj.py
```diff
@@ -1774,6 +1774,63 @@ def imp(ary):
...
+    # TODO: support scalar a (issue #3469)
+    a_min_is_none = a_min is None or isinstance(a_min, types.NoneType)
+    a_max_is_none = a_max is None or isinstance(a_max, types.NoneType)
+
+    def np_clip_impl(a, a_min, a_max, out=None):
+        if a_min_is_none and a_max_is_none:
+            raise ValueError("array_clip: must set either max or min")
+
+        ret = _alloc_out(a, out)
```
stuartarchibald, line 1807: Can probably do:
```python
ret = np.empty_like(a) if out is None else a
```
and the compiler will sort out which branch is valid. This will remove the need for the function/overload `_alloc_out`.

numba/tests/test_array_methods.py
```diff
@@ -1329,6 +1335,45 @@ def test_imag(self):
...
+    def test_clip(self):
+        # TODO: scalars are not tested (issue #3469)
+        for a in (np.linspace(-10, 10, 101),
+                  np.linspace(-10, 10, 40).reshape(5, 2, 4)):
+            for pyfunc in [np_clip, array_clip]:
+                cfunc = jit(nopython=True)(pyfunc)
+
+                with self.assertRaises(ValueError):
+                    pyfunc(a, None, None)
+                with self.assertRaises(ValueError):
```
stuartarchibald, line 1347: Assert the correct error string?

numba/tests/test_array_methods.py
```diff
@@ -1329,6 +1335,45 @@ def test_imag(self):
...
+                cout = np.empty_like(a)
+                np.testing.assert_equal(pyfunc(a, -5, 5, pyout),
+                                        cfunc(a, -5, 5, cout))
+                np.testing.assert_equal(pyout, cout)
+
+                # This test exhibits undefined behaviour since min > max, but
+                # we'll make sure it's consistent with NumPy, which computes
+                # min(max(val, a_min), a_max) since 1.17. Prior to that min and
+                # max were the other way round, so we're not consistent with
+                # older NumPy versions.
```
stuartarchibald, line 1365: Should probably note this in the docs?

numba/tests/test_array_methods.py
```diff
@@ -1329,6 +1335,45 @@ def test_imag(self):
...
+                if numpy_version > (1, 16):
+                    np.testing.assert_equal(pyfunc(a, 8, 5), cfunc(a, 8, 5))
+
+                # verifies that type-inference is working on the return value
+                # this used to trigger issue #3489
+                def lower_clip_result(a):
+                    return np.expm1(cfunc(a, -5, 5))
+                np.testing.assert_almost_equal(
+                    lower_clip_result(a),
+                    jit(nopython=True)(lower_clip_result)(a))
```
stuartarchibald, line 1375: I'm not sure what this is about in relation to #3489, any chance some more explanation could be offered please?

numba/tests/test_array_methods.py
```diff
@@ -232,6 +232,12 @@ def array_real(a):
 def array_imag(a):
     return np.imag(a)
 
+def np_clip(a, a_min, a_max, out=None):
+    return np.clip(a, a_min, a_max, out)
+
+def array_clip(a, a_min=None, a_max=None, out=None):
+    return a.clip(a_min, a_max, out)
```
stuartarchibald, line 239: It'd be good to also include variants where the kwargs are explicitly wired, e.g.
```python
def np_clip_kwargs(a, a_min, a_max, out):
    return np.clip(a, a_min, a_max, out=out)
```
and the same for `array_clip` so as to ensure the kwargs are behaving as expected (along with test modifications to check).

stuartarchibald (commented):
Thanks for the original patch @arvoelke and thanks for carrying this forward @gmarkall. There's just a few small things to look at and once resolved this should be good to merge. Thanks again!

numba/np/arrayobj.py
gmarkall: This seems to work (with `else out` instead of `else a`, which I presume was your intention).

numba/np/arrayobj.py
gmarkall: Added.

numba/tests/test_array_methods.py
gmarkall: Or maybe we shouldn't test this at all? I copied this test from the original PR (then had to modify it to match the current NumPy behaviour) - In this scenario, the user is doing something daft by trying to clip with a minimum greater than the maximum. I think perhaps it would be better to remove this check and in general not attempt to match NumPy when the user does strange things.

Are you happy if I just remove this check instead?

numba/tests/test_array_methods.py
gmarkall: From https://github.com/numba/numba/pull/3468#issuecomment-439137556, it looks like this is testing the case where the returned value is a kwarg, which #3489 reports as not working.

numba/tests/test_array_methods.py
gmarkall: Added.

numba/tests/test_array_methods.py
gmarkall: I've added an assertion for the correct error string. I removed the check of the raise for the pyfunc, as it wasn't really testing anything to do with Numba.

gmarkall:
@stuartarchibald Thanks for the review - all the comments should be addressed now, except for a question about the test where `min > max` - what do you think about removing that (odd) check?

numba/tests/test_array_methods.py
stuartarchibald: As Numba 0.54 is moving to NumPy 1.17 as a minimum this probably becomes moot? I agree however that removing this test on the basis it's checking something strange is probably fine.

numba/tests/test_array_methods.py
stuartarchibald: Thanks.

numba/tests/test_array_methods.py
stuartarchibald: thanks

stuartarchibald:
> @stuartarchibald Thanks for the review - all the comments should be addressed now, except for a question about the test where `min > max` - what do you think about removing that (odd) check?

Thanks for the fixes. I've replied inline above.

numba/tests/test_array_methods.py
gmarkall: I think it does. I've removed the test.

gmarkall:
@stuartarchibald I think the only remaining comment is now addressed.

numba/tests/test_array_methods.py
stuartarchibald: Great, thanks.

stuartarchibald (approved):
Thanks for the patch and fixes @arvoelke @gmarkall .

PR #6810, opened 2021-03-09
Fix tiny formatting error in ROC kernel docs

felixdivo:
Removes a tiny space that caused a the bullet point in the list list to be printed in bold instead. Noticed it [here](https://numba.readthedocs.io/en/stable/roc/kernels.html#choosing-the-workgroup-size).

Screenshot from 2021-03-09 19-51-17


stuartarchibald (approved):
Thanks for the patch. https://numba--6810.org.readthedocs.build/en/6810/roc/kernels.html# confirms fix.

@felixdivo Congratulations on your first contribution to Numba!

PR #6811, opened 2021-03-09
CUDA: Remove test of runtime being a supported version

gmarkall:
This test fails if one uses a CUDA runtime version that is not recognized as being supported by Numba. The problem with this is that a user running the tests against a newer version of CUDA (supposing a version 11.3, or 12.x version existed) would report this test as failing, even though there might not be anything wrong with using the current version of Numba with the future CUDA toolkit version. (Running the version of Numba with a newer toolkit might not be officially supported, but that's no reason to prevent the test suite from passing 100% with it if it otherwise would have done)

This PR removes the offending test.

### DIFF
--- a/numba/cuda/tests/cudadrv/test_runtime.py
+++ b/numba/cuda/tests/cudadrv/test_runtime.py
@@ -27,9 +27,6 @@ else:
 
 
 class TestRuntime(unittest.TestCase):
-    def test_get_version(self):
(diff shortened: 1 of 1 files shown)


stuartarchibald (approved):
Thanks for the patch. Removing this seems reasonable due to the behaviours noted. The rest of the tests in the class also cover a number of similar but more controlled scenarios and so I think we can still have confidence in problems being picked up.

PR #6816, opened 2021-03-11
Update changelog for 0.53.0 final

sklam:
- update release date

PR #6817, opened 2021-03-11
Update changelog for 0.53.0 final release date

sklam:
Same as https://github.com/numba/numba/pull/6816 but for main


stuartarchibald (approved):
Thanks for the patch!

PR #6822, opened 2021-03-14
Add support for np.rot90

braniii:
This PR is adds `np.rot90`, which is currently unimplemented according to #4074
The only problem is, that in numba `np.transpose` the axes argument needs to be a tuple. I am not able to create the tuple dynamically

### DIFF
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -1562,6 +1562,45 @@ def array_T(context, builder, typ, value):
     return impl_ret_borrowed(context, builder, typ, res)
 
 
+@overload(np.rot90)
+def numpy_rot90(arr, k=1):
+    # axes = (0, 1) using other axes needs axis support in flip
+
+    if not isinstance(k, types.Integer):
+        raise errors.TypingError('The second argument "k" must be an integer')
+    if not type_can_asarray(arr):
+        raise errors.TypingError('The first argument "arr" must be array-like')
+
+    def impl(arr, k=1):
+        arr = np.asarray(arr)
+
+        if arr.ndim < 2:
+            raise ValueError('Input must be >= 1-d.')
+
+        k = k % 4
+        if k == 0:
+            return arr[:]
+        if k == 2:
+            return np.flipud(np.fliplr(arr))
+
+        def axes_list_gen(i):
+            if i == 0:
+                return 1
+            if i == 1:
+                return 0
+            return i
+
+        axes_list = (1, 0)
+        for i in range(2, arr.ndim):
+            axes_list = tuple_setitem(axes_list, i, i)
+
+        if k == 1:
+            return np.transpose(np.fliplr(arr), axes_list)
+        # k == 3
+        return np.fliplr(np.transpose(arr, axes_list))
+    return impl
+
+
 def _attempt_nocopy_reshape(context, builder, aryty, ary,
                             newnd, newshape, newstrides):
     """
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -19,6 +19,7 @@ from numba.core import types, utils, typing, errors, cgutils, extending
 from numba.np.numpy_support import (as_dtype, carray, farray, is_contiguous,
                                     is_fortran)
 from numba.np.numpy_support import type_can_asarray, is_nonelike
+from numba.np.unsafe.ndarray import to_fixed_tuple
 from numba.core.imputils import (lower_builtin, lower_getattr,
                                  lower_getattr_generic,
                                  lower_setattr_generic,
@@ -1583,16 +1584,9 @@ def numpy_rot90(arr, k=1):
         if k == 2:
             return np.flipud(np.fliplr(arr))
 
-        def axes_list_gen(i):
-            if i == 0:
-                return 1
-            if i == 1:
-                return 0
-            return i
-
-        axes_list = (1, 0)
-        for i in range(2, arr.ndim):
-            axes_list = tuple_setitem(axes_list, i, i)
+        axes_list = np.arange(arr.ndim)
+        axes_list[:2] = [1, 0]
+        axes_list = to_fixed_tuple(axes_list, arr.ndim)
 
         if k == 1:
             return np.transpose(np.fliplr(arr), axes_list)
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -19,7 +19,6 @@ from numba.core import types, utils, typing, errors, cgutils, extending
 from numba.np.numpy_support import (as_dtype, carray, farray, is_contiguous,
                                     is_fortran)
 from numba.np.numpy_support import type_can_asarray, is_nonelike
-from numba.np.unsafe.ndarray import to_fixed_tuple
 from numba.core.imputils import (lower_builtin, lower_getattr,
                                  lower_getattr_generic,
                                  lower_setattr_generic,
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -1565,12 +1565,13 @@ def array_T(context, builder, typ, value):
 
 @overload(np.rot90)
 def numpy_rot90(arr, k=1):
-    # axes = (0, 1) using other axes needs axis support in flip
-
+    # supporting axes argument it needs to be included in np.flip
     if not isinstance(k, types.Integer):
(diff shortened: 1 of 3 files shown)


esc:
@braniii thank you for submitting this. I have marked it as in-progress for now, until the transpose issues have been resolved.

stuartarchibald:
@braniii this should sort out the fixed tuple issue:
```patch
index 9e81080..ea9b613 100644
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -1569,8 +1569,10 @@ def numpy_rot90(arr, k=1):
 
     if not isinstance(k, types.Integer):
         raise errors.TypingError('The second argument "k" must be an integer')
-    if not type_can_asarray(arr):
-        raise errors.TypingError('The first argument "arr" must be array-like')
+        raise errors.TypingError('The first argument "arr" must be an array')
+
+    const_arr_ndim = arr.ndim
 
     def impl(arr, k=1):
         arr = np.asarray(arr)
@@ -1586,7 +1588,7 @@ def numpy_rot90(arr, k=1):
 
         axes_list = np.arange(arr.ndim)
         axes_list[:2] = [1, 0]
-        axes_list = to_fixed_tuple(axes_list, arr.ndim)
+        axes_list = to_fixed_tuple(axes_list, const_arr_ndim)
 
         if k == 1:
             return np.transpose(np.fliplr(arr), axes_list)
diff --git a/numba/tests/test_np_functions.py b/numba/tests/test_np_functions.py
index 76c7188..3f1e3e5 100644
--- a/numba/tests/test_np_functions.py
+++ b/numba/tests/test_np_functions.py
@@ -2328,7 +2328,6 @@ class TestNPFunctions(MemoryLeakMixin, TestCase):
         def a_variations():
             yield np.arange(10).reshape(5, 2)
             yield np.arange(20).reshape(5, 2, 2)
-            yield ([1, 2], [3, 4],)
 
         for a in a_variations():
             for k in range(-3, 13):
@@ -2339,7 +2338,7 @@ class TestNPFunctions(MemoryLeakMixin, TestCase):
         with self.assertRaises(TypingError) as raises:
             cfunc("abc")
 
-        self.assertIn('The first argument "arr" must be array-like',
+        self.assertIn('The first argument "arr" must be an array',
                       str(raises.exception))
 
         with self.assertRaises(TypingError) as raises:
```

braniii:
@stuartarchibald thank you for your help. Using your changes the test passes (at least locally). I am looking forward to a review :)

stuartarchibald:
> @stuartarchibald thank you for your help. Using your changes the test passes (at least locally). I am looking forward to a review :)

No problem, glad it got this working for you.

stuartarchibald (commented):
Thanks for working on implementing this feature. I've given the patch and initial review and it's a good start. There's a few Numba nuances that I've highlighted and have demoed one way to fix them in this patch https://github.com/stuartarchibald/numba/commit/85ab2e272fa9ecde0a97b5aeef3df3ac04d3986a, most of this revolves around knowing when Numba "sees" a type and what can be inferred at compile time vs. runtime. Thanks again!

numba/tests/test_np_functions.py
```diff
@@ -2317,6 +2321,45 @@ def a_variations():
...
+    def test_rot90_basic(self):
+        pyfunc = rot90
+        cfunc = jit(nopython=True)(pyfunc)
+
+        def a_variations():
+            yield np.arange(10).reshape(5, 2)
+            yield np.arange(20).reshape(5, 2, 2)
+
+        for a in a_variations():
+            for k in range(-3, 13):
```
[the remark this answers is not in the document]
braniii: I've tried to stay close to numpy's test. But you are right, it is not of need.

numba/np/arrayobj.py
```diff
@@ -1562,6 +1563,39 @@ def array_T(context, builder, typ, value):
     return impl_ret_borrowed(context, builder, typ, res)
 
 
+@overload(np.rot90)
+def numpy_rot90(arr, k=1):
+    # supporting axes argument it needs to be included in np.flip
+    if not isinstance(k, types.Integer):
```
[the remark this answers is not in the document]
braniii: This is a valid point, I took your tests.

numba/np/arrayobj.py
```diff
@@ -1562,6 +1563,36 @@ def array_T(context, builder, typ, value):
...
+    if not isinstance(arr, types.Array):
+        raise errors.TypingError('The first argument "arr" must be an array')
+
+    if arr.ndim < 2:
+        raise ValueError('Input must be >= 1-d.')
+
+    axes_list = (1, 0, *range(2, arr.ndim))
+
+    def impl(arr, k=1):
+        arr = np.asarray(arr)
```
braniii, line 1580: Is this needed if I check for `types.Array` in line 1571? Tests passes without this line.

braniii:
> Thanks for working on implementing this feature. I've given the patch and initial review and it's a good start. There's a few Numba nuances that I've highlighted and have demoed one way to fix them in this patch [stuartarchibald@85ab2e2](https://github.com/stuartarchibald/numba/commit/85ab2e272fa9ecde0a97b5aeef3df3ac04d3986a), most of this revolves around knowing when Numba "sees" a type and what can be inferred at compile time vs. runtime. Thanks again!

Thank you for the detailed review and patch, it was really helpful. I changed some flake8 white space issues and simplified the tuple creation. If you prefer the explicit call of `tuple` I will revert the change.

numba/np/arrayobj.py
stuartarchibald: I think that can indeed be removed.

stuartarchibald:
> > Thanks for working on implementing this feature. I've given the patch and initial review and it's a good start. There's a few Numba nuances that I've highlighted and have demoed one way to fix them in this patch [stuartarchibald@85ab2e2](https://github.com/stuartarchibald/numba/commit/85ab2e272fa9ecde0a97b5aeef3df3ac04d3986a), most of this revolves around knowing when Numba "sees" a type and what can be inferred at compile time vs. runtime. Thanks again!
> 
> Thank you for the detailed review and patch, it was really helpful. I changed some flake8 white space issues and simplified the tuple creation. If you prefer the explicit call of `tuple` I will revert the change.

No problem, and thanks for fixing up :)

stuartarchibald:
@esc @sklam this PR contains some bits of code I've authored, please could one of you take a look too, patch is pretty short. Thanks.

numba/np/arrayobj.py
braniii: Ok, I removed it.

numba/np/arrayobj.py
```diff
@@ -1562,6 +1562,35 @@ def array_T(context, builder, typ, value):
...
+        if k == 0:
+            return arr[:]
+        elif k == 1:
+            return np.transpose(np.fliplr(arr), axes_list)
+        elif k == 2:
+            return np.flipud(np.fliplr(arr))
+        elif k == 3:
+            return np.fliplr(np.transpose(arr, axes_list))
+        else:
+            assert 0  # unreachable
```
sklam, line 1589: i thk this should explicitly raise to prevent `python -O` causing the `assert` to be removed and falling through as a return None, which will alter the return type of the function.

braniii: The `else` case is entered only when `k = k %  4` is not in `[0, 1, 2, 3]`, so when modulo operation fails. So an alternative is to change the last `elif k==3` to an `else` and  removing lines 1588-1589. In case you prefer to keep the unreachable else case, could you please help me which kind of error I should use?

sklam: Putting a `raise AssertionError` is suffice. 

Details:

Since this code is compiled by numba, static analysis does not know the range of possible `k`. Type inference (a static analysis) will consider all branches in determining the return type. With `python -O`, the `assert` statement will be omitted; thus, the `else` branch becomes equivalent to a `return None`. The type-inference will then think the output type can be `None` and using an `Optional[Array]` return type unnecessarily.

braniii: I've just fixed it and thx for the helpful explanation.

stuartarchibald: Thanks @sklam I need to remember to stop doing this C-like idiom :)

sklam (approved):
Thanks for your contribution!

stuartarchibald:
@braniii Congratulations on your first contribution to Numba!

PR #6826, opened 2021-03-15
Fix regression on gufunc serialization

### LINKED ISSUE #6821: Regression: guvectorized functions can no longer be pickled in numba 0.53
This is a showstopper regression for users of guvectorized functions on dask.distributed:

```python
import numba
import pickle


@numba.guvectorize(["f8,f8[:]"], "()->()")
def double(x, out):
    out[:] = x * 2


print(pickle.dumps(double))
```

On Linux x64:
```bash
$ conda create -n test52 -c conda-forge python=3.8 numba=0.52
$ conda activate test52
$ python demo.py
b'\x80\x04\x95\x1a\x00\x00\x00\x00\x00\x00\x00\x8c\x0b__mp_main__\x94\x8c\x06double\x94\x93\x94.'
$ conda create -n test53 -c conda-forge python=3.8 numba=0.53
$ conda activate test53
$ python demo.py
  File "demo.py", line 10, in <module>
    print(pickle.dumps(double))
_pickle.PicklingError: Can't pickle <ufunc 'double'>: it's not the same object as __main__.double
```

sklam:
Fixes #6821.

### DIFF
--- a/numba/np/ufunc/gufunc.py
+++ b/numba/np/ufunc/gufunc.py
@@ -38,18 +38,20 @@ class GUFunc(serialize.ReduceMixin):
             is_dynamic=self._is_dynamic,
             targetoptions=gb.targetoptions,
             typesigs=gb._sigs,
+            frozen=self._frozen,
         )
         return dct
 
     @classmethod
     def _rebuild(cls, py_func, signature, identity, cache, is_dynamic,
-                 targetoptions, typesigs):
+                 targetoptions, typesigs, frozen):
         self = cls(py_func=py_func, signature=signature, identity=identity,
                    cache=cache, is_dynamic=is_dynamic,
                    targetoptions=targetoptions)
         for sig in typesigs:
             self.add(sig)
         self.build_ufunc()
+        self._frozen = frozen
         return self
 
     def __repr__(self):
--- a/numba/np/ufunc/gufunc.py
+++ b/numba/np/ufunc/gufunc.py
@@ -3,9 +3,10 @@ from numba.core import types
 from numba.np.ufunc.ufuncbuilder import GUFuncBuilder
 from numba.np.ufunc.sigparse import parse_signature
 from numba.np.numpy_support import ufunc_find_matching_loop
+from numba.core import serialize
 
 
-class GUFunc(object):
+class GUFunc(serialize.ReduceMixin):
     """
     Dynamic generalized universal function (GUFunc)
     intended to act like a normal Numpy gufunc, but capable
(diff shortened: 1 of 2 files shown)


numba/np/ufunc/gufunc.py
```diff
@@ -26,6 +28,30 @@ def __init__(self, py_func, signature, identity=None, cache=None,
...
+    def _reduce_states(self):
+        gb = self.gufunc_builder
+        dct = dict(
+            py_func=gb.py_func,
+            signature=gb.signature,
+            identity=self._identity,
+            cache=gb.cache,
+            is_dynamic=self._is_dynamic,
+            targetoptions=gb.targetoptions,
+            typesigs=gb._sigs,
```
stuartarchibald, line 40: Think `self._frozen` should be in here, whilst it has no impact on dynamic gufunc, it is transmittable state.

numba/tests/npyufunc/test_gufunc.py
```diff
@@ -281,5 +283,90 @@ class TestGUVectorizeScalarParallel(TestGUVectorizeScalar):
...
+        got = np.zeros(1)
+        double(0.5, out=expect)
+        cloned(0.5, out=got)
+        self.assertPreciseEqual(expect, got)
+        # array version
+        expect = np.zeros_like(arr)
+        got = np.zeros_like(arr)
+        double(arr, out=expect)
+        cloned(arr, out=got)
+        self.assertPreciseEqual(expect, got)
```
stuartarchibald, line 368: Wonder if these tests ought to also check, `self._frozen`, `self.gufunc_builder._sigs`, `self._is_dynamic` and `self._identity`, just to make sure the state is carrying appropriately?

stuartarchibald (commented):
Thanks for the patch, couple of comments to address else looks good.

numba/np/ufunc/gufunc.py
sklam: fixed

numba/tests/npyufunc/test_gufunc.py
sklam: These attributes are now tested explicitly

numba/tests/npyufunc/test_gufunc.py
```diff
@@ -281,5 +283,119 @@ class TestGUVectorizeScalarParallel(TestGUVectorizeScalar):
...
+        # attributes carried over
+        self.assertEqual(cloned._frozen, double._frozen)
+        self.assertEqual(cloned.identity, double.identity)
+        self.assertEqual(cloned.is_dynamic, double.is_dynamic)
+        self.assertEqual(cloned.gufunc_builder._sigs,
+                         double.gufunc_builder._sigs)
+        # expected value of attributes
+        self.assertTrue(cloned._frozen)
+
+        cloned.disable_compile()
```
stuartarchibald, line 307: I don't think this has any effect but is harmless.

stuartarchibald (approved):

PR #6828, opened 2021-03-15
Fix regression in CUDA: Set stream in mapped and managed array device_setup

### LINKED ISSUE #6233: CUDA: `mapped_array` accepts a stream but does nothing with it
`mapped_array` accepts a stream parameter:

```python
def mapped_array(shape, dtype=np.float, strides=None, order='C', stream=0,
                 portable=False, wc=False):
```

However, it only passes it to `mappedview.device_setup`:

```python
    managedview.device_setup(buffer, stream=stream)
```

which then does nothing with it:

```python
class MappedNDArray(DeviceNDArrayBase, np.ndarray):
    """ 
    A host array that uses CUDA mapped memory.
    """

    def device_setup(self, gpu_data, stream=0):
        self.gpu_data = gpu_data
```

I think this is misleading - it should either set its default stream to the provided stream (I think this is preferable, to facilitate the use of async operations for non-default streams) or not accept a stream parameter (if there's something preventing the use of async operations on these allocations anyway).

## Reporting a bug

### LINKED ISSUE #6827: CUDA Mapped Array Has No Object "stream" when attempting to access `__cuda_array_interface__`
This appears to be new with Numba 0.53, but when I run:

```python
from numba import cuda

sig = numba.cuda.mapped_array(2**18, dtype=float)
sig.__cuda_array_interface__['data']
```

I get the error: AttributeError: 'MappedNDArray' object has no attribute 'stream'.

When using Numba 0.52 or below, I see the following expected behavior:

```python
from numba import cuda
sig = numba.cuda.mapped_array(2**18, dtype=float)
sig.__cuda_array_interface__['data']
(140332461195264, False)
```

gmarkall:
Since #6827 is a regression in the CUDA Array Interface in 0.53, I'm suggesting this for 0.53.1 - this problem has been known about for a while (#6233) but only started causing a problem now that the stream is exported on the CUDA Array Interface.

Fixes #6827.
Fixes #6233.

@sklam @esc Could this have a buildfarm run if it's still available this week please?

cc @awthomp.

### DIFF
--- a/numba/cuda/cudadrv/devicearray.py
+++ b/numba/cuda/cudadrv/devicearray.py
@@ -681,6 +681,7 @@ class MappedNDArray(DeviceNDArrayBase, np.ndarray):
 
     def device_setup(self, gpu_data, stream=0):
         self.gpu_data = gpu_data
+        self.stream = stream
 
 
 class ManagedNDArray(DeviceNDArrayBase, np.ndarray):
@@ -690,6 +691,7 @@ class ManagedNDArray(DeviceNDArrayBase, np.ndarray):
 
     def device_setup(self, gpu_data, stream=0):
         self.gpu_data = gpu_data
+        self.stream = stream
 
 
 def from_array_like(ary, stream=0, gpu_data=None):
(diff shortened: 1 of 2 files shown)


stuartarchibald (approved):
Thanks for the fix, looks good.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_28`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_28`.

Passed.

PR #6829, opened 2021-03-16
Fix accuracy of np.arange and np.linspace

### LINKED ISSUE #6768: numpy.arange and numpy.linspace not accurate in numba
## Reporting a bug

- [x] I have tried using the latest released version of Numba (using `numba=0.52.0=py38h51da96c_0` from conda-forge which is the latest available version, also tested with `0.51.2` from the defaults channel).
  i.e. it's possible to run as 'python test_numba.py'.

I am getting different results in my code depending on if I run with numba or not. The problem is that the code that numba compiles is extremely simple (calling another compiled numba function in a for loop) but still one of the values comes out different. I have spent some time to create a small reproducible code which is attached below.

[test_numba.zip](https://github.com/numba/numba/files/6049357/test_numba.zip) which contains:
```
- test_numba.py
- env.yml # to reproduce the conda environment though 
#other versions of numba (including the one from the defaults channel) fail too

# 3 numpy arrays inside hdf5 files
- nonsep-epc.hdf5
- nonsep-negatives.hdf5
- nonsep-positives.hdf5

```

The related part of the code is (the `epc` function results are different from `epc.py_func`):
```python
@jit(nopython=True)
def min_weighted_error_rate_threshold(negatives, positives, cost):
    if cost > 1.0:
        cost = 1.0
    elif cost < 0.0:
        cost = 0.0

    return _minimizing_threshold(negatives, positives, cost)


@jit(nopython=True)
def epc(
    dev_negatives,
    dev_positives,
    n_points,
    is_sorted=False,
):
    # if not pre-sorted, copies and sorts
    dev_neg = dev_negatives if is_sorted else numpy.sort(dev_negatives)
    dev_pos = dev_positives if is_sorted else numpy.sort(dev_positives)
    step = 1.0 / (n_points - 1.0)
    alpha = numpy.arange(0, 1 + step, step)
   # something goes wrong from here
    thres = numpy.empty_like(alpha)
    for i, cost in enumerate(alpha):
        thres[i] = min_weighted_error_rate_threshold(dev_neg, dev_pos, cost)

    return thres, alpha
```
with this test function below written to test the `epc` function.
```python
def test():
    def _load(path):
        with h5py.File(path, "r") as f:
            return numpy.array(f["array"])

    positives = _load("nonsep-positives.hdf5")
    negatives = _load("nonsep-negatives.hdf5")
    thresholds_ref = _load("nonsep-epc.hdf5")[2]
    n_points = 100

    dev_negatives = negatives[: (negatives.shape[0] // 2)]
    dev_positives = positives[: (positives.shape[0] // 2)]

    print("Comparing pure Python to reference data.")
    thres, alpha_ref = epc.py_func(
        dev_negatives,
        dev_positives,
        n_points,
        is_sorted=False,
    )
    numpy.testing.assert_allclose(thres, thresholds_ref)
    print("All values are similar")

    print("Comparing sorted in Python to reference data.")
    thres, alpha = epc(
        numpy.sort(dev_negatives),
        numpy.sort(dev_positives),
        n_points,
        is_sorted=True,
    )
    numpy.testing.assert_allclose(alpha, alpha_ref)
    numpy.testing.assert_allclose(thres, thresholds_ref) # code fails here, this is line 181
    print("All values are similar")

    print("Comparing numba variant to reference data.")
    thres, alpha = epc(
        dev_negatives,
        dev_positives,
        n_points,
        is_sorted=False,
    )
    numpy.testing.assert_allclose(alpha, alpha_ref)
    numpy.testing.assert_allclose(thres, thresholds_ref) # it would fail here too if you comment the failing assertion above
    print("All values are similar")
```
If you run `python test_numba.py`, you will get:
```
$ python test_numba.py
Comparing pure Python to reference data.
All values are similar
Comparing sorted in Python to reference data.
  File "test_numba.py", line 197, in <module>
    test()
  File "test_numba.py", line 181, in test
    numpy.testing.assert_allclose(thres, thresholds_ref)
AssertionError: 
Not equal to tolerance rtol=1e-07, atol=0

Mismatched elements: 1 / 100 (1%)
Max absolute difference: 0.4
Max relative difference: 0.06837607
```

If I call `min_weighted_error_rate_threshold` (which is a compiled function) from Python, everything works. But, when it is called from the `epc` function, one of the values turns out to be wrong. I have checked all the computations inside the epc function:
```python
@jit(nopython=True)
def epc(
    dev_negatives,
    dev_positives,
    n_points,
    is_sorted=False,
):
    # if not pre-sorted, copies and sorts
    dev_neg = dev_negatives if is_sorted else numpy.sort(dev_negatives)
    dev_pos = dev_positives if is_sorted else numpy.sort(dev_positives)
    step = 1.0 / (n_points - 1.0)
    alpha = numpy.arange(0, 1 + step, step)
    thres = numpy.empty_like(alpha)
    for i, cost in enumerate(alpha):
        thres[i] = min_weighted_error_rate_threshold(dev_neg, dev_pos, cost)

    return thres, alpha
```
and looks like all the values are correct (according to my tests in the `test()` function) till the code reaches:
```python
    thres = numpy.empty_like(alpha)
    for i, cost in enumerate(alpha):
        thres[i] = min_weighted_error_rate_threshold(dev_neg, dev_pos, cost)
```
Here, numba is just calling another already compiled function, so I don't understand what could go wrong here when numba compiles it. Please let me know if I can provide more information.

stuartarchibald:
As title, this adjusts the implementations so as to not accumulate
error and to better replicate what NumPy does.

Fixes #6768

### DIFF
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -3936,7 +3936,7 @@ def numpy_linspace_3(context, builder, sig, args):
         div = num - 1
         if div > 0:
             delta = stop - start
-            step = np.divide(delta, div)
+            step = delta / div
             for i in range(0, num):
                 arr[i] = start + (i * step)
         else:
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -3907,8 +3907,7 @@ def np_arange(start, stop=None, step=None, dtype=None):
         arr = np.empty(nitems, true_dtype)
         val = _start
         for i in range(nitems):
-            arr[i] = val
-            val += _step
+            arr[i] = val + (i * _step)
         return arr
 
     return impl
@@ -3929,15 +3928,19 @@ def numpy_linspace_2(context, builder, sig, args):
 def numpy_linspace_3(context, builder, sig, args):
     dtype = as_dtype(sig.return_type.dtype)
 
+    # Implementation based on https://github.com/numpy/numpy/blob/v1.20.0/numpy/core/function_base.py#L24 # noqa: E501
     def linspace(start, stop, num):
         arr = np.empty(num, dtype)
         if num == 0:
             return arr
         div = num - 1
-        delta = stop - start
-        arr[0] = start
-        for i in range(1, num):
-            arr[i] = start + delta * (i / div)
+        if div > 0:
+            delta = stop - start
+            step = np.divide(delta, div)
+            for i in range(0, num):
+                arr[i] = start + (i * step)
+        else:
+            arr[0] = start
         return arr
 
     res = context.compile_internal(builder, linspace, sig, args)
--- a/numba/tests/test_array_methods.py
+++ b/numba/tests/test_array_methods.py
@@ -922,6 +922,16 @@ class TestArrayMethods(MemoryLeakMixin, TestCase):
                     self.assertIn("Maximum allowed size exceeded",
                                 str(raises.exception))
 
+    def test_arange_accuracy(self):
+        # Checking arange reasonably replicates NumPy's algorithm
+        # see https://github.com/numba/numba/issues/6768
+        @jit(nopython=True)
+        def foo(step):
+            return np.arange(0, 1 + step, step)
+
+        x = 0.010101010101010102
+        self.assertPreciseEqual(foo(x), foo.py_func(x))
+
     def test_item(self):
         pyfunc = array_item
         cfunc = jit(nopython=True)(pyfunc)
--- a/numba/tests/test_dyn_array.py
+++ b/numba/tests/test_dyn_array.py
@@ -1234,7 +1234,7 @@ class TestNdDiag(TestCase):
             dfunc = nrtjit(self.py_kw)
             dfunc(d, k=3)
 
-class TestNdArange(BaseTest):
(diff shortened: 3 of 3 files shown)


numba/np/arrayobj.py
```diff
@@ -3929,15 +3928,19 @@ def linspace(start, stop):
...
         if num == 0:
             return arr
         div = num - 1
-        delta = stop - start
-        arr[0] = start
-        for i in range(1, num):
-            arr[i] = start + delta * (i / div)
+        if div > 0:
+            delta = stop - start
+            step = np.divide(delta, div)
```
sklam, line 3939: Is `np.divide` used to make sure numpy error model here?

stuartarchibald: Yes, but it's probably not necessary as the predicate would likely propagate and the div-by-zero branch would be DCE. Am happy to change.

sklam: i'd prefer the using `/`

stuartarchibald: Done in 947c744

sklam (approved):
Thanks for the patch.

PR #6830, opened 2021-03-16
CUDA: Use relaxed strides checking to compute contiguity

### LINKED ISSUE #2766: DummyArray Contiguous Flags Are Wrong
````python
import numba.dummyarray
import numpy as np
z = np.empty((1, 2, 3))
z = np.transpose(z, axes=(2, 0, 1))
arr = numba.dummyarray.Array.from_desc(0, z.shape, z.strides, z.itemsize)
assert z.flags['C_CONTIGUOUS'] == arr.flags['C_CONTIGUOUS']
assert z.flags['F_CONTIGUOUS'] == arr.flags['F_CONTIGUOUS']
````

...gives...

````
<ipython-input-8-d132ae3d3b84> in <module>()
      5 arr = numba.dummyarray.Array.from_desc(0, z.shape, z.strides, z.itemsize)
      6 assert z.flags['C_CONTIGUOUS'] == arr.flags['C_CONTIGUOUS']
----> 7 assert z.flags['F_CONTIGUOUS'] == arr.flags['F_CONTIGUOUS']

AssertionError: 
````

### LINKED ISSUE #4943: DeviceNDArray contiguity differs from Numpy array contiguity
## Reporting a bug

Spotted whilst looking into #4628. Similar to #2766, but for `DeviceNDArrays`, the continuity disagrees with Numpy. I'm presently under the impression that this is a bug and not the result of some semantic difference when data is moved over to a CUDA device, so I'm looking into a fix to make them agree. The example code:

```python
from numba import cuda
import numpy as np

def test(shape, order):
    arr = np.ndarray(shape, order=order)
    d_arr = cuda.to_device(arr)
    print("Shape: %s, Order: %s" % (shape, order))
    print("Host:   C: %s, F: %s" % (arr.flags['C_CONTIGUOUS'], arr.flags['F_CONTIGUOUS']))
    print("Device: C: %s, F: %s" % (d_arr.flags['C_CONTIGUOUS'], d_arr.flags['F_CONTIGUOUS']))
    print()

test((1,4), 'C')
test((1,4), 'F')
test((4,1), 'C')
test((4,1), 'F')
```

produces:

```
Shape: (1, 4), Order: C
Host:   C: True, F: True
Device: C: True, F: False

Shape: (1, 4), Order: F
Host:   C: True, F: True
Device: C: False, F: True

Shape: (4, 1), Order: C
Host:   C: True, F: True
Device: C: True, F: False

Shape: (4, 1), Order: F
Host:   C: True, F: True
Device: C: False, F: True
```

### LINKED ISSUE #6824: Array contains non-contiguous buffer error even after using np.ascontiguousarray
## Reporting a bug

It seems numba raises `"Array contains non-contiguous buffer"` error even after using `np.ascontiguousarray`.

### Minimal example:

```python
import dask.array as da
import numpy as np
from numba import guvectorize

x = da.arange(10000, dtype=np.float64).reshape(100, 100).rechunk((1, 10))
f = np.ascontiguousarray(x.blocks[0, 0])
g = np.ascontiguousarray(x.blocks[0, 0])

@guvectorize(["void(float64[:], float64[:], float64[:], float64[:])"],
    "(n),(n),(p)->(p)",
    nopython=True,
    target="cuda",
)

def gpu_error(x, y, _, out) -> None:
    sum_ = 0.0
    for i in range(x.shape[0]):
        sum_ += (x[i] + y[i])
    out[:] = sum_

gpu_error(f, g, np.empty(1, dtype=x.dtype))
```

### Traceback

```python
<ipython-input-29-ed5511acd179> in <module>
----> 1 gpu_error(f, g, np.empty(1, dtype=x.dtype))

~/.conda/envs/sgkit-gpu/lib/python3.8/site-packages/numba/np/ufunc/deviceufunc.py in __call__(self, *args, **kws)
    651         callsteps.adjust_input_types(indtypes)
    652         callsteps.allocate_outputs(schedule, outdtype)
--> 653         callsteps.prepare_kernel_parameters()
    654         newparams, newretval = self._broadcast(schedule,
    655                                                callsteps.kernel_parameters,

~/.conda/envs/sgkit-gpu/lib/python3.8/site-packages/numba/np/ufunc/deviceufunc.py in prepare_kernel_parameters(self)
    810                 params.append(inp)
    811             else:
--> 812                 params.append(self.to_device(inp))
    813         assert all(self.is_device_array(a) for a in params)
    814         self.kernel_parameters = params

~/.conda/envs/sgkit-gpu/lib/python3.8/site-packages/numba/cuda/dispatcher.py in to_device(self, hostary)
    102 
    103     def to_device(self, hostary):
--> 104         return cuda.to_device(hostary, stream=self._stream)
    105 
    106     def to_host(self, devary, hostary):

~/.conda/envs/sgkit-gpu/lib/python3.8/site-packages/numba/cuda/cudadrv/devices.py in _require_cuda_context(*args, **kws)
    222     def _require_cuda_context(*args, **kws):
    223         with _runtime.ensure_context():
--> 224             return fn(*args, **kws)
    225 
    226     return _require_cuda_context

~/.conda/envs/sgkit-gpu/lib/python3.8/site-packages/numba/cuda/api.py in to_device(obj, stream, copy, to)
    102     """
    103     if to is None:
--> 104         to, new = devicearray.auto_device(obj, stream=stream, copy=copy)
    105         return to
    106     if copy:

~/.conda/envs/sgkit-gpu/lib/python3.8/site-packages/numba/cuda/cudadrv/devicearray.py in auto_device(obj, stream, copy)
    774             devobj = from_array_like(obj, stream=stream)
    775         if copy:
--> 776             devobj.copy_to_device(obj, stream=stream)
    777         return devobj, True
    778 

~/.conda/envs/sgkit-gpu/lib/python3.8/site-packages/numba/cuda/cudadrv/devices.py in _require_cuda_context(*args, **kws)
    222     def _require_cuda_context(*args, **kws):
    223         with _runtime.ensure_context():
--> 224             return fn(*args, **kws)
    225 
    226     return _require_cuda_context

~/.conda/envs/sgkit-gpu/lib/python3.8/site-packages/numba/cuda/cudadrv/devicearray.py in copy_to_device(self, ary, stream)
    208             return
    209 
--> 210         sentry_contiguous(self)
    211         stream = self._default_stream(stream)
    212         self_core, ary_core = array_core(self), array_core(ary)

~/.conda/envs/sgkit-gpu/lib/python3.8/site-packages/numba/cuda/cudadrv/devicearray.py in sentry_contiguous(ary)
    745     core = array_core(ary)
    746     if not core.flags['C_CONTIGUOUS'] and not core.flags['F_CONTIGUOUS']:
--> 747         raise ValueError(errmsg_contiguous_buffer)
    748 
    749 

ValueError: Array contains non-contiguous buffer and cannot be transferred as a single memory region. Please ensure contiguous buffer with numpy .ascontiguousarray()
```

### Array flags

```python
f.flags

  C_CONTIGUOUS : True
  F_CONTIGUOUS : True
  OWNDATA : False
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False
```

```python
g.flags

  C_CONTIGUOUS : True
  F_CONTIGUOUS : True
  OWNDATA : False
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False
```

gmarkall:
Since NumPy 1.8, a relaxed strides check has been available for computing the contiguity of arrays (and the relaxed check is the default starting from 1.12). The relaxed check permits additional conditions for arrays to be considered contiguous:

- All 0-size arrays are considered contiguous, even if they are multidimensional.
- Strides for an axis are ignored if its shape is 1 when checking contiguity.

Examples of shapes and strides that would now be considered contiguous:

- Shape: (1, 10), Strides: (800, 80): because the first axis has size 1.
- Shape: (1, 0, 2), Strides (64, 16, 8): because the second axis has size 0.

This PR makes relaxed strides checking the default when computing the C- and F-contiguity of device arrays, and enables more use cases. For example, from the reproducer in Issue #6824:

```python
import dask.array as da
import numpy as np
from numba import guvectorize

x = da.arange(10000, dtype=np.float64).reshape(100, 100).rechunk((1, 10))
f = np.ascontiguousarray(x.blocks[0, 0])
```

`f` is an array that is only contiguous under relaxed strides checking, and prior to this commit could not be transferred to a device array.

Since this may expose latent bugs (the switch in NumPy exposed some issues between versions 1.10.0 and 1.10.2, for example), a config variable, `NPY_RELAXED_STRIDES_CHECKING` is provided to restore the old, strict checking. This is provided on a temporary basis in case any bugs need to be worked out, and will be removed in 0.55 in the current deprecation schedule.

Fixes #6824.
Fixes #4943.
Fixes #2766.

### DIFF
--- a/docs/source/reference/deprecation.rst
+++ b/docs/source/reference/deprecation.rst
@@ -306,3 +306,33 @@ In 0.54.0:
 - ``ptx`` and the inspection methods will always return a dict.
 - Support for indexing into the results of these methods using ``(cc,
   argtypes)`` will be removed.
+
+
+.. _deprecation-strict-strides:
+
+Deprecation of strict strides checking when computing contiguity
+================================================================
+
+The contiguity of device arrays (the ``'C_CONTIGUOUS'`` and ``'F_CONTIGUOUS'``
+elements of the flags of a device array) are computed using relaxed strides
+checking, which matches the default in NumPy since Version 1.12. A config
+variable, :envvar:`NUMBA_NPY_RELAXED_STRIDES_CHECKING`, is provided to force
+computation of these flags using strict strides checking.
+
+This flag is provided to work around any bugs that may be exposed by strict
+strides checking, and will be removed in future.
+
+Schedule
+--------
+
+In 0.54.0:
+
+- Relaxed strides checking will become the default.
+- Strict strides checking will be deprecated.
+
+In 0.55.0:
+
+- Strict strides checking will be removed, if no reports of bugs related to
+  relaxed strides checking in 0.54.0 onwards. This plan will be re-examined if
+  bugs related to relaxed strides checking are reported, but may not necessarily
+  change as a result.
--- a/docs/source/reference/deprecation.rst
+++ b/docs/source/reference/deprecation.rst
@@ -332,7 +332,7 @@ In 0.54.0:
 
 In 0.55.0:
 
-- Strict strides checking will be removed, if no reports of bugs related to
-  relaxed strides checking in 0.54.0 onwards. This plan will be re-examined if
-  bugs related to relaxed strides checking are reported, but may not necessarily
-  change as a result.
+- Strict strides checking will be removed, if there are no reports of bugs
+  related to relaxed strides checking in 0.54.0 onwards. This plan will be
+  re-examined if bugs related to relaxed strides checking are reported, but may
+  not necessarily change as a result.
--- a/docs/source/reference/envvars.rst
+++ b/docs/source/reference/envvars.rst
@@ -430,6 +430,20 @@ GPU support
    Setting this variable to 1 also includes the values of arguments to Driver
    API calls in the logs.
 
+.. envvar:: NUMBA_NPY_RELAXED_STRIDES_CHECKING
+
+   By default Numba device arrays compute their contiguity using relaxed strides
+   checking, which is the default mechanism used by NumPy since version 1.12
+   (see `NPY_RELAXED_STRIDES_CHECKING
+   <https://numpy.org/doc/stable/release/1.8.0-notes.html?highlight=numpy%20sum#npy-relaxed-strides-checking>`_).
+   Setting ``NUMBA_NPY_RELAXED_STRIDES_CHECKING=0`` reverts back to strict
+   strides checking. This option should not normally be needed, but is provided
+   in case it is needed to workaround latent bugs related to strict strides
+   checking.
+
+   Strict strides checking is deprecated and may be removed in future. See
+   :ref:`deprecation-strict-strides`.
+
 
 Threading Control
 -----------------
--- a/docs/source/reference/envvars.rst
+++ b/docs/source/reference/envvars.rst
@@ -432,13 +432,14 @@ GPU support
 
 .. envvar:: NUMBA_NPY_RELAXED_STRIDES_CHECKING
 
-   By default Numba device arrays compute their contiguity using relaxed strides
+   By default arrays that inherit from ``numba.misc.dummyarray.Array` (e.g.
+   CUDA and ROCm device arrays) compute their contiguity using relaxed strides
    checking, which is the default mechanism used by NumPy since version 1.12
    (see `NPY_RELAXED_STRIDES_CHECKING
-   <https://numpy.org/doc/stable/release/1.8.0-notes.html?highlight=numpy%20sum#npy-relaxed-strides-checking>`_).
+   <https://numpy.org/doc/stable/release/1.8.0-notes.html#npy-relaxed-strides-checking>`_).
    Setting ``NUMBA_NPY_RELAXED_STRIDES_CHECKING=0`` reverts back to strict
    strides checking. This option should not normally be needed, but is provided
-   in case it is needed to workaround latent bugs related to strict strides
+   in case it is needed to work around latent bugs related to strict strides
    checking.
 
    Strict strides checking is deprecated and may be removed in future. See
--- a/docs/source/reference/envvars.rst
+++ b/docs/source/reference/envvars.rst
@@ -432,7 +432,7 @@ GPU support
 
 .. envvar:: NUMBA_NPY_RELAXED_STRIDES_CHECKING
 
-   By default arrays that inherit from ``numba.misc.dummyarray.Array` (e.g.
+   By default arrays that inherit from ``numba.misc.dummyarray.Array`` (e.g.
    CUDA and ROCm device arrays) compute their contiguity using relaxed strides
    checking, which is the default mechanism used by NumPy since version 1.12
    (see `NPY_RELAXED_STRIDES_CHECKING
--- a/numba/cuda/tests/cudadrv/test_cuda_ndarray.py
+++ b/numba/cuda/tests/cudadrv/test_cuda_ndarray.py
@@ -1,8 +1,10 @@
+import itertools
 import numpy as np
 from numba.cuda.cudadrv import devicearray
 from numba import cuda
 from numba.cuda.testing import unittest, CUDATestCase
 from numba.cuda.testing import skip_on_cudasim
+from numba.tests.support import override_config
 
 
 class TestCudaNDArray(CUDATestCase):
@@ -279,33 +281,6 @@ class TestCudaNDArray(CUDATestCase):
         self.assertTrue(np.all(d.copy_to_host() == a_f))
 
     def test_devicearray_broadcast_host_copy(self):
-        try:
-            broadcast_to = np.broadcast_to
-        except AttributeError:
-            # numpy<1.10 doesn't have broadcast_to. The following implements
-            # a limited broadcast_to that only works along already existing
-            # dimensions of length 1.
-            def broadcast_to(arr, new_shape):
-                new_strides = []
-                for new_length, length, stride in zip(
-                    new_shape, arr.shape, arr.strides
-                ):
-                    if length == 1 and new_length > 1:
-                        new_strides.append(0)
-                    elif new_length == length:
-                        new_strides.append(stride)
-                    else:
-                        raise ValueError(
-                            "cannot broadcast shape {} to shape {}"
-                            .format(arr.shape, new_shape)
-                        )
-                return np.ndarray(
-                    buffer=np.squeeze(arr),
-                    dtype=arr.dtype,
-                    shape=new_shape,
-                    strides=tuple(new_strides),
-                )
-
         broadsize = 4
         coreshape = (2, 3)
         coresize = np.prod(coreshape)
@@ -314,8 +289,8 @@ class TestCudaNDArray(CUDATestCase):
         for dim in range(len(coreshape)):
             newindex = (slice(None),) * dim + (np.newaxis,)
             broadshape = coreshape[:dim] + (broadsize,) + coreshape[dim:]
-            broad_c = broadcast_to(core_c[newindex], broadshape)
-            broad_f = broadcast_to(core_f[newindex], broadshape)
+            broad_c = np.broadcast_to(core_c[newindex], broadshape)
+            broad_f = np.broadcast_to(core_f[newindex], broadshape)
             dbroad_c = cuda.to_device(broad_c)
             dbroad_f = cuda.to_device(broad_f)
             np.testing.assert_array_equal(dbroad_c.copy_to_host(), broad_c)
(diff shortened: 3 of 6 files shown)


docs/source/reference/envvars.rst
```diff
@@ -430,6 +430,20 @@ GPU support
...
 
+.. envvar:: NUMBA_NPY_RELAXED_STRIDES_CHECKING
+
+   By default Numba device arrays compute their contiguity using relaxed strides
+   checking, which is the default mechanism used by NumPy since version 1.12
+   (see `NPY_RELAXED_STRIDES_CHECKING
+   <https://numpy.org/doc/stable/release/1.8.0-notes.html?highlight=numpy%20sum#npy-relaxed-strides-checking>`_).
+   Setting ``NUMBA_NPY_RELAXED_STRIDES_CHECKING=0`` reverts back to strict
+   strides checking. This option should not normally be needed, but is provided
+   in case it is needed to workaround latent bugs related to strict strides
```
stuartarchibald, line 441:
```suggestion
   in case it is needed to work around latent bugs related to strict strides
```

docs/source/reference/envvars.rst
```diff
@@ -430,6 +430,20 @@ GPU support
    Setting this variable to 1 also includes the values of arguments to Driver
    API calls in the logs.
 
+.. envvar:: NUMBA_NPY_RELAXED_STRIDES_CHECKING
+
+   By default Numba device arrays compute their contiguity using relaxed strides
```
stuartarchibald, line 435: This will impact ROCm and anything else that inherits from `numba.misc.dummyarray.Array`? Perhaps:
```
By default arrays which inherit from `numba.misc.dummyarray.Array` (e.g. CUDA and ROCm device arrays) compute their...
```
?

docs/source/reference/envvars.rst
```diff
@@ -430,6 +430,20 @@ GPU support
...
+   checking, which is the default mechanism used by NumPy since version 1.12
+   (see `NPY_RELAXED_STRIDES_CHECKING
+   <https://numpy.org/doc/stable/release/1.8.0-notes.html?highlight=numpy%20sum#npy-relaxed-strides-checking>`_).
```
stuartarchibald, line 438:
```suggestion
   <https://numpy.org/doc/stable/release/1.8.0-notes.html#npy-relaxed-strides-checking>`_).
```

docs/source/reference/deprecation.rst
```diff
@@ -306,3 +306,33 @@ In 0.54.0:
...
+--------
+
+In 0.54.0:
+
+- Relaxed strides checking will become the default.
+- Strict strides checking will be deprecated.
+
+In 0.55.0:
+
+- Strict strides checking will be removed, if no reports of bugs related to
```
stuartarchibald, line 335:
```suggestion
- Strict strides checking will be removed, if there are no reports of bugs related to
```

numba/misc/dummyarray.py
```diff
@@ -164,7 +165,47 @@ def __init__(self, dims, itemsize):
...
-        self.flags = self._compute_layout()
+        if config.NPY_RELAXED_STRIDES_CHECKING:
+            self.flags = self._relaxed_compute_layout()
+        else:
+            self.flags = self._compute_layout()
+
+    def _relaxed_compute_layout(self):
+        # The logic here is based on that in _UpdateContiguousFlags from
+        # numpy/core/src/multiarray/flagsobject.c in NumPy v1.19.1 (commit
+        # 13661ac70).
```
stuartarchibald, line 176: Is it possible to get an exact URL with the SHA and source section in it?

stuartarchibald: Maybe https://github.com/numpy/numpy/blob/maintenance/1.19.x/numpy/core/src/multiarray/flagsobject.c#L123-L191 ?

numba/misc/dummyarray.py
```diff
@@ -164,7 +165,47 @@ def __init__(self, dims, itemsize):
...
+        # Check F contiguity
+        sd = self.itemsize
+        for dim in self.dims:
+            if dim.size != 1:
+                if dim.stride != sd:
+                    flags['F_CONTIGUOUS'] = False
+                    return flags
+                sd *= dim.size
+
+        return flags
```
stuartarchibald, line 208: On inspection this section matches the logic in the NumPy C code referred to in the comment at the top of the function.

numba/cuda/tests/cudadrv/test_cuda_ndarray.py
```diff
@@ -343,6 +318,49 @@ def test_devicearray_contiguous_device_strided(self):
...
+            # Ensure we still believe the array to be contiguous.
+            self.assertTrue(arr.flags['C_CONTIGUOUS'])
+            self.assertTrue(arr.flags['F_CONTIGUOUS'])
+
+    @skip_on_cudasim('DeviceNDArray class not present in simulator')
+    def test_devicearray_strict_strides(self):
+        # From the reproducer in Issue #6824.
+
+        with override_config('NPY_RELAXED_STRIDES_CHECKING', 0):
+            # Construct a device array that is not contiguous because
```
stuartarchibald, line 341:
```suggestion
            # Construct a device array that is not contiguous under strict stride checking because
```

numba/cuda/tests/cudadrv/test_cuda_ndarray.py
```diff
@@ -343,6 +318,49 @@ def test_devicearray_contiguous_device_strided(self):
             devicearray.errmsg_contiguous_buffer,
             str(e.exception))
 
+    @skip_on_cudasim('DeviceNDArray class not present in simulator')
+    def test_devicearray_relaxed_strides(self):
+        # From the reproducer in Issue #6824.
+
+        with override_config('NPY_RELAXED_STRIDES_CHECKING', 1):
+            # Construct a device array that is contiguous even though
```
stuartarchibald, line 326: Perhaps add something about it being considered contiguous as the stride checking is relaxed?

numba/misc/dummyarray.py
```diff
@@ -164,7 +165,47 @@ def __init__(self, dims, itemsize):
...
+        # numpy/core/src/multiarray/flagsobject.c in NumPy v1.19.1 (commit
+        # 13661ac70).
+
+        # Records have no dims, and we can treat them as contiguous
+        if not self.dims:
+            return {'C_CONTIGUOUS': True, 'F_CONTIGUOUS': True}
+
+        # If this is a broadcast array then it is not contiguous
+        if any([dim.stride == 0 for dim in self.dims]):
+            return {'C_CONTIGUOUS': False, 'F_CONTIGUOUS': False}
```
stuartarchibald, line 184: Are these branches tested?

stuartarchibald (commented):
Thanks for the patch and for working on a fix for this issue. In general it seems good, just a few minor comments/queries to look at. Thanks again!

docs/source/reference/envvars.rst
gmarkall: Oops!

numba/misc/dummyarray.py
gmarkall: Yes, the test is `numba.cuda.tests.cudadrv.test_cuda.ndarray.TestCudaNDArray.test_devicearray_broadcast_host_copy`. The test seems quite sensitive to mistakes in contiguity computation - it took a while to arrive at the correct logic, which is partly why I wanted to have a flag to go back to strict strides checking - this function is correct to the best of my ability and knowledge now, but there may well be blind spots in the test suite and my knowledge.

numba/misc/dummyarray.py
gmarkall: Thanks for checking - even though it doesn't translate to perfectly elegant Python, I tried to keep it close to the original NumPy for easier comparison / future debugging.

numba/cuda/tests/cudadrv/test_cuda_ndarray.py
gmarkall: I've added a note in the comment below before the asserts, because this sentence is already full of clauses and tricky to comprehend.

numba/cuda/tests/cudadrv/test_cuda_ndarray.py
gmarkall: I've added a comment before the asserts below instead (and also corrected it - the expectation was inverted, as we expect the array not to be contiguous, contrary to the original comment.

numba/misc/dummyarray.py
gmarkall: Added.

gmarkall:
@stuartarchibald Many thanks for the review - comments addressed / responded to!

docs/source/reference/envvars.rst
```diff
@@ -432,13 +432,14 @@ GPU support
 
 .. envvar:: NUMBA_NPY_RELAXED_STRIDES_CHECKING
 
-   By default Numba device arrays compute their contiguity using relaxed strides
+   By default arrays that inherit from ``numba.misc.dummyarray.Array` (e.g.
```
stuartarchibald, line 435: The number of backticks here doesn't add up?

gmarkall: Fixed.

stuartarchibald: thanks

stuartarchibald (approved):
Thanks for the fixes.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_29`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_29`.


Passed.

Issue #6656, opened 2021-01-21
numpy constructors aren't typing correctly, invalid cast issued


stuartarchibald:
## Reporting a bug

- [x] I have tried using the latest released version of Numba (most recent is
 visible in the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [x] I have included a self contained code sample to reproduce the problem.
  i.e. it's possible to run as 'python bug.py'.

This:
```python
from numba import njit
import numpy as np

@njit
def foo():
    np.int32(np.zeros((4,)))

foo()
```
does this:
```
numba/core/base.py", line 716, in cast
    raise NotImplementedError(
numba.core.errors.LoweringError: Failed in nopython mode pipeline (step: nopython mode backend)
Cannot cast array(float64, 1d, C) to int32: %".113" = load {i8*, i8*, i64, i64, double*, [1 x i64], [1 x i64]}, {i8*, i8*, i64, i64, double*, [1 x i64], [1 x i64]}* %"$12call_method.5"
```
Typing should catch this, it shouldn't have survived until lowering time.

stuartarchibald:
Starter patch:
```patch
diff --git a/numba/core/typing/builtins.py b/numba/core/typing/builtins.py
index c91078a..ede6fd9 100644
--- a/numba/core/typing/builtins.py
+++ b/numba/core/typing/builtins.py
@@ -800,9 +800,16 @@ class NumberClassAttribute(AttributeTemplate):
                 sig = self.context.resolve_function_type(
                     np.array, (val,), {'dtype': types.DType(ty)})
                 return sig.return_type
-            else:
+            elif isinstance(val, (types.Number, types.Boolean)):
                 # Scalar constructor, e.g. np.int32(42)
                 return ty
+            else:
+                # unsupported
+                msg = f"Casting {val} to {ty} directly is unsupported."
+                if isinstance(val, types.Array):
+                    # array casts are supported a different way.
+                    msg += f" Try doing '<array>.astype(np.{ty})' instead"
+                raise errors.TypingError(msg)
 
         return types.Function(make_callable_template(key=ty, typer=typer))
```

PR #6833, opened 2021-03-17
Raise TypeError exception if numpy array is cast to scalar

testhound:
This pull request addresses https://github.com/numba/numba/issues/6656 (numpy constructors aren't typing correctly, invalid cast issued). The code uses the code introduced by Graham in that issue and adds a test case.


stuartarchibald:
Thanks for the patch @testhound, I think this contains my code so someone who is not me needs to take a look! CC @esc @sklam .

numba/tests/test_casting.py
```diff
@@ -80,6 +84,19 @@ def driver(x):
...
+        Ensure that a TypingError exception is raised if
+        user tries to convert numpy array to scalar
+        """
+
+        try:
+            compile_isolated(numpy_scalar_cast_error, ())
+        except TypingError as e:
+            self.assertIn("Casting array", str(e))
+        else:
+            self.fail("Should raise error")
```
sklam, line 98: You can use [assertRaises](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises) to avoid having to do an explicit `try..except` and `self.fail()`.

testhound: @sklam thanks for the comment, code has been updated.

sklam: we should still have the `assertIn` to make sure it is getting the right exception. 

see example: https://github.com/numba/numba/blob/bb2ec93fe107339d752bb03bde49f505fa2ed02c/numba/cuda/tests/cudadrv/test_context_stack.py#L57-L60

testhound: @sklam your request was incorporated in the most recent commits.

Issue #6615, opened 2021-01-04
FakeCUDAKernel doesn't take "debug"


shaunc:
Using the CUDA simulator:

https://github.com/numba/numba/blob/ba0a6c8e7a7e33ad812bc644b29d2dfd8f3f4f0d/numba/cuda/decorators.py#L81-L82

and several other places pass debug=debug to `FakeCUDAKernel`, but it doesn't take `debug` option:

https://github.com/numba/numba/blob/ba0a6c8e7a7e33ad812bc644b29d2dfd8f3f4f0d/numba/cuda/simulator/kernel.py#L58

esc:
@shaunc thank you for reporting this. Is there a specific use-case that breaks because of this?

shaunc:
Hmm... well I'm now getting this whenever I use CUDASIM. However, I'm not sure exactly what is triggering it. I used CUDASIM about a month ago without problem, and both added a fair amount of code to my repo and also upgraded from 0.51.0 (...ish - I think)

stuartarchibald:
Any chance of a reproducer please, just to make sure everyone is discussing the same thing. Thanks.

shaunc:
I'd suggest you just need to extend your test suite for  to cover `config.ENABLE_CUDASIM`. It looks to me as if there isn't non-error any path through `@jit` wrapper in this case. (Presumably if coverage worked with numba, you would have caught it yourself -- that would be a feature request, but I think there is already an open issue for it.)

stuartarchibald:
I suspect this could be involved.
```
$ python -c 'from numba import config; config.ENABLE_CUDASIM=1; from numba.cuda import jit; print(jit.__module__)'
numba.cuda.simulator.api

$ python -c 'from numba import config; config.ENABLE_CUDASIM=0; from numba.cuda import jit; print(jit.__module__)'
numba.cuda.decorators
```
The test suite already runs with `NUMBA_ENABLE_CUDASIM`:
https://github.com/numba/numba/blob/ba0a6c8e7a7e33ad812bc644b29d2dfd8f3f4f0d/buildscripts/incremental/test.sh#L82-L84
which makes me think that something else is quite possibly going on, either an untested path or an unusual set of circumstances.

A way to reproduce would be really useful.

stuartarchibald:
[...]

PR #6834, opened 2021-03-17
Remove illegal "debug" kw argument

shaunc:
Fix for https://github.com/numba/numba/issues/6615

Removes illegal argument from call to  `FakeCUDAKernel`. Hard to trigger in test -- but currently these lines are untested and clearly illegal, as `FakeCUDAKernel` does not take the `debug` keyword.

With these lines removed, works with my test suite in with `NUMBA_ENABLE_CUDASIM` defined; otherwise doesn't work.

Relevant configuration in conftests.py:
```python
def pytest_configure(config):
    """
    Implement configuration options.
    * If "CUDA" is specified, make sure cuda simulator
      is turned on before collection if real cuda not available.

    * remove featureset cache if specified
    """
    cudasim = config.getoption("--cudasim")
    if cudasim:
        os.environ["NUMBA_ENABLE_CUDASIM"] = "1"
        config.option.cuda = "true"
        return
    cuda = config.getoption("--cuda")
    if cuda:
        check_cuda()


def check_cuda():
    """
    Check if CUDA simulator is necessary.

    Use CUDA simulator if CUDA not available.

    We have to call in subprocess because load of `cuda`
    module requires NUMBA_ENABLE_CUDASIM to be set first.
    """
    available = (
        subprocess.check_output(
            [
                "python",
                "-c",
                "from numba import cuda; print(cuda.is_available())",
            ]
        )
        .decode()
        .strip()
    )
    if available != "True":
        os.environ["NUMBA_ENABLE_CUDASIM"] = "1"
        print(f"Using CUDA Simulator (available: {available})")
    else:
        # make sure we use appropriate device
        from numba import cuda

        c_dev = cuda.get_current_device()
        cd_cc = c_dev.compute_capability
        assert cd_cc[0] >= 8 and (cd_cc[0] > 8 or cd_cc[1] >= 6)

```

### DIFF
--- a/numba/cuda/decorators.py
+++ b/numba/cuda/decorators.py
@@ -76,8 +76,7 @@ def jit(func_or_sig=None, device=False, inline=False, link=[], debug=None,
     if sigutils.is_signature(func_or_sig):
         if config.ENABLE_CUDASIM:
             def jitwrapper(func):
-                return FakeCUDAKernel(func, device=device, fastmath=fastmath,
-                                      debug=debug)
+                return FakeCUDAKernel(func, device=device, fastmath=fastmath)
             return jitwrapper
 
         argtypes, restype = sigutils.normalize_signature(func_or_sig)
--- a/numba/cuda/decorators.py
+++ b/numba/cuda/decorators.py
@@ -106,7 +106,7 @@ def jit(func_or_sig=None, device=False, inline=False, link=[], debug=None,
             if config.ENABLE_CUDASIM:
                 def autojitwrapper(func):
                     return FakeCUDAKernel(func, device=device,
-                                          fastmath=fastmath, debug=debug)
+                                          fastmath=fastmath)
             else:
                 def autojitwrapper(func):
                     return jit(func, device=device, debug=debug, opt=opt, **kws)
@@ -116,7 +116,7 @@ def jit(func_or_sig=None, device=False, inline=False, link=[], debug=None,
         else:
             if config.ENABLE_CUDASIM:
                 return FakeCUDAKernel(func_or_sig, device=device,
-                                      fastmath=fastmath, debug=debug)
+                                      fastmath=fastmath)
             elif device:
                 return jitdevice(func_or_sig, debug=debug, opt=opt, **kws)
             else:
--- a/numba/cuda/tests/cudapy/test_cuda_jit_no_types.py
+++ b/numba/cuda/tests/cudapy/test_cuda_jit_no_types.py
@@ -1,6 +1,7 @@
 from numba import cuda
 import numpy as np
 from numba.cuda.testing import CUDATestCase
+from numba.tests.support import override_config
 import unittest
 
 
@@ -76,6 +77,13 @@ class TestCudaJitNoTypes(CUDATestCase):
 
         self.assertEqual(b[0], (a[0] + 1) + (2 + 1))
 
+    def test_jit_debug_simulator(self):
+        # Ensure that the jit decorator accepts the debug kwarg when the
+        # simulator is in use - see Issue #6615.
+        with override_config('ENABLE_CUDASIM', 1):
+            @cuda.jit(debug=True)
+            def f(x):
+                pass
 
 if __name__ == '__main__':
     unittest.main()
--- a/numba/cuda/tests/cudapy/test_cuda_jit_no_types.py
+++ b/numba/cuda/tests/cudapy/test_cuda_jit_no_types.py
@@ -85,5 +85,6 @@ class TestCudaJitNoTypes(CUDATestCase):
             def f(x):
                 pass
 
+
 if __name__ == '__main__':
     unittest.main()


stuartarchibald:
Thanks for the PR @shaunc.

@gmarkall please could you review, thanks.

gmarkall (changes requested):
Thanks for the fix! This passes tests for me with the simulator and hardware.

I wasn't able to work out exactly what to do with conftest.py to check this (getting errors such as `INTERNALERROR> ValueError: no option named '--cudasim'` when running `pytest`), but I'd suggest the following as a test for this:

```diff
From 2c9313f74360abc9f4af27b855d0cf7f357a55b1 Mon Sep 17 00:00:00 2001
From: Graham Markall <gmarkall@nvidia.com>
Date: Wed, 17 Mar 2021 11:10:44 +0000
Subject: [PATCH] Test debug kwarg to cuda.jit with simulator

---
 numba/cuda/tests/cudapy/test_cuda_jit_no_types.py | 9 +++++++++
 1 file changed, 9 insertions(+)

diff --git a/numba/cuda/tests/cudapy/test_cuda_jit_no_types.py b/numba/cuda/tests/cudapy/test_cuda_jit_no_types.py
index 49018b70e..b57e7649c 100644
--- a/numba/cuda/tests/cudapy/test_cuda_jit_no_types.py
+++ b/numba/cuda/tests/cudapy/test_cuda_jit_no_types.py
@@ -1,6 +1,7 @@
 from numba import cuda
 import numpy as np
 from numba.cuda.testing import CUDATestCase
+from numba.tests.support import override_config
 import unittest
 
 
@@ -76,6 +77,14 @@ class TestCudaJitNoTypes(CUDATestCase):
 
         self.assertEqual(b[0], (a[0] + 1) + (2 + 1))
 
+    def test_jit_debug_simulator(self):
+        # Ensure that the jit decorator accepts the debug kwarg when the
+        # simulator is in use - see Issue #6615.
+        with override_config('ENABLE_CUDASIM', 1):
+            @cuda.jit(debug=True)
+            def f(x):
+                pass
+
 
 if __name__ == '__main__':
     unittest.main()
-- 
2.28.0

```

This additional test fails on master with:

```
$ python -m numba.runtests numba.cuda.tests.cudapy.test_cuda_jit_no_types
...E
ERROR: test_jit_debug_simulator (numba.cuda.tests.cudapy.test_cuda_jit_no_types.TestCudaJitNoTypes)
  File "/home/gmarkall/numbadev/numba/numba/cuda/tests/cudapy/test_cuda_jit_no_types.py", line 83, in test_jit_debug_simulator
    def f(x):
  File "/home/gmarkall/numbadev/numba/numba/cuda/decorators.py", line 108, in autojitwrapper
    return FakeCUDAKernel(func, device=device,
TypeError: __init__() got an unexpected keyword argument 'debug'

Ran 4 tests in 0.292s

FAILED (errors=1)
```

but passes with your patch.

shaunc:
Thanks for the test! For the conftest to work, you also need `pytest_addoption` ... and probably that isn't what is wanted either (we want to test unconditionally -- not depending on an option, and the test should be isolated from other tests). 

I meant that more as an illustration of how I was managing to trigger this problem (though I'm not actually sure in detail what is happening). Still -- the PR does fix the problem for me, so was hopeful it could find its way into the repo.

shaunc:
\[Ok -- @gmarkall -- added your test to PR; Thanks!\]

gmarkall (dismissed):
This now looks good to me - many thanks for the fixes!

stuartarchibald:
@shaunc thanks for the patch, any chance you could fix this `flake8` error please so that CI succeeds? https://dev.azure.com/numba/numba/_build/results?buildId=8269&view=logs&j=9b1c419d-97bd-5aaf-47ca-b83be1ff9bc1&t=d1ed00e6-cc06-521a-e00c-4f5e4a0f55c5&l=84

Many thanks.


@gmarkall thanks for reviewing.

gmarkall (approved):
Thanks for the fix - this still looks good to me.

@stuartarchibald Do you want a buildfarm run?

stuartarchibald:
Think there's another `debug=` here https://github.com/numba/numba/blob/e0ffba72656d317d8ecd4396224d04c9088a39ae/numba/cuda/decorators.py#L76-L80 ?

shaunc:
Ok -- got that one too...

gmarkall (approved):
This still looks good to me with the additional change.

@stuartarchibald pointed out to me that there are reasons why the original issue may not manifest itself in all configurations, particularly if a driver is not installed:

> when you do `@cuda.jit` under CUDASIM and DO NOT have a driver, the `cuda.jit` decorator is the one from `numba.cuda.simulator.api`, and that has a `debug=`  flag in the `jit()` definition, whereas if you had the driver, you'd get `numba.cuda.jit` as your `jit()` definition which is broken, and the `numba.cuda.simulator.api.jit`  function does create a `FakeCUDAKernel` but it doesn't get passed the `debug` flag.

As this PR makes an improvement by fixing cases that presently don't work, I'm inclined not to suggest further changes now given that a whole set of changes to the `jit` decorator to make the logic more explicit and linear are planned for the near future (originally highlighted by the fact that `lineinfo` from PR #6802 "magically" worked, and reinforced by the discussion on this issue).

gmarkall:
@stuartarchibald Are you happy for this to go in, following a successful buildfarm run?

stuartarchibald:
> @stuartarchibald Are you happy for this to go in, following a successful buildfarm run?

Yes, I think this is a fix solely based on inspection of the source/issue reported, the `debug` flag is evidently erroneous in various cases. Also, as noted, the logic needs looking at more generally but that's not a task for this PR. Thanks for reviewing @gmarkall.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_38`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_38`.

Passed.

stuartarchibald:
@shaunc Congratulations on your first contribution to Numba!

PR #6836, opened 2021-03-17
CUDA: Documentation updates

gmarkall:
- PyCUDA now supports the CUDA Array Interface.
- Fix rendering of the documentation for `cuda.atomic.dec` - the indentation caused the description of its operation to render as a subtitle.


stuartarchibald (approved):
Thanks for the patch, manually checked doc rendering, looks fine.

PR #6840, opened 2021-03-18
CUDA: Remove items deprecated in 0.53 + simulator test fixes

gmarkall:
This PR removes items deprecated in 0.53 and scheduled for removal in 0.54. These include:

- The `compute_capabilty` kwarg to inspect methods.
- The ability to index dicts returns by inspect method using a tuple of compute capability and argument types.
- The return of a single item (instead of a dict) from the inspection methods and `ptx` property of specialized dispatchers.
- The `definition` and `definitions` property of dispatchers.

Whilst working on this I discovered a couple of other issues in the simulator:

- The testing of cooperative groups was accidentally skipped because `skip_if_cudadevrt_missing` always skipped tests on the simulator - this is now resolved.
- Some tests inspected some fake PTX from the simulator - this makes no sense, so the `ptx` property of fake kernels has been removed, and the tests which inspected it are now skipped on cudasim.

Some other remarks:

- Various tests made use of the deprecated functionality - these have now been updated to use the current interfaces.
- Relevant documentation (the Cooperative groups example and the deprecation notices) is updated.
- Tests of the deprecation notices have been removed accordingly.


numba/cuda/compiler.py
```diff
@@ -989,23 +977,6 @@ def specialized(self):
...
-            raise ValueError("Dispatcher needs to be specialized to get the "
-                             "single definition")
-        return next(iter(self.overloads.values()))
-
-    @property
-    def definitions(self):
-        warn('Use overloads instead of definitions',
-             category=NumbaDeprecationWarning)
-        return self.overloads
-
```
stuartarchibald, line 1008: These aren't mentioned in the deprecation schedule docs?

numba/cuda/tests/cudapy/test_constmem.py
```diff
@@ -217,46 +218,32 @@ def test_const_record_align(self):
...
             # On earlier versions, a vector of 4 bytes is used to load the
             # first three bytes.
             first_bytes = 'ld.const.v4.u8'
             first_bytes_msg = 'load the first three bytes as a vector'
 
-            self.assertIn(
-                first_bytes,
-                jcuconst.ptx,
-                first_bytes_msg)
+            self.assertIn( first_bytes, ptx, first_bytes_msg)
```
stuartarchibald, line 239:
```suggestion
            self.assertIn(first_bytes, ptx, first_bytes_msg)
```

numba/cuda/tests/cudapy/test_constmem.py
```diff
@@ -217,46 +218,32 @@ def test_const_record_align(self):
...
 
-        self.assertIn(
-            'ld.const.u32',
-            jcuconst.ptx,
-            'load the uint32 natively')
+        self.assertIn( 'ld.const.u32', ptx, 'load the uint32 natively')
```
stuartarchibald, line 241:
```suggestion
        self.assertIn('ld.const.u32', ptx, 'load the uint32 natively')
```

numba/cuda/tests/cudapy/test_constmem.py
```diff
@@ -217,46 +218,32 @@ def test_const_record_align(self):
...
 
         # On 10.2 and above, we already checked for loading the last byte by
         # itself - no need to repeat the check.
         if rtver < (10, 2):
-            self.assertIn(
-                'ld.const.u8',
-                jcuconst.ptx,
-                'load the last byte by itself')
+            self.assertIn( 'ld.const.u8', ptx, 'load the last byte by itself')
```
stuartarchibald, line 246:
```suggestion
            self.assertIn('ld.const.u8', ptx, 'load the last byte by itself')
```

numba/cuda/tests/cudapy/test_cooperative_groups.py
```diff
@@ -130,10 +131,10 @@ def test_max_cooperative_grid_blocks(self):
         # whilst keeping the total number of threads constant doesn't change
         # the maximum to validate some of the logic.
         c_sequential_rows = cuda.jit(void(int32[:,::1]))(sequential_rows)
-        defn = c_sequential_rows.definition
-        blocks1d = defn.max_cooperative_grid_blocks(256)
-        blocks2d = defn.max_cooperative_grid_blocks((16, 16))
-        blocks3d = defn.max_cooperative_grid_blocks((16, 4, 4))
+        overload = c_sequential_rows.overloads[(int32[:,::1],)]
```
stuartarchibald, line 134: factor out `sig = (int32[:, ::1],)` and reuse above and here?

numba/cuda/tests/cudapy/test_cooperative_groups.py
```diff
@@ -113,7 +113,8 @@ def test_sync_at_matrix_row(self):
 
         c_sequential_rows = cuda.jit(void(int32[:,::1]))(sequential_rows)
 
-        mb = c_sequential_rows.definition.max_cooperative_grid_blocks(blockdim)
+        overload = c_sequential_rows.overloads[(int32[:,::1],)]
```
stuartarchibald, line 116: factor out `sig = (int32[:, ::1],)` and reuse above and here?

numba/cuda/tests/cudapy/test_fastmath.py
```diff
@@ -150,11 +147,12 @@ def bar(arr, val):
             if i < arr.size:
                 arr[i] = foo(i, val)
 
-        fastver = cuda.jit("void(float32[:], float32)", fastmath=True)(bar)
-        precver = cuda.jit("void(float32[:], float32)")(bar)
+        sig = (float32[::1], float32)
```
stuartarchibald, line 150: I presume this change from an 'A' to a 'C' ordered `float32` array type is intentional?

stuartarchibald (commented):
Thanks for the patch, great to see this cleaned up. There's a few minor things to look at and once resolved this will be good to merge. Thanks again!

numba/cuda/compiler.py
gmarkall: It turns out that they're not, and that was an oversight - they'll have been warning in any code that uses them since 0.53 though.

numba/cuda/tests/cudapy/test_fastmath.py
gmarkall: Yes - generally I try to strive for the test suite containing good examples of the use of Numba, and in this case I believe it's always better to declare things as C-contiguous until you have a reason not to, so that the compiled code is more efficient than if the contiguity were not explicitly given.

gmarkall:
@stuartarchibald Many thanks for the review - comments should be addressed now.

gmarkall:
Is the fail a Frosty Thompson?

```
INFO: Running slice of discovered tests: (7,None,22)
cc1plus: warning: command line option '-Wstrict-prototypes' is valid for C/ObjC but not for C++
skipped HSA tests
skipped HSA tests
Parallel: 402. Serial: 62
s....................................................................................................................................x...............cc1plus: warning: command line option '-Wstrict-prototypes' is valid for C/ObjC but not for C++
..................................s................ssssssssssssssss..........F...............................................................................................................................................................................sss..ss..........sssssss.s..s....sss..........s............sss
FAIL: test_fork_from_non_main_thread (numba.tests.test_parallel_backend.TestTBBSpecificIssues)
  File "/home/vsts/work/1/s/numba/tests/test_parallel_backend.py", line 921, in test_fork_from_non_main_thread
    self.assertIn(msg_head, err)
AssertionError: 'Attempted to fork from a non-main thread, the TBB library' not found in ''

Ran 464 tests in 159.435s

FAILED (failures=1, skipped=39, expected failures=1)
```

stuartarchibald:
> Is the fail a Frosty Thompson?

No, that's the fork from non-main thread problem reappearing, would guess TBB 2021 related. Frosty Thompson appears as a live lock.

numba/cuda/tests/cudapy/test_cooperative_groups.py
stuartarchibald: Did you wish to do this, I can't seem to find it in the diff? I don't think it's necessarily essential.

stuartarchibald:
> > Is the fail a Frosty Thompson?
> 
> No, that's the fork from non-main thread problem reappearing, would guess TBB 2021 related. Frosty Thompson appears as a live lock.

Fixed in https://github.com/numba/numba/pull/6966

stuartarchibald:
Thanks for the fixes, please could you resolve the conflict against mainline, this can then have a build farm run?

numba/cuda/tests/cudapy/test_cooperative_groups.py
gmarkall: I've done this now, I missed it before.

gmarkall:
@stuartarchibald I've merged master and also done the little refactor I missed before - could this have a buildfarm run now please?

gmarkall:
(Re-running due to conda 500 / 530 errors in the prior test run)

numba/cuda/tests/cudapy/test_fastmath.py
stuartarchibald: Thanks for the explanation.

numba/cuda/tests/cudapy/test_cooperative_groups.py
stuartarchibald: Thanks

numba/cuda/compiler.py
stuartarchibald: Ah yes, they have been.

stuartarchibald (dismissed):

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_61`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_61`.

This failed. All fails are on windows.
* py3.8, np1.17, cuda 9.2
* py3.8, np1.17, cuda 10.0
* py3.9, np1.19, cuda 10.0
* py3.9, np1.19, cuda 9.2

Failure is the same on each:
```
<snip>
FAIL: test_const_record_optimization (numba.cuda.tests.cudapy.test_constmem.TestCudaConstantMemory)
-----------------------------------------------------------------------------
Trace back:
"numba\cuda\tests\test_constmem.py", line 200, in test_const_record_optimization
    self.assertGreaterEqual(u8_load_count, expected_load_count,
AssertionError: 11 not greater than or equal to 12: load record values as individual bytes
<snip>
```

gmarkall:
@stuartarchibald Thanks for the report. I've removed the failing test and a similar one, for the rationale outlined in the commit message - in summary I think these tests are a bit unhelpful / nonsensical and have adjusted them several times over the course of various PRs. Explanation from the commit:

```
These tests check for very specific patterns in the generated PTX that
are unstable across toolkit versions, platforms, and other minor
changes. Since these frequently need adjustment to match the generated
PTX, and have never revealed an underlying issue, is seems more sensible
to remove them instead of continually fixing them up in the presence of
variations.

Note that the latest issue was detected in the PR numba#6840 buildfarm run,
with failures in test_const_record_optimization on Windows for CUDA 9.2
and 10.0. These failures arose because a change to compile the kernel
with an explicit signature instead of specializing the kernel for given
arguments changed the types the kernel was being compiled for on Windows
- the default int on Windows is different to the default int on Linux.
So this test wasn't even testing the same thing on Windows as on Linux,
and this went unnoticed for a very long time!
```

stuartarchibald:
> @stuartarchibald Thanks for the report. I've removed the failing test and a similar one, for the rationale outlined in the commit message - in summary I think these tests are a bit unhelpful / nonsensical and have adjusted them several times over the course of various PRs. Explanation from the commit:
> 
> ```
> These tests check for very specific patterns in the generated PTX that
> are unstable across toolkit versions, platforms, and other minor
> changes. Since these frequently need adjustment to match the generated
> PTX, and have never revealed an underlying issue, is seems more sensible
> to remove them instead of continually fixing them up in the presence of
> variations.
> 
> Note that the latest issue was detected in the PR numba#6840 buildfarm run,
> with failures in test_const_record_optimization on Windows for CUDA 9.2
> and 10.0. These failures arose because a change to compile the kernel
> with an explicit signature instead of specializing the kernel for given
> arguments changed the types the kernel was being compiled for on Windows
> - the default int on Windows is different to the default int on Linux.
> So this test wasn't even testing the same thing on Windows as on Linux,
> and this went unnoticed for a very long time!
> ```

Thanks for the explanation @gmarkall. I agree, if these tests are not really adding anything but very susceptible to external/environmental changes then removing them seems to be the best solution.

stuartarchibald (approved):
Thanks for the fix.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_63`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_63`.

This failed due to the issue in https://github.com/numba/numba/pull/7052. Once that's merged will re-run the build.

gmarkall:
@stuartarchibald / @esc Could this have another run now please?

stuartarchibald:
> @stuartarchibald / @esc Could this have another run now please?

Buildfarm ID: `numba_smoketest_cuda_yaml_67`.

stuartarchibald:
> > @stuartarchibald / @esc Could this have another run now please?
> 
> Buildfarm ID: `numba_smoketest_cuda_yaml_67`.

Passed.

Issue #5311, opened 2020-02-25
[cuda] lineinfo for device function


sklam:
Related gitter discussion starting at [here](https://gitter.im/numba/numba?at=5e4ff1bd8b2d4664ef17be48) and [here](https://gitter.im/numba/numba?at=5e53d9291be0ff01d5a69887).

Reproducer:

```python
# filename: debugme.py
import math
import numpy as np
import numba
from numba import cuda


@cuda.jit(device=True, debug=True)
def expensive(x):
    return math.pow(x, 1.001)


@cuda.jit(device=True, debug=True)
def cheap(x):
    return math.sqrt(x)


@cuda.jit(device=True, debug=True)
def calc(x):
    s1 = expensive(x)
    q = cheap(s1)
    return q + 1


@cuda.jit(device=True, debug=True)
def threadid():
    return cuda.blockDim.x * cuda.blockIdx.x + cuda.threadIdx.x


@cuda.jit(argtypes=[numba.int32[:]], debug=True)
def kernel(arr):
    for i in range(1000):
        where = threadid()
        value = arr[where]
        arr[where] = calc(value)


def main():
    print("SUP")
    z = np.array(range(0, 5000))
    print(z)
    kernel[5000 // 128 + 1, 128](z)
    print(z)

    print(kernel.inspect_llvm())

if __name__ == '__main__':
    main()
```

Run the above with:

```bash
$ nvprof -a instruction_execution --export-profile profile_output python debugme.py
```

On my OSX 10.13 with CUDA 9.2, we can see some line-info in the `profile_output` using NVVP:

<img width="521" alt="separate_devfns" src="https://user-images.githubusercontent.com/1929845/75255060-faac3980-57a6-11ea-8b22-448fea36f026.png">

The "inline-closure" version (this style: https://gist.github.com/c200chromebook/d131322c918b142f14f40c9b3583cc6d) looks like:
<img width="532" alt="inlined_closure" src="https://user-images.githubusercontent.com/1929845/75255017-e7996980-57a6-11ea-8873-7a0f312c860c.png">

@c200chromebook, can you tell us what you see?

c200chromebook:
CUDA 10.1, numba 0.48.0, numpy 1.18.1, llvmlite 0.31.0 - you don't even get near the profiling.

![image](https://user-images.githubusercontent.com/9907030/75358072-4da5ef80-5880-11ea-8e52-db03e415cc86.png)

c200chromebook:
[...]

PR #6841, opened 2021-03-18
CUDA: Fix source location on kernel entry and enable breakpoints to be set on kernels by mangled name

gmarkall:
This PR improves a couple of aspects of debugging in CUDA:

- Breaking on any kernel launch, for example with `set cuda break_on_launch application` now results in GDB being able to find the source line at the break location - this enables the backtrace to show the mangled function name, source file, and line number. It also enables the text user interface to display the source when the kernel launches. From this point, stepping through
  line-by-line is feasible with the `next` command.
- Breakpoints can be set on a specific function and argument types, by setting a breakpoint on the mangled name.

(neither of these worked prior to this change).

This PR also separates the DIBuilder API from the Numba IR API in the first commit, to avoid building Numba IR for no reason (or strange reasons) when generating the kernel wrapper. 

There's still a little API weirdness that bothers me, but I'm not sure I can see the better way - perhaps the reviewer can suggest an improvement (I'll comment on the diff).


numba/cuda/compiler.py
```diff
@@ -479,8 +481,11 @@ def __init__(self, py_func, argtypes, link=None, debug=False, inline=False,
         }
 
         tgt_ctx = cres.target_context
+        filename = cres.type_annotation.filename
```
gmarkall, line 484: It's weird that I have to get the source file name from the type annotation at this stage, but I think the function IR is now lost (it was never saved after the pipeline finished).

numba/cuda/compiler.py
```diff
@@ -479,8 +481,11 @@ def __init__(self, py_func, argtypes, link=None, debug=False, inline=False,
...
+        linenum = int(cres.type_annotation.linenum)
```
gmarkall, line 485: Also, the line number being a string in the type annotation is a minor problem, as LLVM expects line numbers to be integers (e.g. `5`), not strings (`"5"`).

gmarkall:
Example of the text user interface showing source layout and backtrace (start of the kernel is repro.py line 7, the wrapper function name begins `_ZN6cudapy8`):

image

gmarkall:
A couple more notes on the lack of un-mangling in cuda-gdb:

For the CUDA target, `DICompileUnit` metadata looks like:

```
!2 = distinct !DICompileUnit(emissionKind: 1, file: !1, isOptimized: true, language: DW_LANG_Python, producer: "Numba", runtimeVersion: 0)
```

whereas on the CPU target the `emissionKind` is `FullDebug`, i.e. like:

```
!2 = distinct !DICompileUnit(emissionKind: FullDebug, file: !1, isOptimized: true, language: DW_LANG_Python, producer: "Numba", runtimeVersion: 0)
```

At the moment the name and the linkage name are the same for each subprogram, e.g.:

```
!5 = distinct !DISubprogram(file: !1, isDefinition: true, isLocal: false, isOptimized: true, line: 29, linkageName: "_ZN6cudapy8__main__10kernel$241E5ArrayIiLi1E1A7mutable7alignedE", name: "_ZN6cudapy8__main__10kernel$241E5ArrayIiLi1E1A7mutable7alignedE", scope: !1, scopeLine: 29, type: !4, unit: !2)
```

instead of

```
!5 = distinct !DISubprogram(file: !1, isDefinition: true, isLocal: false, isOptimized: true, line: 29, linkageName: "_ZN6cudapy8__main__10kernel$241E5ArrayIiLi1E1A7mutable7alignedE", name: "kernel", scope: !1, scopeLine: 29, type: !4, unit: !2)
```

However, for device functions the `name` is correct, and these are also not un-mangled.

Fixing these alone doesn't seem to help (I've tried locally by applying):

```diff
diff --git a/numba/cuda/cudadrv/nvvm.py b/numba/cuda/cudadrv/nvvm.py
index c77ec456e..35ab0a260 100644
--- a/numba/cuda/cudadrv/nvvm.py
+++ b/numba/cuda/cudadrv/nvvm.py
@@ -776,6 +777,9 @@ def llvm100_to_70_ir(ir):
             attrs = ' '.join(a for a in attrs if a != 'willreturn')
             line = line.replace(m.group(1), attrs)
 
+        if '= distinct !DICompileUnit' in line:
+            line = line.replace('emissionKind: 1', 'emissionKind: FullDebug')
+
         buf.append(line)
 
     return '\n'.join(buf)
diff --git a/numba/cuda/target.py b/numba/cuda/target.py
index 090c58d34..82c21c794 100644
--- a/numba/cuda/target.py
+++ b/numba/cuda/target.py
@@ -145,12 +145,13 @@ class CUDATargetContext(BaseContext):
                                                 nvvm_options=nvvm_options,
                                                 max_registers=max_registers)
         library.add_linking_library(codelib)
-        wrapper = self.generate_kernel_wrapper(library, kernel_name, func_name,
+        wrapper = self.generate_kernel_wrapper(library, codelib.name,
+                                               kernel_name, func_name,
                                                argtypes, debug, filename,
                                                linenum)
         return library, wrapper
 
-    def generate_kernel_wrapper(self, library, kernel_name, func_name,
+    def generate_kernel_wrapper(self, library, py_name, kernel_name, func_name,
                                 argtypes, debug, filename, linenum):
         """
         Generate the kernel wrapper in the given ``library``.
@@ -172,7 +173,10 @@ class CUDATargetContext(BaseContext):
 
         if debug:
             debuginfo = self.DIBuilder(module=wrapper_module, filepath=filename)
-            debuginfo.mark_subprogram(wrapfn, kernel_name, linenum)
+            debuginfo.mark_subprogram(wrapfn, py_name, linenum)
             debuginfo.mark_location(builder, linenum)
 
         # Define error handling variables
```

Since it doesn't help I've not added it to this PR - I'd prefer only to add things that clearly make improvements - are there any other thoughts on what might be missing for mangling to work here?

gmarkall:
(Popping this back to "In progress" because I just noticed it breaks with NVVM 3.4, and needs #5311 fixing for it to work)

gmarkall:
I've included a fix for #5311 in this branch now - this resolves the long-standing "Error: Debugging support cannot be enabled when number of debug compile units is more than 1" message.

gmarkall:
I think this is ready for review - looks like I somehow forgot to mark it as such.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_70`. Just to see if there's any immediate issues with the patch.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_70`. Just to see if there's any immediate issues with the patch.

Passed.

stuartarchibald:
@gmarkall please could you resolve the conflicts? then I'll look to review this, thanks.

gmarkall:
@stuartarchibald conflicts now resolved.

numba/cuda/compiler.py
stuartarchibald: It's probably in the `self.py_func.__code__.co_filename` and `cres.fndesc.lookup_module.__file__` too if they are preferable? IIRC the compile result doesn't hold reference to the IR as the IR could hold reference to e.g. globals, some of which might be large (e.g. NumPy arrays) and this would create memory pressure etc.

numba/cuda/compiler.py
stuartarchibald: `foo.py_func.__code__.co_firstlineno` might help, as might `inspect` module?

numba/cuda/target.py
```diff
@@ -143,6 +144,8 @@ def prepare_cuda_kernel(self, codelib, func_name, argtypes, debug,
         argtypes:      An iterable of the types of arguments to the kernel.
         debug:         Whether to compile with debug.
         nvvm_options:  Dict of NVVM options used when compiling the new library.
+        filename:      The source filename that the function is contained in.
+        linenum:       The source line that the function is on.
```
stuartarchibald, line 148: What happens if someone materialized a function from a string?

numba/cuda/tests/cudapy/test_debuginfo.py
```diff
@@ -61,6 +62,73 @@ def test_issue_5835(self):
...
+            self.assertEqual(len(disubprograms), 1)
+
+            wrapper_disubprogram = disubprograms[0]
+            # Check that the subprogram points to a wrapper (these are all in
+            # the "cudapy::" namespace).
+            self.assertIn('_ZN6cudapy', wrapper_disubprogram)
+
+    def test_debug_function_calls_internal_impl(self):
+        # Calling a function in a module generated from an implementation
+        # internal to Numba reqires multiple modules to be compiled with NVVM -
```
stuartarchibald, line 101:
```suggestion
        # internal to Numba requires multiple modules to be compiled with NVVM -
```

stuartarchibald (commented):
Thanks for the patch, I've taken a first look through it but would like to do some manual testing. There's a few minor things to look at and some questions/suggestions above. Thanks again!

numba/cuda/target.py
gmarkall: `filename` is `<string>` and `linenum` is the line in the string on which the function began - I think this is normal / expected?

gmarkall:
@stuartarchibald Thanks for the review and suggestions! All comments should be addressed now.

esc:
Build Farm ID: `numba_smoketest_cuda_yaml_76`

stuartarchibald (approved):

stuartarchibald:
Note: I manually verified this patch:
```bash
$ cuda-gdb --args python add_one.py
<snip>
Reading symbols from python...
(cuda-gdb) set cuda break_on_launch application
(cuda-gdb) r
Starting program: bin/python add_one.py
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
[Detaching after fork from child process 7719]
[Detaching after fork from child process 7720]
[Detaching after fork from child process 7721]
[Detaching after fork from child process 7724]
[New Thread 0x7fffda69b700 (LWP 7731)]
[New Thread 0x7fffd9e9a700 (LWP 7732)]
[New Thread 0x7fffd9699700 (LWP 7733)]
numba/cuda/compiler.py:865: NumbaPerformanceWarning: Grid size (1) < 2 * SM count (26) will likely result in GPU under utilization due to low occupancy.
  warn(NumbaPerformanceWarning(msg))
[Switching focus to CUDA kernel 0, grid 1, block (0,0,0), thread (0,0,0), device 0, sm 0, warp 0, lane 0]
0x0000555556b98c88 in _ZN6cudapy8__main__20increment_by_one$241E5ArrayIdLi1E1C7mutable7alignedE<<<(1,1,1),(1,1,1)>>> ()
    at add_one.py:4
4       @cuda.jit(debug=True, opt=0)
(cuda-gdb) l
1       import numpy as np
2       from numba import cuda
3
4       @cuda.jit(debug=True, opt=0)
5       def increment_by_one(an_array):
6           # Thread id in a 1D block
7           tx = cuda.threadIdx.x
8           # Block id in a 1D grid
9           ty = cuda.blockIdx.x
10          # Block width, i.e. number of threads per block
(cuda-gdb) s
0x0000555556b7f168 in _ZN8__main__20increment_by_one$241E5ArrayIdLi1E1C7mutable7alignedE () at add_one.py:4
4       @cuda.jit(debug=True, opt=0)
(cuda-gdb) s
7           tx = cuda.threadIdx.x
(cuda-gdb) s
13          pos = tx + ty * bw
(cuda-gdb) s
14          if pos < an_array.size:  # Check array boundaries
```

PR #6844, opened 2021-03-19
CUDA: Remove NUMBAPRO env var warnings, envvars.py + other small tidy-ups

gmarkall:
The `NUMBAPRO_*` env vars have been deprecated and ignored since 0.45 - this PR removes the warning about them. Now the warning is gone, the `get_numba_envvar` function serves no purpose above normal environment variables, so the variables that this was used for are converted to normal config variables, and documented.

I changed the `VERBOSE_CU_JIT_LOG` variable name to `CUDA_VERBOSE_JIT_LOG` for consistency - I don't believe this needs a deprecation cycle as it's never been public (and is likely unused in general).

There are a couple of other small tidy-ups bundled in this PR related to the NumbaPro name, a stale link, and a correction to the check for minumum supported compute capability.


docs/source/reference/envvars.rst
```diff
@@ -430,6 +430,27 @@ GPU support
    Setting this variable to 1 also includes the values of arguments to Driver
    API calls in the logs.
 
+.. envvar:: NUMBA_CUDA_DRIVER
+
+   Path of the directory in which the CUDA driver libraries are to be found.
+   Normally this should not need to be set as Numba can locate the driver in
+   standard locations. However, ths variable can be used if the driver is in a
```
stuartarchibald, line 437:
```suggestion
   standard locations. However, this variable can be used if the driver is in a
```

docs/source/reference/envvars.rst
```diff
@@ -430,6 +430,27 @@ GPU support
...
+   non-standard location.
+
+.. envvar:: NUMBA_CUDA_LOG_SIZE
+
+   Buffer size for logs produced by driver API operations. This defaults to 1024
```
stuartarchibald, line 442:
```suggestion
   Buffer size for logs produced by CUDA driver API operations. This defaults to 1024
```

docs/source/reference/envvars.rst
```diff
@@ -430,6 +430,27 @@ GPU support
...
+   and should not normally need to be modified - however, if an error in an API
+   call produces a large amount of output that appears to be truncated (perhaps
+   due to multiple long function names, for example) then this variable can be
+   used to increase the buffer size and view the full error message.
+
+.. envvar:: NUMBA_CUDA_VERBOSE_JIT_LOG
+
+   Whether the driver should produce verbose log messages. Defaults to 1,
```
stuartarchibald, line 450:
```suggestion
   Whether the CUDA driver should produce verbose log messages. Defaults to 1,
```

numba/cuda/cudadrv/driver.py
```diff
@@ -35,11 +35,9 @@
 from .drvapi import API_PROTOTYPES
 from .drvapi import cu_occupancy_b2d_size, cu_stream_callback_pyobj, cu_uuid
 from numba.cuda.cudadrv import enums, drvapi, _extras
-from numba.cuda.envvars import get_numba_envvar
 
 
-VERBOSE_JIT_LOG = int(get_numba_envvar('VERBOSE_CU_JIT_LOG', 1))
-MIN_REQUIRED_CC = (2, 0)
+MIN_REQUIRED_CC = (3, 0)
```
stuartarchibald, line 40: Whilst unlikely, is this going to break anything?

numba/core/config.py
```diff
@@ -337,6 +337,16 @@ def avx_default():
...
+        # Path of the directory that the CUDA driver libraries are located
+        CUDA_DRIVER = _readenv("NUMBA_CUDA_DRIVER", str, '')
+
+        # Buffer size for logs produced by CUDA driver operations (e.g.
+        # linking)
+        CUDA_LOG_SIZE = _readenv("NUMBA_CUDA_LOG_SIZE", int, 1024)
+
+        # Whether to generate verbose log messages when JIT linking
+        CUDA_VERBOSE_JIT_LOG = _readenv("NUMBA_CUDA_VERBOSE_JIT_LOG", int, 1)
+
```
stuartarchibald, line 349: With moving to `numba.config` I suppose it's possible for users to end up with broken code if they were manipulating the env vars that used to provide this functionality from within python itself. Once `numba.config` is in use, as soon as `numba` is imported I think these configurations are strongly bound to the environment and subsequent manipulation with e.g. `os.environ` will have no effect. The impact is probably minimal and pretty easy to resolve, but just something of which to be aware.

numba/core/config.py
```diff
@@ -337,6 +337,16 @@ def avx_default():
         CUDA_ARRAY_INTERFACE_SYNC = _readenv("NUMBA_CUDA_ARRAY_INTERFACE_SYNC",
                                              int, 1)
 
+        # Path of the directory that the CUDA driver libraries are located
+        CUDA_DRIVER = _readenv("NUMBA_CUDA_DRIVER", str, '')
```
stuartarchibald, line 341: Does this have the effect, when combined with https://github.com/numba/numba/pull/6844/files#diff-7685977fe7c53dfd319942fe1b5405bcf1d7e3488e00568b02597efcfc65e92cR115-R124 of making pwd the `envpath`, e.g. `envpath = os.path.abspath(envpath)` == `envpath = os.path.abspath('')` == `$pwd`.

stuartarchibald (commented):
Thanks for the patch, few minor things to look at else looks good. It's good to see legacy references finally removed!

numba/cuda/cudadrv/driver.py
gmarkall: I don't think so - Nothing pre-3.0 is supported by any CUDA supported toolkit version anyway.

numba/core/config.py
gmarkall: I don't think so, I think the change you linked to should prevent that, because `if envpath` is not true when `config.CUDA_DRIVER` is `''`. If someone sets it to `'.'` then they could intentionally make the current working directory the env path. (If this doesn't answer your question, could you re-phrase it? I found it a bit hard to be sure what you were asking about)

numba/core/config.py
gmarkall: Sometimes these variables are read at import time, and possibly sometimes when CUDA is first used - after that I think manipulating them would have no effect anyway... There's a possibility that code manipulating them could get broken by this change, but that code was manipulating undocumented variables that weren't really publicly exposed before now.

gmarkall:
@stuartarchibald Many thanks for the review - documentation suggestions applied, and other questions responded to.

numba/core/config.py
stuartarchibald: I just interpreted the code entirely incorrectly, ignore this :)

numba/core/config.py
stuartarchibald: Yes, I agree, so long as we're aware of it it's fine.

numba/cuda/cudadrv/driver.py
stuartarchibald: Thanks for confirming.

stuartarchibald (dismissed):

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_33`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_33`.

Passed.

sklam:
@gmarkall , there's a conflict

gmarkall:
Conflicts now resolved.

stuartarchibald:
> Conflicts now resolved.

Thanks @gmarkall, looks good.

stuartarchibald (approved):

Issue #5911, opened 2020-06-23
Feature Request: __hash__ for Enum and IntEnum


HPLegion:
Currently `__hash__` is not implemented for `Enum` and `IntEnum`. Because of this they cannot be used as keys for `typed.Dict` which is a bit unfortunate.

The CPython implementation for hashing Enums is surprising simple, it is simply `hash(name)`

https://github.com/python/cpython/blob/b4e85cadfbc2b1b24ec5f3159e351dbacedaa5e0/Lib/enum.py#L641-L642

But unfortunately within numba Enums are currently lowered as their Value field only, so that the name field is not trivially accessible. Therefore implementing the `__hash__` function may get tricky.

https://github.com/numba/numba/blob/ae598dbf252bad3917b76d9e9844e6ab59bcd218/numba/cpython/enumimpl.py#L55-L60
------------------------

In case someone wants to look into this a few hints (or things that surprised me):
```python
In [1]: from enum import Enum 
   ...:  
   ...: class Color(Enum): 
   ...:     RED = 1 
   ...:     BLUE = 2 
   ...:     GREEN=3 
   ...:     MAGENTA=1 
   ...:                                                                                                              

In [2]: type(Color)                                                                                                  
Out[2]: enum.EnumMeta 
# Enums are of type EnumMeta (but the type is called Enum in numba)

In [3]: type(Color.RED)                                                                                              
Out[3]: <enum 'Color'> 
# Enum members are of type Enum (but the type is called EnumMember in numba)

In [4]: Color.RED.name                                                                                               
Out[4]: 'RED'

In [5]: Color.BLUE.name                                                                                              
Out[5]: 'BLUE'

In [6]: Color.MAGENTA.name                                                                                           
Out[6]: 'RED'
# Aliases (members with same value) seem to have their name set to the first member 
# with this value, this is why all aliases have the same hash

```

esc:
@HPLegion thanks for submitting this, I have labeled it as a feature request.

HPLegion:
Cheers!

In the meantime in case anyone else runs into a similar problem:
I managed to use `IntEnums` as `typed.Dict` keys by overloading `__hash__` based on the value. For dictionary purposes this seems to be good enough, but the hashes will be different from those in CPython - so they should not be used for other purposes.

```python
@numba.extending.overload_method(numba.types.EnumMember, '__hash__')
def enum_hash(val):
    def impl(val):
        return hash(val.value)
    return impl
```

PR #6847, opened 2021-03-22
[WIP][QUESTIONS] An attempt to extend Enum support

HPLegion:
Hi everyone :-)

*While this was once meant to be a real PR, I see it a bit more as a learning opportunity now. So there is absolutely no urgency here. I opened this primarily for archival purposes and to maybe learn a thing or two.*

A while back I had a go at improving the Enum support, triggered by [#5911](https://github.com/numba/numba/issues/5911), but eventually I got stuck.
After reading [this post](https://numba.discourse.group/t/intenum-from-int/625/5) on Discourse I thought it would be a good idea to share what I have come up with so far (this PR would not fix the reported issue there, but maybe it could help someone to pick up the task of implementing a solution.)

## Context

Currently numba deals with Enum fields based on their `value` alone. This cannot fully reflect CPython behaviour, where the member `name` is used for hashing.
Since I needed the hashing capabilities and thought that the name field would be needed to achieve this, I started to turn EnumMembers into Structures with a field each for their name and value.

When I started this work I was unaware that IntEnums (which are sufficient for myself) behave distinctly different since their members are first and foremost integers. As such they use the `int.__hash__` function on their `value`, such that my original [hack](https://github.com/numba/numba/issues/5911#issuecomment-649457131) is actually perfectly acceptable, and extending Enums in general would not be necessary to make this work.

I doubt that adding hashing capabilities for generic Enums justifies complicating the very simple existing implementation to this extent, therefore I will open a proper PR for the IntEnum hashing.

This is maybe more of a resource for the future if extending the capabilities of Enums becomes necessary.

## What this PR tries to achieve

- Extend (Int)Enums so that their value and name are available in jitted code.
- Allow explicit casting of the newly extended IntEnums
- Add hashing of Enums (name based) and IntEnums (value based)
- Block array creation from generic Enums without explicit casting (see below)
- Don't break existing functionality :laughing: 

## Questions / Known issues / Plea for comments

- This is the first time I tried to dig a little deeper into Numba, and I am sure there are some heinous crimes (reference counting :see_no_evil:) hidden in my code.
I would appreciate if someone could share some criticism so that I (and hopefully others) can learn something for the future :slightly_smiling_face: 

- If one were to adopt these changes, should the Enum type move into a different `core/types` submodule since it is no longer strictly scalar?

- What would be needed to implement arithmetic operations and comparisons (maybe general ufuncs) between arrays and IntEnums? The associated tests fail. Operations between IntEnums and scalars from inside arrays seem to work (probably thanks to implicit casting)

- Returning the IntEnums from a vectorized function (without implicit or explicit casting) fails with a segfault (I simplified the associated test, since it contained a comparison that I know to be wonky at this point.)

## Observed existing issues

While working on this patch I noticed that numba currently allows creating arrays of generic Enum members. This is possible because they are lowered as scalar values. So if the datatype is supported by numpy it will happily create an array.

In CPython this actually does not work quite the same; since the enum members are generic Python objects, numpy creates an object array, which for example does not allow arithmetic operations, even if all the Enum values are themselves numbers (so you could argue that numba has a _better_ support :stuck_out_tongue: ).

The culprit is located [here](https://github.com/numba/numba/blob/bb2ec93fe107339d752bb03bde49f505fa2ed02c/numba/np/numpy_support.py#L145-L146).

Does this warrant a separate issue or would you like to keep this as it is currently working? (This PR assumes it shouldn't be possible, but I am aware that breaking things is always risky business.)

```python
from enum import Enum
from numba import njit
import numpy as np

class E(Enum):
    MEMBER = 1

def f(x):
    return np.array((x, x))
fjit = njit(f)

print(f(E.MEMBER).dtype)  # --> object
print(fjit(E.MEMBER).dtype)  # --> int64
```

Cheers
Hannes

/cc @nelson2005

### DIFF
--- a/numba/core/boxing.py
+++ b/numba/core/boxing.py
@@ -162,10 +162,10 @@ def box_enum(typ, val, c):
     """
     Fetch an enum member given its native value.
     """
-    valobj = c.box(typ.dtype, val)
-    # Call the enum class with the value object
+    member = c.context.make_helper(c.builder, typ, val)
+    valueobj = c.box(typ.dtype, member.value)
     cls_obj = c.pyapi.unserialize(c.pyapi.serialize_object(typ.instance_class))
-    return c.pyapi.call_function_objargs(cls_obj, (valobj,))
+    return c.pyapi.call_function_objargs(cls_obj, (valueobj,))
 
 
 @unbox(types.EnumMember)
@@ -173,8 +173,15 @@ def unbox_enum(typ, obj, c):
     """
     Convert an enum member's value to its native value.
     """
+    nameobj = c.pyapi.object_getattr_string(obj, "name")
     valobj = c.pyapi.object_getattr_string(obj, "value")
-    return c.unbox(typ.dtype, valobj)
+    member = c.context.make_helper(c.builder, typ)
+    member.name = c.unbox(typ.ntype, nameobj).value
+    member.value = c.unbox(typ.dtype, valobj).value
+    c.pyapi.decref(nameobj)
+    c.pyapi.decref(valobj)
+    return NativeValue(member._getvalue(), is_error=c.pyapi.c_api_error())
+
 
 #
 # Composite types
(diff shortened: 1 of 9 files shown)


HPLegion:
Feel free to close if this adds noise to the list of open PRs :-)

PR #6849, opened 2021-03-22
Add __hash__ for IntEnum

HPLegion:
Hi  :-)

Partially solving #5911

This patch adds `__hash__` capability for IntEnums. I think this would be a nice addition, because I like to use enums as keys for the typed `Dict`, since there is less risk of introducing typos. 

I have been using this inside my own code for a while, but I had mistakenly thought that the hashes I created are different from the CPython hashes.
Today I realised that the CPython hash for an IntEnum is in fact simply `int.__hash__ (value)` so I think this could move into numba.

Is it frowned upon to use the extension API for internals? 

Cheers!

### DIFF
--- a/numba/cpython/enumimpl.py
+++ b/numba/cpython/enumimpl.py
@@ -80,6 +80,7 @@ def enum_class_getitem(context, builder, sig, args):
     return context.get_constant_generic(builder, enum_cls_typ.dtype,
                                         member.value)
 
+
 @overload_method(types.IntEnumMember, '__hash__')
 def intenum_hash(val):
     # uses the hash of the value, for IntEnums this will be int.__hash__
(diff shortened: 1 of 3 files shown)


stuartarchibald:
Thanks for the patch @HPLegion 

> Is it frowned upon to use the extension API for internals?

It's encouraged!

numba/cpython/enumimpl.py
```diff
@@ -7,6 +7,7 @@
                                  lower_getattr_generic, lower_cast,
                                  lower_constant, impl_ret_untracked)
 from numba.core import types
+from numba.extending import overload_method
```
stuartarchibald, line 10:
```suggestion
from numba.core.extending import overload_method
```

stuartarchibald (commented):

numba/cpython/enumimpl.py
HPLegion: Thanks for the suggestions. I pushed an update :-)

stuartarchibald (approved):

PR #6852, opened 2021-03-23
nanvar with ddof-v1

rawwar:
Added support for `ddof` argument for np.nanvar . Moved `compute_sum_of_square_diffs` out of `nanvar_impl` and decorated with @register_jittable


rawwar:
This is a work in progress. While I am looking for feedback, I am also trying to fix the errors occurred in the builds.

numba/np/arraymath.py
```diff
@@ -155,7 +155,7 @@ def create_full_slice():
     return function_sig, codegen
 
 
-#----------------------------------------------------------------------------
+# ----------------------------------------------------------------------------
```
stuartarchibald, line 158: It's good to try and avoid cosmetic changes in PRs that are about functionality, this makes it easier for reviewers as the change set is smaller, it also somewhat prevents PR `X` making a change one way and PR `Y` making it another.

numba/np/arraymath.py
```diff
@@ -918,29 +918,57 @@ def nanmean_impl(a):
...
+        if not isnan(v):
+            val = (v.item() - m)
+            ssd += np.real(val * np.conj(val))
+            count += 1
+    return ssd, count
+
+
 @overload(np.nanvar)
-def np_nanvar(a):
+def np_nanvar(a, ddof):
```
stuartarchibald, line 936: `ddof` is a kwarg with default value `0`. https://numpy.org/doc/stable/reference/generated/numpy.nanvar.html#numpy.nanvar. The Numba documentation "A guide to using @overload" is probably worth a read as it covers the sort of things that occur in this PR https://numba.readthedocs.io/en/stable/extending/overloading-guide.html

stuartarchibald (commented):
Thanks for the PR. I've provided a bit of feedback/guidance that should hopefully get this to the point of existing tests passing. It'd probably be worth adding tests for the new functionality soon so that they can help guide development.

numba/np/arraymath.py
rawwar: Thank you.  I won't repeat this

rawwar:
Just realized I am causing unnecessary CI builds. Will be avoiding that by using locally running tests

numba/np/arraymath.py
stuartarchibald: Thanks, and no worries, it's just a convention Numba has to try and reduce review time. If you want to do cosmetic changes then feel free to open PRs for just those changes.

rawwar:
@stuartarchibald , I updated the implementation. I also tried to write test cases. But, I keep facing issues. Wondering how can i update test case for nanvar and nansum as the existing test cases don't take any extra argument like ddof. This makes me think if i should move test cases of nanvar and nansum from  reduction_funcs and write separate tests with ddof argument.

esc:
this one was replaced, right?

rawwar:
Yes, @esc

Issue #6860, opened 2021-03-24
Doc error: old reference to hsa


sklam:
The 
```
from numba import hsa
```
should be
```
from numba import roc
```
on this page: https://numba.pydata.org/numba-doc/latest/roc/memory.html

rawwar:
This Issue can be closed?

stuartarchibald:
@rawwar yes, thanks, will close it now. For future you can also close issues via direct linking from a pull request by using predefined "keywords" from [here](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword). Example: https://github.com/numba/numba/pull/6742 has `Fixes: #6737` which means when it gets merged `#6737` gets closed.

PR #6861, opened 2021-03-24
updated reference to hsa with roc

rawwar:
PR for issue #6860. replaced hsa with roc


sklam (approved):
Thanks for the quick patch!

Issue #4064, opened 2019-05-07
Annotate HTML alignment broken in JupyterLab


cdeil:
While trying out Numba `Annotate` (and after fixing #4063) I noticed that the alignment is broken in JupyterLab. It is right-aligned, but should be left-aligned. (I didn't try in Jupyter notebook)

See screenshot below.

@Carreau - Do you know what's wrong?

---

Another usability issue is that executing the cell with the example multiple times makes the HTML output grow and grow, showing the same annotated output multiple times. This looks like a bug, no? When re-executing a `def` statement with `numba.jit` decorator, a new `numba.targets.registry.CPUDispatcher` object should be created, and nothing should accumulate, no?

---

<img width="851" alt="Screenshot 2019-05-07 at 10 22 12" src="https://user-images.githubusercontent.com/852409/57285126-d0bc4600-70b2-11e9-8e42-1ac8fbdde4fb.png">

Carreau:
That is likely an issue in JupyterLab itself; and diverging CSS with classic notebook; do you know with which version of JupyterLab. cc @jasongrout

cdeil:
I have JupyterLab version 0.35.5 and Numba git master (but probably the same as the last release, you should see this issue with Numba 0.43.1 as well).

Carreau:
you may want to try with Jupyterlab 1.0-alpha, using `pip install --pre`, I know the CSS was updated; but havent' been involved much.

cdeil:
I tried `pip install -U --pre jupyterlab` and got `jupyterlab-1.0.0a3-py2.py3-none-any.whl`. Still the same issue.

In Jupyter notebook the output looks fine.

If someone has time to try and reproduce and pinpoint the issue, here's the code snippet to copy & paste to reproduce the issue:
```
import numba
from numba.pretty_annotate import Annotate
@numba.jit
def test(q):
    res = 0
    for i in range(q):
        res += i
    return res
     
test(10)
Annotate(test)
```

jasongrout:
Here is the classic notebook css:
<img width="616" alt="Screen Shot 2019-05-07 at 11 48 32 AM" src="https://user-images.githubusercontent.com/192614/57325668-12241400-70fa-11e9-9999-7338be8dd2b4.png">

Here it is in JLab. You'll notice that there is some jlab css that is overriding the align left. The joys of the global css namespace!
[...]

PR #6871, opened 2021-03-26
Force text-align:left when using Annotate

guilhermeleobas:
This is a simple fix for https://github.com/numba/numba/issues/4064

Before:
image

Result after fix:
image


stuartarchibald (approved):
Thanks for the patch. I verified the change manually in `jupyter-lab`.

PR #6873, opened 2021-03-28
docs: Update reference to @jitclass location

dnadlinger:

### DIFF
--- a/docs/source/user/jitclass.rst
+++ b/docs/source/user/jitclass.rst
@@ -10,13 +10,14 @@ Compiling Python classes with ``@jitclass``
   exposed or implemented, yet.
 
 
-Numba supports code generation for classes via the :func:`numba.jitclass`
-decorator.  A class can be marked for optimization using this decorator along
-with a specification of the types of each field.  We call the resulting class
-object a *jitclass*.  All methods of a jitclass are compiled into nopython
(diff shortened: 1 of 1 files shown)


docs/source/user/jitclass.rst
```diff
@@ -10,7 +10,7 @@ Compiling Python classes with ``@jitclass``
   exposed or implemented, yet.
 
 
-Numba supports code generation for classes via the :func:`numba.jitclass`
+Numba supports code generation for classes via the :func:`numba.experimental.jitclass`
```
stuartarchibald, line 13: Please line wrap to 80 characters, thanks!

stuartarchibald (commented):

docs/source/user/jitclass.rst
dnadlinger: Happy to do so; it will just increase the diff size.

dnadlinger:
Rewrapped (only text change is still adding `.experimental` to the reference).

stuartarchibald (approved):
Thanks for the patch!

docs/source/user/jitclass.rst
stuartarchibald: Thank you :)

PR #6874, opened 2021-03-29
Clarify conversions to numeric types

Fryuni:
The current documentation for `int` support is `only the one-argument form`, but it actually is more restricted than that.

The second argument of the `int` constructor is the base, used only for setting a custom base when converting from `str`, `bytes` or `bytearray`, so just saying it supports one argument only might be interpreted, like I did, as "no custom base is supported", where actually it just supports `Integer`, `Float` or `Boolean` as implemented [here](https://github.com/numba/numba/blob/master/numba/core/typing/builtins.py#L930-L933).

This uses a single argument but fails:
```python
@njit
def foo(val):
    return int(val)

foo('123') # No implementation of function Function(<class 'int'>) found for signature: >>> int(unicode_type)
```

The [python docs](https://docs.python.org/3/library/functions.html#int) also defines the support for casting objects into `int`s, but since passing an object as an argument require using object mode, it works correctly using the python apis, so I don't think it needs to be mentioned here. In object mode string conversions also work correctly:

```python
@jit(forceobj=True)
def foo(val):
    return int(val)

class A:
    def __int__(self):
        return 4

foo(A())
foo('123')
```

The same restriction applies for `float`, but it has a clearer error message:
```python
@njit
def foo(val):
    return float(val)
foo('123')
```
```
No implementation of function Function(<class 'float'>) found for signature:
 
 >>> float(unicode_type)
 
  Overload in function 'Float.generic': File: numba/core/typing/builtins.py: Line 939.
    With argument(s): '(unicode_type)':
     TypeError: float() only support for numbers <=========== The int error message does not have this line
```

I would add that error message on the `int` generic copy [this error](https://github.com/numba/numba/blob/master/numba/core/typing/builtins.py#L945) as a fallback [here](https://github.com/numba/numba/blob/master/numba/core/typing/builtins.py#L934), but I don't know where to test it. I may open another PR for that.

### DIFF
--- a/docs/source/reference/pysupported.rst
+++ b/docs/source/reference/pysupported.rst
@@ -923,7 +923,7 @@ The following built-in functions are supported:
 * :func:`filter`
 * :class:`float`
 * :func:`hash` (see :ref:`pysupported-hashing` below)
-* :class:`int`: only the one-argument form
+* :class:`int`: only the one-argument form, and only for numeric type conversions (Integer, Float and Boolean), string parsing is not supported yet
 * :func:`iter`: only the one-argument form
 * :func:`len`
 * :func:`min`
--- a/docs/source/reference/pysupported.rst
+++ b/docs/source/reference/pysupported.rst
@@ -921,9 +921,9 @@ The following built-in functions are supported:
 * :func:`divmod`
 * :func:`enumerate`
 * :func:`filter`
-* :class:`float`
+* :class:`float`: only for numeric type conversions (Integer, Real)
 * :func:`hash` (see :ref:`pysupported-hashing` below)
-* :class:`int`: only the one-argument form, and only for numeric type conversions (Integer, Float and Boolean), string parsing is not supported yet
+* :class:`int`: only the one-argument form, and only for numeric type conversions (Integer, Float and Boolean)
 * :func:`iter`: only the one-argument form
 * :func:`len`
 * :func:`min`


esc:
@Fryuni thanks for submitting this to the Numba issue tracker. I have added it to the queue for review.

guilhermeleobas:
Hi @Fryuni, thanks for submitting this pull request.

> I would add that error message on the int implementation copying this error as a fallback here, but I don't know where to test it.

You can include the test for `float` and `int` on [`numba/tests/test_builtins.py`.](https://github.com/numba/numba/blob/master/numba/tests/test_builtins.py). See the function [`test_bool`](https://github.com/numba/numba/blob/becab13a5134c92a29b57b1fcde3a308fddc4450/numba/tests/test_builtins.py#L277-L298) to have an idea on how to write tests for `int` and `float` cases.

Let me know if you have any issues

Fryuni:
Now with error message \o/
```
  Overload in function 'Int.generic': File: numba/core/typing/builtins.py: Line 933.
    With argument(s): '(unicode_type, base=int32)':
     TypeError: int() only support for Integer, Float and Boolean
```

I noticed that the the other error messages are not being testes, so I also didn't test  the text, just that there is an error.

I also noticed that the `int` generic was actually not being tested, the test only covered converting from string using pyobject mode and failing without pyobject, so I added tests for all the other cases.

On `float` all the tests are being done inside a single test function, but that causes the second failure condition to not be covered. The `test_float_npm` runs the entire `test_float` without pyobject mode, so it raises a typing error on the string conversion and would not cover the `complex_domain` error case even it was in `test_float`.  
I opted to have each case in a separate test for `int` so I can test all of them both with and without pyobject mode, but I can change it to a single function like `test_float` or organize it in another way.

numba/core/typing/builtins.py
```diff
@@ -940,6 +941,8 @@ def generic(self, args, kws):
         if isinstance(arg, (types.Float, types.Boolean)):
             return signature(types.intp, arg)
 
+        raise TypeError("int() only support for Integer, Float and Boolean")
+
```
guilhermeleobas, line 945: I suppose you just can't raise a TypeError here. This method (`generic`) is called from `AbtractTemplate::apply` (1) and there's a note there saying it is perfectly ok for `generic` to return either a signature or None. In the `Int` case above, if there isn't a match, `generic()` will return nothing (or None). 

(1) https://github.com/numba/numba/blob/3461bcb76a479bffbadbef24b6662107229e2a8a/numba/core/typing/templates.py#L343-L366

If the type you're calling is an `Optional<T>` type, `apply()` will first unwrap it and then try `generic()` again. I suppose it is safe to raise the TypeError only if the type is not an instance of Optional.

Fryuni: Yeah, it was returning `None` before, raising an error causes the error message to be included in the final error/warning message. The idea here is that you might call `int()` with a valid signature by the python docs, but not supported by numba, the error message is not clear in that case.

For example, trying to convert a string to integer, which is allowed in python, says no implementation found for `int(unicode_type)`. In my case when I saw that my thought process was "There is probably some problem converting between unicode, str and bytes like back in Python 2", spent quite a lot of time on that.

This is nothing new, the same is already being done for `float()`, I copied the idea from there.

guilhermeleobas: Yes, but you have a failing test that calls `int()` with an optional type. I suppose the following is ok:
```suggestion
    if not isinstance(arg, types.Optional):
        raise TypeError("int() only support for Integer, Float and Boolean")
```

Fryuni: Now I got what you meant by that. I can't raise it as a default, not that I can't raise it at all. Sorry

guilhermeleobas:
I guess this is ready for review (cc @stuartarchibald, @esc)

Fryuni:
Sorry, I didn't realize this was flagged as waiting on me. Yes, it is ready for review.

PR #6879, opened 2021-03-31
CUDA: NumPy and string dtypes for local and shared arrays

### LINKED ISSUE #5901: cuda: inconsistency between data type specifications in `device_array` and `local.array`

To create `cuda.device_array` of a 32-bit floating-point number type, I can do it as follows:
```python
arr1 = cuda.device_array(100, 'float32')
arr2 = cuda.device_array(100, np.float32)
print(arr1.dtype, arr2.dtype)
```
which prints `float32 float32` as expected.  However, I cannot use `numba.float32` instead:
```python
arr3 = cuda.device_array(100, numba.float32)
````
which yields the error `TypeError: data type not understood`.

By contrast, to create a local array in a Cuda kernel, I have to define it as follows:
```python
@cuda.jit
def test_local_array_kernel():
    arr = cuda.local.array(100, numba.float32)
    print(arr.size)
```
in which neither `np.float32` nor `'float32'` works.  For example, if we use a `np.float32', the following error is produced: 
"type object 'numpy.float32' has no attribute 'is_precise'"

I guess all the above three types refer to the same 32-bit floating-point numbers. Why is there a difference in interfaces of `device_array` and `local.array`? Is it an intentional design?  It would be much better if such an inconsistency can be removed.

gmarkall:
Issue #5901 identified that at some point local arrays would only accept Numba types for their dtype. It appears that they now accept NumPy and string dtypes.

Since this now seems to work, this commit adds tests for passing string and NumPy types as dtypes to both local and shared array constructors.

Fixes #5901.

### DIFF
--- a/numba/cuda/tests/cudapy/test_localmem.py
+++ b/numba/cuda/tests/cudapy/test_localmem.py
@@ -1,7 +1,8 @@
 import numpy as np
 
-from numba import cuda, int32, complex128
-from numba.cuda.testing import unittest, CUDATestCase
+from numba import cuda, int32, complex128, void
+from numba.core.errors import TypingError
+from numba.cuda.testing import unittest, CUDATestCase, skip_on_cudasim
 
 
 def culocal(A, B):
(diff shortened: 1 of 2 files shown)


numba/cuda/tests/cudapy/test_sm_creation.py
```diff
@@ -139,6 +139,56 @@ def test_invalid_4(self):
...
+            s = cuda.shared.array(10, dtype='int32')
+            s[0] = x[0]
+            x[0] = s[0]
+
+        self.check_dtype(f)
+
+    @skip_on_cudasim("Can't check typing in simulator")
+    def test_invalid_string_dtype(self):
+        # Check that strings of invalid dtypes cause a typing error
+        with self.assertRaises(TypingError):
```
stuartarchibald, line 185: Perhaps check the error message string?

numba/cuda/tests/cudapy/test_localmem.py
```diff
@@ -58,6 +59,56 @@ def test_local_array_complex(self):
...
+            l = cuda.local.array(10, dtype='int32')
+            l[0] = x[0]
+            x[0] = l[0]
+
+        self.check_dtype(f)
+
+    @skip_on_cudasim("Can't check typing in simulator")
+    def test_invalid_string_dtype(self):
+        # Check that strings of invalid dtypes cause a typing error
+        with self.assertRaises(TypingError):
```
stuartarchibald, line 105: Perhaps check the error message string?

stuartarchibald (commented):
Thanks for the patch, two minor things to resolve else looks good.

numba/cuda/tests/cudapy/test_sm_creation.py
gmarkall: Done.

numba/cuda/tests/cudapy/test_localmem.py
gmarkall: Done.

stuartarchibald (approved):

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_46`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_46`.

Passed.

PR #6880, opened 2021-03-31
Add attribute lower_extension to CPUContext

reazulhoque:
This PR makes the global `lower_extensions` a part of CPUContext. For Numba IR extension nodes, the pass that adds the extension node can register the method to lower it with the target context.

### DIFF
--- a/numba/core/typed_passes.py
+++ b/numba/core/typed_passes.py
@@ -289,6 +289,13 @@ class ParforPass(FunctionPass):
         """
         Convert data-parallel computations into Parfor nodes
         """
+        # Register lowerer for Parfor Node
+        from numba.parfors.parfor_lowering import _lower_parfor_parallel
+        if hasattr(state.targetctx, "lower_extensions"):
+            state.targetctx.lower_extensions[Parfor] = _lower_parfor_parallel
+        else:
+            raise AttributeError("target_context has no attribute 'lower_extensions'")
+
         # Ensure we have an IR and type information.
         assert state.func_ir
         parfor_pass = _parfor_ParforPass(state.func_ir,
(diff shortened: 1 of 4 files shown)


reazulhoque:
I fixed the issue flake8 was complaining about.

numba/core/typed_passes.py
```diff
@@ -289,6 +289,14 @@ def run_pass(self, state):
         """
         Convert data-parallel computations into Parfor nodes
         """
+        # Register lowerer for Parfor Node
+        from numba.parfors.parfor_lowering import _lower_parfor_parallel
+        if hasattr(state.targetctx, "lower_extensions"):
+            state.targetctx.lower_extensions[Parfor] = _lower_parfor_parallel
```
sklam, line 295: Can we avoid assigning to the target context in the middle the pipeline?

reazulhoque: I absolutely wanted to, but could not find a better place. Please suggest. My thought process: the pass that adds the `special_extension_ir_node` should also add how to lower it.

sklam: I think it make sense for the ``CPUContext.init`` to also insert the ``Parfor`` lowering.

reazulhoque: Should we be inserting the `Parfor` lowering in both places? Or should the initialization be removed from this pass and only be happening in `CPUContext.init`?

sklam: Yes, remove from this pass and only in `CPUContext.init`.

reazulhoque: Done.

sklam (approved):
Thanks for the patch!

PR #6883, opened 2021-04-01
Add support of np.swapaxes #4074

braniii:
This PR is adds `np.swapaxes`, which is currently unimplemented according to #4074 (or `np.swapaxis` which is the mis-spelling version). I had help by @stuartarchibald on discourse: (https://numba.discourse.group/t/how-to-use-types-integer-for-indexing-list/647)

This function can be used to simplify the implementation of `np.rot90` from #6822, which is also the first step of implementing full further arguments of it.

### DIFF
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -5278,14 +5278,22 @@ def numpy_swapaxes(arr, axis1, axis2):
         raise errors.TypingError('The first argument "arr" must be an array')
 
     # create tuple list for transpose
-    axes_list = tuple(range(arr.ndim))
+    ndim = arr.ndim
+    axes_list = tuple(range(ndim))
 
     def impl(arr, axis1, axis2):
+        if axis1 >= ndim or abs(axis1) > ndim:
+            raise ValueError('The second argument "axis1" is out of bounds '
+                             'for array of given dimension')
+        if axis2 >= ndim or abs(axis2) > ndim:
+            raise ValueError('The third argument "axis2" is out of bounds '
+                             'for array of given dimension')
+
         # to ensure tuple_setitem support of negative values
         if axis1 < 0:
-            axis1 += arr.ndim
+            axis1 += ndim
         if axis2 < 0:
-            axis2 += arr.ndim
+            axis2 += ndim
 
         axes_tuple = tuple_setitem(axes_list, axis1, axis2)
         axes_tuple = tuple_setitem(axes_tuple, axis2, axis1)
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -1573,18 +1573,16 @@ def numpy_rot90(arr, k=1):
     if arr.ndim < 2:
         raise ValueError('Input must be >= 2-d.')
 
-    axes_list = (1, 0, *range(2, arr.ndim))
-
     def impl(arr, k=1):
         k = k % 4
         if k == 0:
             return arr[:]
         elif k == 1:
-            return np.transpose(np.fliplr(arr), axes_list)
+            return np.swapaxes(np.fliplr(arr), 0, 1)
         elif k == 2:
             return np.flipud(np.fliplr(arr))
         elif k == 3:
-            return np.fliplr(np.transpose(arr, axes_list))
+            return np.fliplr(np.swapaxes(arr, 0, 1))
         else:
             raise AssertionError  # unreachable
 
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -5266,3 +5266,31 @@ def ol_bool(arr):
                        "is ambiguous. Use a.any() or a.all()")
                 raise ValueError(msg)
         return impl
+
+
+@overload(np.swapaxes)
+def numpy_swapaxes(arr, axis1, axis2):
+    if not isinstance(axis1, (int, types.Integer)):
+        raise errors.TypingError('The second argument "axis1" must be an '
+                                 'integer')
+    if not isinstance(axis2, (int, types.Integer)):
+        raise errors.TypingError('The third argument "axis2" must be an '
+                                 'integer')
+    if not isinstance(arr, types.Array):
+        raise errors.TypingError('The first argument "arr" must be an array')
+
+    # create tuple list for transpose
+    axes_list = tuple(range(arr.ndim))
+
+    def impl(arr, axis1, axis2):
+        # to ensure tuple_setitem support of negative values
+        if axis1 < 0:
+            axis1 += arr.ndim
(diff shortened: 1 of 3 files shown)


esc:
@braniii thank you for submitting this, I have added it to the queue for review.

stuartarchibald:
> This PR is adds `np.swapaxes`, which is currently unimplemented according to #4074 (or `np.swapaxis` which is the name in `numpy <1.16`).

I can't find reference to `np.swapaxis` anywhere in the git log or change log apart from in places where there's people correcting the spelling to `swapaxes`, has it ever existed or is it just a common typo/mis-spelling?

numba/np/arrayobj.py
```diff
@@ -5266,3 +5264,31 @@ def impl(arr):
...
+    # create tuple list for transpose
+    axes_list = tuple(range(arr.ndim))
+
+    def impl(arr, axis1, axis2):
+        # to ensure tuple_setitem support of negative values
+        if axis1 < 0:
+            axis1 += arr.ndim
+        if axis2 < 0:
+            axis2 += arr.ndim
+
```
stuartarchibald, line 5289: I think there probably needs to be some sort of bounds check in here to ensure that the supplied axes are valid. e.g.
```python
@njit
def test():
    return np.swapaxes(np.arange(4), 0, 2)

test()
```
should be illegal.

numba/np/arrayobj.py
```diff
@@ -5266,3 +5264,31 @@ def impl(arr):
...
+@overload(np.swapaxes)
+def numpy_swapaxes(arr, axis1, axis2):
+    if not isinstance(axis1, (int, types.Integer)):
+        raise errors.TypingError('The second argument "axis1" must be an '
+                                 'integer')
+    if not isinstance(axis2, (int, types.Integer)):
+        raise errors.TypingError('The third argument "axis2" must be an '
+                                 'integer')
+    if not isinstance(arr, types.Array):
+        raise errors.TypingError('The first argument "arr" must be an array')
```
stuartarchibald, line 5278: I think that this is permissible given the reliance on `ndim` later, but NumPy supports anything array-like, e.g.:
```python
In [6]: np.swapaxes((1, 2), 0, 0)
Out[6]: array([1, 2])
```

numba/np/arrayobj.py
```diff
@@ -5266,3 +5264,31 @@ def impl(arr):
...
+    def impl(arr, axis1, axis2):
+        # to ensure tuple_setitem support of negative values
+        if axis1 < 0:
+            axis1 += arr.ndim
+        if axis2 < 0:
+            axis2 += arr.ndim
+
+        axes_tuple = tuple_setitem(axes_list, axis1, axis2)
+        axes_tuple = tuple_setitem(axes_tuple, axis2, axis1)
+        return np.transpose(arr, axes_tuple)
```
stuartarchibald, line 5292: I'm reasonably convinced that this will break the "swapaxes returns a view" notion as per the docs, https://numpy.org/doc/stable/reference/generated/numpy.swapaxes.html:
> For NumPy >= 1.10.0, if a is an ndarray, then a view of a is returned; otherwise a new array is created.

Asserting that `a is expected` and `a is got` in the unittests might highlight this.

stuartarchibald (commented):
Thanks for working on this feature. I've given this an initial review, there's a few small things to resolve but the patch is generally good. Thanks again!

braniii:
> > This PR is adds `np.swapaxes`, which is currently unimplemented according to #4074 (or `np.swapaxis` which is the name in `numpy <1.16`).
> 
> I can't find reference to `np.swapaxis` anywhere in the git log or change log apart from in places where there's people correcting the spelling to `swapaxes`, has it ever existed or is it just a common typo/mis-spelling?

When I searched for the function I found multiple results, and I did not consider that numba could be wrong ;) But you are right, I corrected my comment. And thx for the review.

numba/np/arrayobj.py
braniii: I neglected the check, because transpose is checking it, but I included it now.

numba/np/arrayobj.py
braniii: This is true, but I was not able to figure out how to solve it without generating dynamically a tuple. Is it sufficient to mention this in the documentation or does it need to be fixed?

numba/np/arrayobj.py
braniii: `a is expected` and `a is got` should always be false (except when exchanging with the same axis). For being true, not only the data but also the view needs to be the same (which gets changed) or one needs to use the `base` or `np.share_memory`.

You are right, that the definition breaks the "swapaxes returns a view". But the reason is, that the numba version of `np.transpose` does not return a view. Here a short code snippet:

```python
import numpy as np
from numba import njit
b.base is a.base: True
c.base is a.base: False
@njita: False
def transpose(a):
    return np.transpose(a)
c.base is a.base: False

# reference arraype command to continue
a = np.arange(8).reshape(2, -1)

b = np.transpose(a)
c = transpose(a)

# this is always false
print(f'b is a: {b is a}')  # false
print(f'c is a: {c is a}')  # false

# using base it works
print(f'b.base is a.base: {b.base is a.base}')  # true
print(f'c.base is a.base: {c.base is a.base}')  # false
```

numba/np/arrayobj.py
stuartarchibald: I think it's fine to leave for now in the interests of completing the feature. The compiler will catch invalid types and if its really important to someone it's likely they'll open an issue.

numba/np/arrayobj.py
stuartarchibald: I don't think there's much we can do about this right now? Do you feel that this behaviour is prohibitive to it being useful as-is?

braniii: At least from my side there is not much I can do, I have not enough inside in the code base to propose a fix. But I think the function is useful as-is. But in general it could be helpful to document difference from numpy functions, maybe simple at
https://numba.readthedocs.io/en/stable/reference/numpysupported.html

```diff
- `np.transpose()`
+ `np.transpose()` (a copy is returned)
```

stuartarchibald: > At least from my side there is not much I can do, I have not enough inside in the code base to propose a fix. But I think the function is useful as-is. But in general it could be helpful to document difference from numpy functions, maybe simple at
> https://numba.readthedocs.io/en/stable/reference/numpysupported.html
> 
> ```diff
> - `np.transpose()`
> + `np.transpose()` (a copy is returned)
> ```

Agree, this is probably the best thing to do for now.

stuartarchibald (approved):

PR #6888, opened 2021-04-01
Get overload to consider compiler flags in cache lookup

sklam:
as titled

### DIFF
--- a/numba/core/utils.py
+++ b/numba/core/utils.py
@@ -188,6 +188,18 @@ class ConfigStack:
     """
     tls = threading.local()
 
+    @classmethod
+    def top_or_none(cls):
+        """Get the TOS of return None if no config is set.
+        """
+        self = cls()
+        if self:
+            flags = self.top()
+        else:
+            # Note: should this be the default flag for the target instead?
+            flags = None
+        return flags
+
     def __init__(self):
         tls = self.tls
         try:
--- a/numba/core/utils.py
+++ b/numba/core/utils.py
@@ -190,7 +190,7 @@ class ConfigStack:
 
     @classmethod
     def top_or_none(cls):
-        """Get the TOS of return None if no config is set.
+        """Get the TOS or return None if no config is set.
         """
         self = cls()
         if self:
--- a/numba/core/typing/templates.py
+++ b/numba/core/typing/templates.py
@@ -684,11 +684,7 @@ class _OverloadFunctionTemplate(AbstractTemplate):
         Returning a Dispatcher object.  The Dispatcher object is cached
         internally in `self._impl_cache`.
         """
-        cfgstk = utils.ConfigStack()
-        if cfgstk:
-            flags = cfgstk.top()
-        else:
-            flags = None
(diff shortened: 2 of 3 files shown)


numba/core/utils.py
```diff
@@ -188,6 +188,18 @@ class ConfigStack:
...
 
+    @classmethod
+    def top_or_none(cls):
+        """Get the TOS of return None if no config is set.
+        """
+        self = cls()
+        if self:
+            flags = self.top()
+        else:
+            # Note: should this be the default flag for the target instead?
```
stuartarchibald, line 199: I think perhaps it should be the default so it is always known precisely which flags are in use for a given function.

stuartarchibald (commented):
Thanks for the patch, I think this fixes the issue with ensuring that overloads are selected based on flags.

numba/core/utils.py
sklam: I'm getting into a rabbit hole trying to put a default flag there. Currently, the `TargetOptions` is not really part of any contexts. It's floating in the `TargetDescriptor`. I can put the options into the target-context but it doesn't make sense for it to appear in the typing-context. However, I need it during typing. This goes back to the purity problem of needing flags (which I don't think is part of types) during the typing phase.

The only doable hack I can think of is putting the info into the `ConfigStack` to read the default flag later.

sklam: hm.... the other way is to let flags be part of the typing context.

sklam: I sampled the testsuite and found that this branch is only reached from test that bypass the user-facing compiler API---e.g. tests that invokes part of the compiler to test for specific internal behavior. The failure to use a default flag will only result in redundant compilation. I don't think we should spend too much time on this problem for now.

stuartarchibald: From OOB conversation, agree on leaving it for now as this fixes the immediate issue.

sklam:
https://github.com/numba/numba/pull/6898 may provide new facility that this PR can depend on. Also, not sure if this smaller PR will conflict with the work in the larger #6898.

numba/core/utils.py
```diff
@@ -188,6 +188,18 @@ class ConfigStack:
     """
     tls = threading.local()
 
+    @classmethod
+    def top_or_none(cls):
+        """Get the TOS of return None if no config is set.
```
stuartarchibald, line 193:
```suggestion
        """Get the TOS or return None if no config is set.
```

stuartarchibald (commented):
Thanks for the patch. There's a typo to resolve, else I think this should be merged as it'll permit multiple lines of work in figuring out what needs fixing else where!

@sklam IIRC you ran the test suite to check this and `None` for `flags` in the cache key was quite rare and seemed like a harmless recompile?

sklam:
> @sklam IIRC you ran the test suite to check this and `None` for `flags` in the cache key was quite rare and seemed like a harmless recompile?

Correct. It only happens from our testsuite using `compile_isolated()`

stuartarchibald (approved):

Issue #6159, opened 2020-08-23
guvectorize with cuda target is extremely slow


rpopovici:
I am trying to use guvectorize with cuda target. The result is not what I expected. Processig 100Mb of data takes 8s. 
I have a GTX 1070 with 8Gb of RAM. Cuda toolkit installed is 10.2 on windows 10. I tried the official build 0.51 and latest master as well with the same outcome. 

nb.cuda.detect() result
```
Found 1 CUDA devices
id 0     b'GeForce GTX 1070'                              [SUPPORTED]
                      compute capability: 6.1
                           pci device id: 0
                              pci bus id: 1
Summary:
	1/1 devices are supported
```

Example code:
```
import numpy as np
import numba as nb
from numba import jit, njit, vectorize, guvectorize, typeof, prange
from numba.types import float32
from numba import cuda

a = np.random.rand(1024 * 1024 * 32).astype('float32')
b = np.random.rand(1024 * 1024 * 32).astype('float32')
dist = np.zeros(a.shape[0]).astype('float32')

@guvectorize(['void(float32[:], float32[:], float32[:])'], '(n),(n)->(n)', nopython=True, target='cuda')
def numba_dist_cuda(a, b, dist):
    len = a.shape[0]
    for i in range(len):
        dist[i] = a[i] * b[i]

nb.cuda.detect()
print(nb.cuda.is_available())

import time
t = time.time()

numba_dist_cuda(a, b, dist)

elapsed = time.time() - t
print(elapsed)
```

rpopovici:
cpu target 0.03 
parallel target 0.06
cuda target 8s

What am I missing?

gmarkall:
Modified reproducer that doesn't time the copy from host to device:

```python
import numpy as np
import numba as nb
from numba import jit, njit, vectorize, guvectorize, typeof, prange
from numba.types import float32
from numba import cuda

a = np.random.rand(1024 * 1024 * 32).astype('float32')
b = np.random.rand(1024 * 1024 * 32).astype('float32')
dist = np.zeros(a.shape[0]).astype('float32')

@guvectorize(['void(float32[:], float32[:], float32[:])'], '(n),(n)->(n)',
             nopython=True, target='cuda')
def numba_dist_cuda(a, b, dist):
    len = a.shape[0]
    for i in range(len):
        dist[i] = a[i] * b[i]

nb.cuda.detect()
print(nb.cuda.is_available())

import time

d_a = cuda.to_device(a)
[...]

PR #5189, opened 2020-02-03
__cuda_array_interface__ not requiring context

madsbk:
This PR makes `DeviceNDArrayBase.__cuda_array_interface__` not require CUDA context, which can be quite expensive.

Before, it would call `array_core()` which calls `__getitem__()` which is wrapped in `@devices.require_context`

PR #6889, opened 2021-04-02
Address guvectorize too slow for cuda target

testhound:
This pull request address: https://github.com/numba/numba/issues/6159. It addresses the issues by detecting when the blocksize for a kernel would be < 2 * Number of SM's. I also added a environment variable called NUMBA_CUDA_LOW_OCCUPANCY_WARNINGS that is enabled by default. I had to modify the the class CUDATestCase to disable the warning as many of the test cases trigger this warning.

### DIFF
--- a/docs/source/reference/envvars.rst
+++ b/docs/source/reference/envvars.rst
@@ -470,8 +470,7 @@ GPU support
 .. envvar:: NUMBA_CUDA_LOW_OCCUPANCY_WARNINGS
 
    Enable warnings if the grid size is too small relative to the number of
-   steraming multiprocessors (SM). If the block size is too small an inefficient
-   schedule will result and performance will suffer. This option is on by default (1).
+   steraming multiprocessors (SM). This option is on by default (1).
 
    The heuristic checked is whether gridsize < 2 * (number of SMs). NOTE: The absence of
    a warning does not imply a good gridsize relative to the number of SMs. Disabling
--- a/docs/source/reference/envvars.rst
+++ b/docs/source/reference/envvars.rst
@@ -470,12 +470,12 @@ GPU support
 .. envvar:: NUMBA_CUDA_LOW_OCCUPANCY_WARNINGS
 
    Enable warnings if the grid size is too small relative to the number of
-   steraming multiprocessors (SM). This option is on by default (1).
+   streaming multiprocessors (SM). This option is on by default (default  value is 1).
 
-   The heuristic checked is whether gridsize < 2 * (number of SMs). NOTE: The absence of
+   The heuristic checked is whether ``gridsize < 2 * (number of SMs)``. NOTE: The absence of
    a warning does not imply a good gridsize relative to the number of SMs. Disabling
-   this warning will reduce the number of CUDA API calls, as the heuristic needs to
-   check the number of SMs for the device in the current context.
+   this warning will reduce the number of CUDA API calls (during JIT compilation), as the
+   heuristic needs to check the number of SMs for the device in the current context.
 
 
 Threading Control
--- a/docs/source/reference/envvars.rst
+++ b/docs/source/reference/envvars.rst
@@ -469,9 +469,14 @@ GPU support
 
 .. envvar:: NUMBA_CUDA_LOW_OCCUPANCY_WARNINGS
 
-   Enable warnings if the block size is too small relative to the number of
+   Enable warnings if the grid size is too small relative to the number of
    steraming multiprocessors (SM). If the block size is too small an inefficient
-   schedule will result and performance will suffer.
+   schedule will result and performance will suffer. This option is on by default (1).
+
+   The heuristic checked is whether gridsize < 2 * (number of SMs). NOTE: The absence of
+   a warning does not imply a good gridsize relative to the number of SMs. Disabling
+   this warning will reduce the number of CUDA API calls, as the heuristic needs to
+   check the number of SMs for the device in the current context.
 
 
 Threading Control
--- a/docs/source/reference/envvars.rst
+++ b/docs/source/reference/envvars.rst
@@ -467,6 +467,12 @@ GPU support
    Strict strides checking is deprecated and may be removed in future. See
    :ref:`deprecation-strict-strides`.
 
+.. envvar:: NUMBA_CUDA_LOW_OCCUPANCY_WARNINGS
+
+   Enable warnings if the block size is too small relative to the number of
+   steraming multiprocessors (SM). If the block size is too small an inefficient
+   schedule will result and performance will suffer.
+
 
 Threading Control
 -----------------
--- a/docs/source/reference/envvars.rst
+++ b/docs/source/reference/envvars.rst
@@ -470,12 +470,13 @@ GPU support
 .. envvar:: NUMBA_CUDA_LOW_OCCUPANCY_WARNINGS
 
    Enable warnings if the grid size is too small relative to the number of
-   streaming multiprocessors (SM). This option is on by default (default  value is 1).
+   streaming multiprocessors (SM). This option is on by default (default value is 1).
 
    The heuristic checked is whether ``gridsize < 2 * (number of SMs)``. NOTE: The absence of
    a warning does not imply a good gridsize relative to the number of SMs. Disabling
    this warning will reduce the number of CUDA API calls (during JIT compilation), as the
-   heuristic needs to check the number of SMs for the device in the current context.
+   heuristic needs to check the number of SMs available on the device in the
+   current context.
 
 
 Threading Control
--- a/numba/core/config.py
+++ b/numba/core/config.py
@@ -154,6 +154,11 @@ class _EnvReloader(object):
             "NUMBA_ALWAYS_WARN_UNINIT_VAR", int, 0,
         )
 
+        # whether to warn about kernels where the block size will
+        # result in a inefficient mapping to GPUs. On by default.
+        LOW_OCCUPANCY_WARNINGS = _readenv(
+            "NUMBA_CUDA_LOW_OCCUPANCY_WARNINGS", int, 1)
+
         # Debug flag to control compiler debug print
         DEBUG = _readenv("NUMBA_DEBUG", int, 0)
 
--- a/numba/core/config.py
+++ b/numba/core/config.py
@@ -155,8 +155,8 @@ class _EnvReloader(object):
         )
 
         # Whether to warn about kernel launches where the grid size will
-        # underutilize the GPU due to low occupancy. On by default.
-        LOW_OCCUPANCY_WARNINGS = _readenv(
+        # under utilize the GPU due to low occupancy. On by default.
+        CUDA_LOW_OCCUPANCY_WARNINGS = _readenv(
             "NUMBA_CUDA_LOW_OCCUPANCY_WARNINGS", int, 1)
 
         # Debug flag to control compiler debug print
--- a/numba/core/config.py
+++ b/numba/core/config.py
@@ -154,8 +154,8 @@ class _EnvReloader(object):
             "NUMBA_ALWAYS_WARN_UNINIT_VAR", int, 0,
         )
 
-        # whether to warn about kernels where the block size will
-        # result in a inefficient mapping to GPUs. On by default.
+        # Whether to warn about kernel launches where the grid size will
+        # underutilize the GPU due to low occupancy. On by default.
         LOW_OCCUPANCY_WARNINGS = _readenv(
             "NUMBA_CUDA_LOW_OCCUPANCY_WARNINGS", int, 1)
 
--- a/numba/cuda/compiler.py
+++ b/numba/cuda/compiler.py
@@ -787,13 +787,13 @@ class _KernelConfiguration:
         self.stream = stream
         self.sharedmem = sharedmem
 
-        if config.LOW_OCCUPANCY_WARNINGS:
+        if config.CUDA_LOW_OCCUPANCY_WARNINGS:
             ctx = get_context()
             smcount = ctx.device.MULTIPROCESSOR_COUNT
             grid_size = griddim[0] * griddim[1] * griddim[2]
             if grid_size < 2 * smcount:
                 msg = ("Grid size ({grid}) < 2 * SM count ({sm}) "
-                       "will likely result in GPU underutilization due"
+                       "will likely result in GPU under utilization due"
                        "to low occupancy.")
                 msg = msg.format(grid=grid_size, sm=2 * smcount)
                 warn(NumbaPerformanceWarning(msg))
--- a/numba/cuda/compiler.py
+++ b/numba/cuda/compiler.py
@@ -787,15 +787,16 @@ class _KernelConfiguration:
         self.stream = stream
         self.sharedmem = sharedmem
 
-        num_blocks = griddim[0] * griddim[1] * griddim[2]
-        ctx = get_context()
-        smcount = ctx.device.MULTIPROCESSOR_COUNT
-
-        if (config.LOW_OCCUPANCY_WARNINGS) and (num_blocks < 2 * smcount):
-            msg = ("Number of blocks ({blocks}) that are  < 2 * SM ({sm}) "
-                   "will generate inefficient kernel code.")
-            msg = msg.format(blocks=num_blocks, sm=2 * smcount)
-            warn(NumbaPerformanceWarning(msg))
+        if config.LOW_OCCUPANCY_WARNINGS:
+            ctx = get_context()
+            smcount = ctx.device.MULTIPROCESSOR_COUNT
+            grid_size = griddim[0] * griddim[1] * griddim[2]
+            if grid_size < 2 * smcount:
+                msg = ("Grid size ({grid}) < 2 * SM count ({sm}) "
+                       "will likely result in GPU underutilization due"
+                       "to low occupancy.")
+                msg = msg.format(grid=grid_size, sm=2 * smcount)
+                warn(NumbaPerformanceWarning(msg))
 
     def __call__(self, *args):
         return self.dispatcher.call(args, self.griddim, self.blockdim,
--- a/numba/cuda/compiler.py
+++ b/numba/cuda/compiler.py
@@ -791,7 +791,7 @@ class _KernelConfiguration:
         ctx = get_context()
         smcount = ctx.device.MULTIPROCESSOR_COUNT
 
-        if num_blocks < 2 * smcount:
+        if (config.LOW_OCCUPANCY_WARNINGS) and (num_blocks < 2 * smcount):
             msg = ("Number of blocks ({blocks}) that are  < 2 * SM ({sm}) "
                    "will generate inefficient kernel code.")
             msg = msg.format(blocks=num_blocks, sm=2 * smcount)
--- a/numba/cuda/compiler.py
+++ b/numba/cuda/compiler.py
@@ -29,6 +29,7 @@ from .cudadrv import nvvm, driver
 from .errors import missing_launch_config_msg, normalize_kernel_dimensions
 from .api import get_current_device
 from .args import wrap_arg
+from numba.cuda.cudadrv.driver import driver
 
 
 def _nvvm_options_type(x):
@@ -785,6 +786,21 @@ class _KernelConfiguration:
         self.blockdim = blockdim
         self.stream = stream
         self.sharedmem = sharedmem
+
+        num_blocks = griddim[0] * griddim[1] * griddim[2]
+        ctx = get_context()
+        smcount = ctx.device.MULTIPROCESSOR_COUNT
+
+        print (griddim)
+        print (blockdim)
+        print (smcount)
+
+        if num_blocks < 2 * smcount:
+            msg = ("Number of blocks ({blocks}) that are  < 2 * SM ({sm}) "
+                  "will generate inefficient kernel code.")
+            msg = msg.format(blocks=num_blocks, sm=2 * smcount)
+            #warn(msg, category=RuntimeWarning)
+            warn(msg)
 
     def __call__(self, *args):
         return self.dispatcher.call(args, self.griddim, self.blockdim,
--- a/numba/cuda/compiler.py
+++ b/numba/cuda/compiler.py
@@ -29,7 +29,7 @@ from .cudadrv import nvvm, driver
 from .errors import missing_launch_config_msg, normalize_kernel_dimensions
 from .api import get_current_device
 from .args import wrap_arg
-from numba.cuda.cudadrv.driver import driver
+from numba.core.errors import NumbaPerformanceWarning
 
 
 def _nvvm_options_type(x):
@@ -791,16 +791,11 @@ class _KernelConfiguration:
         ctx = get_context()
         smcount = ctx.device.MULTIPROCESSOR_COUNT
 
-        print (griddim)
-        print (blockdim)
-        print (smcount)
-
         if num_blocks < 2 * smcount:
             msg = ("Number of blocks ({blocks}) that are  < 2 * SM ({sm}) "
-                  "will generate inefficient kernel code.")
+                   "will generate inefficient kernel code.")
             msg = msg.format(blocks=num_blocks, sm=2 * smcount)
-            #warn(msg, category=RuntimeWarning)
-            warn(msg)
+            warn(NumbaPerformanceWarning(msg))
 
     def __call__(self, *args):
         return self.dispatcher.call(args, self.griddim, self.blockdim,
--- a/numba/cuda/testing.py
+++ b/numba/cuda/testing.py
@@ -22,6 +22,11 @@ class CUDATestCase(SerialMixin, TestCase):
     the context and destroy resources used by a normal CUDATestCase if any of
     its tests are run between tests from a CUDATestCase.
     """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._low_occupancy_warnings = config.LOW_OCCUPANCY_WARNINGS
+
     def setUp(self):
         self._low_occupancy_warnings = config.LOW_OCCUPANCY_WARNINGS
         # Disable warnings about low gpu utilization for test suite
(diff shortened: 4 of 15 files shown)


sklam:
Started smoketest: `numba_smoketest_cuda_148`

docs/source/reference/envvars.rst
```diff
@@ -467,6 +467,12 @@ GPU support
    Strict strides checking is deprecated and may be removed in future. See
    :ref:`deprecation-strict-strides`.
 
+.. envvar:: NUMBA_CUDA_LOW_OCCUPANCY_WARNINGS
+
+   Enable warnings if the block size is too small relative to the number of
```
gmarkall, line 472: The error is generated when the grid size is too small, not the block size. It's probably worth mentioning that the heuristic used here is `(grid size) < 2 * (number of SMs)`, and that the absence of the warning doesn't necessarily imply a good grid size - larger multiples of the SM count may be more efficient.

numba/cuda/tests/cudapy/test_gufunc.py
```diff
@@ -183,6 +187,28 @@ def copy2d(A, B):
         copy2d(A, out=B)
         self.assertTrue(np.allclose(A, B))
 
+    def test_inefficient_guvectorize(self):
```
gmarkall, line 190: I think this could do with a comment explaining that the test creates an inefficient use of the GPU because the inputs are all mapped onto a single thread in a single block.

docs/source/reference/envvars.rst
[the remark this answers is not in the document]
gmarkall: Also it should be mentioned that this is on by default.

numba/cuda/compiler.py
```diff
@@ -786,6 +787,16 @@ def __init__(self, dispatcher, griddim, blockdim, stream, sharedmem):
         self.stream = stream
         self.sharedmem = sharedmem
 
+        num_blocks = griddim[0] * griddim[1] * griddim[2]
+        ctx = get_context()
+        smcount = ctx.device.MULTIPROCESSOR_COUNT
+
+        if (config.LOW_OCCUPANCY_WARNINGS) and (num_blocks < 2 * smcount):
+            msg = ("Number of blocks ({blocks}) that are  < 2 * SM ({sm}) "
```
gmarkall, line 795: I think `will generate inefficient kernel code` is a bit confusing - how about phrasing the warning as `Grid size (<size>) less than 2 * SM count (2 * (<SM count>) = (<SM count * 2>) will likely result in GPU underutilization due to low occupancy`?

gmarkall (changes requested):
This is looking promising; there are some comments on the diff suggesting some changes.

It's good that there is a test ensuring this fires for inefficient uses of `@guvectorize` functions. I think there should also be a test of this using a normal kernel launch - I can't see a really good place to put this new test, but either in `numba/cuda/tests/cudapy/test_errors.py` or in a new file `test_warnings.py`.

testhound:
@gmarkall I believe all your review comments have been incorporated.

numba/core/config.py
```diff
@@ -154,6 +154,11 @@ def optional_str(x):
             "NUMBA_ALWAYS_WARN_UNINIT_VAR", int, 0,
         )
 
+        # Whether to warn about kernel launches where the grid size will
+        # underutilize the GPU due to low occupancy. On by default.
+        LOW_OCCUPANCY_WARNINGS = _readenv(
```
stuartarchibald, line 159:
```suggestion
        CUDA_LOW_OCCUPANCY_WARNINGS = _readenv(
```
All other CUDA related config vars start with `CUDA_` or have `CUDA` in their name, suggest calling this something like the above.

numba/cuda/testing.py
```diff
@@ -23,6 +23,18 @@ class CUDATestCase(SerialMixin, TestCase):
     its tests are run between tests from a CUDATestCase.
     """
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._low_occupancy_warnings = config.LOW_OCCUPANCY_WARNINGS
+
+    def setUp(self):
+        self._low_occupancy_warnings = config.LOW_OCCUPANCY_WARNINGS
+        # Disable warnings about low gpu utilization for test suite
```
stuartarchibald, line 32:
```suggestion
        # Disable warnings about low gpu utilization in the test suite
```

numba/cuda/compiler.py
```diff
@@ -786,6 +787,17 @@ def __init__(self, dispatcher, griddim, blockdim, stream, sharedmem):
         self.stream = stream
         self.sharedmem = sharedmem
 
+        if config.LOW_OCCUPANCY_WARNINGS:
+            ctx = get_context()
+            smcount = ctx.device.MULTIPROCESSOR_COUNT
+            grid_size = griddim[0] * griddim[1] * griddim[2]
+            if grid_size < 2 * smcount:
+                msg = ("Grid size ({grid}) < 2 * SM count ({sm}) "
+                       "will likely result in GPU underutilization due"
```
stuartarchibald, line 796:
```suggestion
                       "will likely result in GPU under utilization due"
```

numba/core/config.py
```diff
@@ -154,6 +154,11 @@ def optional_str(x):
             "NUMBA_ALWAYS_WARN_UNINIT_VAR", int, 0,
         )
 
+        # Whether to warn about kernel launches where the grid size will
+        # underutilize the GPU due to low occupancy. On by default.
```
stuartarchibald, line 158:
```suggestion
        # under utilize the GPU due to low occupancy. On by default.
```

docs/source/reference/envvars.rst
```diff
@@ -467,6 +467,16 @@ GPU support
    Strict strides checking is deprecated and may be removed in future. See
    :ref:`deprecation-strict-strides`.
 
+.. envvar:: NUMBA_CUDA_LOW_OCCUPANCY_WARNINGS
+
+   Enable warnings if the grid size is too small relative to the number of
+   steraming multiprocessors (SM). This option is on by default (1).
```
stuartarchibald, line 473:
```suggestion
   streaming multiprocessors (SM). This option is on by default (default value is 1).
```

docs/source/reference/envvars.rst
```diff
@@ -467,6 +467,16 @@ GPU support
...
+
+   The heuristic checked is whether gridsize < 2 * (number of SMs). NOTE: The absence of
```
stuartarchibald, line 475:
```suggestion
   The heuristic checked is whether ``gridsize < 2 * (number of SMs)``. NOTE: The absence of
```

docs/source/reference/envvars.rst
```diff
@@ -467,6 +467,16 @@ GPU support
...
+   a warning does not imply a good gridsize relative to the number of SMs. Disabling
+   this warning will reduce the number of CUDA API calls, as the heuristic needs to
+   check the number of SMs for the device in the current context.
```
stuartarchibald, line 478: Does lead to a noticeable compilation speed gain?

stuartarchibald (commented):
Thanks for the patch, this is great feedback to give users, it's good to see this implemented. I've a few minor comments in addition to @gmarkall's review, once resolved looks good to merge. Thanks again!

docs/source/reference/envvars.rst
gmarkall: This is for launch speed rather than compilation speed. In the past there have been performance issues due to excessive getting of contexts, e.g.: https://github.com/numba/numba/pull/5189 - so I think if we add a get of the context in the launch pipeline, it's good to have a way around it as well.

gmarkall:
> @gmarkall I believe all your review comments have been incorporated.

Thanks for looking at these. I think they are all addressed, except this one (from the initial review):

> I think there should also be a test of this using a normal kernel launch - I can't see a really good place to put this new test, but either in numba/cuda/tests/cudapy/test_errors.py or in a new file test_warnings.py.

numba/core/config.py
testhound: Done.

numba/cuda/testing.py
testhound: Done

numba/cuda/compiler.py
testhound: Done.

numba/core/config.py
testhound: Done.

docs/source/reference/envvars.rst
testhound: Done.

docs/source/reference/envvars.rst
testhound: Done.

testhound:
> > @gmarkall I believe all your review comments have been incorporated.
> 
> Thanks for looking at these. I think they are all addressed, except this one (from the initial review):
> 
> > I think there should also be a test of this using a normal kernel launch - I can't see a really good place to put this new test, but either in numba/cuda/tests/cudapy/test_errors.py or in a new file test_warnings.py.

@gmarkall I addressed this with the function 'test_efficient_guvectorize' in the file test_gufunc.py.

gmarkall:
Sorry, I forgot to respond earlier - I mean a test where you call a kernel like `f[1,1](arg)` and check the message is produced (and one that launches with better occupancy and checks that there is no warning). I do think `test_efficient_guvectorize` is a good addition too though.

numba/cuda/tests/cudapy/test_gufunc.py
gmarkall: (Note: this is resolved, but I can't resolve the comment because I don't have write access to this repo)

numba/cuda/tests/cudapy/test_gufunc.py
```diff
@@ -183,6 +186,49 @@ def copy2d(A, B):
...
+        @guvectorize(['void(float32[:], float32[:], float32[:])'],
+                     '(n),(n)->(n)', target='cuda')
+        def numba_dist_cuda(a, b, dist):
+            len = a.shape[0]
+            for i in range(len):
+                dist[i] = a[i] * b[i]
+
+        a = np.random.rand(1024 * 1024 * 32).astype('float32')
+        b = np.random.rand(1024 * 1024 * 32).astype('float32')
```
gmarkall, line 200: Looking at the performance of the kernel test led me to also look into how long the guvectorize tests take - I think the data sizes for this and the test below are too large for unit tests, and they take about 5 seconds to run - can we reduce the size of the data please - e.g.:

```suggestion
        a = np.random.rand(1024 * 32).astype('float32')
        b = np.random.rand(1024 * 32).astype('float32')
```

numba/cuda/tests/cudapy/test_gufunc.py
```diff
@@ -183,6 +186,49 @@ def copy2d(A, B):
...
+                     '(n),(n)->(n)', nopython=True, target='cuda')
+        def numba_dist_cuda2(a, b, dist):
+            len = a.shape[0]
+            for i in range(len):
+                dist[i] = a[i] * b[i]
+
+        a = np.random.rand(1024 * 1024 * 128).astype('float32').\
+            reshape((1024 * 1024, 128))
+        b = np.random.rand(1024 * 1024 * 128).astype('float32').\
+            reshape((1024 * 1024, 128))
```
gmarkall, line 223: Also to reduce the data size here please:

```suggestion
        a = np.random.rand(524288 * 2).astype('float32').reshape((524288, 2))
        b = np.random.rand(524288 * 2).astype('float32').reshape((524288, 2))
```

The leading dimension of 524288 is needed to get a big enough grid size for a large GPU, so I reduced the trailing dimension down to 2 to keep the data size as small as possible.

gmarkall (changes requested):
Thanks for the updated tests! I have a few comments regarding the kernel launch test - rather than annotating lines on the diff, I'll try and summarise here:

- The tests are quite complex compared to what's needed, resulting in the runtime being too long and the purpose of the tests being slightly obscured - the runtime on my machine is 40 seconds for `test_warnings.py`. Since we're only trying to test whether the warning is generated for a low occupancy configuration, we don't need the kernel to do anything or create any arrays to pass into it.
- I think that even if the test runtime is reduced a lot, the test still doesn't work in the CUDA simulator, because it doesn't generate performance warnings (and we shouldn't try to make it do so).
- The names of the tests are a bit vague about what is really tested - `inefficient_kernel` implies that the problem is inefficient code in the kernel, rather than the launch configuration.
- A minor point: I try to not include `CUDA` in the names of CUDA test classes, because it's redundant in the CUDA test suite - the main issue I have with it is having to type more when running individual classes from the command line :-)

I think the above can be resolved with the following implementation of the test class:

```python
@skip_on_cudasim('cudasim does not raise performance warnings')
class TestWarnings(CUDATestCase):
    def test_inefficient_launch_configuration(self):
        @cuda.jit
        def kernel():
            pass

        with override_config('CUDA_LOW_OCCUPANCY_WARNINGS', 1):
            with warnings.catch_warnings(record=True) as w:
                kernel[1, 1]()

        self.assertEqual(w[0].category, NumbaPerformanceWarning)
        self.assertIn('Grid size', str(w[0].message))
        self.assertIn('2 * SM count', str(w[0].message))

    def test_efficient_launch_configuration(self):
        @cuda.jit
        def kernel():
            pass

        with override_config('CUDA_LOW_OCCUPANCY_WARNINGS', 1):
            with warnings.catch_warnings(record=True) as w:
                kernel[256, 256]()

        self.assertEqual(len(w), 0)
```

There are also a couple of comments / questions on the diff to look at - one is related to the way `setUp` works, which turned out to be more complex than I first imagined it would be.

numba/cuda/testing.py
```diff
@@ -23,6 +23,18 @@ class CUDATestCase(SerialMixin, TestCase):
     its tests are run between tests from a CUDATestCase.
     """
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._low_occupancy_warnings = config.CUDA_LOW_OCCUPANCY_WARNINGS
```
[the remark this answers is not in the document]
testhound: @gmarkall good catch on setUpClass. I don't think however, removing the __init __ method and it's implementation is correct. If you remove __init__ then 'self._low_occupancy_warnings" will not be created as a instance variable.

testhound:
@gmarkall should the tests in test_gufunc be skipped for the CUDA simulator (since no warnings are generated in the simulator), similar to the new ones in test_warning.py?

numba/cuda/testing.py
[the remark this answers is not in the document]
gmarkall: It will be created during the `setUp()` method - it doesn't need to be "declared" in `__init__()`.

gmarkall:
> @gmarkall should the tests in test_gufunc be skipped for the CUDA simulator (since no warnings are generated in the simulator), similar to the new ones in test_warning.py?

The whole test class there is skipped, so there's no need for additional decoration of the test methods you've added.

gmarkall (dismissed):
Many thanks for the updates! This looks good to me now, and passes tests locally.

@Stuartarchibald / @esc could this have a buildfarm run please? The test fail showing presently is something broken in the main branch.

sklam:
BFID `numba_smoketest_cuda_yaml_60`

docs/source/reference/envvars.rst
```diff
@@ -467,6 +467,16 @@ GPU support
    Strict strides checking is deprecated and may be removed in future. See
    :ref:`deprecation-strict-strides`.
 
+.. envvar:: NUMBA_CUDA_LOW_OCCUPANCY_WARNINGS
+
+   Enable warnings if the grid size is too small relative to the number of
+   streaming multiprocessors (SM). This option is on by default (default  value is 1).
```
stuartarchibald, line 473:
```suggestion
   streaming multiprocessors (SM). This option is on by default (default value is 1).
```

docs/source/reference/envvars.rst
```diff
@@ -467,6 +467,16 @@ GPU support
...
+
+   The heuristic checked is whether ``gridsize < 2 * (number of SMs)``. NOTE: The absence of
+   a warning does not imply a good gridsize relative to the number of SMs. Disabling
+   this warning will reduce the number of CUDA API calls (during JIT compilation), as the
+   heuristic needs to check the number of SMs for the device in the current context.
```
stuartarchibald, line 478:
```suggestion
   heuristic needs to check the number of SMs available on the device in the
   current context.
```

stuartarchibald (commented):
Thanks for the patch, couple of tiny doc change suggestions else looks good. Thanks for reviewing @gmarkall .

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_60`.

stuartarchibald:
> BFID `numba_smoketest_cuda_yaml_60`

This passed.

docs/source/reference/envvars.rst
testhound: @stuartarchibald change committed.

docs/source/reference/envvars.rst
testhound: @stuartarchibald change committed.

sklam:
`numba.tests.test_config.TestConfig.test_config` is failing.

testhound:
> `numba.tests.test_config.TestConfig.test_config` is failing.

Yes I see that but it's not happening on all platforms or on my local machine.

gmarkall:
> > `numba.tests.test_config.TestConfig.test_config` is failing.
> 
> Yes I see that but it's not happening on all platforms or on my local machine.

It does seem repeatable though. Since the diff is not visible, to debug this I'd push a commit that sets `self.maxDiff` to something greater than 2109 in `numba.tests.test_config.TestConfig` as suggested in the error log to see what the difference is.

testhound:
> > > `numba.tests.test_config.TestConfig.test_config` is failing.
> > 
> > 
> > Yes I see that but it's not happening on all platforms or on my local machine.
> 
> It does seem repeatable though. Since the diff is not visible, to debug this I'd push a commit that sets `self.maxDiff` to something greater than 2109 in `numba.tests.test_config.TestConfig` as suggested in the error log to see what the difference is.

@gmarkall okay done.

numba/cuda/compiler.py
```diff
@@ -823,6 +824,17 @@ def __init__(self, dispatcher, griddim, blockdim, stream, sharedmem):
...
         self.sharedmem = sharedmem
 
+        if config.CUDA_LOW_OCCUPANCY_WARNINGS:
+            ctx = get_context()
+            smcount = ctx.device.MULTIPROCESSOR_COUNT
+            grid_size = griddim[0] * griddim[1] * griddim[2]
+            if grid_size < 2 * smcount:
+                msg = ("Grid size ({grid}) < 2 * SM count ({sm}) "
+                       "will likely result in GPU under utilization due"
+                       "to low occupancy.")
```
gmarkall, line 834: I just noticed there's a space missing here so it ends up printing "dueto low occupancy."

```suggestion
                       "will likely result in GPU under utilization due "
                       "to low occupancy.")
```

gmarkall:
The log shows the difference in configurations between expected and actual now:

```
-  'CUDA_LOW_OCCUPANCY_WARNINGS': 0,
?                                 ^

+  'CUDA_LOW_OCCUPANCY_WARNINGS': 1,
?                                 ^
```

gmarkall:
It seems as though the conditions creating this are (in order of occurrence):

1. A `CUDATestCase` class is setup up, setting `config.CUDA_LOW_OCCUPANCY_WARNINGS` to `0`
2. `TestConfig.test_config` runs
3. The teardown for the `CUDATestCase` happens (or it never happens)

I'm not immediately  sure why this is - I'd expected the tests to be run one class at a time but this expectation may not be correct.

numba/cuda/testing.py
```diff
@@ -23,6 +23,28 @@ class CUDATestCase(SerialMixin, TestCase):
...
+        # CUDATestCase.setUp (to disable low occupancy warnings) and its setUp
+        if cls is not CUDATestCase and cls.setUp is not CUDATestCase.setUp:
+            cls_setUp = cls.setUp
+
+            def setUp_both(self, *args, **kwargs):
+                # Call our setUp first so that low occupancy warnings are
+                # disabled before anything else happens
+                CUDATestCase.setUp(self)
+                return cls_setUp(self, *args, **kwargs)
+            cls.setUp = setUp_both
```
sklam, line 38: Why is this needed? 

If we need to guarantee that this `setUp` is called first, it's better for all the subclasses to have:

```python
def setUp(self):
    super().setUp()
    .... # local setup
```

testhound: @sklam is your suggestion that the setupClass method is not required and that a call to the base class setup is recommended as shown?

sklam: Yes, give that a try.  See if avoiding the `setUpClass` fixes the test failure.

numba/cuda/testing.py
```diff
@@ -31,8 +39,11 @@ class ContextResettingTestCase(CUDATestCase):
     to be internal implementation details (such as the state of allocations and
     deallocations, etc.).
     """
+    def setUp(self):
+        super().setUp()
```
gmarkall, line 43: I don't think this is necessary, because `ContextResettingTestCase.setUp()` isn't doing anything.

numba/cuda/tests/cudadrv/test_cuda_devicerecord.py
```diff
@@ -85,6 +86,7 @@ class TestCudaDeviceRecordWithRecord(TestCudaDeviceRecord):
     Tests the DeviceRecord class with np.record host types
     """
     def setUp(self):
+        super().setUp()
```
gmarkall, line 89: This is now calling `TestCudaDeviceRecord.setUp()`, which wasn't the case previously - the idea here is that `TestCudaDeviceRecordWithRecord` has a distinct `setUp` method. So I think this needs to be:

```suggestion
        CUDATestCase.setUp()
```

in order to bypass `TestCudaDeviceRecord.setUp()`.

gmarkall (changes requested):
Thanks for the changes. This passes locally for me with the CUDA target.

I see that the fix removes the use of `setUpClass` and "manually" invokes `setUp` where required in each derived class. I'm not sure I understand what the problem with `setUpClass` was though - is there some interleaving of the running of tests from different classes that caused the issue? (@sklam)?

There are a couple of small points on the diff - one is more "stylistic" and the other is where the behavior of a `setUp` appears to be changed.

gmarkall:
> `setUpClass` is probably running as soon as the test is loaded. Maybe it should be a `setUp` to surround just the test method.

If this is the case, it seems to be at odds with the documentation (https://docs.python.org/3/library/unittest.html#class-and-module-fixtures), which states:

> When the test suite encounters a test from a new class then tearDownClass() from the previous class (if there is one) is called, followed by setUpClass() from the new class.

Do you see another interpretation of what is happening?

gmarkall (approved):
Thanks for the updates - looks good to me!

@stuartarchibald Do you want another buildfarm run, or are you happy with this as-is?

gmarkall:
... Actually - if a buildfarm run can be squeezed in just to be on the safe side, that would be good - is that possible? (@stuartarchibald / @esc)

numba/cuda/tests/cudadrv/test_cuda_devicerecord.py
```diff
@@ -85,6 +86,7 @@ class TestCudaDeviceRecordWithRecord(TestCudaDeviceRecord):
     Tests the DeviceRecord class with np.record host types
     """
     def setUp(self):
+        CUDATestCase.setUp()
```
sklam, line 89: Missing `self`. Should be `super().setUp()`

sklam (changes requested):
Tests are failing because of the following:

numba/cuda/tests/cudadrv/test_cuda_devicerecord.py
gmarkall: This will call `TestCudaDeviceRecord.setUp()`, which is what I was trying to avoid with the suggestion to use `CUDATestCase.setUp()` instead.

sklam: In that case, it needs to be `CUDATestCase.setUp(self)`

sklam: Example of errors:

```pytb
ERROR: test_auto_device (numba.cuda.tests.cudadrv.test_cuda_devicerecord.TestCudaDeviceRecordWithRecord)
  File "/Users/siu/dev/numba/numba/cuda/tests/cudadrv/test_cuda_devicerecord.py", line 89, in setUp
    CUDATestCase.setUp()
TypeError: setUp() missing 1 required positional argument: 'self'

ERROR: test_device_record_copy (numba.cuda.tests.cudadrv.test_cuda_devicerecord.TestCudaDeviceRecordWithRecord)
  File "/Users/siu/dev/numba/numba/cuda/tests/cudadrv/test_cuda_devicerecord.py", line 89, in setUp
    CUDATestCase.setUp()
TypeError: setUp() missing 1 required positional argument: 'self'

ERROR: test_device_record_interface (numba.cuda.tests.cudadrv.test_cuda_devicerecord.TestCudaDeviceRecordWithRecord)
  File "/Users/siu/dev/numba/numba/cuda/tests/cudadrv/test_cuda_devicerecord.py", line 89, in setUp
    CUDATestCase.setUp()
TypeError: setUp() missing 1 required positional argument: 'self'

ERROR: test_from_record_like (numba.cuda.tests.cudadrv.test_cuda_devicerecord.TestCudaDeviceRecordWithRecord)
  File "/Users/siu/dev/numba/numba/cuda/tests/cudadrv/test_cuda_devicerecord.py", line 89, in setUp
    CUDATestCase.setUp()
TypeError: setUp() missing 1 required positional argument: 'self'

```

gmarkall: Apologies for missing `self` in my earlier suggestion - I thought I tested my suggestions locally, but I must have done something not quite right!

gmarkall (approved):
LGTM!

stuartarchibald (approved):
Thanks for the patch and fixes @testhound.

Thanks for reviewing @gmarkall.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_68`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_68`.

Passed

sklam (approved):

PR #6890, opened 2021-04-02
fixes #6884 

### LINKED ISSUE #6884: CUDA matrix multiplication example is probably wrong
(with thanks to @aktech)

In https://numba.pydata.org/numba-doc/latest/cuda/examples.html#matrix-multiplication

the code:

```python
    if x >= C.shape[0] and y >= C.shape[1]:
        # Quit if (x, y) is outside of valid C boundary
        return
```

should probably use `or` instead of `and`. This mistake may also exist in `test_matmul`.

Ideally this example would be converted to a documentation test as well.

rawwar:
fixes #6884 

I checked `test_matmul`  the boundary conditions check is correct. only the documentation has the typo

### DIFF
--- a/docs/source/cuda/examples.rst
+++ b/docs/source/cuda/examples.rst
@@ -52,7 +52,7 @@ memory::
         ty = cuda.threadIdx.y
         bpg = cuda.gridDim.x    # blocks per grid
 
-        if x >= C.shape[0] and y >= C.shape[1]:
+        if x >= C.shape[0] or y >= C.shape[1]:
             # Quit if (x, y) is outside of valid C boundary
             return
 


gmarkall (approved):
This does make an improvement by fixing something that is clearly an error the example, so this should be merged.

However, I think the example would be more helpful if it were actually executable - at the moment I don't know for certain that the example is correct even with this change. Converting it to a tested example like those in https://github.com/numba/numba/tree/master/numba/cuda/tests/doc_examples would be a further improvement.

gmarkall:
Forgot to say - thanks for the fix!

stuartarchibald (approved):
Thanks for the patch @rawwar , thanks for reviewing @gmarkall .

PR #6911, opened 2021-04-09
CUDA: Add support for activemask(), lanemask_lt(), and nanosleep()

### LINKED ISSUE #6906: CUDA: Support `__activemask`
The aim is to enable support for implementing warp-aggregated atomics, like the following code from [CUDA Pro Tip: Optimized Filtering with Warp-Aggregated Atomics](https://developer.nvidia.com/blog/cuda-pro-tip-optimized-filtering-warp-aggregated-atomics/):

```CUDA
__device__ int atomicAggInc(int *ctr) {
  unsigned int active = __activemask();
  int leader = __ffs(active) - 1;
  int change = __popc(active);
  unsigned int rank = __popc(active & __lanemask_lt());
  int warp_res;
  if(rank == 0)
    warp_res = atomicAdd(ctr, change);
  warp_res = __shfl_sync(active, warp_res, leader);
  return warp_res + rank;
}
```

This should be implemented as `numba.cuda.activemask()`. An equivalent for `__lanemask_lt()` may also be useful.

The `activemask` stub should compile to:

```assembly
activemask.b32 <reg>
```

gmarkall:
These are added to facilitate the implementation of warp-aggregated atomics (see e.g. [1]) and mutexes with exponential back-off (see e.g. [2]).

[1] https://developer.nvidia.com/blog/cuda-pro-tip-optimized-filtering-warp-aggregated-atomics/
[2] https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#scheduling-example

Fixes #6906.

### DIFF
--- a/numba/cuda/cudadecl.py
+++ b/numba/cuda/cudadecl.py
@@ -206,6 +206,18 @@ class Cuda_match_all_sync(ConcreteTemplate):
 
 
 @register
+class Cuda_activemask(ConcreteTemplate):
+    key = cuda.activemask
+    cases = [signature(types.int32)]
+
+
+@register
+class Cuda_lanemask_lt(ConcreteTemplate):
+    key = cuda.lanemask_lt
+    cases = [signature(types.int32)]
+
+
+@register
 class Cuda_popc(ConcreteTemplate):
     """
     Supported types from `llvm.popc`
@@ -369,6 +381,13 @@ class Cuda_atomic_compare_and_swap(AbstractTemplate):
 
         if dty in integer_numba_types and ary.ndim == 1:
             return signature(dty, ary, dty, dty)
+
+
+@register
+class Cuda_nanosleep(ConcreteTemplate):
+    key = cuda.nanosleep
+
+    cases = [signature(types.void, types.uint32)]
 
 
 @register_attr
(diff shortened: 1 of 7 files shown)


numba/cuda/cudadecl.py
```diff
@@ -205,6 +205,18 @@ class Cuda_match_all_sync(ConcreteTemplate):
     ]
 
 
+@register
+class Cuda_activemask(ConcreteTemplate):
+    key = cuda.activemask
+    cases = [signature(types.int32)]
```
sklam, line 211: should the return type be unsigned instead? Consider the situation where the mask is involved in bitshift operation. The current signed int32 type will use arithmetic shift on `>>`. It seems logical shift should be preferred for masks.

numba/cuda/cudadecl.py
```diff
@@ -205,6 +205,18 @@ class Cuda_match_all_sync(ConcreteTemplate):
...
+
+
+@register
+class Cuda_lanemask_lt(ConcreteTemplate):
+    key = cuda.lanemask_lt
+    cases = [signature(types.int32)]
```
sklam, line 217: Same as above on unsigned-ness.

sklam (commented):
There's one question about the signedness of the return type for the masks.

numba/cuda/cudadecl.py
gmarkall: I think it should be, yeah - this is a typo, I actually intended for it to be unsigned, so I'm glad of this comment!

gmarkall:
@sklam Thanks for the observations - I've now fixed the typing of `activemask` and `lanemask_lt`.

sklam (approved):
Thanks for the patch!

Pending buildfarm: `numba_smoketest_cuda_yaml_40`

Issue #6894, opened 2021-04-06
closure inlining not dealing with star-args correctly


stuartarchibald:
## Reporting a bug

- [x] I have tried using the latest released version of Numba (most recent is
 visible in the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [x] I have included a self contained code sample to reproduce the problem.
  i.e. it's possible to run as 'python bug.py'.

This:
```python
from numba import njit

@njit
def foo():
    def bar(a, b):
        pass
    x = 1, 2
    bar(*x)

foo()
```
does this:
```
numba/core/inline_closurecall.py", line 789, in _replace_args_with
    assert(idx < len(args))
AssertionError: Failed in nopython mode pipeline (step: inline calls to locally defined closures)
```

gmarkall:
I can reproduce the issue. Labelling as a bug because even if we decide that starargs is not supported for closures, the error message should be better.

stuartarchibald:
Since merge of https://github.com/numba/numba/pull/6912 the above reproducer now does:
```
numba.core.errors.UnsupportedError: Failed in nopython mode pipeline (step: inline calls to locally defined closures)
Calling a closure with *args is unsupported.

File "issue6894.py", line 8:
    def bar(a, b):
        <source elided>
    x = 1, 2
    bar(*x)
    ^
```
converting this to a feature request to support `*args` in closures.

PR #6912, opened 2021-04-09
Prevent use of varargs in closure calls.

stuartarchibald:
As title. This isn't supported by the closure inliner.

Makes #6894 a feature request not a bug.


sklam (approved):
Thanks for the patch!

PR #6913, opened 2021-04-09
Add runtests option to gitdiff on the common ancestor

sklam:
Example:

```python
python -m numba.runtests -g=ancestor
```

This allows us to test PR HEAD against its first splitting point from main branch. Useful for testing just the changes in the PR instead of including all new additions to the mainline.

### DIFF
--- a/numba/tests/test_runtests.py
+++ b/numba/tests/test_runtests.py
@@ -116,6 +116,23 @@ class TestCase(unittest.TestCase):
         # The tests must be equivalent
         self.assertEqual(sorted(full), sorted(sliced))
 
+    def test_gitdiff(self):
+        # Check for git
+        try:
+            subprocess.call("git", stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+        except FileNotFoundError as e:
+            self.skipTest("no git available")
+
+        # default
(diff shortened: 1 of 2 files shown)


numba/tests/test_runtests.py
```diff
@@ -116,6 +116,23 @@ def filter_test(xs):
...
+
+        # default
+        outs = self.get_testsuite_listing(['-g'])
+        self.assertNotIn("Git diff by common ancestor", outs)
+        # using ancestor
+        outs = self.get_testsuite_listing(['-g=ancestor'])
+        self.assertIn("Git diff by common ancestor", outs)
+        # misspelled ancestor
+        with self.assertRaises(subprocess.CalledProcessError):
+            self.get_testsuite_listing(['-g=ancest'])
```
stuartarchibald, line 134: This adds noise to the test output as the error streams aren't captured/redirected?

stuartarchibald (commented):
Thanks for the patch, this will be useful. One minor comment to resolve else looks good. I've also tested this manually.

numba/tests/test_runtests.py
sklam: ok, i'll sink it to devnull

stuartarchibald (approved):

Issue #6895, opened 2021-04-06
:= operator cannot be jitted by at least the cuda module


Alex-S-H-P:
## Reporting a bug

- [X] I have tried using the latest released version of Numba (most recent is
 visible in the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [X] I have included a self contained code sample to reproduce the problem.
  i.e. it's possible to run as 'python bug.py'.

gmarkall:
I can reproduce the message with the code given above. This also appears not to work with `@njit` or `@jit` - a slightly modified version follows:

```python
from numba import jit
import numpy as np


@jit
def bugged(array, i):
    #i = cuda.grid(1)
    if i < array.shape[0]:
        if (value_used_for_more_complex_computation := array[i]) < 4.:
            array[i] = 1.  # this should work


if __name__ == '__main__':
    array = np.zeros((10), int)
    try:
        bugged(array, 1)
    except Exception as E:
        print(E)
    print(array)
```

which produces the traceback:

```
$ python repro2.py 
repro2.py:5: NumbaWarning: 
Compilation is falling back to object mode WITH looplifting enabled because Function "bugged" failed type inference due to: Type of variable '$20binary_subscr.2' cannot be determined, operation: unknown operation, location: unknown location (0:0)

File "unknown location", line 0:
<source missing, REPL/exec in use?>

  @jit
Failed in object mode pipeline (step: object mode backend)
$20binary_subscr.2

File "repro2.py", line 9:
def bugged(array, i):
    <source elided>
    if i < array.shape[0]:
        if (value_used_for_more_complex_computation := array[i]) < 4.:
        ^

During: lowering "$28compare_op.5 = $20binary_subscr.2 < $const26.4" at repro2.py (9)
[0 0 0 0 0 0 0 0 0 0]
```

gmarkall:
IR dump of the function (when passed a list rather than an array - shouldn't make too much difference):

```
label 0:
    array = arg(0, name=array)               ['array']
    i = arg(1, name=i)                       ['i']
    $6load_attr.2 = getattr(value=array, attr=shape) ['$6load_attr.2', 'array']
    $const8.3 = const(int, 0)                ['$const8.3']
    $10binary_subscr.4 = getitem(value=$6load_attr.2, index=$const8.3, fn=<built-in function getitem>) ['$10binary_subscr.4', '$6load_attr.2', '$const8.3']
[...]

PR #6914, opened 2021-04-09
Fix issue #6895 - walrus operator

guilhermeleobas:
As title. This PR fix a bug (reported in #6895) whereas code using the walrus operator would fail in Numba 0.53:

```python
from numba import jit
import numpy as np


@jit
def bugged(array, i):
    #i = cuda.grid(1)
    if i < array.shape[0]:
        if (value_used_for_more_complex_computation := array[i]) < 4.:
            array[i] = 1.  # this should work


if __name__ == '__main__':
    array = np.zeros((10), int)
    try:
        bugged(array, 1)
    except Exception as E:
        print(E)
    print(array)
```

The interpreter will create a temporary variable `$20binary_subscr.2` for the expression `array[i]` that later will be assigned to `value`. When doing the copy propagation optimization and deleting the temporary variable `$20binary_subscr.2`, numba must replace its uses with `value`

### DIFF
--- a/numba/core/interpreter.py
+++ b/numba/core/interpreter.py
@@ -535,6 +535,24 @@ class Interpreter(object):
         self._inject_call(eh.exception_check, 'exception_check',
                           '$exception_check')
 
+
+    from typing import Dict
+    def replace_dead_vars(self, replaced_var: Dict[str, ir.Var], value):
+        if isinstance(value, ir.Var):
+            if value.name in replaced_var:
+                return replaced_var[value.name]
+            return value
+        elif isinstance(value, (list, tuple)):
+            l = [self.replace_dead_vars(replaced_var, e) for e in value]
+            return type(value)(l)
+        elif isinstance(value, dict):
+            for k, v in value.items():
+                value[k] = self.replace_dead_vars(replaced_var, v)
+        else:
+            return value
+        #     raise ValueError(type(value))
+
+
     def _remove_unused_temporaries(self):
         """
         Remove assignments to unused temporary variables from the
@@ -572,17 +590,8 @@ class Interpreter(object):
 
                 # iterate over RHS operands and replace them iff they
                 # are in replaced_var dict
-                if isinstance(inst.value, ir.Expr):
-                    for k, v in inst.value._kws.items():
-                        if isinstance(v, (tuple, list)):
-                            l = list(v)
-                            for i, e in enumerate(l):
-                                if e.name in replaced_var:
-                                    l[i] = replaced_var[e.name]
-                            inst.value._kws[k] = type(v)(l)
-                        elif isinstance(v, ir.Var):
-                            if v.name in replaced_var:
-                                inst.value._kws[k] = replaced_var[v.name]
+                if len(replaced_var) and isinstance(inst.value, ir.Expr):
+                    self.replace_dead_vars(replaced_var, inst.value._kws)
 
                 # eliminate temporary variables that are assigned to user
                 # variables right after creation. E.g.:
--- a/numba/core/interpreter.py
+++ b/numba/core/interpreter.py
@@ -572,12 +572,17 @@ class Interpreter(object):
 
                 # iterate over RHS operands and replace them iff they
                 # are in replaced_var dict
-                if (isinstance(inst.value, ir.Expr)
-                        and inst.value.op == "binop"):
-                    for k in ('lhs', 'rhs'):
-                        v = inst.value._kws.get(k)
-                        if v.name in replaced_var:
-                            inst.value._kws[k] = replaced_var[v.name]
+                if isinstance(inst.value, ir.Expr):
+                    for k, v in inst.value._kws.items():
+                        if isinstance(v, (tuple, list)):
+                            l = list(v)
+                            for i, e in enumerate(l):
+                                if e.name in replaced_var:
+                                    l[i] = replaced_var[e.name]
+                            inst.value._kws[k] = type(v)(l)
+                        elif isinstance(v, ir.Var):
+                            if v.name in replaced_var:
+                                inst.value._kws[k] = replaced_var[v.name]
 
                 # eliminate temporary variables that are assigned to user
                 # variables right after creation. E.g.:
--- a/numba/core/interpreter.py
+++ b/numba/core/interpreter.py
@@ -569,6 +569,16 @@ class Interpreter(object):
                     inst.value.value = replaced_var[inst.value.value.name]
                     new_body.append(inst)
                     continue
+
+                # iterate over RHS operands and replace them iff they
+                # are in replaced_var dict
+                if (isinstance(inst.value, ir.Expr)
+                        and inst.value.op == "binop"):
+                    for k in ('lhs', 'rhs'):
+                        v = inst.value._kws.get(k)
+                        if v.name in replaced_var:
+                            inst.value._kws[k] = replaced_var[v.name]
+
                 # eliminate temporary variables that are assigned to user
                 # variables right after creation. E.g.:
                 # $1 = f(); a = $1 -> a = f()
--- a/numba/core/interpreter.py
+++ b/numba/core/interpreter.py
@@ -535,9 +535,7 @@ class Interpreter(object):
         self._inject_call(eh.exception_check, 'exception_check',
                           '$exception_check')
 
-
-    from typing import Dict
-    def replace_dead_vars(self, replaced_var: Dict[str, ir.Var], value):
+    def replace_dead_vars(self, replaced_var, value):
         if isinstance(value, ir.Var):
             if value.name in replaced_var:
                 return replaced_var[value.name]
@@ -550,8 +548,6 @@ class Interpreter(object):
                 value[k] = self.replace_dead_vars(replaced_var, v)
         else:
             return value
-        #     raise ValueError(type(value))
-
 
     def _remove_unused_temporaries(self):
         """
--- a/numba/tests/support.py
+++ b/numba/tests/support.py
@@ -63,6 +63,10 @@ skip_py38_or_later = unittest.skipIf(
     utils.PYVERSION >= (3, 8),
     "unsupported on py3.8 or later"
 )
+skip_unless_py38_or_later = unittest.skipIf(
+    utils.PYVERSION < (3, 8),
+    "unsupported on py3.7 or earlier"
+)
 skip_tryexcept_unsupported = unittest.skipIf(
     utils.PYVERSION < (3, 7),
     "try-except unsupported on py3.6 or earlier"
@@ -107,6 +111,7 @@ skip_ppc64le_issue6465 = unittest.skipIf(platform.machine() == 'ppc64le',
 skip_unless_py37_or_later = lambda reason: \
     unittest.skipIf(utils.PYVERSION < (3, 7),
     reason)
+
 
 try:
     import scipy.linalg.cython_lapack
--- a/numba/tests/test_copy_propagate.py
+++ b/numba/tests/test_copy_propagate.py
@@ -155,6 +155,26 @@ class TestCopyPropagate(unittest.TestCase):
         self.assertEqual(impl1(5), njit(impl1)(5))
         self.assertEqual(impl2(np.ones(3), 0, 5), njit(impl2)(np.ones(3), 0, 5))
         self.assertEqual(impl3(C(), 5), jit(forceobj=True)(impl3)(C(), 5))
+    
+    def test_input_ir_replace_temporaries(self):
+        """make sure Interpreter._remove_unused_temporaries() replaces instructions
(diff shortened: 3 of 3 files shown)


guilhermeleobas:
I can't ask for a review from @ehsantn. So, I am pinging him here.

ehsantn:
@guilhermeleobas Unfortunately, this approach is not comprehensive since the walrus operator's output could be used in operations other than binary ops. See example below. I will submit a PR using an alternative fix soon.
```python
@numba.njit
def f():
    print(a:=3)
```

guilhermeleobas:
@ehsantn, I wasn't aware of other possible scenarios. Let me give another shot at this

ehsantn:
@guilhermeleobas go for it. Looks like the problem is with how the `DUP_TOP` op code is translated in Numba:
```python
In [4]: import dis

In [5]: def impl():
   ...:     print(a:=3)
   ...:

In [6]: dis.dis(impl)
  2           0 LOAD_GLOBAL              0 (print)
              2 LOAD_CONST               1 (3)
              4 DUP_TOP
              6 STORE_FAST               0 (a)
              8 CALL_FUNCTION            1
             10 POP_TOP
             12 LOAD_CONST               0 (None)
             14 RETURN_VALUE
```
IR before optimization:
```
    $2load_global.0 = global(print: <built-in function print>) ['$2load_global.0']
    $const4.1 = const(int, 3)                ['$const4.1']
    $6dup_top.2 = $const4.1                  ['$6dup_top.2', '$const4.1']
    a = $const4.1                            ['$const4.1', 'a']
    $10call_function.3 = call $2load_global.0($const4.1, func=$2load_global.0, args=[Var($const4.1, warlus.py:5)], kws=(), vararg=None) ['$10call_function.3', '$2load_global.0', '$const4.1']
    $const14.4 = const(NoneType, None)       ['$const14.4']
    $16return_value.5 = cast(value=$const14.4) ['$16return_value.5', '$const14.4']
    return $16return_value.5                 ['$16return_value.5']
```
IR after optimization
```
$2load_global.0 = global(print: <built-in function print>) ['$2load_global.0']
    a = const(int, 3)                        ['a']
    $10call_function.3 = call $2load_global.0($const4.1, func=$2load_global.0, args=[Var($const4.1, warlus.py:5)], kws=(), vararg=None) ['$10call_function.3', '$2load_global.0', '$const4.1']
    $const14.4 = const(NoneType, None)       ['$const14.4']
    $16return_value.5 = cast(value=$const14.4) ['$16return_value.5', '$const14.4']
    return $16return_value.5                 ['$16return_value.5']
```
The `$6dup_top.2` variable is replaced with `$const4.1` previously which causes this issue.

guilhermeleobas:
CI is all green. Can you take a look @ehsantn?

stuartarchibald:
@ehsantn as author of the original optimisation/patch, any chance you could review this change please? Many thanks.

@guilhermeleobas please could you resolve the conflicts against mainline, thanks!

guilhermeleobas:
> @guilhermeleobas please could you resolve the conflicts against mainline, thanks!

Done!

numba/core/interpreter.py
```diff
@@ -569,6 +583,12 @@ def _remove_unused_temporaries(self):
                     inst.value.value = replaced_var[inst.value.value.name]
                     new_body.append(inst)
                     continue
+
+                # iterate over RHS operands and replace them iff they
+                # are in replaced_var dict
+                if len(replaced_var) and isinstance(inst.value, ir.Expr):
```
ehsantn, line 589: May main concern is that this doesn't handle all nodes and may not be robust. I think we can probably come up with a more robust solution for the byte code pattern of walrus operator.

This may also slow down compilation. Has there been any compile time benchmarking for this?

guilhermeleobas:
Hi @ehsantn, thanks for taking the time of reviewing this PR.

I don't think this issue is strictly related to the walhus operator, but with the optimization approach. See the IR before and after calling `_remove_unused_temporaries` for the example given in the issue:

```python
# Before
label 0:
    array = arg(0, name=array)               ['array']
    i = arg(1, name=i)                       ['i']
    $i2.0 = i                                ['$i2.0', 'i']
    $array4.1 = array                        ['$array4.1', 'array']
    $6load_attr.2 = getattr(value=array, attr=shape) ['$6load_attr.2', 'array']
    $const8.3 = const(int, 0)                ['$const8.3']
    $10binary_subscr.4 = getitem(value=$6load_attr.2, index=$const8.3, fn=<built-in function getitem>) ['$10binary_subscr.4', '$6load_attr.2', '$const8.3']
    $12compare_op.5 = i < $10binary_subscr.4 ['$10binary_subscr.4', '$12compare_op.5', 'i']
    bool14 = global(bool: <class 'bool'>)    ['bool14']
    $14pred = call bool14($12compare_op.5, func=bool14, args=(Var($12compare_op.5, a.py:7),), kws=(), vararg=None, hardware=None) ['$12compare_op.5', '$14pred', 'bool14']
    branch $14pred, 16, 40                   ['$14pred']
label 16:
    $array16.0 = array                       ['$array16.0', 'array']
    $i18.1 = i                               ['$i18.1', 'i']
    $20binary_subscr.2 = getitem(value=array, index=i, fn=<built-in function getitem>) ['$20binary_subscr.2', 'array', 'i']
    $22dup_top.3 = $20binary_subscr.2        ['$20binary_subscr.2', '$22dup_top.3']
    value_used_for_more_complex_computation = $20binary_subscr.2 ['$20binary_subscr.2', 'value_used_for_more_complex_computation']
    $const26.4 = const(float, 4.0)           ['$const26.4']
    $28compare_op.5 = $20binary_subscr.2 < $const26.4 ['$20binary_subscr.2', '$28compare_op.5', '$const26.4']
    bool30 = global(bool: <class 'bool'>)    ['bool30']
    $30pred = call bool30($28compare_op.5, func=bool30, args=(Var($28compare_op.5, a.py:8),), kws=(), vararg=None, hardware=None) ['$28compare_op.5', '$30pred', 'bool30']
    branch $30pred, 32, 40                   ['$30pred']
label 32:
    $const32.0 = const(float, 1.0)           ['$const32.0']
    $array34.1 = array                       ['$array34.1', 'array']
    $i36.2 = i                               ['$i36.2', 'i']
    array[i] = $const32.0                    ['$const32.0', 'array', 'i']
    jump 40                                  []
label 40:
    $const40.0 = const(NoneType, None)       ['$const40.0']
    $42return_value.1 = cast(value=$const40.0) ['$42return_value.1', '$const40.0']
    return $42return_value.1                 ['$42return_value.1']
```

A few points for the code above:

- `$20binary_subscr.2` is the right-hand side of `value_used_for_more_complex_computation := value[i]`
- Numba marks it as a temporary variable
- `$20binary_subscr.2` is later assigned to `value_used_for_more_complex_computation`
- `$20binary_subscr.2` is used after the assignment on `$28compare_op.5 = $20binary_subscr.2 < $const26.4`

```python
# After
label 0:
    array = arg(0, name=array)               ['array']
    i = arg(1, name=i)                       ['i']
    $6load_attr.2 = getattr(value=array, attr=shape) ['$6load_attr.2', 'array']
    $const8.3 = const(int, 0)                ['$const8.3']
    $10binary_subscr.4 = getitem(value=$6load_attr.2, index=$const8.3, fn=<built-in function getitem>) ['$10binary_subscr.4', '$6load_attr.2', '$const8.3']
    $12compare_op.5 = i < $10binary_subscr.4 ['$10binary_subscr.4', '$12compare_op.5', 'i']
    bool14 = global(bool: <class 'bool'>)    ['bool14']
    $14pred = call bool14($12compare_op.5, func=bool14, args=(Var($12compare_op.5, a.py:7),), kws=(), vararg=None, hardware=None) ['$12compare_op.5', '$14pred', 'bool14']
    branch $14pred, 16, 40                   ['$14pred']
label 16:
    value_used_for_more_complex_computation = getitem(value=array, index=i, fn=<built-in function getitem>) ['array', 'i', 'value_used_for_more_complex_computation']
    $const26.4 = const(float, 4.0)           ['$const26.4']
    $28compare_op.5 = $20binary_subscr.2 < $const26.4 ['$20binary_subscr.2', '$28compare_op.5', '$const26.4']
    bool30 = global(bool: <class 'bool'>)    ['bool30']
    $30pred = call bool30($28compare_op.5, func=bool30, args=(Var($28compare_op.5, a.py:8),), kws=(), vararg=None, hardware=None) ['$28compare_op.5', '$30pred', 'bool30']
    branch $30pred, 32, 40                   ['$30pred']
label 32:
    $const32.0 = const(float, 1.0)           ['$const32.0']
    array[i] = $const32.0                    ['$const32.0', 'array', 'i']
    jump 40                                  []
label 40:
    $const40.0 = const(NoneType, None)       ['$const40.0']
    $42return_value.1 = cast(value=$const40.0) ['$42return_value.1', '$const40.0']
    return $42return_value.1                 ['$42return_value.1']

```

The code transformation replaces the assignment `$20binary_subscr.2 = getitem(...)` by `value_used_for_more_complex_computation = getitem(...)` but keeps the users of `$20binary_subscr.2`. This happens because the transformation wrongly assumes any temporary variable is not used elsewhere after the original assignment, which is not always true.

```python
# eliminate temporary variables that are assigned to user
# variables right after creation. E.g.:
# $1 = f(); a = $1 -> a = f()
# the temporary variable is not reused elsewhere since CPython
# bytecode is stack-based and this pattern corresponds to a pop
if (isinstance(inst.value, ir.Var) and inst.value.is_temp
        and new_body and isinstance(new_body[-1], ir.Assign)):
    prev_assign = new_body[-1]
    ...
```

I agree that the proposed solution degrades Numba's performance. Another approach would be to avoid replacing the temporary variable if it has users.

ehsantn:
The current optimization considers all possible byte code patterns for temporaries to make sure removing them is safe. The problem is that the pattern for walrus is different and more complicated currently. Do you have an example without walrus that doesn't work? If so, we may have to change the approach.

ehsantn:
@guilhermeleobas Could you try the patch below instead of your changes? It is more general and also simplifies the code.
In general, the `DUP_TOP` op-code creates the issue with this optimization since it duplicates the top of the stack. This patch should handle it properly.

```python
diff --git a/numba/core/interpreter.py b/numba/core/interpreter.py
index ebd8a47f4..b45df5e56 100644
--- a/numba/core/interpreter.py
+++ b/numba/core/interpreter.py
@@ -4,7 +4,7 @@ import dis
 import operator
 import logging

-from numba.core import errors, dataflow, controlflow, ir, config
+from numba.core import errors, dataflow, controlflow, ir, config, ir_utils
 from numba.core.errors import NotDefinedError, error_extras
 from numba.core.utils import (PYVERSION, BINOPS_TO_OPERATORS,
                               INPLACE_BINOPS_TO_OPERATORS,)
@@ -542,33 +542,12 @@ class Interpreter(object):
         """
         new_body = []
         replaced_var = {}
-        for inst in self.current_block.body:
-            # the same temporary is assigned to multiple variables in cases
-            # like a = b[i] = 1, so need to handle replaced temporaries in
-            # later setitem/setattr nodes
-            if (isinstance(inst, (ir.SetItem, ir.SetAttr))
-                    and inst.value.name in replaced_var):
-                inst.value = replaced_var[inst.value.name]
-            elif isinstance(inst, ir.Assign):
+        for i, inst in enumerate(self.current_block.body):
+            if isinstance(inst, ir.Assign):
                 if (inst.target.is_temp
                         and inst.target.name in self.assigner.unused_dests):
                     continue
-                # the same temporary is assigned to multiple variables in cases
-                # like a = b = 1, so need to handle replaced temporaries in
-                # later assignments
-                if (isinstance(inst.value, ir.Var)
-                        and inst.value.name in replaced_var):
-                    inst.value = replaced_var[inst.value.name]
-                    new_body.append(inst)
-                    continue
-                # chained unpack cases may reuse temporary
-                # e.g. a = (b, c) = (x, y)
-                if (isinstance(inst.value, ir.Expr)
-                        and inst.value.op == "exhaust_iter"
-                        and inst.value.value.name in replaced_var):
-                    inst.value.value = replaced_var[inst.value.value.name]
-                    new_body.append(inst)
-                    continue
+
                 # eliminate temporary variables that are assigned to user
                 # variables right after creation. E.g.:
                 # $1 = f(); a = $1 -> a = f()
@@ -590,6 +569,12 @@ class Interpreter(object):
                         self.definitions[inst.target.name].extend(
                             self.definitions.pop(inst.value.name)
                         )
+                        # replace temporary in the next statement as well since DUP_TOP
+                        # op-code may cause reuse of variable
+                        # examples: a = b = 1, a = b[i] = 1, a = b.c = 1
+                        # a = (b, c) = (x, y), print(a:=3)
+                        if i != len(self.current_block.body) - 1:
+                            ir_utils.visit_vars_stmt(self.current_block.body[i+1], lambda var, _: inst.target if var.name == inst.value.name else var, None)
                         continue

             new_body.append(inst)
```

Issue #6308, opened 2020-10-06
pytest doctest can't discover functions with @guvectorize decorator


guilhermeleobas:
Example:

```python
from numba import guvectorize, int64, njit

@guvectorize([(int64[:], int64, int64[:])], '(n),()->(n)')
def bar(x, y, res):
    """
    Examples
    --------
    >>> x = 1
    """
    for i in range(x.shape[0]):
            res[i] = x[i] + y

@njit
def foo():
    """
    Examples
    --------
    >>> x = 1
    """
    pass
```

```bash
$ pytest -rs -sv -v --doctest-modules a.py
=============================================== test session starts ===============================================
platform darwin -- Python 3.8.3, pytest-6.1.0, py-1.9.0, pluggy-0.13.1 -- /Users/guilhermeleobas/miniconda3/envs/numba/bin/python
cachedir: .pytest_cache
rootdir: /Users/guilhermeleobas/Documents/GitHub/numba
collected 1 item

a.py::a.foo PASSED

================================================ 1 passed in 1.56s ================================================
```

One would pytest to execute the docstring from `foo` and `bar` but only `foo` is discovered.

sklam:
Need to check if it is a problem that numba is not passing `__doc__` of `bar` to the numpy gufunc or that numpy gufunc is somehow preventing pytest to work for other reason.

PR #6915, opened 2021-04-12
Update _Intrinsic for sphinx to capture the inner docstring

guilhermeleobas:
As title.

This also fixes https://github.com/numba/numba/issues/6308

### DIFF
--- a/numba/tests/test_extending.py
+++ b/numba/tests/test_extending.py
@@ -1374,6 +1374,24 @@ class TestIntrinsic(TestCase):
         second = pickle.loads(pickled)
         self.assertIs(rebuilt._defn, second._defn)
 
+    def test_docstring(self):
+
+        @intrinsic
+        def void_func(typingctx, a: int):
+            """void_func docstring"""
+            sig = types.void(types.int32)
+
+            def codegen(context, builder, signature, args):
+                pass  # do nothing, return None, should be turned into
+                # dummy value
+
+            return sig, codegen
+        
+        self.assertEqual("numba.tests.test_extending", void_func.__module__)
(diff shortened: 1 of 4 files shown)


numba/tests/test_extending.py
```diff
@@ -1374,6 +1374,26 @@ def codegen(context, builder, signature, args):
...
+                pass  # do nothing, return None, should be turned into
+                # dummy value
+
+            return sig, codegen
+
+        self.assertEqual("numba.tests.test_extending", void_func.__module__)
+        self.assertEqual("void_func", void_func.__name__)
+        self.assertEqual("TestIntrinsic.test_docstring.<locals>.void_func",
+                         void_func.__qualname__)
+        self.assertDictEqual({'a': int}, void_func.__annotations__)
```
sklam, line 1394: I wasn't aware these were added. I'll need to try in my next PR.

sklam (approved):
Thanks for the patch!

I have manually tested against the doctest case in #6308.

numba/tests/test_extending.py
guilhermeleobas: `functools.update_wrapper` updates the following fields: `__module__`, `__name__`, `__qualname__`, `__annotations__` and `__doc__` ([ref](https://docs.python.org/3/library/functools.html#functools.update_wrapper))

sklam: oops, I mean `assertDictEqual`, not anything particular in your PR.  (I was rushing to finish up the review before starting a meeting and didn't realize the confusing comment.)

guilhermeleobas: Oh, ok 😄

PR #6917, opened 2021-04-12
Add type conversion for StringLiteral to unicode_type and test.

### LINKED ISSUE #6907: Missing conversion from StringLiteral to unicode_type
## Reporting a bug

This:
```python
from numba import typeof, types
from numba.core.typing import Context

x = types.literal('string literal')
print(x, type(x))
print("conversion:", x.can_convert_to(Context(), types.unicode_type))
```
does this:
```
Literal[str](string literal) <class 'numba.core.types.misc.StringLiteral'>
conversion: None
```
I think it should be possible to convert a `StringLiteral` to a `unicode_type` but not the converse.

A `can_convert_to()` method needs adding to this class:
https://github.com/numba/numba/blob/3461bcb76a479bffbadbef24b6662107229e2a8a/numba/core/types/misc.py#L40-L41

xref: https://numba.discourse.group/t/how-to-pass-string-argument-to-numba-njit-python-function-from-inside-another-whilst-specifying-signature-type/663 which hits this bug.

stuartarchibald:
As title.

Fixes #6907

### DIFF
--- a/numba/core/types/misc.py
+++ b/numba/core/types/misc.py
@@ -38,7 +38,10 @@ class RawPointer(Opaque):
 
 
 class StringLiteral(Literal, Dummy):
-    pass
+
+    def can_convert_to(self, typingctx, other):
+        if isinstance(other, UnicodeType):
+            return Conversion.safe
 
 
 Literal.ctor_map[str] = StringLiteral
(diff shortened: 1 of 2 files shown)


esc (approved):
I tested this locally an the code looks good and fixes the issue.

esc:
Thanks for the patch!

PR #6918, opened 2021-04-12
Start section on commonly encounted unsupported parfors code.

stuartarchibald:
As title. This covers a couple of commonly reported problems.

### DIFF
--- a/docs/source/user/parallel.rst
+++ b/docs/source/user/parallel.rst
@@ -230,8 +230,10 @@ Unsupported Operations
 This section contains a non-exhaustive list of commonly encountered but
 currently unsupported features:
 
-#. Concurrent write operations on container types (i.e. lists, sets and
-   dictionaries) in a ``prange`` parallel region, e.g.::
+#. **Mutating a list is not threadsafe**
+
+   Concurrent write operations on container types (i.e. lists, sets and
+   dictionaries) in a ``prange`` parallel region are not threadsafe e.g.::
 
     @njit(parallel=True)
     def invalid():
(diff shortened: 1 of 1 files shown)


docs/source/user/parallel.rst
```diff
@@ -224,6 +224,41 @@ give an equivalence parallel implementation using :func:`~numba.guvectorize`,
 it would require a pervasive change that rewrites the code to extract kernel
 computation that can be parallelized, which was both tedious and challenging.
 
+Unsupported Operations
+======================
+
+This section contains a non-exhaustive list of commonly encountered but
+currently unsupported features:
+
+#. Concurrent write operations on container types (i.e. lists, sets and
```
sklam, line 233: Should we give it a title for quick look up? e.g. "List operations are not thread safe."

docs/source/user/parallel.rst
```diff
@@ -224,6 +224,41 @@ give an equivalence parallel implementation using :func:`~numba.guvectorize`,
...
+        z = []
+        for i in prange(10000):
+            z.append(i)
+        return z
+
+   It is highly likely that the above will result in corruption or an access
+   violation as containers require thread-safety under mutation but this feature
+   is not implemented.
+
+#. The use of the induction variable induced by a ``prange`` based loop in
```
sklam, line 247: maybe title: "Induction variable is not associated with thread id."

sklam (commented):
Thanks for the patch! I think adding a title sentence for each example can allow users to read it faster.

docs/source/user/parallel.rst
stuartarchibald: Good idea, added similar titles in 57b0028

sklam (approved):
LGTM

stuartarchibald:
CI passed. Thanks for the review @sklam.

PR #6924, opened 2021-04-12
CUDA: Fix `ffs`

### LINKED ISSUE #6919: CUDA: ffs is off-by-1
The following Python:

```python
from numba import cuda
import numpy as np


@cuda.jit
def use_ffs(r, x):
    r[0] = cuda.ffs(x[0])


x = np.zeros(1, dtype=np.uint32)
r = np.zeros(1, dtype=np.uint32)

for i in (1, 2, 4, 8, 16):
    x[0] = i
    use_ffs[1, 1](r, x)
    print(r[0])
```

produces

```
$ python repro.py 
0
1
2
3
4
```

whereas the following CUDA C++:

```CUDA
#include <cstdint>
#include <cmath>
#include <iostream>

__global__ void use_ffs(uint32_t *r, uint32_t *x) {
  r[0] = __ffs(x[0]);
}

int main(void) {
  uint32_t *r, *x;

  cudaMallocManaged(&r, sizeof(uint32_t));
  cudaMallocManaged(&x, sizeof(uint32_t));

  for (size_t i = 0; i < 5; i++) {
    x[0] = pow(2, i);
    use_ffs<<<1, 1>>>(r, x);
    cudaDeviceSynchronize();
    std::cout << r[0] << std::endl;
  }

  return 0;
}
```

when compiled with:

```bash
nvcc -gencode arch=compute_75,code=sm_75 test.cu -o test
```

produces:

```
$ ./test
1
2
3
4
5
```

The Python example deviates from [the description in the CUDA Programming Guide](https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH__INTRINSIC__INT.html#group__CUDA__MATH__INTRINSIC__INT_1gaf1eb22243e29e0b7222adee8ae7d4e4), which states:

> Find the position of the first (least significant) bit set to 1 in x, where the least significant bit position is 1.

gmarkall:
The `ffs` intrinsic produced results that were off-by-1 compared to `__ffs()` in CUDA C (and in general compared with `ffs()` implementations in C libraries). This commit fixes the implementation by making use of the libdevice implementations `__nv_ffs` and `__nv_ffsll` for 32- and 64-bit operands respectively. Since these always return a 32-bit integer between 0 and 32 (0 for no bit set), the typing of the function is modified to always return a `uint32`.

Tests are updated accordingly.

The documentation is updated to clarify the index of the least significant bit, and the result of calling `ffs(0)`. Documentation for
other bit manipulation instructions is tidied up and made consistent with the documentation for `ffs`.

Fixes #6919.

### DIFF
--- a/numba/cuda/tests/cudapy/test_intrinsics.py
+++ b/numba/cuda/tests/cudapy/test_intrinsics.py
@@ -363,12 +363,16 @@ class TestCudaIntrinsic(CUDATestCase):
         ary = np.zeros(1, dtype=np.int32)
         compiled[1, 1](ary, 0x00100000)
         self.assertEquals(ary[0], 21)
+        compiled[1, 1](ary, 0x80000000)
+        self.assertEquals(ary[0], 32)
 
     def test_ffs_u4(self):
         compiled = cuda.jit("void(int32[:], uint32)")(simple_ffs)
         ary = np.zeros(1, dtype=np.int32)
         compiled[1, 1](ary, 0x00100000)
         self.assertEquals(ary[0], 21)
+        compiled[1, 1](ary, 0x80000000)
+        self.assertEquals(ary[0], 32)
 
     def test_ffs_i4_1s(self):
         compiled = cuda.jit("void(int32[:], int32)")(simple_ffs)
@@ -388,6 +392,8 @@ class TestCudaIntrinsic(CUDATestCase):
         ary = np.zeros(1, dtype=np.int32)
         compiled[1, 1](ary, 0x000000000010000)
         self.assertEquals(ary[0], 17)
+        compiled[1, 1](ary, 0x100000000)
+        self.assertEquals(ary[0], 33)
 
     def test_simple_laneid(self):
         compiled = cuda.jit("void(int32[:])")(simple_laneid)
--- a/numba/cuda/tests/cudapy/test_intrinsics.py
+++ b/numba/cuda/tests/cudapy/test_intrinsics.py
@@ -362,32 +362,32 @@ class TestCudaIntrinsic(CUDATestCase):
         compiled = cuda.jit("void(int32[:], int32)")(simple_ffs)
         ary = np.zeros(1, dtype=np.int32)
         compiled[1, 1](ary, 0x00100000)
-        self.assertEquals(ary[0], 20)
+        self.assertEquals(ary[0], 21)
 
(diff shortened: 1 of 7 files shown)


gmarkall:
I've also fixed the implementation in the simulator. Whilst doing so, I noticed that the simulator implementation only works for 32-bit values (the test with a 64-bit value is skipped in the simulator), so I've appended a note about this to the simulator documentation.

numba/cuda/tests/cudapy/test_intrinsics.py
```diff
@@ -362,32 +362,32 @@ def test_ffs_i4(self):
...
-        self.assertEquals(ary[0], 32, "CUDA semantics")
+        self.assertEquals(ary[0], 0)
 
     @skip_on_cudasim('only get given a Python "int", assumes 32 bits')
     def test_ffs_i8(self):
         compiled = cuda.jit("void(int32[:], int64)")(simple_ffs)
         ary = np.zeros(1, dtype=np.int32)
         compiled[1, 1](ary, 0x000000000010000)
-        self.assertEquals(ary[0], 16)
+        self.assertEquals(ary[0], 17)
```
stuartarchibald, line 390: Maybe add a test for the wrap around e.g. `1<<32` and `1<<33`?

stuartarchibald (commented):
Thanks for the patch, one minor comment RE testing else looks good.

numba/cuda/tests/cudapy/test_intrinsics.py
gmarkall: I'm not sure what you mean by wrap around, but I've added tests with the 32nd and 33rd bits set.

stuartarchibald: Apologies, I'm referring to the modular arithmetic expr `<position> % 33` in the simulator impl. Thanks for adding the tests.

gmarkall: Ah, I think that gets hit when it's all zeroes.

stuartarchibald (approved):

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_47`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_47`.

Passed.

PR #6929, opened 2021-04-14
Fix CI failure when gitpython is missing.

sklam:
Skip test when `import git` fails indicating that gitpython package is not available thus we cannot run gitdiff


stuartarchibald (approved):
Thanks for the patch.

stuartarchibald:
I tested this manually locally, works as expected.

PR #6939, opened 2021-04-17
Implement `slice` boxing

### LINKED ISSUE #6931: Boxing for `slice`
The native `slice` object doesn't have a `box` implementation:

```python
import numba


@numba.njit
def get_slice(a, b):
    return slice(a, b)
```
```python
>>> get_slice(1, 2)
TypeError: cannot convert native slice<a:b> to Python object
```

## Feature request

A `box` implementation for `slice`.

brandonwillard:
This PR implements a `box` function for `SliceType`.  Closes #6931.

### DIFF
--- a/numba/core/boxing.py
+++ b/numba/core/boxing.py
@@ -363,6 +363,20 @@ def unbox_optional(typ, obj, c):
                        cleanup=cleanup)
 
 
+@box(types.SliceType)
+def box_slice(typ, val, c):
+    start = c.box(types.int64, c.builder.extract_value(val, 0))
+    stop = c.box(types.int64, c.builder.extract_value(val, 1))
+    # if typ.has_step:
+    step = c.box(types.int64, c.builder.extract_value(val, 2))
+    # else:
+    #     step = c.pyapi.get_null_object()
+
+    slice_val = c.pyapi.slice_new(start, stop, step)
+
+    return slice_val
+
+
 @unbox(types.SliceType)
 def unbox_slice(typ, obj, c):
     """
--- a/numba/core/pythonapi.py
+++ b/numba/core/pythonapi.py
@@ -665,6 +665,11 @@ class PythonAPI(object):
     # Concrete slice API
     #
 
+    def slice_new(self, start, stop, step):
+        fnty = Type.function(self.pyobj, [self.pyobj, self.pyobj, self.pyobj])
+        fn = self._get_function(fnty, name="PySlice_New")
+        return self.builder.call(fn, [start, stop, step])
+
     def slice_as_ints(self, obj):
         """
         Read the members of a slice of integers.
--- a/numba/tests/test_slices.py
+++ b/numba/tests/test_slices.py
@@ -16,6 +16,9 @@ def slice_passing(sl):
     return sl.start, sl.stop, sl.step
 
 def slice_constructor(*args):
+    return slice(*args)
+
+def slice_constructor_tuple(*args):
     sl = slice(*args)
     return sl.start, sl.stop, sl.step
 
@@ -72,6 +75,7 @@ class TestSlices(MemoryLeakMixin, TestCase):
         maxnegint = -maxposint - 1
         a = np.arange(10)
         cfunc = jit(nopython=True)(slice_constructor)
+        cfunc_tuple = jit(nopython=True)(slice_constructor_tuple)
         cfunc_use = jit(nopython=True)(slice_construct_and_use)
         for args, expected in [
             ((None,), (0, maxposint, 1)),
@@ -89,11 +93,13 @@ class TestSlices(MemoryLeakMixin, TestCase):
             ((None, 5, -1), (maxposint, 5, -1)),
             ((10, 5, -1), (10, 5, -1)),
         ]:
-            got = cfunc(*args)
+            got = cfunc_tuple(*args)
             self.assertPreciseEqual(got, expected)
             usage = slice_construct_and_use(args, a)
             cusage = cfunc_use(args, a)
             self.assertPreciseEqual(usage, cusage)
+            box = cfunc(*args)
+            self.assertPreciseEqual((box.start, box.stop, box.step), expected)
 
     def test_slice_constructor_cases(self):
         """


brandonwillard:
@guilhermeleobas, sorry, I hadn't noticed you put in a PR for this.  It looks like the fix is pretty straightforward, so we have more or less the exact same thing&mdash;except for your use of `c.context.make_helper`.

Is there something about the IR created when `c.context.make_helper` is used that is preferrable?  If so, I can update this PR to do that instead.

guilhermeleobas:
No problem @brandonwillard.

> Is there something about the IR created when c.context.make_helper is used that is preferrable? If so, I can update this PR to do that instead (or you can re-open yours and I'll close this one).

To be honest I don't know. I guess `c.context.make_helper` just makes the code a bit more easy to read.

guilhermeleobas:
Hi @brandonwillard, your code fails with the following test case:

```python
from numba import njit

@njit
def foo():
    return slice(None, None, -1)

print(foo())
# slice(9223372036854775807, -9223372036854775808, -1)
```

While creating a slice, Numba converts `None` values to special integers and the roundtrip `box(unbox(slice)) == slice` should be an identity function. 

See: https://github.com/numba/numba/blob/12355c16ef70d10b0812c4266a3e143bd135a7f1/numba/_helperlib.c#L755-L783

My understanding of this code is 
* Step defaults to `1`
* Start defaults to `0 if step > 0 else sys.maxsize`
* Stop defaults to `sys.maxsize if step > 0 else (-sys.maxsize - 1)`

When boxing a slice, one must take those values into account to do the correct conversion.

brandonwillard:
Yeah, you can see [some commented code](https://github.com/numba/numba/pull/6939/files#diff-e18497988aefc7989947a3978df271770e5ca2aa8d66afea167edb709a9f2c31R370) where I was looking into a better conversion, but, after noticing the use of `sys.maxsize`, it was apparent that the conversion would be "lossy" without some extra information (e.g. anything that says `sys.maxsize` was originally `None`).

Unless we want to make some strong assumptions and say&mdash;for example&mdash;that values equal to `sys.maxsize` should be converted to `None`, there's really no way to get exact conversions without this extra information.  Even so, such a conversion would be wrong in cases where `sys.maxsize` actually _was_ the original value.

Regardless, the existing tests seem to establish this "lossy" conversion via the tested tuples of start, stop, and step values, so the boxing would appear to be correct under that precedent, as well.

guilhermeleobas:
> Unless we want to make some strong assumptions and say—for example—that values equal to sys.maxsize should be converted to None, there's really no way to get exact conversions without this extra information. Even so, such a conversion would be wrong in cases where sys.maxsize actually was the original value.

I guess it is safe to assume this given Numba already does the mapping when unboxing a slice.

brandonwillard:
> I guess it is safe to assume this given Numba already does the mapping when unboxing a slice.

If you're saying that it's safe to assume `sys.maxsize` is equivalent to `None`, then we would also need to change the existing tests that return `sl.start, sl.stop, sl.step`, because all the boxing is doing is putting those into a `slice` object and returning that.

guilhermeleobas:
Let's wait for one of the core devs to review your PR. Maybe there's a reason for using `sys.maxsize` to represent `None` values.

brandonwillard:
> Maybe there's a reason for using `sys.maxsize` to represent `None` values.

I can understand the potential reason(s) for that (e.g. not wanting to deal with two possible types in a lower-level context).  

I think this just comes down to a design decision: i.e. should the boxing/unboxing mappings be one-to-one for `slice` (and/or all objects)?

That question does seems like something much larger than this specific problem, though, especially since there's already a precedent set for the answer "no".

sklam:
Thanks @brandonwillard and @guilhermeleobas for the discussion. I suggested the `box(unbox(slice)) == slice` property in an OOB discussion to @guilhermeleobas. But the discussion here is causing me to give a second thought. This problem of whether we should treat `sys.maxsize` as `None` in `slice` is getting very interesting.

Here's gathering some interesting points:
- CPython is already making assumption about `sys.maxsize` (or `PY_SSIZE_MAX` in C code): https://github.com/python/cpython/blob/68ba0c67cac10c2545ea3b62d8b161e5b3594edd/Objects/sliceobject.c#L203-L245. In short, bigint get clamped to PY_SSIZE_MAX. 
- In Python, ``sys.maxsize`` is the maximum container size. Would that mean `sys.maxsize` is already equivalent to `None` for `slice` type?
- Operating near ``sys.maxsize`` should be extremely rare if it is even possible; (at the limit of the addressable memory)
- > should the boxing/unboxing mappings be one-to-one for slice (and/or all objects)?... there's already a precedent set for the answer "no". 

  Correct, numba do not always map 1-to-1 with python.

I probably should bring this up for discussion in tomorrow's Numba public meeting.

Like most things in numba, ["this is harder to get right than you might think"](https://github.com/python/cpython/blob/68ba0c67cac10c2545ea3b62d8b161e5b3594edd/Objects/sliceobject.c#L207) 😅

stuartarchibald:
Recollection the outcome of the discussion on this last week at the public meeting (please do correct as needed)...

1. Numba doesn't support `BigInt` which restricts the domain to essentially that addressable by an `int64`, this already limits what can be done with containers and slices.
2. Operating with containers and slices near the maximum addressable size is unlikely/unusual in CPython and incredibly unlikely in Numba as other overflow related problems are likely to occur ahead of this being the main issue.

It was determined that `sys.maxsize` can alias for `None` as a practical solution until proven otherwise, and it is probably the best that can be done with the current type system.

brandonwillard:
It would seems like such a change would be concomitant with a change in the returned values of each slice property/field independently, no?

> It was determined that `sys.maxsize` can alias for `None` as a practical solution until proven otherwise, and it is probably the best that can be done with the current type system.

Would such a change would be concomitant with a change in the returned values of each slice property/field independently?  For example, should an `njit`ed function that returns/`box`es `slice.stop` also return `None` when it's value is `sys.maxsize`?  

Also, what about `slice.step`?  An initial value of `None` appears to be automatically set to `1`.  How would we maintain parity with Python's `slice` in that case?

stuartarchibald:
> > It was determined that `sys.maxsize` can alias for `None` as a practical solution until proven otherwise, and it is probably the best that can be done with the current type system.
> 
> Would such a change be concomitant with a change in the returned values of each slice property/field independently? For example, should an `njit`ed function that returns/`box`es `slice.stop` also return `None` when its value is `sys.maxsize`?
> 
> Also, what about `slice.step`? An initial value of `None` appears to be automatically set to `1`. How would we maintain parity with Python's `slice` in that case, or do we even want/need to handle this case?

This is a good point that was not discussed directly. One discussed solution to the whole issue that I think would also cover the alias of `None` and `1` on `.step` is to carry some `is_none` information on each member in the data structure representing a slice. This will incur a penalty in terms of complexity (which was why it was initially rejected) but would potentially make it possible to "do the right thing" in all these cases without assumption.

brandonwillard:
> One discussed solution to the whole issue that I think would also cover the alias of `None` and `1` on `.step` is to carry some `is_none` information on each member in the data structure representing a slice.

I can add that, if everyone is open to it now.

njriasan:
@brandonwillard If this PR development isn't active can I help it get to the finish line?

brandonwillard:
> @brandonwillard If this PR development isn't active can I help it get to the finish line?

This is waiting on a consensus/decision about how to proceed.  You can inform the Numba team of your interest in having this functionality.

PR #6940, opened 2021-04-19
Fix function resolution for intrinsics across hardware.

stuartarchibald:
As title.

### DIFF
--- a/numba/core/typing/templates.py
+++ b/numba/core/typing/templates.py
@@ -885,12 +885,21 @@ class _IntrinsicTemplate(AbstractTemplate):
         """
         Type the intrinsic by the arguments.
         """
-        from numba.core.extending_hardware import resolve_dispatcher_from_str
+        from numba.core.extending_hardware import (get_local_target,
+                                                   resolve_target_str,
+                                                   dispatcher_registry)
         from numba.core.imputils import builtin_registry
 
         cache_key = self.context, args, tuple(kws.items())
-        hwstr = self.metadata.get('hardware', 'cpu')
-        disp = resolve_dispatcher_from_str(hwstr)
+        hwstr = self.metadata.get('hardware', 'generic')
+        # Get the class for the target declared by the function
+        hw_clazz = resolve_target_str(hwstr)
+        # get the local target
+        target_hw = get_local_target(self.context)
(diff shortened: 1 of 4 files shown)


numba/core/typing/templates.py
```diff
@@ -885,12 +885,21 @@ def generic(self, args, kws):
...
         cache_key = self.context, args, tuple(kws.items())
-        hwstr = self.metadata.get('hardware', 'cpu')
-        disp = resolve_dispatcher_from_str(hwstr)
+        hwstr = self.metadata.get('hardware', 'generic')
+        # Get the class for the target declared by the function
+        hw_clazz = resolve_target_str(hwstr)
+        # get the local target
+        target_hw = get_local_target(self.context)
+        # make sure the target_hw is in the MRO for hw_clazz else bail
+        if hw_clazz not in target_hw.__mro__:
```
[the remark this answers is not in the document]
stuartarchibald: thanks, patched in 54438a0

sklam (commented):
Just one suggestion

numba/core/typing/templates.py
```diff
@@ -885,12 +885,21 @@ def generic(self, args, kws):
...
-        hwstr = self.metadata.get('hardware', 'cpu')
-        disp = resolve_dispatcher_from_str(hwstr)
+        hwstr = self.metadata.get('hardware', 'generic')
+        # Get the class for the target declared by the function
+        hw_clazz = resolve_target_str(hwstr)
+        # get the local target
+        target_hw = get_local_target(self.context)
+        # make sure the target_hw is in the MRO for hw_clazz else bail
+        if not target_hw.inherits_from(hw_clazz):
+            return None
```
sklam, line 901: This is never exercised. It seems this condition will never trigger because the intrinsic will not be found in the wrong hardware. So, maybe this should trigger a ICE instead.

stuartarchibald: I think this is true. The `BaseFunction.get_call_type` would be the way to resolve the intrinsic and it would filter out intrinsics that are invalid for the hardware target. Will make this an ICE.

stuartarchibald: Patch is in 4e8786c

sklam (approved):

PR #6941, opened 2021-04-19
ABC the target descriptor and make consistent throughout.

stuartarchibald:
As title. Needed to enforce the interface for `@intrinsic` template
`generic` methods.

### DIFF
--- a/numba/core/descriptors.py
+++ b/numba/core/descriptors.py
@@ -2,8 +2,20 @@
 Target Descriptors
 """
 
+from abc import ABCMeta, abstractmethod
 
-class TargetDescriptor(object):
+
+class TargetDescriptor(metaclass=ABCMeta):
 
     def __init__(self, target_name):
         self._target_name = target_name
+
+    @property
+    @abstractmethod
+    def typing_context(self):
+        ...
+
+    @property
+    @abstractmethod
+    def target_context(self):
+        ...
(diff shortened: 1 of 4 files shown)


gmarkall (approved):
This passes tests locally for me with hardware and the changes seem to make sense.

We may still encounter issues with the typing and target contexts existing per-dispatcher in the CUDA target (which is presently needed in CUDA because things like the fast math flag are attached to the context) but being class members in the CPU target (as I understood from an earlier OOB conversation with @stuartarchibald ), but this is a move in the right direction.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_48`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_48`.

This passed.

stuartarchibald:
> This passes tests locally for me with hardware and the changes seem to make sense.

> We may still encounter issues with the typing and target contexts existing per-dispatcher in the CUDA target (which is presently needed in CUDA because things like the fast math flag are attached to the context) but being class members in the CPU target (as I understood from an earlier OOB conversation with @stuartarchibald ), but this is a move in the right direction.

Thanks for noting this, definitely something to investigate as part of the next refactor.

sklam (approved):
LGTM

PR #6944, opened 2021-04-19
CUDA: Support for `@overload`

gmarkall:
Joint work with @stuartarchibald.

The implementation consists of:

- Addition of `get_call_template`, `nopython_signatures`, and   `get_overload` to the class that is now `DeviceDispatcher`. These are needed by the overload infrastructure, and are inspired by the equivalents in `_DispatcherBase`.
- Registration of a CUDA `jit` decorator. The registration with `dispatcher_register` is removed and replaced with this new registration, because it did nothing (it was for the `target` kwarg to the `numba.jit` decorator, which is deprecated). The `jit` decorator for overloads is a bit special in that it always adds the `device=True` kwarg to the regular `jit` decorator, because overloads will always be device functions.
- Addition of a `no_cpython_wrapper` kwarg to `jitdevice`, which is needed because the overload infrastructure passes it. No action is required regardless of its value, because there is no CPython wrapper in the CUDA target.

A refactoring is also made: `DeviceFunctionTemplate` is renamed to `DeviceDispatcher`, as it reflects more closely what it is - it is
analagous to the `Dispatcher` class used for kernels, and the `DeviceFunction` class is more analagous to the `_Kernel` class. Eventually `Dispatcher` will subsume these classes as we move towards a more "modern" dispatcher implementation in the CUDA target.

Tests are added; these are all skipped on the simulator because overloading doesn't really exist as a concept without compilation.

### DIFF
--- a/numba/cuda/tests/cudapy/test_overload.py
+++ b/numba/cuda/tests/cudapy/test_overload.py
@@ -1,0 +1,345 @@
+from numba import cuda
+from numba.core.extending import overload
+from numba.cuda.testing import (CUDATestCase, captured_cuda_stdout,
+                                skip_on_cudasim, unittest)
+
+
+def generic_func_1():
+    pass
+
+
+def cuda_func_1():
+    pass
+
+
+def generic_func_2():
+    pass
+
+
+def cuda_func_2():
+    pass
+
+
+def generic_calls_generic():
+    pass
+
+
+def generic_calls_cuda():
+    pass
+
+
+def cuda_calls_generic():
+    pass
+
+
+def cuda_calls_cuda():
+    pass
+
+
+def hardware_overloaded():
+    pass
+
+
+def generic_calls_hardware_overloaded():
+    pass
+
+
+def cuda_calls_hardware_overloaded():
+    pass
+
+
+def hardware_overloaded_calls_hardware_overloaded():
+    pass
+
+
+@overload(generic_func_1, hardware='generic')
+def ol_generic_func_1():
+    def impl():
+        print("Generic function 1")
+    return impl
+
+
+@overload(cuda_func_1, hardware='cuda')
+def ol_cuda_func_1():
+    def impl():
+        print("CUDA function 1")
+    return impl
+
+
+@overload(generic_func_2, hardware='generic')
+def ol_generic_func_2():
+    def impl():
+        print("Generic function 2")
+    return impl
+
+
+@overload(cuda_func_2, hardware='cuda')
+def ol_cuda_func():
+    def impl():
+        print("CUDA function 2")
+    return impl
+
+
+@overload(generic_calls_generic, hardware='generic')
+def ol_generic_calls_generic():
+    def impl():
+        print("Generic calls generic")
+        generic_func_1()
+    return impl
+
+
+@overload(generic_calls_cuda, hardware='generic')
+def ol_generic_calls_cuda():
+    def impl():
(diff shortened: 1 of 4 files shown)


stuartarchibald:
This failure: https://dev.azure.com/numba/ff1fe4d0-ed73-4f1c-b894-1d50a27e048f/_apis/build/builds/8536/logs/130 AFAICT is nothing to do with this PR, fixed in https://github.com/numba/numba/pull/6945

gmarkall:
A quick note on this, mainly for @stuartarchibald: Earlier versions of this had `dispatcher_registry[hardware_registry["cuda"]] = Dispatcher` in `initialize_all()`, but I removed it for this patch because it wasn't used for `@overload` - it will need adding back sometime. My understanding was that this is only needed for the `@jit` decorator with a `hardware` / `target` (whatever we end up calling it) kwarg.

stuartarchibald:
> A quick note on this, mainly for @stuartarchibald: Earlier versions of this had `dispatcher_registry[hardware_registry["cuda"]] = Dispatcher` in `initialize_all()`, but I removed it for this patch because it wasn't used for `@overload` - it will need adding back sometime. My understanding was that this is only needed for the `@jit` decorator with a `hardware` / `target` (whatever we end up calling it) kwarg.


Agree, I think the registrations needed to get `overload` working at present can avoid having a dispatcher registered.

gmarkall:
@stuartarchibald As you suggested OOB I've replaced the printing and captured output in the tests with a value-based check - this is much tidier, and easier to deal with issues.

numba/cuda/tests/cudapy/test_overload.py
```diff
@@ -0,0 +1,290 @@
...
+
+        expected = CUDA_CALLS_HARDWARE_OL * CUDA_HARDWARE_OL
+        self.check_overload(kernel, expected)
+
+    def test_hardware_overloaded_calls_hardware_overloaded(self):
+        def kernel(x):
+            hardware_overloaded_calls_hardware_overloaded(x)
+
+        expected = CUDA_HARDWARE_OL_CALLS_HARDWARE_OL * CUDA_HARDWARE_OL
+        self.check_overload(kernel, expected)
```
stuartarchibald, line 286: In this section, maybe test that the CPU calls would pick up the "generic" variants?

stuartarchibald (commented):
Thanks for the patch @gmarkall, great to see this working :) Once minor comment to resolve else looks good.

numba/cuda/tests/cudapy/test_overload.py
gmarkall: Added.

stuartarchibald:
Thanks for adding the extra tests in f1b54d0, this looks good. There's a conflict to resolve against mainline and once done this can be tested on the farm. Thanks again.

gmarkall:
@stuartarchibald Many thanks, conflict resolved.

stuartarchibald (approved):
Thanks for getting this working, it's great to see the new API in action!

stuartarchibald:
> @stuartarchibald Many thanks, conflict resolved.

Thanks, patch approved.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_49`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_49`.

Passed.

PR #6945, opened 2021-04-19
Fix issue with array analysis tests needing scipy.

stuartarchibald:
This is needed since the implementation of `np.dot` is now an
overload and as a result it must be typed via compilation and if
SciPy is not present this will fail. Fix is to skip this test on
systems with no suitable SciPy.


stuartarchibald:
Context: https://github.com/numba/numba/pull/6944#issuecomment-822632798

sklam (approved):
Thanks for the patch!

PR #6948, opened 2021-04-20
Refactor registry init.

stuartarchibald:
This moves the implementations of core supported features out of
the base context and into explicit targets. This is to ensure that
targets only carry implementations of things they actually
support.


gmarkall:
```python
from numba import cuda
import numpy as np


@cuda.jit
def f(x):
    s = slice(1, 10, 2)
    x[0] = s.start
    x[1] = s.stop
    x[2] = s.step


x = np.zeros(3, dtype=np.int64)

f[1, 1](x)

print(x)
```

prints

```
[ 1 10  2]
```

with this PR even though the slicing registry is not installed on the CUDA target - how is it working now? (Not that it shouldn't be, of course!)

stuartarchibald:
RE slicing. Seems like a lot of registries are being populated as a side effect of import. In the case of `slicing` it's `stencil` from numba's `__init__` which triggers an import chain that reaches it.

numba/tests/test_import.py
```diff
@@ -51,9 +51,63 @@ def test_laziness(self):
 
         out, _ = self.run_in_subproc(code)
         modlist = set(eval(out.strip()))
-        unexpected = set(blacklist) & set(modlist)
+        unexpected = set(banlist) & set(modlist)
         self.assertFalse(unexpected, "some modules unexpectedly imported")
 
+    def test_no_impl_import(self):
```
stuartarchibald, line 57: @gmarkall thoughts on having a test like this for the CUDA target? Essentially something to make sure that `from numba import cuda` / `from numba.cuda import jit` doesn't trigger import of modules that are known to have registration side effects.

stuartarchibald:
> RE slicing. Seems like a lot of registries are being populated as a side effect of import. In the case of `slicing` it's `stencil` from numba's `__init__` which triggers an import chain that reaches it.

The rest of the patches on this PR are to deal with this! They make it so that modules that register lowering implementations are not imported at import time, or at decoration time, but only once compilation starts. The side effect of this is that nothing (not even the NRT) is `jit` compiled on import, this speeds up import times by about 40% locally. These patches also move all the lowering registrations out of the base context and onto the specific targets, such that e.g. the CPU just registers things it needs and CUDA just registers things it needs.

sklam:
@stuartarchibald, the only way I can think of to avoid the move for `typed.Dict` and `typed.List` is to use PEP562 to redirect the import. I tried it at https://github.com/sklam/numba/commit/8230acbb1554f41369689f287a9afffb8c91c979. Is it valid?

numba/tests/test_import.py
gmarkall: Seems that `from numba import cuda` ends up importing `numba.cpython.cmathimpl` and `numba.cpython.mathimpl`.

gmarkall: ... though, this seems to be quite easily fixable.

gmarkall: Test and fix added in https://github.com/numba/numba/commit/0611f058d611ad18276dae57734d4b53062d4180 - what do you think @stuartarchibald ?

stuartarchibald: Thanks @gmarkall, the patch looks good. Are you happy for me to include it in this branch?

stuartarchibald:
> @stuartarchibald, the only way I can think of to avoid the move for `typed.Dict` and `typed.List` is to use PEP562 to redirect the import. I tried it at [sklam@8230acb](https://github.com/sklam/numba/commit/8230acbb1554f41369689f287a9afffb8c91c979). Is it valid?

Thanks @sklam, I think this approach may well work, it's similar to previous shims. As discussed OOB, first Python 3.6 removal is required, will open a new PR for that.

numba/tests/test_import.py
gmarkall: Yes please!

stuartarchibald: Great, thanks for confirming :)

stuartarchibald:
@gmarkall @sklam both your patches are now on this branch, many thanks.

numba/__init__.py
```diff
@@ -56,6 +56,9 @@
 # Initialize hardware
 import numba.core.extending_hardware
 
+# Initialize typed containers
+import numba.typed
```
stuartarchibald, line 60: @sklam Do you recall why this is here? Could it go into the CPU context?

sklam: I think this is needed to register the typed containers for cases where they are used by implicitly; e.g. `d = {}`

sklam (approved):
The changes look good. It's good to see the reduction in import time and that the import side-effects are better controlled.

I profile the import time again from `python -X importtime -c "import numba"`. Here is a descending list of cumulative import time:

```
       self |  cumulative | package
       0.2% |       26.3% | numba.core.config
       0.1% |       21.6% | llvmlite.binding
       0.1% |       21.2% | numba.core.decorators
       0.1% |       20.9% | numba.stencils.stencil
       0.1% |       19.7% | numba.core.types
       0.1% |       19.6% | llvmlite.binding.dylib
       1.1% |       19.0% | llvmlite.binding.ffi
       7.0% |       17.8% | pkg_resources
       5.3% |       17.2% | numpy
       0.1% |       14.7% | numba.core.registry
       0.1% |       14.5% | numba.core.dispatcher
       0.7% |       14.1% | numba.core.compiler
       0.1% |        6.4% | numba.core.callconv
       1.3% |        6.3% | numba.core.base
       0.0% |        5.9% | numba.testing
       0.2% |        5.6% | numba.core.ir_utils
```

If we are interested in further importtime reduction, we should try using PEP562 in `numba/__init__.py` to defer more of the import

stuartarchibald:
Thanks for the review @sklam.

PR #6958, opened 2021-04-21
Inconsistent behavior of reshape between numpy and numba/cuda device array

### LINKED ISSUE #6930: Inconsistent behavior of reshape between numpy and numba/cuda device array
Consider the below code. Numpy will happily reshape using -1 to fill in any missing dimension, numba/cuda will not. Please mirror numpy behavior.
```
import numpy as np
import numba as nb
from numba import cuda
sixty_four = np.zeros(64)

blocks_of_thirty_two = sixty_four.reshape(-1,32)
nb_sixty_four = cuda.to_device(sixty_four)
nb_sixty_four.reshape(-1, 32)
```

laurenarnett:
Fixes #6930. 

A branch is added to check for a single `-1` instance, or an unknown dimension, in the `newdims`. If there is an unknown dimension, the computation is done to work out the missing dimension. Finally, `newdims` is updated with `-1` replaced with the computed dimension at the appropriate index to reflect this.

Unknown dimension inference based off of [numpy's implementation](https://github.com/numpy/numpy/blob/623bc1fae1d47df24e7f1e29321d0c0ba2771ce0/numpy/core/src/multiarray/shape.c#L470-L515).

### DIFF
--- a/numba/misc/dummyarray.py
+++ b/numba/misc/dummyarray.py
@@ -326,6 +326,15 @@ class Array(object):
             else:
                 knownsize *= dim
 
+        # compute the missing dimension
+        if unknownidx >= 0:
+            if knownsize == 0 or self.size % knownsize != 0:
+                raise ValueError("cannot infer valid shape for unknown dimension")
+            else:
+                newdims = newdims[0:unknownidx] \
+                        + tuple((self.size // knownsize,)) \
+                        + newdims[unknownidx + 1:]
+
         newsize = functools.reduce(operator.mul, newdims, 1)
 
         if order == 'A':
--- a/numba/misc/dummyarray.py
+++ b/numba/misc/dummyarray.py
@@ -314,6 +314,18 @@ class Array(object):
         if order not in 'CFA':
             raise ValueError('order not C|F|A')
 
+        # check for exactly one instance of -1 in newdims
+        unknownidx = -1
+        knownsize = 1
+        for i, dim in enumerate(newdims):
+            if dim < 0:
+                if unknownidx == -1:
+                    unknownidx = i
+                else:
+                    raise ValueError("can only specify one unknown dimension")
+            else:
+                knownsize *= dim
+
         newsize = functools.reduce(operator.mul, newdims, 1)
 
         if order == 'A':
--- a/numba/misc/dummyarray.py
+++ b/numba/misc/dummyarray.py
@@ -315,6 +315,7 @@ class Array(object):
             raise ValueError('order not C|F|A')
 
         # check for exactly one instance of -1 in newdims
+        # https://github.com/numpy/numpy/blob/623bc1fae1d47df24e7f1e29321d0c0ba2771ce0/numpy/core/src/multiarray/shape.c#L470-L515   # noqa: E501
         unknownidx = -1
         knownsize = 1
         for i, dim in enumerate(newdims):
--- a/numba/misc/dummyarray.py
+++ b/numba/misc/dummyarray.py
@@ -332,7 +332,7 @@ class Array(object):
                 raise ValueError("cannot infer valid shape for unknown dimension")
             else:
                 newdims = newdims[0:unknownidx] \
-                        + tuple((self.size // knownsize,)) \
+                        + (self.size // knownsize,) \
                         + newdims[unknownidx + 1:]
 
         newsize = functools.reduce(operator.mul, newdims, 1)
--- a/numba/tests/test_dummyarray.py
+++ b/numba/tests/test_dummyarray.py
@@ -208,6 +208,51 @@ class TestReshape(unittest.TestCase):
         self.assertEqual(got.shape, expect.shape)
         self.assertEqual(got.strides, expect.strides)
 
+    def test_reshape_infer2d2d(self):
+        nparr = np.empty((4, 5))
+        arr = Array.from_desc(0, nparr.shape, nparr.strides,
+                              nparr.dtype.itemsize)
+        expect = nparr.reshape(5, 4)
+        got = arr.reshape(-1, 4)[0]
+        self.assertEqual(got.shape, expect.shape)
+        self.assertEqual(got.strides, expect.strides)
+
+    def test_reshape_infer2d1d(self):
+        nparr = np.empty((4, 5))
+        arr = Array.from_desc(0, nparr.shape, nparr.strides,
+                              nparr.dtype.itemsize)
+        expect = nparr.reshape(5 * 4)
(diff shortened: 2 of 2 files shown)


stuartarchibald:
@laurenarnett Many thanks for the patch.

@gmarkall please could you review, thanks!

numba/tests/test_dummyarray.py
```diff
@@ -208,6 +208,51 @@ def test_reshape_3d1d(self):
         self.assertEqual(got.shape, expect.shape)
         self.assertEqual(got.strides, expect.strides)
 
+    def test_reshape_infer2d2d(self):
+        nparr = np.empty((4, 5))
+        arr = Array.from_desc(0, nparr.shape, nparr.strides,
+                              nparr.dtype.itemsize)
+        expect = nparr.reshape(5, 4)
```
gmarkall, line 215: Although this is equal to the shape we'd expect, it does feel a little like we're not quite testing like-for-like here, because we're not creating the expected array with a missing dimension. Perhaps:
```suggestion
        expect = nparr.reshape(-1, 4)
```
would be more explicit that we're testing like-for-like?

(similar applies to the subsequent four `nparr.reshape` calls below too).

gmarkall (changes requested):
Many thanks for the contribution! This is looking great so far:

- CUDA tests pass locally for me with hardware and the simulator.
- The original reproducer from Issue #6930 behaves correctly with this PR (manually verified).
- The logic appears to match the equivalent NumPy implementation (checked by comparing the code side-by-side.

There are a couple of very minor suggestions on the diff, and other than those, this is looking good to me.

gmarkall:
(Moved this to the Numba 0.54 milestone as it seems quite likely to get it in before then)

numba/tests/test_dummyarray.py
laurenarnett: I had been thinking that by explicitly putting the result of the inferred dimension in the expected, it would make it clear that we're ensuring that the logic for working out the missing dimension is working appropriately. But I'm realizing now it's sufficient to match the numpy behavior. I agree that is more explicit in testing like-for-like!

numba/tests/test_dummyarray.py
```diff
@@ -208,6 +208,51 @@ def test_reshape_3d1d(self):
...
+        expect = nparr.reshape(-1, 4)
+        got = arr.reshape(-1, 4)[0]
+        self.assertEqual(got.shape, expect.shape)
+        self.assertEqual(got.strides, expect.strides)
+
+    def test_reshape_infer2d1d(self):
+        nparr = np.empty((4, 5))
+        arr = Array.from_desc(0, nparr.shape, nparr.strides,
+                              nparr.dtype.itemsize)
+        expect = nparr.reshape(5 * 4)
```
gmarkall, line 224: Does this work as `-1` as well?

laurenarnett: Yes it does, just missed that one - good catch.

gmarkall (dismissed):
Many thanks for the updates - this looks good to me!

numba/misc/dummyarray.py
```diff
@@ -314,6 +314,27 @@ def reshape(self, *newdims, **kws):
...
 
+        # check for exactly one instance of -1 in newdims
+        unknownidx = -1
+        knownsize = 1
+        for i, dim in enumerate(newdims):
+            if dim < 0:
+                if unknownidx == -1:
+                    unknownidx = i
+                else:
+                    raise ValueError("can only specify one unknown dimension")
```
stuartarchibald, line 325: Would be good to test that this branch is reached via an appropriate invalid scenario. Numba test suite uses this pattern for asserting exceptions are raised and correct.
https://github.com/numba/numba/blob/d7b1965cb8b5d13a92abd9b8f72f7da8c0fb7354/numba/tests/test_unicode.py#L1677-L1679

numba/misc/dummyarray.py
```diff
@@ -314,6 +314,27 @@ def reshape(self, *newdims, **kws):
         if order not in 'CFA':
             raise ValueError('order not C|F|A')
 
+        # check for exactly one instance of -1 in newdims
```
stuartarchibald, line 317:
```suggestion
        # check for exactly one instance of -1 in newdims
        # https://github.com/numpy/numpy/blob/623bc1fae1d47df24e7f1e29321d0c0ba2771ce0/numpy/core/src/multiarray/shape.c#L470-L515   # noqa: E501
```

Please could you add the above as this is a translation from NumPy? Thanks.

numba/misc/dummyarray.py
```diff
@@ -314,6 +314,27 @@ def reshape(self, *newdims, **kws):
...
+                    unknownidx = i
+                else:
+                    raise ValueError("can only specify one unknown dimension")
+            else:
+                knownsize *= dim
+
+        # compute the missing dimension
+        if unknownidx >= 0:
+            if knownsize == 0 or self.size % knownsize != 0:
+                raise ValueError("cannot infer valid shape for unknown dimension")
```
stuartarchibald, line 332: Would be good to test that this branch is reached via an appropriate invalid scenario.

stuartarchibald (commented):
Thanks for the patch and fixes, this looks good, just a couple of minor things to resolve and then this is good to merge. Many thanks!

stuartarchibald:
> Many thanks for the updates - this looks good to me!

Thanks for the review @gmarkall.

stuartarchibald (approved):

stuartarchibald:
As this doesn't actually modify the CUDA target there's no need to run this through the build farm, CUDA gains the benefit provided through inheritance. @gmarkall would you want a test adding post merge explicitly for CUDA device arrays?

gmarkall:
> @gmarkall would you want a test adding post merge explicitly for CUDA device arrays?

I wouldn't object to one, but I was happy with the patch as-is (without the test) because there are no other CUDA-specific reshape tests. So if someone wants to make some I'd be fine with it, but nobody *needs* to :-)

stuartarchibald:
> > @gmarkall would you want a test adding post merge explicitly for CUDA device arrays?
> 
> I wouldn't object to one, but I was happy with the patch as-is (without the test) because there are no other CUDA-specific reshape tests. So if someone wants to make some I'd be fine with it, but nobody _needs_ to :-)

Thanks @gmarkall . I took a look and couldn't see any so thought I'd ask. The properties are inherited so I think it's fine but if one were to be added at a later time I wouldn't object either :)

This PR should be merged as-is in tonight's merge window.

PR #6961, opened 2021-04-21
Update overload glue to deal with typing_key

stuartarchibald:
As title. Then exercise it working with the `random.random`
module.


sklam (approved):
Thanks for the patch!

PR #6964, opened 2021-04-23
Move minimum supported Python version to 3.7

stuartarchibald:
As title.

### DIFF
--- a/CHANGE_LOG
+++ b/CHANGE_LOG
@@ -3,6 +3,10 @@ Version 0.54.0
 
 In development
 
+NOTES:
+
+* Minimum python version raised to 3.7.
+
 Version 0.53.1 (25 March, 2021)
 -------------------------------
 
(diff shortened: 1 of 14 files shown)


CHANGE_LOG
```diff
@@ -3,6 +3,10 @@ Version 0.54.0
 
 In development
 
+NOTES:
+
+* Minimum python version raised to 3.7.
```
sklam, line 8: We should do this more often. Leaving notes in the changelog for important changes so we don't have to recall 3mos worth of work at release time.

sklam (approved):
Thanks for the patch!

sklam:
@esc, I think we'll need to sync buildfarm before merging this.

esc:
BFID: `numba_smoketest_cpu_yaml` w/o Python 3.6 tasks.

esc:
`numba_smoketest_cpu_yaml_17` was fine.

esc:
Thank you for the patch.

PR #6978, opened 2021-04-28
Implement operator.contains for empty Tuples

brandonwillard:
This PR implements `x in ()`.

### DIFF
--- a/numba/cpython/tupleobj.py
+++ b/numba/cpython/tupleobj.py
@@ -406,3 +406,9 @@ def tuple_index(tup, value):
         raise ValueError("tuple.index(x): x not in tuple")
 
     return tuple_index_impl
+
+
+@overload(operator.contains)
+def in_seq_empty_tuple(x, y):
+    if isinstance(x, types.Tuple) and not x.types:
+        return lambda x, y: False
--- a/numba/tests/test_tuples.py
+++ b/numba/tests/test_tuples.py
@@ -282,7 +282,6 @@ class TestOperations(TestCase):
                "list(undefined)<iv=None>.")
         self.assertIn(msg, str(raises.exception))
 
-
     def test_in(self):
         pyfunc = in_usecase
         cr = compile_isolated(pyfunc,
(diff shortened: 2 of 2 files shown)


esc:
@brandonwillard thank you for submitting this, I have added it to the queue for review. It seems like there are some public CI failures that you could perhaps look into in the meantime.

numba/cpython/tupleobj.py
```diff
@@ -406,3 +406,8 @@ def tuple_index_impl(tup, value):
         raise ValueError("tuple.index(x): x not in tuple")
 
     return tuple_index_impl
+
+
+@overload(operator.contains, types.Tuple([]), types.Any)
```
stuartarchibald, line 411: Providing a signature isn't yet implemented for `@overload`, you'll need to put the type checks into the function body. https://numba.readthedocs.io/en/stable/extending/overloading-guide.html is a good place to start learning about `@overload` and has examples patterns to follow. Due to the lack of guards on the types, this function is accepting any types and is most likely overriding other functions that do similar, as a result some test fail. Good news is that the fix should be quick!

brandonwillard: Ha, yeah, I was thinking of a different decorator when I whipped that together.  I've updated it now.

brandonwillard:
All right, after that change, I'm not sure what those errors are about.  There seems to be [some sort of timeout](https://dev.azure.com/numba/numba/_build/results?buildId=8615&view=logs&j=a506c6d0-8428-5bb5-7d36-c242a5467a36&t=3b81a997-a0fe-5621-27e6-18b6744841b9&l=9741) while attempting to obtain a compiler lock.

stuartarchibald (approved):
Thanks for the patch!

stuartarchibald:
@brandonwillard Congratulations on your first contribution to Numba!

brandonwillard:
My pleasure, and thanks for taking the time to look over this&mdash;especially given that wildly mistaken `overload`. 

Final thought/follow-up: would it actually be better to have an empty tuple inherit from `Sequence` instead?

sklam:
@brandonwillard, the `Sequence` type implies iterable but due to the typing limitation we can't make mix-typed tuple to be iterable; e.g.:

```python
for val in (1, 2.3, "str"):   # type inference can't determine type of `val`
    ...
```

PR #6985, opened 2021-04-30
Implement static set/get items on records with integer index

### LINKED ISSUE #6655: Record staticgetitem with integer index not implemented.
## Reporting a bug

Context: https://numba.discourse.group/t/accessing-structured-array-scalars-by-index/446

This ought to work:
```python
from numba import njit
import numpy as np
arr = np.array([(1, 2., 3.)], dtype='i, f, f')

@njit
def get_elem(arr, idx):
    return arr[idx][0] # record index is const 0

print(get_elem(arr, 0))
```
but doesn't as it is not implemented.

This does work:
```python
from numba import njit
import numpy as np
arr = np.array([(1, 2., 3.)], dtype='i, f, f')

@njit
def get_elem(arr, idx):
    return arr[idx]['f0'] # record index is const 'f0'

print(get_elem(arr, 0))
```
seems strangely inconsistent!

stuartarchibald:
As title.

Closes #6655

### DIFF
--- a/numba/core/typing/arraydecl.py
+++ b/numba/core/typing/arraydecl.py
@@ -555,21 +555,22 @@ class StaticGetItemLiteralRecord(AbstractTemplate):
     def generic(self, args, kws):
         # Resolution of members for records
         record, idx = args
-        if isinstance(record, types.Record) and isinstance(idx, types.StringLiteral):
-            if idx.literal_value not in record.fields:
-                raise KeyError(f"Field '{idx.literal_value}' was not found in record with "
-                               f"fields {tuple(record.fields.keys())}")
-            ret = record.typeof(idx.literal_value)
-            assert ret
-            return signature(ret, *args)
-        if isinstance(record, types.Record) and isinstance(idx, types.IntegerLiteral):
-            if idx.literal_value >= len(record.fields):
-                msg = f"Requested index {idx.literal_value} is out of range"
-                raise IndexError(msg)
-            field_names = list(record.fields)
-            ret = record.typeof(field_names[idx.literal_value])
-            assert ret
-            return signature(ret, *args)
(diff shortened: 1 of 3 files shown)


numba/core/typing/arraydecl.py
```diff
@@ -562,6 +562,14 @@ def generic(self, args, kws):
             ret = record.typeof(idx.literal_value)
             assert ret
             return signature(ret, *args)
+        if isinstance(record, types.Record) and isinstance(idx, types.IntegerLiteral):
```
esc, line 565:
```suggestion
        elif isinstance(record, types.Record) and isinstance(idx, types.IntegerLiteral):
```

To clarify that it is either String or Integer?

numba/core/typing/arraydecl.py
```diff
@@ -575,6 +583,14 @@ def generic(self, args, kws):
             expectedty = record.typeof(idx)
             if self.context.can_convert(value, expectedty) is not None:
                 return signature(types.void, record, types.literal(idx), value)
+        if isinstance(record, types.Record) and isinstance(idx, int):
```
esc, line 586: As above.

esc (changes requested):
Only one statement to consider.

numba/core/typing/arraydecl.py
stuartarchibald: Thanks, in 9d08da6 I've made this an `elif` and factored out the instance check on the `types.Record` to save testing twice!

esc: yeah, that's good!

esc (approved):
Thanks for the patch!

stuartarchibald:
> Thanks for the patch!

Thanks for the review.

PR #6987, opened 2021-04-30
Implement int64() cast and `__hash__` for datetime-like types.

### LINKED ISSUE #6393: Support __hash__ for numpy.datetime64
## Feature request
The numpy datetime64 type can't be used as a typed Dict key currently as Numba doesn't seem to implement a hash function for it.

E.g.
```
import numpy as np
from numba.typed import Dict

d = Dict()
d[np.datetime64("2000-01-01")] = 1
```

Fails with an error message stating that there is no `__hash__` attribute on datetime64[D]

stuartarchibald:
As title.

Closes #6393


stuartarchibald:
Closing as @guilhermeleobas has adopted and fixed this in https://github.com/numba/numba/pull/7047. Thanks @guilhermeleobas !

PR #6992, opened 2021-05-03
Moved typedlist __iter__ into njit code instead of just __getitem__

MegaIng:
This is a very small change rewriting `typed.List.__iter__` to use a new function `_iter` that implements the entire loop in `njit` code, creating a speed-up of 30x (probably because of less calls across the compilation boundaries).

I didn't create new unittest because I assume that typed List is already well tested and I am not changing any user facing behaviour/adding new features.

### DIFF
--- a/numba/typed/typedlist.py
+++ b/numba/typed/typedlist.py
@@ -94,6 +94,12 @@ def _getitem(l, i):
 
 
 @njit
+def _iter(l):
+    for i in range(len(l)):
+        yield l[i]
+
+
+@njit
 def _contains(l, item):
     return item in l
 
@@ -369,8 +375,10 @@ class List(MutableSequence, pt.Generic[T]):
             return _getitem(self, i)
 
     def __iter__(self) -> pt.Iterator[T]:
-        for i in range(len(self)):
-            yield self[i]
+        if not self._typed:
+            return iter([])
+        else:
+            return _iter(self)
 
     def __contains__(self, item: T) -> bool:  # type: ignore[override]
         return _contains(self, item)


esc:
@MegaIng thank you for your contribution! The CI failure is probably unrelated so you can ignore that for now and I assume that the code is probably fine.

esc:
@MegaIng I have added it to the queue for review.

MegaIng:
@esc No, it apparently isn't unrelated. I can reproduce this on my local machine that the change I made breaks this. I still have to figure out why. Somehow the changed implementation creates a memory leak (or it exposes an already existing leak).

esc:
@MegaIng oh good, I recently saw someone mention "master is broken b/c of pickle" so I assume this was it. Memory leaks in Numba are often caused by exceptions. Exceptions still leak memory, which is why we deactivate the leak checker explicitly in tests that check exceptions. IIRC using an iterator raises a `StopIteration` exception, so that may be one avenue to explore.

MegaIng:
@esc Yes generators raise `StopIteration `. But as far as I can tell, this isn't a problem in `test_generators.py`. Is there any info somewhere on how I could debug this further?

esc:
@MegaIng there is, yes, but it is very minimal:

https://numba.pydata.org/numba-doc/dev/developer/numba-runtime.html#debugging-leaks

PS: I'm signing off now for the day.

MegaIng:
Ok, I found the problem. It is actual not related to the code in this PR, but on how numba handles generators in general. I will create  an issue.

stuartarchibald:
On the basis that this PR requires likely involved fixes to generators, I'm going to move this out of 0.54. If upstream issues are fixed then am happy to move it back into the milestone.

Issue #701, opened 2014-08-15
numpy datetime tests fail with Numpy 1.6


pitrou:
The whole numba.npdatetime module fails importing with the following error:

```
  File "/home/build/anaconda/envs/_test/lib/python2.7/site-packages/numba/npdatetime.py", line 30, in <module>
    NAT = np.timedelta64('nat').astype(np.int64)
ValueError: error setting an array element with a sequence
```

We should fix the error on import and instead mark np.datetime64 and np.timedelta64 unsupported with Numpy 1.6.

PR #6997, opened 2021-05-04
CUDA: Remove catch of NotImplementedError in target.py

gmarkall:
Maybe this is more of a question than a PR, but I saw it added in #6948 and I'm not sure what it does. Removing it seems not to cause a problem - all tests still seem to pass without it.

### DIFF
--- a/numba/core/cpu.py
+++ b/numba/core/cpu.py
@@ -77,11 +77,7 @@ class CPUContext(BaseContext):
         from numba.typed import typeddict, dictimpl
         from numba.typed import typedlist, listobject
         from numba.experimental import jitclass, function_type
-
-        try:
-            from numba.np import npdatetime
-        except NotImplementedError:
-            pass
+        from numba.np import npdatetime
 
         # Add target specific implementations
         from numba.np import npyimpl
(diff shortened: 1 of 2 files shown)


gmarkall:
From OOB discussion, it seems like this might have initially been added due to #701, which is no longer relevant because NumPy 1.6 is no longer supported, and it also appears that there is no way to trigger the `NotImplementedError` at import time. The PR has been updated to remove the `try` from the CPU target's import of `numba.np.npdatetime` as well.

stuartarchibald (approved):
Thanks for the patch. I've tried a few ways to trigger the noted exception and it seems like something else is always an issue before a `NotImplementedError` is hit. Further, it appears on inspection that the exception cannot be raised at import time as the control flow required to do so is always in a closure held by a lowering implementation and not actually executed as part of an import sequence.

This patch should get a CPU and CUDA smoke test and assuming both come back fine I think that would further confirm its validity.

Thanks again!

stuartarchibald:
Waiting on merge of https://github.com/numba/numba/pull/7003

stuartarchibald:
Buildfarm ID: `numba_smoketest_cpu_yaml_30`.
Buildfarm ID: `numba_smoketest_cuda_yaml_62`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_62`.

Passed.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cpu_yaml_30`.

Passed

PR #7001, opened 2021-05-06
Make some `numba.core` modules flake8 compliant

gmarkall:
This makes some files in `numba.core` flake8 compliant, but before I could finish them all it got to the point where I've spent a bit too much time on this. The remainder can be addressed later.

This PR also removes some non-existent files from the ignore list.


gmarkall:
To continue this, it will probably be easier to re-do than to attempt to resolve conflicts here - I probably tried to get too much done in one go.

PR #7004, opened 2021-05-06
Test extending the CUDA target

gmarkall:
It occurred to me that people are starting to rely on extending the CUDA target using the low-level API, but there are no specific tests of this API for CUDA. This PR adds tests based on a modification of the Interval example from the documentation for the CUDA target.


stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_55`.

Passed

PR #7007, opened 2021-05-07
Made stencil compilation not fail for arrays of conflicting types.

MegaIng:
Based on a short bug report on [gitter](https://gitter.im/numba/numba?at=60959645c651cb6a001aa559).

Makes `raise_if_incompatible_array_sizes` use `literal_unroll` so that there is no problem when passed arrays 2 and 3 (e.g. `args[0]` and `args[1]`) are of type than can't be unified (i.e. `bool` and `int`).

I also added a test, but I am not completly sure if they are ok in the form they are.

### DIFF
--- a/numba/stencils/stencil.py
+++ b/numba/stencils/stencil.py
@@ -35,6 +35,15 @@ class StencilFuncLowerer(object):
 @register_jitable
 def raise_if_incompatible_array_sizes(a, *args):
     ashape = a.shape
+
+    # We need literal_unroll here because the stencil might take
+    # multiple input arrays with different types that are not compatible
+    # (e.g. values as float[:] and flags as bool[:])
+    # When more than three total arrays are given, the second and third
+    # are iterated over in the loop below. Without literal_unroll, their
+    # types have to match.
+    # An example failing signature without literal_unroll might be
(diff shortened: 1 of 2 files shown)


numba/stencils/stencil.py
```diff
@@ -34,7 +35,7 @@ def __call__(self, context, builder, sig, args):
 @register_jitable
 def raise_if_incompatible_array_sizes(a, *args):
     ashape = a.shape
-    for arg in args:
+    for arg in literal_unroll(args):
```
DrTodd13, line 38: Would you please add a comment and probably include this example from the PR before this line just saying why the literal_unroll is necessary here.

DrTodd13 (changes requested):
Thanks for the PR.  Looks good but I did want just the one additional comment.

numba/stencils/stencil.py
MegaIng: Done.

DrTodd13 (approved):

stuartarchibald (approved):
Thanks for the patch @MegaIng , thanks for reviewing @DrTodd13 .

PR #7008, opened 2021-05-08
Added support for np.random.dirichlet with all size arguments

rishi-kulkarni:
I added support for np.random.dirichlet accepting size arguments of None, int, or tuple. I copied the np.random.multinomial formatting and changed the algorithm to make the Dirichlet distribution.


numba/tests/test_random.py
```diff
@@ -1291,6 +1294,98 @@ def test_multinomial_3_tuple(self):
...
+    def _check_sample(self, alpha, size, sample):
+        '''Check output properties'''
+        if not size is None:
+
+            if type(size) is tuple:
+                self.assertIsInstance(sample, np.ndarray)
+                self.assertEqual(sample.shape, size + (len(alpha),))
+                self.assertEqual(sample.dtype, np.dtype(np.float64))
+                '''Check statistical properties'''
+                self.assertAlmostEqual(sample.sum(), np.prod(size), places=5)
```
stuartarchibald, line 1309: This sort of check seems like it might be susceptible to numerical "accidents" e.g.  unlucky RNG generates a pathological sequence that doesn't meet this condition, are there other properties that could perhaps be used instead?

numba/tests/test_random.py
```diff
@@ -1291,6 +1294,98 @@ def test_multinomial_3_tuple(self):
...
+                    self.assertLessEqual(val, 1)
+        else:
+            self.assertIsInstance(sample, np.ndarray)
+            self.assertEqual(sample.size, len(alpha))
+            self.assertEqual(sample.dtype, np.dtype(np.float64))
+            '''Check statistical properties'''
+            self.assertAlmostEqual(sample.sum(), 1, places=5)
+            for val in np.nditer(sample):
+                self.assertGreaterEqual(val, 0)
+                self.assertLessEqual(val, 1)
```
stuartarchibald, line 1331: The 3 blocks in here look like they could be parameterised into an expected "shape" and "statistical property" ?

numba/cpython/randomimpl.py
```diff
@@ -1506,3 +1506,66 @@ def multinomial_impl(n, pvals, size=None):
...
+    elif isinstance(size, types.Integer):
+        def dirichlet_impl(alpha, size=None):
+            """
+            dirichlet(..., size=int)
+            """
+            out = np.zeros((size, len(alpha)), dtype)
+            dirichlet_arr(alpha, out)
+            return out
+
+    elif isinstance(size, types.BaseTuple):
```
stuartarchibald, line 1558: I'm not sure this is safe? What if the tuple is `('a', 1, 3j)` ?

numba/cpython/randomimpl.py
```diff
@@ -1506,3 +1506,66 @@ def multinomial_impl(n, pvals, size=None):
...
+
+        #Gamma distribution method to generate a Dirichlet distribution
+
+        a_len = len(alpha)
+        size = out.size
+        flat = out.flat
+        for i in range(0, size, a_len):
+            exp = np.zeros(len(alpha), dtype=np.float64)
+            #calculate gamma random numbers per alpha specifications
+            for k, w in enumerate(alpha):
```
stuartarchibald, line 1529: This should explicitly raise `ValueError` for an `alpha <= 0`?

numba/cpython/randomimpl.py
```diff
@@ -1506,3 +1506,66 @@ def multinomial_impl(n, pvals, size=None):
...
+            """
+            dirichlet(..., size=tuple)
+            """
+            out = np.zeros(size + (len(alpha),), dtype)
+            dirichlet_arr(alpha, out)
+            return out
+
+    else:
+        raise TypeError("np.random.dirichlet(): size should be int or "
+                        "tuple or None, got %s" % (size,))
```
stuartarchibald, line 1569: This needs a test to ensure it executes under the correct conditions.

numba/cpython/randomimpl.py
rishi-kulkarni: Good point. This line is also in np.random.multinomial - worth changing in this PR?

numba/cpython/randomimpl.py
```diff
@@ -1506,3 +1506,66 @@ def multinomial_impl(n, pvals, size=None):
...
+        a_len = len(alpha)
+        size = out.size
+        flat = out.flat
+        for i in range(0, size, a_len):
+            exp = np.zeros(len(alpha), dtype=np.float64)
+            #calculate gamma random numbers per alpha specifications
+            for k, w in enumerate(alpha):
+                exp[k] = np.random.gamma(w, 1)
+            #normalize sum to 1
+            exp /= exp.sum()
```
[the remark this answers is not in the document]
rishi-kulkarni: Pardon my misunderstanding, but by this do you mean rewriting it something like:

```
exp = np.empty(0, dtype=np.float64)
for k, w in enumerate(alpha):
    exp = np.append(exp, np.random.gamma(w,1)
exp /= exp.sum()
```

numba/cpython/randomimpl.py
```diff
@@ -1506,3 +1506,66 @@ def multinomial_impl(n, pvals, size=None):
...
+            for j, v in enumerate(exp):
+                flat[i+j] = v.item()
+
+    if not isinstance(alpha, (types.Sequence, types.Array)):
+        raise TypeError("np.random.dirichlet(): alpha should be an "
+                        "array or sequence, got %s" % (alpha,))
+
+    if not isinstance(alpha, types.Array):
+        alpha = np.array(alpha)
```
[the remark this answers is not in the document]
rishi-kulkarni: Honestly, I don't remember - it's gone now.

numba/tests/test_random.py
rishi-kulkarni: Yeah - I've written it so that it now computes the sum along the last axis and makes sure every element of that array is equal to 1.

numba/cpython/randomimpl.py
rishi-kulkarni: Added a set of tests for incorrect size arguments.

numba/cpython/randomimpl.py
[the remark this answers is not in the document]
stuartarchibald: Apologies, that might have not been hugely clear, I was referring to not needing `exp` to be an array, looks like the loop can just accumulate the gamma values and then normalise, all without an allocation.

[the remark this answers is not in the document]
rishi-kulkarni: Ah, yes - exp is long gone now, though the speed difference is pretty minor. You're right that there's no need to make extra memory allocations, though.

numba/tests/test_random.py
```diff
@@ -1291,6 +1294,129 @@ def test_multinomial_3_tuple(self):
...
+        alpha = np.array([1, 1, 10000, 1], dtype=np.float64)
+        res = cfunc(alpha, size)
+        self._check_sample(alpha, size, res)
+        # A case with noninteger probabilities
+        alpha = np.array([1, 1, 1.5, 1], dtype=np.float64)
+        res = cfunc(alpha, size)
+        self._check_sample(alpha, size, res)
+        alpha = tuple(alpha)
+        res = cfunc(alpha, size)
+        self._check_sample(alpha, size, res)
```
stuartarchibald, line 1395: Can this just be a loop in "alphas" x "sizes", i.e.
```python
sizes = (None, (10,), (10, 10))
alphas = (self.alpha, tuple(self.alpha),
          np.array([1, 1, 10000, 1], dtype=np.float64),np.array([1, 1, 1.5, 1], dtype=np.float64) )

for alpha, size in itertools.product(alphas, sizes):
    res = cfunc(alpha, size)
    self._check_sample(alpha, size, res)
```

numba/tests/test_random.py
```diff
@@ -1291,6 +1294,129 @@ def test_multinomial_3_tuple(self):
...
+            cfunc(alpha, size)
+        size = (1.5, 1)
+        with self.assertTypingError():
+            cfunc(alpha, size)
+        size = (3j, 1)
+        with self.assertTypingError():
+            cfunc(alpha, size)
+        size = (3j, 3j)
+        with self.assertTypingError():
+            cfunc(alpha, size)
```
stuartarchibald, line 1419: Whilst not in use in this test module (as it's relatively old), Numba uses this convention for checking that errors (particularly typing errors) come from an expected place:
https://github.com/numba/numba/blob/e395e00cd7720494f68b44e97dc3e51b1a0c9757/numba/tests/test_np_functions.py#L600-L605
essentially, the error message strings are captured and checked to make sure they are as expected.

numba/cpython/randomimpl.py
```diff
@@ -1506,3 +1506,68 @@ def multinomial_impl(n, pvals, size=None):
...
+            norm = 0  # use this to normalize every the group total to 1
+            for k, w in enumerate(alpha):
+                flat[i+k] = np.random.gamma(w, 1)
+                norm += flat[i+k].item()
+            for k, w in enumerate(alpha):
+                flat[i+k] /= norm
+
+    if not isinstance(alpha, (types.Sequence, types.Array)):
+        raise TypeError("np.random.dirichlet(): alpha should be an "
+                          "array or sequence, got %s" % (alpha,))
```
stuartarchibald, line 1539:
```suggestion
        raise TypeError("np.random.dirichlet(): alpha should be an "
                        "array or sequence, got %s" % (alpha,))
```

numba/cpython/randomimpl.py
```diff
@@ -1506,3 +1506,68 @@ def multinomial_impl(n, pvals, size=None):
...
+
+        a_len = len(alpha)
+        size = out.size
+        flat = out.flat
+        for i in range(0, size, a_len):
+            # calculate gamma random numbers per alpha specifications
+            norm = 0  # use this to normalize every the group total to 1
+            for k, w in enumerate(alpha):
+                flat[i+k] = np.random.gamma(w, 1)
+                norm += flat[i+k].item()
```
stuartarchibald, line 1533:
```suggestion
                norm += flat[i + k].item()
```

numba/cpython/randomimpl.py
```diff
@@ -1506,3 +1506,68 @@ def multinomial_impl(n, pvals, size=None):
...
+            for k, w in enumerate(alpha):
+                flat[i+k] /= norm
```
stuartarchibald, line 1535:
```suggestion
                flat[i + k] /= norm
```

numba/cpython/randomimpl.py
```diff
@@ -1506,3 +1506,68 @@ def multinomial_impl(n, pvals, size=None):
...
+            def dirichlet_impl(alpha, size=None):
+                """
+                dirichlet(..., size=tuple)
+                """
+                out = np.empty(size + (len(alpha),))
+                dirichlet_arr(alpha, out)
+                return out
+        else:
+            raise TypeError("np.random.dirichlet(): size should be int or "
+                              "tuple of ints or None, got %s" % (size,))
```
stuartarchibald, line 1567:
```suggestion
            raise TypeError("np.random.dirichlet(): size should be int or "
                            "tuple of ints or None, got %s" % size)
```

numba/cpython/randomimpl.py
```diff
@@ -1506,3 +1506,68 @@ def multinomial_impl(n, pvals, size=None):
...
+                out = np.empty(size + (len(alpha),))
+                dirichlet_arr(alpha, out)
+                return out
+        else:
+            raise TypeError("np.random.dirichlet(): size should be int or "
+                              "tuple of ints or None, got %s" % (size,))           
+
+    else:
+        raise TypeError("np.random.dirichlet(): size should be int or "
+                          "tuple of ints or None, got %s" % (size,))
```
stuartarchibald, line 1571:
```suggestion
        raise TypeError("np.random.dirichlet(): size should be int or "
                        "tuple of ints or None, got %s" % size)
```

numba/tests/test_random.py
```diff
@@ -61,6 +61,9 @@ def numpy_multinomial2(n, pvals):
 def numpy_multinomial3(n, pvals, size):
     return np.random.multinomial(n, pvals=pvals, size=size)
 
+def numpy_dirichlet(alpha, size):
+    return np.random.dirichlet(alpha, size=size)
```
stuartarchibald, line 65: A second variant of this, like:
```python
def numpy_dirichlet(alpha):
    return np.random.dirichlet(alpha)
```
should be tested to ensure that default arguments work as anticipated.

stuartarchibald (commented):
Thanks for the updates, this is looking good. I've made a few suggestions in relation to testing this some more, once resolved I think this will be suitable for merging. Thanks again!

numba/tests/test_random.py
rishi-kulkarni: Added this in.

numba/tests/test_random.py
rishi-kulkarni: Great, I added this in and cleaned up that set of tests.

numba/tests/test_random.py
rishi-kulkarni: Done.

numba/cpython/randomimpl.py
```diff
@@ -1506,3 +1506,68 @@ def multinomial_impl(n, pvals, size=None):
...
+        def dirichlet_impl(alpha, size=None):
+            """
+            dirichlet(..., size=int)
+            """
+            out = np.empty((size, len(alpha)))
+            dirichlet_arr(alpha, out)
+            return out
+
+    elif isinstance(size, types.UniTuple):
+        if size is types.UniTuple(types.intp, len(size)):
```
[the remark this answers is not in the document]
rishi-kulkarni: I couldn't think of a clean way to take mixed tuples as arguments, but I added in a test to make sure it raises an exception. Regardless, I also fixed the intp issue.

numba/cpython/randomimpl.py
```diff
@@ -1506,3 +1506,77 @@ def multinomial_impl(n, pvals, size=None):
...
+        def dirichlet_impl(alpha, size=None):
+            """
+            dirichlet(..., size=int)
+            """
+            out = np.empty((size, len(alpha)))
+            dirichlet_arr(alpha, out)
+            return out
+
+    elif isinstance(size, (types.UniTuple)):
+        if isinstance(size, (types.UniTuple)) and isinstance(size.dtype, types.Integer):
```
stuartarchibald, line 1560:
```suggestion
    elif isinstance(size, (types.UniTuple)) and\
         isinstance(size.dtype, types.Integer):
```
I think the first condition only needs testing once and if both conditions are not met then the below `else` raising the explicit `TypeError` should be hit.

numba/tests/test_random.py
```diff
@@ -1291,6 +1299,77 @@ def test_multinomial_3_tuple(self):
...
+        )
+
+        for alpha, size in itertools.product(alphas, sizes):
+            res = cfunc(alpha, size)
+            self._check_sample(alpha, size, res)
+
+    def test_dirichlet_exceptions(self):
+        cfunc = jit(nopython=True)(numpy_dirichlet)
+        alpha = tuple((0, 1, 1))
+        self.assertRaises(ValueError, cfunc, alpha, 1)
```
stuartarchibald, line 1362: Please could the exception error message be checked? Thanks.

stuartarchibald (commented):
Thanks for the update. There's a couple of minor things in the latest diff to address otherwise the new patch looks good. Thanks for working on this!

numba/tests/test_random.py
rishi-kulkarni: Yep, added that in.

numba/cpython/randomimpl.py
rishi-kulkarni: Good point, I shortened this block.

stuartarchibald (approved):
Thanks for the patch and all your efforts in implementing this.

stuartarchibald:
@rishi-kulkarni Congratulations on your first contribution to Numba!

rishi-kulkarni:
Thanks! Happy to add a little to a project that's made my own job so much easier.

Issue #6864, opened 2021-03-25
guvectorize returns different output after upgrading to Numpy 1.20


RutgerK:
After upgrading from Numpy 1.19.5 to Numpy 1.20.1 a Numba function I wrote using `guvectorize` suddenly returned different results. 

The issue appeared to be related to a mismatch between the datatype of the output array, and the signature of the function. **So at it's core, the bug is in the code I wrote**. But I'm posting it here anyway because I'm not sure if the change in behavior between the Numpy versions is intentional.

I provide a Numpy array initialized with zeros to the Numba function. The result is correct for elements in the output that are explicitly modified, but unmodified elements are returned as semi-random values. The wrong (unmodified) output values resemble what I would expect when initializing an array with `np.empty(...`.

The solution is (of course) to make sure the datatype of the output array matches that of the signature. Both changing the signature to `int8[:]`, or the output array to `np.uint8` would work in this case.

A toy example replicating the behavior is shown below. Notice the `np.int8` output array versus `uint8[:]` in the signature:
```python
import numba
import numpy as np

@numba.guvectorize(["void(float64[:], uint8[:])"], "(n)->(n)", nopython=True)
def func(x, out):
    
    for i in range(x.size):
        # set every fourth element to 1
        if i % 4 == 0:
            out[i] = 1

x = np.random.rand(150,150)
out = np.zeros_like(x, dtype=np.int8)

func(x, out)
```
With Numba 0.52 or 0.53 in combination with **Numpy 1.20.1**, this is the result:
```python
array([[  1,  49,  29, ...,  96,   1,   2],
       [  1,   0, -80, ..., -38,   1,  97],
       [  1,   2,   0, ...,   0,   1,   0],
       ...,
       [  1,  -1,  -1, ...,   0,   1,   0],
       [  1,   0,   0, ...,   0,   1,   0],
       [  1,   0,   0, ..., -34,   1,  97]], dtype=int8)
```

And with Numba 0.52 or 0.53 in combination with **Numpy 1.19.5**, this is the result:
```python
array([[1, 0, 0, ..., 0, 1, 0],
       [1, 0, 0, ..., 0, 1, 0],
       [1, 0, 0, ..., 0, 1, 0],
       ...,
       [1, 0, 0, ..., 0, 1, 0],
       [1, 0, 0, ..., 0, 1, 0],
       [1, 0, 0, ..., 0, 1, 0]], dtype=int8)
```

[...]

PR #7009, opened 2021-05-09
Add writable args

dmbelov:
This pull request adds an option `writable_args` to `guvectorize` that allows one to specify indices of input arguments that should be read/write.  In other words, this PR allows one to write vectorized functions that update arguments in-place with reliability.

Here is simplistic example:
```python
import numba
import numpy as np

def _replace_2nd(x_t, y_1):
    for t in range(0, x_t.size, 2):
        x_t[t] = y_1[0]

replace_2nd = numba.guvectorize(["void(f8[:], f8[:])"], "(t),()", nopython=True)(_replace_2nd)

# replaces every 2nd element by 2 
x_t = np.zeros(5)
replace_2nd(x_t, 2)
assert (x_t[::2] == 2).all()

# does NOT replace every 2nd element by 2 
x_t = np.zeros(5, dtype='f4')
replace_2nd(x_t, 2)
assert (x_t[::2] == 0).all()

# code with writable_args
replace_2nd = numba.guvectorize(["void(f8[:], f8[:])"], "(t),()", nopython=True, writable_args=(0,))(_replace_2nd)

# replace every 2nd element by 2 
x_t = np.zeros(5, dtype='f4')
replace_2nd(x_t, 2)
assert (x_t[::2] == 2).all()

```

Related issue: #6864


gmarkall:

dmbelov:
Why my PR was moved to Numba 0.56RC.   I need this change for my work.   Is it possible to return this PR back to Numba 0.55RC?

gmarkall:
@dmbelov Apologies for the long delay in someone reviewing your PR - it was moved to 0.56 because 0.55 RC was produced before someone could review this.

gmarkall (changes requested):
Many thanks for the PR - some initial thoughts as I've had a quick look through this.

- The code changes generally look good and appear to align with the discussions in #6864 and numpy/numpy#18700
- The documentation presently states that you can't reliably write to input values in this section - https://numba--7009.org.readthedocs.build/en/7009/user/vectorize.html#overwriting-input-values - could you also update this documentation to explain that it is possible using `writable_args`? Adapting your example from the PR description may also be helpful.
- Swapping `nopython=True` for `target='cuda'` in the decorator results in a `TypeError` - I think it won't be as simple to support for the CUDA target so you probably won't want to make it work for CUDA; could you ensure that if `target='cuda'` is used with `writable_args`, then an explanatory error message is presented to the user please?
- How does this interact with dynamic gufuncs, where a signature is not provided up front?
- The PR description mentions that this adds the new kwarg for `vectorize` as well as `guvectorize`, but there aren't any tests or example with `vectorize` - I think it's fine if this is only for `guvectorize` (I'm not sure I see how it could be applied to `vectorize` as the arguments to a `vectorize` function tend to get passed in as scalars) - could you clarify whether there's `vectorize` support please?
- I've only skimmed the tests so far - many thanks for adding tests with the PR.

(The error when trying to use the CUDA target was:

```
  File "/home/gmarkall/numbadev/issues/7009/repro.py", line 8, in <module>
    replace_2nd = numba.guvectorize(["void(f8[:], f8[:])"], "(t),()", target='cuda')(_replace_2nd)
  File "/home/gmarkall/numbadev/numba/numba/np/ufunc/decorators.py", line 201, in wrap
    guvec = GUVectorize(func, signature, **kwargs)
  File "/home/gmarkall/numbadev/numba/numba/np/ufunc/decorators.py", line 60, in __new__
    return imp(func, signature, identity=identity, cache=cache,
TypeError: __init__() got an unexpected keyword argument 'writable_args'
```
)

gmarkall:
@dmbelov Many thanks for the update - I've marked this as waiting for another review.

As a small request regarding the process - could you please avoid force-pushes / rebases and instead add new commits and merge from main please? It is quite hard to tell what changed since the previous review / comments if the history is edited - many thanks in advance.

dmbelov:
I am in process of implementing changes suggested by you on [Dec 30, 2021].  It will take me few days.

The last push was just rebase on the current version of main branch.

dmbelov:
@gmarkall,
I the last 3 commits I addressed all your concerns above:

- [x] Updated doc `vectorize.rst` by adding description of `writable_args` there.
- [x] Fixed code for `target='cuda' `: it raises `TypeError` now.
- [x] interaction with dynamic gufuncs: I think it works because it auto generates function for every given dtype combination.  At least I was not able to find non-working combination.
- [x] I removed `vectorize` from PR.  `writable_args` only make sense for `guvectorize`.
- [ ] Regarding tests: I do not see any failures for CPU tests caused by my changes (I cannot run CUDA tests on my computer).

dmbelov:
@gmarkall,
Can u add this PR request to the nearest numba release?

numba/np/ufunc/_ufunc.c
```diff
@@ -218,5 +220,27 @@ ufunc_fromfunc(PyObject *NPY_UNUSED(dummy), PyObject *args)
         return NULL;
     }
 
+    /* update op_flags */
+    if (writable_args_tuple != NULL) {
+        n_writable_args =PyTuple_Size(writable_args_tuple);
```
gmarkall, line 225:
```suggestion
        n_writable_args = PyTuple_Size(writable_args_tuple);
```

gmarkall (commented):
Thanks for the updates @dmbelov. Some comments / questions on the current iteration:

- I should have checked sooner, but I did have a concern about whether this was a legitimate thing to do or not. Having looked at how NumPy dispatches ufuncs, I'm comfortable that this is safe to do. The function [`_ufunc_setup_flags()`](https://github.com/numpy/numpy/blob/ac39f38987387307ca099ee69fabf33050e10b33/numpy/core/src/umath/ufunc_object.c#L373-L404) explicitly checks for / expects the possibility of inputs being written to. This function is reached during ufunc dispatch e.g. through `ufunc_generic_fastcall()` and `PyUFunc_GeneralizedFunctionInternal()` or `PyUFunc_GenericFunctionInternal()`.
- The changes to the C code in `_ufunc.c` look good (modulo one tiny formatting issue).
- Could you add additional tests for the cases where there are multiple (or 2, at least) writable args, for the exception raised when an invalid argument is specified, and when one attempts to use the CUDA target please?
- The API feels a little awkward - what do you think about the possibility of specifying writable args by name instead of index? How difficult would that be to add / change to?

dmbelov:
@gmarkall , sorry for late reply.  I addressed your suggestions in the commits above.

> The changes to the C code in _ufunc.c look good (modulo one tiny formatting issue).

Fixed

> Could you add additional tests for the cases where there are multiple (or 2, at least) writable args, for the exception raised when an invalid argument is specified, and when one attempts to use the CUDA target please?

Done

> The API feels a little awkward - what do you think about the possibility of specifying writable args by name instead of index? How difficult would that be to add / change to?

As you know, API for Numpy universal and generalized universal functions does **not** support keywords.  Arguments in NumPy C-API are specified by their position.   Current interface for `writable_args` exactly mimics this C-API.

Can you merge my request to the main branch and put this change in the nearest Numba release?

gmarkall:
gpuci run tests

dmbelov:
> gpuci run tests

What do you mean?

gmarkall:
> > gpuci run tests
> 
> What do you mean?

Apologies for the confusion! That phrase triggers the CI system that runs tests with actual GPU hardware - I wanted to see that the test with `target='cuda'` was OK with actual CUDA hardware present.

gmarkall:
Whoops, and quoting it has just triggered it again!

gmarkall:
The consensus from the dev meeting is that the addition of `writable_args` should also allow the arguments to be specified by name: draft minutes are at https://hackmd.io/n5qSqeMTR4Wv4UYkUwCZng but they will be added to the wiki over the next couple of days.

dmbelov:
Thanks.  I added changes as suggested. Please check again.

dmbelov:
@gmarkall, do you think that this PR is ready for merging to the `main` branch?

dmbelov:
@gmarkall, any updates?

docs/source/user/vectorize.rst
```diff
@@ -297,6 +297,28 @@ demonstrate, we can  use an array of `float32` with the `init_values` function::
 In this case, there is no change to the `invals` array because the temporary
 casted array was mutated instead.
 
+.. note::
+   To ensure that ``@guvectorize`` updates `n`-th and `m`-th input arguments pass 
+   ``writable_args=(n, m)``.
+
```
gmarkall, line 303: I think this note can be removed as it seems to serve only to provide a reference-like description of `writable_args`, which is introduced in a more tutorial-like fashion in the following paragraph.

To replace this, there should be a description in the reference manual for the `writable_args` kwarg.

docs/source/user/vectorize.rst
```diff
@@ -297,6 +297,28 @@ demonstrate, we can  use an array of `float32` with the `init_values` function::
...
+To solve this problem, one needs to tell the GUFunc engine that 0-th
+argument is writable. This can be achieved by passing argument position
+``writable_args=(0,)`` or argument name ``writable_args=('invals',)``
+to ``@guvectorize``.  Now, the code above works as expected::
+
+   @guvectorize([(float64[:], float64[:])], '()->()', writable_args=(0,))
```
gmarkall, line 309: Continuing basing the example around argument names:

```suggestion
   @guvectorize([(float64[:], float64[:])], '()->()', writable_args=('invals',))
```

gmarkall (changes requested):
Many thanks for the updates @dmbelov. I think the implementation is looking good. Some minor points to resolve:

- There are some suggestions on the docs. Additionally, the `writable_args` kwarg should be documented in the reference manual in this section: https://numba--7009.org.readthedocs.build/en/7009/reference/jit-compilation.html#numba.guvectorize
- A test of the exception raised when the user specifies a non-existent arg by name is still required.

Many thanks!

dmbelov:
@gmarkall,
Thanks for the review. I made the changes that you suggested:
- updated the documentation;
- added test of the exception raised when the user specifies a non-existent arg by name.

Please check if PR is good to be merged.

dmbelov:
Thanks!

sklam:
BFID: `numba_smoketest_cpu_yaml_171`

sklam:
> BFID: `numba_smoketest_cpu_yaml_171`

passed

PR #7010, opened 2021-05-10
WIP: Add ddof argument to nanvar and nanstd

### LINKED ISSUE #6611: Feature Request: Support ddof argument in np.nan[std | var]
---

## Feature request

Related to https://github.com/numba/numba/issues/6610, it would also be great if the `ddof` argument could be supported in `np.nanstd` and `np.nanvar`

mroeschke:
- [x] closes #6611

### DIFF
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -920,12 +920,12 @@ def np_nanmean(a):
 
 
 @overload(np.nanvar)
-def np_nanvar(a):
+def np_nanvar(a, ddof=0):
     if not isinstance(a, types.Array):
         return
     isnan = get_isnan(a.dtype)
 
-    def nanvar_impl(a):
+    def nanvar_impl(a, ddof=ddof):
         # Compute the mean
         m = np.nanmean(a)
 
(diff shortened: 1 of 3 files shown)


numba/np/arraymath.py
```diff
@@ -939,18 +939,18 @@ def nanvar_impl(a):
...
 
 @overload(np.nanstd)
-def np_nanstd(a):
+def np_nanstd(a, ddof=0):
     if not isinstance(a, types.Array):
         return
 
-    def nanstd_impl(a):
-        return np.nanvar(a) ** 0.5
+    def nanstd_impl(a, ddof=ddof):
```
stuartarchibald, line 952: I think the test failures come down to this:
```suggestion
    def nanstd_impl(a, ddof=0):
```
the outer scope function arguments are in the "numba type" domain, the inner are in the "numba compilation" domain, by using the outer scope definition here it's asking the compiler to compile a type opposed to an instance of the type. There's a user guide here which explains some of this https://numba.readthedocs.io/en/stable/extending/overloading-guide.html hope this helps?

mroeschke:
Sorry I don't have time to pick this back up. Will reopen if I get back around to it

PR #7011, opened 2021-05-10
Allocate new memory in the stack to save array transpose value insteading of sharing memory with original array, this fix numba#6949 .

### LINKED ISSUE #6949: x += x.T and x = x + x.T yield different result.
Numba version : 0.53.1
Python version : 3.8.5
Numpy version : 1.19.2
## Bug Description
When working with 2D array, `x+= x.T `and `x = x + x.T` yield different result.

``` python
from numba import jit
import numpy as np

@jit 
def self_addition1(x):
    x += x.T
    return x

@jit
def self_addition2(x):
    x = x + x.T
    return x

x = np.array([[1,2],[3,4]])
y = np.array([[1,2],[3,4]])

self_addition1(x)
> array([[2, 5],[8, 8]])

self_addition2(y)
> array([[2, 5],[5, 8]])
```

This behaviour don't happen when I remove the numba decorator

guoqiangqi:
The result of array inpalce binop(such as+=`) is ircorrect since the array transpose data are sharing memory with original array, which cased the transposed array data changed during the inplace binop calculation. For example (provided in #6949):
```
@jit 
def self_addition(x):
    x += x.T
    return x

x = np.array([[1,2],[3,4]])

self_addition(x)
> array([[2, 5],[8, 8]])
```
In this case, `x` will be temporarily changed to `[[2,5], [3,4]]` since the first row will be calculated first when matrix addition is performed. Therefore, `x.T` will be `[[2,3], [5,4]]`  (the correct value is `[[1,3], [2,4]]`) when the second row is calculated. Therefore, numba should allocate new memory in the stack to save array transpose data insteading of sharing memory with original array.

### DIFF
--- a/numba/core/lowering.py
+++ b/numba/core/lowering.py
@@ -1004,7 +1004,7 @@ class Lower(BaseLower):
             return self.lower_binop(resty, expr, expr.fn)
         elif expr.op == 'inplace_binop':
             lty = self.typeof(expr.lhs.name)
-            if lty.mutable and self.fndesc.noalias:
+            if lty.mutable:
                 return self.lower_binop(resty, expr, expr.fn)
             else:
                 # inplace operators on non-mutable types reuse the same
--- a/numba/core/lowering.py
+++ b/numba/core/lowering.py
@@ -1004,7 +1004,7 @@ class Lower(BaseLower):
             return self.lower_binop(resty, expr, expr.fn)
         elif expr.op == 'inplace_binop':
             lty = self.typeof(expr.lhs.name)
-            if lty.mutable:
+            if lty.mutable and self.fndesc.noalias:
                 return self.lower_binop(resty, expr, expr.fn)
             else:
                 # inplace operators on non-mutable types reuse the same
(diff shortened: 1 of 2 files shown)


gmarkall (changes requested):
Many thanks for the PR! Unfortunately I think there are a couple of issues with the approach.

Allocating space on the stack works for small data, but larger data exceeds the available stack. For example:

```python
from numba import njit
import numpy as np

x = np.zeros((4096, 4096))


@njit
def self_addition(x):
    x += x.T
    return x


print(self_addition(x))
```

segfaults:

```
$ python repro.py 
Segmentation fault (core dumped)
```

Even for small data, returning a pointer to stack-allocated data is not safe. For example:

```python
from numba import njit
import numpy as np

x = np.array([[1, 2], [3, 4]])


@njit
def get_transpose(x):
    xT = x.T
    return xT


xT = get_transpose(x)

print(xT)
```

gives:

```
$ python dangling_pointer.py 
[[140688323904944 140688323927200]
 [ 94515098062032               0]]
```

(the values are different every time because something else will overwrite the memory pointed to).

In general I think it will be a performance regression to always copy data to create a transpose - ideally a copy should only be made when data overlaps. If it is hard to determine when there is an overlap, it may be that inplace binary operations need to store their result into a temporary array until all results have been computed. It may be worth looking at how NumPy handles the operation `x += x.T`.

guoqiangqi:
@gmarkall Thanks for your advices, i would try to find a better implementation for this. I will ping you again when ready.

guoqiangqi:
Closed this PR since i have not found a good implementation to fix the noticed issue, anyone can feel free to proposal a new PR if they find a correct solution.

PR #7013, opened 2021-05-10
CUDA: Convert some intrinsics to use `@intrinsic`

gmarkall:
This PR converts `grid`, `gridsize`, and `syncthreads[_*]`, from implementations built on a stub, typing, and lowering, to implementations using the `@intrinsic` decorator.

Some additional tests are added to ensure that the casting behaviour for syncthreads functions that accept a predicate is as expected - for predicates narrower than 32 bits we would like them to be cast up - for predicates wider than 32 bits we'd like to present an error so that the predicate isn't truncated.

I plan to convert more of the implementations and add specific tests of the decorator, but this PR is for initial feedback before I proceed with translating too many - once this is approved I will continue with the rest of the translations.

### DIFF
--- a/numba/cuda/cudadecl.py
+++ b/numba/cuda/cudadecl.py
@@ -16,20 +16,6 @@ register_global = registry.register_global
 register_number_classes(register_global)
 
 
-class GridFunction(CallableTemplate):
-    def generic(self):
-        def typer(ndim):
-            val = ndim.literal_value
-            if val == 1:
-                restype = types.int32
-            elif val in (2, 3):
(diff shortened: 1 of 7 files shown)


gmarkall:
Closing this PR for now due to issues in `@intrinsic` for CUDA as identified above. The way forward will be to:

- Add some tests of `@intrinsic` for CUDA to capture what works for now, aiming to get into 0.54
- Resolve the issues that this PR exposes, and generally test `@intrinsic` with some use outside of Numba (e.g. in 3rd party extensions) before rewriting the core intrinsics like `cuda.grid()` using it.

PR #6991, opened 2021-05-03
Move ROCm target status to "unmaintained".

stuartarchibald:
This PR:

* Removes the ROCm target code.
* Fixes up from the above.
* Remove the ROCm target docs.
* Adds a deprecation/unsupported note.
* Adds a reminder to the 0.54 CHANGE_LOG entry.

~Depends on https://github.com/numba/numba/pull/6990.~ Rebased on top of this.

PR #7024, opened 2021-05-13
fixed llvmir usage in create_module method

### LINKED ISSUE #7022: ir not imported
https://github.com/numba/numba/blob/d4d09f378c471b702c5ec0f9a9f0f32a4637f6c7/numba/core/base.py#L1132 Can not find the ir module in the above file.

rawwar:
fixes #7022 

using llvmir instead of ir in `create_module` method of BaseContext

### DIFF
--- a/numba/core/base.py
+++ b/numba/core/base.py
@@ -1129,7 +1129,7 @@ class BaseContext(object):
     def create_module(self, name):
         """Create a LLVM module
         """
-        return ir.Module(name)
+        return llvmir.Module(name)
 
     @property
     def active_code_library(self):
--- a/numba/core/base.py
+++ b/numba/core/base.py
@@ -1130,7 +1130,7 @@ class BaseContext(object):
         """Create a LLVM module
         
         The default implementation in BaseContext always raises a
-        ``NotImplementedError`` exception. SubClass should implement 
+        ``NotImplementedError`` exception. Subclass should implement 
         this method.
         """
         raise NotImplementedError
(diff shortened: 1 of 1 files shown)


sklam:
Makes me wonder if that method is actually dead code. Why is it never exercised?

rawwar:
I see this being used in - https://github.com/numba/numba/blob/d4d09f378c471b702c5ec0f9a9f0f32a4637f6c7/numba/roc/target.py#L134

gmarkall:
> Makes me wonder if that method is actually dead code. Why is it never exercised?

I believe it's dead, and all target contexts override it.

I also came across this when I turned on flake8 for `base.py` (in PR #7001).

> I see this being used in - ...

This may have bitrotted as the ROCm target is unmaintained - see https://github.com/numba/numba/pull/6991

rawwar:
Shall I remove this method from BaseContext?

gmarkall:
I'd be inclined to make it raise `NotImplementedError` instead, because subclasses of `BaseContext` should implement it. (Perhaps `BaseContext` should be an ABC, but I don't think it's worth going to the trouble of making it one for this PR).

rawwar:
Sure, i'll make it raise `NotImplementedError`

numba/core/base.py
```diff
@@ -1128,8 +1128,12 @@ def post_lowering(self, mod, library):
 
     def create_module(self, name):
         """Create a LLVM module
+        
+        The default implementation in BaseContext always raises a
+        ``NotImplementedError`` exception. Subclass should implement
```
stuartarchibald, line 1133:
```suggestion
        ``NotImplementedError`` exception. Subclasses should implement 
```

esc (approved):
Thanks for the patch @stuartarchibald @rawwar !

PR #7027, opened 2021-05-14
Fix nrt debug print

MegaIng:
This PR is a byproduct of me trying to fix generators. Without this PR I get output like this

```
*** NRT_Decref 140720103881220 [000002C1CB230770]
*** NRT_Incref 29778180 [000002C1CB230770] 
*** NRT_Decref 140720103881221 [000002C1CB230770] 
```

This is because the call to `printf` was essentially

```C
printf("*** NRT_Decref %zu [%p]\n", *ptr, ptr)
```
The problem here is that the type of `ptr` is `int8*`., meaning the a value of type `int8` was passed to `printf` when it expected a value of type `uint64`. This is undefined behaviour and apparently didn't break for others. I now changed it to be:

```C
printf("*** NRT_Decref %zu [%p]\n", *((uint64*) ptr), ptr)
```

which does work correctly for me.

At request of @gmarkall I also added a config variable 'NUMBA_DEBUG_NRT' to replace the global `_debug_print` in `nrtdynmod`

### DIFF
--- a/numba/core/config.py
+++ b/numba/core/config.py
@@ -176,6 +176,9 @@ class _EnvReloader(object):
         # (up to and including IR generation)
         DEBUG_FRONTEND = _readenv("NUMBA_DEBUG_FRONTEND", int, 0)
 
+        # Enable debug prints in nrtdynmod
+        DEBUG_NRT = _readenv("NUMBA_DEBUG_NRT", int, 0)
+
         # How many recently deserialized functions to retain regardless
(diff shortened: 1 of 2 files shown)


sklam:
Good catch on the `%zu` mismatch. Thanks!

stuartarchibald (approved):
Thanks for the patch!

sklam (approved):
LGTM!

PR #7030, opened 2021-05-14
Fixes missing typing_extensions dependency on gitpython

stuartarchibald:
As title.


stuartarchibald:
PR not needed, gitpython is now fixed.

PR #7031, opened 2021-05-14
Fix inliner to use a single scope for all blocks

### LINKED ISSUE #6969: Parfor reductions not working when overload uses inline='always' and parallel=True flags

sklam:
closes #6969
supersede #7025

### DIFF
--- a/numba/core/compiler_machinery.py
+++ b/numba/core/compiler_machinery.py
@@ -304,7 +304,7 @@ class PassManager(object):
                 else:  # CFG level changes rebuild CFG
                     internal_state.func_ir.blocks = transforms.canonicalize_cfg(
                         internal_state.func_ir.blocks)
-            # Check the func_ir only has exactly one Scope instance
+            # Check the func_ir has exactly one Scope instance
             if not legalize_single_scope(internal_state.func_ir.blocks):
                 raise errors.CompilerError(
                     f"multiple scope in func_ir detected in {pss}",
--- a/numba/core/compiler_machinery.py
+++ b/numba/core/compiler_machinery.py
@@ -295,12 +295,6 @@ class PassManager(object):
         if isinstance(pss, FunctionPass):
             enforce_no_dels(internal_state.func_ir)
 
-        # Check the func_ir only has exactly one Scope instance
-        if not legalize_single_scope(internal_state.func_ir.blocks):
-            raise errors.CompilerError(
-                f"multiple scope in func_ir detected in {pss}",
-            )
-
         if self._ENFORCING:
             # TODO: Add in self consistency enforcement for
             # `func_ir._definitions` etc
@@ -310,7 +304,11 @@ class PassManager(object):
                 else:  # CFG level changes rebuild CFG
                     internal_state.func_ir.blocks = transforms.canonicalize_cfg(
                         internal_state.func_ir.blocks)
-
+            # Check the func_ir only has exactly one Scope instance
+            if not legalize_single_scope(internal_state.func_ir.blocks):
+                raise errors.CompilerError(
+                    f"multiple scope in func_ir detected in {pss}",
+                )
         # inject runtimes
         pt = pass_timings(init_time.elapsed, pass_time.elapsed,
                           finalize_time.elapsed)
--- a/numba/core/compiler_machinery.py
+++ b/numba/core/compiler_machinery.py
@@ -6,7 +6,7 @@ from numba.core.compiler_lock import global_compiler_lock
 from numba.core import errors, config, transforms
 from numba.core.tracing import event
 from numba.core.postproc import PostProcessor
-from numba.core.ir_utils import enforce_no_dels
+from numba.core.ir_utils import enforce_no_dels, legalize_single_scope
 
 # terminal color markup
 _termcolor = errors.termcolor()
@@ -294,6 +294,12 @@ class PassManager(object):
         # emitted ir.Dels.
         if isinstance(pss, FunctionPass):
             enforce_no_dels(internal_state.func_ir)
+
+        # Check the func_ir only has exactly one Scope instance
+        if not legalize_single_scope(internal_state.func_ir.blocks):
+            raise errors.CompilerError(
+                f"multiple scope in func_ir detected in {pss}",
+            )
 
         if self._ENFORCING:
             # TODO: Add in self consistency enforcement for
--- a/numba/core/ir_utils.py
+++ b/numba/core/ir_utils.py
@@ -2292,19 +2292,17 @@ def fixup_var_define_in_scope(blocks):
                 scope.localvars.define(var.name, var)
 
 
-def transfer_scope(old_block, new_scope):
+def transfer_scope(block, scope):
     """Transfer the ir.Block to use the given ir.Scope.
     """
-    old_scope = old_block.scope
-    if old_scope is new_scope:
+    old_scope = block.scope
+    if old_scope is scope:
         # bypass if the block is already using the given scope
-        return old_block
+        return block
     # Ensure variables are defined in the new scope
-    var_dict = {}
     for var in old_scope.localvars._con.values():
-        new_var = new_scope.redefine(var.name, loc=var.loc)
-        var_dict[var.name] = new_var
+        if var.name not in scope.localvars:
+            scope.localvars.define(var.name, var)
     # replace scope
-    new_block = old_block
-    new_block.scope = new_scope
-    return new_block
+    block.scope = scope
+    return block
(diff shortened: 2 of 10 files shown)


numba/core/compiler_machinery.py
```diff
@@ -304,7 +304,11 @@ def debug_print(pass_name, print_condition, printable_condition):
                 else:  # CFG level changes rebuild CFG
                     internal_state.func_ir.blocks = transforms.canonicalize_cfg(
                         internal_state.func_ir.blocks)
-
+            # Check the func_ir only has exactly one Scope instance
```
stuartarchibald, line 307:
```suggestion
            # Check the func_ir has exactly one Scope instance
```

numba/core/ir_utils.py
```diff
@@ -2179,6 +2179,16 @@ def enforce_no_phis(func_ir):
...
 
 
+def legalize_single_scope(blocks):
+    """Check the given mapping of ir.Block for containing a single scope.
+    """
+    num_of_scope = len({blk.scope for blk in blocks.values()})
+    if num_of_scope == 1:
+        return True
+    else:
+        return False
```
stuartarchibald, line 2189:
```suggestion
    return len({blk.scope for blk in blocks.values()}) == 1
```

numba/core/ir_utils.py
```diff
@@ -2256,3 +2266,43 @@ def convert_code_obj_to_function(code_obj, caller_ir):
     # create the function and return it
     return _create_function_from_code_obj(fcode, func_env, func_arg, func_clo,
                                           glbls)
+
+
+def fixup_var_define_in_scope(blocks):
+    """Fixes the mapping of ir.Block to ensure all ir.Var are defined in all
+    scope.
```
stuartarchibald, line 2273:
```suggestion
    """Fixes the mapping of ir.Block to ensure all ir.Var within a given block
    are defined in the block's associated scope.
```

numba/core/ir_utils.py
```diff
@@ -2256,3 +2266,43 @@ def convert_code_obj_to_function(code_obj, caller_ir):
...
+    """
+    # Scan for all used variables
+    used_var = {}
+    for blk in blocks.values():
+        scope = blk.scope
+        for inst in blk.body:
+            for var in inst.list_vars():
+                used_var[var] = inst
+    # Note: not all blocks share a single scope even though they should.
+    # Ensure the scope of each block defines all used variables.
```
stuartarchibald, line 2285: Should the inverse also be checked? i.e. that there aren't variables in the scope(s) that don't exist in _some_ block. I'm not sure how likely it is for this issue to occur if at all. Am also not sure if this will end up causing more problems if there are cases where blocks are not sharing the same scope.

numba/core/compiler_machinery.py
```diff
@@ -304,7 +304,11 @@ def debug_print(pass_name, print_condition, printable_condition):
...
+            if not legalize_single_scope(internal_state.func_ir.blocks):
+                raise errors.CompilerError(
+                    f"multiple scope in func_ir detected in {pss}",
```
stuartarchibald, line 310: As `self.ENFORCING` is always `False` this never runs. Perhaps it's time to add in `numba.config.enforcing` or something?

stuartarchibald (commented):
Thanks for the patch and for fixing this, it cannot have been an easy thing to figure out! Few minor things to look at but otherwise looks good.

numba/core/compiler_machinery.py
sklam: I'm intentionally putting it in something that doesn't always run. This check is still failing sometimes in a few passes.

numba/core/ir_utils.py
sklam: It probably won't manifest as a problem but it is sign that the passes are not cleaning up properly. I am reluctant to add more checks for now as it is going to be a week long rabbit hole.

numba/core/ir_utils.py
sklam: It's actually making sure that all the scopes referenced in the function know about all the variables used in the function.

sklam: I tried to clarify it in the latest commit.

stuartarchibald: Thanks, latest comment looks good.

numba/core/ir_utils.py
stuartarchibald: Agree, until all passes conform/something else helps out with this, it's probably not a good use of time trying to debug it.

numba/core/compiler_machinery.py
stuartarchibald: ok, thanks for clarifying, that there is a way to run it for debug is what is important.

stuartarchibald (approved):

PR #7036, opened 2021-05-17
Add support for debug_print on cuda

### LINKED ISSUE #7035: `DEBUG_JIT` doesn't work for CUDA
See also #7034.

`config.DEBUG_JIT` it doesn't work for CUDA, giving errors like:

```
TypeError: Failed in cuda mode pipeline (step: native lowering) Type of #1 arg mismatch: i8* != i8 addrspace(4)*
```

This probably needs an appropriate implementation of `debug_print` for the CUDA target context.

If it is added, it should be recommended that kernels are only launched with a small number of threads (perhaps 1) when `DEBUG_JIT` is turned on.

MegaIng:
Closes #7035

Adds support for `DEBUG_JIT` output on the `cuda` target. 

I would like to add an interface to only output on specific threads, but I am not sure how to best design an interface for that.

### DIFF
--- a/numba/cuda/nvvmutils.py
+++ b/numba/cuda/nvvmutils.py
@@ -190,7 +190,7 @@ def declare_string(builder, value):
     return builder.call(conv, [charptr])
 
 
-def declare_vprint(lmod):
+def declare_vprintf(lmod):
     voidptrty = ir.PointerType(ir.IntType(8))
     # NOTE: the second argument to vprintf() points to the variable-length
     # array of arguments (after the format)
--- a/numba/cuda/printimpl.py
+++ b/numba/cuda/printimpl.py
@@ -1,8 +1,7 @@
 from functools import singledispatch
 from llvmlite import ir
-from numba.core import types, cgutils
+from numba.core import types
 from numba.core.imputils import Registry
-from numba.cuda import nvvmutils
 
 registry = Registry()
 lower = registry.lower
--- a/numba/cuda/printimpl.py
+++ b/numba/cuda/printimpl.py
@@ -57,8 +57,6 @@ def print_varargs(context, builder, sig, args):
     It dispatches to the appropriate 'print' implementations above
     depending on the detected real types in the signature."""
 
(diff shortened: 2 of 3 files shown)


gmarkall:
@MegaIng Looks like there's just a couple of flake8 issues to resolve to make CI pass.

MegaIng:
@gmarkall Yes, those were easy to fix. But I still want to include some kind of interface to limit the output to specific threads. There are a few problems I noticed here:

- As far as I can tell, there is no universal function to get the total threadID independent on the number of dimensions used. Since we have to generate the function at compile time, where I don't think we have the dimension count, I am already stuck.
- It would certainly be interesting to just generate the output for some specific (not (0,0,0)) Thread specified by the user. How would the users pass that?
  - It is a numba-developer facing debug option, so probably like many other debug options (e.g. `NUMBA_DEBUG_JIT` itself) as a environment option
  - But it is also function dependent, so it is in theory better as an keyword argument to `cuda.jit` or an attribute of `cuda.compiler.Dispatcher`. But this also feels weird.
 
Does someone else have an idea?

gmarkall:
@MegaIng many thanks for the PR, and the further thoughts. Since this option is most likely to be used when debugging generated code, I think it will be very useful even without a mechanism to limit it to specific threads - I'd imagine the workflow for debugging code generation could easily be based around launching a compiled kernel with a single thread.

To proceed, I'd suggest getting this over the line and merged without support for limiting output to a single thread or group of threads, and then working on a separate PR that adds support for limiting the output.

MegaIng:
@gmarkall Then this PR is finished for the moment.

gmarkall:
@MegaIng could you add some tests for this please?

MegaIng:
@gmarkall How would I structure those tests? There are no tests for CPU `DEBUG_JIT` either, so I probably should add them as well.

MegaIng:
@gmarkall Ok, found the `test_debug` file. I will see what I can do.

PR #7040, opened 2021-05-18
Add Github action to mark issues as stale

gmarkall:
This uses the [stale action](https://github.com/actions/stale).

The intended configuration is:

- At 1:30 every day, any issues that have been inactive for 30 days are:
  - Labelled stale, and
  - Have a comment posted explaining why they're stale and requesting an update or that the issue be closed.
- Labels pertaining to long-term items, feature requests, and good first / second issues do not get marked.
- Stale issues are never closed.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,20 @@
+name: 'Mark stale issues'
+on:
+  schedule:
+    - cron: '30 1 * * *'
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v3
+        with:
+          stale-issue-message: >
+            This issue is marked as stale as it has had no activity in the past
+            30 days. Please close this issue if no further response or action is
+            needed. Otherwise, please respond with any updates and confirm that
+            this issue still needs to be addressed.
+          stale-issue-label: 'stale'
+          exempt-issue-labels: 'good first issue,good second issue,feature_request,Blocked awaiting long term feature,Numba 1.0'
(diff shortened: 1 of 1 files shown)


esc:
@gmarkall thank you for looking into this. FWIW: some time ago we had attempted something similar:

https://github.com/numba/numba/pull/6225

IIRC we encountered some serious issues related to filtering of labels and selecting issues and PRs that are never tobe marked as stale. Further IIRC this was related to bugs within Github that seemed to only manifest when the number of labels was large (like in the Numba repo). Even further IIRC I reached out to Github about this issue, they actually responded several times mentioning that they could reproduce it and were working on it but have gone radio-silent since then.

gmarkall:
@esc thanks for the comments - #6225 looks like it's using something different - the Github Stale app based on Probot seems to be distinct from the Stale action that this uses. Do you have any idea if the bugs in Github were generic related to labels, or were they problems with that app?

(FWIW, I note that other projects such as the RAPIDS ones seem to be able to make use of the Stale action on which this PR is based, but I've no direct experience managing / running that)

sklam (commented):
I'm ok to merge it today so we can test it.

sklam (approved):

esc:
Would add:

* 'bug - documentation incorrect'
* 'bug - numerically incorrect'
* 'caching'
* 'discussion'
* 'doc
* 'installation'
* 'jitclass'
* 'llvm'
* 'question'

And what about the ones with a numerical prefix? Like '0 - Backlog'

esc (approved):
LGTM, let's get this in!

PR #7044, opened 2021-05-19
Fixes for LLVM 11

stuartarchibald:
As title. Adds passes to the cheap optimisation run that permit
vectorisation based optimisation in the full optimisation run.


esc (approved):
I have verified that these fixes are good using experimental builds of llvmdev and llvmlite.

stuartarchibald:
> I have verified that these fixes are good using experimental builds of llvmdev and llvmlite.

Thanks for the review and for testing this.

PR #7047, opened 2021-05-19
Support __hash__ for numpy.datetime64

guilhermeleobas:
This is a follow up PR of #6393 with implementation of `hash` for 32-bit systems.


numba/np/npdatetime.py
```diff
@@ -812,3 +814,29 @@ def _cast_to_timedelta(context, builder, val):
...
+        raise ValueError(msg)
+    return val
+
+
+@overload_method(types.NPTimedelta, '__hash__')
+@overload_method(types.NPDatetime, '__hash__')
+def ol_hash_npdatetime(inst):
+    if IS_32BITS:
+        def impl(inst):
+            hi = (np.int64(inst) & 0xffffffff00000000)
```
stuartarchibald, line 833:
```suggestion
            hi = (np.int64(inst) & 0xffffffff00000000) >> 32
```

guilhermeleobas: If I shift `hi` to the right, then, `f(TD('-4'))` fails

stuartarchibald: I think the types will need to be got exactly right, this might be quicker to do using `builder` or maybe `@njit(locals={})`.

guilhermeleobas: Hum... is there any particular reason for using `locals={}`?

stuartarchibald: Apologies, `{}` is indicative only, you'd need to populate it appropriately. It gives control of the type of variables. There's some examples in https://github.com/numba/numba/blob/master/numba/cpython/hashing.py#L127-L160

guilhermeleobas: Good to know! By the way, the PR is ready for review.

numba/core/typing/builtins.py
```diff
@@ -804,6 +804,15 @@ def typer(val):
...
                  return ty
+            elif isinstance(val, (types.NPDatetime, types.NPTimedelta)):
+                # Constructor cast from datetime-like, e.g.
+                # > np.int64(np.datetime64("2000-01-01"))
+                if ty.bitwidth == 64:
+                    return ty
+                else:
+                    msg = (f"Cannot cast {val} to {ty} as {ty} is not 64 bits "
+                           "wide.")
+                    raise errors.TypingError(msg)
```
guilhermeleobas, line 815: @stuartarchibald, NumPy allows casting datatypes to other integer types. Should Numba follow the same convention?

```python
from numba import njit
import numpy as np

@njit
def foo(a):
    return np.int8(a)

a = np.timedelta64('NaT')
print(foo.py_func(a))
```

numba/np/npdatetime.py
stuartarchibald: Thanks, I'll move the label.

stuartarchibald:
@esc please could you review this, it contains code I wrote in #6987 so it's probably best someone else takes a look. Thanks!

guilhermeleobas:
Here's the NumPy reference to compute `hash` functions: https://github.com/numpy/numpy/blob/main/numpy/core/src/multiarray/scalartypes.c.src#L3123-L3162

esc (approved):
It looks good to me. I was wondering why the cast from integer to DT/TD was not implemented until I realized it's not needed for implementing the hash.

PR #7049, opened 2021-05-19
Make NumPy random module use @overload_glue

stuartarchibald:
As title.

### DIFF
--- a/numba/cpython/randomimpl.py
+++ b/numba/cpython/randomimpl.py
@@ -224,10 +224,10 @@ def random_impl(context, builder, sig, args):
     res = get_next_double(context, builder, state_ptr)
     return impl_ret_untracked(context, builder, sig.return_type, res)
 
-@lower("np.random.random")
-@lower("np.random.random_sample")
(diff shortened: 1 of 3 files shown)


numba/cpython/randomimpl.py
```diff
@@ -1299,7 +1299,12 @@ def random_arr(context, builder, sig, args, typing_key=typing_key):
         arr = arrayobj._empty_nd_impl(context, builder, arrty, shapes)
 
         # ... and populate it in natural order
-        scalar_impl = context.get_function(typing_key, scalar_sig)
+        *mod, fname = typing_key.split('.')
+        np_func = getattr(np.random, fname)
```
sklam, line 1303: `mod` is unused and `np.random` is hardcode in `getattr`. Maybe there should be a assert to make sure `mod` is `np.random`

stuartarchibald: Good point, done in 332c999

sklam (approved):
LGTM

PR #7050, opened 2021-05-19
Add overload_classmethod

sklam:
- Add `overload_classmethod`
- Add `typeof()` on subclasses of `numba.core.types.Type`

### DIFF
--- a/numba/core/extending.py
+++ b/numba/core/extending.py
@@ -247,7 +247,7 @@ def overload_classmethod(typ, attr, **kwargs):
     Similar to ``overload_method``.
 
 
-    Here is an example implementing a classmethod on array types to call
+    Here is an example implementing a classmethod on the Array type to call
     ``np.arange()``::
 
         @overload_classmethod(types.Array, "make")
@@ -256,7 +256,7 @@ def overload_classmethod(typ, attr, **kwargs):
                 return np.arange(nitems)
             return impl
 
-    The above code will allow the following code to work in jit-compiled code::
+    The above code will allow the following to work in jit-compiled code::
 
         @njit
         def foo(n):
--- a/numba/core/extending.py
+++ b/numba/core/extending.py
@@ -232,6 +232,22 @@ def overload_method(typ, attr, **kwargs):
     return decorate
 
 
+def overload_classmethod(typ, attr, **kwargs):
+    from numba.core.typing.templates import make_overload_classmethod_template
+
+    def decorate(overload_func):
+        template = make_overload_classmethod_template(
+            typ, attr, overload_func,
+            inline=kwargs.get('inline', 'never'),
+            prefer_literal=kwargs.get('prefer_literal', False)
+        )
+        infer_getattr(template)
+        overload(overload_func, **kwargs)(overload_func)
+        return overload_func
+
+    return decorate
+
+
 def make_attribute_wrapper(typeclass, struct_attr, python_attr):
     """
     Make an automatic attribute wrapper exposing member named *struct_attr*
--- a/numba/core/extending.py
+++ b/numba/core/extending.py
@@ -245,6 +245,22 @@ def overload_classmethod(typ, attr, **kwargs):
 
 
     Similar to ``overload_method``.
+
+
+    Here is an example implementing a classmethod on array types to call
+    ``np.arange()``::
+
+        @overload_classmethod(types.Array, "make")
+        def ov_make(cls, nitems):
+            def impl(cls, nitems):
+                return np.arange(nitems)
+            return impl
+
+    The above code will allow the following code to work in jit-compiled code::
+
+        @njit
+        def foo(n):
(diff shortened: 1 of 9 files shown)


numba/typed/typeddict.py
```diff
@@ -209,7 +209,7 @@ def copy(self):
 
 
 # XXX: should we have a better way to classmethod
```
stuartarchibald, line 211:
```suggestion
```
we do, this PR!? :)

numba/core/extending.py
```diff
@@ -217,19 +235,18 @@ def take_impl(arr, indices):
...
 
-    return decorate
+def overload_classmethod(typ, attr, **kwargs):
+    """
+    A decorator marking the decorated function as typing and implementing
+    classmethod *attr* for the given Numba type in nopython mode.
+
+
+    Similar to ``overload_method``.
+    """
```
stuartarchibald, line 248: Should add this (and an example) to the high-level extending docs?

numba/typed/typedlist.py
```diff
@@ -444,7 +444,7 @@ def __repr__(self):
 
 
 # XXX: should we have a better way to classmethod
-@overload_method(TypeRef, 'empty_list')
+@overload_classmethod(ListType, 'empty_list')
```
stuartarchibald, line 447: I guess this means `List.empty()` can now written as an alias of this?

numba/typed/typedlist.py
```diff
@@ -444,7 +444,7 @@ def __repr__(self):
 
 
 # XXX: should we have a better way to classmethod
```
stuartarchibald, line 446:
```suggestion
```

numba/core/typing/typeof.py
```diff
@@ -94,6 +94,9 @@ def _typeof_type(val, c):
     if issubclass(val, np.generic):
         return types.NumberClass(numpy_support.from_dtype(val))
 
+    if issubclass(val, types.Type):
+        return types.TypeRef(val)
```
stuartarchibald, line 98: It'd might be useful to make the repr on `TypeRef` more verbose? Using `class MyArray(types.Array): pass`, 
this:
```python
print(typeof(MyArray(types.float64, 3, 'C', True)))
```
gives
```
typeref[readonly array(float64, 3d, C)]
```
Maybe not for this PR though?

stuartarchibald (commented):
Thanks for the patch, few small things to resolve else looks good.

numba/typed/typedlist.py
sklam: Yes, we can finally unify the spelling for Dict and List.

numba/core/typing/typeof.py
sklam: i made an issue to track this: https://github.com/numba/numba/issues/7055

numba/core/extending.py
sklam: Added docs and coverted `overload_method` and `overload_classmethod` to use sphinx autofunction in the high-level extending docs.

numba/typed/typeddict.py
sklam: Yes finally

numba/core/typing/typeof.py
stuartarchibald: thanks

numba/typed/typedlist.py
stuartarchibald: Great!

numba/core/extending.py
```diff
@@ -245,6 +245,22 @@ def overload_classmethod(typ, attr, **kwargs):
 
 
     Similar to ``overload_method``.
+
+
+    Here is an example implementing a classmethod on array types to call
```
stuartarchibald, line 250:
```suggestion
    Here is an example implementing a classmethod the Array type to call
```

numba/core/extending.py
```diff
@@ -217,19 +235,34 @@ def take_impl(arr, indices):
...
+    ``np.arange()``::
+
+        @overload_classmethod(types.Array, "make")
+        def ov_make(cls, nitems):
+            def impl(cls, nitems):
+                return np.arange(nitems)
+            return impl
+
+    The above code will allow the following code to work in jit-compiled code::
```
stuartarchibald, line 259:
```suggestion
    The above code will allow the following to work in jit-compiled code::
```

numba/core/extending.py
stuartarchibald: Thanks, I've checked the rendering locally.

numba/core/extending.py
sklam:
```suggestion
    Here is an example implementing a classmethod on the Array type to call
```

stuartarchibald (approved):
Thanks for the patch and fixes. Great to see this implemented as part of the `numba.extending` API.

PR #7056, opened 2021-05-25
Change prange conversion approach to reuse header block.

### LINKED ISSUE #7051: conditional is not evaluated correctly if parallel=True
## Reporting a bug

The self contained code sample is supplied both inline and as `bug.py` within the attached `bug.zip`.
[bug.zip](https://github.com/numba/numba/files/6512636/bug.zip)

The `else` path in a top level `if-else` in a `@njit(parallel=True)` compiled function with separate `prange`d for loops produces wrong results:
```python
import numpy as np
@njit(parallel=True)
def case_parallel(size,case):
    result = np.zeros((size,))
    if case == 1:
        for i in prange(size):
            result[i] += 1
    else:
        for i in prange(size):
            result[i] += 2
    return result.mean()
@njit(parallel=False)
def case_serial(size,case):
    result = np.zeros((size,))
    if case == 1:
        for i in prange(size):
            result[i] += 1
    else:
        for i in prange(size):
            result[i] += 2
    return result.mean()
for size in [1,10,100]:
    print(f'serial  (size={size:3}), should be 1.0: {case_serial(size=size,case=1)}')
    print(f'parallel(size={size:3}), should be 1.0: {case_parallel(size=size,case=1)}')
``` 
prints
```
serial  (size=  1), should be 1.0: 1.0
parallel(size=  1), should be 1.0: 3.0
serial  (size= 10), should be 1.0: 1.0
parallel(size= 10), should be 1.0: 3.0
serial  (size=100), should be 1.0: 1.0
parallel(size=100), should be 1.0: 3.0
```
As can be seen, this bug is not dependent on actual parallel computations.

I'm running numba 0.53.1 and numpy 1.19.4 on python 3.8.8 on Linux-3.10.0-1160.15.2.el7.x86_64-x86_64-with-glibc2.10

DrTodd13:
Resolves #7051

### DIFF
--- a/numba/parfors/parfor.py
+++ b/numba/parfors/parfor.py
@@ -2301,6 +2301,11 @@ class ConvertLoopPass:
                         and isinstance(inst.value, ir.Expr)
                         and inst.value.op == 'call'
                         and self._is_parallel_loop(inst.value.func.name, call_table)):
+                    # Here we've found a parallel loop, either prange or pndindex.
+                    # We create a parfor from this loop and then overwrite the contents
+                    # of the original loop header block to contain this parfor and then
+                    # a jump to the original loop exit block.  Other blocks in the
+                    # original loop are discarded.
                     body_labels = [ l for l in loop.body if
                                     l in blocks and l != loop.header ]
                     args = inst.value.args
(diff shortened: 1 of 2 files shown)


stuartarchibald (approved):
Thanks for the patch.

PR #6870, opened 2021-03-26
Add support for dispatcher retargeting using with-context

sklam:
Alternative to #6770 for more flexibility

See https://github.com/numba/numba/pull/6770#issuecomment-786887089

~An example that uses the current CUDA target: https://gist.github.com/sklam/1eb89424dda59c8c4a7cb1f1ed2385be~ (outdated)

*Update*

- common retargeting utils are now moved to `numba/core/retarget.py`
- a test that demonstrates the usage is added as `numba/tests/test_retargeting.py`

PR #7061, opened 2021-05-26
Add ndarray allocator classmethod

sklam:
as titled


sklam:
Here's my approval on b683418 to 49f1dd8 which are @stuartarchibald works merged in at 57ebd93

stuartarchibald:
@sklam Started a test of overloading `np.empty` and the allocator here: https://github.com/stuartarchibald/numba/commit/cec66fd5dcb267b6fcf724510a3976b94b3b2b25 It's failing as the function resolution ends up in the CPU target `np.empty` and as the `_allocate` `@overload_classmethod` returns a function returning a literal `1` it fails to lower. However it shouldn't have even got there, the `DPU` extends the `GPU` target class which should know nothing of CPU `np.empty`.

sklam:
@stuartarchibald, I have a hacky fix for the problem at https://github.com/sklam/numba/commit/1469da7245c817ec4a3a09c224986da7f4191271. The commit message has some description of the problems. 

Also, a sidenote that [target_override](https://github.com/numba/numba/blob/d0ca288f24eccff3ba1f86e38f0de07a717f8d4c/numba/core/target_extension.py#L30) will probably needs to be fuse with the target-stack in with-target-switch PR (https://github.com/numba/numba/pull/6870)

sklam:
Trying it on CI before I cleanup the fixes

sklam:
@stuartarchibald, this is ready for another look

sklam:
CI actually passed, but the UI is stuck

numba/core/errors.py
```diff
@@ -651,6 +651,16 @@ def __init__(self, exception):
         self.old_exception = exception
 
 
+class InternalTargetMismatchError(InternalError):
+    """For target mismatch error occurred internally within the compiler.
```
stuartarchibald, line 655:
```suggestion
    """For signalling a target mismatch error occurred internally within the compiler.
```

numba/core/typing/templates.py
```diff
@@ -1038,15 +1016,33 @@ class _OverloadAttributeTemplate(AttributeTemplate):
...
+        disp = dispatcher_registry[target_hw]
+        tgtctx = disp.targetdescr.target_context
+        tgtctx.refresh()
+        if builtin_registry in tgtctx._registries:
+            reg = builtin_registry
+        else:
+            # Pick a registry in which to install intrinsics
+            registries = iter(tgtctx._registries)
+            reg = next(registries)
+        return reg
```
stuartarchibald, line 1038: `_OverloadAttributeTemplate` and `_IntrinsicTemplate` both contain this logic. Perhaps factor out, not necessary for this PR though.

numba/np/ufunc/array_exprs.py
```diff
@@ -376,9 +376,10 @@ def _lower_array_expr(lowerer, expr):
             inner_sig_args.append(argty)
     inner_sig = outer_sig.return_type.dtype(*inner_sig_args)
 
+    flags = utils.ConfigStack().top_or_none() or compiler.Flags()
+    flags = flags.copy() # make sure it's a clone or a fresh instance
```
stuartarchibald, line 380:
```suggestion
    flags = utils.ConfigStack().top_or_none()
    flags = compiler.Flags() if flags is None else flags.copy() # make sure it's a clone or a fresh instance
```
does this save copying a fresh instance in the case TOS is None?

numba/tests/test_svml.py
```diff
@@ -181,62 +183,86 @@ class TestSVMLGeneration(TestCase):
...
+        testname = f"run_{postfix}"
+        setattr(cls, testname, run_template)
+
+        @unittest.skipUnless(not skipped, "Not implemented")
+        def test_runner(self):
+            ctx = mp.get_context("spawn")
+            q = ctx.Queue()
+            p = ctx.Process(target=type(self).mp_runner, args=[testname, q])
+            p.start()
+            p.join()
```
stuartarchibald, line 245: Maybe supply a `timeout` just in case?

numba/tests/test_target_extension.py
```diff
@@ -479,7 +469,9 @@ def cpu_foo():
 
         # DPU chooses the ol_my_func2 as it's most specific, and DPU subtracts
         # for addition, so 10 + x -> 10 - 7 -> 3
-        self.assertPreciseEqual(dpu_foo(), 3)
+        from numba.core.target_extension import target_override
+        with target_override('dpu'):
```
stuartarchibald, line 473: Is this needed? `dpu_foo` is `djit` decorated and so shouldn't need an override ctx?

numba/tests/test_target_extension.py
```diff
@@ -539,8 +531,10 @@ def ol_my_func_cuda(x):
         def dpu_foo():
             my_func(1)
 
-        with self.assertRaises(errors.TypingError) as raises:
-            dpu_foo()
+        from numba.core.target_extension import target_override
+        with target_override('dpu'):
```
stuartarchibald, line 535: Same query as above RE override ctx.

numba/tests/test_target_extension.py
```diff
@@ -682,6 +667,76 @@ def dpu_foo_cpu():
...
+                                                            align_u32)
+                return meminfo
+
+            from numba.core.typing import signature
+            mip = types.MemInfoPointer(types.voidptr)  # return untyped pointer
+            sig = signature(mip, allocsize, align)
+            return sig, codegen
+
+        @overload_classmethod(types.Array, '_allocate', target='dpu',
+                              nopython=True)
```
stuartarchibald, line 707:
```suggestion
                              jit_options={'nopython':True)
```
and the same change throughout this test, whilst accepted I think this should really be an error. This change needs making throughout to prevent objmode fallback on the dpu target when doing function resolution (not that it will fail as the function/target resolution call graph is correct, but should it need to be debugged in future).

numba/tests/test_target_extension.py
stuartarchibald: Related to this? https://github.com/numba/numba/pull/7061/files#diff-3a1fe479f55566dcfb700cf3ea134261b5bbec513f8433b2393bc3bd7ee46645R731 ?

numba/tests/test_target_extension.py
```diff
@@ -682,6 +667,76 @@ def dpu_foo_cpu():
         for msg in msgs:
             self.assertIn(msg, str(raises.exception))
 
+    def test_overload_allocation(self):
+        def cast_integer(context, builder, val, fromty, toty):
+            # XXX Shouldn't require this.
```
stuartarchibald, line 672: Noted. I guess this is due to the casts not being installed in the dpu target and the inheritance model etc not handling this as yet?

stuartarchibald (commented):
Thanks for the patch. I've given the patches that aren't mine a review. In general this looks good, there's a bit of refactoring to do and what I think is a bit of an issue with kwarg handling that needs looking at, however, neither of things things are blockers for merging. The rest of the comments are relatively minor things. Thanks again!

numba/core/typing/context.py
```diff
@@ -284,7 +284,33 @@ def core(typ):
...
+            if hw is not None:
+                hw_clazz = target_registry[hw]
+                if target_hw.inherits_from(hw_clazz):
+                    usable.append((temp_cls, hw_clazz, ix))
+
+        # sort templates based on target specificity
+        def key(x):
+            return target_hw.__mro__.index(x[1])
+
+        order = [x[0] for x in sorted(usable, key=key)]
```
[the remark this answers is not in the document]
sklam: agree. it's a postrelease task

numba/np/ufunc/ufuncbuilder.py
```diff
@@ -140,10 +140,23 @@ def store_overloads_on_success():
...
+                    # Compile
+                    args, return_type = sigutils.normalize_signature(sig)
+                    cres = compiler.compile_extra(typingctx, targetctx,
+                                                  self.py_func, args=args,
+                                                  return_type=return_type,
+                                                  flags=flags, locals=locals)
+
+                    # cache lookup failed before so safe to save
+                    self.cache.save_overload(sig, cres)
+
```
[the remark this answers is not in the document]
sklam: likely that or a mistake in merge conflict resolution

numba/core/typing/context.py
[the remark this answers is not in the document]
sklam: tracked in https://github.com/numba/numba/issues/7150

numba/core/typing/templates.py
sklam: tracked in https://github.com/numba/numba/issues/7151

numba/tests/test_target_extension.py
sklam: you are right. this isn't needed.

numba/tests/test_target_extension.py
sklam: yes, and i'm too lazily to put in the cast

sklam:
Thanks for the review. All comments should be addressed now.

numba/np/ufunc/ufuncbuilder.py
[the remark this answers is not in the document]
sklam: redundant code is removed

numba/tests/test_svml.py
```diff
@@ -181,62 +183,86 @@ class TestSVMLGeneration(TestCase):
...
+    @classmethod
+    def mp_runner(cls, testname, outqueue):
+        method = getattr(cls, testname)
+        try:
+            ok, msg = method()
+        except Exception:
+            msg = traceback.format_exc()
+            ok = False
+        else:
+            fmt = None
```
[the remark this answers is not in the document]
sklam: removed

numba/tests/test_svml.py
sklam: added a 10s timeout

numba/tests/test_target_extension.py
sklam: removed

numba/tests/test_target_extension.py
sklam: I changed all the `nopython=True` for all the `overload_*`

numba/core/typing/context.py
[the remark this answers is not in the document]
stuartarchibald: thanks

numba/core/typing/templates.py
stuartarchibald: thanks

numba/tests/test_target_extension.py
stuartarchibald: Postponing this as discussed OOB.

stuartarchibald (approved):

eric-wieser:
This unfortunately broke the version of #5886 vendored by `clifford`, but it shouldn't be too hard a fix.

stuartarchibald:
@eric-wieser thanks for the note, apologies for breaking it, guess it's using some internals that moved? Maybe someone from https://github.com/pygae/clifford/ could subscribe to the `RC_Testers` group (described here https://numba.discourse.group/t/numba-0-54-0-rc3/835/2) so as to pick up pings about RCs to test when they are available?

PR #7066, opened 2021-05-27
Fix #7065: Add expected exception messages for NumPy 1.20 to tests

### LINKED ISSUE #7065: Cudasim tests will fail when NumPy 1.20 is added to the test matrix
Not quite an issue yet as I don't think we officially support NumPy 1.20, but recording this here as it will need resolving. Running the CUDA testsuite with cudasim and NumPy 1.20 gives a couple of failures:

```
FAIL: test_incompatible_highdim (numba.cuda.tests.cudadrv.test_cuda_array_slicing.CudaArraySetting)
  File "/home/gmarkall/numbadev/numba/numba/cuda/tests/cudadrv/test_cuda_array_slicing.py", line 289, in test_incompatible_highdim
    self.assertIn(
AssertionError: 'could not broadcast input array from shape (2,3) into shape (35,)' not found in ["Can't assign 3-D array to 1-D self", 'could not broadcast input array from shape (2,3) into shape (35)']

FAIL: test_incompatible_shape (numba.cuda.tests.cudadrv.test_cuda_array_slicing.CudaArraySetting)
  File "/home/gmarkall/numbadev/numba/numba/cuda/tests/cudadrv/test_cuda_array_slicing.py", line 303, in test_incompatible_shape
    self.assertIn(
AssertionError: 'could not broadcast input array from shape (2,) into shape (5,)' not found in ["Can't copy sequence with size 2 to array axis 0 with dimension 5", 'cannot copy sequence with size 2 to array axis with dimension 5']

```

IIRC NumPy 1.20 will be supported for 0.54, so I'll put this in the 0.54 milestone.

gmarkall:
The exception messages generated in `test_incompatible_highdim` and
`test_incompatible_shape` vary slightly in NumPy 1.20, so these
variants are added to the list of expected messages.

Fixes #7065.

### DIFF
--- a/numba/cuda/tests/cudadrv/test_cuda_array_slicing.py
+++ b/numba/cuda/tests/cudadrv/test_cuda_array_slicing.py
@@ -291,7 +291,9 @@ class CudaArraySetting(CUDATestCase):
             container=[
                 "Can't assign 3-D array to 1-D self",  # device
                 "could not broadcast input array from shape (2,3) "
-                "into shape (35)",  # simulator
+                "into shape (35)",  # simulator, NP < 1.20
+                "could not broadcast input array from shape (2,3) "
+                "into shape (35,)",  # simulator, NP >= 1.20
             ])
 
     def test_incompatible_shape(self):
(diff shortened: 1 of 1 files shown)


stuartarchibald (approved):
Thanks for the patch.

stuartarchibald:
As this is impacts the CUDA simulator only, no need for a test on real hardware.

PR #6074, opened 2020-08-04
Add support for math.isclose() and numpy.isclose()

njriasan:
I am implementing support for np.isclose and math.isclose in response to issue #5977. I have currently added support for the numpy implementation but the math implementation has not started yet. My changes add a new function to arraymath.py and four new unit tests. I should have complete support for optional arguments, although there may be issues where I am not doing type conversions that should be allowed.

PR #7067, opened 2021-05-28
Implement np.isclose

guilhermeleobas:
As title.


jeertmans:
Hello,
A few months ago I also proposed an implementation of `allclose` and `isclose` but this was put in the PR Backlog milestone, see #6286. Maybe it would be worth merging my request with yours ? Or take the best one ?

gmarkall:
Is this still work-in-progress? @guilhermeleobas Do you plan to add this to the list of supported functions in the docs?

guilhermeleobas:
@gmarkall, I am waiting for a review from @stuartarchibald. The code for this function was copied from another PR (https://github.com/numba/numba/pull/4610)

stuartarchibald:
CC @jpivarski, xref https://github.com/numba/numba/pull/6074

jpivarski:
As promised in the meeting, here's my implementation for cross-checking:

```python
@numba.njit
def _isclose_item(x, y, rtol, atol, equal_nan):
    if numpy.isnan(x) and numpy.isnan(y):
        return equal_nan
    elif numpy.isinf(x) and numpy.isinf(y):
        return (x > 0) == (y > 0)
    elif numpy.isinf(x) or numpy.isinf(y):
        return False
    else:
        return abs(x - y) <= atol + rtol * abs(y)

@numba.extending.overload(numpy.isclose)
def isclose(a, b, rtol=1e-05, atol=1e-08, equal_nan=False):
    if (isinstance(a, numba.types.Array) and a.ndim > 0) or (
        isinstance(b, numba.types.Array) and b.ndim > 0
    ):
        def isclose_impl(a, b, rtol=1e-05, atol=1e-08, equal_nan=False):
            # FIXME: want to broadcast_arrays(a, b) here
            x = a.reshape(-1)
            y = b.reshape(-1)
            out = numpy.zeros(len(y), numpy.bool_)
            for i in range(len(out)):
                out[i] = _isclose_item(x[i], y[i], rtol, atol, equal_nan)
            return out.reshape(b.shape)

    elif isinstance(a, numba.types.Array) or isinstance(b, numba.types.Array):
        def isclose_impl(a, b, rtol=1e-05, atol=1e-08, equal_nan=False):
            return numpy.asarray(
                _isclose_item(a.item(), b.item(), rtol, atol, equal_nan)
            )

    else:
        def isclose_impl(a, b, rtol=1e-05, atol=1e-08, equal_nan=False):
            return _isclose_item(a, b, rtol, atol, equal_nan)

    return isclose_impl
```

The main difference is that @guilhermeleobas's implementation does many passes over the data (following NumPy's implementation) and mine does one pass, and is specialized for non-arrays if given non-arrays. It has a weakness, though: for correctness, my `a` and `b` would have to be broadcasted with

```python
a, b = numpy.broadcast_arrays(a, b)
```

before the `out.reshape(b.shape)` would be fully general. As it is, it managed to pass the tests because all of the tests have the larger shape as the second argument.

The tests can be made more general by adding the following two items:

```python
yield [atol, np.inf, -np.inf, np.nan], [0], kw
yield [atol, np.inf, -np.inf, np.nan], 0, kw
```

which are just reversing the argument order of the last two tests. Then my implementation fails (for lack of lowered `broadcast_arrays`, https://github.com/numba/numba/issues/4074#issuecomment-704691879) and I believe @guilhermeleobas's implementation would pass.

Meta-question: are the "single pass over arrays" or "specialized for non-array types" performance considerations significant?

guilhermeleobas:
@jpivarski, your implementation compiles WAY faster than the one I did. For reference, I am using this [script](https://gist.github.com/guilhermeleobas/52ac0e58dbdea324cf3c6861d276abb1) to benchmark both implementations.

Compare both implementations with:
```
IMPL=jim pytest -rs -sv --disable-warnings --tb=short -x -v --durations=0 a.py
IMPL=guilherme pytest -rs -sv --disable-warnings --tb=short -x -v --durations=0 a.py
```

jpivarski:
The compilation speed probably depends strongly on types—if the arguments to my `isclose` function are scalars, then it resolves to a couple of if-statements. If the arguments are arrays, my implementation is wrong because it lacks broadcasting. (I think the array functions you use don't broadcast, but they at least fail with the right error message when the shapes don't match.)

A "best of both worlds" might be to check for scalar arguments and drop to `_isclose_item` for that case, which is (I'm guessing) what is compiling so quickly. It should also run faster because it won't have to create array structs to evaluate scalars—I doubt LLVM is smart enough to compile that away, but again, I'm just guessing.

guilhermeleobas:
Although the current implementation seems correct (similar to NumPy ones), it takes a while to compile. I'll try to rewrite it based on @jpivarski implementation

guilhermeleobas:
@jpivarski, can you review the code? Once https://github.com/numba/numba/pull/7437 gets merged, one can remove the `broadcast_shapes` part.

guilhermeleobas:
@stuartarchibald or @gmarkall, can one of you folks review this PR when possible?

guilhermeleobas:
One cannot use `np.broadcast_shapes` inside `np.isclose` because the former is available only on NumPy >= (1, 20)

numba/np/arraymath.py
```diff
@@ -1165,6 +1167,75 @@ def impl(a):
...
+            return out.reshape(b.shape)
+
+    elif isinstance(a, types.Array) and isinstance(b, types.Array):
+        m = max(a.ndim, b.ndim)
+        tup_init = (0,) * m
+
+        def isclose_impl(a, b, rtol=1e-05, atol=1e-08, equal_nan=False):
+            # Broadcast arrays of different types - cannot use
+            # np.broadcast_arrays for that
+            # this can be replaced by np.broadcast_shapes once #7437
```
njriasan, line 1212: It looks like that PR is merged. https://github.com/numba/numba/pull/8003

Can you update this section?

numba/np/arraymath.py
```diff
@@ -1165,6 +1167,75 @@ def impl(a):
...
+            for i in range(len(out)):
+                out[i] = _isclose_item(x[i], y, rtol, atol, equal_nan)
+            return out.reshape(a.shape)
+
+    elif isinstance(a, types.Number) and isinstance(b, types.Array):
+        def isclose_impl(a, b, rtol=1e-05, atol=1e-08, equal_nan=False):
+            x = a
+            y = b.reshape(-1)
+            out = np.zeros(len(y), np.bool_)
+            for i in range(len(out)):
```
njriasan, line 1201: Should this be supported with parfors as well? It seems like this should be supported with parallel=True.

njriasan (commented):
Thanks @guilhermeleobas. I left some comments, but this looks pretty good.

numba/np/arraymath.py
njriasan: I see your comment in this PR. Could you update this comment to reflect that it cannot actually be used until the minimum Numpy version is increased?

numba/np/arraymath.py
```diff
@@ -1165,6 +1167,75 @@ def impl(a):
...
+            b_ = np.broadcast_to(b, tup)
+
+            out = np.zeros(len(a_), dtype=np.bool_)
+            for i, (av, bv) in enumerate(np.nditer((a_, b_))):
+                out[i] = _isclose_item(av.item(), bv.item(), rtol, atol,
+                                       equal_nan)
+            return np.broadcast_to(out, tup)
+
+    else:
+        def isclose_impl(a, b, rtol=1e-05, atol=1e-08, equal_nan=False):
```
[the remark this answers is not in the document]
guilhermeleobas: It looks like `np.isclose` supports complex numbers

```python
>>> import numpy as np
>>> print(np.isclose(2, 3j))
False
```

numba/np/arraymath.py
guilhermeleobas: So, you're saying replace `range` by `prange` when `parallel=True`?

apmasell (approved):

sklam (approved):

PR #7074, opened 2021-06-03
Change compile unit language in DWARF to C++

PokhodenkoSA:
Change DWARF language to `DW_LANG_C_plus_plus`.
It helps GDB to understand C++ mangled functions.

Code
```python
from numba import njit

@njit(debug=True)
def foo(a):
    b = a + 1
    print (b)

foo(123)
```
Run with breakpoints:
```bush
gdb --pyargs python br_chk.py
(gdb) break ...
(gdb) run
Stop or not?
```

&nbsp; | DW_LANG_Python | DW_LANG_C_plus_plus
-- | -- | --
`break   njit-basic.py:foo$241` | &nbsp; | +
`break foo$241` | + | +
`break   __main__::foo$241` | + | +
`break   njit-basic.py:__main__::foo$241` | &nbsp; | +

Get function info:
```bash
(gdb) info function foo
```

&nbsp; | DW_LANG_C_plus_plus | DW_LANG_Python
-- | -- | --
`info func foo` | `void   __main__::foo$241(long long);` | `void _ZN8__main__7foo$241Ex(void);`

@vlad-perevezentsev

### DIFF
--- a/numba/core/debuginfo.py
+++ b/numba/core/debuginfo.py
@@ -216,7 +216,7 @@ class DIBuilder(AbstractDIBuilder):
 
     def _di_compile_unit(self):
         return self.module.add_debug_info('DICompileUnit', {
-            'language': ir.DIToken('DW_LANG_Python'),
+            'language': ir.DIToken('DW_LANG_C_plus_plus'),
             'file': self.difile,
             'producer': 'Numba',
             'runtimeVersion': 0,


sklam:
I just give it a try and it is definitely nicer to see the C++ symbols demangled. 

@stuartarchibald, I don't know enough about DWARF or GDB to what to check.

stuartarchibald:
These tests might help as a base for writing something to assert the demangling works:
https://github.com/numba/numba/blob/b29e3072f6dea51a617e48847eb53b34d9a1182f/numba/tests/test_gdb.py

stuartarchibald:
@stuartarchibald note to self: see the impact this has in `radare2` support.

gmarkall (commented):
This is interesting - I feel a bit ambivalent about changing the language in that it's not strictly correct to claim that the source is C++, but I would be happy for demangling to work. One of this things that's a little infuriating about GDB is that when there's no debug info, it does demangle symbols correctly (because it doesn't know that the source was Python). A quick skim of the GDB source suggests that it doesn't have any Python-specific language support (and even if it did it would likely not be implemented with demangling for the Itanium ABI). However, this does suggest that there ought to be a way in GDB to set the demangling method (rather than it being hardcoded as part of language support).

stuartarchibald:
> @stuartarchibald note to self: see the impact this has in `radare2` support.

`r2` seems to just use the language entry for selecting a demangling scheme, this suggesting there's nothing to do when it comes to Python https://github.com/radareorg/radare2/blob/702e1c1adae8d297ce23fab71f7c6b3140d2cc39/libr/anal/dwarf_process.c#L1498-L1519. In practice `r2` seems to manage to demangle fine, I'd guess it maybe has built in Itanium demangle by default?

Given that Numba's source language is technically Python but generates symbols with Itanium mangling I think "lying" about the language by default may be beneficial if it means debugging tools can decode the symbols/it's useful for humans. I also couldn't see a way to set the demangling method in `gdb` either (man pages or looking at the source), it seems strongly linked to the declared language hint.

PokhodenkoSA:
I think the ideal would be to have something like `DW_LANG_Numba`. Reasons:
- Numba uses a subset of Python language
- Generated code is expected to be like generated from C++ sources (i.e. use mangling).

It is just an idea to share. DWARF and GDB support required and it is pretty long way.

PokhodenkoSA:
> However, this does suggest that there ought to be a way in GDB to set the demangling method (rather than it being hardcoded as part of language support).

+1 for splitting mangling and language.

gmarkall:
> I think the ideal would be to have something like DW_LANG_Numba

I like this idea, and agree with your reasoning :-)

stuartarchibald:
I'd also be very pleased to have `DW_LANG_Numba` recognised in DWARF and gdb. However, given that this is a long way off (as noted above) I'm still inclined to go with this PR as an improvement on the current situation. Any further comments on this? If not we can start working out how to test it, probably a simple check of the LLVM IR would be reasonable for now.

PokhodenkoSA:
> If not we can start working out how to test it, probably a simple check of the LLVM IR would be reasonable for now.

In numba-dppy we have to use `DW_LANG_C_plus_plus` for its pipeline and [test](https://github.com/IntelPython/numba-dppy/pull/434/commits/9738e9c3dfc90de78702df48c71af7723b3671b3) it via LLVM IR.
I would like to see examples for testing LLVM IR for numba.

stuartarchibald:
I've pulled this patch into https://github.com/numba/numba/pull/7116, which also fixes #7103. #7116 contains a test for this patch and also tests for more general debug info emission.

stuartarchibald:
Am closing this as the patch is in #7116 along with tests. Thanks for proposing it @PokhodenkoSA.

PR #7082, opened 2021-06-04
Added Add/Sub between datetime64 array and timedelta64 scalar

### LINKED ISSUE #1202: Date arithmetic works when nopython=False, but not when nopython=True
`nopython=False`:

``` python
In [14]: from numba import jit, vectorize

In [15]: f = jit(NPDatetime('D')(NPDatetime('D')))(lambda x: x + np.timedelta64(1, 'D'))

In [16]: x = np.random.randint(10, size=10).astype('datetime64[D]')

In [17]: f(x)
Out[17]:
array(['1970-01-10', '1970-01-04', '1970-01-07', '1970-01-03',
       '1970-01-03', '1970-01-10', '1970-01-04', '1970-01-07',
       '1970-01-03', '1970-01-02'], dtype='datetime64[D]')
```

`nopython=True`:

```
UntypedAttributeError: Caused By:
  File "/Users/pcloud/miniconda/envs/py27/lib/python2.7/site-packages/numba/compiler.py", line 201, in run
    res = stage()
  File "/Users/pcloud/miniconda/envs/py27/lib/python2.7/site-packages/numba/compiler.py", line 415, in stage_nopython_frontend
    self.locals)
  File "/Users/pcloud/miniconda/envs/py27/lib/python2.7/site-packages/numba/compiler.py", line 710, in type_inference_stage
    infer.propagate()
  File "/Users/pcloud/miniconda/envs/py27/lib/python2.7/site-packages/numba/typeinfer.py", line 408, in propagate
    self.constrains.propagate(self.context, self.typevars)
  File "/Users/pcloud/miniconda/envs/py27/lib/python2.7/site-packages/numba/typeinfer.py", line 107, in propagate
    constrain(context, typevars)
  File "/Users/pcloud/miniconda/envs/py27/lib/python2.7/site-packages/numba/typeinfer.py", line 275, in __call__
    attrty = context.resolve_getattr(value=ty, attr=self.attr)
  File "/Users/pcloud/miniconda/envs/py27/lib/python2.7/site-packages/numba/typing/context.py", line 88, in resolve_getattr
    ret = attrinfo.resolve(value, attr)
  File "/Users/pcloud/miniconda/envs/py27/lib/python2.7/site-packages/numba/typing/templates.py", line 335, in resolve
    raise UntypedAttributeError(value=value, attr=attr)
UntypedAttributeError: Unknown attribute "timedelta64" of type Module(<module 'numpy' from '/Users/pcloud/miniconda/envs/py27/lib/python2.7/site-packages/numpy/__init__.pyc'>)

Failed at nopython (nopython frontend)
Unknown attribute "timedelta64" of type Module(<module 'numpy' from '/Users/pcloud/miniconda/envs/py27/lib/python2.7/site-packages/numpy/__init__.pyc'>)
```

njriasan:
Closes #1202.
Adds support for arithmetic between datetime64 array and timedelta64 scalar by supporting unit inference in numpy_support. Also fixes an associated parfor_pass issue where Numba tries to allocate the datetime64 dtype with a unit.


gmarkall:

njriasan:
> Thanks very much for the PR, I've queued it for review.

Thank you @gmarkall  I'm wondering if you think it will be possible to get this merged by 0.54. If it's not possible I completely understand.

gmarkall:
> Thank you @gmarkall I'm wondering if you think it will be possible to get this merged by 0.54. If it's not possible I completely understand.

@stuartarchibald What do you think? (asking as you added it to the backlog rather than a version milestone)

njriasan:
Is it possible to get this moved to a milestone so it can get reviewed? Perhaps Numba 0.55 or Numba 0.56?

gmarkall:
@njriasan I've moved it to the 0.55 milestone.

esc (approved):
I reviewed this by fixing the merge conflict with a merge from master and then I reverted the changes and observed the tests fail predictably.

numba/core/ir_utils.py
```diff
@@ -95,17 +95,24 @@ def mk_alloc(typemap, calltypes, lhs, size_var, dtype, scope, loc):
...
-        # empty doesn't like 'bool' sometimes (e.g. kmeans example)
-        dtype_str = 'bool_'
-    np_typ_getattr = ir.Expr.getattr(g_np_var, dtype_str, loc)
-    typ_var_assign = ir.Assign(np_typ_getattr, typ_var, loc)
+    # If dtype is a datetime/timedelta with a unit,
+    # then it won't return a valid type and instead can be created
+    # with a string. i.e. "datetime64[ns]")
+    if isinstance(dtype, (types.NPDatetime, types.NPTimedelta)) and dtype.unit != '':
+        typename_const = ir.Const(dtype_str, loc)
+        typ_var_assign = ir.Assign(typename_const, typ_var, loc)
```
stuartarchibald, line 109: Not tested?

numba/np/numpy_support.py
```diff
@@ -424,6 +425,17 @@ def make_specific(outputs, unit):
                     new_outputs.append(out)
             return new_outputs
 
+        def make_datetime_specific(outputs, dt_unit, td_unit):
+            new_outputs = []
+            for out in outputs:
+                if isinstance(out, types.NPDatetime) and out.unit == "":
+                    unit = npdatetime_helpers.combine_datetime_timedelta_units(
+                        dt_unit, td_unit)
```
stuartarchibald, line 433: What if unit ends up as `None` as there is no combined type? The result is the `types.NPDatetime()` ctor fails with a key error. Example:
```python
from numba import njit
import numpy as np

@njit
def foo(a, b):
    return a + b

x = np.array([np.datetime64("2011-01-01"),], dtype="datetime64[ns]")
# this is a UFuncTypeError in NumPy without setting a lax casting rule
print(foo.py_func(x, np.timedelta64(1000,'M')))
```

numba/np/numpy_support.py
```diff
@@ -481,9 +507,10 @@ def make_specific(outputs, unit):
             try:
                 inputs = choose_types(input_types, ufunc_inputs)
                 outputs = choose_types(output_types, ufunc_outputs)
-                # if the left operand or both are timedeltas, then the output
+                # if the left operand or both are timedeltas, or we have
+                # 1 datetime and 1 timedelta, then the output
```
stuartarchibald, line 511:
```suggestion
                # 1 datetime and 1 timedelta, then the output
```
the code implies the order is specific but this comment does not?

stuartarchibald (commented):
Thanks for the PR, I've given it a quick review and have noted a few things to take a look at. As mixing datetime/timedelta is tricky, I'd like to test more fully once those issues are fixed. Thanks again!

numba/np/numpy_support.py
njriasan: Thank you for adding this test. It seems clear to me based on the Numpy output that I should be throwing an exception in this case. I will update the code.

njriasan: I could use a little bit of feedback on how to best throw an exception in this case. I believe that in this situation we should cease compilation since this is a typing error that won't be resolved. However this is a bit less conservative and perhaps raising a Numba extended Error is safer.

However, the Numba extended error won't allow me to maintain the existing error message and I think the no supported implementation is very unclear.

However, the TypeError doesn't allow me to display the loc (and I also don't have enough information to determine it at this stage).

Do you have any preference or advice for the best way to handle this? I don't find either of these particularly satisfying, and I would prefer to be able to show both the code location and my error message.

numba/np/numpy_support.py
njriasan: Update the comment. Thank you.

numba/core/ir_utils.py
njriasan: Thank you. I added tests with parallel=True. I think I originally tested with these but forgot to update the testing suite.

numba/tests/test_npdatetime.py
```diff
@@ -823,5 +825,172 @@ def arr_div(a, b):
...
+    def test_add_td_no_match(self):
+        """
+        Tests that attempting to add a datetime64 and timedelta64
+        with types that cannot be cast raises a reasonable exception.
+        """
+        @njit
+        def arr_add(a, b):
+            return a + b
+
+        with pytest.raises(TypeError, match=re.escape("ufunc 'add' is not supported between datetime64[ns] and timedelta64[M]")):
```
gmarkall, line 972: You can use `self.assertRaisesRegex` here.

numba/tests/test_npdatetime.py
```diff
@@ -823,5 +825,172 @@ def arr_div(a, b):
...
+    def test_sub_td_no_match(self):
+        """
+        Tests that attempting to substract a datetime64 and timedelta64
+        with types that cannot be cast raises a reasonable exception.
+        """
+        @njit
+        def arr_sub(a, b):
+            return a - b
+
+        with pytest.raises(TypeError, match=re.escape("ufunc 'subtract' is not supported between datetime64[ns] and timedelta64[M]")):
```
gmarkall, line 988: Also can use `self.assertRaisesRegex`.

numba/tests/test_npdatetime.py
```diff
@@ -8,6 +8,8 @@
 import contextlib
 import itertools
 import warnings
+import pytest
+import re
```
gmarkall, line 12:
```suggestion
```

numba/np/numpy_support.py
stuartarchibald: @njriasan I don't think either option is ideal for the reasons noted. Given the current restrictions, I'd probably go with the `TypeError` as you've implemented, at least the reason is clear even if the location is not available. I think it's possible to wire the location through the stack from `typeinfer.py` (it's in the constraint object), but it's quite involved and likely invasive.

njriasan: Okay thank you. I think for this PR I'll opt with the `TypeError` handling.

numba/tests/test_npdatetime.py
```diff
@@ -823,5 +824,174 @@ def arr_div(a, b):
             self.assertTrue(np.array_equal(py_func(a, b), cfunc(a, b)))
 
 
+class TestDatetimeArrayOps(TestCase):
+
+    def test_add_td(self):
```
stuartarchibald, line 829: There's quite a lot of repetition throughout these test cases.

stuartarchibald (commented):
Thanks for the updates/fixes to this, patch in general looks good. I've added a patch at https://github.com/stuartarchibald/numba/commit/61bb07e1d2e655c4176b23231116d330e3418563 which should address the comments I've made in this review, it's essentially a refactor to remove the duplication in the test cases (and fix up missing `parallel=True`). If you are happy with the suggested patch please feel free to apply it (if not, we can discuss what needs to be done on the PR). Many thanks.

njriasan:
Thanks @stuartarchibald. The patch looked good to me, so I merged it.

stuartarchibald:
> Thanks @stuartarchibald. The patch looked good to me, so I merged it.

@njriasan Great, thanks for the merge.

@sklam / @esc please could one of you look at 8147eda as I wrote it (and also check this PR in general). Many thanks!

sklam:
https://github.com/numba/numba/commit/8147eda0686ed37eeba31c5c02c166e3b93c30c7 LGTM

numba/tests/test_npdatetime.py
```diff
@@ -823,5 +824,85 @@ def arr_div(a, b):
...
+        with types that cannot be cast raises a reasonable exception.
+        """
+        @njit
+        def impl(a, b):
+            return operation(a, b)
+
+        fname = operation.__name__
+        expected = re.escape((f"ufunc '{fname}' is not supported between "
+                              "datetime64[ns] and timedelta64[M]"))
+        with self.assertRaisesRegex(TypeError, expected):
```
sklam, line 893:
```suggestion
        with self.assertRaisesRegex((TypingError, TypeError), expected):
```

sklam (commented):
The following is needed to pass on old_style error

sklam (approved):
Thanks for the patch!

PR #7087, opened 2021-06-07
Add note to docs about zero-initialization of variables.

### LINKED ISSUE #7045: Silent variable creation
## Reporting a bug

- [x] I have tried using the latest released version of Numba (0.53.1).

```python
import numba as nb
import numpy as np

def enum_empty(values):
    # iterate over all values minus last one
    for i, v in enumerate(values[:-1]):
        pass
    print(i, v)

values = np.ones(1)
nb.njit(enum_empty)(values)
enum_empty(values)
```

In this example i want to get last value of `i`, but i didn't think at case when values have a size of one or less.
When you do that with only python you have an error, but with numba there is no warning/error... and you get `0` inplace

Thanks

stuartarchibald:
As title.

Fixes #7045

### DIFF
--- a/docs/source/reference/pysemantics.rst
+++ b/docs/source/reference/pysemantics.rst
@@ -71,4 +71,18 @@ with immutability assumption.  However, large global arrays are not copied to
 conserve memory.  The definition of "small" and "large" may change.
 
 
-.. todo:: This document needs completing.
+Zero initialization of variables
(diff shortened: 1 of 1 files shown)


esc (approved):
Looks good.

stuartarchibald:
Thanks for the review @esc

PR #7088, opened 2021-06-07
Initialize NUMBA_DEFAULT_NUM_THREADS with a batch scheduler aware value

### LINKED ISSUE #7086: Initialize NUMBA_DEFAULT_NUM_THREADS with a batch scheduler aware value
---

## Feature request

As it is `numba` initializes `NUMBA_DEFAULT_NUM_THREADS` with the number of cores of the machine with `multiprocessing.cpu_count()`.
When running a job through a batch scheduler (e.g., SLURM), not all cores might be useable (when the job request does not use an entire compute node).
In this case, using a scheduler aware number of cores sounds a better default, so one does not need to configure the number of threads to use when having requested only part of a machine.

[`multiprocessing.cpu_count`'s doc](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.cpu_count) suggests to use `len(os.sched_getaffinity(0))` ([os.sched_getaffinity](https://docs.python.org/3/library/os.html#os.sched_getaffinity)) when available.

t20100:
This PR updates the value of `NUMBA_DEFAULT_NUM_THREADS` to use a batch-scheduler aware number of CPU cores  through `len(os.sched_getaffinity(0))` when available.

closes  #7086

### DIFF
--- a/docs/source/user/threading-layer.rst
+++ b/docs/source/user/threading-layer.rst
@@ -268,10 +268,10 @@ API Reference
 
 .. py:data:: numba.config.NUMBA_DEFAULT_NUM_THREADS
 
-   The number of CPU cores on the system (as determined by
-   ``multiprocessing.cpu_count()``). This is the default value for
-   :obj:`numba.config.NUMBA_NUM_THREADS` unless the
-   :envvar:`NUMBA_NUM_THREADS` environment variable is set.
+   The number of usable CPU cores on the system (as determined by either
+   ``len(os.sched_getaffinity(0))`` or ``multiprocessing.cpu_count()``).
+   This is the default value for :obj:`numba.config.NUMBA_NUM_THREADS` unless
+   the :envvar:`NUMBA_NUM_THREADS` environment variable is set.
 
 .. autofunction:: numba.set_num_threads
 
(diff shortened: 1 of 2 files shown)


gmarkall:

stuartarchibald (commented):
Thanks for the patch. One minor thing to resolve else good to merge. Thanks again.

t20100:
I applied your patch, thanks.

stuartarchibald (approved):

stuartarchibald:
@t20100 Congratulations on your first contribution to Numba!

stuartarchibald:
Just for reference, this was tested as follows:
```bash
$ cat test_7088.py 
from numba import njit, prange, get_num_threads
import numpy as np

@njit(parallel=True)
def foo(n):
    acc = 0
    for x in prange(n):
        acc += x
    return acc

foo(100)
print(f"get_num_threads(): {get_num_threads()}")

$ python test_7088.py # default, 4 cores on this machine, all accessible
get_num_threads(): 4

$ taskset 0x00000003 python test_7088.py # restrict to cores #0 and #1.
get_num_threads(): 2
```

Issue #4971, opened 2019-12-16
what's the differences between numba.cuda.jit and numba.jit?


WhatAShot:
I'm new at numba. What's the difference between numba.cuda.jit and numba.jit?
If not, why not unify them?

stuartarchibald:
Thanks for the question. `numba.cuda.jit` targets NVIDIA CUDA supporting GPUs, `numba.jit` targets CPUs, `numba.roc.jit` targets AMD ROCm supporting GPUs. There are some elements of these targets which can be unified and at some point they will be, however, the programming model for GPU vs CPU is quite different, and the tool chains needed are also different (CPU uses a slightly customised LLVM from LLVM mainline, CUDA using NVIDIA's LLVM fork, ROCm has a special shim library https://github.com/numba/roctools built against the ROCm fork of LLVM) so total unification is not really feasible. For certain constrained cases, like `@vectorize` and `@guvectorize`, it's possible to use the same kernels on each target by simply changing the `target` kwarg http://numba.pydata.org/numba-doc/latest/reference/jit-compilation.html#vectorized-functions-ufuncs-and-dufuncs.

WhatAShot:
Thanks for replying and thanks for your contribution to communities.

Issue #2571, opened 2017-10-20
Array allocation for CPU and CUDA functions


philippeller:
I'm trying to write some functions that I can interchangeably use for different targets (cpu, cuda, parallel). The probelm I'm having is the allocation of a new array is different for CUDA code, e.g.:

`cuda.local.array(shape, dtype)`

vs. doing something similar for a CPU function, i.e.

`np.empty(shape, dtype)`

Is there a clever way how to deal with this without having to write separate functions?

philippeller:
I have found one dirty workaround for the problem. It is the only way that I could make it work.
Use the `@myjit` decorator instead of `@jit` and `@cuda.jit` and allocate all arrays as `cuda.local.array`.

    def myjit(f):
    '''
    f : function
    Decorator to assign the right jit for different targets
    In case of non-cuda targets, all instances of `cuda.local.array`
    are replaced by `np.empty`. This is a dirty fix, hopefully in the
    near future numba will support numpy array allocation and this will
    not be necessary anymore
    '''
    if target == 'cuda':
        return cuda.jit(f, device=True)
    else:
        source = inspect.getsource(f).splitlines()
        assert '@myjit' in source[0]
        source = '\n'.join(source[1:]) + '\n'
        source = source.replace('cuda.local.array', 'np.empty')
        exec(source)
        fun = eval(f.__name__)
        newfun = jit(fun, nopython=True)
        # needs to be exported to globals
        globals()[f.__name__] = newfun
        return newfun

I really hope this can be addressed in the future and that this issue is taken seriously. One strength of numba, and the main reason i use it in my research, is the ability to run the same code on CPU and GPU targets. The described feature is a severe limitation of this.

seibert:
This is definitely something we want to improve.  We added support for `@numba.jit` functions to automatically recompile to `@numba.cuda.jit(device=True)` a few releases back, but that feature only works if the function body can be compiled for both targets.  We'll be looking for more ways to make code work on both CPU and GPU, including this suggestion.

sgbaird:
A modified version of the workaround described by @philippeller (I'm pretty sure this was on stackoverflow, but I'm having trouble finding it). Converts `prange` to `range` as appropriate, allows for the function to already be indented by an arbitrary amount and allows for some CUDA and non-CUDA specific args. Still hackery, though.

```python
"""Decorator to assign the right jit for different targets."""
from numba import cuda, jit
from Lib import inspect

# HACK
def get_myjit(target="cuda",
[...]

PR #7107, opened 2021-06-11
CUDA: Fold the device dispatcher into the kernel dispatcher

gmarkall:
This PR removes the separate dispatcher for device functions (`DeviceDispatcher`) and uses the `Dispatcher` class for all CUDA functions requiring dispatch. With this change, `device=True` generally has no effect - all CUDA jitted functions are treated the same, and can be called from the host or from kernels. 

There is one exception to this - eager compilation of device functions still goes through a slightly different route, through the `compile_device()` function to create a `DeviceFunction()` instance - this is kept for now because it is needed to support the use case where someone is using Numba to compile Python to PTX device functions. However, a deprecation warning is added that advises the use of `compile_ptx()` so that it is possible in future for the `device` keyword to be completely redundant.

This PR is functionally complete but needs some tidying up and documentation adding - the main reason for making it now is to request a CUDA buildfarm run before assuming that no major changes are required - @esc / @stuartarchibald could this have a run please?

I also noticed various cases where `fastmath` behaviour is not as expected - I'll add xfailing tests along with the other tidy-ups.

### DIFF
--- a/numba/cuda/compiler.py
+++ b/numba/cuda/compiler.py
@@ -956,6 +956,36 @@ class Dispatcher(_dispatcher.Dispatcher, serialize.ReduceMixin):
         typingctx = cuda_target.typing_context
         typingctx.insert_user_function(dispatcher, function_template)
 
+    def get_call_template(self, args, kws):
+        # Copied and simplified from _DispatcherBase.get_call_template.
+        """
+        Get a typing.ConcreteTemplate for this dispatcher and the given
+        *args* and *kws* types.  This allows to resolve the return type.
+
+        A (template, pysig, args, kws) tuple is returned.
+        """
+        # Ensure an overload is available
+        self.compile_device(tuple(args))
+
+        # Create function type for typing
+        func_name = self.py_func.__name__
+        name = "CallTemplate({0})".format(func_name)
+
+        # The `key` isn't really used except for diagnosis here,
+        # so avoid keeping a reference to `cfunc`.
+        call_template = typing.make_concrete_template(
+            name, key=func_name, signatures=self.nopython_signatures)
+        pysig = utils.pysignature(self.py_func)
+
+        return call_template, pysig, args, kws
+
+    def get_overload(self, sig):
+        # NOTE: This dispatcher seems to be used as the key for the dict of
+        # implementations elsewhere in Numba, so we return this dispatcher
+        # instead of a compiled entry point as in
+        # _DispatcherBase.get_overload().
+        return self
+
     def configure(self, griddim, blockdim, stream=0, sharedmem=0):
         griddim, blockdim = normalize_kernel_dimensions(griddim, blockdim)
         return _KernelConfiguration(self, griddim, blockdim, stream, sharedmem)
@@ -1056,7 +1086,7 @@ class Dispatcher(_dispatcher.Dispatcher, serialize.ReduceMixin):
     @property
     def nopython_signatures(self):
         # Based on _DispatcherBase.nopython_signatures
-        return [kernel.signature for kernel in self.overloads.values()]
+        return [overload.signature for overload in self.overloads.values()]
 
     def specialize(self, *args):
         '''
--- a/numba/cuda/compiler.py
+++ b/numba/cuda/compiler.py
@@ -1231,6 +1231,10 @@ class Dispatcher(_dispatcher.Dispatcher, serialize.ReduceMixin):
         for _, defn in self.overloads.items():
             defn.inspect_types(file=file)
 
+    def inspect_ptx(self, args):
+        # XXX: Need to warn here
+        return self.compile_device(args).library.get_asm_str().encode()
+
     @property
     def ptx(self):
         return {sig: overload.ptx for sig, overload in self.overloads.items()}
--- a/numba/cuda/compiler.py
+++ b/numba/cuda/compiler.py
@@ -18,6 +18,7 @@ from numba.core.compiler_lock import global_compiler_lock
 from numba.core.compiler_machinery import (LoweringPass, PassManager,
                                            register_pass)
 from numba.core.dispatcher import OmittedArg
+from numba.core.errors import NumbaDeprecationWarning
 from numba.core.typed_passes import IRLegalization, NativeLowering
 from numba.core.typing.typeof import Purpose, typeof
 from warnings import warn
@@ -1231,8 +1232,17 @@ class Dispatcher(_dispatcher.Dispatcher, serialize.ReduceMixin):
         for _, defn in self.overloads.items():
             defn.inspect_types(file=file)
 
-    def inspect_ptx(self, args):
+    def inspect_ptx(self, args, nvvm_options={}):
         # XXX: Need to warn here
+        msg = ('inspect_ptx for device functions is deprecated. Use '
+               'compile_ptx instead.')
+        warn(msg, category=NumbaDeprecationWarning)
+
+        if nvvm_options:
+            msg = ('nvvm_options are ignored. Use compile_ptx if you want to '
+                   'set NVVM options.')
+            warn(msg, category=NumbaDeprecationWarning)
+
         return self.compile_device(args).library.get_asm_str().encode()
 
     @property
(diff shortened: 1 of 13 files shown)


gmarkall:
@stuartarchibald Could this have a CUDA buildfarm run before I go further with it please?

gmarkall:
Note: as discussed in gitter, signatures for device functions must still be usable, just without eager compilation - this is to force type conversion (e.g. floats to ints for bit operations). 

Reference: https://gitter.im/numba/numba?at=612f8d6cd9a141754dc92a50

gmarkall:
@MegaIng Does the test added in https://github.com/gmarkall/numba/commit/42a438b86edd8179ef281b9f2298b69b30041bd3 capture your use case adequately?

MegaIng:
@gmarkall Jup, that is quite similar to what I have (although I am actually directly using `uint8`, but that shouldn't make a difference).

Although, I am using `inline="always"`. Not sure if that makes a difference (or if it does anything at all).

gmarkall:
@MegaIng Many thanks for the quick reply!

> Jup, that is quite similar to what I have (although I am actually directly using uint8, but that shouldn't make a difference).

Ah - I assumed you were using something wider than `uint8` because of the `& 0xFF` - are those necessary in your use case but I've somehow elided the need for them? (or are they still necessary in my stripped-down example?)

> Although, I am using inline="always". Not sure if that makes a difference (or if it does anything at all).

It makes no difference on CUDA - device functions are always inlined under normal circumstances (they're only not inlined when you have `debug=True, opt=False` set on the decorator).

MegaIng:
> Ah - I assumed you were using something wider than uint8 because of the & 0xFF - are those necessary in your use case but I've somehow elided the need for them? (or are they still necessary in my stripped-down example?)

I think the `& 0xFF` is not necessary (unless the type conversion doesn't work). I think I added them while trying to not use a type signature (which failed because I was passing in floats, and at that point I stopped and asked on gitter)

MegaIng:
After a bit more thorough testing I can say that yes, when using `uint8`, the `& 0xFF` part is unnecessary.

sgbaird:
Seems related to at least #2571 and #4971

gmarkall:
> (just some generally related issues #2571 and #4971)

What is the link between this and #2571?

sgbaird:
> > (just some generally related issues #2571 and #4971)
> 
> What is the link between this and #2571?

Ah, good point.. I guess the only thing is that you can use `@cuda.jit` decorators instead of `@jit`, if I'm understanding correctly, but not that you can `numba.cuda` functions inside like `cuda.local.array`. This would still require either redefining the function or doing something a bit more "hacky" like in #2571. My bad

numba/cuda/compiler.py
```diff
@@ -943,6 +787,66 @@ def __init__(self, py_func, sigs, targetoptions):
...
+        pysig = utils.pysignature(self.py_func)
+
+        return call_template, pysig, args, kws
+
+    def get_overload(self, sig):
+        # NOTE: This dispatcher seems to be used as the key for the dict of
+        # implementations elsewhere in Numba, so we return this dispatcher
+        # instead of a compiled entry point as in
+        # _DispatcherBase.get_overload().
+        return self
```
gmarkall, line 848: This should be understood (or changed) before this PR is considered done:
* Why is `self` acceptable for `get_overload`?
* Why does the dispatcher and not the overload get used as the key for dicts in the CUDA target?

numba/cuda/compiler.py
```diff
@@ -943,6 +787,66 @@ def __init__(self, py_func, sigs, targetoptions):
...
+    def get_call_template(self, args, kws):
+        # Copied and simplified from _DispatcherBase.get_call_template.
+        """
+        Get a typing.ConcreteTemplate for this dispatcher and the given
+        *args* and *kws* types.  This allows to resolve the return type.
+
+        A (template, pysig, args, kws) tuple is returned.
+        """
+        # Ensure an overload is available
+        self.compile_device(tuple(args))
```
gmarkall, line 829: Need to check whether the explicit call to compile is necessary to get the call template. (My gut feel is that it is, so that the typing can be completed, but this should be checked).

numba/cuda/compiler.py
```diff
@@ -255,167 +255,11 @@ def compile_ptx_for_current_device(pyfunc, args, debug=False, lineinfo=False,
...
-                'docstring': pyfunc.__doc__
-            }
-            return info
-
-    typingctx = cuda_target.typing_context
-    typingctx.insert_user_function(dispatcher, device_function_template)
-    return dispatcher
-
-
 def compile_device(pyfunc, return_type, args, inline=True, debug=False,
```
gmarkall, line 258: Need to remove this function - it should no longer be used.

numba/cuda/compiler.py
```diff
@@ -943,6 +787,66 @@ def __init__(self, py_func, sigs, targetoptions):
             self.compile(sigs[0])
```
gmarkall, line 787: Eager compilation here is an issue when inferred typing returns something (because the compilation can only be for a device function. What should the rules be around eager compilation and whether to infer a kernel or device function?

numba/cuda/compiler.py
```diff
@@ -1123,6 +1026,40 @@ def compile(self, sig):
...
 
+    def compile_device(self, args):
+        """Compile the function for the given argument types.
+
+        Each signature is compiled once by caching the compiled function inside
+        this object.
+
+        Returns the `CompileResult`.
+        """
+        if args not in self.overloads:
```
gmarkall, line 1037: What is an overload was compiled for the `args`, but that compilation produced a kernel and not a device function?

gmarkall: Somewhere near to here, we need to enforce casting (or failure to compile) if a signature was provided.

numba/cuda/compiler.py
```diff
@@ -1123,6 +1026,40 @@ def compile(self, sig):
...
+            opt = self.targetoptions.get('opt', True)
+            debug = self.targetoptions.get('debug', False)
+            nvvm_options = {
+                'opt': 3 if opt else 0,
+                'debug': debug,
+            }
+
+            cres = compile_cuda(self.py_func, None, args, debug=debug,
```
gmarkall, line 1045: Why does the `compile_cuda` call not pass `device=True`? (Presumably because we never use the generated PTX, so it does not matter that NVVM doesn't produce a device function here?)

gmarkall: (in which case, it may not matter whether something got compiled as a kernel first then used as a device function (as per  comment above) - perhaps then we wouldn't need separate `compile` / `compile_device` functions, and they could be combined to do everything necessary for kernels and device functions without causing issues).

numba/cuda/compiler.py
```diff
@@ -1133,7 +1070,14 @@ def inspect_llvm(self, signature=None):
 
         '''
         if signature is not None:
-            return self.overloads[signature].inspect_llvm()
+            overload = self.overloads[signature]
+            if isinstance(overload, _Kernel):
```
gmarkall, line 1074: This if/else indicates there is a real distinction between the overloads for kernels and device functions, so combining them may be a problem - unless the `_Kernel` class becomes something more generic, like a `_DeviceFunction` (no judgment here made on how sensible / difficult that may be)

numba/cuda/compiler.py
```diff
@@ -1185,6 +1129,18 @@ def inspect_types(self, file=None):
         for _, defn in self.overloads.items():
             defn.inspect_types(file=file)
 
+    def inspect_ptx(self, args, nvvm_options={}):
```
gmarkall, line 1132: This can be removed now this isn't going into 0.55+, because it was already deprecated in 0.54.

gmarkall (commented):
Added some comments on the diff to note my current thoughts about what needs to be done to get this PR to a "ready to review" form. A couple of additional thoughts on my comments on the initial description:

* We no longer need to support eager compilation of device functions here because that was already deprecated in Numba 0.54 - however, we do need to preserve the use of a signature on a device function to preserve casting when calling functions (or send that through a further deprecation schedule) - a test case was added, `test_device_casting`, that captures this use case.
* There aren't issues with `fastmath` as far as I'm aware - it turned out that I was making some mistakes with my testing / experimentation.

PR #7114, opened 2021-06-14
CUDA: Deprecate eager compilation of device functions

gmarkall:
In future versions of Numba, the `device` kwarg to the `@cuda.jit` decorator will be obviated, and whether a device function or global kernel is compiled will be inferred from the context. With respect to kernel / device functions and lazy / eager compilation, four cases are presently handled:

1. `device=True`, eager compilation with a signature provided
2. `device=False`, eager compilation with a signature provided
3. `device=True`, lazy compilation with no signature
4. `device=False`, lazy compilation with no signature

The latter two cases can be differentiated without the `device` kwarg, because it can be inferred from the calling context - if the call is from the host, then a global kernel should be compiled, and if the call is from a kernel or another device function, then a device function should be compiled.

The first two cases cannot be differentiated in the absence of the `device` kwarg - without it, it will not be clear from a signature alone whether a device function or global kernel should be compiled. In order to resolve, this, support for eager compilation of device functions will be removed. Eager compilation with the `@cuda.jit` decorator will in future always imply the immediate
compilation of a global kernel.

This PR is made for Numba 0.54, in order to prepare for the changes in #7107 for 0.55.

### DIFF
--- a/docs/source/reference/deprecation.rst
+++ b/docs/source/reference/deprecation.rst
@@ -375,3 +375,56 @@ Schedule
 
 - In Numba 0.54: ``inspect_ptx()`` will be deprecated.
 - In Numba 0.55: ``inspect_ptx()`` will be removed.
+
+
+Deprecation of eager compilation of CUDA device functions
+=========================================================
+
+In future versions of Numba, the ``device`` kwarg to the ``@cuda.jit`` decorator
+will be obviated, and whether a device function or global kernel is compiled will
+be inferred from the context. With respect to kernel / device functions and lazy
+/ eager compilation, four cases are presently handled:
+
+1. ``device=True``, eager compilation with a signature provided
+2. ``device=False``, eager compilation with a signature provided
+3. ``device=True``, lazy compilation with no signature
+4. ``device=False``, lazy compilation with no signature
+
+The latter two cases can be differentiated without the ``device`` kwarg, because
+it can be inferred from the calling context - if the call is from the host, then
+a global kernel should be compiled, and if the call is from a kernel or another
+device function, then a device function should be compiled.
+
+The first two cases cannot be differentiated in the absence of the ``device``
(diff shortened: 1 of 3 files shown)


docs/source/reference/deprecation.rst
```diff
@@ -375,3 +375,56 @@ Schedule
...
+4. ``device=False``, lazy compilation with no signature
+
+The latter two cases can be differentiated without the ``device`` kwarg, because
+it can be inferred from the calling context - if the call is from the host, then
+a global kernel should be compiled, and if the call is from a kernel or another
+device function, then a device function should be compiled.
+
+The first two cases cannot be differentiated in the absence of the ``device``
+kwarg - without it, it will not be clear from a signature alone whether a device
+function or global kernel should be compiled. In order to resolve, this, support
```
stuartarchibald, line 400:
```suggestion
function or global kernel should be compiled. In order to resolve this, support
```

numba/cuda/decorators.py
```diff
@@ -104,6 +105,8 @@ def device_jit(func):
                                   debug=debug)
 
         if device:
+            msg = ("Eager compilation of device functions is deprecated")
```
stuartarchibald, line 108: Should this message also remind users that the eagerness is due to signature(s) being provided?

stuartarchibald (commented):

numba/cuda/decorators.py
gmarkall: Thanks for the suggestion - it does now.

stuartarchibald (approved):

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_72`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_73`.

Passed.

PR #7115, opened 2021-06-14
use np.asarray to convert arg to array

### LINKED ISSUE #7101: segmentation fault when calling np.all in ipython
## Reporting a bug

I tested this in the following envs
1. Ubuntu on VM box, - python 3.8 with numba stable and latest master version
2. Anaconda Docker image -with numba stable and latest master version 

This only happens when I am using `ipython` in linux. If I run the following code in `ipython` on windows or script, it works without any error.
```
from numba import njit
import numpy as np
@njit
def func():
     return np.all([[True,False],[True,True]])

```
Output:
```
In [2]: func()
Out[2]: True

Segmentation fault
(numbaenv) root@3080b57977ff:/home/numba_c/numba#
```

rawwar:
used `np.asarray` to convert arg to array

closes #7101

### DIFF
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -791,6 +791,8 @@ def np_all(a):
 @overload(np.any)
 @overload_method(types.Array, "any")
 def np_any(a):
+    if not type_can_asarray(a):
+        raise errors.TypingError('The argument "a" must be array-like')
     def flat_any(a):
         a = np.asarray(a)
         for v in np.nditer(a):
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -779,6 +779,7 @@ def array_argmax(context, builder, sig, args):
 @overload_method(types.Array, "all")
 def np_all(a):
     def flat_all(a):
+        a = np.asarray(a)
         for v in np.nditer(a):
             if not v.item():
                 return False
@@ -791,6 +792,7 @@ def np_all(a):
 @overload_method(types.Array, "any")
 def np_any(a):
     def flat_any(a):
+        a = np.asarray(a)
         for v in np.nditer(a):
             if v.item():
                 return True


stuartarchibald:
Thanks for the patch. This fixes the issue with `np.all` and `np.any` but I think the fundamental problem is still present, that `nditer` is accepting `list` as a valid input when it cannot handle it correctly. I've written a small starter patch to demonstrate one way of preventing the root cause:
```diff
diff --git a/numba/core/types/npytypes.py b/numba/core/types/npytypes.py
index 5267c03..ba38f7e 100644
--- a/numba/core/types/npytypes.py
+++ b/numba/core/types/npytypes.py
@@ -4,6 +4,7 @@ import warnings
 from llvmlite import ir
 
 from .abstract import DTypeSpec, IteratorType, MutableSequence, Number, Type
+from .scalars import Boolean
 from .common import Buffer, Opaque, SimpleIteratorType
 from numba.core.typeconv import Conversion
 from numba.core import utils
@@ -304,6 +305,9 @@ class NumpyNdIterType(IteratorType):
     def __init__(self, arrays):
         # Note inputs arrays can also be scalars, in which case they are
         # broadcast.
+        if not all([isinstance(x, (Number, Boolean, Array)) for x in arrays]):
+            msg = f"nditer only supports scalars and arrays, got: {arrays}"
+            raise TypeError(msg)
         self.arrays = tuple(arrays)
         self.layout = self._compute_layout(self.arrays)
         self.dtypes = tuple(getattr(a, 'dtype', a) for a in self.arrays)
```

numba/np/arraymath.py
```diff
@@ -791,6 +792,7 @@ def flat_all(a):
 @overload_method(types.Array, "any")
 def np_any(a):
     def flat_any(a):
```
stuartarchibald, line 794:
```suggestion
def np_any(a):
    if not type_can_asarray(a):
        raise errors.TypingError('The argument "a" must be array-like')
    def flat_any(a):
```
Suggest catching any type which cannot be converted to an array at typing time, and also testing this.

numba/np/arraymath.py
```diff
@@ -791,6 +792,7 @@ def flat_all(a):
...
+        a = np.asarray(a)
```
stuartarchibald, line 795: Would be good to add a test for this, perhaps with a list as per the original bug?

stuartarchibald (commented):
Thanks for the patch. The two comments I've made apply the same to both the changes in `np.all` and `np.any`. Once resolved this can be merged, but it'd be good to fix the root cause (as noted above) either in this PR or in a subsequent PR. Thanks again!

numba/np/arraymath.py
rawwar: Sure,  I'll add a test case. Thanks

stuartarchibald:
@rawwar thanks for the update, CI build is failing at `flake8`. Errors are:
```
numba/np/arraymath.py:844:15: F821 undefined name 'errors'
numba/np/arraymath.py:845:5: E306 expected 1 blank line before a nested definition, found 0
```
please could you take a look? Thanks!

PR #7119, opened 2021-06-15
Add support for `np.broadcast_to`

guilhermeleobas:
As title.


numba/np/arrayobj.py
```diff
@@ -1326,6 +1365,7 @@ def fancy_setslice(context, builder, sig, args, index_types, indices):
         index_shape = indexer.get_shape()
         src = make_array(srcty)(context, builder, src)
         # Broadcast source array to shape
+        breakpoint()
```
brandonwillard, line 1368:
```suggestion
```

numba/tests/test_array_manipulation.py
```diff
@@ -48,6 +48,10 @@ def numpy_array_reshape(arr, newshape):
     return np.reshape(arr, newshape)
 
 
+def numpy_broadcast_to(arr, shape):
+    return np.broadcast_to(arr, shape)
```
brandonwillard, line 52: Do these tests need a copy of `np.broadcast_to`?

brandonwillard (commented):
I just noticed that this PR isn't linked to the underlying issue (i.e. #7119).

numba/tests/test_array_manipulation.py
guilhermeleobas: What do you mean?

brandonwillard: It seems like one could replace `pyfunc` and `cfunc` below with something like
```python
pyfunc = np.broadcast_to
cfunc = jit(nopython=True)(lambda arr, shape: pyfunc(arr, shape))
```
then there's no need for a function like `numpy_broadcast_to`&mdash;unless I'm missing something.

fanshi118:
Has anyone tried accessing and returning the broadcasted dimensions via index? Encountered the following issue while doing so:

```python
import numba
import numpy as np

b = np.random.normal(size=(2,))
print(b)  # [-0.34972147  1.5952668 ]

@numba.njit
def dev_test(b):
    broadcast_shape = (2, 2)
    b = np.broadcast_to(b, broadcast_shape)
    return b[1]

new_b = dev_test(b)
print(new_b)  # [4.94065646e-324 2.25092138e-314]
```

When I returned `b`, I was able to get the correct output (in this case, two identical rows). However, if I were to return those broadcasted dimensions/indices, I'd get near-zero values.

guilhermeleobas:
Hi @fanshi118,

thanks for the test case, I can confirm it is a bug. No idea why this is happening but I'll investigate

numba/np/arrayobj.py
```diff
@@ -1303,6 +1303,76 @@ def _broadcast_to_shape(context, builder, arrtype, arr, target_shape):
     return new_arrtype, new_arr
 
 
+@intrinsic
+def _numpy_broadcast_to(typingctx, array, shape):
+    ret = types.Array(array.dtype, shape.count, array.layout, readonly=True)
```
gmarkall, line 1308: From the debugging call:

```suggestion
    ret = types.Array(array.dtype, shape.count, 'A', readonly=True)
```

This will prevent strides being ignored when indexing into the broadcasted array (which happens when the layout is `'C'`)

gmarkall (commented):
Suggestion added on the diff that we discussed on the debugging call - this at first glance appears to resolve the issue with indexing into broadcasted arrays.

numba/np/arrayobj.py
guilhermeleobas: Done! I've put both you and Brandon as co-authors in the commit. Hope that's ok

numba/np/arrayobj.py
```diff
@@ -1303,6 +1303,77 @@ def _broadcast_to_shape(context, builder, arrtype, arr, target_shape):
...
+        if src_dim == dest_dim or src_dim == 1:
+            src_index += 1
+            dest_index += 1
+        else:
+            raise ValueError('operands could not be broadcast together '
+                             'with remapped shapes')
+
+
+@overload(np.broadcast_to)
+def numpy_broadcast_to(array, shape):
```
stuartarchibald, line 1362: This also needs to ensure that `array` is an array (or array-like, seems like that's permitted).

numba/tests/test_array_manipulation.py
```diff
@@ -754,6 +762,74 @@ def _assert_raises(arr, val):
...
+            [(3,), (), TypingError, 'The argument "shape" must be a tuple or an integer.'],
+            [(3,), (1,), ValueError, 'operands could not be broadcast together with remapped shapes'],
+            [(3,), (2,), ValueError, 'operands could not be broadcast together with remapped shapes'],
+            [(3,), (4,), ValueError, 'operands could not be broadcast together with remapped shapes'],
+            [(1, 2), (2, 1), ValueError, 'operands could not be broadcast together with remapped shapes'],
+            [(1, 1), (1,), ValueError, 'input operand has more dimensions than allowed by the axis remapping'],
+            [(2, 2), (3,), ValueError, 'input operand has more dimensions than allowed by the axis remapping'],
+            [(1,), -1, ValueError, 'all elements of broadcast shape must be non-negative'],
+            [(1,), (-1,), ValueError, 'all elements of broadcast shape must be non-negative'],
+            [(1, 2), (-1, 2), ValueError, 'all elements of broadcast shape must be non-negative'],
```
stuartarchibald, line 814: line wrapping to 80 chars?

numba/tests/test_array_manipulation.py
```diff
@@ -754,6 +762,74 @@ def _assert_raises(arr, val):
...
+        ]
+        for orig_shape, target_shape, err, msg in data:
+            self.disable_leak_check()
```
stuartarchibald, line 817: Loop hoist this invariant?

numba/tests/test_array_manipulation.py
```diff
@@ -754,6 +762,74 @@ def _assert_raises(arr, val):
...
+            [np.ones(1), (1, 2, 3)],
+            [np.arange(3), (3,)],
+            [np.arange(3), (1, 3)],
+            [np.arange(3), (2, 3)],
+            # test if shape is not a tuple
+            [np.ones(0), 0],
+            [np.ones(1), 1],
+            [np.ones(1), 2],
+            # these cases with size 0 are strange, but they reproduce the behavior
+            # of broadcasting with ufuncs (see test_same_as_ufunc above)
```
stuartarchibald, line 786:
```suggestion
            # of broadcasting with ufuncs (see test_same_as_ufunc above)
```
is this comment relevant to Numba? (that test is in NumPy source?).

numba/np/arrayobj.py
```diff
@@ -1303,6 +1303,77 @@ def _broadcast_to_shape(context, builder, arrtype, arr, target_shape):
     return new_arrtype, new_arr
 
 
+@intrinsic
+def _numpy_broadcast_to(typingctx, array, shape):
+    layout = 'A'  # setting layout to 'A' for broadcast_to_shapes to use strides
+    ret = types.Array(array.dtype, shape.count, layout, readonly=True)
```
stuartarchibald, line 1309:
```suggestion
    # setting layout to 'A' for broadcast_to_shapes to use strides
    ret = array.copy(ndim=shape.count, layout='A', readonly=True)
```

numba/np/arrayobj.py
```diff
@@ -1303,6 +1303,77 @@ def _broadcast_to_shape(context, builder, arrtype, arr, target_shape):
...
+        src, shape_ = args
+        srcty = sig.args[0]
+
+        src = make_array(srcty)(context, builder, src)
+        shape_ = cgutils.unpack_tuple(builder, shape_)
+        _, dest = _broadcast_to_shape(context, builder, srcty, src, shape_,)
+
+        # Hack to get np.broadcast_to to return a read-only array
+        setattr(dest, 'parent', Constant.null(
+            context.get_value_type(dest._datamodel.get_type('parent'))))
```
stuartarchibald, line 1322:
```suggestion
                context.get_value_type(dest._datamodel.get_type('parent'))))
```

numba/tests/test_array_manipulation.py
```diff
@@ -754,6 +762,74 @@ def _assert_raises(arr, val):
...
+    def test_broadcast_to_indexing(self):
+        pyfunc = numpy_broadcast_to_indexing
+        cfunc = jit(nopython=True)(pyfunc)
+        data = [
+            [np.ones(2), (2, 2), (1,)],
+        ]
+        for input_array, shape, idx in data:
+            expected = pyfunc(input_array, shape, idx)
+            got = cfunc(input_array, shape, idx)
+            self.assertPreciseEqual(got, expected)
```
stuartarchibald, line 832: Should there be some testing of the "view" nature of the broadcasted array, i.e. that mutating the original mutates the broadcast? Examples:
```python
from numba import njit
import numpy as np

@njit
def foo(arr, shape):
    return np.broadcast_to(arr, shape)

x = np.ones(2)
ret = foo(x, (6, 2))
print(ret)

x[0] = 10
print(ret)
```
and
```python
from numba import njit
import numpy as np

@njit
def foo(arr, shape):
    tmp = np.broadcast_to(arr, shape)
    print(tmp)
    arr[0] = 10
    print(tmp)

x = np.ones(2)
ret = foo(x, (6, 2))
```

numba/np/arrayobj.py
```diff
@@ -1303,6 +1303,77 @@ def _broadcast_to_shape(context, builder, arrtype, arr, target_shape):
...
+    dest_ndim = len(dest_shape)
+    if src_ndim > dest_ndim:
+        raise ValueError('input operand has more dimensions than allowed '
+                         'by the axis remapping')
+    for size in dest_shape:
+        if size < 0:
+            raise ValueError('all elements of broadcast shape must be '
+                             'non-negative')
+
+    # based on _broadcast_onto function on numba/np/npyimpl.py
```
stuartarchibald, line 1342:
```suggestion
    # based on _broadcast_onto function in numba/np/npyimpl.py
```

stuartarchibald (commented):
Thanks for the patch, great to see this implemented. There's a few things to resolve, mostly in relation to accepted inputs and testing. Thanks again for working on this.

numba/tests/test_array_manipulation.py
guilhermeleobas: Not relevant to numba

numba/np/arrayobj.py
```diff
@@ -1303,6 +1303,85 @@ def _broadcast_to_shape(context, builder, arrtype, arr, target_shape):
...
+@overload(np.broadcast_to)
+def numpy_broadcast_to(array, shape):
+    if not type_can_asarray(array):
+        raise errors.TypingError('The first argument "array" must '
+                                 'be array-like')
+
+    if isinstance(shape, types.UniTuple) and not \
+            isinstance(shape.dtype, types.Integer):
+        raise errors.TypingError('The second argument "shape" must '
+                                 'be a Tuple[int]')
```
stuartarchibald, line 1369:
```suggestion
                                 'be a tuple of integers')
```

numba/tests/test_array_manipulation.py
```diff
@@ -754,6 +762,128 @@ def _assert_raises(arr, val):
...
+                np.zeros((1,)), (-1,), ValueError,
+                'all elements of broadcast shape must be non-negative'
+            ],
+            [
+                np.zeros((1, 2)), (-1, 2), ValueError,
+                'all elements of broadcast shape must be non-negative'
+            ],
+            [
+                np.zeros((1, 2)), (1.1, 2.2), TypingError,
+                'The second argument "shape" must be a Tuple[int]'
```
stuartarchibald, line 853:
```suggestion
                'The second argument "shape" must be a tuple of integers'
```

stuartarchibald (commented):
Thanks for the update. There's a few minor things left to resolve otherwise this looks good. I think the most important thing remaining is making sure that `array-like` arguments are handled correctly in the implementation given that type inference now permits them, this will also need testing. Thanks again!

numba/np/arrayobj.py
```diff
@@ -1303,6 +1303,85 @@ def _broadcast_to_shape(context, builder, arrtype, arr, target_shape):
...
+
+
+@overload(np.broadcast_to)
+def numpy_broadcast_to(array, shape):
+    if not type_can_asarray(array):
+        raise errors.TypingError('The first argument "array" must '
+                                 'be array-like')
+
+    if isinstance(shape, types.UniTuple) and not \
+            isinstance(shape.dtype, types.Integer):
```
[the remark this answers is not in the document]
guilhermeleobas: Done!

stuartarchibald (approved):
Thanks for the patch and fixes. Looks good!

PR #6928, opened 2021-04-13
Add support for axis keyword arg to numpy.argmax()

itamarst:
This adds support for the `axis` keyword argument to `np.argmax` and `np.ndarray.argmax`.

~I could not get arbitrary dimensions working, because `np.transpose()` only works (as far as I can tell) with literals. Suggestions are welcome.~

PR #7129, opened 2021-06-21
Add support for axis keyword argument to np.argmin()

itamarst:
**Note to reviewers:** this PR is actually much simpler than the line count would suggest, for the most part it's just very minor refactoring (refactoring internal functions so they're top-level functions) and reuses a lot of existing code from #6928. 

I have added inline comments noting where code is unchanged and merely moved.

### DIFF
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -632,77 +632,92 @@ def array_max(context, builder, sig, args):
     return impl_ret_borrowed(context, builder, sig.return_type, res)
 
 
-@lower_builtin(np.argmin, types.Array)
-@lower_builtin("array.argmin", types.Array)
-def array_argmin(context, builder, sig, args):
-    ty = sig.args[0].dtype
+@register_jitable
+def array_argmin_impl_datetime(arry):
+    if arry.size == 0:
+        raise ValueError("attempt to get argmin of an empty sequence")
+    it = np.nditer(arry)
+    min_value = next(it).take(0)
+    min_idx = 0
+    if _is_nat(min_value):
+        return min_idx
 
-    if (isinstance(ty, (types.NPDatetime, types.NPTimedelta))):
-        def array_argmin_impl(arry):
-            if arry.size == 0:
-                raise ValueError("attempt to get argmin of an empty sequence")
-            it = np.nditer(arry)
-            min_value = next(it).take(0)
-            min_idx = 0
-            if _is_nat(min_value):
-                return min_idx
-
-            idx = 1
-            for view in it:
-                v = view.item()
-                if _is_nat(v):
-                    if numpy_version >= (1, 18):
-                        return idx
-                    else:
-                        idx += 1
-                        continue
-                if v < min_value:
-                    min_value = v
-                    min_idx = idx
-                idx += 1
-            return min_idx
-
-    elif isinstance(ty, types.Float):
-        def array_argmin_impl(arry):
-            if arry.size == 0:
-                raise ValueError("attempt to get argmin of an empty sequence")
-            for v in arry.flat:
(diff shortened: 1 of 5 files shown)


numba/np/arraymath.py
```diff
@@ -632,77 +632,92 @@ def array_max_impl(arry):
...
-    if (isinstance(ty, (types.NPDatetime, types.NPTimedelta))):
-        def array_argmin_impl(arry):
-            if arry.size == 0:
-                raise ValueError("attempt to get argmin of an empty sequence")
-            it = np.nditer(arry)
-            min_value = next(it).take(0)
-            min_idx = 0
-            if _is_nat(min_value):
-                return min_idx
+@register_jitable
```
itamarst, line 635: This function's code is unchanged, it's simply moved to the top-level whereas previously it was internal to `array_argmin`.

numba/np/arraymath.py
```diff
@@ -632,77 +632,92 @@ def array_max_impl(arry):
...
-            idx = 0
-            for v in arry.flat:
-                if np.isnan(v):
-                    return idx
-                if v < min_value:
-                    min_value = v
-                    min_idx = idx
-                idx += 1
-            return min_idx
+@register_jitable
```
itamarst, line 661: This function's code is unchanged, it's simply moved to the top-level whereas previously it was internal to `array_argmin`.

numba/np/arraymath.py
```diff
@@ -632,77 +632,92 @@ def array_max_impl(arry):
...
+        if np.isnan(v):
+            return idx
+        if v < min_value:
+            min_value = v
+            min_idx = idx
+        idx += 1
+    return min_idx
+
 
+@register_jitable
```
itamarst, line 683: This function's code is unchanged, it's simply moved to the top-level whereas previously it was internal to `array_argmin`.

numba/np/arraymath.py
```diff
@@ -771,6 +786,50 @@ def array_argmax_impl_generic(arry):
     return max_idx
 
 
+def build_argmax_or_argmin_with_axis_impl(arr, axis, flatten_impl):
```
itamarst, line 789: This function's code is unchanged, it's simply moved to the top-level (whereas previously it was internal to `array_argmax`) so it can be used by both argmin and argmax implementations.

numba/np/arraymath.py
```diff
@@ -785,42 +844,9 @@ def array_argmax(arr, axis=None):
         def array_argmax_impl(arr, axis=None):
             return flatten_impl(arr)
     else:
-        _check_is_integer(axis, "axis")
```
itamarst, line 788: This code now lives in `build_argmax_or_argmin_with_axis_impl`, unchanged otherwise.

numba/np/arraymath.py
stuartarchibald: Confirmed, thanks.

numba/np/arraymath.py
stuartarchibald: Confirmed, thanks.

numba/np/arraymath.py
stuartarchibald: Confirmed, thanks.

numba/np/arraymath.py
stuartarchibald: Confirmed. Thanks.

numba/np/arraymath.py
stuartarchibald: Confirmed. Thanks.

stuartarchibald (approved):
Thanks for the patch, looks good! Thanks also for leaving notes to help with the review, they are much appreciated.

PR #7134, opened 2021-06-22
Move minimum LLVM version to 11.

stuartarchibald:
Since Aarch64 is now fixed for LLVM 11 this can be moved.


esc (approved):
LGTM

esc:
this is blocked on producing an `llvmlite` in `0.37.0dev0` which contains the correct version of `llvm` via https://github.com/numba/llvmlite/pull/715

esc:
Thanks for the patch!

PR #7137, opened 2021-06-23
skip pycc test on Python 3.7 + macOS because of distutils issue

esc:
This was detected when using Python 3.7 and NumPy 1.20 on OSX


stuartarchibald (approved):
Thanks for the patch.

PR #7146, opened 2021-06-25
Fix `inline_closurecall` may not be imported

sklam:
Mainline is failing because of this.

See error log at https://dev.azure.com/numba/numba/_build/results?buildId=9107&view=logs&jobId=0b09a2a2-0f6e-53e2-0b1f-738b2a0d11e7&j=0b09a2a2-0f6e-53e2-0b1f-738b2a0d11e7&t=e50a781e-75f1-573d-4144-a4055b22df26

This is likely caused by drifting in test slices that finally reveal the problem because the test is py3.9 only.


stuartarchibald (approved):
Thanks for the patch.

PR #7163, opened 2021-06-29
Split legalization pass to consider IR and features separately.

### LINKED ISSUE #7158: objmode error in IR legalization
## Reporting a bug

Looks like IR legalization is called on objmode IR, potentially throwing error for valid code. Is IR legalization pass needed for objmode IR in general?
```python
import numba
import numpy as np

@numba.njit
def f(A):
    with numba.objmode(B="int64[::1]"):
        B = A.view("int64")
    return B

f(np.ones(11, np.int64))
```
```
  File "/Users/ehsan/dev/numba/numba/core/typeinfer.py", line 1071, in propagate
    raise errors[0]
numba.core.errors.TypingError: Failed in nopython mode pipeline (step: Handle with contexts)
Failed in object mode pipeline (step: ensure IR is legal prior to lowering)
'view' can only be called on NumPy dtypes, try wrapping the variable 'A' with 'np.<dtype>()'

File "../tmp/objmode_ir_legal.py", line 7:
def f(A):
    <source elided>
    with numba.objmode(B="int64[::1]"):
        B = A.view("int64")
        ^

During: resolving callee type: type(ObjModeLiftedWith(<function f at 0x10180d040>))
During: typing of call at /Users/ehsan/dev/bodo/../tmp/objmode_ir_legal.py (6)


File "../tmp/objmode_ir_legal.py", line 6:
def f(A):
    with numba.objmode(B="int64[::1]"):
    ^
```

stuartarchibald:
This splits out legalization into:
* IR legalization
* Feature legalization for nopython mode and object mode.

It puts these into separate compiler passes and adds them to the
standard pipelines.

Adds test based on original reproducer.

Fixes #7158

### DIFF
--- a/numba/core/typed_passes.py
+++ b/numba/core/typed_passes.py
@@ -441,28 +441,13 @@ class NoPythonSupportedFeatureValidation(AnalysisPass):
     """NoPython Mode check: Validates the IR to ensure that features in use are
     in a form that is supported"""
 
-    _name = "no_python_supported_feature_validation"
+    _name = "nopython_supported_feature_validation"
 
     def __init__(self):
         AnalysisPass.__init__(self)
 
     def run_pass(self, state):
         raise_on_unsupported_feature(state.func_ir, state.typemap)
-        warn_deprecated(state.func_ir, state.typemap)
-        return False
-
-
-@register_pass(mutates_CFG=False, analysis_only=True)
-class ObjModeSupportedFeatureValidation(AnalysisPass):
-    """Object mode check: Validates the IR to ensure that features in use are
-    in a form that is supported"""
-
-    _name = "obj_mode_supported_feature_validation"
-
-    def __init__(self):
-        AnalysisPass.__init__(self)
-
-    def run_pass(self, state):
         warn_deprecated(state.func_ir, state.typemap)
         return False
 
--- a/numba/core/typed_passes.py
+++ b/numba/core/typed_passes.py
@@ -437,6 +437,37 @@ class NativeLowering(LoweringPass):
 
 
 @register_pass(mutates_CFG=False, analysis_only=True)
+class NoPythonSupportedFeatureValidation(AnalysisPass):
+    """NoPython Mode check: Validates the IR to ensure that features in use are
+    in a form that is supported"""
+
+    _name = "no_python_supported_feature_validation"
+
+    def __init__(self):
+        AnalysisPass.__init__(self)
+
+    def run_pass(self, state):
+        raise_on_unsupported_feature(state.func_ir, state.typemap)
+        warn_deprecated(state.func_ir, state.typemap)
+        return False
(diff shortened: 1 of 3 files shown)


numba/core/typed_passes.py
```diff
@@ -437,16 +437,45 @@ def run_pass(self, state):
 
 
 @register_pass(mutates_CFG=False, analysis_only=True)
-class IRLegalization(AnalysisPass):
+class NoPythonSupportedFeatureValidation(AnalysisPass):
+    """NoPython Mode check: Validates the IR to ensure that features in use are
+    in a form that is supported"""
 
-    _name = "ir_legalization"
+    _name = "no_python_supported_feature_validation"
```
sklam, line 444:
```suggestion
    _name = "nopython_supported_feature_validation"
```

we always refer to it as one word; else it reads like it not doing validation

numba/core/typed_passes.py
```diff
@@ -437,16 +437,45 @@ def run_pass(self, state):
...
         warn_deprecated(state.func_ir, state.typemap)
+        return False
+
+
+@register_pass(mutates_CFG=False, analysis_only=True)
+class ObjModeSupportedFeatureValidation(AnalysisPass):
+    """Object mode check: Validates the IR to ensure that features in use are
+    in a form that is supported"""
+
+    _name = "obj_mode_supported_feature_validation"
```
sklam, line 460:
```suggestion
    _name = "objmode_supported_feature_validation"
```
one word is better?

numba/core/typed_passes.py
```diff
@@ -437,16 +437,45 @@ def run_pass(self, state):
...
+
+    def __init__(self):
+        AnalysisPass.__init__(self)
+
+    def run_pass(self, state):
+        warn_deprecated(state.func_ir, state.typemap)
```
sklam, line 466: I don't think `warn_deprecated` is needed either. All the deprecated feature are nopython features, and `typemap` is all `pyobject`.

sklam (commented):
Thanks for the patch. I have a few comments.

numba/core/typed_passes.py
stuartarchibald: I think this is also likely true. Obj mode blocks can't return reflected containers and obj mode itself doesn't need to worry about it. Given this pass is now a NOP, remove it?

sklam: yes remove it

numba/core/typed_passes.py
stuartarchibald: indeed, fixed!

numba/core/typed_passes.py
stuartarchibald: done in 8c8222a

sklam (approved):
Thanks for the patch!

PR #7165, opened 2021-06-29
Fix use of np.clip where out is not provided.

stuartarchibald:
This handles the use of `None` being the default arg and the guard
code now permitting that.


sklam (approved):
Thanks for the patch!

PR #7167, opened 2021-06-30
Make aliasing work for tuple (re)construction.

### LINKED ISSUE #7135: Miscompiled parallel access to namedtuple

If I access an array in a parallel loop through a namedtuple, writes have no effect:

```python
import numpy as np
import numba

@numba.njit(parallel=True)
def foo(dat):
    n, = dat.x.shape
    for i in numba.prange(n):
        dat.x[i] = -1

x = np.ones(10)
Data = namedtuple("Data", ["x"])
dat = Data(x)

foo(dat)
dat
```
This should show an array with all values equal -1, but it is still filled with 1.

This does not happen if we access the namedtuple as a tuple, disable parallelization or move the array access into a non-inlined function:

```python
# works fine
@numba.njit(parallel=True)
def foo1(dat):
    n, = dat.x.shape

    x = dat.x
    for i in numba.prange(n):
        x[i] = -1


# works fine
@numba.njit(parallel=True)
def foo2(dat):
    n, = dat[0].shape

    for i in numba.prange(n):
        dat[0][i] = -1


@numba.njit(inline='never')
def bar(dat, i):
    dat.x[i] = -1


# works fine
def foo3(dat):
    n, = dat[0].shape
    
    for i in numba.prange(n):
        bar(dat, i)
```

DrTodd13:
Resolves #7135 

Previously, aliasing only worked when extracting something from a tuple.  In issue 7135, a tuple is passed to a parfor and because it is a tuple, we extract the individual parts and reconstruct the tuple within the gufunc since tuples can't be passed explicitly to gufuncs.  When the tuple is reconstructed, we need the elements that go into the tuple to alias with the tuple so that when elements are subsequently extracted that everything aliases and so the original arguments are known to be live and so that liveness propagates to the tuple and then elements extracted from it.  In short, this PR stops dead code removal from thinking that writes into parts of a reconstructed tuple are dead.

### DIFF
--- a/numba/core/ir_utils.py
+++ b/numba/core/ir_utils.py
@@ -816,6 +816,10 @@ def find_potential_aliases(blocks, args, typemap, func_ir, alias_map=None,
             if isinstance(instr, ir.Assign):
                 expr = instr.value
                 lhs = instr.target.name
+                lhstype = None
+                if lhs in typemap:
+                    lhstype = typemap[lhs]
+
                 # only mutable types can alias
                 if is_immutable_type(lhs, typemap):
                     continue
@@ -851,6 +855,12 @@ def find_potential_aliases(blocks, args, typemap, func_ir, alias_map=None,
                         _add_alias(lhs, expr.args[0].name, alias_map, arg_aliases)
                     if isinstance(fmod, ir.Var) and fname in np_alias_funcs:
                         _add_alias(lhs, fmod.name, alias_map, arg_aliases)
+                    if isinstance(lhstype, types.containers.BaseTuple):
+                        for used_var in expr.args:
+                            _add_alias(lhs, used_var.name, alias_map, arg_aliases)
+                        for used_var in expr.kws:
+                            assert(isinstance(used_var[1], ir.Var))
+                            _add_alias(lhs, used_var[1].name, alias_map, arg_aliases)
 
     # copy to avoid changing size during iteration
     old_alias_map = copy.deepcopy(alias_map)
--- a/numba/tests/test_parfors.py
+++ b/numba/tests/test_parfors.py
@@ -1786,6 +1786,19 @@ class TestParfors(TestParforsBase):
         x = TestNamedTuple3(y=np.zeros(10))
         self.check(test_impl, x, check_arg_equality=[comparer])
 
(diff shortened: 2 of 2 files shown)


gmarkall:
Adding to 0.55 as the linked issue has no milestone (though happy for it to be changed if it should be in 0.54)

numba/core/ir_utils.py
```diff
@@ -851,6 +853,12 @@ def find_potential_aliases(blocks, args, typemap, func_ir, alias_map=None,
                         _add_alias(lhs, expr.args[0].name, alias_map, arg_aliases)
                     if isinstance(fmod, ir.Var) and fname in np_alias_funcs:
                         _add_alias(lhs, fmod.name, alias_map, arg_aliases)
+                    if isinstance(lhstype, types.containers.BaseTuple):
```
ehsantn, line 856: Shouldn't this be limited to namedtuple construction only instead of any function call returning any type of tuple?

DrTodd13: I considered this issue but came to the conclusion that any function given a value and returning a tuple could put that value in the tuple.  I believe this is consistent with our general approach which has been to assume the worst case at function calls (we don't try to recursively analyze those calls to get a more accurate result right?).  I don't think it matters that in this case it is a namedtuple and in fact I think maybe this is not even generic enough because there are other containers in this module for which these rules might possibly apply.

ehsantn: Ok, makes sense. Looks like this is mostly used for dead code elimination and wouldn't affect common cases anyways.

The only other use is in array analysis with `get_canonical_alias`, where this change can potentially cause problems. For example, `A, B = f(C)` assigns aliases for A, B, C and the information is used to replace `A.attr` with `C.attr`. Am I understanding this correctly? If so, our "maybe alias" is used as "definitely alias" in this context which is problematic (a problem that the existing code has, but this PR makes it worse).

DrTodd13: If indeed what we have which is "maybe alias" is used a "definitely alias" then that is a problem.

aseyboldt:
@ehsantn @DrTodd13 Is this fix done? It looks to me as if this does fix the miscompilation issue.

Issue #7170, opened 2021-06-30
Parallel optimization breaks results in a very bad way


davrot:
Hi,

I found an example that is broken by the parallel=true & prange optimization. (broken version: numba_4_debug.py ; workaround with correct results in numba_4.py). It is a nasty problem because there is no indication for the user that such a dramatic error occurred.

The results need to be -- by definition of the problem -- in the value range 0 .... 9999. 
numba_4_debug.py this is slightly out of range but it is order of magnitudes faster :-) :

Result vector: 
[                   0 14685055086129564619 14685055086129564619 ...
                    0                    0                    0]

Somehow the optimization interacts with the initialization of the Result vector 
From the debug information (full info below):
Result = numpy.zeros((NumberOfIterations), dtype=numpy.uint64)---------------------------| #0

In the workaround I take the Result vector outside of the function that is numba.njit-ed and then the results are correct.
 
best wishes
David

[numba_4_debug.py.txt](https://github.com/numba/numba/files/6740586/numba_4_debug.py.txt)
[numba_4.py.txt](https://github.com/numba/numba/files/6740587/numba_4.py.txt)

Tested on Python 3.9.5 and Numba 0.53.1

Debug info:
================================================================================
 Parallel Accelerator Optimizing:  Function main,
/home/davrot/Talk/numba_4_debug.py (22)
================================================================================

Parallel loop listing for  Function main, /home/davrot/Talk/numba_4_debug.py (22)
---------------------------------------------------------------------------------------------|loop #ID
@numba.njit(numba.types.uint64[:](numba.types.uint64, numba.types.uint64), parallel=True)    |
def main(NumberOfIterations, NumerOfNeurons):                                                |
    Result = numpy.zeros((NumberOfIterations), dtype=numpy.uint64)---------------------------| #0
                                                                                             |
    for i in numba.prange(NumberOfIterations):-----------------------------------------------| #3
        H = numpy.random.random(NumerOfNeurons).astype(numpy.float64)------------------------| #1
        H /= numpy.sum(H)--------------------------------------------------------------------| #2
        Temp = GetSpike(H, NumerOfNeurons)                                                   |
        Result[i] = Temp                                                                     |
    return Result                                                                            |
--------------------------------- Fusing loops ---------------------------------
Attempting fusion of parallel loops (combines loops with similar properties)...
  Trying to fuse loops #0 and #3:
    - fusion succeeded: parallel for-loop #3 is fused into for-loop #0.
[...]

PR #7171, opened 2021-06-30
Use ir_utils.find_max_label in parfor gufunc construction.

### LINKED ISSUE #7105: Array returned from prange loop is wrong unless that array is read before the function returns.
- [ X ] I have tried using the latest released version of Numba (most recent is
- [ X] I have included a self contained code sample to reproduce the problem.

Hi,
When returning an array that is generated from a parallel loop using prange, the return value is wrong unless that array is read before the function returns. This seems to depend somewhat on the code executed within the loop before assigning the value to the array. But by trial and error I found various different combinations of code that triggers this problem.
This bug occurs for numba 0.53.1 and python 3.8.10 or 3.9.5. It does not occur for python 3.7.10.
this bug seems to be similar but not necessarily the same as #6962

```
import numpy as np
import math
from numba import jit, prange

def TestBad(TwoDPts, comboarr ):
    comboout = np.zeros(len(comboarr))
    for cindx in prange(0, len(comboarr)):     
        i0 = comboarr[cindx,0]
        i1 = comboarr[cindx,1]
        Pt1 = TwoDPts[i0]
        Pt2 = TwoDPts[i1]
        v = Pt1 - Pt2
        vl2 = v[0]+v[1]
        comboout[cindx] = vl2
    return comboout

def TestGood(TwoDPts, comboarr ):
    comboout = np.zeros(len(comboarr))
    comboout2 = np.zeros(len(comboarr))
    for cindx in prange(0, len(comboarr)):     
        i0 = comboarr[cindx,0]
        i1 = comboarr[cindx,1]
        Pt1 = TwoDPts[i0]
        Pt2 = TwoDPts[i1]
        v = Pt1 - Pt2
        vl2 = v[0]+v[1]
        comboout[cindx] = vl2
        
    for cindx in range(0, len(comboarr)):
      comboout2[cindx] = comboout[cindx]

    return comboout                    

TwoDPts = np.array([[-1., -1.],
       [-1.,  1.],
       [ 0.,  0.],
       [ 1., -1.],
       [ 1.,  0.],
       [ 1.,  1.]])

comboarr = np.array([[0, 1],
       [0, 2],
       [0, 3],
       [0, 4],
       [0, 5],
       [1, 2],
       [1, 3],
       [1, 4],
       [1, 5],
       [2, 3],
       [2, 4],
       [2, 5],
       [3, 4],
       [3, 5],
       [4, 5]])

outputbad = TestBad(TwoDPts, comboarr)
print("Bad", outputbad)
outputgood = TestGood(TwoDPts, comboarr)
print("Good", outputgood)

```
and the output is:

```
Bad [ 0.00000000e+00 -1.36311572e+57  0.00000000e+00 -1.36311572e+57
  0.00000000e+00 -1.36311572e+57  0.00000000e+00 -1.36311572e+57
  0.00000000e+00 -1.36311572e+57  0.00000000e+00 -1.36311572e+57
  0.00000000e+00 -1.36311572e+57  0.00000000e+00]

Good [-2. -2. -2. -3. -4.  0.  0. -1. -2.  0. -1. -2. -1. -2. -1.]
```

DrTodd13:
Resolves #7105 and #7170 

Previously, parfor_lowering incorrectly used the max of the parfor loop body labels to determine the starting point for new labelling.  This is wrong because the loop body can have nested parfors with their own labels.  I switched this to use the ir_utils function find_max_label which looks into the blocks and checks for labels in nested parfors.  This is all that is required to solve #7105.

In the process of debugging, my attention was drawn again to how we had some useless parallel re-processing of parfor gufunc IR.  There is no need to try to convert features to parfors or fuse parfors when we are compiling parfor gufuncs so I added some flags to turn that off.  We can't disable the parallel flag completely because sequential lowering of nested parfors is handled in the parfor pass as well.


stuartarchibald:
Thanks for the patch @DrTodd13. Thoughts on maybe splitting this so that the bug fix part (`find_max_label`) goes int 0.54RC3 (as it's an important fix) and the optimisation part goes into 0.55? I'm inclined to do it this way so as to keep the changes going into the RC as small as possible and for them to just be "fixes", this to minimise the potential of introducing new problems. Am happy to do the patch split work if you agree.

DrTodd13:
Sure, that's fine.  I don't mind.

stuartarchibald:
> Sure, that's fine. I don't mind.

Thanks @DrTodd13, https://github.com/numba/numba/pull/7230 carries the bug fix and preserves history. Once that is merged this can be rebased and the opt part can go in for 0.55. Thanks again!

stuartarchibald:
@DrTodd13 now #7230 has been merged in would you prefer to rebase this PR or open a new PR with the changes?

DrTodd13:
> @DrTodd13 now #7230 has been merged in would you prefer to rebase this PR or open a new PR with the changes?

I'd prefer rebase.  Do you want to do it or want me to attempt it?  This is just a rebase of the PR branch against Numba master?

stuartarchibald:
> > @DrTodd13 now #7230 has been merged in would you prefer to rebase this PR or open a new PR with the changes?
> 
> I'd prefer rebase. Do you want to do it or want me to attempt it? This is just a rebase of the PR branch against Numba master?

Rebase or merge of this PR against mainline should work fine, if you are happy to do it I'll add it to my review queue, thanks.

DrTodd13:
> > > @DrTodd13 now #7230 has been merged in would you prefer to rebase this PR or open a new PR with the changes?
> > 
> > 
> > I'd prefer rebase. Do you want to do it or want me to attempt it? This is just a rebase of the PR branch against Numba master?
> 
> Rebase or merge of this PR against mainline should work fine, if you are happy to do it I'll add it to my review queue, thanks.

Okay...I made an attempt at it...tests are running.

stuartarchibald:
I've taken a look at the remaining code in this patch (since #7230 extracted the bug fix) and I'm a little concerned about the use of a compilation flag option for what appears to just be signalling some internal state to shortcut some unnecessary compilation effort. Have I understood this correctly and is there perhaps a way to achieve the same without using the flag? Or does the notion of a `recursive` vs. `non-recursive` compiled `parallel=True` function have general applicability?

DrTodd13:
> I've taken a look at the remaining code in this patch (since #7230 extracted the bug fix) and I'm a little concerned about the use of a compilation flag option for what appears to just be signalling some internal state to shortcut some unnecessary compilation effort. Have I understood this correctly and is there perhaps a way to achieve the same without using the flag? Or does the notion of a `recursive` vs. `non-recursive` compiled `parallel=True` function have general applicability?

You understand correctly.  I did think to see if I could come up with another solution and couldn't think of one.  I don't think that what is done in the recursive mode is useful separately.  I'm open to other ideas for how to structure this better. @stuartarchibald

stuartarchibald:
As discussed OOB, moving this to the 0.56 milestone to allow more time to consider potential alternative solutions.

DrTodd13:
@stuartarchibald @sklam Closing this as the same effect (but even better) how now been integrated with the PR to split parfor processing into 3 phases.

PR #7177, opened 2021-07-02
Add debug infomation support based on datamodel.

stuartarchibald:
This:
1. Adds debug info for types that are based on `DataModel` by emitting information to make them appear as structures.
2. Fixes another bug in lineinfo emission where `decref` calls at the end of basic blocks were having line information associated with them from the variable declaration site.

Impact of 1. (and 2.!, to sort the lineinfo out) is demonstrated in the following:
```bash
$ cat debuginfo_structures.py
from numba import njit
import numpy as np

@njit(debug=True)
def foo():
    x = np.arange(21, dtype=np.float32).reshape((7, 3))
    y = (1, 2, 'x', 4)
    z = "some string" + str(10) + "some other string"
    return x, y, z

foo()
```
then running the above under `gdb`:
```bash
$ NUMBA_OPT=0 gdb -q  -ex 'b "_ZN8__main__7foo$241E"' -ex 'r' --args python debuginfo_structures.py 
<snip>

Breakpoint 1, __main__::foo$241 () at debuginfo_structures.py:4
4       @njit(debug=True)

(gdb) n
6           x = np.arange(21, dtype=np.float32).reshape((7, 3))
(gdb) n
7           y = (1, 2, 'x', 4)
(gdb) n
8           z = "some string" + str(10) + "some other string"
(gdb) n
9           return x, y, z

(gdb) info locals
x = {meminfo = 0x555556c4a150 "\001", parent = 0x0, nitems = 21, itemsize = 4, data = 0x555556c4a180, shape = {7, 3}, 
  strides = {12, 4}}
y = {f0 = 1, f1 = 2, f2 = 0x0, f3 = 4}
z = {data = 0x555556c4a540 "some string10some other string", length = 30, kind = 1, is_ascii = 1, hash = 18446744073709551615, 
  meminfo = 0x555556c4a510 "\001", parent = 0x0}

(gdb) whatis x
type = array(float32, 2d, C) ({i8*, i8*, i64, i64, float*, [2 x i64], [2 x i64]})

(gdb) whatis y
type = Tuple(Literal[int](1), Literal[int](2), Literal[str](x), Literal[int](4)) ({i64, i64, i8*, i64})

(gdb) whatis z
type = unicode_type ({i8*, i64, i32, i32, i64, i8*, i8*})

(gdb) x.nitems
$1 = 21

(gdb) x/21f x.data # we know there's 21 items in x and x is float32 from whatis above
0x555556c4a180: 0       1       2       3
0x555556c4a190: 4       5       6       7
0x555556c4a1a0: 8       9       10      11
0x555556c4a1b0: 12      13      14      15
0x555556c4a1c0: 16      17      18      19
0x555556c4a1d0: 20
(gdb) c

Continuing.
[Inferior 1 (process 4124) exited normally]
(gdb) q
```
Note that:
1. Calling `whatis` on the variables will yield their Numba typing domain type `repr` and also their datamodel layout in LLVM IR.
2. That the structures are behaving like C structs and the members can be inspected.


numba/tests/gdb_support.py
```diff
@@ -0,0 +1,137 @@
...
+        self._python = sys.executable
+        self._debug = debug
+        self._file_name = file_name
+        self._timeout = timeout
+        self._drive()
+
+    def _drive(self):
+        """This function sets up the caputured gdb instance"""
+        assert os.path.isfile(self._file_name)
+        cmd = [self._gdb_binary, '--interpreter', 'mi', '--args', self._python,
```
stuartarchibald, line 54: Note to self. Patch this to communicate over a new pty.

PR #7182, opened 2021-07-05
DEBUG READ THE DOCS

esc:
As title.

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -6,6 +6,7 @@ channels:
   - numba/label/dev
 dependencies:
   - python=3.7
+  - pip
   - llvmlite=0.37.0rc1
   - numpy
   - numpydoc
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -6,7 +6,7 @@ channels:
   - numba/label/dev
 dependencies:
   - python=3.7
-  - llvmlite
+  - llvmlite=0.37.0rc1
   - numpy
   - numpydoc
   - setuptools
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -2,7 +2,7 @@
 # See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
 
 # Required
-version: 7
+version: 2
 
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -4,6 +4,10 @@
 # Required
 version: 2
 
+# Confiugre the correct image
+build:
+  image: latest
+
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
   configuration: docs/source/conf.py
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -2,7 +2,7 @@
 # See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
 
 # Required
-version: 2
+version: 7
 
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -6,7 +6,7 @@ version: 2
 
 # Confiugre the correct image
 build:
-  image: latest
+  image: testing
 
 # Build documentation in the docs/ directory with Sphinx
 sphinx:


esc:
Strange error on RTD when trying to install the RC:

```
Warning: you have pip-installed dependencies in your environment file, but you do not list pip itself as one of your conda dependencies.  Conda may not use the correct pip to install your packages, and they may end up in the wrong place.  Please add an explicit pip dependency.  I'm adding one for you, but still nagging you.
Collecting package metadata: ...working... done
Solving environment: ...working... failed

ResolvePackageNotFound: 
  - llvmlite=0.37.0rc1 -> libstdcxx-ng[version='>=9.3.0'] -> __glibc[version='>=2.17']
  - llvmlite=0.37.0rc1 -> libgcc-ng[version='>=9.3.0']
 ```

File: docs/environment.yml [hunk 1]
```diff
@@ -6,7 +6,7 @@ channels:
   - numba/label/dev
 dependencies:
   - python=3.7
```
stuartarchibald, line 8:
```suggestion
  - python=3.7
  - pip
```
Pip is missing as noted by `conda`, not sure it's related to the build problem but could well be not helping?

esc: I have made the change but it did not alter any of the described symptoms.

esc:
I have requested support on this one from the kind people at Read the Docs via a support ticket.

astrojuanlu:
Hello folks, Juan Luis from Read the Docs here (and long user and fan of numba), I'm having a look at this issue. I see you have tried several things already, will focus on the conda error shown in @esc second commit here first. The best way to debug this will be to pull the https://github.com/readthedocs/readthedocs-docker-images corresponding to your build (most likely `stable`) and continue from there.

esc:
@astrojuanlu oh wow, when I called in support from Read The Docs, I was expecting to receive THAT kind of support! 😉

astrojuanlu:
Reproduced the conda error locally:

```
$ docker run -u root -it --rm --name rtd -v $(pwd):/src -p 8000:8000 readthedocs/build:stable bash
root@b54100be0026:/# cd /src/
root@b54100be0026:/src# cat environment.yml 
# This environment is used by the RTD config for PR builds. RTD uses this as the
# base environment and then adds in the sphinx etc tools on top.
# See: https://docs.readthedocs.io/en/stable/guides/conda.html
name: rtd
channels:
  - numba/label/dev
dependencies:
  - python=3.7
  - llvmlite=0.37.0rc1
  - numpy
  - numpydoc
  - setuptools
  # https://stackoverflow.com/questions/67542699/readthedocs-sphinx-not-rendering-bullet-list-from-rst-fileA
  - docutils==0.16
  # The following is needed to fix RTD.
  - conda
root@b54100be0026:/src# conda env create --quiet --name numba --file environment.yml
Collecting package metadata: ...working... done
Solving environment: ...working... failed

ResolvePackageNotFound: 
  - llvmlite=0.37.0rc1 -> libstdcxx-ng[version='>=9.3.0'] -> __glibc[version='>=2.17']
  - llvmlite=0.37.0rc1 -> libgcc-ng[version='>=9.3.0']

root@b54100be0026:/
```

I confirm that I am not able to reproduce this on my conda installation outside of Docker. RTD has a version that's 2 years old, I'm upgrading at the moment to see if it fixes the issue.

astrojuanlu:
I confirm that conda 4.10 solves the environment, while conda 4.6 doesn't. That's why switching to the `testing` image has no effect, because it still uses the same version. Still working on it.

astrojuanlu:
Good news: using mamba solves the environment without problems. @esc Do you want us to enable the `CONDA_USES_MAMBA` [feature flag](https://docs.readthedocs.io/en/latest/guides/feature-flags.html#available-flags) for https://readthedocs.org/projects/numba/ ? Then you can revert to 8e3ed82 and I think the problem would be solved.

esc:
> Good news: using mamba solves the environment without problems. @esc Do you want us to enable the `CONDA_USES_MAMBA` [feature flag](https://docs.readthedocs.io/en/latest/guides/feature-flags.html#available-flags) for https://readthedocs.org/projects/numba/ ? Then you can revert to [8e3ed82](https://github.com/numba/numba/commit/8e3ed82bfaac6f00b880394ce6810a080533fc3e) and I think the problem would be solved.

That is very interesting and maybe something for @wolfv -- I'll have to confer with the others about this. @astrojuanlu in any case, thank you very much for your support on this front!

wolfv:
this is probably related to virtual packages (https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-virtual.html#managing-virtual-packages)

The old conda doesn't add a virtual glibc package and thus the solve fails. 

Not sure if they use a newer conda for the mamba image.

wolfv:
If you want to enable the flag, you need to email rtd: https://github.com/patrickfuchs/buildH/issues/68#issuecomment-773365929

astrojuanlu:
> Not sure if they use a newer conda for the mamba image.

No we don't 😓 That's why resorting to mamba is the only solution on RTD.

jezdez:
> > Not sure if they use a newer conda for the mamba image.
> 
> No we don't 😓 That's why resorting to mamba is the only solution on RTD.

Do you take patches to upgrade Conda to 4.10.x? I see there is a comment about [memory usage in Conda 4.7.12.1](https://github.com/readthedocs/readthedocs-docker-images/blob/1ebe82ab639cadb5c544491fa5fbc97f0b0049c5/Dockerfile#L223-L224) and since then there were a number of performance improvements.

astrojuanlu:
I opened an issue about the conda version on the RTD Docker images here https://github.com/readthedocs/readthedocs-docker-images/issues/173

Unless we fast-track the upgrade, it might take us some time to discuss and arrive at a good decision. The short-term solution is still to enable the `CONDA_USES_MAMBA` for this project (it can be turned off later).

esc:
@astrojuanlu we are happy to set "use mamba" temporarily until the issues with a newer conda version have been sorted out.

astrojuanlu:
I just enabled mamba for the numba project on RTD, and looks like https://readthedocs.org/projects/numba/builds/14188031/ is working (triggered from #7193, feel free to close it).

esc:
Solved for now, closing!

astrojuanlu:
As a follow-up to this, we are about to deploy new images with Miniconda3 4.9.2, and I confirmed locally that this fixes the problem with the numba builds without having to use mamba. Let us know if you want us to disable mamba for you after our next deploy.

PR #7185, opened 2021-07-05
Add get_impl_key as abstract method to types.Callable

### LINKED ISSUE #7111: Using @overload with types.Callable fails to find compiled implementation at Lowering

Hi,

I was researching ways to implement classmethods for native numba types that represent python classes and found it can be done pretty easily if a type-reference type (i.e. a metaclass from which concrete types are created) is derived from `types.Callable` and low-leve API is used, but the same code fails to compile when `@overload` decorates the type-refence class (i.e. `CallableTypeRef` below). 

The issue can be reproduced with:
```
from numba import types, njit
from numba.core.typing.templates import signature
from numba.core.imputils import lower_constant
from numba.extending import (
    models,
    register_model,
    typeof_impl,
    type_callable,
    lower_builtin,
    overload,
)
from numba.core.typing.typeof import _typeof_type as numba_typeof_type


class MyClass(object):
    pass


@typeof_impl.register(type)
def mynew_typeof_type(val, c):

    print("DEBUG: sdc replaced val=", val)
    if not issubclass(val, MyClass):
        return numba_typeof_type(val, c)
    else:
        return CallableTypeRef(MyClass)


class CallableTypeRef(types.Callable):    ### this needs update from numba to use impl_key for Callables

    def __init__(self, instance_type):
        self.instance_type = instance_type
        self.impl_key = None
        super(CallableTypeRef, self).__init__('callable_type_ref[{}]'.format(self.instance_type))

    def get_call_type(self, context, args, kws):

        # >>> self == callable_type_ref[<class '__main__.MyClass'>]
        # type(self) == <class '__main__.CallableTypeRef'>
        template = context._functions[type(self)][0]  # for simplicity grab the first template
        res = template.apply(args, kws)

        # we have the signature, but also need to update the impl_key
        # that will be searched in get_function() during lowering
        if getattr(template, '_compiled_overloads', None):
            self.impl_key = template._compiled_overloads[args]
        return res

    def get_call_signatures(self):
        return (), True

    def get_impl_key(self, sig):
        return self.impl_key


@lower_constant(CallableTypeRef)
def constant_sdctyperef(context, builder, ty, pyval):


@register_model(CallableTypeRef)
class CallableTypeModel(models.OpaqueModel):

        models.OpaqueModel.__init__(self, dmm, fe_type)


# first version via low-level API
# @type_callable(CallableTypeRef)
# def typing_callabletype(context):
#     def typer():
#         return types.double
#     return typer
# 
# @lower_builtin(CallableTypeRef, types.VarArg(types.Any))
# def callabletype_impl(context, builder, sig, args):
#     res = context.get_constant(types.double, 42.7)
#     return res


# second version via overloads that fails to compile
@overload(CallableTypeRef)
def callable_type_call_ovld():
    def callable_type_call_impl():
        return 42.5
 
    return callable_type_call_impl


@njit
def test_impl():
    return MyClass()


result = test_impl()
print("DEBUG: result=\n", result)

```

The compilation fails at lowering even though the typing of the overload successfully completed and a compiled function was inserted into the `self._defns` dictionary, but a different key is used in `BaseContext::get_function()` to find the compiled function:
https://github.com/numba/numba/blob/5f6859908e7b73f7e1baf30381bac9e863640210/numba/core/base.py#L544-L550

Here's the debug output from lowering, it can be seen that there are no template versions for `fn == callable_type_ref[<class '__main__.MyClass'>]` as a key, which is selected because, `fn` is neither of these: `(types.Function, types.BoundFunction, types.Dispatcher)`:

```
>>> fn
callable_type_ref[<class '__main__.MyClass'>]
>>> isinstance(fn, (types.Function, types.BoundFunction, types.Dispatcher))
False
>>>
>>> self._defns[fn].versions
[]

>>> type(fn)
<class '__main__.CallableTypeRef'>
>>> self.typing_context._functions[type(fn)]
[<OverloadTemplate_CallableTypeRef C:\Users\akozlov\AppData\Local\Continuum\anaconda3\hpat\test_mindex_5_1.py:82>]
>>> self.typing_context._functions[type(fn)][0]
<OverloadTemplate_CallableTypeRef C:\Users\akozlov\AppData\Local\Continuum\anaconda3\hpat\test_mindex_5_1.py:82>
>>> self.typing_context._functions[type(fn)][0]._compiled_overloads
{(): <built-in method callable_type_call_impl of _dynfunc._Closure object at 0x00000285B7F70BA8>}
```

At the same time version written via low-level API decorators works and produces:
> (NUMBA_MASTER_4) C:\Users\akozlov\AppData\Local\Continuum\anaconda3\hpat>python test_typeref_call.py
> DEBUG: result=
>  42.7
> 

since there's a fallback branch that retires with `type(fn)` in `get_function`:
https://github.com/numba/numba/blob/5f6859908e7b73f7e1baf30381bac9e863640210/numba/core/base.py#L556-L559

The problem can be solved with replacing derived types in the key selection logic with the `types.Callable` class itself and adding `get_impl_key `abstract method to it. I can open a PR, if you think it's a good way to solve this. This issue looks very similar to #5885, but I haven't checked proposed solution will help there or not.

Best Regards,
Alexey.

kozlov-alexey:
This PR adds abstract method get_impl_key to types.Callable
that allows it's subclasses to work with @overload decorator
and find existing implementation during lowering.

Fixes #7111

### DIFF
--- a/numba/tests/test_extending.py
+++ b/numba/tests/test_extending.py
@@ -15,6 +15,7 @@ from numba.core import types, errors, typing, compiler, cgutils
 from numba.core.typed_passes import type_inference_stage
 from numba.core.registry import cpu_target
 from numba.core.compiler import compile_isolated
+from numba.core.imputils import lower_constant
 from numba.tests.support import (
     TestCase,
     captured_stdout,
@@ -490,6 +491,71 @@ class MkFuncTyping(AbstractTemplate):
 
 def mk_func_test_impl():
     mk_func_input(lambda a: a)
+
+
+# -----------------------------------------------------------------------
+# Define a types derived from types.Callable and overloads for them
+
+
+class MyClass(object):
+    pass
+
+
+class CallableTypeRef(types.Callable):
+
+    def __init__(self, instance_type):
+        self.instance_type = instance_type
+        self.sig_to_impl_key = {}
+        self.compiled_templates = []
+        super(CallableTypeRef, self).__init__('callable_type_ref'
+                                              '[{}]'.format(self.instance_type))
+
+    def get_call_type(self, context, args, kws):
+
+        # >>> self == callable_type_ref[<class '__main__.MyClass'>]
+        # type(self) == <class '__main__.CallableTypeRef'>
+        res_sig = None
+        for template in context._functions[type(self)]:
+            try:
+                res_sig = template.apply(args, kws)
+            except Exception:
+                pass  # for simplicity assume args must match exactly
+            else:
+                compiled_ovlds = getattr(template, '_compiled_overloads', {})
+                if args in compiled_ovlds:
+                    self.sig_to_impl_key[res_sig] = compiled_ovlds[args]
+                    self.compiled_templates.append(template)
+                    break
+
+        return res_sig
+
+    def get_call_signatures(self):
+        sigs = list(self.sig_to_impl_key.keys())
+        return sigs, True
+
+    def get_impl_key(self, sig):
+        return self.sig_to_impl_key[sig]
+
+
+@register_model(CallableTypeRef)
+class CallableTypeModel(models.OpaqueModel):
+    def __init__(self, dmm, fe_type):
+
+        models.OpaqueModel.__init__(self, dmm, fe_type)
+
+
+@typeof_impl.register(CallableTypeRef)
+def typeof_callable_typeref(val, c):
+    return val
+
+
+@lower_constant(CallableTypeRef)
+def constant_callable_typeref(context, builder, ty, pyval):
+    return context.get_dummy_value()
+
+
+def get_callable_typeref(instance_type):
+    return CallableTypeRef(instance_type)
 
 
 # -----------------------------------------------------------------------
@@ -1152,6 +1218,33 @@ class TestHighLevelExtending(TestCase):
             "Unknown attribute 'array_alloc' of",
             str(raises.exception),
         )
+
+    def test_overload_callable_typeref(self):
+
+        @overload(CallableTypeRef)
+        def callable_type_call_ovld1(x):
+            if isinstance(x, types.Integer):
+                def impl(x):
+                    return 42.5 + x
+                return impl
+
+        @overload(CallableTypeRef)
+        def callable_type_call_ovld2(x):
+            if isinstance(x, (types.UnicodeType, types.StringLiteral)):
+                def impl(x):
+                    return '42.5' + x
+
+                return impl
+
+        typeref_type = get_callable_typeref(MyClass)
+
+        @njit
+        def foo(a, b):
+            return typeref_type(a), typeref_type(b)
+
+        args = (4, '4')
+        expected = (42.5 + args[0], '42.5' + args[1])
+        self.assertPreciseEqual(foo(*args), expected)
 
 
 def _assert_cache_stats(cfunc, expect_hit, expect_misses):
--- a/numba/tests/test_extending.py
+++ b/numba/tests/test_extending.py
@@ -512,8 +512,6 @@ class CallableTypeRef(types.Callable):
 
     def get_call_type(self, context, args, kws):
 
-        # >>> self == callable_type_ref[<class '__main__.MyClass'>]
-        # type(self) == <class '__main__.CallableTypeRef'>
         res_sig = None
         for template in context._functions[type(self)]:
             try:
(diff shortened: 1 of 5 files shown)


gmarkall:
Can you add a test for this?

kozlov-alexey:
> Can you add a test for this?

Sure, I added one and fixed remaining test fails, so this is more or less finalized and can be reviewed.
The new test is a bit weird though (i.e. requires dummy lowering for CallableTypeRef as a const to be defined), if you have ideas how to simplify it, please let me know.

numba/tests/test_extending.py
```diff
@@ -1153,6 +1219,33 @@ def no_classmethod_in_base(nitems):
...
+                def impl(x):
+                    return '42.5' + x
+
+                return impl
+
+        typeref_type = get_callable_typeref(MyClass)
+
+        @njit
+        def foo(a, b):
+            return typeref_type(a), typeref_type(b)
```
sklam, line 1243: FYI, if you add 

```python
infer_global(MyClass, CallableTypeRef(MyClass))
```

the following can be supported:

```python
 @njit
def foo(a, b):
    return MyClass(a), MyClass(b)
```

numba/tests/test_extending.py
```diff
@@ -1153,6 +1219,33 @@ def no_classmethod_in_base(nitems):
...
+        @overload(CallableTypeRef)
+        def callable_type_call_ovld1(x):
+            if isinstance(x, types.Integer):
+                def impl(x):
+                    return 42.5 + x
+                return impl
+
+        @overload(CallableTypeRef)
+        def callable_type_call_ovld2(x):
+            if isinstance(x, (types.UnicodeType, types.StringLiteral)):
```
sklam, line 1233: `types.StringLiteral` not exercised

numba/tests/test_extending.py
```diff
@@ -492,6 +493,71 @@ def mk_func_test_impl():
...
+        self.instance_type = instance_type
+        self.sig_to_impl_key = {}
+        self.compiled_templates = []
+        super(CallableTypeRef, self).__init__('callable_type_ref'
+                                              '[{}]'.format(self.instance_type))
+
+    def get_call_type(self, context, args, kws):
+
+        # >>> self == callable_type_ref[<class '__main__.MyClass'>]
+        # type(self) == <class '__main__.CallableTypeRef'>
```
sklam, line 516: dead comments?

numba/tests/test_extending.py
```diff
@@ -492,6 +493,71 @@ def mk_func_test_impl():
...
+        return sigs, True
+
+    def get_impl_key(self, sig):
+        return self.sig_to_impl_key[sig]
+
+
+@register_model(CallableTypeRef)
+class CallableTypeModel(models.OpaqueModel):
+    def __init__(self, dmm, fe_type):
+
```
sklam, line 543: extra line?

sklam (commented):
I cannot find a way to remove the dummy lowering for `CallableTypeRef`, but I did find a way to support `MyClass(a)` without the need of `get_callable_typeref`. See diff:

```diff
diff --git a/numba/tests/test_extending.py b/numba/tests/test_extending.py
index 699ab1d21..e72c3826e 100644
--- a/numba/tests/test_extending.py
+++ b/numba/tests/test_extending.py
@@ -544,20 +544,13 @@ class CallableTypeModel(models.OpaqueModel):
         models.OpaqueModel.__init__(self, dmm, fe_type)
 
 
-@typeof_impl.register(CallableTypeRef)
-def typeof_callable_typeref(val, c):
-    return val
-
+infer_global(MyClass, CallableTypeRef(MyClass))
 
 @lower_constant(CallableTypeRef)
 def constant_callable_typeref(context, builder, ty, pyval):
     return context.get_dummy_value()
 
 
-def get_callable_typeref(instance_type):
-    return CallableTypeRef(instance_type)
-
-
 # -----------------------------------------------------------------------
 
 
@@ -1230,17 +1223,15 @@ class TestHighLevelExtending(TestCase):
 
         @overload(CallableTypeRef)
         def callable_type_call_ovld2(x):
-            if isinstance(x, (types.UnicodeType, types.StringLiteral)):
+            if isinstance(x, (types.UnicodeType)):
                 def impl(x):
                     return '42.5' + x
 
                 return impl
 
-        typeref_type = get_callable_typeref(MyClass)
-
         @njit
         def foo(a, b):
-            return typeref_type(a), typeref_type(b)
+            return MyClass(a), MyClass(b)
 
         args = (4, '4')
         expected = (42.5 + args[0], '42.5' + args[1])
```

numba/tests/test_extending.py
kozlov-alexey: Thanks for pointing this! I didn't know `infer_global` can be used that way, so I used `typeof_impl.register(type)` same as it's done for Numba typed containers.

sklam (approved):
Thanks for the patch!

PR #7186, opened 2021-07-06
Add support for np.logspace.

guoqiangqi:
This PR aims to add support for `np.logspace`.
Related to #4074

### DIFF
--- a/docs/source/reference/numpysupported.rst
+++ b/docs/source/reference/numpysupported.rst
@@ -484,6 +484,7 @@ The following top-level functions are supported:
 * :func:`numpy.interp` (only the 3 first arguments)
 * :func:`numpy.intersect1d` (only first 2 arguments, ar1 and ar2)
 * :func:`numpy.linspace` (only the 3-argument form)
+* :func:`numpy.logspace` (only the 3 first arguments)
 * :class:`numpy.ndenumerate`
 * :class:`numpy.ndindex`
 * :class:`numpy.nditer` (only the first argument)
@@ -498,7 +499,6 @@ The following top-level functions are supported:
   must be an integer)
 * :func:`numpy.roots`
 * :func:`numpy.rot90` (only the 2 first arguments)
-* :func:`numpy.logspace` (only the 4-argument form)
 * :func:`numpy.round_`
 * :func:`numpy.searchsorted` (only the 3 first arguments)
 * :func:`numpy.select` (only using homogeneous lists or tuples for the first
--- a/docs/source/reference/numpysupported.rst
+++ b/docs/source/reference/numpysupported.rst
@@ -498,6 +498,7 @@ The following top-level functions are supported:
   must be an integer)
 * :func:`numpy.roots`
 * :func:`numpy.rot90` (only the 2 first arguments)
+* :func:`numpy.logspace`
 * :func:`numpy.round_`
 * :func:`numpy.searchsorted` (only the 3 first arguments)
 * :func:`numpy.select` (only using homogeneous lists or tuples for the first
--- a/docs/source/reference/numpysupported.rst
+++ b/docs/source/reference/numpysupported.rst
@@ -498,7 +498,7 @@ The following top-level functions are supported:
   must be an integer)
 * :func:`numpy.roots`
 * :func:`numpy.rot90` (only the 2 first arguments)
-* :func:`numpy.logspace`
+* :func:`numpy.logspace` (only the 4-argument form)
 * :func:`numpy.round_`
 * :func:`numpy.searchsorted` (only the 3 first arguments)
 * :func:`numpy.select` (only using homogeneous lists or tuples for the first
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -1568,7 +1568,7 @@ def numpy_logspace(start, stop, num=50):
     if not isinstance(start, types.Number):
         raise errors.TypingError('The first argument "start" must be a number')
     if not isinstance(stop, types.Number):
-        raise errors.TypingError('The second argument "end" must be a number')
+        raise errors.TypingError('The second argument "stop" must be a number')
     if not isinstance(num, (int, types.Integer)):
         raise errors.TypingError('The third argument "num" must be an integer')
 
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -1563,6 +1563,19 @@ def array_T(context, builder, typ, value):
     return impl_ret_borrowed(context, builder, typ, res)
 
 
+@overload(np.logspace)
+def numpy_logspace(start, stop, num=50, endpoint=True,
+                   base=10.0, dtype=None, axis=0):
+
+    def impl(start, stop, num=50, endpoint=True, base=10.0, dtype=None, axis=0):
+        y = np.linspace(start, stop, num=num, endpoint=endpoint, axis=axis)
+        if dtype is None:
+            return np.power(base, y)
+        return np.power(base, y).astype(dtype, copy=False)
+
+    return impl
+
+
 @overload(np.rot90)
 def numpy_rot90(arr, k=1):
     # supporting axes argument it needs to be included in np.flip
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -1564,11 +1564,17 @@ def array_T(context, builder, typ, value):
 
 
 @overload(np.logspace)
-def numpy_logspace(start, stop, num=50, base=10.0):
+def numpy_logspace(start, stop, num=50):
+    if not isinstance(start, types.Number):
+        raise errors.TypingError('The first argument "start" must be a number')
+    if not isinstance(stop, types.Number):
+        raise errors.TypingError('The second argument "end" must be a number')
+    if not isinstance(num, (int, types.Integer)):
+        raise errors.TypingError('The third argument "num" must be an integer')
 
-    def impl(start, stop, num=50, base=10.0):
+    def impl(start, stop, num=50):
         y = np.linspace(start, stop, num)
-        return np.power(base, y)
+        return np.power(10.0, y)
 
     return impl
 
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -1564,14 +1564,11 @@ def array_T(context, builder, typ, value):
 
 
 @overload(np.logspace)
-def numpy_logspace(start, stop, num=50, endpoint=True,
-                   base=10.0, dtype=None, axis=0):
+def numpy_logspace(start, stop, num=50, base=10.0):
 
-    def impl(start, stop, num=50, endpoint=True, base=10.0, dtype=None, axis=0):
-        y = np.linspace(start, stop, num=num, endpoint=endpoint, axis=axis)
-        if dtype is None:
-            return np.power(base, y)
-        return np.power(base, y).astype(dtype, copy=False)
+    def impl(start, stop, num=50, base=10.0):
+        y = np.linspace(start, stop, num)
+        return np.power(base, y)
 
     return impl
 
--- a/numba/tests/test_np_functions.py
+++ b/numba/tests/test_np_functions.py
@@ -2458,7 +2458,7 @@ class TestNPFunctions(MemoryLeakMixin, TestCase):
 
     def test_logspace2_basic(self):
 
-        def Inputs():
+        def inputs():
             #start, stop
             yield 1, 60
             yield -1, 60
@@ -2480,9 +2480,8 @@ class TestNPFunctions(MemoryLeakMixin, TestCase):
         pyfunc = logspace2
         cfunc = jit(nopython=True)(pyfunc)
 
-        for start, stop in Inputs():
-            self.assertAlmostEqual(pyfunc(start, stop).all(),
-                                   cfunc(start, stop).all())
+        for start, stop in inputs():
+            np.testing.assert_allclose(pyfunc(start, stop), cfunc(start, stop))
 
     def test_logspace2_exception(self):
         cfunc = jit(nopython=True)(logspace2)
@@ -2501,7 +2500,7 @@ class TestNPFunctions(MemoryLeakMixin, TestCase):
 
     def test_logspace3_basic(self):
 
-        def Inputs():
+        def inputs():
             #start, stop
             yield 1, 60
             yield -1, 60
@@ -2522,15 +2521,13 @@ class TestNPFunctions(MemoryLeakMixin, TestCase):
 
         pyfunc = logspace3
         cfunc = jit(nopython=True)(pyfunc)
-        self.disable_leak_check()
 
-        for start, stop in Inputs():
-            self.assertAlmostEqual(pyfunc(start, stop).all(),
-                                   cfunc(start, stop).all())
+        for start, stop in inputs():
+            np.testing.assert_allclose(pyfunc(start, stop), cfunc(start, stop))
 
     def test_logspace3_with_num_basic(self):
 
-        def Inputs():
+        def inputs():
             #start, stop, num
             yield 1, 60, 20
             yield -1, 60, 30
@@ -2551,11 +2548,10 @@ class TestNPFunctions(MemoryLeakMixin, TestCase):
 
         pyfunc = logspace3
         cfunc = jit(nopython=True)(pyfunc)
-        self.disable_leak_check()
 
-        for start, stop, num in Inputs():
-            self.assertAlmostEqual(pyfunc(start, stop, num).all(),
-                                   cfunc(start, stop, num).all())
+        for start, stop, num in inputs():
+            np.testing.assert_allclose(pyfunc(start, stop, num),
+                                       cfunc(start, stop, num))
 
     def test_logspace3_exception(self):
         cfunc = jit(nopython=True)(logspace3)
--- a/numba/tests/test_np_functions.py
+++ b/numba/tests/test_np_functions.py
@@ -157,7 +157,11 @@ def flip(a):
     return np.flip(a)
 
 
-def logspace(start, stop, num=50):
+def logspace2(start, stop):
+    return np.logspace(start, stop)
+
+
+def logspace3(start, stop, num=50):
     return np.logspace(start, stop, num=num)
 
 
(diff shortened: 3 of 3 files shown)


guoqiangqi:
Hi @stuartarchibald , i have removed the DRAFT tag for this PR, it is ready for reviewing now.

stuartarchibald:
> Hi @stuartarchibald , i have removed the DRAFT tag for this PR, it is ready for reviewing now.

Many thanks. I'll update the labels.

docs/source/reference/numpysupported.rst
```diff
@@ -498,6 +498,7 @@ The following top-level functions are supported:
   must be an integer)
 * :func:`numpy.roots`
 * :func:`numpy.rot90` (only the 2 first arguments)
+* :func:`numpy.logspace` (only the 4-argument form)
```
stuartarchibald, line 501:
```suggestion
* :func:`numpy.logspace` (only the first 4 arguments)
```
given two of the args are kwargs, think it'd be ok to do `logspace(1, 10)` which is a 2-arg form.

numba/np/arrayobj.py
```diff
@@ -1563,6 +1563,16 @@ def array_T(context, builder, typ, value):
     return impl_ret_borrowed(context, builder, typ, res)
 
 
+@overload(np.logspace)
+def numpy_logspace(start, stop, num=50, base=10.0):
```
stuartarchibald, line 1567: I don't think this will replicate the NumPy behaviour in its present form. The `endpoint` kwarg is missing and should be located between `num` and `base`. The impact of this is that if a user calls the Numba `np.logspace` implementation without using kwargs they'll get unexpected behaviour. e.g.
```
logspace(5,  10, 50, 10.0) # 10.0 should be `endpoint`!
```
reference docs: https://numpy.org/doc/stable/reference/generated/numpy.logspace.html

numba/np/arrayobj.py
```diff
@@ -1563,6 +1563,16 @@ def array_T(context, builder, typ, value):
...
+
```
stuartarchibald, line 1568: There's no type checking being done on the types of the arguments in the current implementation, perhaps take a look at the guide to using `@overload` for some idea of what is expected https://numba.readthedocs.io/en/stable/extending/overloading-guide.html#concrete-example (or just look at `np.rot90` immediately below this function)?

numba/tests/test_np_functions.py
```diff
@@ -2448,6 +2452,15 @@ def a_variations():
...
         self.assertIn("Cannot np.flip on UniTuple", str(raises.exception))
 
+    def test_logspace_basic(self):
+        pyfunc = logspace
+        cfunc = jit(nopython=True)(pyfunc)
+
+        self.assertPreciseEqual(pyfunc(0, 6), cfunc(0, 6))
+        self.assertPreciseEqual(pyfunc(0, 6, num=100), cfunc(0, 6, num=100))
+        self.assertPreciseEqual(pyfunc(0, 6, base=2), cfunc(0, 6, base=2))
+        self.assertPreciseEqual(pyfunc(0, 6, 100, 2), cfunc(0, 6, 100, 2))
```
stuartarchibald, line 2462: It'd be good to add some more tests, perhaps consider what is permitted as input into the function and try some variations in value and type. One slightly unusual example might be:
```
In [14]: np.logspace(1, 10, num=4, base=np.array([2, np.e, 3, 10]))
Out[14]: array([2.000000e+00, 5.459815e+01, 2.187000e+03, 1.000000e+10])
```

stuartarchibald (commented):
Thanks for the patch. I've given this an initial review to provide some early feedback. The implementation and tests present are a great start. Thanks again!

numba/np/arrayobj.py
guoqiangqi: I have changed logspace implementation to 3-argument form to avoid confusion between `endpoint` and `base` because `endpoint` is not supported now, i  will add full support for `linspace` and `logspace` in subsequent PR.

docs/source/reference/numpysupported.rst
guoqiangqi: I am not sure about the method for counting argument number of a function with optional argument, so i just follow the `rot90`, which said  `only the 2 first arguments` when it have a necessary argument `arr` and an optional argument `k`.

numba/tests/test_np_functions.py
guoqiangqi: Done.

numba/np/arrayobj.py
guoqiangqi: Appreciate your advices, i have updated this part of code.

esc:
@guoqiangqi thank you for the updates, it seems like CI is failing. I took a quick look at the tests and it seems like there are some resiliency issues towards numerical accuracy. The memory leaks are just a side-effect of the failing tests.

guoqiangqi:
> @guoqiangqi thank you for the updates, it seems like CI is failing. I took a quick look at the tests and it seems like there are some resiliency issues towards numerical accuracy. The memory leaks are just a side-effect of the failing tests.

I have checked the error messages, CI failed only on Windows and Linux with svml, i would try to fix this as soon as possible.

guoqiangqi:
@esc It seems like the resiliency accuracy issues was cased by `SVML` used, checks succeed when set `NUMBA_DISABLE_INTEL_SVML`  `1`. Should i just do these tests without `SVML` because i think it have not more to do to fix this in this simple patch, which call `linspace` directly.

esc:
> @esc It seems like the resiliency accuracy issues was cased by `SVML` used, checks succeed when set `NUMBA_DISABLE_INTEL_SVML` `1`. Should i just do these tests without `SVML` because i think it have not more to do to fix this in this simple patch, which call `linspace` directly.

I am  not sure. My suggestion would be to replace `assertPreciseEqual` with something like `assertAlmostEqual` to increase resilience to numerical accuracy.

guoqiangqi:
> > @esc It seems like the resiliency accuracy issues was cased by `SVML` used, checks succeed when set `NUMBA_DISABLE_INTEL_SVML` `1`. Should i just do these tests without `SVML` because i think it have not more to do to fix this in this simple patch, which call `linspace` directly.
> 
> I am not sure. My suggestion would be to replace `assertPreciseEqual` with something like `assertAlmostEqual` to increase resilience to numerical accuracy.

`assertAlmostEqual` is a good choice, i found a workaround by changing some failed test cases, which one should be better?

esc:
> > > @esc It seems like the resiliency accuracy issues was cased by `SVML` used, checks succeed when set `NUMBA_DISABLE_INTEL_SVML` `1`. Should i just do these tests without `SVML` because i think it have not more to do to fix this in this simple patch, which call `linspace` directly.
> > 
> > 
> > I am not sure. My suggestion would be to replace `assertPreciseEqual` with something like `assertAlmostEqual` to increase resilience to numerical accuracy.
> 
> `assertAlmostEqual` is a good choice, i found a workaround by changing some failed test cases, which one should be better?

I thought about this for a while and it is hard to recommend anything, I'd rather have the extreme valued test cases for example and use `almost` rather than restricting the input and using `precises`. Maybe even both variants could be included?

@gmarkall @sklam what would be your call?

stuartarchibald:
Numba's `assertPreciseEqual` has support for specifying an accepted number of ULPs of error:
https://github.com/numba/numba/blob/5baa1ce53073a786060078a8cbe0ee57b1fb67d2/numba/tests/support.py#L308-L311 I'd recommend using that. If SVML vs. libm is seriously out it's likely either a difference in the way that NumPy implements the alg in comparison to the Numba equivalent (most likely) or a bug in libm/SVML (not so likely). The former obviously needs fixing in the patch!

Also @guoqiangqi, if possible, please could you avoiding force pushing once someone has started to look at/review a patch. It just makes it easier to track changes and keeps the history of which CI builds failed. Many thanks for your help with this.

numba/np/arrayobj.py
```diff
@@ -1563,6 +1563,22 @@ def array_T(context, builder, typ, value):
     return impl_ret_borrowed(context, builder, typ, res)
 
 
+@overload(np.logspace)
+def numpy_logspace(start, stop, num=50):
+    if not isinstance(start, types.Number):
+        raise errors.TypingError('The first argument "start" must be a number')
+    if not isinstance(stop, types.Number):
+        raise errors.TypingError('The second argument "end" must be a number')
```
stuartarchibald, line 1571:
```suggestion
    if not isinstance(stop, types.Number):
        raise errors.TypingError('The second argument "stop" must be a number')
```
The argument is called stop?

numba/tests/test_np_functions.py
```diff
@@ -2448,6 +2452,56 @@ def a_variations():
...
+        self.assertPreciseEqual(pyfunc(1, 60), cfunc(1, 60))
+        self.assertPreciseEqual(pyfunc(-60, 1), cfunc(-60, 1))
+        self.assertPreciseEqual(pyfunc(-1, -60), cfunc(-1, -60))
+
+        self.assertPreciseEqual(pyfunc(1.0, 60.0), cfunc(1.0, 60.0))
+        self.assertPreciseEqual(pyfunc(-60.0, -1.0), cfunc(-60.0, -1.0))
+        self.assertPreciseEqual(pyfunc(-1.0, 60.0), cfunc(-1.0, 60.0))
+
+        self.assertPreciseEqual(pyfunc(0.0, np.e), cfunc(0.0, np.e))
+        self.assertPreciseEqual(pyfunc(0.0, np.pi), cfunc(0.0, np.pi))
```
stuartarchibald, line 2468: Consider writing this as a loop over input tuples? It'd also be good to add in some variation in the type of the input, e.g. complex values.

numba/tests/test_np_functions.py
```diff
@@ -157,6 +157,10 @@ def flip(a):
     return np.flip(a)
 
 
+def logspace(start, stop, num=50):
+    return np.logspace(start, stop, num=num)
```
stuartarchibald, line 161: Similarly to e.g. `bincount` above, both the 2 and 3 argument form should be checked so as to ensure that kwarg defaults are handle correctly.

numba/tests/test_np_functions.py
```diff
@@ -2448,6 +2452,56 @@ def a_variations():
...
+        self.assertPreciseEqual(pyfunc(-1.0, 60.0), cfunc(-1.0, 60.0))
+
+        self.assertPreciseEqual(pyfunc(0.0, np.e), cfunc(0.0, np.e))
+        self.assertPreciseEqual(pyfunc(0.0, np.pi), cfunc(0.0, np.pi))
+
+    def test_logspace_with_num_basic(self):
+        pyfunc = logspace
+        cfunc = jit(nopython=True)(pyfunc)
+
+        self.assertPreciseEqual(pyfunc(1, 50, num=20), cfunc(1, 50, num=20))
```
stuartarchibald, line 2474: Perhaps consider factoring this out as:
```python
def check(start, stop, num):
    self.assertPreciseEqual(pyfunc(start, stop, num=num), cfunc(start, stop, num=num))
    self.assertPreciseEqual(pyfunc(start, stop, num), cfunc(start, stop, num=num))
```
and using that throughout? Also note the comment above about checking the default args by having 2 and 3 arg variations.

stuartarchibald (commented):
Thanks for the patch, just a few minor things mostly testing related to resolve else looks good. Patch https://github.com/numba/numba/pull/7265 should help with at least getting the input to the `power` call to match NumPy. I think most of the numerical issues are coming from differences in the power implementations, suggest using ULPs based tolerances as necessary as noted above.

guoqiangqi:
> Numba's `assertPreciseEqual` has support for specifying an accepted number of ULPs of error:
> https://github.com/numba/numba/blob/5baa1ce53073a786060078a8cbe0ee57b1fb67d2/numba/tests/support.py#L308-L311
> 
> I'd recommend using that. If SVML vs. libm is seriously out it's likely either a difference in the way that NumPy implements the alg in comparison to the Numba equivalent (most likely) or a bug in libm/SVML (not so likely). The former obviously needs fixing in the patch!
> Also @guoqiangqi, if possible, please could you avoiding force pushing once someone has started to look at/review a patch. It just makes it easier to track changes and keeps the history of which CI builds failed. Many thanks for your help with this.

Sorry about the force pushing before, i would avoiding this.   
Back to the `assertPreciseEqual`,  actually,  i have specified `ULPS` to `4` before but it didn't work, still got assertion errors. Should i change `4` to a larger number？  Also, `assertAlmostEqual` may be a better choice before we find out the reason of  accuracy issues.

numba/np/arrayobj.py
guoqiangqi: Yes, the argument is called stop, which is reference to implementation in Numpy.

guoqiangqi:
@stuartarchibald All jobs done in the latest commit.

numba/tests/test_np_functions.py
```diff
@@ -2448,6 +2456,127 @@ def a_variations():
 
         self.assertIn("Cannot np.flip on UniTuple", str(raises.exception))
 
+    def test_logspace2_basic(self):
+
+        def Inputs():
```
stuartarchibald, line 2461:
```suggestion
        def inputs():
```

numba/tests/test_np_functions.py
```diff
@@ -2448,6 +2456,127 @@ def a_variations():
...
+            yield np.complex64(2j), np.complex64(4j)
+            yield np.complex64(2), np.complex64(4j)
+            yield np.complex64(1 + 2j), np.complex64(3 + 4j)
+            yield np.complex64(1 - 2j), np.complex64(3 - 4j)
+            yield np.complex64(-1 + 2j), np.complex64(3 + 4j)
+
+        pyfunc = logspace2
+        cfunc = jit(nopython=True)(pyfunc)
+
+        for start, stop in Inputs():
```
stuartarchibald, line 2483:
```suggestion
        for start, stop in inputs():
```

numba/tests/test_np_functions.py
```diff
@@ -2448,6 +2456,127 @@ def a_variations():
...
+            yield np.complex64(1), np.complex64(2)
+            yield np.complex64(2j), np.complex64(4j)
+            yield np.complex64(2), np.complex64(4j)
+            yield np.complex64(1 + 2j), np.complex64(3 + 4j)
+            yield np.complex64(1 - 2j), np.complex64(3 - 4j)
+            yield np.complex64(-1 + 2j), np.complex64(3 + 4j)
+
+        pyfunc = logspace3
+        cfunc = jit(nopython=True)(pyfunc)
+        self.disable_leak_check()
```
stuartarchibald, line 2525:
```suggestion
```
I think that this should check for leaks

numba/tests/test_np_functions.py
```diff
@@ -2448,6 +2456,127 @@ def a_variations():
...
+            yield np.complex64(1 + 2j), np.complex64(3 + 4j)
+            yield np.complex64(1 - 2j), np.complex64(3 - 4j)
+            yield np.complex64(-1 + 2j), np.complex64(3 + 4j)
+
+        pyfunc = logspace2
+        cfunc = jit(nopython=True)(pyfunc)
+
+        for start, stop in Inputs():
+            self.assertAlmostEqual(pyfunc(start, stop).all(),
+                                   cfunc(start, stop).all())
```
stuartarchibald, line 2485:
```suggestion
            np.testing.assert_allclose(pyfunc(start, stop), cfunc(start, stop))
```
If this code needs a looser testing regime, and it's all NumPy array outputs, I suggest using `np.testing.assert_allclose`.

numba/tests/test_np_functions.py
```diff
@@ -2448,6 +2456,127 @@ def a_variations():
...
+                      str(raises.exception))
+
+        with self.assertRaises(TypingError) as raises:
+            cfunc(5, "abc")
+        self.assertIn('The second argument "stop" must be a number',
+                      str(raises.exception))
+
+    def test_logspace3_basic(self):
+
+        def Inputs():
```
stuartarchibald, line 2504:
```suggestion
        def inputs():
```

numba/tests/test_np_functions.py
```diff
@@ -2448,6 +2456,127 @@ def a_variations():
...
+            yield np.complex64(2), np.complex64(4j)
+            yield np.complex64(1 + 2j), np.complex64(3 + 4j)
+            yield np.complex64(1 - 2j), np.complex64(3 - 4j)
+            yield np.complex64(-1 + 2j), np.complex64(3 + 4j)
+
+        pyfunc = logspace3
+        cfunc = jit(nopython=True)(pyfunc)
+        self.disable_leak_check()
+
+        for start, stop in Inputs():
```
stuartarchibald, line 2527:
```suggestion
        for start, stop in inputs():
```

numba/tests/test_np_functions.py
```diff
@@ -2448,6 +2456,127 @@ def a_variations():
...
+            self.assertAlmostEqual(pyfunc(start, stop).all(),
+                                   cfunc(start, stop).all())
```
stuartarchibald, line 2529:
```suggestion
            np.testing.assert_allclose(pyfunc(start, stop), cfunc(start, stop))
```
```

File: numba/tests/test_np_functions.py [hunk 16]
```diff
@@ -2448,6 +2456,127 @@ def a_variations():
...
+
+    def test_logspace3_with_num_basic(self):
+
+        def Inputs():
```
stuartarchibald, line 2533:
```suggestion
        def inputs():
```

numba/tests/test_np_functions.py
```diff
@@ -2448,6 +2456,127 @@ def a_variations():
...
+            yield np.complex64(1), np.complex64(2), 40
+            yield np.complex64(2j), np.complex64(4j), 50
+            yield np.complex64(2), np.complex64(4j), 60
+            yield np.complex64(1 + 2j), np.complex64(3 + 4j), 70
+            yield np.complex64(1 - 2j), np.complex64(3 - 4j), 80
+            yield np.complex64(-1 + 2j), np.complex64(3 + 4j), 90
+
+        pyfunc = logspace3
+        cfunc = jit(nopython=True)(pyfunc)
+        self.disable_leak_check()
```
stuartarchibald, line 2554:
```suggestion
```

numba/tests/test_np_functions.py
```diff
@@ -2448,6 +2456,127 @@ def a_variations():
...
+
+        for start, stop, num in Inputs():
```
stuartarchibald, line 2556:
```suggestion
        for start, stop, num in inputs():
```

numba/tests/test_np_functions.py
```diff
@@ -2448,6 +2456,127 @@ def a_variations():
...
+            self.assertAlmostEqual(pyfunc(start, stop, num).all(),
+                                   cfunc(start, stop, num).all())
```
stuartarchibald, line 2558:
```suggestion
            np.testing.assert_allclose(pyfunc(start, stop, num),
                                       cfunc(start, stop, num))
```
```

stuartarchibald (commented):
Thanks for the update, there's a few minor (mostly stylistic things) to resolve else this looks good. Thanks again for working on this!

guoqiangqi:
@stuartarchibald Thanks for your review, i have updated  related code to your advices.

stuartarchibald (approved):
Thanks for the patch and fixes, looks good!

Issue #2849, opened 2018-03-26
CUDA test failures on Tegra TX2 (only a few!)


seibert:
Much to my surprise, Numba's CUDA target mostly "just works" on the Tegra TX2 (there are some CPU failures that I'll post in another issue).  Note that a Tegra TX2 is ARMv8 combined with a GP10B CUDA device, which is Pascal (but compute capability 6.2, different from GTX 10XX).  When using the Jetson TX2 dev board, the most common operating system is Ubuntu 16.04, where Numba can run without any modification (not even the small patch we need for ARMv7).  

After compiling LLVM 5.0.1, llvmlite master, and Numba master, you can run the CUDA unit tests with a command like this:
```
NUMBAPRO_LIBDEVICE=/usr/local/cuda/nvvm/libdevice/ NUMBAPRO_NVVM=/usr/local/cuda/nvvm/lib/libnvvm.so NUMBAPRO_CUDALIB=/usr/local/cuda/lib64/ python -m numba.runtests -v numba.cuda.tests
```
(we really need to change those environment variable names, and maybe make a convenience NUMBA_CUDA_SDK that lets you specify the top-level CUDA directory and infers the rest of this.)

This command produces the following failing tests:
<details>

```
======================================================================
ERROR: test_pinned (numba.cuda.tests.cudadrv.test_pinned.TestPinned)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/nvidia/jetconda3/lib/python3.6/site-packages/numba/cuda/tests/cudadrv/test_pinned.py", line 27, in test_pinned
    with cuda.pinned(A):
  File "/home/nvidia/jetconda3/lib/python3.6/contextlib.py", line 82, in __enter__
    return next(self.gen)
  File "/home/nvidia/jetconda3/lib/python3.6/site-packages/numba/cuda/api.py", line 203, in pinned
    mapped=False)
  File "/home/nvidia/jetconda3/lib/python3.6/site-packages/numba/cuda/cudadrv/driver.py", line 737, in mempin
    allocator()
  File "/home/nvidia/jetconda3/lib/python3.6/site-packages/numba/cuda/cudadrv/driver.py", line 732, in allocator
    driver.cuMemHostRegister(pointer, size, flags)
  File "/home/nvidia/jetconda3/lib/python3.6/site-packages/numba/cuda/cudadrv/driver.py", line 288, in safe_cuda_api_call
    self._check_error(fname, retcode)
[...]

PR #7190, opened 2021-07-06
CUDA: Fix test_pinned on Jetson

gmarkall:
Pinning memory with `cuMemHostRegister()` now appears to be supported, but only for memory regions < 4MB. This commit fixes test_pinned by only attempting to pin a 2MB array on ARM.

Marking as draft for now as there is another test case in Issue #2849 that is failing that I need to check.

### DIFF
--- a/numba/cuda/tests/cudadrv/test_pinned.py
+++ b/numba/cuda/tests/cudadrv/test_pinned.py
@@ -1,10 +1,8 @@
 import numpy as np
+import platform
 
 from numba import cuda
 from numba.cuda.testing import unittest, ContextResettingTestCase
-
-
-REPEAT = 25
 
 
 class TestPinned(ContextResettingTestCase):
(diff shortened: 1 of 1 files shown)


gmarkall:
Making this ready for review as it is a complete fix in itself, and I won't have time to look into whether I can reproduce the other (separate) issue.

numba/cuda/tests/cudadrv/test_pinned.py
```diff
@@ -21,7 +19,12 @@ def _run_copies(self, A):
         self.assertTrue(np.allclose(A, A0))
 
     def test_pinned(self):
-        A = np.arange(2 * 1024 * 1024) # 16 MB
+        machine = platform.machine()
+        if machine.startswith('arm') or machine.startswith('aarch64'):
+            count = 262144   # 2MB
+        else:
+            count = 2097152  # 16MB
```
stuartarchibald, line 26: Assumes an 8 byte `intp`? Given a 4 byte `intp` would lead to a smaller allocation size, I think this is still fine.

stuartarchibald (approved):
Thanks for the patch, changes look on on inspection. Need to run this on hardware to test.

sklam:
BFID `numba_smoketest_cuda_yaml_83`

sklam:
BF passed.

numba/cuda/tests/cudadrv/test_pinned.py
gmarkall: Indeed. All systems that support CUDA are 64-bit anyway.

gmarkall:
> BF passed, but i thk we still need a Jetson test(?)

That would be ideal, but it's up to you :-)

sklam:
I actually don't have a Xavier to make it work. Since I'm not seeing any new failures, i'm marking this as ready.

gmarkall:
Does this fail on a Jetson Nano? (I don't have one to test)

PR #7192, opened 2021-07-07
Fix missing import in array.argsort impl and add more tests.

stuartarchibald:
As title.


sklam (approved):
Thanks for the patch!

PR #7195, opened 2021-07-07
Improve flake8 compliance and code style (1)

GeorgeEbberson:
Beginning a campaign to reduce the list of "grandfathered" files in the `flake8` config, starting with `stencil.py`. Since this is probably going to be a big task and take a while, I figured I'd best get some opinions before continuing.

I've reduced `numba/stencils/stencil.py` to only throwing 4 errors:
```
numba\stencils\stencil.py:8:1: F401 'numpy as np' imported but unused
numba\stencils\stencil.py:12:1: F401 'numba.core.typing.templates.CallableTemplate' imported but unused
numba\stencils\stencil.py:12:1: F401 'numba.core.typing.templates.infer_global' imported but unused
numba\stencils\stencil.py:309:25: F841 local variable 'sum_results' is assigned to but never used
```
Before removing all 4, I thought best to check with someone more familiar with the codebase that such imports aren't needed for e.g. side-effect reasons, although I can see nothing obvious.

If you look at the changes I've made, you'll find the style I've used is heavily influenced by [Black](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html), although I've deviated in places where it seemed appropriate - and before someone points out such a deviation, I made the changes by hand which is why the style hasn't been implemented everywhere. But I do wonder how you'd feel about adding something like Black, which includes an autoformatter, in order to tidy up the code style? It would likely throw lots of errors at first, but I'd be relatively happy to focus effort there (as well as in updating older styles) as part of this.

More than anything, I could also do with some general consensus/agreement before I begin on stylistic choices. Things like `%` versus `.format()` string formatting, and whether or not to include parentheses when unpacking tuples, and other such small changes - I don't want to make lots of changes, only to find that someone either a) has major disagreement or, even worse, b) had written using one specific style with good reason!

And finally, in order to track what I've done, I intend to modify and remove a few files per PR to make them more manageable for reviewers.

### DIFF
--- a/numba/core/inline_closurecall.py
+++ b/numba/core/inline_closurecall.py
@@ -1071,9 +1071,9 @@ def _inline_arraycall(func_ir, cfg, visited, loop, swapped, enable_prange=False,
     debug_print("removed variables: ", removed)
 
     # Define an index_var to index the array.
-    # If the range happens to be single step ranges like range(n), or range(m,
-    # n), then the index_var correlates to iterator index; otherwise we'll have
-    # to define a new counter.
+    # If the range happens to be single step ranges like range(n), or 
+    # range(m, n), then the index_var correlates to iterator index; otherwise
+    # we'll have to define a new counter.
     range_def = guard(_find_iter_range, func_ir, iter_var, swapped)
     index_var = ir.Var(scope, mk_unique_var("index"), loc)
     if range_def and range_def[0] == 0:
--- a/numba/core/inline_closurecall.py
+++ b/numba/core/inline_closurecall.py
@@ -2,8 +2,7 @@ import types as pytypes  # avoid confusion with numba.types
 import copy
 import ctypes
 import numba.core.analysis
-from numba.core import utils, types, typing, errors, ir, rewrites, config, ir_utils
-from numba import prange
+from numba.core import types, typing, errors, ir, rewrites, config, ir_utils
 from numba.parfors.parfor import internal_prange
 from numba.core.ir_utils import (
     mk_unique_var,
@@ -25,7 +24,7 @@ from numba.core.ir_utils import (
     simplify_CFG,
     canonicalize_array_math,
     dead_code_elimination,
-    )
+)
 
 from numba.core.analysis import (
     compute_cfg_from_blocks,
@@ -80,24 +79,24 @@ class InlineClosureCallPass(object):
             label, block = work_list.pop()
             for i, instr in enumerate(block.body):
                 if isinstance(instr, ir.Assign):
-                    lhs = instr.target
                     expr = instr.value
                     if isinstance(expr, ir.Expr) and expr.op == 'call':
                         call_name = guard(find_callname, self.func_ir, expr)
-                        func_def = guard(get_definition, self.func_ir, expr.func)
+                        func_def = guard(get_definition, self.func_ir,
+                                         expr.func)
 
                         if guard(self._inline_reduction,
                                  work_list, block, i, expr, call_name):
                             modified = True
                             break # because block structure changed
 
-                        if guard(self._inline_closure,
-                                work_list, block, i, func_def):
+                        if guard(self._inline_closure, work_list, block, i,
+                                 func_def):
                             modified = True
                             break # because block structure changed
 
-                        if guard(self._inline_stencil,
-                                instr, call_name, func_def):
+                        if guard(self._inline_stencil, instr, call_name,
+                                 func_def):
                             modified = True
 
         if enable_inline_arraycall:
@@ -112,11 +111,11 @@ class InlineClosureCallPass(object):
             sized_loops = [(k, len(loops[k].body)) for k in loops.keys()]
             visited = []
             # We go over all loops, bigger loops first (outer first)
-            for k, s in sorted(sized_loops, key=lambda tup: tup[1], reverse=True):
+            for k, s in sorted(sized_loops, key=lambda t: t[1], reverse=True):
                 visited.append(k)
-                if guard(_inline_arraycall, self.func_ir, cfg, visited, loops[k],
-                         self.swapped, self.parallel_options.comprehension,
-                         self.typed):
+                if guard(_inline_arraycall, self.func_ir, cfg, visited,
+                         loops[k], self.swapped,
+                         self.parallel_options.comprehension, self.typed):
                     modified = True
             if modified:
                 _fix_nested_array(self.func_ir)
@@ -150,6 +149,7 @@ class InlineClosureCallPass(object):
                             "two arguments are required (optional initial "
                             "value can also be specified)")
         check_reduce_func(self.func_ir, expr.args[0])
+
         def reduce_func(f, A, v=None):
             it = iter(A)
             if v is not None:
@@ -157,12 +157,12 @@ class InlineClosureCallPass(object):
             else:
                 s = next(it)
             for a in it:
-               s = f(s, a)
+                s = f(s, a)
             return s
         inline_closure_call(self.func_ir,
-                        self.func_ir.func_id.func.__globals__,
-                        block, i, reduce_func, work_list=work_list,
-                        callee_validator=callee_ir_validator)
+                            self.func_ir.func_id.func.__globals__, block, i,
+                            reduce_func, work_list=work_list,
+                            callee_validator=callee_ir_validator)
         return True
 
     def _inline_stencil(self, instr, call_name, func_def):
@@ -173,8 +173,8 @@ class InlineClosureCallPass(object):
         # alive by adding them to the actual kernel call as extra
         # keyword arguments, which is ignored anyway.
         if (isinstance(func_def, ir.Global) and
-            func_def.name == 'stencil' and
-            isinstance(func_def.value, StencilFunc)):
+                func_def.name == 'stencil' and
+                isinstance(func_def.value, StencilFunc)):
             if expr.kws:
                 expr.kws += func_def.value.kws
             else:
@@ -187,23 +187,24 @@ class InlineClosureCallPass(object):
         self._processed_stencils.append(expr)
         if not len(expr.args) == 1:
             raise ValueError("As a minimum Stencil requires"
-                " a kernel as an argument")
+                             " a kernel as an argument")
         stencil_def = guard(get_definition, self.func_ir, expr.args[0])
         require(isinstance(stencil_def, ir.Expr) and
                 stencil_def.op == "make_function")
         kernel_ir = get_ir_of_code(self.func_ir.func_id.func.__globals__,
-                stencil_def.code)
+                                   stencil_def.code)
         options = dict(expr.kws)
         if 'neighborhood' in options:
             fixed = guard(self._fix_stencil_neighborhood, options)
             if not fixed:
-               raise ValueError("stencil neighborhood option should be a tuple"
-                        " with constant structure such as ((-w, w),)")
+                raise ValueError("stencil neighborhood option should be a tuple"
+                                 " with constant structure such as ((-w, w),)")
         if 'index_offsets' in options:
             fixed = guard(self._fix_stencil_index_offsets, options)
             if not fixed:
-               raise ValueError("stencil index_offsets option should be a tuple"
-                        " with constant structure such as (offset, )")
+                msg = ("stencil index_offsets option should be a "
+                       "tuple with constant structure such as (offset, )")
+                raise ValueError(msg)
         sf = StencilFunc(kernel_ir, 'constant', options)
         sf.kws = expr.kws # hack to keep variables live
         sf_global = ir.Global('stencil', sf, expr.loc)
@@ -246,6 +247,7 @@ class InlineClosureCallPass(object):
                             callee_validator=callee_ir_validator)
         return True
 
+
 def check_reduce_func(func_ir, func_var):
     """Checks the function at func_var in func_ir to make sure it's amenable
     for inlining. Returns the function itself"""
@@ -260,10 +262,10 @@ def check_reduce_func(func_ir, func_var):
         # pull out the python function for inlining
         reduce_func = reduce_func.value.py_func
     elif not (hasattr(reduce_func, 'code')
-            or hasattr(reduce_func, '__code__')):
+              or hasattr(reduce_func, '__code__')):
         raise ValueError("Invalid reduction function")
     f_code = (reduce_func.code if hasattr(reduce_func, 'code')
-                                    else reduce_func.__code__)
+              else reduce_func.__code__)
     if not f_code.co_argcount == 2:
         raise TypeError("Reduction function should take 2 arguments")
     return reduce_func
@@ -335,7 +337,6 @@ class InlineWorker(object):
         self.typemap = typemap
         self.calltypes = calltypes
 
-
     def inline_ir(self, caller_ir, block, i, callee_ir, callee_freevars,
                   arg_typs=None):
         """ Inlines the callee_ir in the caller_ir at statement index i of block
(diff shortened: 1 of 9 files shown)


gmarkall:
Thanks for the PR - see also #7001 - if there's anything in that PR that's useful / you want to pull in to here, feel free to do so :-)

gmarkall:
> Before removing all 4, I thought best to check with someone more familiar with the codebase that such imports aren't needed for e.g. side-effect reasons, although I can see nothing obvious.

Given that all tests on CI passed except for flake8, I'd suggest removing these imports, pushing to this branch again, and seeing if CI stays green. My expectation is that these imports are side-effect free.

GEbb4:
> Thanks for the PR - see also #7001 - if there's anything in that PR that's useful / you want to pull in to here, feel free to do so :-)

@gmarkall thanks! I'll certainly bring in some of those changes when I get to those files.

> Given that all tests on CI passed except for flake8, I'd suggest removing these imports, pushing to this branch again, and seeing if CI stays green. My expectation is that these imports are side-effect free.

As it happened, tests started failing because I removed the `numpy` import. I've re-added it, ignored it in `flake8`, and added a comment explaining why.

DrTodd13:
@GEbb4 Was the numpy import needed because of some dynamic code generation?  My suggestion was going to be the same as Graham's.  Just comment out the things that look unnecessary and then try it.  I'm not aware of any imports in parfors that are strictly for side-effects.  Thanks for doing this work.

GEbb4:
@DrTodd13 Uncertain - I just saw "Name np is not defined" from a test on the `stencil.py` file and went back and replaced the import (I probably should have done a bit more digging to be honest). I'm not too familiar with Azure, I'm not sure if it's possible to go back and look at previously failed test runs, but it's likely an easily reproducible failure.

Since that initial file, I've also gone through the process of setting up a suitable local dev environment (although I followed through the tutorial [here](https://numba.readthedocs.io/en/stable/developer/contributing.html), which uses py36, and a luck would have it many `stencil` tests are for py37 and later...), so I can at least run some of the tests before committing.

DrTodd13:
> @DrTodd13 Uncertain - I just saw "Name np is not defined" from a test on the `stencil.py` file and went back and replaced the import (I probably should have done a bit more digging to be honest). I'm not too familiar with Azure, I'm not sure if it's possible to go back and look at previously failed test runs, but it's likely an easily reproducible failure.
> 
> Since that initial file, I've also gone through the process of setting up a suitable local dev environment (although I followed through the tutorial [here](https://numba.readthedocs.io/en/stable/developer/contributing.html), which uses py36, and a luck would have it many `stencil` tests are for py37 and later...), so I can at least run some of the tests before committing.

I just checked and it is due to dynamic code generation.  This happens in stencil and parfor code so if flake says that something is unused then still do a search for that import in the file to see if it occurs in a text string.  If so, then it is still used.

GEbb4:
> I just checked and it is due to dynamic code generation. This happens in stencil and parfor code so if flake says that something is unused then still do a search for that import in the file to see if it occurs in a text string. If so, then it is still used.

@DrTodd13 ah, yes - got it! Thanks, I'll be sure to keep an eye out for this trope in future files, and I'll update my comment to reflect this.

GEbb4:
I've sorted 8 files here (although there are 9 lines removed from `.flake8` - one of them had no errors). The next on the list, `parfor.py`, would push this to being quite a big review so I'll leave these ones to be reviewed and open a new PR for the next few files.

GEbb4:
@stuartarchibald could you advise on what's needed to mark this as "ready for review"?

gmarkall:
@GEbb4 As it's passing tests I've marked this as ready for review.

numba/core/inline_closurecall.py
```diff
@@ -636,12 +647,17 @@ def inline_closure_call(func_ir, glbls, block, i, callee, typingctx=None,
...
-            pass
+            pass_result = typed_passes.type_inference_stage(typingctx,
+                                                            targetctx,
+                                                            callee_ir,
+                                                            arg_typs, None)
+        except Exception:
+            pass_result = typed_passes.type_inference_stage(typingctx,
+                                                            targetctx,
+                                                            callee_ir,
+                                                            arg_typs, None)
```
stuartarchibald, line 658: Whilst this is clearly a valid refactor, makes me wonder why this is like this!

numba/core/inline_closurecall.py
```diff
@@ -751,9 +768,8 @@ def stararg_handler(index, param, default):
                               x in default_tuple.items]
                 args = args + const_vals
             else:
-                raise NotImplementedError(
-                    "Unsupported defaults to make_function: {}".format(
-                        defaults))
+                msg = "Unsupported defaults to make_function: {}"
+                raise NotImplementedError(msg.format(callee_defaults))
```
stuartarchibald, line 772:
```suggestion
                raise NotImplementedError(msg.format(defaults))
```
? Is the issue that `defaults` doesn't exist or is this a mistake?

numba/core/inline_closurecall.py
```diff
@@ -1031,24 +1063,28 @@ def is_removed(val, removed):
...
             stmts.append(stmt)
     debug_print("removed variables: ", removed)
 
     # Define an index_var to index the array.
-    # If the range happens to be single step ranges like range(n), or range(m, n),
-    # then the index_var correlates to iterator index; otherwise we'll have to
-    # define a new counter.
+    # If the range happens to be single step ranges like range(n), or range(m,
+    # n), then the index_var correlates to iterator index; otherwise we'll have
+    # to define a new counter.
```
stuartarchibald, line 1076:
```suggestion
    # If the range happens to be single step ranges like range(n), or 
    # range(m, n), then the index_var correlates to iterator index; otherwise
    # we'll have to define a new counter.
```
keeps the range spelling on the same line.

numba/core/inline_closurecall.py
```diff
@@ -1382,12 +1452,15 @@ def inline_array(array_var, expr, stmts, list_vars, dels):
         # Create a variable to hold the numpy empty function.
         empty_func = ir.Var(scope, mk_unique_var("empty_func"), loc)
         fnty = get_np_ufunc_typ(np.empty)
-        sig = context.resolve_function_type(fnty, (size_typ,), {'dtype':nptype})
+        # XXX: Does this do anything?
```
stuartarchibald, line 1455: Could well be a side effect that's needed. If you remove it, do all the tests still pass?

numba/core/ir_utils.py
```diff
@@ -718,19 +736,25 @@ def has_no_side_effect(rhs, lives, call_table):
...
+            call_list == ['empty', numpy]
+            or call_list == [slice]
+            or call_list == ['stencil', numba]
+            or call_list == ['log', numpy]
+            or call_list == ['dtype', numpy]
+            or call_list == [array_analysis.wrap_index]
+            or call_list == [prange]
+            or call_list == ['prange', numba]
+            or call_list == [parfor.internal_prange]
+        ):
```
stuartarchibald, line 749: Does this help with `flake8`?

numba/core/ir_utils.py
```diff
@@ -637,8 +644,15 @@ def remove_dead_block(block, lives, call_table, arg_aliases, alias_map,
...
-                     typemap)
+            stmt = f(
+                stmt,
+                lives,
+                lives_n_aliases,
+                arg_aliases,
+                alias_map,
+                func_ir,
+                typemap,
+            )
```
stuartarchibald, line 655: Does this help with `flake8`? Would a line wrap suffice?

numba/core/ir_utils.py
```diff
@@ -764,9 +790,11 @@ def is_pure(rhs, lives, call_table):
...
                 return False
             call_list = call_table[func_name]
-            if (call_list == [slice] or
-                call_list == ['log', numpy] or
-                call_list == ['empty', numpy]):
+            if (
+                call_list == [slice]
+                or call_list == ['log', numpy]
+                or call_list == ['empty', numpy]
+            ):
```
stuartarchibald, line 797: Does this help with `flake8`?

numba/core/ir_utils.py
```diff
@@ -819,20 +852,31 @@ def find_potential_aliases(blocks, args, typemap, func_ir, alias_map=None,
...
                 # a = b.c.  a should alias b
-                if (isinstance(expr, ir.Expr) and expr.op == 'getattr'
-                        and expr.attr not in ['shape']
-                        and expr.value.name in arg_aliases):
+                if (
+                    isinstance(expr, ir.Expr)
+                    and expr.op == 'getattr'
+                    and expr.attr not in ['shape']
+                    and expr.value.name in arg_aliases
+                ):
```
stuartarchibald, line 879: Does this help with `flake8`? Is this coming from an auto-formatter? I'm not sure it's actually making it more readable.

numba/core/ir_utils.py
```diff
@@ -848,7 +892,12 @@ def find_potential_aliases(blocks, args, typemap, func_ir, alias_map=None,
                         alias_func = alias_func_extensions[fdef]
                         alias_func(lhs, expr.args, alias_map, arg_aliases)
                     if fmod == 'numpy' and fname in np_alias_funcs:
-                        _add_alias(lhs, expr.args[0].name, alias_map, arg_aliases)
+                        _add_alias(
+                            lhs,
+                            expr.args[0].name,
+                            alias_map,
+                            arg_aliases,
```
stuartarchibald, line 899: Does this help with `flake8`?

numba/core/ir_utils.py
```diff
@@ -1693,8 +1769,15 @@ def compile_to_numba_ir(mk_func, glbls, typingctx=None, targetctx=None,
...
-                typingctx, targetctx, f_ir, arg_typs, None)
+        f_typemap, f_return_type, f_calltypes, _ = (
+            typed_passes.type_inference_stage(
+                typingctx,
+                targetctx,
+                f_ir,
+                arg_typs,
+                None,
+            )
+        )
```
stuartarchibald, line 1780: Maybe create a temporary for the outputs and subsequently unpack? The braces around the function call seems strange.

numba/core/ir_utils.py
```diff
@@ -2010,14 +2118,17 @@ def raise_on_unsupported_feature(func_ir, typemap):
...
+            # Raise an exception when len(tuple) > 1000. The choice of this
+            # number (1000) was entirely arbitrary
+            msg = (
+                "Tuple '{}' length must be smaller than 1000.\nLarge tuples "
+                "lead to the generation of a prohibitively large LLVM IR which "
+                "causes excessive memory pressure and large compile times.\nAs "
+                "an alternative, the use of a 'list' is recommended in place of"
+                " a 'tuple' as lists do not suffer from this problem.".format(
+                    arg_name,
+                )
```
stuartarchibald, line 2130:
```suggestion
            msg = (
                "Tuple '{}' length must be smaller than 1000.\nLarge tuples "
                "lead to the generation of a prohibitively large LLVM IR which "
                "causes excessive memory pressure and large compile times.\nAs "
                "an alternative, the use of a 'list' is recommended in place "
                "of a 'tuple' as lists do not suffer from this problem.".format(
                    arg_name,
                )
```

numba/stencils/stencil.py
```diff
@@ -160,33 +176,47 @@ def add_indices_to_kernel(self, kernel, index_names, ndim,
...
+                        "Assignments to arrays passed to stencil kernels is "
+                        "not allowed."
+                    )
+                if (
+                    isinstance(stmt, ir.Assign)
+                    and isinstance(stmt.value, ir.Expr)
+                    and stmt.value.op in ['getitem', 'static_getitem']
+                    and stmt.value.value.name in kernel.arg_names
+                    and stmt.value.value.name not in standard_indexed
+                ):
```
stuartarchibald, line 209: I'm not sure about this change. Previously, it was easy to spot that the two branches were checking doing very similar things with the difference being a set/get change.

numba/stencils/stencil.py
```diff
@@ -160,33 +176,47 @@ def add_indices_to_kernel(self, kernel, index_names, ndim,
...
                     if stmt.value.op == 'getitem':
                         stmt_index_var = stmt.value.index
                     else:
                         stmt_index_var = stmt.value.index_var
                         # allow static_getitem since rewrite passes are applied
-                        #raise ValueError("Unexpected static_getitem in add_indices_to_kernel.")
+                        # raise ValueError(
+                        #     "Unexpected static_getitem in "
+                        #     "add_indices_to_kernel."
+                        # )
```
stuartarchibald, line 219: Can probably just remove this. CC @DrTodd13

numba/stencils/stencil.py
```diff
@@ -210,36 +242,72 @@ def add_indices_to_kernel(self, kernel, index_names, ndim,
...
+                                ir.Assign(
+                                    ir.Expr.getitem(
+                                        stmt.value.value,
+                                        tmpvar,
+                                        loc,
+                                    ),
+                                    stmt.target,
+                                    loc,
+                                ),
+                            )
```
stuartarchibald, line 306: Same question about autoformatters.

numba/stencils/stencil.py
```diff
@@ -326,38 +438,46 @@ def add_indices_to_kernel(self, kernel, index_names, ndim,
...
+            self._targetctx,
+            self.kernel_ir,
+            argtys,
+            None,
+            {},
+        )
         if isinstance(return_type, types.npytypes.Array):
             raise ValueError(
-                "Stencil kernel must return a scalar and not a numpy array.")
+                "Stencil kernel must return a scalar and not a numpy array."
```
stuartarchibald, line 472:
```suggestion
                "Stencil kernel must return a scalar and not a NumPy array."
```

stuartarchibald (commented):
Thanks for the patch and for working on improving the formatting of the code base, this change largely looks good. I've made a few comments where I'm unsure that a particular change improves the situation and have added some queries about the use of auto-formatters. Many thanks for doing this!

numba/core/inline_closurecall.py
GEbb4: It is rather a strange paradigm... Although not one I intend to understand as part of this change :-)

numba/core/inline_closurecall.py
GEbb4: There is, as far as I can tell, no variable in this context called `defaults`. I presumed it was a typo, because `callee_defaults` is actually either `callee.defaults` or `callee.__defaults__`, which were the only things that could concievably be meant as "defaults", and the `if` statement is looking at `callee_defaults`. See the following elided snippet:
```python
if callee_defaults:
    debug_print("defaults = ", callee_defaults)
    if isinstance(callee_defaults, tuple):  # Python 3.5
        pass # Elided, this is just for GitHub syntax highlighting.
    elif (isinstance(callee_defaults, ir.Var)
                    or isinstance(callee_defaults, str)):
        pass # Elided, this is just for GitHub syntax highlighting.
    else:
        raise NotImplementedError(
            "Unsupported defaults to make_function: {}".format(
                defaults))
```
I'm given further confidence that this was the intended meaning by the log message at the top - printing "defaults =" then `callee_defaults` would suggest that an "Unsupported defaults" would also refer to `callee_defaults`.

numba/core/ir_utils.py
GEbb4: I don't think so, but it does tend towards the code style enforced by Black, which IMO is most readable. In fact, once flake8 compliance is better I think it would benefit the project to add Black as a second linter.

If you strongly disagree, I can change these back, but I think having the logical operators lined up like this makes it much easier to understand what's happening.

numba/core/ir_utils.py
GEbb4: See above comment - flake8 is happy to just have a line wrap, this is done for compliance with Black which IMO is more readable.

numba/core/ir_utils.py
GEbb4: See [above comment](https://github.com/numba/numba/pull/7195#discussion_r706624366).

numba/core/ir_utils.py
GEbb4: It's not from an auto-formatter, but it is inspired by the format of the autoformatter Black - see [above comment](https://github.com/numba/numba/pull/7195#discussion_r706624366). IMO it's considerably more readable to layout conditions like this, but I get the impression you don't agree - I can revert these if you'd prefer.

numba/stencils/stencil.py
GEbb4: I can see that it's slightly more difficult to compare the first and second `if` statements, but equally I think the old version of the first `if` was extremely difficult to understand - particularly the indentation made it troublesome to wrap my head around when I first tried to wrap my head around it. And there were some `flake8` complaints about indentation of conditionals too.

PR #7196, opened 2021-07-09
Fixes for lineinfo emission

stuartarchibald:
As title.

Includes: 2d63784 from https://github.com/numba/numba/pull/7177 as 3631bdb

### DIFF
--- a/numba/tests/support.py
+++ b/numba/tests/support.py
@@ -22,6 +22,7 @@ import traceback
 from contextlib import contextmanager
 import uuid
 import importlib
+import types as pytypes
 
 import numpy as np
 
@@ -112,6 +113,13 @@ except ImportError:
     has_blas = False
 
 needs_blas = unittest.skipUnless(has_blas, "BLAS needs SciPy 1.0+")
+
+# Decorate a test with @needs_subprocess to ensure it doesn't run unless the
+# `SUBPROC_TEST` environment variable is set. Use this in conjunction with:
+# TestCase::subprocess_test_runner which will execute a given test in subprocess
+# with this environment variable set.
+_exec_cond = os.environ.get('SUBPROC_TEST', None) == '1'
+needs_subprocess = unittest.skipUnless(_exec_cond, "needs subprocess harness")
 
 
 def ignore_internal_warnings():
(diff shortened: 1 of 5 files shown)


sklam:
I can confirmed that the decref is no longer jumping back to the line of object creation. I used the following code to test the decref DWARF:

```python


@njit(debug=True)
def foo(n):
    #gdb()
    ar = np.arange(n)
    if ar.sum() == 0:    # lifetime of ar ends after this line
        ret = 2
    else:
        ret = 3
    return ret

```

stuartarchibald:
@sklam thanks for checking. I've patched the disassembly support here: https://github.com/stuartarchibald/numba/commit/ee91dd89747953fe9e0f21c336ed763c5e68e752 to include DWARF info in the disassembly CFG.

sklam:
Cool, i'll try it now

numba/tests/test_debuginfo.py
```diff
@@ -78,6 +78,25 @@ def _get_metadata(self, fn, sig):
...
+        env_copy = os.environ.copy()
+        env_copy['SUBPROC_TEST'] = '1'
+        env_copy['NUMBA_OPT'] = '0'
+        status = subprocess.run(cmd, stdout=subprocess.PIPE,
+                                stderr=subprocess.PIPE, timeout=60,
+                                env=env_copy, universal_newlines=True)
+        self.assertEqual(status.returncode, 0)
+        self.assertIn('OK', status.stderr)
+        self.assertTrue('FAIL' not in status.stderr)
+        self.assertTrue('ERROR' not in status.stderr)
```
sklam, line 98: a duplication of https://github.com/numba/numba/blob/0c499bfff7ebe4fe5d8a6c1d20653e69f1f2a639/numba/tests/test_parfors.py#L73-L86

we should dedup here or in a immediate follow up PR

numba/tests/support.py
```diff
@@ -502,6 +510,46 @@ def run_nullary_func(self, pyfunc, flags):
...
+        env_copy.update(envvars)
+        status = subprocess.run(cmd, stdout=subprocess.PIPE,
+                                stderr=subprocess.PIPE, timeout=timeout,
+                                env=env_copy, universal_newlines=True)
+        streams = (f'\ncaptured stdout: {status.stdout}\n'
+                   f'captured stderr: {status.stderr}')
+        self.assertEqual(status.returncode, 0, streams)
+        self.assertIn('OK', status.stderr)
+        self.assertTrue('FAIL' not in status.stderr)
+        self.assertTrue('ERROR' not in status.stderr)
```
sklam, line 551:
```suggestion
        self.assertNotIn('ERROR', status.stderr)
```

numba/tests/test_debuginfo.py
```diff
@@ -263,24 +265,121 @@ def test_DILocation_entry_blk(self):
...
+        full_ir = self._get_llvmir(foo, sig=sig)
+
+        # make sure decref lines end with `meminfo.<number>)`.
+        count = 0
+        for line in full_ir.splitlines():
+            line_stripped = line.strip()
+            if line_stripped.startswith('call void @NRT_decref'):
+                self.assertRegex(line, r'.*meminfo\.[0-9]+\)$')
+                count += 1
+        self.assertTrue(count > 0) # make sure there were some decrefs!
```
sklam, line 305:
```suggestion
        self.assertGreater(count, 0) # make sure there were some decrefs!
```

numba/tests/test_debuginfo.py
```diff
@@ -263,24 +265,121 @@ def test_DILocation_entry_blk(self):
...
+            if a[0] == 0:
+                sink(x)
+                return 12
+            z = x[0][0]
+            return z
+
+        sig = (types.float64[::1],)
+        full_ir = self._get_llvmir(foo, sig=sig)
+
+        # make sure decref lines end with `meminfo.<number>)`.
```
sklam, line 298:
```suggestion
        # make sure decref lines end with `meminfo.<number>)` without dbg info.
```

sklam:
I wonder if we should leave a note about how the tests are sensitive to SSA and lowering behavior such that they may become outdated when those are changed.

stuartarchibald:
> I wonder if we should leave a note about how the tests are sensitive to SSA and lowering behavior such that they may become outdated when those are changed.

Good point, I've added one as part of 8f1c198. All other comments should be addressed in there too. Thanks for the review.

sklam (approved):
Thanks for the patch!

PR #7197, opened 2021-07-09
don't post to python announce on the first RC

esc:
So as not to spam with many RC announcements.


stuartarchibald (approved):
Thanks for the patch.

PR #7202, opened 2021-07-12
Initial implementation of np.take_along_axis

itamarst:
This adds basic `np.take_along_axis()` support, with some code initially sketched by @stuartarchibald. 

Unfortunately, I haven't figured out how to support non-literal axis values, which is an unfortunate limitation. Suggestions are welcome, but getting arbitrary-sized tuples seems hard to impossible, and the other implementation seems to require fancy indexing Numba doesn't do yet (though I could be wrong).


sklam:
There're some flake8 error: https://dev.azure.com/numba/numba/_build/results?buildId=9256&view=logs&jobId=2be11828-8e3f-59d7-8855-57e072d2b7b4&j=2be11828-8e3f-59d7-8855-57e072d2b7b4&t=569fa040-6c55-5934-00f5-0b9b1c783979

itamarst:
@sklam fixed the flakes, thanks.

gmarkall:
Many thanks for the PR, I've now marked it for review.

stuartarchibald:
@itamarst Any chance you could resolve the merge conflicts please so as to make it possible to review this? Many thanks for your help.

itamarst:
@stuartarchibald ok tried to do that.

itamarst:
I'll go fix flake8 too.

numba/np/arrayobj.py
```diff
@@ -5439,3 +5439,92 @@ def impl(arr, axis1, axis2):
...
+        else:
+            if Nk:
+                def take_along_axis_impl(arr, indices, axis):
+                    return _take_along_axis_impl_set_nk(
+                        arr, indices, axis, Ni, Nk
+                    )
+            else:
+                def take_along_axis_impl(arr, indices, axis):
+                    return _take_along_axis_impl(arr, indices, axis, Ni, Nk)
+    return take_along_axis_impl
```
stuartarchibald, line 5530: I think the above can be collapsed into something like the following: https://github.com/stuartarchibald/numba/commit/44451e3c7c72299c241c18cf8cf8ae8a2152aa41

Numba can spot quite a few common patterns like "is argument X None" or "Is the length of this tuple > const" and specialise the code based on this such that it can survive type inference. Have a look and see what you think, you are welcome to incorporate the patch/ideas.

numba/tests/test_np_functions.py
```diff
@@ -4216,6 +4216,97 @@ def test_swapaxes_exception(self):
...
+        @njit
+        def axis_minus_one(a, i):
+            return np.take_along_axis(a, i, axis=-1)
+
+        self.assertPreciseEqual(axis_minus_one(a, ai3),
+                                axis_minus_one.py_func(a, ai3))
+
+    def test_take_along_axis_exceptions(self):
+        arr2d = np.arange(8).reshape(2, 4)
+        indices = np.array([0, 1], dtype=np.uint64)
```
stuartarchibald, line 4270: Are these inputs valid for any valid axis (I tried with NumPy and it seems not?)? Whilst this section is testing types, for debugging purposes ensuring valid input will likely help.

numba/np/arrayobj.py
```diff
@@ -5439,3 +5439,92 @@ def impl(arr, axis1, axis2):
...
+                a_1d = arr[ii + np_s_ + kk]
+                indices_1d = indices[ii + np_s_ + kk]
+                out_1d = out[ii + np_s_ + kk]
+                for j in range(J):
+                    out_1d[j] = a_1d[indices_1d[j]]
+    return out
+
+
+@overload(np.take_along_axis, jit_options={'boundscheck':True})
+def arr_take_along_axis(arr, indices, axis):
```
stuartarchibald, line 5487: This should check that `arr` and `indices` are array types, along with a unittest adding to check the type handling. I think `indices` also has to have an integer dtype as it's used as an index, appropriate type checks and unittests should be added for this.

numba/tests/test_np_functions.py
```diff
@@ -4216,6 +4216,97 @@ def test_swapaxes_exception(self):
...
+
+        self.assertPreciseEqual(axis_three(a, ai3), axis_three.py_func(a, ai3))
+
+        @njit
+        def axis_minus_one(a, i):
+            return np.take_along_axis(a, i, axis=-1)
+
+        self.assertPreciseEqual(axis_minus_one(a, ai3),
+                                axis_minus_one.py_func(a, ai3))
+
```
stuartarchibald, line 4267: This:
```python
from numba import njit
import numpy as np

# Based on https://github.com/numpy/numpy/blob/v1.21.0/numpy/lib/tests/test_shape_base.py#L74-L79
a  = np.ones((3, 4, 1))
ai = np.ones((1, 2, 5), dtype=np.intp)

@njit
def check(a, i):
    return np.take_along_axis(a, i, axis=1)

actual = check(a, ai)
assert actual.shape == (3, 2, 5)
```
Seems to segfault locally.

numba/tests/test_np_functions.py
```diff
@@ -4216,6 +4216,97 @@ def test_swapaxes_exception(self):
...
+            return np.take_along_axis(a, i, axis=3)
+
+        self.assertPreciseEqual(axis_three(a, ai3), axis_three.py_func(a, ai3))
+
+        @njit
+        def axis_minus_one(a, i):
+            return np.take_along_axis(a, i, axis=-1)
+
+        self.assertPreciseEqual(axis_minus_one(a, ai3),
+                                axis_minus_one.py_func(a, ai3))
```
stuartarchibald, line 4266:
```suggestion
        def gen(axis):
            @njit
            def impl(a, i):
                return np.take_along_axis(a, i, axis)
            return impl

        for i in range(-1, a.ndim):
            jfunc = gen(i)
            ai = np.argsort(a, axis=i)
            self.assertPreciseEqual(jfunc(a, ai), jfunc.py_func(a, ai))
```
Perhaps?

numba/tests/test_np_functions.py
```diff
@@ -4216,6 +4216,97 @@ def test_swapaxes_exception(self):
         self.assertIn('The third argument "axis2" is out of bounds for array'
                       ' of given dimension', str(raises.exception))
 
+    def test_take_along_axis(self):
+        a = np.arange(24).reshape((3, 1, 4, 2))
+
+        # For now axis must be literal, so we need to define a bunch of
+        # functions.
```
stuartarchibald, line 4223:
```suggestion
        # For now axis must be literal, test explicitly defined implementations
```

stuartarchibald (commented):
Many thanks for working on this involved NumPy function implementation. I've given this an initial review and have made some suggestions that will hopefully simplify the implementation. The function seems to work as expected for most use cases, which is great given the complexity (note though that there's a segfault in a more rare case, I've put a MWR in the comments). Other than that there's some type checking stuff to look at. Thanks again for implementing this, am looking forward to getting support for some of these more complex functions into the code base!

numba/tests/test_np_functions.py
stuartarchibald:
```
Program received signal SIGSEGV, Segmentation fault.
0x00007fffd53f0111 in numba::np::arrayobj::_take_along_axis_impl$244 () at arrayobj.py:5482
5482                        out_1d[j] = a_1d[indices_1d[j]]
```

stuartarchibald:
```
$ NUMBA_BOUNDSCHECK=1 python test_7202.py
debug: IndexError: index 1 is out of bounds for axis 0 with size 1
  File "test_7202.py", line 12, in <module>
    actual = check(a, ai)
IndexError: index is out of bounds
```
Looks like it's an invalid array access causing this.

numba/np/arrayobj.py
```diff
@@ -5439,3 +5439,92 @@ def impl(arr, axis1, axis2):
...
+
+@register_jitable
+def _take_along_axis_impl(arr, indices, axis, Ni, Nk):
+    # Based on example code in
+    # https://numpy.org/doc/stable/reference/generated/numpy.take_along_axis.html
+    J = indices.shape[axis]  # Need not equal M
+    out = np.empty(Ni + (J,) + Nk, arr.dtype)
+
+    for ii in np.ndindex(Ni):
+        for ii in np.ndindex(Ni):
```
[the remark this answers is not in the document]
itamarst: That's a (probably copy/paste) bug, should only have one loop over Ni.

LICENSES.third-party
```diff
@@ -456,6 +456,7 @@ NumPy (https://github.com/numpy/numpy)
 Numba source code that references URLs starting with:
 
 https://github.com/numpy/numpy
+https://numpy.org/doc/
```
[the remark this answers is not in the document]
itamarst: Good point, will update link accordingly.

numba/np/arrayobj.py
```diff
@@ -5439,3 +5439,92 @@ def impl(arr, axis1, axis2):
...
+        Ni = tuple_setitem(Ni, i, arr.shape[i])
+    for i in range(len(Nk)):
+        Nk = tuple_setitem(Nk, i, arr.shape[axis + 1 + i])
+    return _take_along_axis_impl(arr, indices, axis, Ni, Nk)
+
+
+@register_jitable
+def _take_along_axis_impl(arr, indices, axis, Ni, Nk):
+    # Based on example code in
+    # https://numpy.org/doc/stable/reference/generated/numpy.take_along_axis.html
```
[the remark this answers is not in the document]
itamarst: Did the equivalent, yeah.

numba/np/arrayobj.py
itamarst: I can try, but my experience is that Numba gets very angry at any use of tuple_setitem where the tuple length is zero, _even_ if the logic is such that it will never be called. Possibly an `if len(t) > 0` will work better than a `for i in range(len(t))` though, if it's catching only specific hardcoded patterns.

stuartarchibald: Dead loop elimination is https://github.com/numba/numba/issues/7245, which could eventually handle removal of known dead loops. The already existing dead branch pruning will pick up what it can infer from types and eliminate anything it can concretely work out as dead. In general, type inference will assume that all code is reachable and will attempt to propagate types, it has to do this static analysis so as to create a concretely typed specialisation that it can compile.


The code in the patch above is a translation of what's in this PR into something that makes use of more of what the compiler can work out so as to shrink the code.

numba/tests/test_np_functions.py
itamarst: They're valid when `axis=None`.

numba/tests/test_np_functions.py
itamarst: Turns out the alternative `take_along_axis` implementation in NumPy docs (which this PR is based on) doesn't match real implementation in some edge cases. I filed https://github.com/numpy/numpy/issues/19704

itamarst:
Going to take a break and see if NumPy maintainers have any suggestions, otherwise I will try to figure out how the two algorithms differ.

numba/tests/test_np_functions.py
itamarst: OK, thanks to @eric-wieser I think I can make this work in regular Python... but Numba is missing `np.broadcast_to()` (though there's a private function?) and is also missing the more complex `nditer()` options. So next is seeing if I can implement matching code some other way that Numba does support.

gmarkall: Numba now has `broadcast_to`: #7119.

itamarst: OK, done, but you should check my logic.

numba/np/arrayobj.py
itamarst: I did more or less the equivalent in my PR. Couldn't quite use it because in the axis=None case the array has to be flattened.

itamarst:
@stuartarchibald should be ready for another review, assuming everything passes.

numba/np/arrayobj.py
```diff
@@ -5534,3 +5534,105 @@ def impl(arr, axis1, axis2):
         return np.transpose(arr, axes_tuple)
 
     return impl
+
+
+@njit
```
stuartarchibald, line 5539:
```suggestion
@register_jitable
```

numba/np/arrayobj.py
```diff
@@ -5534,3 +5534,105 @@ def impl(arr, axis1, axis2):
...
+def _take_along_axis_impl(
+        arr, indices, axis, Ni_orig, Nk_orig, indices_broadcast_shape
+):
+    # Based on example code in
+    # https://github.com/numpy/numpy/blob/623bc1fae1d47df24e7f1e29321d0c0ba2771ce0/numpy/lib/shape_base.py#L90
```
stuartarchibald, line 5544:
```suggestion
    # https://github.com/numpy/numpy/blob/623bc1fae1d47df24e7f1e29321d0c0ba2771ce0/numpy/lib/shape_base.py#L90-L103
```

numba/np/numpy_support.py
```diff
@@ -673,3 +673,9 @@ def type_can_asarray(arr):
           types.Number, types.Boolean, types.containers.ListType)
 
     return isinstance(arr, ok)
+
+
+def check_is_integer(v, name):
+    """Raise TypeError if the value is not an integer."""
```
stuartarchibald, line 679:
```suggestion
    """Raises TypingError if the value is not an integer."""
```

stuartarchibald (commented):
@itamarst many thanks for the update. I've taken a look at the new code (broadcasting part) and other patches since the first review. There's couple of minor things to resolve else looks good. Thanks again for working on this.

stuartarchibald (approved):
@itamarst thanks again for your efforts in implementing this somewhat difficult piece of NumPy functionality. Patch looks good!

Issue #7172, opened 2021-06-30
Prune 0.54.0 RC1 Changelog


esc:
Due to some bugs in the algorithm we now use to automatically generate the change-log, some unexpected PRs remain, especially the PRs that are about changelog changes for 0.53.* should be removed.

esc:
#7203 was merged, closing now.

PR #7203, opened 2021-07-12
remove duplicate changelog entries

esc:
As title

#7172


esc:
I discovered these by carefully looking at the git history, it would probably be good if someone else were to verify them.

esc:
Looks like we missed this one for RC2 -- must go in for RC3.

sklam (approved):
Thanks for the patch!

I have manually confirmed that the removed lines either contain duplicated PRs from 0.53 or that the PRs are corresponding to the 0.53 release but was not mentioned back then.

PR #7204, opened 2021-07-12
Adding implementation of isinstance.

DrTodd13:
I've only checked this for simple cases to get the discussion rolling.
We need to think of cases that we think might break it and even if it doesn't then that is the set of things we can test for.
Both individual and tuples of types are supported.

### DIFF
--- a/numba/cpython/builtins.py
+++ b/numba/cpython/builtins.py
@@ -19,7 +19,6 @@ from numba.core.typeconv import Conversion
 from numba.core.errors import TypingError
 from numba.misc.special import literal_unroll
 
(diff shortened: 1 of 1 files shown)


stuartarchibald:
Thanks for the starter patch @DrTodd13, @guilhermeleobas is taking a look at how to extend this etc, thanks @guilhermeleobas!

stuartarchibald:
@guilhermeleobas implemented this fully in https://github.com/numba/numba/pull/7244, closing as complete. Thanks again for the initial ideas @DrTodd13.

PR #7208, opened 2021-07-14
Fix unicode encoding error in debug metadata in llvmlite.

sklam:
Unicode characters in symbol names are replaced with '?'. Not sure if gdb will allow looking up unicode symbol names.

The original issue can be demonstrated by running: `NUMBA_DEBUGINFO=1 python runtests.py numba.tests.test_unicode_names.TestUnicodeNames`

### DIFF
--- a/numba/core/debuginfo.py
+++ b/numba/core/debuginfo.py
@@ -15,8 +15,13 @@ from numba.core import cgutils
 def _escape_unicode(x):
     """Escape unicode chars used in LLVM debug metadata.
     """
-    # We will simply replace them with '?' for now.
-    return x.encode('ascii', 'replace').decode()
+    def repl(x):
+        # GDB uses \NNN where NNN is octal representation of the character.
+        s = chr(x)
+        return s if s.isascii() else r'\{:03o}'.format(x)
+    utf8chars = x.encode()
+    out = ''.join(map(repl, utf8chars))
+    return out
 
 
 @contextmanager
--- a/numba/core/debuginfo.py
+++ b/numba/core/debuginfo.py
@@ -10,6 +10,15 @@ from contextlib import contextmanager
 from llvmlite import ir
 from numba.core import cgutils
 
+
+
+def _escape_unicode(x):
+    """Escape unicode chars used in LLVM debug metadata.
+    """
+    # We will simply replace them with '?' for now.
+    return x.encode('ascii', 'replace').decode()
+
+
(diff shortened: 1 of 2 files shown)


numba/core/debuginfo.py
```diff
@@ -10,6 +10,15 @@
 from llvmlite import ir
 from numba.core import cgutils
 
+
+
+def _escape_unicode(x):
+    """Escape unicode chars used in LLVM debug metadata.
+    """
+    # We will simply replace them with '?' for now.
+    return x.encode('ascii', 'replace').decode()
```
sklam, line 19: Note to self, gdb uses octal escape sequence: `\nnn` where `nnn` are octal repr of utf8-encoded code point.

e.g.

```
>>> list(map(oct, 'ø'.encode('utf8')))
['0o303', '0o270']
```
the escaped sequence is `\303\270`

stuartarchibald:
Whilst the variable names in `DW_AT_name` appear in something octal like, it seems like the LLVM toolchain encodes them as something bytes-like in the `DILocalVariable`'s `name` field in the LLVM IR. When this IR is compiled, the resulting ELF then ends up with the octal based encoding in the DWARF.

Example:
```
$ cat test_unicode.c 

int main (void){
    int \u00F8 = 10;
    return \u00F8;
}

$ clang -O0 -g -S -emit-llvm -c test_unicode.c

$ cat test_unicode.ll |grep DILocalVariable
!11 = !DILocalVariable(name: "\C3\B8", scope: !7, file: !1, line: 3, type: !10)

$ clang -O0 -g test_unicode.c -o test_unicode

$ llvm-dwarfdump test_unicode|grep DW_AT_name
              DW_AT_name        ("test_unicode.c")
                DW_AT_name      ("main")
                  DW_AT_name    ("\303\270")
                DW_AT_name      ("int")

```

stuartarchibald:
Also suspect that calling this:
https://github.com/numba/llvmlite/blob/09df344ef758174949d582a2c85f2a2d0490e403/llvmlite/ir/values.py#L21
from e.g.
https://github.com/numba/llvmlite/blob/09df344ef758174949d582a2c85f2a2d0490e403/llvmlite/ir/values.py#L276-L277
could well lead to "double" escaping of chars such as `\`.

sklam:
Superseded by https://github.com/numba/llvmlite/pull/751

PR #7209, opened 2021-07-15
Clamp numpy

esc:
As discussed during the developer meeting, we will clamp NumPy to a maximum compatible version for released versions. This implements that.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -50,7 +50,7 @@ Dependencies
 
 * Python versions: 3.7-3.9
 * llvmlite 0.37.*
-* NumPy >=1.17,<=1.20 (can build with 1.11 for ABI compatibility).
+* NumPy >=1.17,<1.21 (can build with 1.11 for ABI compatibility).
 
 Optionally:
 
--- a/README.rst
+++ b/README.rst
@@ -50,7 +50,7 @@ Dependencies
 
 * Python versions: 3.7-3.9
 * llvmlite 0.37.*
-* NumPy >=1.17 (can build with 1.11 for ABI compatibility).
+* NumPy >=1.17,<=1.20 (can build with 1.11 for ABI compatibility).
 
 Optionally:
 
--- a/docs/source/user/installing.rst
+++ b/docs/source/user/installing.rst
@@ -5,8 +5,8 @@ Installation
 Compatibility
 -------------
 
-Numba is compatible with Python 3.7 or later, and Numpy versions from 1.17 to
-1.20.
+Numba is compatible with Python 3.7 or later, and Numpy versions from 1.17 up
+to but excluding 1.21.
 
 Our supported platforms are:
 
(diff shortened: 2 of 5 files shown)


setup.py
```diff
@@ -359,7 +360,7 @@ def check_file_at_path(path2file):
 build_requires = ['numpy >={}'.format(min_numpy_build_version)]
 install_requires = [
     'llvmlite >={},<{}'.format(min_llvmlite_version, max_llvmlite_version),
-    'numpy >={}'.format(min_numpy_run_version),
+    'numpy >={},<={}'.format(min_numpy_run_version, max_numpy_run_version),
```
stuartarchibald, line 363: Think this needs to be `<1.21` so that e.g. `1.20.3` works ok?

esc: oh yes, because `<=1.20` means `<=1.20.0`

esc: fixed in 77f3b

README.rst
```diff
@@ -50,7 +50,7 @@ Dependencies
 
 * Python versions: 3.7-3.9
 * llvmlite 0.37.*
-* NumPy >=1.17 (can build with 1.11 for ABI compatibility).
+* NumPy >=1.17,<=1.20 (can build with 1.11 for ABI compatibility).
```
stuartarchibald, line 53:
```suggestion
* NumPy >=1.17,<1.21 (can build with 1.11 for ABI compatibility).
```

docs/source/user/installing.rst
```diff
@@ -5,7 +5,8 @@ Installation
 Compatibility
 -------------
 
-Numba is compatible with Python 3.7 or later, and Numpy versions 1.17 or later.
+Numba is compatible with Python 3.7 or later, and Numpy versions from 1.17 up
```
stuartarchibald, line 8:
```suggestion
Numba is compatible with Python 3.7 or later, and NumPy versions from 1.17 up
```
though not necessary to get this merged!

stuartarchibald (commented):
Should this get a guard for too high a version as well?
https://github.com/numba/numba/blob/0c499bfff7ebe4fe5d8a6c1d20653e69f1f2a639/numba/__init__.py#L135-L136


Further, what about the conda recipes? e.g.
https://github.com/numba/numba/blob/0c499bfff7ebe4fe5d8a6c1d20653e69f1f2a639/buildscripts/condarecipe.local/meta.yaml#L46
should there be a restriction imposed here?

docs/source/user/installing.rst
esc: fixed in abc4f8d62e94b54221933fc5149aef3a508ede93

sklam:
I confirmed that the wheels are preventing numpy1.21.

```
Installing collected packages: numpy, numba
  Attempting uninstall: numpy
    Found existing installation: numpy 1.21.0
    Uninstalling numpy-1.21.0:
      Successfully uninstalled numpy-1.21.0
Successfully installed numba-0.54.0rc1+9.gabc4f8d62 numpy-1.20.3
```

sklam (approved):
Thanks for the patch!

PR #7216, opened 2021-07-16
Update CHANGE_LOG for 0.54.0rc2

stuartarchibald:
As title.


sklam (approved):
Thanks for the patch!

PR #7222, opened 2021-07-20
CUDA: Warn when debug=True and opt=True

### LINKED ISSUE #7214: cuda segfault nvvm with opt+debug
## Reporting a bug

This segfaults:
```python
from numba import cuda

@cuda.jit(debug=True)
def f(b, n):
    result = 0
    for i in range(n):
        result += b

f[1, 1](1, 2)
```
Under `cuda-gdb` the backtrace points to the `nvvmCompilerProgram` call in `libnvvm.so`. 

Setting `opt=False` or `debug=False` avoids the segfault.

CC @gmarkall

gmarkall:
As per Issue #7214, using `debug=True` with `opt=True` can cause NVVM to crash (or the link can fail with an error message from ptxas). We can't easily stop the user from doing this by default as it might break a lot of existing code (particularly that which uses debug=True to force a check for exceptions) but we can warn them.

This commit adds a warning when `debug` and `opt` are both `True`, and some tests for most cases. The debuginfo tests are also updated so that they don't trigger the warning and make the test suite noisy. However, I chose not to "fix" the tests that check for a raised exception (e.g. in test_exception and test_fastmath) - I think the test suite should be emitting warnings there to highlight/remind about the fact that we need to split `debug=True` into at least two options (e.g. `debuginfo=True` and `exceptions=True`).

Fixes #7214.

### DIFF
--- a/numba/cuda/compiler.py
+++ b/numba/cuda/compiler.py
@@ -18,7 +18,8 @@ from numba.core.compiler_lock import global_compiler_lock
 from numba.core.compiler_machinery import (LoweringPass, PassManager,
                                            register_pass)
 from numba.core.dispatcher import OmittedArg
-from numba.core.errors import NumbaDeprecationWarning
+from numba.core.errors import (NumbaDeprecationWarning,
+                               NumbaInvalidConfigWarning)
 from numba.core.typed_passes import IRLegalization, NativeLowering
 from numba.core.typing.typeof import Purpose, typeof
 from warnings import warn
@@ -212,6 +213,12 @@ def compile_ptx(pyfunc, args, debug=False, lineinfo=False, device=False,
     :return: (ptx, resty): The PTX code and inferred return type
     :rtype: tuple
     """
+    if debug and opt:
+        msg = ("debug=True with opt=True (the default) "
+               "is not supported by CUDA. This may result in a crash"
+               " - set debug=False or opt=False.")
+        warn(NumbaInvalidConfigWarning(msg))
+
     nvvm_options = {
         'debug': debug,
         'lineinfo': lineinfo,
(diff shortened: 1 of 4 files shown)


stuartarchibald (commented):
Thanks for the patch, looks good on inspection. Does `numba.cuda.compile_ptx` need similar, IIRC that's public facing?

gmarkall:
> Thanks for the patch, looks good on inspection. Does `numba.cuda.compile_ptx` need similar, IIRC that's public facing?

Good point - I've added a warning there too. I haven't added a test, because triggering the warning also sends NVVM down the path where it may crash. However, I've verified the change manually:

```python
from numba import cuda

def f():
    pass

cuda.compile_ptx(f, (), debug=True, opt=True)
```

emits:

```
$ python test.py 
/home/gmarkall/numbadev/numba/numba/cuda/compiler.py:220: NumbaInvalidConfigWarning: debug=True with opt=True (the default) is not supported by CUDA. This may result in a crash - set debug=False or opt=False.
  warn(NumbaInvalidConfigWarning(msg))
```

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_88`.

stuartarchibald:
> > Thanks for the patch, looks good on inspection. Does `numba.cuda.compile_ptx` need similar, IIRC that's public facing?
> 
> Good point - I've added a warning there too. I haven't added a test, because triggering the warning also sends NVVM down the path where it may crash. However, I've verified the change manually:
> 
> ```python
> from numba import cuda
> 
> def f():
>     pass
> 
> cuda.compile_ptx(f, (), debug=True, opt=True)
> ```
> 
> emits:
> 
> ```
> $ python test.py 
> /home/gmarkall/numbadev/numba/numba/cuda/compiler.py:220: NumbaInvalidConfigWarning: debug=True with opt=True (the default) is not supported by CUDA. This may result in a crash - set debug=False or opt=False.
>   warn(NumbaInvalidConfigWarning(msg))
> ```

Thanks for adding that, I've also manually verified the same.

stuartarchibald (approved):

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_88`.

Passed.

Issue #7217, opened 2021-07-16
Regression: SSA scope check assertion error


sklam:
If anyone see `AssertionError` from:

https://github.com/numba/numba/blob/master/numba/core/ssa.py#L270

please report to this issue with reproducers.

luk-f-a:
I managed to build a reproducer, probably not as minimal as it can be but I can confirm that they key bit is variable `i`. Removing `i += 1` makes the problem go away.

```python
from collections import namedtuple
import numpy as np
from numba.extending import overload_method
from numba.typed import List
from numba import njit, literal_unroll, typeof, types

first = List([np.array([(1, 2)], dtype=[('val', 'f8'), ('b', 'i8')])])
second = List([np.array([(1, 2)], dtype=[('val', 'f8'), ('d', 'i8')])])

A = namedtuple("A", ["first", "second"])

a = A(first, second)


@overload_method(types.NamedTuple, 'bar')
def bar_ovrl(a):

    def bar_impl(a):
        i = 0
        b = 0
        for x in literal_unroll((a.first, a.second)):
            for i, y in enumerate(x):
                for z in y:
                    b += z['val']
                    i += 1
        return b

    return bar_impl


@njit
def foo(a):
    return a.bar()

print(foo(a))
```

sklam:
@luk-f-a, thank you for finding a reproducer, it makes it a lot easier for us to debug.

stuartarchibald:
Thanks @luk-f-a, much appreciated. I've got this down to:
```python
from numba import njit, literal_unroll

@njit
def foo():
    i = 0
    for _ in literal_unroll((1, 2)):
        for i in range(0):
            i += 1

foo()
```
and think it's coming from when the sentinels in the switch table that's generated are replaced with the versioned loops.

PR #7223, opened 2021-07-20
Replace assertion errors on IR assumption violation

sklam:
with pedantic warnings.

Closes #7217

PR #7228, opened 2021-07-21
Fixes issue reported under #7217

stuartarchibald:
This fixes the issue with ``literal_unroll`` reported in #7217.
Does not close the issue as there's potentially more than just
this problem with ``literal_unroll`` present.

This patch adds a fix to update the scope of blocks in the
versioned loop bodies with variables that have been versioned.

### DIFF
--- a/numba/core/untyped_passes.py
+++ b/numba/core/untyped_passes.py
@@ -910,6 +910,17 @@ class MixedContainerUnroller(FunctionPass):
                 new_var_dict[name] = mk_unique_var(name)
             replace_var_names(loop_blocks, new_var_dict)
 
+            # Patch up the scope with the new variable names, this is horrible
+            new_var_table = get_name_var_table(loop_blocks)
+            for blk in loop_blocks.values():
+                scope = blk.scope
+                real_con = scope.localvars._con
+                for name, var in dict(scope.localvars._con).items():
+                    if name in new_var_dict:
+                        new_name = new_var_dict[name]
+                        del real_con[name]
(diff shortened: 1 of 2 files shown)


stuartarchibald:
NOTE: AZP is maybe having some issues. Have had to request reruns/close-open PR. Deleted comments from `azure-pipelines` bot are just of confirmations of request.

numba/core/untyped_passes.py
```diff
@@ -910,6 +910,17 @@ def inject_loop_body(self, switch_ir, loop_ir, caller_max_label,
...
+            # Patch up the scope with the new variable names, this is horrible
+            new_var_table = get_name_var_table(loop_blocks)
+            for blk in loop_blocks.values():
+                scope = blk.scope
+                real_con = scope.localvars._con
+                for name, var in dict(scope.localvars._con).items():
+                    if name in new_var_dict:
+                        new_name = new_var_dict[name]
+                        del real_con[name]
+                        real_con[new_name] = new_var_table[new_name]
```
sklam, line 922: I probably will make a follow up PR to refactor it with `ir_utils.fixup_var_define_in_scope`

stuartarchibald: Thanks. I think the issue is that the versioned loop bodied have versions of _some_ of the original variables and potentially replace the original.

sklam:
I think https://github.com/numba/numba/pull/7223 resolved the problem this patch is fixing. The test no longer fails on [master](https://github.com/numba/numba/commit/3884e95bdec70c385291bec820beecc5ab64933a)

PR #7229, opened 2021-07-21
DO NOT MERGE: Debug for mypy issue.

stuartarchibald:
As title.


sklam:
`mypy=0.910` is somehow in conflict but conda is not reporting it

PR #7230, opened 2021-07-21
PR #7171 bugfix only

stuartarchibald:
Same as #7171 but just the bug fix components as discussed here: https://github.com/numba/numba/pull/7171#issuecomment-884029933

CC @DrTodd13


DrTodd13 (approved):

sklam (commented):
Thanks for the patch!

sklam (approved):

Issue #7225, opened 2021-07-21
Renamed variable names may not be unique


ehsantn:
## Reporting a bug

- [x] I have tried using the latest ~released~ 0.54rc2 release candidate version of Numba (most recent is
 visible in the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [ ] I have included a self contained code sample to reproduce the problem.
  i.e. it's possible to run as 'python bug.py'.

Using 0.54.0rc2 for one of Bodo's tests causes a typing failure, since some renamed variable names end up being the same in a function (but with different data types). It happens when some function is inlined multiple times (in the same outer function), but with different input types. There is probably some issue with variable scopes in inlining, but making renamed variables globally unique can prevent similar problems in general. Is there a problem with using `ir_utils.next_label()` here instead of starting the counter from one for each variable?https://github.com/numba/numba/blob/d0ca288f24eccff3ba1f86e38f0de07a717f8d4c/numba/core/ir.py#L1142

I have a workaround for this case so this is not a blocker, but I think making renamed variables unique can improve robustness.

esc:
@ehsantn thank you for reporting this. Shame about the lack of reproducer. Do you want me to classify this as a bug or as a feature request?

ehsantn:
A bug probably. Thanks.

PR #7233, opened 2021-07-22
Update postproc to general multiple del nodes if a variable is assigned multiple times in the same basic block

njriasan:
In Bodo we assign to a variable multiple times to support "inplace" updates. For example a sort with `inplace=True` is accomplished by reassigning to the same IR variable after a sort is finished. This change fixes a memory leak that was deteced in Numba 0.54 because only 1 del is generated.

I'll add a reproducer with a Numba pipeline extension shortly.

### DIFF
--- a/numba/core/postproc.py
+++ b/numba/core/postproc.py
@@ -195,6 +195,11 @@ class PostProcessor(object):
                 # used here but not afterwards
                 delete_pts.append((stmt, dead_set))
                 internal_dead_set -= dead_set
+                # Readd the variable to the dead set after any
+                # assignments in case a variable is assigned
+                # to multiple times
+                if isinstance(stmt, ir.Assign):
+                    internal_dead_set.add(stmt.target.name)
 
             # rewrite body and insert dels
             body = []
--- a/numba/core/postproc.py
+++ b/numba/core/postproc.py
@@ -197,9 +197,19 @@ class PostProcessor(object):
                 internal_dead_set -= dead_set
                 # Readd the variable to the dead set after any
                 # assignments in case a variable is assigned
-                # to multiple times
+                # to multiple times. This is not safe if the variable
+                # is also used on the RHS, which while uncommon, seems
+                # to be possible with objmode.
                 if isinstance(stmt, ir.Assign):
-                    internal_dead_set.add(stmt.target.name)
+                    # If the LHS name appears more than once, it must also be on
+                    # the RHS and therefore should not be added. We don't look
+                    # at list_vars for the value because it may not exist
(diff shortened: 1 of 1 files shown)


gmarkall:
Is this to fix #7225?

njriasan:
No this fixes #7258, which I previously described on discourse but hadn't filed the issue. This seems like an easy fix and doesn't seem to interfere with any tests, so could we include it in the numba 0.54 release?

gmarkall:
> I'll add a reproducer with a Numba pipeline extension shortly.

Is this still upcoming?

> This seems like an easy fix and doesn't seem to interfere with any tests, so could we include it in the numba 0.54 release?

I'm not 100% sure that reassignment ought to be legal, as mentioned in #7258 - @sklam's input may be needed here.

njriasan:
> > I'll add a reproducer with a Numba pipeline extension shortly.
> 
> Is this still upcoming?
> 

Yes. I'll attach it in the next comment.

> > This seems like an easy fix and doesn't seem to interfere with any tests, so could we include it in the numba 0.54 release?
> 
> I'm not 100% sure that reassignment ought to be legal, as mentioned in #7258 - @sklam's input may be needed here.

So from our perspective this is incredibly useful because we have some difficulties with inplace updates. For example, we may need to update an immutable array inplace in a way that changes its length (I can elaborate if necessary). We handle this  by updating the reference to where the data is stored, which requires reassignment.

I totally understand if this is not behavior that is needed when just using Numba as a compiler, but since Numba is also a tool for building compilers we believe we have a relevant usecase for why Numba shouldn't complete disallow this behavior. Perhaps there could be a flag used that would allow us to enable this behavior?

njriasan:
Here is the Python file that can be used to generate IR that produces reassign in a way that vaguely resembles our usecase. I tried to simplify it as much as possible.

Github didn't accept just the .py file so I placed it in a zip file. Sorry if that's annoying.
[dummy_reassign_ext.zip](https://github.com/numba/numba/files/6900861/dummy_reassign_ext.zip)

ehsantn:
Currently, I believe Numba's IR is in SSA form only during typing. This change shouldn't affect SSA IR anyways.

sklam:
@njriasan, RE: the reproducer `dummy_reassign_ext.zip`. Where should I be looking to see the effect of the patch?

sklam (commented):
Thank you for the patch and the reproducer. It allowed me to find the true problem. See https://github.com/numba/numba/pull/7267

At every assignment statement in the IR, the old value is always deleted before storing the new value. Therefore, we should never need to have more than 1 del for a variable in a block.

PR #7241, opened 2021-07-23
Set aliasing information for inplace_binops..

### LINKED ISSUE #7201: Setting array elements in parallel after increment fails silently
## Reporting a bug

Setting all elements of an array with the `arr[:] = 1` syntax doesn't seem to work if there is a previous increment statement, *and* the function is compiled with `parallel=True`.  Here's an example:
```python
import numpy as np
import numba

@numba.njit(parallel=True)
def increment_then_set(a):
    a += 1
    a[:] = 3

x = np.ones(1)
increment_then_set(x)
print(x)  # expect 3, get 2
```

Interestingly, numba warns me that no parallel transformations were possible, so I wouldn't expect it to have done anything to the code, but it still seems to introduce the bug!

I saw there were some related-looking parallel bugs in the issue tracker that were fixed by #7171, so I tested on that branch, and it still seems to exhibit the bug (master too).

DrTodd13:
Resolves #7201.

### DIFF
--- a/numba/core/ir_utils.py
+++ b/numba/core/ir_utils.py
@@ -825,6 +825,8 @@ def find_potential_aliases(blocks, args, typemap, func_ir, alias_map=None,
                 if (isinstance(expr, ir.Expr) and (expr.op == 'cast' or
                     expr.op in ['getitem', 'static_getitem'])):
                     _add_alias(lhs, expr.value.name, alias_map, arg_aliases)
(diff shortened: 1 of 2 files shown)


gmarkall:
Queued for 0.55RC as this doesn't appear to be a regression (from the info in the report in #7201 at least)

stuartarchibald (approved):
Thanks for the patch.

PR #7243, opened 2021-07-23
Check for variables in parfor loop nest and init block for global variable pushing.

### LINKED ISSUE #7213: parfors lowering error from missing entry in varmap
## Reporting a bug

This:
```python
from numba import jit, prange
import numpy as np

def foo():
    a = np.empty((2,))
    for k in prange(2):
        b = np.arange(1)
        c = np.zeros((n, n), dtype=np.complex128)
        d = np.expand_dims(b, -1) * c
        e = np.diag(global_arr @ d)
        a[k] = 1
    return a

n = 5
global_arr = np.ones((n, n)).astype(np.complex128)
foo()
```
does this:
```numba.core.errors.LoweringError: Failed in nopython mode pipeline (step: native lowering)
Failed in nopython mode pipeline (step: native lowering)
'$push_global_to_block.300'

File "di85.py", line 9:
def foo():
    <source elided>
        b = np.arange(1)
        c = np.zeros((n, n), dtype=np.complex128)
        ^
```

This is from reduced input as reported here: https://numba.discourse.group/t/jit-decorated-function-fails-compiling-with-parallel-true/793

DrTodd13:
Resolves #7213


gmarkall:
Queued for 0.55 as it doesn't appear to be a regression (or not mentioned in #7213).

numba/tests/test_parfors.py
```diff
@@ -1838,6 +1840,22 @@ def test_impl(x):
...
+                              _GLOBAL_INT_FOR_TESTING2))
+                c = np.zeros((_GLOBAL_INT_FOR_TESTING2,
+                              _GLOBAL_INT_FOR_TESTING2), dtype=np.complex128)
+                d = b * c
+                e = np.diag(_GLOBAL_ISSUE7213_ARR + d)
+                a[k] = 1
+            return a
+
+        self.check(test_impl)
+
```
stuartarchibald, line 1858: I'm struggling to get this to reproduce the original issue. On current [HEAD](https://github.com/numba/numba/tree/5baa1ce53073a786060078a8cbe0ee57b1fb67d2) of mainline this test code seems to work:
```python
$ git rev-parse --short HEAD
5baa1ce

$ cat test_7243.py 
import numpy as np

_GLOBAL_INT_FOR_TESTING1 = 17
_GLOBAL_INT_FOR_TESTING2 = 5
_GLOBAL_ISSUE7213_ARR = np.ones((_GLOBAL_INT_FOR_TESTING2,
                                 _GLOBAL_INT_FOR_TESTING2)).astype(np.complex128)

@njit(parallel=True)
def test_impl():
    a = np.empty((2,))
    for k in prange(2):
        b = np.zeros((_GLOBAL_INT_FOR_TESTING2,
                        _GLOBAL_INT_FOR_TESTING2))
        c = np.zeros((_GLOBAL_INT_FOR_TESTING2,
                        _GLOBAL_INT_FOR_TESTING2), dtype=np.complex128)
        d = b * c
        e = np.diag(_GLOBAL_ISSUE7213_ARR + d)
        a[k] = 1
    return a

print(test_impl())

$ python test_7243.py
[1. 1.]

```

stuartarchibald:
@DrTodd13 thanks for the patch, the fix looks good on inspection, and it fixes the original issue reported in https://github.com/numba/numba/issues/7213#issue-946127754, however [I can't seem to get the unit test that's been added to replicate the problem](https://github.com/numba/numba/pull/7243#discussion_r681703482). Thanks for looking at this.

DrTodd13:
One thing to note is that with the original example it will not work because the expand_dims line has a broadcast in it that is not supported by parallel=True.

numba/tests/test_parfors.py
DrTodd13: @stuartarchibald I also cannot replicate the issue with the latest Numba.  I went back to a July version of Numba and it also didn't replicate there at the time when this issue was originated.  I hesitate to say abandon this PR unless we could prove that the circumstance that this PR is designed to fix can't happen.  I don't know how to force the situation to occur at the moment either.  Suggestions?

stuartarchibald:
Thanks for the update to this patch. Unfortunately I cannot seem to get the new test case to fail against current mainline:
```bash
$ git rev-parse HEAD
8215dffafb7a869280ddd6a0f4c01b49f5cce908

$ cat test7243.py 
import numpy as np

_GLOBAL_INT_FOR_TESTING1 = 17
_GLOBAL_INT_FOR_TESTING2 = 5
_GLOBAL_ISSUE7213_ARR = np.ones((_GLOBAL_INT_FOR_TESTING2,
                                 _GLOBAL_INT_FOR_TESTING2)).astype(np.complex128)

# issue7213
@njit(parallel=True)
def test_impl():
    a = np.empty((2,))
    for k in prange(2):
        b = np.zeros((_GLOBAL_INT_FOR_TESTING2,
                        _GLOBAL_INT_FOR_TESTING2))
        c = np.zeros((_GLOBAL_INT_FOR_TESTING2,
                        _GLOBAL_INT_FOR_TESTING2), dtype=np.complex128)
        d = b * c
        e = np.diag(_GLOBAL_ISSUE7213_ARR + d)
        a[k] = 1
    return a


np.testing.assert_allclose(test_impl(), test_impl.py_func())

$ python test7243.py 

$
```

DrTodd13:
@stuartarchibald I noted the same thing as you in a comment here that I made a couple days ago.  How do you want to proceed here?  I thought a bit about how to create such a test case but I don't know how to do it.  I do know that if somehow a global could be set outside the parfor but used inside that this would be a problem for parallel=True.  Is that impossible now or could some change in the future make it possible then but not now (because certainly people won't be thinking along these lines of moving global lines around and thinking that could be a problem)?  We can remove the test case and commit the rest or just keep this PR in reserve in case this issue comes up again (of course by that point we're likely to forget this PR exists).

stuartarchibald:
> @stuartarchibald I noted the same thing as you in a comment here that I made a couple days ago. 

Just found that, thanks!

> How do you want to proceed here? I thought a bit about how to create such a test case but I don't know how to do it. I do know that if somehow a global could be set outside the parfor but used inside that this would be a problem for parallel=True. Is that impossible now or could some change in the future make it possible then but not now (because certainly people won't be thinking along these lines of moving global lines around and thinking that could be a problem)?

Can you expand on this please? There is a global defined outside the parfor that's used inside in the example?

> We can remove the test case and commit the rest or just keep this PR in reserve in case this issue comes up again (of course by that point we're likely to forget this PR exists).

I'm reluctant to commit code that does not have a test. It may be that we just close this and rely on being able to search for the same error message in the issue tracker which will lead us here.

> One thing to note is that with the original example it will not work because the expand_dims line has a broadcast in it that is not supported by parallel=True.

Seems like there's a few things like this that just lead to incorrect results. Can these be detected and compilation either aborted or maybe the region where the unsupported thing is serialised?

DrTodd13:
> > How do you want to proceed here? I thought a bit about how to create such a test case but I don't know how to do it. I do know that if somehow a global could be set outside the parfor but used inside that this would be a problem for parallel=True. Is that impossible now or could some change in the future make it possible then but not now (because certainly people won't be thinking along these lines of moving global lines around and thinking that could be a problem)?
> 
> Can you expand on this please? There is a global defined outside the parfor that's used inside in the example?

My wording was terrible there....my apologies.  I'm referring to the function's IR.  If you have a prange and inside of that use a global, the question is, where does the ir.Global for that global appear in the function IR?  I believe I saw an instance where the ir.Global, even though it was only used inside the prange, was outside the prange in the IR.  If it is inside then no problem.  If it is outside then you need this PR to move that ir.Global inside so that the lhs of the ir.Global is not passed as a parameter to the parfor gufunc.

> > We can remove the test case and commit the rest or just keep this PR in reserve in case this issue comes up again (of course by that point we're likely to forget this PR exists).
> 
> I'm reluctant to commit code that does not have a test. It may be that we just close this and rely on being able to search for the same error message in the issue tracker which will lead us here.

Maybe you or Siu can think of a case to force the bad situation that I described above.  I think that future issues filed on this bug will not be obvious from what the issue will choose to write.  Maybe we will notice there is a global involved and think back to this PR but we might forget as well.  I'm much rather come up with an example to show this problem than close the PR.

> > One thing to note is that with the original example it will not work because the expand_dims line has a broadcast in it that is not supported by parallel=True.
> 
> Seems like there's a few things like this that just lead to incorrect results. Can these be detected and compilation either aborted or maybe the region where the unsupported thing is serialised?

Currently, attempted usages of broadcasting should be caught and asserts happen instead.  This is obviously not desirable.  You could as you say try to fall back to the non-parallel version of the code for the broadcasted operation but this is very complicated.  If you have a bunch of operations in a row and the first one has a broadcasting check, then to do what you want, the assert becomes a conditional where in the broadcasting case, the first operation is done non-parallel but the remaining operations stay parallel (presumably because array analysis knows the sizes are fine and thus didn't insert a size assert) and in the non-broadcasting case they can all stay parfors.  Thus, each assert would seem to require code duplication of the remaining code in the basic block (I think).  It isn't that hard to imagine the solution but probably quite difficult to get the code right.

Another possibility is to directly supporting broadcasting inside the parfor so that even the broadcasted operation could be parallelized but I think this still requires the code duplication as you can't statically know to generate the broadcasted parfor or the standard parfor.

@stuartarchibald

stuartarchibald:
@DrTodd13 Thanks for the above, given there's no immediately obvious reproducer that is also using supported functionality, shall we push this issue/PR out of the 0.55 milestone?

RE broadcasting problem. My concern at present is that broadcasting like in the original reproducer is being compiled as though it were supported but at runtime it produces invalid output. We should probably discuss further how to handle broadcasting more generally at a Numba public meeting, but my preference right now is to prevent illegal/unsupported code from being compiled.

Also, as an aside, I'm guessing a solution that uses code duplication at each broadcast point could lead to very large IR and the usual associated problems with this?

Thanks for your input.

DrTodd13:
> @DrTodd13 Thanks for the above, given there's no immediately obvious reproducer that is also using supported functionality, shall we push this issue/PR out of the 0.55 milestone?

Sure.

> RE broadcasting problem. My concern at present is that broadcasting like in the original reproducer is being compiled as though it were supported but at runtime it produces invalid output. We should probably discuss further how to handle broadcasting more generally at a Numba public meeting, but my preference right now is to prevent illegal/unsupported code from being compiled.

I think we may be seeing different outcomes here.  For my test code, what I see is the broadcasting assertion firing.  It sounds like you are seeing it going past that and getting an incorrect result.  The thing is that much broadcasting can only be detected at runtime so you don't know if it is illegal or not while you're compiling.

> Also, as an aside, I'm guessing a solution that uses code duplication at each broadcast point could lead to very large IR and the usual associated problems with this?

Yes, that is a problem with this approach.  Another less invasive approach is just to have a conditional for individual operations that don't multiply downstream.  So, you just check if an operation is a broadcast and if so do it the normal way, else do it the parallel way.  The downside of this approach is that you can't then fuse the non-broadcasted operation with others downstream.  So, this could actually slow things down in existing code without broadcasting.  If we did it that way by default then the current way could be an option to turn on and there's yet another option to not even insert the assertions which themselves prevent fusion.

PR #7247, opened 2021-07-25
Solve memory leak to fix issue #7210 

### LINKED ISSUE #7210: Memory leak bugs in module init functions with PyObject_SetAttrString
## Reporting a bug

**(a static analyzer bug report)**

In file `numba/np/ufunc/omppool.cpp`, `numba/np/ufunc/tbbpool.cpp`, `numba/np/ufunc/_num_threads.c`, and `numba/np/ufunc/workqueue.c`, function `PyObject_SetAttrString` is used to add functions to the module being imported in the way shown below.

```c
PyObject_SetAttrString(m, "function_name", PyLong_FromVoidPtr(&function));
```

As the function `PyLong_FromVoidPtr` returns a new reference, it will lead to a memory leak for the `long` objects returned by `PyLong_FromVoidPtr`.

ysheffer:
Added a macro to _pymodule.h and set attributes in _num_threads.c, _omppool.cpp,
tbbpool.cpp, workqueue.c using a temporary variable to solve memory leak and fix #7210 .

### DIFF
--- a/numba/_pymodule.h
+++ b/numba/_pymodule.h
@@ -24,7 +24,6 @@
 #define PyInt_Type PyLong_Type
 #define PyInt_Check PyLong_Check
 #define PyInt_CheckExact PyLong_CheckExact
-// Define object attribute using a temporary variable to avoid memory leak
 #define SetAttrStringFromVoidPointer(m, name, tmp) \
             tmp = PyLong_FromVoidPtr((void *) &name);\
             PyObject_SetAttrString(m, #name, tmp);\
--- a/numba/_pymodule.h
+++ b/numba/_pymodule.h
@@ -24,9 +24,9 @@
 #define PyInt_Type PyLong_Type
 #define PyInt_Check PyLong_Check
 #define PyInt_CheckExact PyLong_CheckExact
-#define SetAttrStringFromVoidPointer(m, name, tmp) \
-            tmp = PyLong_FromVoidPtr((void *) &name);\
-            PyObject_SetAttrString(m, #name, tmp);\
-            Py_DECREF(tmp);
+#define SetAttrStringFromVoidPointer(m, name) do { \
+        PyObject *tmp = PyLong_FromVoidPtr((void *) &name); \
+        PyObject_SetAttrString(m, #name, tmp); \
+        Py_DECREF(tmp); } while (0)
 
 #endif /* NUMBA_PY_MODULE_H_ */
--- a/numba/_pymodule.h
+++ b/numba/_pymodule.h
@@ -27,7 +27,7 @@
 // Define object attribute using a temporary variable to avoid memory leak
 #define SetAttrStringFromVoidPointer(m, name, tmp) \
             tmp = PyLong_FromVoidPtr((void *) &name);\
-            PyObject_SetAttrString(m, "name", tmp);\
-            Py_DECREF(tmp);\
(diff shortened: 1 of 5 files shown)


esc:
@ysheffer thank you for submitting this to the Numba issue tracker. It seems like your PR is doing a lot of cosmetics to the code in addition to implementing the feature. Do you think you could submit this PR with only the feature and not the cosmetics, as this makes it harder to review. Thanks!

ysheffer:
@esc Thank you for your comment. Sure thing, I have updated the PR accordinaglly.

esc:
@ysheffer thank you for making those changes, I have added it to the queue for review.

numba/_pymodule.h
```diff
@@ -24,5 +24,10 @@
 #define PyInt_Type PyLong_Type
 #define PyInt_Check PyLong_Check
 #define PyInt_CheckExact PyLong_CheckExact
+// Define object attribute using a temporary variable to avoid memory leak
+#define SetAttrStringFromVoidPointer(m, name, tmp) \
```
Snape3058, line 28: I think you had better wrap these three lines with a `do { } while (false)` block and declare a temporary variable in the block instead of using an external variable.

numba/_pymodule.h
```diff
@@ -24,5 +24,10 @@
 #define PyInt_Type PyLong_Type
 #define PyInt_Check PyLong_Check
 #define PyInt_CheckExact PyLong_CheckExact
+// Define object attribute using a temporary variable to avoid memory leak
```
stuartarchibald, line 27:
```suggestion
```
Probably don't need this.

stuartarchibald (commented):
@ysheffer Many thanks for the patch, couple of minor things to look at else looks good.

numba/_pymodule.h
gmarkall: This is now applied, but I used `while (0)` instead of `while (false)` because `false` is not defined everywhere the macro is used.

numba/np/ufunc/omppool.cpp
```diff
@@ -239,31 +239,24 @@ static void ready(void)
...
-                           PyLong_FromVoidPtr((void*)&do_scheduling_unsigned));
-    PyObject_SetAttrString(m, "openmp_vendor",
-                           PyString_FromString(_OMP_VENDOR));
-    PyObject_SetAttrString(m, "set_num_threads",
-                           PyLong_FromVoidPtr((void*)&set_num_threads));
-    PyObject_SetAttrString(m, "get_num_threads",
-                           PyLong_FromVoidPtr((void*)&get_num_threads));
-    PyObject_SetAttrString(m, "get_thread_id",
-                           PyLong_FromVoidPtr((void*)&get_thread_id));
+    SetAttrStringFromVoidPointer(m,launch_threads,tmp)
```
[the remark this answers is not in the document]
gmarkall: Comma-consistency has now been achieved.

gmarkall:
@stuartarchibald This should be ready for another look now.

stuartarchibald (approved):
@ysheffer Many thanks for the patch, @gmarkall many thanks for responding to comments and providing fixes against the original patch. Good to see this leak patched.

stuartarchibald:
@ysheffer Congratulations on your first contribution to Numba!

PR #7251, opened 2021-07-27
Fix #6001: typed.List ignores ctor arguments with JIT disabled

### LINKED ISSUE #6001: `typed.List` ignores constructor arguments when jitting is disabled
## Reporting a bug

```python
>>> from numba.typed import List
>>> List([1])
ListType[int64]([1])

>>> from numba import config
>>> config.DISABLE_JIT = True
>>> List([1])
[]  # where did the 1 go?
```

gmarkall:
When `List.__new__()` is called, the constructor arguments get swallowed by `lsttype`, which is the first argument (after the class) and is also a keyword argument. This doesn't matter when JIT is enabled, because `__new__()` returns an instance of `typed.List`, so the interpreter then calls `__init__()` on the result.

However, when JIT is disabled, `List.__new__()` returns an instance of `list`, and therefore the interpreter does not call `__init__()` on it at all - it is the responsibility of `__new__()` to initialize the returned object.

(See https://docs.python.org/3/reference/datamodel.html#object.__new__ for an explanation of the behaviour and requirements described).

To implement correct behaviour with JIT disabled, we need two changes:

- The non-keyword arguments need to be placed ahead of keyword arguments in the parameters to `List.__new__()` (`*args`).
- When JIT is disabled, in addition to creating a new `list`, we also explicitly call its `__init__()` method with our `args` and `kwargs`.

Fixes #6001.

### DIFF
--- a/numba/typed/typedlist.py
+++ b/numba/typed/typedlist.py
@@ -200,12 +200,15 @@ class List(MutableSequence, pt.Generic[T]):
     _legal_kwargs = ["lsttype", "meminfo", "allocated"]
 
     def __new__(cls,
+                *args,
                 lsttype=None,
                 meminfo=None,
                 allocated=DEFAULT_ALLOCATED,
                 **kwargs):
         if config.DISABLE_JIT:
-            return list.__new__(list)
+            pylist = list.__new__(list)
+            pylist.__init__(*args, **kwargs)
+            return pylist
         else:
             return object.__new__(cls)
 
--- a/numba/tests/test_listobject.py
+++ b/numba/tests/test_listobject.py
@@ -20,7 +20,7 @@ from numba.core import types
 from numba.core.errors import TypingError
 from numba.tests.support import (TestCase, MemoryLeakMixin, override_config,
                                  forbid_codegen)
-from numba.typed import listobject
+from numba.typed import listobject, List
 
 
 class TestCreateAppendLength(MemoryLeakMixin, TestCase):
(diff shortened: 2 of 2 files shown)


esc:
@gmarkall thanks for the patch!

esc (approved):
Excellent fix, very insightful!

numba/typed/typedlist.py
```diff
@@ -200,12 +200,15 @@ class List(MutableSequence, pt.Generic[T]):
...
+                *args,
                 lsttype=None,
                 meminfo=None,
                 allocated=DEFAULT_ALLOCATED,
                 **kwargs):
         if config.DISABLE_JIT:
-            return list.__new__(list)
+            pylist = list.__new__(list)
+            pylist.__init__(*args, **kwargs)
+            return pylist
```
eric-wieser, line 211: Wouldn't just `return list(*args, **kwargs)` be less fragile here?

gmarkall: Possibly? I felt that the idiom I used was a bit more natural in the context of being in a `__new__()` method... Is there something that could go wrong with this that `list(*args, **kwargs)` would prevent?

eric-wieser: Yes, theres definitely something that could go wrong. Normally when you call `list(x)`, python calls `l = list.__new__(list, x)` and then `l.__init__(x)`. You're not passing `x` into `__new__`, and it's unlikely that Python specifies what list is supposed to do if you give init and new different arguments.

It's much safer just to call the list constructor - there is no benefit to splitting up the call here, and it makes you vulnerable to implementation details in python itself.

gmarkall: I see - I'd made the assumption that `list.__new__()` did nothing with its arguments, but I do see that this is an implementation detail.

gmarkall: PR #7257 opened to make this change.

Issue #7110, opened 2021-06-13
numpy.delete optional argument 'axis' support


Hanni-ui:
## Feature request

Hi,
I'm working at the moment with numba and numpy, and I wondered, why numba doesn't seems to fully support numpy.delete. If I use the third argument "axis" (described in [Numpy Delete Docu](https://numpy.org/doc/stable/reference/generated/numpy.delete.html)), numba tells me that it doesn't know this argument. Without numba it is working perfectly. (It is very interesting, that numba does support 'axis' from numpy.append, but doesn't support numpy.insert at all.) The problem is, that without this argument, arrays are automatically flattened when deleting a part of it. I appended an code example and a (not very nice) way to go around that problem. Thanks for your help!

`A = np.zeros((2,5))`
`print("original array",A)`
`B = np.delete(A,1)`
`print("without axis",B)`
`C = np.delete(A,1,0)   `
`print("with axis",C)`
`D = np.append(A[:1],A[2:],axis=0)  `
`print("alternative way",D)`

output:

`original array [[0. 0. 0. 0. 0.]
 [0. 0. 0. 0. 0.]]`
`without axis [0. 0. 0. 0. 0. 0. 0. 0. 0.]`
`with axis [[0. 0. 0. 0. 0.]]`
`alternative way [[0. 0. 0. 0. 0.]]`

esc:
@Hanni-ui thank you for asking about this. As stated in: 

https://numba.readthedocs.io/en/stable/reference/numpysupported.html#other-functions

currently `axis` is not supported for `numpy.delete`. Thus I am marking this as a feature request.

esc:
xref: https://github.com/numba/numba/issues/1269

guoqiangqi:
Does anyone have any plan for this issue? If not, i would like to do.

PR #7252, opened 2021-07-27
[WIP] Add numpy.delete optional argument axis support

guoqiangqi:

Add numpy.delete optional argument axis support, related to #7110

### DIFF
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -28,7 +28,7 @@ from numba.np.linalg import ensure_blas
 from numba.core.extending import intrinsic
 from numba.core.errors import RequireLiteralValue, TypingError
 from numba.core.overload_glue import glue_lowering
-from numba.cpython.unsafe.tuple import tuple_setitem
+from numba.cpython.unsafe.tuple import tuple_setitem, build_full_slice_tuple
 
 
 def _check_blas():
@@ -3480,7 +3480,7 @@ np_delete_handler_isarray = register_jitable(lambda x : np.asarray(x))
 
 
 @overload(np.delete)
-def np_delete(arr, obj):
+def np_delete(arr, obj, axis=None):
     # Implementation based on numpy
     # https://github.com/numpy/numpy/blob/af66e487a57bfd4850f4306e3b85d1dac3c70412/numpy/lib/function_base.py#L4065-L4267    # noqa: E501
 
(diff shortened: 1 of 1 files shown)


gmarkall:
@guoqiangqi Is this now ready for review?

guoqiangqi:
> @guoqiangqi Is this now ready for review?

@gmarkall  Not yet, the current implementation has bad performance in some situations, I am still trying to further optimize them.

guoqiangqi:
@gmarkall When all jobs done, i will let you know.

guoqiangqi:
Hi @gmarkall, this patch is ready for reveiew now.

gmarkall:
Thanks, I've now added it to be reviewed!

numba/np/arraymath.py
```diff
@@ -3480,49 +3480,109 @@ def impl(arr, axis=None):
...
 
-        def np_delete_impl(arr, obj):
-            arr = np.ravel(np.asarray(arr))
-            N = arr.size
+                keep = np.ones(N, dtype=np.bool_)
+                obj = handler(obj)
+                keep[obj] = False
+                return arr[keep]
+        else:
+            ndim = arr.ndim
```
kc611, line 3513: This will fail when `arr` is a `types.Sequence`.

```python
import numpy as np
import numba

@numba.njit
def something(x, obj):
    return np.delete(x, obj, axis=1)

x = np.random.default_rng(1).random((3,3))

print(something.py_func(((1,2), (3,4)), slice(1,2))) # [[1][3]]
print(something(((1,2), (3,4)), slice(1,2))) # Fails
```
If we want to keep support for `types.Sequence`, the logic for following case needs to be updated such that doesn't require `ndim` to be calculated here.

numba/np/arraymath.py
```diff
@@ -3480,49 +3480,109 @@ def impl(arr, axis=None):
...
+                N = arr.shape[axis]
+                if (obj < -N or obj >= N):
+                    raise IndexError('obj must be less than arr.shape[axis]')
+                if (obj < 0):
+                    obj += N
+
+                new_shape = tuple_setitem(arr.shape, axis,arr.shape[axis] - 1)
+                new = np.empty(new_shape, arr.dtype)
+
+                slobj = build_full_slice_tuple(arr.ndim)
```
kc611, line 3575: I think this cannot have runtime integers as arguments. (So it will probably fail, since `arr.ndim` is an integer that is considered as not fixed during runtime).

numba/tests/test_np_functions.py
```diff
@@ -771,20 +775,144 @@ def test_delete_exceptions(self):
...
+        )
+
         with self.assertRaises(TypingError) as raises:
             cfunc(2, 3)
         self.assertIn(
             'arr must be either an Array or a Sequence',
             str(raises.exception)
         )
 
+    def test_delete3_basic(self):
```
kc611, line 792: Maybe add those above stated failures in the review as test cases when they're fixed.

kc611 (changes requested):
Hey @guoqiangqi , thank you for the PR, your contributions are highly appreciated :smile:  . Here is some feedback regarding your changes.

sklam:
We have to move this to the next release since it still need some work and too close to our release cut date.

Issue #7104, opened 2021-06-11
CUDA on WSL2 Support


Sahuta:
---

## Feature request
I just tried to work with CUDA on WSL, with Numba on anaconda 3.
Many part of CUDA features works well, such as nvcc, nvidia-smi,  and python libraries such as Cupy, other than Numba CUDA.

Numba works well while it runs simple @jit compiling.  I add the log when I type `numba -s` bellow.

```
__Hardware Information__
Machine                                       : x86_64
CPU Name                                      : skylake
CPU Count                                     : 8
Number of accessible CPUs                     : 8
List of accessible CPUs cores                 : 0 1 2 3 4 5 6 7
CFS Restrictions (CPUs worth of runtime)      : None

CPU Features                                  : 64bit adx aes avx avx2 bmi bmi2
                                                clflushopt cmov cx16 cx8 f16c fma
                                                fsgsbase fxsr invpcid lzcnt mmx
                                                movbe pclmul popcnt prfchw rdrnd
                                                rdseed sahf sse sse2 sse3 sse4.1
                                                sse4.2 ssse3 xsave xsavec xsaveopt
                                                xsaves

Memory Total (MB)                             : 24048
Memory Available (MB)                         : 19547

__OS Information__
Platform Name                                 : Linux-5.10.16.3-microsoft-standard-WSL2-x86_64-with-glibc2.10
Platform Release                              : 5.10.16.3-microsoft-standard-WSL2
OS Name                                       : Linux
OS Version                                    : #1 SMP Fri Apr 2 22:23:49 UTC 2021
OS Specific Version                           : ?
Libc Version                                  : glibc 2.31

__Python Information__
Python Compiler                               : GCC 7.3.0
Python Implementation                         : CPython
Python Version                                : 3.8.8
Python Locale                                 : ja_JP.UTF-8

__LLVM Information__
LLVM Version                                  : 10.0.1

__CUDA Information__
CUDA Device Initialized                       : False
CUDA Driver Version                           : ?
CUDA Detect Output:
None
CUDA Libraries Test Output:
None

__ROC information__
ROC Available                                 : False
ROC Toolchains                                : None
HSA Agents Count                              : 0
HSA Agents:
None
HSA Discrete GPUs Count                       : 0
HSA Discrete GPUs                             : None

__SVML Information__
SVML State, config.USING_SVML                 : False
SVML Library Loaded                           : False
llvmlite Using SVML Patched LLVM              : True
SVML Operational                              : False

[...]

PR #7255, opened 2021-07-27
CUDA: Support CUDA Toolkit conda packages from NVIDIA

gmarkall:
NVIDIA now publishes conda packages containing the CUDA toolkit: https://anaconda.org/nvidia

These packages place components in different locations to the Anaconda- and conda-forge-maintained packages. This PR updates Numba's library search logic so that it can find the libraries in these packages. The locations of the components in these packages are (all relative to `$CONDA_PREFIX`):

- NVVM is placed in `nvvm/lib64` on Linux and `nvvm/bin` on Windows
- Static CUDA libraries are in `lib` on Linux, and `Lib/x64` on Windows
- Dynamic CUDA libraries are in `lib` on Linux, and `bin` on Windows
- Libdevice is in `nvvm/libdevice`

I also noticed that locating libcudadevrt was broken when using a non-conda-installed toolkit on Windows - this is also resolved by this PR.

Along with this, I thought it would also be helpful to display the locations searched for libcuda.so (as the location of libcuda.so can be an issue, as in #7104) - this PR shows the search locations and the path used for a successful load, but unfortunately this is sometimes relative, with the actual path determined by the system loader. There is no easy way to check the absolute path of the loaded library, but something more complex to report the exact path could be added in a future PR. A little refactoring was needed in `cudadrv.py` to separate the path determination from the actual loading, so that the path info could be presented by `libs.test()`.

The logic additions in `cuda_paths.py` are a little unwieldy and contrived - unfortunately as the file has evolved things have got a little out of control - it's difficult to get this logic both correct and minimal, so I've left it as it is for now rather than trying to further refactor the code here - given that this code does not change much, I'm inclined not to spend too much more time thinking about it.


gmarkall:
@esc Could this have a buildfarm run prior to review please? I'm concerned that something might be up (because this has been hard to get right) and I'd rather make sure it works before the review, rather than getting it approved then discovering a fatal flaw in the logic.

numba/cuda/cuda_paths.py
```diff
@@ -118,6 +144,55 @@ def get_conda_ctk():
     return os.path.dirname(max(paths))
 
 
+def get_nvidia_nvvm_ctk():
```
stuartarchibald, line 147: Is there benefit in memozing all these `get` functions with `functools.cache`?

numba/cuda/cudadrv/driver.py
```diff
@@ -150,6 +155,12 @@ def find_driver():
         _raise_driver_error(errmsg)
 
 
+def find_driver():
```
stuartarchibald, line 158: Same question as above RE `functools.cache`.

numba/cuda/cuda_paths.py
```diff
@@ -77,21 +79,45 @@ def _cudalib_path():
...
 
+def _get_static_cudalib_dir_path_decision():
+    options = [
+        ('Conda environment', get_conda_ctk()),
+        ('Conda environment (NVIDIA package)', get_nvidia_static_cudalib_ctk()),
+        ('CUDA_HOME', get_cuda_home(*_static_cudalib_path())),
+        ('System', get_system_ctk(_cudalib_path())),
+    ]
+    by, libdir = _find_valid_path(options)
+    return by, libdir
```
stuartarchibald, line 108: Temporaries needed? Could this just return `_find_valid_path(options)` ?

numba/cuda/cudadrv/driver.py
```diff
@@ -150,6 +155,12 @@ def find_driver():
...
+    dlloader, candidates = locate_driver_and_loader()
+    dll, path = load_driver(dlloader, candidates)
```
stuartarchibald, line 160:
```suggestion
    dll, _ = load_driver(dlloader, candidates)
```
path is unused?

gmarkall:
Moving from the 0.57 milestone pending future developments in the structure and distribution of the CUDA toolkit packages - once a clear route forward is visible this PR can be updated and moved into an appropriate milestone.

numba/cuda/cuda_paths.py
```diff
@@ -77,21 +79,45 @@ def _cudalib_path():
         return 'lib64'
 
 
+def _static_cudalib_path():
+    if IS_WIN32:
+        return ('lib', 'x64')
+    else:
+        return ('lib64',)
```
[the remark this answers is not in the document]
gmarkall: That comment in the description refers to the location of libraries in the Conda packages, but this function is for locating the libraries within user-installed CUDA toolkits specified using the `CUDA_HOME` environment variable. I'll make some change to this function to make it clearer what it's used for.

numba/cuda/cuda_paths.py
gmarkall: As per above, it seems to make no difference.

numba/cuda/cudadrv/driver.py
gmarkall: No difference as above.

numba/cuda/cudadrv/driver.py
gmarkall: It is, but I named it so that you don't have to go and look at `load_driver` to figure out what's being discarded and why.

numba/cuda/cuda_paths.py
gmarkall: It could, but then what's being returned is not evident from this function's code alone. I'd prefer to leave them in because it's already hard enough to follow all the path detection and library-finding logic, without making what's going from `_find_valid_path()` out the return value of this function hidden.

numba/cuda/cuda_paths.py
```diff
@@ -118,6 +144,55 @@ def get_conda_ctk():
     return os.path.dirname(max(paths))
 
 
+def get_nvidia_nvvm_ctk():
+    """Return path to directory containing the NVVM shared library.
+    """
+    is_conda_env = os.path.exists(os.path.join(sys.prefix, 'conda-meta'))
+    if not is_conda_env:
+        return
+    # Asssume the existence of NVVM to imply cudatoolkit installed
```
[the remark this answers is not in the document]
gmarkall: Now fixed.

numba/cuda/cuda_paths.py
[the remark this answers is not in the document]
gmarkall: Now renamed to `_cuda_home_static_cudalib_path()`.

gmarkall:
@stuartarchibald Many thanks for the review. The way forward for publishing CUDA toolkit packages on Anaconda.org is now resolved, and CUDA 12.0 packages are available on the NVIDIA channel. I have updated this PR with `main` and in response to the comments above, so it should be ready for another round of review.

A couple of points on testing - first, it would be good to test with gpuCI, but the driver version in our gpuCI setup doesn't have CUDA 12.0 yet, so that will have to be added in future. Secondly, I've tested this locally with the NVIDIA CUDA 12.0 packages, and tests pass, and the library tests show correct detection:

```
$ python -c "from numba import cuda; cuda.cudadrv.libs.test()"
Finding driver from candidates: libcuda.so, libcuda.so.1, /usr/lib/libcuda.so, /usr/lib/libcuda.so.1, /usr/lib64/libcuda.so, /usr/lib64/libcuda.so.1...
Using loader <class 'ctypes.CDLL'>
	trying to load driver...	ok, loaded from libcuda.so
Finding nvvm from Conda environment (NVIDIA package)
	located at /home/gmarkall/mambaforge/envs/numba-nvidia-channel/nvvm/lib64/libnvvm.so.4.0.0
	trying to open library...	ok
Finding cudart from Conda environment (NVIDIA package)
	located at /home/gmarkall/mambaforge/envs/numba-nvidia-channel/lib/libcudart.so.12.0.146
	trying to open library...	ok
Finding cudadevrt from Conda environment (NVIDIA package)
	located at /home/gmarkall/mambaforge/envs/numba-nvidia-channel/lib/libcudadevrt.a
Finding libdevice from Conda environment (NVIDIA package)
	trying to open library...	ok
```

If you want to test locally with the NVIDIA packages, you can install it with:

```
conda install nvidia::cuda-toolkit=12
```

Let me know if you run into any issues in testing.

gmarkall:
gpuci run tests

numba/cuda/cuda_paths.py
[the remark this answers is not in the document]
stuartarchibald: Thanks.

numba/cuda/cuda_paths.py
```diff
@@ -77,21 +79,45 @@ def _cudalib_path():
...
     by, libdir = _find_valid_path(options)
     return by, libdir
 
 
+def _get_static_cudalib_dir_path_decision():
+    options = [
+        ('Conda environment', get_conda_ctk()),
+        ('Conda environment (NVIDIA package)', get_nvidia_static_cudalib_ctk()),
+        ('CUDA_HOME', get_cuda_home(*_static_cudalib_path())),
+        ('System', get_system_ctk(_cudalib_path())),
```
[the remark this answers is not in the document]
stuartarchibald: Thanks for checking. It also seems like `get_cuda_paths` caches all this locally (might want to update that to use `functools.cache` at some point).

numba/cuda/cudadrv/driver.py
stuartarchibald: Ah ok, thanks for the explanation, am happy to leave as-is.

numba/cuda/cuda_paths.py
stuartarchibald: Thanks for the explanation, am happy to leave as-is.

stuartarchibald:
Thanks for the updates in 09d6fc1, they look good. I think this patch just needs a manual test and a run through the buildfarm.

gmarkall:
@stuartarchibald As discussed OOB I've added the supported CCs to `nvvm.py` for toolkits 12.0 and 12.1 - these are changes related to this PR since the packages it adds support for are from 12.0 onwards. I also made a note in the docs that MVC is not supported on CUDA 12 (when I made the MVC PR CUDA 12 was not out yet so it stated nothing about it at the time).

docs/source/cuda/minor_version_compatibility.rst
```diff
@@ -17,6 +17,8 @@ Numba supports MVC for CUDA 11 on Linux using the external ``cubinlinker`` and
 - Cooperative Groups are unsupported, because they require an archive to be
   linked.
 
+MVC is not yet supported on CUDA 12 - this will be added in a future release of Numba.
```
stuartarchibald, line 20:
```suggestion
MVC is not yet supported on CUDA 12 - this will be added in a future release of
Numba.
```
line wrap to 80 chars?

stuartarchibald:
> @stuartarchibald As discussed OOB I've added the supported CCs to `nvvm.py` for toolkits 12.0 and 12.1 - these are changes related to this PR since the packages it adds support for are from 12.0 onwards. I also made a note in the docs that MVC is not supported on CUDA 12 (when I made the MVC PR CUDA 12 was not out yet so it stated nothing about it at the time).

Many thanks @gmarkall. I've tested this PR at bd92dd2 manually using CUDA Toolkit 12.1 conda packages from NVIDIA, all the CUDA unit tests pass and `numba -s` correctly reports the use of NVIDIA packages.

docs/source/cuda/minor_version_compatibility.rst
gmarkall: Doh!

gmarkall:
gpuci run tests

gmarkall:
@stuartarchibald Many thanks, line-wrap change committed.

stuartarchibald (approved):

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_185`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_185`.

Passed.

PR #7256, opened 2021-07-27
Fix link to the discourse forum in README

kmaehashi:
Just a nitpick to fix the dead link.


gmarkall:
Many thanks for the fix!

sklam (approved):

PR #7257, opened 2021-07-28
Use normal list constructor in List.__new__()

gmarkall:
Calling `list.__new__()` without arguments and then `list.__init__()` with arguments on the resulting object is not guaranteed to work - the fact that `list.__new__()` doesn't refer to its arguments (and therefore doesn't strictly require them to be the same as the arguments to `__init__()`) is an implementation detail.

It is safer to just call `list(*args, **kwargs)` instead.

With thanks to @eric-wieser (ref: https://github.com/numba/numba/pull/7251#discussion_r677856531)


eric-wieser (approved):

esc (approved):
thank you for the patch @gmarkall -- thank you for the review @eric-wieser !

PR #7260, opened 2021-07-30
Support typed lists in `heapq`

### LINKED ISSUE #4926: Numba heapq doesn't work with numba typed lists

gmarkall:
The `heapq` implementation checks for reflected lists, which bans its use with typed lists. This commit modifies the check so that it also allows typed lists. There are no further functional changes required for correct operation with typed lists - the `heapq` tests are updated to run all tests with reflected lists and with typed lists. The modifications to individual tests all convert to/from the chosen list class (either CPython lists for reflected lists, or typed Lists). In some cases typed lists need to be converted back to Python lists for functions like `assertPreciseEqual` to compare them with Python lists used in the reference (CPython) implementations of the functions.

I have checked that the tests with typed lists never use a reflected list by running the test class individually - no warnings about the use of deprecated reflected lists are emitted when running it:

```
$ python -m numba.runtests numba.tests.test_heapq.TestHeapqTypedList -m
Parallel: 21. Serial: 0
.....................
Ran 21 tests in 16.737s

OK
```

cc @truenicoco

Fixes #4926.

### DIFF
--- a/numba/cpython/heapq.py
+++ b/numba/cpython/heapq.py
@@ -104,7 +104,7 @@ def _heapreplace_max(heap, item):
 
 
 def assert_heap_type(heap):
-    if not isinstance(heap, types.List):
+    if not isinstance(heap, (types.List, types.ListType)):
         raise TypingError('heap argument must be a list')
 
     dt = heap.dtype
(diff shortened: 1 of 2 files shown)


gmarkall:
I think the errors are due to the default NumPy int size on Windows being `int32` as opposed to `int64` on Linux. I'll have to fix this.

gmarkall:
Typing errors now fixed, this should be ready for review.

stuartarchibald (approved):
Thanks for the patch, looks good!

PR #7262, opened 2021-07-30
fix order of handling and casting

### LINKED ISSUE #7250: Typed List indices are cast unsafely
## Reporting a bug

- [x ] I have tried using the latest released version of Numba (most recent is

```
import numba as nb
import numpy as np
@nb.njit
def f():
    l = nb.typed.List.empty_list(nb.uint8)
    for i in range(129):
        l.append(nb.uint8(i))
    a = l[-1]
    return l[a],l[nb.int8(a)],l[nb.uint16(a)]

f()
```
Returns (1,1,128) while I expected (128,1,128). 
Seems unsigned integers above 1/2 of the range are not good; so I'm guessing they are cast to the signed version with the same number of bits.

PS: what really threw me off is that this is not the behavior of numpy arrays
```
import numba as nb
@nb.njit
def f():
    N=129
    l = np.empty(N,dtype=np.uint8)
    for i in range(N):
        l[i]=nb.uint8(i)
    a = l[-1]
    return l[a],l[nb.int8(a)],l[nb.uint16(a)]

f()
```
Prints `(128,1,128)`.

esc:
Must cast first and handle later, not the other way around!

Fixes #7250

### DIFF
--- a/numba/tests/test_listobject.py
+++ b/numba/tests/test_listobject.py
@@ -223,12 +223,12 @@ class TestGetitem(MemoryLeakMixin, TestCase):
                   ):
             self.assertEqual(foo((t(0))), 0)
 
-    def test_list_getitem_different_sized_int_index(self):
+    def test_list_getitem_different_sized_uint_index(self):
         # Checks that the index type cast and ext/trunc to the
         # type of the length is correct, both wraparound and
         # direct index is tested via -1/0.
 
-        for ty in types.integer_domain:
+        for ty in types.unsigned_domain:
             @njit
             def foo():
                 l = listobject.new_list(int32)
@@ -236,6 +236,20 @@ class TestGetitem(MemoryLeakMixin, TestCase):
                 return l[ty(0)]
 
             self.assertEqual(foo(), 7)
+
+    def test_list_getitem_different_sized_int_index(self):
+        # Checks that the index type cast and ext/trunc to the
+        # type of the length is correct, both wraparound and
+        # direct index is tested via -1/0.
+
+        for ty in types.signed_domain:
+            @njit
+            def foo():
+                l = listobject.new_list(int32)
+                l.append(7)
+                return l[ty(0)], l[ty(-1)]
+
+            self.assertEqual(foo(), (7, 7))
 
 
 class TestGetitemSlice(MemoryLeakMixin, TestCase):
--- a/numba/tests/test_listobject.py
+++ b/numba/tests/test_listobject.py
@@ -233,9 +233,9 @@ class TestGetitem(MemoryLeakMixin, TestCase):
             def foo():
                 l = listobject.new_list(int32)
                 l.append(7)
-                return l[ty(0)], l[ty(-1)]
+                return l[ty(0)]
 
-            self.assertEqual(foo(), (7, 7))
+            self.assertEqual(foo(), 7)
 
 
 class TestGetitemSlice(MemoryLeakMixin, TestCase):
--- a/numba/tests/test_typedlist.py
+++ b/numba/tests/test_typedlist.py
@@ -7,7 +7,7 @@ import numpy as np
 
 from numba import config
 from numba import njit
-from numba import int32, float32, prange
+from numba import int32, float32, prange, uint8
 from numba.core import types
 from numba import typeof
 from numba.typed import List, Dict
@@ -1439,6 +1439,22 @@ class TestImmutable(MemoryLeakMixin, TestCase):
                     "list is immutable",
                     str(raises.exception),
                 )
+
+
+class TestGetItemIndexType(MemoryLeakMixin, TestCase):
+
+    def test_indexing_with_uint8(self):
+        """ Test for reproducer at https://github.com/numba/numba/issues/7250
+        """
+        @njit
+        def foo():
+            l = List.empty_list(uint8)
+            for i in range(129):
+                l.append(uint8(i))
+            a = uint8(128)
+            return l[a]
+
+        self.assertEqual(foo(), 128)
 
 
 class TestListFromIter(MemoryLeakMixin, TestCase):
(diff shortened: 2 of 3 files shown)


esc:
Seems like this is a breaking change, (Python 3.7 on Windows):

```
ERROR: test_list_getitem_different_sized_int_index (numba.tests.test_listobject.TestGetitem)
  File "D:\a\1\s\numba\tests\test_listobject.py", line 238, in test_list_getitem_different_sized_int_index
    self.assertEqual(foo(), (7, 7))
  File "D:\a\1\s\numba\typed\listobject.py", line 662, in handle_index
    raise IndexError("list index out of range")
IndexError: list index out of range

Stderr:
D:\a\1\s\numba\tests\test_listobject.py:236: NumbaTypeSafetyWarning: unsafe cast from uint64 to int64. Precision may be lost.
  return l[ty(0)], l[ty(-1)]
```

esc:
One of the Linux Python 3.7 tests failed also:

```
ERROR: test_list_getitem_different_sized_int_index (numba.tests.test_listobject.TestGetitem)
  File "/home/vsts/work/1/s/numba/tests/test_listobject.py", line 238, in test_list_getitem_different_sized_int_index
    self.assertEqual(foo(), (7, 7))
  File "/home/vsts/work/1/s/numba/typed/listobject.py", line 662, in handle_index
    raise IndexError("list index out of range")
IndexError: list index out of range

FAIL: test_list_getitem_different_sized_int_index (numba.tests.test_listobject.TestGetitem)
  File "/home/vsts/work/1/s/numba/tests/support.py", line 779, in tearDown
    self.memory_leak_teardown()
  File "/home/vsts/work/1/s/numba/tests/support.py", line 753, in memory_leak_teardown
    self.assert_no_memory_leak()
  File "/home/vsts/work/1/s/numba/tests/support.py", line 762, in assert_no_memory_leak
    self.assertEqual(total_alloc, total_free)
AssertionError: 2 != 1
```

stuartarchibald:
@esc think this might need a merge from `main` to pick up the Azure config which has Python 3.7 removed.

esc:
> @esc think this might need a merge from `main` to pick up the Azure config which has Python 3.7 removed.

yeah, it was like 3k commits behind, merged in `main` and pushed, let's see what azure does.

apmasell:
Please simplify the reproducer in the issue and add it as a test case.

esc:
rebased onto `main` at 08a514b128132a53f99ad104a03a9ce331e97588

esc:
> Please simplify the reproducer in the issue and add it as a test case.

Has been added to this PR.

esc:
So, now it's not breaking on 3.9 but instead on 3.8 -- a little surprising. The error message is:

```
ERROR: test_list_getitem_different_sized_int_index (numba.tests.test_listobject.TestGetitem)
  File "/home/vsts/work/1/s/numba/tests/test_listobject.py", line 238, in test_list_getitem_different_sized_int_index
    self.assertEqual(foo(), (7, 7))
  File "/home/vsts/work/1/s/numba/typed/listobject.py", line 662, in handle_index
    raise IndexError("list index out of range")
IndexError: list index out of range

Stderr:
/home/vsts/work/1/s/numba/tests/test_listobject.py:235: NumbaTypeSafetyWarning: unsafe cast from int64 to int32. Precision may be lost.
  l.append(7)
```

esc:
@apmasell sending this one back to you. I added the required reproducer and it seems to work well. I made sure that it does fail w/o the change of order. But now a test unrelated to this is failing. It may be that eiter `main` is broken or that the change caused some collateral damage elsewhere. Also, as this PR was being developed, I saw failures on both 3.8 and 3.9 -- so there may be some non-detrminism involved here.

apmasell:
I suspect those are collateral damage given they are related to list indexing. `fix_index` will truncate or extend the provided index to the `ssize_t` index length. I guess that means that the larger indices are getting truncatred prematurely and that's causing the problems with those tests. I don't know why the type manipulation is done in two places. Maybe changing `fix_index` to always extend to the larger type will fix the problem?

esc:
I always forget, that the public Numba CI only tests slices of the tests. Hence, I can totally reproduce the above failure using 3.10 on OSX (arm64).

esc:
Here is the debug trace of the exception:

```
-> raise IndexError("list index out of range")
(Pdb) up
> /Users/vhaenel/git/numba/numba/tests/test_listobject.py(238)test_list_getitem_different_sized_int_index()
-> self.assertEqual(foo(), (7, 7))
(Pdb) ty
uint16
(Pdb) ty(-1)
65535
```

So, I guess, this is fine? `int64(-1)` as a `uint16` is the max value of an unsigned uint16 `2**16-1` -- two's complement just means all the bits are set.

Effectively the proposed fix exposes that this test had been working by accident.

@apmasell do you concur?  If so, perhaps we can simply remove the second indexing with `-1` test. I am not sure what benefit it brings.

esc:
> @apmasell do you concur? If so, perhaps we can simply remove the second indexing with `-1` test. I am not sure what benefit it brings.

I pushed a fix.

apmasell:
The -1 for signed integers are useful for the indexing-from-the-end, right? Maybe keep it and do the test only over `types.signed_domain`, and then create a second test as you have for `types.unsigned_domain`. Does that seem like a better test surface?

esc:
> The -1 for signed integers are useful for the indexing-from-the-end, right? Maybe keep it and do the test only over `types.signed_domain`, and then create a second test as you have for `types.unsigned_domain`. Does that seem like a better test surface?

That's great advice. Have pushed another commit to fixup.

esc:
@apmasell thank you for the review, I am marking this as RTM now.

esc:
looks like Azure is choking

esc:
unglitched Azure - now RTM

esc:
@apmasell thank you for the review, @sklam thank you for the merge!

PR #7265, opened 2021-08-03
Fix linspace to use np.divide and clamp to stop.

### LINKED ISSUE #7366: Unexpected behaviour of numpy.linspace within am njitted funciton
## Reporting a bug

- [x ] I have tried using the latest released version of Numba (most recent is
Numba version: 0.54.0

**Summary.**
np.linspace() within a njitted func is not behaving the same than np.linspace() alone. The 'stop' number is surpassed by +1e-6

**Explanation.**
Hi there!
I have a function that generates an array of 'num' values from 0 to 100. I use numpy linspace since 'num' is determined by the size of another array within the function, so for me, it's quite agile. Now, this linspace function is used thousands of times, and very few times, it generates a wrong linspace output array. I can't share the data because it comes from several databases and the error can only be replied in very few (specific) cases, but maybe this image might help:

image

When I use the function without the numba decorator, np.linspace is not giving this problem ever, works fine. When I use it with the njit, it only breaks very few times (like 15 time out of thousands), but I guess that's an unexpected behaviour. It might be related with the lenght of 'num', because it always breaks with the same input datasets

In the image above the length used for np.linspace is num=1298

In my particular case, I can solve this by just making the np.linspace(0,99.9999,num), because the code does not allow any number larger than 100. Yet, I guess that feedback might be helpful, at least for the record

Thanks a lot for your time! And for Numba!

stuartarchibald:
As title. This improves the match to NumPy and as a result the
tests can be more strict.

Fixes #7366

### DIFF
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -4074,11 +4074,13 @@ def numpy_linspace_3(context, builder, sig, args):
         div = num - 1
         if div > 0:
             delta = stop - start
-            step = delta / div
+            step = np.divide(delta, div)
             for i in range(0, num):
                 arr[i] = start + (i * step)
         else:
             arr[0] = start
+        if num > 1:
+            arr[-1] = stop
         return arr
 
     res = context.compile_internal(builder, linspace, sig, args)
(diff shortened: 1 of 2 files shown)


sklam (approved):
Thanks for the patch!

I've manually confirmed the patch resolves https://github.com/numba/numba/issues/7366 and the tests exercise it.

PR #5014, opened 2020-01-02
Enable use of records and bools for shared memory, remove ddt, add additional transpose tests

gmarkall:
This started out with a straightforward replacement of `ddt` with `unittest.subTest` - with the removal of support for Python 2, unittest subTests are available on all supported versions of Python. Since the use of `ddt` was quite minimal (only for the transpose test) I felt it was worth removing in favor of using `subTest` so that the vendored-in version of `ddt` can be dropped.

Whilst doing this I noticed that there weren't many tests for transpose, so I added a few more variants of sizes (around powers of 2, +/-1, where one or the other dimension is one of these, etc.) and noticed that record dtypes and bools were not permitted in shared arrays (as used by the transpose kernel) so I modified the implementation to allow them. This almost worked by simply allowing them through in `cudaimpl._generic_array` - however, NVVM requires shared memory variables to be power-of-2 aligned, which is true when using the sizes of all types in `types.number_domain` and booleans as their alignment, but not for records, which can be of arbitrary size. When the size is not a power of 2, it is rounded up to the next power of 2 to make it alignment to address this concern.

In summary, this PR:

* Adds support for record and bool dtypes for shared arrays and adds appropriate tests
* Adds additional variants of shapes and dtypes for the transpose tests
* Removes `ddt` and replaces it with the now built-in `subTest` function.

PR #7271, opened 2021-08-04
Update 3rd party license text.

stuartarchibald:
As title.


LICENSES.third-party
```diff
@@ -29,37 +29,12 @@ TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
 SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 
-ddt
```
stuartarchibald, line 32: Think this is removed. It used to be in `numba/testing/ddt.py` according to https://github.com/numba/numba/search?q=ddt

sklam: yes, done in https://github.com/numba/numba/pull/5014

sklam (approved):
Thanks for the patch!

PR #7272, opened 2021-08-04
Allow annotations in njit-ed functions

### LINKED ISSUE #7269: Ignore type annotations instead of crashing
---

## Feature request

As it is type annotations in internal functions error out compilation:
```python
import numba as nb
@nb.njit
def f(x):
    def g(x:int):
        return x
    return g(x)
f(1)
``` 
```
...
~/miniconda3/lib/python3.9/site-packages/numba/core/interpreter.py in op_MAKE_FUNCTION(self, inst, name, code, closure, annotations, kwdefaults, defaults, res)
   1685         if annotations is not None:
   1686             msg = "op_MAKE_FUNCTION with annotations is not implemented"
-> 1687             raise NotImplementedError(msg)
   1688         if kwdefaults is not None:
   1689             msg = "op_MAKE_FUNCTION with kwdefaults is not implemented"

NotImplementedError: Failed in nopython mode pipeline (step: analyzing bytecode)
op_MAKE_FUNCTION with annotations is not implemented
```
However annotations enable static analysis tools.
The situation would be improved if numba were to ignore them instead of raising an exception.
In my limited experimentation,commenting out the check seems to work.

LunarLanding:
fixes #7269 
This should not cause any issues, and allows better integration with static analysis tools.

### DIFF
--- a/numba/core/interpreter.py
+++ b/numba/core/interpreter.py
@@ -1715,9 +1715,8 @@ class Interpreter(object):
 
     def op_MAKE_FUNCTION(self, inst, name, code, closure, annotations,
                          kwdefaults, defaults, res):
-        if annotations is not None:
-            msg = "op_MAKE_FUNCTION with annotations is not implemented"
-            raise NotImplementedError(msg)
+        # annotations are ignored by numba but useful for static analysis
+        # re. https://github.com/numba/numba/issues/7269
         if kwdefaults is not None:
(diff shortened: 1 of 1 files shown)


sklam (approved):
Thanks for the patch!

stuartarchibald:
@LunarLanding Congratulations on your first contribution to Numba!

PR #7280, opened 2021-08-05
Fix lifetime of task_scheduler_handle

PokhodenkoSA:
This PR fixes tests in numba-dppy:
https://github.com/IntelPython/numba-dppy/pull/493

Related to TBB.

/CC @Hardcode84

### DIFF
--- a/numba/np/ufunc/tbbpool.cpp
+++ b/numba/np/ufunc/tbbpool.cpp
@@ -224,16 +224,14 @@ static void prepare_fork(void)
     {
         if(is_main_thread())
         {
-            if (tbb::finalize(tsh, std::nothrow))
+            if (!tbb::finalize(tsh, std::nothrow))
             {
-                tsh_was_initialized = false;
-                need_reinit_after_fork = true;
-            }
-            else
-            {
+                tbb::task_scheduler_handle::release(tsh);
                 puts("Unable to join threads to shut down before fork(). "
                      "This can break multithreading in child processes.\n");
             }
+            tsh_was_initialized = false;
+            need_reinit_after_fork = true;
         }
         else
         {
@@ -255,6 +253,7 @@ static void reset_after_fork(void)
     {
         tsh = tbb::task_scheduler_handle::get();
         set_main_thread();
+        tsh_was_initialized = true;
         need_reinit_after_fork = false;
     }
 }


stuartarchibald:
Thanks for the patch @PokhodenkoSA. I think this should be considered for 0.54 release, is there a test case that can be written to replicate the problem seen in `numba-dppy`?

PokhodenkoSA:
We get the error in complex case. I hope reproducer will be later.

Hardcode84:
I am not sure how to test this. We need to make `tbb::finalize` fail and to achieve this we need to take another `task_scheduler_handle` and keep it while doing fork.

stuartarchibald:
> I am not sure how to test this. We need to make `tbb::finalize` fail and to achieve this we need to take another `task_scheduler_handle` and keep it while doing fork.

Just to make sure I've understood... to reproduce, in the same process there needs to be Numba's TBB backend `task_scheduler_handle` and also another `task_scheduler_handle` from somewhere, and then both need to be live whilst `fork(2)` is called?

Hardcode84:
> Just to make sure I've understood... to reproduce, in the same process there needs to be Numba's TBB backend task_scheduler_handle and also another task_scheduler_handle from somewhere, and then both need to be live whilst fork(2) is called?

Yes

Hardcode84:
If there are multiple `task_scheduler_handle` instances in process `finalize` will fail (meaning TBB cannot be safely terminated yet) if there any other live `task_scheduler_handle` instance in process (but it will still free `task_scheduler_handle` passed to finalize, explicit `tbb::task_scheduler_handle::release(tsh);` here is just a precaution), And the last live handle will get the success from finalize.

stuartarchibald:
NOTE: reviewers, this branch is currently targeting the release0.54 branch and will need pointing at mainline before merge, then cherry-pick etc.

stuartarchibald:
@Hardcode84 does this somewhat scrappy example demonstrate the issue?
```python
import ctypes
import os
from tempfile import TemporaryDirectory, NamedTemporaryFile
from numba.pycc.platform import Toolchain

with TemporaryDirectory() as tmpdir:
    with NamedTemporaryFile(dir=tmpdir) as tmpfile:
        src = """
        #define TBB_PREVIEW_WAITING_FOR_WORKERS 1
        #include <tbb/tbb.h>
        static tbb::task_scheduler_handle tsh;
        extern "C"
        {
        void launch(void)
        {
            tsh = tbb::task_scheduler_handle::get();
        }
        }
        """
        cxxfile = f"{tmpfile.name}.cxx"
        with open(cxxfile, 'wt') as f:
            f.write(src)
        tc = Toolchain()
        object_files = tc.compile_objects([cxxfile,], output_dir=tmpdir)
        dso_name = f"{tmpfile.name}.so"
        tc.link_shared(dso_name, object_files, libraries=['tbb',],
                       export_symbols=['launch'])
        DLL = ctypes.CDLL(dso_name)

DLL.launch()

@njit(parallel=True)
def foo(n):
    acc = 0
    for i in prange(n):
        acc += i
    return acc

foo(1)
os.fork()

print(DLL)
```

Hardcode84:
> does this somewhat scrappy example demonstrate the issue?

It should.
Does it crashes with tbb 2021.3 without the the fix?

(We will check with pre-release 2021.4 later today)

stuartarchibald:
> > does this somewhat scrappy example demonstrate the issue?
> 
> It should.
> Does it crashes with tbb 2021.3 without the the fix?

Yes. It does this without the fix:
```
Unable to join threads to shut down before fork(). This can break multithreading in child processes.

<CDLL '/tmp/tmpzd8bvz79/tmp_mmsdcb7.so', handle 555675d79950 at 0x7f1c0dbf99d0>
<CDLL '/tmp/tmpzd8bvz79/tmp_mmsdcb7.so', handle 555675d79950 at 0x7f1c0dbf99d0>
Assertion handle failed (located in the finalize_impl function, line in file: 243)
Detailed description: trying to finalize with null handle
Assertion handle failed (located in the finalize_impl function, line in file: 243)
Detailed description: trying to finalize with null handle
Aborted
```
and this with the fix:
```
Unable to join threads to shut down before fork(). This can break multithreading in child processes.

<CDLL '/tmp/tmpytxji42l/tmputbyvtxx.so', handle 55af74c31950 at 0x7f734bb419d0>
<CDLL '/tmp/tmpytxji42l/tmputbyvtxx.so', handle 55af74c31950 at 0x7f734bb419d0>
```

is this what you'd expect?

> (We will check with pre-release 2021.4 later today)

Great, thanks.

stuartarchibald:
> > is this what you'd expect?
> 
> yes

Great, thanks for confirming @Hardcode84. From OOB conversation, I'll prepare a patch with the fix and unittest based on the above, will ping you and @PokhodenkoSA once it's ready for review.

stuartarchibald:
https://github.com/numba/numba/pull/7294 continues, it adds tests and once demonstrably failing will have this patch appended to prove it fixes the problem.

sklam:
Closed by https://github.com/numba/numba/pull/7294

PR #7289, opened 2021-08-06
Implement np.tensordot()

itamarst:
Implementing in the hopes I can use it as basis for `np.einsum`, but seems useful on its own too.

The code was copy/pasted from NumPy and updated; this relies on #7202 for the NumPy licensing compliance, but the appropriate changes could be done here instead.

My main concern is that this isn't well enough tested, so suggestions for more tests would be appreciated.


stuartarchibald:
Thanks for the PR @itamarst, RE additional testing, I suggest looking at NumPy's test suite for tests of `np.tensordot` functionality for inspiration.

itamarst:
NumPy has surprisingly few tests, unless I'm searching badly, but I added a couple more and a caveat pointed out by a third.

stuartarchibald:
> NumPy has surprisingly few tests, unless I'm searching badly, but I added a couple more and a caveat pointed out by a third.

Great, thanks for checking and adding these. So as to get this reviewed, please could you resolve the conflicts (looks like its just in the documentation) so that CI can re-run? Many thanks!

PR #7296, opened 2021-08-10
Fix generator lowering not casting to the actual yielded type

### LINKED ISSUE #7295: Can't compile tuple generators
```python
@njit
def gen():
    yield 1, 2
    yield 3, 4

for i in gen():
    print(i)
```
I expected the output to be
```
(1, 2)
(3, 4)
```
Instead, I got this error
```
('{i64, i64}', '[2 x i64]')
```

Works fine with only 1 yield
```python
@njit
def gen():
    for i in range(10):
        yield i, i + 1


for i in gen():
    print(i)
```

sklam:
as titled

fixes #7295

### DIFF
--- a/numba/core/lowering.py
+++ b/numba/core/lowering.py
@@ -597,13 +597,15 @@ class Lower(BaseLower):
         # Yield to caller
         val = self.loadvar(inst.value.name)
         typ = self.typeof(inst.value.name)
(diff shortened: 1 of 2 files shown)


stuartarchibald (approved):
Thanks for the patch.

PR #3655, opened 2019-01-06
support numpy.average

slnguyen:
Adding `numpy.average` feature. #3435

PR #7304, opened 2021-08-12
Continue PR#3655: add support for np.average

hadia206:
This is continuation of [PR-3655](https://github.com/numba/numba/pull/3655) to add support for `np.average`


sklam:
I did a review confirming that all review questions originally from https://github.com/numba/numba/pull/3655 are already addressed.

sklam (approved):
Thanks for the patch!

PR #7315, opened 2021-08-18
Link ELF obj as DSO for radare2 disassembly CFG

stuartarchibald:
This improves the amount of analysis r2 can do and also lets
DWARF info be used in the disassembly CFG output.

### DIFF
--- a/numba/core/codegen.py
+++ b/numba/core/codegen.py
@@ -838,16 +838,18 @@ class CPUCodeLibrary(CodeLibrary):
         self._sentry_cache_disable_inspection()
         return _CFG(self, name, py_func, **kwargs)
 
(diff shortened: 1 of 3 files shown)


sklam (approved):
Manually confirmed that radare2 will see the debug information with this patch.

stuartarchibald:
> Manually confirmed that radare2 will see the debug information with this patch.

Thanks @sklam, merging this makes the reversing experience much improved!

stuartarchibald:
@sklam did the seek to the symbol work ok? Since the ABI encoding for the flags was added, the symbols have become a lot longer and it seems like `r2` can only match to the first 61 characters.

sklam:
i only tested on short function names.

PR #7317, opened 2021-08-18
In TBB tsh test switch os.fork for mp fork ctx

stuartarchibald:
As title


sklam:
BFID `numba_smoketest_cpu_yaml_38`

sklam (approved):
Thanks for the patch!

PR #7318, opened 2021-08-18
Defensive copy IR when doing obj looplift.

### LINKED ISSUE #7215: incorrect result on Python 3.7
[Kats](https://github.com/facebookresearch/Kats) ran into a bug with Numba `@jit` (Numba version 0.53.1). This bug doesn't happen on Python 3.8(.3) or Python 3.9(.6), but does happen on Python 3.7.5. I reduced it to the following:

```
from typing import Dict
from numba import jit

@jit(forceobj=True)
def broken(f: Dict[str, float], x: Dict[str, bool]) -> Dict[str, float]:
    result = {}
    for k, v in f.items():
        if x.get(k, True):
            result[k] = v
    return result

if __name__ == "__main__":
    print(broken({"one": 1, "two": 2, "three": 3}, {"two": False}))
```

Expected: `{'one': 1, 'three': 3}`

Actual (on Python 3.7): `{'one': 1}`

Adding additional logging shows that on Python 3.7, only the first entry in `f` is processed.  If the inner `if x.get(k, True)` check is removed, all Python versions produce the same result.

stuartarchibald:
As title. This isolates the mutations made in the various
attempts to compile the looplift IR.

Fixes #7215

### DIFF
--- a/numba/core/compiler.py
+++ b/numba/core/compiler.py
@@ -721,7 +721,9 @@ def compile_ir(typingctx, targetctx, func_ir, args, return_type, flags,
                                        lifted_from=lifted_from)
 
         # compile with rewrites off, IR shouldn't be mutated irreparably
-        norw_cres = compile_local(func_ir.copy(), norw_flags)
+        # deepcopy to create defense against mutation
+        norw_ir_copy = copy.deepcopy(func_ir)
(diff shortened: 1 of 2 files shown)


stuartarchibald:
Closing as #8702 will result in the code paths changing to be more isolated which may just fix the original problem.

Issue #4927, opened 2019-12-05
Importing numba is slow


cgohlke:
Hello,

on my Windows workstation with >1200 pip installed packages, `import numba` is very slow:
```
py -3.7-64 -X importtime -c "import numba"
import time: self [us] | cumulative | imported package
import time:   3069020 |    3235357 |           pkg_resources
import time:      2526 |    3237883 |         numba.entrypoints
import time:     31084 |    4448066 | numba
```
The culprit is https://github.com/numba/numba/blob/310fe841c4744129347c5ef8e68d193a4d26c392/numba/entrypoints.py#L3

Importing `pkg_resources` is known to be excruciatingly slow. See https://github.com/pypa/setuptools/issues/510

How about replacing the use of `pkg_resources` for entrypoints by [`importlib_metadata`](https://importlib-metadata.readthedocs.io/en/latest/using.html#entry-points) or [`entrypoints`](https://entrypoints.readthedocs.io/en/latest/) if possible?

## Reporting a bug

- [x] I am using the latest released version of Numba (most recent is visible in
 the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [x] I have included below a minimal working reproducer (if you are unsure how
 to write one see http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports).

stuartarchibald:
Thanks for the report. IIRC this approach mirrors that taken in pandas for the plotting backends https://github.com/pandas-dev/pandas/blob/0c2b1db198ce628ca889ad25c1179fff4ab3337c/pandas/plotting/_core.py#L1613-L1615 but it would appear they delay this scanning until needed. It seems like the suggested fixes involve taking on dependencies. CC @seibert, thoughts?

seibert:
Adding a dependency on `import_metadata` I think would be fine.  It is pure Python, and already packaged in Anaconda as well.  It also appears in the standard library as of Python 3.8:

https://docs.python.org/3/library/importlib.metadata.html

so this seems like heading toward where Python is already going.

stuartarchibald:
Task here is to remove the use of `pkg_resources` and replace with `importlib_metadata` and fix up the builds for that dependency chain.

PR #7321, opened 2021-08-19
Speed up import time by not importing `unittest` or `multiprocessing` until it is needed

eric-wieser:
This also avoids importing `multiprocessing` for the trivial `cpu_count()` function that [just wraps](https://github.com/python/cpython/blob/bb3e0c240bc60fe08d332ff5955d54197f79751c/Lib/multiprocessing/context.py#L41-L47) `os.cpu_count()` in a way that isn't useful.

This reduces the import time by 7% on my machine (measured with `py -3.8 -S -Ximporttime -c "import numba"`, either by comparing before and after, or subtracting the components due to these imports). Here is the graph before this patch:

image

The two large pale-blue elements are the ones eliminated by this PR.
The next-lowest hanging fruit is #4927, which would contribute substantially to reducing the remaining import time.

### DIFF
--- a/numba/__init__.py
+++ b/numba/__init__.py
@@ -17,7 +17,6 @@ del generate_version_info
 
 
 from numba.core import config
-from numba.testing import _runtests as runtests
 from numba.core import types, errors
 
 # Re-export typeof
@@ -60,8 +59,11 @@ import numba.core.target_extension
 import numba.typed
 
 # Keep this for backward compatibility.
-test = runtests.main
-
+def test(argv, **kwds):
+    # To speed up the import time, avoid importing `unittest` and other test
+    # dependencies unless the user is actually trying to run tests.
+    from numba.testing import _runtests as runtests
+    return runtests.main(argv, **kwds)
 
 __all__ = """
     cfunc
--- a/numba/core/config.py
+++ b/numba/core/config.py
@@ -3,7 +3,6 @@ import sys
 import os
 import re
 import warnings
-import multiprocessing
 
 # YAML needed to use file based Numba config
 try:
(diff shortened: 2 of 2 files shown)


gmarkall:

gmarkall:
gpuci run tests

gmarkall (changes requested):
Many thanks for this! It speeds up import by about 2-3% on my machine (I had to import site though, to allow llvmlite to import the way I've got things set up).

The changes look good to me, with one exception - it appears to remove the configuration variable `NUMBA_DEFAULT_NUM_THREADS`, which is referenced here in the documentation: https://numba.readthedocs.io/en/latest/reference/envvars.html#envvar-NUMBA_NUM_THREADS (in the description).

Since `num_threads_default()` still runs at import time, can its result be kept in `NUMBA_DEFAULT_NUM_THREADS` so that this variable is still available, in case anyone is using it?

gmarkall:
Note that #7627 is RTM, so this PR needs no further action.

eric-wieser:
> it appears to remove the configuration variable `NUMBA_DEFAULT_NUM_THREADS`, which is referenced here in the documentation:

Ah, I assumes these were just local variables, I wasn't expecting something to be iterating through `locals()` and putting them in the global namespace!

gmarkall:
> Ah, I assumed these were just local variables, I wasn't expecting something to be iterating through `locals()` and putting them in the global namespace!

Yes, `config.py` is horrible! :-)

gmarkall:
... We have talked about making `numba.core.config` an object instead of a module, which would make it a bit more sane, but I think nobody's yet found the time to try out making that change.

PR #7329, opened 2021-08-22
Improve documentation in reference to CUDA local memory

### LINKED ISSUE #7328: Possible misnomer in cuda.local.array memory documentation
Under [Numba Docs: Memory Management](https://numba.readthedocs.io/en/stable/cuda-reference/kernel.html?highlight=cuda.local.array#numba.cuda.local.array), for [`numba.cuda.local.array()`](https://numba.pydata.org/numba-doc/latest/cuda/memory.html#local-memory) it states:
> ### Local Memory
> Local memory is an area of memory private to each thread. Using local
> memory helps allocate some scratchpad area when scalar local variables
> are not enough. The memory is allocated once for the duration of the
> kernel, unlike traditional dynamic memory management.
> 
> __numba.cuda.local.array(shape, type)__
>
> Allocate a local array of the
> given shape and type on the device. shape is either an integer or a
> tuple of integers representing the array’s dimensions and must be a
> simple constant expression. type is a Numba type of the elements
> needing to be stored in the array. The array is private to the current
> thread. An array-like object is returned which can be read and written
> to like any standard array (e.g. through indexing).

It was pointed out that in the comments that use of the term "allocation" in this context may be a misnomer: https://stackoverflow.com/questions/68869806/how-is-performance-affected-by-using-numba-cuda-local-array-compared-with-numb

Should I submit a PR for the docs? Suggestions for what to say?

From the [NVIDIA CUDA docs on memory access](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#device-memory-accesses):

> ### Local Memory
> Local memory accesses only occur for some automatic variables as mentioned in Variable Memory Space Specifiers. Automatic variables that the compiler is likely to place in local memory are:
> 
> Arrays for which it cannot determine that they are indexed with constant quantities,
> Large structures or arrays that would consume too much register space,
> Any variable if the kernel uses more registers than available (this is also known as register spilling).
> Inspection of the PTX assembly code (obtained by compiling with the -ptx or-keep option) will tell if a variable has been placed in local memory during the first compilation phases as it will be declared using the .local mnemonic and accessed using the ld.local and st.local mnemonics. Even if it has not, subsequent compilation phases might still decide otherwise though if they find it consumes too much register space for the targeted architecture: Inspection of the cubin object using cuobjdump will tell if this is the case. Also, the compiler reports total local memory usage per kernel (lmem) when compiling with the --ptxas-options=-v option. Note that some mathematical functions have implementation paths that might access local memory.
> 
> The local memory space resides in device memory, so local memory accesses have the same high latency and low bandwidth as global memory accesses and are subject to the same requirements for memory coalescing as described in Device Memory Accesses. Local memory is however organized such that consecutive 32-bit words are accessed by consecutive thread IDs. Accesses are therefore fully coalesced as long as all threads in a warp access the same relative address (e.g., same index in an array variable, same member in a structure variable).
> 
> On some devices of compute capability 3.x local memory accesses are always cached in L1 and L2 in the same way as global memory accesses (see Compute Capability 3.x).
> 
> On devices of compute capability 5.x and 6.x, local memory accesses are always cached in L2 in the same way as global memory accesses (see Compute Capability 5.x and Compute Capability 6.x).

sgbaird:
See https://stackoverflow.com/questions/68869806/how-is-performance-affected-by-using-numba-cuda-local-array-compared-with-numb. As I am new to CUDA computing, I don't know if my suggested edits are correct and am merely passing on the implication in the linked SE post. In the CUDA docs under local memory, I didn't see any use of the word "allocate", so perhaps this is a correct suggestion. If the original usage is fine, please reject this pull request.

Fixes #7328.

### DIFF
--- a/docs/source/cuda/memory.rst
+++ b/docs/source/cuda/memory.rst
@@ -152,21 +152,23 @@ traditional dynamic memory management.
 Local memory
 ============
 
-Local memory is an area of memory private to each thread.  Using local
-memory helps allocate some scratchpad area when scalar local variables
-are not enough.  The memory is allocated once for the duration of the kernel,
-unlike traditional dynamic memory management.
+Local memory is an area of memory private to each thread.  Using local memory
+helps reserve some scratchpad area to variables when scalar local variables are
+not enough.  The memory is allocated once for the duration of the kernel, unlike
+traditional dynamic memory management.
 
 .. function:: numba.cuda.local.array(shape, type)
    :noindex:
 
-   Allocate a local array of the given *shape* and *type* on the device.
+   Create a variable representing a local array of the given *shape* and *type*
+   on the device that corresponds to a chunk of statically defined memory.
    *shape* is either an integer or a tuple of integers representing the array's
-   dimensions and must be a simple constant expression. *type* is a
-   :ref:`Numba type <numba-types>` of the elements needing to be stored in the
-   array. The array is private to the current thread. An array-like object is
-   returned which can be read and written to like any standard array
-   (e.g. through indexing).
+   dimensions and must be a simple constant expression. *type* is a :ref:`Numba
+   type <numba-types>` of the elements needing to be stored in the array. The
+   array is private to the current thread. An array-like object is returned
+   which can be read and written to like any standard array (e.g. through
+   indexing). See also the Local Memory section of `Device Memory Accesses
+   <https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#device-memory-accesses>`_
 
 Constant memory
 ===============
--- a/docs/source/cuda/memory.rst
+++ b/docs/source/cuda/memory.rst
@@ -152,16 +152,15 @@ traditional dynamic memory management.
 Local memory
 ============
 
-Local memory is an area of memory private to each thread.  Using local memory
-helps reserve some scratchpad area to variables when scalar local variables are
-not enough.  The memory is allocated once for the duration of the kernel, unlike
-traditional dynamic memory management.
+Local memory is an area of memory private to each thread.  Using local
+memory helps allocate some scratchpad area when scalar local variables
+are not enough.  The memory is allocated once for the duration of the kernel,
+unlike traditional dynamic memory management.
 
 .. function:: numba.cuda.local.array(shape, type)
    :noindex:
 
-   Create a variable representing a local array of the given *shape* and *type*
-   on the device that corresponds to a chunk of statically defined memory.
+   Allocate a local array of the given *shape* and *type* on the device.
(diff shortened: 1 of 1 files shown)


docs/source/cuda/memory.rst
```diff
@@ -152,21 +152,23 @@ traditional dynamic memory management.
...
-   :ref:`Numba type <numba-types>` of the elements needing to be stored in the
-   array. The array is private to the current thread. An array-like object is
-   returned which can be read and written to like any standard array
-   (e.g. through indexing).
+   dimensions and must be a simple constant expression. *type* is a :ref:`Numba
+   type <numba-types>` of the elements needing to be stored in the array. The
+   array is private to the current thread. An array-like object is returned
+   which can be read and written to like any standard array (e.g. through
+   indexing). See also the Local Memory section of `Device Memory Accesses
+   <https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#device-memory-accesses>`_
```
gmarkall, line 171: This could be rendered using the `seealso` directive to make it stand out a little more clearly, e.g.:

```suggestion
   indexing).

   .. seealso:: The Local Memory section of `Device Memory Accesses
      <https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#device-memory-accesses>`_
      in the CUDA programming guide.
```

gmarkall (changes requested):
Many thanks for the PR! (Also, many thanks for all the other contributions that you're making on Discourse and the issue tracker - it's great to have an examination of all these details that have become somewhat unattended-to over the years)

In my view the local memory is allocated - it's a static allocation done at compile time, that happens to be implemented using the same device memory as global memory is. Therefore, I feel that avoiding the use of "allocate" in the documentation makes the explanation of the concept less clear than it was before, so I'd prefer to keep the original wording. 

I do like the addition of the link to the explanation of local memory in the CUDA Programming guide though, and I think that is a useful addition. As it's a "See also"-type remark, I think it would be good to use the `seealso` RST directive to make it stand out (see the suggestion on the diff for an example of how this could look).

So, to move forward, could you please:

1. Undo the changes to the wording, or, if you feel that the original wording is inaccurate / unhelpful, explain a little more about your thinking behind the changes, or suggest another wording?
2. Adjust the link to the CUDA Programming guide to use the `seealso` directive?

Many thanks so far and in advance!

sgbaird:
I think the changes are removed now, and I accepted the see also edit. It's been very useful to see your suggestions to the PRs. Thank you! Let me know if there's anything else on this one I need to do.

gmarkall:
Many thanks for the update, and your contribution to Numba! This looks good to me now.

gmarkall (approved):

PR #7340, opened 2021-08-24
Add size_t and ssize_t types

bmerry:
These are mainly for use with `cfunc`, to match the defined signature of
a C API. In flat-memory ABIs these are the same as uintptr_t and
intptr_t, but's not guaranteed to hold forever. More practically, having
these names will make code clearer, as it will match the C names for the
interface being implemented, and make it clear that a parameter is
holding a size/count rather than a bit-cast pointer.

### DIFF
--- a/docs/source/reference/types.rst
+++ b/docs/source/reference/types.rst
@@ -65,6 +65,8 @@ intc                    --               C int-sized integer
 uintc                   --               C int-sized unsigned integer
 intp                    --               pointer-sized integer
 uintp                   --               pointer-sized unsigned integer
+ssize_t                 --               C ssize_t
+size_t                  --               C size_t
 
 float32                 f4               single-precision floating-point number
 float64, double         f8               double-precision floating-point number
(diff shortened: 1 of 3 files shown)


bmerry:
I'm not sure what testing would be most appropriate. These are aliases so I'm assuming all the typing machinery will simply work, but I guess a basic smoke test wouldn't hurt. If you have a suggestion for where to put it and what it should do, let me know.

gmarkall (approved):
Looks good to me, pending a successful run on the buildfarm!

stuartarchibald:
Buildfarm ID: `numba_smoketest_cpu_yaml_42`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cpu_yaml_42`.

This passed.

stuartarchibald (approved):
Thanks for the patch @bmerry, thanks for reviewing @gmarkall !

stuartarchibald:
@bmerry Congratulations on your first contribution to Numba!

PR #7347, opened 2021-08-26
fix:updated url for error report and feature rquest using issue template

Stark-developer01:


esc (approved):
Thanks for the patch!

PR #7348, opened 2021-08-26
test_inspect_cli: Decode exception with default (utf-8) codec

### LINKED ISSUE #7337: Test failure due to non-ascii in stdout
From OOB conversation, a user reported the following error from this line:

https://github.com/numba/numba/blob/e031b70c31c8553521ee89b02db8d2d6e995f9eb/numba/tests/test_help.py#L92

Error:
```pytb
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 38: ordinal not in range(128)
```

It decode with 'utf8' instead.

gmarkall:
Fixes #7337.


stuartarchibald (approved):
Thanks for the patch.

PR #7349, opened 2021-08-26
Allow arbitrary walk-back in reduction nodes to find inplace_binop.

### LINKED ISSUE #7344: incoherent behaviour with prange
First, thanks a lot for this very useful job !

Here is the problem: I slightly change the example of prange given in the documentation and get a very strange behaviour. Here is my code
````python
@njit(parallel=True)
def prange_test(A, cond):
    s = 0
    for i in prange(A.shape[0]):
        if cond[i]:
            s += 1
    return s
````
Here are my inputs
````python
A = np.random.randint(0,10,100)
cond = np.random.randint(0,2,len(A)).astype('bool')
````
In this case scenario changing njit option from True to False do not affect the result of `prange_test(A, cond)` (which is what I expect)

However, setting `s=1` as a starting condition in `prange_test()` definition make `prange_test(A, cond)` results to be different depending of the parallel value !
If ``parallel==False``, the result is ok;  If ``parallel==True``, the results is larger of the quantity: 1 x number_threads +1.
Setting `s=2` as a starting condition in `prange_test()` definition makes the result to be larger of quantity 2 x number_of_threads +2.

This strange behavior disappears if there is no check of the ``cond`` value in `prange_test(A, cond)` (it returns the length of A+initial s value) whatever the initial value of ``s`` 
It seems inconsistent as it gives wrong results and threads dependent. I do not understand why checking  ``cond' creates inconsistencies.
Thanks for the help !

### PR DESCRIPTION
**Pull request by DrTodd13**
Resolves #7344 

Previously, there could only be one extra assignment after the inplace_binop node for a reduction before the assignment to the reduction variable.  In this issue, there was an additional assignment.  The solution is to start at the enf with the reduction variable and walk backwards through the reduce nodes so long as they are all assignments and link with each other until we get to inplace_binop.

### DIFF
--- a/numba/parfors/parfor.py
+++ b/numba/parfors/parfor.py
@@ -3570,7 +3570,7 @@ def get_reduction_init(nodes):
     # there could be multiple extra assignments after the reduce node
     # See: test_reduction_var_reuse
     cur_index = 1
-    acc_expr = nodes[cur_index].value
+    acc_expr = nodes[-cur_index].value
     while isinstance(acc_expr, ir.Var):
         require(len(nodes) >= cur_index + 1)
         require(nodes[-(cur_index + 1)].target.name == nodes[-cur_index].value.name)
--- a/numba/parfors/parfor.py
+++ b/numba/parfors/parfor.py
@@ -3569,14 +3569,7 @@ def get_reduction_init(nodes):
     require(len(nodes) >= 1)
     # there could be multiple extra assignments after the reduce node
     # See: test_reduction_var_reuse
-    cur_index = 1
-    acc_expr = nodes[-cur_index].value
-    while isinstance(acc_expr, ir.Var):
-        require(len(nodes) >= cur_index + 1)
-        require(nodes[-(cur_index + 1)].target.name == nodes[-cur_index].value.name)
-        cur_index += 1
-        acc_expr = nodes[-cur_index].value
-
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: numba/parfors/parfor.py [hunk 1]
```diff
@@ -3566,15 +3566,17 @@ def get_reduction_init(nodes):
...
-        acc_expr = nodes[-2].value
-    else:
-        acc_expr = nodes[-1].value
+    cur_index = 1
+    acc_expr = nodes[-cur_index].value
+    while isinstance(acc_expr, ir.Var):
+        require(len(nodes) >= cur_index + 1)
+        require(nodes[-(cur_index + 1)].target.name == nodes[-cur_index].value.name)
+        cur_index += 1
+        acc_expr = nodes[-cur_index].value
```
stuartarchibald, line 3578: It seems like this might be relying on a naming pattern? The following variant on the original MWR breaks for variable `t`:
```python
import numpy as np

@njit(parallel=True)
def test_impl(A, cond):
    s = 1
    t = 10
    for i in prange(A.shape[0]):
        if cond[i]:
            s += 1
            t += 1
        else:
            s += 2
    return s, t

print(test_impl(np.ones(10), (np.arange(10) % 2).astype('bool')))
```

stuartarchibald (commented):
Thanks for the patch, the fix does make the original MWR work correctly but there seems to be a similar issue in a variant. Thanks for taking a look at it!

stuartarchibald (approved):

PR #7345, opened 2021-08-25
Add check for ipykernel file in IPython cache locator

sahil1105:
## Reference an existing issue

Closes #7325 

This is just the potential fix that I attached with the issue itself. 
Looking for feedback regarding the best way to fix this issue.

PR #7357, opened 2021-08-31
modified ipython locator conditions for when ipython points to /tmp

DannyWeitekamp:
Recently on a fresh jupyter/ipython installation on Ubuntu 20 + python 3.8.10, I found that numba wouldn't work w/ njit(cache=True). Usually when running:

```python
import inspect
from numba import njit

def foo():
    return 1

print(inspect.getfile(foo))

njit_foo = njit(cache=True)(foo)
print(njit_foo())
```
You would get something like 
```
<ipython-input-X-XXXXXXXXX>
1
```
For for some reason (maybe a different version of jupyter) on this fresh install it pointed to the /tmp directory like this
```
/tmp/ipykernel_53800/2759104420.py
1
```

This fix just proceeds with the ipython locator if "ipykernel" is in the function's file path.

For reference `jupyter --version` looks like this:
```
jupyter core     : 4.6.3
jupyter-notebook : 6.4.3
qtconsole        : 5.1.1
ipython          : 7.27.0
ipykernel        : 6.3.1
jupyter client   : 7.0.2
jupyter lab      : not installed
nbconvert        : 6.1.0
ipywidgets       : 7.6.3
nbformat         : 5.1.3
traitlets        : 4.3.3
```

I believe this fixes issue #7345

### DIFF
--- a/numba/core/caching.py
+++ b/numba/core/caching.py
@@ -307,7 +307,7 @@ class _IPythonCacheLocator(_CacheLocator):
 
     @classmethod
     def from_function(cls, py_func, py_file):
-        if not py_file.startswith("<ipython-"):
+        if not py_file.startswith("<ipython-") and "ipykernel" not in py_file:
             return
         self = cls(py_func, py_file)
         try:


DannyWeitekamp:
Shoot looks like #7345 is actually already fixed in another PR. Going to close this.

stuartarchibald:
@DannyWeitekamp many thanks for opening this patch. I also think that #7345 will fix, it just needs tests writing then can be merged.

Issue #6473, opened 2020-11-12
Array assignment to a structure's nested array 


ceo-domido:
Impossible to assign an array to the nested array in the structured array. Without numba all the options below are working, but with numba only the last option works. 

- The first option falls with an error "Can only insert i8 at [0] in [1024 x i8]: got i8*". 
- The second option falls with "Buffer dtype cannot be buffer, have dtype: nestedarray(uint8, (1024,))"
- The third option falls with "Buffer dtype cannot be buffer, have dtype: nestedarray(uint8, (1024,))"

```
import numba as nb
import numpy as np

BLOCK = np.dtype([("id", np.uint64), ("props", (np.uint8, 1024))])

@nb.njit()
def set_props1(_source, _target):
    _target[0]["props"] = _source[0]["props"]

@nb.njit()
def set_props2(_source, _target):
    _target["props"][0] = _source["props"][0]

@nb.njit()
def set_props3(_source, _target):
    for i in range(1024):
        _target["props"][0, i] = 1

@nb.njit()
def set_props4(_source, _target):
    for i in range(1024):
        _target[0]["props"][i] = 1

source = np.ones(1, dtype=BLOCK)
target = np.zeros(1, dtype=BLOCK)
set_props1(source, target)
print(target)
set_props2(source, target)
print(target)
set_props3(source, target)
print(target)
set_props4(source, target)
print(target)

```

esc:
@ceo-domido - thank you for reporting this on the Numba issue tracker. I can reproduce this locally and it seems fine with Numpy but broken with Numba.

stuartarchibald:
This is a bug. Something like this probably works fine:
```python
import numba as nb
import numpy as np

BLOCK = np.dtype([("id", np.uint64), ("props", (np.uint8, 1024))])

@nb.njit()
def set_props1(_source, _target):
    _target[0]["props"][:] = _source[0]["props"] # NOTE THE SLICE

source = np.ones(1, dtype=BLOCK)
target = np.zeros(1, dtype=BLOCK)
set_props1(source, target)
print(target)
```

HPLegion:
I think I ran into a related isse: https://numba.discourse.group/t/structured-arrays-with-nd-fields/496

astoeriko:
I think that I am seeing the same issue with the current numba version (0.53.1) when trying to convert to a numpy array:
```python
import numba
import numpy as np

@numba.njit
def foo(val):
    x = np.empty_like(val.x)
    x[:] = val.x
[...]

PR #7359, opened 2021-09-02
Extend support for nested arrays inside numpy records

### LINKED ISSUE #7355: BUG cannot set items in structured array data types in 0.54.0
## Reporting a bug

Here is the code sample:

```
import numpy as np
from numba import njit


@njit
def set_field(arr):
    arr['a'][:] = 0.0


if __name__ == "__main__":
    for _ in range(10):
        arr = np.ones(1, dtype=[('a', 'f8', (6,))])
        set_field(arr[0])
        assert np.all(arr['a'] == 0)
```

This code sample works with 0.53.1 but does not work on 0.54.0. Here is the error

```
  File "/Users/beckermr/Desktop/numba_bug.py", line 13, in <module>
    set_field(arr[0])
  File "/Users/beckermr/miniconda3/envs/numba-bug/lib/python3.9/site-packages/numba/core/dispatcher.py", line 482, in _compile_for_args
    error_rewrite(e, 'typing')
  File "/Users/beckermr/miniconda3/envs/numba-bug/lib/python3.9/site-packages/numba/core/dispatcher.py", line 423, in error_rewrite
    raise e.with_traceback(None)
 
 >>> setitem(nestedarray(float64, (6,)), slice<a:b>, float64)
 
    With argument(s): '(nestedarray(float64, (6,)), slice<a:b>, float64)':
   No match.
  Overload in function 'SetItemBuffer.generic': File: numba/core/typing/arraydecl.py: Line 171.
    With argument(s): '(nestedarray(float64, (6,)), slice<a:b>, float64)':
     TypeError: __init__() got an unexpected keyword argument 'ndim'
  raised from /Users/beckermr/miniconda3/envs/numba-bug/lib/python3.9/site-packages/numba/core/types/abstract.py:66

During: typing of staticsetitem at /Users/beckermr/Desktop/numba_bug.py (7)

File "numba_bug.py", line 7:
def set_field(arr):
    arr['a'][:] = 0.0
```

A conda env with the correct versions pulled from conda-forge should be sufficient to reproduce the bug on either OSX or Linux.

cc @esheldon

luk-f-a:
closes #7355 #6473

This PR extends the support for setitem and getitem on nested array, those embedded in records. These arrays are fixed size, so `shape` is part of the type unlike normal `Array` where only `ndim` is.

The changes are mainly to typing and lowering where scalar fields are assumed, adding branches to split between scalar and nestedarray fields.


luk-f-a:
> Do you need some help looking into the CUDA side of things?

yes, please, @gmarkall . I haven't even written CUDA code, being able to fix a typing error sounds unlikely.

luk-f-a:
@gmarkall thanks for offering to help. A few very basic questions about the cuda target:
- what does "is not aligned" mean in the error message `TypeError: Record(g[type=int32;offset=0],h[type=nestedarray(float32, (2,));offset=4];12;True).h of type nestedarray(float32, (2,)) is not aligned`?
- I thought I was only changing cpu code, but apparently this is shared. What's the relationship between the cuda and cpu targets? is the typing shared and the lowering separate?
- If they are shared, why does the cpu target not show this type error? If they are separate, why did changes I made on the cpu target affect the cuda target?

Thanks!

gmarkall:
@luk-f-a No problem, just starting to have a look at this now.

> what does "is not aligned" mean in the error message TypeError: Record(g[type=int32;offset=0],h[type=nestedarray(float32, (2,));offset=4];12;True).h of type nestedarray(float32, (2,)) is not aligned?

It means that its offset from the base address of the record is not a multiple of its size. I believe the CPU target supports unaligned members, but CUDA doesn't so that may be why we see this error with CUDA but not the CPU target.

> I thought I was only changing cpu code, but apparently this is shared. What's the relationship between the cuda and cpu targets? is the typing shared and the lowering separate?

The CUDA target is (from the typing, lowering, and data model perspective) the same as the nopython mode target, except that it doesn't have some of the typing / lowering registered (for things like typed lists, typed sets, NumPy array functions) for things it doesn't support. It also has some extra typings and lowerings registered for CUDA-specific functions like `cuda.grid()` etc.

> If they are shared, why does the cpu target not show this type error? If they are separate, why did changes I made on the cpu target affect the cuda target?

They are shared, and I think the reason is that mentioned above, that the CPU target supports unaligned record members.

luk-f-a:
thanks, I was looking at the code, and understood the point about alignment.
The error message complains about the record not being aligned. What's the `get_abi_alignment` of the cuda target? The example has a int32 followed by float32[2], so the offset of `h` is 4. For `offset % align` to be false, then `align = self.get_abi_alignment(self.get_data_type(elemty))` cannot 2 or 4. Is it 8?

gmarkall:
(I'll post updates as I make progress, as I don't know how far I'll get tonight)

Here:

https://github.com/numba/numba/blob/fd6a0592a34ed31c76e2f0c212d57cacf4680b5c/numba/core/base.py#L966-L977

The offset of the field is 4.

Prior to your changes, the data type is `[2 x float]` and the ABI alignment for this is computed as `4`.

After your changes, the data type is `{i8*, i8*, i64, i64, float*, [1 x i64], [1 x i64]}` (an array struct) and the alignment of this array struct is `8`.

Therefore, the alignment check fails - something needs to be done about this alignment check.

luk-f-a:
absolutely, I don't mind if you push to this PR.

I was looking at this yesterday, and it's definitely different in CPU, because align is always correct (in this case = 4) whether I used `self.get_abi_alignment(self.get_data_type(elemty))` or `self.get_abi_alignment(self.get_data_type(elemty.dtype))`.

I tried to trace `get_abi_alignment` to see why this happened, but after a few jumps I landed in llvmlite and I don't think there's a logic in either numba or llvmlite that determines this, it must come from llvm, or some C library.

I don't mind the work of copying the tests to into some test case in cuda.tests, but I'm guessing that's the easy part. I still can do it if it helps, but dealing with tests failures would be out of my reach.

numba/tests/test_record_dtype.py
```diff
@@ -223,14 +223,60 @@ def record_write_2d_array(ary):
...
+
+def recarray_getitem_field_return2_2d(ary):
+    return ary.j
+
+
 def record_read_array0(ary):
     return ary.h[0]
 
 
+def record_read_array0_alt(ary):
```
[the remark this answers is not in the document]
luk-f-a: thx, I added this function to a test (you can find the change in the commit "address unused record_read_array0_alt")

numba/tests/test_record_dtype.py
```diff
@@ -223,14 +223,60 @@ def record_write_2d_array(ary):
...
+    return ary[0].h
+
+
 def record_read_array1(ary):
     return ary.h[1]
 
 
+def record_read_whole_array(ary):
```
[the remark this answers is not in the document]
luk-f-a: yes, a test was missing. I added it in commit "added test_record_read_arrays". In commit "gpu test_record_read_arrays" I tried to copy what you did to enable the test on gpu.

numba/core/datamodel/models.py
```diff
@@ -896,9 +896,9 @@ def __init__(self, dmm, fe_type):
         self._be_type = dmm.lookup(fe_type.dtype).get_data_type()
         super(NestedArrayModel, self).__init__(dmm, fe_type)
 
-    def get_data_type(self):
-        ret = ir.ArrayType(self._be_type, self._fe_type.nitems)
-        return ret
+    # def get_data_type(self):
```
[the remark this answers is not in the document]
luk-f-a: done

numba/core/typing/arraydecl.py
```diff
@@ -495,7 +495,11 @@ def generic_resolve(self, ary, attr):
         # Resolution of other attributes, for record arrays
         if isinstance(ary.dtype, types.Record):
             if attr in ary.dtype.fields:
-                return ary.copy(dtype=ary.dtype.typeof(attr), layout='A')
+                attr_dtype = ary.dtype.typeof(attr)
+                if isinstance(attr_dtype, types.NestedArray):
+                    return ary.copy(dtype=attr_dtype.dtype, ndim=ary.ndim + attr_dtype.ndim, layout='A')
```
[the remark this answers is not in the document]
luk-f-a: I'm not sure why flake8 didn't tag it, but I cleaned most of the flake8 warnings in the file (including some from older code).

numba/tests/test_record_dtype.py
```diff
@@ -1362,23 +1342,188 @@ def foo(rec, flag=None):
         self.assertEqual(foo(self.ab_rec1, flag=1), self.ab_rec1[0] + k + 20)
 
 
-class TestRecordArrayExceptions(TestCase):
+class TestNestedArrays(TestCase):
```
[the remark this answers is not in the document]
luk-f-a: done.

luk-f-a:
@gmarkall thanks for your review and added tests for gpu. I have addressed your comments, keeping each change to a different commit to hopefully make it easier to you to check individual changes.
I have added one gpu test so it'd be good to run gpuci again.

gmarkall:
gpuci run tests

gmarkall:
Looks like the GPU test is failing, but I'll have a look into it and push a fix.

gmarkall:
@luk-f-a The test was good, but CUDA couldn't support the read of an entire array just because it doesn't handle returning arrays, so I've xfailed the test.

gpuci run tests.

stuartarchibald:
@luk-f-a @gmarkall thanks for working on this. As CI is passing, is this now ready for review?

gmarkall:
Yes, this is ready for review and I forgot to set the label after gpuCI passed - apologies.

numba/core/types/npytypes.py
```diff
@@ -587,3 +587,15 @@ def strides(self):
...
+        if dtype is None:
+            dtype = self.dtype
+        if ndim is None:
+            ndim = self.ndim
+        if layout is None:
+            layout = self.layout
+        if readonly is None:
+            readonly = not self.mutable
+        cls = type(self)
+        return cls(dtype=dtype, shape=self.shape)
```
gmarkall, line 601: Should `ndim`, `layout`, and `readonly` be used here somehow?

numba/core/typing/arraydecl.py
```diff
@@ -503,7 +504,15 @@ def generic_resolve(self, ary, attr):
         # Resolution of other attributes, for record arrays
         if isinstance(ary.dtype, types.Record):
             if attr in ary.dtype.fields:
-                return ary.copy(dtype=ary.dtype.typeof(attr), layout='A')
+                attr_dtype = ary.dtype.typeof(attr)
+                if isinstance(attr_dtype, types.NestedArray):
+                    return ary.copy(
+                        dtype=attr_dtype.dtype,
+                        ndim=ary.ndim + attr_dtype.ndim,
```
gmarkall, line 511: @stuartarchibald I think you previously expressed some concern to me OOB that the addition of the dimensions of the "outer" array and the nested array were being summed to create the new dimension of the array. However, this makes sense to me. For example:

```python
import numpy as np

dtype = np.dtype([
    ('x', np.float32),
    ('y', np.float32, (4, 2))
])

x = np.zeros((5, 7), dtype=dtype)
print(len(x['y'].shape))
```

prints `4`, because getting an item of an array of records where the item is a nested array "slices" the array such that you get an array of all the nested arrays (the shape of `x['y']` here is `(5, 7, 4, 2)` - the dimensions of the array concatenated with the dimensions of the nested array).

numba/core/types/npytypes.py
gmarkall: Or, is this change absolutely necessary? I can't quite tell why this omission doesn't make tests fail, or maybe it isn't hit by any of the tests?

gmarkall (changes requested):
This broadly looks good to me, but I do have a couple of questions on the diff.

@stuartarchibald / @sklam Since I wrote the CUDA tests (and added a little fix to make it work for CUDA), would you like to check the CUDA tests and other changes in my commits?

numba/core/types/npytypes.py
gmarkall: After some more investigation:

- I think this is copy-pasted from `Array` above, which is why it is the way it presently is?
- It seems this function is only called here: https://github.com/numba/numba/blob/master/numba/core/typing/arraydecl.py#L153

I think it would be more natural to accept the parameters passed by `get_array_index_type()` and validate that they match the existing shape of the `NestedArray`, since we can't reasonably change it for a copy. This is implemented in https://github.com/gmarkall/numba/commit/7d21f75fbb2807860e4a279995e29c4c6003bd3a in the branch https://github.com/gmarkall/numba/tree/feature/nested_arrays_getitem - what do you think about this approach? Would you be happy with adding it to this PR?

It feels like the right thing to do would be

gmarkall:
@luk-f-a This will also fix https://github.com/numba/numba/issues/7526#issuecomment-979325826 - would you be happy to add a test case for `np.shape` on a nested array as well, then we can be confident that that issue will be / stay resolved?

numba/core/types/npytypes.py
luk-f-a: you make a great point. I looked into this and thought of a new test, which I added in this commit: https://github.com/numba/numba/pull/7359/commits/ac492bffd34bc9c2f1152aeab62a960f204d695f

Interestingly, both the first version and the one in your branch fail the test. However, if I remove them (and let the base class method take care of this) the test passes. I need to think what this means. Maybe this operation should be handled differently, or not allowed, but I'm not sure. I'm trying to think of potential negative side effects of allowing the base class method take care of this case.

luk-f-a:
> @luk-f-a This will also fix [#7526 (comment)](https://github.com/numba/numba/issues/7526#issuecomment-979325826) - would you be happy to add a test case for `np.shape` on a nested array as well, then we can be confident that that issue will be / stay resolved?

sure, I have added a test for shape based on your example.

numba/core/types/npytypes.py
gmarkall: An `Array` is more general than a `NestedArray`, so in terms of typing my gut feel is that this is safe. In terms of lowering, I'm wondering whether this would lead to any surprises with NRT, so I'd need to think a little more about that.

luk-f-a:
@gmarkall after our discussion in the call, I'll try to create the edge case @stuartarchibald was talking about. Please review and let me know if you can think of more "edginess".

luk-f-a:
@gmarkall the latest push contains an attempt at a complex test for slicing:
- 7 d array of structs, F order
- struct containing an int64, a 10-bytes string, and a nested 4-d array of float32 dtype
- the slice is not just `[0]` as in the previous example, it's now `[0:2, 0, 1:4, 3:6]`

gmarkall:
gpuci run tests

numba/tests/test_record_dtype.py
```diff
@@ -1553,6 +1566,22 @@ def test_shape(self):
         arr_res = cfunc(arg)
         np.testing.assert_equal(arr_res, arr_expected)
 
+    def test_corner_slice(self):
+        # testing corner cases while slicing nested arrays
+        nbarr = np.recarray((1, 2, 3, 5, 7, 13, 17), dtype=recordwith4darray)
+        for index, _ in np.ndenumerate(nbarr):
+            nbarr[index].p = np.random.randint(0, 1000, (3, 2, 5, 7),
```
gmarkall, line 1574: Can we seed the RNG in this test if we're using `np.random` please? (This just makes it easier when investigating any issues, so that the same numbers are always presented on every run instead of seeing variation from run to run) - e.g. adding `np.random.seed(1)` to the beginning of this test case.

gmarkall:
gpuci run tests

gmarkall:
@luk-f-a Your test looks good to me. I also wondered about a test of a broadcasted recarray - I've added one here: https://github.com/gmarkall/numba/commit/1cce6c32cd771b628a2094b0de559113352ddf7f - what do you think of this test? Do you feel comfortable adding it to this PR branch?

gmarkall:
gpuci run tests

gmarkall (dismissed):
Many thanks for the latest updates and test additions. I think this looks good now - given all the tests we have and ways we've tried to break it I'm pretty confident that this should be robust in the wild (notwithstanding the long-term issue with respect to reference-counting of records that is difficult to solve). I'm happy to approve this but would prefer a buildfarm run to be on the safe side.

@stuartarchibald / @esc Could this have a buildfarm run please?

esc:
`numba_smoketest_cuda_yaml_106` and `numba_smoketest_cpu_yaml_68`

numba/core/typing/arraydecl.py
stuartarchibald: Thanks for the explanation @gmarkall, as noted at the public meeting, my concern was that lowering was working "by accident", however, following the above, I think it does make sense, and also explains why slicing etc works ok.

esc:
> `numba_smoketest_cuda_yaml_106` and `numba_smoketest_cpu_yaml_68`

all green.. 💚

numba/cuda/tests/cudapy/test_record_dtype.py
```diff
@@ -279,5 +343,182 @@ def _createSampleArrays(self):
...
+        np.testing.assert_equal(res, nbval[0].j[0, 1])
+
+        cfunc = self.get_cfunc(record_read_2d_array10, np.float32)
+        res = cfunc(nbval[0])
+        np.testing.assert_equal(res, nbval[0].j[1, 0])
+
+    def test_getitem_idx(self):
+        # Test __getitem__ with numerical index
+
+        # This test returning a record when passing an array and
```
stuartarchibald, line 409:
```suggestion
        # This tests returning a record when passing an array and
```

numba/cuda/tests/cudapy/test_record_dtype.py
```diff
@@ -279,5 +343,182 @@ def _createSampleArrays(self):
...
+        # return the first item when passing a record
```
stuartarchibald, line 410:
```suggestion
        # returning the first item when passing a record
```

numba/cuda/tests/cudapy/test_record_dtype.py
```diff
@@ -279,5 +343,182 @@ def _createSampleArrays(self):
...
+        nbarr = np.recarray(2, dtype=recordwitharray)
+        nbarr[0] = np.array([(1, (2,3))], dtype=recordwitharray)[0]
```
stuartarchibald, line 412:
```suggestion
        nbarr[0] = np.array([(1, (2, 3))], dtype=recordwitharray)[0]
```

numba/np/arrayobj.py
```diff
@@ -2507,14 +2511,30 @@ def array_record_getattr(context, builder, typ, value, attr):
...
+        # new strides = recarray strides + strides of the inner nestedarray
+        strides = cgutils.unpack_tuple(builder, array.strides, typ.ndim)
+        strides += [context.get_constant(types.intp, i) for i in dtype.strides]
+        # new shape = recarray shape + inner dimension from nestedarray
+        shape = cgutils.unpack_tuple(builder, array.shape, typ.ndim)
+        shape += [context.get_constant(types.intp, i) for i in dtype.shape]
+        populate_array(rary,
+                       data=newdataptr,
+                       shape=cgutils.pack_array(builder, shape),
+                       strides=strides,
```
stuartarchibald, line 2525: It seems inconsistent to `pack_array` `shape` but not `strides`?  API looks like it can take either an iterable python container _or_ an LLVM array, and `shape` and `strides` at this call site are both python lists.
https://github.com/numba/numba/blob/3ec12030586647b0b3b81918256eaa4d11d63780/numba/np/arrayobj.py#L180-L201

numba/tests/test_record_dtype.py
```diff
@@ -1371,23 +1368,260 @@ def foo(rec, flag=None):
...
-        self.assertIn(ex2, excstr)
+        nbval = np.recarray(1, dtype=recordwith2darray)
+        nbrecord = numpy_support.from_dtype(recordwith2darray)
+        cfunc = self.get_cfunc(record_write_2d_array, (nbrecord,))
+        cfunc(nbval[0])
+
+        expected = np.recarray(1, dtype=recordwith2darray)
+        expected[0].i = 3
+        expected[0].j[:] = np.asarray([5.0, 6.0, 7.0, 8.0, 9.0, 10.0],
+                                      np.float32).reshape(3, 2)
```
stuartarchibald, line 1402: The expected value should ideally come from execution of the Python function being JIT compiled as that is what this JIT function is attempting to provide a native equivalent of. This comment applies throughout the proposed changes, however, in the interests of getting this functionality on mainline, unless @gmarkall feels strongly otherwise, we can leave it for now.

gmarkall:
@stuartarchibald Were you happy with this too? (re the CUDA changes, etc.)

numba/tests/test_record_dtype.py
```diff
@@ -1371,23 +1368,260 @@ def foo(rec, flag=None):
...
+            np.testing.assert_equal(arr_res, arr_expected)
+
+    def test_getitem_idx(self):
+        # Test __getitem__ with numerical index
+
+        # This test returning a record when passing an array and
+        # return the first item when passing a record
+
+        nbarr = np.recarray(2, dtype=recordwitharray)
+        nbarr[0] = np.array([(1, (2,3))], dtype=recordwitharray)[0]
```
stuartarchibald, line 1505:
```suggestion
        nbarr[0] = np.array([(1, (2, 3))], dtype=recordwitharray)[0]
```

numba/tests/test_record_dtype.py
```diff
@@ -1371,23 +1368,260 @@ def foo(rec, flag=None):
...
+            np.testing.assert_equal(arr_res, arr_expected)
+
+    def test_getitem_idx_2darray(self):
+        # Test __getitem__ with numerical index
+        #
+        # This test returning a record when passing an array and
+        # return the first item when passing a record
+
+        nbarr = np.recarray(2, dtype=recordwith2darray)
+        nbarr[0] = np.array([(1, ((1,2),(4,5),(2,3)))],
```
stuartarchibald, line 1521:
```suggestion
        nbarr[0] = np.array([(1, ((1, 2), (4, 5), (2, 3)))],
```

numba/tests/test_record_dtype.py
```diff
@@ -1371,23 +1368,260 @@ def foo(rec, flag=None):
...
+            np.testing.assert_equal(arr_res, arr_expected)
+
+    def test_return_getattr_getitem_fieldname(self):
+        # Test __getitem__ with field name and getattr .field_name
+        #
+        # This tests returning a array of nestedarrays when passing an array and
+        # returning a nestedarray when passing a record
+
+        nbarr = np.recarray(2, dtype=recordwitharray)
+        nbarr[0] = np.array([(1, (2,3))], dtype=recordwitharray)[0]
```
stuartarchibald, line 1538:
```suggestion
        nbarr[0] = np.array([(1, (2, 3))], dtype=recordwitharray)[0]
```

numba/tests/test_record_dtype.py
```diff
@@ -1371,23 +1368,260 @@ def foo(rec, flag=None):
...
+                arr_expected = pyfunc(arg)
+                cfunc = self.get_cfunc(pyfunc, (ty,))
+                arr_res = cfunc(arg)
+                np.testing.assert_equal(arr_res, arr_expected)
+
+    def test_return_array(self):
+        # Test getitem record AND array within record and returning it
+
+        nbval = np.recarray(2, dtype=recordwitharray)
+        nbval[0] = np.array([(1, (2,3))], dtype=recordwitharray)[0]
```
stuartarchibald, line 1552:
```suggestion
        nbval[0] = np.array([(1, (2, 3))], dtype=recordwitharray)[0]
```

numba/tests/test_record_dtype.py
```diff
@@ -1371,23 +1368,260 @@ def foo(rec, flag=None):
...
+        for pyfunc in [record_read_array0, record_read_array0_alt]:
+            arr_expected = pyfunc(nbval)
+            cfunc = self.get_cfunc(pyfunc, (typeof(nbval),))
+            arr_res = cfunc(nbval)
+            np.testing.assert_equal(arr_expected, arr_res)
+
+    def test_slice_2d_array(self):
+        # test slicing the nestedarray inside a record
+        nbarr = np.recarray(2, dtype=recordwith2darray)
+        nbarr[0] = np.array([(1, ((1,2),(4,5),(2,3)))],
```
stuartarchibald, line 1562:
```suggestion
        nbarr[0] = np.array([(1, ((1, 2), (4, 5), (2, 3)))],
```

numba/tests/test_record_dtype.py
```diff
@@ -1371,23 +1368,260 @@ def foo(rec, flag=None):
...
+            ty = typeof(arg)
+            arr_expected = pyfunc(arg)
+            cfunc = self.get_cfunc(pyfunc, (ty,))
+            arr_res = cfunc(arg)
+            np.testing.assert_equal(arr_res, arr_expected)
+
+    def test_shape(self):
+        # test getting the shape of a nestedarray inside a record
+        nbarr = np.recarray(2, dtype=recordwith2darray)
+        nbarr[0] = np.array([(1, ((1,2),(4,5),(2,3)))],
```
stuartarchibald, line 1576:
```suggestion
        nbarr[0] = np.array([(1, ((1, 2), (4, 5), (2, 3)))],
```

numba/tests/test_record_dtype.py
```diff
@@ -1371,23 +1368,260 @@ def foo(rec, flag=None):
...
+            cfunc = self.get_cfunc(pyfunc, (ty,))
+            arr_res = cfunc(arg)
+            np.testing.assert_equal(arr_res, arr_expected)
+
+    def test_broadcast_slice(self):
+        nbarr = np.recarray(2, dtype=recordwith2darray)
+        nbarr[0] = np.array([(1, ((1,2),(4,5),(2,3)))],
+                            dtype=recordwith2darray)[0]
+        nbarr[1] = np.array([(10, ((10,20),(40,50),(20,30)))],
+                            dtype=recordwith2darray)[0]
```
stuartarchibald, line 1609:
```suggestion
        nbarr[0] = np.array([(1, ((1, 2), (4, 5), (2, 3)))],
                            dtype=recordwith2darray)[0]
        nbarr[1] = np.array([(10, ((10, 20), (40, 50), (20, 30)))],
                            dtype=recordwith2darray)[0]
```

numba/tests/test_record_dtype.py
```diff
@@ -1371,23 +1368,260 @@ def foo(rec, flag=None):
...
+        ty = typeof(arg)
+        arr_expected = pyfunc(arg)
+        cfunc = self.get_cfunc(pyfunc, (ty,))
+        arr_res = cfunc(arg)
+        np.testing.assert_equal(arr_res, arr_expected)
+
+    def test_corner_slice(self):
+        # testing corner cases while slicing nested arrays
+        nbarr = np.recarray((1, 2, 3, 5, 7, 13, 17), dtype=recordwith4darray,
+                            order='F')
```
stuartarchibald, line 1590: Thanks for adding these somewhat unusual test cases as per our discussion at the public meeting, much appreciated.

stuartarchibald (commented):
Thanks for working on this @luk-f-a and @gmarkall, great to see this functionality enabled. I've largely concentrated on reviewing the CUDA changes to this change set on the basis that @gmarkall has already reviewed the compiler and CPU target changes, though I have skimmed these changes too. I've a couple of very minor queries (noted inline) and have highlighted a few minor stylistic/typo things too, otherwise the PR looks good, looking forward to getting this merged! Thanks again!

numba/tests/test_record_dtype.py
gmarkall: I'm happy with it for now.

luk-f-a: fwiw, this code comes from `master`. it shows in the diff because I moved it.

numba/np/arrayobj.py
gmarkall: I think we shouldn't pack the array because `populate_array` will do it for us. Further, I think we needn't have two slightly different calls to `populate_array` - instead we can just determine the correct shape, strides, and datasize for each case in the branches then populate the array after, if we do something like:

```diff
diff --git a/numba/np/arrayobj.py b/numba/np/arrayobj.py
index 7672a2901..97b99d531 100644
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -2511,30 +2511,29 @@ def array_record_getattr(context, builder, typ, value, attr):
     newdataptr = cgutils.pointer_add(
         builder, array.data, constoffset,  return_type=rary.data.type,
     )
+
     if isinstance(dtype, types.NestedArray):
-        datasize = context.get_abi_sizeof(context.get_data_type(dtype.dtype))
-        # new strides = recarray strides + strides of the inner nestedarray
-        strides = cgutils.unpack_tuple(builder, array.strides, typ.ndim)
-        strides += [context.get_constant(types.intp, i) for i in dtype.strides]
-        # new shape = recarray shape + inner dimension from nestedarray
+        # New shape = recarray shape + inner dimension from nestedarray
         shape = cgutils.unpack_tuple(builder, array.shape, typ.ndim)
         shape += [context.get_constant(types.intp, i) for i in dtype.shape]
-        populate_array(rary,
-                       data=newdataptr,
-                       shape=cgutils.pack_array(builder, shape),
-                       strides=strides,
-                       itemsize=context.get_constant(types.intp, datasize),
-                       meminfo=array.meminfo,
-                       parent=array.parent)
+        # New strides = recarray strides + strides of the inner nestedarray
+        strides = cgutils.unpack_tuple(builder, array.strides, typ.ndim)
+        strides += [context.get_constant(types.intp, i) for i in dtype.strides]
+        # New datasize = size of elements of the nestedarray
+        datasize = context.get_abi_sizeof(context.get_data_type(dtype.dtype))
     else:
+        # New shape, strides, and datasize match the underlying array
+        shape = array.shape
+        strides = array.strides
         datasize = context.get_abi_sizeof(context.get_data_type(dtype))
-        populate_array(rary,
-                       data=newdataptr,
-                       shape=array.shape,
-                       strides=array.strides,
-                       itemsize=context.get_constant(types.intp, datasize),
-                       meminfo=array.meminfo,
-                       parent=array.parent)
+
+    populate_array(rary,
+                   data=newdataptr,
+                   shape=shape,
+                   strides=strides,
+                   itemsize=context.get_constant(types.intp, datasize),
+                   meminfo=array.meminfo,
+                   parent=array.parent)
     res = rary._getvalue()
     return impl_ret_borrowed(context, builder, resty, res)
 
```

luk-f-a: you're right, I've now changed it. I probably copied that pattern from another function without realizing it was not necessary to do it here.

luk-f-a: @gmarkall your comment was somehow not showing on screen when I replied to Stuart. so my previous reply was to Stuart's comment. I'll look at yours now.

numba/tests/test_record_dtype.py
luk-f-a: improvement here https://github.com/numba/numba/pull/7359/commits/be6369a07bc83d77de34d1d3425248cfe78c08e0

numba/np/arrayobj.py
luk-f-a: improvement here https://github.com/numba/numba/pull/7359/commits/be6369a07bc83d77de34d1d3425248cfe78c08e0

gmarkall:
gpuci run tests

numba/np/arrayobj.py
```diff
@@ -2507,11 +2511,25 @@ def array_record_getattr(context, builder, typ, value, attr):
     newdataptr = cgutils.pointer_add(
         builder, array.data, constoffset,  return_type=rary.data.type,
     )
-    datasize = context.get_abi_sizeof(context.get_data_type(dtype))
+    if isinstance(dtype, types.NestedArray):
+        datasize = context.get_abi_sizeof(context.get_data_type(dtype.dtype))
```
stuartarchibald, line 2515:
```suggestion
```
This is calculated again below (after the strides calc)?

stuartarchibald:
Thanks for the updates @luk-f-a and @gmarkall, I think it's just https://github.com/numba/numba/pull/7359#discussion_r762921120 to resolve else this looks good. I don't see a need put this through the build farm again given what's changed since the last run and `gpuci` is passing (unless you feel differently @gmarkall?).

numba/np/arrayobj.py
gmarkall: Whoops, I missed that. Are you happy with this @luk-f-a ?

gmarkall:
@stuartarchibald After the last suggestion is implemented I'll run gpuci again and I'll be happy with it if it passes.

gmarkall:
gpuci run tests

stuartarchibald (approved):
@luk-f-a @gmarkall many thanks to you both for all your work on this PR, both in implementing and reviewing. It's great to see this feature working. Patch looks good!

gmarkall:
gpuci run tests

gmarkall (approved):
Many thanks @luk-f-a, this is now looking good!

dlee992:
Ah, this's a nice feature for `NestedArray`! 
But users (like me) could want numba to support more operations for `NestedArray` just as `Array`. I will try to implement some basic broadcast operations between `NestedArray` and primitive types, e.g., `int`, `float`, or a constant as a start for further support.

I also noticed `NestedArray` doesn't have NRT support https://github.com/numba/numba/issues/7526, i.e., valid meminfo management? I guess for simple operations of `NestedArray`, this lack could be no harm (if we don't copy the array?)

PR #7360, opened 2021-09-02
CUDA: Fix potential leaks when initialization fails

gmarkall:
When CUDA driver initialization fails, the driver singleton object persists holding on to an exception object that references calling
frames and potentially other objects. This can create a leak in the case where modules that attempt to initialize the driver and then fail are created and destroyed.

This change rectifies the issue by holding on to a string describing the error instead of the exception object. There are a couple of small related changes:

- `initialize()` is changed to `ensure_initialized()`, and the caller no longer needs to check whether it should be called - it can always call it when it needs to ensure that the driver is initialized.
- `cuda.cuda_error()` returns the error string instead of an exception object. Constructing an exception object here just to maintain the original behavior seems a bit convoluted; it's likely that any code using `cuda_error()` is checking whether its return value is `None` rather than looking for a specific instance of an exception class to see if an exception occurred.

Some initialization tests are added - for the failing cases we need to run in a subprocess to avoid interfering with the initialization of the driver in the process in which we're actually running tests. The failure of `cuInit(0)` is accomplished by a slightly unorthodox patching of `driver.cuInit`, which is needed because driver functions are added to the `Driver` object on-demand, so there is nothing for `mock.patch.object()` to replace at the time we need to set up the mock.

### DIFF
--- a/numba/cuda/cudadrv/driver.py
+++ b/numba/cuda/cudadrv/driver.py
@@ -223,7 +223,10 @@ class Driver(object):
             self.is_initialized = True
             self.initialization_error = e
 
-    def initialize(self):
+    def ensure_initialized(self):
+        if self.is_initialized:
+            return
+
         # lazily initialize logger
         global _logger
         _logger = make_logger()
@@ -233,8 +236,9 @@ class Driver(object):
             _logger.info('init')
             self.cuInit(0)
         except CudaAPIError as e:
-            self.initialization_error = e
-            raise CudaSupportError("Error at driver init: \n%s:" % e)
+            description = f"{e.msg} ({e.code})"
+            self.initialization_error = description
+            raise CudaSupportError(f"Error at driver init: {description}")
         else:
             self.pid = _getpid()
 
@@ -259,8 +263,7 @@ class Driver(object):
 
     @property
     def is_available(self):
-        if not self.is_initialized:
-            self.initialize()
+        self.ensure_initialized()
         return self.initialization_error is None
 
     def __getattr__(self, fname):
@@ -272,9 +275,7 @@ class Driver(object):
         restype = proto[0]
         argtypes = proto[1:]
 
-        # Initialize driver
-        if not self.is_initialized:
-            self.initialize()
+        self.ensure_initialized()
 
         if self.initialization_error is not None:
             raise CudaSupportError("Error at driver init: \n%s:" %
--- a/numba/cuda/cudadrv/driver.py
+++ b/numba/cuda/cudadrv/driver.py
@@ -221,7 +221,7 @@ class Driver(object):
             self.lib = find_driver()
         except CudaSupportError as e:
             self.is_initialized = True
-            self.initialization_error = e
+            self.initialization_error = e.msg
 
     def ensure_initialized(self):
         if self.is_initialized:
--- a/numba/cuda/tests/cudadrv/test_init.py
+++ b/numba/cuda/tests/cudadrv/test_init.py
@@ -1,4 +1,5 @@
 import multiprocessing as mp
+import os
 
 from numba import cuda
 from numba.cuda.cudadrv.driver import CudaAPIError, driver
@@ -45,8 +46,39 @@ def initialization_error_test(result_queue):
         cuda.device_array(1)
     except CudaSupportError:
         success = True
-        msg = cuda.cuda_error()
 
+    msg = cuda.cuda_error()
+    result_queue.put((success, msg))
+
+
+# For testing the path where Driver.__init__() catches a CudaSupportError
+def cuda_disabled_test(result_queue):
+    success = False
+    msg = None
+
+    try:
+        # A CUDA operation that forces initialization of the device
+        cuda.device_array(1)
+    except CudaSupportError as e:
+        success = True
+        msg = e.msg
+
+    result_queue.put((success, msg))
+
+
+# Similar to cuda_disabled_test, but checks cuda.cuda_error() instead of the
+# exception raised on initialization
+def cuda_disabled_error_test(result_queue):
+    success = False
+    msg = None
+
+    try:
+        # A CUDA operation that forces initialization of the device
+        cuda.device_array(1)
+    except CudaSupportError:
+        success = True
+
+    msg = cuda.cuda_error()
     result_queue.put((success, msg))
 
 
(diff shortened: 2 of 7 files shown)


gmarkall:
@stuartarchibald @sklam I've tested this with cuDF:

* In the normal case (`cuInit(0)` succeeds), all is well.
* In the failing case (`cuInit(0)` fails) cuDF errors on all tests before it can even reach Numba - every test errors in an RMM error instead.

So I believe this change will have no effect on cuDF (and the other RAPIDS libraries).

gmarkall:
@philippjfr Are you able to verify that this resolves holoviz/panel#2640?

philippjfr:
Thanks for checking with me, what's the best way to install the PR? I've never installed numba from source, can I just check it out and `pip install -e .`?

gmarkall:
> Thanks for checking with me, what's the best way to install the PR? I've never installed numba from source, can I just check it out and `pip install -e .`?

That might work - you will need the latest dev version of llvmlite though, and I'm not sure if that will be a problem with pip - I usually install the dev version of llvmlite with:

```
conda install numba/label/dev::llvmlite
```

If you're in a conda env, then I would guess you could do the above then run `pip install --no-deps -e .`.

numba/cuda/cudadrv/driver.py
```diff
@@ -223,7 +223,10 @@ def __init__(self):
             self.is_initialized = True
             self.initialization_error = e
```
philippjfr, line 224: Still seeing the same issue because of this line (when I comment it out the issue goes away).

philippjfr: Presumably `.initialize()` is never exercised in the scenario I'm testing (which is simply to `import datashader` on a machine without CUDA)

gmarkall: Argh! Sorry I missed this, will fix up tomorrow.

gmarkall: @philippjfr This should now be resolved - could you let me know if it's fixed up all the issues in your test please?

philippjfr: Can confirm, fixed now.

gmarkall: Many thanks for re-testing!

numba/cuda/cudadrv/driver.py
```diff
@@ -233,8 +236,9 @@ def initialize(self):
             _logger.info('init')
             self.cuInit(0)
         except CudaAPIError as e:
-            self.initialization_error = e
-            raise CudaSupportError("Error at driver init: \n%s:" % e)
+            description = f"{e.msg} ({e.code})"
```
stuartarchibald, line 239: Could this just be `str(e)`?

numba/cuda/cudadrv/driver.py
```diff
@@ -221,9 +221,12 @@ def __init__(self):
             self.lib = find_driver()
         except CudaSupportError as e:
             self.is_initialized = True
-            self.initialization_error = e
+            self.initialization_error = e.msg
+
+    def ensure_initialized(self):
```
stuartarchibald, line 226: As noted out of band, am not convinced this (or the init) is threadsafe in its current state, seems like the `is_initialized` bit is set too eagerly. Suggest fixing for the next release. xref https://github.com/numba/numba/issues/7387

numba/cuda/tests/cudadrv/test_init.py
```diff
@@ -0,0 +1,139 @@
...
+
+@skip_on_cudasim('CUDA Simulator does not initialize driver')
+class TestInit(CUDATestCase):
+    def _test_init_failure(self, target, expected):
+        # Run the initialization failure test in a separate subprocess
+        ctx = mp.get_context('spawn')
+        result_queue = ctx.Queue()
+        proc = ctx.Process(target=target, args=(result_queue,))
+        proc.start()
+        proc.join()
```
stuartarchibald, line 93:
```suggestion
        proc.join(30) # should complete within 30s
```
Perhaps add a timeout just in case something gets stuck so the test suite doesn't hang?

numba/cuda/tests/cudadrv/test_init.py
```diff
@@ -0,0 +1,139 @@
...
+        success, msg = result_queue.get()
+
+        # Ensure the child process raised an except during initialization
```
stuartarchibald, line 96:
```suggestion
        # Ensure the child process raised an exception during initialization
```

stuartarchibald (commented):
Thanks for the patch, looks good and on inspection should stop the issue reported regarding holding reference to objects via the numba cuda driver singleton->exception->backtrace->frame path. There's a few minor things to look at but otherwise is ready for testing on the buildfarm.

stuartarchibald:
CC @sklam, did you perhaps want to take a look at this patch too? I helped debug the original issue etc. so some additional eyes on it might be good especially given it's scheduled the 0.54.1 patch release. Thanks!

numba/cuda/cudadrv/driver.py
gmarkall: I suppose so, but I thought writing the message first was a tiny bit nicer for the user, e.g.

```
CUDA_ERROR_UNKNOWN (999)
```

vs

```
[999] CUDA_ERROR_UNKNOWN
```

since the user tends to think in terms of the names of the errors rather than the error code numbers (which are an implementation detail to some extent).

gmarkall:
@stuartarchibald Many thanks for the review - comments now addressed.

numba/cuda/cudadrv/driver.py
stuartarchibald: Agree, let's leave it.

stuartarchibald:
> @stuartarchibald Many thanks for the review - comments now addressed.

Thanks, looks good.

stuartarchibald (approved):
Thanks for the patch!

philippjfr:
Appreciate the quick turnaround, thanks everyone!

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_94`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_94`.

Passed.

stuartarchibald:
> Appreciate the quick turnaround, thanks everyone!

@philippjfr No problem, thanks for testing it! If you still have the set up available, is there any chance you could please test 17e112b against the original problem just to make sure the patch that will get merged does indeed still fix it!? Many thanks.

Issue #7314, opened 2021-08-18
cuda.local.array produces TypingError even though the shape argument is defined at the top of the script (i.e. globally)


sgbaird:
## Example
Here is an example that works, but only by hard-coding (i.e. typing directly into the input function using the keyboard keys `2` and `0`) the number of columns of the concatenated array.

concatenate-gpu-test.py
```python
# -*- coding: utf-8 -*-
"""
Created on Tue Aug 17 23:16:43 2021.

@author: sterg
"""

import numpy as np
from numba import cuda
from numba.types import float32, int32

bits = 32
nb_float_type = float32
np_float_type = np.float32

# sample data
a = np.random.rand(100, 10)
b = np.random.rand(100, 10)

# I want to be able to programmatically define the total number of columns for later array allocation
cols = a.shape[1]
tot_cols = np.int32(a.shape[1] + b.shape[1])


@cuda.jit(device=True)
def vec_concatenate(vec, vec2, vec3):

    # number of elements in each vector
    n = len(vec)
    n2 = len(vec2)

    # populate vec3 with element-wise concatenation
    for i in range(n):
        vec3[i] = vec[i]
    for i in range(n2):
        vec3[i + n] = vec2[i]


@cuda.jit("void(float{0}[:, :], float{0}[:, :], float{0}[:, :])".format(bits))
def foo_stuff(mat, mat2, out):

    i = cuda.grid(1)

    # unpack vectors
    vec = mat[i]
    vec2 = mat2[i]

    # note the `20` is hardcoded into the shape argument
    vec3 = cuda.local.array(20, float32)
    vec_concatenate(vec, vec2, vec3)
    for j in range(tot_cols):
        out[i, j] = vec3[j]


def mat_concatenate(mat, mat2):

    # black and grid dimensions
    rows = mat.shape[0]
    block_dim = 16
    grid_dim = np.int32(rows / block_dim + 1)

    # output initialization
    cuda_out = cuda.device_array((rows, tot_cols), np.float32)

    # CUDA setup
    stream = cuda.stream()

    # copying data to GPU
    cuda_mat = cuda.to_device(np.asarray(mat, dtype=np_float_type), stream=stream)
    cuda_mat2 = cuda.to_device(np.asarray(mat2, dtype=np_float_type), stream=stream)
    foo_stuff[grid_dim, block_dim](cuda_mat, cuda_mat2, cuda_out)

    # copy it to the CPU
    out = cuda_out.copy_to_host(stream=stream)

    return out


[...]

PR #7363, opened 2021-09-03
Update cuda.local.array to clarify "simple constant expression" (e.g. no NumPy ints)

sgbaird:
Update the docs to clarify that NumPy ints aren't supported when defined as global variables and passed into cuda.local.array.

#7314

### DIFF
--- a/docs/source/cuda/memory.rst
+++ b/docs/source/cuda/memory.rst
@@ -125,8 +125,17 @@ traditional dynamic memory management.
    This function must be called on the device (i.e. from a kernel or
    device function). *shape* is either an integer or a tuple of integers
    representing the array's dimensions and must be a simple constant
-   expression. *type* is a :ref:`Numba type <numba-types>` of the elements
-   needing to be stored in the array.
+   expression. A "simple constant expression" includes, but is not limited to:
+   1. A literal (e.g. ``10``)
+   2. A local variable whose right-hand side is a literal or a simple constant
+   expression (e.g. ``shape``, where ``shape`` is defined earlier in the function as 
+   `shape = 10``)
+   3. A global variable that is defined in the jitted function's globals by the time of 
+   compilation (e.g. ``shape``, where ``shape`` is defined using any expression at
+   global scope).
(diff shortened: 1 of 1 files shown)


sgbaird:
The checks are passing now.

esc (approved):
I looked at #7314 and then at this. It appears to be fine.

Thank you for the patch!

esc (dismissed):
After OOB discussion with @gmarkall -- it seems a few more things could be added. Comments to follow soon, sorry for the confusion.

gmarkall (changes requested):
Thanks for the PR! I think the suggested wording is a bit misleading, in that it doesn't quite characterize exactly what a "simple constant expression" is. (Note that it's actually quite hard to define exactly due to the behavior of constant propagation, etc).

I would suggest stating that a simple constant expression includes, but is not limited to:

* A literal (e.g. `10`)
* A local variable whose right-hand side is a literal or a simple constant expression (e.g. `shape`, where `shape` is defined earlier in the function as `shape = 10`)
* A global variable that is defined in the jitted function's globals by the time of compilation (e.g. `shape`, where `shape` is defined using any expression at global scope).

where the definition results in a Python `int` (i.e. not a NumPy scalar or other scalar / integer-like type).

gmarkall:
@sgbaird Did you want to revise this shortly to try to get it into the 0.55 release in early November? Or shall we bump the target to 0.56 to give a bit more time?

stuartarchibald:
Just checking in on the state of this PR. @sgbaird was wondering if you have time to address the changes above, if not, it's no problem, one of the core developers can adopt the patch as needed. Many thanks.

gmarkall (changes requested):
Many thanks for the update - I think the wording is fine, but there's some issues with the rendering - I think the intention was for the changes to render as a numbered list, but they come out all as one paragraph. Also, there's a backtick missing which causes a problem rendering some of the code as code. Example shown:

image

sgbaird:
@gmarkall good catch. I think this should be resolved now

image

image

gmarkall (approved):
Looks good. Many thanks for your patience on this one!

PR #7364, opened 2021-09-03
Removes an instance of signed integer overflow undefined behaviour.

folded:
The UB is benign but may interact badly with optimization and is
flagged by UB sanitizers.

* Values that are used as array indices become size_t.
* TCCMap::hash can be private, as the usage of the hash value is
  an internal implementation detail.
* Use std::hash for hashing the elements of TCCRecord. Double hash
  the first element so that {a,b} hashes differently to {b,a}


gmarkall:
@folded Many thanks for the PR! Are there any other instances of UB or other issues flagged by static analysis that you're presently aware of, or is the issue addressed in this PR the only one?

gmarkall:
Also, do you have guidelines for running the analysis to check what the analysis output is before / after this PR?

folded:
This was brought to my attention by another team whose tests were failing when run under asan. It's the only case that I'm aware of currently, but I haven't tried running the numba test suite with asan yet. I would pessimistically assume that there are other cases.

folded:
I'm not sure how you would reproduce this outside of our internal environment. I've landed the patch internally, and it's also possible to inspect the patch to see that it removes the multiply. The error that we intially saw was:

`numba/core/typeconv/typeconv.cpp:19:25: runtime error: signed integer overflow: 3430005 * 1000003 cannot be represented in type 'int'`

Which overflows 32 bits (but not 64)

gmarkall:
Many thanks for the details! I'll add this to the review queue.

folded:
Found one more trivial instance of UB and filed an issue for one that is less so. After that there's an unaligned access from numba (which seems intended).

stuartarchibald:
@gmarkall @sklam this patch looks good to me. The key space is now `size_t` which is still massive in comparison to what is ever going to be feasible in a program. The use of `std::hash` seems like a reasonable and maintenance free alternative to the existing python-hash based alg. Any thoughts/objections to merging this?

gmarkall (approved):
This looks good to me.

stuartarchibald (approved):

stuartarchibald:
> This looks good to me.

Thanks for taking a look @gmarkall .

numba/core/typeconv/typeconv.cpp
```diff
@@ -13,32 +13,26 @@ TCCMap::TCCMap()
...
 }
 
-unsigned int TCCMap::hash(const TypePair &key) const {
-    const int mult = 1000003;
-    int x = 0x345678;
-    x = (x ^ key.first) * mult;
-    x = (x ^ key.second);
-    return x;
+size_t TCCMap::hash(const TypePair &key) const {
+    return std::hash<size_t>()(std::hash<Type>()(key.first)) ^
```
gmarkall, line 17: Why do we hash (hash(first) ^ hash(second))? (i.e. why the "outer" hash?)

folded: because ^ is commutative, this is done to avoid the hash of (a,b) being the same as the hash of (b,a). Originally that was done by the mult param, but it's better just to double hash.

gmarkall: Understood - many thanks!

gmarkall:
@stuartarchibald The double-hash question seems to be resolved now (as in comment on the line of code).

@stuartarchibald / @esc Could this have a full buildfarm run please?

esc:
> `Build numba_smoketest_cpu_yaml_101 has started`

Was green. 🟢

PR #7384, opened 2021-09-08
Add wrapper to cuda.jit

### LINKED ISSUE #5902: cuda.jit does NOT preserve the original __doc__ and __module__ of the decorated function

I noticed this issue when using [pdoc3](https://pdoc3.github.io/pdoc/) to generate documentation. `pdoc3` works by extracting the `__doc__` attribute. Consider the following example.
```python
from numba import cuda, jit, njit

@jit
def add1(a: int, b: int) -> int:
    """Add two integers

    Args:
        a (int): one integer
        b (int): the other integer

    Returns:
        int: the sum
    """
    return a + b


@njit
def add2(a: int, b: int) -> int:
    """Add two integers

    Args:
        a (int): one integer
        b (int): the other integer

    Returns:
        int: the sum
    """
    return a + b


@cuda.jit
def add3(a: int, b: int) -> int:
    """Add two integers

    Args:
        a (int): one integer
        b (int): the other integer

    Returns:
        int: the sum
    """
    return a + b


print('add1\n', add1.__doc__)
print('add2\n', add2.__doc__)
print('add3\n', add3.__doc__)
```

Both `add1` and `add2` keep the `__doc__` of the original function. However, `add3` does not.
```
add1
 Add two integers

    Args:
        a (int): one integer
        b (int): the other integer

    Returns:
        int: the sum

add2
 Add two integers

    Args:
        a (int): one integer
        b (int): the other integer

    Returns:
        int: the sum

add3

    CUDA Kernel object. When called, the kernel object will specialize itself
    for the given arguments (if no suitable specialized version already exists)
    & compute capability, and launch on the device associated with the current
    context.

    Kernel objects are not to be constructed by the user, but instead are
```
Is it quick or possible to fix this issue? Thank you.

MarcinBaranek:
## Issue #5902
I added an update wrapper to cuda.jit to fix issue # 5902

### DIFF
--- a/numba/cuda/compiler.py
+++ b/numba/cuda/compiler.py
@@ -937,6 +937,8 @@ class Dispatcher(_dispatcher.Dispatcher, serialize.ReduceMixin):
                                         defargs, can_fallback, has_stararg,
                                         exact_match_required)
 
+        functools.update_wrapper(self, py_func)
+
         if sigs:
             if len(sigs) > 1:
                 raise TypeError("Only one signature supported at present")
--- a/numba/tests/test_decorated_functions.py
+++ b/numba/tests/test_decorated_functions.py
@@ -1,0 +1,70 @@
+import unittest
+
+from numba import cuda, jit, njit
+
+
+@jit
+def add1(a: int, b: int) -> int:
+    """Add two integers
+    Args:
+        a (int): one integer
+        b (int): the other integer
+    Returns:
+        int: the sum
+    """
+    return a + b
+
+
+@njit
+def add2(a: int, b: int) -> int:
+    """Add two integers
+    Args:
+        a (int): one integer
+        b (int): the other integer
+    Returns:
+        int: the sum
+    """
+    return a + b
+
+
+@cuda.jit
+def add3(a: int, b: int) -> int:
+    """Add two integers
+    Args:
+        a (int): one integer
+        b (int): the other integer
+    Returns:
+        int: the sum
+    """
+    return a + b
+
+
+DOC = """Add two integers
+    Args:
+        a (int): one integer
+        b (int): the other integer
+    Returns:
+        int: the sum
+    """
+
+
+class TestCooperativeGroups(unittest.TestCase):
+    def test_doc_decorated_functions(self):
+        """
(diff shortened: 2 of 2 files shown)


gmarkall (changes requested):
Many thanks for the PR! From a quick skim, I have a couple of requests:

- Could you fix the flake8 error please? It seems there's a whitespace error introduced.
- Could you add a test of the docstrings for a decorated function please?

gmarkall (changes requested):
Many thanks for adding the tests. It looks like some attention may also need to be given for this in cudasim.

When you push future commits, could you please add commits on top of those already reviewed, and not force push? Force-pushing tends to mess up the review history which makes it more difficult to pick up the review again. Many thanks in advance.

gmarkall:
@MarcinBaranek Do you plan to continue with this PR? If so, could you please:

- Fix the issue causing the tests to fail when run under the CUDA simulator (you can replicate this locally by running with the env var `NUMBA_ENABLE_CUDASIM=1` set)
- Move the tests somewhere appropriate - all CUDA tests should be under the `numba.cuda.tests` module - I'd suggest putting the tests in `numba.cuda.tests.cudapy.test_dispatcher`, since they're testing a property of the CUDA `Dispatcher` objects.
- Rename the test to something more appropriate (`TestCooperativeGroups` doesn't reflect the nature of the test).

If you don't plan to continue with the PR, could you please close it?

Either way - many thanks for your efforts so far, and in anticipation of your response.

gmarkall:
Bumping to 0.56 as 0.55 is close.

stuartarchibald:
Just checking in on the state of this PR. @MarcinBaranek was wondering if you have time to address the changes above, if not, it's no problem, one of the core developers can adopt the patch as needed. Many thanks.

sklam:
This is one of four PRs that are affected by our renaming of the default branch and Github failed to automatic update the target branch. It is now in a state that cannot be reopened. OP please open a new PR. Sorry for the inconvenience.

gmarkall:
This should be resolved by #7815, so it should be OK for this to remain closed.

Issue #7356, opened 2021-08-31
Numba 0.54.0: unexpected assignment behavior within `objmode` context.


BenjaminChoou:
## Reporting a bug

- [Y] I have tried using the latest released version of Numba (most recent is
 visible in the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [Y] I have included a self contained code sample to reproduce the problem.

```
import numba
from numba import objmode


class Test(object):
    a = None

@numba.jit(nopython=True)
def test():
    with objmode():
        Test.a = 100
        print(Test.a)
    
    with objmode():
        print(Test.a)


test()
```

The above code runs as expected if Numba <= 0.53.1. The expected outputs are:
```
100
100
```
However, after update Numba to 0.54.0, the outputs are:
```
100
None
```
, which is not consistent with pure python.

stuartarchibald:
Thanks for the report, I can reproduce.

`git bisect` suggests:

```
The first bad commit could be any of:
```
b9a7fc148ad298e60b0b1197ef75eb04c4109b2c
e13d80c9af2238ae2d510b5669272debce7c3ef3
033452bc2b16e06121a3fa78dda592eb17fba32d
58c4e3282607199ad3e260919c3a4849b8088153
34db404262b50c1b90ef12108da4d4b2fa092900
```
We cannot bisect more!
```
Which are all #6977.

CC @sklam

sklam:
I found out that `cloudpickle` is serializing `Test` when it is not importable. A class (e.g. `Test`) is usually importable when it is defined as a global of a module. However, the failing case is when the class is defined in the `__main__` module. The cloudpickle code is rejecting to import from `__main__` even though it appears to be possible.

Since `Test` is serialized allow with the code in `objmode()`, it is creating a fresh `Test` class everytime.

sklam:
The problem is pretty much down to this if-statement in cloudpickle: https://github.com/cloudpipe/cloudpickle/blob/343da119685f622da2d1658ef7b3e2516a01817f/cloudpickle/cloudpickle.py#L300-L301
Why does cloudpickle reject handling `__main__`? Commenting it out appears to work for this issue.

sklam:
> Why does cloudpickle reject handling __main__? Commenting it out appears to work for this issue.

answer: https://github.com/cloudpipe/cloudpickle/issues/131

PR #7388, opened 2021-09-08
Patch cloudpickle to not reset dynamic class each time it is unpickled

sklam:
Fixes a problem with cloudpickle when dynamic class is unpickled and the class is reset to the state when it is pickled.

A short reproducer for the cloudpickle problem is available here:

```python
# Tested with cloudpickle 1.6.0
from cloudpickle import dumps, loads


class Klass:
    classvar = None

def mutator():
    Klass.classvar = 100

def check():
    print("checking....")
    print(f"   Klass.classvar [{hex(id(Klass))}] = {Klass.classvar}")


def failing_case():
    print("Klass", hex(id(Klass)))
    saved = dumps(Klass)
    mutator()
    check()
    loads(saved)
    check()
    loads(saved)
    check()



if __name__ == '__main__':
    failing_case()
```

Note, cloudpickle treats any class created in `__main__` as dynamic.

The patch makes cloudpickle remember whether a dynamic class is being reused. Reuse can come from the class being unpickled in the same process as it being pickled; or, a previously unpickled class is still around.

Fixes https://github.com/numba/numba/issues/7356

### DIFF
--- a/numba/tests/test_serialize.py
+++ b/numba/tests/test_serialize.py
@@ -5,6 +5,7 @@ import runpy
 import subprocess
 import sys
 import unittest
+from multiprocessing import get_context
 
 from numba.core.errors import TypingError
 from numba.tests.support import TestCase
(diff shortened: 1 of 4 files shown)


numba/tests/test_serialize.py
```diff
@@ -205,5 +208,99 @@ def test_numba_unpickle(self):
         self.assertIs(got1, got2)
 
 
+
+class TestCloudPickleIssues(TestCase):
+    """This testcase includes all issues specific to cloudpickle implementation.
```
stuartarchibald, line 213:
```suggestion
    """This test case includes issues specific to the cloudpickle implementation.
```

numba/tests/test_serialize.py
```diff
@@ -205,5 +208,99 @@ def test_numba_unpickle(self):
         self.assertIs(got1, got2)
 
 
+
+class TestCloudPickleIssues(TestCase):
```
stuartarchibald, line 212: Maybe this ought to also include the original reproducer as a test? https://github.com/numba/numba/issues/7356#issue-983961914

stuartarchibald (commented):
Thanks for the patch, a couple of minor things to resolve else looks good.

numba/tests/test_serialize.py
sklam: Good point. Test added in https://github.com/numba/numba/pull/7388/commits/8f6ffdcafbfdbb5317605950c4f57f43dae83111

stuartarchibald (approved):

Issue #6550, opened 2020-12-05
Array comprehension fails against multidimensional array with nopython and typed List


dizcza:
## Reporting a bug

I'm reporting a regression bug for https://github.com/numba/numba/issues/3659 when `@njit` and typed lists are used.

- [x] I have tried using the latest released version of Numba (most recent is
 visible in the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [x] I have included a self contained code sample to reproduce the problem.
  i.e. it's possible to run as 'python bug.py'.

```python
import numba
import numpy as np
from numba.typed import List


@numba.njit
def main(a):
    return np.array([x[0] for x in a])


a = [np.array([1,2]), np.array([3, 4, 5])]
a = List(a)
print(main(a))
```

```
Traceback (most recent call last):
  File "/home/ulianych/PycharmProjects/elephant/elephant/temp2.py", line 13, in <module>
    print(main(a))
  File "/home/ulianych/anaconda3/envs/elephant/lib/python3.8/site-packages/numba/core/dispatcher.py", line 414, in _compile_for_args
    error_rewrite(e, 'typing')
  File "/home/ulianych/anaconda3/envs/elephant/lib/python3.8/site-packages/numba/core/dispatcher.py", line 357, in error_rewrite
    raise e.with_traceback(None)
numba.core.errors.TypingError: Failed in nopython mode pipeline (step: nopython frontend)
No implementation of function Function(<intrinsic range_iter_len>) found for signature:
 
 >>> <unknown function>(iter[ListType[array(int64, 1d, C)]->array(int64, 1d, C)])
 
There are 2 candidate implementations:
   - Of which 2 did not match due to:
   Intrinsic of function '<unknown function>': File: numba/cpython/rangeobj.py: Line 183.
     With argument(s): '(iter[ListType[array(int64, 1d, C)]->array(int64, 1d, C)])':
    No match.

During: resolving callee type: Function(<intrinsic range_iter_len>)
During: typing of call at /home/ulianych/PycharmProjects/elephant/elephant/temp2.py (8)


File "temp2.py", line 8:
def main(a):
    return np.array([x[0] for x in a])
    ^
```

Works fine for ragged lists (not typed).

Numba v0.52.0
Python 3.8.5

stuartarchibald:
Thanks for the report. Think this is a missing feature, the `typed.List` isn't known to the `range_iter_len` intrinsic and so the closure inlining used for list comp raises.

Starter patch to fix:
[...]

PR #7395, opened 2021-09-10
Fix #7394 and #6550 & Added test & improved error message

### LINKED ISSUE #7394: iteration fails in tuple comprehension
Hello,

As discussed on [gitter](https://gitter.im/numba/numba?at=613a2952d206b85e4f8c176e), I found a bug in list comprehension when passing a tuple as input:
```
@numba.njit
def f1_fine(array_tuple):
    vals = []
    for a in array_tuple:
        vals.append(a[0])
    return np.array(vals)

@numba.njit
def f2_fine(array_tuple):
    return np.array([array_tuple[i][0] for i in range(len(array_tuple))])


@numba.njit
def f3_crash(array_tuple):
    return np.array([a[0] for a in array_tuple])

at = (np.ones(4), np.ones(3))

f1_fine(at)
f2_fine(at)
f3_crash(at)
```

While these functions are pretty much the same, `f3` does not compile with error
```
No implementation of function Function(<intrinsic range_iter_len>) found for signature:
 
 >>> <unknown function>(iter(UniTuple(array(float64, 1d, C) x 2)))
 
      Intrinsic of function '<unknown function>': File: numba/cpython/rangeobj.py: Line 183.
        With argument(s): '(iter(UniTuple(array(float64, 1d, C) x 2)))':
       No match.

During: resolving callee type: Function(<intrinsic range_iter_len>)
During: typing of call at <ipython-input-20-4047ac44d0b7> (15)


File "<ipython-input-20-4047ac44d0b7>", line 15:
def f3_crash(array_tuple):
    return np.array([a[0] for a in array_tuple])
    ^
```
numba version `0.53.1`

MegaIng:
also renames `range_iter_len` -> `comprehension_iter_len`

As discussed with @stuartarchibald, this adds the code he suggest in both the issues to (the renamed) `comprehension_iter_len` and gives a better error message. It might be worth considering to move `comprehension_iter_len` to another module, since it is now even less tightly coupled to the `rangeobj`.

### DIFF
--- a/numba/core/inline_closurecall.py
+++ b/numba/core/inline_closurecall.py
@@ -1068,9 +1068,11 @@ def _inline_arraycall(func_ir, cfg, visited, loop, swapped, enable_prange=False,
         # this doesn't work in objmode as it's effectively untyped
         if typed:
             len_func_var = ir.Var(scope, mk_unique_var("len_func"), loc)
-            from numba.cpython.rangeobj import comprehension_iter_len
+            from numba.cpython.rangeobj import length_of_iterator
             stmts.append(_new_definition(func_ir, len_func_var,
-                                         ir.Global('comprehension_iter_len', comprehension_iter_len, loc=loc),
+                                         ir.Global('length_of_iterator',
+                                                   length_of_iterator,
+                                                   loc=loc),
                                          loc))
             size_val = ir.Expr.call(len_func_var, (iter_var,), (), loc=loc)
         else:
--- a/numba/core/inline_closurecall.py
+++ b/numba/core/inline_closurecall.py
@@ -1068,10 +1068,10 @@ def _inline_arraycall(func_ir, cfg, visited, loop, swapped, enable_prange=False,
         # this doesn't work in objmode as it's effectively untyped
         if typed:
             len_func_var = ir.Var(scope, mk_unique_var("len_func"), loc)
-            from numba.cpython.rangeobj import range_iter_len
+            from numba.cpython.rangeobj import comprehension_iter_len
             stmts.append(_new_definition(func_ir, len_func_var,
-                        ir.Global('range_iter_len', range_iter_len, loc=loc),
-                        loc))
+                                         ir.Global('comprehension_iter_len', comprehension_iter_len, loc=loc),
+                                         loc))
             size_val = ir.Expr.call(len_func_var, (iter_var,), (), loc=loc)
         else:
             raise GuardException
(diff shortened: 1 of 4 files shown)


stuartarchibald:
Thanks for the patch @MegaIng.

As this contains my patch from https://github.com/numba/numba/issues/7394#issuecomment-916865579 please could one of you take a look at this @esc @sklam @gmarkall ? Thanks!

stuartarchibald:
> also renames `range_iter_len` -> `comprehension_iter_len`

@MegaIng RE: the above, I've left some comments here on the original issue: https://github.com/numba/numba/issues/7394#issuecomment-916904842

numba/core/inline_closurecall.py
```diff
@@ -1068,10 +1068,10 @@ def is_removed(val, removed):
         # this doesn't work in objmode as it's effectively untyped
         if typed:
             len_func_var = ir.Var(scope, mk_unique_var("len_func"), loc)
-            from numba.cpython.rangeobj import range_iter_len
+            from numba.cpython.rangeobj import comprehension_iter_len
             stmts.append(_new_definition(func_ir, len_func_var,
-                        ir.Global('range_iter_len', range_iter_len, loc=loc),
-                        loc))
+                                         ir.Global('comprehension_iter_len', comprehension_iter_len, loc=loc),
```
sklam, line 1073: flake8---line too long
```suggestion
                                         ir.Global('comprehension_iter_len', 
                                                   comprehension_iter_len, 
                                                   loc=loc),
```

sklam (commented):
Just one flake8 nitpick

numba/core/inline_closurecall.py
MegaIng: Interesting. For me flake8 did not complain about this. But since we are still maybe changing the name I would wait. What do you think about the name?

sklam: Some older files have been added to the exclude-list for flake8 because there are too much problems in it (which is unfortunate).

As for the name, I agree with https://github.com/numba/numba/issues/7394#issuecomment-916904842 and the function is not comprehension specific.

MegaIng: I decided on the more correct `length_of_iterator`, since we don't input the `iterable`, but the actual `iterator` (the `iterable` would be the `range`/`list` itself)

sklam (approved):
Thanks for the patch!

PR #7397, opened 2021-09-13
Add option to catch only Numba `numba.core.errors` derived exceptions.

### LINKED ISSUE #7278: Make non-`NumbaError` exceptions during compilations hard errors
This issue aims to improve error handling. When any instances of `Exception` is raised during compilation, Numba would try to recover by continuing the function selection to find a working implementation. However, this behavior can be confusing. Some errors should be treated as hard error to indicate to the developer that the code is incorrect and Numba should not fallback to other alternatives. For example, a typo error can cause an `AttributeError`, resulting in the implementation to be never used. 

With the recent addition of target-specific overload support, the problem has became more apparent.  For instance, a target specific `overload` implementation could be skipped because an error in the target-selection, but Numba proceeds to fallback to the generic implementation. The fallback machinery can often hide or obfuscated the true error, making debugging harder. 

The proposed solution is to:

- Adopt the model that any non-`NumbaError` derived error are consider as hard errors.
- Initially, allow extensions (`@overloads`, `@lower_bultins`, etc) to opt in to the new model.
- Replace internal implementations
- Deprecate the old model.

stuartarchibald:
As title.

Closes #7278


numba/core/overload_glue.py
```diff
@@ -136,6 +138,24 @@ def body(tyctx):
...
+                    if isinstance(typing, CallableTemplate):
+                        generic_inst = generic()
+                        try:
+                            match_sig = inspect.signature(generic_inst)
+                            match_sig.bind(*ol_args, **ol_kwargs)
+                        except TypeError as e:
+                            # ctor or bind failed, raise, if there's a
+                            # ValueError then there's likely unrecoverable
+                            # problems
+                            raise errors.TypingError(str(e)) from e
```
sklam, line 154: why can't this be part of `CallableTemplate.apply`?

numba/cpython/unsafe/tuple.py
```diff
@@ -36,6 +36,9 @@ def codegen(context, builder, signature, args):
 @intrinsic
 def build_full_slice_tuple(tyctx, sz):
     """Creates a sz-tuple of full slices."""
+    if not isinstance(sz, types.IntegerLiteral):
+        raise errors.RequireLiteralValue(sz)
```
sklam, line 40: Self reminder: need to test this in CUDA

sklam (commented):
Some early comment.

numba/tests/test_parfors.py
```diff
@@ -2409,7 +2411,12 @@ def run_pass(self, state):
                             # point it needs to be a set so e.g. set.difference
                             # can be computed, this therefore creates an error
                             # in the right location.
-                            stmt.races = []
+                            class Broken(list):
+
+                                def difference(self, other):
+                                    raise errors.LoweringError(BROKEN_MSG)
+
+                            stmt.races = Broken()
```
sklam, line 2419: Why is test necessary? When I undo the changes to this test, i can't make it fail either error style.

stuartarchibald: Is the test running or skipping? The parfors tests need the subprocess test harness to actually run, else they "run" but just skip internally. Invocation I'm using is:
```
SUBPROC_TEST=1 NUMBA_CAPTURED_ERRORS='new_style' ./runtests.py numba.tests.test_parfors.TestParforsMisc.test_no_state_change_in_gufunc_lowering_on_error
```
which gives:
```
umba/parfors/parfor_lowering.py", line 989, in _create_gufunc_for_parfor_body
    races = races.difference(set(parfor_redvars))
AttributeError: 'list' object has no attribute 'difference'
```
without this change. What this change does is regularise the failure.

numba/core/overload_glue.py
stuartarchibald: It can and probably should be, so long as it's generic enough?! Done in 810b6bf

numba/core/compiler.py
```diff
@@ -466,6 +466,10 @@ def _compile_core(self):
                     res = e.result
                     break
                 except Exception as e:
+                    if (config.CAPTURED_ERRORS == 'new_style' and not
```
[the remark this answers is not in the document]
stuartarchibald: Good idea, done in fe731a6, there's a couple of places I've left where there's an `if <new style> elif <old style> else assert 0.` which I've left as it seems more natural.

stuartarchibald:
@sklam ba74a06 merges mainline, 3e2f20c resolves issues introduced on mainline and 3a55616 resolve issues introduced in fix ups.

numba/tests/test_parfors.py
sklam: Ok, i see the point on regularizing the failure.

(btw, i am amazed at my inability to type at 7pm. the original comment is incomprehensible 🤦 .)

sklam:
oldstyle smoketest: `numba_smoketest_cpu_yaml_47`

numba/core/typing/templates.py
```diff
@@ -402,6 +402,15 @@ class CallableTemplate(FunctionTemplate):
     def apply(self, args, kws):
         generic = getattr(self, "generic")
         typer = generic()
+        try:
+            match_sig = inspect.signature(typer)
```
sklam, line 406: i don't think it should be catching `TypeError` from `.signature`

```suggestion
        match_sig = inspect.signature(typer)
        try:       
```

numba/core/errors.py
```diff
@@ -783,9 +821,17 @@ def new_error_context(fmt_, *args, **kwargs):
         # Let assertion error pass through for shorter traceback in debugging
         raise
     except Exception as e:
-        newerr = errcls(e).add_context(_format_msg(fmt_, args, kwargs))
-        tb = sys.exc_info()[2] if numba.core.config.FULL_TRACEBACKS else None
-        raise newerr.with_traceback(tb)
+        if numba.core.config.CAPTURED_ERRORS == 'old_style':
```
sklam, line 824: needs a `use_old_style_errors()`

numba/core/errors.py
```diff
@@ -783,9 +821,17 @@ def new_error_context(fmt_, *args, **kwargs):
...
+            newerr = errcls(e).add_context(_format_msg(fmt_, args, kwargs))
+            if numba.core.config.FULL_TRACEBACKS:
+                tb = sys.exc_info()[2]
+            else:
+                tb = None
+            raise newerr.with_traceback(tb)
+        elif numba.core.config.CAPTURED_ERRORS == 'new_style':
```
sklam, line 831: replace with `use_new_style_errors()`

numba/core/typeinfer.py
```diff
@@ -162,15 +162,20 @@ def propagate(self, typeinfer):
...
+                        _logger.debug("captured error", exc_info=e)
+                        msg = ("Internal error at {con}.\n{err}\n"
+                               "Enable logging at debug level for details.")
+                        new_exc = TypingError(
+                            msg.format(con=constraint, err=str(e)),
+                            loc=constraint.loc,
+                            highlighting=False,
+                        )
+                        errors.append(utils.chain_exception(new_exc, e))
+                    elif config.CAPTURED_ERRORS == 'new_style':
```
sklam, line 175: replace with `use_new_style_errors()`

numba/core/typeinfer.py
```diff
@@ -162,15 +162,20 @@ def propagate(self, typeinfer):
...
-                    _logger.debug("captured error", exc_info=e)
-                    msg = ("Internal error at {con}.\n"
-                           "{err}\nEnable logging at debug level for details.")
-                    new_exc = TypingError(
-                        msg.format(con=constraint, err=str(e)),
-                        loc=constraint.loc,
-                        highlighting=False,
-                    )
-                    errors.append(utils.chain_exception(new_exc, e))
+                    if config.CAPTURED_ERRORS == 'old_style':
```
sklam, line 165: replace with `use_old_style_errors()`

numba/core/typing/templates.py
stuartarchibald: I guess this can only come from the arg to `signature` being non-callable, which would indicate a more fundamental problem, will make the change, thanks!

numba/core/typeinfer.py
stuartarchibald: 124f86f

numba/core/typeinfer.py
stuartarchibald: 124f86f

numba/core/errors.py
stuartarchibald: 124f86f

numba/core/errors.py
stuartarchibald: d63dab5

numba/core/typing/templates.py
stuartarchibald: d63dab5

sklam:
`numba.tests.test_np_functions.TestNPFunctions.test_take_along_axis` is failing on the farm on power8 and rpi64

numba/core/typing/templates.py
```diff
@@ -402,6 +402,15 @@ class CallableTemplate(FunctionTemplate):
     def apply(self, args, kws):
         generic = getattr(self, "generic")
         typer = generic()
+        match_sig = inspect.signature(typer)
+        try:
+            match_sig.bind(*args, **kws)
+        except TypeError as e:
+            # ctor or bind failed, raise, if there's a
```
gmarkall, line 409: I think we only catch a type error in the bind now with recent changes?

numba/core/typing/templates.py
```diff
@@ -772,7 +781,17 @@ def _build_impl(self, cache_key, args, kws):
         jitter = self._get_jit_decorator()
 
         # Get the overload implementation for the given types
-        ovf_result = self._overload_func(*args, **kws)
+        try:
+            ov_sig = inspect.signature(self._overload_func)
```
gmarkall, line 785: Should this be moved out of the try block as above?

numba/cpython/unsafe/tuple.py
gmarkall: `NUMBA_CAPTURED_ERRORS="new_style" python -m numba.runtests numba.cuda.tests` passes all tests for me locally.

gmarkall (changes requested):
This looks good and I'm happy to see things move in this direction! There are a couple of comments / questions on the diff.

I've tested with `NUMBA_CAPTURED_ERRORS="new_style" python -m numba.runtests numba.cuda.tests` locally (and without setting the env var) and all seems to be well.

> 3. Update the PR to set NUMBA_CAPTURED_ERROR="new_style" in the conda-build test phase.

This should say "`NUMBA_CAPTURED_ERRORS`" - for a while I was setting `NUMBA_CAPTURED_ERROR` from copy/pasting this, and wondering why there was no effect.

The other thing is that there should be some validation that `NUMBA_CAPTURED_ERRORS` is legitimate value, otherwise it's possible to hit the asserts in some of the branches if the user makes a typo setting the value.

numba/core/typing/templates.py
stuartarchibald: yes, thanks!

numba/core/typing/templates.py
stuartarchibald: Agree, will fix.

numba/core/errors.py
```diff
@@ -783,9 +822,17 @@ def new_error_context(fmt_, *args, **kwargs):
...
+            newerr = errcls(e).add_context(_format_msg(fmt_, args, kwargs))
+            if numba.core.config.FULL_TRACEBACKS:
+                tb = sys.exc_info()[2]
+            else:
+                tb = None
+            raise newerr.with_traceback(tb)
+        elif use_new_style_errors():
+            raise e
+        else:
+            assert 0
```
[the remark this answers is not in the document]
stuartarchibald: I've added handling of invalid values in d83cfc8, it uses the existing mechanism which is to warn if something invalid is found and proceed with the default.

stuartarchibald:
@gmarkall thanks for the review, I think I've address the comments.

sklam:
testing again on `numba_smoketest_cpu_yaml_51`

sklam:
BF passed on old-style errors. 
Also passed in manual testing on `-j"None,None,5"`slice of the testsuite with both flags.

sklam:
Looking at the errors on windows, maybe the batch file shouldn't have the quotes `"` around `new_style`.

```
AssertionError: 'ERROR' unexpectedly found in 'D:\\a\\1\\s\\numba\\core\\config.py:130: RuntimeWarning: environ NUMBA_CAPTURED_ERRORS defined but failed to parse \'"new_style"\'\n  warnings.warn("environ %s defined but failed to parse \'%s\'" %\nD:\\a\\1\\s\\numba\\core\\typed_passes.py:331: NumbaPerformanceWarning: \x1b[1m\nThe keyword argument \'parallel=True\' was specified but no transformation for parallel execution was possible.\n\nTo find out why, try turning on parallel diagnostics, see https://numba.pydata.org/numba-doc/latest/user/parallel.html#diagnostics for help.\n\x1b[1m\nFile "numba\\tests\\test_parfors.py", line 3803:\x1b[0m\n\x1b[1m        # Test for issue4007.\n\x1b[1m        def test_impl(n):\n\x1b[0m        \x1b[1m^\x1b[0m\x1b[0m\n\x1b[0m\n  warnings.warn(errors.NumbaPerformanceWarning(msg,\n..s....D:\\a\\1\\s\\numba\\core\\ir_utils.py:2141: NumbaPendingDeprecationWarning: \x1b[1m\nEncountered the use of a type that is scheduled for deprecation: type \'reflected list\' found for argument \'x\' of function \'TestPrangeSpecific.test_list_comprehension_prange.<locals>.test_impl\'.\n\nFor more information visit https://numba.pydata.org/numba-doc/latest/reference/deprecation.html#deprecation-of-reflection-for-list-and-set-types\n\x1b[1m\nFile "numba\\tests\\test_parfors.py", line 3852:\x1b[0m\n\x1b[1m        # issue4569\n\x1b[1m        def test_impl(x):\n\x1b[0m        \x1b[1m^\x1b[0m\x1b[0m\n\x1b[0m\n  warnings.warn(NumbaPendingDeprecationWarning(msg, loc=loc))\n...D:\\a\\1\\s\\numba\\core\\ir_utils.py:2141: NumbaPendingDeprecationWarning: \x1b[1m\nEncountered the use of a type that is scheduled for deprecation: type \'reflected list\' found for argument \'X\' of function \'TestPrangeSpecific.test_mutable_list_param.<locals>.list_check\'.\n\nFor more information visit 
```

PR #7398, opened 2021-09-13
Add support for arrayanalysis of tuple args.

DrTodd13:
Previously, if a tuple of integers were passed in and then used in a pndindex, the parfor pass would fail because the tuple of dimension sizes didn't have an entry in array analysis.  This PR adds integer tuple arguments to array analysis whereas previously only integer tuple variables were supported.

### DIFF
--- a/numba/parfors/array_analysis.py
+++ b/numba/parfors/array_analysis.py
@@ -1325,6 +1325,12 @@ class ArrayAnalysis(object):
                     shape = gvalue
                 elif isinstance(gvalue, int):
                     shape = (gvalue,)
+            elif isinstance(inst.value, ir.Arg):
+                if (
+                    isinstance(typ, types.containers.UniTuple)
+                    and isinstance(typ.dtype, types.Integer)
+                ):
+                    shape = inst.value
 
             if isinstance(shape, ir.Const):
                 if isinstance(shape.value, tuple):
@@ -3128,9 +3134,14 @@ class ArrayAnalysis(object):
         # attr call: A_sh_attr = getattr(A, shape)
         if isinstance(shape, ir.Var):
             shape = equiv_set.get_shape(shape)
+
         # already a tuple variable that contains size
         if isinstance(shape, ir.Var):
             attr_var = shape
+            shape_attr_call = None
+            shape = None
+        elif isinstance(shape, ir.Arg):
+            attr_var = var
             shape_attr_call = None
             shape = None
         else:
--- a/numba/tests/test_parfors.py
+++ b/numba/tests/test_parfors.py
@@ -1850,6 +1850,13 @@ class TestParfors(TestParforsBase):
         x = np.ones((2, 2, 2, 2, 2, 15))
         self.check(test_impl, x)
 
+    def test_tuple_arg(self):
+        def test_impl(x, sz):
+            for i in numba.pndindex(sz):
+                x[i] = 1
+            return x
+        sz = (10,10)
+        self.check(test_impl, np.empty(sz), sz)
 
 @skip_parfors_unsupported
 class TestParforsLeaks(MemoryLeakMixin, TestParforsBase):


gmarkall:
@DrTodd13 Thanks for the PR! Is there a related issue for this PR?

stuartarchibald:
Thanks for the patch. I wonder if adding the expression to equivalence analysis creates a problem for legitimate cases such as this:
```python
from numba import njit, pndindex
import numpy as np

@njit(parallel=True)
def test_impl(x, sz):
    for i in pndindex(sz):
        x[i] = 1
    return x

@njit(parallel=True)
def call(x):
    return test_impl(x, (10, 3)) # Only want to iterate the 3rd column

print(call(np.empty((10, 10))))
```
where a user wants to just iterate to the 3rd column of the 2d array.

DrTodd13:
> @DrTodd13 Thanks for the PR! Is there a related issue for this PR?

No.  It was just something that I ran across in the course of my regular work that I needed to work in Numba.

DrTodd13:
@stuartarchibald If I don't make "call" njit then your example works.  If I make "call" njit then I get an internal typing error at numba.core.typeinfer.CallConstraint which doesn't look related to my changes.

stuartarchibald:
> @stuartarchibald If I don't make "call" njit then your example works. If I make "call" njit then I get an internal typing error at numba.core.typeinfer.CallConstraint which doesn't look related to my changes.

I see the same, it's a chained exception, it looks a bit like:
```
<snip>
numba/parfors/parfor.py", line 2441, in run
    assert(equiv_set.has_shape(args[0]))
AssertionError: Failed in nopython mode pipeline (step: convert to parfors)


The above exception was the direct cause of the following exception:

<snip>
numba/core/typeinfer.py", line 1074, in propagate
    raise errors[0]
Internal error at <numba.core.typeinfer.CallConstraint object at 0x7fe012437940>.
Failed in nopython mode pipeline (step: convert to parfors
```
i.e. an `AssertionError` in parfors triggered it.

DrTodd13:
@stuartarchibald When you make the caller Numba then the tuple (10,3) doesn't become a UniTuple of Literal[int] but a regular Tuple of two Literal[int].  Is there a reason for that?  In any case, maybe that is better because you can also have a Tuple with one literal and the other a non-literal which my recent patch to this PR supports.  Please have another look.  I added an additional test for this mixed case (which subsumes the case of two literals).

stuartarchibald:
> @stuartarchibald When you make the caller Numba then the tuple (10,3) doesn't become a UniTuple of Literal[int] but a regular Tuple of two Literal[int]. Is there a reason for that?

I think it's because the IR preserves literals into the tuple type during type inference. 

https://github.com/numba/numba/blob/4a3d575f20c1f2ef7d4a661a6f52102b4039a40e/numba/core/typeinfer.py#L230-L241

This bit:
```
if vals and all(vals[0] == v for v in vals): 
```
will be False as the literal values will compare False. I think this might be from trying to preserve as much constant type information as possible.

> In any case, maybe that is better because you can also have a Tuple with one literal and the other a non-literal which my recent patch to this PR supports. Please have another look. I added an additional test for this mixed case (which subsumes the case of two literals).

I think that this is indeed the case.

numba/tests/test_parfors.py
```diff
@@ -1850,6 +1850,37 @@ def test_impl(x):
...
+            return x
+        sz = (10,10)
+        self.check(test_impl, np.empty(sz), sz)
+
+    def test_tuple_arg_literal(self):
+        def test_impl(x, sz):
+            for i in numba.pndindex(sz):
+                x[i] = 1
+            return x
+        sz = (10,numba.literally(10))
```
stuartarchibald, line 1866: I'm not sure that this has the desired effect. `numba.literally` outside of compiled code is a pass through NOP:
https://github.com/numba/numba/blob/4a3d575f20c1f2ef7d4a661a6f52102b4039a40e/numba/misc/special.py#L64-L87

it might be necessary to close over or locally define a const to get a literal if that's the intent?

numba/parfors/array_analysis.py
```diff
@@ -1325,6 +1325,20 @@ def _analyze_inst(self, label, scope, equiv_set, inst, redefined):
                     shape = gvalue
                 elif isinstance(gvalue, int):
                     shape = (gvalue,)
+            elif isinstance(inst.value, ir.Arg):
+                if (
+                    isinstance(typ, types.containers.UniTuple)
+                    and isinstance(typ.dtype, types.Integer)
+                ):
+                    shape = inst.value
+                elif (
```
stuartarchibald, line 1334: This branch doesn't appear to be taken by the new unittests?

numba/parfors/array_analysis.py
```diff
@@ -1379,6 +1393,16 @@ def _analyze_inst(self, label, scope, equiv_set, inst, redefined):
                     shape = self._gen_shape_call(
                         equiv_set, lhs, len(typ), shape, post
                     )
+            elif (
```
stuartarchibald, line 1396: This branch doesn't appear to be taken by the new unittests?

numba/parfors/parfor.py
```diff
@@ -1873,10 +1873,10 @@ def _type_getitem(self, args):
         return self.pass_states.typingctx.resolve_function_type(fnty, tuple(args), {})
 
 
-def _make_index_var(typemap, scope, index_vars, body_block):
+def _make_index_var(typemap, scope, index_vars, body_block, force_tuple=False):
```
stuartarchibald, line 1876: Might be a good idea to add a comment/docstring about what `force_tuple` is going to do and/or the situations where it is needed?

stuartarchibald (commented):
Thanks for the update. I've taken a look through and have left some new comments inline. I also wonder if it'd be a good idea to include the test case from here: https://github.com/numba/numba/pull/7398#issuecomment-919173454 somewhere so as to ensure this common pattern works.

Thanks again!

numba/tests/test_parfors.py
DrTodd13: I've added some tests and tried to generate the right tuples using what I think you were suggesting above with one literal arg but there is some lowering error I don't understand.  Can you take a look at it?

numba/parfors/array_analysis.py
DrTodd13: I added some tests that are supposed to take these paths and I confirmed it in isolated tests locally.

numba/tests/test_parfors.py
```diff
@@ -1850,6 +1850,57 @@ def test_impl(x):
...
+
+    def test_tuple_arg_1d_literal(self):
+        def test_impl(x):
+            sz = (10,)
+            for i in numba.pndindex(sz):
+                x[i] = 1
+            return x
+        sz = (10,)
+        self.check(test_impl, np.zeros(sz))
+
```
stuartarchibald, line 1903: I think this test:
```python
    def test_tuple_of_literal_nonliteral(self):
        # This test has to be done manually as the self.check uses
        # compile_isolated and one function cannot "see" the other

        def test_impl(x, sz):
            for i in numba.pndindex(sz):
                x[i] = 1
            return x

        def call(x, fn):
            return fn(x, (10, 3)) # Only want to iterate to the 3rd

        get_input = lambda: np.zeros((10, 10))
        expected = call(get_input(), test_impl)

        def check(dec):
            f1 = dec(test_impl)
            f2 = njit(call) # no parallel semantics in the caller
            got = f2(get_input(), f1)

        for d in (njit, njit(parallel=True)):
            check(d)
```
will exercise the path where there's an `ir.Arg` of type `Tuple(<int type>, <literal int>)`.

numba/tests/test_parfors.py
```diff
@@ -24,7 +24,7 @@
 import subprocess as subp
 
 import numba.parfors.parfor
-from numba import njit, prange, set_num_threads, get_num_threads, typeof
+from numba import njit, prange, set_num_threads, get_num_threads, typeof, literally
```
stuartarchibald, line 27:
```suggestion
from numba import njit, prange, set_num_threads, get_num_threads, typeof
```
think `literally` is now unused?

PR #7403, opened 2021-09-15
Fix for issue 7402: implement missing numpy ufunc interface

guilhermeleobas:
As title. cc @gmarkall

### DIFF
--- a/numba/np/ufunc/gufunc.py
+++ b/numba/np/ufunc/gufunc.py
@@ -105,27 +105,27 @@ class GUFunc(serialize.ReduceMixin):
     @property
     def signature(self):
         return self.ufunc.signature
-    
+
     @property
     def accumulate(self):
         return self.ufunc.accumulate
-    
+
     @property
     def at(self):
         return self.ufunc.at
-    
+
     @property
     def outer(self):
         return self.ufunc.outer
-    
+
     @property
     def reduce(self):
         return self.ufunc.reduce
-    
+
     @property
     def reduceat(self):
         return self.ufunc.reduceat
-    
+
     def _get_ewise_dtypes(self, args):
         argtys = map(lambda x: typeof(x), args)
         tys = []
--- a/numba/np/ufunc/gufunc.py
+++ b/numba/np/ufunc/gufunc.py
@@ -105,7 +105,27 @@ class GUFunc(serialize.ReduceMixin):
     @property
     def signature(self):
         return self.ufunc.signature
-
+    
+    @property
+    def accumulate(self):
+        return self.ufunc.accumulate
+    
+    @property
+    def at(self):
+        return self.ufunc.at
+    
+    @property
+    def outer(self):
+        return self.ufunc.outer
+    
+    @property
+    def reduce(self):
+        return self.ufunc.reduce
+    
+    @property
+    def reduceat(self):
+        return self.ufunc.reduceat
+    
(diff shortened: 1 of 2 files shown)


gmarkall:
@guilhermeleobas Many thanks for the PR! There are a few other members / properties that ufuncs have that dynamic gufuncs do not, including:

- `accumulate`
- `at`
- `outer`
- `reduce`
- `reduceat`

Does it make sense to add / pass through any of these too?

numba/np/ufunc/gufunc.py
```diff
@@ -102,6 +102,30 @@ def types(self):
...
+    def outer(self):
+        return self.ufunc.outer
+
+    @property
+    def reduce(self):
+        return self.ufunc.reduce
+
+    @property
+    def reduceat(self):
+        return self.ufunc.reduceat
```
stuartarchibald, line 127: I think it'd be a good idea to add tests for these properties such that future refactors don't reintroduce this bug or similar.

stuartarchibald (commented):
Thanks for the patch, think this needs the noted tests (in comments), else looks good.

gmarkall (commented):
The tests test that dynamic GUFuncs export the attributes, but they don't test that they do the right thing. It's not obvious to me that just because they're exported, they do the right thing - what happens if they are used with a data type not encountered previously?

sklam:
I can confirm that segfault issue is a numpy problem. For any gufunc that has a non-scalar signature, it will the `.at` will segfault.  For example:

```python
import numpy as np
import numpy.core.umath_tests as ut

x = np.arange(16).reshape(4, 4)
y = x * 2
print(ut.matrix_multiply.signature)
ut.matrix_multiply.at(x, (), y)
```
will segfault.

The reason is that [numpy `ufunc_at()` C implementation is putting `0xDEADBEEF` as the shape and strides](https://github.com/numpy/numpy/blob/d4d0584ccb508cad69288b504e6c8b04aa0d5caf/numpy/core/src/umath/ufunc_object.c#L6044-L6045). We can observe the `0xDEADBEEF` by:

```python
from numba import guvectorize
import numpy as np

@guvectorize("(x)->(x)")
def gufunc(x, res):
    acc = 0
    print(x.shape, x.strides)
    res[0] = x[0]


a = np.array([1, 2, 3, 4])
res = np.array([0, 0, 0, 0])
gufunc(a, res)

x = np.arange(24).reshape(4, 6)
y = x * 2
gufunc.at(x, [0, 0], y[0, :])   # this will show 0xdeadbeef (3735928559) for shape and strides
```

sklam (approved):
Thanks for the patch!

This patch will fix the regression of missing ufunc interface. However, requiring an invocation to `GUFunc.__call__` before a `GUFunc.at` can be used in the case of dynamic gufunc is not very user friendly. There needs to be a follow up PR to add automatic compiling of the required signature.

sklam:
@gmarkall, do you agree with https://github.com/numba/numba/pull/7403#pullrequestreview-811752860 or have other concerns?

gmarkall (approved):
Looks good now, though I agree with https://github.com/numba/numba/pull/7403#pullrequestreview-811752860.

stuartarchibald:
https://github.com/numba/numba/issues/7590 tracks the remaining task. Thanks for the patch @guilhermeleobas and reviews @gmarkall @sklam.

PR #7404, opened 2021-09-15
fix typo in literal_unroll docs

esc:
As title


stuartarchibald (approved):
Thanks for the patch.

PR #7419, opened 2021-09-20
insert missing backtick in comment

esc:
As title


stuartarchibald (approved):
Thanks for the patch.

PR #7421, opened 2021-09-20
Add option to extend variable lifetimes to the end of their block.

### LINKED ISSUE #7420: Debugger incorrectly shows zero value for variables
## Reporting a bug

Python code:
```
from numba import njit

@njit(debug=True)
def func_sum(a, b):
    c = a + b
    result = c
    return result

sum = func_sum(1, 2)
print(sum)
print("Done...")
```

Annotation:
```
$ NUMBA_DUMP_ANNOTATION=1 python bug.py
-----------------------------------ANNOTATION-----------------------------------
# File: bug.py
# --- LINE 3 ---

@njit(debug=True)

# --- LINE 4 ---

def func_sum(a, b):

    # --- LINE 5 ---
    # label 0
    #   a = arg(0, name=a)  :: int64
    #   b = arg(1, name=b)  :: int64
    #   c = a + b  :: int64
    #   del b
    #   del a

    c = a + b

    # --- LINE 6 ---
    #   result = c  :: int64
    #   del c

    result = c

    # --- LINE 7 ---
    #   $16return_value.5 = cast(value=result)  :: int64
    #   del result
    #   return $16return_value.5

    return result


================================================================================
3
Done...
```

Debugger:
```
$ NUMBA_OPT=0 gdb -q --args python bug.py
(gdb) break bug.py:6
(gdb) run
< ... >
Thread 1 "python" hit Breakpoint 1, __main__::func_sum$241 () at bug.py:6
6           result = c
(gdb) info locals
a = 0
b = 0
c = 3
result = 0
(gdb) next
7           return result
(gdb) info locals
a = 0
b = 0
c = 0
result = 3
(gdb)
```

When debugger hits the Python code line 6, variables _a_ and _b_ are already deleted and their contents in memory cleared. However, for Python code point of view, variables should still be alive, and user would expect debugger shows the values passed to the function (a = 1, b = 2).

Similarly, at line 7, debugger unexpectedly shows zero value for _a_, _b_ and _c_.
If variable lifetime would continue to the end of the function, as they do in Python, debugger should be able to show the actual values of the corresponding variables.

stuartarchibald:
This:
* Adds a new env var ``NUMBA_EXTEND_VARIABLE_LIFETIMES`` which when
  set extends the lifetime of variables to the end of the block in
  which their lifetime ends.
* Fixes up the Dels and their location information in the case of
  the above being set.
* Wires the above into the post processor call in the legalizer. The
  legalizer is the only place which should change the variable life
  times as it is the last thing that is executed in the pipeline
  before lowering. Changing the lifetimes in general will break
  optimisation passes that rely on lifetime analysis via Del nodes.

Fixes #7420

### DIFF
--- a/docs/source/reference/envvars.rst
+++ b/docs/source/reference/envvars.rst
@@ -125,6 +125,13 @@ These variables influence what is printed out during compilation of
    for each compiled function.
    Default value equals to the value of `NUMBA_ENABLE_PROFILING`.
 
+.. envvar:: NUMBA_EXTEND_VARIABLE_LIFETIMES
+
+    If set to non-zero, extend the lifetime of variables to the end of the block
+    in which their lifetime ends. This is particularly useful in conjunction
+    with `NUMBA_DEBUGINFO` as it helps with introspection of values. Default is
+    zero.
+
 .. envvar:: NUMBA_GDB_BINARY
 
    Set the ``gdb`` binary for use in Numba's ``gdb`` support, this takes the
--- a/numba/core/config.py
+++ b/numba/core/config.py
@@ -417,6 +417,9 @@ class _EnvReloader(object):
         DEBUGINFO_DEFAULT = _readenv("NUMBA_DEBUGINFO", int, ENABLE_PROFILING)
         CUDA_DEBUGINFO_DEFAULT = _readenv("NUMBA_CUDA_DEBUGINFO", int, 0)
 
+        EXTEND_VARIABLE_LIFETIMES = _readenv("NUMBA_EXTEND_VARIABLE_LIFETIMES",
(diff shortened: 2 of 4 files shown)


stuartarchibald:
c4c6baf enables the new env var in the CI set up, looks like object lifetime tests/refcount tests are failing, which is to be expected. There does however appear to be something related to generators that needs checking, e.g. `numba.tests.test_obj_lifetime.TestObjLifetime.test_generator1`.

stuartarchibald:
> numba.tests.test_obj_lifetime.TestObjLifetime.test_generator1

Seems like this also is down to changing lifetimes.

PR #7425, opened 2021-09-21
fix: preserve decorated function's type annotations

### LINKED ISSUE #7424: numba.jit strips away function type annotations, breaking mypy's analysis
## Reporting a bug

`numba.jit` decorator strips away function's type annotations. This is problematic when using `mypy` for static analysis.

Example:

```python
import numba as nb
import numpy as np

@nb.jit(nb.float64(nb.float64, nb.float64), nopython=True, nogil=True, inline="always")
def my_function(a: float, b: float) -> float:
    return 0.5 if np.isnan(a) or np.isnan(b) else 1.0
```

Mypy's error:
```
/path/to/bug.py:4: error: Untyped decorator makes function "my_function" untyped
```

castarco:
This PR introduces type annotations to the `@jit` decorator in order to preserve decorated functions' type annotations and avoid some problems with `mypy`.

~Closes #7424 .~

**EDIT:** I tried to apply these same changes in the installed library I'm using in my virtualenv... and it's not working as intended (`mypy` keeps complaining about missing types, I'm not sure if it's because I "hide" the type annotations under the `if typing.TYPE_CHECKING:` guard).

So I would say it's still a work in progress.

### DIFF
--- a/numba/core/decorators.py
+++ b/numba/core/decorators.py
@@ -7,6 +7,7 @@ import sys
 import warnings
 import inspect
 import logging
+import typing
 
 from numba.core.errors import DeprecationError, NumbaDeprecationWarning
 from numba.stencils.stencil import stencil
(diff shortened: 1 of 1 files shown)


esc:
@castarco thank you for submitting this. I have labelled it as i progress for now. Do make yourself known, if you would like this to be queued for review.

castarco:
I'm not working on this issue, I couldn't find an elegant way to patch the problem, and my focus shifted quite a bit these past weeks. Sorry about that.

Issue #7426, opened 2021-09-21
Loss of functionality in .54 - change in slicing behavior on user defined types


c200chromebook:
possible dupe of : https://github.com/numba/numba/issues/7355
The following code worked prior to 0.54, but no longer will run:

```
import numpy as np
import numba as nb

ALT_TYPE = [('Z', np.float64, 3)]
np_alt = np.dtype(ALT_TYPE, align=True)
nb_alt = nb.from_dtype(np_alt)
print(np_alt)
print(nb_alt)


@nb.njit(nb.void(nb_alt[:]))
def new_kern(z):
    z[0]['Z'][:] = 7


dta = np.zeros((1,), dtype=nb_alt)
new_kern(dta)
print(dta)
```

Yields:

```
No implementation of function Function(<built-in function setitem>) found for signature:
 
 >>> setitem(nestedarray(float64, (3,)), slice<a:b>, Literal[int](7))
 
There are 16 candidate implementations:
  - Of which 14 did not match due to:
  Overload of function 'setitem': File: <numerous>: Line N/A.
    With argument(s): '(nestedarray(float64, (3,)), slice<a:b>, int64)':
   No match.
  - Of which 2 did not match due to:
  Overload in function 'SetItemBuffer.generic': File: numba\core\typing\arraydecl.py: Line 171.
    With argument(s): '(nestedarray(float64, (3,)), slice<a:b>, int64)':
   Rejected as the implementation raised a specific error:
     TypeError: __init__() got an unexpected keyword argument 'ndim'
```

Expected behavior:
``` [([7,7,7],)]```

The bug affects CUDA as well but this is probably incidental.

**Comment by c200chromebook** [comment]
have confirmed still happens in `numba-0.55.0.dev0+299.gf299c6aa7-py3.7-win-amd64`

**Comment by esc** [comment]
@c200chromebook thank you for reporting this. I can reproduce, here is the full stack trace.

```
 💣 zsh» python issue7426.py
{'names':['Z'], 'formats':[('<f8', (3,))], 'offsets':[0], 'itemsize':24, 'aligned':True}
Record(Z[type=nestedarray(float64, (3,));offset=0];24;True)
Traceback (most recent call last):
  File "/Users/vhaenel/git/numba/numba/core/typeinfer.py", line 154, in propagate
    constraint(typeinfer)
  File "/Users/vhaenel/git/numba/numba/core/typeinfer.py", line 807, in __call__
[...]

PR #7428, opened 2021-09-21
Fix regression #7355: cannot set items in structured array data types 

sklam:
This is a minimal fix for #7355 and does not replace #7359. We need a minimal fix for 0.54.1

**Edit, adding details:**

This is the easiest fix because:

- making the base `Array.copy()` return a covariant type is the reason of the regression for `nestedarray`.
- The new _array subclass_ feature can override `copy()` to specialize accordingly.
- Requiring `copy()` is less invasive and _array subclass_ support is new and experimental.

This change will require _array subclass_ to override `copy()`.

### DIFF
--- a/numba/core/types/npytypes.py
+++ b/numba/core/types/npytypes.py
@@ -452,9 +452,8 @@ class Array(Buffer):
             layout = self.layout
         if readonly is None:
             readonly = not self.mutable
-        cls = type(self)
-        return cls(dtype=dtype, ndim=ndim, layout=layout, readonly=readonly,
-                   aligned=self.aligned)
+        return Array(dtype=dtype, ndim=ndim, layout=layout, readonly=readonly,
+                     aligned=self.aligned)
 
     @property
     def key(self):
--- a/numba/tests/test_ndarray_subclasses.py
+++ b/numba/tests/test_ndarray_subclasses.py
@@ -89,6 +89,11 @@ class MyArrayType(types.Array):
         super().__init__(dtype, ndim, layout, readonly=readonly,
                          aligned=aligned, name=name)
 
+    def copy(self, *args, **kwargs):
+        # This is here to future-proof.
+        # The test here never uses this.
+        raise NotImplementedError
+
     # Tell Numba typing how to combine MyArrayType with other ndarray types.
     def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
         if method == "__call__":
--- a/numba/tests/test_record_dtype.py
+++ b/numba/tests/test_record_dtype.py
@@ -315,6 +315,10 @@ def set_field4(rec):
     return rec
 
 
+def set_field_slice(arr):
+    arr['k'][:] = 0.0
+
+
 recordtype = np.dtype([('a', np.float64),
                        ('b', np.int16),
                        ('c', np.complex64),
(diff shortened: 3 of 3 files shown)


numba/core/types/npytypes.py
```diff
@@ -452,9 +452,8 @@ def copy(self, dtype=None, ndim=None, layout=None, readonly=None):
             layout = self.layout
         if readonly is None:
             readonly = not self.mutable
-        cls = type(self)
-        return cls(dtype=dtype, ndim=ndim, layout=layout, readonly=readonly,
-                   aligned=self.aligned)
+        return Array(dtype=dtype, ndim=ndim, layout=layout, readonly=readonly,
+                     aligned=self.aligned)
```
stuartarchibald, line 456: Note: This is as 0.53 https://github.com/numba/numba/blob/release0.53/numba/core/types/npytypes.py#L441-L451

stuartarchibald (approved):
Thanks for the patch. I've tested it with this... on HEAD of mainline with the diff to the tests that this patch introduces, the corresponding test fails. Later this PR is checked out and the same runs fine:
```
$ git rev-parse HEAD
ed6ae6d3c2a615d329604c96e034efc876cf8230

$ git diff
diff --git a/numba/tests/test_record_dtype.py b/numba/tests/test_record_dtype.py
index 8f1137f..699017e 100644
--- a/numba/tests/test_record_dtype.py
+++ b/numba/tests/test_record_dtype.py
@@ -315,6 +315,11 @@ def set_field4(rec):
     return rec
 
 
+def set_field_slice(arr):
+    arr['k'][:] = 0.0
+    return arr
+
+
 recordtype = np.dtype([('a', np.float64),
                        ('b', np.int16),
                        ('c', np.complex64),
@@ -865,6 +870,10 @@ class TestRecordDtype(unittest.TestCase):
         cfunc = self.get_cfunc(pyfunc, (nbrecord,))
         self.assertEqual(cfunc(rec), pyfunc(rec))
 
+        pyfunc = set_field_slice
+        cfunc = self.get_cfunc(pyfunc, (nbrecord,))
+        np.testing.assert_array_equal(cfunc(rec), pyfunc(rec))
+
     def test_structure_dtype_with_titles(self):
         # the following is the definition of int4 vector type from pyopencl
         vecint4 = np.dtype([(('x', 's0'), 'i4'), (('y', 's1'), 'i4'),

$ ./runtests.py numba.tests.test_record_dtype.TestRecordDtype.test_record_two_arrays
E
ERROR: test_record_two_arrays (numba.tests.test_record_dtype.TestRecordDtype)
Tests that comparison of NestedArrays by key is working correctly. If
<snip>
numba.core.errors.TypingError: No implementation of function Function(<built-in function setitem>) found for signature:
 
 >>> setitem(nestedarray(int32, (10, 20)), slice<a:b>, float64)
 
    With argument(s): '(nestedarray(int32, (10, 20)), slice<a:b>, float64)':
   No match.
  Overload in function 'SetItemBuffer.generic': File: numba/core/typing/arraydecl.py: Line 171.
    With argument(s): '(nestedarray(int32, (10, 20)), slice<a:b>, float64)':
     TypeError: __init__() got an unexpected keyword argument 'ndim'

$ #git stash/checkout this PR etc.

$ git rev-parse HEAD
1c3c10a528fb3e5ca0071886af49cc700a811577

$ ./runtests.py numba.tests.test_record_dtype.TestRecordDtype.test_record_two_arrays
.
Ran 1 test in 0.584s

OK
```
I've also manually tested this PR against all the examples in https://github.com/numba/numba/issues/7355 and https://github.com/numba/numba/issues/7426, and those reported here: https://numba.discourse.group/t/883/7

PR #7430, opened 2021-09-22
Cherrypicks for release 0.54.1

sklam:
as titled

### DIFF
--- a/CHANGE_LOG
+++ b/CHANGE_LOG
@@ -1,3 +1,26 @@
+Version 0.54.1 (23 September, 2021)
+-----------------------------------
+
+This is a bugfix release for 0.54.0. It fixes a regression in structured array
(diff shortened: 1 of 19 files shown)


stuartarchibald:
Thanks for the patch. I've replayed the same merge commits onto a local copy of the `release0.54` branch and there's no diff when compared to this branch. The merge commits tie out with the milestone and the change log.

stuartarchibald:
Note, 793a2ec adds the change log, checks in https://github.com/numba/numba/pull/7430#issuecomment-925291720 have been rechecked, it still matches.

stuartarchibald (approved):
Thanks for assembling the patches.

stuartarchibald:
NOTE: Not waiting on CI pass for merge as the "git diff" style tests will end up testing most of the test suite due to the amount of churn against mainline, net result is CI will likely fail due to timeout.

PR #7437, opened 2021-09-27
Add support for `np.broadcast_shapes`

guilhermeleobas:
As title. This PR requires `isinstance` support (#7204) to be merged first.


guilhermeleobas:
Failure doesn't seem to be related to this PR
```
ERROR: test_truediv_float (numba.tests.npyufunc.test_errors.TestFloatingPointExceptions)
  File "/home/vsts/work/1/s/numba/tests/npyufunc/test_errors.py", line 95, in test_truediv_float
    self.check_truediv_real(np.float64)
  File "/home/vsts/work/1/s/numba/tests/npyufunc/test_errors.py", line 91, in check_truediv_real
    res = f(a, b)
numpy.core._exceptions._UFuncNoLoopError: ufunc 'truediv' did not contain a loop with signature matching types (<class 'numpy.dtype[float64]'>, <class 'numpy.dtype[float64]'>) -> None

ERROR: test_math_erf (numba.cuda.tests.cudapy.test_math.TestCudaMath)
  File "/home/vsts/work/1/s/numba/cuda/tests/cudapy/test_math.py", line 433, in test_math_erf
    self.unary_template_float32(math_erf, ufunc)
  File "/home/vsts/work/1/s/numba/cuda/tests/cudapy/test_math.py", line 210, in unary_template_float32
    self.unary_template(func, npfunc, np.float32, np.float32, start, stop)
  File "/home/vsts/work/1/s/numba/cuda/tests/cudapy/test_math.py", line 238, in unary_template
    np.testing.assert_allclose(npfunc(A), B, rtol=rtol)
numpy.core._exceptions._UFuncNoLoopError: ufunc 'ufunc' did not contain a loop with signature matching types <class 'numpy.dtype[float32]'> -> None
```

numba/cpython/builtins.py
```diff
@@ -751,7 +751,8 @@ def false_impl(var, typs):
         if isinstance(typ, types.TypeRef):
             # Use of Numba type classes is in general not supported as they do
             # not work when the jit is disabled.
-            if key not in (types.ListType, types.DictType):
+            if key not in (types.ListType, types.DictType, types.Tuple,
+                           types.UniTuple):
```
stuartarchibald, line 755: Is this change actually demonstrating a potential bug in the impl that went out in 0.55 ?

stuartarchibald: Thanks for confirming, indeed this refuses to compile in  0.55!

```python
from numba import njit, types
import numpy as np

@njit
def foo(x):
    if isinstance(x, types.UniTuple):
        print("is tuple")
    else:
        print("is NOT tuple")

foo((1,2))
foo(1)
```
The note above RE JIT transparency suggests that it ought to not work either?

>            # Use of Numba type classes is in general not supported as they do
>            # not work when the jit is disabled.

guilhermeleobas: the problem with using `isinstance(x, tuple)` is that it is hard to differentiate between an empty tuple and a non-empty one.

guilhermeleobas: I've update `rewrite_semantic_constants` to recognize non-arg tuples. This change allows the JIT to recognize and optimize away cases like `isisntance(x, tuple) and len(x) == 0`. Let's see if I didn't break Numba with this change

stuartarchibald:
NOTE: it was decided OOB that the divergence of this branch against main had become too high/involved to make it worth the rebase effort, therefore the implementation of `np.broadcast_shapes` that is in this patch is presented again for review in #8003. Thanks @guilhermeleobas.

PR #7441, opened 2021-09-29
Permit any string as a key in literalstrkeydict type.

### LINKED ISSUE #6518: LiteralStrKeyDict error for non-identifier key names
## Reporting a bug

`LiteralStrKeyDict` creates a `namedtuple` in its constructor which requires key names to be valid identifiers, leading to unexpected errors. Seems to me like this can be avoided.
https://github.com/numba/numba/blob/d12bb5946527e2ef75f25136e89a53b156cb9b54/numba/core/types/containers.py#L775
```python
import numba

@numba.njit
def f():
    a = {"A+1": 4, "B": "C"}
    return a["A"]

print(f())
```
```
  File "../tmp/numba_litstrdict.py", line 9, in <module>
    print(f())
  File "/Users/ehsan/dev/numba/numba/core/dispatcher.py", line 415, in _compile_for_args
    error_rewrite(e, 'typing')
  File "/Users/ehsan/dev/numba/numba/core/dispatcher.py", line 358, in error_rewrite
    reraise(type(e), e, None)
  File "/Users/ehsan/dev/numba/numba/core/utils.py", line 80, in reraise
Internal error at <numba.core.typeinfer.BuildMapConstraint object at 0x7fecd29301f0>.
Type names and field names must be valid identifiers: 'A+1'
During: typing of dict at ../tmp/numba_litstrdict.py (5)

File "../tmp/numba_litstrdict.py", line 5:
def f():
    a = {"A+1": 4, "B": "C"}
    ^
```

### LINKED ISSUE #7416: Using literal dicts fails to compile when string keys hold numeric data

Hi!

Trying to use a LiteralStrKeyDict we found an issue, when string keys with numeric data are forbidden, e.g.

```
import numpy as np
import numba
from numba import types, njit

@njit
def tested():
    lit_dict = {'0': np.ones(5),
                    '1': ['a', 'b', 'c']}
    return len(lit_dict)

print("Dict size: ", tested())

# # below commented version works as expected
# @njit
# def tested():
#     lit_dict = {'f0': np.ones(5),
#                 'f1': ['a', 'b', 'c']}
#     return len(lit_dict)
#   
# print("Dict size: ", tested())
```

produces:
>   File "test_literal_dict.py", line 12, in <module>
>     print("Dict size: ", tested())
>   File "C:\Users\akozlov\AppData\Local\Continuum\anaconda3\numba_master\numba\numba\core\dispatcher.py", line 420, in _compile_for_args
>     error_rewrite(e, 'typing')
>   File "C:\Users\akozlov\AppData\Local\Continuum\anaconda3\numba_master\numba\numba\core\dispatcher.py", line 361, in error_rewrite
>     raise e.with_traceback(None)
> numba.core.errors.TypingError: Failed in nopython mode pipeline (step: nopython frontend)
> Internal error at <numba.core.typeinfer.BuildMapConstraint object at 0x000001F7C7783288>.
> Type names and field names must be valid identifiers: '0'
> During: typing of dict at test_literal_dict.py (9)
> Enable logging at debug level for details.
> 
> File "test_literal_dict.py", line 9:
> def tested():
>     <source elided>
>     lit_dict = {'0': np.ones(5),
>                 '1': ['a', 'b', 'c']}
>                 ^

It seems the reason is that namedtuple used in impl requires field names to be non-numeric, here:
https://github.com/numba/numba/blob/4911192c5b51f3ecbb76a12b74372979378c43a1/numba/core/types/containers.py#L775

When the above code executes `strkeys == ['0', '1']` and this line raises above exception. Is there any easy fix for this?

Best Regards,
Alexey.

stuartarchibald:
As title.

Fixes #7416
Fixes #6518

### DIFF
--- a/numba/core/types/containers.py
+++ b/numba/core/types/containers.py
@@ -1,4 +1,3 @@
-from collections import namedtuple
 from collections.abc import Iterable
 from types import MappingProxyType
 
@@ -766,13 +765,22 @@ class LiteralStrKeyDict(Literal, ConstSized, Hashable):
     namedtuple with dict semantics).
     """
 
+    class FakeNamedTuple(object):
+        # This is namedtuple-like and is a workaround for #6518 and #7416.
+        # This has the couple of namedtuple properties that are used by Numba's
+        # internals but avoids use of an actual namedtuple as it cannot have
+        # numeric field names, i.e. `namedtuple('foo', '0 1')` is invalid.
+        def __init__(self, name, keys):
+            self.__name__ = name
+            self._fields = tuple(keys)
+
     mutable = False
 
     def __init__(self, literal_value, value_index=None):
         self._literal_init(literal_value)
         self.value_index = value_index
         strkeys = [x.literal_value for x in literal_value.keys()]
-        self.tuple_ty = namedtuple("_ntclazz", " ".join(strkeys))
+        self.tuple_ty = self.FakeNamedTuple("_ntclazz", strkeys)
         tys = [x for x in literal_value.values()]
         self.types = tuple(tys)
         self.count = len(self.types)
--- a/numba/core/types/containers.py
+++ b/numba/core/types/containers.py
@@ -1,4 +1,5 @@
 from collections.abc import Iterable
+from collections.abc import Sequence as pySequence
 from types import MappingProxyType
 
 from .abstract import (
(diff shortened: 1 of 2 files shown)


numba/core/types/containers.py
```diff
@@ -766,13 +765,22 @@ class LiteralStrKeyDict(Literal, ConstSized, Hashable):
     namedtuple with dict semantics).
     """
 
+    class FakeNamedTuple(object):
```
sklam, line 768: Shouldn't this inherit from `Sequence`? The old type for `tuple_ty` is a `namedtuple` and it is not hidden (no `_` prefix), so users might have code expecting it has tuple-like API.

stuartarchibald: I've updated this to inherit from `collections.abc.Sequence` and to implement `__len__` and `__getitem__` by delegating to the `._fields` member.

sklam (approved):
LGTM

PR #7444, opened 2021-09-29
use correct variable name in closure

esc:
The argument name in the closure is `start` but `s` is used instead. The
only reason this works is because `s` is defined in the same scope that
also defines the closure. This fixes that.This fixes that.


sklam (approved):
Good catch!

PR #7447, opened 2021-09-29
cleanup Numba metadata

esc:
Cleanup the author and email listing on setup.py, the mailinglist is no
longer used and the company has been renamed. Also, we now host our docs
on readthedocs, so update that too.

In addition remove the mention of the "old" way of contact from the
README and the Sphinx based documentation.


setup.py
```diff
@@ -395,9 +395,7 @@ def check_file_at_path(path2file):
         "numba.typed": ["py.typed"],
     },
     scripts=["numba/pycc/pycc", "bin/numba"],
-    author="Anaconda, Inc.",
-    author_email="numba-users@continuum.io",
-    url="https://numba.pydata.org",
+    url="https://numba.readthedocs.io/",
```
stuartarchibald, line 398: Isn't https://numba.pydata.org Numba's home page? Is changing this to the docs page consistent with other packages?

esc: reverting

Issue #7423, opened 2021-09-20
numpy 1.21 ufunc reduce segfault


sklam:
Failing test: `numba.tests.npyufunc.test_ufunc.TestUFuncs.test_ufunc_attributes`

```pytb
Fatal Python error: Segmentation fault

Current thread 0x000000010dbdbe00 (most recent call first):
  File "/path/to/numba/numba/tests/npyufunc/test_ufunc.py", line 35 in ufunc_reduce
  File "/path/to/numba/numba/tests/npyufunc/test_ufunc.py", line 58 in _test_ufunc_attributes
  File "/path/to/numba/numba/tests/npyufunc/test_ufunc.py", line 74 in test_ufunc_attributes
```

sklam:
Related to https://github.com/numba/numba/issues/7175

stuartarchibald:
Closed by #7449 via https://github.com/numba/numba/pull/7483.

PR #7449, opened 2021-09-30
BUG: Fastcall fix ufunc methods (NumPy 1.21 support)

seberg:
This is a start to support NumPy 1.21 use of FASTCALL methods
in ufuncs.  There are probably better ways to achieve this support.

NumPy further uses `tp_vectorcall_offset` on newer versions.  This
PR does not add support for this (meaning that a DUfunc will be
unnecessarily slow on NumPy >=1.21 when kwargs are used).

Closes gh-7423

---

This is a start to close the NumPy 1.21 issues.  I am not extremely happy with this dance of duplicating the wrapper functions, but I did not have a better idea.
(Sorry, this is not really tested, yet aside from the fact that it compiles, so I may abuse CI to see if it actually is correct...)

I am sure there will be fix-ups necessary, please don't hesitate to just push/overwrite the changes here directly.

### DIFF
--- a/buildscripts/incremental/setup_conda_environment.sh
+++ b/buildscripts/incremental/setup_conda_environment.sh
@@ -40,7 +40,7 @@ conda list
 # NOTE: pyyaml is used to ensure that the Azure CI config is valid
 # NOTE: 32 bit linux... do not install NumPy, there's no conda package for >1.15
 # so it has to come from pip later
-if [[ "$CONDA_SUBDIR" == "linux-32" || "$BITS32" == "yes" ]]; then
+if [[ "$CONDA_SUBDIR" == "linux-32" || "$BITS32" == "yes" || "$NUMPY" == "1.21" ]]; then
     conda create -n $CONDA_ENV -q -y ${EXTRA_CHANNELS} python=$PYTHON pip gitpython pyyaml
 else
     conda create -n $CONDA_ENV -q -y ${EXTRA_CHANNELS} python=$PYTHON numpy=$NUMPY pip gitpython pyyaml
@@ -76,7 +76,7 @@ elif  [[ $(uname) == Darwin ]]; then
 fi
 
 # If on 32bit linux, now pip install NumPy and SciPy (no conda package)
-if [[ "$CONDA_SUBDIR" == "linux-32" || "$BITS32" == "yes" ]] ; then
+if [[ "$CONDA_SUBDIR" == "linux-32" || "$BITS32" == "yes" || "$NUMPY" == "1.21" ]] ; then
     $PIP_INSTALL numpy==$NUMPY scipy
 fi
 
--- a/numba/np/ufunc/_internal.c
+++ b/numba/np/ufunc/_internal.c
@@ -276,6 +276,7 @@ static PyMemberDef dufunc_members[] = {
  */
 
 static struct _ufunc_dispatch {
+    /* Note that the following may also hold `_PyCFunctionFastWithKeywords` */
     PyCFunctionWithKeywords ufunc_reduce;
     PyCFunctionWithKeywords ufunc_accumulate;
     PyCFunctionWithKeywords ufunc_reduceat;
@@ -286,7 +287,7 @@ static struct _ufunc_dispatch {
 } ufunc_dispatch;
 
 static int
-init_ufunc_dispatch(void)
+init_ufunc_dispatch(int *numpy_uses_fastcall)
 {
     int result = 0;
     PyMethodDef * crnt = PyUFunc_Type.tp_methods;
@@ -329,6 +330,16 @@ init_ufunc_dispatch(void)
             result = -1; /* Unknown method */
         }
         if (result < 0) break;
+
+        /* Check whether NumPy uses fastcall (ufunc.at never uses it) */
+        if (!strncmp(crnt_name, "at", 3)) {
+            if (*numpy_uses_fastcall == -1) {
+                *numpy_uses_fastcall = crnt->ml_flags & METH_FASTCALL;
+            }
+            else if (*numpy_uses_fastcall != (crnt->ml_flags & METH_FASTCALL)) {
+                return -1;
+            }
+        }
     }
     if (result == 0) {
         /* Sanity check. */
@@ -343,6 +354,7 @@ init_ufunc_dispatch(void)
     }
     return result;
 }
+
 
 static PyObject *
 dufunc_reduce(PyDUFuncObject * self, PyObject * args, PyObject *kws)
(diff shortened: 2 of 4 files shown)


numba/np/ufunc/_internal.c
```diff
@@ -329,6 +330,20 @@ init_ufunc_dispatch(void)
             result = -1; /* Unknown method */
         }
         if (result < 0) break;
+
+#if PY_VERSION_HEX >= 0x03070000
```
seberg, line 334: I suppose you probably only support >=3.7 and the `#ifdef`'s are actually unnecessary?

seberg:
Sorry, but the CI setup is confusing me with the fact that NumPy 1.21 cannot be found when added this way, not sure how to fix/approach that.

numba/np/ufunc/_internal.c
stuartarchibald: Numba supports >= 3.7, so no need to guard.

stuartarchibald:
> Sorry, but the CI setup is confusing me with the fact that NumPy 1.21 cannot be found when added this way, not sure how to fix/approach that.

Thanks for opening the PR and working on this, much appreciated.

Numba CI uses conda packages where possible, this line will be what's creating the envs for the additions to the build matrix made in this PR:
https://github.com/numba/numba/blob/e617b39a0b4b23d7b69d16f482fd66b4ac6cc307/buildscripts/incremental/setup_conda_environment.sh#L46

`pip` is available via:
https://github.com/numba/numba/blob/e617b39a0b4b23d7b69d16f482fd66b4ac6cc307/buildscripts/incremental/setup_conda_environment.sh#L21

Something like this might work (untested, needs the NumPy pip part filling in):
```diff
--- a/buildscripts/incremental/setup_conda_environment.sh
+++ b/buildscripts/incremental/setup_conda_environment.sh
@@ -40,7 +40,7 @@ conda list
 # NOTE: pyyaml is used to ensure that the Azure CI config is valid
 # NOTE: 32 bit linux... do not install NumPy, there's no conda package for >1.15
 # so it has to come from pip later
-if [[ "$CONDA_SUBDIR" == "linux-32" || "$BITS32" == "yes" ]]; then
+if [[ "$CONDA_SUBDIR" == "linux-32" || "$BITS32" == "yes" || "$NUMPY" == "1.21"]]; then
     conda create -n $CONDA_ENV -q -y ${EXTRA_CHANNELS} python=$PYTHON pip gitpython pyyaml
 else
     conda create -n $CONDA_ENV -q -y ${EXTRA_CHANNELS} python=$PYTHON numpy=$NUMPY pip gitpython pyyaml
@@ -80,6 +80,11 @@ if [[ "$CONDA_SUBDIR" == "linux-32" || "$BITS32" == "yes" ]] ; then
     $PIP_INSTALL numpy==$NUMPY scipy
 fi
 
+# if needing NumPy 1.21
+if [[ "$NUMPY" == "1.21" ]] ; then
+    $PIP_INSTALL <NOT SURE WHERE TO GET NUMPY=1.21?> scipy
+fi
+
 # Install latest llvmlite build
 $CONDA_INSTALL -c numba/label/dev llvmlite
 ```

File: buildscripts/incremental/setup_conda_environment.sh [hunk 2]
```diff
@@ -40,7 +40,7 @@ conda list
 # NOTE: pyyaml is used to ensure that the Azure CI config is valid
 # NOTE: 32 bit linux... do not install NumPy, there's no conda package for >1.15
 # so it has to come from pip later
-if [[ "$CONDA_SUBDIR" == "linux-32" || "$BITS32" == "yes" ]]; then
+if [[ "$CONDA_SUBDIR" == "linux-32" || "$BITS32" == "yes" || "$NUMPY" == "1.21"]]; then
```
stuartarchibald:
```suggestion
if [[ "$CONDA_SUBDIR" == "linux-32" || "$BITS32" == "yes" || "$NUMPY" == "1.21" ]]; then
```

buildscripts/incremental/setup_conda_environment.sh
```diff
@@ -76,7 +76,7 @@ elif  [[ $(uname) == Darwin ]]; then
 fi
 
 # If on 32bit linux, now pip install NumPy and SciPy (no conda package)
-if [[ "$CONDA_SUBDIR" == "linux-32" || "$BITS32" == "yes" ]] ; then
+if [[ "$CONDA_SUBDIR" == "linux-32" || "$BITS32" == "yes" || "$NUMPY" == "1.21"]] ; then
```
stuartarchibald:
```suggestion
if [[ "$CONDA_SUBDIR" == "linux-32" || "$BITS32" == "yes" || "$NUMPY" == "1.21" ]] ; then
```

stuartarchibald (commented):
might need a space at the end?

buildscripts/incremental/setup_conda_environment.sh
seberg: Thanks, had just managed to google it :).

seberg:
The fact that 1.21 broke numba is a bit ironic, because my new changes in the ufuncs carefully work around the fact that Numba sometimes mutates type list to avoid just that...
If this (unrelated, and from my side accidental/unexpected) change broke Numba without breaking the world, I should likely just pull out that compatibility layer soon...   Next time around we should try to have the compatible numba released _before_ the NumPy release, though!

The current NumPy main branch seems to run into a problem: not finding the loop that should be found by legacy-fallback.  That should be easy enough to fix, but I am curious if I could convince you to create a (non-failing) CI job against https://anaconda.org/scipy-wheels-nightly/numpy or using `pip install ... --pre`?

stuartarchibald:
> The fact that 1.21 broke numba is a bit ironic, because my new changes in the ufuncs carefully work around the fact that Numba sometimes mutates type list to avoid just that... If this (unrelated, and from my side accidental/unexpected) change broke Numba without breaking the world, I should likely just pull out that compatibility layer soon... Next time around we should try to have the compatible numba released _before_ the NumPy release, though!
>
> The current NumPy main branch seems to run into a problem: not finding the loop that should be found by legacy-fallback. That should be easy enough to fix, but I am curious if I could convince you to create a (non-failing) CI job against https://anaconda.org/scipy-wheels-nightly/numpy or using `pip install ... --pre`?

@seberg Thanks for taking extra care over the ufunc work to try and not break Numba, it's very much appreciated.

RE: future changes/testing against NumPy mainline, I wonder if https://github.com/numba/numba-integration-testing could help here? CC @esc

stuartarchibald:
This C code changes needed for NumPy 1.21 (df8442f) were committed as cff3dd6 and merged as part of https://github.com/numba/numba/pull/7483. That same PR updated the rest of Numba to accommodate NumPy 1.21 too. Thanks again for submitting the patches for this @seberg, much appreciated.

PR #7453, opened 2021-10-01
CUDA: Provide stream in async_done result

gmarkall:
It's awkward to await on a `Stream.async_done()` for the purpose of enqueuing something new onto the stream, because there is no way to tell which stream finished from the completed future - this forces the user to maintain a mapping between futures and streams.

This PR makes the result of the future be the stream on which operations completed, so that the user can obtain the stream from the completed future.

### DIFF
--- a/numba/cuda/cudadrv/driver.py
+++ b/numba/cuda/cudadrv/driver.py
@@ -1920,7 +1920,7 @@ class Stream(object):
     def async_done(self) -> asyncio.futures.Future:
         """
         Return an awaitable that resolves once all preceding stream operations
-        are complete.
+        are complete. The result of the awaitable is the current stream.
         """
         loop = asyncio.get_running_loop() if utils.PYVERSION >= (3, 7) \
             else asyncio.get_event_loop()
@@ -1930,7 +1930,7 @@ class Stream(object):
             if future.done():
                 return
             elif status == 0:
-                future.set_result(None)
+                future.set_result(self)
             else:
                 future.set_exception(Exception(f"Stream error {status}"))
 
--- a/numba/cuda/tests/cudadrv/test_streams.py
+++ b/numba/cuda/tests/cudadrv/test_streams.py
@@ -47,7 +47,8 @@ class TestCudaStream(CUDATestCase):
             h_src[:] = value_in
             d_ary = cuda.to_device(h_src, stream=stream)
             d_ary.copy_to_host(h_dst, stream=stream)
-            await stream.async_done()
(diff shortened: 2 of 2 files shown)


numba/cuda/cudadrv/driver.py
```diff
@@ -1930,7 +1930,7 @@ def resolver(future, status):
             if future.done():
                 return
             elif status == 0:
-                future.set_result(None)
+                future.set_result(self)
```
stuartarchibald, line 1933: Is there any chance that the API might need to return something more than just the stream in the future, if so should this return a container of some sort of which `stream` is a member?

numba/cuda/tests/cudadrv/test_streams.py
```diff
@@ -59,7 +60,18 @@ async def async_cuda_fn(value_in: float) -> float:
...
         stream = cuda.stream()
         done_aws = [stream.async_done() for _ in range(4)]
-        await asyncio.gather(*done_aws)
+        done = await asyncio.gather(*done_aws)
+        for d in done:
+            self.assertEqual(d, stream)
+
+    @with_asyncio_loop
+    async def test_multiple_async_done_multiple_streams(self):
+        streams = [ cuda.stream() for _ in range(4)]
```
stuartarchibald, line 69:
```suggestion
        streams = [cuda.stream() for _ in range(4)]
```

stuartarchibald (commented):
Thanks for the patch, couple of minor things to look at else looks good. This new capability will be useful for e.g. data streams with non-uniform workloads.

numba/cuda/cudadrv/driver.py
gmarkall: As discussed OOB, I think I'd prefer to keep the interface simple for now and just return the stream. If extra info is needed in future, it could be enabled by an optional default-`False` kwarg on the `async_done` method.

stuartarchibald: Thanks for recording this outcome.

stuartarchibald (approved):
Thanks for the patch!

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_97`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_97`.

Passed.

MurrayData:
This is great, thank you @gmarkall, really useful for streams with non-uniform workloads, as @stuartarchibald said, such as the satellite data feeds I work with.

PR #7456, opened 2021-10-05
Fix invalid codegen for #7451.

### LINKED ISSUE #7451: Corruption on ppc64le/arm64 (potential codegen/LLVM opt bug)
## Reporting a bug

The test `numba.tests.test_np_functions.TestNPFunctions.test_take_along_axis` currently fails against e617b39a0b4b23d7b69d16f482fd66b4ac6cc307 (HEAD of current mainline) with `SIGSEGV` on `linux-ppc64le` and `linux-arm64`.

This is a somewhat minimised reproducer:
```python
from numba import njit, types, config
from numba.extending import register_jitable
import numpy as np

if config.DISABLE_JIT:
    def tuple_setitem(tup, idx, val):
        return tup[:idx] + (val,) + tup[idx+1:]
else:
    from numba.cpython.unsafe.tuple import tuple_setitem

@register_jitable
def _tal_impl(arr, indices, axis, indices_broadcast_shape):
    arr_shape = list(arr.shape) # comment just this and a ValueError is raised from the broadcast (ppc64le)
    print(indices_broadcast_shape) # comment just this and it "works" on ppc64le, ValueError in broadcast on arm64
    arr_broadcast_shape = tuple_setitem(indices_broadcast_shape, axis, arr.shape[axis])
    # Call these in either order, effect is the same, corruption
    np.broadcast_to(indices, indices_broadcast_shape)
    np.broadcast_to(arr, arr_broadcast_shape)

a = np.arange(24).reshape(3, 1, 4, 2)

@njit
def impl(a, i):
    return _tal_impl(a, i, -1, (3, 1, 4, 2))

ai = np.argsort(a, axis=-1)
impl(a, ai)
```
Failure table:
| `NUMBA_OPT` level|ARM64       |PPC64LE|
|----------------------------|-------------------|----------|
| 0 | ok | ok |
| 1 | `ValueError` from `np.broadcast_to` impl | ok |
| 2 | `SIGBUS` | `SIGSEGV` |
| 3 | `SIGBUS` | `SIGSEGV` |

Notes:
 * Running the script under `valgrind` on `ppc64le` with `NUMBA_OPT=3` leads to a `SIGBUS`.
 * Running the script under `valgrind` on `x86_64` with `NUMBA_OPT=3` has nothing suspicious.
 * Commenting out various lines in the script (as noted) triggers different behaviour.
 * Running the script with `NUMBA_DISABLE_JIT=1` has no errors.

I think this somewhat hints at corruption from an invalid optimisation.

stuartarchibald:
As title.

Fixes #7451

### DIFF
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -5545,6 +5545,11 @@ def _take_along_axis_impl(
     # With addition of pre-broadcasting:
     # https://github.com/numpy/numpy/issues/19704
 
+    # Wrap axis, it's used in tuple_setitem so must be (axis >= 0) to ensure
+    # the GEP is in bounds.
+    if axis < 0:
+        axis = arr.ndim + axis
+
     # Broadcast the two arrays to matching shapes:
     arr_shape = list(arr.shape)
     arr_shape[axis] = 1
@@ -5620,7 +5625,7 @@ def arr_take_along_axis(arr, indices, axis):
         check_is_integer(axis, "axis")
         if not isinstance(axis, types.IntegerLiteral):
             raise ValueError(
-                "axis must be a literal value (i.e. not an argument) for now"
+                "axis must be a literal value (i.e. not an argument)"
             )
         axis = axis.literal_value
         if axis < 0:
--- a/numba/tests/test_np_functions.py
+++ b/numba/tests/test_np_functions.py
@@ -4461,14 +4461,19 @@ class TestNPFunctions(MemoryLeakMixin, TestCase):
         arr = np.ones((3, 4, 1))
         ai = np.ones((1, 2, 5), dtype=np.intp)
 
-        @njit
-        def check(a, i):
-            return np.take_along_axis(a, i, axis=1)
+        def gen(axis):
+            @njit
+            def impl(a, i):
+                return np.take_along_axis(a, i, axis)
+            return impl
 
-        expected = check.py_func(arr, ai)
-        actual = check(arr, ai)
-        self.assertPreciseEqual(expected, actual)
-        self.assertEqual(actual.shape, (3, 2, 5))
+        # Check same axis but expressed as positive/negative value
+        for i in (1, -2):
+            check = gen(i)
+            expected = check.py_func(arr, ai)
+            actual = check(arr, ai)
+            self.assertPreciseEqual(expected, actual)
+            self.assertEqual(actual.shape, (3, 2, 5))
 
     def test_take_along_axis_exceptions(self):
         arr2d = np.arange(8).reshape(2, 4)


stuartarchibald:
6562250 extends a current unit test to include prescribing a negative axis value to trigger the bug described here: https://github.com/numba/numba/issues/7451#issuecomment-933703118. This will manifest as not being able to broadcast.

stuartarchibald:
The addition of f3755f2 fixes the problem demonstrated in 6562250.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cpu_yaml_50`.

gmarkall:
This fixes this particular issue, but should we instead change `tuple_setitem` so that it supports negative indexing? What if there are other latent instances of this issue in the codebase?

gmarkall:
> This fixes this particular issue, but should we instead change `tuple_setitem` so that it supports negative indexing? What if there are other latent instances of this issue in the codebase?

Or maybe "also", since I see why the change is needed to allow broadcasting here.

stuartarchibald:
> This fixes this particular issue, but should we instead change `tuple_setitem` so that it supports negative indexing? What if there are other latent instances of this issue in the codebase?

Thanks for checking the patch. Suggest talking about this at the public meeting today. The function is "unsafe" by design, I guess the question is how unsafe that should be.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cpu_yaml_50`.

Passed on f3755f2

gmarkall:
> The function is "unsafe" by design, I guess the question is how unsafe that should be.

I think at the moment it's a bit of a footgun, because a user would reasonably expect it to handle negative indexing (consistent with lots of other indexing use cases) but its documentation makes no mention of it not supporting negative indexing. If we decide not to support negative indexing, then I think it would be good to mention that it doesn't support negative indexing in the docs.

gmarkall (approved):
Looks good!

stuartarchibald:
> > The function is "unsafe" by design, I guess the question is how unsafe that should be.
> 
> I think at the moment it's a bit of a footgun, because a user would reasonably expect it to handle negative indexing (consistent with lots of other indexing use cases) but its documentation makes no mention of it not supporting negative indexing. If we decide not to support negative indexing, then I think it would be good to mention that it doesn't support negative indexing in the docs.

Agree, it either needs the checks or some big warning about what "in-bounds" means, i.e. not the Python version.

I think given the current PR backlog, as this fixes the immediate issue it should be merged in as it makes mainline stable again and permits merges of other PRs. However, I'll raise the question of "how safe is `numba.*.unsafe.*`?"  at the public meeting today. This also makes me wonder if there should be a "safe" mode where things like this are checked.

stuartarchibald:
> Looks good!

sklam:
I see all unsafe intrinsics as the same level of safety as inlineasm; thus, no safety checks at all. We can provide a "safe" version of it though but leave dev a choice to opt into the unsafe and faster one.

sklam (approved):
LGTM

PR #7459, opened 2021-10-05
Include compiler flags in symbol mangling

sklam:
TBD

### DIFF
--- a/numba/core/base.py
+++ b/numba/core/base.py
@@ -288,11 +288,11 @@ class BaseContext(object):
         Load target-specific registries.  Can be overridden by subclasses.
         """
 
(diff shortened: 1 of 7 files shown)


sklam:
With the last commit, gdb works again. It turns out [`$` is reserved in the itanium mangling](https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangling-general) and only works in private implementation. So, gdb is not able to parse it correctly.

stuartarchibald:
Should probably review and merge https://github.com/numba/numba/pull/7177 ahead of this.

sklam:
TODO: 

- [ ] add `__repr__` to parfor options and fastmath options

PR #7460, opened 2021-10-06
Testhound/fp16 support

testhound:
This PR begins the process of adding 16-bit floating point support for the cuda target. This PR adds support for the following 16-bit fp operators: add, subtract, multiply, multiply accumulate, negate and absolute value. The divide operator was omitted as it is more involved and will be included in the next PR.


testhound:
> Looks like the CI failure is because the fp16 intrinsics is not available in cudasim

@sklam pushed a patch for cudasim.

gmarkall:
This needs some fixups to stop the segfault on NVVM 3.4, so I've changed the status back from waiting on CI to waiting on author (cc @sklam).

testhound:
gpuci run tests

testhound:
gpuci run tests

numba/cuda/tests/cudapy/test_intrinsics.py
```diff
@@ -292,6 +341,150 @@ def test_fma_f8(self):
         compiled[1, 1](ary, 2., 3., 4.)
         np.testing.assert_allclose(ary[0], 2 * 3 + 4)
 
+    def toolkit_and_compute_supported(self, compute_major, compute_minor,
+                                      toolkit_major, toolkit_minor):
```
stuartarchibald, line 345: Perhaps the interface could be:
```python
def toolkit_and_compute_supported(compute=(major, minor), toolkit=(major, minor))
```
so as to make it easier to identify which values belong to which requirement?

stuartarchibald: Or even two decorators, `toolkit_support(major, minor)` and `compute_support(major, minor)`. CC @gmarkall as code owner.

gmarkall:
The following:

```python
from numba import cuda
import numpy as np


@cuda.jit
def f(r, h1, h2):
    i = cuda.grid(1)
    if i >= len(r):
        return
    r[i] = cuda.fp16.hadd(h1[i],  h2[i])


np.random.seed(1)
x = np.random.rand(10).astype(np.float16)
y = np.random.rand(10).astype(np.float16)
r = np.zeros(x.shape, dtype=np.float32)

f[1, 32](r, x, y)

ref = (x + y).astype(np.float32)
np.testing.assert_array_equal(r, ref)
```

results in:

```
  File "/home/gmarkall/numbadev/issues/7460/cast_issue.py", line 17, in <module>
    f[1, 32](r, x, y)
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 676, in __call__
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 864, in call
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 872, in _compile_for_args
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 1033, in compile
    kernel = _Kernel(self.py_func, argtypes, link=self.link,
    return func(*args, **kwargs)
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 373, in __init__
    self.cooperative = 'cudaCGGetIntrinsicHandle' in lib.get_asm_str()
  File "/home/gmarkall/numbadev/numba/numba/cuda/codegen.py", line 103, in get_asm_str
    return self._join_ptxes(self._get_ptxes(cc=cc))
  File "/home/gmarkall/numbadev/numba/numba/cuda/codegen.py", line 141, in _get_ptxes
    ptxes = [nvvm.llvm_to_ptx(irs, **options)]
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudadrv/nvvm.py", line 711, in llvm_to_ptx
    return cu.compile(**opts)
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudadrv/nvvm.py", line 297, in compile
    self._try_error(err, 'Failed to compile\n')
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudadrv/nvvm.py", line 315, in _try_error
    self.driver.check_error(err, "%s\n%s" % (msg, self.get_log()))
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudadrv/nvvm.py", line 192, in check_error
    raise exc

<unnamed> (184, 18): parse invalid cast opcode for cast from 'i16' to 'float'
NVVM_ERROR_COMPILATION
```

I think this is because there are no specific implementations of lowering casts for float16 types. A sketch of an appropriate cast looks like:

```python
@lower_cast(types.float16, types.Float)
def float16_cast(context, builder, fromty, toty, val):
    if fromty.bitwidth == toty.bitwidth:
        return val
    casted = builder.bitcast(val, ir.HalfType())
    if fromty.bitwidth < toty.bitwidth:
        return builder.fpext(casted, context.get_value_type(toty))
    else:
        return builder.fptrunc(casted, context.get_value_type(toty))
```

where `lower_cast` is defined as `registry.lower_cast` in `numba/cuda/cudaimpl.py`. This cast (and other appropriate casts as required - potentially to integers, complex values, etc.) need to be added, along with tests that mix float16 with other types in artihmetic operations.

numba/cuda/tests/cudapy/test_intrinsics.py
gmarkall: I think separate functions `toolkit_support` and `compute_support` would be easier to read at the call site so I'd prefer that option.

testhound:
gpuci run tests

gmarkall:
When I run the following:

```python
from numba import cuda
import numpy as np


@cuda.jit
def fadd(r, x, y):
    r[0] = cuda.fp16.hadd(x[0], y[0])

x = np.ones(1, dtype=np.float16) * 2.5
y = np.ones(1, dtype=np.float16) * 3.7
r = np.zeros_like(x)

fadd[1, 1](r, x, y)
```

I get:

```
  File "/home/gmarkall/numbadev/issues/7460/int_to_float_cast.py", line 13, in <module>
    f[1, 1](r, x)
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 727, in __call__
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 915, in call
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 923, in _compile_for_args
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 1084, in compile
    kernel = _Kernel(self.py_func, argtypes, link=self.link,
    return func(*args, **kwargs)
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 390, in __init__
    cres = compile_cuda(self.py_func, types.void, self.argtypes,
    return func(*args, **kwargs)
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 200, in compile_cuda
    cres = compiler.compile_extra(typingctx=typingctx,
  File "/home/gmarkall/numbadev/numba/numba/core/compiler.py", line 690, in compile_extra
  File "/home/gmarkall/numbadev/numba/numba/core/compiler.py", line 428, in compile_extra
  File "/home/gmarkall/numbadev/numba/numba/core/compiler.py", line 496, in _compile_bytecode
    return self._compile_core()
  File "/home/gmarkall/numbadev/numba/numba/core/compiler.py", line 471, in _compile_core
    raise e
  File "/home/gmarkall/numbadev/numba/numba/core/compiler.py", line 462, in _compile_core
    pm.run(self.state)
  File "/home/gmarkall/numbadev/numba/numba/core/compiler_machinery.py", line 342, in run
    raise e
  File "/home/gmarkall/numbadev/numba/numba/core/compiler_machinery.py", line 334, in run
    return func(*args, **kwargs)
  File "/home/gmarkall/numbadev/numba/numba/core/compiler_machinery.py", line 289, in _runPass
  File "/home/gmarkall/numbadev/numba/numba/core/compiler_machinery.py", line 262, in check
  File "/home/gmarkall/numbadev/numba/numba/core/typed_passes.py", line 396, in run_pass
    lower.lower()
  File "/home/gmarkall/numbadev/numba/numba/core/lowering.py", line 183, in lower
  File "/home/gmarkall/numbadev/numba/numba/core/lowering.py", line 237, in lower_normal_function
  File "/home/gmarkall/numbadev/numba/numba/core/lowering.py", line 266, in lower_function_body
    self.lower_block(block)
  File "/home/gmarkall/numbadev/numba/numba/core/lowering.py", line 280, in lower_block
    self.lower_inst(inst)
  File "/home/gmarkall/numbadev/numba/numba/core/lowering.py", line 425, in lower_inst
    val = self.lower_assign(ty, inst)
  File "/home/gmarkall/numbadev/numba/numba/core/lowering.py", line 611, in lower_assign
    return self.lower_expr(ty, value)
  File "/home/gmarkall/numbadev/numba/numba/core/lowering.py", line 1146, in lower_expr
    res = self.lower_call(resty, expr)
  File "/home/gmarkall/numbadev/numba/numba/core/lowering.py", line 876, in lower_call
    res = self._lower_call_normal(fnty, expr, signature)
  File "/home/gmarkall/numbadev/numba/numba/core/lowering.py", line 1117, in _lower_call_normal
    res = impl(self.builder, argvals, self.loc)
  File "/home/gmarkall/numbadev/numba/numba/core/base.py", line 1196, in __call__
    res = self._imp(self._context, builder, self._sig, args, loc=loc)
  File "/home/gmarkall/numbadev/numba/numba/core/base.py", line 1226, in wrapper
    return fn(*args, **kwargs)
  File "/home/gmarkall/numbadev/numba/numba/cpython/builtins.py", line 341, in number_constructor
    return context.cast(builder, val, valty, sig.return_type)
  File "/home/gmarkall/numbadev/numba/numba/core/base.py", line 713, in cast
    return impl(self, builder, fromty, toty, val)
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudaimpl.py", line 508, in integer_to_float16_cast
    if toty.signed:
AttributeError: 'Float' object has no attribute 'signed'
```
```

**Comment by gmarkall** [comment]
Running:

```python
from numba import cuda
import numpy as np


@cuda.jit
def f(r, x):
    r[0] = cuda.fp16.hmul(np.float16(x[0]), 2)


x = np.ones(1, dtype=np.float16)
r = np.zeros_like(x)

f[1, 1](r, x)
```

results in:

```
  File "/home/gmarkall/numbadev/issues/7460/hmul_int.py", line 13, in <module>
    f[1, 1](r, x)
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 727, in __call__
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 915, in call
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 923, in _compile_for_args
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 1084, in compile
    kernel = _Kernel(self.py_func, argtypes, link=self.link,
    return func(*args, **kwargs)
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 390, in __init__
    cres = compile_cuda(self.py_func, types.void, self.argtypes,
    return func(*args, **kwargs)
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 200, in compile_cuda
    cres = compiler.compile_extra(typingctx=typingctx,
  File "/home/gmarkall/numbadev/numba/numba/core/compiler.py", line 690, in compile_extra
  File "/home/gmarkall/numbadev/numba/numba/core/compiler.py", line 428, in compile_extra
  File "/home/gmarkall/numbadev/numba/numba/core/compiler.py", line 496, in _compile_bytecode
    return self._compile_core()
  File "/home/gmarkall/numbadev/numba/numba/core/compiler.py", line 475, in _compile_core
    raise e
  File "/home/gmarkall/numbadev/numba/numba/core/compiler.py", line 462, in _compile_core
    pm.run(self.state)
  File "/home/gmarkall/numbadev/numba/numba/core/compiler_machinery.py", line 346, in run
    raise patched_exception
  File "/home/gmarkall/numbadev/numba/numba/core/compiler_machinery.py", line 334, in run
    return func(*args, **kwargs)
  File "/home/gmarkall/numbadev/numba/numba/core/compiler_machinery.py", line 289, in _runPass
  File "/home/gmarkall/numbadev/numba/numba/core/compiler_machinery.py", line 262, in check
  File "/home/gmarkall/numbadev/numba/numba/core/typed_passes.py", line 396, in run_pass
    lower.lower()
  File "/home/gmarkall/numbadev/numba/numba/core/lowering.py", line 183, in lower
  File "/home/gmarkall/numbadev/numba/numba/core/lowering.py", line 237, in lower_normal_function
  File "/home/gmarkall/numbadev/numba/numba/core/lowering.py", line 266, in lower_function_body
    self.lower_block(block)
  File "/home/gmarkall/numbadev/numba/numba/core/lowering.py", line 280, in lower_block
    self.lower_inst(inst)
  File "/home/gmarkall/numbadev/numba/numba/core/lowering.py", line 425, in lower_inst
    val = self.lower_assign(ty, inst)
  File "/home/gmarkall/numbadev/numba/numba/core/lowering.py", line 611, in lower_assign
    return self.lower_expr(ty, value)
  File "/home/gmarkall/numbadev/numba/numba/core/lowering.py", line 1146, in lower_expr
    res = self.lower_call(resty, expr)
  File "/home/gmarkall/numbadev/numba/numba/core/lowering.py", line 876, in lower_call
    res = self._lower_call_normal(fnty, expr, signature)
  File "/home/gmarkall/numbadev/numba/numba/core/lowering.py", line 1099, in _lower_call_normal
    argvals = self.fold_call_args(
  File "/home/gmarkall/numbadev/numba/numba/core/lowering.py", line 798, in fold_call_args
    argvals = [self._cast_var(var, sigty)
  File "/home/gmarkall/numbadev/numba/numba/core/lowering.py", line 798, in <listcomp>
    argvals = [self._cast_var(var, sigty)
  File "/home/gmarkall/numbadev/numba/numba/core/lowering.py", line 781, in _cast_var
    return self.context.cast(self.builder, val, varty, ty)
  File "/home/gmarkall/numbadev/numba/numba/core/base.py", line 712, in cast
    impl = self._casts.find((fromty, toty))
  File "/home/gmarkall/numbadev/numba/numba/core/base.py", line 49, in find
    out = self._find(sig)
  File "/home/gmarkall/numbadev/numba/numba/core/base.py", line 56, in _find
    return candidates[self._best_signature(candidates)]
  File "/home/gmarkall/numbadev/numba/numba/core/base.py", line 84, in _best_signature
    raise errors.NumbaTypeError('\n'.join(msg))
numba.core.errors.NumbaTypeError: Failed in cuda mode pipeline (step: native lowering)
2 ambiguous signatures
(<class 'numba.core.types.scalars.IntegerLiteral'>, <class 'numba.core.types.scalars.Float'>) => <function literal_int_to_number at 0x7f6631f4a430>
(<class 'numba.core.types.scalars.Integer'>, float16) => <function integer_to_float16_cast at 0x7f6631cc7700>
During: lowering "$22call_method.10 = call $6load_method.2($18call_method.8, $const20.9, func=$6load_method.2, args=[Var($18call_method.8, hmul_int.py:7), Var($const20.9, hmul_int.py:7)], kws=(), vararg=None, target=None)" at /home/gmarkall/numbadev/issues/7460/hmul_int.py (7)
```

numba/cuda/cudaimpl.py
```diff
@@ -405,6 +405,213 @@ def ptx_popc(context, builder, sig, args):
 def ptx_fma(context, builder, sig, args):
     return builder.fma(*args)
 
+
+@lower_cast(types.float16, types.Float)
+def float16_to_float_cast(context, builder, fromty, toty, val):
+    if fromty.bitwidth == toty.bitwidth:
+        return val
+
+    conv_type = ""
```
gmarkall, line 414: This is superfluous:
```suggestion
```

numba/cuda/cudaimpl.py
```diff
@@ -405,6 +405,213 @@ def ptx_popc(context, builder, sig, args):
...
+                                 f"={conv_constraint},h")
+    return builder.call(cvt2float_asm, [val])
+
+
+@lower_cast(types.Float, types.float16)
+def float_to_float16_cast(context, builder, fromty, toty, val):
+    if fromty.bitwidth == toty.bitwidth:
+        return val
+
+    conv_type = ""
```
gmarkall, line 439: This is superfluous:
```suggestion
```

numba/cuda/cudaimpl.py
```diff
@@ -405,6 +405,213 @@ def ptx_popc(context, builder, sig, args):
...
+    elif toty.bitwidth == 64:
+        conv_constraint = "l"
+    elif toty.bitwidth == 8:
+        conv_constraint = "c"
+    else:
+        raise error.CudaSupportError(f"Cuda target does not support "
+                                     f"from fp16 converting to integer type "
+                                     f"with width {toty.bitwidth}-bits")
+
+    signedness = ""
```
gmarkall, line 476: This is superfluous:
```suggestion
```

numba/cuda/cudaimpl.py
```diff
@@ -405,6 +405,213 @@ def ptx_popc(context, builder, sig, args):
...
+    elif fromty.bitwidth == 64:
+        conv_constraint = "l"
+    elif fromty.bitwidth == 8:
+        conv_constraint = "c"
+    else:
+        raise error.CudaSupportError(f"Cuda target does not support "
+                                     f"converting from integer type with "
+                                     f"width {toty.bitwidth}-bits to fp16")
+
+    signedness = ""
```
gmarkall, line 507: This is superfluous:
```suggestion
```

numba/cuda/cudaimpl.py
```diff
@@ -405,6 +405,213 @@ def ptx_popc(context, builder, sig, args):
...
+                                f".f16 $0, $1;",
+                                f"=h,{conv_constraint}")
+    return builder.call(cvt2fp16_asm, [val])
+
+
+def lower_fp16_binary(fn, op):
+    @lower(fn, types.float16, types.float16)
+    def ptx_fp16_binary(context, builder, sig, args):
+        compute_capability_supported = \
+            cuda.current_context().device.compute_capability >= (5, 3)
```
gmarkall, line 526: I realised recently that we can't check the current context during compilation - we also support compilation to PTX with `compile_ptx` (https://numba.readthedocs.io/en/latest/cuda-reference/host.html#numba.cuda.compile_ptx), so there isn't guaranteed to be a context to check.

numba/cuda/cudaimpl.py
```diff
@@ -405,6 +405,213 @@ def ptx_popc(context, builder, sig, args):
...
+lower_fp16_binary(stubs.fp16.hadd, 'add')
+lower_fp16_binary(stubs.fp16.hsub, 'sub')
+lower_fp16_binary(stubs.fp16.hmul, 'mul')
+
+
+def lower_fp16_unary(fn, op):
+    @lower(fn, types.float16)
+    def ptx_fp16_unary(context, builder, sig, args):
+        operation_supported = False
+        compute_capability_supported = \
```
gmarkall, line 557: This check will need to be removed too.

numba/cuda/cudaimpl.py
```diff
@@ -405,6 +405,213 @@ def ptx_popc(context, builder, sig, args):
...
+    @lower(fn, types.float16, types.float16)
+    def ptx_fp16_binary(context, builder, sig, args):
+        compute_capability_supported = \
+            cuda.current_context().device.compute_capability >= (5, 3)
+        toolkit_version = cuda.runtime.get_version()
+
+        # 16-bit Floating point binary operations add, sub, and mul were
+        # introduced in PTX ISA 4.2 (Cuda 7.0) and SM 5.3
+
+        if toolkit_version >= (7, 0) and compute_capability_supported:
```
gmarkall, line 532: The minimum toolkit version is 9.2 so we don't need to check for 7.0.

numba/cuda/cudaimpl.py
```diff
@@ -405,6 +405,213 @@ def ptx_popc(context, builder, sig, args):
...
+def lower_fp16_unary(fn, op):
+    @lower(fn, types.float16)
+    def ptx_fp16_unary(context, builder, sig, args):
+        operation_supported = False
+        compute_capability_supported = \
+            cuda.current_context().device.compute_capability >= (5, 3)
+        toolkit_version = cuda.runtime.get_version()
+
+        if op == 'neg':
+            if toolkit_version >= (9,0) and compute_capability_supported:
```
gmarkall, line 562: We also don't need to check for toolkit 9.0.

numba/cuda/cudaimpl.py
```diff
@@ -405,6 +405,213 @@ def ptx_popc(context, builder, sig, args):
...
+@lower(stubs.fp16.hfma, types.float16, types.float16, types.float16)
+def ptx_hfma(context, builder, sig, args):
+    compute_capability_supported = \
+        cuda.current_context().device.compute_capability >= (5, 3)
+    toolkit_version = cuda.runtime.get_version()
+
+    # 16-bit Floating point fused multiply accumulate was
+    # introduced in PTX ISA 4.2 (Cuda 7.0) and SM 5.3
+
+    if toolkit_version >= (7, 0) and compute_capability_supported:
```
gmarkall, line 596: Check for toolkit 7.0 not needed.

gmarkall (changes requested):
Many thanks for the updates.

It seems that there are still some issues in the implementation, which I've added as comments earlier. In general the testing of casts in the CUDA target and this PR is quite minimal (in its existing form, and in this PR). This was sufficient for testing of types that are practically shared with the CPU target, but adding a new type that has no prior implementation tested throughout the rest of the code base and in various user implementations requires a bit more rigour - testing casting to / from all types (within reason) is really needed to have some confidence that the implementation will work in most use cases.

I've thought a bit more about the checks for the compute capability used to raise exceptions when the functionality is unsupported by the current device / toolkit. Unfortuntately we're not guaranteed to have a context available because we could be compiling a function through `compile_ptx` - I think we will have to remove the checks for a supported compute capability in lowering, and just rely on the test skipping to not lower anything unsupported by the toolkit version used with the test.

There are also some other smaller changes I'd suggest (some marked on the diff too)

- Raising `CudaSupportError` shouldn't happen in lowering - this exception is used to signal issues with the driver (c.f. its location / other uses in `cudadrv`). For raising an exception here, it would probably be appropriate to raise something that's a subclass of `numba.core.errors.LoweringError`  - e.g. `CudaLoweringError`, which I'd suggest could be defined in `numba.cuda.errors`.
- There are some changes to numba/tests/test_casting.py that need removing, as it needn't be edited here.
- Adding a test that tests compiling some fp16 operations with `compile_ptx` is probably a good idea, to ensure that it works through that pathway.
- There are some superfluous assignments marked on the diff.

I pushed a more comprehensive set of casting tests to this branch: https://github.com/gmarkall/numba/tree/fp16-support-tests - there is only one commit, https://github.com/gmarkall/numba/commit/a3118470546a55f3d281975e9175744eb05339ac, which exposes the two issues that I mentioned in previous comments - I'd suggest merging this commit into the branch for this PR.

numba/cuda/tests/cudapy/test_casting.py
```diff
@@ -64,12 +104,116 @@ def test_float_to_unsigned(self):
         self.assertEqual(cfunc(3.21),
                          struct.unpack('I', struct.pack('i', 3))[0])
 
-    def test_float_to_complex(self):
+    def test_float16_to_unsigned(self):
+        pyfunc = float_to_unsigned
+        cfunc = self._create_wrapped(pyfunc, np.float16, np.uint32)
+
+        self.assertEqual(cfunc(3.21), pyfunc(3.21))
+        self.assertEqual(cfunc(3.21),
```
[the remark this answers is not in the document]
sklam: I think this might be to get the wraparound for negative numbers in unsigned domain:

```python
>>> struct.unpack("I", struct.pack("i", 3))[0]
3
>>> struct.unpack("I", struct.pack("i", -3))[0]
4294967293
```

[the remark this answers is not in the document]
gmarkall: I think we saw platform-specific differences doing that (https://github.com/numba/numba/pull/1048) because it wasn't defined behaviour.

testhound:
gpuci run tests

numba/cuda/cudaimpl.py
```diff
@@ -405,6 +405,187 @@ def ptx_popc(context, builder, sig, args):
...
+@lower_cast(types.IntegerLiteral, types.float16)
+def literal_int_to_number(context, builder, fromty, toty, val):
+    lit = context.get_constant_generic(
+        builder,
+        fromty.literal_type,
+        fromty.literal_value,
+    )
+    return context.cast(builder, lit, fromty.literal_type, toty)
+
+
```
gmarkall, line 442: I'm not sure this is necessary - I think this can reuse the `integer_to_float16_cast` below:
```suggestion
```

numba/cuda/cudaimpl.py
```diff
@@ -405,6 +405,187 @@ def ptx_popc(context, builder, sig, args):
...
+    cvt2int_fnty = ir.FunctionType(context.get_value_type(toty),
+                                   [ir.IntType(16)])
+    cvt2int_asm = ir.InlineAsm(cvt2int_fnty,
+                               f"cvt.rni.{signedness}{toty.bitwidth}"
+                               f".f16 $0, $1;",
+                               f"={conv_constraint},h")
+    return builder.call(cvt2int_asm, [val])
+
+
+@lower_cast(types.Integer, types.float16)
```
gmarkall, line 498: This can also lower the integer literal case:

```suggestion
@lower_cast(types.Integer, types.float16)
@lower_cast(types.IntegerLiteral, types.float16)
```

numba/cuda/cudamath.py
```diff
@@ -127,5 +127,5 @@ class Math_isnan(ConcreteTemplate):
 class Math_modf(ConcreteTemplate):
     cases = [
         signature(types.UniTuple(types.float64, 2), types.float64),
-        signature(types.UniTuple(types.float32, 2), types.float32)
+        signature(types.UniTuple(types.float32, 2), types.float32),
```
gmarkall, line 130: A stray change seems to have crept in here:
```suggestion
        signature(types.UniTuple(types.float32, 2), types.float32)
```

numba/cuda/testing.py
```diff
@@ -106,6 +107,29 @@ def skip_on_arm(reason):
...
+def skip_unless_toolkit_70(fn):
+    return unittest.skipUnless(cuda_X_or_above(7, 0),
+                               "requires toolkit >= 7.0")(fn)
+
+
+def skip_unless_toolkit_90(fn):
+    return unittest.skipUnless(cuda_X_or_above(9, 0),
+                               "requires toolkit >= 9.0")(fn)
+
+
```
gmarkall, line 127: These are no longer needed.
```suggestion
```

numba/cuda/tests/cudapy/test_intrinsics.py
```diff
@@ -292,6 +343,205 @@ def test_fma_f8(self):
...
+        return cuda_X_or_above(toolkit_major, toolkit_minor)
+
+    def compute_supported(self, compute_major, compute_minor):
+        return cc_X_or_above(compute_major, compute_minor)
+
+    def test_hadd(self):
+        # 16-bit Floating point binary operations add, sub, and mul were
+        # introduced in PTX ISA 4.2 (Cuda 7.0) and SM 5.3
+
+        if self.compute_supported(5,3) and self.toolkit_supported(7,0):
```
gmarkall, line 356: No need to check for toolkit 7.0 here.

numba/cuda/tests/cudapy/test_intrinsics.py
```diff
@@ -292,6 +343,205 @@ def test_fma_f8(self):
...
+            ary = np.zeros(1, dtype=np.float16)
+            arg1 = np.array([3.], dtype=np.float16)
+            arg2 = np.array([4.], dtype=np.float16)
+            compiled[1, 1](ary, arg1, arg2)
+            np.testing.assert_allclose(ary[0], arg1 + arg2)
+        else:
+            self.skipTest("Test requires toolkit >= 7.0 and SM >= 5.3")
+
+    def test_hadd_scalar(self):
+        if self.compute_supported(5,3) and self.toolkit_supported(7,0):
```
gmarkall, line 367: No need to check for toolkit 7.0 here.

numba/cuda/tests/cudapy/test_intrinsics.py
```diff
@@ -292,6 +343,205 @@ def test_fma_f8(self):
...
+            arg2 = np.float16(3.)
+            compiled[1, 1](ary, arg1, arg2)
+            ref = arg1 + arg2
+            np.testing.assert_allclose(ary[0], ref)
+        else:
+            self.skipTest("Test requires toolkit >= 7.0 and SM >= 5.3")
+
+    @skip_on_cudasim('Compilation unsupported in the simulator')
+    def test_hadd_ptx(self):
+        if self.compute_supported(5,3) and self.toolkit_supported(7,0):
```
gmarkall, line 380: No need to check for toolkit 7.0 here.

numba/cuda/tests/cudapy/test_intrinsics.py
```diff
@@ -292,6 +343,205 @@ def test_fma_f8(self):
...
+            args = (f2[:], f2, f2)
+            ptx, _ = compile_ptx(simple_hadd_scalar, args)
+            self.assertIn('add.f16', ptx)
+        else:
+            self.skipTest("Test requires toolkit >= 7.0 and SM >= 5.3")
+
+    def test_hfma(self):
+        if self.compute_supported(5,3) and self.toolkit_supported(7,0):
```
gmarkall, line 388: No need to check for toolkit 7.0 here.

numba/cuda/tests/cudapy/test_intrinsics.py
```diff
@@ -292,6 +343,205 @@ def test_fma_f8(self):
...
+            arg1 = np.array([2.], dtype=np.float16)
+            arg2 = np.array([3.], dtype=np.float16)
+            arg3 = np.array([4.], dtype=np.float16)
+            compiled[1, 1](ary, arg1, arg2, arg3)
+            np.testing.assert_allclose(ary[0], arg1 * arg2 + arg3)
+        else:
+            self.skipTest("Test requires toolkit >= 7.0 and SM >= 5.3")
+
+    def test_hfma_scalar(self):
+        if self.compute_supported(5,3) and self.toolkit_supported(7,0):
```
gmarkall, line 400: No need to check for toolkit 7.0 here.

gmarkall: (there are more instances of this - I'll stop repeating this comment now :-)

numba/cuda/tests/cudapy/test_intrinsics.py
```diff
@@ -292,6 +343,205 @@ def test_fma_f8(self):
...
+    def test_hmul_ptx(self):
+        if self.compute_supported(5,3) and self.toolkit_supported(7,0):
+            args = (f2[:], f2, f2)
+            ptx, _ = compile_ptx(simple_hmul_scalar, args)
+            self.assertIn('mul.f16', ptx)
+        else:
+            self.skipTest("Test requires toolkit >= 7.0 and SM >= 5.3")
+
+    def test_hneg(self):
+        if self.compute_supported(5,3) and self.toolkit_supported(9,0):
```
gmarkall, line 486: Similar for toolkit 9.0 - no need to check for it.

numba/cuda/tests/cudapy/test_intrinsics.py
gmarkall: Also, because we're only compiling PTX for this test, we don't need a device that supports it - `compile_ptx` takes an argument for the compute .capability to use, so instead of skipping these tests, we can always run them and pass `cc=(5, 3)` to `compile_ptx()`: https://numba.readthedocs.io/en/latest/cuda-reference/host.html#numba.cuda.compile_ptx

gmarkall (changes requested):
Many thanks for the updates - I think this is getting closer to finished! There are some comments on the diff, and the main points are:

- Casting from an integer literal isn't tested - it's possible to delete the cast from an integer literal, and I don't see any tests fail when I do this. You have a test of a cast from a floating point literal because of `to_float16()` (which passes without any issue, possibly float literals will implicitly cast to something else that is castable to `float16`).
- We don't need to skip any tests in the test suite due to toolkits 7.0 or 9.0 - these checks can be removed.
- For tests that compile PTX only, we can always run them - since `compile_ptx()` only compiles, we shouldn't skip them due to the CC of the current device, but instead pass `cc=(5, 3)` to `compile_ptx()` - see https://numba.readthedocs.io/en/latest/cuda-reference/host.html#numba.cuda.compile_ptx

numba/cuda/cudaimpl.py
sklam: This can be removed when the change in https://github.com/numba/numba/pull/7460#discussion_r763945372 is added. Otherwise, it will confuse the overload-selector.

numba/cuda/cudaimpl.py
sklam: Need to add a test that runs `cuda.fp16.hmul(x, 2)`. The `2` will need to be implicitly casted to `float16`.

testhound:
gpuci run tests

gmarkall:
/azp run
gpuci run tests

gmarkall (changes requested):
Many thanks for the changes! There are a couple of small points on the diff, summarised as:

- Tests of casting literals to float16 - I think it could be clearer what the tests are doing (the comments are identical to those in other functions and I think they've lost their relevance to the functions as they are).
- The CUDA and reference functions for testing casts of integer literals aren't quite doing the same thing - an explicit cast to `float16` of the literal is needed on the reference version to ensure its behaviour matches the CUDA version.
- It would also be good to add a test of implicitly casting from a float literal - even though this needs no special handling now, the way literals work could change in future and testing this will ensure that support in the CUDA target stays robust.

I also noticed a couple of other things during this round of review:

- I think we've ended up simplifying the skipping logic enough that it's safe to use the decorators like `@skip_unless_cc_53` for skipping the tests in `test_intrinsics`.
- There are some simplifications that can be made to make the lowering implementations more concise and hopefully maintainable (there is some duplication between functions).
- The type check in the `FloatModel` is out of order with respect to the dtype size, which is quite a minor point but I found it a little confusing when I first read it.

Since we're right up against the deadline for 0.55RC rather than adding more comments to the diff and going through another round, I'm going to push the changes I'd suggest to this branch, then hand over to @stuartarchibald for a final review / signoff - I hope this is OK.

gmarkall:
gpuci run tests

stuartarchibald (commented):
Thanks for the patch and fixes @testhound @gmarkall. I've taken a look at the code and it largely looks good. I've made a few comments in line for some minor things I noticed. Thanks again for working on this.

docs/source/cuda-reference/kernel.rst
```diff
@@ -463,6 +463,51 @@ precision parts of the CUDA Toolkit documentation.
    ``cbrt`` and ``cbrtf`` in the C api. Supports float32, and float64 arguments
    only.
 
+16-bit Floating Point Intrinsics
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
```
stuartarchibald, line 468: Does this doc section need a small introduction?

numba/core/itanium_mangler.py
```diff
@@ -78,6 +79,7 @@
     types.int32: 'int',
     types.uint64: 'unsigned long long',
     types.int64: 'long long',
+    types.float16: 'float16',
```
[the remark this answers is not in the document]
gmarkall: I can't see anywhere where `half` is used - have you seen this somewhere? CUDA uses `__half` - perhaps that would be more appropriate?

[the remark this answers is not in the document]
stuartarchibald: I think LLVM refers to it as "half precision". I'm guessing there's no real map to a C name and I presume that this map just needs a unique (and ideally obvious) string for the type to use in mangling?

numba/cuda/cudaimpl.py
```diff
@@ -405,6 +405,116 @@ def ptx_popc(context, builder, sig, args):
...
+lower_fp16_binary(stubs.fp16.hsub, 'sub')
+lower_fp16_binary(stubs.fp16.hmul, 'mul')
+
+
+def lower_fp16_unary(fn, op):
+    @lower(fn, types.float16)
+    def ptx_fp16_unary(context, builder, sig, args):
+        if op == 'abs' and cuda.runtime.get_version() < (10, 2):
+            msg = f"CUDA toolkit >= 10.2 required for {op}(float16)"
+            raise errors.CudaLoweringError(msg)
```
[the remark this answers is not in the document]
gmarkall: Following OOB discussion - we don't generally always know what's going to be done with the PTX we produce (e.g. in `compile_ptx`, what the toolkit / driver version used with the output will be) - so rather than erroring, it would be better to provide an alternative implementation for toolkits that don't support `abs.f16`. In this case we'll mask off the signbit, and we'll still use the runtime version to determine which code to generate, because this will result in code generation consistent with the NVVM version currently being used to go from NVVM IR to PTX.

(Also in general we should not introduce errors that are intended to be user-facing in lowering)

gmarkall:
gpuci run tests

numba/cuda/models.py
```diff
@@ -23,4 +24,18 @@ def __init__(self, dmm, fe_type):
         super().__init__(dmm, fe_type, be_type)
 
 
+@register_default(types.Float)
```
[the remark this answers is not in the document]
gmarkall: I think it will be doing that. In the long term I think we will need to allow targets to have their own `DataModelManager` to avoid this. Perhaps this ought to be brought up in the dev meeting?

docs/source/cuda-reference/kernel.rst
gmarkall: I've added a small introduction.

gmarkall:
gpuci run tests

numba/cuda/cudaimpl.py
```diff
@@ -405,6 +405,116 @@ def ptx_popc(context, builder, sig, args):
...
+
+
+@lower_cast(types.Integer, types.float16)
+@lower_cast(types.IntegerLiteral, types.float16)
+def integer_to_float16_cast(context, builder, fromty, toty, val):
+    bitwidth = fromty.bitwidth
+    signedness = 's' if fromty.signed else 'u'
+    constraint = float16_int_constraint(bitwidth)
+
+    fnty = ir.FunctionType(context.get_value_type(toty),
```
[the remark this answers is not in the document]
gmarkall: Now changed.

numba/core/itanium_mangler.py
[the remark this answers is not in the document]
gmarkall: Now changed to `half`.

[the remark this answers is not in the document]
gmarkall: I agree with your assessment, btw.

numba/cuda/cudaimpl.py
```diff
@@ -405,6 +405,116 @@ def ptx_popc(context, builder, sig, args):
...
+    except KeyError:
+        msg = f"Conversion between float16 and int{bitwidth} unsupported"
+        raise errors.CudaLoweringError(msg)
+
+
+@lower_cast(types.float16, types.Integer)
+def float16_to_integer_cast(context, builder, fromty, toty, val):
+    bitwidth = toty.bitwidth
+    signedness = 's' if toty.signed else 'u'
+    constraint = float16_int_constraint(bitwidth)
```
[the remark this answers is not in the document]
gmarkall: Now swapped, and also another instance of the same has been swapped.

numba/cuda/cudaimpl.py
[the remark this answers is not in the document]
gmarkall: This is now implemented.

stuartarchibald (dismissed):
Thanks for the updates to address the review @gmarkall, patch looks good.

docs/source/cuda-reference/kernel.rst
stuartarchibald: Thanks

numba/cuda/models.py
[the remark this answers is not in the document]
stuartarchibald: I think it'd be good to raise that at the next meeting, thanks.

numba/cuda/cudaimpl.py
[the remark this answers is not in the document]
stuartarchibald: Thanks, implementation looks good.

numba/cuda/cudaimpl.py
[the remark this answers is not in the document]
stuartarchibald: Thanks.

numba/cuda/cudaimpl.py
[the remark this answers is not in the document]
stuartarchibald: Noted, thanks.

numba/core/itanium_mangler.py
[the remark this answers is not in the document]
stuartarchibald: Great, thanks for doing that.

esc:
`Build numba_smoketest_cuda_yaml_108 has started`

gmarkall (approved):
LGTM pending buildfarm.

esc:
Interestingly enough this breaks across all Windows builds with the following errors:

```
ERROR: test_literal_to_float16 (numba.cuda.tests.cudapy.test_casting.TestCasting) (func=<function cuda_int_literal_to_float16 at 0x000002332EB61550>)
  File "F:\ci_envs\64\Miniconda3\envs\testenv_03f65b4d-5932-4ea8-9027-cc24ca0c5d36\lib\site-packages\numba\cuda\cudadrv\driver.py", line 2682, in add_ptx
  File "F:\ci_envs\64\Miniconda3\envs\testenv_03f65b4d-5932-4ea8-9027-cc24ca0c5d36\lib\site-packages\numba\cuda\cudadrv\driver.py", line 319, in safe_cuda_api_call
  File "F:\ci_envs\64\Miniconda3\envs\testenv_03f65b4d-5932-4ea8-9027-cc24ca0c5d36\lib\site-packages\numba\cuda\cudadrv\driver.py", line 384, in _check_ctypes_error

  File "F:\ci_envs\64\Miniconda3\envs\testenv_03f65b4d-5932-4ea8-9027-cc24ca0c5d36\lib\site-packages\numba\cuda\tests\cudapy\test_casting.py", line 172, in test_literal_to_float16
    self.assertEqual(cfunc(321), hostfunc(321))
  File "F:\ci_envs\64\Miniconda3\envs\testenv_03f65b4d-5932-4ea8-9027-cc24ca0c5d36\lib\site-packages\numba\cuda\tests\cudapy\test_casting.py", line 105, in wrapper_fn
    cuda_wrapper_fn[1, 1](argarray, resarray)
  File "F:\ci_envs\64\Miniconda3\envs\testenv_03f65b4d-5932-4ea8-9027-cc24ca0c5d36\lib\site-packages\numba\cuda\compiler.py", line 727, in __call__
  File "F:\ci_envs\64\Miniconda3\envs\testenv_03f65b4d-5932-4ea8-9027-cc24ca0c5d36\lib\site-packages\numba\cuda\compiler.py", line 915, in call
  File "F:\ci_envs\64\Miniconda3\envs\testenv_03f65b4d-5932-4ea8-9027-cc24ca0c5d36\lib\site-packages\numba\cuda\compiler.py", line 923, in _compile_for_args
  File "F:\ci_envs\64\Miniconda3\envs\testenv_03f65b4d-5932-4ea8-9027-cc24ca0c5d36\lib\site-packages\numba\cuda\compiler.py", line 1093, in compile
    kernel.bind()
  File "F:\ci_envs\64\Miniconda3\envs\testenv_03f65b4d-5932-4ea8-9027-cc24ca0c5d36\lib\site-packages\numba\cuda\compiler.py", line 466, in bind
  File "F:\ci_envs\64\Miniconda3\envs\testenv_03f65b4d-5932-4ea8-9027-cc24ca0c5d36\lib\site-packages\numba\cuda\codegen.py", line 201, in get_cufunc
  File "F:\ci_envs\64\Miniconda3\envs\testenv_03f65b4d-5932-4ea8-9027-cc24ca0c5d36\lib\site-packages\numba\cuda\codegen.py", line 174, in get_cubin
    linker.add_ptx(ptx.encode())
  File "F:\ci_envs\64\Miniconda3\envs\testenv_03f65b4d-5932-4ea8-9027-cc24ca0c5d36\lib\site-packages\numba\cuda\cudadrv\driver.py", line 2685, in add_ptx
ptxas application ptx input, line 54; error   : Feature 'f16 arithemetic and compare instructions' requires .target sm_53 or higher

ERROR: test_literal_to_float16 (numba.cuda.tests.cudapy.test_casting.TestCasting) (func=<function cuda_float_literal_to_float16 at 0x000002332EB61670>)
  File "F:\ci_envs\64\Miniconda3\envs\testenv_03f65b4d-5932-4ea8-9027-cc24ca0c5d36\lib\site-packages\numba\cuda\cudadrv\driver.py", line 2682, in add_ptx
  File "F:\ci_envs\64\Miniconda3\envs\testenv_03f65b4d-5932-4ea8-9027-cc24ca0c5d36\lib\site-packages\numba\cuda\cudadrv\driver.py", line 319, in safe_cuda_api_call
  File "F:\ci_envs\64\Miniconda3\envs\testenv_03f65b4d-5932-4ea8-9027-cc24ca0c5d36\lib\site-packages\numba\cuda\cudadrv\driver.py", line 384, in _check_ctypes_error

  File "F:\ci_envs\64\Miniconda3\envs\testenv_03f65b4d-5932-4ea8-9027-cc24ca0c5d36\lib\site-packages\numba\cuda\tests\cudapy\test_casting.py", line 172, in test_literal_to_float16
    self.assertEqual(cfunc(321), hostfunc(321))
  File "F:\ci_envs\64\Miniconda3\envs\testenv_03f65b4d-5932-4ea8-9027-cc24ca0c5d36\lib\site-packages\numba\cuda\tests\cudapy\test_casting.py", line 105, in wrapper_fn
    cuda_wrapper_fn[1, 1](argarray, resarray)
  File "F:\ci_envs\64\Miniconda3\envs\testenv_03f65b4d-5932-4ea8-9027-cc24ca0c5d36\lib\site-packages\numba\cuda\compiler.py", line 727, in __call__
  File "F:\ci_envs\64\Miniconda3\envs\testenv_03f65b4d-5932-4ea8-9027-cc24ca0c5d36\lib\site-packages\numba\cuda\compiler.py", line 915, in call
  File "F:\ci_envs\64\Miniconda3\envs\testenv_03f65b4d-5932-4ea8-9027-cc24ca0c5d36\lib\site-packages\numba\cuda\compiler.py", line 923, in _compile_for_args
  File "F:\ci_envs\64\Miniconda3\envs\testenv_03f65b4d-5932-4ea8-9027-cc24ca0c5d36\lib\site-packages\numba\cuda\compiler.py", line 1093, in compile
    kernel.bind()
  File "F:\ci_envs\64\Miniconda3\envs\testenv_03f65b4d-5932-4ea8-9027-cc24ca0c5d36\lib\site-packages\numba\cuda\compiler.py", line 466, in bind
  File "F:\ci_envs\64\Miniconda3\envs\testenv_03f65b4d-5932-4ea8-9027-cc24ca0c5d36\lib\site-packages\numba\cuda\codegen.py", line 201, in get_cufunc
  File "F:\ci_envs\64\Miniconda3\envs\testenv_03f65b4d-5932-4ea8-9027-cc24ca0c5d36\lib\site-packages\numba\cuda\codegen.py", line 174, in get_cubin
    linker.add_ptx(ptx.encode())
  File "F:\ci_envs\64\Miniconda3\envs\testenv_03f65b4d-5932-4ea8-9027-cc24ca0c5d36\lib\site-packages\numba\cuda\cudadrv\driver.py", line 2685, in add_ptx
ptxas application ptx input, line 55; error   : Feature 'f16 arithemetic and compare instructions' requires .target sm_53 or higher

Ran 1257 tests in 231.363s

FAILED (errors=2, skipped=53, expected failures=7)
```

gmarkall:
gpuci run tests

seibert:
This shows a limitation of the internal Numba Windows test server, which currently has a `sm_35` card (Tesla K40c) in it.  We need to upgrade that server to a newer card anyway since NVIDIA has already dropped support for this card in the latest CUDA release.  For now, I think we need to mark these tests to skip on older architectures.

gmarkall:
I've pushed a fix that skips the test if the CC is not new enough. I think we should consider dropping support for CC less than at least 5.2 after Numba 0.55.

stuartarchibald (approved):
Thanks for addressing the issue with CC < 5.3.

sklam:
BFID `numba_smoketest_cuda_yaml_109`

PR #7479, opened 2021-10-11
CUDA: Print format string and warn for > 32 print() args

gmarkall:
CUDA vprintf prints garbage for arguments past the 32nd (a similar limitation on `printf` is outlined in the CUDA C++ Programming Guide at https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#limitations).

This commit matches the C/C++ behavior in Python by emitting the format string when more than 32 arguments are provided, instead of printing garbage results. A warning is also produced to explain the behavior that users will see.


gmarkall:
gpuci run tests

sklam:
BFID `numba_smoketest_cuda_yaml_101`

sklam:
build farm passed

sklam (approved):

PR #7498, opened 2021-10-20
refactor  with-detection logic

esc:
As title

### DIFF
--- a/numba/core/transforms.py
+++ b/numba/core/transforms.py
@@ -501,11 +501,7 @@ def find_setupwiths(blocks):
             return isinstance(stmt, ir.Return)
 
         def is_pop_block(stmt):
-            try:
-                if hasattr(stmt, "value"):
-                    return str(stmt.value).startswith("POP_BLOCK_INFO")
-            except KeyError:
-                return False
+            return isinstance(stmt, ir.PopBlock)
 
         cfg = compute_cfg_from_blocks(blocks)
         sus_setups, sus_pops = [], []
--- a/numba/core/transforms.py
+++ b/numba/core/transforms.py
@@ -572,16 +572,32 @@ def find_setupwiths(blocks):
         for p,w in pop_block_to_setup_map.items():
             yield w, p
 
-    def previously_occurred(start, known_ranges):
+    known_ranges = []
+
+    def within_known_range(start, end, known_ranges):
         for a, b in known_ranges:
-            if start >= a and start < b:
+            # FIXME: this should be a comparison in topological order, right
+            # now we are comparing the integers of the blocks, stuff probably
+            # works by accident.
+            if start > a and end < b:
                 return True
         return False
 
-    known_ranges = []
+    setup_to_pop_block_map = {}
     for s, e in sorted(find_ranges(blocks)):
-        if not previously_occurred(s, known_ranges):
-            assert s in blocks, 'starting offset is not a label'
+        # Look for setup_withs that have multiple pop_blocks each with a
+        # different target.
+        if s not in setup_to_pop_block_map:
+            setup_to_pop_block_map[s] = e
+        else:
+            if setup_to_pop_block_map[s] != e:
+                raise errors.CompilerError(
+                    "Does not support with-context that contain branches "
+                    "(i.e. break/return/raise) that can leave the with-context. "
+                )
+        # Elminate all withs contained within withs, we are only interested in
+        # the outermost with statements so we can lift them.
+        if not within_known_range(s, e, known_ranges):
             known_ranges.append((s, e))
 
     return known_ranges
--- a/numba/core/transforms.py
+++ b/numba/core/transforms.py
@@ -486,10 +486,8 @@ def find_setupwiths(blocks):
         def is_setup_with(stmt):
             return isinstance(stmt, ir.EnterWith)
 
-
         def is_branch(stmt):
             return isinstance(stmt, ir.Branch)
-
 
         def is_terminator(stmt):
             return isinstance(stmt, ir.Terminator)
--- a/numba/core/transforms.py
+++ b/numba/core/transforms.py
@@ -579,13 +579,13 @@ def find_setupwiths(blocks):
                     # raise detected before pop_block
                     if is_raise(stmt):
                             raise errors.CompilerError(
-                                'unsupported controlflow due to return/raise '
+                                'unsupported controlflow due to raise '
                                 'statements inside with block'
                                 )
                     # special case 3.7, return before POP_BLOCK
                     if PYVERSION < (3, 8) and is_return(stmt):
                             raise errors.CompilerError(
-                                'unsupported controlflow due to return/raise '
+                                'unsupported controlflow due to return '
                                 'statements inside with block'
                                 )
                     # if a pop_block, process it
@@ -596,7 +596,7 @@ def find_setupwiths(blocks):
                         target_block = blocks[pop_block_targets[0]]
                         if not target_block.terminator.get_targets():
                             raise errors.CompilerError(
-                                'unsupported controlflow due to return/raise '
+                                'unsupported controlflow due to return '
                                 'statements inside with block'
                                 )
                         pop_block_to_setup_map[pop_block_targets[0]] = setup_block
@@ -637,7 +637,7 @@ def find_setupwiths(blocks):
             if e not in blocks:
                 # this's possible if there's an exit path in the with-block
                 raise errors.CompilerError(
-                    'unsupported controlflow due to return/raise '
+                    'unsupported controlflow due to foo'
                     'statements inside with block'
                     )
             assert s in blocks, 'starting offset is not a label'
--- a/numba/core/transforms.py
+++ b/numba/core/transforms.py
@@ -544,6 +544,9 @@ def find_setupwiths(blocks):
         def is_raise(stmt):
             return isinstance(stmt, ir.Raise)
 
+        def is_return(stmt):
+            return isinstance(stmt, ir.Return)
+
         def is_pop_block(stmt):
             try:
                 if hasattr(stmt, "value"):
@@ -575,6 +578,12 @@ def find_setupwiths(blocks):
                 for stmt in blocks[block].body:
                     # raise detected before pop_block
                     if is_raise(stmt):
+                            raise errors.CompilerError(
+                                'unsupported controlflow due to return/raise '
+                                'statements inside with block'
+                                )
+                    # special case 3.7, return before POP_BLOCK
+                    if PYVERSION < (3, 8) and is_return(stmt):
                             raise errors.CompilerError(
                                 'unsupported controlflow due to return/raise '
                                 'statements inside with block'
--- a/numba/core/transforms.py
+++ b/numba/core/transforms.py
@@ -503,6 +503,8 @@ def find_setupwiths(blocks):
 
         cfg = compute_cfg_from_blocks(blocks)
         sus_setups, sus_pops = [], []
+        # traverse the cfg and collect all suspected SETUP_WITH and POP_BLOCK
+        # statements so that we can iterate over them
         for label, block in blocks.items():
             for stmt in block.body:
                 if is_setup_with(stmt):
--- a/numba/core/transforms.py
+++ b/numba/core/transforms.py
@@ -545,9 +545,11 @@ def find_setupwiths(blocks):
             return isinstance(stmt, ir.Raise)
 
         def is_pop_block(stmt):
-            if hasattr(stmt, "value"):
-                return str(stmt.value).startswith("POP_BLOCK_INFO")
-
+            try:
+                if hasattr(stmt, "value"):
+                    return str(stmt.value).startswith("POP_BLOCK_INFO")
+            except KeyError:
+                return False
 
         cfg = compute_cfg_from_blocks(blocks)
         sus_setups, sus_pops = [], []
--- a/numba/core/transforms.py
+++ b/numba/core/transforms.py
@@ -528,13 +528,13 @@ def find_setupwiths(blocks):
                     # raise detected before pop_block
                     if is_raise(stmt):
                             raise errors.CompilerError(
-                                'unsupported controlflow due to raise '
+                                'unsupported control flow due to raise '
                                 'statements inside with block'
                                 )
                     # special case 3.7, return before POP_BLOCK
                     if PYVERSION < (3, 8) and is_return(stmt):
                             raise errors.CompilerError(
-                                'unsupported controlflow due to return '
+                                'unsupported control flow due to return '
                                 'statements inside with block'
                                 )
                     # if a pop_block, process it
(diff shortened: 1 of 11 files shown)


sklam:
@esc, i found some cases that the new with-handling fail to report unsupported control-flow. See https://gist.github.com/sklam/a373bf4d601fd27a93a330f66770eb63#gistcomment-3947520

The `before.log` is from 944dceee2136ab55b595319aa19611e3699a32a5.  The `after.log` is from https://github.com/numba/numba/pull/7498/commits/fd113423eeb33587a5d84630abd90ff668e781ae

esc:
@sklam this PR is now green and ready for review.

stuartarchibald:
@esc in https://github.com/stuartarchibald/numba/commit/037c28a3d37e8d011e843e191fe409be8cb52ebd I've refactored this patch slightly to make the `pop_block` opcode interpret as a new `ir.PopBlock` class which means it's a statement and not an expr, this eliminating the need to provide dummy types/storage etc. What do you think?

numba/core/transforms.py
```diff
@@ -477,73 +476,130 @@ def _cfg_nodes_in_region(cfg, region_begin, region_end):
...
+            if setup_to_pop_block_map[s] != e:
                 raise errors.CompilerError(
-                    'unsupported controlflow due to return/raise '
-                    'statements inside with block'
-                    )
-            assert s in blocks, 'starting offset is not a label'
+                    "Does not support with-context that contain branches "
+                    "(i.e. break/return/raise) that can leave the with-context. "
+                )
+        # Elminate all withs contained within withs, we are only interested in
```
stuartarchibald, line 596:
```suggestion
        # Eliminate all withs contained within withs, we are only interested in
```

numba/core/transforms.py
```diff
@@ -477,73 +476,130 @@ def _cfg_nodes_in_region(cfg, region_begin, region_end):
...
+                                'statements inside with block'
+                                )
+                    # if a pop_block, process it
+                    if is_pop_block(stmt) and block in sus_pops:
+                        # record the jump target of this block belonging to this setup
+                        pop_block_targets = blocks[block].terminator.get_targets()
+                        if len(pop_block_targets) != 1:
+                            raise errors.CompilerError(
+                                "Does not support with-context that contain branches "
+                                "(i.e. break/return/raise) that can leave the with-context. "
```
stuartarchibald, line 547:
```suggestion
                                "unsupported control flow: with-context contains branches "
                                "(i.e. break/return/raise) that can leave the with-context. "
```

numba/core/transforms.py
```diff
@@ -477,73 +476,130 @@ def _cfg_nodes_in_region(cfg, region_begin, region_end):
...
+                    # raise detected before pop_block
+                    if is_raise(stmt):
+                            raise errors.CompilerError(
+                                'unsupported controlflow due to raise '
+                                'statements inside with block'
+                                )
+                    # special case 3.7, return before POP_BLOCK
+                    if PYVERSION < (3, 8) and is_return(stmt):
+                            raise errors.CompilerError(
+                                'unsupported controlflow due to return '
```
stuartarchibald, line 537:
```suggestion
                                'unsupported control flow due to return '
```

numba/core/transforms.py
```diff
@@ -477,73 +476,130 @@ def _cfg_nodes_in_region(cfg, region_begin, region_end):
...
+            while to_visit:
+                # get whatever is next and record that we have seen it
+                block = to_visit.pop()
+                seen.append(block)
+                # go through the body of the block, looking for statements
+                for stmt in blocks[block].body:
+                    # raise detected before pop_block
+                    if is_raise(stmt):
+                            raise errors.CompilerError(
+                                'unsupported controlflow due to raise '
```
stuartarchibald, line 531:
```suggestion
                                'unsupported control flow due to raise '
```

numba/core/transforms.py
```diff
@@ -477,73 +476,130 @@ def _cfg_nodes_in_region(cfg, region_begin, region_end):
...
+                        pop_block_targets = blocks[block].terminator.get_targets()
+                        if len(pop_block_targets) != 1:
+                            raise errors.CompilerError(
+                                "Does not support with-context that contain branches "
+                                "(i.e. break/return/raise) that can leave the with-context. "
+                            )
+                        target_block = blocks[pop_block_targets[0]]
+                        if not target_block.terminator.get_targets():
+                            raise errors.CompilerError(
+                                'unsupported controlflow due to return '
```
stuartarchibald, line 552:
```suggestion
                                'unsupported control flow due to return '
```

numba/core/transforms.py
```diff
@@ -477,73 +476,130 @@ def _cfg_nodes_in_region(cfg, region_begin, region_end):
...
+                    if is_pop_block(stmt) and block in sus_pops:
+                        # record the jump target of this block belonging to this setup
+                        pop_block_targets = blocks[block].terminator.get_targets()
+                        if len(pop_block_targets) != 1:
+                            raise errors.CompilerError(
+                                "Does not support with-context that contain branches "
+                                "(i.e. break/return/raise) that can leave the with-context. "
+                            )
+                        target_block = blocks[pop_block_targets[0]]
+                        if not target_block.terminator.get_targets():
```
stuartarchibald, line 550: Would it be clearer to check if the target block terminator is an instance of an `ir.Return`, i.e.
```python
if is_return(target_block.terminator):
    ...
```
?

numba/core/transforms.py
```diff
@@ -477,73 +476,130 @@ def _cfg_nodes_in_region(cfg, region_begin, region_end):
...
+                    # if a pop_block, process it
+                    if is_pop_block(stmt) and block in sus_pops:
+                        # record the jump target of this block belonging to this setup
+                        pop_block_targets = blocks[block].terminator.get_targets()
+                        if len(pop_block_targets) != 1:
+                            raise errors.CompilerError(
+                                "Does not support with-context that contain branches "
+                                "(i.e. break/return/raise) that can leave the with-context. "
+                            )
+                        target_block = blocks[pop_block_targets[0]]
```
stuartarchibald, line 549:
```suggestion
                        pop_block_target = pop_block_targets[0]
                        target_block = blocks[pop_block_target]
```

numba/core/transforms.py
```diff
@@ -477,73 +476,130 @@ def _cfg_nodes_in_region(cfg, region_begin, region_end):
...
+                        if not target_block.terminator.get_targets():
+                            raise errors.CompilerError(
+                                'unsupported controlflow due to return '
+                                'statements inside with block'
+                                )
+                        setup_with_to_pop_blocks_map[setup_block].add(pop_block_targets[0])
```
stuartarchibald, line 555:
```suggestion
                        setup_with_to_pop_blocks_map[setup_block].add(pop_block_target)
```

numba/core/transforms.py
```diff
@@ -477,73 +476,126 @@ def _cfg_nodes_in_region(cfg, region_begin, region_end):
...
+                        # remove the block from blocks to be matched
+                        sus_pops.remove(block)
+                        # stop looking, we have reached the frontier
+                        break
+                    # if we are still here, by the terminator block,
```
stuartarchibald, line 560:
```suggestion
                    # if we are still here, by the block terminator,
```

numba/core/transforms.py
```diff
@@ -477,73 +476,126 @@ def _cfg_nodes_in_region(cfg, region_begin, region_end):
...
+                    # add all it's targets to the to_visit stack, unless we
```
stuartarchibald, line 561:
```suggestion
                    # add all its targets to the to_visit stack, unless we
```

numba/core/transforms.py
```diff
@@ -477,73 +476,126 @@ def _cfg_nodes_in_region(cfg, region_begin, region_end):
...
+        def is_setup_with(stmt):
+            return isinstance(stmt, ir.EnterWith)
+
+
+        def is_branch(stmt):
+            return isinstance(stmt, ir.Branch)
+
+
+        def is_terminator(stmt):
+            return isinstance(stmt, ir.Terminator)
```
stuartarchibald, line 495: Given the use of this function later to determine which targets to add to the `to_visit` stack, is it really looking for **_any_** terminator, or just `Jump` and `Branch`? I suppose given those block terminators which terminate control flow, like `Raise`, have an empty target list it probably makes no difference.

numba/core/transforms.py
```diff
@@ -477,73 +476,126 @@ def _cfg_nodes_in_region(cfg, region_begin, region_end):
...
+
+
+    known_ranges = []
+
+    def within_known_range(start, end, known_ranges):
         for a, b in known_ranges:
-            if start >= a and start < b:
+            # FIXME: this should be a comparison in topological order, right
+            # now we are comparing the integers of the blocks, stuff probably
+            # works by accident.
```
stuartarchibald, line 579: Is this statement still true?

numba/core/transforms.py
```diff
@@ -477,73 +476,126 @@ def _cfg_nodes_in_region(cfg, region_begin, region_end):
...
+        if s not in setup_to_pop_block_map:
+            setup_to_pop_block_map[s] = e
+        else:
+            if setup_to_pop_block_map[s] != e:
                 raise errors.CompilerError(
-                    'unsupported controlflow due to return/raise '
-                    'statements inside with block'
-                    )
-            assert s in blocks, 'starting offset is not a label'
+                    "Does not support with-context that contain branches "
```
stuartarchibald, line 593:
```suggestion
                    ""unsupported control flow: with-context contains branches "
```

numba/core/transforms.py
```diff
@@ -477,73 +476,126 @@ def _cfg_nodes_in_region(cfg, region_begin, region_end):
...
+
+    known_ranges = []
+
+    def within_known_range(start, end, known_ranges):
         for a, b in known_ranges:
-            if start >= a and start < b:
+            # FIXME: this should be a comparison in topological order, right
+            # now we are comparing the integers of the blocks, stuff probably
+            # works by accident.
+            if start > a and end < b:
```
stuartarchibald, line 580: This is now strictly `>` and `<`, i.e. no equality, is this because of a change in the block emission, i.e. starts and ends have to be in their own unique blocks?

numba/tests/test_withlifting.py
```diff
@@ -577,7 +597,7 @@ def foo(x):
         with self.assertRaises(errors.CompilerError) as raises:
             cfoo(x)
         self.assertIn(
-            ('unsupported controlflow due to return/raise statements inside '
+            ('unsupported controlflow due to raise statements inside '
```
stuartarchibald, line 600: It's good that the difference can now be detected.

stuartarchibald (commented):
Many thanks for all your efforts on this patch @esc. This is a really important refactor for helping accommodate Python 3.10 bytecode changes (inlined returns!). I've reviewed the code and it looks good, I particularly appreciated the inline comments in the `find_setupwiths` function rewrite. I've left a some comments inline, most of which are typo level things, but there are a few which might require further consideration. Thanks again!

numba/core/transforms.py
```diff
@@ -477,73 +476,130 @@ def _cfg_nodes_in_region(cfg, region_begin, region_end):
...
-    def previously_occurred(start, known_ranges):
+
+        def is_setup_with(stmt):
+            return isinstance(stmt, ir.EnterWith)
+
+
+        def is_branch(stmt):
+            return isinstance(stmt, ir.Branch)
+
+
```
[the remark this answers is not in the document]
esc: done in ff17b1ebcbf27810ab6557f57fffbc08453296d9

numba/core/transforms.py
```diff
@@ -477,73 +476,130 @@ def _cfg_nodes_in_region(cfg, region_begin, region_end):
...
+                    return str(stmt.value).startswith("POP_BLOCK_INFO")
+            except KeyError:
+                return False
+
+        cfg = compute_cfg_from_blocks(blocks)
+        sus_setups, sus_pops = [], []
+        for label, block in blocks.items():
+            for stmt in block.body:
+                if is_setup_with(stmt):
+                    sus_setups.append(label)
```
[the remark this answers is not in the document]
esc: yes, they can both be sets.

numba/core/transforms.py
esc: this may break tests, but probably a good idea.

numba/core/transforms.py
```diff
@@ -477,73 +476,130 @@ def _cfg_nodes_in_region(cfg, region_begin, region_end):
...
+                if is_pop_block(stmt):
+                    sus_pops.append(label)
```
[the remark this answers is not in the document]
esc: done in faab51edcf1fc809292c85c3ccb5fd9cf1174b60

numba/core/transforms.py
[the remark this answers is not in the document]
esc: fixed in cd94d0918a7e4c4c0c4b527b4857ef25e8d9cd5f

numba/core/transforms.py
esc: done in a1eb14a0f2f044623366036177af991cc9528a4d

numba/tests/test_withlifting.py
esc: it is, isn't it!

numba/core/transforms.py
esc: It may very well be, I am not certain. I have been unable to find python code to break this so far.

esc: for many cases topological order equals linear ordering, so this is probably why this works.

numba/core/transforms.py
esc: Yes, it is a bit contrived. The previous representation checked if `start` was between `a` and `b`. Because having a `SETUP_WITH` inside the bounds (as determined by the offset argument to `SETUP_WITH`). Now that we have the exact locations of `SETUP_WITH` AND `POP_BLOCK` we can check that a given `SETUP_WITH`/`POP_BLOCK` pair  is contained entirely within the bounds of an enclosing pair, i.e. nested. 

Anyway, to answer the original question, we never have duplicate `SETUP_WITH`s or `POP_BLOCKS` because there will only ever be one per block. This means, we can use strict comparison.

stuartarchibald: Thanks for the explanation, this makes sense with respect to the proposed changes.

numba/core/transforms.py
stuartarchibald: I think it is probably still true and is being "got away with" because of the noted common case of topological order is linear order in these sorts of regions. I think this comment should stay as it, it's a good reminder that the check isn't quite as robust as it perhaps ought to be.

esc: I agree

numba/core/transforms.py
stuartarchibald: ok, let's leave it as-is.

stuartarchibald (approved):

esc:
`Build numba_smoketest_cpu_yaml_67 has started`

esc:
> `Build numba_smoketest_cpu_yaml_67 has started`
Screen Shot 2021-11-24 at 21 00 07

esc:
@stuartarchibald thank you for the review!

PR #7506, opened 2021-10-21
Enhance Flags mangling/demangling

sklam:
## changes
demangle:
- to deal with `_` escaped string
- to become a classmethod

mangling:
- include more values into the zdict for shorter output

## To manually demangle Numba symbol and extract the compilation flags:

1. mangled symbol `_ZN5numba7cpython7unicode11integer_str12_3clocals_3e9impl_2425B40c8tJTIeFCjyCbUFRqqOAK_2f6h0phxApbQBAA_3dEi` (probably from debugger)
2.  use `c++filt` to demangle:
     ```bash
     % c++filt -n _ZN5numba7cpython7unicode11integer_str12_3clocals_3e9impl_2425B40c8tJTIeFCjyCbUFRqqOAK_2f6h0phxApbQBAA_3dEi
     numba::cpython::unicode::integer_str::_3clocals_3e::impl_2425[abi:c8tJTIeFCjyCbUFRqqOAK_2f6h0phxApbQBAA_3d](int)
3. extract the part in `[abi:...]`
4. put it through:
```python
>>> from numba.core.compiler import Flags
>>> Flags.demangle("c8tJTIeFCjyCbUFRqqOAK_2f6h0phxApbQBAA_3d")
b'Flags(enable_looplift=True, enable_pyobject_looplift=True, no_cpython_wrapper=True)'
```

### DIFF
--- a/numba/core/targetconfig.py
+++ b/numba/core/targetconfig.py
@@ -2,6 +2,7 @@
 This module contains utils for manipulating target configurations such as
 compiler flags.
 """
+import re
 import zlib
 import base64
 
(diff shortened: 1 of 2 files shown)


stuartarchibald (approved):

PR #7515, opened 2021-10-28
Make *var work for randn for parfors.

DrTodd13:
Sometimes, the shape of an output is inferred from its varargs and not the regular args.  Randn is one such example.  Randn gets converted to parfor but the varargs also need to be cleared along with args so that randn gets a single-item within each loop iteration.

I included another related fix here where a UniTuple(int32 x 2) was coming in as an argument but the size variable created was int64 and this led to 32-64-bit assignment problems.  I've passed the typ to the code that extracts sizes as individual variables so that the type created there can match the input type.

This is not based on any pending issue but something I ran across.

### DIFF
--- a/numba/parfors/array_analysis.py
+++ b/numba/parfors/array_analysis.py
@@ -1568,7 +1568,7 @@ class ArrayAnalysis(object):
         # TODO: getattr of npytypes.Record
         if expr.attr == "T" and self._isarray(expr.value.name):
             return self._analyze_op_call_numpy_transpose(
-                scope, equiv_set, expr.loc, [expr.value], {}
+                scope, equiv_set, expr.loc, [expr.value], {}, None
             )
         elif expr.attr == "shape":
             shape = equiv_set.get_shape(expr.value)
@@ -2233,8 +2233,10 @@ class ArrayAnalysis(object):
             mod_name = "numpy"
             # Remember that args and expr.args don't alias.
             added_mod_name = True
+            varargs = None
         else:
             args = expr.args
+            varargs = expr.vararg
         fname = "_analyze_op_call_{}_{}".format(mod_name, fname).replace(
             ".", "_"
         )
@@ -2253,6 +2255,7 @@ class ArrayAnalysis(object):
                 loc=expr.loc,
                 args=args,
                 kws=dict(expr.kws),
+                varargs=varargs
             )
             # We want the ability for function fn to modify arguments.
             # If args and expr.args don't alias then we need the extra
@@ -2262,7 +2265,7 @@ class ArrayAnalysis(object):
                 expr.args = args[1:]
             return result
 
-    def _analyze_op_call_builtins_len(self, scope, equiv_set, loc, args, kws):
+    def _analyze_op_call_builtins_len(self, scope, equiv_set, loc, args, kws, varargs):
         # python 3 version of len()
         require(len(args) == 1)
         var = args[0]
@@ -2272,13 +2275,13 @@ class ArrayAnalysis(object):
         return ArrayAnalysis.AnalyzeResult(shape=shape[0], rhs=shape[0])
 
     def _analyze_op_call_numba_parfors_array_analysis_assert_equiv(
-        self, scope, equiv_set, loc, args, kws
+        self, scope, equiv_set, loc, args, kws, varargs
     ):
         equiv_set.insert_equiv(*args[1:])
         return None
 
     def _analyze_op_call_numba_parfors_array_analysis_wrap_index(
-        self, scope, equiv_set, loc, args, kws
+        self, scope, equiv_set, loc, args, kws, varargs
     ):
         """ Analyze wrap_index calls added by a previous run of
             Array Analysis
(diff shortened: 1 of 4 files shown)


guilhermeleobas (approved):
I'm not that familiar with how the parallel accelerator code works, but I checked the implementation and seems to make sense.

numba/parfors/array_analysis.py
```diff
@@ -3148,7 +3185,19 @@ def _make_assert_equiv(self, scope, loc, equiv_set, _args, names=None):
...
+        if isinstance(typ, types.containers.UniTuple):
+            return typ.dtype
+        elif isinstance(typ, types.ArrayCompatible):
+            return types.intp
+        elif isinstance(typ, types.Number):
+            return typ
+        else:
+            if config.DEBUG_ARRAY_OPT >= 1:
+                print(1, "Unhandled case in _get_typ_part:", typ, i)
+            return types.intp
```
guilhermeleobas, line 3198: In case the `else` statement is executed, the 32-64 bit assignment problem results in a compilation error?

DrTodd13: Most likely no.  The "else" part is essentially what was happening in the prior version of the code which worked in most cases.  I tried to think of what other cases there might be that would fail with this default behavior and put those checks in the preceding lines.  I can't say 100% that there aren't other cases and that's why I added the debug code on line 3197 where in the future if there are cases that fail for a similar reason then we can catch and add checks for those above as well.

numba/parfors/array_analysis.py
```diff
@@ -2272,13 +2277,13 @@ def _analyze_op_call_builtins_len(self, scope, equiv_set, loc, args, kws):
         return ArrayAnalysis.AnalyzeResult(shape=shape[0], rhs=shape[0])
 
     def _analyze_op_call_numba_parfors_array_analysis_assert_equiv(
-        self, scope, equiv_set, loc, args, kws
+        self, scope, equiv_set, loc, args, kws, varargs
```
stuartarchibald, line 2280: At some point perhaps these `_analyze_op*` methods could be refactored to just take a dictionary? Probably 80% of the code change in this patch is adding the same extra arg to many methods?

stuartarchibald: Also, methods like `_analyze_numpy_array_like` do not have this extra arg as they are presumably never used or called in that way, however this makes it quite hard to work out whether that's a mistake or by design.

numba/tests/test_parfors.py
```diff
@@ -1018,6 +1018,24 @@ def test_impl(n):
         self.assertEqual(parfor_output, py_output)
         self.assertEqual(countParfors(test_impl, (types.int64, )), 0)
 
+    def test_stararg(self):
+        def test_impl(n):
+            A = np.random.randn(*n)
```
stuartarchibald, line 1023:
```suggestion
            np.random.seed(0)
            A = np.random.randn(*n)
```
suggest setting the seed for reproducible testing.

numba/tests/test_parfors.py
```diff
@@ -1018,6 +1018,24 @@ def test_impl(n):
...
+            B = np.sum(A)
+            return B - B
+
+        n = 10
+        self.check(test_impl, (n,n))
+
+    def test_stararg_mix(self):
+        def test_impl(n, r):
+            A = np.random.randn(r, *n)
```
stuartarchibald, line 1032:
```suggestion
            np.random.seed(0)
            A = np.random.randn(r, *n)
```
suggest setting the seed for reproducible testing.

numba/tests/test_parfors.py
```diff
@@ -1018,6 +1018,24 @@ def test_impl(n):
...
+            B = np.sum(A)
+            return B - B
+
+        n = 10
+        self.check(test_impl, (n,n), n)
```
stuartarchibald, line 1037:
```suggestion
        self.check(test_impl, (n, n), n)
```

numba/tests/test_parfors.py
```diff
@@ -1018,6 +1018,24 @@ def test_impl(n):
...
         self.assertEqual(countParfors(test_impl, (types.int64, )), 0)
 
+    def test_stararg(self):
+        def test_impl(n):
+            A = np.random.randn(*n)
+            B = np.sum(A)
+            return B - B
+
+        n = 10
+        self.check(test_impl, (n,n))
```
stuartarchibald, line 1028:
```suggestion
        self.check(test_impl, (n, n))
```

numba/tests/test_parfors.py
```diff
@@ -1018,6 +1018,24 @@ def test_impl(n):
         self.assertEqual(parfor_output, py_output)
         self.assertEqual(countParfors(test_impl, (types.int64, )), 0)
 
+    def test_stararg(self):
```
stuartarchibald, line 1021: RE testing, was there the intention that this should work:
```python
from numba import njit
import numpy as np

@njit(parallel=True)
def foo(*args):
    x = np.random.randn(2, *args)
    return x

print(foo(1, 2))
```
at present this won't lower, it produces:
```
numba/core/base.py", line 616, in get_getattr
    raise NotImplementedError("No definition for lowering %s.%s" % (typ, attr))
NotImplementedError: No definition for lowering UniTuple(int64 x 3).shape
```

numba/parfors/array_analysis.py
```diff
@@ -3148,7 +3185,19 @@ def _make_assert_equiv(self, scope, loc, equiv_set, _args, names=None):
...
+    def _get_typ_part(self, typ, i):
+        if isinstance(typ, types.containers.UniTuple):
+            return typ.dtype
+        elif isinstance(typ, types.ArrayCompatible):
+            return types.intp
+        elif isinstance(typ, types.Number):
+            return typ
+        else:
+            if config.DEBUG_ARRAY_OPT >= 1:
+                print(1, "Unhandled case in _get_typ_part:", typ, i)
```
stuartarchibald, line 3197:
```suggestion
                print("Unhandled case in _get_typ_part:", typ, i)
```
Was the 1 intentional?

numba/parfors/array_analysis.py
stuartarchibald, line 3198: This will end up in this case:
```python
from numba import njit
import numpy as np

@njit(parallel=True)
def foo(*args):
    x = np.random.randn(*args)
    return x

print(foo(np.int32(1), np.int64(2)))
```
it manifests as a lowering error:
```
TypeError: Can only insert i64 at [0] in [2 x i64]: got i32
```

stuartarchibald (commented):
Thanks for the patch @DrTodd13 and thanks for reviewing this @guilhermeleobas. I've noted inline a few additional things to look at, most of which are minor. Thanks again!

DrTodd13:
@stuartarchibald @esc @sklam There's too many issues with this one to get into 0.56.  It isn't critical for us (Intel) and we probably shouldn't half way add support for varargs.  Stuart found some easy counter-examples that you think should work and were close to the original PR and don't.  If the contents of the varags is not just shape but additional arguments then my PR is broken.  For something like np.zeros(*varargs), there's really nothing for array analysis to analyze (maybe extract first element of varargs and that should be shape and analyze that?) and like with Nick's latest PR, we could try a strategy of if we can't analyze then insert a shape call to get it at runtime BUT we only insert shape calls before the statement at the moment and then only on RHS entities.  I think this would require adding shape calls after the np.zeros has finished running.

Issue #7507, opened 2021-10-22
Numba is raising a `NotImplementedError` instead of the `NumbaNotImplementedError` equivalent


guilhermeleobas:
## Reporting a bug

- [x] I have tried using the latest released version of Numba (most recent is
 visible in the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [x] I have included a self contained code sample to reproduce the problem.
  i.e. it's possible to run as 'python bug.py'.

While trying the new error model with [RBC](https://github.com/xnd-project/rbc/), I found a `NotImplementedError` left on `numba/core/base.py` which I suspect it should be a `errors.NumbaNotImplementedError`. This causes the compilation process to crash because the exception is not captured anymore.

PR #7516, opened 2021-10-28
Refresh the registries in get_function early and always.

stuartarchibald:
As title.

### DIFF
--- a/numba/core/base.py
+++ b/numba/core/base.py
@@ -534,13 +534,16 @@ class BaseContext(object):
         lty = self.get_value_type(ty)
         return Constant.null(lty)
 
-    def get_function(self, fn, sig, _firstcall=True):
+    def get_function(self, fn, sig):
         """
         Return the implementation of function *fn* for signature *sig*.
         The return value is a callable with the signature (builder, args).
         """
         assert sig is not None
         sig = sig.as_function()
+        # Reload registries in case something has injected a lowering impl in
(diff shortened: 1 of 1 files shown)


gmarkall:
xref: https://github.com/numba/numba/issues/7507

gmarkall:
Does this fix the issue for the reproducer in #7507?

stuartarchibald:
> Does this fix the issue for the reproducer in #7507?

Yes. But it's costly in performance.

gmarkall:
It does seem to do a lot of redundant work, like loading everything again, when really all we need to do is install new things from the registry added since last time. I wonder if on each refresh we can clear the registry from which typing refreshes, so that next time we refresh, it doesn't do the same things over and over?

stuartarchibald:
> It does seem to do a lot of redundant work, like loading everything again, when really all we need to do is install new things from the registry added since last time. I wonder if on each refresh we can clear the registry from which typing refreshes, so that next time we refresh, it doesn't do the same things over and over?

This lazy approach is carried out to some extent at a relatively low level via: https://github.com/numba/numba/blob/805e24fbd895d90634cca68c13f4c439609e9286/numba/core/utils.py#L466-L482
it might be possible to improve the performance of the logic ahead of this though.

stuartarchibald:
Closing for now, need to reduce the cost of the reload to make this feasible.

PR #7522, opened 2021-10-30
Don't embed current thread count into parallel=True functions so that they can be cacheable.

### LINKED ISSUE #7518: Segfault with caching and parallel=True.
```
import numba
import numpy as np

@numba.njit(cache=True, parallel=True)
def test(a):
    b = a+5
    print(b.shape)
    return b

a = np.ones(1000)
b = test(a)
print(b.shape)
```

Run that with the following:
```
NUMBA_NUM_THREADS=8 python /tmp/caching1.py
NUMBA_NUM_THREADS=16 python /tmp/caching1.py
```

That results in a segfault.

DrTodd13:
Resolves #7518 

Move the printing of the parfor schedule into the C code in debug mode.

Use the dynamic thread count when constructing the schedule so that the parallel=True function can be correctly cacheable.

### DIFF
--- a/numba/np/ufunc/gufunc_scheduler.cpp
+++ b/numba/np/ufunc/gufunc_scheduler.cpp
@@ -326,6 +326,20 @@ std::vector<RangeActual> create_schedule(const RangeActual &full_space, uintp nu
     }
 }
 
+void print_rangeactual_vector(const std::vector<RangeActual> &rav) {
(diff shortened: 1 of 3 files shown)


DrTodd13:
@sklam @stuartarchibald This PR at this point should work for parallel=True functions that don't contain reductions.  There are several possible strategies to resolve this issue for when the functions do contain reductions and I'd like to discuss the possibilities with you before I do any implementation.

Issue #2751, opened 2018-02-14
Support for Numpy.ufunc.reduce


tsoernes:
Would be welcome. For example for methods that are already supported, like `np.bitwise_or.reduce`

PR #7524, opened 2021-11-01
Added overload methods for Numpy's ufunc.reduce

### LINKED ISSUE #6313: [Request] Implement numpy.ufunc.METHOD
## Feature request

There are 5 method for ufuncs that allows in place operations, which could implement in lower level uniformly instead of overwriting one by one. `ufunc.at` and `ufunc.reduce` (https://github.com/numba/numba/issues/2751) are quite welcome.

1. [`ufunc.accumulate`](https://numpy.org/doc/stable/reference/generated/numpy.ufunc.accumulate.html)
2. [`ufunc.at`](https://numpy.org/doc/stable/reference/generated/numpy.ufunc.at.html)
3. [`ufunc.reduce`](https://numpy.org/doc/stable/reference/generated/numpy.ufunc.reduce.html)
4. [`ufunc.reduceat`](https://numpy.org/doc/stable/reference/generated/numpy.ufunc.reduceat.html)
5. [`ufunc.outer`](https://numpy.org/doc/stable/reference/generated/numpy.ufunc.outer.html)

kc611:

Resolves #6313 #2751

This PR aims to implement the `numpy.ufunc.METHOD` as a callable inside `njit`-ed functions so that we can do stuff like: 

```python
import numba as nb
import numpy as np

@nb.njit
def add_twice(x, y):
    return 2*x + 2*y

custom_vectorize = nb.vectorize([], identity=None, target="cpu")
vec_add = custom_vectorize(add_twice)

@nb.njit
def test_fn(x, axis):
    return vec_add.reduce(x, axis) 

print(test_fn(np.ones((10, 20)), 0))
```

### DIFF
--- a/numba/np/ufunc/_internal.c
+++ b/numba/np/ufunc/_internal.c
@@ -344,10 +344,12 @@ dufunc_reduce(PyDUFuncObject * self, PyObject * args, PyObject *kws)
 }
 
 static PyObject *
-dufunc_reduce_direct(PyDUFuncObject * self, PyObject * args, int axis)
-{
-  PyObject *kwargs = Py_BuildValue("{s:L}", "axis", axis);
-  return dufunc_reduce(self, args, kwargs);
+dufunc_reduce_direct(PyDUFuncObject * self, arystruct_t * args, int axis, PyTypeObject *retty,  PyArray_Descr *descr)
+{   
+    PyObject * res = NRT_adapt_ndarray_to_python_acqref(args, retty, 2, 1, descr);
+    PyObject *kwargs = Py_BuildValue("{s:L}", "axis", axis);
+    // The code segfaults on the following call.
+    return dufunc_reduce(self, res, kwargs);
 }
 
 static PyObject *
--- a/numba/np/ufunc/_internal.c
+++ b/numba/np/ufunc/_internal.c
@@ -344,10 +344,9 @@ dufunc_reduce(PyDUFuncObject * self, PyObject * args, PyObject *kws)
 }
 
 static PyObject *
-dufunc_reduce_direct(PyDUFuncObject * self, int axis, PyObject * args)
+dufunc_reduce_direct(PyDUFuncObject * self, PyObject * args, int axis)
 {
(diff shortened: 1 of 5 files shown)


numba/np/ufunc/dufunc.py
```diff
@@ -321,3 +323,33 @@ def _install_cg(self, targetctx=None):
...
+    def codegen(context, builder, signature, args):
+        f_ir, x_ir, axis_ir = args
+
+        fnty = lc.Type.function(
+            x_ir.type,
+            [f_ir.type, x_ir.type, axis_ir.type]
+        )
+
+        fn = cgutils.get_or_insert_function(
+            builder.module, fnty,  "dufunc_reduce_direct"
```
kc611, line 343: The current issue I was facing is that this call cannot detect the implemented `dufunc_reduce_direct` method in C code. Is it a `type` issue ?

numba/np/ufunc/_internal.c
```diff
@@ -344,6 +344,29 @@ init_ufunc_dispatch(void)
     return result;
 }
 
+
+NUMBA_EXPORT_FUNC(PyObject)
```
kc611, line 348: I'm also not too familiar with this C function declarations of Numba so I might be doing something wrong here. 
(Looking at the failing build)

numba/np/ufunc/dufunc.py
gmarkall: You need to add the symbol - try using `llvmlite.binding.add_symbol('dufunc_reduce_direct', <addr of dufunc_reduce_direct>)`

numba/np/ufunc/_internal.c
gmarkall: I don't think you need to wrap the return type in `NUMBA_EXPORT_FUNC` (at least for now) - can you try without it?

gmarkall: Also, I think this should be returning a `PyObject*` since, that's what a `PyCFunctionWithKeywords` returns: https://docs.python.org/3/c-api/structures.html#c.PyCFunctionWithKeywords

PR #7537, opened 2021-11-04
Add chrome tracing

hadia206:
Add support for chrome tracing for run pass events.


hadia206:
@sklam test fails in CI. It works on my machine so I am unable to replicate. Any tips?

stuartarchibald:
@hadia206 Looking at the pattern in the failures, seems like Python 3.7 is the common factor? Which python version are you using?

hadia206:
> @hadia206 Looking at the pattern in the failures, seems like Python 3.7 is the common factor? Which python version are you using?

@stuartarchibald I am using `Python 3.9.5`

numba/core/event.py
```diff
@@ -423,3 +426,44 @@ def on_exit(*exc_details):
...
+def _write_chrome_trace(rec):
+    """ If chrome tracing is enabled,
+        write event information (process/thread id, event name,
+        and duration time to json file.
+    """
+    import os
+    import threading
+
+    pid = os.getpid()
+    tid = threading.get_native_id()
```
stuartarchibald, line 440: This is Python 3.8+ https://docs.python.org/3/library/threading.html#threading.get_native_id hence the code fails on Python 3.7. If you take a look at the output of `run_cmd` in https://github.com/numba/numba/pull/7537/files#diff-45abfee05930c5c2e330559e75b0b5ed4a7e1b56b1223f99674ad7a6f1b9a64fR20 you'll see:
```
 'line 440, in _write_chrome_trace\n'
 '    tid = threading.get_native_id()\n'
 "AttributeError: module 'threading' has no attribute 'get_native_id'\n"
```

hadia206: Thanks a lot @stuartarchibald. I fixed that.

numba/core/event.py
```diff
@@ -423,3 +426,42 @@ def on_exit(*exc_details):
...
+        name = data['name']
+        args = data
+        ev = dict(
+            cat=cat, pid=pid, tid=tid, ts=ts, ph=ph, name=name, args=args,
+        )
+        evs.append(ev)
+    return evs
+
+
+if config.CHROME_TRACE:
```
stuartarchibald, line 458: Should this be at module level? I wonder if Numba needs a process level hook mechanism.

numba/core/event.py
```diff
@@ -423,3 +426,42 @@ def on_exit(*exc_details):
 
         start_event(kind, data=data)
         yield
+
+
+def _write_chrome_trace(rec):
```
stuartarchibald, line 431:
```suggestion
def _chrome_trace_parse_records(rec):
```
suggest changing the name to something that doesn't involve writing, as it doesn't write?

numba/core/event.py
```diff
@@ -423,3 +426,42 @@ def on_exit(*exc_details):
...
+    """ If chrome tracing is enabled,
+        write event information (process/thread id, event name,
+        and duration time to json file.
```
stuartarchibald, line 434: This doc string is outdated for the contents of this function? There's no write or json involved, looks like it's just parsing event records into a list of dicts for consumption by the json dumping function later?

numba/tests/test_chrome_trace.py
```diff
@@ -0,0 +1,23 @@
...
+    Test chrome tracing generated file(s).
+    """
+
+    def test_trace_output(self):
+        with TemporaryDirectory() as tmpdir:
+            path = os.path.join(tmpdir, "test_trace.json")
+            with override_env_config('NUMBA_CHROME_TRACE', path):
+                from .chrome_trace_usecase import __file__
+                cmdline = [sys.executable, __file__]
+                run_cmd(cmdline)
```
stuartarchibald, line 20: Can probably make use of `numba.tests.support::run_in_subprocess` and define the contents of `chrome_trace_usecase.py` as a string. Perhaps something like this produces the equivalent:
```python
    def test_trace_output(self):
        code = """
                from numba import njit
                import numpy as np

                x = np.arange(100).reshape(10, 10)

                @njit
                def go_fast(a):
                    trace = 0.0
                    for i in range(a.shape[0]):
                        trace += np.tanh(a[i, i])
                    return a + trace

                go_fast(x)
                """

        src = dedent(code)
        with TemporaryDirectory() as tmpdir:
            path = os.path.join(tmpdir, "test_trace.json")
            env = os.environ.copy()
            env['NUMBA_CHROME_TRACE'] = path
            run_in_subprocess(src, env=env)
            with open(path) as file:
                jfile = json.load(file)
                self.assertTrue(jfile)
```

numba/tests/test_chrome_trace.py
```diff
@@ -0,0 +1,23 @@
...
+                with open(path) as file:
+                    jfile = json.load(file)
+                    self.assertTrue(jfile)
```
stuartarchibald, line 23: This test module needs:
```python
if __name__ == "__main__":
    unittest.main()

```
at the end.

docs/source/reference/envvars.rst
```diff
@@ -176,6 +176,12 @@ These variables influence what is printed out during compilation of
    If set to non-zero, trace certain function calls (function entry and exit
    events, including arguments and return values).
 
+.. envvar:: NUMBA_CHROME_TRACE
+
+   Specify the location of the chrome tracing json file output. 
+
+   If not defined, chrome tracing is disabled.
```
stuartarchibald, line 183:
```suggestion
   Specify the location of the chrome tracing json output file. If defined
   chrome tracing is enabled.
```

numba/core/event.py
```diff
@@ -423,3 +426,42 @@ def on_exit(*exc_details):
...
+            cat=cat, pid=pid, tid=tid, ts=ts, ph=ph, name=name, args=args,
+        )
+        evs.append(ev)
+    return evs
+
+
+if config.CHROME_TRACE:
+    listener = RecordingListener()
+    register("nb:run_pass", listener)
+    filename = config.CHROME_TRACE
```
stuartarchibald, line 461: I don't think this is going to be multiprocessing safe? Should the filename be combined with the `pid` of the parent process.

numba/core/event.py
stuartarchibald: I don't think this will work with multiprocessing either, have been testing with:
```python
import os
import multiprocessing as mp
from numba import njit
import numpy as np

@njit
def go_fast(a):
    trace = 0.0
    for i in range(a.shape[0]):
        trace += np.tanh(a[i, i])
    return a + trace

def foo(q, arr):
    q.put(go_fast(arr))

if __name__ == '__main__':
    arr = np.arange(100).reshape(10, 10)
    mp.set_start_method('fork')
    q = mp.Queue()
    procs = []
    n = os.environ.get('NPROCS')
    assert n
    n = int(n)
    for i in range(n):
        procs.append(mp.Process(target=foo, args=(q, arr)))

    for proc in procs:
        proc.start()

    for proc in procs:
        proc.join()

    acc = 0
    for i in range(n):
        acc += q.get()

    assert np.allclose(acc[0, 0], 9 * n)
```
then:
```
NPROCS=4 NUMBA_CHROME_TRACE='/tmp/foo.json' python test_7537.py 
```
This may not matter/just be a limitation of this feature, in which case it should be added to the docs as such?

stuartarchibald (commented):
@hadia206 Many thanks for the patch. I've given this an initial review to provide some feedback and can confirm that this does produce a json repr of the event log in a file. I'd really like to see some docs adding if possible please, mainly to provide guidance on how to make good use of this feature. Thanks again!

numba/core/event.py
sklam: Added warning in docs about the limitation for multiprocessing

sklam:
@stuartarchibald, I've addressed your review. 

There is a bug that I am working on and this PR provides a useful view to the compiler that helps me understand the problem.

stuartarchibald:
@sklam thanks for patching this. Looks like CI is failing on Python 3.7 due to the json output file not being found.

sklam:
oh no, that's not it. why isn't the file there?!

sklam:
It's working now

docs/source/reference/envvars.rst
```diff
@@ -176,6 +176,14 @@ These variables influence what is printed out during compilation of
    If set to non-zero, trace certain function calls (function entry and exit
    events, including arguments and return values).
 
+.. envvar:: NUMBA_CHROME_TRACE
+
+   If defined, chrome tracing is enabled and this variable specify the filepath
```
stuartarchibald, line 181:
```suggestion
   If defined, chrome tracing is enabled and this variable specifies the filepath
```

docs/source/reference/envvars.rst
```diff
@@ -176,6 +176,14 @@ These variables influence what is printed out during compilation of
...
+   of the chrome tracing json file output. The emitted file can be opened by
+   Chromium-based browser using the profile viewer at `chrome://tracing/`.
```
stuartarchibald, line 183:
```suggestion
   a Chromium-based browser using the profile viewer at `chrome://tracing/`.
```

docs/source/reference/envvars.rst
```diff
@@ -176,6 +176,14 @@ These variables influence what is printed out during compilation of
...
+
+   .. warning:: This feature is not supported on multi-process application.
```
stuartarchibald, line 185:
```suggestion
   .. warning:: This feature is not supported in multi-process applications. 
```

numba/core/event.py
```diff
@@ -423,3 +429,53 @@ def on_exit(*exc_details):
...
+    except AttributeError:
+        # Fallback for python <3.8
+        return threading.get_ident()
+
+
+def _prepare_chrome_trace_data(listener: RecordingListener):
+    """Prepare events in `listener` for serializing as chrome trace data.
+    """
+    # The spec for the trace event format can be found at:
+    # https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/edit   # noqa
```
stuartarchibald, line 446: Is there a non-google doc version of the spec?

numba/core/event.py
```diff
@@ -455,13 +461,21 @@ def _write_chrome_trace(rec):
     return evs
 
 
-if config.CHROME_TRACE:
+def _setup_chrome_trace_exit_handler():
+    """Setup a RecordingListener and a exit handler to write the captured events
```
stuartarchibald, line 465:
```suggestion
    """Setup a RecordingListener and an exit handler to write the captured events
```

numba/tests/test_chrome_trace.py
```diff
@@ -0,0 +1,51 @@
...
+                for i in range(a.shape[0]):
+                    trace += np.tanh(a[i, i])
+                return a + trace
+
+            go_fast(x)
+        """
+
+        src = dedent(code)
+        with TemporaryDirectory() as tmpdir:
+            path = os.path.join(tmpdir, "test_trace.json")
```
stuartarchibald, line 34: Could this just be `tempfile.NamedTemporaryFile`?

numba/core/event.py
```diff
@@ -423,3 +429,53 @@ def on_exit(*exc_details):
...
+
+    @atexit.register
+    def _write_chrome_trace():
+        # The following output file is not multi-process safe.
+        evs = _prepare_chrome_trace_data(listener)
+        with open(filename, "w") as out:
+            json.dump(evs, out)
+
+
+if config.CHROME_TRACE:
```
stuartarchibald, line 480: Should this be at module level?

stuartarchibald (commented):
Thanks for the update, there's a few minor things to look at (mostly typos), once resolved should be good to merge. Thanks again!

numba/core/event.py
sklam: I wanted the env-var to enable the atexit hook and one can call `_setup_chrome_trace_exit_handler()` programmatically if the env-var is not set.

numba/tests/test_chrome_trace.py
sklam: No, `NamedTemporaryFile` will create and open the file. We want the chrometrace code to do that. The tempdir guarantee a safe empty directory.

numba/core/event.py
sklam: I tried to find it but no. All other tools I found are referring to it via the google doc link. Google tends to use google doc for spec.

numba/tests/test_chrome_trace.py
stuartarchibald: Ah yes, good point.

numba/core/event.py
stuartarchibald: ok, thanks for checking.

stuartarchibald (approved):
Thanks for the patch and fixes @hadia206 and @sklam.

Issue #7533, opened 2021-11-04
Missed optimization on typed lists


zephyr111:
Numba generate JIT calls for lists that are not optimized out by LLVM-Lite in practice. This include reference counting calls and especially function calls to check the size of the list in the following example. The fact that the functions related to the list are not inlined is a missing optimization (especially because of the `alwaysinline` and `readonly` JIT tags). This can have a relatively huge impact on the performance of codes using lists intensively. A detailled analysis (including the generated assembly code) can be found on: https://stackoverflow.com/questions/69738653/why-is-element-access-for-typed-lists-so-much-slower-than-for-arrays-with-numba.

```python
import numpy as np
import numba as nb

@nb.njit
def only_rand(N):
    s = 0
    for _ in range(10000):
        i = np.random.randint(N)
        j = np.random.randint(N)
        s += i + j
    return s

@nb.njit
def foo(pos, N):
    s = 0
    for _ in range(10000):
        i = np.random.randint(N)
        j = np.random.randint(N)
        
        dx = pos[i][0] - pos[j][0]
        dy = pos[i][1] - pos[j][1]
        dz = pos[i][2] - pos[j][2]
        s += dx + dy + dz
    return s

N = 100

Array = np.random.rand(N,3)
List = nb.typed.List(Array)

print('Random number generation:')
%timeit only_rand(N)
print('Numpy Array:')
%timeit foo(Array, N)
print('Typed List:')
%timeit foo(List, N)
```

Here is the result on my machine (i5-9600KF):

```
Random number generation:
109 µs ± 2.55 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
Numpy Array:
119 µs ± 566 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
Typed List:
739 µs ± 4.54 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

The version of Numba tested is 0.53.1 (on Windows with CPython 3.8.1).

sklam:
Thank you for reporting the performance issue. I can confirm that this continues to be a problem on 0.54.x and current [mainline](https://github.com/numba/numba/commit/6c95d38d945bb5ee77e52662bed416e24a2770ef). 

Unfortunately, the non-typedlist version has also gotten slower in new version:

post 0.53
```python
Random number generation:
217 µs ± 2.87 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
Numpy Array:
[...]

PR #7538, opened 2021-11-05
Fix `@overload_glue` performance regression.

stuartarchibald:
The code generated by `@overload_glue` is too involved for LLVM
to inline by default, this patch forces inlining to end up with
generated code similar to that prior to `overload_glue` where IR
was directly injected at the call site.

### DIFF
--- a/numba/core/options.py
+++ b/numba/core/options.py
@@ -88,6 +88,7 @@ class DefaultOptions:
     fastmath = _mapping("fastmath")
     error_model = _mapping("error_model")
     inline = _mapping("inline")
+    forceinline = _mapping("forceinline")
 
     target_backend = _mapping("target_backend")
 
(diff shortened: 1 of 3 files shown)


sklam (approved):
I tested it with `only_rand()` from https://github.com/numba/numba/issues/7533#issue-1044232613.  This patch provides a ~3x speedup.

The before and after CFG is shown below:

Before:
before

After:
after

This patch opens up more optimization opportunity for LLVM.

numba/core/options.py
```diff
@@ -88,6 +88,7 @@ class DefaultOptions:
     fastmath = _mapping("fastmath")
     error_model = _mapping("error_model")
     inline = _mapping("inline")
+    forceinline = _mapping("forceinline")
```
sklam, line 91: NOTE: the `forceinline` flag pre-existed this PR. It was previously internal only and it is used by the `FunctionDescriptor.inline` at https://github.com/numba/numba/blob/58426cf7658177b792b820ff66c0945985b72da0/numba/core/typed_passes.py#L390

PR #7543, opened 2021-11-06
Support for all modes in np.correlate and np.convolve

### LINKED ISSUE #7406: Support for Numpy convolve mode 'valid'
## Feature request

I've been working with model simulations and there's the need for `np.convolve(a, k, mode='valid')`.

I wish to have support for this for use with `@jit`/`@njit` decorators.

I have this example implementation from one of the discussions about using Numba for such situation, but I'd rather not have to resort to it in projects which require valid mode.
```python
def numba_convolve_mode_valid_as_loop(arr, kernel):
        m = arr.size
        n = kernel.size
        out_size = m - n + 1
        out = np.empty(out_size, dtype=np.complex128)
        kernel = np.flip(kernel)  # wrong results otherwise
        for i in range(out_size):
            out[i] = np.dot(arr[i:i + n], kernel)
        return out
```

jeertmans:
Hello everyone,

With this PR, I hope to bring the missing features from `np.convolve` and `np.correlate` into Numba. As you can see in the edited files, the initial solution was not correct, and I had to modify a bit the `_np_correlate_core_impl` to make it work.
I also added some tests to verify that it works and it should close #7406.

Last but not least, I think that `_np_correlate_core_impl` should allow for arguments where `len(ap1) < len(ap2)`. Here, it is assumed that the arrays are ordered properly, but in my opinion that should be the job of `_np_correlate_core_impl`. Same could apply when checking for the `mode` value. Don't hesitate to give me your opinion on that?

Thanks in advance!


stuartarchibald:
@jeertmans Many thanks for the PR, great to see this implemented, IIRC there's been quite a few requests for it!

Some initial feedback...

> As you can see in the edited files, the initial solution was not correct

Any chance you could please demonstrate how the original code was incorrect, was there some input that produced an invalid result or have I perhaps misunderstood? For reference, the implementation in Numba is based on that in NumPy referenced here: https://github.com/numba/numba/blob/master/numba/np/arraymath.py#L4137.

> RE: Last but not least, I think that _np_correlate_core_impl should allow for arguments where len(ap1) < len(ap2).

Numba tries where possible to translate the implementation in NumPy, this is so it is easier to track changes in NumPy's behaviour, Numba tries to replicate NumPy behaviour right down to implementation bugs! It looks like the `_pyarray_correlate` function does a variable swap in the function itself if an inversion is needed, and as a side effect updates the `inverted` variable:
https://github.com/numpy/numpy/blob/3bce2be74f228684ca2895ad02b63953f37e2a9d/numpy/core/src/multiarray/multiarraymodule.c#L1205-L1216
later this variable is read and if the input was decided to be inverted then the output is reversed:
https://github.com/numpy/numpy/blob/3bce2be74f228684ca2895ad02b63953f37e2a9d/numpy/core/src/multiarray/multiarraymodule.c#L1389-L1403

Given passing a scalar by reference is not something to consider from Python, it's probably ok to have this deviation from the original algorithms. Essentially, Numba is lifting the inversion logic out of the callee.

jeertmans:
Thanks for your review @stuartarchibald ! I will answer to it in two parts.

**First:**
> Any chance you could please demonstrate how the original code was incorrect, was there some input that produced an invalid result or have I perhaps misunderstood? For reference, the implementation in Numba is based on that in NumPy referenced here: https://github.com/numba/numba/blob/master/numba/np/arraymath.py#L4137.

Well, in my opinion, it was incorrect for two reasons:

1. It did not implement the part for `mode="same"`, which the NumPy reference does;
2. and simply adding the possibility to change the `mode` was not sufficient to make it work.

Actually, `np.convolve` and `np.correlate` implementation only worked for their default value of `mode` (which are different).

The "incorrectness" of the implementation comes from how the side effects are handled, i.e., the left and right sides. See example below, tests are run on [this commit version](https://github.com/jeertmans/numba/commit/466c49ebe94327edf736bd26acf4a61b2686c1e6). Here, the convolution of `[0 1]` with `[0 1 2]` failed for `mode="full"`.
```python
FF
FAIL: test_correlate (numba.tests.test_np_functions.TestNPFunctions)
  File "/home/jeertmans/repositories/numba/numba/tests/test_np_functions.py", line 1208, in test_correlate
    self._test_correlate_convolve(correlate)
  File "/home/jeertmans/repositories/numba/numba/tests/test_np_functions.py", line 1197, in _test_correlate_convolve
  File "/home/jeertmans/repositories/numba/numba/tests/support.py", line 361, in assertPreciseEqual
    self.fail("when comparing %s and %s: %s" % (first, second, failure_msg))
AssertionError: when comparing [0 2 1 0] and [  0   2   1 -67]: 0 != -67
...
```
The error come from a dot product of arrays with weird lengths (either zeros or different vector lengths).

Generally speaking, the convolution method can be split into two parts:

1. The main (central) part where the smaller (inner) vector can totally fit into the other bigger (outer) vector
2. Then two subparts (left and right) where the inner vector is has to be truncated. What I mean by `truncation` is the number of elements that have to be discarded during the dot product. From left to right, the truncation decreases (left part), then stays constant (central), and finally increases (right part).
The reason why it fails it that the truncation sizes are not correct in all possible cases. Maybe it is easier to do a small drawing about this... :'-) Tell me if you would like more detailed information.

**Second:**
> > RE: Last but not least, I think that _np_correlate_core_impl should allow for arguments where len(ap1) < len(ap2).
> 
> Numba tries where possible to translate the implementation in NumPy, this is so it is easier to track changes in NumPy's behaviour, Numba tries to replicate NumPy behaviour right down to implementation bugs! It looks like the `_pyarray_correlate` function does a variable swap in the function itself if an inversion is needed, and as a side effect updates the `inverted` variable: https://github.com/numpy/numpy/blob/3bce2be74f228684ca2895ad02b63953f37e2a9d/numpy/core/src/multiarray/multiarraymodule.c#L1205-L1216 later this variable is read and if the input was decided to be inverted then the output is reversed: https://github.com/numpy/numpy/blob/3bce2be74f228684ca2895ad02b63953f37e2a9d/numpy/core/src/multiarray/multiarraymodule.c#L1389-L1403
> 
> Given passing a scalar by reference is not something to consider from Python, it's probably ok to have this deviation from the original algorithms. Essentially, Numba is lifting the inversion logic out of the callee.

Ok for me. I understand your point, and I think this deviation is not a major problem if we keep raising an error in case the parameters lengths do not match the predicate `len(ap1) >= len(ap2)`.

**Finally:**

Tell me if there is anything left to do, I would be glad to help!

stuartarchibald:
> Thanks for your review @stuartarchibald ! I will answer to it in two parts.

No problem.

<snip>

> Well, in my opinion, it was incorrect for two reasons:
> 
>     1. It did not implement the part for `mode="same"`, which the NumPy reference does;
> 
>     2. and simply adding the possibility to change the `mode` was not sufficient to make it work.
> 
> 
> Actually, `np.convolve` and `np.correlate` implementation only worked for their default value of `mode` (which are different).
> 
> The "incorrectness" of the implementation comes from how the side effects are handled, i.e., the left and right sides. See example below, tests are run on [this commit version](https://github.com/jeertmans/numba/commit/466c49ebe94327edf736bd26acf4a61b2686c1e6). Here, the convolution of `[0 1]` with `[0 1 2]` failed for `mode="full"`.
<snip> 

For reference, Numba doesn't always fully implement everything, this doesn't make it incorrect, just limited. The aim is to try and implement as much as possible and document the limitations. Sometimes resource constraints limit how much gets implemented, and this also influences the implementation, i.e. a corner that cannot be reached due to a limitation might be cut (which looks the the case in the issue described). So long as the problem cannot manifest in production through user input it's effectively working as intended.

> **Second:**
<snip>

> Ok for me. I understand your point, and I think this deviation is not a major problem if we keep raising an error in case the parameters lengths do not match the predicate `len(ap1) >= len(ap2)`.

Thanks for your understanding. I'm not sure that the branch here can be reached https://github.com/numba/numba/blob/a24e70aef600c72d74369844f0309dea057b0613/numba/np/arraymath.py#L4147-L4148 because the the guard code here would prevent it https://github.com/numba/numba/blob/a24e70aef600c72d74369844f0309dea057b0613/numba/np/arraymath.py#L4247-L4250 ?

> 
> **Finally:**
> 
> Tell me if there is anything left to do, I would be glad to help!

Having just had a quick look at the code, I'd be interested in your thoughts on whether the loop ranges and indexing in the proposed patch here:
https://github.com/numba/numba/blob/a24e70aef600c72d74369844f0309dea057b0613/numba/np/arraymath.py#L4177-L4187
could be made to closer match the expression in the NumPy impl: https://github.com/numba/numba/blob/5fc9d3c56da4e4c6aef7189e588ce9c44263d4a6/numba/np/arraymath.py#L4174-L4184
the original Numba code manages to do this to some degree:
https://github.com/numba/numba/blob/5fc9d3c56da4e4c6aef7189e588ce9c44263d4a6/numba/np/arraymath.py#L4174-L4184
it's this sort of thing which helps make it easier to track changes upstream.

Once this is resolved I'll take a closer look at the patch itself.

Many thanks.

jeertmans:
> For reference, Numba doesn't always fully implement everything, this doesn't make it incorrect, just limited. The aim is to try and implement as much as possible and document the limitations. Sometimes resource constraints limit how much gets implemented, and this also influences the implementation, i.e. a corner that cannot be reached due to a limitation might be cut (which looks the the case in the issue described). So long as the problem cannot manifest in production through user input it's effectively working as intended.

Yes I agree, but then it was (in my opinion) clearly misleading up to what point the reference NumPy function had been reimplemented. For example, the function would not work for correlate with full mode, but well for convole. Even if this function is not directly exposed to users, I think that the API should be clear to programmers, in case of future improvements (as it was the case here).

> Thanks for your understanding. I'm not sure that the branch here can be reached
> 
> https://github.com/numba/numba/blob/a24e70aef600c72d74369844f0309dea057b0613/numba/np/arraymath.py#L4147-L4148
> 
> because the the guard code here would prevent it
> https://github.com/numba/numba/blob/a24e70aef600c72d74369844f0309dea057b0613/numba/np/arraymath.py#L4247-L4250
> 
> ?

This directly follows what I mentioned above. Here the conditional statement is useless, but I added it for future developers. This was a way, and there are many, to document the fact that this function derives a bit from reference function and only accepts some lengths.

I thought it would be useful as I had a lot of hard time trying to understand why some would not work, and such a documentation would have saved me a lot of time :)

> Having just had a quick look at the code, I'd be interested in your thoughts on whether the loop ranges and indexing in the proposed patch here:
> 
> https://github.com/numba/numba/blob/a24e70aef600c72d74369844f0309dea057b0613/numba/np/arraymath.py#L4177-L4187
> 
> 
> could be made to closer match the expression in the NumPy impl:
> https://github.com/numba/numba/blob/5fc9d3c56da4e4c6aef7189e588ce9c44263d4a6/numba/np/arraymath.py#L4174-L4184
> 
> 
> the original Numba code manages to do this to some degree:
> https://github.com/numba/numba/blob/5fc9d3c56da4e4c6aef7189e588ce9c44263d4a6/numba/np/arraymath.py#L4174-L4184
> 
> 
> it's this sort of thing which helps make it easier to track changes upstream.
> Once this is resolved I'll take a closer look at the patch itself.
> 
> Many thanks.

For the `idx = idx + inc`, sure it can do this.

But for the rest of the implementation, I am not sure how. Part of the original implementation did not work for all modes and had to be changed, and part of the implementation was also changed to increase readability. Of course, we can debate on this and I am up to discuss this.
My philosophy was to respect NumPy’s implementation in terms of how it computes everything, but also to make the program as readable as possible. 
First, it did find useful to keep the increment into the array indexation, as it makes you do additional computation while you could avoid it. Second, as the algorithm is (almost) symmetric, I made my implementation also symmetric. Third, directly changing the ranges was, for me, the good compromise between changes in the code and readability.
The way I implemented this seemed very natural to me when I started to draw on a paper the different cases I had to handle.

Please let me know what you think about that.

jeertmans:
Hello @stuartarchibald, just coming for potential update :-)
Is there something that I should do, or is my PR still to be reviewed ?
Thanks!

stuartarchibald:
> Hello @stuartarchibald, just coming for potential update :-) Is there something that I should do, or is my PR still to be reviewed ? Thanks!

Hi @jeertmans, apologies for the delay, the Numba engineers are trying to get a release candidate for Numba 0.55 prepared and so there is a general PR freeze for anything not in that milestone (https://github.com/numba/numba/milestone/56). We will begin reviewing PRs again once the RC is out and we thank you in advance for your patience and understanding!

jeertmans:
Hello @stuartarchibald , I am sorry to "up" this PR, but I was wondering you would now have time for reviewing my PR? Thanks in advance ;-)

jeertmans:
Hi @stuartarchibald, any update on this review? Sorry to "UP" again

falseywinchnet:
I would also like to see this implemented. I use multiple convolve modes and I have to looplift to use convolve with my numba code. It would allow me to make my entire program jit'd, at the moment, the only reason it is not is because of this.

jeertmans:
Hello @stuartarchibald, do you think you would have some time for reviewing this?

Thanks :-)

Makhaos:
This implementation would be great, I would like to be able to use all modes from Numpy's correlate. Hopefully the reviewers will have some time for it soon

jeertmans:
Hello @stuartarchibald! As v0.57 was recently released, I was wondering if you could review this?

Do not hesitate to tell me if I can help make this process easier :-)

seibert:
@kc611: Can you pick up the review on this PR?

numba/np/arraymath.py
```diff
@@ -4116,11 +4116,12 @@ def impl(ap1, ap2, mode, direction):
...
         # -1 to write them out N->0.
 
-        if not (mode == Mode.VALID or mode == Mode.FULL):
-            raise ValueError("Invalid mode")
-
         n1 = len(ap1)
         n2 = len(ap2)
+
+        if n1 < n2:
+            raise ValueError("'len(ap1)' must greate than 'len(ap2)'")
```
kc611, line 4123:
```suggestion
            raise ValueError("'len(ap1)' must greater than 'len(ap2)'")
```

Also a good idea to add a test for this error, so that we have coverage.

numba/np/arraymath.py
```diff
@@ -4192,41 +4196,69 @@ def op_nop(x):
             a_op = op_conj
             b_op = op_nop
 
-    def impl(a, v):
+    _NP_PRED = numpy_version > (1, 17)
```
kc611, line 4199: The minimum supported NumPy is `1.20` as of now, so this can simply be removed.

numba/np/arraymath.py
```diff
@@ -4192,41 +4196,69 @@ def op_nop(x):
...
+    def impl(a, v, mode="valid"):
         la = len(a)
         lv = len(v)
-        if la == 0:
-            raise ValueError("'a' cannot be empty")
-        if lv == 0:
-            raise ValueError("'v' cannot be empty")
+
+        if mode == "full":
+            corr_mode = Mode.FULL
```
kc611, line 4206: Since we're now introducing the `mode` as an argument I think the `Mode` Enum framework can be removed entirely.

We can simply pass the `mode` in the `_np_correlate_core` and compare it over there as a string?

numba/np/arraymath.py
kc611, line 4123:
```suggestion
            raise ValueError("'len(ap1)' must greater than 'len(ap2)'")
```

Also a good idea is to add a test for this error, so that we have coverage.

numba/np/arraymath.py
kc611, line 4199: The minimum supported NumPy is `1.20` as of now, so this can simply be removed.

numba/np/arraymath.py
kc611, line 4206: Since we're now introducing the `mode` as an argument in both the functions where `_np_correlate_core` is being used. I think the `Mode` Enum class can be removed entirely.

We can simply pass the `mode` in the `_np_correlate_core` and compare it over there as required? (Same with the `ValueError` being thrown over here)

numba/np/arraymath.py
```diff
@@ -4146,24 +4149,25 @@ def impl(ap1, ap2, mode, direction):
         else:
             raise ValueError("Invalid direction")
 
-        for i in range(n_left):
-            ret[idx] = innerprod(ap1[:idx + 1], ap2[-(idx + 1):])
+        for i in range(n - n_left, n):
+            ret[idx] = innerprod(ap1[:i], ap2[-i:])
```
kc611, line 4153: I would recommend keeping this particular part of implementation closer to NumPy if possible. It serves as a reference point for most of the users, even if it is a bit less readable/natural. 

As for the optimization of the increment being made over here, the constant increment within the loop (`+1` part) as a general rule of thumb is supposed to be optimized out by LLVM/compiler passes. (Although I'll need to check for this before making this a claim.)

kc611 (changes requested):

kc611:
Hi @jeertmans, I gave the PR another round of review. It largely looks good to me, except for a few minor nitpicks that I've highlighted above.

numba/np/arraymath.py
jeertmans: As now documented by the comments, I don't see how to cover this in tests, because this should never fail when called with `np.correlate` or `np.convolve`. This check is only here to avoid future errors if one change the upstream code of `_np_correlate.imp` or else.

numba/np/arraymath.py
jeertmans: This is the best I can do, without making it a total mess (in my opinion, making it closer to Numpy's implementation would dramatically decrease readability, and maybe performances):

https://github.com/numba/numba/blob/0c5e47ee43053d52c2e526ecd7cfe702e9c9be79/numba/np/arraymath.py#L4146-L4158

numba/np/arraymath.py
jeertmans: See, e.g.:

https://github.com/numba/numba/blob/0c5e47ee43053d52c2e526ecd7cfe702e9c9be79/numba/np/arraymath.py#L4202-L4205

jeertmans:
Hi @kc611, thanks for your review! I have addressed all your comments in (mostly) separate commits. Please feel free to give your opinion on those :-)

numba/np/arraymath.py
kc611: Got it, thanks for the explanation.

numba/np/arraymath.py
```diff
@@ -4192,29 +4190,31 @@ def op_nop(x):
...
     return impl
 
 
 @overload(np.convolve)
-def np_convolve(a, v):
+def np_convolve(a, v, mode="full"):
     _assert_1d(a, 'np.convolve')
     _assert_1d(v, 'np.convolve')
 
     Mode = _corr_conv_Mode
```
kc611, line 4215:
```suggestion
```

Seems like accidentally leftover code.

kc611 (changes requested):

numba/np/arraymath.py
jeertmans: Good catch! Thanks

jeertmans:
Thanks for your review @kc611! I have made some fixes to make the tests pass. It should now be all good :-)

kc611 (approved):
Yep, LGTM! Thank you @jeertmans.

esc (approved):
as per @kc611

esc:
thank you for the patch!

danilo-bc:
Thank you everyone for this, specially @jeertmans for taking the initiative!

jeertmans:
Thank you for the review! Happy to contribute to this great work!

stuartarchibald:
@kc611 thanks for picking up and finishing the review on this. Thank you @jeertmans for all your effort and patience, it's great to see this finished. This will be in Numba 0.58.0.

PR #7546, opened 2021-11-09
Fix handling of missing const key in LiteralStrKeyDict

### LINKED ISSUE #7545: LiteralStrKeyDict incorrectly handles missing const-key at compile time.
## Reporting a bug

This:
```python
from numba import njit
@njit
def foo():
    a = {'not_a': 2j, 'c': 'd', 'e': np.zeros(4)}
    return a['a']

foo()
```
does this (or wraps it as an `InternalError` in the `old_style` error capturing):
```
numba/core/typing/builtins.py", line 648, in generic
    lookup = tup.fields.index(idx)
ValueError: tuple.index(x): x not in tuple
```

stuartarchibald:
As title.

Fixes #7545

### DIFF
--- a/numba/core/typing/builtins.py
+++ b/numba/core/typing/builtins.py
@@ -645,7 +645,10 @@ class StaticGetItemLiteralStrKeyDict(AbstractTemplate):
         if not isinstance(tup, types.LiteralStrKeyDict):
             return
         if isinstance(idx, str):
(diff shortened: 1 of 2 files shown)


sklam (dismissed):
Thanks for the patch!

sklam:
Attempted to resolve merge conflict

sklam (approved):

PR #7548, opened 2021-11-09
Correct evaluation order in assert statement

gmarkall:
This fixes the condition in the `assert` such that the `|` is evaluated before the `==`.


stuartarchibald (approved):
Thanks for fixing this, looks good.

Issue #3426, opened 2018-10-21
np.random.multinomial throws ValueError: binomial(): p outside of [0, 1]


al5250:
## Reporting a bug

- [ ] I am using the latest released version of Numba (most recent is visible in
 the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [ ] I have included below a minimal working reproducer (if you are unsure how
 to write one see http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports).

Suppose I have a vector of probabilities that add up to 1.  Sometimes, I have found that using `np.random.multinomial` throws an error.  For example,
```
A = np.array([0.     , 0.91954, 0.01867, 0.     , 0.     , 0.01745, 0.04434,
              0.     , 0.     , 0.     , 0.     , 0.     ])
@jit(nopython=True)
def samp_multi(x):
    return np.random.multinomial(1, x)

for _ in range(100):
    samp_multi(A)
```
leads to the following error:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-794-f6a667d67d4f> in <module>()
      1 for _ in range(100):
----> 2     samp_multi(A)

//anaconda/envs/py35/lib/python3.5/site-packages/numba/targets/randomimpl.py in binomial_impl()
    812             raise ValueError("binomial(): n <= 0")
    813         if not (0.0 <= p <= 1.0):
--> 814             raise ValueError("binomial(): p outside of [0, 1]")
    815         if p == 0.0:
    816             return 0

ValueError: binomial(): p outside of [0, 1]
```

rjenc29:
Just a thought: the fraction **p_j / p_sum** may be affected by float precision behaviour (numba/targets/randomimpl.py:1444):

```python
n_j = fl[i + j] = np.random.binomial(n_experiments, p_j / p_sum)
```
If so, perhaps capping the ratio at 1 would make sense / be justified:

```python
n_j = fl[i + j] = np.random.binomial(n_experiments, min(p_j / p_sum, 1))
```

stuartarchibald:
Thanks for the report. I think Numba needs to follow NumPy better here and compute the more stable Kahan sum when assessing this condition (and also check out the numerical stability of the call to `np.random.binomial` as @rjenc29 suggests).

https://github.com/numpy/numpy/blob/db5750f6cdc2715f1c65be31f985e2cd2699d2e0/numpy/random/mtrand/mtrand.pyx#L4627-L4628

[...]

PR #7551, opened 2021-11-11
Fixes to random.multinomial

juliusbierk:
First attempt at fixing https://github.com/numba/numba/issues/3426

### DIFF
--- a/numba/cpython/randomimpl.py
+++ b/numba/cpython/randomimpl.py
@@ -1453,7 +1453,7 @@ def multinomial(n, pvals, size=None):
         
         psum = 0.0
         for i in range(plen):
-            psum += pval[i]
+            psum += pvals[i]
         if psum > 1 + 1e-10:
             raise AssertionError("np.random.multinomial(): Sum of probabilities "
                                  "cannot exceed 1.")
--- a/numba/cpython/randomimpl.py
+++ b/numba/cpython/randomimpl.py
@@ -1450,7 +1450,11 @@ def multinomial(n, pvals, size=None):
         fl = out.flat
         sz = out.size
         plen = len(pvals)
-        if np.sum(pvals) > 1 + 1e-10:
+        
+        psum = 0.0
+        for i in range(plen):
+            psum += pval[i]
+        if psum > 1 + 1e-10:
             raise AssertionError("np.random.multinomial(): Sum of probabilities "
                                  "cannot exceed 1.")
 
--- a/numba/cpython/randomimpl.py
+++ b/numba/cpython/randomimpl.py
@@ -1465,7 +1465,7 @@ def multinomial(n, pvals, size=None):
             # distribution over the remaining number of experiments.
             for j in range(0, plen - 1):
                 p_j = pvals[j]
-                n_j = fl[i + j] = np.random.binomial(n_experiments, p_j / p_sum)
+                n_j = fl[i + j] = np.random.binomial(n_experiments, min(p_j / p_sum, 1))
                 n_experiments -= n_j
                 if n_experiments <= 0:
                     # Note the output was initialized to zero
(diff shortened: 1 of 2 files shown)


esc:
It may be worth checking how the Numpy variant of this is implemented. Perhaps they have some numerical tricky too.

juliusbierk:
> It may be worth checking how the Numpy variant of this is implemented. Perhaps they have some numerical tricky too.

Hhmmm... they definitely also have issues with it:
https://github.com/numpy/numpy/blob/main/numpy/random/mtrand.pyx#L4239

But they don't seem to have to cap the probability at 1:
https://github.com/numpy/numpy/blob/ea7b0e281a8d221c847abb2cd760cb067fa8543b/numpy/random/src/distributions/distributions.c#L1679

esc:
@juliusbierk thank you for your efforts on this front! Your efforts to improve Numba are appreciated. I've moved the PR back to in-progress. Please do let us know when you would like a review!

juliusbierk:
Thanks @esc. Sorry for all the commits. I thought this was a really simple change, so I just did it online in github, but apparently I had misunderstood some inner workings of numba.
Anyway: I think I fixed it all now, and the PR is ready for review.

esc:
@juliusbierk thank you for submitting this. I have added it to the queue for review. It may be some time however until someone finds the time to review it. Thank you for your patience.

juliusbierk:
@esc Any update?

esc:
> @esc Any update?

Not at this time, no. 

@stuartarchibald do you think this could profit from https://en.wikipedia.org/wiki/Kahan_summation_algorithm ?

esc:
Judging from https://github.com/numpy/numpy/blob/main/numpy/random/mtrand.pyx#L4274 -- it does seem like NumPy is using Kahan sum for this? At least for some aspects?

stuartarchibald:
@esc @juliusbierk Apologies for the delay, I've just taken a look at this patch and the reference implementation (NumPy). 

There's evidently issues with numerical sensitivity and these will likely be exacerbated through algorithmic difference, therefore I think the first thing to do is adjust the underlying Numba implementation so as to better match the implementation in NumPy.

I also suspect that whilst entirely well intentioned, adding absolute value checks like https://github.com/numba/numba/pull/7551/files#diff-342c40f4687a52b5620fde4998e02c94263bbc46f9b0802d1a5e60d5c99311deR1457 that do not correlate to the reference implementation will unfortunately lead to hard to track down numerical discrepancies.

YannisZa:
Hi there @stuartarchibald , @esc . Are there any plans to fix this soon? Thank you.

georgewambold:
@juliusbierk Just curious if you have any updates on this PR. I just ran into that same `ValueError: binomial(): p outside of [0, 1]` issue.

juliusbierk:
@georgewambold I am not moving on this at the moment. The submitted PR does fix the issue, but as pointed out by 
@esc  and @stuartarchibald, it does not match exactly the numpy implementation, which e.g. uses Kahan summutation.
To be clear, the current implementation certainly also does not match the numpy implementation (since it fails).
Hopefully someone will take over the issue and fix it in line with how it is done in numpy. Otherwise, at some point I will have a look again.

stuartarchibald:
Given the efforts being made to support `np.random.Generator` and in the process update all the RNG implementations, it might be a good idea to roll fixing this issue into that work. CC @kc611 what do you think?

kc611:
I think going with a implementation for Kahan's summation would be the way to go over here. In the past, we did have instances where slight numerical differences ended up being a problem. 

Even with respect to efforts towards  `np.random.Generator`, when `multinomial` is implemented for `Generator`s we will anyways require a `kahan_sum` method as it is a part of both implementations. 

https://github.com/numpy/numpy/blob/7a3066c39002404e5c0b7ff316317604ef3f603b/numpy/random/_generator.pyx#L3924

@juliusbierk If you're fine with it, Can I take over this PR and go by Kahan's summation instead ?

PR #7552, opened 2021-11-11
Prepend the inlined function name to inlined variables.

### LINKED ISSUE #7365: Variable name uniqueness issue in inlining (introduced in Numba 0.54)
## Reporting a bug

`mk_unique_var` was removed from the line below in Numba 0.54. This is causing variable name conflict issues in Bodo due to heavy use of inlining. Reproducer is not easy to provide but possible if necessary (doesn't seem necessary to me at the moment). What was the reason for removing `mk_unique_var`?

https://github.com/numba/numba/blob/7f72315f0bf672729b5a6eaf004aa9061e8bf0f5/numba/core/inline_closurecall.py#L398

### LINKED ISSUE #7380: Inlining functions with loops with the same induction variable name in python source causes parfors fail.
## Reporting a bug

This:
```python
import numpy as np

@njit(inline="always")
def bar(x):
    for i in range(x.size):
        x[i] += 1

@njit(parallel=True)
def foo(a):
    for i in prange(a.shape[0]):
        bar(a[i])

a = np.ones((10, 10))
foo(a)
```
Does this:
```
numba.core.errors.UnsupportedRewriteError: Failed in nopython mode pipeline (step: convert to parfors)
Overwrite of parallel loop index
```
debug info is here:
https://github.com/numba/numba/issues/7365#issuecomment-913774965

this is technically a regression in 0.54, however it seemed to rely on a side effect/something invalid to work. Discussion required...!

CC @DrTodd13 

CC @JSKenyon

stuartarchibald:
As title.

Fixes #7380
Fixes #7365

### DIFF
--- a/numba/core/inline_closurecall.py
+++ b/numba/core/inline_closurecall.py
@@ -53,6 +53,20 @@ def callee_ir_validator(func_ir):
                 raise errors.UnsupportedError(msg, loc=stmt.loc)
 
 
+def _created_inlined_var_name(function_name, var_name):
+    """Creates a name for an inlined variable based on the function name and the
+    variable name. It does this "safely" to avoid the use of characters that are
+    illegal in python variable names as there are occasions when function
+    generation needs valid python name tokens."""
+    inlined_name = f'{function_name}.{var_name}'
+    # Replace angle brackets, e.g. "<locals>" is replaced with "_locals_"
+    new_name = inlined_name.replace('<', '_').replace('>', '_')
+    # The version "version" of the closure function e.g. foo$2 (id 2) is
+    # rewritten as "foo_v2". Further "." is also replaced with "_".
+    new_name = new_name.replace('.', '_').replace('$', '_v')
+    return new_name
+
+
 class InlineClosureCallPass(object):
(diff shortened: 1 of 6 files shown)


numba/core/inline_closurecall.py
```diff
@@ -599,7 +600,8 @@ def inline_closure_call(func_ir, glbls, block, i, callee, typingctx=None,
     var_dict = {}
     for var in callee_scope.localvars._con.values():
         if not (var.name in callee_code.co_freevars):
-            new_var = scope.redefine(mk_unique_var(var.name), loc=var.loc)
+            inlined_name = f'{callee_ir.func_id.unique_name}.{var.name}'
+            new_var = scope.redefine(inlined_name, loc=var.loc)
```
stuartarchibald, line 604: Note: probably need to write a test for explicit closure inlining, it uses this path opposed to the function inliner.

esc (approved):
I checked this by trying all the unit tests to make sure they do test the correct thing. LGTM!

stuartarchibald:
Thanks for the review @esc.

Issue #4402, opened 2019-08-02
meta issue: fp16 support


sklam:
This meta issue lists the concrete tasks needed to implement half-precision floating type. (#4395).

llvmlite: 
- [ ] add `half` type

numba:
- [ ] add `float16` type and conversion to/from `numpy.float16`.
- [ ] setup implicit casting rules.  (note: survey what numpy does with float16 casting).
- [ ] implement basic arithmetic ops.
    - [ ] add, sub, mul
    - [ ] div, rem (separate item because CUDA PTX do not provide these it seems)
- [ ] decide what to do with other math functions? implicit cast to float32 variant?

seibert:
Related issue is how to handle bfloat16, which seems to be gaining traction and will have limited AVX-512 support on very recent Intel CPU architectures, and is also used on the Google TPU.  Haven't seen any GPU hardware support for bfloat16 yet, although AMD's ROCm seems to have added software support.

njwhite:
- [ ] Nvidia [to add support](https://docs.nvidia.com/cuda/nvvm-ir-spec/index.html#type-system) for `f16` ("`half`") types in NVVM IR - otherwise need to sprinkle `llvm.convert.to.fp16.f32` / `llvm.convert.from.fp16.f32` everywhere (and lower `types.f16` in numba code to `i16`s in IR).

MurrayData:
Just a note that CuPy now supports float16 in user defined kernels.

```
>>> import cupy as cp
>>> c = cp.random.random(1000000).astype(cp.float16)

@cp.fuse()
... def squared_diff(x, y):
...     return (x - y) * (x - y)

>>> square(c)

array([0.848  , 0.637  , 0.05655, ..., 0.02043, 0.5337 , 0.8125 ],
      dtype=float16)
```

It would be good to have float16 in Numba kernels to prepare data for deep learning models and RAPIDS Numba based UDFs. At present RAPIDS converts float32 to float16 as a workaround.

gmarkall:
I can't edit the issue to tick the box, but it looks like:

> llvmlite:
> 
> * [ ]  add `half` type

is done in https://github.com/numba/llvmlite/pull/509, and

[...]

PR #7556, opened 2021-11-11
Testhound/fp16 comparison

testhound:
This pull request adds support for the following 16-bit floating point operations to the CUDA target:

heq - (fp16 equal)
hne - (fp16 not equal)
hgt - (fp16 greater than)
hge - (fp16 greater than or equal to)
hlt - (fp16 less than)
hle - (fp16 less than or equal to)
hmax -(fp16 maximum)
hmin - (fp16 minimum)

This PR partly addresses the meta issue "add fp16 support": https://github.com/numba/numba/issues/4402.

Note: this PR depends on this fp16 PR which is currently in review: https://github.com/numba/numba/pull/7460


testhound:
gpuci run tests

testhound:
gpuci run tests

testhound:
gpuci run tests

testhound:
gpuci run tests

numba/cuda/cudaimpl.py
```diff
@@ -523,6 +523,92 @@ def ptx_hfma(context, builder, sig, args):
     asm = ir.InlineAsm(fnty, "fma.rn.f16 $0,$1,$2,$3;", "=h,h,h,h")
     return builder.call(asm, args)
 
+
+def lower_fp16_comparison(fn, op):
+    @lower(fn, types.float16, types.float16)
+    def ptx_fp16_comparison(context, builder, sig, args):
+        temp_reg = '__$$temp3'
```
gmarkall, line 530: Is this hard coding a register name? Does nvcc use `__$$temp3` every time?

gmarkall (changes requested):
I started reviewing the code and I've left some comments. I noticed that the following:

```python
from numba import cuda

@cuda.jit('void(boolean[:], float16, float16, float16)')
def test(r, a, b, c):
    r[0] = cuda.fp16.hlt(a, b) and cuda.fp16.hlt(b, c)
```

results in:

```
  File "/home/gmarkall/numbadev/issues/7556/repro.py", line 5, in <module>
    def test(r, a, b, c):
  File "/home/gmarkall/numbadev/numba/numba/cuda/decorators.py", line 99, in _jit
    return Dispatcher(func, [func_or_sig], targetoptions=targetoptions)
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 805, in __init__
    self.compile(sigs[0])
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 1084, in compile
    kernel = _Kernel(self.py_func, argtypes, link=self.link,
    return func(*args, **kwargs)
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 390, in __init__
    cres = compile_cuda(self.py_func, types.void, self.argtypes,
    return func(*args, **kwargs)
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 200, in compile_cuda
    cres = compiler.compile_extra(typingctx=typingctx,
  File "/home/gmarkall/numbadev/numba/numba/core/compiler.py", line 693, in compile_extra
  File "/home/gmarkall/numbadev/numba/numba/core/compiler.py", line 429, in compile_extra
  File "/home/gmarkall/numbadev/numba/numba/core/compiler.py", line 497, in _compile_bytecode
    return self._compile_core()
  File "/home/gmarkall/numbadev/numba/numba/core/compiler.py", line 472, in _compile_core
    raise e
  File "/home/gmarkall/numbadev/numba/numba/core/compiler.py", line 463, in _compile_core
    pm.run(self.state)
  File "/home/gmarkall/numbadev/numba/numba/core/compiler_machinery.py", line 349, in run
    raise e
  File "/home/gmarkall/numbadev/numba/numba/core/compiler_machinery.py", line 341, in run
    return func(*args, **kwargs)
  File "/home/gmarkall/numbadev/numba/numba/core/compiler_machinery.py", line 296, in _runPass
  File "/home/gmarkall/numbadev/numba/numba/core/compiler_machinery.py", line 269, in check
  File "/home/gmarkall/numbadev/numba/numba/core/typed_passes.py", line 396, in run_pass
    lower.lower()
  File "/home/gmarkall/numbadev/numba/numba/core/lowering.py", line 183, in lower
  File "/home/gmarkall/numbadev/numba/numba/core/lowering.py", line 237, in lower_normal_function
  File "/home/gmarkall/numbadev/numba/numba/core/lowering.py", line 266, in lower_function_body
    self.lower_block(block)
  File "/home/gmarkall/numbadev/numba/numba/core/lowering.py", line 280, in lower_block
    self.lower_inst(inst)
  File "/home/gmarkall/numbadev/numba/numba/core/lowering.py", line 433, in lower_inst
    self.storevar(val, inst.target.name, argidx=argidx)
  File "/home/gmarkall/numbadev/numba/numba/core/lowering.py", line 1427, in storevar
    raise AssertionError(msg)
AssertionError: Storing i8 to ptr of i1 ('$phi28.0.2'). FE type bool
```

docs/source/cuda-reference/kernel.rst
```diff
@@ -502,17 +502,51 @@ These functions return a 16-bit floating point result.
...
 
    Perform the negation operation ``-a`` on the 16-bit floating point argument.
-   Maps to the ``neg.f16`` PTX instruction.
 
    Returns the 16-bit floating point result of the negation.
 
 .. function:: numba.cuda.fp16.habs (a)
 
    Perform the absolute value operation ``|a|`` on the 16-bit floating point argument.
-   Maps to the ``abs.f16`` PTX instruction.
```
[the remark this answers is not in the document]
testhound: Not accidental since lowering does not necessarily produce a 'abs.f16 instruction. It can generate either a 'abs.f16' or 'and.b16'.

testhound:
gpuci run tests

testhound:
gpuci run tests

gmarkall (changes requested):
Many thanks for the updates.

I now note that:

```python
from numba import cuda


@cuda.jit(device=True)
def test1(x, y):
    return cuda.fp16.hlt(x, y)


@cuda.jit(device=True)
def test2(x, y):
    return cuda.fp16.hlt(x, y)


@cuda.jit('void(boolean[:], float16, float16, float16)')
def test(r, a, b, c):
    r[0] = test1(a, b) and test2(b, c)
```

results in:

```
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudadrv/driver.py", line 2682, in add_ptx
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudadrv/driver.py", line 319, in safe_cuda_api_call
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudadrv/driver.py", line 384, in _check_ctypes_error

  File "/home/gmarkall/numbadev/issues/7556/repro2.py", line 15, in <module>
    def test(r, a, b, c):
  File "/home/gmarkall/numbadev/numba/numba/cuda/decorators.py", line 99, in _jit
    return Dispatcher(func, [func_or_sig], targetoptions=targetoptions)
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 807, in __init__
    self.compile(sigs[0])
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 1095, in compile
    kernel.bind()
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 468, in bind
  File "/home/gmarkall/numbadev/numba/numba/cuda/codegen.py", line 201, in get_cufunc
  File "/home/gmarkall/numbadev/numba/numba/cuda/codegen.py", line 174, in get_cubin
    linker.add_ptx(ptx.encode())
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudadrv/driver.py", line 2685, in add_ptx
ptxas application ptx input, line 61; error   : Duplicate definition of variable '__$temp1'
```

Could you comment on this please?

Although the original failing case now works, it needs a test case adding to ensure its continued correctness (I don't think I see one - did I miss it?). If it is not present, could you add it please?

numba/cuda/cudaimpl.py
```diff
@@ -523,6 +524,114 @@ def ptx_hfma(context, builder, sig, args):
...
+                                f"selp.u16 $0, 1, 0, {temp_reg};",
+                                '=h')
+        selp = builder.call(selp_asm, [])
+
+        zero = context.get_constant(types.int16, 0)
+        cmp = builder.icmp_unsigned("!=",
+                                    builder.bitcast(selp, ir.IntType(16)),
+                                    zero)
+
+        #zext = builder.zext(cmp, ir.IntType(8))
```
gmarkall, line 579:
```suggestion
```

gmarkall:
I believe the declaration and use of your temporary predicate registers needs to be scoped with `{}`: https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#control-flow-instructions-curly-braces

For example, in NVCC-generated PTX using temporary predicate registers, I see:

```asm
	{ .reg .pred __$temp3;
  setp.lt.f16  __$temp3, %rs8, %rs9;
  selp.u16 %rs7, 1, 0, __$temp3;}
```

testhound:
gpuci run tests

gmarkall:
gpuci run tests

testhound:
gpuci run tests

sklam:
@gmarkall, it looks like all the review comments are addressed. Can you confirm?

numba/cuda/cudaimpl.py
isVoid: It looks like it's declaring a register with predicate type named `__$$temp3`.

gmarkall:
gpuci run tests

numba/cuda/cudaimpl.py
gmarkall: The scoping of the predicate (in the inline assembly on line 534) resolves this.

gmarkall (changes requested):
Thanks for the updates. I think there are some further changes to make here:

* A couple of small comments on the diff about syntax in a docstring.
* Additional tests for using multiple predicates within a function, as suggested in https://github.com/numba/numba/pull/7556#pullrequestreview-850587849
   - Also, some of the present names of the test functions are a bit too generic given that they're in a file with lots of other tests - e.g. `test1` and `test2` don't mean much in the context of a file with tests for all different kinds of intrinsics.
* Simplification of the lowering - the present implementations seem quite complex:
   -  Multiple inline assembly blocks are used for each function - this doesn't seem to be necessary, and complicates both the generating code and the generated PTX.
   - There's no need to assign a result then discard it (instead of `_= builder.call(...)`, just write `builder.call(...)`.
   - The temporary predicate register could be renamed so it's clearer what its purpose is (instead of just `__$temp3`, which, although similar to nvcc, is a little arbitrary)
   - The comparison code used by the comparisons and min / max is practically identical - min and max should be able to reuse the comparisons.
* We can't use the current device's compute capability to control the lowering (as established when we were working through #7460) - we don't know that we're lowering for the current device, or that there even is a current device. This unfortunately means we don't have a way to use the PTX float16 min/max instructions when the compute capability is 8.0 - I'll open an issue for this.

Implementations of the suggested changes are in this branch: https://github.com/gmarkall/numba/tree/fp16-comparison, with the changes on top of this PR looking like: https://github.com/numba/numba/compare/96d31b9713a85c3e7c6cfccfaffba862e39ba417...ae0998a66e6e7069db16571ca36d2e8edf8a0e45 - feel free to pull those commits into this branch if you agree with the changes.

gmarkall:
> * We can't use the current device's compute capability to control the lowering (as established when we were working through #7460) - we don't know that we're lowering for the current device, or that there even is a current device. This unfortunately means we don't have a way to use the PTX float16 min/max instructions when the compute capability is 8.0 - I'll open an issue for this.



Created #7954 for this.

gmarkall:
> * We can't use the current device's compute capability to control the lowering (as established when we were working through #7460) - we don't know that we're lowering for the current device, or that there even is a current device. This unfortunately means we don't have a way to use the PTX float16 min/max instructions when the compute capability is 8.0 - I'll open an issue for this.



Created #7954 for this.

testhound:
gpuci run tests

numba/cuda/tests/cudapy/test_intrinsics.py
```diff
@@ -118,6 +119,77 @@ def simple_habs_scalar(ary, a):
...
+def hlt_func_1(x, y):
+    return cuda.fp16.hlt(x, y)
+
+
+@cuda.jit(device=True)
+def hlt_func_2(x, y):
+    return cuda.fp16.hlt(x, y)
+
+
+def test_predicate_reg(r, a, b, c):
```
gmarkall, line 156: Is this function dead code?

gmarkall:
Thanks for the updates. I have verified  that https://github.com/numba/numba/pull/7556/commits/48b0c324cc5e2707e4b65bde7d65356722e9b7ba subsumes the changes in https://github.com/gmarkall/numba/tree/fp16-comparison. For future PRs if this situation arises where there are suggested commits, could you please merge / cherry-pick the suggested commits into your branch rather than re-committing the same changes as a new commit? This will make it easy to verify the changes are the same (or will highlight where there are additions / differences in subsequent commits) and also preserve authorship for the purpose of the review - because those changes come verbatim from my commits, I think they're significant enough that I should get someone else to review them too (as a review of my own code is limited compared to someone else's view on it).

numba/cuda/tests/cudapy/test_intrinsics.py
testhound: Yes. Will fix and push update shortly.

testhound:
> Thanks for the updates. I have verified that [48b0c32](https://github.com/numba/numba/commit/48b0c324cc5e2707e4b65bde7d65356722e9b7ba) subsumes the changes in https://github.com/gmarkall/numba/tree/fp16-comparison. For future PRs if this situation arises where there are suggested commits, could you please merge / cherry-pick the suggested commits into your branch rather than re-committing the same changes as a new commit? This will make it easy to verify the changes are the same (or will highlight where there are additions / differences in subsequent commits) and also preserve authorship for the purpose of the review - because those changes come verbatim from my commits, I think they're significant enough that I should get someone else to review them too (as a review of my own code is limited compared to someone else's view on it).

@gmarkall input received and I will merge/cherry-pick in the future.

testhound:
gpuci run tests

gmarkall (approved):
Many thanks for the updates. I think this looks good.

gmarkall:
@esc could this have a CUDA buildfarm run please?

esc:
> @esc could this have a CUDA buildfarm run please?

`Build numba_smoketest_cuda_yaml_129 has started`

esc:
> > @esc could this have a CUDA buildfarm run please?
> 
> `Build numba_smoketest_cuda_yaml_129 has started`

All green 🟢

PR #7564, opened 2021-11-15
Update Windows and OSX CI images.

stuartarchibald:
macOS-10.14 and vs2017-win2016 are being deprecated by Azure CI.
This patch updates the images used to the oldest image that is
still supported.


stuartarchibald:
Implemented in https://github.com/numba/numba/pull/7650 and https://github.com/numba/numba/pull/7628, closing.

PR #7570, opened 2021-11-16
Update the docs and error message for errors when importing Numba.

stuartarchibald:
This updates the error message users see when the import of Numba
fails. It directs users to new documentation that outlines common
issues and has suggested fixes. This error is now considered
non-reportable as pretty much all occurences are due to one of the
few cases that are now documented. Should users have an issue that
is not covered, the docs direct them to the discourse forum to ask
for assistance.

### DIFF
--- a/docs/source/user/faq.rst
+++ b/docs/source/user/faq.rst
@@ -3,6 +3,75 @@
 Frequently Asked Questions
 ==========================
 
+Installation
+============
+
+Numba could not be imported
+---------------------------
+
+If you are seeing an exception on importing Numba with an error message
+that starts with::
+
+    ImportError: Numba could not be imported.
+
+here are some common issues and things to try to fix it.
+
+#. Your installation has more than one version of Numba a given environment.
+
+   Common ways this occurs include:
+
+   * Installing Numba with conda and then installing again with pip.
+   * Installing Numba with pip and then updating to a new version with pip (pip
+     re-installations don't seem to always clean up very well).
+
+   To fix this the best approach is to create an entirely new environment and
+   install a single version of Numba in that environment using a package manager
+   of your choice.
+
+#. Your installation has Numba for Python version X but you are running with
+   Python version Y.
+
+   This occurs due to a variety of Python environment mixup/mismatch problems.
+   The most common mismatch comes from installing Numba into the
+   site-packages/environment of one version of Python by using a base or
+   system installation of Python that is a different version, this typically
+   happens through the use of the "wrong" ``pip`` binary. This will obviously
+   cause problems as the C-Extensions on which Numba relies are bound to
+   specific Python versions. A way to check if this likely the problem is to
+   see if the path to the python at::
+
+       python -c 'import sys; print(sys.executable)'
+
+   matches the path to your installation tool and/or matches the reported
+   installation location and if the Python versions match up across all of
+   these. Note that Python version ``X.Y.A`` is compatible with ``X.Y.B``.
+
+   To fix this the best approach is to create an entirely new environment and
+   ensure that the installation tool used to install Numba is the one from that
+   environment/the Python versions at install and run time match.
+
+#. Your core system libraries are too old.
+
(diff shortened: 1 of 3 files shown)


numba/core/typeconv/typeconv.py
```diff
@@ -3,18 +3,18 @@
...
-    from numba.core.errors import feedback_details as reportme
-    import sys
-    url = "https://numba.pydata.org/numba-doc/latest/developer/contributing.html"
+    base_url = "https://numba.readthedocs.io/en/stable"
+    dev_url = f"{base_url}/developer/contributing.html"
+    user_url = f"{base_url}/user/faq.html#numba-could-not-be-imported"
     dashes = '-' * 80
-    msg = ("Numba could not be imported.\nIf you are seeing this message and "
-           "are undertaking Numba development work, you may need to re-run:\n\n"
-           "python setup.py build_ext --inplace\n\n(Also, please check the "
```
esc, line 12: I would argue that including build instructions here may be a form of unneeded replication. I think it will suffice to just point folks at the development docs. Everyone will end up having their own way to build Numba and having multiple sources of instructions will only be confusing.

docs/source/user/faq.rst
```diff
@@ -3,6 +3,75 @@
...
+   installation location and if the Python versions match up across all of
+   these. Note that Python version ``X.Y.A`` is compatible with ``X.Y.B``.
+
+   To fix this the best approach is to create an entirely new environment and
+   ensure that the installation tool used to install Numba is the one from that
+   environment/the Python versions at install and run time match.
+
+#. Your core system libraries are too old.
+
+   This is a somewhat rare occurence, but there are occasions when a very old
```
esc, line 55:
```suggestion
   This is a somewhat rare occurrence, but there are occasions when a very old
```

docs/source/user/faq.rst
```diff
@@ -3,6 +3,75 @@
...
+   Python version Y.
+
+   This occurs due to a variety of Python environment mixup/mismatch problems.
+   The most common mismatch comes from installing Numba into the
+   site-packages/environment of one version of Python by using a base or
+   system installation of Python that is a different version, this typically
+   happens through the use of the "wrong" ``pip`` binary. This will obviously
+   cause problems as the C-Extensions on which Numba relies are bound to
+   specific Python versions. A way to check if this likely the problem is to
+   see if the path to the python at::
```
esc, line 41:
```suggestion
   see if the path to the ``python`` executable at::
```

esc: maybe 'binary' is even better than ''executable'.  It would be coherent with the vocabulary already in use.

esc (changes requested):
It's looking good, thank you! I left a few remarks for your consideration.

numba/core/typeconv/typeconv.py
stuartarchibald: They were here originally as an incompatible local source build was quite common (and maybe still is, we can't know because this message tells people what to do). Typically, a local source based development environment couldn't load this DSO because of python version mismatch or internal API changes due to development which leads to missing symbols. Perhaps this should go into the FAQ doc?

esc: Perhaps, yes.

stuartarchibald: Done as part of: 9d4a5f0

The error message now reads:
```
ImportError: Numba could not be imported.
If you are seeing this message and are undertaking Numba development work, you may need to rebuild Numba.
Please see the development set up guide:

https://numba.readthedocs.io/en/stable/developer/contributing.html.

If you are not working on Numba development, the original error was: 'some problem'.
For help, please visit:

https://numba.readthedocs.io/en/stable/user/faq.html#numba-could-not-be-imported

```

esc: perfect!

esc (approved):
LGTM

PR #7580, opened 2021-11-18
Fix #7507. catch `NotImplementedError ` in `.get_function()` 

### LINKED ISSUE #7507: Numba is raising a `NotImplementedError` instead of the `NumbaNotImplementedError` equivalent

sklam:
Exception handler should be catching `NotImplementedError` instead.

### DIFF
--- a/numba/core/base.py
+++ b/numba/core/base.py
@@ -560,7 +560,7 @@ class BaseContext(object):
             # It's a type instance => try to find a definition for the type class
             try:
                 return self.get_function(type(fn), sig)
-            except errors.NumbaNotImplementedError:
+            except NotImplementedError:
                 # Raise exception for the type instance, for a better error message
                 pass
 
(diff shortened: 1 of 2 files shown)


stuartarchibald:
CC @guilhermeleobas Does this fix the issue you were seeing a few weeks back? xref: https://github.com/numba/numba/issues/7507

stuartarchibald:
Local testing of MWR in https://github.com/numba/numba/issues/7507#issuecomment-953165889 confirms this patch fixes that particular issue.

stuartarchibald (approved):
Thanks for finding and fixing this @sklam, patch and test (which is based on the MWR) look good. Waiting on @guilhermeleobas to confirm against RBC (thanks in advance for testing!).

guilhermeleobas:
I can confirm this change fix the issue I faced in RBC.

stuartarchibald:
> I can confirm this change fix the issue I faced in RBC.

Great, thanks for checking @guilhermeleobas, much appreciated!

PR #7582, opened 2021-11-18
Use get_parfor_writes to detect illegal array access that prevents fusion.

### LINKED ISSUE #7578: Compilation with `parallel=True` gives garbage results for a very simply parallelizable function
I have a relatively simple (and seemingly easily parallelizable) function for which I get different results when compiling with `parallel=True` and `parallel=False`. I guess that there must be some race conditions at play, but I really cannot see why.

I am not 100% sure whether it is a bug or just some fault in my code, but it was suggested to me on [SO](https://stackoverflow.com/questions/69958049/numba-parallel-true-produces-corrupted-results) that this might be a bug.

Below is a reproducible example.

```python
import numpy as np

def ubcm_jacobian(
    D: np.ndarray,
    b: np.ndarray
) -> np.ndarray:
    """Jacobian of UBCM negative log-likelihood.

    Parameters
    ----------
    D
        Degree sequence array.
    b
        Model parameters.
    """
    B = np.empty_like(b)
    b = np.exp(b)

    for i in prange(len(b)):
        bi   = b[i]
        x    = bi * b
        B[i] = (x / (1 + x)).sum()

    b2 = b**2
    B -= b2 / (1 + b2)

    return B - D


ubcm_not_parallel = njit(ubcm_jacobian, boundscheck=False, nogil=True, parallel=False)
ubcm_parallel = njit(ubcm_jacobian, boundscheck=False, nogil=True, parallel=True)

# Sample data
D = np.arange(1, 11)
b = np.log(D)

# Results
x1 = ubcm_not_parallel(D, b)
x2 = ubcm_parallel(D, b)

assert np.allclose(x1, x2), "results are not equal"
```

# Expected behavior

Of course the expected behavior is that the function should return identical results (up to the floating point precision) in both cases.

DrTodd13:
Resolves #7578 

In parfor fusion, there is a check that something written in one parfor not be used in the second parfor.  There was a case in issue #7578 that an array was written in the first parfor and analysis of that parfor said that the array was "used" but not written in that parfor.  Therefore, the second parfor which also used the same array didn't see a conflict and did the fusion.  If you just make the array written in the first parfor loop body as a "def" then the array isn't passed to the parfor gufunc and ends up undefined.  So, what I am trying here is to make the array both def and use so that the parfors don't fuse but the array is passed as a parameter to the gufunc.

This is in-progress at the moment as I expect this to break legitimate instances of fusion because the real problem with this example is that indices outside of the parfor index are used in the second parfor which is the only thing that should prevent fusion in this case.  I want to see what the tests say to see which tests fail due to this and if none do then to create one that should be allowed.


ehsantn:
Can something like `get_stmt_writes` be used instead? Use/def and liveness are used in many places and messing with them is potentially very dangerous. Setitem doesn't really have def semantics.

DrTodd13:
> Can something like `get_stmt_writes` be used instead? Use/def and liveness are used in many places and messing with them is potentially very dangerous. Setitem doesn't really have def semantics.

I did something similar now.  Please have a look.

ehsantn:
Didn't go through the details but seems like this is on the right track. Thanks.

numba/parfors/parfor.py
```diff
@@ -4145,13 +4217,33 @@ def get_user_varname(v):
...
-        dprint("try_fuse: parfor2 depends on parfor1 body")
-        msg = ("- fusion failed: parallel loop %s has a dependency on the "
-                "body of parallel loop %s. ")
-        report = FusionReport(parfor1.id, parfor2.id,
-                                msg % (parfor1.id, parfor2.id))
-        return None, report
+    overlap = p1_body_defs.intersection(p2_uses)
+    # overlap are those variable defined in first parfor and used in the second
+    if len(overlap) != 0:
+        is_safe = True # Assume safe until proven otherwise.
```
sklam, line 4223: Can it default to be unsafe instead for a more conservative behavior?

It seems it just need all `overlap` variables to be of `ArrayCompatible` and in `p2arraynotindexed`

DrTodd13: I can do the following:

unsafe_var = [not isinstance(typemap[x], types.ArrayCompatible) or x in p2arraynotindexed for x in overlap]
if any(unsafe_var):

That is simpler looking but this does have potential to be slower than what I wrote that gives up as soon as it finds the first problem.  The new line above would potentially compute some useless stuff but the amount of work is trivial so I've checked in this change.  See if you like it.

lgarrison: Would a generator comprehension solve this problem? If you change the brackets to parentheses, then the body will be lazily evaluated by `any()`: `unsafe_var = (not isinstance(typemap[x], types.ArrayCompatible) or x in p2arraynotindexed for x in overlap)`

DrTodd13: Thanks.  Done.

sklam:
@DrTodd13, the changes look good. One last thing that is needed is to make a towncrier note for this PR. It will need a `docs/upcoming_changes/7582.bug_fix.rst` file with the PR description. See https://github.com/numba/numba/blob/main/docs/upcoming_changes/8792.new_feature.rst for example. Detail instructions in https://github.com/numba/numba/blob/main/docs/upcoming_changes/README.rst.

DrTodd13:
> @DrTodd13, the changes look good. One last thing that is needed is to make a towncrier note for this PR. It will need a `docs/upcoming_changes/7582.bug_fix.rst` file with the PR description. See https://github.com/numba/numba/blob/main/docs/upcoming_changes/8792.new_feature.rst for example. Detail instructions in https://github.com/numba/numba/blob/main/docs/upcoming_changes/README.rst.

It seems to show an error in the towncrier source now.  Suggestions?

sklam:
towncrier is broken. don't worry about the error.

esc:
towncrier support was fixed with https://github.com/numba/numba/pull/9069

esc:
> towncrier support was fixed with #9069

I think you'll need to merge in `main` to get towncrier to work again with the patches on `main`.

esc:
@DrTodd13 there is a bug in the filename for towncrier: `7582.big_fix.rst` should be `7582.bug_fix.rst`, can you fix?

sklam (approved):
LGTM

esc:
@DrTodd13 thank you for the patch. @sklam thank you for the review. I have marked this as RTM now.

esc:
@DrTodd13 after merging #9037 this now conflicts with `main`.

DrTodd13:
> @DrTodd13 after merging #9037 this now conflicts with `main`.

Resolved.

sklam:
(CI failed due to conda timeout. i've restarted failing builds)

sklam (approved):

PR #7583, opened 2021-11-19
Make numba.types.Optional __str__ less verbose.

stuartarchibald:
As title. Whilst well intentioned, the verbose string repr makes
it harder to debug issues with Optional types.

### DIFF
--- a/numba/core/types/misc.py
+++ b/numba/core/types/misc.py
@@ -218,9 +218,6 @@ class Optional(Type):
         name = "OptionalType(%s)" % self.type
         super(Optional, self).__init__(name)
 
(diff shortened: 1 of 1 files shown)


stuartarchibald:
Impact on the representation is as follows:
Without this patch:
```
$ python -c 'from numba import types; print(types.Optional(types.float64))'
OptionalType(float64) i.e. the type 'float64 or None'
```
With this patch:
```
$ python -c 'from numba import types; print(types.Optional(types.float64))'
OptionalType(float64)
```

sklam (approved):
LGTM!

PR #7584, opened 2021-11-21
Export NRT_MemInfo_new

Hardcode84:
* We need it for our extension to allocate arrays with custom dtor

### DIFF
--- a/numba/core/runtime/_nrt_pythonmod.c
+++ b/numba/core/runtime/_nrt_pythonmod.c
@@ -168,6 +168,7 @@ declmethod(MemInfo_alloc_safe_aligned_external);
 declmethod_internal(_nrt_get_sample_external_allocator);
 declmethod(MemInfo_alloc_dtor_safe);
 declmethod(MemInfo_call_dtor);
+declmethod(MemInfo_new);
 declmethod(MemInfo_new_varsize);
 declmethod(MemInfo_new_varsize_dtor);
 declmethod(MemInfo_varsize_alloc);


sklam:
`nrt_external.h` from https://github.com/numba/numba/pull/4554 should satisfy the usecase for this.

DrTodd13:
My suggestion in the meeting was to add a function, manage_memory_external, that takes does the same thing as manage_memory but takes an extra NRT_ExternalAllocator parameter and then puts that into the allocated object.

Hardcode84:
https://github.com/numba/numba/blob/master/numba/core/runtime/nrt_external.h#L46

* `manage_memory` won't work for me as I need to pass additional data to dtor
* `allocate_external` should probably work, I need to check it
* Most convenient way would be probably to have smth like `NRT_MemInfo* (*manage_memory2)(void *data, NRT_managed_dtor_opaque_data dtor, void* dtor_data);`

Hardcode84:
Also, why `manage_memory` don't have `size` parameter (and how `MemInfo->size` is used internally)?

DrTodd13:
> https://github.com/numba/numba/blob/master/numba/core/runtime/nrt_external.h#L46
> 
> * `manage_memory` won't work for me as I need to pass additional data to dtor
> * `allocate_external` should probably work, I need to check it
> * Most convenient way would be probably to have smth like `NRT_MemInfo* (*manage_memory2)(void *data, NRT_managed_dtor_opaque_data dtor, void* dtor_data);`

I think you'd find that the extra opaque data pointer in manage_memory2 would not be sufficient.  You'd end up back at NRT_MemInfo * (manage_memory_external)(void *data, void *dtor_data, NRT_ExternalAllocator *allocator).  If you can use allocate_external to create it in the first place then that is fine but if Numba must take ownership of externally allocated memory then I think we need this new manage_memory_external function.

stuartarchibald:
Just following up on this PR. Is it still needed in the proposed form or was there some conclusion over a new functionality required?

sklam:
Is this still needed?

stuartarchibald:
> Is this still needed?

CC @Hardcode84

stuartarchibald:
Following OOB discussion with @Hardcode84, closing this as it is anticipated that it is not needed. Should it be needed in future this can be reopened. Thanks!

PR #7596, opened 2021-11-24
Fix max symbol match length for r2

stuartarchibald:
Turns out that r2 `is~` only matches up to 61 chars. This fixes
the immediate problem of making sure that a match is found. It
does not fix potential issues with aliasing symbols.

### DIFF
--- a/numba/misc/inspection.py
+++ b/numba/misc/inspection.py
@@ -62,10 +62,12 @@ def disassemble_elf_to_cfg(elf, mangled_symbol):
                 # If the elf is linked then it's necessary to seek as the
                 # DSO ctor/dtor is at the default position
                 if linked:
+                    # r2 only matches up to 61 chars?! found this by experiment!
+                    mangled_symbol_61char = mangled_symbol[:61]
                     # switch off demangle, the seek is on a mangled symbol
                     r.cmd('e bin.demangle=false')
                     # seek to the mangled symbol address
-                    r.cmd(f's `is~ {mangled_symbol}[1]`')
+                    r.cmd(f's `is~ {mangled_symbol_61char}[1]`')
                     # switch demangling back on for output purposes
                     r.cmd('e bin.demangle=true')
                 data = r.cmd('%s' % cmd) # print graph


stuartarchibald:
Without this patch:
```python
from numba import njit
import numpy as np

@njit(debug=True)
def foo(x):
    y = np.cos(x) + np.sin(x)
    z = np.sqrt(y)
    return x + y + z

foo(1.23)

print(foo.inspect_disasm_cfg(foo.signatures[0]))
```
gives:
```
$ python test_r2.py 
[0x00001050]>  # entry0 ();
... the graph of entry0
```
with this patch, the same gives:
```
$ python test_r2.py 
[0x00001120]>  # dbg.__main__::foo_241[abi:c8tJTC_2fWgEeGLSgydRTQUgiqKEZ6gEoDvQJmaQIA](double) (float64 x, int64_t arg1);
... the graph of the selected symbol
```

sklam:
Hm... how is r2 working with C++? 61 char is not that long at all with C++

stuartarchibald:
> Hm... how is r2 working with C++? 61 char is not that long at all with C++

It's the symbol table list match that has the restriction. IIRC I couldn't get `sf` to work, I'll give that another try now, that'd avoid having to do a partial match.

sklam (approved):
LGTM. Manually tested.

sklam:
> > Hm... how is r2 working with C++? 61 char is not that long at all with C++
> 
> It's the symbol table list match that has the restriction. IIRC I couldn't get `sf` to work, I'll give that another try now, that'd avoid having to do a partial match.

@stuartarchibald, any update? should I merge this PR now?

stuartarchibald:
> @stuartarchibald, any update? should I merge this PR now?

@sklam given this is a pretty advanced debug feature, I think merge this as-is because it fixes the immediate issue to a reasonable degree. I think in practice an aliasing situation would be quite rare (would require two compilations with different flags, which has only recently been accommodated and is quite hard to achieve). I'll try and fix it properly if I get a chance to before 0.55RC1 or even during the RC period as it would be a bug fix.

PR #7603, opened 2021-11-27
Fix list.insert() for refcounted values

### LINKED ISSUE #7553: List.insert() crashes for Numpy arrays
## Reporting a bug

This code segfaults for me (on Mac OS). Didn't see an obvious issue in insert's implementation with a quick look.
```python
import numpy as np
import numba

@numba.njit
def f():
    l = [np.ones(3)]
    l.insert(0, np.ones(3))
    return l

f()
```

ehsantn:
Fixes #7553.

### DIFF
--- a/numba/cpython/listobj.py
+++ b/numba/cpython/listobj.py
@@ -191,9 +191,10 @@ class ListInstance(_ListPayloadMixin):
         # it's necessary for the dtor which just decref every slot on it.
         self.zfill(idx, self._builder.add(idx, idx.type(1)))
 
-    def setitem(self, idx, val, incref):
+    def setitem(self, idx, val, incref, decref_old_value=True):
         # Decref old data
-        self.decref_value(self.getitem(idx))
+        if decref_old_value:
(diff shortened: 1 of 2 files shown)


esc (approved):
Thank you for this fix. I have tried running the test w/o the fix applied and I receive:

```
python(26644,0x112ca8dc0) malloc: Incorrect checksum for freed object 0x7f957ce5ec80: probably modified after being freed.
Corrupt value: 0x9fffffffffffffff
python(26644,0x112ca8dc0) malloc: *** set a breakpoint in malloc_error_break to debug
Fatal Python error: Aborted

Current thread 0x0000000112ca8dc0 (most recent call first):
<no Python frame>
```

But, with the fix applied, the test passes fine. So I am giving this the green light to merge.

sklam (approved):
Thanks for the patch!

I give a 15minute attempt to see if calling `zfill` in `move` would be more elegant and I am stuck with the case of moving the slots backward. So, it can be a future thing to revisit but we need to fix this bug now.

PR #7604, opened 2021-11-28
Added multiple axes support for np.expand_dims

kc611:
This PR adds support for passing multiple axes as a tuple to `np.expand_dims()`.

```python
import numpy as np
import numba

orig_shape = (3,4,5)
orig = np.random.normal(size=orig_shape)
new_order_tuple = (1, 3)

ref = np.expand_dims(orig, new_order_tuple)

@numba.njit
def expand_dims_fn(x, new_order):
    return np.expand_dims(x, new_order)

res = expand_dims_fn(orig, new_order_tuple)

print(ref.shape) # (3, 1, 4, 1, 5)
print(res.shape) # (3, 1, 4, 1, 5)
```

### DIFF
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -5986,99 +5986,76 @@ def impl_np_array(object, dtype=None):
     return impl
 
 
-def _normalize_axis(context, builder, func_name, ndim, axis):
-    zero = axis.type(0)
-    ll_ndim = axis.type(ndim)
-
-    # Normalize negative axis
-    is_neg_axis = builder.icmp_signed('<', axis, zero)
-    axis = builder.select(is_neg_axis, builder.add(axis, ll_ndim), axis)
-
-    # Check axis for bounds
-    axis_out_of_bounds = builder.or_(
-        builder.icmp_signed('<', axis, zero),
-        builder.icmp_signed('>=', axis, ll_ndim))
-    with builder.if_then(axis_out_of_bounds, likely=False):
-        msg = "%s(): axis out of bounds" % func_name
-        context.call_conv.return_user_exc(builder, IndexError, (msg,))
-
-    return axis
-
-
-def _insert_axis_in_shape(context, builder, orig_shape, ndim, axis):
-    """
-    Compute shape with the new axis inserted
-    e.g. given original shape (2, 3, 4) and axis=2,
-    the returned new shape is (2, 3, 1, 4).
-    """
-    assert len(orig_shape) == ndim - 1
-
-    ll_shty = ir.ArrayType(cgutils.intp_t, ndim)
-    shapes = cgutils.alloca_once(builder, ll_shty)
-
-    one = cgutils.intp_t(1)
-
-    # 1. copy original sizes at appropriate places
-    for dim in range(ndim - 1):
-        ll_dim = cgutils.intp_t(dim)
-        after_axis = builder.icmp_signed('>=', ll_dim, axis)
-        sh = orig_shape[dim]
-        idx = builder.select(after_axis,
-                             builder.add(ll_dim, one),
-                             ll_dim)
-        builder.store(sh, cgutils.gep_inbounds(builder, shapes, 0, idx))
-
-    # 2. insert new size (1) at axis dimension
-    builder.store(one, cgutils.gep_inbounds(builder, shapes, 0, axis))
-
-    return cgutils.unpack_tuple(builder, builder.load(shapes))
-
-
-def _insert_axis_in_strides(context, builder, orig_strides, ndim, axis):
-    """
-    Same as _insert_axis_in_shape(), but with a strides array.
-    """
-    assert len(orig_strides) == ndim - 1
-
-    ll_shty = ir.ArrayType(cgutils.intp_t, ndim)
-    strides = cgutils.alloca_once(builder, ll_shty)
-
-    one = cgutils.intp_t(1)
-    zero = cgutils.intp_t(0)
-
-    # 1. copy original strides at appropriate places
-    for dim in range(ndim - 1):
-        ll_dim = cgutils.intp_t(dim)
-        after_axis = builder.icmp_signed('>=', ll_dim, axis)
-        idx = builder.select(after_axis,
-                             builder.add(ll_dim, one),
-                             ll_dim)
-        builder.store(orig_strides[dim],
-                      cgutils.gep_inbounds(builder, strides, 0, idx))
-
-    # 2. insert new stride at axis dimension
-    # (the value is indifferent for a 1-sized dimension, we use 0)
-    builder.store(zero, cgutils.gep_inbounds(builder, strides, 0, axis))
-
-    return cgutils.unpack_tuple(builder, builder.load(strides))
-
-
-def expand_dims(context, builder, sig, args, axis):
+def expand_dims(context, builder, sig, args):
     """
     np.expand_dims() with the given axis.
     """
     retty = sig.return_type
     ndim = retty.ndim
     arrty = sig.args[0]
+    _empty_tuple = tuple([-1] * ndim)
+
+    if isinstance(sig.args[1], types.Integer):
+        to_tuple = register_jitable(lambda axis: (axis,))
+    else:
+        to_tuple = register_jitable(lambda axis: axis)
+
+    @register_jitable
+    def normalize_axis(axis, ndim):
+        for i in range(len(axis)):
+            # In case of negative axes, we wrap them around
+            if axis[i] < 0:
+                axis = tuple_setitem(axis, i, axis[i] + ndim)
+
+            # Check for out of bound axes
+            if axis[i] < 0 or axis[i] >= ndim:
+                raise IndexError("axis out of bounds")
+        return axis
+
+    @register_jitable
+    def _build_shape_stride_tuples(_curr_tuple, ndim, axis, value):
+        axis = to_tuple(axis)
+        axis = normalize_axis(axis, ndim)
+        # 1. Initialize new tuple with -1
+        res = _empty_tuple
+
+        # 2. Replace the new tuple with given value using
+        # their index array i.e. axis
+        for i in axis:
+            if res[i] == value:
+                raise ValueError("An axis is being repeated")
+            res = tuple_setitem(res, i, value)
+
+        # 3. Incrementally put original tuple values into
+        # indices that are -1
+        idx = 0
+        for i in range(ndim):
+            if res[i] == -1:
+                res = tuple_setitem(res, i, _curr_tuple[idx])
+                idx = idx + 1
+
+        return res
 
     arr = make_array(arrty)(context, builder, value=args[0])
     ret = make_array(retty)(context, builder)
+    temp_ty = types.UniTuple(types.intp, count=sig.args[0].ndim)
 
-    shapes = cgutils.unpack_tuple(builder, arr.shape)
-    strides = cgutils.unpack_tuple(builder, arr.strides)
+    ndim_const = context.get_constant(types.intp, ndim)
+    zero = context.get_constant(types.intp, 0)
+    one = context.get_constant(types.intp, 1)
 
-    new_shapes = _insert_axis_in_shape(context, builder, shapes, ndim, axis)
-    new_strides = _insert_axis_in_strides(context, builder, strides, ndim, axis)
+    fnty = context.typing_context.resolve_value_type(_build_shape_stride_tuples)
+    sig = fnty.get_call_type(
+        context.typing_context,
+        (temp_ty, types.intp, sig.args[1], types.intp),
+        {}
+    )
+    new_shapes = context.get_function(fnty, sig)(
+        builder, (arr.shape, ndim_const, args[1], one)
+    )
+    new_strides = context.get_function(fnty, sig)(
+        builder, (arr.strides, ndim_const, args[1], zero)
+    )
 
     populate_array(ret,
                    data=arr.data,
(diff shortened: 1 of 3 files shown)


kc611:
I also wanted to add a similar support (of multiple user defined axes) for np.squeeze but I couldn't find the code where Numba has binded that function to it's respective  implementation. (At least nothing that goes by the name of sqeeze) 

Can anyone point me towards the Numba implementation for that particular function?

stuartarchibald:
@kc611 many thanks for the PR, it's now queued for review.

RE: `np.squeeze` I don't think Numba supports this at present (what is supported is largely noted in https://numba.readthedocs.io/en/stable/reference/numpysupported.html), maybe you could write the initial implementation and then add support for axes if this is important for your use case?!

The use of Numba's extension API `@overload` decorator is strongly recommended for this task. A guide to using `@overload` is here: https://numba.readthedocs.io/en/stable/extending/overloading-guide.html and the API documentation is here: https://numba.readthedocs.io/en/stable/extending/high-level.html#implementing-functions. Hope this helps.

kc611:
Hi @stuartarchibald , So I tried using `@overload` for `np.squeeze` and I think I'd be a great idea to also add it as an bound method attribute to arrays too,  i.e. we would also be able to do something like `array.squeeze()` in njit-ed functions.

So I plan to something more or less similar to following code:

```python
# This code goes in numba/np/arrayobj.py
@lower_builtin('array.squeeze', types.Array)
def array_squeeze(context, builder, sig, args):
    return args[0] #This will return the squeezed array instead

@lower_builtin('array.squeeze', types.Array, types.BaseTuple)
def array_squeeze_tuple(context, builder, sig, args):
    return args[0] #This will return the squeezed array instead

@lower_builtin('array.squeeze', types.Array, types.VarArg(types.Any))
def array_squeeze_vararg(context, builder, sig, args):
    new_sig, new_args = vararg_to_tuple(context, builder, sig, args)
    return array_squeeze_tuple(context, builder, new_sig, new_args)

@overload(np.squeeze)
def numpy_transpose(a, axes=None):

    if axes is None:
        def np_transpose_impl(a, axes=None):
            return a.squeeze()
    else:
        def np_transpose_impl(a, axes=None):
            return a.squeeze(axes)

    return np_transpose_impl
```

However the immediate problem that I run into with this approach is that we need to know the dimensions of the return signature of the array to properly bind this method, similar to how other methods are bound, for example: 

https://github.com/numba/numba/blob/8be3456a271de3e1edb9e996a722376bc249b3ae/numba/core/typing/arraydecl.py#L479-L484

In `np.squeeze` (with just the array argument) the dimensions of return signature might differ based on how many axes are 1.

So my question here is basically is it possible for us to know at the type level, which of these axes are `1` (or has zero strides or are 'broadcastable', if Numba has something like that) ? Alternatively is there a better approach for this (like not using `@lower_builtin`) ?

brandonwillard (commented):
> However the immediate problem that I run into with this approach is that we need to know the dimensions of the return signature of the array to properly bind this method, similar to how other methods are bound, for example:

Not just the dimensions, but you would need to know the explicit _shape_ of the array in order to determine whether or not a dimension could be dropped.

In Aesara, `TensorType`s include "broadcastable" information, which is essentially just partial shape information (i.e. whether or not a dimension has a shape value equal to one).  Tracking this information makes it possible to perform `squeeze` operations at the type level; however, it also requires type-level inference (e.g. all the `Op`-specific logic in `Op.make_node` that determines `TensorType.broadcastable` for the outputs).

This might not be possible in Numba without similar mechanisms.

> Alternatively is there a better approach for this (like not using `@lower_builtin`) ?

Can't `overload_method` be used instead?

kc611:
> Can't overload_method be used instead?

There are a few issues with `overload_method` too.

For instance, We cannot have return types of different dimensions in a single overload method.

Also It seems that we can't have overloads of same method with different signatures and having default arguments i.e. `dims=None` is not allowed so basically we won't be able to do `arr.squeeze()` and `arr.squeeze(dims)`.

kc611:
We can go ahead for a review with the current functionality of this PR. (Implementing `np.squeeze` will probably require more discussion and planning than the scope of this particular PR.)

gmarkall:
OK - adding this to 0.56, to try and get it in for then.

numba/tests/test_array_manipulation.py
```diff
@@ -408,11 +408,13 @@ def check_all_axes(arr):
         # 1d
         arr = np.arange(5)
         check_all_axes(arr)
+        check(arr, (1, 2))
         # 3d (C, F, A)
         arr = np.arange(24).reshape((2, 3, 4))
         check_all_axes(arr)
         check_all_axes(arr.T)
         check_all_axes(arr[::-1])
+        check(arr, (1, 3))
```
stuartarchibald, line 417: It would be good to include:
* tests ensuring that repeated axis values raise `ValueError('repeated axis')`.
* tests for out of bound axis handling
* tests for negative and zero axes (and e.g. wrap around hits the repeated axis value exception)
* tests for a large number of axes e.g. `(0, 1, 2, 3, 4, 5, 6, 7, 8)`
* tests for empty tuple e.g. `np.expand_dims(np.arange(5), ())`
* tests involving strided arrays
* tests involving 0d arrays.

stuartarchibald (commented):
Thanks for the patch. I've given this an initial review so as to provide some early feedback. I've largely concentrated on the functionality via type checking and testing and have made some suggestions with regards to these. Once those comments are addressed I'll take a look at the LLVM code gen. Thanks again!

numba/tests/test_array_manipulation.py
kc611: Added more tests in [60d59ab](https://github.com/numba/numba/pull/7604/commits/60d59abc7cdf0b3654604f24b2024dca7bf2043c)

numba/core/typing/npydecl.py
```diff
@@ -737,12 +737,14 @@ class NdExpandDims(CallableTemplate):
...
-            layout = a.layout if a.ndim <= 1 else 'A'
-            return a.copy(ndim=a.ndim + 1, layout=layout)
+            if isinstance(a, types.Array):
+                layout = a.layout if a.ndim <= 1 else 'A'
+                if isinstance(axis, types.Integer):
+                    return a.copy(ndim=a.ndim + 1, layout=layout)
+                elif isinstance(axis, types.BaseTuple):
+                    return a.copy(ndim=a.ndim + axis.count, layout=layout)
+                else:
+                    return
```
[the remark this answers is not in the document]
kc611: Resolved in [60d59ab](https://github.com/numba/numba/pull/7604/commits/60d59abc7cdf0b3654604f24b2024dca7bf2043c)

numba/core/typing/npydecl.py
```diff
@@ -737,12 +737,14 @@ class NdExpandDims(CallableTemplate):
...
-                return
-
-            layout = a.layout if a.ndim <= 1 else 'A'
-            return a.copy(ndim=a.ndim + 1, layout=layout)
+            if isinstance(a, types.Array):
+                layout = a.layout if a.ndim <= 1 else 'A'
+                if isinstance(axis, types.Integer):
+                    return a.copy(ndim=a.ndim + 1, layout=layout)
+                elif isinstance(axis, types.BaseTuple):
+                    return a.copy(ndim=a.ndim + axis.count, layout=layout)
```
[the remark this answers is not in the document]
kc611: Resolved in [60d59ab](https://github.com/numba/numba/pull/7604/commits/60d59abc7cdf0b3654604f24b2024dca7bf2043c)

kc611:
Depends on/currently built on top of: https://github.com/numba/numba/pull/8234

numba/np/arrayobj.py
```diff
@@ -5168,80 +5168,72 @@ def _normalize_axis(context, builder, func_name, ndim, axis):
...
     np.expand_dims() with the given axis.
     """
     retty = sig.return_type
     ndim = retty.ndim
     arrty = sig.args[0]
+    _empty_tuple = tuple([-1] * ndim)
+
+    if isinstance(sig.args[1], types.Integer):
+        to_tuple = register_jitable(lambda axis: (axis,))
+    else:
```
esc, line 5182: The documentation on https://numpy.org/doc/stable/reference/generated/numpy.expand_dims.html seems to suggest `axis` can either be an int or a tuple of ints. In this case however it could also be an array or a string?

numba/np/arrayobj.py
```diff
@@ -5168,80 +5168,72 @@ def _normalize_axis(context, builder, func_name, ndim, axis):
...
+        to_tuple = register_jitable(lambda axis: axis)
+
+    @register_jitable
+    def normalize_axis(axis, ndim):
+        for i in range(len(axis)):
+            if axis[i] < 0:
+                axis = tuple_setitem(axis, i, axis[i] + ndim)
+
+            if axis[i] < 0 or axis[i] >= ndim:
```
esc, line 5191: This is a bit hard to read, it's not clear from the code how `axis[i]` is still less than zero after the statement above. Suggest to either add a comment or refactor to something more obvious.

numba/np/arrayobj.py
```diff
@@ -5168,80 +5168,72 @@ def _normalize_axis(context, builder, func_name, ndim, axis):
...
+                raise IndexError("axis out of bounds")
+        return axis
+
+    def _insert_value_in_tuple(_curr_tuple, ndim, axis, value):
```
esc, line 5195: Perhaps `insert` isn't the best name here. It creates a new tuple, sets the values in that tuple and returns it.

esc (changes requested):

esc:
@kc611 I've given this an initial pass and left a few questions for you to answer.

numba/np/arrayobj.py
kc611: Adressed in [0ca6eb6](https://github.com/numba/numba/pull/7604/commits/0ca6eb6c4ee2c25ba609e7f0bc722a2527303de6)

numba/np/arrayobj.py
kc611: Addressed in [0ca6eb6](https://github.com/numba/numba/pull/7604/commits/0ca6eb6c4ee2c25ba609e7f0bc722a2527303de6)

numba/np/arrayobj.py
kc611: The validity of arguments is being checked at type level:

https://github.com/numba/numba/pull/7604/files#diff-d0aefd08783016c4e2a19e29d4725c7870806ea2aa1ab4caa6e8f0adc08c6e64R5287-R5293

numba/tests/test_array_manipulation.py
```diff
@@ -439,6 +448,13 @@ def test_expand_dims_exceptions(self):
             cfunc(arr, 'hello')
         self.assertIn('Argument "axis" must be an integer', str(raises.exception))
```
esc, line 449:
```suggestion
        self.assertIn('Argument "axis" must be an integer or tuple of integers', str(raises.exception))
```

Suggestion the above change the error message test to be somewhat more specific.

numba/np/arrayobj.py
```diff
@@ -5257,15 +5253,26 @@ def expand_dims(context, builder, sig, args, axis):
...
     sig = ret(a, axis)
 
     def codegen(context, builder, sig, args):
-        axis = context.cast(builder, args[1], sig.args[1], types.intp)
-        axis = _normalize_axis(context, builder, "np.expand_dims",
-                               sig.return_type.ndim, axis)
-
-        ret = expand_dims(context, builder, sig, args, axis)
+        axes_types = sig.args[1]
+        if (
```
esc, line 5268: The formatting here seems a little off?

esc (changes requested):
I have a few more things to fix. Also, 

I was wondering about the comment on:

Screen Shot 2022-11-15 at 11 55 49

https://numpy.org/doc/stable/reference/generated/numpy.expand_dims.html

So I tried the following:

```
In [25]: @nb.jit
    ...: def foo(a, axis):
    ...:     return np.expand_dims(a, axis)
    ...:

In [26]: a = np.ones((2,2))

In [27]: b = foo(a, -4)
IndexError                                Traceback (most recent call last)
Input In [27], in <cell line: 1>()
----> 1 b = foo(a, -4)

IndexError: tuple index out of range
```

And my question is, should this not raise:

```
IndexError: axis out of bounds
```

Rather than the error it raises right now?

numba/tests/test_array_manipulation.py
kc611: Resolved in [4e4c820](https://github.com/numba/numba/pull/7604/commits/4e4c8200aae40b629579780fdcd1aa8b64096536)

numba/np/arrayobj.py
kc611: Resolved in [4e4c820](https://github.com/numba/numba/pull/7604/commits/4e4c8200aae40b629579780fdcd1aa8b64096536)

kc611:
The above mentioned issue has also been resolved in [4e4c820](https://github.com/numba/numba/pull/7604/commits/4e4c8200aae40b629579780fdcd1aa8b64096536)

esc:
As release manager, I have decided that this isn't mature enough to be ready for the upcoming 0.58 release (next week) and as such I am moving this into the next release.

numba/np/arrayobj.py
```diff
@@ -5986,99 +5986,76 @@ def impl(object, dtype=None):
...
-    shapes = cgutils.unpack_tuple(builder, arr.shape)
-    strides = cgutils.unpack_tuple(builder, arr.strides)
+    ndim_const = context.get_constant(types.intp, ndim)
+    zero = context.get_constant(types.intp, 0)
+    one = context.get_constant(types.intp, 1)
 
-    new_shapes = _insert_axis_in_shape(context, builder, shapes, ndim, axis)
-    new_strides = _insert_axis_in_strides(context, builder, strides, ndim, axis)
+    fnty = context.typing_context.resolve_value_type(_build_shape_stride_tuples)
+    sig = fnty.get_call_type(
```
esc:
```suggestion
    fsig = fnty.get_call_type(
```

So as to keep the `sig` which was an argument and the new signature distinct.

numba/np/arrayobj.py
```diff
@@ -5986,99 +5986,76 @@ def impl(object, dtype=None):
...
+                raise ValueError("An axis is being repeated")
+            res = tuple_setitem(res, i, value)
+
+        # 3. Incrementally put original tuple values into
+        # indices that are -1
+        idx = 0
+        for i in range(ndim):
+            if res[i] == -1:
+                res = tuple_setitem(res, i, _curr_tuple[idx])
+                idx = idx + 1
```
esc:
```suggestion
                idx += 1
```

For simplicity.

esc (changes requested):

PR #7605, opened 2021-11-29
Fix TBB 2021 DSO names on OSX/Win and make TBB reporting consistent

### LINKED ISSUE #7600: TBB backend not working on Win due to wrong dll name

Hi, 

It seems that compiling with `NUMBA_THREADING_LAYER=tbb` doesn't work on windows at all. The attempt to load the dll is done in `_check_tbb_version_compatible` but it fails since on Windows it's named: "tbb12.dll" now:
> akozlov@akozlov-MOBL1: ~/AppData/Local/Continuum/anaconda3/envs/test_numba_tbb_backend$ find . -name "*tbb*" | grep dll
> ./Library/bin/mkl_tbb_thread.1.dll
> ./Library/bin/tbb12.dll
> ./Library/bin/tbbmalloc.dll
> ./Library/bin/tbbmalloc_proxy.dll

Below are the steps to reproduce the issue:
```
conda create -n test_numba_tbb_backend -y -c numba -c conda-forge python=3.7 numpy=1.20 llvmlite>=0.38 vc vs2015_runtime vs2017_win-64
conda activate test_numba_tbb_backend
conda install -c conda-forge --override-channels tbb=2021.4 tbb-devel=2021.4
set NUMBA_THREADING_LAYER=tbb
python test_tbb.py
```
where test_tbb.py is:
```
import numpy as np

@njit(parallel=True)
def test_impl(arr):
    return arr.sum()

data = np.arange(1000)
print(test_impl(data))
```

Now it fails with:
> numba.core.errors.LoweringError: Failed in nopython mode pipeline (step: native lowering)
> No threading layer could be loaded.
> HINT:
> Intel TBB is required, try:
> $ conda/pip install tbb
> 
> File "numba\parfors\parfor.py", line 256:
>         def sum_1(in_arr):
>             <source elided>
>             val = zero
>             for i in numba.parfors.parfor.internal_prange(len(in_arr)):
>             ^
> 
> During: lowering "id=0[LoopNest(index_variable = parfor_index.33, range = (0, arr_size0.32, 1))]{42: <ir.Block at C:\Users\akozlov\AppData\Local\Continuum\anaconda3\numba_master\numba\numba\parfors\parfor.py (256)>}Var(parfor_index.33, parfor.py:256)" at C:\Users\akozlov\AppData\Local\Continuum\anaconda3\numba_master\numba\numba\parfors\parfor.py (256)

The following patch should fix this (also changes lib name on MacOS, as tbb package on conda-forge suggests this name was also incorrect):
```diff
diff --git a/numba/np/ufunc/parallel.py b/numba/np/ufunc/parallel.py
index 5d85bca9f..9c744fa83 100644
--- a/numba/np/ufunc/parallel.py
+++ b/numba/np/ufunc/parallel.py
@@ -344,9 +344,9 @@ def _check_tbb_version_compatible():
     try:
         # first check that the TBB version is new enough
         if _IS_WINDOWS:
-            libtbb_name = 'tbb'
+            libtbb_name = 'tbb12.dll'
         elif _IS_OSX:
-            libtbb_name = 'libtbb.dylib'
+            libtbb_name = 'libtbb.12.dylib'
         elif _IS_LINUX:
             libtbb_name = 'libtbb.so.12'
         else:
```

Best Regards,
Alexey.

stuartarchibald:
As title.

Fixes #7600

### DIFF
--- a/numba/np/ufunc/parallel.py
+++ b/numba/np/ufunc/parallel.py
@@ -344,9 +344,9 @@ def _check_tbb_version_compatible():
     try:
         # first check that the TBB version is new enough
         if _IS_WINDOWS:
-            libtbb_name = 'tbb'
+            libtbb_name = 'tbb12.dll'
         elif _IS_OSX:
-            libtbb_name = 'libtbb.dylib'
+            libtbb_name = 'libtbb.12.dylib'
         elif _IS_LINUX:
             libtbb_name = 'libtbb.so.12'
         else:
--- a/numba/misc/numba_sysinfo.py
+++ b/numba/misc/numba_sysinfo.py
@@ -421,7 +421,13 @@ def get_sysinfo():
         return "Unknown import problem."
 
     try:
+        # check import is ok, this means the DSO linkage is working
         from numba.np.ufunc import tbbpool  # NOQA
+        # check that the version is compatible, this is a check performed at
+        # runtime (well, compile time), it will also ImportError if there's
+        # a problem.
+        from numba.np.ufunc.parallel import _check_tbb_version_compatible
+        _check_tbb_version_compatible()
         sys_info[_tbb_thread] = True
     except ImportError as e:
         # might be a missing symbol due to e.g. tbb libraries missing


numba/np/ufunc/parallel.py
```diff
@@ -344,9 +344,9 @@ def _check_tbb_version_compatible():
     try:
         # first check that the TBB version is new enough
         if _IS_WINDOWS:
-            libtbb_name = 'tbb'
+            libtbb_name = 'tbb12.dll'
```
gmarkall, line 347: What happens if there is an older TBB version? Was there a previous version of TBB where the library was named `tbb.dll` that would be missed now?

gmarkall: ... or was this just always wrong?

stuartarchibald: Am trying to enforce some consistency in https://github.com/numba/numba/pull/7606, whereby the CI system checks that if a threading layer is requested that it can actually be loaded.

numba/np/ufunc/parallel.py
```diff
@@ -344,9 +344,9 @@ def _check_tbb_version_compatible():
...
         elif _IS_OSX:
-            libtbb_name = 'libtbb.dylib'
+            libtbb_name = 'libtbb.12.dylib'
```
sklam, line 349: confirmed this on OSX

sklam (approved):
LGTM

stuartarchibald:
Thanks for the review @sklam.

PR #7619, opened 2021-12-03
CUDA: Fix linking with PTX when compiling lazily

gmarkall:
When no signature was provided, the `link` kwarg was accidentally dropped in the wrapper function calling `jit` inside the decorator.

Also includes the removal of some un-needed decorators in the linker test.

Fixes #7607.

Issue #7528, opened 2021-11-02
Implement array reshape for CUDA


gmarkall:
The following:

```python
from numba import cuda
import numpy as np


@cuda.jit('void()')
def f():
    x = cuda.shared.array(0, dtype=np.int32)
    x.reshape(2, 2)
```

Gives:

```
Traceback (most recent call last):
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudadrv/driver.py", line 2222, in complete
    driver.cuLinkComplete(self.handle, byref(cubin), byref(size))
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudadrv/driver.py", line 303, in safe_cuda_api_call
    self._check_error(fname, retcode)
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudadrv/driver.py", line 343, in _check_error
    raise CudaAPIError(retcode, msg)
numba.cuda.cudadrv.driver.CudaAPIError: [300] Call to cuLinkComplete results in CUDA_ERROR_INVALID_SOURCE

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/gmarkall/numbadev/issues/gabri94/repro.py", line 6, in <module>
    def f():
  File "/home/gmarkall/numbadev/numba/numba/cuda/decorators.py", line 109, in kernel_jit
    return Dispatcher(func, [func_or_sig], targetoptions=targetoptions)
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 938, in __init__
    self.compile(sigs[0])
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 1117, in compile
    kernel.bind()
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 628, in bind
    self._codelibrary.get_cufunc()
  File "/home/gmarkall/numbadev/numba/numba/cuda/codegen.py", line 201, in get_cufunc
    cubin = self.get_cubin(cc=device.compute_capability)
  File "/home/gmarkall/numbadev/numba/numba/cuda/codegen.py", line 178, in get_cubin
    cubin_buf, size = linker.complete()
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudadrv/driver.py", line 2224, in complete
    raise LinkerError("%s\n%s" % (e, self.error_log))
[...]

PR #7621, opened 2021-12-03
Add support for linking CUDA C / C++ with `@cuda.jit` kernels

gmarkall:
This adds support for linking CUDA C/C++ sources using NVRTC. One links them in the same way as PTX files, i.e.:

```python
@cuda.jit(link=['source.cu'])
def f(x):
    # bar() is defined in source.cu
    bar(x)
```

Some design decisions I've made for this initial implementation:

- This uses the NVIDIA binding to access NVRTC, so that issues around locating NVRTC are delegated to it, rather than adding logic to Numba that duplicates this functionality (and is potentially difficult and error-prone to implement and test).
- There is no support for the ctypes API. Ideally I'd avoid adding it, with the current plan being to eventually default to the NVIDIA bindings and remove the ctypes bindings in the long run.
- Whilst not all CUDA sources require the CUDA include directory, some do. Towards providing this, Numba assumes a default location for the include path, and allows it to be set if necessary though `NUMBA_CUDA_INCLUDE_PATH`. Some more sophisticated logic could be added to detect the include path in future if this seems necessary / useful to do and a standard / sensible way to do it can be found. (An example of requiring the CUDA includes is in numba/numba-examples#40, where cuRAND includes are used).
- No support for adding additional include paths is present for this initial iteration - sources must only rely on CUDA includes for now. Support for specifying include paths may be added in future.

Based on #7619, which fixes issues with linking when lazily compiling (though it is not strictly necessary for this PR, it makes sense to consider on top of it).


gmarkall:
gpuci run tests

seibert:
Will this provide a solution for accessing cuRandom?  Would love to (slowly) deprecate our existing CUDA random number generator.

gmarkall:
@seibert It makes it easier to call cuRAND (it was always possible, but you'd have to compile to PTX first). Here's an example: https://github.com/numba/numba-examples/pull/40 - I think we could work through the cuRAND API and provide a nice wrapper for it using this functionality, it's likely just a case of finding the hours to put into it.

(is cuRandom something different? My searches for it turn up mostly `pycuda.curandom`)

To compile code using cuRAND, I also had to make a small hack on this PR, to include the CUDA include directory, e.g.:

```diff
diff --git a/numba/cuda/cudadrv/driver.py b/numba/cuda/cudadrv/driver.py
index 81e7d608a..a78bcc4f1 100644
--- a/numba/cuda/cudadrv/driver.py
+++ b/numba/cuda/cudadrv/driver.py
@@ -2810,9 +2810,10 @@ class CudaPythonLinker(Linker):
             major, minor = dev.compute_capability
 
         arch = bytes(f'--gpu-architecture=compute_{major}{minor}', 'ascii')
+        include = b'-I/usr/local/cuda-11.4/include'
 
         # rdc is needed to prevent device functions being optimized away
-        opts = [arch, b'-rdc', b'true']
+        opts = [arch, b'-rdc', b'true', include]
         err, = nvrtc.nvrtcCompileProgram(prog, len(opts), opts)
         compile_error = (err == nvrtc.nvrtcResult.NVRTC_ERROR_COMPILATION)
 
```

so I probably need to handle finding an appropriate CUDA include dir in this PR too.

seibert:
oops, I got the name wrong.  Yes, I meant cuRAND.  :)

gmarkall:
This will also help enable https://github.com/numba/numba/issues/7528

gmarkall:
Note to anyone who's pulled locally or recently observed changes: I did a force push to squash out some commits where I had a file called `aux.cu`, which will cause pain for Windows users if they ever try to checkout those commits. The history diverges just after commit 7fe9422 (Add NvrtcError to simulator).

gmarkall:
(gpuci fail seems to be due to bad Conda downloads)

gpuci run tests

gmarkall:
Although gpuCI seems to be having a bad day, I'm going to mark this as ready for review because it's certainly in a state to receive feedback. When I get back to work on Monday I'll rerun the build (and hopefully all will be well since the last gpuCI pass I only added documentation and Windows-specific path handling).

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

docs/source/cuda/cuda_ffi.rst
```diff
@@ -0,0 +1,148 @@
...
+-----------------------------
+
+The ``link`` keyword argument of the :func:`@cuda.jit <numba.cuda.jit>`
+decorator accepts a list of files; files whose name ends in ``.cu`` will be
+compiled with the `NVIDIA Runtime Compiler (NVRTC)
+<https://docs.nvidia.com/cuda/nvrtc/index.html>`_ and linked into the kernel as
+PTX; other files will be passed directly to the CUDA Linker.
+
+For example, the following kernel calls the ``mul()`` function declared above
+with the implementation ``mul_f32_f32()`` in a file called ``functions.cu``:
```
stuartarchibald, line 94: Does something need to be said about:
* The required presence of `nvrtc` on the path (and where to get it).
* How specified files are resolved via paths, am guessing working directory only or full path required?

docs/source/cuda/cuda_ffi.rst
```diff
@@ -0,0 +1,148 @@
...
+
+Support for compiling and linking of CUDA C/C++ code is provided through the use
+of NVRTC subject to the following considerations:
+
+- It is only available when using the NVIDIA Bindings. See
+  :envvar:`NUMBA_CUDA_USE_NVIDIA_BINDING`.
+- The CUDA include path is assumed by default to be ``/usr/local/cuda/include``
+  on Linux and ``$env:CUDA_PATH\include`` on Windows. It can be modified using
+  the environment variable :envvar:`NUMBA_CUDA_INCLUDE_PATH`.
+- The CUDA include folder will be made available to NVRTC on the include path;
```
stuartarchibald, line 117:
```suggestion
- The CUDA include directory will be made available to NVRTC on the include path;
```

numba/core/config.py
```diff
@@ -412,6 +412,20 @@ def avx_default():
...
 
+        # Location of the CUDA include files
+        if IS_WIN32:
+            cuda_path = os.environ.get('CUDA_PATH')
+            if cuda_path:
+                default_cuda_include_path = os.path.join(cuda_path, "include")
+            else:
+                default_cuda_include_path = "cuda_include_not_found"
+        else:
+            default_cuda_include_path = "/usr/local/cuda/include"
```
stuartarchibald, line 423: Should this be `os.path.join(['usr', 'local', 'cuda', 'include'])`?

numba/cuda/cudadrv/driver.py
```diff
@@ -2605,15 +2609,27 @@ def error_log(self):
     def add_ptx(self, ptx, name):
         """Add PTX source in a string to the link"""
 
+    @abstractmethod
+    def add_cu(self, cu, name):
```
stuartarchibald, line 2613: Is the `name` here e.g. the file name from which the `cu` string was obtained?

numba/cuda/cudadrv/driver.py
```diff
@@ -2715,6 +2735,88 @@ def complete(self):
...
+        # Get log from compilation
+        err, log_size = nvrtc.nvrtcGetProgramLogSize(self._program)
+        self.check(err)
+        log = b' ' * log_size
+        err, = nvrtc.nvrtcGetProgramLog(self._program, log)
+        self.check(err)
+
+        # If the compile failed, provide the log in an exception
+        if compile_error:
+            msg = (f'NVRTC Compilation failure compiling {name}:\n\n'
```
stuartarchibald, line 2782:
```suggestion
            msg = (f'NVRTC Compilation failure whilst compiling {name}:\n\n'
```

numba/cuda/cudadrv/driver.py
```diff
@@ -2715,6 +2735,88 @@ def complete(self):
...
+                   f'{log.decode()}')
+            raise NvrtcError(msg)
+
+        # Otherwise, if there's any content in the log, present it as a warning
+        if log_size > 1:
+            msg = f"NVRTC log messages compiling {name}:\n\n{log.decode()}"
```
stuartarchibald, line 2788:
```suggestion
            msg = f"NVRTC log messages whilst compiling {name}:\n\n{log.decode()}"
```

numba/cuda/cudadrv/driver.py
```diff
@@ -2715,6 +2735,88 @@ def complete(self):
...
+            warnings.warn(msg)
```
stuartarchibald, line 2789: Should this be a Numba CUDA specific warning class?

numba/cuda/decorators.py
```diff
@@ -142,6 +142,15 @@ def autojitwrapper(func):
...
 
 def declare_device(name, sig):
+    """
+    Declare the signature of a foreign function. Returns a descriptor that can
+    be used to call the function from a Python kernel.
+
+    :param name: The name of the foreign function.
+    :type name: str
+    :param sig: The Numba signature of the function.
+    :type sig: str
```
stuartarchibald, line 152: Should this also accept a `numba.core.typing.templates.Signature` such as that derived from composing a signature with `numba.types` types?

numba/cuda/tests/cudadrv/data/cuda_include.cu
```diff
@@ -0,0 +1,5 @@
+// Not all CUDA includes are safe to include in device code compiled by NVRTC,
```
stuartarchibald, line 1: I think these `.cu` files will need adding to `package_data` in `setup.py`.

numba/cuda/tests/doc_examples/ffi/functions.cu
```diff
@@ -0,0 +1,15 @@
+// Foreign function example: multiplication of a pair of floats
```
stuartarchibald, line 1: I think this `.cu` file will need adding to `package_data` in `setup.py`.

numba/cuda/tests/cudadrv/test_linker.py
```diff
@@ -95,6 +97,76 @@ def test_linking_lazy_compile(self):
     def test_linking_eager_compile(self):
         self._test_linking(eager=True)
 
+    @skip_unless_cuda_python('NVIDIA Binding needed for NVRTC')
+    def test_linking_cu(self):
+        bar = cuda.declare_device('bar', 'int32(int32)')
+
+        link = os.path.join(os.path.dirname(__file__), 'data', 'jitlink.cu')
```
stuartarchibald, line 104: Is `jitlink.cu` missing from this patch?

numba/cuda/tests/cudadrv/test_linker.py
```diff
@@ -95,6 +97,76 @@ def test_linking_lazy_compile(self):
...
+        np.testing.assert_array_equal(r, expected)
+
+    @skip_unless_cuda_python('NVIDIA Binding needed for NVRTC')
+    def test_linking_cu_log_warning(self):
+        bar = cuda.declare_device('bar', 'int32(int32)')
+
+        link = os.path.join(os.path.dirname(__file__), 'data', 'warn.cu')
+
+        with warnings.catch_warnings(record=True) as w:
+            @cuda.jit('void(int32)', link=[link])
```
stuartarchibald, line 129: context managed block should probably have a call to this to filter known internal warnings:
https://github.com/numba/numba/blob/59fa2e335be68148b3bd72a29de3ff011430038d/numba/tests/support.py#L141-L144

numba/cuda/tests/doc_examples/test_ffi.py
```diff
@@ -0,0 +1,50 @@
...
+        # magictoken.ex_linking_cu.begin
+        from numba import cuda
+        import numpy as np
+        import os
+
+        # Declaration of the foreign function
+        mul = cuda.declare_device('mul_f32_f32', 'float32(float32, float32)')
+
+        # Path to the source containing the foreign function
+        # (here assumed to be in a subfolder called "ffi"
```
stuartarchibald, line 22:
```suggestion
        # (here assumed to be in a sub-directory called "ffi")
```

stuartarchibald (commented):
Thanks for implementing this functionality @gmarkall. On inspection, the code in general looks good, I've left a few minor comments/queries inline. Will manually tests this next to try it out! Thanks again!

numba/core/config.py
gmarkall: Yes, that would be better.

numba/cuda/cudadrv/driver.py
gmarkall: It could be, but it's not required.

numba/cuda/cudadrv/driver.py
gmarkall: None of the other warnings in driver.py use a specific class. I think there's not too great a need for this - I'd imagine you always want to resolve the warnings rather than filter them, but I'm open to a suggestion as to what I might be missing / misjudging here.

numba/cuda/decorators.py
gmarkall: It will already accept one, I think I just got carried away with the docs here. I'll just remove this type documentation since a Numba signature can be a `Signature` or a string.

numba/cuda/tests/cudadrv/test_linker.py
gmarkall: It already exists: https://github.com/numba/numba/blob/main/numba/cuda/tests/cudadrv/data/jitlink.cu

numba/cuda/tests/cudadrv/test_linker.py
gmarkall: I've now added the call.

numba/cuda/tests/doc_examples/ffi/functions.cu
gmarkall: Added.

numba/cuda/tests/cudadrv/data/cuda_include.cu
gmarkall: Added.

docs/source/cuda/cuda_ffi.rst
gmarkall: I've now added a note about the relative / absolute path specification for linked files, and also about the need for NVRTC. I didn't get too specific about NVRTC because I felt it would get needlessly complex and confusing trying to be specific about the possibilities across platforms and versions.

gmarkall:
gpuci run tests

gmarkall:
@stuartarchibald Many thanks for the review - I have addressed / responded to all comments, so I think this is ready for another look.

docs/source/cuda/cuda_ffi.rst
stuartarchibald: Thanks for adding the path specification. RE the `nvrtc` binary location, this seems reasonable, it's quite an advanced use case and it needing to be on the path would likely be an implicit requirement, how this is resolved appropriately is understandably user specific.

numba/cuda/cudadrv/driver.py
stuartarchibald: ok, thanks, I think for consistency with the current code base it's fine to leave this as is in this PR, but might be something to consider at a later time.

numba/cuda/decorators.py
stuartarchibald: Thanks, seems like that's the easiest fix!

numba/cuda/tests/cudadrv/test_linker.py
stuartarchibald: ah yes, thanks.

numba/cuda/cudadrv/driver.py
stuartarchibald: Should the docstring explain what it is? If it's not required should there be a default?

gmarkall: Sorry, I mean it's not required to be the filename from which the cu string was obtained. However, it should be provided and should end in .cu - I've updated the docstring.

gmarkall:
gpuci run tests

gmarkall:
@stuartarchibald All should be ready for another look now.

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

numba/cuda/cudadrv/driver.py
gmarkall: Since I'm now using splitext (as you suggested OOB) it doesn't really matter what the extension is, so I've removed that requirement from the docstring.

gmarkall:
@stuartarchibald I think this is ready again for re-reviewing the responses to the last couple of comments.

stuartarchibald:
Notes from manual testing:
1. Checking mis-spelling of file name and extension behave reasonably but error handling could be improved.
2. Checking mis-matched function prototypes breaks in an expected way.
3. Checking two `.cu` files defining the same named function is spotted as a multiple definition.
4. CUDA does unsafe conversion of e.g. float64 to float32 if the `.cu` file definition takes a `float32`.

stuartarchibald (commented):
Thanks for the updates @gmarkall, couple of minor other things else looks good. Thanks again!

numba/cuda/cudadrv/driver.py
```diff
@@ -2605,15 +2609,28 @@ def error_log(self):
...
+    def add_cu(self, cu, name):
+        """Add CUDA source in a string to the link. The name of the source
+        file should be specified in `name`."""
+
     @abstractmethod
     def add_file(self, path, kind):
         """Add code from a file to the link"""
 
+    def add_cu_file(self, path):
+        with open(path, 'rb') as f:
```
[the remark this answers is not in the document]
gmarkall: I think it might as well just try, rather than reimplementing the error-handling logic of `open()`. For example, if you try to link a nonexistent file, the last line of the traceback is pretty clear:

```
FileNotFoundError: [Errno 2] No such file or directory: '/home/gmarkall/numbadev/numba/numba/cuda/tests/doc_examples/ffi/functioddns.cu'
```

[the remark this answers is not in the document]
stuartarchibald: ok, let's leave it as-is. It does state the issue, which is probably sufficient to start a bug report for e.g. the maintainers of a library that is using this functionality.

gmarkall:
gpuci run tests

numba/cuda/cudadrv/driver.py
```diff
@@ -2605,15 +2609,28 @@ def error_log(self):
...
+            cu = f.read()
+        self.add_cu(cu, os.path.basename(path))
+
     def add_file_guess_ext(self, path):
         """Add a file to the link, guessing its type from its extension."""
         ext = path.rsplit('.', 1)[1]
```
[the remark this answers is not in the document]
gmarkall: Using it now.

numba/cuda/cudadrv/driver.py
```diff
@@ -2605,15 +2609,28 @@ def error_log(self):
...
-        kind = FILE_EXTENSION_MAP[ext]
-        self.add_file(path, kind)
+        if ext == 'cu':
+            self.add_cu_file(path)
+        else:
+            kind = FILE_EXTENSION_MAP[ext]
```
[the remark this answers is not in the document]
gmarkall: I've changed it anyway, it now raises a `RuntimeError` describing the problem (depending on whether the extension is unknown, or non-existent) and there's a couple of tests for this.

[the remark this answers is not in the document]
stuartarchibald: Thanks!

stuartarchibald (dismissed):

stuartarchibald:
> @stuartarchibald I've made updates responding to all the feedback - assuming gpuCI and the buildfarm pass, I'd like to give this another manual check on Windows too, because I've changed the code since I last checked on Windows with the CUDA Python bindings.

@gmarkall No problem, thanks for working through all of the review comments. I'll mark this for a build farm run now as GPU CI has passed, will also wait for your feedback on how a manual test on Windows with the CUDA Python bindings goes. Thanks again!

numba/cuda/cudadrv/driver.py
```diff
@@ -2700,6 +2722,10 @@ def add_file(self, path, kind):
                 msg = "%s\n%s" % (e, self.error_log)
             raise LinkerError(msg)
 
+    def add_cu(self, path, name):
+        raise NotImplementedError("Linking CUDA source files is not supported "
+                                  "with the ctypes binding. ")
```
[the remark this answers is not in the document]
gmarkall: I've added a test for this now. The linking cu tests that actually try to link a cu file are skipped if the ctypes binding is in use.

gmarkall:
gpuci run tests

numba/cuda/tests/cudadrv/test_linker.py
```diff
@@ -158,6 +159,14 @@ def kernel(x):
         # Check the filename is reported correctly
         self.assertIn('in the compilation of "error.cu"', msg)
 
+    @skip_with_cuda_python
+    def test_linking_cu_ctypes_unsupported(self):
+        msg = 'not supported with the ctypes binding'
```
stuartarchibald, line 164: Perhaps:
```suggestion
        msg = ('Linking CUDA source files is not supported with the ctypes '
               'binding')
```
just to make absolutely sure it's that specific exception from linking and not a different but similar one?

stuartarchibald (approved):
Thanks for adding the additional test.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_125`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_125`.

Passed.

Issue #7611, opened 2021-12-02
njit with parallel=True causes fatal Python error and aborted


dlee992:
# Reporting a bug

```
import time
import unittest
import pandas as pd
import numpy as np
from numba import njit, prange, typed

num_columns, num_rows, num_borders = 50, 100_000, 20
target_col = 'col0'

features = [f'col{i}' for i in range(num_columns)]
input_df = pd.DataFrame(np.random.rand(num_rows, num_columns), columns=features)
borders = tuple([x/100 for x in range(0, 100, 3)][:num_borders])


# ndarray_2d: N * M, ndarray_1d: N
def _compute(ndarray_2d, ndarray_1d, numba_list):
    for i in prange(len(borders)):
        filtered = (ndarray_2d >= borders[i]).astype(np.int_) # filtered: N * M
        filtered_1d = filtered.sum(axis=0).astype(np.float_) # filtered_1d: M
        another_filtered_1d = (filtered.transpose() * ndarray_1d).sum(axis=1) # another_filtered_1d: M
        for item in zip(filtered_1d, another_filtered_1d):
            numba_list.extend(item)


class MyTestCase(unittest.TestCase):
    @staticmethod
    def test_something():
        numba_list = typed.List()
        numba_list.append(0.1)
        ndarray_2d = input_df.to_numpy()
        ndarray_1d = input_df[target_col].to_numpy()

        njit_parallel_func = njit(parallel=True)(_compute)
        # njit_parallel_func(ndarray_2d, ndarray_1d, numba_list)

        python_func = _compute
        numba_list = []
        start_time = time.time()
        python_func(ndarray_2d, ndarray_1d, numba_list)
        print(f'python time = {time.time() - start_time}s')

        njit_func = njit(_compute)
        for _ in range(5):
            numba_list = typed.List()
            numba_list.append(0.1)
            middle_time = time.time()
            njit_func(ndarray_2d, ndarray_1d, numba_list)
            print(f'njit time = {time.time() - middle_time}s')


if __name__ == '__main__':
    unittest.main()

```

In this test code:

1. if uncomment `njit_parallel_func(ndarray_2d, ndarray_1d, numba_list)`, python process will be killed by sth bad happend in numba internal.
```
test_parallel.py::MyTestCase::test_something Fatal Python error: Aborted

Thread 0x000000011094f600 (most recent call first):
  File "/Users/.../debug-numba/test_parallel.py", line 36 in test_something
[...]

PR #7622, opened 2021-12-04
Support fortran loop ordering for ufunc generation

sklam:
Numba currently generates C ordered loop for ufunc only. This adds the support for F ordered loop generation when arrayexpr inputs are mostly F ordered.

Consider the case:

```python
import numpy as np
from numba import njit


# Make horribly strided arrays
num_columns = 50
num_rows = 100_000
inp_data_2d = np.random.rand(num_rows, num_columns)
inp_data_1d = (inp_data_2d)[:,0]
print(inp_data_2d.strides)
print(inp_data_1d.strides)


def compute(data2d, data1d):
    out = (data2d.transpose() * data1d)
    return out


@njit
def jit_compute(data2d, data1d):
    out = (data2d.transpose() * data1d)
    return out

# warm up
expect = compute(inp_data_2d, inp_data_1d)
got = jit_compute(inp_data_2d, inp_data_1d)
np.testing.assert_array_equal(expect, got)

%timeit -n 10 compute(inp_data_2d, inp_data_1d)
%timeit -n 10 jit_compute(inp_data_2d, inp_data_1d)
```

Without this patch, Numba is 10x slower than NumPy.

Relates to issue in https://github.com/numba/numba/issues/7611

### DIFF
--- a/numba/np/npyimpl.py
+++ b/numba/np/npyimpl.py
@@ -383,7 +383,17 @@ def numpy_ufunc_kernel(context, builder, sig, args, ufunc, kernel_class):
     # assume outputs are all the same size, which numpy requires
 
     loopshape = outputs[0].shape
-    with cgutils.loop_nest(builder, loopshape, intp=intpty) as loop_indices:
+
+    input_layouts = [inp.layout for inp in inputs if isinstance(inp, _ArrayHelper)]
+    num_c_layout = len([x for x in input_layouts if x == 'C'])
+    num_f_layout = len([x for x in input_layouts if x == 'F'])
+
+    if num_f_layout > num_c_layout:
+        order = 'F'
+    else:
+        order = 'C'
+
+    with cgutils.loop_nest(builder, loopshape, intp=intpty, order=order) as loop_indices:
         vals_in = []
         for i, (index, arg) in enumerate(zip(indices, inputs)):
             index.update_indices(loop_indices, i)
--- a/numba/np/npyimpl.py
+++ b/numba/np/npyimpl.py
@@ -384,10 +384,16 @@ def numpy_ufunc_kernel(context, builder, sig, args, ufunc, kernel_class):
 
     loopshape = outputs[0].shape
 
-    input_layouts = [inp.layout for inp in inputs if isinstance(inp, _ArrayHelper)]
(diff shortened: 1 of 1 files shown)


numba/np/npyimpl.py
```diff
@@ -383,7 +383,17 @@ def numpy_ufunc_kernel(context, builder, sig, args, ufunc, kernel_class):
     # assume outputs are all the same size, which numpy requires
 
     loopshape = outputs[0].shape
-    with cgutils.loop_nest(builder, loopshape, intp=intpty) as loop_indices:
+
+    input_layouts = [inp.layout for inp in inputs if isinstance(inp, _ArrayHelper)]
+    num_c_layout = len([x for x in input_layouts if x == 'C'])
+    num_f_layout = len([x for x in input_layouts if x == 'F'])
```
sklam, line 389: TODO check what numpy does

sklam: There are some information around https://github.com/numpy/numpy/blob/ff4b33edcf6819f24f9a1b4a483d64873cea2d5c/numpy/core/src/common/lowlevel_strided_loops.h#L621.

Numpy will perform a trivial loop iteration if arguments are all C or F. Otherwise, it will use the array iterator which we don't have. 

The logic here is doing the best effort. If more arrays are in F layout then in C layout, we will use that F iteration order; otherwise, we will use C order by default.

stuartarchibald: Perhaps not for this PR, but maybe sometime... given `ndim` is known for each array, if there are 0D arrays should they get special treatment? e.g. imagine `input_layouts` evaluated as `['F', 'C', 'C', 'C', 'C']`, but the `'C'` arrays were all `0D` and the `'F'` array was `>1D`, it'd be most efficient to iterate in `'F'` order?

sklam: You are right. The NumPy code has logic for array scalars.

stuartarchibald (approved):
Thanks for the patch. Manual performance testing suggests this fixes the issue described above.

numba/np/npyimpl.py
stuartarchibald: #8618 tracks.

Issue #5890, opened 2020-06-19
Seemingly random segfault on macOS if function is in larger library


max3-2:
# Reporting a bug
- [x] I am using the latest released version of Numba (most recent is visible in
 the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [ x] I have included below a minimal working reproducer (if you are unsure how
 to write one see http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports).

Hi,

first of all, sorry for this small report and few examples, but at this point this seems to untraceable to me that im hoping for any input to trace down the issue. Maybe its even a severely stupid mistake by myself that I just can't find. I was getting segfaults from a Numba function and traced it down to the state I will outline here, but at this point I can't find anything anymore.

I have a function which operates on arrays. I have simplified it very far so I know its not making so much sense - but here it goes.

```python
import numpy as np
import numba

@numba.jit("f8[:,:](f8[:,:,:],f8[:,:],f8,f8,f8[:])", nopython=True, parallel=True,
           nogil=True)
def evalManyIndividual(individuals, X,p1, p2, p3):
    fitnesses = np.zeros((individuals.shape[0], 4))
    nF = individuals.shape[0]
    for i in numba.prange(nF):
        individual = individuals[i]
        P = np.random.random((3,4))
        fitnesses[i] = np.random.random((4,))
    return fitnesses
```

For debugging, Im using synthetic input data

```python
n = 15
m = 3
inputInd = np.random.random((500, n, m))
inputArray = np.random.random((n, m))
p1 = 25e-3
p2 = 55.
p3 = np.array([320., 240.])

ret = evalManyIndividualQ3D(inputInd, inputArray, p1, p2, p3)
```

Running this as a small script works. Running this from interactive works. However, I have a large library with Numba functions in which the one above is included. Just somewhere in there. Same syntax, copy & paste. If I then add the execution **with the same synthetic input data** after the library (compiling the full library including the above function), only calling the function as above

```python
ret = evalManyIndividualQ3D(inputInd, inputArray, p1, p2, p3)
```

Im getting a segfault. No traceback, no nothing. In terminal its `zsh: segmentation fault`, Jupyter just hangs completely. 

This happens on macOS 10.15. A difference to mention would be that during compilation of the library, Im getting some warnings

```python
NumbaPerformanceWarning: '@' is faster on contiguous arrays, called on (array(float64, 2d, A), array(float64, 2d, A))
[...]

PR #7625, opened 2021-12-06
Combined parfor chunking and caching PRs.

### LINKED ISSUE #2556: prange() scheduling is suboptimal when workload varies between iterations
When using `prange()` to parallelize a loop with a body that takes highly variable amounts of time per iteration, the scheduler will tend to leave some threads idle.  This is because the iteration range is divided between the threads upfront to maximize chunk size.  If the chunk size were smaller and fed into a work queue, it would allow better load balancing between the worker threads.

This is also likely something which becomes more efficient when we have TBB to do the thread dispatch.

See email thread: https://groups.google.com/a/continuum.io/forum/#!msg/numba-users/ZDWRtE_U2Y8/S-SbZt6YBQAJ

### LINKED ISSUE #3144: [Question] Would it be possible to support larger step size in `prange`?
I noticed that `numba.prange` in `parallel=True` accepts `step=1` only, but it would be extremely useful to have a larger step size (so functions like `all`, `min` can be performed along specific axis by specifying step size larger than 1)

DrTodd13:
This replaces #6025 and #7522.  There was overlap between these two PRs around using the dynamic thread count so rather than delaying the merge I went ahead and combined them.

This combined PR provides an API for selecting a parfor chunk size to deal with load balancing issues and it eliminates all use of static thread counts in generated parfor code.  Thus, parfor code (even with reductions) is now cacheable and if you change the chunksize or thread count after reloading from cache then you will use the new values as they are applied correctly in the code now.

Closes #2556
Closes #3144


stuartarchibald:
NOTE: I think this might also fix this: https://github.com/numba/numba/issues/5890#issuecomment-661848040

numba/parfors/parfor_lowering.py
```diff
@@ -1634,28 +1682,23 @@ def load_range(v):
...
+        llvmlite.ir.FunctionType(llvmlite.ir.VoidType(), [uintp_t]),
+        name="set_parallel_chunksize")
 
     get_num_threads = cgutils.get_or_insert_function(
         builder.module,
         llvmlite.ir.FunctionType(llvmlite.ir.IntType(types.intp.bitwidth), []),
         "get_num_threads")
 
     num_threads = builder.call(get_num_threads, [])
+    current_chunksize = builder.call(get_chunksize, [])
```
stuartarchibald, line 1701: I think this new IR generation would benefit from some commentary regarding the handling of the chunksize. It looks like the current chunksize is stored, then later, once the divisions have been calculated, the chunksize is zeroed during  kernel execution and then finally restored to the original value post execution of the kernel. What's the purpose of doing this? Is it to ensure subsequent chunksize use is zero?

numba/tests/doc_examples/test_parallel_chunksize.py
```diff
@@ -0,0 +1,75 @@
...
+                    acc += i
+                return acc
+
+            @njit(parallel=True)
+            def func2(n):
+                acc = 0
+                with parallel_chunksize(8):
+                    for i in prange(n):
+                        acc += i
+                return acc # Note: this cannot go inside the with
```
stuartarchibald, line 65: I think since 0.55 it can go inside the context managed block.

numba/tests/test_parfors.py
```diff
@@ -3981,6 +3983,107 @@ def test_issue_due_to_max_label(self):
         self.assertIn("TEST PASSED", out.decode())
 
 
+class TestParforChunksizing(TestCase):
+    """
+    Tests miscellaneous parts of ParallelAccelerator use.
```
stuartarchibald, line 3988:
```suggestion
    Tests chunksize handling in ParallelAccelerator.
```

numba/tests/test_parfors.py
```diff
@@ -3981,6 +3983,107 @@ def test_issue_due_to_max_label(self):
...
+    """
+    _numba_parallel_test_ = False
+
+    @skip_parfors_unsupported
+    def test_python_parallel_chunksize_basic(self):
```
stuartarchibald, line 3993: Should all these tests be context managed or include a `setUp/tearDown` which sets the chunksize to 0? This is so that if e.g. a test fails after it has set the chunksize to some non-default, that subsequent tests are not impacted by this state.

numba/tests/test_parfors.py
```diff
@@ -3981,6 +3983,107 @@ def test_issue_due_to_max_label(self):
...
+
+    @skip_parfors_unsupported
+    def test_all_iterations_reset_chunksize(self):
+        """ Test that all the iterations get run if you set the
+            chunksize.  Also check that the chunksize that each
+            worker thread sees has been reset to 0. """
+
+        @njit(parallel=True)
+        def test_impl(cs):
+            n = 1000
```
stuartarchibald, line 4051: Perhaps check a number of different array sizes and chunk sizes. 1000/8=125 so it's a relatively safe case?

numba/tests/test_parfors.py
```diff
@@ -3981,6 +3983,107 @@ def test_issue_due_to_max_label(self):
...
+                    res[i] = 13
+            return res, inner_cs
+
+        res, inner_cs = test_impl(8)
+        self.assertTrue(np.all(res == 13))
+        self.assertTrue(np.all(inner_cs == 0))
+
+    @skip_parfors_unsupported
+    def test_python_parallel_chunksize_negative(self):
+        # Test negative set_parallel_chunksize outside njit.
```
stuartarchibald, line 4066:
```suggestion
        # Test negative set_parallel_chunksize inside njit.
```

stuartarchibald (commented):
@DrTodd13 Many thanks for the patch and implementing this long awaited feature. I've given this an initial review. Outstanding more broad questions include:
* Should this also expose `get_thread_id` as part of the public API?
* Given that it is intended that these changes also fix issues with caching in relation to the number of threads in use, how should this be tested?
* What other testing is needed for this:
  * Across nested regions
  * Across processes
  * Involving checking that the schedule is as expected.
  * Anything else?

Thanks again!

numba/parfors/parfor_lowering.py
```diff
@@ -154,11 +181,11 @@ def _lower_parfor_parallel(lowerer, parfor):
...
-            num_threads_var = pfbdr.make_const_variable(
-                cval=thread_count,
-                typ=types.intp,
-                name='num_threads',
-            )
+            #num_threads_var = pfbdr.make_const_variable(
+            #    cval=thread_count,
+            #    typ=types.intp,
+            #    name='num_threads',
+            #)
```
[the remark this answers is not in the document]
DrTodd13: Done.

numba/core/analysis.py
```diff
@@ -689,3 +689,19 @@ def find_literally_calls(func_ir, argtypes):
         if not isinstance(query_arg, (types.Literal, types.InitialValue)):
             loc = first_loc[pos]
             raise errors.ForceLiteralArg(marked_args, loc=loc)
+
+
+ir_extension_use_alloca = {}
+
+
+def must_use_alloca(blocks):
```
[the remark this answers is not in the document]
DrTodd13: Done.

numba/parfors/parfor.py
```diff
@@ -3780,6 +3782,18 @@ def parfor_defs(parfor, use_set=None, def_set=None):
 analysis.ir_extension_usedefs[Parfor] = parfor_defs
 
 
+def parfor_use_alloca(parfor, alloca_set):
```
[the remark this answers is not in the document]
DrTodd13: Done.

numba/tests/test_parfors.py
DrTodd13: @stuartarchibald I'm not sure what to do here.  The context manager that would do this is what we are testing in many cases.  Could you make this change as you'd like it to be?  Thanks.

numba/tests/doc_examples/test_parallel_chunksize.py
DrTodd13: Just removed the comment.

numba/parfors/parfor_lowering.py
DrTodd13: I think I am forgetting some details here but what I remember is that the idea is to create an invariant that 0 is the default (which the docs state).  I don't remember if there was some issue such that nested cases had a problem if you don't do this reset to 0 at this boundary or if I made the decision that the invariant should also be true inside parfor regions because it is unlikely that a nested parfor would need the same chunksize.  I'm open to discussion here.  Please advise. @stuartarchibald

numba/parfors/parfor_lowering.py
```diff
@@ -1255,17 +1305,20 @@ def _create_gufunc_for_parfor_body(
...
-        # temp and accumulate on that temp to prevent false sharing.
-        if redtyp_is_scalar(typemap[var]):
-            gufunc_txt += "    " + param_dict[var] + \
-                 "=" + param_dict[arr] + "[0]\n"
-        else:
-            # The reduction variable is an array so np.copy it to a temp.
-            gufunc_txt += "    " + param_dict[var] + \
-                 "=np.copy(" + param_dict[arr] + ")\n"
+        gufunc_txt += "    " + param_dict[var] + \
+             "=" + param_dict[arr] + "[" + gufunc_thread_id_var + "]\n"
```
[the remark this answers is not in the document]
DrTodd13: The change is due to change in the way we handle the thread dimension of the reduction array.

numba/parfors/parfor_lowering.py
```diff
@@ -1255,17 +1305,20 @@ def _create_gufunc_for_parfor_body(
             gufunc_txt += ("    " + parfor_params_orig[pindex]
                 + " = np.ascontiguousarray(" + parfor_params[pindex] + ")\n")
 
+    gufunc_thread_id_var = "ParallelAcceleratorGufuncThreadId"
+
+    if len(parfor_redarrs) > 0:
+        gufunc_txt += "    " + gufunc_thread_id_var + " = "
+        gufunc_txt += "numba.np.ufunc.parallel._iget_thread_id()\n"
```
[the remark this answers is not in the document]
DrTodd13: Done.

numba/parfors/parfor_lowering.py
```diff
@@ -252,12 +279,33 @@ def _lower_parfor_parallel(lowerer, parfor):
...
+            numba_ir_loop_index_var = ir.Var(scope,
+                                             mk_unique_var("loop_index"), loc)
+            # Give that variable the right type.
+            typemap[numba_ir_loop_index_var.name] = num_thread_type
+            # Associate this Numba variable to the LLVm variable in the
+            # lowerer's varmap.
+            lowerer.varmap[numba_ir_loop_index_var.name] = alloc_loop_var
+            # Insert a loop into the outputed LLVM that goes from 0 to
+            # the current thread count.
+            with cgutils.for_range(builder, lowerer.loadvar(num_threads_var.name), intp=ntllvm_type) as loop:
```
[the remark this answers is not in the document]
DrTodd13: Someone should explore this later but this is a minor potential optimization in this particular context.

numba/parfors/parfor_lowering.py
```diff
@@ -258,12 +285,33 @@ def _lower_parfor_parallel(lowerer, parfor):
...
+            num_thread_type = typemap[num_threads_var.name]
+            # Get the LLVM type of the thread count variable.
+            ntllvm_type = targetctx.get_value_type(num_thread_type)
+            # Create a LLVM variable to hold the loop index.
+            alloc_loop_var = cgutils.alloca_once(builder, ntllvm_type)
+            # Associate this LLVM variable to a Numba IR variable so that
+            # we can use setitem IR builder.
+            # Create a Numba IR variable.
+            numba_ir_loop_index_var = ir.Var(scope,
+                                             mk_unique_var("loop_index"), loc)
```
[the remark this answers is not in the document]
DrTodd13: Done.

numba/core/withcontexts.py
```diff
@@ -464,3 +465,64 @@ def _mutate_with_block_callee(blocks, blk_start, blk_end, inputs, outputs):
...
+        scope = blocks[blk_start].scope
+        loc = blocks[blk_start].loc
+        if isinstance(arg, ir.Arg):
+            arg = ir.Var(scope, arg.name, loc)
+
+        set_state = []
+        restore_state = []
+
+        # global for Numba itself
+        gvar = ir.Var(scope, ir_utils.mk_unique_var("$ngvar"), loc)
```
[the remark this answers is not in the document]
DrTodd13: Done.

numba/core/withcontexts.py
```diff
@@ -464,3 +465,64 @@ def _mutate_with_block_callee(blocks, blk_start, blk_end, inputs, outputs):
...
+        set_state.append(ir.Assign(spcattr, spcvar, loc))
+        # call set_parallel_chunksize
+        orig_pc_var = ir.Var(scope, ir_utils.mk_unique_var("$save_pc"), loc)
+        cs_var = ir.Var(scope, ir_utils.mk_unique_var("$cs_var"), loc)
+        set_state.append(ir.Assign(arg, cs_var, loc))
+        spc_call = ir.Expr.call(spcvar, [cs_var], (), loc)
+        set_state.append(ir.Assign(spc_call, orig_pc_var, loc))
+
+        restore_spc_call = ir.Expr.call(spcvar, [orig_pc_var], (), loc)
+        restore_state.append(ir.Assign(restore_spc_call, orig_pc_var, loc))
```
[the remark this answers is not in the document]
DrTodd13: I think loc at restore_state as referring to start block location is correct.  It would seem even weirder if you did the loc after the block which may be some unrelated line of code.

numba/core/withcontexts.py
```diff
@@ -464,3 +465,64 @@ def _mutate_with_block_callee(blocks, blk_start, blk_end, inputs, outputs):
...
+
+        blocks[blk_start].body = (blocks[blk_start].body[1:-1] + 
+                                  set_state + 
+                                  [blocks[blk_start].body[-1]])
+        blocks[blk_end].body = restore_state + blocks[blk_end].body
+        ir_utils.dprint_func_ir(func_ir, "After with changes", blocks=blocks)
```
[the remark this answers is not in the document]
DrTodd13: This scheme is really ad-hoc at the moment and I don't know where to do it but I added it here for you.

numba/core/withcontexts.py
```diff
@@ -464,3 +465,64 @@ def _mutate_with_block_callee(blocks, blk_start, blk_end, inputs, outputs):
...
+
+    def __call__(self, *args, **kwargs):
+        assert len(args) == 1
+        assert not kwargs
+        self.chunksize = args[0]
+        return self
```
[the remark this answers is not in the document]
DrTodd13: Done.

numba/core/ir_utils.py
```diff
@@ -1989,7 +1989,7 @@ def find_global_value(func_ir, var):
     or raise GuardException otherwise.
     """
     dfn = get_definition(func_ir, var)
-    if isinstance(dfn, ir.Global):
+    if isinstance(dfn, (ir.Global, ir.FreeVar)):
```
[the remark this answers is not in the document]
DrTodd13: I think the intent of this function is really "get value outside current scope".  I've changed named to reflect as it is only used in one place as far as I can tell.

numba/core/types/functions.py
```diff
@@ -628,6 +628,10 @@ def generic(self, args, kws):
     def key(self):
         return self.sig, self.cconv, self.get_pointer
 
+    def unify(self, typingctx, other):
+        if isinstance(other, Function):
+            return self
```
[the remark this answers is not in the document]
DrTodd13: Weird I admit.  Other suggestions welcome.  Pointers can be function pointers...but of course they could be anything else as well.  I don't remember the details of where this came up but comment it out and I'm sure it will pop up again. 
 @stuartarchibald

numba/__init__.py
```diff
@@ -40,7 +40,8 @@
 
 # Re-export vectorize decorators and the thread layer querying function
 from numba.np.ufunc import (vectorize, guvectorize, threading_layer,
-                            get_num_threads, set_num_threads)
+                            get_num_threads, set_num_threads,
+                            set_parallel_chunksize, get_parallel_chunksize)
```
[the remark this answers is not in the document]
DrTodd13: We don't currently support the full complexity of OpenMP in this regard.  "0" is somewhat akin to openmp static with chunk 0.  Non-0 is somewhat akin, but not exactly like, openmp dynamic with the given chunk size.  To support the openmp api, we'd need to implement the two combinations that we don't currently support, e.g., static with non-0 chunk.  I'm not opposed to this but perhaps we can consider doing this later.  @stuartarchibald

numba/__init__.py
```diff
@@ -51,6 +52,7 @@
 # Initialize withcontexts
 import numba.core.withcontexts
 from numba.core.withcontexts import objmode_context as objmode
+from numba.core.withcontexts import parallel_chunksize
```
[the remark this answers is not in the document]
DrTodd13: Same as above.

numba/np/ufunc/workqueue.c
```diff
@@ -490,6 +523,12 @@ add_task(void *fn, void *args, void *dims, void *steps, void *data)
     }
 }
 
+static void
+add_task(void *fn, void *args, void *dims, void *steps, void *data)
+{
+    add_task_internal(fn, args, dims, steps, data, 0);
+}
+
```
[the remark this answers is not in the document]
DrTodd13: This one is up to you but I propose we don't do it in this PR.

numba/np/ufunc/workqueue.c
```diff
@@ -62,9 +62,14 @@ static int _nesting_level = 0;
    free the task-queue, too. */
 static void reset_after_fork(void);
 
+static void
+add_task_internal(void *fn, void *args, void *dims, void *steps, void *data, int tid);
+
 /* PThread */
 #ifdef NUMBA_PTHREAD
 
+static pthread_key_t tidkey;
```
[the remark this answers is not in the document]
DrTodd13: It has been so long again that I can't remember.  Here's what I can say.  My first instinct would have been to use a language construct like THREAD_LOCAL and if I didn't and instead used an API that I wasn't really familiar with before then there must have been a reason.  I can't remember the reason now.

numba/parfors/parfor_lowering.py
```diff
@@ -114,19 +114,45 @@ def _lower_parfor_parallel(lowerer, parfor):
...
+
         # For each reduction variable...
         for i in range(nredvars):
-            redvar_typ = lowerer.fndesc.typemap[parfor_redvars[i]]
-            redvar = ir.Var(scope, parfor_redvars[i], loc)
+            red_name = parfor_redvars[i]
+            # Get the type of the reduction variable.
+            redvar_typ = lowerer.fndesc.typemap[red_name]
+            # Get the ir.Var for the reduction variable.
+            redvar = ir.Var(scope, red_name, loc)
```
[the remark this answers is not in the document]
DrTodd13: It is already in the typemap.  See like 138.  There's two cases, 1) constructing a new Var and 2) constructing a Var object from a name (string) that is already a variable.  This one here is type 2 I believe.

sklam:
@DrTodd13, patch to make the reduction use runtime threadcount is at https://github.com/IntelLabs/numba/pull/70

stuartarchibald:
> Can you make a suggestion here? I am out of office attending to my father who is very ill. Is there no way to test in get_thread_id if we are not in a parallel region?

@DrTodd13 I've a couple of ideas about how to do this and will attempt those, I can take care of this and report back here when a solution is found. Will catch up OOB.

stuartarchibald:
@DrTodd13 I've added a few patches in https://github.com/stuartarchibald/numba/commits/pr_7625_cont_1 on top of this PR, the commit messages should hopefully explain what they do. 
*  https://github.com/stuartarchibald/numba/commit/5950626c898224baa22390432e100f823fb7067f
* https://github.com/stuartarchibald/numba/commit/166593a9dedf80d49276a43df2e1309b2fab8125
* https://github.com/stuartarchibald/numba/commit/b27ffdb1e0c171833ae2b8e7d1ca964fb7ab35d8
* https://github.com/stuartarchibald/numba/commit/f6e91835e6d12d8aac173be41de854da70cfd493

They essentially fix a few bugs I found and do a small amount of refactoring. I think the first two are necessary for correctness. If you are happy with them I can make a PR against this branch for you to look at? Thanks!

stuartarchibald:
> @stuartarchibald Looks good to me so please create the PR. Thanks!

Thanks for checking @DrTodd13, PR is https://github.com/IntelLabs/numba/pull/71

DrTodd13:
> @DrTodd13 As discussed, this is a test which is producing behaviour that I'm not sure is expected:
> 
> ```python
> from numba import njit, parallel_chunksize, get_parallel_chunksize, prange
> import numpy as np
> 
> @njit(parallel=True)
> def bar():
>     acc = 0
>     print("Bar entry:", get_parallel_chunksize())
>     with parallel_chunksize(30):
>         print("Bar ctx entry:", get_parallel_chunksize())
>         for i in prange(1):
>             acc += i
>         print("Bar ctx exit:", get_parallel_chunksize())
>     print("Bar exit:", get_parallel_chunksize())
>     return acc
> 
> 
> @njit(parallel=True)
> def foo():
>     acc = 0
>     print("Foo entry:", get_parallel_chunksize())
>     with parallel_chunksize(10):
>         print("Foo ctx entry:", get_parallel_chunksize())
>         for i in prange(1):
>             acc += bar()
>         print("Foo ctx exit:", get_parallel_chunksize())
>     print("Foo exit:", get_parallel_chunksize())
> 
> foo()
> ```
> 
> which for me gives:
> 
> ```
> Foo entry: 0
> Foo ctx entry: 10
> Bar entry: 0
> Bar ctx entry: 30
> Bar ctx exit: 30
> Bar exit: 0
> Foo ctx exit: 10
> Foo exit: 0
> ```
> 
> I'm not sure why `Bar entry/exit` has a value of 0. For comparison, this is what happens under `NUMBA_DISABLE_JIT=1`:
> 
> ```
> $ NUMBA_DISABLE_JIT=1 python test_7625_nested_ctx.py 
> Foo entry: 0
> Foo ctx entry: 10
> Bar entry: 10
> Bar ctx entry: 30
> Bar ctx exit: 30
> Bar exit: 10
> Foo ctx exit: 10
> Foo exit: 10
> ```

@stuartarchibald 
Now that I see your example, this is the expected behavior.  Well, at least it is what I intended so the question remains is will you agree that this should be the expected behavior.  What is going on is that when the parallel executor starts, it resets the chunksize to 0 (do the default simple equal division of work to threads) for each worker.  The reason that I did this is that I was thinking that a custom chunksize for one parallel region is unlikely to be right for a nested parallel region so best to reset to the default behavior.  So, if you don't enter any parallel regions then it makes sense that the chunksizes are maintained in the non-jit (or non-parallel) mode.  Obviously, this goes against your expectations and if I hadn't been the one to implement it then that might have been my expectation as well.  I can easily make it do what you expect but my only reluctance for me to do that is that I think that being the default is unlikely to be best performance wise and it will be easy for programmers to forget they are coming from a chunksize context that may have an effect they don't want.  So, if they give it no thought then my approach may be better.  If they give it a modicum of thought then they might expect the chunksize to carry over into the parallel workers, but the question is if they give it even more thought whether they will want it to carry over by default or not.  Curious as to your opinion.

stuartarchibald:
Capturing the failing test from CI. OSX python 3.9, NumPy 1.21 failed with:

<details>

```
FAIL: test_TestParforNumPy (numba.tests.test_parfors.TestParforsRunner)
  File "/Users/runner/work/1/s/numba/tests/test_parfors.py", line 91, in test_TestParforNumPy
    self.runner()
  File "/Users/runner/work/1/s/numba/tests/test_parfors.py", line 80, in runner
    self.subprocess_test_runner(test_module=themod,
  File "/Users/runner/work/1/s/numba/tests/support.py", line 581, in subprocess_test_runner
AssertionError: 1 != 0 : 
captured stdout: 
captured stderr: /Users/runner/work/1/s/numba/core/typed_passes.py:329: NumbaPerformanceWarning: 
The keyword argument 'parallel=True' was specified but no transformation for parallel execution was possible.

To find out why, try turning on parallel diagnostics, see https://numba.readthedocs.io/en/stable/user/parallel.html#diagnostics for help.

File "numba/tests/test_parfors.py", line 1247:
    def test_0d_array(self):
        def test_impl(n):
        ^

  warnings.warn(errors.NumbaPerformanceWarning(msg,
..F./Users/runner/work/1/s/numba/core/typed_passes.py:329: NumbaPerformanceWarning: 
The keyword argument 'parallel=True' was specified but no transformation for parallel execution was possible.

To find out why, try turning on parallel diagnostics, see https://numba.readthedocs.io/en/stable/user/parallel.html#diagnostics for help.

File "numba/tests/test_parfors.py", line 1014:
    def test_dead_randoms(self):
        def test_impl(n):
        ^

  warnings.warn(errors.NumbaPerformanceWarning(msg,
..................
FAIL: test_argmax (numba.tests.test_parfors.TestParforNumPy)
  File "/Users/runner/work/1/s/numba/tests/test_parfors.py", line 1115, in test_argmax
    self.check(test_impl1, A)
  File "/Users/runner/work/1/s/numba/tests/test_parfors.py", line 282, in check
    self.check_parfors_vs_others(pyfunc, cfunc, cpfunc, *args, **kwargs)
  File "/Users/runner/work/1/s/numba/tests/test_parfors.py", line 260, in check_parfors_vs_others
    np.testing.assert_almost_equal(parfor_output, py_expected, **kwargs)
  File "/Users/runner/miniconda3/envs/azure_ci/lib/python3.9/site-packages/numpy/testing/_private/utils.py", line 599, in assert_almost_equal
    raise AssertionError(_build_err_msg())
AssertionError: 
Arrays are not almost equal to 7 decimals
 ACTUAL: 4
 DESIRED: 2

Ran 22 tests in 235.433s

FAILED (failures=1)

```

</details>

i.e. incorrect result from `numba.tests.test_parfors.TestParforNumPy.test_argmax`

I cannot reproduce this locally.

Looking at the test implementation:
https://github.com/numba/numba/blob/b1be2f12c83c01f57fe34fab9a9d77334f9baa1d/numba/tests/test_parfors.py#L1102-L1133
seems like there's use of `np.random.randint` and `np.random.ranf` without a `np.seed` being set, which means this test is not testing with the same values across runs. I'm not convinced that this test failure is a result of this PR vs. some combination of values from the RNG that trigger an already present bug. Will rerun the test job and see if it passes.

numba/__init__.py
```diff
@@ -84,6 +87,8 @@ def test(argv, **kwds):
     literal_unroll
     get_num_threads
     set_num_threads
+    set_parallel_chunksize
+    get_parallel_chunksize
```
stuartarchibald, line 91:
```suggestion
    get_parallel_chunksize
    parallel_chunksize
```

numba/core/withcontexts.py
```diff
@@ -464,3 +466,70 @@ def _mutate_with_block_callee(blocks, blk_start, blk_end, inputs, outputs):
         block=ir.Block(scope=scope, loc=loc),
         outputs=outputs,
     )
+
+class _ParallelChunksize(WithContext):
```
stuartarchibald, line 470: In the interests of keeping parfors specific code together, should this be moved into a module under `numba.parfors` ?

stuartarchibald (commented):
Thanks for the updates @DrTodd13 there's a few minor things to look at following the changes, I'm also going to add some more general questions/discussion directly to the PR. Thanks again.

stuartarchibald:
> > @DrTodd13 As discussed, this is a test which is producing behaviour that I'm not sure is expected:
> > ```python
> > from numba import njit, parallel_chunksize, get_parallel_chunksize, prange
> > import numpy as np
> > 
> > @njit(parallel=True)
> > def bar():
> >     acc = 0
> >     print("Bar entry:", get_parallel_chunksize())
> >     with parallel_chunksize(30):
> >         print("Bar ctx entry:", get_parallel_chunksize())
> >         for i in prange(1):
> >             acc += i
> >         print("Bar ctx exit:", get_parallel_chunksize())
> >     print("Bar exit:", get_parallel_chunksize())
> >     return acc
> > 
> > 
> > @njit(parallel=True)
> > def foo():
> >     acc = 0
> >     print("Foo entry:", get_parallel_chunksize())
> >     with parallel_chunksize(10):
> >         print("Foo ctx entry:", get_parallel_chunksize())
> >         for i in prange(1):
> >             acc += bar()
> >         print("Foo ctx exit:", get_parallel_chunksize())
> >     print("Foo exit:", get_parallel_chunksize())
> > 
> > foo()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > which for me gives:
> > ```
> > Foo entry: 0
> > Foo ctx entry: 10
> > Bar entry: 0
> > Bar ctx entry: 30
> > Bar ctx exit: 30
> > Bar exit: 0
> > Foo ctx exit: 10
> > Foo exit: 0
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'm not sure why `Bar entry/exit` has a value of 0. For comparison, this is what happens under `NUMBA_DISABLE_JIT=1`:
> > ```
> > $ NUMBA_DISABLE_JIT=1 python test_7625_nested_ctx.py 
> > Foo entry: 0
> > Foo ctx entry: 10
> > Bar entry: 10
> > Bar ctx entry: 30
> > Bar ctx exit: 30
> > Bar exit: 10
> > Foo ctx exit: 10
> > Foo exit: 10
> > ```
> 
> @stuartarchibald Now that I see your example, this is the expected behavior. Well, at least it is what I intended so the question remains is will you agree that this should be the expected behavior. What is going on is that when the parallel executor starts, it resets the chunksize to 0 (do the default simple equal division of work to threads) for each worker. The reason that I did this is that I was thinking that a custom chunksize for one parallel region is unlikely to be right for a nested parallel region so best to reset to the default behavior. So, if you don't enter any parallel regions then it makes sense that the chunksizes are maintained in the non-jit (or non-parallel) mode. Obviously, this goes against your expectations and if I hadn't been the one to implement it then that might have been my expectation as well. I can easily make it do what you expect but my only reluctance for me to do that is that I think that being the default is unlikely to be best performance wise and it will be easy for programmers to forget they are coming from a chunksize context that may have an effect they don't want. So, if they give it no thought then my approach may be better. If they give it a modicum of thought then they might expect the chunksize to carry over into the parallel workers, but the question is if they give it even more thought whether they will want it to carry over by default or not. Curious as to your opinion.

@DrTodd13 I think what you describe makes sense in terms of picking defaults that are likely to give reasonable performance in these situations and am happy to leave the implementation behaving in this way. If nested regions would benefit from some specific chunksize then it's relatively easy to just add this in to user code. My only suggestion would be to ensure that this is all explicitly documented (as it does differ to all the other TLS properties, which are transferred into nested regions) and that the deviation between the context manager behaviour when called from Python and compiled code is noted. Thanks.

stuartarchibald:
NOTE: Tested the nesting behaviour and `get_thread_id` with:
```python
from numba import njit, prange, get_thread_id
import numpy as np

@njit(parallel=True)
def inner(n):
    acc = 0
    for i in prange(n):
        print("inner ", get_thread_id())
        acc += 1
    print("nested serial", get_thread_id())
    for i in prange(n):
        print("inner ", get_thread_id())
        acc += 1

@njit(parallel=True)
def outer(n):
    acc = 0
    for i in prange(n):
        print("outer ", get_thread_id())
        acc += 1
        inner(3)
    print("outer serial", get_thread_id())
    for i in prange(n):
        print("outer ", get_thread_id())
        acc += 1
        inner(3)

outer(3)
```
and the C code:
```C
#include <omp.h>
#include <stdlib.h>
#include <stdio.h>


double inner(size_t n)
{
    double acc = 0.;

#pragma omp parallel for
    for (size_t i = 0; i < n; i++)
    {
        printf("inner %d\n", omp_get_thread_num());
        acc += 1.;
    }

    printf("nested serial %d\n", omp_get_thread_num());

#pragma omp parallel
    for (size_t i = 0; i < n; i++)
    {
        printf("inner %d\n", omp_get_thread_num());
        acc += 1.;
    }
    return acc;
}

double outer(size_t n)
{
    double acc = 0.;
#pragma omp parallel for
    for (size_t i = 0; i < n; i++)
    {
        printf("outer %d\n", omp_get_thread_num());
        acc += 1.;
        acc += inner(2);
    }

    printf("outer serial %d\n", omp_get_thread_num());

#pragma omp parallel for
    for (size_t i = 0; i < n; i++)
    {
        printf("outer %d\n", omp_get_thread_num());
        acc += 1.;
        acc += inner(2);
    }

    return acc;
}


int main(void){
    omp_set_nested(0x1);
    outer(3);
}
```
In Python, the reported `thread ID` when using either the OpenMP or TBB backed does approximately the same as the C code. i.e. the call in the "nested serial" region returns the ID of the thread that is the "main" thread in the nested region.

numba/core/withcontexts.py
DrTodd13: I'm torn.  I can see arguments on both sides.  I'll leave the decision up to you.

numba/np/ufunc/parallel.py
```diff
@@ -640,35 +655,97 @@ def ol_get_num_threads():
...
+    return signature(types.intp), codegen
+
+
+def get_thread_id():
+    """
+    Returns a unique ID for each thread in the range 0 to N.
+    If we get to this point then we are not in a Numba compiled function.
+    Return the result from get_thread_id() from the threading layer.
+    If we are in a Numba compiled function then the overload below will
+    be exercised.
```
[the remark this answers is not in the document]
DrTodd13: Done.

DrTodd13:
> 

I added a big block of descriptive text in the documentation about this resetting of chunksize inside parallel regions.

numba/core/withcontexts.py
```diff
@@ -464,3 +466,70 @@ def _mutate_with_block_callee(blocks, blk_start, blk_end, inputs, outputs):
...
+        ir_utils.dprint_func_ir(func_ir, "After with changes", blocks=blocks)
+
+    def __call__(self, *args, **kwargs):
+        """Act like a function and enforce the contract that
+        setting the chunksize takes only one integer input.
+        """
+        if len(args) != 1 or kwargs or not isinstance(args[0], int):
+            raise ValueError("parallel_chunksize takes only a "
+                             "single integer argument.")
+
```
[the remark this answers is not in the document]
DrTodd13: Despite the different implementations, I'd like the behavior of the with context to be the same both inside and outside njit.  Inside njit, you might not have a literal value you can check at compile time but the value is checked in all paths in the C code.  So, I would suggest that we just keep the singular check there.

stuartarchibald:
Thanks for the update on the docs @DrTodd13, I'll take a look at it shortly. Also, public CI is failing, looks like it's just some `flake8` problems:
```
numba/tests/doc_examples/test_parallel_chunksize.py:23:17: F841 local variable 'pre_prange_chunksize' is assigned to but never used
numba/tests/doc_examples/test_parallel_chunksize.py:25:21: F841 local variable 'in_prange_chunksize' is assigned to but never used
numba/tests/doc_examples/test_parallel_chunksize.py:25:81: E501 line too long (83 > 80 characters)
numba/tests/doc_examples/test_parallel_chunksize.py:27:17: F841 local variable 'post_prange_chunksize' is assigned to but never used
numba/tests/doc_examples/test_parallel_chunksize.py:27:81: E501 line too long (81 > 80 characters)
numba/np/ufunc/parallel.py:680:81: E501 line too long (108 > 80 characters)
```

docs/source/user/parallel.rst
```diff
@@ -601,4 +601,105 @@ The report is split into the following sections:
...
+However, upon entering a parallel region, Numba sets the chunk size thread local variable
+for each of the threads executing that parallel region back to the default of 0,
+since it is unlikely
+that any nested parallel regions would require the same chunk size.  If the same thread is
+used to execute a sequential and parallel region then that thread's chunk size
+variable is set to 0 at the beginning of the parallel region and restored to
+its original value upon exiting the parallel region.
+This behavior is demonstrated in ``func1`` in the example below in that the
+reported chunk size inside the ``prange`` parallel region is 0 but is 4 outside
+the parallel region.  Note that if the prange is not executed in parallel for
```
stuartarchibald, line 662:
```suggestion
the parallel region.  Note that if the ``prange`` is not executed in parallel for
```

stuartarchibald (commented):
Thanks for the doc update @DrTodd13, this explanation makes sense and should hopefully clarify the behaviour for users. There's a single markup issue but otherwise the docs change looks good.

sklam (commented):
Two groups of comments:

1. cxx code that compares int to size_t 
2. no need for `mk_unique_var` if `scope.redefine` is used

numba/core/withcontexts.py
```diff
@@ -464,3 +466,70 @@ def _mutate_with_block_callee(blocks, blk_start, blk_end, inputs, outputs):
...
+        # global for Numba itself
+        gvar = scope.define(ir_utils.mk_unique_var("$ngvar"), loc)
+        set_state.append(ir.Assign(ir.Global('numba', numba, loc), gvar, loc))
+        # getattr for set chunksize function in Numba
+        spcattr = ir.Expr.getattr(gvar, 'set_parallel_chunksize', loc)
+        spcvar = scope.define(ir_utils.mk_unique_var("$spc"), loc)
+        set_state.append(ir.Assign(spcattr, spcvar, loc))
+        # call set_parallel_chunksize
+        orig_pc_var = scope.define(ir_utils.mk_unique_var("$save_pc"), loc)
+        cs_var = scope.define(ir_utils.mk_unique_var("$cs_var"), loc)
```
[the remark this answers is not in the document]
DrTodd13: Done.

sklam:
I found an compilation error with a common pattern to use `get_thread_id()`:

```python
import numpy as np
from numba import njit, prange, set_num_threads, get_num_threads
from numba import get_thread_id


@njit(parallel=True)
def foo(arr, data):
    # In this version,
    set_num_threads(data.size)
    for i in prange(arr.size):
        tid = get_thread_id()
        ## Uncomment the print to stop `tid` from apparently
        ## optimized to 0 for all threads
        # print(tid)
        data[tid] += 1
        arr[i] = tid

arr = np.arange(32)
data = np.zeros(4)
foo(arr, data)

print(arr)
print(data)
```

Without the `print(tid)`, the output is:

```
[0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
[32.  0.  0.  0.]
```
as if `tid` is assumed to be `0` all the time.

With the `print(tid)`, the output is:

```
[0 0 0 0 0 0 0 0 3 3 3 3 3 3 3 3 2 2 2 2 2 2 2 2 0 0 0 0 0 0 0 0]
[16.  0.  8.  8.]
```

sklam:
[The previous error](https://github.com/numba/numba/pull/7625#issuecomment-1151688972) occurs on `main` as well

stuartarchibald:
> I found an compilation error with a common pattern using `get_thread_id()`:
> 
> ```python
> import numpy as np
> from numba import njit, prange, set_num_threads, get_num_threads
> from numba import get_thread_id
> 
> 
> @njit(parallel=True)
> def foo(arr, data):
>     # In this version,
>     set_num_threads(data.size)
>     for i in prange(arr.size):
>         tid = get_thread_id()
>         ## Uncomment the print to stop `tid` from apparently
>         ## optimized to 0 for all threads
>         # print(tid)
>         data[tid] += 1
>         arr[i] = tid
> 
> arr = np.arange(32)
> data = np.zeros(4)
> foo(arr, data)
> 
> print(arr)
> print(data)
> ```
> 
> Without the `print(tid)`, the output is:
> 
> ```
> [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
> [32.  0.  0.  0.]
> ```
> 
> as if `tid` is assumed to be `0` all the time.
> 
> With the `print(tid)`, the output is:
> 
> ```
> [0 0 0 0 0 0 0 0 3 3 3 3 3 3 3 3 2 2 2 2 2 2 2 2 0 0 0 0 0 0 0 0]
> [16.  0.  8.  8.]
> ```

I'm not sure yet that this is a bug. Using the TBB threading layer I can reproduce the above result, but I cannot reproduce with OpenMP or workqueue. Assuming the TBB threading layer is being used, there's no guarantee that threads will spin up and/or join the task arena for a given parallel region unless there's sufficient work, _without_ the `print` the amount of work is tiny so it probably "decides" it's not worth adding more threads, but _with_ the `print` it might just go over some threshold for enough work to add more threads. 

The "adding some more threads" can be observed in the output:

> [0 0 0 0 0 0 0 0 3 3 3 3 3 3 3 3 2 2 2 2 2 2 2 2 0 0 0 0 0 0 0 0]

There's no thread ID of 1 (assuming 4 cores), it's decided to use thread 0 (main thread) and then add in a couple of others. I also note that making the array `arr` progressively larger and running _without_ the `print` leads to more and more threads joining to do work. Does this make sense or are you thinking of something else?

sklam:
Ok, i agree. Both omp and workqueue backends are doing the right thing. So it is probably just TBB scheduling at work in my original report.

DrTodd13:
@sklam I fixed some problems with your suggestions in the C file relating to printf format specifiers but now 5 of the tests are failing "before install."  I haven't seen this before so don't know what to suggest but perhaps you know.  I don't think the problems are related to your suggestions or my fixes.

stuartarchibald:
> @sklam I fixed some problems with your suggestions in the C file relating to printf format specifiers but now 5 of the tests are failing "before install." I haven't seen this before so don't know what to suggest but perhaps you know. I don't think the problems are related to your suggestions or my fixes.

@DrTodd13 don't think it's this patch, think it's #8152 (fixed by #8151).

stuartarchibald:
On a further look, whilst this test has some deterministic issues (the RNG seed isn't set) I think this failure on the Azure OSX build is genuine and I can reproduce it locally using this branch and the TBB backend (but not on mainline, or with the OpenMP backend). The error is:
```
2022-06-14T14:03:06.8684310Z ======================================================================
2022-06-14T14:03:06.8685090Z FAIL: test_argmax (numba.tests.test_parfors.TestParforNumPy)
2022-06-14T14:03:06.8686580Z ----------------------------------------------------------------------
2022-06-14T14:03:06.8687270Z Traceback (most recent call last):
2022-06-14T14:03:06.8687900Z   File "/Users/runner/work/1/s/numba/tests/test_parfors.py", line 1118, in test_argmax
2022-06-14T14:03:06.8688540Z     self.check(test_impl2, A)
2022-06-14T14:03:06.8689140Z   File "/Users/runner/work/1/s/numba/tests/test_parfors.py", line 282, in check
2022-06-14T14:03:06.8689850Z     self.check_parfors_vs_others(pyfunc, cfunc, cpfunc, *args, **kwargs)
2022-06-14T14:03:06.8690580Z   File "/Users/runner/work/1/s/numba/tests/test_parfors.py", line 260, in check_parfors_vs_others
2022-06-14T14:03:06.8691370Z     np.testing.assert_almost_equal(parfor_output, py_expected, **kwargs)
2022-06-14T14:03:06.8692790Z   File "/Users/runner/miniconda3/envs/azure_ci/lib/python3.10/site-packages/numpy/testing/_private/utils.py", line 599, in assert_almost_equal
2022-06-14T14:03:06.8693900Z     raise AssertionError(_build_err_msg())
2022-06-14T14:03:06.8694440Z AssertionError: 
2022-06-14T14:03:06.8694980Z Arrays are not almost equal to 7 decimals
2022-06-14T14:03:06.8695620Z  ACTUAL: 4
2022-06-14T14:03:06.8696090Z  DESIRED: 2
```
I used this locally to reproduce (it's a trimmed down version of the failing test in the test suite), it usually hits the breakpoint within 50 iterations:
```python
from numba import njit
import numpy as np

def check(impl, arg):
    njitted = njit(impl)
    pjitted = njit(parallel=True)(impl)
    python = impl

    expected = python(arg.copy())
    got_njitted = njitted(arg.copy())
    got_pjitted = pjitted(arg.copy())

    try:
        np.testing.assert_allclose(expected, got_njitted)
        np.testing.assert_allclose(expected, got_pjitted)
    except Exception as e:
        import pdb; pdb.set_trace()
        pass
    else:
        print("ok")

def test_argmax():
    def test_impl1(A):
        return A.argmax()

    def test_impl2(A):
        return np.argmax(A)

    n = 211
    A = np.array([1., 0., 3., 2., 3.])
    B = np.random.randint(10, size=n).astype(np.int32)
    C = np.random.ranf((n, n))  # test multi-dimensional array
    check(test_impl1, A)
    check(test_impl1, B)
    check(test_impl1, C)
    check(test_impl2, A)
    check(test_impl2, B)
    check(test_impl2, C)

for i in range(1000):
    test_argmax()
```
output:
```
(Pdb) l
 13         try:
 14             np.testing.assert_allclose(expected, got_njitted)
 15             np.testing.assert_allclose(expected, got_pjitted)
 16         except Exception as e:
 17             import pdb; pdb.set_trace()
 18  ->         pass
 19         else:
 20             print("ok")
 21  
 22     def test_argmax():
 23         def test_impl1(A):
(Pdb) p e
AssertionError('\nNot equal to tolerance rtol=1e-07, atol=0\n\nMismatched elements: 1 / 1 (100%)\nMax absolute difference: 2\nMax relative difference: 0.5\n x: array(2)\n y: array(4)')
(Pdb) p arg
array([1., 0., 3., 2., 3.])
```
and it looks like it's failing using the inlined array input `array([1., 0., 3., 2., 3.])`, where indexes 2 and 4 both hold the same value.

I think this will need resolving as either "existing bug" or "bug in this patch" (and fixed) prior to merge.

stuartarchibald:
@DrTodd13 as discussed OOB, the few outstanding items on this PR are:
* Set the OpenMP backend parallel for-loop to use dynamic scheduling (else the chunksize won't have as much impact?)
* Add a motivating example to the user docs with a really biased work load, e.g. work load increasing with respect to loop index, and offer some basic examples of the use of setting the chunk size and the context manager.
* Move the existing detailed documentation on the implementation into the developer docs (suggest into here: https://github.com/numba/numba/blob/158b117234db18c9bf415a4624340c04a375fdf4/docs/source/developer/threading_implementation.rst?plain=1#L1-L3)
* Add a test for the issue reported in #7518

Many thanks!

DrTodd13:
> * Set the OpenMP backend parallel for-loop to use dynamic scheduling (else the chunksize won't have as much impact?

The default chunksize of 1 with dynamic would work well for parfor use cases but is there going to be too much overhead for other uses of the vectorize/guvectorize path?

> * Move the existing detailed documentation on the implementation into the developer docs (suggest into here: https://github.com/numba/numba/blob/158b117234db18c9bf415a4624340c04a375fdf4/docs/source/developer/threading_implementation.rst?plain=1#L1-L3
>   )

Which existing detailed documentation?

stuartarchibald:
> > * Set the OpenMP backend parallel for-loop to use dynamic scheduling (else the chunksize won't have as much impact?
> 
> The default chunksize of 1 with dynamic would work well for parfor use cases but is there going to be too much overhead for other uses of the vectorize/guvectorize path?

This is a good point, I think for low intensity kernels the dynamic scheduler would end up causing overhead. What do you think about doing this: the `parallel_for` function API could have an additional argument `int dynamic_scheduling` which if set can be used to inform the scheduling at runtime. I think if this were implemented the `omp parallel for` could have `schedule(runtime)` with it configured at runtime via `omp_set_schedule(dynamic/static)` based on the `dynamic_scheduling` argument? Parfors can set this argument as part of its code generation and the standard `guvectorize/vectorize` can elect to not set it. I think this will "fix" the issue with OpenMP, and also let the TBB backend specify a more specific partitioning.

> > * Move the existing detailed documentation on the implementation into the developer docs (suggest into here: https://github.com/numba/numba/blob/158b117234db18c9bf415a4624340c04a375fdf4/docs/source/developer/threading_implementation.rst?plain=1#L1-L3
> >   
> >   )
> 
> Which existing detailed documentation?

The docs here: https://github.com/numba/numba/pull/7625/files#diff-03a635d2d8fbe72934bd351660296e5bb40b60490860eb63b6490db8fbc8c02bR614, these are very detailed and great for those wanting to know about all about the implementation details. For the first item in the user facing docs, was thinking it might be good to provide a more simple introductory overview with a motivating example of a poorly balanced load, and then to reference out to the more detailed documentation, what do you think?

numba/cpython/builtins.py
```diff
@@ -550,8 +550,29 @@ def indval_min(indval1, indval2):
     if isinstance(indval1, IndexValueType) and \
        isinstance(indval2, IndexValueType):
         def min_impl(indval1, indval2):
-            if indval1.value > indval2.value:
+            # nan check
```
stuartarchibald, line 553: Thanks for fixing this, suggest adding unit tests that include `np.nan` values as part of their input to `np.argmax`/`np.argmin`.

DrTodd13:
@stuartarchibald Per our discussion, we're holding off on all the changes for dynamic scheduling across all backends.  I moved some of the documentation details about chunk sizing to the developer files.  I added a test of caching of a parfor function with a reduction with different thread numbers but that really needs a good review as I'm not very confident I used the infrastructure in test_caching.py correctly.  I also added a np.nan test in argmin and argmax.

numba/cpython/builtins.py
DrTodd13: Done.

stuartarchibald:
> @stuartarchibald Per our discussion, we're holding off on all the changes for dynamic scheduling across all backends. I moved some of the documentation details about chunk sizing to the developer files. I added a test of caching of a parfor function with a reduction with different thread numbers but that really needs a good review as I'm not very confident I used the infrastructure in test_caching.py correctly. I also added a np.nan test in argmin and argmax.

@DrTodd13 Thanks for adding these updates. I'll make sure that I take a good look at the caching tests. It turns out that the public CI build is currently failing, seems like the test:
```
numba.tests.doc_examples.test_parallel_chunksize.ChunksizeExamplesTest.test_unbalanced_example
```
is producing:
```
numba.core.errors.UnsupportedRewriteError: Failed in nopython mode pipeline (step: convert to parfors)
Overwrite of parallel loop index

File "numba/tests/doc_examples/test_parallel_chunksize.py", line 29:
            def func1():
                <source elided>
                        if cur % 2 == 0:
                            cur //= 2
                            ^
```
which doesn't look quite right given `cur` is not the loop index var, but it is an alias of it?

numba/tests/doc_examples/test_parallel_chunksize.py
```diff
@@ -0,0 +1,105 @@
...
+                return acc
+
+            # This version saves the previous chunksize as returned
+            # by set_parallel_chunksize.
+            old_chunksize = set_parallel_chunksize(4)
+            func1(12)
+            func2(12)
+            func1(12)
+            set_parallel_chunksize(old_chunksize)
+            # magictoken.ex_chunksize_manual.end
```
stuartarchibald, line 75: I think this needs something like:
```suggestion
            result1 = func1(12)
            result2 = func2(12)
            result3 = func1(12)
            set_parallel_chunksize(old_chunksize)
            # magictoken.ex_chunksize_manual.end
            self.assertPreciseEqual(result1, func1.py_func())
            self.assertPreciseEqual(result2, func2.py_func())
            self.assertPreciseEqual(result3, func1.py_func())
```
so as to ensure that the test asserts something (suggest the same in the `test_chunksize_with` test below).

numba/tests/test_parfors.py
```diff
@@ -4307,5 +4438,70 @@ def strip_instrs(asm):
                         self.assertEqual(b[op[-2]:op[-1]], 'u')
 
 
+@skip_parfors_unsupported
+class TestParforReductionSetNumThreads(TestCase):
+    """Test exceution correctness on reductions with set_num_threads.
```
stuartarchibald, line 4443:
```suggestion
    """Test execution correctness on reductions with set_num_threads.
```

numba/tests/test_parfors.py
```diff
@@ -4307,5 +4438,70 @@ def strip_instrs(asm):
...
+
+        @njit(parallel=True)
+        def udt(nthreads):
+            acc = 1
+            set_num_threads(nthreads)
+            for i in prange(M):
+                local_mask = 1 + i % mask
+                set_num_threads(local_mask)
+                gnt = get_num_threads()
+                acc *= gnt
```
stuartarchibald, line 4478: For a machine with e.g. 64 cores, this will overflow?

numba/cpython/builtins.py
```diff
@@ -550,8 +550,29 @@ def indval_min(indval1, indval2):
...
+            if indval1.value != indval1.value:
+                # nan check
+                if indval2.value != indval2.value:
```
stuartarchibald, line 556: Should this use `np.isnan`?

stuartarchibald (commented):
Thanks for the update @DrTodd13, I've given the caching tests a review along with looking at the doc changes, there's a few minor things to resolve noted in line else looks good. Many thanks.

numba/cpython/builtins.py
DrTodd13: Up to you.  What I did should work for any kind of nan, not just np.nan but if that is the only kind possible then they are equivalent.

docs/source/user/parallel.rst
```diff
@@ -619,9 +619,19 @@ sized chunks and gives one such chunk to each configured thread.
 (See :ref:`setting_the_number_of_threads`).
 This scheduling approach is equivalent to OpenMP's static schedule with no specified
 chunk size and is appropriate when the work required for each iteration is nearly constant.
+Conversely, if the work required per iteration, as shown in the ``prange`` loop below,
+varies significantly then this static
+scheduling approach can lead to load imbalances and longer execution times.
```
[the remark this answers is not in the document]
DrTodd13: Done.

numba/tests/parfor_cache_usecases.py
```diff
@@ -0,0 +1,41 @@
...
+    def check_module(self, mod):
+        import numpy as np
+        f = mod.f
+        a = np.ones(100)
+        b = np.ones(100)
+        f(a, b)
+
+    # For 2.x
+    def runTest(self):
+        raise NotImplementedError
```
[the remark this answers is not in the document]
DrTodd13: Removed from the 4 files that had them.

numba/tests/test_parfors.py
DrTodd13: Maybe "acc = acc % (2**64)" at the end of each iteration?

numba/tests/doc_examples/test_parallel_chunksize.py
stuartarchibald: I think the same needs applying to `test_chunksize_with` ?

numba/cpython/builtins.py
stuartarchibald: IIRC `np.isnan` does a check for IEEE 754 defined `nan`, so should work for any conformant `nan`.
```pycon
In [1]: import numpy as np

In [2]: np.isnan(float('nan'))
Out[2]: True

In [3]: np.isnan(np.nan)
Out[3]: True

In [4]: qnan = (np.float64(float('nan')).view(np.uint64) | np.uint8(1)).view(np.float64)

In [5]: hex(qnan.view(np.uint64))
Out[5]: '0x7ff8000000000001'

In [6]: np.isnan(qnan)
Out[6]: True
```
my preference towards using `np.nan` is that it means that there's a single point in the code base to catch use of `fastmath` and "is this a nan", which means such an issue can be reported back to users from there.

numba/tests/test_parfors.py
stuartarchibald: Perhaps bound it like `N = min(4, config.NUMBA_NUM_THREADS)`?

stuartarchibald:
@DrTodd13 thanks for the fixes, I've reviewed the code up to f70175b (changes look fine, thanks), and have responded to your questions in line.

numba/tests/test_parfors.py
DrTodd13: Done.

sklam:
The latest changes looks good to me

stuartarchibald (commented):
Thanks for the update @DrTodd13. I just noticed there's a near duplicate cache test file name the needs addressing (see inline comment). I also reviewed all the outstanding queries that got lost in the long review and have commented on those, all are resolved with the exception of https://github.com/numba/numba/pull/7625#discussion_r844410107 which is still of concern. I'm going to give this a run through the build farm now on the basis that public CI will be sufficient to deal with the minor change resulting in merging the cache test files. Thanks again!

stuartarchibald:
@DrTodd13 I've opened https://github.com/IntelLabs/numba/pull/72 to address the duplication of test files, please could you take a look and if you approve merge in, many thanks.

numba/tests/parfor_cache_usecases.py
```diff
@@ -0,0 +1,37 @@
+"""
```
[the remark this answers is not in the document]
stuartarchibald: See https://github.com/IntelLabs/numba/pull/72

stuartarchibald:
RE the outstanding comment from: https://github.com/numba/numba/pull/7625#discussion_r836577063, PR https://github.com/numba/numba/pull/8186 has 991a965 which removes the proposed unification method on the `ExternalFunctionPointer` type and the tests at least pass in public CI. Is this sufficient evidence to suggest it is not needed?

gmarkall:
gpuci run tests (just running this as there are some changes to `numba.core` - I'm pretty sure it won't affect anything negatively in CUDA, but just double-checking here)

numba/core/types/functions.py
```diff
@@ -628,6 +628,10 @@ def generic(self, args, kws):
...
+
```
stuartarchibald, line 634:
```suggestion
```
@DrTodd13 xref: https://github.com/numba/numba/pull/7625#issuecomment-1164307865 suggest removing this.

sklam:
one unresolved comment: https://github.com/numba/numba/pull/7625/files#r904748398

sklam:
A windows test failed:

```

stuartarchibald:
> A windows test failed:
> 
> ```
> 
> FAIL: test_caller (numba.tests.test_parfors_caching.TestParforsCache)
>   File "D:\a\1\s\numba\tests\test_parfors_caching.py", line 45, in test_caller
>     self.run_test(f, num_funcs=3)
>   File "D:\a\1\s\numba\tests\test_parfors_caching.py", line 22, in run_test
>     self.assertPreciseEqual(f(ary), f.py_func(ary))
>   File "D:\a\1\s\numba\tests\support.py", line 390, in assertPreciseEqual
>     self.fail("when comparing %s and %s: %s" % (first, second, failure_msg))
> AssertionError: when comparing [0.1 0.1 0.1 0.1 0.1 0.1 0.1 0.1 0.1 0.1] and [0.1 0.1 0.1 0.1 0.1 0.1 0.1 0.1 0.1 0.1]: 0.1 != 0.10000000000000003
> 
> ```

It's strange that this should suddenly start failing. It looks like a minor numerical error, probably just from using reductions/accumulation in here:
https://github.com/numba/numba/blob/33e94b0773b88b92fa8f59abfe51c246be2d472a/numba/tests/parfors_cache_usecases.py#L9-L25

I think the "fix" is to use `np.testing.assert_allclose` in the `check_module` and `run_module` methods in https://github.com/numba/numba/pull/7625/files#diff-89333c093ac43075778fe1a5bdd16ed10fd6380e4a76dc5e9e8fdb148596f679

stuartarchibald (approved):
Thanks for all your work on this @DrTodd13!

sklam:
smoketesting at BFID `numba_smoketest_cpu_yaml_110`

gmarkall:
:tada:

PR #7632, opened 2021-12-08
Capture output in CUDA matmul doctest

gmarkall:
This prevents matrices being printed when running the testsuite - we want to retain the print statements for the purpose of the documentation example.


gmarkall:
gpuci run tests

gmarkall:
Just realised I have a bug, I think `tearDown()` should call the method from `super()` otherwise the occupancy warnings settings don't get put back.

stuartarchibald (approved):
Thanks for the patch.

stuartarchibald:
Marking as ready to merge. I've manually tested the patch and can confirm it suppresses output in `test_matmul.*`, `gpuci` passed too.

esc:
@gmarkall thank you for the patch! @stuartarchibald thank you for the review and testing!

PR #7636, opened 2021-12-09
Copy prange loop header to after the parfor.

### LINKED ISSUE #7630: Function with two logic paths and two pranges gives erroneous result when compiled with parallel=True
## Reporting a bug

- [ x ] I have tried using the latest released version of Numba (most recent is
- [ x ] I have included a self contained code sample to reproduce the problem.

Minimal reproducing code sample:
```
def proof(flag):
    result = 0
    if flag:
        for i in numba.prange(1000000):
            result += 1
    else:
        for i in numba.prange(1000000):
            result -= 3
    return result
```

I expect this function to return the same result with flag set to False whether compiled with our without parallelism.
Results:
```
Python 3.8.12 (default, Dec  6 2021, 03:13:14) 
[GCC 7.5.0] on linux
>>> import numpy
>>> import numba
>>> numpy.__version__
'1.20.3'
>>> numba.__version__
'0.54.1'
>>> def proof(flag):
...     result = 0
...     if flag:
...         for i in numba.prange(1000000):
...             result += 1
...     else:
...         for i in numba.prange(1000000):
...             result -= 3
...     return result
... 
>>> good = numba.njit(parallel=False)(proof)
>>> bad  = numba.njit(parallel=True)(proof)
>>> good(True)   #correct
1000000
>>> good(False)  #correct
-3000000
>>> bad(True)  # this is ok
1000000
>>> bad(False)  # fails here
0
```

DrTodd13:
Resolves #7630 

Numba can insert statements with side-effects into the prange loop header and expect those to be executed after the last loop iteration for correctness.  We copy the loop header after the parfor here then for correctness.

### DIFF
--- a/numba/parfors/parfor.py
+++ b/numba/parfors/parfor.py
@@ -2568,7 +2568,13 @@ class ConvertLoopPass:
                                     ("prange", loop_kind, loop_replacing),
                                     pass_states.flags, races=races)
 
-                    blocks[loop.header].body = [parfor, ir.Jump(list(loop.exits)[0], loc)]
+                    blocks[loop.header].body = [parfor]
+                    # We have to insert the header_body after the parfor because in
+                    # a Numba loop this will be executed one more time before the
+                    # branch and may contain instructions such as variable renamings
+                    # that are relied upon later.
+                    blocks[loop.header].body.extend(header_body)
+                    blocks[loop.header].body.append(ir.Jump(list(loop.exits)[0], loc))
                     self.rewritten.append(dict(
                         old_loop=loop,
                         new=parfor,
--- a/numba/parfors/parfor.py
+++ b/numba/parfors/parfor.py
@@ -2570,7 +2570,7 @@ class ConvertLoopPass:
 
                     blocks[loop.header].body = [parfor]
                     # We have to insert the header_body after the parfor because in
-                    # a Numba loop this will be executed one more time before the
+                    # a Numba loop this will be executed one more times before the
                     # branch and may contain instructions such as variable renamings
                     # that are relied upon later.
                     blocks[loop.header].body.extend(header_body)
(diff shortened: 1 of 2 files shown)


numba/tests/test_parfors.py
```diff
@@ -3554,6 +3554,21 @@ def test_impl(x, y):
         self.prange_tester(test_impl, X, Y, scheduler_type='unsigned',
                            check_fastmath=True, check_fastmath_result=True)
 
+    def test_prange29(self):
+        # issue7630: SSA renaming in prange header
+        def test_impl(flag):
+            result = 0
+            if flag:
+                for i in range(1000000):
```
stuartarchibald, line 3562:
```suggestion
                for i in range(1):
```
The reproducer doesn't actually need to execute more than one iteration for the bug in #7630 to manifest.

numba/tests/test_parfors.py
```diff
@@ -3554,6 +3554,21 @@ def test_impl(x, y):
...
+                    result += 1
+            else:
+                for i in range(1000000):
```
stuartarchibald, line 3565:
```suggestion
                for i in range(1):
```

numba/parfors/parfor.py
```diff
@@ -2568,7 +2568,13 @@ def find_mask_from_size(size_var):
                                     ("prange", loop_kind, loop_replacing),
                                     pass_states.flags, races=races)
 
-                    blocks[loop.header].body = [parfor, ir.Jump(list(loop.exits)[0], loc)]
+                    blocks[loop.header].body = [parfor]
+                    # We have to insert the header_body after the parfor because in
+                    # a Numba loop this will be executed one more time before the
```
stuartarchibald, line 2573:
```suggestion
                    # a Numba loop this will be executed one more times before the
```

stuartarchibald (commented):
Thanks for the patch, couple of minor things to resolve else looks good. I can confirm this fixes the original issue in #7630.

stuartarchibald (approved):

PR #7637, opened 2021-12-09
Increase the timeout on the SVML tests for loaded machines.

stuartarchibald:
As title. If a machine is under heavy load 10s might not be
enough.


esc (approved):
LGTM

esc:
@stuartarchibald thank you for the patch.

PR #7645, opened 2021-12-10
In debuginfo, do not add noinline to functions marked alwaysinline

### LINKED ISSUE #7642: debuginfo flag attempts to set noinline on functions with alwaysinline
## Reporting a bug

This (and in fact any function implemented via the `@overload_glue` wrappers):
```python
from numba import njit
import numpy as np

@njit
def foo():
    a = np.zeros(1)

print(foo())
```
will not compile if `NUMBA_DEBUGINFO=1` is set. End of the stack trace:
```
  File "<numba_path>/numba/core/lowering.py", line 255, in lower_function_body
    self.pre_lower()
  File "<numba_path>/numba/core/lowering.py", line 133, in pre_lower
    self.debuginfo.mark_subprogram(function=self.builder.function,
  File "<numba_path>/numba/core/debuginfo.py", line 319, in mark_subprogram
    function.attributes.add('noinline')
  File "<llvmlite_path>/llvmlite/ir/values.py", line 887, in add
    raise ValueError("Can't have alwaysinline and noinline")
ValueError: Can't have alwaysinline and noinline
```
this comes from the `overload_glue` functions being forced as `alwaysinline` and the `debuginfo` flag trying to prevent inlining.

stuartarchibald:
As title.

Fixes #7642

### DIFF
--- a/numba/core/debuginfo.py
+++ b/numba/core/debuginfo.py
@@ -315,8 +315,11 @@ class DIBuilder(AbstractDIBuilder):
                                        line=line, function=function,
                                        argmap=argmap)
         function.set_metadata("dbg", di_subp)
-        # disable inlining for this function for easier debugging
-        function.attributes.add('noinline')
+
+        # Don't marked alwaysinline functions as noinline.
+        if 'alwaysinline' not in function.attributes:
+            # disable inlining for this function for easier debugging
+            function.attributes.add('noinline')
 
     def finalize(self):
         dbgcu = cgutils.get_or_insert_named_metadata(self.module, self.DBG_CU_NAME)
--- a/numba/tests/test_debuginfo.py
+++ b/numba/tests/test_debuginfo.py
@@ -2,6 +2,7 @@ from collections import namedtuple
 import inspect
 import re
 import numpy as np
+import math
 
 from numba.tests.support import TestCase, override_config, needs_subprocess
 from numba import jit, njit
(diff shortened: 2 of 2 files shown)


numba/core/debuginfo.py
```diff
@@ -315,8 +315,11 @@ def mark_subprogram(self, function, qualname, argnames, argtypes, line):
                                        line=line, function=function,
                                        argmap=argmap)
         function.set_metadata("dbg", di_subp)
-        # disable inlining for this function for easier debugging
-        function.attributes.add('noinline')
+
+        # Don't marked alwaysinline functions as noinline.
+        if 'alwaysinline' not in function.attributes:
+            # disable inlining for this function for easier debugging
+            function.attributes.add('noinline')
```
sklam, line 322: Why not the other way around? Discard `alwaysinline` is `noinline` is present. The `noinline` is being added for better debugging experience.

stuartarchibald: I left it this way around because it matches up with the "direct lowering" experience that is present in Numba < 0.54, and also will be in 0.55. I guess we can swap it but it might get weird as e.g. potentially trivial things like `operator.add` would end up as a `noinline` function call, this making the IR pretty strange looking?

sklam: hm... i'll need to play around with the other way.

sklam: With everything as `noinline`, I don't find it strange, but retaining the user specified `forceinline` seems more intuitive and give users the control.

With code like:

```python
import math
import numpy as np
from numba import njit

@njit(forceinline=True)
def bar(x):
    return math.sin(x)

@njit(forceinline=False)
def baz(x):
    return math.cos(x)

@njit(forceinline=True)
def array_work(a, b):
    c = a + np.zeros(10) * b
    return np.sum(c)

@njit
def foo(x):
    # gdb()
    a = bar(x)
    b = baz(x)
    return array_work(a, b)

result = foo(np.pi)
foo.inspect_cfg(foo.signatures[0]).display(view=True)
```

If everything is `noinline`, it will show:

Screen Shot 2021-12-10 at 5 34 05 PM

If we use the changes in this PR, it will show:

Screen Shot 2021-12-10 at 5 34 35 PM

If I'm debugging, both versions are useful.

sklam (approved):
LGTM.

As discussed OOB,  keeping `forceinline` in debug mode makes more sense for users.

PR #7639, opened 2021-12-09
Move Azure builds to OSX 10.15.

stuartarchibald:
As title.

PR #7650, opened 2021-12-13
Move Azure builds to OSX 10.15

sklam:
Closes https://github.com/numba/numba/pull/7639

Azure moved minimal OSX support to 10.15, but anaconda compiler needs the MacOSX10.10 SDK to work.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -10,7 +10,7 @@ jobs:
 - template: buildscripts/azure/azure-linux-macos.yml
   parameters:
     name: macOS
-    vmImage: macOS-10.14
+    vmImage: macOS-10.15
     matrix:
       py37_np118:
         PYTHON: '3.7'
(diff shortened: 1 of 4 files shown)


gmarkall (commented):
If I understand correctly, this is intended to be a temporary workaround so that we can start testing properly again and move towards the 0.55RC - is that the case?

If that is the case, then I'd suggest that we should store hash of the SDK we're downloading from Github and check it, so that we detect if the SDK changes or the download is corrupted. Is there a way we can check what an authentic SDK package would hash as? For example, is it possible to install the relevant xcode version on a local machine and extract the SDK to check its hash using `xcodelegacy.sh` from https://github.com/devernay/xcodelegacy ?

If this is not intended to be a quick fix but a permanent resolution, then it certainly feels strange to have to rely on a Github repository that we don't control that's hosting the SDK - maybe this isn't an issue if it's the entire mechanism that the Anaconda compilers / toolchain relies on in general, since we're dependent on them, but I think I'd need to better understand the situation to make a proper comment if this is the case.

esc (dismissed):
You can merge this immediately, because it fixes our Azure CI MacOSX builds. I agree with the concerns of @gmarkall however about downloading a "random blob" and would appreciate those being addressed in a future PR.

chenghlee (approved):

gmarkall (approved):

PR #7660, opened 2021-12-15
Add support for np.broadcast_arrays

guilhermeleobas:
As title


guilhermeleobas:
PR is ready for review. @gmarkall, can you please change the label?

numba/np/arrayobj.py
```diff
@@ -1382,6 +1384,49 @@ def impl(array, shape):
...
+    def impl(*args):
+        # find out the output shape
+        # we can't call np.broadcast_shapes here since args may have arrays
+        # with different shapes and it is not possible to create a list
+        # with those shapes dinamically
+        shape = [1] * m
+        for array in literal_unroll(args):
+            numpy_broadcast_shapes_list(shape, m, array.shape)
+
+        tup = build_int_tuple(m)
```
stuartarchibald, line 1416: I think the tuple size is based on a compile time constant, such that a change like:
```diff
diff --git a/numba/np/arrayobj.py b/numba/np/arrayobj.py
index da81005..42d8ef5 100644
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -1404,6 +1404,8 @@ def numpy_broadcast_arrays(*args):
     # number of dimensions
     m = max([arr.ndim for arr in args])
 
+    tup_init = (0,) * m
+
     def impl(*args):
         # find out the output shape
         # we can't call np.broadcast_shapes here since args may have arrays
@@ -1413,8 +1415,7 @@ def numpy_broadcast_arrays(*args):
         for array in literal_unroll(args):
             numpy_broadcast_shapes_list(shape, m, array.shape)
 
-        tup = build_int_tuple(m)
-
+        tup = tup_init
         for i in range(m):
             tup = tuple_setitem(tup, i, shape[i])
```
would make it possible to remove the `build_int_tuple` function entirely.

guilhermeleobas:
@stuartarchibald done!

numba/tests/test_array_manipulation.py
```diff
@@ -762,6 +766,118 @@ def _assert_raises(arr, val):
         val = np.array([-1e100])
         _assert_raises(arr, val)
 
+    def test_shape(self):
```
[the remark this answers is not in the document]
guilhermeleobas: done!

numba/np/arrayobj.py
guilhermeleobas: thanks for the tip

guilhermeleobas:
@stuartarchibald, when you have some time, can you update the label in this PR?

numba/np/arrayobj.py
```diff
@@ -1382,6 +1382,50 @@ def impl(array, shape):
...
+            r[k] = tmp
+        elif r[k] != tmp:
+            raise ValueError("shape mismatch: objects"
+                             " cannot be broadcast"
+                             " to a single shape")
+
+
+@overload(np.broadcast_arrays)
+def numpy_broadcast_arrays(*args):
+    # number of dimensions
```
stuartarchibald, line 1402: I think this implementation needs a couple of things:
1. Some type checking to ensure that `*args` contains things that are suitable for use in the implementation.
2. Decide how much of what is accepted in NumPy's `broadcast_arrays` should be accepted by Numba. For example, the np impl accepts pretty much anything that `np.array` will accept. Should Numba do similar?

stuartarchibald (commented):
Thanks for the patch. I've given this an initial review to provide some feedback and have left this inline. Once addressed I'd like to take another look to see if there's any corner cases that are potentially missing as the tests equivalent to those from NumPy seem to cover a lot of things but there may be some Numba specific issues. Thanks again!

numba/np/arrayobj.py
guilhermeleobas: I'll restrict the implementation to what `type_can_asarray` accepts

stuartarchibald: Sounds good, thanks.

numba/np/arrayobj.py
```diff
@@ -1382,6 +1382,64 @@ def impl(array, shape):
...
+        if not type_can_asarray(arg):
+            raise errors.TypingError(f'Argument "{idx}" must '
+                                     'be array-like')
+
+    # number of dimensions
+    m = 0
+    for idx, arg in enumerate(args):
+        if isinstance(arg, types.ArrayCompatible):
+            m = max(m, arg.ndim)
+        elif isinstance(arg, (types.Number, types.StringLiteral,
```
stuartarchibald, line 1413: `StringLiteral`? Do you know of a usecase for this? e.g.
```python
np.broadcast_arrays(np.arange(4), 'oops')
```
is invalid?

numba/np/arrayobj.py
```diff
@@ -1382,6 +1382,64 @@ def impl(array, shape):
...
+                              types.Boolean)):
```
stuartarchibald, line 1414: Bool inputs are not tested? This is failing for me:
```python
from numba import njit
import numpy as np

@njit
def foo(x,):
    return np.broadcast_arrays(x, True)

b = np.arange(6).reshape((2, 3))
print(foo(b,))
```
the internal list is typed for int array but a bool array is being appended. This raises the question... does this implementation only work if the inputs are homogeneous in dtype? I think this might be the case, e.g.
```python
from numba import njit
import numpy as np

@njit
def foo(x, y):
    return np.broadcast_arrays(x, y)

a = np.arange(3, dtype=np.complex128)
b = np.arange(6).reshape((2, 3))
print(foo(a, b))
```
to which effect this probably ought to be caught at typing time or fixed in the implementation?

stuartarchibald (commented):
Thanks for the updates @guilhermeleobas. I've made a couple more remarks in relation to what is supported/not supported, please could you take a look? Many thanks.

numba/np/arrayobj.py
guilhermeleobas: To be honest, I don't know why I put `StringLiteral` here 😄

numba/np/arrayobj.py
guilhermeleobas: The code above crashes because the current implementation appends each intermediary result to a list. Is there any way one could append different types to the list (or tuple) without unifying them?

stuartarchibald: I think invariant of the implementation, the issue is that the return type of the NumPy function is a list of lists with no guarantee that the nested list(s) are of the same type? Numba cannot construct such a list dynamically, the list must be homogeneous in type. Perhaps we could discuss this at the developer's meeting or similar, I don't think there's a work around at present.

guilhermeleobas: Could this be done using another data structure? I know one can return a heterogeneous tuple from a jitted function.

stuartarchibald: Technically I think it could be done, but it would break the semantics of the NumPy function as the return type would not match that of NumPy. I'm not sure that there's any reason for this to be a list other than the list comprehension that generates it is idiomatic:
https://github.com/numpy/numpy/blob/73e48ae177c4b29cdbdbcfd39e22cf5de6c7f125/numpy/lib/stride_tricks.py#L546-L547

guilhermeleobas: Got it! If breaking the API is not a problem, I think I can do it with tuples.

stuartarchibald: Unfortunately, for Numba, breaking the NumPy API is a problem. Perhaps raise an issue/make a suggestion on NumPy's issue tracker about it returning a tuple?

guilhermeleobas: I don't think this will change in the near future. The array-api spec also says that `broadcast_arrays` return type is a `list[array]`:
https://data-apis.org/array-api/latest/API_specification/generated/signatures.data_type_functions.broadcast_arrays.html?highlight=broadcast_arrays

stuartarchibald: RE data-api spec, I suppose the question is: was this explicitly designed or was it inherited from/informed by NumPy semantics? Seems like the result is a fixed size that is derived solely from number of args so maybe a tuple would be a better fit? It certainly helps compilers with respect to typing/type stability if fixed sized containers are used where possible.

For this PR. I agree, this can be implemented with the homogeneous type restriction.

guilhermeleobas: > For this PR. I agree, this can be implemented with the homogeneous type restriction.

Do you have any more comments about the PR?

stuartarchibald: Once this is fixed I'll retest it, main thing is to ensure that only types/combinations of types that are supported are permitted through and that the result is correct in value and type! More testing/checking in this area is welcomed. Thanks.

guilhermeleobas: The issue with the example you posted is because you're mixing different types. The following works

```python
from numba import njit
import numpy as np

@njit
def foo(x,):
    return np.broadcast_arrays(x, False)

b = np.full(6, fill_value=True).reshape((2, 3))
print(foo(b,))
```

stuartarchibald: Apologies, I wasn't very clear. Fixing this means ensuring that type guards are present to prevent the code running with the unsupported "mixing different types".

numba/np/arrayobj.py
```diff
@@ -1382,6 +1382,77 @@ def impl(array, shape):
...
+                                     'broadcast arrays with different types. '
+                                     f'Got {args}')
+
+    # number of dimensions
+    m = 0
+    for idx, arg in enumerate(args):
+        if isinstance(arg, types.ArrayCompatible):
+            m = max(m, arg.ndim)
+        elif isinstance(arg, (types.Number, types.Boolean)):
+            m = max(m, 1)
```
stuartarchibald, line 1429: This unfortunately doesn't work:
```python
from numba import njit
import numpy as np

@njit
def foo(tupl):
    return np.broadcast_arrays(tupl, tupl)

print(foo((1,2)))
```
I think the issue is that this loop is not handling tuple types? Probably needs something like:
```python
        elif isinstance(arg, types.BaseTuple):
            m = max(m, arg.count)
```
?

stuartarchibald (commented):
Thanks for the update. There seems to be a remaining issue with handling tuples but otherwise this looks good.

numba/np/arrayobj.py
guilhermeleobas: Fixed

guilhermeleobas:
Failure is not related to the changes introduced in this PR.

numba/tests/test_array_manipulation.py
```diff
@@ -965,6 +969,184 @@ def test_argwhere_array_like(self):
...
+            [(1, 3, 4), (2, 3, 3)],
+        ]
+        for input_shapes in data:
+            for shape in [input_shapes, input_shapes[::-1]]:
+                # Reverse the input shapes since broadcasting should be symmetric.
+                with self.assertRaises(ValueError) as raises:
+                    inarrays = [np.zeros(s) for s in shape]
+                    cfunc(*inarrays)
+                self.assertIn("shape mismatch: objects cannot be broadcast to a single shape",
+                            str(raises.exception))
```
stuartarchibald, line 1147:
```suggestion
                              str(raises.exception))
```

stuartarchibald (commented):
Thanks for the update. One minor style alignment issue to fix, then will mark for merge.

guilhermeleobas:
@stuartarchibald done!

PR #7664, opened 2021-12-16
Flatten mangling dicts into a single dict

gmarkall:
Since we are attempting to mangle Numba types, mapping straight from Numba types to type codes is more direct than going through C type names, and avoids the issue of dealing with adding C type names for which there is no standard name (e.g. 16-bit floats).

See https://github.com/numba/numba/pull/7460#discussion_r768668206 for a discussion that spurred this.


gmarkall:
gpuci run tests

stuartarchibald (commented):
Thanks for the patch. I think this flattening makes sense, it removes a seemingly unnecessary layer of indirection. I'm not sure if there's any use case where the C map would be needed, am going to wait for a response from @sklam in case there's some rare case that I've missed (maybe external C interop?).

sklam (dismissed):
LGTM.

`C2CODE` is probably from legacy code. We don't need it anymore.

sklam:
There's conflict with main

stuartarchibald:
> LGTM.
> 
> `C2CODE` is probably from legacy code. We don't need it anymore.

Thanks for confirming @sklam, it's from around 7 years ago https://github.com/numba/numba/commit/a257cb1a26ef3dd7f69f5160f97abf37960bb1ea, and seems entirely unused!

stuartarchibald (approved):

stuartarchibald:
@gmarkall this is approved but needs a conflict resolving. Please could you take a look when you have a moment? Many thanks.

gmarkall:
@stuartarchibald Many thanks for the review. The conflict should now be resolved.

stuartarchibald (approved):
Thanks for fixing the conflict against mainline.

Issue #7679, opened 2021-12-20
Numba CUDA Matrix Multiplication Example


susuter:
## Reporting a bug

- [] I have tried using the latest released version of Numba (most recent is
 visible in the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [y] I have included a self contained code sample to reproduce the problem.
  i.e. it's possible to run as 'python bug.py'.

When executing the example https://numba.pydata.org/numba-doc/dev/cuda/examples.html, the resulting matrix is not filled entirely - only first column. The grid index j is always 0. I was trying it on different computing clusters with jupyter notebooks, e.g., Colab, where numba is of version: 0.54.1. I was not able to run it with the latest numba 0.55.0rc1, but the example seems the same for varying versions. On another cluster, I had the same problems with numba 0.53.1. Input were np arrays of float32 - varying shapes.

See Colab Notebook: https://colab.research.google.com/drive/1uewkzBaGEcrr9CCPeyR0YtJus9I3Nvwf

Would you check that? Do you have any idea on that?

Thanks in advance!

gmarkall:
The documentation linked is for an old version (unfortunately Google tends to prioritise it in search results) - the current documentation, which has undergone some fixes to the matrix multiplication example, is at: https://numba.readthedocs.io/en/latest/cuda/examples.html

Could you try the example from the current version of the docs with and let me know if you have any further issues with it please?

susuter:
Thank you for your fast reply. I checked the example on the link. The code is actually the same and I experience the same problem. I saw there is a sample program call on this page. Hence, I tried it with that sample program call. I get the same problem. It really surprises me that the second grid-index is always 0. Would you look into that?

gmarkall:
Many thanks for following up!

> I checked the example on the link. The code is actually the same and I experience the same problem.

I'm seeing different code on https://numba.pydata.org/numba-doc/dev/cuda/examples.html (the outdated version) and https://numba.readthedocs.io/en/latest/cuda/examples.html (the latest version) - the `matmul` kernel looks the same but otherwise the examples are different and there are no examples of launching the kernel on the old documentation. 

> I saw there is a sample program call on this page. Hence, I tried it with that sample program call.

Can you indicate where the sample program call is please (if it's somewhere in the latest docs)? Maybe we have something else in the docs that needs fixing.

> See Colab Notebook: https://colab.research.google.com/drive/1uewkzBaGEcrr9CCPeyR0YtJus9I3Nvwf

[...]

PR #7680, opened 2021-12-21
CUDA Docs: include example calling slow matmul

gmarkall:
The CUDA matrix multiplcation example lacked an example of how to call the naive matrix multiplication kernel (there was only an example of a call of the shared-memory optimized version - identified in https://github.com/numba/numba/issues/7679#issuecomment-998830327). This is a little unhelpful for someone encountering the example for the first time. This PR adds an example calling the naive matrix multiplication kernel prior to discussing optimizations.

### DIFF
--- a/docs/source/cuda/examples.rst
+++ b/docs/source/cuda/examples.rst
@@ -27,6 +27,15 @@ Here is a naïve implementation of matrix multiplication using a CUDA kernel:
    :dedent: 8
    :linenos:
 
+An example usage of this function is as follows:
+
+.. literalinclude:: ../../../numba/cuda/tests/doc_examples/test_matmul.py
+   :language: python
+   :caption: from ``test_ex_matmul`` in ``numba/cuda/tests/doc_examples/test_matmul.py``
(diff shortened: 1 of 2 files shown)


gmarkall:
Something is not quite right with Python 3.10 unrelated to this PR:

```
CondaVerificationError: The package for python located at /home/vsts/miniconda3/pkgs/python-3.10.0-h12debd9_4
appears to be corrupted. The path 'lib/python3.1'
specified in the package manifest cannot be found.
```

Hopefully this is just a transient issue.

seibert:
This is due a conda / conda-build bug that was fixed, but I'm not sure if we need to update conda in the build process or if Anaconda needs to rebuild some packages.

gmarkall:
(cc @susuter)

gmarkall:
gpuci run tests

stuartarchibald (approved):
Thanks for the patch.

stuartarchibald:
Once the build passes this can be merged (failures are unrelated).

PR #7682, opened 2021-12-21
performance improvements to np.full and np.ones

### LINKED ISSUE #7665: Poor np.full performace with jit
numba version: 0.54.1

This is the testing code.
```python
import numpy as np
from numba import jit, prange, set_num_threads
import time

set_num_threads(4)

@jit
def f1(shape):
    return np.full(shape, np.nan, np.float64)

@jit
def f2(shape):
    ans = np.empty(shape, np.float64)
    for i in range(ans.shape[0]):
        for j in range(ans.shape[1]):
            ans[i, j] = np.nan

@jit(parallel=True)
def f3(shape):
    ans = np.empty(shape, np.float64)
    for i in prange(ans.shape[0]):
        for j in range(ans.shape[1]):
            ans[i, j] = np.nan

# Warm up
shape = (12, 10)
f1(shape)
f2(shape)
f3(shape)

shape = (128000, 1000)

t0 = time.time()
np.full(shape, np.nan, np.float64)
print(time.time() - t0)

t0 = time.time()
f1(shape)
print(time.time() - t0)

t0 = time.time()
f2(shape)
print(time.time() - t0)

t0 = time.time()
f3(shape)
print(time.time() - t0)
```
This is the output
```
0.24103260040283203 # numpy
0.5575239658355713 # np.full in jit
0.45681166648864746 # fill with for-loop
0.18029499053955078 # fill with parallel for-loop
```
The speed of `np.full` in jit is less than half compared with the normal one from my testing.

rishi-kulkarni:
Closes #7665. np.full and np.ones use the slow np.ndindex iterator, which results in both functions being about twice as slow as their numpy counterparts. Iterating over the flat array instead makes the functions as fast as their numpy versions.

### DIFF
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -3789,8 +3789,9 @@ def numpy_full_nd(context, builder, sig, args):
 
     def full(shape, value):
         arr = np.empty(shape, type(value))
-        for idx in np.ndindex(arr.shape):
-            arr[idx] = value
+        arr_flat = arr.flat
+        for idx, _ in enumerate(arr_flat):
+            arr_flat[idx] = value
         return arr
 
     res = context.compile_internal(builder, full, sig, args)
@@ -3803,8 +3804,9 @@ def numpy_full_dtype_nd(context, builder, sig, args):
 
     def full(shape, value, dtype):
         arr = np.empty(shape, dtype)
-        for idx in np.ndindex(arr.shape):
-            arr[idx] = value
+        arr_flat = arr.flat
+        for idx, _ in enumerate(arr_flat):
+            arr_flat[idx] = value
         return arr
 
     res = context.compile_internal(builder, full, sig, args)
@@ -3816,8 +3818,9 @@ def numpy_full_like_nd(context, builder, sig, args):
 
     def full_like(arr, value):
         arr = np.empty_like(arr)
-        for idx in np.ndindex(arr.shape):
-            arr[idx] = value
+        arr_flat = arr.flat
+        for idx, _ in enumerate(arr_flat):
+            arr_flat[idx] = value
         return arr
 
     res = context.compile_internal(builder, full_like, sig, args)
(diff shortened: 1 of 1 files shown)


gmarkall:
Many thanks for the PR! I'll take a look at this on Wednesday when I'm back at work.

esc:
It will probably be of relevance to take the following into account when reviewing this issue:

https://github.com/numba/numba/issues/7753

esc:
@rishi-kulkarni do you have any benchmarks to illustrate the performance gains here?

gmarkall (dismissed):
This looks good to me. Whilst fixing #7753 might also fix this issue, this change doesn't appear to add significant complexity and fixes #7665 right now, so I think it would be OK to merge.

Note that we still have a big difference between Numba and NumPy for `zeros`, because NumPy uses `calloc`:

```python
import numpy as np
from numba import jit


@jit
def f0(shape):
    return np.zeros(shape, dtype=np.float64)


shape = (128000, 1000)

f0(shape)

%timeit f0(shape)

%timeit np.zeros(shape, dtype=np.float64)
```

produces:

```
401 ms ± 119 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
8.33 µs ± 40.6 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

(and is unaffected by this branch).

numba/np/arrayobj.py
```diff
@@ -3803,8 +3804,9 @@ def numpy_full_dtype_nd(context, builder, sig, args):
 
     def full(shape, value, dtype):
         arr = np.empty(shape, dtype)
-        for idx in np.ndindex(arr.shape):
-            arr[idx] = value
+        arr_flat = arr.flat
+        for idx, _ in enumerate(arr_flat):
```
stuartarchibald, line 3808: Suggest using `for idx in range(len(arr_flat))` (or similar)? There's no need to set up the more complex loop structure?

rishi-kulkarni: Sure, done. Performance is identical.

stuartarchibald (approved):

PR #5209, opened 2020-02-06
Use importlib to load numba extensions

svrakitin:
Closes #4927

PR #7687, opened 2021-12-23
- remove setuptools dependency

franzhaas:
similar to https://github.com/numba/llvmlite/pull/801#issuecomment-1000445283, this is an effort to remove usage / dependency of setuptools.

The motivation is that it makes it easier to use tools like pyoxidizer, replaces a library dependency with a builtin depenency, decreases import latency and reduces overall dependancies.

also like the mentioned pull request, not a killer feauture not a bug fix, but it is expected to speedup import times.

At some point i noticed the PR [fix/4927-slow-import](https://github.com/numba/numba/pull/5209) and used the developoments from there.

This is not a successfull simplification. A dependency of a backport and two implementations how to find the entrypoints have been added. On  the other side the dependancy of setuptools and 1 implementation got removed.

If this yields improvements on importtime, it propably is still be worth it, especially for python 3.8 and newer. Also for me personaly droping setuptools as a requirement is a big plus as packages tend to use it if it is there, which ocationally causes issues...

I used input from: https://github.com/python/importlib_metadata/issues/298
And again, the original PR from @svrakitin.

These code changes would need to be touched when suport for 3.7 and 3.8 gets removed.

Unfortunately I was unable to test it with real extensions.

### DIFF
--- a/buildscripts/condarecipe.local/meta.yaml
+++ b/buildscripts/condarecipe.local/meta.yaml
@@ -36,7 +36,7 @@ requirements:
     - setuptools
     # On channel https://anaconda.org/numba/
     - llvmlite >=0.38.0dev0,<0.38
-    - importlib_metadata # [py<39]
+    - importlib_metadata # [py<38]
     # TBB devel version is to match TBB libs.
     # 2020.3 is the last version with the "old" ABI
     # NOTE: ppc64le exclusion is temporary until packages are more generally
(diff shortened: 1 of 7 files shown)


gmarkall:
Many thanks for the PR! I've queued it for review.

I do need to go back and get #5209 over the line - it got a bit fiddly so I didn't yet find a good moment to get everything as it should be.

franzhaas:
Sry for the late reply.

I did not check, but this would be extremely unlikely as this is a derived work of the mentioned pull request.

stuartarchibald:
@franzhaas thanks for checking back and electing to close. #5209 should gain the same performance improvement, please do check 0.56 when it's released and report any issues in relation to this. Thanks again!

PR #7688, opened 2021-12-24
Fix/4927 slow import

franzhaas:
this is a experiment to look if the original pull request can be improved by using backports.entry-points-selectable


franzhaas:
- did go nowhere...

PR #7689, opened 2021-12-24
Fix/4927 slow import take2

franzhaas:
hopefully this tame with the new name the ci pipeline starts to run...


franzhaas:
the name was not the issue... i cherry picked into main instead...

Issue #1335, opened 2015-07-29
Improving support for numpy.random


xiaoweiz:
I've noticed that there are four distributions in numpy.random that are not supported by Numba yet, i.e. 

```python
multinomial
multivariate_normal
noncentral_chisquare
noncentral_f
```

Is there any particular difficulty?

seibert:
This can certainly be done (and you can often write the implementations in Python rather than use the LLVM builder syntax).  Here is where the implementation of the beta distribution is defined, as an example:

https://github.com/numba/numba/blob/master/numba/targets/randomimpl.py#L557-L579

New functions also need to be registered with the front end, as seen here:

https://github.com/numba/numba/blob/master/numba/typing/randomdecl.py#L105

seibert:
I haven't looked at the implementation of these distributions in numpy, though.  They might depend on some external data or functions that are harder to put into Numba.  If you do attempt one of these functions, please let us know if you run into trouble.

pitrou:
Some of those distributions return vectors, which wasn't possible in nopython mode when the random module was first implemented. We would actually have to add the possibility of return random arrays to all existing methods.

xiaoweiz:
@seibert 
The pseudo-code for implementing  `noncentral_chisquare` and `noncentral_f` is quite simple. I gave one at numpy/numpy#5766 for improving `noncentral_chisquare`. According to 
[https://github.com/numpy/numpy/blob/master/numpy/random/mtrand/distributions.c#L256-L260](https://github.com/numpy/numpy/blob/master/numpy/random/mtrand/distributions.c#L256-L260) 
a `noncentral-f` random variate with parameters `dfnum`, `dfden`, and `nonc` can be generated by 
```python
t = np.random.noncentral_chisquare(dfnum, nonc) * dfden 
t / (np.random.chisquare(dfden) * dfnum)
```

My python skill is very limited (otherwise I would have submitted a PR). Unless there's something more subtle, is it possible to add support for these two methods in Numba? 

Btw, an important application of noncentral chisquare distribution is in financial engineering, where it is used to simulate sample paths of the popular [Cox-Ross-Ingersoll model](https://en.wikipedia.org/wiki/Cox%E2%80%93Ingersoll%E2%80%93Ross_model).

pitrou:
multinomial is now supported as of PR #1758.

spring-haru:
`multivariate_normal`? 
Any workaround?

esc:
[...]

PR #7690, opened 2021-12-24
Implemented np.random.noncentral_chisquare for all size arguments

rishi-kulkarni:
This is an implementation of np.random.noncentral_chisquare that accepts size arguments. This implementation uses the same algorithm found in numpy's distributions.c. Part of #1335 - I have plans to knock out the other two functions shortly.

### DIFF
--- a/numba/cpython/randomimpl.py
+++ b/numba/cpython/randomimpl.py
@@ -1586,3 +1586,58 @@ def dirichlet(alpha, size=None):
         )
 
     return dirichlet_impl
+
+# ------------------------------------------------------------------------
+# np.random.noncentral_chisquare
+
+@overload(np.random.noncentral_chisquare)
+def noncentral_chisquare(df, nonc, size=None):
+
+    @register_jitable
+    def noncentral_chisquare_single(df, nonc):
+
+        if df <= 0:
+            raise ValueError("df <= 0")
+        if nonc < 0:
+            raise ValueError("nonc < 0")
+
+        if np.isnan(nonc):
+            return np.nan
+
+        if df > 1:
+            chi2 = np.random.chisquare(df-1)
+            n = np.random.standard_normal() + np.sqrt(nonc)
+            return chi2 + n * n
+        else:
+            i = np.random.poisson(nonc/2.0)
+            return np.random.chisquare(df + 2 * i)
+
+    if size in (None, types.none):
+        def noncentral_chisquare_impl(df, nonc, size=None):
+            return noncentral_chisquare_single(df, nonc)
+
+    elif isinstance(size, types.Integer):
+
+        def noncentral_chisquare_impl(df, nonc, size=None):
+            out = np.empty(size)
+            out_flat = out.flat
+            for idx, _ in enumerate(out_flat):
+                out_flat[idx] = noncentral_chisquare_single(df, nonc)
+            return out
+
+    elif isinstance(size, (types.UniTuple)) and isinstance(size.dtype, types.Integer):
+
+        def noncentral_chisquare_impl(df, nonc, size=None):
+            out = np.empty(size)
+            out_flat = out.flat
+            for idx, _ in enumerate(out_flat):
+                out_flat[idx] = noncentral_chisquare_single(df, nonc)
+            return out
+
+    else:
+        raise NumbaTypeError(
+            "np.random.noncentral_chisquare(): size should be int or "
+            "tuple of ints or None, got %s" % size
+        )
+
+    return noncentral_chisquare_impl
--- a/numba/cpython/randomimpl.py
+++ b/numba/cpython/randomimpl.py
@@ -1594,15 +1594,17 @@ def dirichlet(alpha, size=None):
 def noncentral_chisquare(df, nonc, size=None):
 
     @register_jitable
-    def noncentral_chisquare_single(df, nonc):
-        # identical to numpy implementation from distributions.c
-        # https://github.com/numpy/numpy/blob/c65bc212ec1987caefba0ea7efe6a55803318de9/numpy/random/src/distributions/distributions.c#L797
-        
+    def validate_input(df, nonc):
         if df <= 0:
             raise ValueError("df <= 0")
         if nonc < 0:
             raise ValueError("nonc < 0")
 
+    @register_jitable
+    def noncentral_chisquare_single(df, nonc):
+        # identical to numpy implementation from distributions.c
+        # https://github.com/numpy/numpy/blob/c65bc212ec1987caefba0ea7efe6a55803318de9/numpy/random/src/distributions/distributions.c#L797
+        
         if np.isnan(nonc):
             return np.nan
 
@@ -1617,23 +1619,18 @@ def noncentral_chisquare(df, nonc, size=None):
 
     if size in (None, types.none):
         def noncentral_chisquare_impl(df, nonc, size=None):
+            validate_input(df, nonc)
             return noncentral_chisquare_single(df, nonc)
 
-    elif isinstance(size, types.Integer):
+    elif isinstance(size, types.Integer) or (
+        (isinstance(size, (types.UniTuple)) and isinstance(size.dtype, types.Integer))
+        ):
 
         def noncentral_chisquare_impl(df, nonc, size=None):
+            validate_input(df, nonc)
             out = np.empty(size)
             out_flat = out.flat
-            for idx, _ in enumerate(out_flat):
-                out_flat[idx] = noncentral_chisquare_single(df, nonc)
-            return out
-
-    elif isinstance(size, (types.UniTuple)) and isinstance(size.dtype, types.Integer):
-
-        def noncentral_chisquare_impl(df, nonc, size=None):
-            out = np.empty(size)
-            out_flat = out.flat
-            for idx, _ in enumerate(out_flat):
+            for idx in range(out.size):
                 out_flat[idx] = noncentral_chisquare_single(df, nonc)
             return out
 
@@ -1644,3 +1641,4 @@ def noncentral_chisquare(df, nonc, size=None):
         )
 
     return noncentral_chisquare_impl
+
--- a/numba/cpython/randomimpl.py
+++ b/numba/cpython/randomimpl.py
@@ -1595,7 +1595,8 @@ def noncentral_chisquare(df, nonc, size=None):
 
     @register_jitable
     def noncentral_chisquare_single(df, nonc):
-
+        # identical to numpy implementation 
+        
         if df <= 0:
             raise ValueError("df <= 0")
         if nonc < 0:
@@ -1608,6 +1609,7 @@ def noncentral_chisquare(df, nonc, size=None):
             chi2 = np.random.chisquare(df-1)
             n = np.random.standard_normal() + np.sqrt(nonc)
             return chi2 + n * n
+
         else:
             i = np.random.poisson(nonc/2.0)
             return np.random.chisquare(df + 2 * i)
--- a/numba/cpython/randomimpl.py
+++ b/numba/cpython/randomimpl.py
@@ -1596,6 +1596,7 @@ def noncentral_chisquare(df, nonc, size=None):
     @register_jitable
     def noncentral_chisquare_single(df, nonc):
         # identical to numpy implementation from distributions.c
+        # https://github.com/numpy/numpy/blob/c65bc212ec1987caefba0ea7efe6a55803318de9/numpy/random/src/distributions/distributions.c
         
         if df <= 0:
             raise ValueError("df <= 0")
--- a/numba/cpython/randomimpl.py
+++ b/numba/cpython/randomimpl.py
@@ -1596,7 +1596,7 @@ def noncentral_chisquare(df, nonc, size=None):
     @register_jitable
     def noncentral_chisquare_single(df, nonc):
         # identical to numpy implementation from distributions.c
-        # https://github.com/numpy/numpy/blob/c65bc212ec1987caefba0ea7efe6a55803318de9/numpy/random/src/distributions/distributions.c
+        # https://github.com/numpy/numpy/blob/c65bc212ec1987caefba0ea7efe6a55803318de9/numpy/random/src/distributions/distributions.c#L797
         
         if df <= 0:
             raise ValueError("df <= 0")
--- a/numba/cpython/randomimpl.py
+++ b/numba/cpython/randomimpl.py
@@ -1595,7 +1595,7 @@ def noncentral_chisquare(df, nonc, size=None):
 
     @register_jitable
     def noncentral_chisquare_single(df, nonc):
-        # identical to numpy implementation 
+        # identical to numpy implementation from distributions.c
         
         if df <= 0:
             raise ValueError("df <= 0")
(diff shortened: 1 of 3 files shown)


gmarkall:
Many thanks for the PR - I've queued it for review (though I have given it a quick skim and I see that the changes look generally sane, that the docs are updated and that there are tests to go with the new functionality).

numba/cpython/randomimpl.py
```diff
@@ -1586,3 +1586,60 @@ def dirichlet_impl(alpha, size=None):
...
+
+# ------------------------------------------------------------------------
+# np.random.noncentral_chisquare
+
+@overload(np.random.noncentral_chisquare)
+def noncentral_chisquare(df, nonc, size=None):
+
+    @register_jitable
+    def noncentral_chisquare_single(df, nonc):
+        # identical to numpy implementation
```
esc, line 1598: Am I correct in assuming that the NumPy implementation that this matches is:

https://github.com/numpy/numpy/blob/623bc1fae1d47df24e7f1e29321d0c0ba2771ce0/numpy/random/src/legacy/legacy-distributions.c#L119-L139

esc: If this is the case, it will be great if you can add an appropriate link to the comments.

numba/cpython/randomimpl.py
```diff
@@ -1586,3 +1586,60 @@ def dirichlet_impl(alpha, size=None):
...
+        
+        if df <= 0:
+            raise ValueError("df <= 0")
+        if nonc < 0:
+            raise ValueError("nonc < 0")
+
+        if np.isnan(nonc):
+            return np.nan
+
+        if df > 1:
```
esc, line 1608: This condition appears reversed in the original NumPy implementation:

https://github.com/numpy/numpy/blob/623bc1fae1d47df24e7f1e29321d0c0ba2771ce0/numpy/random/src/legacy/legacy-distributions.c#L125

Or did I miss something here?

esc (changes requested):
I have given this an initial review and would like to make sure we are considering the same original NumPy implementation.

numba/cpython/randomimpl.py
rishi-kulkarni: It's actually from:
https://github.com/numpy/numpy/blob/623bc1fae1d47df24e7f1e29321d0c0ba2771ce0/numpy/random/src/distributions/distributions.c
However, it seems that you're right that the default np.random calls legacy-distributions.c, so I'll update my implementation to reflect that.

numba/cpython/randomimpl.py
rishi-kulkarni: See above.

numba/cpython/randomimpl.py
esc: OK;  but if legacy is deprecated, maybe the better choice would be to move forwards? @stuartarchibald what do you think?

stuartarchibald: IIRC it's legacy because of the new style RNG classes in NumPy, it still works fine. Suggest going for as close as possible to the NumPy impl as it makes debug/maintenance easier long term.

rishi-kulkarni: After playing with it some more, I think the implementation from distributions.c is safer. For reference:

```c
double legacy_noncentral_chisquare(aug_bitgen_t *aug_state, double df,
                                   double nonc) {
  double out;
  if (nonc == 0) {
    return legacy_chisquare(aug_state, df);
  }
  if (1 < df) {
    const double Chi2 = legacy_chisquare(aug_state, df - 1);
    const double n = legacy_gauss(aug_state) + sqrt(nonc);
    return Chi2 + n * n;
  } else {
    const long i = random_poisson(aug_state->bit_generator, nonc / 2.0);
    out = legacy_chisquare(aug_state, df + 2 * i);
    /* Insert nan guard here to avoid changing the stream */
    if (npy_isnan(nonc)){
      return NPY_NAN;
    } else {
    return out;
    }
  }
}
```

Putting the np.nan return there doesn't play well with numba. Otherwise, however, the implementations are the same. Nonetheless, I added a comment referencing distributions.c.

numba/cpython/randomimpl.py
```diff
@@ -1595,7 +1595,7 @@ def noncentral_chisquare(df, nonc, size=None):
 
     @register_jitable
     def noncentral_chisquare_single(df, nonc):
-        # identical to numpy implementation 
+        # identical to numpy implementation from distributions.c
```
esc, line 1598: Thank you for making this update. I think it would be useful to provide a full link to a stable github URL. I.e. one with line-numbers and a SHA in the URL. Thank you!

rishi-kulkarni: Yep, not a problem. I've added it in.

esc:
@rishi-kulkarni apologies for dropping the ball on this I was busy with other things. I have now schedule an hour on Monday to complete this review.

numba/cpython/randomimpl.py
```diff
@@ -1586,3 +1586,61 @@ def dirichlet_impl(alpha, size=None):
...
+    if size in (None, types.none):
+        def noncentral_chisquare_impl(df, nonc, size=None):
+            return noncentral_chisquare_single(df, nonc)
+
+    elif isinstance(size, types.Integer):
+
+        def noncentral_chisquare_impl(df, nonc, size=None):
+            out = np.empty(size)
+            out_flat = out.flat
+            for idx, _ in enumerate(out_flat):
```
esc, line 1627: I have two questions about this loop:

a) Since we only need the indices anyway, can we use `for i in range(size)`
b) It look as though `out_flat` will receive identical values? Is this correct? Can we try to use broadcasting for the assignment?

numba/cpython/randomimpl.py
```diff
@@ -1586,3 +1586,61 @@ def dirichlet_impl(alpha, size=None):
...
+                out_flat[idx] = noncentral_chisquare_single(df, nonc)
+            return out
+
+    elif isinstance(size, (types.UniTuple)) and isinstance(size.dtype, types.Integer):
+
+        def noncentral_chisquare_impl(df, nonc, size=None):
```
esc, line 1633: I don't see a difference here between this implementation and the one where `size` is just an integer above. Is this perhaps an opportunity to reduce duplication?

numba/tests/test_random.py
```diff
@@ -1374,6 +1380,92 @@ def test_dirichlet_exceptions(self):
...
+class TestRandomNoncentralChiSquare(BaseTest):
+
+    def _check_sample(self, size, sample):
+
+        """Check output structure"""
+        if size is not None:
+            self.assertIsInstance(sample, np.ndarray)
+            self.assertEqual(sample.dtype, np.float64)
+            
+            if type(size) is int:
```
esc, line 1392: In this case, I believe that `isinstance` is the preferred idiom over using `type`.

numba/tests/test_random.py
```diff
@@ -1374,6 +1380,92 @@ def test_dirichlet_exceptions(self):
                 str(raises.exception),
             )
 
+class TestRandomNoncentralChiSquare(BaseTest):
+
+    def _check_sample(self, size, sample):
+
+        """Check output structure"""
```
esc, line 1387: Using triple quotes `"""` is commonly associated with docstrings rather than inline comments. I would recommend to us the hash-symbol instead: `#`.

numba/tests/test_random.py
```diff
@@ -1374,6 +1380,92 @@ def test_dirichlet_exceptions(self):
...
+            self.assertEqual(sample.dtype, np.float64)
+            
+            if type(size) is int:
+                self.assertEqual(sample.shape, (size,))
+            else:
+                self.assertEqual(sample.shape, size)
+        else:
+             self.assertIsInstance(sample, float)
+
+        """Check statistical properties"""
```
esc, line 1399: As above, use `#` instead.

numba/tests/test_random.py
```diff
@@ -1374,6 +1380,92 @@ def test_dirichlet_exceptions(self):
...
+            (1, 5),
+            (5, 1),
+            (100000, 1),
+            (1, 10000),
+        )
+        for input in inputs:
+            df, nonc = input
+            res = cfunc(df, nonc)
+            self._check_sample(None, res)
+            nonc = np.nan
```
esc, line 1419: I don't think you need this assignment, just pass in `np.nan` -- the comment makes it clear what is going on.

numba/tests/test_random.py
```diff
@@ -1374,6 +1380,92 @@ def test_dirichlet_exceptions(self):
...
+        sizes = (None, (10,), (10, 10))
+        inputs = (
+            (0.5, 1),
+            (1, 5),
+            (5, 1),
+            (100000, 1),
+            (1, 10000),
+        )
+
+        for input, size in itertools.product(inputs, sizes):
```
esc, line 1438: I think you can do:

```suggestion
        for (df, nonc), size in itertools.product(inputs, sizes):
```

And delete the line below.

numba/tests/test_random.py
```diff
@@ -1374,6 +1380,92 @@ def test_dirichlet_exceptions(self):
...
+            df, nonc = input
+            res = cfunc(df, nonc, size)
+            self._check_sample(size, res)
+            nonc = np.nan
+            res = cfunc(df, nonc, size)
```
esc, line 1443: Would suggest to hand in `np.nan` here too and add the comment as above.

numba/tests/test_random.py
```diff
@@ -1374,6 +1380,92 @@ def test_dirichlet_exceptions(self):
...
+        """
+        cfunc = jit(nopython=True)(numpy_noncentral_chisquare_default)
+        inputs = (
+            (0.5, 1), # test branch when df < 1
+            (1, 5),
+            (5, 1),
+            (100000, 1),
+            (1, 10000),
+        )
+        for input in inputs:
```
esc, line 1415:
```suggestion
        for df, nonc in inputs:
```

You can unpack directly in the loop and thus delete the line below.

esc (changes requested):
The functionality looks good. I have left some stylistic comments. I believe this will be ready to merge after those have been adressed.

numba/cpython/randomimpl.py
rishi-kulkarni: Thank you for your comments. 
a) Yes, I've made it so. 
b) Each element will be a different realization of the RNG, so unfortunately, no. I tried to rewrite it so that it could take advantage of the fact that its underlying distributions can take size arguments, but it did not work so well, as none of the current random distribution implementations can handle arrays for their location and scale arguments.

numba/cpython/randomimpl.py
rishi-kulkarni: Yes, I've condensed these now.

numba/tests/test_random.py
rishi-kulkarni: Good idea.

rishi-kulkarni:
@esc Thank you for your comments. I've made the requested changes. 

As an aside, I encountered some mysterious behavior with the current np.random.chisquare implementation - it can handle size arguments of int, tuple of ints, or not being given one, but it throws an error if size=None is given to it (despite that nominally being its default size argument). Any thoughts?

esc:
> @esc Thank you for your comments. I've made the requested changes.

Thank you, I will take a look tomorrow.

> As an aside, I encountered some mysterious behavior with the current np.random.chisquare implementation - it can handle size arguments of int, tuple of ints, or not being given one, but it throws an error if size=None is given to it (despite that nominally being its default size argument). Any thoughts?

Hmmm, usually we vow to emulate Numpy as far as possible. But in this case, that may sound like a Numpy bug. Would recommend to check their issues tarcker and we can ask @seberg for advice on this right here: Should the above ☝️ perhaps be reported to Numpy?

seberg:
Sounds weird, please open a bug-report with a clear repro (a simple try works for me).  I wonder if it is just some strange cython + new Python (i.e. cython version dependend) or similar issue though, and not really related to NumPy.

rishi-kulkarni:
Sorry, I mean numba's np.random.chisquare implementation has this issue, not numpy's. As in:

```python
import numpy as np
import numba as nb

# this works
@nb.njit
def nb_chi2(df):
    return np.random.chisquare(df)

nb_chi2(2)

# this also works
np.random.chisquare(2)

# this does not work
@nb.njit
def nb_chi2(df, size):
    return np.random.chisquare(df, size)

nb_chi2(2, None)

# while this does
np.random.chisquare(2, None)
```
This means you have to write something like this everywhere if you want to use np.random.chisquare in a jitted function, though you'd hope that numba's dispatch mechanism would be handling this for you. 

```python
if size is None:
    return np.random.chisquare(df)
else:
    return np.random.chisquare(df, None)
```

esc:
> ```
> # this does not work
> @nb.njit
> def nb_chi2(df, size):
>     return np.random.chisquare(df, size)
> 
> nb_chi2(2, None)
> ```


I can confirm this is the case. For reference, here is the error:

```
                File "/Users/vhaenel/git/numba/numba/core/types/functions.py", line 308, in get_call_type
                  sig = temp.apply(nolitargs, nolitkws)
                File "/Users/vhaenel/git/numba/numba/core/typing/templates.py", line 351, in apply
                  sig = generic(args, kws)
                File "/Users/vhaenel/git/numba/numba/core/typing/templates.py", line 972, in generic
                  result = self._definition_func(self.context, *args, **kws)
                File "/Users/vhaenel/git/numba/numba/core/overload_glue.py", line 37, in stub
                  return body(tyctx)  # noqa: F821
                File "/Users/vhaenel/git/numba/numba/core/overload_glue.py", line 140, in body
                  sig = typing.apply(ol_args, ol_kwargs)
                File "/Users/vhaenel/git/numba/numba/core/typing/templates.py", line 414, in apply
                  sig = typer(*args, **kws)
                File "/Users/vhaenel/git/numba/numba/core/typing/randomdecl.py", line 277, in typer
                  return self.array_typer(size)(a)
                File "/Users/vhaenel/git/numba/numba/core/typing/randomdecl.py", line 63, in typer
                  shape, ndim = normalize_shape(size)
                File "/Users/vhaenel/git/numba/numba/core/typing/randomdecl.py", line 37, in normalize_shape
                  ndim = len(shape)
              TypeError: object of type 'NoneType' has no len()
```

My guess would be that the `None` isn't being properly handled.

esc:
@rishi-kulkarni it's probably worth opening an issue about this.

numba/cpython/randomimpl.py
esc: OK, thank you for clearning this up, should be fine then.

stuartarchibald:
> @rishi-kulkarni it's probably worth opening an issue about this.

It's likely coming from here: https://github.com/numba/numba/blob/290bcaa63804fb99b1fe238b0f8fe1caf7686899/numba/core/typing/randomdecl.py#L264-L278

The typer needs to reject the case of `None`, this is an issue on mainline so need not be fixed here unless it prevents this PR from working. I've a feeling the fix might be challenging as it mixes old and new style APIs.

esc (approved):
@rishi-kulkarni thank you for this PR and making the required changes. I reviewed this by checking that the requested changes have been made and by checking the code locally.

I did some statistical eyeballing:

`df=2, nonc=1`

Screen Shot 2022-02-22 at 12 54 05

`df=4, nonc=3`

Screen Shot 2022-02-22 at 12 58 16

Comparing this to the wikipedia article at: https://en.wikipedia.org/wiki/Noncentral_chi-squared_distribution -- I think this looks fine.

rishi-kulkarni:
Thanks! I opened an issue about the problems with size arguments.

PR #7691, opened 2021-12-27
Fix `NUMBA_DUMP_ANNOTATION` with `NUMBA_EXTEND_VARIABLE_LIFETIMES`

PokhodenkoSA:

This PR fixes how `NUMBA_DUMP_ANNOTATION=1` and `NUMBA_EXTEND_VARIABLE_LIFETIMES=1` works together.

Source code `chk_annotate.py`:
```python
from numba import njit

@njit
def func(a):
  b = a
  return b

b = func(10)
```
Before:
```bash
$ NUMBA_DUMP_ANNOTATION=1 NUMBA_EXTEND_VARIABLE_LIFETIMES=1 python chk_annotate.py
-----------------------------------ANNOTATION-----------------------------------
# File: /.../chk_annotate.py
# --- LINE 3 --- 

@njit

# --- LINE 4 --- 

def func(a):

  # --- LINE 5 --- 
  # label 0
  #   a = arg(0, name=a)  :: int64
  #   b = a  :: int64
  #   del a

  b = a

  # --- LINE 6 --- 
  #   $8return_value.2 = cast(value=b)  :: int64
  #   del b
  #   return $8return_value.2

  return b
```

After:
```bash
$ NUMBA_DUMP_ANNOTATION=1 NUMBA_EXTEND_VARIABLE_LIFETIMES=1 python chk_annotate.py
-----------------------------------ANNOTATION-----------------------------------
# File: /.../chk_annotate.py
# --- LINE 3 --- 

@njit

# --- LINE 4 --- 

def func(a):

  # --- LINE 5 --- 
  # label 0
  #   a = arg(0, name=a)  :: int64
  #   b = a  :: int64

  b = a

  # --- LINE 6 --- 
  #   $8return_value.2 = cast(value=b)  :: int64
  #   del a
  #   del b
  #   return $8return_value.2

  return b

```

### DIFF
--- a/numba/core/typed_passes.py
+++ b/numba/core/typed_passes.py
@@ -186,7 +186,7 @@ class AnnotateTypes(AnalysisPass):
         """
         # add back in dels.
         post_proc = postproc.PostProcessor(state.func_ir)
-        post_proc.run(emit_dels=True)
+        post_proc.run(emit_dels=True, extend_lifetimes=config.EXTEND_VARIABLE_LIFETIMES)
 
         state.type_annotation = type_annotations.TypeAnnotation(
             func_ir=state.func_ir.copy(),


sklam:
This would mutate the IR in unexpected ways during the `AnnotateTypes` pass. The lifetime extension is currently occurring at `IRLegalization`, which happens right before lowering. Perhaps we should be moving `AnnotateTypes` to run after the `IRLegalization` pass and before the lowering pass.

PokhodenkoSA:
> This would mutate the IR in unexpected ways during the `AnnotateTypes` pass. The lifetime extension is currently occurring at `IRLegalization`, which happens right before lowering. Perhaps we should be moving `AnnotateTypes` to run after the `IRLegalization` pass and before the lowering pass.

I agree. Your proposal is better.

stuartarchibald:
> > This would mutate the IR in unexpected ways during the `AnnotateTypes` pass. The lifetime extension is currently occurring at `IRLegalization`, which happens right before lowering. Perhaps we should be moving `AnnotateTypes` to run after the `IRLegalization` pass and before the lowering pass.
> 
> I agree. Your proposal is better.

I've implemented this here: https://github.com/numba/numba/pull/7704

sklam:
Closed by https://github.com/numba/numba/pull/7704

Issue #6631, opened 2021-01-12
Numpy does not support Enum / IntEnum


korlaz:
numba 0.52
numpy 1.19.5
python 3.7

Numpy methods do not handle IntEnum type correctly.

```python
from enum import IntEnum
import numpy as np
from numba import njit

class Matrix(IntEnum):
    r = 4
    c = 4
    
@njit
def test():
    return np.zeros((Matrix.r, Matrix.c), dtype='i8')

test()
```
Output:

```
TypingError                               Traceback (most recent call last)
<ipython-input-9-fbd55f77ab7c> in <module>
----> 1 test()

~/miniconda3/lib/python3.7/site-packages/numba/core/dispatcher.py in _compile_for_args(self, *args, **kws)
    412                 e.patch_message(msg)
    413 
--> 414             error_rewrite(e, 'typing')
    415         except errors.UnsupportedError as e:
    416             # Something unsupported is present in the user code, add help info

~/miniconda3/lib/python3.7/site-packages/numba/core/dispatcher.py in error_rewrite(e, issue_type)
    355                 raise e
    356             else:
--> 357                 raise e.with_traceback(None)
    358 
    359         argtypes = []

TypingError: Failed in nopython mode pipeline (step: nopython frontend)
No implementation of function Function(<built-in function zeros>) found for signature:
 
 >>> zeros(UniTuple(IntEnum<int64>(Matrix) x 2), dtype=Literal[str](i8))
 
There are 2 candidate implementations:
  - Of which 2 did not match due to:
  Overload of function 'zeros': File: numba/core/typing/npydecl.py: Line 507.
    With argument(s): '(UniTuple(IntEnum<int64>(Matrix) x 2), dtype=unicode_type)':
   No match.

During: resolving callee type: Function(<built-in function zeros>)
During: typing of call at <ipython-input-8-0e7fb952e8a4> (11)


File "<ipython-input-8-0e7fb952e8a4>", line 11:
def test():
    return np.zeros((Matrix.r, Matrix.c), dtype='i8')
```

stuartarchibald:
Thanks for the report. This is currently unsupported so labelled as a `feature_request`. Casting the `IntEnum` to `intp` should suffice as a workaround:
```python
from enum import IntEnum
import numpy as np
from numba import njit

class Matrix(IntEnum):
    r = 4
    c = 4

@njit
def test():
    return np.zeros((np.intp(Matrix.r), np.intp(Matrix.c)), dtype='int64')

z = test()
[...]

PR #7695, opened 2021-12-29
`IntEnumMember` support for  `np.empty`, `np.zeros`, and `np.ones`

benwilliamgraham:
Adds `IntEnumMember` support for `np.empty`, `np.zeros`, and `np.ones`.

Closes [issue 6631](https://github.com/numba/numba/issues/6631), as well as the implicitly-casted version of the code in question:
```python
import numba
from enum import IntEnum
import numpy as np
from numba import njit

class Matrix(IntEnum):
    r = 4
    c = 4

@njit
def test():
    return np.zeros((Matrix.r, Matrix.c), dtype='int64')

z = test()
print(z, z.dtype)
```

Is there a more general solution to this problem that would allow support for IntEnumMember in other places, or a good way to automate its inclusions throughout the codebase?

### DIFF
--- a/numba/core/typing/npydecl.py
+++ b/numba/core/typing/npydecl.py
@@ -455,7 +455,7 @@ def parse_shape(shape):
     if isinstance(shape, types.Integer):
         ndim = 1
     elif isinstance(shape, (types.Tuple, types.UniTuple)):
-        if all(isinstance(s, types.Integer) for s in shape):
+        if all(isinstance(s, (types.Integer, types.IntEnumMember)) for s in shape):
             ndim = len(shape)
     return ndim
 
--- a/numba/core/typing/npydecl.py
+++ b/numba/core/typing/npydecl.py
@@ -455,7 +455,8 @@ def parse_shape(shape):
     if isinstance(shape, types.Integer):
         ndim = 1
     elif isinstance(shape, (types.Tuple, types.UniTuple)):
-        if all(isinstance(s, (types.Integer, types.IntEnumMember)) for s in shape):
+        int_tys = (types.Integer, types.IntEnumMember)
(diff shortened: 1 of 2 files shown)


numba/core/typing/npydecl.py
```diff
@@ -455,7 +455,7 @@ def parse_shape(shape):
     if isinstance(shape, types.Integer):
         ndim = 1
     elif isinstance(shape, (types.Tuple, types.UniTuple)):
-        if all(isinstance(s, types.Integer) for s in shape):
+        if all(isinstance(s, (types.Integer, types.IntEnumMember)) for s in shape):
```
gmarkall, line 458: Although this file is not automatically checked by flake8 (it should be, but needs fixing and removing from the ignore list) this line is over-long (more than 80 chars) - perhaps this could be resolved e.g. by:

```suggestion
        int_tys = (types.Integer, types.IntEnumMember)
        if all(isinstance(s, int_tys) for s in shape):
```

gmarkall (changes requested):
Many thanks for the PR! This does indeed appear to resolve issue #6631 (checked with a manual test).

I have one small suggestion on the diff, only for a non-automatically-checked flake8 issue.

> Is there a more general solution to this problem that would allow support for IntEnumMember in other places, or a good way to automate its inclusions throughout the codebase?

We'd probably have to rethink the typing of `IntEnum` members for this, which might be a little bit fiddly.

numba/core/typing/npydecl.py
benwilliamgraham: Sounds good, I'll add that!

gmarkall (approved):
Many thanks for the PR - this looks good to me!

Issue #7353, opened 2021-08-29
Implement `np.ascontiguousarray` for Numba scalars


brandonwillard:
The conversion for `np.ascontiguousarray` doesn't appear to handle scalar inputs:
```python
import numpy as np
import numba as nb


@nb.njit
def test_fn(x):
    return np.ascontiguousarray(x)


test_fn(1)
# ...
# TypingError: Failed in nopython mode pipeline (step: nopython frontend)
# No implementation of function Function(<function ascontiguousarray at 0x7fdd294aab00>) found for signature:
#
#  >>> ascontiguousarray(int64)


np.ascontiguousarray(1)
# array([1])
```

JohamyB:
Hi @gmarkall ,
I'm new here and want to take this one as my first issue?

JohamyB:
Hi,
I have been checking the documentation and the code, but I'm still not able to debug, can you please provide me a guide on this?

JohamyB:
@stuartarchibald  I have tried with this: https://llvmlite.readthedocs.io/en/v0.25.0/admin-guide/install.html but I'm still not able to update llvmlite.  Do you know if I'm missing something else?

Installed c:\users\laptop1\anaconda3\lib\site-packages\numba-0.55.0.dev0+502.g453765375.dirty-py3.7-win-amd64.egg
Processing dependencies for numba==0.55.0.dev0+502.g453765375.dirty
Searching for llvmlite<0.39,>=0.38.0dev0
Reading https://pypi.org/simple/llvmlite/
No local packages or working download links found for llvmlite<0.39,>=0.38.0dev0
error: Could not find suitable distribution for Requirement.parse('llvmlite<0.39,>=0.38.0dev0')

stuartarchibald:
@JohamyB the issue is that you need a "development" version of `llvmlite`. `conda` packages are the way to obtain these: https://llvmlite.readthedocs.io/en/stable/admin-guide/install.html#pre-built-binaries and as noted in the docs:
> Development releases of binary wheels are not made available.

Alternatively you can build them yourself but they are quite challenging: https://llvmlite.readthedocs.io/en/stable/admin-guide/install.html#building-manually

DhruvPatel01:
@JohamyB Are you still working on this? I would like to work on it, if you are not.

JohamyB:
Hi @DhruvPatel01 ,
Sure, you can work on this.

jeertmans:
[...]

PR #7700, opened 2022-01-01
Support for scalar arguments in Np.ascontiguousarray 

DhruvPatel01:

#7353 

## Details

I have added a code to overload scalar arguments to `np.ascontiguousarray` and `np.asfortranarray`. Existing implementations use low level API `glue_lowering`, but I couldn't figure out how to modify the code to handle scalars. Hopefully mixing high level API (`overload`) and low level API is fine.

### DIFF
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -4331,26 +4331,28 @@ def array_ascontiguousarray(context, builder, sig, args):
     return _as_layout_array(context, builder, sig, args, output_layout='C')
 
 
-def _array_ascontiguous_scalar(a):
-    """common between ascontiguousarray and asfortranarray"""
-    if isinstance(a, (types.Number, types.Boolean, )):
-        dtype = a
-
+@overload(np.ascontiguousarray)
+def array_ascontiguousarray_scalar(a):
+    """
+    This is an implementation for scalar.
+    For arrays, see `array_ascontiguousarray`.
+    """
+    if isinstance(a, (types.Number, types.Boolean,)):
         def impl(a):
-            ary = np.empty((1, ), dtype=dtype)
-            ary[0] = a
-            return ary
+            return np.ascontiguousarray(np.array(a))
         return impl
 
 
-@overload(np.ascontiguousarray)
-def array_ascontiguousarray_scalar(a):
-    return _array_ascontiguous_scalar(a)
-
-
 @overload(np.asfortranarray)
-def array_asfortran_scalar(a):
-    return _array_ascontiguous_scalar(a)
+def array_asfortranarray_scalar(a):
+    """
+    This is an implementation for scalar.
+    For arrays, see `array_asfortranarray`.
+    """
+    if isinstance(a, (types.Number, types.Boolean,)):
+        def impl(a):
+            return np.asfortranarray(np.array(a))
+        return impl
 
 
 @lower_builtin("array.astype", types.Array, types.DTypeSpec)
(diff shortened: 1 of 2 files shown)


DhruvPatel01:
I am not sure why build is failing in Python 3.10 Linux. The failed cases seem unrelated to this commit.

njriasan:
Hi @DhruvPatel01. Regarding I'll defer to maintainers for a more complete answer to `Hopefully mixing high level API (overload) and low level API is fine`, but I think it should be okay in this case because the type checks will never overlap.

For my testing this implementation/review this implementation seems correct but, it would be great if you could add a test to verify your implementation is working as expected.

It also seems to me that the build is failing in Python 3.10 Linux due to an unrelated error.

Just as a note I think many of the maintainers are still on a holiday break this week, so there may be a delay in getting more detailed feedback.

numba/np/arrayobj.py
```diff
@@ -4331,6 +4331,28 @@ def array_ascontiguousarray(context, builder, sig, args):
...
+        return impl
+
+
+@overload(np.ascontiguousarray)
+def array_ascontiguousarray_scalar(a):
+    return _array_ascontiguous_scalar(a)
+
+
+@overload(np.asfortranarray)
+def array_asfortran_scalar(a):
```
njriasan, line 4352: Personally I think it would be helpful if you could add a docstring here explaining this handles the scalar implementation and that the array cases are handled with the lower level glue_* APIs. Someone reading through this who is use to searching for overloads may be surprised when this is all they find.

numba/np/arrayobj.py
```diff
@@ -4331,6 +4331,28 @@ def array_ascontiguousarray(context, builder, sig, args):
     return _as_layout_array(context, builder, sig, args, output_layout='C')
 
 
+def _array_ascontiguous_scalar(a):
```
njriasan, line 4334: I'm not too familiar with the Fortran array layout (I never use it), but I suspect reusing the same function here is not completely correct because although it shouldn't be meaningful, the types won't accurately reflect the desired layout.

In particular based on existing code, this will always give the `layout='C'`, which won't be correct for the Fortran function. While this probably isn't meaningful in this case, I'm not sure what impact it might have when mixed with other APIs.

I'm not sure if this can be done with the higher level APIs, but I'll defer to people who are more knowledgable (perhaps @stuartarchibald).

njriasan (commented):
Thanks for submitting this. I'm not a maintainer, but I work with Numba a decent amount and figured I'd try help with this PR. The biggest issue is that in Numba, an array layout is part of type (i.e. C style or Fortran style) and reusing this higher level API will cause the fortran style to have an incorrect type.

I think this requires the lower level API at this time, but I'll let @stuartarchibald give the final word about that. If you need some help modifying your solution to integrate with the existing solution and include the correct layout I'm happy to help guide you through the code.

numba/np/arrayobj.py
DhruvPatel01: Agree, will add docstrings for it.

numba/np/arrayobj.py
DhruvPatel01: For single element arrays, I think they should be the same. I printed `np.ascontiguousarray(1).flags`. Both  `C_CONTIGUOUS ` and `F_CONTIGUOUS` were `True`. Same flags apart from `OWNDATA` flag were returned by this implementation. This also applies for `asfortranarray`.

stuartarchibald:
> I am not sure why build is failing in Python 3.10 Linux. The failed cases seem unrelated to this commit.

https://github.com/numba/numba/pull/7702 should hopefully fix.

numba/np/arrayobj.py
DhruvPatel01: Added the docstrings.

esc (approved):
Thank you for the patch!

stuartarchibald:
Thanks for the patch @DhruvPatel01, thanks for reviewing @njriasan @esc.

PR #7702, opened 2022-01-04
CI testing: Pin NumPy to 1.21 for Python 3.10.

stuartarchibald:
As title.

This should fix mainline, it's currently pulling NumPy 1.22 down from PyPI. This is breaking PRs etc, e.g. https://github.com/numba/numba/pull/7700#issuecomment-1003527445

`numba -s` has this in it (from the log of the above build):
```
2022-01-01T08:23:24.7306256Z numpy                     1.22.0                   pypi_0    pypi
```


esc:
I had a quick look at this and it seems Numpy is now being built from source and the wheels are not being picked up. Looking at the output, I see:

```
  setup.py:63: RuntimeWarning: NumPy 1.21.0 may not yet support Python 3.10.
```

So my suggestion would be to try Numpy `1.21.5` that is the latest version in the `1.21` series and it does indeed have 3.10 wheels.

esc:
I would also like to propose we move to pinning Numpy on `master` the same way we pin Python. The breakage caused by 1.21 and 1.22 caused quite some friction.

esc (approved):
LGTM

esc:
@stuartarchibald please feel free to merge as convenient.

Issue #6478, opened 2020-11-13
Array analysis error for constant float tuple creation


ehsantn:
## Reporting a bug

- [x] I have tried using the latest released version of Numba (most recent is
 visible in the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [x] I have included below a minimal working reproducer (if you are unsure how
 to write one see http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports).

Minimal example:
```python
import numba

@numba.njit(parallel=True)
def f():
    a = (1.2, 1.3)
    return a[0]

f()
```
```
  File "/Users/ehsan/dev/numba/numba/parfors/parfor.py", line 2708, in run
    self._pre_run()
  File "/Users/ehsan/dev/numba/numba/parfors/parfor.py", line 2700, in _pre_run
    self.array_analysis.run(self.func_ir.blocks)
  File "/Users/ehsan/dev/numba/numba/parfors/array_analysis.py", line 1119, in run
    self._run_on_blocks(topo_order, blocks, cfg, init_equiv_set)
  File "/Users/ehsan/dev/numba/numba/parfors/array_analysis.py", line 1139, in _run_on_blocks
    pending_transforms = self._determine_transform(
  File "/Users/ehsan/dev/numba/numba/parfors/array_analysis.py", line 1220, in _determine_transform
    pre, post = self._analyze_inst(
  File "/Users/ehsan/dev/numba/numba/parfors/array_analysis.py", line 1337, in _analyze_inst
    needs_define = equiv_set.insert_equiv(lhs, shape)
  File "/Users/ehsan/dev/numba/numba/parfors/array_analysis.py", line 518, in insert_equiv
    obj_names = [self._get_names(x) for x in objs]
  File "/Users/ehsan/dev/numba/numba/parfors/array_analysis.py", line 518, in <listcomp>
    obj_names = [self._get_names(x) for x in objs]
  File "/Users/ehsan/dev/numba/numba/parfors/array_analysis.py", line 431, in _get_names
    raise NotImplementedError(
NotImplementedError: Failed in nopython mode pipeline (step: convert to parfors)
ShapeEquivSet does not support 1.2
```

esc:
@ehsantn thanks for reporting, I can reproduce. cc @DrTodd13

stuartarchibald:
Suspect this will "fix" it:
```patch
diff --git a/numba/parfors/array_analysis.py b/numba/parfors/array_analysis.py
[...]

PR #7703, opened 2022-01-04
Ignore unsupported types in `ShapeEquivSet._getnames()`

benwilliamgraham:
Fix for issue #6478

### DIFF
--- a/numba/parfors/array_analysis.py
+++ b/numba/parfors/array_analysis.py
@@ -5,7 +5,6 @@
 
 import numpy
 import operator
-import logging
 from numba.core import types, ir, config, cgutils, errors
 from numba.core.ir_utils import (
     mk_unique_var,
@@ -31,8 +30,6 @@ from numba.core.extending import intrinsic, register_jitable
 import llvmlite.llvmpy.core as lc
 import llvmlite
 from numba.np.unsafe.ndarray import to_fixed_tuple
-
-_logger = logging.getLogger(__name__)
 
 UNKNOWN_CLASS = -1
 CONST_CLASS = 0
@@ -450,11 +447,19 @@ class ShapeEquivSet(EquivSet):
             else:
                 return (obj.value,)
         elif isinstance(obj, tuple):
-            return tuple(self._get_names(x)[0] for x in obj)
+
+            def get_names(x):
+                names = self._get_names(x)
+                if len(names) != 0:
+                    return names[0]
+                return names
+
+            return tuple(get_names(x) for x in obj)
         elif isinstance(obj, int):
             return (obj,)
-        _logger.debug(
-            f"Ignoring untracked object type {type(obj)} in ShapeEquivSet")
+        if config.DEBUG_ARRAY_OPT >= 1:
+            print(
+                f"Ignoring untracked object type {type(obj)} in ShapeEquivSet")
         return ()
 
     def is_equiv(self, *objs):
--- a/numba/parfors/array_analysis.py
+++ b/numba/parfors/array_analysis.py
@@ -5,6 +5,7 @@
 
 import numpy
 import operator
+import warnings
 from numba.core import types, ir, config, cgutils, errors
 from numba.core.ir_utils import (
     mk_unique_var,
@@ -450,10 +451,9 @@ class ShapeEquivSet(EquivSet):
             return tuple(self._get_names(x)[0] for x in obj)
         elif isinstance(obj, int):
             return (obj,)
-        else:
-            raise NotImplementedError(
-                "ShapeEquivSet does not support {}".format(obj)
-            )
+        warnings.warn(errors.NumbaPerformanceWarning(
+            f"Ignoring untracked object type {type(obj)} in ShapeEquivSet"))
+        return ()
 
     def is_equiv(self, *objs):
         """Overload EquivSet.is_equiv to handle Numba IR variables and
--- a/numba/parfors/array_analysis.py
+++ b/numba/parfors/array_analysis.py
@@ -483,7 +483,7 @@ class ShapeEquivSet(EquivSet):
         return the scalar value, or None otherwise.
         """
         names = self._get_names(obj)
-        if len(names) > 1:
+        if len(names) != 1:
             return None
         return super(ShapeEquivSet, self).get_equiv_const(names[0])
 
@@ -502,7 +502,7 @@ class ShapeEquivSet(EquivSet):
         """Return the set of equivalent objects.
         """
         names = self._get_names(obj)
-        if len(names) > 1:
+        if len(names) != 1:
             return None
         return super(ShapeEquivSet, self).get_equiv_set(names[0])
 
--- a/numba/parfors/array_analysis.py
+++ b/numba/parfors/array_analysis.py
@@ -5,7 +5,7 @@
 
 import numpy
 import operator
-import warnings
+import logging
 from numba.core import types, ir, config, cgutils, errors
 from numba.core.ir_utils import (
     mk_unique_var,
@@ -31,6 +31,8 @@ from numba.core.extending import intrinsic, register_jitable
 import llvmlite.llvmpy.core as lc
 import llvmlite
 from numba.np.unsafe.ndarray import to_fixed_tuple
+
+_logger = logging.getLogger(__name__)
 
 UNKNOWN_CLASS = -1
 CONST_CLASS = 0
@@ -451,8 +453,8 @@ class ShapeEquivSet(EquivSet):
             return tuple(self._get_names(x)[0] for x in obj)
         elif isinstance(obj, int):
             return (obj,)
-        warnings.warn(errors.NumbaPerformanceWarning(
-            f"Ignoring untracked object type {type(obj)} in ShapeEquivSet"))
+        _logger.debug(
+            f"Ignoring untracked object type {type(obj)} in ShapeEquivSet")
         return ()
 
     def is_equiv(self, *objs):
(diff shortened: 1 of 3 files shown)


numba/tests/test_tuples.py
```diff
@@ -772,6 +772,15 @@ def pyfunc(a):
 
         check((10, 20))
 
+    def test_untraced_value_tuple(self):
+        # This is a test for issue #6478.
+        def pyfunc():
+            a = (1.2, 1.3)
+            return a[0]
+
+        cfunc = jit(nopython=True)(pyfunc)
```
njriasan, line 781: I think you need `parallel=True` to test this issue.

njriasan: You can also just use njit here to avoid `nopython=True`.

numba/tests/test_tuples.py
```diff
@@ -772,6 +772,36 @@ def pyfunc(a):
...
+                arr3[j] = arr2[j] - a[1]
+            total = 0.0
+            n3 = len(arr3)
+            for k in prange(n3):
+                total += arr3[k]
+            return total + a[0]
+
+        arr = np.arange(100)
+        cfunc = jit(nopython=True, parallel=True)(pyfunc)
+        self.assertEqual(cfunc(arr), pyfunc(arr))
```
njriasan, line 804: This check is leading to failures because `self.assertEqual` is looking for exact equality and floating point computation may have slight differences. If you replace this with `self.assertPreciseEqual` you should be able to specify approximate correctness.

```suggestion
        self.assertPreciseEqual(cfunc(arr), pyfunc(arr), prec="approximate")
```

numba/tests/test_tuples.py
```diff
@@ -772,6 +772,36 @@ def pyfunc(a):
 
         check((10, 20))
 
+    def test_untraced_value_tuple(self):
+        # This is a test for issue #6478.
+        def pyfunc():
+            a = (1.2, 1.3)
+            return a[0]
+
+        cfunc = jit(nopython=True, parallel=True)(pyfunc)
```
njriasan, line 781: Not all targets support `parallel=True`, so this is failing on the 32 bit architecture. You should mark all tests with `parallel=True` as `@skip_parfors_unsupported`

numba/tests/test_tuples.py
```diff
@@ -772,6 +772,36 @@ def pyfunc(a):
 
         check((10, 20))
 
+    def test_untraced_value_tuple(self):
```
njriasan, line 775:
```suggestion
    @skip_parfors_unsupported
    def test_untraced_value_tuple(self):
```

numba/tests/test_tuples.py
```diff
@@ -772,6 +772,36 @@ def pyfunc(a):
...
+        # This is a test for issue #6478.
+        def pyfunc():
+            a = (1.2, 1.3)
+            return a[0]
+
+        cfunc = jit(nopython=True, parallel=True)(pyfunc)
+        self.assertEqual(cfunc(), pyfunc())
+
+    def test_untraced_value_parfor(self):
```
njriasan, line 784:
```suggestion
    @skip_parfors_unsupported
    def test_untraced_value_parfor(self):
```

njriasan (commented):
@benwilliamgraham See my comments for changes that I think should help get this passing CI.

numba/parfors/array_analysis.py
```diff
@@ -450,10 +451,9 @@ def _get_names(self, obj):
             return tuple(self._get_names(x)[0] for x in obj)
         elif isinstance(obj, int):
             return (obj,)
-        else:
-            raise NotImplementedError(
-                "ShapeEquivSet does not support {}".format(obj)
-            )
+        warnings.warn(errors.NumbaPerformanceWarning(
```
ehsantn, line 454: I think throwing a warning here may be too much, since these cases are not likely to affect performance. Maybe some debug/trace print is enough.

numba/parfors/array_analysis.py
```diff
@@ -450,10 +451,9 @@ def _get_names(self, obj):
...
+            f"Ignoring untracked object type {type(obj)} in ShapeEquivSet"))
+        return ()
```
ehsantn, line 456: We have to check all the uses of `_get_names` in the code to make sure this empty tuple output works as expected. For example, `get_equiv_const` uses `names[0]` for the output of `_get_names` which can fail for empty tuple. Also, some places may expect the length of this to be the same as the tuple.

njriasan: From what I see, the places that assume size of at least 1 are:
- get_equiv_const
- get_equiv_set
- A recursive call to `_get_names` with a tuple

To test the recursive call to `_get_names` you can probably use `((1.2, 1.1))`

I'm not sure how to test the equiv ones, but since they clearly break in the empty tuple case, we can probably replace `> 1` with `!= 1` without any consequences.

benwilliamgraham:
@stuartarchibald Is there a better way to test internal functions and ensure the parfor fusion is being performed as expected?

numba/parfors/array_analysis.py
```diff
@@ -5,6 +5,7 @@
 
 import numpy
 import operator
+import logging
```
DrTodd13, line 8: Array analysis should use NUMBA_DEBUG_ARRAY_OPT and not logging.

numba/parfors/array_analysis.py
```diff
@@ -31,6 +32,8 @@
 import llvmlite
 from numba.np.unsafe.ndarray import to_fixed_tuple
 
+_logger = logging.getLogger(__name__)
```
DrTodd13, line 35: Either we should switch everything to logger or use the existing NUMBA_DEBUG_ARRAY_OPT.  I suggest the latter because I see no compelling interest to switch at this point.

numba/parfors/array_analysis.py
```diff
@@ -450,10 +453,9 @@ def _get_names(self, obj):
             return tuple(self._get_names(x)[0] for x in obj)
         elif isinstance(obj, int):
             return (obj,)
-        else:
-            raise NotImplementedError(
-                "ShapeEquivSet does not support {}".format(obj)
-            )
+        _logger.debug(
+            f"Ignoring untracked object type {type(obj)} in ShapeEquivSet")
```
DrTodd13, line 457: Same comment as above about not using logger.

numba/tests/test_tuples.py
```diff
@@ -772,6 +772,48 @@ def pyfunc(a):
 
         check((10, 20))
 
+    @skip_parfors_unsupported
+    def test_untraced_value_tuple(self):
```
DrTodd13, line 776: Please move this test to test_parfors.py.  All parfor related tests go in that file.  That file has infrastructure for running in parallel and non-parallel and comparing results which you are doing manually here.

DrTodd13 (changes requested):

DrTodd13:
> @stuartarchibald Is there a better way to test internal functions and ensure the parfor fusion is being performed as expected?

Look in test_parfors.py and you'll see tests that check the parfor count after fusion.  That is the BKM for this check at the present.

benwilliamgraham:
> > @stuartarchibald Is there a better way to test internal functions and ensure the parfor fusion is being performed as expected?
> 
> Look in test_parfors.py and you'll see tests that check the parfor count after fusion. That is the BKM for this check at the present.

Great, thanks for the review! It should be updated now!

numba/tests/test_parfors.py
```diff
@@ -1956,6 +1956,47 @@ def test_impl(x, sz):
...
         self.check(test_impl, np.zeros((10, 10)), 3)
 
+    def test_untraced_value_tuple(self):
+        # This is a test for issue #6478.
+        def pyfunc():
+            a = (1.2, 1.3)
+            return a[0]
+
+        cfunc = jit(nopython=True, parallel=True)(pyfunc)
+        self.assertEqual(cfunc(), pyfunc())
```
DrTodd13: Replace this line and the one above with "self.check(pyfunc)".

numba/tests/test_parfors.py
```diff
@@ -1956,6 +1956,47 @@ def test_impl(x, sz):
...
+
+    def test_recursive_untraced_value_tuple(self):
+        # This is a test for issue #6478.
+        def pyfunc():
+            a = ((1.2, 1.3),)
+            return a[0][0]
+
+        cfunc = jit(nopython=True, parallel=True)(pyfunc)
+        self.assertEqual(cfunc(), pyfunc())
```
DrTodd13: Same comment as above.  Replace with builtin mechanism for running both and testing results with self.check.

DrTodd13 (approved):
Wow...somehow you made the exact changes I wanted you to make while I was writing the review.

stuartarchibald:
Thanks for the reviews so far. Looks like CI is broken at present.

numba/tests/test_parfors.py
```diff
@@ -1956,6 +1956,47 @@ def test_impl(x, sz):
...
+            for j in prange(n2):
+                arr3[j] = arr2[j] - a[1]
+            total = 0.0
+            n3 = len(arr3)
+            for k in prange(n3):
+                total += arr3[k]
+            return total + a[0]
+
+        arg = (types.Array(types.int64, 1, 'C'), )
+        self.assertEqual(countParfors(test_impl, arg), 1)
```
stuartarchibald, line 1998: This test should also execute the sample code with some meaningful input to make sure the answer is correct?

stuartarchibald (commented):
Thanks for the patch. I've made one suggestion RE testing, once resolve looks good to merge. NOTE: I've also manually checked the `DEBUG_ARRAY_OPT` output.

numba/tests/test_parfors.py
benwilliamgraham: Sounds good!

stuartarchibald (approved):

stuartarchibald:
@njriasan @ehsantn @DrTodd13 thanks for your input on this.

PR #7704, opened 2022-01-05
Move the type annotation pass to post legalization.

stuartarchibald:
This PR moves the type annotation pass to run after legalization,
this is so as to pick up the variable lifetimes correctly. The
`AnnotateTypes` pass now depends on the `IRLegalization` pass. The
rest of the changes are fixes to change use of members of the type
annotations object (e.g. `typemap`) to the same member on the
pipeline `state` object.

<!--

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -173,7 +173,6 @@ exclude =
     numba/tests/test_profiler.py
     numba/tests/test_numpyadapt.py
     numba/tests/test_stencils.py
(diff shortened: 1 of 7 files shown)


sklam (approved):
The changes make sense. 

Side note, i'm a bit surprised by why we used to have `state.type_annotation.typemap`.

Issue #270, opened 2013-07-02
AttributeError: 'Name' object has no attribute 'args' when using range


asmeurer:
``` pytb
Traceback (most recent call last):
  File "test4.py", line 3, in <module>
    @jit('void()')
  File "/Users/aaronmeurer/anaconda/envs/numba-tests/lib/python2.7/site-packages/numba/decorators.py", line 222, in _jit_decorator
    env, func, argtys, restype=return_type, nopython=nopython, **kwargs)
  File "/Users/aaronmeurer/anaconda/envs/numba-tests/lib/python2.7/site-packages/numba/decorators.py", line 133, in compile_function
    func_env = pipeline.compile2(env, func, restype, argtypes, **kwds)
  File "/Users/aaronmeurer/anaconda/envs/numba-tests/lib/python2.7/site-packages/numba/pipeline.py", line 137, in compile2
    post_ast = pipeline(func_ast, env)
  File "/Users/aaronmeurer/anaconda/envs/numba-tests/lib/python2.7/site-packages/numba/pipeline.py", line 184, in __call__
    ast = self.transform(ast, env)
  File "/Users/aaronmeurer/anaconda/envs/numba-tests/lib/python2.7/site-packages/numba/pipeline.py", line 621, in transform
    ast = stage(ast, env)
  File "/Users/aaronmeurer/anaconda/envs/numba-tests/lib/python2.7/site-packages/numba/pipeline.py", line 606, in _stage
    return _check_stage_object(stage_obj)(ast, env)
  File "/Users/aaronmeurer/anaconda/envs/numba-tests/lib/python2.7/site-packages/numba/pipeline.py", line 187, in __call__
    ast = self.transform(ast, env)
  File "/Users/aaronmeurer/anaconda/envs/numba-tests/lib/python2.7/site-packages/numba/pipeline.py", line 417, in transform
    return transform.visit(ast)
  File "/Users/aaronmeurer/anaconda/envs/numba-tests/lib/python2.7/ast.py", line 241, in visit
    return visitor(node)
  File "/Users/aaronmeurer/anaconda/envs/numba-tests/lib/python2.7/ast.py", line 297, in generic_visit
    value = self.visit(value)
  File "/Users/aaronmeurer/anaconda/envs/numba-tests/lib/python2.7/ast.py", line 241, in visit
    return visitor(node)
[...]

Issue #45, opened 2012-10-25
Unary Operator Error - Negative (-) Symbol on Variables


mactabish:
<b>Getting Unary Operator error when running this code:</b>

```
def minimumFilter(image, filt):
    rows, columns = image.shape
    result = numpy.zeros(image.shape)

    filterSize = 5
    halfFilterSize = filterSize // 2    

    for x in range(halfFilterSize, rows - halfFilterSize):
        for y in range(halfFilterSize, columns - halfFilterSize):
            sMin = 255
            for j in range(-halfFilterSize, halfFilterSize):
                for i in range(-halfFilterSize, halfFilterSize):
                    filt[i + halfFilterSize, j + halfFilterSize] = image[x + i, y + j]
            for j in range(filterSize - 1):
                for i in range(filterSize - 1):
                    if filt[i, j] < sMin:
                         sMin = filt[i, j]
            result[x, y] = sMin

     return result
```

The problem is when I use: 

```
-halfFilterSize
```

To fix the issue, I used:

```
(-1) * halfFilterSize
```

which executed fine. Just thought this should be fixed. It works fine with constant numbers, but not for variables.

<b>Here is error output:</b>

Traceback (most recent call last):
  File "minimumFilter.py", line 9, in <module>
    @jit(argtypes=[int32[:,:], int32[:,:]], restype=int32[:,:])
  File "/opt/anaconda/lib/python2.7/site-packages/numba/decorators.py", line 268, in _jit2_decorator
    *_kwargs)
  File "/opt/anaconda/lib/python2.7/site-packages/numba/functions.py", line 183, in compile_function
    ctypes=ctypes, *_kwds)
  File "/opt/anaconda/lib/python2.7/site-packages/numba/pipeline.py", line 169, in compile
    context, func, restype, argtypes, codegen=True, *_kwds)
  File "/opt/anaconda/lib/python2.7/site-packages/numba/pipeline.py", line 130, in _infer_types
    return run_pipeline(context, func, ast, func_signature, *_kwargs)
  File "/opt/anaconda/lib/python2.7/site-packages/numba/pipeline.py", line 124, in run_pipeline
    return pipeline, pipeline.run_pipeline()
  File "/opt/anaconda/lib/python2.7/site-packages/numba/pipeline.py", line 70, in run_pipeline
    ast = getattr(self, method_name)(ast)
  File "/opt/anaconda/lib/python2.7/site-packages/numba/pipeline.py", line 81, in type_infer
[...]

PR #7705, opened 2022-01-05
Use `omp_set_max_active_levels()` instead of `omp_set_nested()`

### LINKED ISSUE #5275: Threading backend using `omp_set_nested` which is deprecated in OpenMP 5?
## Reporting a bug

This
```python
import numpy as np

@njit(parallel=True)
def foo(n):
    acc = 0
    for i in prange(n):
        acc += i
    return acc

print(foo(10))
```
run like:
```
NUMBA_THREADING_LAYER=omp python example.py
```
does this:
```
OMP: Info #273: omp_set_nested routine deprecated, please use omp_set_max_active_levels instead.
```

Think this is because OpenMP 5.0 deprecated `omp_set_nested` and Anaconda MKL 2020 packages depends on in Intel-OpenMP which is 5.0 compliant.

gmarkall:
`omp_set_nested()` is deprecated and often results in a warning on recent installations. Following the discussion in Issue #5275, this commit changes the call to `omp_set_max_active_levels()`. Ideally we'd set it to the maximum supported nesting level, but we can't determine what that is; the function `omp_get_supported_active_levels()` was only added in OpenMP 5.0, which is quite recent and was not supported by some implementations until very recently (e.g. it was added to GNU OpenMP in mid-2020). We arbitrarily set the nesting level to 32, which is far greater than any real requirement, and the implementation will limit the nesting levels to the supported number as per the spec if it is less than 32.

Fixes #5275.

### DIFF
--- a/docs/source/user/threading-layer.rst
+++ b/docs/source/user/threading-layer.rst
@@ -26,9 +26,9 @@ There are three threading layers available and they are named as follows:
 * ``workqueue`` -A simple built-in work-sharing task scheduler.
 
 In practice, the only threading layer guaranteed to be present is ``workqueue``.
-The ``omp`` layer requires the presence of a suitable OpenMP runtime library.
-The ``tbb`` layer requires the presence of Intel's TBB libraries, these can be
-obtained via the conda command::
+The ``omp`` layer requires the presence of a suitable (3.0 or greater) OpenMP
+runtime library.  The ``tbb`` layer requires the presence of Intel's TBB
+libraries, these can be obtained via the conda command::
 
     $ conda install tbb
 
--- a/numba/np/ufunc/omppool.cpp
+++ b/numba/np/ufunc/omppool.cpp
@@ -224,8 +224,21 @@ static void launch_threads(int count)
     if(count < 1)
         return;
     omp_set_num_threads(count);
-    omp_set_nested(0x1); // enable nesting, control depth with OMP env var
     _INIT_NUM_THREADS = count;
+
+    // Enable up to 32 levels of nested parallelism, or whatever the
+    // implementation supports. From the OpenMP 3.0 spec, p.127:
+    //
+    //     "If the number of parallel levels requested exceeds the number of
+    //     levels of parallelism supported by the implementation, the value of
+    //     the max-active-levels-var ICV will be set to the number of parallel
+    //     levels support by the implementation."
+    //
+    // (https://www.openmp.org/wp-content/uploads/spec30.pdf)
+    //
+    // 32 levels is likely to be far in excess of any practical use case, and
+    // therefore should not present any real-world limitation.
+    omp_set_max_active_levels(32);
 }
 
 static void synchronize(void)


gmarkall:
(cc @seanlaw)

gmarkall:
Argh: "Visual C++ supports the OpenMP 2.0 standard." - https://docs.microsoft.com/en-us/cpp/parallel/openmp/openmp-in-visual-cpp?view=msvc-170

gmarkall:
The Linux failures look pretty weird, but I wonder if this is just a knock-on effect of the function not being supported in the environmentfor that test:

<details>
<summary>Errors</summary>
```
Parallel: 463. Serial: 73
ss.........E.....................................................................s..........................................................................................................s...........sssssssssssssssss.............................................................................
Stderr:
/home/vsts/work/1/s/numba/tests/test_vectorization_type_inference.py:10: NumbaWarning: 
Compilation is falling back to object mode WITH looplifting enabled because Function "func" failed type inference due to: Untyped global name 'typeof': Cannot determine Numba type of <class 'function'>

File "numba/tests/test_vectorization_type_inference.py", line 13:
def func(dtypeA, dtypeB):
    <source elided>
    B = np.arange(10, dtype=dtypeB)
    return typeof(vector_add(A, B))
    ^

'typeof' looks like a Numba internal function, has it been imported (i.e. 'from numba import typeof')?

  def func(dtypeA, dtypeB):
/home/vsts/work/1/s/numba/core/object_mode_passes.py:151: NumbaWarning: Function "func" was compiled in object mode without forceobj=True.

File "numba/tests/test_vectorization_type_inference.py", line 10:

def func(dtypeA, dtypeB):
^

  warnings.warn(errors.NumbaWarning(warn_msg,
/home/vsts/work/1/s/numba/core/object_mode_passes.py:161: NumbaDeprecationWarning: 
Fall-back from the nopython compilation path to the object mode compilation path has been detected, this is deprecated behaviour.

For more information visit https://numba.pydata.org/numba-doc/latest/reference/deprecation.html#deprecation-of-object-mode-fall-back-behaviour-when-using-jit

File "numba/tests/test_vectorization_type_inference.py", line 10:

def func(dtypeA, dtypeB):
^
Ran 536 tests in 140.188s

FAILED (failures=1, errors=3, skipped=60, expected failures=1)
```</details>

**Comment by gmarkall** [comment]
@stuartarchibald Shall we give this a buildfarm run? I'd like to get some disappointments in early on in the year.

**Review by stuartarchibald — COMMENTED** [review]
Thanks for the patch and for investigating this. I think in principle this is a good change, but the proof is likely in testing on the farm as noted. It is of concern that in practice there may be edge cases that are not picked up, however the limitations are now documented.

I wonder if at compile time the OpenMP version could be detected and added to the current 2 vs 3+ split implied by `_MSC_VER`, this to accommodate e.g. software distributions with older OpenMP?

**Comment by gmarkall** [comment]
> I wonder if at compile time the OpenMP version could be detected and added to the current 2 vs 3+ split implied by _MSC_VER, this to accommodate e.g. software distributions with older OpenMP?

Would that add any tangible benefit? For binary packages, they'd generally be built with an OpenMP 3.0 or later (e.g. by us or conda-forge or Anaconda or a Linux distro maintainer, none of whom are not likely to use such an old OpenMP), and therefore all the users who don't build from source would still need to have OpenMP 3.0 - for those that do build from source, I don't imagine there are many who would build from source but have an ancient (pre-2008) OpenMP - although HPC clusters can be slow to update (and are a setting where Numba might be built instead of a binary version used), I would expect the OpenMP installation wouldn't be that old on any of them.

@seanlaw Are any of the reports of the OpenMP warning that you're aware of coming from Windows users?

**Comment by seanlaw** [comment]
> Are any of the reports of the OpenMP warning that you're aware of coming from Windows users?

@gmarkall So far, I think it has been primarily Mac OSX users (don't know if it affects both Intel and M1). I certainly see it on my Apple M1

**Comment by stuartarchibald** [comment]
> > I wonder if at compile time the OpenMP version could be detected and added to the current 2 vs 3+ split implied by _MSC_VER, this to accommodate e.g. software distributions with older OpenMP?
> 
> Would that add any tangible benefit? For binary packages, they'd generally be built with an OpenMP 3.0 or later (e.g. by us or conda-forge or Anaconda or a Linux distro maintainer, none of whom are not likely to use such an old OpenMP), and therefore all the users who don't build from source would still need to have OpenMP 3.0 - for those that do build from source, I don't imagine there are many who would build from source but have an ancient (pre-2008) OpenMP - although HPC clusters can be slow to update (and are a setting where Numba might be built instead of a binary version used), I would expect the OpenMP installation wouldn't be that old on any of them.
> 

@gmarkall thanks for your thoughts, this is essentially what I was wondering, is there some known use case where there are packagers of Numba that only have access to a 2.x OpenMP? I cannot think of any other cases than those mentioned either, so I guess we can leave it. There's also the option of disabling OpenMP support at compile time as a way out (using this https://numba.readthedocs.io/en/stable/user/installing.html#envvar-NUMBA_DISABLE_OPENMP-default-not-set).

File: docs/source/user/threading-layer.rst [hunk 1]
```diff
@@ -26,9 +26,10 @@ There are three threading layers available and they are named as follows:
 * ``workqueue`` -A simple built-in work-sharing task scheduler.
 
 In practice, the only threading layer guaranteed to be present is ``workqueue``.
-The ``omp`` layer requires the presence of a suitable OpenMP runtime library.
-The ``tbb`` layer requires the presence of Intel's TBB libraries, these can be
-obtained via the conda command::
+The ``omp`` layer requires the presence of a suitable (2.0 or greater on
+Windows, 3.0 or greater otherwise) OpenMP runtime library.  The ``tbb`` layer
```
stuartarchibald, line 30: Should probably mention the constraint in the installation/dependency list too? https://numba.readthedocs.io/en/stable/user/installing.html#dependency-list

gmarkall: I've added them there - the descriptions there were pretty vague already, but I believe I've improved it a little with some more specific info.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cpu_yaml_70`. This should highlight any immediate issues ahead of reviewing the patch.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cpu_yaml_70`. This should highlight any immediate issues ahead of reviewing the patch.

This is failing on ppc64le, at test: `numba.tests.test_num_threads.TestNumThreadsBackends.test_nested_parallelism_3_omp_4` (this test is auto generated) with:
```
Arrays are not equal, x: array(3), y: array([1.])
```
The actual test code that fails is:
https://github.com/numba/numba/blob/ef1ba4c10b753ca31e9d4ff3de874d85be1be9c2/numba/tests/test_num_threads.py#L459-L465

which maybe hints at `OMP_MAX_ACTIVE_LEVELS` being involved. What is somewhat strange is that this test fails but the very similar:
https://github.com/numba/numba/blob/ef1ba4c10b753ca31e9d4ff3de874d85be1be9c2/numba/tests/test_num_threads.py#L356-L360
passes.

stuartarchibald:
The failure reported in https://github.com/numba/numba/pull/7705#issuecomment-1018482015 appears to be consistent, a rerun of the test suite hits it again. It also appears on all configurations for ppc64le i.e. it's independent of Python/NumPy version, which is no surprise but also highlights that it's probably not transient, 3 builds fail the same way!

ziofil:
@gmarkall bump 🥺 (I have a recursive function with parallel loops)

gmarkall:
@ziofil I'd be surprised if the message were related to the crash - I'd suggest starting a thread on https://numba.discourse.group describing your problem, and hopefully you can get some more assistance there.

stuartarchibald:
With view of getting this PR merged, perhaps the `ppc64le` target can continue using `omp_set_nested` (along with windows), and a ticket is opened to record the issue and require investigation into what is happening. This would make it so that linux (which makes up the vast majority of downloads on PyPI) and OSX users will not see this message any more, which is an improvement. What do you think @gmarkall ?

gmarkall:
> What do you think @gmarkall ?

That sounds good to me. I'll update this PR to do as suggested.

stuartarchibald:
> > What do you think @gmarkall ?
> 
> That sounds good to me. I'll update this PR to do as suggested.

Great, thanks!

stuartarchibald:
https://github.com/numba/numba/issues/7981 is a record of the issue with ppc64le.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cpu_yaml_81`.

marioroy:
I'm seeing the deprecation notice on Ubuntu 20.04 (x86_64) using the Numba 0.56 branch and `NUMBA_THREADING_LAYER=omp`.
```text
OMP: Info #270: omp_set_nested routine deprecated, please use omp_set_max_active_levels instead.
```
Abstract from numba -s output:
```text
__OS Information__
Platform Name                                 : Linux-5.15.0-42-lowlatency-x86_64-with-glibc2.29
Platform Release                              : 5.15.0-42-lowlatency
OS Name                                       : Linux
OS Version                                    : #45~20.04.1-Ubuntu SMP PREEMPT Mon Jul 11 11:25:26 UTC 2022
OS Specific Version                           : ?
Libc Version                                  : glibc 2.31

Python Compiler                               : GCC 9.4.0
Python Implementation                         : CPython
Python Version                                : 3.8.13
Python Locale                                 : en_US.UTF-8

__Numba Toolchain Versions__
Numba Version                                 : 0.56.0rc1
llvmlite Version                              : 0.39.0rc1

__LLVM Information__
LLVM Version                                  : 11.0.0

__Threading Layer Information__
TBB Threading Layer Available                 : True
+-->TBB imported successfully.
OpenMP Threading Layer Available              : True
+-->Vendor: GNU
Workqueue Threading Layer Available           : True
+-->Workqueue imported successfully.

__Installed Packages__
Package            Version
------------------ ---------
llvmlite           0.39.0rc1
numba              0.56.0rc1
numpy              1.22.4
tbb                2021.5.1
tbb-devel          2021.5.1
```

stuartarchibald:
> Re: reproducing the fail may be possible on an x86 box with many cores.
> 
> I have a multi-core x86_64 system. The deprecation notice is appearing on Ubuntu 20.04 (x86_64) using the Numba 0.56 branch and `NUMBA_THREADING_LAYER=omp`.
> 
> ```
> OMP: Info #270: omp_set_nested routine deprecated, please use omp_set_max_active_levels instead.
> ```
> 
> Abstract from numba -s output:
> 
> ```
> __OS Information__
> Platform Name                                 : Linux-5.15.0-42-lowlatency-x86_64-with-glibc2.29
> Platform Release                              : 5.15.0-42-lowlatency
> OS Name                                       : Linux
> OS Version                                    : #45~20.04.1-Ubuntu SMP PREEMPT Mon Jul 11 11:25:26 UTC 2022
> OS Specific Version                           : ?
> Libc Version                                  : glibc 2.31
> 
> Python Compiler                               : GCC 9.4.0
> Python Implementation                         : CPython
> Python Version                                : 3.8.13
> Python Locale                                 : en_US.UTF-8
> 
> __Numba Toolchain Versions__
> Numba Version                                 : 0.56.0rc1
> llvmlite Version                              : 0.39.0rc1
> 
> __LLVM Information__
> LLVM Version                                  : 11.0.0
> 
> __Threading Layer Information__
> TBB Threading Layer Available                 : True
> +-->TBB imported successfully.
> OpenMP Threading Layer Available              : True
> +-->Vendor: GNU
> Workqueue Threading Layer Available           : True
> +-->Workqueue imported successfully.
> 
> __Installed Packages__
> Package            Version
> ------------------ ---------
> llvmlite           0.39.0rc1
> numba              0.56.0rc1
> numpy              1.22.4
> tbb                2021.5.1
> tbb-devel          2021.5.1
> ```

Thanks for taking a look @marioroy, the deprecation notice is expected and will still be present in the `release0.56` branch as it does not contain this patch. This patch is attempting to use the newer OpenMP API that won't raise the warning where we know it is supported, but there's a failing test that needs looking at before it can be merged.

samaid:
I checked with OpenMP experts. They recommend to check the presence of the `omp_set_max_active_levels` symbol on `dlopen()` (Linux, OSX)/`LoadLibrary()` (Windows). On Linux & OSX you can do this with `dlsym()` function. On Windows using `GetProcAddress()`. If the symbol is present then you don't need `omp_set_nested`. Otherwise you continue using it. That should work with any OpenMP on any platform.

sklam:
`numba_smoketest_cpu_yaml_160`

gmarkall:
@sklam How did it go on the buildfarm?

jbednar:
The Datashader team is encountering this warning; is there any chance this PR can be merged and included in the next release? Is there anything else that needs to be done for it?

gmarkall:
From discussion in the triage meeting - we'll give this another buildfarm run in case the potential latent parfors bugs the change seems to expose have been fixed in the interim.

gmarkall:
A quick note that I heard this passed on the buildfarm, but we really shouldn't be hiding the potential problem on PPC (because it could still exist in other scenarios outside the buildfarm), so I've pushed a revert of the change that used `omp_set_nested()` on PPC in preparation for another buildfarm run.

gmarkall:
If it passes the buildfarm I'll add release notes describing exactly what we end up doing.

sklam:
latest BFID: `numba_yaml_473`

gmarkall:
Closing to tidy things up. I think a new PR could be made if someone wants to take this over.

PR #7712, opened 2022-01-07
Fixing issue 7693

### LINKED ISSUE #7693: Incorrect result copying array-typed field of structured array in 0.55.0rc1
This works as expected in plain python but is incorrect in 0.55.0rc1 on Windows 10 with python 3.7.1

Correct plain python result: (1.2, [1, 2, 3])
Incorrect njit result: (1.2, [0, 0, 0])

```python
import numpy as np
from numba import njit

src_dtype = np.dtype([
    ("user", np.float64),
    ("array", np.int16, (3,))
], align=True)


dest_dtype = np.dtype([
    ("user1", np.float64),
    ("array1", np.int16, (3,))
], align=True)

@njit
def copy(index, src, dest):
    dest['user1'] = src[index]['user']
    dest['array1'] = src[index]['array']

source = np.empty(5, dtype=src_dtype)
dest = np.empty(5, dtype=dest_dtype)

source[0] = (1.2, [1, 2, 3])
copy(0, source, dest[0])
print(dest[0])

```

This was first discussed in [discourse](https://numba.discourse.group/t/copying-array-typed-field-of-structured-array/1093) using 0.53.1 with a different manifestation.  Shout-out to @HPLegion for suggesting trying the latest release  candidate.

luk-f-a:
Fixes the lowering of setitem of a nestedarray into the nestedarray field of a record. Changes to the generated LLVM IR in those cases, to prevent the metadata (strides, etc) of the source being copied into the data buffer of the destination array. 

fixes #7693

### DIFF
--- a/numba/cuda/tests/cudapy/test_record_dtype.py
+++ b/numba/cuda/tests/cudapy/test_record_dtype.py
@@ -99,6 +99,8 @@ recordwith2darray = np.dtype([('i', np.int32),
 
 nested_array1_dtype = np.dtype([("array1", np.int16, (3,))], align=True)
 
+nested_array2_dtype = np.dtype([("array2", np.int16, (3, 2))], align=True)
+
 
 # Functions used for "full array" tests
 
@@ -170,6 +172,10 @@ def record_read_2d_array01(ary):
 
 def assign_array_to_nested(dest, src):
     dest['array1'] = src
+
+
+def assign_array_to_nested_2d(dest, src):
+    dest['array2'] = src
 
 
 class TestRecordDtype(CUDATestCase):
@@ -461,6 +467,19 @@ class TestNestedArrays(CUDATestCase):
         expected = np.zeros(2, dtype=nested_array1_dtype)
 
         pyfunc = assign_array_to_nested
+        kernel = cuda.jit(pyfunc)
+
+        kernel[1, 1](got[0], src)
+        pyfunc(expected[0], src)
+
+        np.testing.assert_array_equal(expected, got)
+
+    def test_assign_array_to_nested_2d(self):
+        src = (np.arange(6) + 1).astype(np.int16).reshape((3, 2))
+        got = np.zeros(2, dtype=nested_array2_dtype)
+        expected = np.zeros(2, dtype=nested_array2_dtype)
+
+        pyfunc = assign_array_to_nested_2d
         kernel = cuda.jit(pyfunc)
 
         kernel[1, 1](got[0], src)
--- a/numba/cuda/tests/cudapy/test_record_dtype.py
+++ b/numba/cuda/tests/cudapy/test_record_dtype.py
@@ -97,6 +97,8 @@ recordwitharray = np.dtype(
 recordwith2darray = np.dtype([('i', np.int32),
                               ('j', np.float32, (3, 2))])
 
+nested_array1_dtype = np.dtype([("array1", np.int16, (3,))], align=True)
+
 
 # Functions used for "full array" tests
 
@@ -115,6 +117,10 @@ def recarray_set_record(ary, rec):
 def recarray_write_array_of_nestedarray_broadcast(ary):
     ary.j[:, :, :] = 1
     return ary
+
+
+def record_setitem_array(rec_source, rec_dest):
+    rec_dest['j'] = rec_source['j']
 
 
 def recarray_write_array_of_nestedarray(ary):
@@ -160,6 +166,10 @@ def record_read_2d_array10(ary):
 
 def record_read_2d_array01(ary):
     return ary.j[0, 1]
+
+
+def assign_array_to_nested(dest, src):
+    dest['array1'] = src
 
 
 class TestRecordDtype(CUDATestCase):
@@ -403,6 +413,20 @@ class TestNestedArrays(CUDATestCase):
         res = cfunc(nbval[0])
         np.testing.assert_equal(res, nbval[0].j[1, 0])
 
+    def test_setitem(self):
+        nbarr1 = np.recarray(2, dtype=recordwith2darray)
+        nbarr1[0] = np.array([(1, ((1, 2), (4, 5), (2, 3)))],
+                             dtype=recordwith2darray)[0]
+        nbarr2 = np.recarray(2, dtype=recordwith2darray)
+        nbarr2[0] = np.array([(10, ((10, 20), (40, 50), (20, 30)))],
+                             dtype=recordwith2darray)[0]
+        pyfunc = record_setitem_array
+        args = nbarr1[0], nbarr2[0]
+        arr_expected = pyfunc(*args)
+        cfunc = cuda.jit(pyfunc)
+        arr_res = cfunc[1, 1](*args)
+        np.testing.assert_equal(arr_res, arr_expected)
+
     def test_getitem_idx(self):
         # Test __getitem__ with numerical index
 
@@ -430,6 +454,45 @@ class TestNestedArrays(CUDATestCase):
         kernel = cuda.jit(pyfunc)
         kernel[1, 1](nbarr, rec)
         np.testing.assert_equal(nbarr, arr)
+
+    def test_assign_array_to_nested(self):
+        src = (np.arange(3) + 1).astype(np.int16)
+        got = np.zeros(2, dtype=nested_array1_dtype)
+        expected = np.zeros(2, dtype=nested_array1_dtype)
+
+        pyfunc = assign_array_to_nested
+        kernel = cuda.jit(pyfunc)
+
+        kernel[1, 1](got[0], src)
+        pyfunc(expected[0], src)
+
+        np.testing.assert_array_equal(expected, got)
+
+    def test_issue_7693(self):
+        src_dtype = np.dtype([
+            ("user", np.float64),
+            ("array", np.int16, (3,))],
+            align=True)
+
+        dest_dtype = np.dtype([
+            ("user1", np.float64),
+            ("array1", np.int16, (3,))],
+            align=True)
+
+        @cuda.jit
+        def copy(index, src, dest):
+            dest['user1'] = src[index]['user']
+            dest['array1'] = src[index]['array']
+
+        source = np.zeros(2, dtype=src_dtype)
+        got = np.zeros(2, dtype=dest_dtype)
+        expected = np.zeros(2, dtype=dest_dtype)
+
+        source[0] = (1.2, [1, 2, 3])
+        copy[1, 1](0, source, got[0])
+        copy.py_func(0, source, expected[0])
+
+        np.testing.assert_array_equal(expected, got)
 
     # Reading and returning arrays from recarrays - the following functions are
     # all xfailed because CUDA cannot handle returning arrays from device
--- a/numba/cuda/tests/cudapy/test_record_dtype.py
+++ b/numba/cuda/tests/cudapy/test_record_dtype.py
@@ -420,18 +420,22 @@ class TestNestedArrays(CUDATestCase):
         np.testing.assert_equal(res, nbval[0].j[1, 0])
 
     def test_setitem(self):
-        nbarr1 = np.recarray(2, dtype=recordwith2darray)
-        nbarr1[0] = np.array([(1, ((1, 2), (4, 5), (2, 3)))],
-                             dtype=recordwith2darray)[0]
-        nbarr2 = np.recarray(2, dtype=recordwith2darray)
-        nbarr2[0] = np.array([(10, ((10, 20), (40, 50), (20, 30)))],
-                             dtype=recordwith2darray)[0]
+        def gen():
+            nbarr1 = np.recarray(1, dtype=recordwith2darray)
+            nbarr1[0] = np.array([(1, ((1, 2), (4, 5), (2, 3)))],
+                                 dtype=recordwith2darray)[0]
+            nbarr2 = np.recarray(1, dtype=recordwith2darray)
+            nbarr2[0] = np.array([(10, ((10, 20), (40, 50), (20, 30)))],
+                                 dtype=recordwith2darray)[0]
+            return nbarr1[0], nbarr2[0]
         pyfunc = record_setitem_array
-        args = nbarr1[0], nbarr2[0]
-        arr_expected = pyfunc(*args)
+        pyargs = gen()
+        pyfunc(*pyargs)
+
         cfunc = cuda.jit(pyfunc)
-        arr_res = cfunc[1, 1](*args)
-        np.testing.assert_equal(arr_res, arr_expected)
+        cuargs = gen()
+        cfunc[1, 1](*cuargs)
+        np.testing.assert_equal(pyargs, cuargs)
 
     def test_getitem_idx(self):
         # Test __getitem__ with numerical index
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -2548,8 +2548,7 @@ def array_record_getitem(context, builder, sig, args):
 @lower_getattr_generic(types.Record)
 def record_getattr(context, builder, typ, value, attr):
     """
-    Generic getattr() implementation for records: fetch the given
-    record member, i.e. a scalar.
+    Generic getattr() implementation for records: get the given record member.
     """
     context.sentry_record_alignment(typ, attr)
     offset = typ.offset(attr)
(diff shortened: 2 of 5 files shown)


gmarkall:
This also fixes the [other reproducer from 7693](https://github.com/numba/numba/issues/7693#issuecomment-1005583958)

```python
import numpy as np
from numba import njit

dest_dtype = np.dtype([("array1", np.int16, (3,))], align=True)


@njit
def ex1(dest):
    tmp = (np.arange(3) + 1).astype(np.int16)
    dest['array1'] = tmp 


dest = np.empty(5, dtype=dest_dtype)
expected = np.empty(5, dtype=dest_dtype)
ex1(dest[0])
ex1.py_func(expected[0])

print(expected[0])
print(dest[0])
```

which gives:

```
$ python ex1.py 
([1, 2, 3],)
([1, 2, 3],)
```

with this PR.

gmarkall:
Thanks for looking at this so far! I hope you don't mind that I pushed a fix for the flake8 issues, just to make it a little clearer where the "true" failures are.

gmarkall:
Note that the failures are all variations on the same issue:

```
TypeError: Can only insert float at [0] in [6 x float]: got i8*

```

where `[6x float]` is the data for the nested array.

gmarkall:
@luk-f-a I have added the two reproducers we have for #7693 as tests as well - they seem to be passing for me locally.

luk-f-a:
> @luk-f-a I have added the two reproducers we have for #7693 as tests as well - they seem to be passing for me locally.

oh, I didn't see the message and added one too. I'll remove it later. For now, I'm trying to fix the tests that the previous push broke. Some of them are passing locally, I hope I've fixed all of them.

luk-f-a:
open question for a reviewer:
- is the `if` in function `record_setattr` the best way to product the special behaviour needed for nested arrays? That's as close to LLVM IR as I could take it. Alternatively, it could be done upstream from there, maybe here? https://github.com/numba/numba/blob/bf480b9e0da858a65508c2b17759a72ee6a44c51/numba/np/arrayobj.py#L2630-L2640

gmarkall:
gpuci run tests

gmarkall:
@luk-f-a Many thanks for the fixes - from a preliminary glance this is looking promising - do you consider this non-draft / ready for review now?

> * is the `if` in function `record_setattr` the best way to product the special behaviour needed for nested arrays?

I think it's as good as we can do given the way the `NestedArray` abstraction is - I don't see a problem with this pattern. It's also consistent with some of the other implementation from #7359.

> * Alternatively, it could be done upstream from there, maybe here?

I think it's better to keep it where it is because some code sets attrs on structured types rather than using `setitem`, and the current implementation will handle both cases.

luk-f-a:
thanks for the feedback, I'll remove the duplicated test and promote to non-draft

luk-f-a:
@gmarkall do the new tests need a cuda version?

gmarkall:
> @gmarkall do the new tests need a cuda version?

Ah, that's a good point - I'm not a sure if this woudl work with regualr arrays in CUDA, so I#ll have a quick check to see.

gmarkall:
gpuci run tests

gmarkall:
It turns out that the new tests do work in CUDA, so I've added them.

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall (approved):
@luk-f-a Many thanks for tracking down the issue and implementing the fix! I did find it quite hard to understand the implementation, as it was possibly a little unidiomatic - I've pushed a modification to it that makes the implementation use a `memcpy` to move the data from the RHS to the LHS - this avoids involving the data model, because it uses information already at hand during the lowering. Other than that, I think that the use of branching on the nested array / scalar case is perfectly fine, and is also symmetric with `record_getattr`. Are you comfortable with these changes? Please do let me know if you'd prefer any changes to the approach or if it needs some further explanation.

As a general observation, it looks like there's long been an asymmetry between `record_getattr` and `record_setattr`, which is why we had the initial manifestation of #7693 as a lowering error - `record_getattr` handled the nested array case, but `record_setattr` did not - now with this PR they are aligned, and I think this adds weight to the case that your approach of having a branch for the nested array case is the right one.

I also added a couple more tests to make sure that this worked with nested arrays of more than one dimension - although I expected it to, it doesn't hurt to be sure.

I'm now reasonably confident that this is good to go, but since I've had such a large hand in writing the code, someone else should also review it.

luk-f-a:
hi @gmarkall , I like the changes. It's a more direct way of doing it, and closer to what I had in mind when I started. Since I lacked experience on lowering, I limited myself to copying examples and ended up with something which didn't feel 100% optimal.
When I have more time later, I'm curious to see what LLVM IR is produced by this solution.

gmarkall:
This is some IR generated by it (diff of the IR before and after the lowering runs):

```diff
--- before.ll	2022-01-13 16:28:36.704435809 +0000
+++ after.ll	2022-01-13 16:29:08.932563321 +0000
@@ -42,6 +42,8 @@
   store i8* null, i8** %".246"
   %"excinfo.5" = alloca {i8*, i32, i8*}*
   store {i8*, i32, i8*}* null, {i8*, i32, i8*}** %"excinfo.5"
+  %".318" = alloca {i8*, i8*, i64, i64, i16*, [1 x i64], [1 x i64]}
+  store {i8*, i8*, i64, i64, i16*, [1 x i64], [1 x i64]} zeroinitializer, {i8*, i8*, i64, i64, i16*, [1 x i64], [1 x i64]}* %".318"
 B0:
   store {i8*, i8*, i64, i64, i64*, [1 x i64], [1 x i64]} zeroinitializer, {i8*, i8*, i64, i64, i64*, [1 x i64], [1 x i64]}* %".5"
   %".9" = call i32 @"_ZN5numba2np8arrayobj9np_arange12_3clocals_3e8impl_242B44c8tJTIeFCjyCbUFRqqOAK_2f6h0phxApMogijRBAA_3dEx27omitted_28default_3dNone_2927omitted_28default_3dNone_2927omitted_28default_3dNone_29"({i8*, i8*, i64, i64, i64*, [1 x i64], [1 x i64]}* %".5", {i8*, i32, i8*}** %"excinfo.1", i64 3)
@@ -399,6 +401,25 @@
   %"extracted.strides.3" = extractvalue {i8*, i8*, i64, i64, i64*, [1 x i64], [1 x i64]} %".155", 6
   %".316" = extractvalue [1 x i64] %"extracted.strides.3", 0
   call void @"NRT_decref"(i8* %"extracted.meminfo.3")
+  store {i8*, i8*, i64, i64, i16*, [1 x i64], [1 x i64]} zeroinitializer, {i8*, i8*, i64, i64, i16*, [1 x i64], [1 x i64]}* %".318"
+  store {i8*, i8*, i64, i64, i16*, [1 x i64], [1 x i64]} %".314", {i8*, i8*, i64, i64, i16*, [1 x i64], [1 x i64]}* %".318"
+  %".322" = getelementptr inbounds {i8*, i8*, i64, i64, i16*, [1 x i64], [1 x i64]}, {i8*, i8*, i64, i64, i16*, [1 x i64], [1 x i64]}* %".318", i32 0, i32 4
+  %".323" = load i16*, i16** %".322"
+  %".324" = getelementptr inbounds [6 x i8], [6 x i8]* %"arg.dest", i32 0, i32 0
+  %".325" = bitcast i8* %".324" to i16*
+  br label %"for.cond.2"
+for.cond.2:
+  %"loop.index.2" = phi i64 [0, %"for.end.1"], [%".333", %"for.body.2"]
+  %".327" = icmp slt i64 %"loop.index.2", 3
+  br i1 %".327", label %"for.body.2", label %"for.end.2"
+for.body.2:
+  %".329" = getelementptr i16, i16* %".325", i64 %"loop.index.2"
+  %".330" = getelementptr i16, i16* %".323", i64 %"loop.index.2"
+  %".331" = load i16, i16* %".330"
+  store i16 %".331", i16* %".329"
+  %".333" = add nsw i64 %"loop.index.2", 1
+  br label %"for.cond.2"
+for.end.2:
 }
 
 declare i32 @"_ZN5numba2np8arrayobj9np_arange12_3clocals_3e8impl_242B44c8tJTIeFCjyCbUFRqqOAK_2f6h0phxApMogijRBAA_3dEx27omitted_28default_3dNone_2927omitted_28default_3dNone_2927omitted_28default_3dNone_29"({i8*, i8*, i64, i64, i64*, [1 x i64], [1 x i64]}* noalias nocapture %"retptr", {i8*, i32, i8*}** noalias nocapture %"excinfo", i64 %"arg.start") 
```

for the Python example:

```python
import numpy as np
from numba import njit

dest_dtype = np.dtype([("array1", np.int16, (3,))], align=True)


@njit
def ex1(dest):
    tmp = (np.arange(3) + 1).astype(np.int16)
    dest['array1'] = tmp 


dest = np.empty(5, dtype=dest_dtype)
expected = np.empty(5, dtype=dest_dtype)
ex1(dest[0])
ex1.py_func(expected[0])

print(expected[0])
print(dest[0])
```

gmarkall:
I do wonder if it would be better to use a raw memcopy (I hadn't initially spotted that `cgutils.memcpy` generates a loop). If we did, that would look like this:

```llvm
  store {i8*, i8*, i64, i64, i16*, [1 x i64], [1 x i64]} zeroinitializer, {i8*, i8*, i64, i64, i16*, [1 x i64], [1 x i64]}* %".318"
  store {i8*, i8*, i64, i64, i16*, [1 x i64], [1 x i64]} %".314", {i8*, i8*, i64, i64, i16*, [1 x i64], [1 x i64]}* %".318"
  %".322" = getelementptr inbounds {i8*, i8*, i64, i64, i16*, [1 x i64], [1 x i64]}, {i8*, i8*, i64, i64, i16*, [1 x i64], [1 x i64]}* %".318", i32 0, i32 4
  %".323" = load i16*, i16** %".322"
  %".324" = getelementptr inbounds [6 x i8], [6 x i8]* %"arg.dest", i32 0, i32 0
  %".325" = bitcast i8* %".324" to i16*
  %".326" = bitcast i16* %".325" to i8*
  %".327" = bitcast i16* %".323" to i8*
  %".328" = mul i64 3, 2
  call void @"llvm.memcpy.p0i8.p0i8.i64"(i8* %".326", i8* %".327", i64 %".328", i1 0)
```

instead of the for-loops. The above could be generated by applying the following patch:

```diff
diff --git a/numba/np/arrayobj.py b/numba/np/arrayobj.py
index 5f25aa1dc..4c82a8e2f 100644
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -2602,10 +2602,11 @@ def record_setattr(context, builder, sig, args, attr):
         val_struct = cgutils.create_struct_proxy(valty)(context, builder,
                                                         value=args[1])
         src = val_struct.data
-        dest = cgutils.get_record_member(builder, target, offset,
-                                         src.type.pointee)
-        cgutils.memcpy(builder, dest, src,
-                       context.get_constant(types.intp, elemty.nitems))
+        llty = src.type.pointee
+        dest = cgutils.get_record_member(builder, target, offset, llty)
+        count = context.get_constant(types.intp, elemty.nitems)
+        itemsize = context.get_constant(types.intp, llty.width // 8)
+        cgutils.raw_memcpy(builder, dest, src, count, itemsize)
     else:
         # Set the given scalar record member
         dptr = cgutils.get_record_member(builder, target, offset,
```

A question for the review (e.g. @esc / @sklam / @stuartarchibald) - do you have a preference for one or the other?

numba/tests/test_record_dtype.py
```diff
@@ -1495,6 +1513,20 @@ def test_set_arrays(self):
             arr_res = cfunc(nbarr)
             np.testing.assert_equal(arr_res, arr_expected)
 
+    def test_setitem(self):
+        nbarr1 = np.recarray(2, dtype=recordwith2darray)
```
stuartarchibald, line 1517: Why are these size 2 if only the first index is assigned below, should they be size 1?

numba/tests/test_record_dtype.py
```diff
@@ -1495,6 +1513,20 @@ def test_set_arrays(self):
...
+        nbarr1[0] = np.array([(1, ((1, 2), (4, 5), (2, 3)))],
+                             dtype=recordwith2darray)[0]
+        nbarr2 = np.recarray(2, dtype=recordwith2darray)
+        nbarr2[0] = np.array([(10, ((10, 20), (40, 50), (20, 30)))],
+                             dtype=recordwith2darray)[0]
+        pyfunc = record_setitem_array
+        args = nbarr1[0], nbarr2[0]
+        arr_expected = pyfunc(*args)
```
stuartarchibald, line 1525: The function `record_setitem_array` doesn't return anything so the default return of `None` is used. As a result the check at the end of this test is just checking that `None` equals `None`?

stuartarchibald (commented):
Thanks for the patches @luk-f-a @gmarkall. I think there might be an issue in the testing and in updating the test locally to fix it this uncovers further issues. Please do take a look and see what you think. Thanks.

numba/tests/test_record_dtype.py
stuartarchibald: I think https://github.com/stuartarchibald/numba/commit/6042b7fbf1719de9136ca5acdfaaa7673f371308 fixes, feel free to incorporate.

numba/cuda/tests/cudapy/test_record_dtype.py
```diff
@@ -403,6 +419,20 @@ def test_record_read_2d_array(self):
...
+                             dtype=recordwith2darray)[0]
+        nbarr2 = np.recarray(2, dtype=recordwith2darray)
+        nbarr2[0] = np.array([(10, ((10, 20), (40, 50), (20, 30)))],
+                             dtype=recordwith2darray)[0]
+        pyfunc = record_setitem_array
+        args = nbarr1[0], nbarr2[0]
+        arr_expected = pyfunc(*args)
+        cfunc = cuda.jit(pyfunc)
+        arr_res = cfunc[1, 1](*args)
+        np.testing.assert_equal(arr_res, arr_expected)
```
stuartarchibald, line 434: This suffers from the same testing `None` == `None` issue as noted here: https://github.com/numba/numba/pull/7712/files#r789550428?

numba/np/arrayobj.py
```diff
@@ -2599,11 +2597,22 @@ def record_setattr(context, builder, sig, args, attr):
...
-    context.pack_value(builder, elemty, val, dptr, align=align)
+    if isinstance(elemty, types.NestedArray):
+        # Copy the data from the RHS into the nested array
+        val_struct = cgutils.create_struct_proxy(valty)(context, builder,
+                                                        value=args[1])
+        src = val_struct.data
+        dest = cgutils.get_record_member(builder, target, offset,
+                                         src.type.pointee)
+        cgutils.memcpy(builder, dest, src,
+                       context.get_constant(types.intp, elemty.nitems))
```
stuartarchibald, line 2608: I've checked this a bit locally, it seems like this is ok because the ctor for the record type copies in the data as C-contig when constructing the array from NumPy, and Numba correctly types the nested array as C-contig.

numba/cuda/tests/cudapy/test_record_dtype.py
gmarkall: Good point - I've just added a change modelled on your fix that resolves this.

numba/tests/test_record_dtype.py
gmarkall: Thanks for the fix, I've just incorporated it.

numba/tests/test_record_dtype.py
gmarkall: Now resolved using your fix.

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

stuartarchibald (approved):

stuartarchibald:
@gmarkall please could you review/approve too RE patch https://github.com/numba/numba/pull/7712/commits/364fe7ba06f06c39e5725aaf6edcea6a2c142400 from https://github.com/numba/numba/pull/7712#discussion_r791925854 as I'm author. Thanks.

gmarkall (approved):
Approved re patch 364fe7b, of which @stuartarchibald is the author.

stuartarchibald:
> Approved re patch [364fe7b](https://github.com/numba/numba/commit/364fe7ba06f06c39e5725aaf6edcea6a2c142400), of which @stuartarchibald is the author.

Thanks @gmarkall

esc:
`Build numba_smoketest_cuda_yaml_114 has started`

esc:
> `Build numba_smoketest_cuda_yaml_114 has started`

all green

stuartarchibald:
> > `Build numba_smoketest_cuda_yaml_114 has started`
> 
> all green

Thanks @esc

stuartarchibald:
Thanks for all your work on this @luk-f-a and @gmarkall, glad to see this going into 0.55.1!

PR #7722, opened 2022-01-11
Testhound/fp16 divide

testhound:
This PR adds support for fp16 divide and is part of the overall effort to add support for fp16 operations. This PR partly addresses the meta issue "add fp16 support": https://github.com/numba/numba/issues/4402.


testhound:
gpuci run tests

testhound:
gpuci run tests

gmarkall (changes requested):
This has a similar problem to #7556:

```python
from numba import cuda


@cuda.jit(device=True)
def test1(x, y):
    return cuda.fp16.hdiv(x, y)


@cuda.jit(device=True)
def test2(x, y):
    return cuda.fp16.hdiv(x, y)


@cuda.jit('void(float16[:], float16, float16, float16)')
def test(r, a, b, c):
    r[0] = test1(a, b) + test2(b, c)
```

gives:

```
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudadrv/driver.py", line 2682, in add_ptx
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudadrv/driver.py", line 319, in safe_cuda_api_call
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudadrv/driver.py", line 384, in _check_ctypes_error

  File "/home/gmarkall/numbadev/issues/7722/test.py", line 15, in <module>
    def test(r, a, b, c):
  File "/home/gmarkall/numbadev/numba/numba/cuda/decorators.py", line 99, in _jit
    return Dispatcher(func, [func_or_sig], targetoptions=targetoptions)
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 807, in __init__
    self.compile(sigs[0])
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 1095, in compile
    kernel.bind()
  File "/home/gmarkall/numbadev/numba/numba/cuda/compiler.py", line 468, in bind
  File "/home/gmarkall/numbadev/numba/numba/cuda/codegen.py", line 201, in get_cufunc
  File "/home/gmarkall/numbadev/numba/numba/cuda/codegen.py", line 174, in get_cubin
    linker.add_ptx(ptx.encode())
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudadrv/driver.py", line 2685, in add_ptx
ptxas application ptx input, line 101; error   : Duplicate definition of variable '__$temp1'
```

testhound:
gpuci run tests

numba/cuda/tests/cudapy/test_intrinsics.py
```diff
@@ -445,6 +458,37 @@ def test_hmul_ptx(self):
...
+        ary = np.zeros(1, dtype=np.float16)
+        arry1 = np.random.randint(-65504, 65505, size=500).astype(np.float16)
+        arry2 = np.random.randint(-65504, 65505, size=500).astype(np.float16)
+        for arg1, arg2 in np.nditer([arry1, arry2]):
+            compiled[1, 1](ary, arg1, arg2)
+            ref = arg1 / arg2
+            np.testing.assert_allclose(ary[0], ref)
+
+    @skip_unless_cc_53
+    def test_hdiv_unique_reg_predicate(self):
```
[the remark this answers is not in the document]
stuartarchibald: Further, is there value in checking for the predicate in the PTX?

numba/cuda/cudaimpl.py
```diff
@@ -492,6 +492,104 @@ def ptx_fp16_binary(context, builder, sig, args):
...
+                            ir.FloatType()))
+    fma_func = cgutils.get_or_insert_function(lmod, fnty, fname)
+    fma1 = builder.call(fma_func, [neg_f, fmul, arg2_fp32])
+    fma2 = builder.call(fma_func, [arg2_rcp, fma1, fmul])
+    fp32_to_f16 = builder.call(cvt2fp16_asm, [fma2])
+    builder.branch(true_bb)
+
+    #True branch
+    builder.position_at_start(true_bb)
+    phi_node = builder.phi(ir.IntType(16))
```
stuartarchibald, line 587: Does the `phi` need to be a 16 bit int?

docs/source/cuda-reference/kernel.rst
```diff
@@ -499,6 +499,13 @@ These functions return a 16-bit floating point result.
 
    Returns the 16-bit floating point result of the multiplication.
 
+.. function:: numba.cuda.fp16.hdiv (a, b)
+
+   Perform the divide operation ``a / b`` on 16-bit floating point arguments in
+   round to nearest mode.
```
stuartarchibald, line 505: Should this also note that it is implemented as multiply-by-reciprocal?

numba/cuda/stubs.py
```diff
@@ -629,6 +629,16 @@ class hmul(Stub):
 
         """
 
+    class hdiv(Stub):
+        """hdiv(a, b)
+
+        Perform fp16 division, (a / b) in round to nearest mode. Supported
+        on fp16 operands only.
```
stuartarchibald, line 636: As per the docs, note this is implemented as multiply-by-reciprocal?

stuartarchibald (commented):
Thanks for the patch, it's good to see the CUDA f16 operation support expanding! I've given this an initial review, comments are inline. @gmarkall please could you also take a look, especially at the use of inline asm/resulting PTX. Many thanks.

numba/cuda/cudaimpl.py
testhound: Yes it needs to be a 16-bit integer type because the Float model for fp16 is implemented as a 16-bit type:

@register_default(types.Float)
class FloatModel(models.PrimitiveModel):
        if fe_type == types.float16:
            be_type = ir.IntType(16)
        elif fe_type == types.float32:
            be_type = ir.FloatType()
        elif fe_type == types.float64:
            be_type = ir.DoubleType()
        else:
            raise NotImplementedError(fe_type)
        super(FloatModel, self).__init__(dmm, fe_type, be_type)

docs/source/cuda-reference/kernel.rst
testhound: I personally think this is an implementation detail that it is not necessary to specify.

testhound:
gpuci run tests

testhound:
gpuci run tests

numba/cuda/tests/cudapy/test_intrinsics.py
```diff
@@ -469,25 +477,60 @@ def test_hdiv_scalar(self):
...
-        arg2 = np.float16(1.57)
-        compiled[1, 1](ary, arg1, arg2)
-        ref = (arg1 / arg2) + (arg2 / arg1)
-        np.testing.assert_allclose(ary[0], ref)
+    def test_hdiv_denormal(self):
+        compiled = cuda.jit("void(f2[:], f2[:], f2[:])")(simple_hdiv_kernel)
+
+        # This test is designed to test divison of fp16 constants that
+        # result in small denormal numbers that should not be further
+        # iterated using the Newton Raphson algorithm used in the hdiv
```
gmarkall, line 496: It appears that `arry1 / arry2` is:

```
array([0.004738, 0.00912 , 0.010574, 0.0164  , 0.01787 , 0.01932 ,
       0.02078 , 0.02223 , 0.03244 , 0.0339  , 0.03537 , 0.03683 ,
       0.03827 , 0.03973 , 0.0412  , 0.04266 , 0.04413 , 0.04556 ,
       0.003685, 0.00841 , 0.009   , 0.00958 , 0.01548 , 0.0173  ,
       0.01843 , 0.01875 , 0.02008 , 0.02138 , 0.02727 , 0.03021 ,
       0.03116 ], dtype=float16)
```

Where do the small denormals appear?

numba/cuda/tests/cudapy/test_intrinsics.py
```diff
@@ -469,25 +477,60 @@ def test_hdiv_scalar(self):
...
-        ary = np.zeros(1, dtype=np.float16)
-        arg1 = np.float16(3.1415926)
-        arg2 = np.float16(1.57)
-        compiled[1, 1](ary, arg1, arg2)
-        ref = (arg1 / arg2) + (arg2 / arg1)
-        np.testing.assert_allclose(ary[0], ref)
+    def test_hdiv_denormal(self):
+        compiled = cuda.jit("void(f2[:], f2[:], f2[:])")(simple_hdiv_kernel)
+
+        # This test is designed to test divison of fp16 constants that
```
gmarkall, line 494: Which values are constants?

numba/cuda/tests/cudapy/test_intrinsics.py
```diff
@@ -469,25 +477,60 @@ def test_hdiv_scalar(self):
...
+        # result in small denormal numbers that should not be further
+        # iterated using the Newton Raphson algorithm used in the hdiv
+        # intrinisc. A smaple of values were extracted from the Nvidia hdiv
+        # test suite
+        arry1 = np.array([0x5b, 0xaf, 0xcb, 0x13b, 0x157, 0x173, 0x18f,
```
gmarkall, line 499: Why are these values written in hex?

numba/cuda/tests/cudapy/test_intrinsics.py
testhound: The denormal being checked occurs after the first iteration of the Newton Raphson (NR) algorithm; if a denormal is found the second iteration of the NR algorithm is skipped

numba/cuda/tests/cudapy/test_intrinsics.py
testhound: The constants in the the arrays 'arry1' and 'arry2'.

numba/cuda/tests/cudapy/test_intrinsics.py
testhound: That is how the constants were provided in the hdiv test suite.

numba/cuda/cudaimpl.py
```diff
@@ -492,6 +492,141 @@ def ptx_fp16_binary(context, builder, sig, args):
...
+    entry_bb = builder.block
+    fall_thru_bb = builder.append_basic_block()
+    true_bb = builder.append_basic_block()
+    builder.cbranch(or_cond, true_bb, fall_thru_bb)
+
+    # Fall through branch (Second Newton Rapson iteration)
+    builder.position_at_start(fall_thru_bb)
+
+    float_zero = context.get_constant(types.float32, 0.0)
+    neg_f = builder.fsub(float_zero, arg2_fp32)
```
isVoid, line 603: I think I would like to see this part moved to L617 because `-y` was not used till then. Also I would recommend naming the variables same as the math formula.

numba/cuda/cudaimpl.py
```diff
@@ -492,6 +492,141 @@ def ptx_fp16_binary(context, builder, sig, args):
...
+    fma_func = cgutils.get_or_insert_function(lmod, fnty, fname)
+
+    # Mathematically we have: q0 = x * 1/y * (1 - err)
+    # So err (correction_term) = x - y*q0 = fma(-y, q0, x)
+    # And q1 = err*yn + q0 = fma(err, yn, q0)
+
+    fma1 = builder.call(fma_func, [neg_f, fmul, arg2_fp32])
+    fma2 = builder.call(fma_func, [arg2_rcp, fma1, fmul])
+    fp32_to_f16 = builder.call(cvt2fp16_asm, [fma2])
+    builder.branch(true_bb)
```
isVoid, line 620: This should be moved to the next block of code (after the comment, that is).

numba/cuda/cudaimpl.py
```diff
@@ -492,6 +492,141 @@ def ptx_fp16_binary(context, builder, sig, args):
...
+
+    # Our objective is to get a quotient x/y, correctly rounded to nearest
+    # even in fp16 format, so we can do q0 = x * yn.
+
+    # We need to perform one more Newton iteration to eliminate the error
+    # from the multiplication.
+
+    # Mathematically we have: q0 = x * 1/y * (1 - err)
+    # So err (correction_term) = x - y*q0 = fma(-y, q0, x)
+    # And q1 = err*yn + q0 = fma(err, yn, q0)
```
isVoid, line 541: These comments seems out of place. They describe steps later in the process. The first part is conditional on if the result from the first iteration has denormals. And the second part is duplicate of what's mentioned below. I suggest we can remove this part.

numba/cuda/cudaimpl.py
```diff
@@ -492,6 +492,141 @@ def ptx_fp16_binary(context, builder, sig, args):
...
+    rcp_asm = ir.InlineAsm(rcp_fnty,
+                           "rcp.approx.ftz.f32 $0, $1;",
+                           "=f,f")
+    arg2_rcp = builder.call(rcp_asm, [arg2_fp32])
+
+    # Implement fp32 division via multiplication by
+    # reciprocal of arg2
+
+    # Our objective is to get a quotient x/y, correctly rounded to nearest
+    # even in fp16 format, so we can do q0 = x * yn.
```
isVoid, line 534: What's `yn`?

isVoid (commented):
I have some style-wise comments. I would suggest naming the result of each `builder.call` to their corresponding variable names used in the formula. Second, the pattern of declare function type, write inline asms, then call function is repeated several times. I suggest one of below:
- Abstract this pattern in to a helper function so each function calls can be reduced to 1 line, or
- Move all declare and inline asm to the top of the method, give them a sensible name. Then have all `builder.call`s in one place, each takes one line.

numba/cuda/tests/cudapy/test_intrinsics.py
```diff
@@ -445,6 +466,72 @@ def test_hmul_ptx(self):
...
+        args = (f2[:], f2, f2)
+        ptx, _ = compile_ptx(simple_hdiv_unique, args, cc=(5, 3))
+
+        # Look for two .reg .pred __$temp3 lines preceded by a '{'
+        # which disambiguates (via nesting) the variable names.
+
+        predicate_count = 0
+        for line in ptx.split('\n'):
+            if '{ .reg .pred __$temp3;' in line:
+                predicate_count += 1
```
isVoid, line 528: Can be simplified to:
```suggestion
        predicate_count = ptx.count('{ .reg .pred __$temp3;')
```

numba/cuda/tests/cudapy/test_intrinsics.py
```diff
@@ -445,6 +466,72 @@ def test_hmul_ptx(self):
...
+
+    @skip_unless_cc_53
+    def test_hdiv_denormal(self):
+        compiled = cuda.jit("void(f2[:], f2[:], f2[:])")(simple_hdiv_kernel)
+
+        # This test is designed to test divison of fp16 constants that
+        # result in small denormal numbers that should not be further
+        # iterated using the Newton Raphson algorithm used in the hdiv
+        # intrinisc. A smaple of values were extracted from the Nvidia hdiv
+        # test suite
```
isVoid, line 498:
```suggestion
        # intrinsics. A sample of values were extracted from the Nvidia hdiv
        # test suite
```

numba/cuda/cudaimpl.py
```diff
@@ -492,6 +492,141 @@ def ptx_fp16_binary(context, builder, sig, args):
 lower_fp16_binary(stubs.fp16.hmul, 'mul')
 
 
+@lower(stubs.fp16.hdiv, types.float16, types.float16)
+def lower_fp16_divide(context, builder, sig, args):
+    arg1 = args[0]
+    arg2 = args[1]
+    # Floating point 16 division algorithm implemented
+    # using Newton Rapson algorithm.
```
isVoid, line 500:
```suggestion
    # using Newton Raphson algorithm.
```

numba/cuda/cudaimpl.py
```diff
@@ -492,6 +492,141 @@ def ptx_fp16_binary(context, builder, sig, args):
...
+                                 "cvt.f32.f16 $0, $1;",
+                                 "=f,h")
+    arg1_fp32 = builder.call(cvt2float_asm, [arg1])
+    arg2_fp32 = builder.call(cvt2float_asm, [arg2])
+
+    rcp_fnty = ir.FunctionType(ir.FloatType(), [ir.FloatType()])
+
+    # Obtain a gross estimate of the reciprocal of argument 2 (y)
+    # which will be refined. This is the first step in the Newton
+    # Rapson method.
```
isVoid, line 519:
```suggestion
    # Raphson method.
```

numba/cuda/cudaimpl.py
testhound: @isVoid I don't understand this comment.

isVoid: Apologies, I wasn't aware this is the unconditional jump to the true branch. Then this line should stay where it is.

testhound:
gpuci run tests

numba/cuda/cudaimpl.py
```diff
@@ -493,6 +493,134 @@ def ptx_fp16_binary(context, builder, sig, args):
...
 
+def inline_asm_helper(builder, func_ret_ty, func_arg_types,
+                      asm_str, constraints_str, asm_args):
+    fnty = ir.FunctionType(func_ret_ty, func_arg_types)
+    asm = ir.InlineAsm(fnty, asm_str, constraints_str)
+    return builder.call(asm, asm_args)
+
+
+@lower(stubs.fp16.hdiv, types.float16, types.float16)
+def lower_fp16_divide(context, builder, sig, args):
```
isVoid, line 504:
```suggestion
def lower_fp16_divide(context, builder, sig, args):
    """Implements division between two fp16 using Newton Raphson Algorithm."""
```

numba/cuda/cudaimpl.py
```diff
@@ -493,6 +493,134 @@ def ptx_fp16_binary(context, builder, sig, args):
...
+    x, y = args
+
+    # Floating point 16 division algorithm implemented
+    # using Newton Raphson algorithm.
+
+    # x = arg1, y = arg2
+    # Implement x/y
```
isVoid, line 511:
```suggestion
```

numba/cuda/cudaimpl.py
```diff
@@ -493,6 +493,134 @@ def ptx_fp16_binary(context, builder, sig, args):
...
+    y_fp32 = inline_asm_helper(builder, ir.FloatType(), [ir.IntType(16)],
+                               asm, constr, [y])
+
+    # Obtain a gross estimate of the reciprocal of argument 2 (y)
+    # which will be refined. This is the first step in the Newton
+    # Raphson method.
+
+    # y0 = 1/y * (1-err), here y0 is not a precise 1/y
+    # it is an approximation and the term 'err' stands for the
+    # relative error of the approximation.
```
isVoid, line 530: I wish we can call the reciprocal of `y` not as `yn` but as something else... (perhaps `y_rcp`?)
```suggestion
    # Use fast approximation reciprocal to obtain an estimate of `y`s reciprocal `yn`.
    # yn=1/y * (1 - err), where `err` is the relative error term.
```

numba/cuda/cudaimpl.py
```diff
@@ -493,6 +493,134 @@ def ptx_fp16_binary(context, builder, sig, args):
...
+
+    yn = inline_asm_helper(builder, ir.FloatType(), [ir.FloatType()],
+                           "rcp.approx.ftz.f32 $0, $1;", "=f,f", [y_fp32])
+
+    # Implement fp32 division via multiplication by
+    # reciprocal of arg2
+
+    # Our objective is to get a quotient x/y, correctly rounded to nearest
+    # even in fp16 format, so we can do q0 = x * yn.
```
isVoid, line 539: I've had some difficulty understanding how to interpret this comment: is it "correctly rounded to nearest, even in fp16 format", or is it "correctly rounded nearest even (digit), in fp16 foramt"?

numba/cuda/cudaimpl.py
```diff
@@ -493,6 +493,134 @@ def ptx_fp16_binary(context, builder, sig, args):
...
+
+    # y0 = 1/y * (1-err), here y0 is not a precise 1/y
+    # it is an approximation and the term 'err' stands for the
+    # relative error of the approximation.
+
+    yn = inline_asm_helper(builder, ir.FloatType(), [ir.FloatType()],
+                           "rcp.approx.ftz.f32 $0, $1;", "=f,f", [y_fp32])
+
+    # Implement fp32 division via multiplication by
+    # reciprocal of arg2
```
isVoid, line 536:
```suggestion
    # reciprocal of y
```

numba/cuda/cudaimpl.py
```diff
@@ -493,6 +493,134 @@ def ptx_fp16_binary(context, builder, sig, args):
...
+                           (ir.FloatType(),
+                            ir.FloatType(),
+                            ir.FloatType()))
+    fma_func = cgutils.get_or_insert_function(lmod, fnty, fname)
+
+    # We need to perform one more Newton iteration to eliminate the error
+    # from the multiplication.
+    #
+    # Mathematically we have: q0 = x * 1/y * (1 - err)
+    # So err (correction_term) = x - y*q0 = fma(-y, q0, x)
```
isVoid, line 603: The variable `err` here means something different to the previous formula. If they mean different thing then it shouldn't share the same name. Probably `correction_term` is better naming.

numba/cuda/cudaimpl.py
```diff
@@ -493,6 +493,134 @@ def ptx_fp16_binary(context, builder, sig, args):
...
+    # And q1 = err*yn + q0 = fma(err, yn, q0)
```
isVoid, line 604: From the formula above, `correction_term=x*err`, I could not quite figure out why `correction_term * yn + q0` gives the final result.

isVoid (commented):
Possibly due to my limited knowledge on Newton-Raphson Division algorithm, I looked it up from wikipedia: https://en.wikipedia.org/wiki/Division_algorithm#Newton%E2%80%93Raphson_division. The Newtion's method is applied only when computing reciprocals. In this PR, the reciprocals was computed with an inline asm instruction. It feels like the use of "iteration" isn't a standard usage. Perhaps calling the algorithm as a 2-stage algorithm: "compute and correct error" is more appropriate?

numba/cuda/cudaimpl.py
testhound: No this is the same 'err' variable. This comment spells out it's derivation. I can drop the use of correction_term to avoid confusion.

isVoid: I think the term on the left is `x * err` (deriving from the first formula). Replacing this with the first term in the next formula:
`x * y_n * err` gives an estimation of the correction term. Adding this to `q0` should bring the result closer to the actual value. Because `q0 = x * 1/y * (1 - err) = x * 1/y - x * 1/y * err`, the second term is the error term that requires correction. We don't have the actual `1/y` but we do have the estimate of it `yn`. Using that to correct should reduce the error by an order of magnitude.

isVoid: I guess this is why it's called an "iteration" of newton's method because the net result of an iteration is reducing the error `epsilon` to `epsilon^2`?

testhound: Actually @isVoid you were correct, I am about to push another change which should hopefully clear this up.

testhound:
gpuci run tests

numba/cuda/cudaimpl.py
isVoid: Thanks. I think this should eliminate the ambituity.

numba/cuda/cudaimpl.py
```diff
@@ -493,6 +493,125 @@ def ptx_fp16_binary(context, builder, sig, args):
...
+                      asm_str, constraints_str, asm_args):
+    fnty = ir.FunctionType(func_ret_ty, func_arg_types)
+    asm = ir.InlineAsm(fnty, asm_str, constraints_str)
+    return builder.call(asm, asm_args)
+
+
+@lower(stubs.fp16.hdiv, types.float16, types.float16)
+def lower_fp16_divide(context, builder, sig, args):
+    x, y = args
+    """Implements division between two fp16 using Newton Raphson Algorithm."""
```
isVoid, line 506:
```suggestion
def lower_fp16_divide(context, builder, sig, args):
    """Implements division between two fp16 using Newton Raphson Algorithm."""
    x, y = args
```

isVoid (approved):
Thanks, just having one small issue with docstring above. I think the N-R methods isn't implemented in canonical way here because we want to utilize hardware accelerator like fast approximate reciprocal and fma modules etc. But it seems like the convergence rate is the same as N-R method, see: https://github.com/numba/numba/pull/7722#discussion_r813559853. More comments here are welcome.

testhound:
gpuci run tests

gmarkall:
I'm conscious that a lot of work has already gone into this PR, but I think we should put it on hold until #8001 has gone through - it seems that it might be more appropriate to use a similar mechanism to reuse the implementation in the C++ headers rather than re-implementing it in Numba (assuming that this approach proves to work well in #8001, which I'm hopeful it will).

testhound:
> I'm conscious that a lot of work has already gone into this PR, but I think we should put it on hold until #8001 has gone through - it seems that it might be more appropriate to use a similar mechanism to reuse the implementation in the C++ headers rather than re-implementing it in Numba (assuming that this approach proves to work well in #8001, which I'm hopeful it will).

@gmarkall I added support for fp16 divide to the math operator branch: https://github.com/numba/numba/pull/8001

gmarkall:
> @gmarkall I added support for fp16 divide to the math operator branch: https://github.com/numba/numba/pull/8001

Many thanks - I'll mark this as "abandoned" for now as it's superseded by that.

gmarkall:
Closing as abandoned in favour of #8001.

PR #7724, opened 2022-01-12
Update URLs in error messages to refer to RTD docs.

stuartarchibald:
As title. This replaces references to documentation at
numba.pydata.org with numba.readthedocs.io. It also fixes URLs
that were invalid.


gmarkall (approved):
All new URLs tested and working - LGTM!

Issue #7725, opened 2022-01-12
Raise a error when an array of the wrong types are used for function compiled AOT


ChiCheng45:
We have some functions we want to compile AOT using numba which uses arrays as its input. In some cases users of our code may accidentally use an array of float64 instead of int64 for example. In these cases the numba code appears to continue and does not raise an error so that the results will be wrong. Is there a something we should be doing to ensure that the user uses the right types?

See the minimal example below,

```
from numba.pycc import CC
from numba import complex64, float64, int64
from numba.core.types.npytypes import Array


cc = CC('my_module')


@cc.export('multc', Array(complex64, 1, "A")(
    Array(complex64, 1, "A"), Array(complex64, 1, "A")
))
@cc.export('multf', Array(float64, 1, "A")(
    Array(float64, 1, "A"), Array(float64, 1, "A")
))
@cc.export('multi', Array(int64, 1, "A")(
    Array(int64, 1, "A"), Array(int64, 1, "A")
))
def mult(a, b):
    return a * b


if __name__ == "__main__":
    cc.compile()
```

```
>>> import numpy as np
>>> a = np.array([1, 2, 3, 4], dtype=np.int64)
>>> b = np.array([1.1, 2.2, 3.3, 4.4], dtype=np.float64)
>>> c = np.array([1.1+1.1j, 2.2+2.2j, 3.3+3.3j, 4.4+4.4j], dtype=np.complex64)
>>> import my_module
>>> my_module.multi(a, a)
array([ 1,  4,  9, 16])
>>> my_module.multi(b, b)   # should raise a error
array([-8119809984163909468, -4426858289720102748,  6638485994729205924,
        -733906595276296028])
>>> my_module.multi(c, c)    # should raise a error
array([ 2586564337444281385, -4745352150906872791, -8446381298434483159,
        6369616171939879977])
>>> my_module.multf(a, a)     # should raise a error
array([0., 0., 0., 0.])
>>> my_module.multf(b, b)
array([ 1.21,  4.84, 10.89, 19.36])
>>> my_module.multf(c, c)    # should raise a error
array([1.97754000e-04, 1.29600062e+01, 5.89824048e+03, 8.49346967e+05])
>>> my_module.multc(a, a)    # should raise a error
array([0.+0.j, 0.+0.j, 0.+0.j, 0.+0.j], dtype=complex64)
>>> my_module.multc(b, b)    # should raise a error
[...]

PR #7728, opened 2022-01-12
Document that AOT-compiled functions do not check arg types

gmarkall:
This explicitly documents the limitation described in #7725, that AOT-compiled functions don't check their argument types.

### DIFF
--- a/docs/source/user/pycc.rst
+++ b/docs/source/user/pycc.rst
@@ -39,7 +39,7 @@ Limitations
    several different signatures under different names).
 
 #. Exported functions do not check the types of the arguments that are passed
-   to them; the callee is expected to provide arguments of the correct type.
+   to them; the caller is expected to provide arguments of the correct type.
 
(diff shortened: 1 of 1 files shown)


docs/source/user/pycc.rst
```diff
@@ -38,6 +38,9 @@ Limitations
 #. Each exported function can have only one signature (but you can export
    several different signatures under different names).
 
+#. Exported functions do not check the types of the arguments that are passed
+   to them; the callee is expected to provide arguments of the correct type.
```
stuartarchibald, line 42:
```suggestion
   to them; the caller is expected to provide arguments of the correct type.
```
is it not that the caller of the AOT function has to ensure the data is of suitable type for use as arguments?

gmarkall: Doh!

stuartarchibald (approved):

PR #7729, opened 2022-01-12
Handle Omitted/OmittedArgDataModel in DI generation.

### LINKED ISSUE #7726: DWARF generation failing for OmittedArgs
## Reporting a bug

- [x] I have tried using the latest RC of Numba (0.55rc1)

This:
```python
from numba import njit
import numpy as np

@njit
def foo():
    x = np.arange(5.)

foo()
```
run like this:
```
NUMBA_DEBUGINFO=1 python omitted1.py
```
does this:
```
numba/core/datamodel/models.py", line 43, in get_value_type
    raise NotImplementedError(self)
NotImplementedError: <numba.core.datamodel.models.OmittedArgDataModel object at 0x7f2a5e883df0>
```

stuartarchibald:
As title.

Fixes #7726

### DIFF
--- a/numba/core/debuginfo.py
+++ b/numba/core/debuginfo.py
@@ -422,8 +422,12 @@ class DIBuilder(AbstractDIBuilder):
 
         for idx, (name, nbtype) in enumerate(argmap.items()):
             name = name.replace('.', '$')    # for gdb to work correctly
-            datamodel = self.cgctx.data_model_manager[nbtype]
-            lltype = self.cgctx.get_value_type(nbtype)
+            use_ty = nbtype
+            if isinstance(nbtype, types.Omitted):
+                resolver = self.cgctx.typing_context.resolve_value_type
+                use_ty = resolver(nbtype.value)
+            datamodel = self.cgctx.data_model_manager[use_ty]
+            lltype = self.cgctx.get_value_type(use_ty)
             size = self.cgctx.get_abi_sizeof(lltype)
             mdtype = self._var_type(lltype, size, datamodel=datamodel)
             md.append(mdtype)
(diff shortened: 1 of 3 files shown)


gmarkall:
@stuartarchibald Is the ready for review now (were you just waiting for CI?)

I was initially a little tentative about the idea of the being in 0.55RC2, but since the change is limited to debuginfo I'm comfortable with the risk (because I believe it's extremely low).

numba/core/debuginfo.py
```diff
@@ -422,8 +422,12 @@ def _di_subroutine_type(self, line, function, argmap):
 
         for idx, (name, nbtype) in enumerate(argmap.items()):
             name = name.replace('.', '$')    # for gdb to work correctly
-            datamodel = self.cgctx.data_model_manager[nbtype]
-            lltype = self.cgctx.get_value_type(nbtype)
+            use_ty = nbtype
+            if isinstance(nbtype, types.Omitted):
+                resolver = self.cgctx.typing_context.resolve_value_type
+                use_ty = resolver(nbtype.value)
```
sklam, line 428: hm... Maybe `types.Omitted` should use some dummy type for the value model

stuartarchibald: Could just use an `ir.IntType(8).as_pointer()` or were you thinking more along the lines of creating a special dummy type?

sklam: `i8*` works just fine. other choices are a empty struct (`llvmlite.ir.LiteralStructType([])`.

stuartarchibald:
I think the failing test is this https://github.com/numba/numba/pull/7764#issuecomment-1022140110

PR #7737, opened 2022-01-13
[WIP] Tidy up memmove/memcpy APIs

gmarkall:
PR to test on CI as I'm working; description to be updated when the PR is ready.

### DIFF
--- a/numba/core/cgutils.py
+++ b/numba/core/cgutils.py
@@ -1034,7 +1034,7 @@ def memcpy(builder, dst, src, count):
         builder.store(builder.load(in_ptr), out_ptr)
 
 
-def _raw_memcpy(builder, func_name, dst, src, count, itemsize, align):
+def _raw_memcpy(builder, func_name, dst, src, count, itemsize):
     size_t = count.type
     if isinstance(itemsize, int):
         itemsize = ir.Constant(size_t, itemsize)
@@ -1048,21 +1048,21 @@ def _raw_memcpy(builder, func_name, dst, src, count, itemsize, align):
                           is_volatile])
 
 
-def raw_memcpy(builder, dst, src, count, itemsize, align=1):
+def raw_memcpy(builder, dst, src, count, itemsize):
     """
     Emit a raw memcpy() call for `count` items of size `itemsize`
     from `src` to `dest`.
     """
-    return _raw_memcpy(builder, 'llvm.memcpy', dst, src, count, itemsize, align)
+    return _raw_memcpy(builder, 'llvm.memcpy', dst, src, count, itemsize)
 
 
-def raw_memmove(builder, dst, src, count, itemsize, align=1):
+def raw_memmove(builder, dst, src, count, itemsize):
     """
     Emit a raw memmove() call for `count` items of size `itemsize`
     from `src` to `dest`.
     """
     return _raw_memcpy(builder, 'llvm.memmove', dst, src, count,
-                       itemsize, align)
+                       itemsize)
 
 
 def muladd_with_overflow(builder, a, b, c):
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -4227,7 +4227,7 @@ def _array_copy(context, builder, sig, args):
     if arytype.layout == rettype.layout:
         # Fast path: memcpy
         cgutils.raw_memcpy(builder, dest_data, src_data, ary.nitems,
-                           ary.itemsize, align=1)
+                           ary.itemsize)
 
     else:
         src_strides = cgutils.unpack_tuple(builder, ary.strides)


gmarkall:
gpuci run tests

gmarkall:
The gpuCI fail is unrelated to this PR and is only the result of an unexpected success due to the latest CUDA Python bindings release fixing an upstream issue (profiling APIs are now available) - I'll be making a separate PR to fix this. Otherwise, all tests pass so this is ready for review now.

gmarkall:
gpuci run tests

numba/core/cgutils.py
```diff
@@ -1013,56 +1014,59 @@ def cbranch_or_continue(builder, cond, bbtrue):
...
     """
-    # Note this does seem to be optimized as a raw memcpy() by LLVM
-    # whenever possible...
-    assert dst.type == src.type
-    with for_range(builder, count, intp=count.type) as loop:
-        out_ptr = builder.gep(dst, [loop.index])
-        in_ptr = builder.gep(src, [loop.index])
-        builder.store(builder.load(in_ptr), out_ptr)
+    if dst.type != src.type:
+        msg = f'memcpy requires the same types; got {dst.type} and {src.type}'
```
stuartarchibald, line 1030:
```suggestion
        msg = f'memcpy requires the same types; got source type {src.type} and destination type {dst.type}'
```
would it be useful to specify which way around these are?

numba/core/cgutils.py
```diff
@@ -1013,56 +1014,59 @@ def cbranch_or_continue(builder, cond, bbtrue):
...
+        raise TypingError(msg)
```
stuartarchibald, line 1031: Is this a typing problem? Think this should be a lowering error, the caller has passed in invalid llvm types?

numba/core/cgutils.py
```diff
@@ -1013,56 +1014,59 @@ def cbranch_or_continue(builder, cond, bbtrue):
...
-    builder.call(memcpy, [builder.bitcast(dst, voidptr_t),
-                          builder.bitcast(src, voidptr_t),
-                          builder.mul(count, itemsize),
-                          is_volatile])
+def memmove(builder, dst, src, count, itemsize=1):
+    """
+    Emit a memmove to the builder.
 
+    Source and destination types can be any type, but the types must be the
+    same.
```
stuartarchibald, line 1041: Should this contain a similar check for source type matching dest type as present in the `memcpy` impl above?

numba/core/cgutils.py
```diff
@@ -1013,56 +1014,59 @@ def cbranch_or_continue(builder, cond, bbtrue):
...
+    if isinstance(count, int):
+        count = ir.Constant(intp_t, count)
+
+    if isinstance(itemsize, int):
+        itemsize = ir.Constant(intp_t, itemsize)
+
+    if isinstance(itemsize, ir.Constant) and itemsize.constant == 1:
+        length = count
+    else:
+        length = builder.mul(count, itemsize)
```
stuartarchibald, line 1061: I think LLVM would just optimise a multiple by constant one away, but think it's fine to leave this as-is too.

numba/core/cgutils.py
```diff
@@ -1013,56 +1014,59 @@ def cbranch_or_continue(builder, cond, bbtrue):
...
+
+    memcpy = builder.module.declare_intrinsic(func_name,
```
stuartarchibald, line 1063:
```suggestion
    func_impl = builder.module.declare_intrinsic(func_name,
```
(and refactor, this interface is intended to be generic for `memcpy` and `memmove`)?

numba/core/cgutils.py
stuartarchibald: Also, should this be tested?

stuartarchibald (commented):
Thanks for undertaking this refactoring. The interface is a lot more consistent with the signatures and also with the functions themselves now! I've left a few small comments to address inline but otherwise looks good. Thanks again!

numba/core/cgutils.py
gmarkall: Now I've looked into it, there's no requirement in LLVM for the types to be the same (https://llvm.org/docs/LangRef.html#llvm-memcpy-intrinsic) - this was only needed by the old hand-rolled implementation. In this PR I added the check to enforce the description in the docstring, but I now see it's not necessary I'll remove it from this PR.

numba/core/cgutils.py
gmarkall: No, this implementation was OK (although the docstring is wrong, and will be updated).

numba/core/cgutils.py
gmarkall: Now you mention it I do think it would be better to have simpler code here, so I'll make it just always generate the mul instead of trying to be clever.

numba/core/cgutils.py
gmarkall: Will refactor to make this generic.

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
@stuartarchibald Many thanks for the review and comments. Everything is fixed up from your review, but as can be seen from gpuCI there's a new issue from things that were added to main since I made this PR that I've not yet resolved. Annoyingly it passes on my machines, but only fails on gpuCI so it's taking a little longer than I hoped to resolve this.

gmarkall:
gpuci run tests

gmarkall:
OK, I got to the bottom of this issue. Factors involved in the problem / fix are:

- This PR replaces all uses of `memcpy` in Numba with calls to the `@llvm.memcpy` intrinsic.
- The `@llvm.memcpy` intrinsic seems to assume its arguments are aligned to the alignment of an `i8`, which varies across architectures. (The docs are quite vague about this, they do not specify the assumption when the alignment is not specified - https://llvm.org/docs/LangRef.html#llvm-memcpy-intrinsic)
- We need to specify alignment of the source and destination parameters to prevent this assumption.
- In order to specify alignment, we need to add the `align` parameter attribute to the source and destination parameters.
- llvmlite does not support the `align` parameter attribute.
- The form of `memcpy` and `memcmp` differs between LLVM 3.4 and 7.0, so for CUDA Toolkits less than 11.2, an additional transformation needs adding to the NVVM IR to translate the parameter alignment attributes into an additional parameter (it looks like no code has ever generated `@llvm.memcpy` in conjunction with the CUDA target before.

So, the next steps to get this finished are:

- Add support for the `align` parameter attribute in llvmlite.
- Modify the implementation of `_raw_mem_intrinsic` to add alignment parameter attributes to the source and destination parameters.
- Add a transform to rewrite the `@llvm.memcpy` intrinsic as-needed for NVVM 3.4.

gmarkall:
Rather than attempt to fix support for NVVM 3.4, instead I'll wait to revisit this when NVVM 3.4 support is dropped.

PR #7740, opened 2022-01-14
CUDA Python 11.6 support

gmarkall:
CUDA Python now supports the profiler APIs (as of 11.6), so the xfail needs removing.

Description to be updated once this PR is ready for review

gpuci run tests

### DIFF
--- a/buildscripts/condarecipe.local/meta.yaml
+++ b/buildscripts/condarecipe.local/meta.yaml
@@ -57,7 +57,7 @@ requirements:
     # scipy 1.0 or later
     - scipy >=1.0
     # CUDA Python 11.6 or later
-    - cuda-python >= 11.6
+    - cuda-python >=11.6
 
 test:
   requires:
--- a/buildscripts/condarecipe.local/meta.yaml
+++ b/buildscripts/condarecipe.local/meta.yaml
@@ -56,6 +56,8 @@ requirements:
     - cudatoolkit >=9.2
     # scipy 1.0 or later
     - scipy >=1.0
+    # CUDA Python 11.6 or later
+    - cuda-python >= 11.6
 
 test:
   requires:
--- a/docs/source/cuda/bindings.rst
+++ b/docs/source/cuda/bindings.rst
@@ -4,12 +4,7 @@ CUDA Bindings
 Numba supports two bindings to the CUDA Driver APIs: its own internal bindings
 based on ctypes, and the official `NVIDIA CUDA Python bindings
 <https://nvidia.github.io/cuda-python/>`_. Functionality is equivalent between
-the two bindings, with two exceptions:
-
-* the NVIDIA bindings presently do not support Per-Thread Default Streams
-  (PTDS), and an exception will be raised on import if PTDS is enabled along
-  with the NVIDIA bindings.
-* The profiling APIs are not available with the NVIDIA bindings.
+the two bindings.
 
 The internal bindings are used by default. If the NVIDIA bindings are installed,
 then they can be used by setting the environment variable
@@ -17,15 +12,32 @@ then they can be used by setting the environment variable
 Numba has been imported, the selected binding cannot be changed.
 
 
+Per-Thread Default Streams
+--------------------------
+
+Responsibility for handling Per-Thread Default Streams (PTDS) is delegated to
+the NVIDIA bindings when they are in use. To use PTDS with the NVIDIA bindings,
+set the environment variable ``CUDA_PYTHON_CUDA_PER_THREAD_DEFAULT_STREAM`` to
+``1`` instead of Numba's environmnent variable
+:envvar:`NUMBA_CUDA_PER_THREAD_DEFAULT_STREAM`.
+
+.. seealso::
+
+   The `Default Stream section
+   <https://nvidia.github.io/cuda-python/release/11.6.0-notes.html#default-stream>`_
+   in the NVIDIA Bindings documentation.
+
+
 Roadmap
 -------
 
+In Numba 0.56, the NVIDIA Bindings will be used by default, if they are
+installed.
+
 In future versions of Numba:
 
-- The NVIDIA Bindings will be used by default, if they are installed.
 - The internal bindings will be deprecated.
 - The internal bindings will be removed.
 
-It is expected that the NVIDIA bindings will be the default in Numba 0.56; at
-present, no specific release is planned for the deprecation or removal of the
-internal bindings.
+At present, no specific release is planned for the deprecation or removal of
+the internal bindings.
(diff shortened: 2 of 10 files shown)


gmarkall:
gpuci run tests

gmarkall:
To do:

- Add documentation specifying that the minimum CUDA Python version is 11.6
- Skip instead of xfailing the PTDS test on CUDA Python (because the responsibility for setting up PTDS is pushed down to CUDA Python and out of Numba :tada:)

jakirkham:
cc @vzhurba01 (for awareness)

stuartarchibald:
Note: this is probably going to need a `run_constrained` on the Nvidia bindings.

gmarkall:
gpuci run tests

docs/source/cuda/bindings.rst
```diff
@@ -4,28 +4,40 @@ CUDA Bindings
...
 
 
+Per-Thread Default Streams
+--------------------------
+
+Responsibility for handling Per-Thread Default Streams (PTDS) is delegated to
+the NVIDIA bindings when they are in use. To use PTDS with the NVIDIA bindings,
+set the environment variable ``CUDA_PYTHON_CUDA_PER_THREAD_DEFAULT_STREAM`` to
+``1`` instead of Numba's environmnent variable
+:envvar:`NUMBA_CUDA_PER_THREAD_DEFAULT_STREAM`.
```
stuartarchibald, line 22: If both are set by accident, what happens?

stuartarchibald (commented):
Thanks for the patch, one minor comment else looks good. I'll manually verify the behaviours with the PTDS configuration at a later point. Thanks again.

docs/source/cuda/bindings.rst
gmarkall: That may not be an accident - suppose you're using RMM (which always uses the NVIDIA bindings) and Numba with the ctypes bindings, and you want both to use PTDS, then that's what you should do. If you try to use `NUMBA_CUDA_PER_THREAD_DEFAULT_STREAM` when Numba is using the NVIDIA bindings, you will get a warning though, because that won't have the effect you want.

stuartarchibald: Many thanks for the clarification.

numba/core/config.py
```diff
@@ -135,7 +135,11 @@ def validate(self):
                 CUDA_USE_NVIDIA_BINDING = False
 
             if CUDA_PER_THREAD_DEFAULT_STREAM:  # noqa: F821
-                warnings.warn("PTDS is not supported with CUDA Python")
+                warnings.warn("PTDS support is handled by CUDA Python when "
+                              "using the NVIDIA binding. Please set the "
+                              "environment variable "
+                              "CUDA_PYTHON_CUDA_PER_THREAD_DEFAULT_STREAM to 1 "
+                              "instead.")
```
stuartarchibald, line 142: I've manually tested that this warning appears correctly. Two questions 1. is it worth adding a test for it to the test suite 2. Should it be a warning of class `RuntimeWarning`?

buildscripts/condarecipe.local/meta.yaml
```diff
@@ -56,6 +56,8 @@ requirements:
     - cudatoolkit >=9.2
     # scipy 1.0 or later
     - scipy >=1.0
+    # CUDA Python 11.6 or later
+    - cuda-python >= 11.6
```
stuartarchibald, line 60:
```suggestion
    - cuda-python >=11.6
```

docs/source/cuda/bindings.rst
```diff
@@ -4,28 +4,40 @@ CUDA Bindings
 Numba supports two bindings to the CUDA Driver APIs: its own internal bindings
 based on ctypes, and the official `NVIDIA CUDA Python bindings
 <https://nvidia.github.io/cuda-python/>`_. Functionality is equivalent between
-the two bindings, with two exceptions:
-
-* the NVIDIA bindings presently do not support Per-Thread Default Streams
-  (PTDS), and an exception will be raised on import if PTDS is enabled along
-  with the NVIDIA bindings.
-* The profiling APIs are not available with the NVIDIA bindings.
+the two bindings.
```
stuartarchibald, line 7: Side note... should this section on bindings refer to the cuda overview doc on how to install/obtain the NVIDIA bindings? As part of review I looked at the rendered RTD docs and thought perhaps this section ought to link through to how to get them.

docs/source/reference/envvars.rst
```diff
@@ -485,11 +485,20 @@ GPU support
...
    When set to 1, the default stream is the per-thread default stream. When set
    to 0, the default stream is the legacy default stream. This defaults to 0,
-   for the legacy default stream. It may default to 1 in a future release of
-   Numba. See `Stream Synchronization Behavior
+   for the legacy default stream. See `Stream Synchronization Behavior
    <https://docs.nvidia.com/cuda/cuda-runtime-api/stream-sync-behavior.html>`_
    for an explanation of the legacy and per-thread default streams.
 
+   This variable only takes effect when using Numba's internal CUDA bindings;
+   when using the NVIDIA bindings, use the variable
```
stuartarchibald, line 493:
```suggestion
   when using the NVIDIA bindings, use the environment variable
```

docs/source/cuda/bindings.rst
gmarkall: There's a link in the first sentence to the docs for CUDA Python, which begins with how to install them: https://nvidia.github.io/cuda-python/

numba/core/config.py
gmarkall: 1. This will be temporary, PTDS is presently a pretty niche use case, and I don't expect the code around this to change much, so I was comfortable without the test - my estimation was that the time saved by adding it wouldn't exceed the time consumed implementing it. I can add one if you would really like one though.
2. Most of the other warnings in `config.py` are not - is there a reason it should be?

stuartarchibald: 1. Thanks, I agree RE effort given the longevity of the change and have manually tested too. These comments record the reasoning :)
2. I don't think so, was just wondering if you had a preference, let's leave as-is. The warning classes are a little inconsistent in `config.py` and should probably be all the same/something special to Numba to permit filtering.

docs/source/cuda/bindings.rst
stuartarchibald: Yes :) I found that after writing the comment! I'm not sure it's evident that the installation instructions are at the linked location but am happy to leave for now.

stuartarchibald (approved):
Thanks for the patch and fixes. This looks good. I think this needs a run through gpuci and the Numba build farm, scheduling the latter now.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_112`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_112`.

Passed.

gmarkall:
gpuci run tests

PR #7744, opened 2022-01-14
Fix issues with locating/parsing source during DebugInfo emission.

### LINKED ISSUE #7730: "Unexpected" indentation can break DI emission
## Reporting a bug

- [x] I have tried using the RC1 version of Numba 0.55

This:
```python
from numba import njit
import numpy as np

def gen():
    @njit(debug=True)
    def foo():
# comment
        impl = 1
        return impl
    return foo

gen()()
```
does this:
```
numba/core/lowering.py", line 112, in _compute_def_location
    src_ast = ast.parse(source_str)
  File "<pypath>/lib/python3.9/ast.py", line 50, in parse
    return compile(source, filename, mode, flags,
  File "<unknown>", line 1
    @njit(debug=True)
IndentationError: unexpected indent
```
the issue is the `#comment` at a different indent level.

stuartarchibald:
This works around issues with source that cannot be parsed when
looking for a line containing `def` so as to get the correct line
info for use in DebugInfo emission. It adds tests for this and
also the case where the function source is simply unobtainable.

Fixes #7730

### DIFF
--- a/numba/tests/test_debuginfo.py
+++ b/numba/tests/test_debuginfo.py
@@ -3,12 +3,16 @@ import inspect
 import re
 import numpy as np
 import math
+from textwrap import dedent
+import unittest
+import warnings
 
-from numba.tests.support import TestCase, override_config, needs_subprocess
+from numba.tests.support import (TestCase, override_config, needs_subprocess,
+                                 ignore_internal_warnings)
 from numba import jit, njit
 from numba.core import types, utils
 from numba.core.datamodel import default_manager
-import unittest
+from numba.core.errors import NumbaDebugInfoWarning
 import llvmlite.binding as llvm
 
 #NOTE: These tests are potentially sensitive to changes in SSA or lowering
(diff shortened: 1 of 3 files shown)


numba/tests/test_debuginfo.py
```diff
@@ -595,6 +599,50 @@ def foo():
...
+            foo()
+
+        self.assertEqual(len(w), 1)
+        found = w[0]
+        self.assertEqual(found.category, NumbaDebugInfoWarning)
+        msg = str(found.message)
+        # make sure the warning contains the right message
+        self.assertIn('Could not find source for function', msg)
+        # and refers to the offending function
+        self.assertIn(str(foo.py_func), msg)
```
esc, line 623: will it make sense to check that the last line of the warning is: "Debug line information may be inaccurate."?

numba/tests/test_debuginfo.py
```diff
@@ -595,6 +599,50 @@ def foo():
...
+            foo()
+
+        self.assertEqual(len(w), 1)
+        found = w[0]
+        self.assertEqual(found.category, NumbaDebugInfoWarning)
+        msg = str(found.message)
+        # make sure the warning contains the right message
+        self.assertIn('Could not parse the source for function', msg)
+        # and refers to the offending function
+        self.assertIn(str(foo.py_func), msg)
```
esc, line 644: The warning for indentation issues is slightly different. Do we want to check here that something like `that is at a lesser indent level than the source.` is contained within the message too?

esc (dismissed):
Two suggestions for slightly improved testing. Looks good otherwise.

esc (dismissed):
LGTM!

stuartarchibald:
776f5df fixes conflicts against mainline.

esc (approved):
Re-approved.

PR #7749, opened 2022-01-17
Refactor threading layer priority tests to not use stdout/stderr

### LINKED ISSUE #7458: Refactor unit tests scraping stdout/stderr in cases where alternatives are available.
As noted here:
https://github.com/numba/numba/pull/7234#discussion_r706303872

Scraping `stdout` and `stderr`, often from running a code sample in a subprocess, is error prone and ideally this should be avoided where an alternative exists (e.g. closing over the expected string and asserting its presence in some other way as part of the code sample being executed).

stuartarchibald:
As title, this prevents output from e.g diagnostics being able to
influence the outcome of the tests.

Fixes #7458

### DIFF
--- a/numba/tests/test_parallel_backend.py
+++ b/numba/tests/test_parallel_backend.py
@@ -536,7 +536,7 @@ class TestThreadingLayerPriority(ThreadLayerTestHelper):
         env['NUMBA_THREADING_LAYER'] = 'default'
         env['NUMBA_THREADING_LAYER_PRIORITY'] = env_var
 
-        code = """import sys
+        code = f"""
 import numba
 
 # trigger threading layer decision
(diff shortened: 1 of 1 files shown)


esc:
Although the PR didn't touch this line, I would recommend to change:

```
env = os.environ.copy()
```

into

```
env = {}
```

Copying the entire environment is a bit dangerous. For tests, it is better to start with a clean slate to avoid whatever variables exist in the environment to leak into the test. May not become an issue in practice, but I believe it's prudent to make sure it really doesn't.

numba/tests/test_parallel_backend.py
```diff
@@ -549,23 +549,29 @@ def each_env_var(self, env_var: str):
...
+assert numba.threading_layer() == captured_envvar[0],\
+    "selected backend mismatch"
 """
         cmd = [
             sys.executable,
             '-c',
             code,
         ]
-        return self.run_cmd(cmd, env=env)
+        self.run_cmd(cmd, env=env)
```
esc, line 563: Since you remove the return statement here, please also update the docstring to match, it currently still contains the documentation for the return.

numba/tests/test_parallel_backend.py
```diff
@@ -536,7 +536,7 @@ def each_env_var(self, env_var: str):
         env['NUMBA_THREADING_LAYER'] = 'default'
         env['NUMBA_THREADING_LAYER_PRIORITY'] = env_var
 
-        code = """import sys
```
esc, line 539: Perhaps use `dedent` for improved readability? 

https://docs.python.org/3/library/textwrap.html#textwrap.dedent

stuartarchibald:
> Although the PR didn't touch this line, I would recommend to change:
> 
> ```
> env = os.environ.copy()
> ```
> 
> into
> 
> ```
> env = {}
> ```
> 
> Copying the entire environment is a bit dangerous. For tests, it is better to start with a clean slate to avoid whatever variables exist in the environment to leak into the test. May not become an issue in practice, but I believe it's prudent to make sure it really doesn't.

Agree, am happy to make this change whilst editing these tests. It was, however, the env copy that actually triggered and demonstrated the problem with these tests in the first place!

numba/tests/test_parallel_backend.py
stuartarchibald: 7363275

numba/tests/test_parallel_backend.py
stuartarchibald: 7363275

esc (approved):
Thank you for the changes, looks good now!

stuartarchibald:
Turns out that doing https://github.com/numba/numba/pull/7749#issuecomment-1022065489 leads to:
```
Fatal Python error: _Py_HashRandomization_Init: failed to get random numbers to initialize Python
Python runtime state: preinitialized
```
on windows, probably because the sysroot is cleared. 1d7f47b puts back the copy of the env as it's likely the easiest way to obtain a suitable environment that doesn't require tracking required Python env vars.

esc (approved):
OK, let's go with the original.

PR #7750, opened 2022-01-17
"Iteration" over empty containers

### LINKED ISSUE #7245: [Feature Request] "Iteration" over empty tuple
Hi,

Currently trying to iterate over an empty tuple raises a typing error. Luckily there is a workaround (see below) so this is not entirely a deal breaker. But maybe it would be possible to generate some kind of generic empty iterator (or prune the loop entirely) such that no error is raised? (I hope I am not overlooking some weird consequences that could have)

Cheers

PS: Somewhere in the back of my mind I feel like I may have adressed this before, but I cannot seem to find an issue, sorry if this is a duplicate :-)

### Reproducer
```python 
from numba import njit

@njit
def f_guarded(x):
    if len(x)>0:
        for c in x:
            pass

@njit
def f_unguarded(x):
    for c in x:
        pass


cases = {
    "Guarded iterator with filled tuple": lambda: f_guarded((1,2,3)),
    "Guarded iterator with empty tuple": lambda: f_guarded(tuple()),
    "Unguarded iterator with filled tuple": lambda: f_unguarded((1,2,3)),
    "Unguarded iterator with empty tuple": lambda: f_unguarded(tuple()),
}

for desc, c in cases.items():
    try:
        c()
        print(desc + ": success")
    except Exception as e:
        print(desc + ": failed")
        print(e)
```
### Output
```
Guarded iterator with filled tuple: success
Guarded iterator with empty tuple: success
Unguarded iterator with filled tuple: success
Unguarded iterator with empty tuple: failed
Invalid use of getiter with parameters (Tuple())

During: typing of intrinsic-call at bugs/iter_empty_tuple.py (12)

File "bugs/iter_empty_tuple.py", line 12:
def f_unguarded(x):
    for c in x:
    ^
```

guilhermeleobas:
Fixes #7245. This PR includes an optimization pass that removes the loop if it is done over an empty container


numba/core/untyped_passes.py
```diff
@@ -1680,6 +1681,103 @@ def run_pass(self, state):
...
+    def __init__(self):
+        FunctionPass.__init__(self)
+        self.dead_vars = []
+
+    def is_arg_constsized(self, state, arg):
+        # True if len(type(arg)) == 0
+        t = state.typemap.get(arg.name)
+        return isinstance(t, types.ConstSized) and len(t) == 0
+
+    def match_iter(self, state, getiter):
```
stuartarchibald, line 1697: Docstring?

numba/core/untyped_passes.py
```diff
@@ -1680,6 +1681,103 @@ def run_pass(self, state):
...
+
+        for arg in call_expr.args:
+            # only one arg?
+            if self.is_arg_constsized(state, arg):
+                # store the call expr to be removed later
+                self.dead_vars.append(func_ir.get_assignee(call_expr))
+                return True
+        return False
+
+    def find_dead_loops(self, state):
```
stuartarchibald, line 1718: Docstring?

numba/core/untyped_passes.py
```diff
@@ -1680,6 +1681,103 @@ def run_pass(self, state):
...
+                expr = guard(get_definition, state.func_ir, expr.value.name)
+                if expr:
+                    if isinstance(expr, ir.Expr) and expr.op == 'getiter':
+                        # expr := getiter(value=Var)
+                        if self.is_arg_constsized(state, expr.value) or \
+                           self.match_iter(state, expr):
+                            dead_loops.append(loop)
+        return dead_loops
+
+    def remove_dead_loops(self, state, dead_loops):
```
stuartarchibald, line 1733: Docstring?

numba/core/untyped_passes.py
```diff
@@ -1680,6 +1681,103 @@ def run_pass(self, state):
...
+            term = header.terminator
+            header.insert_before_terminator(ir.Jump(term.falsebr, term.loc))
+            header.remove(term)
+
+            # iterate over header and remove "pair_first", since
+            # dead_code_elimination won't remove it
+            for inst in header.find_insts(ir.Assign):
+                expr = inst.value
+                if isinstance(expr, ir.Expr) and expr.op == 'pair_first':
+                    header.remove(inst)
```
stuartarchibald, line 1746: `func_ir._definitions` will contain reference to variables that do not exist following this and the `header.remove` above?

numba/core/untyped_passes.py
```diff
@@ -1680,6 +1681,103 @@ def run_pass(self, state):
...
+            pm.add_pass(RewriteSemanticConstants, "Semantic const")
+            pm.add_pass(DeadBranchPrune, "Dead branch prune")
+            pm.add_pass(GenericRewrites, "Generic Rewrites")
+        pm.add_pass(PartialTypeInference, "performs partial type inference")
+        pm.finalize()
+        pm.run(state)
+
+        dead_loops = self.find_dead_loops(state)
+        if dead_loops:
+            self.remove_dead_loops(state, dead_loops)
```
stuartarchibald, line 1773: I think this ends up with an invalid intermediate IR as there's a `phi` referring to the target of the `pair_first`, it just so happens that DCE removes it, but it is invalid. Example:
```python
from numba import njit
import numpy as np

@njit
def foo():
    for x in ():
        pass
foo()
```
has this state at this point:
```
label 0:
    $const2.0 = build_tuple(items=[])        ['$const2.0']
    $4get_iter.1 = getiter(value=$const2.0)  ['$4get_iter.1', '$const2.0']
    $phi6.0 = $4get_iter.1                   ['$4get_iter.1', '$phi6.0']
    jump 6                                   []
label 6:
    $6for_iter.1 = iternext(value=$phi6.0)   ['$6for_iter.1', '$phi6.0']
    $6for_iter.3 = pair_second(value=$6for_iter.1) ['$6for_iter.1', '$6for_iter.3']
    $phi8.1 = $6for_iter.2                   ['$6for_iter.2', '$phi8.1'] # THIS REFERS TO SOMETHING THAT DOES NOT EXIST!
    jump 12                                  []
label 12:
    $const12.0 = const(NoneType, None)       ['$const12.0']
    $14return_value.1 = cast(value=$const12.0) ['$14return_value.1', '$const12.0']
    return $14return_value.1                 ['$14return_value.1']
```

numba/tests/test_iteration.py
```diff
@@ -250,5 +250,84 @@ def foo(sequence):
...
+            return s
+
+        inps = [
+            ((), ()),
+            ((), (0, 2, 3)),
+            ((0, 2, 3), ()),
+        ]
+        for ca, cb in inps:
+            self.assertPreciseEqual(func(ca, cb), func.py_func(ca, cb))
+
```
stuartarchibald, line 330: Is it worth doing anything about cases like:
```python
@njit
def foo():

    for x in ():
        pass

    print(x)

foo()
```
the interpreter does `UnboundLocalError: local variable 'x' referenced before assignment`.
Numba doesn't know what the type of `'x'` is and can't work out where it is either (which is technically correct!).

numba/core/untyped_passes.py
```diff
@@ -1680,6 +1681,103 @@ def run_pass(self, state):
...
+
+        dead_branch_prune(state.func_ir, state.args)
+
+        for var in self.dead_vars:
+            if var is None:
+                continue
+            for block in state.func_ir.blocks.values():
+                assign = block.find_variable_assignment(var.name)
+                if assign:
+                    block.remove(assign)
```
stuartarchibald, line 1756: Same comment RE `func_ir._definitions`.

numba/core/untyped_passes.py
```diff
@@ -1680,6 +1681,103 @@ def run_pass(self, state):
         return mutated
 
 
+@register_pass(mutates_CFG=True, analysis_only=False)
+class DeadLoopElimination(FunctionPass):
```
stuartarchibald, line 1685: Docstring?

stuartarchibald (commented):
Thanks for working on this feature @guilhermeleobas, it's good to see this use case can now be handled. I've given this an initial review, I have a few concerns about the validity of certain parts and have raised these concerns inline, however, in general, the patch looks good. Thanks again!

numba/core/untyped_passes.py
```diff
@@ -1680,6 +1681,103 @@ def run_pass(self, state):
...
+            pm.add_pass(GenericRewrites, "Generic Rewrites")
+        pm.add_pass(PartialTypeInference, "performs partial type inference")
+        pm.finalize()
+        pm.run(state)
+
+        dead_loops = self.find_dead_loops(state)
+        if dead_loops:
+            self.remove_dead_loops(state, dead_loops)
+            state.func_ir.blocks = simplify_CFG(state.func_ir.blocks)
+            dead_code_elimination(state.func_ir)
```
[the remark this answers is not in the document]
DrTodd13: @stuartarchibald From first principles, I don't think that DCE should absolutely require types.  However, I looked at the code and our DCE does use typemap because it can improve alias analysis and do a better job at DCE.  Typemap is passed around to various other functions and I spot checked some of them and the story is the same there that typemap can improve accuracy but isn't 100% required.  I didn't check absolutely every call path so can't say that right now the typemap is optional but if not then it should be possible to make that happen.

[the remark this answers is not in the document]
ehsantn: I remember having issues with DCE without typemap but don't remember the details. Ideally, it should be more conservative but work without typemap.

numba/core/untyped_passes.py
guilhermeleobas: I'll reset the definitions once the pass execute

numba/core/untyped_passes.py
guilhermeleobas: But DCE will make the IR valid again, no? This is the IR I get after dead code elimination

```
label 0:
    $const12.0 = const(NoneType, None)       ['$const12.0']
    $14return_value.1 = cast(value=$const12.0) ['$14return_value.1', '$const12.0']
    return $14return_value.1                 ['$14return_value.1']
```

numba/core/compiler.py
```diff
@@ -628,6 +629,7 @@ def define_untyped_pipeline(state, name='untyped'):
             pm.add_pass(ReconstructSSA, "ssa")
 
         pm.add_pass(LiteralPropagationSubPipelinePass, "Literal propagation")
+        pm.add_pass(DeadLoopElimination, "remove dead loops")
```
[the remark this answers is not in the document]
guilhermeleobas: >  edit: having looked at the tests, is this here to accommodate literal_unroll?
Actually, no. There's no particular reason, I just place the pass there.

Looking now, the only requirement is that the pass must run after `LiteralUnroll`

numba/tests/test_iteration.py
```diff
@@ -250,5 +250,84 @@ def foo(sequence):
...
+                    break
+                else:
+                    pass
+            for e in (((),)[-1]):
+                if e:
+                    break
+                else:
+                    pass
+
+        empty_things()
```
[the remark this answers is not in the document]
guilhermeleobas: fixed

numba/core/untyped_passes.py
guilhermeleobas: @stuartarchibald, can you check if what I did seems correct?

numba/core/untyped_passes.py
stuartarchibald: It does happen to "fix it", but I wonder if this is always the case?

stuartarchibald:
The test suite is currently failing. An example that seems to trigger the problem is:
```python
from numba import njit
import numpy as np

@njit
def a():
    return 2

@njit
def b():
    return 3

@njit
def g(arg):
    if not arg:
        f = b
    else:
        f = a
    return f()

print(g(True))
```
it looks like this is down to errors being raise during partial type inference that are arising during the building of the constraint network here:
https://github.com/numba/numba/blob/59fa2e335be68148b3bd72a29de3ff011430038d/numba/core/typeinfer.py#L1051-L1052
Adding in SSA would probably "fix" it.

It might be time to start enforcing analysis usage and using the dependency trees, the current state is relying too much on memory of what can work with what. CC @sklam am going to raise this at a dev meeting.

guilhermeleobas:
> Adding in SSA would probably "fix" it.

So, DeadLoopElimination needs to run after SSA.

stuartarchibald:
> > Adding in SSA would probably "fix" it.
> 
> So, DeadLoopElimination needs to run after SSA.

This is because (partial) type inference needs SSA?

stuartarchibald:
> > > > Adding in SSA would probably "fix" it.
> > > 
> > > 
> > > So, DeadLoopElimination needs to run after SSA.
> > 
> > 
> > This is because (partial) type inference needs SSA?
> 
> It seems that this is the case. But I guess it is only needed if first-class function type is used

I think the partial type inference pass (`PartialTypeInference`) needs SSA form. Locally I tried making this pass depend on SSA through analysis usage but unfortunately it breaks the `LiteralUnroll` pass which uses the `PartialTypeInference` pass in its subpipeline. Therefore, I think for now, whilst a bit overreaching, making this pass depend on SSA is the only option.

numba/core/untyped_passes.py
```diff
@@ -1680,6 +1681,135 @@ def run_pass(self, state):
...
+
+        self.dead_vars = []
+        dead_loops = self.find_dead_loops(state)
+        if dead_loops:
+            self.remove_dead_loops(state, dead_loops)
+            state.func_ir.blocks = simplify_CFG(state.func_ir.blocks)
+            dead_code_elimination(state.func_ir)
+
+            # reset type inference now we are done with the partial results
+            state.typemap = None
```
stuartarchibald, line 1802:
```suggestion
            state.typemap = {}
```

numba/core/untyped_passes.py
stuartarchibald: Thanks, the code added to rebuild looks approximately as I'd expect it to.

numba/core/untyped_passes.py
```diff
@@ -1680,6 +1681,135 @@ def run_pass(self, state):
...
+
+        call_fn = guard(get_definition, func_ir, call_expr.func)
+        if call_fn and call_fn.value not in (iter, zip):
+            return False
+
+        for arg in call_expr.args:
+            # only one arg?
+            if self.is_arg_constsized(state, arg):
+                # store the call expr to be removed later
+                self.dead_vars.append(func_ir.get_assignee(call_expr))
```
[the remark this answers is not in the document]
stuartarchibald: I also think that there also may be a case where a phi node needs its incoming values updating, where one of the incoming is from a "removed" var.

stuartarchibald (commented):
Thanks for the updates, I've given this a more in depth review and have tried to break it. There seems to be a few issues that result in mis-compiles that I've highlighted inline, I think the next step is to fix these and get the reproducers added as tests. Thanks again!

numba/core/untyped_passes.py
guilhermeleobas: In other places, `state.typemap` is assigned to `None`. Maybe as an effort to standardize, one should change None values to `{}`.

numba/core/untyped_passes.py
```diff
@@ -1680,6 +1681,135 @@ def run_pass(self, state):
...
+        t = state.typemap.get(arg.name)
+        return isinstance(t, types.ConstSized) and len(t) == 0
+
+    def match_iter(self, state, getiter):
+        """
+        Return True if getiter operates on 'zip' or 'iter'
+
+        call_fn  := global(zip: <class 'zip'>)
+        call_ret := call_expr(func=call_fn, args=[...])
+               _ := getiter(value=call_ret)
```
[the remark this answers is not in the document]
guilhermeleobas: I changed the code to support `zip(iter)` and I'm working on adding `enumerate` support as well.

numba/core/untyped_passes.py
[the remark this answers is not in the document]
guilhermeleobas: I've removed the `self.dead_vars` list as it doesn't seem to be necessary. About the test case, I'll put it in a test case to make sure we don't lose track of it.

numba/core/untyped_passes.py
```diff
@@ -1680,6 +1681,135 @@ def run_pass(self, state):
...
+        """
+        Find and return all dead loops. A loop is considered dead if it is
+        iterating over an empty container (i.e. empty tuple)
+        """
+        dead_loops = list()
+        cfg = compute_cfg_from_blocks(state.func_ir.blocks)
+        for loop in cfg.loops().values():
+            header = state.func_ir.blocks[loop.header]
+            for expr in header.find_exprs('iternext'):
+                expr = guard(get_definition, state.func_ir, expr.value.name)
```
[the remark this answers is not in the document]
guilhermeleobas: "Fixed". Numba will not run the pruning pass if the variable is a PHI node.

PR #7617, opened 2021-12-03
Numba gdb-python extension for printing

stuartarchibald:
This adds support for printing Numba types as their python
equivalents from gdb by using its python extension.

It includes fixes from https://github.com/numba/numba/pull/7729 and https://github.com/numba/numba/pull/7744

Pythonic printing support is added for:
* Numeric scalars
* Strings
* Tuples
* NumPy arrays of primitive numeric types, and very basic record/structured dtypes.

Also:
* docs - the documentation for gdb needs is updated to reflect these changes
* `numba -g` has been added to provide a simple way of displaying the location of the gdb printing extension file along with gdb configuration information.

PR #7764, opened 2022-01-19
Add dbg_optnone and dbg_extend_lifetimes flags

sklam:
- Adds `dbg_optnone` to disable optimization on specific functions for better debugging experience.
- Adds `dbg_extend_lifetimes` to extend value lifetimes for better debugging.

### DIFF
--- a/docs/source/user/troubleshoot.rst
+++ b/docs/source/user/troubleshoot.rst
@@ -443,7 +443,9 @@ JIT options for debug
 * ``_dbg_optnone`` (bool). Set to ``True`` to disable all LLVM optimization passes 
   on the function. Defaults to ``False``. See :envvar:`NUMBA_OPT` for a global setting
   to disable optimization.
-* ``_dbg_extend_lifetimes`` (bool). Set to ``True`` to extend the lifetime of objects such that they more closely follow the semantics of Python. Automatically set to ``True`` when 
+* ``_dbg_extend_lifetimes`` (bool). Set to ``True`` to extend the lifetime of
+  objects such that they more closely follow the semantics of Python.
+  Automatically set to ``True`` when 
   ``debug=True``; otherwise, defaults to ``False``. Users can explicitly set this option 
   to ``False`` to retain the normal execution semantics of compiled code.
   See :envvar:`NUMBA_EXTEND_VARIABLE_LIFETIMES` for a global option to extend object 
--- a/docs/source/user/troubleshoot.rst
+++ b/docs/source/user/troubleshoot.rst
@@ -443,10 +443,9 @@ JIT options for debug
 * ``_dbg_optnone`` (bool). Set to ``True`` to disable all LLVM optimization passes 
   on the function. Defaults to ``False``. See :envvar:`NUMBA_OPT` for a global setting
   to disable optimization.
-* ``_dbg_extend_lifetimes`` (bool). Set to ``True`` to extend object lifetimes to match 
-  closer to Python semantic for the function. Automatically set to ``True`` when 
+* ``_dbg_extend_lifetimes`` (bool). Set to ``True`` to extend the lifetime of objects such that they more closely follow the semantics of Python. Automatically set to ``True`` when 
   ``debug=True``; otherwise, defaults to ``False``. Users can explicitly set this option 
-  to ``False`` to retain  semantic of normal execution of compiled code.
+  to ``False`` to retain the normal execution semantics of compiled code.
   See :envvar:`NUMBA_EXTEND_VARIABLE_LIFETIMES` for a global option to extend object 
   lifetimes.
 
--- a/docs/source/user/troubleshoot.rst
+++ b/docs/source/user/troubleshoot.rst
@@ -443,8 +443,8 @@ JIT options for debug
 * ``_dbg_optnone`` (bool). Set to ``True`` to disable all LLVM optimization passes 
   on the function. Defaults to ``False``. See :envvar:`NUMBA_OPT` for a global setting
   to disable optimization.
-* ``_dbg_extend_lifetimes`` (bool). Set to ``True` to extend object lifetimes to match 
-  closer to Python semantic for the function. Automatically set to ``True` when 
+* ``_dbg_extend_lifetimes`` (bool). Set to ``True`` to extend object lifetimes to match 
+  closer to Python semantic for the function. Automatically set to ``True`` when 
   ``debug=True``; otherwise, defaults to ``False``. Users can explicitly set this option 
   to ``False`` to retain  semantic of normal execution of compiled code.
   See :envvar:`NUMBA_EXTEND_VARIABLE_LIFETIMES` for a global option to extend object 
--- a/docs/source/user/troubleshoot.rst
+++ b/docs/source/user/troubleshoot.rst
@@ -407,8 +407,11 @@ Known issues:
 
 * Stepping depends heavily on optimization level. At full optimization
   (equivalent to O3), most of the variables are optimized out. It is often
-  beneficial to use the environment variable :envvar:`NUMBA_OPT` to adjust the
-  optimization level and :envvar:`NUMBA_EXTEND_VARIABLE_LIFETIMES` to extend
+  beneficial to use the jit option ``_dbg_optnone=True`` 
+  or the environment variable :envvar:`NUMBA_OPT` to adjust the 
+  optimization level and the jit option ``_dbg_extend_lifetimes=True`` 
+  (which is on by default if ``debug=True``) or
+  :envvar:`NUMBA_EXTEND_VARIABLE_LIFETIMES` to extend
   the lifetime of variables to the end of their scope so as to get a debugging
   experience closer to the semantics of Python execution.
 
@@ -432,6 +435,20 @@ Internal details:
   ``x``, ``x.1`` and ``x.2``.)
 
 * When debug is enabled, inlining of functions at LLVM IR level is disabled.
+
+JIT options for debug
+---------------------
+
+* ``debug`` (bool). Set to ``True`` to enable debug info. Defaults to ``False``.
+* ``_dbg_optnone`` (bool). Set to ``True`` to disable all LLVM optimization passes 
+  on the function. Defaults to ``False``. See :envvar:`NUMBA_OPT` for a global setting
+  to disable optimization.
+* ``_dbg_extend_lifetimes`` (bool). Set to ``True` to extend object lifetimes to match 
+  closer to Python semantic for the function. Automatically set to ``True` when 
+  ``debug=True``; otherwise, defaults to ``False``. Users can explicitly set this option 
+  to ``False`` to retain  semantic of normal execution of compiled code.
+  See :envvar:`NUMBA_EXTEND_VARIABLE_LIFETIMES` for a global option to extend object 
+  lifetimes.
 
 Example debug usage
 -------------------
--- a/numba/core/base.py
+++ b/numba/core/base.py
@@ -431,7 +431,8 @@ class BaseContext(object):
         fn = cgutils.get_or_insert_function(module, fnty, fndesc.mangled_name)
         self.call_conv.decorate_function(fn, fndesc.args, fndesc.argtypes, noalias=fndesc.noalias)
         if fndesc.inline:
-            fn.attributes.add('alwaysinline')
+            if 'noinline' not in fn.attributes:
+                fn.attributes.add('alwaysinline')
         return fn
 
     def declare_external_function(self, module, fndesc):
--- a/numba/core/base.py
+++ b/numba/core/base.py
@@ -431,8 +431,10 @@ class BaseContext(object):
         fn = cgutils.get_or_insert_function(module, fnty, fndesc.mangled_name)
         self.call_conv.decorate_function(fn, fndesc.args, fndesc.argtypes, noalias=fndesc.noalias)
         if fndesc.inline:
-            if 'noinline' not in fn.attributes:
-                fn.attributes.add('alwaysinline')
+            fn.attributes.add('alwaysinline')
+            # alwaysinline overrides optnone
+            fn.attributes.discard('noinline')
+            fn.attributes.discard('optnone')
         return fn
 
     def declare_external_function(self, module, fndesc):
--- a/numba/core/compiler.py
+++ b/numba/core/compiler.py
@@ -143,6 +143,12 @@ detail""",
         doc="backend"
     )
 
+    dbg_extend_lifetimes = Option(
+        type=bool,
+        default=False,
+        doc="Extend variable lifetime for debugging",
+    )
+
 
 DEFAULT_FLAGS = Flags()
 DEFAULT_FLAGS.nrt = True
--- a/numba/core/compiler.py
+++ b/numba/core/compiler.py
@@ -149,6 +149,12 @@ detail""",
         doc="Extend variable lifetime for debugging",
     )
 
+    dbg_optnone = Option(
+        type=bool,
+        default=False,
+        doc="Disable opt for debug. Put optnone attribute in LLVM Function."
+    )
+
 
 DEFAULT_FLAGS = Flags()
 DEFAULT_FLAGS.nrt = True
--- a/numba/core/compiler.py
+++ b/numba/core/compiler.py
@@ -87,7 +87,7 @@ class Flags(TargetConfig):
     forceinline = Option(
         type=bool,
         default=False,
-        doc="TODO",
+        doc="Force inlining of the function. Overrides dbg_optnone.",
     )
     no_cpython_wrapper = Option(
         type=bool,
@@ -147,13 +147,15 @@ detail""",
     dbg_extend_lifetimes = Option(
         type=bool,
         default=False,
-        doc="Extend variable lifetime for debugging",
+        doc=("Extend variable lifetime for debugging. "
+             "This automatically turns on with debug=True."),
     )
 
     dbg_optnone = Option(
         type=bool,
         default=False,
-        doc="Disable opt for debug. Put optnone attribute in LLVM Function."
+        doc=("Disable optimization for debug. "
+             "Equivalent to adding optnone attribute in the LLVM Function.")
     )
 
 
--- a/numba/core/compiler.py
+++ b/numba/core/compiler.py
@@ -87,7 +87,7 @@ class Flags(TargetConfig):
     forceinline = Option(
         type=bool,
         default=False,
-        doc="Force inlining of the function. Overrides dbg_optnone.",
+        doc="Force inlining of the function. Overrides _dbg_optnone.",
     )
     no_cpython_wrapper = Option(
         type=bool,
--- a/numba/core/cpu.py
+++ b/numba/core/cpu.py
@@ -296,8 +296,6 @@ class CPUTargetOptions(_options_mixin, TargetOptions):
             if not flags.is_set("dbg_extend_lifetimes"):
                 # auto turn on extend-lifetimes if debuginfo is on.
                 flags.dbg_extend_lifetimes = True
-            if not flags.is_set("dbg_optnone"):
-                flags.dbg_optnone = True
 
         if not flags.is_set("boundscheck"):
             flags.boundscheck = flags.debuginfo
@@ -312,6 +310,10 @@ class CPUTargetOptions(_options_mixin, TargetOptions):
         flags.inherit_if_not_set("target_backend")
 
         flags.inherit_if_not_set("forceinline")
+
+        if flags.forceinline:
+            # forceinline turns off optnone, just like clang.
+            flags.optnone = False
 
 # ----------------------------------------------------------------------------
 # Internal
--- a/numba/core/cpu.py
+++ b/numba/core/cpu.py
@@ -294,6 +294,7 @@ class CPUTargetOptions(_options_mixin, TargetOptions):
 
         if flags.debuginfo:
             if not flags.is_set("dbg_extend_lifetimes"):
+                # auto turn on extend-lifetimes if debuginfo is on.
                 flags.dbg_extend_lifetimes = True
             if not flags.is_set("dbg_optnone"):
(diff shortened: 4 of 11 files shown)


sklam:
```python
from numba import njit

@njit(debug=True, dbg_optnone=True)
def foo(x):
    c = 0
    for i in range(10):
        c+= i
    y = x + c
    print(y)
    return y


r = foo(123)
```

without optnone, most of the code will be optimized out:
```
(numba_dev) [root@cd000bcd2aef numba]# gdb -q --args python test_gdb.py
Reading symbols from python...
(gdb) b __main__::foo_241
Function "__main__::foo_241" not defined.
Make breakpoint pending on future shared library load? (y or [n]) y
Breakpoint 1 (__main__::foo_241) pending.
(gdb) r
Starting program: /root/miniconda3/envs/numba_dev/bin/python test_gdb.py

Breakpoint 1, 0x00007ffff6e5d140 in cpython::__main__::foo_241[abi:c8tJTC_2fWwKbcFpQAdBTgEQrlo6UyVFGMNAWVBgYHmKUJAA_3d_3d](long long) ()
(gdb) n
Single stepping until exit from function _ZN7cpython8__main__7foo_241B54c8tJTC_2fWwKbcFpQAdBTgEQrlo6UyVFGMNAWVBgYHmKUJAA_3d_3dEx,
which has no line number information.

Breakpoint 1, __main__::foo_241[abi:c8tJTC_2fWwKbcFpQAdBTgEQrlo6UyVFGMNAWVBgYHmKUJAA_3d_3d](long long) (x=0) at test_gdb.py:10
10	    print(y)
```

with optnone, gdb will stop at the first line when breaking on the function:
```
(gdb) b __main__::foo_241
Function "__main__::foo_241" not defined.
Make breakpoint pending on future shared library load? (y or [n]) y
Breakpoint 1 (__main__::foo_241) pending.
(gdb) r
Starting program: /root/miniconda3/envs/numba_dev/bin/python test_gdb.py

Breakpoint 1, 0x00007ffff6e5d930 in cpython::__main__::foo_241[abi:c8tJTC_2fWwKbcFpQAdBSQNMBEYFEM5aOlO1RRjFQGlQYGEJilCQA_3d](long long) ()
(gdb) n
Single stepping until exit from function _ZN7cpython8__main__7foo_241B56c8tJTC_2fWwKbcFpQAdBSQNMBEYFEM5aOlO1RRjFQGlQYGEJilCQA_3dEx,
which has no line number information.

Breakpoint 1, __main__::foo_241[abi:c8tJTC_2fWwKbcFpQAdBSQNMBEYFEM5aOlO1RRjFQGlQYGEJilCQA_3d](long long) (x=123) at test_gdb.py:6
6	    c = 0
(gdb) n
7	    for i in range(10):
(gdb) n
8	        c+= i
```

stuartarchibald:
Thanks for the patch. Looks like the typeguard tests are failing, expects a bool found an int (failing build is `Linux py38_np120_typeguard`):
```
TypeError: type of argument "extend_lifetimes" must be bool; got int instead
```

numba/core/base.py
```diff
@@ -431,7 +431,8 @@ def declare_function(self, module, fndesc):
         fn = cgutils.get_or_insert_function(module, fnty, fndesc.mangled_name)
         self.call_conv.decorate_function(fn, fndesc.args, fndesc.argtypes, noalias=fndesc.noalias)
         if fndesc.inline:
-            fn.attributes.add('alwaysinline')
+            if 'noinline' not in fn.attributes:
+                fn.attributes.add('alwaysinline')
```
stuartarchibald, line 435: How does this situation occur? Does it imply that if a function is described with `inline` but has attr `noinline` it will remain so even if a user specifies `forceinline`, this state seems a bit confusing?

numba/core/compiler.py
```diff
@@ -144,6 +144,20 @@ class Flags(TargetConfig):
         doc="backend"
     )
 
+    dbg_extend_lifetimes = Option(
+        type=bool,
+        default=False,
+        doc=("Extend variable lifetime for debugging. "
+             "This automatically turns on with debug=True."),
```
stuartarchibald, line 151: Whilst not introduced by this PR, it seems confusing that user flag `debug` maps to `debuginfo` on the target. But `debuginfo` kind of implies DI for DWARF emission.

numba/core/cpu.py
```diff
@@ -290,6 +292,11 @@ def finalize(self, flags, options):
         if not flags.is_set("debuginfo"):
             flags.debuginfo = config.DEBUGINFO_DEFAULT
 
+        if flags.debuginfo:
+            if not flags.is_set("dbg_extend_lifetimes"):
+                # auto turn on extend-lifetimes if debuginfo is on.
+                flags.dbg_extend_lifetimes = True
```
stuartarchibald, line 298: Does this present a problem if the code being debugged is suspected to have a bug related to life times? Is it possible to instead do the above but also include: if the user explicitly sets  `dbg_extend_lifetimes` to `False` and `debug` to `True`, then life times are not extended. i.e. the explicit request to not extend lifetimes is acknowledged but the rest of `debuginfo` works as usual.

numba/core/ir_utils.py
```diff
@@ -2215,15 +2210,16 @@ def legalize_single_scope(blocks):
...
+def check_and_legalize_ir(func_ir, flags: "numba.core.compiler.Flags"):
     """
     This checks that the IR presented is legal
     """
     enforce_no_phis(func_ir)
     enforce_no_dels(func_ir)
     # postprocess and emit ir.Dels
     post_proc = postproc.PostProcessor(func_ir)
-    post_proc.run(True, extend_lifetimes=config.EXTEND_VARIABLE_LIFETIMES)
+    ext_life = flags.dbg_extend_lifetimes or config.EXTEND_VARIABLE_LIFETIMES
```
stuartarchibald, line 2221: IIRC the environment variables are usually overridden by in-source definition?

numba/core/lowering.py
```diff
@@ -305,6 +305,12 @@ def create_cfunc_wrapper(self):
     def setup_function(self, fndesc):
         # Setup function
         self.function = self.context.declare_function(self.module, fndesc)
+        if self.flags.dbg_optnone:
+            attrset = self.function.attributes
+            attrset.add("optnone")
+            # optnone requires noinline unless it has alwaysinline
+            attrset.discard("alwaysinline")
+            attrset.add("noinline")
```
stuartarchibald, line 313: It looks like this always adds `optnone noinline` and removes `alwaysinline`, should it be:
```
if not ("alwaysinline" in attrset):
    attrset.add("optnone")
    attrset.add("noinline")
```
i.e. if`alwaysinline` is present, do nothing, else add `optnone noinline`?

numba/core/options.py
```diff
@@ -92,6 +92,9 @@ class DefaultOptions:
 
     target_backend = _mapping("target_backend")
 
+    dbg_extend_lifetimes = _mapping("dbg_extend_lifetimes")
+    dbg_optnone = _mapping("dbg_optnone")
```
stuartarchibald, line 96:
```suggestion
    _dbg_extend_lifetimes = _mapping("dbg_extend_lifetimes")
    _dbg_optnone = _mapping("dbg_optnone")
```
(and refactor to match). Suggest making these options available but not explicitly public on the basis that they really need designing into the `debug=` flag as part of a debugging configuration object for that flag to consume.

numba/tests/test_debuginfo.py
```diff
@@ -595,6 +595,42 @@ def foo():
...
+            # The optnone version have all the lines still in there.
+            mdlist = self._get_metadata(udt, udt.signatures[0])
+            lines = set()
+            for md in mdlist:
+                m = re.match(r"!\d+ = !DILocation\(line: (\d+),", md)
+                if m:
+                    ln = int(m.group(1))
+                    lines.add(ln)
+            self.assertEqual(lines, expected_lines)
+
```
stuartarchibald, line 633: Should the metadata also be scanned to look for function attrs like `optnone`?

Also, might be good to test the logic and arising from combinations of the flags: `forceinline`, `debug`, `dbg_opt_none` and `dbg_extend_lifetimes`.

stuartarchibald (commented):
Thanks for the PR. As requested OOB I've given this an early review to provide some initial feedback given the complexity of incorporating this change. The patch in general looks good and I think in large will enable the emission of appropriate LLVM IR for improving the debugging experience. Comments are inline, my largest concerns at present surround the interaction of flags. Thanks again!

numba/core/cpu.py
sklam: Yes, this is what is happening. The comment needs updating.

numba/core/compiler.py
sklam: Yes, this goes back to the discussion to better define what `debug` flag does.

numba/core/ir_utils.py
sklam: good point

numba/core/options.py
sklam: done in d036b1e65

numba/core/base.py
sklam: I changed the logic. Now `forceinline` will always override `optnone`.

numba/core/lowering.py
sklam: applied in 9c9a56b

stuartarchibald:
This is currently failing CI: https://dev.azure.com/numba/ff1fe4d0-ed73-4f1c-b894-1d50a27e048f/_apis/build/builds/10901/logs/66

<details>
<summary>Details:</summary>

```
2022-01-26T02:33:08.8964499Z ======================================================================
2022-01-26T02:33:08.8966094Z FAIL: test_array (numba.tests.test_gdb_dwarf.TestGDBDwarf)
2022-01-26T02:33:08.8968458Z ----------------------------------------------------------------------
2022-01-26T02:33:08.8969919Z Traceback (most recent call last):
2022-01-26T02:33:08.8975643Z   File "/home/vsts/work/1/s/numba/tests/test_gdb_dwarf.py", line 30, in test_array
2022-01-26T02:33:08.8977678Z     self._subprocess_test_runner('test_array_arg')
2022-01-26T02:33:08.8979199Z   File "/home/vsts/work/1/s/numba/tests/test_gdb_dwarf.py", line 24, in _subprocess_test_runner
2022-01-26T02:33:08.8980569Z     envvars=self._NUMBA_OPT_0_ENV)
2022-01-26T02:33:08.8981879Z   File "/home/vsts/work/1/s/numba/tests/support.py", line 564, in subprocess_test_runner
2022-01-26T02:33:08.8983399Z     self.assertEqual(status.returncode, 0, streams)
2022-01-26T02:33:08.8987270Z AssertionError: 1 != 0 : 
2022-01-26T02:33:08.8988513Z captured stdout: 
2022-01-26T02:33:08.8989660Z captured stderr: F
2022-01-26T02:33:08.8990802Z ======================================================================
2022-01-26T02:33:08.8991563Z FAIL: test (numba.tests.gdb.test_array_arg.Test)
2022-01-26T02:33:08.8992684Z ----------------------------------------------------------------------
2022-01-26T02:33:08.8993490Z Traceback (most recent call last):
2022-01-26T02:33:08.8994768Z   File "/home/vsts/work/1/s/numba/tests/gdb/test_array_arg.py", line 40, in test
2022-01-26T02:33:08.8995510Z     driver.check_hit_breakpoint(2)
2022-01-26T02:33:08.8996261Z   File "/home/vsts/work/1/s/numba/tests/gdb_support.py", line 127, in check_hit_breakpoint
2022-01-26T02:33:08.8997301Z     self.assert_output(f'bkptno="{number}"')
2022-01-26T02:33:08.8998093Z   File "/home/vsts/work/1/s/numba/tests/gdb_support.py", line 71, in assert_output
2022-01-26T02:33:08.8999125Z     assert expected in decoded, f'decoded={decoded}\nexpected={expected})'
2022-01-26T02:33:08.9001320Z AssertionError: decoded=*stopped,reason="breakpoint-hit",disp="keep",bkptno="1",frame={addr="0x00007ffff6922047",func="_3cdynamic_3e::__numba_array_expr_0x7fffdfbd4090_242[abi:c8tJTC_2fWwKbcFpQAdBTgEQrlo6UyVFGMNAWTRnjbFhyiwIBFi2OoSkRyQhJATSIwiSKI6ZoA](double)",args=[{name="x_1",value="1"}],file="test_array_arg.py",fullname="/home/vsts/work/1/s/numba/tests/gdb/test_array_arg.py",line="15",arch="i386:x86-64"},thread-id="1",stopped-threads="all",core="1"
2022-01-26T02:33:08.9006892Z (gdb) 
2022-01-26T02:33:08.9007835Z 
2022-01-26T02:33:08.9008655Z expected=bkptno="2")
2022-01-26T02:33:08.9009174Z 
2022-01-26T02:33:08.9010289Z ----------------------------------------------------------------------
2022-01-26T02:33:08.9011027Z Ran 1 test in 8.923s
2022-01-26T02:33:08.9011405Z 
2022-01-26T02:33:08.9011905Z FAILED (failures=1)
2022-01-26T02:33:08.9012281Z 
2022-01-26T02:33:08.9012620Z 
2022-01-26T02:33:08.9013433Z ----------------------------------------------------------------------
2022-01-26T02:33:08.9014090Z Ran 539 tests in 188.147s
2022-01-26T02:33:08.9014499Z 
2022-01-26T02:33:08.9015030Z FAILED (failures=1, skipped=57, expected failures=1)
2022-01-26T02:33:09.5174775Z ##[error]Bash exited with code '1'.
2022-01-26T02:33:09.5217046Z ##[section]Finishing: Test
```

</details>

Cannot reproduce yet locally, would guess at the cause being: https://github.com/numba/numba/pull/7752, looks like the first breakpoint is being hit twice in succession, but the reported func for the second hit is `_3cdynamic_3e::__numba_array_expr_0x7fffdfbd4090_242`, i.e. something dynamically generated that should perhaps not be associated with this file/line.

stuartarchibald:
> Cannot reproduce yet locally,

Reproduction occurs on at least Python 3.7.

stuartarchibald:
Python 3.7 gdb session sees the same:
```bash
$ cat test_7764.py
from numba import njit
import numpy as np










@njit(debug=True) # Declare here to match location in `test_array_arg.py`
def foo(x):
    z = 7 + x # break here
    return x, z

tmp = np.ones(5)
foo(tmp)

$ NUMBA_OPT=0 gdb -q -ex 'b test_7764.py:15' -ex 'r' --args python test_7764.py
<snip>

Breakpoint 1, __main__::foo_241[abi:c8tJTC_2fWwKbcFpQAdBTgEQrlo6UyVFGMNAWVBgYHmKUJAA_3d_3d](Array<double, 1, C, mutable, aligned>) (x=...) at test_7764.py:15
15          z = 7 + x # break here

(gdb) b 16
Breakpoint 2 at 0x7fffdaf589f5: file test_7764.py, line 16.
(gdb) c
Continuing.

Breakpoint 1, _3cdynamic_3e::__numba_array_expr_0x7fffdb2fa1d0_242[abi:c8tJTC_2fWwKbcFpQAdBTgEQrlo6UyVFGMNAWTRnjbFhyiwIBFi2OoSkRyQhJATSIwiSKI6ZoA](double) (x_1=1) at test_7764.py:15
15          z = 7 + x # break here
(gdb) c
Continuing.

Breakpoint 1, _3cdynamic_3e::__numba_array_expr_0x7fffdb2fa1d0_242[abi:c8tJTC_2fWwKbcFpQAdBTgEQrlo6UyVFGMNAWTRnjbFhyiwIBFi2OoSkRyQhJATSIwiSKI6ZoA](double) (x_1=1) at test_7764.py:15
15          z = 7 + x # break here
(gdb) c
Continuing.

Breakpoint 1, _3cdynamic_3e::__numba_array_expr_0x7fffdb2fa1d0_242[abi:c8tJTC_2fWwKbcFpQAdBTgEQrlo6UyVFGMNAWTRnjbFhyiwIBFi2OoSkRyQhJATSIwiSKI6ZoA](double) (x_1=1) at test_7764.py:15
15          z = 7 + x # break here
(gdb) c
Continuing.

Breakpoint 1, _3cdynamic_3e::__numba_array_expr_0x7fffdb2fa1d0_242[abi:c8tJTC_2fWwKbcFpQAdBTgEQrlo6UyVFGMNAWTRnjbFhyiwIBFi2OoSkRyQhJATSIwiSKI6ZoA](double) (x_1=1) at test_7764.py:15
15          z = 7 + x # break here
(gdb) c
Continuing.

Breakpoint 1, _3cdynamic_3e::__numba_array_expr_0x7fffdb2fa1d0_242[abi:c8tJTC_2fWwKbcFpQAdBTgEQrlo6UyVFGMNAWTRnjbFhyiwIBFi2OoSkRyQhJATSIwiSKI6ZoA](double) (x_1=1) at test_7764.py:15
15          z = 7 + x # break here
(gdb) c
Continuing.

Breakpoint 2, __main__::foo_241[abi:c8tJTC_2fWwKbcFpQAdBTgEQrlo6UyVFGMNAWVBgYHmKUJAA_3d_3d](Array<double, 1, C, mutable, aligned>) (x=...) at test_7764.py:16
16          return x, z
```

stuartarchibald:
Reverting https://github.com/numba/numba/pull/7752 seems to fix it.

stuartarchibald:
Suggest changing the test as follows:
```python
diff --git a/numba/tests/gdb/test_array_arg.py b/numba/tests/gdb/test_array_arg.py
index f4b4b88..e4a969b 100644
--- a/numba/tests/gdb/test_array_arg.py
+++ b/numba/tests/gdb/test_array_arg.py
@@ -12,7 +12,7 @@ class Test(TestCase):
     def test(self):
         @njit(debug=True)
         def foo(x):
-            z = 7 + x # break here
+            z = np.ones_like(x) # break here
             return x, z
 
         tmp = np.ones(5)
```
this retains the testing of the break points and array update, but avoids the use of an `arrayexpr`.

stuartarchibald:
https://github.com/numba/numba/pull/7788 fixes.

numba/tests/test_debuginfo.py
```diff
@@ -596,6 +596,24 @@ def foo():
...
+        def get_debug_lines(fn):
+            # Get the lines contained in the debug info
+            mdlist = self._get_metadata(fn, fn.signatures[0])
+            lines = set()
+            for md in mdlist:
+                m = re.match(r"!\d+ = !DILocation\(line: (\d+),", md)
+                if m:
+                    ln = int(m.group(1))
+                    lines.add(ln)
+            return lines
```
stuartarchibald, line 608: This is quite like: https://github.com/numba/numba/pull/7617/files#diff-f5c7dace4d41363f673092ed1b4030c54136136260a87cfe2916faeeac985c4fR151-R162 which is a refactor of these tests to pull out the code that extracts the metadata.

stuartarchibald (commented):
Thanks for the fixes. This looks good. I've also tested this manually a load in `gdb` and it seems to have the desired effect(s). The remaining question here is how best to document this, options are probably in the docstring of the `@jit` decorator and/or in the docs on troubleshooting/debugging, both may be appropriate. Probably also worth noting that these are "private" debug flags and are likely to change in future releases.

numba/core/cpu.py
```diff
@@ -290,6 +292,15 @@ def finalize(self, flags, options):
         if not flags.is_set("debuginfo"):
             flags.debuginfo = config.DEBUGINFO_DEFAULT
 
+        if not flags.is_set("dbg_extend_lifetimes"):
```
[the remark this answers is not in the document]
sklam: Correct. no inheriting lifetime flags. I will make sure to highlight in the docs about setting the env-var for extending lifetimes for the entire program.

docs/source/user/troubleshoot.rst
```diff
@@ -433,6 +436,20 @@ Internal details:
...
+---------------------
+
+* ``debug`` (bool). Set to ``True`` to enable debug info. Defaults to ``False``.
+* ``_dbg_optnone`` (bool). Set to ``True`` to disable all LLVM optimization passes 
+  on the function. Defaults to ``False``. See :envvar:`NUMBA_OPT` for a global setting
+  to disable optimization.
+* ``_dbg_extend_lifetimes`` (bool). Set to ``True`` to extend object lifetimes to match 
+  closer to Python semantic for the function. Automatically set to ``True`` when 
+  ``debug=True``; otherwise, defaults to ``False``. Users can explicitly set this option 
+  to ``False`` to retain  semantic of normal execution of compiled code.
```
stuartarchibald, line 449:
```suggestion
  to ``False`` to retain the normal execution semantics of compiled code.
```

stuartarchibald (commented):
Thanks for the update. Couple of minor things to look at in the docs but looks good else.

numba/tests/test_debuginfo.py
sklam: refactored in https://github.com/numba/numba/pull/7764/commits/ece8ce1b2d96aa7914a27312a5fdb8c986f5b239

docs/source/user/troubleshoot.rst
```diff
@@ -433,6 +436,19 @@ Internal details:
...
 * When debug is enabled, inlining of functions at LLVM IR level is disabled.
 
+JIT options for debug
+---------------------
+
+* ``debug`` (bool). Set to ``True`` to enable debug info. Defaults to ``False``.
+* ``_dbg_optnone`` (bool). Set to ``True`` to disable all LLVM optimization passes 
+  on the function. Defaults to ``False``. See :envvar:`NUMBA_OPT` for a global setting
+  to disable optimization.
+* ``_dbg_extend_lifetimes`` (bool). Set to ``True`` to extend the lifetime of objects such that they more closely follow the semantics of Python. Automatically set to ``True`` when
```
stuartarchibald, line 446:
```suggestion
* ``_dbg_extend_lifetimes`` (bool). Set to ``True`` to extend the lifetime of
  objects such that they more closely follow the semantics of Python.
  Automatically set to ``True`` when 
```
suggest line wrap.

stuartarchibald (approved):

PR #7765, opened 2022-01-19
Updated versioneer to enable Python 3.11

oscargus:
versioneer must be updated as the older versions use stuff that is removed in Python 3.11.

So while this may not be enough to run on Python 3.11, it is at least required at some stage.


esc:
@oscargus thank you, we will begin to work on 3.11 support soon, so this will come in handy!

esc:
@oscargus question: will this update of the vendored versioneer be needed for llvmlite: https://github.com/numba/llvmlite too? And also, we could just merge this now as the change is backwards compatible with previous Python versions?

oscargus:
Updated to latest version of versioneer. (Note that this is in no way dependent on 3.11, so it can be merged earlier if you want. More the opposite: you will not get 3.11 to work without it.)

oscargus:
As discussed in https://github.com/numba/llvmlite/pull/812 the current version in numba is also too old to have the problem I was trying to solve.

esc:
I think this can be closed now. A new PR porting versionner to 3.14 might be in order? But then again, there isn't anything broken by the current version so no technical need to update.

PR #7771, opened 2022-01-21
Move function unique ID to abi-tags

sklam:
- Add unique-id to `FunctionIdentity` and `FunctionDescriptor` classes.
- `RecursionCall` class API is improved. 
- Mangler function now takes an optional `uid` argument for the unique id.

### DIFF
--- a/numba/tests/gdb/test_break_on_symbol_version.py
+++ b/numba/tests/gdb/test_break_on_symbol_version.py
@@ -1,0 +1,65 @@
+# NOTE: This test is sensitive to line numbers as it checks breakpoints
+from numba import njit
+from numba.tests.gdb_support import GdbMIDriver
+from numba.tests.support import TestCase, needs_subprocess
+import unittest
+
+
+def foo_factory(n):
+    @njit(debug=True)
+    def foo(x):
+        z = 7 + n
+        return x, z
+
+    return foo
+
+
+foo1, foo2, foo3 = [foo_factory(x) for x in range(3)]
+
+
+@njit(debug=True)
+def call_foo():
+    a = foo1(10)
+    b = foo2(20)
+    c = foo3(30)
+    return a, b, c
+
+
+@needs_subprocess
+class Test(TestCase):
+
+    def test(self):
+        call_foo()
+        driver = GdbMIDriver(__file__)
+        # A specific foo, the first one, it has uid=2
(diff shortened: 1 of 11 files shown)


stuartarchibald:
@sklam This patch updates and has a new test for this change, feel free to incorporate etc: https://github.com/stuartarchibald/numba/commit/81fad4f5f859a162396beaee16f51c9d114f69aa

sklam:
@stuartarchibald, i cherrypicked your tests. (It's showing up before your message about it in this page.)

numba/tests/gdb/test_break_on_symbol_version.py
```diff
@@ -0,0 +1,65 @@
+# NOTE: This test is sensitive to line numbers as it checks breakpoints
+from numba import njit
```
stuartarchibald, line 2: @sklam as I wrote this, please can you review the test file? Thanks.

stuartarchibald (commented):
Thanks for the patch @sklam. Two comments 1. is inline, about reviewing the test I wrote. 2. Should an example be added to the docs?

numba/tests/gdb/test_break_on_symbol_version.py
sklam: Yes, i have reviewed it and play around with it as i cherrypicked it over.

sklam:
@stuartarchibald, regarding docs and example, I think it's better if we convert the gdb docs to a more generative approach once we merge #76er 17, #7764 and #7771 (this). I find the current docs quite scattered and requires too much manual effort each time we the debuginfo.

numba/tests/gdb/test_break_on_symbol_version.py
stuartarchibald: Great, thanks for confirming.

stuartarchibald:
> @stuartarchibald, regarding docs and example, I think it's better if we convert the gdb docs to a more generative approach once we merge #7617, #7764 and #7771 (this). I find the current docs quite scattered and requires too much manual effort each time we the debuginfo.

Having manually updated this region of the docs a few times, I agree, a more generative approach would be beneficial. I also think it might be worth having a section on native debugging that is stand-alone. Let's leave doc changes for now with view of getting this functionality merged.

stuartarchibald (approved):
Thanks for the patch!

PR #7776, opened 2022-01-24
Updates coverage.py config

stuartarchibald:
As title.


esc (approved):
Thank you for the patch!

PR #7777, opened 2022-01-24
Remove reference existing issue from GH template.

stuartarchibald:
As title.


stuartarchibald:
Whilst well intentioned, this hasn't really been used, elected to removed.

esc (approved):
Yes, time has proved this to be superfluous it doesn't incentivize us.

Issue #7468, opened 2021-10-08
Building w/OpenMP on macOS


jakirkham:
This changeset ( https://github.com/numba/numba/commit/a23bc195096178e5849ebc0fb6dcedbd6f0cbeea ) was added to fix building on macOS previously. While packaging 0.54.1, we needed to switch to `llvm-openmp`. This seemed to work well despite issues previously. Given this, am wondering do we still need the custom handling here for macOS?

stuartarchibald:
This PR attempts to remove the `intel-openmp` dependency: https://github.com/numba/numba/pull/7786 will have to see how it does in testing etc.

PR #7779, opened 2022-01-24
M1 build script changes

seibert:
Similar to https://github.com/numba/llvmlite/pull/813

### DIFF
--- a/buildscripts/condarecipe.local/meta.yaml
+++ b/buildscripts/condarecipe.local/meta.yaml
@@ -29,7 +29,7 @@ requirements:
     - {{ compiler('cxx') }}    # [not (armv6l or armv7l or aarch64)]
     # both of these are needed on osx, llvm for the headers, Intel for the lib
     - llvm-openmp              # [osx]
-    - intel-openmp             # [osx]
+    - intel-openmp             # [osx and x86_64]
   host:
     - python
     - numpy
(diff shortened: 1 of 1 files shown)


buildscripts/condarecipe.local/meta.yaml
```diff
@@ -29,7 +29,7 @@ requirements:
     - {{ compiler('cxx') }}    # [not (armv6l or armv7l or aarch64)]
     # both of these are needed on osx, llvm for the headers, Intel for the lib
     - llvm-openmp              # [osx]
-    - intel-openmp             # [osx]
+    - intel-openmp             # [osx and x86_64]
```
stuartarchibald, line 32: xref: https://github.com/numba/numba/issues/7468 there's a reasonably high chance that this dep can just be deleted?

stuartarchibald:
https://github.com/numba/numba/pull/7786 attempts to remove the `intel-openmp` dependency.

seibert:
OK, great.  I'll let that one finish review and merge before updating this PR.

seibert:
I should also note that a few Numba tests fail on my M1, but I don't think those are build related.  (A few look like BLAS issues on ARM, so look forward to that...)

esc:
This branch is now conflicting against `master`.

esc:
I think this pr can be abandoned. It turns out that we can build fine when all the underlying changes to llvmdev and llvmlite have been applied. However, trying to compile and run and test Numba locally on an M1 machine yields multiple errors, which I will open a separate ticket for.

stuartarchibald:
I think https://github.com/numba/numba/pull/7786 renders this patch unneeded as the `intel-openmp` is no longer a dependency.

PR #4586, opened 2019-09-18
[WIP] Fix Set for Unicode items

ehsantn:
As title. Adds reference counting to add/remove of Set items, and fixes function resolution for `eq`.

PR #7780, opened 2022-01-24
Fix sets with reference counted items

benwilliamgraham:
Continuation of https://github.com/numba/numba/pull/4586

### DIFF
--- a/numba/cpython/setobj.py
+++ b/numba/cpython/setobj.py
@@ -461,10 +461,11 @@ class SetInstance(object):
                 self.upsize(used)
             self.set_dirty(True)
 
-    def _remove_entry(self, payload, entry, do_resize=True):
+    def _remove_entry(self, payload, entry, do_resize=True, do_decref=True):
         # Mark entry deleted
         entry.hash = ir.Constant(entry.hash.type, DELETED)
-        self.decref_value(entry.key)
+        if do_decref:
+            self.decref_value(entry.key)
         # used--
         used = payload.used
         one = ir.Constant(used.type, 1)
@@ -551,9 +552,8 @@ class SetInstance(object):
         payload = self.payload
         with payload._next_entry() as entry:
             builder.store(entry.key, key)
-            # incref since the value is returned but _remove_entry() decrefs
-            self.incref_value(entry.key)
-            self._remove_entry(payload, entry)
+            # since the value is returned don't decref in _remove_entry()
+            self._remove_entry(payload, entry, do_decref=False)
 
         return builder.load(key)
 
--- a/numba/cpython/setobj.py
+++ b/numba/cpython/setobj.py
@@ -1246,7 +1246,10 @@ def set_constructor(context, builder, sig, args):
 
     # If the argument has a len(), preallocate the set so as to
     # avoid resizes.
-    # `for_iter` increfs each item in the set, so don't incref in add
+    # `for_iter` increfs each item in the set, so a `decref` is required each
+    # iteration to balance. Because the `incref` from `.add` is dependant on
+    # the item not already existing in the set, just removing its incref is not
+    # enough to guarantee all memory is freed
     n = call_len(context, builder, items_type, items)
     inst = SetInstance.allocate(context, builder, set_type, n)
     with for_iter(context, builder, items_type, items) as loop:
@@ -1383,8 +1386,10 @@ def set_update(context, builder, sig, args):
         # set instance
         casted = context.cast(builder, loop.value, items_type.dtype, inst.dtype)
         inst.add(casted)
-        # decref each item to counter balance the double incref from `add` +
-        # `for_iter`
+        # decref each item to counter balance the incref from `for_iter`
+        # `.add` will conditionally incref when the item does not already exist
+        # in the set, therefore removing its incref is not enough to guarantee
+        # all memory is freed
         context.nrt.decref(builder, items_type.dtype, loop.value)
 
     if n is not None:
--- a/numba/cpython/setobj.py
+++ b/numba/cpython/setobj.py
@@ -418,7 +418,7 @@ class SetInstance(object):
 
         old_hash = entry.hash
         entry.hash = h
-        context.nrt.incref(builder, self._ty.dtype, item)
+        self.incref_value(item)
         entry.key = item
         # used++
         used = payload.used
@@ -445,7 +445,7 @@ class SetInstance(object):
             entry = payload.get_entry(i)
             old_hash = entry.hash
             entry.hash = h
-            context.nrt.incref(builder, self._ty.dtype, item)
+            self.incref_value(item)
             entry.key = item
             # used++
             used = payload.used
@@ -463,7 +463,7 @@ class SetInstance(object):
     def _remove_entry(self, payload, entry, do_resize=True):
         # Mark entry deleted
         entry.hash = ir.Constant(entry.hash.type, DELETED)
-        self._context.nrt.decref(self._builder, self._ty.dtype, entry.key)
+        self.decref_value(entry.key)
         # used--
         used = payload.used
         one = ir.Constant(used.type, 1)
@@ -551,7 +551,7 @@ class SetInstance(object):
         with payload._next_entry() as entry:
             builder.store(entry.key, key)
             # incref since the value is returned but _remove_entry() decrefs
-            context.nrt.incref(builder, self._ty.dtype, entry.key)
+            self.incref_value(entry.key)
             self._remove_entry(payload, entry)
 
         return builder.load(key)
@@ -953,7 +953,7 @@ class SetInstance(object):
             entry = loop.entry
             self._add_key(payload, entry.key, entry.hash,
                           do_resize=False)
-            context.nrt.decref(builder, self._ty.dtype, entry.key)
+            self.decref_value(entry.key)
 
         self._free_payload(old_payload.ptr)
 
@@ -970,7 +970,7 @@ class SetInstance(object):
         # decref all of the previous entries
         with self.payload._iterate() as loop:
             entry = loop.entry
-            context.nrt.decref(builder, self._ty.dtype, entry.key)
+            self.decref_value(entry.key)
 
         # Free old payload
         self._free_payload(self.payload.ptr)
@@ -1017,7 +1017,7 @@ class SetInstance(object):
                 alloc_ok = cgutils.is_null(builder, ptr)
             else:
                 # create destructor to be called upon set destruction
-                dtor = _imp_dtor(context, builder.module, self._ty)
+                dtor = self._imp_dtor(context, builder.module)
                 meminfo = context.nrt.meminfo_new_varsize_dtor(
                     builder, allocsize, builder.bitcast(dtor, cgutils.voidptr_t))
                 alloc_ok = cgutils.is_null(builder, meminfo)
@@ -1083,7 +1083,7 @@ class SetInstance(object):
 
         with builder.if_then(builder.load(ok), likely=True):
             # create destructor for new meminfo
-            dtor = _imp_dtor(context, builder.module, self._ty)
+            dtor = self._imp_dtor(context, builder.module)
             meminfo = context.nrt.meminfo_new_varsize_dtor(builder, allocsize, builder.bitcast(dtor, cgutils.voidptr_t))
             alloc_ok = cgutils.is_null(builder, meminfo)
 
@@ -1107,7 +1107,7 @@ class SetInstance(object):
                     # increment the refcounts to simulate `_add_key` for each
                     # element
                     with src_payload._iterate() as loop:
-                        context.nrt.incref(builder, self._ty.dtype, loop.entry.key)
+                        self.incref_value(loop.entry.key)
 
                     if DEBUG_ALLOCS:
                         context.printf(builder,
@@ -1116,34 +1116,44 @@ class SetInstance(object):
 
         return builder.load(ok)
 
-def _imp_dtor(context, module, set_type):
-    """Define the dtor for set
-    """
-    llvoidptr = context.get_value_type(types.voidptr)
-    llsize = context.get_value_type(types.uintp)
-    # create a dtor function that takes (void* set, size_t size, void* dtor_info)
-    fnty = ir.FunctionType(
-        ir.VoidType(),
-        [llvoidptr, llsize, llvoidptr],
-    )
-    # create type-specific name
-    fname = f"_numba_set_dtor_{set_type}"
+    def _imp_dtor(self, context, module):
+        """Define the dtor for set
+        """
+        llvoidptr = cgutils.voidptr_t
+        llsize_t= context.get_value_type(types.size_t)
+        # create a dtor function that takes (void* set, size_t size, void* dtor_info)
+        fnty = ir.FunctionType(
+            ir.VoidType(),
+            [llvoidptr, llsize_t, llvoidptr],
+        )
+        # create type-specific name
+        fname = f".dtor.set.{self._ty.dtype}"
+    
+        fn = cgutils.get_or_insert_function(module, fnty, name=fname)
+    
+        if fn.is_declaration:
+            # Set linkage
+            fn.linkage = 'linkonce_odr'
+            # Define
+            builder = ir.IRBuilder(fn.append_basic_block())
+            payload = _SetPayload(context, builder, self._ty, fn.args[0])
+            with payload._iterate() as loop:
+                entry = loop.entry
+                self.decref_value(entry.key)
+            builder.ret_void()
+    
+        return fn
 
-    fn = cgutils.get_or_insert_function(module, fnty, name=fname)
+    def incref_value(self, val):
+        """Incref an element value
+        """
+        self._context.nrt.incref(self._builder, self._ty.dtype, val)
 
-    if fn.is_declaration:
-        # Set linkage
-        fn.linkage = 'linkonce_odr'
-        # Define
-        builder = ir.IRBuilder(fn.append_basic_block())
-        set_ptr = builder.bitcast(fn.args[0], cgutils.voidptr_t.as_pointer())
-        payload = _SetPayload(context, builder, set_type, set_ptr)
-        with payload._iterate() as loop:
-            entry = loop.entry
-            context.nrt.decref(builder, set_type.dtype, entry.key)
-        builder.ret_void()
+    def decref_value(self, val):
+        """Decref an element value
+        """
+        self._context.nrt.decref(self._builder, self._ty.dtype, val)
 
-    return fn
 
 class SetIterInstance(object):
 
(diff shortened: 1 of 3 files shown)


numba/cpython/setobj.py
```diff
@@ -1005,11 +1008,13 @@ def _allocate_payload(self, nentries, realloc=False):
         with builder.if_then(builder.load(ok), likely=True):
             if realloc:
                 meminfo = self._set.meminfo
-                ptr = context.nrt.meminfo_varsize_alloc(builder, meminfo,
-                                                        size=allocsize)
+                ptr = context.nrt.meminfo_varsize_realloc(builder, meminfo,
```
njriasan, line 1011: It doesn't seem like alloc_ok is used in either path. It seems like the contents of the builder.if_else should be replaced with `alloc_ok`.

njriasan, line 1011: Shouldn't the meminfo be updated with this new pointer?

numba/cpython/setobj.py
```diff
@@ -952,6 +950,7 @@ def _resize(self, payload, nentries, errmsg):
             entry = loop.entry
             self._add_key(payload, entry.key, entry.hash,
                           do_resize=False)
+            context.nrt.decref(builder, self._ty.dtype, entry.key)
 
         self._free_payload(old_payload.ptr)
```
ehsantn, line 955: This seems to deallocate the pointer after realloc which doesn't seem valid to me. This whole code section isn't necessary since realloc copies the original data: https://en.cppreference.com/w/c/memory/realloc

numba/cpython/setobj.py
```diff
@@ -1089,6 +1094,9 @@ def _copy_payload(self, src_payload):
                     cgutils.raw_memcpy(builder, payload.entries,
                                        src_payload.entries, nentries,
                                        entry_size)
+                    with src_payload._iterate() as loop:
```
ehsantn, line 1097: Shouldn't `meminfo_new_varsize` above be replaced with `meminfo_new_varsize_dtor` similar to the original PR?

njriasan:
Summarizing our discussion here @benwilliamgraham. The memory leak in the tests seems to result from the iterable constructor with either duplicates or returning a length. We don't see the same issues when adding each element, which suggests to me its an issue with how we handle references when iterating over the iterable type.

njriasan:
@benwilliamgraham Here is a simple reproducer from yesterday:

```python3
import gc
import numba
from numba.core.runtime import rtsys

@numba.njit
def f(lst):
    s = set(lst)
    return len(s)

lst = ["a"]


gc.collect()
old = rtsys.get_allocation_stats()
f(lst)
gc.collect()
new = rtsys.get_allocation_stats()
old_stats = [old]
new_stats = [new]
total_alloc = sum([m[0] for m in new_stats]) - sum([m[0] for m in old_stats])
total_free = sum([m[1] for m in new_stats]) - sum([m[1] for m in old_stats])
total_mi_alloc = sum([m[2] for m in new_stats]) - sum([m[2] for m in old_stats])
total_mi_free = sum([m[3] for m in new_stats]) - sum([m[3] for m in old_stats])
print(total_alloc, total_free)
assert total_alloc == total_free
assert total_mi_alloc == total_mi_free
```

numba/cpython/setobj.py
```diff
@@ -946,12 +944,13 @@ def _resize(self, payload, nentries, errmsg):
             context.call_conv.return_user_exc(builder, MemoryError,
                                               (errmsg,))
 
-        # Re-insert old entries
+        # Re-insert old entries, decrefing each one due to `_iterate`'s increfs
```
njriasan, line 947: I might add, that `_add_key` adds an incref, so we remove the extra incref from `_iterate`

numba/cpython/setobj.py
```diff
@@ -960,11 +959,16 @@ def _replace_payload(self, nentries):
         Replace the payload with a new empty payload with the given number
         of entries.
 
-        CAUTION: *nentries* must be a power of 2!
+        eAUTION: *nentries* must be a power of 2!
```
njriasan, line 962: This seems like a typo.

numba/cpython/setobj.py
```diff
@@ -1072,7 +1079,9 @@ def _copy_payload(self, src_payload):
                                             nentries))
 
         with builder.if_then(builder.load(ok), likely=True):
-            meminfo = context.nrt.meminfo_new_varsize(builder, size=allocsize)
+            # create destructor for new meminfo
+            dtor = _imp_dtor(context, builder.module, self._ty)
+            meminfo = context.nrt.meminfo_new_varsize_dtor(builder, allocsize, builder.bitcast(dtor, cgutils.voidptr_t))
             alloc_ok = cgutils.is_null(builder, meminfo)
 
             with builder.if_else(cgutils.is_null(builder, meminfo),
```
njriasan, line 1087: Replace this is_null with alloc_ok as well.

numba/cpython/setobj.py
```diff
@@ -1089,6 +1098,9 @@ def _copy_payload(self, src_payload):
                     cgutils.raw_memcpy(builder, payload.entries,
                                        src_payload.entries, nentries,
                                        entry_size)
+                    with src_payload._iterate() as loop:
+                        entry = loop.entry
+                        context.nrt.incref(builder, self._ty.dtype, entry.key)
```
njriasan, line 1103: Is this case tested? I think this is correct, but I would add a comment about whey you aren't using `_add_key`.

numba/cpython/setobj.py
```diff
@@ -1097,6 +1109,33 @@ def _copy_payload(self, src_payload):
...
         return builder.load(ok)
 
+def _imp_dtor(context, module, set_type):
+    """Define the dtor for set
+    """
+    llvoidptr = context.get_value_type(types.voidptr)
+    llsize = context.get_value_type(types.uintp)
+    fnty = ir.FunctionType(
+        ir.VoidType(),
+        [llvoidptr, llsize, llvoidptr],
```
njriasan, line 1119: When we were reviewing this code we found determining the arguments confusing, so it might help to explain what the arguments are when you define the function type.

numba/cpython/setobj.py
```diff
@@ -1188,10 +1227,12 @@ def set_constructor(context, builder, sig, args):
 
     # If the argument has a len(), preallocate the set so as to
     # avoid resizes.
+    # decref each item to avoid leak from iterator's incref
```
njriasan, line 1230: I be more specific in this comment and say something like both `add` and `for_iter` incref each element, so we add a decref to avoid a leak.

numba/cpython/setobj.py
```diff
@@ -1323,6 +1364,7 @@ def set_update(context, builder, sig, args):
         # set instance
         casted = context.cast(builder, loop.value, items_type.dtype, inst.dtype)
         inst.add(casted)
+        context.nrt.decref(builder, items_type.dtype, loop.value)
```
njriasan, line 1367: I think you need the same comment above for_iter as in the previous constructor.

numba/tests/test_sets.py
```diff
@@ -666,6 +653,14 @@ def _range(self, stop):
         return list(zip(b, c, d))
 
 
+class TestUnicodeSets(OtherTypesTest, BaseTest):
```
njriasan, line 656: I would like to see a couple additional tests to check all the cases where incref/decrefs are being inserted. Can you verify which methods are tested and add additional tests as need?

njriasan (commented):
@benwilliamgraham I left some comments. Overall it looks really good and probably just needs some extra comments and tests.

In addition could you test the two set bugs the Numba people sent us (they should be in attached in our JIRA). I'm not sure if they will be fixed by this PR, but it would be nice if they were.

ehsantn (commented):
Thanks for the effort so far. See comments. I think this needs more unittests. A couple of issues I found:

- Unboxing doesn't work for me:

```python
@numba.njit
def f(s):
    return len(s)
```
```
TypeError: cannot store {i8*, i32, i8*}* to i8*: mismatching types
```

- Does `iternext` need an incref? Doesn't look like `List` does it but I'm not sure. Can't produce an error.

- Similarly, does `pop` need an incref?

ehsantn:
I think we could also add more comments about the internal data structures (e.g. what `mask`, `used`, `fill` mean).

stuartarchibald:
> @benwilliamgraham I left some comments. Overall it looks really good and probably just needs some extra comments and tests.
> 
> In addition could you test the two set bugs the Numba people sent us (they should be in attached in our JIRA). I'm not sure if they will be fixed by this PR, but it would be nice if they were.

From memory, I don't think this PR will fix these bugs (IIRC https://github.com/numba/numba/issues/6543 is one) as they are probably caused by an issue in the set impl logic, to which this this change is solely additive. However, there is no harm in taking a look. Should they suddenly be fixed, it'll be important to figure out why so as to ensure that the fixes are true fixes opposed to e.g. just moving some corruption to a location where it has no impact on the reproducer(s).

ehsantn:
I agree with @stuartarchibald . This PR won't change the logic so it won't fix other set issues. But now that we went through the pain of understanding the logic, we might as well fix other issues too. I'd keep the scope of this particular PR limited though.

numba/tests/test_sets.py
```diff
@@ -59,8 +59,8 @@ def iterator_usecase(arg):
 def update_usecase(a, b, c):
     s = set()
     s.update(a)
-    s.update(b)
-    s.update(c)
+    # s.update(b)
+    # s.update(c)
```
ehsantn, line 63: Why are these commented out?

numba/tests/test_sets.py
```diff
@@ -369,52 +369,41 @@ def test_constructor(self):
...
         check(self.duplicates_array(200))
         check(self.sparse_array(200))
 
     def test_set_return(self):
         pyfunc = set_return_usecase
         cfunc = jit(nopython=True)(pyfunc)
 
-        arg = (1, 2, 3, 2, 7)
+        arg = self.duplicates_array(200)
+        print(arg)
```
ehsantn, line 380: remove print?

ehsantn (commented):
Looks like the test file needs a bit of clean up

numba/tests/test_sets.py
```diff
@@ -369,52 +369,40 @@ def test_constructor(self):
...
-    def test_bool(self):
-        pyfunc = bool_usecase
-        check = self.unordered_checker(pyfunc)
-
-        check([1])
-        check([1, 2])
-        check([False, False])
-        check([True, False])
+        import gc
+        gc.collect()
```
ehsantn, line 398: remove?

ehsantn (dismissed):
Thanks for the effort! Looks good to me.

stuartarchibald:
Thanks for the patch and initial reviews/support/advice provided. I've marked this as ready for review now as the build is stable and the PR is out of draft status.

njriasan (approved):
LGTM. Thanks Ben

numba/cpython/setobj.py
```diff
@@ -1072,11 +1082,12 @@ def _copy_payload(self, src_payload):
                                             nentries))
 
         with builder.if_then(builder.load(ok), likely=True):
-            meminfo = context.nrt.meminfo_new_varsize(builder, size=allocsize)
+            # create destructor for new meminfo
+            dtor = _imp_dtor(context, builder.module, self._ty)
+            meminfo = context.nrt.meminfo_new_varsize_dtor(builder, allocsize, builder.bitcast(dtor, cgutils.voidptr_t))
```
stuartarchibald, line 1087: please line wrap this to 80 chars.

numba/tests/test_sets.py
```diff
@@ -554,6 +535,9 @@ def _test_set_operator(self, pyfunc):
         for na, nb in itertools.product(sizes, sizes):
             a = self.sparse_array(na)
             b = self.sparse_array(nb)
+            # empty lists cannot be passed to njit functions due to typing restrictions
```
stuartarchibald, line 538: Same comment as above RE generating non-empty lists.

numba/tests/test_sets.py
```diff
@@ -666,6 +632,14 @@ def _range(self, stop):
         return list(zip(b, c, d))
 
 
+class TestUnicodeSets(TestSets):
+    """
+    Test sets with unicode keys.
```
stuartarchibald, line 637:
```suggestion
    Test sets with unicode keys. For the purpose of testing refcounted items.
```

numba/tests/test_sets.py
```diff
@@ -610,40 +594,22 @@ def test_isub(self):
...
         check = self.unordered_checker(pyfunc)
 
-        a = self.sparse_array(50)
-        b = self.duplicates_array(50)
-        c = self.sparse_array(50)
-        check(a, b, c)
+        check([1])
+        check([1, 2])
+        check([False, False])
+        check([True, False])
```
stuartarchibald, line 609: I think this should have a check with unicode types too.

stuartarchibald (changes requested):
@benwilliamgraham many thanks for working on this and figuring out how to accommodate reference counted items in sets! I've made a few comments on the PR, once addressed this should be good to merge.

I also note in the history of this PR that there is some force-push/rewriting. In future, if you could please try and avoid doing this once a PR has been marked as ready for review/been reviewed it would be much appreciated,. The reason for this is that review is asynchronous and github doesn't do a great job of tracking comments/reviews across force pushes. If you are force pushing to reduce the number of commits, maybe do the rewrites and run of the relevant tests locally and then push once you are happy. Thanks for your help with this.

numba/tests/test_sets.py
```diff
@@ -369,52 +369,38 @@ def test_constructor(self):
...
         check(self.duplicates_array(200))
         check(self.sparse_array(200))
 
     def test_update(self):
         pyfunc = update_usecase
         check = self.unordered_checker(pyfunc)
 
-        a, b, c = (1, 2, 4, 9), (2, 3, 5, 11, 42), (4, 5, 6, 42)
-        check(a, b, c)
-
```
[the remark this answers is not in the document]
benwilliamgraham: Yeah, integer values should be tested under the generic TestSets implementation

numba/cpython/setobj.py
```diff
@@ -947,11 +947,13 @@ def _resize(self, payload, nentries, errmsg):
                                               (errmsg,))
 
         # Re-insert old entries
+        # Decref to balance incref from `_add_key`
```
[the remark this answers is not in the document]
benwilliamgraham: Although I don't think it would ever be an issue because the only time there's the incref/decref combo would be when copying the data from one set to another where the item should always be `not_found`, I do think having a kwarg makes the code simpler and more understandable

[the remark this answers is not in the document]
stuartarchibald: I agree, in this specific case it shouldn't be a problem by the nature of its construction, but in general I think would be. Thanks for adding to the API, seems like it'd make sense to be able to toggle it at the call site.

numba/cpython/setobj.py
```diff
@@ -1188,10 +1235,13 @@ def set_constructor(context, builder, sig, args):
 
     # If the argument has a len(), preallocate the set so as to
     # avoid resizes.
+    # both `add` and `for_iter` incref each item in the set, so manually decref
+    # the items to avoid a leak from the double incref
```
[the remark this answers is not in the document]
benwilliamgraham: Unfortunately in this case and in the `set_update` case the item must be decref-ed because each item could occur multiple times. The decref is more a response to `for_iter`'s incref than it is to `add`'s incref

numba/cpython/setobj.py
```diff
@@ -1323,6 +1373,9 @@ def set_update(context, builder, sig, args):
         # set instance
         casted = context.cast(builder, loop.value, items_type.dtype, inst.dtype)
         inst.add(casted)
+        # decref each item to counter balance the double incref from `add` +
+        # `for_iter`
```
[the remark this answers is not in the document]
stuartarchibald: Has this been addressed?

stuartarchibald (changes requested):
Thanks for the update. There's a couple more cases where I think a kwarg indicating whether to incref/decref should be added to elide the need for handling a "double count", other than that, patch looks good. Many thanks for working on this.

numba/cpython/setobj.py
```diff
@@ -1188,10 +1246,12 @@ def set_constructor(context, builder, sig, args):
 
     # If the argument has a len(), preallocate the set so as to
     # avoid resizes.
+    # `for_iter` increfs each item in the set, so don't incref in add
```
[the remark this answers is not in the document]
benwilliamgraham: I don't think it would ever be used, because here the decref is necessary. This is because the constructor could be a list of repeating items, in which case we would want the number of increfs to be 1 because the set only contains the item once. Therefore, the `decref` is needed to balance the `incref` from `for_iter`, and the incref from `.add` will only occur the one time a repeated item is added to the set and skipped every other time because the item is already in the set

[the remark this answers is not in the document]
stuartarchibald: Thanks for the explanation, this makes sense. Please could the comment be updated with/to reflect this explanation, many thanks!

numba/cpython/setobj.py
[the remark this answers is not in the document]
stuartarchibald: As per: https://github.com/numba/numba/pull/7780#discussion_r815796769 given the explanation here is likely similar, please could the comment be updated similarly. Many thanks!

numba/cpython/setobj.py
```diff
@@ -1188,10 +1246,15 @@ def set_constructor(context, builder, sig, args):
 
     # If the argument has a len(), preallocate the set so as to
     # avoid resizes.
+    # `for_iter` increfs each item in the set, so a `decref` is required each
+    # iteration to balance. Because the `incref` from `.add` is dependant on
```
stuartarchibald, line 1250:
```suggestion
    # iteration to balance. Because the `incref` from `.add` is dependent on
```

numba/tests/test_sets.py
```diff
@@ -544,13 +519,23 @@ def test_copy(self):
...
 
+    def test_bool(self):
+        pyfunc = bool_usecase
+        check = self.unordered_checker(pyfunc)
+
+        check(self.sparse_array(1))
+        check(self.sparse_array(2))
+
+    def _is_empty_list(self, a):
+        return isinstance(a, list) and len(a) == 0
```
stuartarchibald, line 530: Dead code?

stuartarchibald (commented):
Thanks for the fix-ups @benwilliamgraham, this patch looks good, just need to remove (what I think is) this bit of dead code: https://github.com/numba/numba/pull/7780#discussion_r816046163 and then this can be marked for merge. Thanks again for your help with this!

numba/tests/test_sets.py
benwilliamgraham: Good catch!

PR #7782, opened 2022-01-25
adding reminder to check on deprecations

esc:
As title


stuartarchibald (approved):
Thanks for the patch.

PR #7783, opened 2022-01-25
remove upper limit on Python version

esc:
As suggested on
https://discuss.python.org/t/requires-python-upper-limits/12663 and on

https://github.com/numba/llvmlite/issues/693#issuecomment-1011411673

we should not use `Requires-Python` to set an upper limit.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -400,7 +400,7 @@ metadata = dict(
     packages=packages,
     setup_requires=build_requires,
     install_requires=install_requires,
-    python_requires=">={},<{}".format(min_python_version, max_python_version),
+    python_requires=">={}".format(min_python_version),
     license="BSD",
     cmdclass=cmdclass,
 )


esc:
@henryiii can you take a look at this one when you have a minute? Thank you!

henryiii:
This looks good to me, though it needs to happen on llvmlite, too. https://github.com/numba/llvmlite/blob/aeddf447d4befc336b26bde8c46042553a13cf75/setup.py#L226-L227 Though even without that, this is probably better than the current behavior.

stuartarchibald:
Any suggestions for how to test this?

henryiii:
You can do a one-time test by using `pip download` to grab a couple of existing releases (one with the limit, one before the limit) and put it in a wheelhouse folder. You can also make a wheel for this PR with `python -m build --wheel` or `pip wheel .` and drop that into the wheelhouse folder too. Then try to pip install with `--no-index` & `--find-links=<wheelhouse>`.

You could bundle these into a fixture for PyTest, which I've done for other things, but this is not something that is probably worth adding to continuous testing for this? See https://github.com/scikit-build/scikit-build/blob/47a61e7aca8e4b37f390503df49363013829f81a/tests/conftest.py#L14-L29.

esc:
The `llvmlite` followup is here: https://github.com/numba/llvmlite/pull/817

sklam:
https://github.com/numba/llvmlite/pull/817 merged

esc:
@sklam will this one be merged soon too?

sklam (approved):
LGTM

Issue #3116, opened 2018-07-15
np.norm axis support


vyasr:
Following up on issues #3114 and #3115, another function that is not yet implemented is the axis argument to numpy.norm. The ability to specify an axis can be useful when computing norms over high-dimensional arrays. To see how it might be used, see the file rowan/functions in the [rowan repository](https://bitbucket.org/glotzer/rowan/src)

stuartarchibald:
Thanks, this is a duplicate of https://github.com/numba/numba/issues/2558 (assuming `np.linalg.norm` is meant?)?

vyasr:
Yes you are right. Feel free to close this one, I just added a few in bulk when I was discussing with Stan and the others didn't appear to have prior issues, so I didn't check this one. Thanks!

stuartarchibald:
Great, thanks :)

Issue #2558, opened 2017-10-09
support np.linalg.norm


nunocalaim:
as in the title! Thanks

stuartarchibald:
Thanks for the request, I've edited the title to reflect your comment as vanilla `np.linalg.norm` is supported.

nunocalaim:
In the meantime can you point to a way to normalize columns in a numpy array

normal numpy code:

```
F = np.random.randn(4, 5)
F /= np.linalg.norm(F, axis = 0)
```

stuartarchibald:
How about `impl1` defined below, does that what you want?
```python
import numpy as np
from numba import njit
np.random.seed(0)

A = np.random.randn(4, 5)

@njit
def impl1(A):
    F = A.copy()
    for i in range(F.shape[1]):
        nrm = np.linalg.norm(F[:, i])
        F[:, i] /= nrm
    return F

def impl2(A):
    F = A.copy()
    F = F / np.linalg.norm(F, axis=0)
    return F

out1 = impl1(A)
out2 = impl2(A)
np.testing.assert_allclose(out1, out2)
```

seibert:
Is the copy necessary since the original code modifies the array in place anyway?

stuartarchibald:
The copy doesn't need to be in the function, just did it because it was convenient :) I'll hoist it and update the gist.

flying-sheep:
Also `a` should be `x` as in numba, some people pass everything as named argument.

wtfzambo:
Yes please, this feat would be great to have.

a7744hsc:
> ```python
>     for i in range(F.shape[1]):
>         nrm = np.linalg.norm(F[:, i])
>         F[:, i] /= nrm
> ```

Thanks for your solution, It works fine, but very slowly. In my case(10 core 20threads),  it is slower than a single thread numpy.
```
dimension = 256
data_size = 100_000
query_size = 100
np.random.seed(1234)
retrieval_k = 10
data_vec = np.random.random((data_size, dimension)).astype(np.float32)
query_vec = np.random.random((query_size, dimension)).astype(np.float32)


# numpy 
for query in query_vec[:50]:
    search_index = np.argpartition(np.linalg.norm(query-data_vec,axis=1),retrieval_k)

# numba
@njit(parallel=True)
def numba_func(q,d):
    first = True
    for query in q[:50]:
        residual = query-d
        normd_array = np.empty(residual.shape[0])
        for i in range(residual.shape[0]):
[...]

Issue #2181, opened 2016-10-31
np.linalg.norm() does not accept axis argument in nopython mode 


breedlun:
When I try to run the following

    @jit(nopython = True)
    def numba_func(x)
        x_norm = np.linalg.norm(dx, axis = 1)
        return(x_norm)

I get `InternalError: norm_impl() got an unexpected keyword argument 'axis'`

I am running numba 0.28.1 in python 2.7.9.

stuartarchibald:
Thanks for the report. This is true at present, as per the docs http://numba.pydata.org/numba-doc/0.28.1/reference/numpysupported.html#linear-algebra
`numpy.linalg.norm() (only the 2 first arguments and only non string values in ord).`
Most `numpy.linalg` support is basic at present as it's only been around for a short while.

Should you develop a fix for this, patches are most welcome :-)

Thanks again...

stuartarchibald:
duplicated https://github.com/numba/numba/issues/2558

PR #7785, opened 2022-01-26
Add axis option to norm computation

PABannier:
The goal of this PR is to add the axis option to the `np.linalg.norm` function in NoPython mode. 

Some issues request the support of this feature: #2181 #2558 #3116

Roadmap: 
- [x] Implement axis option for `ord=None`
- [x] Implement axis option for other orders
- [x] Add unit tests

### DIFF
--- a/numba/np/linalg.py
+++ b/numba/np/linalg.py
@@ -2309,7 +2309,7 @@ def _get_norm_impl(a, ord_flag, axis):
                     m = a.shape[-2]
 
                     if a.size == 0:
-                        return np.zeros(1, dtype=a.dtype)
+                        return np.zeros(1, dtype=np_ret_type)
 
                     if axis not in (0, 1):
                         raise ValueError("Invalid axis for array.")
@@ -2317,7 +2317,7 @@ def _get_norm_impl(a, ord_flag, axis):
                     if ord == np.inf and axis == 0:
                         # max of abs across rows
                         # max(abs(a), axis=0)
-                        axis_nrms = np.zeros(n)
+                        axis_nrms = np.zeros(n, dtype=np_ret_type)
                         for j in range(n):
                             for i in range(m):
                                 axis_nrms[j] = max(abs(a[i, j]), axis_nrms[j])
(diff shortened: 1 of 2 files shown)


PABannier:
So far, the norm implemented in Numba returns a scalar. Passing an axis argument would potentially make the norm function return a 1d array (e.g. norm of the rows/columns). Since the function signature must be unique, I can’t have 2 return types: scalar and 1d-array.

stuartarchibald:
> 

It's not possible to have return types that won't unify returned by a given implementation, but it is possible to dispatch different implementations based on type. Example:
```python
from numba import njit, types
import numpy as np
from numba.np.numpy_support import is_nonelike

def something(*args):
    pass

@overload(something)
def ol_something(array, axis=None):
    # do type checking etc

    if is_nonelike(axis):
        def impl(array, axis=None):
            print("Doing something with a None axis")
            return 1.234 # scalar return
        return impl
    elif isinstance(axis, (types.Integer, int)):
        def impl(array, axis=None):
            print("Doing something with an int axis")
            return np.ones((6,)) # 1d array return
        return impl

@njit
def foo(axis):
    return something(np.ones(1), axis)

print(foo(None))
print(foo(1))
```
Hope this helps? There's also an example here (which so happens to be `norm`): 
https://numba.readthedocs.io/en/stable/extending/overloading-guide.html

PABannier:
@stuartarchibald awesome! thanks!

PABannier:
Quick update. All the tests pass in C order. Next steps are:

- [ ] Make norm computation work with Fortran order
- [ ] Write the -1 and -2 norms

-2 norm is a bit tricky since `_compute_singular_values` is only acceptable for 2d arrays. Slicing a 2d array yields a 1d array and adding an axis with `np.newaxis` makes Numba raise an error.

PABannier:
@stuartarchibald @esc @sklam I'm a new contributor. Do I need to place my PR in a queue before review?

esc:
> @stuartarchibald @esc @sklam I'm a new contributor. Do I need to place my PR in a queue before review?

Thank you for asking about this. The queue is implicit and this PR will be reviewed as soon as review capacity becomes available for this PR.

PABannier:
@esc alright! thanks for the quick response :D

sklam:
We have to bump this to the next release.

sklam:
Note-to-self: implementation may need compressing

PABannier:
@sklam do you want me to compress the implementation? We can discuss it in more details if needed. This feature would still be a very valuable addition if it could be merged 😊

sklam:
@PABannier, a lot of code in `_get_norm_impl` is quite similar. Would be worth finding common patterns there.

Felice27:
This would be really helpful to add; I encountered this exact issue in a project I'm working on.

plutonium-239:
Hi, I would also like to accept this PR since `norm` is used in a lot of places. 
Is the repetition of code the only problem being faced here?

PR #7799, opened 2022-01-28
update the release checklist post 0.55.1

esc:
Changed the order of items and re-worded the bullet for cleaning out
c_artifacts. "clean out" doesn't mean "remove", but instead "move files to
subdirectory".


stuartarchibald (approved):
Thanks for making these changes to the release check list. Patch looks good.

PR #7814, opened 2022-02-04
CUDA Dispatcher refactor

gmarkall:
This PR moves classes around in the CUDA target to better mirror the structure of the CPU target and core of Numba (see #7739)

Fixes #7739.

PR #7815, opened 2022-02-04
CUDA Dispatcher refactor 2: inherit from `dispatcher.Dispatcher`

### LINKED ISSUE #5902: cuda.jit does NOT preserve the original __doc__ and __module__ of the decorated function

### LINKED ISSUE #7741: Refactor the CUDA target `Dispatcher` class to inherit from `numba.core.dispatcher::Dispatcher`
The CUDA target dispatcher class currently inherits from `_Dispatcher`, this task is to extend this up one conceptual level such that inherits from `numba.core.dispatcher::Dispatcher`. By doing this the CPU and CUDA targets will be much more aligned, further the CUDA target will inherit a lot of the functionality that is available from the `Dispatcher` class itself. It will also reduce maintenance burden as the CUDA target will carry less custom code.

For simplicity, this depends on https://github.com/numba/numba/issues/7739

gmarkall:
(Note: based on #7814 - it should be easier to tell the differences once that is merged - in the meantime [this comparison](https://github.com/gmarkall/numba/compare/dispatcher-refactor-20220204...gmarkall:cuda-dispatcher-base-20220204) shows the changes unique to this PR)

The main accomplishment of this PR is to make the CUDA dispatcher inherit from `numba.core.dispatcher.Dispatcher` and reuse as much of its logic as possible, rather than duplicating it (and also to serve as a base for implementing on-disk caching, AOT compilation, better support for the high-level extension API, etc.). 

This necessitates a set of individual changes that accompany the main change:

- Creation of a `CUDACompileResult`, which reports its entry point as its `id` - this is because certain functions expect compile results to have entry points - however, this isn't true for CUDA because you can't enter a CUDA kernel through the C-layer dispatcher. Using the `id` gives each CUDA compile result a unique identifier.
- A small change to `compile_result` to factor out sanitizing the entries that go into the compile result, so that this logic can be shared between the core and the CUDA target.
- Changes to `numba.cuda.decorators` so it is responsible for handling signatures and compilation - the CUDA dispatcher no longer needs to be aware of signatures, and this behaviour mirrors the CPU target. CUDA still only supports one signature, but this can now be more easily changed in future.
- Fixes to the setup of the `extensions` target option, so that the CUDA dispatcher no longer needs to make a strange "defensive copy".
- Two additions to `_Kernel` so that they have the `objectmode` and `entry_point` properties, as if they were compile results - this is because overloads are expected to be compile results by some of the core dispatcher logic.
- Some refactors to `ForAll` that make the terminology more representative (but no functional change).
- Fix a very strange "how did it ever work"-style bug in `resolve_value_type`, where it was wrapping a CPU dispatcher in a CUDA dispatcher instead of wrapping the Python function in a CUDA dispatcher when calling an `@njit` function from a `@cuda.jit` function.
- Add test of docstrings for CUDA dispatchers (#5209).

Fixes #7741.

## Description prior to editing, kept for reference:

Fixes #5902.

To do:

- [X] Resolve the issue with `_make_finalizer` - possibly one of:
  - be certain that a dummy finalizer is fine, 
  - remove the finalization altogether if it's no longer needed (perhaps this is an old Python / Python 2 remnant)
  - implement a custom non-dummy finalizer
  - make `_Kernel` pretend to be a `CompileResult` a bit more by giving it an `entry_point`.
- [X] Add a test that #5902 is fixed (I have only manually verified it so far).

### DIFF
--- a/docs/source/cuda-reference/kernel.rst
+++ b/docs/source/cuda-reference/kernel.rst
@@ -56,7 +56,7 @@ This is similar to launch configuration in CUDA C/C++:
 Dispatcher objects also provide several utility methods for inspection and
 creating a specialized instance:
 
-.. autoclass:: numba.cuda.dispatcher.Dispatcher
+.. autoclass:: numba.cuda.dispatcher.CUDADispatcher
    :members: inspect_asm, inspect_llvm, inspect_sass, inspect_types,
              get_regs_per_thread, specialize, specialized, extensions, forall
 
--- a/numba/core/compiler.py
+++ b/numba/core/compiler.py
@@ -269,20 +269,25 @@ _LowerResult = namedtuple("_LowerResult", [
 ])
 
 
-def compile_result(**kws):
-    keys = set(kws.keys())
+def sanitize_compile_result_entries(entries):
+    keys = set(entries.keys())
     fieldset = set(CR_FIELDS)
     badnames = keys - fieldset
     if badnames:
         raise NameError(*badnames)
     missing = fieldset - keys
     for k in missing:
-        kws[k] = None
+        entries[k] = None
     # Avoid keeping alive traceback variables
-    err = kws['typing_error']
+    err = entries['typing_error']
     if err is not None:
-        kws['typing_error'] = err.with_traceback(None)
-    return CompileResult(**kws)
+        entries['typing_error'] = err.with_traceback(None)
+    return entries
+
+
+def compile_result(**entries):
+    entries = sanitize_compile_result_entries(entries)
+    return CompileResult(**entries)
 
 
 def compile_isolated(func, args, return_type=None, flags=DEFAULT_FLAGS,
--- a/numba/cuda/compiler.py
+++ b/numba/cuda/compiler.py
@@ -1,7 +1,8 @@
 from numba.core.typing.templates import ConcreteTemplate
 from numba.core import types, typing, funcdesc, config, compiler
-from numba.core.compiler import (CompilerBase, DefaultPassBuilder,
-                                 Flags, Option, CompileResult, CR_FIELDS)
+from numba.core.compiler import (sanitize_compile_result_entries, CompilerBase,
+                                 DefaultPassBuilder, Flags, Option,
+                                 CompileResult)
 from numba.core.compiler_lock import global_compiler_lock
 from numba.core.compiler_machinery import (LoweringPass, AnalysisPass,
                                            PassManager, register_pass)
@@ -51,20 +52,9 @@ class CUDACompileResult(CompileResult):
         return id(self)
 
 
-def compile_result(**kws):
-    keys = set(kws.keys())
-    fieldset = set(CR_FIELDS)
-    badnames = keys - fieldset
-    if badnames:
-        raise NameError(*badnames)
-    missing = fieldset - keys
-    for k in missing:
-        kws[k] = None
-    # Avoid keeping alive traceback variables
-    err = kws['typing_error']
-    if err is not None:
-        kws['typing_error'] = err.with_traceback(None)
-    return CUDACompileResult(**kws)
+def cuda_compile_result(**entries):
+    entries = sanitize_compile_result_entries(entries)
+    return CUDACompileResult(**entries)
 
 
 @register_pass(mutates_CFG=True, analysis_only=False)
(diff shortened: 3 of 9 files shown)


gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
> * Resolve the issue with `_make_finalizer`

Regarding this, my present thinking is that the way to go is to give `_Kernel` objects an `entry_point` of `None`, because they were never inserted into the target context. We still should have the finalizer for device functions, because they are inserted into the target context, and are actual compile results with an entry point.

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
@stuartarchibald I've merged main into this PR so the diff is now straightforward (compared to what it was :laughing:)

gpuci run tests

stuartarchibald:
> @stuartarchibald I've merged main into this PR so the diff is now straightforward

Thanks @gmarkall, the patch is much easier too look at now!

stuartarchibald (commented):
Thanks for the patch, great to see all the prior refactoring effort finally manifest in this PR!

Given the logical changes herein were discussed extensively OOB whilst implementing this, I'm of the view that they are correct! The only comments I have are that there's a couple of typos to look at but otherwise this looks good.

I think due to the nature of the change, it'd be good to build this PR locally and run the RAPIDS test suite against it just to provide another layer of confidence on top of GPU CI and the Numba build farm. Thanks again!!

gmarkall:
gpuci run tests

gmarkall:
Status marked as:

* "Waiting on CI" for BuildFarm, and
* "Waiting on author" for a run with RAPIDS.

gmarkall:
This passed a run with RAPIDS, so I'll remove "Waiting on author" - I think it's only waiting on the BuildFarm run now?

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_117`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_117`.

This failed as mainline is currently failing on CUDA due to issues hopefully resolved in https://github.com/numba/numba/pull/7846

gmarkall:
gpuci run tests

gmarkall:
@stuartarchibald I've merged from master and re-tested with CI and gpuCI now that https://github.com/numba/numba/pull/7846 is merged. Could this have another buildfarm run please?

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_119`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_119`.

Passed!

stuartarchibald:
> This passed a run with RAPIDS, so I'll remove "Waiting on author" - I think it's only waiting on the BuildFarm run now?

@gmarkall Thanks for testing this against RAPIDS, as this patch is passing the test suite on an external code base it increases my confidence that we got this right!

The Numba buildfarm also passed, have marked this as ready to merged!

PR #7817, opened 2022-02-07
Update intersphinx URLs for NumPy and llvmlite.

stuartarchibald:
As title.


sklam (approved):
LGTM. Tested doc building locally.

Issue #7824, opened 2022-02-09
`_helperlib.c::import_cython_function` erroneously checks result of `PyCapsule_GetName` against `NULL`


stuartarchibald:
## Reporting a bug

- [x] I have tried using the latest released version of Numba (most recent is
 visible in the change log (https://github.com/numba/numba/blob/main/CHANGE_LOG).
- [x] I have included a self contained code sample to reproduce the problem.
  i.e. it's possible to run as 'python bug.py'.

As title. Reported here: https://github.com/numba/numba/issues/7818#issuecomment-1033649811

The problematic code is:
https://github.com/numba/numba/blob/39f03307113eaf76f0b1308cfbe0c8bf78a5c3e9/numba/_helperlib.c#L565-L568

Variable `capsule_name` should not be checked against `NULL`, `PyCapsule_IsValid` or `PyErr_Occurred` should be used as it is legal for a capsule to have name as `NULL`.

Python docs for the C-API call https://docs.python.org/3/c-api/capsule.html#c.PyCapsule_GetName.

Issue #7818, opened 2022-02-08
Calling f2py wrapped code (add an example to documentation)


max9111:
Especially in scipy quite a lot functions are wrapped using f2py. Fortran functions which are wrapped using f2py always  have an [`_cpointer `](https://numpy.org/devdocs/f2py/python-usage.html)` attribute.
It would be good to show an example in the documentation of the [High-level extension API](https://numba.pydata.org/numba-doc/latest/extending/high-level.html#high-level-extension-api) and maybe also implement some useful functions in the extension API.

**Code to extract the function pointer**

```Python
from ctypes import pythonapi
#https://docs.python.org/3/c-api/capsule.html

def capsule_name(capsule):
    pythonapi.PyCapsule_GetName.restype = ctypes.c_char_p
    pythonapi.PyCapsule_GetName.argtypes = [ctypes.py_object]
    return pythonapi.PyCapsule_GetName(capsule)

def get_f2py_function_address(capsule):
    name = capsule_name(capsule)
    pythonapi.PyCapsule_GetPointer.restype = ctypes.c_void_p
    pythonapi.PyCapsule_GetPointer.argtypes = [ctypes.py_object, ctypes.c_char_p]
    return pythonapi.PyCapsule_GetPointer(capsule, name)
```
**Example on scipy.interpolate.dfitpack.splev**

```Python
import numba as nb
import numpy as np
import ctypes
from scipy import interpolate

dble_p=ctypes.POINTER(ctypes.c_double)
int_p =ctypes.POINTER(ctypes.c_longlong)

#Calling siganture
functype = ctypes.CFUNCTYPE(ctypes.c_void_p,
                            dble_p,int_p,dble_p,int_p,dble_p,dble_p,int_p,int_p,int_p)

#Get function pointer and create ctypes function
func_ptr=get_f2py_function_address(interpolate.dfitpack.splev._cpointer)
SPLEV = functype(func_ptr)


#########Intrinsics to efficiently to pass scalars by ref#############
#Workaround using stack-allocated arrays, faster than heap-allocated arrays.
#This could be relevant on functions with short runtimes.
from numba import types
from numba.extending import intrinsic
from numba.core import cgutils

@intrinsic
def val_to_ptr(typingctx, data):
    def impl(context, builder, signature, args):
        ptr = cgutils.alloca_once_value(builder,args[0])
        return ptr
    sig = types.CPointer(nb.typeof(data).instance_type)(nb.typeof(data).instance_type)
    return sig, impl

[...]

PR #7825, opened 2022-02-09
F2PY Support

max9111:
Especially in scipy quite a lot functions are wrapped using f2py. Fortran functions which are wrapped using f2py always have an [_cpointer ](https://numpy.org/devdocs/f2py/python-usage.html)` attribute.

This pull request adds a new function `get_f2py_function_address`, which should work in the same way as `get_cython_function_address`.

Fix for #7818 and #7824 .

### DIFF
--- a/docs/source/extending/high-level.rst
+++ b/docs/source/extending/high-level.rst
@@ -119,6 +119,52 @@ the function's name will be mangled. To find out the mangled name of
 your function you can check the extension module's ``__pyx_capi__``
 attribute.
 
+Importing F2PY Functions
+--------------------------
+
+The function ``get_f2py_function_address`` obtains the address of a
+C function in a F2PY extension module. The address can be used to
+access the C function via a :func:`ctypes.CFUNCTYPE` callback, thus
+allowing use of the C function inside a Numba jitted function. For
+example, call scipy.interpolate.dfitpack.splev::
+
+    import numba as nb
+    import numpy as np
+    import ctypes
+    from scipy import interpolate
+    from numba.core.extending import get_f2py_function_address
+
+    dble_p=ctypes.POINTER(ctypes.c_double)
+    int_p =ctypes.POINTER(ctypes.c_longlong)
+
+    #Calling siganture
+    functype = ctypes.CFUNCTYPE(ctypes.c_void_p,
+                                dble_p,int_p,dble_p,int_p,dble_p,dble_p,int_p,int_p,int_p)
+
+    #Get function pointer and create ctypes function
+    func_ptr=get_f2py_function_address("scipy.interpolate.dfitpack","splev")
+    SPLEV = functype(func_ptr)
+
+You can access ``SPLEV`` from Numba in the following way::
+
+    from numba.cpython.unsafe.stack_arr import val_to_ptr,ptr_to_val
+
+    #float64[::1] -> ensure contiguous array, call np.ascontiguousarray or 
+    #np.asfortranarray instead to ensure to pass a pointer to a contiguous array
+    @nb.njit('float64[::1](float64[::1],float64[::1],int64,float64[::1],int64)')
+    def splev_wrapped(t,c,k,x,e):
+        y=np.empty(x.shape[0],dtype=np.float64)
+
+        n_arr=val_to_ptr(nb.int64(t.shape[0]))
+        k_arr=val_to_ptr(nb.int64(k))
+        m_arr=val_to_ptr(nb.int64(x.shape[0]))
+        e_arr=val_to_ptr(nb.int64(e))
+        ier_arr=val_to_ptr(nb.int64(0))
+
+        SPLEV(t.ctypes,n_arr,c.ctypes,k_arr,x.ctypes,
+            y.ctypes,m_arr,e_arr,ier_arr)
+        return y
+
 Implementing intrinsics
 -----------------------
 
--- a/docs/source/extending/high-level.rst
+++ b/docs/source/extending/high-level.rst
@@ -126,7 +126,7 @@ The function ``get_f2py_function_address`` obtains the address of a
 C function in a F2PY extension module. The address can be used to
 access the C function via a :func:`ctypes.CFUNCTYPE` callback, thus
 allowing use of the C function inside a Numba jitted function. For
-example, call scipy.interpolate.dfitpack.splev::
+example, call ``scipy.interpolate.dfitpack.splev``::
 
     import numba as nb
     import numpy as np
--- a/docs/source/extending/high-level.rst
+++ b/docs/source/extending/high-level.rst
@@ -137,7 +137,7 @@ example, call scipy.interpolate.dfitpack.splev::
     dble_p=ctypes.POINTER(ctypes.c_double)
     int_p =ctypes.POINTER(ctypes.c_longlong)
 
-    #Calling siganture
+    #Calling signature
     functype = ctypes.CFUNCTYPE(ctypes.c_void_p,
                                 dble_p,int_p,dble_p,int_p,dble_p,dble_p,int_p,int_p,int_p)
 
--- a/numba/_helperlib.c
+++ b/numba/_helperlib.c
@@ -533,7 +533,7 @@ numba_attempt_nocopy_reshape(npy_intp nd, const npy_intp *dims, const npy_intp *
 }
 
 /*
- * Cython utilities.
+ * Cython and f2py utilities
  */
 
 /* Fetch the address of the given function, as exposed by
@@ -563,7 +563,42 @@ import_cython_function(const char *module_name, const char *function_name)
     }
     /* 2.7+ => Cython exports a PyCapsule */
     capsule_name = PyCapsule_GetName(cobj);
-    if (capsule_name != NULL) {
+    if (PyCapsule_IsValid(cobj,capsule_name)!=0) {
+        res = PyCapsule_GetPointer(cobj, capsule_name);
+    }
+    Py_DECREF(cobj);
+    return res;
+}
+
+static void *
+import_f2py_function(const char *module_name, const char *function_name)
+{
+    PyObject *module,*func, *cobj;
+    void *res = NULL;
+    const char *capsule_name;
+
+    module = PyImport_ImportModule(module_name);
+    if (module == NULL)
+        return NULL;
+    
+    func = PyObject_GetAttrString(module, (char *)function_name);
+    Py_DECREF(module);
+    if (func == NULL)
+        return NULL;
+    
+    cobj = PyObject_GetAttrString(func, "_cpointer");
+    Py_DECREF(func);
+    
+    if (cobj == NULL) {
+        PyErr_Clear();
+        PyErr_Format(PyExc_ValueError,
+                     "No _cpointer attribute found in function '%s' of '%s'",
+                     function_name,module_name);
+        return NULL;
+    }
+    /* f2py exports PyCapsule */
+    capsule_name = PyCapsule_GetName(cobj);
+    if (PyCapsule_IsValid(cobj,capsule_name)!=0){
         res = PyCapsule_GetPointer(cobj, capsule_name);
     }
     Py_DECREF(cobj);
@@ -582,6 +617,31 @@ _numba_import_cython_function(PyObject *self, PyObject *args)
         return NULL;
     }
     p = import_cython_function(module_name, function_name);
+    if (p == NULL) {
+        return NULL;
+    }
+    res = PyLong_FromVoidPtr(p);
+    if (res == NULL) {
+      PyErr_SetString(PyExc_RuntimeError,
+                      "Could not convert function address to int");
+      return NULL;
+    }
+    return res;
+}
+
+NUMBA_EXPORT_FUNC(PyObject *)
+_numba_import_f2py_function(PyObject *self, PyObject *args)
+{
+    const char *module_name;
+    const char *function_name;
+    
+    void *p = NULL;
+    PyObject *res;
+
+    if (!PyArg_ParseTuple(args, "ss", &module_name, &function_name)) {
+        return NULL;
+    }
+    p = import_f2py_function(module_name, function_name);
     if (p == NULL) {
         return NULL;
     }
--- a/numba/_helperlib.c
+++ b/numba/_helperlib.c
@@ -563,7 +563,7 @@ import_cython_function(const char *module_name, const char *function_name)
     }
     /* 2.7+ => Cython exports a PyCapsule */
     capsule_name = PyCapsule_GetName(cobj);
-    if (PyCapsule_IsValid(cobj,capsule_name)!=0) {
+    if (PyCapsule_IsValid(cobj, capsule_name) != 0) {
         res = PyCapsule_GetPointer(cobj, capsule_name);
     }
     Py_DECREF(cobj);
--- a/numba/_helperlib.c
+++ b/numba/_helperlib.c
@@ -593,7 +593,7 @@ import_f2py_function(const char *module_name, const char *function_name)
         PyErr_Clear();
         PyErr_Format(PyExc_ValueError,
                      "No _cpointer attribute found in function '%s' of '%s'",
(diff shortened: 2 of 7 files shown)


max9111:
@stuartarchibald 

Regarding to #7825 I have a few questions

- Should a general method be also implemented to get a function address of a PyCapsule (more or less ctypes code)?
- Especially when wrapping fortran code (pass by reference), something like `val_to_ptr` and `ptr_to_val` could be seen as an indirect dependency. This is also connected to an official support of arrays allocated on the stack.  I am aware of the np.array workaround, but when calling functions with short runtimes this adds quite a lot of unnecessary overhead which could be higher than the runtime of the called function (maybe in extension.py)
- Is there some place for a wrapper generator, which just gives a basic idea of how to wrap eg. Fortran functions (not expected to work automatically due to multiple expected bugs/missing features, but only to get a basic idea on how a wrapper for a whole package may look like)?

stuartarchibald:
> Thanks for the review @stuartarchibald

No problem.

> Regarding to #7825 I have a few questions
> 
>     * Should a general method be also implemented to get a function address of a PyCapsule (more or less ctypes code)?

Could you describe the motivation/use case for this please?

>     * Especially when wrapping fortran code (pass by reference), something like `val_to_ptr` and `ptr_to_val` could be seen as an indirect dependency. This is also connected to an official support of arrays allocated on the stack.  I am aware of the np.array workaround, but when calling functions with short runtimes this adds quite a lot of unnecessary overhead which could be higher than the runtime of the called function (maybe in extension.py)

hmmm, this is like `ctypes.byref`? I wonder longer term if it'd make sense to extend the `ctypes` support in Numba a bit such that `ctypes.byref` works in JIT compiled code. The `ctypes` type constructors would also probably need overloading too such that e.g. `ctypes.byref(ctypes.c_int(some_fortran_parameter))` would work. For now, it might be best to add those two stack based helper methods to `numba.cpython.unsafe` or similar, a future extension to `ctypes` support could then just consume them.

>     * Is there some place for a wrapper generator, which just gives a basic idea of how to wrap eg. Fortran functions (not expected to work automatically due to multiple expected bugs/missing features, but only to get a basic idea on how a wrapper for a whole package may look like)?

Probably in the `numba-extras` repo for now? https://github.com/numba/numba-extras

max9111:
@stuartarchibald The Pull request is ready for review.

numba/cpython/unsafe/stack_arr.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+    def impl(context, builder, signature, args):
+        val = builder.load(args[0])
+        return val
+    sig = data.dtype(types.CPointer(data.dtype))
+    return sig, impl
+
+
+@intrinsic
+def intr_stack_empty_alloc(typingctx,shape,dtype):
```
stuartarchibald, line 35:
```suggestion
def intr_stack_empty_alloc(typingctx, shape, dtype):
```
This also needs a docstring please.

stuartarchibald: Suggest moving this and the `stack_empty` function to a `numba/np/unsafe/stack.py` module. This explicitly concerns arrays-like behaviours?

numba/cpython/unsafe/stack_arr.py
```diff
@@ -0,0 +1,74 @@
...
+    size = 1
+    if isinstance(shape,types.scalars.IntegerLiteral):
+        size = shape.literal_value
+        sig = types.CPointer(dtype.dtype)(types.int64,dtype)
```
stuartarchibald, line 39:
```suggestion
        sig = types.CPointer(dtype.dtype)(types.int64, dtype)
```

numba/cpython/unsafe/stack_arr.py
```diff
@@ -0,0 +1,74 @@
...
+
+    elif isinstance(shape,(types.containers.Tuple,
+                           types.containers.UniTuple)):
```
stuartarchibald, line 42:
```suggestion
    elif isinstance(shape, (types.containers.Tuple,
                            types.containers.UniTuple)):
```

numba/cpython/unsafe/stack_arr.py
```diff
@@ -0,0 +1,74 @@
...
+        for i in range(len(shape)):
+            size *= shape[i].literal_value
```
stuartarchibald, line 44: I think this requires that the container comprises literal values, this should be asserted through the type checks. I'm also not sure how to create a `UniTuple` that comprises literals?

numba/cpython/unsafe/stack_arr.py
```diff
@@ -0,0 +1,74 @@
...
+
+        sig = types.CPointer(dtype.dtype)(typeof(shape).instance_type,dtype)
```
stuartarchibald, line 46:
```suggestion
        sig = types.CPointer(dtype.dtype)(typeof(shape).instance_type, dtype)
```

numba/cpython/unsafe/stack_arr.py
```diff
@@ -0,0 +1,74 @@
...
+    else:
+        raise errors.TypingError(
+            "Shape must be IntegerLiteral " +
+            "or a tuple of IntegerLiterals")
+
+    def impl(context, builder, signature, args):
+        ty = context.get_value_type(dtype.dtype)
```
stuartarchibald, line 53: It seems like there's no check made to ensure that the `dtype` arg is a `dtype`?

numba/cpython/unsafe/stack_arr.py
```diff
@@ -0,0 +1,74 @@
...
+        ptr = cgutils.alloca_once(builder, ty,size=size)
```
stuartarchibald, line 54:
```suggestion
        ptr = cgutils.alloca_once(builder, ty, size=size)
```

numba/_helperlib.c
```diff
@@ -563,7 +563,42 @@ import_cython_function(const char *module_name, const char *function_name)
...
+        return NULL;
+    
+    cobj = PyObject_GetAttrString(func, "_cpointer");
+    Py_DECREF(func);
+    
+    if (cobj == NULL) {
+        PyErr_Clear();
+        PyErr_Format(PyExc_ValueError,
+                     "No _cpointer attribute found in function '%s' of '%s'",
+                     function_name,module_name);
```
stuartarchibald, line 596:
```suggestion
                     function_name, module_name);
```

numba/_helperlib.c
```diff
@@ -563,7 +563,42 @@ import_cython_function(const char *module_name, const char *function_name)
...
+        return NULL;
+    }
+    /* f2py exports PyCapsule */
+    capsule_name = PyCapsule_GetName(cobj);
+    if (PyCapsule_IsValid(cobj,capsule_name)!=0){
```
stuartarchibald, line 601:
```suggestion
    if (PyCapsule_IsValid(cobj, capsule_name) != 0){
```

stuartarchibald: Should an exception be raised if it is not valid?

docs/source/extending/high-level.rst
```diff
@@ -119,6 +119,52 @@ the function's name will be mangled. To find out the mangled name of
...
+        n_arr=val_to_ptr(nb.int64(t.shape[0]))
+        k_arr=val_to_ptr(nb.int64(k))
+        m_arr=val_to_ptr(nb.int64(x.shape[0]))
+        e_arr=val_to_ptr(nb.int64(e))
+        ier_arr=val_to_ptr(nb.int64(0))
+
+        SPLEV(t.ctypes,n_arr,c.ctypes,k_arr,x.ctypes,
+            y.ctypes,m_arr,e_arr,ier_arr)
+        return y
+
```
stuartarchibald, line 167: This may need adjusting with respect to suggested changes in this PR. I think it also makes sense to add this as a unit test in `numba.tests.doc_examples` and then render that straight into the docs, this will ensure that the example use case of this public API is unlikely to be silently broken. Suggest following something like this for writing the test (the `magictoken` comments indicate included region boundaries for the docs):
https://github.com/numba/numba/blob/3cc5e5e97664c65d92172b8f9401e835ce0825ca/numba/tests/doc_examples/test_literal_container_usage.py#L58-L90
and including in the docs like: https://github.com/numba/numba/blob/3cc5e5e97664c65d92172b8f9401e835ce0825ca/docs/source/reference/pysupported.rst?plain=1#L838-L846

numba/_helperlib.c
```diff
@@ -563,7 +563,42 @@ import_cython_function(const char *module_name, const char *function_name)
     }
     /* 2.7+ => Cython exports a PyCapsule */
     capsule_name = PyCapsule_GetName(cobj);
-    if (capsule_name != NULL) {
+    if (PyCapsule_IsValid(cobj,capsule_name)!=0) {
```
stuartarchibald, line 566:
```suggestion
    if (PyCapsule_IsValid(cobj, capsule_name) != 0) {
```
also, should this raise if invalid?

numba/cpython/unsafe/stack_arr.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+    Allocate an array on the stack.
+    Please note: - Arrays allocated on the stack can't be returned.
+                 - Stack size is limited
+    """
+    arr_ptr = intr_stack_empty_alloc(shape,dtype)
+    if order == "C":
+        return carray(arr_ptr,shape)
+    elif order == "F":
+        return farray(arr_ptr,shape)
```
stuartarchibald, line 71:
```suggestion
    if order == "C":
        return carray(arr_ptr, shape)
    elif order == "F":
        return farray(arr_ptr, shape)
```

numba/cpython/unsafe/stack_arr.py
```diff
@@ -0,0 +1,74 @@
...
+
+#inline always, stack memory can't be returned
+@njit(inline="always")
+def stack_empty(shape,dtype,order="C"):
+    """
+    Allocate an array on the stack.
+    Please note: - Arrays allocated on the stack can't be returned.
+                 - Stack size is limited
+    """
+    arr_ptr = intr_stack_empty_alloc(shape,dtype)
```
stuartarchibald, line 67:
```suggestion
    arr_ptr = intr_stack_empty_alloc(shape, dtype)
```

docs/source/extending/high-level.rst
```diff
@@ -119,6 +119,52 @@ the function's name will be mangled. To find out the mangled name of
...
+    import numba as nb
+    import numpy as np
+    import ctypes
+    from scipy import interpolate
+    from numba.core.extending import get_f2py_function_address
+
+    dble_p=ctypes.POINTER(ctypes.c_double)
+    int_p =ctypes.POINTER(ctypes.c_longlong)
+
+    #Calling siganture
```
stuartarchibald, line 140:
```suggestion
    #Calling signature
```

docs/source/extending/high-level.rst
```diff
@@ -119,6 +119,52 @@ the function's name will be mangled. To find out the mangled name of
...
 attribute.
 
+Importing F2PY Functions
+--------------------------
+
+The function ``get_f2py_function_address`` obtains the address of a
+C function in a F2PY extension module. The address can be used to
+access the C function via a :func:`ctypes.CFUNCTYPE` callback, thus
+allowing use of the C function inside a Numba jitted function. For
+example, call scipy.interpolate.dfitpack.splev::
```
stuartarchibald, line 129:
```suggestion
example, call ``scipy.interpolate.dfitpack.splev``::
```

stuartarchibald (commented):
Many thanks for the updates to this patch. I've given this an initial review and it largely looks good. I've made a quite a number of suggestions, the majority of which are stylistic but there are a few to do with code organisation and some more general questions. Thanks again for your work on this.

numba/cpython/unsafe/stack_arr.py
```diff
@@ -0,0 +1,74 @@
...
+    def impl(context, builder, signature, args):
+        ty = context.get_value_type(dtype.dtype)
+        ptr = cgutils.alloca_once(builder, ty,size=size)
+        return ptr
+    return sig, impl
+
+
+#inline always, stack memory can't be returned
+@njit(inline="always")
+def stack_empty(shape,dtype,order="C"):
```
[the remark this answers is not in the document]
max9111: In a stack module, I may would call this function `empty` . There could be also a `zero` function and an `array` function which creates a stack_array from a tuple for example.  On the other hand I think it is quite a good idea to remind the user that he is allocating memory on the stack, which can't be returned. (except workaround with inline="always"). There is no further check if this is happening...

numba/tests/test_extending.py
```diff
@@ -1584,6 +1586,67 @@ def test_missing_function(self):
...
 
 
+class TestImportF2PYFunction(unittest.TestCase):
+    @unittest.skipIf(sc is None, "Only run if SciPy >= 0.19 is installed")
+    def test_getting_function(self):
+        addr = get_f2py_function_address(
+            "scipy.interpolate.dfitpack", "splev"
+        )
+        dble_p = ctypes.POINTER(ctypes.c_double)
+        int_p = ctypes.POINTER(ctypes.c_longlong)
```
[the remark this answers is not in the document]
max9111: [It depends on the BLAS int size.](https://github.com/scipy/scipy/blob/b06254e5a08a0fad05f6e5a09b7c847c893ff355/scipy/interpolate/setup.py)

numba/tests/test_extending.py
```diff
@@ -1584,6 +1586,67 @@ def test_missing_function(self):
...
+        x2 = np.linspace(0,110,1000)
+        np.random.shuffle(x2)
+        isclose = np.allclose(interpolate.splev(x2, coeff_1), splev_wrapped(
+                              coeff_1[0],coeff_1[1],coeff_1[2],x2,0))
+        self.assertEqual(isclose,True)
+
+    def test_missing_module(self):
+        with self.assertRaises(ImportError) as raises:
+            get_f2py_function_address("fakemodule", "fakefunction")
+        # The quotes are not there in Python 2
```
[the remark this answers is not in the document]
max9111: I will remove this comment in `TestImportCythonFunction` and `TestImportF2PYFunction`.

numba/tests/test_extending.py
```diff
@@ -1584,6 +1586,67 @@ def test_missing_function(self):
...
+        msg = "No module named '?fakemodule'?"
+        match = re.match(msg, str(raises.exception))
+        self.assertIsNotNone(match)
```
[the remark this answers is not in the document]
max9111: I am also changing this in `TestImportCythonFunction`.

stuartarchibald:
Thanks for the updates @max9111. Looks like the build is failing in `flake8`:
```
numba/tests/test_extending.py:9:1: F401 're' imported but unused
```
once that is fixed and the build is passing I'll take another look. Thanks again!

numba/cpython/unsafe/stack_arr.py
```diff
@@ -0,0 +1,74 @@
...
+                 - Stack size is limited
+    """
+    arr_ptr = intr_stack_empty_alloc(shape, dtype)
+    if order == "C":
+        return carray(arr_ptr, shape)
+    elif order == "F":
+        return farray(arr_ptr, shape)
+    else:
+        raise errors.UnsupportedError(
+            "order must be one of 'C', 'F'")
```
stuartarchibald, line 74:
```suggestion
            "order must be one of 'C' or 'F'")
```

numba/tests/test_extending.py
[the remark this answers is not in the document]
stuartarchibald: Can this be queried programmatically?

stuartarchibald (commented):
Thanks for the updates to this, I've taken another look. There's a few comments that didn't get addressed/were lost from the initial review that I've tried to highlight. There's also a few questions/decisions to be made and some stylistic changes. Once these are resolved this should be good to merge. Thanks again for your help with this.

@sklam as this is adding to the public facing extension API please could you also take a look, particularly at the interface(s). Many thanks.

PR #4373, opened 2019-07-27
Set maximum name size to maximum allowable value

toddrme2178:
Fix for #3876 without needing to patch LLVM.

Issue #3876, opened 2019-03-18
LLVM verify error: Terminator found in the middle of a basic block


ehsantn:
## Reporting a bug

- [x] I am using the latest ~~released~~ master version of Numba (most recent is visible in
 the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [ ] I have included below a minimal working reproducer (if you are unsure how
 to write one see http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports).

I see the error below in a large piece of code that's hard to reproduce. I'm using the master of Numba and dev version of llvmlite. Any ideas where to look for issues?

```
  File "/Users/etotoni/pse-hpc/python/numba/numba/compiler.py", line 731, in stage_nopython_backend
    self._backend(lowerfn, objectmode=False)
  File "/Users/etotoni/pse-hpc/python/numba/numba/compiler.py", line 681, in _backend
    lowered = lowerfn()
  File "/Users/etotoni/pse-hpc/python/numba/numba/compiler.py", line 668, in backend_nopython_mode
    self.metadata)
  File "/Users/etotoni/pse-hpc/python/numba/numba/compiler.py", line 1074, in native_lowering_stage
    cfunc = targetctx.get_executable(library, fndesc, env)
  File "/Users/etotoni/pse-hpc/python/numba/numba/targets/cpu.py", line 166, in get_executable
    baseptr = library.get_pointer_to_function(fndesc.llvm_func_name)
  File "/Users/etotoni/pse-hpc/python/numba/numba/targets/codegen.py", line 469, in get_pointer_to_function
    self._ensure_finalized()
  File "/Users/etotoni/pse-hpc/python/numba/numba/targets/codegen.py", line 104, in _ensure_finalized
    self.finalize()
  File "/Users/etotoni/pse-hpc/python/numba/numba/targets/codegen.py", line 221, in finalize
    self._final_module.verify()
  File "/anaconda3/envs/PA37/lib/python3.7/site-packages/llvmlite/binding/module.py", line 116, in verify
    raise RuntimeError(str(outmsg))
RuntimeError: Failed in nopython mode pipeline (step: nopython mode backend)
Terminator found in the middle of a basic block!
[...]

PR #7828, opened 2022-02-09
Set maximum name size to maximum allowable value

esc:
Fix for #3876 without needing to patch LLVM.

Followup from: https://github.com/numba/numba/pull/4373


esc:
Closing this now as I don't think this is needed anymore for LLVM 14 -- can you double check @apmasell ?

PR #7829, opened 2022-02-10
CUDA: Support `Enum/IntEnum` in Kernel

isVoid:
This PR adds support to `Enum/IntEnum` to kernel. Tested cases include: compares, use as constants, explicit/implict convert to ints and vectorize.

### DIFF
--- a/numba/cuda/dispatcher.py
+++ b/numba/cuda/dispatcher.py
@@ -358,18 +358,6 @@ class _Kernel(serialize.ReduceMixin):
             kernelargs.append(ctypes.c_double(val.real))
             kernelargs.append(ctypes.c_double(val.imag))
 
-        elif isinstance(ty, types.EnumMember):
-            try:
-                self._prepare_args(
-                    ty.dtype, val.value, stream, retr, kernelargs
-                )
-            except Exception as e:
-                if isinstance(e, NotImplementedError):
-                    raise NotImplementedError(
-                        f"Unsupported value for Enum {val}"
-                    ) from e
-                raise
-
         elif isinstance(ty, (types.NPDatetime, types.NPTimedelta)):
             kernelargs.append(ctypes.c_int64(val.view(np.int64)))
 
@@ -384,6 +372,14 @@ class _Kernel(serialize.ReduceMixin):
             assert len(ty) == len(val)
             for t, v in zip(ty, val):
                 self._prepare_args(t, v, stream, retr, kernelargs)
+
+        elif isinstance(ty, types.EnumMember):
+            try:
+                self._prepare_args(
+                    ty.dtype, val.value, stream, retr, kernelargs
+                )
+            except NotImplementedError:
+                raise NotImplementedError(ty, val)
 
         else:
             raise NotImplementedError(ty, val)
--- a/numba/cuda/dispatcher.py
+++ b/numba/cuda/dispatcher.py
@@ -358,6 +358,18 @@ class _Kernel(serialize.ReduceMixin):
             kernelargs.append(ctypes.c_double(val.real))
             kernelargs.append(ctypes.c_double(val.imag))
 
+        elif isinstance(ty, types.EnumMember):
+            try:
+                self._prepare_args(
+                    ty.dtype, val.value, stream, retr, kernelargs
+                )
+            except Exception as e:
+                if isinstance(e, NotImplementedError):
+                    raise NotImplementedError(
+                        f"Unsupported value for Enum {val}"
+                    ) from e
+                raise
+
         elif isinstance(ty, (types.NPDatetime, types.NPTimedelta)):
             kernelargs.append(ctypes.c_int64(val.view(np.int64)))
 
--- a/numba/cuda/target.py
+++ b/numba/cuda/target.py
@@ -20,11 +20,13 @@ from numba.cuda import codegen, nvvmutils
 class CUDATypingContext(typing.BaseContext):
     def load_additional_registries(self):
         from . import cudadecl, cudamath, libdevicedecl
+        from numba.core.typing import enumdecl
 
         self.install_registry(cudadecl.registry)
         self.install_registry(cudamath.registry)
         self.install_registry(cmathdecl.registry)
         self.install_registry(libdevicedecl.registry)
+        self.install_registry(enumdecl.registry)
 
     def resolve_value_type(self, val):
         # treat dispatcher object as another device function
@@ -88,7 +90,7 @@ class CUDATargetContext(BaseContext):
     def load_additional_registries(self):
         # side effect of import needed for numba.cpython.*, the builtins
         # registry is updated at import time.
-        from numba.cpython import numbers, tupleobj, slicing # noqa: F401
+        from numba.cpython import numbers, tupleobj, slicing, enumimpl # noqa: F401, E501
         from numba.cpython import rangeobj, iterators # noqa: F401
         from numba.cpython import unicode, charseq # noqa: F401
         from numba.cpython import cmathimpl
--- a/numba/cuda/target.py
+++ b/numba/cuda/target.py
@@ -90,8 +90,8 @@ class CUDATargetContext(BaseContext):
     def load_additional_registries(self):
         # side effect of import needed for numba.cpython.*, the builtins
         # registry is updated at import time.
-        from numba.cpython import numbers, tupleobj, slicing, enumimpl # noqa: F401, E501
-        from numba.cpython import rangeobj, iterators # noqa: F401
+        from numba.cpython import numbers, tupleobj, slicing # noqa: F401
+        from numba.cpython import rangeobj, iterators, enumimpl # noqa: F401
         from numba.cpython import unicode, charseq # noqa: F401
         from numba.cpython import cmathimpl
         from numba.np import arrayobj # noqa: F401
--- a/numba/cuda/tests/cudapy/test_enums.py
+++ b/numba/cuda/tests/cudapy/test_enums.py
@@ -1,0 +1,116 @@
+"""
+Test cases adapted from numba/tests/test_enums.py
+"""
+
+import numpy as np
+
+from numba import int8, int16, int32
+from numba import cuda, vectorize
+from numba.tests.support import TestCase
+from numba.tests.enum_usecases import (
+    Color,
+    Shape,
+    Planet,
+    RequestError,
+    IntEnumWithNegatives
+)
+
+
+class EnumTest(TestCase):
+
+    pairs = [
+        (Color.red, Color.red),
+        (Color.red, Color.green),
+        (Planet.EARTH, Planet.EARTH),
+        (Planet.VENUS, Planet.MARS),
+        (Shape.circle, IntEnumWithNegatives.two) # IntEnum, same value
+    ]
+
+    def test_compare(self):
+        @cuda.jit
+        def f(a, b, out):
+            out[0] = a == b
+            out[1] = a != b
+            out[2] = a is b
+            out[3] = a is not b
+
+        arr = np.zeros((4,), dtype=np.bool_)
+        for a, b in self.pairs:
+            f[1, 1](a, b, arr)
+            self.assertPreciseEqual(arr, np.array([
+                a == b, a != b, a is b, a is not b
+            ]))
+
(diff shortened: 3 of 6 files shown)


numba/cuda/dispatcher.py
```diff
@@ -358,6 +358,18 @@ def _prepare_args(self, ty, val, stream, retr, kernelargs):
             kernelargs.append(ctypes.c_double(val.real))
             kernelargs.append(ctypes.c_double(val.imag))
 
+        elif isinstance(ty, types.EnumMember):
```
gmarkall, line 361: Can we move this beneath the `Record` and `BaseTuple` cases? `_prepare_args` is the slowest part of the launch and my concern is that each case adds a measurable impact on the launch time - if we move this check last it should avoid the impact on existing types.

numba/cuda/dispatcher.py
```diff
@@ -358,6 +358,18 @@ def _prepare_args(self, ty, val, stream, retr, kernelargs):
...
+            try:
+                self._prepare_args(
+                    ty.dtype, val.value, stream, retr, kernelargs
+                )
+            except Exception as e:
+                if isinstance(e, NotImplementedError):
+                    raise NotImplementedError(
+                        f"Unsupported value for Enum {val}"
+                    ) from e
+                raise
```
gmarkall, line 371: Could this be simplified to:

```suggestion
            except NotImplementedError as e:
                raise NotImplementedError(f"Unsupported value for Enum {val}")
```

(or is there a different effect I've missed here?)

numba/cuda/target.py
```diff
@@ -88,7 +90,7 @@ def init(self):
     def load_additional_registries(self):
         # side effect of import needed for numba.cpython.*, the builtins
         # registry is updated at import time.
-        from numba.cpython import numbers, tupleobj, slicing # noqa: F401
+        from numba.cpython import numbers, tupleobj, slicing, enumimpl # noqa: F401, E501
```
gmarkall, line 93: Just to keep the lines within the limit and avoid an extra `noqa`, something along the lines of:

```suggestion
        from numba.cpython import numbers, tupleobj, slicing # noqa: F401
        from numba.cpython import enumimpl # noqa: F401
```

(or it could be added to the following line with `rangeobj` and `iterators`)

numba/cuda/tests/cudapy/test_enums.py
```diff
@@ -0,0 +1,116 @@
...
+            out[0] = helper(pred) == Color.red
+            out[1] = helper(not pred) == Color.green
+
+        arr = np.zeros((2,), dtype=np.bool_)
+        f[1, 1](True, arr)
+        self.assertPreciseEqual(arr, np.array([
+            Color.red == Color.red, Color.green == Color.green
+        ]))
+
+    def test_int_coerse(self):
```
gmarkall, line 73: Possibly a typo from the original?

```suggestion
    def test_int_coerce(self):
```

gmarkall:
gpuci run tests

numba/cuda/tests/cudapy/test_enums.py
```diff
@@ -0,0 +1,116 @@
...
+            out[1] = Color['red'] == Color['green']
+
+        arr = np.zeros((2,), dtype=np.bool_)
+        f[1, 1](arr)
+        self.assertPreciseEqual(arr, np.array([
+            Color.red == Color.green, Color['red'] == Color['green']
+        ]))
+
+    def test_return_from_device_func(self):
+        @cuda.jit
```
gmarkall, line 58: This should probably be:

```suggestion
        @cuda.jit(device=True)
```

though if it works without it (except on the simulator) then that's interesting - I've been working for some time on fixing the CUDA target so that it's not necessary to specify which functions are device functions, and it could well be that this now works "by accident".

numba/cuda/tests/cudapy/test_enums.py
```diff
@@ -0,0 +1,116 @@
+"""
+Test cases adapted from numba/tests/test_enums.py
+"""
+
+import numpy as np
+
+from numba import int8, int16, int32
+from numba import cuda, vectorize
+from numba.tests.support import TestCase
```
gmarkall, line 9:
```suggestion
from numba.cuda.testing import unittest, CUDATestCase
```

numba/cuda/tests/cudapy/test_enums.py
```diff
@@ -0,0 +1,116 @@
...
+from numba.tests.enum_usecases import (
+    Color,
+    Shape,
+    Planet,
+    RequestError,
+    IntEnumWithNegatives
+)
+
+
+class EnumTest(TestCase):
```
gmarkall, line 19:
```suggestion
class EnumTest(CUDATestCase):
```

numba/cuda/tests/cudapy/test_enums.py
```diff
@@ -0,0 +1,116 @@
...
+            if x != RequestError.not_found:
+                return RequestError['internal_error']
+            else:
+                return RequestError.dummy
+
+        cuda_func = vectorize("int64(int64)", target='cuda')(f)
+        arr = np.array([2, 404, 500, 404])
+        expected = np.array([f(x) for x in arr])
+        got = cuda_func(arr)
+        self.assertPreciseEqual(expected, got)
```
gmarkall, line 116:
```suggestion
        self.assertPreciseEqual(expected, got)


if __name__ == '__main__':
    unittest.main()
```

numba/cuda/tests/cudapy/test_enums.py
gmarkall: This needs to be a `CUDATestCase` to ensure that the tests run in the right process - otherwise we get the "CUDA initialized before forking" error we see in GPU CI.

numba/cuda/tests/cudapy/test_enums.py
gmarkall: (ah, it wasn't a typo in the original :slightly_smiling_face:)

numba/cuda/tests/cudapy/test_enums.py
gmarkall: This pattern is common at least to all the CUDA tests, so that a file can be run with e.g. `python numba/cuda/tests/cudapy/test_enums.py` (though I usually use the test harness module with `python -m numba.runtests numba.cuda.tests.cudapy.test_enums`)

numba/cuda/dispatcher.py
isVoid: I was basically unfamiliar with the type of exception may be raised here. If only `NotImplementedError` is thrown then we can simplify this.

isVoid (commented):
I think I addressed all reviews. Some comments below. And `hash` was not supported for integer types on cuda kernel.

```python
In [2]: from numba import cuda
   ...: @cuda.jit
   ...: def f():
   ...:     a = hash(128)
   ...: 
   ...: f[1, 1]()

LoweringError: Failed in cuda mode pipeline (step: native lowering)
No definition for lowering <built-in method impl of _dynfunc._Closure object at 0x7f705fc3c100>(int64,) -> int64

File "<ipython-input-2-aaeba3127a8b>", line 4:
def f():
    a = hash(128)
```

I don't think it's trivial to support this just yet. From this [page](https://numba.pydata.org/numba-doc/dev/developer/hashing.html?highlight=hash) we may need to implement some functions on device?

gmarkall:
> I don't think it's trivial to support this just yet. From this [page](https://numba.pydata.org/numba-doc/dev/developer/hashing.html?highlight=hash) we may need to implement some functions on device?

Ah - once overloads and the high-level extension API works on the CUDA target, this should be easier to re-visit (if there are hashes written only using `overload_method` and `intrinsic`s then support for them should come "for free").

numba/cuda/tests/cudapy/test_enums.py
```diff
@@ -0,0 +1,126 @@
+"""
+Test cases adapted from numba/tests/test_enums.py
+"""
+
+import numpy as np
+
+from numba import int8, int16, int32
+from numba import cuda, vectorize
```
gmarkall, line 8:
```suggestion
from numba import cuda, njit, vectorize
```

numba/cuda/tests/cudapy/test_enums.py
```diff
@@ -0,0 +1,126 @@
...
+        cuda_f[1, 1](got)
+        f(expected)
+        self.assertPreciseEqual(expected, got)
+
+    def test_return_from_device_func(self):
+        def inner(pred):
+            return Color.red if pred else Color.green
+
+        def _make_test_func(inner, jitter):
+            inner = jitter(inner)
```
gmarkall, line 62:
```suggestion
            inner = njit(inner)
```

numba/cuda/tests/cudapy/test_enums.py
```diff
@@ -0,0 +1,126 @@
...
+    ]
+
+    def test_compare(self):
+        def f(a, b, out):
+            out[0] = a == b
+            out[1] = a != b
+            out[2] = a is b
+            out[3] = a is not b
+
+        cuda_f = cuda.jit()(f)
```
gmarkall, line 36:
```suggestion
        cuda_f = cuda.jit(f)
```

numba/cuda/tests/cudapy/test_enums.py
```diff
@@ -0,0 +1,126 @@
...
+            f(a, b, expected)
+            self.assertPreciseEqual(expected, got)
+
+    def test_getattr_getitem(self):
+        def f(out):
+            # Lookup of an enum member on its class
+            out[0] = Color.red == Color.green
+            out[1] = Color['red'] == Color['green']
+
+        cuda_f = cuda.jit()(f)
```
gmarkall, line 50:
```suggestion
        cuda_f = cuda.jit(f)
```

numba/cuda/tests/cudapy/test_enums.py
```diff
@@ -0,0 +1,126 @@
...
+        def _make_test_func(inner, jitter):
+            inner = jitter(inner)
+
+            def f(pred, out):
+                out[0] = inner(pred) == Color.red
+                out[1] = inner(not pred) == Color.green
+            return jitter(f)
+
+        f = _make_test_func(inner, lambda x:x)
+        cuda_f = _make_test_func(inner, cuda.jit())
```
gmarkall, line 70:
```suggestion
        cuda_f = _make_test_func(inner, cuda.jit)
```

numba/cuda/tests/cudapy/test_enums.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def test_int_coerce(self):
+        def f(x, out):
+            # Implicit coercion of intenums to ints
+            if x > RequestError.internal_error:
+                out[0] = x - RequestError.not_found
+            else:
+                out[0] = x + Shape.circle
+
+        cuda_f = cuda.jit()(f)
```
gmarkall, line 86:
```suggestion
        cuda_f = cuda.jit(f)
```

numba/cuda/tests/cudapy/test_enums.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def test_int_cast(self):
+        def f(x, out):
+            # Explicit coercion of intenums to ints
+            if x > int16(RequestError.internal_error):
+                out[0] = x - int32(RequestError.not_found)
+            else:
+                out[0] = x + int8(Shape.circle)
+
+        cuda_f = cuda.jit()(f)
```
gmarkall, line 102:
```suggestion
        cuda_f = cuda.jit(f)
```

gmarkall (changes requested):
Many thanks for the changes - all looks good with the exception of:

- For `test_device_func`, if you use `njit` to wrap the inner function, this avoids it expecting a launch configuration when run with the simulator - this works around a limitation in the simulator.
- It's not necessary to call `cuda.jit()` with no arguments to wrap a function - `cuda.jit` is sufficient.

I believe that applying all the suggestions on the diff should resolve these two issues - if you're happy with them, then once applied we can do another gpuci run.

gmarkall:
gpu ci run tests

gmarkall:
gpuci run tests

gmarkall (approved):
Many thanks for the recent fixes - this looks good!

PR #7833, opened 2022-02-11
Add version support information table to docs.

stuartarchibald:
This adds a table to the docs that shows which versions of Numba's
dependencies were tested and known to work against which version of
Numba. It also updates the `first_rc_checklist.md` issue template
to ensure that this information is provided along side future
releases.


docs/source/reference/support_info.rst
```diff
@@ -0,0 +1,37 @@
+.. _support_info:
```
sklam, line 1: I think this page should be referenced in the installation doc: https://numba--7833.org.readthedocs.build/en/7833/user/installing.html

esc: I would actual raise the argument, that the version support information table, should go alongside our canonical reference for the other dependencies, here:

https://numba.readthedocs.io/en/stable/user/installing.html#dependency-list

This would mean, all dependency information is in a single place and can percolate into the various places (`setup.py` et.c..) from there. How does that sound?

stuartarchibald: 867223f makes this move, I'm happy either way, am just glad this reference now exists!

sklam (approved):
LGTM now

esc (approved):
thank you!

PR #5390, opened 2020-03-16
add error handling for lookup_module

ssikdar1:
# Description
Closes #4914 . Adds error handling for when `self.modname` cannot be found in `sys.modules`. 

# Testing 

Added a unit test `test_funcdesc.py` with the example case from the issue. 
```
$ python -m numba.runtests  numba.tests.test_funcdesc
.
----------------------------------------------------------------------
Ran 1 test in 0.224s
```

Also by testing in the shell:
```
>>> filename = 'test.py'
>>> name = 'mypackage'
>>> code = """
... def f(x):
...     return x
... """
>>> 
>>> objs = dict(__file__=filename, __name__=name)
>>> compiled = compile(code, filename, 'exec')
>>> exec(compiled, objs)
>>> 
>>> from numba import njit
>>> compiled_f = njit(objs['f'])
>>> print(compiled_f(0))
Traceback (most recent call last):
  File "/Users/ssikdar1/numba/numba/core/lowering.py", line 30, in from_fndesc
    return cls._memo[fndesc.env_name]
  File "/Users/ssikdar1/anaconda3/envs/numbaenv/lib/python3.6/weakref.py", line 137, in __getitem__
    o = self.data[key]()
KeyError: '_ZN08NumbaEnv9mypackage5f$241Ex'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/ssikdar1/numba/numba/core/funcdesc.py", line 94, in lookup_module
    return sys.modules[self.modname]
KeyError: 'mypackage'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/ssikdar1/numba/numba/core/dispatcher.py", line 420, in _compile_for_args
    raise e
...
  File "/Users/ssikdar1/numba/numba/core/funcdesc.py", line 96, in lookup_module
    raise ModuleNotFoundError(f"can't compile {self.qualname}: import of module {self.modname} failed")
ModuleNotFoundError: can't compile f: import of module mypackage failed
>>> 
```

PR #7835, opened 2022-02-11
Fix pickling error when module cannot be imported

idorrington:
The [change to using cloudpickle ](https://github.com/numba/numba/pull/6977) also changed the behaviour when unpickling a function from a module that cannot be imported. Previously Numba used the _rebuild_function in serialize.py, which handled the ImportErrors like this

```
def _rebuild_function(code_reduced, globals, name, cell_values, defaults):
    """
    Rebuild a function from its _reduce_function() results.
    """
    if cell_values:
        cells = tuple(_make_cell(v) for v in cell_values)
    else:
        cells = ()
    code = _rebuild_code(*code_reduced)
    modname = globals['__name__']
    try:
        _rebuild_module(modname)
    except ImportError:
        # If the module can't be found, avoid passing it (it would produce
        # errors when lowering).
        del globals['__name__']
    return FunctionType(code, globals, name, defaults, cells)
```

This PR attempts to reimplement this behaviour within cloudpickle. 

Below is a minimal example to show the issue. The real life context for wanting this feature is deploying the code using dask distributed, where the pickled function is sent to worker nodes but the original module is not. 

```
import pickle
import sys
import os


source = """
from numba import vectorize

@vectorize(["float64(float64)"])
def inc1(x):
    return x + 1
"""

modname = "example"
filename = f"{modname}.py"
f = open(filename, "a")
f.write(source)
f.close()

from example import inc1

os.remove(filename)
del sys.modules[modname]
pkl = pickle.dumps(inc1)

f = pickle.loads(pkl)
print(f(2))
```

I'm not very familiar with the Numba source code, so please let me know if you think there is a better way to implement this feature. If it seems ok, I'm happy to make a unit test for it.

### DIFF
--- a/numba/cloudpickle/cloudpickle.py
+++ b/numba/cloudpickle/cloudpickle.py
@@ -220,7 +220,7 @@ def _lookup_module_and_qualname(obj, name=None):
         # types.ModuleType. The other possibility is that module was removed
         # from sys.modules after obj was created/imported. But this case is not
         # supported, as the standard pickle does not support it either.
-        del obj.__module__
+        obj.__module__ = None
         return None
 
     try:
--- a/numba/cloudpickle/cloudpickle.py
+++ b/numba/cloudpickle/cloudpickle.py
@@ -203,6 +203,10 @@ def _lookup_module_and_qualname(obj, name=None):
 
     module_name = _whichmodule(obj, name)
 
+    if module_name not in sys.modules:
+        del obj.__module__
+        return None
+
     if module_name is None:
         # In this case, obj.__module__ is None AND obj was not found in any
         # imported module. obj is thus treated as dynamic.
--- a/numba/cloudpickle/cloudpickle.py
+++ b/numba/cloudpickle/cloudpickle.py
@@ -220,7 +220,6 @@ def _lookup_module_and_qualname(obj, name=None):
         # types.ModuleType. The other possibility is that module was removed
         # from sys.modules after obj was created/imported. But this case is not
         # supported, as the standard pickle does not support it either.
-        obj.__module__ = None
         return None
 
     try:
--- a/numba/cloudpickle/cloudpickle.py
+++ b/numba/cloudpickle/cloudpickle.py
@@ -203,10 +203,6 @@ def _lookup_module_and_qualname(obj, name=None):
 
     module_name = _whichmodule(obj, name)
 
-    if module_name not in sys.modules:
-        del obj.__module__
-        return None
-
     if module_name is None:
         # In this case, obj.__module__ is None AND obj was not found in any
         # imported module. obj is thus treated as dynamic.
@@ -224,6 +220,7 @@ def _lookup_module_and_qualname(obj, name=None):
         # types.ModuleType. The other possibility is that module was removed
         # from sys.modules after obj was created/imported. But this case is not
         # supported, as the standard pickle does not support it either.
+        del obj.__module__
         return None
 
     try:
--- a/numba/core/funcdesc.py
+++ b/numba/core/funcdesc.py
@@ -87,7 +87,7 @@ class FunctionDescriptor(object):
         """
         Return the module in which this function is supposed to exist.
         This may be a dummy module if the function was dynamically
-        generated. Raise exception if the module can't be found.
+        generated or the module can't be found.
         """
         if self.modname == _dynamic_modname:
             return _dynamic_module
--- a/numba/core/funcdesc.py
+++ b/numba/core/funcdesc.py
@@ -96,9 +96,7 @@ class FunctionDescriptor(object):
                 # ensure module exist
                 return importlib.import_module(self.modname)
             except ImportError:
-                raise ModuleNotFoundError(
-                    f"can't compile {self.qualname}: "
-                    f"import of module {self.modname} failed") from None
+                return _dynamic_module
 
     def lookup_function(self):
         """
--- a/numba/tests/test_funcdesc.py
+++ b/numba/tests/test_funcdesc.py
@@ -25,7 +25,7 @@ def f(x):
         exec(compiled, objs)
 
         compiled_f = njit(objs['f'])
-        assert compliled_f(3) == 3
+        assert compiled_f(3) == 3
 
 
 class TestFuncDescMangledName(unittest.TestCase):
--- a/numba/tests/test_funcdesc.py
+++ b/numba/tests/test_funcdesc.py
@@ -25,10 +25,7 @@ def f(x):
         exec(compiled, objs)
 
         compiled_f = njit(objs['f'])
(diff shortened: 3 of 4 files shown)


numba/cloudpickle/cloudpickle.py
```diff
@@ -203,6 +203,10 @@ def _lookup_module_and_qualname(obj, name=None):
 
     module_name = _whichmodule(obj, name)
 
+    if module_name not in sys.modules:
```
sklam, line 206: Why isn't this case cover by the check at line 221?

https://github.com/numba/numba/blob/abffa7060eedff66cfa3e7b3876018c1f22e5e0e/numba/cloudpickle/cloudpickle.py#L220-L227

numba/cloudpickle/cloudpickle.py
```diff
@@ -203,6 +203,10 @@ def _lookup_module_and_qualname(obj, name=None):
...
+        del obj.__module__
```
sklam, line 207: Why is it necessary to do this?

(To my surprise, I didn't know that deleting `__module__` is the same as setting it to `None`.)

numba/cloudpickle/cloudpickle.py
idorrington: You're right. I've moved the delete obj.__module__ into this block now

numba/cloudpickle/cloudpickle.py
idorrington: If we don't delete the module from the object then it will try to import it the module at line 97 in fundesc.py. It will then raise the module import error and fail. We could maybe deal with it at that point, but I'm not very familiar with the code and so tried to make it as similar as possible to the old serialisation code.

idorrington:
Hi, is there any update on this?

numba/cloudpickle/cloudpickle.py
```diff
@@ -220,6 +220,7 @@ def _lookup_module_and_qualname(obj, name=None):
         # types.ModuleType. The other possibility is that module was removed
         # from sys.modules after obj was created/imported. But this case is not
         # supported, as the standard pickle does not support it either.
+        del obj.__module__
```
sklam, line 223:
```suggestion
        obj.__module__ = None
```

Some object type forbids the deletion of `__module__`. For example, any class type will reject this.

On the other hand, it is odd for this function to mutate the object.

sklam (commented):
I am wondering if the better fix is instead to modify https://github.com/numba/numba/blob/2e825b000e609e090005f8fb3d567999f1f1940c/numba/core/funcdesc.py#L100 so it returns `_dynamic_module`. We can probably do that if the error you are seeing is:

```pytb
  File "/path/to/numba/core/funcdesc.py", line 99, in lookup_module
    raise ModuleNotFoundError(
ModuleNotFoundError: can't compile inc1: import of module example failed
```

sklam:
Related PR: https://github.com/numba/numba/pull/5390

idorrington:
> I am wondering if the better fix is instead to modify
> 
> https://github.com/numba/numba/blob/2e825b000e609e090005f8fb3d567999f1f1940c/numba/core/funcdesc.py#L100
> 
> so it returns `_dynamic_module`. We can probably do that if the error you are seeing is:
> ```
>   File "/path/to/numba/core/funcdesc.py", line 99, in lookup_module
>     raise ModuleNotFoundError(
> ModuleNotFoundError: can't compile inc1: import of module example failed
> ```

I tested this with our codebase using dask distributed and it also works. I don't mind which solution we use.

idorrington:
This is still marked "waiting on author". Do you want me to change the ModuleNotFoundError to return a dynamic module instead? I tested and it also works for our code. It really just depends on what your preferences are as you have to maintain this code.

idorrington:
@sklam Just pinging again with this. Do you have a preference for using the dynamic module? If so I can make that change. It worked for us. 

Also, do you know when the next release will be and if we could get this into it? This is preventing us from updating Numba which means we also cannot update to Python 3.10

idorrington:
@sklam I've change this to use _dynamic_module as you suggested

sklam:
@idorrington, can you push your test for this patch?

numba/tests/test_funcdesc.py
```diff
@@ -25,10 +25,7 @@ def f(x):
         exec(compiled, objs)
 
         compiled_f = njit(objs['f'])
-        with self.assertRaises(ModuleNotFoundError) as raises:
-            compiled_f(0)
-        msg = "can't compile f: import of module mypackage failed"
-        self.assertIn(msg, str(raises.exception))
+        assert compiled_f(3) == 3
```
sklam, line 28:
```suggestion
        self.assertEqual(compiled_f(3), 3)
```

numba/core/funcdesc.py
```diff
@@ -96,9 +96,7 @@ def lookup_module(self):
                 # ensure module exist
                 return importlib.import_module(self.modname)
             except ImportError:
-                raise ModuleNotFoundError(
-                    f"can't compile {self.qualname}: "
-                    f"import of module {self.modname} failed") from None
+                return _dynamic_module
```
sklam, line 99: With this change, the doc string needs to be updated to remove `Raise exception if the module can't be found.` and update it with the new behavior.

numba/tests/test_unpickle_without_module.py
```diff
@@ -0,0 +1,49 @@
...
+            # Import the temporary module before file is deleted
+            from tmp_module import inc1
+
+        # Remove from imported libraries
+        del sys.modules[modname]
+
+        # Pickle function and assert that it loads correctly
+        pkl = pickle.dumps(inc1)
+        f = pickle.loads(pkl)
+        assert f(2) == 3
```
sklam, line 49:
```suggestion
        self.assertEqual(f(2), 3)
```

numba/tests/test_funcdesc.py
```diff
@@ -25,10 +25,7 @@ def f(x):
         exec(compiled, objs)
 
         compiled_f = njit(objs['f'])
-        with self.assertRaises(ModuleNotFoundError) as raises:
-            compiled_f(0)
-        msg = "can't compile f: import of module mypackage failed"
-        self.assertIn(msg, str(raises.exception))
```
sklam, line 31: Ah yes, the change is necessary because funcdesc no longer raise if the module is not found.

sklam (commented):
That changes to the test makes sense.

The remaining issues are just a docstring needs updating because `ModuleNotFoundError` is no longer raised, and we use unittest style to the test should not use bare `assert`.

numba/core/funcdesc.py
idorrington: Done :)

sklam (approved):
Thanks for the patch!

Issue #4134, opened 2019-06-01
ParallelAccelerator error for array of dt64 min()


ehsantn:
## Reporting a bug

- [x] I am using the latest ~~released~~ master version of Numba (most recent is visible in
 the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [x] I have included below a minimal working reproducer (if you are unsure how
 to write one see http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports).

Example below, which calls `min()` on an array of datetime64s, fails in ParallelAccelerator. This particular error seems related to `get_type_max_value`, but in general, datetime64/timedelta64 support is not supported/tested properly.
```python
@numba.njit(parallel=True) 
def f(A): 
     return A.min() 
f(np.array([np.datetime64(0, 'ns')]))
```
```
~/dev/numba/numba/errors.py in new_error_context(fmt_, *args, **kwargs)
    660     try:
--> 661         yield
    662     except NumbaError as e:

~/dev/numba/numba/lowering.py in lower_block(self, block)
    257                                    loc=self.loc, errcls_=defaulterrcls):
--> 258                 self.lower_inst(inst)
    259 

~/dev/numba/numba/lowering.py in lower_inst(self, inst)
    409                 if isinstance(inst, _class):
--> 410                     func(self, inst)
    411                     return

~/dev/numba/numba/npyufunc/parfor.py in _lower_parfor_parallel(lowerer, parfor)
     74             print("lower init_block instr = ", instr)
---> 75         lowerer.lower_inst(instr)
     76 

~/dev/numba/numba/lowering.py in lower_inst(self, inst)
    300             ty = self.typeof(inst.target.name)
--> 301             val = self.lower_assign(ty, inst)
    302             self.storevar(val, inst.target.name)

~/dev/numba/numba/lowering.py in lower_assign(self, ty, inst)
    458         elif isinstance(value, ir.Expr):
--> 459             return self.lower_expr(ty, value)
    460 

~/dev/numba/numba/lowering.py in lower_expr(self, resty, expr)
    917         elif expr.op == 'call':
--> 918             res = self.lower_call(resty, expr)
    919             return res

~/dev/numba/numba/lowering.py in lower_call(self, resty, expr)
    710         else:
--> 711             res = self._lower_call_normal(fnty, expr, signature)
    712 

~/dev/numba/numba/lowering.py in _lower_call_normal(self, fnty, expr, signature)
    889 
[...]

PR #7836, opened 2022-02-11
min() and max() support for np.datetime and np.timedelta

benwilliamgraham:
Fix for https://github.com/numba/numba/issues/4134

### DIFF
--- a/numba/core/typing/builtins.py
+++ b/numba/core/typing/builtins.py
@@ -1105,7 +1105,6 @@ class MinValInfer(AbstractTemplate):
         assert len(args) == 1
         if isinstance(args[0], (types.DType, types.NumberClass)):
             return signature(args[0].dtype, *args)
-        assert None
 
 
 #------------------------------------------------------------------------------
--- a/numba/core/typing/builtins.py
+++ b/numba/core/typing/builtins.py
@@ -881,7 +881,7 @@ class MinMaxBase(AbstractTemplate):
 
     def _unify_minmax(self, tys):
         for ty in tys:
-            if not isinstance(ty, types.Number):
+            if not isinstance(ty, (types.Number, types.NPDatetime, types.NPTimedelta)):
                 return
         return self.context.unify_types(*tys)
 
@@ -1103,8 +1103,11 @@ class MinValInfer(AbstractTemplate):
     def generic(self, args, kws):
         assert not kws
         assert len(args) == 1
-        assert isinstance(args[0], (types.DType, types.NumberClass))
-        return signature(args[0].dtype, *args)
+        if isinstance(args[0], (types.DType, types.NumberClass)):
+            return signature(args[0].dtype, *args)
+        elif isinstance(args[0], (types.NPDatetime, types.NPTimedelta)):
+            return signature(args[0], *args)
+        assert False
 
 
 #------------------------------------------------------------------------------
--- a/numba/core/typing/builtins.py
+++ b/numba/core/typing/builtins.py
@@ -1105,8 +1105,6 @@ class MinValInfer(AbstractTemplate):
         assert len(args) == 1
         if isinstance(args[0], (types.DType, types.NumberClass)):
             return signature(args[0].dtype, *args)
-        elif isinstance(args[0], (types.NPDatetime, types.NPTimedelta)):
-            return signature(args[0], *args)
         assert False
 
 
--- a/numba/core/typing/builtins.py
+++ b/numba/core/typing/builtins.py
@@ -1105,7 +1105,7 @@ class MinValInfer(AbstractTemplate):
         assert len(args) == 1
         if isinstance(args[0], (types.DType, types.NumberClass)):
             return signature(args[0].dtype, *args)
-        assert False
+        assert None
 
 
 #------------------------------------------------------------------------------
--- a/numba/core/typing/npdatetime.py
+++ b/numba/core/typing/npdatetime.py
@@ -266,3 +266,14 @@ class DatetimeCmpGt(DatetimeCmpOp):
 @infer_global(operator.ge)
 class DatetimeCmpGE(DatetimeCmpOp):
     key = operator.ge
+
+
+@infer_global(npdatetime_helpers.datetime_minimum)
+@infer_global(npdatetime_helpers.datetime_maximum)
+class DatetimeMinMax(AbstractTemplate):
+    def generic(self, args, kws):
+        assert not kws
+        assert len(args) == 2
+        assert isinstance(args[0], (types.NPDatetime, types.NPTimedelta))
+        assert isinstance(args[1], (types.NPDatetime, types.NPTimedelta))
+        return signature(args[0], *args)
--- a/numba/core/typing/npdatetime.py
+++ b/numba/core/typing/npdatetime.py
@@ -277,7 +277,9 @@ class DatetimeMinMax(AbstractTemplate):
         error_msg = "DatetimeMinMax requires both arguments to be NPDatetime type or both arguments to be NPTimedelta types"
         assert isinstance(args[0], (types.NPDatetime, types.NPTimedelta)), error_msg
         if isinstance(args[0], types.NPDatetime):
-            assert isinstance(args[1], types.NPDatetime), error_msg
+            if not isinstance(args[1], types.NPDatetime):
+                raise TypeError(error_msg)
         else:
-            assert isinstance(args[1], types.NPTimedelta), error_msg
+            if not isinstance(args[1], types.NPTimedelta):
+                raise TypeError(error_msg)
         return signature(args[0], *args)
--- a/numba/core/typing/npdatetime.py
+++ b/numba/core/typing/npdatetime.py
@@ -274,6 +274,10 @@ class DatetimeMinMax(AbstractTemplate):
     def generic(self, args, kws):
         assert not kws
         assert len(args) == 2
-        assert isinstance(args[0], (types.NPDatetime, types.NPTimedelta))
-        assert isinstance(args[1], (types.NPDatetime, types.NPTimedelta))
+        error_msg = "DatetimeMinMax requires both arguments to be NPDatetime type or both arguments to be NPTimedelta types"
+        assert isinstance(args[0], (types.NPDatetime, types.NPTimedelta)), error_msg
+        if isinstance(args[0], types.NPDatetime):
+            assert isinstance(args[1], types.NPDatetime), error_msg
+        else:
+            assert isinstance(args[1], types.NPTimedelta), error_msg
         return signature(args[0], *args)
--- a/numba/cpython/builtins.py
+++ b/numba/cpython/builtins.py
@@ -481,13 +481,14 @@ def get_type_min_value(typ):
 @lower_builtin(get_type_min_value, types.DType)
 def lower_get_type_min_value(context, builder, sig, args):
     typ = sig.args[0].dtype
-    bw = typ.bitwidth
 
     if isinstance(typ, types.Integer):
+        bw = typ.bitwidth
         lty = ir.IntType(bw)
         val = typ.minval
         res = ir.Constant(lty, val)
     elif isinstance(typ, types.Float):
+        bw = typ.bitwidth
         if bw == 32:
             lty = ir.FloatType()
         elif bw == 64:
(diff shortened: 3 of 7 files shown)


numba/cpython/builtins.py
```diff
@@ -496,19 +497,25 @@ def lower_get_type_min_value(context, builder, sig, args):
             raise NotImplementedError("llvmlite only supports 32 and 64 bit floats")
         npty = getattr(np, 'float{}'.format(bw))
         res = ir.Constant(lty, -np.inf)
+    elif isinstance(typ, (types.NPDatetime, types.NPTimedelta)):
+        bw = 64
+        lty = ir.IntType(bw)
+        val = types.uint64.minval
```
njriasan, line 503: I think the minimum value for a timedelta/datetime is defined to be types.int64.minval. Is this a typo? Is this tested?

numba/tests/test_npdatetime.py
```diff
@@ -544,6 +550,21 @@ def check(a):
...
             check(typ(*case))
 
+    def test_min_max(self):
+        min_f = self.jit(min_usecase)
+        max_f = self.jit(max_usecase)
+        def check(a, b):
+            self.assertPreciseEqual(min_f(a, b), min(a, b))
+            self.assertPreciseEqual(max_f(a, b), max(a, b))
+
+        check(TD(1), TD(2))
```
njriasan, line 560: I think you could simplify this could by have an iterable of arguments and call check on the pairs.

For example:

```python
for x,y in pairs
    check(x, y)
```

numba/tests/test_npdatetime.py
```diff
@@ -772,6 +793,21 @@ def check_lt(a, b, expected):
...
                 check_lt(b - np.timedelta64(1, unit), a, True)
 
+    def test_min_max(self):
+        min_f = self.jit(min_usecase)
+        max_f = self.jit(max_usecase)
+        def check(a, b):
+            self.assertPreciseEqual(min_f(a, b), min(a, b))
+            self.assertPreciseEqual(max_f(a, b), max(a, b))
+
+        check(DT(1, 'ns'), DT(2, 'ns'))
```
njriasan, line 803: Same comment as above.

numba/tests/test_npdatetime.py
```diff
@@ -903,6 +939,81 @@ def test_add_td_no_match(self):
...
+            ]),
+            np.array([
+                TD(1, "D"),
+                TD(2, "D"),
+                TD(3, "D"),
+                TD(4, "D"),
+            ]),
+        ]
+
+        for arr in test_cases:
```
njriasan, line 984: Can you add a test with NaT in each array type? In addition can you add a test for the edge case where the data is all NaT.

njriasan (commented):
This looks pretty good Ben. I left some comments with some changes.

In addition could you change the title of this PR. As it is right now I think the name is unclear, but I think a name like `min() and max() support for np.datetime and np.timedelta` makes it much easier to understand what this PR contains.

numba/parfors/parfor.py
```diff
@@ -128,13 +128,35 @@ def min_parallel_impl(return_type, arg):
...
+                        val = min(val, in_arr[i])
+                    return val
+            else:
+                def min_1(in_arr):
+                    numba.parfors.parfor.init_prange()
+                    min_checker(len(in_arr))
+                    val = numba.cpython.builtins.get_type_max_value(in_arr.dtype)
+                    for i in numba.parfors.parfor.internal_prange(len(in_arr)):
+                        if np.isnat(in_arr[i]):
+                            continue
```
njriasan, line 149: This doesn't seem like it will work correctly if all the values are NaT. Have you tested this with Numpy version 1.17?

numba/parfors/parfor.py
```diff
@@ -128,13 +128,35 @@ def min_parallel_impl(return_type, arg):
...
-            return val
+        if isinstance(arg.dtype, (types.NPDatetime, types.NPTimedelta)):
+            if numpy_version >= (1, 18):
+                def min_1(in_arr):
+                    numba.parfors.parfor.init_prange()
+                    min_checker(len(in_arr))
+                    val = numba.cpython.builtins.get_type_max_value(in_arr.dtype)
+                    for i in numba.parfors.parfor.internal_prange(len(in_arr)):
+                        if np.isnat(in_arr[i]):
+                            return in_arr[i]
```
njriasan, line 139: I don't think its safe to return from within a `prange`. This should be equivalent to putting a `break` within a parallel for loop, which is generally unsafe.

Instead I think you should maintain the value of `val` in the output based on `isnat`. This would require increasing control flow, but to minimize control flow you could do:
```python
if np.isnat(in_arr[i]):
    val = in_arr[i]
if not np.isnat(val):
    val = min(val, in_arr[i])
```

numba/parfors/parfor.py
```diff
@@ -150,13 +172,35 @@ def max_parallel_impl(return_type, arg):
...
```
njriasan, line 183: See comment above.

numba/parfors/parfor.py
```diff
@@ -150,13 +172,35 @@ def max_parallel_impl(return_type, arg):
...
+                        val = max(val, in_arr[i])
+                    return val
+            else:
+                def min_1(in_arr):
+                    numba.parfors.parfor.init_prange()
+                    min_checker(len(in_arr))
+                    val = numba.cpython.builtins.get_type_max_value(in_arr.dtype)
+                    for i in numba.parfors.parfor.internal_prange(len(in_arr)):
+                        if np.isnat(in_arr[i]):
+                            continue
```
njriasan, line 193: See comment above.

numba/parfors/parfor.py
```diff
@@ -128,13 +128,35 @@ def min_parallel_impl(return_type, arg):
...
     elif arg.ndim == 1:
-        def min_1(in_arr):
-            numba.parfors.parfor.init_prange()
-            min_checker(len(in_arr))
-            val = numba.cpython.builtins.get_type_max_value(in_arr.dtype)
-            for i in numba.parfors.parfor.internal_prange(len(in_arr)):
-                val = min(val, in_arr[i])
-            return val
+        if isinstance(arg.dtype, (types.NPDatetime, types.NPTimedelta)):
+            if numpy_version >= (1, 18):
```
njriasan, line 132: I think it would be helpful if you could add a comment explain how the behavior differs in Numpy versions. For example: `Starting with Numpy 1.18 NaT values are skipped if encountered rather than returned.`

numba/parfors/parfor.py
```diff
@@ -150,13 +172,35 @@ def max_parallel_impl(return_type, arg):
...
     elif arg.ndim == 1:
-        def max_1(in_arr):
-            numba.parfors.parfor.init_prange()
-            max_checker(len(in_arr))
-            val = numba.cpython.builtins.get_type_min_value(in_arr.dtype)
-            for i in numba.parfors.parfor.internal_prange(len(in_arr)):
-                val = max(val, in_arr[i])
-            return val
+        if isinstance(arg.dtype, (types.NPDatetime, types.NPTimedelta)):
+            if numpy_version >= (1, 18):
```
njriasan, line 176: See comment above.

numba/tests/test_npdatetime.py
```diff
@@ -903,6 +944,153 @@ def test_add_td_no_match(self):
...
+        self._test_min_max(np.min, False, True)
+
+    @skip_parfors_unsupported
+    def test_min_method_parallel(self):
+        self._test_min_max(np.min, True, True)
+
+    def test_max_method(self):
+        self._test_min_max(np.max, False, True)
+
+    @skip_parfors_unsupported
```
njriasan, line 1090: I think it might be helpful if you group the parallel tests together. Switching between parallel and regular makes it hard to keep track of which tests are needed.

njriasan (commented):
@benwilliamgraham I left some feedback on what I think are possible issues. Good work so far!

numba/parfors/parfor.py
```diff
@@ -121,20 +124,50 @@ class internal_prange(object):
     def __new__(cls, *args):
         return range(*args)
 
+def _datetime_minimum(a, b):
```
njriasan, line 127: I would move these implementations out of this file. I think it would be better if you placed it in `npdatetime_helpers.py`

njriasan (commented):
Overall I think this looks pretty good. Have you determined the source of the CI failure?

numba/core/typing/npdatetime.py
```diff
@@ -266,3 +266,14 @@ class DatetimeCmpGt(DatetimeCmpOp):
...
+
+
+@infer_global(npdatetime_helpers.datetime_minimum)
+@infer_global(npdatetime_helpers.datetime_maximum)
+class DatetimeMinMax(AbstractTemplate):
+    def generic(self, args, kws):
+        assert not kws
+        assert len(args) == 2
+        assert isinstance(args[0], (types.NPDatetime, types.NPTimedelta))
+        assert isinstance(args[1], (types.NPDatetime, types.NPTimedelta))
```
njriasan, line 278: This isn't fully accurate because it allows comparison between types.NPDatetime and types.NPTimedelta. I would refactor it as such.

```suggestion
        error_msg = "DatetimeMinMax requires both arguments to be NPDatetime type or both arguments to be NPTimedelta types"
        assert isinstance(args[0], (types.NPDatetime, types.NPTimedelta)), error_msg
        if isinstance(args[0], types.NPDatetime):
            assert isinstance(args[1], types.NPDatetime), error_msg
        else:
            assert isinstance(args[1], types.NPTimedelta), error_msg
```

numba/np/npdatetime.py
```diff
@@ -847,3 +847,7 @@ def impl(x):
                 return np.int64(-2)
             return np.int64(x)
     return impl
+
+
+lower_builtin(npdatetime_helpers.datetime_minimum, types.VarArg(types.Any))(datetime_minimum_impl)
+lower_builtin(npdatetime_helpers.datetime_maximum, types.VarArg(types.Any))(datetime_maximum_impl)
```
njriasan, line 853: I would be more specific on the input types to avoid future issues.

```suggestion
lower_builtin(npdatetime_helpers.datetime_minimum, types.NPDatetime, types.NPDatetime)(datetime_minimum_impl)
lower_builtin(npdatetime_helpers.datetime_minimum, types.NPTimedelta, types.NPTimedelta)(datetime_minimum_impl)
lower_builtin(npdatetime_helpers.datetime_minimum, types.NPDatetime, types.NPDatetime)(datetime_maximum_impl)
lower_builtin(npdatetime_helpers.datetime_minimum, types.NPTimedelta, types.NPTimedelta)(datetime_maximum_impl)
```

njriasan (commented):
I left a couple of comments but it looks very close. Thank you Ben!

njriasan (dismissed):
LGMT @benwilliamgraham. Thank you!

benwilliamgraham:
@stuartarchibald could you have the label switched from "in progress" to "ready for review" please?

numba/core/typing/builtins.py
```diff
@@ -1103,8 +1103,11 @@ class MinValInfer(AbstractTemplate):
     def generic(self, args, kws):
         assert not kws
         assert len(args) == 1
-        assert isinstance(args[0], (types.DType, types.NumberClass))
-        return signature(args[0].dtype, *args)
+        if isinstance(args[0], (types.DType, types.NumberClass)):
+            return signature(args[0].dtype, *args)
+        elif isinstance(args[0], (types.NPDatetime, types.NPTimedelta)):
+            return signature(args[0], *args)
```
sklam, line 1109: This doesn't seems to be needed. the datetime types seems to always be wrapped in a `DType`; e.g. `dtype(timedelta64[D])`.

numba/tests/test_npdatetime.py
```diff
@@ -544,6 +550,24 @@ def check(a):
...
             check(typ(*case))
 
+    def _test_min_max(self, usecase):
+        f = self.jit(usecase)
+        def check(a, b):
+            self.assertPreciseEqual(f(a, b), usecase(a, b))
+
+        for cases in (
+            (TD(0), TD(1), TD(2), TD('NaT')),
+            (TD(0, 's'), TD(1, 's'), TD(2, 's'), TD('NaT', 's')),
```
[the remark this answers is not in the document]
benwilliamgraham: Mixing times with and without units is currently beyond the scope of this PR, but I'd be happy to open a separate issue for supporting it

esc:
@benwilliamgraham thank you for working on this, your efforts to improve Numba are appreciated. A quick note about force-pushes. In general, we discourage contributors to force push after a review has begun. This is because our experience has shown, that this makes it harder for Github to line up the comments and also makes it harder for reviewers to follow what has been changed etc.. Thank you in advance for complying with this practice! 🙏

njriasan:
@esc Is it possible to get this added to the 0.56 milestone?

esc:
@njriasan it looks like it's pretty far along and probably only needs a sign-off from @sklam ? I'll add it it proactively for now (we can always remove it again, in case things don't line up).

numba/core/typing/builtins.py
```diff
@@ -1103,8 +1103,9 @@ class MinValInfer(AbstractTemplate):
     def generic(self, args, kws):
         assert not kws
         assert len(args) == 1
-        assert isinstance(args[0], (types.DType, types.NumberClass))
-        return signature(args[0].dtype, *args)
+        if isinstance(args[0], (types.DType, types.NumberClass)):
+            return signature(args[0].dtype, *args)
+        assert None
```
sklam, line 1108:
```suggestion
```

sklam (commented):
One last thing

PR #7845, opened 2022-02-15
change time.time() to time.perf_counter() in docs

Nopileos2:
Minor changes to the 5 minute guide:
- time.perf_counter() is preferred way to measure execution time, so best to use it here since it is exactly what should be done
- changed style of string formatting and added 's' as the unit


gmarkall:
Many thanks for the PR - I've queued it for review.

stuartarchibald (approved):
Thanks for the patch, good to keep these examples in sync with common practice.

PR #7846, opened 2022-02-16
Fix CUDA enum vectorize test on Windows

gmarkall:
The default int on Windows is `int32`.

This is to fix the Windows CUDA Buildfarm failures caused by the merge of #7829 without a buildfarm run - I erred in marking that PR RTM without a buildfarm run.


gmarkall:
gpuci run tests

stuartarchibald (approved):
Patch looks good, thanks.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_118`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_118`.

Passed.

PR #7848, opened 2022-02-16
Support for int * list

### LINKED ISSUE #7847: int * list isn't supported
---

## Feature request

operator.mul(int, list) doesn't work inside JIT code. Seems like an easy fix.

```python3
import numba

@numba.njit
def f():
    return 5 * [1]

print(f())
```

Output:
```python3
No implementation of function Function(<built-in function mul>) found for signature:
 
 >>> mul(Literal[int](5), list(int64)<iv=[1]>)
 
There are 12 candidate implementations:
  - Of which 10 did not match due to:
  Overload of function 'mul': File: <numerous>: Line N/A.
    With argument(s): '(int64, list(int64)<iv=None>)':
   No match.
  Operator Overload in function 'mul': File: unknown: Line unknown.
    With argument(s): '(int64, list(int64)<iv=None>)':
   No match for registered cases:
    * (int64, int64) -> int64
    * (int64, uint64) -> int64
    * (uint64, int64) -> int64
    * (uint64, uint64) -> uint64
    * (float32, float32) -> float32
    * (float64, float64) -> float64
    * (complex64, complex64) -> complex64
    * (complex128, complex128) -> complex128
       

During: typing of intrinsic-call at test.py (5)

File "test.py", line 5:
def f():
    return 5 * [1]


<!--

Please include details of the feature you would like to see, why you would
like to see it/the use case.
```

-->

njriasan:
Closes #7847

### DIFF
--- a/numba/core/typing/listdecl.py
+++ b/numba/core/typing/listdecl.py
@@ -153,6 +153,8 @@ class MulList(AbstractTemplate):
         a, b = args
         if isinstance(a, types.List) and isinstance(b, types.Integer):
             return signature(a, a, types.intp)
+        elif isinstance(a, types.Integer) and isinstance(b, types.List):
+            return signature(b, types.intp, b)
 
 
 @infer_global(operator.imul)
--- a/numba/cpython/listobj.py
+++ b/numba/cpython/listobj.py
@@ -714,11 +714,16 @@ def list_add_inplace(context, builder, sig, args):
 
 
 @lower_builtin(operator.mul, types.List, types.Integer)
+@lower_builtin(operator.mul, types.Integer, types.List)
 def list_mul(context, builder, sig, args):
-    src = ListInstance(context, builder, sig.args[0], args[0])
+    if isinstance(sig.args[0], types.List):
+        list_idx, int_idx = 0, 1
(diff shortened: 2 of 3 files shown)


ehsantn (approved):
Looks good to me.

gmarkall:
This seems to be for reflected lists, which are deprecated (https://numba.readthedocs.io/en/stable/reference/deprecation.html#deprecation-of-reflection-for-list-and-set-types) - does it work with typed lists as well?

njriasan:
@gmarkall `operator.mul` doesn't appear to be supported on typed lists at all. I can add support for that, but as a bridge I would like to maintain parity with reflected lists even if they are deprecated.

esc:
> @gmarkall `operator.mul` doesn't appear to be supported on typed lists at all. I can add support for that, but as a bridge I would like to maintain parity with reflected lists even if they are deprecated.

I agree. We don't know when we can finally get rid of the reflected list, so I am also of the opinion that this should go in!

esc (approved):
I tried this locally and can confirm the patch does implement the feature correctly and the test does exercise it. Thank you!

Issue #7849, opened 2022-02-17
CUDA: `ptx` attribute fails for device function


isVoid:
## Reporting a bug

- [x] I have tried using the latest released version of Numba (most recent is
 visible in the change log (https://github.com/numba/numba/blob/main/CHANGE_LOG).
- [x] I have included a self contained code sample to reproduce the problem.
  i.e. it's possible to run as 'python bug.py'.

Min repro:
```python
from numba import cuda

def f(x):
    return x + 3.14

compiled = cuda.jit("float32(float32)", device=True)(f)
print(compiled.ptx)
```

Failure:
```
Traceback (most recent call last):
  File "/home/yhw/dev/numba_explore/cuda_device_get_ptx.py", line 8, in <module>
    print(compiled.ptx)
  File "/home/yhw/dev/numba/numba/cuda/dispatcher.py", line 909, in ptx
    return {sig: overload.ptx for sig, overload in self.overloads.items()}
  File "/home/yhw/dev/numba/numba/cuda/dispatcher.py", line 909, in <dictcomp>
    return {sig: overload.ptx for sig, overload in self.overloads.items()}
AttributeError: 'CompileResult' object has no attribute 'ptx'
```

Expected result (work around, assuming compiled for current device):

Same as following output.
```python
cc = get_current_device().compute_capability
args, return_type = sigutils.normalize_signature("float32(float32)")
lib = compiled.overloads[args].library
ptx = lib.get_asm_str(cc=cc)
```

gmarkall:
This has accidentally survived - getting PTX for decorated device functions was intended to be removed because it had various problems that meant the PTX you got might not be the PTX you expected: https://numba.readthedocs.io/en/stable/reference/deprecation.html#deprecation-of-the-inspect-ptx-method

I documented the removal of `inspect_ptx()` but missed the `ptx` property. The recommended way to get `ptx` for a device function is to use `compile_ptx` or `compile_ptx_for_current_device` instead, e.g.:

```python
from numba import cuda, float32

def f(x):
    return x + 3.14

ptx, resty = cuda.compile_ptx_for_current_device(f, (float32,),
                                                 device=True)
print(ptx)
```

Does this suit your use cases, or is there a use case I've overlooked here?

isVoid:
Got it. `compute_ptx_*` works for my scenario. I wasn't aware `ptx` attribute is about to be deprecated. Is it still recommended to retrieve ptx for kernels with `ptx` attribute?

[...]

Issue #7168, opened 2021-06-30
Impact of flags in function selection/recompilation


stuartarchibald:
Discussion: Impact of flags in function selection/recompilation.

This:
```python
from numba import njit
import numpy as np

@njit
def foo(a, b):
    return a + b

@njit
def caller(a, b):
    return foo(a, b)

@njit(fastmath=True)
def fast_caller(a, b):
    return foo(a, b)

@njit
def main(a, b):
    # SWAP THE ORDER OF THESE CALLS AND RERUN TO SEE THE EFFECT
    x = caller(a, b)
    y = fast_caller(a, b)
    return x, y

print(main(10., 20.))
print(main.inspect_llvm(main.signatures[0]))
```
produces a `fadd double %arg.b, %arg.a` or a `fadd fast double %arg.b, %arg.a` for `foo` that gets inlined into `main` depending on the order of the calling functions.

The flags are propagating from the caller to the callee, but not impacting function selection and potentially triggering recompilation for new variants based on flags. To do this I think typing has to be made aware of the flags, probably by them being part of a function signature, and then this would trigger a recompilation if there's no match. In the above example, were typing aware of the flags for the different calls to `foo`, two variants would be generated, one fast and one normal, and this would be reflected in the LLVM IR.

Discussion points:
1. Performance impact on compile and runtime.
2. Difficulty of implementation.
3. Is this desirable behaviour!?
4. Something else.

sklam:
Here's the traceback on the second time `<fast_caller_dispatcher>.get_call_template()` is called:

```pytb
  /path/to/numba/core/typeinfer.py(589)resolve()
-> sig = typeinfer.resolve_call(fnty, pos_args, kw_args)
  /path/to/numba/core/typeinfer.py(1541)resolve_call()
-> return self.context.resolve_function_type(fnty, pos_args, kw_args)
  /path/to/numba/core/typing/context.py(195)resolve_function_type()
-> res = self._resolve_user_function_type(func, args, kws)
  /path/to/numba/core/typing/context.py(247)_resolve_user_function_type()
-> return func.get_call_type(self, args, kws)
  /path/to/numba/core/types/functions.py(555)get_call_type()
-> self.dispatcher.get_call_template(args, kws)
> /path/to/numba/core/dispatcher.py(376)get_call_template()
-> pysig, args = self._compiler.fold_argument_types(args, kws)
```

The problem is apparently that a `Dispatcher` type do not know or consider the compiler flags.

[...]

PR #7850, opened 2022-02-17
CUDA: Pass `fastmath` compiler flag down to `compile_ptx` and `compile_device`; Improve `fastmath` tests

isVoid:
This PR pass `fastmath` compiler flag down to `compile_ptx` and `compile_device` method. This PR also rewrites `test_fastmath.py` to reduce duplicates and adds coverage of above newly supported methods.


gmarkall:
Looks like the fastmath test needs guarding against some code execution when the simulator is enabled?

isVoid:
I think the issue was because prior to python3.9, type annotations require using type classes imported from the typing module. I was running tests locally with 3.9 so this issue didn't pop up.

isVoid:
In addressing https://github.com/numba/numba/issues/7849#issuecomment-1046328534, `ptx` attribute is no longer accessed, instead `inspect_asm` is used to retrieve machine code.

stuartarchibald:
Thanks for the patch @isVoid, is this now ready for review?

isVoid:
@stuartarchibald Yup, it's ready for review!

stuartarchibald:
NOTE: Must go into the same release as #7858 due to https://github.com/numba/numba/pull/7858#discussion_r813132453

numba/cuda/tests/cudapy/test_fastmath.py
```diff
@@ -136,24 +171,6 @@ def f10(r, x, y):
...
-            if i < arr.size:
-                arr[i] = foo(i, val)
-
-        sig = (float32[::1], float32)
-        fastver = cuda.jit(sig, fastmath=True)(bar)
-        precver = cuda.jit(sig)(bar)
-
-        self.assertIn('div.full.ftz.f32', fastver.ptx[sig])
-        self.assertNotIn('div.full.ftz.f32', precver.ptx[sig])
-
```
gmarkall, line 156: I think this test tested something a little different to all the other test cases in this file - the propagation of fastmath from callers into callees. Presently this doesn't actually work - compiling `bar` with fastmath doesn't result in the same code for `foo` as if it were compiled with fastmath, but it should. 

I think it might be worth reinstating this test case, in a slightly different form:

```python
    @unittest.expectedFailure     
    def test_device_fastmath_propagation(self):    
        # The fastmath option doesn't presently propagate to device functions     
        # from their callees - arguably it should do, so this test is presently     
        # an xfail.     
        @cuda.jit("float32(float32, float32)", device=True)    
        def foo(a, b):    
            return a / b    
      
        def bar(arr, val):    
            i = cuda.grid(1)    
            if i < arr.size:    
                arr[i] = foo(i, val)    
            
        sig = (float32[::1], float32)    
        fastver = cuda.jit(sig, fastmath=True)(bar)    
        precver = cuda.jit(sig)(bar)    
            
        # Variants of the div instruction are further documented at:     
        # https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#floating-point-instructions-div     
    
        # The fast version should use the "fast, approximate divide" variant     
        self.assertIn('div.approx.f32', fastver.ptx[sig])    
        # The precise version should use the "IEEE 754 compliant rounding"     
        # variant, and neither of the "approximate divide" variants.     
        self.assertIn('div.rn.f32', precver.ptx[sig])    
        self.assertNotIn('div.approx.f32', precver.ptx[sig])    
        self.assertNotIn('div.full.f32', precver.ptx[sig])
```

The test case previously ensured our behaviour was consistent with what it was in the past, but I think it's more appropriate to mark this as an expected failure because we know the present behaviour is wrong. However, when changes to the propagation of flags is made (potentially to the core of Numba rather than specific to the CUDA target) it will be good to detect that this gets fixed.

gmarkall (changes requested):
Many thanks for the this - the fastmath tests are a lot easier to read and comprehend now! 

I think this highlights that we might also be missing passing of the `lineinfo` flag to `compile_cuda` in `compile_device`, but I'm happy for us to open a new issue to track investigating that rather than adding it on to this PR.

I have a couple of remarks on the diff, but they're not major :-)

gmarkall:
gpuci run tests

numba/cuda/tests/cudapy/test_fastmath.py
isVoid: > the propagation of fastmath from callers into callees

Ah ok, I didn't realize the test was to test the propagation property. It seems to me propagation should be configurable. But I'd like to investigate more about how much granularity does nvcc allows user to configure different compilation unit with the fast-math flag and Numba should provide similar level of granularity.

gmarkall: There's an issue for some general discussion around the propagation of flags here: https://github.com/numba/numba/issues/7168

gmarkall:
gpuci run tests

gmarkall (changes requested):
This now looks good, except that `TestCompileToPTX.test_fastmath` also needs a fix:

```patch
diff --git a/numba/cuda/tests/cudapy/test_compiler.py b/numba/cuda/tests/cudapy/test_compiler.py
index b8ea4a2a7..cfd549f8b 100644
--- a/numba/cuda/tests/cudapy/test_compiler.py
+++ b/numba/cuda/tests/cudapy/test_compiler.py
@@ -61,7 +61,7 @@ class TestCompileToPTX(unittest.TestCase):
         # With fastmath, ftz and approximate div / sqrt are enabled
         self.assertIn('fma.rn.ftz.f32', ptx)
         # "full" refers to a full-range approximate divide
-        self.assertIn('div.full.ftz.f32', ptx)
+        self.assertIn('div.approx.ftz.f32', ptx)
         self.assertIn('sqrt.approx.ftz.f32', ptx)
 
     def check_debug_info(self, ptx):
```

The test was testing for an incorrect behaviour that you've now fixed in this PR. I recall being a bit confused about why `div.full` and not `div.approx` appeared in the PTX when I wrote the test but never pursued it all the way down.

isVoid:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall (dismissed):
Many thanks for the fixes!

gmarkall:

numba/cuda/tests/cudapy/test_compiler.py
```diff
@@ -61,7 +61,7 @@ def f(x, y, z, d):
         # With fastmath, ftz and approximate div / sqrt are enabled
         self.assertIn('fma.rn.ftz.f32', ptx)
         # "full" refers to a full-range approximate divide
```
stuartarchibald, line 63: Is this comment still applicable?

numba/cuda/tests/cudapy/test_fastmath.py
```diff
@@ -151,8 +188,16 @@ def bar(arr, val):
...
+        # Variants of the div instruction are further documented at:
+        # https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#floating-point-instructions-div
+
+        # The fast version should use the "fast, approximate divide" variant
+        self.assertIn('div.approx.f32', fastver.ptx[sig])
+        # The precise version should use the "IEEE 754 compliant rounding"
+        # variant, and neither of the "approximate divide" variants.
+        self.assertIn('div.rn.f32', precver.ptx[sig])
+        self.assertNotIn('div.approx.f32', precver.ptx[sig])
+        self.assertNotIn('div.full.f32', precver.ptx[sig])
```
stuartarchibald, line 200: `.ptx` is deprecated?

isVoid: Oops, that's my oversight. Fixed.

numba/cuda/tests/cudapy/test_compiler.py
isVoid: Apparently not. `full` isn't seen in the ptx any more.

isVoid:
I reworded the version in contrast from `slow` to `prec` (precise) as I think that's the more accurate term.

esc:
`Build numba_smoketest_cuda_yaml_121 has started`

esc:
I found that with 11.2 the tests the following fails:

```
INFO - ======================================================================
INFO - FAIL: test_wrapper_has_debuginfo (numba.cuda.tests.cudapy.test_debuginfo.TestCudaDebugInfo)
INFO - ----------------------------------------------------------------------
INFO - Traceback (most recent call last):
INFO -   File "F:\ci_envs\64\Miniconda3\envs\testenv_d2a4f607-b781-45cf-b6d9-351951a92ac6\lib\site-packages\numba\cuda\tests\cudapy\test_debuginfo.py", line 90, in test_wrapper_has_debuginfo
INFO -     self.assertIn('noinline !dbg', wrapper_define)
INFO - AssertionError: 'noinline !dbg' not found in 'define void @"_ZN6cudapy5numba4cuda5tests6cudapy14test_debuginfo17TestCudaDebugInfo26test_wrapper_has_debuginfo12_3clocals_3e1fB4v632B96cw51cXTLSUwv1oDHpC0oJQBDAi1SkIQh8Y8kgBqnMAkkq2yr1cHmq1spQCTVQd4C2QYUgQaoOixskISA2oE8g1pNAA_3d_3dE5ArrayIiLi1E1C7mutable7alignedE"(i8* %".1", i8* %".2", i64 %".3", i64 %".4", i32* %".5", i64 %".6", i64 %".7") noinline!dbg !38'
INFO - 
INFO - ----------------------------------------------------------------------
INFO - Ran 1278 tests in 231.506s
INFO - 
INFO - FAILED (failures=1, skipped=50, expected failures=8)
INFO - Command exited with return code 1

```

But I don't think this is a result of this PR, perhaps?

gmarkall:
@esc #7915 was to resolve an issue where the debuginfo looked slightly different due to a change in llvmlite - does the Windows builder have the latest llvmlite?

esc:
> @esc #7915 was to resolve an issue where the debuginfo looked slightly different due to a change in llvmlite - does the Windows builder have the latest llvmlite?

I believe this is the case, yes. However, `main` is now also showing these issues, so it's probably best that I go and look there.

esc:
@gmarkall thank you again for the pointer. It was indeed down to a CI misconfiguration.

stuartarchibald (approved):

stuartarchibald:
@esc thanks for the build farm run.

PR #7855, opened 2022-02-22
Ensure np.argmin/no.argmax return type is intp

### LINKED ISSUE #7853: numpy.argmin(arr, axis=int64), where (arr.ndim > 1), returns passes through the dtype of arr
When operating on a 1D array within a JIT-ed function, `numpy.argmin` always returns an array of type int64, regardless of whether or not you specify `axis=0`. However, when operating on a higher-dimension array, the dtype of the array is passed through to the return value. Here is a simple demonstration script:

```python
import numpy as np
from numba import njit


def test():
    one = np.argmin(np.arange(1024) < 512, axis=0)
    two_zero = np.argmin(np.arange(9).reshape((3, 3)).T < 5, axis=0)
    two_one = np.argmin(np.arange(9).reshape((3, 3)) < 4, axis=1)
    x = (np.arange(1023, -1, -1).reshape((512, 2)) // 4).astype(np.uint8)
    another_test = np.argmin(x, axis=0)
    return one, two_zero, two_one, another_test


if __name__ == '__main__':
    print(test())  # (512, array([0, 2, 0], dtype=int64), array([0, 1, 0], dtype=int64), array([510, 510], dtype=int64))
    print(njit(test)())  # (512, array([False,  True, False]), array([False,  True, False]), array([254, 254], dtype=uint8))
```

This can be avoided by casting the array with `.astype('int64')`, but its rather strange that you would need to do this.

stuartarchibald:
As title. Includes test cases.

Fixes #7853

### DIFF
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -795,7 +795,7 @@ def build_argmax_or_argmin_with_axis_impl(arr, axis, flatten_impl):
     array, return the implementation function.
     """
     check_is_integer(axis, "axis")
-    retty = arr.dtype
+    retty = types.intp
 
     tuple_buffer = tuple(range(arr.ndim))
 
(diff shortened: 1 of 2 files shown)


sklam (approved):
LGTM. 

Also confirmed that the reproducer in #7853 works correctly with this patch.

PR #7858, opened 2022-02-22
CUDA: Deprecate `ptx` Attribute and Update Tests

isVoid:
After speaking with @gmarkall , `ptx` attribute is supposed to be removed as part of effort in https://numba.readthedocs.io/en/stable/reference/deprecation.html#deprecation-of-the-inspect-ptx-method, but was missed. This PR adds deprecation warning to `ptx` attribute and updates all usages in tests.

Part of #7849

### DIFF
--- a/numba/cuda/dispatcher.py
+++ b/numba/cuda/dispatcher.py
@@ -157,8 +157,8 @@ class _Kernel(serialize.ReduceMixin):
         '''
         warnings.warn(
             "Attribute `ptx` is deprecated and will be removed in the future. "
-            "To retrieve compiled machine code of the cuda function, use "
-            "`inpect_asm`."
+            "To retrieve the compiled machine code of the CUDA function, use "
+            "the `inpect_asm` method."
             , FutureWarning
         )
         return self._codelibrary.get_asm_str()
@@ -818,8 +818,8 @@ class CUDADispatcher(Dispatcher, serialize.ReduceMixin):
     def ptx(self):
         warnings.warn(
             "Attribute `ptx` is deprecated and will be removed in the future. "
-            "To retrieve compiled machine code of the cuda function of certain"
-            "signature `sig`, use `inspect_asm(sig)`."
+            "To retrieve the compiled machine code of the CUDA function for a"
+            "given signature `sig`, use the method `inspect_asm(sig)`."
             , FutureWarning
         )
         return {sig: overload.ptx for sig, overload in self.overloads.items()}
--- a/numba/cuda/dispatcher.py
+++ b/numba/cuda/dispatcher.py
@@ -8,7 +8,7 @@ import warnings
 from numba.core import config, serialize, sigutils, types, typing, utils
 from numba.core.compiler_lock import global_compiler_lock
 from numba.core.dispatcher import Dispatcher
-from numba.core.errors import NumbaPerformanceWarning
+from numba.core.errors import NumbaPerformanceWarning, NumbaDeprecationWarning
 from numba.core.typing.typeof import Purpose, typeof
 
 from numba.cuda.api import get_current_device
@@ -159,7 +159,7 @@ class _Kernel(serialize.ReduceMixin):
             "Attribute `ptx` is deprecated and will be removed in the future. "
             "To retrieve the compiled machine code of the CUDA function, use "
             "the `inpect_asm` method."
-            , FutureWarning
+            , NumbaDeprecationWarning
         )
         return self._codelibrary.get_asm_str()
 
@@ -820,7 +820,7 @@ class CUDADispatcher(Dispatcher, serialize.ReduceMixin):
             "Attribute `ptx` is deprecated and will be removed in the future. "
             "To retrieve the compiled machine code of the CUDA function for a"
             "given signature `sig`, use the method `inspect_asm(sig)`."
-            , FutureWarning
+            , NumbaDeprecationWarning
         )
         return {sig: overload.ptx for sig, overload in self.overloads.items()}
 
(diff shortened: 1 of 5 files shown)


numba/cuda/dispatcher.py
```diff
@@ -154,6 +155,12 @@ def ptx(self):
         '''
         PTX code for this kernel.
         '''
+        warnings.warn(
+            "Attribute `ptx` is deprecated and will be removed in the future. "
+            "To retrieve compiled machine code of the cuda function, use "
+            "`inpect_asm`."
+            , FutureWarning
```
stuartarchibald, line 162: Think this should be a `numba.core.errors.NumbaDeprecationWarning` ?

numba/cuda/dispatcher.py
```diff
@@ -809,6 +816,12 @@ def inspect_types(self, file=None):
 
     @property
     def ptx(self):
+        warnings.warn(
```
stuartarchibald, line 819: Suggest adding tests to `numba.cuda.tests.cudapy.test_dispatcher` to ensure that these warnings appear correctly (both from the dispatcher object and the kernel object). A pattern to follow might be similar to:
https://github.com/numba/numba/blob/290bcaa63804fb99b1fe238b0f8fe1caf7686899/numba/tests/test_deprecations.py#L16-L29

stuartarchibald (commented):
Thanks for the patch and performing maintenance on the CUDA target. There's a few minor things to resolve but otherwise looks good.

numba/cuda/tests/cudapy/test_sync.py
```diff
@@ -206,7 +206,7 @@ def test_threadfence_system_codegen(self):
         compiled[1, 1](ary)
         self.assertEqual(123 + 321, ary[0])
         if not ENABLE_CUDASIM:
-            self.assertIn("membar.sys;", compiled.ptx[sig])
+            self.assertIn("membar.sys;", compiled.inspect_asm(sig))
```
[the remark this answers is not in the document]
isVoid: Ah, these tests are about to be rewritten in https://github.com/numba/numba/pull/7850, that's why I didn't cover them in this PR.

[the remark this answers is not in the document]
stuartarchibald: I see, I think so long as #7850 and this patch go into the same release that's fine. It seems like it'd create work to change the `.ptx` usage here and then immediately remove it in the next logical patch to the CUDA target. @gmarkall as code owner are you ok with this approach?

numba/cuda/dispatcher.py
isVoid: Test is added.

numba/cuda/tests/cudapy/test_dispatcher.py
```diff
@@ -288,5 +290,33 @@ def add_device(a, b):
...
+@skip_on_cudasim('Dispatcher objects not used in the simulator')
+class TestDispatcherDeprecation(CUDATestCase):
+    def check_warning(self, warnings, expected_str, category):
+        self.assertEqual(len(warnings), 3)
+        for w in warnings:
+            self.assertEqual(w.category, category)
+            self.assertIn(expected_str, str(w.message))
+
+    def test_ptx_deprecation(self):
+        with warnings.catch_warnings(record=True) as w:
```
stuartarchibald, line 302: This probably needs to call: https://github.com/numba/numba/blob/53ea89fee1a19471ce1b1907f9f3d98acd1b6730/numba/tests/support.py#L141 so as to not break if any of the internal, and safe to ignore, warnings are raised.

stuartarchibald (commented):
Thanks for the update, there's a couple of minor things to resolve else looks good. Thanks again!

numba/cuda/tests/cudapy/test_sync.py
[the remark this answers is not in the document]
gmarkall: I'm OK with this approach.

numba/cuda/tests/cudapy/test_dispatcher.py
```diff
@@ -288,5 +291,42 @@ def add_device(a, b):
...
+
+        with warnings.catch_warnings(record=True) as w:
+            ignore_internal_warnings()
+            warnings.simplefilter("always", category=NumbaDeprecationWarning)
+
+            msg = "Attribute `ptx` is deprecated and will be removed in the "
+            "future. "
+
+            # Kernel code path
+            with self.subTest(name="DispatcherOnCudaKernel"):
```
stuartarchibald, line 314: I'm wondering if the warnings list is stateful and as a result these tests could break in strange ways? e.g. should some part of one of these `subTest`s fail such that the `subTest` context is exited early, the warnings list would not be cleared (as `w.clear()` wouldn't be reached) and as a result the next `subTest` would erroneously fail as the warnings list is too long. I think the fix is maybe to copy the statements `with warnings.catch_warnings...` through to the `warnings.simplefilter...` into each subtest, and then to remove the `w.clear()`?

isVoid: Since we will recreate the warning catcher context manager repeatedly, I created a custom ctx manager for re-usability.

stuartarchibald (commented):
Thanks for the update, there's a single typo `cuda`->`CUDA` to fix, else looks good.

@gmarkall are you happy with this being tested solely on `gpu ci`?

gmarkall:
@stuartarchibald I'm happy with just the gpuci tests.

gpuci run tests

gmarkall:
gpuci run tests

stuartarchibald (approved):

PR #7860, opened 2022-02-23
Path finding for JIT compiling CUDA launch sites

stuartarchibald:
**NOTE: Do not merge or review this with view to merge, it's exploratory.**

As per the title, this somewhat scrappy patch is path finding for what it takes to JIT compile the launch sites for the CUDA target. The basic idea is to ctypes bind to the CUDA kernel launch function, close over a kernel pointer and then generate a JIT call that splats the arguments from a CPU dispatcher matching the launch site into the `void**` that a CUDA kernel expects. There's some additional complexity involved in that NumPy arrays need round tripping to the device if they are present, `@overload`s calling `objectmode` handle this but it would also be possible to just JIT a similar call to the CUDA library function to copy the memory etc.

Current things that work:
* Standard kernel launches (not cooperative launch) going via CPU JIT code.
* Automatic to/from device handling of NumPy arrays (see caveat below about tuples).
* Most of the unit tests!

Current issues:
* 36 of the unittests are skipped due to problems (see 256ac65). Of those which are more serious failures, there seems to be three or four common themes in the failing tests so it's probably not that many individual problems.
* CUDA's device array needs to be its own type, but it's a bit too close to `types.Array` in some cases which prevents specialisation. e.g. if you compile a function with a NumPy array first it won't recompile when a CUDADeviceArray type is presented.
* CUDA's device array boxing is too raw, the `MemoryPointer` needs more consideration, at present it's just using an int as the address and MemoryPointer is hacked to cope with this. I've also some unspecified concerns about references and lifetimes of these managed pointers in relation to this, it just seems likely an area that could present issues.
* Record arrays are entirely broken as they are ignored by the current patch (cause of some of the unittest skips).
* Cooperative launches are still going via the ctypes path.
* Exception handling is still done via ctypes, does not propagate through the JIT launch.
* Passing `tuple`s containing `np.ndarray` as kernel args is not handled, the tuple needs duplicating at runtime with a switch out for a device array and back (cause of some of the unittest skips). It's doable but prohibitively complicated for this pathfinding.
* "extensions" arg to the kernel does not work. This should probably be handled via the type system as it looks like it's just a workaround for externally defined types.

Current issues that could be problems in the future:
* Closing over the kernel function pointer will silently break on disk caching (addresses baked in).
* There's something going on with types of arrays that ends up with them often being "A" order which is a less specific than the CPU dispatcher default/preference. I think this is all fixable (see outcomes), but it might alter the way users code is typed.

Outcomes:
* This patch proves that JIT kernel launch is possible. On a NOP 4 arg kernel launch is about 1.75x faster.
* I'm now pretty convinced that this refactor cannot be completed without also changing the way kernel launches are structured. It seems like the kernel is compiled and specialised on arguments too early in the launch sequence, which means a CPU JIT launcher can struggle to match the args at the call site for the kernel. This comes from the kernel essentially being compiled with a signature which then blocks further compilation and permits various degrees of casting to the arguments, this is mostly fine, but there's cases where it is not. For example, if a kernel ends up compiled with "A" ordered arrays but the CPU JIT code decides they are "C" ordered, it breaks. Similarly if a kernel has a signature specified saying it e.g. takes a float32, but then the user executes it with a python float, the CPU JIT function will type that as float64 and then the kernel packed args will be of the wrong types which ends up with incorrect results/segfaults. I'm of the view that the kernel, launch config and CUDA dispatcher all need to be type specialising dispatchers in their own right such that types and signatures correctly propagate through a call graph like `cuda dispatcher - > launch config -> kernel`. This also has the secondary effect of making is reasonably likley that launching CUDA kernels from CPU jitted functions would "just work" or at least work with a minor amount of effort.

### DIFF
--- a/numba/cuda/dispatcher.py
+++ b/numba/cuda/dispatcher.py
@@ -4,10 +4,13 @@ import sys
 import ctypes
 import functools
 
-from numba.core import config, serialize, sigutils, types, typing, utils
+from numba.core import (config, serialize, sigutils, types, typing, utils,
+                        cgutils)
 from numba.core.compiler_lock import global_compiler_lock
 from numba.core.dispatcher import Dispatcher
 from numba.core.errors import NumbaPerformanceWarning
+from numba.core.extending import overload
+from numba import objmode
 from numba.core.typing.typeof import Purpose, typeof
 
 from numba.cuda.api import get_current_device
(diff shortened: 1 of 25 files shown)


numba/cuda/dispatcher.py
```diff
@@ -230,11 +396,13 @@ def max_cooperative_grid_blocks(self, blockdim, dynsmemsize=0):
         return active_per_sm * sm_count
 
     def launch(self, args, griddim, blockdim, stream=0, sharedmem=0):
+        from .cudadrv import enums
         # Prepare kernel
         cufunc = self._codelibrary.get_cufunc()
```
stuartarchibald, line 401: Is the `cufunc` sufficiently invariant for it to be cached and/or baked into the launcher?

gmarkall: It's invariant for the duration of a single process, but can't be cached on-disk as it will be different each time the module is loaded to the device.

gmarkall: (Also. it's not invariant across devices)

stuartarchibald:
Closing as path finding work is complete.

PR #7861, opened 2022-02-23
Fix a spelling mistake in README

gzz2000:
instuction => instruction


stuartarchibald (approved):
Well spotted, thanks for the patch!

PR #7865, opened 2022-02-25
Remove add_user_function

gmarkall:
As per
https://numba.readthedocs.io/en/stable/reference/deprecation.html#deprecation-of-numba-core-base-basecontext-add-user-function


stuartarchibald (approved):
Thanks for following up on this deprecation notice and supplying the patch!

PR #7866, opened 2022-02-25
Support for large numbers of args/kws with Python 3.10

### LINKED ISSUE #7812: Providing many keyword arguments triggers assertion error with Python 3.10 due to unsupported bytecode
It seems that Python 3.10 produces a different bytecode instruction, `CALL_FUNCTION_EX` when using a large number of keyword arguments. From my testing it seems like this occurs once you reach 16 keyword arguments. I only see this behavior on Python 3.10 and not Python 3.9. Here is a reproducer:

```python
import numba

@numba.njit
def my_func1(a=None, b=None, c=None, d=None, e=None, f=None, g=None, h=None, i=None, j=None, k=None, l=None, m=None, n=None, o=None, p=None):
    return my_func2(
        a=a,
        b=b,
        c=c,
        d=d,
        e=e,
        f=f,
        g=g,
        h=h,
        i=i,
        j=j,
        k=k,
        l=l,
        m=m,
        n=n,
        o=o,
        p=p
    )


@numba.njit
def my_func2(a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p):
    return


print(my_func1())
```

njriasan:
Closes #7812. This seems to be working for the set of tests I have covered. I think the code quality probably needs to improve considerably. In addition any code to test for unintended side effects is greatly appreciated.

### DIFF
--- a/numba/core/interpreter.py
+++ b/numba/core/interpreter.py
@@ -5,7 +5,7 @@ import operator
 import logging
 
 from numba.core import errors, dataflow, controlflow, ir, config
-from numba.core.errors import NotDefinedError, error_extras
+from numba.core.errors import NotDefinedError, UnsupportedError, error_extras
 from numba.core.utils import (PYVERSION, BINOPS_TO_OPERATORS,
                               INPLACE_BINOPS_TO_OPERATORS,)
 from numba.core.byteflow import Flow, AdaptDFA, AdaptCFA
@@ -77,6 +77,153 @@ class Assigner(object):
             return self.dest_to_src[destname]
         self.unused_dests.discard(destname)
         return None
+
+
+def peep_hole_call_function_ex_to_call_function_kw(func_ir):
+    """
+    This peephole rewrites a bytecode sequence unique to Python 3.10
+    where CALL_FUNCTION_EX is used instead of CALL_FUNCTION_KW because of
+    stack limitations set by CPython. This limitation is imposed whenever
+    a function call has too many arguments or keyword arguments.
+
+    https://github.com/python/cpython/blob/a58ebcc701dd6c43630df941481475ff0f615a81/Python/compile.c#L55
+    https://github.com/python/cpython/blob/a58ebcc701dd6c43630df941481475ff0f615a81/Python/compile.c#L4442
+
+    In particular, change is imposed whenever (n_args / 2) + n_kws > 15.
+
+    Different bytecode is generated for args depending on if n_args > 30.
+    With n_args <= 30, the bytecode looks like:
+
+        # Start for each argument
+        LOAD_FAST  # Load each argument.
+        # End for each argument
+        ...
+        BUILD_TUPLE # Create a tuple of the arguments
+
+    whereas with n_args > 30 the bytecode looks like:
+
+        BUILD_TUPLE # Create a list to append to
+        # Start for each argument
+        LOAD_FAST  # Load each argument.
+        LIST_APPEND # Add the argument to the list
+        # End for each argument
+        ...
+        LIST_TO_TUPLE # Convert the args to a tuple.
+
+    Similar the bytecode is different for n_kws > 15.
+    With n_kws <= 15, the bytecode looks like:
+
+        # Start for each argument
+        LOAD_FAST  # Load each argument.
+        # End for each argument
+        ...
+        BUILD_CONST_KEY_MAP # Build a map
+
+    whereas with n_kws > 15, the bytecode looks like:
+
+        BUILD_MAP # Construct the map
+        # Start for each argument
+        LOAD_CONST # Load a constant for the name of the argument
+        LOAD_FAST  # Load each argument.
+        MAP_ADD # Append the (key, value) pair to the map
+        # End for each argument
+
+    Finally at the end the bytecode will contain a CALL_FUNCTION_EX instruction.
+
+    This optimizations unwraps the *args and **kwargs in the function call
+    and places these values directly into the args and kwargs.
+
+    """
+    # All changes are local to the a single block
+    # so it can be traversed in any order.
+    for blk in func_ir.blocks.values():
+        new_body = []
+        for i, stmt in enumerate(blk.body):
+            if (
+                isinstance(stmt, ir.Assign)
+                and isinstance(stmt.value, ir.Expr)
+                and stmt.value.op == "call"
+            ):
+                call = stmt.value
+                args = call.args
+                kws = call.kws
+                # We search for replace if a call has either vararg
+                # or varkwarg.
+                vararg = call.vararg
+                varkwarg = call.varkwarg
+                start_search = i - 1
+                if varkwarg is not None:
+                    # varkwarg should be defined second so we start there.
+                    varkwarg_loc = start_search
+                    keyword_def = None
+                    not_found = True
+                    while varkwarg_loc >= 0 and not_found:
+                        keyword_def = blk.body[varkwarg_loc]
+                        if (
+                            isinstance(keyword_def, ir.Assign)
+                            and keyword_def.target.name == varkwarg.name
+                        ):
+                            not_found = False
+                        else:
+                            varkwarg_loc -= 1
+                    start_search = varkwarg_loc
+                    if (
+                        kws
+                        or not_found
+                        or not (
+                            isinstance(keyword_def.value, ir.Expr)
+                            and keyword_def.value.op == "build_map"
+                        )
+                    ):
+                        # If we couldn't find where the kwargs are created
+                        # then it should be a normal **kwargs call
+                        # so we produce an unsupported message.
+                        errmsg = "CALL_FUNCTION_EX with **kwargs not supported"
+                        raise UnsupportedError(errmsg)
+                    # Extract the kws from the build_map
+                    kws = keyword_def.value.items.copy()
+                    # kws are required to have constants for all of the strings.
+                    # We update these with the value_indexes
+                    for key, index in keyword_def.value.value_indexes.items():
+                        kws[index] = (key, kws[index][1])
+
+                    # Remove the build_map from the body.
+                    new_body[varkwarg_loc] = None
+                if vararg is not None:
+                    # Vararg isn't required to be provided.
+                    vararg_loc = start_search
+                    args_def = None
+                    not_found = True
+                    while vararg_loc >= 0 and not_found:
+                        args_def = blk.body[vararg_loc]
+                        if (
+                            isinstance(args_def, ir.Assign)
+                            and args_def.target.name == vararg.name
+                        ):
+                            not_found = False
+                        else:
+                            vararg_loc -= 1
+                    # If we found args update them
+                    if not not_found:
+                        if (
+                            isinstance(args_def.value, ir.Expr)
+                            and args_def.value.op == "build_tuple"
+                        ):
+                            args = args + args_def.value.items
+                            # Remove the build_tuple from the body.
+                            new_body[vararg_loc] = None
+                if vararg or varkwarg:
+                    # Create a new call updating the args and kws
+                    new_call = ir.Expr.call(
+                        call.func, args, kws, call.loc, target=call.target
+                    )
+                    # Update the statement
+                    stmt = ir.Assign(new_call, stmt.target, stmt.loc)
+
+            new_body.append(stmt)
+        # Update the block body
+        blk.body = [x for x in new_body if x is not None]
+    return func_ir
 
 
 def peep_hole_list_to_tuple(func_ir):
@@ -383,6 +530,8 @@ class Interpreter(object):
         peepholes = []
         if PYVERSION in [(3, 9), (3, 10)]:
             peepholes.append(peep_hole_list_to_tuple)
+        if PYVERSION == (3, 10):
+            peepholes.append(peep_hole_call_function_ex_to_call_function_kw)
         peepholes.append(peep_hole_delete_with_exit)
 
         post_processed_ir = self.post_process(peepholes, func_ir)
@@ -1217,10 +1366,13 @@ class Interpreter(object):
             expr = ir.Expr.call(func, posvals, keyvalues, loc=self.loc)
             self.store(expr, res)
 
-        def op_CALL_FUNCTION_EX(self, inst, func, vararg, res):
+        def op_CALL_FUNCTION_EX(self, inst, func, vararg, varkwarg, res):
             func = self.get(func)
             vararg = self.get(vararg)
-            expr = ir.Expr.call(func, [], [], loc=self.loc, vararg=vararg)
+            varkwarg = self.get(varkwarg)
+            expr = ir.Expr.call(
+                func, [], [], loc=self.loc, vararg=vararg, varkwarg=varkwarg
+            )
             self.store(expr, res)
 
     def _build_tuple_unpack(self, inst, tuples, temps, is_assign):
--- a/numba/core/interpreter.py
+++ b/numba/core/interpreter.py
@@ -345,10 +345,8 @@ def _call_function_ex_replace_args_large(
     else:
         # There must always be an initial assignement
         # https://github.com/numba/numba/blob/59fa2e335be68148b3bd72a29de3ff011430038d/numba/core/interpreter.py#L259-L260
-        # If this changes we may need to support this branch. We can
-        # reach this section if there is an unsupported conversion due
-        # to a failure in the list_to_tuple conversion.
-        raise UnsupportedError(errmsg)
+        # If this changes we may need to support this branch.
+        raise AssertionError("unreachable")
     start_not_found = True
     # Traverse backwards to find all concatentations
     # until eventually reaching the original empty tuple.
@@ -623,6 +621,14 @@ Replace that with
                             vararg_loc,
                             func_ir,
                         )
+                    elif (
+                        isinstance(args_def.value, ir.Expr)
+                        and args_def.value.op == "list_to_tuple"
+                    ):
+                        # If there is a call with vararg we need to check
+                        # if the list -> tuple conversion failed and if so
+                        # throw an error.
+                        raise UnsupportedError(errmsg)
                     else:
                         # Here the IR is an initial empty build_tuple.
                         # Then for each arg, a new tuple with a single
--- a/numba/core/interpreter.py
+++ b/numba/core/interpreter.py
@@ -345,7 +345,10 @@ def _call_function_ex_replace_args_large(
         _remove_assignment_definition(old_body, search_end, func_ir)
         search_end -= 1
     else:
-        target_name = vararg_stmt.target.name
+        # There must always be an initial assignement
+        # https://github.com/numba/numba/blob/59fa2e335be68148b3bd72a29de3ff011430038d/numba/core/interpreter.py#L259-L260
+        # If this changes we may need to support this branch.
+        raise AssertionError("unreachable")
     start_not_found = True
     # Traverse backwards to find all concatentations
     # until eventually reaching the original empty tuple.
@@ -400,7 +403,8 @@ def _call_function_ex_replace_args_large(
                 # We cannot handle this format.
                 raise UnsupportedError(errmsg)
             if arg_tuple_stmt.target.name == lhs_name:
-                target_name = rhs_name
+                # The tuple should always be generated on the RHS.
+                raise AssertionError("unreachable")
             elif arg_tuple_stmt.target.name == rhs_name:
                 target_name = lhs_name
             else:
--- a/numba/core/interpreter.py
+++ b/numba/core/interpreter.py
@@ -137,6 +137,7 @@ def peep_hole_call_function_ex_to_call_function_kw(func_ir):
     # All changes are local to the a single block
     # so it can be traversed in any order.
     for blk in func_ir.blocks.values():
+        blk_changed = False
         new_body = []
         errmsg = "CALL_FUNCTION_EX with **kwargs not supported"
         for i, stmt in enumerate(blk.body):
@@ -144,7 +145,9 @@ def peep_hole_call_function_ex_to_call_function_kw(func_ir):
                 isinstance(stmt, ir.Assign)
                 and isinstance(stmt.value, ir.Expr)
                 and stmt.value.op == "call"
+                and stmt.value.varkwarg is not None
             ):
+                blk_changed = True
                 call = stmt.value
                 args = call.args
                 kws = call.kws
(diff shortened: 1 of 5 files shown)


benwilliamgraham (dismissed):
Looks good!

numba/core/interpreter.py
```diff
@@ -79,6 +79,432 @@ def get_assignment_source(self, destname):
...
+                        and keyword_def.value.op == "build_map"
+                    )
+                ):
+                    # If we couldn't find where the kwargs are created
+                    # then it should be a normal **kwargs call
+                    # so we produce an unsupported message.
+                    raise UnsupportedError(errmsg)
+                # Determine the kws
+                if keyword_def.value.items:
+                    # n_kws <= 15 case.
```
ehsantn, line 442: I think it would be helpful to add example IRs in comments for each case that is being matched here.

numba/core/interpreter.py
```diff
@@ -79,6 +79,432 @@ def get_assignment_source(self, destname):
...
+                        and args_def.value.op == "build_tuple"
+                    ):
+                        # n_args <= 30 case.
+                        args = _call_function_ex_replace_args_small(
+                            args_def.value,
+                            new_body,
+                            vararg_loc
+                        )
+                    else:
+                        args = _call_function_ex_replace_args_large(
```
ehsantn, line 491: Add comment on the expected IR node for args_def.

numba/core/interpreter.py
```diff
@@ -79,6 +79,432 @@ def get_assignment_source(self, destname):
...
+    n_kws <= 15 and the bytecode looks like:
+
+        BUILD_MAP # Construct the map
+        # Start for each argument
+        LOAD_CONST # Load a constant for the name of the argument
+        LOAD_FAST  # Load each argument.
+        MAP_ADD # Append the (key, value) pair to the map
+        # End for each argument
+
+    In the IR generated, the inital build map is empty and a series
```
ehsantn, line 127:
```suggestion
    In the IR generated, the initial build map is empty and a series
```

ehsantn (commented):
Thanks for the effort @njriasan ! I took a quick look and the code makes sense to me.

numba/core/interpreter.py
njriasan: Done!

numba/core/interpreter.py
njriasan: I added a comment explaining my assumptions. Please let me know if it needs further clarification.

njriasan:
@ehsantn and @benwilliamgraham thanks for doing an initial review.

@stuartarchibald could you mark this PR as ready for review? Thanks.

numba/core/interpreter.py
```diff
@@ -79,6 +79,488 @@ def get_assignment_source(self, destname):
...
+                and getattr_stmt.value.op == "getattr"
+                and (
+                    getattr_stmt.value.value.name
+                    == buildmap_name
+                )
+                and getattr_stmt.value.attr == "__setitem__"
+            ):
+                not_found_getattr = False
+            else:
+                search_start += 1
```
sklam, line 180: Why is this needed and what would trigger this line?

numba/core/interpreter.py
```diff
@@ -79,6 +79,488 @@ def get_assignment_source(self, destname):
...
+    if (
+        isinstance(vararg_stmt, ir.Assign)
+        and isinstance(vararg_stmt.value, ir.Var)
+    ):
+        target_name = vararg_stmt.value.name
+        # If there is an initial assignment, delete it
+        new_body[search_end] = None
+        search_end -= 1
+    else:
+        target_name = vararg_stmt.target.name
```
sklam, line 288: Why is this needed and what would trigger this line?

numba/core/interpreter.py
```diff
@@ -79,6 +79,488 @@ def get_assignment_source(self, destname):
...
+                and (
+                    arg_tuple_stmt.value.op
+                    == "build_tuple"
+                )
+                and len(arg_tuple_stmt.value.items) == 1
+            ):
+                # We cannot handle this format.
+                raise UnsupportedError(errmsg)
+            if arg_tuple_stmt.target.name == lhs_name:
+                target_name = rhs_name
```
sklam, line 341: Impossible case? cpython doesn't seem to put the tuple on the LHS.

numba/core/interpreter.py
```diff
@@ -79,6 +79,488 @@ def get_assignment_source(self, destname):
...
+                if (
+                    isinstance(next_stmt, ir.Assign)
+                    and (
+                        next_stmt.target.name
+                        == target_name
+                    )
+                ):
+                    keep_looking = False
+                else:
+                    search_end -= 1
```
sklam, line 366: Why is this needed and what would trigger this line?

sklam (commented):
I did a quick check on coverage and found the following lines to be never covered. I only checked a random 20% of the test suite and all the new tests thus the coverage data may not be 100% accurate. If these lines are not really dead, they should have more comments to explain them.

numba/core/interpreter.py
njriasan: This is necessary if some intermediate operation is needed to create the argument. For example, if one of the args was `arg5 = bool(n)`, then search_start skips over the function call.

I can update the comment and the tests to include a function call.

numba/core/interpreter.py
njriasan: This may never be needed, but I wasn't sure the code will always produce an assignment (although I can't construct a case where it hasn't happened).

Basically the existing IR looks like:

```python
...
$final_var = build_tuple(items=[Var(argn, test.py:6)])
$final_tuple = $prev_tuple + $final_var
$varargs_var = $final_tuple
```
but it would be completely valid if instead the IR generated was
```python
...
$final_var = build_tuple(items=[Var(argn, test.py:6)])
$varargs_var = $prev_tuple + $final_var
```
so I was trying to protect against this ever changing.

njriasan: If you are confident this extra assignment can't occur I'm happy to remove it.

numba/core/interpreter.py
njriasan: I agree I also never saw this on the LHS. I was simply trying to be conservative against any slight changes in the generated code. Since this is a property of the bytecode should I just remove this case?

numba/core/interpreter.py
njriasan: Same reason as the KWS case. If there is a function call like `bool(n)`

njriasan: If there is anything need to "create" the arg, for example a function call.

njriasan:
> I did a quick check on coverage and found the following lines to be never covered. I only checked a random 20% of the test suite and all the new tests thus the coverage data may not be 100% accurate. If these lines are not really dead, they should have more comments to explain them.

Thanks @sklam. I'll update my tests to add checks to two of the regions and I'll expand the comments on those regions.

Two of the areas that are uncovered were my attempts to be conservative against minor changes in CPython or the IR generated because I wasn't sure if my tests are exhaustive. If you feel like they are unnecessary I'm happy to remove them.

njriasan:
> > I did a quick check on coverage and found the following lines to be never covered. I only checked a random 20% of the test suite and all the new tests thus the coverage data may not be 100% accurate. If these lines are not really dead, they should have more comments to explain them.
> 
> Thanks @sklam. I'll update my tests to add checks to two of the regions and I'll expand the comments on those regions.
> 
> Two of the areas that are uncovered were my attempts to be conservative against minor changes in CPython or the IR generated because I wasn't sure if my tests are exhaustive. If you feel like they are unnecessary I'm happy to remove them.

@sklam I pushed an update to the tests and updated my comments.

numba/core/interpreter.py
sklam: The extra assignment is coming from https://github.com/numba/numba/blob/59fa2e335be68148b3bd72a29de3ff011430038d/numba/core/interpreter.py#L259-L260. 

Let's make this branch raise an error to indicate it is unreachable and add a comment using your description in https://github.com/numba/numba/pull/7866#discussion_r817233996.

numba/core/interpreter.py
sklam: Let's make this raise an error indicating it is unreachable/impossible.

stuartarchibald:
I don't think at present the transform will be able to handle inline control flow at the call site. For example, the rather contrived:
```python
import numba
import dis

@numba.njit
def my_func1(a=None, b=None, c=None, d=None, e=None, f=None, g=None, h=None, i=None, j=None, k=None, l=None, m=None, n=None, o=None, p=None):
    return my_func2(
        a=a,
        b=b,
        c=c,
        d=d,
        e=e,
        f=f,
        g=g,
        h=h,
        i=i,
        j=j,
        k=k,
        l=l,
        m=m,
        n=n,
        o=o,
        p = False if a else True
    )


@numba.njit
def my_func2(a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p):
    return

print(dis.dis(my_func1.py_func))
print(my_func1())
```
which gives a bytecode that has:
```
  6          96 LOAD_CONST              16 ('p')

 22          98 LOAD_FAST                0 (a)
            100 POP_JUMP_IF_FALSE       55 (to 110)
            102 LOAD_CONST              17 (False)
            104 MAP_ADD                  1

  6         106 CALL_FUNCTION_EX         1
            108 RETURN_VALUE

 22     >>  110 LOAD_CONST              18 (True)
            112 MAP_ADD                  1

  6         114 CALL_FUNCTION_EX         1
```
at the location where `p` is resolved and added to the map prior to the call.

stuartarchibald:
> @stuartarchibald Thank you for finding these cases! You are definitely correct and I need to work on a more complete solution.

No problem.

I think given the rarity of the occurrence of these patterns, it would acceptable to do the following:
* Update this patch to catch the cases where control flow occurs. Inform users that control flow on arguments within call sites is not supported and refer them to docs on how to adjust their code.
* Add tests for the above. 

The most common use cases of this somewhat rare pattern would then be covered and workarounds for the uncommon use cases of this rare patter would be shown to users. Essentially, this patch makes the rare patterns work much more often, so I think it's better that it is reviewed and merged with acknowledgement of the imperfection instead of trying to cover everything. This of course also doesn't preclude further fixes later. What do you think?

njriasan:
> > @stuartarchibald Thank you for finding these cases! You are definitely correct and I need to work on a more complete solution.
> 
> No problem.
> 
> I think given the rarity of the occurrence of these patterns, it would acceptable to do the following:
> 
> * Update this patch to catch the cases where control flow occurs. Inform users that control flow on arguments within call sites is not supported and refer them to docs on how to adjust their code.
> * Add tests for the above.
> 
> The most common use cases of this somewhat rare pattern would then be covered and workarounds for the uncommon use cases of this rare pattern would be shown to users. Essentially, this patch makes the rare patterns work much more often, so I think it's better that it is reviewed and merged with acknowledgement of the imperfection instead of trying to cover everything. This of course also doesn't preclude further fixes later. What do you think?

@stuartarchibald I think this approach is reasonable and I agree with it.

njriasan:
@sklam I made your requested changes.

@stuartarchibald I made your changes as well and added tests. Proper error handling requires some extra code, so there may need to be to be a more detailed review. I also added a test for objmode which fails when the definitions are not properly updated.

numba/core/interpreter.py
```diff
@@ -79,6 +79,647 @@ def get_assignment_source(self, destname):
...
+                not_found = True
+                while vararg_loc >= 0 and not_found:
+                    args_def = blk.body[vararg_loc]
+                    if (
+                        isinstance(args_def, ir.Assign)
+                        and args_def.target.name == vararg.name
+                    ):
+                        not_found = False
+                    else:
+                        vararg_loc -= 1
```
sklam, line 704: (Leaving thought trail in case we need it in the future. This probably doesn't present to be a problem)

With bytecode from:

```python
args = (*(0, 1, 2),)   # the starunpack triggers LIST_TO_TUPLE
y = my_func2(*args)
```
```
 16           0 BUILD_LIST               0
              2 LOAD_CONST               1 ((0, 1, 2))
              4 LIST_EXTEND              1
              6 LIST_TO_TUPLE
              8 STORE_FAST              31 (args)

 17          10 LOAD_GLOBAL              0 (my_func2)
             12 LOAD_FAST               31 (args)
             14 CALL_FUNCTION_EX         0
             16 STORE_FAST              32 (y)
```
the code here scan back from the call to `my_func2` to find the definition of `args`, which is suppose to be a `LIST_TO_TUPLE` opcode, but is replaced by `peep_hole_list_to_tuple`; otherwise, the later check will reject this call.

sklam: I have also tried to trick this part of the code with the walrus operator but Numba is already miscompiling this on mainline:

```python
my_func2(args:=(*(0, 1, 2),), *args)
```

numba/core/interpreter.py
```diff
@@ -79,7 +79,38 @@ def get_assignment_source(self, destname):
...
+        #   172list_to_tuple.116 = $170list_append.115
+        #
+        # but the definition is:
+        #
+        #   list_to_tuple(info=('$4build_list.1',))
+        #
+        func_ir._definitions[lhs].clear()
+    else:
+        raise UnsupportedError(
+            "Inconsistency found the in definitions trying to remove varkwargs"
```
sklam, line 107: Is the error message still accurate? This function seem to be applied generally not just on varkwargs.

njriasan: How is this message? Also thank you for the ping, I missed the original review.

numba/core/interpreter.py
```diff
@@ -104,7 +104,9 @@ def _remove_assignment_definition(old_body, idx, func_ir):
         func_ir._definitions[lhs].clear()
     else:
         raise UnsupportedError(
-            "Inconsistency found the in definitions trying to remove varkwargs"
+            "Inconsistency found in the definitions while electuting"
```
sklam, line 107:
```suggestion
            "Inconsistency found in the definitions while executing"
```

numba/core/interpreter.py
```diff
@@ -79,6 +79,649 @@ def get_assignment_source(self, destname):
...
+you have a large number of kwargs and are using inlined control
+flow. You can resolve this issue by moving the control flow out of
+the function call. For example, if you have
+
+    f(a=1 if flag else 0, ...)
+
+Replace that with
+
+    a_val = 1 if flag else 0
+    f(a=a_val, ...)
```
stuartarchibald, line 494: Perhaps `textwrap.dedent` would let this be indented at a more expected level?

numba/core/interpreter.py
```diff
@@ -79,6 +79,649 @@ def get_assignment_source(self, destname):
...
+                kws = call.kws
+                # We search for replace if a call has either vararg
+                # or varkwarg.
+                vararg = call.vararg
+                varkwarg = call.varkwarg
+                start_search = i - 1
+                # varkwarg should be defined second so we start there.
+                varkwarg_loc = start_search
+                keyword_def = None
+                not_found = True
```
stuartarchibald, line 518:
```suggestion
                found = False
```
(and fix up)?

numba/tests/test_interpreter.py
```diff
@@ -0,0 +1,641 @@
...
+            'You can resolve this issue by moving the control flow out',
+            str(raises.exception)
+        )
+        # Uncomment sections when inlined control flow is
+        # actually supported.
+        # py_func = inline_func
+        # cfunc = njit()(inline_func)
+        # a = py_func(False)
+        # b = cfunc(False)
+        # self.assertEqual(a, b)
```
stuartarchibald, line 396: Can probably just remove this, it's easy to add in if/when needed.

numba/tests/test_interpreter.py
```diff
@@ -0,0 +1,641 @@
...
```
stuartarchibald, line 513: Remove? Same reasoning as above.

numba/tests/test_interpreter.py
```diff
@@ -0,0 +1,641 @@
...
+
+    @skip_unless_py10
+    def test_large_args_uninlined_controlflow(self):
```
stuartarchibald, line 399:
```suggestion
    def test_large_args_noninlined_controlflow(self):
```
(suggest same change throughout).

numba/tests/test_interpreter.py
```diff
@@ -0,0 +1,641 @@
...
+            'You can resolve this issue by moving the control flow out',
+            str(raises.exception)
+        )
+        # Uncomment sections when inlined control flow is
+        # actually supported.
+        # py_func = inline_func
+        # cfunc = njit()(inline_func)
+        # a = py_func(False)
+        # b = cfunc(False)
+        # self.assertEqual(a, b)
```
stuartarchibald, line 607: Remove?

stuartarchibald (commented):
I've re-reviewed this patch following the changes to catch problems associated with control flow at call sites. It now seems to be handling the cases outlined in the original ticket as expected and failing in a controlled manner else where. Given @sklam has run coverage on this patch I've not checked this further. There's a few minor things to look at in line and once resolved I think this should be ready to merge. Thanks again for your efforts on this @njriasan.

numba/core/interpreter.py
```diff
@@ -79,6 +79,649 @@ def get_assignment_source(self, destname):
...
+    return kws
+
+
+def _call_function_ex_replace_kws_large(
+    old_body, buildmap_name, buildmap_idx, search_end, new_body, func_ir, errmsg
+):
+    """
+    Extracts the kws args passed as varkwarg
+    for CALL_FUNCTION_EX. This pass is taken when
+    n_kws <= 15 and the bytecode looks like:
```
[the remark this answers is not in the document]
njriasan: This is an error in the docstring. Thank you for catching this.

numba/core/interpreter.py
```diff
@@ -79,6 +79,649 @@ def get_assignment_source(self, destname):
...
+
+        BUILD_MAP # Construct the map
+        # Start for each argument
+        LOAD_CONST # Load a constant for the name of the argument
+        LOAD_FAST  # Load each argument.
+        MAP_ADD # Append the (key, value) pair to the map
+        # End for each argument
+
+    In the IR generated, the initial build map is empty and a series
+    of setitems are applied afterwards. THE IR looks like:
```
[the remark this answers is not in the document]
njriasan: You are correct. I will update the description.

numba/core/interpreter.py
```diff
@@ -79,6 +79,649 @@ def get_assignment_source(self, destname):
...
+            and isinstance(const_stmt.value, ir.Const)
+        ):
+            # We cannot handle this format so raise the
+            # original error message.
+            raise UnsupportedError(errmsg)
+        key_var_name = const_stmt.target.name
+        key_val = const_stmt.value.value
+        search_start += 1
+        # Now we need to search for a getattr with setitem
+        not_found_getattr = True
```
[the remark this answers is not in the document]
njriasan: Agreed.

numba/core/interpreter.py
```diff
@@ -79,6 +79,649 @@ def get_assignment_source(self, destname):
...
+            )
+            and len(setitem_stmt.value.args) == 2
+            and (
+                setitem_stmt.value.args[0].name
+                == key_var_name
+            )
+        ):
+            # We cannot handle this format so raise the
+            # original error message.
+            raise UnsupportedError(errmsg)
```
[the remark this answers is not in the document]
njriasan: Actually for now I will just add the comment. If you would like me to change to an AssertionError I would be happy to, but I'm not certain this is actually unreachable.

numba/core/interpreter.py
```diff
@@ -79,6 +79,649 @@ def get_assignment_source(self, destname):
...
+    look like:
+
+        $orig_tuple = build_tuple(items=[])
+        $first_var = build_tuple(items=[Var(arg0, test.py:6)])
+        $next_tuple = $orig_tuple + $first_var
+        ...
+        $final_var = build_tuple(items=[Var(argn, test.py:6)])
+        $final_tuple = $prev_tuple + $final_var
+        $varargs_var = $final_tuple
+
```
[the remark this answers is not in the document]
njriasan: Yes this does. I will update the comment.

numba/core/interpreter.py
```diff
@@ -79,6 +79,649 @@ def get_assignment_source(self, destname):
...
+        if (
+            isinstance(concat_stmt, ir.Assign)
+            and concat_stmt.target.name == target_name
+            and isinstance(concat_stmt.value, ir.Expr)
+            and concat_stmt.value.op == "build_tuple"
+            and not concat_stmt.value.items
+        ):
+            # If we have reached the build_tuple
+            # we exit.
+            start_not_found = False
```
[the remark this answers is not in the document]
njriasan: Yes it is. It is also functioning as a break for the loop, but that can be replaced with a real break.

numba/tests/test_interpreter.py
njriasan: Done.

numba/tests/test_interpreter.py
njriasan: Done.

numba/tests/test_interpreter.py
njriasan: Done.

numba/core/interpreter.py
njriasan: Agreed. Thank you!

numba/core/interpreter.py
```diff
@@ -79,6 +79,649 @@ def get_assignment_source(self, destname):
...
+    """
+    This peephole rewrites a bytecode sequence unique to Python 3.10
+    where CALL_FUNCTION_EX is used instead of CALL_FUNCTION_KW because of
+    stack limitations set by CPython. This limitation is imposed whenever
+    a function call has too many arguments or keyword arguments.
+
+    https://github.com/python/cpython/blob/a58ebcc701dd6c43630df941481475ff0f615a81/Python/compile.c#L55
+    https://github.com/python/cpython/blob/a58ebcc701dd6c43630df941481475ff0f615a81/Python/compile.c#L4442
+
+    In particular, change is imposed whenever (n_args / 2) + n_kws > 15.
```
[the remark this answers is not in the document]
njriasan: Done.

numba/core/interpreter.py
```diff
@@ -79,6 +79,649 @@ def get_assignment_source(self, destname):
...
+        ):
+            # If we have reached the build_tuple
+            # we exit.
+            start_not_found = False
+            new_body[search_end] = None
+            # Remove the definition.
+            _remove_assignment_definition(old_body, search_end, func_ir)
+        else:
+            # We expect to find another arg to append.
+            # The first stmt must be an add
```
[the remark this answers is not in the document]
njriasan: Done.

numba/core/interpreter.py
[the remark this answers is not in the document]
njriasan: Yes.

numba/core/interpreter.py
```diff
@@ -79,6 +79,649 @@ def get_assignment_source(self, destname):
...
+    for CALL_FUNCTION_EX. This pass is taken when
+    n_kws <= 15 and the bytecode looks like:
+
+        # Start for each argument
+        LOAD_FAST  # Load each argument.
+        # End for each argument
+        ...
+        BUILD_CONST_KEY_MAP # Build a map
+
+    In the IR generated, the varkwarg refer
```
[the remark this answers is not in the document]
njriasan: Done.

numba/core/interpreter.py
```diff
@@ -79,6 +79,649 @@ def get_assignment_source(self, destname):
...
+
+        $orig_tuple = build_tuple(items=[])
+        $first_var = build_tuple(items=[Var(arg0, test.py:6)])
+        $next_tuple = $orig_tuple + $first_var
+        ...
+        $final_var = build_tuple(items=[Var(argn, test.py:6)])
+        $final_tuple = $prev_tuple + $final_var
+        $varargs_var = $final_tuple
+
+    It unclear if the extra assignment at the end is always present.
```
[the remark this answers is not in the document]
njriasan: Yes this included in the AssertionError, so I will remove this from the comments.

numba/core/interpreter.py
```diff
@@ -79,6 +79,649 @@ def get_assignment_source(self, destname):
...
+                and isinstance(stmt.value, ir.Expr)
+                and stmt.value.op == "call"
+                and stmt.value.varkwarg is not None
+            ):
+                blk_changed = True
+                call = stmt.value
+                args = call.args
+                kws = call.kws
+                # We search for replace if a call has either vararg
+                # or varkwarg.
```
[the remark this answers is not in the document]
njriasan: I have updated this with a longer comment explaining the meaning of this. The summary is that we investigate a call expression whenever it contains either vararg or varkwarg. varkwarg requires the bytecode changes and vararg is used for generating a nicer error message.

Please let me know if you think my more detailed comment is still unclear.

njriasan:
@stuartarchibald I made your requested changes. There are possibly two things that need to be looked at:

1. Should the "abundance of caution" check contain a different error like an unreachable assertion. I don't have enough confidence with the code structure to say its truly unreachable so I just added a comment.
2. Is the comment explaining when to check the call expression clear enough?

If both these changes seem fine, I think everything else should be a direct application of your feedback and we should be ready to merge. Thank you so much for the detailed review!

numba/core/interpreter.py
```diff
@@ -79,6 +80,646 @@ def get_assignment_source(self, destname):
...
+                isinstance(stmt, ir.Assign)
+                and isinstance(stmt.value, ir.Expr)
+                and stmt.value.op == "call"
+                and stmt.value.varkwarg is not None
+            ):
+                blk_changed = True
+                call = stmt.value
+                args = call.args
+                kws = call.kws
+                # We need to to check the call expression contents if
```
stuartarchibald, line 507:
```suggestion
                # We need to check the call expression contents if
```

numba/core/interpreter.py
```diff
@@ -79,6 +80,646 @@ def get_assignment_source(self, destname):
...
+                # it contains either vararg or varkwarg. If it contains
+                # varkwarg we need to update the actual bytecode. If it
+                # just contains vararg we don't need to update the bytecode,
```
stuartarchibald, line 510: Do you mean `bytecode` or IR?

stuartarchibald (commented):
Thanks for the fixes @njriasan, there's a typo and a query to resolve else this looks good.

stuartarchibald:
> @stuartarchibald I made your requested changes.

@njriasan Many thanks.

> There are possibly two things that need to be looked at:
> 
>     1. Should the "abundance of caution" check contain a different error like an unreachable assertion. I don't have enough confidence with the code structure to say its truly unreachable so I just added a comment.

I think it can be left as-is for now. Given it cannot be handled at present, if it turns out that there's use cases in practice that will trigger it, I would guess someone will make a request for it/report it. I don't think it warrants an exception containing the reportable condition text because it's already known that it might happen and it can't be handled.
 
>     2. Is the comment explaining when to check the call expression clear enough?

Couple of minor things I've noted in https://github.com/numba/numba/pull/7866#pullrequestreview-964652343 else looks good thanks.

> If both these changes seem fine, I think everything else should be a direct application of your feedback and we should be ready to merge. Thank you so much for the detailed review!

No problem, thanks for your efforts on this.

numba/core/interpreter.py
njriasan: Done.

njriasan:
Thank you @stuartarchibald. I made the requested changes.

stuartarchibald (approved):
Thanks for the patch and fixes @njriasan, great to get this merged!

PR #7871, opened 2022-03-01
accept kwargs for cuda kernels. cache kernel configurations

### LINKED ISSUE #7870: Add support of keyword argument to Dispatchers from `numba.cuda.jit`
---

## Feature request

Currently, the `numba.njit` decorated functions can support keyword arguments and can be invoked as regular python functions (where positional arguments can be interpreted as keywords and etc.)

However, the same is not possible with `numba.cuda.jit` results. This is likely due to the definition of `_KernelConfiguration` (now called [`_LaunchConfiguration`](https://github.com/numba/numba/blob/a38cfe537269f3bdc9c3401bceba588c7ae07c85/numba/cuda/dispatcher.py#L218-L240) that is returned as a result of the `__getitem__` call to a jitted cuda kernel. 

Not sure how this could be accomplished, would it be possible to wrap the original function with the `_LaunchConfiguration` instance that is returned from `__getitem__`?

```py
import numba
import numpy as np
from numba import cuda

@numba.njit
def func_cpu(a, b=0.):
    for i in range(a.shape[0]):
        a[i]+= b

@cuda.jit
def func_cuda(a, b=0.):
    for i in range(a.shape[0]):
        a[i]+= b


# these all work
a = np.random.randn(1000000)
_ = func_cpu(a)
func_cpu(a, b=1.)
func_cpu(a, 2.)

a_g = cuda.to_device(a)
# 1. This works
func_cuda[1,1](a_g, 2.) 

# 2. Raises TypeError:
#    TypeError: Failed in cuda mode pipeline (step: fix up args)
#    Signature mismatch: 1 argument types given, but function takes 2 arguments
func_cuda[1,1](a_g) 

# 3. Raise TypeError:
#    __call__() got an unexpected keyword argument 'b'
func_cuda[1,1](a_g, b=1.)
```

TK-21st:
This PR:
1. Supports kwargs for cuda-jitted kernel invocations by using `inspect.signature.bind`, which converts any `**kwargs` to `*args` before calling `Dispatcher.call()`.  As a result of this change, the call signature of `_LaunchConfiguration` and `Dispatcher.call` have been changed to include `**kwargs`.
2. Use `lru_cache` for `Dispatcher.configure` to avoid creating new `_LaunchConfiguration` instances at every `__getitem__`call.
For a simple kernel like
```py
@cuda.jit
def func_cuda(a, b=0.):
    for i in range(cuda.grid(1), a.shape[0], cuda.gridsize(1)):
        a[i]+= b
```
The invocation of `func_cuda[4,256](a_g, b=1.)` changes from 70 us to 40 us per loop with the lru_cache.

#### EDIT
`lru_cache` change is removed from the PR.

Closes #7870


numba/cuda/dispatcher.py
```diff
@@ -561,10 +573,16 @@ def __call__(self, *args, **kwargs):
...
-    def call(self, args, griddim, blockdim, stream, sharedmem):
+    def call(self, args, kwargs, griddim, blockdim, stream, sharedmem):
         '''
-        Compile if necessary and invoke this kernel with *args*.
+        Compile if necessary and invoke this kernel with *args, kwargs*.
+
+        *kwargs* are first converted to args using *self.py_func_signature*
         '''
+        bound_args = self.py_func_signature.bind(*args, **kwargs)
+        bound_args.apply_defaults()
```
TK-21st, line 591: The `apply_defaults` call here makes sure that the every argument (including kwargs) is populated for kernel invocation. However, this does not do any type coercion, and the types of the default kwargs that is specfied at kernel specification time will remain unchanged. This will lead to an error in `_Kernel.prepare_args` for specialized kernels, see [test case](https://github.com/numba/numba/blob/b62ce48e0d3b215eba9881432a15e410dac4135c/numba/cuda/tests/cudapy/test_dispatcher.py#L340-L347).

gmarkall:
Thanks for the PR - can you split this into two PRs please (e.g. by removing the caching from this PR and making another one that just does the caching)? It's much easier for the review process to consider each improvement separately.

TK-21st:
> Thanks for the PR - can you split this into two PRs please (e.g. by removing the caching from this PR and making another one that just does the caching)? It's much easier for the review process to consider each improvement separately.

Hey @gmarkall, just did and waiting for the checks. Not sure if this is what I should do or not but I basically just rolled back the change for caching. Will open another one for the caching.

gmarkall:
> Hey @gmarkall, just did and waiting for the checks. Not sure if this is what I should do or not but I basically just rolled back the change for caching. Will open another one for the caching.

That's great, many thanks!

gmarkall:
Now that the CUDA target shares a lot of the infrastructure with the core / CPU implementations, I'm not sure it's necessary to have special handling for default arguments in CUDA - my present hypothesis is that we ought to just be able to enable argument folding and then pass args and kwargs through the dispatcher machinery. I have a quick experiment with this idea here: https://github.com/gmarkall/numba/commit/1db1c44be1a541e73269ba12131762891d8410f5 (branch is https://github.com/numba/numba/compare/main...gmarkall:cuda-kwargs)

In general I'd prefer the approach of reusing the core/CPU machinery for this - are there test cases for which the implementation in the commit above fails, whereas the implementation in this PR succeeds?

Note that my understanding of argument folding is as follows:

> If you are compiling a function where not all the kwargs are specified, the dispatcher compiles a variant where the argument list consists only of positional args and specified kwargs. So for example:
> - You have an njitted function `def f(a, b, c=1, d=2, e=3)`
> - You call `f(1, 2, d=13)`
> - A version `f(a, b, d)` is compiled with `c=1` and `e=3` baked in
> - The call executes `f(1, 2, 13)`

Can @sklam or @stuartarchibald comment on this description and the idea above?

TK-21st:
> Now that the CUDA target shares a lot of the infrastructure with the core / CPU implementations, I'm not sure it's necessary to have special handling for default arguments in CUDA - my present hypothesis is that we ought to just be able to enable argument folding and then pass args and kwargs through the dispatcher machinery. I have a quick experiment with this idea here: [gmarkall@1db1c44](https://github.com/gmarkall/numba/commit/1db1c44be1a541e73269ba12131762891d8410f5) (branch is [main...gmarkall:cuda-kwargs](https://github.com/numba/numba/compare/main...gmarkall:cuda-kwargs))
> 
> In general I'd prefer the approach of reusing the core/CPU machinery for this - are there test cases for which the implementation in the commit above fails, whereas the implementation in this PR succeeds?
> 
> Note that my understanding of argument folding is as follows:
> 
> > If you are compiling a function where not all the kwargs are specified, the dispatcher compiles a variant where the argument list consists only of positional args and specified kwargs. So for example:
> > 
> > * You have an njitted function `def f(a, b, c=1, d=2, e=3)`
> > * You call `f(1, 2, d=13)`
> > * A version `f(a, b, d)` is compiled with `c=1` and `e=3` baked in
> > * The call executes `f(1, 2, 13)`
> 
> Can @sklam or @stuartarchibald comment on this description and the idea above?


I'm not familiar with how the dispatchers are setup, if there's a more native way to do this that'd be great!

I just gave your approach a quick try and a test case with kwargs added in the wrong order will fail because of argtypes mismatch with arguments:

```py

def add_multi(x, y, z):
    return x + y + z

def add_kernel_with_multiple_default(r, x, y=789., z=1):
    r[0] = x + y + z

...

def test_kernel_with_defaults(self):
        c_add = cuda.jit(add_kernel)
        c_add_w_default = cuda.jit(add_kernel_with_default)
        c_add_w_multi_default = cuda.jit(add_kernel_with_multiple_default)

        # Using a complex128 allows us to represent any result produced by the
        # test
        r = np.zeros(1, dtype=np.complex128)

        for x,y,z in zip([123, 12.3, 12.3, 12300000000], [456, 45.6, 45.6j, 456], [1, 2, 3, 4]):
            c_add[1, 1](r, x, y)
            self.assertEqual(r[0], add(x, y))

            # invocation with keyword
            c_add_w_default[1, 1](r, x, y=y)
            self.assertEqual(r[0], add(x, y))

            # invocation as position
            c_add_w_default[1, 1](r, x, y)
            self.assertEqual(r[0], add(x, y))

            # invocation using default, y=789
            c_add_w_default[1, 1](r, x)
            self.assertEqual(r[0], add(x, 789))

            # multi defaults
            c_add_w_multi_default[1, 1](r, x, y, z)
            self.assertEqual(r[0], add_multi(x, y, z))

            c_add_w_multi_default[1, 1](r, x, y, z=z)
            self.assertEqual(r[0], add_multi(x, y, z))

            c_add_w_multi_default[1, 1](r, x, y=y, z=z)
            self.assertEqual(r[0], add_multi(x, y, z))

            c_add_w_multi_default[1, 1](r, x=x, y=y, z=z)
            self.assertEqual(r[0], add_multi(x, y, z))
           
            c_add_w_multi_default[1, 1](r, y=y, z=z, x=x)     #  <------ this fails
            self.assertEqual(r[0], add_multi(x, y, z))

```

I suspect the lines [here](https://github.com/gmarkall/numba/blob/1db1c44be1a541e73269ba12131762891d8410f5/numba/cuda/dispatcher.py#L255-L259) is the issue.

gmarkall:
Just adding a quick note, will respond to your earlier comments later - I noticed we'll also need to do something for `float16` and kwargs as mentioned in https://github.com/numba/numba/pull/7972#pullrequestreview-938324170

stuartarchibald:
@gmarkall RE: https://github.com/numba/numba/pull/7871#issuecomment-1093074823 I think the description regarding argument folding is about right, the compiled function will have the missing args baked in and the function signature appropriately reduced. RE: the reuse of core machinery, I can't think of anything special about CUDA in this case that would warrant a custom code path for handling kwargs. Therefore, if possible, I think the application/use of existing the functionality from the CPU target is the best thing in terms of maintenance and consistency.

gmarkall:
> if possible, I think the application/use of existing the functionality from the CPU target is the best thing in terms of maintenance and consistency.

I didn't yet finish understanding how the CPU target does this, but it seems that it uses the [`ArgPacker`](https://github.com/numba/numba/blob/main/numba/core/datamodel/packer.py#L60) in the data model, as part of the generated function. I was going to try and understand this to suggest that @TK-21st duplicate the logic in the CUDA target for now, with a view to it using the same code generation as the CPU when call sides are jitted.

gmarkall:
I think the CPU logic begins here: https://github.com/numba/numba/blob/main/numba/core/callconv.py#L547, but there's a few layers of abstraction to go through to understand what's happening - the builder API, the argpacker, and whatever it does to create the list of `realargs`.

gmarkall:
@TK-21st Since the kernel caching part of this went in a separate PR, and I don't think we can accept the support for kwargs in this form (that works a different way to the CPU target), could we close this PR? Many thanks for your efforts so far and thankyou for contributing to Numba.

TK-21st:
Sure! Thank you Graham for following up. On Mar 15, 2023, at 12:47 AM, Graham Markall ***@***.***> wrote:﻿
@TK-21st Since the kernel caching part of this went in a separate PR, and I don't think we can accept the support for kwargs in this form (that works a different way to the CPU target), could we close this PR? Many thanks for your efforts so far and thankyou for contributing to Numba.

PR #6577, opened 2020-12-14
Linear probe in sets only when appropriate.

stuartarchibald:
As title.

Fixes #6543

PR #7875, opened 2022-03-01
Merged changes from fix/6543 and added comments

benwilliamgraham:
Continuation of https://github.com/numba/numba/pull/6577 (fix for https://github.com/numba/numba/issues/6543)


esc:
@benwilliamgraham thank you for submitting this. I see only a single commit in this PR. If this is a continuation of #6577 I would have expected at least one commit from @stuartarchibald ? Did you perhaps use `git commit --amend`? For the code, is it simply the code from #6577 with additional source code comments?

benwilliamgraham:
> For the code, is it simply the code from https://github.com/numba/numba/pull/6577 with additional source code comments?

Yeah, it's the original code along with the comment and the update from https://github.com/numba/numba/pull/6577#discussion_r542713967

esc:
> > For the code, is it simply the code from #6577 with additional source code comments?
> 
> Yeah, it's the original code along with the comment and the update from [#6577 (comment)](https://github.com/numba/numba/pull/6577#discussion_r542713967)

OK; thank you! Can you redo this PR and split it into two commits please:

a) The original commit from @stuartarchibald 
b) The additional commit with the changes you made

Otherwise we can not review this since the attribution of @stuartarchibald for that work/commit would be lost. Thank you!

stuartarchibald:
@esc do you recall if patch 735f10d actually fixed the problem or just moved it somewhere else? I've a vague recollection of it only being a partial fix.

esc:
> @esc do you recall if patch [735f10d](https://github.com/numba/numba/commit/735f10d3414685013b4aa05f32771551c61c6fbf) actually fixed the problem or just moved it somewhere else? I've a vague recollection of it only being a partial fix.

This is my recollection also. I would tend to argue, this PR will also need at least my patch from:

https://github.com/numba/numba/issues/6543#issuecomment-740726668

Or a variant of that. And also, all the additional test-cases listed in #6543 to really make sure the fix is solid.

stuartarchibald:
> > @esc do you recall if patch [735f10d](https://github.com/numba/numba/commit/735f10d3414685013b4aa05f32771551c61c6fbf) actually fixed the problem or just moved it somewhere else? I've a vague recollection of it only being a partial fix.
> 
> This is my recollection also. I would tend to argue, this PR will also need at least my patch from:
> 
> [#6543 (comment)](https://github.com/numba/numba/issues/6543#issuecomment-740726668)
> 

Ah good, glad we have the same recollection of this, and now it's recorded in comments it is less likely to be forgotten!

> Or a variant of that. And also, all the additional test-cases listed in #6543 to really make sure the fix is solid.

I think that until the root causes are discovered and explanations offered for them, that it'd be very hard to review a patch. Without this, a "fix" could just be moving the problem elsewhere.

esc:
> I think that until the root causes are discovered and explanations offered for them, that it'd be very hard to review a patch. Without this, a "fix" could just be moving the problem elsewhere.

@stuartarchibald I concur. For example, now that the tests have been added, I see that the linear probing tests hangs on linux on public CI.

esc:
> I think that until the root causes are discovered and explanations offered for them, that it'd be very hard to review a patch. Without this, a "fix" could just be moving the problem elsewhere.

@stuartarchibald I concur. For example, now that the tests have been added, I see that the linear probing tests hangs on linux on public CI.

@benwilliamgraham it's probably worth considering my patch from here: https://github.com/numba/numba/issues/6543#issuecomment-740726668 which may help.

benwilliamgraham:
I added the patch but it seems that it causes some memory to not be freed. Also, is there a good way to attribute the patch to you in git?

esc:
> I added the patch but it seems that it causes some memory to not be freed. Also, is there a good way to attribute the patch to you in git?

I think using my github handle in the commit message is fine. I think you might need to take a closer look here. IIRC this bug was quite tricky, maybe you get more lucky than we did so far.

esc:
@benwilliamgraham we will wait on you. Please do let us know if you make progress on this and @stuartarchibald and I will be happy to review! Thank you so much for looking into this.

PR #7877, opened 2022-03-01
Fix invalid resolution with multiple ConcreteTemplate

### LINKED ISSUE #7863: Multiple ConcreteTemplates should not be considered separately
Functions with multiple ConcreteTemplates must be combined and considered at the same time during function resolution.

The problem is discovered by @testhound in https://github.com/testhound/numba/commit/5188b287d9d7384c388846085ea8416bd7f210a4.

Reproducer:

```python
import numpy as np
from numba import cuda

@cuda.jit
def kernel(arr):
    v = arr[0]
    exp = -v                 # fp16 negate is being considered first; making exp a float64
    arr[0] = exp & 1    # operator.and_ does not support float64

arr = np.zeros(1, dtype=np.intp)
kernel[1, 1](arr)
```

sklam:
Fixes #7863.

### DIFF
--- a/numba/core/types/functions.py
+++ b/numba/core/types/functions.py
@@ -293,6 +293,21 @@ class BaseFunction(Callable):
 
         self._depth += 1
 
+        from numba.core.typing.templates import ConcreteTemplate
+        concretes = []
+        abstracts = []
+        for temp_cls in order:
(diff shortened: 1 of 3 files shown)


sklam:
Reviving for #8010. Relates to https://github.com/numba/numba/pull/8010#issuecomment-1112345537

PR #7884, opened 2022-03-04
Implement getattr builtin.

stuartarchibald:
As title.


numba/cpython/builtins.py
```diff
@@ -777,3 +777,93 @@ def false_impl(var, typs):
...
+# getattr with default arg present, obj is an open type, name is forced as a
+# literal string, the "default" is again an open type. Note that the CPython
+# definition is: `getattr(object, name[, default]) -> value`, the `default`
+# is not a kwarg.
+@overload(getattr)
+def ol_getattr(obj, name, default):
+    def impl(obj, name, default):
+        try:
+            return resolve_getattr(obj, name)
+        except Exception:
```
stuartarchibald, line 866: This is not strictly correct, it should only accept `AttributeError`, but `try-except` support cannot handle this. Given the only place an exception can be raised is from the intrinsic `resolve_getattr` and Numba controls this I'm reasonably sure that this is safe.

sklam:
I got a test failure when I run `python runtests.py numba/tests/test_builtins.py`

```
FAIL: test_experimental_warning (numba.tests.test_builtins.TestIsinstanceBuiltin)
```

sklam:
Nevermind, the problem exist on `main`

numba/cpython/builtins.py
sklam: Worth leaving this note into the code as comment

sklam (commented):
Looks good. 

I'll do a second pass review to try breaking it.

stuartarchibald:
> Nevermind, the problem exist on `main`

What was the issue and is there a ticket tracking it?

sklam (commented):
Also, the 3-arg version of `getattr` has unexpected typing for the return type. (This is very likely a problem beyond this PR):

```python
from numba import njit
import numpy as np


@njit
def foo(obj, name):
    return getattr(obj, name)


@njit
def fake_hash():
    return 12345

@njit
def bar():
    ndim = foo(np.ones(1), "ndim")
    hasher = foo("a_string", "__hash__")
    print(ndim, hasher())
    array_hash = getattr(np.ones(1), "__hash__", fake_hash)  # this is ok
    print(array_hash())                                      # optional type?

    # No type info available for OptionalType(type(CPUDispatcher(<function fake_hash at 0x7fe36843e430>))) as a callable.
    # During: resolving callee type: OptionalType(type(CPUDispatcher(<function fake_hash at 0x7fe36843e430>)))
    #
    # File "chk.py", line 20:
    # def bar():
    #     <source elided>
    #     array_hash = getattr(np.ones(1), "__hash__", fake_hash)
    #     print(array_hash())
    #     ^

bar()
```

stuartarchibald:
> Also, the 3-arg version of getattr has unexpected typing for the return type. (This is very likely a problem beyond this PR):

I've got a fix for this which involves making the intrinsic implementation always take 3 args, and then in the "no default" case a specific dummy `namedtuple` is used as a marker to signal that no default has been supplied. I think this preserves the semantics and avoids the `OptionalType` problem. Does that sound acceptable?

numba/cpython/builtins.py
stuartarchibald: I removed the exception handling impl in 5886a8c as it was causing the `OptionalType` problem noted below.

docs/source/reference/pysupported.rst
```diff
@@ -930,6 +930,7 @@ The following built-in functions are supported:
 * :func:`enumerate`
 * :func:`filter`
 * :class:`float`
+* :func:`getattr`
```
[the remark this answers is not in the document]
stuartarchibald: Both the literal string requirement and the lack of support for Function type are noted in the docs as part of 5886a8c.

stuartarchibald:
Conflicts resolved in 5151bbd.

sklam:
https://github.com/numba/numba/pull/7884/commits/5886a8c7e056ec3d6a6d60e6c801113149d105fc is smart

sklam (approved):
Just one more merge conflict to resolve.
Otherwise, everything looks good.

stuartarchibald:
> Just one more merge conflict to resolve. Otherwise, everything looks good.

Thanks for the review, conflict should be resolved in 3d9981d.

sklam (approved):

PR #7892, opened 2022-03-07
Add support for default rcond=None in np.linalg.lstsq

### LINKED ISSUE #7790: np1.14+ np.linalg.lstsq has `rcond="warn"` as default and a "new" proposed default of None
## Reporting a bug

- [N/A] I have included a self contained code sample to reproduce the problem.

NumPy 1.14+ function `np.linalg.lstsq` has `rcond="warn"` as default and a "new" proposed default of `None` see:
https://github.com/numpy/numpy/blob/maintenance/1.21.x/numpy/linalg/linalg.py#L2170 
and
https://github.com/numpy/numpy/blob/maintenance/1.21.x/numpy/linalg/linalg.py#L2282-L2294 

whereas the Numba impl has `rcond` as `-1.0` which is the default prior to np 1.14.
https://github.com/numba/numba/blob/324dd8f14aa83404ab6d0e51c5e25665dd89ed8d/numba/np/linalg.py#L1583

stuartarchibald:
As title.

Fixes #7790

### DIFF
--- a/numba/np/linalg.py
+++ b/numba/np/linalg.py
@@ -1579,8 +1579,34 @@ def _lstsq_solution_impl(b, bcpy, n):
         return twoD_impl
 
 
+def _lstsq_resolve_rcond(rcond, dtype, m, n):
+    pass
+
+
+@overload(_lstsq_resolve_rcond, prefer_literal=True)
+def ol_lstsq_resolve_rcond(rcond, dtype, m, n):
+    if isinstance(rcond, types.StringLiteral):
+        if rcond.literal_value == "warn": # it's the default
+            def impl(rcond, dtype, m, n):
+                return np.float64(-1.0)
+            return impl
+        else:
+            # This is an error, user has set the rcond as a string
+            raise TypingError("Unsupported literal string found in rcond")
+    elif np_support.is_nonelike(rcond):
+        def impl(rcond, dtype, m, n):
+            return np.float64(np.finfo(dtype).eps * max(m, n))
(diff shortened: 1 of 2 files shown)


numba/np/linalg.py
```diff
@@ -1579,8 +1579,34 @@ def twoD_impl(b, bcpy, n):
...
+    pass
+
+
+@overload(_lstsq_resolve_rcond, prefer_literal=True)
+def ol_lstsq_resolve_rcond(rcond, dtype, m, n):
+    if isinstance(rcond, types.StringLiteral):
+        if rcond.literal_value == "warn": # it's the default
+            def impl(rcond, dtype, m, n):
+                return np.float64(-1.0)
+            return impl
```
sklam, line 1592: There are no tests for this.
And, should it raise a warning?

stuartarchibald: Thanks, I've now written a test. I'm not sure about Numba re-implementing NumPy's warnings, the same question occurs for NumPy 1.22 in `np.MachAr` which is deprecated. I think the conclusion from an OOB conversation was to replicate the warnings from the typing layer such that users are warned about changes to NumPy behaviour/API even though the NumPy implementation itself is not in use. I'll have a go at implementing that here.

stuartarchibald: Implemented in bfad183

esc:
@stuartarchibald seems like this one went stale. Can you merge in `main` to resolve the conflicts?

PR #7893, opened 2022-03-08
Use uuid.uuid4() as the key in serialization.

### LINKED ISSUE #7881: objmode() triggers fork()
## Reporting a bug

It seems that using [`with numba.objmode()`](https://numba.readthedocs.io/en/stable/user/withobjmode.html) is triggering a call to `fork()` somewhere. In an MPI context, this can be problematic – Open MPI is complaining about this with a message of the following form:

>     --------------------------------------------------------------------------
>     A process has executed an operation involving a call to the
>     "fork()" system call to create a child process.  Open MPI is currently
>     operating in a condition that could result in memory corruption or
>     other system errors; your job may hang, crash, or produce silent
>     data corruption.  The use of fork() (or system() or other calls that
>     create child processes) is strongly discouraged.
>     
>     The process that invoked fork was:
>     
>       Local host:          [[53156,1],2] (PID 26851)
>     
>     If you are *absolutely sure* that your application will successfully
>     and correctly survive a call to fork(), you may disable this warning
>     by setting the mpi_warn_on_fork MCA parameter to 0.
>     --------------------------------------------------------------------------

Using the same code, but omitting `with numba.objmode()`, Open MPI doesn’t complain. As far as I can remember, this issue only started to appear recently, so there seems to have been a change in the implementation causing this? If there is some way to stop `objmode()` from triggering a `fork()`, that would be great.

<details>
  <summary>(Trivial) code sample</summary>

```
import numba
from mpi4py import MPI

mpi_comm = MPI.COMM_WORLD
mpi_rank = mpi_comm.Get_rank()

@numba.njit
def f():
    with numba.objmode():
        print(mpi_rank)

f()
```
</details>

<details>
  <summary>Running the sample and output</summary>

>     $ mpiexec -n 3 python3 test.py
>     --------------------------------------------------------------------------
>     A process has executed an operation involving a call to the
>     "fork()" system call to create a child process.  Open MPI is currently
>     operating in a condition that could result in memory corruption or
>     other system errors; your job may hang, crash, or produce silent
>     data corruption.  The use of fork() (or system() or other calls that
>     create child processes) is strongly discouraged.
>     
>     The process that invoked fork was:
>     
>       Local host:          [[53156,1],2] (PID 26851)
>     
>     If you are *absolutely sure* that your application will successfully
>     and correctly survive a call to fork(), you may disable this warning
>     by setting the mpi_warn_on_fork MCA parameter to 0.
>     --------------------------------------------------------------------------
>     1
>     2
>     0
>     [host:26836] 2 more processes have sent help message help-opal-runtime.txt / opal_init:warn-fork
>     [host:26836] Set MCA parameter "orte_base_help_aggregate" to 0 to see all help / error messages
</details>

stuartarchibald:
As title, this is to avoid fork()+exec() calls made by uuid1().

Fixes #7881

### DIFF
--- a/numba/tests/support.py
+++ b/numba/tests/support.py
@@ -1010,7 +1010,7 @@ def run_in_subprocess(code, flags=None, env=None, timeout=30):
 
 
 def strace(work, syscalls, timeout=10):
-    """ Runs strace whilst executing the function work() in the current process,
+    """Runs strace whilst executing the function work() in the current process,
     captures the listed syscalls (list of strings). Takes an optional timeout in
     seconds, default is 10, if this is exceeded the process will be sent a
     SIGKILL. Returns a list of lines that are output by strace.
@@ -1064,12 +1064,6 @@ def strace(work, syscalls, timeout=10):
             check_return()
             # collect the data
             strace_data = ntf.readlines()
-            # stop the process
-            popen.terminate()
-            if popen.poll() is None:
-                # Something has gone wrong, force stop the process
-                popen.kill()
-                raise RuntimeError("strace did not terminate.")
         finally:
             # join communication, should be stopped now as process has
             # exited
@@ -1085,14 +1079,12 @@ def _strace_supported():
     def force_clone(): # subprocess triggers a clone
         subprocess.run([sys.executable, '-c', 'exit()'])
 
+    syscall = 'clone'
     try:
-        syscall = 'clone'
         trace = strace(force_clone, [syscall,])
-        assert syscall in ''.join(trace)
     except Exception:
         return False
-    else:
-        return True
+    return syscall in ''.join(trace)
 
 
 _HAVE_STRACE = _strace_supported()
--- a/numba/tests/support.py
+++ b/numba/tests/support.py
@@ -9,10 +9,12 @@ import gc
 import math
 import platform
 import os
+import signal
 import shutil
 import subprocess
 import sys
 import tempfile
+import threading
 import time
 import io
 import ctypes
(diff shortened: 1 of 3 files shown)


numba/tests/test_withlifting.py
```diff
@@ -1213,5 +1220,30 @@ def foo():
...
 
 
+class TestMisc(TestCase):
+    # Tests for miscellaneous objmode issues. Run serially.
+
+    _numba_parallel_test_ = False
+
+    @linux_only
+    @needs_strace
+    def test_no_fork_in_compilation(self):
```
stuartarchibald, line 1230: Note to self. This test probably needs to be run as a subprocess test on the basis that previous execution of something which triggers a stateful call e.g. `uuid.uuid1.getnode()` is pretty likely.

stuartarchibald: Addressed in d114552

numba/tests/support.py
```diff
@@ -1007,6 +1009,98 @@ def run_in_subprocess(code, flags=None, env=None, timeout=30):
...
+                check_return(problem)
+            # Make sure the return code is 0, SIGINT to detach is considered
+            # a successful exit.
+            popen.wait()
+            check_return()
+            # collect the data
+            strace_data = ntf.readlines()
+            # stop the process
+            popen.terminate()
+            if popen.poll() is None:
```
sklam, line 1069: Would there ever be a delay in `terminate()` that `poll()` will see the process not terminated but actually in the process of doing so? Should this be a `wait()`?

numba/tests/support.py
```diff
@@ -1007,6 +1009,98 @@ def run_in_subprocess(code, flags=None, env=None, timeout=30):
...
+
+def _strace_supported():
+    """Checks if strace is supported and working"""
+    def force_clone(): # subprocess triggers a clone
+        subprocess.run([sys.executable, '-c', 'exit()'])
+
+    try:
+        syscall = 'clone'
+        trace = strace(force_clone, [syscall,])
+        assert syscall in ''.join(trace)
```
sklam, line 1091: not why just `return`?

numba/tests/support.py
```diff
@@ -1007,6 +1009,98 @@ def run_in_subprocess(code, flags=None, env=None, timeout=30):
...
+
+
+def _strace_supported():
+    """Checks if strace is supported and working"""
+    def force_clone(): # subprocess triggers a clone
+        subprocess.run([sys.executable, '-c', 'exit()'])
+
+    try:
+        syscall = 'clone'
+        trace = strace(force_clone, [syscall,])
```
sklam, line 1090: try..except should only be applied to this line

sklam (commented):
Thanks for the patch! `uuid4` should be sufficient given its use in cloudpickle and its implementation should never depend on fork (hopefully).

There are a few comments esp. in the strace code regarding the handling of the subprocess

numba/tests/support.py
sklam: right, i missed the earlier `wait()`.  and yes for the removal

numba/tests/support.py
stuartarchibald: Ah I see, yes, thanks, will fix.

sklam (approved):
Thanks for the patch!

PR #7898, opened 2022-03-10
Skip test_ptds under cuda-memcheck

gmarkall:
This is another test that hangs under cuda-memcheck, so a skip is added for it.


gmarkall:
gpuci run tests

stuartarchibald:
From OOB conversation with @gmarkall, as I could not reproduce the process hang locally, this is probably hardware/driver specific. Further the test being skipped doesn't have anything to do with out of bounds access so it's probably fine to skip this under memcheck as the convenience of being able to run the entire suite under memcheck outweighs the risk involved with this skipping this PTDS test.

stuartarchibald (approved):
Thanks for the patch!

PR #7901, opened 2022-03-10
Pyston compatibility for the test suite

kmod:
numba already works on Pyston, but there were two small snags in the test suite that caused the tests to not pass.

### DIFF
--- a/numba/tests/test_nrt.py
+++ b/numba/tests/test_nrt.py
@@ -226,7 +226,10 @@ class TestTracemalloc(unittest.TestCase):
     """
 
     def measure_memory_diff(self, func):
-        import tracemalloc
+        try:
+            import tracemalloc
+        except ImportError:
+            self.skipTest("tracemalloc not available")
         tracemalloc.start()
         try:
             before = tracemalloc.take_snapshot()
--- a/numba/tests/test_pycc.py
+++ b/numba/tests/test_pycc.py
@@ -210,7 +210,7 @@ class TestCC(BasePYCCTest):
         self.assertTrue(os.path.basename(f).startswith('pycc_test_simple.'), f)
         if sys.platform.startswith('linux'):
             self.assertTrue(f.endswith('.so'), f)
-            self.assertIn('.cpython', f)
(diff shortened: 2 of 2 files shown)


gmarkall:
This is exciting, I didn't even know Numba worked on Pyston! Slightly tangential, but how easy is it to add Pyston to the Azure CI tests?

stuartarchibald:
> This is exciting, I didn't even know Numba worked on Pyston! Slightly tangential, but how easy is it to add Pyston to the Azure CI tests?

Indeed, great to see Numba works with Pyston.

RE Azure. I think it'd require a `llvmlite` development conda package that works under Pyston adding to the `numba` channel somewhere. Other than that I think it's just selectively adding in some conda channels and probably a e.g. `$PYTHON_IMPL` identifier to the build scripts/Azure config.

stuartarchibald (approved):
I tested this manually on Pyston with/without this patch. This patch fixes the noted failing tests. Many thanks.

stuartarchibald:
Notes:
* Initially CUDA tests were failing on Pyston, segfault very early on in library loading, I then installed SciPy from `conda-forge` and it seemed to "fix it" and all the CUDA tests passed, even though the SciPy build itself was broken, no idea why as yet (imagine it's to do with changing some underlying libraries to get SciPy to install)!
* Seems like there's no suitable SciPy build for Pyston so could not run the `numba.tests.test_linalg` tests which use SciPy's BLAS/LAPACK bindings.

seibert:
One note (mentioned OOB, but putting here for posterity) is that current Pyston conda packages are built in the conda-forge style, using the python_abi method for managing alternative interpreters.  Anaconda's packages don't yet do this, so testing with Pyston should be done in an environment built this way to avoid contamination by packages with an incompatible ABI:
```
conda create -n pyston_env  # create an empty environment
conda activate pyston_env
echo 'channels: [pyston,conda-forge]' | conda config --env --stdin
conda install pyston # add whatever other packages you want
```
(reference: https://github.com/pyston/pyston/wiki/Using-conda)

Issue #7886, opened 2022-03-05
Numba implementation of `numpy.random.vonmises` does not match NumPy


brandonwillard:
It looks like the Numba (version 0.55.1) implementation of `numpy.random.vonmises` is using a different parameterization than NumPy (version 1.21.5):
```python
import numpy as np
import numba as nb


@nb.njit
def vonmises(mu, kappa, size):
    np.random.seed(232)

    res = np.empty(size, dtype=np.float64)
    for n in range(size):
        res[n] = np.random.vonmises(mu, kappa)

    return res


numba_res = vonmises(0.0, 1.0, 1000)
numba_res.mean()
# 3.032362867944328

np.random.seed(232)
exp_res = np.random.vonmises(0.0, 1.0, size=1000)
exp_res.mean()
# 0.047849847034024555
```

If Numba's implementation follows `random.vonmisesvariate`, then the range probably needs to be shifted from `[0, 2 * pi]` to `[-pi, pi]` in order to match NumPy.

stuartarchibald:
Thanks for the report @brandonwillard, think this is down to algorithmic differences, the comments in the implementation in Numba hint at this: https://github.com/numba/numba/blob/59fa2e335be68148b3bd72a29de3ff011430038d/numba/cpython/randomimpl.py#L776-L778 and having looked at the NumPy implementation, they are considerably different. I think the NumPy implementation will need translating explicitly into Numba's `@overload` of this function to get a match.

brandonwillard:
> and having looked at the NumPy implementation, they are considerably different. I think the NumPy implementation will need translating explicitly into Numba's `@overload` of this function to get a match.

I had a general question about this in `numba-dev`:
> what's the consensus on making the `numpy.random` implementations match NumPy's exactly?
> is there a previous thread on this that someone can point me toward?

stuartarchibald:
> > and having looked at the NumPy implementation, they are considerably different. I think the NumPy implementation will need translating explicitly into Numba's `@overload` of this function to get a match.
> 
> I had a general question about this in `numba-dev`:
> 
> > what's the consensus on making the `numpy.random` implementations match NumPy's exactly?
> > is there a previous thread on this that someone can point me toward?
> 
> and [in the Discourse](https://numba.discourse.group/t/random-sampling-parity-with-numpy/1226).

[...]

Issue #7856, opened 2022-02-22
Some random distributions do not accept size=None arguments


rishi-kulkarni:
From a discussion in #7690 

Some of numba's random distribution functions cannot handle size=None arguments, despite size=None being the default value for all of these functions. 

```python
import numpy as np
import numba as nb

# this works
@nb.njit
def nb_chi2(df):
    return np.random.chisquare(df)

nb_chi2(2)

# this also works
np.random.chisquare(2)

# this does not work
@nb.njit
def nb_chi2(df, size):
    return np.random.chisquare(df, size)

nb_chi2(2, None)

# while this does
np.random.chisquare(2, None)
```
This means you have to write something like this everywhere if you want to use np.random.chisquare in a jitted function, though you'd hope that numba's dispatch mechanism would be handling this for you. 

```python
if size is None:
    return np.random.chisquare(df)
else:
    return np.random.chisquare(df, size)
```
Trying the above nonfunctioning function gives this error:
```
              Traceback (most recent call last):
                File "/Users/vhaenel/git/numba/numba/core/types/functions.py", line 308, in get_call_type
                  sig = temp.apply(nolitargs, nolitkws)
                File "/Users/vhaenel/git/numba/numba/core/typing/templates.py", line 351, in apply
                  sig = generic(args, kws)
                File "/Users/vhaenel/git/numba/numba/core/typing/templates.py", line 972, in generic
                  result = self._definition_func(self.context, *args, **kws)
                File "/Users/vhaenel/git/numba/numba/core/overload_glue.py", line 37, in stub
                  return body(tyctx)  # noqa: F821
                File "/Users/vhaenel/git/numba/numba/core/overload_glue.py", line 140, in body
                  sig = typing.apply(ol_args, ol_kwargs)
                File "/Users/vhaenel/git/numba/numba/core/typing/templates.py", line 414, in apply
                  sig = typer(*args, **kws)
                File "/Users/vhaenel/git/numba/numba/core/typing/randomdecl.py", line 277, in typer
                  return self.array_typer(size)(a)
                File "/Users/vhaenel/git/numba/numba/core/typing/randomdecl.py", line 63, in typer
                  shape, ndim = normalize_shape(size)
                File "/Users/vhaenel/git/numba/numba/core/typing/randomdecl.py", line 37, in normalize_shape
                  ndim = len(shape)
              TypeError: object of type 'NoneType' has no len()
```

@stuartarchibald points out that the issue is likely in here.

[...]

PR #7902, opened 2022-03-11
Update np.random.random_sample and aliases to handle size=None

rishi-kulkarni:
This PR relates to #7856 and (barely) to #7886.

Currently, numba's implementations of np.random functions can handle `size` arguments if they are integers or tuples of integers, or being given no size arguments at all. However, they throw the following error when given `size=None` explicitly:

```python
No implementation of function Function(<built-in method random_sample of numpy.random.mtrand.RandomState object at 0x7fc070323040>) found for signature:
 
 >>> random_sample(none)
 
There are 4 candidate implementations:
    - Of which 4 did not match due to:
    Overload in function '_OverloadWrapper._build.<locals>.ol_generated': File: numba/core/overload_glue.py: Line 131.
      With argument(s): '(none)':
     No implementation of function Function(<intrinsic stub>) found for signature:
      
      >>> stub(none)
      
       Intrinsic in function 'stub': File: numba/core/overload_glue.py: Line 35.
         With argument(s): '(none)':
          TypeError: object of type 'NoneType' has no len()
       raised from /home/rishik/Projects/numba/numba/numba/core/typing/randomdecl.py:37
     
     During: resolving callee type: Function(<intrinsic stub>)
     During: typing of call at <string> (3)
     
     
     File "<string>", line 3:

  raised from /home/rishik/Projects/numba/numba/numba/core/typeinfer.py:1086

During: resolving callee type: Function(<built-in method random_sample of numpy.random.mtrand.RandomState object at 0x7fc070323040>)
During: typing of call at <ipython-input-12-be755df48a33> (3)


File "<ipython-input-12-be755df48a33>", line 3:
def nb_rand_sample(size):
    return np.random.random_sample(size)
```

After discussions with @stuartarchibald  and @esc in #7886 and #7690, I decided that the softest touch here would be to just add an `@overload` for `np.random.random_sample` and its aliases that handles the `size=None` case. This should make it easier to make the other random distributions handle `size=None` arguments in the near future. 

On this branch, the following code snippet works:

```python
import numba as nb
import numpy as np
from numpy.testing import assert_almost_equal

@nb.njit
def set_nb_seed(seed):
    np.random.seed(seed)

@nb.njit
def nb_random_sample(size):
    return np.random.random_sample(size)



np.random.seed(1)
set_nb_seed(1)

for size in (None, 1, (5,), (5, 5, 5)):

    assert_almost_equal(nb_random_sample(size), np.random.random_sample(size))
```
I have added tests explicitly testing size=None random number generation, as well.


rishi-kulkarni:
I admit I'm puzzled re: the failing tests, as these same test cases pass on my machine:

```python
...........................................................................................
Ran 104 tests in 45.939s

OK
```

Any thoughts?

rishi-kulkarni:
https://dev.azure.com/numba/numba/_build/results?buildId=11273&view=logs&j=917c5bd5-a345-5b64-2c71-15c698d91dc1&t=5acb7173-a7eb-5b83-a52c-58af5e7f3ce0&l=865

These error messages make it look like the new overloads are being ignored entirely.

rishi-kulkarni:
@stuartarchibald I seem to have fixed it for reasons I neither understand, nor am happy with. I was never able to reproduce the errors on either of my machines, which makes me unsure I've completely fixed the issue. 

It seems that once I've added an overload for `np.random.random` to handle `size=None`, the other three aliases should just be able to be delegated through either the overload or the old-school method. I suspect there's still some incongruence between what's happening in the CI runners and what's happening on my development machine. 

Indeed, I think it'd be best to remove the old-style stuff - both for clarity's sake and maintainability. Are there any examples in the code of `@overload` handling a typing key, or would this have to be new functionality?

stuartarchibald:
> @stuartarchibald I seem to have fixed it for reasons I neither understand, nor am happy with. I was never able to reproduce the errors on either of my machines, which makes me unsure I've completely fixed the issue.

Thanks for working on this. Having looked at the logs from where it was failing, there's this:
```
AssertionError: No entries in the typing->lowering map for <built-in method random of numpy.random.mtrand.RandomState object at 0x7f3921ffa5a0>
```
which suggests that the `@overload_glue` was trying to resolve for the `random` method on a `numpy.random.mtrand.RandomState` instance, which implies that the typing key part was not being picked up.

> It seems that once I've added an overload for `np.random.random` to handle `size=None`, the other three aliases should just be able to be delegated through either the overload or the old-school method. I suspect there's still some incongruence between what's happening in the CI runners and what's happening on my development machine.

I'm slightly surprised this works without there being typing keys. However, I think it is working as the single "old style" function of `np.random.random` has a typing key _and_ will only work for the case where size is not `None`, then everything else is going through `@overload` and all of these delegate to the "old style" `np.random.random` with some variant of `size` that is accepted. What I'm not sure about is whether the NumPy API has been plumbed through in the way it was before, and will need to go in with a debugger and look.

> Indeed, I think it'd be best to remove the old-style stuff - both for clarity's sake and maintainability. Are there any examples in the code of `@overload` handling a typing key, or would this have to be new functionality?

I think so too. `@overload` doesn't handle typing keys yet, but this can be added if there's a demonstrable need (which I think there is).

rishi-kulkarni:
>I'm slightly surprised this works without there being typing keys. However, I think it is working as the single "old style" function of np.random.random has a typing key and will only work for the case where size is not None,

This makes sense. What doesn't make sense is that it works completely fine (including with `np.random.random(size=None)`) on my machine. I suspect that hints at some other underlying issue that I haven't grasped.

stuartarchibald:
> > I'm slightly surprised this works without there being typing keys. However, I think it is working as the single "old style" function of np.random.random has a typing key and will only work for the case where size is not None,
> 
> This makes sense. What doesn't make sense is that it works completely fine (including with `np.random.random(size=None)`) on my machine. I suspect that hints at some other underlying issue that I haven't grasped.

I'm not sure, there's some complex state in the "glue" code which is part of the reason it's intended to be temporary just whilst code is ported. It's possible that something locally was imported/updated in a different order to that on CI which yielded a different state.

rishi-kulkarni:
It'd be convenient for this effort (as well as for implementing `BitGenerator`, I suspect) if `@overload`  functions could directly consume `get_next_int`, `get_next_double`, etc. For example, an implementation of `np.random.random` might be:

```python
@overload(np.random.random)
def random(size=None):
    if size in (None, types.none):
        def impl(size=None):
            return get_next_double()
    else:
        def impl(size=None):
            out = np.empty(size)
            flat = out.flat
            for idx in range(out.size):
                flat[idx] = get_next_double()
            return out
    return impl

stuartarchibald:
@rishi-kulkarni thanks, I see. PR https://github.com/numba/numba/pull/7900 is adding support for this, see file `numba/np/npy_randomgen.py` in that PR, the function `generate_next_binding` is generating the bindings for `next_double` etc based on the underlying `BitGenerator`. As I understand it, a degree of caution over code reuse is needed as there seems to be some divergence between `Generator.<distribution_method>()` and `np.random.<distribution_function>`.

CC @brandonwillard and @kc611, they are working on the `Generator` support. I think some coordination could be needed here so that work isn't duplicated?

rishi-kulkarni:
@stuartarchibald I have some time to give this some thought again. I'm a little unclear on how typing keys work - as I understand it, they're needed in this instance because the actual `np.random` functions are bound methods of a class, is that right?

rishi-kulkarni:
Closing this, as it seems like significant progress on migrating `random` to `@overload` has been made.

stuartarchibald:
Thanks @rishi-kulkarni. The next release of Numba (0.57) will have support for a lot of the `Generator`/`BitGenerator` objects/methods, would be good to see if that provides some equivalent functionality.

PR #7904, opened 2022-03-11
Support m1

esc:

### DIFF
--- a/buildscripts/condarecipe.local/meta.yaml
+++ b/buildscripts/condarecipe.local/meta.yaml
@@ -54,7 +54,7 @@ requirements:
     # avoid confusion from openblas bugs
     - libopenblas !=0.3.6      # [x86_64]
     # 0.3.17 buggy on M1 silicon
-    # https://github.com/xianyi/OpenBLAS/blob/develop/Changelog.txt#L118
+    # https://github.com/xianyi/OpenBLAS/blob/v0.3.20/Changelog.txt#L118
     # https://github.com/numba/numba/issues/7822#issuecomment-1063229855
     - libopenblas >=0.3.18     # [arm64]
     # CUDA 9.2 or later is required for CUDA support
--- a/buildscripts/condarecipe.local/meta.yaml
+++ b/buildscripts/condarecipe.local/meta.yaml
@@ -55,6 +55,7 @@ requirements:
     - libopenblas !=0.3.6      # [x86_64]
     # 0.3.17 buggy on M1 silicon
     # https://github.com/xianyi/OpenBLAS/blob/develop/Changelog.txt#L118
+    # https://github.com/numba/numba/issues/7822#issuecomment-1063229855
     - libopenblas >=0.3.18     # [arm64]
     # CUDA 9.2 or later is required for CUDA support
     - cudatoolkit >=9.2
--- a/buildscripts/condarecipe.local/meta.yaml
+++ b/buildscripts/condarecipe.local/meta.yaml
@@ -53,6 +53,9 @@ requirements:
     - tbb >=2021    # [not (armv6l or armv7l or aarch64 or linux32 or ppc64le)]
     # avoid confusion from openblas bugs
     - libopenblas !=0.3.6      # [x86_64]
+    # 0.3.17 buggy on M1 silicon
+    # https://github.com/xianyi/OpenBLAS/blob/develop/Changelog.txt#L118
+    - libopenblas >=0.3.18     # [arm64]
     # CUDA 9.2 or later is required for CUDA support
     - cudatoolkit >=9.2
     # scipy 1.0 or later
--- a/numba/core/config.py
+++ b/numba/core/config.py
@@ -15,8 +15,11 @@ except ImportError:
 
 import llvmlite.binding as ll
 
+uname = platform.uname()
+
 IS_WIN32 = sys.platform.startswith('win32')
 IS_OSX = sys.platform.startswith('darwin')
+IS_OSX_ARM64 = uname.system == 'Darwin' and uname.machine == 'arm64'
 MACHINE_BITS = tuple.__itemsize__ * 8
 IS_32BITS = MACHINE_BITS == 32
 # Python version in (major, minor) tuple
--- a/numba/core/config.py
+++ b/numba/core/config.py
@@ -15,11 +15,11 @@ except ImportError:
 
 import llvmlite.binding as ll
 
-uname = platform.uname()
+_uname = platform.uname()
 
 IS_WIN32 = sys.platform.startswith('win32')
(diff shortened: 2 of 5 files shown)


numba/tests/support.py
```diff
@@ -107,6 +107,8 @@
                                           "parameter area' in "
                                           "LowerCall_64SVR4"))
 
+skip_m1 = unittest.skipIf(config.IS_OSX_ARM64, "Unreliable on OSX arm64")
```
stuartarchibald, line 110: This seems like quite a generic name for a function that is skipping for a specific reason? Suggest either making this `skip_m1` take the "reason" string such that it is generic, or making this function name a lot more specific.

numba/tests/npyufunc/test_errors.py
```diff
@@ -115,6 +115,7 @@ def test_floordiv_float(self):
                                 ["divide by zero encountered",
                                  "invalid value encountered"])
 
+    @skip_m1
```
stuartarchibald, line 118: Would be good to add a note here about the FPE warning tests being a little unreliable in general, with a link to the discussion on the M1 tracking ticket.

numba/core/config.py
```diff
@@ -15,8 +15,11 @@
 
 import llvmlite.binding as ll
 
+uname = platform.uname()
```
stuartarchibald, line 18:
```suggestion
_uname = platform.uname()
```

numba/core/config.py
```diff
@@ -15,8 +15,11 @@
...
+
 IS_WIN32 = sys.platform.startswith('win32')
 IS_OSX = sys.platform.startswith('darwin')
+IS_OSX_ARM64 = uname.system == 'Darwin' and uname.machine == 'arm64'
```
stuartarchibald, line 22:
```suggestion
IS_OSX_ARM64 = _uname.system == 'Darwin' and _uname.machine == 'arm64'
```

buildscripts/condarecipe.local/meta.yaml
```diff
@@ -53,6 +53,9 @@ requirements:
     - tbb >=2021    # [not (armv6l or armv7l or aarch64 or linux32 or ppc64le)]
     # avoid confusion from openblas bugs
     - libopenblas !=0.3.6      # [x86_64]
+    # 0.3.17 buggy on M1 silicon
```
stuartarchibald, line 56: Suggest adding a cross reference to the discussion on this as it demonstrates how the bugs materialise.

stuartarchibald (commented):
Thanks for the patch, few minor things to look at inline but otherwise looks good.

numba/tests/support.py
esc: I will make the skip name more specific.

numba/tests/npyufunc/test_errors.py
esc: I will add a comment to the definition of the skip

numba/core/config.py
esc: Sure.

numba/core/config.py
esc: Sure.

esc:
Screen Shot 2022-03-23 at 16 32 18

This is now passing on the Anaconda internal build-farm (albeit with conda-forge packages) -- but this PR is now in a solid state and needs a final round of review.

buildscripts/condarecipe.local/meta.yaml
```diff
@@ -53,6 +53,10 @@ requirements:
...
+    # https://github.com/xianyi/OpenBLAS/blob/develop/Changelog.txt#L118
```
stuartarchibald, line 57: This reference isn't stable, it refers to the `develop` branch?

numba/core/config.py
```diff
@@ -15,8 +15,11 @@
 
 import llvmlite.binding as ll
 
+_uname = platform.uname()
+
 IS_WIN32 = sys.platform.startswith('win32')
 IS_OSX = sys.platform.startswith('darwin')
+IS_OSX_ARM64 = _uname.system == 'Darwin' and _uname.machine == 'arm64'
```
stuartarchibald, line 22: Thinking about the impact this has on import time... does this need to be in here opposed to in e.g. `numba.tests.support` ? `uname` will make a call to `uname(3P)` and IIRC `core/config.py` is on the main import path from `__init__.py`.

numba/tests/support.py
```diff
@@ -107,6 +107,10 @@
                                           "parameter area' in "
                                           "LowerCall_64SVR4"))
 
+# fenv.h on M1 may have various issues:
+# https://github.com/numba/numba/issues/7822#issuecomment-1065356758
+skip_m1_fenv_errors = unittest.skipIf(config.IS_OSX_ARM64, "Unreliable on OSX arm64")
```
stuartarchibald, line 112:
```suggestion
skip_m1_fenv_errors = unittest.skipIf(config.IS_OSX_ARM64, "fenv.h-like functionality unreliable on OSX arm64")
```
(and line wrap).

stuartarchibald (commented):
Thanks for persisting with this patch to get the M1 support working. I've made a few suggestions inline, once resolved looks good to merge.

buildscripts/condarecipe.local/meta.yaml
esc: Good catch, thank you.

numba/tests/support.py
esc: sure

numba/core/config.py
stuartarchibald: I'd also rather they were all in one place (e.g. `numba.core.system` or something), and also lazy, but that's out of scope for this PR. As this check is solely used in `numba.tests.support`, I think it's ok to move it there, there are similar system checks there already.

esc:
Quick update here. After slogging through everything, I went to test this again on the Anaconda internal build farm and it appears as though there are more "issues":

```
% /Users/ci/miniconda3-arm64/envs/testenv_765afdac-a85f-42f2-b63a-472a3a7c62ee/bin/python
Python 3.9.12 | packaged by conda-forge | (main, Mar 24 2022, 07:11:11)
[Clang 12.0.1 ] on darwin
zsh: killed
```

I will continue to look into this.

esc:
> I will continue to look into this.

I found that the current `python` package provided by `conda-forge`, the `3.9.12` package is actually broken:

Screen Shot 2022-03-25 at 11 25 59

When testing with Python `3.9.6` -- which was the prior available `python` for `osx-arm64` everything passed. So I am marking this as having passed internal CI

stuartarchibald (approved):
Thanks for the patch and all the work on getting the M1 supported!

PR #7911, opened 2022-03-15
added sys import

Nightfurex:

### DIFF
--- a/numba/parfors/parfor_lowering.py
+++ b/numba/parfors/parfor_lowering.py
@@ -1,6 +1,5 @@
 import copy
 import operator
-import sys
 
 import types as pytypes
 import operator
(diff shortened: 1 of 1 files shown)


DrTodd13 (changes requested):
See my comment about removing the line that uses sys.

Nightfurex:
Ooh sorry my bad i misunderstood .

numba/parfors/parfor_lowering.py
```diff
@@ -323,8 +322,7 @@ def _lower_parfor_parallel(lowerer, parfor):
         index_var_typ,
         parfor.races,
         exp_name_to_tuple_var)
-    if config.DEBUG_ARRAY_OPT:
-        sys.stdout.flush()
+
```
DrTodd13, line 325: Should only be one blank line here.

DrTodd13 (changes requested):

DrTodd13:
Thanks for making this change.  Just that one additional picky thing about the extra line and I think we're done.

DrTodd13 (approved):
Thanks!

stuartarchibald:
Thanks for the patch @Nightfurex, thanks for reviewing @DrTodd13.

Issue #4266, opened 2019-07-03
Add numba.cuda.jit examples


mrocklin:
I'm sitting with a group of people and we started talking about numba.cuda.jit. I turned a recent blogpost into a Google Colab notebook, and it seems to be somewhat effective in getting people started.  https://gist.github.com/mrocklin/cdb99419552bd8df2f2bea2adfd6231f

However, they don't know how to extend this, except with other stencil operations (my example is a stencil computation).  I think that if they had a few examples for reductions, matmul ([which we have already](https://numba.pydata.org/numba-doc/dev/cuda/examples.html#examples)), elementwise, and so on, then they would be able to pull from those and generalize a bit.  

So concretely, I'd love to see an expansion of the numba.cuda.jit examples documentation.  Secondarily I'd love to see this in notebook form that could run on Colab (or some other GPU-cloud service).

mrocklin:
cc @stuartarchibald (who I think will be interacting with this same group tomorrow)

stuartarchibald:
Does the GTC tutorial series help: https://github.com/ContinuumIO/gtc2019-numba ?

stuartarchibald:
xref: https://github.com/numba/numba/issues/4288

gmarkall:
Potentially useful examples:

- [ ] Use of `@jit` function from both CPU- and GPU-jitted functions
- [ ] CuPy / Numba interop examples
- [ ] Transcribe / translate appropriate examples from CUDA release notebooks: https://github.com/numba/numba-examples/tree/master/notebooks

quasiben:
@seanlaw do you have an thoughts on good examples of using numba GPU kernels ?

seanlaw:
@quasiben I don't have any particularly good examples that are easily approachable but, from my limited experience, I've found that the biggest challenge/barrier to entry for users is being able identify appropriate SIMD scenarios to fit the GPU programming paradigm. It almost feels like there are some basic patterns that the user needs to get exposed to first and then, later, be able to recognize it as a chunk of CPU code that would be great on a GPU.

brandon-b-miller:
I'm working on this and should have a PR shortly if thats ok :)

sklam:
Please note that examples should now go into https://github.com/numba/numba-examples.

PR #7918, opened 2022-03-16
Add JIT examples to CUDA docs

brandon-b-miller:
Hi! I thought I'd add some examples I have pulled together around some use cases for `cuda.jit` since there is an issue for it. I added some of my own as well as those I could glean from the discussion in the issue. 

There was some discussion of moving these to the numba examples repo but after discussion with @gmarkall we decided to submit them here. 

Closes https://github.com/numba/numba/issues/4266


sklam (commented):
Thank you for the contribution. 

One worry that I have is ease of testing the validity of the examples as API may drift. I'd suggest include the examples as a runnable `.py` script and use `literalinclude` to reference the lines. See example in https://github.com/numba/numba/blame/b57da5b52cb330ec03837a721e49eb8c9b01aa66/docs/source/cuda/examples.rst#L485

gmarkall:
I have looked over the vector example, and in https://github.com/gmarkall/numba/tree/jit-examples in the commit https://github.com/gmarkall/numba/commit/7632af7e241d8d5d68468742485b4622ea3cb32e I have some suggested edits. The main aim behind the edits is to:

- Simplify the wording / shorten it,
- Change the example code into a test - the same process should be repeatable for the other examples,
- Present more idiomatic usage of CUDA (e.g. not passing in array bounds, using a grid of reasonable size, using `device_array_like` for empty allocations, etc.)

I'll continue to look at the examples but wanted to push this feedback sooner rather than later.

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,479 @@
...
+               next_temp = curr_temp + k * (data[i + 1] - (2 * curr_temp))
+         elif i == size - 1:
+               # Right wall is held at T = 0
+               next_temp = curr_temp + k * (data[i - 1] - (2 * curr_temp))
+         else:
+               next_temp = curr_temp + k * (data[i - 1] - (2 * curr_temp) + data[i + 1])
+         data[i] = next_temp
+         
+         # wait for every thread to write before moving on
+         cuda.syncthreads()
```
gmarkall, line 146: How does syncing at the block level have any meaningful effect here?

It feels like this example needs two copies of the domain along with a cooperative grid sync, or one kernel launch per timestep in order to ensure that all inputs are taken from the previous timestep and not some values coming from the current timestep at the boundaries between blocks.

Am I misunderstanding the operation of the example?

gmarkall, line 146: How does syncing at the block level have any meaningful effect here?

It feels like this example needs two copies of the domain along with a cooperative grid sync, or one kernel launch per timestep in order to ensure that all inputs are taken from the previous timestep and not some values coming from the current timestep at the boundaries between blocks.

Am I misunderstanding the operation of the example?

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,479 @@
...
+         2.72682981,  2.18464289,  1.64033805,  1.09444297,  0.54748687])
+
+Plotting this data with a graphing library shows an arc shape highest where
+the object was hot initially and gradually sloping down to zero towards the
+edges where the temperature is fixed at zero. In the limit of infinite time,
+the arc will flatten out completely.
+
+.. _cuda_reduction_shared:
+
+Shared Memory Reduction
```
gmarkall, line 193: We should probably note that this as an example of how to use shared memory rather than the way in which reductions should actually be implemented using Numba - we could point to the `@cuda.reduce` decorator (https://numba.readthedocs.io/en/stable/cuda/reduction.html) and/or demonstrate its use, as it provides an easier interface for building a reduction and uses a slightly more sophisticated (hopefully more performant) implementation: https://github.com/numba/numba/blob/main/numba/cuda/kernels/reduction.py

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,479 @@
...
+.. code-block:: python
+
+   solve_heat_equation.forall(len(data))(data_gpu, len(data_gpu), 10000, 0.25)
+
+Since the kernel reused the original data vector to work on the problem, copying it back
+to the host yields the results:
+
+.. code-block:: python
+
+   print(data_gpu.copy_to_host())
```
gmarkall, line 159: It might be nice to include before and after plots generated with something like:

```python
import matplotlib.pyplot as plt

plt.plot(np.arange(len(data)), data_gpu.copy_to_host())
plt.title('Heat in a bar')
plt.xlabel('Position')
plt.ylabel('Temperature')
plt.show()
```

to give something like:

temp

gmarkall, line 159: It might be nice to include before and after plots generated with something like:

```python
import matplotlib.pyplot as plt

plt.plot(np.arange(len(data)), data_gpu.copy_to_host())
plt.title('Heat in a bar')
plt.xlabel('Position')
plt.ylabel('Temperature')
plt.show()
```

to give something like:

temp

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,479 @@
...
+      ### DETERMINE SESSION LABELS
+      # this thread marks the start of a session
+      if gid < size:
+         if is_sess_boundary:
+               results[gid] = gid
+               look_ahead = 1
+
+               # check elements 'forward' of this one 
+               # until a new session boundary is found
+               while results[gid + look_ahead] == 0:
```
gmarkall, line 320: How is it guaranteed that if another session boundary is written by the next thread, it is correctly seen by the current thread?

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,479 @@
...
+      if 0 < gid < size:
+         if (
+               user_id[gid] != user_id[gid - 1]
+               or timestamp[gid] - timestamp[gid - 1] > session_timeout
+         ):
+               is_sess_boundary = 1
+         else:
+               is_sess_boundary = 0
+      else:
+         is_sess_boundary = 1
```
gmarkall, line 308: Could it be made a little clearer what this is doing by writing it as:


```suggestion
         # Exit early if our gid is beyond the bounds of the data
         if (gid >= len(user_id)):
             return
  
         # Determine whether we're on a session boundary
         different_user = (user_id[gid] != user_id[gid - 1]) 
         time_diff = timestamp[gid] - timestamp[gid - 1]
         timed_out = time_diff > session_timeout
         first_datapoint = (gid == 0)
         is_sess_boundary = different_user or timed_out or first_datapoint
```

i.e. the variable names describe the checks and the condition for being a session boundary "reads" in plain language as opposed to being a more complicated expression.

Also this would allow the next `if` to be removed and the code under it to be moved to the same level as the rest of the function. 

(Note that my indentation in the edited code may not match that in the diff at present - there's some weird 3-space indentation going on in places)

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,479 @@
...
+   # set the timeout to one hour
+   session_timeout = np.int64(np.timedelta64('3600', 's'))
+   
+Here is a solution using Numba:
+
+.. code-block:: python
+
+   @cuda.jit
+   def sessionize(
+      user_id, timestamp, results, size
```
gmarkall, line 292: Can we not pass in the `size` and instead use `len(user_id)` in the kernel?

gmarkall, line 292: Can we not pass in the `size` and instead use `len(user_id)` in the kernel?

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,479 @@
...
+..code-block:: python
+
+   print(business_logic(1,2,3)) # -126.79644737231007
+
+It can also be directly reused threadwise inside a GPU kernel. For example one may 
+generate some vectors to represent ``x``, ``y``, and ``z``:
+
+.. code-block:: python
+
+   X = cp.array([1, 10, 234])
```
gmarkall, line 390: Although I do love to champion CuPy / Numba interoperability, we should probably keep the Numba examples self-contained and use Numba device arrays here.

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,479 @@
...
+
+   @numba.jit
+   def business_logic(x, y, z):
+      return 4 * z * (2 * x - (4 * y) / 2 * pi)
+
+The function ``business_logic`` can be run standalone in compiled form on the CPU:
+
+..code-block:: python
+
+   print(business_logic(1,2,3)) # -126.79644737231007
```
gmarkall, line 383: I think the 4-space indentation makes this fail to render as code (but it should probably be converted to a tested example and `literalinclude`d anyway)

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,479 @@
...
+   Z = cp.array([3,14,2211])
+
+   results = cp.array([0.,0.,0.])
+
+And a numba kernel referencing the decorated function:
+
+.. code-block:: python
+
+   @cuda.jit
+   def f(res, xarr, yarr, zarr, size):
```
gmarkall, line 401: Can we use `len(xarr)` or `len(res)` instead of passing in the size?

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,479 @@
...
+=======================
+
+This example shows how to use numba to approximate the value of a definite integral by rapidly generating 
+random numbers on the GPU. A detailed description of the mathematical mechanics of monte carlo integeration 
+is out of the scope of the example, but it can briefly be described as an averaging process where the area 
+under the curve is approximated by taking the average of many rectangles formed by its function values.
+
+.. code-block:: python
+
+   nsamps = 1000000 # numer of samples, higher will lead to a more accurate answer
```
gmarkall, line 427:
```suggestion
   nsamps = 1000000 # number of samples, higher will lead to a more accurate answer
```

gmarkall (changes requested):
Many thanks for the PR - in addition to the commit I linked to in the previous comment, I've left some additional comments on the diff. I think that each of the examples needs to be converted to a test and literalincluded as @sklam suggests, and this will also highlight a couple of things that prevent the code running as-is. 

The use of CuPy is nice but I'd prefer if we kept to Numba code for these examples, although it would be great if we also had a Numba +CuPy interop example (I'm not asking for one to be included in this PR though :-) )

Once the examples are converted to tests I think we might be able to streamline some of the explanations and some of the code too, but I'd prefer to iterate on that once we have the code converted to tested samples.

docs/source/cuda/examples.rst
brandon-b-miller: Yeah, this example was one that I remembered from academia. I think I derived this from some memory of a pure c++ example which used a grid wide barrier. This will be updated.

docs/source/cuda/examples.rst
brandon-b-miller: I will add a comment here, and I used the `cuda.reduce` technique later on in the monte-carlo example.

docs/source/cuda/examples.rst
brandon-b-miller: I really wanted to include some plots here but didn't see any precedent for that elsewhere in the docs. Happy to add them in though.

docs/source/cuda/examples.rst
brandon-b-miller: Ahh, you are right. The reason is these used to be two separate kernels that I fused into one, forgetting that the synchronization that comes from two distinct kernel launches is actually critical. The first used to write the boundaries and the second built out the IDs. Will look into alternatives.

brandon-b-miller:
I think this should be ready for a second pass.

gmarkall:
gpuci run tests

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,398 @@
...
+.. _cuda_reduction_shared:
+
+Shared Memory Reduction
+=======================
+Numba exposes many CUDA features including shared memory. Shared memory is high speed on chip memory
+that represents a shared address space that all the threads in a block can see. It is local to each
+block and is limited in size, but extremely quick to access, similar to a cache. More can be read 
+about shared memory in the CUDA documentation. As a way of demonstrating shared memory, let's reimplement 
+a famous CUDA solution for summing a vector which works by "folding" the data up using a sucessively
+smaller number of threads.
```
gmarkall, line 179: I think we can skip explaining shared memory here, and link to the section documenting it:
```suggestion
Numba exposes many CUDA features including :ref:`shared memory
<cuda-shared-memory>`. To demonstrate shared memory, let's reimplement a
famous CUDA solution for summing a vector which works by "folding" the data up
using a sucessively smaller number of threads.

```

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,398 @@
...
+
+Note that this is a fairly naive implementation and more efficient ways of summing an array exist
+within numba. See :ref:`cuda_montecarlo` for an example.
```
gmarkall, line 182:
```suggestion
Note that this is a fairly naive implementation, and there are more efficient ways of implementing reductions
using Numba - see :ref:`cuda_montecarlo` for an example.
```

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,398 @@
...
+   :language: python
+   :caption: from ``test_ex_reduction`` in ``numba/cuda/tests/doc_examples/test_reduction.py``
+   :start-after: ex_reduction.launch.begin
+   :end-before: ex_reduction.launch.end
+   :dedent: 8
+   :linenos:
+
+This algorithm can be greatly improved upon by redesigning the inner loop
+to use sequential memory accesses, or even furthermore using strategies that
+keep more threads active and working since in this example most threads quickly
```
gmarkall, line 215:
```suggestion
keep more threads active and working, since in this example most threads quickly
```

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,398 @@
...
+.. literalinclude:: ../../../numba/cuda/tests/doc_examples/test_reduction.py
+   :language: python
+   :caption: from ``test_ex_reduction`` in ``numba/cuda/tests/doc_examples/test_reduction.py``
+   :start-after: ex_reduction.launch.begin
+   :end-before: ex_reduction.launch.end
+   :dedent: 8
+   :linenos:
+
+This algorithm can be greatly improved upon by redesigning the inner loop
+to use sequential memory accesses, or even furthermore using strategies that
```
gmarkall, line 214:
```suggestion
to use sequential memory accesses, and even further by using strategies that
```

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,398 @@
...
+
+.. _cuda_sessionization:
+
+Dividing Click Data into Sessions
+=================================
+
+
+A common problem in business analytics is that of grouping the activity of users of an online platform into
+sessions, called "sessionization". The idea is that users generally traverse through a website and perform
+various actions (clicking something, filling out a form, etc) in discrete groups. Perhaps a customer spends
```
gmarkall, line 226:
```suggestion
various actions (clicking something, filling out a form, etc.) in discrete groups. Perhaps a customer spends
```

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,398 @@
...
+some time shopping for an item in the morning and then again at night - often the business is interested in
+treating these periods as separate interactions with their service, and this creates the problem of 
+programmatically splitting up activity in some agreed upon way.
```
gmarkall, line 229:
```suggestion
programmatically splitting up activity in some agreed-upon way.
```

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,398 @@
...
+
+Here we'll illustrate how to write a group of numba kernels to solve this problem. We'll start with data
```
gmarkall, line 231: I only see one kernel:
```suggestion
Here we'll illustrate how to write a Numba kernel to solve this problem. We'll start with data 
```

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,398 @@
...
+
+A common problem in business analytics is that of grouping the activity of users of an online platform into
+sessions, called "sessionization". The idea is that users generally traverse through a website and perform
+various actions (clicking something, filling out a form, etc) in discrete groups. Perhaps a customer spends
+some time shopping for an item in the morning and then again at night - often the business is interested in
+treating these periods as separate interactions with their service, and this creates the problem of 
+programmatically splitting up activity in some agreed upon way.
+
+Here we'll illustrate how to write a group of numba kernels to solve this problem. We'll start with data 
+containing two fields: let user_id represent a unique ID corresponding to an individual customer, and let
```
gmarkall, line 232:
```suggestion
containing two fields: let ``user_id`` represent a unique ID corresponding to an individual customer, and let 
```

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,398 @@
...
+A common problem in business analytics is that of grouping the activity of users of an online platform into
+sessions, called "sessionization". The idea is that users generally traverse through a website and perform
+various actions (clicking something, filling out a form, etc) in discrete groups. Perhaps a customer spends
+some time shopping for an item in the morning and then again at night - often the business is interested in
+treating these periods as separate interactions with their service, and this creates the problem of 
+programmatically splitting up activity in some agreed upon way.
+
+Here we'll illustrate how to write a group of numba kernels to solve this problem. We'll start with data 
+containing two fields: let user_id represent a unique ID corresponding to an individual customer, and let 
+action_time be a time that some unknown action was taken on the service. Right now, we'll assume there's
```
gmarkall, line 233:
```suggestion
``action_time`` be a time that some unknown action was taken on the service. Right now, we'll assume there's 
```

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,398 @@
...
+some time shopping for an item in the morning and then again at night - often the business is interested in
+treating these periods as separate interactions with their service, and this creates the problem of 
+programmatically splitting up activity in some agreed upon way.
+
+Here we'll illustrate how to write a group of numba kernels to solve this problem. We'll start with data 
+containing two fields: let user_id represent a unique ID corresponding to an individual customer, and let 
+action_time be a time that some unknown action was taken on the service. Right now, we'll assume there's 
+only one type of action, so all there is to know is when it happened.
+
+Our goal will be to create a new column called session_id which contains a label corresponding to a unique
```
gmarkall, line 236:
```suggestion
Our goal will be to create a new column called ``session_id``, which contains a label corresponding to a unique 
```

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,398 @@
...
+
+.. literalinclude:: ../../../numba/cuda/tests/doc_examples/test_sessionize.py
+   :language: python
+   :caption: from ``test_ex_sessionize`` in ``numba/cuda/tests/doc_examples/test_sessionize.py``
+   :start-after: ex_sessionize.kernel.begin
+   :end-before: ex_sessionize.kernel.end
+   :dedent: 8
+   :linenos:
+
+Let's generate some data and try out the kernel.
```
gmarkall, line 258:
```suggestion
Let's generate some data and try out the kernel:
```

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,398 @@
...
+.. literalinclude:: ../../../numba/cuda/tests/doc_examples/test_sessionize.py
+   :language: python
+   :caption: from ``test_ex_sessionize`` in ``numba/cuda/tests/doc_examples/test_sessionize.py``
+   :start-after: ex_sessionize.allocate.begin
+   :end-before: ex_sessionize.allocate.end
+   :dedent: 8
+   :linenos:
+
+As can be seen above, the kernel successfully divided the first three datapoints from the second three for the first user ID,
+and a similar pattern is seen througout. This example has wide applicability in a business setting and can be used to study
```
gmarkall, line 269: The sentence beginning "This example has..." seems out-of-place here and partially re-states the introduction to the problem - can we either somehow move what it says into the first paragraph, or just remove it?

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,398 @@
...
+   :end-before: ex_cpu_gpu_compat.launch.end
+   :dedent: 8
+   :linenos:
+
+.. _cuda_montecarlo:
+
+Monte Carlo Integration
+=======================
+
+This example shows how to use numba to approximate the value of a definite integral by rapidly generating
```
gmarkall, line 337:
```suggestion
This example shows how to use Numba to approximate the value of a definite integral by rapidly generating 
```

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,398 @@
...
+   :dedent: 8
+   :linenos:
+
+.. _cuda_montecarlo:
+
+Monte Carlo Integration
+=======================
+
+This example shows how to use numba to approximate the value of a definite integral by rapidly generating 
+random numbers on the GPU. A detailed description of the mathematical mechanics of monte carlo integeration
```
gmarkall, line 338:
```suggestion
random numbers on the GPU. A detailed description of the mathematical mechanics of Monte Carlo integeration 
```

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,398 @@
...
+
+Monte Carlo Integration
+=======================
+
+This example shows how to use numba to approximate the value of a definite integral by rapidly generating 
+random numbers on the GPU. A detailed description of the mathematical mechanics of monte carlo integeration 
+is out of the scope of the example, but it can briefly be described as an averaging process where the area 
+under the curve is approximated by taking the average of many rectangles formed by its function values.
+
+In addition this example shows how to perform reductions in numba using the :func:`numba.cuda.reduce` API.
```
gmarkall, line 342:
```suggestion
In addition, this example shows how to perform reductions in numba using the :func:`numba.cuda.reduce` API.
```

gmarkall: (Note that the link to the reduction docs isn't working here either)

gmarkall (changes requested):
Many thanks for the updates - I've had another pass at the review - mostly everything is small suggestions on the diff, but in addition to that:

- Could you commit the code that you used to generate the plots? It will be a pain if we have to regenerate them and the code isn't there. I think it would be fine to add it to the `test_laplace.py` file, guarded by a conditional which is always false (then if the plots need to be regenerated, the current editor of the docs can set the conditional to True temporarily and run the test and commit the new output)
- Can you convert the Monte Carlo integration code to a test?

Many thanks in advance!

gmarkall:
Ah, I forgot to mention - the gpuCI fails are with toolkits that were incomplete, because they are missing libcudadevrt.a - this was fixed in the cudatoolkit recipe, but was never resolved for old toolkits like 9.0 and 10.0. Since we're dropping support these for the next release, it doesn't matter that the tests fail with these toolkits due to their incompleteness.

gmarkall:
gpuci run tests

numba/cuda/tests/doc_examples/test_vecadd.py
```diff
@@ -0,0 +1,69 @@
...
+        @cuda.jit
+        def f(a, b, c):
+            # like threadIdx.x + (blockIdx.x * blockDim.x)
+            tid = cuda.grid(1)
+            size = len(c)
+
+            if tid < size:
+                c[tid] = a[tid] + b[tid]
+
+        # ex_vecadd.kernel.end
```
gmarkall, line 41:
```suggestion
        # ex_vecadd.kernel.end
```

numba/cuda/tests/doc_examples/test_vecadd.py
```diff
@@ -0,0 +1,69 @@
...
+        # No exception type, value, or traceback
+        self._captured_stdout.__exit__(None, None, None)
+        super().tearDown()
+
+    def test_ex_vecadd(self):
+        # ex_vecadd.import.begin
+        import numpy as np
+        from numba import cuda
+
+        # ex_vecadd.import.end
```
gmarkall, line 30:
```suggestion
        # ex_vecadd.import.end

```

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,397 @@
...
+conditions and boundary conditions. A full discussion of Laplace's equation is out of
+scope for this documentation, but it will suffice to say that it describes how heat
+propagates through an object over time. It works by discretizing the problem in two ways:
+
+1. The domain is partitioned into a mesh of points that each have an individual temperature.
+2. Time is partitioned into discrete intervals that are advanced forward sequentially.
+
+Then, the following assumption is applied: The temperature of a point after some interval 
+has passed is some weighted average of the temperature of the points that are directly
+touching it. Intuitively, this means if all the points in the domain are very hot
```
gmarkall, line 106:
```suggestion
adjacent to it. Intuitively, if all the points in the domain are very hot
```

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,397 @@
...
+and a single point in the middle is very cold, as time goes on, the hot points will cause
```
gmarkall, line 107:
```suggestion
and a single point in the middle is very cold, as time passes, the hot points will cause
```

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,397 @@
...
+the cold one to heat up and the cold point will cause the surrounding hot pieces to cool
+slightly. Simply put, the heat spreads out around the object.
```
gmarkall, line 109:
```suggestion
slightly. Simply put, the heat spreads throughout the object.
```

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,397 @@
...
+
+We can implement this simulation using a Numba kernel. Let's start simple by assuming
+we have a one dimensional object which we'll represent with an array of values. The position 
+of the element in the array is like the position of a point within the object, and the value
```
gmarkall, line 113:
```suggestion
of the element in the array is the position of a point within the object, and the value
```

numba/cuda/tests/doc_examples/test_laplace.py
```diff
@@ -0,0 +1,129 @@
...
+
+        # set True to regenerate the figures that
+        # accompany this example
+        plot = False
+
+        # ex_laplace.import.begin
+        import numpy as np
+        from numba import cuda
+
+        # ex_laplace.import.end
```
gmarkall, line 34:
```suggestion
        # ex_laplace.import.end

```

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,397 @@
...
+
+.. literalinclude:: ../../../numba/cuda/tests/doc_examples/test_laplace.py
+   :language: python
+   :caption: from ``test_ex_laplace`` in ``numba/cuda/tests/doc_examples/test_laplace.py``
+   :start-after: ex_laplace.allocate.begin
+   :end-before: ex_laplace.allocate.end
+   :dedent: 8
+   :linenos:
+
+The initial state of the problem can be visualized with ``matplotlib``:
```
gmarkall, line 135:
```suggestion
The initial state of the problem can be visualized as:
```

numba/cuda/tests/doc_examples/test_laplace.py
```diff
@@ -0,0 +1,129 @@
...
+                        data[i - 1] - (2 * curr_temp) + data[i + 1]
+                    )
+                tmp[i] = next_temp
+                # wait for every thread to write before moving on
+                grid.sync()
+
+                # swap data vectors for the next iteration
+                data[i] = tmp[i]
+
+        # ex_laplace.kernel.end
```
gmarkall, line 104:
```suggestion
        # ex_laplace.kernel.end
```

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,397 @@
...
+.. image:: laplace_final.svg
+
+.. _cuda_reduction_shared:
+
+Shared Memory Reduction
+=======================
+Numba exposes many CUDA features including :ref:`shared memory
+<cuda-shared-memory>`. To demonstrate shared memory, let's reimplement a
+famous CUDA solution for summing a vector which works by "folding" the data up
+using a sucessively smaller number of threads.
```
gmarkall, line 178:
```suggestion
using a successively smaller number of threads.
```

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,397 @@
...
+edges where the temperature is fixed at zero. In the limit of infinite time,
+the arc will flatten out completely.
+
+.. image:: laplace_final.svg
+
+.. _cuda_reduction_shared:
+
+Shared Memory Reduction
+=======================
+Numba exposes many CUDA features including :ref:`shared memory
```
gmarkall, line 175:
```suggestion
Numba exposes many CUDA features, including :ref:`shared memory
```

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,397 @@
...
+.. literalinclude:: ../../../numba/cuda/tests/doc_examples/test_reduction.py
+   :language: python
+   :caption: from ``test_ex_reduction`` in ``numba/cuda/tests/doc_examples/test_reduction.py``
+   :start-after: ex_reduction.kernel.begin
+   :end-before: ex_reduction.kernel.end
+   :dedent: 8
+   :linenos:
+
+This kernel can be run as follows and the same result is obtained through
+summing the host data by traditional means.
```
gmarkall, line 203:
```suggestion
We can run kernel and verify that the same result is obtained through
summing data on the host as follows:
```

docs/source/cuda/examples.rst
gmarkall: What did you think of this suggestion? I think the original version of the code is in the test file.

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,397 @@
...
+
+Monte Carlo Integration
+=======================
+
+This example shows how to use Numba to approximate the value of a definite integral by rapidly generating 
+random numbers on the GPU. A detailed description of the mathematical mechanics of Monte Carlo integeration 
+is out of the scope of the example, but it can briefly be described as an averaging process where the area 
+under the curve is approximated by taking the average of many rectangles formed by its function values.
+
+In addition, this example shows how to perform reductions in numba using the :func:`numba.cuda.reduce` API.
```
gmarkall, line 341: The link doesn't work because `reduce` is an alias, I think:

```suggestion
In addition, this example shows how to perform reductions in numba using the 
:func:`cuda.reduce() <numba.cuda.Reduce>` API.
```

numba/cuda/tests/doc_examples/test_laplace.py
```diff
@@ -0,0 +1,129 @@
...
+
+        # ex_laplace.import.begin
+        import numpy as np
+        from numba import cuda
+
+        # ex_laplace.import.end
+        # ex_laplace.allocate.begin
+        # use an odd problem size
+        # this is so there can be an element truly
+        # in the "middle" for symmetry
```
gmarkall, line 38:
```suggestion
        # Use an odd problem size.
        # This is so there can be an element truly in the "middle" for symmetry.
```

numba/cuda/tests/doc_examples/test_laplace.py
```diff
@@ -0,0 +1,129 @@
...
+        size = 1001
+        data = np.zeros(size)
+
+        # middle element is made very hot
```
gmarkall, line 42:
```suggestion
        # Middle element is made very hot
```

numba/cuda/tests/doc_examples/test_laplace.py
```diff
@@ -0,0 +1,129 @@
...
+        data[500] = 10000
+        data_gpu = cuda.to_device(data)
+
+        # this extra array is used in the algorithm for
+        # synchronization purposes
```
gmarkall, line 47:
```suggestion
        # This extra array is used for synchronization purposes
```

numba/cuda/tests/doc_examples/test_laplace.py
```diff
@@ -0,0 +1,129 @@
...
+            plt.xlim(0, len(data))
+            plt.ylim(0, 10001)
+            plt.savefig('laplace_initial.svg')
+
+        # ex_laplace.kernel.begin
+        @cuda.jit
+        def solve_heat_equation(data, tmp, timesteps, k):
+            i = cuda.grid(1)
+
+            # prepare to do a grid-wide synchronization later
```
gmarkall, line 79:
```suggestion
            # Prepare to do a grid-wide synchronization later
```

numba/cuda/tests/doc_examples/test_laplace.py
```diff
@@ -0,0 +1,129 @@
...
+            grid = cuda.cg.this_grid()
+
+            for step in range(timesteps):
+                # get the current temperature associated with this segment
```
gmarkall, line 83:
```suggestion
                # Get the current temperature associated with this point
```

numba/cuda/tests/doc_examples/test_laplace.py
```diff
@@ -0,0 +1,129 @@
...
+                curr_temp = data[i]
+
+                # apply formula from finite difference equation
```
gmarkall, line 86:
```suggestion
                # Apply formula from finite difference equation
```

numba/cuda/tests/doc_examples/test_laplace.py
```diff
@@ -0,0 +1,129 @@
...
+                    next_temp = curr_temp + k * (data[i + 1] - (2 * curr_temp))
+                elif i == len(data) - 1:
+                    # Right wall is held at T = 0
+                    next_temp = curr_temp + k * (data[i - 1] - (2 * curr_temp))
+                else:
+                    next_temp = curr_temp + k * (
+                        data[i - 1] - (2 * curr_temp) + data[i + 1]
+                    )
+                tmp[i] = next_temp
+                # wait for every thread to write before moving on
```
gmarkall, line 98:
```suggestion

                # Wait for every thread to write before moving on
```

numba/cuda/tests/doc_examples/test_laplace.py
```diff
@@ -0,0 +1,129 @@
...
+                grid.sync()
+
+                # swap data vectors for the next iteration
```
gmarkall, line 101:
```suggestion
                # Swap data for the next iteration
```

numba/cuda/tests/doc_examples/test_reduction.py
```diff
@@ -0,0 +1,78 @@
...
+        self._captured_stdout.__exit__(None, None, None)
+        super().tearDown()
+
+    def test_ex_reduction(self):
+        # ex_reduction.import.begin
+        import numpy as np
+        from numba import cuda
+        from numba.types import int32
+
+        # ex_reduction.import.end
```
gmarkall, line 30:
```suggestion
        # ex_reduction.import.end
```

numba/cuda/tests/doc_examples/test_reduction.py
```diff
@@ -0,0 +1,78 @@
...
+                        shr[tid] += shr[tid + s]
+                    s *= 2
+                    cuda.syncthreads()
+
+                # after the loop, the zeroth
+                # element contains the sum
+                if tid == 0:
+                    data[tid] = shr[tid]
+
+        # ex_reduction.kernel.end
```
gmarkall, line 68:
```suggestion
        # ex_reduction.kernel.end
```

numba/cuda/tests/doc_examples/test_reduction.py
```diff
@@ -0,0 +1,78 @@
...
+
+        # ex_reduction.kernel.begin
+        @cuda.jit
+        def array_sum(data):
+            tid = cuda.threadIdx.x
+            size = len(data)
+            if tid < size:
+                i = cuda.grid(1)
+
+                # declare an array in shared memory
```
gmarkall, line 46:
```suggestion
                # Declare an array in shared memory
```

numba/cuda/tests/doc_examples/test_reduction.py
```diff
@@ -0,0 +1,78 @@
...
+        from numba import cuda
+        from numba.types import int32
+
+        # ex_reduction.import.end
+
+        # ex_reduction.allocate.begin
+        # generate data
+        a = cuda.to_device(np.arange(1024))
+        nelem = len(a)
+        # ex_reduction.allocate.end
```
gmarkall, line 36: I don't see this used in the documentation, so I think it gets omitted, and the user can't just copy and paste the example to run it.

numba/cuda/tests/doc_examples/test_reduction.py
```diff
@@ -0,0 +1,78 @@
...
+            if tid < size:
+                i = cuda.grid(1)
+
+                # declare an array in shared memory
+                shr = cuda.shared.array(nelem, int32)
+                shr[tid] = data[i]
+
+                # make sure every thread has written
+                # its value to shared memory
+                # before we start reducing
```
gmarkall, line 52:
```suggestion
                # Ensure writes to shared memory are visible to all threads before reducing
```

numba/cuda/tests/doc_examples/test_reduction.py
```diff
@@ -0,0 +1,78 @@
...
+                cuda.syncthreads()
+
+                s = 1
+                while s < cuda.blockDim.x:
+                    if tid % (2 * s) == 0:
+                        # stride by `s` and add
```
gmarkall, line 58:
```suggestion
                        # Stride by `s` and add
```

numba/cuda/tests/doc_examples/test_reduction.py
```diff
@@ -0,0 +1,78 @@
...
+                        shr[tid] += shr[tid + s]
+                    s *= 2
+                    cuda.syncthreads()
+
+                # after the loop, the zeroth
+                # element contains the sum
```
gmarkall, line 64:
```suggestion
                # After the loop, the zeroth  element contains the sum
```

numba/cuda/tests/doc_examples/test_reduction.py
```diff
@@ -0,0 +1,78 @@
...
+                if tid == 0:
+                    data[tid] = shr[tid]
+
+        # ex_reduction.kernel.end
+
+        # ex_reduction.launch.begin
+        array_sum.forall(len(a))(a)
+        print(a[0])  # array(523776)
```
gmarkall, line 72: Why does that print `array(...)`?

numba/cuda/tests/doc_examples/test_sessionize.py
```diff
@@ -0,0 +1,119 @@
...
+        # No exception type, value, or traceback
+        self._captured_stdout.__exit__(None, None, None)
+        super().tearDown()
+
+    def test_ex_sessionize(self):
+        # ex_sessionize.import.begin
+        import numpy as np
+        from numba import cuda
+
+        # set the timeout to one hour
```
gmarkall, line 29:
```suggestion
        # Set the timeout to one hour
```

numba/cuda/tests/doc_examples/test_sessionize.py
```diff
@@ -0,0 +1,119 @@
...
+        session_timeout = np.int64(np.timedelta64("3600", "s"))
+        # ex_sessionize.import.end
+
+        # ex_sessionize.allocate.begin
+        # generate data
```
gmarkall, line 34:
```suggestion
        # Generate data
```

numba/cuda/tests/doc_examples/test_sessionize.py
```diff
@@ -0,0 +1,119 @@
...
+                    1, 2, 3, 5000, 5001, 5002, 1,
+                    2, 3, 1, 2, 5000, 5001, 10000,
+                    10001, 10002, 10003, 15000, 150001,
+                    1, 5000, 50001, 15000, 20000,
+                    25000, 25001, 25002, 25003,
+                ],
+                dtype="datetime64[ns]",
+            ).astype(
+                "int64"
+            )  # cast to int64 for compatibility
```
gmarkall, line 57:
```suggestion
            )  # Cast to int64 for compatibility
```

numba/cuda/tests/doc_examples/test_sessionize.py
```diff
@@ -0,0 +1,119 @@
...
+        )
+        # create a vector to hold the results
```
gmarkall, line 59:
```suggestion
        # Create a vector to hold the results
```

numba/cuda/tests/doc_examples/test_cpu_gpu_compat.py
```diff
@@ -0,0 +1,74 @@
...
+        print(business_logic(1, 2, 3))  # -126.79644737231007
+        # ex_cpu_gpu_compat.usecpu.end
+
+        # ex_cpu_gpu_compat.usegpu.begin
+        @cuda.jit
+        def f(res, xarr, yarr, zarr):
+            tid = cuda.grid(1)
+            if tid < len(xarr):
+                # the function decorated with numba.jit
+                # may be directly reused
```
gmarkall, line 61:
```suggestion
                # The function decorated with numba.jit may be directly reused
```

gmarkall (changes requested):
Many thanks for the updates! The majority of comments are small stylistic issues - some prevent blank lines appearing in the code excerpts in the docs. Other things to consider are:

- It looks like you gave my suggestion for another way of writing some of the sessionize logic a thumbs up, but the new test file still contains the old code. Did you want to change it?
- The Monte Carlo integration code still needs moving into a test file
- There is a portion of the shared memory example that doesn't appear in the docs - it does have magic markers already, but I guess there is no reference to it in the source.

numba/cuda/tests/doc_examples/test_reduction.py
brandon-b-miller: Nice catch, this was left over from what `a` was a cupy array. I updated this.

numba/cuda/tests/doc_examples/test_reduction.py
brandon-b-miller: Fixed this

gmarkall:
gpuci run tests

numba/cuda/tests/doc_examples/test_sessionize.py
```diff
@@ -0,0 +1,119 @@
...
+        # ex_sessionize.kernel.begin
+        @cuda.jit
+        def sessionize(user_id, timestamp, results):
+            gid = cuda.grid(1)
+            size = len(user_id)
+
+            if gid >= size:
+                return
+
+            ### DETERMINE SESSION BOUNDARIES
```
gmarkall, line 72:
```suggestion
            # Determine session boundaries
```

numba/cuda/tests/doc_examples/test_sessionize.py
```diff
@@ -0,0 +1,119 @@
...
+                new_user = user_id[gid] != user_id[gid - 1]
+                timed_out = (
+                    timestamp[gid] - timestamp[gid - 1] > session_timeout
+                )
+                is_sess_boundary = new_user or timed_out
+            else:
+                is_sess_boundary = True
+
+            ### DETERMINE SESSION LABELS
+            # this thread marks the start of a session
```
gmarkall, line 84:
```suggestion
            # Determine session labels
```

numba/cuda/tests/doc_examples/test_sessionize.py
```diff
@@ -0,0 +1,119 @@
...
+            if gid < size:
+                if is_sess_boundary:
```
gmarkall, line 86:
```suggestion
                if is_sess_boundary:
                # This thread marks the start of a session
```

numba/cuda/tests/doc_examples/test_sessionize.py
```diff
@@ -0,0 +1,119 @@
...
+                    results[gid] = gid
+
+                    # make sure all session boundaries are written
```
gmarkall, line 89:
```suggestion
                    # Make sure all session boundaries are written
```

numba/cuda/tests/doc_examples/test_sessionize.py
```diff
@@ -0,0 +1,119 @@
...
+                    # before populating the session id
+                    grid = cuda.cg.this_grid()
+                    grid.sync()
+
+                    look_ahead = 1
+                    # check elements 'forward' of this one
```
gmarkall, line 95:
```suggestion
                    # Check elements 'forward' of this one
```

numba/cuda/tests/doc_examples/test_sessionize.py
```diff
@@ -0,0 +1,119 @@
...
+                    # until a new session boundary is found
+                    while results[gid + look_ahead] == 0:
+                        results[gid + look_ahead] = gid
+                        look_ahead += 1
+                        # don't segfault if I'm the last thread
```
gmarkall, line 100:
```suggestion
                        # Avoid out-of-bounds accesses by the last thread
```

gmarkall (changes requested):
Many thanks for the updates - there's a few more formatting comments for the code on the diff. I think the text is looking good, so there are a couple of points about the code:

* Some of the CI is failing due to a flake8 error.
* The test files check that the examples run, but not that their output is correct - could you add some checks for the correctness of the output to them please? (e.g. `test_cg.py` checking the output: https://github.com/numba/numba/blob/main/numba/cuda/tests/doc_examples/test_cg.py#L71)

brandon-b-miller:
👍  I'll update these with some assertions that actually test the numerical values.

brandon-b-miller:
The only one I didn't add an actual numerical tests for is the laplace example. I can do this one like the others but it feels possibly weird to hardcode the results of a simulation. I don't have any better ideas though. What do you think @gmarkall

numba/cuda/tests/doc_examples/test_montecarlo.py
```diff
@@ -0,0 +1,105 @@
...
+        def func(x):
+            return 1.0 / x
+
+        mc_integrate(1, 2, nsamps)  # array(0.6929643, dtype=float32)
+        mc_integrate(2, 3, nsamps)  # array(0.4054021, dtype=float32)
+        # ex_montecarlo.launch.end
+
+        # values computed independently using maple
+        assert np.isclose(mc_integrate(1, 2, nsamps), 0.69315, atol=0.001)
+        assert np.isclose(mc_integrate(2, 3, nsamps), 0.4055, atol=0.001)
```
gmarkall, line 101: Can these use `np.testing.assert_allclose` instead of `isclose` with a bare assert please?

numba/cuda/tests/doc_examples/test_reduction.py
```diff
@@ -0,0 +1,76 @@
...
+                    data[tid] = shr[tid]
+        # ex_reduction.kernel.end
+
+        # ex_reduction.launch.begin
+        array_sum.forall(len(a))(a)
+        print(a[0])  # 523776
+        sum(np.arange(1024))  # 523776
+        # ex_reduction.launch.end
+
+        assert a[0] == sum(np.arange(1024))
```
gmarkall, line 72: Can this use `np.testing.assert_equal` instead please?

numba/cuda/tests/doc_examples/test_sessionize.py
```diff
@@ -0,0 +1,126 @@
...
+                new_user = user_id[gid] != user_id[gid - 1]
+                timed_out = (
+                    timestamp[gid] - timestamp[gid - 1] > session_timeout
+                )
+                is_sess_boundary = new_user or timed_out
+            else:
+                is_sess_boundary = True
+
+            # Determine session labels
+            if gid < size:
```
gmarkall, line 84: I think this must always be true because of the `if gid >= size: return` above - can this conditional be removed?

docs/source/cuda/examples.rst
```diff
@@ -3,6 +3,410 @@
...
+   :dedent: 8
+   :linenos:
+
+The initial state of the problem can be visualized as:
+
+.. image:: laplace_initial.svg
+
+In our kernel each thread will be responsible for managing the temperature update for a single element
+in a loop over the desired number of timesteps. The kernel is below. Note the use of cooperative group
+synchronization and the temporary array used to avoid race conditions. See
```
gmarkall, line 141:
```suggestion
synchronization and the use of two buffers swapped at each iteration to avoid race conditions. See 
```

gmarkall (changes requested):
Many thanks for the updates again. There are some small suggestions on the diff.

With respect to the Laplace example, I have a commit that fixes some issues in it and adds a check: https://github.com/gmarkall/numba/commit/e57cb26046943b901fbcc6214b88f35b6604b899 - see the commit for details. Please do feel free to pull it in if you agree with the suggestions.

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall (approved):
Many thanks!

gmarkall:
@stuartarchibald / @esc Could this have a CUDA buildfarm run please?

esc:
`Build numba_smoketest_cuda_yaml_127 has started`

gmarkall:
gpuci run tests

gmarkall:
@esc could this have another buildfarm run please?

esc:
> @esc could this have another buildfarm run please?

`Build numba_smoketest_cuda_yaml_128 has started`

esc:
> > @esc could this have another buildfarm run please?
> 
> `Build numba_smoketest_cuda_yaml_128 has started`

all green 💚

gmarkall:
> all green green_heart

Many thanks!

sklam (approved):

PR #7919, opened 2022-03-16
Disallow //= reductions in pranges.

### LINKED ISSUE #7867: Prange incorrectly compute `ifloordiv` reduction
MWR:
```python
import numpy as np

@njit(parallel=True)
def foo():
    acc = 10000
    for i in prange(4):
        acc //= 2
    return acc

print(foo())
print(foo.py_func())
```

_Originally posted by @stuartarchibald in https://github.com/numba/numba/pull/7837#discussion_r812944548_

DrTodd13:
Resolves #7867

Turns out that //= reductions of real divisors is not parallelizable.  //= reductions of integer divisors is parallelizable with some difficulty.  We just say that we don't support them at all but when present we raise an exception that suggests integer divisors can be re-written as a product reduction followed by one floor div after the prange.


numba/parfors/parfor.py
```diff
@@ -3611,13 +3611,18 @@ def get_reduction_init(nodes):
...
     if (  acc_expr.fn == operator.imul
-       or acc_expr.fn == operator.itruediv
-       or acc_expr.fn == operator.ifloordiv ):
+       or acc_expr.fn == operator.itruediv ):
         return 1, acc_expr.fn
     return None, None
 
 def supported_reduction(x, func_ir):
     if x.op == 'inplace_binop' or x.op == 'binop':
+        if x.fn == operator.ifloordiv or x.fn == operator.floordiv:
```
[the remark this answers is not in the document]
DrTodd13: Done.

numba/parfors/parfor.py
```diff
@@ -3611,14 +3611,27 @@ def get_reduction_init(nodes):
...
-       or acc_expr.fn == operator.ifloordiv ):
+       or acc_expr.fn == operator.itruediv ):
         return 1, acc_expr.fn
     return None, None
 
 def supported_reduction(x, func_ir):
     if x.op == 'inplace_binop' or x.op == 'binop':
-        return True
+        if x.fn == operator.ifloordiv or x.fn == operator.floordiv:
+            raise ValueError(("Parallel floordiv reductions are not supported. "
```
stuartarchibald, line 3621:
```suggestion
            raise errors.NumbaValueError(("Parallel floordiv reductions are not supported. "
```
(and fix line wrap)

numba/parfors/parfor.py
```diff
@@ -3611,14 +3611,27 @@ def get_reduction_init(nodes):
...
+                              "If all divisors are integers then a floordiv "
+                              "reduction can in some cases be parallelized as "
+                              "a multiply reduction followed by a floordiv of "
+                              "the resulting product."))
```
stuartarchibald, line 3625:
```suggestion
                              "the resulting product."), x.loc)
```
Add in the location such that the code causing the issue is highlighted?

docs/source/user/parallel.rst
```diff
@@ -82,7 +82,12 @@ supported reductions.
 A reduction is inferred automatically if a variable is updated by a binary
 function/operator using its previous value in the loop body. The initial value
 of the reduction is inferred automatically for the ``+=``, ``-=``,  ``*=``,
-and ``/=`` operators.
```
stuartarchibald, line 85: Should this section explicitly declare which reductions are supported, I think the opening statement is a bit too general given what is available in practice?

numba/parfors/parfor.py
```diff
@@ -3611,14 +3611,27 @@ def get_reduction_init(nodes):
...
+        supps = [operator.iadd,
+                 operator.isub,
+                 operator.imul,
+                 operator.itruediv,
+                 operator.add,
+                 operator.sub,
+                 operator.mul,
+                 operator.truediv]
+        return x.fn in supps
```
stuartarchibald, line 3634: I think it would be good to update the `get_reduce_nodes` function to raise a `NumbaValueError` so that the exception handling is consistent. Would also be good to add a test for a reduction like `acc ^= True` or similar to ensure that this branch runs and is tested.

stuartarchibald (commented):
Thanks for the patch. There's a few comments to address but otherwise looks good. Thanks for adding this to improve correctness/the user experience around parallel reductions.

docs/source/user/parallel.rst
DrTodd13: Done.

numba/parfors/parfor.py
DrTodd13: I added the ^= test.  For the specific error to raise, there are numerous places in parfors.py where ValueError is raised and none where NumbaValueError is raised.  Do you want all those places changed in this PR or defer to a separate PR?

numba/tests/test_parfors.py
```diff
@@ -1395,6 +1395,35 @@ def test_impl():
...
+                acc //= 2
+            return acc
+
+        # checks that invalid use of ifloordiv reduction operator is detected
+        msg = ("Parallel floordiv reductions are not supported. "
+               "If all divisors are integers then a floordiv "
+               "reduction can in some cases be parallelized as "
+               "a multiply reduction followed by a floordiv of "
+               "the resulting product.")
+        with self.assertRaises(ValueError) as e:
```
stuartarchibald, line 1411: This probably needs to check for a `errors.NumbaValueError`?

numba/tests/test_parfors.py
```diff
@@ -1395,6 +1395,35 @@ def test_impl():
...
+
+    def test_unsupported_xor1(self):
+        def test_impl():
+            acc = 100
+            for i in prange(2):
+                acc ^= i + 2
+            return acc
+
+        msg = ("Use of reduction variable acc in an unsupported reduction function.")
+        with self.assertRaises(ValueError) as e:
```
stuartarchibald, line 1423: This probably needs to check for a `errors.NumbaValueError`?

stuartarchibald (commented):
Thanks for the update. Think the CI build is failing because the unit tests are looking for standard Python error classes opposed to Numba versions. I've suggested inline where this might be the case. Thanks again.

numba/parfors/parfor.py
stuartarchibald: Suggest fixing these couple of cases in this PR and then deferring the rest to a separate one. How does that sound?

numba/tests/test_parfors.py
DrTodd13: This one needed it.  Done.

PR #7924, opened 2022-03-19
Retain non-modified index tuple components.

### LINKED ISSUE #7843: Incorrect result when numpy broadcasting in parallel
I stumbled into an incorrect behavior with jit parallel=True when writing into a numpy array using broadcasting.
I reduced my code to the following minimal example to demonstrate the incorrect behavior:
```python
import numpy as np
from numba import jit
N=3
nxy=np.zeros((N,1))
@jit(nopython=True,parallel=True)
def example(nxy):
    count=0
    for n in range(nxy.shape[0]):
        #Pointless if statement that does nothing. If I remove it, it works ok.
        if n:
            n
        #writting to array:
        nxy[count,:]=1
        print(nxy[count,:]) #nxy[count,:] should contain [1.]
        count+=1
example(nxy)
```
The expected result would be:
```
[1.]
[1.]
[1.]
```
however the example prints:
```
[1.]
[0.]
[0.]
```
I understand that the example seems weird with having a counter in a for loop, but it is only meant to demonstrate an issue.

If I run it non-jitted or jitted with parallel=False it works ok.
If I remove the pointless if statement, it works ok.
If I replace `nxy[count,:]=1` with `nxy[count,...]=1`  it works ok, because that part then does not seem to get parallelized. I am currently using that as my solution, because I get parallelization elsewhere, however, it makes me uneasy to not know why it behaves incorrectly in the demonstrated case.

DrTodd13:
Resolves #7843

If an index tuple needs to be partially reconstructed by array analysis, then for the unmodified parts, use the original variables used in the tuple.

### DIFF
--- a/numba/parfors/array_analysis.py
+++ b/numba/parfors/array_analysis.py
@@ -2056,6 +2056,7 @@ class ArrayAnalysis(object):
         var_shape = equiv_set._get_shape(var)
         if isinstance(ind_typ, types.SliceType):
             seq_typs = (ind_typ,)
+            seq = (ind_var,)
         else:
             require(isinstance(ind_typ, types.BaseTuple))
             seq, op = find_build_sequence(self.func_ir, ind_var)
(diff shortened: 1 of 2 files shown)


sklam (approved):
LGTM.

PR #7938, opened 2022-03-25
array_equal new impl

### LINKED ISSUE #7937: enhancement for `array_equal`
---

## Feature request

I find current `array_equal` doesn't support `equal_nan` argument, so I made an implementation as follows:
```python
@njit
def eq1(a, b, equal_nan=False):
    a = np.asarray(a)
    b = np.asarray(b)
    if a.shape != b.shape:
        return False
    for _a, _b in zip(a.flat, b.flat):
        if _a != _b:
            if equal_nan and _a != _a and _b != _b:
                continue
            else:
                return False
    return True

@njit
def eq2(a, b):
    return np.array_equal(a, b)

eq3 = np.array_equal
```

Besides, I find my implementation is both faster than original numba implemention and numpy's:

```python
# worst case: all equal
a, b = np.ones((2, 1000))

%timeit -n 10000 _ = eq1(a, b)
# 2.1 µs ± 77.3 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

%timeit -n 10000 _ = eq2(a, b)
# 2.6 µs ± 60.5 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

%timeit -n 10000 _ = eq3(a, b)
# 4.14 µs ± 969 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

# best case: all non-equal
c = a + 1

%timeit -n 10000 _ = eq1(a, c)
# 658 ns ± 12.8 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

%timeit -n 10000 _ = eq2(a, c)
# 1.05 µs ± 22.4 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

%timeit -n 10000 _ = eq3(a, c)
# 5.25 µs ± 140 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```
FYI: numpy impl:

https://github.com/numpy/numpy/blob/4eb53425ecdad412411d6d174064405f511494eb/numpy/core/numeric.py#L2396-L2464

In my impl, I use a loop instead of `np.all(a==b)`, which may help reducing memory alloc and can return early. 

I can make a PR If you think this new impl is OK.

auderson:
closes #7937

### DIFF
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -3626,23 +3626,41 @@ def np_diff_impl(a, n=1):
 
 
 @overload(np.array_equal)
-def np_array_equal(a, b):
+def np_array_equal(a, b, equal_nan=False):
 
     if not (type_can_asarray(a) and type_can_asarray(b)):
         raise TypingError('Both arguments to "array_equals" must be array-like')
 
+    if not isinstance(equal_nan, types.Boolean):
+        raise TypingError('"equal_nan" must be a boolean')
+
     accepted = (types.Boolean, types.Number)
     if isinstance(a, accepted) and isinstance(b, accepted):
         # special case
-        def impl(a, b):
+        def impl(a, b, equal_nan=False):
+            if equal_nan:
+                if a != a and b != b:
+                    return True
             return a == b
     else:
-        def impl(a, b):
+        def impl(a, b, equal_nan=False):
             a = np.asarray(a)
             b = np.asarray(b)
-            if a.shape == b.shape:
-                return np.all(a == b)
-            return False
+            if a.shape != b.shape:
+                return False
+            if equal_nan:
+                for _a, _b in zip(a.flat, b.flat):
+                    if _a != _b:
+                        # NaN != NaN
+                        if _a != _a and _b != _b:
+                            continue
+                        else:
+                            return False
+            else:
+                for _a, _b in zip(a.flat, b.flat):
+                    if _a != _b:
+                        return False
+            return True
 
     return impl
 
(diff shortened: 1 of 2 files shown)


auderson:
I have to point out that for numpy versions (pre 1.19) without equal_nan option, the impl should also work, but it can't pass numba's signature check in overload.

 So I wrote 2 impl (with & without equal_nan) for these numpy versions.

numba/np/arraymath.py
```diff
@@ -3625,26 +3625,71 @@ def diff_impl(a, n=1):
...
+                return a == b
+        else:
+            def impl(a, b):
+                a = np.asarray(a)
+                b = np.asarray(b)
+                if a.shape != b.shape:
+                    return False
+                for _a, _b in zip(a.flat, b.flat):
+                    if _a != _b:
+                        return False
```
sklam, line 3689: You can use [@register_jittable](https://github.com/numba/numba/blob/61ec1fd0f69aeadece218dccf4c39ebc5c7dfbc4/numba/core/extending.py#L138) to avoid duplicating. The two versions can share this loop; example:

```python
@register_jitable
def _array_equal_simple(a, b):
    for _a, _b in zip(a.flat, b.flat):
        if _a != _b:
            return False
```
so this `impl` can just be:

```python
 def impl(a, b):
    a = np.asarray(a)
    b = np.asarray(b)
    if a.shape != b.shape:
        return False
    return _array_equal_simple(a, b)
```

sklam (commented):
Thanks for the patch!

I think the body of the overload can share more to reduce code duplication

numba/np/arraymath.py
auderson: updated

PR #7940, opened 2022-03-26
Implemented `np.allclose` in `numba/np/arraymath.py`

czgdp1807:
https://github.com/numba/numba/issues/4074

I am implementing `np.allclose` in `numba/np/arraymath.py` as I noticed that it hasn't been added yet and I was unable to find any open PR already working on it. Please let me know if its already there or someone recently is already working on it, I will close the PR then.

I am working on adding unit test and therefore I added `[skip ci]` in my commit so that my PR doesn't consume CI resources unnecessarily. Meanwhile if you have any comments/reviews for the implementation, please let me know. Thanks.


czgdp1807:
Seems like `[skip ci]` didn't work.

numba/np/arraymath.py
```diff
@@ -864,6 +864,56 @@ def flat_all(a):
...
+    if not isinstance(atol, types.Float):
+        raise TypingError('The fifth argument "atol" must be a '
+                          'floating point')
+
+    if not isinstance(equal_nan, types.Boolean):
+        raise TypeError('The sixth argument "equal_nan" must be a '
+                        'boolean')
+
+    def np_allclose_impl(a, b, rtol=1e-05, atol=1e-08, equal_nan=False):
+        for av, bv in zip(np.nditer(a), np.nditer(b)):
```
czgdp1807, line 890: May be we need to broadcast `a` or `b` if needed. I am yet to do that.

numba/tests/test_np_functions.py
```diff
@@ -3172,6 +3176,33 @@ def test_1D_weights_axis(data, axis, weights):
...
+        b = np.asarray([1.0001e10, 1e-9])
+        assert not cfunc(a, b)
+
+        a = np.asarray([1.0, np.nan])
+        b = np.asarray([1.0, np.nan])
+        assert not cfunc(a, b)
+        assert cfunc(a, b, equal_nan=True)
+
+        b = np.asarray([np.nan, 1.0])
+        assert not cfunc(a, b)
```
czgdp1807, line 3203: I will be adding stronger tests using `np.random.ranf`.

numba/np/arraymath.py
```diff
@@ -864,6 +864,98 @@ def flat_all(a):
...
+        a = np.asarray(a)
+        b = np.asarray(b)
+        if a.shape != b.shape:
+            if _can_broadcast_returns_bool(a, b.shape):
+                a = np.broadcast_to(a, b.shape)
+            elif _can_broadcast_returns_bool(b, a.shape):
+                b = np.broadcast_to(b, a.shape)
+            else:
+                raise ValueError('operands could not be broadcast together '
+                                 'with remapped shapes')
```
czgdp1807, line 930: The broadcasting here fails if one of `a` or `b` is a scalar. The reason being, `.shape` will be `()`(an empty `UniTuple`) and it will fail while JITing the code.

numba/np/arraymath.py
czgdp1807: This doesn't work when `a` and `b` have different ranks because rank is a part of array type which gives an error like, `Failed to unify Array(float64, 2d, c) and Array(float64, 0d, c)`.  @sklam Any ideas for a workaround for the above problem?

czgdp1807:
@guilhermeleobas Could you please review this PR and provide your thoughts/opinions on the above comments? Thanks.

numba/np/arraymath.py
guilhermeleobas: As you notice, one cannot call `np.broadcast_to` with an empty shape. In the case one of the arguments is an integer, you can iterate over the array and check if they're the same:

```python
if isinstance(b, types.Integer):
    def impl(a, b, rtol=1e-05, atol=1e-08, equal_nan=False):
        for av in np.nditer(a):
            if not allclose_inner(av, b, rtol, atol, equal_nan):
                return False
            return True
    return impl
else:
    def np_allclose_impl(a, b, rtol=1e-05, atol=1e-08, equal_nan=False):
        ...
    return np_allclose_impl
```

You can replicate the idea above for the cases where
* both `a` and `b` are scalars
* either `a` or `b` is scalar
* both `a` and `b` are arrays

numba/np/arraymath.py
guilhermeleobas: Do you have an input that replicate this error?

numba/tests/test_np_functions.py
```diff
@@ -3172,6 +3176,66 @@ def test_1D_weights_axis(data, axis, weights):
...
+
+        a = np.asarray([1e10])
+        b = np.asarray([1.0001e10, 1e-9])
+        assert not cfunc(a, b)
+
+        a = np.asarray([1e10, 1e-9, np.nan])
+        b = np.asarray([1.0001e10, 1e-9])
+        with self.assertRaises(ValueError):
+            cfunc(a, b)
```
guilhermeleobas, line 3203: Move the test that raises an exception to its own function. If I'm not wrong, one need to disable the leak check when asserting exceptions:

```python

def test_allcase_exception(self):
    self.disable_leak_check()
    a = np.asarray([1e10, 1e-9, np.nan])
    b = np.asarray([1.0001e10, 1e-9])
    with self.assertRaises(ValueError) as e:
        cfunc(a, b)
    self.assertIn("put here the expected exception msg", str(e.exception))
```

numba/tests/test_np_functions.py
```diff
@@ -3172,6 +3176,66 @@ def test_1D_weights_axis(data, axis, weights):
...
+        zero_array = np.zeros((25,))
+        a = np.random.ranf((25, 4))
+        for noise in noise_levels:
+            for rtol in noise_levels:
+                for atol in noise_levels:
+                    py_result = pyfunc(zero_array, np.asarray([noise]),
+                                       atol=atol, rtol=rtol)
+                    c_result = cfunc(zero_array, np.asarray([noise]),
+                                     atol=atol, rtol=rtol)
+                    assert py_result == c_result
```
guilhermeleobas, line 3223: Replace the `assert` by `self.assertEqual(py_result, c_result)`

numba/tests/test_np_functions.py
```diff
@@ -3172,6 +3176,66 @@ def test_1D_weights_axis(data, axis, weights):
...
+
+                    py_result = pyfunc(np.asarray([noise]), zero_array,
+                                       atol=atol, rtol=rtol)
+                    c_result = cfunc(np.asarray([noise]), zero_array,
+                                     atol=atol, rtol=rtol)
+                    assert py_result == c_result
```
guilhermeleobas, line 3229: same.

numba/tests/test_np_functions.py
```diff
@@ -3172,6 +3176,66 @@ def test_1D_weights_axis(data, axis, weights):
...
+
+                    py_result = pyfunc(a, a + noise, atol=atol, rtol=rtol)
+                    c_result = cfunc(a, a + noise, atol=atol, rtol=rtol)
+                    assert py_result == c_result
```
guilhermeleobas, line 3233: same.

numba/tests/test_np_functions.py
```diff
@@ -3172,6 +3176,66 @@ def test_1D_weights_axis(data, axis, weights):
...
+
+                    py_result = pyfunc(a + noise, a, atol=atol, rtol=rtol)
+                    c_result = cfunc(a + noise, a, atol=atol, rtol=rtol)
+                    assert py_result == c_result
```
guilhermeleobas, line 3237: same.

numba/tests/test_np_functions.py
```diff
@@ -3172,6 +3176,66 @@ def test_1D_weights_axis(data, axis, weights):
...
+
+        a = np.asarray([1e10, 1e-9, np.nan])
+        b = np.asarray([1.0001e10, 1e-9])
+        with self.assertRaises(ValueError):
+            cfunc(a, b)
+
+        a = np.asarray([1.0, np.nan])
+        b = np.asarray([1.0, np.nan])
+        assert not cfunc(a, b)
+        assert cfunc(a, b, equal_nan=True)
```
guilhermeleobas, line 3208: Replace the `assert` by `self.assertEqual`

numba/tests/test_np_functions.py
```diff
@@ -3172,6 +3176,66 @@ def test_1D_weights_axis(data, axis, weights):
...
+        pyfunc = np_allclose
+        cfunc = jit(nopython=True)(pyfunc)
+
+        a = np.asarray([1e10, 1e-7])
+        b = np.asarray([1.00001e10, 1e-8])
+        assert not cfunc(a, b)
+
+        a = np.asarray([1e10, 1e-8])
+        b = np.asarray([1.00001e10, 1e-9])
+        assert cfunc(a, b)
```
guilhermeleobas, line 3190: Replace the `assert` by `self.assertEqual`.

numba/tests/test_np_functions.py
```diff
@@ -3172,6 +3176,66 @@ def test_1D_weights_axis(data, axis, weights):
...
+
+        a = np.asarray([1e10, 1e-8])
+        b = np.asarray([1.0001e10, 1e-9])
+        assert not cfunc(a, b)
```
guilhermeleobas, line 3194: Replace the `assert` by `self.assertEqual`.

numba/tests/test_np_functions.py
```diff
@@ -3172,6 +3176,66 @@ def test_1D_weights_axis(data, axis, weights):
...
+
+        a = np.asarray([1e10])
+        b = np.asarray([1.0001e10, 1e-9])
+        assert not cfunc(a, b)
```
guilhermeleobas, line 3198: Replace the `assert` by `self.assertEqual`.

guilhermeleobas (changes requested):
Hi @czgdp1807, thanks for the contribution to the Numba codebase. This is a nice addition. I've made a few comments regarding the usage of `assert` and answered your questions on how you can handle mixed input types (i.e. array and int).

Also, you can copy the inputs NumPy uses to test `np.allclose` to make sure we're covering all cases.
https://github.com/numpy/numpy/blob/aac965af6032b69d5cb515ad785cc9a331e816f4/numpy/core/tests/test_numeric.py#L2203-L2285

numba/np/arraymath.py
czgdp1807: **Code**

```python
from numba import jit
import numpy as np

def np_allclose(a, b, rtol=1e-05, atol=1e-08, equal_nan=False):
    return np.allclose(a, b, rtol, atol, equal_nan)

def test_allclose_1():

    pyfunc = np_allclose

    zero_array = np.zeros((25, 4))
    b = np.asarray([1e-6])
    print("Shapes: ", zero_array.shape, b.shape)
    print(cfunc(zero_array, b))

test_allclose_1()
```

**Error**

```python
Shapes:  (25, 4) (1,)
  File "../debug.py", line 75, in <module>
    test_allclose_1()
  File "../debug.py", line 73, in test_allclose_1
    print(cfunc(zero_array, b))
  File "/Users/czgdp1807/numba_project/numba/numba/core/dispatcher.py", line 468, in _compile_for_args
    error_rewrite(e, 'typing')
  File "/Users/czgdp1807/numba_project/numba/numba/core/dispatcher.py", line 409, in error_rewrite
    raise e.with_traceback(None)
No implementation of function Function(<function allclose at 0x7f7e9072f560>) found for signature:
 
 >>> allclose(array(float64, 2d, C), array(float64, 1d, C), float64, float64, Literal[bool](False))
 
  Overload in function 'np_allclose': File: numba/np/arraymath.py: Line 898.
    With argument(s): '(array(float64, 2d, C), array(float64, 1d, C), float64, float64, bool)':
   Cannot unify array(float64, 1d, C) and readonly array(float64, 2d, A) for 'b.1.2', defined at /Users/czgdp1807/numba_project/numba/numba/np/arraymath.py (932)
   
   File "numba/np/arraymath.py", line 932:
       def np_allclose_impl(a, b, rtol=1e-05, atol=1e-08, equal_nan=False):
           <source elided>
   
           for av, bv in zip(np.nditer(a), np.nditer(b)):
           ^
   
   During: typing of assignment at /Users/czgdp1807/numba_project/numba/numba/np/arraymath.py (932)
   
   File "numba/np/arraymath.py", line 932:
       def np_allclose_impl(a, b, rtol=1e-05, atol=1e-08, equal_nan=False):
           <source elided>
   
           for av, bv in zip(np.nditer(a), np.nditer(b)):
           ^

  raised from /Users/czgdp1807/numba_project/numba/numba/core/typeinfer.py:1086

During: resolving callee type: Function(<function allclose at 0x7f7e9072f560>)
During: typing of call at ../debug.py (6)


File "../debug.py", line 6:
def np_allclose(a, b, rtol=1e-05, atol=1e-08, equal_nan=False):
    return np.allclose(a, b, rtol, atol, equal_nan)
    ^
```

numba/np/arraymath.py
czgdp1807: This sounds like a nice approach. Thanks for the idea.

numba/tests/test_np_functions.py
```diff
@@ -3172,6 +3176,66 @@ def test_1D_weights_axis(data, axis, weights):
...
         test_1D_weights_axis(data, axis=1, weights=w)
 
+    def test_allclose(self):
+
+        pyfunc = np_allclose
+        cfunc = jit(nopython=True)(pyfunc)
+
+        a = np.asarray([1e10, 1e-7])
+        b = np.asarray([1.00001e10, 1e-8])
+        assert not cfunc(a, b)
```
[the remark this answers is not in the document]
czgdp1807: I though that since, output of `cfunc` and `pyfunc` is a `bool` object so using `assert` should be fine. Though I am fine with `self.assertEqual` as well. Will do the replacements.

czgdp1807:
@guilhermeleobas I have addressed the reviews and responded to the questions.

numba/np/arraymath.py
czgdp1807: I tried to workaround this problem in https://github.com/numba/numba/pull/7940/commits/f09d8ffd2555e59a70ec0618b1532e8d95511ea2 and it seems to work locally. `a_b`/`b_b` both have updates types (with correct ranks in the type), so `np_allclose_impl_no_broadcast` receives `a` and `b` parameters of the same shape and hence, `np.nditer((a, b))` works without issues.

czgdp1807: I used recently added `np.broadcast_arrays` functionality, so this conversation might not be relevant anymore.

numba/np/arraymath.py
```diff
@@ -864,6 +864,100 @@ def flat_all(a):
...
+    elif is_a_scalar and not is_b_scalar:
+        def np_allclose_impl_scalar_array(a, b, rtol=1e-05, atol=1e-08,
+                                          equal_nan=False):
+            b = np.asarray(b)
+            for bv in np.nditer(b):
+                if not _allclose_scalars(a, bv.item(), rtol=rtol, atol=atol,
+                                         equal_nan=equal_nan):
+                    return False
+            return True
+        return np_allclose_impl_scalar_array
```
guilhermeleobas, line 933: You can replace this block of code by:
```suggestion
    elif is_a_scalar and not is_b_scalar:
        def np_allclose_impl_scalar_array(a, b, rtol=1e-05, atol=1e-08,
                                          equal_nan=False):
            return np.allclose(b, a, rtol=rtol, atol=atol, equal_nan=equal_nan)
        return np_allclose_impl_scalar_array
```

numba/tests/test_np_functions.py
```diff
@@ -3172,6 +3176,126 @@ def test_1D_weights_axis(data, axis, weights):
...
+
+        a = np.asarray([1.0, np.nan])
+        b = np.asarray([1.0, np.nan])
+        self.assertFalse(cfunc(a, b))
+        self.assertTrue(cfunc(a, b, equal_nan=True))
+
+        b = np.asarray([np.nan, 1.0])
+        self.assertFalse(cfunc(a, b))
+
+        # NumPy test data
```
guilhermeleobas, line 3210: Can you also include the link to NumPy repository pointing to the tests?

guilhermeleobas, line 3210: Maybe split the tests into different functions. This makes it easy to debug when there's a test failure.

numba/tests/test_np_functions.py
```diff
@@ -3172,6 +3176,126 @@ def test_1D_weights_axis(data, axis, weights):
...
+
+                    py_result = pyfunc(a, a + noise, atol=atol, rtol=rtol)
+                    c_result = cfunc(a, a + noise, atol=atol, rtol=rtol)
+                    self.assertEqual(py_result, c_result)
+
+                    py_result = pyfunc(a + noise, a, atol=atol, rtol=rtol)
+                    c_result = cfunc(a + noise, a, atol=atol, rtol=rtol)
+                    self.assertEqual(py_result, c_result)
+
+    def test_allcase_exception(self):
```
guilhermeleobas, line 3284:
```suggestion
    def test_allclose_exception(self):
```

guilhermeleobas (changes requested):
Good work @czgdp1807. I have a few more comments that should be easy to address.

numba/np/arraymath.py
czgdp1807: I don't think this will give right results all the time, because the check depends on the ordering of `a` and `b`. For `(a, b)` the check is `abs(a - b) > atol + rtol * abs(b)`, where as for `(b, a)` the check is `abs(b - a) > atol + rtol * abs(a)`. See the RHS in both the inequalities, one uses `b` and the other uses `a`. So, `np.allclose(a, b)` might give different result from `np.allclose(b, a)` for edge cases. 

Quoting from "Notes" section of https://numpy.org/doc/stable/reference/generated/numpy.allclose.html

> The above equation is not symmetric in a and b, so that `allclose(a, b)` might be different from `allclose(b, a)` in some rare cases.

Please let me know if should still apply the above code suggestion.

czgdp1807:
@guilhermeleobas @gmarkall I have addressed/responded to the new reviews and the tests pass. Please let me know if something else is to be done.

numba/np/arraymath.py
guilhermeleobas: That's fine. I wasn't aware of this info.

numba/tests/test_np_functions.py
```diff
@@ -3172,6 +3176,134 @@ def test_1D_weights_axis(data, axis, weights):
...
+                    c_result = cfunc(a, a + noise, atol=atol, rtol=rtol)
+                    self.assertEqual(py_result, c_result)
+
+                    py_result = pyfunc(a + noise, a, atol=atol, rtol=rtol)
+                    c_result = cfunc(a + noise, a, atol=atol, rtol=rtol)
+                    self.assertEqual(py_result, c_result)
+
+    def test_allclose_numpy_data(self):
+
+        pyfunc = np_allclose
```
guilhermeleobas, line 3244: Can you split the content of this test case into different functions? Similar to what NumPy does.

numba/tests/test_np_functions.py
```diff
@@ -3172,6 +3176,134 @@ def test_1D_weights_axis(data, axis, weights):
...
+            (np.asarray([1.0]), np.asarray([1 + rtol + atol])),
+            (arr, arr + arr * rtol),
+            (arr, arr + arr * rtol + atol * 2),
+            (aran, aran + aran * rtol),
+            (np.inf, np.inf),
+            (np.inf, np.asarray([np.inf]))
+        ]
+
+        for (x, y) in numpy_data:
+            self.assertTrue(cfunc(x, y))
```
guilhermeleobas, line 3267: Instead of asserting to `True`, compare the outputs of `cfunc(x, y)` and `pyfunc(x, y)` are equals.

numba/tests/test_np_functions.py
```diff
@@ -3172,6 +3176,134 @@ def test_1D_weights_axis(data, axis, weights):
...
+            (np.asarray([-np.inf, 0.0]), np.asarray([np.inf, 0.0])),
+            (np.asarray([np.nan, 0.0]), np.asarray([np.nan, 0.0])),
+            (np.asarray([atol * 2]), np.asarray([0.0])),
+            (np.asarray([1.0]), np.asarray([1 + rtol + atol * 2])),
+            (aran, aran + aran * atol + atol * 2),
+            (np.array([np.inf, 1.0]), np.array([0.0, np.inf]))
+        ]
+
+        for (x, y) in numpy_data:
+            self.assertFalse(cfunc(x, y))
```
guilhermeleobas, line 3283: Same here. Compare the outputs of `cfunc(x, y)` and `pyfunc(x, y)`.

numba/np/arraymath.py
```diff
@@ -864,6 +864,100 @@ def flat_all(a):
...
+def np_allclose(a, b, rtol=1e-05, atol=1e-08, equal_nan=False):
+
+    if not type_can_asarray(a):
+        raise TypeError('The first argument "a" must be array-like')
+
+    if not type_can_asarray(b):
+        raise TypeError('The first argument "b" must be array-like')
+
+    if not isinstance(rtol, types.Float):
+        raise TypeError('The fourth argument "rtol" must be a '
```
guilhermeleobas, line 904: `rtol` is the third argument, right?
```suggestion
        raise TypeError('The third argument "rtol" must be a '
```

numba/np/arraymath.py
```diff
@@ -864,6 +864,100 @@ def flat_all(a):
...
+                        'floating point')
+
+    if not isinstance(atol, types.Float):
+        raise TypingError('The fifth argument "atol" must be a '
```
guilhermeleobas, line 908:
```suggestion
        raise TypingError('The fourth argument "atol" must be a '
```

numba/np/arraymath.py
```diff
@@ -864,6 +864,100 @@ def flat_all(a):
...
+                          'floating point')
+
+    if not isinstance(equal_nan, types.Boolean):
+        raise TypeError('The sixth argument "equal_nan" must be a '
```
guilhermeleobas, line 912:
```suggestion
        raise TypeError('The fifth argument "equal_nan" must be a '
```

numba/tests/test_np_functions.py
```diff
@@ -3172,6 +3176,134 @@ def test_1D_weights_axis(data, axis, weights):
...
+
+        for a, b in simple_data:
+            py_result = pyfunc(a, b)
+            c_result = cfunc(a, b)
+            self.assertEqual(py_result, c_result)
+
+        a = np.asarray([1.0, np.nan])
+        b = np.asarray([1.0, np.nan])
+        self.assertFalse(cfunc(a, b))
+        self.assertTrue(cfunc(a, b, equal_nan=True))
```
guilhermeleobas, line 3205: Compare the output of `cfunc` with `pyfunc` instead.

numba/tests/test_np_functions.py
```diff
@@ -3172,6 +3176,134 @@ def test_1D_weights_axis(data, axis, weights):
...
+
+        b = np.asarray([np.nan, 1.0])
+        self.assertFalse(cfunc(a, b))
```
guilhermeleobas, line 3208: Compare the output of `cfunc` with `pyfunc` instead.

guilhermeleobas (changes requested):
Thanks, @czgdp1807, I only have a few more requests.

czgdp1807:
> I just remembered that you need to update the file `numpysupported.rst` to include `np.allclose`.

Done.

guilhermeleobas (dismissed):
LGTM. Thanks, @czgdp1807

czgdp1807:
@sklam @stuartarchibald @esc Same for this PR. Is it ready for merge? Please let me know if anything is left to be done here. Thanks.

numba/np/arraymath.py
```diff
@@ -864,6 +864,100 @@ def flat_all(a):
...
+
+@overload(np.allclose)
+@overload_method(types.Array, "allclose")
+def np_allclose(a, b, rtol=1e-05, atol=1e-08, equal_nan=False):
+
+    if not type_can_asarray(a):
+        raise TypeError('The first argument "a" must be array-like')
+
+    if not type_can_asarray(b):
+        raise TypeError('The first argument "b" must be array-like')
```
sklam, line 901:
```suggestion
        raise TypeError('The second argument "b" must be array-like')
```

sklam (commented):
This should be the last needed change

numba/np/arraymath.py
czgdp1807: I have addressed the review and the tests are passing. Please let me know if anything else is to be done here. Thanks.

sklam (approved):
Thanks for the patch!

PR #7949, opened 2022-03-29
Add Cuda Vector Types

### LINKED ISSUE #7943: CUDA Vector Types
CUDA supports writing [builtin vector types](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#built-in-vector-types) in kernels. Such as `float2`, `uint3`, etc. Each element of the vector type can be accessed via `x`, `y`, `z`, `w`. And each vector type has its own `make_` factory function. It would be nice if Numba cuda supports them.

isVoid:
This is the initial introduction of cuda vector types. This PR adds support to vector type creation and readouts. Note that not only vector types can be constructed a number of primitive types, but also with all valid combinations of vector type and primitives. Such as:

```python
f3 = float32x3(float32(1.0), float32x2(float32(2.0), float32(3.0)))
u4 = uint32x4(uint32x2(4, 3), uint32x2(2, 1))
```

Contributes to #7943 .


gmarkall:
gpuci run tests

docs/source/cuda-reference/kernel.rst
```diff
@@ -537,3 +537,35 @@ Timer Intrinsics
...
+    zero = uint32(0)
+    u2 = uint2(1, 2)
+    # Construct a 3-component vector with primitive type and a 2-component vector
+    u3 = uint3(zero, u2)
+    # Construct a 4-component vector with 2 2-component vectors
+    u4 = uint4(u2, u2)
+
+The 1st, 2nd, 3rd and 4th component of the vector type can be accessed through fields 
+``x``, ``y``, ``z``, ``w``, respectively. The components are currently immutable
+after construction.
```
gmarkall, line 571: Just to make it clear that they're immutable as a limitation of the present implementation in Numba:

```suggestion
``x``, ``y``, ``z``, and ``w`` respectively. The components are immutable after
construction in the present version of Numba; it is expected that support for
mutating vector components will be added in a future release.
```

numba/cuda/device_init.py
```diff
@@ -31,6 +33,10 @@
 
 reduce = Reduce = reduction.Reduce
 
+# Expose vector types constructors as module level attributes.
```
gmarkall, line 36:
```suggestion
# Expose vector type constructors as module level attributes.
```

docs/source/cuda-reference/kernel.rst
```diff
@@ -537,3 +537,35 @@ Timer Intrinsics
 
     Suspends the thread for a sleep duration approximately close to the delay
     ``ns``, specified in nanoseconds.
+
```
gmarkall, line 540: It feels like "Types" should be a higher-level section all of its own:

```suggestion


Types
-----
```

numba/cuda/device_init.py
```diff
@@ -31,6 +33,10 @@
...
+for vector_type_stub in _vector_type_stubs:
```
gmarkall, line 37: Should we `del` `vector_type_stub` and maybe `_vector_type_stubs` after this loop so they don't appear in the module globals afterwards?

numba/cuda/stubs.py
```diff
@@ -656,3 +657,40 @@ class habs(Stub):
         Returns the fp16 result of the absolute value.
 
         """
+
+
+#-------------------------------------------------------------------------------
+# vector types
```
gmarkall, line 663: Should all of the implementation of the generation of `_vector_type_stubs` go in a function to make it easier to avoid leaking into the module namespace? e.g. `cuda.stubs.name` is `"double"` and `cuda.stubs.nelem` is `4` after import at present.

gmarkall (changes requested):
I've partially reviewed this so far, but I think it's worth posting some comments now with more to follow - there are some comments / suggestions on the diff.

Can we support casting between vectors that are essentially the same? For example, the following:

```python
from numba import cuda, types

def f(c):
    if c:
        return cuda.int2(2, 2)
    else:
        return cuda.long2(2, 2)

cuda.compile_ptx(f, (types.boolean,), device=True)
```

presently gives:

```
numba.core.errors.TypingError: Failed in cuda mode pipeline (step: nopython frontend)
Can't unify return type from the following types: int2, long2
Return of: IR name '$16return_value.5', type 'int2', location: 
File "cast_same.py", line 6:
def f(c):
    <source elided>
    if c:
        return cuda.int2(2, 2)
        ^
Return of: IR name '$28return_value.5', type 'long2', location: 
File "cast_same.py", line 8:
def f(c):
    <source elided>
    else:
        return cuda.long2(2, 2)
        ^
```

Do we need the `once` decorator in `vector_types.py`? Is it to prevent a circular import? If it is needed, can we just call the function `initialize()`? (I feel like the "once" part of it is an implementation detail, in a sense).

I've still yet to properly look at the majority of `vector_types.py` and the tests - I'll look at these in the next pass.

docs/source/cuda-reference/kernel.rst
isVoid: Should this be under kernel.rst or should it have its own page? Like host kernel, types has its own page: https://numba.pydata.org/numba-doc/dev/reference/types.html.

numba/cuda/stubs.py
```diff
@@ -656,3 +657,40 @@ class habs(Stub):
...
+):
+    type_name = f"{name}{nelem}"
+    attr_names = _vector_type_attribute_names[:nelem]
+
+    vector_type_stub = type(
+        type_name, (Stub,),
+        {
+            **{attr: lambda self: None for attr in attr_names},
+            **{"_description_": f"<{type_name}>"}
+        }
```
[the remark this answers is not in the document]
isVoid: I think the suggested change is better to override `__signature__` to make init signature look correct. The stub docstring mentions the type is meaningless outside of a cuda kernel - which is correct and should be mentioned.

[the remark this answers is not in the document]
isVoid: My change:
```
In [1]: from numba import cuda

In [2]: cuda.float3?
Init signature: cuda.float3(x, y, z, /)
Docstring:      A stub for float3 to be used in CUDA kernel.
File:           /raid/wangm/dev/numba/numba/cuda/stubs.py
Type:           type
Subclasses:   
```

isVoid:
> Can we support casting between vectors that are essentially the same?

To clarify - `int32` happens to be the base type for both `intx` and `longx`, are you asking that for the vector types that has the same base type and number of elements should be recognized as the same type? Not asking for conversion rules such as `intx` -> `longx`?

> Do we need the once decorator in vector_types.py? 

It's called twice in `target.py`. Once for impl and once for decl. Since I believe the order with which decl and impl are called is an implementation detail, the `once` guard is set to prevent the vector types being initialized twice.

numba/cuda/stubs.py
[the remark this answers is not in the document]
gmarkall: Thanks, this looks better - I wasn't aware of the `__signature__` property :-)

docs/source/cuda-reference/kernel.rst
gmarkall: I think perhaps starting a new page is a good idea - once that is done, we should at some point add reference to `float16` in it as well (but not for this PR / as part of this work).

isVoid:
> Can we support casting between vectors that are essentially the same?

@gmarkall and I discussed offline on this and there are two solutions: one is to implement the same integer types as aliases and the other is to implement the casting rules. We favored implemented as aliases; it aligns more with how numba types are implemented. The main cons of this solutions are that the kernels won't be portable across platforms. For example `int` and `long` are aliased as the same type on one platform but could be different on another so the example given in @gmarkall 's comment above cannot compile for some platforms.

To address this concern - we plan to define the vector type for python kernels with a new set of names more aligned with numpy's naming conventions: it would look like `int32x4` for 32 bit vectors with 4 elements. This makes bitwidth of the base type unambiguous. Developers who writes new kernels in numba cuda is encouraged to follow this convention.

In addition, all names from native CUDA will be aliased on this new naming convention (based on machine definition of `int` and `long`). This is for convenience for people who want to transcribe their cuda kernel to python easily. And since [there isn't a casting rule](https://godbolt.org/z/nnn5Msq45) in native cuda between `intx` in `longx`, we don't expect this aliasing should break during these migration.

numba/cuda/tests/cudapy/test_vector_type.py
```diff
@@ -0,0 +1,155 @@
...
+            res[i * 4 + j + 1] = f.y
+            res[i * 4 + j + 2] = f.z
+            res[i * 4 + j + 3] = f.w
+
+    return cuda.jit(kernel)
+
+
+class TestCudaVectorType(CUDATestCase):
+
+    def setUp(self):
```
gmarkall, line 115: This seems to increase the runtime of the tests by 5-8 seconds on my system. Removing it seems not to have any ill-effects (is there a potential race?). In case it is needed, I tried using `setUpClass` instead but that is just as slow. I'm not sure why though.

- Why is this needed?
- If it is needed, can it be made quicker?

docs/source/cuda/index.rst
```diff
@@ -7,6 +7,7 @@ Numba for CUDA GPUs
 .. toctree::
 
    overview.rst
+   types.rst
```
gmarkall, line 10: I think it would be better to link to this in the reference guide, underneath the "CUDA Kernel API" section. IIRC this is where the content previously was when it was in the "Intrinsic Attributes and Functions" section.

docs/source/cuda/types.rst
```diff
@@ -0,0 +1,49 @@
+Types on CUDA target
```
gmarkall, line 1:
```suggestion
CUDA-Specific Types
```

docs/source/cuda/types.rst
```diff
@@ -0,0 +1,49 @@
...
+====================
+
+Note: This page is about types specific to CUDA targets. Most types on CPU targets are also
+available in CUDA target, for information on these types, see :ref:`numba-types`.
```
gmarkall, line 5: There is a list of Python / NumPy types supported in the CUDA target at https://numba--7949.org.readthedocs.build/en/7949/cuda/cudapysupported.html#built-in-types - could you update the reference to point there?

docs/source/cuda/types.rst
```diff
@@ -0,0 +1,49 @@
...
+
+Vector Types
+~~~~~~~~~~~~
+
+`CUDA Vector Types <https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#built-in-vector-types>`_
+are available in the kernel. There are 2 important distinctions from vector types in numba CUDA:
```
gmarkall, line 11:
```suggestion
are usable in kernels. There are two important distinctions from vector types in CUDA C/C++:
```

docs/source/cuda/types.rst
```diff
@@ -0,0 +1,49 @@
...
+
+First, the recommended names for vector types in Numba CUDA is formatted as ``base_typexN``,
```
gmarkall, line 13:
```suggestion
First, the recommended names for vector types in Numba CUDA is formatted as ``<base_type>x<N>``,
```

docs/source/cuda/types.rst
```diff
@@ -0,0 +1,49 @@
...
+where `base_type` is the base type of the vector, and `N` is the number of elements in the vector.
```
gmarkall, line 14: Sphinx formatting for code instead of emphasis:

```suggestion
where ``base_type`` is the base type of the vector, and ``N`` is the number of elements in the vector.
```

docs/source/cuda/types.rst
```diff
@@ -0,0 +1,49 @@
...
+Examples include ``int64x3``, ``uint16x4``, ``float32x4``, etc. This is consistent in type names with
+the pydata community where bit widths are explicit. For new Numba CUDA kernels, this is the recommended
+way to write vector types.
```
gmarkall, line 17:
```suggestion
Examples include ``int64x3``, ``uint16x4``, ``float32x4``, etc. For new Numba CUDA kernels,
this is the recommended way to instantiate vector types.
```

docs/source/cuda/types.rst
```diff
@@ -0,0 +1,49 @@
...
+
+For convenience, users who prefer to adapt existing kernels from native CUDA to Numba CUDA may use
+aliases consistent with its native namings. For example, ``float3`` aliases ``float32x3``,
+``long3`` aliases ``int32x3`` or ``int64x3``, depending on platform specification, etc.
```
gmarkall, line 21:
```suggestion
For convenience, users adapting existing kernels from CUDA C/C++ to Python may use
aliases consistent with the C/C++ namings. For example, ``float3`` aliases ``float32x3``,
``long3`` aliases ``int32x3`` or ``int64x3`` (depending on the platform), etc. 
```

docs/source/cuda/types.rst
```diff
@@ -0,0 +1,49 @@
...
+where `base_type` is the base type of the vector, and `N` is the number of elements in the vector.
+Examples include ``int64x3``, ``uint16x4``, ``float32x4``, etc. This is consistent in type names with
+the pydata community where bit widths are explicit. For new Numba CUDA kernels, this is the recommended
+way to write vector types.
+
+For convenience, users who prefer to adapt existing kernels from native CUDA to Numba CUDA may use
+aliases consistent with its native namings. For example, ``float3`` aliases ``float32x3``,
+``long3`` aliases ``int32x3`` or ``int64x3``, depending on platform specification, etc. 
+
+Second, unlike native CUDA where factory functions are used, vector types are constructed directly
```
gmarkall, line 23:
```suggestion
Second, unlike CUDA C/C++ where factory functions are used, vector types are constructed directly
```

docs/source/cuda/types.rst
```diff
@@ -0,0 +1,49 @@
...
+with their constructor. For example, to construct a ``float32x3``:
+
+.. code-block:: python3
+
+    from numba.cuda import float32x3
+
+    # In kernel
+    f3 = float32x3(float32(0.0), float32(-1.0), float32(1.0))
```
gmarkall, line 31: Are the casts necessary? What happens if the user does `float32x3(0.0, -1.0, 1.0)`?

gmarkall (changes requested):
Many thanks for the changes - I've made some comments on the documentation on the diff.

Let me know if you need any help with looking into any of the test fails.

numba/cuda/tests/cudapy/test_vector_type.py
```diff
@@ -0,0 +1,255 @@
...
+                    arr, np.array(range(vty.num_elements))
+                )
+
+    @skip_on_cudasim("Vector type isn't defined in simulator.")
+    def test_fancy_creation_readout(self):
+        for vty in vector_types.values():
+            kernel = make_fancy_creation_kernel(vty)
+
+            expected = np.array([
+                1,
```
gmarkall, line 198:
```suggestion
                # 1-component vectors
                1,
```

numba/cuda/tests/cudapy/test_vector_type.py
```diff
@@ -0,0 +1,255 @@
...
+                1,
+                2, 3,
```
gmarkall, line 200:
```suggestion
                # 2-component vectors
                2, 3,
```

gmarkall (commented):
Submitting review to show comments, because I accidentally started one

docs/source/cuda/types.rst
isVoid: It turns out they are not necessary, since I only defined `float32` as the argument, I believe the floating point literals were cast to `float32`.

docs/source/cuda/types.rst
```diff
@@ -0,0 +1,49 @@
...
+
+    zero = uint32(0)
+    u2 = uint32x2(1, 2)
+    # Construct a 3-component vector with primitive type and a 2-component vector
+    u3 = uint32x3(zero, u2)
+    # Construct a 4-component vector with 2 2-component vectors
+    u4 = uint32x4(u2, u2)
+
+The 1st, 2nd, 3rd and 4th component of the vector type can be accessed through fields 
+``x``, ``y``, ``z``, and ``w`` respectively. The components are immutable after
```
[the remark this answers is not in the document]
isVoid: Great idea

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

numba/cuda/simulator/vector_types.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def __init__(self, *args):
+        args_flattened = []
+        for arg in args:
+            if isinstance(arg, SimulatedVectorType):
+                args_flattened += arg.as_list()
+            else:
+                args_flattened.append(arg)
+        self._attrs = self.attributes[:len(args_flattened)]
+        assert len(self._attrs) == len(args_flattened), \
```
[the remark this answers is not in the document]
isVoid: Ah, that's right. I meant to test that `num_elements` should be the same as `len(args_flattened)`. Since the number of elements is an intrinsic property of the vector type and the number of argument should match it. Updating.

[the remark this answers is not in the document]
isVoid: Done. With 7deb8dbbd we are actually testing the number of arguments matches with the number of element:
```python
In [11]: float32x3(1.0, 2.0, 3.0)
Out[11]: <numba.cuda.simulator.vector_types.float32x3 at 0x7f15810b5910>

In [12]: float32x3(1.0, 2.0, 3.0, 4.0)
Input In [12], in <cell line: 1>()
----> 1 float32x3(1.0, 2.0, 3.0, 4.0)

File /raid/wangm/dev/numba/numba/cuda/simulator/vector_types.py:16, in SimulatedVectorType.__init__(self, *args)
     14         args_flattened.append(arg)
     15 self._attrs = self.attributes[:len(args_flattened)]
---> 16 assert self.num_elements == len(args_flattened), \
     17     f"{self.num_elements} != {len(args_flattened)}"
     19 for arg, attr in zip(args_flattened, self._attrs):
     20     setattr(self, attr, arg)

AssertionError: 3 != 4

In [13]: float32x3(1.0, 2.0)
Input In [13], in <cell line: 1>()
----> 1 float32x3(1.0, 2.0)

File /raid/wangm/dev/numba/numba/cuda/simulator/vector_types.py:16, in SimulatedVectorType.__init__(self, *args)
     14         args_flattened.append(arg)
     15 self._attrs = self.attributes[:len(args_flattened)]
---> 16 assert self.num_elements == len(args_flattened), \
     17     f"{self.num_elements} != {len(args_flattened)}"
     19 for arg, attr in zip(args_flattened, self._attrs):
     20     setattr(self, attr, arg)

AssertionError: 3 != 2
```

numba/cuda/tests/cudapy/test_vector_type.py
```diff
@@ -0,0 +1,265 @@
+import os
+
+import numpy as np
+
+from numba.cuda.testing import CUDATestCase
+
+from numba import cuda
+
+if os.environ.get("NUMBA_ENABLE_CUDASIM", 0):
+    vector_types = cuda.vector_types._simulated_vector_types
```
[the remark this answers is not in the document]
isVoid: Yes. the strange part is the line that follows, the vector types needs to be injected into `cuda` namespace in user domain. I fixed this with importing the vector types and exposing them to `numba.cuda` module in `simulator.__init__.py`.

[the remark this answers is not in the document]
isVoid: For this import line I think it's fine, because `vector_type` is internal and should never be accessed by user, we are importing the list of vector types here so that we can loop over the vector types and test them programmatically.

numba/cuda/simulator/vector_types.py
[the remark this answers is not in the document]
gmarkall: Thanks for clarifying and the update. Is this an error that a user can make? If so, would it be more appropriate to provide a `ValueError` or `RuntimeError` with a message like `float32x3 expected 3 values, received 2`?

numba/cuda/simulator/vector_types.py
```diff
@@ -0,0 +1,60 @@
...
+            else:
+                args_flattened.append(arg)
+        self._attrs = self.attributes[:len(args_flattened)]
+        if not self.num_elements == len(args_flattened):
+            raise TypeError(
+                f"{self.name} expects {self.num_elements}"
+                f" elements, got {len(args_flattened)}"
+            )
+        assert self.num_elements == len(args_flattened), \
+            f"{self.num_elements} != {len(args_flattened)}"
```
gmarkall, line 22: I think this is now superfluous given the check and raise above?

numba/cuda/stubs.py
```diff
@@ -736,3 +740,88 @@ class hmin(Stub):
...
+                **{attr: lambda self: None for attr in attr_names},
+                **{
+                    "_description_": f"<{type_name}>",
+                    "__signature__": Signature(parameters=[
+                        Parameter(
+                            name=attr_name, kind=Parameter.POSITIONAL_ONLY
+                        ) for attr_name in attr_names[:nelem]
+                    ]),
+                    "__doc__": f"A stub for {type_name} to be used in "
+                    "CUDA kernel."
```
gmarkall, line 784:
```suggestion
                    "CUDA kernels."
```

gmarkall:
gpuci run tests

numba/cuda/simulator/vector_types.py
```diff
@@ -0,0 +1,60 @@
...
+
+        for arg, attr in zip(args_flattened, self._attrs):
+            setattr(self, attr, arg)
+
+    @property
+    def name(self):
```
gmarkall, line 28: I think this and `num_elements` get replaced when a new type is made from `SimulatedVectorType` in `make_simulated_vector_type()` below, and that these methods are only placeholders. Is this correct?

gmarkall (changes requested):
Many thanks for the updates - just some very small comments / questions on the diff.

numba/cuda/simulator/vector_types.py
isVoid: That's correct, they are replaced at https://github.com/isVoid/numba/blob/4faf3f58b3192212b5ae91eff07cedd2f0ae3b2a/numba/cuda/simulator/vector_types.py#L41-L43. On a second note they seem unrelated though.

numba/cuda/simulator/vector_types.py
isVoid: Oops yes.

gmarkall:
gpuci run tests

gmarkall (approved):
Many thanks for addressing the comments!

gmarkall:
@esc / @stuartarchibald Could this have a CUDA buildfarm smoketest please?

esc:
> @esc / @stuartarchibald Could this have a CUDA buildfarm smoketest please?

`Build numba_smoketest_cuda_yaml_131 has started`

stuartarchibald:
> > @esc / @stuartarchibald Could this have a CUDA buildfarm smoketest please?
> 
> `Build numba_smoketest_cuda_yaml_131 has started`

Thanks @esc, this passed.

gmarkall:
Thanks for the buildfarm run @esc / @stuartarchibald.

From OOB discussion, there was some concern about the length of time the Windows builders in the buildfarm took - however, I tested this on my machine locally (Win11, RTX 6000, Xeon 6128, 96GB RAM) and the increase in test time on Windows was negligible on a modern quiescent machine - it is likely that what was observed was just the general slowness of the Windows buildfarm machines.

PR #7957, opened 2022-04-04
Use cloudpickle for disk caches

### LINKED ISSUE #2868: cache=True & generators?
reproducing code:
``` python
import numba
import numpy as np

@numba.njit(cache=True)
def _thisfunc(vals):
    for v in vals:
        yield v

# calling code:
_thisfunc(np.random.rand(500))

```
results in: 
```
C:\Anaconda3\lib\site-packages\numba\caching.py in _dump(self, obj)
    563 
    564     def _dump(self, obj):
--> 565         return pickle.dumps(obj, protocol=-1)
    566 
    567     @contextlib.contextmanager

PicklingError: Can't pickle <function _thisfunc at 0x0000014AF82B4D08>: it's not the same object as __main__._thisfunc
```
but with `cache=False` the function works. Is there any workaround for this?

sklam:
Fixes #2868

### DIFF
--- a/numba/core/caching.py
+++ b/numba/core/caching.py
@@ -560,7 +560,7 @@ class IndexDataCacheFile(object):
         return os.path.join(self._cache_path, name)
 
     def _dump(self, obj):
-        return pickle.dumps(obj, protocol=-1)
+        return dumps(obj)
 
     @contextlib.contextmanager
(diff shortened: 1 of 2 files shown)


stuartarchibald (approved):
Thanks for the patch, looks good. Also tested manually against #2868 reproducers.

Issue #6382, opened 2020-10-16
Release global interpreter lock (GIL) in ahead-of-time-compiled code


99991:
## Feature request

It would be nice if the global interpreter lock could be released for AOT-compiled code so that multiple functions can be run in parallel.

The syntax could be `@cc.export("foo", nogil=True)`, similar to how jit currently works:`@jit(nogil=True)`

The following code calls three test functions 10 times in parallel (run on a quad-core CPU) to demonstrate that releasing the GIL is a useful feature to improve performance on multi-core systems.

- Jitted function with GIL (0.95 seconds)
- Jitted function with GIL released (0.25 seconds)
- AOT-compiled function with GIL (1.05 seconds)

```python3
from numba import jit
from numba.pycc import CC
from threading import Thread
import time

@jit(nogil=False)
def testfunc_jit_withgil(n, x):
    s = 0.0
    for _ in range(n):
        s += x
    return s

@jit(nogil=True)
def testfunc_jit_nogil(n, x):
    s = 0.0
    for _ in range(n):
        s += x
    return s

cc = CC("mymodule")

@cc.export("testfunc_aot", "f8(i8, f8)")
def testfunc_aot(n, x):
    s = 0.0
    for _ in range(n):
        s += x
    return s

cc.compile()

from mymodule import testfunc_aot

def benchmark(func):
    # warmup
    for _ in range(10):
        func(10**6, 1.0)

    start_time = time.perf_counter()

    threads = [Thread(target=func, args=(10**8, 1.0)) for _ in range(10)]
    for thread in threads: thread.start()
    for thread in threads: thread.join()

    elapsed_time = time.perf_counter() - start_time
    print(elapsed_time, "seconds\n")

print("with JIT, with GIL:")
benchmark(testfunc_jit_withgil)
print("with JIT, without GIL:")
benchmark(testfunc_jit_nogil)
print("AOT-compiled:")
benchmark(testfunc_aot)
```

esc:
@99991 thanks for submitting this, I have labelled it as a feature request accordingly.

auderson:
I guess the following codes could be a workaround to enable `nogil`:
What I did here is just adding `flags.release_gil = True` to `_ModuleCompiler._cull_exports` and let `CC` use the modified compiler.

```python
import os
from distutils import log
from llvmlite.binding import Linkage
import llvmlite.llvmpy.core as lc

from numba.pycc import CC
[...]

PR #7962, opened 2022-04-06
Added `nogil` option in `cc.export`

czgdp1807:
Closes https://github.com/numba/numba/issues/6382

Results from the benchmark in gh-6382,

**Before this PR**

```python
with JIT, with GIL:
0.9407890839999999 seconds

with JIT, without GIL:
0.16593487500000004 seconds

AOT-compiled:
0.9418166669999999 seconds
```

**After the diff in this PR**

```python
with JIT, with GIL:
0.9413587919999999 seconds

with JIT, without GIL:
0.15825958299999998 seconds

AOT-compiled:
0.15359266699999985 seconds
```

cc: @gmarkall @esc @auderson

### DIFF
--- a/numba/tests/test_gil.py
+++ b/numba/tests/test_gil.py
@@ -164,6 +164,34 @@ class TestGILRelease(TestCase):
             compiled_f(a, i)
         self.check_gil_held(caller)
 
+    def test_cc_export_nogil(self):
+        from numba.pycc import CC
+        cc = CC('pycc_test_nogil')
+        @cc.export("f_cc_export", f_sig, nogil=True)
+        def f_nogil(a, indices):
+            # same as f but for cc.export
+            for idx in indices:
+                # Let another thread run
+                sleep(10 * sleep_factor)
+                a[idx] = PyThread_get_thread_ident()
+        cc.compile()
+        from pycc_test_nogil import f_cc_export
+        self.check_gil_released(f_cc_export)
+
+    def test_cc_export_gil(self):
+        from numba.pycc import CC
+        cc = CC('pycc_test_gil')
+        @cc.export("f_cc_export", f_sig)
+        def f_gil(a, indices):
+            # same as f but for cc.export
+            for idx in indices:
+                # Let another thread run
+                sleep(10 * sleep_factor)
+                a[idx] = PyThread_get_thread_ident()
+        cc.compile()
+        from pycc_test_gil import f_cc_export
+        self.check_gil_held(f_cc_export)
+
     def test_object_mode(self):
         """
         When the function is compiled in object mode, a warning is
--- a/numba/tests/test_gil.py
+++ b/numba/tests/test_gil.py
@@ -164,34 +164,6 @@ class TestGILRelease(TestCase):
             compiled_f(a, i)
         self.check_gil_held(caller)
 
-    def test_cc_export_nogil(self):
-        from numba.pycc import CC
-        cc = CC('pycc_test_nogil')
-        @cc.export("f_cc_export", f_sig, nogil=True)
-        def f_nogil(a, indices):
-            # same as f but for cc.export
-            for idx in indices:
-                # Let another thread run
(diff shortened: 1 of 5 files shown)


gmarkall:
Many thanks for the PR! Once the CI is passing, I'll mark it for review (I briefly skimmed the CI output - is there a file or two that needs committing?)

numba/tests/test_gil.py
```diff
@@ -164,6 +164,34 @@ def caller(a, i):
...
+        cc = CC('pycc_test_nogil')
+        @cc.export("f_cc_export", f_sig, nogil=True)
+        def f_nogil(a, indices):
+            # same as f but for cc.export
+            for idx in indices:
+                # Let another thread run
+                sleep(10 * sleep_factor)
+                a[idx] = PyThread_get_thread_ident()
+        cc.compile()
+        from pycc_test_nogil import f_cc_export
```
czgdp1807, line 178: I think this module isn't detected while importing on CI. Though, locally on my system it works fine. @gmarkall Any ideas how to make this work? I am trying to figure out the pattern in `compile_with_pycc.py` and `test_pycc.py` but unable to succeed. I would appreciate any hints/ideas for this. Thanks.

stuartarchibald: I suggest moving these tests to `test_pycc` as they are testing `CC` capabilities. As a benefit of doing that, the `test_pycc`  module contains facilities/infrastructure to help with testing AOT compilation (e.g. patching `sys.modules` appropriately) which might help?

stuartarchibald: If the `cc.export` decorated functions in these tests are moved to `numba/tests/compile_with_pycc.py` and then the checking part of unit test added to new tests in `test_pycc::TestCC` then I think it would likely "just work" as the infrastructure is already set up in there. I think addition of a `nogil` flag is similar to having a `use_nrt` setting cf. https://github.com/numba/numba/blob/974a5ca78182d2b35f32df6b395d38ef636f26aa/numba/tests/compile_with_pycc.py#L16-L17

numba/tests/test_pycc.py
```diff
@@ -233,6 +257,35 @@ def check_compile_for_cpu(self, cpu_name):
...
+
+    def check_gil_held(self, func):
+        arr = self.run_in_threads(func, n_threads=4)
+        distinct = set(arr)
+        self.assertEqual(len(distinct), 1, distinct)
+
+    def test_cc_export_nogil(self):
+        with self.check_cc_compiled(self._test_module.cc) as lib:
+            self.assertEqual(lib.f_nogil.__module__, 'pycc_test_simple')
+            self.check_gil_released(lib.f_nogil)
```
czgdp1807, line 282: This test fails locally. However, the `nogil` flag works fine locally when used in `test_gil.py`. I tried removing the context manager but that changed nothing.

czgdp1807: It fails with `n_threads=32` in `self.check_gil_released`. For 4 and 12 it works fine.

numba/tests/test_pycc.py
```diff
@@ -233,6 +233,34 @@ def check_compile_for_cpu(self, cpu_name):
...
+
+        @cc_gil.export("f_nogil", f_sig)
+        def f_gil(n, x):
+            s = 0.0
+            for _ in range(n):
+                s += x
+            return s
+
+        self.assertFalse(cc_gil._release_gil)
+
```
czgdp1807, line 263: @stuartarchibald After discussing with @guilhermeleobas I implemented the approach to test the correctness of GIL flag according the user provided value of `nogil` parameter in `cc.export`. Please let me know if its okay.

czgdp1807, line 263: @stuartarchibald After discussing with @guilhermeleobas I implemented the approach to test the correctness of GIL flag according the user provided value of `nogil` parameter in `cc.export`. Please let me know if its okay.

PR #7964, opened 2022-04-06
Support for large constant dictionaries in Python 3.10

### LINKED ISSUE #7894: Limit in Constant Dictionary Size with Python 3.10
## Reporting a bug

When using a large constant dictionary Python 3.10 the bytecode changes. Here is the reproducer:

```python3
import numba

@numba.njit
def f():
    d = {
        "a": 1,
        "b": 2,
        "c": 3,
        "d": 4,
        "e": 5,
        "f": 6,
        "g": 7,
        "h": 8,
        "i": 9,
        "j": 10,
        "k": 11,
        "l": 12,
        "m": 13,
        "n": 14,
        "o": 15,
        "p": 16,
        "q": 17,
        "r": 18,
        "s": 19,
    }
    return d

f()
```

In particular, when a dictionary gets too large it appears to construct smaller smaller dictionaries on the stack and then use `DICT_UPDATE` to combine the dictionaries. Numba doesn't seem to support the `update` method of dictionaries or `DICT_UPDATE`.

I think this should be treated as a regression as it works in earlier Python versions.

njriasan:
Closes #7894 

Provides support for large literal dictionaries in Python 3.10 by fusing `MAP_ADD` and `DICT_UPDATE` operations within the same basic block into an initialized dictionary. This is not a complete implementation (for example it can't handle control flow), but it should handle the common cases for Python 3.10.

Marking this as WIP because I still need to write exhaustive tests.


numba/core/interpreter.py
```diff
@@ -314,6 +314,164 @@ def peep_hole_delete_with_exit(func_ir):
...
+        """
+        old_assign = blk.body[old_lineno]
+        old_target = old_assign.target
+        old_bm = old_assign.value
+        # Build the literals
+        literal_keys = []
+        # Track the values
+        values = []
+        for pair in new_items:
+            k, v = pair
```
benwilliamgraham, line 345: why not `for k, v in new_items`?

benwilliamgraham (commented):
Looks good!

numba/core/interpreter.py
```diff
@@ -314,6 +314,164 @@ def peep_hole_delete_with_exit(func_ir):
...
+                            # Delete from the const maps
+                            del lit_maps[var.name]
+                            del map_updates[var.name]
+                new_body.append(stmt)
+        # Update we any dictionaries that weren't used in the block yet.
+        # We pop off the last statement because this needs to remain last
+        # as it might be a jump, branch, or return
+        end = new_body.pop()
+        for var_name, lineno in lit_maps.items():
+            items = map_updates[var_name]
```
njriasan, line 468: I think I may need to also update definitions whenever a node is deleted or when adding the new build maps (since the build_map changes to include more information).

numba/core/interpreter.py
njriasan: This was just from debugging. Will update.

numba/core/interpreter.py
njriasan: Not completely sure, but it seems like the map definitions should be updated based on the other optimizations.

numba/typed/dictobject.py
```diff
@@ -912,6 +912,19 @@ def impl(d):
     return impl
 
 
+@overload_method(types.DictType, 'update')
+def ol_dict_update(d, other):
+    if not isinstance(d, types.DictType):
+        return
+    if not isinstance(other, types.DictType):
+        return
```
stuartarchibald, line 920: Should `d = {'a':1, 'b':2}; d.update(a=10)` be supported too?

numba/core/interpreter.py
stuartarchibald: Assuming this is referring to `func_ir._definitions`, then I think the definitions need to be kept consistent. I don't think it hugely matters whether this is fixed at the time of invalidation or buffered and fixed at the end of the transform, as nothing else will be accessing it until the transform is complete. The `list_to_tuple` rewrite elects to do this at invalidation time, e.g. (and others): https://github.com/numba/numba/blob/f04f4b164887e5aa5ffeeed1bef317e13ec995f8/numba/core/interpreter.py#L176-L182

numba/typed/dictobject.py
njriasan: I agree this should be supported eventually, but this is meant as a fallback in case the peephole fails and therefore isn't needed for this PR. I think its beyond the scope of this issue and we should open a followup.

njriasan:
Okay I have added the tests for this PR. It should be ready for review.

numba/tests/test_interpreter.py
```diff
@@ -618,3 +618,202 @@ def inline_func(flag):
...
+        self.assertEqual(a, b)
+
+    @skip_unless_py10
+    def test_large_heterogeneous_const_keys_dict(self):
+        """
+        Tests that a function with a large heterogenous constant
+        dictionary remains a constant.
+        """
+        def const_keys_func(a):
+            # D is a heterogenous dictionary
```
stuartarchibald, line 684:
```suggestion
            # D is a heterogeneous dictionary
```

numba/tests/test_interpreter.py
```diff
@@ -618,3 +618,202 @@ def inline_func(flag):
...
+
+    This optimization modifies the IR to rejoin dictionaries
+    and remove the update() calls. This then allows code that
+    depends on literal dictionaries or literal keys to succeed.
+    """
+
+    @skip_unless_py10
+    def test_large_heterogeneous_const_dict(self):
+        """
+        Tests that a function with a large heterogenous constant
```
stuartarchibald, line 641:
```suggestion
        Tests that a function with a large heterogeneous constant
```

numba/tests/test_interpreter.py
```diff
@@ -618,3 +618,202 @@ def inline_func(flag):
...
+        dictionary remains a constant.
+        """
+        def const_func():
+            # D is a heterogenous dictionary
```
stuartarchibald, line 645:
```suggestion
            # D is a heterogeneous dictionary
```

numba/tests/test_interpreter.py
```diff
@@ -618,3 +618,202 @@ def inline_func(flag):
...
+
+        py_func = const_keys_func
+        cfunc = njit()(const_keys_func)
+        value = "efwf"
+        a = py_func(value)
+        b = cfunc(value)
+        self.assertEqual(a, b)
+
+    @skip_unless_py10
+    def test_usercode_update_heterogenous_keys(self):
```
stuartarchibald, line 718:
```suggestion
    def test_usercode_update_heterogeneous_keys(self):
```

stuartarchibald (commented):
Thanks for the patch @njriasan. I've concentrated this part of the review on the docs and the proposed behaviour of this change and the impact it potentially has across various use cases. I've left a few questions/cases for consideration inline, suggest these are discussed and resolved prior to the implementation itself being reviewed. Thanks again for your work on this!

numba/core/interpreter.py
```diff
@@ -955,6 +956,277 @@ def peep_hole_delete_with_exit(func_ir):
...
+    with a new definition that combines the two dicitonaries.
+
+    Python 3.10 may also rewrite an individual dictionary as an empty
+    build_map + many map_add, so those expressions are also replaced
+    with a constant build map.
+
+    When running this algorithm we do not remove any maps, even if they
+    were fused into an update to protect against strange choices in user
+    code. Here we depend on dead code elimination to run later
+    and remove any unnecessary maps.
```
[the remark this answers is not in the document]
njriasan: Yes. Since the implementation only looks at particular basic block, the implementation does not/cannot determine that d2 is unused in the entire IR. This change seems like it should be handled with some dead code elimination.

There is nothing in this code the guarantees DCE runs. I didn't run DCE yet in case anything relied on explicit pattern matching and I assumed DCE would run later in the pipeline. I'm guessing this is an erroneous assumption and causes a unused dictionary to be allocated consisting of all of d2's entries.

numba/tests/test_interpreter.py
```diff
@@ -618,3 +618,202 @@ def inline_func(flag):
...
+                "P": 1,
+                "Q": 1,
+                "R": 1,
+                "S": a,
+            }
+            return d["S"]
+
+        py_func = const_keys_func
+        cfunc = njit()(const_keys_func)
+        value = "efwf"
```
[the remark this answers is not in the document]
njriasan: This is unused code from revising the test. I will remove it.

numba/tests/test_interpreter.py
```diff
@@ -618,3 +618,202 @@ def inline_func(flag):
...
+                "a": 1,
+                "b": 2,
+                "c": 3,
+            }
+            d2 = {
+                "d": 4,
+                "e": a
+            }
+            d1.update(d2)
+            return d1["e"]
```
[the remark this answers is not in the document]
njriasan: I agree this is potentially confusing. I added these tests because from my perspective this user code is the closest example that resembles the Python 3.10 bytecode changes (and thus cannot be distinguished in the current implementation). Writing code in this style also proved to be the easiest way to test edge cases not possible in just the bytecode changes (for example d2 being used after the update).

Personally I think this case is uncommon and its seems unlikely that people would write code like this, although I'm open to discussion about possible changes.

numba/core/interpreter.py
```diff
@@ -1031,6 +1303,7 @@ def interpret(self, bytecode):
             # any large number of arguments from a list to a
             # tuple.
             peepholes.append(peep_hole_call_function_ex_to_call_function_kw)
+            peepholes.append(peep_hole_fuse_dict_add_updates)
```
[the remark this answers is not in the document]
njriasan: No. If run before `peep_hole_call_function_ex_to_call_function_kw` you can delete `_call_function_ex_replace_kws_large` because the dictionary used in varkwargs will always be a single build_map that is fully populated.

numba/tests/test_interpreter.py
```diff
@@ -618,3 +618,202 @@ def inline_func(flag):
...
+    into multiple dictionaries that are joined by a update()
+    call.
+
+    This optimization modifies the IR to rejoin dictionaries
+    and remove the update() calls. This then allows code that
+    depends on literal dictionaries or literal keys to succeed.
+    """
+
+    @skip_unless_py10
+    def test_large_heterogeneous_const_dict(self):
```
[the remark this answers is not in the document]
njriasan: Sounds good. Is the only test here to check `literally` with both homogeneous and heterogeneous dictionaries?

numba/core/interpreter.py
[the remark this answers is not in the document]
njriasan: I also think this change is worth making to reduce code complexity.

numba/tests/test_interpreter.py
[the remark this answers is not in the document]
stuartarchibald: I can understand implementing this from a "convenient way to test this specific transform" point of view, however, it has the unfortunate side effect of leaking into user facing functionality.

When designing and implementing the `LiteralStrKey` behaviour, it turned out there's a load of cases where you could potentially track some mutation of a const dictionary. However, from a usability perspective, it was decided that it'd be fairly hard to explain what's going when there's an invalid case even when the user actually wrote the control flow in their source (it being even harder when it occurs as a result of some bytecode optimisation!). All mutation of const types was therefore banned.

As a result of the above I do not think that supporting this should happen until:
1. a considerably improved method of tracking, explaining and reporting specific errors in relation to this is introduced.
2. all mutating methods in the type class are tracked and do the "mutation", this so as to provide consistency (it'd be strange to permit "update" but not "setitem").

Both of which I think are out of scope for this PR and would be a lot of work in themselves.

numba/core/interpreter.py
```diff
@@ -955,6 +956,277 @@ def peep_hole_delete_with_exit(func_ir):
     return func_ir
 
 
+def peep_hole_fuse_dict_add_updates(func_ir):
+    """
+    This rewrite removes d1.update(d2) calls that
+    are between two dictionaries, d1 and d2 in the
+    same basic block. This pattern can appear as a
```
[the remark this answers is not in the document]
njriasan: I generated this example:

```python3
from numba import njit

@njit
def foo(a):
    d1 = {'a':1,
          'b':2,
          'c':3,
          'd':4,
          'e':5,
          'f':6,
          'g':7,
          'h':8,
          'i':9,
          'p':'change type'}
    if a:
        val = 10
    else:
        val = 100
    d1['j'] = val
    return d1['a']

foo(False)
```

Based on this there are a couple of things I noticed (before the peephole optimzations run).
1. The key is defined after the branch (as opposed to before). This may not be consequential.
2. The assignment comes from an explicit `ir.SetItem`. I don't know much about this, but this seems more promising. Perhaps we could generate an error based on the source of the setitem and type of IR. Do you know if user code should always generate the `ir.SetItem` (assuming someone doesn't try explicitly call the setitem attribute, which seems like an uncommon/unlikely pattern)?

Regarding the error, I also don't think it be reliably raised before typing. Its fine for this optimization to fail so long as you don't need the dictionary to be a literal. I don't believe you should be able to determine that before typing.

njriasan (commented):
Thank you @stuartarchibald. I applied the immediate feedback and added some followup comments. I will make more updates, especially to the testing, later today.

numba/tests/test_interpreter.py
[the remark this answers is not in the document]
stuartarchibald: This is about ensuring that the available constant information is preserved through these transforms. The `literal_value` is available on the `LiteralStrKeyDict` type, its equivalent for a `DictType` is `initial_value`. A > 15 entry const string keys ->homogeneous literal value type dictionary should trigger the path through to being able to check the `initial_value` on a `DictType` is present and correct. Example:
```python
from numba import njit

@njit
def foo():
    d1 = {'a': 1,
          'b': 2,
          'c': 3,
          'd': 4,
          'e': 5,
          'f': 6,
          'g': 7,
          'h': 8,
          'i': 9,
          'j': 10,
          'k': 11,
          'l': 12,
          'm': 13,
          'n': 14,
          'o': 15,
          'p': 16,}

foo()
foo.inspect_types()
```
Similarly the `literal_value` of a `LiteralStrKeyDict` from changing the above to have `'p': 'abc'` should permit a check along the same lines as described but against `.literal_value`.

Note that the `initial_value` of a `DictType` is carried across mutation(s).

numba/core/interpreter.py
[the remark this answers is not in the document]
stuartarchibald: I've been thinking about a generalisation of this problem, the question "Can we know if the the dictionary being built via updates etc is from an optimisation or from a user choice/definition"? Using this:
```python

def user_source():
    d1 = {'a': 1,
          'b': 2,
          'c': 3,
          'd': 4,
          'e': 5,
          'f': 6,
          'g': 7,
          'h': 8,
          'i': 9,
          'j': 10,
          'k': 11,
          'l': 12,
          'm': 13,
          'n': 14,
          'o': 15,
          'p': 16,
          'q': 17,}

    d2 = {"r": 18,
          "s": 19,}

    d1.update(d2)


def internal_opt_source():
    d3 = {
        "a": 1,
        "b": 2,
        "c": 3,
        "d": 4,
        "e": 5,
        "f": 6,
        "g": 7,
        "h": 8,
        "i": 9,
        "j": 10,
        "k": 11,
        "l": 12,
        "m": 13,
        "n": 14,
        "o": 15,
        "p": 16,
        "q": 17,
        "r": 18,
        "s": 19,
    }
```
it seems like the `DICT_UPDATE` only occurs in `internal_opt_source`, in the `user_source` it's evidently a method call. Combining this information with the information here: https://github.com/numba/numba/issues/7894#issuecomment-1062915850 about the source of this change from the bytecode compiler, it seems to indicate that the presence of a `DICT_UPDATE` op code can only arise in the case of a CPython induced byte code optimisation i.e. it's not reachable from user code. Is this the piece of information that helps with answering the question?

Were this information about the presence of a `DICT_UPDATE` opcode encoded into the `ir.Expr.getattr` as e.g. `getattr(dict_inst, 'update_internal')` or something similar, would that make it possible to correctly implement a standard `dict.update` as normal, then have a special overload for `dict.update_internal` that wires into `dict.update` for a standard dict but the implementation for the `LiteralStrKeyDict.update_internal` wires into an overload that raises and provides an explanation of how it should not have been reached and there must be something that's preventing the transform?

This is quite long as might be best discussed at the next Numba public meeting.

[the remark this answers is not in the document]
njriasan: @stuartarchibald I think this sounds like the right approach and it would also fix some of the current issues in the code (namely failing to delete the other dictionary/dictionaries). I think it would be great if we could discuss that at the next Numba meeting. I'm going to pause on any other updates until we have that discussion.

esc:
> > @njriasan may I inquire about the status of this PR? It is now scheduled to be included in 0.56 so we'd like to get this over the line within the next week. Are you currently waiting on anything (e.g. feedback?) or blocked for this one?
> 
> Hi @esc I need to make the requested changes to only implement the algorithm on bytecode generated code (not user code). I will have time to work on this on Saturday and Sunday and can have it ready for review by Monday. I will make sure to be responsive all of next week to get this to the finish line.

Awesome, sounds great!!

numba/core/interpreter.py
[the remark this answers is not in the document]
njriasan: This is now just no.

numba/tests/test_interpreter.py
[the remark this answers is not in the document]
njriasan: Actually to clarify this is the value to make the dictionary heterogenous and is passed to the function. its type is significant but its value is not.

njriasan:
I have made the initial changes to refactor the optimization to only run on bytecode generated updates. I still need to add the additional tests that @stuartarchibald requested to confirm the initial values are correct.

numba/tests/test_interpreter.py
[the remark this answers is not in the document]
njriasan: @stuartarchibald I added a couple of these tests in the same style as `numba.tests.test_dictobject.TestTypedDictInitialValues`. I think these should be sufficient given the scope of the optimization, but please let me know if you need more.

njriasan:
@stuartarchibald I added the requested tests, and I think this is once again ready for review.

esc:
Quick question: did you ever work out which change in CPython for 3.10 lead to this change being required.

numba/core/interpreter.py
```diff
@@ -955,6 +956,271 @@ def peep_hole_delete_with_exit(func_ir):
...
+    neither is used between the update call, then we replace d1
+    with a new definition that combines the two dictionaries. At
+    the bytecode translation stage we convert DICT_UPDATE into
+    _update_from_bytecode, so we know that _update_from_bytecode
+    always comes from the bytecode change and not user code.
+
+    Python 3.10 may also rewrite the individual dictionaries
+    as an empty build_map + many map_add. Here we again look
+    that a _update_from_bytecode, and if so we replace these
+    with a single constant dictionary.
```
esc, line 977: I am not sure this paragraph makes sense. Should it be perhaps:

```suggestion
    Python 3.10 may also rewrite the individual dictionaries
    as an empty build_map + many map_add. Here we again look
    for an _update_from_bytecode, and if so we replace these
    with a single constant dictionary.
```

?

esc:
FWIW: I did run this through the Anaconda Internal Build Farm as `numba_smoketest_cpu_yaml_102` and it was all green 🟢

numba/core/interpreter.py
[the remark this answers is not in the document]
stuartarchibald: > This is now just no.

Thanks for confirming!

njriasan:
> Quick question: did you ever work out which change in CPython for 3.10 lead to this change being required.

Yes. This is basically the same change as the kwargs refactoring. https://github.com/python/cpython/blob/3.10/Python/compile.c#L4048. Is there anywhere you would like me to include this in the code?

numba/core/interpreter.py
njriasan: Yes absolutely. Thank you.

esc:
> > Quick question: did you ever work out which change in CPython for 3.10 lead to this change being required.
> 
> Yes. This is basically the same change as the kwargs refactoring. https://github.com/python/cpython/blob/3.10/Python/compile.c#L4048. Is there anywhere you would like me to include this in the code?

Perhaps it makes sense to include a link to that change in the docstring for the peephole optimizer function?

njriasan:
> > > Quick question: did you ever work out which change in CPython for 3.10 lead to this change being required.
> > 
> > 
> > Yes. This is basically the same change as the kwargs refactoring. https://github.com/python/cpython/blob/3.10/Python/compile.c#L4048. Is there anywhere you would like me to include this in the code?
> 
> Perhaps it makes sense to include a link to that change in the docstring for the peephole optimizer function?

Sure Happy to do so.

njriasan:
Done @esc

esc:
> Done @esc

Thank you!

numba/core/interpreter.py
```diff
@@ -955,6 +956,275 @@ def peep_hole_delete_with_exit(func_ir):
...
+                        del lit_map_def_idx[var.name]
+                        del lit_map_use_idx[var.name]
+                        del map_updates[var.name]
+
+            # Append the instruction to the new block
+            new_body.append(new_inst)
+
+        if blk_changed:
+            # If the block is changed replace the block body.
+            blk.body = [x for x in new_body if x is not None]
```
stuartarchibald, line 1170: Suggest clearing the existing body and extending it opposed to replacement.

numba/core/interpreter.py
```diff
@@ -955,6 +956,275 @@ def peep_hole_delete_with_exit(func_ir):
...
+        # is updated.
+        lit_map_use_idx = {}
+        # literal map var name -> list of key/value items for build map
+        map_updates = {}
+        blk_changed = False
+
+        for i, stmt in enumerate(blk.body):
+            # What instruction should we append
+            new_inst = stmt
+            # Name that shoud be skipped when tracking used
```
stuartarchibald, line 1055:
```suggestion
            # Name that should be skipped when tracking used
```

stuartarchibald (commented):
Thanks for the updates @njriasan. I think this patch now follows what we discussed at a public Numba meeting a few weeks back with regards to preserving the semantics of a `LiteralStrKeyDict` being immutable, I'm glad the suggestion/ideas worked out ok. With regards to the state of the current patch, in general it looks good, there's a few minor things to resolve in testing and a few typos etc. My main concern at present is that I think there's a bug for dictionaries with >34 elements, there's a reproducer inline. Thanks again for working on this.

numba/tests/test_interpreter.py
```diff
@@ -618,3 +620,355 @@ def inline_func(flag):
...
+    gh #7894
+
+    Tests that check a peephole optimization for constant
+    dictionaries in Python 3.10. The bytecode changes when
+    number elements > 15, which splits the constant dictionary
+    into multiple dictionaries that are joined by a update()
+    call.
+
+    This optimization modifies the IR to rejoin dictionaries
+    and remove the update() calls. This then allows code that
```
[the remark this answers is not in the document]
njriasan: Yes. I will correct this docstring.

numba/tests/test_interpreter.py
```diff
@@ -618,3 +620,355 @@ def inline_func(flag):
...
+            return d["S"]
+
+        py_func = const_func
+        cfunc = njit()(const_func)
+        a = py_func()
+        b = cfunc()
+        self.assertEqual(a, b)
+
+    @skip_unless_py10
+    def test_large_heterogeneous_const_keys_initial_values(self):
```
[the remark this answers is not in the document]
njriasan: Yes. I will remove the `initial_values` from the name/docstring of the relevant tests and add a comment explaining this.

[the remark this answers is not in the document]
njriasan: I should clarify, this just tests that operations that "require" a literal dictionary work, not the actual literal value.

numba/core/interpreter.py
```diff
@@ -955,6 +956,275 @@ def peep_hole_delete_with_exit(func_ir):
...
+                                # Delete the old IR for d1 and d2
+                                lit_map_use_idx[update_map_name].extend(
+                                    lit_map_use_idx[d2_map_name]
+                                )
+                                lit_map_use_idx[update_map_name].append(i - 1)
+                                for linenum in lit_map_use_idx[update_map_name]:
+                                    # Drop the existing definition.
+                                    _remove_assignment_definition(
+                                        blk.body, linenum, func_ir
+                                    )
```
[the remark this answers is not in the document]
njriasan: Thanks @stuartarchibald. This occurred because I attempted to delete the original dictionary definition for the first part of the dictionary multiple times. I will update the code to handle this case.

numba/tests/test_interpreter.py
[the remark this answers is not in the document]
njriasan: I got confused and thought this was referring to a different test. This checks the initial value for LiteralStrKeyDict, which becomes necessary as the dictionary is heterogeneous. I will update the test accordingly.

numba/core/interpreter.py
njriasan: Okay I made the change (and updated the previous peephole I wrote as well with this suggestion as well).

numba/tests/test_interpreter.py
```diff
@@ -618,3 +620,355 @@ def inline_func(flag):
...
+            d1 = {
+                "a": 1,
+                "b": 2,
+                "c": 3,
+            }
+            d2 = {
+                "d": 4,
+                "e": 4
+            }
+            d1.update(d2)
```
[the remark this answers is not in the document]
njriasan: @esc I tried adding this test but I had some issues. I'm hoping you can help me with. In the prefer_literal path, both `check_before` and `check_after` produce the same literal value. However, the function falls back to the non-literal version, I assume because the returned result is not stable when returning literals.

Does this assessment seem consistent? How should I test this?

numba/typed/dictobject.py
njriasan: Done.

numba/core/interpreter.py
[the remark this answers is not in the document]
stuartarchibald: Great, thanks.

numba/core/interpreter.py
stuartarchibald: Thanks, and thanks for doing the additional adjustment.

njriasan:
@esc @stuartarchibald I tried implementing this test:

```python3
def check_before(x):
    pass

def check_after(x):
    pass

@overload(check_before, prefer_literal=True)
def ol_check_before(x):
    # Assert initial value here
    def impl(x):
        pass
    return impl

@overload(check_after, prefer_literal=True)
def ol_check_after(x):
    # Assert initial value here
    def impl(x):
        pass
    return impl

@njit
def foo():

    d1 = {
        "a": 1,
        "b": 2,
        "c": 3,
    }
    d2 = {
        "d": 4,
        "e": 4
    }
    check_before(d1)
    d1.update(d2)
    check_after(d1)
    # Create a use of d2 in a new block.
    if len(d1) > 4:
        return d2
    return d1
```

However I'm having issues structuring a good test. It appears that `check_before` is correct on the `prefer_literal` attempt. However, it seems to retry compiling `check_before` without literals, so that break my test. Is this the expected behavior? Any advice on how to clean up this test? The relevant test if you want to play with it is `test_usercode_update_use_d2`.

stuartarchibald:
@njriasan how about something like:

<details>

```python
    @skip_unless_py10
    def test_usercode_update_use_d2(self):
        """
        Tests an example using a regular update is
        not modified by the optimization.
        """

        def check_before(x):
            pass

        def check_after(x):
            pass

        checked_before = False
        checked_after = False

        @overload(check_before, prefer_literal=True)
        def ol_check_before(d):
            nonlocal checked_before
            impl = lambda d: d
            if not checked_before:
                checked_before = True
                a = {
                    "a": 1,
                    "b": 2,
                    "c": 3,
                }
                self.assertTrue(isinstance(d, types.DictType))
                self.assertEqual(d.initial_value, a)
            return impl

        @overload(check_after, prefer_literal=True)
        def ol_check_after(d):
            nonlocal checked_after
            impl = lambda d: d
            if not checked_after:
                checked_after = True
                self.assertTrue(isinstance(d, types.DictType))
                self.assertTrue(d.initial_value is None)
            return impl

        def const_dict_func():
            """
            Dictionary update between two constant
            dictionaries. This verifies d2 doesn't
            get incorrectly removed.
            """
            d1 = {
                "a": 1,
                "b": 2,
                "c": 3,
            }
            d2 = {
                "d": 4,
                "e": 4
            }
            check_before(d1)
            d1.update(d2)
            check_after(d1)
            # Create a use of d2 in a new block.
            if len(d1) > 4:
                return d2
            return d1

        py_func = const_dict_func
        cfunc = njit()(const_dict_func)
        a = py_func()
        b = cfunc()
        self.assertEqual(a, b)
```
</details>

I think this is probably what's going on: As type inference is iterative, the first pass through `check_before` will yield `d1` as a `DictType` with an `initial_value` of the literal keys/values inferred from the bytecode, but as type inference progresses it will encounter the `d1.update(d2)`. This will change the type of `d1` to a `DictType` with an `initial_value` of `None`, such that when type inference tries again and passes through `check_before` for the second time it will now have `d1` presented as a `DictType` with an `initial_value` of `None`. Capturing this state in `nonlocal`s helps manage the assertion behaviour.

njriasan:
> @njriasan how about something like:
> 
> ```python
>     @skip_unless_py10
>     def test_usercode_update_use_d2(self):
>         """
>         Tests an example using a regular update is
>         not modified by the optimization.
>         """
> 
>         def check_before(x):
>             pass
> 
>         def check_after(x):
>             pass
> 
>         checked_before = False
>         checked_after = False
> 
>         @overload(check_before, prefer_literal=True)
>         def ol_check_before(d):
>             nonlocal checked_before
>             impl = lambda d: d
>             if not checked_before:
>                 checked_before = True
>                 a = {
>                     "a": 1,
>                     "b": 2,
>                     "c": 3,
>                 }
>                 self.assertTrue(isinstance(d, types.DictType))
>                 self.assertEqual(d.initial_value, a)
>             return impl
> 
>         @overload(check_after, prefer_literal=True)
>         def ol_check_after(d):
>             nonlocal checked_after
>             impl = lambda d: d
>             if not checked_after:
>                 checked_after = True
>                 self.assertTrue(isinstance(d, types.DictType))
>                 self.assertTrue(d.initial_value is None)
>             return impl
> 
>         def const_dict_func():
>             """
>             Dictionary update between two constant
>             dictionaries. This verifies d2 doesn't
>             get incorrectly removed.
>             """
>             d1 = {
>                 "a": 1,
>                 "b": 2,
>                 "c": 3,
>             }
>             d2 = {
>                 "d": 4,
>                 "e": 4
>             }
>             check_before(d1)
>             d1.update(d2)
>             check_after(d1)
>             # Create a use of d2 in a new block.
>             if len(d1) > 4:
>                 return d2
>             return d1
> 
>         py_func = const_dict_func
>         cfunc = njit()(const_dict_func)
>         a = py_func()
>         b = cfunc()
>         self.assertEqual(a, b)
> ```
> 
> I think this is probably what's going on: As type inference is iterative, the first pass through `check_before` will yield `d1` as a `DictType` with an `initial_value` of the literal keys/values inferred from the bytecode, but as type inference progresses it will encounter the `d1.update(d2)`. This will change the type of `d1` to a `DictType` with an `initial_value` of `None`, such that when type inference tries again and passes through `check_before` for the second time it will now have `d1` presented as a `DictType` with an `initial_value` of `None`. Capturing this state in `nonlocal`s helps manage the assertion behaviour.

This makes sense to me @stuartarchibald. I'll update my test. Thank you.

numba/core/interpreter.py
```diff
@@ -943,6 +948,279 @@ def peep_hole_delete_with_exit(func_ir):
...
+            # a build_map.
+            stmt_build_map_out = None
+            if isinstance(stmt, ir.Assign) and isinstance(stmt.value, ir.Expr):
+                if stmt.value.op == "build_map":
+                    # Skip the output build_map when looking for used vars.
+                    stmt_build_map_out = stmt.target.name
+                    # If we encounter a build map add it to the
+                    # tracked maps.
+                    lit_map_def_idx[stmt.target.name] = i
+                    lit_map_use_idx[stmt.target.name] = [i]
```
stuartarchibald, line 1058:
```suggestion
                    lit_map_use_idx[stmt.target.name].append(i)
```
this is acting as the initial state constructor ?

njriasan: This should always be the first element, but I'm happy to make this change. It seems simpler to me.

stuartarchibald: It was seeing this that made me think it was matching the use case of a `defaultdict(list)`, without this change, it could just be a `dict`as before. I'm ok with it either way so long as it's safe, but as noted, at present it's a `defaultdict(list)` but without making use of that behaviour, so I think the change makes sense?

njriasan:
@stuartarchibald Thank you that fixed the test. @esc This PR should be completely ready now.

numba/tests/test_interpreter.py
```diff
@@ -622,6 +623,504 @@ def inline_func(flag):
...
+        py_func = const_func
+        cfunc = njit()(const_func)
+        a = py_func()
+        b = cfunc()
+        self.assertEqual(a, b)
+
+    @skip_unless_py10
+    def test_large_heterogenous_LiteralStrKeyDict_initial_values(self):
+        """Check the initial values for a LiteralStrKeyDict requiring
+        optimizations because it is heterogenous.
```
stuartarchibald, line 684:
```suggestion
    def test_large_heterogeneous_LiteralStrKeyDict_literal_values(self):
        """Check the literal values for a LiteralStrKeyDict requiring
        optimizations because it is heterogeneous.
```

numba/core/interpreter.py
```diff
@@ -80,21 +81,24 @@ def get_assignment_source(self, destname):
         return None
 
 
-def _remove_assignment_definition(old_body, idx, func_ir):
+def _remove_assignment_definition(old_body, idx, func_ir, ignore_missing=False):
```
stuartarchibald, line 84: I think ideally this should involve tracking what has previously been removed opposed to ignoring duplicate removals? It seems like `ignore_missing=True` essentially aliases "Remove a previously existing valid `rhs` from `lhs`'s definitions" and "Remove a non-sensical `rhs` (i.e. from a bug elsewhere!) from `lhs`'s definitions".

The question here is really... what's the risk? Is the only way to trigger such an issue via a bug in the IR state from else where? If so, leave as is? What do you think @njriasan?

numba/tests/test_interpreter.py
```diff
@@ -622,6 +623,504 @@ def inline_func(flag):
...
+            self.assertTrue(isinstance(d, types.LiteralStrKeyDict))
+            self.assertEqual(d.literal_value, expected)
+            self.assertEqual(hasattr(d, 'initial_value'), False)
+            return lambda d: d
+
+        @njit
+        def foo():
+            # D is a heterogeneous dictionary
+            # so this code can only compile if
+            # d has the correct initial values.
```
stuartarchibald, line 726:
```suggestion
            # d has the correct literal values.
```

stuartarchibald (commented):
Thanks for the update @njriasan. I've a question (inline) on which I'd like your opinion please, other than that there's a couple of typos to resolve then this looks good to merge. Thanks again!

numba/core/interpreter.py
njriasan: @stuartarchibald I've pushed a change that keeps a dictionary of which variables have been deleted. Let me know if this changes looks sufficient to you.

stuartarchibald: > I don't know if there is a realistic risk, but I suggestion is reasonable and also don't very difficult to implement. Since in the past I found issues with other peepholes in the definitions and this stuff is done manually, I think its worth effort to add a set of deleted definitions.

OK, if the effort is short then please feel free to add this, many thanks. I guess having this also helps with ensuring self-consistency as it's enforcing a prescribed behaviour, it'll probably be more easily debugged as a result too.

numba/core/interpreter.py
```diff
@@ -81,20 +81,23 @@ def get_assignment_source(self, destname):
...
     Deletes the definition defined for old_body at index idx
     from func_ir. We assume this stmt will be deleted from
     new_body.
 
-    We pass ignore_missing if we may attempt to delete the
-    same definition multiple times.
+    In some optimizations we may update the same variable multiple times.
+    In this situation, we only need to delete a particular definition once,
+    so we deleted_vars, which is a default dict for each assignment name
+    to the set of values that have already been deleted.
```
stuartarchibald, line 93:
```suggestion
    this is tracked in already_deleted_def, which is a map from
    assignment name to the set of values that have already been
    deleted.
```

stuartarchibald (commented):
@njriasan code update looks good RE: https://github.com/numba/numba/pull/7964#discussion_r901835839, thanks! There's a minor docstring fix suggested else looks good.

njriasan:
Thanks @stuartarchibald! Looking forward to merging this.

stuartarchibald (approved):
Many thanks for all your effort on this @njriasan, it was a tricky problem and your persistence in fixing it is much appreciated. Thanks also for your input in reviewing this @esc. Great to see this fixed, patch is approved!

esc:
Thank you for the patch!

PR #7965, opened 2022-04-06
Use uuid4 instead of PID in cache temp name to prevent collisions.

### LINKED ISSUE #7952: Cache temp filename not unique across nodes
## Feature request

In numba/core/caching.py, tmpname is supposed to be unique to avoid race conditions.
Please change this to use uuid.uuid1() instead of os.getpid().  I'm using numba in either
a slurm or ray cluster where jobs on different nodes may have the same pid.

Thanks, Mark W

    def _open_for_write(self, filepath):
         tmpname = '%s.tmp.%d' % (filepath, os.getpid().)

should be:

    def _open_for_write(self, filepath):
         tmpname = '%s.tmp.%s' % (filepath, uuid.uuid1())

stuartarchibald:
As title.

Fixes #7952

### DIFF
--- a/numba/core/caching.py
+++ b/numba/core/caching.py
@@ -13,6 +13,7 @@ import os
 import pickle
 import sys
 import tempfile
+import uuid
 import warnings
 
 from numba.misc.appdirs import AppDirs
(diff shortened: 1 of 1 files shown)


markweissman (approved):
Looks good

sklam (approved):
LGTM

PR #7976, opened 2022-04-11
[WIP] Add axis option to `np.unique`

### LINKED ISSUE #7663: Feature request numpy.unique axis=0
---

## Feature request

I want to go from
```
a = np.array([[1, 1, 1, 0, 0, 0],
              [0, 1, 1, 1, 0, 0],
              [0, 1, 1, 1, 0, 0],
              [1, 1, 1, 0, 0, 0],
              [1, 1, 1, 1, 1, 0]])
```

to 
```
array([[1, 1, 1, 0, 0, 0],
       [0, 1, 1, 1, 0, 0],
       [1, 1, 1, 1, 1, 0]])
```

Using numpy I would use 

`unique_rows = np.unique(a, axis=0)`

however, the [axis parameter is not supported yet](https://numba.pydata.org/numba-doc/dev/reference/numpysupported.html). I tried some workarounds based on [this SO question](https://stackoverflow.com/questions/16970982/find-unique-rows-in-numpy-array), but I havn't figured out a sollution yet.

PABannier:
The goal of this PR is to add the axis option for 2-dimensional arrays to the `np.unique` overloaded function.

A minimum working example: 

```python
import numpy as np
from numba import njit

a = np.array([[1, 1, 1, 0, 0, 0],
              [0, 1, 1, 1, 0, 0],
              [0, 1, 1, 1, 0, 0],
              [1, 1, 1, 0, 0, 0],
              [1, 1, 1, 1, 1, 0]])

@njit
def f(x):
    return np.unique(a, axis=0)

print(f(a))
```

Expected behavior:
```python
[[1, 1, 1, 0, 0, 0],
 [0, 1, 1, 1, 0, 0],
 [1, 1, 1, 1, 1, 0]]
```

This PR also adds the `axis` option to the `np.all` overloaded function, since it is needed in the implementation of `np.unique`.

Closes #7663


rishi-kulkarni:
Here are some thoughts based on the initial implementation in #7663:

For `np.all`, I don't think the implementation I wrote is a robust way to handle the fact that the operation you want to do is sometimes F-ordered and sometimes C-ordered. Maybe #7622 fixes this, @sklam? Though as far as I'm aware, you can't call custom ufuncs from `@overload`ed code. 

As a result of the above, _any_ numpy function that has an `axis` argument ends up having a pretty messy implementation. The major roadblocks I ran into with this function  were as follows:

1. The axis problem in most numpy functions could be solved in a general way by making `ufunc` methods accept axis arguments - specifically `ufunc.reduce` and `ufunc.accumulate`
2. In order to `overload` a `ufunc.reduce` method with an `axis` argument, the number of dimensions needs to be known at compile time 
3. Even knowing the output dimensions, some kind of code generation needs to be done to properly set up the reduction loop (or, as is done here, just give up on handling anything outside of a narrow set of use cases)

I guess I decided the first step to supporting `axis` arguments for `np.unique` would be to figure out a flexible strategy for `ufunc.reduce` - which I couldn't quite figure out. It seems like @kc611 started working on something in #7524. I'm not sure if I missed anything about the background of this problem - maybe @stuartarchibald knows more.

sjauhri:
Hello team,
Yes, this functionality is still needed

gmarkall:
@sjauhri Given that this PR never had any activity from the initial author following the comments from @rishi-kulkarni, I'd suggest that if this is to be implemented it should be done in a new PR.

PR #7979, opened 2022-04-12
Add support of np.geomspace

braniii:
This PR aims to add support for `np.geomspace`.
Related to https://github.com/numba/numba/issues/4074

### DIFF
--- a/numba/tests/test_np_functions.py
+++ b/numba/tests/test_np_functions.py
@@ -165,6 +165,14 @@ def logspace3(start, stop, num=50):
     return np.logspace(start, stop, num=num)
 
 
+def geomspace(start, stop):
+    return np.geomspace(start, stop)
+
+
+def geomspace_num(start, stop, num=50):
+    return np.geomspace(start, stop, num=num)
+
+
 def rot90(a):
     return np.rot90(a)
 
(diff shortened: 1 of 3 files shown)


braniii:
I've needed to increase the `rtol` by a factor of 10 to make the tests pass for complex numbers. I have no idea if this is an issue of the `np.logspace` implementation or of taking the `np.log10` of an complex number. But probably it is the latter.

numba/tests/test_np_functions.py
```diff
@@ -2577,6 +2585,80 @@ def test_logspace3_exception(self):
         self.assertIn('The third argument "num" must be an integer',
                       str(raises.exception))
 
+    def test_geomspace(self):
+
+        def inputs():
+            #start, stop
```
stuartarchibald, line 2591: I think some tests for negative numbers would be good. Might also be worth updating the implementation to reflect NumPy's implementation, this might help with handling of complex numbers? xref: https://github.com/numpy/numpy/blob/ae13307732b430daa56c6a85ab56006611a5b6ba/numpy/core/function_base.py#L287

stuartarchibald (commented):
Thanks for the patch, this is a great start, I've added a comment in line with regards to the implementation and testing which will hopefully help resolve issues around complex numbers.

numba/tests/test_np_functions.py
braniii: Thx, this fixed the issue.

braniii:
I have also specified in the docs that the first two arguments of `np.linspace`, `np.logspace`, and `np.geomspace` have to be scalars. The numpy implementation supports also interpolation between arrays.

guilhermeleobas:
I tested this PR with the inputs used by NumPy and the code is currently failing for 
```python
@njit
def geom(a, b, c):
    return np.geomspace(a, b, num=c)

a = -4j
b = -324j
c = 5
np.testing.assert_allclose(geom(a, b, c), geom.py_func(a, b, c))
```

@braniii I've put all the converted tests in a gist that you can copy and paste in the test file.
https://gist.github.com/guilhermeleobas/a94511fe9d7825b60c965b9d69e37136

braniii:
I need to admit that I was using geomspace only for real values. @guilhermeleobas thank you for finding this bug in my implementation and providing this helpful gist, but I guess (or better hope) this is an issue of the current `np.logspace` implementation which is used under the hood. I will adjust my PR once I find some time to fix the issue.

Issue #7903, opened 2022-03-11
error from getitem with optional array


auderson:
## Reporting a bug

- [ ] I have tried using the latest released version of Numba (most recent is
 visible in the change log (https://github.com/numba/numba/blob/main/CHANGE_LOG).
- [ ] I have included a self contained code sample to reproduce the problem.
  i.e. it's possible to run as 'python bug.py'.

Hi, I found the following codes will fail:
```python
import numpy as np
from numba import njit


@njit("float64[:](float64[:], optional(boolean[:]))")
def f1(x, mask=None):
    if mask is None:
        mask = np.isfinite(x)
    return x[mask]
```
```
numba.core.errors.TypingError: Failed in nopython mode pipeline (step: nopython frontend)
No implementation of function Function(<built-in function getitem>) found for signature:
 
 >>> getitem(array(float64, 1d, A), OptionalType(array(bool, 1d, A)))
 
There are 22 candidate implementations:
  - Of which 20 did not match due to:
  Overload of function 'getitem': File: <numerous>: Line N/A.
    With argument(s): '(array(float64, 1d, A), OptionalType(array(bool, 1d, A)))':
   No match.
  - Of which 2 did not match due to:
  Overload in function 'GetItemBuffer.generic': File: numba/core/typing/arraydecl.py: Line 166.
    With argument(s): '(array(float64, 1d, A), OptionalType(array(bool, 1d, A)))':
   Rejected as the implementation raised a specific error:
     NumbaTypeError: unsupported array index type OptionalType(array(bool, 1d, A)) in [OptionalType(array(bool, 1d, A))]
  raised from /home/auderson/miniconda3/lib/python3.8/site-packages/numba/core/typing/arraydecl.py:72
During: typing of intrinsic-call at <ipython-input-4-70443e7e4add> (5)
File "<ipython-input-4-70443e7e4add>", line 5:
def f1(x, mask=None):
    <source elided>
        mask = np.isfinite(x)
    return x[mask]
    ^
```
```python

@njit("float64[:](float64[:], optional(int64[:]))")
def f2(x, mask=None):
    if mask is None:
        mask = np.arange(x.size)
    return x[mask]
```
```
numba.core.errors.TypingError: Failed in nopython mode pipeline (step: nopython frontend)
No implementation of function Function(<built-in function getitem>) found for signature:
 
 >>> getitem(array(float64, 1d, A), OptionalType(array(int64, 1d, A)))
 
There are 22 candidate implementations:
[...]

PR #7980, opened 2022-04-13
Fix: optional array index

1844144:

### DIFF
--- a/numba/core/typing/arraydecl.py
+++ b/numba/core/typing/arraydecl.py
@@ -27,6 +27,9 @@ def get_array_index_type(ary, idx):
     if not isinstance(ary, types.Buffer):
         return
 
+    if isinstance(idx, types.Optional):
+        return
+
     ndim = ary.ndim
 
     left_indices = []


numba/core/dispatcher.py
```diff
@@ -969,6 +969,24 @@ def folded(args, kws):
...
+                    if self.overloads.get(tuple(args_variant)) is not None:
+                        continue
+                    try:
+                        cres_variant = self._compiler.compile(args_variant,
+                                                              return_type)
+                    except Exception:
+                        pass
+                    else:
+                        self.add_overload(cres_variant)
+
```
sklam, line 989: I don't think adding overloads here is the right approach. The problem in https://github.com/numba/numba/issues/7903 needs to be addressed at runtime---the array indexing implementation needs to check whether the index is an optional.

1844144: Not the right approach - exactly.  I thought you will not read "draft" PR's so thank you for reading this.  
I've provided additional description above. First commit in PR solves index trouble, but there is one more issue. By changes in dispatcher i've tried to deal with OmittedArg, and yes that was wrong direction (but half of tests is green :) )
 
so I still diving in code and figuring out OmittedArg issue.

1844144:
Summary:

If index array is Optional , we should return None as array index can not be None.

When `njit` called with signature it disables compiling, so we need to extend decorator to compile cases when each Optional argument will be omitted.

sklam:
This PR is implementing two concepts. I'd want to consider them separately

1) array-index being optional:

> Summary:
> 
> If array index is Optional , we should return None as array index can not be None.

NumPy indexing is tricky. It allows `None`; e.g.

```python
In [1]: import numpy as np
np
In [2]: arr = np.ones(4)

In [3]: arr[None]
Out[3]: array([[1., 1., 1., 1.]])

In [4]: arr = np.ones((2, 3, 4))

In [5]: arr[None]
Out[5]:
array([[[[1., 1., 1., 1.],
         [1., 1., 1., 1.],
         [1., 1., 1., 1.]],

        [[1., 1., 1., 1.],
         [1., 1., 1., 1.],
         [1., 1., 1., 1.]]]])
```

I need to check if the ongoing advance indexing PRs will add support for this.


2) spelling in signature for default arguments:

> 
> When `njit` called with signature it disables compiling, so we need to extend decorator to compile cases when each Optional argument will be omitted.

I'll need to discuss this with the maintainer group as this is a significant change.

sklam:
Thank you for the contribution.

After some discussions, we conclude that:

1. Support of optional index is more involved and is related to the in-flight advance indexing PRs. Tracked in https://github.com/numba/numba/issues/8941#issuecomment-1540508603
2. An alternative spelling for default argument is preferred: https://github.com/numba/numba/issues/8950 

Due to the above, we are closing this PR.

PR #7944, opened 2022-03-29
[WIP] Pathfinding for CUDA kernel caching

gmarkall:
This mostly just needed the missing pieces hooking up, but the final implementation could do with some refactoring. This incorporates the fix for PTX picking from @c200chromebook at https://gist.github.com/c200chromebook/1ee304161a39b247e8a029ff9d2b3cd7

This seems to work for simple examples. Some thought may need to be given about how to ensure loaded PTX and cubins are valid - e.g. checking the current driver and toolkit version.

PR #7986, opened 2022-04-15
Small caching refactor to support target cache implementations

gmarkall:
This PR refactors the caching implementation such that a target can implement its own cache without modifying the core of Numba. These changes are derived from the experience of writing #7944. There is no functional change. The changes include:

- Fix the arguments to `BaseContext.get_executable()` - if a target context forgets to implement `get_executable()`, then a call to it fails because the default implementation was missing the `env` argument.
- Removal of the `_get_codegen()` function `caching.py` - the design of this function forces it to be updated for every target. To replace it, `_load_overload()` and `_save_overload()` use a method/property on the objects they handle to get the codegen instead. This required adding the `codegen` property to `CompileResult`.
- making `_CacheImpl` public - a target might want its own cache implementation, so it shouldn't really be private in `caching.py`.

A PR based on this one that adds caching for CUDA by only making changes in the `numba.cuda` module will follow, but this is ready for review as-is.

### DIFF
--- a/numba/core/base.py
+++ b/numba/core/base.py
@@ -948,7 +948,7 @@ class BaseContext(object):
         res = imputils.fix_returning_optional(self, builder, sig, status, res)
         return res
 
-    def get_executable(self, func, fndesc):
+    def get_executable(self, func, fndesc, env):
         raise NotImplementedError
 
     def get_python_api(self, builder):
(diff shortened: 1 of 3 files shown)


gmarkall:
I am still working on the CUDA caching PR, but a proof of refactor in this branch is implemented in https://github.com/numba/numba/compare/main...gmarkall:cuda-caching-2 - only changes the CUDA target on top of the commits in this PR.

stuartarchibald (approved):
Thanks for the patch, the changes look good and follow what we guessed in relation to what might need changing to the core to support on disk caching for e.g. the CUDA target.

PR #7990, opened 2022-04-18
Stop patching error

sklam:
Error messages in the pattern of "Failed in ... mode pipeline (step: ...)" is not helpful to users.
The patched message can be confusing.

### DIFF
--- a/numba/core/compiler_machinery.py
+++ b/numba/core/compiler_machinery.py
@@ -3,7 +3,7 @@ from abc import abstractmethod, ABCMeta
 from collections import namedtuple, OrderedDict
 import inspect
 from numba.core.compiler_lock import global_compiler_lock
-from numba.core import errors, config, transforms, utils
+from numba.core import errors, config, transforms
 from numba.core.tracing import event
 from numba.core.postproc import PostProcessor
 from numba.core.ir_utils import enforce_no_dels, legalize_single_scope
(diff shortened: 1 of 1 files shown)


stuartarchibald (commented):
Thanks for the patch. I was in two minds about removing this on the basis that it can provide that little bit of essential additional context in a bug report or when providing assistance to a user that cannot share their reproducing code. However, I think the same information can be obtained from setting `NUMBA_TRACE=1` on the command line, it lists the pipeline and function being compiled and then each pipeline stage as it is entered, so provides the same information but without patching user facing error messages. As a result, I think removing it is a good idea as it is superfluous on the common path.

The only further comment I have is about how the exceptions are laid out on screen, taking this example (broken due to a `TypingError`):
```python
from numba import njit

@njit
def foo(x,):
    return x == x

foo(b'')
```
It's probably easiest to observe the following by running this locally with `colorama` installed, however, I've added in markers with `<HIGHLIGHTING>` to show where highlighting starts and ends.

With the standard patched exception from `main` it shows:
```
numba.core.errors.TypingError: Failed in nopython mode pipeline (step: native lowering)
<START HIGHLIGHTING>
No conversion from array(bool, 1d, C) to bool for '$8return_value.3', defined at None
<END HIGHLIGHTING>
```
whereas with this patch it shows:
```
numba.core.errors.TypingError: <START HIGHLIGHTING> No conversion from array(bool, 1d, C) to bool for '$8return_value.3', defined at None <END HIGHLIGHTING>
```
I'm wondering if editing the exception so that the error message appears on a new line helps with making it clear from a compiler output sense opposed to a what-python-does sense? i.e.
```
numba.core.errors.TypingError: 
<START HIGHLIGHTING>
No conversion from array(bool, 1d, C) to bool for '$8return_value.3', defined at None
<END HIGHLIGHTING>
```
the type of the error probably doesn't matter so much as they are often rewritten/compounded within the compiler pipelines, but the message does?

Issue #7767, opened 2022-01-20
Partial support only for `np.ndarray.sort` / support `np.ndarray.sort(axis=...)`


erezinman:
## Reporting a bug
In numba version 0.55 python 3.8, there's no support for sortting a multidimensional array. Consider the following code:

```python
@njit
def sort_multidimensional_arr(arr: np.ndarray) -> np.ndarray:
    arr = arr.reshape(-1, 2)
    arr.sort()

sort_multidimensional_arr(np.arange(10, dtype=int))
```

It results in the following result:
```
Invalid use of BoundFunction(array.sort for array(int64, 2d, C)) with parameters ()

During: resolving callee type: BoundFunction(array.sort for array(int64, 2d, C))
During: typing of call at /home/razor/PycharmProjects/datamind_train_infra/dmcommon-submodule/dmcommon/data_classes/time_types/optimized.py (9)


File "optimized.py", line 9:
def sort_multidimensional_arr(arr: np.ndarray) -> np.ndarray:
    <source elided>
    arr = arr.reshape(-1, 2)
    arr.sort()
    ^


Process finished with exit code 1
```

I may be able to get around it using `arr[:, 0].argsort()`, but for larger arrays the overhead in this becomes substantial.

stuartarchibald:
Thanks for the report. As noted in the docs: https://numba.readthedocs.io/en/0.55.0/reference/numpysupported.html#other-methods `np.ndarray.sort()` with arguments is currently not supported, will mark this as a feature request.

czgdp1807:
I can work on this I think.

czgdp1807:
I investigated. So, `np.ndarray.sort` doesn't support `axis` argument. However, according to NumPy behaviour, `np.ndarray.sort` in Numba should sort the last axis by default. So, I added that feature in the following diff. An example after the diff shows the change,

```diff
diff --git a/numba/core/typing/arraydecl.py b/numba/core/typing/arraydecl.py
index 43cdeeed4..629263aad 100644
--- a/numba/core/typing/arraydecl.py
+++ b/numba/core/typing/arraydecl.py
@@ -418,8 +418,7 @@ class ArrayAttribute(AttributeTemplate):
     def resolve_sort(self, ary, args, kws):
         assert not args
         assert not kws
-        if ary.ndim == 1:
-            return signature(types.none)
+        return signature(types.none)
 
     @bound_function("array.argsort")
     def resolve_argsort(self, ary, args, kws):
[...]

PR #7994, opened 2022-04-20
Supporting multidimensional arrays in quick sort

czgdp1807:
https://github.com/numba/numba/issues/7767

As of now I have supported multi-dimensional arrays in quick sort algorithm of Numba. The behaviour is same as NumPy i.e., the last axis is sorted.

cc: @guilhermeleobas

### DIFF
--- a/numba/tests/test_sort.py
+++ b/numba/tests/test_sort.py
@@ -738,10 +738,10 @@ class TestQuicksortMultidimensionalArrays(BaseSortingTest, TestCase):
             return shapes
 
         for i in range(2, int(math.sqrt(n)) + 1):
-            if n%i == 0:
-                shapes.append((n//i, i))
-                shapes.append((i, n//i))
-                _shapes = self.get_shapes(n//i)
+            if n % i == 0:
+                shapes.append((n // i, i))
+                shapes.append((i, n // i))
+                _shapes = self.get_shapes(n // i)
                 for _shape in _shapes:
                     shapes.append((i,) + _shape)
                     shapes.append(_shape + (i,))
--- a/numba/tests/test_sort.py
+++ b/numba/tests/test_sort.py
@@ -721,17 +721,35 @@ class TestQuicksortMultidimensionalArrays(BaseSortingTest, TestCase):
     quicksort = make_jit_quicksort(is_np_array=True)
     make_quicksort = staticmethod(make_jit_quicksort)
 
-    def assertSorted(self, orig, result, last_index=None):
-        # sorted() returns a list, so make sure we compare to another list
-        if last_index is not None:
-            self.assertTrue((orig == result[0][:last_index]).all())
+    def assertSorted(self, orig, result, contains_nans=False):
+        if contains_nans is True:
+            self.assertPreciseEqual(orig, result)
         else:
             self.assertEqual(orig.shape, result.shape)
             self.assertTrue((orig == result).all())
 
-    def array_factory(self, lst):
+    def array_factory(self, lst, shape=None):
         array = np.array(lst, dtype=np.float64)
-        return array.reshape(-1, array.shape[0])
+        if shape is None:
+            return array.reshape(-1, array.shape[0])
+        else:
+            return array.reshape(shape)
+
+    def get_shapes(self, n):
+        shapes = []
+        if n == 1:
+            return shapes
+
+        for i in range(2, int(math.sqrt(n)) + 1):
+            if n%i == 0:
+                shapes.append((n//i, i))
+                shapes.append((i, n//i))
+                _shapes = self.get_shapes(n//i)
+                for _shape in _shapes:
+                    shapes.append((i,) + _shape)
+                    shapes.append(_shape + (i,))
+
+        return shapes
 
     def test_run_quicksort(self):
         f = self.quicksort.run_quicksort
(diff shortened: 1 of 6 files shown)


numba/tests/test_sort.py
```diff
@@ -715,6 +716,80 @@ class TestQuicksortArrays(BaseQuicksortTest, TestCase):
...
         return np.array(lst, dtype=np.float64)
 
+class TestQuicksortMultidimensionalArrays(BaseSortingTest, TestCase):
+
+    quicksort = make_jit_quicksort(is_np_array=True)
+    make_quicksort = staticmethod(make_jit_quicksort)
+
+    def assertSorted(self, orig, result, last_index=None):
+        # sorted() returns a list, so make sure we compare to another list
+        print("orig: ", orig, "result: ", result)
```
guilhermeleobas, line 726: Remove `print` stmt.

guilhermeleobas (changes requested):
Hi @czgdp1807, thanks for working on this feature.

Correct if I'm wrong, but this PR is adding support for the `axis=` keyword for the `sort` method, right? And sorting is done by using `quicksort`? Can you update the `numpysupported.rst` doc to include this change?

czgdp1807:
> Correct if I'm wrong, but this PR is adding support for the `axis=` keyword for the `sort` method, right?

No, as of now I haven't added supported for `axis=` keyword. I have allowed multi-dimensional arrays. We know that in case of multi-dimensional arrays as input and no axis keyword argument, numpy sorts only the last axis. So, I am doing the same here. However, note that I can try out adding support for `axis` keyword argument in a follow up PR.



> And sorting is done by using `quicksort`?

Yes, exactly, for now only `quicksort` accepts multi-dimensional arrays.



> Can you update the `numpysupported.rst` doc to include this change?

Sure.

czgdp1807:
> However, note that I can try out adding support for `axis` keyword argument in a follow up PR.

I think I can do this. On it.

czgdp1807:
I have added axis and kind in signature of `array.sort` and `array.argsort`. Now I will try to fix the implementations to give desired results.

guilhermeleobas:
Thaks @czgdp1807. It is fine to limit the scope of this PR to what you originally proposed.

czgdp1807:
All the previous commits are available in https://github.com/czgdp1807/numba/tree/ndarray_sort_axiskind. Will continue work on them once this PR concludes.

guilhermeleobas (approved):

guilhermeleobas:
Thanks for the work @czgdp1807. PR looks good to me.

czgdp1807:
@esc @stuartarchibald @sklam Can we merge this? Its ready from my side.

numba/tests/test_sort.py
```diff
@@ -715,6 +716,79 @@ class TestQuicksortArrays(BaseQuicksortTest, TestCase):
...
+        # sorted() returns a list, so make sure we compare to another list
+        if last_index is not None:
+            self.assertTrue((orig == result[0][:last_index]).all())
+        else:
+            self.assertEqual(orig.shape, result.shape)
+            self.assertTrue((orig == result).all())
+
+    def array_factory(self, lst):
+        array = np.array(lst, dtype=np.float64)
+        return array.reshape(-1, array.shape[0])
```
sklam, line 734: Are we only testing on arrays with shape `(1, n)`? There needs to be test for non-one dimension and on various number of dimension.

numba/tests/test_sort.py
```diff
@@ -715,6 +716,79 @@ class TestQuicksortArrays(BaseQuicksortTest, TestCase):
     def array_factory(self, lst):
         return np.array(lst, dtype=np.float64)
 
+class TestQuicksortMultidimensionalArrays(BaseSortingTest, TestCase):
+
+    quicksort = make_jit_quicksort(is_np_array=True)
+    make_quicksort = staticmethod(make_jit_quicksort)
+
+    def assertSorted(self, orig, result, last_index=None):
+        # sorted() returns a list, so make sure we compare to another list
```
sklam, line 725: This comment is probably leftover from copying previous code and can be removed.

numba/tests/test_sort.py
```diff
@@ -715,6 +716,79 @@ class TestQuicksortArrays(BaseQuicksortTest, TestCase):
...
+        if last_index is not None:
+            self.assertTrue((orig == result[0][:last_index]).all())
```
sklam, line 727: `self.assertPreciseEqual` or `np.testing.assert_equal` will deal with nans so you don't need to skip to the non-nans.

sklam (changes requested):

numba/tests/test_sort.py
czgdp1807: Updated. Now for a give size, I test on all possible shapes for that size.

numba/tests/test_sort.py
czgdp1807: Removed. Thanks for pointing out.

sklam (dismissed):

sklam (commented):
some pep8 issues

numba/tests/test_sort.py
```diff
@@ -715,6 +716,103 @@ class TestQuicksortArrays(BaseQuicksortTest, TestCase):
...
+    def get_shapes(self, n):
+        shapes = []
+        if n == 1:
+            return shapes
+
+        for i in range(2, int(math.sqrt(n)) + 1):
+            if n%i == 0:
+                shapes.append((n//i, i))
+                shapes.append((i, n//i))
+                _shapes = self.get_shapes(n//i)
```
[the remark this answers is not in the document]
czgdp1807: Ah I see. Thanks for doing this.

stuartarchibald:
@sklam your patch 635a70d is approved. Thanks.

PR #7996, opened 2022-04-20
Fix binding logic in `@overload_glue`.

### LINKED ISSUE #7982: np.full ignores keyword argument names
## In `np.full`, numba ignores the named keyword arguments. 

```python
import numba
import numpy as np

@numba.jit(nopython=True)
def test1():
    x = np.full(shape=123, fill_value=456)
    print(x.shape)

@numba.jit(nopython=True)
def test2():
    x = np.full(fill_value=456, shape=123)
    print(x.shape)

test1()
test2()
```

outputs

```python
(123,)
(456,)  # WRONG! The shape is given as `shape=123` in the function call!
```

## Bug

Both `test1` and `test2` would be interpereted the same by numpy alone. The only difference between them is the order of the keyword arguments `fill_value` and `shape`. Numba seems to silently ignore this order. I am not sure this is a bug, but this just cost me nearly one hour of debugging. (I am still a very happy Numba user)

## Numba version

```numba                0.55.1```

### PR DESCRIPTION
**Pull request by stuartarchibald**
Updates the binding logic to accommodate calling `@overload_glue`
wrapped functions with named arguments.
    
Fixes #7982

### CODE REVIEW

File: numba/core/typing/npydecl.py [hunk 1]
```diff
@@ -1254,7 +1254,7 @@ class NumbaCArray(CallableTemplate):
     def generic(self):
         func_name = self.key.__name__
 
-        def typer(ptr, shape, dtype=types.none):
+        def typer(ptr, shape, dtype=None):
```
stuartarchibald, line 1257: This change needs some thinking about. I think the default now matches what an `@overload` would implement, but I'm not sure if this is correct WRT `@overload_glue`. There's also the question of whether the subsequent type check needs to be a `is_nonelike()` type check for the case of a `types.none` appearing, not sure if this is actually possible yet.

numba/core/overload_glue.py
```diff
@@ -151,16 +150,69 @@ def body(tyctx):
...
+            # selection of the lowering etc, so it just needs to be a list of
+            # the argument names.
+
+            if sig.pysig: # CallableTemplate, has pysig
+                pysig_params = sig.pysig.parameters
+
+                # Define the generator for the var names
+                def gen_var_names():
+                    for x in pysig_params.keys():
+                        yield x
```
sklam, line 190: Seems overly complicated to use a generator. It can just be a sequence.

numba/core/overload_glue.py
```diff
@@ -151,16 +150,69 @@ def body(tyctx):
...
+                        buf.append(k)
+                    else: # is kwarg, wire in default
+                        buf.append(f'{k} = {v.default}')
+                call_str_specific = ', '.join(buf)
+                call_str = ', '.join(pysig_params.keys())
+            else: # AbstractTemplate, need to bind 1:1 vars to the arg count
+                # Define the generator for the var names
+                def gen_var_names():
+                    for x in range(len(ol_args)):
+                        yield f'tmp{x}'
```
sklam, line 206: same here, can easily be a list-comprehension

numba/core/typing/npydecl.py
sklam: Hm... that's a problem. It will be a `types.none` when called with `c_arr = numba.carray(arr, (3,), dtype=None)`.

stuartarchibald: From OOB chat, time would be better spent translating this to `@overload` than trying to fix this edge case in `@overload_glue`, added this as issue: https://github.com/numba/numba/issues/8009

numba/core/overload_glue.py
stuartarchibald: list-comp used in 65ca6a0

numba/core/overload_glue.py
stuartarchibald: list-comp used in 65ca6a0

numba/core/overload_glue.py
```diff
@@ -27,28 +27,28 @@ def __init__(self, function, typing_key=None):
         # is all lazy.
         self._build()
 
-    def _stub_generator(self, nargs, body_func, kwargs=None):
-        """This generates a function that takes "nargs" count of arguments
-        and the presented kwargs, the "body_func" is the function that'll
+    def _stub_generator(self, body_func, varnames_gen):
```
sklam, line 30:
```suggestion
    def _stub_generator(self, body_func, varnames):
```

numba/core/overload_glue.py
```diff
@@ -27,28 +27,28 @@ def __init__(self, function, typing_key=None):
...
+        """This generates a function based on the argnames generated by the
+        varnames_gen generator, the "body_func" is the function that'll
```
sklam, line 32:
```suggestion
        """This generates a function based on the argnames provided by
        varnames, the "body_func" is the function that'll
```

numba/core/overload_glue.py
```diff
@@ -27,28 +27,28 @@ def __init__(self, function, typing_key=None):
...
-            kwargs = {}
-        # create new code parts
-        stub_code = stub.__code__
-        co_args = [stub_code.co_argcount + nargs + len(kwargs)]
 
+        stub_code = stub.__code__
         new_varnames = [*stub_code.co_varnames]
-        new_varnames.extend([f'tmp{x}' for x in range(nargs)])
-        new_varnames.extend([x for x, _ in kwargs.items()])
+        additional_var_names = [*varnames_gen]
```
sklam, line 41:
```suggestion
        additional_var_names = [*varnames]
```

sklam (changes requested):

numba/core/overload_glue.py
stuartarchibald: Thanks, fixed these and the side-effects of fixing these in 2e833f7

sklam (approved):
Thanks for the patch!

Issue #7998, opened 2022-04-21
Allow registering and using extra jit options in Numba


dlee992:
## Feature request
Hi, recently, I met a problem when coding SDC, but I think this is a general issue for Numba. Pandas have a flexible way to handle `series["A"]`:
```
import pandas as pd
s1 = pd.Series([1,2,3,4], index=list("ABCD"))
print(s1["A"])
1
s2 = pd.Series([1,2,3,4], index=list("AAAD"))
print(s2["A"])
A    1
A    2
A    3
dtype: int64
```
But when I implement this behavior on SDC, I cannot decide to return which type without runtime info. So I want to let users to tell compiler whether return series or not. 
Basically, I want to use extra options for type infer pass, since Numba doesn't allow one function with the same type inputs to return different types of outputs as above.

Can we open a "gate" to allow developers to register extra jit options in Numba? I still want to compile this function in CPU target. I don't want to override the whole Compiler class just for this simple purpose, as CUDACompiler class does.

```
class CUDAFlags(Flags):
    nvvm_options = Option(
        type=_nvvm_options_type,
        default=None,
        doc="NVVM options",
    )

@global_compiler_lock
def compile_cuda(pyfunc, return_type, args, debug=False, lineinfo=False,
                 inline=False, fastmath=False, nvvm_options=None):
    from .descriptor import cuda_target
    typingctx = cuda_target.typing_context
    targetctx = cuda_target.target_context

    flags = CUDAFlags()
```

If I misunderstand sth, please correct me. If I am right, I can help to extend Numba in this aspect.

sklam:
For the usecase described here, I would imagine the need for a custom compiler pass.  Perhaps, compiler passes should be able to export compiler flags.

dlee992:
@sklam , but Numba will throw an exception before any compiler pass goes on, since it will check options first before compiling.
```
        unused = set(options) - used
        if unused:
            # Unread options?
            m = (f"Unrecognized options: {unused}. "
                 f"Known options are {mappings.keys()}")
>           raise KeyError(m)
E           KeyError: ...
```

PR #8000, opened 2022-04-22
Add flag_class argument in jit decorator

dlee992:
As I said in the issue https://github.com/numba/numba/issues/7998, I hope Numba can allow other users/developers to register and obtain extra flags, which are not defined in CPU target flags.

But I met a segfault related typedlist:
```
@njit
def _make_list(itemty, allocated=DEFAULT_ALLOCATED):
    return listobject._as_meminfo(listobject.new_list(itemty,
                                                      allocated=allocated))
```
When calling `_make_list`, it crashed:
```
    def _parse_arg(self, lsttype, meminfo=None, allocated=DEFAULT_ALLOCATED):
        if not isinstance(lsttype, ListType):
            raise TypeError('*lsttype* must be a ListType')

        if meminfo is not None:
            opaque = meminfo
        else:
            opaque = _make_list(lsttype.item_type, allocated=allocated)
        return lsttype, opaque
```
I doesn't find the root cause, any idea about this? @sklam 

It seems I did something bad, which infecting intrinsic or overload method. But ... how to debug this kind of segfault fault?

### DIFF
--- a/numba/core/ccallback.py
+++ b/numba/core/ccallback.py
@@ -37,7 +37,7 @@ class CFunc(object):
 
     def __init__(self, pyfunc, sig, locals, options,
                  pipeline_class=compiler.Compiler,
-                 flags=compiler.Flags()):
+                 flags=None):
         args, return_type = sig
         if return_type is None:
             raise TypeError("C callback needs an explicit return type")
--- a/numba/core/ccallback.py
+++ b/numba/core/ccallback.py
@@ -36,8 +36,8 @@ class CFunc(object):
     _targetdescr = registry.cpu_target
 
     def __init__(self, pyfunc, sig, locals, options,
-                 flag_class=compiler.Flags(),
-                 pipeline_class=compiler.Compiler):
+                 pipeline_class=compiler.Compiler,
+                 flags=compiler.Flags()):
         args, return_type = sig
         if return_type is None:
             raise TypeError("C callback needs an explicit return type")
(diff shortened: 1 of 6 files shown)


numba/core/ccallback.py
```diff
@@ -36,7 +36,8 @@ class CFunc(object):
     _targetdescr = registry.cpu_target
 
     def __init__(self, pyfunc, sig, locals, options,
-                 pipeline_class=compiler.Compiler):
+                 pipeline_class=compiler.Compiler,
+                 flags=compiler.Flags()):
```
sklam, line 40: This is probably sharing the flag object.

numba/core/dispatcher.py
```diff
@@ -805,7 +809,9 @@ class attribute.
     __numba__ = 'py_func'
 
     def __init__(self, py_func, locals={}, targetoptions={},
-                 impl_kind='direct', pipeline_class=compiler.Compiler):
+                 impl_kind='direct',
+                 pipeline_class=compiler.Compiler,
+                 flags=compiler.Flags()):
```
sklam, line 814: same here. sharing of flags

sklam: This should be written as:

```python
def __init__(..., flags=None):
    ...
    flags = compiler.Flags() if flags is None else flags
```

numba/tests/test_flags.py
```diff
@@ -0,0 +1,39 @@
...
+    def test_define_flag_class(self):
+        my_flags = MyFlags()
+
+        # extremely strange. If I put empty_list outside JIT, segfaults.
+        # l = List.empty_list(numba.int64)
+
+        @jit(flags=my_flags, debug=True)
+        def foo():
+            # put it here, don't segfault.
+            l = List.empty_list(numba.int64)
```
sklam, line 30: this segfault problem is solved by: https://github.com/numba/numba/pull/8000/files#r859196496

dlee992: Many thanks for your time! I updated these codes as u pointed out.

numba/core/dispatcher.py
dlee992: Yes, I made a silly mistake...

PR #8001, opened 2022-04-22
CUDA fp16 math functions

testhound:
This pull request adds support for the following floating point 16 math functions:

- hsin
- hcos
- hlog
- hlog10
- hlog2
- hexp
- hexp10
- hexp2
- hfloor
- hceil
- hsqrt
- hrsqrt
- hrcp
- hrint
- htrunc

NOTE: This pull request is based on this PR: https://github.com/numba/numba/pull/7621

which is pending APPROVAL as of the creation date of this PR.


gmarkall:
gpuci run tests

gmarkall:
The tests need to be avoided when the ctypes binding is in use - additionally, we need to provide a more appropriate error message when an attempt to use the fp16 math functions when using the ctypes bindings is made - at present the error states that linking .cu files is not supported with the ctypes binding, which is confusing for the user because they didn't explicitly attempt to link a .cu file.

isVoid (commented):
Some comments on the documentation:
1. We should avoid unnecessary abbreviations in docs.
2. "base 10 logarithm" is the natural spelling of `log10` in texts.
3. "Calculating a logarithm" sounds better than "calculating an operation".

The below are some example suggestions.

testhound:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
(I've just merged main into this branch to do the review, because Github is confused about how to generate the diff since the merge of #7621)

numba/cuda/cudadecl.py
```diff
@@ -361,6 +361,39 @@ class Cuda_fp16_binary(ConcreteTemplate):
 Cuda_hneg = _genfp16_unary(cuda.fp16.hneg)
 Cuda_habs = _genfp16_unary(cuda.fp16.habs)
 
+hsin_device = declare_device_function('hsin_wrapper', types.float16,
```
gmarkall, line 364: Rather than adding all these at global scope, I think it would save some boilerplate to declare a function in `CudaFp16Template` like:

```python
    def _resolve_wrapped_unary(self, fname):
        decl = declare_device_function(f'{fname}_wrapper',
                                       types.float16,
                                       (types.float16,))
        return types.Function(decl)
```

then use it as shown below for `hsin`.

numba/cuda/cudadecl.py
```diff
@@ -529,6 +565,51 @@ def resolve_habs(self, mod):
     def resolve_hfma(self, mod):
         return types.Function(Cuda_hfma)
 
+    def resolve_hsin(self, mod):
+        return hsin_device
```
gmarkall, line 569: Example use of `resolve_wrapped_unary`:

```suggestion
        return self._resolve_wrapped_unary('hsin')
```

numba/cuda/dispatcher.py
```diff
@@ -96,6 +96,26 @@ def __init__(self, py_func, argtypes, link=None, debug=False,
         if self.cooperative:
             link.append(get_cudalib('cudadevrt', static=True))
 
+        cuda_fp16_math_funcs = ['hsin_wrapper', 'hcos_wrapper',
```
gmarkall, line 99: I'd suggest moving this to global scope - in future the list will get longer. It would also be nice to avoid baking the `_wrapper` part into the striongs so we can just list the actual function names.

numba/cuda/functions.cu
```diff
@@ -0,0 +1,211 @@
+#include "cuda_fp16.h"
```
gmarkall, line 1: I think `package_data` in `setup.py` will need modifying to ensure that this file is present in installations, otherwise we won't be able to use the wrapper outside of running from source.

numba/cuda/functions.cu
```diff
@@ -0,0 +1,211 @@
...
+
+extern "C" __device__ int
+hdiv_wrapper(
```
gmarkall, line 4: I'm not sure `_wrapper` is a sufficient guard against name collisions - can we rename these so they're of the form `__numba_wrapper_{fn}` instead please?

numba/cuda/functions.cu
gmarkall: Also, perhaps `cpp_function_wrappers.cu` would be a more descriptive name for this file?

numba/cuda/functions.cu
```diff
@@ -0,0 +1,211 @@
...
+{
+  __half retval = __hdiv(__short_as_half (x), __short_as_half (y));
+  
+  *return_value = __half_as_short (retval);
+  // Signal that no Python exception occurred
+  return 0;
+}
+
+extern "C" __device__ int
+hsin_wrapper(
```
gmarkall, line 18: Although C preprocessor macros are generally to be avoided where possible, I can't help but think it'd be a worthwhile tradeoff to implement a macro for all the unary function wrappers, rather than duplicating lots of boilerplate. What do you think?

Alternatively perhaps a template function with multiple explicit instantiations would be nicer, but I'm not sufficiently experience in using NVRTC and/or extern C functions to see ahead of trying it whether that could actually work in this situation - could that be worth a try?

gmarkall (changes requested):
I've added some comments / thoughts on the diff - it's not a full review, but as a first pass I hope it can help things move in the right direction.

I think we'll also need to document that these functions are only available when using the CUDA Python binding, and have a reasonable failure mode for when the ctypes binding is in use.

testhound:
gpuci run tests

numba/cuda/functions.cu
testhound: Done.

numba/cuda/functions.cu
testhound: Done.

numba/cuda/functions.cu
testhound: Implemented both comments.

numba/cuda/cudadecl.py
testhound: @gmarkall I partially implemented your recommendation: I created two global functions _resolve_wrapped_unary and _resolve_wrapped_binary. I had to implement and use them at global scope because device functions needs to be singletions otherwise you get ambigious errors for multiple singatures.

numba/cuda/compiler.py
```diff
@@ -289,7 +289,7 @@ def compile_ptx_for_current_device(pyfunc, args, debug=False, lineinfo=False,
                        device=device, fastmath=fastmath, cc=cc, opt=True)
 
 
-def declare_device_function(name, restype, argtypes):
+def declare_device_function(name, restype, argtypes, ret_template=False):
```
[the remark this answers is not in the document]
testhound: Done

testhound:
gpuci run tests

testhound:
gpuci run tests

testhound:
gpuci run tests

testhound:
gpuci run tests

numba/cuda/dispatcher.py
```diff
@@ -96,6 +106,17 @@ def __init__(self, py_func, argtypes, link=None, debug=False,
         if self.cooperative:
             link.append(get_cudalib('cudadevrt', static=True))
 
+        res = [ele for ele in cuda_fp16_math_funcs
```
gmarkall, line 109: `ele` is a strange choice of name for the loop variable here. Perhaps `f` or `fn` are a closer representation of what the variable holds?

gmarkall (changes requested):
I've had another look at this today with the aim of moving this forward another iteration. Some questions / comments:

- Do we need all the duplication for testing every function in test_intrinsics.py? Can we reduce this a bit? Perhaps with subtests that iterate through the functions to test?
- Similarly there's some duplication in the division tests, in particular of the test values and the associated documentation-like comments. Could the division tests be re-organised to reduce the need to copy / paste test values and documentary comments?
- `cpp_function_wrappers.cu` still duplicates almost everything across every function. I imagined that the use of a template or macro could go further such that you could define each of the wrapper functions using only a macro instantiation, and not have to write the entire function manually in each case. Can this be done? Is there anything preventing this that I've missed?
- There's a few "random" whitespace changes in this PR to unrelated files or sections of files (e.g. device_init.py).
- There are some comments from @isVoid on the diff that appear to be unaddressed at present.
- In addition, there are one or two new comments on the diff that I've added here.

Once the above are addressed / discussed I'll go through another review.

testhound:
gpuci run tests

testhound:
@gmarkall I updated the code to reflect your comments as well as the outstanding ones from @isVoid. Note that I removed the additional divide operator tests because the fp16 divide operator is implemented via the fp16 divide intrinsic and those tests are duplicates. Additionally I removed the corner case checking test case as that was only required when the divide algorithm was being code via inline assembly.

numba/cuda/stubs.py
```diff
@@ -584,6 +594,156 @@ class habs(Stub):
...
+        on fp16 operands only.
+
+        Returns the logarithm base 2 result.
+
+        """
+
+    class hexp(Stub):
+        """hexp(a)
+
+        Calculate natural exponential, (2.71828 ** a) in round to nearest mode.
```
gmarkall, line 650: I think this calculates with more precision than the docstring implies. How about:

```suggestion
        Calculate natural exponential, exp(a), in round to nearest mode.
```

numba/cuda/stubs.py
```diff
@@ -584,6 +594,156 @@ class habs(Stub):
...
+        operands only.
+
+        Returns the rounded result.
+
+        """
+
+    class htrunc(Stub):
+        """htrunc(a)
+
+        Truncate the input argument to it's integer portion. Supported
```
gmarkall, line 740:
```suggestion
        Truncate the input argument to its integer portion. Supported
```

gmarkall (changes requested):
Thanks for the updates. On the whole I think this looks good, there are some minor points suggested on the diff.

testhound:
@gmarkall all requested changes committed.

testhound:
gpuci run tests

gmarkall (approved):
Many thanks for the updates - I think this is looking good now.

sklam:
BFID `numba_smoketest_cuda_yaml_163`

gmarkall:
@sklam How did the buildfarm run go?

gmarkall:
@sklam Do we have a buildfarm result?

sklam:
The buildfarm has never fully finish running this PR due to interruptions from unstable machines and the release. It has passed enough linux tests, but i'm still trying to get it run on windows machines.

gmarkall:
Hopefully #8572 will help with testing on the buildfarm - perhaps this could have another try once that is merged?

sklam:
> Hopefully #8572 will help with testing on the buildfarm - perhaps this could have another try once that is merged?

Good idea. Reruning test at `numba_smoketest_cuda_yaml_171`

sklam:
> > Hopefully #8572 will help with testing on the buildfarm - perhaps this could have another try once that is merged?
> 
> Good idea. Reruning test at `numba_smoketest_cuda_yaml_171`

All green

PR #8004, opened 2022-04-25
CUDA fixes for Windows

gmarkall:
This fixes a couple of issues on Windows:

- IPC is not actually supported on Windows, there was just a small range of drivers where it appeared to work in some circumstances. Because it is not properly supported, we limit its support back to just Linux again - this includes skipping the tests, erroring if an attempt to use IPC on non-Linux is used, and updating the documentation.
- Sometimes on Windows if `nvdisasm` is not found, the `FileNotFoundError`'s filename is blank. There doesn't appear to be any value in checking for `nvdisasm` in the name, and when it's not found it resulted in incorrect logic, so I just removed the check for the filename.

One issue is still outstanding on Windows with recent drivers (tests of printing functionality, #8005), which will be fixed in a separate PR.

### DIFF
--- a/docs/source/cuda/ipc.rst
+++ b/docs/source/cuda/ipc.rst
@@ -7,6 +7,10 @@ Sharing CUDA Memory
 Sharing between process
 =======================
 
+Sharing between processes is implemented using the Legacy CUDA IPC API
+(functions whose names begin with ``cuIpc``), and is supported only on Linux.
+
+
 Export device array to another process
 --------------------------------------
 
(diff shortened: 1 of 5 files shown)


gmarkall:
gpuci run tests

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_123`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_123`.

Passed.

PR #8010, opened 2022-04-27
Add support for fp16 comparison native operators

testhound:
This pull request adds support for the fp16 comparison operators (>, >=, <, <=, ==, !=). Note this PR depends on a not yet merged PR: https://github.com/numba/numba/pull/7556.


gmarkall:
The following:

```python
from numba import cuda
import numpy as np

@cuda.jit
def f(r, x ,y):
    i = cuda.grid(1)
    r[i] = x[i] < y[i]

N = 1
x = np.ones(N, dtype=np.float16)
y = np.ones(N, dtype=np.float32) + 1.0e-6
r = np.zeros(N, dtype=np.bool)

f[1, N](r, x, y)

print(r)
print(x < y)
```

prints:

```
[False]
[ True]
```

i.e. the kernel seems to determine that `1 < 1.000001` is `False`. With the `main` branch the kernel evaluates the result to `True`. Am I doing anything incorrect in this experiment?

numba/cuda/cudadecl.py
```diff
@@ -355,11 +355,50 @@ class Cuda_fp16_binary(ConcreteTemplate):
...
+
+        cases = [
+            signature(types.b1, types.float16, types.float16)
+        ]
+    return Cuda_fp16_cmp
+
+
+def _genfp16_comparison_operator(l_key):
+    @register_global(l_key)
+    class Cuda_fp16_cmp2(AbstractTemplate):
```
gmarkall, line 371: Why the `2`?

numba/cuda/cudadecl.py
```diff
@@ -355,11 +355,50 @@ class Cuda_fp16_binary(ConcreteTemplate):
...
+        key = l_key
+
+        def generic(self, args, kws):
+            assert not kws
+            if len(args) == 2 and \
+                    (args[0] == types.float16 or args[1] == types.float16):
+                return signature(types.b1, types.float16, types.float16)
+
+    return Cuda_fp16_cmp2
```
gmarkall, line 380: I don't think there's any need to return the generated class (since we don't need to assign the result of this function to anything).

gmarkall (changes requested):
I've had a quick pass over this. In addition to the issue where some comparisons appear to evaluate incorrectly (in my previous comment), I think the testing is a bit too rudimentary:

- What happens if there are multiple operator comparisons in a single kernel / function?
- How do comparisons between `float16` and other types (e.g. `float32`, `int16`, etc.) turn out?

There are also some comments and questions on the diff.

gmarkall:
gpuci run tests

testhound:
> The following:
> 
> ```python
> from numba import cuda
> import numpy as np
> 
> @cuda.jit
> def f(r, x ,y):
>     i = cuda.grid(1)
>     r[i] = x[i] < y[i]
> 
> N = 1
> x = np.ones(N, dtype=np.float16)
> y = np.ones(N, dtype=np.float32) + 1.0e-6
> r = np.zeros(N, dtype=np.bool)
> 
> f[1, N](r, x, y)
> 
> print(r)
> print(x < y)
> ```
> 
> prints:
> 
> ```
> [False]
> [ True]
> ```
> 
> i.e. the kernel seems to determine that `1 < 1.000001` is `False`. With the `main` branch the kernel evaluates the result to `True`. Am I doing anything incorrect in this experiment?

@gmarkall based on conversations with @sklam about argument promotion we determined that '1.000001' is not representable in float16. 

We also determined that the code should have been promoted to float32 (wider mode). I will update the code when I return from vacation.

testhound:
gpuci run tests

testhound:
gpuci run tests

numba/cuda/cudadecl.py
```diff
@@ -355,11 +356,56 @@ class Cuda_fp16_binary(ConcreteTemplate):
...
+
+def _genfp16_comparison_operator(l_key):
+    @register_global(l_key)
+    class Cuda_fp16_operator_cmp(AbstractTemplate):
+        key = l_key
+
+        def generic(self, args, kws):
+            assert not kws
+
+            if len(args) == 2 and \
```
gmarkall, line 378: Can you explain the thinking behind the logic here? Some questions that spring to mind when I read this implementation:

- Why not just explicitly require both arguments to be float16? If they are convertible, won't a cast be inserted by typing? 
- What types are convertible to float16 safely that would result in `convertible` being `True`?

gmarkall: Further to this, I notice that the following implementation passes all tests:

```python
          def generic(self, args, kws):
              assert not kws  
              if args == (types.float16, types.float16):
                  return signature(types.b1, types.float16, types.float16)
```

Does this imply that there is some missing test coverage, or that the simple implementation suffices? (my suspicion is that the simple implementation suffices, but I don't think I've understood yet why the more complex implementation could be needed, so I'm not certain either way).

gmarkall (changes requested):
Many thanks for the update - I note that the changes are relatively small on top of #7556, so there is just one area commented on the diff - my suspicion is that the typing implementation could be simplified. I think everything else looks OK.

numba/cuda/cudadecl.py
testhound: @gmarkall requiring both arguments to be float16 will fail, forcing one argument to be float16 ended up generating incorrect mixed type code. As an example comparing a float16 and float32 converted the float32 to float16 and then doing the comparison which is incorrect. The types int16 and int8 can be converted to float16.

gmarkall: Thankyou for the information - however, I don't really understand the explanation. I have the following patch applied on top of your commit d9abed603:

```diff
diff --git a/numba/cuda/cudadecl.py b/numba/cuda/cudadecl.py
index c0ef42652..a54740efd 100644
--- a/numba/cuda/cudadecl.py
+++ b/numba/cuda/cudadecl.py
@@ -5,7 +5,6 @@ from numba.core.typing.templates import (AttributeTemplate, ConcreteTemplate,
                                          AbstractTemplate, CallableTemplate,
                                          signature, Registry)
 from numba.cuda.types import dim3, grid_group
-from numba.core.typeconv import Conversion
 from numba import cuda
 import operator
 
@@ -374,17 +373,8 @@ def _genfp16_comparison_operator(l_key):
 
         def generic(self, args, kws):
             assert not kws
-
-            if len(args) == 2 and \
-                    (args[0] == types.float16 or args[1] == types.float16):
-                if (args[0] == types.float16):
-                    convertible = self.context.can_convert(args[1], args[0])
-                else:
-                    convertible = self.context.can_convert(args[0], args[1])
-
-                if (convertible == Conversion.exact) or \
-                        (convertible == Conversion.promote):
-                    return signature(types.b1, types.float16, types.float16)
+            if args == (types.float16, types.float16):
+                return signature(types.b1, types.float16, types.float16)
 
 
 Cuda_hadd = _genfp16_binary(cuda.fp16.hadd)
```

I run the testsuite like this:

```
python -m numba.runtests numba.cuda.tests  -m
```

and I get the following summary as a result:

```
Ran 1295 tests in 115.148s

OK (skipped=18, expected failures=8)
```

From this I can see three possibilities:

- The implementation could just require both arguments to be `float16`, and everything will still work. If this is the case, then please could you change to the simple implementation outlined above.
- Requiring both arguments to be `float16` is in fact incorrect, but there is no test that demonstrates the need for the more complex implementation in this PR. If this is the case, could you please add a test that will pass with the implementation in this PR, and fail with the simple implementation outlined above.
- There's something else going on that I don't understand - if this is the case, could you please help me to understand where I'm going wrong? :-)

testhound: @gmarkall I added two test cases that verify the generated ptx. One of these test cases fails when using your code that requires both arguments to be float16.

testhound:
gpuci run tests

numba/core/typeconv/rules.py
```diff
@@ -30,9 +30,11 @@ def _init_casting_rules(tm):
     tcr.safe_unsafe(types.uint16, types.int32)
     tcr.safe_unsafe(types.uint32, types.int64)
 
+    tcr.safe_unsafe(types.int8, types.float16)
```
gmarkall, line 33: Is conversion from `int16` to `float16` also possible? I note from the limits of a `float16` and `int16` that the range of values of an `int16` seems to fit into the range of a `float16`:

```
print(np.finfo(np.float16))
# Machine parameters for float16
# ---------------------------------------------------------------
# precision =   3   resolution = 1.00040e-03
# machep =    -10   eps =        9.76562e-04
# negep =     -11   epsneg =     4.88281e-04
# minexp =    -14   tiny =       6.10352e-05
# maxexp =     16   max =        6.55040e+04
# nexp =        5   min =        -max
# ---------------------------------------------------------------

print(np.iinfo(np.int16))
# Machine parameters for int16
# ---------------------------------------------------------------
# min = -32768
# max = 32767
# ---------------------------------------------------------------
```

gmarkall: Actually nevermind, I see that that conversion is already registered on line L36/R38.

numba/cuda/cudadecl.py
gmarkall: Thanks - this now makes sense. I now recall that the complexity here is to work around the fact that we can't have multiple concrete templates for a single global. I think what's going on here is going to be quite unclear to anyone coming across this for the first time - can you please add some comments explaining the background and how the workaround works to the `Cuda_fp16_operator_cmp` class?

numba/core/typeconv/rules.py
testhound: @gmarkall I modeled int16 to float16 after int32 to float32 which is considered unsafe:

tcr.unsafe_unsafe(types.int32, types.float32)

numba/cuda/tests/cudapy/test_operator.py
```diff
@@ -36,6 +100,137 @@ def test_truediv(self):
...
+                     simple_fp16_lt, simple_fp16_le,
+                     simple_fp16_eq, simple_fp16_ne)
+        ops = (operator.gt, operator.ge, operator.lt, operator.le,
+               operator.eq, operator.ne)
+
+        for fn, op in zip(functions, ops):
+            with self.subTest(op=op):
+                kernel = cuda.jit("void(b1[:], f2, f2)")(fn)
+
+                expected = np.zeros(1, dtype=np.bool8)
```
gmarkall, line 115: This appears to be superfluous - it is re-defined later.

numba/cuda/tests/cudapy/test_operator.py
```diff
@@ -36,6 +100,137 @@ def test_truediv(self):
...
+    @skip_unless_cc_53
+    def test_mixed_fp16_comparison(self):
+        functions = (simple_fp16_gt, simple_fp16_ge,
+                     simple_fp16_lt, simple_fp16_le,
+                     simple_fp16_eq, simple_fp16_ne)
+        ops = (operator.gt, operator.ge, operator.lt, operator.le,
+               operator.eq, operator.ne)
+        types = (np.int8, np.int16, np.int32, np.int64,
+                 np.float32, np.float64)
+        for fn, op, ty in zip(functions, ops, types):
```
gmarkall, line 133: I don't think zipping the types with the `fn` and `op` has the intended effect here. Did you mean to test with the cross-product of `(fn, op)` pairs with the types?

numba/cuda/tests/cudapy/test_operator.py
```diff
@@ -36,6 +100,137 @@ def test_truediv(self):
...
+            with self.subTest(op=op):
+                kernel = cuda.jit(fn)
+
+                expected = np.zeros(1, dtype=np.bool8)
```
gmarkall, line 137: Looks like another superfluous assignment?

numba/cuda/tests/cudapy/test_operator.py
```diff
@@ -36,6 +100,137 @@ def test_truediv(self):
...
+    def test_mixed_fp16_comparison(self):
+        functions = (simple_fp16_gt, simple_fp16_ge,
+                     simple_fp16_lt, simple_fp16_le,
+                     simple_fp16_eq, simple_fp16_ne)
+        ops = (operator.gt, operator.ge, operator.lt, operator.le,
+               operator.eq, operator.ne)
+        types = (np.int8, np.int16, np.int32, np.int64,
+                 np.float32, np.float64)
+        for fn, op, ty in zip(functions, ops, types):
+            with self.subTest(op=op):
```
gmarkall, line 134: I think the subtest should be parameterised on both the operation and the type (assuming my assumption about the test being with all pairs of operators and types is correct)

numba/cuda/tests/cudapy/test_operator.py
```diff
@@ -36,6 +100,137 @@ def test_truediv(self):
...
+                    operator.lt:'setp.lt.f16',
+                    operator.le:'setp.le.f16',
+                    operator.eq:'setp.eq.f16',
+                    operator.ne:'setp.ne.f16'}
+        for (fn, op), ty in itertools.product(zip(functions, ops),
+                                              types_convert):
+            with self.subTest(op=op):
+                args = (b1[:], f2, from_dtype(ty))
+                ptx, _ = compile_ptx(fn, args, cc=(5, 3))
+                print(ptx)
```
gmarkall, line 200: No need to print the ptx here.

numba/cuda/tests/cudapy/test_operator.py
```diff
@@ -36,6 +100,137 @@ def test_truediv(self):
...
+                     test_multiple_hcmp_5)
+        for fn in functions:
+            with self.subTest(fn=fn):
+                compiled = cuda.jit("void(b1[:], f2, f2, f2)")(fn)
+                ary = np.zeros(1, dtype=np.bool8)
+                arg1 = np.float16(2.)
+                arg2 = np.float16(3.)
+                arg3 = np.float16(4.)
+                compiled[1, 1](ary, arg1, arg2, arg3)
+                self.assertTrue(ary[0])
```
gmarkall, line 161: It would be good to also test a case where the expected result is not true - a bug which manifests as multiple comparisons always results in `True` would not be caught by this test.

numba/cuda/tests/cudapy/test_operator.py
```diff
@@ -36,6 +100,137 @@ def test_truediv(self):
...
+
+        opstring = {operator.gt:'setp.gt.f16',
+                    operator.ge:'setp.ge.f16',
+                    operator.lt:'setp.lt.f16',
+                    operator.le:'setp.le.f16',
+                    operator.eq:'setp.eq.f16',
+                    operator.ne:'setp.ne.f16'}
+        for (fn, op), ty in itertools.product(zip(functions, ops),
+                                              types_convert):
+            with self.subTest(op=op):
```
gmarkall, line 197: This subtest should also be parameterised on the type as well as the op.

numba/cuda/tests/cudapy/test_operator.py
```diff
@@ -36,6 +100,137 @@ def test_truediv(self):
...
+                self.assertIn(s, ptx)
+
+    @skip_on_cudasim('Compilation unsupported in the simulator')
+    def test_mixed_fp16_comparison_ptx(self):
+        functions = (simple_fp16_gt, simple_fp16_ge,
+                     simple_fp16_lt, simple_fp16_le,
+                     simple_fp16_eq, simple_fp16_ne)
+        ops = (operator.gt, operator.ge, operator.lt, operator.le,
+               operator.eq, operator.ne)
+        types_convert = (np.int8,)
```
gmarkall, line 187: Is this missing some types?

numba/cuda/tests/cudapy/test_operator.py
```diff
@@ -36,6 +100,137 @@ def test_truediv(self):
...
+                    operator.eq:'setp.eq.',
+                    operator.ne:'setp.neu.'}
+        opsuffix = {np.dtype('int32'): 'f64',
+                    np.dtype('int64'): 'f64',
+                    np.dtype('float32'): 'f32',
+                    np.dtype('float64'): 'f64'}
+
+        for (fn, op), ty in itertools.product(zip(functions, ops),
+                                              types_promote):
+            with self.subTest(op=op):
```
gmarkall, line 226: I think this should also parameterise on the op and type.

numba/cuda/tests/cudapy/test_operator.py
```diff
@@ -36,6 +100,137 @@ def test_truediv(self):
...
+        for (fn, op), ty in itertools.product(zip(functions, ops),
+                                              types_convert):
+            with self.subTest(op=op):
+                args = (b1[:], f2, from_dtype(ty))
+                ptx, _ = compile_ptx(fn, args, cc=(5, 3))
+                print(ptx)
+                self.assertIn(opstring[op], ptx)
+
+    @skip_on_cudasim('Compilation unsupported in the simulator')
+    def test_mixed_fp16_comparison_ptx2(self):
```
gmarkall, line 204: This might be a more appropriate name:

```suggestion
    def test_mixed_fp16_comparison_promotion_ptx(self):
```

(seeing that `test_mixed_fp16_comparison_promotion_ptx` as opposed to `test_mixed_fp16_comparison_ptx2` has failed gives more information/context)

numba/cuda/cudadecl.py
```diff
@@ -366,6 +367,28 @@ class Cuda_fp16_cmp(ConcreteTemplate):
...
+                    (args[0] == types.float16 or args[1] == types.float16):
+                if (args[0] == types.float16):
+                    convertible = self.context.can_convert(args[1], args[0])
+                else:
+                    convertible = self.context.can_convert(args[0], args[1])
+
+                #print(args, convertible)
```
gmarkall, line 385: Please remove commented code.

```suggestion
                #print(args, convertible)
```

gmarkall (changes requested):
Many thanks for the updates - I think the implementation itself is looking good. There are some comments about the the tests on the diff. In addition:

- Several of the tests use random data - could you please add a `setUp()` method to the class that seeds the RNG consistently across runs (or make sure it is seeded in every test that uses random data).
- The reason for the complexity in `_genfp16_comparison_operator` is totally non-obvious (hence the long discussion around it) - can  you please add commentary to it that explains why it works the way it does (as requested in https://github.com/numba/numba/pull/8010/files#r882689328)

testhound:
gpuci run tests

testhound:
@gmarkall I updated the PR with all your requested changes.

numba/cuda/tests/cudapy/test_operator.py
gmarkall: @testhound What are your thoughts on this comment?

numba/cuda/tests/cudapy/test_operator.py
gmarkall: Why did you change the name of the test to `test_fp16_int8_comparison_ptx`? Is there only value in testing with `int8`? I'm trying to understand the reasoning behind testing only this particular type.

If it does only make sense to test with `int8` here, then I don't think you need to iterate over the product of the functions/ops and `types_convert` at all.

testhound: @gmarkall this test is only comparing against int8 to match the "safe" conversion part of this if condition:

https://github.com/testhound/numba/blob/0b6d045a6c814fa046bb5a5f0c6bb2923712d9e5/numba/cuda/cudadecl.py#L407

I will modify the code to remove the iteration.

testhound: Code update to remove iteration and comment clarification.

numba/cuda/tests/cudapy/test_operator.py
testhound: Additional test case added.

numba/cuda/cudadecl.py
```diff
@@ -378,6 +379,39 @@ class Cuda_fp16_cmp(ConcreteTemplate):
     return Cuda_fp16_cmp
 
 
+# Until issue 7863 is resolved, if multiple concrete templates are utilized
+# function resolution will pick the wrong implementation and errors will occur.
+# We avoid this by using ConcreteTempalte for the intrinisics and
+# AbstractTemplate
```
gmarkall, line 385: Thanks for adding more description. I'm still concerned that it's difficult for the reader to understand the situation because the comment is quite vague and this is a complex issue. In particular my thoughts on the explanation are:

- It's not clear what "multiple concrete templates are utilized" means - I know right now that it's for a single function, but in future I may not remember, and someone else may not know.
- Why does function resolution pick the wrong implementation?
- What errors occur? Failure to compile, incorrect behaviour of compiled code, or just suboptimal performance?
- `ConcreteTemplate` and `intrinsics` are mis-spelled.

I know we're going through a lot of iterations here, but my concern is that this code is going to live for a long time and it should be evident from looking only at the code in `cudadecl.py` what is going on for readers into the future - having to go back and dig through old pull requests and comments is inconvenient enough, but especially so when trying to decipher what is happening and why for such a complex problem that we're dealing with here (especially when it's in general very hard to understand typing and function resolution in Numba at the best of times).

I'd suggest using the following comment to describe the situation - what do you think of it?

```suggestion
# If multiple ConcreteTemplates provide typing for a single function, then
# function resolution will pick the first compatible typing it finds even if it
# involves inserting a cast that would be considered undesirable (in this
# specific case, float16s could be cast to float32s for comparisons).
#
# To work around this, we instead use an AbstractTemplate that implements
# exactly the casting logic that we desire. The AbstractTemplate gets
# considered in preference to ConcreteTemplates during typing.
#
# This is tracked as Issue #7863 (https://github.com/numba/numba/issues/7863) -
# once this is resolved it should be possible to replace this AbstractTemplate
# with a ConcreteTemplate to simplify the logic.
```

gmarkall (changes requested):
Many thanks for the updates. There is one remaining comment / suggestion about the description of the workaround for typing - after that is resolved, I think this should be good to approve.

testhound:
gpuci run tests

gmarkall (approved):
Many thanks for the update. I think this now looks good.

gmarkall:
(I think the CI issue here is the general one right now, not to do with this PR).

esc:
> (I think the CI issue here is the general one right now, not to do with this PR).

This is (was) correct and was related to an outage on anaconda.org

gmarkall:
gpuci run tests

stuartarchibald (approved):

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_144`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_144`.

Passed.

PR #8014, opened 2022-04-29
Fix support for {real,imag} array attrs in Parfors.

### LINKED ISSUE #8012: Parallel fails when using .real and .imag attributes with numpy functions
## Reporting a bug

```
import numpy as np
from numba import njit

@njit(parallel=True)
def fails(z):
    return np.sum(z.real**2 + z.imag**2)

@njit(parallel=False)
def works(z):
    return np.sum(z.real**2 + z.imag**2)

@njit(parallel=True) 
def also_works(z):
    return np.sum(np.real(z)**2 + np.imag(z)**2)

n = 10
z = np.arange(n) + 1j*np.arange(n)
print(works(z))
print(also_works(z))
print(fails(z)) # this one throws an error
```

`parallel=True` fails when sending values derived from `.real` and `.imag` array attribute to a numpy reduction or math function. The example given here uses `np.sum()`, but other functions also show the same error (`np.prod`, `np.min`, `np.max`, `np.argmin`, `np.argmax`, `np.mean`, `np.var`, and `np.std`). Using `np.real()` and `np.imag()` to access the real and imaginary components, instead of `.real` and `.imag` array attributes, works as expected.

stuartarchibald:
As title. ArrayAnalysis was missing handling for real and imag
attributes on arrays. Adds tests for equivalence analysis and
parfors generation/execution.

Fixes #8012

### DIFF
--- a/numba/parfors/array_analysis.py
+++ b/numba/parfors/array_analysis.py
@@ -1579,6 +1579,10 @@ class ArrayAnalysis(object):
         elif expr.attr == "shape":
             shape = equiv_set.get_shape(expr.value)
             return ArrayAnalysis.AnalyzeResult(shape=shape)
+        elif expr.attr in ("real", "imag") and self._isarray(expr.value.name):
+            # Shape of real or imag attr is the same as the shape of the array
+            # itself.
+            return ArrayAnalysis.AnalyzeResult(shape=expr.value)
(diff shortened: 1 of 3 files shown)


sklam (approved):
Thanks for the quick patch!

PR #8016, opened 2022-05-01
[Docs] [Very Minor] Make `numba.jit` boundscheck doc line consistent

martinky24:
The rest of the docs for `jit` above on this same page use italicized paramater names instead of code block. Similar story with the boolean at the start, the rest of the page uses lowercase word.

Very minor nitpicky stuff, but hopefully an improvement nevertheless.


stuartarchibald (approved):
This makes the doc style more consistent! Thanks for the patch.

PR #8017, opened 2022-05-02
Update FAQ to include details about using debug-only option

### LINKED ISSUE #7358: Debug output for LLVM loop-vectorize needs an LLVM build with assertions
I see comments suggesting adding this to understand how loops are being handled by numba, and in the their own FAQ (https://numba.pydata.org/numba-doc/latest/user/faq.html)

    llvm.set_option('','--debug-only=loop-vectorize')

You would then create your njit function and run it, and I believe the idea is that it prints debug information about whether (and why) the loop might be vectorized.

    @njit
    def test(a):
        for i in range(a.shape[0]):
            a[i] = a[i] * 2

    print(test(np.asarray([1.0, 2.0, 3.0])

I find that this code changes nothing - there is no additional debug output from llvm indicating whether a loop is being vectorized / enhanced with SSE instructions, or if not why not.

I’ve tried this on Windows, Linux and Mac, all using Anaconda distributions of Python 3.8 and with the conda, conda-forge and numba channels.

numba version 0.53.1
llvmlite version 0.36.0

guilhermeleobas:
Fix #7358. This PR includes an info section to use llvmlite from numba channel in order for `debug-only` to work.

Edit: example
image

### DIFF
--- a/docs/source/user/faq.rst
+++ b/docs/source/user/faq.rst
@@ -188,6 +188,12 @@ add the following lines:
 This tells LLVM to print debug information from the **loop-vectorize**
 pass to stderr.  Each function entry looks like:
 
+
+.. note::
+   Using ``--debug-only`` requires LLVM to be build with assertions enabled to
+   work. Use the build of llvmlite in the `Numba channel <https://anaconda.org/numba/llvmlite>`_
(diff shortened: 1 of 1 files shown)


sklam (approved):
LGTM

PR #8020, opened 2022-05-03
Add support for `array.resize`

### LINKED ISSUE #8015: "resize" attribute of numpy array
## Feature request
Dear Developer Team,

Could you please add the "resize" attribute to numpy arrays? When increasing the size of a numpy array, "resize" seems to be more memory-efficient than creating a new numpy array and deleting the old one.

Best,

R.H.

czgdp1807:
Closes #8015 

I have added initial support `array.resize` for numpy arrays in Numba. I have taken inspiration from `reshape` implementation. I have considered 3 cases,

1. New size of array is less than or equal to the original size of array - Then there is no need to allocation extra amount of data. Just fill the shape and strides (according to order of the input array).
2. New size of arrays is greater than the size of the original size of array - Here, other than filling the shape and strides according to the new shape, we have to allocate new block of memory and then fill the extra values with zero.

I have currently implemented and tested the first case. Will update the PR for the second case as well.

Please let me know if you have any other approach in mind. Will be happy to implement that as well. Thanks.

### DIFF
--- a/numba/core/typing/arraydecl.py
+++ b/numba/core/typing/arraydecl.py
@@ -414,6 +414,49 @@ class ArrayAttribute(AttributeTemplate):
             retty = ary.copy(ndim=len(args))
             return signature(retty, *args)
 
+    @bound_function("array.resize")
+    def resolve_resize(self, ary, args, kws):
+        def sentry_shape_scalar(ty):
+            if ty in types.number_domain:
+                # Guard against non integer type
+                if not isinstance(ty, types.Integer):
+                    raise TypeError("resize() arg cannot be {0}".format(ty))
+                return True
+            else:
+                return False
+
+        assert not kws
+        if ary.layout not in 'CF':
+            # only work for contiguous array
+            raise TypeError("resize() supports contiguous array only")
+
+        if len(args) == 1:
+            # single arg
+            shape, = args
+
+            if sentry_shape_scalar(shape):
+                ndim = 1
+            else:
+                shape = normalize_shape(shape)
+                if shape is None:
+                    return
+                ndim = shape.count
+            retty = ary.copy(ndim=ndim)
+            return signature(retty, shape)
+
+        elif len(args) == 0:
+            # no arg
+            raise TypeError("resize() take at least one arg")
+
+        else:
+            # vararg case
+            if any(not sentry_shape_scalar(a) for a in args):
+                raise TypeError("resize({0}) is not supported".format(
+                    ', '.join(map(str, args))))
+
+            retty = ary.copy(ndim=len(args))
+            return signature(retty, *args)
+
     @bound_function("array.sort")
     def resolve_sort(self, ary, args, kws):
         assert not args
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -1914,6 +1914,94 @@ def np_reshape(a, shape):
     return np_reshape_impl
 
 
+def check_input_shape(shape):
+    num_neg_value = 0
+    known_size = 1
+    for _, s in enumerate(shape):
+        if s < 0:
+            num_neg_value += 1
+        else:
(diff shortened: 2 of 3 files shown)


stuartarchibald (commented):
@czgdp1807 thanks for starting work on this. Having quickly looked at the proposed changes, it seems like the public extension APIs could work here? I think `@overload_method` https://numba.readthedocs.io/en/stable/extending/high-level.html#implementing-methods might be the thing to use (probably along with some `@intrinsic` as needed). What do you think?

czgdp1807:
I feel that you are right. So that means we don't need to add `resolve_resize` function to `numba/core/typing/arraydecl.py` file I think? Plus does that also mean that we need to make changes to `reshape`'s implementation as well? Please let me know, I can do that in a separate PR. Thanks.

stuartarchibald:
> I feel that you are right. So that means we don't need to add `resolve_resize` function to `numba/core/typing/arraydecl.py` file I think?

I think that's the case. The typing part of the `@overload*` functions can do any type checking or analysis needed.

The idea behind using the extension APIs internally is to make it so that Numba's implementation of functionality is itself written using it's own APIs, it means there's less to maintain and ensures that the APIs are appropriate for real world use! Perhaps take a look at the guide to using `@overload` if you are not familiar with it: https://numba.readthedocs.io/en/stable/extending/overloading-guide.html

> Plus does that also mean that we need to make changes to `reshape`'s implementation as well?

I'd try and restrict the scope of the PR to the minimum. What's the issue you've run into with `reshape` or is it just that it uses the old-style APIs?

Thanks for your help.

czgdp1807:
> Perhaps take a look at the guide to using `@overload` if you are not familiar with it: https://numba.readthedocs.io/en/stable/extending/overloading-guide.html

Thanks for sharing. I read this and its a wonderful documentation. Reinforces my learnings from previous PRs here. 



> What's the issue you've run into with `reshape` or is it just that it uses the old-style APIs?

No conceptual issues as such. Just that it helps in maintaining consistency across the code base. No worries though if we don't want to make changes to `reshape`, its already working so we can leave as it is.

czgdp1807:
Well while try to use `overload_method` in `resize`, we can't do it. The reason is we need lower level access to `shape` and `strides` so that we can just update these two things when new size is less than or equal to the original size. `overload_method` will not allow something like, `a.shape = new_shape`. So may be that's why we need `lower_builtin` to update the arrays directly, the way `reshape` does.

stuartarchibald:
@czgdp1807 Thanks for looking at this some more. Some observations:

1. `np.ndarray.resize` (the method on the NumPy array object) updates the array in-place and returns `None`. It also has a default of checking CPython's reference counter (kwarg `refcheck=True`). This is problematic on numerous fronts:
   * Numba's type inference system cannot handle in-place type changes, e.g.
     ```python
     x = np.ones((3, 4, 5)) # types.Array(float64, 3d, 'C')
     x.resize((2,)) # types.Array(float64, 1d, 'C') but there's no assignment!
     ```
   * Numba has no access to or knowledge of CPython's reference counter, it has its own reference counting system.
   * The array type is represented as a LLVM SSA value of type struct, so the parts that need mutating (e.g. shape) cannot be mutated.
   
   Due to the above, it's unfortunately not possible to implement this functionality in Numba.
2. `np.resize` (the NumPy module API function) returns a new array with the given shape. This should be possible to implement in Numba.

czgdp1807:
> 2\. `np.resize` (the NumPy module API function) returns a new array with the given shape. This should be possible to implement in Numba.

Yeah. It would be not that difficult I think to do. I can do this in a separate PR.

Thanks for explaining `@intrinsic` in https://github.com/numba/numba/pull/8020#issuecomment-1118791450. 

I will close this PR then and will make a new one for `np.resize`.

PR #8024, opened 2022-05-04
Allow converting NumPy datetimes to int

apmasell:
Add support for `int(numpy.datetime64)` and `int(numpy.timedelta64)`.

### DIFF
--- a/numba/core/typing/builtins.py
+++ b/numba/core/typing/builtins.py
@@ -964,7 +964,7 @@ class Int(AbstractTemplate):
             if arg.unit == 'ns':
                 return signature(types.int64, arg)
             else:
-                raise errors.NumbaTypeError(f"Only datetime64[ns] can be converted, but got {arg.unit}")
+                raise errors.NumbaTypeError(f"Only datetime64[ns] can be converted, but got datetime64[{arg.unit}]")
         if isinstance(arg, types.NPTimedelta):
             return signature(types.int64, arg)
 
--- a/numba/core/typing/builtins.py
+++ b/numba/core/typing/builtins.py
@@ -796,7 +796,7 @@ class NumberClassAttribute(AttributeTemplate):
 
     def resolve___call__(self, classty):
         """
-        Resolve a number class's constructor (e.g. calling int(...))
+        Resolve a NumPy number class's constructor (e.g. calling numpy.int32(...))
         """
         ty = classty.instance_type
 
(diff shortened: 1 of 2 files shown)


stuartarchibald:
Thanks for the patch @apmasell, I think that converting `datetime64`/`timedelta64` to `int` is not permitted unless the unit is `'ns'`, whereas I think the `.view()` method should work all the time (to e.g. `.view(np.int64)`).

sklam:
`int(timedelta64)` has different rules for the units than `int(datetime64)`:

```python
In [3]: td = np.timedelta64(1, 'Y')

In [4]: td
Out[4]: numpy.timedelta64(1,'Y')

In [5]: int(td)
Out[5]: 1
```

We'll need to check what numpy is doing to be sure.

numba/core/typing/builtins.py
```diff
@@ -960,6 +960,13 @@ def generic(self, args, kws):
             return signature(arg, arg)
         if isinstance(arg, (types.Float, types.Boolean)):
             return signature(types.intp, arg)
+        if isinstance(arg, types.NPDatetime):
+            if arg.unit == 'ns':
+                return signature(types.int64, arg)
+            else:
+                raise errors.NumbaTypeError(f"Only datetime64[ns] can be converted, but got {arg.unit}")
```
sklam, line 967: nitpicking
```suggestion
                raise errors.NumbaTypeError(f"Only datetime64[ns] can be converted, but got datetime64[{arg.unit}]")
```

numba/tests/test_npdatetime.py
```diff
@@ -544,6 +547,25 @@ def check(a):
...
+
+        for time in ('2014', '2016', '2000', '2014-02', '2014-03', '2014-04',
+                     '2016-02', '2000-12-31', '2014-01-16', '2014-01-05',
+                     '2014-01-07', '2014-01-06', '2014-02-02', '2014-02-27',
+                     '2014-02-16', '2014-03-01', '2000-01-01T01:02:03.002Z',
+                     '2000-01-01T01:02:03Z'):
+            check(DT(time).astype('datetime64[ns]'))
+
+        with self.assertRaises((TypeError, TypingError)):
+            f(DT('2014'))
```
sklam, line 567: Can you add a test when `TypingError` is raised that the expected message is shown?

sklam (approved):
Sorry for the delay in review. Patch looks good.

PR #8029, opened 2022-05-09
convert NamedTuple to NamedUniTuple when possible

### LINKED ISSUE #8026: When dict value_type is namedtuple, cannot set item for the dict
```

from numba import njit
from numba.core import types
from numba.typed.typeddict import Dict

def test_misc():
        @njit
        def use_tuple_dict(d):
            return d

        Node = namedtuple("Node", ["ID", "Name"])
        node1 = Node("1", "abc")
        dct = Dict.empty(
            key_type  =types.unicode_type, 
            value_type=types.NamedTuple(
                (types.unicode_type, types.unicode_type), 
                Node)
        )
        dct["1"] = node1
        pv = use_tuple_dict.py_func(dct)
        jv = use_tuple_dict(dct)
        assert pv == jv
```
And an Error was reported:
```
  File "/Users/ld/Code/numba/numba/tests/test_unicode_array.py", line 919, in test_misc4
    dct["1"] = node1
  File "/Users/ld/Code/numba/numba/typed/typeddict.py", line 155, in __setitem__
    return _setitem(self, key, value)
  File "/Users/ld/Code/numba/numba/core/dispatcher.py", line 468, in _compile_for_args
    error_rewrite(e, 'typing')
  File "/Users/ld/Code/numba/numba/core/dispatcher.py", line 409, in error_rewrite
    raise e.with_traceback(None)
 
 >>> setitem(DictType[unicode_type,Node(unicode_type, unicode_type)]<iv=None>, unicode_type, Node(unicode_type x 2))
 
    With argument(s): '(DictType[unicode_type,Node(unicode_type, unicode_type)]<iv=None>, unicode_type, Node(unicode_type x 2))':
   No match.
  Overload in function 'impl_setitem': File: numba/typed/dictobject.py: Line 695.
    With argument(s): '(DictType[unicode_type,Node(unicode_type, unicode_type)]<iv=None>, unicode_type, Node(unicode_type x 2))':
     LoweringError: Failed in nopython mode pipeline (step: native lowering)
   
   
   File "../typed/dictobject.py", line 705:
       def impl(d, key, value):
           <source elided>
           castedkey = _cast(key, keyty)
           castedval = _cast(value, valty)
           ^
   
   During: lowering "castedval = call $12load_global.4(value, $16load_deref.6, func=$12load_global.4, args=[Var(value, dictobject.py:704), Var($16load_deref.6, dictobject.py:705)], kws=(), vararg=None, target=None)" at /Users/ld/Code/numba/numba/typed/dictobject.py (705)
  raised from /Users/ld/Code/numba/numba/core/errors.py:837

During: typing of setitem at /Users/ld/Code/numba/numba/typed/typeddict.py (33)

File "../typed/typeddict.py", line 33:
def _setitem(d, key, value):
    d[key] = value
    ^

```
cc @luk-f-a

dlee992:
fixes #8026 .

- [x] Add \_\_new\_\_ for `NamedTuple`, similar with \_\_new\_\_ for `Tuple`.
- [x] Add a test for #8026 .
- [x] Change `cls` to `clss`, to avoid duplicate names between positional args and kwargs.
- [x] forbid `LiteralDict.FakeNamedTuple` use NamedUniTuple
- [x] CI pass

### DIFF
--- a/numba/core/types/containers.py
+++ b/numba/core/types/containers.py
@@ -393,7 +393,12 @@ class NamedUniTuple(_HomogeneousTuple, BaseNamedTuple):
 
 
 class NamedTuple(_HeterogeneousTuple, BaseNamedTuple):
-    def __new__(cls, types, clss):
+    def __new__(cls, types, clss, convert=True):
+        """
+        arg convert: whether allow NamedTuple convert to Uni, default True.
+        """
+        if not convert:
+            return object.__new__(NamedTuple)
 
         t = utils.unified_function_type(types, require_precise=True)
         if t is not None:
@@ -406,8 +411,7 @@ class NamedTuple(_HeterogeneousTuple, BaseNamedTuple):
         else:
             return object.__new__(NamedTuple)
 
-    def __init__(self, types, clss):
-        _HeterogeneousTuple.is_types_iterable(types)
+    def __init__(self, types, clss, convert=True):
         self.types = tuple(types)
         self.count = len(self.types)
         self.fields = tuple(clss._fields)
--- a/numba/core/types/containers.py
+++ b/numba/core/types/containers.py
@@ -400,7 +400,9 @@ class NamedTuple(_HeterogeneousTuple, BaseNamedTuple):
 
         _HeterogeneousTuple.is_types_iterable(types)
 
-        if types and all(t == types[0] for t in types[1:]):
+        all_literals = all(isinstance(ty, Literal) for ty in types) \
+            if types else False
+        if types and not all_literals and all(t == types[0] for t in types[1:]):
             return NamedUniTuple(dtype=types[0], count=len(types), clss=clss)
         else:
             return object.__new__(NamedTuple)
--- a/numba/core/types/containers.py
+++ b/numba/core/types/containers.py
@@ -375,12 +375,12 @@ class BaseNamedTuple(BaseTuple):
 
 
 class NamedUniTuple(_HomogeneousTuple, BaseNamedTuple):
-    def __init__(self, dtype, count, cls):
+    def __init__(self, dtype, count, clss):
         self.dtype = dtype
         self.count = count
-        self.fields = tuple(cls._fields)
-        self.instance_class = cls
-        name = "%s(%s x %d)" % (cls.__name__, dtype, count)
+        self.fields = tuple(clss._fields)
+        self.instance_class = clss
+        name = "%s(%s x %d)" % (clss.__name__, dtype, count)
         super(NamedUniTuple, self).__init__(name)
 
     @property
@@ -393,14 +393,27 @@ class NamedUniTuple(_HomogeneousTuple, BaseNamedTuple):
 
 
 class NamedTuple(_HeterogeneousTuple, BaseNamedTuple):
-    def __init__(self, types, cls):
+    def __new__(cls, types, clss):
+
+        t = utils.unified_function_type(types, require_precise=True)
+        if t is not None:
+            return NamedUniTuple(dtype=t, count=len(types), clss=clss)
+
         _HeterogeneousTuple.is_types_iterable(types)
 
+        if types and all(t == types[0] for t in types[1:]):
+            return NamedUniTuple(dtype=types[0], count=len(types), clss=clss)
+        else:
+            return object.__new__(NamedTuple)
+
+    def __init__(self, types, clss):
+        _HeterogeneousTuple.is_types_iterable(types)
         self.types = tuple(types)
         self.count = len(self.types)
-        self.fields = tuple(cls._fields)
-        self.instance_class = cls
-        name = "%s(%s)" % (cls.__name__, ", ".join(str(i) for i in self.types))
+        self.fields = tuple(clss._fields)
+        self.instance_class = clss
+
+        name = "%s(%s)" % (clss.__name__, ", ".join(str(i) for i in self.types))
         super(NamedTuple, self).__init__(name)
 
     @property
--- a/numba/core/types/containers.py
+++ b/numba/core/types/containers.py
@@ -400,9 +400,8 @@ class NamedTuple(_HeterogeneousTuple, BaseNamedTuple):
 
         _HeterogeneousTuple.is_types_iterable(types)
 
-        all_literals = all(isinstance(ty, Literal) for ty in types) \
-            if types else False
-        if types and not all_literals and all(t == types[0] for t in types[1:]):
+        if types and all(t == types[0] for t in types[1:])\
+                and not isinstance(clss, LiteralStrKeyDict.FakeNamedTuple):
             return NamedUniTuple(dtype=types[0], count=len(types), clss=clss)
         else:
             return object.__new__(NamedTuple)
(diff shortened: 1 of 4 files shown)


dlee992:
Ask a question about failed test cases, CI reported that:`TypeError: expecting {{i64, i8*}} but got [1 x {i64, i8*}]`
in the test case `File "/home/vsts/work/1/s/numba/tests/test_dictobject.py", line 2283, in test_build_map_op_code`. 

I found the reason may be the variable `nbty` type change, in function `def build_map(context, builder, dict_type, item_types, items):`. After my PR, `nbty` type changed from `NamedTuple` to `NamedUniTuple`. I think we need to change the ctor in `typed.Dict` implementation. But I haven't found the real place to change.

dlee992:
Now, I do sth tricky to workaround the `TypeError` in del instruction lowering process. I add another argument in NamedTuple new and init method, named `convert`. I hope the variable `nbty` in the function `build_map`  is still an instance of `NamedTuple` type, instead of `NamedUniTuple`. Otherwise, when checking type consistency at the `numba/core/datamodel/models.py:710`, numba will raise an inconsistent error, between `LiteralStrucutType` and `ArrayType` defined in `ir` module.

numba/core/types/containers.py
```diff
@@ -393,14 +393,31 @@ def key(self):
...
 class NamedTuple(_HeterogeneousTuple, BaseNamedTuple):
-    def __init__(self, types, cls):
+    def __new__(cls, types, clss, convert=True):
+        """
+        arg convert: whether allow NamedTuple convert to Uni, default True.
+        """
+        if not convert:
+            return object.__new__(NamedTuple)
+
+        t = utils.unified_function_type(types, require_precise=True)
```
luk-f-a, line 403: If I'm not mistaken this line would allow the casting of one type into another, prioritizing the creation of NamedUniTuple over maintaning the original types. I might be wrong but I think that if the input is (int64, float64), then the types will be unified as (float64, float64) and then a NamedUniTuple will be created. Shouldn't we prefer to preserve the original types by using a NamedTuple in this case? And only create NamedUniTuple when the types are the same.

stuartarchibald: @dlee992 following on from the discussion at the Numba meeting, I'd guess that the types are being constructed differently for some reason, this is done here for maps: https://github.com/numba/numba/blob/4684b9cf9d5e42ec0e144c3f238432b078627a1c/numba/core/typeinfer.py#L317-L326 and here for lists: https://github.com/numba/numba/blob/4684b9cf9d5e42ec0e144c3f238432b078627a1c/numba/core/typeinfer.py#L278-L285 it's the `__call__` methods that do the construction logic. That might be the place to start debugging.

dlee992: @stuartarchibald , thanks! I will debug from here ASAP.

luk-f-a: I was wrong in my comment above. That line will only apply to function types, and it will not affect other types like float or int.

dlee992: @luk-f-a , thanks for ur new finding. @stuartarchibald, I tried with ur suggestions. However, these two test cases don't call into the `call` method of `BuildListConstraint`. I now tried to debug about `InsertValue` and `TupleModel` stuff, since they are in the code location when error happens.

dlee992: I think I have found the root cause. Since I change `NamedTuple` could be converted to `NamedUniTuple` by `new` method, but `LiteralStrKeyDict` is registered to use `TupleModel`. Actually, it should can also use `UniTupleModel`, since sometimes `LiteralStrKeyDict` will use `NamedUniTuple` as underlying datastruct.

dlee992:
Hi, @stuartarchibald, this PR is ready for review again. I think right now the trick is good, and not use new argument.

numba/core/types/containers.py
```diff
@@ -393,14 +393,26 @@ def key(self):
...
-    def __init__(self, types, cls):
+    def __new__(cls, types, clss):
+        t = utils.unified_function_type(types, require_precise=True)
+        if t is not None:
+            return NamedUniTuple(dtype=t, count=len(types), clss=clss)
+
         _HeterogeneousTuple.is_types_iterable(types)
 
+        if types and all(t == types[0] for t in types[1:])\
+                and not isinstance(clss, LiteralStrKeyDict.FakeNamedTuple):
```
dlee992, line 404: I forbid `LiteralStrKeyDict` to use `NamedUniType` here.

dlee992:
@stuartarchibald , hi stuart. Could u begin to review this, if u have some time? Many thanks.

stuartarchibald:
I think the reason this patch is running into difficulties is that the new `NamedTuple` type construction behaviour leads to the wrong representation when lowering the `build_map` expression in the case of a `LiteralStrKeyDict` with a single key. e.g. 
```python
@njit
def foo():
    a = {'a': {'b1': 1.23, 'b2': 'string'}}

foo()
```
This patch (do not use it as it's just a workaround!):
```diff
diff --git a/numba/typed/dictobject.py b/numba/typed/dictobject.py
--- a/numba/typed/dictobject.py
+++ b/numba/typed/dictobject.py
@@ -1099,8 +1099,13 @@ def build_map(context, builder, dict_type, item_types, items):
     if isinstance(dict_type, types.LiteralStrKeyDict):
         unliteral_tys = [x for x in
                          dict_type.literal_value.values()]
+        if len(dict_type.tuple_ty._fields) == 1:
+            unliteral_tys += ("dummy",)
         nbty = types.NamedTuple(unliteral_tys,
                                 dict_type.tuple_ty)
+        if len(dict_type.tuple_ty._fields) == 1:
+            nbty.types = nbty.types[:-1]
+            nbty.count -= 1
         values = [x[1] for x in items]
         # replace with make_tuple call?
         tup = context.get_constant_undef(nbty)
```
fixes it, because it's creating a new "dummy" field to trigger creation of a `NamedTuple` through the constructor opposed to a `NamedUniTuple`.

dlee992:
@stuartarchibald, thank u for catching up. You nailed the key point. I feel my current PR lets NamedTuple \_\_new\_\_ method know if it's called from build_map for `LiteralStrKeyDict`, then don't convert to NamedUniTuple. And ur suggested fix (or workaround) lets NamedTuple \_\_new\_\_  not be aware about this info, instead let build_map to do the same trick. I feel these two ways are only different implementations for the same goal. If u don't like my way, I would like to change to ur way to fix this.

The root cause is here as I said in the comments above:
> I think I have found the root cause. Since I change `NamedTuple` could be converted to `NamedUniTuple`, but `LiteralStrKeyDict` is registered to use `TupleModel`. Actually, it should also use `UniTupleModel`, since sometimes `LiteralStrKeyDict` will use `NamedUniTuple` as underlying datastruct.

dlee992:
I think this issue again, I feel we don't have a nice way to handle `homo` and `hete` literal string key dictionary using just one type at the same time. 

So, I propose we might create two types for this issue, say, a `HomoLiteralStrKeyDict` for `homo` values (binding to `UniTupleModel`, and a `HeteLiteralStrKeyDict` for `hete` values (binding to `TupleModel`). 

What do u think? @stuartarchibald

dlee992:
I decide to give up this PR, since this issue is not bother many users. If anyone is interested in fixing this issue, can go on with it.

PR #8031, opened 2022-05-10
Support for Numpy BitGenerators PR#1 - Core Generator Support

kc611:
This PR adds the core `BitGenerator` implementation from #7900 and adds support for `np.random.Generator().random()`

cc @esc @stuartarchibald

### DIFF
--- a/docs/source/reference/numpysupported.rst
+++ b/docs/source/reference/numpysupported.rst
@@ -585,7 +585,7 @@ Modules
 
 Generator Objects
 '''''''''''''''''
-Numba supports ``numpy.random.Generator()`` objects. Users can now pass
+Numba supports ``numpy.random.Generator()`` objects. As of version 0.56, users can pass
 individual Numpy ``Generator`` objects into Numba functions and use their
 methods inside the functions. The same algorithms are used as Numpy for
 random number generation hence maintaining parity between the random
@@ -598,8 +598,8 @@ execution logic.
 .. note::
   Numpy's ``Generator`` objects rely on ``BitGenerator`` to manage state
   and generate the random bits, which are then transformed into random
-  values from useful distributions. Numba ``unbox``es the ``Generator`` objects
-  and maintians a reference to the underlying ``BitGenerator`` objects using Numpy's
+  values from useful distributions. Numba will ``unbox`` the ``Generator`` objects
+  and will maintain a reference to the underlying ``BitGenerator`` objects using Numpy's
   ``ctypes`` interface bindings. Hence ``Generator`` objects can cross the JIT boundary
   and their functions be used within Numba-Jit code. Note that since only references
   to ``BitGenerator`` objects are maintained, any change to the state of a particular
--- a/docs/source/reference/numpysupported.rst
+++ b/docs/source/reference/numpysupported.rst
@@ -586,20 +586,20 @@ Modules
 Generator Objects
 '''''''''''''''''
 Numba supports :py:class:`numpy.random.Generator()` objects. As of version 0.56, users can pass
-individual Numpy :py:class:`Generator` objects into Numba functions and use their
-methods inside the functions. The same algorithms are used as Numpy for
+individual NumPy :py:class:`Generator` objects into Numba functions and use their
+methods inside the functions. The same algorithms are used as NumPy for
 random number generation hence maintaining parity between the random
-number generated using Numpy and Numba under identical arguments 
-(also the same documentation notes as Numpy :py:class:`Generator` methods apply).
+number generated using NumPy and Numba under identical arguments 
+(also the same documentation notes as NumPy :py:class:`Generator` methods apply).
 The current Numba support for :py:class:`Generator` is not thread-safe, hence we
 do not recommend using :py:class:`Generator` methods in methods with parallel
 execution logic.
 
 .. note::
-  Numpy's :py:class:`Generator` objects rely on :py:class:`BitGenerator` to manage state
+  NumPy's :py:class:`Generator` objects rely on :py:class:`BitGenerator` to manage state
   and generate the random bits, which are then transformed into random
   values from useful distributions. Numba will ``unbox`` the :py:class:`Generator` objects
-  and will maintain a reference to the underlying :py:class:`BitGenerator` objects using Numpy's
+  and will maintain a reference to the underlying :py:class:`BitGenerator` objects using NumPy's
   ``ctypes`` interface bindings. Hence :py:class:`Generator` objects can cross the JIT boundary
   and their functions be used within Numba-Jit code. Note that since only references
   to :py:class:`BitGenerator` objects are maintained, any change to the state of a particular
--- a/docs/source/reference/numpysupported.rst
+++ b/docs/source/reference/numpysupported.rst
@@ -585,39 +585,32 @@ Modules
 
 Generator Objects
 '''''''''''''''''
-Numba supports ``numpy.random.Generator()`` objects and their respective
-methods. Numpy's ``Generator`` objects rely on ``BitGenerator`` to manage state
-and generate the random bits, which are then transformed into random
-values from useful distributions. Numba ``unbox``es the ``Generator`` objects
-and maintians a reference to the underlying ``BitGenerator`` objects using Numpy's
-``ctypes`` interface bindings. Hence ``Generator`` objects can cross the JIT boundary
-and their functions be used within Numba-Jit code. Note that since only references
-to ``BitGenerator`` objects are maintained, any change to the state of a particular
-``Generator`` object outside Numba code would affect the state of ``Generator``
-inside the Numba code.
+Numba supports ``numpy.random.Generator()`` objects. Users can now pass
+individual Numpy ``Generator`` objects into Numba functions and use their
+methods inside the functions. The same algorithms are used as Numpy for
+random number generation hence maintaining parity between the random
+number generated using Numpy and Numba under identical arguments 
+(also the same documentation notes as Numpy ``Generator`` methods apply).
+The current Numba support for ``Generator`` is not thread-safe, hence we
+do not recommend using ``Generator`` methods in methods with parallel
+execution logic.
 
-.. code-block:: python
+.. note::
+  Numpy's ``Generator`` objects rely on ``BitGenerator`` to manage state
+  and generate the random bits, which are then transformed into random
+  values from useful distributions. Numba ``unbox``es the ``Generator`` objects
+  and maintians a reference to the underlying ``BitGenerator`` objects using Numpy's
+  ``ctypes`` interface bindings. Hence ``Generator`` objects can cross the JIT boundary
+  and their functions be used within Numba-Jit code. Note that since only references
+  to ``BitGenerator`` objects are maintained, any change to the state of a particular
+  ``Generator`` object outside Numba code would affect the state of ``Generator``
+  inside the Numba code.
 
-  import numpy as np
-  import numba
-
-  x = np.random.default_rng(1)
-  y = np.random.default_rng(1)
-
-  size = 10
-
-  def do_stuff(gen):
-      return gen.random(size=int(size/2))
-
-  print(x.random(size=size))
-  # [0.51182162 0.9504637  0.14415961 0.94864945 0.31183145
-  #  0.42332645 0.82770259 0.40919914 0.54959369 0.02755911]
-
-  print(do_stuff(y))
-  # [0.51182162 0.9504637  0.14415961 0.94864945 0.31183145]
-
-  print(y.random(size=int(size/2)))
-  # [0.42332645 0.82770259 0.40919914 0.54959369 0.02755911]
+.. literalinclude:: ../../../numba/tests/doc_examples/test_numpy_generators.py
+   :language: python
+   :start-after: magictoken.npgen_usage.begin
+   :end-before: magictoken.npgen_usage.end
+   :dedent: 8
 
 The following ``Generator`` methods are supported:
 
--- a/docs/source/reference/numpysupported.rst
+++ b/docs/source/reference/numpysupported.rst
@@ -585,15 +585,15 @@ Modules
 
 Generator Objects
 '''''''''''''''''
-Numba supports `numpy.random.Generator()` objects and their respective
-methods. Numpy's `Generator` objects rely on `BitGenerator` to manage state
+Numba supports ``numpy.random.Generator()`` objects and their respective
+methods. Numpy's ``Generator`` objects rely on ``BitGenerator`` to manage state
 and generate the random bits, which are then transformed into random
-values from useful distributions. Numba `unbox`es the `Generator` objects
-and maintians a reference to the underlying `BitGenerator` objects using Numpy's
-`ctypes` interface bindings. Hence `Generator` objects can cross the JIT boundary
+values from useful distributions. Numba ``unbox``es the ``Generator`` objects
+and maintians a reference to the underlying ``BitGenerator`` objects using Numpy's
+``ctypes`` interface bindings. Hence ``Generator`` objects can cross the JIT boundary
 and their functions be used within Numba-Jit code. Note that since only references
-to `BitGenerator` objects are maintained, any change to the state of a particular
-`Generator` object outside Numba code would affect the state of `Generator`
+to ``BitGenerator`` objects are maintained, any change to the state of a particular
+``Generator`` object outside Numba code would affect the state of ``Generator``
 inside the Numba code.
 
 .. code-block:: python
@@ -619,7 +619,7 @@ inside the Numba code.
   print(y.random(size=int(size/2)))
   # [0.42332645 0.82770259 0.40919914 0.54959369 0.02755911]
 
-The following `Generator` methods are supported:
+The following ``Generator`` methods are supported:
 
 * :func:`numpy.random.Generator().random()`
 
--- a/docs/source/reference/numpysupported.rst
+++ b/docs/source/reference/numpysupported.rst
@@ -583,6 +583,46 @@ Modules
 ``random``
 ----------
 
+Generator Objects
+''''''''''''
+Numba supports `numpy.random.Generator()` objects and their respective
+methods. Numpy's `Generator` objects rely on `BitGenerator` to manage state
+and generate the random bits, which are then transformed into random
+values from useful distributions. Numba `unbox`es the `Generator` objects
+and maintians a reference to the underlying `BitGenerator` objects using Numpy's
+`ctypes` interface bindings. Hence `Generator` objects can cross the JIT boundary
+and their functions be used within Numba-Jit code. Note that since only references
+to `BitGenerator` objects are maintained, any change to the state of a particular
+`Generator` object outside Numba code would affect the state of `Generator`
+inside the Numba code.
+
+  import numpy as np
+  import numba
+
+  x = np.random.default_rng(1)
+  y = np.random.default_rng(1)
+
+  size = 10
+
+  def do_stuff(gen):
+      return gen.random(size=int(size/2))
+
+  print(x.random(size=size))
+  # [0.51182162 0.9504637  0.14415961 0.94864945 0.31183145
+  #  0.42332645 0.82770259 0.40919914 0.54959369 0.02755911]
+
+  print(do_stuff(y))
+  # [0.51182162 0.9504637  0.14415961 0.94864945 0.31183145]
+
+  print(y.random(size=int(size/2)))
+  # [0.42332645 0.82770259 0.40919914 0.54959369 0.02755911]
+
+The following `Generator` methods are supported:
+
+* :func:`numpy.random.Generator().random()`
+
+RandomState and legacy Random number generation
+''''''''''''
 Numba supports top-level functions from the
 `numpy.random <http://docs.scipy.org/doc/numpy/reference/routines.random.html>`_
 module, but does not allow you to create individual RandomState instances.
--- a/docs/source/reference/numpysupported.rst
+++ b/docs/source/reference/numpysupported.rst
@@ -585,25 +585,25 @@ Modules
 
 Generator Objects
 '''''''''''''''''
-Numba supports ``numpy.random.Generator()`` objects. As of version 0.56, users can pass
-individual Numpy ``Generator`` objects into Numba functions and use their
+Numba supports :py:class:`numpy.random.Generator()` objects. As of version 0.56, users can pass
+individual Numpy :py:class:`Generator` objects into Numba functions and use their
 methods inside the functions. The same algorithms are used as Numpy for
 random number generation hence maintaining parity between the random
 number generated using Numpy and Numba under identical arguments 
-(also the same documentation notes as Numpy ``Generator`` methods apply).
-The current Numba support for ``Generator`` is not thread-safe, hence we
-do not recommend using ``Generator`` methods in methods with parallel
+(also the same documentation notes as Numpy :py:class:`Generator` methods apply).
+The current Numba support for :py:class:`Generator` is not thread-safe, hence we
+do not recommend using :py:class:`Generator` methods in methods with parallel
 execution logic.
 
 .. note::
-  Numpy's ``Generator`` objects rely on ``BitGenerator`` to manage state
+  Numpy's :py:class:`Generator` objects rely on :py:class:`BitGenerator` to manage state
   and generate the random bits, which are then transformed into random
-  values from useful distributions. Numba will ``unbox`` the ``Generator`` objects
-  and will maintain a reference to the underlying ``BitGenerator`` objects using Numpy's
-  ``ctypes`` interface bindings. Hence ``Generator`` objects can cross the JIT boundary
+  values from useful distributions. Numba will ``unbox`` the :py:class:`Generator` objects
+  and will maintain a reference to the underlying :py:class:`BitGenerator` objects using Numpy's
+  ``ctypes`` interface bindings. Hence :py:class:`Generator` objects can cross the JIT boundary
   and their functions be used within Numba-Jit code. Note that since only references
-  to ``BitGenerator`` objects are maintained, any change to the state of a particular
-  ``Generator`` object outside Numba code would affect the state of ``Generator``
+  to :py:class:`BitGenerator` objects are maintained, any change to the state of a particular
+  :py:class:`Generator` object outside Numba code would affect the state of :py:class:`Generator`
   inside the Numba code.
 
 .. literalinclude:: ../../../numba/tests/doc_examples/test_numpy_generators.py
@@ -612,7 +612,7 @@ execution logic.
    :end-before: magictoken.npgen_usage.end
    :dedent: 8
 
-The following ``Generator`` methods are supported:
+The following :py:class:`Generator` methods are supported:
 
 * :func:`numpy.random.Generator().random()`
 
--- a/docs/source/reference/numpysupported.rst
+++ b/docs/source/reference/numpysupported.rst
@@ -584,7 +584,7 @@ Modules
 ----------
 
 Generator Objects
-''''''''''''
+'''''''''''''''''
 Numba supports `numpy.random.Generator()` objects and their respective
 methods. Numpy's `Generator` objects rely on `BitGenerator` to manage state
 and generate the random bits, which are then transformed into random
@@ -595,6 +595,8 @@ and their functions be used within Numba-Jit code. Note that since only referenc
 to `BitGenerator` objects are maintained, any change to the state of a particular
 `Generator` object outside Numba code would affect the state of `Generator`
 inside the Numba code.
+
+.. code-block:: python
 
   import numpy as np
   import numba
@@ -622,7 +624,8 @@ The following `Generator` methods are supported:
 * :func:`numpy.random.Generator().random()`
 
 RandomState and legacy Random number generation
-''''''''''''
+'''''''''''''''''''''''''''''''''''''''''''''''
+
 Numba supports top-level functions from the
 `numpy.random <http://docs.scipy.org/doc/numpy/reference/routines.random.html>`_
 module, but does not allow you to create individual RandomState instances.
--- a/numba/core/boxing.py
+++ b/numba/core/boxing.py
@@ -1177,19 +1177,6 @@ def unbox_numpy_random_bitgenerator(typ, obj, c):
     wire_in_fnptrs('next_uint64')
     wire_in_fnptrs('next_uint32')
 
-    # This is the same as the `state` member but it's the bit_generator address,
-    # it's probably never needed.
-    interface_bit_generator = c.pyapi.object_getattr_string(
-        ctypes_binding, 'bit_generator')
-    interface_bit_generator_value = c.pyapi.object_getattr_string(
-        interface_bit_generator, 'value')
-    setattr(
-        struct_ptr,
-        'bit_generator',
-        c.unbox(
-            types.uintp,
-            interface_bit_generator_value).value)
-
     return NativeValue(struct_ptr._getvalue())
 
 _bit_gen_type = types.NumPyRandomBitGeneratorType('bit_generator')
@@ -1200,12 +1187,15 @@ def unbox_numpy_random_generator(typ, obj, c):
     bit_gen_inst = c.pyapi.object_getattr_string(obj, 'bit_generator')
     unboxed = c.unbox(_bit_gen_type, bit_gen_inst).value
     struct_ptr.bit_generator = unboxed
-    blank_ptr = cgutils.alloca_once(c.builder, ir.IntType(8))
-    struct_ptr.meminfo = c.pyapi.nrt_meminfo_new_from_pyobject(
-        blank_ptr, obj
-    )
     struct_ptr.parent = obj
-    return NativeValue(struct_ptr._getvalue())
+    NULL = cgutils.voidptr_t(None)
+    struct_ptr.meminfo = c.pyapi.nrt_meminfo_new_from_pyobject(
+        NULL,  # there's no data
+        obj,   # the python object, the call to nrt_meminfo_new_from_pyobject
+               # will py_incref
+    )
+    is_py_error = cgutils.is_not_null(c.builder, c.pyapi.err_occurred())
+    return NativeValue(struct_ptr._getvalue(), is_error=is_py_error)
 
 
 @box(types.NumPyRandomGeneratorType)
--- a/numba/core/boxing.py
+++ b/numba/core/boxing.py
@@ -1260,7 +1260,6 @@ def unbox_numpy_random_generator(typ, obj, c):
         where deletion happens in Python runtime without NRT being awareness of it. 
     """
     is_error_ptr = cgutils.alloca_once_value(c.builder, cgutils.false_bit)
-    fail_obj = c.context.get_constant_null(typ)
 
     with ExitStack() as stack:
         struct_ptr = cgutils.create_struct_proxy(typ)(c.context, c.builder)
--- a/numba/core/boxing.py
+++ b/numba/core/boxing.py
@@ -1199,7 +1199,7 @@ def unbox_numpy_random_generator(typ, obj, c):
 
 
 @box(types.NumPyRandomGeneratorType)
-def box_numpy_random_bitgenerator(typ, val, c):
+def box_numpy_random_generator(typ, val, c):
     inst = c.context.make_helper(c.builder, typ, val)
     obj = inst.parent
     res = cgutils.alloca_once_value(c.builder, obj)
--- a/numba/core/boxing.py
+++ b/numba/core/boxing.py
@@ -1216,6 +1216,8 @@ def unbox_numpy_random_bitgenerator(typ, obj, c):
             # args tuple for that.
             extra_refs.append(ct_voidptr_ty)
             args = c.pyapi.tuple_pack([interface_next_fn, ct_voidptr_ty])
+            with early_exit_if_null(c.builder, stack, args):
+                handle_failure()
             extra_refs.append(ct_voidptr_ty)
 
             # Call ctypes.cast()
--- a/numba/core/boxing.py
+++ b/numba/core/boxing.py
@@ -1200,6 +1200,10 @@ def unbox_numpy_random_generator(typ, obj, c):
     bit_gen_inst = c.pyapi.object_getattr_string(obj, 'bit_generator')
     unboxed = c.unbox(_bit_gen_type, bit_gen_inst).value
     struct_ptr.bit_generator = unboxed
+    blank_ptr = cgutils.alloca_once(c.builder, ir.IntType(8))
+    struct_ptr.meminfo = c.pyapi.nrt_meminfo_new_from_pyobject(
+        blank_ptr, obj
+    )
     struct_ptr.parent = obj
     return NativeValue(struct_ptr._getvalue())
 
--- a/numba/core/boxing.py
+++ b/numba/core/boxing.py
@@ -10,6 +10,8 @@ from numba.core.errors import NumbaNotImplementedError
 
 from numba.cpython import setobj, listobj
 from numba.np import numpy_support
+from contextlib import contextmanager, ExitStack
+
 
 #
 # Scalar types
@@ -1106,7 +1108,18 @@ def unbox_typeref(typ, val, c):
 def box_LiteralStrKeyDict(typ, val, c):
     return box_unsupported(typ, val, c)
 
-import sys
+
+@contextmanager
+def early_exit_if(builder, stack: ExitStack, cond):
+    then, otherwise = stack.enter_context(builder.if_else(cond, likely=False))
+    with then:
+        yield
+    stack.enter_context(otherwise)
+
+
+def early_exit_if_null(builder, stack, obj):
+    return early_exit_if(builder, stack, cgutils.is_null(builder, obj))
+
 
 # Original implementation at: https://github.com/numba/numba/issues/4499#issuecomment-1063138477
 @unbox(types.NumPyRandomBitGeneratorType)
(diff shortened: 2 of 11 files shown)


esc:
> I just checked the CI logs, it doesn't look like the tests I added are actually being run in the pipeline (probably because of them being in a new file). Where are the configuration files located for registering that ?

You will need to make the test class inherit from `TestCase` so as to mark it as a test and such that it is picked up by the testing framework. I created a temporary PR at:

https://github.com/numba/numba/pull/8033

Which does that and also deliberately makes the test fail to provide an easy way to tell if it did indeed fail. You can probably just grab the following diff to make it work:

```
diff --git a/numba/tests/test_np_randomgen.py b/numba/tests/test_np_randomgen.py
index f0830b10c2..9d33d62914 100644
--- a/numba/tests/test_np_randomgen.py
+++ b/numba/tests/test_np_randomgen.py
@@ -1,10 +1,12 @@
 import numpy as np
 import numba

 from numpy.random import MT19937, Generator


-class TestRandomGenerators:
+class TestRandomGenerators(TestCase):
+
     def check_numpy_parity(self, distribution_func,
                            bitgen_instance=None, seed=1):
         distribution_func = numba.njit(distribution_func)
```

kc611:
What kind of documentation should we add for this PR ?

numba/np/random/generator_methods.py
```diff
@@ -0,0 +1,51 @@
...
+from numba.core.extending import overload_method
+from numba.core.imputils import Registry
+from numba.np.numpy_support import as_dtype, from_dtype
+from numba.np.random.generator_core import next_float, next_double
+
+
+registry = Registry('generator_methods')
+
+
+def get_proper_func(func_32, func_64, dtype):
```
esc, line 12: This function will need some comments, to explain what is going on. Also, it may be wise to unit-test this, the function does seem to be quite straightforward to test.

numba/tests/test_np_randomgen.py
```diff
@@ -0,0 +1,38 @@
...
+
+            assert np.allclose(numba_res, numpy_res)
+
+        # Check if the end state of both BitGenerators is same
+        # after drawing the distributions
+        numba_gen_state = numba_rng_instance.__getstate__()['state']
+        numpy_gen_state = numpy_rng_instance.__getstate__()['state']
+
+        for _state_key in numpy_gen_state:
+            assert np.all(numba_gen_state[_state_key]
```
esc, line 32: As above, recommend to use `numpy.testing`.

numba/tests/test_np_randomgen.py
```diff
@@ -0,0 +1,38 @@
...
+                          == numpy_gen_state[_state_key])
+
+    def test_random(self):
+        dist_func = lambda x, size:x.random(size=size)
```
esc, line 36: I see that the `random` overload above also has a `dtype` keyword argument, so it will be good to test that too.

Also I would recommend to try to run just `x.random` without any keyword arguments, we sometimes have issues with keyword and non-keyword argument handling.

esc (commented):
Looking good. I left some comments for your consideration.

esc:
> What kind of documentation should we add for this PR ?

I would argue that the section in: https://numba.readthedocs.io/en/stable/reference/numpysupported.html#modules

will need an overhaul to describe how the NumPy random number generator state can now traverse the JIT boundary.

numba/core/typing/typeof.py
```diff
@@ -11,6 +11,11 @@
 # terminal color markup
 _termcolor = errors.termcolor()
 
+try:
+    from numpy.random._bit_generator import BitGenerator
```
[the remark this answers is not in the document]
kc611: Ah this is primarily because of: https://github.com/numba/numba/pull/7900#issuecomment-1065026619
Will add a comment here pointing to the same.

numba/tests/test_np_randomgen.py
```diff
@@ -0,0 +1,38 @@
...
+        else:
+            numba_rng_instance = Generator(bitgen_instance(seed))
+            numpy_rng_instance = Generator(bitgen_instance(seed))
+
+        # Check parity for different size cases
+        for size in [None, (), (100,), (10,20,30)]:
+            numba_res = distribution_func(numba_rng_instance, size)
+            numpy_res = distribution_func.py_func(numpy_rng_instance, size)
+
+            assert np.allclose(numba_res, numpy_res)
```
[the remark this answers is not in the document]
kc611: Yes. Theoretically, they should be equal down to the last bit, unless it's some voodoo precision differences are introduced due to change in frameworks/optimizations.

esc:
@kc611 thank you for the updates. One immediate thing to note is that the doc build seems to be failing. This is because for the public CI in Numba we have elected to treat warnings as errors and so the build fails. The error is

```
/home/vsts/work/1/s/docs/source/reference/numpysupported.rst:586:Title underline too short.
```

which ought to be quite straightforward to fix.

docs/source/reference/numpysupported.rst
```diff
@@ -582,6 +582,46 @@ Modules
...
+methods. Numpy's `Generator` objects rely on `BitGenerator` to manage state
+and generate the random bits, which are then transformed into random
+values from useful distributions. Numba `unbox`es the `Generator` objects
+and maintians a reference to the underlying `BitGenerator` objects using Numpy's
+`ctypes` interface bindings. Hence `Generator` objects can cross the JIT boundary
+and their functions be used within Numba-Jit code. Note that since only references
+to `BitGenerator` objects are maintained, any change to the state of a particular
+`Generator` object outside Numba code would affect the state of `Generator`
+inside the Numba code.
+
```
esc, line 597: Look at the build at:

https://numba--8031.org.readthedocs.build/en/8031/reference/numpysupported.html

You will need the following:

```suggestion
.. code-block:: python

```

To render the python example correctly. Otherwise it will look like:

Screen Shot 2022-05-16 at 10 28 47

esc (changes requested):
The doc needs some changes for it to render correctly.

docs/source/reference/numpysupported.rst
```diff
@@ -582,6 +582,49 @@ Modules
 ``random``
 ----------
 
+Generator Objects
+'''''''''''''''''
+Numba supports `numpy.random.Generator()` objects and their respective
```
esc, line 587: IIRC the correct markup in our Sphinx setup for things that should be in fixed-font is to use two backticks ` `` ` instead of a single one on either side of the term. Looking at how this renders, the technical terms are in italics:

Screen Shot 2022-05-17 at 12 02 09

numba/core/typing/typeof.py
```diff
@@ -11,6 +11,12 @@
 # terminal color markup
 _termcolor = errors.termcolor()
 
+# https://github.com/numba/numba/pull/7900#issuecomment-1065026619
```
esc, line 14: I think it may make sense (and is safer in the long run) to include a verbatim statement of why this change is needed. We never know when Github will become obsolete and this comment will end up pointing no-where.

numba/np/random/generator_methods.py
```diff
@@ -0,0 +1,58 @@
...
+from numba.core.extending import overload_method
+from numba.core.imputils import Registry
+from numba.np.numpy_support import as_dtype, from_dtype
+from numba.np.random.generator_core import next_float, next_double
+
+
+registry = Registry('generator_methods')
+
+
+# Most of the standard Numpy distributions only support either
```
esc, line 12: This comment could go into the docstring, what do you think?

esc (changes requested):
Three minor documentation nitpicks.

numba/core/boxing.py
```diff
@@ -1105,3 +1105,99 @@ def unbox_typeref(typ, val, c):
...
+        'bit_generator',
+        c.unbox(
+            types.uintp,
+            interface_bit_generator_value).value)
+
+    return NativeValue(struct_ptr._getvalue())
+
+_bit_gen_type = types.NumPyRandomBitGeneratorType('bit_generator')
+
+@unbox(types.NumPyRandomGeneratorType)
```
stuartarchibald, line 1197: A `@box` counterpart needs writing so that this sort of thing works:
```python
@njit
def foo(gen):
    return gen

foo(np.random.default_rng(1))
```

numba/core/types/npytypes.py
```diff
@@ -587,3 +587,14 @@ def strides(self):
     @property
     def key(self):
         return self.dtype, self.shape
+
+class NumPyRandomBitGeneratorType(Type):
```
stuartarchibald, line 591:
```suggestion

class NumPyRandomBitGeneratorType(Type):
```

numba/core/typing/typeof.py
```diff
@@ -11,6 +11,12 @@
...
+try:
+    from numpy.random._bit_generator import BitGenerator
+except ImportError:
+    from numpy.random.bit_generator import BitGenerator
```
stuartarchibald, line 19: Can this perhaps be more explicit and based on NumPy version via `numba.np.numpy_support.numpy_version`? That way as older NumPy versions are dropped the split is likely to be found when grepping for a NumPy version tuple and the split path can be removed.

numba/np/random/generator_core.py
```diff
@@ -0,0 +1,113 @@
+
+from llvmlite import ir
+from numba.core import cgutils, types
+from numba.core.extending import (intrinsic, make_attribute_wrapper, models,
+                                  overload, register_jitable,
+                                  register_model)
+from numba.core.imputils import Registry
+
+registry = Registry('generator_core')
```
stuartarchibald, line 9: Can remove this?

numba/np/random/generator_core.py
```diff
@@ -0,0 +1,113 @@
...
+    def __init__(self, dmm, fe_type):
+        members = [
+            ('bit_generator', _bit_gen_type),
+        ]
+        super(
+            NumPyRandomGeneratorTypeModel,
+            self).__init__(
+            dmm,
+            fe_type,
+            members)
```
stuartarchibald, line 41: Whilst data models normally reside in `numba.core.datamodel` I think it's ok to put this here, keeping NumPy specific models with the NumPy implementation seems to make sense.

numba/np/random/generator_core.py
```diff
@@ -0,0 +1,113 @@
...
+
+# The Generator instances have a bit_generator attr
+make_attribute_wrapper(
+    types.NumPyRandomGeneratorType,
+    'bit_generator',
+    'bit_generator')
+
+
+# Generate the overloads for "next_(some type)" functions
+def generate_next_binding(overloadable_function, return_type):
```
stuartarchibald, line 52:
```suggestion
def _generate_next_binding(overloadable_function, return_type):
```
make this private?

numba/np/random/generator_core.py
```diff
@@ -0,0 +1,113 @@
...
+            next_double_addr = getattr(struct_ptr, f'fnptr_{name}')
+
+            # LLVM IR types needed
+            ll_void_ptr_t = cgctx.get_value_type(types.voidptr)
+            ll_return_t = cgctx.get_value_type(return_type)
+            ll_uintp_t = cgctx.get_value_type(types.uintp)
+
+            # Convert the stored Generator function address to a pointer
+            next_fn_fnptr = builder.inttoptr(
+                next_double_addr, ll_void_ptr_t)
```
stuartarchibald, line 73: That the function address is stored as an `int` is likely going to break caching as LLVM will bake it in as a constant.

numba/np/random/generator_core.py
```diff
@@ -0,0 +1,113 @@
...
+            # Add the function to the module
+            fnty = ir.FunctionType(ll_return_t, (ll_uintp_t,))
+            next_fn = cgutils.get_or_insert_function(
+                builder.module, fnty, name)
+            # Bit cast the function pointer to the function type
+            hack = builder.bitcast(next_fn_fnptr, next_fn.type)
```
stuartarchibald, line 79:
```suggestion
            fnptr_as_fntype = builder.bitcast(next_fn_fnptr, next_fn.type)
```
and fix up

numba/np/random/generator_core.py
```diff
@@ -0,0 +1,113 @@
...
+def next_double(bitgen):
+    return bitgen.ctypes.next_double(bitgen.ctypes.state)
+
+
+def next_uint32(bitgen):
+    return bitgen.ctypes.next_uint32(bitgen.ctypes.state)
+
+
+def next_uint64(bitgen):
+    return bitgen.ctypes.next_uint64(bitgen.ctypes.state)
```
stuartarchibald, line 103: Are these tested anywhere?

numba/np/random/generator_core.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+generate_next_binding(next_double, types.double)
+generate_next_binding(next_uint32, types.uint32)
+generate_next_binding(next_uint64, types.uint64)
+
+
+@register_jitable
+def next_float(bitgen):
+    return (next_uint32(bitgen) >> 9) * (1.0 / 8388608.0)
```
stuartarchibald, line 113: Two questions:
1. Is a `local`s dict needed to ensure the correctness of the types involved match those in the C alg?
2. I've a recollection of these constants changing in recent NumPy versions, see https://github.com/numpy/numpy/pull/20314, think a branch is needed to match?

numba/tests/test_np_randomgen.py
```diff
@@ -0,0 +1,79 @@
...
+            assert_equal(numba_gen_state[_state_key],
+                         numpy_gen_state[_state_key])
+
+    def test_random(self):
+        # Test with no arguments
+        dist_func = lambda x, size, dtype:x.random()
+        # Provide single values so this test would run exactly once
+        self.check_numpy_parity(dist_func, test_sizes=[100],
+                                test_dtypes=[np.float64])
+        self.check_numpy_parity(dist_func, bitgen_instance=MT19937)
```
stuartarchibald, line 75: Should other generator type instances be checked?

numba/tests/test_np_randomgen.py
```diff
@@ -0,0 +1,79 @@
...
+                                 types.float32)[0](), 32)
+
+    # With any other datatype it should return a TypeError
+    with assert_raises(TypeError):
+        get_proper_func(test_32bit_func, test_64bit_func, np.int32)
+
+
+class TestRandomGenerators(TestCase):
+    def check_numpy_parity(self, distribution_func,
+                           bitgen_instance=None, seed=1,
```
stuartarchibald, line 34:
```suggestion
                           bitgen_type=None, seed=1,
```
it's a type not an instance?

numba/tests/test_np_randomgen.py
```diff
@@ -0,0 +1,79 @@
...
+        distribution_func = numba.njit(distribution_func)
+        if bitgen_instance is None:
+            numba_rng_instance = np.random.default_rng(seed=seed)
+            numpy_rng_instance = np.random.default_rng(seed=seed)
+        else:
+            numba_rng_instance = Generator(bitgen_instance(seed))
+            numpy_rng_instance = Generator(bitgen_instance(seed))
+
+        if test_sizes is None:
+            test_sizes = [None, (), (100,), (10,20,30)]
```
stuartarchibald, line 46:
```suggestion
            test_sizes = [None, (), (100,), (10, 20, 30)]
```

numba/tests/test_np_randomgen.py
```diff
@@ -0,0 +1,79 @@
...
+
+        # Check if the end state of both BitGenerators is same
+        # after drawing the distributions
+        numba_gen_state = numba_rng_instance.__getstate__()['state']
+        numpy_gen_state = numpy_rng_instance.__getstate__()['state']
+
+        for _state_key in numpy_gen_state:
+            assert_equal(numba_gen_state[_state_key],
+                         numpy_gen_state[_state_key])
+
```
stuartarchibald, line 68: I think it'd be a good idea to write a number of tests with invalid types to ensure that the compiler aborts compilation appropriately.

numba/np/random/generator_methods.py
```diff
@@ -0,0 +1,58 @@
...
+        np_dt = as_dtype(nb_dt)
+
+    np_dt = np.dtype(np_dt)
+
+    if np_dt == np.float32:
+        next_func = func_32
+    elif np_dt == np.float64:
+        next_func = func_64
+    else:
+        raise TypeError(f"Unsupported dtype {np_dt} for\
```
stuartarchibald, line 35:
```suggestion
        raise numba.errors.TypingError(f"Unsupported dtype {np_dt} for\
```
this needs to be a Numba error type to get syntax highlighting etc.

numba/np/random/generator_methods.py
```diff
@@ -0,0 +1,58 @@
...
+        {dist_name} distribution")
```
stuartarchibald, line 36: The break in the line above leads to this 8 character indent rendering directly in the exception string.

numba/np/random/generator_methods.py
```diff
@@ -0,0 +1,58 @@
...
+        nb_dt = dtype
+        np_dt = as_dtype(nb_dt)
+
+    np_dt = np.dtype(np_dt)
+
+    if np_dt == np.float32:
+        next_func = func_32
+    elif np_dt == np.float64:
+        next_func = func_64
+    else:
```
stuartarchibald, line 34: I'm not sure about this, 1. is it unique to random Generator and therefore necessary? 2. If not does something like this already exist in Numba that can be reused?

numba/np/random/generator_methods.py
stuartarchibald, line 12:
```suggestion
# Most of the standard NumPy distributions only support either
```

numba/tests/test_np_randomgen.py
```diff
@@ -0,0 +1,79 @@
...
+                                 np.float64)[0](), 64)
+    assert_equal(get_proper_func(test_32bit_func, test_64bit_func,
+                                 np.float32)[0](), 32)
+    assert_equal(get_proper_func(test_32bit_func, test_64bit_func,
+                                 types.float64)[0](), 64)
+    assert_equal(get_proper_func(test_32bit_func, test_64bit_func,
+                                 types.float32)[0](), 32)
+
+    # With any other datatype it should return a TypeError
+    with assert_raises(TypeError):
```
stuartarchibald, line 28: The Numba test suite "spelling" for this pattern is like this:
https://github.com/numba/numba/blob/4684b9cf9d5e42ec0e144c3f238432b078627a1c/numba/tests/test_np_functions.py#L644-L649
the error type is checked, and then the exception message checked to ensure that it came from the expected place and that it rendered as expected.

numba/tests/test_np_randomgen.py
```diff
@@ -0,0 +1,79 @@
...
+        # Test with no arguments
+        dist_func = lambda x, size, dtype:x.random()
+        # Provide single values so this test would run exactly once
+        self.check_numpy_parity(dist_func, test_sizes=[100],
+                                test_dtypes=[np.float64])
+        self.check_numpy_parity(dist_func, bitgen_instance=MT19937)
+
+        dist_func = lambda x, size, dtype:x.random(size=size, dtype=dtype)
+        self.check_numpy_parity(dist_func)
+        self.check_numpy_parity(dist_func, bitgen_instance=MT19937)
```
stuartarchibald, line 79: Some "lower level" tests checking that calling e.g. the `next_*` methods on the unboxed instance matches the NumPy instance would be good. This would make it possible to spot e.g. that NumPy changed something about the way a distribution is computed vs. it changing something about the way `next_*` is implemented.

numba/tests/test_np_randomgen.py
```diff
@@ -0,0 +1,79 @@
...
+        return 64
+
+    assert_equal(get_proper_func(test_32bit_func, test_64bit_func,
+                                 np.float64)[0](), 64)
+    assert_equal(get_proper_func(test_32bit_func, test_64bit_func,
+                                 np.float32)[0](), 32)
+    assert_equal(get_proper_func(test_32bit_func, test_64bit_func,
+                                 types.float64)[0](), 64)
+    assert_equal(get_proper_func(test_32bit_func, test_64bit_func,
+                                 types.float32)[0](), 32)
```
stuartarchibald, line 25: Numba uses the `unittest.TestCase` method based assertions and those methods available from `numba.tests.support.TestCase` which extends `unittest.TestCase`. Suggest using `self.assertEqual(...)` here and updating similar throughout? To get an array bitwise equality assertion I think `self.assertPreciseEqual(..., ulps=0)` should work.

numba/np/random/generator_methods.py
```diff
@@ -0,0 +1,58 @@
...
+def NumPyRandomGeneratorType_random(inst, size=None, dtype=np.float64):
+    dist_func, nb_dt = get_proper_func(next_float, next_double, dtype, "random")
+    if isinstance(size, types.Omitted):
+        size = size.value
+
+    if isinstance(size, (types.NoneType,)) or size is None:
+        def impl(inst, size=None, dtype=np.float64):
+            return nb_dt(dist_func(inst.bit_generator))
+        return impl
+    else:
```
stuartarchibald, line 52: Assert that `size` is a tuple of ints?

numba/core/boxing.py
```diff
@@ -1105,3 +1105,99 @@ def unbox_typeref(typ, val, c):
 @box(types.LiteralStrKeyDict)
 def box_LiteralStrKeyDict(typ, val, c):
     return box_unsupported(typ, val, c)
+
+
```
stuartarchibald, line 1109: Should probably reference the original source of this patch? https://github.com/numba/numba/issues/4499#issuecomment-1063138477

numba/np/random/generator_methods.py
kc611: I don't think so, this one is purely an internal functionality. The distributions in Numpy itself are the ones which support either `np.float32` or `np.float64` as `dtype` arguments. For instance: https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.standard_normal.html

stuartarchibald (commented):
Thanks for working on this @kc611, great to see it implemented. Having written a some of the code for this I've done a limited review to comment on general design, docs and ensuring correctness. Outstanding issues for discussion include:
* object life times: what happens if a bit generator instance is deleted whilst in use by an `njit` function?
* caching `@njit(cache=True)`: what happens to those pointers when the code is lowered, think they end up as runtime specific ints, cache replay will likely cause segfaults!
* Is there a good way to test the bit generator sequences so as to separate their correctness from the correctness of derived properties such as the random distributions?

Thanks again!

numba/core/boxing.py
kc611: (We discussed about storing a pointer to the original `Generator` in the `NumPyRandomGeneratorType`'s model and returning it during boxing)

Can we use `cgutils.alloca_once` to get the pointer to the `Generator` ? As far as I know it allocates memory to store a copy of that object which might not be what we want, am I correct ?

numba/np/random/generator_core.py
kc611: I don't know if (or how?) we can test these w.r.t Numba. This seems to be something that is expected to be already tested in Numpy (as this is a direct functionality of it's c-interfacing)

numba/tests/test_np_randomgen.py
kc611: We can but it'll just increase the run time unnecessarily. The only reason I added an extra `MT19937` is because it was the 'legacy' generator for random bits, otherwise here we're just expected to check if the pipeline from `next_*` ctypes interface to the final output from Numba side matches the expected Numpy outputs. If there are any changes under the hood regarding how a specific `BitGenerator` is expected to give out random bits, the change is reflected in both libraries.

numba/tests/test_np_randomgen.py
kc611: If by invalid types you mean passing in a `dtype` that is incorrect, I think that logic is already being tested separately when testing the `_get_proper_func`

numba/np/random/generator_methods.py
kc611: As far as I know, this is unique atleast to certain functions within the random Generator, simply because there exists separate implementations for `float32` and `float64` (and the functions which have such separation between implementations do not allow any other datatype to be passed to it.). The other functions either do not have a datatype argument (see `Generator().normal()`) or handle each of the datatypes separately (for instance `Generator().integers()`). 

I don't know if anything like this exists within Numba. (Should be unlikely considering most Numpy functions do not have such limitations.)

numba/np/random/generator_core.py
kc611: Do you have a specific case in which this should break ?

I tried doing this and it runs as expected (not sure if this is what you meant):

```python
import numpy as np
import numba

x = np.random.default_rng(1)
y = np.random.default_rng(1)

size = 2

@numba.njit(cache=True)
def do_stuff(gen):
    return gen.random(size)

print(do_stuff(y))
print(do_stuff(y))
print(do_stuff(y))
print(do_stuff(y))
# [0.51182162 0.9504637 ]
# [0.14415961 0.94864945]
# [0.31183145 0.42332645]
# [0.82770259 0.40919914]

print(x.random(8))
# [0.51182162 0.9504637  
#  0.14415961 0.94864945 
#  0.31183145 0.42332645
#  0.82770259 0.40919914]

```

numba/core/boxing.py
```diff
@@ -1105,3 +1105,111 @@ def unbox_typeref(typ, val, c):
...
+    struct_ptr = cgutils.create_struct_proxy(typ)(c.context, c.builder)
+    bit_gen_inst = c.pyapi.object_getattr_string(obj, 'bit_generator')
+    unboxed = c.unbox(_bit_gen_type, bit_gen_inst).value
+    struct_ptr.bit_generator = unboxed
+    struct_ptr.parent = obj
+    return NativeValue(struct_ptr._getvalue())
+
+
+@box(types.NumPyRandomGeneratorType)
+def box_numpy_random_bitgenerator(typ, val, c):
```
esc, line 1208:
```suggestion
def box_numpy_random_generator(typ, val, c):
```

Since this is boxing the `NumPyRandomGeneratorType` not the `NumPyRandomBitGeneratorType`

stuartarchibald:
I think this might summarise the current issue observed with reference counting for the `Generator` type. This shows what happens with a NumPy array and what happens with a custom type that does something similar to that required in the case of the Generator type. It might be easier debugging with this than with a full implementation?
```python
from contextlib import ContextDecorator
import gc
import sys
import numpy as np

from numba import njit, types, objmode
from numba.extending import typeof_impl, models, register_model
from numba.extending import box, unbox, NativeValue
from numba.core.runtime import rtsys


class Thing(object):
    def __init__(self):
        pass


class ThingType(types.Type):
    def __init__(self):
        super(ThingType, self).__init__(name='Thing')


thing_type = ThingType()


@typeof_impl.register(Thing)
def typeof_thing(val, c):
    return thing_type


@register_model(ThingType)
class ThingModel(models.StructModel):
        members = [
            ('meminfo', types.MemInfoPointer(types.voidptr)),
            ('parent', types.pyobject),
        ]


@unbox(ThingType)
def unbox_thing(typ, obj, c):
    thing_struct = cgutils.create_struct_proxy(typ)(c.context, c.builder)
    thing_struct.parent = obj
    NULL = cgutils.voidptr_t(None)
    thing_struct.meminfo = c.pyapi.nrt_meminfo_new_from_pyobject(
        NULL,  # there's no data
        obj,   # the python object, the call to nrt_meminfo_new_from_pyobject
               # will py_incref
    )
    is_py_error = cgutils.is_not_null(c.builder, c.pyapi.err_occurred())
    return NativeValue(thing_struct._getvalue(), is_error=is_py_error)


@box(ThingType)
def box_thing(typ, val, c):
    thing_struct = cgutils.create_struct_proxy(typ)(c.context, c.builder,
                                                    value=val)
    res = thing_struct.parent
    return res


class memory_leak_check(ContextDecorator):

    def __init__(self, the_obj):
        self._the_obj = the_obj

    def __enter__(self):
        # Clean up any NRT-backed objects hanging in a dead reference cycle
        gc.collect()
        print(f"Python refcount: {sys.getrefcount(self._the_obj)}")
        self.__init_stats = rtsys.get_allocation_stats()

    def __exit__(self, *exc):
        gc.collect()
        old = self.__init_stats
        new = rtsys.get_allocation_stats()
        total_alloc = new.alloc - old.alloc
        total_free = new.free - old.free
        total_mi_alloc = new.mi_alloc - old.mi_alloc
        total_mi_free = new.mi_free - old.mi_free
        print(f"Allocs: {total_alloc}. Frees: {total_free}")
        print(f"MI Allocs: {total_mi_alloc}. MI Frees: {total_mi_free}")
        gc.collect()
        print(f"Python refcount: {sys.getrefcount(self._the_obj)}")


def np_ty(): return np.ones(3)


for Clazz in (np_ty, Thing):
    print('\n', f"Testing Type: {Clazz}".center(80, '='), '\n')
    print('Testing no-op'.center(80, '_'))

    thing_inst1 = Clazz()
    print(f"Python refcount: {sys.getrefcount(thing_inst1)}")  # should be 2
    with memory_leak_check(thing_inst1):
        thing_inst1  # Python refcount should be 3, the ctxmanager holds 1 more

    @njit
    def foo(x):
        return x

    thing_inst2 = Clazz()

    print('Testing foo'.center(80, '_'))
    with memory_leak_check(thing_inst2):
        initial_id = id(thing_inst2)
        result = foo(thing_inst2)
        result_id = id(result)
        print(f"ID check: {result is thing_inst2}")

    @njit
    def bar(x):
        pass  # x unused

    print('Testing bar'.center(80, '_'))
    thing_inst3 = Clazz()
    with memory_leak_check(thing_inst3):
        bar(thing_inst3)

    @njit
    def baz(x):
        with objmode():
            print(f"In objmode block {sys.getrefcount(x)}")

    print('Testing baz'.center(80, '_'))
    thing_inst4 = Clazz()
    with memory_leak_check(thing_inst4):
        baz(thing_inst4)
```

kc611:
If we have the `incref` for the `parent` during boxing, the stats seems exactly the same. But does that cause further problems ?

(In the above example that would be equivalent to having `c.pyapi.incref(res)` during `boxing`)

numba/np/random/generator_core.py
kc611: The second problem is fixed by a version based branching for the function. Not sure about the first one though. My best guess is those datatypes will be handled correctly when we allocate this number to the resulting array in original function according to the specified datatype.

numba/tests/test_np_randomgen.py
```diff
@@ -0,0 +1,117 @@
...
+        numba_func = numba.njit(lambda x: x)
+        self.assertEqual(rng_instance, numba_func(rng_instance))
+        self.assertEqual(id(rng_instance), id(numba_func(rng_instance)))
+
+    def test_bitgen_funcs(self):
+        self._test_bitgen_func_parity("next_uint32", next_uint32)
+        self._test_bitgen_func_parity("next_uint64", next_uint64)
+        self._test_bitgen_func_parity("next_double", next_double)
+
+    def test_randomgen_caching(self):
```
esc, line 98: This test should also check that the function can be retrieved successfully from the cache. You can probably make use of https://github.com/numba/numba/blob/468647d/numba/tests/support.py#L881

kc611:
So I tried using `run_in_new_process_caching`, and since we don't yet have support for constructors for `Generator` inside Jit functions, I modified `run_in_new_process_caching` to support arguments. Apparently Python's `multiprocessing` is not happy with it and says (from azure logs): 

```
  File "/home/vsts/work/1/s/numba/tests/test_np_randomgen.py", line 108, in test_randomgen_caching
    res = run_in_new_process_caching(numba_func, func_args=(nb_rng,))
  File "/home/vsts/work/1/s/numba/tests/support.py", line 895, in run_in_new_process_caching
    return run_in_new_process_in_cache_dir(func, cache_dir, func_args, verbose=verbose)
  File "/home/vsts/work/1/s/numba/tests/support.py", line 918, in run_in_new_process_in_cache_dir
    proc.start()
  File "/home/vsts/miniconda3/envs/azure_ci/lib/python3.7/multiprocessing/process.py", line 110, in start
    'daemonic processes are not allowed to have children'
AssertionError: daemonic processes are not allowed to have children

```

kc611:
I tried using `run_in_new_process_caching`, and getting the following error in Azure pipelines (and unable to replicate locally):

```
  File "/home/vsts/work/1/s/numba/tests/test_np_randomgen.py", line 116, in test_randomgen_caching
    res = run_in_new_process_caching(test_generator_caching)
  File "/home/vsts/work/1/s/numba/tests/support.py", line 895, in run_in_new_process_caching
    return run_in_new_process_in_cache_dir(func, cache_dir, verbose=verbose)
  File "/home/vsts/work/1/s/numba/tests/support.py", line 918, in run_in_new_process_in_cache_dir
    proc.start()
  File "/home/vsts/miniconda3/envs/azure_ci/lib/python3.7/multiprocessing/process.py", line 110, in start
    'daemonic processes are not allowed to have children'
AssertionError: daemonic processes are not allowed to have children

```

numba/core/boxing.py
```diff
@@ -1105,3 +1105,105 @@ def unbox_typeref(typ, val, c):
...
+
+        # Wire up
+        setattr(struct_ptr, f'fnptr_{name}',
+                c.unbox(types.uintp, interface_next_fn_casted_value).value)
+
+    wire_in_fnptrs('next_double')
+    wire_in_fnptrs('next_uint64')
+    wire_in_fnptrs('next_uint32')
+
+    return NativeValue(struct_ptr._getvalue())
```
esc, line 1180:
```suggestion
    is_py_error = cgutils.is_not_null(c.builder, c.pyapi.err_occurred())
    return NativeValue(struct_ptr._getvalue(), is_error=is_py_error)
```

Since there are a lot of calls here to the pyapi, it may be a good idea to check for any errors on the python side.

wsjeon:
Hi, thanks for this commit, and I checked it worked well on my problems. While it worked well for JIT compilation, I'd like to move my implementation for AOT compilation, but I cannot figure out how to specify the datatype, e.g., I tried
```
import numpy as np
from numba import njit
from numba.core.types.npytypes import NumPyRandomGeneratorType
from numba import int64 as i8, boolean as b1, uint8 as u1, float64 as f8

@njit
@cc.export('cumsum', f8[:](f8[:]))
def cumsum(p):
    p = np.copy(p)
    for i in range(1, len(p)):
        p[i] += p[i -1]
    return p

@njit
@cc.export('choice', i8(f8[:], NumPyRandomGeneratorType))
def choice(p, rng):
    # Reference: https://github.com/numpy/numpy/blob/main/numpy/random/_generator.pyx#L762-L767
    cdf = cumsum(p)
    return np.searchsorted(cdf, rng.random(), side='right')
```

but it didn't work. Can I get a way to AOT compile with RNG?

Thanks!.

stuartarchibald:
@wsjeon thank you for testing this patch. RE AOT: As hinted at in the error message, the signature expects type instances not type classes. Something like this should work:
```python
from numba import njit, types
import numpy as np
from numba.pycc import CC

cc = CC('test8031')

# Create a NumPyRandomGeneratorType instance
type_inst = types.npytypes.NumPyRandomGeneratorType('rng')

@njit
@cc.export('foo', types.void(type_inst))
def foo(rng):
    print(rng.random())

cc.compile()

import test8031

x = np.random.default_rng()
test8031.foo(x)
```

@kc611 @esc wonder if it might be a good idea to create a default type instance for these types, I don't think there's any per-instance specialisation other than the "name"?

stuartarchibald:
@kc611 as discussed OOB, this patch fixes a few issues but it is _NOT_ for use in its current form! I've pasted it as-is as it hopefully conveys the ideas needed to fix the numerical problems:
1. float constants (like `1.0`) are typed as `float64` which will cause `float32` type promotion in a lot of operations.
2. ensuring the type of the argument to functions like `np.exp` and `np.log` is correct will ensure dispatch to the appropriate `libm` function. For example `log(float32(1.23))` is different to `float32(log(1.23))`, former calls `logf` and gets a `float32` result, latter calls `log` and casts the result to `float32`.
3. the use if `log1p` and `expm1` is recommended in cases where there's expressions like `log(1 + x)` and `exp(x) - 1` respectively. NumPy fixed up its distribution algs in version 1.21 to do this, suggest looking in NumPy's `distributions.c` to see precisely where.

Patch:
```diff
diff --git a/numba/np/random/distributions.py b/numba/np/random/distributions.py
index d2aa2f2..b216780 100644
--- a/numba/np/random/distributions.py
+++ b/numba/np/random/distributions.py
@@ -87,7 +87,7 @@ def random_standard_exponential(bitgen):
             return x
         else:
             if idx == 0:
-                return ziggurat_exp_r - np.log(1.0 - next_double(bitgen))
+                return ziggurat_exp_r - np.log1p(-next_double(bitgen))
             elif ((fe_double[idx - 1] - fe_double[idx]) * next_double(bitgen) +
                   fe_double[idx] < np.exp(-x)):
                 return x
@@ -105,16 +105,18 @@ def random_standard_exponential_f(bitgen):
             return x
         else:
             if (idx == 0):
-                return float32(ziggurat_exp_r_f -
-                               float32(np.log(1.0 - next_float(bitgen))))
+                # Don't do this, use log1p with a float32 arg.
+                # This is for demonstrative purposes only.
+                f32_one = np.float32(1.0)
+                return float32(ziggurat_exp_r_f - float32(np.log(np.float32(f32_one - next_float(bitgen)))))
             elif ((fe_float[idx - 1] - fe_float[idx]) * next_float(bitgen) +
-                  fe_float[idx] < np.exp(-x)):
+                  fe_float[idx] < np.float32(np.exp(np.float32(-x)))):
                 return x
 
 
 @register_jitable
 def random_standard_exponential_inv(bitgen):
-    return -np.log(1.0 - next_double(bitgen))
+    return -np.log1p(- next_double(bitgen))

```

stuartarchibald:
Also, additional thoughts... given there's branches on floating point values that could well be derived from expressions that have diverged across NumPy releases (e.g. the `log1p` fixes), some consideration is going to be needed about how to test the bit-wise replication. It may be good to focus on matching a single NumPy version to start with (maybe 1.21) and then look at how much algorithmic churn has occurred against the other versions, if it's not too bad then it might be possible to "port the bugs", else testing against "the most fixed" version might be the thing to do (and declare that in the docs).

esc:
Running a smoketest at: `Build numba_smoketest_cpu_yaml_103 has started`.

esc:
> @kc611 @esc wonder if it might be a good idea to create a default type instance for these types, I don't think there's any per-instance specialisation other than the "name"?

@stuartarchibald @wsjeon thanks to the work of @kc611 AOT appears to work now too:

```python
zsh» ls
test_8031.py

zsh» cat test_8031.py
import numpy as np
from numba import njit
from numba.core.types import npy_rng
from numba import int64 as i8, boolean as b1, uint8 as u1, float64 as f8
from numba.pycc import CC

cc = CC('test')
cc.output_dir = "CC_OUTPUT"

@njit
@cc.export('cumsum', f8[:](f8[:]))
def cumsum(p):
    p = np.copy(p)
    for i in range(1, len(p)):
        p[i] += p[i -1]
    return p

@njit
@cc.export('choice', i8(f8[:], npy_rng))
def choice(p, rng):
    # Reference: https://github.com/numpy/numpy/blob/main/numpy/random/_generator.pyx#L762-L767
    cdf = cumsum(p)
    return np.searchsorted(cdf, rng.random(), side='right')

print(choice(np.zeros(23, dtype=np.int64), np.random.default_rng()))

cc.compile()

zsh» python test_8031.py
23

 zsh» ls
CC_OUTPUT  test_8031.py

zsh» cd CC_OUTPUT
zsh» ls
test.cpython-39-darwin.so

zsh» python
Python 3.9.7 (default, Sep 16 2021, 08:50:36)
[Clang 10.0.0 ] :: Anaconda, Inc. on darwin
>>> import test
>>> dir(test)
['__doc__', '__file__', '__loader__', '__name__', '__package__', '__spec__', 'choice', 'cumsum']
>>>
```

esc:
> Running a smoketest at: `Build numba_smoketest_cpu_yaml_103 has started`.

FWIW: this was fine 🟢

numba/core/boxing.py
```diff
@@ -1105,3 +1107,186 @@ def unbox_typeref(typ, val, c):
...
+    * ('bit_generator', _bit_gen_type): The unboxed BitGenerator associated with
+                                        this Generator object instance.
+    * ('parent', types.pyobject): Pointer to the original Generator PyObject.
+    * ('meminfo', types.MemInfoPointer(types.voidptr)): The information about the memory
+        stored at the pointer (to the original Generator PyObject). This is useful for
+        keeping track of reference counts within the Python runtime. Helps prevent cases
+        where deletion happens in Python runtime without NRT being awareness of it. 
+    """
+    is_error_ptr = cgutils.alloca_once_value(c.builder, cgutils.false_bit)
+    fail_obj = c.context.get_constant_null(typ)
```
sklam, line 1263: unused
```suggestion
```

numba/np/random/generator_core.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+if numpy_version >= (1, 22):
+    @register_jitable
+    def next_float(bitgen):
+        return (next_uint32(bitgen) >> 8) * (1.0 / 16777216.0)
+else:
+    @register_jitable
+    def next_float(bitgen):
+        return (next_uint32(bitgen) >> 9) * (1.0 / 8388608.0)
```
sklam, line 123: Where are these constants (`16777216.0`, `8388608.0`) from? Can you add a reference to it?

numba/core/typing/typeof.py
```diff
@@ -265,3 +272,13 @@ def _typeof_nb_type(val, c):
         return types.NumberClass(val)
     else:
         return types.TypeRef(val)
+
+
+@typeof_impl.register(BitGenerator)
+def typeof_numpy_random_bitgen(val, c):
+    return types.NumPyRandomBitGeneratorType(val)
```
sklam, line 279: seems like this is not covered by any tests

numba/np/random/generator_core.py
kc611: These are from: https://github.com/numba/numba/pull/8031#discussion_r874689065

numba/core/typing/typeof.py
esc: I think it was added in [40e87e9](https://github.com/numba/numba/pull/8031/commits/40e87e903789460997195bb3d639ee48213fd13b)

numba/core/boxing.py
esc: Was removed in [40e87e9](https://github.com/numba/numba/pull/8031/commits/40e87e903789460997195bb3d639ee48213fd13b)

esc:
`Build numba_smoketest_cpu_yaml_105 has started`

esc (approved):

esc:
@kc611 thanks for the patch!

pwuertz:
The documentation added in this PR states that

> The current Numba support for :py:class:`Generator` is not thread-safe, hence we
do not recommend using :py:class:`Generator` methods in methods with parallel
execution logic.

Could you elaborate a bit on this? I understand that if I pass a `Generator` into a Numba function, Numba will take a reference to its `BitGenerator` for reading/writing its state - so naturally, one shouldn't use it concurrently from multiple threads. Are there any dangers beyond this type of misuse?

More specifically, is it safe to run multiple `nogil=True` Numba functions concurrently if each call receives a separate Generator,  e.g. created by `Generator.spawn()`? Or does the Numba `Generator` implementation use some sort of global state management that prohibits any sort of parallelism?

PR #8032, opened 2022-05-10
Pass null pointer args through as is

cluePrints:
Bumped into this
https://reviews.llvm.org/D67122

```
OMP: Info #271: omp_set_nested routine deprecated, please use omp_set_max_active_levels instead.
tmp/numba/0.53.1/numba/np/ufunc/omppool.cpp:185: runtime error: applying non-zero offset 96 to null pointer
    #0 0x7f86f24bc95a in .omp_outlined. tmp/numba/0.53.1/numba/np/ufunc/omppool.cpp:185
    #1 0x7f8706cfaa22 in __kmp_invoke_microtask /tmp/openmp/12/src/openmp/runtime/src/z_Linux_asm.S:1166
    #2 0x7f8706c8e5eb in __kmp_invoke_task_func /tmp/openmp/12/src/openmp/runtime/src/kmp_runtime.cpp:7308:30
    #3 0x7f8706c8d288 in __kmp_launch_thread /tmp/openmp/12/src/openmp/runtime/src/kmp_runtime.cpp:5829:34
    #4 0x7f8706ce115a in __kmp_launch_worker(void*) /tmp/openmp/12/src/openmp/runtime/src/z_Linux_util.cpp:591:33
    #5 0x7f872e497c2e in start_thread /tmp/glibc/2.34/src/glibc-2.34/nptl/pthread_create.c:434:8
    #6 0x7f872e52a1fb in clone3 /tmp/glibc/2.34/src/glibc-2.34/misc/../sysdeps/unix/sysv/linux/x86_64/clone3.S:81

SUMMARY: UndefinedBehaviorSanitizer: nullptr-with-nonzero-offset tmp/numba/0.53.1/numba/np/ufunc/omppool.cpp:185 in 
```

### DIFF
--- a/numba/np/ufunc/omppool.cpp
+++ b/numba/np/ufunc/omppool.cpp
@@ -182,7 +182,7 @@ parallel_for(void *fn, char **args, size_t *dimensions, size_t *steps, void *dat
                 char * base = args[j];
                 size_t step = steps[j];
                 ptrdiff_t offset = step * r;
-                array_arg_space[j] = base + offset;
+                array_arg_space[j] = base ? base + offset : base;
 
                 if(0&&_DEBUG)
                 {


esc:
@cluePrints thank you for submitting this. Do you have a reproducer by any chance?

cluePrints:
@esc no, unfortunately, not a cpp / llvm expert here. This happens a few levels down the stack from me at runtime in asan-enabled builds

Happy to play remote hands role here if that might be helpful. Maybe there's some sort of additional logging I can enable?

gmarkall:
@cluePrints What code / commands did you run to cause this error to be emitted?

gmarkall:
Also - I think we need to instead figure out why we're ending up with a null pointer there that we're attempting to offset. I don't think that just avoiding doing it at the last minute is the right fix, it will just mask the error you've originally seen without addressing the underlying issue.

cluePrints:
Added numba user code here (it's actually trimmed down version of https://github.com/facebookexperimental/fronni/blob/main/tests.py)
https://colab.research.google.com/drive/1gdKr7M0WAg6MTQVBFzyZWnhNjZJjpphO

I'm also trying to set up end-to-end repro but stuck rebuilding numba to add some logging and confirm - anything I'm missing here? (following https://numba.readthedocs.io/en/stable/user/installing.html#installing-from-source)

esc:
> Added numba user code here (it's actually trimmed down version of https://github.com/facebookexperimental/fronni/blob/main/tests.py) https://colab.research.google.com/drive/1gdKr7M0WAg6MTQVBFzyZWnhNjZJjpphO
> 
> I'm also trying to set up end-to-end repro but stuck rebuilding numba to add some logging and confirm - anything I'm missing here? (following https://numba.readthedocs.io/en/stable/user/installing.html#installing-from-source)

Hi, and thank you for adding this to the ticket. I had a look at the reproduction notebook, from what I can tell there seem to be only warnings which I think can be ignored? For example, the lack of `conda` is probably not a problem. Also, it seems like you were able to run the modified code w/o issues? 

One thing I don't understand is if you were able to reproduce the error using the patch you supplied?

esc:
> > One thing I don't understand is if you were able to reproduce the error using the patch you supplied?
> 
> No, I wasn't able to install numba from source and was looking for some help with that if possible

Oh, that's not good. Do you have an error message by any chance?

cluePrints:
> Oh, that's not good. Do you have an error message by any chance?

sorry, I should have been more precise. Install did ran well, it's just that numba was not usable: importing numba.config was failing so I concluded the build was the problem)

I've fixed 2 things in the build process since then: 1) git cloning to a folder which won't be targeted by python import & 2) llvmlite

it still appears to be not usable the way I've built it, the error message is below

P.S. can you see this output in colab?
```
----> 2 report = classification_report(label, predicted)

15 frames
[<ipython-input-27-f46a84422e4b>](https://localhost:8080/#) in classification_report(label, predicted, n, confidence_level, as_dict, sort_by_sample_size)
    176     df_length = len(df)
    177 
--> 178     labels = create_numba_list(np.unique(df["label"]))
    179     n_labels = len(labels)
    180     mapping_df = pd.DataFrame({"class": labels, "value": range(len(labels))})

[<ipython-input-27-f46a84422e4b>](https://localhost:8080/#) in create_numba_list(py_list)
    133     numba_list = List()
    134     for item in py_list:
--> 135         numba_list.append(item)
    136     return numba_list
    137 

/usr/local/lib/python3.7/dist-packages/numba/typed/typedlist.py in append(self, item)
/usr/local/lib/python3.7/dist-packages/numba/typed/typedlist.py in _initialise_list(self, item)
/usr/local/lib/python3.7/dist-packages/numba/typed/typedlist.py in _parse_arg(self, lsttype, meminfo, allocated)
/usr/local/lib/python3.7/dist-packages/numba/core/dispatcher.py in _compile_for_args(self, *args, **kws)
/usr/local/lib/python3.7/dist-packages/numba/core/dispatcher.py in _compile_for_args(self, *args, **kws)
/usr/local/lib/python3.7/dist-packages/numba/core/compiler_lock.py in _acquire_compile_lock(*args, **kwargs)
/usr/local/lib/python3.7/dist-packages/numba/core/dispatcher.py in compile(self, sig)
/usr/local/lib/python3.7/dist-packages/numba/core/dispatcher.py in compile(self, args, return_type)
/usr/local/lib/python3.7/dist-packages/numba/core/dispatcher.py in _compile_cached(self, args, return_type)
/usr/local/lib/python3.7/dist-packages/numba/core/dispatcher.py in _compile_core(self, args, return_type)
/usr/local/lib/python3.7/dist-packages/numba/core/compiler.py in compile_extra(typingctx, targetctx, func, args, return_type, flags, locals, library, pipeline_class)
/usr/local/lib/python3.7/dist-packages/numba/core/compiler.py in __init__(self, typingctx, targetctx, library, args, return_type, flags, locals)
/usr/local/lib/python3.7/dist-packages/numba/core/typing/context.py in refresh(self)
/usr/local/lib/python3.7/dist-packages/numba/core/typing/context.py in load_additional_registries(self)
ImportError: cannot import name 'cffi_utils' from 'numba.core.typing' (/usr/local/lib/python3.7/dist-packages/numba/core/typing/__init__.py)
```

esc:
> P.S. can you see this output in colab?

No, I couldn't find that error in colab, but also, I keep getting error messages about missing cookies and an expired google account so I am not convinced it is working correctly on my end. It is very unusual and I haven't seen this before. We have seen similar errors in the past though and they always pointed to a failed installation. I am not sure though what is going on in your case and I am not sure what to recommend.

stuartarchibald:
@cluePrints Thanks for reporting this, a couple of questions:
1. Does the code that produces the originally reported problem execute correctly when the environment variable `NUMBA_DISABLE_JIT=1` is set? Am asking this to make sure that this isn't being caused by corruption from e.g. out of bounds access in user code/libraries or from elsewhere in Numba.
2. If the threading layer is changed to TBB or workqueue, does UBSan report the same issue? (docs on how to do this: https://numba.readthedocs.io/en/stable/user/threading-layer.html#setting-the-threading-layer). Am asking this as I think all the threading backends would wire the gufunc args through in the same way, so all of them should report the same thing.

Many thanks for your help!

stuartarchibald:
For those debugging and for reference, the variable in question `base`:
https://github.com/numba/numba/blob/4684b9cf9d5e42ec0e144c3f238432b078627a1c/numba/np/ufunc/omppool.cpp#L185
is assigned from the `args` variable :
https://github.com/numba/numba/blob/4684b9cf9d5e42ec0e144c3f238432b078627a1c/numba/np/ufunc/omppool.cpp#L182
which is a `char **` argument to the `parallel_for` loop:
https://github.com/numba/numba/blob/4684b9cf9d5e42ec0e144c3f238432b078627a1c/numba/np/ufunc/omppool.cpp#L94-L95

IIRC this arg should contain the argument and output array(s). I'm not sure it can be `null` from a lowered Parfors node having too few inputs, but I guess it could be from the `array_count` exceeding the number of items supplied in `args`.

cluePrints:
@stuartarchibald thanks for looking into this and all the additional context, please see my my answers below

> Does the code that produces the originally reported problem execute correctly when the environment variable NUMBA_DISABLE_JIT=1 is set?

yup, it alleviates the reported problem

> If the threading layer is changed to TBB or workqueue, does UBSan report the same issue?

tbb - we don't built that internally, unfortunately
workqueue - yes, reports the same, traceback is a bit longer ([gist](https://gist.githubusercontent.com/cluePrints/b741c8dde5095209eb5e93d45b92c37a/raw/be750348a001b5a1a3e835cf2a172903d2c6ffe6/gistfile1.txt))

stuartarchibald:
> @stuartarchibald thanks for looking into this and all the additional context, please see my my answers below

No problem, many thanks for your help in debugging this.
 
> > Does the code that produces the originally reported problem execute correctly when the environment variable NUMBA_DISABLE_JIT=1 is set?
> 
> yup, it alleviates the reported problem

Thanks, that confirms that the code in use is likely valid/not corrupting the state, as things like out of bounds would be picked up by NumPy's APIs. It does not mean however that Numba's not some how corrupting its own state.
 
> > If the threading layer is changed to TBB or workqueue, does UBSan report the same issue?
> 
> tbb - we don't built that internally, unfortunately workqueue - yes, reports the same, traceback is a bit longer ([gist](https://gist.githubusercontent.com/cluePrints/b741c8dde5095209eb5e93d45b92c37a/raw/be750348a001b5a1a3e835cf2a172903d2c6ffe6/gistfile1.txt))

OK, thanks, that confirms that it is not backend specific and is likely related to incorrect code gen or corruption.

I think there might be a debug flag that can be switched on to give an idea of what the call site(s) look like, I'll take a look and see if it can show anything useful and report back.

stuartarchibald:
@cluePrints I think I've caught it.

I'm reasonably sure that the issue is related to this assumption:
https://github.com/numba/numba/blob/468647dddde27ee8af124c97dfcd20c35c4a2bc6/numba/np/ufunc/parallel.py#L132-L133

it makes the array count one larger than the length of the signature through assuming there's an output array, but in the case of a parfors lowered loop I'm not sure that's the case.

Trivial reproducer:
```python
from numba import njit
import numpy as np

@njit(parallel=True)
def foo(n):
    x = np.ones((n,n))
    return x

print(foo(1))
```
run with:
```
NUMBA_THREADING_LAYER=omp python test8032.py
```
The `base` is `NULL` (or sometimes random junk if you change `n`) on the last index (i.e. the off-by-one read)
https://github.com/numba/numba/blob/4684b9cf9d5e42ec0e144c3f238432b078627a1c/numba/np/ufunc/omppool.cpp#L182

I think the fix has to come from code generation, the threading backend seems ok. , 

@cluePrints I'm pretty sure it's a code generation issue, are you happy to leave it with me and @DrTodd13 (the codeowner)?

stuartarchibald:
After OOB discussion with @DrTodd13, I think this is a more general problem with gufunc code generation that extends above the parfors call sites, as a result this is not a parfors issue (so un-CC @DrTodd13, and thanks for the input/discussion!).

Example:
```python
from numba import guvectorize
import numpy as np

@guvectorize(['void(int64[::1], int64, int64[::1])',], '(n),()->(n)', target='parallel')
def g(x, y, res):
    for i in range(x.shape[0]):
        res[i] = x[i] + y

x = np.arange(5, dtype=np.int64)

y = 10

res = np.zeros_like(x)

g(x, y, res)
print(res)
print(g.types)
```
and applying this patch to get debug:
```diff
diff --git a/numba/np/ufunc/omppool.cpp b/numba/np/ufunc/omppool.cpp
index d9208e5..1142f29 100644
--- a/numba/np/ufunc/omppool.cpp
+++ b/numba/np/ufunc/omppool.cpp
@@ -24,7 +24,7 @@ Threading layer on top of OpenMP.
 #include <signal.h>
 #endif
 
-#define _DEBUG 0
+#define _DEBUG 2
 #define _DEBUG_FORK 0
 
 // OpenMP vendor strings
@@ -184,7 +184,7 @@ parallel_for(void *fn, char **args, size_t *dimensions, size_t *steps, void *dat
                 ptrdiff_t offset = step * r;
                 array_arg_space[j] = base + offset;
 
-                if(0&&_DEBUG)
+                if(1&&_DEBUG)
                 {
                     printf("Index %zu\n", j);
                     printf("-->Got base %p\n", (void *)base);
```
Then gives:
```
Using parallel_for
inner_ndim: 1
arg_len: 2
total: 1
dimensions: 1, 5, 
steps: 0, 0, 0, 8, 
*args: 0x55ea9ae04330, 0x55ea9a460cf0, 0x55ea9ae0dea0, (nil), 
THREAD 0x7fffe3638dd0:count_space: 1, 5, 
Index 0
-->Got base 0x55ea9ae04330
-->Got step 0
-->Got offset 0
-->Got addr 0x55ea9ae04330
Index 1
-->Got base 0x55ea9a460cf0
-->Got step 0
-->Got offset 0
-->Got addr 0x55ea9a460cf0
Index 2
-->Got base 0x55ea9ae0dea0
-->Got step 0
-->Got offset 0
-->Got addr 0x55ea9ae0dea0
Index 3
-->Got base (nil)
-->Got step 8
-->Got offset 0
-->Got addr (nil)
array_arg_space: 0x55ea9ae04330, 0x55ea9a460cf0, 0x55ea9ae0dea0, (nil), 
[10 11 12 13 14]
['ll->l']
```
i.e. the final index is erroneous, and this happens outside of parfors, so it's a more general issue.

Good news is that I'm reasonably convinced that https://github.com/numba/numba/pull/8081 fixes it.

cluePrints:
> are you happy to leave it with me and @DrTodd13 (the codeowner)?

yes, please!

> Good news is that I'm reasonably convinced that https://github.com/numba/numba/pull/8081 fixes it.

thanks for the patch - fixes the problem on my internal build too

stuartarchibald:
> > are you happy to leave it with me and @DrTodd13 (the codeowner)?
> 
> yes, please!

Great, thanks for confirming.

> > Good news is that I'm reasonably convinced that #8081 fixes it.
> 
> thanks for the patch - fixes the problem on my internal build too

No problem, thanks for checking that it fixed the issue. With reference to https://github.com/numba/numba/pull/8032#issuecomment-1128650530, in practice I think it's an unnecessary, but likely safe, over allocation by one element that occurs when lowering `gufunc`s for execution on a parallel backend. Whatever happens to result from `args[j] + offset` gets written into the "extra" element, but it's not accessed (if it were, because it's nonsense, most likely outcome is segfault!).

cluePrints:
Closed in favor of #8081

PR #8044, opened 2022-05-12
Make Python 3.10 kwarg peephole less restrictive

### LINKED ISSUE #8043: Python 3.10 kwarg peephole rewrite failing in valid case (breaks mainline)
## Reporting a bug

This breaks on mainline and is a blocking issue for 0.56.
```python
from numba import njit, objmode

def func(*args):
    pass

@njit
def log(*args):
    with objmode():
        func(*args)

log()
```
the above reproducer is extracted from the failing tests found  by running on Python 3.10:
```
./runtests.py numba.tests.test_ndarray_subclasses
```

njriasan:
Closes #8043

Fixes a bug in peep_hole_call_function_ex_to_call_function_kw where we raised the error if we couldn't find the varargs in the basic block and no kwargs are passed. This is an inaccurate check and we should only raise an exception if we find the args have the wrong definition.

This also fixes a bug in `peep_hole_list_to_tuple` where the definition of the final assignment wasn't properly set.

### DIFF
--- a/numba/core/interpreter.py
+++ b/numba/core/interpreter.py
@@ -702,11 +702,7 @@ def peep_hole_call_function_ex_to_call_function_kw(func_ir):
                         found = True
                     else:
                         vararg_loc -= 1
-                if not found:
-                    # If we couldn't find where the args are created
-                    # then we can't handle this format.
-                    raise UnsupportedError(errmsg)
-                else:
+                if found:
                     # If this value is still a list to tuple raise the
                     # exception.
                     expr = blk.body[vararg_loc].value
--- a/numba/core/interpreter.py
+++ b/numba/core/interpreter.py
@@ -90,19 +90,6 @@ def _remove_assignment_definition(old_body, idx, func_ir):
     rhs = old_body[idx].value
     if rhs in func_ir._definitions[lhs]:
         func_ir._definitions[lhs].remove(rhs)
-    elif len(func_ir._definitions[lhs]) == 1:
-        # There appears to be an issue where the RHS of the
-        # definitions doesn't match after the list -> tuple
-        # conversion.
-        # For example the statement is:
-        #
-        #   172list_to_tuple.116 = $170list_append.115
-        #
-        # but the definition is:
-        #
-        #   list_to_tuple(info=('$4build_list.1',))
-        #
-        func_ir._definitions[lhs].clear()
     else:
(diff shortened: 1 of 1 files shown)


sklam:
I did a quick check. It does resolve the failure in `numba.tests.test_ndarray_subclasses.TestNdarraySubclasses`

sklam:
smoketest BFID: `numba_smoketest_cpu_yaml_83`

njriasan:
> > smoketest BFID: `numba_smoketest_cpu_yaml_83`
> 
> This passed on multiple build jobs with Python 3.10, and all other build jobs not impacted by some other unrelated issue.
> 
> Just checked into the few failing Python 3.10 jobs, there's still a couple that are reporting the issue in #8043. Need to take a closer look as to what has happened.

@stuartarchibald If you post the list of tests I'm happy to take a look.

stuartarchibald:
> @stuartarchibald If you post the list of tests I'm happy to take a look.

Thanks @njriasan it's these two tests, same failures as in #8043:
* `numba.tests.test_ndarray_subclasses.TestNdarraySubclasses.test_my_array_return`
* `numba.tests.test_ndarray_subclasses.TestNdarraySubclasses.test_my_array_allocator_override`

What I'm not sure about is if this is caused by a genuine issue with this patch or there's something unusual going on in the farm builds.

Set up was Python 3.10, np 1.21, linux 64 bit system.

njriasan:
@stuartarchibald I reran the whole test file manually using a ubuntu x86 docker image. I could not reproduce any failures. I'm inclined to believe this may be a build farm issue.

sklam:
@stuartarchibald, I think we might be mixing up the result of two builds. I'm going to combine the two PRs (#8044 and #8046) to tests.

stuartarchibald:
> @stuartarchibald, I think we might be mixing up the result of two builds. I'm going to combine the two PRs (#8044 and #8046) to tests.

Thanks @sklam

stuartarchibald:
> @stuartarchibald I reran the whole test file manually using a ubuntu x86 docker image. I could not reproduce any failures. I'm inclined to believe this may be a build farm issue.

@njriasan Thanks for checking.

sklam:
tested as part of #8053 and its all green

sklam (approved):
Thanks for the patch!

PR #8049, opened 2022-05-13
support str(bool) syntax

dlee992:
Support convert Boolean/LiteralBoolean value to corresponding string value, like python does.


esc (approved):
LGTM thank you for the patch!

Issue #7302, opened 2021-08-12
PPC64LE undefined symbol `pthread_atfork` error in helperlib


sklam:
On ppc64le platform, some Numba C-extension libraries will fail on import with `undefined symbol: pthread_atfork`.

Upon investigation, the upgrade from `python 3.9.5-h836d2c2_3` to `python 3.9.6-h836d2c2_0` (conda packages) is the start of the problem in our internal build farm.

sklam:
Might be easier to fix this by setting `CC` to include `-pthread`.

The problem in python3.9.6 is coming from https://github.com/conda-forge/python-feedstock/pull/477/files#diff-8c19aabdb077038b2aea79fb69cb607c7bf2406a0a687b1d5c23d947d4a2b24fR16 for https://bugs.python.org/issue24935. The conda build of cpython includes `-pthread` as part of the `CC` in sysconfig (e.g. `<compiler_name> -pthread`). The referenced line will strip `-pthread` from the later compiler invocation.

PR #8052, opened 2022-05-13
Ensure pthread is linked in when building for ppc64le.

sklam:
This problem is a variation of #7302.


sklam:
tested as part of #8053 and its all green

apmasell (approved):

gmarkall (approved):
I've just looked at the notes and discussion on #8053 and the changes here including the addition of the pycc fix make sense, and #8053 suggests that this PR helps towards stabilization.

PR #8056, opened 2022-05-16
Move caching tests from test_dispatcher to test_caching

gmarkall:
This moves the cache tests from `test_dispatcher.py` and `test_cfunc.py` into `test_caching.py` - this organises things a little better as `test_dispatcher.py` was huge, `test_caching.py` contained very few cache tests, and the cfunc cache tests were a small part of `test_cfunc.py`. Other tests that test caching are already in their own dedicated files, so these are left as-is except for updating imports.

@stuartarchibald and I did discuss refactoring to collapse the `BaseCacheTest` / `BaseCacheUseCasesTest` split, but now I see that both of these are used in different places I decided against trying to do that - instead, it seems to make the code more clearly-structured to just rename `BaseCacheUseCasesTest` to `DispatcherCacheUseCasesTest`.


gmarkall:
gpuci run tests

(because you just never know!... :-) )

numba/tests/test_dispatcher.py
```diff
@@ -1003,846 +949,6 @@ def pyfunc(A, i, j):
...
-        out, err = popen.communicate()
-        if popen.returncode != 0:
-            raise AssertionError(
-                "process failed with code %s: \n"
-                "stdout follows\n%s\n"
-                "stderr follows\n%s\n"
-                % (popen.returncode, out.decode(), err.decode()),
-            )
-
-    def check_module(self, mod):
```
stuartarchibald, line 1090: @gmarkall from OOB discussion, this was the function that we established did nothing/was never run?

numba/tests/test_caching.py
```diff
@@ -133,3 +192,886 @@ def test(num=1000):
...
+
+        with warnings.catch_warnings():
+            warnings.simplefilter('error', NumbaWarning)
+
+            f = mod.closure1
+            self.assertPreciseEqual(f(3), 6) # 3 + 3 = 6
+            f = mod.closure2
+            self.assertPreciseEqual(f(3), 8) # 3 + 5 = 8
+            f = mod.closure3
+            self.assertPreciseEqual(f(3), 10) # 3 + 7 = 8
```
stuartarchibald, line 435: 3 + 7 = 10 !

numba/tests/test_caching.py
```diff
@@ -133,3 +192,886 @@ def test(num=1000):
...
+        # Test caching in an IPython session
+        base_cmd = [sys.executable, '-m', 'IPython']
+        base_cmd += ['--quiet', '--quick', '--no-banner', '--colors=NoColor']
+        try:
+            ver = subprocess.check_output(base_cmd + ['--version'])
+        except subprocess.CalledProcessError as e:
+            self.skipTest("ipython not available: return code %d"
+                          % e.returncode)
+        ver = ver.strip().decode()
+        print("ipython version:", ver)
```
stuartarchibald, line 605: delete?

numba/tests/test_dispatcher.py
gmarkall: It is.

gmarkall:
@stuartarchibald Many thanks for the review, your comments should be addressed now.

numba/tests/test_dispatcher.py
stuartarchibald: Thanks, its removal makes sense!

esc:
Thank you for the patch!

PR #8057, opened 2022-05-16
Fix coverage checking

gmarkall:
`run_coverage.py` seemed to be broken in a couple of ways - these updates allow me to run it locally again - for example, invoking

```
./run_coverage.py numba.cuda.tests
```

results in running the CUDA tests, with a coverage report produced in the `htmlcov` directory.


run_coverage.py
```diff
@@ -16,22 +16,36 @@
...
+                  "requested. Aborting.")
+            sys.exit(1)
+
     # We must start coverage before importing the package under test,
     # otherwise some lines will be missed.
     config_file = os.path.join(
-        os.path.dirname(os.path.dirname(__file__)),
-        '.coveragerc')
+        os.path.dirname(__file__),
+        'coverage.conf')
```
sklam, line 39: There's already a `.coveragerc` (https://github.com/numba/numba/blob/main/.coveragerc)

coverage.conf
```diff
@@ -0,0 +1,15 @@
+# configuration file used by run_coverage.py
```
sklam, line 1: There's already a `.coveragerc`

gmarkall:
@sklam thanks for the feedback. WIth testing across Linux and Windows, and different ways of invoking it, e.g.

```
python run_coverage.py
python ./run_coverage.pu
./run_coverage.py
# etc.
```

I found there's a bunch of fiddly issues still to resolve that seem a bit pointless to plough much time into. Instead, I think it's better to document the workflow using the coverage CLI tool, so I've updated this PR accordingly.

- Do you follow these exact steps?
- Do you have any other coverage-related tips?

PR #7992, opened 2022-04-19
Pin llvmlite=0.39.0dev0 builds to build 63 on OSX

stuartarchibald:
Builds later than 63 are breaking public CI for OSX due to
looking for a linux based soname when resolving library
linkage of libllvmlite.dylib. This pins the version
appropriately.

PR #8058, opened 2022-05-16
Fix/pin llvmlite 039

esc:
Debug #7992


esc:
After further investigation, the build `80` of llvmlite in the channel `numba/label/dev` was built fine on the 27th April. The public CI for `main` has been using these builds for sometime, so we assume that the issue was transient and is no longer present after a rebuild of llvmlite via the Anaconda build farm.

PR #8027, opened 2022-05-06
Support for NumPy 1.22

stuartarchibald:
As title.

Fixes #7756
Fixes #7757
Fixes #7759
Fixes #7760

PR #8067, opened 2022-05-19
Backport #8027: Support for NumPy 1.22

esc:
As title


stuartarchibald (approved):
Thanks for doing the back port. I've locally replayed the same cherry-pick of the merge from #8027 onto `release0.55` and resolved the conflict in the `installing.rst`, the diff against this PR is empty which suggests it's as expected.

buildscripts/condarecipe.local/meta.yaml
```diff
@@ -58,6 +58,8 @@ requirements:
     - scipy >=1.0
     # CUDA Python 11.6 or later
     - cuda-python >=11.6
+    # NumPy 1.22.0, 1.22.1, 1.22.2 are all broken for ufuncs, see #7756
+    - numpy !=1.22.0, !=1.22.1, !=1.22.2
```
sklam, line 62: this should be move into `run` requirement as per #8102

PR #8077, opened 2022-05-19
change return type of np.broadcast_shapes to a tuple

guilhermeleobas:
This is a follow-up PR of https://github.com/numba/numba/pull/8003. It changes the return type of `np.broadcast_shapes` to a tuple, which matches the [NumPy spec](https://numpy.org/doc/stable/reference/generated/numpy.broadcast_shapes.html#numpy.broadcast_shapes).

I also notice in PR https://github.com/numba/numba/pull/7067 that by returning a list, `np.broadcast_shapes` is pretty much unusable.

### DIFF
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -1425,7 +1425,7 @@ def ol_numpy_broadcast_shapes(*args):
 
     # discover the number of dimensions
     m = 0
-    for arg in literal_unroll(args):
+    for arg in args:
         if isinstance(arg, types.Integer):
             m = max(m, 1)
         elif isinstance(arg, types.BaseTuple):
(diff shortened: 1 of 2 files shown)


numba/np/arrayobj.py
```diff
@@ -1423,24 +1423,32 @@ def ol_numpy_broadcast_shapes(*args):
...
-        for arg in literal_unroll(args):
-            if isinstance(arg, tuple) and len(arg) > 0:
-                numpy_broadcast_shapes_list(r, m, arg)
-            elif isinstance(arg, int):
-                numpy_broadcast_shapes_list(r, m, (arg,))
-        return r
-    return impl
+    # discover the number of dimensions
+    m = 0
+    for arg in literal_unroll(args):
```
stuartarchibald, line 1428:
```suggestion
    for arg in args:
```
This is pure python code?

stuartarchibald (commented):

numba/np/arrayobj.py
guilhermeleobas: Done!

stuartarchibald (approved):

PR #7804, opened 2022-01-31
update local references from master -> main

esc:
As title.

PR #8078, opened 2022-05-19
Backport #7804: update local references from master -> main

esc:
As title.

### DIFF
--- a/numba/testing/main.py
+++ b/numba/testing/main.py
@@ -207,7 +207,7 @@ class NumbaTestProgram(unittest.main):
         parser.add_argument('-g', '--gitdiff', dest='gitdiff', type=git_diff_str,
                             default=False, nargs='?',
                             help=('Run tests from changes made against '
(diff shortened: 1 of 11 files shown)


numba/testing/main.py
```diff
@@ -399,9 +399,9 @@ def _choose_gitdiff_tests(tests, *, use_common_ancestor=False):
     path = os.path.join('numba', 'tests')
     if use_common_ancestor:
         print(f"Git diff by common ancestor")
-        target = 'origin/master...HEAD'
+        target = 'origin/release0.55...HEAD'
     else:
-        target = 'origin/master..HEAD'
+        target = 'origin/release0.55..HEAD'
```
stuartarchibald, line 404: Am wondering if this should be done on every release branch as part of the release process.

esc: probably a good idea

sklam (approved):

PR #8080, opened 2022-05-19
Fix windows test failure due to timeout when the machine is slow poss…

sklam:
…ibly due to high load.


stuartarchibald (approved):
Thanks for the patch. Confirmed by @esc to fix the issue seen on the farm.

PR #8081, opened 2022-05-19
Fix erroneous array count in parallel gufunc kernel generation.

stuartarchibald:
As title.


stuartarchibald:
Observation: The code path for ufunc and gufunc both run through the code changed in the diff, I think the +1 to the array count is only needed in the case of a ufunc. Next thing to try is branching on `types.none` return type to skip adding the 1.

stuartarchibald:
xref reproducer: https://github.com/numba/numba/pull/8032#issuecomment-1131991661
xref debug: https://github.com/numba/numba/pull/8032#issuecomment-1132662142

sklam (approved):
Make sense, and existing code already cover testing it.

Without this patch, an extra slot would be allocated when the type signature has a "void" return (output array declared).

PR #8082, opened 2022-05-19
Backport #8080: fix windows failure due to timeout

sklam:
as titled


stuartarchibald (approved):
Thanks for the patch. Matches #8080 on inspection.

PR #8089, opened 2022-05-22
[WIP] CUDA caching

gmarkall:
For testing on CI as I go along


gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
@stuartarchibald This is now ready for review.

numba/cuda/dispatcher.py
```diff
@@ -106,23 +107,50 @@ def __init__(self, py_func, argtypes, link=None, debug=False,
         self._codelibrary = lib
         self.call_helper = cres.call_helper
 
+        # The following are referred to by the cache implementation. Note:
+        # - There are no referenced environments in CUDA.
+        # - Kernels don't have lifted code.
+        # - reload_init is only for parfors.
```
stuartarchibald, line 113: Currently it's only used by parfors, but the use extends to anything which needs some sort of "state" reinitialising.

numba/cuda/tests/cudapy/cache_usecases.py
```diff
@@ -0,0 +1,170 @@
...
+from numba.cuda.testing import CUDATestCase
+import numpy as np
+import sys
+
+
+class UseCase:
+    """
+    Provide a way to call a kernel as if it were a function.
+
+    This allows the CUDA cache tests to match closely the CPU cache tests, and
```
stuartarchibald, line 11:
```suggestion
    This allows the CUDA cache tests to closely match the CPU cache tests, and
```

numba/cuda/tests/cudapy/cache_usecases.py
```diff
@@ -0,0 +1,170 @@
+from numba import cuda
+from numba.cuda.testing import CUDATestCase
+import numpy as np
+import sys
+
+
+class UseCase:
+    """
+    Provide a way to call a kernel as if it were a function.
```
stuartarchibald, line 9: Thanks for doing this, it makes it easier to verify and maintain!

gmarkall: You're welcome... I did think about trying to align the CUDA and CPU tests even more closely, but I worried it would result in a tangled incomprehensible mess. I think there's no perfect solution, but hopefully this whole setup doesn't cause too much maintenance burden in future :slightly_smiling_face:

numba/cuda/tests/cudapy/test_caching.py
```diff
@@ -0,0 +1,416 @@
...
+                     "cannot easily make a directory read-only on Windows")
+    def test_non_writable_pycache(self):
+        # Make it impossible to write to the __pycache__ directory
+        pycache = os.path.join(self.tempdir, '__pycache__')
+        os.mkdir(pycache)
+        old_perms = os.stat(pycache).st_mode
+        os.chmod(pycache, 0o500)
+        self.addCleanup(os.chmod, pycache, old_perms)
+
+        self._test_pycache_fallback()
```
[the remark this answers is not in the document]
gmarkall: I had a vague suspicion that there could be some issue related to the fallback dir being shared by kernels, and the keying of cache files being different that could potentially cause some issue. I think that I could possibly delete this check and it not be too much of an issue, but I'd be more comfortable just leaving it in given that it passes and exercises the fallback in a slightly different way.

Are you OK with keeping it?

gmarkall:
gpuci run tests

gmarkall:
@stuartarchibald Many thanks for the review - I've applied the two suggestions and responded to the other comments.

numba/cuda/tests/cudapy/test_caching.py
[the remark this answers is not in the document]
stuartarchibald: I think it's fine to keep it as it has value as noted. Thanks!

numba/cuda/tests/cudapy/cache_usecases.py
stuartarchibald: I think this is a good compromise in terms of alignment vs. complexity, am happy to leave as is!

stuartarchibald (dismissed):
Thanks for the fixes. Patch approved pending manual testing.

sklam (approved):
I did manual testing on a multi-GPUs (CC 6.0, CC 3.5) machine and can confirm that the cache is working as expected.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_135`.

sklam:
@gmarkall, there's now a small merge conflict

gmarkall:
gpuci run tests

stuartarchibald (approved):
Merge commit looks good, thanks @gmarkall.

PR #8093, opened 2022-05-22
Backport #7904: Support m1

esc:
As title.


stuartarchibald (approved):
Thanks for the patch. Checked this locally by cherry-picking the merge commit to `main` of #7904 onto the `release0.55` branch, after which `git diff` reported no difference against this patch. Manual inspection confirms the same. Thanks again!

PR #8094, opened 2022-05-22
Backport #8052 Ensure pthread is linked in when building for ppc64le.

esc:
As title.


stuartarchibald (approved):
Thanks for the patch. Backported source looks the same as #8052 on inspection.

PR #8097, opened 2022-05-23
Exclude libopenblas 0.3.20 on osx-arm64

esc:
Fixes: #8096

PR #8098, opened 2022-05-23
Backport #8097: Exclude libopenblas 0.3.20 on osx-arm64

esc:
As title


stuartarchibald (approved):
Thanks for the patch, on inspection this matches https://github.com/numba/numba/pull/8097.

PR #7786, opened 2022-01-26
Remove dependency on intel-openmp for OSX

stuartarchibald:
Solely the llvm-openmp package can provide the necessary headers
and runtime.

Fixes #7468

PR #8100, opened 2022-05-23
Backport #7786 for 0.55.2: Remove dependency on intel-openmp for OSX

esc:
As title.


stuartarchibald (approved):
Thanks for the patch. Checked this locally by cherry-picking the merge commit to `main` of #7786 onto the `release0.55` branch and resolving the conflict in `buildscripts/condarecipe.local/meta.yaml`, after which `git diff` reported no difference against this patch. Manual inspection confirms the same. Thanks again!

PR #8104, opened 2022-05-24
WIP: support numpy array comparison with string

### LINKED ISSUE #7859: Incorrect behavior with numpy structured array
Hello,
I have a different behavior when executing the following code with or with numba:

import numpy as np
import numba as nb
table = np.zeros(2, dtype=([('field1', np.uint8), ('field2', '<U1')]))
table[0] = (1, 'A')
table[1] = (2, 'B')
@nb.njit
def foo(table):
    print(table['field1'] == 1)
    print(table['field2'] == 'A')
    return
foo(table)

Without numba: (prints two vectors of boolean, OK)
[ True False]
[ True False]

With numba: (second print return just a boolean, NOK)
[ True False]
False

Thanks for your help!
Anthony

dlee992:
fix #7859 .

Choose an alternative to implement this feature:

- [x] Aligned with NumPy, use special case to handle string ufunc (I don't want to choose this way, because I guess this makes string case cannot use ufunc-related optimization?)
- [x] Pending on NumPy new version (1.23?), then use an unified way to handle all ufuncs
  -  tested on NumPy==1.23.0rc2, it doesn't unify string case and others. [2022-6-6]

Updates on 2022-6-21:

- [ ] I choose to help NumPy to complete the string ufunc unification PRs. 
- [ ] Then back to Numba string ufuncs.

### DIFF
--- a/numba/tests/test_array_exprs.py
+++ b/numba/tests/test_array_exprs.py
@@ -738,5 +738,21 @@ class TestExternalTypes(MemoryLeakMixin, unittest.TestCase):
             np.testing.assert_array_equal(arr_expr_neg(4), np.array([-4]))
 
 
+class TestStringExpressions(MemoryLeakMixin, unittest.TestCase):
+
+    def test_eq(self):
+        @njit
+        def foo(a, s):
+            return a == s
+
+        a = np.array(["A", "B"])
+        s = "A"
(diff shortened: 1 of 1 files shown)


dlee992:
Indeed, I didn't follow this thread these days. So I guess I should just close it.

PR #8109, opened 2022-05-24
Pin TBB support with respect to incompatible 2021.6 API.

stuartarchibald:
As title. Numba is not yet compatible with TBB 2021.6 API, but the
ABI seems ok.

PR #8114, opened 2022-05-25
Backport #8109 Pin TBB support with respect to incompatible 2021.6 API.

stuartarchibald:
As title. Numba is not yet compatible with TBB 2021.6 API, but the
ABI seems ok.

### DIFF
--- a/buildscripts/condarecipe.local/meta.yaml
+++ b/buildscripts/condarecipe.local/meta.yaml
@@ -37,9 +37,10 @@ requirements:
     - llvmlite 0.38.*
     # TBB devel version is to match TBB libs.
     # 2020.3 is the last version with the "old" ABI
+    # NOTE: 2021.1..2021.5 are API compatible for Numba's purposes.
     # NOTE: ppc64le exclusion is temporary until packages are more generally
     #       available.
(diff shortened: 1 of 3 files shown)


ZhenShuo2021:
I have similar problem like this, but I don't know whether this problem belongs to tbb or numba.
I tried to installed tbb by conda and pip and both of the `threading_layer()` prints omp, and import tbb says no module name 'tbb'. Same problem also post in [stack overflow](https://stackoverflow.com/questions/62518874/no-module-named-tbb) but there was no answer.

CPU: i9-10900k
Numba version: 0.55.1
tbb version: 2021.6.0

esc:
> I have similar problem like this, but I don't know whether this problem belongs to tbb or numba. I tried to installed tbb by conda and pip and both of the `threading_layer()` prints omp, and import tbb says no module name 'tbb'. Same problem also post in [stack overflow](https://stackoverflow.com/questions/62518874/no-module-named-tbb) but there was no answer.
> 
> CPU: i9-10900k Numba version: 0.55.1 tbb version: 2021.6.0

@ZhenShuo2021 can you open a new issues for this? Please include:

* What commands you entered
* What you expected
* What happened instead

Thank you.

PR #8120, opened 2022-05-30
Support nesting of nested array types

### LINKED ISSUE #1469: Support nestedarray as dtype of Array
This is not supported:

``` python
import numpy as np
import numba
qtype = np.dtype((np.float64,(4,)))
ntype = numba.types.Array(numba.from_dtype(qtype),2,'C')
```

and results in:

```
numba/types.py in __init__(self, dtype, ndim, layout, readonly, name)
    728         if isinstance(dtype, Buffer):
    729             print(type)
--> 730             raise TypeError("Buffer dtype cannot be buffer")
    731         if layout not in self.LAYOUTS:
    732             raise ValueError("Invalid layout '%s'" % layout)

TypeError: Buffer dtype cannot be buffer
```

Note that:

``` python
In [7]: numba.from_dtype(qtype)
Out[7]: nestedarray(float64, (4,))
In [16]: a = np.zeros(4, dtype=qtype)

In [17]: a.shape
Out[17]: (4, 4)

In [18]: a.dtype
Out[18]: dtype('float64')
```

### LINKED ISSUE #3158: Structured dtypes with multi-dim fields: numba either throws or crashes
Given this (quite trivial) example:

```python
dtype1 = np.dtype([('a', 'i8'), ('b', 'i4')])
dtype2 = np.dtype((dtype1, (2, 2)))
dtype3 = np.dtype([('x', '?'), ('y', dtype2)])
arr = np.zeros(2, dtype3)

@numba.jit(nopython=True)
def foo(arr):
    return arr[0]

foo(arr)
```

numba will most likely throw like so:

```
.../numba/types/npytypes.py in __init__(self, dtype, shape)
    419 
    420     def __init__(self, dtype, shape):
--> 421         assert dtype.bitwidth % 8 == 0, \
    422             "Dtype bitwidth must be a multiple of bytes"
    423         self._shape = shape

AttributeError: 'Record' object has no attribute 'bitwidth'
```

... or you could also get

```
SystemError: 
CPUDispatcher(<function foo at 0x7f7741382d08>) returned NULL without setting an error
```

gmarkall:
Adds support for nesting of nested array types. Test cases are based on those provided in Issues #1469 and #3158.

Fixes #1469.
Fixes #3158.

### DIFF
--- a/numba/core/types/npytypes.py
+++ b/numba/core/types/npytypes.py
@@ -212,6 +212,10 @@ class Record(Type):
 
         return as_struct_dtype(self)
 
+    @property
+    def bitwidth(self):
+        return self.dtype.itemsize * 8
+
     def can_convert_to(self, typingctx, other):
         """
         Convert this Record to the *other*.
@@ -430,6 +434,10 @@ class Array(Buffer):
         if (not aligned or
             (isinstance(dtype, Record) and not dtype.aligned)):
             self.aligned = False
+        if isinstance(dtype, NestedArray):
+            tmp = Array(dtype.dtype, dtype.ndim, 'C')
+            ndim += tmp.ndim
+            dtype = tmp.dtype
         if name is None:
             type_name = "array"
             if not self.mutable:
@@ -556,6 +564,10 @@ class NestedArray(Array):
     """
 
     def __init__(self, dtype, shape):
+        if isinstance(dtype, NestedArray):
+            tmp = Array(dtype.dtype, dtype.ndim, 'C')
+            shape += dtype.shape
+            dtype = tmp.dtype
         assert dtype.bitwidth % 8 == 0, \
             "Dtype bitwidth must be a multiple of bytes"
         self._shape = shape
--- a/numba/core/types/npytypes.py
+++ b/numba/core/types/npytypes.py
@@ -134,6 +134,8 @@ class Record(Type):
         name = 'Record({};{};{})'.format(desc, self.size, self.aligned)
         super(Record, self).__init__(name)
 
+        self.bitwidth = self.dtype.itemsize * 8
+
     @classmethod
     def _normalize_fields(cls, fields):
         """
@@ -211,10 +213,6 @@ class Record(Type):
         from numba.np.numpy_support import as_struct_dtype
 
         return as_struct_dtype(self)
-
-    @property
-    def bitwidth(self):
-        return self.dtype.itemsize * 8
 
     def can_convert_to(self, typingctx, other):
         """
@@ -435,9 +433,8 @@ class Array(Buffer):
             (isinstance(dtype, Record) and not dtype.aligned)):
             self.aligned = False
         if isinstance(dtype, NestedArray):
-            tmp = Array(dtype.dtype, dtype.ndim, 'C')
-            ndim += tmp.ndim
-            dtype = tmp.dtype
+            ndim += dtype.ndim
+            dtype = dtype.dtype
         if name is None:
             type_name = "array"
             if not self.mutable:
--- a/numba/tests/test_record_dtype.py
+++ b/numba/tests/test_record_dtype.py
@@ -1773,7 +1773,7 @@ class TestNestedArrays(TestCase):
             return x[0]
 
         arr = np.asarray([([(0,), (1,), (2,)],),
-                          ([(4,), (5,), (6,)],)],
+                          ([(3,), (4,), (5,)],)],
                          dtype=items)
 
         expected = fn.py_func(arr)
--- a/numba/tests/test_record_dtype.py
+++ b/numba/tests/test_record_dtype.py
@@ -1766,13 +1766,16 @@ class TestNestedArrays(TestCase):
     def test_issue_3158_1(self):
         # A nested array dtype.
         item = np.dtype([('some_field', np.int32)])
-        items = np.dtype([('items', item, 10)])
+        items = np.dtype([('items', item, 3)])
 
         @njit
         def fn(x):
             return x[0]
 
-        arr = np.zeros((2,), items)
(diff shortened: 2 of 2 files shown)


gmarkall:
@luk-f-a It seems like this doesn't break any tests - what do you think about the elegance (or lack thereof) of this patch?

gavinpotter:
Any idea when this might be implemented?  thanks.

gmarkall:
@gavinpotter Thanks for the prod - I've now finished up this patch, so it should be done whenever we can get it reviewed / merged. I've added it to the list to consider for today's dev meeting.

@luk-f-a Would you be happy to have a look over this patch please?

gavinpotter:
Many thanks

gavin

luk-f-a:
@gmarkall sure, I'll have a look

DylanModesitt:
FWIW, we have run into this issue and this patch seems to resolve it. Anything we can do to help this get merged? (need for additional tests, etc?)

luk-f-a:
thanks for the offer @DylanModesitt  I'll review this next week and after that we'll see what's needed.

numba/core/types/npytypes.py
```diff
@@ -430,6 +434,10 @@ def __init__(self, dtype, ndim, layout, readonly=False, name=None,
         if (not aligned or
             (isinstance(dtype, Record) and not dtype.aligned)):
             self.aligned = False
+        if isinstance(dtype, NestedArray):
+            tmp = Array(dtype.dtype, dtype.ndim, 'C')
```
luk-f-a, line 438: is `tmp` necessary? since the `Array` constructor does not transform the inputs it should be possible to do 
```python
            ndim += dtype.ndim
            dtype = dtype.dtype
```

luk-f-a (commented):
@gmarkall sorry for the 7 months delay, I reviewed and I think the PR is good, only two small comments/questions.

gmarkall:
@luk-f-a Many thanks for the suggestions - I've implemented them both.

gmarkall:
gpuci run tests

numba/tests/test_record_dtype.py
```diff
@@ -1732,6 +1732,68 @@ def copy(index, src, dest):
 
         np.testing.assert_array_equal(expected, got)
 
+    def test_issue_1469_1(self):
+        # Dimensions of nested arrays as a dtype are concatenated with the
+        # dimensions of the array.
+        nptype = np.dtype((np.float64,(4,)))
+        nbtype = types.Array(numpy_support.from_dtype(nptype),2,'C')
```
stuartarchibald, line 1739:
```suggestion
        nptype = np.dtype((np.float64, (4,)))
        nbtype = types.Array(numpy_support.from_dtype(nptype), 2, 'C')
```

numba/tests/test_record_dtype.py
```diff
@@ -1732,6 +1732,68 @@ def copy(index, src, dest):
...
+        nptype = np.dtype((np.dtype((np.float64, (5, 2))), (3, 6)))
+        nbtype = types.Array(numpy_support.from_dtype(nptype), 3, 'C')
+        expected = types.Array(types.float64, 7, 'C')
+        self.assertEqual(nbtype, expected)
+
+    def test_issue_1469_3(self):
+        # Nested arrays in record dtypes are accepted, but alignment is not
+        # guaranteed.
+        nptype = np.dtype([('a', np.float64,(4,))])
+        nbtype = types.Array(numpy_support.from_dtype(nptype),2,'C')
```
stuartarchibald, line 1756:
```suggestion
        nbtype = types.Array(numpy_support.from_dtype(nptype), 2, 'C')
```

numba/tests/test_record_dtype.py
```diff
@@ -1732,6 +1732,68 @@ def copy(index, src, dest):
...
+    def test_issue_3158_1(self):
+        # A nested array dtype.
+        item = np.dtype([('some_field', np.int32)])
+        items = np.dtype([('items', item, 10)])
+
+        @njit
+        def fn(x):
+            return x[0]
+
+        arr = np.zeros((2,), items)
```
stuartarchibald, line 1775: Should this use data with varying values? Were there a bug in e.g. the read implementation, if the entire memory region is zero it is less likely to be apparent.

numba/tests/test_record_dtype.py
```diff
@@ -1732,6 +1732,68 @@ def copy(index, src, dest):
...
+        # A slightly more complex nested array dtype example.
+        dtype1 = np.dtype([('a', 'i8'), ('b', 'i4')])
+        dtype2 = np.dtype((dtype1, (2, 2)))
+        dtype3 = np.dtype([('x', '?'), ('y', dtype2)])
+
+        @njit
+        def fn(arr):
+            return arr[0]
+
+        arr = np.zeros(2, dtype3)
```
stuartarchibald, line 1791: Same question RE data with varying values as above.

stuartarchibald (commented):
Thanks for the patch @gmarkall and thanks for checking/reviewing this @luk-f-a. This looks good in general, couple of minor queries to resolve then can be merged.

numba/tests/test_record_dtype.py
gmarkall: Yes - the test now uses varying values.

numba/tests/test_record_dtype.py
gmarkall: This test also uses varying values.

gmarkall:
@stuartarchibald Many thanks for the review - I've implemented all your suggestions, so assuming CI passes this should be ready for another review.

stuartarchibald (approved):

stuartarchibald:
> Any idea when this might be implemented? thanks.

@gavinpotter patch is approved, it'll be in Numba version 0.57.

PR #8121, opened 2022-05-31
Testing NumPy 1.23 RC 2 on public CI.

stuartarchibald:
As title. Do not merge.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -12,14 +12,14 @@ jobs:
     name: macOS
     vmImage: macOS-10.15
     matrix:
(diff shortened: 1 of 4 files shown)


stuartarchibald:
This CI run seems to indicate that NumPy 1.23rc2 doesn't have any major issues under Numba's test suite. Whilst this only tests a small subset of what the Numba buildfarm can, if there are more serious issues they do usually materialize in public CI. CC @seberg.

There's a few CUDA simulator test fails, but they don't look too bad, maybe a few minor semantic changes/error strings being altered. CC @gmarkall.

PR #8122, opened 2022-06-01
support register_jitable-ed function as njit function argument

### LINKED ISSUE #7043: register_jitable -ed function not recognized when used as arg to njit function
## Reporting a bug

register_jitable-ed function is not recognized when used as arg to njit-ed function.

Numba version: 0.53.1
Code:
```py
import numba as nb
import numba.extending

@nb.extending.register_jitable
def g(x):
    return 2*x

@nb.njit
def f(g,x):
    x+=1
    return g(x)

f(g,2)
```
```
TypingError                               Traceback (most recent call last)
<ipython-input-1-cbb9b70cf0f8> in <module>
     11     return g(x)
     12 
---> 13 f(g,2)

~/miniconda3/lib/python3.9/site-packages/numba/core/dispatcher.py in _compile_for_args(self, *args, **kws)
    418                 e.patch_message(msg)
    419 
--> 420             error_rewrite(e, 'typing')
    421         except errors.UnsupportedError as e:
    422             # Something unsupported is present in the user code, add help info

~/miniconda3/lib/python3.9/site-packages/numba/core/dispatcher.py in error_rewrite(e, issue_type)
    359                 raise e
    360             else:
--> 361                 raise e.with_traceback(None)
    362 
    363         argtypes = []

non-precise type pyobject
During: typing of argument at <ipython-input-1-cbb9b70cf0f8> (10)

File "<ipython-input-1-cbb9b70cf0f8>", line 10:
def f(g,x):
    x+=1
    ^

This error may have been caused by the following argument(s):
- argument 0: Cannot determine Numba type of <class 'function'>
```

Swaping register_jitable for njit works:

```py
import numba as nb
import numba.extending

# @nb.extending.register_jitable
@nb.njit
def g(x):
    return 2*x

@nb.njit
def f(g,x):
    x+=1
    return g(x)

f(g,2)
```
```
6
```

dlee992:
fix #7043 . 

- [x] typeof/typer OK: getting `global_type` via `typingctx` will avoid the first `TypingError`
- [x] unbox OK: add `unbox` for `types.Function`
- [x] CI pass
- [ ] because of holding lock on `typeof_pyval`, it results in much time comsumption in some CUDA test cases

### DIFF
--- a/numba/core/boxing.py
+++ b/numba/core/boxing.py
@@ -1063,6 +1063,11 @@ def box_pyobject(typ, val, c):
     return val
 
 
+@unbox(types.Function)
+def unbox_function(typ, obj, c):
+    return NativeValue(obj)
+
+
 def unbox_unsupported(typ, obj, c):
     c.pyapi.err_set_string("PyExc_TypeError",
                            "can't unbox {!r} type".format(typ))
--- a/numba/core/dispatcher.py
+++ b/numba/core/dispatcher.py
@@ -732,7 +732,8 @@ class _DispatcherBase(_dispatcher.Dispatcher):
         try:
             tp = typeof(val, Purpose.argument)
         except ValueError:
-            self.targetctx.refresh()
+            with global_compiler_lock:
+                self.targetctx.refresh()
             tp = self.typingctx._get_global_type(val)
             if tp is None:
                 tp = types.pyobject
--- a/numba/core/dispatcher.py
+++ b/numba/core/dispatcher.py
@@ -732,7 +732,10 @@ class _DispatcherBase(_dispatcher.Dispatcher):
         try:
             tp = typeof(val, Purpose.argument)
         except ValueError:
-            tp = types.pyobject
+            self.targetctx.refresh()
+            tp = self.typingctx._get_global_type(val)
+            if tp is None:
+                tp = types.pyobject
         else:
             if tp is None:
                 tp = types.pyobject
--- a/numba/core/typing/templates.py
+++ b/numba/core/typing/templates.py
@@ -1,7 +1,6 @@
 """
 Define typing templates
 """
-import threading
 from abc import ABC, abstractmethod
 import functools
 import sys
@@ -1303,12 +1302,10 @@ class BaseRegistryLoader(object):
         self._registrations = dict(
             (name, utils.stream_list(getattr(registry, name)))
             for name in self.registry_items)
-        self.lock = threading.lock()
 
     def new_registrations(self, name):
-        with self.lock:
-            for item in next(self._registrations[name]):
-                yield item
+        for item in next(self._registrations[name]):
+            yield item
 
 
 class RegistryLoader(BaseRegistryLoader):
--- a/numba/core/typing/templates.py
+++ b/numba/core/typing/templates.py
@@ -1,6 +1,7 @@
 """
 Define typing templates
 """
+
 from abc import ABC, abstractmethod
 import functools
 import sys
--- a/numba/core/typing/templates.py
+++ b/numba/core/typing/templates.py
@@ -1,7 +1,7 @@
 """
 Define typing templates
 """
-
+import threading
 from abc import ABC, abstractmethod
 import functools
 import sys
@@ -1303,10 +1303,12 @@ class BaseRegistryLoader(object):
         self._registrations = dict(
             (name, utils.stream_list(getattr(registry, name)))
             for name in self.registry_items)
+        self.lock = threading.lock()
 
     def new_registrations(self, name):
-        for item in next(self._registrations[name]):
-            yield item
+        with self.lock:
+            for item in next(self._registrations[name]):
+                yield item
 
 
 class RegistryLoader(BaseRegistryLoader):
--- a/numba/tests/test_extending.py
+++ b/numba/tests/test_extending.py
@@ -1551,6 +1551,19 @@ class TestRegisterJitable(unittest.TestCase):
         )
         self.assertIn(msg, str(raises.exception))
 
+    def test_as_njit_argument(self):
+        @register_jitable
+        def g():
+            return 1
+
+        @njit
+        def f(x):
+            return g(), x()
+
+        pv = f.py_func(g)
+        jv = f(g)
+        self.assertEqual(pv, jv)
+
 
 class TestImportCythonFunction(unittest.TestCase):
     @unittest.skipIf(sc is None, "Only run if SciPy >= 0.19 is installed")


gmarkall:
gpuci run tests

numba/core/dispatcher.py
```diff
@@ -732,7 +732,10 @@ def typeof_pyval(self, val):
         try:
             tp = typeof(val, Purpose.argument)
         except ValueError:
-            tp = types.pyobject
+            self.targetctx.refresh()
```
gmarkall, line 735: This causes the issue in the CUDA simulator, because refreshing the target context is not thread-safe. This isn't usually an issue because it's rarely or never (I'm not sure which :slightly_smiling_face:) refreshed during multithreaded execution.

I'm not sure whether avoiding refreshing from multiple threads, or making refreshing thread-safe is the way to go - perhaps @stuartarchibald or @sklam might have some thoughts about this?

gmarkall: Note that applying the following:

```diff
diff --git a/numba/core/dispatcher.py b/numba/core/dispatcher.py
index b93ca3481..40a9d1d90 100644
--- a/numba/core/dispatcher.py
+++ b/numba/core/dispatcher.py
@@ -732,7 +732,8 @@ class _DispatcherBase(_dispatcher.Dispatcher):
         try:
             tp = typeof(val, Purpose.argument)
         except ValueError:
-            self.targetctx.refresh()
+            with global_compiler_lock:
+                self.targetctx.refresh()
             tp = self.typingctx._get_global_type(val)
             if tp is None:
                 tp = types.pyobject
```

fixes the test, but has a huge overhead. Prior to this PR being applied:

```
$ NUMBA_ENABLE_CUDASIM=1 python runtests.py \
    numba.cuda.tests.cudapy.test_random.TestCudaRandomXoroshiro128p.test_uniform_float32

Ran 1 test in 8.774s

OK
```

With this PR and the above patch:

```
$ NUMBA_ENABLE_CUDASIM=1 python runtests.py \
    numba.cuda.tests.cudapy.test_random.TestCudaRandomXoroshiro128p.test_uniform_float32

Ran 1 test in 49.165s

OK
```

stuartarchibald:
Thanks for the patch @dlee992.

I think this extends the intended behaviour of public API function `@register_jitable`? As I understand it the intention was to support compilation when used "inside" another `jit` compiled function by direct reference, not as an argument:
https://github.com/numba/numba/blob/290a693641b85c327ab79e2430d60fe82666aa38/numba/core/extending.py#L140-L142

Whilst this change is solely additive in behaviour, I think this probably needs discussion to ensure it is a considered change and that it will not create problems elsewhere/in the future.

To open the discussion... what is a practical the use case for this? Is it for example a user wanting to use a function as an argument to both a python and a JIT compiled function, but in the case of the python function the argument function needs to run in pure python? Example:
```python
from numba import njit
from numba.extending import register_jitable
import numpy as np


@register_jitable
def func_arg(x):
    return x + 1


@njit
def jit_consumer(func, arg):
    return func(arg)


def python_consumer(func, arg):
    return func(arg)


arg = 10
print(jit_consumer(func_arg, arg))
print(python_consumer(func_arg, arg))
```

stuartarchibald:
> @stuartarchibald, I met this case because I developed SDC for Pandas `dataframe.apply(func, ...)`. I do need this feature for `register_jitable` in some situation like this.

Could you please expand on this perhaps with a motivating example to help explain why `@register_jitable` is useful to you in this case? Why not just use `njit`, wouldn't that be faster in the case of Panda's calls made through the Python interpreter and also work fine with calls through `JIT` compiled functions?

> I met this issue yesterday, so I searched in numba issues, and find #7043, so I opened this PR for this.

Thanks for checking the issue tracker first. I see that #7043 has no discussion in relation to the use cases/motivation either.

> Besides, if numba allows users to mark a python function as `register_jitable`, users may not know to avoid this function as a njit argument from python develop experience.

This adds some motivation, but arguably you can just turn the JIT off by setting the env var `NUMBA_DISABLE_JIT=1`, or use the `.py_func` attribute on any `@jit` family decorated function to achieve similar.

I suppose the most beneficial result of permitting this behaviour would be that it makes the notion of having a method of declaring that functions are `JIT` compatible more complete. I think it also makes it easier to have the entire call graph run using either all Python interpreted functions or all JIT functions.

Any thoughts?

-------------

Example of the use of `.py_func`:
```python
from numba import njit
from numba.extending import register_jitable
import numpy as np


@njit
def func_arg(x):
    return x + 1


@njit
def jit_consumer(func, arg):
    return func(arg)


def python_consumer(func, arg):
    return func(arg)


arg = 10
print(jit_consumer(func_arg, arg))
print(python_consumer(func_arg.py_func, arg))
```

dlee992:
> Could you please expand on this perhaps with a motivating example to help explain why `@register_jitable` is useful to you in this case? Why not just use `njit`, wouldn't that be faster in the case of Panda's calls made through the Python interpreter and also work fine with calls through `JIT` compiled functions?

I use `register_jitable` in a test case related with `df.apply(func, ...)`, so I can easily get the Pandas result and SDC result, and assert these.

> Thanks for checking the issue tracker first. I see that #7043 has no discussion in relation to the use cases/motivation either.

I saw the tag `feature request`, then I guess u guys maybe already discuss this issue, and think this is a feature on the todo list.

> This adds some motivation, but arguably you can just turn the JIT off by setting the env var `NUMBA_DISABLE_JIT=1`, or use the `.py_func` attribute on any `@jit` family decorated function to achieve similar.

Yeah, I know `.py_func` and use it a lot.

> I suppose the most beneficial result of permitting this behaviour would be that it makes the notion of having a method of declaring that functions are `JIT` compatible more complete. I think it also makes it easier to have the entire call graph run using either all Python interpreted functions or all JIT functions.

Yes! I also think these two reasons are most important/solid to support this behaviour.

numba/core/boxing.py
```diff
@@ -1063,6 +1063,11 @@ def box_pyobject(typ, val, c):
     return val
 
 
+@unbox(types.Function)
+def unbox_function(typ, obj, c):
+    return NativeValue(obj)
```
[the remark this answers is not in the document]
dlee992: Thanks for ur explaination!

numba/core/dispatcher.py
dlee992: @gmarkall, hi, thanks again. I compared the calling times of `BaseContext.refresh()` method, the test `test_uniform_float32` called 4319 times, while the new test `test_as_njit_argument` called 175 times. In fact, I never notice this method `refresh` could be called so many times. @stuartarchibald, @sklam any thought about this?

stuartarchibald:
> > I suppose the most beneficial result of permitting this behaviour would be that it makes the notion of having a method of declaring that functions are `JIT` compatible more complete. I think it also makes it easier to have the entire call graph run using either all Python interpreted functions or all JIT functions.
> 
> Yes! I also think these two reasons are most important/solid to support this behaviour.

I agree, this is a good motivation for getting this to work! Many thanks for your input on the discussion!

numba/core/dispatcher.py
dlee992: > I'd guess it's getting called so many times because the simulator uses a lot of threads, and they all hit the njit function at roughly the same time.

Emmm, it must be fixed or is acceptable?

sklam: I think it is acceptable if it is just the cudasim running it that many times and regular cuda execution is not greatly affected. @gmarkall, what's your thought?

sklam: @dlee992, what is driving the need for the `.refresh()`?
@gmarkall, should the lock happen earlier?

dlee992: > should the lock happen earlier?

@sklam, yes, I also think about moving the lock earlier, to avoid holding/releasing the lock too many times. But I haven't figured out where is the good place to move...

gmarkall:
Notes from discussion during the Numba dev meeting on 21 June 2022:

* The use of a lock around refreshing the typing context leads to performance degradation in the simulator due to many threads getting bottlenecked on the same call.
* In terms of design it feels preferable to move the typing context refresh to the point of registration - i.e. when `register_jitable()` is called.
* However, the reason for having the `refresh()` method is to improve performance - to lazily add to the typing context, rather than at every function decoration (which can have an impact on import time due to many decorations being at module top levels).
* The suggested route forward is still to try moving the refresh of the typing context out of the typeof function and towards the point of function registration, but it may take a lot of investigation / experimentation to find how the typing context can be refreshed without having a performance impact somewhere.
* Related issue: https://github.com/numba/numba/pull/7580

dlee992:
Yes, I think this PR and the related issue is almost related with #7507, #7516, and #7580. I will try to think this design again, to find another light-weight way to fix this.

dlee992:
I decide to give up this PR, since this issue is not bothering many users. If anyone is interested in fixing this issue, can go on with it.

PR #8126, opened 2022-06-01
Mark gufuncs as having mutable outputs

### LINKED ISSUE #6864: guvectorize returns different output after upgrading to Numpy 1.20

apmasell:
Indicate that functions generated with `guvectorize` should be allowed to
mutate their inputs so that NumPy will copy any mutations and prevent
uninitialized values from being exposed.

Closes #6864

### DIFF
--- a/numba/np/ufunc/_internal.c
+++ b/numba/np/ufunc/_internal.c
@@ -182,7 +182,7 @@ dufunc_init(PyDUFuncObject *self, PyObject *args, PyObject *kws)
     PyObject *dispatcher=NULL, *keepalive=NULL, *py_func_obj=NULL, *tmp;
     PyUFuncObject *ufunc=NULL;
     int identity=PyUFunc_None;
-    int nin=-1, nout=1;
+    int nin=-1, nout=1, idx;
     const char *name=NULL, *doc=NULL;
 
     static char * kwlist[] = {"dispatcher", "identity", "_keepalive", "nin",
@@ -229,6 +229,10 @@ dufunc_init(PyDUFuncObject *self, PyObject *args, PyObject *kws)
                                                      name, doc, 0);
     if (!ufunc) {
         return -1;
+    }
+
+    for (idx = 0; idx < nout; idx++) {
+        ufunc->op_flags[idx + nin] |= NPY_ITER_READWRITE | NPY_ITER_UPDATEIFCOPY | NPY_ITER_ALLOCATE;
     }
 
     /* Construct a keepalive list if none was given. */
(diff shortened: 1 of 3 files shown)


sklam (approved):
Thanks for the patch!
I have confirmed this on numpy1.22.

Issue #8127, opened 2022-06-02
No out of bounds check during advanced array indexing


kc611:
Numba ignores any indices that are out of bounds for a given array shape. This behavior is present in both basic and advanced indexing. 

```python 
import numpy as np
import numba

@numba.njit
def something():
    arr = np.ones((3,3))
    arr[100] = 0
    arr[:, np.array([2,3,5,100])] = 0
    return arr

print(something()) # Does not raise error, ignores the out of bound indices completely.
print(something.py_func()) # Raises out of bounds error
```

Not sure if we should be considering this as a bug or a feature :-) 

Should be easily fixed by adding a bounds check during indexing during function runtime.

esc:
@kc611 thank you for reporting this. IIRC there is a keyword argument that can be passed to the `njit` decorator:

https://numba.readthedocs.io/en/stable/reference/pysemantics.html?highlight=boundscheck#bounds-checking

Is this perhaps what you are looking for?

kc611:
Thank you that was exactly what I was looking for.

kc611:
Marking this as a good first issue for any newcomer.

PR #8129, opened 2022-06-03
FIXED :: No out of bounds check during advanced array indexing #8127

sanketny8:
No out of bounds check during advanced array indexing #8127

Added fix according to discussion.


sklam:
@sanketny8, Thanks for the patch. Can you include add a test using the reproducer at https://github.com/numba/numba/issues/8127#issuecomment-1144578419?

PR #8133, opened 2022-06-06
Fix #8132. Regression in Record.make_c_struct for handling nestedarray

### LINKED ISSUE #8132: Record not recognized as a data type
The Record.make_c_struct throws error in some special cases in the new releases of numba:

```
import numba
import numpy as np
from numba.types import Record, NestedArray, float64

print(numba.__version__)
print(np.__version__)

cell_shape_poly = np.array([[0,0],[1,1], [1,1],[1,0], [1,1]])

args_dtype = Record.make_c_struct([('cell_shape_poly', NestedArray(dtype=float64, shape=cell_shape_poly.shape)),])
print(args_dtype)
args=np.array((cell_shape_poly,), dtype=args_dtype)
```

The error:
```
0.55.2
1.22.4
Record(cell_shape_poly[type=nestedarray(float64, (5, 2));offset=0;alignment=8];72;True)
  File "D:\Projects\Python\tessellation\TorchSpatial\test.py", line 12, in <module>
    args=np.array((cell_shape_poly,), dtype=args_dtype)
TypeError: Cannot interpret 'Record(cell_shape_poly[type=nestedarray(float64, (5, 2));offset=0;alignment=8];72;True)' as a data type
```

Surprisingly the following code works correctly:
```
import numba
import numpy as np
from numba.types import Record, NestedArray, float64

print(numba.__version__)
print(np.__version__)

cell_shape_poly = np.array([[0,0],[1,1], [1,1],[1,0]])

args_dtype = Record.make_c_struct([('cell_shape_poly', NestedArray(dtype=float64, shape=cell_shape_poly.shape)),])
print(args_dtype)
args=np.array((cell_shape_poly,), dtype=args_dtype)
```

```
0.55.2
1.22.4
Record(cell_shape_poly[type=nestedarray(float64, (4, 2));offset=0;alignment=8];72;True)
```

All of the above codes work perfectly on numba 0.51.2 and numpy 1.21.6, however numba 0.51.2 is not compatible with python 3.10.4 that I'm currently using.

sklam:
Fix #8132.

### DIFF
--- a/numba/core/datamodel/models.py
+++ b/numba/core/datamodel/models.py
@@ -899,6 +899,13 @@ class NestedArrayModel(ArrayModel):
         self._be_type = dmm.lookup(fe_type.dtype).get_data_type()
         super(NestedArrayModel, self).__init__(dmm, fe_type)
 
+    def as_storage_type(self):
+        """Return the LLVM type representation for the storage of
+        the nestedarray.
+        """
+        ret = ir.ArrayType(self._be_type, self._fe_type.nitems)
+        return ret
+
 
 @register_default(types.Optional)
 class OptionalModel(StructModel):
--- a/numba/core/types/npytypes.py
+++ b/numba/core/types/npytypes.py
@@ -94,7 +94,10 @@ class Record(Type):
             if not isinstance(ty, (Number, NestedArray)):
                 msg = "Only Number and NestedArray types are supported, found: {}. "
                 raise TypeError(msg.format(ty))
-            datatype = ctx.get_data_type(ty)
+            if isinstance(ty, NestedArray):
+                datatype = ctx.data_model_manager[ty].as_storage_type()
+            else:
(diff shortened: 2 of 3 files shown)


numba/core/datamodel/models.py
```diff
@@ -899,6 +899,13 @@ def __init__(self, dmm, fe_type):
         self._be_type = dmm.lookup(fe_type.dtype).get_data_type()
         super(NestedArrayModel, self).__init__(dmm, fe_type)
 
+    def as_storage_type(self):
+        """Return the LLVM type representation for the storage of
+        the nestedarray.
+        """
+        ret = ir.ArrayType(self._be_type, self._fe_type.nitems)
```
stuartarchibald, line 906: NOTE: I think this is ok as the type system only allows nested arrays of simple numerical types.

stuartarchibald (approved):
Thanks for the patch.

PR #8136, opened 2022-06-07
Fix some C++ 11 Issues

njriasan:
I was experimenting with the AOT module inside a Python package that sets requires C++ 11. In doing so, I got a variety of errors related to C code that fails to conform to the C++ 11 standard. These are some of the changes that seemed straight forward because they just require simple casts.

### DIFF
--- a/numba/cext/listobject.c
+++ b/numba/cext/listobject.c
@@ -164,7 +164,7 @@ numba_list_new(NB_List **out, Py_ssize_t item_size, Py_ssize_t allocated){
     NB_List *lp;
     char *items;
     // allocate memory to hold the struct
-    lp = malloc(aligned_size(sizeof(NB_List)));
+    lp = (NB_List *) malloc(aligned_size(sizeof(NB_List)));
     if (lp == NULL) {
         return LIST_ERR_NO_MEMORY;
     }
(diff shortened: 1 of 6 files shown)


numba/cext/listobject.c
```diff
@@ -623,7 +623,7 @@ numba_test_list(void) {
     Py_ssize_t it_count;
     const char *it_item = NULL;
     NB_ListIter iter;
-    char got_item[4] = "\x00\x00\x00\x00";
+    char got_item[4] = {'\x00', '\x00', '\x00', '\x00'};
```
njriasan, line 626: This issue arose because C++ 11 complains about a size mismatch between the string and the array. Since the string had a null terminator I believe its length was actually 5.

stuartarchibald:
Thanks for the patch. It's probably going to need a rebase on https://github.com/numba/numba/pull/8124 which is scheduled for 0.56, will hold off review until that makes it to mainline.

njriasan:
> Thanks for the patch. It's probably going to need a rebase on #8124 which is scheduled for 0.56, will hold off review until that makes it to mainline.

My apologies for the force push as I needed that to do a proper rebase. Please let me know if there was a better/preferred way to do this.

This PR should now be ready for review.

stuartarchibald:
> > Thanks for the patch. It's probably going to need a rebase on #8124 which is scheduled for 0.56, will hold off review until that makes it to mainline.
> 
> My apologies for the force push as I needed that to do a proper rebase. Please let me know if there was a better/preferred way to do this.

No problem, in general, until review has taken place, it's ok to rebase. However, once a review has been done, if needed, it's typical to merge changes in from `main` otherwise github struggles to track comments and it adds work for the reviewer(s) as entire patches have to be re-checked.

On which note, this patch currently has conflicts, please could you resolve them so it can be scheduled for review? Many thanks!

PR #8137, opened 2022-06-07
CUDA: Fix #7806, Division by zero stops the kernel

### LINKED ISSUE #7806: CUDA: Division by zero stops the kernel
The following:

```python
from numba import cuda
import numpy as np

r = np.zeros(1)
x = np.zeros(1)
y = np.ones(1)


@cuda.jit
def f(r, x, y):
    r[0] = y[0] / x[0]
    print("Here")


f[1, 1](r, x, y)
cuda.synchronize()

print(r)
```

prints:

```
[0.]
```

instead of the expected:

```
Here
[inf]
```

gmarkall:
The expected behaviour is that execution proceeds after division by zero, as the CUDA target doesn't normally have a way to signal the exception. When debug is enabled, it can signal the exception so an exception should be raised and execution should stop. This is implemented by setting CUDA kernels to use the NumPy error model in general, and only switching back to the Python error model when debug is on.

This PR also adds support for raising on divide by zero when debug is on in the simulator - this helps align the behaviour of the simulator with that of the real device. The error behaviour in NumPy is configured on a per-thread basis, so it's essential to use `np.seterr()` when the target thread starts running. Since threads are ephemeral for the life of the kernel launch, no attempt is made to save and restore the initial state of the error behaviour.

Finally, there were some illegal uses of debug with optimization in the `test_exception.py` file I was working on, so I fixed that while I was working in this area.

Fixes #7806.

### DIFF
--- a/docs/source/cuda/cudapysupported.rst
+++ b/docs/source/cuda/cudapysupported.rst
@@ -26,9 +26,14 @@ For details please consult the
 Floating Point Error Model
 --------------------------
 
-By default, CUDA Python kernels execute with the NumPy error model. In this model, division by zero raises no exception and instead produces a result of ``inf`` / ``-inf`` or ``nan``. This differs from the normal Python error model, in which divisions by zero raise a ``ZeroDivisionError``.
+By default, CUDA Python kernels execute with the NumPy error model. In this
+model, division by zero raises no exception and instead produces a result of
+``inf``, ``-inf`` or ``nan``. This differs from the normal Python error model,
+in which division by zero raises a ``ZeroDivisionError``.
 
-When debug is enabled (by passing ``debug=True`` to the :func:`@cuda.jit <numba.cuda.jit>` decorator), the Python error model is used. This allows division-by-zero errors during kernel execution to be identified.
+When debug is enabled (by passing ``debug=True`` to the
+:func:`@cuda.jit <numba.cuda.jit>` decorator), the Python error model is used.
+This allows division-by-zero errors during kernel execution to be identified.
 
 Constructs
 ----------
--- a/docs/source/cuda/cudapysupported.rst
+++ b/docs/source/cuda/cudapysupported.rst
@@ -23,6 +23,13 @@ For details please consult the
 `CUDA Programming Guide
 <http://docs.nvidia.com/cuda/cuda-c-programming-guide/#programming-model>`_.
 
+Floating Point Error Model
+--------------------------
+
+By default, CUDA Python kernels execute with the NumPy error model. In this model, division by zero raises no exception and instead produces a result of ``inf`` / ``-inf`` or ``nan``. This differs from the normal Python error model, in which divisions by zero raise a ``ZeroDivisionError``.
+
+When debug is enabled (by passing ``debug=True`` to the :func:`@cuda.jit <numba.cuda.jit>` decorator), the Python error model is used. This allows division-by-zero errors during kernel execution to be identified.
+
 Constructs
 ----------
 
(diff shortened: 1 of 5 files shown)


gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

numba/cuda/tests/cudapy/test_exception.py
```diff
@@ -91,6 +93,56 @@ def test_raise_causing_warp_diverge(self):
         """
         self.case_raise_causing_warp_diverge(with_debug_mode=False)
 
+    # The following two cases relate to Issue #7806: Division by zero stops the
+    # kernel. https://github.com/numba/numba/issues/7806.
+
+    def test_no_zero_division_error(self):
+        # When debug is False:
+        # - Zero by division raises no exception
```
stuartarchibald, line 101:
```suggestion
        # - Division by zero raises no exception
```

stuartarchibald (commented):
Thanks for the patch, I've suggested minor wording change else looks good. I have two additional questions:
1. Is there ever a case where it's useful to have debug enabled and also for divide by zero to not raise?
2. Should this be documented, it seems like it's quite important behaviour?

Thanks again!

numba/cuda/tests/cudapy/test_exception.py
gmarkall: Thanks, not quite sure how I managed to write that!

gmarkall:
> Is there ever a case where it's useful to have debug enabled and also for divide by zero to not raise?

Yes, I think this would be a useful enhancement for the simulator. For example, you might be debugging an issue not related to division by zero, with the expectation that your code does divide by zero in places and that it's not a problem. I'll open an issue to track this.

> Should this be documented, it seems like it's quite important behaviour?

I'll add some notes about this to the documentation.

gmarkall:
@stuartarchibald I've added some documentation for the error model in CUDA and opened an issue for requesting support for more flexibility in the CUDA error model (https://github.com/numba/numba/issues/8155) - this should be ready for another look now.

docs/source/cuda/cudapysupported.rst
```diff
@@ -23,6 +23,13 @@ For details please consult the
 `CUDA Programming Guide
 <http://docs.nvidia.com/cuda/cuda-c-programming-guide/#programming-model>`_.
 
+Floating Point Error Model
+--------------------------
+
+By default, CUDA Python kernels execute with the NumPy error model. In this model, division by zero raises no exception and instead produces a result of ``inf`` / ``-inf`` or ``nan``. This differs from the normal Python error model, in which divisions by zero raise a ``ZeroDivisionError``.
+
+When debug is enabled (by passing ``debug=True`` to the :func:`@cuda.jit <numba.cuda.jit>` decorator), the Python error model is used. This allows division-by-zero errors during kernel execution to be identified.
```
stuartarchibald, line 31:
```suggestion
By default, CUDA Python kernels execute with the NumPy error model. In this
model, division by zero raises no exception and instead produces a result of
``inf``, ``-inf`` or ``nan``. This differs from the normal Python error model,
in which division by zero raises a ``ZeroDivisionError``.

When debug is enabled (by passing ``debug=True`` to the
:func:`@cuda.jit <numba.cuda.jit>` decorator), the Python error model is used.
This allows division-by-zero errors during kernel execution to be identified.
```

This line wraps and also alters the `inf / -inf` to `inf, -inf` and changes `in which divisions by a zero raise a` to `in which division by zero raises a`.

stuartarchibald (commented):
Thanks for the update, a minor change (mostly line wrap) is noted inline else looks good.

gmarkall:
gpuci run tests

stuartarchibald (dismissed):

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_134`.

Passed.

sklam:
@gmarkall, there's now a merge conflict

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

PR #8139, opened 2022-06-07
Stencil neighborhood option doesn't support numpy.integer

### LINKED ISSUE #6901: Stencil neighborhood option doesn't support numpy.int64.
In numba/stencils/stencil.py, there are various places (like line 552, "if isinstance(kernel_size[i][0], int):") where we check for "int" in relation to neighborhoods.  I ran across a case where I was creating a neighborhood tuple by extracting values from a Numpy array.    This causes a problem because those Numpy values will not match in these isinstance int checks.  I worked around it by converting to int during neighborhood tuple creation but perhaps a more flexible check in Numba would eliminate this conversion.

andrijdavid:
This PR fixes #6901. It adds a fix in `stencil.py` and a new test case for the stencil operator in `test_stencil.py`.

### DIFF
--- a/numba/stencils/stencil.py
+++ b/numba/stencils/stencil.py
@@ -84,7 +84,8 @@ class StencilFunc(object):
         self._targetctx.refresh()
         self._install_type(self._typingctx)
         neighborhood = self.options.get("neighborhood")
-        self.neighborhood = tuple([tuple(map(int, n)) for n in neighborhood]) if neighborhood else None
+        _cast_numpy_to_int = lambda x: int(x) if isinstance(x, np.integer) else x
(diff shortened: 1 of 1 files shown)


stuartarchibald:
@andrijdavid Thanks for the patch, I've queued it for review.

stuartarchibald:
CC @DrTodd13, might be worth trying to get https://github.com/numba/numba/pull/6490 in ahead of this?

andrijdavid:
I noticed that there hasn't been any activity on this PR for a while now, and it's currently stale. Any review about this after #6490 was merged

stuartarchibald:
@DrTodd13 now #6490 is merged, please could you give this a review? Many thanks.

PR #7878, opened 2022-03-02
CUDA: Remove some deprecated support, add CC 8.6 and 8.7

gmarkall:
This removes deprecated and various un-needed parts of the CUDA target, and adds support for Compute Capabilities 8.6 and 8.7 (dependent upon the toolkit version). The changes include:

- Remove support for CUDA toolkits 9.2 - 10.1.
- Test on gpuCI with 10.2, 11.1, 11.2, and 11.5:
   - 10.2: The oldest-supported, and only 10.x version supported
   - 11.1: The last version using NVVM 3.4
   - 11.2: The first version using NVVM 7.0
   - 11.5: The newest version supported on gpuCI (used for CUDA Python)
- Update the documentation to mention that CC 5.3 and CUDA Toolkit 10.2 are the minimum required.
   - Also update the supported device list to include newer devices like AGX Orin and H100.
   - Deprecation notice updated accordingly.
- Remove the 32-bit data layout, which has been unsupported for years.
- Remove vestigial parts of OSX support from the CUDA target (e.g. in `cuda_paths.py`).
- Update the libdevice search to not check for multiple virtual architectures (`compute_XX` etc) - these have not been used since CUDA 9.0.
- Remove asyncio support for Python < 3.7 (used in support for async operations on streams).
- Normalise the presentation of version numbers in `numba -s`:
   - Before: `(11, 6)` / `11050` for driver 11.6 with runtime 11.5
   - After: `11.6` / `11.5` for the same versions
- Always implement `math.remainder`, as it was only elided for Python < 3.7.
-  Bump the simulator reported CC to 5.3.
- Remove test skips for now-unsupported compute capabilities (<3.2, <5.0, and <5.3).

PR #8142, opened 2022-06-08
CUDA: Fix some missed changes from dropping 9.2

gmarkall:
When toolkit 9.2-10.1 support was dropped in #7878, a couple of bits got missed.

- Most importantly, the Conda recipe still required only 9.2 - it should be 10.2.
- The missing function error message stated that it required CUDA 9.2. Since all APIs we use are available in all supported toolkit versions, the error message need no longer state a CUDA requirement.
- All supported toolkit versions implement `nvvmLazyAddModuleToProgram`, so we no longer need a workaround for it.

### DIFF
--- a/buildscripts/condarecipe.local/meta.yaml
+++ b/buildscripts/condarecipe.local/meta.yaml
@@ -61,8 +61,8 @@ requirements:
     # Exclude 0.3.20 too
     # https://github.com/numba/numba/issues/8096
     - libopenblas >=0.3.18, !=0.3.20     # [arm64]
-    # CUDA 9.2 or later is required for CUDA support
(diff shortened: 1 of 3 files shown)


gmarkall:
gpuci run tests

stuartarchibald (approved):
Thanks for the patch, looks good on inspection (have also grepped for other 9.2 references and there don't appear to be any that are relevant).

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_133`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_133`.

Passed.

PR #8144, opened 2022-06-10
Fix NumPy capitalisation in docs.

stuartarchibald:
As title. NumPy is written as "NumPy".


esc (changes requested):
I used `git grep` in the `docs` subdirectory and only found the following, in a source code comment:

https://github.com/numba/numba/blob/main/docs/source/user/cfunc.rst?plain=1#L175

esc:
Thank you for the patch!

esc (approved):

PR #8145, opened 2022-06-10
Allow ufunc builder to use previously JITed function

### LINKED ISSUE #7409: `numba.vectorize` interface inconsistency between `target="cpu"` and `target="parallel"`
`numba.vectorize` on 4911192c5 (and further back) appears to have the following interface inconsistency:
```python
import numpy as np
import numba as nb


@nb.jit
def add(x, y):
    return x + y


custom_vectorize = nb.vectorize([], identity=None, target="cpu")

# No problem here
test_fn = custom_vectorize(add)

custom_vectorize_2 = nb.vectorize([], identity=None, target="parallel")

# Problem here
test_fn_2 = custom_vectorize_2(add)

# TypeError: A jit decorator was called on an already jitted function CPUDispatcher(<function add at 0x7fa399ca2ef0>).  If trying to access the original python function, use the CPUDispatcher(<function add at 0x7fa399ca2ef0>).py_func attribute.
```

apmasell:
This permits the ufunc builder to look inside a dispatcher and JIT the original
Python function so that `vectorize` can be applied to a JITed function when in
non-CPU mode (CPU alread works).

Closes #7409

### DIFF
--- a/numba/np/ufunc/dufunc.py
+++ b/numba/np/ufunc/dufunc.py
@@ -1,9 +1,9 @@
 from numba import jit, typeof
 from numba.core import cgutils, types, serialize, sigutils
+from numba.core.extending import is_jitted
 from numba.core.typing import npydecl
 from numba.core.typing.templates import AbstractTemplate, signature
 from numba.np.ufunc import _internal
-from numba.core.dispatcher import Dispatcher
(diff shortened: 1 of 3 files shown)


sklam (approved):
LGTM

PR #8158, opened 2022-06-14
adding -pthread for linux-ppc64le in setup.py

### LINKED ISSUE #8157: -pthread link flag missing in setup.py for linux-ppc64le
As part of https://github.com/numba/numba/pull/8052 the `-pthread` was injected for the linux-ppc64le platform. However, this was only done for the conda recipe and as setting for AOT complation. What is missing is adding this injection into `setup.py` so that the regular build process succeeds on that platform too. Here is a reproducer:

```
(debug_gamma) ci@atx-power8-01:~/ci/numba$ python setup.py build_ext -i && python setup.py develop --no-deps
...
(debug_gamma) ci@atx-power8-01:~/ci/numba$ python -m numba.runtests -m 4 numba.tests.test_np_randomgen
  File "/home/ci/miniconda3/envs/debug_gamma/lib/python3.9/runpy.py", line 188, in _run_module_as_main
    mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
  File "/home/ci/miniconda3/envs/debug_gamma/lib/python3.9/runpy.py", line 111, in _get_module_details
    __import__(pkg_name)
  File "/home/ci/ci/numba/numba/__init__.py", line 38, in <module>
    from numba.core.decorators import (cfunc, generated_jit, jit, njit, stencil,
  File "/home/ci/ci/numba/numba/core/decorators.py", line 12, in <module>
    from numba.stencils.stencil import stencil
  File "/home/ci/ci/numba/numba/stencils/stencil.py", line 11, in <module>
    from numba.core import types, typing, utils, ir, config, ir_utils, registry
  File "/home/ci/ci/numba/numba/core/ir_utils.py", line 13, in <module>
    from numba.core.extending import _Intrinsic
  File "/home/ci/ci/numba/numba/core/extending.py", line 19, in <module>
    from numba.core.pythonapi import box, unbox, reflect, Nativeciue  # noqa: F401
  File "/home/ci/ci/numba/numba/core/pythonapi.py", line 11, in <module>
    from numba import _helperlib
ImportError: /home/ci/ci/numba/numba/_helperlib.cpython-39-powerpc64le-linux-gnu.so: undefined symbol: pthread_atfork
```

## Reporting a bug

esc:
Fixes #8157


setup.py
```diff
@@ -323,7 +325,7 @@ def check_file_at_path(path2file):
     ext_np_ufunc_backends.append(ext_np_ufunc_workqueue_backend)
 
     ext_mviewbuf = Extension(name='numba.mviewbuf',
-                             extra_link_args=install_name_tool_fixer,
+                             extra_link_args=extra_link_args,
```
stuartarchibald, line 328: Does this library need pthreads?

setup.py
```diff
@@ -323,7 +325,7 @@ def check_file_at_path(path2file):
     ext_np_ufunc_backends.append(ext_np_ufunc_workqueue_backend)
```
stuartarchibald, line 325: The `workqueue` threading layer definitely needs pthreads?

stuartarchibald (commented):
Thanks for the patch and for fixing this. Few comments inline, this also needs a buildfarm run and a manual test on a ppc64le machine, particularly against the threading layers.

setup.py
```diff
@@ -66,9 +66,11 @@ def run(self):
 cmdclass = versioneer.get_cmdclass()
 cmdclass['build_doc'] = build_doc
 
-install_name_tool_fixer = []
+extra_link_args = []
 if sys.platform == 'darwin':
-    install_name_tool_fixer += ['-headerpad_max_install_names']
+    extra_link_args += ['-headerpad_max_install_names']
+elif platform.machine() == 'ppc64le':
+    extra_link_args = ['-pthread']
```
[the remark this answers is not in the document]
esc: fixed

esc:
@stuartarchibald I have fixed up the items from the review and tested the patch on a `ppc64le` machine. The Numba test suite now passes on that platform w/o linking errors.

stuartarchibald:
> @stuartarchibald I have fixed up the items from the review and tested the patch on a `ppc64le` machine. The Numba test suite now passes on that platform w/o linking errors.

Thanks @esc, does this need to be in the 0.56 milestone?

setup.py
```diff
@@ -278,6 +282,7 @@ def check_file_at_path(path2file):
                 depends=['numba/np/ufunc/workqueue.h'],
                 include_dirs=[os.path.join(tbb_root, 'include')],
                 extra_compile_args=cpp11flags,
+                extra_link_args=extra_link_args,
```
stuartarchibald, line 285: Why is this needed for TBB pool but not OpenMP one? Am guessing the `pthread_atfork` call in TBB and no `pthread.h` API use in OpenMP (even though the header file is `#include`d https://github.com/numba/numba/blob/9a7820cb5f21776587d23e7facf3b95fcbe6097d/numba/np/ufunc/omppool.cpp#L21 probably from adding debug code during development, e.g. calls to `pthread_self()`).

setup.py
```diff
@@ -172,7 +175,8 @@ def get_ext_modules():
                                        "numba/cext/dictobject.c",
                                        "numba/cext/listobject.c",
                                        ],
-                              extra_link_args=install_name_tool_fixer,
+                              extra_link_args=install_name_tool_fixer +
+                              extra_link_args,
```
stuartarchibald, line 179: NOTE: it's `_random.c` that needs `pthread` as it uses the fork handler.

stuartarchibald (commented):
Thanks for the update, this looks about right and the proof is in the build on the PPC64LE architecture. There's a question inline but otherwise this seems ok. Thanks again!

setup.py
esc: Yes, the OpenMP extension does not appear to need the linker flags.

setup.py
esc: Should that be included as a comment?

setup.py
esc: no, it appears as though it doesn't.

setup.py
stuartarchibald: Thanks for checking. I suggest removing the unused header as well as the use of the link time flags.

setup.py
stuartarchibald: Yes, I think that's a good idea, saves searching next time. Thanks!

esc: done in dd33b3da3caac8d511aa5ceb20485e88e85d2c90

setup.py
esc: I have removed the header in 5646137a2d8630d92af8426c931025dec47ad804

stuartarchibald (approved):

esc:
@stuartarchibald thank you for the review. I double checked b17fe1a manually again on a power8 machine. I was able to compile Numba and the test suite ran to completion w/o issues.

stuartarchibald:
> @stuartarchibald thank you for the review. I double checked [b17fe1a](https://github.com/numba/numba/commit/b17fe1a4d74ebb55c3d698e67b2b54d0529d4f8e) manually again on a power8 machine. I was able to compile Numba and the test suite ran to completion w/o issues.

Great, thanks for checking again @esc.

PR #8163, opened 2022-06-15
CUDA: Remove context query in launch config

### LINKED ISSUE #7926: CUDA: remove context query in launch configuration
There's a call made to query the device context every time a kernel is configured for launch as `config.CUDA_LOW_OCCUPANCY_WARNINGS` is on by default.
https://github.com/numba/numba/blob/fa3630edab447f38f81c97cbd943809ebe7a9840/numba/cuda/dispatcher.py#L453-L462

from OOB discussion with @gmarkall, this task is to explore removing this as it's a debug/help message that impacts the performance of the most common use case. Replacement would be to instead just produce the message for suspiciously low grid sizes (based on some arbitrary value) without the context query. Another alternative is to have a `config.CUDA_PRODUCTION` flag to disable the check (and other future similar ones) in production.

gmarkall:
This swaps the context query in the launch configuration for a statically-chosen threshold of 128 blocks, removing a source of overhead in kernel configuration.

This also fixes a failure in `TestWarnings.test_efficient_launch_configuration`, which fails with:

```
FAIL: test_efficient_launch_configuration (numba.cuda.tests.cudapy.test_warning.TestWarnings)
  File "/home/swqa/NUMBA/envs/NumbaGPUTest/lib/python3.9/site-packages/numba/cuda/tests/cudapy/test_warning.py", line 33, in test_efficient_launch_configuration
    self.assertEqual(len(w), 0)
AssertionError: 1 != 0
```

when run on a device with 132 SMs - this is because the test launched 256 blocks, but the check for emitting the warning requires 2 * (SM count), which is 268 (the test was written assuming such a large GPU did not exist).

Fixes #7926.

### DIFF
--- a/numba/cuda/dispatcher.py
+++ b/numba/cuda/dispatcher.py
@@ -459,14 +459,19 @@ class _LaunchConfiguration:
         self.sharedmem = sharedmem
 
         if config.CUDA_LOW_OCCUPANCY_WARNINGS:
-            ctx = get_context()
-            smcount = ctx.device.MULTIPROCESSOR_COUNT
+            # Warn when the grid has fewer than 128 blocks. This number is
+            # chosen somewhat heuristically - ideally the minimum is 2 times
+            # the number of SMs, but the number of SMs varies between devices -
+            # some very small GPUs might only have 4 SMs, but an H100-SXM5 has
+            # 132. In general kernels should be launched with large grids
+            # (hundreds or thousands of blocks), so warning when fewer than 128
+            # blocks are used will likely catch most beginner errors, where the
+            # grid tends to be very small (single-digit or low tens of blocks).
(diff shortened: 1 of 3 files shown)


gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

stuartarchibald (approved):
Thanks for the patch. As discussed OOB a while back, completely agree with this change as:
* it's diagnostic/offering a suggestion only
* querying the context on every kernel launch adds significantly to the launch overhead

Thanks again!

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_137`.

sklam:
build farm passed

PR #8164, opened 2022-06-15
remove myself from automatic reviewer assignment

esc:
With the new Burndown model and the weekly assignment of reviewers, I don't want to be automatically assigned as a reviewer for every incoming PR. I'd rather be assigned manually to only those PRs that I will actually review.


stuartarchibald (approved):
Thanks for the patch. For discussion, see Numba public meeting notes from 28th June 2022 in the wiki (once published).

PR #8165, opened 2022-06-15
Restrict strace based tests to be linux only via support feature.

stuartarchibald:
As title, also updates the command line used for strace to use the
binary found by `which`.


esc (approved):
I tested this on the Windows machines of the Anaconda Build Farm. It works as expected and also fixes an SVML timeout issue I have been seeing!

PR #8167, opened 2022-06-16
CUDA: Facilitate and document passing arrays / pointers to foreign functions

gmarkall:
This PR fills in a gap in the CUDA FFI functionality, making it easier to pass arrays without hacks to get at the data pointer, and adding relevant documentation. The main change is to enable CFFI support in the CUDA target - although this seems like a broad base of support to add, in effect adding CFFI typing and lowering only adds implementations for the `cffi.FFI.from_buffer()` method, which is exactly what is needed (and makes sense to use in kernels to get the pointer).

Some peripheral changes were also needed:

- The data for linking tests was moved so it is side by side with the CUDA driver and CUDA Python tests, since it is now needed by both.
- The `jitlink.ptx` file was updated and regenerated for the current oldest-supported device and toolkit (CC 5.3 with toolkit 10.2).
- A small fix to a prototype declaration in the CUDA FFI docs.


gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

docs/source/cuda/cuda_ffi.rst
```diff
@@ -87,6 +87,51 @@ For example, when:
...
+
+
+would be declared as follows:
+
+.. code::
+
+   signature = 'float32(CPointer(float32), int32)'
+   sum_reduce = cuda.declare_device('sum_reduce', signature)
+
+To obtain a pointer to array data for passing to C functions, use the
```
stuartarchibald, line 120:
```suggestion
To obtain a pointer to array data for passing to foreign functions, use the
```
(and line wrap), I guess it doesn't really matter what the source language of the function is?

docs/source/cuda/cuda_ffi.rst
```diff
@@ -87,6 +87,51 @@ For example, when:
...
+
+.. code::
+
+   import cffi
+   ffi = cffi.FFI()
+
+   @cuda.jit
+   def reduction_caller(result, array):
+       array_ptr = ffi.from_buffer(array)
+       result[0] = sum_reduce(array, len(array))
```
stuartarchibald, line 132:
```suggestion
       array_ptr = ffi.from_buffer(array)
       result[0] = sum_reduce(array_ptr, len(array))
```
doesn't this need to be the data pointer? Also... should this all be turned into a doc-test style test such that the code is executed and asserted to be correct?

numba/cuda/tests/cudadrv/test_linker.py
```diff
@@ -104,6 +104,10 @@ def simple_lmem(A, B, dty):
         B[i] = C[i]
 
 
+parent_dir = os.path.dirname(os.path.dirname(__file__))
+data_dir = os.path.join(parent_dir, 'data')
```
stuartarchibald, line 108: Given this is "spelled" to be at a fixed location in `setup.py` (`numba.cuda.tests.data`), should there be a variable in e.g. `numba.cuda.tests.support` called something like `TEST_DATA_PATH` which is set to this path and then that used throughout the CUDA tests?

numba/cuda/tests/cudadrv/test_linker.py
```diff
@@ -104,6 +104,10 @@ def simple_lmem(A, B, dty):
         B[i] = C[i]
 
 
+parent_dir = os.path.dirname(os.path.dirname(__file__))
```
stuartarchibald, line 107: Maybe use `pathlib.Path` here and the `.parent` property instead?

numba/cuda/tests/cudapy/test_cffi.py
```diff
@@ -0,0 +1,34 @@
...
+
+from numba import cuda, types
+from numba.cuda.testing import skip_on_cudasim, unittest, CUDATestCase
+
+parent_dir = os.path.dirname(os.path.dirname(__file__))
+data_dir = os.path.join(parent_dir, 'data')
+
+
+@skip_on_cudasim('Simulator does not support linking')
+class TestCFFI(CUDATestCase):
```
stuartarchibald, line 13: Does this also need to be skipped on CC=(5,2) devices?

numba/cuda/tests/data/jitlink.cu
```diff
@@ -0,0 +1,20 @@
+// Compile with:
+//
+//   nvcc -gencode arch=compute_53,code=compute_53 -rdc true -ptx jitlink.cu
+//
+// using the oldest supported toolkit version (10.2 at the time of writing).
+
```
stuartarchibald, line 6: Suggest adding a note in here about the necessity of the `out` argument due to CUDAs minimal CC.

stuartarchibald (commented):
Thanks for implementing this feature @gmarkall, great to see the interoperability between Numba compiled code and functions from external sources expanded. I've given this a review, there's a few queries to take a look at, once resolved should be good to merge. Thanks again!

docs/source/cuda/cuda_ffi.rst
gmarkall: Good point!

numba/cuda/tests/cudadrv/test_linker.py
gmarkall: That would be better, yes.

numba/cuda/tests/cudadrv/test_linker.py
gmarkall: Yeah, that would be better,

numba/cuda/tests/data/jitlink.cu
gmarkall: Will do.

numba/cuda/tests/cudapy/test_cffi.py
gmarkall: I don't think so - the difference between 5.2 and 5.3 is support for `float16`. Could I be missing something here?

stuartarchibald: From OOB conversation. The issue is that the baseline CC is 5,3 and whilst the PTX can be re-targetted for higher versions it cannot be for lower. Suggested fixes are either skip or use 5,0 as the baseline, with the latter being chosen.

numba/cuda/tests/cudadrv/test_linker.py
gmarkall: Now used in the declaration of `test_data_dir` in `numba.tests.support`.

numba/cuda/tests/cudadrv/test_linker.py
gmarkall: Now added `test_data_path` to `numba.cuda.testing`, which is where lots of utility bits for CUDA tests have been put.

docs/source/cuda/cuda_ffi.rst
gmarkall: Now changed.

docs/source/cuda/cuda_ffi.rst
gmarkall: Doctest now created, demonstrating that your suggested code change is correct.

numba/cuda/tests/data/jitlink.cu
gmarkall: Done.

gmarkall:
@stuartarchibald Many thanks for the review - I have attempted to address all comments, so this should be ready for another look pending a successful gpuCI run.

gmarkall:
gpuci run tests

numba/cuda/tests/doc_examples/test_ffi.py
```diff
@@ -45,6 +45,38 @@ def multiply_vectors(r, x, y):
         np.testing.assert_array_equal(r, x * y)
         # magictoken.ex_linking_cu.end
 
+    def test_ex_from_buffer(self):
```
stuartarchibald, line 48: Does this test need some `@skip` for environments in which it is not designed to work?

stuartarchibald (commented):
Thanks for the update @gmarkall. One minor comment to resolve else the patch looks good. Am just going to test this patch manually.

numba/cuda/tests/doc_examples/test_ffi.py
gmarkall: It's already skipped on cudasim and in environments where the NVIDIA bindings aren't available - the CC doesn't matter because it's linked with C++ source that is compiled for whatever the current CC is.

gmarkall:
@stuartarchibald Many thanks for the review - comment responded to (similar to OOB discussion as I recall it).

stuartarchibald (approved):

stuartarchibald:
> Thanks for the update @gmarkall. One minor comment to resolve else the patch looks good. Am just going to test this patch manually.

A manual testing of this patch was ok (essentially wrote something similar to the doc example).

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_187`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_187`.

Passed.

PR #8170, opened 2022-06-17
CUDA: Fix missing space in low occupancy warning

gmarkall:
The message was printing:

> Grid size N will likely result in GPUunder-utilization due to low occupancy.

instead of:

> Grid size N will likely result in GPU under-utilization due to low occupancy.


stuartarchibald (approved):
Thanks for the patch.

esc:
Thank you for the patch!

PR #8183, opened 2022-06-22
Add `n_keys` option to Dict.empty()

### LINKED ISSUE #8110: Minsize argument for Dict.empty()
## Feature request

Being able to allocate memory for a Dict based on problem-specific information would minimize resizes and increase performance. A Dict is initialized by a call to the function `numba_dict_new_minsize` in `dictobject.c`. The starting number of buckets is fixed  at `D_MINSIZE = 8`, which only allows for 5 entries without a resize.

The code comments state that this is suitable for the common case of a small dictionary used for passing keyword arguments. I believe an optional size argument would give users the ability to achieve significantly better performance with large hash tables. This would be in line with Numba's focus on fast numerical computation.

I've tried to modify Numba so that I could pass a size argument to `Dict.empty()` and have it call `numba_dict_new` instead of `numba_dict_new_minsize`. Unfortunately, I haven't been successful so far. Please consider adding such an option in a future release. Thank you.

stefanfed:
Adds the optional argument `n_keys` to the `Dict.empty()` method of typed dictionaries. `n_keys` is the number of keys to make space for so that resizes aren't required during insertion. To calculate the required number of slots, the passed value is divided by the load factor and rounded up to the nearest power of two. 

When the number of keys to be inserted can be known in advance, performance is improved (especially for larger dictionaries) because unnecessary allocations and copies are avoided. 

As discussed with @gmarkall, the kinds of tests to write will be determined after a preliminary review.

Resolves #8110.


gmarkall:
@stefanfed Many thanks for the PR! 

Some quick questions for now:

* Did you expect to see the failures in CI?
* Do you need guidance on resolving them?
* In the related issue and the description here, you mention some performance experiments you've done - could you share the details of the experiments you did please? (Ideally if I could try running them on my own machines, that would be great)

stefanfed:
@gmarkall,

I should have expected the tests to fail, because I renamed one of the core dictionary creation functions in `dictobject.py` from `new_dict()` to `new_dict_minsize()` for clarity. Looking at the logs, this seems to be the only cause of the failures:
`AttributeError: module 'numba.typed.dictobject' has no attribute 'new_dict'`

I can revert the name to what it was or modify the tests. Which would you prefer? 

I will make the required changes, test locally, and hopefully all of the tests pass here as well. I've had no problems building or using it on my Windows 10 x64 machine so far.

This is the kind of performance test I used:
```python
import numba as nb
from numba import njit
from numba.typed import Dict
import numpy as np

@njit
def test_minsize(keys):    
    d = Dict.empty(nb.int64, nb.int64, n_keys=None)
    
    for i in range(keys.shape[0]):
        k = keys[i]
        d[k] = k+1
    
    return d

@njit
def test_n_keys(keys):    
    d = Dict.empty(nb.int64, nb.int64, n_keys=keys.shape[0])
    
    for i in range(keys.shape[0]):
        k = keys[i]
        d[k] = k+1
    
    return d

# Compile first
test_minsize(np.arange(8, dtype='i8'))
test_n_keys(np.arange(8, dtype='i8'))

# Random ascending ints with no duplicates, size 2**n
n = 20
np.random.seed(0)
keys = np.arange(2**(n+2), step=4, dtype='i8') + \
    np.random.choice(np.arange(4, dtype='i8'), 2**n)

%timeit test_minsize(keys)
%timeit test_n_keys(keys)
```
On my computer, `test_minsize` takes 88.2ms while `test_n_keys` takes 41.2ms to add around 1 million keys. The performance will decrease substantially for both when the keys are shuffled, but the difference between the two stays more or less the same.

stefanfed:
Hi @gmarkall,

I substantially simplified the changes to the Python implementation. It now calls a single C function for both dictionary creation cases (minimum and custom sizes). The default value of `n_keys` has been changed to from `None` to `0`, which will always return a dictionary with minimum size. I restored the name of `new_dict()`, resolving the earlier attribute errors. Everything passes CI now, and I believe it's easier to reason about too.

I had a look at the tests in [test_dictobject.py](https://github.com/numba/numba/blob/94916113d92cb519a28adb10c30d8afda8049d07/numba/tests/test_dictobject.py) and [test_dictimpl.py](https://github.com/numba/numba/blob/main/numba/tests/test_dictimpl.py). I suppose I could repeat the tests with some custom allocation size passed through `n_keys`. I could also push changes I already made that remove the reliance of `test_dictimpl` on the C function `numba_dict_new_minsize()`. This is the last place it's used, and it could then be removed from the LLVM symbol table with `numba_dict_new_sized()` fully replacing it. Please let me know what you think.

gmarkall:
Many thanks for the updates! I think these changes are also along lines that @stuartarchibald asked me about (in a private chat message, not in this PR), suggesting to use a single function for both cases.

> I had a look at the tests in [test_dictobject.py](https://github.com/numba/numba/blob/94916113d92cb519a28adb10c30d8afda8049d07/numba/tests/test_dictobject.py) and [test_dictimpl.py](https://github.com/numba/numba/blob/main/numba/tests/test_dictimpl.py). I suppose I could repeat the tests with some custom allocation size passed through `n_keys`. I could also push changes I already made that remove the reliance of `test_dictimpl` on the C function `numba_dict_new_minsize()`. This is the last place it's used, and it could then be removed from the LLVM symbol table with `numba_dict_new_sized()` fully replacing it. Please let me know what you think.

I do think it would be good to fully replace `numba_dict_new_minsize()`, and remove it from the tests / symbol table. If you could add this change too, I think that would be a further improvement.

For the performance benchmarks on my system (Core i7-6700K, 64GB DDR4 2133 RAM), I get:

```
$ ipython repro.ipy 
81.5 ms ± 142 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
47 ms ± 216 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

so I see an improvement in line with your observations.

stuartarchibald:
> Many thanks for the updates! I think these changes are also along lines that @stuartarchibald asked me about (in a private chat message, not in this PR), suggesting to use a single function for both cases.
> 
> > I had a look at the tests in [test_dictobject.py](https://github.com/numba/numba/blob/94916113d92cb519a28adb10c30d8afda8049d07/numba/tests/test_dictobject.py) and [test_dictimpl.py](https://github.com/numba/numba/blob/main/numba/tests/test_dictimpl.py). I suppose I could repeat the tests with some custom allocation size passed through `n_keys`. I could also push changes I already made that remove the reliance of `test_dictimpl` on the C function `numba_dict_new_minsize()`. This is the last place it's used, and it could then be removed from the LLVM symbol table with `numba_dict_new_sized()` fully replacing it. Please let me know what you think.
> 
> I do think it would be good to fully replace `numba_dict_new_minsize()`, and remove it from the tests / symbol table. If you could add this change too, I think that would be a further improvement.
> 
> For the performance benchmarks on my system (Core i7-6700K, 64GB DDR4 2133 RAM), I get:
> 
> ```
> $ ipython repro.ipy 
> 81.5 ms ± 142 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
> 47 ms ± 216 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
> ```
> 
> so I see an improvement in line with your observations.

Thanks for looking at this @gmarkall @stefanfed.

As noted, I think the dictionary allocation could all go through a single function and use `0` as a default to specify `D_MINSIZE`. I think `0` is reasonable to use as a reserved item for this as `malloc(0)` behaviour is awkward and a typed truly empty dictionary is probably not that useful in practice.

RE: preventing resize on insertion, I see that the `numba_dict_new_sized` attempts to defeat the `USABLE_FRACTION` macro, would it be a good idea to create a macro that helps with this (i.e. calculates the inverse) and then pull out the constant values etc such that they are always kept in sync?

stefanfed:
Hi @gmarkall @stuartarchibald,

Thanks for the feedback. I created a new macro, `INV_USABLE_FRACTION`, and removed all dependencies on `numba_dict_new_minsize()` from the Python side. 

I also added some tests to [dictobject.c](https://github.com/numba/numba/blob/e4676841d54ff0f8d8b0430dd8c1757a704adfb1/numba/cext/dictobject.c) to ensure that the sizing mechanism is working correctly. They are run by [test_dictimpl.py](https://github.com/numba/numba/blob/e4676841d54ff0f8d8b0430dd8c1757a704adfb1/numba/tests/test_dictimpl.py). In my view, if the implementation passes all existing tests and the sizing mechanism works correctly, there's no room for other problems to creep in. Still, please let me know if any other tests are needed.

The Read the Docs build failed with the following message:
`error: Could not find suitable distribution for Requirement.parse('llvmlite<0.40,>=0.39.0dev0')`
This doesn't appear to be related to any modifications I had made on my side, and I don't know how to fix it.

esc:
> The Read the Docs build failed with the following message: `error: Could not find suitable distribution for Requirement.parse('llvmlite<0.40,>=0.39.0dev0')` This doesn't appear to be related to any modifications I had made on my side, and I don't know how to fix it.

This is a side-effect of the recent release. You can fix this by merging `main` into your branch `dict_empty_allocated`.

stefanfed:
@gmarkall I've marked this pull request as ready for review. I'm available if you would like any further changes to be made.

gmarkall (approved):
I think this is now looking good - as far as I can see all previous comments are resolved. I think it should also be possible to remove the `numba_dict_new_minsize()` function as well, e.g.:

```diff
diff --git a/numba/cext/dictobject.c b/numba/cext/dictobject.c
index 6cc641495..3557757ba 100644
--- a/numba/cext/dictobject.c
+++ b/numba/cext/dictobject.c
@@ -1024,12 +1024,6 @@ numba_dict_insert_ez(
     return numba_dict_insert(d, key_bytes, hash, val_bytes, old);
 }
 
-int
-numba_dict_new_minsize(NB_Dict **out, Py_ssize_t key_size, Py_ssize_t val_size)
-{
-    return numba_dict_new(out, D_MINSIZE, key_size, val_size);
-}
-
 
 /* Allocate a new dictionary with enough space to hold n_keys without resizes */
 int
@@ -1037,7 +1031,7 @@ numba_dict_new_sized(NB_Dict **out, Py_ssize_t n_keys, Py_ssize_t key_size, Py_s
 
     /* Respect D_MINSIZE */
     if (n_keys <= USABLE_FRACTION(D_MINSIZE)) {
-        return numba_dict_new_minsize(out, key_size, val_size);
+        return numba_dict_new(out, D_MINSIZE, key_size, val_size);
     }
 
     /*  Adjust for load factor */
diff --git a/numba/cext/dictobject.h b/numba/cext/dictobject.h
index 1c289b3dd..f635360a9 100644
--- a/numba/cext/dictobject.h
+++ b/numba/cext/dictobject.h
@@ -113,12 +113,6 @@ numba_dict_free(NB_Dict *d);
 NUMBA_EXPORT_FUNC(Py_ssize_t)
 numba_dict_length(NB_Dict *d);
 
-/* Allocates a new dict at the minimal size
-See numba_dict_new().
-*/
-NUMBA_EXPORT_FUNC(int)
-numba_dict_new_minsize(NB_Dict **out, Py_ssize_t key_size, Py_ssize_t val_size);
-
 /* Set the method table for type specific operations
 */
 NUMBA_EXPORT_FUNC(void)
```

but I don't feel strongly about it either way because it's quite a minimal bit of code.

I think this should probably have a quick look / sanity check from another reviewer to make sure I'm not missing something obvious that's outstanding.

stuartarchibald (commented):
Apologies for the delay, I've taken a look at this as requested and it looks good (just one query I've left in-line), the reported performance numbers are encouraging too. RE: the note above about cleaning up the C code https://github.com/numba/numba/pull/8183#pullrequestreview-1145844577, seems like this would be a good idea to do as part of this given the changes in this patch provide a superset of the existing API. Many thanks for working on this @stefanfed and thanks for reviewing @gmarkall.

numba/cext/dictobject.c
```diff
@@ -368,15 +368,33 @@ set_index(NB_DictKeys *dk, Py_ssize_t i, Py_ssize_t ix)
...
+ * of 2, must add 1 to the result of the inverse for correct sizing.
+ * 
+ * For example, when USABLE_FRACTION ratio = 5/8 (8 is a power of 2):
+ * #define INV_USABLE_FRACTION(n) (((n) << 3)/5)  // inv_ratio: 8/5
+ * 
+ * When USABLE_FRACTION ratio = 5/7 (7 is not a power of 2):
+ * #define INV_USABLE_FRACTION(n) ((7*(n))/5 + 1)  // inv_ratio: 7/5
+ */
+
+#define INV_USABLE_FRACTION(n) ((n) + ((n) >> 1) + 1)  // inv_ratio: 3/2
```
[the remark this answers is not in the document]
gmarkall: I'm happy with it. I can't remember my thinking / checking at the time, but I do recall that the majority of time I spent here was thinking about this and convincing myself it was OK :slightly_smiling_face:

numba/typed/dictobject.py
```diff
@@ -242,7 +247,7 @@ def codegen(context, builder, sig, args):
         refdp = cgutils.alloca_once(builder, ll_dict_type, zfill=True)
         status = builder.call(
             fn,
-            [refdp, ll_ssize_t(sz_key), ll_ssize_t(sz_val)],
+            [refdp, args[2], ll_ssize_t(sz_key), ll_ssize_t(sz_val)],
```
stuartarchibald, line 250: Does `args[2]` need a cast to a `size_t`?

stuartarchibald (commented):
Thanks for applying the discussed merge/fixes @gmarkall, they look good. I just had a last check over this and wondered if there needs to be a cast made in the intrinsic, or at the point of calling the intrinsic (highlighted in-line). Final thing before merge is whether this needs a run through the build farm, I'd probably say yes given it's C code changes, what do you think?

numba/cext/dictobject.c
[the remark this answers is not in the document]
stuartarchibald: Great, thanks for checking this!

numba/typed/dictobject.py
gmarkall: `numba_dict_new_sized` takes a `Py_ssize_t` for `n_keys`:

```C
int numba_dict_new_sized(NB_Dict **out, Py_ssize_t n_keys, Py_ssize_t key_size, Py_ssize_t val_size);
```

so I've changed the code to cast to that. We were getting away with it at least in CI because `args[2]` was an `int64`. Whilst making this change I also edited the `codegen` function so it's more of a straight-line read rather than multi-line-things-happening-in-the-indents code.

stefanfed: I want to point out that `listobject.py` also passes the size parameter without casting with `ll_ssize_t()` (as far as I can tell). 

`_list_new` passes `args[1]` to `_list_new_codegen` as `new_size`. There is no casting of this variable. It is then passed to `numba_list_new` in `listobject.c` as the argument `allocated`, which is a `Py_ssize_t`. I imitated this approach. Perhaps this needs to be fixed as well?

Sorry for the lack of responses lately. I've been busy.

gmarkall: No problem, and thanks for all your efforts so far! If `_list_new_codegen` does this, then it probably ought to be fixed too. However, as this PR is nearly over the line I'd suggest we do that in a separate PR (if we're going to... As nobody noticed now it seems like we're getting away with it 🙂 )

gmarkall:
@stuartarchibald I've additionally updated the order of args in the intrinsic as discussed OOB yesterday - this should be ready for another look.

gmarkall:
(Also added the Pending Buildfarm label to track this having a run on it)

stuartarchibald:
> @stuartarchibald I've additionally updated the order of args in the intrinsic as discussed OOB yesterday - this should be ready for another look.

Many thanks. Patch looks good.

stuartarchibald (approved):
@stefanfed and @gmarkall many thanks to you both for your efforts on this patch. Am glad to see this change as it is a step towards higher performance dictionaries.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cpu_yaml_168`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cpu_yaml_168`.

This "passed" sufficiently (failed on `osx-64` due to: #8567 and on `win32` due to an unrelated problem in `np.argpartition` tests which shows up on `main` too).

PR #8185, opened 2022-06-23
Resolve overload ambiguity

sklam:
- Adds `@impl_for` to mark overload implementation that is specializing for a type class. The new feature is enabled when the `@overload` to set `use_impl_for=True`. e.g.
  ```python
        @overload(foo, use_impl_for=True)
        def ov_foo_base(x, y):
            if isinstance(x, types.Array) and isinstance(y, types.Array):
                @impl_for(types.Array)
                def impl(x, y):
                    return "base"
                return impl

        @overload(foo, use_impl_for=True)
        def ov_foo_tensor(x, y):
            if isinstance(x, types.Array) and isinstance(y, types.Array):
                 # Tensor is subtype of Array
                if isinstance(x, Tensor) or isinstance(y, Tensor):
                    @impl_for(Tensor)
                    def impl(x, y):
                        return "tensor"
                    return impl
  ```
  The `@impl_for(Tensor)` overload is more specific than the `@impl_for(types.Array)`
- Old-style can set `impl_for` in the returning `Signature` in the typing logic.
- The new behavior changes the overload resolution when one of the _matching_ overloads (incl old-styles) defines `impl_for`.
   - All _matching_ overloads  are ranked by their specificity as defined by `impl_for`. The most specific one is used.
   - If there's a tie, resolution fails.
   - If the "impl_for" types in all _matching_ overloads do not share the same type ancestry, resolution fails.


sklam:
hmm... it's blowing up the python stack in half of the tests.

sklam:
https://github.com/numba/numba/pull/8185/commits/89c95de7d34bf4fd218288e2a59680e01d8aa0f2 needed because of https://github.com/python/cpython/issues/89726

sklam:
Meeting notes:

a) What if two implementation are on the same level in the type hierarchy? i.e. `Dense <: Array` and `Sparse <: Array` and `gemm(Dense, Sparse)`.
b) Is the below better?
```python
      @overload(foo, impl_for=types.Array)
      def ov_foo_base(x, y):
          if isinstance(x, types.Array) and isinstance(y, types.Array):
              def impl(x, y):
                  return "base"
              return impl
```

sklam:
note to self:

- [ ] [Fix failure with disjoint type hierarchy](https://github.com/numba/numba/pull/8185/commits/66314e287ea7bcba6dc335c7041164742e1422b1) needs a test

PR #8193, opened 2022-06-24
Add interface to allow extracting remark information

apmasell:
Allows setting an environment variable to dump LLVM remark information for later analysis.


stuartarchibald:
@apmasell Thanks for the patch. Was wondering if it would be preferable to expose remarks information in the function's per-signature metadata dictionary? I think it'd be possible to collect information from both the "cheap" and "full" module optimisation passes, as well as the function level optimisation passes. Perhaps something to discuss at the next public Numba meeting?

apmasell:
We could. The goal of this was to make something that could be fed into the external LLVM tools with little effort for our own debugging rather than something user friendly. Also, we don't know what the memory cost of these will be for larger programs.

numba/core/codegen.py
```diff
@@ -659,9 +660,13 @@ def _optimize_functions(self, ll_module):
             # module-level optimization.
             for func in ll_module.functions:
                 k = f"Function passes on {func.name!r}"
+                remarks_dir = os.environ.get("NUMBA_LLVM_OPT_REMARKS_DIR", None)
+                remarks_file = os.path.join(remarks_dir,
+                                            "%s-remarks.yaml" % func.name
+                                            ) if remarks_dir else None
```
sklam, line 666: This needs to be moved to https://github.com/numba/numba/blob/203d6548a6f84e51f7346c47da271a47799f187d/numba/core/codegen.py#L687. It is where all the aggressive optimization take place.

Something like:

```diff
diff --git a/numba/core/codegen.py b/numba/core/codegen.py
index 472c7aede..17a1342ff 100644
--- a/numba/core/codegen.py
+++ b/numba/core/codegen.py
@@ -660,13 +660,9 @@ class CPUCodeLibrary(CodeLibrary):
             # module-level optimization.
             for func in ll_module.functions:
                 k = f"Function passes on {func.name!r}"
-                remarks_dir = os.environ.get("NUMBA_LLVM_OPT_REMARKS_DIR", None)
-                remarks_file = os.path.join(remarks_dir,
-                                            "%s-remarks.yaml" % func.name
-                                            ) if remarks_dir else None
                 with self._recorded_timings.record(k):
                     fpm.initialize()
-                    fpm.run(func, remarks_file=remarks_file)
+                    fpm.run(func)
                     fpm.finalize()

     def _optimize_final_module(self):
@@ -682,9 +678,13 @@ class CPUCodeLibrary(CodeLibrary):
         if not config.LLVM_REFPRUNE_PASS:
             self._final_module = remove_redundant_nrt_refct(self._final_module)
         full_name = "Module passes (full optimization)"
+        remarks_dir = os.environ.get("NUMBA_LLVM_OPT_REMARKS_DIR", None)
+        remarks_file = os.path.join(remarks_dir,
+                                    "%s-remarks.yaml" % self.name
+                                    ) if remarks_dir else None
         with self._recorded_timings.record(full_name):
             # The full optimisation suite is then run on the refop pruned IR
-            self._codegen._mpm_full.run(self._final_module)
+            self._codegen._mpm_full.run(self._final_module, remarks_file=remarks_file)

     def _get_module_for_linking(self):
         """
```

sklam: with that, i get:

```
--- !Analysis
Pass:            loop-vectorize
Name:            CFGNotUnderstood
DebugLoc:        { File: optremarks.py, Line: 7, Column: 1 }
Function:        _ZN8__main__3fooB2v1B54c8tJTC_2fWwKbcFpQAdBTgEQrlo6UyVFGMNAWVBgYHmKUJAA_3d_3dE5ArrayIfLi1E1C7mutable7alignedE
Args:
  - String:          'loop not vectorized: '
  - String:          loop control flow is not understood by vectorizer
...
--- !Analysis
Pass:            loop-vectorize
Name:            CFGNotUnderstood
DebugLoc:        { File: optremarks.py, Line: 7, Column: 1 }
Function:        _ZN8__main__3fooB2v1B54c8tJTC_2fWwKbcFpQAdBTgEQrlo6UyVFGMNAWVBgYHmKUJAA_3d_3dE5ArrayIfLi1E1C7mutable7alignedE
Args:
  - String:          'loop not vectorized: '
  - String:          loop control flow is not understood by vectorizer
...
```
from source:

```python
from numba import njit
import numpy as np


@njit(debug=True)
def foo(arr):
    c = arr + arr.sum()
    return c

res = foo(np.arange(100).astype(np.float32))
print(res)
```

numba/core/codegen.py
```diff
@@ -668,18 +673,58 @@ def _optimize_final_module(self):
...
+                # A cheaper optimisation pass is run first to try and get as
+                # many refops into the same function as possible via inlining
+                self._codegen._mpm_cheap.run(self._final_module,
+                                             remarks_file=remarks_file,
+                                             remarks_filter=remarks_filter)
+            if remarks_file:
+                import yaml
+                with open(remarks_file, 'rt') as f:
+                    remarks_data.append([*yaml.load_all(f,
+                                                        Loader=_llvm_yaml())])
```
sklam, line 704: maybe worth annotating what pass run is generating these remarks; e.g. this will be from the cheap pass. For most users, remarks generated at this phase is not as interesting as the later one.

numba/core/codegen.py
```diff
@@ -668,18 +673,58 @@ def _optimize_final_module(self):
...
+        remarks_data = []
+        try:
+            if optimization_processors:
+                remarks_filter = "|".join("(%s)" % f for p in
+                                          optimization_processors for f in
+                                          p.filters())
+                remarkdesc, remarks_file = mkstemp()
+                # Close file
+                with os.fdopen(remarkdesc, 'r'):
+                    pass
```
sklam, line 690: couldn't we let llvmlite handle the file by using `run_with_remarks`?

numba/core/codegen.py
```diff
@@ -668,18 +673,58 @@ def _optimize_final_module(self):
...
+            with self._recorded_timings.record(full_name):
+                # The full optimisation suite is then run on the refop pruned IR
+                self._codegen._mpm_full.run(self._final_module,
+                                            remarks_file=remarks_file,
+                                            remarks_filter=remarks_filter)
+            if remarks_file:
+                import yaml
+                with open(remarks_file, 'rt') as f:
+                    remarks_data.append([*yaml.load_all(f,
+                                                        Loader=_llvm_yaml())])
```
sklam, line 718: Can we refactor this out so we don't duplicate? Also the refactoring may remove the need to importing `yaml` twice.

numba/core/opt_info.py
```diff
@@ -0,0 +1,183 @@
...
+    def needs_debug_info(self) -> bool:
+        return any(p.needs_debug_info() for p in self._processors)
+
+    def process(self, remarks_data: List[Any], full_name: str) ->\
+            Iterable[Tuple[str, Any]]:
+        return (o for p in self._processors for o in p.process(remarks_data,
+                                                               full_name))
+
+
+class RawOptimizationRemarks(OptimizationProcessor):
```
sklam, line 66: For consistency, would it be better if the class name matches the "key" in the `process()` method?

numba/core/opt_info.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def needs_debug_info(self) -> bool:
+        return False
+
+    def process(self, remarks_data: List[Any], function) -> Iterable[
+            Tuple[str, Any]]:
+        return ('raw', remarks_data),
+
+
+class LoopDeLoop(OptimizationProcessor):
```
sklam, line 89: Why not just name this `LoopVectorizationDetector` since it is entirely focusing on loop vectorization?

sklam (commented):
I reviewed from the perspective of generating annotations for users. 

In addition to my review comments, i made a hacky script to annotate source code with the remarks at https://gist.github.com/sklam/3890954b5335a1e4907ef1e75c73cd31. It generates something like below:

<details>
<summary>source+remarks</summary>

```python
# 18
# ...
# !!python/object/new:numba.core.opt_info.Passed
# - Args:
#   - String: 'sinking '
#   - DebugLoc:
#       Column: 1
#       File: chk.py
#       Line: 18
#     Inst: getelementptr
#   DebugLoc:
#     Column: 1
#     File: chk.py
#     Line: 18
#   Function: _ZN8__main__3fooB2v1B128c8tJTC_2fWwKbRFpQUdBTgUQvlo6U3VFGM1AWVBgYMlAWz2TbaJg_2bUsPWS84tS9WCiekGJ5SBf5mZWJYL8GpSam1iUXawAMVUhsUTBoMI8LTnFyMLAOCUtxcAuVhMAEx
#   Name: InstSunk
#   Pass: licm
@njit(opt_info=[RawOptimizationRemarks()], debug=True)

# 18
# ...
# !!python/object/new:numba.core.opt_info.Missed
# - Args:
#   - String: failed to move load with loop-invariant address because the loop may invalidate
#       its value
#   DebugLoc:
#     Column: 1
#     File: chk.py
#     Line: 18
#   Function: _ZN8__main__3fooB2v1B128c8tJTC_2fWwKbRFpQUdBTgUQvlo6U3VFGM1AWVBgYMlAWz2TbaJg_2bUsPWS84tS9WCiekGJ5SBf5mZWJYL8GpSam1iUXawAMVUhsUTBoMI8LTnFyMLAOCUtxcAuVhMAEx
#   Name: LoadWithLoopInvariantAddressInvalidated
#   Pass: licm
def foo(n):

    # 18
    # ...
    # !!python/object/new:numba.core.opt_info.Passed
    # - Args:
    #   - String: 'sinking '
    #   - DebugLoc:
    #       Column: 1
    #       File: chk.py
    #       Line: 18
    #     Inst: getelementptr
    #   DebugLoc:
    #     Column: 1
    #     File: chk.py
    #     Line: 18
    #   Function: _ZN8__main__3fooB2v1B128c8tJTC_2fWwKbRFpQUdBTgUQvlo6U3VFGM1AWVBgYMlAWz2TbaJg_2bUsPWS84tS9WCiekGJ5SBf5mZWJYL8GpSam1iUXawAMVUhsUTBoMI8LTnFyMLAOCUtxcAuVhMAEx
    #   Name: InstSunk
    #   Pass: licm
    c = np.zeros(n)

    # 18
    # ...
    # !!python/object/new:numba.core.opt_info.Passed
    # - Args:
    #   - String: 'sinking '
    #   - DebugLoc:
    #       Column: 1
    #       File: chk.py
    #       Line: 18
    #     Inst: getelementptr
    #   DebugLoc:
    #     Column: 1
    #     File: chk.py
    #     Line: 18
    #   Function: _ZN8__main__3fooB2v1B128c8tJTC_2fWwKbRFpQUdBTgUQvlo6U3VFGM1AWVBgYMlAWz2TbaJg_2bUsPWS84tS9WCiekGJ5SBf5mZWJYL8GpSam1iUXawAMVUhsUTBoMI8LTnFyMLAOCUtxcAuVhMAEx
    #   Name: InstSunk
    #   Pass: licm
    for i in range(n):

        # 18
        # ...
        # !!python/object/new:numba.core.opt_info.Passed
        # - Args:
        #   - String: 'sinking '
        #   - DebugLoc:
        #       Column: 1
        #       File: chk.py
        #       Line: 18
        #     Inst: getelementptr
        #   DebugLoc:
        #     Column: 1
        #     File: chk.py
        #     Line: 18
        #   Function: _ZN8__main__3fooB2v1B128c8tJTC_2fWwKbRFpQUdBTgUQvlo6U3VFGM1AWVBgYMlAWz2TbaJg_2bUsPWS84tS9WCiekGJ5SBf5mZWJYL8GpSam1iUXawAMVUhsUTBoMI8LTnFyMLAOCUtxcAuVhMAEx
        #   Name: InstSunk
        #   Pass: licm
        c += i

    # 18
    # ...
    # !!python/object/new:numba.core.opt_info.Missed
    # - Args:
    #   - String: failed to move load with loop-invariant address because the loop may invalidate
    #       its value
    #   DebugLoc:
    #     Column: 1
    #     File: chk.py
    #     Line: 18
    #   Function: _ZN8__main__3fooB2v1B128c8tJTC_2fWwKbRFpQUdBTgUQvlo6U3VFGM1AWVBgYMlAWz2TbaJg_2bUsPWS84tS9WCiekGJ5SBf5mZWJYL8GpSam1iUXawAMVUhsUTBoMI8LTnFyMLAOCUtxcAuVhMAEx
    #   Name: LoadWithLoopInvariantAddressInvalidated
    #   Pass: licm
    return c


```

</details>

numba/core/codegen.py
apmasell: Not without having totally independent logic for the two paths and I'd rather not do that in case there are any changes to the pass manager logic.

numba/core/opt_info.py
apmasell: Since they can emit multiple values, no. We can force them to only emit one if you prefer.

PR #8195, opened 2022-06-27
Make the workqueue threading backend once again fork safe.

### LINKED ISSUE #7872: Interpreter hangs when running parallel function in subprocess with workqueue backend
## Reporting a bug

## Description

The Python interpreter hangs when using Numba in the following way:
1. Select the "forksafe" `workqueue` threading backend
1. Run a Numba parallel function
1. Start a subprocess (e.g. using Python `multiprocessing`)
1. Run a Numba parallel function in the subprocess

If I understand correctly, this pattern should be allowed when selecting a forksafe threading layer. Is this correct? Or is one only allowed to run parallel functions in a subprocess, but not mix parent and subprocess work?

## Versions

I ran into the issue on Numba 0.55.1 (installed from pip), and then I installed from master (0.56.0dev0+298.g53ea89fee) and found the same issue.  I thought it might be a specific issue to my local machine, but it also fails on [GitHub CI](https://github.com/abacusorg/abacusutils/runs/5363451592?check_suite_focus=true) in the same way.

I've been testing using Python 3.9 on Linux.  GitHub CI fails on Python 3.7-3.10.

https://github.com/numba/numba/issues/5890 seemed related, so I tried installing the PR from https://github.com/numba/numba/pull/7625, but it has no effect on this problem.

I haven't tried a conda install or a TBB threading layer yet; will try that and report back.

## Debugging Attempts

Valgrind gives a pleasantly coherent view of what's happening. It appears the subprocess segfaults inside a workqueue mutex, thus leaving the parent process hanging, waiting for the child to return its result.  This segfault story also meshes with the GitHub CI result from above, where pytest seems to notice that something segfaulted.

Note the PID changes after a few lines, as Valgrind starts tracking the child.

```console
(venv) lgarrison@ccalin008:~/abacusutils$ valgrind --suppressions=valgrind-python.supp python3 repro_gh47.py
==1400719== Memcheck, a memory error detector
==1400719== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==1400719== Using Valgrind-3.17.0 and LibVEX; rerun with -h for copyright info
==1400719== Command: python3 repro_gh47.py
==1400719== 
Single-process result: [ 0  2  4  6  8 10 12 14 16 18]
numba.threading_layer()='workqueue'
==1401030== Invalid read of size 4
==1401030==    at 0x58B7D00: pthread_mutex_lock (in /usr/lib64/libpthread-2.17.so)
==1401030==    by 0xE845DA63: queue_condition_lock (workqueue.c:89)
==1401030==    by 0xE845DA63: queue_state_wait (workqueue.c:260)
==1401030==    by 0xE845DA63: ready (workqueue.c:554)
==1401030==    by 0xE845DA63: parallel_for (workqueue.c:401)
==1401030==    by 0x404E19A: ???
==1401030==    by 0x3: ???
==1401030==    by 0x17: ???
==1401030==    by 0xFEFF706F: ???
==1401030==    by 0x8: ???
==1401030==  Address 0x40 is not stack'd, malloc'd or (recently) free'd
==1401030== 
==1401030== 
==1401030== Process terminating with default action of signal 11 (SIGSEGV)
==1401030==  Access not within mapped region at address 0x40
==1401030==    at 0x58B7D00: pthread_mutex_lock (in /usr/lib64/libpthread-2.17.so)
==1401030==    by 0xE845DA63: queue_condition_lock (workqueue.c:89)
==1401030==    by 0xE845DA63: queue_state_wait (workqueue.c:260)
==1401030==    by 0xE845DA63: ready (workqueue.c:554)
==1401030==    by 0xE845DA63: parallel_for (workqueue.c:401)
==1401030==    by 0x404E19A: ???
==1401030==    by 0x3: ???
==1401030==    by 0x17: ???
==1401030==    by 0xFEFF706F: ???
==1401030==    by 0x8: ???
==1401030==  If you believe this happened as a result of a stack
==1401030==  overflow in your program's main thread (unlikely but
==1401030==  possible), you can try to increase the size of the
==1401030==  main thread stack using the --main-stacksize= flag.
==1401030==  The main thread stack size used in this run was 8388608.
==1401030== 
==1401030== HEAP SUMMARY:
==1401030==     in use at exit: 24,269,054 bytes in 24,666 blocks
==1401030==   total heap usage: 639,280 allocs, 614,614 frees, 1,540,075,539 bytes allocated
==1401030== 
==1401030== LEAK SUMMARY:
==1401030==    definitely lost: 576 bytes in 12 blocks
==1401030==    indirectly lost: 0 bytes in 0 blocks
==1401030==      possibly lost: 985,934 bytes in 4,064 blocks
==1401030==    still reachable: 23,282,544 bytes in 20,590 blocks
==1401030==                       of which reachable via heuristic:
==1401030==                         multipleinheritance: 6,896 bytes in 8 blocks
==1401030==         suppressed: 0 bytes in 0 blocks
==1401030== Rerun with --leak-check=full to see details of leaked memory
==1401030== 
==1401030== For lists of detected and suppressed errors, rerun with: -s
==1401030== ERROR SUMMARY: 1 errors from 1 contexts (suppressed: 84136 from 1044)
```

## Reproducer

```python
#!/usr/bin/env python3

import numba
numba.config.THREADING_LAYER = 'workqueue'

import numpy as np

# === Run a Numba parallel function in a single process ===

@numba.njit(parallel=True)
def f(a):
    return 2*a

res = f(np.arange(10))
print(f'Single-process result: {res}', flush=True)
print(f'{numba.threading_layer()=}')

# === Now run a Numba parallel function in a forked process ===

@numba.njit(parallel=True)
def g(a):
    return 3*a
    
import multiprocessing
with multiprocessing.Pool(1) as p:
    mres = p.map(g, [np.arange(10)])  # hangs

print(f'Forked process result: {mres}', flush=True)
```

x-ref: https://github.com/abacusorg/abacusutils/issues/47

stuartarchibald:
As title, this fixes the workqueue threading backend such that it
is fork safe from _any_ thread, not just the main thread.

Fixes #7872

### DIFF
--- a/numba/tests/test_parallel_backend.py
+++ b/numba/tests/test_parallel_backend.py
@@ -690,6 +690,7 @@ class TestMiscBackendIssues(ThreadLayerTestHelper):
             self.assertIn("Terminating: Nested parallel kernel launch detected",
                           e_msg)
 
+    @unittest.skipUnless(_HAVE_OS_FORK, "Test needs fork(2)")
     def test_workqueue_handles_fork_from_non_main_thread(self):
         # For context see #7872, but essentially the multiprocessing pool
         # implementation has a number of Python threads for handling the worker
@@ -702,6 +703,8 @@ class TestMiscBackendIssues(ThreadLayerTestHelper):
             import multiprocessing
 
             if __name__ == "__main__":
+                # Need for force fork context (OSX default is "spawn")
+                multiprocessing.set_start_method('fork')
 
                 @njit(parallel=True)
                 def func(x):
@@ -718,6 +721,8 @@ class TestMiscBackendIssues(ThreadLayerTestHelper):
                     result = p.map(func, [arr])
                 np.testing.assert_allclose(result,
                                            func.py_func(np.expand_dims(arr, 0)))
+
+                assert threading_layer() == "workqueue"
         """
         cmdline = [sys.executable, '-c', runme]
         env = os.environ.copy()
--- a/numba/tests/test_parallel_backend.py
+++ b/numba/tests/test_parallel_backend.py
@@ -690,6 +690,42 @@ class TestMiscBackendIssues(ThreadLayerTestHelper):
             self.assertIn("Terminating: Nested parallel kernel launch detected",
                           e_msg)
 
+    def test_workqueue_handles_fork_from_non_main_thread(self):
+        # For context see #7872, but essentially the multiprocessing pool
+        # implementation has a number of Python threads for handling the worker
+        # processes, one of which calls fork(2), this results in a fork from a
+        # non-main thread.
+
+        runme = """if 1:
+            from numba import njit, prange, threading_layer
+            import numpy as np
+            import multiprocessing
+
+            if __name__ == "__main__":
+
+                @njit(parallel=True)
+                def func(x):
+                    return 10. * x
+
+                arr = np.arange(2.)
+
+                # run in single process to start Numba's thread pool
+                np.testing.assert_allclose(func(arr), func.py_func(arr))
+
+                # now run in a multiprocessing pool to get a fork from a
+                # non-main thread
+                with multiprocessing.Pool(10) as p:
+                    result = p.map(func, [arr])
+                np.testing.assert_allclose(result,
+                                           func.py_func(np.expand_dims(arr, 0)))
+        """
+        cmdline = [sys.executable, '-c', runme]
+        env = os.environ.copy()
+        env['NUMBA_THREADING_LAYER'] = "workqueue"
+        env['NUMBA_NUM_THREADS'] = "4"
+
+        self.run_cmd(cmdline, env=env)
+
 
 # 32bit or windows py27 (not that this runs on windows)
 @skip_parfors_unsupported
--- a/numba/np/ufunc/workqueue.c
+++ b/numba/np/ufunc/workqueue.c
@@ -157,7 +157,7 @@ platform_launch(void)
     pthread_key_create(&tidkey, NULL);
 }
 
-#endif
+#endif /* pthread threading */
 
 /* Win Thread */
 #ifdef NUMBA_WINTHREAD
@@ -262,7 +262,7 @@ platform_launch(void)
     tidkey = TlsAlloc();
 }
 
-#endif
+#endif /* Windows threading */
 
 typedef struct Task
 {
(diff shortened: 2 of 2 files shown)


lgarrison (dismissed):
Looks good to me! The logic seems pretty straightforward: if a parallel function is called and the thread pool is not yet initialized, start it. `queues = NULL` is used to signal that the thread pool is not active, via a `reset_after_fork()` handler.

The other piece seems to be that the global `NUM_THREADS` must be set by the caller before `launch_threads()` is entered (i.e. set by `_launch_threads()` in Python). But once it is set, it will be initialized in any subprocesses (because of `NUM_THREADS = _INIT_NUM_THREADS;` in the fork handler), so subprocesses don't need to call `_launch_threads()` again.

I guess this means that subprocesses inherit the same upper limit on threads as the parent process. But this is probably fine, since the upper limit is set explicitly by `NUMBA_NUM_THREADS` or inferred from the parent process affinity mask. Changing either of these in a child process would be a strange use-case, and so I think the current behavior makes sense.

I checked that this PR fixes the original issue https://github.com/abacusorg/abacusutils/issues/47, and fixes the test cases derived from that issue.

numba/tests/test_parallel_backend.py
```diff
@@ -690,6 +690,42 @@ def main():
...
+                    return 10. * x
+
+                arr = np.arange(2.)
+
+                # run in single process to start Numba's thread pool
+                np.testing.assert_allclose(func(arr), func.py_func(arr))
+
+                # now run in a multiprocessing pool to get a fork from a
+                # non-main thread
+                with multiprocessing.Pool(10) as p:
```
sklam, line 717: needs to force a fork context here

stuartarchibald: Good point, OSX defaults to `'spawn'`.

stuartarchibald: b07d687 fixes, also adds a test skip for OS which don't have `fork(2)` and adds an assert to ensure the expected threading layer was used.

stuartarchibald:
> Looks good to me! The logic seems pretty straightforward: if a parallel function is called and the thread pool is not yet initialized, start it. `queues = NULL` is used to signal that the thread pool is not active, via a `reset_after_fork()` handler.
> 
> The other piece seems to be that the global `NUM_THREADS` must be set by the caller before `launch_threads()` is entered (i.e. set by `_launch_threads()` in Python). But once it is set, it will be initialized in any subprocesses (because of `NUM_THREADS = _INIT_NUM_THREADS;` in the fork handler), so subprocesses don't need to call `_launch_threads()` again.

Thanks for the detailed review @lgarrison, much appreciated.

> I guess this means that subprocesses inherit the same upper limit on threads as the parent process. But this is probably fine, since the upper limit is set explicitly by `NUMBA_NUM_THREADS` or inferred from the parent process affinity mask. Changing either of these in a child process would be a strange use-case, and so I think the current behavior makes sense.

I thought the same, an upper thread limit that changes would be unusual, and also quite tricky to implement. Such a behaviour could be emulated by creating a larger than needed upper limit and then using thread masks if desired.

> I checked that this PR fixes the original issue [abacusorg/abacusutils#47](https://github.com/abacusorg/abacusutils/issues/47), and fixes the test cases derived from that issue.

Great, thanks for checking.

sklam (approved):
LGTM

PR #8197, opened 2022-06-27
Fix `isinstance` warning check test.

stuartarchibald:
The `isinstance` warning is raised from an `@overload`, as a
result its appearance is stateful with respect to the use of
`isinstance` in a JIT function. This patch pushes the test into
a subprocess to guarantee the state, i.e. has not been used in
the subprocess so the warning will always appear.


esc:
`Build numba_smoketest_cpu_yaml_112 has started`

PR #8200, opened 2022-06-28
Make the NRT use the "unsafe" allocation API by default.

### LINKED ISSUE #7887: Performance of Numba Runtime Allocations
Hello,

I have tested the performance creating very short Numpy array in Numba  njit function and I am a bit surprised of the computation time. I found out that a significant part of the time is spent in NRT functions defined in `core/runtime/nrt.c`, and more specifically, in `memset` calls (in `NRT_MemInfo_alloc_safe_aligned` and `nrt_internal_dtor_safe`) that appear to only be useful for debugging purpose. Is it the case? If so, I propose to add a debug boolean parameter to the function so to enable this only in debug mode. A faster alternative solution is to add functions with the _debug prefix that can be called by the JIT only in debug mode. Such `memset` calls takes 20-25% of the time of very short Numpy allocations on my machine (with a i5-9600KF processor).

Additionally, I am wondering why such calls are not inlined by the JIT. Some calls like this could be faster if they were inlined because of constant propagation and the expression optimization of the provided function arguments. Is this behavior intended?

Finally, I found that `nrt_allocate_meminfo_and_data_align` performs a modulus that tends to be quite slow. Since the divider is the memory alignment which is generally a power of two, this modulus can be optimized. We can check if the alignment is a power of two with `if((align & (align - 1)) == 0)`, and if this is the case, then use a simple `remainder = intptr & (align - 1)`. Otherwise, a modulus can be used (rare case so the processor can predict the conditional). If the alignment is guaranteed to be always a power of two, then the conditional can even be removed! Note that this strategy is already used by most compilers for stack alignment of optimized programs and optimized memory allocators.

Here is a simple code to test this:

```python
import numba as nb
import numpy as np

@nb.njit('void()')
def test():
    for i in range(1000):
        np.empty(2)

%timeit  -n 20_000 test()
```

stuartarchibald:
This patch adds to the NRT C API to create equivalent "safe" and
"unsafe" functions for allocation associated with meminfo. The
NRT context is updated to only use the "safe" functions if
`NRT_DEBUG` is enabled. This is to increase performance by only
calling `memset(3)` to generate debug markers at allocation time
when in `NRT_DEBUG` mode. The "aligned" allocators are updated to
calculate alignment offsets avoiding modulo if the alignment is
specified as a power of 2.

Fixes #7887


numba/core/config.py
```diff
@@ -220,7 +220,7 @@ def optional_str(x):
         # (up to and including IR generation)
         DEBUG_FRONTEND = _readenv("NUMBA_DEBUG_FRONTEND", int, 0)
 
-        # Enable debug prints in nrtdynmod
+        # Enable debug prints in nrtdynmod and use of "safe" API functions
```
gmarkall, line 223: I was going to suggest updating the documentation for `NUMBA_DEBUG_NRT` but it appears not to be documented at all. Is it worth adding a note on this to the documentation (in the [list of environment variables](https://numba.readthedocs.io/en/stable/reference/envvars.html)) and/or in the [Debugging Leaks](https://numba.readthedocs.io/en/stable/developer/numba-runtime.html#debugging-leaks) section?

gmarkall:
WRT benchmarking, I modified the example slightly to use `timeit`:

```python
import numba as nb
import numpy as np

@nb.njit('void()')
def test():
    for i in range(1000):
        np.empty(2)


def call_test():
    n = 20000
    for x in range(n):
        test()

%timeit call_test()
```

I get:

```
1.1 s ± 5.08 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
833 ms ± 1.65 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

before and after, which seems to be a reduction in execution time of about 24%.

gmarkall:
I also tried undoing the alignment optimization - for the above benchmark it makes about 1% difference in execution time on an i7-6700K, so I think it's definitely worth keeping considering other machines (small ARM SBCs, etc) will probably have a much worse modulo operation.

gmarkall (approved):
Thanks for the updates. This now looks good!

PR #8202, opened 2022-06-28
Searchsorted unichr

### LINKED ISSUE #8190: `searchsorted` not working with unicode arrays
## Reporting a bug

The `np.searchsorted` function does not work with `@njit` when the query array `v` is of unicode dtype. This seems to be because the `np.isnan` function is not implemented for this type. I would like to get this working so that I can use `np.searchsorted` in parallel using `prange` and `@njit(parallel=True)`.

MWE:
```python
# imports
import numpy as np

# attributes
UNICODE_a = 97
ALPHABET_SIZE = 26

# parameters
n_reference_string = 100
n_query_string = 10
string_length = 32

# create and sort random reference strings
reference_strings_unicode = np.random.choice(
    np.arange(UNICODE_a, UNICODE_a + ALPHABET_SIZE, dtype=np.uint32),
    size=(n_reference_string, string_length),
)
reference_strings = reference_strings_unicode.view(f"<U{string_length}").reshape(-1)
sorted_reference_strings = np.sort(reference_strings)

# create random query strings
query_strings_unicode = np.random.choice(
    np.arange(UNICODE_a, UNICODE_a + ALPHABET_SIZE, dtype=np.uint32),
    size=(n_query_string, string_length),
)
query_strings = query_strings_unicode.view(f"<U{string_length}").reshape(-1)

# JIT-compiled function
@njit
def query(sorted_reference_strings, query_strings):
    return np.searchsorted(sorted_reference_strings, query_strings)

# Execution
query(sorted_reference_strings, query_strings)
```

Error message:
```
TypingError                               Traceback (most recent call last)
Input In [1], in <cell line: 37>()
     32     return np.searchsorted(sorted_reference_strings, query_strings)
     35 # Execution
---> 37 query(sorted_reference_strings, query_strings)

File /opt/anaconda/envs/numba_test/lib/python3.10/site-packages/numba/core/dispatcher.py:468, in _DispatcherBase._compile_for_args(self, *args, **kws)
    464         msg = (f"{str(e).rstrip()} \n\nThis error may have been caused "
    465                f"by the following argument(s):\n{args_str}\n")
    466         e.patch_message(msg)
--> 468     error_rewrite(e, 'typing')
    469 except errors.UnsupportedError as e:
    470     # Something unsupported is present in the user code, add help info
    471     error_rewrite(e, 'unsupported_error')

File /opt/anaconda/envs/numba_test/lib/python3.10/site-packages/numba/core/dispatcher.py:409, in _DispatcherBase._compile_for_args.<locals>.error_rewrite(e, issue_type)
    407     raise e
    408 else:
--> 409     raise e.with_traceback(None)

No implementation of function Function(<function searchsorted at 0x7f67d40d37f0>) found for signature:
 
 >>> searchsorted(array([unichr x 32], 1d, C), array([unichr x 32], 1d, C))
 
  Overload in function 'searchsorted': File: numba/np/arraymath.py: Line 3773.
    With argument(s): '(array([unichr x 32], 1d, C), array([unichr x 32], 1d, C))':
   No implementation of function Function(<function _searchsorted.<locals>.searchsorted_inner at 0x7f67ae15e830>) found for signature:
    
    >>> searchsorted_inner(array([unichr x 32], 1d, C), [unichr x 32])
    
         Overload in function 'register_jitable.<locals>.wrap.<locals>.ov_wrap': File: numba/core/extending.py: Line 159.
           With argument(s): '(array([unichr x 32], 1d, C), [unichr x 32])':
          No implementation of function Function(<ufunc 'isnan'>) found for signature:
           
           >>> isnan([unichr x 32])
           
                Overload in function 'Numpy_rules_ufunc.generic': File: numba/core/typing/npydecl.py: Line 98.
                  With argument(s): '([unichr x 32])':
                   TypingError: can't resolve ufunc isnan for types [[unichr x 32]]
            raised from /opt/anaconda/envs/numba_test/lib/python3.10/site-packages/numba/core/typing/npydecl.py:107
          
          During: resolving callee type: Function(<ufunc 'isnan'>)
          During: typing of call at /opt/anaconda/envs/numba_test/lib/python3.10/site-packages/numba/np/arraymath.py (3746)
          
          
          File "../opt/anaconda/envs/numba_test/lib/python3.10/site-packages/numba/np/arraymath.py", line 3746:
              def searchsorted_inner(a, v):
                  <source elided>
                  n = len(a)
                  if np.isnan(v):
                  ^
   
     raised from /opt/anaconda/envs/numba_test/lib/python3.10/site-packages/numba/core/typeinfer.py:1086
   
   During: resolving callee type: Function(<function _searchsorted.<locals>.searchsorted_inner at 0x7f67ae15e830>)
   During: typing of call at /opt/anaconda/envs/numba_test/lib/python3.10/site-packages/numba/np/arraymath.py (3788)
   
   
   File "../opt/anaconda/envs/numba_test/lib/python3.10/site-packages/numba/np/arraymath.py", line 3788:
           def searchsorted_impl(a, v, side='left'):
               <source elided>
               for view, outview in np.nditer((v, out)):
                   index = loop_impl(a, view.item())
                   ^

  raised from /opt/anaconda/envs/numba_test/lib/python3.10/site-packages/numba/core/typeinfer.py:1086

During: resolving callee type: Function(<function searchsorted at 0x7f67d40d37f0>)
During: typing of call at /tmp/ipykernel_7977/1750162599.py (32)


File "../tmp/ipykernel_7977/1750162599.py", line 32:
```

Temporary Fix:
```python
def _searchsorted(func):
    def searchsorted_inner(a, v):
        n = len(a)
        # Commenting out this block solves the problem.
        # if np.isnan(v):
        #     # Find the first nan (i.e. the last from the end of a,
        #     # since there shouldn't be many of them in practice)
        #     for i in range(n, 0, -1):
        #         if not np.isnan(a[i - 1]):
        #             return i
        #     return 0
        lo = 0
        hi = n
        while hi > lo:
            mid = (lo + hi) >> 1
            if func(a[mid], (v)):
                # mid is too low => go up
                lo = mid + 1
            else:
                # mid is too high, or is a NaN => go down
                hi = mid
        return lo
    return searchsorted_inner
```

As I see it, there are couple of options here, but I'm open to better ideas:
1. Implement the `np.isnan` function for the unicode dtype. Needs thinking about because there is no null option for this type.
2. Put a wrapper around the `if` block which prevents calling `np.isnan` if the user has specified (through an option) that there are no null items in the sorted array.

I think 1. is preferable to keep the function as close to that in numpy as possible.

This is the first issue I've raised on an open-source project so please do give me some tips! I'm happy to get involved with the fix if somebody could give me some pointers.

stonescw:
Closes #8190. Fixes `searchsorted` so that it works for types where `np.isnan` is not implemented, motivated by the `unichr` type.

Summary of changes:
1. Add additional `if` statement to `searchsorted_inner` so that inserted item `v` is only checked using `np.isnan` when it is of the `float` or `complex` types.
2. Update `isinstance` implementation to support `UnicodeCharSeq`.
3. Test `searchsorted` using `unichr` type.

### DIFF
--- a/numba/cpython/builtins.py
+++ b/numba/cpython/builtins.py
@@ -750,7 +750,8 @@ def ol_isinstance(var, typs):
                         types.DictType, types.LiteralStrKeyDict, types.List,
                         types.ListType, types.Tuple, types.UniTuple, types.Set,
                         types.Function, types.ClassType, types.UnicodeType,
-                        types.ClassInstanceType, types.NoneType, types.Array)
+                        types.ClassInstanceType, types.NoneType, types.Array,
+                        types.UnicodeCharSeq)
     if not isinstance(var_ty, supported_var_ty):
         msg = f'isinstance() does not support variables of type "{var_ty}".'
         raise NumbaTypeError(msg)
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -3857,13 +3857,15 @@ def _searchsorted(func):
 
         .. [1] https://github.com/numpy/numpy/blob/809e8d26b03f549fd0b812a17b8a166bcd966889/numpy/core/src/npysort/binsearch.cpp#L173
         """  # noqa: E501
-        if np.isnan(v):
-            # Find the first nan (i.e. the last from the end of a,
-            # since there shouldn't be many of them in practice)
-            for i in range(n, 0, -1):
-                if not np.isnan(a[i - 1]):
-                    return i
-            return 0
+        # handle NaN where appropriate
+        if isinstance(v, (float, complex)):
+            if np.isnan(v):
+                # Find the first nan (i.e. the last from the end of a,
+                # since there shouldn't be many of them in practice)
+                for i in range(n, 0, -1):
+                    if not np.isnan(a[i - 1]):
+                        return i
+                return 0
 
         if v_last < v:
             hi = n
--- a/numba/tests/test_np_functions.py
+++ b/numba/tests/test_np_functions.py
@@ -1088,6 +1088,42 @@ class TestNPFunctions(MemoryLeakMixin, TestCase):
             # Sequence values
             check(a, list(values))
 
+        # Third with unichr (no NaNs)
+        unicode_a = 97
+        alphabet_size = 26
+        string_length = 8
+
+        np.random.seed(42)
+        bins_unicode = np.random.choice(
+            np.arange(
+                unicode_a,
+                unicode_a + alphabet_size,
+                dtype=np.uint32,
+            ),
+            size=(5, string_length),
+        )
+        bins_unsorted = bins_unicode.view(f"<U{string_length}").reshape(-1)
+        bins = np.sort(bins_unsorted)
+        values_unicode = np.random.choice(
+            np.arange(
+                unicode_a,
+                unicode_a + alphabet_size,
+                dtype=np.uint32,
+            ),
+            size=(20, string_length),
+        )
+        values = values_unicode.view(f"<U{string_length}").reshape(-1)
+
+        for a in (bins, list(bins)):
+            # Scalar values
+            for v in values:
+                check(a, v)
+            # Array values
+            for v in (values, values.reshape((4, 5))):
+                check(a, v)
+            # Sequence values
+            check(a, list(values))
+
         # nonsense value for 'side' raises TypingError
         def bad_side(a, v):
             return np.searchsorted(a, v, side='nonsense')


stuartarchibald:
@stonescw Thanks for the patch. Please ignore dependency related CI failures at present, the Numba maintainers are in the middle of building release candidates for Numba 0.56/llvmlite 0.39 and there's a small window (a few hours) during this time where the state of Numba and llvmlite are out of sync. When they are back in sync someone will rerun the checks.

numba/np/arraymath.py
```diff
@@ -3857,13 +3857,15 @@ def searchsorted_inner(a, v, v_last, lo, hi, n):
...
         """  # noqa: E501
-        if np.isnan(v):
-            # Find the first nan (i.e. the last from the end of a,
-            # since there shouldn't be many of them in practice)
-            for i in range(n, 0, -1):
-                if not np.isnan(a[i - 1]):
-                    return i
-            return 0
+        # handle NaN where appropriate
+        if isinstance(v, (float, complex)):
```
gmarkall, line 3861: @guilhermeleobas Does the `isinstance` check here look safe?

guilhermeleobas: @gmarkall Yes, it should be safe to use `isisntance` here

numba/tests/test_np_functions.py
```diff
@@ -1078,6 +1078,42 @@ def check(a, v):
...
+        bins = np.sort(bins_unsorted)
+        values_unicode = np.random.choice(
+            np.arange(
+                unicode_a,
+                unicode_a + alphabet_size,
+                dtype=np.uint32,
+            ),
+            size=(20, string_length),
+        )
+        values = values_unicode.view(f"<U{string_length}").reshape(-1)
```
guilhermeleobas, line 1115: Not necessary for this PR but perhaps one should split the entire function into different test cases.

gmarkall:
I think the RTD build failed because of a transient issue with the versions / tags - do we have a way to re-run this @esc?

numba/tests/test_np_functions.py
stuartarchibald, line 1115: Suggest making this a separate test and more simple as the logic being tested is essentially that the `np.isnan` branch is correctly elided? Would the following be sufficient to demonstrate the issue as in #8190? 
```python
from numba import njit
import numpy as np

@njit
def foo(a, b):
    return np.searchsorted(a, b)

a = np.array(['a','b','c'])
b = np.array('a')

foo(a, b)
```
Further, are there concerns over whether the underlying sorting algorithms will work correctly on unicode char arrays, if so, it seems appropriate to add this testing but it can probably be refactored to make use of things like the `string` module's `ascii_letters` etc?

numba/cpython/builtins.py
```diff
@@ -750,7 +750,8 @@ def false_impl(var, typs):
                         types.DictType, types.LiteralStrKeyDict, types.List,
                         types.ListType, types.Tuple, types.UniTuple, types.Set,
                         types.Function, types.ClassType, types.UnicodeType,
-                        types.ClassInstanceType, types.NoneType, types.Array)
+                        types.ClassInstanceType, types.NoneType, types.Array,
+                        types.UnicodeCharSeq)
```
stuartarchibald, line 754: It would be good to add a test for this additional type, perhaps something based on:
```python
from numba import njit
import numpy as np

@njit
def foo(a):
    unichr_inst = a.take(0)
    isinstance(unichr_inst, float)

a = np.array(['a']).take(0)

foo(a)
```
as a new test within https://github.com/numba/numba/blob/77882b19460110d28bcd617e1aca37a2899d0513/numba/tests/test_builtins.py#L1209

stuartarchibald (commented):
Thanks for the patch, I've made a couple of suggestions inline in relation to testing else looks good. Thanks again!

stuartarchibald:
> I think the RTD build failed because of a transient issue with the versions / tags - do we have a way to re-run this @esc?

Think `main` would need merging into this branch to pick up the version change fixes. IIRC readthedocs builds against HEAD of the branch without merging in `main` first.

esc:
> > I think the RTD build failed because of a transient issue with the versions / tags - do we have a way to re-run this @esc?
> 
> Think `main` would need merging into this branch to pick up the version change fixes. IIRC readthedocs builds against HEAD of the branch without merging in `main` first.

This is correct.

stonescw:
Would anybody object to my merging of main into this branch?

gmarkall:
> Would anybody object to my merging of main into this branch?

No, I'd encourage you to merge main into this branch :-)

PR #8203, opened 2022-06-28
pin llvmlite 0.39 for public CI builds

esc:
As title


esc:
This ensures that llvmlite 0.40.0dev0 isn't picked up by public CI.

stuartarchibald (approved):
Thanks for the patch.

PR #8208, opened 2022-06-29
Attempt to ensure that all C++ files build as C++11

gmarkall:
`typeconv.cpp` now uses C++11 features, but it manages to skip getting the `-std=c++11` flag on Darwin because it's built as an extension that includes both C and C++ files, so we can't just add the flag to `extra_compile_args`. Instead this attempts to sneak the flag into the C++ compiler specification just before extensions are built.

Testing on CI for now.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -106,6 +106,11 @@ class NumbaBuildExt(build_ext):
 
         super().run()
 
+    def build_extensions(self):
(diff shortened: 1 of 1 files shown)


gmarkall:
cc @dlee992

gmarkall:
I think the issues this attempted to solve were addressed by #8389.

PR #8209, opened 2022-06-29
Adds checklists for doing initial PR review.

stuartarchibald:
As title.

### DIFF
--- a/docs/source/developer/contributing.rst
+++ b/docs/source/developer/contributing.rst
@@ -6,6 +6,13 @@ We welcome people who want to make contributions to Numba, big or small!
 Even simple documentation improvements are encouraged.  If you have
 questions, don't hesitate to ask them (see below).
 
+Quick links for how you can help with the Numba project...
+
+* Contribute a patch (see: :ref:`getting set up <getting_set_up>`).
+* Review a patch (see:
+  :ref:`reviewing pull requests <reviewing_pull_requests>`).
+* Answer questions for other users on the discussion forum (see:
+  :ref:`Numba's discussion forum <discourse_forum>`).
 
 Communication
 -------------
@@ -27,6 +34,8 @@ developers to keep up with.
 
 Note that the Github issue tracker is the best place to report bugs.  Bug
 reports in chat are difficult to track and likely to be lost.
+
+.. _discourse_forum:
 
 Forum
 .....
@@ -61,6 +70,9 @@ please include specifics:
   results you are getting;
 * as far as possible, a code snippet that allows full reproduction of your
   problem.
+
+
+.. _getting_set_up:
 
 Getting set up
 --------------
@@ -240,8 +252,8 @@ Development rules
 Code reviews
 ''''''''''''
 
-Any non-trivial change should go through a code review by one or several of
-the core developers.  The recommended process is to submit a pull request
+All changes will go through a code review undertaken by the Numba community,
+including core developers. The recommended process is to submit a pull request
 on github.
 
 A code review should try to assess the following criteria:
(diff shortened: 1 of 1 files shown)


docs/source/developer/contributing.rst
```diff
@@ -422,6 +439,145 @@ as to free up core developer time. Examples of ways to help:
...
+        * [ ] The implementation has references to the region of the NumPy code
+              base that provides the same functionality. The references are URLs
+              of the form:
+              `https://github.com/numpy/numpy/tree/<SHA>/path/to/file.py:#<lines>`
+        * [ ] The code "looks" reasonably similar to the NumPy implementation.
+        * [ ] The code uses the same algorithm as the NumPy implementation, if
+              it does not it explains why and this justification is reasonable.
+        * [ ] The type inference part of the Numba implementation restricts the
+              types to those declared in the NumPy docs (or if not present, what
+              is accepted in the NumPy implementation itself).
```
esc, line 485: Probably worth adding that those restrictions are tested by supplying bogus types and checking that the error message is indeed the expected one.

docs/source/developer/contributing.rst
```diff
@@ -422,6 +439,145 @@ as to free up core developer time. Examples of ways to help:
...
+          please highlight this and one of the core developers will pick it up.)
+    * [ ] The proposed additional NumPy functionality is not deprecated by
+          NumPy.
+    * [ ] Public CI is passing.
+    * [ ] The patch does not touch files in `numba.core`.
+    * [ ] The patch is implemented using the high-level extension API only
+          (https://numba.readthedocs.io/en/stable/extending/high-level.html).
+    * [ ] The patch updates the documentation to reflect the new functionality.
+          (https://github.com/numba/numba/blob/main/docs/source/reference/numpysupported.rst)
+    * [ ] The patch is line wrapped to 80 characters.
```
stuartarchibald, line 471: Need to add a note about running a "spell checker", and making sure that NumPy/SciPy/Python are correctly capitalised.

sklam: hm... can we automate these type of checks as part of CI?

docs/source/developer/contributing.rst
```diff
@@ -422,6 +439,145 @@ as to free up core developer time. Examples of ways to help:
...
+              them to track their origin. The references are URLs of the form:
+              `https://github.com/numpy/numpy/tree/<SHA>/path/to/file.py:#<lines>`
+        * [ ] Use `unittest` style assertions (not pytest style!).
+        * [ ] Check that unsupported/invalid input types are handled as expected
+              using the `with self.assertRaises()` pattern.
+        * [ ] Explicitly check custom error messages are raised appropriately.
+        * [ ] Hit all the branches of the proprosed change (both at type
+              inference and run time).
+        * [ ] Check that numerically awkward values work as expected e.g.
+              `NaN`, `+/-Inf`, `0`, `INT_MAX`, `INT_MIN` etc.
```
sklam, line 508: I wonder if we should reference Boundary Value Testing in addition to this as this is a subset of BVT. BVT will also include empty containers.

sklam (commented):
This is very thorough, but i worry that it is so long it can be overwhelming. I don't think we need to mention things that are already automated. As for testing, would it be better if there's a link/page to explain our testing standard?

docs/source/developer/contributing.rst
```diff
@@ -422,6 +439,145 @@ as to free up core developer time. Examples of ways to help:
 
 The core developers thank everyone for their understanding with the above!
 
+.. _reviewing_pull_requests:
+
+Reviewing Pull Requests
+-----------------------
+
+Help with reviewing pull requests is welcomed. There are time-effort categorised
+review tasks available on this project board: **TODO: ADD LINK**.
```
esc, line 448:
```suggestion
review tasks available on the `Contributor self-service board <https://github.com/orgs/numba/projects/7>`_.
```

esc (changes requested):
Adding the link to the contributor self-service board.

PR #8216, opened 2022-07-02
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -3,8 +3,14 @@ on:
   schedule:
     - cron: '30 1 * * *'
 
+permissions:
+  contents: read
+
 jobs:
   stale:
+    permissions:
+      issues: write  # for actions/stale to close stale issues
+      pull-requests: write  # for actions/stale to close stale PRs
     runs-on: ubuntu-latest
     steps:
       - uses: actions/stale@v3


naveensrinivasan:
> We're really not sure what's going on here. Can you provide some detail about why this change is useful?

The summary is to set permissions for the Workflow.

The links in the PR should help with the details. HTH. Thanks

esc:
> Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

@naveensrinivasan thank you very much for bringing this to our attention, your input to the Numba project is appreciated. Thank you also for attached the links to the documentation. After reading through the information and discussing with my colleagues, it is not clear to me, what kind of malicious action this change would prevent from. I.e. what would be the consequence in Numba's case if this patch is not applied?

naveensrinivasan:
> > Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.
> 
> @naveensrinivasan thank you very much for bringing this to our attention, your input to the Numba project is appreciated. Thank you also for attached the links to the documentation. After reading through the information and discussing with my colleagues, it is not clear to me, what kind of malicious action this change would prevent from. I.e. what would be the consequence in Numba's case if this patch is not applied?

It is a good practice as it helps future changes to workflow not be compromised. If you don't think it is beneficial I can close it. Thanks.

esc:
@naveensrinivasan thank you. After some OOB deliberation amongs developers, we decided to accept this patch!

sklam (approved):

PR #8219, opened 2022-07-04
Permit use of 'literal_unroll' via getattr on 'numba' module.

### LINKED ISSUE #5344: literal_unroll != numba.literal_unroll?
Hi, I'm seeing a strange behaviour, confirmed in two v0.48 environments. The same function behaves differently when using `literal_unroll` or `numba.literal_unroll`.

```python
def make_foo():
    x = (1, 2)
    @njit
    def _foo_unrolled_global():
        a = 0
        for val in numba.literal_unroll(x):
            a += val
        return a
    return _foo_unrolled_global


def make_bar():
    x = (1, 2)
    @njit
    def _bar_unrolled_global():
        a = 0
        for val in literal_unroll(x):
            a += val
        return a
    return _bar_unrolled_global

foo = make_foo()
foo()
# OK

bar = make_bar()
bar()
UnsupportedError: Failed in nopython mode pipeline (step: handles literal_unroll)
Failed in literal_unroll_subpipeline mode pipeline (step: switch const list for tuples)
Invalid use of literal_unroll, argument should be a tuple or a list of constant values, found unknown problem

```

I know the second error is related (and fixed by) #5182, so it should not fail in v0.49, but the puzzle is why the first one does not fail in v0.48.

cc: @stuartarchibald

stuartarchibald:
As title.

Fixes #5344

### DIFF
--- a/numba/core/untyped_passes.py
+++ b/numba/core/untyped_passes.py
@@ -21,6 +21,7 @@ from numba.core.ir_utils import (guard, resolve_func_from_module, simplify_CFG,
                                  )
 from numba.core.ssa import reconstruct_ssa
 from numba.core import interpreter
+import numba # Need this to check for attributes on the Numba module itself.
 
 
 @contextmanager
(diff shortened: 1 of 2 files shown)


numba/core/untyped_passes.py
```diff
@@ -626,109 +627,123 @@ def run_pass(self, state):
...
+                            if got_attr.value == numba:
+                                glbl = literal_unroll
+                else:
+                    if isinstance(glbl_tmp, (ir.Global, ir.FreeVar)):
+                        glbl = glbl_tmp.value
+
+                # find a literal_unroll
+                if glbl is literal_unroll:
+
+                    if len(call.args) > 1:
```
stuartarchibald, line 648: NOTE: most of the diff from here onward is just code motion, the block has one less level of indentation.

numba/tests/test_mixed_tuple_unroller.py
```diff
@@ -1246,6 +1326,7 @@ def foo():
                     acc -= 1
                 else:
                     break
+            _check_is_tuple(a)
```
stuartarchibald, line 1329: NOTE: This check is added to all tests to ensure transformation of const-list to `tuple` occurs. Prior to this the tests would pass regardless of whether `literal_unroll` was used as the list type used in the test would generally resolve to e.g. `list(complex128)` and work fine in operations such as reduction.

numba/core/untyped_passes.py
```diff
@@ -626,109 +627,123 @@ def run_pass(self, state):
...
-                            if to_unroll_lhs is None:
+                glbl = None
+                glbl_tmp = guard(get_definition, func_ir, call.func)
+                if isinstance(glbl_tmp, ir.Expr) and glbl_tmp.op == "getattr":
+                    if glbl_tmp.attr == "literal_unroll":
+                        # resolve the getattr, make sure it's getting
+                        # attr "literal_unroll" from the numba module.
+                        got_attr = guard(get_definition, func_ir,
+                                         glbl_tmp.value)
+                        if isinstance(got_attr, (ir.Global, ir.FreeVar)):
```
sklam, line 638: should there be a `else` branch that raise unreachable error? i can't imagine `got_attr` being anything but `(ir.Global, ir.FreeVar)`?

numba/core/untyped_passes.py
```diff
@@ -1079,7 +1094,17 @@ def get_call_args(init_arg, want):
...
                     continue
-                call_func = call_func.value
+                if isinstance(call_func, ir.Expr) and call_func.op == "getattr":
+                    if call_func.attr == "literal_unroll":
+                        # resolve the getattr, make sure it's getting
+                        # attr "literal_unroll" from the numba module.
+                        got_attr = guard(get_definition, func_ir,
+                                         call_func.value)
+                        if isinstance(got_attr, (ir.Global, ir.FreeVar)):
+                            if got_attr.value == numba:
```
sklam, line 1104: same with these two branches? should there be a unreachable error else branch?

sklam (commented):
There is a lot of `isinstance(x, (ir.Global, ir.FreeVar))` and `x.value == numba` and `x.attr == 'literal_unroll' scatter in this patch. Can all of these be consolidated to avoid code duplication?

PR #8220, opened 2022-07-04
Added the interval example as doctest

kc611:
Adds the `Interval` example as doctest.

### DIFF
--- a/docs/source/extending/interval-example.rst
+++ b/docs/source/extending/interval-example.rst
@@ -3,7 +3,7 @@ Example: An Interval Type
 =========================
 
 In this example, we will extend the Numba frontend to add support for a user-defined
-class that it does not internally supported. This will allow:
+class that it does not internally support. This will allow:
 
 * Passing an instance of the class to a Numba function
 * Accessing attributes of the class in a Numba function
--- a/docs/source/extending/interval-example.rst
+++ b/docs/source/extending/interval-example.rst
@@ -1,9 +1,9 @@
 
-Example: an interval type
+Example: An Interval Type
 =========================
 
-We will extend the Numba frontend to support a class that it does not
-currently support so as to allow:
+In this example, we will extend the Numba frontend to add support for a user-defined
+class that it does not internally supported. This will allow:
 
 * Passing an instance of the class to a Numba function
 * Accessing attributes of the class in a Numba function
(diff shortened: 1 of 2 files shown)


numba/tests/doc_examples/test_interval_example.py
```diff
@@ -0,0 +1,221 @@
...
+
+        def early_exit_if_null(builder, stack, obj):
+            return early_exit_if(builder, stack, cgutils.is_null(builder, obj))
+
+        @unbox(IntervalType)
+        def unbox_interval(typ, obj, c):
+            """
+            Convert a Interval object to a native interval structure.
+            """
+            is_error_ptr = cgutils.alloca_once_value(c.builder,
```
kc611, line 136: The original `fail_obj` logic over here that replaced the object by a null object of the same type wasn't working. Not sure if that was how to do it or if it's required, pseudocode:

```
       ret_ptr = cgutils.alloca_once(c.builder, c.context.get_value_type(typ))
       fail_obj = c.context.get_constant_null(typ)

       with ExitStack() as stack:
           if fail:
               c.builder.store(fail_obj, ret_ptr)
           interval = cgutils.create_struct_proxy(typ)(c.context, c.builder)
           c.builder.store(interval._getvalue(), ret_ptr)
       return NativeValue(c.builder.load(ret_ptr), is_error=c.builder.load(is_error_ptr))
```

numba/tests/doc_examples/test_interval_example.py
```diff
@@ -0,0 +1,221 @@
...
+                                                     cgutils.false_bit)
+
+            with ExitStack() as stack:
```
kc611, line 139: I remember @stuartarchibald suggesting we could have a generalized solution for such Exit-if-null strategy. The main issue with this particular approach is that we can't abstract out the context definitions and exits. 

Is there anything that comes to mind regarding how we could do this with a more simpler method ?

eric-wieser:
Any chance you could split this PR into two: the doc-test-ification, and the the change I suggested in the other PR? It would make it easier to see the effect of my changes.

kc611:
Good idea. It'll be easier to review too.

docs/source/extending/interval-example.rst
```diff
@@ -1,9 +1,9 @@
 
-Example: an interval type
+Example: An Interval Type
 =========================
 
-We will extend the Numba frontend to support a class that it does not
-currently support so as to allow:
+In this example, we will extend the Numba frontend to add support for a user-defined
+class that it does not internally supported. This will allow:
```
eric-wieser, line 6:
```suggestion
class that it does not internally support. This will allow:
```

numba/tests/doc_examples/test_interval_example.py
```diff
@@ -0,0 +1,183 @@
+# "magictoken" is used for markers as beginning and ending of example text.
```
eric-wieser, line 1:
```suggestion
"""
This test is used by `docs/source/extending/interval-example.rst`.

The "magictoken" comments are used as markers for the beginning and ending of example code.
"""
```

eric-wieser (dismissed):
I haven't checked that the old and new python code are identical, but this looks like a very nice way to extract a doctest.

esc:
@kc611 is there any chance you could include @eric-wieser 's original commit as a basis, in order to preserve their attribution?

eric-wieser:
I don't think any of this patch is my contribution at this point, only the next one is

numba/tests/doc_examples/test_interval_example.py
```diff
@@ -0,0 +1,187 @@
+"""
+This test is used by `docs/source/extending/interval-example.rst`.
+
+The "magictoken" comments are used as markers for the beginning and ending of example code.
```
esc, line 4: Sphinx is complaing here:

numba/tests/doc_examples/test_interval_example.py:4:81: E501 line too long (91 > 80 characters)

```suggestion
The "magictoken" comments are used as markers for the beginning and ending of
example code.
```

esc (changes requested):
just some final flake8 nitpick

esc:
@kc611 thank you for the patch!

esc (approved):

PR #8228, opened 2022-07-06
Fix typo in @vectorize docstring and a NumPy spelling.

### LINKED ISSUE #8227: vectorize docstring example has incorrect identity
## Reporting a bug

This is a documentation bug more than a functionality bug, so reproducibility does not exactly apply.

The docstring example in `vectorize` provides a ufunc of `sum` that has `identity=1`.  I believe this should be `0` (additive identity):

https://github.com/numba/numba/blob/15e48788660f5c387d8d1160cc4be5d2363510fc/numba/np/ufunc/decorators.py#L100-L103

The later example for `mul` correctly sets `identity=1`.

- [ ] ~I have included a self contained code sample to reproduce the problem.
  i.e. it's possible to run as 'python bug.py'.~ N/A

stuartarchibald:
As title.

Fixes #8227

### DIFF
--- a/numba/np/ufunc/decorators.py
+++ b/numba/np/ufunc/decorators.py
@@ -58,7 +58,7 @@ class GUVectorize(_BaseVectorize):
 def vectorize(ftylist_or_function=(), **kws):
     """vectorize(ftylist_or_function=(), target='cpu', identity=None, **kws)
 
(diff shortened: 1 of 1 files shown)


gmarkall (approved):
Looks good.

PR #8235, opened 2022-07-08
Make the NRT stats counters optional.

### LINKED ISSUE #8156: Make NRT stats counters optional, off by default
## Reporting a bug

As discussed at last week's public Numba meeting, https://github.com/numba/numba/wiki/Minutes_2022_06_07. Code where:
* there are potentially quite a few allocations/deallocations in a function, often caused by array temporaries 

and
* the function is called by many native threads

results in a lot of pressure on the atomic stats counters defined here:
https://github.com/numba/numba/blob/080633aa333e69781888eee6521f5e2fa6751c75/numba/core/runtime/nrt.c#L53
and their use as is commented out here:
```diff
diff --git a/numba/core/runtime/nrt.c b/numba/core/runtime/nrt.c
index 3a65c9b..213251a 100644
--- a/numba/core/runtime/nrt.c
+++ b/numba/core/runtime/nrt.c
@@ -182,7 +182,7 @@ void NRT_MemInfo_init(NRT_MemInfo *mi,void *data, size_t size,
     mi->external_allocator = external_allocator;
     NRT_Debug(nrt_debug_print("NRT_MemInfo_init mi=%p external_allocator=%p\n", mi, external_allocator));
     /* Update stats */
-    TheMSys.atomic_inc(&TheMSys.stats_mi_alloc);
+//     TheMSys.atomic_inc(&TheMSys.stats_mi_alloc);
 }
 
 NRT_MemInfo *NRT_MemInfo_new(void *data, size_t size,
@@ -321,7 +321,7 @@ void NRT_dealloc(NRT_MemInfo *mi) {
     NRT_Debug(nrt_debug_print("NRT_dealloc meminfo: %p external_allocator: %p\n", mi, mi->external_allocator));
     if (mi->external_allocator) {
         mi->external_allocator->free(mi, mi->external_allocator->opaque_data);
-        TheMSys.atomic_inc(&TheMSys.stats_free);
+//         TheMSys.atomic_inc(&TheMSys.stats_free);
     } else {
         NRT_Free(mi);
     }
@@ -329,7 +329,7 @@ void NRT_dealloc(NRT_MemInfo *mi) {
 
 void NRT_MemInfo_destroy(NRT_MemInfo *mi) {
     NRT_dealloc(mi);
-    TheMSys.atomic_inc(&TheMSys.stats_mi_free);
+//     TheMSys.atomic_inc(&TheMSys.stats_mi_free);
 }
 
 void NRT_MemInfo_acquire(NRT_MemInfo *mi) {
@@ -472,7 +472,7 @@ void* NRT_Allocate_External(size_t size, NRT_ExternalAllocator *allocator) {
         ptr = TheMSys.allocator.malloc(size);
         NRT_Debug(nrt_debug_print("NRT_Allocate_External bytes=%zu ptr=%p\n", size, ptr));
     }
-    TheMSys.atomic_inc(&TheMSys.stats_alloc);
+//     TheMSys.atomic_inc(&TheMSys.stats_alloc);
     return ptr;
 }
 
@@ -486,7 +486,7 @@ void *NRT_Reallocate(void *ptr, size_t size) {
 void NRT_Free(void *ptr) {
     NRT_Debug(nrt_debug_print("NRT_Free %p\n", ptr));
     TheMSys.allocator.free(ptr);
-    TheMSys.atomic_inc(&TheMSys.stats_free);
+//     TheMSys.atomic_inc(&TheMSys.stats_free);
 }
```

I think this code that follows reproduces the situation described above (and at the public meeting) that applies pressure on the atomic counters (using the `count_muons` example as given at the public meeting):

```python
import ctypes
import numba
from numba import cfunc, carray, types, njit, prange
from numba.extending import intrinsic
import numpy as np
from timeit import default_timer


@cfunc(types.intp(types.CPointer(types.float32)))
def count_muons(ptr):
    arr = carray(ptr, 10)
    return np.count_nonzero((arr > 1.) & (np.abs(arr) < 7.) & (arr > 0.))


x = np.arange(10, dtype=np.float32)

count = count_muons(x.ctypes.data_as(ctypes.POINTER(ctypes.c_float)))
print("cfunc call", count)


@intrinsic
def addr_as_float32ptr(tyctx, addr):
    sig = types.CPointer(types.float32)(types.intp)

    def codegen(cgctx, builder, sig, llargs):
        ret = builder.inttoptr(llargs[0], cgctx.get_value_type(sig.return_type))
        return ret
    return sig, codegen


@njit
def liveness_sink(*args):
    # keeps things that are only referenced by pointers alive
    pass


@njit(parallel=True)
def foo(n):
    acc0 = acc1 = acc2 = acc3 = 0
    for i in prange(n):
        input_arr = np.arange(10).astype(np.float32)
        addr = input_arr.ctypes.data
        ptr = addr_as_float32ptr(addr)
        acc0 += count_muons(ptr)
        acc1 += count_muons(ptr)
        acc2 += count_muons(ptr)
        acc3 += count_muons(ptr)
        liveness_sink(input_arr)
    return acc0 + acc1 + acc2 + acc3


# make this bigger to run more arrays
n = 4000000

tstart = default_timer()
out = foo(n)
tend = default_timer()
print("jit call", out)
assert out == n * 5 * 4 # n loops with 5 things counted in 4 accumulators

print(f"Elapsed time: {tend-tstart}")
```

Without the patch above the result is:
```
Elapsed time: 19.056538556000305
```
but with the patch above (i.e. atomic counters are disabled)
```
Elapsed time: 6.221216844001901
```
It will probably be easier to fix this after https://github.com/numba/numba/pull/8106 is merged.

CC @jpivarski

stuartarchibald:
As title. They can be switched on/off with env var:
`NUMBA_NRT_STATS`. Testing is updated to ensure that they are on
when running CI/conda build test phase.

Fixes #8156

~Depends on https://github.com/numba/numba/pull/8106, only the final patch (b3c7d18) is new.~ Rebased, dependency resolved, a73612b is b3c7d18.


gmarkall (commented):
The code changes look fine to me.

I have one query, which is about the use of the canary value for stats values when counting is disabled - wouldn't it be be safer / more Pythonic to raise a Python exception if the counter values are requested when counting is disabled? Presently the risk is that an invalid value is returned unnoticed - or, it needs to be checked for and dealt with anyway.

gmarkall:
/azp run

(Checking if the test running issues are resolved)

stuartarchibald:
> The code changes look fine to me.
> 
> I have one query, which is about the use of the canary value for stats values when counting is disabled - wouldn't it be be safer / more Pythonic to raise a Python exception if the counter values are requested when counting is disabled? Presently the risk is that an invalid value is returned unnoticed - or, it needs to be checked for and dealt with anyway.

From an OOB conversation, it was decided that the more Pythonic thing is to raise a Python exception if the counter values are requested when counting is disabled. This however would break the test suite out-of-the-box as there's a lot of tests that check for reference leaks and need the stats counters to be enabled. This is addressed by updating the test suite to enable the stats counters if the reference leak checking mixin is used, it also adds a new mixin to enable the stats counters as some tests use them for other purposes.

gmarkall (approved):
Thanks for the updates - I think this now looks good. As agreed per OOB discussion, the best path is for the `EnableNRTStatsMixin` to always disable stats counters when it tears down - generally stats counters should be off, and any test that requires them should be using the mixin to make sure they're on - therefore the current state is most likely to catch any errors in the expectations of tests about the enablement of counters.

@esc / @stuartarchibald could this have a full buildfarm run please?

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cpu_yaml_117`.

Passed.

Issue #6448, opened 2020-11-03
CUDA: Add accessors for function attributes beyond just register use per thread


gmarkall:
PR #6447 adds a public API to get the maximum number of registers per thread (`numba.cuda.Dispatcher.get_regs_per_thread()`). There are other attributes that might be nice to provide - shared memory per block, local memory per thread, const memory usage, maximum block size.

These are all available in the `FuncAttr` named tuple: https://github.com/numba/numba/blob/master/numba/cuda/cudadrv/driver.py#L2023, set by `_read_func_attr_all`: https://github.com/numba/numba/blob/master/numba/cuda/cudadrv/driver.py#L2067

Cross ref: adding these would provide some of the info in the ptxas output requested in #3482.

edwinlim0919:
Hi Graham! I'm a student @ The University of California, Berkeley and this is my first time contributing to open source. I am interested in addressing this issue.

tejash5489-lang:
It looks like this has already been fully addressed by follow-on PRs to #6447, which just weren't linked back to this issue:

- \`get_shared_mem_per_block\` — #8240 (merged)
- \`get_const_mem_size\` — #8295 (merged)
- \`get_local_mem_per_thread\` — #8315 (merged)
- \`get_max_threads_per_block\` — #8348 (merged)

All four are present on \`Dispatcher\` in current \`main\` alongside \`get_regs_per_thread\`. This can probably be closed.

PR #8240, opened 2022-07-13
Add get_shared_mem_per_block method to Dispatcher 

testhound:
This pull request partially address issue: https://github.com/numba/numba/issues/6448 by adding support for retrieving shared memory per block from the Dispatcher.


testhound:
gpuci run tests

testhound:
gpuci run tests

testhound:
gpuci run tests

numba/cuda/tests/cudapy/test_dispatcher.py
```diff
@@ -272,6 +272,78 @@ def pi_sin_array(x, n):
         self.assertIsInstance(regs_per_thread, int)
         self.assertGreater(regs_per_thread, 0)
 
+    def test_get_shared_mem_per_block_unspecialized(self):
```
gmarkall, line 275: A few comments for this test:

- There are some copy / paste errors that cause it to refer to registers per thread.
- The test now doesn't have one kernel with two different specializations, but instead uses two separate jitted functions. I think this deviates from the pattern copied from above, to test multiple specializations of one kernel.
- There is some hardcoding of the `N` value in the kernel.
- The test code is a little more complex than necessary - the minimum we need is something where different specializations have different shared memory usages.

I was experimenting locally with this test, so I would suggest you could just apply the following to resolve all the above:

```diff
diff --git a/numba/cuda/tests/cudapy/test_dispatcher.py b/numba/cuda/tests/cudapy/test_dispatcher.py
index 1376ca950..2db9b0996 100644
--- a/numba/cuda/tests/cudapy/test_dispatcher.py
+++ b/numba/cuda/tests/cudapy/test_dispatcher.py
@@ -273,60 +273,43 @@ class TestDispatcher(CUDATestCase):
         self.assertGreater(regs_per_thread, 0)
 
     def test_get_shared_mem_per_block_unspecialized(self):
-        # A kernel where the register usage per thread is likely to differ
-        # between different specializations
-        @cuda.jit
-        def simple_smem_float(ary):
-            sm = cuda.shared.array(100, dtype=float32)
-            i = cuda.grid(1)
-            if i == 0:
-                for j in range(100):
-                    sm[j] = j
-            cuda.syncthreads()
-            ary[i] = sm[i]
+        N = 10
 
+        # A kernel where the shared memory per block is likely to differ
+        # between different specializations
         @cuda.jit
-        def simple_smem_double(ary):
-            sm = cuda.shared.array(100, dtype=float64)
-            i = cuda.grid(1)
-            if i == 0:
-                for j in range(100):
-                    sm[j] = j
-            cuda.syncthreads()
-            ary[i] = sm[i]
+        def simple_smem(ary):
+            sm = cuda.shared.array(N, dtype=ary.dtype)
+            for j in range(N):
+                sm[j] = j
+            for j in range(N):
+                ary[j] = sm[j]
 
         # Call the kernel with different arguments to create two different
         # definitions within the Dispatcher object
-        N = 100
         arr_f32 = np.zeros(N, dtype=np.float32)
         arr_f64 = np.zeros(N, dtype=np.float64)
 
-        simple_smem_float[1, N](arr_f32)
-        simple_smem_double[1, N](arr_f64)
+        simple_smem[1, 1](arr_f32)
+        simple_smem[1, 1](arr_f64)
 
         sig_f32 = void(float32[::1])
         sig_f64 = void(float64[::1])
-        shared_mem_per_block_f32 = \
-            simple_smem_float.get_shared_mem_per_block(sig_f32)
-        shared_mem_per_block_f64 = \
-            simple_smem_double.get_shared_mem_per_block(sig_f64)
-        self.assertIsInstance(shared_mem_per_block_f32, int)
-        self.assertIsInstance(shared_mem_per_block_f64, int)
+        sh_mem_f32 = simple_smem.get_shared_mem_per_block(sig_f32)
+        sh_mem_f64 = simple_smem.get_shared_mem_per_block(sig_f64)
+        self.assertIsInstance(sh_mem_f32, int)
+        self.assertIsInstance(sh_mem_f64, int)
 
-        self.assertEqual(shared_mem_per_block_f32, 400)
-        self.assertEqual(shared_mem_per_block_f64, 800)
+        self.assertEqual(sh_mem_f32, N * 4)
+        self.assertEqual(sh_mem_f64, N * 8)
 
-        # Check that getting the shared memory per blockfor all signatures
+        # Check that getting the shared memory per block for all signatures
         # provides the same values as getting the shared mem per block for
         # individual signatures.
-        shared_mem_per_block_f32_all = \
-            simple_smem_float.get_shared_mem_per_block()
-        shared_mem_per_block_f64_all = \
-            simple_smem_double.get_shared_mem_per_block()
-        self.assertEqual(shared_mem_per_block_f32_all[sig_f32.args],
-                         shared_mem_per_block_f32)
-        self.assertEqual(shared_mem_per_block_f64_all[sig_f64.args],
-                         shared_mem_per_block_f64)
+        sh_mem_f32_all = simple_smem.get_shared_mem_per_block()
+        sh_mem_f64_all = simple_smem.get_shared_mem_per_block()
+        self.assertEqual(sh_mem_f32_all[sig_f32.args], sh_mem_f32)
+        self.assertEqual(sh_mem_f64_all[sig_f64.args], sh_mem_f64)
 
     def test_get_shared_mem_per_block_specialized(self):
         @cuda.jit(void(float32[::1]))
```

numba/cuda/tests/cudapy/test_dispatcher.py
```diff
@@ -272,6 +272,78 @@ def pi_sin_array(x, n):
...
+        def simple_smem(ary):
+            sm = cuda.shared.array(100, dtype=float32)
+            i = cuda.grid(1)
+            if i == 0:
+                for j in range(100):
+                    sm[j] = j
+            cuda.syncthreads()
+            ary[i] = sm[i]
+
+        # Check we get a positive integer for the specialized variation
```
gmarkall, line 342: I think we know what shared memory usage to expect (unlike the register usage test above) so we should be able to assert that we have exactly the shared memory usage we expect.

gmarkall (changes requested):
Many thanks for the PR - the implementation looks fine. There seems to be an issue with the docs generation, possibly because the implementation has no docstring. There are some suggestions for the tests on the diff as well.

testhound:
@gmarkall addressed all issues that were commented on.

testhound:
gpuci run tests

numba/cuda/tests/cudapy/test_dispatcher.py
```diff
@@ -272,6 +272,61 @@ def pi_sin_array(x, n):
...
```
gmarkall, line 325: I don't think this comment aligns with what the test is doing, and now the test is more straightforward (a direct check of the exact shared memory usage) it seems like a comment shouldn't be needed.
```suggestion
```

docs/source/cuda-reference/kernel.rst
```diff
@@ -58,7 +58,8 @@ creating a specialized instance:
 
 .. autoclass:: numba.cuda.dispatcher.CUDADispatcher
    :members: inspect_asm, inspect_llvm, inspect_sass, inspect_types,
-             get_regs_per_thread, specialize, specialized, extensions, forall
+             get_regs_per_thread, specialize, specialized, extensions, forall,
+             get_shared_mem_per_block
```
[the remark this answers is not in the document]
gmarkall: Just checked, this now renders as expected.

gmarkall (changes requested):
Many thanks for the updates. This now looks better - there is just one suggestion on the diff, then I think this will be good.

gmarkall:
The test runner for Linux py38_np119_tbb seems to have had a transient issue, so I think it can be ignored. I won't kick off the pipeline again because another commit will probably added that will trigger it anyway.

testhound:
gpuci run tests

testhound:
@gmarkall addressed the spurious comment.

gmarkall (approved):
Many thanks for the update - I think this looks good now. I think the CI issues are only due to the current general outage.

gmarkall:
gpuci run tests

stuartarchibald (approved):

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_143`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_143`.

Passed.

PR #8241, opened 2022-07-13
Reorder typeof checks to avoid infinite loops on StructrefProxy  __hash__

DannyWeitekamp:
I was having an issue where I couldn't implement `__hash__` for a StructRefProxy by calling into a dispatcher that takes the structref as an argument because the dispatcher calls typeof on the proxy instance which in turn calls cffi_utils.is_cffi_func() which calls `__hash__` on the instance creating an infinite loop. 

I'm not sure how often typeof is called (I suspect a lot), but my guess is that putting the `_numba_type_` check first might provide a small compile-time performance benefit. I would imagine objects with `_numba_type_` are more common than cffi functions and thus should be checked first. 

I don't have a ton of time, but I'm happy to add a test if ya'll think it would be worth it.

### DIFF
--- a/numba/core/typing/typeof.py
+++ b/numba/core/typing/typeof.py
@@ -53,7 +53,7 @@ def typeof_impl(val, c):
         return tp
 
     tp = getattr(val, "_numba_type_", None)
-    if(tp is not None):
+    if tp is not None:
         return tp
 
     # cffi is handled here as it does not expose a public base class
--- a/numba/core/typing/typeof.py
+++ b/numba/core/typing/typeof.py
@@ -52,6 +52,10 @@ def typeof_impl(val, c):
     if tp is not None:
         return tp
 
+    tp = getattr(val, "_numba_type_", None)
+    if(tp is not None):
+        return tp
+
     # cffi is handled here as it does not expose a public base class
     # for exported functions or CompiledFFI instances.
     from numba.core.typing import cffi_utils
@@ -61,7 +65,7 @@ def typeof_impl(val, c):
         if cffi_utils.is_ffi_instance(val):
             return types.ffi
 
-    return getattr(val, "_numba_type_", None)
+    return None
 
 
 def _typeof_buffer(val, c):
--- a/numba/tests/test_struct_ref.py
+++ b/numba/tests/test_struct_ref.py
@@ -230,8 +230,8 @@ class TestStructRefBasic(MemoryLeakMixin, TestCase):
             td['b'] = MyStruct(2.3, 1)
 
     def test_MyStructType_hash_no_typeof_recursion(self):
-        '''Tests that __hash__ is not called prematurely in typeof 
-           causing infinite recursion (see #8241).'''
+        # Tests that __hash__ is not called prematurely in typeof
+        # causing infinite recursion (see #8241).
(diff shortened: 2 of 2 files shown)


stuartarchibald (commented):
Thanks for the patch. The proposed change seems like a good idea in general, `typeof` does indeed get called a lot and CFFI use is probably less likely than something with a `_numba_type_` attribute! I think adding a test would be a good idea as it would help prevent potential future changes causing the same problem.

I presume the issue you saw was something along these lines:
```python
from numba import njit, typeof
from numba.core import types


@structref.register
class MyStructType(types.StructRef):
    def preprocess_fields(self, fields):
        return tuple()


class MyStruct(structref.StructRefProxy):
    def __new__(cls):
        return structref.StructRefProxy.__new__(cls)

    def __hash__(self):
        return MyStructHashImpl(self)


@njit
def MyStructHashImpl(self):
    return 10


structref.define_proxy(MyStruct, MyStructType, [])


struct_inst = MyStruct()
typeof(struct_inst)
```
which overflows the stack due to recursion, if so, perhaps this could be used as a basis for a test?

Many thanks.

numba/core/typing/typeof.py
```diff
@@ -52,6 +52,10 @@ def typeof_impl(val, c):
     if tp is not None:
         return tp
 
+    tp = getattr(val, "_numba_type_", None)
+    if(tp is not None):
```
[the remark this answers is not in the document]
DannyWeitekamp: Incorperated in commit below.

gmarkall:
Also needs a quick flake8 fix.

numba/tests/test_struct_ref.py
```diff
@@ -226,6 +229,14 @@ def test_MyStructType_in_dict_mixed_type_error(self):
             # the second field is now an integer.
             td['b'] = MyStruct(2.3, 1)
 
+    def test_MyStructType_hash_no_typeof_recursion(self):
+        '''Tests that __hash__ is not called prematurely in typeof 
+           causing infinite recursion (see #8241).'''
```
stuartarchibald, line 234:
```suggestion
        # Tests that __hash__ is not called prematurely in typeof
        # causing infinite recursion (see #8241).
```
docstrings end up polluting Numba's test runner output in verbose mode (also fixes `flake8` issue).

numba/tests/test_struct_ref.py
```diff
@@ -226,6 +229,14 @@ def test_MyStructType_in_dict_mixed_type_error(self):
...
+        st = MyStruct(1, 2)
+        typeof(st)
+
+        assert hash(st) == 3
```
stuartarchibald, line 238:
```suggestion
        self.assertEqual(hash(st), 3)
```
Numba uses `unittest` style assertions.

stuartarchibald (commented):
Thanks for the patch, two minor suggestions/fixes else looks good.

numba/tests/test_struct_ref.py
DannyWeitekamp: Ahh getting too used to pytest.

stuartarchibald (approved):

esc:
Thank you for the patch!

PR #8243, opened 2022-07-14
Add a note to `reference/numpysupported.rst`

### LINKED ISSUE #2787: Issue with np.concatenate
Hello,

I'm working with Python 3.6, Numpy 1.14.1 and Numba 0.37.0

I have to tile up a 2-D array, and since  the Numpy concatenate function is supported by Numba, I'm trying to use it. But I'm not able to make it work.

Here is an example:

    pts=np.random.randn(1000, 2)
    @jit(nopython=True)
    def myf(points,  n_begin, n_end):
        expd=np.expand_dims(points,2)
        tiled=np.concatenate([expd for i in range(n_end-n_begin)], axis=2)

        return tiled

    myf(pts, 1, 3)

Here the error I get:

    TypingError: Failed at nopython (nopython frontend)
    Invalid usage of Function(<built-in function concatenate>) with parameters (list(array(float64, 3d, A)), axis=int64)
     * parameterized
    In definition 0:
        TypeError: np.concatenate(): expecting a non-empty tuple of arrays, got list(array(float64, 3d, A))
        raised from /usr/local/lib/python3.6/site-packages/numba/typing/npydecl.py:789
    [1] During: resolving callee type: Function(<built-in function concatenate>)
    [2] During: typing of call at <ipython-input-8-b567293363b0> (8)

Can you tell me where I'm doing wrong?

Thank you!

P.S.
this is the workaround I am using while understanding where I'm wrong

    pts=np.random.randn(1000, 2)
    @jit(nopython=True)
    def myf(points,  n_begin, n_end):
        tiled = np.zeros((points.shape[0], points.shape[1], n_end-n_begin))
        for dim in range(n_end-n_begin):
            tiled[:,:,dim]=points

        return tiled

    myf(pts, 1, 3)

sasakamaq:
Add a note to the line of `numpy.concatenate` that only supports tuple arguments. Fixes #2787.

### DIFF
--- a/docs/source/reference/numpysupported.rst
+++ b/docs/source/reference/numpysupported.rst
@@ -439,7 +439,7 @@ The following top-level functions are supported:
 * :func:`numpy.broadcast_arrays` (only the first argument)
 * :func:`numpy.broadcast_shapes`
 * :func:`numpy.column_stack`
-* :func:`numpy.concatenate`
+* :func:`numpy.concatenate` (only supports tuple arguments)
 * :func:`numpy.convolve` (only the 2 first arguments)
 * :func:`numpy.copy` (only the first argument)
 * :func:`numpy.corrcoef` (only the 3 first arguments, requires SciPy)


stuartarchibald (approved):
Thanks for the patch!

PR #8247, opened 2022-07-16
Fix issue 8127

### LINKED ISSUE #8127: No out of bounds check during advanced array indexing

bszollosinagy:
Fixes an issue with bounds checking when using advanced indexing, as described in [Issue 8127](https://github.com/numba/numba/issues/8127)

Added a unit test to reproduce the issue. The unit test looks to see if IndexError assertion is raised or not. It fails without the modification in numba\np\arrayobj.py, and passes if you use the version in this pull request.

Apparently the issue only manifests itself if you try to assign to the indexed array, and that is why it was not visible with the existing unit test "test_fancy_indexing_boundscheck"

Fixes #8127

### DIFF
--- a/numba/tests/test_boundscheck.py
+++ b/numba/tests/test_boundscheck.py
@@ -26,6 +26,12 @@ def fancy_array_access(x):
     return x[a]
 
 
+def fancy_array_modify(x):
+    a = np.array([1, 2, 3])
+    x[a] = 0
+    return x
+
+
 class TestBoundsCheckNoError(MemoryLeakMixin, unittest.TestCase):
     def setUp(self):
         self.old_boundscheck = config.BOUNDSCHECK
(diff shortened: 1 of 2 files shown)


numba/tests/test_boundscheck.py
```diff
@@ -204,6 +210,26 @@ def test_fancy_indexing_boundscheck(self):
         with self.assertRaises(IndexError):
             boundscheck(a)
 
+    def test_fancy_indexing_with_modification_boundscheck(self):
+        a = np.arange(3)
+        b = np.arange(4)
+
+        # Check the numpy behavior to ensure the test is correct.
+        with self.assertRaises(IndexError):
+            # TODO: When we raise the same error message as numpy, test that
```
njriasan, line 219: Is there a github issue associated with this? I'm worried this TODO will be lost without a proper github issue.

njriasan (approved):
This looks good to me. Thank you so much for the contribution.

numba/tests/test_boundscheck.py
bszollosinagy: A quick look tells me no, there is no issue, but I would not worry about forgetting about it. Basically Numba now simply outputs: IndexError: index is out of bounds, where as Numpy will also print some details: IndexError: index 100 is out of bounds for axis 0 with size 3.  Bounds checking is optional, and only used during development, because it slows vectorization. Printing more details is a conveniencve function, and if it really block somebody, I'm sure they will raise the issue. Don't worry I say :D

stuartarchibald:
> Apparently the issue only manifests itself if you try to assign to the indexed array, and that is why it was not visible with the existing unit test "test_fancy_indexing_boundscheck"

Adding the `boundscheck` kwarg to `get_item_pointer2` was done in the function `fancy_setslice` which implements "fancy" indexing whilst "setting" items, hence an assignment was needed to trigger this code path via `__setitem__`.

PR #8061, opened 2022-05-17
Migrate random `glue_lowering` to `overload` where easy

apmasell:
Convert all the `glue_lowering` to `overload_method` that is using
`compile_internal` to do basically exactly that.

PR #8040, opened 2022-05-11
Support for Numpy BitGenerators PR#3: Advanced Distributions Support.

kc611:
This PR builds on top of #8038

Adds the following distributions as `Generator` methods:
- `Generator().beta()`
- `Generator().f()`
- `Generator().chisquare()`
- `Generator().standard_cauchy()`
- `Generator().pareto()`
- `Generator().weibull()`
- `Generator().power()`
- `Generator().laplace()`
- `Generator().logistic()`
- `Generator().lognormal()`
- `Generator().rayleigh()`
- `Generator().standard_t()`
- `Generator().wald()`
- `Generator().geometric()`
- `Generator().zipf()`
- `Generator().triangular()`
- `Generator().poisson()`
- `Generator().negative_binomial()`

PR #8250, opened 2022-07-17
Fix issue 8161

### LINKED ISSUE #8161: Argument names for random.uniform are different in Numba and numpy

I think I have discovered a very minor bug - or rather inconsistency with numpy -  in Numba's implementation of np.random.uniform. In numpy the arguments for the lower- and upperbound are named `low` and `high`, in Numba they are named `a` and `b`. My expectation was that the argument names would be the same across Numba and numpy. Of course this problem can be completely avoided by specifying positional arguments instead of keyword arguments.

As a sidenote, Numba also doesn't accept a call to np.random.uniform without arguments, while in numpy this results in a draw from Uniform[0, 1]. I suspect this is because np.random.uniform mirrors Python's implementation of random.uniform as listed here: https://docs.python.org/3/library/random.html#random.uniform which also explains the use of the argument names `a` and `b`.

Please see below for a minimal working example.

Edit: Updated code for clarity.

```python
import numpy as np
import numba as nb

def uniform_pos_arg():
    return np.random.uniform(0.0, 1.0)

uniform_pos_arg_njit = nb.njit(uniform_pos_arg)

def uniform_numpy_arg():
    return np.random.uniform(low=0.0, high=1.0)

uniform_numpy_arg_njit = nb.njit(uniform_numpy_arg)

def uniform_numba_arg():
    return np.random.uniform(a=0.0, b=1.0)

uniform_numba_arg_njit = nb.njit(uniform_numba_arg)

print('numba version:', nb.__version__)
print('numpy version:', np.__version__)

print(uniform_pos_arg())  # works

print(uniform_pos_arg_njit())  # works

print(uniform_numpy_arg())  # works
try:
    print(uniform_numpy_arg_njit())
except:
    print('Calling uniform_numpy_arg_njit() does not work: Numba expects arguments a and b')

try:
    print(uniform_numba_arg())
except:
    print('Calling uniform_numba_arg() does not work: Numpy expects arguments low and high')
print(uniform_numba_arg_njit())
```

Output:
```
numba version: 0.55.2
numpy version: 1.22.3
0.8939617128078611
0.9575756779190199
0.6867936141928337
Calling uniform_numpy_arg_njit() does not work: Numba expects arguments a and b
Calling uniform_numba_arg() does not work: Numpy expects arguments low and high
0.6859305961634644
```

bszollosinagy:
Fixes #8161 

Keyword parameter names are now consistent between Numba and Numpy for these random distributions:

- np.random.uniform
- np.random.gumbel
- np.random.wald

Unit tests were also made to test that the named keyword arguments are properly handled.

The solutions closely follow the suggestion given by @guilhermeleobas

The issue presently could not be solved for these:
- np.random.f
- np.random.vonmises

Reason: It turns out that their unit tests tested Numba against the Python implementation and not Numpy. In itself this is not a big problem, because the python implementations are also delivering the characteristics of the chosen random distribution, but in practice, if started from the same seed, the Numpy and Python implementations give differing samples (which makes the unit test fail, if you really want to compare Numba to Numpy and not python).

For example TestRandom.test_numpy_vonmises() compares the Numba implementation against the Python not the Numpy "vonmises" distribution. It should use "_follow_numpy"  instead of "_follow_cpython"

Why does this matter? Because the keyword arguments have different names in Python Random, and Numpy Random, so one cannot simply keep using the Python distributions as baselines if the keyword names are important. One also cannot use the Numpy versions, because they give different samples than Numba, if started from the same seed. This is most likely caused by Numba and Numpy using the random number generator slightly differently for the same algorithm. For example if you need three random numbers A,B and C, maybe one of them asks three samples and puts them into A,B and C, maybe the other simply initializes C before A and B, thus putting the first random sample into C instead of A. This results in a distribution with the appropriate characteristics, yet numerically different output. This issue only affects the Numba implementations of the distributions F and VonMises  as far as I can tell.

Maybe this is not a big issue, and it is perfectly okay for Numba to give the same pseudo-random stream as Python, and not the same as Numpy. And someone could just make a wrapper function for testing, that wraps the Python functions to have the keywords used by Numpy.


numba/core/typing/randomdecl.py
```diff
@@ -228,6 +228,36 @@ def typer(shape, scale=None, size=None):
...
+
+
+@glue_typing(np.random.wald, typing_key="np.random.wald")
+class Numpy_wald(ConcreteRandomTemplate):
+    cases = [signature(tp, tp, tp) for tp in _float_types]
+
+    def generic(self):
+        def typer(mean, scale, size=None):
+            return self.array_typer(size)(mean, scale)
+        return typer
```
guilhermeleobas, line 258: Numba is transitioning to use the [high-level API](https://numba.pydata.org/numba-doc/latest/extending/high-level.html) for implementing NumPy functions. Once the transition is done, the `glue_*` machinery will not be necessary anymore. @apmasell is working on migrating the `np.random.*` module to use the high-level API on https://github.com/numba/numba/pull/8061. With that said, I wonder if this change worth or if it better to just try using the high level API.

guilhermeleobas (commented):
Hi @bszollosinagy, thanks for working on this issue. As I mentioned in the review, Numba is migrating some code from the old API into the high level one.

Also, can you include the tests that NumPy uses to tests the three functions you're changing? You can find them in the file below:
https://github.com/numpy/numpy/blob/main/numpy/random/tests/test_generator_mt19937.py

numba/tests/test_random.py
```diff
@@ -483,12 +503,31 @@ def _check_uniform(self, func, ptr):
         self._check_dist(func, r.uniform,
                          [(1.5, 1e6), (-2.5, 1e3), (1.5, -2.5)])
 
+    def _check_arbitrary_distrib_kwargs(self, func, ptr, distrib, paramlist):
+        """
+        Check any numpy distribution function. Does Numba use the same keyword argument names as Numpy?
+        And given a fixed seed, do they both return the same samples?
```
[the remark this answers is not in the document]
bszollosinagy: done

numba/core/typing/randomdecl.py
bszollosinagy: Hi @guilhermeleobas, thanks for your review. I took a look at [#8061](https://github.com/numba/numba/pull/8061), and yes, the implementations using new API solve #8161 perfectly fine. I see that some distributions are not yet implemented, and for those ones #8161 remains an issue (8161 specifically concerns np.random.uniform, and that one is not yet converted to the new API).  On the other hand, Gumbel and Wald are already implemented, and my unit tests pass on that branch. The question is: how can I include the high level API without plagiarizing the work done by @apmasell ?

bszollosinagy:
Hi @guilhermeleobas, concerning the Numpy tests: you recommend including the unit tests that Numpy uses. Those tests verify that for some pre-defined seed, Numpy outputs the expected series of numbers. As far as I can tell, you are already peforming the equivalent test for Numba. The Numba BaseTest._follow_numpy(ptr) creates a Numpy generator, sets the seed from Numpy into Numba, and then you will sample some values from both functions (Numpy and Numba) and compare if you got the same sequence. This is done in TestRandom._check_dist_kwargs and TestRandom.\_check_dist, and by default you will sample niters=3 times.

The issue I mentionned happens because some tests named test_numpy\_\* (for example: test_numpy_vonmises) cheat a little, and do not actually  test against Numpy, because they call self._follow_cpython() which creates a standard Python generator, which,  for a pre-defined seed, gives a different sequence of numbers than Numpy. So your tests exist, but for F and VonMises, they have a typo, and compare the sequence against Python's own implementation, not Numpy's.

This means that for F and VonMises, **Numba is no longer a drop in replacement for Numpy**. Should I raise a separate issue about this on GitHub?

numba/core/typing/randomdecl.py
guilhermeleobas: Hi @bszollosinagy, to avoid re-doing @apmasell work. I think it is better to just wait his PR to land. As for `np.random.uniform`, if you can, you can do the migration for this particular function in a new PR.

guilhermeleobas:
> This means that for F and VonMises, **Numba is no longer a drop in replacement for Numpy**. Should I raise a separate issue about this on GitHub?

@kc611, is your work on https://github.com/numba/numba/pull/8232 solves this?

numba/core/typing/randomdecl.py
apmasell: Hopefully my changes will land soon, but there will be another round of changes to push the migrate the random functions that generate IR directly.

bszollosinagy: Okay, I'll hold on, and when #8061 lands, then I'll merge the main into this PR and adapt it.

bszollosinagy:
@guilhermeleobas concerning your question to @kc611 : I went and looked, and as far as I can tell, the issue I mentioned comes from test_random.py, and the "old way of calling Numpy random", and is **not an issue with** the new BitGenerator based approach that was merged with PR #8040 . There, another test suite is used : test_np_random_gen.py, which correctly compares the numbafied Numpy against the original Numpy implementation. 

I know Numpy has been trying to deprecate the "old way" of generating random number  (ie. simply calling np.random.vonmises(mu, kappa)), but it is still available. I don't know if it makes sense to put any effort into bringing the Numba version of this old style into a bit-perfect drop-in replacement... if nobody raised the issue, maybe they are happy with it? (after all, the test does verify that it is a bit-perfect replacement for the Python implementation).

kc611:
@bszollosinagy is right. This issue was initially discussed at the start of the `Generator` project in Numba, the older NumPy random API in Numba is not a exact replacement of the original and hence efforts were made to prevent that while implementing the newer API in Numba. (This was also raised before as issues, for instance: https://github.com/numba/numba/issues/7886 and the current https://github.com/numba/numba/issues/8161)  

As for the question, do we want to put efforts towards improving support for older API, as we can clearly see NumPy wanting to move away from it and actively promoting use of the newer API in it's documentation. And the general consensus for this during the discussion was the first priority would be to develop and refine the newer API as with time more and more people will be switching to it. But any efforts to improving the older API would be appreciated as well.

numba/core/typing/randomdecl.py
bszollosinagy: Okay, I see that #8061 is merged. So now I take it by merging main into my branch, and solve the merge conflicts by dropping the old glue code.  For the uniform distrib I keep the glue code, because that is not yet implemented in the new style.

bszollosinagy: The new unit tests pass (those that test that keyword based calling works as intended). 
So the issue is solved for the uniform distribution (which was the main task for #8161 ). In total, these distributions now work well with keyword arguments: (Wald and Gumbel were done by @apmasell in PR #8061  by using override instead of glue, which implicitly solved the issue)

- np.random.uniform
- np.random.gumbel
- np.random.wald

This PR is now ready for review

bszollosinagy:
The new unit tests pass (those that test that keyword based calling works as intended).
So the issue is solved for the uniform distribution (which was the main task for https://github.com/numba/numba/issues/8161 ). In total, these distributions now work well with keyword arguments: (Wald and Gumbel were done by @apmasell in PR https://github.com/numba/numba/pull/8061 by using override instead of glue, which implicitly solved the issue)

    np.random.uniform
    np.random.gumbel
    np.random.wald

This PR is now ready for review

apmasell:
There are updates coming for those remaining ones in #8368

bszollosinagy:
Okay, CI issues are solved, and now it is clear that all unit tests pass. 

PR ready for final review.

bszollosinagy:
After the recent merging of  #8368 this PR no longer contains any implementation related to the keywords, because they are not needed (as discussed above). Unit tests that test the keyword based calls are the main contribution now, and may be used to test for regressions in the future, if needed.

The reviewer can decide if he/she wants to merge these tests to the main branch (but they don't take long).

guilhermeleobas:
HI @bszollosinagy, I'll put this one for discussion at the next Numba meeting, which happens next tuesday.

apmasell (approved):
The changes look good. I think we can extend this to other functions and do any parameter renaming if required. When refactoring the other random functions, I used the names that were present without checking back to the original NumPy code.

sklam (approved):

PR #8253, opened 2022-07-18
CUDA: Verify NVVM IR prior to compilation

gmarkall:
It is a good idea to verify the NVVM IR before compilation. The NVVM IR verifier provides the same function as the LLVM IR verifier with some additional CUDA-specific checks.

I had originally planned to make the verification optional, but given that the CPU target always runs the verifier I have decided to always run verification - I expect this might help catch future issues more easily.

There were some small fix-ups to the tests that needed making in order for them to pass verification, but this was only because they left the target triple unspecified or invalid.

### DIFF
--- a/numba/cuda/cudadrv/nvvm.py
+++ b/numba/cuda/cudadrv/nvvm.py
@@ -113,6 +113,10 @@ class NVVM(object):
         #                           int* minorDbg )
         'nvvmIRVersion': (nvvm_result, POINTER(c_int), POINTER(c_int),
                           POINTER(c_int), POINTER(c_int)),
+        # nvvmResult nvvmVerifyProgram (nvvmProgram prog, int numOptions,
+        #                               const char** options)
+        'nvvmVerifyProgram': (nvvm_result, nvvm_program, c_int,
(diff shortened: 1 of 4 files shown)


gmarkall:
gpuci run tests

stuartarchibald (commented):
Thanks for the patch. The changes seem good, only question is whether there's merit in adding a test attempting to compile some invalid IR so as to ensure that verification is working correctly?

In testing this PR I also checked that the CPU target llvm verification was not inadvertently running for the CUDA target (as this would largely be duplicated effort) and it is not.

gmarkall:
> The changes seem good, only question is whether there's merit in adding a test attempting to compile some invalid IR so as to ensure that verification is working correctly?

Good point, I've now added an additional test of this.

gpuci run tests

stuartarchibald (approved):

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_142`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_142`.

Passed.

PR #8256, opened 2022-07-19
Migrate linear algebra functions from glue_lowering

apmasell:
Change functions in `linalg` module using `glue_lowering` to new `overload`
annotation.

### DIFF
--- a/numba/np/linalg.py
+++ b/numba/np/linalg.py
@@ -555,7 +555,7 @@ def dot_2_impl(name, left, right):
                     elif ndims == (1, 1):
                         return dot_2_vv(context, builder, sig, args)
                     else:
-                        assert 0
+                        raise AssertionError('unreachable')
 
             if left.dtype != right.dtype:
                 raise TypingError(
@@ -788,7 +788,7 @@ def dot_3(left, right, out):
                     elif ndims == {1, 2}:
                         return dot_3_vm(context, builder, sig, args)
                     else:
-                        assert 0
+                        raise AssertionError('unreachable')
             if left.dtype != right.dtype or left.dtype != out.dtype:
                 raise TypingError(
                     "np.dot() arguments must all have the same dtype")
--- a/numba/np/linalg.py
+++ b/numba/np/linalg.py
@@ -4,6 +4,7 @@ Implementation of linear algebra operations.
 
 
 import contextlib
+import warnings
 
 from llvmlite import ir
 
@@ -15,7 +16,8 @@ from numba.core.imputils import (lower_builtin, impl_ret_borrowed,
 from numba.core.typing import signature
 from numba.core.extending import intrinsic, overload, register_jitable
 from numba.core import types, cgutils
-from numba.core.errors import TypingError, NumbaTypeError
+from numba.core.errors import TypingError, NumbaTypeError, \
+    NumbaPerformanceWarning
 from .arrayobj import make_array, _empty_nd_impl, array_copy
 from numba.np import numpy_support as np_support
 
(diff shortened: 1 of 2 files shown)


numba/np/linalg.py
```diff
@@ -521,39 +520,52 @@ def check_args(a, b):
...
+                    elif ndims == [2, 1]:
+                        return dot_2_mv(context, builder, sig, args)
+                    elif ndims == [1, 2]:
+                        return dot_2_vm(context, builder, sig, args)
+                    elif ndims == [1, 1]:
+                        return dot_2_vv(context, builder, sig, args)
+                    else:
+                        assert 0
+
+            return signature(types.Array, left, right), _dot2_codegen
```
sklam, line 549:
```
            return signature(types.Array, left, right), _dot2_codegen
                             ^^^^^^^^^^^
```
needs to be a instance of `types.Array`. you can get it from the logic in npydecl.py

numba/np/linalg.py
```diff
@@ -521,39 +520,52 @@ def check_args(a, b):
...
+        def _impl(typingcontext, left, right):
+            def codegen(context, builder, sig, args):
+                ensure_blas()
 
-    with make_contiguous(context, builder, sig, args) as (sig, args):
-        return dot_2_vv(context, builder, sig, args, conjugate=True)
+                with make_contiguous(context, builder, sig, args) as\
+                        (sig, args):
+                    return dot_2_vv(context, builder, sig, args, conjugate=True)
+            return signature(types.Array, left, right), _impl
```
sklam, line 567: same problem here

numba/np/linalg.py
```diff
@@ -724,24 +736,32 @@ def check_args(a, b, out):
...
+                with make_contiguous(context, builder, sig, args) as (sig, args):
+                    ndims = set(x.ndim for x in sig.args[:2])
+                    if ndims == set([2]):
+                        return dot_3_mm(context, builder, sig, args)
+                    elif ndims == set([1, 2]):
+                        return dot_3_vm(context, builder, sig, args)
+                    else:
+                        assert 0
+
+            return signature(types.Array, left, right, out), codegen
```
sklam, line 760: here too

numba/np/linalg.py
```diff
@@ -521,39 +520,52 @@ def check_args(a, b):
...
-@glue_lowering(np.vdot, types.Array, types.Array)
-def vdot(context, builder, sig, args):
+@overload(np.vdot)
+def vdot(left, right):
     """
     np.vdot(a, b)
     """
-    ensure_blas()
+    if isinstance(left, types.Array) and isinstance(right, types.Array):
+        def _impl(typingcontext, left, right):
```
sklam, line 560:
```suggestion
    if isinstance(left, types.Array) and isinstance(right, types.Array):
        @intrinsic
        def _impl(typingcontext, left, right):
```

numba/np/linalg.py
```diff
@@ -521,39 +520,84 @@ def check_args(a, b):
...
+                    if ndims == (2, 2):
+                        return dot_2_mm(context, builder, sig, args)
+                    elif ndims == (2, 1):
+                        return dot_2_mv(context, builder, sig, args)
+                    elif ndims == (1, 2):
+                        return dot_2_vm(context, builder, sig, args)
+                    elif ndims == (1, 1):
+                        return dot_2_vv(context, builder, sig, args)
+                    else:
+                        assert 0
```
sklam, line 558:
```suggestion
                        raise AssertionError('unreachable')
```
because Python can skip `assert` statement with `python -O <script>`

sklam (commented):
one nitpick.

numba/np/linalg.py
apmasell: No problem. There's actually a bunch of these assertions in this file.  This one was in the original code block. Maybe some of the others need to be replaced.

sklam (approved):
Thanks for the patch!

PR #8259, opened 2022-07-19
Add `np.broadcast_to(scalar_array, ())`

guilhermeleobas:
This PR adds a special case where the input to `np.broadcast_to` is a scalar array + empty shape.


numba/np/arrayobj.py
```diff
@@ -1339,6 +1339,26 @@ def codegen(context, builder, sig, args):
...
+
+    def codegen(context, builder, sig, args):
+        [src] = args
+        srcty = sig.args[0]
+
+        dest = make_array(srcty)(context, builder, src)
+        # Hack to return a read-only array
+        setattr(dest, 'parent', Constant(
+                context.get_value_type(dest._datamodel.get_type('parent')),
+                None))
```
kc611, line 1356: I have a question here. What's the difference between `parent` as an attribute and `parent` as a type within the datamodel ?

numba/np/arrayobj.py
```diff
@@ -1389,6 +1409,14 @@ def impl(array, shape):
     elif isinstance(shape, types.Integer):
         def impl(array, shape):
             return np.broadcast_to(array, (shape,))
+    elif isinstance(shape, types.Tuple):
```
kc611, line 1412: This branch could use a simple check to see if it's actually an empty tuple and not something like `(1, 'x')`, the current logic accepts that shape which is invalid.

numba/np/arrayobj.py
```diff
@@ -1389,6 +1409,14 @@ def impl(array, shape):
...
+        if array.ndim == 0:
+            # broadcast_to(0d_array, )
+            def impl(array, shape):
+                return get_readonly_array(array)
+        else:
+            msg = 'cannot broadcast a non-scalar to a scalar array'
```
kc611, line 1418: Minor nitpick ? 
```suggestion
            msg = 'Cannot broadcast a non-scalar to a scalar array'
```

guilhermeleobas: Fixed

numba/np/arrayobj.py
guilhermeleobas: Thanks, fixed!

numba/np/arrayobj.py
guilhermeleobas: `parent` as an attribute is a member of the ndarray struct, whereas the datamodel says how `parent` should be typed. See:
https://github.com/numba/numba/blob/2e9b58cfce0391b968ab8c7c84393cc12c4d1bfb/numba/core/datamodel/models.py#L876

kc611 (approved):
The test failure looks unrelated, rest LGTM!

esc:
> The test failure looks unrelated, rest LGTM!

Yes, this should be fixed with #8354 -- but you may need to merge in `main` b/c RTD will build a PR against the branch head and not the branch merge, so the build doesn't include any fixes from `main` (unlike Azure).

esc (changes requested):
Some nitpicks I found.

numba/np/arrayobj.py
```diff
@@ -1339,6 +1339,28 @@ def codegen(context, builder, sig, args):
...
+        [src] = args
+        srcty = sig.args[0]
+
+        dest = make_array(srcty)(context, builder, src)
+        # Hack to return a read-only array
+        # breakpoint()
+        dest.parent = cgutils.get_null_value(dest.parent.type)
+        # setattr(dest, 'parent', Constant(
+        #         context.get_value_type(dest._datamodel.get_type('parent')),
+        #         None))
```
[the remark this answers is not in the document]
guilhermeleobas: Yes! Removed

numba/np/arrayobj.py
```diff
@@ -1339,6 +1339,28 @@ def codegen(context, builder, sig, args):
...
+    ret = arr.copy(readonly=True)
+    sig = ret(arr)
+
+    def codegen(context, builder, sig, args):
+        [src] = args
+        srcty = sig.args[0]
+
+        dest = make_array(srcty)(context, builder, src)
+        # Hack to return a read-only array
+        # breakpoint()
```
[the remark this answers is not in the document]
guilhermeleobas: Removed

numba/np/arrayobj.py
```diff
@@ -1389,6 +1407,14 @@ def impl(array, shape):
     elif isinstance(shape, types.Integer):
         def impl(array, shape):
             return np.broadcast_to(array, (shape,))
+    elif isinstance(shape, types.Tuple) and shape.count == 0:
+        if array.ndim == 0:
```
stuartarchibald, line 1411: This:
```python
from numba import njit
import numpy as np

@njit
def foo():
    return np.broadcast_to(123, ())

print(foo())
```
produces:
```
    if array.ndim == 0:
AttributeError: 'Integer' object has no attribute 'ndim'
```

guilhermeleobas:
CI failure is not related to this PR.

stuartarchibald:
> CI failure is not related to this PR.

Failed test was: `ERROR: test_np_MachAr_deprecation_np122 (numba.tests.test_np_functions.TestNPMachineParameters)`. It indeed seems unrelated.

esc:
Running on the Anaconda Internal Build farm as `numba_smoketest_cpu_yaml_138`.

numba/np/arrayobj.py
```diff
@@ -1389,6 +1408,16 @@ def impl(array, shape):
     elif isinstance(shape, types.Integer):
         def impl(array, shape):
             return np.broadcast_to(array, (shape,))
+    elif isinstance(shape, types.Tuple) and shape.count == 0:
+        is_scalar_array = isinstance(array, types.Array) and array.ndim == 0
+        if type_is_scalar(array) or is_scalar_array:
```
stuartarchibald, line 1413: Perhaps add a comment here saying that `array` must be supported by `type_can_asarray` as checked above. I just spent a few minutes trying to work out why inputs such as `'hello'` are commented in the test below, it's because whilst they are permitted scalar types, they fail the `type_can_asarray` check so cannot reach here.

This also lead to the discovery of this bug:
```python
from numba import njit
import numpy as np


@njit
def foo():
    return np.broadcast_to('a', (2, 3))

print(foo())
```
During type inference, for the non-literal type attempt, this function correctly fails to accept `'a'` as an input (as it's a `unicode_type`). For the subsequent literal type attempt, the types are: `(Literal[str](a), Tuple(Literal[int](2), Literal[int](3)))`. The literal string 'a', is correctly handled and accepted, but the shape tuple is now a `types.Tuple` of literal integers (i.e. heterogeneous), and this is not handled correctly and leads to calling the `else:` branch fall through below.

stuartarchibald (commented):
Thanks for the update @guilhermeleobas. There's a couple of issues to resolve else looks good.

esc:
> Running on the Anaconda Internal Build farm as `numba_smoketest_cpu_yaml_138`.

This was all green 🍏 -- but probably will need another review after @stuartarchibald 's comments have been addressed.

numba/np/arrayobj.py
guilhermeleobas: Hi Stuart, I've addressed the issue.

esc:
`Build numba_smoketest_cpu_yaml_141 has started`

guilhermeleobas:
@esc, can you please re-run this PR in the build farm?

esc:
> `Build numba_smoketest_cpu_yaml_141 has started`

This was fine, again! Now waiting on a final sign-off from @stuartarchibald .

stuartarchibald (approved):

PR #8267, opened 2022-07-20
Fix segfault for invalid axes in np.split

### LINKED ISSUE #8257: Segfault in np.split with negative axis

```python
import numba
import numpy as np

@numba.njit
def foo():
    return np.split(np.zeros((10, 10)), np.array([5]), axis=-1)

foo()  # segfault
```

According to gdb the segfault occurs [in this function](https://github.com/numba/numba/blob/61ec1fd0f69aeadece218dccf4c39ebc5c7dfbc4/numba/core/runtime/_nrt_python.c#L332)

numba version '0.55.2'

aseyboldt:
This adds a new function `normalize_axis` to `arrayobj.py`, and uses this function to verify axis arguments in `np.split`. There already is a similar function that could be wrapped in an intrinsic, but because the error message needs to be constant that seems more complicated than necessary to me.

I also use the new `normalize_axis` function to simplify `np.swapaxes` and `np.take_along_axis` a bit.

Fixes #8257

### DIFF
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -14,6 +14,7 @@ import numpy as np
 
 from numba import pndindex, literal_unroll
 from numba.core import types, utils, typing, errors, cgutils, extending
+from numba.core.decorators import generated_jit
 from numba.np.numpy_support import (as_dtype, carray, farray, is_contiguous,
                                     is_fortran, check_is_integer,
                                     type_is_scalar)
@@ -26,7 +27,7 @@ from numba.core.imputils import (lower_builtin, lower_getattr,
                                  impl_ret_new_ref, impl_ret_untracked,
                                  RefType)
 from numba.core.typing import signature
-from numba.core.types import Literal
+from numba.core.types import StringLiteral
 from numba.core.extending import (register_jitable, overload, overload_method,
                                   intrinsic)
 from numba.misc import quicksort, mergesort
@@ -271,14 +272,10 @@ def update_array_info(aryty, array):
                                           get_itemsize(context, aryty))
 
 
+@generated_jit(nopython=True)
 def normalize_axis(func_name, ndim, axis):
     """Constrain axis values to valid positive values."""
-    raise NotImplementedError()
-
-
-@overload(normalize_axis)
-def _normalize_axis_overloads(func_name, ndim, axis):
-    if not isinstance(func_name, Literal):
+    if not isinstance(func_name, StringLiteral):
         raise errors.TypingError("func_name must be a str literal.")
 
     msg = f"{func_name.literal_value}: Axis out of bounds"
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -26,6 +26,7 @@ from numba.core.imputils import (lower_builtin, lower_getattr,
                                  impl_ret_new_ref, impl_ret_untracked,
                                  RefType)
 from numba.core.typing import signature
+from numba.core.types import Literal
 from numba.core.extending import (register_jitable, overload, overload_method,
                                   intrinsic)
 from numba.misc import quicksort, mergesort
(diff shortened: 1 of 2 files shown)


aseyboldt:
There are still some test failures, but maybe they are unrelated? The workflow fails even before the tests are started.

numba/np/arrayobj.py
```diff
@@ -266,6 +267,29 @@ def update_array_info(aryty, array):
...
+    """Constrain axis values to valid positive values."""
+    raise NotImplementedError()
+
+
+@overload(normalize_axis)
+def _normalize_axis_overloads(func_name, ndim, axis):
+    if not isinstance(func_name, Literal):
+        raise errors.TypingError("func_name must be a str literal.")
+
+    msg = f"{func_name.literal_value}: Axis out of bounds"
```
guilhermeleobas, line 280: Instead of an overload, you can either use `@generated_jit` or `@register_jitable`

guilhermeleobas (commented):
HI @aseyboldt, thanks for working on this issue. I gave an initial review

numba/np/arrayobj.py
aseyboldt: I switched to `generated_jit`. I don't see how I could use `register_jitable` though, because I think the error message needs to be known at compile time, and register_jitable can't dispatch on literal types, can it?

numba/np/arrayobj.py
```diff
@@ -266,6 +267,29 @@ def update_array_info(aryty, array):
...
 
 
+def normalize_axis(func_name, ndim, axis):
+    """Constrain axis values to valid positive values."""
+    raise NotImplementedError()
+
+
+@overload(normalize_axis)
+def _normalize_axis_overloads(func_name, ndim, axis):
+    if not isinstance(func_name, Literal):
```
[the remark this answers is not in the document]
aseyboldt: fixed

numba/np/arrayobj.py
guilhermeleobas: That's correct! But it should be possible after https://github.com/numba/numba/pull/8134 gets merged.

numba/np/arrayobj.py
```diff
@@ -5641,8 +5662,10 @@ def impl(ary, indices_or_sections, axis=0):
     ):
         def impl(ary, indices_or_sections, axis=0):
             slice_tup = build_full_slice_tuple(ary.ndim)
-            out = list()
+            axis = normalize_axis("np.split", ary.ndim, axis)
+            out = []
             prev = 0
+            cur = 0
```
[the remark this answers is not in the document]
aseyboldt: fixed

aseyboldt:
Completely forgot about this PR :-)

> The reduced level of information in the error message is a little unfortunate - would you be able to modify the normalize_axis() function so it takes the function name and argument name and bakes them both into the exception message?

I changed the message to:
```
f"{func_name.literal_value}: Argument {arg_name.literal_value} out of bounds for dimensions of the array"
```

And reverted back to overload

numba/tests/test_np_functions.py
```diff
@@ -2768,6 +2772,10 @@ def test_split_basic(self):
             str(raises.exception)
         )
 
+        with self.assertRaises(ValueError) as raises:
+            njit(split)(np.ones(5), [3], axis=-3)
+        self.assertIn("np.split", str(raises.exception))
```
gmarkall, line 2777:
```suggestion
        self.assertIn("np.split: Argument axis out of bounds",
                      str(raises.exception))
```

numba/tests/test_np_functions.py
```diff
@@ -4651,14 +4659,12 @@ def test_swapaxes_exception(self):
         with self.assertRaises(ValueError) as raises:
             cfunc(np.arange(4), 1, 0)
 
-        self.assertIn('The second argument "axis1" is out of bounds for array'
-                      ' of given dimension', str(raises.exception))
+        self.assertIn('np.swapaxes', str(raises.exception))
```
gmarkall, line 4662:
```suggestion
        self.assertIn('np.swapaxes: Argument axis1 out of bounds',
                      str(raises.exception))
```

numba/tests/test_np_functions.py
```diff
@@ -4651,14 +4659,12 @@ def test_swapaxes_exception(self):
...
 
         with self.assertRaises(ValueError) as raises:
             cfunc(np.arange(8).reshape(2, 4), 0, -3)
 
-        self.assertIn('The third argument "axis2" is out of bounds for array'
-                      ' of given dimension', str(raises.exception))
+        self.assertIn('np.swapaxes', str(raises.exception))
```
gmarkall, line 4667:
```suggestion
        self.assertIn('np.swapaxes: Argument axis2 out of bounds',
                      str(raises.exception))
```

gmarkall (changes requested):
Many thanks for the updates - I think this is looking pretty good. I have made some small suggestions on the diff just to ensure the reporting of the argument name as well as the function name - are you happy with these? If so, then once they are accepted I can approve this PR.

A quick note on process for the future - could you use merge commits and avoid rebasing / force-pushes please? GIthub reviews get a bit confused when there is a force push, which makes it harder to check what the changes since the last review were - many thanks in advance!

aseyboldt:
>  so, then once they are accepted I can approve this PR.

done

> A quick note on process for the future - could you use merge commits and avoid rebasing / force-pushes please?

Sure, I will to that in the future :-)

gmarkall (approved):
Many thanks for the fix!

PR #8271, opened 2022-07-22
Implement some CUDA intrinsics with `@overload`, `@overload_attribute`, and `@intrinsic`

gmarkall:
This continues on the work from PR #7013, updated to use current keywords ('target' instead of 'hardware') and to fix it so that it actually works.


gmarkall:
gpuci run tests

leofang:
This is so much nicer!

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
Now that this has passed CI, I'm making this Ready for Review - after some OOB discussion with @stuartarchibald, we have decided that it's reasonable to start moving towards this, given that we're early in the 0.57 cycle, and getting it merged will help to turn up issues in support for the high-level extension API in the CUDA target with plenty of time to resolve them before it gets close to the release.

numba/cuda/intrinsics.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+def _syncthreads_predicate(typingctx, predicate, fname):
+    if not isinstance(predicate, types.Integer):
+        return None
+
+    # Don't allow larger ints to be downcast automatically - force the user to
+    # think about what they want instead
+    if predicate.bitwidth > 32:
+        return None
```
stuartarchibald, line 156: Should this perhaps raise an informative error about refusing to downcast, else I think this will just manifest as a refusal to compile with no reason given? Also, this is a change from the existing code base which permits the cast, should it be deprecated first?

numba/cuda/extending.py
```diff
@@ -0,0 +1,7 @@
+"""
+Added for symmetry with the core API
```
stuartarchibald, line 2: Should the rest of `numba.extending.overload*` also perhaps be added?

stuartarchibald (commented):
Thanks for the patch, the code changes look good and are what I'd expect from switching APIs, which is encouraging. I've left a few comments about the changes inline, next thing is to manually test the implementation further. Thanks again!

stuartarchibald:
I've tried to break this feature in a number of ways, including involved things like calling a CUDA intrinsic, from an `njit` function with no specified targets, from a CUDA kernel, just to make sure that involved call resolution works as expected. I'm not wholly convinced that the `@overload` decorator should accept incompatible flags for the CUDA target, e.g. `parallel=True`, and I don't think that e.g. `fastmath` is necessarily working. However, I think these issues should be resolved subsequently and not as part of this PR which is about enabling basic support of the extension API for the CUDA target. Another future piece of work is to document this as working (once we're suitably happy with it!).

numba/cuda/compiler.py
```diff
@@ -207,14 +207,16 @@ def compile_cuda(pyfunc, return_type, args, debug=False, lineinfo=False,
...
     # Run compilation pipeline
-    cres = compiler.compile_extra(typingctx=typingctx,
-                                  targetctx=targetctx,
-                                  func=pyfunc,
-                                  args=args,
-                                  return_type=return_type,
-                                  flags=flags,
-                                  locals={},
-                                  pipeline_class=CUDACompiler)
+    from numba.core.target_extension import target_override
```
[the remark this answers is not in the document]
gmarkall: It creates import cycles.

numba/cuda/intrinsics.py
```diff
@@ -0,0 +1,201 @@
...
+    Computation of the first integer is as follows::
+
+        cuda.threadIdx.x + cuda.blockIdx.x * cuda.blockDim.x
+
+    and is similar for the other two indices, but using the ``y`` and ``z``
+    attributes.
+    '''
+
+    if not isinstance(ndim, types.IntegerLiteral):
+        return None
```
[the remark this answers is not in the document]
gmarkall: Ah, I wasn't aware of this - thanks for pointing to it!

numba/cuda/intrinsics.py
gmarkall: I hadn't realised this was changing behaviour - I thought that the old typing that only provided signatures for `i4` would prevent the downcast, but that turns out not to be the case. I don't really want to change the API for this PR, so I'm just going to remove this check instead,

numba/cuda/extending.py
gmarkall: I'd like to add it, but I'd like to do it incrementally with tests for each component that I add here in future PRs, rather than adding it all at once.

gmarkall:
gpuci run tests

gmarkall:
@stuartarchibald Many thanks for the review and testing - I've responded to each of the comments and made some changes to raise `RequireLiteralValue` where appropriate and undo the accidental API change.

numba/cuda/intrinsics.py
stuartarchibald: I agree with keeping the API and with removing the check, thanks!

numba/cuda/intrinsics.py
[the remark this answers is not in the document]
stuartarchibald: No problem, thanks for incorporating into the patch.

numba/cuda/extending.py
stuartarchibald: OK, sounds like a safe plan!

stuartarchibald (approved):

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_157`.

Passed (on all non-CUDA-10.2 builds as expected since #8287)!

After many months of effort to get to this point... `@overload` works for CUDA!

Many thanks @gmarkall, marking as ready to merge !!!

Issue #6249, opened 2020-09-16
LoweringError assigning array to `.real` part of NumPy array


stuartarchibald:
## Reporting a bug

- [x] I have tried using the latest released version of Numba (most recent is
 visible in the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [x] I have included below a minimal working reproducer (if you are unsure how
 to write one see http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports).

From: https://gitter.im/numba/numba?at=5f620a11ce5bbc7ffdcb4da9

This:
```python
from numba import njit
import numpy as np

@njit
def foo(input_tensor):
    complex_input = np.empty(input_tensor.shape[-1:], dtype=np.complex128)
    complex_input.real = input_tensor[0]
    complex_input.imag = input_tensor[1]

foo(np.empty((3, 2)))
```
does this:
```
numba.core.errors.LoweringError: Failed in nopython mode pipeline (step: nopython mode backend)
No definition for lowering array(complex128, 1d, C).real = array(float64, 1d, A)

File "gi286.py", line 7:
def foo(input_tensor):
    <source elided>
    complex_input = np.empty(input_tensor.shape[-1:], dtype=np.complex128)
    complex_input.real = input_tensor[0]
    ^
```

MothNik:
@stuartarchibald Was there any progress on this? I'm facing the same problem in version `0.60.0`.

PR #8283, opened 2022-07-26
Clean up / remove support for old NumPy versions

gmarkall:
Removes some support / legacy code for old NumPy versions (and one instance where NumPy differs between Python 2 and 3). 

The comments in `np.interp` support warn not to refactor it, but I think the reason for not refactoring is to avoid diverging from the NumPy source - however, because the Numba implementation supported multiple NumPy versions, and it now does not, after the refactor it is actually closer to the C source of the NumPy versions we do support, so I think this is an acceptable refactoring.

### DIFF
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -456,19 +456,6 @@ def zero_dim_msg(fn_name):
     return msg
 
 
-def _is_nat(x):
-    pass
-
-
-@overload(_is_nat)
-def ol_is_nat(x):
-    if numpy_version >= (1, 18):
-        return lambda x: np.isnat(x)
-    else:
-        nat = x('NaT')
-        return lambda x: x == nat
-
-
 @lower_builtin(np.min, types.Array)
 @lower_builtin("array.min", types.Array)
 def array_min(context, builder, sig, args):
@@ -476,25 +463,19 @@ def array_min(context, builder, sig, args):
     MSG = zero_dim_msg('minimum')
 
     if isinstance(ty, (types.NPDatetime, types.NPTimedelta)):
-        # NP < 1.18: NaT is smaller than every other value, but it is
-        # ignored as far as min() is concerned.
-        # NP >= 1.18: NaT dominates like NaN
         def array_min_impl(arry):
             if arry.size == 0:
                 raise ValueError(MSG)
 
             it = np.nditer(arry)
             min_value = next(it).take(0)
-            if _is_nat(min_value):
+            if np.isnat(min_value):
                 return min_value
 
             for view in it:
                 v = view.item()
-                if _is_nat(v):
-                    if numpy_version >= (1, 18):
-                        return v
-                    else:
-                        continue
+                if np.isnat(v):
+                    return v
                 if v < min_value:
                     min_value = v
             return min_value
@@ -559,25 +540,19 @@ def array_max(context, builder, sig, args):
     MSG = zero_dim_msg('maximum')
 
     if isinstance(ty, (types.NPDatetime, types.NPTimedelta)):
-        # NP < 1.18: NaT is smaller than every other value, but it is
-        # ignored as far as min() is concerned.
-        # NP >= 1.18: NaT dominates like NaN
         def array_max_impl(arry):
             if arry.size == 0:
                 raise ValueError(MSG)
 
             it = np.nditer(arry)
             max_value = next(it).take(0)
-            if _is_nat(max_value):
+            if np.isnat(max_value):
                 return max_value
 
             for view in it:
                 v = view.item()
-                if _is_nat(v):
-                    if numpy_version >= (1, 18):
-                        return v
-                    else:
-                        continue
+                if np.isnat(v):
+                    return v
                 if v > max_value:
                     max_value = v
             return max_value
@@ -642,13 +617,13 @@ def array_argmin_impl_datetime(arry):
     it = np.nditer(arry)
     min_value = next(it).take(0)
     min_idx = 0
-    if _is_nat(min_value):
+    if np.isnat(min_value):
         return min_idx
 
     idx = 1
     for view in it:
         v = view.item()
-        if _is_nat(v):
+        if np.isnat(v):
             if numpy_version >= (1, 18):
                 return idx
             else:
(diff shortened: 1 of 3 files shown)


jamesobutler:
Hi and thanks for your continued maintenance support of `numba` 🙏🏻 . 

Does `numba` follow [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html)? I see this work is now linked to the Numba 0.57 release milestone which I read on another thread is estimated to be released in January 2023. As of today NEP 29 would indicate only supporting numpy 1.20+ and in January 2023 supporting numpy 1.21+. Should this PR go further in cleaning up/removing support of numpy <1.21 in preparation of the Numba 0.57 January 2023 release date?

> ```
> On Dec 26, 2021 drop support for Python 3.7 (initially released on Jun 27, 2018)
> On Jun 21, 2022 drop support for NumPy 1.19 (initially released on Jun 20, 2020)
> On Jan 31, 2023 drop support for NumPy 1.20 (initially released on Jan 31, 2021)
> ```
https://numpy.org/neps/nep-0029-deprecation_policy.html#drop-schedule

gmarkall:
> Hi and thanks for your continued maintenance support of numba 🙏🏻 .

You're welcome :slightly_smiling_face: - thanks for looking over this PR and sharing your thoughts!

> Does numba follow [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html)? I see this work is now linked to the Numba 0.57 release milestone which I read on another thread is estimated to be released in January 2023. As of today NEP 29 would indicate only supporting numpy 1.20+ and in January 2023 only supporting numpy 1.21+.

We've used NEP 29 as a guide for dropping support for versions before, see e.g. https://github.com/numba/numba/wiki/Minutes_2021_12_07, https://github.com/numba/numba/wiki/Minutes_2021_04_27.

> Should this PR go further in cleaning up/removing support of numpy <1.21 in preparation of the Numba 0.57 January 2023 release date?

This PR is only removing code that is dead because it handled situations in which Numba was being used in conjunction with NumPy versions that it has already dropped support for.

I do think it's possible (desirable, even?) to drop support for more versions, especially since we'll need to add 1.23 to the matrix soon, but that work should happen in a separate PR.

gmarkall:
Also, I note CuPy is following NEP 29 for its 12.x plan, which is planned to release in February 2023: https://github.com/cupy/cupy/issues/6866

numba/np/arraymath.py
```diff
@@ -2397,156 +2374,37 @@ def np_interp_impl_complex_fp_inner(x, xp, fp, dtype):
...
+                    if np.isnan(imag) and dy[j].imag == dy[j + 1].imag:
+                        imag = dy[j].imag
 
-                # NOTE: there's a change in master which is not
-                # in any released version of 1.16.x yet... as
-                # per the real value implementation, but
-                # interpolate real and imaginary parts
-                # independently; this will need to be added in
-                # due course
+                dres.flat[i] = real + 1j * imag
```
stuartarchibald, line 2396: I think this should ideally be direct assignment of the real/imag parts to prevent potential multiply by NaN, however it seems like there's an issue with lowering that at present.

numba/np/arraymath.py
```diff
@@ -2397,156 +2374,37 @@ def np_interp_impl_complex_fp_inner(x, xp, fp, dtype):
                     imag = (dy[j + 1].imag - dy[j].imag) * inv_dx
                     slope = real + 1j * imag
 
+                # Numpy 1.17 handles NaN correctly - this is a copy of
+                # innermost part of arr_interp_complex post 1.17:
+                # https://github.com/numpy/numpy/blob/maintenance/1.17.x/numpy/core/src/multiarray/compiled_base.c    # noqa: E501
+                # Permanent reference:
+                # https://github.com/numpy/numpy/blob/91fbe4dde246559fa5b085ebf4bc268e2b89eea8/numpy/core/src/multiarray/compiled_base.c#L798-L812    # noqa: E501
```
stuartarchibald, line 2381: Checked, implementation appears to be translated correctly.

numba/np/arraymath.py
```diff
@@ -2612,142 +2473,25 @@ def np_interp_impl_inner(x, xp, fp, dtype):
...
                 # If we get nan in one direction, try the other
-                # if np.isnan(dres.flat[i]):
-                #     dres.flat[i] = slope * (x_val - dx[j + 1]) + dy[j + 1]
-                #
-                #     if np.isnan(dres.flat[i]) and dy[j] == dy[j + 1]:
-                #         dres.flat[i] = dy[j]
+                if np.isnan(dres.flat[i]):
+                    dres.flat[i] = slope * (x_val - dx[j + 1]) + dy[j + 1]    # noqa: E501
+                    if np.isnan(dres.flat[i]) and dy[j] == dy[j + 1]:
+                        dres.flat[i] = dy[j]
```
stuartarchibald, line 2485: Checked changes against NumPy, the algorithms appear to be the same.

numba/np/arraymath.py
```diff
@@ -2388,6 +2362,9 @@ def np_interp_impl_complex_fp_inner(x, xp, fp, dtype):
                 dres.flat[i] = rval
             elif j == lenxp - 1:
                 dres.flat[i] = dy[j]
+            elif dx[j] == x_val:
+                # Avoid potential non-finite interpolation
+                dres.flat[i] = dy[j]
```
stuartarchibald, line 2367: Have checked unit tests hit this.

numba/np/arraymath.py
```diff
@@ -2604,6 +2462,9 @@ def np_interp_impl_inner(x, xp, fp, dtype):
                 dres.flat[i] = rval
             elif j == lenxp - 1:
                 dres.flat[i] = dy[j]
+            elif dx[j] == x_val:
+                # Avoid potential non-finite interpolation
```
stuartarchibald, line 2466: Have checked unit tests hit this.

stuartarchibald (commented):
Thanks for the patch. As requested I've taken a long look at the refactoring of `np.interp` and I'm of the view that the refactor is good, it appears to match with NumPy 1.18+ and the unit tests are fairly extensive which adds to confidence. The rest of the patch looks fine, the only actionable item is a `Numpy`->`NumPy` typo. Many thanks for doing this refactoring!

numba/np/arraymath.py
stuartarchibald: xref: #6249

stuartarchibald (approved):

PR #8287, opened 2022-07-27
Drop CUDA 10.2

gmarkall:
By the time of the next release (0.57, early 2023) CUDA 10.2 will be more than three years old. Supporting it will require extra maintenance overhead for changes planned in the near term, so it is dropped to reduce this burden.

Notes:

- I removed a comment about `cudadevrt` from the Cooperative Groups docs. This comment was only relevant when there was the possibility of `cudadevrt` being missing for an installation, when (broken) older CUDA toolkit packages that happened to be missing it were still supported. Now they are unsupported, `cudadevrt` can be assumed to be present so there is no need to explicitly mention it in the CG docs.
- Previously there was a concern about dropping support for Jetsons by dropping CUDA 10.2 - however, Jetpack 5 (presently in developer preview, [on the roadmap for release](https://developer.nvidia.com/embedded/develop/roadmap), prior to the Numba 0.57 release) has CUDA 11.4 so this would no longer be an issue.
- Some code to handle pre-11.0 toolkits has been simplified / removed.


gmarkall:
gpuci run tests

gmarkall:
(The fail is the current parfors fail that is present on master).

sklam (approved):
The patch looks good.

Before it's merged, the numba buildfarm needs to be adjusted. We should waiting until 0.56.1 release is done before changing the farm

esc:
`Build numba_smoketest_cuda_yaml_154 has started`

gmarkall:
gpuci run tests

esc:
`a9c095e82` has passed on the anaconda internal build farm for all CUDA versions except 10.2 as expected.

esc:
Thank you for the patch!

Issue #8288, opened 2022-07-27
`main` failing in parfors testing as of 6f70229


stuartarchibald:
This test (and others in the same `TestCase`) are failing: `FAIL: test_issue5942_1 (numba.tests.test_parfors.TestParforsMisc)` with:

<details>
<summary>Details:</summary>

```
======================================================================
FAIL: test_issue5942_1 (numba.tests.test_parfors.TestParforsMisc)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/vsts/work/1/s/numba/tests/test_parfors.py", line 2988, in test_issue5942_1
    self.check(test_impl, gg, gg_next)
  File "/home/vsts/work/1/s/numba/tests/test_parfors.py", line 282, in check
    cfunc, cpfunc = self.compile_all(pyfunc, *args)
  File "/home/vsts/work/1/s/numba/tests/test_parfors.py", line 184, in compile_all
    cpfunc = self.compile_parallel(pyfunc, sig)
  File "/home/vsts/work/1/s/numba/tests/test_parfors.py", line 172, in compile_parallel
    return self._compile_this(func, sig, flags=self.pflags)
  File "/home/vsts/work/1/s/numba/tests/test_parfors.py", line 169, in _compile_this
    return compile_isolated(func, sig, flags=flags)
  File "/home/vsts/work/1/s/numba/core/compiler.py", line 309, in compile_isolated
    return compile_extra(typingctx, targetctx, func, args, return_type,
  File "/home/vsts/work/1/s/numba/core/compiler.py", line 716, in compile_extra
    return pipeline.compile_extra(func)
  File "/home/vsts/work/1/s/numba/core/compiler.py", line 452, in compile_extra
    return self._compile_bytecode()
  File "/home/vsts/work/1/s/numba/core/compiler.py", line 520, in _compile_bytecode
    return self._compile_core()
  File "/home/vsts/work/1/s/numba/core/compiler.py", line 495, in _compile_core
    raise e
  File "/home/vsts/work/1/s/numba/core/compiler.py", line 486, in _compile_core
    pm.run(self.state)
  File "/home/vsts/work/1/s/numba/core/compiler_machinery.py", line 364, in run
    raise e
  File "/home/vsts/work/1/s/numba/core/compiler_machinery.py", line 356, in run
    self._runPass(idx, pass_inst, state)
  File "/home/vsts/work/1/s/numba/core/compiler_lock.py", line 35, in _acquire_compile_lock
[...]

PR #8265, opened 2022-07-20
Remove `mk_unique_var` from `array_analysis.py`

guilhermeleobas:
xref: https://github.com/numba/numba/issues/8230

PR #8289, opened 2022-07-27
Revert #8265.

stuartarchibald:
This reverts #8265 as it causes mainline builds to fail in parfors
testing. The cause seems to be related to scope transfer for the
gufunc and resulting name collisions in the type map during array
analysis as a result. Tracking issue #8288.

This reverts commit 6f702299ba2cbe66a196b05a43a49d06bc529008, reversing
changes made to e3d4912f8e6b1ddf8f7b42d73b963c18595d8824.

### DIFF
--- a/numba/parfors/array_analysis.py
+++ b/numba/parfors/array_analysis.py
@@ -7,6 +7,7 @@ import numpy
 import operator
 from numba.core import types, ir, config, cgutils, errors
 from numba.core.ir_utils import (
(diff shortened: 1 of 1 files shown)


esc (approved):
I can confirm this is indeed a revert of #8265 and public CI is passing. This looks good.

sklam (approved):
i can confirm this revert unbreak the failing test locally.

PR #8292, opened 2022-07-27
update checklist

esc:
Rephrase the task to update the version support table and include it as
a cherry-pick instead.


stuartarchibald (approved):
Thanks for the patch.

PR #8295, opened 2022-07-28
Add get_const_mem_size method to Dispatcher

testhound:
This pull request partially address issue: https://github.com/numba/numba/issues/6448 by adding support for retrieving the constant memory size utilized by a kernel from the Dispatcher.


testhound:
gpuci run tests

numba/cuda/dispatcher.py
```diff
@@ -703,6 +710,26 @@ def get_regs_per_thread(self, signature=None):
...
+        '''
+        Returns the size in bytes of user-allocated constant memory in this
+        kernel for the device in the current context.
+
+        :param signature: The signature of the compiled kernel to get constant
+                          memory usage for. This may be omitted for a
+                          specialized kernel.
+        :return: The size in bytes of constant memory allocated by the
+         compiled variant of the kernel for the given signature and current
+        device.
```
stuartarchibald, line 723:
```suggestion
        :return: The size in bytes of constant memory allocated by the
                 compiled variant of the kernel for the given signature and
                 current device.
```
Think something like this should fix the doc build.

numba/cuda/dispatcher.py
```diff
@@ -191,6 +191,13 @@ def regs_per_thread(self):
         '''
         return self._codelibrary.get_cufunc().attrs.regs
 
+    @property
+    def const_mem_size(self):
+        '''
+        The size of user-allocated const memory for this kernel.
```
gmarkall, line 197: What does "user-allocated" mean in this context?

testhound: @gmarkall I am referring to uses of "cuda.const.array_like" in a kernel.

testhound:
gpuci run tests

testhound:
@gmarkall I updated the docstring with you recommendation.

numba/cuda/tests/cudapy/test_dispatcher.py
```diff
@@ -272,6 +280,67 @@ def pi_sin_array(x, n):
...
+
+        @cuda.jit
+        def const_record_double(A, B):
+            C = cuda.const.array_like(CONST_RECORD_DOUBLE)
+            i = cuda.grid(1)
+            A[i] = C[i]['x']
+            B[i] = C[i]['y']
+
+        # Call the kernel with different arguments to create two different
+        # definitions within the Dispatcher object
```
gmarkall, line 299: This test isn't calling a kernel with different arguments to create two different definitions within the Dispatcher object - instead it's using two different Dispatchers with one definition each, so this test isn't really testing the differentiation between specializations (a similar issue to that described in https://github.com/numba/numba/pull/8240#discussion_r924362469).

As with the tests above, it would be better to check that const memory usage is at least that expected, rather than exactly equal to the minimum possible const memory usage based on the values used in the kernel.

To create a test where different specializations use different amounts of const memory, we can define a kernel that prints values of different types, because the lengths of their format strings will differ. To do this (and address comments about simplifying `test_get_const_mem_specialized()` below you could apply:

```diff
diff --git a/numba/cuda/tests/cudapy/test_dispatcher.py b/numba/cuda/tests/cudapy/test_dispatcher.py
index d5889e7c8..ac61b9719 100644
--- a/numba/cuda/tests/cudapy/test_dispatcher.py
+++ b/numba/cuda/tests/cudapy/test_dispatcher.py
@@ -1,18 +1,10 @@
 import numpy as np
 import threading
 
-from numba import cuda, float32, float64, int32, int64, void
+from numba import boolean, cuda, float32, float64, int32, int64, void
 from numba.cuda.testing import skip_on_cudasim, unittest, CUDATestCase
 import math
 
-CONST_RECORD_FLOAT = np.array(
-    [(1.0, 2), (3.0, 4)],
-    dtype=[('x', np.float32), ('y', np.int32)])
-
-CONST_RECORD_DOUBLE = np.array(
-    [(1.0, 2), (3.0, 4)],
-    dtype=[('x', np.float64), ('y', np.int32)])
-
 
 def add(x, y):
     return x + y
@@ -282,64 +274,52 @@ class TestDispatcher(CUDATestCase):
 
     def test_get_const_mem_unspecialized(self):
         @cuda.jit
-        def const_record_float(A, B):
-            C = cuda.const.array_like(CONST_RECORD_FLOAT)
-            i = cuda.grid(1)
-            A[i] = C[i]['x']
-            B[i] = C[i]['y']
-
-        @cuda.jit
-        def const_record_double(A, B):
-            C = cuda.const.array_like(CONST_RECORD_DOUBLE)
-            i = cuda.grid(1)
-            A[i] = C[i]['x']
-            B[i] = C[i]['y']
+        def const_fmt_string(val, to_print):
+            # We guard the print with a conditional to prevent noise from the
+            # test suite
+            if to_print:
+                print(val)
 
         # Call the kernel with different arguments to create two different
         # definitions within the Dispatcher object
-        N = 2
-        arr_f32 = np.zeros(N, dtype=np.float32)
-        arr_f64 = np.zeros(N, dtype=np.float64)
-        arr_int = np.zeros(N, dtype=np.int32)
-
-        const_record_float[1, N](arr_f32, arr_int)
-        const_record_double[1, N](arr_f64, arr_int)
+        const_fmt_string[1, 1](1, False)
+        const_fmt_string[1, 1](1.0, False)
 
         # Check we get a positive integer for the two different variations
-        sig_f32 = void(float32[::1], int32[::1])
-        sig_f64 = void(float64[::1], int32[::1])
-        const_mem_size_f32 = const_record_float.get_const_mem_size(sig_f32)
-        const_mem_size_f64 = const_record_double.get_const_mem_size(sig_f64)
+        sig_i64 = void(int64, boolean)
+        sig_f64 = void(float64, boolean)
+        const_mem_size_i64 = const_fmt_string.get_const_mem_size(sig_i64)
+        const_mem_size_f64 = const_fmt_string.get_const_mem_size(sig_f64)
 
-        self.assertIsInstance(const_mem_size_f32, int)
+        self.assertIsInstance(const_mem_size_i64, int)
         self.assertIsInstance(const_mem_size_f64, int)
 
-        self.assertEqual(const_mem_size_f32, CONST_RECORD_FLOAT.nbytes)
-        self.assertEqual(const_mem_size_f64, CONST_RECORD_DOUBLE.nbytes)
+        # 6 bytes for the equivalent of b'%lld\n\0'
+        self.assertGreaterEqual(const_mem_size_i64, 6)
+        # 4 bytes for the equivalent of b'%f\n\0'
+        self.assertGreaterEqual(const_mem_size_f64, 4)
 
         # Check that getting the const memory size for all signatures
         # provides the same values as getting the const memory size for
         # individual signatures.
 
-        const_mem_size_f32_all = const_record_float.get_const_mem_size()
-        const_mem_size_f64_all = const_record_double.get_const_mem_size()
-        self.assertEqual(const_mem_size_f32_all[sig_f32.args],
-                         const_mem_size_f32)
-        self.assertEqual(const_mem_size_f64_all[sig_f64.args],
-                         const_mem_size_f64)
+        const_mem_size_all = const_fmt_string.get_const_mem_size()
+        self.assertEqual(const_mem_size_all[sig_i64.args], const_mem_size_i64)
+        self.assertEqual(const_mem_size_all[sig_f64.args], const_mem_size_f64)
 
     def test_get_const_mem_specialized(self):
-        @cuda.jit(void(float32[::1], int32[::1]))
-        def const_record_float(A, B):
-            C = cuda.const.array_like(CONST_RECORD_FLOAT)
+        arr = np.arange(32, dtype=np.int64)
+        sig = void(int64[::1])
+
+        @cuda.jit(sig)
+        def const_array_use(x):
+            C = cuda.const.array_like(arr)
             i = cuda.grid(1)
-            A[i] = C[i]['x']
-            B[i] = C[i]['y']
+            x[i] = C[i]
 
-        sig_f32 = void(float32[::1], int32[::1])
-        const_mem_size_f32 = const_record_float.get_const_mem_size(sig_f32)
-        self.assertIsInstance(const_mem_size_f32, int)
-        self.assertEqual(const_mem_size_f32, CONST_RECORD_FLOAT.nbytes)
+        const_mem_size = const_array_use.get_const_mem_size(sig)
+        self.assertIsInstance(const_mem_size, int)
+        self.assertGreaterEqual(const_mem_size, arr.nbytes)
 
     def test_get_shared_mem_per_block_unspecialized(self):
         N = 10
```

numba/cuda/tests/cudapy/test_dispatcher.py
```diff
@@ -272,6 +280,67 @@ def pi_sin_array(x, n):
...
+        # individual signatures.
+
+        const_mem_size_f32_all = const_record_float.get_const_mem_size()
+        const_mem_size_f64_all = const_record_double.get_const_mem_size()
+        self.assertEqual(const_mem_size_f32_all[sig_f32.args],
+                         const_mem_size_f32)
+        self.assertEqual(const_mem_size_f64_all[sig_f64.args],
+                         const_mem_size_f64)
+
+    def test_get_const_mem_specialized(self):
```
gmarkall, line 331: I think this test could be simplified / improved:

- The second argument seems superfluous.
- It seems unnecessary to use a record array, and there's no need to keep the definition of the array it freezes as a constant at global scope.
- It should check that const memory usage is at least the size of the array, rather than equal to it.

gmarkall (changes requested):
Many thanks for the PR - there are some comments on the diff to address.

testhound:
gpuci run tests

testhound:
@gmarkall I updated the PR with all your comments and your patch.

gmarkall (dismissed):
Many thanks for resolving everything - this looks good now.

numba/cuda/dispatcher.py
```diff
@@ -191,6 +191,13 @@ def regs_per_thread(self):
         '''
         return self._codelibrary.get_cufunc().attrs.regs
 
+    @property
+    def const_mem_size(self):
+        '''
+        The amount of const memory used by this kernel.
```
stuartarchibald, line 197:
```suggestion
        The amount of constant memory used by this kernel.
```

stuartarchibald (dismissed):
Thanks for the patch. Approved for the purposes of testing on the build farm.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_149`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_149`.

Passed.

gmarkall:
gpuci run tests

stuartarchibald (approved):

PR #8299, opened 2022-07-29
Fix build for mingw toolchain

Biswa96:

### DIFF
--- a/numba/np/ufunc/omppool.cpp
+++ b/numba/np/ufunc/omppool.cpp
@@ -15,7 +15,7 @@ Threading layer on top of OpenMP.
 #include "workqueue.h"
 #include "gufunc_scheduler.h"
 
-#ifdef _MSC_VER
+#ifdef _WIN32
 #include <malloc.h>
 #else
 #include <pthread.h>
@@ -33,7 +33,9 @@ Threading layer on top of OpenMP.
 #elif defined(__clang__)
 #define _OMP_VENDOR "Intel"
 #elif defined(__GNUC__) // NOTE: clang also defines this, but it's checked above
+#ifndef _WIN32
 #define _NOT_FORKSAFE 1 // GNU OpenMP Not forksafe
+#endif
 #define _OMP_VENDOR "GNU"
 #endif
 
--- a/numba/np/ufunc/workqueue.c
+++ b/numba/np/ufunc/workqueue.c
@@ -16,7 +16,7 @@ race conditions.
 #undef _XOPEN_SOURCE
 #endif
 
-#ifdef _MSC_VER
+#ifdef _WIN32
 /* Windows */
 #include <windows.h>
 #include <process.h>
--- a/setup.py
+++ b/setup.py
@@ -235,9 +235,15 @@ def get_ext_modules():
     # Set various flags for use in TBB and openmp. On OSX, also find OpenMP!
     have_openmp = True
     if sys.platform.startswith('win'):
-        cpp11flags = []
-        ompcompileflags = ['-openmp']
-        omplinkflags = []
+        if 'MSC' in sys.version:
+            cpp11flags = []
+            ompcompileflags = ['-openmp']
+            omplinkflags = []
+        else:
+            # For non-MSVC toolchain e.g. gcc and clang with mingw
+            cpp11flags = ['-std=c++11']
+            ompcompileflags = ['-fopenmp']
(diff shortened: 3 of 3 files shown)


gmarkall:
Many thanks for the PR! Could you please provide some details about the environment / setup in which this patch is useful? E.g. how is MinGW installed, how was LLVM / llvmlite setup, what versions of tools and Numba dependencies are in use?

(Basically - how would one get set up to test these fixes?)

Biswa96:
I am using clang in msys2/mingw-w64 environment.

* Some info about the toolchain https://www.msys2.org/docs/environments/
* llvmlite setup https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-python-llvmlite
* numba setup https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-python-numba

gmarkall:
@Biswa96 Many thanks for the info - unfortunately none of the Numba / llvmlite maintainers are users of MinGW, so it's quite hard for us to see how to get set up to test this patch - can you provide some guidance as to how to set up an environment to test the patch please?

Biswa96:
Understood. Let me check if I can setup any GitHub Actions for mingw environment.

guilhermeleobas:
Hi @Biswa96, I have a Windows machine and can try your patch. If you provide a high level description of the necessary steps, that would be helpful.

Biswa96:
Verbose documentation can be found here https://www.msys2.org/. Here is the gist:
* Download the msys2 installer from that page and install it.
* Run `C:\msys64\clang64.exe` or the clang64 shortcut in start menu.
* Update packages with `pacman -Syyu` **twice**.
* Install required packages `pacman -S pacboy; pacboy -S git cc:p python:p python-setuptools:p` and other required dependencies for this project. The `:p` is necessary because it is translated into `mingw-w64-clang64-x86_64-` prefix.
* Make sure there is no other python in `$PATH`. `which python.exe` will show `/clang64/bin/python.exe`.
* Now do the regular development steps.

guilhermeleobas:
Thanks, @Biswa96. I'll try it in the upcoming days.

guilhermeleobas:
HI @Biswa96, thanks for the info. I can confirm your patch works and I was able to successfully build Numba with it. I also tried to build Numba main branch, which failed because it could not build `omppool.o`. 

Below I will copy the steps I did for the "**regular development steps**" to compile Numba on Windows with msys2:
* pacboy -S python-llvmlite:p python-numpy:p
* Since llvmlite on msys2 is old (v0.38), I had to manually disable the `_ensure_llvm()` check on `numba/__init__.py`
* Build numba with `python setup.py build_ext --inplace`
* Tried a simple test case:

```python
$ cat file.py
from numba import njit
import numpy as np

@njit
def foo(a):
    return np.arange(a)

print(foo(3))

$ python file.py
[0, 1, 2]
```

guilhermeleobas:
Hi @Biswa96, thanks for the info. I can confirm your patch works and I was able to successfully build Numba with it. I also tried to build Numba main branch, which failed because it could not build `omppool.o`. 

Below I will copy the steps I did for the "**regular development steps**" to compile Numba on Windows with msys2:
* pacboy -S python-llvmlite:p python-numpy:p
* Since llvmlite on msys2 is old (v0.38), I had to manually disable the `_ensure_llvm()` check on `numba/__init__.py`
* Build numba with `python setup.py build_ext --inplace`
* Tried a simple test case:

```python
$ cat file.py
from numba import njit
import numpy as np

@njit
def foo(a):
    return np.arange(a)

print(foo(3))

$ python file.py
[0, 1, 2]
```

cc @esc

gmarkall (approved):
Assuming this is also good on the build farm, this looks good to me. I think it is OK to accept patches to build on platforms we don't explicitly support, as we do for llvmlite (e.g. it has accepted patches for building on BSDs in the past).

Biswa96:
mingw is not an entirely different "platform". In simple terms,
1. The first commit applies the same compiler flags from Linux to Windows.
2. The second commit removes the misconception that `_MSC_VER` means Windows platform. [`_WIN32`][1] should be used to check the target OS.

[1]: https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros

guilhermeleobas (approved):

esc:
I ran this on the Anaconda build farm and the build was all green (`numba_smoketest_cpu_yaml_119`).

PR #8302, opened 2022-08-01
CUDA: Revert numba_nvvm intrinsic name workaround

gmarkall:
This is a small tidy-up of an old workaround (added in #6080) that prevented auto-upgrade of the IR when it passed through the llvmlite LLVM version - since this is no longer done, the workaround is no longer needed.

### DIFF
--- a/numba/cuda/cudadrv/nvvm.py
+++ b/numba/cuda/cudadrv/nvvm.py
@@ -651,12 +651,6 @@ def llvm_replace(llvmir):
     for decl, fn in replacements:
         llvmir = llvmir.replace(decl, fn)
 
-    # llvm.numba_nvvm.atomic is used to prevent LLVM 9 onwards auto-upgrading
-    # these intrinsics into atomicrmw instructions, which are not recognized by
-    # NVVM. We can now replace them with the real intrinsic names, ready to
-    # pass to NVVM.
-    llvmir = llvmir.replace('llvm.numba_nvvm.atomic', 'llvm.nvvm.atomic')
-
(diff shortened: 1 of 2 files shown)


gmarkall:
gpuci run tests

guilhermeleobas (approved):

guilhermeleobas:
> since this is no longer done, the workaround is no longer needed.

No longer done because of the LLVM version used by llvmlite?

gmarkall:
> No longer done because of the LLVM version used by llvmlite?

Numba used to generate LLVM IR, parse and optimize it with the llvmlite binding, then print out the IR again before feeding it to NVVM. This resulted in super-weird and hard-to-debug issues, probably because of some subtle mismatch of semantics/bugs between the LLVM versions in llvmlite and NVVM, so Numba doesn't do this anymore - it just generates the text-based IR with llvmlite and then passes it straight to NVVM.

The auto-upgrade happened during the parse process, which doesn't happen anymore.

gmarkall:
@esc could this have a CUDA buildfarm run please? (and possibly an approving review so it becomes mergeable?)

stuartarchibald (approved):
Thanks for the patch, this should indeed not be needed now.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_145`.

esc:
> Buildfarm ID: `numba_smoketest_cuda_yaml_145`.

Was all green.

esc (approved):
LGTM

Issue #8307, opened 2022-08-03
Ambiguous overloads are allowed for calls to jitted functions inside other jitted functions


gmarkall:
For example:

```python
from numba import njit

@njit(['(float32, float64)', '(float64, float32)'])
def add(x, y):
    return x + y

# (float64, float64) is an equal match for the (float32, float64) and (float64, float32) compiled versions
try:
    add(1.0, 2.0)
except TypeError as te:
    print(f"Got TypeError as expected: {str(te)}")

# Also an equal match for the compiled versions of add, but is allowed to compile
@njit('(float64, float64)')
def add_caller(x, y):
    return add(x, y)

print(add_caller(1.0, 2.0))
```

```
Got TypeError as expected: Ambiguous overloading for <function add at 0x7fa5c51b9cf0> (float64, float64):
(float32, float64) -> float64
(float64, float32) -> float64
3.0
```

So the dispatcher disallows dispatch of ambiguous overloads, but the compiler permits compilation with them. Is this intended?

Why do ambiguous overloads even matter? Could we just pick the first one instead of giving an error?

sklam:
the problem seems to be at https://github.com/numba/numba/blob/main/numba/core/typing/context.py#L620 and `allows_ambiguous` is `True` when resolving `add()` in `add_caller()`.

stuartarchibald:
This sample code produces the wrong answer due to the above.
```python
from numba import njit
import numpy as np

@njit(['(float32, float64)', '(float64, float32)'])
def add(x, y):
    return x + y

# Also an equal match for the compiled versions of add, but is allowed to compile
@njit('(float64, float64)')
def add_caller(x, y):
    return add(x, y) # x will get converted to float32 as (float32, float64) is a match

a = np.uint64(0x0010000000000000).view(np.float64)
b = np.float64(0x0)
print(add_caller(a, b))

# Now use open compilation for add
@njit
def add_open(x, y):
    return x + y


@njit('(float64, float64)')
def add_caller_to_open(x, y):
    return add_open(x, y)

print(add_caller_to_open(a, b))
```
gives:
```
0.0
2.2250738585072014e-308
```

sklam:
turning off the ambiguous overload leads to error like:

PR #8308, opened 2022-08-03
CUDA: Support for multiple signatures

### LINKED ISSUE #3226: CUDA JIT errors when multiple signatures provided
With Numba's CPU JIT, one is able to supply multiple signatures (e.g. single and double precision floating values) for a function. However when trying the same with Numba's CUDA JIT, this results in an error. Limiting the CUDA JIT to a single signature fixes the issue. Code snippet demonstrating this below.

<details>
<summary>Example:</summary>

```python
In [1]: from numba import jit

In [2]: from numba.cuda import jit as cuda_jit

In [3]: @jit(["f4(f4, f4)", "f8(f8, f8)"])
   ...: def f(a, b):
   ...:     return a + b
   ...: 
   ...: 

In [4]: @cuda_jit(["f4(f4, f4)", "f8(f8, f8)"])
   ...: def g(a, b):
   ...:     return a + b
   ...: 
   ...: 
<ipython-input-4-5a86fdfdf6d9> in <module>()
----> 1 @cuda_jit(["f4(f4, f4)", "f8(f8, f8)"])
      2 def g(a, b):
      3     return a + b

~/.conda/envs/numba/lib/python3.6/site-packages/numba/cuda/compiler.py in __call__(self, *args)
    700         Specialize and invoke this kernel with *args*.
    701         '''
--> 702         kernel = self.specialize(*args)
    703         cfg = kernel[self.griddim, self.blockdim, self.stream, self.sharedmem]
    704         cfg(*args)

~/.conda/envs/numba/lib/python3.6/site-packages/numba/cuda/compiler.py in specialize(self, *args)
    710         '''
    711         argtypes = tuple(
--> 712             [self.typingctx.resolve_argument_type(a) for a in args])
    713         kernel = self.compile(argtypes)
    714         return kernel

~/.conda/envs/numba/lib/python3.6/site-packages/numba/cuda/compiler.py in <listcomp>(.0)
    710         '''
    711         argtypes = tuple(
--> 712             [self.typingctx.resolve_argument_type(a) for a in args])
    713         kernel = self.compile(argtypes)
    714         return kernel

~/.conda/envs/numba/lib/python3.6/site-packages/numba/typing/context.py in resolve_argument_type(self, val)
    288         ValueError is raised for unsupported types.
    289         """
--> 290         return typeof(val, Purpose.argument)
    291 
    292     def resolve_value_type(self, val):

~/.conda/envs/numba/lib/python3.6/site-packages/numba/typing/typeof.py in typeof(val, purpose)
     31         msg = errors.termcolor.errmsg(
     32             "cannot determine Numba type of %r") % (type(val),)
---> 33         raise ValueError(msg)
     34     return ty
     35 

ValueError: cannot determine Numba type of <class 'function'>
```

</details>

gmarkall:
When a list of signatures is passed to the `@cuda.jit` decorator, the dispatcher is compiled for each signature. This does not result in a specialized kernel, even if the list contained just one signature, though compilation will be disabled for the dispatcher.

This ensures that:

- An exact match is required when a list of signatures is provided (matching the dispatch behaviour of the CPU target)
- An exact match is required, but not checked for, when a single signature is provided (matching the current CUDA target behaviour).
- Other functionality that depends on whether a kernel is specalialized (getting shared memory per block, for example) retains its existing behaviour for specialized kernels, such as returning the shared memory per block directly instead of in a dict for each implementation. For dispatchers where multiple signatures were provided, a dict should be returned by these functions.

Tests for multiple signatures are ported from the CPU multiple signature tests in `numba.tests.test_dispatcher`. Note that there is a difference between expectations for kernels and device functions when ambiguity cannot be resolved for an overload - launching a kernel with ambiguous overloads is not permitted, but compiling a call to a device function with ambiguous overloads is permitted - this feels a little odd but is the way Numba currently seems to work for dispatch vs. function calls, and is recorded separately as Issue #8307.

Fixes #3226.


gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
@s-m-e Are you able to try with this branch to check whether it addresses your use case for multiple signatures please?

s-m-e:
@gmarkall Just checked [how to set up a dev environment](https://numba.readthedocs.io/en/stable/developer/contributing.html). Looks doable, so yes.

s-m-e:
I'll report back.

numba/cuda/decorators.py
```diff
@@ -107,14 +112,20 @@ def _jit(func):
...
+
+                if restype and not device and restype != types.void:
+                    raise TypeError("CUDA kernel must have void return type.")
+
+                if device:
+                    from numba.core import typeinfer
+                    with typeinfer.register_dispatcher(disp):
+                        disp.compile_device(argtypes)
+                else:
+                    disp.compile(argtypes)
```
stuartarchibald, line 126: Is it worth splitting this loop to:
1. iterate through the signatures, normalizing them, checking they are compliant and then storing the normalized result.
2. iterate through the stored normalized results invoking compilation on each.

This to create a fast path for ensuring signatures are compliant before starting to compile them.

numba/cuda/tests/cudapy/test_dispatcher.py
```diff
@@ -213,6 +214,171 @@ def wrapper():
...
+
+        r = np.zeros(1, dtype=np.float64)
+        f[1, 1](r, 1, 2**-25)
+        self.assertPreciseEqual(r[0], 1.0000000298023224)
+
+        # Fail to resolve ambiguity between the two best overloads
+        # (Also deliberate float64[::1] for the first argument in all cases)
+        f = cuda.jit(["(float64[::1], float32,float64)",
+                      "(float64[::1], float64,float32)",
+                      "(float64[::1], int64,int64)"])(add_kernel)
```
stuartarchibald, line 259:
```suggestion
        f = cuda.jit(["(float64[::1], float32, float64)",
                      "(float64[::1], float64, float32)",
                      "(float64[::1], int64, int64)"])(add_kernel)
```

numba/cuda/tests/cudapy/test_dispatcher.py
```diff
@@ -213,6 +214,171 @@ def wrapper():
...
+        )
+        # The integer signature is not part of the best matches
+        self.assertNotIn("int64", str(cm.exception))
+
+    @unittest.expectedFailure
+    def test_explicit_signatures_unsafe(self):
+        # These tests are from test_explicit_signatures, but have to be xfail
+        # at present because _prepare_args in the CUDA target cannot handle
+        # unsafe conversions of arguments.
+        f = cuda.jit("(int64[::1], int64,int64)")(add_kernel)
```
stuartarchibald, line 281:
```suggestion
        f = cuda.jit("(int64[::1], int64, int64)")(add_kernel)
```

numba/cuda/tests/cudapy/test_dispatcher.py
```diff
@@ -213,6 +214,171 @@ def wrapper():
...
+        r = np.zeros(1, dtype=np.int64)
+
+        # Approximate match (unsafe conversion)
+        f[1, 1](r, 1.5, 2.5)
+        self.assertPreciseEqual(r[0], 3)
+        self.assertEqual(len(f.overloads), 1, f.overloads)
+
+        sigs = ["(int64[::1], int64,int64)", "(float64[::1], float64, float64)"]
```
stuartarchibald, line 289:
```suggestion
        sigs = ["(int64[::1], int64, int64)", "(float64[::1], float64, float64)"]
```

numba/cuda/tests/cudapy/test_dispatcher.py
```diff
@@ -213,6 +214,171 @@ def wrapper():
...
+        r = np.zeros(1, dtype=np.float64)
+        f[1, 1](r, 1, 2**-25)
+        self.assertPreciseEqual(r[0], 1.0000000298023224)
+
+        # Ambiguity between the two best overloads resolves. This is somewhat
+        # surprising given that ambiguity is not permitted for dispatching
+        # overloads when launching a kernel, but seems to be the general
+        # behaviour of Numba (See Issue #8307:
+        # https://github.com/numba/numba/issues/8307).
+        sigs = ["(float32,float64)", "(float64,float32)", "(int64,int64)"]
```
stuartarchibald, line 353:
```suggestion
        sigs = ["(float32, float64)", "(float64, float32)", "(int64, int64)"]
```

stuartarchibald (commented):
Thanks for the patch, glad to see this implemented, it further closes the gap between the CPU and CUDA targets! Couple of queries to resolve and a few minor flake8 things else looks good.

s-m-e:
@gmarkall @stuartarchibald First of all, cool work, thanks a lot and sorry for the delay in checking this out.

I just spent a couple of hours taking it for a spin and found an interesting issue with the data types of return values.

The following imports are required:

```python
import numba as nb
from numba import cuda
import numpy as np
```

An initial version of my test works:

```python
for target in ('cpu', 'parallel', 'cuda'):

    @nb.vectorize(['f4(f4)', 'f8(f8)'], target = target)
    def demo(x):
        return x ** 2

    for dtype in (np.float32, np.float64):
        
        data = np.arange(2**14, dtype = dtype)
        result = demo(data)
        print(target, type(demo), data.dtype, result.dtype)
        
        assert result.dtype == dtype
    
    del demo

```

Notice the assert statement. I am trying to check if the data type that I am putting in is also the one that is getting out, i.e. matching `['f4(f4)', 'f8(f8)']`.

However, when used with the following code, a "load simulator" of some kind ...

```python
COMPLEXITY = 2 ** 11

from math import cos, sin

for target in ('cpu', 'parallel', 'cuda'):

    if target == 'cuda':
    
        @cuda.jit(['f4(f4)', 'f8(f8)'], device = True)
        def helper(scalar: float) -> float:
            res: float = 0.0
            for idx in range(COMPLEXITY):
                if idx % 2 == round(scalar) % 2:
                    res += sin(idx)
                else:
                    res -= cos(idx)
            return res
    
    else:
        
        @nb.jit(['f4(f4)', 'f8(f8)'], nopython = True)
        def helper(scalar: float) -> float:
            res: float = 0.0
            for idx in range(COMPLEXITY):
                if idx % 2 == round(scalar) % 2:
                    res += sin(idx)
                else:
                    res -= cos(idx)
            return res

    @nb.vectorize(['f4(f4)', 'f8(f8)'], nopython = True, target = target)
    def demo(d: float) -> float:
        return helper(d)

    for dtype in (np.float32, np.float64):
        
        data = np.arange(2**14, dtype = dtype)
        result = demo(data)
        print(target, type(demo), data.dtype, result.dtype)
        
        assert result.dtype == dtype
    
    del demo, helper
```

... I get an assertion error:

```
cpu <class 'numba.np.ufunc.dufunc.DUFunc'> float32 float32
cpu <class 'numba.np.ufunc.dufunc.DUFunc'> float64 float64
parallel <class 'numpy.ufunc'> float32 float32
parallel <class 'numpy.ufunc'> float64 float64
cuda <class 'numba.cuda.vectorizers.CUDAUFuncDispatcher'> float32 float64

/github.numba/numba/np/ufunc/deviceufunc.py:364: RuntimeWarning: nopython kwarg for cuda target is redundant
  warnings.warn("nopython kwarg for cuda target is redundant",
/github.numba/numba/cuda/dispatcher.py:495: NumbaPerformanceWarning: Grid size 26 will likely result in GPU under-utilization due to low occupancy.
  warn(NumbaPerformanceWarning(msg))

Input In [43], in <cell line: 5>()
     38     result = demo(data)
     39     print(target, type(demo), data.dtype, result.dtype)
---> 41     assert result.dtype == dtype
     43 del demo, helper

AssertionError: 
```
I am putting a single in and getting a double out, but only on the `cuda` target. Interestingly though, this might not even be related to this PR. It happens even when providing only a single signature (and then again only for target `cuda`):

```python
COMPLEXITY = 2 ** 11

from math import cos, sin

for target in ('cpu', 'parallel', 'cuda'):

    if target == 'cuda':
    
        @cuda.jit('f4(f4)', device = True)
        def helper(scalar: float) -> float:
            res: float = 0.0
            for idx in range(COMPLEXITY):
                if idx % 2 == round(scalar) % 2:
                    res += sin(idx)
                else:
                    res -= cos(idx)
            return res
    
    else:
        
        @nb.jit('f4(f4)', nopython = True)
        def helper(scalar: float) -> float:
            res: float = 0.0
            for idx in range(COMPLEXITY):
                if idx % 2 == round(scalar) % 2:
                    res += sin(idx)
                else:
                    res -= cos(idx)
            return res

    @nb.vectorize('f4(f4)', nopython = True, target = target)
    def demo(d: float) -> float:
        return helper(d)

    for dtype in (np.float32,):
        
        data = np.arange(2**14, dtype = dtype)
        result = demo(data)
        print(target, type(demo), data.dtype, result.dtype)
        
        assert result.dtype == dtype
    
    del demo, helper
```
... basically results in the same traceback.

So the PR is ok from my end, but I guess I just ran into another unrelated bug (?) ...

s-m-e:
Just checked against `main` - also present there. I'll file a bug.

gmarkall:
gpuci run tests

(quick check that all is well after the merge, I haven't yet addressed comments)

numba/cuda/decorators.py
gmarkall: I'd rather keep the loop as it is for simplicity - if the decorator is going to raise an exception, I'm not sure we need to optimize performance for that case.

gmarkall:
gpuci run tests

gmarkall:
@stuartarchibald Many thanks for the review - I believe I have addressed / responded to all feedback.

numba/cuda/decorators.py
```diff
@@ -83,16 +84,20 @@ def jit(func_or_sig=None, device=False, inline=False, link=[], debug=None,
...
+        signatures = func_or_sig
+        specialized = False
+    else:
+        signatures = None
+
+    if signatures:
         if config.ENABLE_CUDASIM:
             def jitwrapper(func):
                 return FakeCUDAKernel(func, device=device, fastmath=fastmath)
             return jitwrapper
```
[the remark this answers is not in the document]
stuartarchibald: Great, thanks for checking through the impact of this and making the necessary changes!

numba/cuda/decorators.py
stuartarchibald: Thanks for considering it, I agree, let's not optimise until there's a proven need.

numba/cuda/simulator/api.py
```diff
@@ -85,9 +85,14 @@ def jit(func_or_sig=None, device=False, debug=False, argtypes=None,
 
     if link is not None:
         raise NotImplementedError('Cannot link PTX in the simulator')
+
+    # Types would imply that the first argument is a signature or sequence of
+    # signatures
+    sig_types = (str, tuple, list, Signature)
```
stuartarchibald, line 91: I'm not sure that this matches what's accepted here in the dispatcher? https://github.com/numba/numba/blob/0b1ef84c0d7311e7ff47d3bdfa72c02bb6905b94/numba/cuda/decorators.py#L86-L93

Also, are the unit tests checking that string, `Signature`, list and tuple all work as methods of describing a number of signatures (both for the real dispatcher and the simulator dispatcher)? The new tests seem to cover "list of strings"? Should the accepted "containers" of signatures just be anything that implements the iterator protocol?

stuartarchibald (commented):
Thanks for the updates @gmarkall. I have a query I've left inline, thanks for looking!

numba/cuda/simulator/api.py
gmarkall: > I'm not sure that this matches what's accepted here in the dispatcher? 

What do you think the mismatch is? `is_signature()` returns `True` for strings, tuples, and `Signature`s, and the list case is handled in the `elif` clause - am I missing something?

> Also, are the unit tests checking that string, Signature, list and tuple all work as methods of describing a number of signatures (both for the real dispatcher and the simulator dispatcher)? The new tests seem to cover "list of strings"? 

I wrote the new tests by copying and modifying all the relevant ones for the CPU dispatcher, so maybe the testing there is lacking these other cases too.

> Should the accepted "containers" of signatures just be anything that implements the iterator protocol?

I think we generally try to steer people away from eager compilation, so I think I'd not lean towards expanding the API for it without good reason - I think that would be best addressed in a separate PR / discussion if we are going to modify the API.

gmarkall:
@stuartarchibald Thanks for the review! I've answered most of the questions in your comment on the diff, and I'm just looking at expanding the coverage of the cases when a list of signatures is provided.

gmarkall:
gpuci run tests

gmarkall:
@stuartarchibald I've added additional tests that check for lists of each type of signature, as well as mixing types within a list of signatures (which is quite unidiomatic but easy to check!). I think this should be ready for another look now.

gmarkall:
(Note that the fail is #8628, which seems to have suddenly started happening on `main` - PR to fix is #8630).

numba/cuda/simulator/api.py
stuartarchibald: > > I'm not sure that this matches what's accepted here in the dispatcher?
> 
> What do you think the mismatch is? `is_signature()` returns `True` for strings, tuples, and `Signature`s, and the list case is handled in the `elif` clause - am I missing something?

Thanks for pointing this out, in which case, would something like this be a possible improvement in that it doesn't explicitly encode the behaviour of `is_signature` into types?

```python
if func_or_sig is None or sigutils.is_signature(func_or_sig) or isinstance(func_or_sig, list):
```


> > Also, are the unit tests checking that string, Signature, list and tuple all work as methods of describing a number of signatures (both for the real dispatcher and the simulator dispatcher)? The new tests seem to cover "list of strings"?
> 
> I wrote the new tests by copying and modifying all the relevant ones for the CPU dispatcher, so maybe the testing there is lacking these other cases too.

This may indeed be the case. Fixing is probably out of scope for this PR.

> > Should the accepted "containers" of signatures just be anything that implements the iterator protocol?
> 
> I think we generally try to steer people away from eager compilation, so I think I'd not lean towards expanding the API for it without good reason - I think that would be best addressed in a separate PR / discussion if we are going to modify the API.

OK, seems like it's best to leave this for now.

gmarkall: > Thanks for pointing this out, in which case, would something like this be a possible improvement in that it doesn't explicitly encode the behaviour of `is_signature` into types?

Given that different things happen in the different branches, I'm not sure I see how that would work... Can you clarify what you're asking please?

gmarkall:
gpuci run tests

numba/cuda/simulator/api.py
gmarkall: OK, thanks for OOB clarification - change pushed.

stuartarchibald: From OOB conversation. The guard presently looks a bit like:
```python
if X is None or isinstance(X, <types>)...:
```
where the `<types>` are implicitly encoded by what `is_signature` determines to be a signature, so perhaps this should actually just use `is_signature` so as to remove the implicitness. i.e.
```python
if X is None or sigutils.is_signature(X)...:
```
Thanks for addressing this in acd8e53!

stuartarchibald (approved):

esc:
`Build numba_smoketest_cuda_yaml_180 has started`

esc:
> How did the Buildfarm run go?

It was all green. 🍏

PR #8310, opened 2022-08-04
CUDA: Fix Issue #8309 - atomics don't work on complex components

### LINKED ISSUE #8309: Numba 0.56 does not support `atomic.add` on arrays of complexs anymore
## Reporting a bug

The following code snippet uses an `atomic.add` on an array of complex numbers. 
It runs without any issues with `numba 0.55.2`, but fails with `numba 0.56`

```python
from numba import cuda
import numpy as np

@cuda.jit
def atomic_add_one(values):
    """
    Adds 1 to each element values, with an atomic add

    Parameters:
    -----------
    values: 1darray of `complex128`
    """
    tid = cuda.threadIdx.x
    bid = cuda.blockIdx.x
    bdim = cuda.blockDim.x
    i = (bid * bdim) + tid
    cuda.atomic.add(values.real, i, 1)

arr = np.random.rand(16384) + 1.j*np.random.rand(16384)
print(arr.dtype)

atomic_add_one[256,64](arr)
```

gmarkall:
This resolves the issue by reverting the changes from #7999 that make `array.real()` and `array.imag()` overloads.

Current status: this appears to work around the issue reported in #8309 - the reproducer there runs without failing.

To do:
- [X] Resolve CI fails
- [X] Investigate whether `conj` also needs to not be an overload - it does not, because there were no actual changes to `conj`, just diff noise.
- [X] Tidy up the patch so it doesn't make unrelated changes
- [X] Add relevant CUDA tests
- [x] Discussion to determine if this is the right / best way forward to fix the issue in the near term, and what the long term plan is.
   - Planned for the dev meeting on 9th August 2022: https://hackmd.io/P1y0q3BcStKfIZky7CLHHg?view

Fixes #8309.

### DIFF
--- a/numba/cuda/tests/cudapy/test_complex.py
+++ b/numba/cuda/tests/cudapy/test_complex.py
@@ -262,7 +262,7 @@ class TestCMath(BaseComplexTest):
 class TestAtomicOnComplexComponents(CUDATestCase):
     # Based on the reproducer from Issue #8309. array.real and array.imag could
     # not be used because they required returning an array from a generated
-    # function, and even if this was permitted, they coudl not be resolved from
+    # function, and even if this was permitted, they could not be resolved from
     # the atomic lowering when they were overloads.
     #
     # See https://github.com/numba/numba/issues/8309
--- a/numba/cuda/tests/cudapy/test_complex.py
+++ b/numba/cuda/tests/cudapy/test_complex.py
@@ -259,5 +259,38 @@ class TestCMath(BaseComplexTest):
                               ignore_sign_on_zero=True)
 
 
+class TestAtomicOnComplexComponents(CUDATestCase):
+    # Based on the reproducer from Issue #8309. array.real and array.imag could
+    # not be used because they required returning an array from a generated
+    # function, and even if this was permitted, they coudl not be resolved from
+    # the atomic lowering when they were overloads.
+    #
+    # See https://github.com/numba/numba/issues/8309
+
+    def test_atomic_on_real(self):
+        @cuda.jit
+        def atomic_add_one(values):
+            i = cuda.grid(1)
+            cuda.atomic.add(values.real, i, 1)
+
+        N = 32
+        arr1 = np.arange(N) + np.arange(N) * 1j
+        arr2 = arr1.copy()
+        atomic_add_one[1, N](arr2)
+        np.testing.assert_equal(arr1 + 1, arr2)
+
+    def test_atomic_on_imag(self):
+        @cuda.jit
+        def atomic_add_one_j(values):
+            i = cuda.grid(1)
+            cuda.atomic.add(values.imag, i, 1)
+
+        N = 32
+        arr1 = np.arange(N) + np.arange(N) * 1j
+        arr2 = arr1.copy()
+        atomic_add_one_j[1, N](arr2)
+        np.testing.assert_equal(arr1 + 1j, arr2)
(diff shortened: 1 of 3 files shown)


gmarkall:
gpuci run tests

numba/cuda/tests/cudapy/test_complex.py
```diff
@@ -259,5 +259,38 @@ def test_tanh(self):
                               ignore_sign_on_zero=True)
 
 
+class TestAtomicOnComplexComponents(CUDATestCase):
+    # Based on the reproducer from Issue #8309. array.real and array.imag could
+    # not be used because they required returning an array from a generated
+    # function, and even if this was permitted, they coudl not be resolved from
```
stuartarchibald, line 265:
```suggestion
    # function, and even if this was permitted, they could not be resolved from
```

stuartarchibald (commented):
Thanks for fixing this @gmarkall. I've manually checked that the revert covers the parts of the changes introduced in #7999 that broke this functionality on the CUDA target. There's one typo to resolve after which this needs testing on the build farm. Thanks again!

stuartarchibald:
I've manually verified this patch on real hardware, including testing against the original reproducer in #8309.

gmarkall:
@stuartarchibald Thanks for the review - typo fix committed.

gpuci run tests

stuartarchibald (approved):

esc:
`Build numba_smoketest_cuda_yaml_147 has started`

esc:
> `Build numba_smoketest_cuda_yaml_147 has started`

Passed.

stuartarchibald:
Thanks for running this through the farm @esc, am wondering if it ought to also have CPU run too as this patch impacts both CPU and CUDA?

esc:
> Thanks for running this through the farm @esc, am wondering if it ought to also have CPU run too as this patch impacts both CPU and CUDA?

`Build numba_smoketest_cpu_yaml_120 has started`

stuartarchibald:
> > Thanks for running this through the farm @esc, am wondering if it ought to also have CPU run too as this patch impacts both CPU and CUDA?
> 
> `Build numba_smoketest_cpu_yaml_120 has started`

This test build is being impacted by the `setuptools==65` issues, #8356 will be needed to fix the conda recipes.

gmarkall:
Main is now merged.

gmarkall:
`main` has been merged in again, since #8356 was merged - hopefully this is now good for the CPU buildfarm @stuartarchibald / @esc?

gpuci run tests

stuartarchibald (approved):
Thanks for merging in `main` to pick up build fixes @gmarkall .

stuartarchibald:
Buildfarm ID: `numba_smoketest_cpu_yaml_122`.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_148`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_148`.

Passed.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cpu_yaml_122`.

This "passed", the quotes are because `numpy=1.23` tests failed (expected, it's not supported yet) and there was a single test failure in `numba.tests.test_gdb_dwarf` which is known to sometimes fail on a heavily loaded machine, the proposed patch wouldn't impact this.

PR #8315, opened 2022-08-05
Add get_local_mem_per_thread method to Dispatcher

testhound:
This pull request partially address issue: https://github.com/numba/numba/issues/6448 by adding support for retrieving the local memory size utilized by a kernel from the Dispatcher.


testhound:
gpuci run tests

gmarkall (changes requested):
Some quick review points:

* Local memory may be allocated by the compiler or the user, so it's fragile to check for exact local memory usage - instead you need to check that memory usage is greater than or equal to the minimum expected.
* Similarly, `test_get_no_local_memory` is not safe - especially for a GPU with limited resources, spilling to local memory might occur for `func_with_lots_of_registers`.
* Is the level of duplication / copy-pasting starting to get out of hand in test_dispatcher.py? Are there enough commonalities in testing these attributes that we should think about reducing the amount of duplication?
* Please check the spelling in your comments - there are several typos.

testhound:
gpuci run tests

testhound:
@gmarkall I addressed three out of your four requests. After reviewing, I don't agree their is sufficient commonality as the kernels and conditions differ sufficiently.

testhound:
gpuci run tests

numba/cuda/tests/cudadrv/test_linker.py
```diff
@@ -269,6 +284,29 @@ def test_get_shared_mem_per_specialized(self):
         shared_mem_size = compiled_specialized.get_shared_mem_per_block()
         self.assertEqual(shared_mem_size, 800)
 
+    def test_get_no_local_memory(self):
```
gmarkall, line 287: I still don't think this test is safe, as the compiler isn't guaranteed to allocate zero local memory even for an empty kernel - similar to the analagous test in https://github.com/numba/numba/pull/8295, I think this should just be removed.

gmarkall (changes requested):
Thanks for the updates. I think the `test_get_no_local_memory` test should be removed, otherwise this looks OK.

numba/cuda/tests/cudadrv/test_linker.py
testhound: @gmarkall test removed.

testhound:
gpuci run tests

PR #8321, opened 2022-08-08
Fix literal_unroll pass erroneously exiting on non-conformant loop.

### LINKED ISSUE #8311: presence of while loop breaks literal_unroll compilation
I reproduce this issue by running on numba 0.56.0 on google colab.

I cannot understand why the presence of a while loop later in the block creates a `getitem` problem for literal_unroll on heterogenous tuples.

```python
from numba import njit,literal_unroll
@njit
def f1( t ):
    for t_i in literal_unroll(t):
        print('.')

    return 0

@njit
def f2( t ):
    for t_i in literal_unroll(t):
        print('.')

    x=0
    while x>0:
        x= x-1
    
    return 0

print('f1')
f1( (0,0.0) ) # works
print('f2')
f2( (0,0.0) ) # fails
```

The unexpected compile fail for `f2` is the following:

```
 
 >>> getitem(Tuple(int64, float64), int64)
 
There are 22 candidate implementations:
      - Of which 22 did not match due to:
        With argument(s): '(Tuple(int64, float64), int64)':
       No match.

During: typing of intrinsic-call at C:\Users\pconlon\AppData\Local\Temp/ipykernel_18736/1219060107.py (11)

File "..\..\..\..\..\AppData\Local\Temp\ipykernel_18736\1219060107.py", line 11:
```

What I expect to happen:
- the f2 should compiles just like f1 and does not have a compilation failure

stuartarchibald:
The guard code for checking a loop is conformant to what
literal_unroll can handle was incorrectly exiting the pass if it
encountered a non-conformant loop, e.g. a while loop, opposed to
just ignoring it.

Fixes #8311

### DIFF
--- a/numba/core/untyped_passes.py
+++ b/numba/core/untyped_passes.py
@@ -1043,8 +1043,8 @@ class MixedContainerUnroller(FunctionPass):
             # scan loop header
             iternexts = [_ for _ in
                          func_ir.blocks[loop.header].find_exprs('iternext')]
-            if len(iternexts) != 1:
-                return False
+            if len(iternexts) != 1: # needs to be an single iternext driven loop
+                continue
             for iternext in iternexts:
                 # Walk the canonicalised loop structure and check it
                 # Check loop form range(literal_unroll(container)))
--- a/numba/tests/test_mixed_tuple_unroller.py
+++ b/numba/tests/test_mixed_tuple_unroller.py
@@ -1818,6 +1818,18 @@ class TestMore(TestCase):
 
         self.assertEqual(foo(), foo.py_func())
 
+    def test_unroll_with_non_conformant_loops_present(self):
+        # See issue #8311
+
+        @njit('(Tuple((int64, float64)),)')
+        def foo(tup):
+            for t in literal_unroll(tup):
+                pass
+
+            x = 1
+            while x == 1:
+                x = 0
+
 
 def capture(real_pass):
     """ Returns a compiler pass that captures the mutation state reported


gmarkall (commented):
I think I'm ready to approve this, but I first want to confirm my understanding of the issue and patch is correct. I have resisted understanding entirely how the literal unroll pass works in detail, because there's a lot to it, but my general understanding is that it unrolls the loop and generates a switch table on the type of the "current" item containing the loop body so it can be typed for each member type as appropriate.

Anyway, I think what happened was:

- On the first iteration of applying the transform in `MixedContainerUnroller.run_pass()`, `apply_transform()` runs, transforms the `literal_unroll()` loop, and then moves on to looking at the `while` loop.
- It sees that the `while` loop is not in the correct form for a literal unroll, so it returns `False`
- Back in `run_pass`, `stat` gets set to `False`, and therefore `mutated` does too, and the iteration of applying the transform exits its "infinite loop".
- The function was changed, but `run_pass` returns `False` and signals that it did not change anything, and chaos ensues from the resulting inconsistency.

With the fix, the big `for` loop in `apply_transform` carries on looking at all loops even if some of them aren't in a suitable form for transformation, and therefore correctly reports that a change was made, so the subsequent pipeline operations get a consistent / correct view of things.

If this is accurate then I think the fix and test are good. If it's inaccurate I'll have another attempt at understanding the situation :slightly_smiling_face:

stuartarchibald:
> I think I'm ready to approve this, but I first want to confirm my understanding of the issue and patch is correct. I have resisted understanding entirely how the literal unroll pass works in detail, because there's a lot to it, but my general understanding is that it unrolls the loop and generates a switch table on the type of the "current" item containing the loop body so it can be typed for each member type as appropriate.
> 
> Anyway, I think what happened was:
> 
>     * On the first iteration of applying the transform in `MixedContainerUnroller.run_pass()`, `apply_transform()` runs, transforms the `literal_unroll()` loop, and then moves on to looking at the `while` loop.
> 
>     * It sees that the `while` loop is not in the correct form for a literal unroll, so it returns `False`
> 
>     * Back in `run_pass`, `stat` gets set to `False`, and therefore `mutated` does too, and the iteration of applying the transform exits its "infinite loop".
> 
>     * The function was changed, but `run_pass` returns `False` and signals that it did not change anything, and chaos ensues from the resulting inconsistency.
> 
> 
> With the fix, the big `for` loop in `apply_transform` carries on looking at all loops even if some of them aren't in a suitable form for transformation, and therefore correctly reports that a change was made, so the subsequent pipeline operations get a consistent / correct view of things.
> 
> If this is accurate then I think the fix and test are good. If it's inaccurate I'll have another attempt at understanding the situation slightly_smiling_face

Thanks for taking a look at this @gmarkall. I think your description is close to my assessment, but I think the root cause comes from this large loop:
https://github.com/numba/numba/blob/a06693d22fd2eced4d802af02809e41b08013e03/numba/core/untyped_passes.py#L1039-L1132
doing a scan of all loops to build up some analysis info into variable `literal_unroll_info` about the "literal_unroll" driven loops. The bug is that this analysis loop exits the entire function here if a non-conformant loop is found:
https://github.com/numba/numba/blob/a06693d22fd2eced4d802af02809e41b08013e03/numba/core/untyped_passes.py#L1046-L1047
which means there's no transformation made, hence the `literal_unroll` pass has no effect and the `getitem` on a heterogeneous tuple still exists which leads to the typing error.

gmarkall:
Ah, that explanation makes the error message make way more sense. The only way I could rationalise it was that something inconsistent made a lot of confusion down the line. Thanks for clarifying!

gmarkall (approved):
This looks good to me, now that I understand the issue properly! :slightly_smiling_face:

stuartarchibald:
> Ah, that explanation makes the error message make way more sense. The only way I could rationalise it was that something inconsistent made a lot of confusion down the line. Thanks for clarifying!

No problem. Thanks for the assessment and review.

PR #8266, opened 2022-07-20
Remove `mk_unique_var` in `untyped_passes.py`

guilhermeleobas:
xref: https://github.com/numba/numba/issues/8230

Issue #8230, opened 2022-07-06
Refactor tasks: remove `mk_unique_var`


sklam:
Reason: `mk_unique_var` uses global state that lives outside of the `Scope` object. `Scope.redefine` can be used instead. `Scope.redefine` provides a better way to manage variable names.

**NOTE**: The list below is left for historical reasons, but the plan is now to move `mk_unique_var()` out of `ir_utils` and into somewhere in parfors - see https://github.com/numba/numba/issues/8230#issuecomment-1478330246 - (edit by @gmarkall).

Files that references `mk_unique_var`:

- [x] `numba/core/inline_closurecall.py` (patch https://github.com/numba/numba/pull/8229)
- [x] `numba/core/untyped_passes.py`
- [x] `numba/parfors/array_analysis.py`
- [ ] `numba/parfors/parfor_lowering_utils.py`
- [ ] `numba/parfors/parfor_lowering.py`
- [ ] `numba/parfors/parfor.py`
- [x] `numba/stencils/stencil.py`
- [ ] `numba/stencils/stencilparfors.py`

Finally, remove the definition:

- [ ] `numba/core/ir_utils.py`

sklam:
Self reminder to address https://github.com/numba/numba/pull/8229#discussion_r915987381

sklam:
We now realize remaining of the task is very hard. See comment at https://github.com/numba/numba/pull/8820#issuecomment-1476450175

PR #8324, opened 2022-08-08
Remove use of mk_unique_var in untyped_passes.py

bszollosinagy:
Doing the necessary for Issue #8230 to remove the use of mk_unique_var in untyped_passes.py

I think this code section is exercised by the unit tests in test_mixed_tuple_unroller.py, so that is what I used to make sure I was not doing something harmful. It runs well.

Edit: I just saw that #8266 also addresses this topic.


bszollosinagy:
Only for Python 3.8, a unit test fails : 

- numba.tests.test_record_dtype.test_literal_unroll_dynamic_to_static_getitem_transform

I have reproduced, and it also fails on may machine for Python 3.8

bszollosinagy:
#8266 seems to cover what is needed, so closing this duplicate

esc:
@bszollosinagy thank you for closing!

Issue #8272, opened 2022-07-24
'CUDAUFuncDispatcher' object has no attribute '__name__'


s-m-e:
This might be a simple oversight, but anyway, I stumbled across this while testing and benchmarking:

```python
import numba as nb

@nb.vectorize('f8(f8)', target = 'cpu')
def foo(x):
    return x ** 2

@nb.vectorize('f8(f8)', target = 'cuda')
def bar(x):
    return x ** 2

assert foo.__name__ == 'foo'
assert bar.__name__ == 'bar'
```

The second assertion for `bar` fails, causing `AttributeError: 'CUDAUFuncDispatcher' object has no attribute '__name__'`. It would be useful if `numba` was consistent here.

stuartarchibald:
Thanks for the report. @testhound / @gmarkall please could you take a look at this as it's CUDA target, many thanks.

testhound:
@stuartarchibald created a PR to address this issue: https://github.com/numba/numba/pull/8297

PR #8320, opened 2022-08-08
Add __name__ support for GUFuncs

testhound:
This pull request addresses issue 8272: https://github.com/numba/numba/issues/8272 for GUFuncs

PR #8329, opened 2022-08-09
Add __name__ attribute to CUDAGUFuncVectorize plus test case

s-m-e:
This solves what is basically a variation of #8272: `CUDAGUFuncVectorize` does not have a `__name__` attribute.


gmarkall:
@s-m-e Many thanks for the PR! Unfortunately I think this duplicates #8320, which is already approved, buildfarm passed, and ready-to-merge.

PR #8331, opened 2022-08-09
Extend docs with info on how to call C functions from Numba

guilhermeleobas:
As suggested on [discourse](https://numba.discourse.group/t/how-to-call-the-externel-c-function-in-ir/1477/2), I'm extending the docs with information on how to call C functions from Numba

### DIFF
--- a/docs/source/user/cfunc.rst
+++ b/docs/source/user/cfunc.rst
@@ -234,7 +234,7 @@ integers and calling it within Numba ``@jit`` code
    }
 
 
-Compile the code with ``gcc lib.c -fpic -shared -o lib.so`` to generate a
+Compile the code with ``gcc lib.c -fPIC -shared -o shared_library.so`` to generate a
 shared library.
 
 .. code-block:: python
@@ -245,7 +245,7 @@ shared library.
    import os
 
    # load the library into LLVM
-   path = os.path.abspath('.//path/to/shared_library.so')
+   path = os.path.abspath('./shared_library.so')
    binding.load_library_permanently(path)
 
    # Adds typing information
--- a/docs/source/user/cfunc.rst
+++ b/docs/source/user/cfunc.rst
@@ -216,3 +216,41 @@ Compilation options
 A number of keyword-only arguments can be passed to the ``@cfunc``
 decorator: ``nopython`` and ``cache``.  Their meaning is similar to those
 in the ``@jit`` decorator.
+
+
+Calling C code from Numba
+=========================
+
+It is also possible to call C code from Numba ``@jit`` functions. In this
+example, we are going to be compiling a simple function ``sum`` that adds two
+integers and calling it within Numba ``@jit`` code::
+
+   #include <stdint.h>
+
+   int64_t sum(int64_t a, int64_t b){
+      return a + b;
+   }
+
+
+Compile the code with ``gcc lib.c -fpic -shared -o lib.so`` to generate a
+shared library. ::
+
+   from numba import njit
+   from numba.core import types, typing
+   from llvmlite import binding
+
+   # load the library into LLVM
+   binding.load_library_permanently('./path/to/lib.so')
+
+   # Adds typing information
+   c_func_name = 'sum'
+   return_type = types.int64
+   argty = types.int64
+   c_sig = typing.signature(return_type, argty, argty)
+   c_func = types.ExternalFunction(c_func_name, c_sig)
+
+   @njit
+   def example(x, y):
+      return c_func(x, y)
+
+   print(example(3, 4))
--- a/docs/source/user/cfunc.rst
+++ b/docs/source/user/cfunc.rst
@@ -226,6 +226,7 @@ example, we are going to be compiling a simple function ``sum`` that adds two
 integers and calling it within Numba ``@jit`` code
 
 .. code-block:: C
+
    #include <stdint.h>
 
    int64_t sum(int64_t a, int64_t b){
--- a/docs/source/user/cfunc.rst
+++ b/docs/source/user/cfunc.rst
@@ -242,9 +242,11 @@ shared library.
    from numba import njit
    from numba.core import types, typing
    from llvmlite import binding
+   import os
 
    # load the library into LLVM
-   binding.load_library_permanently('./path/to/lib.so')
+   path = os.path.abspath('.//path/to/shared_library.so')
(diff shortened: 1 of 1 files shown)


esc:
@guilhermeleobas this looks great! Can you use `code-block` for the highlighting!

https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-code-block

docs/source/user/cfunc.rst
```diff
@@ -216,3 +216,44 @@ Compilation options
...
+
+
+Calling C code from Numba
+=========================
+
+It is also possible to call C code from Numba ``@jit`` functions. In this
+example, we are going to be compiling a simple function ``sum`` that adds two
+integers and calling it within Numba ``@jit`` code
+
+.. code-block:: C
```
esc, line 228:
```suggestion
.. code-block:: C

```

I think this is the cause of the doc build error here:

```
Warning, treated as error:
/home/vsts/work/1/s/docs/source/user/cfunc.rst:228:Error in "code-block" directive:
maximum 1 argument(s) allowed, 3 supplied.

.. code-block:: C
   #include <stdint.h>

   int64_t sum(int64_t a, int64_t b){
      return a + b;
   }
```

esc (changes requested):
Some minor whitespace issues, then it's good to go.

stuartarchibald:
Out of interest, why not use Numba's `ctypes` support, it's more idiomatic python?

docs/source/user/cfunc.rst
guilhermeleobas: Hi @esc, just fixed the issue.

guilhermeleobas:
> Out of interest, why not use Numba's `ctypes` support, it's more idiomatic python?

Good question. No particular reason. I just didn't find in the docs a way to do this.

esc:
@guilhermeleobas this looks good. I think we should merge this now and then perhaps a second example that uses the ctypes variant? I think this could be interesting for many people.

docs/source/user/cfunc.rst
```diff
@@ -216,3 +216,45 @@ Compilation options
...
+shared library.
+
+.. code-block:: python
+
+   from numba import njit
+   from numba.core import types, typing
+   from llvmlite import binding
+
+   # load the library into LLVM
+   binding.load_library_permanently('./path/to/lib.so')
```
esc, line 247:
```suggestion
   binding.load_library_permanently('.//lib.so')  # path to lib.so
```

The reason I suggest this is because, when I tried this in my `/tmp` I stumbled over it. I think it would be better to be able to copy and paste the example and just have it work. I strongly presume that most uses will stuff the library next to the python file, so maybe the above would improve the user experience.

esc (changes requested):
Just one nitpick now.

docs/source/user/cfunc.rst
stuartarchibald: Suggest using `os.path` utilities to handle paths.

stuartarchibald:
The issue with binding to C libraries as demonstrated in this example is that if the `@njit` decorator is removed, the code stops working correctly. Whereas the use of `ctypes` works invariant of the JIT:
```python
from numba import njit
import ctypes
DSO = ctypes.CDLL('./libfoo.so')

# Add typing information
c_func = DSO.sum
c_func.restype = ctypes.c_int
c_func.argtypes = [ctypes.c_int, ctypes.c_int]

@njit
def example(x, y):
    return c_func(x, y)

print(example(3, 4)) # 7
print(example.py_func(3, 4)) # 7
```

guilhermeleobas:
Thanks, Stuart. I'll update the docs to use ctypes.

docs/source/user/cfunc.rst
```diff
@@ -216,3 +216,47 @@ Compilation options
...
+
+.. code-block:: python
+
+   from numba import njit
+   from numba.core import types, typing
+   from llvmlite import binding
+   import os
+
+   # load the library into LLVM
+   path = os.path.abspath('.//path/to/shared_library.so')
```
esc, line 248:
```suggestion
   path = os.path.abspath('./shared_library.so')
```

docs/source/user/cfunc.rst
```diff
@@ -216,3 +216,47 @@ Compilation options
...
+.. code-block:: C
+
+   #include <stdint.h>
+
+   int64_t sum(int64_t a, int64_t b){
+      return a + b;
+   }
+
+
+Compile the code with ``gcc lib.c -fpic -shared -o lib.so`` to generate a
```
esc, line 237:
```suggestion
Compile the code with ``gcc lib.c -fpic -shared -o shared_library.so`` to generate a
```

esc (changes requested):
I suggest making the following changes so as to make the example more user-friendly. Like this, the files can just be downloaded into the same directory to execute the example. No modification of the example files will be required.

docs/source/user/cfunc.rst
esc:
```suggestion
Compile the code with ``gcc lib.c -fPIC -shared -o shared_lib.so`` to generate a
```

esc (changes requested):
I think `PIC` will need to be capitalized.

esc:
@guilhermeleobas I just took another look at this and would like to make the following suggestions:

* Add a ctypes example and mention why this is superior (can run w/o Numba)
* Mention that this is likely to only work on a LInux (or at least UNIX) style OS

docs/source/user/cfunc.rst
```diff
@@ -216,3 +216,73 @@ Compilation options
...
+Calling C code from Numba
+=========================
+
+It is also possible to call C code from Numba ``@jit`` functions. In this
+example, we are going to be compiling a simple function ``sum`` that adds two
+integers and calling it within Numba ``@jit`` code.
+
+.. note::
+   The example below was tested on Linux and will likely work on Unix-like
+   Operating Systems.
```
esc, line 230:
```suggestion
   operating systems.
```

esc (changes requested):
I don't think Operating System is to be capitalized in this context. It is not a name of a product  or company or such, but rather a description of a certain class of program.

esc:
I tested the example locally and both variants did work fine.

esc (approved):

esc:
Thank you for the patch!

PR #8333, opened 2022-08-09
Remove dtype from kwarg list during function replacement in pre-parfor pass.

### LINKED ISSUE #8225: np.arange with kwargs not supported when parallel=True
## Reporting a bug

`parallel=True` with `np.arange` doesn't seem to support passing any arguments with keywords. Based on the code I believe this exists for all functions that are replaced, but here is a reproducer for `np.arange`.

```python3
import numpy as np
import numba

@numba.njit
def f():
    return np.arange(1, 100, dtype=np.float32)

@numba.njit(parallel=True)
def g():
    return np.arange(1, 100, dtype=np.float32)

print(f())
print(g())
```

DrTodd13:
Resolves #8225

Generally, dtype kwargs to NumPy functions only have an effect on the typing aspect of Numba and not any runtime ramifications.  So, in the parfor code, we can filter out dtype kwargs and rely on typing to handle them.  I wrote this in a generic way but the only function it may now affect is arange.  Many other functions that the pre-parfor pass replaces don't take a dtype parameter and others like linspace don't support dtype for typing either.

### DIFF
--- a/numba/parfors/parfor.py
+++ b/numba/parfors/parfor.py
@@ -1478,6 +1478,10 @@ class PreParforPass(object):
 
                             require(repl_func is not None)
                             typs = tuple(self.typemap[x.name] for x in expr.args)
+                            # The dtype keyword argument is handled by typing and can
+                            # be removed here during the implementation replacement so
+                            # that subsequent typing and argument handling won't fail.
+                            expr.kws = list(filter(lambda x: x[0] != "dtype", expr.kws))
                             try:
                                 new_func =  repl_func(lhs_typ, *typs)
                             except:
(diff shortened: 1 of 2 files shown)


eiis1000:
Any news on this? The issue it resolves is quite old and annoying

DrTodd13:
> Any news on this? The issue it resolves is quite old and annoying

Just added to 0.58 milestone so hopefully we can get it in that release.

sklam:
> I wrote this in a generic way but the only function it may now affect is arange. Many other functions that the pre-parfor pass replaces don't take a dtype parameter and others like linspace don't support dtype for typing either.

Given it only affects `arange`, I made an alternative PR to add `dtype` keyword handling in `array_parallel_impl`. That way we are not rewriting the `call` IR-node regardless and we don't have to worry that future extensions will add functions that uses `dtype` keyword into `replace_functions_map`.

Issue #8327, opened 2022-08-09
Support List as a value type in njit dictionaries


njriasan:
## Feature request

It doesn't appear to be possible to pass a list as a value inside a dictionary. Here is a simple reproducer.

```python3
import numba

@numba.njit
def f():
    d = {}
    for i in range(10):
        d[str(i)] = [i]
    return d

print(f())
```

Produces: `numba.core.errors.TypingError: list(int64)<iv=None> as value is forbidden`

guilhermeleobas:
It doesn't seem to be possible to support this one as reflected containers don't play well with typed containers.

```python
>>> from numba import typed
>>> d = typed.Dict()
>>> d[0] = [0]
>>> d[1] = [1]
>>>print(d)
{0: [], 1: ['0: []']}
```

esc:
@njriasan @guilhermeleobas you could use the `numba.typed.List` as showen below:

```
import numba


@numba.njit
def f():
    d = {}
    for i in range(10):
        d[str(i)] = numba.typed.List((1,))
    return d


print(f())
```

esc:
I do hope that at some point in the not too distant future `[]` in JIT-ed code will result in a typed List.

SomeAuthor123:
It seems a little weird to me, that while `typed.List` is allowed as value as per @esc example
```python
@numba.njit
def f():
    d = {}
    for i in range(10):
        d[str(i)] = numba.typed.List((1,))
    return d
print(f())
```
and seems to work just fine in my application 

The same code fails when assigning a set as value.
```python
@numba.njit
def f2():
    d = {}
    for i in range(10):
        d[str(i)] = {1}
    return d
print(f2())
```
 I know that the documentation states that neither List nor Set are supported. It still makes me wonder why one works and the other doesn't, especially because you can easily put `typed.Dict` in `typed.Dict` as value

```python
@numba.njit
def f3():
    d = {}
    for i in range(10):
       tmp = numba.typed.Dict()
       tmp[i] = True
        d[str(i)] = tmp
    return d
print(f3())
```

Not sure whether this is the right place or i should put it in a different issue ;)

esc:
> It seems a little weird to me, that while `typed.List` is allowed by the compiler as value as per @esc's example
> 
> ```python
> @numba.njit
> def f():
>     d = {}
>     for i in range(10):
>         d[str(i)] = numba.typed.List((1,))
>     return d
> print(f())
> ```
> 
> and seems to work just fine in my application
> 
[...]

PR #8337, opened 2022-08-10
Support list/sets as dict value type

guilhermeleobas:
Fix issue #8327


guilhermeleobas:
I'll close this one since reflected containers are unstable in jitted code.

ehsantn:
Can we check the reflected flags and allow non-reflected cases?

PR #8338, opened 2022-08-10
Calculate reductions before fusion so that use of reduction vars can stop fusion.

### LINKED ISSUE #8007: Wrong results  when using jit(parallel=True)
This bug report is very similar to issue #7984 .
Using Python 3.8.7,  numba-0.55.1 with llvmlite-0.38.0, Under Win-10:

```
from numpy import ndarray
import numpy as np
import numba
from numba import jit


def test_par(img: ndarray) -> ndarray:

    rad = 5
    w2 = rad + rad

    min_value = np.amin(img)
    # print("")
    img = img - min_value
    s = img.shape
    a_p = np.zeros((s[0] + w2, s[1] + w2), dtype=np.int32)
    a_p[rad:rad+s[0], rad:rad+s[1]] = img

    b = np.zeros(s, dtype=np.float32)

    for i in range(rad):
        for j in range(rad):
            a_1 = a_p[i:i+s[0], j:j+s[1]]
            b = b + a_1
    
    return b

test_np = test_par
test_nb = numba.jit(nopython=True, fastmath=True, parallel=False)(test_par)
test_nb_p = numba.jit(nopython=True, fastmath=True, parallel=True)(test_par)

if __name__ == '__main__':
    image = np.empty((256, 256), dtype=np.uint8)
    for k in range(256):
        for m in range(256):
            image[k, m] = k

    res0 = test_np(image)
    res1 = test_nb(image)
    res2 = test_nb_p(image)

    diff1 = np.max(np.abs(res1-res0))
    print(diff1)
    diff2 = np.max(np.abs(res2-res0))
    print(diff2)

```

The results are:
0.0
4250.0

There seems to be a race condition that causes different results for` parallel=true.`
Strangely enough, when uncommenting the  `print("") `command, the results are OK.
Looks like the print command stops somehow the race condition (like openning the box of Schrödinger's cat:-)).

Tracking back the problem,  on numba-0.52.0 &  llvmlite-0.35.0 this problem does not exist,
but on  numba-0.53.0 &  llvmlite-0.36.0 it appears.

I hope that you can reproduce this bug.

Thanks,

DrTodd13:
Resolves #8007

### DIFF
--- a/numba/parfors/parfor.py
+++ b/numba/parfors/parfor.py
@@ -2902,6 +2902,20 @@ class ParforPass(ParforPassStates):
             self.func_ir._definitions = build_definitions(self.func_ir.blocks)
             self.array_analysis.equiv_sets = dict()
             self.array_analysis.run(self.func_ir.blocks)
+
+            # Get parfor params to calculate reductions below.
+            _, parfors = get_parfor_params(self.func_ir.blocks,
+                                           self.options.fusion,
+                                           self.nested_fusion_info)
+
+            # Find reductions so that fusion can be disallowed if a
+            # subsequent parfor read a reduction variable.
+            for p in parfors:
+                p.redvars, p.reddict = get_parfor_reductions(self.func_ir,
+                                                             p,
+                                                             p.params,
+                                                             self.calltypes)
+
             # reorder statements to maximize fusion
             # push non-parfors down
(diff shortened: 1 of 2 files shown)


stuartarchibald (approved):
Thanks for the patch, the changes make sense as a method of preventing fusion through reduction variables.

Issue #7710, opened 2022-01-06
CUDA: Don't use typemap for function name and line number 


gmarkall:
Following discussion around #7704 and #7709 it becomes clear that we shouldn't really rely on the type annotation for the function name and line number in CUDA's `compile_ptx` function. We should remove the use of the type annotation for this and instead get them from the Python function.

We will also need to add a test that the type annotation is present when compiling in the CUDA target, to ensure it is available for display (e.g. with `numba --annotate-html`).

PR #8343, opened 2022-08-12
Eliminate references to type annotation in compile_ptx

testhound:
Eliminate use of type annotation when referencing filenumber and line number of a python function. This issue partially addresses issue [7710](https://github.com/numba/numba/issues/7710)

### DIFF
--- a/numba/cuda/compiler.py
+++ b/numba/cuda/compiler.py
@@ -11,7 +11,6 @@ from numba.core.typed_passes import (IRLegalization, NativeLowering,
                                      AnnotateTypes)
 from warnings import warn
 from numba.cuda.api import get_current_device
-from numba.core.bytecode import FunctionIdentity
 
 
 def _nvvm_options_type(x):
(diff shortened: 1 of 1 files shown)


testhound:
gpuci run tests

numba/cuda/compiler.py
```diff
@@ -272,8 +273,9 @@ def compile_ptx(pyfunc, args, debug=False, lineinfo=False, device=False,
         lib = cres.library
     else:
         tgt = cres.target_context
-        filename = cres.type_annotation.filename
-        linenum = int(cres.type_annotation.linenum)
+        func_identity = FunctionIdentity.from_function(pyfunc)
+        filename = func_identity.filename
+        linenum = func_identity.firstlineno
```
gmarkall, line 278: Could we just do this similarly to how  `_Kernel` does it instead of introducing a dependency between the `bytecode` module and the CUDA `compiler` module,

https://github.com/numba/numba/blob/05e5e3407ba0ff3fb454943e03f6d76255556cbe/numba/cuda/dispatcher.py#L82-L84

 i.e.:

```suggestion
        code = pyfunc.__code__
        filename = code.co_filename
        linenum = code.co_firstlineno
```
?

gmarkall (changes requested):
Thanks for the PR - one question on the diff.

testhound:
gpuci run tests

numba/cuda/compiler.py
testhound: @gmarkall PR updated as requested.

gmarkall (approved):
This looks good to me. Many thanks!

PR #8347, opened 2022-08-13
Use pre-commit to make Flake8 pass again

jamesobutler:
I am issuing this PR after reviewing the contributing section regarding coding conventions (see below) and then noticing that the `main` branch was failing with the current configuration. 

https://github.com/numba/numba/blob/2e9b58cfce0391b968ab8c7c84393cc12c4d1bfb/docs/source/developer/contributing.rst#L255-L281

I've refreshed the `.pre-commit-config.yml` which hasn't had much interaction in the past few years and likely a reason for `main` to not be in compliance with the stated coding conventions. To keep the code base in compliance, I suggest a code owner attaching the numba repository to https://pre-commit.ci/ which will run an additional status check (see image below) and has the benefit of being able to contribute fixes to identified style issues automatically to PRs. This keeps developers doing the more important things by automating the fix me up style tasks. I utilized it in https://github.com/jamesobutler/numba/pull/1 by making a change to the pre-commit-config and then letting pre-commit.ci push a commit to fix things up and get things back in compliance with the new configuration. 

image

I let pre-commit-ci do the changes so that the contribution would be associated with a bot rather than my personal account. These bulk changes could also be added to a new `.git-blame-ignore-revs` file in this repository so that they could be ignored when git blaming locally and even while on GitHub.com (see https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/).

This PR aims to supersede #7195 and #7224


jamesobutler:
I’m currently exploring the test failures and will follow up here. Once passing I will mark as ready for review. Of course any early feedback is appreciated

apmasell:
Thank you for the contribution. It would be really nice to finish formatting the codebase. This is a great change, but it's a big larger than we can review at once. Could we split it into smaller pieces? We know that's going to be a _lot_ of pieces.

The changed files select which unit tests get activated, so the failed unit tests are likely because too many are trying to run and hitting the timeout. Also, a change this big is going to disrupt a lot of other inflight PRs, so, if we move in smaller chunks, we can minimize the impact.

Although you ran autopep8 and it's an automated, we still need to review all the changes as if they were done by hand, partially to decide if some of the places we have weird formatting should remain so because they are more semantically meaningful.

Could we start with just the formatting configuration changes?

jamesobutler:
Hi, I will plan to break this up into bite sized PRs. I can aim to limit each PR to a certain set of formatting changes that are all related where it is the same type change just repeated over and over across many files. I think that will be easier review rather than fully formatting specific files which could have many different formatting changes contained in the specific file.

Will the above approach work considering the below note?

> The changed files select which unit tests get activated, so the failed unit tests are likely because too many are trying to run and hitting the timeout.

apmasell:
I don't think there's an easy way to know how big it can get before the unit tests will timeout since not all the unit tests are equally fast. I think the human limit is probably lower than the unit test limit though.

PR #8351, opened 2022-08-15
switch RTD from mamba to miniconda

esc:
As title

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -2,7 +2,7 @@ version: 2
 build:
   os: ubuntu-20.04
   tools:
-    python: mambaforge-4.10
+    python: miniconda3-4.7
 sphinx:
   configuration: docs/source/conf.py
 python:


esc:
Offered as an alternative to #8352 

We had switch to mamba on RTD some time ago to workaround a missing feature in conda. We had always wanted to switch back to miniconda, since this is more consistent with out public CI setup (Azure Pipelines).

stuartarchibald:
@esc I think it'd make sense to accept both this and #8352. This patch is "working" as Anaconda distribution channels don't have the `setuptools` version which is broken for Numba's purposes, but as soon as `setuptools` version 65 is released into Anaconda distribution channels it'll break again.

1. I think accepting this patch is a good idea as it means the doc builds are performed using the same distribution/tooling as the rest of Numba's public CI.
2. I think accepting #8352 is a good idea as it prevents the use of a package broken for Numba's use case (I suspect NumPy/Numba package constraints might need also adjusting to accommodate).

stuartarchibald (approved):
Thanks for the patch. I've checked the RTD docs, the `miniconda` version used in this change is the one listed in the `build.tools.python` section. Also think it's a good idea to use the same tools/channels for building the CI docs as the rest of Numba's public CI, i.e. `miniconda`.

stuartarchibald:
Closing, merged as part of #8354.

PR #8353, opened 2022-08-15
Test setuptools < 65 with gpuCI

gmarkall:
Based on #8352

PR #8352, opened 2022-08-15
pin setuptools to 64 for doc builds

esc:
As title

PR #8354, opened 2022-08-16
pin setuptools to < 65 and switch from mamba to conda on RTD

esc:
This is a combination #8351 #8352 and #8353 -- in order to get docs building on `main` again.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -2,7 +2,7 @@ version: 2
 build:
   os: ubuntu-20.04
   tools:
-    python: mambaforge-4.10
+    python: miniconda3-4.7
 sphinx:
   configuration: docs/source/conf.py
 python:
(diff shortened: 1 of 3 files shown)


gmarkall:
gpuci run tests

esc:
Note: this should probably be path of 0.56.1 too, to make sure we can get a clean docs build there too.

stuartarchibald (approved):
Thanks for the patch. Assuming the docs builds now pass, this can be merged.

stuartarchibald:
> Note: this should probably be part of 0.56.1 too, to make sure we can get a clean docs build there too.

I agree, it's a good idea to preempt the issue for 0.56.1.

gmarkall (approved):
gpuCI-related changes look good to me.

esc:
With #8356 -- we prefer that PR over this one. Hence this is no longer included in the 0.56.1 milestone.

Issue #7774, opened 2022-01-24
inconsistent array order  for bare and qualified jit decorator


yt87:
Consider the following code:

```
import numpy as np
import numba as nb

N = 4
M = 2
a = np.arange(M*N, dtype=np.float64).reshape(M, N).T
b = np.ones(M, dtype=np.float64)

def calc0(x, y, out=None):
    op = np.multiply
    it = np.nditer((x, y, None), [], [['readonly'], ['readonly'], ['writeonly', 'allocate']])
    for (a, b, c) in it:                                
        op(a, b, out=c)
    return it.operands[2]

@nb.njit
def calc1(x, y, out):
    it = np.nditer((x, y))
    for i, (u, v) in enumerate(it):
        out[i] = u * v
        
@nb.njit(nb.void(nb.float64[:,:], nb.float64[:], nb.float64[:]))
def calc2(x, y, out):
    it = np.nditer((x, y))
    for i, (u, v) in enumerate(it):
        out[i] = u * v
        
def calc1_wrap(x, y):
    shape = np.broadcast_shapes(x.shape, y.shape)
    size = functools.reduce(lambda u, v: u*v, shape)
    out = np.empty(size, dtype=np.float64)
    calc1(x, y, out)
    return out.reshape(shape)
        
def calc2_wrap(x, y):
    shape = np.broadcast_shapes(x.shape, y.shape)
    size = functools.reduce(lambda u, v: u*v, shape)
    out = np.empty(size, dtype=np.float64)
    calc2(x, y, out)
    return out.reshape(shape)

a * b
calc0(a, b)
calc1_wrap(a, b)
calc2_wrap(a, b)
        
calc1.signatures
calc2.signatures
```
The output is:
```
array([[0., 4.],
       [1., 5.],
       [2., 6.],
       [3., 7.]])

array([[0., 4.],
       [1., 5.],
       [2., 6.],
       [3., 7.]])

array([[0., 1.],
       [2., 3.],
       [4., 5.],
       [6., 7.]])

array([[0., 4.],
       [1., 5.],
       [2., 6.],
       [3., 7.]])

[(array(float64, 2d, F), array(float64, 1d, C), array(float64, 1d, C))]
[(array(float64, 2d, A), array(float64, 1d, A), array(float64, 1d, A))]
```
The third array stands out. It seems that the bare ``@njit`` assumes that the ``a`` is in Fortran order. This is not consistent with numpy behaviour. Is there a reason for this, or a minor bug? 

[...]

PR #8349, opened 2022-08-15
NumPy 1.23 testing patch

stuartarchibald:
Do not merge!

PR #8358, opened 2022-08-16
Added runtime check for np.nditer's input array layouts

kc611:
Fixes: https://github.com/numba/numba/issues/7774

A simple run-time check to see if array layouts are defined (either `C` or `F` continuous) in `np.nditer`. It's implementation is made under assumption that layout of arguments are either `C` or `F` continuous but there aren't any checks that enforced this.

### DIFF
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -3829,6 +3829,10 @@ def make_array_nditer(context, builder, sig, args):
     nditerty = sig.return_type
     arrtys = nditerty.arrays
 
+    if any(_arrty.layout not in 'CF' for _arrty in arrtys):
+        raise TypeError('Only C or F contiguous arrays are '
+                        'accepted by Numba implementation of np.nditer')
+
     if isinstance(sig.args[0], types.BaseTuple):
         arrays = cgutils.unpack_tuple(builder, args[0])
     else:


numba/np/arrayobj.py
```diff
@@ -3829,6 +3829,10 @@ def make_array_nditer(context, builder, sig, args):
     nditerty = sig.return_type
     arrtys = nditerty.arrays
 
+    if any(_arrty.layout not in 'CF' for _arrty in arrtys):
+        raise TypeError('Only C or F contiguous arrays are '
+                        'accepted by Numba implementation of np.nditer')
```
stuartarchibald, line 3834: This check is being made at compile time? What if the array is typed as `A` ordered but has a runtime layout of `F`? Example:
```python
from numba import njit, typeof
import numpy as np

data = np.array([[1., 3, 2, 7,],
                 [-5, 4, 2, 3,],
                 [9, -3, 1, 1,],
                 [2, -2, 2, 8,]], order='F')

reshaped_data = data.reshape((1, 4, 4))

@njit
def foo(arr):
    tmp = arr[0]
    acc = 0.
    for ix in np.nditer(tmp):
        acc += np.take(ix, 0)
    return acc

foo(reshaped_data)

foo.inspect_types()
```
Also, if an assertion is being made about supported types it needs to be done during type inference, not lowering.

stuartarchibald:
Thanks for the patch @kc611,  I've provided some initial feedback inline.

numba/np/arrayobj.py
stuartarchibald: IIRC `np.ndarray.flags.{c, f}_contiguous` works in `nopython` mode, had to do something similar here #8349.

esc:
@kc611 can you convert this from draft please? Thank you.

esc:
@stuartarchibald @kc611 -- this can probably be in the 0.57 milestone?

numba/np/arrayobj.py
```diff
@@ -3235,7 +3250,21 @@ def check_shape(shape, main_shape):
                                     main_shape_ty)
                     context.compile_internal(builder, check_shape,
                                              sig, (arr.shape, main_shape))
-
+                    tyctx = context.typing_context
+                    fnty = tyctx.resolve_value_type(_arr_contig)
+                    _arr_contig_sig = fnty.get_call_type(tyctx, (arrty,), {})
+                    impl = context.get_function(fnty, _arr_contig_sig)
+                    _arr_contig_result = impl(builder, (arr,))
```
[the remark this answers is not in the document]
stuartarchibald: Variable `arrays` is being reassigned here (which impacts the type of induction variable `arr`):
https://github.com/numba/numba/blob/64c06beb0551117efc3b3687d9d66734e391f603/numba/np/arrayobj.py#L3243
if you work around that, it'll probably work fine as the types will then be what's expected.

PR #8359, opened 2022-08-16
adding warnings about cache behaviour

luk-f-a:
Copies warning about cache behaviour from developer section of docs into user section of docs.

### DIFF
--- a/docs/source/user/jit.rst
+++ b/docs/source/user/jit.rst
@@ -199,7 +199,7 @@ a file-based cache.  This is done by passing ``cache=True``::
       This carries the risk that "old" function code might be used in the
       calculations.
     - Global variables are treated as constants. The cache will remember the value
-      in the global variable used at compilation. On cache load, the cached
+      of the global variable at compilation time. On cache load, the cached
       function will not rebind to the new value of the global variable.
 
 .. _parallel_jit_option:
--- a/docs/source/user/jit.rst
+++ b/docs/source/user/jit.rst
@@ -192,7 +192,7 @@ a file-based cache.  This is done by passing ``cache=True``::
       function-by-function basis. The cached function is the the main jit
       function, and all secondary functions (those called by the main
       function) are incorporated in the cache of the main function.
-    - Cache invalidation fails to recognize changes in symbols defined in a
+    - Cache invalidation fails to recognize changes in functions defined in a
       different file. This means that when a main jit function calls
       functions that were imported from a different module, a change in those
       other modules will not be detected and the cache will not be updated.
--- a/docs/source/user/jit.rst
+++ b/docs/source/user/jit.rst
@@ -183,6 +183,25 @@ a file-based cache.  This is done by passing ``cache=True``::
    def f(x, y):
        return x + y
 
+
+
+.. note::
+    Caching of compiled functions has several known limitations:
+
+    - The caching of compiled functions is not performed on a
+      function-by-function basis. The cached function is the the main jit
+      function, and all secondary functions (those called by the main
+      function) are incorporated in the cache of the main function.
+    - Cache invalidation fails to recognize changes in symbols defined in a
+      different file. This means that when a main jit function calls
+      functions that were imported from a different module, a change in those
+      other modules will not be detected and the cache will not be updated.
+      This carries the risk that "old" function code might be used in the
+      calculations.
+    - Global variables are treated as constants. The cache will remember the value
+      in the global variable used at compilation. On cache load, the cached
+      function will not rebind to the new value of the global variable.
+
 .. _parallel_jit_option:
 
 ``parallel``
--- a/docs/source/developer/caching.rst
+++ b/docs/source/developer/caching.rst
@@ -69,7 +69,7 @@ This is a list of known limitation of the cache:
 - Cache invalidation fails to recognize changes in symbols defined in a
   different file.
 - Global variables are treated as constants. The cache will remember the value
-  in the global variable used at compilation. On cache load, the cached
+  of the global variable at compilation time. On cache load, the cached
   function will not rebind to the new value of the global variable.
 
 


gmarkall (dismissed):
Thanks for the PR! This looks good to me:

- The rendered docs look good, no formatting errors.
- The language is clear and appropriate for the early section of the docs that these notes are in.
- It's not an exact duplicate of the developer docs notes, which are more concise. I think it's appropriate to both add this introductory warning section, and to keep the notes in the developer section.

gmarkall:
@esc / @stuartarchibald any further opinions on this before making this RTM?

docs/source/user/jit.rst
```diff
@@ -183,6 +183,25 @@ a file-based cache.  This is done by passing ``cache=True``::
...
+
+
+.. note::
+    Caching of compiled functions has several known limitations:
+
+    - The caching of compiled functions is not performed on a
+      function-by-function basis. The cached function is the the main jit
+      function, and all secondary functions (those called by the main
+      function) are incorporated in the cache of the main function.
+    - Cache invalidation fails to recognize changes in symbols defined in a
```
stuartarchibald, line 195:
```suggestion
    - Cache invalidation fails to recognize changes in functions defined in a
```
whilst `symbols` is probably more correct given what goes on, as this are user facing documentation, it might not be known to users what actually happens when code is compiled and the associated terminology, so I wonder if `functions` better serves this purpose?

docs/source/user/jit.rst
```diff
@@ -183,6 +183,25 @@ a file-based cache.  This is done by passing ``cache=True``::
...
+      different file. This means that when a main jit function calls
+      functions that were imported from a different module, a change in those
+      other modules will not be detected and the cache will not be updated.
+      This carries the risk that "old" function code might be used in the
+      calculations.
+    - Global variables are treated as constants. The cache will remember the value
+      in the global variable used at compilation. On cache load, the cached
+      function will not rebind to the new value of the global variable.
```
stuartarchibald, line 203: I agree with "globals are constants", this is something Numba does invariant of cache. Out of curiosity, I'm wondering if there is a situation where mutating a global would not impact the `mtime` of the source file (or similar property), which would end up invalidating the cache anyway? Can this perhaps be done with closures?

stuartarchibald (commented):
@luk-f-a thanks for adding this user documentation, these notes certainly help to better explain the current behaviour of the on-disk cache. Couple of queries to resolve else looks good.

stuartarchibald:
> @esc / @stuartarchibald any further opinions on this before making this RTM?

I've a couple of queries that have been added as a review. Thanks for taking a look at this @gmarkall.

docs/source/user/jit.rst
luk-f-a: yes, you're right. This will make it clearer.

stuartarchibald: Ah, you can do it of course by mutating a reference.

`reference_lib.py`:
```python
import numpy as np
some_array = np.ones(10)
```
`test8359.py`:
```python
from numba import njit
import numpy as np
from reference_lib import some_array

@njit(cache=True)
def foo(idx):
    return some_array[idx]
```
`call_test8359.py`:
```python
from test8359 import foo

foo(3)
```

Changes can be made to `reference_lib.some_array` and the cache won't notice because the module in which the cached function is defined isn't changing.

luk-f-a: what if the global is defined in a separate file? would numba track the source file of the global or only the source file of the function? 

>  Can this perhaps be done with closures?

I assume "can be done" refers to a future PR on improving the cache invalidation, is that correct? it's not a question on this PR

stuartarchibald: > what if the global is defined in a separate file? would numba track the source file of the global or only the source file of the function?

Example above, I guess we were typing at the same time!

> > Can this perhaps be done with closures?
> 
> I assume "can be done" refers to a future PR on improving the cache invalidation, is that correct? it's not a question on this PR

Apologies, this wasn't particularly clear, I was wondering if creating the state where Numba captures a global and the cache isn't invalidated on changing this global could be achieved in the same module through the use of a closure. After some additional thought about this, I think it'd essentially just be an extension of state coming from an outside source, e.g. something providing the freevar in a closure generating a cached jit function.

docs/source/user/jit.rst
```diff
@@ -183,6 +183,25 @@ a file-based cache.  This is done by passing ``cache=True``::
...
+      function) are incorporated in the cache of the main function.
+    - Cache invalidation fails to recognize changes in functions defined in a
+      different file. This means that when a main jit function calls
+      functions that were imported from a different module, a change in those
+      other modules will not be detected and the cache will not be updated.
+      This carries the risk that "old" function code might be used in the
+      calculations.
+    - Global variables are treated as constants. The cache will remember the value
+      in the global variable used at compilation. On cache load, the cached
+      function will not rebind to the new value of the global variable.
```
stuartarchibald, line 203:
```suggestion
    - Global variables are treated as constants. The cache will remember the value
      of the global variable at compilation time. On cache load, the cached
      function will not rebind to the new value of the global variable.
```
Suggest this edit.

docs/source/user/jit.rst
stuartarchibald: I've suggested a minor edit as a further review comment, I think the rest of this paragraph is sufficient, thanks for the discussion. The reason I am of the view

> On cache load, the cached function will not rebind to the new value of the global variable.

is correct is as follows:

1. if the module is mutated there will not be a _cache load_.
2. if the module is not mutated and the global value is a reference which has been mutated, then there will not be a rebind.

luk-f-a: This part I copied from the developer documentation. If it's ok for you as it stands, then no need to adjust it here or in the developer section.

docs/source/user/jit.rst
luk-f-a: I merged your suggestion. As this text came from the developer docs, I changed it also there and pushed the change in my latest commit. I didn't make any other changes since your review, only backporting this change.

docs/source/user/jit.rst
stuartarchibald: I agree, thanks!

PR #8361, opened 2022-08-17
Enforce Flake8 E7 statement rules with help of autopep8

jamesobutler:
This is work branched off original PR https://github.com/numba/numba/pull/8347. This utilizes autopep8 by way of https://pre-commit.ci/ automatically pushing fixes as completed in https://github.com/jamesobutler/numba/pull/3. Work will continue in future PRs to remove the ignored error codes which will then trigger autopep8 to fix things up automatically.


jamesobutler:
All checks have passed so I have converted this from draft to ready for review

.flake8
```diff
@@ -1,248 +1,46 @@
...
+    E26,
+    # Assigning lambda expression
+    E731,
+    # Ambiguous variable names
+    E741,
+    # line break before binary operator
+    W503,
+    # line break after binary operator
+    W504,
+    E1,
```
gmarkall, line 18: Does the addition of E1 and the following lines mean that it will now be possible to introduce violations into the CUDA target files (`numba/cuda`)? They were not part of the exclude list (I fixed all violations in the CUDA target a couple of years ago and removed them from the exclude list), but will this these ignores will now apply to them?

jamesobutler: Yes the ignore list was increased as part of `Temporarily ignore all flake8 codes` so that pre-commit-ci could run to autofix things. I could add follow up commits that do revert both `Temporarily ignore all flake8 codes` and `Remove grandfathered flake8 file exceptions` to keep the current state. I'm in a tough situation where I can't apply all bulk changes at once (see #8347), but to iterate in tiny bits I need to ignore more and then remove them again in subsequent PRs.

gmarkall:
Moved back to "In progress" whilst checks are failing.

jamesobutler:
@gmarkall It seems unusual that the previous passing CI now has only Linux py310_np121 failing after including the latest commits to `main`. It doesn't seem like have the priviliges to "re-run failed" jobs. Can you trigger that for me? I would like to see it try without hitting a timeout one more time, but only the failed test running. I wouldn't want to unnecessarily trigger all the other jobs that are already passing to be run again.

gmarkall:
Re-running the failed one now.

jamesobutler:
@gmarkall after the re-run it looks like everything is passing again.

numba/tests/test_operators.py
```diff
@@ -1518,10 +1518,10 @@ def test_impl2(a):
             return a == b_val
 
         def test_impl3():
-            r1 = True == True
-            r2 = True == False
-            r3 = False == True
-            r4 = False == False
+            r1 = True is True
```
stuartarchibald, line 1521: I think this is incorrect, `operator.eq` is being tested not `operator.is_` (same for the change below with `s/!=/is not/`).

jamesobutler: Since there will be no place in the code comparing to a singleton like True/False/None, then this test can be removed?

stuartarchibald: There will indeed be no place in _Numba's_ code base that does this. But Numba is a compiler and the input space to the compiler is "literally any valid Python code a user can write", regardless of whether it is `flake8` compliant or even compilable by Numba.

Whilst it might be desirable for input code to not contain things like `operator.eq` on booleans, in practice, it probably does, and whilst breaking support for this might enforce code bases into `flake8` compliance, it'd also potentially break code already in production. Further, part of Numba's aim is to replicate the Python language semantics and `operator.eq(boolean, boolean)` is  valid Python.

As a result, I think this test and implementation has to stay as-is for as long as it is legal in Python to do e.g. `True == True`, as the test is checking that Numba can correctly compile this expression.

jamesobutler: I've removed these diff changes from the `Manually fix additional Flake8 E7 issues not fixed by autopep8` commit and updated `Enforce Flake8 E7 statement rules` to include the inline exception for E712 in these specific cases.

PR #8368, opened 2022-08-18
Remove `glue_lowering` in random math that requires IR

apmasell:
Change any random number generation code that's using `@glue_lowering` with
explicit IR generation to use `@intrinsic` and `@overload`.


guilhermeleobas:
Hi @apmasell, I don't think the CI failures are related to your changes, but can you restart CI just to be sure?

apmasell:
It's a know problem with packaging right now. Unfortunately, restarting it won't fix the problem. We will have to wait until they've fixed the packages.

apmasell:
Yay. Packages are fixed!

numba/cpython/randomimpl.py
```diff
@@ -203,57 +209,146 @@ def _fill_defaults(context, builder, sig, args, defaults):
...
+            scale_preprocessor = _double_preprocessor(scale)
+            return signature(types.float64, loc, scale),\
+                   _gauss_impl("np", loc_preprocessor, scale_preprocessor)
+        return lambda loc, scale: _impl(loc, scale)
+
+
+@overload(np.random.standard_normal)
+def standard_normal_impl1(size):
+    if isinstance(size, types.NoneType):
+        return lambda size: np.random.standard_normal()
```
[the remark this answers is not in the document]
guilhermeleobas: I just notice this pattern happens in other functions. You can choose if you merge all the common overloads into a single one or leave the way it is now.

guilhermeleobas (changes requested):
Hi @apmasell, thanks for working on this. I did a round of review. One comment I have which you can choose to address is wheter split the implementation of a function into different overloads or merge them all in a single one.

Also, I notice the `_fill_defaults` function is not used anymore and can be removed.

numba/cpython/randomimpl.py
```diff
@@ -203,57 +209,146 @@ def _fill_defaults(context, builder, sig, args, defaults):
...
+    def _impl(typingcontext, seed):
+        def codegen(context, builder, sig, args):
+            seed_value, = args
+            fnty = ir.FunctionType(ir.VoidType(), (rnd_state_ptr_t, int32_t))
+            fn = cgutils.get_or_insert_function(builder.function.module, fnty,
+                                                'numba_rnd_init')
+            builder.call(fn, (get_state_ptr(context, builder, state_type),
+                              seed_value))
+            return context.get_constant(types.none, None)
+        return signature(types.void, types.uint32), codegen
```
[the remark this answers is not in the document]
apmasell: The function prototype is unsigned though:

```
NUMBA_EXPORT_FUNC(void)
numba_rnd_init(rnd_state_t *state, unsigned int seed)
```

numba/cpython/randomimpl.py
```diff
@@ -203,57 +209,146 @@ def _fill_defaults(context, builder, sig, args, defaults):
...
+            return get_next_double(context, builder, state_ptr)
+        return signature(types.double), codegen
+    return lambda: _impl()
+
+
+@overload(np.random.random)
+@overload(np.random.random_sample)
+@overload(np.random.sample)
+@overload(np.random.ranf)
+def random_impl(size):
```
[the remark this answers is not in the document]
apmasell: There's a version with no arguments before. I will rename to avoid confusion.

numba/cpython/randomimpl.py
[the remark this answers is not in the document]
apmasell: I'd rather leave it. Some of the functions change the argument order depending on the number of provided arguments.

apmasell:
I think I've made all the necessary changes.

guilhermeleobas (dismissed):
Looks good to me. Thanks @apmasell

gmarkall:
Can this go to RTM?

numba/cpython/randomimpl.py
```diff
@@ -191,69 +199,146 @@ def get_shifted_int(nbits):
...
+@overload(np.random.random_sample)
+@overload(np.random.sample)
+@overload(np.random.ranf)
+def random_impl0():
+    @intrinsic
+    def _impl(typingcontext):
+        def codegen(context, builder, sig, args):
+            state_ptr = get_state_ptr(context, builder, "np")
+            return get_next_double(context, builder, state_ptr)
+        return signature(types.double), codegen
```
stuartarchibald, line 250: Whilst aliased, these `types.double` should probably be `types.float64` to reflect the NumPy semantics of a `float64` return type.

numba/cpython/randomimpl.py
```diff
@@ -392,68 +507,188 @@ def get_num():
...
+        return lambda low, high, size: np.random.randint(low, high)
+    if (isinstance(low, types.Integer) and isinstance(high, types.Integer) and
+       (isinstance(size, types.Integer) or (isinstance(size, types.UniTuple)
+                                            and isinstance(size.dtype,
+                                                           types.Integer)))):
+        bitwidth = max(low.bitwidth, high.bitwidth)
+        result_type = {8: np.int8,
+                       16: np.int16,
+                       32: np.int32,
+                       64: np.int64}[bitwidth]
```
stuartarchibald, line 607: perhaps use `getattr(np, f'int{bitwidth}')` ?

numba/cpython/randomimpl.py
```diff
@@ -392,68 +507,188 @@ def get_num():
...
+
+@overload(np.random.uniform)
+def np_uniform_impl1(low):
+    if isinstance(low, types.Number):
+        return lambda low: np.random.uniform(low, 1.0)
+
+
+@overload(random.uniform)
+def uniform_impl2(low, high):
+    if isinstance(low, types.Number) and isinstance(high, types.Number):
```
stuartarchibald, line 642: I'm slightly uncomfortable with accepting `Number` here (and throughout, it's a common occurrence in this PR) and then deferring the type check to e.g. `_double_preprocessor` because it's pushing the error further down the stack and relying on the presence of a call to something else to catch a problem. Would checking if the types are in `types.real_domain | types.integer_domain` be sufficient and ensure the error is raised from this overload opposed to the intrinsic call?

numba/cpython/randomimpl.py
```diff
@@ -1304,66 +1539,91 @@ def _impl(n, p):
...
+                    X = 0
+                    prod = 1.0
+                    while 1:
+                        U = _random()
+                        prod *= U
+                        if prod <= enlam:
+                            return X
+                        X += 1
+
+                ret = context.compile_internal(builder, poisson_impl, sig, args)
```
stuartarchibald, line 1603: Not for this PR, but in general, whilst this is a faithful port of the existing code, use of `compile_internal` can and should be avoided. The call should either be resolved via the context or ideally the overload should be entirely rewritten to instead have the overload implementation written as a JIT compilable Python implementation with intrinsic calls where needed.

stuartarchibald (commented):
@apmasell thanks for the patch and for undertaking this refactoring. In general this looks good and it's nice to see these as `@overload`s! I've made a few minor comments, the one about catching errors I think needs resolving ahead of this being merged.

@gmarkall Thinking about CUDA's tentative steps towards enabling `@overload` etc... I assume this won't impact the CUDA target as the RNG support via e.g. stdlib `random` module doesn't work on CUDA as it involves use of pointers from a DSO?

numba/cpython/randomimpl.py
apmasell: Yeah, random complex numbers are probably not a good idea. Fixed.

stuartarchibald:
> @gmarkall Thinking about CUDA's tentative steps towards enabling `@overload` etc... I assume this won't impact the CUDA target as the RNG support via e.g. stdlib `random` module doesn't work on CUDA as it involves use of pointers from a DSO?

This suggests there's no dependency:
https://github.com/numba/numba/blob/447fd9628c66ef18623fc5b65291a8845ed9841d/numba/cuda/tests/nocuda/test_import.py#L5-L25

stuartarchibald (approved):

esc:
Thank you for the patch!

PR #8371, opened 2022-08-19
Added binomial distribution

kc611:
As titled, adds `np.random.Generator().binomial()` distribution support.


numba/np/random/distributions.py
```diff
@@ -529,3 +529,165 @@ def random_poisson(bitgen, lam):
 def random_negative_binomial(bitgen, n, p):
     Y = random_gamma(bitgen, n, (1 - p) / p)
     return random_poisson(bitgen, Y)
+
+
+@register_jitable
+def random_binomial_btpe(bitgen, n, p):
```
kc611, line 535: This particular function (which is implemented using `goto` statements in C in NumPy source code) leads to massive slowdown during compilation, and potentially needs to be optimized somehow.

numba/np/random/distributions.py
```diff
@@ -529,3 +529,165 @@ def random_poisson(bitgen, lam):
...
+            U = next_double(bitgen)
+        else:
+            U -= px
+            px = ((n - X + 1) * p * px) / (X * q)
+
+    return X
+
+
+@register_jitable
+def random_binomial(bitgen, p, n):
```
bszollosinagy, line 679: A typo I guess : the order of the arguments "p" and "n" are reversed here in the fn signature, compared to NumPyRandomGeneratorType_binomial which calls random_binomial() like this: random_binomial(inst.bit_generator, n, p)

numba/np/random/distributions.py
```diff
@@ -529,3 +529,165 @@ def random_poisson(bitgen, lam):
...
+                      / x2) / x1 / 166320. +
+                     (13680. - (462. - (132. - (99. - 140. / w2) / w2) / w2)
+                      / w2) / w / 66320.)):
+                case = 0
+                continue
+        elif case == 6:
+            if (p > 0.5):
+                y = n - y
+
+    return y
```
bszollosinagy, line 652: a typo here: the "return y"  ought to be under "elif case == 6:" and not outside the "while 1:" loop, because like this it never actually returns.

numba/np/random/distributions.py
bszollosinagy: Hi @kc611 , the curiosity of seeing a bunch of goto was too interesting to skip. I took your approach, and then converted your code into a finite state machine using various approaches, but all of them take many minutes to compile. Your "while" and "if" based switch table is the winner for now.

bszollosinagy: I have managed to decrease the compile time, but only by about 10%, by re-arranging your code. See [my fork](https://github.com/bszollosinagy/numba/commits/refactor_binomial_goto)

I have traced the algo to its original paper, where they discuss how efficient it will be at executing, but did not consider compilation times:
Kachitvichyanukul, V., & Schmeiser, B. W. (1988). Binomial random variate generation. Communications of the ACM, 31(2), 216-222.

Now sure how to improve on the compilation time any more than this. AFAIK the finite state machine is about the only acceptable use of GOTO nowadays, because aligns well with the problem domain (literally jumping around states).

numba/np/random/distributions.py
kc611: This one too. The return outght to be in that case.

numba/np/random/distributions.py
kc611: Yes, this seems to be a typo.

numba/np/random/distributions.py
kc611: Hi, @bszollosinagy Thank you for for your interest in this PR. :smiley: 

> I took your approach, and then converted your code into a finite state machine using various approaches, but all of them take many minutes to compile. Your "while" and "if" based switch table is the winner for now.

Yep, same was the case for me. Of the all the logical approaches I took, this particular one seems to compile and execute the fastest, hence I drafted the PR with this one. 

The changes you made on your fork look pretty great. I think those might pretty much be the optimizations we could make with the 'while' + 'if' approach. Do you want to continue development on the issue ? If so, I'll leave this PR be for a while, so your fork can take over. I think the tests have already been written and the only work remaining would be making sure the types that Numba lowers the logic into are correct to prevent overflows.

kc611: No probs, I'll adopt changes from your fork,, they look good. 

The type verification was required in previous such PRs to avoid using the wrong floating precision such as accidentally using 32 bit float in a 64 bit algorithmic implementation, since without intervention, Numba will assign the default precision bits, same for signed - unsigned floats.

numba/np/random/distributions.py
```diff
@@ -574,3 +574,164 @@ def random_logseries(bitgen, p):
...
+            y = int(xm - p1 * v + u)
+            case = 6
+            continue
+        elif case == 1:
+            if (u > p2):
+                case = 2
+                continue
+            x = xl + (u - p1) / c
+            v = v * c + 1.0 - abs(m - x + 0.5) / p1
+            if (v > 1.0):
```
kc611, line 617: So far I've been able to figure out the failure as follows:

The logic is not going through this branch, i.e. The logic jumps from case 0 to case 1 and is supposed to go through here but doesn't and ends up returning incorrect values in the next iteration. 

Probably we're overlooking some sort of precision issue over here.

stuartarchibald: Is the issue in the expression `p1 = int(2.195 * np.sqrt(n * r * q) - 4.6 * q) + 0.5`, seems like in the NumPy source it's `p1 = floor(<same expr>)`, which has a different behaviour in the negative case.

kc611: I corrected this, but doesn't seem like the issue at hand is being fixed.

numba/np/random/distributions.py
```diff
@@ -574,3 +574,164 @@ def random_logseries(bitgen, p):
...
         else:
             return 2
+
+
+@register_jitable
+def random_binomial_btpe(bitgen, n, p):
+    r = min(p, 1.0 - p)
+    q = 1.0 - r
+    fm = n * r + r
+    m = np.floor(fm)
```
stuartarchibald, line 584: Think this should be cast to the Numba equivalent of `RAND_INT_TYPE`?

numba/np/random/distributions.py
```diff
@@ -574,3 +574,164 @@ def random_logseries(bitgen, p):
...
+    y = k = 0
+    while 1:
+        if case == 0:
+            nrq = n * r * q
+            u = next_double(bitgen) * p4
+            v = next_double(bitgen)
+            if (u > p1):
+                case = 1
+                continue
+            y = np.floor(xm - p1 * v + u)
```
stuartarchibald, line 608: `y` should be `int` type, need to cast to `RAND_INT_TYPE`?

numba/np/random/distributions.py
```diff
@@ -574,3 +574,164 @@ def random_logseries(bitgen, p):
...
+
+    case = 0
+    y = k = 0
+    while 1:
+        if case == 0:
+            nrq = n * r * q
+            u = next_double(bitgen) * p4
+            v = next_double(bitgen)
+            if (u > p1):
+                case = 1
```
stuartarchibald, line 606: Maybe update all these case names to reflect the labels in NumPy. For example, `case = 1` is a reflection of the statement `goto Step20`, so perhaps refer to this as `case = 20`? It might  make it easier to track changes/validate the alg.

numba/np/random/distributions.py
```diff
@@ -574,3 +574,164 @@ def random_logseries(bitgen, p):
...
+                continue
+            y = np.floor(xm - p1 * v + u)
+            case = 6
+            continue
+        elif case == 1:
+            if (u > p2):
+                case = 2
+                continue
+            x = xl + (u - p1) / c
+            v = v * c + 1.0 - abs(m - x + 0.5) / p1
```
stuartarchibald, line 616:
```suggestion
            v = v * c + 1.0 - np.fabs(m - x + 0.5) / p1
```

numba/np/random/distributions.py
```diff
@@ -574,3 +574,164 @@ def random_logseries(bitgen, p):
...
+            if (v > 1.0):
+                case = 0
+                continue
+            y = np.floor(x)
```
stuartarchibald, line 620: `y` should be `int`, cast needed?

numba/np/random/distributions.py
```diff
@@ -574,3 +574,164 @@ def random_logseries(bitgen, p):
...
+                continue
+            y = np.floor(xl + np.log(v) / laml)
+            if ((y < 0) or (v == 0.0)):
+                case = 0
+                continue
+            v = v * (u - p2) * laml
+            case = 4
+            continue
+        elif case == 3:
+            y = np.floor(xr - np.log(v) / lamr)
```
stuartarchibald, line 635: `y` should be `int`, cast needed?

numba/np/random/distributions.py
```diff
@@ -574,3 +574,164 @@ def random_logseries(bitgen, p):
...
+                case = 0
+                continue
+            y = np.floor(x)
+            case = 4
+            continue
+        elif case == 2:
+            if (u > p3):
+                case = 3
+                continue
+            y = np.floor(xl + np.log(v) / laml)
```
stuartarchibald, line 627: `y` should be `int`, cast needed?

numba/np/random/distributions.py
```diff
@@ -574,3 +574,164 @@ def random_logseries(bitgen, p):
...
+                case = 5
+                continue
+            s = r / q
+            a = s * (n + 1)
+            F = 1.0
+            if (m < y):
+                for i in range(m + 1, y + 1):
+                    F = F * (a / i - s)
+            elif (m > y):
+                for i in range(m + 1, y + 1):
```
stuartarchibald, line 654: I think this range expression is incorrect compared to NumPy (it's backwards). Consider the `if..else`, the range domain must reflect the nature of what is in the predicates.

stuartarchibald (commented):
Thanks for the update @kc611, I've found a few more potential areas of algorithmic difference that could be causing the issues you report, they are noted inline, hope this helps?!

numba/np/random/distributions.py
kc611: Resolved in [3dfb3ca](https://github.com/numba/numba/pull/8371/commits/3dfb3cad5a2629fe52672ad71534e215ee3063e1)

numba/np/random/distributions.py
kc611: Resolved in [3dfb3ca](https://github.com/numba/numba/pull/8371/commits/3dfb3cad5a2629fe52672ad71534e215ee3063e1)

numba/np/random/distributions.py
kc611: Resolved in [3dfb3ca](https://github.com/numba/numba/pull/8371/commits/3dfb3cad5a2629fe52672ad71534e215ee3063e1)

numba/np/random/distributions.py
kc611: Resolved in [3dfb3ca](https://github.com/numba/numba/pull/8371/commits/3dfb3cad5a2629fe52672ad71534e215ee3063e1)

numba/np/random/distributions.py
kc611: Resolved in [3dfb3ca](https://github.com/numba/numba/pull/8371/commits/3dfb3cad5a2629fe52672ad71534e215ee3063e1)

numba/np/random/distributions.py
kc611: Resolved in [3dfb3ca](https://github.com/numba/numba/pull/8371/commits/3dfb3cad5a2629fe52672ad71534e215ee3063e1)

numba/np/random/distributions.py
kc611: Resolved in [3dfb3ca](https://github.com/numba/numba/pull/8371/commits/3dfb3cad5a2629fe52672ad71534e215ee3063e1)

kc611:
Yep, it seems to be working now. The looping was the issue I suppose.

esc:
This was over 1k commits behind `main` -- so I merged it in. It was basically no longer possible to build this PR with the environment configuration for `main` -- because the dependency on the `tbb` and `tbb-devel` packages had changed.

esc:
@kc611 I am not sure this is worth a towncrier entry, what do you think?

kc611:
Yep, we can skip release notes for this one.

esc:
I ran this through the build farm as `numba_smoketest_cpu_nprgen_39` -- a NumPy RNG specific test -- and it came back all green.

PR #8375, opened 2022-08-19
[WIP] Extend cache invalidation to include changes in other files

luk-f-a:
This PR extends the cache invalidation mechanism to detect changes in functions imported from other files and used in the main function. It does that by including the code and closure variables of any dispatcher called in the main function in the index key used in the cache.

This PR is still a draft.

Implementation notes:
 - The process is recursive. Each function hashes its own code and closures, and that of its direct dependencies.
 - The merging of the code and closure is not done at the content level and then hashed. Since the process is recursive, the content of one functions is concatenated with the **hashed** content of its dependencies and then hashed. Intuitively this is ok, but it should be confirmed by someone with better understanding of hash functions.
 - since the index_key is now a property of the dispatcher object, it's necessary to pass it to the cache object.

Open questions:
- what to do with ever growing cache size?
- should invalidation of cache index be based on code signature, not on file timestamp?

TODO:
- clean tests

### DIFF
--- a/numba/core/caching.py
+++ b/numba/core/caching.py
@@ -592,14 +592,12 @@ class Cache(_Cache):
     # The following class variables must be overridden by subclass.
     _impl_class = None
 
-    def __init__(self, py_func,
-                 index_info_callback: pt.Callable[[], pt.Tuple[str, str]]):
+    def __init__(self, py_func, index_key: pt.Callable[[], pt.Tuple[str, str]]):
         # self._dispatcher_index_key is function that returns
         # a tuple of hashed code and hashed content of closure variables
         # it's a callable instead of a tuple to delay the calculation of the
         # index key until all globals have been linked to the function
-        # If None is provided, the index_key will be based only
-        self._dispatcher_index_info = index_info_callback
+        self._dispatcher_index_key = index_key
         self._name = repr(py_func)
         self._py_func = py_func
         self._impl = self._impl_class(py_func)
@@ -688,8 +686,7 @@ class Cache(_Cache):
         """
         # self._dispatcher_index_key is a function that returns a tuple of
         # hashed code and hashed content of closure variables
-        index_key = self._dispatcher_index_info(sig).index_key
-        return (sig, codegen.magic_tuple(), index_key)
+        return (sig, codegen.magic_tuple(), self._dispatcher_index_key())
 
 
 class FunctionCache(Cache):
--- a/numba/core/caching.py
+++ b/numba/core/caching.py
@@ -15,6 +15,7 @@ import sys
 import tempfile
 import uuid
 import warnings
+import typing as pt
 
 from numba.misc.appdirs import AppDirs
 
@@ -595,7 +596,10 @@ class Cache(_Cache):
     # The following class variables must be overridden by subclass.
     _impl_class = None
 
-    def __init__(self, py_func):
+    def __init__(self, py_func, index_key: pt.Tuple[str, str]):
+        # self._dispatcher_index_key is a tuple of hashed code and
+        # hashed content of closure variables
+        self._dispatcher_index_key = index_key
         self._name = repr(py_func)
         self._py_func = py_func
         self._impl = self._impl_class(py_func)
@@ -682,18 +686,9 @@ class Cache(_Cache):
         the bytecode for the function and, if the function has a __closure__,
         a hash of the cell_contents.
         """
-        codebytes = self._py_func.__code__.co_code
-        if self._py_func.__closure__ is not None:
-            cvars = tuple([x.cell_contents for x in self._py_func.__closure__])
-            # Note: cloudpickle serializes a function differently depending
-            #       on how the process is launched; e.g. multiprocessing.Process
-            cvarbytes = dumps(cvars)
-        else:
-            cvarbytes = b''
-
-        hasher = lambda x: hashlib.sha256(x).hexdigest()
-        return (sig, codegen.magic_tuple(), (hasher(codebytes),
-                                             hasher(cvarbytes),))
+        # self._dispatcher_index_key is a tuple of hashed code and
+        # hashed content of closure variables
+        return (sig, codegen.magic_tuple(), self._dispatcher_index_key)
 
 
 class FunctionCache(Cache):
--- a/numba/core/caching.py
+++ b/numba/core/caching.py
@@ -596,9 +596,11 @@ class Cache(_Cache):
     # The following class variables must be overridden by subclass.
     _impl_class = None
 
-    def __init__(self, py_func, index_key: pt.Tuple[str, str]):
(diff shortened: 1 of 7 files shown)


stuartarchibald:
RE:
> should invalidation of cache index be based on code signature, not on file timestamp?

Would this lead to undesirable behaviour in cases like this:
```python
from numba import njit

glbl = 10

@njit(cache=True)
def foo():
  return glbl
```
with `foo` used from another module. If `glbl` is changed, without a timestamp `foo`'s cache isn't invalidated?

stuartarchibald:
@luk-f-a as discussed at the public development meeting on 2022-08-23, this might help with finding the IR and typemap without actually lowering anything. There's probably a better way to do this but it's a start.

```python
from numba import njit, types
from numba.core.compiler import CompilerBase, DefaultPassBuilder
from numba.core.typed_passes import NopythonTypeInference
from numba.core.compiler_machinery import AnalysisPass, register_pass
from numba.core import ir
from pprint import pprint


def get_type_map(function, arg_tys):

    save = dict()

    # Define a compiler pass to simply store the IR and typemap
    @register_pass(mutates_CFG=False, analysis_only=True)
    class SaveTypeMapAndIR(AnalysisPass):
        _name = "save_type_map_and_ir"

        def __init__(self):
            AnalysisPass.__init__(self)

        def run_pass(self, state):
            save['func_ir'] = state.func_ir
            save['typemap'] = state.typemap
            return False

    # Compiler pass to swap the calls to njit functions to a new dispatcher
    # with the TypeInferCompiler pipeline
    @register_pass(mutates_CFG=False, analysis_only=True)
    class SwitchPipelineClass(AnalysisPass):
        _name = "switchpipelineclass"

        def __init__(self):
            AnalysisPass.__init__(self)

        def run_pass(self, state):
            mutated = False
            func_ir = state.func_ir
            targets = set()
            # find the call targets
            for blk in func_ir.blocks.values():
                calls = [x for x in blk.find_exprs('call')]
                for call in calls:
                    glbl = func_ir.get_definition(calls[0].func.name)
                    target = func_ir.get_assignee(glbl)
                    targets.add(target)
            # replace the globals of the call targets with custom njit pipeline
            for blk in func_ir.blocks.values():
                for stmt in blk.body:
                    if isinstance(stmt, ir.Assign):
                        if stmt.target in targets:
                            glbl = stmt.value
                            pyfunc = glbl.value.py_func
                            new = njit(pipeline_class=TypeInferCompiler)(pyfunc)
                            glbl.value = new
                    mutated = True
            return mutated

    # create a compiler pipeline that runs the untyped part of the pipeline,
    # switch out the dispatchers for custom ones using this pipeline, then run
    # type inference, and then the custom passes above to store the IR/typemap
    class TypeInferCompiler(CompilerBase):

        def define_pipelines(self):
            pm = DefaultPassBuilder.define_untyped_pipeline(self.state)
            pm.add_pass(SwitchPipelineClass, "PipelineSwitcher")
            pm.add_pass(NopythonTypeInference, "nopython frontend")
            pm.add_pass_after(SaveTypeMapAndIR, NopythonTypeInference)
            pm.finalize()
            return [pm]

    # run the supplied function with the given args using the custom pipeline
    jitted = njit(pipeline_class=TypeInferCompiler)(function)
    try:
        jitted.compile(arg_tys)
    except AttributeError as e:
        # Compilation will always fail as no compile result is being produced
        assert str(e) == 'cr' # compile result attr missing

    return save


@njit
def bar(z):
    return z * 1j

# Want to get the IR + typemap of this, assuming the arg is an int8
def foo(x):
    tmp = bar(x + 1)
    return tmp


arg_tys = (types.int8,)
saved = get_type_map(foo, arg_tys)


saved['func_ir'].dump()
pprint(saved['typemap'])
```

luk-f-a:
@stuartarchibald  thanks a lot for the code example. I'm testing a new approach in #8396 but I'll keep this PR for experimenting with this type inference without lowering. I couldn't run your example (on latest main), I get an error
```python
  File "/home/.../.config/.../scratches/numba_cache_invalidation2.py", line 93, in <module>
    saved = get_type_map(foo, arg_tys)
  File "/home/.../.config/.../scratches/numba_cache_invalidation2.py", line 74, in get_type_map
    jitted.compile(arg_tys)
  File "/home/.../mypyprojects/numba/numba/core/dispatcher.py", line 982, in compile
    cres = self._compiler.compile(args, return_type)
  File "/home/.../mypyprojects/numba/numba/core/dispatcher.py", line 135, in compile
    raise retval
  File "/home/.../mypyprojects/numba/numba/core/dispatcher.py", line 145, in _compile_cached
    retval = self._compile_core(args, return_type)
  File "/home/.../mypyprojects/numba/numba/core/dispatcher.py", line 158, in _compile_core
    cres = compiler.compile_extra(self.targetdescr.typing_context,
  File "/home/.../mypyprojects/numba/numba/core/compiler.py", line 716, in compile_extra
  File "/home/.../mypyprojects/numba/numba/core/compiler.py", line 452, in compile_extra
  File "/home/.../mypyprojects/numba/numba/core/compiler.py", line 520, in _compile_bytecode
    return self._compile_core()
  File "/home/.../mypyprojects/numba/numba/core/compiler.py", line 499, in _compile_core
    raise e
  File "/home/.../mypyprojects/numba/numba/core/compiler.py", line 486, in _compile_core
    pm.run(self.state)
  File "/home/.../mypyprojects/numba/numba/core/compiler_machinery.py", line 368, in run
    raise patched_exception
  File "/home/.../mypyprojects/numba/numba/core/compiler_machinery.py", line 356, in run
  File "/home/.../mypyprojects/numba/numba/core/compiler_lock.py", line 35, in _acquire_compile_lock
    return func(*args, **kwargs)
  File "/home/.../mypyprojects/numba/numba/core/compiler_machinery.py", line 311, in _runPass
  File "/home/.../mypyprojects/numba/numba/core/compiler_machinery.py", line 273, in check
  File "/home/.../mypyprojects/numba/numba/core/typed_passes.py", line 105, in run_pass
    typemap, return_type, calltypes, errs = type_inference_stage(
  File "/home/.../mypyprojects/numba/numba/core/typed_passes.py", line 83, in type_inference_stage
    errs = infer.propagate(raise_errors=raise_errors)
  File "/home/.../mypyprojects/numba/numba/core/typeinfer.py", line 1086, in propagate
    raise errors[0]
numba.core.errors.TypingError: Failed in untyped mode pipeline (step: nopython frontend)
Internal error at <numba.core.typeinfer.CallConstraint object at 0x7f75bcd9d9f0>.
cr
During: resolving callee type: type(CPUDispatcher(<function bar at 0x7f75be635d80>))
During: typing of call at /home/.../.config/.../scratches/numba_cache_invalidation2.py (88)

File "numba_cache_invalidation2.py", line 88:
def foo(x):
    tmp = bar(x + 1)
```

PR #8376, opened 2022-08-21
Fix issue 8370

### LINKED ISSUE #8370: NumbaTypeError: Cannot modify readonly array of type: readonly array(int64, 2d, C)
```
import numpy as np
from numba import njit

@njit
def unfold(x, y):
    r, c = x.shape
    a = np.broadcast_to(x, (y, r, c))
#     a = a.copy()
    b = np.swapaxes(a, 0, 1)
#     b = b.copy()
    cc = b.flatten()
#     c = c.copy()
    d = np.reshape(cc, (-1, c))
    # assign indicator to the right end of the period
    # in case of look forward bias
#     d = d.copy()
    d[y - 1 :, :] = d[: 1 - y]
    return d


unfold(np.array([[1, 1, 1], [2, 2, 2], [3,3,3]]), 2)
```
This will reproduce the error above. Under plain numpy this works fine.
By uncommenting any one of the copy() calls, the issue is resolved. So I have no idea which numpy call returns this readonly array.
Can it be implemented without an explicit copy call?

bszollosinagy:
Fixes #8370 

New unit tests were created, and the solution added, as suggested by @gmarkall 

All unit tests pass locally that would exercise ravel() and flatten()

- python -m numba.runtests -m -v numba.tests.test_array_reductions.TestArrayReductions
- python -m numba.runtests -m -v numba.tests.test_array_manipulation.TestArrayManipulation
- python -m numba.runtests -m -v numba.tests.test_closure.TestObjmodeFallback
- python -m numba.runtests -m -v numba.tests.test_dyn_array.TestNpStack
- python -m numba.runtests -m -v numba.tests.test_np_functions.TestNPFunctions.test_extract_basic
- python -m numba.runtests -m -v numba.tests.test_support.TestAssertPreciseEqual

CUDA tests may need to be run, but I simply cannot download the CUDA drivers from Nvidia today for some reason, so maybe a CI test would be more appropriate.


gmarkall:
Many thanks for the PR!

> CUDA tests may need to be run, but I simply cannot download the CUDA drivers from Nvidia today for some reason, so maybe a CI test would be more appropriate.

I'm not sure `flatten` would work on CUDA since its implementation returns a copy in Numba, and it's not possible to do allocation for the CUDA target - am I missing something here?

gmarkall:
gpuci run tests

gmarkall:
(Marking as RFR because the test fail is a CI issue, not an issue with this PR)

numba/core/typing/arraydecl.py
```diff
@@ -474,14 +474,14 @@ def resolve_ravel(self, ary, args, kws):
         # Only support no argument version (default order='C')
         assert not kws
         assert not args
-        return signature(ary.copy(ndim=1, layout='C'))
+        return signature(ary.copy(ndim=1, layout='C', readonly=False))
```
stuartarchibald, line 477: I'm not sure that this is correct? I think the mutability of the array resulting from `ravel()` would be determined by whether a copy is needed?
```pycon
In [35]: import numpy as np

In [36]: a = np.ones((2, 3))

In [37]: a.flags.writeable = False

In [38]: a.flags
Out[38]: 
  C_CONTIGUOUS : True
  F_CONTIGUOUS : False
  OWNDATA : True
  WRITEABLE : False
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False

In [39]: a_c = np.ravel(a)

In [40]: a_c.flags
Out[40]: 
  C_CONTIGUOUS : True
  F_CONTIGUOUS : True
  OWNDATA : False
  WRITEABLE : False
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False

In [41]: a_f = np.ravel(a, "F")

In [42]: a_f.flags
Out[42]: 
  C_CONTIGUOUS : True
  F_CONTIGUOUS : True
  OWNDATA : True
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False

In [43]: a.ravel('F').flags
Out[43]: 
  C_CONTIGUOUS : True
  F_CONTIGUOUS : True
  OWNDATA : True
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False

In [44]: a.ravel('C').flags
Out[44]: 
  C_CONTIGUOUS : True
  F_CONTIGUOUS : True
  OWNDATA : False
  WRITEABLE : False
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False
```

gmarkall: Note that in Numba, it appears that no copy is made if the array is C-contiguous, and a copy is made if it is not C-contiguous: https://github.com/numba/numba/blob/main/numba/np/arrayobj.py#L1992

bszollosinagy: Okay, updated the code to be sensitive to this, and added a unit test to verify.

bszollosinagy:
Resolved issue picked up by an automated test: if the source array is mutable, then a simple ravel() should not make it read-only.

bszollosinagy:
Ready for review.

bszollosinagy:
Note: the algo modification is not a big deal: 

- on calling flatten(), a copy is made by Numba, and therefore the read-only is removed (just like in Numpy, see ref 1).  
- In the case of ravel(), the copy is only made if the "order" of the dimensions has to be converted from Fortran to C, otherwise no copy is made. 
  - In case there was a copy, then the read-only is removed (just like in Numpy, see ref 1),
  - in case no copy was needed, then we look at the source array: was it read-only.

That's all.

ref: [Numpy removing writeable=False when doing a copy()](https://numpy.org/devdocs/reference/generated/numpy.copy.html)

gmarkall:
gpuci run tests

numba/tests/test_array_manipulation.py
```diff
@@ -1250,6 +1250,69 @@ def test_broadcast_arrays_incompatible_shapes_raise_valueerror(self):
...
+            cc = b.flatten()
+            d = np.reshape(cc, (-1, c))
+            d[y - 1:, :] = d[: 1 - y]
+            return d
+
+        pyfunc = unfold_flatten
+        cfunc = jit(nopython=True)(pyfunc)
+
+        # If issue #8370 is not fixed: This will fail with "NumbaTypeError:
+        # Cannot modify readonly array of type: readonly array(int32, 2d, C)"
```
gmarkall, line 1268: This isn't the message I get - I get:

```
 
 >>> setitem(readonly array(int64, 2d, C), UniTuple(slice<a:b> x 2), readonly array(int64, 2d, C))
```

I think it'd be OK to just delete this comment rather than update it.

numba/tests/test_array_manipulation.py
```diff
@@ -1250,6 +1250,69 @@ def test_broadcast_arrays_incompatible_shapes_raise_valueerror(self):
...
+            cc = b.ravel()
+            d = np.reshape(cc, (-1, c))
+            d[y - 1:, :] = d[: 1 - y]
+            return d
+
+        pyfunc = unfold_ravel
+        cfunc = jit(nopython=True)(pyfunc)
+
+        # If issue #8370 is not fixed: This will fail with "NumbaTypeError:
+        # Cannot modify readonly array of type: readonly array(int32, 2d, C)"
```
gmarkall, line 1287: Similar to above, I think this comment can / should be removed.

numba/core/typing/arraydecl.py
```diff
@@ -474,14 +474,19 @@ def resolve_ravel(self, ary, args, kws):
...
+        readonly = not (copy_will_be_made or ary.mutable)
+        return signature(ary.copy(ndim=1, layout='C', readonly=readonly))
 
     @bound_function("array.flatten")
     def resolve_flatten(self, ary, args, kws):
         # Only support no argument version (default order='C')
         assert not kws
         assert not args
-        return signature(ary.copy(ndim=1, layout='C'))
+        # Just like in Numpy: "np.copy clears previously
```
gmarkall, line 486:
```suggestion
        # Just like in NumPy: "np.copy clears previously
```

numba/core/typing/arraydecl.py
```diff
@@ -474,14 +474,19 @@ def resolve_ravel(self, ary, args, kws):
...
+        # set WRITEABLE=False flag"
+        # Numba also sets readonly=False (b/c flatten returns a copy())
```
gmarkall, line 488: I'm unclear about what this comment is saying. Is there any deviation from NumPy here, or do Numba and NumPy have similar behaviour now?

gmarkall (changes requested):
Many thanks for the updates and your continued efforts! There are a couple of comments on the PR.

In addition, in `test_readonly_after_ravel` and `test_readonly_after_flatten`, there is no assertion that checks that an expected output is produced - could you also assert that `cfunc` and `pyfunc` both produce the same return value please?

gmarkall:
Also, apologies for the delay in this round of review.

numba/core/typing/arraydecl.py
bszollosinagy: Numba and NumPy will have identical behaviour. The comment explains why the next line has a hardcoded readonly=False  argument.

bszollosinagy: I've now rephrased the comment.

numba/tests/test_array_manipulation.py
```diff
@@ -1315,6 +1315,74 @@ def test_broadcast_arrays_incompatible_shapes_raise_valueerror(self):
...
+            return d
+
+        pyfunc = unfold_flatten
+        cfunc = jit(nopython=True)(pyfunc)
+
+        # If issue #8370 is not fixed: This will fail.
+        res_nb = cfunc(np.array([[1, 1, 1], [2, 2, 2], [3, 3, 3]]), 2)
+        res_py = pyfunc(np.array([[1, 1, 1], [2, 2, 2], [3, 3, 3]]), 2)
+
+        self.assertTrue(np.array_equal(res_py, res_nb),
```
gmarkall, line 1336: If you use `np.testing.assert_array_equal(res_py, res_nb)` here instead of `self.assertTrue(np.array_equal(...`, then if the test fails, NumPy outputs the difference.

E.g. for a trivial example:

```
In [2]: np.testing.assert_array_equal(1, 2)

# ... snip ...

AssertionError: 
Arrays are not equal

Mismatched elements: 1 / 1 (100%)
Max absolute difference: 1
Max relative difference: 0.5
 x: array(1)
 y: array(2)
```

gmarkall (changes requested):
Many thanks for the quick update - just a very small comment on the diff - would you be happy to make this change, then this should be good to approve?

numba/tests/test_array_manipulation.py
bszollosinagy: sounds good, done!

gmarkall:
Many thanks - just waiting on CI before I approve.

gmarkall:
gpuci run tests

gmarkall (approved):
Many thanks for the PR and updates! I think this looks good now.

PR #8377, opened 2022-08-22
Add support for NumPy 1.23.

stuartarchibald:
Adds support for NumPy 1.23 to Numba. With thanks to @gmarkall for patches for the CUDA target.

The Azure CI matrix now looks like this:
```
$ python -c 'from numba.tests.support import print_azure_matrix; print_azure_matrix()'
NumPy | Python | Count
-----------------------
 1.18 |  3.7   |   4
 1.18 |  3.8   |   1
 1.19.2=*_0 |  3.8   |   3
 1.19.2=*_0 |  3.9   |   1
 1.20 |  3.8   |   3
 1.20 |  3.9   |   1
 1.21 |  3.9   |   1
 1.21 |  3.10  |   1
 1.22 |  3.10  |   1
 1.22 |  3.8   |   1
 1.23 |  3.9   |   1
 1.23 |  3.10  |   2
```


gmarkall:
gpuci run tests

stuartarchibald:
Am marking this as ready for review as the 32bit build fail on Azure CI is unrelated.

numba/tests/test_array_methods.py
```diff
@@ -383,6 +383,17 @@ def check_err(arr, dtype):
                 run(arr, dtype)
             self.assertEqual(str(raises.exception),
                              "new type not compatible with array")
+        def check_err_noncontig_last_axis(arr, dtype):
+            # check NumPy interpreted version raises
```
gmarkall, line 387: On first glance this seems a tiny bit odd to be checking NumPy, but I think the purpose that this serves is to explicitly test / document our own understanding of the NumPy behaviour we replicate in Numba. So I'm happy with this check staying in.

Out of curiousity, do we expect further NumPy changes around this area that we might detect in future releases with the NumPy check?

gmarkall (commented):
I think the code is good in the sense that it gets the test suite passing with 1.23 in light of the changed behaviour on taking views, but I still think we permit a little more than NumPy does - I'm not sure whether this matters or not.

For example:

```python
import numpy as np
from numba import njit

x = np.zeros((4, 4))
view = x.T[:2:, :2:]


@njit
def f(arr):
    return arr.view(np.int8)

print("Running jitted function...")
f(view)
print("Running Python function...")
f.py_func(view)
```

gives:

```
$ python repro.py 
Running jitted function...
Running Python function...
  File "/home/gmarkall/numbadev/issues/8377/repro.py", line 15, in <module>
    f.py_func(view)
  File "/home/gmarkall/numbadev/issues/8377/repro.py", line 10, in f
    return arr.view(np.int8)
ValueError: To change to a dtype of a different size, the last axis must be contiguous
```

I think this discrepancy comes because we only look at C-/F-contiguity (`if flag_f and not flag_c:`) rather than looking specifically at the strides as NumPy does:

https://github.com/numpy/numpy/blob/50a74fb65fc752e77a2f9e9e2b7227629c2ba953/numpy/core/src/multiarray/getset.c#L516-L525

numba/tests/test_array_methods.py
stuartarchibald: I am unsure if there will be changes in NumPy in this area in future releases, but put this in catch NumPy either changing this or the associated error message(s).

stuartarchibald:
Thanks for finding the patch breaking example in https://github.com/numba/numba/pull/8377#pullrequestreview-1083461152 @gmarkall, I've added 8b0d0d2 in an attempt to address this. I noticed when writing it that the existing Numba implementation doesn't exactly replicate NumPy either with respect to error messages, but suggest leaving that for another time (an error is raised, it's just not the same as that in NumPy < 1.23).

gmarkall:
gpuci run tests

gmarkall (approved):
I think this is good to approve - it looks like behaviour is consistent for NumPy 1.23 now as far as I can tell and the code of the checks appears to be equivalent between Numba and NumPy. I did notice that there are still some discrepancies between the messages produced by Numba and those produced by NumPy, but they are only for NumPy < 1.23 so I don't think it's really worth worrying about.

I tried to push a commit that reduced some duplication in the `check_err_*` methods but it seems I can't push to Stuart's branch so may just add this as a small refactoring PR later.

sklam (approved):

Issue #8304, opened 2022-08-02
Python 3.11


esc:
This is the meta/coordination issue for Python 3.11 support for Numba and llvmlite.

Note: The issue text will be updated continuously to reflect the current process and progress. 

As Numba interfaces with cpython on the bytecode and C-level, Python minor version upgrades present a significant compatibility challenge and have -- in the past -- required multiple PRs over multiple months of work. For Python 3.11 this issue will serve as a central point of coordination of the efforts. As such it will list the current branches pertaining to Python 3.11 support in addition to using the tag `Python 3.11`.

In order to support testing against beta releases and release candidates, I created a small utility that allows easy setup of a Numba development environment against a Python pre-release docker container. https://github.com/numba/numba-hatchery this can be used with development branches/repositories of Numba and llvmlite.

* llvmlite status
  * Using the PR https://github.com/numba/llvmlite/pull/869 passes all unit tests when used with numba-hatchery

* numba status
  * Numba Python 3.11 support is now being worked on across multiple branches:
    * https://github.com/esc/numba/tree/python3.11_partial/version_bump  -- version bump
    * https://github.com/esc/numba/tree/python3.11_partial/compile-hack -- compile hacks (WIP)
    * https://github.com/esc/numba/tree/python3.11_partial/implement_resume -- RESUME implementation
  * Numba would not compile initially. The https://github.com/esc/numba/tree/python3.11_partial/compile-hack branch  was created to allow. Numba to compile
  * The new `RESUME` opcode wasn't being handled. https://github.com/esc/numba/tree/python3.11_partial/implement_resume  implements this.
  * The branch `esc/python3.11` now contains a merge of all involved feature branches and can be used for testing all required patches in unison: https://github.com/numba/numba/pull/8381
  * @sklam is making progress with the bytecode at: https://github.com/sklam/numba/tree/py3.11wip

* TODO
  * Cleanup the compile-hacks PR so that it is sane and not just something to allow compilation
  * Port distutils to use setuptools as per https://peps.python.org/pep-0632/

sklam:
I'm making progress on the bytecode support at https://github.com/sklam/numba/tree/py3.11wip

sklam:
I'm making progress at https://github.com/numba/numba/pull/8545. The PR adapts Numba for the major bytecode changes in Py3.11.

binary-husky:
Still no official support for python 3.11 ?

alonme:
Any news?

extesy:
[...]

PR #8378, opened 2022-08-22
bump max python version for Python 3.11

esc:
As title


esc:

PR #8379, opened 2022-08-22
Comment out all C-level changes

esc:
This will make Numba compile against the Python 3.11 code-base and
C-level API. Note that some of the Numba functionality may not be
available as a result.


esc:

PR #8380, opened 2022-08-22
implement new Python 3.11 opcode: RESUME

esc:
With Python 3.11 a new Bytecode has been introduced: `RESUME`.

The documentation states that:

```
RESUME has been added. It is a no-op. Performs internal tracing, debugging and optimization checks.
```

And so, this is implemented as a `NOP` (no-operation) in Numba.


esc:

PR #8384, opened 2022-08-23
Move strace() check into tests that actually need it.

stuartarchibald:
Having the strace() check on the import path for
`numba.tests.support` leads to the check occurring at every run of
a test in a subprocess. This moves the check for `strace()`
support to the point of use.


gmarkall (approved):
I no longer see the strace message when running the CUDA test suite (which used to happen multiple times) with this patch. This is a nice "quality of life" improvement in that regard!

PR #8385, opened 2022-08-23
DO NOT MERGE. debugging ci timeout

stuartarchibald:


stuartarchibald:
Copy of #8241 which is failing on the 32bit tests, but with the `-v` flag added at test time to try and identify the test(s) that is/are stuck!

PR #8386, opened 2022-08-23
Fix the docs for numba.get_thread_id

stuartarchibald:
* Adds the function to the API reference.
* Adds associated links/reference wiring.
* Formats the associated docs.


sklam (approved):
Thanks for the patch!

Issue #7954, opened 2022-04-01
Provide a way for lowering to know what Compute Capability is being targeted


gmarkall:
As discussed in https://github.com/numba/numba/pull/7556#pullrequestreview-928732504, it's not possible for lowering to generate specific code based on the compute capability of the target device, because there's no way to know what it is.

An example use of this would be to generate inline assembly for CC 8.0 devices for max/min instructions on 16-bit floating point operands - presently we have to instead lower to a generic implementation that works for all compute capabilities.

One way to do this would be to add it to the `CUDAFlags`:

https://github.com/numba/numba/blob/637a75023bae889467ebfc0e8489be547b62a209/numba/cuda/compiler.py#L25-L30

but then we'd also need to provide a route for lowering to be able to tell what the flags are in the current compiler pipeline.

gmarkall:
From discussion in the triage meeting - flags can be obtained using the `ConfigStack`, e.g. as in `_compile_subroutine_no_cache()` in `base.py`.

dlee992:
@gmarkall, I met the similar problem when coding SDC. Pandas have a flexible way to handle `series["A"]`:
```
import pandas as pd
s1 = pd.Series([1,2,3,4], index=list("ABCD"))
print(s1["A"])
1
s2 = pd.Series([1,2,3,4], index=list("AAAD"))
print(s2["A"])
A    1
A    2
A    3
dtype: int64
```
But when I implement this behavior on SDC, I cannot decide to return which type without runtime info. So I want to let users to tell compiler whether return series or not.

dlee992:
@sklam @stuartarchibald , hi, can we open a "back gate" to allow developers to register an extra jit option in Numba? As above, I still want to compile the function in CPU target, just want to define and use an extra jit option in type infer pass. I don't want to override the whole Compiler class just for this simple purpose.
If I misunderstand sth, please correct me.

stuartarchibald:
@dlee992 Hi! Please could you open a new issue using the "feature request" template asking about the above? In this new issue please do include a description of what you are asking for, ideally with motivating example/sample code. This issue (#7954) is about specialising lowering on CUDA compute capability and the above seems unrelated or maybe a generalisation relating to the use of flags, hence it's probably best discussed in a new issue? Many thanks for your help.

dlee992:
Yes, of course.

PR #8387, opened 2022-08-23
Add support for compute capability in IR Lowering

testhound:
This PR addresses issue 7954: https://github.com/numba/numba/issues/7954 by adding compute capability to the compiler flags which can be checked during lowering. It also adds a test case where tanh is converted to tanh.approx.f32 when fastmath is enabled and compute capability >= 7.5.

One consequence of this change is that several test case had to be converted to CUDATestCase because of the driver call in compile_cuda to get the compute capability.


testhound:
gpuci run tests

gmarkall (changes requested):
The fact that the tests of `compile_ptx` needed to be converted to `CUDATestCase` subclasses  suggests an issue - compilation to PTX shouldn't require a CUDA device to be present / cause any interaction with the driver API.

Before this can proceed, it needs to be changed such that `compile_ptx` doesn't cause any interactions with the driver / hardware.

testhound:
gpuci run tests

testhound:
@gmarkall fixed so that `compile_ptx` does not interact with the driver. I did have to convert `TestCudaMandel `to derive from `CUDATestCase` because set device=true in the jit decorator.

testhound:
gpuci run tests

testhound:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall (changes requested):
Many thanks for the PR! I have made a small suggestion on the diff about the flag type (I also think the `nvvm_options_type` which probably inspired it should go, but that's something for another PR). I have a couple of other concerns:

The addition of use of the default compute capability in `compile_cuda()` now means the default is potentially used in two places (there and in `compile_ptx()`), which is a bit harder to reason about the safety of. I think it'd be safer if the compute capability always had to be passed in to `compile_cuda()`, then the default only need be used for `compile_ptx()`. I think the following change would achieve this:

```diff
diff --git a/numba/cuda/compiler.py b/numba/cuda/compiler.py
index 664d82c8c..99c5b7e2d 100644
--- a/numba/cuda/compiler.py
+++ b/numba/cuda/compiler.py
@@ -191,6 +191,9 @@ class CUDACompiler(CompilerBase):
 def compile_cuda(pyfunc, return_type, args, debug=False, lineinfo=False,
                  inline=False, fastmath=False, nvvm_options=None,
                  cc=None):
+    if cc is None:
+        raise ValueError('Compute Capability must be supplied')
+
     from .descriptor import cuda_target
     typingctx = cuda_target.typing_context
     targetctx = cuda_target.target_context
@@ -215,7 +218,7 @@ def compile_cuda(pyfunc, return_type, args, debug=False, lineinfo=False,
         flags.fastmath = True
     if nvvm_options:
         flags.nvvm_options = nvvm_options
-    flags.compute_capability = cc or config.CUDA_DEFAULT_PTX_CC
+    flags.compute_capability = cc
 
     # Run compilation pipeline
     from numba.core.target_extension import target_override
@@ -277,6 +280,7 @@ def compile_ptx(pyfunc, args, debug=False, lineinfo=False, device=False,
         'opt': 3 if opt else 0
     }
 
+    cc = cc or config.CUDA_DEFAULT_PTX_CC
     cres = compile_cuda(pyfunc, None, args, debug=debug, lineinfo=lineinfo,
                         fastmath=fastmath,
                         nvvm_options=nvvm_options, cc=cc)
@@ -292,7 +296,6 @@ def compile_ptx(pyfunc, args, debug=False, lineinfo=False, device=False,
         lib, kernel = tgt.prepare_cuda_kernel(cres.library, cres.fndesc, debug,
                                               nvvm_options, filename, linenum)
 
-    cc = cc or config.CUDA_DEFAULT_PTX_CC
     ptx = lib.get_asm_str(cc=cc)
     return ptx, resty
 
```
What do you think of this suggestion?

My second question is - what happens with the `tanh` implementation when used in a function compiled with `compile_ptx`? Can you add a test that compiles for two different CCs (one less than and one at least 7.5) please?

testhound:
@gmarkall I incorporated all your requested changes.

testhound:
gpuci run tests

gmarkall:
> My second question is - what happens with the tanh implementation when used in a function compiled with compile_ptx? Can you add a test that compiles for two different CCs (one less than and one at least 7.5) please?

Can you add the tests of `compile_ptx` please?

testhound:
> > My second question is - what happens with the tanh implementation when used in a function compiled with compile_ptx? Can you add a test that compiles for two different CCs (one less than and one at least 7.5) please?
> 
> Can you add the tests of `compile_ptx` please?

@gmarkall I added a new test "test_tanhf_no_fastmath" that is skipped if compute capability is < 7.5.

numba/cuda/mathimpl.py
```diff
@@ -249,13 +248,53 @@ def lower_ldexp_impl(context, builder, sig, args):
...
 
+def impl_tanh(ty, libfunc):
+    def lower_tanh_impl(context, builder, sig, args):
+        def get_compute_capability():
+            cstk = targetconfig.ConfigStack()
+            if cstk:
+                flags = cstk.top()
+                return flags.compute_capability
+            else:
+                return None
```
gmarkall, line 259: From OOB discussion, I think the conclusion is that the top of the config stack can never be `None` (an `IndexError` would be raised in this case).

```suggestion
            flags = targetconfig.ConfigStack().top()
            return flags.compute_capability
```

gmarkall (changes requested):
Following some further OOB discussion, it seems like the remaining step is to add a test of compiling `tanh` with and without fastmath using `compile_ptx()`, specifying compute capabilities 7.0 in one instance and 7.5 in another - this will mean that both paths through the tanh implementation get tested on all systems, regardless of the device present in the machine.

This should also ensure that there isn't a case we've missed where the config stack is empty when `compile_ptx()` is called - I don't think there is one, bit I think the config stack stuff is very hard to reason about due to its use of TLS and hidden / global state.

testhound:
gpuci run tests

testhound:
@gmarkall new testcase added as you requested.

testhound:
@gmarkall fixed merge conflicts and waiting on your review.

gmarkall (changes requested):
Many thanks for the updates - some comments on the diff, but on the whole I think this is almost done.

testhound:
@gmarkall all request changes made to the PR.

gmarkall:
gpuci run tests

gmarkall:
@testhound what about enabling `test_tanhf_compile_ptx` to run regardless of the CC of the installed device?

testhound:
@gmarkall okay I was planning to try that next, I misunderstood you comment and applied it to test_tanhf.

testhound:
gpuci run tests

testhound:
gpuci run tests

testhound:
@gmarkall the gpuci test failed in a way unrelated to my latest change.

gmarkall (approved):
I think this looks good now - many thanks for the updates!

PR #8389, opened 2022-08-23
Make C++ extensions compile with correct compiler

apmasell:
Some extension files are C++ but have `.c` extensions. GCC will compile these
as C++, but clang will not.

### DIFF
--- a/numba/_typeof.cpp
+++ b/numba/_typeof.cpp
@@ -101,9 +101,9 @@ string_writer_ensure(string_writer_t *w, size_t bytes)
     if (newsize < bytes)
         newsize = bytes;
     if (w->buf == w->static_buf)
-        w->buf = malloc(newsize);
+        w->buf = (char *) malloc(newsize);
     else
-        w->buf = realloc(w->buf, newsize);
+        w->buf = (char *) realloc(w->buf, newsize);
     if (w->buf) {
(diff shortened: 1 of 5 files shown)


numba/_typeof.cpp
```diff
@@ -842,6 +842,10 @@ int typecode_devicendarray(PyObject *dispatcher, PyObject *ary)
     int dtype;
     int ndim;
     int layout = 0;
+    PyObject *ndim_obj = nullptr;
+    PyObject* num_obj = nullptr;
+    PyObject* dtype_obj = nullptr;
```
stuartarchibald, line 847:
```suggestion
    PyObject *num_obj = nullptr;
    PyObject *dtype_obj = nullptr;
```

stuartarchibald (commented):
Thanks for the patch @apmasell. I've taken a look through the C/C++ code changes, specifically at things like type casts and the proposed changes seem valid on inspection.

gmarkall (approved):
This looks good to me - I'm approving it, though I note Stuart's suggestion about the formatting of the pointers on the diff.

apmasell:
Accepted the suggestion. Might be worth extending the formatting from llvmlite to numba.

stuartarchibald (approved):

Issue #8222, opened 2022-07-05
compilation error: expected expression when handling typeconv.cpp on mac


dlee992:
## Reporting a bug
- [x] I have tried using the latest released version of Numba (most recent is
 visible in the change log (https://github.com/numba/numba/blob/main/CHANGE_LOG).
- [x] I have included a self contained code sample to reproduce the problem.
  i.e. it's possible to run as 'python bug.py'.

## Code
I build numba main branch code on my x86_64 mac machine, but get a compilation error.

## Compilation errors
```
(numba-dev) ➜  numba (main✔)python setup.py build_ext --inplace
TBB not found
OpenMP not found
/Users/ld/opt/anaconda3/envs/numba-dev/lib/python3.8/site-packages/setuptools/dist.py:530: UserWarning: The version specified ('cuda-dispatcher-base+1148.g15e487886') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.
  warnings.warn(
running build_ext
building 'numba._dynfunc' extension
Warning: Can't read registry to find the necessary compiler setting
Make sure that Python modules winreg, win32api or win32con are installed.
C compiler: gcc -Wno-unused-result -Wsign-compare -Wunreachable-code -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -I/Users/ld/opt/anaconda3/envs/numba-dev/include -arch x86_64 -I/Users/ld/opt/anaconda3/envs/numba-dev/include -arch x86_64

creating build
creating build/temp.macosx-10.9-x86_64-3.8
creating build/temp.macosx-10.9-x86_64-3.8/numba
compile options: '-I/Users/ld/opt/anaconda3/envs/numba-dev/include/python3.8 -c'
gcc: numba/_dynfuncmod.c
In file included from numba/_dynfuncmod.c:1:
numba/_dynfunc.c:143:5: warning: 'tp_print' is deprecated [-Wdeprecated-declarations]
    0,                         /* tp_print */
    ^
/Users/ld/opt/anaconda3/envs/numba-dev/include/python3.8/cpython/object.h:260:5: note: 'tp_print' has been explicitly marked deprecated here
    Py_DEPRECATED(3.8) int (*tp_print)(PyObject *, FILE *, int);
    ^
/Users/ld/opt/anaconda3/envs/numba-dev/include/python3.8/pyport.h:515:54: note: expanded from macro 'Py_DEPRECATED'
[...]

PR #8409, opened 2022-08-31
fix compilation error on mac

dlee992:
Fixes https://github.com/numba/numba/issues/8222.

I tried on my MacOS, with clang 13.1.6 installed.

Just revert several changed line in https://github.com/numba/numba/pull/7364.

But I still to set envvar `NUMBA_DISABLE_OPENMP=1` to avoid another compilation error `clang: error: unsupported option '-fopenmp'`. I feel the new error is unrelated with this PR.

### DIFF
--- a/numba/core/typeconv/typeconv.cpp
+++ b/numba/core/typeconv/typeconv.cpp
@@ -27,7 +27,11 @@ void TCCMap::insert(const TypePair &key, TypeCompatibleCode val) {
             return;
         }
     }
-    bin.push_back({key, val});
+    TCCRecord data;
+    data.key = key;
+    data.val = val;
+    bin.push_back(data);
+//    bin.push_back({key, val});
     nb_records++;
 }
 
--- a/numba/core/typeconv/typeconv.cpp
+++ b/numba/core/typeconv/typeconv.cpp
@@ -31,7 +31,6 @@ void TCCMap::insert(const TypePair &key, TypeCompatibleCode val) {
     data.key = key;
     data.val = val;
     bin.push_back(data);
-//    bin.push_back({key, val});
     nb_records++;
 }
 


dlee992:
@gmarkall @stuartarchibald I guess u guys are familar with this part of code, so ping u, ah.

dlee992:
Yeah, I agree. Should use #8389 style to fix this. But that PR is only related with `extra_compile_args=['-std=c++11'],` with my issue. I tried but still not work. 

clang did have some behavior deviations compared with gcc, which need some cares to handle with. But I feel this general problem should be addressed by other projects already, maybe we can obtain the related experience.

guilhermeleobas:
I can reproduce the same error @dlee992 is having on osx.
```bash
INFO: clang -bundle -undefined dynamic_lookup -Wl,-rpath,/Users/guilhermeleobas/miniconda3/envs/numba/lib -L/Users/guilhermeleobas/miniconda3/envs/numba/lib -Wl,-rpath,/Users/guilhermeleobas/miniconda3/envs/numba/lib -L/Users/guilhermeleobas/miniconda3/envs/numba/lib build/temp.macosx-11.0-arm64-cpython-39/numba/experimental/jitclass/_box.o -o /Users/guilhermeleobas/git/numba/numba/experimental/jitclass/_box.cpython-39-darwin.so
numba/core/typeconv/typeconv.cpp:30:19: error: expected expression
    bin.push_back({key, val});
                  ^
numba/_dispatcher.cpp:1104:37: warning: offset of on non-POD type 'Dispatcher' [-Winvalid-offsetof]
    {(char*)"_can_compile", T_BOOL, offsetof(Dispatcher, can_compile), 0, NULL },
                                    ^                    ~~~~~~~~~~~
/Library/Developer/CommandLineTools/usr/lib/clang/13.1.6/include/stddef.h:104:24: note: expanded from macro 'offsetof'
#define offsetof(t, d) __builtin_offsetof(t, d)
                       ^                     ~
```

And I can confirm his patch works.

gmarkall:
@guilhermeleobas Does #8389 still give you the same error when building with it? Or do you get a different error?

guilhermeleobas:
@gmarkall No errors when building #8389. Just had to disable openmp when building it.

stuartarchibald:
Following #8389, is this suggested patch still needed?

dlee992:
Thanks for your patch. Closing this now.

Issue #4689, opened 2019-10-10
The fastmath SVML option has been broken since somewhere between 0.43.1 and 0.44


stuartarchibald:
## Reporting a bug

- [x] I am using the latest released version of Numba (most recent is visible in
 the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [x] I have included below a minimal working reproducer (if you are unsure how
 to write one see http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports).

Numba 0.43.1 was the last version of Numba with a working SVML option accessible via the standard JIT decorators. The patch 310b3cbb1 broke SVML working, but it's not especially the fault of that patch. The issue seems to be caused by decorating a function with `njit` ahead of setting the vector library option to LLVM via the llvmlite bindings.

This is broken:
```python
from llvmlite import binding

# import before set
from numba import njit
binding.set_option('SVML', '-vector-library=SVML')

import numpy as np
import math

@njit(fastmath={'fast'})
def foo(n):
    x   = np.empty(n*8, dtype=np.float64)
    ret = np.empty_like(x)
    for i in range(ret.size):
            ret[i] += math.cosh(x[i])
    return ret

foo(1000)

if 'intel_svmlcc' in foo.inspect_llvm(foo.signatures[0]):
    print("found")
    exit(0)
else:
    print("not found")
    exit(1)
```
whereas this works as expected:
```python
from llvmlite import binding

# set before import
binding.set_option('SVML', '-vector-library=SVML')
from numba import njit

import numpy as np
import math

@njit(fastmath={'fast'})
def foo(n):
    x   = np.empty(n*8, dtype=np.float64)
    ret = np.empty_like(x)
    for i in range(ret.size):
            ret[i] += math.cosh(x[i])
    return ret

foo(1000)

if 'intel_svmlcc' in foo.inspect_llvm(foo.signatures[0]):
    print("found")
    exit(0)
else:
    print("not found")
    exit(1)
```

The issue manifests in the Numba code base during init as there's now all this:
https://github.com/numba/numba/blob/bdd8e312a24c5b76ee4fe480abe553cff349aa51/numba/__init__.py#L46-L52
which are imports of modules that somewhere in their import tree import `njit` and then decorate one or more functions with it, this prior to setting the vector-library option in the llvmlite DSO to SVML.

This is most likely an issue with internal state of the llvmlite DSO. Fixing options:
[...]

PR #8410, opened 2022-08-31
Remove old warning about imports after SVML init

gmarkall:
This warning applied to imports that followed SVML initialization, but now there are none.

Related question - do we have a test that ensures that nothing is `@njit`-ed during Numba import?

### DIFF
--- a/numba/__init__.py
+++ b/numba/__init__.py
@@ -211,12 +211,3 @@ import llvmlite
 Is set to True if Intel SVML is in use.
 """
 config.USING_SVML = _try_enable_svml()
-
-
-# ---------------------- WARNING WARNING WARNING ----------------------------
-# The following imports occur below here (SVML init) because somewhere in their
-# import sequence they have a `@njit` wrapped function. This triggers too early
-# a bind to the underlying LLVM libraries which then irretrievably sets the LLVM
-# SVML state to "no SVML". See https://github.com/numba/numba/issues/4689 for
-# context.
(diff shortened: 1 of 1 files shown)


stuartarchibald:
Thanks for the patch.

> This warning applied to imports that followed SVML initialization, but now there are none.

I'm happy to remove this warning, conditional on the following test existing, which I do not think it does:

> Related question - do we have a test that ensures that nothing is `@njit`-ed during Numba import?

Tracking down the issue that resulted in that warning being written was exceedingly hard (think it was issue #4689) so I think a test is required. Maybe the code samples demonstrating the issue in #4689 could be used as a base for this test?

stuartarchibald:
In #8424 I've attempted to make the initialization of LLVM lazy (probably still needs more work). The result of this patch is that the import of `numba` and the use of the `jit` decorator family of functions does not result in LLVM being initialized. As a side effect, nothing gets JIT compiled when a `jit` decorated function is interpreted, prior to this merely interpreting a `jit` decorated function lead to the NRT being compiled.

I think this means that the warning in this PR could indeed be removed. It also seems possible to test this by seeing if:
```python
from llvmlite import binding as b
print(b.Target.from_triple(b.get_process_triple()))
```
results in a `RuntimeError`, if it does, it means it's likely that LLVM was not initialized for CPU JIT purposes. It should therefore be possible to write some tests to run in subprocesses that do things like `import numba`, and exercise the `jit` decorators, and then check that the `RuntimeError` still occurs.

PR #8414, opened 2022-09-01
Hotfix for 0.56.2

sklam:
as titled

PR #8419, opened 2022-09-04
increment version support table entry for 0.56.* to 0.56.2

esc:
With https://github.com/numba/numba/pull/8414/files the version support table on `main` must also be updated.


Sohlstyce (approved):
Fixing an overlook, should be fine.

docs/source/user/installing.rst
```diff
@@ -278,7 +278,7 @@ information.
 +===========+==============+===========================+============================+==============================+===================+=============================+
 | 0.57.x    | TBC          | 3.8.x <= version < 3.12   | 1.19 <= version < 1.24     | 0.40.x                       | 11.x              | 2021.x                      |
 +-----------+--------------+---------------------------+----------------------------+------------------------------+-------------------+-----------------------------+
-| 0.56.1    | 2022-09-01   | 3.7.x <= version < 3.11   | 1.18 <= version < 1.24     | 0.39.x                       | 11.x              | 2021.x                      |
+| 0.56.2    | 2022-09-01   | 3.7.x <= version < 3.11   | 1.18 <= version < 1.24     | 0.39.x                       | 11.x              | 2021.x                      |
```
stuartarchibald, line 281: Is it worth including a line about 0.56.1 with a note about it not being shipped?

esc: maybe, yes, it does appear like that in the changelog

stuartarchibald: Whilst perhaps not strictly necessary, given this table is the reference point for package maintainers to work out what works/is tested, I'm inclined to say include it, on the basis it'll save them having to go looking further to figure out where 0.56.1 went. Thanks!

stuartarchibald (dismissed):
Thanks for the patch.

gmarkall:
Marked as abandoned following triage discussion - new PR to follow

PR #8415, opened 2022-09-02
Refactor `numba.np.arraymath` methods from lower_builtins to overloads

kc611:
This PR refactors methods in `numba.np.arraymath` to use the new `@overload` API. It refactors the following functions:
- [x] `np.prod`
- [x] `np.cumsum`
- [x] `np.cumprod`
- [x] `np.mean`
- [x] `np.var`
- [x] `np.std`
- [x] `np.min`
- [x] `np.max`

The two remaining functions in `arraymath` that use `@lower_builtin` are:
- `np.nonzero` (which would be trivial once #8258 goes in) 
- `np.sum` (which is big enough to need it's own separate PR, due to it's support for `axis` argument). 

Ultimately the goal with these PRs would be to get `axis` support for all the methods i.e. extending `axis` framework used in `np.sum` to all the other methods in `arraymath`.

PR #4480, opened 2019-08-23
[WIP] Reimplementing np.mean with overload and adding axis parameter

luk-f-a:
Partially addresses #1269 and #4245 
Requires #4472 (first 3 commits come from that PR), #5060 and #5711, which fix intermediate problems.

PR #8420, opened 2022-09-05
Refactor np.sum to use @overload decorator and support for axis for various functions

kc611:
Builds on top of #8415

This PR refactors the array method `np.sum` to use the new `@overload` API. It also refactors the function such that it supports  axis values > 3 for the method. 

Additionally, This framework is then extended onto the following functions refactored in #8415:

- np.prod
- np.cumsum
- np.cumprod
- np.mean
- np.var
- np.std
- np.min
- np.max

### DIFF
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -14,18 +14,19 @@ import llvmlite.ir
 import numpy as np
 
 from numba.core import types, cgutils
-from numba.core.extending import overload, overload_method, register_jitable
+from numba.core.extending import (overload, overload_method,
+                                  register_jitable, intrinsic)
 from numba.np.numpy_support import (as_dtype, type_can_asarray, type_is_scalar,
                                     numpy_version, is_nonelike,
                                     check_is_integer)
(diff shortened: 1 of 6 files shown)


numba/np/arraymath.py
```diff
@@ -44,388 +44,121 @@ def _check_blas():
...
-    zero = sig.return_type(0)
 
+@overload(np.sum)
+@overload_method(types.Array, "sum")
+def array_sum(arr, axis=None):
+
+    @register_jitable
     def array_sum_impl(arr):
-        c = zero
+        c = 0
```
stuartarchibald, line 57: I think it's necessary to ensure that the type of the accumulator matches the dtype of the array. IIRC `np.arange(100, dtype=np.float32).sum()` uses a float32 accumulator and returns a value of `np.float32` type.

luk-f-a:
if you managed to solve the problem with `npydecl.registry` that would make `test_parfor.test_var` fail (describe here https://github.com/numba/numba/pull/4480#issuecomment-629698098), maybe this PR unlocks #4480

PR #8421, opened 2022-09-06
Update release checklist: add a task to check dependency pinnings on subsequent releases (e.g. PATCH)

esc:
As title


stuartarchibald (approved):
Thanks for the patch.

PR #8422, opened 2022-09-06
Switch public CI builds to use gdb from conda packages.

stuartarchibald:
As title.


gmarkall (approved):
This looks good to me.

esc:
Thank you for the patch!

PR #8424, opened 2022-09-07
Make LLVM initialization lazy in the CPU target.

stuartarchibald:
This patch pushes the LLVM initialization and all compilation off the import path and the `jit` decorator path for the CPU target.

The result of this patch is that both LLVM and the NRT are not initialized until Numba is definitely going to be required to compile something.

### DIFF
--- a/numba/core/cpu.py
+++ b/numba/core/cpu.py
@@ -39,14 +39,27 @@ class CPUContext(BaseContext):
     def __init__(self, typingctx, target='cpu'):
         super().__init__(typingctx, target)
 
+    @property
+    def _internal_codegen(self):
+        # Only need the lock whilst this isn't set up
+        if self._internal_codegen_ref is None:
+            with global_compiler_lock:
+                # make sure another thread hasn't set the state whilst the lock
+                # was blocking.
+                if self._internal_codegen_ref is None:
+                    self._internal_codegen_ref = \
+                        codegen.JITCPUCodegen("numba.exec")
+        return self._internal_codegen_ref
+
     # Overrides
     def create_module(self, name):
         return self._internal_codegen._create_empty_module(name)
 
     @global_compiler_lock
     def init(self):
+        self._nrt_initialized = False
         self.is32bit = (utils.MACHINE_BITS == 32)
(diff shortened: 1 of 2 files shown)


stuartarchibald:
One failing test as of 4677ddd:
```
ERROR: test_liftcall5 (numba.tests.test_withlifting.TestLiftCall)
  File "/home/vsts/work/1/s/numba/tests/test_withlifting.py", line 291, in test_liftcall5
    self.check_extracted_with(liftcall5, expect_count=1,
  File "/home/vsts/work/1/s/numba/tests/test_withlifting.py", line 211, in check_extracted_with
    cres.entry_point()
RuntimeError: missing Environment: _ZN08NumbaEnv5numba5tests16test_withlifting9liftcall5B4v216B32c8tJTC_2fWQHKmLSjegf4uKgGzNAE_3dEx
```
Notes:
1. Replacing the nested context to just use the standard `cpu_target.<appropriate context>` here https://github.com/numba/numba/blob/447fd9628c66ef18623fc5b65291a8845ed9841d/numba/tests/test_withlifting.py#L197-L198 means the nested context ends up being not actually nested as the contexts are the same: https://github.com/numba/numba/blob/447fd9628c66ef18623fc5b65291a8845ed9841d/numba/tests/test_withlifting.py#L215-L222
   which seems to fix it.
2. This test runs by itself successfully, it's just when run with another one of the `test_liftcall{<5}` tests it fails, e.g.
   ```
   ./runtests.py numba.tests.test_withlifting.TestLiftCall.test_liftcall1 numba.tests.test_withlifting.TestLiftCall.test_liftcall5 -v 
   ```

sklam:
RE: https://github.com/numba/numba/pull/8424#issuecomment-1239322747

I think it has to do with https://github.com/numba/numba/blob/c61921276a0ae27d74c82126070014607d6bf460/numba/core/base.py#L350. It's creating the codegen object before it's actually needing one.

I'm guessing the environment is being inserted into the wrong codegen.

stuartarchibald:
The lazy init of the NRT is extracted out into #8438. Dealing with the lazy init of LLVM itself is going to be difficult as found above! Closing this for now.

PR #8423, opened 2022-09-06
Remove public facing and CI references to 32 bit linux support.

stuartarchibald:
As noted in the release notes for 0.56, it is the last release to support `linux-32` (i.e. linux running on 32 bit x86 hardware). This patch series:
* Removes 32 bit linux support from the build scripts/conda recipes.
* Removes 32 bit linux builds from Azure CI.
* Fixes up the documentation to not refer to `linux-32` as supported.
* Removes references to other 32 bit linux systems which are no longer tested/supported, i.e. `armv6l` and `armv7l`.

This patch series does not address changes needed to the code base itself that are necessary to remove `linux-32` support.

This PR depends on merging #8422, that patch is included as the first in the series for this PR as it touches the same code.

PR #8426, opened 2022-09-07
Try using a single conda solve in Azure CI.

stuartarchibald:
Depends on merge of #8423.


stuartarchibald:
Closing as it doesn't seem to make much material difference at present. This patch should be easy enough to recreate if needed.

PR #8427, opened 2022-09-07
Shorten the time to verify test discovery.

stuartarchibald:
The Numba project uses `python -m numba.tests.test_runtests` in build scripts to verify that test discovery and the test suite itself is working as expected. This is a guard to make sure that when the test suite is run, it's actually running what is expected.

This patch tries to shorten the time it takes to get to the point of the test suite being validated. It does two things:

1. Hoists a costly loop invariant from a loop in one of the tests in the test_runtests module.
2. Caches the mtime of test files based on class as this should be invariant (gains about 3-4% improvement on runtests.py -l).

### DIFF
--- a/numba/testing/main.py
+++ b/numba/testing/main.py
@@ -11,6 +11,7 @@ import time
 import unittest
 import warnings
 
+from functools import cache
 from io import StringIO
 from unittest import result, runner, signals, suite, loader, case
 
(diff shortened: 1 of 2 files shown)


apmasell (approved):

esc (approved):
I did a crude benchmark of this branch with `main`.

For `main` I get the following times: `70.326s` and `69.016s` and `64.027s`.

For the feature branch I get: `62.353s` and `54.295s` and `55.061s`.

This is sufficient evidence for me regarding the effect this change has, thank you for the patch!

Issue #8254, opened 2022-07-18
meta-issue: migrate `@glue_*` functions to `@overload`


guilhermeleobas:
Following some recent discussion, this is a meta-issue to track the migration from the `@glue_typing/glue_lowering` API to use `@overload`.

### `numba/core/typing/npydecl.py`
- [x] np.array - [PR 8234](https://github.com/numba/numba/pull/8234)
- [x] np.full - [PR 8234](https://github.com/numba/numba/pull/8234)
- [x] np.full_like - [PR 8234](https://github.com/numba/numba/pull/8234)
- [x] np.identity - [PR 8234](https://github.com/numba/numba/pull/8234)
- [x] np.frombuffer - [PR 8234](https://github.com/numba/numba/pull/8234)
- [x] np.sort - [PR 8234](https://github.com/numba/numba/pull/8234)
- [x] np.asfortranarray - [PR 8234](https://github.com/numba/numba/pull/8234)
- [x] np.ascontiguousarray - [PR 8234](https://github.com/numba/numba/pull/8234)
- [x] np.copy - [PR 8234](https://github.com/numba/numba/pull/8234)
- [x] np.expand_dims - [PR 8234](https://github.com/numba/numba/pull/8234)
- [x] np.concatenate - [PR 8234](https://github.com/numba/numba/pull/8234)
- [x] np.stack - [PR 8234](https://github.com/numba/numba/pull/8234)
- [x] np.hstack - [PR 8234](https://github.com/numba/numba/pull/8234)
- [x] np.vstack - [PR 8234](https://github.com/numba/numba/pull/8234)
- [x] np.dstack - [PR 8234](https://github.com/numba/numba/pull/8234)
- [x] np.column_stack - [PR 8234](https://github.com/numba/numba/pull/8234)
- [x] np.round - [PR 8234](https://github.com/numba/numba/pull/8234)
- [x] np.around - [PR 8234](https://github.com/numba/numba/pull/8234)
- [x] np.sinc - [PR 8234](https://github.com/numba/numba/pull/8234)
- [x] np.angle - [PR 8234](https://github.com/numba/numba/pull/8234)
- [x] np.diag - [PR 8234](https://github.com/numba/numba/pull/8234)
- [x] np.linspace - [PR 8431](https://github.com/numba/numba/pull/8431)
- [x] np.atleast_1d - [PR 8435](https://github.com/numba/numba/pull/8435)
[...]

PR #8431, opened 2022-09-09
Replace `@overload_glue` by `@overload` for `np.linspace` and `np.take`

guilhermeleobas:
ref: #8254 

Change `np.take` and `np.linspace` to use `@overload`

### DIFF
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -4389,66 +4389,51 @@ def numpy_diag_kwarg(context, builder, sig, args):
     return impl_ret_new_ref(context, builder, sig.return_type, res)
 
 
-@lower_builtin('array.take', types.Array, types.Integer)
-@glue_lowering(np.take, types.Array, types.Integer)
-def numpy_take_1(context, builder, sig, args):
+@overload(np.take)
+@overload_method(types.Array, 'take')
+def numpy_take(a, indices):
 
-    def take_impl(a, indices):
-        if indices > (a.size - 1) or indices < -a.size:
-            raise IndexError("Index out of bounds")
-        return a.ravel()[indices]
-
-    res = context.compile_internal(builder, take_impl, sig, args)
-    return impl_ret_new_ref(context, builder, sig.return_type, res)
-
-
-@lower_builtin('array.take', types.Array, types.Array)
-@glue_lowering(np.take, types.Array, types.Array)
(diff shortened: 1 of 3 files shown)


numba/np/arrayobj.py
```diff
@@ -4545,21 +4531,28 @@ def impl(start, stop=None, step=None, dtype=None):
...
 
 
-@glue_lowering(np.linspace, types.Number, types.Number)
-def numpy_linspace_2(context, builder, sig, args):
+@overload(np.linspace)
+def numpy_linspace_2(start, stop):
+    if isinstance(start, types.Number) and isinstance(stop, types.Number):
+        def impl(start, stop):
+            return np.linspace(start, stop, 50)
+        return impl
```
stuartarchibald, line 4539: I don't think that this version should exist. The default value for `num` should be set as a kwarg in the 3 arg variant, like `start, stop, num=50`.

numba/np/arrayobj.py
```diff
@@ -4545,21 +4531,28 @@ def impl(start, stop=None, step=None, dtype=None):
...
+        return
 
-@glue_lowering(np.linspace, types.Number, types.Number, types.Integer)
-def numpy_linspace_3(context, builder, sig, args):
-    dtype = as_dtype(sig.return_type.dtype)
+    if any(isinstance(arg, types.Complex) for arg in [start, stop]):
+        dtype = types.complex128
+    else:
+        dtype = types.float64
+    dtype = as_dtype(dtype)
```
stuartarchibald, line 4554:
```suggestion
    if any(isinstance(arg, types.Complex) for arg in [start, stop]):
        dtype = np.complex128
    else:
        dtype = np.float64
```
Just set the dtype directly?

numba/np/arrayobj.py
```diff
@@ -4545,21 +4531,28 @@ def impl(start, stop=None, step=None, dtype=None):
...
 
-    # Implementation based on https://github.com/numpy/numpy/blob/v1.20.0/numpy/core/function_base.py#L24 # noqa: E501
```
stuartarchibald, line 4562: This reference should be maintained?

guilhermeleobas: Done!

stuartarchibald (approved):
Thanks for the patch!

PR #8438, opened 2022-09-14
Make the initialisation of the NRT more lazy for the njit decorator.

stuartarchibald:
This patch makes it so that the NRT is only compiled when compilation starts as the result of the user electing to trigger compilation. A new test is added to ensure that decorating a function with `njit` doesn't trigger NRT compilation.

### DIFF
--- a/numba/core/cpu.py
+++ b/numba/core/cpu.py
@@ -48,7 +48,6 @@ class CPUContext(BaseContext):
 
     @global_compiler_lock
     def init(self):
-        self._nrt_initialized = False
         self.is32bit = (utils.MACHINE_BITS == 32)
         self._internal_codegen = codegen.JITCPUCodegen("numba.exec")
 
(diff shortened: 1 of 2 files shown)


stuartarchibald:
As this hasn't been reviewed yet, have force pushed 8804f8d to 645c84c to rebase on `main` and resolve conflicts.

numba/core/cpu.py
```diff
@@ -58,10 +59,16 @@ def init(self):
...
-        # Initialize NRT runtime
-        rtsys.initialize(self)
-
     def load_additional_registries(self):
+        # Only initialize the NRT once something is about to be compiled. The
+        # "initialized" state doesn't need to be threadsafe, there's a lock
+        # around the internal compilation and the rtsys.initialize call can be
+        # made multiple times, worse case this just gets called a bit more often
+        # than optimal.
+        if not self._nrt_initialized:
```
sklam, line 68: `_nrt_initialized` is not needed. `rtsys.initialize` already checks for that.

sklam (commented):
Just one minor suggestion. The patch works as expected.

numba/core/cpu.py
stuartarchibald: Good point, fixed in 1a07d27.

sklam (approved):

PR #8439, opened 2022-09-14
Update the contributing docs to include a policy on formatting changes.

stuartarchibald:
This updates the contributing docs to highlight that Numba's policy on large scale formatting changes is to not accept such patches. It removes the fix a file to be flake8 compliant suggestion and adds a link to the time boxed tasks available on the self-service board.

### DIFF
--- a/docs/source/developer/contributing.rst
+++ b/docs/source/developer/contributing.rst
@@ -252,6 +252,25 @@ A code review should try to assess the following criteria:
 * docstrings, comments
 * test coverage
 
+
+Policy on large scale changes to code formatting
+''''''''''''''''''''''''''''''''''''''''''''''''
+
+Please note that pull requests making large scale changes to format the code
+base are in general not accepted. Such changes often increase the likelihood of
+merge conflicts for other pull requests, which inevitably take time and
+resources to resolve. They also require a lot of effort to check as Numba aims
+to compile code that is valid even if it is not ideal, i.e. testing
+``operator.eq``::
+
+    if x == None: # should be `if x is None:` but it's valid so don't change it!
+
+This policy has been adopted by the core developers so as to try and make best
+use of limited resources. Whilst it would be great to have an extremely tidy
+code base, priority is given to fixes and features over code formatting changes.
+Thanks for your understanding on this matter!
+
+
 Coding conventions
 ''''''''''''''''''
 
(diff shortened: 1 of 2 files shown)


docs/source/developer/contributing.rst
```diff
@@ -252,6 +252,25 @@ A code review should try to assess the following criteria:
...
+''''''''''''''''''''''''''''''''''''''''''''''''
+
+Please note that pull requests making large scale changes to format the code
+base are in general not accepted. Such changes often increase the likelihood of
+merge conflicts for other pull requests, which inevitably take time and
+resources to resolve. They also require a lot of effort to check as Numba aims
+to compile code that is valid even if it is not ideal, i.e. testing
+``operator.eq``::
+
+    if x == None: # should be `if x is None:` but it's valid so don't change it!
```
gmarkall, line 266: I think it's a little hard to understand this explanation as the explanation of the context is quite terse - perhaps the following would explain more clearly to a broader set of people:

```suggestion
to compile code that is valid even if it is not ideal. For example, in a test of
``operator.eq``::

    if x == None: # Valid code, even if the recommended form is `if x is None:`

This tests Numba's compilation of comparison with ``None``, and therefore
should not be changed, even though most style checkers will suggest it should.
```

What do you think?

docs/source/developer/contributing.rst
```diff
@@ -252,6 +252,25 @@ A code review should try to assess the following criteria:
...
+
+This policy has been adopted by the core developers so as to try and make best
+use of limited resources. Whilst it would be great to have an extremely tidy
+code base, priority is given to fixes and features over code formatting changes.
+Thanks for your understanding on this matter!
```
gmarkall, line 271: I think this comes from a place of genuine gratitude, but I'm concerned it reads a bit passive-aggressively. I also think the explanation up to this point was a clear and calm explanation of the situation, so perhaps it is unnecessary?

```suggestion
```

gmarkall (commented):
I've left a couple of comments on the diff - what do you think?

docs/source/developer/contributing.rst
stuartarchibald: Thanks for raising this, in the interests of ensuring that it cannot be interpreted with a meaning/in a manner that is unintended, I'll remove it.

docs/source/developer/contributing.rst
stuartarchibald: I think this makes it clearer, thanks for suggesting it.

stuartarchibald:
> I've left a couple of comments on the diff - what do you think?

Thanks for the review, both changes are accepted in b024011.

gmarkall (approved):
Thanks for the updated policy - it looks good to me. I've checked that this renders well on RTD also.

PR #8442, opened 2022-09-16
Implement hasattr(), str() and repr().

stuartarchibald:
This patch:

* Implements the `hasattr()` builtin.
* Reimplements the `str()` builtin to behave more like Python in that `__str__` and `repr` are used internally.
* Implements `repr` through the use of `__repr__`.
* Fixes up `hash` to check the result of a `__hash__` method before returning and adds some more hash implementations.
* Refactors a common "create a dummy type" method into the standard Numba `numba.tests.support.TestCase`.
* Updates the `LiteralPropagation` compilation pass to recognised inferred constants from `hasattr` calls.


numba/cpython/builtins.py
```diff
@@ -928,3 +929,76 @@ def impl(obj, name, default):
...
+    def impl(obj):
+        attr = '__repr__'
+        if hasattr(obj, attr) == True:
+            return getattr(obj, attr)()
+        else:
+            # There's no __str__ or __repr__ defined for this object, go via
+            # the interpreter.
+            with objmode(pyrepr='unicode_type'):
+                pyrepr = _get_py_repr_and_warn(obj)
+                return pyrepr
```
stuartarchibald, line 991: TODO: Need to decide about this! Another option is to raise.

numba/cpython/builtins.py
```diff
@@ -928,3 +929,77 @@ def impl(obj, name, default):
...
+# hasattr cannot be implemented as a getattr call and then catching
+# AttributeError because Numba doesn't support catching anything other than
+# "Exception", so lacks the specificity required. Instead this implementation
+# tries to resolve the attribute via typing information and returns True/False
+# based on that.
+@overload(hasattr)
+def ol_hasattr(obj, name):
+    def impl(obj, name):
+        return resolve_hasattr(obj, name)
+    return impl
```
guilhermeleobas, line 963: I think the implementation for `hasattr` is correct but it might produce false negatives in some cases, due to how Numba overloads some builtin functions. i.e. 

```python
@njit
def test():
    s = set([1, 2, 3])
    return hasattr(s, '__len__'), len(s)

print(test())
```

guilhermeleobas (dismissed):
LGMT. I only have one concern regarding `hasattr`. Although the impl. for it seems correct. It might produce false negatives in due to how Numba overloads some functions:

```python
@njit
def test():
    s = set([1, 2, 3])
    return hasattr(s, '__len__'), len(s)

print(test())
```

Should Numba emit a warning for users to use this feature with caution? Or even an acceptlist of functions we know are safe to use (str/repr/hash). Other functions should be safe once one migrates `@overload(func)` to `@overload_method(type, func)`

stuartarchibald:
Thanks for the review @guilhermeleobas. Two remarks:

1. Yes, `len` needs to defer to `obj.__len__`. I think that's it though for protocol based function support (ignoring `operator.*`)? Can you think of any others in the Numba code base at present?
2. Your comment makes me wonder if there ought to be a concept of "protected" overloads to warn users about overloading an implementation at an unusual level, i.e. warn on overloading `len` that overload `obj.__len__` is the preferred behaviour, that sort of thing.

guilhermeleobas:
I searched for `__` on https://docs.python.org/3/library/functions.html and these are the functions I found that has a corresponding dunder method:
* `abs` -> `__abs__()`
* `bin(x)` -> `__index__()`, when `x` is not an int
* `callable` -> `__call__()`
* `complex` -> `__complex__()` or `__float__()` or `__index__()`
* `enumerate` -> `__next__()`
* `float` -> `__float__()` or `__index__()`
* `hex` -> `__index__`()
* `int` -> `__int__()` or `__index__()` or `__trunc__()`
* `iter` -> Must support either the iterable or the sequence protocol
* `next` -> `__next__()`
* `reversed` -> `__reversed__()`
* `round` -> `__round__()` for python object

stuartarchibald:
> I searched for `__` on https://docs.python.org/3/library/functions.html and these are the functions I found that has a corresponding dunder method:

Thanks @guilhermeleobas, from memory, I think that Numba has implementations for a reduced set of these, namely:

>     * `abs` -> `__abs__()`
>     * `complex` -> `__complex__()` or `__float__()` or `__index__()`
>     * `enumerate` -> `__next__()`
>     * `float` -> `__float__()` or `__index__()`
>     * `int` -> `__int__()` or `__index__()` or `__trunc__()`
>     * `iter` -> Must support either the iterable or the sequence protocol
>     * `next` -> `__next__()`
>     * `round` -> `__round__()` for python object

I'm not sure if it's possible to handle `iter`, `enumerate` and `next` very easily due to the stack based nature of their implementations, the rest, however, should be doable.

stuartarchibald:
In the interests of making this feature available for Numba 0.57, what needs to be done to get this patch to a state where it's sufficiently useful to be able to merge it? Is it potentially "just" resolving these questions: https://github.com/numba/numba/pull/8442#issuecomment-1266770728 ?

I don't think that it's going to be possible to handle quite a few of those dunder methods above trivially in this PR. For example `abs` uses concrete templates which makes it hard to translate to an `overload`, the constructors like `float` and `int` are entangled with the NumPy machinery and others like `next` are stack based, which makes them reliant on direct lowering (though could go via the overload inliner eventually).

numba/cpython/builtins.py
sklam: one more option is to print something like `<object XYZType>` so it doesn't need objmode

numba/core/untyped_passes.py
```diff
@@ -1529,7 +1531,8 @@ def run_pass(self, state):
                     fn = guard(get_definition, func_ir, value.func.name)
                     if fn is None:
                         continue
-                    if not (isinstance(fn, ir.Global) and fn.name == 'isinstance'):  # noqa: E501
+                    if not (isinstance(fn, ir.Global) and fn.name in
+                            accepted_functions):  # noqa: E501
```
sklam, line 1535: The error msg that comes after should not be referring to hardcoded `"isinstance"` anymore

numba/cpython/hashing.py
```diff
@@ -47,8 +47,23 @@
...
+                           f"'{objty}' and a generic hash() cannot be "
+                           "performed as there is no suitable object "
+                           "represention in Numba compiled code!")
+
     def impl(obj):
-        return obj.__hash__()
+        if hasattr(obj, '__hash__'):
+            hash_func = getattr(obj, '__hash__')
+            hash_value = hash_func()
+            if hash_value is not None:
```
sklam, line 61: `__hash__` should never return non-integer according to python. Why not let it raise?

numba/np/arrayobj.py
```diff
@@ -3026,6 +3026,14 @@ def array_is_impl(a, b):
 
     return context.compile_internal(builder, array_is_impl, sig, args)
 
+# ------------------------------------------------------------------------------
+# Hash
+
+
+@overload_method(types.Array, "__hash__")
+def ol_array_hash(arr):
+    return lambda arr: None
```
sklam, line 3035: Why not raise?

numba/typed/dictobject.py
```diff
@@ -746,6 +746,13 @@ def impl(dct, key, default=None):
     return impl
 
 
+@overload_method(types.DictType, '__hash__')
+def impl_hash(dct):
+    if not isinstance(dct, types.DictType):
+        return
+    return lambda dct: None
```
sklam, line 753: why not raise?

numba/typed/listobject.py
```diff
@@ -1309,6 +1309,13 @@ def impl(lst, index):
     return impl
 
 
+@overload_method(types.ListType, '__hash__')
+def ol_list_hash(lst):
+    if not isinstance(lst, types.ListType):
+        return
+    return lambda lst: None
```
sklam, line 1316: why not raise?

sklam (commented):
Besides the minor suggestion to the error message in the `untyped_passes.py`, the main question is why allow `__hash__` to return `None`  instead of `raising`.

Once false negatives problems and the above suggestions are addressed, i think this patch is safe to be included.

numba/np/arrayobj.py
stuartarchibald: The intention is to replicate the behaviour in Python, the `__hash__` attribute of unhashable types is often `None`, e.g. `np.ndarray`:
```pycon
In [1]: import numpy as np

In [2]: x = np.ones(3)

In [3]: print(x.__hash__)
None
```

numba/typed/listobject.py
stuartarchibald: Comment as per `np.ndarray`:
```pycon
In [4]: print([].__hash__)
None
```

numba/typed/dictobject.py
stuartarchibald: Comment as per `np.ndarray`:
```pycon
In [5]: print({}.__hash__)
None
```

numba/cpython/hashing.py
stuartarchibald: Indeed, thanks for picking this up, I think it may be just working around a bug in `getattr`. Calls like `getattr(obj, method)` return in a `BoundFunction` against the `obj`, this probably needs to go through another round of resolution to provide the actual type/function. I'll take a look.

stuartarchibald: I think this was caused by the implementations using `@overload_method(<type>, '__hash__')` where they should have used `@overload_attribute(<type>, '__hash__')`.

numba/core/untyped_passes.py
stuartarchibald: Fix and test in 1c5970d.

stuartarchibald:
Thanks for the review @sklam, comments should now be addressed.

stuartarchibald:
1c5970d failed in Azure CI on the `macos_py37_np118` build with:
```
FAIL: test_reflect_exception (numba.tests.test_sets.TestSetReflection)
When the function exits with an exception, sets should still be
  File "/Users/runner/work/1/s/numba/tests/test_sets.py", line 813, in test_reflect_exception
    self.assertPreciseEqual(s, set([1, 2, 3, 42]))
  File "/Users/runner/miniconda3/envs/azure_ci/lib/python3.7/contextlib.py", line 119, in __exit__
    next(self.gen)
  File "/Users/runner/work/1/s/numba/tests/support.py", line 266, in assertRefCount
    % (old, new, obj))
AssertionError: Refcount changed from 4 to 6 for object: {1, 2, 3, 42}
```

stuartarchibald:
> This doesn't reproduce locally. Rerunning the CI job to see if it's transient.

It's not transient.

stuartarchibald:
9ef54ce fixes the transient issue.

numba/cpython/builtins.py
stuartarchibald: From Numba public meeting discussion on 2023-01-10, the use of `objmode` is going to be replaced with a simple compiled in string cf. https://github.com/numba/numba/pull/8442#discussion_r1029956563. This is because users are potentially going to `str()` or `repr()` objects to essentially stringify them for use elsewhere and reacquiring the GIL in such a case would be an unnecessary performance penalty.

stuartarchibald:
Need to do: https://github.com/numba/numba/pull/8442#discussion_r1066097995

numba/cpython/builtins.py
stuartarchibald: This is implemented and tested in b819f01.

sklam (approved):

PR #6881, opened 2021-03-31
support the 'reversed' function for ranges, lists, and arrays

farahhariri:
closes #2774

PR #8448, opened 2022-09-20
support the 'reversed' function

guilhermeleobas:
This PR continues the work done in https://github.com/numba/numba/pull/6881. I address the comments made by the reviewer and add new tests.

The commit history is preserved.

### DIFF
--- a/numba/cpython/builtins.py
+++ b/numba/cpython/builtins.py
@@ -700,7 +700,7 @@ def ol_sum(iterable, start=0):
 
 
 # ------------------------------------------------------------------------------
-# map, filter, reduce
+# map, filter, reduce, reversed
 
 
 @overload(map)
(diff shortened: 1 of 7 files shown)


stuartarchibald:
Thanks for continuing this @guilhermeleobas. Coincidentally, as a result of #8442, I also wrote `reversed()` yesterday, it's here: https://github.com/stuartarchibald/numba/commit/0c02483bad1df3aa9bca555a2c455046c5de004f. The difference in approach being that the implementation in the aforementioned patch follows the Python protocol of checking for `__reversed__` on an object and calling that ahead of performing a generic sequence protocol based reversal. This behaviour was enabled by implementing `hasattr` and making it cooperate with branch pruning in #8442.

guilhermeleobas:
Closing in favor of https://github.com/stuartarchibald/numba/commit/0c02483bad1df3aa9bca555a2c455046c5de004f

PR #8449, opened 2022-09-20
Try using libmamba experimental solver option in conda commands.

stuartarchibald:
As title, with view of speeding up CI env setup.

Depends on #8423.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -3,7 +3,7 @@ trigger:
 
 variables:
   # Change the following along with adding new TEST_START_INDEX.
-  TEST_COUNT: 20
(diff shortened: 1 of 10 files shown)


stuartarchibald:
Doesn't seem like this has a huge impact on the time it takes to get the build into a position to test changes. Depending on the CPU, somewhere in the 90-120s region is what happens on `main`, this patch has similar.

stuartarchibald:
Closing as the impact is not large enough to be a noticeable improvement.

esc:
> Closing as the impact is not large enough to be a noticeable improvement.

Good to know, thank you for checking.

Issue #8437, opened 2022-09-14
Remove user name scraping from changelog generator


sklam:
Referring to https://github.com/numba/numba/pull/8429 and roughly this part https://github.com/numba/numba/pull/8429/files#diff-c07552e1b5f50e764b6bba45566846cd148aabdb103ba23bb82a993eeea38fcdR34-R37

To remove scraping of user profile information from Github and to only rely on user name in the git log.

esc:
closing as #8450 has been merged!

PR #8450, opened 2022-09-20
remove GitHub username from changelog generation script

esc:
This modifies the changelog such that the configured username isn't used. Only the user login is used.

If the login is `maxm` and the configured username is `Max Mustermann`, the changelog entry would have contained:

```
(`Max Mustermann  <https://github.com/maxm>`_)
```

And with this change the entry would now contain:

```
(`maxm  <https://github.com/maxm>`_)
```

### DIFF
--- a/maint/gitlog2changelog.py
+++ b/maint/gitlog2changelog.py
@@ -32,9 +32,7 @@ def get_pr(pr_number):
 
 
 def hyperlink_user(user_obj):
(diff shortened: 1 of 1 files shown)


esc:
Fixes #8437

guilhermeleobas (approved):
LGTM. Tried locally and it is working as expected

PR #8454, opened 2022-09-21
Add string comparison operators from the numpy.char module.

nmehran:
This patch adds support for the following string array operations:

- 'char.equal'
- 'char.not_equal'
- 'char.greater_equal'
- 'char.less_equal'
- 'char.greater'
- 'char.less'
- 'char.compare_chararrays'

Included is support for UTF-32 strings and ASCII bytes on contiguous arrays of 1-dimension and scalars, a light framework for future supported string array functions, tests, and updated documentation.

### DIFF
--- a/docs/source/reference/numpysupported.rst
+++ b/docs/source/reference/numpysupported.rst
@@ -411,6 +411,19 @@ The following reduction functions are supported:
 * :func:`numpy.quantile` (only the 2 first arguments, complex dtypes
   unsupported)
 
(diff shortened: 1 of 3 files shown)


nmehran:
In addition to the above, development of 'string information' methods has been completed, found here: https://numpy.org/doc/stable/reference/routines.char.html#string-information.  

Given this pull requests current state, I will kindly close this request and initiate one which includes new and current methods.

Issue #8455, opened 2022-09-21
pythonapi function for PyBytes_AsStringAndSize?


jpivarski:
It looks like [pythonapi.py](https://github.com/numba/numba/blob/main/numba/core/pythonapi.py) doesn't have a function for [PyBytes_AsStringAndSize](https://docs.python.org/3/c-api/bytes.html#c.PyBytes_AsStringAndSize) (or `PyBytes_AsString` or `PyBytes_Size`, for that matter).

We could implement it manually, following an example like

https://github.com/numba/numba/blob/60723cb5a31d0b6ef516af95602a463322991420/numba/core/pythonapi.py#L470-L475

But are there any plans to fill out this part of the `PythonAPI` class with these methods?

Thanks!

Cc: @ianna

apmasell:
No plans, but I know some of these were driven by requests, so I'll check if there's any reason we didn't.

ianna:
I'd be happy to create a PR for it. Thanks!

stuartarchibald:
Closing as implemented in #8462 with thanks to @ianna!

PR #8462, opened 2022-09-26
Add PyBytes_AsString and PyBytes_AsStringAndSize

ianna:
Resolve issue https://github.com/numba/numba/issues/8455:
add `bytes_as_string` and `bytes_as_string_and_size`


numba/core/pythonapi.py
```diff
@@ -1150,6 +1150,23 @@ def string_from_kind_and_data(self, kind, string, size):
...
+
+    def bytes_as_string_and_size(self, obj, p_buffer, p_length):
+        fnty = ir.FunctionType(
+            self.pyobj,
+            [self.pyobj, self.cstring.as_pointer(), self.py_ssize_t.as_pointer()],
+        )
+        fname = "PyBytes_AsStringAndSize"
+        fn = self._get_function(fnty, name=fname)
+        result = self.builder.call(fn, [obj, p_buffer, p_length])
+        ok = self.builder.icmp_unsigned("!=", Constant(result.type, None), result)
```
guilhermeleobas, line 1167: In case of error, [`PyBytes_AsStringAndSize`](https://docs.python.org/3/c-api/bytes.html#c.PyBytes_AsStringAndSize) returns `-1`.

```suggestion
        ok = self.builder.icmp_unsigned("!=", Constant(result.type, -1), result)
```

guilhermeleobas (changes requested):
Hi @ianna, thanks for working on this issue. I only have one small comment regarding the return value of `PyBytes_AsStringAndSize`

numba/core/pythonapi.py
ianna: Sorry, I should have tested it before committing it - `-1` causes it to fail at runtime:
```python
>               raise RuntimeError("LLVM IR parsing error\n{0}".format(errmsg))
E               RuntimeError: Failed in nopython mode pipeline (step: native lowering)
E               LLVM IR parsing error
E               <string>:69:24: error: integer constant must have integer type
E                 %".42" = icmp ne i8* -1, %".41"
E                                      ^

```

ianna:
@guilhermeleobas - Where shall I add a unit test for this PR? Thanks!

guilhermeleobas:
Hi @ianna, I don't think Numba has any tests that covers the Python API. Perhaps you can create a file for that. Let me know if you need any help

numba/core/pythonapi.py
guilhermeleobas: My bad as well! The return type must be changed from `pyobj` to `py_ssize_t`. Also, I think the comparison instruction should be a signed integer comparison (icmp_signed)

guilhermeleobas: Hi @ianna, from the [docs](https://docs.python.org/3/c-api/bytes.html#c.PyBytes_AsStringAndSize), `PyBytes_AsStringAndSize` returns an integer. My suggestion is to change the return type in the signature to match CPython:

```python
        fnty = ir.FunctionType(
            ir.IntType(32),
            [self.pyobj, self.cstring.as_pointer(), self.py_ssize_t.as_pointer()],
        )
```

guilhermeleobas:
HI @ianna, I wrote a small snippet of code which tests `PyBytes_AsString`. You can use it as a template to test `PyBytes_AsStringAndSize`. The idea is that we create a Python Bytes object from a unicode value and use it as argument to `api.bytes_as_string`.

Perhaps there's a better way to test the API (cc @sklam).

```python
"""
Test Python API
"""


import ctypes
import unittest
from numba.core import types
from numba.core.extending import intrinsic
from numba import jit


@intrinsic
def _pyapi_bytes_as_string(typingctx, csrc, size):
    sig = types.voidptr(csrc, size)  # cstring == void*

        [csrc, size] = args
        api = context.get_python_api(builder)
        b = api.bytes_from_string_and_size(csrc, size)
        return api.bytes_as_string(b)
    return sig, codegen


def PyBytes_AsString(uni):
    # test_PyBytes_AsString will call this function with a unicode type.
    # We then use the underlying buffer to create a PyBytes object and call the
    # PyBytes_AsString function with PyBytes object as argument
    return _pyapi_bytes_as_string(uni._data, uni._length)


class TestPythonAPI(unittest.TestCase):
    def test_PyBytes_AsString(self):
        cfunc = jit(nopython=True)(PyBytes_AsString)
        cstr = cfunc('hello')  # returns a cstring

        fn = ctypes.pythonapi.PyBytes_FromString
        fn.argtypes = [ctypes.c_void_p]
        fn.restype = ctypes.py_object
        obj = fn(cstr)

        # Use the cstring created from bytes_as_string to create a python
        # bytes object
        self.assertEqual(obj, b'hello')


if __name__ == '__main__':
    unittest.main()

```

guilhermeleobas:
Hi @ianna, sorry for the really late review. I just saw you made progress in the PR. Could you merge Numba main branch into your branch? So that CI can run again

guilhermeleobas (changes requested):
Hi @ianna, could you add a test for `PyBytes_AsStringAndSize`? The idea for the test is similar to what you wrote:
* Get a pointer to a c string and size by calling PyBytes_AsStringAndSize` inside an intrinsic
* Use ctypes to call `PyBytes_FromStringAndSize` with the return values of the previous function

```python
@intrinsic
def _pyapi_bytes_as_string_and_size(typingctx, csrc, size):
    # return a tuple containing the c-string and size
    retty = types.Tuple.from_types((csrc, size))
    sig = retty(csrc, size)

        [csrc, size] = args
        pyapi = context.get_python_api(builder)
        b = pyapi.bytes_from_string_and_size(csrc, size)
        p_cstr = builder.alloca(pyapi.cstring)
        p_size = builder.alloca(pyapi.py_ssize_t)
        pyapi.bytes_as_string_and_size(b, p_cstr, p_size)

        cstr = builder.load(p_cstr)
        size = builder.load(p_size)
        tup = context.make_tuple(builder, sig.return_type, (cstr, size))
        return tup
    return sig, codegen


def PyBytes_AsStringAndSize(uni):
    return _pyapi_bytes_as_string_and_size(uni._data, uni._length)
```

ianna (commented):
@guilhermeleobas - I think, I'm done with this PR. Both functions have tests. Please, have a look. Thanks!

numba/tests/test_pythonapi.py
```diff
@@ -0,0 +1,82 @@
...
+        # Use the cstring created from bytes_as_string to create a python
+        # bytes object
+        self.assertEqual(obj, b'hello')
+
+    def test_PyBytes_AsStringAndSize(self):
+        cfunc = jit(nopython=True)(PyBytes_AsStringAndSize)
+        tup = cfunc('hello\x00world')  # returns a tuple: cstring and its size
+
+        fn = ctypes.pythonapi.PyBytes_FromStringAndSize
+        fn.argtypes = [ctypes.c_void_p]
```
guilhermeleobas, line 72:
```suggestion
        fn.argtypes = [ctypes.c_void_p, ctypes.c_size_t]
```

guilhermeleobas (changes requested):
Hi @ianna, thanks. Just one last change.

guilhermeleobas (approved):
LGTM. Thanks @ianna

PR #8467, opened 2022-09-27
Convert implementations using generated_jit to overload

gmarkall:
Mixing `generated_jit()` and `overload()` causes fallbacks to object mode - this should not be happening in Numba internal implementations. This commit converts all uses of `generated_jit()` into `overload()` for function implementations in Numba.

### DIFF
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -4620,7 +4620,7 @@ def cross2d(a, b):
     pass
 
 
-@overload
+@overload(cross2d)
 def cross2d_impl(a, b):
     if not type_can_asarray(a) or not type_can_asarray(b):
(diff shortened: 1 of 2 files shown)


gmarkall:
I originally wrote these changes as part of an attempt at removing pipeline fallback to see how it would go, where these tests passed. However, from the fails here it looks like the removal of the fallback harmonized `overload` and `generated_jit` somewhat, so some further change will be needed here to make tests pass again (if it is possible to make this change work without removing fallback).

gmarkall:
I must have been testing badly locally. The additional commit should(:crossed_fingers:) resolve the errors in the original commit.

guilhermeleobas (approved):
LGTM. Thanks @gmarkall

PR #8468, opened 2022-09-27
Reference test suite in installation documentation

### LINKED ISSUE #8452: Documentation doesn't explain how to run tests
## Reporting a bug

The [installing section](https://numba.readthedocs.io/en/stable/user/installing.html) in the docs doesn't explain how to run tests.

I guessed that the script might be ```runtests.py```. But it doesn't print progress info, unlike pytest does.

Documentation should have the command that runs tests.

apmasell:
As titled. Closes #8452.


docs/source/user/installing.rst
```diff
@@ -152,6 +152,8 @@ Then you can build and install Numba from the top level of the source tree::
 
     $ python setup.py install
 
+If you wish to run the test suite, see the :doc:`developer documentation <../developer/contributing>`.
```
stuartarchibald, line 155: Please could this be line wrapped to 80 chars?

Also, if there's no anchor creating a link to the "how to run the test suite" part of the contributing docs, then perhaps it'd be a good idea to add one in there so there's a direct link that is available that can be easily shared (it also would mean this reference could jump straight to the relevant part of the docs).

stuartarchibald, line 155:
```suggestion
If you wish to run the test suite, see the instructions in :doc:`developer documentation <../developer/contributing>`.
```

stuartarchibald (commented):

docs/source/user/installing.rst
stuartarchibald: Thanks, I see in 79c2b3a there's no specific ref/link pair. An example of this for e.g. references and linking to the "continuous integration testing" section is:
**anchor**:
https://github.com/numba/numba/blob/061f21a73a574ea1b3451910a36db4f9b3ae788b/docs/source/developer/contributing.rst?plain=1#L325

**reference**:
https://github.com/numba/numba/blob/061f21a73a574ea1b3451910a36db4f9b3ae788b/docs/source/developer/contributing.rst?plain=1#L75

apmasell: Sorry, didn't get pushed.

docs/source/user/installing.rst
```diff
@@ -152,6 +152,9 @@ Then you can build and install Numba from the top level of the source tree::
 
     $ python setup.py install
 
+If you wish to run the test suite, see the instructions in
+:doc:`developer documentation <running-tests>`.
```
stuartarchibald, line 156: This doesn't seem to render correctly in the RTD build, might need to be a `:ref:` ?

stuartarchibald:
NOTE: build failures are not related to this PR. #8486 is an attempt to fix.

docs/source/user/installing.rst
```diff
@@ -152,6 +152,9 @@ Then you can build and install Numba from the top level of the source tree::
 
     $ python setup.py install
 
+If you wish to run the test suite, see the instructions in
+:ref:`developer documentation <running-tests>`.
```
stuartarchibald, line 156:
```suggestion
If you wish to run the test suite, see the instructions in the
:ref:`developer documentation <running-tests>`.
```

stuartarchibald:
@apmasell I think mainline needs merging in to pick up the changes to the Azure CI job config for the windows build from #8486 else this PR will continue using the job config that is "broken". Once that's done builds should pass.

stuartarchibald (approved):
Thanks for the patch, the docs appear to be rendering correctly on inspection!

PR #8469, opened 2022-09-27
Correctly handle optional types in parfors lowering

### LINKED ISSUE #8453: Reproducible bug in parallel mode when prange loop contains a condition on an optional argument
numba version 0.55.2
python 3.9.13
linux debian

When running parallel mode and the prange contains a condition on an optional argument the compilation fails.

**This fails:**

@nb.njit("f8[:](f8[:], optional(u1))", nogil=True, cache=False, parallel=True, fastmath=False) 
def test(arr, test=None):
    for i in numba.prange(10):
        if test is not None:
            arr[i] = 0.0
    return arr

**This works:**

@nb.njit("f8[:](f8[:], optional(u1))", nogil=True, cache=False, parallel=True, fastmath=False) 
def test(arr, test=None):
    for i in numba.prange(10):
            arr[i] = 0.0
    return arr

**Error:**

cannot store {i8, i1} to {i8, i8}*: mismatching types

File "realizedMeasures.py", line 134:
def test(arr, ft=None):
    for i in nb.prange(10):
    ^

During: lowering "id=0[LoopNest(index_variable = parfor_index.2, range = (0, $const6.2, 1))]

apmasell:
Closes #8453

### DIFF
--- a/numba/tests/test_parfors.py
+++ b/numba/tests/test_parfors.py
@@ -236,6 +236,8 @@ class TestParforsBase(TestCase):
                     new_args.append(x.copy())
                 elif isinstance(x, numbers.Number):
                     new_args.append(x)
+                elif x is None:
+                    new_args.append(x)
                 elif isinstance(x, tuple):
                     new_args.append(copy.deepcopy(x))
                 elif isinstance(x, list):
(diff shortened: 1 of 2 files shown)


numba/tests/test_parfors.py
```diff
@@ -2130,6 +2130,16 @@ def test_impl(a, b, size):
...
 
+    def test_prange_optional(self):
+        @njit("f8[:](optional(u1))", parallel=True)
+        def foo(pred=None):
+            arr = np.ones(10)
+            for i in prange(1):
+                if pred is not None:
+                    arr[i] = 0.0
+            return arr
+        self.assertEqual(foo().sum(), 10.0)
```
stuartarchibald, line 2141: @apmasell Perhaps take a look at the first code sample in: https://github.com/numba/numba/issues/8453#issuecomment-1253372598, this function needs to be called with an `types.Optional`, a default kwarg goes through as `types.Omitted`.

Also, this should use the `self.check` functionality to assert the validity of the result.

stuartarchibald:
@apmasell thanks for applying the patch and adding the test as noted in/to resolve #8453.

@DrTodd13 as this contains a fix suggested by me in https://github.com/numba/numba/issues/8453#issuecomment-1253372598, please could you review this? Many thanks!

DrTodd13 (approved):

stuartarchibald:
Thanks for the review @DrTodd13.

PR #8473, opened 2022-09-28
change the include style in _pymodule.h and remove unused or duplicate headers in two header files

dlee992:
Hi, guys.

In this PR, I changed the include style from <..> to ".." in `_pymodule.h`. Since I want to use some Numba headers in my numba extension, and meanwhile I use bazel to build a `cc` file in my extension (because I also depends on a 3rd C++ library). But because of these "ill-formed" include usages, bazel will output some errors to me, like this:
```
external/numba/cext/../_pymodule.h:6:10: error: 'Python.h' file not found with <angled> include; use "quotes" instead
```

Meanwhile, I also deleted `Python.h` in `listobject.h` and `dictobject.h`, because `cext.h` will include this for them. I also deleted `../_pymodule.h` in `dictobject.h`, since they seem not use any marcos or other things in it. 

After these changes, the building process seems still good. 

PS, I hope these changes in `_pymodule.h` can be merged into a recent Numba release version at least, thanks.

=============
updates [Sept. 29]: After changing the include style in the `_pymodule.h`, bazel can successfully compile the `cc` file for me.

Sure, I have done some tricky things in `WORKSPACE`:
```
# Convert the python include dir to a local repo
new_local_repository(
    name = "python",
    # FIXME: DON'T use hardcode python include path
    path = "/Users/ld/opt/anaconda3/envs/numba-pandas/include/python3.10",
    build_file_content = """
package(default_visibility = ["//visibility:public"])
cc_library(
    name = "headers",
    hdrs = glob(["**/*.h"])
)
"""
)

# Convert the site-packages/numba dir to a local repo
new_local_repository(
    name = "numba",
    path = "/Users/ld/opt/anaconda3/envs/numba-pandas/lib/python3.10/site-packages/numba",
    build_file_content = """
package(default_visibility = ["//visibility:public"])
cc_library(
    name = "headers",
    hdrs = glob(["**/*.h"])
)
"""
)
```
Then, add these two headers into the cc_binary arguments.

### DIFF
--- a/numba/_pymodule.h
+++ b/numba/_pymodule.h
@@ -3,9 +3,9 @@
 
 #define PY_SSIZE_T_CLEAN
 
-#include <Python.h>
-#include <structmember.h>
-#include <frameobject.h>
+#include "Python.h"
+#include "structmember.h"
+#include "frameobject.h"
 
 #define MOD_ERROR_VAL NULL
 #define MOD_SUCCESS_VAL(val) val
--- a/numba/cext/dictobject.h
+++ b/numba/cext/dictobject.h
@@ -1,6 +1,4 @@
 /* Adapted from CPython3.7 Objects/dict-common.h */
-#include "Python.h"
-#include "../_pymodule.h"
 #include "cext.h"
 
 #ifndef NUMBA_DICT_COMMON_H
--- a/numba/cext/listobject.h
+++ b/numba/cext/listobject.h
@@ -10,7 +10,6 @@
 #ifndef NUMBA_LIST_H
 #define NUMBA_LIST_H
 
-#include "Python.h"
 #include "cext.h"
 
 typedef void (*list_refcount_op_t)(const void*);


gmarkall (approved):
This looks good to me - I think that the change is a move in the right direction - my understanding is that in general the search order of both `<` / `>` and `"` / `"` includes is implementation-defined, but in general one ought to only expect system paths to be searched when using angle brackets. For example from the [GCC Search Path docs](https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html):

> For the angle-bracket form `#include <file>`, the preprocessor’s default behavior is to look only in the standard system directories.

We have been getting away with this because the distutils / setuptools / Python build setup (I'm not sure exactly which / what) adds the Python include dir as a system dir - for example, I see `-isystem /home/gmarkall/mambaforge/envs/numbadev/include` on the compiler's command line when building Numba. I am actually a little surprised that this never caused an issue before :-)

I do think this should have a full buildfarm run to ensure it doesn't cause a surprise build failure on any of the platforms we support, but other than that I have confidence in this change.

gmarkall:
@esc / @stuartarchibald Could this have a buildfarm run please?

stuartarchibald:
Thanks for writing up our OOB discussion @gmarkall. The `cext.h` header already uses `#include "Python.h"` and as a result the explicit inclusion of this in the typed container headers is indeed superfluous. Changing the search path also makes sense for the reasons outlined by @gmarkall.

Thanks for the patch @dlee992 !

gmarkall:
@sklam / @stuartarchibald Can this have a buildfarm run please? I think it's the only thing it's waiting on since we decided this change was OK, but want to check there's no surprises on all platforms.

sklam:
BFID: `numba_smoketest_cpu_yaml_158`

stuartarchibald:
> BFID: `numba_smoketest_cpu_yaml_158`

This passed with the exception of a known failure on `osx-arm64` due to #8567.

PR #8474, opened 2022-09-29
Remove setuptools as a runtime dependency

cpcloud:
This PR removes setuptools as a runtime dependency, to make downstream packaging
with tools that don't allow dependency cycles (e.g., nix) easier.

I couldn't find any actual runtime uses of `setuptools`, only build time, so
I'm happy to close this out if somehow `setuptools` is required at runtime.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -371,11 +371,13 @@ def get_ext_modules():
 
 packages = find_packages(include=["numba", "numba.*"])
 
-build_requires = ['numpy >={}'.format(min_numpy_build_version)]
+build_requires = [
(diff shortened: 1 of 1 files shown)


esc:
Related: https://github.com/numba/numba/pull/8366

cpcloud:
Closing this out in favor of whatever comes of #8366.

esc:
> Closing this out in favor of whatever comes of #8366.

Thank you for following up, I think the patch here should be cherry-picked, I don't see this covered there.

PR #8475, opened 2022-09-29
Remove setuptools version pin

gmarkall:

### DIFF
--- a/buildscripts/condarecipe.local/meta.yaml
+++ b/buildscripts/condarecipe.local/meta.yaml
@@ -31,7 +31,7 @@ requirements:
   host:
     - python
     - numpy
-    - setuptools <60
+    - setuptools
     - importlib_metadata       # [py<39]
     # On channel https://anaconda.org/numba/
     - llvmlite >=0.40.0dev0,<0.40
@@ -45,7 +45,7 @@ requirements:
     - python >=3.7
     # NumPy 1.22.0, 1.22.1, 1.22.2 are all broken for ufuncs, see #7756
     - numpy >=1.18, !=1.22.0, !=1.22.1, !=1.22.2
-    - setuptools <60
+    - setuptools
     - importlib_metadata       # [py<39]
     # On channel https://anaconda.org/numba/
     - llvmlite >=0.40.0dev0,<0.40
@@ -74,7 +74,7 @@ test:
     - cffi
     - scipy
     - ipython                  # [not aarch64]
-    - setuptools <60
+    - setuptools
     - tbb  >=2021              # [not (aarch64 or ppc64le)]
     - llvm-openmp              # [osx]
     # This is for driving gdb tests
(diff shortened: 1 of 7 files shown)


gmarkall:
> @gmarkall is this a re-roll of #8366 and #8474 ? Do we want to use this one instead and close #8366 ?

No, it is still a dependency but not on any particular version. This is just a quick fix for the issues it causes (with potentially some new test issues introduced in tests for warnings, that may show up on the buildfarm).

gmarkall:
Marking is RFR to see what people (and potentially the buildfarm run) think.

stuartarchibald:
Thanks for the patch @gmarkall, as discussed OOB and in various issues, this appears to be an approximate revert of #8356 which makes the version of the `setuptools` runtime dependency unrestricted. On inspection only the patch looks good.

I was considering that this test might fail:
https://github.com/numba/numba/blob/061f21a73a574ea1b3451910a36db4f9b3ae788b/numba/tests/test_import.py#L96-L103
but on second look, I don't think it will as importing `numba` doesn't import `numba.pycc` and so the `numpy.distutils` module won't be accessed.

sklam:
I tested it on my mac and on win64 with setuptools=65 and numpy=1.23.

gmarkall:
How did the buildfarm run go @esc?

gmarkall:
I've put this in the 0.56.3RC milestone for now - I think this is the change we intend to consider for it (please let me know if this understanding was not right).

esc:
> How did the buildfarm run go @esc?

There were some issues that @sklam is looking into.

gmarkall:
> > How did the buildfarm run go @esc?
> 
> There were some issues that @sklam is looking into.

Just to check, these are expected to be problems with the buildfarm and not this PR?

esc:
> > > How did the buildfarm run go @esc?
> > 
> > 
> > There were some issues that @sklam is looking into.
> 
> Just to check, these are expected to be problems with the buildfarm and not this PR?

This seems likely, yes, but isn't 100% conclusive, yet.

sklam (approved):
Ok on the farm

mbargull:
It probably makes sense to restrict the build time dependency (runtime dependency might be fine to be unrestricted) given that `setuptools` can break `numpy.distutils` as happened again with the latest `setuptools=65.6.0` release and NumPy themselves pin to `setuptools<60.0`.
See https://github.com/numpy/numpy/issues/22623#issuecomment-1321012633 and issues linked therein.

stuartarchibald:
> It probably makes sense to restrict the build time dependency (runtime dependency might be fine to be unrestricted) given that `setuptools` can break `numpy.distutils` as happened again with the latest `setuptools=65.6.0` release and NumPy themselves pin to `setuptools<60.0`. See [numpy/numpy#22623 (comment)](https://github.com/numpy/numpy/issues/22623#issuecomment-1321012633) and issues linked therein.

Thanks for raising this @mbargull, I've opened #8617 to ensure that this is discussed by the maintainers at the next triage meeting. There are a few patches pending review for the next Numba release that are hopefully going to remove a lot of the reliance on `setuptools` and `numpy.distutils` which might decrease the exposure to upstream changes. Thanks again!

PR #8482, opened 2022-10-03
Fix #8477: Allow copies with different strides for 0-length data

### LINKED ISSUE #8477: CUDA: Copying zero-size device array to zero-size host array can fail if strides differ
The following:

```python
from numba import cuda
import numpy as np

shape = (0,)
dtype = 'f4'
order = 'C'

ary = np.arange(np.product(shape), dtype=dtype)
ary = np.ndarray(shape, dtype=ary.dtype, buffer=ary.data, order=order)
x = cuda.to_device(ary)

hx = np.empty_like(ary)
x.copy_to_host(hx)
```

when run with NumPy 1.23 results in:

```
$ python repro.py 
  File "/home/gmarkall/numbadev/issues/distributed-serialize/repro.py", line 13, in <module>
    x.copy_to_host(hx)
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudadrv/devices.py", line 232, in _require_cuda_context
    return fn(*args, **kws)
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudadrv/devicearray.py", line 273, in copy_to_host
    check_array_compatibility(self, ary)
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudadrv/devicearray.py", line 899, in check_array_compatibility
    raise ValueError('incompatible strides: %s vs. %s' %
ValueError: incompatible strides: (4,) vs. (0,)
```

One workaround for this is:

```diff
diff --git a/numba/cuda/cudadrv/devicearray.py b/numba/cuda/cudadrv/devicearray.py
index 651f06bf6..dd8878cab 100644
--- a/numba/cuda/cudadrv/devicearray.py
+++ b/numba/cuda/cudadrv/devicearray.py
@@ -895,6 +895,6 @@ def check_array_compatibility(ary1, ary2):
     if ary1sq.shape != ary2sq.shape:
         raise ValueError('incompatible shape: %s vs. %s' %
                          (ary1.shape, ary2.shape))
-    if ary1sq.strides != ary2sq.strides:
+    if (ary1sq.strides != ary2sq.strides) and ary1.size:
         raise ValueError('incompatible strides: %s vs. %s' %
                          (ary1.strides, ary2.strides))
```

which is probably a reasonable workaround, given that the strides don't matter for zero-sized arrays.

One question is whether `numpy.empty_like()` should be copying the strides of its argument. If I add:

```python
print(f"{ary.shape}, {ary.strides}")
print(f"{x.shape}, {x.strides}")
print(f"{hx.shape}, {hx.strides}")
```

to the above example, I get:

```
(0,), (4,)
(0,), (4,)
(0,), (0,)
```

Maybe this should come out as `(0,), (4,)` on all lines (i.e. the strides of `hx` should be `(4,)` too).

gmarkall:
Since NumPy 1.23, it is possible for zero-length ndarrays to have different strides (e.g. `(0,)`, and `(8,)`). If we attempt to copy from a zero-length device array to a zero-length host array where the strides differ, our compatibility check fails because it compares strides.

This commit fixes the issue by only considering strides when checking compatibility of nonzero-length arrays. I believe this to be valid because the following works normally with NumPy 1.23:

```python
import numpy as np

ary1 = np.arange(0)
ary2 = np.ndarray((0,), buffer=ary1.data)
ary3 = np.empty_like(ary2)

ary3[:] = ary2
ary3[...] = ary2
np.copyto(ary2, ary3)
```

i.e. copying zero-length arrays with different strides generally works as expected.

The included test is written in such a way that it should test this change in behaviour regardless of the installed NumPy version - we explicitly construct zero-length device and host arrays with differing strides. The additional sanity check ensures that the host array has the strides we expect, just in case there is some version of NumPy in which setting the strides explicitly didn't result in the expected strides - I have observed that requesting nonzero strides for a zero length array can result still in zero strides (a separate but related behaviour), so this sanity check is provided to account for any other unexpected behaviour of this nature. I have tested locally with NumPy 1.22 and 1.23 (pre- and post-changes to strides).

See also https://github.com/dask/distributed/pull/7089 where a workaround for an observation of this issue was needed. This would not be needed with the fix in this commit.

Fixes #8477.

### DIFF
--- a/numba/cuda/cudadrv/devicearray.py
+++ b/numba/cuda/cudadrv/devicearray.py
@@ -895,6 +895,8 @@ def check_array_compatibility(ary1, ary2):
     if ary1sq.shape != ary2sq.shape:
         raise ValueError('incompatible shape: %s vs. %s' %
                          (ary1.shape, ary2.shape))
-    if ary1sq.strides != ary2sq.strides:
+    # We check strides only if the size is nonzero, because strides are
+    # irrelevant (and can differ) for zero-length copies.
+    if ary1.size and ary1sq.strides != ary2sq.strides:
         raise ValueError('incompatible strides: %s vs. %s' %
                          (ary1.strides, ary2.strides))
--- a/numba/cuda/tests/cudadrv/test_cuda_ndarray.py
+++ b/numba/cuda/tests/cudadrv/test_cuda_ndarray.py
@@ -428,19 +428,31 @@ class TestCudaNDArray(CUDATestCase):
         # https://github.com/numba/numba/issues/8477.
 
         # Create a device array with shape (0,) and strides (8,)
-        src = devicearray.DeviceNDArray(shape=(0,), strides=(8,), dtype=np.int8)
+        dev_array = devicearray.DeviceNDArray(shape=(0,), strides=(8,),
+                                              dtype=np.int8)
 
         # Create a host array with shape (0,) and strides (0,)
-        dest = np.ndarray(shape=(0,), strides=(0,), dtype=np.int8)
+        host_array = np.ndarray(shape=(0,), strides=(0,), dtype=np.int8)
 
         # Sanity check for this test - ensure our destination has the strides
         # we expect, because strides can be ignored in some cases by the
         # ndarray constructor - checking here ensures that we haven't failed to
         # account for unexpected behaviour across different versions of NumPy
-        self.assertEqual(dest.strides, (0,))
+        self.assertEqual(host_array.strides, (0,))
 
-        # Ensure that the copy succeeds
-        src.copy_to_host(dest)
+        # Ensure that the copy succeeds in both directions
+        dev_array.copy_to_host(host_array)
+        dev_array.copy_to_device(host_array)
+
+        # Ensure that a device-to-device copy also succeeds when the strides
+        # differ - one way of doing this is to copy the host array across and
+        # use that for copies in both directions.
+        dev_array_from_host = cuda.to_device(host_array)
+        self.assertEqual(dev_array_from_host.shape, (0,))
+        self.assertEqual(dev_array_from_host.strides, (0,))
+
+        dev_array.copy_to_device(dev_array_from_host)
+        dev_array_from_host.copy_to_device(dev_array)
 
 
 class TestRecarray(CUDATestCase):
--- a/numba/cuda/tests/cudadrv/test_cuda_ndarray.py
+++ b/numba/cuda/tests/cudadrv/test_cuda_ndarray.py
@@ -420,6 +420,28 @@ class TestCudaNDArray(CUDATestCase):
         got = np.asarray(dary)
         self.assertEqual(got.dtype, dary.dtype)
 
+    def test_issue_8477(self):
+        # Ensure that we can copy a zero-length device array to a zero-length
+        # host array when the strides of the device and host arrays differ -
+        # this should be possible because the strides are irrelevant when the
+        # length is zero. For more info see
+        # https://github.com/numba/numba/issues/8477.
+
+        # Create a device array with shape (0,) and strides (8,)
+        src = devicearray.DeviceNDArray(shape=(0,), strides=(8,), dtype=np.int8)
+
+        # Create a host array with shape (0,) and strides (0,)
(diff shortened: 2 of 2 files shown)


gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
Since making this PR I also decided it would be better to test copies in all directions, so the additional commit adds other cases (host-to-device, and device-to-device).

gmarkall:
gpuci run tests

numba/cuda/cudadrv/devicearray.py
```diff
@@ -895,6 +895,8 @@ def check_array_compatibility(ary1, ary2):
     if ary1sq.shape != ary2sq.shape:
         raise ValueError('incompatible shape: %s vs. %s' %
                          (ary1.shape, ary2.shape))
-    if ary1sq.strides != ary2sq.strides:
+    # We check strides only if the size is nonzero, because strides are
+    # irrelevant (and can differ) for zero-length copies.
+    if ary1.size and ary1sq.strides != ary2sq.strides:
```
stuartarchibald, line 900: NOTE: cases like:
```python
ary1 = np.ndarray(shape=(1, 1, 1), strides=(0, 2, 4), dtype=np.float64)
ary2 = np.ndarray(shape=(), strides=(), dtype=np.float64)
```
are "safe" due to the `squeeze` above. It shouldn't be possible to reach here and need to check `ary2.size`.

numba/cuda/tests/cudadrv/test_cuda_ndarray.py
```diff
@@ -420,6 +420,41 @@ def test_bug6697(self):
...
+
+        # Ensure that a device-to-device copy also succeeds when the strides
+        # differ - one way of doing this is to copy the host array across and
+        # use that for copies in both directions.
+        dev_array_from_host = cuda.to_device(host_array)
+        self.assertEqual(dev_array_from_host.shape, (0,))
+        self.assertEqual(dev_array_from_host.strides, (0,))
+
+        dev_array.copy_to_device(dev_array_from_host)
+        dev_array_from_host.copy_to_device(dev_array)
```
stuartarchibald, line 456: I guess there's no merit in checking the values of the copies as this is about testing that the copies have succeeded without error?

stuartarchibald (commented):
Thanks for the patch @gmarkall, this looks like an appropriate fix for the issue described. One minor query on the testing to resolve but otherwise looks good. Thanks again!

numba/cuda/tests/cudadrv/test_cuda_ndarray.py
gmarkall: I think there's no merit - I think there's already tests that copies work, this test is focused on an edge case in checking compatibility. Additionally, since these are empty arrays, there's not a lot to look at to check.

numba/cuda/cudadrv/devicearray.py
gmarkall: Also checked manually with:

```python
from numba import cuda
import numpy as np

x  = np.ndarray(shape = (1, 1, 1), strides = (0, 2, 4), dtype=np.float64)
y  = np.ndarray(shape = (), strides = (), dtype=np.float64)

print(x.shape, x.strides, x.size)
print(y.shape, y.strides, y.size)

dx = cuda.to_device(x)
dy = cuda.to_device(y)

dx.copy_to_host(y)
dy.copy_to_host(x)
```

without issue.

stuartarchibald: Great, thanks for checking this too!

numba/cuda/tests/cudadrv/test_cuda_ndarray.py
stuartarchibald: Thanks for confirming, I'm inclined to agree, but thought it best to check!

stuartarchibald (approved):
Thanks for the patch!

sklam:
BFID `numba_smoketest_cuda_yaml_161`

gmarkall:
How did the buildfarm run go?

esc:
> How did the buildfarm run go?

It was fine.

Issue #8036, opened 2022-05-11
Raising exceptions in CUDA device functions does not work correctly


gmarkall:
The following:

```python
from numba import cuda

@cuda.jit(device=True)
def f():
    raise ValueError('Error')

@cuda.jit(debug=True, opt=False)
def kernel():
    f()

kernel[1, 1]()
```

results in an error like:

```
Traceback (most recent call last):
  File "/home/gmarkall/numbadev/issues/dispatcher-refactor/raise_in_devfunc.py", line 14, in <module>
    kernel[1, 1]()
  File "/home/gmarkall/mambaforge/envs/numba0531/lib/python3.9/site-packages/numba/cuda/compiler.py", line 821, in __call__
    return self.dispatcher.call(args, self.griddim, self.blockdim,
  File "/home/gmarkall/mambaforge/envs/numba0531/lib/python3.9/site-packages/numba/cuda/compiler.py", line 966, in call
    kernel.launch(args, griddim, blockdim, stream, sharedmem)
  File "/home/gmarkall/mambaforge/envs/numba0531/lib/python3.9/site-packages/numba/cuda/compiler.py", line 716, in launch
    exccls, exc_args, loc = self.call_helper.get_exception(code)
ValueError: not enough values to unpack (expected 3, got 2)
```

If I fix the `MinimalCallHelper` so that reporting the error works correctly (`get_exception()` should return three things, not two):

```diff
diff --git a/numba/core/callconv.py b/numba/core/callconv.py
index b3fa6cbbd..e99494524 100644
--- a/numba/core/callconv.py
+++ b/numba/core/callconv.py
@@ -315,7 +315,10 @@ class _MinimalCallHelper(object):
             return self.exceptions[exc_id]
         except KeyError:
             msg = "unknown error %d in native function" % exc_id
-            return SystemError, (msg,)
+            exc = SystemError
+            exc_args = (msg,)
+            locinfo = None
+            return exc, exc_args, locinfo
 
 # The structure type constructed by PythonAPI.serialize_uncached()
 # i.e a {i8* pickle_buf, i32 pickle_bufsz, i8* hash_buf}
```

then the exception becomes:

```
Traceback (most recent call last):
  File "/home/gmarkall/numbadev/issues/dispatcher-refactor/raise_in_devfunc.py", line 14, in <module>
    kernel[1, 1]()
  File "/home/gmarkall/numbadev/numba/numba/cuda/dispatcher.py", line 473, in __call__
[...]

PR #8484, opened 2022-10-03
Initial commit for raise/assert in device functions

testhound:
This fixes raise/assert so it works with device functions.

### DIFF
--- a/numba/core/callconv.py
+++ b/numba/core/callconv.py
@@ -286,6 +286,14 @@ class MinimalCallConv(BaseCallConv):
 
 
 class _MinimalCallHelper(object):
+    _instance = None
+    exceptions = {}
+    def __new__(cls, *args, **kwargs):
+        if not cls._instance:
+            cls._instance = super(_MinimalCallHelper, cls).__new__(
+                                cls, *args, **kwargs)
+        return cls._instance
+
     """
     A call helper object for the "minimal" calling convention.
     User exceptions are represented as integer codes and stored in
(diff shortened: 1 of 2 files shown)


gmarkall (changes requested):
Having reflected on this, I don't think this making the minimal call helper a singleton is the right way to go. I believe the issue (#8036) is because we have multiple call helper objects when a device function is called, and the exception in the device function gets added to the wrong one (they should all get added to the kernel's call helper). Making the call helper a singleton doesn't address why we have this behaviour - something is wrong in the design around the use of call helpers - it just hides the issue that results from this design error / bug.

I don't know what the correct fix is, but I think it will become clearer by answering:

- Why do we have different call helpers for callers and callees?
- Why does the exception get added to the device function's call helper?
- How is it working correctly on the CPU target?

testhound:
> Having reflected on this, I don't think this making the minimal call helper a singleton is the right way to go. I believe the issue (#8036) is because we have multiple call helper objects when a device function is called, and the exception in the device function gets added to the wrong one (they should all get added to the kernel's call helper). Making the call helper a singleton doesn't address why we have this behaviour - something is wrong in the design around the use of call helpers - it just hides the issue that results from this design error / bug.
> 
> I don't know what the correct fix is, but I think it will become clearer by answering:
> 
>     * Why do we have different call helpers for callers and callees?
> 
>     * Why does the exception get added to the device function's call helper?
> 
>     * How is it working correctly on the CPU target?

One thing is clear: the cpu target does not use call helper objects. The _make_call_helper for the CpuCallConv looks like this:

def _make_call_helper(self, builder):
        return None

gmarkall:
OK, so how does the CPU target correlate information about exceptions raised during execution of jitted code with the exceptions as written in the source? Perhaps we shouldn't be using call helpers for exceptions in the CUDA target - maybe there is a design flaw here?

PR #8413, opened 2022-09-01
Fix Azure CI for NumPy 1.23 and use conda-forge scipy

sklam:

PR #8490, opened 2022-10-05
Restore installing SciPy from defaults instead of conda-forge on public CI

esc:
This fixes fetching SciPy from defaults instead of conda-forge, now that a suitable SciPy 1.9 has become available.

This is a partial revert of #8413 since only the changes for the public ci build scripts, that pertain to the installation of SciPy are included.


stuartarchibald (approved):
Thanks for the patch.

sklam:
there are merge conflict

esc:
@sklam @stuartarchibald I opted to rebase onto `main` in this case, since the change was quite short and no comments had been made. This will need a re-review now.

PR #8238, opened 2022-07-10
Advanced Indexing Support #1

kc611:
This PR refactors the fancy index selection of of NumPy array, which fixes the following cases:

- When an integer index is just before or just after an array index, the case is supported by our current advanced indexing logic, however is blocked at typing level.

```python
import numba
import numpy as np

@numba.njit
def func(x, idx):
    return x[:, 2, idx, :, :]

a = np.random.random_integers(0, 100, (10, 11, 12, 13, 14))
b = np.random.random_integers(0, 10, (4,))

print(np.allclose(func(a, b), func.py_func(a, b)))
# Before PR: Error, but we have the support
# After PR: True
```

- When an integer index is after an array index with slices in between (multiple indexing subspaces), the case is not supported by  our current advanced indexing logic, however is not being blocked at typing level.

```python
import numba
import numpy as np

@numba.njit
def func(x, idx):
    return x[:, :, idx, :, 4]

a = np.random.random_integers(0, 100, (10, 11, 12, 13, 14))
b = np.random.random_integers(0, 10, (4,))

print(np.allclose(func(a, b), func.py_func(a, b)))
# Before PR: False, but runs so it's a bug
# After PR: raises Error correctly
```

This also clearly separates out the various limitations of advanced indexing in Numba so that we can add support/remove the limitation one at a time.

PR #8491, opened 2022-10-05
Added support for multidimensional array indexing

kc611:
This PR builds on top of #8238 

As titled, this PR adds support for multidimensional indices while indexing NumPy arrays in Numba JIT functions: 

```python
import numba
import numpy as np

@numba.njit
def func(x, idx):
    return x[:, 2, idx, :, :]

a = np.random.randint(0, 100, (10, 11, 12, 13, 14))
b = np.random.randint(0, 10, (4, 5)) # Previously, these were limited to 1-D.

print(func(a, b).shape) 
# (10, 4, 5, 13, 14)
print(np.allclose(func(a, b), func.py_func(a, b)))
# True
```


kc611:
The memory leak I was facing in this issue has been isolated: 

```python
import numpy as np
import unittest
from numba import njit
from numba.tests.support import MemoryLeakMixin, TestCase

class TestFancyIndexingMultiDim(MemoryLeakMixin, TestCase):
    shape = (4, 5, 6)
    
    def check_setitem_indices(self, arr_shape, index):
        @njit     
        def set_item(array, idx, item):
            array[idx] = item

        arr = np.random.randint(0, 11, size=arr_shape)
        src = arr[index]
        expected = np.zeros_like(arr)
        got = np.zeros_like(arr)

        set_item.py_func(expected, index, src)
        set_item(got, index, src)

        np.testing.assert_equal(got, expected)

    def test_setitem_with_tuple(self):
        # When index is an array within a tuple
        idx = (np.array([1,2,3]),)
        # No memory leak
        self.check_setitem_indices(self.shape, idx)

    def test_setitem_without_tuple(self):
        # When index itself is an array, it is
        # supposed to be implicitly cast to same form as
        # the example above i.e. both should and do 
        # give same answer, except for the leak
        idx = np.array([1,2,3])
        # Memory leaks
        self.check_setitem_indices(self.shape, idx)
    
if __name__ == '__main__':
    unittest.main()

```

stuartarchibald:
@kc611 please could you resolve conflicts against `main`? As #8238 is merged the diff will hopefully shrink. Many thanks.

numba/np/arrayobj.py
```diff
@@ -1690,67 +1699,74 @@ def fancy_setslice(context, builder, sig, args, index_types, indices):
...
-    # Loop on destination and copy from source to destination
-    dest_indices, counts = indexer.begin_loops()
-
-    # Source is iterated in natural order
-
-    # Counts represent a counter for the number of times a specified axis
-    # is being accessed, during setitem they are used as source
-    # indices
-    counts = list(counts)
+    if src_type == 'buffer':
```
stuartarchibald, line 1711: I don't think that this path in its present form can be supported on CUDA as the `flat_imp*` function compiled below relies on:
* NumPy functions
* CPU only implementations for `reshape`
* returns an array
* needs the NRT

I'm not sure of the best way to "detect" this, immediate thoughts are to either require the NRT and raise if the context doesn't have it, or alternatively declare these functions as overloads targetting CPU only which will then fail for CUDA users. CC @gmarkall do you have an opinion on this?

numba/np/arrayobj.py
```diff
@@ -1131,6 +1150,12 @@ def fancy_getitem(context, builder, sig, args,
 
     indexer.end_loops()
 
+    for _indexer in indexer.indexers:
+        if isinstance(_indexer, IntegerArrayIndexer) \
+           and hasattr(_indexer, "idxary_instr"):
+            context.nrt.decref(builder, _indexer.idxty,
+                               _indexer.idxary_instr)
```
stuartarchibald, line 1157: As per a later comment, this _requires_ the presence of the NRT which isn't guaranteed and needs guarding against.

numba/np/arrayobj.py
```diff
@@ -1658,25 +1684,8 @@ def fancy_setslice(context, builder, sig, args, index_types, indices):
             msg = "cannot assign slice from input of different size"
             context.call_conv.return_user_exc(builder, ValueError, (msg,))
 
-        # Check for array overlap
-        src_start, src_end = get_array_memory_extents(context, builder, srcty,
```
stuartarchibald, line 1662: Given this check is removed, does the proposed implementation correctly handle "overlap"?

numba/tests/test_fancy_indexing.py
stuartarchibald, line 1: It would be good for the testing to now also include indexing into F-order and A-order arrays?

stuartarchibald (commented):
Thanks for the PR @kc611, great to see this implemented. I've given this an initial review and have also done some initial manual testing. Once the comments are address I'll look at the implementation details more closely, but from a cursory inspection the approach seems like it should work. Thanks again!

numba/core/typing/arraydecl.py
```diff
@@ -85,9 +85,12 @@ def get_array_index_type(ary, idx):
         elif (isinstance(ty, types.Array)
               and isinstance(ty.dtype, (types.Integer, types.Boolean))):
             if ty.ndim > 1:
-                # Advanced indexing limitation # 1
-                raise NumbaTypeError(
-                    "Multi-dimensional indices are not supported.")
+                if isinstance(ty.dtype, (types.Boolean,)):
+                    msg = ("You may only use one dimensional boolean index per axis."
```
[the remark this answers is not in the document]
kc611: 'One-dimensional boolean indexing array per axis of source array / multidimensional boolean index array in a single axis of source array are not allowed.' is what it's trying to convey.

I'll add clarification in the `msg`

numba/np/arrayobj.py
```diff
@@ -745,15 +746,44 @@ class IntegerArrayIndexer(Indexer):
...
+
+            # If the index array is contigous, use the nocopy version
+            if idxty.is_contig:
+                flat_imp = flat_imp_nocopy
+            # otherwise, use copy version since we don't support
+            # reshaping non-contigous arrays
+            else:
+                flat_imp = flat_imp_copy
+
+            retty = types.Array(idxty.dtype, 1, idxty.layout, readonly=True)
```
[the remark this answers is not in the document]
kc611: This is the flattened indexing array not the resultant array, it's used only for internal operations of `Indexer`s and discarded when the operation completes.

numba/np/arrayobj.py
```diff
@@ -745,15 +746,44 @@ class IntegerArrayIndexer(Indexer):
...
+            self.idxary = idxary
+
+        assert self.idxty.ndim == 1
+
     def prepare(self):
         builder = self.builder
-        self.idx_size = cgutils.unpack_tuple(builder, self.idxary.shape)[0]
+        self.idx_size = self.ll_intp(1)
+        for _shape in self.idx_shape:
+            self.idx_size = self.builder.mul(self.idx_size, _shape)
```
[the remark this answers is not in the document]
kc611: I'd say improbable, theoretically you'll need to have an array of size 2^32 to overflow in a 32-bit system.

numba/np/arrayobj.py
kc611: Define overlap in context of arrays ?

numba/tests/test_fancy_indexing.py
kc611: Alright will add that.

numba/np/arrayobj.py
```diff
@@ -1690,67 +1699,74 @@ def fancy_setslice(context, builder, sig, args, index_types, indices):
...
-    source_indices = [c for c in counts if c is not None]
-
-    val = src_getitem(source_indices)
+        # If the source array is contigous, use the nocopy version
+        if srcty.is_contig:
+            flat_imp = flat_imp_nocopy
+        # otherwise, use copy version since we don't support
+        # reshaping non-contigous arrays
+        else:
+            flat_imp = flat_imp_copy
```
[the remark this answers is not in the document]
kc611: I checked and for some reason, the source index array is always non-contiguous. So it's the other way around, the former branch (no_copy) isn't being hit.

numba/core/typing/arraydecl.py
```diff
@@ -85,9 +85,12 @@ def get_array_index_type(ary, idx):
...
+                           " NumPy behaviour of multidimensional boolean indices for "
+                           "implicitly indexing multiple axes is not supported.")
+                    raise NumbaTypeError(msg)
```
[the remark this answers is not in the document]
kc611: Probably not, will add one.

numba/np/arrayobj.py
```diff
@@ -745,15 +746,44 @@ class IntegerArrayIndexer(Indexer):
...
+            res = context.compile_internal(builder, flat_imp, sig,
+                                           (idxary._getvalue(),))
+            self.idxty = retty
+            self.idxary = make_array(retty)(context, builder, res)
+            self.idxary_instr = res
+        else:
+            self.idxty = idxty
+            self.idxary = idxary
+
+        assert self.idxty.ndim == 1, <message>
```
sklam, line 780: Bad syntax

numba/np/arrayobj.py
kc611: With the current implementation, I don't think that's possible since fancy indexing always creates a copy of the data. i.e. In fancy indexing the source and destination array never share memory. This is stated in the NumPy documentation as well. 

I'm in-fact curious as of why it was the case over here that they did.

stuartarchibald (commented):
Thanks for the updates @kc611. I've given them a review and I think there's a few things to be resolved but otherwise looks good. The unit tests that are failing CI are failing due to the use of `compile_isolated` and `Flags()`. To fix, this:
https://github.com/numba/numba/blob/74cb715132519cbe5dcc4ea963c6953ef424861d/numba/tests/test_indexing.py#L15-L16
could have:
```
enable_pyobj_flags.nrt = True
```
adding. Or alternatively, as object mode fallback is deprecated, forced object mode testing could be achieved by using alternative flags such as those available through importing these from testing support:
https://github.com/numba/numba/blob/74cb715132519cbe5dcc4ea963c6953ef424861d/numba/tests/support.py#L70-L71
Hope this helps?

numba/tests/test_fancy_indexing.py
```diff
@@ -3,11 +3,13 @@
...
 import unittest
-from numba import jit, typeof, njit
-from numba.core import types
+from numba import jit, njit
 from numba.core.errors import TypingError
 from numba.tests.support import MemoryLeakMixin, TestCase
+from numba.core.compiler import compile_isolated, Flags
 
+enable_pyobj_flags = Flags()
+enable_pyobj_flags.nrt = True
```
stuartarchibald, line 12: dead code/imports?

numba/tests/test_fancy_indexing.py
```diff
@@ -479,7 +482,12 @@ def test_unsupported_condition_exceptions(self):
             ("Using more than one indexing subspace is unsupported." + \
              " An indexing subspace is a group of one or more consecutive" + \
              " indices comprising integer or array types.",
-             (0, np.array([1, 2]), slice(None), 3, 4))
+             (0, np.array([1, 2]), slice(None), 3, 4)),
+            (("Only a single one-dimensional boolean array" + \
+              " index is permitted for each axis of the source array being indexed." + \
+              " NumPy behaviour of multi-dimensional boolean indices for " + \
+              "implicitly indexing multiple axes is not supported."),
+             (0, np.array([[0,1,0,1,0,1], [1,0,1,0,1,0]], dtype=bool), slice(None), 3,))
```
stuartarchibald, line 490:
```suggestion
             (0, np.array([[0, 1, 0, 1, 0, 1], [1, 0, 1, 0, 1, 0]], dtype=bool), slice(None), 3,))
```

numba/tests/test_fancy_indexing.py
```diff
@@ -303,32 +305,43 @@ def np_new_axis_setitem(a, idx, item):
...
 class TestFancyIndexingMultiDim(MemoryLeakMixin, TestCase):
     # Every case has exactly one, one-dimensional array,
-    # otherwise it's not fancy indexing.
+    # Otherwise it's not fancy indexing
     shape = (5, 6, 7, 8, 9, 10)
     indexing_cases = [
         # Slices + Integers
         (slice(4, 5), 3, np.array([0, 1, 3, 4, 2]), 1),
         (3, np.array([0,1,3,4,2]), slice(None), slice(4)),
+        (3, np.array([[0,1,3,4,2], [0,1,2,3,2], [3,1,3,4,1]]),
```
stuartarchibald, line 314:
```suggestion
        (3, np.array([[0, 1, 3, 4, 2], [0, 1, 2, 3, 2], [3, 1, 3, 4, 1]]),
```

numba/tests/test_fancy_indexing.py
```diff
@@ -303,32 +305,43 @@ def np_new_axis_setitem(a, idx, item):
...
+         slice(None), slice(4)), # multi-dimensional
 
         # Ellipsis + Integers
         (Ellipsis, 1, np.array([0,1,3,4,2])),
         (np.array([0,1,3,4,2]), 3, Ellipsis),
+        (np.array([[0,1,3,4,2], [0,1,2,3,2], [3,1,3,4,1]]),
```
stuartarchibald, line 320:
```suggestion
        (np.array([[0, 1, 3, 4, 2], [0, 1, 2, 3, 2], [3, 1, 3, 4, 1]]),
```

numba/tests/test_fancy_indexing.py
```diff
@@ -303,32 +305,43 @@ def np_new_axis_setitem(a, idx, item):
...
+         3, Ellipsis), # multi-dimensional
 
         # Ellipsis + Slices + Integers
         (Ellipsis, 1, np.array([0,1,3,4,2]), 3, slice(1,5)),
         (np.array([0,1,3,4,2]), 3, Ellipsis, slice(1,5)),
+        (np.array([[0,1,3,4,2], [0,1,2,3,2], [3,1,3,4,1]]),
+         3, Ellipsis, slice(1,5)), # multi-dimensional
```
stuartarchibald, line 327:
```suggestion
        (np.array([[0, 1, 3, 4, 2], [0, 1, 2, 3, 2], [3, 1, 3, 4, 1]]),
         3, Ellipsis, slice(1, 5)), # multi-dimensional
```

numba/tests/test_fancy_indexing.py
```diff
@@ -303,32 +305,43 @@ def np_new_axis_setitem(a, idx, item):
...
          1),
         (3, np.array([True, False, True, False, True, False]),
          slice(None), slice(4)),
+
+        # Differently ordered arrays
+        # Ellipsis + Slices + Integers
+        (Ellipsis, 1, np.array([0,1,3,4,2], order='A'), 3, slice(1,5)),
+        (np.array([0,1,3,4,2], order='F'), 3, Ellipsis, slice(1,5)),
+        (np.array([[0,1,3,4,2], [0,1,2,3,2], [3,1,3,4,1]], order='A'),
+         3, Ellipsis, slice(1,5)), # multidimensional
```
stuartarchibald, line 341:
```suggestion
        (Ellipsis, 1, np.array([0, 1, 3, 4, 2], order='A'), 3, slice(1, 5)),
        (np.array([0, 1, 3, 4, 2], order='F'), 3, Ellipsis, slice(1, 5)),
        (np.array([[0, 1, 3, 4, 2], [0, 1, 2, 3, 2], [3, 1, 3, 4, 1]], order='A'),
         3, Ellipsis, slice(1, 5)), # multidimensional
```

stuartarchibald (commented):
Thanks for the updates @kc611, few minor things to resolve now else looks good. OOB I spoke with @gmarkall about the impact of the use of the NRT on the CUDA target. The conclusion is as follows:
1. It's probably best to concentrate on getting this and the next related PR (#8912) merged as the implementation needs to be figured out to start with and it's almost certainly easier to do this on the CPU with the NRT present.
2. Go back to e.g. the `release0.57` branch and see what sort of "fancy indexing" actually worked on CUDA, write tests for this.
3. Get the tests in 2. working on `main` once these PRs are merged. I suspect it'll only be the case where the index is a non-contiguous array that's a problem as the copy is currently needed as part of "flattening"  it. Once we have the tests I expect the options and limitations will become more clear.

gmarkall:
gpuci run tests

numba/np/arrayobj.py
```diff
@@ -745,15 +751,53 @@ class IntegerArrayIndexer(Indexer):
...
+        self.idx_shape = cgutils.unpack_tuple(builder, idxary.shape)
         self.size = size
-        assert idxty.ndim == 1
         self.ll_intp = self.context.get_value_type(types.intp)
 
+        if idxty.ndim > 1:
+            if not context.enable_nrt:
+                raise NotImplementedError("This type of indexing is not"
+                                          " currently supported for"
+                                          " given compiler target.")
```
stuartarchibald, line 763: Ideally this would be tested, but I'm inclined to leave this to a subsequent PR as I am relatively sure it's going to need to be moved.

numba/np/arrayobj.py
```diff
@@ -1633,8 +1672,13 @@ def fancy_setslice(context, builder, sig, args, index_types, indices):
     indexer = FancyIndexer(context, builder, aryty, ary,
                            index_types, indices)
     indexer.prepare()
+    src_type = None
 
     if isinstance(srcty, types.Buffer):
+        if not context.enable_nrt:
+            raise NotImplementedError("This type of indexing is not currently"
+                                      " supported for given compiler target.")
```
stuartarchibald, line 1680: As above, ideally this would be tested, but I'm inclined to leave this to a subsequent PR as I am relatively sure it's going to need to be moved.

numba/np/arrayobj.py
stuartarchibald: Resolving this can be postponed as the 3 points in https://github.com/numba/numba/pull/8491#pullrequestreview-1411169708 will cover it.

numba/np/arrayobj.py
```diff
@@ -1690,67 +1716,80 @@ def fancy_setslice(context, builder, sig, args, index_types, indices):
...
+        if srcty.is_contig:
+            flat_imp = flat_imp_nocopy
+        # otherwise, use copy version since we don't support
+        # reshaping non-contiguous arrays
+        else:
+            flat_imp = flat_imp_copy
 
-    # Cast to the destination dtype (cross-dtype slice assignment is allowed)
-    val = context.cast(builder, val, src_dtype, aryty.dtype)
+        flat_imp = njit(flat_imp)
```
stuartarchibald, line 1743: This needs moving to a module global in a subsequent PR.

numba/np/arrayobj.py
```diff
@@ -1690,67 +1716,80 @@ def fancy_setslice(context, builder, sig, args, index_types, indices):
...
 
-    src_cleanup()
+    indexer.end_loops()
 
-    return context.get_dummy_value()
+    if src_type == 'buffer':
+        for _indexer in indexer.indexers:
+            if isinstance(_indexer, IntegerArrayIndexer) \
+               and hasattr(_indexer, "idxary_instr"):
+                context.nrt.decref(builder, _indexer.idxty,
```
stuartarchibald, line 1790: Use of NRT needs assessing, defer to later PR.

numba/tests/test_fancy_indexing.py
```diff
@@ -303,32 +302,43 @@ def np_new_axis_setitem(a, idx, item):
...
          np.array([True, False, True, False, True, False, False]),
          1),
         (3, np.array([True, False, True, False, True, False]),
          slice(None), slice(4)),
+
+        # Differently ordered arrays
+        # Ellipsis + Slices + Integers
+        (Ellipsis, 1, np.array([0, 1, 3, 4, 2], order='A'), 3, slice(1, 5)),
+        (np.array([0, 1, 3, 4, 2], order='F'), 3, Ellipsis, slice(1, 5)),
+        (np.array([[0, 1, 3, 4, 2], [0, 1, 2, 3, 2], [3, 1, 3, 4, 1]], order='A'),
```
stuartarchibald, line 337: Setting `order='A'` in the NumPy constructor won't have the effect of it being `A` ordered within Numba's type system (assuming that was the intent?) Suggest deferring fixing this to a later PR.

stuartarchibald (approved):
Many thanks for all your efforts on this @kc611. It's great to see this feature implemented. There's a few outstanding issues that I've left as a "review", but these should be captured in a new issue for completion in subsequent work. Most of the outstanding items relate to use of the Numba runtime which is going to take careful assessment. As alluded to previously, this will manifest in having to write a number of tests for the CUDA target to ensure that no regressions are introduced. Thanks again for work on this challenging implementation and feature, patch is approved!

PR #8493, opened 2022-10-06
Removes `context.compile_internal` where easy

guilhermeleobas:
Removes context.compile_internal where easy

* np.ravel
* array.flatten
* array.sort
* round
* not in
* ~math.gcd~
* np.argsort


numba/np/arrayobj.py
```diff
@@ -6015,21 +6008,22 @@ def np_sort_impl(a):
...
 
 
-@lower_builtin("array.argsort", types.Array, types.StringLiteral)
-@lower_builtin(np.argsort, types.Array, types.StringLiteral)
-def array_argsort(context, builder, sig, args):
-    arytype, kind = sig.args
+@overload(np.argsort)
+@overload_method(types.Array, "argsort")
+def impl_arr_argsort(arr, kind=None):
+    if kind is None or isinstance(kind, types.Omitted):
```
kc611, line 6014: You can also use `numba.np.numpy_support.is_nonelike` over here.

numba/np/arrayobj.py
```diff
@@ -5989,17 +5984,15 @@ def get_sort_func(kind, is_float, is_argsort=False):
         return func
```
kc611, line 5984: Unrelated to this PR but seems like there is no error being raised here if a key is supplied other than those listed in these branches ?

```python
import numpy as np
import numba

@numba.njit
def foo(x):
    return x.argsort(kind='somesort')

x = np.random.default_rng(1).integers(10, size=10)

print(foo(x))
# Error: func accessed before assignment.
```

numba/np/arrayobj.py
```diff
@@ -1992,46 +1992,41 @@ def impl(arr, values, axis=None):
...
-    def imp(ary):
-        return ary.copy().reshape(ary.size)
-
-    res = context.compile_internal(builder, imp, sig, args)
-    res = impl_ret_new_ref(context, builder, sig.return_type, res)
-    return res
+    if order == 'C' or isinstance(order, types.Omitted):
+        def imp(arr, order='C'):
+            return arr.copy().reshape(arr.size)
+        return imp
```
kc611, line 2029: Would be better if we raised an Error here instead of simply letting the default Not-implemented error take over, either that or simply don't have the order argument cause it seems unnecessary otherwise, since we won't support anything else.

This particular condition fails if user tries to pass 'C' itself as order.

kc611 (commented):
Looks Good to me!

Few minor nitpicks below.

guilhermeleobas:
Hi @kc611, thanks for the review. I have addressed all your comments.

numba/cpython/builtins.py
```diff
@@ -401,13 +403,11 @@ def next_impl(context, builder, sig, args):
...
-def not_in(context, builder, sig, args):
-    def in_impl(a, b):
+@overload("not in")
+def impl_not_in(a, b):
+    def impl(a, b):
         return operator.contains(b, a)
-
-    res = context.compile_internal(builder, in_impl, sig, args)
-    return builder.not_(res)
+    return impl
```
stuartarchibald, line 410: Not sure that this is correct.
1. String `"not in"` looks like a workaround for something that should now be done differently. Am a little surprised that `@overload` tolerates this.
2. The original implementation has a negated result which is not reflected in these changes, is that intentional?

Further, I'm not convinced that this code is even reachable any more. The interpreter directly handles the `CONTAINS` bytecode and inverts the condition based on the op case flag:
https://github.com/numba/numba/blob/deeeaa273c7b2645f83d14f092eb124a19bfe40c/numba/core/interpreter.py#L2584-L2591

numba/np/arrayobj.py
```diff
@@ -6015,21 +6012,22 @@ def np_sort_impl(a):
...
 
-@lower_builtin("array.argsort", types.Array, types.StringLiteral)
-@lower_builtin(np.argsort, types.Array, types.StringLiteral)
-def array_argsort(context, builder, sig, args):
-    arytype, kind = sig.args
+@overload(np.argsort)
+@overload_method(types.Array, "argsort")
+def impl_arr_argsort(arr, kind=None):
+    if is_nonelike(kind):
+        kind = types.StringLiteral(value='quicksort')
```
stuartarchibald, line 6019:
```suggestion
        kind = types.literal('quicksort')
```

stuartarchibald (commented):
Thanks for doing this refactoring @guilhermeleobas. A couple of comments to resolve else looks good.

stuartarchibald:
@gmarkall same comment as in https://github.com/numba/numba/pull/8494#issuecomment-1274527541, IIRC the CUDA target has its own `round` but maybe borrows the typing information from the CPU? I think the rest of the functions that are updated in this PR are either dead (`"not in"`) or not currently supported by the CUDA target.

stuartarchibald:
@guilhermeleobas I think https://github.com/numba/numba/pull/8493#issuecomment-1274535474 is going to need resolving before this can be merged. The test suite is likely to fail on the CUDA target for the reasons noted.

gmarkall:
gpuci run tests

numba/cuda/extending.py
```diff
@@ -2,6 +2,9 @@
 Added for symmetry with the core API
 """
 
+from functools import partial
 from numba.core.extending import intrinsic as _intrinsic
+from numba.core.extending import overload as _overload
 
 intrinsic = _intrinsic(target='cuda')
```
gmarkall, line 9: Should I have used `functools.partial` for `intrinsic` as well?

numba/cuda/cudaimpl.py
```diff
@@ -627,27 +628,43 @@ def ptx_min_f8(context, builder, sig, args):
...
+def round_to_impl(x):
+    if isinstance(x, types.Float):
+        def impl(x):
+            return ptx_round(x)
+        return impl
+
+
+@overload(round)
+def round_to_impl2(x, ndigits):
+    # XXX: it doesn't look like the CUDA target supports ndigits=None
```
gmarkall, line 664: Can it support `ndigits=None` with something like:
```python
    if isinstance(ndigits, NoneType):
        def impl(x):
            return ptx_round(x)
        return impl
```
?

gmarkall (commented):
The CUDA changes look generally good and gpuCI seems to be fine with it. There are a couple of thoughts / questions on the diff.

numba/cuda/extending.py
stuartarchibald: Is it worth doing `functools.update_wrapper` on these so as to inherit the docstrings etc?

numba/cuda/cudaimpl.py
stuartarchibald: I think there's at least two problems going on here, one of which is #8528, the other is what is causing this lowering error. It seems like non-matching default kwargs might be part of the cause, it's also strange that the overload validation step doesn't catch this.

stuartarchibald: This is also involved: #8530

gmarkall:
gpuci run tests

gmarkall (dismissed):
Thanks for the update - this looks good to me from a CUDA perspective now.

gmarkall:
(Note that the gpuCI fail is just an issue inherited from `main`, not this PR)

gmarkall:
gpuci run tests

(just to make everything look nice and green now #8537 is merged)

stuartarchibald:
@guilhermeleobas please could you resolve the conflicts against `main`? Many thanks.

guilhermeleobas:
I'm seeing some failures in this PR after migrating np.ravel to `@overload`:
https://gist.github.com/guilhermeleobas/91fc7a4137b4fa27b95ff4b8830d0952

stuartarchibald:
> I'm seeing some failures in this PR after migrating np.ravel to `@overload`: https://gist.github.com/guilhermeleobas/91fc7a4137b4fa27b95ff4b8830d0952

Public CI is passing, is this still an issue or was it resolved by the latest two commits (or just missed by public CI)?

guilhermeleobas:
> > I'm seeing some failures in this PR after migrating np.ravel to `@overload`: https://gist.github.com/guilhermeleobas/91fc7a4137b4fa27b95ff4b8830d0952
> 
> Public CI is passing, is this still an issue or was it resolved by the latest two commits (or just missed by public CI)?

This is still an "issue".

gmarkall:
This message in the traceback: `Only accept returning of array passed into the function as argument` I see popping up when NRT is not available (usually in the CUDA target, but these messages indicate `nopython` mode) - is there something disabling NRT going on there?

stuartarchibald:
> This message in the traceback: `Only accept returning of array passed into the function as argument` I see popping up when NRT is not available (usually in the CUDA target, but these messages indicate `nopython` mode) - is there something disabling NRT going on there?

I think this is indeed the root cause of the problem. In the test this is present:
https://github.com/numba/numba/blob/036d9561dc4188222febadb3fcb621519e241990/numba/tests/test_array_manipulation.py#L547
which is set via this:
https://github.com/numba/numba/blob/036d9561dc4188222febadb3fcb621519e241990/numba/tests/test_array_manipulation.py#L15-L16

The `Flags()` ctor won't enable the NRT by default.

This problem exists on `main` too, it's just far less "noisy" as the split typing/lowering API doesn't have such tracebacks. I think the test needs updating to run with the NRT enabled.

PR #8494, opened 2022-10-07
Remove `context.compile_internal` where easy on `numba/cpython/cmathimpl.py`

guilhermeleobas:
Note: Some functions were not converted because ufunc_db uses them.

List of functions converted:
- ~cmath.exp~
- ~cmath.sqrt~
- cmath.acosh
- cmath.tanh
- cmath.sinh
- ~cmath.sin~
- cmath.cosh
- ~cmath.cos~
- cmath.polar
- cmath.phase
- ~cmath.log~
- cmath.log10
- cmath.rect

I tried to keep the diff as small as possible and made one commit for each function migrated to `@overload`


numba/cpython/cmathimpl.py
```diff
@@ -183,21 +180,31 @@ def log10_impl(z):
...
+    return impl
+
+
+@overload(cmath.polar)
+def polar_impl(x):
+    if not isinstance(x, types.Complex):
+        return
+
+    def impl(x):
+        r, i = x.real, x.imag
```
kc611, line 205: Just confirming, previously these were assigned using the `intrinsic_complex_unary` decorator, and now it's being done during compile time.

numba/cpython/cmathimpl.py
```diff
@@ -53,39 +54,32 @@ def isfinite_float_impl(context, builder, sig, args):
...
-            # ditto
-            imag /= r
-        else:
-            imag *= r
-        return complex(real, imag)
+@overload(cmath.rect)
+def impl_cmath_rect(r, phi):
+    if all([isinstance(typ, types.Float) for typ in [r, phi]]):
+        def impl(r, phi):
+            if not math.isfinite(phi):
```
kc611, line 61: Are we sure this works in every case ? 

(Referring to comment ` # We can't call math.isfinite() inside rect() below because it only exists on 3.2+.`, if that is still valid. )

kc611 (approved):
This one LGTM!

Just a few clarifications to be done below.

numba/cpython/cmathimpl.py
guilhermeleobas: From the docs, it says that Numba requires Python 3.7 or newer. See:
https://numba.readthedocs.io/en/stable/user/installing.html#compatibility

numba/cpython/cmathimpl.py
guilhermeleobas: This is fine. `intrinsic_complex_unary` unpacks the complex value and call the decorated function with both `real` and `imag` values as well as if they're finite or not.

stuartarchibald:
@gmarkall Am wondering if loss of the typing definitions for the `cmath` functions is going to cause problems for the CUDA target? This either due to there being no registration against which a CUDA lowering can be associated OR if CUDA is relying on using the "old style" implementation injected directly into its registries there's now no reachable implementation as these `@overload` conversions default to targeting the CPU.

gmarkall:
gpuci run tests

gmarkall:
It could cause problems - I'm not sure if the `cmath` module has tests in the CUDA test suite or if it's just "assumed" to work by reusing the CPU target's implementation, so I don't know whether the gpuCI run will fail. If it doesn't fail and there are no explicit tests in CUDA, then I think a test for each modified function will need to be added to the CUDA tests, otherwise we might find that our tests pass but we break code in the wild. The list of supported `cmath` functions in CUDA is at https://numba.readthedocs.io/en/stable/cuda/cudapysupported.html#cmath which could be a useful reference for checking against.

stuartarchibald:
> @gmarkall, I've re-enabled cmath on CUDA by basically overloading the same function from `numba/cpython/cmathimpl` for the CUDA target.

I think the target hierarchy should be used to allow the sharing of implementations if they are literally the same functions. For example, if the implementations of the `cmath` functions in question are truly "generic" then they should have the target set as `"generic"` and both the CPU and CUDA targets can use them. If the CUDA target then needs to specialize some `cmath` functions further via `libdevice` it can do so by having an overload with `target="CUDA"`. If this doesn't work, I think it should, and it probably means there are bug(s)!

gmarkall:
gpuci run tests

stuartarchibald:
Note: #8554 will make specification of the `@overload` target as `'generic'` unnecessary as it makes it the default.

numba/cpython/cmathimpl.py
```diff
@@ -53,39 +54,32 @@ def isfinite_float_impl(context, builder, sig, args):
...
-            real /= r
-        else:
-            real *= r
-        if imag == 0. and math.isinf(r):
-            # ditto
-            imag /= r
-        else:
-            imag *= r
-        return complex(real, imag)
+@overload(cmath.rect, target='generic')
```
stuartarchibald, line 57:
```suggestion
@overload(cmath.rect)
```

stuartarchibald (commented):
Some suggested changes WRT #8554 which makes the `'generic'` target the default.

numba/cpython/cmathimpl.py
```diff
@@ -183,21 +180,31 @@ def log10_impl(z):
...
-@intrinsic_complex_unary
-def polar_impl(x, y, x_is_finite, y_is_finite):
-    """cmath.polar(x + y j)"""
-    return math.hypot(x, y), math.atan2(y, x)
+    if not isinstance(x, types.Complex):
+        return
+
+    def impl(x):
+        r, i = x.real, x.imag
+        return math.atan2(i, r)
```
stuartarchibald, line 195:
```suggestion
        r, i = x.real, x.imag
        return math.atan2(x.imag, x.real)
```
is the unpack needed?

stuartarchibald (commented):
One minor thing to resolve (along with the previous suggestions RE removing the `target='generic'`) else looks good, thanks @guilhermeleobas.

stuartarchibald (approved):

PR #8495, opened 2022-10-07
Removes context.compile_internal where easy on `numba/cpython/listobj.py`

guilhermeleobas:
- list.copy
- list.reverse
- list.count
- list.index
- list.remove
- operator.ne
- operator.lt
- operator.le
- operator.gt
- operator.ge

### DIFF
--- a/numba/cpython/listobj.py
+++ b/numba/cpython/listobj.py
@@ -888,8 +888,9 @@ def list_copy(lst):
 
     return list_copy_impl
 
-@lower_builtin("list.count", types.List, types.Any)
-def list_count(context, builder, sig, args):
+
+@overload_method(types.List, "count")
+def list_count(lst, value):
 
     def list_count_impl(lst, value):
         res = 0
@@ -898,7 +899,8 @@ def list_count(context, builder, sig, args):
                 res += 1
         return res
 
-    return context.compile_internal(builder, list_count_impl, sig, args)
+    return list_count_impl
+
 
 def _list_extend_list(context, builder, sig, args):
     src = ListInstance(context, builder, sig.args[1], args[1])
--- a/numba/cpython/listobj.py
+++ b/numba/cpython/listobj.py
@@ -5,7 +5,6 @@ Support for native homogeneous lists.
 
 import math
 import operator
-import numpy as np
 
 from llvmlite import ir
 from numba.core import types, typing, errors, cgutils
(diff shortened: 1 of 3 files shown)


numba/cpython/listobj.py
```diff
@@ -794,16 +795,20 @@ def list_eq(context, builder, sig, args):
 
     return builder.load(res)
 
-@lower_builtin(operator.ne, types.List, types.List)
-def list_ne(context, builder, sig, args):
+@overload(operator.ne)
+def impl_list_ne(a, b):
+    if not all([isinstance(typ, types.List) for typ in [a, b]]):
+        return
```
stuartarchibald, line 801: Perhaps factor this out into a function `all_lists(*args)` or similar?

numba/cpython/listobj.py
```diff
@@ -921,57 +935,49 @@ def list_extend(lst, iterable):
...
-@lower_builtin("list.index", types.List, types.Any,
-           types.Integer)
-def list_index(context, builder, sig, args):
 
-    def list_index_impl(lst, value, start):
-        n = len(lst)
-        if start < 0:
-            start += n
+@overload_method(types.List, "index")
+def list_index(lst, value, start=None, end=None):
```
stuartarchibald, line 940: This isn't the signature for `list.index`? It's, e.g.:
```pycon
In [1]: list.index?
Signature: list.index(self, value, start=0, stop=9223372036854775807, /)
```
I suspect the `stop` is `max(intp)`.

stuartarchibald (commented):
Thanks for the patch @guilhermeleobas. The refactor in general looks good, but I think some more work needs doing on `list.index` to match the CPython implementation (as noted inline). Thanks again!

numba/cpython/listobj.py
guilhermeleobas: Done!

numba/cpython/listobj.py
```diff
@@ -921,57 +940,35 @@ def list_extend(lst, iterable):
...
-@lower_builtin("list.index", types.List, types.Any)
-def list_index(context, builder, sig, args):
 
-    def list_index_impl(lst, value):
-        for i in range(len(lst)):
-            if lst[i] == value:
-                return i
-        # XXX references are leaked when raising
-        raise ValueError("value not in list")
+intp_max = np.iinfo(as_dtype(types.intp)).max
```
stuartarchibald, line 944: If possible, the `cpython` implementations should not rely on or refer to NumPy. In this case I think the `intp_max` can be found via:
```
>>> from numba import types
>>> types.intp.maxval
9223372036854775807
```

numba/cpython/listobj.py
```diff
@@ -921,57 +940,35 @@ def list_extend(lst, iterable):
...
 
-    return context.compile_internal(builder, list_index_impl, sig, args)
 
-@lower_builtin("list.index", types.List, types.Any,
-           types.Integer)
-def list_index(context, builder, sig, args):
+@overload_method(types.List, "index")
+def list_index(lst, value, start=0, end=intp_max):
```
stuartarchibald, line 948:
```
In [1]: list.index?
Signature: list.index(self, value, start=0, stop=9223372036854775807, /)
```
The last kwarg is named `stop` not `end`?

stuartarchibald (commented):
Thanks for the updates @guilhermeleobas, couple of minor things to look at else looks good.

numba/cpython/listobj.py
```diff
@@ -17,6 +17,7 @@
 from numba.misc import quicksort
 from numba.cpython import slicing
 from numba import literal_unroll
+from numba.np.numpy_support import as_dtype
```
stuartarchibald, line 20:
```suggestion
```
dead code?

numba/cpython/listobj.py
```diff
@@ -921,41 +938,19 @@ def list_extend(lst, iterable):
...
-                start = 0
-        for i in range(start, len(lst)):
-            if lst[i] == value:
-                return i
-        # XXX references are leaked when raising
-        raise ValueError("value not in list")
+    if not isinstance(start, (int, types.Integer, types.Omitted)):
+        raise TypeError(f'arg "start" must be an Integer. Got {start}')
+    if not isinstance(stop, (int, types.Integer, types.Omitted)):
+        raise TypeError(f'arg "end" must be an Integer. Got {end}')
```
stuartarchibald, line 951: Just noticed that this refers to `end`, which is undefined. That it's not been caught by testing suggests that it is untested. Perhaps add a test to hit these two branches?

stuartarchibald (commented):
Thanks for the updates, I just gave this a final check and noticed that there's a potential bug in the error reporting (noted inline). Once fixed, should be good to merge. Thanks again.

PR #8496, opened 2022-10-09
Rewrite most of the set API to use overloads

guilhermeleobas:
Methods changed:
- set.copy
- set.clear
- set.discard
- set.pop
- set.remove
- set.difference_update
- set.intersection_update
- set.symmetric_difference_update
- set.isdisjoint
- set.issubset
- set.issuperset
- set.union
- set.symetryc_difference
- set.intersection
- set.difference
- operator.sub
- operator.and_
- operator.xor
- operator.or_
- operator.iand
- operator.ior
- operator.isub
- operator.ixor
- operator.le
- operator.gt
- operator.lt
- operator.ge
- operator.ne
- operator.eq

### DIFF
--- a/numba/core/typing/setdecl.py
+++ b/numba/core/typing/setdecl.py
@@ -59,14 +59,6 @@ class SetAttribute(AttributeTemplate):
             sig = sig.replace(recvr=set.copy(dtype=unified))
             return sig
 
-    def _resolve_xxx_update(self, set, args, kws):
-        assert not kws
-        iterable, = args
-        # Set arguments only supported for now
-        # (note we can mix non-reflected and reflected arguments)
-        if isinstance(iterable, types.Set) and iterable.dtype == set.dtype:
-            return signature(types.none, iterable)
-
     def _resolve_operator(self, set, args, kws):
         assert not kws
         iterable, = args
--- a/numba/core/typing/setdecl.py
+++ b/numba/core/typing/setdecl.py
@@ -131,10 +131,6 @@ class SetAttribute(AttributeTemplate):
     def resolve_issubset(self, set, args, kws):
         return self._resolve_comparator(set, args, kws)
 
-    @bound_function("set.issuperset")
-    def resolve_issuperset(self, set, args, kws):
-        return self._resolve_comparator(set, args, kws)
-
 
 class SetOperator(AbstractTemplate):
 
@@ -169,7 +165,7 @@ for op_key in (operator.iadd, operator.isub, operator.iand, operator.ior, operat
         key = op_key
 
 
-for op_key in (operator.eq, operator.ne, operator.lt, operator.le, operator.ge, operator.gt):
+for op_key in (operator.eq, operator.ne, operator.lt, operator.le, operator.gt):
     @infer_global(op_key)
     class ConcreteSetComparison(SetComparison):
         key = op_key
--- a/numba/core/typing/setdecl.py
+++ b/numba/core/typing/setdecl.py
@@ -117,10 +117,6 @@ class SetAttribute(AttributeTemplate):
         if isinstance(iterable, types.Set) and iterable.dtype == set.dtype:
             return signature(set, iterable)
 
-    @bound_function("set.union")
-    def resolve_union(self, set, args, kws):
-        return self._resolve_operator(set, args, kws)
-
     def _resolve_comparator(self, set, args, kws):
         assert not kws
         arg, = args
@@ -161,7 +157,7 @@ class SetComparison(AbstractTemplate):
             return signature(types.boolean, *args)
 
 
-for op_key in (operator.add, operator.or_, operator.invert):
+for op_key in (operator.add, operator.invert):
     @infer_global(op_key)
     class ConcreteSetOperator(SetOperator):
         key = op_key
--- a/numba/core/typing/setdecl.py
+++ b/numba/core/typing/setdecl.py
@@ -165,7 +165,7 @@ for op_key in (operator.iadd, operator.isub, operator.iand, operator.ior, operat
         key = op_key
 
 
-for op_key in (operator.le, operator.gt):
+for op_key in (operator.le,):
     @infer_global(op_key)
     class ConcreteSetComparison(SetComparison):
         key = op_key
--- a/numba/core/typing/setdecl.py
+++ b/numba/core/typing/setdecl.py
@@ -45,12 +45,6 @@ class SetAttribute(AttributeTemplate):
             sig = sig.replace(recvr=set.copy(dtype=unified))
             return sig
 
-    @bound_function("set.clear")
-    def resolve_clear(self, set, args, kws):
-        assert not kws
-        if not args:
-            return signature(types.none)
-
     @bound_function("set.copy")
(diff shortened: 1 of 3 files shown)


numba/cpython/setobj.py
```diff
@@ -1351,24 +1406,60 @@ def set_difference_update(context, builder, sig, args):
...
+@intrinsic
+def _set_difference_update(typingctx, a, b):
+    sig = types.none(a, b)
+    return sig, set_difference_update
+
+
+@overload_method(types.Set, "difference_update")
+def set_difference_update_impl(a, b):
+    if not all([isinstance(typ, types.Set) for typ in (a, b)]):
+        return
```
stuartarchibald, line 1419: Factor this out as function `all_sets(*args)` ?

numba/cpython/setobj.py
```diff
@@ -1298,15 +1298,26 @@ def set_add(context, builder, sig, args):
 
     return context.get_dummy_value()
 
-@lower_builtin("set.discard", types.Set, types.Any)
+
 def set_discard(context, builder, sig, args):
```
stuartarchibald, line 1302: Move this into the intrinsic `_set_discard` as the `codegen` (and same comment for all the occurrences of this pattern)? i.e.
```python
@intrinsic
def _set_<something>(typingctx, <args>):
    sig = <>
    def codegen(cgctx, builder, sig, args): # This is the `set_<something>` function at present.
        ...
    return sig, codegen
```

numba/cpython/setobj.py
```diff
@@ -1328,21 +1350,54 @@ def set_remove(context, builder, sig, args):
...
 
+@intrinsic
+def _set_remove(typingctx, s, item):
+    sig = types.none(s, item)
+    return sig, set_remove
+
+
+@overload_method(types.Set, "remove")
+def ol_set_remove(s, item):
+    # XXX: should item have the same type as s.dtype?
```
stuartarchibald, line 1361: I think yes given the current implementation. e.g.:
```python
from numba import njit
import numpy as np

@njit
def foo():
    x = {1, 2, 3, 4}
    x.remove('a')

foo()
```
ends up with:
```
TypeError: Type of #3 arg mismatch: i64 != {i8*, i64, i32, i32, i64, i8*, i8*}
```
As the reflected set implementation is pending deprecation I don't think there's any need to add support for this, but the issue in lowering can be prevented at type inference time so it should be. Suggest fixing the typing and adding a test.

numba/core/typing/setdecl.py
```diff
@@ -97,18 +67,6 @@ def _resolve_xxx_update(self, set, args, kws):
         if isinstance(iterable, types.Set) and iterable.dtype == set.dtype:
             return signature(types.none, iterable)
 
-    @bound_function("set.difference_update")
-    def resolve_difference_update(self, set, args, kws):
-        return self._resolve_xxx_update(set, args, kws)
```
stuartarchibald, line 102: Is `self._resolve_xxx_update` now dead code?

numba/cpython/setobj.py
stuartarchibald: Also, the original `_resolve_xxx_update` method requires that the `dtype` of the two sets is the same:
https://github.com/numba/numba/blob/4bc04ceb28a3bce0461e23db16b8826745e96aef/numba/core/typing/setdecl.py#L92-L98

numba/cpython/setobj.py
```diff
@@ -1399,34 +1490,53 @@ def set_update(context, builder, sig, args):
...
 
 # Set operations creating a new set
 
-@lower_builtin(operator.sub, types.Set, types.Set)
-@lower_builtin("set.difference", types.Set, types.Set)
-def set_difference(context, builder, sig, args):
+@overload(operator.sub)
+@overload_method(types.Set, "difference")
+def impl_set_difference(a, b):
+    if not all([isinstance(typ, types.Set) for typ in (a, b)]):
```
stuartarchibald, line 1524: Think these also need `dtype` checks. Suggest checking all of these to make sure the `dtype` check is present.

stuartarchibald (commented):
Thanks for the patch. This refactoring seems broadly correct on inspection. There's a few minor things to resolve particularly in relation to ensuring `dtype` checks are made, once resolved should be good to merge.

Note: There should be no impact on the CUDA target as `set` isn't supported due to the use of the NRT.

Thanks again!

numba/core/typing/setdecl.py
guilhermeleobas: Yes, it is! I'll remove it

numba/cpython/setobj.py
guilhermeleobas: Ok! I kept this way to reduce the patch size and to make it easier to review.

numba/cpython/setobj.py
stuartarchibald: Thanks for making the change, I think it would be good to add a test for it.

guilhermeleobas: Oops, sorry. I forgot to include a test for it

numba/cpython/setobj.py
```diff
@@ -103,6 +103,11 @@ def is_hash_used(context, builder, h):
     return builder.icmp_unsigned('<', h, deleted)
 
 
+def all_set(*args):
+    return all([isinstance(typ, types.Set) for typ in args]) and \
+        all([args[0].dtype == s.dtype for s in args])
```
stuartarchibald, line 108: I just ran the unit tests for this and found that whilst a mixed dtype operation is correctly rejected, the error message does not describe the problem very well, it just reports `No implementation of function...<some function>... found for signature...` with no indication as to the actual problem. Maybe this function should:
* raise something like `TypingError(f"All arguments must be Sets, got {args}")` if the types are not all `Set`s.
* raise  something like `TypingError(f"All Sets must be of the same type, got {args}")` if the dtypes are not all the same.

What do you think?

stuartarchibald: Thanks for adding the differentiating error messages in 5510cb6, any chance you could also please add test(s) to demonstrate that each path is correctly taken, i.e. check the correct exceptions are raised for the known erroneous input? Many thanks.

guilhermeleobas: Done!

numba/tests/test_sets.py
```diff
@@ -171,36 +171,25 @@ def union_usecase(a, b):
...
-    """ % dict(op=op)
-    return compile_function('inplace_operator_usecase', code, globals())
-
-def make_comparison_usecase(op):
-    code = """if 1:
-    def comparison_usecase(a, b):
-        return set(a) %(op)s set(b)
-    """ % dict(op=op)
-    return compile_function('comparison_usecase', code, globals())
+def union_usecase__2(a, b):
```
stuartarchibald, line 186: These `__2` function names don't really note their intended use. Given they are used only in the unit tests in `TestSetsInvalid` perhaps just define them locally in those unit tests?

stuartarchibald (commented):
Thanks for the updates @guilhermeleobas, couple of minor things to look at else looks good.

stuartarchibald (approved):

PR #8498, opened 2022-10-11
Remove _DispatcherBase._types_active_call

gmarkall:
Keeping a list of the argument types for the current call appears to be pointless - even if it weren't pointless, I'm not convinced it wouldn't be a source of race conditions when an njitted function is called by multiple threads.


gmarkall:
OK, I see it's not pointless - it stops the things it holds references to getting GCed.

PR #8499, opened 2022-10-11
Deprecate numba.generated_jit

### LINKED ISSUE #8466: Consider deprecating `numba.generated_jit`
The decorator `@numba.generated_jit` provides similar functionality to the much more powerful `@numba.extending.overload` decorator. This issue discusses issues with the `@generated_jit` functionality and is to track decisions, if any, over deprecation in favour of just using `@overload`.

The `@generated_jit` decorator essentially provides functionality to allow type-based selection of an implementation at compile time. Example:

```python
from numba import njit, generated_jit, types

@generated_jit
def select(x):
    if isinstance(x, types.Float):
        def impl(x):
            return x + 1
        return impl
    elif isinstance(x, types.UnicodeType):
        def impl(x):
            return x + " the number one"
        return impl
    else:
        raise TypeError("Unsupported Type")

@njit
def foo(x):
    return select(x)

print(foo(1.))
print(foo("a string"))
```
A few issues exist with the concept/implementation:
1. `generated_jit` use tends to break "JIT transparency", i.e. if the `@jit` family decorators are removed or the environment variable `NUMBA_DISABLE_JIT` is set, then the source code does not execute in the same way as it would were the JIT compiler present (xref: #4759).
2. Various flags etc are not supported by `@generated_jit`, for example inlining (xref: #4692).
3. I've not proved this as yet, but would guess based on the implementation, that the target specific overload functionality would probably not work as expected.

Conceptually, `@generated_jit` is like `@overload` but the overloaded function is the decorated function, as a result, there already exists a replacement for `@generated_jit`, it's `@overload`. Taking the example above and turning it into an `@overload`:
```python
from numba import njit, types

def select(x):
    if isinstance(x, float):
        return x + 1
    elif isinstance(x, str):
        return x + " the number one"
    else:
        raise TypeError("Unsupported Type")

@overload(select)
def ol_select(x):
    if isinstance(x, types.Float):
        def impl(x):
            return x + 1
        return impl
    elif isinstance(x, types.UnicodeType):
        def impl(x):
            return x + " the number one"
        return impl
    else:
        raise TypeError("Unsupported Type")

@njit
def foo(x):
    return select(x)

print(foo(1.))
print(foo("a string"))
```

Further, since preliminary support for `isinstance` has been added, along with advances in branch pruning, the necessity for `@generated_jit` in basic cases like the above has decreased, for example, this is a shorter version of the above.

```python
from numba import njit, types

def select(x):
    if isinstance(x, float):
        return x + 1
    elif isinstance(x, str):
        return x + " the number one"
    else:
        raise TypeError("Unsupported Type")

overload(select)(lambda x: select)

@njit
def foo(x):
    return select(x)

print(foo(1.))
print(foo("a string"))
```

Final points...
* The CUDA target (and any other target implemented through the target specific functionality) now supports `@overload` and `@intrinsic` and will inherit all the effort that has been put into those decorators, trying to make `@generated_jit` work seems unnecessary given the replacement already exists.
* From a maintenance perspective, reducing the public facing API reduces maintenance burden, however, `@generated_jit` is not particularly involved in part because it is not as heavily developed as `@overload`.

Considerations:
1. Is there any functionality that `@generated_jit` provides that `@overload` does not?
2. Does `@overload` perhaps need an update to assist with 1. ?
3. How much of an impact would deprecation have against existing code bases?
    1. Need to find prevalence of use.
    2. Need to work out how hard the translation would be/should a wrapper through `@overload` be provided?
5. If deprecated, how long a cycle is required?

stuartarchibald:
This deprecates numba.generated_jit as outlined in #8466

Closes #8466

### DIFF
--- a/docs/source/reference/deprecation.rst
+++ b/docs/source/reference/deprecation.rst
@@ -224,7 +224,146 @@ supply the keyword argument ``forceobj=True`` to ensure the function is always
 compiled in :term:`object mode`.
 
 
-.. _deprecation-strict-strides:
+.. _deprecation-of-generated-jit:
+
+Deprecation of ``generated_jit``
+================================
+The top level API function ``numba.generated_jit`` provides functionality that
+allows users to write JIT compilable functions that have different
+implementations based on the types of the arguments to the function. This is a
+hugely useful concept and is also key to Numba's internal implementation.
+
+Reason for deprecation
+----------------------
+
+There are a number of reasons for this deprecation.
+
+First, ``generated_jit`` breaks the concept of "JIT transparency" in that if the
+JIT compiler is disabled, the source code does not execute the same way as it
+would were the JIT compiler present.
+
+Second, internally Numba uses the ``numba.extending.overload`` family of
+decorators to access an equivalent functionality to ``generated_jit``. The
+``overload`` family of decorators are more powerful than ``generated_jit`` as
+they support far more options and both the CPU and CUDA targets. Essentially a
+replacement for ``generated_jit`` already exists and has been recommended and
+preferred for a long while.
+
+Third, the public extension API decorators are far better maintained than
+``generated_jit``. This is an important consideration due to Numba's limited
+resources, fewer duplicated pieces of functionality to maintain will reduce
+pressure on these resources.
+
+For more information on the ``overload`` family of decorators see the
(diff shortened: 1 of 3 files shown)


sklam (approved):
Thanks for the patch!

PR #8513, opened 2022-10-14
Added support for numpy.newaxis

kc611:
As titled, this PR adds support for `numpy.newaxis` (which is simply an alias for `None`) for indexing NumPy arrays in Numba JIT functions.

```python
import numpy as np
import numba

@numba.njit
def foo(x):
    # Basic Indexing
    return x[:, np.newaxis]

@numba.njit
def foo_fancy(x):
    # Fancy Indexing
    return x[:, np.newaxis, np.array([1,2,1])]

x = np.random.randint(0, 10, (4,5,6))

print(foo(x).shape == foo.py_func(x).shape)
# True

print(foo_fancy(x).shape == foo_fancy.py_func(x).shape)
# True
```

### DIFF
--- a/docs/source/reference/numpysupported.rst
+++ b/docs/source/reference/numpysupported.rst
@@ -161,7 +161,7 @@ Array access
 ------------
 
 Arrays support normal iteration.  Full basic indexing and slicing is
-supported along with passing ``None``/ ``np.newaxis`` as indices for
+supported along with passing ``None`` / ``np.newaxis`` as indices for
 additional resulting dimensions. A subset of advanced indexing is also
 supported: only one advanced index is allowed, and it has to be a
 one-dimensional array (it can be combined with an arbitrary number
--- a/docs/source/reference/numpysupported.rst
+++ b/docs/source/reference/numpysupported.rst
@@ -161,9 +161,11 @@ Array access
 ------------
 
 Arrays support normal iteration.  Full basic indexing and slicing is
-supported.  A subset of advanced indexing is also supported: only one
-advanced index is allowed, and it has to be a one-dimensional array
-(it can be combined with an arbitrary number of basic indices as well).
+supported along with passing ``None``/ ``np.newaxis`` as indices for
+additional resulting dimensions. A subset of advanced indexing is also
+supported: only one advanced index is allowed, and it has to be a
+one-dimensional array (it can be combined with an arbitrary number
+of basic indices as well).
 
 .. seealso::
    `NumPy indexing <http://docs.scipy.org/doc/numpy/reference/arrays.indexing.html>`_
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -994,12 +994,21 @@ class FancyIndexer(object):
     def prepare(self):
         for i in self.indexers:
             i.prepare()
-        # Compute the resulting shape
-        shape_list = [i.get_shape() for i in self.indexers]
+
         one = self.context.get_constant(types.intp, 1)
+
+        # Compute the resulting shape given by the indices
+        res_shape = [i.get_shape() for i in self.indexers]
+
+        # At every position where newaxis/None is present insert
+        # one as a constant shape in the resulting list of shapes.
         for i in self.newaxes:
-            shape_list.insert(i, (one,))
-        self.indexers_shape = sum(shape_list, ())
+            res_shape.insert(i, (one,))
+
+        # Store the shape as a tuple, we can't do a simple
+        # tuple(res_shape) here since res_shape is a list
+        # of tuples which may be differently sized.
+        self.indexers_shape = sum(res_shape, ())
 
     def get_shape(self):
         """
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -351,6 +351,7 @@ def basic_indexing(context, builder, aryty, ary, index_types, indices,
     the corresponding view.
     """
     zero = context.get_constant(types.intp, 0)
+    one = context.get_constant(types.intp, 1)
 
     shapes = cgutils.unpack_tuple(builder, ary.shape, aryty.ndim)
     strides = cgutils.unpack_tuple(builder, ary.strides, aryty.ndim)
@@ -386,6 +387,10 @@ def basic_indexing(context, builder, aryty, ary, index_types, indices,
             if boundscheck:
                 cgutils.do_boundscheck(context, builder, ind, shapes[ax], ax)
             output_indices.append(ind)
+        elif is_nonelike(idxty):
+            output_shapes.append(one)
+            output_strides.append(zero)
+            ax -= 1
         else:
             raise NotImplementedError("unexpected index type: %s" % (idxty,))
         ax += 1
@@ -926,6 +931,7 @@ class FancyIndexer(object):
         self.shapes = cgutils.unpack_tuple(builder, ary.shape, aryty.ndim)
         self.strides = cgutils.unpack_tuple(builder, ary.strides, aryty.ndim)
         self.ll_intp = self.context.get_value_type(types.intp)
+        self.newaxes = []
 
         indexers = []
 
@@ -963,6 +969,9 @@ class FancyIndexer(object):
                 else:
                     assert 0
                 indexers.append(indexer)
+            elif is_nonelike(idxty):
+                self.newaxes.append(ax)
+                ax -= 1
             else:
                 raise AssertionError("unexpected index type: %s" % (idxty,))
             ax += 1
@@ -981,7 +990,11 @@ class FancyIndexer(object):
         for i in self.indexers:
             i.prepare()
         # Compute the resulting shape
-        self.indexers_shape = sum([i.get_shape() for i in self.indexers], ())
+        shape_list = [i.get_shape() for i in self.indexers]
+        one = self.context.get_constant(types.intp, 1)
+        for i in self.newaxes:
+            shape_list.insert(i, (one,))
+        self.indexers_shape = sum(shape_list, ())
 
     def get_shape(self):
         """
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -360,11 +360,12 @@ def basic_indexing(context, builder, aryty, ary, index_types, indices,
     output_shapes = []
     output_strides = []
 
+    num_newaxes = len([idx for idx in index_types if is_nonelike(idx)])
     ax = 0
     for indexval, idxty in zip(indices, index_types):
         if idxty is types.ellipsis:
             # Fill up missing dimensions at the middle
-            n_missing = aryty.ndim - len(indices) + 1
+            n_missing = aryty.ndim - len(indices) + 1 + num_newaxes
             for i in range(n_missing):
                 output_indices.append(zero)
                 output_shapes.append(shapes[ax])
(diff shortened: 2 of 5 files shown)


numba/tests/test_fancy_indexing.py
```diff
@@ -255,6 +255,19 @@ def check(arr, ind):
...
+            return a[idx, np.newaxis, :]
+        
+        arr = np.random.randint(0, 10, (4,5,6))
+        idx = np.array([1,2,1])
+        py_func = np_new_axis
+        cfunc = njit(py_func)
+
+        expected = py_func(arr, idx)
+        got = cfunc(arr, idx)
+        np.testing.assert_equal(expected, got)
```
stuartarchibald, line 270: Think it'd be a good idea to include more tests:
* Multiple `np.newaxis` in a single indexing call.
* `np.newaxis` with other involved indexes.
* `np.newaxis` in a `__setitem__` context.
* Use of `None` in place of `np.newaxis`.

numba/np/arrayobj.py
```diff
@@ -981,7 +990,11 @@ def prepare(self):
         for i in self.indexers:
             i.prepare()
         # Compute the resulting shape
-        self.indexers_shape = sum([i.get_shape() for i in self.indexers], ())
+        shape_list = [i.get_shape() for i in self.indexers]
+        one = self.context.get_constant(types.intp, 1)
+        for i in self.newaxes:
+            shape_list.insert(i, (one,))
+        self.indexers_shape = sum(shape_list, ())
```
stuartarchibald, line 997: Is this correct? I think this might trigger an issue present here:
```python
from numba import njit
import numpy as np

@njit
def foo(a):
    return a[np.newaxis, :, np.newaxis,  ..., np.newaxis]

def gen(n):
    return np.arange(np.prod(range(1, n + 1))).reshape((*range(n, 0, -1)))

x = gen(4)
got, expected = foo(x), foo.py_func(x)
np.testing.assert_allclose(got, expected)
```

stuartarchibald (commented):
Thanks for the patch. I've given this an initial review, I think that this makes most of the common uses for `np.newaxis` work but there's some edge cases that need looking at. Adding some more tests would probably be a good idea given how indexing is always an involved and nuanced topic! Thanks again!

numba/np/arrayobj.py
kc611: Resolved in [be5d0cc](https://github.com/numba/numba/pull/8513/commits/be5d0ccbbff8b33615fc6e2a6eb88c104fc5dbf2)

numba/tests/test_fancy_indexing.py
kc611: Resolved in [be5d0cc](https://github.com/numba/numba/pull/8513/commits/be5d0ccbbff8b33615fc6e2a6eb88c104fc5dbf2)

numba/tests/test_array_manipulation.py
```diff
@@ -633,23 +629,40 @@ def test_squeeze_array_npm(self):
...
+        def np_new_axis_getitem(a, idx):
+            return a[idx]
+
+        def np_new_axis_setitem(a, idx, item):
+            a[idx] = item
+            return a
+
+        a = np.arange(4*5*6*7).reshape((4, 5, 6, 7))
+        idx_cases = [
+            (np.newaxis, slice(None)),
```
stuartarchibald, line 642: Perhaps add some use cases where `np.newaxis` isn't the first item?

numba/tests/test_fancy_indexing.py
```diff
@@ -255,6 +255,41 @@ def check(arr, ind):
...
+        def np_new_axis_getitem(a, idx):
+            return a[idx]
+
+        def np_new_axis_setitem(a, idx, item):
+            a[idx] = item
+            return a
+
+        a = np.arange(4*5*6*7).reshape((4, 5, 6, 7))
+        idx_cases = [
+            (np.newaxis, slice(None)),
```
stuartarchibald, line 269: As above, perhaps add some cases where `np.newaxis` isn't the first item?

numba/np/arrayobj.py
```diff
@@ -981,7 +995,11 @@ def prepare(self):
         for i in self.indexers:
             i.prepare()
         # Compute the resulting shape
-        self.indexers_shape = sum([i.get_shape() for i in self.indexers], ())
+        shape_list = [i.get_shape() for i in self.indexers]
+        one = self.context.get_constant(types.intp, 1)
+        for i in self.newaxes:
+            shape_list.insert(i, (one,))
```
stuartarchibald, line 1001: This seems to be inserting at indexes based on a previous loop index into a list which has changing length. I guess this could work but it's quite hard to follow. This made me wonder if there's merit in implementing a `NewAxisIndexer` class as a way to handle the `newaxis` without there being a custom path and custom tracking/fixing of the shapes etc?

numba/tests/test_array_manipulation.py
```diff
@@ -633,23 +629,40 @@ def test_squeeze_array_npm(self):
...
-        a = np.arange(9).reshape(3, 3)
+    def test_add_axis(self):
+        def np_new_axis_getitem(a, idx):
+            return a[idx]
+
+        def np_new_axis_setitem(a, idx, item):
+            a[idx] = item
+            return a
+
+        a = np.arange(4*5*6*7).reshape((4, 5, 6, 7))
```
stuartarchibald, line 640:
```suggestion
        a = np.arange(4 * 5 * 6 * 7).reshape((4, 5, 6, 7))
```

stuartarchibald (commented):
Thanks for the updates @kc611, I've left a few comments inline to look at, the most important of which is RE a potential new `NewAxisIndexer` class as an alternative implementation method, would be interested in hearing your thoughts. Thanks again!

numba/np/arrayobj.py
kc611: I tried out this approach. The problem with it is, the way `Indexer`s built, they are supposed to provide an index for every axis in the 'source' array and not the resulting array. The number of `Indexer`s are supposed to correspond to the number of axes in the 'source'. (This is also represented by the following, which asserts that number of `Indexer` objects should always match the original dimensions of array.)

https://github.com/numba/numba/blob/bfe1270e23f8daf7fd95cfb6ef385f6c832fd927/numba/np/arrayobj.py#L1005

By adding a `np.newaxis`/`None`, we are adding a dimension in the resulting array and not the original one from where we're accessing the elements from. The original remains unchanged. 

So adding a `NewAxisIndexer`will leads to breaking of this `assert`s and requires special case handling in the index provision logic where we'll need to exclude the `NewAxisIndexer` objects since they don't have a existence in original array.

https://github.com/numba/numba/blob/bfe1270e23f8daf7fd95cfb6ef385f6c832fd927/numba/np/arrayobj.py#L1059

So it's possible, but will only add more complexity. Also we would want to take into consideration that currently the logic for both basic and advanced indexing is nearly the same. The above `Indexer`s will only apply to advanced indexing and not basic one. I think instead, we can somehow work on simplifying the current logic to make it less bug-prone/readable.

numba/tests/test_fancy_indexing.py
kc611: Addressed in 89ba549

numba/tests/test_array_manipulation.py
kc611: Addressed in 89ba549

numba/tests/test_fancy_indexing.py
```diff
@@ -255,6 +255,41 @@ def check(arr, ind):
...
+            (np.newaxis, slice(1), np.newaxis, np.array([1,2,1])),
+            (np.newaxis, slice(1), Ellipsis, np.newaxis, np.array([1,2,1])),
+            (np.newaxis, np.array([1,2,1]), np.newaxis, Ellipsis),
+            (np.newaxis, np.array([1,2,1]), Ellipsis, None),
+        ]
+        pyfunc_getitem = np_new_axis_getitem
+        cfunc_getitem = njit(pyfunc_getitem)
+
+        pyfunc_setitem = np_new_axis_setitem
+        cfunc_setitem = njit(pyfunc_setitem)
```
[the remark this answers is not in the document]
kc611: Addressed in 89ba549

numba/np/arrayobj.py
stuartarchibald: Ah ok, thanks for considering it. I would expect `assert`s to need updating if logic changes, so I am not convinced that is an issue in itself, however additional/increased complexity is a problem and something to avoid. I think perhaps making the current logic have a more clear intent seems like the best thing to do?

stuartarchibald (commented):
Additional thought... please could you add a note to the NumPy reference docs to state that `np.newaxis`/use of `None` as an axis is now supported? Many thanks!

docs/source/reference/numpysupported.rst
```diff
@@ -161,9 +161,11 @@ Array access
 ------------
 
 Arrays support normal iteration.  Full basic indexing and slicing is
-supported.  A subset of advanced indexing is also supported: only one
-advanced index is allowed, and it has to be a one-dimensional array
-(it can be combined with an arbitrary number of basic indices as well).
+supported along with passing ``None``/ ``np.newaxis`` as indices for
```
stuartarchibald, line 164:
```suggestion
supported along with passing ``None`` / ``np.newaxis`` as indices for
```

numba/np/arrayobj.py
```diff
@@ -1208,7 +1235,6 @@ def maybe_copy_source(context, builder, use_copy,
             builder.store(builder.load(src_ptr), dest_ptr)
 
     def src_getitem(source_indices):
-        assert len(source_indices) == srcty.ndim
```
stuartarchibald, line 1211: Should this be checking something like `len(source_indices) == srcty.ndim + len(self.newaxes)` or is this sort of invariant now entirely lost?

kc611: I removed it because during that function call we do not have the information regarding the number of `newaxes`/`None`s, since that is being unpacked/processed while calling the `Indexer`'s constructor which happens much later after this call. So at this specific line, yeah we can't do it unless we unpack the indices specifically for this `assert`.

stuartarchibald: Ah I see, thanks for the explanation. I think it's ok to leave this as-is then.

numba/tests/test_array_manipulation.py
```diff
@@ -633,23 +629,49 @@ def test_squeeze_array_npm(self):
...
+        for idx in idx_cases:
+            expected = pyfunc_getitem(a, idx)
+            got = cfunc_getitem(a, idx)
+            np.testing.assert_equal(expected, got)
+
+            a_empty = np.zeros_like(a)
+            item = a[idx]
+
+            expected = pyfunc_setitem(a_empty, idx, item)
+            got = cfunc_setitem(a_empty, idx, item)
```
stuartarchibald, line 673:
```suggestion
            expected = pyfunc_setitem(a_empty.copy(), idx, item)
            got = cfunc_setitem(a_empty.copy(), idx, item)
```
else they are both doing a `setitem` on the same array so the first could work and the second not and it wouldn't be possible to tell?

numba/tests/test_fancy_indexing.py
```diff
@@ -256,6 +256,50 @@ def check(arr, ind):
...
+        for idx in idx_cases:
+            expected = pyfunc_getitem(a, idx)
+            got = cfunc_getitem(a, idx)
+            np.testing.assert_equal(expected, got)
+
+            a_empty = np.zeros_like(a)
+            item = a[idx]
+
+            expected = pyfunc_setitem(a_empty, idx, item)
+            got = cfunc_setitem(a_empty, idx, item)
```
stuartarchibald, line 300:
```suggestion
            expected = pyfunc_setitem(a_empty.copy(), idx, item)
            got = cfunc_setitem(a_empty.copy(), idx, item)
```

stuartarchibald (commented):
Thanks for the updates, I think there's one remaining issue in the testing (noted inline) else this looks good.

numba/tests/test_array_manipulation.py
kc611: Yeah, this is a mistake.

stuartarchibald (approved):
Thanks for the patch and fixes, great to see this long standing feature request implemented!

PR #8517, opened 2022-10-17
make some typedlist C-APIs public

dlee992:
Hi. I make some typedlist C-APIs public, since I want to use typedlist C-APIs in my C++ code, which will be called from jitted numba code, via `intrinsic`, `lower_builtin`, and etc.

At least, (I feel) I need `new`, `free`, `get_length`, `setitem`, and `getitem` APIs. I hope this PR won't bother other parts in numba or the design principle of Numba. If not, please shout.

I mark this PR as a draft, when I finish my extension and test it, then I will mark this as an official PR.

### DIFF
--- a/docs/source/reference/pysupported.rst
+++ b/docs/source/reference/pysupported.rst
@@ -552,11 +552,6 @@ Typed List
   functionality or suffer from unexpectedly bad performance, please report
   this, ideally by opening an issue on the Numba issue tracker.
 
-.. warning::
-  Most interfaces listed in ``numba/cext/listobject.h`` are exported,
-  but they're not supported/stable, and these interfaces or underlying
-  implementations could be changed or removed in future without notice.
-
 As of version 0.45.0 a new implementation of the list data type is available,
 the so-called *typed-list*. This is compiled library backed, type-homogeneous
 list data type that is an improvement over the *reflected-list* mentioned
--- a/docs/source/reference/pysupported.rst
+++ b/docs/source/reference/pysupported.rst
@@ -552,6 +552,11 @@ Typed List
   functionality or suffer from unexpectedly bad performance, please report
   this, ideally by opening an issue on the Numba issue tracker.
 
+.. warning::
+  Most interfaces listed in ``numba/cext/listobject.h`` are exported,
+  but they're not supported/stable, and these interfaces or underlying
+  implementations could be changed or removed in future without notice.
+
 As of version 0.45.0 a new implementation of the list data type is available,
 the so-called *typed-list*. This is compiled library backed, type-homogeneous
 list data type that is an improvement over the *reflected-list* mentioned
--- a/numba/cext/cext.h
+++ b/numba/cext/cext.h
@@ -9,8 +9,7 @@
 #define NUMBA_EXPORT_FUNC(_rettype) VISIBILITY_HIDDEN _rettype
 #define NUMBA_EXPORT_DATA(_vartype) VISIBILITY_HIDDEN _vartype
 
-/* Define symbols in this C module, and export them outside the
-   shared library in various platforms, including Windows. */
+/* Use to declare a symbol as exported (global). */
 #define NUMBA_GLOBAL_FUNC(_rettype) VISIBILITY_GLOBAL _rettype
 
 NUMBA_EXPORT_FUNC(Py_ssize_t)
--- a/numba/cext/cext.h
+++ b/numba/cext/cext.h
@@ -9,6 +9,10 @@
 #define NUMBA_EXPORT_FUNC(_rettype) VISIBILITY_HIDDEN _rettype
 #define NUMBA_EXPORT_DATA(_vartype) VISIBILITY_HIDDEN _vartype
 
+/* Define symbols in this C module, and export them outside the
+   shared library in various platforms, including Windows. */
+#define NUMBA_GLOBAL_FUNC(_rettype) VISIBILITY_GLOBAL _rettype
+
 NUMBA_EXPORT_FUNC(Py_ssize_t)
 aligned_size(Py_ssize_t sz);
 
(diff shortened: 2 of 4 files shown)


gmarkall:
The consensus from the dev meeting on 2022-10-18 was that it's OK to make these APIs public, as long as it's documented that they're not supported / stable, and the interfaces or implementation could be changed in future without notice. So with some documentation of them not being supported / stable, this PR should be OK to merge.

gmarkall:
Notes: https://hackmd.io/SXbcwPbCSwKgo4i52XH5zw

gmarkall:
Sorry, I hit the wrong button and closed the PR.

stuartarchibald:
> The consensus from the dev meeting on 2022-10-18 was that it's OK to make these APIs public, as long as it's documented that they're not supported / stable, and the interfaces or implementation could be changed in future without notice. So with some documentation of them not being supported / stable, this PR should be OK to merge.

I think "_changed_ **OR** _removed_ in future without notice" may better cover the potential liability associated with their use?

dlee992:
Thanks, I've already made some changes in docs/header comments/codes, following your suggestions. Can begin to review.

docs/source/reference/pysupported.rst
```diff
@@ -552,6 +552,11 @@ Typed List
   functionality or suffer from unexpectedly bad performance, please report
   this, ideally by opening an issue on the Numba issue tracker.
 
+.. warning::
+  Most interfaces listed in ``numba/cext/listobject.h`` are exported,
+  but they're not supported/stable, and these interfaces or underlying
+  implementations could be changed or removed in future without notice.
+
```
gmarkall, line 559: I think this note is a little out-of-place in the section of the manual that discusses support for Python features. Given that the same warning / disclaimer is present in `listobject.h`, I'd suggest just removing this change.

```suggestion
```

gmarkall (changes requested):
This looks good to me (once the modification to pysupported.rst is removed), but given the nature of the changes and prior discussion, someone else should take a look too.

docs/source/reference/pysupported.rst
dlee992: Okay, solved as you requested.

gmarkall (approved):
Thanks for the update - this looks good to me (though still should have a second look from someone else).

numba/cext/listobject.h
```diff
@@ -4,7 +4,10 @@
  *
  * https://github.com/python/cpython/blob/51ddab8dae056867f3595ab3400bffc93f67c8d4/Include/listobject.h
  *
- *
+ * WARNING:
+ * Most interfaces listed here are exported, but they're not supported/stable,
+ * and these interfaces or underlying implementations could be changed or
+ * removed in future without notice.
```
stuartarchibald, line 10:
```suggestion
 * Most interfaces listed here are exported (global), but they are not
 * supported, stable, or part of Numba's public API. These interfaces and their
 * underlying implementations may be changed or removed in future without
 * notice.
```
suggest this to note that the symbols are global, and to make it explicit that they are not part of Numba's public API.

stuartarchibald (commented):
Thanks for the patch @dlee992. I've suggested a minor change to the warning text in an attempt to be more explicit with respect to a couple of points.

The only other thought I have is whether the exported symbols should have `protected` visibility just for additional safety (though this may be excessive). Do you have a view on this @gmarkall ?

gmarkall:
> The only other thought I have is whether the exported symbols should have `protected` visibility just for additional safety (though this may be excessive). Do you have a view on this @gmarkall ?

It's not clear to me how that would make anything safer - can you elaborate?

stuartarchibald:
> > The only other thought I have is whether the exported symbols should have `protected` visibility just for additional safety (though this may be excessive). Do you have a view on this @gmarkall ?
> 
> It's not clear to me how that would make anything safer - can you elaborate?

From OOB discussion, conclusion is that it probably makes insufficient difference. 

However, as part of this discussion, it was noted that the "default" behaviour in relation to symbol export varies across platforms and that it would be a good idea to ensure that symbols are exported consistently. For example, I think this patch as of ac31d64 will result in some typed list symbols being `global` on unix systems, but on windows I think they will need explicitly exporting to achieve the same. Thoughts?

gmarkall:
I think the default visibility on OS X / Linux was already fine - on Windows I think `__declspec(dllexport)` is needed: https://learn.microsoft.com/en-us/cpp/build/exporting-from-a-dll-using-declspec-dllexport?view=msvc-170

dlee992:
> I think the default visibility on OS X / Linux was already fine - on Windows I think `__declspec(dllexport)` is needed: https://learn.microsoft.com/en-us/cpp/build/exporting-from-a-dll-using-declspec-dllexport?view=msvc-170

Cool, I should revise it again based on this info.

gmarkall (dismissed):
Thanks for the revisions - this looks good to me.

gmarkall:
@stuartarchibald Are you happy with this PR at this point?

stuartarchibald:
> @stuartarchibald Are you happy with this PR at this point?

I think what is now present makes the exporting of declared symbols as globally visible consistent across platforms, thanks for addressing this @dlee992.

numba/cext/cext.h
```diff
@@ -9,6 +9,10 @@
 #define NUMBA_EXPORT_FUNC(_rettype) VISIBILITY_HIDDEN _rettype
 #define NUMBA_EXPORT_DATA(_vartype) VISIBILITY_HIDDEN _vartype
 
+/* Define symbols in this C module, and export them outside the
+   shared library in various platforms, including Windows. */
```
stuartarchibald, line 13:
```suggestion
/* Use to declare a symbol as exported (global). */
```

stuartarchibald (commented):
Thanks for the updates @dlee992, one minor suggestion RE commentary for the C macro else looks good.

numba/cext/cext.h
dlee992: Thanks for the review!

gmarkall (approved):
Following the small comment update, I still approve.

stuartarchibald (approved):
Thanks for the patch and fixes. I manually checked on a linux system that the supposedly exported symbols are indeed global in the `_helperlib.<version>.so`.

PR #8522, opened 2022-10-19
Import jitclass from numba.experimental in jitclass documentation

armgabrielyan:
`jitclass` is still part of experimental support, so it should be imported from `numba.experimental`. Updating the documentation for `jitclass` to reflect it.


stuartarchibald (approved):
Thanks for fixing these mistakes in the docs.

armgabrielyan:
Thank you!

PR #8525, opened 2022-10-20
Making CUDA specific datamodel manager

sklam:
so that CUDA datamodel (like the fp16 model) does not infect the default datamodel in case CPU needs to have a different datamodel

### DIFF
--- a/numba/core/datamodel/manager.py
+++ b/numba/core/datamodel/manager.py
@@ -9,6 +9,12 @@ class DataModelManager(object):
     """
 
     def __init__(self, handlers=None):
+        """
+        Parameters
+        -----------
+        handlers: Mapping[Type, DataModel] or None
+            Optionally provide the initial handlers mapping.
+        """
         # { numba type class -> model factory }
         self._handlers = handlers or {}
         # { numba type instance -> model instance }
@@ -42,11 +48,21 @@ class DataModelManager(object):
         Use this to inherit from the default data model and specialize it
         for custom target.
         """
-        dmm = DataModelManager()
-        dmm._handlers = self._handlers.copy()
-        return dmm
+        return DataModelManager(self._handlers.copy())
 
     def chain(self, other_manager):
+        """Create a new DataModelManager by chaining the handlers mapping of
+        `other_manager` with a fresh handlers mapping.
+
+        Any existing and new handlers inserted to `other_manager` will be
+        visible to the new manager. Any handlers inserted to the new manager
+        can override existing handlers in `other_manager` without actually
+        mutating `other_manager`.
+
+        Parameters
+        ----------
+        other_manager: DataModelManager
+        """
         chained = ChainMap(self._handlers, other_manager._handlers)
         return DataModelManager(chained)
 
--- a/numba/core/datamodel/manager.py
+++ b/numba/core/datamodel/manager.py
@@ -1,4 +1,5 @@
 import weakref
+from collections import ChainMap
 
 from numba.core import types
 
@@ -7,9 +8,9 @@ class DataModelManager(object):
     """Manages mapping of FE types to their corresponding data model
     """
 
-    def __init__(self):
+    def __init__(self, handlers=None):
         # { numba type class -> model factory }
-        self._handlers = {}
+        self._handlers = handlers or {}
         # { numba type instance -> model instance }
         self._cache = weakref.WeakKeyDictionary()
 
@@ -44,4 +45,8 @@ class DataModelManager(object):
         dmm = DataModelManager()
         dmm._handlers = self._handlers.copy()
         return dmm
+
+    def chain(self, other_manager):
+        chained = ChainMap(self._handlers, other_manager._handlers)
+        return DataModelManager(chained)
 
--- a/numba/cuda/models.py
+++ b/numba/cuda/models.py
@@ -1,9 +1,16 @@
+import functools
+
 from llvmlite import ir
 
-from numba.core.datamodel.registry import register_default
-from numba.core.extending import register_model, models
+from numba.core.datamodel.registry import DataModelManager, register
+from numba.core.extending import models
 from numba.core import types
 from numba.cuda.types import Dim3, GridGroup, CUDADispatcher
+
+
+cuda_data_manager = DataModelManager()
+
+register_model = functools.partial(register, cuda_data_manager)
 
 
 @register_model(Dim3)
@@ -24,7 +31,7 @@ class GridGroupModel(models.PrimitiveModel):
         super().__init__(dmm, fe_type, be_type)
 
 
-@register_default(types.Float)
+@register_model(types.Float)
 class FloatModel(models.PrimitiveModel):
     def __init__(self, dmm, fe_type):
         if fe_type == types.float16:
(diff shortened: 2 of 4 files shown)


gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

stuartarchibald:
Should this be integrated into the target extension API in `numba.core.target_extension`?

gmarkall (changes requested):
I think this looks good, but I do note that there's not an additional test that registering models with the CUDA target doesn't enable the CPU target to find them, which is probably a good idea.

@testhound What do you think of this changes? Does this adequately support the addition of an FP16 CPU implementation that uses a different model to the CUDA target?

stuartarchibald:
> > Should this be integrated into the target extension API in `numba.core.target_extension`?
> 
> As for now, the datamodel is decided by the TargetContext. I'm not sure how it will go into target_extension.

It could look something like this (untested, and on top of this patch):
```diff
$ git diff
diff --git a/numba/core/base.py b/numba/core/base.py
--- a/numba/core/base.py
+++ b/numba/core/base.py
@@ -230,7 +230,8 @@ class BaseContext(object):
 
         self.address_size = utils.MACHINE_BITS
         self.typing_context = typing_context
-        from numba.core.target_extension import target_registry
+        from numba.core.target_extension import (target_registry,
+                                                 datamodel_registry)
         self.target_name = target
         self.target = target_registry[target]
 
@@ -251,7 +252,7 @@ class BaseContext(object):
 
         self._boundscheck = False
 
-        self.data_model_manager = datamodel.default_manager
+        self.data_model_manager = datamodel_registry[target]
 
         # Initialize
         self.init()
diff --git a/numba/core/target_extension.py b/numba/core/target_extension.py
--- a/numba/core/target_extension.py
+++ b/numba/core/target_extension.py
@@ -166,3 +166,6 @@ dispatcher_registry = DelayedRegistry(key_type=Target)
 cpu_target = target_registry['cpu']
 dispatcher_registry[cpu_target] = CPUDispatcher
 jit_registry[cpu_target] = jit
+
+# Registry mapping Target classes to the datamodel manager for that target.
+datamodel_registry = DelayedRegistry(key_type=Target)
diff --git a/numba/cuda/models.py b/numba/cuda/models.py
index 21d1151..f075d63 100644
--- a/numba/cuda/models.py
+++ b/numba/cuda/models.py
@@ -6,10 +6,12 @@ from numba.core.datamodel.registry import DataModelManager, register
 from numba.core.extending import models
 from numba.core import types
 from numba.cuda.types import Dim3, GridGroup, CUDADispatcher
-
+from numba.core.target_extension import target_registry, datamodel_registry
 
 cuda_data_manager = DataModelManager()
 
+datamodel_registry[target_registry["cuda"]] = cuda_data_manager
+
 register_model = functools.partial(register, cuda_data_manager)
```
I think the potential advantage of having this is that it requires extensions to consider where their data models are being registered opposed to the default at present which is adding to the existing/CPU DMM.

testhound:
@stuartarchibald I took a stab at testing your changes with target_extension and received a mapping error which I resolved with this one line patch:

diff --git a/numba/core/base.py b/numba/core/base.py
index 1c8227e11..a8f83f793 100644
--- a/numba/core/base.py
+++ b/numba/core/base.py
@@ -252,7 +252,7 @@ class BaseContext(object):
 
        ```
 self._boundscheck = False
 
-        self.data_model_manager = datamodel_registry[target]
+        self.data_model_manager = datamodel_registry[target_registry[target]]
 
         # Initialize
         self.init()
```
After making that change I tried running CUDA test but received this error and traceback:

  File "/home/collison/Nvidia/testhound_numba_fork/numba/testing/__init__.py", line 29, in load_testsuite
    suite.addTests(loader.loadTestsFromName(f))
  File "/home/collison/anaconda3/envs/numbaenv/lib/python3.8/unittest/loader.py", line 154, in loadTestsFromName
  File "/home/collison/Nvidia/testhound_numba_fork/numba/cuda/tests/cudapy/test_extending.py", line 24, in <module>
    def interval_width(interval):
  File "/home/collison/Nvidia/testhound_numba_fork/numba/core/decorators.py", line 258, in njit
    return jit(*args, **kws)
  File "/home/collison/Nvidia/testhound_numba_fork/numba/core/decorators.py", line 179, in jit
    return wrapper(pyfunc)
  File "/home/collison/Nvidia/testhound_numba_fork/numba/core/decorators.py", line 208, in wrapper
    disp = dispatcher(py_func=func, locals=locals,
  File "/home/collison/Nvidia/testhound_numba_fork/numba/core/dispatcher.py", line 824, in __init__
    self.targetctx = self.targetdescr.target_context
  File "/home/collison/Nvidia/testhound_numba_fork/numba/core/registry.py", line 47, in target_context
    return self._toplevel_target_context
  File "/home/collison/anaconda3/envs/numbaenv/lib/python3.8/functools.py", line 967, in __get__
    val = self.func(instance)
  File "/home/collison/Nvidia/testhound_numba_fork/numba/core/registry.py", line 31, in _toplevel_target_context
    return cpu.CPUContext(self.typing_context, self._target_name)
  File "/home/collison/Nvidia/testhound_numba_fork/numba/core/cpu.py", line 43, in __init__
    super().__init__(typingctx, target)
  File "/home/collison/Nvidia/testhound_numba_fork/numba/core/base.py", line 255, in __init__
    self.data_model_manager = datamodel_registry[target_registry[target]]
  File "/home/collison/Nvidia/testhound_numba_fork/numba/core/registry.py", line 99, in __getitem__
    return super(DelayedRegistry, self).__getitem__(item)
KeyError: <class 'numba.core.target_extension.CPU'>

Are we missing registration of a datamodel for the cpu with your recent change?

stuartarchibald:
> Are we missing registration of a datamodel for the cpu with your recent change?

Yes, I think a similar registration for the CPU target is necessary. This patch was a bit of a "guess" at what would be needed and is/was untested, thanks for testing it!

This now also makes me wonder if the standard DMM needs to be bound to the "generic" target, with the CPU target then having a chained DMM registry cf. the CUDA target as in the original patch. In this way a new DM could be shared between targets when needed and kept separate otherwise. Further, consider the case of e.g. a CPU specific DM, the CUDA target shouldn't be able to "see" that any more than a CUDA specific DM should be "seen" by the CPU.

Suggest this is discussed at the public Numba meeting today.

sklam:
Summarizing discussion from today's public meeting, I'll make a new PR based on this one to try out the [target-extension suggestion](https://github.com/numba/numba/pull/8525#issuecomment-1313639254). This should not affect merging this PR for the short-term.

I will add tests as per review at https://github.com/numba/numba/pull/8525#pullrequestreview-1178990722 so to keep this PR moving along.

(ping @testhound)

sklam:
@gmarkall, I added a generic test for chaining datamodel such that it doesn't need any target at https://github.com/numba/numba/pull/8525/commits/c3187f82cd0c70e953751ae99f779eb395596bd0

sklam:
> Yes, I think a similar registration for the CPU target is necessary.

I tried the patch and attempted to fix it. It leads to some spectacular circular import problem that I don't have time to fix.

Issue #8529, opened 2022-10-21
CUDA test: `test_reinterpret_array_type` is failing on NumPy 1.23


stuartarchibald:
## Reporting a bug

- [x] I have tried using the latest released version of Numba (most recent is
 visible in the change log (https://github.com/numba/numba/blob/main/CHANGE_LOG).
- [x] I have included a self contained code sample to reproduce the problem.
  i.e. it's possible to run as 'python bug.py'.

Reproducer is simply install Numba (from 0.56 series or current `main`) alongside NumPy>=1.23 and then run the test:
```
./runtests.py numba.cuda.tests.cudapy.test_array_methods.TestCudaArrayMethod.test_reinterpret_array_type
```
Failure is roughly:
```
<snip>
numba/np/arrayobj.py, line 2540, in array_view
numba/core/base.py, line 559, in get_function
numba/core/base.py, line 561, in get_function

NotImplementedError: No definition for lowering <built-in method impl of ....>(array(uint8, 1d, C), class(int32)) -> none
```

I suspect it's this that's causing the issue:
https://github.com/numba/numba/blob/f4b6fe0124fa7734ba2bb28c7685db5b6306c486/numba/np/arrayobj.py#L2559-L2569
the lowering for `array.view` is being "borrowed" by the CUDA target, but this branch for NumPy 1.23 resolves an `@overload` (following) to determine if it's a compatible view and something goes wrong. I would guess that it either refuses to compile on CUDA (which is probably correct) or the current compilation target gets mixed up leading to similar.
https://github.com/numba/numba/blob/f4b6fe0124fa7734ba2bb28c7685db5b6306c486/numba/np/arrayobj.py#L2498-L2539

The same test runs fine on NumPy 1.22.

CC @gmarkall.

stuartarchibald:
Think this will fix it:
```python
diff --git a/numba/np/arrayobj.py b/numba/np/arrayobj.py
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -2467,7 +2467,7 @@ def _compatible_view(a, dtype):
     pass
 
 
-@overload(_compatible_view)
+@overload(_compatible_view, target='generic')
 def ol_compatible_view(a, dtype):
     """Determines if the array and dtype are compatible for forming a view."""
     # NOTE: NumPy 1.23+ uses this check.
```

PR #8532, opened 2022-10-24
Vary NumPy version on gpuCI

gmarkall:
Test with different NumPy versions on gpuCI. It doesn't matter which NumPy version goes with which CUDA toolkit, and there's no point having the cartesian product tested.

This tests the oldest supported NumPy (1.19 for0.57+), and then the most recent versions (1.21, 1.22, and 1.23).

As one would expect, NumPy 1.23 is failing because that's broken on mainline.

### DIFF
--- a/buildscripts/gpuci/build.sh
+++ b/buildscripts/gpuci/build.sh
@@ -24,6 +24,13 @@ else
   export NUMBA_CUDA_USE_NVIDIA_BINDING=0;
 fi;
 
+# Test with different NumPy versions with each toolkit (it's not worth testing
+# the cartesian product of versions here, we just need to test with different
+# CUDA and NumPy versions).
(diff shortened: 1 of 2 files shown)


gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

buildscripts/gpuci/build.sh
```diff
@@ -24,6 +24,13 @@ else
   export NUMBA_CUDA_USE_NVIDIA_BINDING=0;
 fi;
 
+# Test with different NumPy versions with each toolkit (it's not worth testing
+# the cartesian product of versions here, we just need to test with different
```
stuartarchibald, line 28:
```suggestion
# the Cartesian product of versions here, we just need to test with different
```

stuartarchibald (approved):
Approving with knowledge that `gpuci` is not passing the NumPy 1.23 testing variant due to #8529, this reflects the true state of mainline as of f4b6fe0.

PR #8537, opened 2022-10-25
Make ol_compatible_view accessible on all targets

### LINKED ISSUE #8529: CUDA test: `test_reinterpret_array_type` is failing on NumPy 1.23

gmarkall:
This is to fix Issue #8529, where test_reinterpret_array_type fails on CUDA with NumPy 1.23 because this overload is needed. Making it accessible to the CUDA target should resolve the issue.

Fixes #8529.

### DIFF
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -2495,7 +2495,7 @@ def _compatible_view(a, dtype):
     pass
 
 
-@overload(_compatible_view)
+@overload(_compatible_view, target='generic')
 def ol_compatible_view(a, dtype):
     """Determines if the array and dtype are compatible for forming a view."""
     # NOTE: NumPy 1.23+ uses this check.


gmarkall:
gpuci run tests

stuartarchibald (approved):
Thanks for the patch. This has been manually tested locally as part of debugging: https://github.com/numba/numba/issues/8529#issuecomment-1287079195. I think it also makes sense as the most minimally invasive method to resolve the problem, i.e. the CUDA target using some lowering from the CPU target that calls out to an `@overload`... that `@overload` needs to be for the `generic` target else the CUDA target cannot "see" it.

Further, as `gpuci` is now passing (it wasn't in #8532, or by induction the merge to `main` in 9d034e9) this also suggests the fix is working.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_164`.

Passed.

PR #8539, opened 2022-10-25
Fix #8534, np.broadcast_to should update array size attr.

### LINKED ISSUE #8534: Numba's `np.broadcast_to` doesn't change the size of the array
## Reporting a bug

As titled, Numba's `np.broadcast_to` doesn't update the size of an array after broadcasting it. 

MWE as follows:

```python
import numpy as np
import numba

@numba.njit     
def bdcast(array):
    bdcast_array = np.broadcast_to(array, (2, 3))
    return bdcast_array.size

array = np.array([1,2,0])

expected = bdcast.py_func(array)
got = bdcast(array)

# Initial size of array
print(array.size) # 3

# Expected size after broadcast
print(expected) # 6
# Returned size after broadcast
print(got) # 3

np.testing.assert_equal(got, expected)

```

stuartarchibald:
As title.

Fixes #8534


sklam (approved):
Thanks for the patch!

PR #8541, opened 2022-10-25
Remove restoration of "free" channel in Azure CI windows builds.

stuartarchibald:
This was added for continuing NumPy 1.10 support, which has long since been removed.


sklam (approved):
CI passed. Looks good

PR #8542, opened 2022-10-26
CUDA: Make arg optional for Stream.add_callback()

gmarkall:
It's a bit annoying to have to pass an argument if your callback doesn't need one. To make this use case a little more pleasant, we default `arg` to `None`.


gmarkall:
gpuci run tests

gmarkall:
Note that this should be considered a gpuCI pass, it only failed because #8537 is not yet merged.

stuartarchibald (approved):
Thanks for the patch.

PR #8550, opened 2022-10-28
Changes how tests are split between test instances

apmasell:
Since tests are dynamically generated on different platforms, uniformly slicing a list of tests can cause individual tests to be lost or tested in multiple platforms. This changes tests to be sharded rather than sliced so that each worker is independently making the same choice on a per-test basis and all dynamically generated tests are always run.

### DIFF
--- a/numba/testing/main.py
+++ b/numba/testing/main.py
@@ -10,6 +10,7 @@ import sys
 import time
 import unittest
 import warnings
+import zlib
 
 from functools import lru_cache
 from io import StringIO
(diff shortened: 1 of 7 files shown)


sklam (commented):
The chunking logic looks good to me. My only comments are for the `always_test` attribute. I'd suggest adding a test for it and marking the test itself as `always_test`, so there can be a `test_runtests` test that check if `always_test` logic is working.

sklam (dismissed):
LGTM

stuartarchibald:
Does this statement need any alterations?
https://github.com/numba/numba/blob/0441bb17c7820efc2eba4fd141b68dac2afa4740/docs/source/developer/contributing.rst?plain=1#L376-L379

numba/testing/main.py
```diff
@@ -84,19 +85,36 @@ def cuda_sensitive_mtime(x):
...
-        return l['sl']
+        (index, count) = useslice.split(":")
+        index = int(index)
+        count = int(count)
     except Exception:
-        msg = ("Expected arguments consumable by 'slice' to follow "
-                "option `-j`, found '%s'" % useslice)
+        msg = (
+                    "Expected arguments shard index and count to follow "
+                    "option `-j`, found '%s'" % useslice)
```
stuartarchibald, line 106: This doesn't explain how to specify them correctly? It's expecting syntax `-j index:count` ?

numba/tests/test_runtests.py
```diff
@@ -120,20 +120,33 @@ def get_count(arg_list):
...
-        l = [x for x in tmp if x.startswith('numba.')]
-        self.assertEqual(len(l), 5)
+        tmpAll = self.get_testsuite_listing([])
+        tmp1 = self.get_testsuite_listing(['-j', '0:2'])
+        tmp2 = self.get_testsuite_listing(['-j', '1:2'])
+        lAll = {x for x in tmpAll if x.startswith('numba.')}
+        l1 = {x for x in tmp1 if x.startswith('numba.')}
+        l2 = {x for x in tmp2 if x.startswith('numba.')}
+        # The variance between two adjacent shards should be less than 5% of
+        # the total
```
stuartarchibald, line 130: Is this just a guess based on what happens or is there some fundamental property involved here? Also, is this the variance or the difference?

numba/tests/test_runtests.py
```diff
@@ -120,20 +120,33 @@ def get_count(arg_list):
...
+        # Test should always run
+        self.assertIn("numba.tests.test_api.TestNumbaModule.test_numba_module",
+                      l1)
+        self.assertIn("numba.tests.test_api.TestNumbaModule.test_numba_module",
+                      l2)
 
     def test_check_slicing_equivalent(self):
         def filter_test(xs):
-            return [x for x in xs if x.startswith('numba.')]
+            return {x for x in xs if x.startswith('numba.')}
```
stuartarchibald, line 142: Is the use of `set`s in this test potentially hiding bugs in that a set implicitly filters duplicates?

stuartarchibald (commented):
Thanks for the patch. I've a few questions left in line to resolve but otherwise looks good.

numba/tests/test_runtests.py
apmasell: It's fundamental. If you have a hash distribution, you don't extend it to be perfectly flat.

numba/tests/test_runtests.py
apmasell: There will be duplicates for any tests that are marked to run on all shards.

apmasell:
The documentation in the contribution document is still accurate.

numba/tests/test_runtests.py
sklam: btw, the duplicates will be the always running tests.

sklam (approved):
All review comments are addressed.

PR #8553, opened 2022-10-31
Update CHANGE_LOG for 0.56.4

stuartarchibald:
As title.


esc (approved):
I manually inspected the source and looked at the generated text via readthedocs. It looks good to me.

PR #8554, opened 2022-10-31
Make target for `@overload` have 'generic' as default.

### LINKED ISSUE #8530: Cross target use of target-less `@overload` leads to lowering errors.
## Reporting a bug

This incorrect use of `@overload` is actually a simplified version of the issue in #8529. Note that the `@overload` has no `target=` kwarg, as such the CUDA target cannot find it, but _also_ as the `target` is merely implied as CPU to preserve existing behaviours, this manifests as a lowering error as there's no `bar` function implementation in the target context for CUDA.
```python
import numpy as np
from numba.cuda import jit

def bar():
    pass


@overload(bar)
def ol_bar(x):
    def impl(x):
        return 12
    return impl


@jit
def call_bar(arr):
    arr[0] = bar(3)

tmp = np.zeros((1,))
call_bar[1, 1](tmp)
print(tmp)
```
The error message ends with:
```
NotImplementedError: No definition for lowering <... Closure>(int64,) -> Literal[int](12)
```
which is not ideal.

If the example is updated so that the overload is declared as `cpu` only via `@overload(bar, target='cpu')`, then the error message is more helpful:
```
Function resolution cannot find any matches for function '<function bar at 0x<snip>>` for the current target: '<class 'numba.core.target_extension.CUDA'>'.
```

stuartarchibald:
This makes the default `target` for `@overload` the `'generic'` target, as, in effect, very few `@overload` implementations are genuinely target specific. Further, the use of the standard `numba.jit` wrapper (with no `target` set) for `@overload` implementations leads to strange cases where some function cannot be resolved on e.g. the CUDA target, but because there's no target set in the `jit` wrapper the error message cannot tell you why, it just materialises as a lowering error (i.e. this should fix #8530).

The first commit in this PR (a3ecb1f) is from @gmarkall's Numba fork on branch `overload-target-generic`, the corresponding commit is: 38b537993dbcf7c73c95c815e1bc955d6f917010, authorship is preserved, many thanks @gmarkall!

### DIFF
--- a/numba/core/extending.py
+++ b/numba/core/extending.py
@@ -53,8 +53,10 @@ def type_callable(func):
 
 
 # By default, an *overload* does not have a cpython wrapper because it is not
-# callable from python.
-_overload_default_jit_options = {'no_cpython_wrapper': True}
+# callable from python. It also has `nopython=True`, this has been default since
+# its inception!
+_overload_default_jit_options = {'no_cpython_wrapper': True,
+                                 'nopython':True}
 
 
 def overload(func, jit_options={}, strict=True, inline='never',
--- a/numba/core/typing/templates.py
+++ b/numba/core/typing/templates.py
@@ -716,37 +716,30 @@ class _OverloadFunctionTemplate(AbstractTemplate):
     def _get_jit_decorator(self):
         """Gets a jit decorator suitable for the current target"""
 
-        jitter_str = self.metadata.get('target', None)
-        if jitter_str is None:
-            from numba import jit
-            # There is no target requested, use default, this preserves
-            # original behaviour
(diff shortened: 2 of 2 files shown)


gmarkall:
gpuci run tests

stuartarchibald:
RE patch approval. I approve a3ecb1f which was authored by @gmarkall. I had made identical changes locally before realising that @gmarkall already had a patch for it so am happy with what is present!

gmarkall (approved):
This looks good to me, though I agree with what was said OOB about a buildfarm run being a good idea first.

esc:
`Build numba_smoketest_cuda_yaml_169 has started`

gmarkall:
@esc Many thanks!

PR #8555, opened 2022-10-31
Try xml test reporting in Azure

apmasell:


setup.py
```diff
@@ -371,7 +371,7 @@ def check_file_at_path(path2file):
 
 packages = find_packages(include=["numba", "numba.*"])
 
-build_requires = ['numpy >={}'.format(min_numpy_build_version)]
+build_requires = ['numpy >={}'.format(min_numpy_build_version), 'unittest-xml-reporting']
```
sklam, line 374: The test failure may be caused by some broken packages with xml. Try removing 'unittest-xml-reporting' on this line so the py3.10 build is not trying to install it from `pip`. 

Later, we will need to compare the xml packages for py3.9 vs py3.10

apmasell: Thanks! That worked

buildscripts/incremental/setup_conda_environment.cmd
```diff
@@ -21,9 +21,11 @@ call deactivate
 @rem Display root environment (for debugging)
 conda list
 @rem Scipy, CFFI, jinja2 and IPython are optional dependencies, but exercised in the test suite
-conda create -n %CONDA_ENV% -q -y python=%PYTHON% numpy=%NUMPY% cffi pip scipy jinja2 ipython gitpython pyyaml
+conda create -n %CONDA_ENV% -q -y python=%PYTHON% numpy=%NUMPY% cffi pip scipy jinja2 ipython gitpython pyyaml libxml2
 
 call activate %CONDA_ENV%
+@rem Install latest unittest-xml-reporting
+%CONDA_INSTALL% -c conda-forge unittest-xml-reporting
```
sklam, line 28:
```suggestion
%CONDA_INSTALL% conda-forge::unittest-xml-reporting
```
so it won't try to change anything to use conda-forge unnecessarily.

sklam: and that it matches https://github.com/numba/numba/pull/8555/files#diff-319afae18279ecbf9d01cb17a6091e1103ddccadda228ace0a011a5590aa3667R84

numba/testing/main.py
```diff
@@ -807,3 +808,62 @@ def run(self, test):
         # This will call self._run_inner() on the created result object,
         # and print out the detailed test results at the end.
         return super(ParallelTestRunner, self).run(self._run_inner)
+
+
+
+
+import xmlrunner
```
sklam, line 815: Let's make this dependency (unittest-xml-reporting) optional. 
And, put it at the top of the file. (sorry, this is probably from my hacky commit )

numba/testing/__init__.py
```diff
@@ -34,7 +34,7 @@ def load_testsuite(loader, dir):
 
 
 def run_tests(argv=None, defaultTest=None, topleveldir=None,
-              xmloutput=None, verbosity=1, nomultiproc=False):
+              xmloutput="junit_reports", verbosity=1, nomultiproc=False):
```
sklam, line 37: Let's change the argparser so that xmloutput is read from the cmdline. Keep the default to not producing xml.

docs/environment.yml
```diff
@@ -4,12 +4,14 @@
...
 channels:
   - numba/label/dev
+  - conda-forge
 dependencies:
   - python=3.7
   - llvmlite=0.40
   - numpy
   - numpydoc
   - setuptools
+  - unittest-xml-reporting
```
sklam, line 14: We don't really need these changes once xmlrunner is optional

sklam (changes requested):
We should keep the testsuite behavior the same by default, and add something like a `--xml=junits_report` to make it write out XML. The dependency on unittest-xml-reporting should be optional.

Let's treat changes from my commits as hack. They need some cleanup.

numba/testing/main.py
```diff
@@ -807,3 +808,62 @@ def run(self, test):
...
+
+            for idx, status in pool.imap_unordered(xml_child_runner, enumerate(splitted_tests)):
+                if not status:
+                    case = type(f"ParallelTest{idx}", (_FailedTest,), {})
+                    ss = loader.defaultTestLoader.loadTestsFromTestCase(case)
+                    ss.run(result)
+
+        stests = SerialSuite(self._stests)
+        stests.run(result)
+        return result
```
sklam, line 854: This part may have to look a bit like:


```

    def _run_inner(self, result):
        pool = multiprocessing.get_context("spawn").Pool(self.nprocs)

        ptests = self._ptests
        print("len(ptests)", len(ptests))

        # Split the tests and recycle the worker process to tame memory usage.
        chunk_size = 100
        splitted_tests = [self._ptests[i:i + chunk_size]
                        for i in range(0, len(self._ptests), chunk_size)]

        for tests in splitted_tests:
            pool = multiprocessing.Pool(self.nprocs)
            try:
                result = xml_child_runner(tests)
            except:
                # On exception, kill still active workers immediately
                pool.terminate()
                # Make sure exception is reported and not ignored
                raise
            else:
                # Close the pool cleanly unless asked to early out
                if result.shouldStop:
                    pool.terminate()
                    break
                else:
                    pool.close()
            finally:
                # Always join the pool (this is necessary for coverage.py)
                pool.join()
        if not result.shouldStop:
            stests = SerialSuite(self._stests)
            stests.run(result)
            return result
        return result
```

which is mostly a copy of https://github.com/numba/numba/blob/bfe1270e23f8daf7fd95cfb6ef385f6c832fd927/numba/testing/main.py#L744-L776

sklam (commented):
I suspect the slowdown is due to the missing chunking logic:

sklam:
https://github.com/numba/numba/pull/8630#issuecomment-1334114164 is another case why this is needed

apmasell:
This is going to be redesigned.

PR #8556, opened 2022-10-31
Fix `np.fmod` to match numpy behavior

benwilliamgraham:
Fix for `np.fmod` so that `np.fmod` with negative numbers matches numpy's behavior

### DIFF
--- a/numba/np/npyfuncs.py
+++ b/numba/np/npyfuncs.py
@@ -269,7 +269,7 @@ def np_int_fmod_impl(context, builder, sig, args):
     result.add_incoming(mod, bb_if)
 
     # set the sign to match the numerator
-    num_gt_zero = builder.icmp(lc.ICMP_SGT, num, ZERO)
+    num_gt_zero = builder.icmp_signed(">", num, ZERO)
 
     return builder.select(num_gt_zero, result, builder.neg(result))
 
--- a/numba/np/npyfuncs.py
+++ b/numba/np/npyfuncs.py
@@ -252,26 +252,17 @@ def np_int_fmod_impl(context, builder, sig, args):
     ty = sig.args[0]  # any arg type will do, homogeneous
 
     ZERO = context.get_constant(ty, 0)
-    den_not_zero = builder.icmp_signed("!=", ZERO, den)
+    den_not_zero = builder.icmp_unsigned('!=', ZERO, den)
     bb_no_if = builder.basic_block
     with cgutils.if_unlikely(builder, den_not_zero):
         bb_if = builder.basic_block
-
-        # use the abs of the numerator and denominator
-        num_neg = builder.icmp_signed("<", num, ZERO)
-        den_neg = builder.icmp_signed("<", den, ZERO)
-        num_abs = builder.select(num_neg, builder.neg(num), num)
-        den_abs = builder.select(den_neg, builder.neg(den), den)
-        mod = builder.urem(num_abs, den_abs)
+        mod = builder.srem(num,den)
 
     result = builder.phi(ZERO.type)
     result.add_incoming(ZERO, bb_no_if)
     result.add_incoming(mod, bb_if)
 
-    # set the sign to match the numerator
-    num_gt_zero = builder.icmp_signed(">", num, ZERO)
-
-    return builder.select(num_gt_zero, result, builder.neg(result))
+    return result
 
 
 def np_real_div_impl(context, builder, sig, args):
--- a/numba/np/npyfuncs.py
+++ b/numba/np/npyfuncs.py
@@ -243,9 +243,35 @@ def np_int_udivrem_impl(context, builder, sig, args):
     return context.make_tuple(builder, sig.return_type, [div, rem])
 
 
-# implementation of int_fmod is in fact the same as the unsigned remainder,
-# that is: srem with a special case returning 0 when the denominator is 0.
-np_int_fmod_impl = np_int_urem_impl
+def np_int_fmod_impl(context, builder, sig, args):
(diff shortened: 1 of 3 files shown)


benwilliamgraham:
Need to update to use `icmp_signed` (was initially tested on numba 0.55)

numba/np/npyfuncs.py
```diff
@@ -243,9 +243,35 @@ def np_int_udivrem_impl(context, builder, sig, args):
...
+
+    ZERO = context.get_constant(ty, 0)
+    den_not_zero = builder.icmp(lc.ICMP_NE, ZERO, den)
+    bb_no_if = builder.basic_block
+    with cgutils.if_unlikely(builder, den_not_zero):
+        bb_if = builder.basic_block
+
+        # use the abs of the numerator and denominator
+        num_neg = builder.icmp(lc.ICMP_SLT, num, ZERO)
+        den_neg = builder.icmp(lc.ICMP_SLT, den, ZERO)
```
benwilliamgraham, line 262:
```suggestion
    den_not_zero = builder.icmp_signed("!=", ZERO, den)
    bb_no_if = builder.basic_block
    with cgutils.if_unlikely(builder, den_not_zero):
        bb_if = builder.basic_block

        # use the abs of the numerator and denominator
        num_neg = builder.icmp_signed("<", num, ZERO)
        den_neg = builder.icmp_signed("<", den, ZERO)
```

numba/np/npyfuncs.py
```diff
@@ -243,9 +243,35 @@ def np_int_udivrem_impl(context, builder, sig, args):
...
+        num_abs = builder.select(num_neg, builder.neg(num), num)
+        den_abs = builder.select(den_neg, builder.neg(den), den)
+        mod = builder.urem(num_abs, den_abs)
+
+    result = builder.phi(ZERO.type)
+    result.add_incoming(ZERO, bb_no_if)
+    result.add_incoming(mod, bb_if)
+
+    # set the sign to match the numerator
+    num_gt_zero = builder.icmp(lc.ICMP_SGT, num, ZERO)
```
benwilliamgraham, line 272:
```suggestion
    num_gt_zero = builder.icmp_signed(">", num, ZERO)
```

stuartarchibald:
@benwilliamgraham thanks for the patch. Please could open an issue that demonstrates the bug that this is fixing (assuming there isn't one already open, if there is, please could you reference it)? Many thanks!

benwilliamgraham:
Sounds good, here's a new issue discussing it: https://github.com/numba/numba/issues/8574

guilhermeleobas:
Hi @benwilliamgraham, thanks for providing a test case. Can you update your patch to include this test case?

numba/tests/test_ufuncs.py
```diff
@@ -332,7 +332,10 @@ def test_mod_ufunc(self, flags=no_pyobj_flags):
             ])
 
     def test_fmod_ufunc(self, flags=no_pyobj_flags):
-        self.basic_ufunc_test(np.fmod, flags=flags)
+        self.basic_ufunc_test(np.fmod, flags=flags,
+            additional_inputs = [
+                ((np.int64(-4), np.int64(10)), types.uint64)
+            ])
```
benwilliamgraham, line 338:
```suggestion
                ((np.int64(-4), np.int64(10)), types.int64)
            ])
```

numba/np/npyfuncs.py
```diff
@@ -243,9 +243,35 @@ def np_int_udivrem_impl(context, builder, sig, args):
...
+    bb_no_if = builder.basic_block
+    with cgutils.if_unlikely(builder, den_not_zero):
+        bb_if = builder.basic_block
+
+        # use the abs of the numerator and denominator
+        num_neg = builder.icmp_signed("<", num, ZERO)
+        den_neg = builder.icmp_signed("<", den, ZERO)
+        num_abs = builder.select(num_neg, builder.neg(num), num)
+        den_abs = builder.select(den_neg, builder.neg(den), den)
+        mod = builder.urem(num_abs, den_abs)
```
guilhermeleobas, line 265: Rather than checking if the arguments are negative, can't you just use [`llvm.srem`](https://llvm.org/docs/LangRef.html#srem-instruction) instead?

benwilliamgraham: Good catch! I had previously misread the docs for `srem` and had thought it used the sign of the denominator. That'll simplify the code a bunch

stuartarchibald:
I think this needs addressing:
```python
from numba import njit
import numpy as np

@njit
def foo(x, y):
    return np.fmod(x, y)

a = np.uint32(20)
b = np.uint32(-1)
print(f"a = {a}, b = {b}")
print("got", foo(a, b))
print("exp", foo.py_func(a, b))
```
which currently gives:
```
a = 20, b = 4294967295
got 0
exp 20
```

guilhermeleobas:
The `ufunc_db[np.fmod]` table on numba/np/ufunc_db.py needs update as the `uint*` should be mapped now to `npyfuncs.np_int_urem_impl`

stuartarchibald:
> The `ufunc_db[np.fmod]` table on numba/np/ufunc_db.py needs update as the `uint*` should be mapped now to `npyfuncs.np_int_urem_impl`

Given the tests didn't catch this, do they need an update?

guilhermeleobas:
@benwilliamgraham, can you add a couple of tests for the test case reported by Stuart?

guilhermeleobas:
Hi @benwilliamgraham, I'm happy to continue this work if needed. Just let me know.

PR #8557, opened 2022-11-01
[Unicode] support startswith with args, start and end.

dlee992:
As it is, support `start` and `end` optional arguments for `startswith`.

Most added code comes from the implementation of `endswith`, with marginal changes.

PS, I also have aother PR about supporting `int(str)` based on this one, and I will open that one after merging this.

BTW, I have a question about `_nrt=False`, I didn't find any document about this, what is it for in the Numba Runtime level? Disable any memory allocation/refct/release?

### DIFF
--- a/numba/cpython/unicode.py
+++ b/numba/cpython/unicode.py
@@ -839,7 +839,7 @@ def unicode_startswith(s, prefix, start=None, end=None):
 
     if not is_nonelike(end) and not isinstance(end, types.Integer):
         raise TypingError(
-            "When specified, the arg 'end' must be an Integer or nonelike")
+            "When specified, the arg 'end' must be an Integer or None")
 
     if isinstance(prefix, (types.Tuple, types.UniTuple)):
         if isinstance(prefix, types.Tuple) or \
--- a/numba/cpython/unicode.py
+++ b/numba/cpython/unicode.py
@@ -835,7 +835,7 @@ def _adjust_indices(length, start, end):
 def unicode_startswith(s, prefix, start=None, end=None):
     if not is_nonelike(start) and not isinstance(start, types.Integer):
         raise TypingError(
-            "When specified, the arg 'start' must be an Integer or nonelike")
+            "When specified, the arg 'start' must be an Integer or None")
 
     if not is_nonelike(end) and not isinstance(end, types.Integer):
         raise TypingError(
--- a/numba/cpython/unicode.py
+++ b/numba/cpython/unicode.py
@@ -4,6 +4,7 @@ import operator
 import numpy as np
 from llvmlite.ir import IntType, Constant
 
+from numba.core.cgutils import is_nonelike
 from numba.core.extending import (
     models,
     register_model,
@@ -832,17 +833,20 @@ def _adjust_indices(length, start, end):
 
 @overload_method(types.UnicodeType, 'startswith')
 def unicode_startswith(s, prefix, start=None, end=None):
-    if not (start is None or isinstance(start, (types.Omitted,
-                                                types.Integer,
-                                                types.NoneType))):
-        raise TypingError('When specified, the arg "start" must be an Integer')
+    if not is_nonelike(start) and not isinstance(start, types.Integer):
+        raise TypingError(
+            "When specified, the arg 'start' must be an Integer or nonelike")
 
-    if not (end is None or isinstance(end, (types.Omitted,
-                                            types.Integer,
-                                            types.NoneType))):
-        raise TypingError('When specified, the arg "end" must be an Integer')
+    if not is_nonelike(end) and not isinstance(end, types.Integer):
+        raise TypingError(
+            "When specified, the arg 'end' must be an Integer or nonelike")
 
     if isinstance(prefix, (types.Tuple, types.UniTuple)):
+        if isinstance(prefix, types.Tuple) or \
+                not isinstance(prefix.dtype, types.UnicodeType):
+            raise TypingError(
+                "The arg 'prefix' should a UniTuple of UnicodeType")
+
         def startswith_tuple_impl(s, prefix, start=None, end=None):
             for item in prefix:
                 if s.startswith(item, start, end):
@@ -859,23 +863,22 @@ def unicode_startswith(s, prefix, start=None, end=None):
 
     if isinstance(prefix, types.UnicodeType):
         def startswith_unicode_impl(s, prefix, start=None, end=None):
-            length = len(s)
-            sub_length = len(prefix)
+            length, prefix_length = len(s), len(prefix)
             if start is None:
                 start = 0
             if end is None:
                 end = length
 
             start, end = _adjust_indices(length, start, end)
-            if end - start < sub_length:
+            if end - start < prefix_length:
                 return False
 
-            if sub_length == 0:
+            if prefix_length == 0:
                 return True
 
-            s = s[start:end]
+            s_slice = s[start:end]
 
-            return _cmp_region(s, 0, prefix, 0, sub_length) == 0
+            return _cmp_region(s_slice, 0, prefix, 0, prefix_length) == 0
 
         return startswith_unicode_impl
 
--- a/numba/cpython/unicode.py
+++ b/numba/cpython/unicode.py
@@ -841,11 +841,8 @@ def unicode_startswith(s, prefix, start=None, end=None):
         raise TypingError(
             "When specified, the arg 'end' must be an Integer or None")
 
-    if isinstance(prefix, (types.Tuple, types.UniTuple)):
-        if isinstance(prefix, types.Tuple) or \
-                not isinstance(prefix.dtype, types.UnicodeType):
-            raise TypingError(
-                "The arg 'prefix' should a UniTuple of UnicodeType")
+    if isinstance(prefix, types.UniTuple) and \
+            isinstance(prefix.dtype, types.UnicodeType):
 
         def startswith_tuple_impl(s, prefix, start=None, end=None):
             for item in prefix:
@@ -881,6 +878,10 @@ def unicode_startswith(s, prefix, start=None, end=None):
             return _cmp_region(s_slice, 0, prefix, 0, prefix_length) == 0
 
         return startswith_unicode_impl
+
+    else:
+        raise TypingError(
+            "The arg 'prefix' should be a string or a tuple of strings")
 
 
 @overload_method(types.UnicodeType, 'endswith')
--- a/numba/cpython/unicode.py
+++ b/numba/cpython/unicode.py
@@ -813,18 +813,6 @@ def unicode_rpartition(data, sep):
     return impl
 
 
-@overload_method(types.UnicodeType, 'startswith')
-def unicode_startswith(a, b):
-    if isinstance(b, types.UnicodeType):
-        def startswith_impl(a, b):
-            return _cmp_region(a, 0, b, 0, len(b)) == 0
-        return startswith_impl
-    if isinstance(b, types.UnicodeCharSeq):
-        def startswith_impl(a, b):
-            return a.startswith(str(b))
-        return startswith_impl
-
-
 # https://github.com/python/cpython/blob/201c8f79450628241574fba940e08107178dc3a5/Objects/unicodeobject.c#L9342-L9354    # noqa: E501
 @register_jitable
 def _adjust_indices(length, start, end):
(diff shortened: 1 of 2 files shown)


numba/cpython/unicode.py
```diff
@@ -842,6 +830,56 @@ def _adjust_indices(length, start, end):
     return start, end
 
 
+@overload_method(types.UnicodeType, 'startswith')
+def unicode_startswith(s, substr, start=None, end=None):
+    if not (start is None or isinstance(start, (types.Omitted,
+                                                types.Integer,
+                                                types.NoneType))):
+        raise TypingError('The arg must be a Integer or None')
```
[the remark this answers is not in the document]
dlee992: No problem. Added, wait for CI.

guilhermeleobas (changes requested):

guilhermeleobas:
I have checked that:

- [x]  There is no other Pull Request proposing the same change. (If there is
please highlight this and one of the core developers will pick it up.)
- [ ]  The proposed additional NumPy functionality is not deprecated by
NumPy.
- [x]  Public CI is passing. (With the exception of something unrelated to this PR)
- [x]  The patch does not touch files in `numba.core`.
- [x]  The patch is implemented using the high-level extension API only
([https://numba.readthedocs.io/en/stable/extending/high-level.html](https://numba.readthedocs.io/en/stable/extending/high-level.html)).
- [ ]  The patch updates the documentation to reflect the new functionality.
([https://github.com/numba/numba/blob/main/docs/source/reference/numpysupported.rst](https://github.com/numba/numba/blob/main/docs/source/reference/numpysupported.rst))
- [x]  The patch is line wrapped to 80 characters.
- [x]  Changes to Python source in the patch are `flake8` compliant.
- [x]  The unit tests for the implementation:
    - [x]  Explicitly check custom error messages are raised appropriately.
    - [x]  Hit all the branches of the proprosed change (both at type
    inference and run time).
- [x]  The patch works in a simple manual test in my local checkout.
- [x]  I cannot break the code in the patch in my local checkout after
trying to do so for a minimum of 10 minutes. The error messages seem
reasonable and there are no "incorrect" results.
- [x]  I think the code introduced in this change is easy to maintain.
- [x]  I understand all the code in this PR.
- [ ]  The implementation of the NumPy/CPython function reflects that present in the
NumPy implementation itself (look in the NumPy repository:
[https://github.com/numpy/numpy](https://github.com/numpy/numpy))
    - [ ]  The implementation has references to the region of the NumPy code
    base that provides the same functionality. The references are URLs
    of the form:
    `https://github.com/numpy/numpy/tree/<SHA>/path/to/file.py:#<lines>`
    - [ ]  The code "looks" reasonably similar to the NumPy/CPython implementation.
    - [ ]  The code uses the same algorithm as the NumPy/CPython implementation, if
    it does not it explains why and this justification is reasonable.
    - [x]  The type inference part of the Numba implementation restricts the
    types to those declared in the NumPy/CPython docs (or if not present, what
    is accepted in the NumPy/CPython implementation itself).
    - [x]  The Numba implementation uses the same variable names for
    arguments as in the NumPy/CPython implementation.
    - [ ]  The Numba code has commentary present for anything not-obvious by
    inspection.
    - [ ]  The Numba code accommodates differences that might occur across
    different NumPy versions in a clear fashion and uses
    `numba.np.numpy_support.numpy_version` to separate the difference
    in implementations.
- [x]  The unit tests for the implementation:
    - [x]  Use the same test inputs as the NumPy/CPython unit tests for the same
    functionality. If this is not the case there are comments which
    explain why and this justification is reasonable.
    - [ ]  The NumPy/CPython unit test inputs have reference URLs associated with
    them to track their origin. The references are URLs of the form:
    `https://github.com/numpy/numpy/tree/<SHA>/path/to/file.py:#<lines>`
    - [x]  Use `unittest` style assertions (not pytest style!).
    - [x]  Check that unsupported/invalid input types are handled as expected
    using the `with self.assertRaises()` pattern.
    - [ ]  Check that numerically awkward values work as expected e.g.
    `NaN`, `+/-Inf`, `0`, `INT_MAX`, `INT_MIN` etc.
    - [ ]  Check that types which can be converted to arrays (assuming NumPy
    accepts this behaviour) work as expected.
    - [ ]  Check that passing empty versions of containers e.g. empty tuple
    or 0d array works as expected.

numba/cpython/unicode.py
```diff
@@ -842,6 +830,56 @@ def _adjust_indices(length, start, end):
     return start, end
 
 
+@overload_method(types.UnicodeType, 'startswith')
+def unicode_startswith(s, substr, start=None, end=None):
```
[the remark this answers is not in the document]
dlee992: Sure. Keeping argument names consistent is a good habit.

numba/tests/test_unicode.py
```diff
@@ -551,23 +559,95 @@ def test_expandtabs_exception_noninteger_tabsize(self):
...
+        ]
+        extra_subs = ['hellohellohello', ' ']
+        for s in cpython_str + UNICODE_EXAMPLES:
+            default_subs = ['', 'x', s[:-2], s[3:], s, s + s]
+            for sub_str in cpython_subs + default_subs + extra_subs:
+                for start in list(range(-20, 20)) + [None]:
+                    self.assertEqual(pyfunc(s, sub_str, start),
+                                     cfunc(s, sub_str, start))
+        with self.assertRaises(TypingError):
+            cfunc('hello', 'he', 0.1)
```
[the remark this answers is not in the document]
dlee992: Okay. Revised as it should be.

dlee992:
Actually, the API `endswith` implementation contains similar "nitpicks" as my previous version for `startswith`. But I guess we could change that later, after this one is merged.

guilhermeleobas (approved):

numba/cpython/unicode.py
```diff
@@ -842,6 +830,56 @@ def _adjust_indices(length, start, end):
     return start, end
 
 
+@overload_method(types.UnicodeType, 'startswith')
+def unicode_startswith(s, prefix, start=None, end=None):
+    if not (start is None or isinstance(start, (types.Omitted,
+                                                types.Integer,
+                                                types.NoneType))):
+        raise TypingError('When specified, the arg "start" must be an Integer')
```
stuartarchibald, line 838: I'm not sure this error message is correct? `None` is accepted so it would be entirely legal for a user to write:
```python
'abcdef'.startswith('f', None)
```
which is not an integer.

numba/cpython/unicode.py
```diff
@@ -842,6 +830,56 @@ def _adjust_indices(length, start, end):
...
+
+    if not (end is None or isinstance(end, (types.Omitted,
+                                            types.Integer,
+                                            types.NoneType))):
+        raise TypingError('When specified, the arg "end" must be an Integer')
```
stuartarchibald, line 843: Same comment as above.

numba/cpython/unicode.py
```diff
@@ -842,6 +830,56 @@ def _adjust_indices(length, start, end):
     return start, end
 
 
+@overload_method(types.UnicodeType, 'startswith')
+def unicode_startswith(s, prefix, start=None, end=None):
+    if not (start is None or isinstance(start, (types.Omitted,
```
stuartarchibald, line 835: Should this use `numba.core.cgutils.is_nonelike` oppose to both checking for `None`, `types.NoneType` and `types.Omitted`?

numba/cpython/unicode.py
```diff
@@ -842,6 +830,56 @@ def _adjust_indices(length, start, end):
...
+                                                types.Integer,
+                                                types.NoneType))):
+        raise TypingError('When specified, the arg "start" must be an Integer')
+
+    if not (end is None or isinstance(end, (types.Omitted,
+                                            types.Integer,
+                                            types.NoneType))):
+        raise TypingError('When specified, the arg "end" must be an Integer')
+
+    if isinstance(prefix, (types.Tuple, types.UniTuple)):
```
stuartarchibald, line 845: `Tuple` cannot be accepted here unless the contents of the `Tuple` are all `types.StringLiteral`. Consider:

```python
@njit
def foo():
    print('abcde'.startswith(('a', 1)))

foo()
```
it may be easier to just not accept `Tuple` to force type inference to try with a non-literal types instead.

numba/cpython/unicode.py
```diff
@@ -842,6 +830,56 @@ def _adjust_indices(length, start, end):
...
+            start, end = _adjust_indices(length, start, end)
+            if end - start < sub_length:
+                return False
+
+            if sub_length == 0:
+                return True
+
+            s = s[start:end]
+
+            return _cmp_region(s, 0, prefix, 0, sub_length) == 0
```
stuartarchibald, line 878:
```suggestion
            s_slice = s[start:end]

            return _cmp_region(s_slice, 0, prefix, 0, sub_length) == 0
```
just to make it clear that this is addressing a slice of the original string.

stuartarchibald (changes requested):
Thanks for the patch. Few minor things to look at else looks good.

numba/cpython/unicode.py
dlee992: I prefer to accept `Tuple` of `StringLiteral`, 1) it's more consistent with cpython, 2) I do need this to simplify my next PR https://github.com/numba/numba/pull/8558. So I prefer to add a check to make sure every item in `Tuple` (or `UniTuple`) is a `StringLiteral`.

numba/cpython/unicode.py
dlee992: Revised as expected.

numba/cpython/unicode.py
dlee992: Revised using this func.

numba/cpython/unicode.py
dlee992: Revised, please review this issue again.

numba/cpython/unicode.py
dlee992: OK.

numba/cpython/unicode.py
```diff
@@ -831,51 +832,53 @@ def _adjust_indices(length, start, end):
...
+            "When specified, the arg 'end' must be an Integer or nonelike")
 
-    if isinstance(substr, (types.Tuple, types.UniTuple)):
-        def startswith_tuple_impl(s, substr, start=None, end=None):
-            for item in substr:
+    if isinstance(prefix, (types.Tuple, types.UniTuple)):
+        if isinstance(prefix, types.Tuple) or \
+                not isinstance(prefix.dtype, types.UnicodeType):
+            raise TypingError(
+                "The arg 'prefix' should a UniTuple of UnicodeType")
```
stuartarchibald, line 848: Just accept `types.UniTuple` with a `.dtype` which is `types.UnicodeType` ? e.g.

```python
if isinstance(prefix, types.UniTuple) and isinstance(prefix.dtype, types.UnicodeType):
    ... # then return the implementation
```

numba/cpython/unicode.py
```diff
@@ -842,6 +831,58 @@ def _adjust_indices(length, start, end):
...
+
+        def startswith_tuple_impl(s, prefix, start=None, end=None):
+            for item in prefix:
+                if s.startswith(item, start, end):
+                    return True
+            return False
+
+        return startswith_tuple_impl
+
+    if isinstance(prefix, types.UnicodeCharSeq):
```
stuartarchibald, line 858:
```suggestion
    elif isinstance(prefix, types.UnicodeCharSeq):
```

numba/cpython/unicode.py
```diff
@@ -842,6 +831,58 @@ def _adjust_indices(length, start, end):
...
+        def startswith_char_seq_impl(s, prefix, start=None, end=None):
+            return s.startswith(str(prefix), start, end)
+
+        return startswith_char_seq_impl
+
+    if isinstance(prefix, types.UnicodeType):
```
stuartarchibald, line 864:
```suggestion
    elif isinstance(prefix, types.UnicodeType):
```

numba/cpython/unicode.py
```diff
@@ -842,6 +831,58 @@ def _adjust_indices(length, start, end):
...
+
+            if prefix_length == 0:
+                return True
+
+            s_slice = s[start:end]
+
+            return _cmp_region(s_slice, 0, prefix, 0, prefix_length) == 0
+
+        return startswith_unicode_impl
+
```
stuartarchibald, line 884: I think there should be an `else` here for the case where `prefix` is of an unsupported type, which raises:
```python
    else:
        raise TypingError("The arg 'prefix' should be a string or a tuple of strings")
```

stuartarchibald (commented):
Thanks for the update @dlee992. I've made a few suggestions about restructuring the type checking with respect to the type of `prefix` (and a few other minor things). I think once resolved this should be ready to merge. 

Note: I'll answer your questions on the diff inline now so as to not get them confused with this review.

numba/cpython/unicode.py
stuartarchibald: By default, type inference will try using non-literal types first, and then if that fails, it will try again using literal types. Python source containing tuples of string literals will be inferred as `types.Tuple(<types.StringLiteral>, <types.StringLiteral>, ...)`. e.g.
```python
from numba import njit
import numpy as np

@njit
def foo():
    x = ('a', 'b')

foo()
foo.inspect_types()
```
-------------------------------------
You can see the "try non-literal then literal" types behaviour by doing something like:
```python
from numba import njit, types
import numpy as np

def bar(tup):
    pass

@overload(bar)
def ol_bar(tup):
    print(f"bar:: tup type is {tup}")

@njit
def foo():
    x = ('a', 'b')
    bar(x)

foo()
```
(ignore the compilation error, just look at what gets printed to stdout).

-------------------------------------
RE: the `'a'.startswith(('a', 5))`, there are implementation details like this in the CPython implementation. They don't need replicating unless there's some proven use case!

stuartarchibald (approved):

PR #8558, opened 2022-11-01
[Unicode] support `int(str, base=10)` as cpython does

### LINKED ISSUE #5650: Unable to create integer from unicode string with int()
I am trying to convert a unicode string to an integer like this:

```
import numba

@numba.jit(nopython=True)
def str_to_int(str_date):
	return int(str_date[0:4])

print(str_to_int("2016 foo bar"))
```

and receive the following error:

```Invalid use of Function(<class 'int'>) with argument(s) of type(s): (unicode_type)
 * parameterized
In definition 0:
In definition 1:
[1] During: resolving callee type: Function(<class 'int'>)
[2] During: typing of call at numba-bug.py (5)


File "numba-bug.py", line 5:
def str_to_int(str_date):
	return int(str_date[0:4])```


I have also tried to replace the `int(str_date[0:4])` with `int(str_date[0]) * 1000 + int(str_date[1]) * 100 + int(str_date[2]) * 10 + int(str_date[3])` but the same error appears.

Is this a real bug or am I using the library wrong? Unfortunately I was unable to find more info on substring handling in the docs. Thank you!

### LINKED ISSUE #5723: Overload float(), int() and str() to do number <-> string conversion
So currently all three of these jitted functions will fail

```python
@njit
def failed_str():
	str(1.2)
@njit
def failed_float():
	float("1.2")
@njit
def failed_int():
	int("1.2")
```
I've tried re-implementing these and overloading them.
Strings -> Numbers
```python
@njit
def str_to_posint(s):
	final_index, result = len(s) - 1, 0
	for i,v in enumerate(s):
		result += (ord(v) - 48) * (10 ** (final_index - i))
	return result

@njit
def str_to_int(s):
	neg = (s[0] == "-")
	if(neg): s = s[1:]
	result = str_to_posint(s)
	return -result if neg else result

@njit
def str_to_float(s):
	neg = (s[0] == "-")
	if(neg): s = s[1:]

	exp,dec_loc,exp_loc = 1,-1, len(s)
	for i,c in enumerate(s):
		if(c == "."): 
			dec_loc=i
		elif(c == "e"):
			exp_loc=i
			exp = str_to_int(s[exp_loc+1:])

	if(dec_loc != -1):
		result = str_to_posint(s[:dec_loc])
		result += float(str_to_posint(s[dec_loc+1:exp_loc])) / np.power(10,(exp_loc-dec_loc-1))
	else:
		result = str_to_posint(s)

	if(exp != 1):
		result *= 10.**exp

	if(neg): result = -result
	return result
``` 
Numbers -> Strings
```python
@overload(str)
def _str(x):
	if isinstance(x,types.Integer):
		def int_to_str(x):
			isneg = int(x < 0.0)
			x = np.abs(x)
			l = 0 
			_x = x
			while _x > 0:
				_x = _x // 10
				l += 1
			s = _empty_string(PY_UNICODE_1BYTE_KIND,l+isneg)
			if(isneg): _set_code_point(s,0,DASH)
			for i in range(l):
				digit = x % 10
				_set_code_point(s,isneg+l-i-1,digit + DIGITS_START)
				x = x // 10
			return s
		return int_to_str
	elif isinstance(x,types.Float):
		def float_to_str(x):
			if(x == np.inf):
				return 'inf'
			elif(x == -np.inf):
				return '-inf'

			isneg = int(x < 0.0)
			x = np.abs(x)
			l1,l2 = 0,-1
			_x = x
			while _x > 0:
				_x = _x // 10
				l1 += 1
			_x = x % 10 
			while _x > 1e-10:
				_x = (_x * 10) % 10
				l2 += 1

			l2 = max(1,l2)
			l = l1+l2+1

			_x = x-np.floor(x)
			s = _empty_string(PY_UNICODE_1BYTE_KIND,l+isneg)
			if(isneg): _set_code_point(s,0,DASH)


			_x = x
			for i in range(l1):
				digit = int(_x % 10)
				# print('a',digit,(isneg+l1)-i-1)
				_set_code_point(s,(isneg+l1)-i-1,digit + DIGITS_START)
				_x = _x // 10

			# if(l2 > 0):
			_set_code_point(s,l1+isneg,DOT)

			_x = x % 10 
			for i in range(l2):
				_x = (_x * 10) % 10
				digit = int(_x)  
				# print('here',i,digit)
				# print('b',digit,(isneg+l1)+i+1)
				_set_code_point(s,(isneg+l1)+i+1,digit + DIGITS_START)
			return s
		return float_to_str
```

But it seems that just letting numba loop lift the native python is pretty consistently faster than anything that I can write.

```python
import timeit
N=1000
def time_ms(f):
		f() #warm start
		return " %0.6f ms" % (1000.0*(timeit.timeit(f, number=N)/float(N)))

@njit
def numba_to_str(x):
	return str(x)

def py_to_str(x):
	return str(x)

from warnings import catch_warnings

with catch_warnings(record=True) as w:
	@jit
	def nb_lotsa_strings():
		for i in range(1000):
			numba_to_str(14.58)

	@jit
	def nb_lotsa_strings_int():
		for i in range(1000):
			numba_to_str(14)

	@jit
	def py_lotsa_strings():
		for i in range(1000):
			py_to_str(14.58)

	@jit
	def py_lotsa_strings_int():
		for i in range(1000):
			py_to_str(14)

	print("nb_to_str",time_ms(nb_lotsa_strings))
	print("nb_int_to_str",time_ms(nb_lotsa_strings_int))
	print("py_to_str",time_ms(py_lotsa_strings))
	print("py_int_to_str",time_ms(py_lotsa_strings_int))
```
```bash
----numba---
nb_to_str  1.287220 ms
nb_int_to_str  0.047568 ms
---python---
py_to_str  0.424743 ms
py_int_to_str  0.246021 ms

```

So my feature request is can str(int/float), float(str) and int(str) just be overloaded to grab the GIL and just call native python? Or might there be a way for these to just call out directly to the cpython implementations?

This is related to #5650 and #4758

dlee992:
I refer to https://docs.python.org/3/reference/lexical_analysis.html#integer-literals, and https://docs.python.org/3/library/functions.html#int. 
But I'm not 100% sure about the consistency between my numba impl. and python impl.

I found I don't have to wait for another PR, after some revisions...

Closes #5650 
And also might close #5723


dlee992:
Hi, @esc. I feel this one can be marked as ready for the initial review too.

dlee992:
Hi, @stuartarchibald , @guilhermeleobas , this one that handles `int(str, base=10)` can be reviewed too.

dlee992:
Hi, @stuartarchibald @guilhermeleobas , this one is ready for an initial review, I guess.

dlee992:
Yes, you are absolutely right. I don't know why I didn't notice these requirements, even I have already posted the link... I will revise this ASAP.

dlee992:
I have updated this PR according python docs and cpython source code in `listobject.c` and `unicodeobject.c`. And I will add similar testcases borrowed from `test_int.py` in cpython, which contains a huge amount of various tests.

Besides, since Python has a `big-int` concept, while numba only has fixed-bit int, long,  as C language. So I simplify the conversion algorithm used in cpython function `long_from_binary_base` and especially in `long_from_non_binary_base`. I am not sure if the conversion performance is degrading too much, but since numba can only hold a 32- or 64-bit of int/long, I guess the performance won't change too much.

dlee992:
> 1. The argument must be a string, bytes or bytearray instance representing an integer literal in the given base.
> 2. The literal can be preceded by `+` or `-`
> 3. The literal can be surrounded by `whitespace`.
> 4. The default base is 10.
> 5. Valid bases are 0 and 2-36, base 0 means interpret the base from the string.

After full changes, the requirements of 2, 3, 4, 5 are met nicely. About the first one, right now I only support the argument is a `string`, I guess we can defer the supporting for other types later.

About if the string literal overflows after converting to int, I didn't check this restriction, except that I check a condition, `digits > 63`, to prevent the digits in the string is too big to store in a `int64`.

And I add many testcases copied or modified from `test_int.py` in cpython, and remove non `string-to-int` conversion testcases. I think @stuartarchibald  or others interested in this PR can take a second review, thanks.

dlee992:
Hi, @stuartarchibald , @sklam do u have some time to re-review this? Many thanks.

dlee992:
For FYI, 

1. this is the python blog in 2006, describing how they implemented the exact BM algorithm in that period of Python: https://web.archive.org/web/20201107074620/http://effbot.org/zone/stringlib.htm
2. this is about all kinds of string fast matching algorithms: http://www-igm.univ-mlv.fr/~lecroq/string/index.html

stuartarchibald:
@dlee992 I edited the PR description to note that this closes #5650.

dlee992:
Hi, @stuartarchibald and @guilhermeleobas .

This PR is ready for second review after my fully revised changes, if u guys have some time for it. Thanks!

docs/source/reference/pysupported.rst
```diff
@@ -934,7 +934,7 @@ The following built-in functions are supported:
   cannot be a function type (e.g. ``getattr(numpy, 'cos')`` is not supported as
   it returns a function type).
 * :func:`hash` (see :ref:`pysupported-hashing` below)
-* :class:`int`: only the one-argument form
+* :class:`int`
```
sklam, line 937: Can you add the implementation limitations here? e.g. only support `str` argument.

numba/cpython/unicode.py
```diff
@@ -31,37 +31,20 @@
...
+    _PyUnicode_IsXidStart, _PyUnicode_IsXidContinue, _PyUnicode_IsCased,
+    _PyUnicode_IsCaseIgnorable, _PyUnicode_IsUppercase, _PyUnicode_IsLowercase,
+    _PyUnicode_IsLineBreak, _Py_ISLINEBREAK, _Py_ISLINEFEED,
+    _Py_ISCARRIAGERETURN, _PyUnicode_IsTitlecase, _Py_ISLOWER, _Py_ISUPPER,
+    _Py_TAB, _Py_LINEFEED, _Py_CARRIAGE_RETURN, _Py_SPACE, _PyUnicode_IsAlpha,
+    _PyUnicode_IsNumeric, _Py_ISALPHA, _PyUnicode_IsDigit,
+    _PyUnicode_IsDecimalDigit, _Ord_Zero, _PyUnicode_ToDecimalDigit, _Ord_Plus,
+    _Ord_Minus, _Ord_x, _Ord_o, _Ord_b, _Ord_B, _Ord_O, _Ord_X, _Ord_Underscore,
+    _PyLong_DigitValue, _Py_CHARMASK,
+)
```
guilhermeleobas, line 47: Can you undo this change, as it is not related to the main patch?

numba/tests/test_unicode.py
```diff
@@ -2701,5 +2710,347 @@ def impl5():
...
+            for sign in "", "+", "-":
+                for prefix in "", " ", "\t", "  \t\t  ":
+                    ss = prefix + sign + s
+                    vv = v
+                    if sign == "-" and v is not ValueError:
+                        vv = -v
+                    try:
+                        self.assertEqual(cfunc(ss), vv)
+                    except ValueError:
+                        pass
```
guilhermeleobas, line 2931: For the cases `cfunc(ss)` raises an exception, can you check the exception message to see if it is what you're expecting.

numba/core/typing/builtins.py
```diff
@@ -954,7 +954,7 @@ def generic(self, args, kws):
         if kws:
             raise errors.NumbaAssertionError('kws not supported')
 
-        [arg] = args
+        arg = args[0]
```
guilhermeleobas, line 957: Can you add a comment here saying that support for `int(str)` is done via `@overload`

guilhermeleobas (changes requested):
Hi @dlee992, thanks for working on this.

It looks like you covered everything in the implementation. I tried it locally and it is working as expected. I only have a couple of comments that should be easy to address

numba/cpython/unicode.py
guilhermeleobas: That's fine. I thought the change was just formatting the lines.

numba/tests/test_unicode.py
dlee992: This one is tricky, I will try to revise it.

dlee992:
Hi, I have revised this patch following your advice. Please re-review this again. And Can you put this one into `0.57` as well?

guilhermeleobas (dismissed):
Hi @dlee992, I'll label your PR as waiting for second review as Stuart was reviewing it before.

dlee992:
Hi, @stuartarchibald , this one can use your thoughtful second review!

stuartarchibald:
> Hi, @stuartarchibald , this one can use your thoughtful second review!

@dlee992 I've made some progress on reviewing this (mostly looked at the implementation), there's a lot of involved code (esp. C pointer arithmetic to Python equivalent) so I'm afraid it's taking a while. I hope to find some more time to complete this next week. Thanks in advance for your understanding and patience.

dlee992:
> > Hi, @stuartarchibald , this one can use your thoughtful second review!
> 
> @dlee992 I've made some progress on reviewing this (mostly looked at the implementation), there's a lot of involved code (esp. C pointer arithmetic to Python equivalent) so I'm afraid it's taking a while. I hope to find some more time to complete this next week. Thanks in advance for your understanding and patience.

Thanks! Looking forward to your further comments! If u have any question about the implementation, please discuss with me at any time.

numba/cpython/unicode.py
```diff
@@ -2470,6 +2453,219 @@ def boolean_str(b):
         return lambda b: "True" if b else "False"
 
 
+# WARNING: Integer in [-1<<63, 1<<63) is supported, which differs from cpython.
+# Global error messages for conversion (unicode -> integer)
+_INVALID_LITERAL_MSG = "invalid literal for int() with base"
```
stuartarchibald, line 2458: This exception string appears even when used in a case where no `base` is present, which is a little confusing. For example:
```python
@njit
def foo():
    return int('a') # there is no base given
```
It would be good to differentiate between the case of "no base" and "base given but not compatible".

stuartarchibald: Same as above applies for a case like:
```python
@njit
def foo():
    return int('1', -2) # invalid base
```
would be good to have a more explicit message about what the problem is (i.e. base is out of supported range).

docs/source/reference/pysupported.rst
```diff
@@ -934,7 +934,8 @@ The following built-in functions are supported:
   cannot be a function type (e.g. ``getattr(numpy, 'cos')`` is not supported as
   it returns a function type).
 * :func:`hash` (see :ref:`pysupported-hashing` below)
-* :class:`int`: only the one-argument form
+* :class:`int`: support `int`, `float`, or `str` as the first argument,
+  and support an optional `base` argument with `str`
```
stuartarchibald, line 938: Depending on how `int` overflow ends up being resolved in this PR, the docs might need associated limitations noting.

stuartarchibald (commented):
Thanks for the updates to this PR @dlee992, it's great to see so many involved unit tests derived from CPython's test suite passing! I've reviewed the changes, most of the outstanding things are small changes, but there are a few questions about resolving TODO's and algorithmic differences that need addressing. Thanks again for working on this!

PR #8560, opened 2022-11-01
CUDA: Get rid of `_nvvm_options_type`

gmarkall:
This isn't really necessary - it only existed because there were sometimes no options as `None` instead of an empty dict. This commit ensures it is always a dict and uses the standard dict type for it.

### DIFF
--- a/numba/cuda/compiler.py
+++ b/numba/cuda/compiler.py
@@ -16,7 +16,6 @@ from numba.cuda.api import get_current_device
 class CUDAFlags(Flags):
     nvvm_options = Option(
         type=dict,
-        default=None,
         doc="NVVM options",
     )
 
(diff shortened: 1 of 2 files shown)


gmarkall:
gpuci run tests

numba/cuda/compiler.py
```diff
@@ -13,18 +13,9 @@
...
-    else:
-        assert isinstance(x, dict)
-        return x
-
-
 class CUDAFlags(Flags):
     nvvm_options = Option(
-        type=_nvvm_options_type,
+        type=dict,
         default=None,
```
gmarkall, line 19: Need to check if default can be removed

gmarkall:
@stuartarchibald I was able to remove the default by providing a default of `None` for `default` in the `Option` class. This feels a little better to me from the point of view of the declaration of an option, where omitting `default=` implies to the reader that there is no default. What do you think?

gmarkall:
gpuci run tests

gmarkall:
I think this is a bit of a bikeshed refactor and not worth prioritising.

PR #8552, opened 2022-10-31
Update version support table for 0.56.4.

stuartarchibald:
As title.

PR #8549, opened 2022-10-28
Fix rstcheck in Azure CI builds, update sphinx dep and docs to match

stuartarchibald:
As title.

Issue #8548, opened 2022-10-28
Azure CI builds of `main` failing on `rstcheck`


stuartarchibald:
As of 598aadf, `main` is failing to build in Azure CI in Linux build job configuration `py38_np120_doc`. The `Test` step fails with:
```
# Ensure the README is correctly formatted
if [ "$BUILD_DOC" == "yes" ]; then rstcheck README.rst; fi
Traceback (most recent call last):
  File "/home/vsts/miniconda3/envs/azure_ci/bin/rstcheck", line 5, in <module>
    from rstcheck._cli import main
  File "/home/vsts/miniconda3/envs/azure_ci/lib/python3.8/site-packages/rstcheck/_cli.py", line 7, in <module>
    from rstcheck_core import _extras, config as config_mod, runner
  File "/home/vsts/miniconda3/envs/azure_ci/lib/python3.8/site-packages/rstcheck_core/runner.py", line 10, in <module>
    from . import _sphinx, checker, config, types
  File "/home/vsts/miniconda3/envs/azure_ci/lib/python3.8/site-packages/rstcheck_core/_sphinx.py", line 14, in <module>
    import sphinx.application
  File "/home/vsts/miniconda3/envs/azure_ci/lib/python3.8/site-packages/sphinx/application.py", line 42, in <module>
    from sphinx.highlighting import lexer_classes, lexers
  File "/home/vsts/miniconda3/envs/azure_ci/lib/python3.8/site-packages/sphinx/highlighting.py", line 30, in <module>
    from sphinx.ext import doctest
  File "/home/vsts/miniconda3/envs/azure_ci/lib/python3.8/site-packages/sphinx/ext/doctest.py", line 28, in <module>
    from sphinx.builders import Builder
  File "/home/vsts/miniconda3/envs/azure_ci/lib/python3.8/site-packages/sphinx/builders/__init__.py", line 24, in <module>
    from sphinx.io import read_doc
  File "/home/vsts/miniconda3/envs/azure_ci/lib/python3.8/site-packages/sphinx/io.py", line 42, in <module>
    from sphinx.util.rst import append_epilog, docinfo_re, prepend_prolog
  File "/home/vsts/miniconda3/envs/azure_ci/lib/python3.8/site-packages/sphinx/util/rst.py", line 22, in <module>
    from jinja2 import environmentfilter
[...]

PR #8562, opened 2022-11-01
Cherry picks 0.56.4

esc:
As Title


stuartarchibald (approved):
Thanks for the patch. I replayed the merges from #8537, #8552 and #8553 manually onto a local copy of the `release0.56` branch and the `git diff` against this PR is empty, which suggests it is correct. Manually inspecting the changes, they look as expected. The RTD build looks ok on manual inspection too.

stuartarchibald:
NOTE: The build failure on the `Linux py39_np120_doc` build is due to #8548 which is fixed on branch `main` by #8549 but not on target branch for this PR (branch `release0.56`). The issue is `rstcheck` related which only occurs on Azure CI, hence RTD and the buildfarm should be ok.

esc:
@stuartarchibald thank you for the review!

gmarkall:
gpuci run tests

PR #8570, opened 2022-11-02
Release 0.56 branch: Fix overloads with `target="generic"` for CUDA

gmarkall:
As identified in #8271, the CUDA target needs to be set as the target at the bottom of the call stack, otherwise overloads for the generic target cannot be resolved. This is required so that the fix applied in #8562 (using the generic target for `ol_compatible_view` from #8537) actually works.

### DIFF
--- a/numba/cuda/compiler.py
+++ b/numba/cuda/compiler.py
@@ -207,14 +207,16 @@ def compile_cuda(pyfunc, return_type, args, debug=False, lineinfo=False,
         flags.nvvm_options = nvvm_options
 
     # Run compilation pipeline
-    cres = compiler.compile_extra(typingctx=typingctx,
-                                  targetctx=targetctx,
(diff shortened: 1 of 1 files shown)


stuartarchibald (approved):
Thanks for the patch, I've tested this locally and it fixes the noted issue. I think it also makes sense that this would be needed as the path that's failing is quite likely unaware of what the current target is as it has CUDA lowering calling typing and lowering on a `'generic'` target.

esc:
`Build numba_smoketest_cuda_yaml_167 has started`

esc:
> `Build numba_smoketest_cuda_yaml_167 has started` `Build numba_smoketest_cpu_yaml_147 has started`

These were all green.

esc (approved):
Checked via CI

PR #8572, opened 2022-11-02
CUDA: Reduce memory pressure from local memory tests

gmarkall:
Local memory usage requires a reservation of (bytes per thread) * (number of SMs) * (Max resident threads per SM), which for kernels with local memory requirements of 80KB per thread (as is the case for these tests) forces allocations into the low GBs - this is a problem on the buildfarm as multiple tests run concurrently and can result in out-of-memory conditions.

This commit reduces the local memory allocations by a factor of 10 - I worry that reducing by another order of magnitude might cause local memory not to be allocated in some cases.

Some copy/paste errors that slipped through in these tests are also resolved (including the name of one of the tests!).

### DIFF
--- a/numba/cuda/tests/cudapy/test_dispatcher.py
+++ b/numba/cuda/tests/cudapy/test_dispatcher.py
@@ -377,11 +377,11 @@ class TestDispatcher(CUDATestCase):
         self.assertIsInstance(shared_mem_per_block, int)
         self.assertEqual(shared_mem_per_block, 400)
 
-    def test_get_local_mem_per_block_unspecialized(self):
+    def test_get_local_mem_per_thread_unspecialized(self):
         # NOTE: A large amount of local memory must be allocated
         # otherwise the compiler will optimize out the call to
         # cuda.local.array and use local registers instead
-        N = 10000
+        N = 1000
 
         @cuda.jit
         def simple_lmem(ary):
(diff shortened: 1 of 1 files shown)


gmarkall:
gpuci run tests

numba/cuda/tests/cudapy/test_dispatcher.py
```diff
@@ -412,16 +412,16 @@ def simple_lmem(ary):
...
         # provides the same values as getting the shared mem per block for
         # individual signatures.
-        sh_mem_f32_all = simple_lmem.get_local_mem_per_thread()
-        sh_mem_f64_all = simple_lmem.get_local_mem_per_thread()
-        self.assertEqual(sh_mem_f32_all[sig_f32.args], local_mem_f32)
-        self.assertEqual(sh_mem_f64_all[sig_f64.args], local_mem_f64)
+        local_mem_f32_all = simple_lmem.get_local_mem_per_thread()
+        local_mem_f64_all = simple_lmem.get_local_mem_per_thread()
+        self.assertEqual(local_mem_f32_all[sig_f32.args], local_mem_f32)
+        self.assertEqual(local_mem_f64_all[sig_f64.args], local_mem_f64)
```
stuartarchibald, line 418:
```suggestion
        local_mem_all = simple_lmem.get_local_mem_per_thread()
        self.assertEqual(local_mem_all[sig_f32.args], local_mem_f32)
        self.assertEqual(local_mem_all[sig_f64.args], local_mem_f64)
```
Is this just the same dictionary being assigned to different variables, in which case the above should be equivalent?

stuartarchibald (commented):
Thanks for the patch. One minor thing to resolve else looks good. The main test of this will be in CI to see if the resource pressure is reduced when the tests are run in a multi-process environment against the same device(s).

numba/cuda/tests/cudapy/test_dispatcher.py
gmarkall: Whoops, yes.

gmarkall:
gpuci run tests

gmarkall:
@stuartarchibald Many thanks for the review, suggestion committed.

stuartarchibald (approved):

esc:
`Build numba_smoketest_cuda_yaml_170 has started`

stuartarchibald:
> `Build numba_smoketest_cuda_yaml_170 has started`

Passed.

PR #8575, opened 2022-11-02
Migrate `cmath.*` to use overload

guilhermeleobas:
This is a follow-up of #8494 and migrate the rest of cmath functions to use overloads.

### DIFF
--- a/numba/core/typing/cmathdecl.py
+++ b/numba/core/typing/cmathdecl.py
@@ -16,7 +16,6 @@ infer_global = registry.register_global
 @infer_global(cmath.asinh)
 @infer_global(cmath.atan)
 @infer_global(cmath.atanh)
-@infer_global(cmath.cos)
 @infer_global(cmath.cosh)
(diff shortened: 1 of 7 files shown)


gmarkall:
gpuci run tests

guilhermeleobas:
Small reproducer for the CI failure:

```python

llvm.set_option('', '--debug-only=loop-vectorize')

from numba import njit
import numpy as np


@njit('complex64[:](int64)', fastmath={'fast'}, error_model='numpy', no_cfunc_wrapper=True, no_cpython_wrapper=True)
def fn(n):
    x = np.empty(n*8, dtype=np.complex64)
    return np.exp(x)


pattern = ('__svml_expf',)
asm = fn.inspect_asm(fn.signatures[0])

for p in pattern:
    print(f"{p} in asm? {p in asm}")
```

guilhermeleobas:
I've removed cmath.exp from this PR and will create a new one to track the current failure.

Issue #6513, opened 2020-11-23
Can't compile recursive function ahead of time


nubol23:
When trying to compile a simple recursive function

```python
@cc.export('flood_fill', 'b1(u1[:,:], i8, i8, f8[:])')
def flood_fill(mat, i, j, coord):
    if mat[i, j] == 0:
        return
    
    directions = np.asarray([(-1, -1), (0, -1), (1, -1), (-1, 0), (1, 0), (-1, 1), (0, 1), (1, 1)])
    
    mat[i, j] = 0
    coord[0] = min(coord[0], j)
    coord[1] = min(coord[1], i)
    coord[2] = max(coord[2], j)
    coord[3] = max(coord[3], i)
    
    for dx, dy in directions:
        flood_fill(mat, i+dy, j+dx, coord)
        
    return True
```

i get

```bash
Untyped global name 'flood_fill': cannot determine Numba type of <class 'function'>
```

with the arrow pointing at the recursive call, added a dummy boolean return to ensure explicit typing of the function but still the error persists.

gmarkall:
I can reproduce the issue with the following:

```python
from numba.pycc import CC
import numpy as np


cc = CC('repro')


@cc.export('flood_fill', 'b1(u1[:,:], i8, i8, f8[:])')
def flood_fill(mat, i, j, coord):
    if mat[i, j] == 0:
        return False

    directions = np.asarray([(-1, -1), (0, -1), (1, -1), (-1, 0), (1, 0),
                             (-1, 1), (0, 1), (1, 1)])

    mat[i, j] = 0
    coord[0] = min(coord[0], j)
    coord[1] = min(coord[1], i)
    coord[2] = max(coord[2], j)
    coord[3] = max(coord[3], i)

    for dx, dy in directions:
        flood_fill(mat, i+dy, j+dx, coord)

    return True


if __name__ == '__main__':
    cc.compile()
```

(note that the original was inconsistent in its return type - the `return` statement returned `None`, whereas the declared return type is `bool`, so I edited that to `return False`).

Compiling the function with `@njit` works as expected.

Full error with AOT compilation:

```
Traceback (most recent call last):
  File "repro.py", line 29, in <module>
    cc.compile()
  File "/home/gmarkall/numbadev/numba/numba/core/compiler_lock.py", line 32, in _acquire_compile_lock
    return func(*args, **kwargs)
  File "/home/gmarkall/numbadev/numba/numba/pycc/cc.py", line 214, in compile
[...]

PR #8577, opened 2022-11-03
Document that AOT compilation of recursive functions is unsupported

gmarkall:
See Issue #6513 - also this popped up on Discourse recently: https://numba.discourse.group/t/how-to-compile-ahead-of-time-a-recursive-function/1630

### DIFF
--- a/docs/source/user/pycc.rst
+++ b/docs/source/user/pycc.rst
@@ -33,6 +33,8 @@ Limitations
 
 #. AOT compilation only allows for regular functions, not :term:`ufuncs <ufunc>`.
 
+#. Compilation of recursive functions is unsupported.
+
 #. You have to specify function signatures explicitly.
 
 #. Each exported function can have only one signature (but you can export


stuartarchibald:
I think the issue is that `pycc` needs to have a completely resolvable call graph (and recursion isn't handled by `pycc` compilation path). There is a work around though, defer the recursion part to a compilation path that supports it i.e.:
```python
from numba import njit
import numpy as np
from numba.pycc import CC
cc = CC('fib')
cc.verbose = True

@cc.export('exported_name', 'i8(i8)')
def wrap(n):
    return fib(n)

@njit
def fib(n):
    if n < 2:
        return n
    return fib(n - 1) + fib(n - 2)


if __name__ == "__main__":
    cc.compile()
```

stuartarchibald (commented):
Thanks for the patch. Noting this limition is a good idea given a there have been a few queries in relation to it. Is it worth adding an example of the workaround https://github.com/numba/numba/pull/8577#issuecomment-1302102477 too?

PR #8579, opened 2022-11-04
CUDA: Add CUDA 11.8 / Hopper support and required fixes

gmarkall:
See individual commit messages for full descriptions. The required fixes consist of:

- Correcting the logic for determining supported compute capabilities, which was completely broken on `main`.
- Using the correct NVVM IR version in all tests.

### DIFF
--- a/numba/cuda/cudadrv/nvvm.py
+++ b/numba/cuda/cudadrv/nvvm.py
@@ -380,7 +380,8 @@ def ccs_supported_by_ctk(ctk_version):
     except KeyError:
         # For unsupported CUDA toolkit versions, all we can do is assume all
         # non-deprecated versions we are aware of are supported.
-        return tuple([cc for cc in COMPUTE_CAPABILITIES if cc >= (5, 3)])
+        return tuple([cc for cc in COMPUTE_CAPABILITIES
+                      if cc >= config.CUDA_DEFAULT_PTX_CC])
 
 
 def get_supported_ccs():
@@ -394,11 +395,13 @@ def get_supported_ccs():
         return _supported_cc
 
     # Ensure the minimum CTK version requirement is met
-    if cudart_version < (11, 0):
+    min_cudart = min(CTK_SUPPORTED)
+    if cudart_version < min_cudart:
         _supported_cc = ()
         ctk_ver = f"{cudart_version[0]}.{cudart_version[1]}"
         unsupported_ver = (f"CUDA Toolkit {ctk_ver} is unsupported by Numba - "
-                           "11.0 is the minimum required version.")
+                           f"{min_cudart[0]}.{min_cudart[1]} is the minimum "
+                           "required version.")
         warnings.warn(unsupported_ver)
         return _supported_cc
 
--- a/numba/cuda/cudadrv/nvvm.py
+++ b/numba/cuda/cudadrv/nvvm.py
@@ -343,51 +343,69 @@ class CompilationUnit(object):
         return ''
 
 
+COMPUTE_CAPABILITIES = (
+    (3, 5), (3, 7),
+    (5, 0), (5, 2), (5, 3),
+    (6, 0), (6, 1), (6, 2),
+    (7, 0), (7, 2), (7, 5),
+    (8, 0), (8, 6), (8, 7), (8, 9),
+    (9, 0)
+)
+
+# Maps CTK version -> (min supported cc, max supported cc)
+CTK_SUPPORTED = {
+    (11, 0): ((3, 5), (8, 0)),
+    (11, 1): ((3, 5), (8, 6)),
+    (11, 2): ((3, 5), (8, 6)),
+    (11, 3): ((3, 5), (8, 6)),
+    (11, 4): ((3, 5), (8, 7)),
+    (11, 5): ((3, 5), (8, 7)),
+    (11, 6): ((3, 5), (8, 7)),
+    (11, 7): ((3, 5), (8, 7)),
+    (11, 8): ((3, 5), (9, 0)),
+}
+
+
+def ccs_supported_by_ctk(ctk_version):
+    try:
+        # For supported versions, we look up the range of supported CCs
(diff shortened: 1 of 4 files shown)


gmarkall:
gpuci run tests

numba/cuda/cudadrv/nvvm.py
```diff
@@ -343,51 +343,69 @@ def get_log(self):
...
+COMPUTE_CAPABILITIES = (
+    (3, 5), (3, 7),
+    (5, 0), (5, 2), (5, 3),
+    (6, 0), (6, 1), (6, 2),
+    (7, 0), (7, 2), (7, 5),
+    (8, 0), (8, 6), (8, 7), (8, 9),
+    (9, 0)
+)
+
+# Maps CTK version -> (min supported cc, max supported cc)
```
stuartarchibald, line 355: Are the bounds inclusive/exclusive, should this be clarified?

numba/cuda/cudadrv/nvvm.py
```diff
@@ -343,51 +343,69 @@ def get_log(self):
...
+        # For unsupported CUDA toolkit versions, all we can do is assume all
+        # non-deprecated versions we are aware of are supported.
+        return tuple([cc for cc in COMPUTE_CAPABILITIES if cc >= (5, 3)])
+
+
 _supported_cc = None
 
 
 def get_supported_ccs():
+    # Attempt to return cached list
```
stuartarchibald, line 385: If there's an in-process cache need, should something like `functools.cache` be used opposed to module globals?

numba/cuda/cudadrv/nvvm.py
```diff
@@ -343,51 +343,69 @@ def get_log(self):
...
-                         (8, 0), (8, 6), (8, 7))
-    else:
+        return _supported_cc
+
+    # Ensure the minimum CTK version requirement is met
+    if cudart_version < (11, 0):
         _supported_cc = ()
+        ctk_ver = f"{cudart_version[0]}.{cudart_version[1]}"
+        unsupported_ver = (f"CUDA Toolkit {ctk_ver} is unsupported by Numba - "
+                           "11.0 is the minimum required version.")
```
stuartarchibald, line 404: Should the minimum version be specified as a constant somewhere (maybe a module global, or perhaps even all this encoding of supported CCs etc ought to go into it's own module or YAML file?) and then code be updated in terms of that constant?

numba/cuda/tests/cudadrv/test_nvvm_driver.py
```diff
@@ -15,7 +14,8 @@
 class TestNvvmDriver(unittest.TestCase):
     def get_nvvmir(self):
         if NVVM().is_nvvm70:
-            metadata = metadata_nvvm70
+            versions = NVVM().get_ir_version()
+            metadata = metadata_nvvm70 % versions
```
stuartarchibald, line 18: Perhaps update to use `.format()` or an `f-string`?

numba/cuda/tests/cudadrv/test_nvvm_driver.py
```diff
@@ -25,21 +25,22 @@ def get_nvvmir(self):
...
         nvvmir = self.get_nvvmir()
-        ptx = llvm_to_ptx(nvvmir).decode('utf8')
+        ptx = nvvm.llvm_to_ptx(nvvmir).decode('utf8')
         self.assertTrue('simple' in ptx)
         self.assertTrue('ave' in ptx)
 
     def test_nvvm_from_llvm(self):
         m = ir.Module("test_nvvm_from_llvm")
         m.triple = 'nvptx64-nvidia-cuda'
+        nvvm.add_ir_version(m)
```
stuartarchibald, line 35: Is this 3 line "setup" for a CUDA-target-LLVM-module sufficiently common to make it worth creating a utility function? The main concern here is that the `m.triple` specification is likely always necessary whereas the `nvvm.add_ir_version(m)` appears not to be? I suppose the underlying question is "what is the need driving the addition of the `add_ir_version()`" call?

stuartarchibald (commented):
Thanks for the patch. This largely looks good, just a few queries to resolve inline. Thanks again!

numba/cuda/cudadrv/nvvm.py
gmarkall: That would be way better than the current bodge!

numba/cuda/tests/cudadrv/test_nvvm_driver.py
gmarkall: I decided against that because it makes the IR less legible due to the presence of `{` and `}`.

numba/cuda/cudadrv/nvvm.py
gmarkall: Actually it doesn't even need to be cached, this just makes no sense. The supported CCs is a property of the NVVM version, so I'm going to just make `supported_ccs` a property of the `NVVM` instance, because that just makes more sense anyway (and implicitly "caches" the values too because it's a singleton).

I think there could be an argument for the logic in `get_supported_ccs()` and its callees to be part of the `NVVM` class, but I think that functionality will also be common with NVPTX so I don't want to hoist the code into `NVVM` for now.

numba/cuda/cudadrv/nvvm.py
gmarkall: Now clarified.

stuartarchibald (approved):
@gmarkall Thanks for the fixes! I agree with the comments you left in response to review, thanks for answering/addressing them.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_172`.

gmarkall:
Any news on the buildfarm?

esc:
@gmarkall @stuartarchibald apologies again for the delay here. Some hardware on the buildfram had to be replaced, hence the delay here. I have scheduled a fresh run for this PR at:

`Build numba_smoketest_cuda_yaml_176 has started`

PR #8580, opened 2022-11-04
adding note about doing a wheel test build prior to tagging

esc:
As title


stuartarchibald (approved):
Thanks for the patch.

PR #8588, opened 2022-11-09
Remove lowering extension hooks, replace with pass infrastructure.

stuartarchibald:
This removes the "extension hook" for handling custom IR nodes in the native target lowering. The same behaviour can be achieved through use of the compiler pass infrastructure which is the declared extension interface for building compiler customisations.

The changes to the code base comprise:

* Creating a new pass to handle lowering of "parfor" nodes, this is based on the standard lowering pass. Both the new "parfor lowering" and the current "lowering" passes are now implemented as an extension of a "base" lowering pass. The "parfor lowering" pass is used as the lowerer if `parallel=True`.
* Removing the lowering "extension hook" as it's not needed.
* Removing the registration of the above "extension hook" for Parfor nodes in the CPU context as it is not needed.

### DIFF
--- a/numba/core/typed_passes.py
+++ b/numba/core/typed_passes.py
@@ -9,6 +9,7 @@ from numba.core import (errors, types, typing, ir, funcdesc, rewrites,
 from numba.parfors.parfor import PreParforPass as _parfor_PreParforPass
 from numba.parfors.parfor import ParforPass as _parfor_ParforPass
 from numba.parfors.parfor import Parfor
+from numba.parfors.parfor_lowering import ParforLower
 
 from numba.core.compiler_machinery import (FunctionPass, LoweringPass,
                                            AnalysisPass, register_pass)
@@ -352,10 +353,10 @@ class DumpParforDiagnostics(AnalysisPass):
         return True
 
 
-@register_pass(mutates_CFG=True, analysis_only=False)
-class NativeLowering(LoweringPass):
+class BaseNativeLowering(LoweringPass):
 
-    _name = "native_lowering"
+    _name = None
+    _lowering_class = None
 
     def __init__(self):
         LoweringPass.__init__(self)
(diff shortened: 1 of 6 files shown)


numba/core/typed_passes.py
```diff
@@ -434,6 +435,20 @@ def run_pass(self, state):
         return True
 
 
+@register_pass(mutates_CFG=True, analysis_only=False)
+class NativeLowering(BaseNativeLowering):
+
```
sklam, line 440: empty line. maybe put a simple docstring here.

numba/core/typed_passes.py
```diff
@@ -434,6 +435,20 @@ def run_pass(self, state):
...
+    _name = "native_lowering"
+    _lowering_class = lowering.Lower
+
+
+@register_pass(mutates_CFG=True, analysis_only=False)
+class NativeParforLowering(BaseNativeLowering):
+
```
sklam, line 447: empty line. maybe put a simple docstring here.

numba/core/typed_passes.py
```diff
@@ -352,10 +353,10 @@ def run_pass(self, state):
         return True
 
 
-@register_pass(mutates_CFG=True, analysis_only=False)
-class NativeLowering(LoweringPass):
+class BaseNativeLowering(LoweringPass):
 
-    _name = "native_lowering"
+    _name = None
+    _lowering_class = None
```
sklam, line 359: Maybe making this abstractclassmethod so we lock down how it can be modified a little bit.

sklam (commented):
The changes are unexpectedly simply. I only have a few minor comments:

stuartarchibald:
> The changes are unexpectedly simply. I only have a few minor comments:

Thanks for the review, comments should be resolved in 444108f.

Issue #7775, opened 2022-01-24
Fix DI generation for array exprs


stuartarchibald:
Numba's DI generation for array expressions leads to strange behaviour in gdb. Issues include:
* Stepping into an array expression lands here: https://github.com/numba/numba/blob/14b170a11c9b3e546b24943ffc446e152093080d/numba/np/npyimpl.py#L190-L205
* If https://github.com/numba/numba/blob/14b170a11c9b3e546b24943ffc446e152093080d/numba/np/npyimpl.py#L190 is stepped through until finish and then stepped over the next executable line is reported as line 1 of the file containing the array expr.
* Once on line 1, continued stepping repeatedly reports the same location with varying input values, though it will eventually step back into the calling function.

As noted here: https://github.com/numba/numba/pull/7752#issuecomment-1018394185 the fix may be to just omit DI for array expressions.

stuartarchibald:
xref: https://github.com/numba/numba/issues/7787 which demonstrates the problem.

PR #8590, opened 2022-11-09
Py3.11 support continues

sklam:
based on https://github.com/numba/numba/pull/8545


sklam:
rebased to pickup CI fixes

numba/core/byteflow.py
```diff
@@ -445,6 +445,9 @@ def op_LOAD_GLOBAL(self, state, inst):
     def op_COPY_FREE_VARS(self, state, inst):
         state.append(inst)
 
+    def op_MAKE_CELL(self, state: "State", inst):
```
stuartarchibald, line 448:
```suggestion
    def op_MAKE_CELL(self, state, inst):
```
not sure this adds anything?

numba/core/interpreter.py
```diff
@@ -2239,14 +2239,30 @@ def op_LOAD_DEREF(self, inst, res):
...
+            self.store(value=value, name=dstname)
+    elif PYVERSION < (3, 11):
+        def op_STORE_DEREF(self, inst, value):
+            n_cellvars = len(self.code_cellvars)
+            if inst.arg < n_cellvars:
+                dstname = self.code_cellvars[inst.arg]
+            else:
+                dstname = self.code_freevars[inst.arg - n_cellvars]
+            value = self.get(value)
+            self.store(value=value, name=dstname)
```
stuartarchibald, line 2263: I don't think this is quite right, but it is fixed in 485a1e89 on #8621.

numba/core/interpreter.py
```diff
@@ -2994,21 +3010,43 @@ def op_MAKE_CLOSURE(self, inst, name, code, closure, annotations,
...
+                    msg = "Unsupported use of op_LOAD_CLOSURE encountered"
+                    raise NotImplementedError(msg)
+            else:
+                idx = inst.arg - n_cellvars
+                name = self.code_freevars[idx]
+                value = self.get_closure_value(idx)
+                gl = ir.FreeVar(idx, name, value, loc=self.loc)
+            self.store(gl, res)
+    else:
+        raise NotImplementedError(PYVERSION)
```
stuartarchibald, line 3049: I don't think this is quite right, but it is fixed in 485a1e89 on #8621.

numba/core/ir.py
```diff
@@ -138,7 +138,7 @@ def strformat(self, nlines_up=2):
 
 
         ret = [] # accumulates output
-        if lines and use_line:
+        if lines and use_line > 0:
```
stuartarchibald, line 141: I think e0c54bde from #8621 "fixes" the root cause of the problem (because it also hits DWARF output, can't have negative line numbers) but putting in this protection in the user facing part is probably a good idea anyway.

numba/misc/codetype.py
```diff
@@ -0,0 +1,42 @@
+import types
+
+from numba.core import utils
+
+
+def copy_code_type(code: types.CodeType, **kwargs) -> types.CodeType:
```
stuartarchibald, line 6: Is this essentially provided by stdlib `types.CodeType.replace()`? It's new in Python 3.8 but as that'll be the baseline version by the time 3.11 support is added should be ok to use?

numba/np/ufunc/array_exprs.py
```diff
@@ -329,6 +329,28 @@ def _legalize_parameter_names(var_list):
             var.name = old_name
 
 
+class _EraseInvalidLineRanges(ast.NodeTransformer):
```
stuartarchibald, line 332: What is this fixing? Is there an example? I ask this as I'm wondering if needing this is simply an effect of "strange" line number information being generated earlier on in the compilation pipeline?

numba/tests/test_parfors.py
```diff
@@ -3346,7 +3350,7 @@ def generate_prange_func(self, pyfunc, patch_instance):
             # look for LOAD_GLOBALs that point to 'range'
             for instr in dis.Bytecode(pyfunc_code):
                 if instr.opname == 'LOAD_GLOBAL':
-                    if instr.arg == range_idx:
+                    if _fix_LOAD_GLOBAL_arg(instr.arg) == range_idx:
```
stuartarchibald, line 3353: Not for now, but this also needs 6b101e2 from #8621 to address the correct index following the injection of the `prange` into `co_names`.

stuartarchibald (commented):
Thanks for the patch @sklam. I've reviewed this with view of it making progress in terms of fixing issues related to Python 3.11. I've noted proposed changes which partially fix issues where I think a subsequent fix is needed that is present in #8621, it's probably ok to merge these as-is so as to keep the diff small. The main queries in the review are:
1. Are some of these changes due to the line numbering being incorrect, in which case fixing that (as in #8621) might make the problems disappear.
2. Use of stdlib `types.CodeType.replace()` instead of creating something Numba specific.

Thanks again for working on this!

numba/core/byteflow.py
sklam: oh just to get vscode to provide proper hints at one point. let's remove it.

numba/core/interpreter.py
sklam: stay as is in this pr

numba/misc/codetype.py
sklam: ok. i'll switch to it

numba/np/ufunc/array_exprs.py
sklam: Anything with arrayexprs and a lot of stencil tests fail without this. This has nothing to do with line number information being wrong. It's completely caused by inlining and the way arrayexpr generate ASTs and splicing ASTs. In Py3.11, AST nodes contain start and end linenos. Arrayexpr splices AST in a way that can generate nodes with outer AST node saying the line range is (1, 10) and the inner nodes are at line 100. This fixes it.

numba/tests/test_parfors.py
sklam: yup, i saw that when i reviewed #8621. keeping this as is in this pr

numba/np/ufunc/array_exprs.py
stuartarchibald: Ah I see. Thanks for the explanation. IIRC the `arrayexpr` node line numbers aren't quite right anyway, e.g. #7775.

sklam:
I've addressed everything

stuartarchibald (approved):
Thanks for the fixes. This is approved with view of it being a further improvement on the way to Python 3.11 support.

PR #8594, opened 2022-11-10
Fix various CUDA lineinfo issues

### LINKED ISSUE #8581: CUDA: Lineinfo doesn't work for device functions
For example, running:

```python
import numba.cuda
import numpy as np
import PIL.Image

@numba.cuda.jit(device=True, inline=False, lineinfo=True)
def get_pixel(img, x, y, size):
    if x >= 0 and y >= 0 and x < size and y < size:
        return img[y][x]
    else:
        return 0.
    
# 9x9 gaussian blur kernel
@numba.cuda.jit(lineinfo=True)
def gaussian_blur(img_in, img_out, size):
    tx, ty = numba.cuda.grid(2)

    value = 0.0
    radius = 16
    num = (radius + 1) * (radius + 1)

    for dy in range(-radius, radius + 1):
        for dx in range(-radius, radius + 1):
            x = tx + dx
            y = ty + dy
            value += get_pixel(img_in, x, y, size) * (1.0 / num)

    img_out[ty][tx] = value


size = 256
tpb = (16, 16)
threads = (size, size)
blocks = (threads[0]//tpb[0], threads[1]//tpb[1])

img = np.random.normal(0, 1.0, size*size).astype(np.float32).reshape(threads)

gpu_in = numba.cuda.to_device(img)
gpu_out = numba.cuda.device_array_like(img)

gaussian_blur[blocks, tpb](gpu_in, gpu_out, size)
numba.cuda.synchronize()
img = gpu_out.copy_to_host()
```

gives a view in the source window of NSight compute that looks like the attached image - line info for `get_pixel()` is missing.

missing_line_info_issue

### LINKED ISSUE #8582: CUDA: Enabling lineinfo causes many redundant moves in SASS
These redundant moves can be seen in the attached image, from profiling

```python
import numba.cuda
import numpy as np
import PIL.Image

@numba.cuda.jit(device=True, inline=False, lineinfo=True)
def get_pixel(img, x, y, size):
    if x >= 0 and y >= 0 and x < size and y < size:
        return img[y][x]
    else:
        return 0.
    
# 9x9 gaussian blur kernel
@numba.cuda.jit(lineinfo=True)
def gaussian_blur(img_in, img_out, size):
    tx, ty = numba.cuda.grid(2)

    value = 0.0
    radius = 16
    num = (radius + 1) * (radius + 1)

    for dy in range(-radius, radius + 1):
        for dx in range(-radius, radius + 1):
            x = tx + dx
            y = ty + dy
            value += get_pixel(img_in, x, y, size) * (1.0 / num)

    img_out[ty][tx] = value


size = 256
tpb = (16, 16)
threads = (size, size)
blocks = (threads[0]//tpb[0], threads[1]//tpb[1])

img = np.random.normal(0, 1.0, size*size).astype(np.float32).reshape(threads)

gpu_in = numba.cuda.to_device(img)
gpu_out = numba.cuda.device_array_like(img)

gaussian_blur[blocks, tpb](gpu_in, gpu_out, size)
numba.cuda.synchronize()
img = gpu_out.copy_to_host()
```

these are of the form:

```
MOV R4, R4
MOV R5, R5
```

etc.

missing_line_info_issue

This is consistent with debug being enabled when compiling to SASS. This probably happens because the PTX generated by NVVM begins with:

```
//
// Generated by NVIDIA NVVM Compiler
//
// Compiler Build ID: CL-30794723
// Cuda compilation tools, release 11.6, V11.6.55
// Based on NVVM 7.0.1
//

.version 7.6
.target sm_75, debug
.address_size 64
```

(`.target sm_75, debug`) being the relevant part. The next step is probably to find out what's triggering the introduction of this target specification.

gmarkall:
The main aim here was to fix #8581 and #8582, but I also discovered a couple more issues along the way that needed fixing so that I could properly write tests for these fixes. I also spotted that adding lineinfo changed the error model from NumPy to Python, which is inappropriate - requesting lineinfo should not change the generated code.

I also removed the `NvvmDIBuilder` because it isn't really needed / used anymore and I didn't want to unnecessarily update it to align with these fixes - it was only needed for NVVM 3.4, for which we don't support debugging. There were some tests that still checked it, so I modified tests appropriately for the removal.

Individual commit messages go into the details of each fix / change.

As an example of the improvements resulting from this PR, the following is the source view in NSight Compute for an example before this PR is applied:

source-before

Note how there is no lineinfo for the device function (nothing listed in the "Instructions Executed" column for any of the device function's lines) and how there are many redundant moves in the SASS (e.g. `MOV R90, R90`, etc.)

After applying this PR:

source-after

We see lineinfo ("Instructions executed") for the device function lines, and no redundant moves in the SASS.

Many thanks to @fs-nv for identifying these issues, guidance on running NSight Compute in the correct way, and providing an explanation of how/why the redundant moves were getting generated.

Fixes #8581.
Fixes #8582.


gmarkall:
gpuci run tests

fs-nv:
I looked over the changes and everything made sense to me (to the extend that I understood the numba code). Thanks for fixing this, it will be very useful for future performance analysis of numba.cuda programs.

gmarkall:
gpuci run tests

numba/cuda/tests/cudapy/test_lineinfo.py
```diff
@@ -1,42 +1,207 @@
...
+        ptx = fn.inspect_asm(sig)
+        assertfn = self.assertIsNotNone if expect else self.assertIsNone
+
+        # DICompileUnit debug info metadata should all be of the
+        # DebugDirectivesOnly kind, and not the FullDebug kind
+        pat = (
+            r'!DICompileUnit\(.*'    # Opening of DICompileUnit metadata. Since
+                                     # the order of attributes is not
+                                     # guaranteed, we need to match arbitrarily
+                                     # aftwerwards.
```
stuartarchibald, line 34:
```suggestion
                                     # afterwards.
```

numba/cuda/tests/cudapy/test_lineinfo.py
```diff
@@ -1,42 +1,207 @@
...
+
+        if start is None:
+            self.fail(f'Could not identify device function in:\n\n{ptx}')
+
+        # Identify the end of the function
+        end = None
+
+        for offset, line in enumerate(ptxlines[start:]):
+            # Assume the end of the function is a line with an unindented '}'
+            if line[:1] == '}':
```
stuartarchibald, line 165: This matches `}<anything can go here>`, should the length of the line be checked or the `.strip()`ed version matched instead?

numba/cuda/tests/cudapy/test_lineinfo.py
```diff
@@ -1,42 +1,207 @@
...
+                end = start + offset
```
stuartarchibald, line 166: Should there be a `break` here once the match is found?

numba/cuda/tests/cudapy/test_lineinfo.py
```diff
@@ -1,42 +1,207 @@
...
+        devfn_start = re.compile(r'^\.weak\s*\.func')
+
+        # Identify the beginning of the function.
+        start = None
+
+        for lineno, line in enumerate(ptxlines):
+            if devfn_start.match(line) is not None:
+                # We will begin our search on the line following the
+                # declaration
+                start = lineno + 1
```
stuartarchibald, line 155: Should there be a `break` here once the match is found?

numba/cuda/tests/cudapy/test_lineinfo.py
```diff
@@ -1,42 +1,207 @@
...
+        for line in ptxlines[start:end]:
+            if loc_directive.search(line) is not None:
+                found = True
+
+        if not found:
+            # Join one line either side so the function as a whole is shown,
+            # i.e. including the declaration and parameter list, and the
+            # closing brace.
+            devfn = "\n".join(ptxlines[start - 1:end + 1])
+            self.fail(f'.loc directive not found in:\n\n{devfn}')
```
stuartarchibald, line 184:
```suggestion
        for line in ptxlines[start:end]:
            if loc_directive.search(line) is not None:
                break
        else:
            # Join one line either side so the function as a whole is shown,
            # i.e. including the declaration and parameter list, and the
            # closing brace.
            devfn = "\n".join(ptxlines[start - 1:end + 1])
            self.fail(f'.loc directive not found in:\n\n{devfn}')
```
suggest using `for...else...`

numba/cuda/tests/cudapy/test_lineinfo.py
```diff
@@ -1,42 +1,207 @@
...
+        # checks on the generated LLVM and PTX.
+
+        @cuda.jit(lineinfo=True)
+        def callee(x):
+            x[0] += 1
+
+        @cuda.jit(lineinfo=True)
+        def caller(x):
+            x[0] = 1
+            callee(x)
```
stuartarchibald, line 133: As noted OOB, I'm not sure it's not clear what the inheritance model is for the `lineinfo` flag. Does the device function need `lineinfo=True` if the calling kernel has `lineinfo=True` already. Further, what if the device function has `False` but the kernel has `True` (or vice versa).

stuartarchibald (commented):
Thanks for the patch @gmarkall, this in general looks good. I've tried to capture some of the content of queries made OOB in the review, there's a few (mostly) minor things to address, once resolved should be good to merge. The remaining review task I have is to go and manually test this patch to make sure it does as expected! Thanks again!

numba/cuda/compiler.py
```diff
@@ -190,15 +190,23 @@ def compile_cuda(pyfunc, return_type, args, debug=False, lineinfo=False,
     flags.no_compile = True
     flags.no_cpython_wrapper = True
     flags.no_cfunc_wrapper = True
+
+    # Both debug and lineinfo turn on debug information in the compiled code,
```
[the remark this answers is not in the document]
gmarkall: Yes, I think so.

numba/cuda/cudadrv/nvvm.py
```diff
@@ -273,12 +272,6 @@ def compile(self, **options):
 
         # stringify options
         opts = []
-        if 'debug' in options:
-            if options.pop('debug'):
-                opts.append('-g')
-
-        if options.pop('lineinfo', False):
-            opts.append('-generate-line-info')
```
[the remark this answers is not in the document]
gmarkall: Probably not - I think this isn't a public API, so I don't think the use of these options need to be guarded for.

numba/cuda/tests/cudapy/test_lineinfo.py
gmarkall: I'm not a big fan of `for .. else ..` because it makes the condition ("found") implicit rather than explicit, and therefore harder for a reader to understand. Do you feel strongly the other way?

numba/cuda/tests/cudapy/test_lineinfo.py
gmarkall: It is not necessary for the device function to have it (and `False` is the default). I think perhaps if this test had two variants, that check with `lineinfo=True` on the device function and without it, both expecting the exact same behaviour, that would cover the cases - does that sound suitable to you?

numba/cuda/tests/cudapy/test_lineinfo.py
gmarkall: I guess there could be. I was working on the assumption that there would only be one matching line (as it would be surprising to have multiple device functions) but adding a `break` probably communicates more clearly that we need look no further.

numba/cuda/tests/cudapy/test_lineinfo.py
gmarkall: It would be legitimate for something to follow it, and I've sometimes seen trailing whitespace in different versions' output - I think the check is still as strong but more robust if we just check for the closing brace regardless of what follows.

numba/cuda/tests/cudapy/test_lineinfo.py
gmarkall: I think we might as well do that here for the same reason as above.

numba/cuda/tests/cudapy/test_lineinfo.py
stuartarchibald: I don't feel strongly either way, but would prefer explicit `break` on the condition being found. I made the suggestion after noticing there was no `break` on being found, then realised it was a `for...else..`,

numba/cuda/tests/cudapy/test_lineinfo.py
stuartarchibald: ah ok, thanks for explaining, perhaps update the comment above to say `# Assume the end of the function is a line starting with an unindented '}'`? I read the original comment to mean it was looking for a line with just a `}` on it.

numba/cuda/cudadrv/nvvm.py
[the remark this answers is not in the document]
stuartarchibald: Thanks for confirming, this seems reasonable.

numba/cuda/compiler.py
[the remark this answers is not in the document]
gmarkall: I decided to go with a warning here rather than raising an error - thinking about it, some people might have both in existing code (especially in test suites). I'll think about upgrading the warning to an exception in a following release rather than breaking things immediately.

numba/cuda/target.py
```diff
@@ -200,10 +205,12 @@ def generate_kernel_wrapper(self, library, fndesc, kernel_name, debug,
         wrapfn = ir.Function(wrapper_module, wrapfnty, prefixed)
         builder = ir.IRBuilder(wrapfn.append_basic_block(''))
 
-        if debug:
-            debuginfo = self.DIBuilder(
-                module=wrapper_module, filepath=filename, cgctx=self,
-            )
+        if debug or lineinfo:
+            directives_only = lineinfo and not debug
```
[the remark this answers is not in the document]
gmarkall: As I decided to make it a warning rather than an error, this logic is kept intact for now.

numba/cuda/tests/cudapy/test_lineinfo.py
gmarkall: Turns out I'm wrong, and you do need to add `lineinfo=True` to every function you want lineinfo for - it doesn't propagate.

numba/cuda/tests/cudapy/test_lineinfo.py
```diff
@@ -1,42 +1,207 @@
...
-        fn.compile(sig)
-        return fn.inspect_asm(sig)
+    def _loc_directive_regex(self):
+        # This is used in several tests
+
+        pat = (
+            r'\.loc'      # .loc directive beginning
+            r'\s*[0-9]*'  # whitespace then file index
+            r'\s*[0-9]*'  # whitespace then line number
+            r'\s*[0-9]*'  # whitespace then column position
```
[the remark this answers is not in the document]
gmarkall: There should be, yeah - in addition to the suggestions on this PR, I found a couple more places where whitespace could be aggressively enforced too.

gmarkall:
gpuci run tests

gmarkall:
@stuartarchibald Many thanks for the review - I've replied to some of the comments and attempted to address some with code. I think we might need another round on some of the discussions on the diff, if you're not comfortable with every response.

gmarkall:
gpuci run tests

numba/cuda/tests/cudapy/test_lineinfo.py
gmarkall: I've now removed this code anyway (it's not needed when we're generating code "correctly" without performance regression with lineinfo).

numba/cuda/tests/cudapy/test_lineinfo.py
gmarkall: This code is now removed as it was no longer relevant to the test after the performance regression fix.

numba/cuda/tests/cudapy/test_lineinfo.py
gmarkall: This code is now removed as it is no longer needed anyway. (As per https://github.com/numba/numba/pull/8594#discussion_r1108670106)

numba/cuda/tests/cudapy/test_lineinfo.py
gmarkall: From OOB discussion: Changing this behaviour requires quite a large rework of the way flags and options work, and although I tried to fix that for this PR, it is quite a big scope bloat, so I'll have to follow up with another PR if we decide to change this soon. Some of the work towards that can be found in this branch (only provided for info, there's no need related to this PR to look at it): https://github.com/gmarkall/numba/tree/cuda-lineinfo-fixes-wip-20230216

gmarkall:
@stuartarchibald Many thanks for the review - the performance regression identified by @fs-nv due to prevention of inlining when lineinfo is enabled has now been fixed in this branch, and all comments above responded-to. Most of them pertained to code that I have now deleted because the test needed to change once inlining was working again, so there should be less to re-review than there would have been.

numba/cuda/tests/cudapy/test_lineinfo.py
stuartarchibald: Thanks for capturing the discussion/outcomes over the handling of flags @gmarkall.

numba/cuda/compiler.py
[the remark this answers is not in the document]
stuartarchibald: > I decided to go with a warning here rather than raising an error - thinking about it, some people might have both in existing code (especially in test suites). I'll think about upgrading the warning to an exception in a following release rather than breaking things immediately.

Perhaps open a ticket to track?

stuartarchibald:
I've tested 0dafa70 manually.
* Issue #8582 looks fixed, the reproducer in the OP has a lot of redundant  `MOV` instructions when running on `main` branch, on this branch that it not the case, the SASS looks a lot more optimised.
* Issue #8581 also appears to be fixed, the reproducer in the OP has nothing DI related generated for the device function on `main` branch, on this branch something relatively minimal is generated (it could potentially be smaller but the requirement is to make sure that the function is fully optimised and DI lineinfo is present which now appears to be the case).

numba/cuda/tests/cudapy/test_lineinfo.py
```diff
@@ -1,42 +1,205 @@
...
+        # Check that there is no device function in the PTX
+
+        # A line beginning with ".weak .func" that identifies a device function
+        devfn_start = re.compile(r'^\.weak\s*\.func')
+
+        for line in ptxlines:
+            if devfn_start.match(line) is not None:
+                self.fail(f"Found device function in PTX:\n\n{ptx}")
+
+        # Scan for .loc directives taht refer to an inlined device function
```
stuartarchibald, line 153:
```suggestion
        # Scan for .loc directives that refer to an inlined device function
```

numba/cuda/tests/cudapy/test_lineinfo.py
```diff
@@ -1,42 +1,205 @@
...
+
+        # Now we can check the PTX of the device function specifically.
+
+        ptx = caller.inspect_asm(sig)
+        ptxlines = ptx.splitlines()
+
+        # Check that there is no device function in the PTX
+
+        # A line beginning with ".weak .func" that identifies a device function
+        devfn_start = re.compile(r'^\.weak\s*\.func')
```
stuartarchibald, line 147:
```suggestion
        devfn_start = re.compile(r'^\.weak\s+\.func')
```

numba/cuda/tests/cudapy/test_lineinfo.py
```diff
@@ -1,42 +1,205 @@
...
+        for line in ptxlines:
+            if loc_directive.search(line) is not None:
+                if 'inlined_at' in line:
+                    found = True
+                    break
+
+        if not found:
+            # Join one line either side so the function as a whole is shown,
+            # i.e. including the declaration and parameter list, and the
+            # closing brace.
```
stuartarchibald, line 167: Is this comment applicable? It looks like it's just showing the full ptx?

numba/cuda/tests/cudapy/test_lineinfo.py
```diff
@@ -1,42 +1,205 @@
...
+        # - The callee
+        expected_subprograms = 3
+
+        self.assertEqual(subprograms, expected_subprograms,
+                         f'"Expected {expected_subprograms} DISubprograms; '
+                         f'got {subprograms}')
+
+    def test_debug_and_lineinfo_warning(self):
+        with warnings.catch_warnings(record=True) as w:
+            # We pass opt=False to prevent the warning about opt and debug
```
stuartarchibald, line 193: This block should probably have a call to `numba.tests.support.ignore_internal_warnings()` to suppress known warnings.

stuartarchibald (commented):
Thanks for the updates @gmarkall, few minor things to address else looks good.

numba/cuda/tests/cudapy/test_lineinfo.py
gmarkall: Ah, not anymore - it was from before the function was (correctly) inlined.

numba/cuda/tests/cudapy/test_lineinfo.py
gmarkall: OK - will add.

numba/cuda/tests/cudapy/test_lineinfo.py
gmarkall: Changed.

numba/cuda/tests/cudapy/test_lineinfo.py
gmarkall: Fixed.

numba/cuda/tests/cudapy/test_lineinfo.py
gmarkall: Now removed.

numba/cuda/tests/cudapy/test_lineinfo.py
gmarkall: I've added a call to `ignore_internal_warnings()`.

gmarkall:
gpuci run tests

gmarkall:
@stuartarchibald Many thanks - all minor points addressed.

stuartarchibald (approved):

numba/cuda/tests/cudapy/test_lineinfo.py
stuartarchibald: thanks.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_186`.

Passed.

PR #8598, opened 2022-11-11
Setup Azure CI to test py3.11

sklam:
(closes #8595)

- only test linux
- see #8595 for example of the CI working.  https://github.com/numba/numba/pull/8598/commits/d8f5c8c7048cdee7aeb7ba22fe75b973b8b47291 is cherrypick of https://github.com/numba/numba/pull/8595/commits/3c3142740084abaa63b25f92b361c69a18aec709 in #8595

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -6,117 +6,141 @@ variables:
   TEST_COUNT: 19
 
 jobs:
-# Mac and Linux use the same template with different matrixes
-- template: buildscripts/azure/azure-linux-macos.yml
-  parameters:
-    name: macOS
-    vmImage: macos-11
-    matrix:
-      py37_np118:
-        PYTHON: '3.7'
-        NUMPY: '1.18'
-        CONDA_ENV: 'azure_ci'
-        TEST_START_INDEX: 0
-      py310_np123:
-        PYTHON: '3.10'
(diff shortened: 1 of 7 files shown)


stuartarchibald (approved):
Thanks for the patch. There's some superfluous changes with respect to windows builds (given this targets linux only), but as discussed OOB this patch is going to be reverted on the `py3.11` development branch prior to merge to `main` as it is just to get some testing/feedback running.

Seems like the only difference between this patch and 3c3142740084abaa63b25f92b361c69a18aec709 from #8595 is this:
```patch
<  if [[ "$(uname -s)" == *"Linux"* ]] && [[ "$(uname -p)" == *"86"* ]]; then
<      EXTRA_BUILD_EXT_FLAGS="--werror --wall"
---
>  # if [[ "$(uname -s)" == *"Linux"* ]] && [[ "$(uname -p)" == *"86"* ]]; then
>  #     EXTRA_BUILD_EXT_FLAGS="--werror --wall"
```
whereby this patch does not comment out the build-with-warnings-as-errors flags, which is probably a good thing!

stuartarchibald:
Note: The build is expected to fail, this PR is changing public CI to start working with Python 3.11 exclusively on the merge target branch of `py3.11`.

Issue #8586, opened 2022-11-09
np.kron behaves differently when compiled with numba


SnoodMcGuinness:
np.kron of an int array with a float array will return an int array when compiled with numba, but a float when run in normal numpy:

def kron_result(size=3):
    a = np.sqrt(np.diag(np.arange(size), k=1))
    one = np.array([[1]])
    kron = np.kron(one, a)
    return kron

returns 
array([[0.        , 0.        , 0.        , 0.        ],
       [0.        , 0.        , 1.        , 0.        ],
       [0.        , 0.        , 0.        , 1.41421356],
       [0.        , 0.        , 0.        , 0.        ]])
in normal numpy, but 
array([[0, 0, 0, 0],
       [0, 0, 1, 0],
       [0, 0, 0, 1],
       [0, 0, 0, 0]])

when compiled with njit

I am working on a mac M1 processor

Conda list:
`anyio                     3.6.2              pyhd8ed1ab_0    conda-forge
appnope                   0.1.3              pyhd8ed1ab_0    conda-forge
argon2-cffi               21.3.0             pyhd8ed1ab_0    conda-forge
argon2-cffi-bindings      21.2.0          py310h90acd4f_3    conda-forge
asttokens                 2.1.0              pyhd8ed1ab_0    conda-forge
attrs                     22.1.0             pyh71513ae_1    conda-forge
backcall                  0.2.0              pyh9f0ad1d_0    conda-forge
backports                 1.0                        py_2    conda-forge
backports.functools_lru_cache 1.6.4              pyhd8ed1ab_0    conda-forge
beautifulsoup4            4.11.1             pyha770c72_0    conda-forge
bleach                    5.0.1              pyhd8ed1ab_0    conda-forge
brotli                    1.0.9                hb7f2c08_8    conda-forge
brotli-bin                1.0.9                hb7f2c08_8    conda-forge
bzip2                     1.0.8                h0d85af4_4    conda-forge
ca-certificates           2022.9.24            h033912b_0    conda-forge
certifi                   2022.9.24          pyhd8ed1ab_0    conda-forge
cffi                      1.15.1          py310ha78151a_2    conda-forge
contourpy                 1.0.6           py310ha23aa8a_0    conda-forge
cycler                    0.11.0             pyhd8ed1ab_0    conda-forge
debugpy                   1.6.3           py310h7a76584_1    conda-forge
decorator                 5.1.1              pyhd8ed1ab_0    conda-forge
[...]

PR #8601, opened 2022-11-15
Check for type mismatch in `np.kron`

apmasell:
Convert a comment into an assertion that two array types must match.

### DIFF
--- a/numba/np/linalg.py
+++ b/numba/np/linalg.py
@@ -2792,9 +2792,10 @@ def kron_impl(a, b):
     fix_b = _kron_normaliser_impl(b)
     ret_c = _kron_return(a, b)
 
-    # this is fine because the ufunc for the Hadamard product
-    # will reject differing dtypes in a and b.
     dt = getattr(a, 'dtype', a)
+    if (isinstance(a, types.Array) and isinstance(b, types.Array) and
+       a.dtype != b.dtype):
+        raise NumbaTypeError("Both types must be the same in numpy.kron.")
 
     def kron_impl(a, b):
 
--- a/numba/tests/test_linalg.py
+++ b/numba/tests/test_linalg.py
@@ -2643,6 +2643,9 @@ class TestBasics(TestLinalgSystems):  # TestLinalgSystems for 1d test
 
         self._assert_wrong_dim("kron", cfunc)
 
+        with self.assertRaises(errors.TypingError):
+            cfunc(np.ones(10, dtype=np.intp), np.ones(10, dtype=np.float32))
+
         args = (np.empty(10)[::2], np.empty(10)[::2])
         msg = "only supports 'C' or 'F' layout"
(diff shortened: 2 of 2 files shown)


numba/np/linalg.py
```diff
@@ -2792,9 +2792,10 @@ def kron_impl(a, b):
     fix_b = _kron_normaliser_impl(b)
     ret_c = _kron_return(a, b)
 
-    # this is fine because the ufunc for the Hadamard product
-    # will reject differing dtypes in a and b.
     dt = getattr(a, 'dtype', a)
+    if (isinstance(a, types.Array) and isinstance(b, types.Array) and
+       a.dtype != b.dtype):
+        raise NumbaTypeError("Both types must be the same in numpy.kron.")
```
stuartarchibald, line 2798: Do the `dtype`s need to be _the same_ or just of the same family, i.e. does this implementation work ok for two different floating point types?

apmasell: I have no idea. I'm matching the comment as best I understand it.

stuartarchibald: The issue is here, the `dtype` is taken from `a`, the first arg:
https://github.com/numba/numba/blob/fb952f8f5959ce31583da612a91b204bca58da81/numba/np/linalg.py#L2797
which is then used as the `dtype` in the allocation of the return value (array `c`):
https://github.com/numba/numba/blob/fb952f8f5959ce31583da612a91b204bca58da81/numba/np/linalg.py#L2812-L2813
It is safest to require them to be the same, anything else would probably require analysis similar to that needed to just fix the issue!

stuartarchibald, line 2798: I'm not sure that this check is currently sufficient. Consider:
```python
from numba import njit
import numpy as np

@njit
def foo(x, y):
    return np.kron(x, y)

args = (1, np.array([1.2, 3.4]))
print(foo(*args))
print(foo.py_func(*args))
```
which exhibits the same problem as in #8586. Perhaps something like this might work?

```suggestion
    dta = getattr(a, 'dtype', a)
    dtb = getattr(b, 'dtype', b)
    if dta != dtb:
        raise NumbaTypeError(("All the arguments must have the same "
                              f"type/dtype. Have: '{dta}' and '{dtb}'."))
    dt = dta
```

stuartarchibald (commented):
@apmasell I've found a case that I don't think will be handled by the current patch and have noted it with a potential fix inline, please take a look at see what you think. Many thanks.

numba/np/linalg.py
apmasell: I initially did that, but it breaks all the unit tests.

stuartarchibald:
From OOB discussion with @apmasell, we elect to close this as the problem with scalars aliasing in the type system is beyond the scope of this patch.

PR #8602, opened 2022-11-15
Throw error for unsupported dunder methods

apmasell:
Make jitclass throw an error rather than ignore unhandled `__X__` methods.

### DIFF
--- a/numba/experimental/jitclass/boxing.py
+++ b/numba/experimental/jitclass/boxing.py
@@ -151,7 +151,7 @@ def _specialize_box(typ):
             and name.endswith("__")
             and name not in supported_dunders
         ):
-            raise TypeError("Method {0} is not supported".format(name))
+            raise TypeError(f"Method '{name}' is not supported.")
         dct[name] = _generate_method(name, func)
 
     # Inject static methods as class members
--- a/numba/experimental/jitclass/boxing.py
+++ b/numba/experimental/jitclass/boxing.py
@@ -144,12 +144,14 @@ def _specialize_box(typ):
         "__ixor__",
     }
     for name, func in typ.methods.items():
+        if name == "__init__":
+            continue
         if (
             name.startswith("__")
             and name.endswith("__")
             and name not in supported_dunders
         ):
-            continue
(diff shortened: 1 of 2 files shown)


numba/experimental/jitclass/boxing.py
```diff
@@ -144,12 +144,14 @@ def _specialize_box(typ):
...
     for name, func in typ.methods.items():
+        if name == "__init__":
+            continue
         if (
             name.startswith("__")
             and name.endswith("__")
             and name not in supported_dunders
         ):
-            continue
+            raise TypeError("Method {0} is not supported".format(name))
```
stuartarchibald, line 154: Suggest adding a unit test that checks this error message is raised as expected to `numba.tests.test_jitclasses`.

stuartarchibald (commented):
Thanks for the patch. One minor comment to resolve else looks good.

numba/experimental/jitclass/boxing.py
stuartarchibald, line 154:
```suggestion
            raise TypeError(f"Method '{name}' is not supported.")
```
suggest fstring and also putting the name into quotes such that it reads e.g.:
```
Method '__enter__' is not supported.
```

numba/tests/test_jitclasses.py
```diff
@@ -1154,6 +1154,18 @@ def test_jitclass_isinstance(obj):
...
+        with self.assertRaises(TypeError) as e:
+            @jitclass
+            class Foo(object):
+                def __init__(self):
+                    return
+
+                def __enter__(self):
+                    return None
+            Foo()
+        self.assertIn("__enter__", str(e.exception))
```
stuartarchibald, line 1167:
```suggestion
        self.assertIn("Method '__enter__' is not supported.", str(e.exception))
```
I think being more specific here would be beneficial so as to ensure the exception definitely came from the expected location?

stuartarchibald (commented):
@apmasell many thanks for adding the test. Couple of minor things to resolve else looks good.

stuartarchibald (approved):

PR #8604, opened 2022-11-16
Allow empty tuple as size for numpy rng

aseyboldt:
Currently random draws from `np.random` methods fail in jited code if `size=()`.

```python
@numba.njit
def foo():
    return np.random.normal(0, 1, ())

foo()
```
leads to a typing error, because the overloads don't know how to deal with
the empty tuple. Note that `size=None` works fine.

This PR expands the checks in the dispatch to handle the additional case.

### DIFF
--- a/numba/cpython/randomimpl.py
+++ b/numba/cpython/randomimpl.py
@@ -52,6 +52,20 @@ rnd_state_t = ir.LiteralStructType([
 rnd_state_ptr_t = ir.PointerType(rnd_state_t)
 
 
+def _is_valid_size(size):
+    """Check if a size has a valid type for a random function."""
+    if isinstance(size, types.Integer):
+        return True
+
+    if isinstance(size, types.UniTuple):
+        return isinstance(size.dtype, types.Integer)
+
+    if isinstance(size, types.Tuple):
+        return len(size) == 0
+
+    return False
+
+
 def get_state_ptr(context, builder, name):
     """
     Get a pointer to the given thread-local random state
@@ -259,15 +273,15 @@ def random_impl0():
 def random_impl1(size):
     if is_nonelike(size):
         return lambda size: np.random.random()
-    if isinstance(size, types.Integer) or (isinstance(size, types.UniTuple)
-                                           and isinstance(size.dtype,
-                                                          types.Integer)):
+    if _is_valid_size(size):
+
         def _impl(size):
             out = np.empty(size)
             out_flat = out.flat
             for idx in range(out.size):
                 out_flat[idx] = np.random.random()
             return out
+
         return _impl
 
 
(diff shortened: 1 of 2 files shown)


aseyboldt:
On a separate but related note: The current code also doesn't deal with vector inputs for parameters. So 
```python
@numba.njit
def foo():
    return np.random.normal(0, np.zeros(3), 3)

foo()
```
fails as well. Maybe instead of the custom (and identical) code in each overload that does a rough approximation of vectorization it would be better to use numba.vectorize directly? So something along the lines of

```python
@numba.extending.overload(np.random.normal)
def overload_rng_normal(loc, scale, size=None):
    # check types

    def rng_scalar(loc, scale):
        return np.random.normal(loc, scale)

    if size is None:
        return normal_scalar

    if size is not None:
        rng_vec = numba.vectorize(rng_scalar)

        def _impl(loc, scale, size=None):
            out = np.empty(size)
            rng_vec(loc, scale, out=out)
            return out

        return _impl
```

This doesn't currently work however, because as far as I know vectorized functions don't support the `out` argument in nopython mode. Or is there a workaround for this?

numba/cpython/randomimpl.py
```diff
@@ -52,6 +52,20 @@ def const_int(x):
 rnd_state_ptr_t = ir.PointerType(rnd_state_t)
 
 
+def _is_valid_size(size):
```
kc611, line 55: There's a function with similar functionality. Maybe it could be reused.

https://github.com/numba/numba/blob/3e96339a523b3cf4b737b8fff7dba6ee64927584/numba/np/random/generator_methods.py#L59

Otherwise, The current implementation, while being efficient might be hard to read straightaway, and would be better if it simply was a bunch of conditions:

```python
if isinstance(size, types.Integer) or (isinstance(size, types.UniTuple) and isinstance(size.dtype, types.Integer)).....:
    return True
```

numba/cpython/randomimpl.py
```diff
@@ -52,6 +52,20 @@ def const_int(x):
...
+    if isinstance(size, types.Integer):
+        return True
+
+    if isinstance(size, types.UniTuple):
+        return isinstance(size.dtype, types.Integer)
+
+    if isinstance(size, types.Tuple):
+        return len(size) == 0
+
+    return False
```
kc611, line 66: A good addition to this can be a mechanism that raises an error stating that the `size` argument type is wrong.

kc611 (commented):
Left a few suggestions regarding the improvements that could be made to this PR. Otherwise LGTM !

esc:
@aseyboldt this is scheduled for inclusion in 0.58 now -- this is due about 4-6 weeks from now. Do you think you will have time to complete the requested changes by then?

esc:
@kc611 do you want to adopt this one? Don't think there is much left to complete?

Issue #7967, opened 2022-04-07
doc and impl of RewriteArrayExprs are inconsistent


dlee992:
```
def test_single_array_expr():
    def _test_impl(array):
        ret = array >= 0.1
        return ret

    jitted = numba.jit()(_test_impl)
    arr = np.random.rand(10)
    jit_ret = jitted(arr)
```

Code above will be rewritten by `RewriteArrayExprs` Pass. But `array >= 0.1` only has one operator. Is it necessary to rewrite this  expression, since loop fusion and shortcut deforestation optimizations don't seem to work?
And in numba developer doc, it wrote (https://numba.readthedocs.io/en/latest/developer/rewrites.html):

> RewriteArrayExprs.match(): The rewrite pass looks for two or more array operations that form an array expression.

Is it inconsistent between doc and impl?

If this can be confirmed as a real inconsistency, I would like to contribute some code to fix it.

gmarkall:
Is this an SDC issue, or a Numba issue? If it's a Numba issue, can you paste an executable variant of the code that uses Numba to reproduce the issue please?

dlee992:
Oh, sry, I didn't notice it. U can just replace `sdc.jit` with `numba.jit` in this case. SDC didn't re-implement `RewriteArrayExprs` Pass.

Below is the corresponding Numba IR:
```
----------------SSA IR DUMP: test_array_expr.<locals>._test_impl----------------
label 0:
    array1 = arg(0, name=array1)             ['array1']
    $const4.1 = const(float, 0.1)            ['$const4.1']
    ret = array1 >= $const4.1                ['$const4.1', 'array1', 'ret']
    $12return_value.4 = cast(value=ret)      ['$12return_value.4', 'ret']
    return $12return_value.4                 ['$12return_value.4']

______________________________________________________________________
REWRITING (RewriteArrayExprs):
    array1 = arg(0, name=array1)             ['array1']
    $const4.1 = const(float, 0.1)            ['$const4.1']
    ret = array1 >= $const4.1                ['$const4.1', 'array1', 'ret']
    del array1                               []
    del $const4.1                            []
    $12return_value.4 = cast(value=ret)      ['$12return_value.4', 'ret']
    del ret                                  []
    return $12return_value.4                 ['$12return_value.4']
____________________________________________________________
    array1 = arg(0, name=array1)             ['array1']
    $const4.1 = const(float, 0.1)            ['$const4.1']
[...]

PR #8606, opened 2022-11-17
[Doc] Make the RewriteArrayExprs doc more precise

dlee992:
Fix the issue https://github.com/numba/numba/issues/7967#issuecomment-1163891208.

A very little change.


sklam (approved):
Thanks for the patch!

PR #8614, opened 2022-11-22
py3.11 hashing updates

stuartarchibald:
This is based on #8590 but with an attempt to fix the issues with `hash` applied.


sklam (approved):
b7e5aeb14b1b2acdbdb07f76dd303442b08f2f07..3ce7f36dbdbd787c744e7d0b4105a03e6050fe41 looks good

PR #8619, opened 2022-11-23
Added flat iteration logic for random distributions

kc611:
As titled, this PR provides a small optimisation to the loops of random distributions by flattening their iterators.


stuartarchibald (approved):
Thanks for the patch, this should help with making loop generation and analysis easier!

PR #8621, opened 2022-11-24
py3.11 closure updates

stuartarchibald:
Attempting to fix closures on Python 3.11. This branch is #8614 rebased on #8545 at commit 455fd62, with fixes for closures added in 485a1e8.


stuartarchibald:
Failing tests as of 6b101e2 (potential reason/fix in brackets):
1. `numba.tests.test_exceptions.TestRaising.test_reraise_nopython` (think that this is because the Python 3.11 adjustments to exception handling don't know if there's a "live" handling region, error is `UnsupportedError(The re-raising of an exception is not yet supported`)).
2. `numba.tests.test_exceptions.TestRaising.test_reraise_objmode` (as above)
3. `numba.tests.test_profiler.TestProfiler.test_profiler` (Numba `C` code for tracing needs fixing).
4. `numba.tests.test_try_except.TestTryExceptOtherControlFlow.test_objmode` (unknown, error is: `AssertionError: 'unsupported control flow: with-context contains branches...'`).
5. `numba.tests.test_exceptions.TestRaising.test_user_code_error_traceback_nopython` (error message in 3.11 has changed to include markers, fix is trivial).
6. `numba.tests.test_try_except.TestTryExceptOtherControlFlow.test_objmode_output_type` (error message changed, fix is potentially trivial).
7. `numba.tests.test_sort.TestTimsortArrays.test_merge_lo_hi` (this is potentially complicated, the bytecode offsets are being computed incorrectly in the case of an `EXTENDED_ARG`, further, Python 3.11 seems to use `EXTENDED_ARG` opcodes as jump targets.
8. `numba.tests.test_sort.TestTimsortArrays.test_run_timsort` (as above)
9. `numba.tests.test_sort.TestTimsortArrays.test_run_timsort_with_values` (as above)
10. `numba.tests.test_operators.TestMixedIntsOperatorModule.test_pow` (fails with, `AssertionError: when comparing -1261475310744950528 and -1261475310744950487`, there is something a little strange about the compilation of this function (uses `compile_isolated`)).
11. `numba.tests.test_withlifting.TestLiftObj.test_case19_recursion` (unknown, error is `ValueError: The truth value of an array with more than one element is ambiguous.` so it may just be a simply NumPy fix).
12. `numba.tests.test_withlifting.TestLiftCall.test_liftcall5` (unknown, probably CF analysis related, error is `ValueError: Cannot add edge as dest node 34 not in nodes {0, 104}`)
13. `numba.tests.test_import.TestNumbaImport.test_no_accidental_warnings` (needs a `llvmlite` level fix, there's a `DeprecationWarning` coming from the FFI bindings' use of `importlib`.)

stuartarchibald:
5. is fixed by b7b2b1e.

stuartarchibald:
7., 8. and 9. are fixed by 2ec5b0b.

numba/core/byteflow.py
```diff
@@ -836,6 +1055,11 @@ def op_CALL_FUNCTION_EX(self, state, inst):
             varkwarg = None
         vararg = state.pop()
         func = state.pop()
+
+        if PYVERSION == (3, 11):
+            if "$null" in state.peek(1):
```
sklam, line 1060:
```suggestion
            if _is_null_temp_reg(state.peek(1)):
```

numba/tests/test_ir.py
```diff
@@ -432,7 +432,7 @@ def check_diffstr(string, pointing_at=[]):
             blk = z_ir.blocks[label]
             ref = blk.body[:-1]
             idx = None
-            for i in range(len(ref)):
+            for i in range(len(ref) - 1):
```
sklam, line 435: Is this just error in test logic?

sklam (commented):
An early review of the changes:

sklam:
That's some interesting conflict because of my changes to the base branch. Here's the merge commit with all conflict resolved: https://github.com/sklam/numba/commit/e43aea7be9d67bfd3af49133d34bb360d179edb1

numba/tests/test_ir.py
stuartarchibald: Yes. It's supposed to be looking for two adjacent `Del` nodes, the loop body contains an index into `ref` at `i + 1` so as to look ahead one index, hence the `range` needs limiting so as to not lookahead out of bounds.

Issue #6857, opened 2021-03-23
Implement np.nan_to_num


jpivarski:
## Feature request

This is a formal request to implement [np.nan_to_num](https://numpy.org/doc/stable/reference/generated/numpy.nan_to_num.html), as a first step in contributing a PR. (I knew I'd be contributing something to Numba at some point...) I need to read over the [contributing guidelines](https://numba.pydata.org/numba-doc/dev/developer/contributing.html) to set up my environment, but the implementation would just be this:

```python
@numba.extending.overload(numpy.nan_to_num)
def nan_to_num(x, copy=True, nan=0.0, posinf=None, neginf=None):
    if isinstance(x, numba.types.Array):
        def nan_to_num_impl(x, copy=True, nan=0.0, posinf=None, neginf=None):
            if copy:
                out = numpy.copy(x).reshape(-1)
            else:
                out = x.reshape(-1)
            for i in range(len(out)):
                if numpy.isnan(out[i]):
                    out[i] = nan
                if posinf is not None and numpy.isinf(out[i]) and out[i] > 0:
                    out[i] = posinf
                if neginf is not None and numpy.isinf(out[i]) and out[i] < 0:
                    out[i] = neginf
            return out.reshape(x.shape)

    else:
        def nan_to_num_impl(x, copy=True, nan=0.0, posinf=None, neginf=None):
            if numpy.isnan(x):
                return nan
            if posinf is not None and numpy.isinf(x) and x > 0:
                return posinf
            if neginf is not None and numpy.isinf(x) and x < 0:
                return neginf
            return x

    return nan_to_num_impl
```

(Of course, if you want to add this before I have a chance to open a PR myself, please do so!)

It might be better for this to be implemented on `numba.types.ArrayCompatible`, rather than `numba.types.Array`, but I don't know if you're doing that yet.

sklam:
👍 

Regarding `ArrayCompatible`, the code base is using that in quite a few places already. It would be a good idea for new implementation to use that.

jpivarski:
Elsewhere, I see `type_can_asarray`:

https://github.com/numba/numba/blob/d4b18d66e7454ce7c643d382f0d1fc51b79f8959/numba/np/numpy_support.py#L667-L675

That would be convenient because it's one function call in the typing layer and `np.asarray` in the lowered layer. Is this what should be used for new functions?

sklam:
> Is this what should be used for new functions?

Yes.

jpivarski:
[...]

PR #8623, opened 2022-11-28
Adds support for np.nan_to_num

thomasjpfan:
Closes https://github.com/numba/numba/issues/6857

This PR adds support for `np.nan_to_num` with only the 3 first arguments. Note, this implementation only supports real numbers. I would be happy to add complex support if reviewers think it is required.


numba/np/arrayobj.py
```diff
@@ -6376,3 +6376,27 @@ def take_along_axis_impl(arr, indices, axis):
             return _take_along_axis_impl(arr, indices, axis, Ni, Nk,
                                          indices_broadcast_shape)
     return take_along_axis_impl
+
+
+@overload(np.nan_to_num)
+def nan_to_num_impl(X, copy=True, nan=0.0):
```
guilhermeleobas, line 6382: Can you rename `X` -> `x`?

numba/tests/test_np_functions.py
```diff
@@ -4862,6 +4867,35 @@ def not_literal_axis(a, i, axis):
...
+    def test_nan_to_num(self):
+        values = [
+            np.nan,
+            1.0,
+            np.asarray([0.1, 1.0, 0.4]),
+            np.asarray([[0.1, 1.0, 0.4], [0.4, 1.2, 4.0]]),
+            np.asarray([0.1, np.nan, 0.4]),
+            np.asarray([[0.1, np.nan, 0.4], [np.nan, 1.2, 4.0]]),
+        ]
+        nans = [0.0, 4.2]
```
guilhermeleobas, line 4879: Could you copy the tests used by NumPy? You can find them [here](https://github.com/numpy/numpy/blob/8ff45c5bb520db04af8720bf1d34a392a8d2561a/numpy/lib/tests/test_type_check.py#L350-L452). Not all tests will work with Numba. Copy the ones that will work with your changes.

numba/np/arrayobj.py
```diff
@@ -6376,3 +6376,27 @@ def take_along_axis_impl(arr, indices, axis):
...
+            X_ = np.asarray(X)
+            output = np.copy(X_) if copy else X_
+            output_flat = output.flat
+            for i in range(output.size):
+                if np.isnan(output_flat[i]):
+                    output_flat[i] = nan
+            return output
+    else:
+        raise errors.TypingError("The first argument must be a scalar or an "
+                                 "array-like")
```
guilhermeleobas, line 6400: Can you add a test for this branch?

guilhermeleobas (changes requested):
Hi @thomasjpfan, thanks for working on this pull request.  The numba implementation of `np.nan_to_num` is currently failing on a specific test case:

```python
from numba import njit
import numpy as np

@njit
def foo(a):
    return np.nan_to_num(a)

vals = np.array([-np.inf, np.nan, np.inf])
print(foo(vals))  # [-inf   0.  inf]
print(foo.py_func(vals))  # [-1.79769313e+308  0.00000000e+000  1.79769313e+308]
```

Also, can you update `numpysupported.rst` to include `np.nan_to_num`?

numba/np/arrayobj.py
thomasjpfan: This PR was updated to include more input validation and tests to include each code path.

numba/np/arrayobj.py
```diff
@@ -6376,3 +6376,53 @@ def take_along_axis_impl(arr, indices, axis):
...
+            def impl(x, copy=True, nan=0.0):
+                if np.isnan(x):
+                    return nan
+                elif np.isneginf(x):
+                    return np.finfo(np.float64).min
+                elif np.isposinf(x):
+                    return np.finfo(np.float64).max
+                return x
+        else:
+            raise errors.TypingError("Only Integers and Floats are accepted")
```
guilhermeleobas, line 6399: Any reason for not supporting complex values as well? I'm fine with restricting the PR to only int and float values, but you could support complex values as well by reusing the code from float:

```python
if isinstance(x, types.Integer):
    ...
elif isinstance(x, types.Float):
    ...
elif isinstance(x, types.Complex):
    def impl(x, copy=True, nan=0.0):
        r = np.nan_to_num(x.real)
        i = np.nan_to_num(x.imag)
        return complex(r, i)
```

numba/np/arrayobj.py
```diff
@@ -6376,3 +6376,53 @@ def take_along_axis_impl(arr, indices, axis):
...
+                for i in range(output.size):
+                    if np.isnan(output_flat[i]):
+                        output_flat[i] = nan
+                    elif np.isneginf(output_flat[i]):
+                        output_flat[i] = min_inf
+                    elif np.isposinf(output_flat[i]):
+                        output_flat[i] = max_inf
+                return output
+        else:
+            raise errors.TypingError("Only Integers and Floats are accepted")
```
guilhermeleobas, line 6424: Same thing here. You could reuse the code from float to support complex arrays

numba/tests/test_np_functions.py
```diff
@@ -4862,6 +4867,63 @@ def not_literal_axis(a, i, axis):
             gen(0)(arr2d, np.ones((2, 3), dtype=np.uint64))
         self.assertIn("dimensions don't match", str(raises.exception))
 
+    def test_nan_to_num(self):
+        values = [
```
guilhermeleobas, line 4871: Can you add a comment with a link to the NumPy tests?

```python
# https://github.com/numpy/numpy/blob/8ff45c5bb520db04af8720bf1d34a392a8d2561a/numpy/lib/tests/test_type_check.py#L350-L452
```

docs/source/reference/numpysupported.rst
```diff
@@ -499,6 +499,7 @@ The following top-level functions are supported:
 * :func:`numpy.intersect1d` (only first 2 arguments, ar1 and ar2)
 * :func:`numpy.linspace` (only the 3-argument form)
 * :func:`numpy.logspace` (only the 3 first arguments)
+* :func:`numpy.nan_to_num` (only the 3 first arguments)
```
guilhermeleobas, line 502: In case you decide to not support complex values, update this message to say it only support integer and float values.

guilhermeleobas (commented):
Hi @thomasjpfan, thanks again for working on this. I made a few more comments in the PR regarding support complex values or not. It is up to you to add support for them or not.

docs/source/reference/numpysupported.rst
thomasjpfan: I updated the PR to support complex numbers, so this statement should be okay now.

guilhermeleobas (approved):
LGTM. Thanks @thomasjpfan

PR #8624, opened 2022-11-28
DOC: Add guvectorize scalar return example

### LINKED ISSUE #2935: Documentation: guvectorize scalar return value
At my first experience with guvectorize, I wasted a solid 2 hours trying to figure out how you could have the wrapped function return a scalar. It's really not trivial, as if you pass "float64" without brackes in the signature, the wrapped function will receive a plain float.
An example in the documentation would have helped a lot.

e.g.
# Scalar return values
If you want to return a scalar from a function wrapped with guvectorize, you'll need to:
- in the signatures, declare it as ``[:]`` like with a 1-dimensional array
- in the layout, declare it as ``()``
- in the implementation, write to it with ``varname[0] = ...``

Example:
```
@guvectorize(["float64[:], float64[:]"], "(i)->()")
def mysum(x, y):
    "Reimplement x.sum(axis=-1)"
    acc = 0
    for xi in x:
        acc += xi
    y[0] = acc
```

Matt711:
This documentation PR closes #2935. The new example is returning a scalar from guvectorize wrapped function.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,5 @@
 repos:
-- repo: https://gitlab.com/pycqa/flake8
+- repo: https://github.com/PyCQA/flake8
   rev: 3.7.8
   hooks:
   - id: flake8
--- a/docs/source/user/vectorize.rst
+++ b/docs/source/user/vectorize.rst
@@ -221,15 +221,18 @@ complicated inputs, depending on their shapes::
 Scalar return values
 --------------------
 
-Now suppose we want to return a scalar value from :func:`~numba.guvectorize`. To do this, we need to:
+Now suppose we want to return a scalar value from 
+:func:`~numba.guvectorize`. To do this, we need to:
 
-* in the signatures, declare the scalar return with ``[:]`` like a 1-dimensional array (eg. ``int64[:]``),
+* in the signatures, declare the scalar return with ``[:]`` like 
+  a 1-dimensional array (eg. ``int64[:]``),
 
 * in the layout, declare it as ``()``,
 
 * in the implementation, write to the first element (e.g. ``res[0] = acc``).
 
-The following example function computes the sum of the 1-dimensional array (``x``) plus the scalar (``y``) and returns it as a scalar:
+The following example function computes the sum of the 1-dimensional 
+array (``x``) plus the scalar (``y``) and returns it as a scalar:
 
 .. literalinclude:: ../../../numba/tests/doc_examples/test_examples.py
    :language: python
@@ -239,11 +242,16 @@ The following example function computes the sum of the 1-dimensional array (``x`
    :dedent: 12
    :linenos:
 
-Now apply the wrapped function over the array, we get a scalar value as the output.
+Now if we apply the wrapped function over the array, we get a scalar 
+value as the output:
 
-    >>> a = np.arange(5)
-    >>> g(a,2)
-    20
+.. literalinclude:: ../../../numba/tests/doc_examples/test_examples.py
+   :language: python
+   :caption: from ``test_guvectorize_scalar_return`` of ``numba/tests/doc_examples/test_examples.py``
+   :start-after: magictoken.ex_guvectorize_scalar_return_call.begin
+   :end-before: magictoken.ex_guvectorize_scalar_return_call.end
+   :dedent: 12
+   :linenos:
 
 
 .. _overwriting-input-values:
--- a/docs/source/user/vectorize.rst
+++ b/docs/source/user/vectorize.rst
@@ -219,23 +219,27 @@ complicated inputs, depending on their shapes::
 .. _scalar-return-values:
 
 Scalar return values
-------------------------
+--------------------
 
-Now suppose we want to return a scalar value from :func:`~numba.guvectorize`::
+Now suppose we want to return a scalar value from :func:`~numba.guvectorize`. To do this, we need to:
 
-   @guvectorize([(int64[:], int64, int64[:])], '(n),()->()')
-   def g(x, y, res):
-       acc=0
-       for i in range(x.shape[0]):
-           acc += x[i] + y
-       res[0]=acc
+* in the signatures, declare the scalar return with ``[:]`` like a 1-dimensional array (eg. ``int64[:]``),
 
-The modified function takes the sum of the 1-dimensional array (``x``) plus the scalar (``y``).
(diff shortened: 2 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -1,5 +1,5 @@
 repos:
-- repo: https://gitlab.com/pycqa/flake8
+- repo: https://github.com/PyCQA/flake8
```
Matt711, line 2: I made the suggested change to the link from [here](https://github.com/AICoE/aicoe-ci/issues/53)

docs/source/user/vectorize.rst
```diff
@@ -388,7 +415,7 @@ compilation works for a :class:`~numba.GUFunc`::
    >>> res = np.zeros_like(x)
    >>> g(x, y, res)
    >>> res
-   array([5, 6, 7, 8, 9])
+   array([10, 11, 12, 13, 14])
```
Matt711, line 418: Ther output in this example was incorrect.

docs/source/user/vectorize.rst
```diff
@@ -215,6 +215,33 @@ complicated inputs, depending on their shapes::
...
+       for i in range(x.shape[0]):
+           acc += x[i] + y
+       res[0]=acc
+
+The modified function takes the sum of the 1-dimensional array (``x``) plus the scalar (``y``).
+There are a few important changes that allow the wrapped function to return a scalar:
+
+* in the signatures, declare the scalar return with ``[:]`` like a 1-dimensional array (eg. ``int64[:]``)
+* in the layout, declare it as ``()``
+* in the implementation, write the return value to output. (e.g. ``res[0]=acc``)
```
gmarkall, line 238: I think it would be easier to digest if we write the explanation of how to return scalars then follow with an example. E.g.:

```suggestion
Now suppose we want to return a scalar value from :func:`~numba.guvectorize`. To do this, we need to:

* in the signatures, declare the scalar return with ``[:]`` like a 1-dimensional array (eg. ``int64[:]``),
* in the layout, declare it as ``()``,
* in the implementation, write to the first element (e.g. ``res[0] = acc``).

The following example function computes the sum of the 1-dimensional array (``x``) plus the scalar (``y``) and returns it as a scalar:
```

(then the example follows afterwards). Note there are some minor edits to each point in the suggestion above too.

gmarkall (changes requested):
@Matt711 Many thanks for the PR!

For new examples, it would be best if we could create them as doctests, rather than inline in the documentation source - that way, their continued correctness (and flake8 compliance etc.) is ensured by CI testing. Examples of doctests can be seen in the Examples section:

- Doc source: https://raw.githubusercontent.com/numba/numba/main/docs/source/user/examples.rst
- Code: https://github.com/numba/numba/blob/main/numba/tests/doc_examples/test_examples.py

In a nutshell, the example code is written as a test that gets executed by the testsuite, and the relevant portions for inclusion in the docs are delimited with `# magictoken.*` comments. These portions are then included in the documentation source using the `literalinclude` directive.

Would you be willing to convert your new example to the doctest format? (If you'd like to go further, I think it would also be great to convert other examples from the page to doctests, but I'd suggest not doing that as part of this PR in order to make it easier to finish quickly - conversion of the other examples could be considered for a separate PR).

docs/source/user/vectorize.rst
Matt711: Thanks for the review. I'll add this example as a doc test and address the other samples in a follow-up PR.

docs/source/user/vectorize.rst
```diff
@@ -215,6 +215,37 @@ complicated inputs, depending on their shapes::
...
+
+.. literalinclude:: ../../../numba/tests/doc_examples/test_examples.py
+   :language: python
+   :caption: from ``test_guvectorize_scalar_return`` of ``numba/tests/doc_examples/test_examples.py``
+   :start-after: magictoken.ex_guvectorize_scalar_return.begin
+   :end-before: magictoken.ex_guvectorize_scalar_return.end
+   :dedent: 12
+   :linenos:
+
+Now apply the wrapped function over the array, we get a scalar value as the output.
```
gmarkall, line 242:
```suggestion
Now if we apply the wrapped function over the array, we get a scalar value as the output:
```

numba/tests/doc_examples/test_examples.py
```diff
@@ -197,6 +197,19 @@ def func_mt(*args):
...
+            from numba import guvectorize, int64
+
+            @guvectorize([(int64[:], int64, int64[:])], '(n),()->()')
+            def g(x, y, res):
+                acc = 0
+                for i in range(x.shape[0]):
+                    acc += x[i] + y
+                res[0] = acc
+            # magictoken.ex_guvectorize_scalar_return.end
+
```
gmarkall, line 212: In addition to checking that the example code executes without error, we can also ensure properties of the example hold - in this case, we can check that the result is an integer and that its value is as expected. This also helps to ensure that the example does what we state it does in the documentation.

For this particular example, we can convert the invocation of `g()` to be part of the doctest, and create a second block enclosed by magic tokens for inclusion into the example documentation - then we can assert that the correct result is given:

```suggestion

            # magictoken.ex_guvectorize_scalar_return_call.begin
            a = np.arange(5)
            result = g(a, 2)
            # At this point, result == 20.
            # magictoken.ex_guvectorize_scalar_return.end

            self.assertIsInstance(result, np.integer)
            self.assertEqual(result, 20)


```

docs/source/user/vectorize.rst
```diff
@@ -215,6 +215,37 @@ complicated inputs, depending on their shapes::
...
+
+    >>> a = np.arange(5)
+    >>> g(a,2)
+    20
```
gmarkall, line 246: Could you convert this to a `literalinclude` of the `magictoken.ex_guvectorize_scalar_return_call.begin`-enclosed block suggested below please?

gmarkall (changes requested):
Many thanks for converting the example to a doctest - this is a good step foward. I have a couple more suggestions:

- As noted on the diff, we could test more in the doctest, and I've added a suggestion of how.
- Generally we wrap documentation source at 80 characters - I know the file you're contributing to already isn't perfect, but if you could wrap your added text at 80 characters (where appropriate, I realise this can be hard to do for some of the lines in the `literalinclude` directive) so at least it isn't introducing any new violations of that ideal, that would be great.

gmarkall (approved):
Many thanks for the updates! This looks good to me!

PR #8625, opened 2022-11-28
Refactor `test_ufuncs`

gmarkall:
This PR refactors `test_ufuncs` in preparation for the ufunc tests being reused for CUDA ufunc tests, based on work in PR #8294. The original motivation was to replace usage of the `CompilationCache`, as I did not want to implement one for CUDA (and the nested context work that would need to be done to implement `compile_isolated`, on which it relies, for CUDA). This has the additional benefit of being a move towards a single path through Numba compilation (the jit decorators) rather than the use of `compile_isolated`, and hopefully eventual removal of nested contexts as used by `compile_isolated`. However, the total changes now encompass:

- Fix flake8 for `test_ufuncs.py` and enforce it - because the noncompliant format was quite annoying to work with.
- Fix some errors in the tests - some dtypes that we called compiled functions with did not actually match the dtypes they were compiled for. This was possible with `compile_isolated` because jumping straight into the resulting `CompileResult`'s entry point does no type checking - when we compile with the `@jit` or `@njit` decorators in eager mode, type checking is performed.
- Add an `lru_cache` for the compiled functions in `BaseUFuncTest`. This was to recover some lost speed from dropping `CompilationCache`, but it seems like the end result is faster than before.

Timings on my system (i7-6700K) for `python runtests.py numba.tests.test_ufuncs -vf -m`:

- `main`: `Ran 4107 tests in 59.191s`
- This branch, without `@lru_cache`: `Ran 4108 tests in 71.164s`
- This branch as-is: `Ran 4108 tests in 54.814s`

Note there is now an extra test running - one of the flake8 errors was a duplicated method name, so one of the tests shadowed another and prevented it from running.

For aid of reviewing, commits are:

- First commit - flake8 fixes
- Second commit - the substance of the changes
- Final commit - add the LRU cache.

### DIFF
--- a/numba/tests/test_ufuncs.py
+++ b/numba/tests/test_ufuncs.py
@@ -8,11 +8,9 @@ import operator
 import numpy as np
 
 import unittest
-from numba import typeof, njit, guvectorize
+from numba import guvectorize, jit, njit, typeof, vectorize
 from numba.core import types
-from numba.core.compiler import compile_isolated, Flags
 from numba.np.numpy_support import from_dtype
-from numba import jit, vectorize
 from numba.core.errors import LoweringError, TypingError, NumbaTypeError
 from numba.tests.support import TestCase, CompilationCache, MemoryLeakMixin
 from numba.core.typing.npydecl import supported_ufuncs
@@ -23,20 +21,6 @@ from numba.np import ufunc_db
 
 is32bits = tuple.__itemsize__ == 4
 iswindows = sys.platform.startswith('win32')
-
-# NOTE: to test the implementation of Numpy ufuncs, we disable rewriting
-# of array expressions.
-
-enable_pyobj_flags = Flags()
-enable_pyobj_flags.enable_pyobject = True
-enable_pyobj_flags.no_rewrites = True
-
-no_pyobj_flags = Flags()
-no_pyobj_flags.no_rewrites = True
-
-enable_nrt_flags = Flags()
-enable_nrt_flags.nrt = True
-enable_nrt_flags.no_rewrites = True
 
 
 def _unimplemented(func):
@@ -142,6 +126,11 @@ class BaseUFuncTest(MemoryLeakMixin):
 
         self.cache = CompilationCache()
 
+    def _compile(self, pyfunc, args, nopython=True, nrt=False):
+        # NOTE: to test the implementation of Numpy ufuncs, we disable
+        # rewriting of array expressions.
+        return jit(args, nopython=nopython, _nrt=nrt, no_rewrites=True)(pyfunc)
+
     def _determine_output_type(self, input_type, int_output_type=None,
                                float_output_type=None):
         ty = input_type
@@ -168,7 +157,7 @@ class BaseUFuncTest(MemoryLeakMixin):
 
 class TestUFuncs(BaseUFuncTest, TestCase):
 
-    def basic_ufunc_test(self, ufunc, flags=no_pyobj_flags,
+    def basic_ufunc_test(self, ufunc, nopython=True,
                          skip_inputs=[], additional_inputs=[],
                          int_output_type=None, float_output_type=None,
                          kinds='ifc', positive_only=False):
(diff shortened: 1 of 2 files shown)


numba/tests/test_ufuncs.py
```diff
@@ -751,38 +753,42 @@ def test_broadcasting(self):
...
+            return np.add(a0, a1)
+        arr_ty = types.Array(types.uint64, 1, 'C')
+
         with self.assertRaises(NumbaTypeError):
-            def myadd(a0, a1):
-                return np.add(a0, a1)
-            arr_ty = types.Array(types.uint64, 1, 'C')
-            cr = compile_isolated(myadd, (arr_ty, arr_ty),
-                                  flags=no_pyobj_flags)
+            njit((arr_ty, arr_ty), _nrt=False)(myadd)
```
stuartarchibald, line 766: Should this use `self._compile` ?

numba/tests/test_ufuncs.py
```diff
@@ -1371,7 +1386,8 @@ def _arg_for_type(self, a_letter_type, index=0):
                             dtype=a_letter_type)
         elif a_letter_type in 'FD':
             # complex
-            # Note `-1j` is different on 2.x and 3.x, hence the explicit spelling
+            # Note `-1j` is different on 2.x and 3.x, hence the explicit
+            # spelling
```
stuartarchibald, line 1390:
```suggestion
```
remove python 2 reference?

numba/tests/test_ufuncs.py
```diff
@@ -1178,12 +1191,11 @@ class TestScalarUFuncs(TestCase):
     This class tests regular uses. A subclass tests the no python backend.
     """
 
-    _compile_flags = enable_pyobj_flags
+    _nopython = False
```
stuartarchibald, line 1194: Whilst this is just a refactor, I'm puzzled by the use of `enable_pyobj_flags`/`nopython=False`. IIRC these flags essentially permit "fallback" to `object mode` should the `nopython mode` pipeline fail in compilation, but everything being tested looks like it would succeed under `nopython mode`, this is tested again in `TestScalarUFuncsNoPython` where `nopython mode` is enforced? Is there any need for this?

stuartarchibald (commented):
Thanks for doing this refactoring @gmarkall. There's a couple of minor things to resolve and a question about the use of flags, otherwise looks good.

numba/tests/test_ufuncs.py
gmarkall: It's been this way since the tests were added in 2014, in 7c20a342cde38793a735382e033d17f190a49291. There's probably no need for it now, but I can see that back in 2014 it might have been handy to be able to tell if nopython mode stopped working whilst object mode succeeded. Do you want this removing and all tests to use just nopython mode always? Or shall we leave test semantics as they are, as this is just a refactor to get towards being able to remove `compiled_isolated` / `NestedContext`, and use the ufunc tests for CUDA ufunc testing?

numba/tests/test_ufuncs.py
gmarkall: I'd say it shouldn't - the result won't be cached because it's expected to fail, and it's more transparent what's going on in the test when the `njit` function is used directly.

stuartarchibald: ah I see, thanks, makes sense to leave as-is.

gmarkall:
@stuartarchibald As discussed OOB, I've made it so that all tests run in nopython mode only. A lot of boilerplate went into passing a nopython mode flag around! I've left the NRT uses in for now as changing / removing that could be a little more complex, as its presence seems to be used as a proxy to catch / cause other issues.

stuartarchibald:
> @stuartarchibald As discussed OOB, I've made it so that all tests run in nopython mode only. A lot of boilerplate went into passing a nopython mode flag around! I've left the NRT uses in for now as changing / removing that could be a little more complex, as its presence seems to be used as a proxy to catch / cause other issues.

Thanks @gmarkall, this in response to https://github.com/numba/numba/pull/8625#discussion_r1035740633. The use of the object mode "fallback" seems unnecessary as the tests were all compiling in `nopython` mode anyway. In future, were we to want to explicitly test object mode ufuncs we'd need to force object mode explicitly.

numba/tests/test_ufuncs.py
gmarkall: Well, as we concluded OOB, this is just daft, so https://github.com/numba/numba/pull/8625/commits/9ecfe70d3408836dc72b30a43fbded52b326c2ea changes the test to not need to disable NRT. I've still left the use of `njit` here so it's clear what's going on.

gmarkall:
@stuartarchibald All discussed changes are now pushed - hopefully CI remains green!

stuartarchibald (approved):

PR #8626, opened 2022-11-30
[unicode-PERF]: use optmized BM algorithm to replace the brute-force finder

dlee992:
Hi, after some digging, I found cpython has at least 3 ways to tackle string-matching operations, based on some metrics on `needle` and `haystack` lengths:
1. `find_char`: use `memchr` as possible as we can, which using hand-tuned assembly code with vectorized instructions; 	
2. `default_find`: use `Boyer-Moore` algorithm with `Horspool` (`O(1)` space) and `Sunday` (`bloom filter`) optimizations (merged in 2006);
3. `_two_way_find`: a more advanced but longer preprocessing algorithm (merged in 2021).

Indeed, I haven't known how to implement a vectorized find for char search (any idea? I would like to contribute, but don't know exactly the details), so right now I just implemented the second algorithm for numba default string find underlying impl.

I haven't perf the speedup ratio, but choose to trust cpython's decision. If needed, I could make up some perf tests to show how much reduced comparsions w/ this PR.

PS, during the development, I found numba doesn't support python `reversed` keyword, is this owing to some numba limitations? Or we just didn't have time to implement it?

### DIFF
--- a/numba/cpython/unicode.py
+++ b/numba/cpython/unicode.py
@@ -71,9 +71,7 @@ _py38_or_later = utils.PYVERSION >= (3, 8)
 _MAX_UNICODE = 0x10ffff
 
 # https://github.com/python/cpython/blob/1960eb005e04b7ad8a91018088cfdb0646bc1ca0/Objects/stringlib/fastsearch.h#L31    # noqa: E501
-# I fix the _bloom_width, not chosen from (32, 64, 128) based on machines.
-#   Not sure if impacting the bloom-filter effectiveness.
-_BLOOM_WIDTH = 64
+_BLOOM_WIDTH = types.intp.bitwidth
 
 # DATA MODEL
 
@@ -612,7 +610,6 @@ def _bloom_check(mask, ch):
 
 
 # https://github.com/python/cpython/blob/1960eb005e04b7ad8a91018088cfdb0646bc1ca0/Objects/stringlib/fastsearch.h#L550    # noqa: E501
-# XXX: Why we don't have the FAST_COUNT flag as cpython?
 @register_jitable
 def _default_find(data, substr, start, end):
     """Left finder."""
@@ -623,8 +620,8 @@ def _default_find(data, substr, start, end):
     gap = mlast = m - 1
     last = _get_code_point(substr, mlast)
 
-    # How to coerce mask to be an uint64 as cpython does?
-    mask = _bloom_add(0, last)
+    zero = types.intp(0)
+    mask = _bloom_add(zero, last)
     for i in range(mlast):
         ch = _get_code_point(substr, i)
         mask = _bloom_add(mask, ch)
--- a/numba/cpython/unicode.py
+++ b/numba/cpython/unicode.py
@@ -70,6 +70,11 @@ _py38_or_later = utils.PYVERSION >= (3, 8)
 # https://github.com/python/cpython/blob/1d4b6ba19466aba0eb91c4ba01ba509acf18c723/Objects/unicodeobject.c#L84-L85    # noqa: E501
 _MAX_UNICODE = 0x10ffff
 
+# https://github.com/python/cpython/blob/1960eb005e04b7ad8a91018088cfdb0646bc1ca0/Objects/stringlib/fastsearch.h#L31    # noqa: E501
+# I fix the _bloom_width, not chosen from (32, 64, 128) based on machines.
+#   Not sure if impacting the bloom-filter effectiveness.
+_BLOOM_WIDTH = 64
+
 # DATA MODEL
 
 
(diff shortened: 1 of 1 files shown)


dlee992:
FYI, 

1. this is the python blog in 2006, describing how they implemented the exact BM algorithm in that period of Python: https://web.archive.org/web/20201107074620/http://effbot.org/zone/stringlib.htm
2. this is about all kinds of string fast matching algorithms: http://www-igm.univ-mlv.fr/~lecroq/string/index.html

numba/cpython/unicode.py
```diff
@@ -593,6 +598,117 @@ def unicode_sub_check_type(ty, name):
...
+    return mask
+
+
+@register_jitable
+def _bloom_check(mask, ch):
+    return mask & (1 << (ch & (_BLOOM_WIDTH - 1)))
+
+
+# https://github.com/python/cpython/blob/1960eb005e04b7ad8a91018088cfdb0646bc1ca0/Objects/stringlib/fastsearch.h#L550    # noqa: E501
+# XXX: Why we don't have the FAST_COUNT flag as cpython?
```
guilhermeleobas, line 615: CPython defines `FAST_COUNT` as 0. I think you can do the same as well.
https://github.com/python/cpython/blob/1960eb005e04b7ad8a91018088cfdb0646bc1ca0/Objects/stringlib/fastsearch.h#L24

guilhermeleobas: Actually, I think you implemented the `FAST_SEARCH` mode. So, the `mode` variable is not needed, and thus, unless you implement the three mode, the flags (`FAST_COUNT`, `FAST_SEARCH`, `FAST_RSEARCH`) defined in CPython will not be used.

guilhermeleobas (changes requested):
Hi @dlee992, thanks for working on this. I've done an initial review, but the code looks pretty similar to the one implemented in CPython.

Question: Have you measure the time using the new algorithm vs the older implementation?

dlee992:
> Question: Have you measure the time using the new algorithm vs the older implementation?

No, not yet. I can work on that, make up some test cases to see the time difference. Will update it here when the perf test is ready.

numba/cpython/unicode.py
dlee992: OK, I'll remove this question comment.

numba/cpython/unicode.py
```diff
@@ -593,6 +598,117 @@ def unicode_sub_check_type(ty, name):
...
+    """Left finder."""
+    m = len(substr)
+    if m == 0:
+        return start
+
+    gap = mlast = m - 1
+    last = _get_code_point(substr, mlast)
+
+    # How to coerce mask to be an uint64 as cpython does?
+    mask = _bloom_add(0, last)
```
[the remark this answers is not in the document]
dlee992: OK, will modify it.

numba/cpython/unicode.py
```diff
@@ -593,6 +598,117 @@ def unicode_sub_check_type(ty, name):
...
+    for i in range(mlast):
+        ch = _get_code_point(substr, i)
+        mask = _bloom_add(mask, ch)
+        if ch == last:
+            gap = mlast - i - 1
```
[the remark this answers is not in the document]
dlee992: Indeed, I move the `mask = _bloom_add(mask, last)` before the loop.

dlee992:
Here's a made-up test case, copied from the method `test_find` in `test_unicode.py`, and add a very long test case in it. U can put  this as a method in the class `UnicodeTest` in `test_unicode.py` and rerun this.

<details>

```python
    def test_find_perf(self):
        pyfunc = find_usecase
        cfunc = njit(pyfunc)

        # pre-run cfunc
        cfunc("abc", "a")

        default_subs = [
            (s, ['', 'xx', s[:-2], s[3:], s]) for s in UNICODE_EXAMPLES
        ]

        cpython_subs = [
            ('a' * 100 + '\u0102', ['\u0102', '\u0201', '\u0120', '\u0220']),
            ('a' * 100 + '\U00100304', ['\U00100304', '\U00100204',
                                        '\U00102004']),
            ('\u0102' * 100 + 'a', ['a']),
            ('\U00100304' * 100 + 'a', ['a']),
            ('\U00100304' * 100 + '\u0102', ['\u0102']),
            ('a' * 100, ['\u0102', '\U00100304', 'a\u0102', 'a\U00100304']),
            ('\u0102' * 100, ['\U00100304', '\u0102\U00100304']),
            ('\u0102' * 100 + 'a_', ['a_']),
            ('\U00100304' * 100 + 'a_', ['a_']),
            ('\U00100304' * 100 + '\u0102_', ['\u0102_']),
        ]

        start_time = time.time()
        for _ in range(1000):
            for s, subs in default_subs + cpython_subs:
                for sub_str in subs:
                    cfunc(s, sub_str)
        print(f"consumed time in short test cases: {time.time() - start_time}")

        # made-up cases for fast search
        for leng in [10, 100, 1000, 10000]:
            s = "abcdefg" * leng + "hijk"
            sub_str = "hijk"
            start_time = time.time()
            for _ in range(1000):
                cfunc(s, sub_str)
            print(f"consumed time with len={len(s)}: {time.time() - start_time}")
``` 
</details>

Consumed time in Numba 0.56.3:
```bash
consumed time in short test cases: 0.19890618324279785
consumed time with len=74: 0.003802061080932617
consumed time with len=704: 0.006798982620239258
consumed time with len=7004: 0.03879284858703613
consumed time with len=70004: 0.44359922409057617
```
Consumed time in Numba main with this PR:
```bash
total time consumed: 0.18944501876831055
consumed time with len=74: 0.004054069519042969
consumed time with len=704: 0.003863811492919922
consumed time with len=7004: 0.007459163665771484
consumed time with len=70004: 0.04593682289123535
```
In short test cases, consumed time are similar (perhaps a minor degradation with this PR, but marginal, I think).
When putting the `sub_str` in the end of a very long `s`, the acceleration ratio is 1.76, 5.2, and 9.66, which is nice.

guilhermeleobas:
>  When putting the sub_str in the end of a very long s, the acceleration ratio is 1.76, 5.2, and 9.66, which is nice.

Can you share the source code for these?

dlee992:
> Can you share the source code for these?

I have already shared the code, u can unfold the `Details` button in https://github.com/numba/numba/pull/8626#issuecomment-1346083189, then u will see the source code!

@guilhermeleobas

guilhermeleobas:
Oh, sorry. I'll check it tomorrow morning.

guilhermeleobas:
Thanks @dlee992. I ran the code and your approach is way faster than the old one:

```bash
# your branch
consumed time in short test cases: 0.1025538444519043
consumed time with len=74: 0.0019850730895996094
consumed time with len=704: 0.002277851104736328
consumed time with len=7004: 0.006390094757080078
consumed time with len=70004: 0.04852104187011719
consumed time with len=700004: 0.45890283584594727
consumed time with len=7000004: 4.588933944702148

# Numba main
consumed time in short test cases: 0.10606002807617188
consumed time with len=74: 0.002021312713623047
consumed time with len=704: 0.003590106964111328
consumed time with len=7004: 0.01938009262084961
consumed time with len=70004: 0.17744994163513184
consumed time with len=700004: 1.744581937789917
consumed time with len=7000004: 17.453103065490723
```

guilhermeleobas (approved):

dlee992:
Hi, @guilhermeleobas and @sklam 

Since no obvious side effects with this PR, I feel this one can also be put into [Numba 0.57 RC](https://github.com/numba/numba/milestone/63) if possible.

guilhermeleobas:
Yes! I just tagged to 0.57

PR #8633, opened 2022-12-02
DOC: Convert vectorize and guvectorize examples to doctests

Matt711:
This is the follow-up PR to #8624. I'm converting the other `vectorize` and `guvectorize` examples to doctests.


gmarkall:
gpuci run tests

numba/tests/doc_examples/test_examples.py
```diff
@@ -220,6 +353,189 @@ def g(x, y, res):
...
+            result = f(3,4)
+            # result == 12
+
+            print(f.types)
+            # ['ll->l']
+            # magictoken.ex_vectorize_dynamic_call_one.end
+
+            self.assertEqual(result, 12)
+            correct = [['ll->l'], ['ll->q']]
+            self.assertIn(f.types, correct)
```
Matt711, line 439: Windows test we're failing due to differences in type signatures. On Linux, the output of `f.types == ['ll->l']`, but on Windows 64 the output of `f.types == ['ll->q']`. Do you know what 'q' is? I believe 'l' is for long. I created an issue related to this here #8668. Should I raise it in Numpy instead because there may be other inconsistencies between Windows and Linux?

For now, I'm just testing for both type signatures, but I can change it if needed.

gmarkall: `'q'` is a `long long int`. I think this is different on Windows because it has the LLP64 (long long integers and pointers are 64-bit, long integers are 32 bits) data model, whereas 64-bit Linux is LP64 (long integers and pointers are 64 bits) - so it needs to be `q` because it it were `l` on Windows that would be a 32-bit return value. (See the 64-bit Data Models table in https://en.wikipedia.org/wiki/64-bit_computing and the section that it's in).

There are inconsistencies between Windows and Linux, but in this case I believe it to be a platform difference rather than an unexpected issue / bug, so the test will need to check differently on Windows (we have quite a few other Windows-specific deviations in the test suite).

numba/tests/doc_examples/test_examples.py
```diff
@@ -197,6 +198,139 @@ def func_mt(*args):
...
+            # result == array([2, 3, 4, 5, 6])
+            # magictoken.ex_guvectorize_call_one.end
+
+            self.assertIsInstance(result, np.ndarray)
+            correct = np.array([2, 3, 4, 5, 6])
+            np.testing.assert_array_equal(result, correct)
+
+            # magictoken.ex_guvectorize_call_two.begin
+            a = np.arange(6).reshape(2, 3)
+            # a == array([[0, 1, 2], [3, 4, 5]])
```
gmarkall, line 316: It might be nice for readability to preserve the visual shape of the arrays in the output:

```suggestion
            # a == array([[0, 1, 2], 
            #             [3, 4, 5]])
```

numba/tests/doc_examples/test_examples.py
```diff
@@ -197,6 +198,139 @@ def func_mt(*args):
...
+
+            result1 = g(a, 10)
+            # result1 == array([[10, 11, 12], [13, 14, 15]])
```
gmarkall, line 319:
```suggestion
            # result1 == array([[10, 11, 12], 
            #                   [13, 14, 15]])
```

numba/tests/doc_examples/test_examples.py
```diff
@@ -220,6 +354,204 @@ def g(x, y, res):
...
+
+            # magictoken.ex_guvectorize_overwrite_call_two.begin
+            invals = np.zeros(shape=(3, 3), dtype=np.float32)
+            # invals == array([[0., 0., 0.],
+            #                  [0., 0., 0.],
+            #                  [0., 0., 0.]], dtype=float32)
+            outvals = init_values(invals)
+            # outvals == array([[4.2, 4.2, 4.2],
+            #                   [4.2, 4.2, 4.2],
+            #                   [4.2, 4.2, 4.2]])
```
gmarkall, line 403: I would suggest also showing `invals` is still all zeros here - it's no longer clear that this is the case - in the original text, `invals` was printed out after the call to `init_values()`.

docs/source/user/vectorize.rst
```diff
@@ -260,39 +264,40 @@ Overwriting input values
...
-   array([[0., 0., 0.],
-          [0., 0., 0.],
-          [0., 0., 0.]], dtype=float32)
+.. literalinclude:: ../../../numba/tests/doc_examples/test_examples.py
+   :language: python
+   :caption: from ``test_guvectorize_overwrite`` of ``numba/tests/doc_examples/test_examples.py``
+   :start-after: magictoken.ex_guvectorize_overwrite_call_two.begin
+   :end-before: magictoken.ex_guvectorize_overwrite_call_two.end
+   :dedent: 12
+   :linenos:
```
gmarkall, line 300: I can't write a comment on a line unmodified by this PR, but it looks like there is one block below (beginning on line 305 / 310 below) that is unconverted. This block was probably added after you started this PR, which might be why it's unconverted.

numba/tests/doc_examples/test_examples.py
```diff
@@ -220,6 +354,204 @@ def g(x, y, res):
...
+            def g(x, y, res):
+                for i in range(x.shape[0]):
+                    res[i] = x[i] + y
+            # magictoken.ex_guvectorize_dynamic.end
+
+            # magictoken.ex_guvectorize_dynamic_call_one.begin
+            x = np.arange(5, dtype=np.int64)
+            y = 10
+            res = np.zeros_like(x)
+            # res == array([10, 11, 12, 13, 14])
```
gmarkall, line 509: I think this comment should be after the call to `g()` - at this point in the code it will be zeroes.

numba/tests/doc_examples/test_examples.py
```diff
@@ -220,6 +354,204 @@ def g(x, y, res):
...
+                correct = ['qq->q']
+            else:
+                correct = ['ll->l']
+            self.assertEqual(g.types, correct)
+
+            # magictoken.ex_guvectorize_dynamic_call_two.begin
+            x = np.arange(5, dtype=np.double)
+            y = 2.2
+            res = np.zeros_like(x)
+            # res == array([2.2, 3.2, 4.2, 5.2, 6.2])
```
gmarkall, line 528: This also needs moving after `g()`.

numba/tests/doc_examples/test_examples.py
```diff
@@ -197,6 +198,139 @@ def func_mt(*args):
...
+
+            # magictoken.ex_vectorize_return_call_two.begin
+            a = np.linspace(0, 1, 6)
+            result = f(a, a)
+            # Now, result == array([0. , 0.4, 0.8, 1.2, 1.6, 2. ])
+            # magictoken.ex_vectorize_return_call_two.end
+
+            self.assertIsInstance(result, np.ndarray)
+            correct = np.array([0., 0.4, 0.8, 1.2, 1.6, 2. ])
+            np.testing.assert_array_almost_equal(result, correct, -16)
```
gmarkall, line 243: In most cases we use `assert_allclose()` - it's probably a little more tolerant than the check here - I fear this might be a little too stringent and introduce a spurious failure at some point in the future:

```suggestion
            np.testing.assert_allclose(result, correct)
```

gmarkall (changes requested):
Many thanks for the update! I think this is in pretty good shape overall, with just a few minor comments on the diff.

gmarkall (approved):
Thanks for the fixes - this is now looking good!

PR #8634, opened 2022-12-02
Add new method is_fp16_supported

testhound:
This pull requests adds support for a new method 'is_fp16_supported' so that user code can determine whether fp16 support is available on the gpu.


testhound:
gpuci run tests

testhound:
gpuci run tests

testhound:
@gmarkall updated based on your feedback.

esc:
I ran this on the farm as `numba_smoketest_cuda_yaml_138` and it was fine. 🧪

docs/source/cuda-reference/kernel.rst
```diff
@@ -469,10 +469,13 @@ precision parts of the CUDA Toolkit documentation.
 16-bit Floating Point Intrinsics
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+.. function:: numba.cuda.is_fp16_supported ()
+
+   Return ``True`` if 16-bit floatint point is supported, ``False`` otherwise.
```
gmarkall, line 474:
```suggestion
   Return ``True`` if 16-bit floats are supported, ``False`` otherwise.
```

docs/source/cuda-reference/kernel.rst
```diff
@@ -469,10 +469,13 @@ precision parts of the CUDA Toolkit documentation.
 16-bit Floating Point Intrinsics
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+.. function:: numba.cuda.is_fp16_supported ()
```
gmarkall, line 472: I did make a comment on #8744 about not wanting to introduce additional uses of the term `fp16`, but I'm a bit torn here because the the module with the intrinsics is named `numba.cuda.fp16`. However, given that a user would normally expect to use operators rather than intrinsics, I think I'd suggest we call it `is_float16_supported()`:

```suggestion
.. function:: numba.cuda.is_float16_supported ()
```

numba/cuda/api.py
```diff
@@ -85,6 +85,14 @@ def is_cuda_array(obj):
     return hasattr(obj, '__cuda_array_interface__')
 
 
+def is_fp16_supported():
+    """ Determine if IEEE floating-point 16-bit is supported.
+
+        Returns True if fp16 is supported, False otherwise.
```
gmarkall, line 91: As per comment above:

```suggestion
def is_float16_supported():
    """Whether 16-bit floats are supported.

    Returns True if float16 is supported, False otherwise.
```

numba/cuda/simulator/api.py
```diff
@@ -16,6 +17,10 @@ def select_device(dev=0):
     assert dev == 0, 'Only a single device supported by the simulator'
 
 
+def is_fp16_supported():
```
gmarkall, line 20:
```suggestion
def is_float16_supported():
```

numba/cuda/tests/cudadrv/test_is_fp16.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def tearDown(self):
+        super().tearDown()
+        config.CUDA_USE_NVIDIA_BINDING = self._use_cuda_bindings
+
+    @skip_on_cudasim
+    @skip_unless_cc_53
+    def test_is_fp16_supported(self):
+        self.assertEqual(config.CUDA_USE_NVIDIA_BINDING, 0)
+        self.assertFalse(cuda.is_fp16_supported())
```
gmarkall, line 25:
```suggestion
        self.assertFalse(cuda.is_float16_supported())
```

numba/cuda/tests/cudadrv/test_is_fp16.py
```diff
@@ -0,0 +1,31 @@
...
+        with override_config('CUDA_USE_NVIDIA_BINDING', 1):
+            self.assertTrue(cuda.is_fp16_supported())
```
gmarkall, line 27:
```suggestion
            self.assertTrue(cuda.is_float16_supported())
```

numba/cuda/tests/cudadrv/test_is_fp16.py
```diff
@@ -0,0 +1,31 @@
...
+
+    @skip_unless_cudasim
+    def test_is_fp16_supported_on_simulator(self):
+        self.assertTrue(cuda.is_fp16_supported())
```
gmarkall, line 31:
```suggestion
        self.assertTrue(cuda.is_float16_supported())
```

numba/cuda/tests/cudadrv/test_is_fp16.py
```diff
@@ -0,0 +1,31 @@
...
+        config.CUDA_USE_NVIDIA_BINDING = 0
+
+    def tearDown(self):
+        super().tearDown()
+        config.CUDA_USE_NVIDIA_BINDING = self._use_cuda_bindings
+
+    @skip_on_cudasim
+    @skip_unless_cc_53
+    def test_is_fp16_supported(self):
+        self.assertEqual(config.CUDA_USE_NVIDIA_BINDING, 0)
```
gmarkall, line 24: This seems to be testing the `setUp()` method and not the method under test.
```suggestion
```

gmarkall (changes requested):
Thanks for the PR. There are some comments on the diff.

In addition to the comments on the diff, I think the documentation needs to be clear that this represents whether Numba supports float16 in the current configuration. If the user needs to know about the capabilities of the device they're using, they themselves should check whether the compute capability is at least 5.3 (e.g. with `cuda.get_current_device().compute_capability >= (5, 3)`). The rationale for this (as discussed OOB previously) is to avoid duplicating the purpose of the compute capability in Numba by carrying around a set of checks for particular features of the hardware.

numba/cuda/api.py
```diff
@@ -85,6 +85,13 @@ def is_cuda_array(obj):
     return hasattr(obj, '__cuda_array_interface__')
 
 
+def is_fp16_supported(device_id=None):
```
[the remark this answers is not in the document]
testhound: > Thanks for the PR. There are some comments on the diff.
> 
> In addition to the comments on the diff, I think the documentation needs to be clear that this represents whether Numba supports float16 in the current configuration. If the user needs to know about the capabilities of the device they're using, they themselves should check whether the compute capability is at least 5.3 (e.g. with `cuda.get_current_device().compute_capability >= (5, 3)`). The rationale for this (as discussed OOB previously) is to avoid duplicating the purpose of the compute capability in Numba by carrying around a set of checks for particular features of the hardware.

@gmarkall the more I think about this I disagree. The attribute `cuda.get_current_device().compute_capability` is implemented in Numba, why not just add it to the `is_float16_supported method`? We are already "carrying around" this code and I am concerned developers may forget to add the additional check for the compute capability/

testhound:
gpuci run tests

testhound:
@gmarkall all your feedback incorporated.

numba/cuda/api.py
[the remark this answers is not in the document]
gmarkall: I could live with an additional property of a device that indicates its support for `float16`, e.g. so you could do:

```python
# Check whether Numba supports float16 in the current configuration
cuda.is_float16_supported()
# Check whether the current device supports float16
cuda.get_current_device().supports_float16
```

where `supports_float16` is a property that returns `self.compute_capability >= (5, 3)`. I'm not keen on trying to shoehorn the check of both the bindings and the device capability into one method, because if it returns `False` there are two possible reasons why it's not supported, and it won't be clear which of them are the case.

A developer would have to check both things. I think trying to give them a single function to call in an attempt to provide a simple interface actually creates a more complicated one because it promotes ambiguity.

numba/cuda/api.py
```diff
@@ -85,6 +85,14 @@ def is_cuda_array(obj):
     return hasattr(obj, '__cuda_array_interface__')
 
 
+def is_float16_supported():
+    """Whether 16-bit floats are  supported.
+
+        Returns True if float16 is supported, False otherwise.
```
gmarkall, line 91: The indentation is out-of-place here:

```suggestion
    Returns True if float16 is supported, False otherwise.
```

numba/cuda/api.py
```diff
@@ -85,6 +85,14 @@ def is_cuda_array(obj):
     return hasattr(obj, '__cuda_array_interface__')
 
 
+def is_float16_supported():
+    """Whether 16-bit floats are  supported.
```
gmarkall, line 89:
```suggestion
    """Whether 16-bit floats are supported.
```

gmarkall (changes requested):
Thanks for the updates - some additional comments and follow-up on the diff.

testhound:
gpuci run tests

testhound:
@gmarkall indentation and grammar fixes committed.

gmarkall:
@testhound What about https://github.com/numba/numba/pull/8634#discussion_r1108358436?

testhound:
> @testhound What about [#8634 (comment)](https://github.com/numba/numba/pull/8634#discussion_r1108358436)?

@gmarkall I did not forget; I prioritized the compiler-rt integration with llvmlite this week. I will address the missing test case next week.

gmarkall:
> > @testhound What about [#8634 (comment)](https://github.com/numba/numba/pull/8634#discussion_r1108358436)?
> 
> @gmarkall I did not forget; I prioritized the compiler-rt integration with llvmlite this week. I will address the missing test case next week.

Thanks, I just wasn't sure whether a review was expected at this point.

testhound:
gpuci run tests

testhound:
@gmarkall ready for review

gmarkall (changes requested):
Thanks for the updates. The code and tests look OK in general, but the documentation still needs updating to be clear that the `is_float16_supported()` method refers to Numba's support for `float16` in the current configuration, and the docs should also mention the now-added `supports_float16` property.

testhound:
gpuci run tests

testhound:
@gmarkall ready for review.

docs/source/cuda-reference/host.rst
```diff
@@ -95,6 +95,10 @@ the functionality of the selected device:
       Delete the context for the device. This will destroy all memory
       allocations, events, and streams created within the context.
 
+   .. attribute:: supports_float16
+
+      Return ``True`` if the device supports float16 operations, ``False`` otherwise.
```
gmarkall, line 100: This should have a line-wrap:
```suggestion
      Return ``True`` if the device supports float16 operations, ``False``
      otherwise.
```

gmarkall (changes requested):
Thanks for the update. There's a line wrap needed in one section of the docs, which is marked on the diff.

For the documentation in the Kernel API section, the order in which it's presented seems a bit mixed up. How about:

```diff
diff --git a/docs/source/cuda-reference/kernel.rst b/docs/source/cuda-reference/kernel.rst
index 68dca9b17..67abef2f6 100644
--- a/docs/source/cuda-reference/kernel.rst
+++ b/docs/source/cuda-reference/kernel.rst
@@ -480,16 +480,19 @@ precision parts of the CUDA Toolkit documentation.
 16-bit Floating Point Intrinsics
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-.. function:: numba.cuda.is_float16_supported ()
+The functions in the ``cuda.fp16`` module are used to operate on 16-bit
+floating point operands. These functions return a 16-bit floating point result.
 
-   Return ``True`` if 16-bit floats are supported, ``False`` otherwise.
+To determine whether Numba supports compiling code that uses the ``float16``
+type in the current configuration, use:
+
+   .. function:: numba.cuda.is_float16_supported ()
 
-The following functions are used to operate on 16-bit floating point operands.
-These functions return a 16-bit floating point result.
+   Return ``True`` if 16-bit floats are supported, ``False`` otherwise.
 
-Determining whether float16 is supported additionally requires ascertaining if the
-device in question supports float16 via
-:attr:`supports_float16 <numba.cuda.cudadrv.driver.Device.supports_float16>`.
+To check whether a device supports ``float16``, use its
+:attr:`supports_float16 <numba.cuda.cudadrv.driver.Device.supports_float16>`
+attribute.
 
 .. function:: numba.cuda.fp16.hfma (a, b, c)
 
```

?

testhound:
@gmarkall ready for review. New test failure relating to typeguard import error. This appears to be a CI issue.

gmarkall:
gpuci run tests

gmarkall (approved):
Thanks for the update. This looks good - just waiting on CI.

testhound:
gpuci run tests

testhound:
@gmarkall merged main into this branch and all tests now pass.

gmarkall:
I've tested this locally on both Linux and Windows, both in the Numba tree (`python runtests.py ...`) and outside of it (`python -m numba.runtests ...`) so I'm reasonably confident this should be OK on the buildfarm now.

@sklam / @stuartarchibald Could this have a buildfarm run please?

sklam:
BFID `numba_smoketest_cuda_yaml_194`

PR #8644, opened 2022-12-06
Fix bare reraise support

sklam:
- restore lost logic that we only ban reraise within a try-except block, but bare reraise (causes the interpreter to reraise existing exception value) is supported.
- fix C logic in handling reraise. C API changed and old code needs access to now opaque struct (`_PyErr_StackItem`). Use [PyErr_GetExcInfo](https://docs.python.org/3/c-api/exceptions.html#c.PyErr_GetExcInfo) instead.

### DIFF
--- a/numba/_helperlib.c
+++ b/numba/_helperlib.c
@@ -891,18 +891,19 @@ void traceback_add_loc(PyObject *loc) {
 static
 int reraise_exc_is_none(void) {
     /* Reraise */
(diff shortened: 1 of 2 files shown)


stuartarchibald (approved):
Thanks for the patch, the changes seem appropriate given what is being fixed. The two tests failing on Python 3.11 related to `reraise` now pass!

PR #8645, opened 2022-12-06
PR #8232 continued

stuartarchibald:
A continuation of #8232 (it's rebased on current `main` c910705) with view of debugging/fixing the issues cause by `fma` instructions present on certain NumPy versions compiled for M1 arch.


stuartarchibald:
Closing, investigation is complete. It's possible to numerically detect the probable presence of FMA instructions but not with the granularity required to be useful.

PR #8648, opened 2022-12-06
Add overload to handle int(s: str)

### LINKED ISSUE #5650: Unable to create integer from unicode string with int()

k-hanke:
Resolves #5650

### DIFF
--- a/numba/cpython/unicode.py
+++ b/numba/cpython/unicode.py
@@ -2399,6 +2399,26 @@ def ol_ord(c):
         return impl
 
 
+# near the speed of native python (including python overhead) when called from
+# jitted code; however, getting significantly faster would prohibit implementing
+# it in python like this
+@overload(int)
+def str_integer(s):
+    if isinstance(s, numba.types.UnicodeType):
+        def impl(s):
+            if s[0] == "-":
+                sgn = -1
+                s = s[1:]
+            else:
+                sgn = 1
+            final_index, result = len(s) - 1, 0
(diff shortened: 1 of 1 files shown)


stuartarchibald:
Thanks for the PR @k-hanke. As mentioned above, @dlee992 has already created a PR that is under review that aims to fully implement `int(string, base)`, I think this is a superset of the functionality proposed in this PR? As this enabling this feature is clearly of interest to you, perhaps a route forward might be to close this PR and assist in reviewing #8558? What do you think? Thanks again!

PR #8649, opened 2022-12-06
Remove `glue_lowering` aka `overload_glue`

apmasell:
Delete the implementation of `glue_lowering` now that all uses have been removed.

### DIFF
--- a/numba/core/overload_glue.py
+++ b/numba/core/overload_glue.py
@@ -253,17 +253,6 @@ class _Gluer:
                 self._registered[key] = wrapper
                 return wrapper
 
-    def defer_lowering(self, key, lower_fn):
-        """Defer lowering of the given key and lowering function.
-        """
-        with self._lock:
-            if key in self._no_defer:
-                # Key is marked as no defer, register lowering now
-                lower_fn()
-            else:
-                # Defer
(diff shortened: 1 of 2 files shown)


stuartarchibald (commented):
Thanks for the patch @apmasell. I'm not sure that the `numba.core.overload_glue` module contains anything that is used anywhere in the code base on the `main` branch as of 0441bb1, could the module perhaps be removed entirely? All the references I can see in the code base are unused or comments/tests about some aspect of the module, all of which become dead on removing the module. Do you see the same or have I missed something? Thanks for checking!

stuartarchibald:
Couple more comments/references to `overload_glue`:

Suggest changing this
https://github.com/numba/numba/blob/ded015a852329cab10b72738215a7ccbcbababe0/numba/tests/test_errorhandling.py#L281
to:
```
are mostly "@overload"s they'll just appear as "No match".
```


Suggest changing the name of this test:
https://github.com/numba/numba/blob/ded015a852329cab10b72738215a7ccbcbababe0/numba/tests/test_extending.py#L1966-L1978
to
```
def test_overload_arg_binding
```

stuartarchibald:
Note: PR title updated to reflect final content.

PR #8650, opened 2022-12-06
Split up parfor pass

sklam:
Split up the passes to allow custom compiler pipelines to change parfor behavior while reusing some existing behavior.

### DIFF
--- a/numba/core/compiler.py
+++ b/numba/core/compiler.py
@@ -26,7 +26,10 @@ from numba.core.untyped_passes import (ExtractByteCode, TranslateByteCode,
 
 from numba.core.typed_passes import (NopythonTypeInference, AnnotateTypes,
                                      NopythonRewrites, PreParforPass,
-                                     ParforPass, DumpParforDiagnostics,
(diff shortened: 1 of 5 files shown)


DrTodd13 (changes requested):
Please see some experiments I did in github.com/IntelLabs/numba.  Look at the compute follows data pull request.  One thing I think is mandatory here is the change to the test code that you can see in my work for the fusion pass (at least) and maybe the one after that to run so that the parfor counts will be correct in the tests.

In my code, I also added the parfor tag (called "lowerer").  You can choose to incorporate that here or hold off and make it a separate PR.

Also, in my PR I show more code movement for example the two fusion functions out of the conversion pass and into the fusion pass.  I don't see that here in this PR.

In general, our approaches were basically identical so I'm happy but just want to make sure the details match.

sklam:
@DrTodd13, i like your code changes more. I was just trying to do the minimal and see if everything is still ok.

DrTodd13 (commented):
Yes, this way is better.

PR #8655, opened 2022-12-08
Support `Optional(...)` in Typed List

dlee992:
Hi, I'm implementing this feature. Don't need to review right now.

I have already achieved some kinds of initial success for supporting this feature (see the testcase `test_optional_in_python(self)`, it has already worked, although still flawed). 

But I do have a question about `_gen_getitem`. Why do we need to wrap a not-none item type into an optional type?

https://github.com/numba/numba/blob/c91070592448ebc2243442677f5358988f5feb62/numba/typed/listobject.py#L689-L700

It does confuse me. From the git history, perhaps @stuartarchibald can give me some tips? Indeed, I want to remove this wrapper logic, since I'm trying to support optional item_type natively.

### DIFF
--- a/numba/core/types/containers.py
+++ b/numba/core/types/containers.py
@@ -618,9 +618,6 @@ class ListType(IterableType):
     def __init__(self, itemty):
         assert not isinstance(itemty, TypeRef)
         itemty = unliteral(itemty)
-        if isinstance(itemty, Optional):
-            fmt = "List.item_type cannot be of type {}"
-            raise TypingError(fmt.format(itemty))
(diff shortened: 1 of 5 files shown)


sklam:
> But I do have a question about `_gen_getitem`. Why do we need to wrap a not-none item type into an optional type?

Optional type is used here as an equivalent to returning nothing (e.g. like return NULL in C).  It can be restructured to make it less confusing when the element type is already an optional type.

dlee992:
Sry, I haven't focused on this feature for more than half a year, thus I will close this. If anyone is interested, maybe you can get some hints from this unfinished draft.

PR #8661, opened 2022-12-09
Make external compiler discovery lazy in the test suite.

stuartarchibald:
This aims to make it so that external compilers are only probed when absolutely necessary and that the result of the probe is cached. This should speed up initialisation of the testing framework and consequently speed up tests running in subprocesses.

### DIFF
--- a/numba/tests/test_pycc.py
+++ b/numba/tests/test_pycc.py
@@ -261,7 +261,6 @@ class TestCC(BasePYCCTest):
 
 class TestDistutilsSupport(TestCase):
 
-
     def setUp(self):
         super().setUp()
         self.skip_if_no_external_compiler() # external compiler needed
--- a/numba/tests/test_pycc.py
+++ b/numba/tests/test_pycc.py
@@ -15,10 +15,10 @@ import llvmlite.binding as ll
 from numba.core import utils
 from numba.pycc.decorators import clear_export_registry
 from numba.pycc.platform import find_shared_ending, find_pyext_ending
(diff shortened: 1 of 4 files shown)


numba/tests/test_pycc.py
```diff
@@ -259,10 +259,13 @@ def test_c_extension_usecase(self):
             self.assertPreciseEqual(got, expect)
 
 
-@needs_external_compilers
 class TestDistutilsSupport(TestCase):
 
+
     def setUp(self):
+        super().setUp()
+        self.skip_if_no_external_compiler() # external compiler needed
```
gmarkall, line 267: I just checked whether it's OK to call `skipTest()` in a `setUp()` method, and it is: https://docs.python.org/3/library/unittest.html#skipping-tests-and-expected-failures

numba/tests/test_pycc.py
```diff
@@ -259,10 +259,13 @@ def test_c_extension_usecase(self):
             self.assertPreciseEqual(got, expect)
 
 
-@needs_external_compilers
 class TestDistutilsSupport(TestCase):
 
+
```
sklam, line 264:
```suggestion
```
empty line

sklam (commented):
Manually verified that the `external_compiler_works` is now lazy. 
There's just one nitpick regarding an empty line added.

numba/tests/test_pycc.py
stuartarchibald: Thanks, fixed in 8e43a67.

sklam (approved):

PR #8662, opened 2022-12-10
Add support for `.nbytes` accessor for numpy arrays

alanhdu:
The `arr.nbytes` accessor is the same as `arr.size * arr.itemsize`, so it's fairly easy to add in. I haven't figured out how to actually build Numba locally yet, so I only tested this by manually editing an installed conda build.


stuartarchibald (commented):
Thanks for the patch. The implementation looks good. Please could you update the docs here to declare this attribute as supported:
https://github.com/numba/numba/blob/61e4b01a0e317abdfb5193347d820547f8cc0acd/docs/source/reference/numpysupported.rst?plain=1#L237-L238
after which this can be merged. Thanks again!

alanhdu:
@stuartarchibald Done!

stuartarchibald (approved):

PR #8666, opened 2022-12-12
Updates for Python 3.8 baseline/Python 3.11 migration

stuartarchibald:
This series of patches contains two things
1. Clean-up following removal of Python 3.7/migration of the baseline Python version to 3.8
2. Clean-up with view of minimising the diff against Python 3.11 work, as located on the `py3.11` branch.

Patches do the following:
* Fix `PyTypeObject` structs/initializers with respect to the now supported Python 3.8, 3.9 and 3.10.
* Remove the custom code object manipulation code in favour of `CodeType.replace()` calls (already in the `py3.11` branch).
* Fixes a bug in the loop bound of a test in `test_ir.py` (already in the `py3.11` branch)
* Marks `test_extended_arg.py` as `flake8` compliant (already in the `py3.11` branch).

### DIFF
--- a/numba/_devicearray.cpp
+++ b/numba/_devicearray.cpp
@@ -41,10 +41,10 @@ PyTypeObject DeviceArrayType = {
     sizeof(DeviceArray),                         /* tp_basicsize */
     0,                                           /* tp_itemsize */
     0,                                           /* tp_dealloc */
-    0,                                           /* tp_print */
+    0,                                           /* tp_vectorcall_offset */
     0,                                           /* tp_getattr */
     0,                                           /* tp_setattr */
-    0,                                           /* tp_compare */
+    0,                                           /* tp_as_async */
     0,                                           /* tp_repr */
     0,                                           /* tp_as_number */
     0,                                           /* tp_as_sequence */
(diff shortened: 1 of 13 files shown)


numba/_devicearray.cpp
```diff
@@ -85,10 +85,31 @@ PyTypeObject DeviceArrayType = {
     0,                                           /* tp_del */
     0,                                           /* tp_version_tag */
     0,                                           /* tp_finalize */
-#if PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION == 8
+/* The docs suggest Python 3.8 has no tp_vectorcall
+ * https://github.com/python/cpython/blob/d917cfe4051d45b2b755c726c096ecfcc4869ceb/Doc/c-api/typeobj.rst?plain=1#L146
+ * but the header has it:
```
sklam, line 90: 😮

Given how crazy it is already, would it actually worth moving to the [PyType_FromSpec](https://pythoncapi.readthedocs.io/type_object.html#pytype-fromspec) API so we never have to change this again?

sklam (commented):
One comment. everything else looks good

numba/_devicearray.cpp
stuartarchibald: I think that this could potentially work and would certainly have the noted advantage. I've done some preliminary work in this area and the code changes are unfortunately much larger than the scope of this PR, so have deferred this to #8669.

stuartarchibald:
> One comment. everything else looks good

Thanks for the review. I've replied in-line.

sklam (approved):

PR #8544, opened 2022-10-26
Remove reliance on npy_<impl> ufunc loops.

stuartarchibald:
This removes Numba's reliance on NumPy ufunc inner loops. All the cases where a ufunc inner loop (or something from `numpy/npy_math.h`) was present, have been replaced as follows:

* `np.signbit` -> Numba custom builder implementation.
* `np.nextafter` -> Numba helperlib wrapped C library call.
* `np.spacing` -> Numba custom builder implementation.
* `np.log2` -> llvm intrinsic.
* `np.exp2` -> llvm intrinsic.
* `np.logaddexp` -> custom Numba overload translation of NumPy C code.
* `np.logaddexp2` -> custom Numba overload translation of NumPy C code.
* Numba helperlib `numba_cpowf` -> now uses `<complex.h>` functions in place of NumPy functions.

Edit: This also:
* Removes static linkage to `libnpymath.a` and replaces use of `numpy.distutils.misc_util`  with `np.get_include()` as there's now no for link information.

PR #8671, opened 2022-12-15
Debugging #8544

stuartarchibald:
As title. Do not merge.


numba/pycc/platform.py
```diff
@@ -58,10 +58,10 @@ def _check_external_compiler():
                 ntf.close()
                 # *output_dir* is set to avoid the compiler putting temp files
                 # in the current directory.
-                compiler.compile([ntf.name], output_dir=Path(ntf.name).anchor)
+                #compiler.compile([ntf.name], output_dir=Path(ntf.name).anchor)
```
stuartarchibald, line 61: Potential problem line.

stuartarchibald:
ebc9299 has no `temp.{c, cxx}` in the log, which suggests this line https://github.com/numba/numba/pull/8671#discussion_r1049452169 is potentially the cause.

stuartarchibald:
e221307 has no `temp.{c, cxx}` in the log but does have this line https://github.com/numba/numba/pull/8671#discussion_r1049452169 present. This perhaps suggests it's NumPy's distutils patching standard distutils that "fixes" the problem. Seems like this could be due to the replacement of `.spawn` in the `CCompiler` https://github.com/numpy/numpy/blob/62e075c1c22530528bf5b6778ef51e212f65acb7/numpy/distutils/ccompiler.py#L112-L183.

The `stdlib` vendored `distutils` variant of `.spawn` uses a `Popen` with no specification for `stderr` handling (the new process will inherit the file descriptors from the parent): https://github.com/python/cpython/blob/3939a4b7d95a403ee587ff5d222db53ec83bec2c/Lib/distutils/spawn.py#L79 whereas the NumPy patched version of `.spawn` has redirection to `stdout` in a `checked_output` call https://github.com/numpy/numpy/blob/62e075c1c22530528bf5b6778ef51e212f65acb7/numpy/distutils/ccompiler.py#L145. i.e. any output to `stdout/stderr` from the subprocess in the NumPy patch version will be "hidden".

stuartarchibald:
Closing, investigation is complete.

PR #8673, opened 2022-12-16
Enable the CUDA simulator tests on Windows builds in Azure CI.

stuartarchibald:
As title.

### DIFF
--- a/buildscripts/azure/azure-windows.yml
+++ b/buildscripts/azure/azure-windows.yml
@@ -51,6 +51,7 @@ jobs:
     - script: |
         call activate %CONDA_ENV%
         set NUMBA_CAPTURED_ERRORS=new_style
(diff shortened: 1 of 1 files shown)


esc (dismissed):
I have checked that the setting of the environment variable does indeed have the intended effect. I have checked the output of the test-listing-task and of the shard-running-task and have detected sufficient indication that cuda tests would run via the simulator. Please revert the `-v` switch and this is ready to merge.

stuartarchibald:
> I have checked that the setting of the environment variable does indeed have the intended effect. I have checked the output of the test-listing-task and of the shard-running-task and have detected sufficient indication that cuda tests would run via the simulator. Please revert the `-v` switch and this is ready to merge.

Thanks for the review and for checking @esc, 75b8328 is the revert of 81cd582 to remove the verbose flag.

esc (approved):

PR #8674, opened 2022-12-16
Enforce that no warnings are raised on listing of tests.

stuartarchibald:
As title. This is minor clean up to ensure that there are no warnings raised when the tests are listed e.g. via `./runtests.py -l`.

### DIFF
--- a/numba/tests/test_serialize.py
+++ b/numba/tests/test_serialize.py
@@ -12,10 +12,17 @@ from numba.core.errors import TypingError
 from numba.tests.support import TestCase
 from numba.core.target_extension import resolve_dispatcher_from_str
 from numba.cloudpickle import dumps, loads
-from .serialize_usecases import *
 
 
 class TestDispatcherPickling(TestCase):
+
+    def setUp(self):
+        super().setUp()
+        # delay import of these so as to not trigger e.g. deprecation warnings
+        # present on Numba decorator APIs on import.
+        import numba.tests.serialize_usecases as ucs
+        g = globals()
+        g.update({x: getattr(ucs, x) for x in ucs.__all__})
 
     def run_with_protocols(self, meth, *args, **kwargs):
         for proto in range(pickle.HIGHEST_PROTOCOL + 1):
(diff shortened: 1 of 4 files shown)


stuartarchibald:
Depends on #8544 which removes the use of `numpy.distutils` that currently raises a deprecation warning (at least on NumPy 1.23).

numba/tests/test_serialize.py
```diff
@@ -12,11 +12,18 @@
...
 
 class TestDispatcherPickling(TestCase):
 
+    def setUp(self):
+        super().setUp()
+        # delay import of these so as to not trigger e.g. deprecation warnings
+        # present on Numba decorator APIs on import.
+        import numba.tests.serialize_usecases as ucs
+        g = globals()
+        g.update({x: getattr(ucs, x) for x in ucs.__all__})
```
sklam, line 25: shouldn't it be a `setUpClass`? or is that too early? Every test method will trigger a global update.

and wouldn't it be better to turn all the references to `serialize_usecase.xyz` instead of hacking the globals.

stuartarchibald: I wasn't sure if the original import form had a specific intent (it's very unusual to see `from .<module> import *` in the code base) and given this is testing serialisation thought it best to attempt to emulate that behaviour. If it's not important, I think it should be ok to put the `serialize_usecases` module as a reference into the test class itself and then refer to it as `self._serialize_usecases.xyz` as noted. What do you think?

stuartarchibald:
Close/Reopen to get a build against `main` with #8544 merged.

esc:
@stuartarchibald wanna revive this one, seems like it got quite far already.

stuartarchibald:
> @stuartarchibald wanna revive this one, seems like it got quite far already.

Am waiting for a response to https://github.com/numba/numba/pull/8674#discussion_r1071438001.

stuartarchibald:
@esc I don't think that this is a critical patch for any particular release, it's more of a nice-to-have, but I'd still like to get it done sometime.

PR #8675, opened 2022-12-16
Make `always_run` test decorator a tag and improve shard tests.

### LINKED ISSUE #8657: Check test shards do not accidentally overlap
From: https://github.com/numba/numba/pull/8550#discussion_r1043636415_
     
To guard against future mistakes, there should be a check that shards do not overlap anything that is not marked as [always_test](https://github.com/numba/numba/blob/9d4f82df7f0426d89b189eafc7acdc6bf719c716/numba/testing/__init__.py#L64)

stuartarchibald:
This patch:

* Makes an `always_run` test decorator to use on tests that should run in all test shards, it's based on the existing test `tag` mechanism.
* Updates the test runner to recognise the `always_run` tag when sharding so as to ensure that tests tagged with this tag will always be included in a shard.
* Updates the unit tests to ensure fundamental properties with regards to sharding. This includes shard uniqueness, the presence of the `always_run` tests in shards and also that the sum of the test shards and `always_run` tests is the same as the full test listing.
* Fixes up a few references slice->shard.

Fixes #8657

### DIFF
--- a/numba/tests/test_runtests.py
+++ b/numba/tests/test_runtests.py
@@ -159,8 +159,8 @@ class TestCase(unittest.TestCase):
 
         # check that each shard contains no repeats
         sharded_sets = [set(x) for x in sharded]
-        [self.assertEqual(len(sharded_sets[i]), len(sharded[i]))
-         for i in range(len(sharded))]
+        for i in range(len(sharded)):
+            self.assertEqual(len(sharded_sets[i]), len(sharded[i]))
 
         # check that the always running tests are in every shard, and then
         # remove them from the shards
@@ -178,7 +178,8 @@ class TestCase(unittest.TestCase):
         # same as the full listing
 
         sum_of_parts = set()
-        [sum_of_parts.update(x) for x in sharded_sets]
+        for x in sharded_sets:
+            sum_of_parts.update(x)
         sum_of_parts.update(always_running)
 
         full_listing = set(self._get_numba_tests_from_listing(
--- a/numba/tests/test_runtests.py
+++ b/numba/tests/test_runtests.py
@@ -4,6 +4,7 @@ import subprocess
 
 from numba import cuda
 import unittest
+import itertools
 
 try:
     import git  # noqa: F401 from gitpython package
@@ -64,6 +65,11 @@ class TestCase(unittest.TestCase):
         self.assertTrue(any('numba.tests.npyufunc.test_' in line
                             for line in lines),)
 
+    def _get_numba_tests_from_listing(self, listing):
+        """returns a filter on strings starting with 'numba.', useful for
+        selecting the 'numba' test names from a test listing."""
+        return filter(lambda x: x.startswith('numba.'), listing)
+
     def test_default(self):
         self.check_all([])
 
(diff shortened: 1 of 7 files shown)


numba/tests/test_runtests.py
```diff
@@ -119,34 +125,70 @@ def get_count(arg_list):
...
-        sliced = set()
+
+        # Check the tests that should always run are present
+        tmp = self.get_testsuite_listing(['--tag', 'always_test'])
+        always_running = set(self._get_numba_tests_from_listing(tmp))
+        self.assertGreaterEqual(len(always_running), 1)
+
+        for test in always_running:
+            self.assertIn(test, l1)
+            self.assertIn(test, l2)
```
stuartarchibald, line 148: This is covered by the test below, is it needed?

numba/tests/test_runtests.py
```diff
@@ -119,34 +125,66 @@ def get_count(arg_list):
...
+        # get some shards
+        sharded = list()
         for i in range(3):
             subset = self.get_testsuite_listing(['-j', '{}:3'.format(i)])
-            sliced.update(filter_test(subset))
-        # The tests must be equivalent
-        self.assertEqual(full, sliced)
+            slist = [*self._get_numba_tests_from_listing(subset)]
+            # there's at least 100 tests in a shard of roughly 1/3!
+            self.assertGreater(len(slist), 100)
```
apmasell, line 150: This seems fragile to put in this value.

numba/tests/test_runtests.py
```diff
@@ -119,34 +125,66 @@ def get_count(arg_list):
...
+
+        # check that there is no overlap between the shards
+        for a, b in itertools.combinations(sharded_sets, 2):
+            self.assertFalse(a & b)
+
+        # check that the sum of the shards and the always running tests is the
+        # same as the full listing
+
+        sum_of_parts = set()
+        [sum_of_parts.update(x) for x in sharded_sets]
```
apmasell, line 181: Why is this a list comprehension?

numba/tests/test_runtests.py
stuartarchibald: The intent here is to make sure that there's not an "accident" whereby the sharding logic has produced an unreasonable quantity of tests in a shard given the size of the test suite. Having seen a few mistakes made through the history of the test listing logic, they tend to manifest as there being be zero or just a few tests listed. As a result I chose 100 as a minimum based on this and as there are 10k+ tests in total it's unlikely a ~1/3 split would ever be less than 100. Also note there are other baked in constants (see `check_all` above which makes sure there are more than 5k tests listed).

Do you think there might be something less "fragile" to check or some other way to assert similar? I suppose querying the size of the test suite and taking a percentage of 1/3 of that might be another way?

stuartarchibald:
@apmasell thanks for the review. Note that since review:
* 3772028 merges `main` and fixes conflicts.
* 228112a fixes a comment `slice`->`shard`.
* 47be4be changes list comp to loops as per review comment.

numba/tests/test_runtests.py
apmasell: Take the total and divide by 3. Then it doesn't matter if the test count changes. Or check that the slices are approximately the same size. If they sum to the total and are about equal, then the actual number doesn't matter.

stuartarchibald: Maybe it's best to just remove this assertion. In this module:
* `test_check_shard` asserts that shards are "approximately the same size".
* `test_default`/`test_all` both check that the test suite is being listed as a sensible size (>5000 tests).
* this test checks that the shard sizes add up to the total which is the same as obtained in `test_default`.

I think this renders the assertion under discussion well intentioned but superfluous.

stuartarchibald: The above is done in b4081c1.

apmasell (approved):

sklam (approved):

PR #8677, opened 2022-12-16
Add support for min and max on boolean types.

DrTodd13:
Veros ocean simulator through Ramba generates min on booleans so adding support here.

### DIFF
--- a/numba/cpython/builtins.py
+++ b/numba/cpython/builtins.py
@@ -573,7 +573,11 @@ def indval_min(indval1, indval2):
                     return indval2
             return indval1
         return min_impl
-    elif isinstance(indval1, types.scalars.Boolean) and \
+
+
+@overload(min)
+def indval_min(indval1, indval2):
+    if isinstance(indval1, types.scalars.Boolean) and \
          isinstance(indval2, types.scalars.Boolean):
         def bool_min_impl(indval1, indval2):
             return indval1 and indval2
@@ -607,7 +611,11 @@ def indval_max(indval1, indval2):
                     return indval2
             return indval1
         return max_impl
-    elif isinstance(indval1, types.scalars.Boolean) and \
+
+
+@overload(max)
+def indval_max(indval1, indval2):
+    if isinstance(indval1, types.scalars.Boolean) and \
          isinstance(indval2, types.scalars.Boolean):
         def bool_max_impl(indval1, indval2):
             return indval1 or indval2
--- a/numba/cpython/builtins.py
+++ b/numba/cpython/builtins.py
@@ -577,8 +577,8 @@ def indval_min(indval1, indval2):
 
 @overload(min)
 def boolval_min(val1, val2):
-    if isinstance(val1, types.scalars.Boolean) and \
-       isinstance(val2, types.scalars.Boolean):
+    if isinstance(val1, types.Boolean) and \
+       isinstance(val2, types.Boolean):
         def bool_min_impl(val1, val2):
             return val1 and val2
         return bool_min_impl
--- a/numba/cpython/builtins.py
+++ b/numba/cpython/builtins.py
@@ -615,8 +615,8 @@ def indval_max(indval1, indval2):
 
 @overload(max)
 def boolval_max(val1, val2):
-    if isinstance(val1, types.scalars.Boolean) and \
-       isinstance(val2, types.scalars.Boolean):
+    if isinstance(val1, types.Boolean) and \
+       isinstance(val2, types.Boolean):
         def bool_max_impl(val1, val2):
             return val1 or val2
         return bool_max_impl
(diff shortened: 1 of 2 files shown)


gmarkall:
Just had a quick look at this - doesn't it seem a little odd to squeeze the implementation for bools into the overloads for min and max on `IndexValue` types?

DrTodd13:
> Just had a quick look at this - doesn't it seem a little odd to squeeze the implementation for bools into the overloads for min and max on `IndexValue` types?

I don't normally add these kind of overloads.  What is the standard approach in this case?  Should I just add a whole new function instead for each separate type?

gmarkall:
> I don't normally add these kind of overloads. What is the standard approach in this case? Should I just add a whole new function instead for each separate type?

Yes - adding a new overload function for each separate type fits in well with the standard pattern for overloads (in my view / experience).

numba/cpython/builtins.py
```diff
@@ -573,7 +573,11 @@ def min_impl(indval1, indval2):
                     return indval2
             return indval1
         return min_impl
-    elif isinstance(indval1, types.scalars.Boolean) and \
+
+
+@overload(min)
+def indval_min(indval1, indval2):
```
gmarkall, line 579: I'd suggest naming the function and variables to reflect the fact they are for booleans, and not `IndexValue` pairs, e.g.
```suggestion
def bool_min(bool1, bool2):
```

etc.

gmarkall (dismissed):
Many thanks for the fix! This looks good to me, but maybe someone else with a little more overload experience should give it a quick check over too.

Side note: this does highlight how it's a bit weird that our boolean values are not integers, unlike Python - e.g.:

```
In [2]: bool.mro()
Out[2]: [bool, int, object]
```

numba/tests/test_builtins.py
```diff
@@ -838,6 +838,27 @@ def test_min_1_invalid_types_npm(self):
...
+
+        x_operands = [False, True]
+        y_operands = [False, True]
+        for x, y in itertools.product(x_operands, y_operands):
+            self.assertPreciseEqual(cfunc(x, y), pyfunc(x, y))
+
+    def test_max_bool1(self, flags=enable_pyobj_flags):
+        """
+        max(*args)
+        """
```
stuartarchibald, line 853:
```suggestion
        # tests max(<booleans>)
```
suggest this, else the doc strings end up in the test log which adds noise.

numba/tests/test_builtins.py
```diff
@@ -838,6 +838,27 @@ def test_min_1_invalid_types_npm(self):
...
+        self.check_minmax_bool1(max_usecase1, flags)
+
+    def test_min_bool1(self, flags=enable_pyobj_flags):
+        """
+        min(*args)
+        """
```
stuartarchibald, line 859:
```suggestion
        # tests min(<booleans>)
```
as above.

numba/tests/test_builtins.py
```diff
@@ -838,6 +838,27 @@ def test_min_1_invalid_types_npm(self):
         with self.assertTypingError():
             self.check_min_max_invalid_types(min_usecase1, flags=no_pyobj_flags)
 
+    def check_minmax_bool1(self, pyfunc, flags):
+        cr = compile_isolated(pyfunc, (types.bool_, types.bool_), flags=flags)
+        cfunc = cr.entry_point
+
+        x_operands = [False, True]
+        y_operands = [False, True]
+        for x, y in itertools.product(x_operands, y_operands):
```
stuartarchibald, line 847:
```suggestion
        operands = (False, True)
        for x, y in itertools.product(operands, operands):
```
`x_operands` and `y_operands` are the same, only need one definition?

numba/cpython/builtins.py
```diff
@@ -575,6 +575,15 @@ def min_impl(indval1, indval2):
         return min_impl
 
 
+@overload(min)
+def boolval_min(val1, val2):
+    if isinstance(val1, types.scalars.Boolean) and \
+       isinstance(val2, types.scalars.Boolean):
```
stuartarchibald, line 581:
```suggestion
    if isinstance(val1, types.Boolean) and \
       isinstance(val2, types.Boolean):
```
suggest using the top level re-export for consistency.

numba/cpython/builtins.py
```diff
@@ -604,6 +613,15 @@ def max_impl(indval1, indval2):
         return max_impl
 
 
+@overload(max)
+def boolval_max(val1, val2):
+    if isinstance(val1, types.scalars.Boolean) and \
+       isinstance(val2, types.scalars.Boolean):
```
stuartarchibald, line 619:
```suggestion
    if isinstance(val1, types.Boolean) and \
       isinstance(val2, types.Boolean):
```
suggest using the top level re-export for consistency.

stuartarchibald (commented):
Thanks for the patch @DrTodd13, a few minor things to resolve else looks good.

stuartarchibald (approved):
Thanks for the patch!

PR #8680, opened 2022-12-20
Adjust flake8 config to be compatible with flake8=6.0.0

thomasjpfan:
This PR adjusts the comments in the flake8 configuration to be compatible with `flake8=6.0.0`. Currently, running `flake8=6.0.0` will result in the following error:

```python
  File "../lib/python3.10/site-packages/flake8/options/config.py", line 131, in parse_config
    raise ValueError(
ValueError: Error code '#' supplied to 'ignore' option does not match '^[A-Z]{1,3}[0-9]{0,3}$'
```

According to the [flake8 docs](https://flake8.pycqa.org/en/latest/user/configuration.html) inline comments are not supported and the recommend way to comment is to have the comments be in it's own line. `flake8=6.0.0` introduced a stricter check to enforce this requirement.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,12 +1,19 @@
 [flake8]
 ignore =
-    E20,   # Extra space in brackets
-    E231,E241,  # Multiple spaces around ","
-    E26,   # Comments
-    E731,  # Assigning lambda expression
-    E741,  # Ambiguous variable names
-    W503,  # line break before binary operator
-    W504,  # line break after binary operator
+    # Extra space in brackets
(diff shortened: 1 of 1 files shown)


gmarkall:
Many thanks for the PR! As the Numba maintainers are all out until the new year, this will be triaged during the next triage meeting on Jan 10th.

stuartarchibald (approved):
Thanks for making this update, I've manually tested this against `flake8=6.0.0` and it fixes the problem described.

PR #8685, opened 2022-12-28
Implement `__repr__` for numba types

luk-f-a:
Implements a custom `__repr__` for certain types to enable reconstruction of the type object from the repr string. From python docs: repr returns "a string that would yield an object with the same value when passed to `eval()`"

Note: some error messages were explicitly using repr (`%r`) for displaying type information. I assumed that this was a mistake and `str` (`%s`) is the desired behaviour. An alternative is to restore displaying repr() instead of str() and to adjust those tests that assert against the content of the error message.
Open question: should all error messages relating to type information be changed to `%s`? There are 100+ uses of `%r`, although not all of them related to type information.


guilhermeleobas:
Hi @luk-f-a, I don't think the tests are being executed on CI. If I try to run them locally, I get the following:

```
$ ./runtests.py -v -f numba.tests.python_typing.test_repr
python_typing (unittest.loader._FailedTest) ... ERROR

ERROR: python_typing (unittest.loader._FailedTest)
ImportError: Failed to import test module: python_typing
  File "/Users/guilhermeleobas/miniconda3/envs/numba/lib/python3.9/unittest/loader.py", line 154, in loadTestsFromName
ModuleNotFoundError: No module named 'numba.tests.python_typing'


Ran 1 test in 0.000s

FAILED (errors=1)
```

luk-f-a:
> Hi @luk-f-a, I don't think the tests are being executed on CI. If I try to run them locally, I get the following:

you're right, but it's weird. I have fixed the test that was failing, and I have added a new test that should always fail. I should see the first test pass and the second fail, otherwise something is still wrong.

guilhermeleobas:
@luk-f-a, you removed the test file in your last commit and forget to include the new one 😅

luk-f-a:
> @luk-f-a, you removed the test file in your last commit and forgot to include the new one sweat_smile

lol, yeah, I forgot to `add` in git, I shouldn't fix thing so late at night.

luk-f-a:
test failed as expected, due to the aded "always fail" test. I'm removing it in the next commit.

luk-f-a:
now all tests pass

numba/tests/test_repr.py
```diff
@@ -0,0 +1,53 @@
...
+    types.List, types.Tuple, types.DictType, types.ListType, types.Set,
+    types.bool_
+]
+
+
+class TestRepr(TestCase):
+    def check_repr(self, val):
+        ty = typeof(val)
+        tys_ns = {ty.__name__: ty for ty in NB_TYPES if hasattr(ty, "__name__")}
+        tys_ns.update({ty.name: ty for ty in NB_TYPES if hasattr(ty, "name")})
```
guilhermeleobas, line 21: This is invariant to `check_repr` and can be moved to `setUp`:

```python
class TestRepr(TestCase):
    def setUp(self) -> None:
        tys_ns = {ty.__name__: ty for ty in NB_TYPES if hasattr(ty, "__name__")}
        tys_ns.update({ty.name: ty for ty in NB_TYPES if hasattr(ty, "name")})
        self.tys_ns = tys_ns
        return super().setUp()

    def check_repr(self, val):
        ty = typeof(val)
        ty2 = eval(repr(ty), self.tys_ns)
        self.assertEqual(ty, ty2)
```

guilhermeleobas (changes requested):
Thanks again for working on this, @luk-f-a. Code looks good. I have two more comments, to which one of them is about testing more types.

numba/tests/test_repr.py
luk-f-a: good idea, I've added it now.

numba/tests/test_repr.py
```diff
@@ -0,0 +1,53 @@
+import unittest
+import numpy as np
+
+from numba.tests.support import TestCase
+from numba import typeof
+from numba.core import types
+from numba.typed import List, Dict
+
+NB_TYPES = [
+    types.Array, types.NestedArray, types.int64, types.float64,
```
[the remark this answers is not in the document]
luk-f-a: I have added the types in `number_domain`. The single types seem mostly fine without modification, but it was the containers and compound types that I had to work on. I have tested so far: typed list, reflected list, unituple, tuple, set, typed dict, record, array and nested array. Can you think of any other container I might be missing?

luk-f-a:
@guilhermeleobas I've applied the changes and I think it's "ready for review" again.

guilhermeleobas (approved):
LGTM. Thanks @luk-f-a

PR #8688, opened 2023-01-02
Implement np.nanargmin/nanargmax

Tobi995:
#4074 

Implements np.nanargmin/nanargmax similar to the np.argmin/argmax imeplementations

### DIFF
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -658,6 +658,8 @@ def array_argmin_impl_generic(arry):
 
 @register_jitable
 def array_nanargmin_impl_float(arry):
+    if numpy_version < (1, 22) and arry.size == 0:
+        raise ValueError("attempt to get argmin of an empty sequence")
     has_non_nan_index = False
     for v in arry.flat:
         if np.isnan(v):
@@ -766,6 +768,8 @@ def array_argmax_impl_float(arry):
 
 @register_jitable
 def array_nanargmax_impl_float(arry):
+    if numpy_version < (1, 22) and arry.size == 0:
+        raise ValueError("attempt to get argmax of an empty sequence")
     has_non_nan_index = False
     for v in arry.flat:
         if not np.isnan(v):
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -658,8 +658,6 @@ def array_argmin_impl_generic(arry):
 
 @register_jitable
 def array_nanargmin_impl_float(arry):
-    if arry.size == 0:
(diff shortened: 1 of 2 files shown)


Tobi995:
The error message raised by numpy changed between the numpy versions. How would you handle that?

gmarkall:
> The error message raised by numpy changed between the numpy versions. How would you handle that?

By making the test behaviour conditional on the NumPy version. You can get it from:

```python
from numba.np.numpy_support import numpy_version
```

and then check it like

```python
if numpy_version < (1, 20):
    expected = # ... something
else:
    expected = # ... something else
```

(see e.g. `test_np_functions.py`).

Tobi995:
Thanks, the tests are green now

numba/np/arraymath.py
```diff
@@ -656,6 +656,33 @@ def array_argmin_impl_generic(arry):
     return min_idx
 
 
+@register_jitable
+def array_nanargmin_impl_float(arry):
```
guilhermeleobas, line 660: The code here and in `array_nanargmax_impl_float` are practically the same. You can refactor it to something like:

```python
def build_array_nanarg_impl_float(name, operator):
    @register_jitable
    def array_nanargmax_impl_float(arry):
        if numpy_version < (1, 22) and arry.size == 0:
            if name == 'argmax':
                raise ValueError("attempt to get argmax of an empty sequence")
            else:
                raise ValueError("attempt to get argmin of an empty sequence")
        has_non_nan_index = False
        for v in arry.flat:
            if not np.isnan(v):
                max_value = v
                max_idx = 0
                has_non_nan_index = True
                break
        if not has_non_nan_index:
            raise ValueError("All-NaN slice encountered")

        idx = 0
        for v in arry.flat:
            if np.isnan(v):
                continue
            # if v > max_value:
            if operator(v, max_value):
                max_value = v
                max_idx = idx
            idx += 1
        return max_idx
    return array_nanargmax_impl_float
```

And inside the `@overload` function, do:

```python
    ...
    fn = build_array_nanarg_impl_float('argmax', operator.gt)  # or operator.lt for nanargmin
    if is_nonelike(axis):
        def array_argmax_impl(arr, axis=None):
            return fn(arr)
            # return array_nanargmax_impl_float(arr)
    else:
        array_argmax_impl = build_argmax_or_argmin_with_axis_impl(
            arr, axis, fn
        )
```

guilhermeleobas (changes requested):
Hi @Tobi995, thanks for the contribution. I've made a few comments regarding code duplication but it should be easy to handle.

PR #8620, opened 2022-11-24
[WIP] NumPy 1.24

gmarkall:
Current status: All is well with NumPy 1.24, so this needs some tidying up and other accompanying changes to be ready for review. To-do list:

- [ ] Update docs
- [ ] Reconfigure CI appropriately
- [ ] Rewrite history for coherency
- [ ] Create new PR for review, linking to this for history / edits

**Original, outdated description follows:**

For testing against NumPy 1.24.0rc1. Presently all seems to be good (all tests passing with NumPy 1.24) so it is not expected that any major changes are needed, assuming there are no further NumPy changes requiring attention.

I will keep this up to date with `main` and update to use newer RCs when they appear - when the final release is available, I will make this Ready For Review in order to support NumPy 1.24.

The use of `resolve_dtypes` will be added as a separate PR, because it is not needed to make all existing functionality in Numba work with 1.24 - its use will eventually simplify Numba' ufunc typing logic.

PR #8690, opened 2023-01-03
[CI only] Testing #8620 with NumPy versions < 1.24

gmarkall:
Only opened to try the code changes in #8620 with the original test matrix.


gmarkall:
Now that #8691 is ready to go except for CI on that branch itself, I think this is no longer needed.

PR #8701, opened 2023-01-10
Relaxed ULP testing precision for NumPy Generator tests across all systems

kc611:
As titled,

This is a temporary fix for ULP issues in NumPy's Generator support and will remain in effect till the issue is resolved.

### DIFF
--- a/docs/source/reference/numpysupported.rst
+++ b/docs/source/reference/numpysupported.rst
@@ -668,11 +668,12 @@ The following :py:class:`Generator` methods are supported:
 
 .. note::
   Due to instruction selection differences across compilers, there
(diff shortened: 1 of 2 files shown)


stuartarchibald (commented):
Thanks for the patch @kc611. Couple of minor things to resolve else looks good.

stuartarchibald (approved):

stuartarchibald:
Buildfarm ID: `numba_smoketest_cpu_nprgen_37`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cpu_nprgen_37`.

This "passed", the config still includes Python 3.7 builds, which have been removed from `main`, and so those are expected to fail. The only other failure is due to missing packages on one `ppc64le` configuration.

PR #8702, opened 2023-01-11
Supply concrete timeline for objmode fallback deprecation/removal.

stuartarchibald:
This patch:

* Updates the deprecation notices for the objmode fallback removal to include a concrete deadline for removal (0.59.0). It updates the advice in the same.
* Adds a "prominent" warning message for users of `@jit` that do not supply the `nopython` kwarg, or supply it with value False, that the default is changing. The warning links to the deprecation notice in the docs.
* Updates the deprecation warning issued when objmode fallback occurs to tell users which release will remove this behaviour.
* Adds unit tests for the above.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -133,7 +133,6 @@ exclude =
     numba/tests/test_slices.py
     numba/tests/test_mandelbrot.py
     numba/tests/compile_with_pycc.py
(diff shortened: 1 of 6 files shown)


sklam (approved):
Thanks for doing this.

PR #8704, opened 2023-01-11
Added traceback printing capability while NRT debugging

### LINKED ISSUE #8559: Add file and line number information in NRT Debug statements.
Currently NRT Debugging mode outputs the memory addresses of pointers to the objects being `incref`-ed/`decref`-ed. 

https://github.com/numba/numba/blob/bdb238410946f5ef26c83629f36b5ec3cc6cf336/numba/core/runtime/nrtdynmod.py#L57-L58

A good idea for this feature would be to work out if we can somehow add file and line number information into these `incref`/`decref` calls in the NRT debug mode (or some other debug mode). This way we can get a "trace" of every time one was called and where it originated from the Numba source. 

cc @stuartarchibald

kc611:
Resolves #8559

As titled, 
This PR adds traceback printing capability while NRT debugging is enabled. 

Also adds a flag that helps user configure the stack limit for traceback. This functionality can be disabled by setting the same stack limit to zero.

An example NRT Debug with stack limit as 3:

```python
import numpy as np
import numba

@numba.njit
def foo(x):
    return x + 5

x = np.array([0,1,0])
foo(x)
```

Outputs the follows:
```
*** NRT_Decref 1 [0x3245f80]
Traceback:  File "/home/kaustubh/Desktop/WorkSpaces/Numba Workspace/nrt_debug/numba/numba/core/callwrapper.py", line 180, in build_wrapper
    cleanup_manager.emit_cleanup()
  File "/home/kaustubh/Desktop/WorkSpaces/Numba Workspace/nrt_debug/numba/numba/core/callwrapper.py", line 71, in emit_cleanup
    dtor()
  File "/home/kaustubh/Desktop/WorkSpaces/Numba Workspace/nrt_debug/numba/numba/core/callwrapper.py", line 50, in cleanup_arg
    self.context.nrt.decref(self.builder, ty, native.value)

*** NRT_Decref 2 [0x326b5f0]
Traceback:  File "/home/kaustubh/Desktop/WorkSpaces/Numba Workspace/nrt_debug/numba/numba/core/pythonapi.py", line 1405, in from_native_return
    out = self.from_native_value(typ, val, env_manager)
  File "/home/kaustubh/Desktop/WorkSpaces/Numba Workspace/nrt_debug/numba/numba/core/pythonapi.py", line 1419, in from_native_value
    return impl(typ, val, c)
  File "/home/kaustubh/Desktop/WorkSpaces/Numba Workspace/nrt_debug/numba/numba/core/boxing.py", line 432, in box_array
```

### DIFF
--- a/docs/source/reference/envvars.rst
+++ b/docs/source/reference/envvars.rst
@@ -128,9 +128,10 @@ These variables influence what is printed out during compilation of
 
 .. envvar:: NUMBA_DEBUG_NRT_STACK_LIMIT
 
-   Enables user to set traceback stack limit when NUMBA_DEBUG_NRT set to non-zero.
+   Enables user to set traceback stack limit when ``NUMBA_DEBUG_NRT`` set to non-zero.
    By default set to 0 to disable traceback print. Increasing it to a positive non-zero
-   value enables traceback stack printing upto the respective value.
+   value enables traceback stack printing upto the respective value. Does not have any
+   active effects when ``NUMBA_DEBUG_NRT`` flag is set to disabled/zero.
 
 .. envvar:: NUMBA_NRT_STATS
 
--- a/docs/source/reference/envvars.rst
+++ b/docs/source/reference/envvars.rst
@@ -128,10 +128,12 @@ These variables influence what is printed out during compilation of
 
 .. envvar:: NUMBA_DEBUG_NRT_STACK_LIMIT
 
-   Enables user to set traceback stack limit when ``NUMBA_DEBUG_NRT`` set to non-zero.
-   By default set to 0 to disable traceback print. Increasing it to a positive non-zero
-   value enables traceback stack printing upto the respective value. Does not have any
-   active effects when ``NUMBA_DEBUG_NRT`` flag is set to disabled/zero.
+   :ref:`Numba run time (NRT) <arch-numba-runtime>` calls to reference counting
+   operations will print `n` frames of backtrace from the stack present at the time
+   the call was generated during compilation. The purpose of this is to make it
+   easier to locate the code responsible for generating reference counting
+   operations. This environment variable has no effect unless the environment
+   variable ``NUMBA_DEBUG_NRT`` is enabled. Default value is zero.
 
 .. envvar:: NUMBA_NRT_STATS
 
--- a/docs/source/reference/envvars.rst
+++ b/docs/source/reference/envvars.rst
@@ -126,6 +126,11 @@ These variables influence what is printed out during compilation of
    allocation and ``0xDE`` on deallocation, both to help with debugging memory
    leaks.
 
+.. envvar:: NUMBA_DEBUG_NRT_STACK_LIMIT
+
+   Enables user to set traceback stack limit when NUMBA_DEBUG_NRT set to non-zero.
+   By default set to 3. Can also be set to zero to disable traceback print
+
 .. envvar:: NUMBA_NRT_STATS
 
    If set to non-zero, enable the
--- a/docs/source/reference/envvars.rst
+++ b/docs/source/reference/envvars.rst
@@ -129,7 +129,8 @@ These variables influence what is printed out during compilation of
 .. envvar:: NUMBA_DEBUG_NRT_STACK_LIMIT
 
    Enables user to set traceback stack limit when NUMBA_DEBUG_NRT set to non-zero.
-   By default set to 3. Can also be set to zero to disable traceback print
+   By default set to 0 to disable traceback print. Increasing it to a positive non-zero
+   value enables traceback stack printing upto the respective value.
 
 .. envvar:: NUMBA_NRT_STATS
 
--- a/docs/source/reference/envvars.rst
+++ b/docs/source/reference/envvars.rst
@@ -135,6 +135,10 @@ These variables influence what is printed out during compilation of
    operations. This environment variable has no effect unless the environment
    variable ``NUMBA_DEBUG_NRT`` is enabled. Default value is zero.
 
+   Note that this functionality isn't cache friendly, so it should be used with
+   caution. It changes ABI call interface so users are required to clean the cache
+   and recompile their code after setting the flag.
+
 .. envvar:: NUMBA_NRT_STATS
 
    If set to non-zero, enable the
--- a/numba/core/config.py
+++ b/numba/core/config.py
@@ -259,6 +259,9 @@ class _EnvReloader(object):
         # Enable debug prints in nrtdynmod and use of "safe" API functions
         DEBUG_NRT = _readenv("NUMBA_DEBUG_NRT", int, 0)
 
+        # Stack Traceback limit when DEBUG_NRT is enabled
+        DEBUG_NRT_STACK_LIMIT = _readenv("NUMBA_DEBUG_NRT_STACK_LIMIT", int, 3)
+
         # Enable NRT statistics counters
         NRT_STATS = _readenv("NUMBA_NRT_STATS", int, 0)
 
--- a/numba/core/config.py
+++ b/numba/core/config.py
@@ -260,7 +260,7 @@ class _EnvReloader(object):
         DEBUG_NRT = _readenv("NUMBA_DEBUG_NRT", int, 0)
 
         # Stack Traceback limit when DEBUG_NRT is enabled
-        DEBUG_NRT_STACK_LIMIT = _readenv("NUMBA_DEBUG_NRT_STACK_LIMIT", int, 3)
+        DEBUG_NRT_STACK_LIMIT = _readenv("NUMBA_DEBUG_NRT_STACK_LIMIT", int, 0)
 
         # Enable NRT statistics counters
         NRT_STATS = _readenv("NUMBA_NRT_STATS", int, 0)
--- a/numba/core/runtime/context.py
+++ b/numba/core/runtime/context.py
@@ -1,4 +1,8 @@
 import functools
+
+import io
+import traceback
+
 from collections import namedtuple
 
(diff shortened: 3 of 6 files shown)


numba/core/config.py
```diff
@@ -223,6 +223,9 @@ def optional_str(x):
         # Enable debug prints in nrtdynmod and use of "safe" API functions
         DEBUG_NRT = _readenv("NUMBA_DEBUG_NRT", int, 0)
 
+        # Stack Traceback limit when DEBUG_NRT is enabled
+        DEBUG_NRT_STACK_LIMIT = _readenv("NUMBA_DEBUG_NRT_STACK_LIMIT", int, 3)
```
kc611, line 227: Should the default be set to 0, to prevent existing users from getting unnecessary stack trace prints ?

stuartarchibald: I think yes else it's altering a default behaviour. Most of the time NRT debugging is looking for a mismatched pair and it's often known where the inc/decs are coming from, this is adding a new capacity to help in the case where it's not obvious/really involved. Further, the use of `0` as default would mean it can just not be generated _at all_ by default so as to keep the IR smaller?

guilhermeleobas (dismissed):
LGTM. Just wondering if one should add a test for the new env variable.

numba/core/runtime/context.py
```diff
@@ -372,7 +376,27 @@ def _call_incref_decref(self, builder, typ, value, funcname):
...
+                clean_trace = trace.getvalue().split('\n')[:-5]
+
+                if clean_trace:
+                    trace_str += 'Traceback:' + '\n'.join(clean_trace)+'\n\n'
+
+            mod = builder.module
+            # Make global constant for format string
+            cstring = cgutils.voidptr_t
+            fmt_bytes = cgutils.make_bytearray((trace_str + '\00').encode('ascii'))
+            global_fmt = cgutils.global_constant(mod, "printf_format", fmt_bytes)
```
stuartarchibald, line 396:
```suggestion
            global_fmt = cgutils.global_constant(mod, "nrt_debug_printf_format", fmt_bytes)
```
suggest this to make it more clear in the resulting LLVM IR what the global is for.

numba/core/runtime/context.py
```diff
@@ -372,7 +376,27 @@ def _call_incref_decref(self, builder, typ, value, funcname):
...
-            builder.call(fn, [mi])
+
+            trace_str = ""
+
+            if config.DEBUG_NRT and config.DEBUG_NRT_STACK_LIMIT:
+                trace = io.StringIO()
+                traceback.print_stack(limit=config.DEBUG_NRT_STACK_LIMIT+2, file=trace)
+                # The last two stack frames are `_call_incref_decref`
+                # and `traceback.print_stack` which we ignore.
+                clean_trace = trace.getvalue().split('\n')[:-5]
```
stuartarchibald, line 387: If the trace format changes `-5` might not be appropriate, should this instead scan backwards for the `_call_incref_decref` name?

numba/core/runtime/context.py
```diff
@@ -372,7 +376,27 @@ def _call_incref_decref(self, builder, typ, value, funcname):
             # function be called with a NULL pointer?
             fn.args[0].add_attribute("noalias")
             fn.args[0].add_attribute("nocapture")
-            builder.call(fn, [mi])
+
+            trace_str = ""
+
+            if config.DEBUG_NRT and config.DEBUG_NRT_STACK_LIMIT:
+                trace = io.StringIO()
+                traceback.print_stack(limit=config.DEBUG_NRT_STACK_LIMIT+2, file=trace)
```
stuartarchibald, line 384:
```suggestion
                traceback.print_stack(limit=config.DEBUG_NRT_STACK_LIMIT + 2, file=trace)
```

numba/core/runtime/context.py
```diff
@@ -372,7 +376,27 @@ def _call_incref_decref(self, builder, typ, value, funcname):
...
+                # The last two stack frames are `_call_incref_decref`
+                # and `traceback.print_stack` which we ignore.
+                clean_trace = trace.getvalue().split('\n')[:-5]
+
+                if clean_trace:
+                    trace_str += 'Traceback:' + '\n'.join(clean_trace)+'\n\n'
```
stuartarchibald, line 390:
```suggestion
                    trace_str += 'Traceback:' + '\n'.join(clean_trace) + '\n\n'
```
also, perhaps just use an `f-string`?

numba/core/runtime/context.py
```diff
@@ -372,7 +376,27 @@ def _call_incref_decref(self, builder, typ, value, funcname):
...
+
+            mod = builder.module
+            # Make global constant for format string
+            cstring = cgutils.voidptr_t
```
stuartarchibald, line 394: Is this variable needed? Could the subsequent `bitcast` just refer to `cgutils.voidptr_t` directly?

stuartarchibald (commented):
Thanks for the patch, I've given this a review and tested it manually, it's going to be useful! As @guilhermeleobas noted, some tests are needed. I also wonder if it'd be a good idea to actually change the NRT refop function signature only if this env var is enabled, such that the refop functions are called with `(meminfo,)` by default (as they are in `main`), and then if this env var is set the function generation changes the signature to take `(meminfo, void * trace)` and a trace is supplied.

numba/core/runtime/context.py
kc611: Addressed in 170f84f

numba/core/runtime/context.py
kc611: Addressed in 170f84f

numba/core/runtime/context.py
kc611: Addressed in 170f84f

numba/core/runtime/context.py
kc611: Addressed in 170f84f

numba/core/runtime/context.py
```diff
@@ -372,7 +376,27 @@ def _call_incref_decref(self, builder, typ, value, funcname):
...
+            fmt_bytes = cgutils.make_bytearray((trace_str + '\00').encode('ascii'))
+            global_fmt = cgutils.global_constant(mod, "printf_format", fmt_bytes)
+            ptr_fmt = builder.bitcast(global_fmt, cstring)
+
+            builder.call(fn, [mi, ptr_fmt])
```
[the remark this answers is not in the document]
kc611: Addressed in 170f84f

docs/source/reference/envvars.rst
```diff
@@ -148,6 +148,13 @@ These variables influence what is printed out during compilation of
    allocation and ``0xDE`` on deallocation, both to help with debugging memory
    leaks.
 
+.. envvar:: NUMBA_DEBUG_NRT_STACK_LIMIT
+
+   Enables user to set traceback stack limit when ``NUMBA_DEBUG_NRT`` set to non-zero.
+   By default set to 0 to disable traceback print. Increasing it to a positive non-zero
+   value enables traceback stack printing upto the respective value. Does not have any
+   active effects when ``NUMBA_DEBUG_NRT`` flag is set to disabled/zero.
```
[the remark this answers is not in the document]
kc611: Addressed in 170f84f

esc:
@kc611 @guilhermeleobas -- folks, what is the state here? Is this waiting to be signed off by @stuartarchibald ?

guilhermeleobas:
> @kc611 @guilhermeleobas -- folks, what is the state here? Is this waiting to be signed off by @stuartarchibald ?

I think so

swap357:
@kc611 could you merge latest main on this branch ?

numba/core/runtime/context.py
```diff
@@ -1,4 +1,8 @@
 import functools
+
+import io
+import traceback
```
swap357: can we lazy import these into if block if not used anywhere else ?

swap357 (commented):
I've reviewed and tested this PR. The debug feature is properly gated behind `NUMBA_DEBUG_NRT` and `NUMBA_DEBUG_NRT_STACK_LIMIT`, so the normal code path is not affected.
Known issue - cached functions compiled without `NUMBA_DEBUG_NRT_STACK_LIMIT` will use the 1-arg NRT call. When cache is loaded with flag enabled, it expects 2 args, and traceback might contain garbage. Not a blocker since this is debug feature.

numba/core/runtime/context.py
kc611: Fixed in e7e4ced

swap357 (approved):
I went through PR again, the changes look good. Note for usage with caching is noted on docs. Looks good to me. Thank you @kc611 for the efforts!

swap357:
test usage-
```
❯ cat issues/test.py                                                   
import numpy as np
from numba import jit


@jit
def foo():
    a = np.arange(10.0)
    return a[::2]

print(foo())

❯ NUMBA_DEBUG_NRT=1 NUMBA_DEBUG_NRT_STACK_LIMIT=3 python issues/test.py
*** NRT_Decref 2 [0x13c94a5d0]
Traceback:  File "/Users/mac357/dev/numba/numba/core/pythonapi.py", line 1443, in from_native_return
    out = self.from_native_value(typ, val, env_manager)
  File "/Users/mac357/dev/numba/numba/core/pythonapi.py", line 1457, in from_native_value
    return impl(typ, val, c)
  File "/Users/mac357/dev/numba/numba/core/boxing.py", line 440, in box_array
  File "/Users/mac357/dev/numba/numba/core/runtime/context.py", line 417, in decref
    self._call_incref_decref(builder, typ, value, "NRT_decref")

[0. 2. 4. 6. 8.]
```

PR #8706, opened 2023-01-11
Fix doctest for `@vectorize`

### LINKED ISSUE #8694: pytest doctest can't discover functions with @vectorize decorator
## Reporting a bug

This problem seems very similar to the issue raised in #6308, which has already been fixed.

Here's my minimum working example:

```python
"""
Test that all docstrings are the same:

>>> len({f.__doc__ for f in (a, b, c, d)})
1
"""
from numba import guvectorize, int64, njit, vectorize


def a():
    """>>> x = 1"""
    return 1


@njit
def b():
    """>>> x = 1"""
    return 1


@guvectorize([(int64[:], int64, int64[:])], "(n),()->(n)")
def c(x, y, res):
    """>>> x = 1"""
    for i in range(x.shape[0]):
        res[i] = x[i] + y


@vectorize([int64(int64, int64)])
def d(x, y):
    """>>> x = 1"""
    return x + y

```

When I run pytest I get the following results:

```
pytest -rs -sv --doctest-modules mwe.py
============================== test session starts ===============================
platform win32 -- Python 3.10.8, pytest-7.2.0, pluggy-1.0.0 -- C:\Users\MYUSER\Miniconda3\envs\fdm\python.exe
cachedir: .pytest_cache
rootdir: C:\Users\MYUSER\local_work\sketches\temp
plugins: cov-4.0.0
collected 4 items

mwe.py::mwe PASSED
mwe.py::mwe.a PASSED
mwe.py::mwe.b PASSED
mwe.py::mwe.c PASSED

=============================== 4 passed in 0.71s ================================
```
As all docstrings are the same, I would expect pytest to run a doctest for the function `d`, as it did for `a`, `b`, and `c`.

sklam:
Closes #8694

- Apply same patch as in https://github.com/numba/numba/blob/009ae4e84ede1b6a750682ed5c357601c496495a/numba/np/ufunc/gufunc.py#L31 for gufunc
- Add a `test_doctest.py` to exercise doctest functionality for decorators, including `njit`, `vectorize`, `guvectorize`

### DIFF
--- a/numba/np/ufunc/dufunc.py
+++ b/numba/np/ufunc/dufunc.py
@@ -1,3 +1,5 @@
+import functools
+
 from numba import jit, typeof
 from numba.core import cgutils, types, serialize, sigutils
 from numba.core.extending import is_jitted
@@ -83,6 +85,7 @@ class DUFunc(serialize.ReduceMixin, _internal._DUFunc):
                          cache=cache,
                          **targetoptions)(py_func)
         self._initialize(dispatcher, identity)
+        functools.update_wrapper(self, py_func)
 
     def _initialize(self, dispatcher, identity):
         identity = ufuncbuilder.parse_identity(identity)
--- a/numba/tests/doctest_usecase.py
+++ b/numba/tests/doctest_usecase.py
@@ -1,0 +1,31 @@
+"""
+Test that all docstrings are the same:
+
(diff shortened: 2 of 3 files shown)


guilhermeleobas (approved):
LGTM

PR #8670, opened 2022-12-14
[WIP] Python 3.11 tracing

AndrewVallette:
This is a work-in-progress for enabling tracing support with Python 3.11. The initial effort has been to get this to cleanly compile with an updated `C_TRACE` and related functions.

PR #8711, opened 2023-01-12
Python 3.11 tracing support (continuation of #8670).

sklam:


sklam:
(just did the close-reopen dance to kick CI)

numba/_dispatcher.cpp
```diff
@@ -27,15 +28,96 @@
...
+    if (tstate->tracing) {
+        return 0;
+    }
+    if (frame == NULL) {
+        return -1;
+    }
+    int old_what = tstate->tracing_what;
+    tstate->tracing_what = what;
+    PyThreadState_EnterTracing(tstate);
+    result = func(obj, frame, PyTrace_CALL, NULL);
```
stuartarchibald, line 56: There's also nothing that updates the frame's line number.

numba/_dispatcher.cpp
```diff
@@ -568,8 +650,12 @@ call_cfunc(Dispatcher *self, PyObject *cfunc, PyObject *args, PyObject *kws, PyO
     tstate = PyThreadState_GET();
 
 #if (PY_MAJOR_VERSION >= 3) && (PY_MINOR_VERSION >= 11)
-    #warning "FIXME: THIS IS STUB CODE"
-    if (tstate->tracing && tstate->c_profilefunc)
+    /*
+     * On Python 3.11, _PyEval_EvalFrameDefault stop using PyTraceInfo since
```
stuartarchibald, line 654:
```suggestion
     * On Python 3.11, _PyEval_EvalFrameDefault stops using PyTraceInfo since 
```

stuartarchibald (commented):
Thanks for the patch. Couple of small things to look at else looks about as expected. I think with a lot of effort some of these variations between Python minor versions could be combined but it may not be worth the risk of doing so and potentially introducing mistakes as a result.

numba/_dispatcher.cpp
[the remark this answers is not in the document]
sklam: Right, need to fix it. But, `what` cause segfault.

sklam: I don't think we need to anymore. The test passes and it's asserting that the lineno is in there.

stuartarchibald: From OOB discussion, @sklam confirms that the line numbers are present already. The CPython Python 3.11 implementation has the frame as a `_PyInterpreterFrame` struct which is internal. I think it's possible to move the code here to work more like the code in CPython, but it will come at the cost of Numba copying in more internal CPython implementation details. Considering copying-in as the alternative, as the current method seems to work, I think it is an acceptable trade off.

stuartarchibald (approved):
Thanks for the patch and fixes. I think this completes the Python 3.11 support!

zanejgr:
> Thanks for the patch and fixes. I think this completes the Python 3.11 support!

Apologies if this is the wrong place to ask this, but does this mean Numba will be updated on PyPI/Conda-forge today?

stuartarchibald:
> > Thanks for the patch and fixes. I think this completes the Python 3.11 support!
> 
> Apologies if this is the wrong place to ask this, but does this mean Numba will be updated on PyPI/Conda-forge today?

It means the code part is potentially finished, Numba won't release a new version until it has been through testing and an RC (which should be soon). Perhaps subscribe to #8304 for updates?

PR #8716, opened 2023-01-16
CI: Use `set -e` in "Before Install" step and fix install

gmarkall:
`miniconda.sh` needs to be run with `bash` because it uses `[[`, which is not available in all `sh` implementations.

We also add `set -e` to the "Before Install" step so that problems installing Miniconda cause the build to fail.

### DIFF
--- a/buildscripts/azure/azure-linux-macos.yml
+++ b/buildscripts/azure/azure-linux-macos.yml
@@ -22,11 +22,13 @@ jobs:
       displayName: 'Before Install'
 
     - script: |
+        set -e
         export PATH=$HOME/miniconda3/bin:$PATH
         buildscripts/incremental/build.sh
       displayName: 'Build'
 
     - script: |
+        set -e
         export PATH=$HOME/miniconda3/bin:$PATH
         conda install -y flake8
         flake8 numba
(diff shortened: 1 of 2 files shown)


buildscripts/azure/azure-linux-macos.yml
```diff
@@ -13,6 +13,7 @@ jobs:
 
   steps:
     - script: |
+        set -e
```
stuartarchibald, line 16: Would it be worth adding `set -e` to the start of all of these `script` steps to catch similar issues in future?

stuartarchibald (commented):
Thanks for the patch @gmarkall, it looks good. I've left one minor comment in-line that may be worth addressing now but could also be deferred.

buildscripts/azure/azure-linux-macos.yml
gmarkall: Yes, I hadn't noticed that it was on every step in the llvmlite list of steps. Now added.

stuartarchibald (approved):

gmarkall:
Is there any other CI that needs running for this? (just checking about the "Waiting for CI" label)

stuartarchibald:
> Is there any other CI that needs running for this? (just checking about the "Waiting for CI" label)

Not that I'm aware of, I just approved ahead of CI completing so left the label just in case! Will mark as ready to merge now.

PR #8720, opened 2023-01-18
Enable coverage for subprocess testing

sklam:
I noticed that in recent PR reviews that coverage tool do not see into subprocess test (e.g. `TestCase.run_test_in_subprocess`). This patch fixes it.

### DIFF
--- a/numba/tests/support.py
+++ b/numba/tests/support.py
@@ -57,7 +57,7 @@ try:
 except ImportError:
     scipy = None
 
-# Make sure that coverage is setup.
+# Make sure that coverage is set up.
 try:
     import coverage
 except ImportError:
--- a/numba/tests/support.py
+++ b/numba/tests/support.py
@@ -52,12 +52,18 @@ from numba.core.extending import (
 from numba.core.datamodel.models import OpaqueModel
 from numba.pycc.platform import external_compiler_works
 
-
 try:
     import scipy
 except ImportError:
     scipy = None
 
+# Make sure that coverage is setup.
+try:
+    import coverage
+except ImportError:
+    pass
+else:
+    coverage.process_startup()
 
 enable_pyobj_flags = Flags()
 enable_pyobj_flags.enable_pyobject = True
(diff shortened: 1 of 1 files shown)


numba/tests/support.py
```diff
@@ -583,6 +589,10 @@ def subprocess_test_runner(self, test_module, test_class=None,
         cmd = [sys.executable, '-m', 'numba.runtests', fully_qualified_test]
         env_copy = os.environ.copy()
         env_copy['SUBPROC_TEST'] = '1'
+        try:
+            env_copy['COVERAGE_PROCESS_START'] = os.environ['COVERAGE_RCFILE']
```
stuartarchibald, line 593: What sets `COVERAGE_RCFILE` ?

numba/tests/support.py
```diff
@@ -52,12 +52,18 @@
 from numba.core.datamodel.models import OpaqueModel
 from numba.pycc.platform import external_compiler_works
 
-
```
stuartarchibald, line 55: Should this be removed? There's usually 2 blank lines after `import` decls?

stuartarchibald (commented):
Thanks for the patch. The changes make sense with respect to the coverage docs on "Measuring sub-processes", couple of minor questions to resolve else looks good.

numba/tests/support.py
sklam: `coverage run` sets it.

numba/tests/support.py
sklam: flake8 didn't complain when i removed the file from the ignore list.

numba/tests/support.py
stuartarchibald: Thanks for confirming.

numba/tests/support.py
stuartarchibald: ok, thanks for checking.

numba/tests/support.py
```diff
@@ -52,12 +52,18 @@
...
 try:
     import scipy
 except ImportError:
     scipy = None
 
+# Make sure that coverage is setup.
```
stuartarchibald, line 60:
```suggestion
# Make sure that coverage is set up.
```

stuartarchibald (commented):
A typo to fix, else looks good, thanks @sklam

sklam:
i've accepted the typo fix

PR #8723, opened 2023-01-21
Check for void return type in `cuda.compile_ptx`

### LINKED ISSUE #8717: `compile_ptx()` allows compilation of kernel with non-`void` return type
For example:

```python
from numba import cuda, int32

def f(x, y):
    return x[0] + y[0]

ptx, resty = cuda.compile_ptx(f, (int32[::1], int32[::1]))

print(resty)
# int64
```

compiles, and generates PTX where the function body returns nothing (heavily edited for clarity, but the idea is represented:

```assembly
.visible .entry f (
    // args omitted for brevity
)
{
	ret;
}
```

Usually we check that the kernel has a void return type in the `@cuda.jit` decorator, which is why this slips by in `compile_ptx`. The check should probably be pushed a bit deeper to cover both uses.

(cc @brandonwillard )

brandonwillard:
This PR adds a check for void return types to `cuda.compile_ptx`.

Closes #8717

### DIFF
--- a/numba/cuda/tests/cudapy/test_compiler.py
+++ b/numba/cuda/tests/cudapy/test_compiler.py
@@ -116,6 +116,13 @@ class TestCompileToPTX(unittest.TestCase):
         ptx, resty = compile_ptx(f, [], lineinfo=True)
         self.check_line_info(ptx)
 
+    def test_non_void_return_type(self):
+        def f(x, y):
+            return x[0] + y[0]
+
+        with self.assertRaisesRegex(TypeError, 'must have void return type'):
+            compile_ptx(f, (uint32[::1], uint32[::1]))
+
 
 @skip_on_cudasim('Compilation unsupported in the simulator')
 class TestCompileToPTXForCurrentDevice(CUDATestCase):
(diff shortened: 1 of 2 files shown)


numba/cuda/tests/cudapy/test_compiler.py
```diff
@@ -116,6 +116,13 @@ def f():
         ptx, resty = compile_ptx(f, [], lineinfo=True)
         self.check_line_info(ptx)
 
+    def test_non_void_return_type(self):
+        def f(x, y):
+            return x[0] + y[0]
+
+        with self.assertRaises(TypeError):
```
gmarkall, line 123: It would be handy also to check the error message, just in case something bizarre happens and causes some other `TypeError`:

```suggestion
        with self.assertRaisesRegex(TypeError, 'must have void return type'):
```

gmarkall (changes requested):
Many thanks for the fix! I just made a small suggestion on the diff.

I do think there's some further fiddling around we could do so that there's only one instance of this check needed (rather than the added one and the one in decorators.py), but I'm doing some heavy refactoring in various areas where options are handled in order to fix the propagation of options to jit decorators, so I'd like to avoid suggesting that for now.

gmarkall:
gpuci run tests

gmarkall (approved):
Many thanks for the fix!

PR #8725, opened 2023-01-24
Split parfor pass into 3 parts. (old)

DrTodd13:
Split parfor pass into 1) parfor conversion, 2) fusion, and 3) post-fusion miscellaneous passes.  Adding a lowerer tag to the parfor node, which defaults to the existing lowerer.  Applications that change the pipeline can insert passes between 1) and 2) that change the lowerer.  Parfors with different lowerers will then not fuse.  The lowerer tag is then used during the lowering phase to lowerer this parfor.  This allows different parfors in the same function to be lowered differently, e.g., for CPU or GPU.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,12 +1,19 @@
 [flake8]
 ignore =
-    E20,   # Extra space in brackets
-    E231,E241,  # Multiple spaces around ","
-    E26,   # Comments
-    E731,  # Assigning lambda expression
-    E741,  # Ambiguous variable names
-    W503,  # line break before binary operator
-    W504,  # line break after binary operator
+    # Extra space in brackets
+    E20,
+    # Multiple spaces around ","
+    E231,E241,
+    # Comments
+    E26,
+    # Assigning lambda expression
+    E731,
+    # Ambiguous variable names
+    E741,
+    # line break before binary operator
+    W503,
+    # line break after binary operator
+    W504,
 max-line-length = 80
 
 exclude =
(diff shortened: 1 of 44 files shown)


DrTodd13:
@sklam @stuartarchibald This PR is failing on tests.test_inlining.TestInlining.test_inline_call_after_parfor.  This test is:

sklam:
@DrTodd13, Do you need this PR for the 0.57 release?

DrTodd13:
@sklam I was looking at another PR of mine (#7171) that is tagged for inclusion in 0.57 and @stuartarchibald didn't like the approach I took there but I just realized that this PR is the solution that Stuart wanted.  If during the gufunc compile we change the pipeline to include the third parfor stage (but not the first two) then I think we get the same effect as PR 7171.  I will experiment with this in the afternoon so given that we were going to include 7171 in 0.57 and this PR is very likely to be correct then perhaps we should combine them for 0.57.  What do you think?

DrTodd13:
@sklam Okay....I got parfor gufunc compilation working with a very minimal pipeline that includes the prelowering parfor phase but not the other parfor phases.

sklam:
ok, i'm adding this for 0.57 for now.

sklam:
The latest commit is failing because of missing `cr` (the compiler result):

https://dev.azure.com/numba/numba/_build/results?buildId=14377&view=logs&j=cdf58ab6-1542-56e8-2141-f11736c48688&t=a6ad49d1-c022-53a8-dec0-947c577c759a&l=18299

DrTodd13:
> The latest commit is failing because of missing `cr` (the compiler result):
> 
> https://dev.azure.com/numba/numba/_build/results?buildId=14377&view=logs&j=cdf58ab6-1542-56e8-2141-f11736c48688&t=a6ad49d1-c022-53a8-dec0-947c577c759a&l=18299

Fixed.

sklam:
superseded by https://github.com/numba/numba/pull/8804

Issue #8707, opened 2023-01-11
Move baseline supported NumPy version to 1.21.


stuartarchibald:
Tracking issue as per title from discussion at Numba public meeting on 2022-12-06 "Based on NEP-029, 0.57 will have the baseline of 1.21.". 
* [ ] May be merge #8691 (not essential to do this move, but makes it easier)
* [ ] Update CI matrix
* [ ] grep for `1\.X` and `1, X` for `X` in `18`, `19`, `20` and fix up.
* [ ] Fix NumPy guard in `__init__.py`
* [ ] Fix NumPy guard in `__setup__.py`

PR #8726, opened 2023-01-25
Make Numba dependency check run ahead of Numba internal imports.

### LINKED ISSUE #8718: Numba `__init__.py` not catching incompatible NumPy version early enough
## Reporting a bug

Following on from numerous issues reported against #8615, no released version of Numba is compatible with NumPy 1.24, this is stated in the docs e.g.:
https://github.com/numba/numba/blob/288a38bbd5a15418a211bf067878dfdf3c139509/docs/source/user/installing.rst?plain=1#L281

The problem in practice seems to be as follows:
1. As of the version tags on the `release0.56` branch of Numba (though this also probably impacts all Numba versions), all builds made against these tags are not compatible with NumPy 1.24 due to some changes in NumPy `ufunc` objects, see #8691 for necessary adjustments. Essentially the `resolve_dtypes` method was added which tripped a piece of guard code in Numba that returns `-1` without an `Exception` being set, this manifests as a `SystemError` on import of the C extension containing this code.
2. Tools installing Numba wheels e.g. `pip` are not constrained by the maximum NumPy version in metadata as they are in `conda` packages. The mechanism to prevent use of incompatible installations in this case is to raise an `Exception` at runtime from Numba's `__init__.py`: https://github.com/numba/numba/blob/288a38bbd5a15418a211bf067878dfdf3c139509/numba/__init__.py#L133-L155 called here: https://github.com/numba/numba/blob/288a38bbd5a15418a211bf067878dfdf3c139509/numba/__init__.py#L206

I think the issue is that the mechanism in 2. fails if a C extension with the problem in 1. is imported as part of Numba's `__init__` ahead of the call to ` _ensure_critical_deps`. i.e. the check for compatibility of dependencies is too late.

stuartarchibald:
This patch makes the runtime check that ensures the Python, NumPy and SciPy verions are accepted by Numba occur before importing any of Numba's internal modules. This is to prevent a case where one of Numba's modules is incompatible with some version of a critical dependency but the error mode is e.g. SystemError, this preventing the correct reporting of the actual problem to the user.

Fixes #8718

### DIFF
--- a/numba/__init__.py
+++ b/numba/__init__.py
@@ -7,6 +7,55 @@ import re
 import sys
 import warnings
 
+
+# ---------------------- WARNING WARNING WARNING ----------------------------
+# THIS MUST RUN FIRST, DO NOT MOVE... SEE DOCSTRING IN _ensure_critical_deps
+def _ensure_critical_deps():
+    """
+    Make sure the Python, NumPy and SciPy present are supported versions.
+    This has to be done _before_ importing anything from Numba such that
+    incompatible versions can be reported to the user. If this occurs _after_
+    importing things from Numba and there's an issue in e.g. a Numba c-ext, a
+    SystemError might have occurred which prevents reporting the likely cause of
+    the problem (incompatible versions of critical dependencies).
+    """
+    #NOTE THIS CODE SHOULD NOT IMPORT ANYTHING FROM NUMBA!
+
+    def extract_version(mod):
+        return tuple(map(int, mod.__version__.split('.')[:2]))
+
+    PYVERSION = sys.version_info[:2]
+
+    if PYVERSION < (3, 8):
+        msg = ("Numba needs Python 3.8 or greater. Got Python "
+               f"{PYVERSION[0]}.{PYVERSION[1]}.")
+        raise ImportError(msg)
+
+    import numpy as np
+    numpy_version = extract_version(np)
+
+    if numpy_version < (1, 18):
+        msg = (f"Numba needs NumPy 1.18 or greater. Got NumPy "
+               f"{numpy_version[0]}.{numpy_version[1]}.")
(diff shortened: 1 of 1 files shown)


numba/__init__.py
```diff
@@ -7,6 +7,55 @@
...
+
+    if PYVERSION < (3, 8):
+        msg = ("Numba needs Python 3.8 or greater. Got Python "
+               f"{PYVERSION[0]}.{PYVERSION[1]}.")
+        raise ImportError(msg)
+
+    import numpy as np
+    numpy_version = extract_version(np)
+
+    if numpy_version < (1, 18):
```
gmarkall, line 37: Should this be 1.21? I think we concluded recently that 0.57 would support 1.21-1.24.

gmarkall (approved):
I think this does the right things at the right time - I was thinking when I commented on the diff that it should check for NumPy 1.21, but I realise now this will break current CI. Once this is merged I'll then update #8691, which modifies the matrix so that versions less than 1.21 are dropped.

numba/__init__.py
gmarkall: I concluded for now that it shouldn't be as it will break CI - but it will need updating in #8691.

gmarkall:
(I think the CI fail is due to #8680 not yet being merged rather than a real issue)

numba/__init__.py
stuartarchibald: Thanks for checking, I think it needs to be moved to NumPy 1.21 as part of a PR that closes #8707, but for now needs to stay as-is else the current CI configuration will break.

stuartarchibald:
> (I think the CI fail is due to #8680 not yet being merged rather than a real issue)

I think so, the rebuild against `main` (at e0c1e4c) has passed.

PR #7608, opened 2021-11-30
[WIP] Updating TBB backend calls for tbb=2022.1

kozlov-alexey:
Adapts TBB parallel backend to API changes in tbb=2022.1 (task_scheduler_handle::get and task_scheduler_handle::release are removed)

PR #8727, opened 2023-01-25
Move minimum supported TBB version to 2021.6

stuartarchibald:
This is a continuation of PR #7608 (the starting point is that PR rebased on `main`). It moves the minimum supported TBB version to 2021.6 and updates source, docs, packaging configuration and CI to reflect this.

With thanks to @kozlov-alexey and @Hardcode84 for the original patches and review in #7608.

### DIFF
--- a/buildscripts/condarecipe.local/meta.yaml
+++ b/buildscripts/condarecipe.local/meta.yaml
@@ -40,7 +40,7 @@ requirements:
     # NOTE: 2021.1..2021.5 are API compatible for Numba's purposes.
     # NOTE: ppc64le exclusion is temporary until packages are more generally
     #       available.
-    - tbb-devel >=2021,<2021.6       # [not (aarch64 or ppc64le)]
+    - tbb-devel >=2021.6       # [not (aarch64 or ppc64le)]
   run:
     - python >=3.8
     # NumPy 1.22.0, 1.22.1, 1.22.2 are all broken for ufuncs, see #7756
@@ -50,8 +50,8 @@ requirements:
     # On channel https://anaconda.org/numba/
     - llvmlite >=0.40.0dev0,<0.40
   run_constrained:
-    # If TBB is present it must be at least version 2021
-    - tbb >=2021    # [not (aarch64 or ppc64le)]
+    # If TBB is present it must be at least version 2021.6
+    - tbb >=2021.6    # [not (aarch64 or ppc64le)]
     # avoid confusion from openblas bugs
     - libopenblas !=0.3.6      # [x86_64]
     # 0.3.17 buggy on M1 silicon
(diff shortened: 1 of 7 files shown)


sklam:
BFID `numba_smoketest_cpu_yaml_159`

sklam:
No tbb-devel 2021.6 on win-32. Should we drop TBB support for win-32?

tbb-devel availability:
```
2021.2.0: {'win-32', 'win-64', 'osx-64', 'linux-ppc64le', 'linux-aarch64', 'linux-64'}
2021.3.0: {'win-64', 'osx-64', 'linux-ppc64le', 'osx-arm64', 'linux-64'}
2021.4.0: {'win-32', 'win-64', 'osx-64', 'linux-ppc64le', 'linux-aarch64', 'osx-arm64', 'linux-64'}
2021.5.0: {'win-32', 'win-64', 'osx-64', 'linux-ppc64le', 'linux-s390x', 'linux-aarch64', 'osx-arm64', 'linux-64'}
2021.6.0: {'win-64', 'osx-64', 'linux-ppc64le', 'linux-s390x', 'linux-aarch64', 'osx-arm64', 'linux-64'}
```

seibert:
Isn't the parallel target already disabled on win-32, so having TBB as a thread pool option on that platform doesn't matter either way?

sklam (commented):
Buildfarm is mostly ok. Just need to disable tbb on win32 due to lack of tbb package.

stuartarchibald:
> Isn't the parallel target already disabled on win-32, so having TBB as a thread pool option on that platform doesn't matter either way?

That is the case. I think originally the 32-bit route was left open in case `parfors` could support it, but that's still not the case and 32-bit in general has been dropped.

@sklam RE: the https://github.com/numba/numba/pull/8727#pullrequestreview-1291772543, I presume this needs needs a conda recipe update to reflect that there's no TBB support on win32? This will be an update to `meta.yaml` to add in the package constraint and then an update to `bld.bat` to add logic for "if this is win32 then export env var `NUMBA_DISABLE_TBB=1`". Is that what you were thinking?

stuartarchibald:
This also needs fixing up with respect to this change:
https://github.com/numba/numba/blob/7b2637f875901d27513550c92c44541dfebad74e/numba/np/ufunc/parallel.py#L344-L375

stuartarchibald:
@sklam 72c0d7a updates the conda recipe and 39b1c7d fixes up the runtime version check.

sklam (approved):
Changes look good. I just need to verify the buildscript change in the buildfarm again.

stuartarchibald:
@sklam as discussed OOB, this should be rebased on `main` as of 291269b.

stuartarchibald:
Am going to close this so as to fix up the history, will open a new PR.

PR #8728, opened 2023-01-26
Fix flake8 checks since upgrade to flake8=6.x

stuartarchibald:
It seems like public CI updating to use flake8=6.x has resulted in a few additional formatting violations being picked up. This patch simply addresses these and nothing more.


esc (approved):
The changes look fine, now waiting on CI. I am pretty sure it will pass, as I tried locally with `flake8 numba` and received no further errors and the process exited fine with a `0`.

guilhermeleobas (approved):

esc:
@stuartarchibald thank you for the fix and @guilhermeleobas thank you for the review!

PR #7367, opened 2021-09-04
developer contributing docs - llvm, sqlite3, sphinx-rtd-theme, and make html

sgbaird:
Here's a summary:
- There seems to be an issue in the docs related to the channel used for llvm
- There seems to be a Windows-specific bug where conda can't find sqlite3.dll when trying to commit using the pre-commit hooks, so I added some instructions for overcoming that issue.
- The docs can't build because it `sphinx-rtd-theme` isn't installed, so I added that to the pip install
- I added instructions that `make html` is run under `docs` (should be obvious to anyone familiar with make, seeing as there is no MAKEFILE under `docs/source`, but saves someone the extra step)

## Reference an existing issue

#7362 #7354

PR #8730, opened 2023-01-27
Update dev-docs

esc:
This is variant of #7367 -- but with conflicts resolved to `main` on 2022-01-27.


esc:
@gmarkall I just read through it again, and perhaps we can remove the channel setup. The docs now list using the `numba/label/dev::` syntax.

gmarkall (changes requested):
I agree, we can remove the section about the channel setup since the docs already specify installing llvmlite from the `numba/label/dev` channel:

```
$ conda create -n numbaenv python=3.10 numba/label/dev::llvmlite numpy scipy jinja2 cffi
```

esc:
> I agree, we can remove the section about the channel setup since the docs already specify installing llvmlite from the `numba/label/dev` channel:
> 
> ```
> $ conda create -n numbaenv python=3.10 numba/label/dev::llvmlite numpy scipy jinja2 cffi
> ```

Thank you @gmarkall -- I opted to remove the section. I also noticed, while writing, that the installation instructions for the dependencies were somewhat conflated. With examples in the paragraph above and then a concrete command below. In the interest of good reading flow, I opted to list all the dependencies for the docs within a single command. I am hoping this is less confusing and makes it very clear what dependencies are needed to build the docs for Numba. May I please have a re-review? Thank you.

docs/source/developer/contributing.rst
```diff
@@ -92,8 +92,9 @@ When working with a source checkout of Numba you will also need a development
 build of llvmlite. These are available from the ``numba/label/dev`` channel on
 `anaconda.org <https://anaconda.org/numba/llvmlite>`_.
 
-
-Then, to create an environment with a few of the most common dependencies::
+Then, to create an environment with a few of the most common dependencies. Note
+the use of the double-colon syntax (``numba/label/dev::llvmlite``) to install
+the latest development version of the llvmlite library.
```
gmarkall, line 97: I think "Then" doesn't apply because nothing was done prior. The text also needs a `::` to render the shell command as code:

```suggestion
To create an environment with the required dependencies, noting the use of the 
double-colon syntax (``numba/label/dev::llvmlite``) to install the latest
development version of the llvmlite library::
```

docs/source/developer/contributing.rst
```diff
@@ -461,16 +462,16 @@ Main documentation
 ''''''''''''''''''
 
 This documentation is under the ``docs`` directory of the `Numba repository`_.
-It is built with `Sphinx <http://sphinx-doc.org/>`_ and
-`numpydoc <https://numpydoc.readthedocs.io/>`_, which are available using
-conda or pip; i.e. ``conda install sphinx numpydoc``.
+It is built with `Sphinx <http://sphinx-doc.org/>`_, `numpydoc
+<https://numpydoc.readthedocs.io/>`_ and the `sphinx-rtd-theme
+<https://sphinx-rtd-theme.readthedocs.io/en/stable/>`.
```
gmarkall, line 467: I think this needs a following underscore to render as a link:

```suggestion
<https://sphinx-rtd-theme.readthedocs.io/en/stable/>_`.
```

gmarkall (changes requested):
There are a couple of rendering errors - suggestions added on the diff that I think will resolve them.

docs/source/developer/contributing.rst
esc: Screen Shot 2023-05-04 at 10 23 26

Looks good.

docs/source/developer/contributing.rst
esc: Screen Shot 2023-05-04 at 10 24 17
 
this one also

esc:
@gmarkall thank you for the feedback, I have addressed the review comments and this is now ready for another round of review.

docs/source/developer/contributing.rst
esc: The underscore must come after the backtick. I fixed this in 4a030ee85f716e5b2c523ff2f02a57a27560a79b

gmarkall (approved):
Many thanks for the updates!

esc:
> Many thanks for the updates!

Thank your for the review!

PR #8731, opened 2023-01-27
Add incubator project policy

sklam:


docs/source/developer/incubator.rst
```diff
@@ -0,0 +1,36 @@
+.. Adapted from the dask-contrib policy document https://github.com/dask/dask/blob/89ecb076bbbbd767eeef7e8cd1040cb838e33a5c/docs/source/contrib.rst
```
stuartarchibald, line 1: Does `<project root>/LICENSES.third-party` need an update?

stuartarchibald (commented):
Thanks for writing this @sklam. I've given it an initial review with view of providing early feedback. My main concern at present is ensuring that the "graduation" process for an incubator project is well defined. Thanks again!

sklam:
I closed this to rethink how the policy should look like.

Issue #2445, opened 2017-07-01
Please add np.argpartition implementation in nopython mode


ethen8181:
Hi, this is more of a feature request. I checked the documentation and found that `np.argpartition` is not supported at the moment, it would be great if this functionality is included!! Thanks for all the hard work.

ethen8181:
Hi team, just wanted to check whether there's an update on this? Like in this in the works or something. Thanks!

seibert:
No updates.  We'd happily accept a contributed implementation, but have not had time to do it ourselves.

sleighsoft:
Would love to see this. What would I have to do to add support for this to numba?

sleighsoft:
Would love to see this. What would I have to do to add support for this to numba?

stuartarchibald:
@sleighsoft The use of Numba's extension API `@overload` decorator is strongly recommended for this task. A guide to using `@overload` is [here](http://numba.pydata.org/numba-doc/dev/extending/overloading-guide.html?highlight=overload) and API documentation is [here](http://numba.pydata.org/numba-doc/latest/extending/high-level.html#implementing-functions).

It may be useful to take a look at examples of pull requests adding support for a NumPy function that have already been done if guidance is needed... here's a couple:

* `np.roll`: https://github.com/numba/numba/pull/3563/files
* `np.roots`: https://github.com/numba/numba/pull/2135/files

sharthZ23:
@sleighsoft Have you made any progress?

@stuartarchibald numba already has a partial sorting algorithm - [(L1212).](https://github.com/numba/numba/blob/ec6fa07c12c3703a52c2b4ac8a225a68539c3cc8/numba/np/arraymath.py#L1212) Do I need to try to extend it to the case of argpartition, or can I duplicate the functionality (`np.partition - _select_factory`, `np.argpartition - _select_factory2`)?

sleighsoft:
I haven't had the time to pursue this any further

rragundez:
Close by #8732

stuartarchibald:
> Closed by #8732

Thanks for following up on this @rragundez. This was indeed implemented in #8732 with thanks to @rragundez. Closing as fixed, the functionality will be in the next version of Numba (0.57.x).

PR #8732, opened 2023-01-28
Add numpy argpartition function support

rragundez:
Add numpy argpartition function support. 

I used some of the implementation of `partition`. Something to note it that I modified the already existing `_partition_factory` to add an optional argument if the user would need `np.argpartition` instead of `np.partition`, as you will see in the code it doesn't look the most pretty, but it was either that or write a `_argpartition_factory` function which has exactly the same logic it just swaps index elements at the same time it swaps the array with the values being sorted.

Let me know if you prefer to leave it as is or go the duplication route. I think this is the best compromise to not repeat code and have a single source of truth of partition logic.

Solves #2445
#4074 would need to be updated.

cheers

### DIFF
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -1474,8 +1474,9 @@ def nan_aware_less_than(a, b):
 
 
 def _partition_factory(pivotimpl):
-    def _partition(A, low, high):
+    def _partition(A, low, high, I=np.arange(0)):
         mid = (low + high) >> 1
+        argpartition = True if len(I) else False
         # NOTE: the pattern of swaps below for the pivot choice and the
         # partitioning gives good results (i.e. regular O(n log n))
         # on sorted, reverse-sorted, and uniform arrays.  Subtle changes
@@ -1484,13 +1485,21 @@ def _partition_factory(pivotimpl):
         # Use median of three {low, middle, high} as the pivot
         if pivotimpl(A[mid], A[low]):
             A[low], A[mid] = A[mid], A[low]
+            if argpartition:
+                I[low], I[mid] = I[mid], I[low]
         if pivotimpl(A[high], A[mid]):
             A[high], A[mid] = A[mid], A[high]
+            if argpartition:
+                I[high], I[mid] = I[mid], I[high]
         if pivotimpl(A[mid], A[low]):
             A[low], A[mid] = A[mid], A[low]
+            if argpartition:
+                I[low], I[mid] = I[mid], I[low]
         pivot = A[mid]
 
         A[high], A[mid] = A[mid], A[high]
+        if argpartition:
+            I[high], I[mid] = I[mid], I[high]
         i = low
         j = high - 1
         while True:
@@ -1501,11 +1510,15 @@ def _partition_factory(pivotimpl):
             if i >= j:
                 break
             A[i], A[j] = A[j], A[i]
+            if argpartition:
+                I[i], I[j] = I[j], I[i]
             i += 1
             j -= 1
         # Put the pivot back in its final place (all items before `i`
         # are smaller than the pivot, all items at/after `i` are larger)
         A[i], A[high] = A[high], A[i]
+        if argpartition:
+            I[i], I[high] = I[high], I[i]
         return i
     return _partition
 
@@ -1515,18 +1528,18 @@ _partition_w_nan = register_jitable(_partition_factory(nan_aware_less_than))
 
 
 def _select_factory(partitionimpl):
-    def _select(arry, k, low, high):
+    def _select(arry, k, low, high, idx=np.arange(0)):
         """
         Select the k'th smallest element in array[low:high + 1].
         """
-        i = partitionimpl(arry, low, high)
+        i = partitionimpl(arry, low, high, idx)
         while i != k:
             if i < k:
                 low = i + 1
-                i = partitionimpl(arry, low, high)
+                i = partitionimpl(arry, low, high, idx)
             else:
                 high = i - 1
-                i = partitionimpl(arry, low, high)
+                i = partitionimpl(arry, low, high, idx)
         return arry[k]
     return _select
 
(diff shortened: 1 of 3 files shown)


kc611 (approved):
Hi @rragundez , Thank you for the PR :smile: . Your contributions to Numba are highly appreciated. 

I gave the PR an initial review and it mostly LGTM!

A few points to note:
1. The logic does seem a bit off within the `_partition_factory`/`_partition`, but as you mentioned and from a maintainer's point of view, it's less code to manage and reuses existing logic so it does makes sense to keep it this way.

2. The `np.argpartition`, based on the existing `np.partition` logic does seem to deviate from NumPy behaviour, they are implemented using different algorithms (see https://github.com/numba/numba/pull/3320#issuecomment-422747016 for original `np.partition` implementation discussion) and whilst this might seem a bit implicit, since the ordering of the elements in the two partitions is supposed to be undefined, but it might be a good idea to mention the behaviour in the docs that the outputs may deviate from NumPy.

numba/np/arraymath.py
```diff
@@ -1474,8 +1474,9 @@ def nan_aware_less_than(a, b):
 
 
 def _partition_factory(pivotimpl):
-    def _partition(A, low, high):
+    def _partition(A, low, high, I=np.arange(0)):
```
kc611, line 1477:
```suggestion
    def _partition(A, low, high, I=None):
```

numba/np/arraymath.py
```diff
@@ -1474,8 +1474,9 @@ def nan_aware_less_than(a, b):
...
         mid = (low + high) >> 1
+        argpartition = True if len(I) else False
```
kc611, line 1479:
```suggestion
        argpartition = True if I is None else False
        if argpartition:
            assert len(I) == len(A) # The corresponding index array should be valid for the array provided.
```

Just to make sure every upstream function uses the factory correctly.

numba/np/arraymath.py
```diff
@@ -1515,18 +1528,18 @@ def _partition(A, low, high):
 
 
 def _select_factory(partitionimpl):
-    def _select(arry, k, low, high):
+    def _select(arry, k, low, high, idx=np.arange(0)):
```
kc611, line 1531:
```suggestion
    def _select(arry, k, low, high, idx=None):
```

kc611 (changes requested):
A small suggestion regarding default function arguments. 

On this note is there any particular reason you went with the default argument for `idx` being `np.arange(0)` ?

rragundez:
@kc611  I also do not like it at all, but since the function is passed to `register_jitable` it needs to make sense of the types of the variables and if left with default as  `None` it cannot make sense of the `None` type and the array of type `int64` which `I` or `idx` can become. So I decided to use the `np.arange(0)` as default value which is an empty array of type `int64`. 
Does that make sense? Let me know what you think.

kc611:
Hi @rragundez, apologies for the delay. Following a small internal discussion, it'd be better for us to go with `None` as default argument. The reason being having a mutable default arguments for Python functions is generally not recommended due to the kind of unexpected behaviours it leads to down the line.

Additionally, the type inference within Numba is expected to figure out if the argument type is `None` and prune the branches accordingly during compile time.

rragundez:
@kc611 Yeap, agree with the Python anti-pattern. I will make the changes and let's see what the the CI says, hopefully I made come mistake when testing and it will be able to resolve the `None` type.

rragundez:
@kc611  done, please review. Thanks.

rragundez:
@kc611 By the way I just remembered that indeed is a bad practice to set default parameters that are mutable but even though numpy arrays are mutable, an array of size 0 is immutable because it has no elements to change, and numpy arrays are of fixed size, therefore the default value of an int array of size zero is immutable. Let me know if this makes sense and how you would like me to proceed.

numba/np/arraymath.py
```diff
@@ -1474,8 +1474,9 @@ def nan_aware_less_than(a, b):
 
 
 def _partition_factory(pivotimpl):
```
kc611, line 1476:
```suggestion
def _partition_factory(pivotimpl, argpartition=False):
```

numba/np/arraymath.py
```diff
@@ -1474,8 +1474,9 @@ def nan_aware_less_than(a, b):
...
-    def _partition(A, low, high):
+    def _partition(A, low, high, I=None):
         mid = (low + high) >> 1
+        argpartition = False if I is None else True
```
kc611, line 1479:
```suggestion
```

numba/np/arraymath.py
```diff
@@ -1515,18 +1528,18 @@ def _partition(A, low, high):
```
kc611, line 1528:
```suggestion
_argpartition_w_nan = register_jitable(_partition_factory(nan_aware_less_than, argpartition=True))
```

numba/np/arraymath.py
```diff
@@ -1515,18 +1528,18 @@ def _partition(A, low, high):
...
             if i < k:
                 low = i + 1
-                i = partitionimpl(arry, low, high)
+                i = partitionimpl(arry, low, high, idx)
             else:
                 high = i - 1
-                i = partitionimpl(arry, low, high)
+                i = partitionimpl(arry, low, high, idx)
         return arry[k]
     return _select
```
kc611, line 1545:
```suggestion

_arg_select_w_nan = register_jitable(_select_factory(_argpartition_w_nan))
```

numba/np/arraymath.py
```diff
@@ -1808,6 +1821,28 @@ def np_partition_impl_inner(a, kth_array):
...
+
+    idx = np.ndindex(a.shape[:-1])  # Numpy default partition axis is -1
+    for s in idx:
+        arry = a[s].copy()
+        idx_arry = np.arange(len(arry))
+        low = 0
+        high = len(arry) - 1
+
+        for kth in kth_array:
+            _select_w_nan(arry, kth, low, high, idx_arry)
```
kc611, line 1839:
```suggestion
            _arg_select_w_nan(arry, kth, low, high, idx_arry)
```

kc611 (changes requested):
Hi @rragundez , Thank you for making the changes. :smiley: . I think the reason for failure is that Numba isn't able to infer `I` as `None` during compile time and ends up trying to lower the branches with `argpartition` boolean. A way around this would be to make `argpartition` a compile time constant instead of runtime, this would make sure branch pruner absolutely removes the branches with the given boolean so that they aren't lowered into IR. 

I agree that an array of size 0 is immutable because it has no elements to change, but we can actually change it's attributes making it a 'mutable as a structure'. :slightly_smiling_face:  However, the issue at hand is that this practice is generally not recommended and even more so in Numba because it might lead to weird and unexpected IR issues down the line, even if it seems to work just fine right now.

rragundez:
@kc611 thanks for the feedback. I made the requested changes and the CI is passing now. Please review.

kc611 (approved):
Thanks @rragundez, these changes LGTM!

rragundez:
@kc611  Thanks for the approval. I cannot merge though, still says "Merging is blocked Merging can be performed automatically with 1 approving review."

rragundez:
Hi @kc611, can you help me understand if this PR will be able to be merged? Thanks.

PR #8736, opened 2023-02-03
Implement np.lib.stride_tricks.sliding_window_view

### LINKED ISSUE #7674: FR: Support `numpy.lib.stride_tricks.sliding_window_view`
Hi.
Please add support for numpy's `sliding_window_view` ([docs](https://numpy.org/doc/stable/reference/generated/numpy.lib.stride_tricks.sliding_window_view.html#numpy.lib.stride_tricks.sliding_window_view)). It is essentially a safe and useful wrapper to `.as_strided()` that you already support so that shouldn't be much of an issue. 

Thanks!

bcbnz:
This pull request adds an overload for [numpy.lib.stride_tricks.sliding_window_view](https://numpy.org/doc/stable/reference/generated/numpy.lib.stride_tricks.sliding_window_view.html) ([current source](https://github.com/numpy/numpy/blob/9db4a50c7310744f00ae83f93ac936047d9c9165/numpy/lib/stride_tricks.py#L124)) and in doing so closes #7674.

The supported function arguments are:

* `window_size`: must be integer or tuple of integers (same as listed in NumPy docstring).
* `axis`: must be None (default, meaning all axes), integer or tuple of integers (same as listed in NumPy docstring).
* `subok`: not supported.
* `writeable`: not supported; returned views are always writeable.

Two sets of unit tests added: one to compare output with the NumPy version (only runs with NumPy >= 1.20, the release the function was added), and one to check that exceptions are raised for invalid inputs (always runs as it does not need to run the NumPy version).

I have two outstanding questions:

* Since `reshape_unchecked` only accepts tuples for ``shape`` and ``strides``, I used a pattern I found elsewhere in the code to create tuples: at compile time generate tuples of the appropriate length (`x.ndim + len(window_shape)`) and then use `tuple_setitem` to modify them at runtime. Is this acceptable / is there a better way of doing this?

* I have replicated the exception messages from the NumPy version (with the removal of the string interpolation). These messages are not consistent, e.g., one puts argument names in backticks, others don't. Is it preferred to replicate exception messages where possible, or should I reword them to be consistent with each other?

### DIFF
--- a/numba/tests/test_array_manipulation.py
+++ b/numba/tests/test_array_manipulation.py
@@ -558,6 +558,7 @@ class TestArrayManipulation(MemoryLeakMixin, TestCase):
         # Repeated axis.
         check(arr2, (5, 3, 3), (0, 1, 0))
 
+    @unittest.skipIf(numpy_version < (1, 20), "requires NumPy 1.20 or newer")
     def test_sliding_window_view_errors(self):
         cfunc = jit(nopython=True)(sliding_window_view)
 
--- a/numba/tests/test_array_manipulation.py
+++ b/numba/tests/test_array_manipulation.py
@@ -147,6 +147,10 @@ def as_strided2(a):
     return np.lib.stride_tricks.as_strided(a, shape=shape, strides=strides)
 
 
+def sliding_window_view(x, window_shape, axis=None):
+    return np.lib.stride_tricks.sliding_window_view(x, window_shape, axis=axis)
+
+
 def bad_index(arr, arr2d):
     x = arr.x,
     y = arr.y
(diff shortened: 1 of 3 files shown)


bcbnz:
Of course, the code is not valid as is on NumPy < 1.20 since there is no function to overload. Not sure how that didn't cross my mind when I added the version check to one of the tests...

I guess this means the entire implementation should be inside a version-checked block. Or is there some other way of dealing with it?

kc611:
Hi @bcbnz, thank you for your contributions to Numba.  

As for your question, since the function is a new addition to NumPy, it makes sense for it to be implemented within a version checked block. Similar to how we have below:

https://github.com/numba/numba/blob/28573bb985cf4816e1c6f54ce2f4bd1e44617bcd/numba/np/random/distributions.py#L53

I'd also recommend documenting this behaviour within Numba docs just to be sure there isn't any confusion regarding it's support for Numba users.

bcbnz:
Thanks for the confirmation @kc611. I have added a version check around the overload and the tests, and have added a note to the docs stating NumPy 1.20 or later is required for this function.

numba/tests/test_array_manipulation.py
```diff
@@ -516,6 +520,99 @@ def test_as_strided(self):
         self.check_as_strided(as_strided1)
         self.check_as_strided(as_strided2)
 
+    @unittest.skipIf(numpy_version < (1, 20), "requires NumPy 1.20 or newer")
+    def test_sliding_window_view(self):
+        cfunc = jit(nopython=True)(sliding_window_view)
```
kc611, line 525: Could maybe use new `njit` decorator upon the original function itself and using thhe `.py_func` to get the python version of the function ?

numba/tests/test_array_manipulation.py
```diff
@@ -516,6 +520,99 @@ def test_as_strided(self):
...
+            self.assertIn(msg, str(raises.exception))
+
+        # Exceptions leak references
+        self.disable_leak_check()
+
+        arr1 = np.arange(24)
+        arr2 = np.arange(200).reshape(10, 20)
+
+        # Window shape validity.
+        _raises("window_shape cannot be larger", arr1, 25, None)
```
kc611, line 582: These could be seperate subtests just so that if anything fails we know what exactly breaks on first glance. (See `TestCase.subTest`).

kc611 (commented):
The PR itself LGTM!

A few minor nitpicks below regarding testing and it'll be good to go.

bcbnz:
Thanks for the review @kc611 -- hopefully the new commit I've made takes care of the tests.

kc611 (approved):

sklam (approved):

sklam:
This was merged as part of 0.57 and should not be part of 0.58 milestone

PR #8744, opened 2023-02-08
Update CtypesLinker::add_cu error message to include fp16 usage

testhound:
This pull request update the error message on the add_cu method in CtypesLinker to include fp16. The reason for the update is CUDA fp16 support requires linking of cu files in it's internal implementation.


testhound:
gpuci run tests

testhound:
gpuci run tests

gmarkall:
gpuci run tests

numba/cuda/cudadrv/driver.py
```diff
@@ -2721,8 +2721,11 @@ def add_file(self, path, kind):
             raise LinkerError(msg)
 
     def add_cu(self, path, name):
-        raise NotImplementedError("Linking CUDA source files is not supported "
-                                  "with the ctypes binding. ")
+        raise NotImplementedError("Support for fp16 or linking CUDA source "
+                                  "files requires the use of the Nvidia CUDA "
+                                  "bindings and enabling the "
+                                  "NUMBA_CUDA_USE_NVIDIA_BINDING environment "
+                                  "variable. See documentation for me details.")
```
gmarkall, line 2728: Suggested changes:

- `fp16` should be `float16` -I noticed that the code in Numba is unusual in Python referring to 16-bit floating point values as `fp16` rather than `float16` which seems to be used everywhere else - for example, the NumPy dtype. This is an inconsistency I should probably have realised much sooner, so I'd like to try not introducing more instances of it.
- `Nvidia` is usually written `NVIDIA`
- Instead of saying the environment variable should be enabled, say that it should be set to `1`.
- `See documentation for me details` needs the grammar correcting, and it would be helpful to provide a link to the relevant documentation. Where is the user supposed to look?

gmarkall (changes requested):
Thanks for the PR. In addition to the changed test, I think it would be wise to have a test that causes the exception to be raised through the use of an unsupported operation on `float16` data.

A couple of other questions:

- Is it possible to identify what the user is doing (either using `float16` or linking in additional `.cu` files) and provide them an error message that references specifically what they're doing instead of providing a catch-all generic message?
- Is the documentation of this restriction sufficient, or are some additions needed in the docs?

testhound:
@gmarkall error message updated.

gmarkall:
Thanks. What do you think about the other two questions from the review?

testhound:
> Thanks. What do you think about the other two questions from the review?

@gmarkall I don't think it's possible to determine if this is a float 16 issue or the user is linking in .cu files; in both cases we end up in the CtypesLinker::add_cu which throws the exception. We would need to catch this higher up in the stack somewhere.

I think the clarified error message along with the link to the documentation is sufficient.

gmarkall:
>  I don't think it's possible to determine if this is a float 16 issue or the user is linking in .cu files; in both cases we end up in the CtypesLinker::add_cu which throws the exception. We would need to catch this higher up in the stack somewhere.

Is there anything wrong with this approach:

```diff
diff --git a/numba/cuda/dispatcher.py b/numba/cuda/dispatcher.py
index a79120d85..bb85093f9 100644
--- a/numba/cuda/dispatcher.py
+++ b/numba/cuda/dispatcher.py
@@ -110,6 +110,9 @@ class _Kernel(serialize.ReduceMixin):
                if (f'__numba_wrapper_{fn}' in lib.get_asm_str())]
 
         if res:
+            if not config.CUDA_USE_NVIDIA_BINDING:
+                raise NotImplementedError('float16 needs NV Binding')
+
             # Path to the source containing the foreign function
 
             basedir = os.path.dirname(os.path.abspath(__file__))
```

which causes the example

```python
from numba import cuda, types

@cuda.jit('void()')
def f():
    types.float16(1) / types.float16(2)
```

to emit:

```
  File "/home/gmarkall/numbadev/numba/numba/cuda/dispatcher.py", line 114, in __init__
    raise NotImplementedError('float16 needs NV Binding')
NotImplementedError: float16 needs NV Binding
```

?

testhound:
@gmarkall I was thinking about that after I mentioned "higher up in the stack", and yes that works as well. I'll make that change.

testhound:
gpuci run tests

testhound:
@gmarkall error message updated.

numba/cuda/dispatcher.py
```diff
@@ -110,6 +110,15 @@ def __init__(self, py_func, argtypes, link=None, debug=False,
                if (f'__numba_wrapper_{fn}' in lib.get_asm_str())]
 
         if res:
+            if not config.CUDA_USE_NVIDIA_BINDING:
+                s = "https://numba.readthedocs.io/en/stable/cuda/bindings.html"
+                msg = ("Use of float16 requires the use of the NVIDIA CUDA "
+                       "bindings and settng the "
```
gmarkall, line 116:
```suggestion
                       "bindings and setting the "
```

numba/cuda/tests/cudadrv/test_linker.py
```diff
@@ -191,8 +191,8 @@ def kernel(x):
 
     @skip_with_cuda_python
     def test_linking_cu_ctypes_unsupported(self):
-        msg = ('Linking CUDA source files is not supported with the ctypes '
-               'binding')
+        msg = ('Support for fp16 or linking CUDA source files '
```
gmarkall, line 194: The expected message doesn't seem to match the message added in `dispatcher.py`. Shouldn't this test be failing?

gmarkall:
@testhound Some questions on the diff.

numba/cuda/tests/cudadrv/test_linker.py
testhound: @gmarkall the test was not failing it appears because it was not being run due to the skip_with_cuda_python decorator.

testhound:
gpuci run tests

testhound:
gpuci run tests

testhound:
@gmarkall test case updates made to address your comments.

gmarkall (changes requested):
Thanks for the changes. Whilst running `test_linking_cu_ctypes_unsupported` in a subprocess is a positive change (so it's good to keep it in), my point was that it doesn't test the changes you're making in this PR, which raises when the automatic linking of a `.cu`file is induced by the use of a `float16` function that requires it - so, another test needs to be added for the new exception that's added here.

There are some other suggestions on the diff relating to spelling errors and simplification.

testhound:
gpuci run tests

testhound:
@gmarkall ready for review.

gmarkall (changes requested):
Thanks for the updates and adding the additional test. The additional test seems more complicated than necessary - since we're just checking for something at compilation time, we needn't create arguments and launch a kernel just to test it. For example, instead of its present form it could look like:

```python
    @TestCase.run_test_in_subprocess(envvars=_NUMBA_NVIDIA_BINDING_0_ENV)
    def test_linking_float16_unsupported(self):
        msg = ("Use of float16 requires the use of the NVIDIA CUDA "
               "bindings and settng the ")
        with self.assertRaisesRegex(NotImplementedError, msg):
            @cuda.jit('void(float16[::1])')
            def hexp10_vectors(x):
                cuda.fp16.hexp10(x[0])
```

(note that that example is still checking for an error message that needs fixing). There are some other issues to look at:

- There is still an unresolved comment on the diff: https://github.com/numba/numba/pull/8744/files#r1109694856
- CI is failing with CUDA 11.8 - why is this?
- There are some additional comments added on the diff.

testhound:
gpuci run tests

testhound:
gpuci run tests

gmarkall (dismissed):
Thanks for the updates - I'm approving this, but it needs conflicts resolving before it's mergeable.

The problem with the `test_linking_cu_types_unsupported` test under MVC is actually a bug (https://github.com/numba/numba/issues/8801) so once that's fixed this test can be converted to run under normal circumstances.

gmarkall:
This should also have a buildfarm run once conflicts are resolved, so I've added "Waiting for CI" too.

testhound:
gpuci run tests

testhound:
@gmarkall merge conflict fixed and ready for review.

gmarkall:
@stuartarchibald Could this have a CUDA BF run please?

gmarkall (approved):
This looks good - just needs a BF run to move to RTM.

sklam:
BFID `numba_smoketest_cuda_yaml_192`

sklam:
Buildfarm failed with missing `jitlink.cu`

```
FAIL: test_linking_cu_ctypes_unsupported (numba.cuda.tests.cudadrv.test_linker.TestLinker)
File "/opt/conda/envs/testenv_3e4abae3/lib/python3.9/site-packages/numba/tests/support.py", line 644, in inner
self.subprocess_test_runner(test_module=self.__module__,
File "/opt/conda/envs/testenv_3e4abae3/lib/python3.9/site-packages/numba/tests/support.py", line 629, in subprocess_test_runner
AssertionError: 1 != 0 :
captured stdout:
captured stderr: E
ERROR: test_linking_cu_ctypes_unsupported (numba.cuda.tests.cudadrv.test_linker.TestLinker)
File "/opt/conda/envs/testenv_3e4abae3/lib/python3.9/site-packages/numba/tests/support.py", line 652, in inner
func(self)
File "/opt/conda/envs/testenv_3e4abae3/lib/python3.9/site-packages/numba/cuda/tests/cudadrv/test_linker.py", line 223, in test_linking_cu_ctypes_unsupported
def f():
File "/opt/conda/envs/testenv_3e4abae3/lib/python3.9/site-packages/numba/cuda/decorators.py", line 133, in _jit
disp.compile(argtypes)
File "/opt/conda/envs/testenv_3e4abae3/lib/python3.9/site-packages/numba/cuda/dispatcher.py", line 929, in compile
kernel.bind()
File "/opt/conda/envs/testenv_3e4abae3/lib/python3.9/site-packages/numba/cuda/dispatcher.py", line 208, in bind
File "/opt/conda/envs/testenv_3e4abae3/lib/python3.9/site-packages/numba/cuda/codegen.py", line 179, in get_cufunc
File "/opt/conda/envs/testenv_3e4abae3/lib/python3.9/site-packages/numba/cuda/codegen.py", line 158, in get_cubin
linker.add_file_guess_ext(path)
File "/opt/conda/envs/testenv_3e4abae3/lib/python3.9/site-packages/numba/cuda/cudadrv/driver.py", line 2652, in add_file_guess_ext
self.add_cu_file(path)
File "/opt/conda/envs/testenv_3e4abae3/lib/python3.9/site-packages/numba/cuda/cudadrv/driver.py", line 2642, in add_cu_file
with open(path, 'rb') as f:
FileNotFoundError: [Errno 2] No such file or directory: 'numba/cuda/tests/data/jitlink.cu'
```

testhound:
> Buildfarm failed with missing `jitlink.cu`
> 
> ```
> FAIL: test_linking_cu_ctypes_unsupported (numba.cuda.tests.cudadrv.test_linker.TestLinker)
> File "/opt/conda/envs/testenv_3e4abae3/lib/python3.9/site-packages/numba/tests/support.py", line 644, in inner
> self.subprocess_test_runner(test_module=self.__module__,
> File "/opt/conda/envs/testenv_3e4abae3/lib/python3.9/site-packages/numba/tests/support.py", line 629, in subprocess_test_runner
> self.assertEqual(status.returncode, 0, streams)
> AssertionError: 1 != 0 :
> captured stdout:
> captured stderr: E
> ERROR: test_linking_cu_ctypes_unsupported (numba.cuda.tests.cudadrv.test_linker.TestLinker)
> File "/opt/conda/envs/testenv_3e4abae3/lib/python3.9/site-packages/numba/tests/support.py", line 652, in inner
> func(self)
> File "/opt/conda/envs/testenv_3e4abae3/lib/python3.9/site-packages/numba/cuda/tests/cudadrv/test_linker.py", line 223, in test_linking_cu_ctypes_unsupported
> def f():
> File "/opt/conda/envs/testenv_3e4abae3/lib/python3.9/site-packages/numba/cuda/decorators.py", line 133, in _jit
> disp.compile(argtypes)
> File "/opt/conda/envs/testenv_3e4abae3/lib/python3.9/site-packages/numba/cuda/dispatcher.py", line 929, in compile
> kernel.bind()
> File "/opt/conda/envs/testenv_3e4abae3/lib/python3.9/site-packages/numba/cuda/dispatcher.py", line 208, in bind
> self._codelibrary.get_cufunc()
> File "/opt/conda/envs/testenv_3e4abae3/lib/python3.9/site-packages/numba/cuda/codegen.py", line 179, in get_cufunc
> cubin = self.get_cubin(cc=device.compute_capability)
> File "/opt/conda/envs/testenv_3e4abae3/lib/python3.9/site-packages/numba/cuda/codegen.py", line 158, in get_cubin
> linker.add_file_guess_ext(path)
> File "/opt/conda/envs/testenv_3e4abae3/lib/python3.9/site-packages/numba/cuda/cudadrv/driver.py", line 2652, in add_file_guess_ext
> self.add_cu_file(path)
> File "/opt/conda/envs/testenv_3e4abae3/lib/python3.9/site-packages/numba/cuda/cudadrv/driver.py", line 2642, in add_cu_file
> with open(path, 'rb') as f:
> FileNotFoundError: [Errno 2] No such file or directory: 'numba/cuda/tests/data/jitlink.cu'
> ```

@sklam this error confuses me as I explicitly fixed the test case to refer to the new location of jitlink.cu and this passes all CI tests which refers to this location.

sklam:
new BFID: `numba_smoketest_cuda_yaml_193`

sklam:
> new BFID: `numba_smoketest_cuda_yaml_193`

passed

testhound:
> > @sklam this error confuses me as I explicitly fixed the test case to refer to the new location of jitlink.cu and this passes all CI tests which refers to this location.
> 
> The test contained a hardcoded expectation of the working directory, which doesn't hold in all cases (in the buildfarm in this instance). Test data locations should be determined using `test_data_dir`. As we're close to the wire for 0.57 I've pushed a fix for this branch.

@gmarkall thank you for the fix.

gmarkall (approved):
I think this is good to merge.

Issue #8745, opened 2023-02-09
Failure in test_dispatcher due change in representation of types


testhound:
The CUDA test case cudapy/test_dispatcher.py::test_explicit_signatures_ambiguous_resolution is failing due to a change that implemented `__repr__` for numba types. The issue appears to be related to this pull request:

https://github.com/numba/numba/pull/8685

PR #8746, opened 2023-02-09
Fix failing test_dispatcher test case

testhound:
This pull request fixes a failing test case in test_dispatch due to a change in the way cuda types are represented. The corresponding issue is here: https://github.com/numba/numba/issues/8745


testhound:
gpuci run tests

gmarkall (approved):
Many thanks for spotting and fixing this!

PR #8748, opened 2023-02-09
Suppress known test failures for py3.11

sklam:
- add expectedFailure to try-except and with-lifting tests
- reduce test accuracy for test_pow

### DIFF
--- a/numba/tests/test_operators.py
+++ b/numba/tests/test_operators.py
@@ -1311,8 +1311,9 @@ class TestMixedInts(TestCase):
         self.run_arith_binop(self.op.mod_usecase, 'mod', samples)
 
     def test_pow(self):
+        extra_cast = {}
         if utils.PYVERSION == (3, 11):
-            extra_cast = dict(force_type=float)
+            extra_cast["force_type"] = float
         pyfunc = self.op.pow_usecase
         # Only test with positive values, as otherwise trying to write the
         # control function in terms of Python or Numpy power turns out insane.
--- a/numba/tests/test_operators.py
+++ b/numba/tests/test_operators.py
@@ -1248,7 +1248,8 @@ class TestMixedInts(TestCase):
         return control_unsigned
 
     def run_binary(self, pyfunc, control_func, operands, types,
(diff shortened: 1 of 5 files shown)


stuartarchibald:
Closed and reopened to get a rebuild against `py3.11` branch with #8752 now present.

stuartarchibald:
This patch reduces the reported Python 3.11 related test failures to the following:
* `test_no_accidental_warnings (numba.tests.test_import.TestNumbaImport.test_no_accidental_warnings` which complains of `DeprecationWarning: path is deprecated. Use files() instead.` which is from `importlib`. xref fix in: https://github.com/numba/llvmlite/pull/906 .
* `FAIL: test_profiler (numba.tests.test_profiler.TestProfiler.test_profiler)` as #8711(or a variant of) is needed.

numba/tests/test_operators.py
```diff
@@ -1304,11 +1311,13 @@ def test_mod(self):
         self.run_arith_binop(self.op.mod_usecase, 'mod', samples)
 
     def test_pow(self):
+        if utils.PYVERSION == (3, 11):
+            extra_cast = dict(force_type=float)
```
stuartarchibald, line 1315: Is an `else` branch needed to define `extra_cast` else this will fail on < 3.11 with a `NameError`?

Another option to "fix" this would be to just restrict the input on 3.11 to numerically smaller values. As noted in the comments, there are already challenges present in the inputs/testing/implementation.

sklam: oops, i'll fix that. and check if smaller numbers make a easier fix.

sklam:
> * `FAIL: test_profiler (numba.tests.test_profiler.TestProfiler.test_profiler)` as [[CI only] testing patch for #8670 #8711](https://github.com/numba/numba/pull/8711)(or a variant of) is needed.

I still need to suppress this

stuartarchibald (approved):

sklam:
https://github.com/numba/numba/pull/8761 passing shows that this PR work

PR #8751, opened 2023-02-10
Recycle test runners more aggressively

apmasell:
This changes the multiprocess pool to use a single long-lived pool with short-lived workers that are automatically restarted by the pool rather than operating in batches. This is to reduce idle time between tests.

### DIFF
--- a/numba/testing/main.py
+++ b/numba/testing/main.py
@@ -770,28 +770,24 @@ class ParallelTestRunner(runner.TextTestRunner):
 
         # Split the tests and recycle the worker process to tame memory usage.
         chunk_size = 100
-        splitted_tests = [self._ptests[i:i + chunk_size]
-                          for i in range(0, len(self._ptests), chunk_size)]
 
-        for tests in splitted_tests:
-            pool = multiprocessing.Pool(self.nprocs)
-            try:
-                self._run_parallel_tests(result, pool, child_runner, tests)
-            except:
-                # On exception, kill still active workers immediately
+        pool = multiprocessing.Pool(self.nprocs, maxtasksperchild=chunk_size)
+        try:
+            self._run_parallel_tests(result, pool, child_runner, self._ptests)
+        except:
+            # On exception, kill still active workers immediately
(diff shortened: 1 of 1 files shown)


gmarkall:
gpuci run tests

gmarkall (commented):
I tried this with the CUDA testsuite (`python runtests.py numba.cuda.tests -m`) and it didn't speed things up, but it didn't slow them down either. So it's OK from the CUDA side of things.

apmasell:
Okay. I assume the CUDA suite is smaller and bottlenecks on different resources, so I guess that isn't surprising.

gmarkall:
I had hoped it would at least shave a couple of seconds off, because there are some tests that just compile a lot of things in parallel without touching the GPU. Never mind :-)

sklam:
I also didn't observe much speed difference, but the simplification of the code is nice.

sklam:
I tested this again on a 160 core power8 machine and the result is so much more interesting. The command that i used: `time python runtests.py -j0:10 -m -vb`

On main [c12bbf4](https://github.com/numba/numba/commit/c12bbf4bb0bb383b576a0e356a2e23afe476aa8e):
```
Ran 1037 tests in 295.618s

OK (skipped=85, expected failures=3)

real	5m0.635s
user	25m58.936s
sys	1m6.356s
```

On this PR branch:

```
Ran 1037 tests in 220.317s

OK (skipped=85, expected failures=3)

real	3m43.361s
user	44m3.620s
sys	0m48.908s
```

There's a 1.3x improvement. Reduction in `sys` time is good indication of less syscall overhead. Not sure how to interpret the `user` time though.

stuartarchibald:
> Not sure how to interpret the user time though.

I'm not sure how to interpret this either. The tests that ran in `-j 0:10` include some of stress tests for the threading backends which are deliberately left to do unpredictable things, could that be involved? Maybe a rerun would yield a different result?

Either way, the code simplification is worthwhile.

sklam (approved):

PR #8757, opened 2023-02-15
Fix refcount issue with reflected list pop/del.

### LINKED ISSUE #8753: Memory overwriting other memory issue?
## Reporting a bug

Take this piece of code:
```
import numba

@numba.njit
def foo():
  lst1 = [[1,2,3,4,5], [6,7,8,9,10]]

  for i in range(300):
    lst2 = []
    lst2.append(1)

    del lst1[1]
    del lst1[0]

    lst1.append([1,2,3,4,5])
    lst1.append([6,7,8,9,10])

    if len(lst2) > 1:
      print("This should never print!")
      print("Length of lst2(should not ever be greater than 1): ", len(lst2))

for i in range(200):
  foo()
```
If you run this, you will get that sometimes the prints happen even though they shouldn't. 'lst2' should never have length more than 1 since it only gets appended to once and then reset in each iteration. It seems that appending to lst1 is somehow "interfering" with lst2. My suspicion is some kind of memory problem where 'lst1' appending is overwriting other memory? But I am not sure.. When I run this, the length of lst2 somehow ends up being 5 sometimes which is super weird.. Thank you for the help!

stuartarchibald:
This adds a missing incref for the `pop()`'d item of a reflected list of reference counted objects. It impacts `__delitem__` too as that is simply implemented as a call to `pop(<index to delete>)`.

Fixes #8753

### DIFF
--- a/numba/cpython/listobj.py
+++ b/numba/cpython/listobj.py
@@ -1010,7 +1010,7 @@ def list_pop(context, builder, sig, args):
     inst.guard_index(idx, "pop index out of range")
 
     res = inst.getitem(idx)
-
+    inst.incref_value(res)
     one = ir.Constant(n.type, 1)
     n = builder.sub(n, ir.Constant(n.type, 1))
     inst.move(idx, builder.add(idx, one), builder.sub(n, idx))
--- a/numba/tests/test_lists.py
+++ b/numba/tests/test_lists.py
@@ -463,6 +463,19 @@ class TestLists(MemoryLeakMixin, TestCase):
             cfunc(1, 5)
         self.assertEqual(str(cm.exception), "pop index out of range")
 
+    def test_pop_refcounted(self):
+        # See issue #8753.
+
+        @njit
+        def foo():
+            lst = [[1,]]
+            x = [10,]
+            lst.pop(0)
+            lst.append(x)
+            return lst[0]
+
+        self.assertPreciseEqual(foo(), foo.py_func())
+
     def test_insert(self):
         pyfunc = list_insert
         cfunc = jit(nopython=True)(pyfunc)


stuartarchibald:
This didn't quite fix it but probably confirms the problem, the segfault/broken reference count issue went away, but it now leaks.

seberg:
Very randomly came here and thus pass-by comment.  I suspect "a" fix would be:
```diff
diff --git a/numba/cpython/listobj.py b/numba/cpython/listobj.py
index e28113e09..9843d175e 100644
--- a/numba/cpython/listobj.py
+++ b/numba/cpython/listobj.py
@@ -873,7 +873,7 @@ def list_append(context, builder, sig, args):
     n = inst.size
     new_size = builder.add(n, ir.Constant(n.type, 1))
     inst.resize(new_size)
-    inst.setitem(n, item, incref=True)
+    inst.setitem(n, item, incref=True, decref_old_value=False)
 
     return context.get_dummy_value()
 
@@ -1010,7 +1010,7 @@ def list_pop(context, builder, sig, args):
     inst.guard_index(idx, "pop index out of range")
 
     res = inst.getitem(idx)
-    inst.incref_value(res)
+
     one = ir.Constant(n.type, 1)
     n = builder.sub(n, ir.Constant(n.type, 1))
     inst.move(idx, builder.add(idx, one), builder.sub(n, idx))
```

An additional thing might be to make sure that `resize()` zero-fills when shrinking the size to ensure that the now unused data is NULLed (in case the list grows again later)?  (not sure if you use xdecref, otherwise it hardly matters, I guess.)

I suspect that `list_mul_inplace` would be in this boat and also requires the `decref_old_value=False`.

seberg:
Sorry, realized that the above is maybe not quite wrong, but not how things seem designed.  The normal pop does clearing of the now free location.

So I think the issues is rather that calling `move()` will not `zfill` the (newly) unused space.  That job could also fall to `resize()`, but it is probably cleaner to do in `move()` and likely affects all calls of move (with references).

stuartarchibald:
@seberg thanks for taking a look. The fix could well be to do with zeroing out memory for the unused space. The code change at present hints at the underlying segfault being due to inc/dec ref not behaving as expected, probably due to inconsistent state (it's the most common case!).

PR #8764, opened 2023-02-16
CUDA tidy-up: remove some unneeded methods

gmarkall:
- The `_Kernel.device` property is unused and just weird - it's not a property of the kernel at all, and just gave back the current device.
- The `CUDADispatcher.bind()` method was unused and un-needed. It was called needlessly in a couple of test methods - definitions owned by the dispatcher were already bound when they were created anyway.

Neither of these functions were documented or public APIs.

### DIFF
--- a/numba/cuda/dispatcher.py
+++ b/numba/cuda/dispatcher.py
@@ -199,13 +199,6 @@ class _Kernel(serialize.ReduceMixin):
         self._codelibrary.get_cufunc()
 
     @property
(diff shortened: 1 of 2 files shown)


gmarkall:
gpuci run tests

stuartarchibald (approved):
Thanks for the patch. I've taken a look at the proposed removals and agree with the assessment made in the patch commentary.

PR #8765, opened 2023-02-16
BLD: remove distutils

fangchenli:
distutils will be removed in Python 3.12. This PR removes all of its usage from setup.py.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,12 +1,11 @@
 import os
 import platform
 import sys
-from distutils import sysconfig
-from distutils.command import build
-from distutils.command.build_ext import build_ext
-from distutils.spawn import spawn
+from sysconfig import get_config_vars
 
 from setuptools import Extension, find_packages, setup
+from setuptools.command import build_ext
+
 import versioneer
 
 _version_module = None
@@ -28,7 +27,7 @@ max_llvmlite_version = "0.41"
 
 if sys.platform.startswith('linux'):
     # Patch for #2555 to make wheels without libpython
-    sysconfig.get_config_vars()['Py_ENABLE_SHARED'] = 0
+    get_config_vars()['Py_ENABLE_SHARED'] = 0
 
 
 def _guard_py_ver():
@@ -49,14 +48,6 @@ def _guard_py_ver():
 
 _guard_py_ver()
 
-
-class build_doc(build.build):
-    description = "build documentation"
-
-    def run(self):
-        spawn(['make', '-C', 'docs', 'html'])
-
-
 versioneer.VCS = 'git'
 versioneer.versionfile_source = 'numba/_version.py'
 versioneer.versionfile_build = 'numba/_version.py'
@@ -64,7 +55,6 @@ versioneer.tag_prefix = ''
 versioneer.parentdir_prefix = 'numba-'
 
 cmdclass = versioneer.get_cmdclass()
-cmdclass['build_doc'] = build_doc
 
 extra_link_args = []
 install_name_tool_fixer = []
@@ -128,7 +118,7 @@ def is_building():
                       'build_scripts', 'install', 'install_lib',
                       'install_headers', 'install_scripts', 'install_data',
                       'sdist', 'bdist', 'bdist_dumb', 'bdist_rpm',
-                      'bdist_wininst', 'check', 'build_doc', 'bdist_wheel',
+                      'bdist_wininst', 'check', 'bdist_wheel',
                       'bdist_egg', 'develop', 'easy_install', 'test']
     return any(bc in sys.argv[1:] for bc in build_commands)
 
--- a/setup.py
+++ b/setup.py
@@ -73,13 +73,34 @@ if sys.platform == 'darwin':
 if platform.machine() == 'ppc64le':
     extra_link_args += ['-pthread']
 
+numba_be_user_options = [
+    ('werror', None, 'Build extensions with -Werror'),
+    ('wall', None, 'Build extensions with -Wall'),
+    ('noopt', None, 'Build extensions without optimization'),
+]
 
 class NumbaBuildExt(build_ext):
+
+    user_options = build_ext.user_options + numba_be_user_options
+    boolean_options = build_ext.boolean_options + ['werror', 'wall', 'noopt']
+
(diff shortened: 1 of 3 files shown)


gmarkall:
Many thanks for the PR. Why does it remove the `build_doc` command?

gmarkall:
The issue that RTD seems to have also seems to crop up for me when building locally:

```
$ python setup.py develop
  File "/home/gmarkall/numbadev/numba/setup.py", line 75, in <module>
    class NumbaBuildExt(build_ext):
  File "/home/gmarkall/numbadev/numba/setup.py", line 77, in NumbaBuildExt
    user_options = build_ext.user_options + numba_be_user_options
AttributeError: module 'setuptools.command.build_ext' has no attribute 'user_options'
```

fangchenli:
> Many thanks for the PR. Why does it remove the `build_doc` command?

It's not used in the codebase. I could add it back if people still want it.

setup.py
```diff
@@ -73,40 +63,13 @@ def run(self):
...
-            if sys.platform == 'win32':
-                extra_compile_args.append('/Od')
-            else:
-                extra_compile_args.append('-O0')
-        if self.werror:
-            extra_compile_args.append('-Werror')
-        if self.wall:
-            extra_compile_args.append('-Wall')
+        optimization_level = "/Od" if sys.platform == 'win32' else "-O0"
+        extra_compile_args = ["-Werror", "-Wall", optimization_level]
```
gmarkall, line 70: I think we don't have these on in all cases and instead guard them behind the `-wall` and `-werror` flags because some users might have a newer C compiler than we use / test with, which triggers unexpected warnings and causes the build to fail.

gmarkall:
> It's not used in the codebase. I could add it back if people still want it.

Thanks for clarifying. I would imagine we can probably do without it, but I don't know if any downstream packagers are using it to build docs as part of their packaging setup (e.g. in Debian, Fedora, etc.). Another maintainer might be able to provide more clarity here.

gmarkall:
> > It's not used in the codebase. I could add it back if people still want it.
> 
> Thanks for clarifying. I would imagine we can probably do without it, but I don't know if any downstream packagers are using it to build docs as part of their packaging setup (e.g. in Debian, Fedora, etc.). Another maintainer might be able to provide more clarity here.

To follow on a little more thought here, I certainly think it would be easier to get over the line here if the PR just focused on removing distutils usage and not making other functional changes, then were followed up by any other cleanups / improvements (such as removing the doc build command).

fangchenli:
> > > It's not used in the codebase. I could add it back if people still want it.
> > 
> > 
> > Thanks for clarifying. I would imagine we can probably do without it, but I don't know if any downstream packagers are using it to build docs as part of their packaging setup (e.g. in Debian, Fedora, etc.). Another maintainer might be able to provide more clarity here.
> 
> To follow on a little more thought here, I certainly think it would be easier to get over the line here if the PR just focused on removing distutils usage and not making other functional changes, then were followed up by any other cleanups / improvements (such as removing the doc build command).

Agree, I've put it back.

setup.py
```diff
@@ -73,40 +73,15 @@ def run(self):
...
-        if self.noopt:
-            if sys.platform == 'win32':
-                extra_compile_args.append('/Od')
-            else:
-                extra_compile_args.append('-O0')
-        if self.werror:
-            extra_compile_args.append('-Werror')
-        if self.wall:
-            extra_compile_args.append('-Wall')
+        if sys.platform.startswith("linux") and platform.machine() == "x86_64":
```
gmarkall, line 80: I think this is not the behaviour we want - we don't want these flags to always be added on Linux / x86_64 when we build, only on Linux / x86_64 in CI. This still removes the option for the user to add these flags through options to `build_ext` (and additionally the `noopt` flag as well..

gmarkall:
In the present state, the `--noopt`, `--werror` and `--wall` flags are removed from `build_ext`, but they need to be retained. They are described in the documentation: https://numba.readthedocs.io/en/latest/developer/contributing.html#building-numba

gmarkall:
Many thanks for the updates. Will have another look at this.

setup.py
```diff
@@ -73,8 +73,6 @@ def run(self):
 if platform.machine() == 'ppc64le':
     extra_link_args += ['-pthread']
 
-build_ext = cmdclass.get('build_ext', build_ext)
```
gmarkall, line 76: This came from the vendored Versioneer's `cmdclass`, which I think raises two questions:

- The vendored Versioneer we have is still using `distutils`. What's the right course of action here? Should we update the vendored version, or try to un-vendor it?
- Is `setuptools.command.build_ext.build_ext` the same thing as returned by `cmdclass.get('build_ext', build_ext)` when `cmdclass` is Versioneer's `cmdclass`?

(I also think there's the question of whether continuing to use Versioneer is the right way to go, but I *definitely* don't want to muddy the waters with that right now)

gmarkall (changes requested):
Thanks for the updates - I've had a look and before proceeding I'd like to understand the Versioneer situation a bit better - see comments on the diff.

setup.py
fangchenli: I reverted it. And will address versioneer issue later.

gmarkall (approved):
Many thanks for the updates. I think these changes look good. Considerations:

- The `spawn` command was just a wrapper around `subprocess.Popen`, so replacing it with `run` makes sense and makes things simpler.
- The `build` command is just a subclass of `Command`, but adds a whole load of behaviour completely unnecessary for the docs build, so making `build_doc` a `Command` is an improvement.

Other notes:

- The vendored Versioneer still uses distutils, but we should consider what to do about that separately from this PR.
- #8476 handles the other (non-setup.py) uses of distutils.
- This should maybe have a buildfarm run just in case of any surprises that might show up (@esc / @stuartarchibald thoughts maybe?) - marking as "Waiting on CI" for now before we decide to do that / do that.

stuartarchibald:
Buildfarm ID: `numba_smoketest_cpu_yaml_166`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cpu_yaml_166`.

Passed.

PR #8771, opened 2023-02-20
Update vendored Versioneer from 0.14 to 0.28

gmarkall:
This is primarily done to remove its distutils dependency.

Per the upgrade notes (see https://github.com/python-versioneer/python-versioneer/blob/master/UPGRADING.md), the configuration of Versioneer is moved into `setup.cfg` from `setup.py`.

The other notes on upgrading between Versions 0.14 and 0.28 don't apply to our particular configuration as we're not presently using `pyproject.toml`.

`versioneer.py` is also added to the Flake8 ignore list, as its format no longer conforms to our rules.

Note that there are other options we could consider, including:

- Not vendoring Versioneer
- Not using Versioneer anymore

However, as I'm not aware of any issues with our current setup / approach, I think this PR is the simplest way forward with minimum effort to help get us away from having a distutils dependency.

Changes to `numba/_version.py` and `versioneer.py` were made by Versioneer, when I ran the `versioneer install` command as per [Updating Versioneer](https://github.com/python-versioneer/python-versioneer#updating-versioneer).

This also includes @oscargus's fix from #7765 that updates `inspector._get_commit()` for recent versions of Versioneer. Note that this PR updates to version 0.28, whereas #7765 updates to 0.22 (which I believe was current at the time).


gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

PR #8773, opened 2023-02-21
CUDA: Refactor handling of target options

gmarkall:
The broad aim of this PR is to handle target options in CUDA more closely to how they are handled by the CPU target, with the longer-term aim of fixing the inheritance of options (for example, so that a device function could inherit `lineinfo=True` if its caller had `lineinfo=True`), and also to minimise the differences between the way decorators work in the CPU and CUDA targets. This PR doesn't reach the point where the decorator implementations are quite ready to unify, but it does move in the right direction.

I checked the generated code from this branch compared with `main`, and no difference is made to code generation.

This work was originally done with a view towards integrating it into #8594 so that the inheritance of lineinfo could be rectified there, but it has turned out to be too large a chunk of work to dump on top of an already-reviewed PR.

Individual commit messages provide more details, and the patches should be reviewable individually (this may be an easier way of reading them than as a whole).

### DIFF
--- a/docs/source/cuda-reference/kernel.rst
+++ b/docs/source/cuda-reference/kernel.rst
@@ -9,6 +9,34 @@ be configured and launched:
 
 .. autofunction:: numba.cuda.jit
 
+Kernel launch parameter handling extensions
+-------------------------------------------
+
+Numba supports passing a limited set of types of objects (scalars, arrays, and
+tuples) to CUDA kernels. The set of supported types can be augmented by passing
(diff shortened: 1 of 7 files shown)


gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
No need to do this now.

PR #8775, opened 2023-02-22
Revert PR#8751 for buildfarm stability

sklam:
Revert "Merge pull request #8751 from apmasell/mulitprocess_recycling" as an attempt to establish the baseline.

#8751 exposed many latent bug in the testsuite and fixing all of them may take a while. That's why I suggest revert the PR for now and continue to fix the problems in #8776.

### DIFF
--- a/numba/testing/main.py
+++ b/numba/testing/main.py
@@ -770,24 +770,28 @@ class ParallelTestRunner(runner.TextTestRunner):
 
         # Split the tests and recycle the worker process to tame memory usage.
         chunk_size = 100
+        splitted_tests = [self._ptests[i:i + chunk_size]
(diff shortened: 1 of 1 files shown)


sklam:
`numba_smoketest_cpu_yaml_164`

sklam:
all linux x64 and ppc64le platforms have passed.

sklam:
BFID: `numba_yaml_312`

stuartarchibald (approved):
Thanks for creating the patch. As discussed OOB, this should stabilise the build and create a point from which this change (and the potential bugs it exposed) can be explored again.

sklam:
> BFID: `numba_yaml_312`

Buildfarm is stable enough with this PR. Merging

PR #8776, opened 2023-02-22
Fix test suite

sklam:
- fix oversubscription of CPU resources 
- fix stability issue with tests that checks refcount

PR #8781, opened 2023-02-23
Ensure gc.collect() is called before checking refcount in tests.

sklam:
Splitted from https://github.com/numba/numba/pull/8776#issuecomment-1442164785 as requested by reviewer.

### DIFF
--- a/numba/tests/support.py
+++ b/numba/tests/support.py
@@ -266,6 +266,16 @@ class TestCase(unittest.TestCase):
                 self.fail("Refcount changed from %d to %d for object: %r"
                           % (old, new, obj))
 
+    def assertRefCountEqual(self, *objects):
+        gc.collect()
+        rc = [sys.getrefcount(x) for x in objects]
+        for i, x in enumerate(objects[1:], start=1):
+            rc_0 = rc[0]
+            rc_i = rc[i]
+            if rc_0 != rc_i:
+                self.fail(f"Refcount for objects does not match. "
+                          f"#0({rc_0}) != #{i}({rc_i}) does not match.")
+
     @contextlib.contextmanager
     def assertNoNRTLeak(self):
         """
--- a/numba/tests/support.py
+++ b/numba/tests/support.py
@@ -269,8 +269,8 @@ class TestCase(unittest.TestCase):
     def assertRefCountEqual(self, *objects):
         gc.collect()
         rc = [sys.getrefcount(x) for x in objects]
-        for i, x in enumerate(objects[1:], start=1):
-            rc_0 = rc[0]
+        rc_0 = rc[0]
+        for i in range(len(objects))[1:]:
             rc_i = rc[i]
             if rc_0 != rc_i:
                 self.fail(f"Refcount for objects does not match. "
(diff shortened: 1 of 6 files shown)


numba/tests/support.py
```diff
@@ -266,6 +266,16 @@ def assertRefCount(self, *objects):
...
 
+    def assertRefCountEqual(self, *objects):
+        gc.collect()
+        rc = [sys.getrefcount(x) for x in objects]
+        for i, x in enumerate(objects[1:], start=1):
+            rc_0 = rc[0]
+            rc_i = rc[i]
+            if rc_0 != rc_i:
+                self.fail(f"Refcount for objects does not match. "
+                          f"#0({rc_0}) != #{i}({rc_i}) does not match.")
```
stuartarchibald, line 277:
```suggestion
        rc_0 = rc[0]
        for i in range(1, len(objects) - 1):
            rc_i = rc[i]
            if rc_0 != rc_i:
                self.fail(f"Refcount for objects does not match. "
                          f"#0({rc_0}) != #{i}({rc_i}) does not match.")
```
hoist loop invariant and don't think `enumerate` is needed?

numba/tests/test_np_randomgen.py
```diff
@@ -231,9 +232,11 @@ def test_npgen_boxing_refcount(self):
         do_box = numba.njit(lambda x:x)
 
         y = do_box(rng_instance)
+        gc.collect()
```
stuartarchibald, line 235: Could this use `with self.assertRefCount(...):` ?

stuartarchibald (commented):
Couple of minor things else looks good, thanks for moving this patch to a separate PR @sklam.

numba/tests/test_np_randomgen.py
sklam: it's expecting a `+ 1` so it cannot use that testmethod

stuartarchibald: ah ok, thanks for checking.

numba/tests/support.py
sklam: shouldn't be `len(objects) - 1`.  addressed in https://github.com/numba/numba/pull/8781/commits/70cedbc9de634ef8e7f50c8cd974fb3113a8e69e

numba/tests/support.py
```diff
@@ -266,6 +266,16 @@ def assertRefCount(self, *objects):
                 self.fail("Refcount changed from %d to %d for object: %r"
                           % (old, new, obj))
 
+    def assertRefCountEqual(self, *objects):
+        gc.collect()
+        rc = [sys.getrefcount(x) for x in objects]
+        rc_0 = rc[0]
+        for i in range(len(objects))[1:]:
```
stuartarchibald, line 273:
```suggestion
        for i in range(1, len(objects)):
```
is it just this?

sklam: yup. you can imagine `range()` returning a list. that's allowed use from py2 when range() does return a list.

stuartarchibald: Can leave it, it's just a preference.

stuartarchibald (approved):

PR #8783, opened 2023-02-24
Add support for NamedTuple casting and constructor default arguments

danieldjohnson:
This PR adds support for some additional features for named tuples.

### Casting

Enables casts between NamedTuple types with the same instance class and compatible element types, similar to how regular tuples can be cast if they have compatible element types. Casting is supported for both NamedTuple and NamedUniTuple. Casts remain unsupported between tuples of different instance classes (i.e. with different Python types).

This makes it possible to use namedtuples in place of regular tuples in more situations. For instance, we can pass a namedtuple containing ints to a function that expects floats, or a namedtuple containing None to a function expecting that element to be optional. Before this change, both of those resulted in an error unless each element was cast manually.

### Constructor default arguments

NamedTuple classes can have default arguments in Python 3.7 and later. This is especially convenient when defining a NamedTuple class by subclassing `typing.NamedTuple`.

However, Numba does not currently process the default arguments in the correct way. This appears to be due to the use of `make_callable_template` to build the callable template for the constructor, which inspects the signature of the `typer` function itself and then ["hacks omitted parameters out"](https://github.com/numba/numba/blob/4535cb650da992a4c06bd252f94ee50095795f46/numba/core/typing/templates.py#L448-L454). In this case the lowering function for namedtuples expects to receive values for these parameters, and in some cases this can lead to trying to access memory out of the allocated bounds, e.g. if we define
```
class Foo(typing.NamedTuple):
    foo: int
    bar: int = 0
```
then construct `foo = Foo(1)` and try to access `foo.bar` inside `nopython` jit code.

This change modifies the call template for the namedtuple class object to infer default arguments from the python signature, by using `fold_arguments` and `generic` to build an explicit template instead of deferring to `make_callable_template`.

### DIFF
--- a/numba/core/types/containers.py
+++ b/numba/core/types/containers.py
@@ -377,7 +377,24 @@ class StarArgUniTuple(_StarArgTupleMixin, UniTuple):
 
 
 class BaseNamedTuple(BaseTuple):
-    pass
+    def can_convert_to(self, typingctx, other):
+        if not isinstance(other, BaseNamedTuple):
+            return
+        if (
+            self.instance_class != other.instance_class
+            or self.count != other.count
+        ):
+            return
+        if self.count == 0:
+            return Conversion.safe
+
+        kinds = [
+            typingctx.can_convert(ta, tb)
+            for ta, tb in zip(self.types, other.types)
+        ]
+        if any(kind is None for kind in kinds):
+            return
+        return max(kinds)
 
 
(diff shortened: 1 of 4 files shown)


kc611:
Hi @danieldjohnson, Thanks for the PR, your contributions to Numba are much appreciated. I've marked the PR as ready for review.

PR #8802, opened 2023-03-09
Move the minimum supported NumPy version to 1.21

### LINKED ISSUE #8707: Move baseline supported NumPy version to 1.21.

stuartarchibald:
This moves the minimum supported NumPy version to 1.21. The patch removes code specialising on any version of NumPy prior to 1.21 (the oldest of which was from 1.16). It also updates code to remove workarounds necessary due to changes in the NumPy API (e.g. np.broadcast_shapes). Docs, buildscript and `setup.py`/`__init__.py` are also updated.

Closes #8707

### DIFF
--- a/numba/np/npdatetime.py
+++ b/numba/np/npdatetime.py
@@ -384,16 +384,12 @@ def _create_timedelta_comparison_impl(ll_op, default_value):
                 else:
                     builder.store(builder.icmp_unsigned(ll_op, norm_a, norm_b), ret)
             with otherwise:
-                if numpy_support.numpy_version < (1, 16):
-                    # No scaling when comparing NaTs
-                    builder.store(builder.icmp_unsigned(ll_op, va, vb), ret)
+                # NumPy >= 1.16 switched to NaT ==/>=/>/</<= NaT being
+                # False and NaT != <anything, including NaT> being True
+                if ll_op == '!=':
+                    builder.store(cgutils.true_bit, ret)
                 else:
-                    # NumPy >= 1.16 switched to NaT ==/>=/>/</<= NaT being
-                    # False and NaT != <anything, including NaT> being True
-                    if ll_op == '!=':
-                        builder.store(cgutils.true_bit, ret)
-                    else:
-                        builder.store(cgutils.false_bit, ret)
+                    builder.store(cgutils.false_bit, ret)
         res = builder.load(ret)
         return impl_ret_untracked(context, builder, sig.return_type, res)
 
@@ -411,14 +407,8 @@ def _create_timedelta_ordering_impl(ll_op):
                     context, builder, va, vb, ta, tb)
                 builder.store(builder.icmp_signed(ll_op, norm_a, norm_b), ret)
             with otherwise:
-                if numpy_support.numpy_version < (1, 16):
-                    # No scaling when comparing NaT with something else
-                    # (i.e. NaT is <= everything else, since it's the smallest
-                    #  int64 value)
-                    builder.store(builder.icmp_signed(ll_op, va, vb), ret)
-                else:
-                    # NumPy >= 1.16 switched to NaT >=/>/</<= NaT being False
-                    builder.store(cgutils.false_bit, ret)
+                # NumPy >= 1.16 switched to NaT >=/>/</<= NaT being False
+                builder.store(cgutils.false_bit, ret)
         res = builder.load(ret)
         return impl_ret_untracked(context, builder, sig.return_type, res)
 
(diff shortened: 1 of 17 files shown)


gmarkall:
gpuci run tests

numba/tests/test_np_functions.py
```diff
@@ -2811,27 +2796,21 @@ def input_variations():
         ## fixed here: https://github.com/numpy/numpy/pull/12713 for np 1.16
         to_begin = np.array([1, 2, 3.142, np.nan, 5, 6, 7, -8, np.nan])
         params = {'ary': np.arange(-4, 6), 'to_begin': to_begin}
-        if numpy_version < (1, 16):
-            _check(params)
-        else:
-            # np 1.16 raises, cannot cast float64 array to intp array
-            _check_raises_type_error(params, 'to_begin')
+        # np >1.16 raises, cannot cast float64 array to intp array
```
gmarkall, line 2799: Possibly redundant comment?

numba/tests/test_npdatetime.py
```diff
@@ -385,12 +385,11 @@ def check(a, b, expected):
             expected_val = expected
             not_expected_val = not expected
 
-            if numpy_version >= (1, 16):
-                # since np 1.16 all NaT == comparisons are False, including
-                # NaT==NaT, conversely != is True
-                if np.isnat(a) or np.isnat(a):
-                    expected_val = False
-                    not_expected_val = True
+            # since np 1.16 all NaT == comparisons are False, including
```
gmarkall, line 388: Multiple comments contrasting NP 1.16 with older behaviour (I won't add more comments on this file to save noise)

gmarkall (commented):
I've read through the changes and they look good - the only comment I had is it looks like there are some comments that wouldn't have been written if there weren't divergent behaviour between NumPy versions - what do you think about their necessity now?

I need to go through this tomorrow to check if anything is missed, but other than that, this is looking good so far.

numba/np/npdatetime.py
```diff
@@ -384,16 +384,12 @@ def impl(context, builder, sig, args):
                 else:
                     builder.store(builder.icmp_unsigned(ll_op, norm_a, norm_b), ret)
             with otherwise:
-                if numpy_support.numpy_version < (1, 16):
-                    # No scaling when comparing NaTs
-                    builder.store(builder.icmp_unsigned(ll_op, va, vb), ret)
+                # NumPy >= 1.16 switched to NaT ==/>=/>/</<= NaT being
```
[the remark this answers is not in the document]
stuartarchibald: I considered deleting these comments but chose to leave them in as a reminder for anyone revisiting this code as to what the behaviour is around `NaT`... if you think it would be better/more clean to remove them am happy to do so.

gmarkall (changes requested):
I don't feel strongly about the comments.

I think this test will never run again, so could be removed:

https://github.com/numba/numba/blob/f54d55345a27197469be172a444d59384f18bf5e/numba/tests/test_np_functions.py#L2780-L2786

I also wonder whether we need to carry this macro around anymore (my assumption is that if we require NumPy 1.21 nobody should be building Numba with a version less than 1.21), but I'd maybe lean towards punting it to a separate issue / PR following this one, given it's a change at the C level and I'd rather not hold this one up:

https://github.com/numba/numba/blob/f54d55345a27197469be172a444d59384f18bf5e/numba/_numba_common.h#L24-L41

stuartarchibald:
Thanks for reviewing this @gmarkall.

> I don't feel strongly about the comments.

I elected to modify them, I think they are useful notes RE behaviour but the NumPy 1.16 part was/is a distraction. In the process I found another couple of redundant 1.16 references which are now removed. Commit 5d7b0fb  contains these changes.

> I think this test will never run again, so could be removed:
> 
> https://github.com/numba/numba/blob/f54d55345a27197469be172a444d59384f18bf5e/numba/tests/test_np_functions.py#L2780-L2786

Thanks for catching this, removed in dac8800.

> 
> I also wonder whether we need to carry this macro around anymore (my assumption is that if we require NumPy 1.21 nobody should be building Numba with a version less than 1.21), but I'd maybe lean towards punting it to a separate issue / PR following this one, given it's a change at the C level and I'd rather not hold this one up:
> 
> https://github.com/numba/numba/blob/f54d55345a27197469be172a444d59384f18bf5e/numba/_numba_common.h#L24-L41

I also wondered about this as it appeared when searching for references to older NumPy versions. IIRC Numba can be compiled using a much older version of NumPy than it supports at runtime, so I elected to leave it in with view of making this patch potentially a lower risk.

Finally b746d41 is a flake8 violation fix-up. Thanks again for the review.

gmarkall:
> I also wondered about this as it appeared when searching for references to older NumPy versions. IIRC Numba can be compiled using a much older version of NumPy than it supports at runtime, so I elected to leave it in with view of making this patch potentially a lower risk.

Sure, I didn't think it should be done in this PR. It's not clear to me why we support building with an older-than-supported NumPy  - I think that's not documented? Is it mentioned somewhere?

gmarkall (approved):
Thanks for the updates, I think this is good to merge assuming gpuCI is fine.

gmarkall:
gpuci run tests

stuartarchibald:
> > I also wondered about this as it appeared when searching for references to older NumPy versions. IIRC Numba can be compiled using a much older version of NumPy than it supports at runtime, so I elected to leave it in with view of making this patch potentially a lower risk.
> 
> Sure, I didn't think it should be done in this PR. It's not clear to me why we support building with an older-than-supported NumPy - I think that's not documented? Is it mentioned somewhere?

I'd have to go and check but I think it is/was implicit through the use of `conda build` which has/had a default NumPy version at build time. This may not be applicable or of relevance to Numba any more.

PR #8803, opened 2023-03-09
Attempted fix to #8789 by changing `compile_ptx` to accept a signature instead of argument tuple

### LINKED ISSUE #8789: Support passing a signature to `compile_ptx`
As noted in https://github.com/numba/numba/issues/8400#issuecomment-1438485725, argument types can be passed to `compile_ptx` but the return type cannot be specified. If `compile_ptx` took a signature then the return type could also be specified. The signature could be a signature object, a tuple of argument types, or a string representation of the signature - this would be facilitated by calling `sigutils.normalize_signature()` as the `@cuda.jit` decorator does.

This would allow one to write:

```python
def add(x, y):
    return x + y

ptx, resty = cuda.compile_ptx_for_current_device(add, int32(int32, int32), device=True)
```

and get a function that returns an `int32` back instead of the `int64` currently implied by NBEP 1 typing.

cc @cstyl, who originally mentioned a scenario in which this would be desirable in https://github.com/numba/numba/issues/8400#issuecomment-1229502785.

KyanCheung:
Think this should work. This fix would be backwards compatible with how it worked previously (when one only specified the argument types of the inputs only) but also allows specifying output type.

Fixes #8789

### DIFF
--- a/numba/cuda/compiler.py
+++ b/numba/cuda/compiler.py
@@ -1,5 +1,5 @@
 from numba.core.typing.templates import ConcreteTemplate
-from numba.core import types, typing, funcdesc, config, compiler
+from numba.core import types, typing, funcdesc, config, compiler, sigutils
 from numba.core.compiler import (sanitize_compile_result_entries, CompilerBase,
                                  DefaultPassBuilder, Flags, Option,
                                  CompileResult)
@@ -235,12 +235,13 @@ def compile_cuda(pyfunc, return_type, args, debug=False, lineinfo=False,
 
 
 @global_compiler_lock
-def compile_ptx(pyfunc, args, debug=False, lineinfo=False, device=False,
+def compile_ptx(pyfunc, sig, debug=False, lineinfo=False, device=False,
                 fastmath=False, cc=None, opt=True):
     """Compile a Python function to PTX for a given set of argument types.
 
     :param pyfunc: The Python function to compile.
-    :param args: A tuple of argument types to compile for.
+    :param sig: The signature representing the function's input and output
+                     types.
     :param debug: Whether to include debug info in the generated PTX.
     :type debug: bool
     :param lineinfo: Whether to include a line mapping from the generated PTX
@@ -255,8 +256,8 @@ def compile_ptx(pyfunc, args, debug=False, lineinfo=False, device=False,
     :param fastmath: Whether to enable fast math flags (ftz=1, prec_sqrt=0,
                      prec_div=, and fma=1)
     :type fastmath: bool
-    :param cc: Compute capability to compile for, as a tuple ``(MAJOR, MINOR)``.
-               Defaults to ``(5, 3)``.
+    :param cc: Compute capability to compile for, as a tuple
+               ``(MAJOR, MINOR)``. Defaults to ``(5, 3)``.
     :type cc: tuple
     :param opt: Enable optimizations. Defaults to ``True``.
     :type opt: bool
@@ -276,9 +277,11 @@ def compile_ptx(pyfunc, args, debug=False, lineinfo=False, device=False,
         'opt': 3 if opt else 0
     }
 
+    args, return_type = sigutils.normalize_signature(sig)
+
     cc = cc or config.CUDA_DEFAULT_PTX_CC
-    cres = compile_cuda(pyfunc, None, args, debug=debug, lineinfo=lineinfo,
-                        fastmath=fastmath,
+    cres = compile_cuda(pyfunc, return_type, args, debug=debug,
+                        lineinfo=lineinfo, fastmath=fastmath,
                         nvvm_options=nvvm_options, cc=cc)
     resty = cres.signature.return_type
 
@@ -300,13 +303,13 @@ def compile_ptx(pyfunc, args, debug=False, lineinfo=False, device=False,
     return ptx, resty
 
 
-def compile_ptx_for_current_device(pyfunc, args, debug=False, lineinfo=False,
+def compile_ptx_for_current_device(pyfunc, sig, debug=False, lineinfo=False,
                                    device=False, fastmath=False, opt=True):
     """Compile a Python function to PTX for a given set of argument types for
     the current device's compute capabilility. This calls :func:`compile_ptx`
     with an appropriate ``cc`` value for the current device."""
     cc = get_current_device().compute_capability
-    return compile_ptx(pyfunc, args, debug=debug, lineinfo=lineinfo,
+    return compile_ptx(pyfunc, sig, debug=debug, lineinfo=lineinfo,
                        device=device, fastmath=fastmath, cc=cc, opt=True)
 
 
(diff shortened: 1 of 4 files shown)


KyanCheung:
Wait - am I required to write a test for this? If so, I'd like to apologise for making the PR without following the correct procedure. I don't see an option to undo a PR, unfortunately.

gmarkall:
gpuci run tests

gmarkall:
(Just running the tests on gpuCI to check that nothing is broken by this - I don't expect it would break anything, your changes look correct based on a quick inspection).

KyanCheung:
Should I only test for if the signature is `int32(int32, int32)`? Or should I also test for something like `int16(int16, int16)`, or using string input, or checking that it works with an argument tuple only?

gmarkall:
Good question - If I were writing it, I'd add tests that use both `int16` and `int32` for a little reassurance there isn't some chance case where only `int32` works. There should be tests using the `signature` objects generated by writing e.g. `int32(int32, int32)` and tests using the string notation `"int32(int32, int32)"` to check that works as well. The existing tests already serve the case where only a tuple of argument types is provided, so I'd not add any new tests that do that.

I don't think it's necessary to add multiple different function bodies, I think just addition is sufficient since it demonstrates the case where we're effectively "overriding" the natural return type `int64` of `int32 + int32`.

KyanCheung:
gpuci run tests

KyanCheung:
All right, seems like the code has passed the checks. Should we proceed with merging?

gmarkall:
gpuci run tests

gmarkall:
> gpuci run tests

Only maintainers can trigger gpuCI, so although it was a good proactive step, it unfortunately didn't test with GPU hardware following this comment.

> All right, seems like the code has passed the checks. Should we proceed with merging?

I appreciate your enthusiasm! The next step is for the code to have a review - I've not yet reviewed it. The tests passing is a good place to be in to start from, though.

KyanCheung:
Probably best not to, since the original function wasn't supposed to allow lists.

All lists should be tuples now.

gmarkall:
Thanks - I didn't get to responding to your comment before you made the change, but I was going to suggest that it's not a good idea to convert lists to a tuple because it would be inconsistent with the `@jit` and `@vectorize` decorators, where if you pass a list it's interpreted as a list of signatures, e.g. `[float32(float32, float32), float64(float64, float64)]` - for `compile_ptx` I think we really only want to compile one thing at a time, so interpreting it as a list of signatures wouldn't be a good path forward either.

gpuci run tests

KyanCheung:
For the `test_libdevices.py` test, is `pyreturns` meant to be the return types of the function? I thought `compile_ptx` would use whatever types it was given as input.

gmarkall:
I think the problems here stem from the transformations of the form `[ty]` to `(ty)` - the former is a list construction, the latter is a parenthesised reference to the variable `ty`. To create a tuple, you would need to write `(ty,)` instead.

gmarkall:
gpuci run tests

numba/cuda/tests/cudapy/test_compiler.py
```diff
@@ -30,8 +30,8 @@ def test_device_function(self):
         def add(x, y):
             return x + y
 
-        args = (float32, float32)
-        ptx, resty = compile_ptx(add, args, device=True)
+        sig = (float32, float32)
+        ptx, resty = compile_ptx(add, sig, device=True)
```
gmarkall, line 34: Since you're passing a tuple of args and not a signature object, I think this change makes the code less succinct - I'd suggest going back to its original form:

```suggestion
        args = (float32, float32)
        ptx, resty = compile_ptx(add, args, device=True)
```

numba/cuda/tests/cudapy/test_compiler.py
```diff
@@ -43,6 +43,19 @@ def add(x, y):
...
+        # Check that function's output matches signature
+        sig_int32 = int32(int32, int32)
+        ptx, resty = compile_ptx(add, sig_int32, device=True)
+        self.assertEqual(resty, int32)
+
+        sig_int16 = int16(int16, int16)
+        ptx, resty = compile_ptx(add, sig_int16, device=True)
+        self.assertEqual(resty, int16)
+
+        sig_string = "uint32(uint32, uint32)"
```
gmarkall, line 55: Might be worth adding a comment to explicitly point out that this is also testing with a string-based signature in contrast to the signature objects above.

numba/cuda/tests/cudapy/test_compiler.py
```diff
@@ -130,8 +143,8 @@ def test_compile_ptx_for_current_device(self):
         def add(x, y):
             return x + y
 
-        args = (float32, float32)
-        ptx, resty = compile_ptx_for_current_device(add, args, device=True)
+        sig = (float32, float32)
+        ptx, resty = compile_ptx_for_current_device(add, sig, device=True)
```
gmarkall, line 147: Similar to the comment about the tuple of args above:
```suggestion
        args = (float32, float32)
        ptx, resty = compile_ptx_for_current_device(add, args, device=True)
```

numba/cuda/compiler.py
```diff
@@ -235,12 +235,13 @@ def compile_cuda(pyfunc, return_type, args, debug=False, lineinfo=False,
...
 @global_compiler_lock
-def compile_ptx(pyfunc, args, debug=False, lineinfo=False, device=False,
+def compile_ptx(pyfunc, sig, debug=False, lineinfo=False, device=False,
                 fastmath=False, cc=None, opt=True):
     """Compile a Python function to PTX for a given set of argument types.
 
     :param pyfunc: The Python function to compile.
-    :param args: A tuple of argument types to compile for.
+    :param sig: The signature representing the function's input and output
+                     types.
```
gmarkall, line 244: Indentation seems a little off:
```suggestion
                types.
```

numba/cuda/compiler.py
```diff
@@ -255,8 +256,8 @@ def compile_ptx(pyfunc, args, debug=False, lineinfo=False, device=False,
     :param fastmath: Whether to enable fast math flags (ftz=1, prec_sqrt=0,
                      prec_div=, and fma=1)
     :type fastmath: bool
-    :param cc: Compute capability to compile for, as a tuple ``(MAJOR, MINOR)``.
-               Defaults to ``(5, 3)``.
+    :param cc: Compute capability to compile for, as a tuple
+               ``(MAJOR, MINOR)``. Defaults to ``(5, 3)``.
```
gmarkall, line 260: I'm not sure this change was necessary (I think the Numba flake8 config permits 80 chars as opposed to 79) but I also don't think it matters much either way :-)

gmarkall (changes requested):
Many thanks for the updates! There are just some very small style comments on the diff - once these are resolved I think this should be good to merge.

KyanCheung:
OK, all of these things should be fixed now.

KyanCheung:
All right, thank you for working with me for my first PR!

stuartarchibald:
@KyanCheung Congratulations on your first contribution to Numba!

PR #8804, opened 2023-03-09
Pr/8725

DrTodd13:


sklam (approved):
The refactoring done here is nice and will make the parfor passes more extensible

PR #8809, opened 2023-03-13
Update LLVM versions for 0.57 release

apmasell:
Follow up for #8535


sklam:
holding this for @gmarkall's llvm14 fix.

docs/source/user/installing.rst
```diff
@@ -260,7 +260,9 @@ information.
 +----------++--------------+---------------------------+----------------------------+------------------------------+-------------------+-----------------------------+
 | Numba     | Release date | Python                    | NumPy                      | llvmlite                     | LLVM              | TBB                         |
 +===========+==============+===========================+============================+==============================+===================+=============================+
-| 0.57.0    | TBC          | 3.8.x <= version < 3.12   | 1.21 <= version < 1.25     | 0.40.x                       | 11.x              | 2021.x                      |
+| 0.57.0    | TBC          | 3.8.x <= version < 3.12   | 1.21 <= version < 1.25     | 0.40.x                       | 14.x              | 2021.x                      |
+|           |              |                           |                            |                              | (11.x for Linux   |                             |
```
gmarkall, line 264: Do we still plan to go with only requiring 11.x, assuming the llvmdev PR I make today resolves the LLVM 14 issues to everyone's satisfaction?

gmarkall: Oops, just saw @sklam's comment on the main PR!

sklam: Now that GOT patch is merged, we can drop the part about llvm 11

apmasell:
Updated.

stuartarchibald:
Should https://github.com/numba/numba/blob/e520da897f1eb08cf9d3912b1ef67a1081f3a786/numba/__init__.py#L146 also be updated?

apmasell:
Updated to 14 per discussion in meeting.

sklam:
no idea why RTD is failing. the docs/environment.yml gets the correct llvmlite when i run it locally.

sklam:
I confirmed that conda-4.7 has a bug that would cause it to select an old build of llvmlite. This is no longer a problem in conda-4.8.

sklam (approved):

esc (approved):
I checked the RTD build and the docs are generated fine. The version table and min versions are fine too, thank you for the patch!

PR #8816, opened 2023-03-14
Fix some buildfarm test failures

sklam:
as titled


sklam:
`numba_smoketest_cpu_yaml_169`

sklam:
Passed buildfarm smoketest (minus known arm64/aarch64 segfault)

stuartarchibald (approved):
Thanks for the patch.

PR #8817, opened 2023-03-14
Remove NRT decref in generators

### LINKED ISSUE #8800: double free malloc error in generator returning reference
## Reporting a bug

I'm trying to make a generator that returns a reference to the same array, stored in the generator (i.e. not a copy):

```python
from numba import njit
import numpy as np

@njit
def generator():
    z = np.zeros(shape=(3,), dtype=np.int32)
    for i in range(3):    # bug hits on i=2, i.e. after 2 iterations
        z[i] = 1
        yield z    # no bug if return z.copy()

for g in generator():
    print(g)
```

The bug does not happen if I return a copy. Also it occurs at the 3rd iteration, which maybe implies that something related to `malloc` is happening at each loop?

apmasell:
As titled. Closes #8800

### DIFF
--- a/numba/core/generators.py
+++ b/numba/core/generators.py
@@ -350,7 +350,4 @@ class LowerYield(object):
             ty = self.gentype.state_types[state_index]
             val = self.context.unpack_value(self.builder, ty, state_slot)
             self.lower.storevar(val, name)
-            # Previous storevar is making an extra incref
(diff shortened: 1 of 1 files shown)


gmarkall:
Closing as the hypothesis this was testing seems to have been disproved by the negative side-effects of removing this decref.

PR #8820, opened 2023-03-16
Removed `mk_unique_var` from `parfor.py`

KyanCheung:
Related to #8230

### DIFF
--- a/numba/parfors/parfor.py
+++ b/numba/parfors/parfor.py
@@ -37,7 +37,6 @@ from numba.stencils.stencilparfor import StencilPass
 from numba.core.extending import register_jitable, lower_builtin
 
 from numba.core.ir_utils import (
-    mk_unique_var,
     next_label,
     mk_alloc,
     get_np_ufunc_typ,
@@ -1526,7 +1525,7 @@ class PreParforPass(object):
                             dtype = typ.dtype
                             scope = block.scope
                             loc = instr.loc
-                            g_np_var = ir.Var(scope, mk_unique_var("$np_g_var"), loc)
+                            g_np_var = scope.redefine("$np_g_var", loc)
                             self.typemap[g_np_var.name] = types.misc.Module(numpy)
                             g_np = ir.Global('np', numpy, loc)
                             g_np_assign = ir.Assign(g_np, g_np_var, loc)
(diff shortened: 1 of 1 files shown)


gmarkall:
Thanks for working on this! There are a few other `mk_unique_var()` removal PRs that went in in the past, which might be helpful references: https://github.com/numba/numba/pulls?q=is%3Apr+mk_unique_var+is%3Aclosed

KyanCheung:
All of the other `mk_unique_var()` PRs I saw involved editing something of the form `ir.Var(scope, mk_unique_var(name), loc)` to `scope.redefine(name, loc)`, or `mk_unique_var(name)` to `scope.redefine(name, loc).name`, so I couldn't really draw any conclusions from the other PRs.

I'm still not sure why the same variable names are being assigned by `scope.redefine` - presumably the scope itself is changing, but I don't have a broad enough understanding of the code to figure out where.

Nonetheless, I've decided to use a while loop to check if the variable name has been used, and if so, reassign another - this is similar to what `Scope.redefine` is doing anyways, so I'm assuming it's fine to use it here too. Do let me know if you think it should be handled differently though.

Unfortunately, my code changes don't fix the issue with `test_basic94` in `test_stencils` failing, but the commits fix the other test suites that caused errors in the Github check.

KyanCheung:
With all errors fixed, I think this is ready for review.

DrTodd13 (changes requested):
I suggest that these 3 places where you're doing a redefine and then have a while loop to check if it is already in the typemap be made into a function.  Can you say why these loops are necessary here but not necessary in the other places?  Is it just that the tests fail for those locations and not the others?  Could something change to make the other locations start to fail as well?

KyanCheung:
Alright, added a function.

I only did those 3 places initially because they were causing tests to fail, but I've also decided to edit the other places because it's possible that there might be duplication issues that prop up outside of the tests.

KyanCheung:
I'm guessing `ModuleNotFoundError: No module named 'typeguard.importhook'` is likely to be an error on Azure's end?

DrTodd13:
@KyanCheung @stuartarchibald @sklam There used to be mk_unique_var which would create a unique var across the whole function.  Then, we tried scope.redefine but that only redefined within the current scope and generated some conflicts across the whole function.  Now, we have a new function that tries multiple scope.redefine calls until it finds one that is unique across the whole function.  How is this better than what we started with?  I never did this kind of analysis when originally writing the code but we could think about each variable and its usage to see if it is only used within the scope or also outside the scope.  I suspect that most of them are used outside the scope and given that making a function-wide unique var is always safe I'm wondering the benefits of making these changes.

gmarkall:
@KyanCheung Many thanks for your effort so far, which has helped progress our thinking about the way forward for Issue #8230 - unfortunately it looks like the approach in this PR doesn't quite solve the problem we were hoping, but there is a way forward to resolve issue #8230, outlined in https://github.com/numba/numba/issues/8230#issuecomment-1478330246. If you still feel enhusiastic, a PR that moves `mk_unique_var()` to parfors would be welcomed - if not, I understand. Either way, thankyou for your contributions, and I hope we can encourage you to continue to contribute!

KyanCheung:
Alright, I'll close this PR then. I'll be a bit busy for the next few days, but if I feel like it I'll see if I can move the definition of `mk_unique_vars` to `parfors` later.

gmarkall:
Many thanks for your understanding - do let us know if we can help out in any way with the new PR.

PR #8833, opened 2023-03-21
Fix typeguard import hook location.

stuartarchibald:
Package ``typeguard`` moved the function ``install_import_hook`` to the top level module for version 3+. This adapts the test runner to accept both the old and the new location.


sklam (approved):
Changes make sense and fix confirmed in CI

PR #8837, opened 2023-03-22
Update AzureCI matrix for Python 3.11/NumPy 1.21..1.24

### LINKED ISSUE #8834: Clean up AzureCI matrix
> The CI matrix will probably need some further adjustment, as will general build configuration, but not for this PR as it seems out of scope.

_Originally posted by @stuartarchibald in https://github.com/numba/numba/issues/8822#issuecomment-1478350653_

#8822 PR may have increased the number of CI configuration too much. We should consider reducing it and perhaps remove testing of older python versions.

stuartarchibald:
This updates the AzureCI test matrix. It aims to ensure that:

* The latest supported NumPy and Python are tested on all OS.
* The oldest supported NumPy and Python are tested on all OS.
* That the rest of the "matrix" formed from the outer product of supported NumPys x supported Pythons has at least one build if packages are available.

Closes #8834

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -3,7 +3,7 @@ trigger:
 
 variables:
   # Change the following along with adding new TEST_START_INDEX.
-  TEST_COUNT: 22
+  TEST_COUNT: 21
 
 jobs:
 # Mac and Linux use the same template with different matrixes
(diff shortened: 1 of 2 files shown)


stuartarchibald:
Close/Open to trigger new CI build against fixed `main`.

sklam:
CI is failing because there is no scipy for numpy=1.22 and python=3.11

sklam:
The window llvm error is tracked at https://github.com/numba/numba/issues/8840

stuartarchibald:
> CI is failing because there is no scipy for numpy=1.22 and python=3.11

Thanks, 5424212 fixes.

azure-pipelines.yml
```diff
@@ -17,17 +17,11 @@ jobs:
         NUMPY: '1.21'
         CONDA_ENV: 'azure_ci'
         TEST_START_INDEX: 0
-      py310_np124:
-        PYTHON: '3.10'
-        NUMPY: '1.24'
-        CONDA_ENV: 'azure_ci'
-        TEST_THREADING: 'tbb'
```
sklam, line 24: This remove TBB testing on OSX

stuartarchibald: Should probably keep on doing this as it tests a different loading path. Reinstated in 1bd9dd0.

PR #8839, opened 2023-03-22
Add Dynamic Shared Memory example.

### LINKED ISSUE #5025: Provide documentation, docstrings, tests, examples for dynamic shared memory
Dynamic shared memory is supported in CUDA Python but there appears to be little in the way of documentation and examples. Some things to consider adding:

- User guide / reference documentation
- Examples
- Docstrings
- Additional tests

Some useful references:

- [CUDA C Programming guide: Execution Configuration](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#execution-configuration)
- [CUDA C Programming guide: Shared Memory](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#shared)
- [Numba-users thread in which I previously described dynamic shared memory](https://groups.google.com/a/continuum.io/forum/#!topic/numba-users/yckiwJZ0y-A)
- Issue #2463 - support for more than one dimension would be nice to add.

k1m190r:
@gmarkall have provided excellent demonstration of dynamic shared memory as answer to my question on [numba discourse](https://numba.discourse.group/t/how-do-i-use-sharedmem-parameter-in-kernel/1841/2). It is essentially copy and paste of @gmarkall 's answer. 

This PR partially addresses #5025. 

Fixes #5025.

### DIFF
--- a/docs/source/cuda/memory.rst
+++ b/docs/source/cuda/memory.rst
@@ -156,6 +156,81 @@ traditional dynamic memory management.
 .. seealso::
    :ref:`Matrix multiplication example <cuda-matmul>`.
 
+Dynamic Shared Memory
+---------------------
+
+In order to use dynamic shared memory in kernel code declare shared array of size 0:
+
+.. code-block:: python
+
+   @cuda.jit
+   def kernel_func(x):
+      dyn_arr = cuda.shared.array(0, dtype=np.float32)
+      ...
+
+and specify the size of dynamic shared array in bytes during kernel invokation (dispatcher object instantiation):
+
+.. code-block:: python
+
+   kernel_func[32, 32, 0, 128](x)
+
+In above code dispatcher object is instantiated with 4 parameters:
+
+.. code-block:: python
+
+   kernel_func[grid_dim, block_dim, stream_id, dyn_shared_mem_size]
+
+**Note:** that all dynamic shared memory arrays *alias*, so if you want to have multiple dynamic shared arrays, you need to take *disjoint* views of the arrays. For example, consider:
+
+.. code-block:: python
+
+   from numba import cuda
+   import numpy as np
+
+   @cuda.jit
+   def f():
+      f32_arr = cuda.shared.array(0, dtype=np.float32)
+      i32_arr = cuda.shared.array(0, dtype=np.int32)
+      f32_arr[0] = 3.14
+      print(f32_arr[0])
+      print(i32_arr[0])
+
+   f[1, 1, 0, 4]()
+   cuda.synchronize()
+
+This allocates 4 bytes of shared memory (large enough for one int32 or one float32) and declares dynamic shared memory arrays of type int32 and of type float32. When `f32_arr[0]` is set, this also sets the value of `i32_arr[0]`, because they're pointing at the same memory. So we see as output:
+
+.. code-block:: pycon
+
+   3.140000
+   1078523331
+
+because `1078523331` is the `int32` represented by the bits of the `float32` IEEE754 `3.14`.
+
+If we take disjoint views of the dynamic shared memory:
+
+.. code-block:: python
+
+   @cuda.jit
+   def f_with_view():
+      f32_arr = cuda.shared.array(0, dtype=np.float32)
+      i32_arr = cuda.shared.array(0, dtype=np.int32)[1:] # 1 int32 = 4 bytes
+      f32_arr[0] = 3.14
+      i32_arr[0] = 1
+      print(f32_arr[0])
+      print(i32_arr[0])
+
+   f_with_view[1, 1, 0, 8]()
+   cuda.synchronize()
+
+This time we declare 8 dynamic shared memory bytes, using the first 4 for a `float32` value and the next 4 for an `int32` value. Now we can set both the `int32` and `float32` value without them aliasing:
+
+.. code-block:: pycon
+
+   3.140000
+   1
+
+
 .. _cuda-local-memory:
 
 Local memory
--- a/docs/source/cuda/memory.rst
+++ b/docs/source/cuda/memory.rst
@@ -159,7 +159,8 @@ traditional dynamic memory management.
 Dynamic Shared Memory
 ---------------------
 
-In order to use dynamic shared memory in kernel code declare shared array of size 0:
+In order to use dynamic shared memory in kernel code declare a shared array of
+size 0:
 
 .. code-block:: python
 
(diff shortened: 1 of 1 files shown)


docs/source/cuda/memory.rst
```diff
@@ -156,6 +156,81 @@ traditional dynamic memory management.
 .. seealso::
    :ref:`Matrix multiplication example <cuda-matmul>`.
 
+Dynamic Shared Memory
+---------------------
+
+In order to use dynamic shared memory in kernel code declare shared array of size 0:
```
gmarkall, line 162:
```suggestion
In order to use dynamic shared memory in kernel code declare a shared array of size 0:
```

docs/source/cuda/memory.rst
```diff
@@ -156,6 +156,81 @@ traditional dynamic memory management.
...
+
+.. code-block:: python
+
+   kernel_func[32, 32, 0, 128](x)
+
+In above code dispatcher object is instantiated with 4 parameters:
+
+.. code-block:: python
+
+   kernel_func[grid_dim, block_dim, stream_id, dyn_shared_mem_size]
```
gmarkall, line 181: The third parameter should be 0 or a stream object, rather than the ID of a stream, so I think it would be more illustrative to just refer to it as `stream`. (Perhaps in future we should encourage people to use `cuda.default_stream()` instead and warn when `0` is passed to mean the default stream, which is a historical idea and not very ergonomic IMHO)

```suggestion
   kernel_func[grid_dim, block_dim, stream, dyn_shared_mem_size]
```

docs/source/cuda/memory.rst
```diff
@@ -156,6 +156,81 @@ traditional dynamic memory management.
...
+      f32_arr = cuda.shared.array(0, dtype=np.float32)
+      i32_arr = cuda.shared.array(0, dtype=np.int32)
+      f32_arr[0] = 3.14
+      print(f32_arr[0])
+      print(i32_arr[0])
+
+   f[1, 1, 0, 4]()
+   cuda.synchronize()
+
+This allocates 4 bytes of shared memory (large enough for one int32 or one float32) and declares dynamic shared memory arrays of type int32 and of type float32. When `f32_arr[0]` is set, this also sets the value of `i32_arr[0]`, because they're pointing at the same memory. So we see as output:
```
gmarkall, line 201:
```suggestion
This allocates 4 bytes of shared memory (large enough for one ``int32`` or one ``float32``) and declares dynamic shared memory arrays of type ``int32`` and of type ``float32``. When ``f32_arr[0]`` is set, this also sets the value of ``i32_arr[0]``, because they're pointing at the same memory. So we see as output:
```

docs/source/cuda/memory.rst
```diff
@@ -156,6 +156,81 @@ traditional dynamic memory management.
...
+
+.. code-block:: pycon
```
gmarkall, line 203: Looks like a popular conference is trying to sneak a little bit of self-promotion into the docs here :slightly_smiling_face: 

```suggestion
.. code-block:: python
```

docs/source/cuda/memory.rst
```diff
@@ -156,6 +156,81 @@ traditional dynamic memory management.
...
+      i32_arr = cuda.shared.array(0, dtype=np.int32)[1:] # 1 int32 = 4 bytes
+      f32_arr[0] = 3.14
+      i32_arr[0] = 1
+      print(f32_arr[0])
+      print(i32_arr[0])
+
+   f_with_view[1, 1, 0, 8]()
+   cuda.synchronize()
+
+This time we declare 8 dynamic shared memory bytes, using the first 4 for a `float32` value and the next 4 for an `int32` value. Now we can set both the `int32` and `float32` value without them aliasing:
```
gmarkall, line 226: Code formatting:

```suggestion
This time we declare 8 dynamic shared memory bytes, using the first 4 for a ``float32`` value and the next 4 for an ``int32`` value. Now we can set both the ``int32`` and ``float32`` value without them aliasing:
```

docs/source/cuda/memory.rst
```diff
@@ -156,6 +156,81 @@ traditional dynamic memory management.
...
+
+.. code-block:: pycon
```
gmarkall, line 228:
```suggestion
.. code-block:: python
```

docs/source/cuda/memory.rst
```diff
@@ -156,6 +156,81 @@ traditional dynamic memory management.
...
+   3.140000
+   1078523331
+
+because `1078523331` is the `int32` represented by the bits of the `float32` IEEE754 `3.14`.
+
+If we take disjoint views of the dynamic shared memory:
+
+.. code-block:: python
+
+   @cuda.jit
```
gmarkall, line 214: It would be good to add the imports at the top of this code block as in the previous one, so that the whole thing can easily be copy/pasted like the above one.

gmarkall (changes requested):
Many thanks for this! This is something I've been meaning to get round to for literally years but never quite managed it (instead answering the same question multiple times over the years!). I think this is looking generally great - I have made some suggestions on the diff, which are mostly about formatting and grammar / spelling, but a couple of more substantive suggestions too. In addition to the comments on the diff, the lines need wrapping at 80 characters as well - this isn't enforced by a style check in CI, but is the convention for the docs.

I have manually checked that the code snippets execute correctly and demonstrate the correct output. Ideally new code in the docs would be a doctest, but I'd suggest we just fix up the things referred to in the comments here and then could convert to a doctest in a future PR, so that we can get this over the line quickly.

Once you've made adjustments, please ping me and I can do another round of review!

docs/source/cuda/memory.rst
```diff
@@ -156,6 +156,81 @@ traditional dynamic memory management.
...
+In order to use dynamic shared memory in kernel code declare shared array of size 0:
+
+.. code-block:: python
+
+   @cuda.jit
+   def kernel_func(x):
+      dyn_arr = cuda.shared.array(0, dtype=np.float32)
+      ...
+
+and specify the size of dynamic shared array in bytes during kernel invokation (dispatcher object instantiation):
```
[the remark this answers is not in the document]
k1m190r: Got it thanks!

docs/source/cuda/memory.rst
```diff
@@ -156,6 +156,81 @@ traditional dynamic memory management.
...
+
+.. code-block:: python
+
+   kernel_func[32, 32, 0, 128](x)
+
+In above code dispatcher object is instantiated with 4 parameters:
```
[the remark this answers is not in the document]
k1m190r: changed.

docs/source/cuda/memory.rst
k1m190r: :) I thought it's a python console output formatter is more appropriate, will adjust as suggested.

gmarkall: Ah, I actually had no idea about this, and I'm sorry I was flippant about it. I've just learned about the python console outputter... I think in this case it makes no difference, but feel free to change this, and the other instance back to `pycon` - apologies for the mess-up!

docs/source/cuda/memory.rst
gmarkall: As with previous comment, please fee free to flip this back if you'd like.

docs/source/cuda/memory.rst
```diff
@@ -156,6 +156,94 @@ traditional dynamic memory management.
...
+size 0:
+
+.. code-block:: python
+
+   @cuda.jit
+   def kernel_func(x):
+      dyn_arr = cuda.shared.array(0, dtype=np.float32)
+      ...
+
+and specify the size of dynamic shared array in bytes during kernel invocation:
```
gmarkall, line 172: Sorry, I missed this last night - I'd suggest we're specifying the size of dynamic shared memory, rather than an array, to be consistent with the terminology we're generally using when talking about the declaration:
```suggestion
and specify the size of dynamic shared memory in bytes during kernel invocation:
```

k1m190r:
Changes committed. I've decided to  use `pycon` to convey meaning of the code-blocks. Though there is no  differences in appearance :).

Thanks for the review @gmarkall.

gmarkall (approved):
Looks great, many thanks for fixing this longstanding issue!

k1m190r:
Closed it accidently. Should I keep this open and it'll auto-close when merged?

gmarkall:
Yeah, it will close when merged - the current state of things is all good.

gmarkall:
I've linked this to #5025 so it will close too - I think this is sufficient documentation for dynamic shared memory, so we might as well have one less issue on the tracker :slightly_smiling_face:

PR #8842, opened 2023-03-23
Fix buildscripts, setup.py, docs for setuptools becoming optional.

stuartarchibald:
This builds on #8476.

As the setuptools package is moved to an optional runtime
dependency, various packaging, docs and testing infrastructure has
to be changed to reflect this. This patch contains the necessary
changes.


stuartarchibald:
fca421e squashes all the trial-and-error to involved in trying to find out how to run tests without `setuptools` being present.

sklam (approved):

PR #8843, opened 2023-03-23
Pin typeguard to 3.0.1 in AzureCI.

stuartarchibald:
As title. 3.0.2 seems to have a problem in `typeguard.importhook` when run against Numba.


sklam (commented):
Easy enough change. We should merge it as soon as build passes.

sklam (approved):

PR #8848, opened 2023-03-25
added lifted loops to glossary term

### LINKED ISSUE #8828: Explain what "lifted loops" are in the docs
## Feature request
I'm currently getting lots of warnings that look like:
```python
 Function "get_non_zero_ull_min" was compiled in object mode without forceobj=True, but has lifted loops.
```

I'm not sure what "lifted loops" means in this context though, and the docs don't seem to have anything on them. Searching for "lifted" just gives a couple of changelog entries: https://numba.readthedocs.io/en/stable/search.html?q=lifted&check_keywords=yes&area=default#

It would be great if someone who knows what's going on here could either improve the error message, or add some explanation about "lifted loops" to the docs

cherieliu:
This commit adds the term "lift loops" to the Numba glossary, with the same definitions as "loop-lifting" and "loop-jifting".

Resolves #8828

### DIFF
--- a/docs/source/glossary.rst
+++ b/docs/source/glossary.rst
@@ -33,6 +33,7 @@ Glossary
       Shorthand for "a function :term:`JIT-compiled <JIT>` with Numba using
       the :ref:`@jit <jit>` decorator."
 
+   lifted loops
    loop-lifting
    loop-jitting
       A feature of compilation in :term:`object mode` where a loop can be


stuartarchibald (approved):
Thanks for the patch! I've checked the build of the docs with this patch and searching for the term "lifted loops" now finds this entry in the glossary.

stuartarchibald:
@cherieliu congratulations on your first contribution to Numba!

PR #8854, opened 2023-03-28
Updated mk_alloc to support Numba-Dpex compute follows data.

mingjie-intel:
Change is made in mk_alloc to support Numba-Dpex compute follows data. There will be no affect to Numba main function.

### DIFF
--- a/numba/core/ir_utils.py
+++ b/numba/core/ir_utils.py
@@ -91,6 +91,20 @@ def mk_alloc(typingctx, typemap, calltypes, lhs, size_var, dtype, scope, loc,
             out.append(tuple_assign)
             size_var = tuple_var
             size_typ = types.containers.UniTuple(types.intp, ndims)
+    if hasattr(lhs_typ, "__allocate__"):
+        return lhs_typ.__allocate__(
+            typingctx,
+            typemap,
+            calltypes,
+            lhs,
+            size_var,
+            dtype,
+            scope,
+            loc,
+            lhs_typ,
+            size_typ,
+            out,
+        )
     # g_np_var = Global(numpy)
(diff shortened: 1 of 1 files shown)


DrTodd13 (changes requested):
Please add this __allocate__ support in the appropriate place in the Numba developer documentation.

diptorupd:
Some more motivating context behind these changes. The `mk_alloc` function is used inside the Numba parfor machinery to create a new empty NumPy array (by calling numpy.empty) whenever a temporary array is needed to allocate the left-hand-side sub-expression of a parfor node.

In our work on using parfors to support other NumPy-like libraries we want to change the actual library call used to allocate the empty array. These changes extend the `mk_alloc` function to first look if the array type provides an `__allocate__` function and use that instead of the default logic in the original function.

@DrTodd13 can you point me to the file where the documentation should be updated?

DrTodd13:
@diptorupd @sklam @stuartarchibald I don't know if the documentation belongs in numba/docs/source/extending or numba/docs/source/developer.  Unbelievably, we don't have a ParallelAccelerator page in the developer docs which was my first idea.  I'm soliciting ideas from Siu and Stuart as to where this should be documented.

sklam:
> @diptorupd @sklam @stuartarchibald I don't know if the documentation belongs in numba/docs/source/extending or numba/docs/source/developer. Unbelievably, we don't have a ParallelAccelerator page in the developer docs which was my first idea. I'm soliciting ideas from Siu and Stuart as to where this should be documented.

I think this is ParallelAccelerator specific enough to have its own page. I think we can add that ParallelAccelerator page in a different PR.

PR #8857, opened 2023-03-29
Fix broken np.allclose, add np.isclose

lericson:
It was previously impossible to use `np.allclose` as the default arguments `rtol`, `atol`, and `equal_nan` had types `float` and `bool`, whereas the code was testing for the Numba types `Float` and `Boolean`.

Also there was no `np.isclose`, the element-wise equivalent of `np.allclose`.

It is also debatable whether `allclose` should be an array method; it is not in numpy. I have not made `isclose` a method.

### DIFF
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -809,8 +809,20 @@ def np_all(a):
     return flat_all
 
 
+@overload(np.any)
+@overload_method(types.Array, "any")
+def np_any(a):
+    def flat_any(a):
+        for v in np.nditer(a):
+            if v.item():
+                return True
+        return False
+
+    return flat_any
+
+
 @register_jitable
-def _allclose_scalars(a_v, b_v, rtol=1e-05, atol=1e-08, equal_nan=False):
+def _isclose_scalars(a_v, b_v, rtol=1e-05, atol=1e-08, equal_nan=False):
     a_v_isnan = np.isnan(a_v)
     b_v_isnan = np.isnan(b_v)
 
@@ -845,15 +857,15 @@ def np_allclose(a, b, rtol=1e-05, atol=1e-08, equal_nan=False):
     if not type_can_asarray(b):
         raise TypeError('The second argument "b" must be array-like')
 
-    if not isinstance(rtol, types.Float):
+    if not isinstance(rtol, (types.Float, float)):
         raise TypeError('The third argument "rtol" must be a '
                         'floating point')
 
-    if not isinstance(atol, types.Float):
+    if not isinstance(atol, (types.Float, float)):
         raise TypingError('The fourth argument "atol" must be a '
                           'floating point')
 
-    if not isinstance(equal_nan, types.Boolean):
+    if not isinstance(equal_nan, (types.Boolean, bool)):
         raise TypeError('The fifth argument "equal_nan" must be a '
                         'boolean')
 
@@ -863,16 +875,16 @@ def np_allclose(a, b, rtol=1e-05, atol=1e-08, equal_nan=False):
     if is_a_scalar and is_b_scalar:
         def np_allclose_impl_scalar_scalar(a, b, rtol=1e-05, atol=1e-08,
                                            equal_nan=False):
-            return _allclose_scalars(a, b, rtol=rtol, atol=atol,
-                                     equal_nan=equal_nan)
+            return _isclose_scalars(a, b, rtol=rtol, atol=atol,
+                                    equal_nan=equal_nan)
         return np_allclose_impl_scalar_scalar
     elif is_a_scalar and not is_b_scalar:
         def np_allclose_impl_scalar_array(a, b, rtol=1e-05, atol=1e-08,
                                           equal_nan=False):
             b = np.asarray(b)
             for bv in np.nditer(b):
-                if not _allclose_scalars(a, bv.item(), rtol=rtol, atol=atol,
-                                         equal_nan=equal_nan):
+                if not _isclose_scalars(a, bv.item(), rtol=rtol, atol=atol,
+                                        equal_nan=equal_nan):
                     return False
             return True
         return np_allclose_impl_scalar_array
@@ -881,8 +893,8 @@ def np_allclose(a, b, rtol=1e-05, atol=1e-08, equal_nan=False):
                                           equal_nan=False):
             a = np.asarray(a)
             for av in np.nditer(a):
-                if not _allclose_scalars(av.item(), b, rtol=rtol, atol=atol,
-                                         equal_nan=equal_nan):
+                if not _isclose_scalars(av.item(), b, rtol=rtol, atol=atol,
+                                        equal_nan=equal_nan):
                     return False
             return True
         return np_allclose_impl_array_scalar
@@ -894,8 +906,8 @@ def np_allclose(a, b, rtol=1e-05, atol=1e-08, equal_nan=False):
             a_a, b_b = np.broadcast_arrays(a, b)
 
             for av, bv in np.nditer((a_a, b_b)):
-                if not _allclose_scalars(av.item(), bv.item(), rtol=rtol,
-                                         atol=atol, equal_nan=equal_nan):
+                if not _isclose_scalars(av.item(), bv.item(), rtol=rtol,
+                                        atol=atol, equal_nan=equal_nan):
                     return False
 
             return True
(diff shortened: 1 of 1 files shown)


stuartarchibald:
@lericson Many thanks for the patch. Please could you open an issue containing a minimal working reproducer for the bug(s) this is fixing and also please could appropriate tests be added for the proposed changes/bug fixes? Thanks again.

xref: #6286 which implements similar?

lericson:
Here is a reproduction…

```sh
lericson@strixy:~$ ~/.pyenv/versions/3.10.6/bin/python3 -m venv tmp_env
lericson@strixy:~$ cd tmp_env
lericson@strixy:~/tmp_env$ ./bin/pip install numba
…
Successfully installed llvmlite-0.39.1 numba-0.56.4 numpy-1.23.5
lericson@strixy:~/tmp_env$ ./bin/python3
Python 3.10.6 (main, Aug 22 2022, 12:13:43) [GCC 7.5.0] on linux
```

Scalar-scalar case

```python3
>>> from numba import njit
>>> @njit
... def f(a, b):
...   return np.allclose(a, b)
...
>>> import numpy as np
>>> f(0.0, 1.0)
  File "/home/lericson/tmp_env/lib/python3.10/site-packages/numba/core/dispatcher.py", line 468, in _compile_for_args
    error_rewrite(e, 'typing')
  File "/home/lericson/tmp_env/lib/python3.10/site-packages/numba/core/dispatcher.py", line 409, in error_rewrite
    raise e.with_traceback(None)
No implementation of function Function(<function allclose at 0x7f5bba179c60>) found for signature:

 >>> allclose(float64, float64)

  Overload in function 'np_allclose': File: numba/np/arraymath.py: Line 894.
    With argument(s): '(float64, float64)':
     TypeError: The third argument "rtol" must be a floating point
  raised from /home/lericson/tmp_env/lib/python3.10/site-packages/numba/np/arraymath.py:905

During: resolving callee type: Function(<function allclose at 0x7f5bba179c60>)
During: typing of call at <stdin> (3)


File "<stdin>", line 3:
```

Array-array case

```python3
>>> f(np.array([0.0]), np.array([1.0]))
  File "/home/lericson/tmp_env/lib/python3.10/site-packages/numba/core/dispatcher.py", line 468, in _compile_for_args
    error_rewrite(e, 'typing')
  File "/home/lericson/tmp_env/lib/python3.10/site-packages/numba/core/dispatcher.py", line 409, in error_rewrite
    raise e.with_traceback(None)
No implementation of function Function(<function allclose at 0x7f5bba179c60>) found for signature:

 >>> allclose(array(float64, 1d, C), array(float64, 1d, C))

    Overload in function 'np_allclose': File: numba/np/arraymath.py: Line 894.
      With argument(s): '(array(float64, 1d, C), array(float64, 1d, C))':
       TypeError: The third argument "rtol" must be a floating point
  raised from /home/lericson/tmp_env/lib/python3.10/site-packages/numba/np/arraymath.py:905

During: resolving callee type: Function(<function allclose at 0x7f5bba179c60>)
During: typing of call at <stdin> (3)


File "<stdin>", line 3:

>>>
```

lericson:
Note that the PR you xref'd seems to reimplement allclose also, maybe the PR is out of date.

Numba has np.allclose already, it just doesn't work at all.

stuartarchibald:
Having just updated the release notes for the upcoming 0.57 release, I noticed that `np.isclose` is already implemented via #7067.

stuartarchibald:
@lericson thanks for the reproducer. #8860 tracks.

numba/np/arraymath.py
```diff
@@ -894,25 +906,80 @@ def np_allclose_impl_array_array(a, b, rtol=1e-05, atol=1e-08,
...
-@overload(np.any)
-@overload_method(types.Array, "any")
-def np_any(a):
-    def flat_any(a):
-        for v in np.nditer(a):
-            if v.item():
-                return True
-        return False
+@overload(np.isclose)
+def np_isclose(a, b, rtol=1e-05, atol=1e-08, equal_nan=False):
```
guilhermeleobas, line 919: This file already contains an implementation of `np.isclose` in line 1191. Could you update the existing implementation? Or remove it in favor of this one?

lericson:
@guilhermeleobas I think the existing implementation is just as good, let's ignore this PR since the issue seems resolved. As for #8860, it is also likely fixed with the next release. Thanks for a great project!

lericson:
Actually on closer inspection, the code in master should be rewritten to reuse its `_isclose_item` between `allclose` and `isclose`. It may actually be preferable to use my code instead, as I use `np.nditer`, which doesn't accidentally densify 0-stride arrays. Let me demonstrate:

```python3
>>> import numpy as np
>>> a = np.arange(10)[:, None]
>>> b = np.arange(10)[None, :]
>>> ab, ba = np.broadcast_arrays(a, b)
```

The current implementation uses `ndarray.reshape(-1)`, which sometimes copies memory:

```python3
>>> ab.shape
(10, 10)
>>> ab.strides
(8, 0)
>>> ab_flat = ab.reshape(-1)
>>> ab_flat.shape
(100,)
>>> ab_flat.strides
(8,)
```
>>> ab.base.nbytes
80
>>> ab_flat.base.nbytes
800
```

**Comment by guilhermeleobas** [comment]
> It may actually be preferable to use my code instead, as I use `np.nditer`

Any chance you could replace the existing implementation by yours?

**Comment by lericson** [comment]
Certainly, just give me a little time- LudvigOn 30 Mar 2023, at 15:48, Guilherme Leobas ***@***.***> wrote:﻿

It may actually be preferable to use my code instead, as I use np.nditer

Any chance you could replace the existing implementation by yours?

**Comment by lericson** [comment]
I suggest deprecating `arr1.allclose(arr2)`, it's not in numpy and the existing `isclose` is not an array method either.

File: numba/np/arraymath.py [hunk 2]
```diff
@@ -894,25 +897,80 @@ def np_allclose_impl_array_array(a, b, rtol=1e-05, atol=1e-08,
...
+            a = np.asarray(a)
+            b = np.asarray(b)
+            a_a, b_b = np.broadcast_arrays(a, b)
+            c_c = np.empty(a_a.shape, dtype=np.bool_)
+
+            for av, bv, cv in np.nditer((a_a, b_b, c_c)):
+                cv[()] = _isclose_scalars(av.item(), bv.item(), rtol=rtol,
+                                          atol=atol, equal_nan=equal_nan)
+
+            return c_c
```
lericson, line 971: By the way, wouldn't this variant work for all array-scalar combinations?

lericson: In fact I think this works for scalar-scalar also. At least in non-numba Numpy.

lericson:
I tested it and this version uses way less code, and does exactly the same thing.

numba/np/arraymath.py
```diff
@@ -809,110 +809,56 @@ def flat_all(a):
...
-        for v in np.nditer(a):
-            if v.item():
-                return True
-        return False
 
-    return flat_any
+@overload(np.isclose)
+def np_isclose(a, b, rtol=1e-05, atol=1e-08, equal_nan=False):
+    def impl(a, b, rtol=1e-05, atol=1e-08, equal_nan=False):
+        ab, ba = np.broadcast_arrays(a, b)
```
guilhermeleobas, line 855: rather than calling `np.broadcast_arrays`, compute the final shape with `np.broadcast_shapes` and call `np.broadcast_arrays` in each array individually. This might solve the current CI issue.

guilhermeleobas (changes requested):
Hi @lericson, there's currently an error on CI regarding using `np.broadcast_arrays` with arrays with different dtype. This is currently not supported in Numba as it is not possible to infer the output list dtype.

numba/np/arraymath.py
lericson: I would also note that `np.nditer` doesn't work quite the same as the true numpy version; my use of it here seems to fail. I have to manually broadcast the arrays.

guilhermeleobas: > Hold on, why can't broadcast_arrays just return a Tuple(ArrayType1, ArrayType2)?

We follow NumPy spec where `np.broadcast_arrays` returns a list.

numba/np/arraymath.py
```diff
@@ -809,110 +809,65 @@ def flat_all(a):
...
-        raise TypeError('The third argument "rtol" must be a '
-                        'floating point')
-
-    if not isinstance(atol, types.Float):
-        raise TypingError('The fourth argument "atol" must be a '
-                          'floating point')
-
-    if not isinstance(equal_nan, types.Boolean):
-        raise TypeError('The fifth argument "equal_nan" must be a '
-                        'boolean')
```
guilhermeleobas, line 858: When `np.allclose` is used with unsupported types, these exceptions provide a useful error message. Could we keep them?

numba/np/arraymath.py
```diff
@@ -1109,60 +1064,6 @@ def impl(a):
...
-    elif np.isinf(x) or np.isinf(y):
-        return False
-    else:
-        return abs(x - y) <= atol + rtol * abs(y)
-
-
-@overload(np.isclose)
-def isclose(a, b, rtol=1e-05, atol=1e-08, equal_nan=False):
-    if not (type_can_asarray(a) and type_can_asarray(b)):
-        raise TypingError("Inputs for `np.isclose` must be array-like.")
```
guilhermeleobas, line 1127: Same thing here. Having the exception has its upsides.

guilhermeleobas (changes requested):
Hi @lericson, thanks for working on these changes.

Current changes are almost 2x slower (`np.allclose`) than previous code:

```bash
$ ./runtests.py -f -v numba.tests.test_np_functions -k allclose
test_allclose (numba.tests.test_np_functions.TestNPFunctions) ... ok
test_allclose_exception (numba.tests.test_np_functions.TestNPFunctions) ... ok
test_ip_allclose_numpy (numba.tests.test_np_functions.TestNPFunctions) ... ok
test_ip_not_allclose_numpy (numba.tests.test_np_functions.TestNPFunctions) ... ok

Ran 4 tests in 3.802s

$ time ./runtests.py -f -v numba.tests.test_np_functions -k isclose
test_isclose (numba.tests.test_np_functions.TestNPFunctions) ... /Users/guilhermeleobas/git/numba/numba/tests/test_np_functions.py:135: NumbaExperimentalFeatureWarning: Use of isinstance() detected. This is an experimental feature.
  return np.isclose(a, b, rtol, atol, equal_nan)
/Users/guilhermeleobas/git/numba/numba/tests/test_np_functions.py:135: NumbaExperimentalFeatureWarning: Use of isinstance() detected. This is an experimental feature.
  return np.isclose(a, b, rtol, atol, equal_nan)
ok

Ran 1 test in 4.077s
```

```bash
$ ./runtests.py -f -v numba.tests.test_np_functions -k allclose
test_allclose (numba.tests.test_np_functions.TestNPFunctions) ... /Users/guilhermeleobas/git/numba/numba/tests/test_np_functions.py:343: NumbaExperimentalFeatureWarning: Use of isinstance() detected. This is an experimental feature.
  return np.allclose(a, b, rtol, atol, equal_nan)
/Users/guilhermeleobas/git/numba/numba/tests/test_np_functions.py:343: NumbaExperimentalFeatureWarning: Use of isinstance() detected. This is an experimental feature.
  return np.allclose(a, b, rtol, atol, equal_nan)
ok
test_allclose_exception (numba.tests.test_np_functions.TestNPFunctions) ... ok
test_ip_allclose_numpy (numba.tests.test_np_functions.TestNPFunctions) ... /Users/guilhermeleobas/git/numba/numba/tests/test_np_functions.py:343: NumbaExperimentalFeatureWarning: Use of isinstance() detected. This is an experimental feature.
  return np.allclose(a, b, rtol, atol, equal_nan)
ok
test_ip_not_allclose_numpy (numba.tests.test_np_functions.TestNPFunctions) ... ok

Ran 4 tests in 7.379s

$ ./runtests.py -f -v numba.tests.test_np_functions -k isclose
test_isclose (numba.tests.test_np_functions.TestNPFunctions) ... /Users/guilhermeleobas/git/numba/numba/tests/test_np_functions.py:135: NumbaExperimentalFeatureWarning: Use of isinstance() detected. This is an experimental feature.
  return np.isclose(a, b, rtol, atol, equal_nan)
/Users/guilhermeleobas/git/numba/numba/tests/test_np_functions.py:135: NumbaExperimentalFeatureWarning: Use of isinstance() detected. This is an experimental feature.
  return np.isclose(a, b, rtol, atol, equal_nan)
/Users/guilhermeleobas/git/numba/numba/tests/test_np_functions.py:135: NumbaExperimentalFeatureWarning: Use of isinstance() detected. This is an experimental feature.
  return np.isclose(a, b, rtol, atol, equal_nan)
ok

Ran 1 test in 5.415s
```

It's not necessary to create and broadcast an array for scalar items, as it can be an expensive operation in terms of both compilation and execution. Could you investigate this a bit further?

guilhermeleobas:
@lericson could you add a couple of tests for the case you reported in your first comment?

> It was previously impossible to use np.allclose as the default arguments rtol, atol, and equal_nan had types float and bool, whereas the code was testing for the Numba types Float and Boolean.

lericson:
Good luck!

PR #8861, opened 2023-03-30
CUDA: Don't add device kwarg for jit registry

gmarkall:
The device kwarg has no effect - the need for a device function is inferred by logic in the dispatcher. Therefore, we no longer need to get in front of the `cuda.jit` decorator to insert it for the JIT registry.

### DIFF
--- a/numba/cuda/initialize.py
+++ b/numba/cuda/initialize.py
@@ -2,16 +2,12 @@ def initialize_all():
     # Import models to register them with the data model manager
     import numba.cuda.models  # noqa: F401
 
-    from numba import cuda
+    from numba.cuda.decorators import jit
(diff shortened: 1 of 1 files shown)


gmarkall:
gpuci run tests

stuartarchibald (approved):
Thanks for the patch. I think this makes sense as a CUDA-dispatcher-to-CUDA-dispatcher call would imply the callee is a device function and `CUDADispatcher::get_call_template()` handles this.

sklam:
BFID: `numba_smoketest_cuda_yaml_198`

stuartarchibald:
> BFID: `numba_smoketest_cuda_yaml_198`

Passed.

PR #8864, opened 2023-03-30
Remove obsolete deprecation notices

gmarkall:
- Eager compilation of device functions
- Removal of `add_user_function()`


stuartarchibald (approved):
Thanks for the patch!

PR #8870, opened 2023-03-31
Fix opcode "spelling" change since Python 3.11 in CUDA debug test.

### LINKED ISSUE #8867: Test `numba.cuda.tests.cudapy.test_debug.TestDebugOutput.test_dump_bytecode` fails on Python 3.11
## Reporting a bug

This:
```
./runtests.py numba.cuda.tests.cudapy.test_debug.TestDebugOutput.test_dump_bytecode
```
fails on Python 3.11. I think it's due to this:
https://github.com/numba/numba/blob/e7a1e8315d74b0f92ba8b26a40bbb1f28db94ba2/numba/cuda/tests/cudapy/test_debug.py#L49-L50
which is looking for a `BINARY_ADD` which in Python 3.11 is a `BINARY_OP` with the argument to the instruction determining the "OP" kind.

stuartarchibald:
As title.

Closes #8867

### DIFF
--- a/numba/cuda/tests/cudapy/test_debug.py
+++ b/numba/cuda/tests/cudapy/test_debug.py
@@ -1,5 +1,6 @@
 import numpy as np
 
+from numba.core.utils import PYVERSION
 from numba.cuda.testing import skip_on_cudasim, CUDATestCase
 from numba.tests.support import (override_config, captured_stderr,
                                  captured_stdout)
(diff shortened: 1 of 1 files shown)


numba/cuda/tests/cudapy/test_debug.py
```diff
@@ -47,7 +48,11 @@ def check_debug_output(self, out, enabled_dumps):
                 self.assertRaises(AssertionError, check_meth, out)
 
     def _check_dump_bytecode(self, out):
-        self.assertIn('BINARY_ADD', out)
+        if PYVERSION == (3, 11):
```
gmarkall, line 51: Are we expecting this to change back for 3.12? If not, perhaps
```suggestion
        if PYVERSION >= (3, 11):
```

will be more robust for the future?

gmarkall: Suggesting leaving it as-is following OOB discussion that this is in line with other similar changes. If it breaks for 3.12 I'd like to change this test to compare against the bytecode from `dis.dis()` anyway so it's Python-version-invariant.

gmarkall (approved):
LGTM!

numba/cuda/tests/cudapy/test_debug.py
stuartarchibald: From OOB discussion, electing to leave this as-is such that if it breaks again for 3.12 the test itself might need a change to the way in which it works.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_197`.

Passed, including Python 3.11 variants.

PR #8871, opened 2023-03-31
Don't return the function in CallConv.decorate_function()

gmarkall:
This function mutates its input function, and its return value is ignored (only called by `BaseContext.declare_function()`) - returning it is pointless and in some sense a returned value implies not mutating the input, so this pattern is a bit confusing.

This small refactor stops the return of the input function in order to tidy up this ambiguity.

### DIFF
--- a/numba/core/callconv.py
+++ b/numba/core/callconv.py
@@ -264,7 +264,6 @@ class MinimalCallConv(BaseCallConv):
         arginfo.assign_names(self.get_arguments(fn),
                              ['arg.' + a for a in args])
         fn.args[0].name = ".ret"
-        return fn
 
     def get_arguments(self, func):
         """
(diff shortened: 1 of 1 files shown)


gmarkall:
gpuci run tests

stuartarchibald:
> This function mutates its input function, and its return value is ignored (only called by `BaseContext.declare_function()`) - returning it is pointless and in some sense a returned value implies not mutating the input, so this pattern is a bit confusing.

Given the name of the function contains `decorate` I wondered if the original intent was to perhaps use it as a decorator, which might perhaps explain some part of returning the function. It seems like it was introduced in 86d30b71 and has changed very little since, and as far as I can tell, it's also never been used as a decorator. I'm left to conclude that this is just "decorating" the LLVM function with respect to adding attributes and names etc. I think the proposed change is good as is clears up the confusion in relation to returning a mutated input.

stuartarchibald (approved):
Thanks for the patch!

PR #8876, opened 2023-04-03
Force disabling of vfork in multiproc testing

sklam:


sklam:
BFID `numba_pynext_cpu_yaml_14` passed. So disabling vfork also make the segfault go away.

sklam:
we won't need this

Issue #5000, opened 2019-12-21
Closure inliner not handling keyword args correctly


stuartarchibald:
## Reporting a bug

- [x] I am using the ~latest released~ master version of Numba (most recent is visible in
 the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [x] I have included below a minimal working reproducer (if you are unsure how
 to write one see http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports).

This:
```python
from numba import njit
import numpy as np

@njit
def pennsylvania6():

    def foo(a, b, kw1=False):
        print(a, b, kw1)

    foo(1, 2, "x")
    foo(1, 2, kw1="y")
    foo(1, 2, kw1=True)


pennsylvania6()
print('-' * 80)
pennsylvania6.py_func()
```
produces:
```
1 2 x                                                                                                                           
1 2 False                                                                                                                       
1 2 False                                                                                                                       
--------------------------------------------------------------------------------                                                
1 2 x                                                                                                                           
1 2 y                                                                                                                           
1 2 True                                                                                                                        
```

suspect there's something wrong with the inliner/a missing guard against this.

PR #8877, opened 2023-04-03
Add support for keyword and default arguments during closure inlining.

danieldjohnson:
Currently, numba has only partial support for keyword and default arguments in closures. Specifically, when the closure object is a `make_function` op in the IR, the keyword arguments are silently ignored, and default arguments are appended to the argument list even when values for those arguments were actually provided. This can produce incorrect and surprising behavior when writing code that uses closures with keyword or default arguments.

This change reworks the `_get_callee_args` function to build a full Python signature for closures, and then use `fold_arguments` to handle them in the same way that arguments are handled for other Python functions.

To avoid reimplementing the existing `inspect` logic for determining the signature of a Python callable, this change constructs a new `_FakeFunction` object that has all the fields of a normal Python callable and then fetches the signature of that object. (It's not possible to directly construct the Python callable that the closure corresponds to, because we do not have concrete values for any closed-over variables yet.)

### DIFF
--- a/numba/core/inline_closurecall.py
+++ b/numba/core/inline_closurecall.py
@@ -2,7 +2,9 @@ import types as pytypes  # avoid confusion with numba.types
 import copy
 import ctypes
 import numba.core.analysis
-from numba.core import types, typing, errors, ir, rewrites, config, ir_utils
+from numba.core import (
+    types, typing, errors, ir, rewrites, config, ir_utils, utils
+)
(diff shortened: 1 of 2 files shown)


stuartarchibald:
Many thanks for the patch, I'll queue it for review. I think there's at least one ticket open about this, will try and find it to cross reference.

stuartarchibald:
xref #5000

PR #8879, opened 2023-04-04
Remove use of `compile_isolated` from generator tests.

### LINKED ISSUE #8873: Generator test segfault on linux py3.11
`test_gen9` or `test_gen9_objmode` in `numba.tests.test_generators` are either segfaulting or hanging and causing the multiproc testsuite to timeout. Problem only occur on Linux and py3.11. Problem has occurred on x64, ppc64le, and aarch64.

Issue first occurred with the commit 201a2f8, merge of https://github.com/numba/numba/pull/8134.

The issue is hard to reproduce. Only consistent reproduction is in the buildfarm and with the buildfarm created package. 

Scripts to reproduce:

```bash
$ conda create -y -q -n testenv_f8901e13 -c numba/label/pynext_ci -c numba/label/pynext_dev -c conda-forge python=3.11 numba=0.57.0dev0=np1.23py3.11h3611171_g201a2f8d6_1571 numpy=1.24 pyyaml colorama scipy jinja2 cffi ipython flake8
$ conda activate testenv_f8901e13
$ NUMBA_ENABLE_CUDASIM python -m numba.runtests -m 4 -v numba.tests
```

`conda-forge` channel is not necessary. Problem has occurred on py3.11 + np1.23 from conda defaults.

stuartarchibald:
As title. Use of `compile_isolated` leads to hard to track down life-time bugs such as that in #8873. Essentially the `CodeLibrary` containing the generator destructors was being free'd before the generators had actually had their destructors invoked such that the code that should have been executed was missing. This patch replaces `compile_isolated` with `jit` which has an easier to reason about lifetime behaviour for `CodeLibrary` instances etc.

Fixes #8873

### DIFF
--- a/numba/tests/test_generators.py
+++ b/numba/tests/test_generators.py
@@ -1,20 +1,13 @@
 import numpy as np
 
 import unittest
-from numba.core.compiler import compile_isolated, Flags
 from numba import jit, njit
 from numba.core import types
 from numba.tests.support import TestCase, MemoryLeakMixin
 from numba.core.datamodel.testing import test_factory
 
-
-enable_pyobj_flags = Flags()
-enable_pyobj_flags.enable_pyobject = True
-
-forceobj_flags = Flags()
-forceobj_flags.force_pyobject = True
-
-no_pyobj_flags = Flags()
+forceobj_flags = {'nopython': False, 'forceobj': True}
+nopython_flags = {'nopython': True}
 
 
 def make_consumer(gen_func):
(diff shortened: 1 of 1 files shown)


sklam:
BFID `numba_pynext_cpu_yaml_15`

gmarkall (dismissed):
This looks good to me:

 - The `tryanother.py` example from #8873 has not segfaulted once in 20 runs.
 - I increased the iteration count of `tryanother.py` to 2000 (an order of magnitude increase) and it still has not segfaulted after another 20 runs, though it does take much longer to run (obviously). Memory usage slowly climbs to about 2GB during each run, which doesn't alarm me given the artificial nature of the test and the fact it's compiling over and over.
 - There are no more instances of `compile_isolated` in `test_generators.py`.

There's one small suggestion on the diff where I think a name got altered, but it's just a very minor point - I think if this runs fine on the build farm then it's good for unblocking the release from my perspective.

stuartarchibald:
@gmarkall thanks for reviewing, the "minor point" should be addressed in 0dc45d9.

sklam (approved):
LGTM. Passed manual testing. Just waiting for buildfarm to finish.

sklam:
> BFID `numba_pynext_cpu_yaml_15`

passed

PR #8881, opened 2023-04-05
Replace use of compile_isolated in test_obj_lifetime

sklam:
For another instance of https://github.com/numba/numba/issues/8873 but this time in `numba.tests.test_obj_lifetime.TestObjLifetime.test_inf_loop_multiple_back_edge` and only appear in the buildfarm with osx-arm64 py3.8 npy1.21.

### DIFF
--- a/numba/tests/test_obj_lifetime.py
+++ b/numba/tests/test_obj_lifetime.py
@@ -6,7 +6,7 @@ import operator
 from itertools import takewhile
 
 import unittest
-from numba import njit
+from numba import njit, jit
 from numba.core.controlflow import CFGraph, Loop
 from numba.core.compiler import (compile_extra, compile_isolated, Flags,
                                  CompilerBase, DefaultPassBuilder)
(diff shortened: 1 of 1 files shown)


sklam:
BFID: `numba_yaml_332`

numba/tests/test_obj_lifetime.py
```diff
@@ -248,8 +248,10 @@ class TestObjLifetime(TestCase):
     """
 
     def compile(self, pyfunc):
-        cr = compile_isolated(pyfunc, (types.pyobject,), flags=forceobj_flags)
-        return cr.entry_point
+        # Note: looplift must be disabled. The test require the function
+        #       control-flow to be unchanged.
+        cfunc = jit((types.pyobject,), forceobj=True, looplift=False)(pyfunc)
```
stuartarchibald, line 253:
```suggestion
        cfunc = jit((types.pyobject,), nopython=False, forceobj=True,
                    looplift=False)(pyfunc)
```
else it raises a warning.

stuartarchibald (approved):
Thanks for the patch. I think this change is technically valid as a fix for a problem similar to that in #8873. In the interests of tagging a release candidate, given that this has been through testing on the build farm as-is, the couple of minor things in the comments **do not** need to be addressed in this PR (they have no material impact on the test validity) and can go into a subsequent patch to remove all use of `compile_isolated` .

stuartarchibald:
NOTE: Fixed typo in PR description.

gmarkall (approved):
Also LGTM modulo Stuart's comments that don't need to be addressed here.

I've also enjoyed the irony of an "object lifetime test" having an object lifetime issue :slightly_smiling_face:

stuartarchibald:
> BFID: `numba_yaml_332` and `numba_pynext_cpu_yaml_17`

These "passed" (a couple of jobs failed for known configuration/LLVM bug reasons).

Issue #8860, opened 2023-03-29
`np.allclose` implementation does not handle default arguments correctly.


stuartarchibald:
## Reporting a bug

- [x] I have tried using the latest released version of Numba (most recent is
 visible in the change log (https://github.com/numba/numba/blob/main/CHANGE_LOG).
- [x] I have included a self contained code sample to reproduce the problem.
  i.e. it's possible to run as 'python bug.py'.

The `np.allclose` implementation does not handle default arguments correctly.

MWR:
```python
from numba import njit
import numpy as np

@njit
def foo(a, b):
    np.allclose(a, b)

foo(1., 1.)
```
The implementation section here:
https://github.com/numba/numba/blob/9878790437b16718ac18cb38e9ee1a711b8a9017/numba/np/arraymath.py#L842-L858
Should:
1. Only be raising exceptions that derive from `numba.core.errors.NumbaError`, i.e. `TypingError`.
2. Accommodate both the Numba type system type and the Python type when checking the argument types. i.e. should be at least `if not isinstance(rtol, (float, types.Float)):`

guilhermeleobas:
> Only be raising exceptions that derive from numba.core.errors.NumbaError, i.e. TypingError.

Just FYI, there are other places that raise exceptions not deriving from `NumbaError`. I'll open an issue to track this.

cherieliu:
Has anyone worked on this? If not, I'm interested!

esc:
This was fixed in https://github.com/numba/numba/pull/8885

PR #8885, opened 2023-04-05
Fix np.allclose not handling default args

guilhermeleobas:
Fix issue #8860:
* Replace `TypeError` by `TypingError`
* Add a few tests to ensure exceptions are raised
* Add `float` class to `isinstance` comparison

Also, wondering if Numba should have the same checks on `np.isclose` implementation?

### DIFF
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -840,22 +840,22 @@ def _allclose_scalars(a_v, b_v, rtol=1e-05, atol=1e-08, equal_nan=False):
 def np_allclose(a, b, rtol=1e-05, atol=1e-08, equal_nan=False):
 
     if not type_can_asarray(a):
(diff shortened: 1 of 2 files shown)


kc611 (approved):
LGTM!

kc611:
> Also, wondering if Numba should have the same checks on np.isclose implementation?

Definitely would make sense to do so. `np.isclose` and `np.allclose` pretty much share a common input argument format so the error handling at type level can also be shared.

gmarkall:
@kc611 if this looks good, is it ready to be RTM? Or does it need an additional look?

kc611:
In it's current state it's RTM unless @guilhermeleobas wants to add the things he mentioned in the same PR (Although those ca be part of another PR as well.)

sklam (approved):

PR #8895, opened 2023-04-12
CUDA: Enable caching functions that use CG

### LINKED ISSUE #8888: Feature: support compilation caching for CUDA target
## Feature request

As mentioned here: https://numba.discourse.group/t/cuda-jit-is-there-cache-true/163/4

there is no caching for CUDA jit compilation. To confirm in most recent CUDA, when I try, I get:
```sh
    def _reduce_states(self):
        """
        Reduce the instance for serialization. We retain the PTX and cubins,
        but loaded functions are discarded. They are recreated when needed
        after deserialization.
        """
        if self._linking_files:
            msg = 'Cannot pickle CUDACodeLibrary with linking files'
>           raise RuntimeError(msg)
E           RuntimeError: Cannot pickle CUDACodeLibrary with linking files
```

Our test suite now has >60 kernels for unit testing device functions. Many are taking ~5-10 sec to compile, so this is becoming a big issue for us.

gmarkall:
Caching functions that use Cooperative Groups was blocked because the dispatcher adds a file to link, the `cudadevrt` cudalib, to the code library, and pickling code libraries with linking files is not supported.

To enable caching of functions that use cooperative groups, we push the linking of `cudadevrt` into the code library. We push it in here rather than special-casing the check for linked libraries because a recompile of a cached function (e.g. for a different compute capability) might require `cudadevrt` from a different location to the one found during the initial compile; locating the library at the cubin linking time ensures the correct version will be used.

Fixes #8888.

### DIFF
--- a/buildscripts/incremental/setup_conda_environment.sh
+++ b/buildscripts/incremental/setup_conda_environment.sh
@@ -80,8 +80,8 @@ if [ $PYTHON \< "3.9" ]; then $CONDA_INSTALL importlib_metadata; fi
 # Install dependencies for building the documentation
 if [ "$BUILD_DOC" == "yes" ]; then $CONDA_INSTALL sphinx docutils sphinx_rtd_theme pygments numpydoc; fi
 if [ "$BUILD_DOC" == "yes" ]; then $PIP_INSTALL rstcheck; fi
-# Install dependencies for code coverage (codecov.io)
-if [ "$RUN_COVERAGE" == "yes" ]; then $PIP_INSTALL codecov; fi
+# Install dependencies for code coverage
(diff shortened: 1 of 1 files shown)


gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

guilhermeleobas (approved):
LGTM. Tried the patch locally and works as expected.

shaunc:
Not in 0.57.1, @sklam  🥺 ?  It is arguably a bug fix.... 

I was hoping to be able to include in production and depending on a patch branch is dicey.

shaunc:
Hooray!

PR #8905, opened 2023-04-18
FIX: Add missing `**targetoptions` to `jit` in `GUFuncBuilder`

### LINKED ISSUE #8903: 0.57.0rc1 NumbaDeprecationWarning: `@guvectorize` even with `nopython` keyword argument
The following code (note the `nopython=True` argument)

```py

@guvectorize([(int64[:], int64, int64[:])], '(n),()->(n)', nopython=True)
def g(x, y, res):
    for i in range(x.shape[0]):
        res[i] = x[i] + y
```

raises `NumbaDeprecationWarning`:

`/Applications/anaconda3/envs/py311/lib/python3.11/site-packages/numba/np/ufunc/ufuncbuilder.py:328: NumbaDeprecationWarning: The 'nopython' keyword argument was not supplied to the 'numba.jit' decorator. The implicit default value for this argument is currently False, but it will be changed to True in Numba 0.59.0. See https://numba.readthedocs.io/en/stable/reference/deprecation.html#deprecation-of-object-mode-fall-back-behaviour-when-using-jit for details.
  self.nb_func = jit(_target='npyufunc', cache=cache)(py_func)`

This does not occur with `@vectorize`:

```py
from numba import vectorize, float64

@vectorize([float64(float64, float64)], nopython=True)
def f(x, y):
    return x + y
```

```
>>> numba.__version__
'0.57.0rc1'
```

oyamad:
Fix #8903

* Add missing `**targetoptions` to `self.nb_func = jit(...)` in `GUFuncBuilder.__init__`, as in the corresponding part in `UFuncBuilder.__init__`: https://github.com/numba/numba/blob/320a9f533201c642ef6fd01baf7b3376bd4daa9b/numba/np/ufunc/ufuncbuilder.py#L263-L265

### DIFF
--- a/numba/np/ufunc/ufuncbuilder.py
+++ b/numba/np/ufunc/ufuncbuilder.py
@@ -325,7 +325,9 @@ class GUFuncBuilder(_BaseUFuncBuilder):
                  targetoptions={}, writable_args=()):
         self.py_func = py_func
         self.identity = parse_identity(identity)
-        self.nb_func = jit(_target='npyufunc', cache=cache)(py_func)
+        self.nb_func = jit(_target='npyufunc',
+                           cache=cache,
+                           **targetoptions)(py_func)
         self.signature = signature
         self.sin, self.sout = parse_signature(signature)
         self.targetoptions = targetoptions


guilhermeleobas:
Closing this as it was fixed in https://github.com/numba/numba/issues/8928

Issue #8898, opened 2023-04-13
numba 0.57.0rc1 segfaults: missing ldexpf for np.exp2 on Windows


sklam:
from https://github.com/numba/numba/issues/8841#issuecomment-1506280834:

Original message:

I am finding that numba 0.57.0rc1 segfaults on Windows when installed with pip or conda for the following:

```python
import numpy as np
from numba import njit

f = njit(lambda x: np.exp2(x))
f(np.int8(2))
```
I create the environments via
```
conda create -n py310pip -c conda-forge python=3.10
pip install numba==0.57.0rc1
```
and
```
conda create -n py310conda -c numba -c conda-forge python=3.10 numba
```
I have not found any other function for which this occurs.

I discovered this when testing [`python-graphblas`](https://github.com/python-graphblas/python-graphblas/issues/431). Here's an example traceback:
<details>
<summary>LLVM ERROR: Symbol not found: ldexpf</summary>

```
LLVM ERROR: Symbol not found: ldexpf
Fatal Python error: Aborted

Current thread 0x00000bb0 (most recent call first):
  File "C:\Miniconda3\envs\graphblas\lib\site-packages\llvmlite\binding\ffi.py", line 152 in __call__
  File "C:\Miniconda3\envs\graphblas\lib\site-packages\llvmlite\binding\executionengine.py", line 92 in finalize_object
  File "C:\Miniconda3\envs\graphblas\lib\site-packages\numba\core\codegen.py", line 1060 in wrapper
  File "C:\Miniconda3\envs\graphblas\lib\site-packages\numba\core\codegen.py", line 999 in _finalize_specific
  File "C:\Miniconda3\envs\graphblas\lib\site-packages\numba\core\codegen.py", line 797 in _finalize_final_module
  File "C:\Miniconda3\envs\graphblas\lib\site-packages\numba\core\codegen.py", line 765 in finalize
  File "C:\Miniconda3\envs\graphblas\lib\site-packages\numba\core\codegen.py", line 567 in _ensure_finalized
  File "C:\Miniconda3\envs\graphblas\lib\site-packages\numba\core\codegen.py", line 989 in get_pointer_to_function
  File "C:\Miniconda3\envs\graphblas\lib\site-packages\numba\core\ccallback.py", line 75 in compile
  File "C:\Miniconda3\envs\graphblas\lib\site-packages\numba\core\compiler_lock.py", line 35 in _acquire_compile_lock
[...]

PR #8907, opened 2023-04-18
Work around issue #8898. Defer `exp2` (and `log2`) calls to Numba internal symbols.

stuartarchibald:
As title.

### DIFF
--- a/numba/np/npyfuncs.py
+++ b/numba/np/npyfuncs.py
@@ -648,11 +648,13 @@ def np_complex_exp_impl(context, builder, sig, args):
 def np_real_exp2_impl(context, builder, sig, args):
     _check_arity_and_homogeneity(sig, args, 1)
 
-    ll_ty = args[0].type
-    fnty = llvmlite.ir.FunctionType(ll_ty, [ll_ty,])
-    fn = cgutils.insert_pure_function(builder.module, fnty,
-                                      name='llvm.exp2')
-    return builder.call(fn, [args[0]])
+    dispatch_table = {
+        types.float32: 'numba_exp2f',
+        types.float64: 'numba_exp2',
+    }
+
+    return _dispatch_func_by_name_type(context, builder, sig, args,
+                                       dispatch_table, 'exp2')
 
 
 def np_complex_exp2_impl(context, builder, sig, args):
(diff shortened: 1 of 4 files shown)


numba/np/npyfuncs.py
```diff
@@ -648,11 +648,13 @@ def np_complex_exp_impl(context, builder, sig, args):
...
-    fn = cgutils.insert_pure_function(builder.module, fnty,
-                                      name='llvm.exp2')
-    return builder.call(fn, [args[0]])
+    dispatch_table = {
+        types.float32: 'numba_exp2f',
+        types.float64: 'numba_exp2',
+    }
+
+    return _dispatch_func_by_name_type(context, builder, sig, args,
+                                       dispatch_table, 'exp2')
```
sklam, line 657: Should we split the impl depending on the platform? Are we disabling potential optimization (vectorization) on non-windows?

numba/np/npyfuncs.py
```diff
@@ -685,11 +687,13 @@ def np_complex_log_impl(context, builder, sig, args):
...
-    fn = cgutils.insert_pure_function(builder.module, fnty,
-                                      name='llvm.log2')
-    return builder.call(fn, [args[0]])
+    dispatch_table = {
+        types.float32: 'numba_log2f',
+        types.float64: 'numba_log2',
+    }
+
+    return _dispatch_func_by_name_type(context, builder, sig, args,
+                                       dispatch_table, 'log2')
```
sklam, line 696: same split impl question

sklam (commented):
I've tested the changes and it works. The only question is whether we should split the impl so we only changes windows.

numba/np/npyfuncs.py
stuartarchibald: I think this will potentially limit optimisation as it generates 'unknown' library calls. However, with a "minimising risk for 0.57.0 release" view, I think perhaps it should be left as written as it's essentially doing the same as what was shipped in 0.56.4 (deferred symbol calls). Am happy to consider views from other folks that are interested, the patch is small and the changes to accommodate a split implementation are also small.

sklam: Ok, minimal risk is a good reason

sklam (approved):
Thanks for the patch

PR #8909, opened 2023-04-19
Fix #8903. `NumbaDeprecationWarning`s raised from `@{gu,}vectorize`.

### LINKED ISSUE #8903: 0.57.0rc1 NumbaDeprecationWarning: `@guvectorize` even with `nopython` keyword argument

stuartarchibald:
This adds localised suppressions for `NumbaDeprecationWarning`s that would have been raised from the code path for `@{gu,}vectorize` due to Numba's internal use of `@jit()` with the `nopython` kwarg not present.

This is considered acceptable due to the following: Deprecation warnings from Numba's internal use of its own API should not be visible to users. Unfortunately there's no trivial programmatic way to work around the internal use of `@jit`-with-no-kwargs in this case as numerous code paths go through the same operations and user code may well be relying on "object mode" fall-back. What this patch does it makes it so that users will not see `NumbaDeprecationWarning`s for this internal use. However, if a user defines a `@{gu,}vectorize` function that ends up using the fall-back to the object mode pipeline during compilation, Numba will continue to raise numerous loud warnings about this happening which should provide a hint as to the problem and how to fix it.

Numerous tests have been added to check combinations of kwargs/use of default kwargs being passed through to `@{gu,}vectorize`. These tests make sure that there are no additional `NumbaDeprecationWarning`s raised to do with Numba's internal use of `@jit()` in `@{gu,}vectorize` where the `nopython` kwarg is not supplied.

Closes #8903

### DIFF
--- a/numba/np/ufunc/dufunc.py
+++ b/numba/np/ufunc/dufunc.py
@@ -81,9 +81,10 @@ class DUFunc(serialize.ReduceMixin, _internal._DUFunc):
     def __init__(self, py_func, identity=None, cache=False, targetoptions={}):
         if is_jitted(py_func):
             py_func = py_func.py_func
-        dispatcher = jit(_target='npyufunc',
-                         cache=cache,
-                         **targetoptions)(py_func)
+        with ufuncbuilder._suppress_deprecation_warning_nopython_not_supplied():
+            dispatcher = jit(_target='npyufunc',
+                             cache=cache,
+                             **targetoptions)(py_func)
         self._initialize(dispatcher, identity)
         functools.update_wrapper(self, py_func)
 
--- a/numba/np/ufunc/ufuncbuilder.py
+++ b/numba/np/ufunc/ufuncbuilder.py
@@ -1,12 +1,14 @@
 # -*- coding: utf-8 -*-
 
 import inspect
+import warnings
 from contextlib import contextmanager
 
 from numba.core import config, targetconfig
 from numba.core.decorators import jit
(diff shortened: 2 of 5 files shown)


stuartarchibald:
Note that 0d1969a fixes `flake8` and also fixes an issue picked up in CI where the `pycc` deprecation tests need `setuptools` to run, but there's no guarantee of it being present, so the tests are now conditionally skipped.

PR #8911, opened 2023-04-21
Remove isinstance experimental feature warning

guilhermeleobas:
As title.


stuartarchibald:
Thanks for the patch. I think the docs also need an update, here:
https://github.com/numba/numba/blob/d44573b43dec9a7b66e9a0d24ef8db94c3dc346c/docs/source/reference/pysupported.rst?plain=1#L920-L922
and here:
https://github.com/numba/numba/blob/d44573b43dec9a7b66e9a0d24ef8db94c3dc346c/docs/source/reference/pysupported.rst?plain=1#L939

guilhermeleobas:
@stuartarchibald, done!

numba/tests/test_builtins.py
```diff
@@ -1387,7 +1387,7 @@ def foo(x):
             self.assertEqual(got, expected)
 
     @TestCase.run_test_in_subprocess
-    def test_experimental_warning(self):
+    def test_no_experimental_warning(self):
```
stuartarchibald, line 1390: Now CI has run and this change has been effectively confirmed... what do you think to just deleting this whole test? I don't think there's a need to check that there is no warning raised here any more than in any other piece of code where no warning is expected?

guilhermeleobas: Done!

guilhermeleobas: Oops, my bad!

guilhermeleobas:
@gmarkall, could you take another look at this PR once you have some free cycles to spare?

gmarkall (commented):
The changes look good to me, but should this also undo the changes from #8947 which ought to be un-needed now?

guilhermeleobas:
Thanks for the review

PR #8918, opened 2023-04-25
Make the Azure CI runtests testing run using multiprocessing.

stuartarchibald:
As title, aim is to speed up the runtests check.

### DIFF
--- a/buildscripts/incremental/test.cmd
+++ b/buildscripts/incremental/test.cmd
@@ -18,7 +18,7 @@ set PYTHONFAULTHANDLER=1
 set NUMBA_CAPTURED_ERRORS=new_style
 
 @rem First check that the test discovery works
-python -m numba.tests.test_runtests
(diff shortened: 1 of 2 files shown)


stuartarchibald:
CI is failing, this is because the `runtests` test isn't using Numba's test runner but rather running the test module itself directly through it's `__main__` (hence flags like `-m` don't work). Needs discussion.

stuartarchibald:
Conclusion from discussion at the maintainer meeting was that the "test `runtests`" part of the CI testing can be massively shortened to essentially just check that:
* the test suite reports it is at least X tests in size
* that at least one thing can be compiled and executed correctly.

This should reduce the "test `runtests`" run time from ~90s to probably nearer 10s.

stuartarchibald:
Closing, replaced by #8927, which implements https://github.com/numba/numba/pull/8918#issuecomment-1523761168.

PR #8924, opened 2023-04-26
Change llvmlite version pinning.

sklam:
To prevent picking up next version's dev build.


sklam:
BFID ``numba_smoketest_cpu_yaml_174``

stuartarchibald (approved):
Thanks for the patch. As noted in https://github.com/numba/numba/pull/8925#pullrequestreview-1402624617 which updates the related docs, this change is consistent with the logic in prior releases.

stuartarchibald:
> BFID `numba_smoketest_cpu_yaml_174`

Spot checks on a few builds suggest this used `llvmlite-0.40.0rc1-<pyversion+hash>_<build number>`. The builds jobs largely passed, just a couple of known failures.

PR #8925, opened 2023-04-26
Update release checklist template

sklam:
Changes pinning of llvmlite to avoid picking up the next version's dev build.

Relates to https://github.com/numba/numba/pull/8924


stuartarchibald (approved):
Thanks for the patch. I think this updates the instructions to be consistent with what has happened in practice (at least for the last few releases). xref:
* `release0.56` branch f789aab
* `release0.55` branch faff3dc

PR #8927, opened 2023-04-26
Add a Numba power-on-self-test script and use in CI.

stuartarchibald:
This replaces the much broader "test of runtests" check performed in CI with a more simple check aiming to ensure that "runtests" works and Numba can correctly compile and execute at least one thing. The intent of this change is to drastically reduce the run time spent on this check.

### DIFF
--- a/buildscripts/incremental/test.cmd
+++ b/buildscripts/incremental/test.cmd
@@ -17,8 +17,10 @@ set PYTHONFAULTHANDLER=1
 @rem enable new style error handling
 set NUMBA_CAPTURED_ERRORS=new_style
 
-@rem First check that the test discovery works
-python -m numba.tests.test_runtests
+@rem First run Numba's Power-On-Self-Test to make sure testing will likely work
+python -m numba.misc.POST
+if %errorlevel% neq 0 exit /b %errorlevel%
+
 @rem Now run the Numba test suite
 @rem Note that coverage is run from the checkout dir to match the "source"
 @rem directive in .coveragerc
--- a/numba/misc/POST.py
+++ b/numba/misc/POST.py
@@ -1,0 +1,34 @@
+""" Numba's POWER ON SELF TEST script. Used by CI to check:
+0. That Numba imports ok!
+1. That Numba can find an appropriate number of its own tests to run.
+2. That Numba can manage to correctly compile and execute at least one thing.
+"""
+from numba.tests import test_runtests
+from numba import njit
+
+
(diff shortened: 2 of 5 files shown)


numba/misc/POST.py
```diff
@@ -0,0 +1,34 @@
...
+        raise AssertionError(msg)
+
+
+def check():
+    _check_runtests()
+    _check_cpu_compilation()
+
+
+if __name__ == "__main__":
+    check()
```
stuartarchibald, line 34: Maybe this should print a status?

stuartarchibald:
For run time comparison, it's probably best to look at the Windows builds as they have an isolated "Verify runtests" step on `main` which is the comparable step to the `Power-On-Self-Test...` step in this branch. The windows build machines are at present always skylake-avx512 chips with 4 cores available, so CPU differences are not present as they are in e.g. the linux build machines. Results:
* `main` (at 4f16b8f) build in Azure CI has step "Verify runtests" taking 2m 16s.
* this branch (at da12f2c) build in Azure CI has step "Power-On-Self-Test..." taking 9s.

Essentially, this patch makes it so that the actual testing of a change for e.g. a PR build will now start 2 minutes earlier (on windows at least)!

stuartarchibald:
Marking this for review so as to get some feedback.

buildscripts/incremental/test.cmd
```diff
@@ -17,8 +17,10 @@ set PYTHONFAULTHANDLER=1
 @rem enable new style error handling
 set NUMBA_CAPTURED_ERRORS=new_style
 
-@rem First check that the test discovery works
-python -m numba.tests.test_runtests
+@rem First run Numba's Power-On-Self-Test to make sure testing will likely work
```
sklam, line 20: Is this file used?

numba/misc/POST.py
sklam: nah, the traceback is probably good enough

buildscripts/incremental/test.cmd
sklam: The `test.cmd` doesn’t seem to be referenced anywhere. But, i’ll ignore it for the context of this PR and open a different PR to try to remove it.

stuartarchibald:
This is continued in #9191, it removes the changes to `test.cmd` so that it can be merged alongside #9155, squashes the patch series and rebases on `main` to accommodate changes in testing due to the RVSDG work.

PR #8929, opened 2023-04-27
Update CHANGE_LOG for 0.57.0 final.

stuartarchibald:
As title. Also notes release date in version support table.


sklam (approved):
Manually verified docs.

PR #8930, opened 2023-04-28
Fix year in change log

jtilly:
I just stumbled across this typo.

Same issue on the 0.57 release branch (https://github.com/numba/numba/blob/release0.57/CHANGE_LOG).


stuartarchibald:
Thanks for catching this typo @jtilly (and for fixing it). @sklam has carried the patch into #8932 to update the change log with this patch included and added you to the authors list, #8932 is approved for merge and this PR should end up as "merged" once that is merged.

PR #8948, opened 2023-05-08
Fix issue #8939

### LINKED ISSUE #8939: Regression with `literal_unroll` in 0.57.0
The following code used to work on `0.56.4`, but fails on the new version `0.57.0` (installed via `pip` on macOS if that matters):

```python
from numba import literal_unroll, njit
import numpy as np

test = np.array([(1, 2), (2, 3)], dtype=[("a1", "f8"), ("a2", "f8")])
fields = tuple(test.dtype.fields.keys())

@njit
def f(arr):
    res = 0
    for k in literal_unroll(fields):
        res = res + np.abs(arr[k]).sum()
    return res


print(f(test))
```

The script is supposed to print `8.0`, but it raised the following error in the newest version:

```
'vLiteral[str](a1)__28binary_subscr_7' is not a valid parameter name

File "literal_unroll.py", line 11:
def f(arr):
    <source elided>
    for k in literal_unroll(fields):
        res = res + np.abs(arr[k]).sum()
        ^

During: lowering "vLiteral[str](a1)_$30call_method.8 = arrayexpr(expr=(<ufunc 'absolute'>, [Var(vLiteral[str](a1)_$28binary_subscr.7, literal_unroll.py:11)]), ty=array(float64, 1d, C))" at [...]/literal_unroll.py (11)
```

guilhermeleobas:
Fixes #8939: Legalize var name when defining a new variable in the scope.

### DIFF
--- a/numba/core/untyped_passes.py
+++ b/numba/core/untyped_passes.py
@@ -906,8 +906,12 @@ class MixedContainerUnroller(FunctionPass):
             new_var_dict = {}
             for name, var in var_table.items():
                 scope = switch_ir.blocks[lbl].scope
-                new_var_dict[name] = scope.define(
-                    f"v{branch_ty}_{name}", var.loc).name
+                if isinstance(branch_ty, types.Literal):
+                    new_var_dict[name] = scope.define(
+                        f"v{branch_ty.literal_value}_{name}", var.loc).name
+                else:
+                    new_var_dict[name] = scope.define(
+                        f"v{branch_ty}_{name}", var.loc).name
             replace_var_names(loop_blocks, new_var_dict)
 
             # clobber the sentinel body and then stuff in the rest
--- a/numba/tests/test_mixed_tuple_unroller.py
+++ b/numba/tests/test_mixed_tuple_unroller.py
@@ -1830,6 +1830,20 @@ class TestMore(TestCase):
             while x == 1:
                 x = 0
 
+    def test_literal_unroll_legalize_var_names(self):
+        # See issue #8939
+        test = np.array([(1, 2), (2, 3)], dtype=[("a1", "f8"), ("a2", "f8")])
+        fields = tuple(test.dtype.fields.keys())
+
+        @njit
+        def foo(arr):
+            res = 0
+            for k in literal_unroll(fields):
+                res = res + np.abs(arr[k]).sum()
+            return res
+
+        self.assertEqual(foo(test), 8.0)
+
 
 def capture(real_pass):
     """ Returns a compiler pass that captures the mutation state reported


esc:
@gmarkall I updated the labels on this one and wanted to ask: is it good to merge now that a test has been added?

gmarkall:
@esc I think it's ready for review now that a test has been added - I was just explaining my labelling decision as part of FPOC duties above.

esc:
I'm picking this one up now.

esc (approved):
I have tested this fix locally and it does indeed resolve the reproducer submitted as part of #8939. I also tried to change the variant of the reproducer included with the test-suite locally to ensure that it does indeed run as part of the test-suite. I saw a failure locally, which indicates the included reproducer is valid and it means that the test is good. 

👍

esc:
@stuartarchibald suggests to take a closer look, I'm moving this back into `waiting on reviewer`.

esc:
Taking a closer look here, it seems like this may not suffice yet.

Using the following test-input:

```
test = np.array([(1, 2), (2, 3)], dtype=[("[a1", "f8"), ("a2", "f8")])
```

I.e. inclusion of a square bracket. We get following issue:

```
'v[a1__28binary_subscr_7' is not a valid parameter name

File "issue_8939.py", line 13:
def f(arr):
    <source elided>
    for k in literal_unroll(fields):
        res = res + np.abs(arr[k]).sum()
        ^

During: lowering "v[a1_$30call_method.8 = arrayexpr(expr=(<ufunc 'absolute'>, [Var(v[a1_$28binary_subscr.7, issue_8939.py:13)]), ty=array(float64, 1d, C))" at /Users/vhaenel/git/numba/issue_8939.py (13)
```

esc (changes requested):
I think this'll need a "better" fix after all.

esc:
As an alternative, using `main` at dd082ca03169382ac7d2ffd7d4711a5d485ccce2 with the revert of https://github.com/numba/numba/commit/c6a25069deabe982bbc589bc98c4290808f8a4e4 

```diff
diff --git i/numba/core/untyped_passes.py w/numba/core/untyped_passes.py
index f76b2d1b79..d563207cd1 100644
--- i/numba/core/untyped_passes.py
+++ w/numba/core/untyped_passes.py
@@ -906,8 +906,14 @@ class MixedContainerUnroller(FunctionPass):
             new_var_dict = {}
             for name, var in var_table.items():
                 scope = switch_ir.blocks[lbl].scope
-                new_var_dict[name] = scope.define(
-                    f"v{branch_ty}_{name}", var.loc).name
+                try:
+                    scope.get_exact(name)
+                except errors.NotDefinedError:
+                    # is this correct? In case the scope doesn't have the
+                    # variable, we need to define it prior creating new
+                    # copies of it!
+                    scope.define(name, var.loc)
+                new_var_dict[name] = scope.redefine(name, var.loc).name
             replace_var_names(loop_blocks, new_var_dict)

             # clobber the sentinel body and then stuff in the rest
```

And the reproducer:

```python
from numba import literal_unroll, njit
import numpy as np

test = np.array([(1, 2), (2, 3)], dtype=[("a1[0]", "f8"), ("a2", "f8")])
fields = tuple(test.dtype.fields.keys())


@njit
def f(arr):
    res = 0
    for k in literal_unroll(fields):
        res = res + np.abs(arr[k]).sum()
    return res


print(f(test))
```

Does yield the correct answer.

esc:
> As an alternative, using `main` at [dd082ca](https://github.com/numba/numba/commit/dd082ca03169382ac7d2ffd7d4711a5d485ccce2) with the revert of [c6a2506](https://github.com/numba/numba/commit/c6a25069deabe982bbc589bc98c4290808f8a4e4)


I ran the Numba test-suite locally, and didn't see any failures with this revert.

esc:
Removing this from the milestone 0.57.1 again as it has been superseded by: #8998

Issue #8940, opened 2023-05-04
Regression: NameError: name 'unsafe_empty_inferred' is not defined


sk1p:
## Reporting a bug

- [x] I have tried using the latest released version of Numba (most recent is
 visible in the change log (https://github.com/numba/numba/blob/main/CHANGE_LOG).
- [x] I have included a self contained code sample to reproduce the problem.
  i.e. it's possible to run as 'python bug.py'.

The following reproducer crashes with numba 0.57.0:

```python
import numba
import numpy as np

# uncomment one of these lines to work around the problem:
# import numba.np.unsafe.ndarray
# import numba.cuda  # indirectly imports numba.np.unsafe.ndarray

@numba.njit()
def _fn():
    np.array([1 for _ in range(1)])

_fn()
```

Backtrace: 

```
Traceback (most recent call last):
  File "/home/alex/numba-debug/minmin.py", line 12, in <module>
    _fn()
  File "/home/alex/source/numba/numba/core/dispatcher.py", line 468, in _compile_for_args
    error_rewrite(e, 'typing')
  File "/home/alex/source/numba/numba/core/dispatcher.py", line 407, in error_rewrite
    raise e
  File "/home/alex/source/numba/numba/core/dispatcher.py", line 420, in _compile_for_args
    return_val = self.compile(tuple(argtypes))
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alex/source/numba/numba/core/dispatcher.py", line 965, in compile
    cres = self._compiler.compile(args, return_type)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alex/source/numba/numba/core/dispatcher.py", line 129, in compile
    raise retval
  File "/home/alex/source/numba/numba/core/dispatcher.py", line 139, in _compile_cached
    retval = self._compile_core(args, return_type)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alex/source/numba/numba/core/dispatcher.py", line 152, in _compile_core
    cres = compiler.compile_extra(self.targetdescr.typing_context,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alex/source/numba/numba/core/compiler.py", line 742, in compile_extra
    return pipeline.compile_extra(func)
[...]

PR #8958, opened 2023-05-15
resurrect the import, this time in the registry initialization

esc:
As title.

Fixes: #8940


guilhermeleobas (approved):
Confirm the patch fixes the issue reported in #8940

esc:
> Confirm the patch fixes the issue reported in #8940

Does this still need a test, I am not sure. @guilhermeleobas if not can you move this to RTM, please, thank you!

Issue #8960, opened 2023-05-15
Jupyter/Ipython Magic tools for Numba


Matt711:
## Feature request
Hi, I'm working with @gmarkall as an intern on the Numba team at NVIDIA this summer. As a part of my internship, I would like to build some jupyter magic tools for Numba. I'd like to use this issue for discussion and to track my progress.

1. The first, is a annotation command similar to what Cython does that could display different parts of Numba's compilation pipeline. For example, calling `%%numba --annotate` could default to providing the LLVM. We could also display the control flow graph (CFG) and for the CUDA target, the PTX / SASS. There's already support for providing the LLVM in an html file, but I'd like to extend it in the way I've described. I also would like to change how it's displayed in Jupyter (some CSS stuff mostly).
2. In addition, I'd also like to provide Numba users more insight into the type inference process. I could imagine if they use the magic command `%%numba --annotate --cfg --debug`, it could print out the IR and inferred types at every step of the type inference process. Essentially providing the user a better way to visualize the output when `NUMBA_DEBUG_TYPEINFER` is set. For example, the user could see
- What got updated at each step. Let the user navigate between the steps using the CFG of the IR.
- Why certain variables take particular types - for example, signed + unsigned int64 operations can result in a float64 - being able to pinpoint the step at which this occurs can help the user understand what happened.
- Why a unification error occurs - what two branches produced the typings for a value that then needs to be unified but can't. 
3. I could also work on other Jupyter magic commands for Numba related the the compilation pipeline. How else could I extend this?

I hope this is the right place for this. If not I'm happy to move it. Thanks for any advice or discussion.

gmarkall:
I just saw the stale bot closed this: https://github.com/numba/numba/pull/4453 - (not sure if it's any source of info / inspiration for this work)

Matt711:
> I just saw the stale bot closed this: #4453 - (not sure if it's any source of info / inspiration for this work)

I didn't know about that issue, but it's definitely related. I'll look into it more when I start the CFG stuff.

stuartarchibald:
Maybe also take a look at the `.inspect_*` methods on the Numba dispatcher (base) class:
https://github.com/numba/numba/blob/e4ff3cf5fcb59e91fc7d46c340f2e7eb664dd0c0/numba/core/dispatcher.py#L237
there might be some things that could easily be wired into jupyter "magic" tools?

PR #8962, opened 2023-05-15
Jupyter magic command for displaying Numba IR

Matt711:
As a part of #8960, I'd like to add a Jupyter magic command that displays the LLVM of the jupyter cell. This PR implements the first magic command `%%numba --annotate`. It works by writing the text to of the Jupyter cell to an IPython cache directory, importing, and then passing the "jitted" function to a `numba.core.annotations.Annotate` object which is returned.


Matt711:
image

numba/__init__.py
```diff
@@ -246,3 +246,9 @@ def _try_enable_svml():
 # SVML state to "no SVML". See https://github.com/numba/numba/issues/4689 for
 # context.
 # ---------------------- WARNING WARNING WARNING ----------------------------
+
+
+def load_ipython_extension(ip):
```
gmarkall, line 251: Is this to support `%load_ext numba`? If so, could you add a comment mentioning this, for the benefit of readers unfamiliar with how magics work? (They are new to me, and I imagine others might also benefit from the explanation)

gmarkall (commented):
Thanks for sharing progress with this! I made some initial high level comments on the PR. I need also to learn / think a bit more about magics - my main concern is over the robustness of writing out a temp file to annotate (whether we will have trouble capturing all that's required) but we can discuss / explore this a bit more outside this PR before another round of review.

numba/core/annotations/IpythonMagic.py
```diff
@@ -0,0 +1,85 @@
...
+import importlib.util
+import inspect
+
+from IPython.core import magic_arguments
+from IPython.core.magic import Magics, magics_class, cell_magic
+from IPython.paths import get_ipython_cache_dir
+from numba.core.annotations.pretty_annotate import Annotate
+
+
+@magics_class
```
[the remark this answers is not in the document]
Matt711: I'm not sure we need to handle anything as the default error is clear. You should get a error like below if you use cell magic as line magic.
```
In [3]: %numba --annotate
UsageError: Line magic function `%numba` not found (But cell magic `%%numba` exists, did you mean that instead?).
```

gmarkall:
**Note to self:** The next step reviewing here is to assess how safe the approach of writing out a temporary source file is; whether there are cases in which it could break, or if there's a way to get at the source code without writing to disk. One thing to look at for inspiration might be the Jupyter cache locator, to see how that handles code in cells of notebooks.

@stuartarchibald / @siu do you have any thoughts / suggestions off the top of your head about getting access to the source of functions in notebook cells?

numba/core/annotations/ipython_magic.py
```diff
@@ -0,0 +1,85 @@
...
+        cell = cell.strip()
+        return cell
+
+    def _get_function_names(self, cell):
+        return re.findall(r"def (.*)\(", cell)[0]
+
+    def _get_jitted_functions(self, module, function_names):
+        members = inspect.getmembers(module)
+        functions = [
+            member[1] for member in members if hasattr(member[1], "py_func")
```
gmarkall, line 56: I think the canonical way to check whether a function is jitted would be to see if it's an instance of `numba.core.dispatcher.Dispatcher`.

numba/core/annotations/ipython_magic.py
```diff
@@ -0,0 +1,85 @@
...
+            spec.loader.exec_module(numba_module)
+            self._import_all(numba_module)
+
+            numba_functions = self._get_jitted_functions(
+                numba_module, function_names
+            )
+            ann = Annotate(numba_functions[0])
+            return ann
+
+    def _preprocess_cell(self, cell):
```
gmarkall, line 46: Having a function just to strip the cell seems a bit heavyweight - perhaps this could just be rolled into the `_build_module_spec()` function.

numba/core/annotations/ipython_magic.py
```diff
@@ -0,0 +1,85 @@
...
+        cell = cell.strip()
+        return cell
+
+    def _get_function_names(self, cell):
+        return re.findall(r"def (.*)\(", cell)[0]
+
```
gmarkall, line 52: From suggestion above that this may not be needed:

```suggestion
```

numba/core/annotations/ipython_magic.py
```diff
@@ -0,0 +1,85 @@
...
+        functions = [
+            member[1] for member in members if hasattr(member[1], "py_func")
+        ]
+        return functions
+
+    def _build_module_spec(self, line, code):
+        lib_dir = os.path.join(get_ipython_cache_dir(), 'numba')
+        key = (code, line)
+        if not os.path.exists(lib_dir):
+            os.makedirs(lib_dir)
```
gmarkall, line 64: Do we need to do something to gracefully handle the case where creating the directory fails?

numba/core/annotations/ipython_magic.py
```diff
@@ -0,0 +1,85 @@
...
+        module_name = "_numba_magic_" + hashlib.sha1(
+            str(key).encode('utf-8')).hexdigest()
+        numba_file = os.path.join(lib_dir, module_name + '.py')
+        with open(numba_file, 'w', encoding='utf-8') as f:
+            f.write(code)
```
gmarkall, line 69: What if opening the file fails? E.g. if the `lib_dir` is a file that already exists - do we need to handle that gracefully too?

numba/core/annotations/ipython_magic.py
```diff
@@ -0,0 +1,85 @@
...
+        """
+        args = magic_arguments.parse_argstring(self.numba, line)
+        if args.annotate:
+            function_names = self._get_function_names(cell)
+            code = self._preprocess_cell(cell)
+            spec = self._build_module_spec(line, code)
+
+            # Import the module
+            numba_module = importlib.util.module_from_spec(spec)
+            spec.loader.exec_module(numba_module)
```
gmarkall, line 37: This seems to execute the module in its own namespace - What happens if it refers to globals in the notebook that aren't defined in the module?

numba/core/annotations/ipython_magic.py
```diff
@@ -0,0 +1,85 @@
...
+            f.write(code)
+        spec = importlib.util.spec_from_file_location(module_name, numba_file)
+        return spec
+
+    def _import_all(self, module):
+        mdict = module.__dict__
+        if '__all__' in mdict:
+            keys = mdict['__all__']
+        else:
+            keys = [k for k in mdict if not k.startswith('_')]
```
gmarkall, line 78: Similarly, isn't this presenting anything with an underscore prefix being added to the namespace of the notebook? Do we want to prevent that?

gmarkall (changes requested):
I've taken another look at this. My present thinking is that writing to a temp file so that the source can be found is an OK approach - I can't see any serious robustness issues with it, with the caveat this is outside my area of experience / expertise.

However, I think the way in which we execute the imported module in its own namespace seems likely to cause issues. I think the general expectation is that code in the cell runs in the same namespace as the rest of the notebook, so isolating its namespace might stop it accessing globals from elsewhere in the notebook, and also make it hard to ensure everything from its namespace gets into the notebook's namespace. (Again with the caveat that this is my understanding, but I'm not an expert here 🙂 )

numba/core/annotations/ipython_magic.py
```diff
@@ -0,0 +1,85 @@
...
+
+            # Import the module
+            numba_module = importlib.util.module_from_spec(spec)
+            spec.loader.exec_module(numba_module)
+            self._import_all(numba_module)
+
+            numba_functions = self._get_jitted_functions(
+                numba_module, function_names
+            )
+            ann = Annotate(numba_functions[0])
```
gmarkall, line 43: Noted during discussion that this produces an annotation for only the first jitted function.

gmarkall: (Due to potential corruption of output when displaying a lot in the notebook)

numba/tests/test_ipython_magic.py
```diff
@@ -0,0 +1,22 @@
+from unittest import TestCase
+from IPython.terminal.interactiveshell import TerminalInteractiveShell
+
+code = u"""\
+import numba
+
+@numba.njit
+def f(x):
```
gmarkall, line 8: Note from discussion: we should also add a test where something defined outside the cell is referred to.

Issue #8218, opened 2022-07-04
CUDA: Numerical differences between CPython / CPU target and the CUDA target in `math.pow`


gmarkall:
Originally noticed [on Discourse](https://numba.discourse.group/t/some-trouble-of-using-cuda-kernel-function-to-calculate-heap-sort/1421).

The following:

```python
from numba import cuda, njit
import numpy as np
import math


def inner(x, y):
    return int(math.pow(x, y))


cpu_jitted = njit(inner)


def cuda_jitted(x, y):
    inner_device = cuda.jit(device=True)(inner)

    @cuda.jit
    def outer(r, x, y):
        r[()] = inner_device(x[()], y[()])

    r_arr = np.array(0, dtype=np.int64)
    x_arr = np.array(0, dtype=np.int64)
    y_arr = np.array(0, dtype=np.int64)
    x_arr[()] = x
    y_arr[()] = y
    outer[1, 1](r_arr, x_arr, y_arr)

    return r_arr[()]


x = 3
y = 1

print(inner(x, y))
print(cpu_jitted(x, y))
print(cuda_jitted(x, y))
```

prints

```
3
3
2
```

This can cause some surprise.

gmarkall:
Similar occurs in CUDA C / C++:

```cuda
// Compile with:
// nvcc -gencode arch=compute_75,code=sm_75 -std c++17 mini.cu -o mini
//
// produces:
//
// 3
// 2

#include <iostream>
#include <cstdint>

template<typename T>
__global__ void kernel(T* r, T* x, T* y)
{
  r[0] = std::pow(x[0], y[0]);
}

template <typename T>
T host(T x, T y)
{
  return std::pow(x, y);
}

__managed__ uint32_t r_32;
__managed__ uint32_t x_32;
__managed__ uint32_t y_32;

int main(int argc, char **argv)
{
  x_32 = 3;
  y_32 = 1;
  kernel<<<1, 1>>>(&r_32, &x_32, &y_32);
  cudaDeviceSynchronize();
  std::cout << host(x_32, y_32) << std::endl;
  std::cout << r_32 << std::endl;
}
```

gmarkall:
The CUDA target doesn't end up using `int_power_impl`, but the CPU target does. This will likely need some adjustments in the typing / lowering for `math.pow` for CUDA, but because of the mix of implementations (those from the CPU target, `impl_pow_int` and `impl_binary` in `mathimpl.py`) means that it will need a bit of careful thinking about / testing.

[...]

Issue #8169, opened 2022-06-17
x**(-1) is zero when x is integer


ruoyu0088:
Version: Numba 0.55.1

The result of f() and f_jit() is different:

```
import numba

def f(x):
    return x**(-1)

f_jit = numba.jit(f)
f(2), f_jit(2)
```

output:

```
(0.5, 0)
```

guilhermeleobas:
Hi @ruoyu0088, thanks for the report, I can reproduce the error.

This bug happens due to a combination of two things. First, `a ^ (-b) = (1 / a) ^ b`, results in a float value, but Numba types `math.pow` as `(a: int, b: int) -> int`. Second, at the impl. function, when `b < 0`, Numba truncates the result of `a ^ (-b)` to the nearest integer, `0`.

https://github.com/numba/numba/blob/57ce5528af436d5269dd4aa34f24b58a2063cef2/numba/core/typing/builtins.py#L251-L253

https://github.com/numba/numba/blob/57ce5528af436d5269dd4aa34f24b58a2063cef2/numba/cpython/numbers.py#L218-L230

If you know the exponent is `< 0`, I suggest changing the code to do the division first to avoid this kind of problem.

I am going to close this issue as it is a duplicate of https://github.com/numba/numba/issues/7012 which tracks the same issue.

Issue #7012, opened 2021-05-10
pow() producing incorrect results for negative integer exponent 


stuartarchibald:
## Reporting a bug

- [x] I have tried using the latest released version of Numba (most recent is
 visible in the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [x] I have included a self contained code sample to reproduce the problem.
  i.e. it's possible to run as 'python bug.py'.

```python
from numba import njit

@njit
def foo(x):
    print(pow(2, x))

foo(-2)
foo.py_func(-2)
```
produces:
```
0
0.25
```

from a quick look at the source, not sure that this is quite right
https://github.com/numba/numba/blob/51978769c319b50404ac2ea972aa1e8c0aa9f0a4/numba/cpython/numbers.py#L230-L231 also, am not convinced this is valid in the case of a negative exponent https://github.com/numba/numba/blob/51978769c319b50404ac2ea972aa1e8c0aa9f0a4/numba/cpython/numbers.py#L212

stuartarchibald:
Original reference: https://gitter.im/numba/numba?at=6096bfdad261cc4d74b64283

guilhermeleobas:
Hum...isn't this code right? I mean, the function is called `int_power_impl` and the docstring says the following:
```
    a ^ b, where a is an integer or real, and b an integer
```

If b is a negative value, then, `a^b` is equivalent to `1 / a^b`, and the result is between `0 < (1 / a^b) < 1`. Given the return type is already defined to be an integral type, it is not possible to return `1 / (a ^ b)` without truncating the result.

stuartarchibald:
I think the fundamental issue is that the answer is incorrect in the case presented in the OP, part of the issue is indeed that the return type is defined as an integral type.

goxeq:
I also encountered the reported problem, although not with the `pow()` but with the `**` operator.  I consider, it would be much better to raise ValueError for negative exponent as numpy does. (i.e. replace [this](https://github.com/numba/numba/blob/f04f4b164887e5aa5ffeeed1bef317e13ec995f8/numba/cpython/numbers.py#L230) `return 0`  with something similar to `raise ValueError("Integers to negative integer powers are not allowed.")` Otherwise it silently causes incorrect results, which are not always caught in testing.

```python
import numba; njit=numba.njit; import numpy as np
def exponent(b):
    b=2**(48-b[0])
    return b
l=np.array([49,])

print(f"{numba.__version__=}")
try:
[...]

PR #8963, opened 2023-05-16
Fix math.pow with int64 argument for CUDA

testhound:
This pull request addresses an issue on the CUDA target where a [u]int64 argument to math.pow can produce incorrect results. The issue is detailed in this [bug report](https://github.com/numba/numba/issues/8218).

### DIFF
--- a/numba/cuda/mathimpl.py
+++ b/numba/cuda/mathimpl.py
@@ -331,55 +331,15 @@ def impl_pow_int(ty, libfunc):
 
     lower(math.pow, ty, types.int32)(lower_pow_impl_int)
 
-# Copied from cpython/numbers.py
-# Copied to eliminate test case failure in
-# test_import.py
-#
-# Also copied and modified to remove exception
-# raising which is not supported in CUDA.
+def register_int_power_impls():
+    # int_power_impl is used because libdevice.powi has slightly different
+    # semantics to math.pow. See Issue #8218 
+    from numba.cpython.numbers import int_power_impl
+    def impl_pow_int64(ty, int_ty):
+        lower(math.pow, ty, int_ty)(int_power_impl)
 
-
-def int_power_impl(context, builder, sig, args):
-    """
-    a ^ b, where a is an integer or real, and b an integer
-    """
-    is_integer = isinstance(sig.args[0], types.Integer)
-    tp = sig.return_type
-
-    def int_power(a, b):
-        # Ensure computations are done with a large enough width
-        r = tp(1)
-        a = tp(a)
-        if b < 0:
-            invert = True
-            exp = -b
-            if exp < 0:
-                return np.nan
-            if is_integer:
-                if a == 0:
-                    return -1 << (tp.bitwidth - 1)
-                if a != 1 and a != -1:
-                    return 0
-        else:
-            invert = False
-            exp = b
-        if exp > 0x10000:
-            # Optimization cutoff: fallback on the generic algorithm
-            return math.pow(a, float(b))
-        while exp != 0:
-            if exp & 1:
-                r *= a
-            exp >>= 1
-            a *= a
-
-        return 1.0 / r if invert else r
-
-    res = context.compile_internal(builder, int_power, sig, args)
-    return res
-
-
-def impl_pow_int64(ty, int_ty):
-    lower(math.pow, ty, int_ty)(int_power_impl)
+    impl_pow_int64(types.float64, int64)
+    impl_pow_int64(types.float64, uint64)
 
 
 impl_pow_int(types.float32, libdevice.powif)
--- a/numba/cuda/mathimpl.py
+++ b/numba/cuda/mathimpl.py
@@ -334,7 +334,6 @@ def impl_pow_int(ty, libfunc):
 
 def impl_pow_int64(ty, int_ty):
     def lower_pow_impl_int64(context, builder, sig, args):
-        print("In lower_pow_impl_int64")
         return int_power_impl(context, builder, sig, args)
 
     lower(math.pow, ty, int_ty)(lower_pow_impl_int64)
--- a/numba/cuda/mathimpl.py
+++ b/numba/cuda/mathimpl.py
@@ -3,10 +3,10 @@ import operator
 from llvmlite import ir
 from numba.core import types, typing, cgutils, targetconfig
 from numba.core.imputils import Registry
-from numba.cpython.numbers import int_power_impl
 from numba.types import float32, float64, int64, uint64
 from numba.cuda import libdevice
 from numba import cuda
+import numpy as np
 
 registry = Registry()
 lower = registry.lower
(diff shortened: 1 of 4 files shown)


testhound:
gpuci run tests

numba/cuda/mathimpl.py
```diff
@@ -329,9 +331,61 @@ def lower_pow_impl_int(context, builder, sig, args):
...
+            a *= a
+
+        return 1.0 / r if invert else r
+
+    res = context.compile_internal(builder, int_power, sig, args)
+    return res
+
+
+def impl_pow_int64(ty, int_ty):
+    lower(math.pow, ty, int_ty)(int_power_impl)
```
gmarkall, line 382: The test failure in `test_import.py` should be addressed by deferring the import of `numba.cpython.numbers` until the CUDA target is initialized, not by copy/pasting code from it. This can be done by:

```suggestion
def register_int_power_impls():
    # int_power_impl is used because libdevice.powi has slightly different
    # semantics to math.pow. See Issue #8218 
    from numba.cpython.numbers import int_power_impl
    def impl_pow_int64(ty, int_ty):
        lower(math.pow, ty, int_ty)(int_power_impl)

    impl_pow_int64(types.float64, int64)
    impl_pow_int64(types.float64, uint64)
```
and then calling `register_int_power_impls()` in `numba.cuda.target.CUDATargetContext.load_additional_registries()`.

On the topic of exceptions - they are supported in the CUDA target, but are ignored unless `debug=True`.

numba/cuda/tests/cudapy/test_powi.py
```diff
@@ -61,6 +71,32 @@ def test_powi(self):
...
+        kernel = dec(cu_mat_power_v2)
+
+        A = np.arange(10, dtype=np.float64).reshape(2, 5)
+        Aout = np.empty_like(A)
+
+        for power in range(1, 11):
+            power = np.int64(power)
+            kernel[1, A.shape](A, power, Aout)
+            result = np.allclose(Aout, A ** power)
+            self.assertTrue(result, f"Failed on int64 power {power}")
```
gmarkall, line 85: * A combination of `np.allclose` and `self.assertTrue` is odd - the `np.testing` module contains functions such as `assert_allclose()` etc that report the numerical issue detected.
* If you want to report the test data for which a test failed, you can use `self.subTest()` (as in other uses in the test suite) rather than adding your own message.
* Using `allclose()` for a test of integer arithmetic seems surprising - why not check for exact equality?

numba/cuda/tests/cudapy/test_powi.py
```diff
@@ -61,6 +71,32 @@ def test_powi(self):
         kernel[1, A.shape](A, power, Aout)
         self.assertTrue(np.allclose(Aout, A ** power))
 
+    def test_powi_int64(self):
+        dec = cuda.jit(void(float64[:, :], int64, float64[:, :]))
+        kernel = dec(cu_mat_power_v2)
+
+        A = np.arange(10, dtype=np.float64).reshape(2, 5)
```
gmarkall, line 78: Why use `float64` data for what (appears to be / I would expect to be) to be a test of integer functionality?

gmarkall (changes requested):
Thanks for the PR - there are some comments on the diff. In addition, I note that the new tests always pass on `main`, so they don't seem to be testing changes in this PR. I would suggest using the reproducer from #8218 as a starting point for adding tests.

numba/cuda/tests/cudapy/test_powi.py
testhound: @gmarkall I used float64 because the typing for math.pow in cudaimpl.py the first argument is either a floating point value or will be converted to a floating point value. In particular the test case that is failing is a float64 first argument and a int64 second argument. I wanted to isolate the exact typing for this test case.

gmarkall:
Noting also this will cause the CUDA target to be affected by #7012, which might be a regression in itself.

numba/cuda/tests/cudapy/test_powi.py
gmarkall: Can you show me the test case that is failing? I don't see a use of `float64` in #8218, and this test case seems not to fail on `main`, so I'm not sure which test case you're referring to.

testhound:
gpuci run tests

testhound:
gpuci run tests

testhound:
@gmarkall I incorporated your suggestions on not copying the integer pow implementation and using assert_allclose. Waiting on feedback on the float64 vs int64 question.

gmarkall:
Noting also we should consider whether the issue described in #8169 could have an effect here,

gmarkall (changes requested):
Thanks for the updates. I'm still finding that the added tests pass on `main`, so they're not testing that anything gets fixed with these changes. Using `test_powi.py` from this branch with `main`, I get:

```
0 gmarkall@gatnash 12:42:02 (main) ~/numbadev/numba
$ python runtests.py numba.cuda.tests.cudapy.test_powi -v
test_cpow_complex128_binop (numba.cuda.tests.cudapy.test_powi.TestCudaPowi) ... ok
test_cpow_complex128_inplace_binop (numba.cuda.tests.cudapy.test_powi.TestCudaPowi) ... ok
test_cpow_complex128_pow (numba.cuda.tests.cudapy.test_powi.TestCudaPowi) ... ok
test_cpow_complex64_binop (numba.cuda.tests.cudapy.test_powi.TestCudaPowi) ... ok
test_cpow_complex64_inplace_binop (numba.cuda.tests.cudapy.test_powi.TestCudaPowi) ... ok
test_cpow_complex64_pow (numba.cuda.tests.cudapy.test_powi.TestCudaPowi) ... ok
test_powi (numba.cuda.tests.cudapy.test_powi.TestCudaPowi) ... ok
test_powi_binop (numba.cuda.tests.cudapy.test_powi.TestCudaPowi) ... oktest_powi_int64 (numba.cuda.tests.cudapy.test_powi.TestCudaPowi) ... ok
test_powi_uint64 (numba.cuda.tests.cudapy.test_powi.TestCudaPowi) ... ok

Ran 10 tests in 2.020s

OK
```

We need a test that fails on `main` and passes with the other changes here, otherwise we're not verifying that this PR fixes anything.

Additionally, we should consider issues #7012 and #8169 here, since there's a risk that by fixing one numerical issue in the CUDA target we bring in another. Here's what I've been using to test things out locally:

```diff
diff --git a/numba/cuda/tests/cudapy/test_powi.py b/numba/cuda/tests/cudapy/test_powi.py
index 42c5fff1e..a0f81259b 100644
--- a/numba/cuda/tests/cudapy/test_powi.py
+++ b/numba/cuda/tests/cudapy/test_powi.py
@@ -99,6 +99,36 @@ class TestCudaPowi(CUDATestCase):
         kernel[1, A.shape](A, power, Aout)
         self.assertTrue(np.allclose(Aout, A ** power))
 
+    def test_powi_issue_7012_math_pow(self):
+        @cuda.jit
+        def f(r, x):
+            r[0] = math.pow(2, x)
+
+        r = np.zeros(1, dtype=np.float64)
+        x = -2
+        f[1, 1](r, x)
+        self.assertEqual(r[0], math.pow(2, x))
+
+    def test_powi_issue_7012_pow_operator(self):
+        @cuda.jit
+        def f(r, x):
+            r[0] = 2 ** x
+
+        r = np.zeros(1, dtype=np.float64)
+        x = -2
+        f[1, 1](r, x)
+        self.assertEqual(r[0], math.pow(2, x))
+
+    def test_powi_issue_8169(self):
+        @cuda.jit
+        def f(r, x):
+            r[0] = 2 ** x
+
+        r = np.zeros(1, dtype=np.float64)
+        x = -1
+        f[1, 1](r, x)
+        self.assertEqual(r[0], math.pow(2, x))
+
     # Relative tolerance kwarg is provided because 1.0e-7 (the default for
     # assert_allclose) is a bit tight for single precision.
     def _test_cpow(self, dtype, func, rtol=1.0e-7):
```

Here's what I'm finding with these tests with this branch at present:

```
test_powi_issue_7012_math_pow (numba.cuda.tests.cudapy.test_powi.TestCudaPowi) ... ok
test_powi_issue_7012_pow_operator (numba.cuda.tests.cudapy.test_powi.TestCudaPowi) ... FAIL
test_powi_issue_8169 (numba.cuda.tests.cudapy.test_powi.TestCudaPowi) ... FAIL

FAIL: test_powi_issue_7012_pow_operator (numba.cuda.tests.cudapy.test_powi.TestCudaPowi)
  File "/home/gmarkall/numbadev/numba/numba/cuda/tests/cudapy/test_powi.py", line 120, in test_powi_issue_7012_pow_operator
    self.assertEqual(r[0], math.pow(2, x))
AssertionError: 0.0 != 0.25

FAIL: test_powi_issue_8169 (numba.cuda.tests.cudapy.test_powi.TestCudaPowi)
  File "/home/gmarkall/numbadev/numba/numba/cuda/tests/cudapy/test_powi.py", line 130, in test_powi_issue_8169
    self.assertEqual(r[0], math.pow(2, x))
AssertionError: 0.0 != 0.5
```

Perhaps this behaviour is already present and these are not regressions - if that's the case then it would be OK for the failing tests to be marked as `xfail`. However, if they are regressions in this branch, that would need to be addressed too.

To clarify my earlier suggestion - my preference is for exact equality checking, not that you use `allclose` - this is because the original report exhibits surprising behaviour for the user due to the result of the operation in the reproducer being very close to, but just below 3. I would still like to see a test case that looks like the reproducer from that issue.

The discussion on #7012 goes over the discrepancy and challenge with typing and numerical accuracy here - that the type returned by `math.pow` is integral or float depending on the input values in Python, but that's not easy behaviour to follow in Numba.

gmarkall:
Closing as per discussion in the dev meeting on 2023-07-04: https://github.com/numba/numba/wiki/Minutes_2023_07_04

Summary: this still requires more work on the tests and doesn't address other issues that may be introduced by the fix, so it'd be better to close this rather than attempt to pick it up from here.

Issue #8951, opened 2023-05-09
`NumbaDeprecationWarning` from `numba/cuda/random.py`


RemiLehe:
## Reporting a bug

- [x] I have tried using the latest released version of Numba (most recent is
 visible in the change log (https://github.com/numba/numba/blob/main/CHANGE_LOG).
- [x] I have included a self contained code sample to reproduce the problem.
  i.e. it's possible to run as 'python bug.py'.

When running the example for [random number generation on GPU](https://numba.pydata.org/numba-doc/0.36.1/cuda/random.html#example), I get a bunch of `DeprecationWarning`s, similar to this:
```
numba/cuda/random.py:45: NumbaDeprecationWarning: The 'nopython' keyword argument was not supplied to the 'numba.jit' decorator. The implicit default value for this argument is currently False, but it will be changed to True in Numba 0.59.0. See https://numba.readthedocs.io/en/stable/reference/deprecation.html#deprecation-of-object-mode-fall-back-behaviour-when-using-jit for details.
```
It seems that some of the decorators would need to be updated?

guilhermeleobas:
Thanks for the report @RemiLehe. This issue was labelled to be fixed in the next release.

esc:
For reference, here are all the warnings, when running the example:

```
/home/esc/git/numba/numba/cuda/random.py:45: NumbaDeprecationWarning: The 'nopython' keyword argument was not supplied to the 'numba.jit' decorator. The implicit default value for this argument is currently False, but it will be changed to True in Numba 0.59.0. See https://numba.readthedocs.io/en/stable/reference/deprecation.html#deprecation-of-object-mode-fall-back-behaviour-when-using-jit for details.
  @jit(forceobj=_forceobj, looplift=_looplift)
/home/esc/git/numba/numba/cuda/random.py:71: NumbaDeprecationWarning: The 'nopython' keyword argument was not supplied to the 'numba.jit' decorator. The implicit default value for this argument is currently False, but it will be changed to True in Numba 0.59.0. See https://numba.readthedocs.io/en/stable/reference/deprecation.html#deprecation-of-object-mode-fall-back-behaviour-when-using-jit for details.
  @jit(forceobj=_forceobj, looplift=_looplift)
[...]

PR #8964, opened 2023-05-16
fix missing nopython keyword in cuda random module

esc:
Fixes: #8951

### DIFF
--- a/numba/cuda/random.py
+++ b/numba/cuda/random.py
@@ -34,7 +34,7 @@ xoroshiro128p_dtype = np.dtype([('s0', np.uint64), ('s1', np.uint64)],
 xoroshiro128p_type = from_dtype(xoroshiro128p_dtype)
 
 # When cudasim is enabled, Fake CUDA arrays are passed to some of the
-# @jit-decorated functions. This required the a fallback to object mode. With
+# @jit-decorated functions. This required fallback to object mode. With
 # Numba 0.59.0 this fallback must be explicitly enabled.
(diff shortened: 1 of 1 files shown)


gmarkall:
gpuci run tests

numba/cuda/random.py
```diff
@@ -34,15 +34,16 @@
 xoroshiro128p_type = from_dtype(xoroshiro128p_dtype)
 
 # When cudasim is enabled, Fake CUDA arrays are passed to some of the
-# @jit-decorated functions. This causes a fallback to object mode and the
-# emission of a warning at present - this is visible in the test suite with
-# cudasim, but also fallback-by-default will not be permitted in future. In
-# order to avoid the warning / future error, we explicitly specify that
+# @jit-decorated functions. This required the a fallback to object mode. With
```
gmarkall, line 37: "the a" - should this be:
```suggestion
# @jit-decorated functions. This required fallback to object mode. With
```

numba/cuda/random.py
```diff
@@ -34,15 +34,16 @@
...
+# Numba 0.59.0 this fallback must be explicitly enabled.
```
gmarkall, line 38: I think there is no fallback in the future:

```suggestion
# Numba 0.59.0 object mode must be explicitly enabled.
```

gmarkall (changes requested):
The code changes look good to me in their present form (assuming gpuCI passes) - just a couple of comments on the explanatory comments.

esc:
@gmarkall thank you for the review, I have addressed the review comments accordingly and pushed two new commits.

gmarkall:
gpuci run tests

gmarkall (approved):
Many thanks for the fix - this looks good to me!

PR #8965, opened 2023-05-16
fix return dtype for np.angle

esc:
Fixes: 8949


guilhermeleobas (approved):

guilhermeleobas:
Pending a test

esc:
@guilhermeleobas this looks good, unfortunately I can not approve my own PR, so you'll have to wait for @apmasell to do so. Anyway, all things considered, I think this is RTM.

apmasell (approved):

PR #8968, opened 2023-05-17
Cherry-pick #8895 into release0.57 (CUDA: Enable caching functions that use CG)

gmarkall:
This is based on PR #8895 which was merged into `main`.

I've tested this with RAPIDS and it appears not to introduce any regressions, so I have more confidence in introducing this to 0.57.1 (as discussed in last week's Numba dev / triage meetings).

Original commit / PR message follows:

> Caching functions that use Cooperative Groups was blocked because the dispatcher adds a file to link, the `cudadevrt` cudalib, to the code library, and pickling code libraries with linking files is not supported.
>
> To enable caching of functions that use cooperative groups, we push the linking of `cudadevrt` into the code library. We push it in here rather than special-casing the check for linked libraries because a recompile of a cached function (e.g. for a different compute capability) might require `cudadevrt` from a different location to the one found during the initial compile; locating the library at the cubin linking time ensures the correct version will be used.

### DIFF
--- a/numba/cuda/codegen.py
+++ b/numba/cuda/codegen.py
@@ -3,6 +3,7 @@ from llvmlite import ir
 from numba.core import config, serialize
 from numba.core.codegen import Codegen, CodeLibrary
 from .cudadrv import devices, driver, nvvm, runtime
+from numba.cuda.cudadrv.libs import get_cudalib
 
 import os
 import subprocess
@@ -74,6 +75,8 @@ class CUDACodeLibrary(serialize.ReduceMixin, CodeLibrary):
         # Files to link with the generated PTX. These are linked using the
         # Driver API at link time.
         self._linking_files = set()
+        # Should we link libcudadevrt?
+        self.needs_cudadevrt = False
 
         # Cache the LLVM IR string
         self._llvm_strs = None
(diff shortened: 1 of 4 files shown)


sklam:
BFID `numba_smoketest_cuda_yaml_199`

sklam:
> BFID `numba_smoketest_cuda_yaml_199`

passed

esc:
@gmarkall I can "review" this -- however, since a variant of this has already hit `main` -- to what extent must this be "re-reviewed"?

gmarkall:
@esc Many thanks - if our roles were reversed and I were doing the review, I'd only check that this PR implements the same thing as #8895, and approve if so.

esc:
> @esc Many thanks - if our roles were reversed and I were doing the review, I'd only check that this PR implements the same thing as #8895, and approve if so.

OK, great, thank you!

esc (approved):
I reviewed this patch the following way:

I used `git format-patch` to generate a patch file for the single commit included in this PR #8968. I then looked at #8895 and saw that it consists of three commits and a merge commit bringing in some upstream changes. I then created a temporary branch from #8895 and used `git rebase --interactive` to squash together the three original commits into a single one, so as to get only the changes included in this PR and not those introduced by the upstream-merge. Then I used `git format-patch` on that single commit and finally ran `diff` on the two patch files. Only the comments and SHA-1 sums are different, so the delta of hunks appears to be equal between the original and the cherry-pick PR:

```
 💣 zsh» diff patch01 patch02
1c1
< From cedb21ad2b3cd4f49711783dbde47003164feed7 Mon Sep 17 00:00:00 2001
---
> From 4da60e8ade7ea026a89a43e53d43c2eaed0eb522 Mon Sep 17 00:00:00 2001
19c19
< EDIT: all subsequent patches included too
---
> This is based on PR #8895 which was merged into `main`.
```

Hence I am giving this PR the thumbs up!

PR #8741, opened 2023-02-06
Added initial typed set implementation

kc611:
As titled, this PR adds initial support for `typed.set`s.

The following set methods are currently supported:
- `len(set())`
- `set().add()`
- `set().discard()`
- `set().__contains__()`
- `set().__iter__()`

PR #8973, opened 2023-05-22
[Experimental] Removed Set as type constraint from typed Dict key-value initialisation

kc611:
As titled,

This PR builds on top of typed set support PR #8741 and removes Set as a type constraint for dictionaries. This PR is separated from the original as the scope differs from the original PR.

### DIFF
--- a/docs/source/reference/pysupported.rst
+++ b/docs/source/reference/pysupported.rst
@@ -711,6 +711,49 @@ The use of reference counted types, e.g. strings, in sets is unsupported.
    made to the set will not be visible to the Python interpreter until
    the function returns.
 
+.. _feature-typed-set:
+
+Typed Set
+----------
+
+.. note::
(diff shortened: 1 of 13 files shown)


lgray:
I will try this soon. Thanks!

kc611:
Closed this after discussion in the triage meeting. This patch actually removes the wrong constraints, (i.e. it un-restricts reflected sets instead of the newly added typed sets). 

Also, See: https://github.com/numba/numba/pull/8741#issuecomment-1557208561

lgray:
@kc611 If this is not the correct way of doing things, how do I go about properly using a typed set back with #8741 (taking the example I posted, I have a full repro if you need)? If that PR is simply not enough for the functionality I require to move forward what can be done to get there?

lgray:
Ah, reading through again. Here I can still use typed.Set and leave in the sentry on the python side. I'll see how far this PR gets me.

PR #8974, opened 2023-05-23
CUDA: Add binary ufunc support

Matt711:
Following from #8294, this PR adds support for binary ufuncs for the CUDA target (`bitwise_and`, `bitwise_or`, etc.). Currently there aren't any unit tests for `np.left_shift` and `np.right_shift`. I didn't add them because there was a comment [here](https://github.com/numba/numba/blob/bca99bcde4abf9df48af9095ce4d0e4d6d76df80/numba/tests/test_ufuncs.py#L495) saying the tests were not added for the CPU target either.

### DIFF
--- a/numba/cuda/tests/cudapy/test_ufuncs.py
+++ b/numba/cuda/tests/cudapy/test_ufuncs.py
@@ -62,6 +62,29 @@ class TestUFuncs(BasicUFuncTest, TestCase):
              types.Array(types.float64, 2, 'F')),
         ])
 
+        # Add tests for other integer types
+        self.inputs.extend([
+            (np.uint8(0), types.uint8),
+            (np.uint8(1), types.uint8),
+            (np.int8(-1), types.int8),
+            (np.int8(0), types.int8),
+
+            (np.uint16(0), types.uint16),
+            (np.uint16(1), types.uint16),
+            (np.int16(-1), types.int16),
+            (np.int16(0), types.int16),
+
+            (np.ulonglong(0), types.ulonglong),
+            (np.ulonglong(1), types.ulonglong),
+            (np.longlong(-1), types.longlong),
+            (np.longlong(0), types.longlong),
+
+            (np.array([0,1], dtype=np.ulonglong),
+             types.Array(types.ulonglong, 1, 'C')),
+            (np.array([0,1], dtype=np.longlong),
+             types.Array(types.longlong, 1, 'C')),
+        ])
+
         self._low_occupancy_warnings = config.CUDA_LOW_OCCUPANCY_WARNINGS
         self._warn_on_implicit_copy = config.CUDA_WARN_ON_IMPLICIT_COPY
 
@@ -187,6 +210,11 @@ class TestUFuncs(BasicUFuncTest, TestCase):
 
     def test_bitwise_not_ufunc(self):
         self.basic_int_ufunc_test(np.bitwise_not)
+
+    # Note: there is no entry for np.left_shift and np.right_shift
+    # because their implementations in NumPy have undefined behavior
+    # when the second argument is a negative. See the comment in
+    # numba/tests/test_ufuncs.py for more details.
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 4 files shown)


numba/cuda/tests/cudapy/test_ufuncs.py
```diff
@@ -154,6 +173,21 @@ def test_degrees_ufunc(self):
...
+
+    def test_bitwise_xor_ufunc(self):
+        self.basic_int_ufunc_test(np.bitwise_xor)
+
+    def test_invert_ufunc(self):
+        self.basic_int_ufunc_test(np.invert)
+
+    def test_bitwise_not_ufunc(self):
+        self.basic_int_ufunc_test(np.bitwise_not)
+
```
gmarkall, line 190: Could you add a comment here explaining why there's no test for left and right shift? I don't think it's necessary to duplicate the entire comment from `numba/tests/test_ufuncs.py`, just referring to it would be fine.

numba/cuda/ufuncs.py
```diff
@@ -272,4 +272,55 @@ def np_real_atanh_impl(context, builder, sig, args):
 
     db[np.degrees] = db[np.rad2deg]
 
+    db[np.bitwise_and] = {
+        'll->l': numbers.int_and_impl,
```
gmarkall, line 276: Why do we not have implementations for unsigned types?

gmarkall (changes requested):
From a first pass over this, I think this is looking good like the pattern for implementing ufuncs in CUDA is followed correctly - I have a suggestion and a comment on the diff for now, and will have another pass once I understand the responses.

gmarkall:
gpuci run tests

numba/cuda/ufuncs.py
Matt711: Thanks! I added the missing types from `np.<ufunc>.types` except `O` which is a generic python object

gmarkall:
gpuci run tests

numba/cuda/ufuncs.py
```diff
@@ -272,4 +272,86 @@ def np_real_atanh_impl(context, builder, sig, args):
...
+        'LL->L': numbers.int_xor_impl,
+        'qq->q': numbers.int_xor_impl,
+        'QQ->Q': numbers.int_xor_impl,
+    }
+
+    db[np.invert] = {
+        'l->l': numbers.int_invert_impl,
+        'i->i': numbers.int_invert_impl,
+        '?->?': numbers.int_invert_impl,
+        'b->b': numbers.int_invert_impl,
```
gmarkall, line 321: Are we missing:

```suggestion
        'b->b': numbers.int_invert_impl,
        'B->B': numbers.int_invert_impl,
```

numba/cuda/ufuncs.py
```diff
@@ -272,4 +272,86 @@ def np_real_atanh_impl(context, builder, sig, args):
...
+        'h->h': numbers.int_invert_impl,
+        'q->q': numbers.int_invert_impl,
+        'L->L': numbers.int_invert_impl,
+        'I->I': numbers.int_invert_impl,
+        'H->H': numbers.int_invert_impl,
+        'Q->Q': numbers.int_invert_impl,
+    }
+
+    db[np.bitwise_not] = db[np.invert]
+
```
gmarkall, line 331: I think this is superflous, and just reassigns the same key in the dict - these two keys are the same:

```
$ python -c "import numpy as np; print(np.invert is np.bitwise_not)"
True
```

I think this can be removed.

gmarkall (changes requested):
Many thanks for the updates! A couple of quick questions / observations on the diff.

gmarkall (approved):
Many thanks for the updates! This looks good now.

I think I'd like a buildfarm run for this, since it's the first set of new ufuncs we've added since the addition of the feature - maybe eventually these additions will seem "safe" / generally expected to work everywhere if they work in gpuCI, but I'm not confident there yet.

@stuartarchibald / @sklam Could this have a buildfarm run please?

PR #8978, opened 2023-05-24
Import MVC packages when using MVCLinker.

### LINKED ISSUE #8977: NameError from CubinLinker with CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY
## Reporting a bug

In Numba 0.57.0, the config option `CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY` must be set _before_ importing `numba.cuda`. However, it's not always possible to know if minor version compatibility is needed at import time, and changing the value of `CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY` can result in a `NameError`.

This minimal repro:
```python
from numba import config, cuda

# Change the MVC config after importing numba.cuda
config.CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY = 1

@cuda.jit
def f():
    pass

f[1, 1, 0]()
```
gives
```
NameError: name 'CubinLinker' is not defined
```

I have a fix for this issue in #8978.

bdice:
Closes #8977.

The previous behavior conditionally imported `ptxcompiler` and `cubinlinker` when importing `numba.cuda`. This PR changes that behavior to perform the imports of `ptxcompiler` and `cubinlinker` as late as possible, when instantiating `MVCLinker`.

This change fixes the `NameError` while preserving a helpful error message if `ptxcompiler` or `cubinlinker` are not available. Now that error occurs when instantiating the linker rather than at `numba.cuda` import time.

### DIFF
--- a/numba/cuda/cudadrv/driver.py
+++ b/numba/cuda/cudadrv/driver.py
@@ -2661,14 +2661,10 @@ class Linker(metaclass=ABCMeta):
         """
 
 
-def _import_mvc_packages():
-    try:
-        from ptxcompiler import compile_ptx
-        from cubinlinker import CubinLinker, CubinLinkerError
-    except ImportError as err:
-        msg = ("Minor version compatibility requires ptxcompiler and "
-               "cubinlinker packages to be available")
-        raise ImportError(msg) from err
+_MVC_ERROR_MESSAGE = (
+    "Minor version compatibility requires ptxcompiler and cubinlinker packages "
+    "to be available"
+)
 
 
 class MVCLinker(Linker):
@@ -2677,9 +2673,11 @@ class MVCLinker(Linker):
     package.
     """
     def __init__(self, max_registers=None, lineinfo=False, cc=None):
-        # At this point, we know MVC is requested so we lazily import
-        # ptxcompiler and cubinlinker.
-        _import_mvc_packages()
+        try:
+            from cubinlinker import CubinLinker
+        except ImportError as err:
+            raise ImportError(_MVC_ERROR_MESSAGE) from err
+
         if cc is None:
             raise RuntimeError("MVCLinker requires Compute Capability to be "
                                "specified, but cc is None")
@@ -2704,6 +2702,11 @@ class MVCLinker(Linker):
         return self._linker.error_log
 
     def add_ptx(self, ptx, name='<cudapy-ptx>'):
+        try:
+            from ptxcompiler import compile_ptx
+            from cubinlinker import CubinLinkerError
+        except ImportError as err:
+            raise ImportError(_MVC_ERROR_MESSAGE) from err
         compile_result = compile_ptx(ptx.decode(), self.ptx_compile_options)
         try:
             self._linker.add_cubin(compile_result.compiled_program, name)
@@ -2711,6 +2714,11 @@ class MVCLinker(Linker):
             raise LinkerError from e
 
     def add_file(self, path, kind):
+        try:
+            from cubinlinker import CubinLinkerError
+        except ImportError as err:
+            raise ImportError(_MVC_ERROR_MESSAGE) from err
+
         try:
             with open(path, 'rb') as f:
                 data = f.read()
@@ -2747,6 +2755,11 @@ class MVCLinker(Linker):
         self.add_ptx(program.ptx.rstrip(b'\x00'), ptx_name)
 
     def complete(self):
+        try:
+            from cubinlinker import CubinLinkerError
+        except ImportError as err:
+            raise ImportError(_MVC_ERROR_MESSAGE) from err
+
         try:
             return self._linker.complete()
         except CubinLinkerError as e:
--- a/numba/cuda/cudadrv/driver.py
+++ b/numba/cuda/cudadrv/driver.py
@@ -37,15 +37,6 @@ from .drvapi import API_PROTOTYPES
 from .drvapi import cu_occupancy_b2d_size, cu_stream_callback_pyobj, cu_uuid
 from numba.cuda.cudadrv import enums, drvapi, _extras
 
-if config.CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY:
-    try:
-        from ptxcompiler import compile_ptx
-        from cubinlinker import CubinLinker, CubinLinkerError
-    except ImportError as ie:
-        msg = ("Minor version compatibility requires ptxcompiler and "
-               "cubinlinker packages to be available")
-        raise ImportError(msg) from ie
-
 USE_NV_BINDING = config.CUDA_USE_NVIDIA_BINDING
 
 if USE_NV_BINDING:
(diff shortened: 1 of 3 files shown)


gmarkall:
gpuci run tests

gmarkall (changes requested):
Many thanks for the PR! A couple of notes:

- The sysinfo tool determines whether MVC is enabled by checking whether `numba.cuda.cudadrv.driver` has a `CubinLinker` attr: https://github.com/numba/numba/blob/bca99bcde4abf9df48af9095ce4d0e4d6d76df80/numba/misc/numba_sysinfo.py#L395 - this check needs modifying as the attr is now no longer present just after import.
- I'm still looking into why gpuCI now fails in the "MVC-enabled" config (CUDA 11.8) - see https://gpuci.gpuopenanalytics.com/job/numba/job/numba/job/prb/job/numba-prb/510/CUDA_TOOLKIT_VER=11.8,CUDA_VER=11.2,LINUX_VER=ubuntu18.04,PYTHON_VER=3.8,RAPIDS_VER=21.12/

numba/cuda/cudadrv/driver.py
```diff
@@ -2670,12 +2661,25 @@ def complete(self):
         """
 
 
+def _import_mvc_packages():
+    try:
+        from ptxcompiler import compile_ptx
+        from cubinlinker import CubinLinker, CubinLinkerError
```
gmarkall, line 2667: I think this is importing these into the local scope of `_import_mvc_packages()`, rather than the `driver` module?

bdice: Oh, that's possible. I was still in the mindset of the other fix for cuDF where we were patching the module directly.

We have two options:
1. Declare the variables as `global` like this simplified example -- the `global` scope is the module scope here:
```
def _import_mvc_packages():
    global compile_ptx
    from ptx_compiler import compile_ptx
```
2. Import the names more precisely, only in the functions that require them.

gmarkall: I'd prefer option 2, because it keeps the effects of the import more localised, rather than stashing something into a module global when it's only used in one other place anyway.

gmarkall:
@bdice During our earlier (private) conversation, I mentioned I was confused about why we make a new linker for every kernel - I've now realised whilst looking at this code, that we do it because different kernels can have different linker configuration options (e..g max registers, lineinfo / debug info, etc.) so it doesn't make sense for kernels to share linker objects.

numba/cuda/cudadrv/driver.py
bdice: Fantastic. I agree that's best and I will implement those changes. Probably tomorrow.

gmarkall:
I have checked the docs for the MVC feature: https://numba.readthedocs.io/en/latest/cuda/minor_version_compatibility.html#enabling-mvc-support

It states that:

> MVC support is enabled [...] by setting a configuration variable prior to using any CUDA functionality

so arguably I think the docs already implied that this should work after importing `numba.cuda` but before actually using it, and therefore no documentation update is required to accompany this PR.

gmarkall:
gpuci run tests

numba/cuda/cudadrv/driver.py
bdice: Resolved in b062fee.

numba/misc/numba_sysinfo.py
```diff
@@ -392,8 +392,7 @@ def get_sysinfo():
                 sys_info[_cu_mvc_available] = False
 
             sys_info[_cu_mvc_needed] = cu_rt_ver > cu_drv_ver
-            mvc_used = hasattr(cudadrv.driver, 'CubinLinker')
-            sys_info[_cu_mvc_in_use] = mvc_used
+            sys_info[_cu_mvc_in_use] = config.CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY
```
gmarkall, line 395: The flag is an `int`, so it needs converting to `bool` to present consistently with the other CUDA settings reported by the sysinfo tool:

```suggestion
            sys_info[_cu_mvc_in_use] = bool(
                config.CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY)
```

to give e.g.:

```
CUDA Minor Version Compatibility Available    : True
CUDA Minor Version Compatibility Needed       : False
CUDA Minor Version Compatibility In Use       : False
```

as opposed to:

```
CUDA Minor Version Compatibility Available    : True
CUDA Minor Version Compatibility Needed       : False
CUDA Minor Version Compatibility In Use       : 0
```

gmarkall (changes requested):
Soe notes on the diff - I noticed it become outdated whilst doing the review so I'm posting it now to make my WIP comments visible

numba/cuda/tests/cudadrv/test_mvc.py
```diff
@@ -0,0 +1,47 @@
...
+        success = True
+    # Catch anything raised so it can be propagated
+    except: # noqa: E722
+        output = traceback.format_exc()
+        success = False
+
+    result_queue.put((success, output))
+
+
+class TestMinorVersionCompatibility(CUDATestCase):
```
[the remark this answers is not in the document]
gmarkall: BTW I suspect the use of CUDA in subprocesses is probably what interferes with the memcheck tool, so I think it's safer to skip it if we detect it's being used, in the same way that `test_ptds.py` does.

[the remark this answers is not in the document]
gmarkall: Also, no need to edit the remark for skipping on cudasim - your existing description is fine, it just happens to differ from mine.

gmarkall:
gpuci run tests

numba/cuda/tests/cudadrv/test_mvc.py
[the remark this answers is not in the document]
bdice: I like your comment better! Applied in cba0e1a.

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

numba/cuda/tests/cudadrv/test_mvc.py
```diff
@@ -0,0 +1,52 @@
...
+        success = False
+
+    result_queue.put((success, output))
+
+
+@linux_only
+@skip_under_cuda_memcheck('May hang CUDA memcheck')
+@skip_on_cudasim('Simulator does not require or implement MVC')
+class TestMinorVersionCompatibility(CUDATestCase):
+    def test_mvc(self):
```
gmarkall, line 37: Confirming this fails on `main`:

```
FAIL: test_mvc (numba.cuda.tests.cudadrv.test_mvc.TestMinorVersionCompatibility)
  File "/home/gmarkall/numbadev/numba/numba/cuda/tests/cudadrv/test_mvc.py", line 48, in test_mvc
    self.fail(output)
AssertionError: Traceback (most recent call last):
  File "/home/gmarkall/numbadev/numba/numba/cuda/tests/cudadrv/test_mvc.py", line 23, in child_test_wrapper
    output = child_test()
  File "/home/gmarkall/numbadev/numba/numba/cuda/tests/cudadrv/test_mvc.py", line 18, in child_test
    f[1, 1]()
  File "/home/gmarkall/numbadev/numba/numba/cuda/dispatcher.py", line 541, in __call__
  File "/home/gmarkall/numbadev/numba/numba/cuda/dispatcher.py", line 675, in call
  File "/home/gmarkall/numbadev/numba/numba/cuda/dispatcher.py", line 683, in _compile_for_args
  File "/home/gmarkall/numbadev/numba/numba/cuda/dispatcher.py", line 928, in compile
    kernel.bind()
  File "/home/gmarkall/numbadev/numba/numba/cuda/dispatcher.py", line 207, in bind
  File "/home/gmarkall/numbadev/numba/numba/cuda/codegen.py", line 184, in get_cufunc
  File "/home/gmarkall/numbadev/numba/numba/cuda/codegen.py", line 155, in get_cubin
    linker = driver.Linker.new(max_registers=self._max_registers, cc=cc)
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudadrv/driver.py", line 2612, in new
    return MVCLinker(max_registers, lineinfo, cc)
  File "/home/gmarkall/numbadev/numba/numba/cuda/cudadrv/driver.py", line 2692, in __init__
    self._linker = CubinLinker(f"--arch={arch}")
NameError: name 'CubinLinker' is not defined
```

and passes locally on this branch for me, so the test appears to be sound.

gmarkall (approved):
Many thanks - this looks good!

PR #8980, opened 2023-05-25
[CI only] Use nightly build of cubinlinker

gmarkall:
This is to test cubinlinker 0.3.0 on CI ahead of promoting it to the main RAPIDS channel. It is not intended to be merged into Numba.


gmarkall:
gpuci run tests

gmarkall:
Test output listing the conda environment on CUDA 11.8 shows:

```
11:50:48 cubinlinker               0.3.0            py38hbccff83_0    rapidsai-nightly
```

and it's passed, so it looks like this is OK from the pure Numba perspective - closing.

PR #8982, opened 2023-05-25
Don't do the parfor diagnostics pass for the parfor gufunc.

### LINKED ISSUE #8967: RuntimeError with NUMBA_PARALLEL_DIAGNOSTICS = 4
Self contained code sample to reproduce the problem:
```
import os
os.environ['NUMBA_PARALLEL_DIAGNOSTICS']='4'
import numba as nb
import numpy as np

@nb.njit(parallel=True)
def g():
    n = 10000
    b = np.zeros((n), dtype=np.float64)
    for i in nb.prange(n):
        b[i] = 1
    return b

g()
```

Returns following error:
```
LoweringError: Failed in nopython mode pipeline (step: native parfor lowering)
Failed in full_parfor_gufunc mode pipeline (step: dump parfor diagnostics)
self.setup has not been called

File "../../../../../../../../var/folders/bh/g86lgxvj67x8_qdx30vwmw500000gn/T/ipykernel_33020/3850650943.py", line 20:


During: lowering "id=0[LoopNest(index_variable = parfor_index.1, range = (0, n, 1))]{47: }Var(parfor_index.1, 3850650943.py:20)" at /var/folders/bh/g86lgxvj67x8_qdx30vwmw500000gn/T/ipykernel_33020/3850650943.py (20)
```
I would expect the the code to run without errors and returning some parallel diagnostic output.

Run with Numba 0.57.0 on MacBook Air M1

DrTodd13:
Resolves #8967

### DIFF
--- a/numba/core/compiler.py
+++ b/numba/core/compiler.py
@@ -602,7 +602,8 @@ class DefaultPassBuilder(object):
         return pm
 
     @staticmethod
-    def define_parfor_gufunc_nopython_lowering_pipeline(state, name='parfor_gufunc_nopython_lowering'):
+    def define_parfor_gufunc_nopython_lowering_pipeline(
+        state, name='parfor_gufunc_nopython_lowering'):
         pm = PassManager(name)
         # legalise
         pm.add_pass(NoPythonSupportedFeatureValidation,
--- a/numba/core/compiler.py
+++ b/numba/core/compiler.py
@@ -603,7 +603,7 @@ class DefaultPassBuilder(object):
 
     @staticmethod
     def define_parfor_gufunc_nopython_lowering_pipeline(
-        state, name='parfor_gufunc_nopython_lowering'):
+            state, name='parfor_gufunc_nopython_lowering'):
         pm = PassManager(name)
         # legalise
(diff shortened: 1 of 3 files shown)


numba/core/compiler.py
```diff
@@ -601,6 +601,25 @@ def define_nopython_lowering_pipeline(state, name='nopython_lowering'):
         pm.finalize()
         return pm
 
+    @staticmethod
+    def define_parfor_gufunc_nopython_lowering_pipeline(state, name='parfor_gufunc_nopython_lowering'):
```
esc, line 605:
```suggestion
    def define_parfor_gufunc_nopython_lowering_pipeline(
        state, name='parfor_gufunc_nopython_lowering'):
```

☝️ To fix the flake8 error 👇 

numba/core/compiler.py:605:81: E501 line too long (103 > 80 characters)

esc (changes requested):
Thank you for the patch. There was a flake8 violation, for which I made a suggestion. Additionally, I am requesting a test to include the original reproducer. Thank you.

esc (approved):
I tested this locally and can confirm the test does test what it is supposed to and the code is fine too.

PR #8988, opened 2023-05-28
support for latest CUDA driver codes #8363

### LINKED ISSUE #8363: Update `numba\cuda\cudadrv\enums.py`
## Feature request

Additional enums, e.g. for device attributes, have been added in recent cuda versions but are not yet represented in `numba\cuda\cudadrv\enums.py`.

s1Sharp:
fix #8363
I have added support for the latest CUDA driver codes. 

During the addition, I was guided by the values from the official documentation at the link: **https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES**
I also tried to make the file with the codes a little more readable.

p.s. UNKNOWN_CUDA_ERROR was returning to me, although I understood what error was in my code. But it didn't exist in the file numba/cuda/cudadrv/enums.py. This signaled me to update this file

### DIFF
--- a/numba/cuda/cudadrv/enums.py
+++ b/numba/cuda/cudadrv/enums.py
@@ -1,8 +1,10 @@
 """
 Enum values for CUDA driver. Information about the values
 can be found on the official NVIDIA documentation website.
-ref: https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES
+ref: https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html
+anchor: #group__CUDA__TYPES
 """
+
 
 # Error codes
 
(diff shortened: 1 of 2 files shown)


numba/cuda/cudadrv/enums.py
```diff
@@ -180,14 +309,16 @@
 # Applicable options: PTX compiler options, ::CU_JIT_FALLBACK_STRATEGY
 CU_JIT_INPUT_LIBRARY = 4
 
+CU_JIT_NUM_INPUT_TYPES = 6
```
gmarkall, line 312: I note that `CU_JIT_INPUT_NVVM` is not added here, but I'm happy with its omission given that:
- It's already deprecated.
- It can never be used in Numba. For CUDA 12, we will support NVVM input using the nvjitlink library. I'm working on a Python binding for this with @brandon-b-miller at https://github.com/gmarkall/nvjitlink

gmarkall (approved):
These additions look good - I've cross-checked with the documentation you linked to.

Many thanks for the PR!

Issue #8989, opened 2023-05-29
Spurious deep copies in inlining


JSKenyon:
## Reporting a bug

- [X] I have tried using the latest released version of Numba (most recent is
 visible in the change log (https://github.com/numba/numba/blob/main/CHANGE_LOG).
- [ ] I have included a self contained code sample to reproduce the problem.
  i.e. it's possible to run as 'python bug.py'.

I recently profiled the compilation of some code in an effort be better understand where time is spent inside Numba. For my specific use case, I saw that the majority of compilation time was spent inside `deepcopy`. After some investigation, I figured out that those deep copies are associated with inlining. The code which produces these copies is here: https://github.com/numba/numba/blob/0b85096be667ed817a52f5989bf252bda552b8f1/numba/core/inline_closurecall.py#L371-L381

The problem (perhaps a bug, although I may have missed some subtlety) is that the first `deepcopy` (line 375) is followed by immediately discarding the contents of the body (line 376) and then repopulating it with additional deep copies (lines 377-379). It is this second set of copies that I am struggling to explain - surely those elements would already have been copied in the original `deepcopy` call?

The reason I am raising this is because inlining (specifically these deep copies) is currently dominating my compilation times. As an experiment, I commented out lines 376-379 and saw roughly an x2.5 speedup in compilation with no decrease in performance of the compiled code. Note that disabling inlining completely (by removing inline='always' in my code) typically results in 40% slower code in my specific use case (which is somewhat unexpected, but clearly inlining into the Numba IR helps LLVM optimize the code in some way).

To summarize: inlining produces many expensive deep copy operations - are they all necessary?

kc611:
Thank you @JSKenyon for reporting the issue.

On closer inspection this deep-copy logic was added in https://github.com/numba/numba/pull/5812.

The problem was related to the IR being mutated in-place, a complete description for the original issue can be found at: https://github.com/numba/numba/issues/5824 and a reproducer is at: https://github.com/numba/numba/issues/5824#issuecomment-641202064

I've opened a PR #8990 to check if any known tests fails with the suggestions that you've given.

JSKenyon:
I don't believe this issue should be marked as stale until there is an explanation of the current code.

Issue #5824, opened 2020-06-06
Overload inliner regression in 0.50.0rc1


ehsantn:
## Reporting a bug

- [ ] I am using the latest released version of Numba (most recent is visible in
 the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [ ] I have included below a minimal working reproducer (if you are unsure how
 to write one see http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports).

Overload inliner in 0.50.0rc1 reuses the implementation function's IR if the original function is called multiple times: https://github.com/numba/numba/blob/015dbf83df8b11bb25d0aa6b96cae809939dfaf6/numba/core/typed_passes.py#L681
Since Numba IR is mutable, reuse of IR nodes results in errors (I don't have a small reproducer yet). Therefore, a new IR should be generated each time (since IR cannot fully be copied yet).

stuartarchibald:
Thanks for the report. Likely fixed by: https://github.com/numba/numba/pull/5812 ?

ehsantn:
@stuartarchibald I don't believe #5812 actually solves the problem since the IR is not fully copied. Patching up a few places to avoid mutation is not enough, since the assumption that the IR is mutable is present everywhere (including user passes, not just Numba passes). Regenerating the IR each time is the safest option I believe.

stuartarchibald:
The patch in #5812 solves it for the cases known but perhaps not for everything, suggests more tests are needed. I'm reluctant to re-run the entire untyped compilation pipeline every time a function is to be inlined as this is costly, I will look at copying the IR at the statement level just prior to inlining which ought to have the same effect. If you have a use case that triggers the issue then please could you share it? Thanks.

ehsantn:
ParallelAccelerator uses `deepcopy` of IR already (e.g. stencils), but I'm not sure if it works for everything. Unfortunately, I can't create a small reproducer easily since it needs non-trivial optimization passes to trigger the issue.

stuartarchibald:
This triggers it:
```python
from numba import njit, types
import numpy as np
from numba.core.extending import overload
from numba.core.typed_passes import InlineOverloads
from numba.core.compiler import CompilerBase, DefaultPassBuilder

class CustomCompiler(CompilerBase):

    def define_pipelines(self):
        pm = DefaultPassBuilder.define_nopython_pipeline(self.state)
        pm.add_pass_after(InlineOverloads, InlineOverloads)
        pm.finalize()
[...]

PR #8990, opened 2023-05-30
Removed extra block copying in InlineWorker

kc611:
Based on #8989 ,

This PR checks if removing the secondary deep-copying for IR in `InlineWorker` fails any known tests in our test suite.

### DIFF
--- a/numba/core/inline_closurecall.py
+++ b/numba/core/inline_closurecall.py
@@ -373,10 +373,6 @@ class InlineWorker(object):
             kernel_copy.blocks = {}
             for block_label, block in the_ir.blocks.items():
                 new_block = copy.deepcopy(the_ir.blocks[block_label])
-                new_block.body = []
-                for stmt in the_ir.blocks[block_label].body:
-                    scopy = copy.deepcopy(stmt)
-                    new_block.body.append(scopy)
                 kernel_copy.blocks[block_label] = new_block
             return kernel_copy
 


JSKenyon:
Just checking in here @esc and @kc611 - I see this has been marked as abandoned. Isn't this an easy win for compilation performance? Or has it been superseded by other planned inlining improvements?

gmarkall:
Putting this into the 0.59 milestone - this PR is considered a bit safer for that release as the object mode fallback pipeline will be removed in that release, and this deep copy may have made some operations in that mode safe. (even though CI passes for this at present)

gmarkall:
Noting also that this looks like it will be good to merge once the object mode fallback has been removed.

stuartarchibald:
Outcome of maintainer discussion is here: https://github.com/numba/numba/issues/8989#issuecomment-1715519353 and noted by @gmarkall here https://github.com/numba/numba/pull/8990#issuecomment-1669775041, essentially, the tests continue to pass and the concerns about mutation of IR and object-mode fallback will be mitigated by removal of this behaviour in 0.59 anyway. This PR should be ok as-is.

gmarkall (approved):
I'm happy to approve this based on the discussions we've had and the removal of object-mode fallback for 0.59.

ehsantn:
The issue raised in https://github.com/numba/numba/issues/5824 is not a concern anymore? Would love to see compilation time improve, but Numba IR is mutable and optimization passes may change nodes. Making the IR immutable would require time to update all the optimization code.

gmarkall:
> The issue raised in #5824 is not a concern anymore?

Can you be more specific here? I'm not sure I follow.

> Would love to see compilation time improve, but Numba IR is mutable and optimization passes may change nodes. Making the IR immutable would require time to update all the optimization code.

Can you provide an example of code that needs updating?

ehsantn:
The concern is also with external code that uses Numba as a compiler toolkit and adds transformation/optimization passes, not just code inside the Numba code base. Unfortunately, this issue is too hard to reproduce with a small reproducer easily and needs full optimization passes to demonstrate.

Even inside Numba codebase, there are a lot of IR mutations (e.g. [here](https://github.com/numba/numba/blob/596e8a55334cc46854e3192766e643767bd7c934/numba/core/ir_utils.py#L406)). Unit tests may not be comprehensive enough to detect issues that require specific interactions of transformations.

gmarkall:
@ehsantn In my understanding, this PR is not making Numba IR immutable. It is removing a redundant deep copy that discards / replaces something already copied by a prior deep copy:

> The problem (perhaps a bug, although I may have missed some subtlety) is that the first `deepcopy` (line 375) is followed by immediately discarding the contents of the body (line 376) and then repopulating it with additional deep copies (lines 377-379). It is this second set of copies that I am struggling to explain - surely those elements would already have been copied in the original `deepcopy` call?

(from @JSKenyon's description in #8989)

> Unfortunately, this issue is too hard to reproduce with a small reproducer easily and needs full optimization passes to demonstrate.

For the purpose of demonstrating that this PR breaks something, I think showing a full optimization pass that breaks under the application of this PR would be an appropriate starting point.

gmarkall:
Following discussion in the public dev meeting: this will be marked RTM for now and merged, on the basis that it can easily be reverted if it proves to be an issue.

@ehsantn Following the merge, please do open an issue with a reproducer (large or small, as a starting point) if this turns out to be problematic.

ehsantn:
Thanks for following up @gmarkall. The fact that this PR is just removing redundant copies makes sense. The comment about waiting for removal of object mode fallback was confusing to me since implied that the copy isn't redundant.

`Block` class doesn't have an explicit `__deepcopy__` method but I just checked CPython source code and manually verified that it does the right thing (attributes are deepcopied properly). I think merging this should be fine. I'll follow up if there were issues.

JSKenyon:
Thank you to everyone who contributed to this - looking forward to testing it out.

PR #8991, opened 2023-05-30
Cherry-pick and squash PR #8978 (Import MVC packages when using MVCLinker) for Release 0.57.1

gmarkall:
This cherry-pick and squash incorporates the fix for Issue #8977 (NameError from CubinLinker with `CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY`) from PR #8978 (Import MVC packages when using MVCLinker). As discussed during triage / FPOC discussions, this is for 0.57.1 because it forces cuDF to monkey-patch Numba to avoid `NameError`s in a supported use case: importing `numba.cuda` then enabling Minor Version Compatibility (MVC) prior to use of the CUDA functionality.

Tested locally with and without MVC enabled, by running the CUDA tests (`python runtests.py numba.cuda.tests`) and verifying the output of the sysinfo tool by inspection.

cc @bdice.

### DIFF
--- a/numba/cuda/cudadrv/driver.py
+++ b/numba/cuda/cudadrv/driver.py
@@ -37,15 +37,6 @@ from .drvapi import API_PROTOTYPES
 from .drvapi import cu_occupancy_b2d_size, cu_stream_callback_pyobj, cu_uuid
 from numba.cuda.cudadrv import enums, drvapi, _extras
 
-if config.CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY:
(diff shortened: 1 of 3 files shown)


esc (approved):
I can confirm this cherry-picks the changes from #8978 

```
 💣 zsh» git diff 6799f6d~1...import-mvc-packages > patch01 && git diff release0.57...release0.57-8978 > patch02 && diff patch01 patch02
 💣 zsh» echo $?
0
```

PR #8992, opened 2023-05-30
Add OrcJIT backend to Numba

apmasell:
Creates a new code generator and library pair to work with OrcJIT and also AOT compilation that is compatible with OrcJIT.

### DIFF
--- a/numba/core/cpu.py
+++ b/numba/core/cpu.py
@@ -235,20 +235,19 @@ class CPUContext(BaseContext):
         """
         Returns
         -------
-        (cfunc, fnptr)
+        cfunc
 
         - cfunc
-            callable function (Can be None)
-        - fnptr
-            callable function address
-        - env
-            an execution environment (from _dynfunc)
+            callable function.
+            It's used as dictionary key thus must uniquely identifiable.
         """
-        if not fndesc.llvm_cpython_wrapper_name:
-            return None
-        # Code generation
-        fnptr = library.get_pointer_to_function(
-            fndesc.llvm_cpython_wrapper_name)
+
+        if fndesc.llvm_cpython_wrapper_name:
+            # Code generation
+            fnptr = library.get_pointer_to_function(
+                fndesc.llvm_cpython_wrapper_name)
+        else:
+            fnptr = 0
 
         # Note: we avoid reusing the original docstring to avoid encoding
         # issues on Python 2, see issue #1908
@@ -259,7 +258,8 @@ class CPUContext(BaseContext):
                                        # objects to keepalive with the function
                                        (library,)
                                        )
-        library.set_env(self.get_env_name(fndesc), env)
+        if fndesc.llvm_cpython_wrapper_name:
+            library.set_env(self.get_env_name(fndesc), env)
         return cfunc
 
     def calc_array_sizeof(self, ndim):
--- a/numba/core/cpu.py
+++ b/numba/core/cpu.py
@@ -1,9 +1,11 @@
+import os
 import platform
 
 import llvmlite.binding as ll
 from llvmlite import ir
 
 from numba import _dynfunc
+from numba.core.bad_gen import BadCodegen
 from numba.core.callwrapper import PyCallWrapper
 from numba.core.base import BaseContext
 from numba.core import (utils, types, config, cgutils, callconv, codegen,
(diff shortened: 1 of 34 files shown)


numba/core/cpu.py
```diff
@@ -235,6 +244,8 @@ def get_executable(self, library, fndesc, env):
         - env
             an execution environment (from _dynfunc)
         """
+        if not fndesc.llvm_cpython_wrapper_name:
+            return None
```
sklam, line 248: @apmasell, This is the source of the imputils `AssertionError`. The return value of this function is later used as a key (`func`) in https://github.com/numba/numba/blob/22c322acd943901769ba480fc6d2686642719ae9/numba/core/base.py#L390
So this value needs to be a unique object that is suitable as dict key and will be stored in the Dispatcher as the "entry_point" (`cres.entry_point`) for the overload.

numba/core/imputils.py
```diff
@@ -196,7 +196,13 @@ def imp(context, builder, sig, args):
...
         with cgutils.if_unlikely(builder, status.is_error):
             context.call_conv.return_status_propagate(builder, status)
-        assert sig.return_type == fndesc.restype, \
+        ret_ty = (sig.return_type.literal_type
+                  if isinstance(sig.return_type, types.Literal)
+                  else sig.return_type)
+        res_ty = (fndesc.restype.literal_type
+                  if isinstance(fndesc.restype, types.Literal)
+                  else fndesc.restype)
+        assert ret_ty == res_ty, \
```
sklam, line 205: This change needs reverting

numba/core/cpu.py
sklam: The following patch will make it work for `NUMBA_JIT_BACKEND=orc python runtests.py numba.tests.test_builtins.TestOperatorMixedTypes.test_eq_ne` but I am still seeing a different segfault with `NUMBA_JIT_BACKEND=orc python runtests.py numba.tests.test_builtins.TestBuiltins.test_abs`

Btw, the docstring on the function is outdated and misleading.

```diff
diff --git a/numba/core/cpu.py b/numba/core/cpu.py
index 234eb05ffd..7c4e259743 100644
--- a/numba/core/cpu.py
+++ b/numba/core/cpu.py
@@ -235,20 +235,19 @@ class CPUContext(BaseContext):
         """
         Returns
         -------
-        (cfunc, fnptr)
+        cfunc
 
         - cfunc
-            callable function (Can be None)
-        - fnptr
-            callable function address
-        - env
-            an execution environment (from _dynfunc)
+            callable function.
+            It's used as dictionary key thus must uniquely identifiable.
         """
-        if not fndesc.llvm_cpython_wrapper_name:
-            return None
-        # Code generation
-        fnptr = library.get_pointer_to_function(
-            fndesc.llvm_cpython_wrapper_name)
+
+        if fndesc.llvm_cpython_wrapper_name:
+            # Code generation
+            fnptr = library.get_pointer_to_function(
+                fndesc.llvm_cpython_wrapper_name)
+        else:
+            fnptr = 0
 
         # Note: we avoid reusing the original docstring to avoid encoding
         # issues on Python 2, see issue #1908
@@ -259,7 +258,8 @@ class CPUContext(BaseContext):
                                        # objects to keepalive with the function
                                        (library,)
                                        )
-        library.set_env(self.get_env_name(fndesc), env)
+        if fndesc.llvm_cpython_wrapper_name:
+            library.set_env(self.get_env_name(fndesc), env)
         return cfunc
 
     def calc_array_sizeof(self, ndim):
```

numba/core/image_gen.py
```diff
@@ -0,0 +1,480 @@
...
+        tm_options = dict(opt=config.OPT)
+        if image_configuration.cpu_name == 'host':
+            cpu_name = (ll.get_host_cpu_name()
+                        if config.CPU_NAME is None
+                        else config.CPU_NAME)
+        else:
+            cpu_name = image_configuration.cpu_name
+        tm_options['cpu'] = cpu_name
+        tm_options['reloc'] = 'pic'
+        tm_options['codemodel'] = 'default'
```
sklam, line 207:
```suggestion
        tm_options['codemodel'] = 'jitdefault'
```

On the M1 Mac, the default code model ended up as "small" and causes failure in OrcJIT to find in-process symbols (e.g. `puts`). I made a reproducer at https://gist.github.com/sklam/108e89c745de9d9cbc4fa5154721d760, which tries to use `puts` (optimized from `printf`) and `lldb` shows generated machine code binds to the wrong address. The program will segfault.

Searching for the problem led me to https://github.com/llvm/llvm-project/issues/84#issuecomment-576900826 which suggested using the "jitdefault" codemodel. I verified this change worked on my M1 Mac. But we need to verify how portable this really is on different OS+arch given past experience with MCJIT.

PR #8996, opened 2023-06-01
adding a test for 8940

esc:
As title

Without the import from #8958 this reproducer from #8940 will fail.


esc:
Using `exec` only allows the test to fail when using the test-suite with `-m`. Using `test_in_subprocess` allows the test to fail, regardless of the invocation of the test-suite runner. 7e128634c90bc3970037e62288354878b06e4807 fixes.

apmasell (approved):

esc:
@apmasell thank you for the review.

PR #8998, opened 2023-06-02
Fix 8939 (second attempt)

### LINKED ISSUE #8939: Regression with `literal_unroll` in 0.57.0

esc:
Fixes #8939

See also: https://github.com/numba/numba/issues/8939#issuecomment-1573751234

### DIFF
--- a/numba/core/untyped_passes.py
+++ b/numba/core/untyped_passes.py
@@ -909,11 +909,10 @@ class MixedContainerUnroller(FunctionPass):
                 try:
                     scope.get_exact(name)
                 except errors.NotDefinedError:
-                    # In case the scope doesn't have the
-                    # variable, we need to define it prior creating new
-                    # copies of it! This is because the scope of the function
-                    # and the scope of the function and the scope of the loop
-                    # are different and the variable needs to be redefined
+                    # In case the scope doesn't have the variable, we need to
+                    # define it prior creating new copies of it! This is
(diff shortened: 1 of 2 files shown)


apmasell (approved):

esc:
@apmasell thank you for the review!

Issue #9000, opened 2023-06-05
Numba 0.57.1-rc1 Checklist


esc:
## numba 0.57.1-rc1

* [x] Cherry-pick items from the 0.57.1-rc1 milestone into a PR.

Non cherry-picked issues, these are issues that have not been cherry-picked yet individually and so they need to become part of the cherry-pick PR.

Issue |PR    |Merge
------|------|----------
#8940 |#8958 |0c8b59e106
#8940 |#8996 |0c752441b3
#8936 |#8947 |ec09902714
#8939 |#8998 |24c89c25dd
#8967 |#8982 |dd082ca031
#8949 |#8965 |bca99bcde4
#8951 |#8964 |98ab12e35c
None  |#9004 |5f1ea887113
None  |#9006 |9710916650
None  |#9005. |c9cc06ba141

Issues that have already been cherry-picked. These have had the author create individual cherry-pick PRs, so they do not need to be included in the cherry-pick PR, but they DO need to mentioned in the change log.

Issue |PR    |Cherry
------|------|------
#8994 |#None |#None
#8977 |#8978 |#8991
#8888 |#8895 |#8968
#None |#8976 |#8979

Cherry-pick PR is now: https://github.com/numba/numba/pull/9003

* [x] Update the "version support table" in the documentation with the final
  release date .
  * [ ] <strike>Make, approve and merge a PR against the `main` branch.</strike>
  * [ ] <strike>Create a cherry-pick from the merge and include in the cherry-pick-PR</strike>
    for the `release0.57` branch.</strike?
  * [x] add to cherry-pick PR.
* [x] Check if any dependency pinnings need an update (e.g. NumPy) and update on cherry-pick branch.
* [ ] <strike>Approve change log modifications and cherry-pick.</strike>
* [x] Add changelog modifications to cherry-pick PR.
* [x] Merge change log modifications and cherry-picks to 0.57 release branch.
  * [x] https://github.com/numba/numba/pull/9003
* [ ] <strike>Review, merge and check execution of release notebook. (FINAL ONLY)</strike>
* [x] Run the HEAD of the release branch through the build farm and confirm: (pre-tag testing)
  * [x] Build farm CPU testing has passed. (`numba_yamla_363`)
  * [x] Build farm CUDA testing has passed (`numba_smoketest_cuda_yaml_203`)
  * [x] Build farm wheel testing has passed (`numba_wheel_83`)
* [x] Annotated tag 0.57.1rc1 on release branch (no `v` prefix).
[...]

Issue #8993, opened 2023-05-31
Different results with Numba 0.56 and 0.57


hameerabbasi:
## Reporting a bug

- [x] I have tried using the latest released version of Numba (most recent is
 visible in the change log (https://github.com/numba/numba/blob/main/CHANGE_LOG).
- [ ] I have included a self contained code sample to reproduce the problem.
  i.e. it's possible to run as 'python bug.py'.

It seems that Numba 0.56 is more consistent in what would happen without JIT than Numba 0.57. At PyData/Sparse, we have a CI job that disables JIT and runs the tests, that runs fine. However, we just received a bug report (pydata/sparse#594) which I highly suspect comes from improper code generation in Numba.

esc:
@hameerabbasi thank you for reporting this!  A few things to consider:

* We are about to push out a 0.57.1 -- but the window for including fixes is closing very quickly. If this does turn out to be a regression for 0.57.0 -- it is worth considering adding a fix for this issue.
* I glanced at the report, but nothing immediate came to mind. Would it be possible to make a small, self-contained reproducer that uses only Numba w/o sparse?

hameerabbasi:
Thanks @esc, I'll do my best to come up with a reproducer.

esc:
> Thanks @esc, I'll do my best to come up with a reproducer.

Any luck on this? FYI: I'd like to push a 0.57.1 out this week.

hameerabbasi:
Hey, unfortunately not yet. I haven't had time as my main PC needed repairs.

esc:
@kc611 and @stuartarchibald debugged this for some time. The conclusions were as follows:

There are two separate issues:

One is where the failing tests complain about lists being negative(integer overflow) or memory buffer error. These are numerical test failures, the underlying cause is an incorrect initialization of an array with garbage memory instead of zeros.

fixed by:

```diff
diff --git a/sparse/_common.py b/sparse/_common.py
 index 96f43c7..2f7a67e 100644
 --- a/sparse/_common.py
 +++ b/sparse/_common.py
 @@ -844,7 +844,7 @@ def _dot_csc_ndarray_type_sparse(dt1, dt2):
          a_shape, b_shape : Tuple[int]
              The shapes of the input arrays.
          """
 -        indptr = np.empty(b_shape[1] + 1, dtype=np.intp)
 +        indptr = np.zeros(b_shape[1] + 1, dtype=np.intp)
[...]

PR #9003, opened 2023-06-06
0.57.1 rel cherry pick

esc:
As per the table at:

https://github.com/numba/numba/issues/9000

I placed the following merge commits in the file `CP`:

```
0c8b59e106
0c752441b3
ec09902714
24c89c25dd
dd082ca031
bca99bcde4
98ab12e35c
```

And then ran:

```
cat CP | xargs git cp -m1
```

### DIFF
--- a/numba/core/compiler.py
+++ b/numba/core/compiler.py
@@ -602,6 +602,26 @@ class DefaultPassBuilder(object):
         return pm
 
     @staticmethod
+    def define_parfor_gufunc_nopython_lowering_pipeline(
+            state, name='parfor_gufunc_nopython_lowering'):
+        pm = PassManager(name)
+        # legalise
+        pm.add_pass(NoPythonSupportedFeatureValidation,
+                    "ensure features that are in use are in a valid form")
+        pm.add_pass(IRLegalization,
+                    "ensure IR is legal prior to lowering")
+        # Annotate only once legalized
+        pm.add_pass(AnnotateTypes, "annotate types")
+        # lower
+        if state.flags.auto_parallel.enabled:
+            pm.add_pass(NativeParforLowering, "native parfor lowering")
+        else:
+            pm.add_pass(NativeLowering, "native lowering")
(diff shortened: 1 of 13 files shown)


CHANGE_LOG
```diff
@@ -1,3 +1,27 @@
+Version 0.57.1 (8 June, 2023)
+----------------------------
```
DrTodd13, line 2:
```suggestion
-----------------------------
```

DrTodd13 (changes requested):

esc:
Added #9004 from c9cc06ba141

CHANGE_LOG
```diff
@@ -1,3 +1,30 @@
+Version 0.57.1 (8 June, 2023)
```
stuartarchibald, line 1: NOTE: date will need fixing to whenever the tag is made.

esc: indeed, we are currently estimating Monday, so will fix that now.

stuartarchibald: IIRC for the 0.57.0 RC release I left it as `TBC`, on the basis that it'll inevitably need patching for the final version. Perhaps the same could be done here?

esc: sure, I decided to reverse the order of change-log updates are applied. I am collelcting cherry-picks and updating the change-log in the same PR and will then cherry-pick those changes onto `main` once the FINAL is tagged. It seems much easier this way, especially since I can respond quickly to additional incoming cherry-picks.

esc:
This had been delayed, waiting for a fix for https://github.com/numba/numba/issues/8993 -- unfortunately no final verdict on this issue was reached. 

Hence, we will proceed now, with the cherry-picks in this PR as it is. This needs a review and can then be merged and go into pre-tag testing.

apmasell (approved):
All patches match originals

esc:
@apmasell thank you, merging this now and then will proceed with pre-tag testing.

Issue #8923, opened 2023-04-26
CUDA Vectorize appears to induce spurious device-to-host copies even with all inputs on device


gmarkall:
The following:

```python
from numba import cuda, vectorize
import numpy as np


@vectorize(['float32(float32)'], target='cuda')
def func(noise):
    return noise + 1.0


noise = np.random.randn(1, 3, 64, 64).astype(np.float32)
noise = cuda.to_device(noise)
out = func(noise)
```

seems to generate a device-to-host copy for no apparent reason:

```
$ NUMBA_CUDA_LOG_LEVEL=DEBUG python repro.py
== CUDA [103]  INFO -- init
== CUDA [104] DEBUG -- call driver api: cuInit
== CUDA [3205] DEBUG -- call driver api: cuCtxGetCurrent
== CUDA [3206] DEBUG -- call driver api: cuDeviceGetCount
== CUDA [3206] DEBUG -- call driver api: cuDeviceGet
== CUDA [3206] DEBUG -- call driver api: cuDeviceGetAttribute
== CUDA [3206] DEBUG -- call driver api: cuDeviceGetAttribute
== CUDA [3206] DEBUG -- call driver api: cuDeviceGetName
== CUDA [3206] DEBUG -- call driver api: cuDeviceGetUuid_v2
== CUDA [3206] DEBUG -- call driver api: cuDevicePrimaryCtxRetain
== CUDA [3253] DEBUG -- call driver api: cuCtxPushCurrent_v2
== CUDA [3253] DEBUG -- call driver api: cuMemGetInfo_v2
== CUDA [3358] DEBUG -- call runtime api: cudaRuntimeGetVersion
== CUDA [3359] DEBUG -- call runtime api: cudaRuntimeGetVersion
== CUDA [3359] DEBUG -- call runtime api: cudaRuntimeGetVersion
== CUDA [3361] DEBUG -- call driver api: cuCtxGetCurrent
== CUDA [3362] DEBUG -- call driver api: cuCtxGetDevice
== CUDA [3362] DEBUG -- call driver api: cuMemAlloc_v2
== CUDA [3362] DEBUG -- call driver api: cuMemcpyHtoD_v2
== CUDA [3362] DEBUG -- call driver api: cuCtxGetCurrent
== CUDA [3362] DEBUG -- call driver api: cuCtxGetDevice
== CUDA [3362] DEBUG -- call driver api: cuMemcpyDtoH_v2
[...]

PR #6878, opened 2021-03-31
CUDA: Support passing tuples to ufuncs

gmarkall:
This PR adds support for use cases like the following to CUDA ufuncs created with `@vectorize`:

```python
from numba import vectorize, float64
import math

@vectorize([float64(float64)], target='cuda')
def cuda_cos(x):
    return math.cos(x)

print(cuda_cos((1.0, 2.0, 3.0)))
```

Passing tuples, namedtuples, tuples of arrays, and nesting of these is supported - tests are added to cover some of these cases.

It seems that this already worked for gufuncs, so some tests are added for the equivalent in gufuncs, but no functional change is made there.

Fixes #6563.

PR #9005, opened 2023-06-06
Fix: Issue #8923 - avoid spurious device-to-host transfers in CUDA ufuncs

gmarkall:
Only convert an object to a NumPy array when we can't work out the dtype, so that we don't end up converting all arguments to NumPy arrays on the host.

Fixes Issue https://github.com/numba/numba/issues/8923.

### DIFF
--- a/numba/cuda/tests/cudapy/test_vectorize.py
+++ b/numba/cuda/tests/cudapy/test_vectorize.py
@@ -4,6 +4,7 @@ from collections import namedtuple
 from itertools import product
 from numba import vectorize
 from numba import cuda, int32, float32, float64
+from numba.cuda.cudadrv.driver import CudaAPIError, driver
 from numba.cuda.testing import skip_on_cudasim
 from numba.cuda.testing import CUDATestCase
 import unittest
@@ -226,6 +227,36 @@ class TestCUDAVectorize(CUDATestCase):
             return x ** 2
 
         self.assertEqual(bar.__name__, 'bar')
+
+    def test_no_transfer_for_device_data(self):
+        noise = np.random.randn(1, 3, 64, 64).astype(np.float32)
+        noise = cuda.to_device(noise)
+
+        # A mock of a CUDA function that always raises a CudaAPIError
+        def raising_transfer(*args, **kwargs):
+            raise CudaAPIError(999, 'Transfer not allowed')
+
+        old_HtoD = getattr(driver, 'cuMemcpyHtoD', None)
+        old_DtoH = getattr(driver, 'cuMemcpyDtoH', None)
+
+        setattr(driver, 'cuMemcpyHtoD', raising_transfer)
+        setattr(driver, 'cuMemcpyDtoH', raising_transfer)
+
+        try:
+            @vectorize(['float32(float32)'], target='cuda')
+            def func(noise):
+                return noise + 1.0
+
+            func(noise)
+        finally:
+            if old_HtoD is not None:
+                setattr(driver, 'cuMemcpyHtoD', old_HtoD)
+            else:
+                del driver.cuMemcpyHtoD_v2
+            if old_DtoH is not None:
+                setattr(driver, 'cuMemcpyDtoH', old_DtoH)
+            else:
+                del driver.cuMemcpyDtoH_v2
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


esc:
`numba_smoketest_cuda_yaml_200` <-- Build Farm ID.

gmarkall (commented):
Looking at the changes in https://github.com/numba/numba/pull/6878/files#diff-be83bb737c3b43603dfd2780d692709f3a95b36b9b86b77b88172404b97fab5c I am concerned I might have missed a case or two where spurious transfers could be induced in my haste to make this PR. I need to check whether the other modified cases in that PR could also induce unneeded transfers.

numba/cuda/tests/cudapy/test_vectorize.py
```diff
@@ -227,6 +228,36 @@ def bar(x):
...
+
+        # A mock of a CUDA function that always raises a CudaAPIError
+        def raising_transfer(*args, **kwargs):
+            raise CudaAPIError(999, 'Transfer not allowed')
+
+        old_HtoD = getattr(driver, 'cuMemcpyHtoD', None)
+        old_DtoH = getattr(driver, 'cuMemcpyDtoH', None)
+
+        setattr(driver, 'cuMemcpyHtoD', raising_transfer)
+        setattr(driver, 'cuMemcpyDtoH', raising_transfer)
```
sklam, line 243: Suggestion for future (when not in a hurry), try `unittest.mock`

numba/np/ufunc/deviceufunc.py
```diff
@@ -105,7 +105,10 @@ def _fill_argtypes(self):
         """
         for i, ary in enumerate(self.arrays):
             if ary is not None:
-                self.argtypes[i] = np.asarray(ary).dtype
+                dtype = getattr(ary, 'dtype')
+                if dtype is None:
+                    dtype = np.asarray(ary).dtype
+                self.argtypes[i] = dtype
```
sklam, line 111: I see. The `np.asarray` is forcing a transfer to the host while we only care about getting the `.dtype`.

numba/cuda/tests/cudapy/test_vectorize.py
```diff
@@ -227,6 +228,36 @@ def bar(x):
...
+            @vectorize(['float32(float32)'], target='cuda')
+            def func(noise):
+                return noise + 1.0
+
+            func(noise)
+        finally:
+            if old_HtoD is not None:
+                setattr(driver, 'cuMemcpyHtoD', old_HtoD)
+            else:
+                del driver.cuMemcpyHtoD_v2
```
sklam, line 255: Why are we deleting the `_v2` version but we never have set it?

numba/cuda/tests/cudapy/test_vectorize.py
```diff
@@ -227,6 +228,36 @@ def bar(x):
...
+        setattr(driver, 'cuMemcpyHtoD', raising_transfer)
+        setattr(driver, 'cuMemcpyDtoH', raising_transfer)
+
+        try:
+            @vectorize(['float32(float32)'], target='cuda')
+            def func(noise):
+                return noise + 1.0
+
+            func(noise)
+        finally:
```
sklam, line 251: I'll suggest adding a case where `raising_transfer` is called and the `CudaAPIError` is raised so we know the monkeypatch is working as expected.

numba/cuda/tests/cudapy/test_vectorize.py
gmarkall: I do use it in some other places but I wasn't quite sure I could reliably replace the function on the driver singleton and remove it again afterwards if it wasn't present before with `mock`.

numba/cuda/tests/cudapy/test_vectorize.py
gmarkall: It's a mistake, thanks for catching it.

numba/cuda/tests/cudapy/test_vectorize.py
gmarkall: Good point!

gmarkall:
gpuci run tests

gmarkall:
Having looked at this again I don't think I overlooked a case, I misunderstood the original changes in my earlier comment. So I think this is now ready for another look @sklam - many thanks for the first review!

esc:
Build Farm ID: `numba_smoketest_cuda_yaml_202`

sklam (approved):

PR #9007, opened 2023-06-08
CUDA: Add comparison ufunc support

Matt711:
Following from #8294, this PR adds support for comparison ufuncs for the CUDA target (`numpy.greater`, `numpy.greater_equal`, `numpy.less_equal`, etc.).

### DIFF
--- a/numba/cuda/ufuncs.py
+++ b/numba/cuda/ufuncs.py
@@ -21,7 +21,8 @@ def get_ufunc_info(ufunc_key):
 @lru_cache
 def ufunc_db():
     # Imports here are at function scope to avoid circular imports
-    from numba.cpython import cmathimpl, mathimpl
+    from numba.cpython import cmathimpl, mathimpl, numbers
+    from numba.np import npyfuncs
     from numba.np.npyfuncs import _check_arity_and_homogeneity
     from numba.np.npyfuncs import (np_complex_acosh_impl, np_complex_cos_impl,
                                    np_complex_sin_impl)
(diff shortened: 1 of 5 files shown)


gmarkall:
gpuci run tests

apmasell:
This code is going to be changed to support NumPy 1.25 in #9011 and which ever of this gets in second will have to be updated.

gmarkall:
gpuci run tests

numba/cuda/ufuncs.py
```diff
@@ -22,6 +22,7 @@ def get_ufunc_info(ufunc_key):
 def ufunc_db():
     # Imports here are at function scope to avoid circular imports
     from numba.cpython import cmathimpl, mathimpl, numbers
+    from numba.np import npyfuncs
     from numba.np.npyfuncs import _check_arity_and_homogeneity
     from numba.np.npyfuncs import (np_complex_acosh_impl, np_complex_cos_impl,
                                    np_complex_sin_impl)
```
gmarkall, line 28: Just a very minor suggestion: now that we're importing all of `npyfuncs` it would be a nice little tidy-up to remove these imports and prefix references to these functions with `npyufuncs.`

gmarkall (changes requested):
Many thanks for the PR - this is looking good! One very minor suggestion on the diff then I think this should be good to approve.

gmarkall:
gpuci run tests

gmarkall (approved):
Many thanks for the PR! I've also manually tested this on Windows, and all the functionality added here has precedent for the functionality used (e.g. bits from the `numbers` and `npyimpl` modules) so I think this is good to merge!

Matt711:
@esc @sklam Can we merge this?

esc:
I have retroactively added this to the 0.58RC milestone.

PR #9011, opened 2023-06-12
Implementation and test updates to support NumPy 1.25

apmasell:
NumPy 1.25 has two major changes:

- allow mixed signed comparisons ufuncs (the `qQ->?` and `Qq->?` implementations for `>`, `<`, `>=`, `<=`, `==`, `!=`); this mode was not supported by Numba and an appropriate implementation is added to match the C implementation in NumPy.
- the implementation of `sin` and `cos` are changed in a way that produces slightly different numeric results. This either lowers the comparison precision where these functions are used in tests or replaces them with another function (mostly `exp`) to avoid the differences.


numba/cpython/numbers.py
```diff
@@ -361,6 +361,58 @@ def int_ne_impl(context, builder, sig, args):
...
+
+        with builder.goto_block(bb_lt_zero):
+            res = builder.icmp_unsigned(op, left, Constant(left.type, 0))
+            phi.add_incoming(res, bb_lt_zero)
+            builder.branch(bb_exit)
+
+        with builder.goto_block(bb_ge_zero):
+            res = builder.icmp_unsigned(op, left, right)
+            phi.add_incoming(res, bb_ge_zero)
+            builder.branch(bb_exit)
```
sklam, line 384: Given how common comparison is, does it worth checking the generated machine code for to make sure the extra basic-blocks are all optimized-out as to not affect vectorization?  The alternative is to use `select`.

stuartarchibald:
Thanks for the patch. RE `sin`/`cos` changes, I think this patch https://github.com/numpy/numpy/pull/23925 reverts the implementation that was causing Numba's tests to fail, assuming this is backported to the NumPy 1.25 release branch, Numba will not need to make any related changes.

numba/cpython/numbers.py
apmasell: Sure. I can replace this with `select`. I was using the code as output by clang from the NumPy implementation.

buildscripts/incremental/setup_conda_environment.sh
```diff
@@ -30,7 +30,7 @@ conda list
 # Use conda-forge for NumPy 1.24 - at the time of writing it is not available
 # on the defaults channel.
 
-if [ "${NUMPY}" == "1.24" ]; then
+if [ "${NUMPY}" == "1.24" -o "${NUMPY}" == "1.25" ]; then
```
gmarkall, line 33: I think conda-forge is only needed for 1.25 these days (1.24 is in the default channels).

buildscripts/incremental/setup_conda_environment.sh
```diff
@@ -30,7 +30,7 @@ conda list
 # Use conda-forge for NumPy 1.24 - at the time of writing it is not available
```
gmarkall, line 30:
```suggestion
# Use conda-forge for NumPy 1.25 - at the time of writing it is not available
```

buildscripts/incremental/setup_conda_environment.cmd
```diff
@@ -16,7 +16,7 @@ set PIP_INSTALL=pip install -q
 
 @rem Use conda-forge for NumPy 1.24 - at the time of writing it is not available
 @rem on the defaults channel.
-if %NUMPY%==1.24 (set NUMPY_CHANNEL_PKG="conda-forge::numpy") else (set NUMPY_CHANNEL_PKG="numpy")
+if %NUMPY% GEQ 1.24 (set NUMPY_CHANNEL_PKG="conda-forge::numpy") else (set NUMPY_CHANNEL_PKG="numpy")
```
gmarkall, line 19: I think this should be OK with NP 1.24 in default channels now:

```suggestion
if %NUMPY%==1.25 (set NUMPY_CHANNEL_PKG="conda-forge::numpy") else (set NUMPY_CHANNEL_PKG="numpy")
```

numba/cpython/numbers.py
```diff
@@ -361,6 +361,38 @@ def int_ne_impl(context, builder, sig, args):
     return impl_ret_untracked(context, builder, sig.return_type, res)
 
 
+def int_su_cmp(op):
+    def impl(context, builder, sig, args):
+        (left, right) = args
+        # This code is entirely too clever. It's taken from the NumPy source.
```
gmarkall, line 367: Could you add a link to the NumPy source it's taken from?

numba/cpython/numbers.py
```diff
@@ -361,6 +361,38 @@ def int_ne_impl(context, builder, sig, args):
     return impl_ret_untracked(context, builder, sig.return_type, res)
 
 
+def int_su_cmp(op):
```
gmarkall, line 364: Perhaps:

```suggestion
def int_signed_unsigned_cmp(op):
```

would make it a bit easier to guess what the function does from the name? (I had to read the description in the implementation to figure out that s = signed and u = unsigned)

numba/cpython/numbers.py
```diff
@@ -361,6 +361,38 @@ def int_ne_impl(context, builder, sig, args):
...
+        # the expected unsigned-unsigned comparison operation.
+        cmp_zero = builder.icmp_signed('<', left, Constant(left.type, 0))
+        lt_zero = builder.icmp_signed(op, left, Constant(left.type, 0))
+        ge_zero = builder.icmp_unsigned(op, left, right)
+        res = builder.select(cmp_zero, lt_zero, ge_zero)
+        return impl_ret_untracked(context, builder, sig.return_type, res)
+    return impl
+
+
+def int_us_cmp(op):
```
gmarkall, line 383: Similar to above,

```suggestion
def int_unsigned_signed_cmp(op):
```

gmarkall (changes requested):
Things mostly look good from a quick skim of the diff - there are some comments on it.

The main issue I've come across is that building locally with:

```
python setup.py develop
```

results in:

```
$ python runtests.py -m
RuntimeError: The new DType API is currently in an exploratory phase and should NOT be used for production code.  Expect modifications and crashes!  To experiment with the new API you must set `NUMPY_EXPERIMENTAL_DTYPE_API=1` as an environment variable.
  File "/home/gmarkall/numbadev/numba/numba/testing/__init__.py", line 29, in load_testsuite
    suite.addTests(loader.loadTestsFromName(f))
  File "/home/gmarkall/mambaforge/envs/numbanp125/lib/python3.10/unittest/loader.py", line 154, in loadTestsFromName
  File "/home/gmarkall/numbadev/numba/numba/tests/npyufunc/test_gufunc.py", line 5, in <module>
    import numpy.core.umath_tests as ut
  File "/home/gmarkall/mambaforge/envs/numbanp125/lib/python3.10/site-packages/numpy/core/umath_tests.py", line 13, in <module>
    from ._umath_tests import *
RuntimeError: cannot load _umath_tests module.
```

I think requiring the user to set an environment variable pertaining to NumPy to run the testsuite is a problem - can / should we skip the relevant tests if `numpy.core.umath_tests` fails to import?

buildscripts/incremental/setup_conda_environment.sh
apmasell: This raises a bit of a concern because we've been using conda-forge packages until now for 1.24. Should we switch this off for 1.25 once Anaconda packages are ready?

numba/cpython/numbers.py
apmasell: No problem. I had it this way to match the other comparison functions.

numba/__init__.py
```diff
@@ -142,7 +142,7 @@ def test(argv, **kwds):
     """.split() + types.__all__ + errors.__all__
 
 
-_min_llvmlite_version = (0, 41, 0)
+_min_llvmlite_version = (0, 40, 0)
```
gmarkall, line 145: Was this accidentally added?

seberg:
> RuntimeError: cannot load _umath_tests module.

Odd, `conftest.py` should set that environment variable so that you don't run into that (although, I will admit that may not be a great solution).  Ping me if it remains/seems like a problem, then I will look into it.

gmarkall:
@seberg Thanks for the comment - perhaps we're getting around that environment variable setting because we were directly importing `numpy.core.umath_tests`? It shouldn't be a problem anymore as @apmasell updated our test suite not to use it (we shouldn't have been using it anyway)

numba/__init__.py
apmasell: Oops. Yes.

apmasell:
After discussion, I've updated to use the Anaconda-built packages. Work is proceeding on adding 1.25 to the build farm.

buildscripts/incremental/setup_conda_environment.cmd
```diff
@@ -14,9 +14,7 @@ cmd /C conda info
 set CONDA_INSTALL=cmd /C conda install -q -y
 set PIP_INSTALL=pip install -q
 
-@rem Use conda-forge for NumPy 1.24 - at the time of writing it is not available
-@rem on the defaults channel.
-if %NUMPY%==1.24 (set NUMPY_CHANNEL_PKG="conda-forge::numpy") else (set NUMPY_CHANNEL_PKG="numpy")
+set NUMPY_CHANNEL_PKG="numpy"
```
gmarkall, line 17: I can't make a suggestion on line L28/R26, but that line should change to:

```
conda create -n %CONDA_ENV% -q -y python=%PYTHON% numpy=%NUMPY% cffi pip scipy jinja2 ipython gitpython pyyaml
```

and this line be deleted.

buildscripts/incremental/setup_conda_environment.sh
```diff
@@ -27,14 +27,10 @@ source deactivate
...
-# Use conda-forge for NumPy 1.24 - at the time of writing it is not available
+# Use conda-forge for NumPy 1.25 - at the time of writing it is not available
 # on the defaults channel.
 
-if [ "${NUMPY}" == "1.24" ]; then
-  NUMPY_CHANNEL_PKG=conda-forge::numpy
-else
-  NUMPY_CHANNEL_PKG=numpy
-fi
+NUMPY_CHANNEL_PKG=numpy
```
gmarkall, line 33: This should also be deleted and change line L46/R42 to:

```
conda create -n $CONDA_ENV -q -y ${EXTRA_CHANNELS} python=$PYTHON numpy=$NUMPY pip gitpython pyyaml
```

numba/cpython/numbers.py
gmarkall: Reading through your description in the comments, I actually think this code is "just the right amount of clever" :slightly_smiling_face:

numba/cpython/numbers.py
```diff
@@ -361,6 +361,39 @@ def int_ne_impl(context, builder, sig, args):
...
+        (left, right) = args
+        # This code is entirely too clever. It's taken from the NumPy source.
+        # What we're going to do is divide the range of a signed value at zero.
+        # If the signed value is less than zero, then we can treat zero as the
+        # unsigned value since the unsigned value is necessarily zero or larger
+        # and any signed comparison between a negative value and zero/infinity
+        # will yield the same result. If the signed value is greater than or
+        # equal to zero, then we can safely cast it to an unsigned value and do
+        # the expected unsigned-unsigned comparison operation.
+        # Original: https://github.com/numpy/numpy/pull/23713
```
gmarkall, line 375: Thanks for adding this link - I appreciate the link to the PR because it also gives the context for the addition to NumPy, which is nice.

gmarkall (changes requested):
This is looking pretty good to me - there's a couple of small suggestions for tidying up the build scripts on the diff.

Could you also modify the gpuCI setup so that it tests 1.25 please? If you change this line:

https://github.com/numba/numba/blob/249c8ff3206928b486346443ec148508f8c25f8e/buildscripts/gpuci/build.sh#L38

so that `["11.8"]="1.24"` becomes `["11.8"]="1.25"` and drop this bit:

https://github.com/numba/numba/blob/249c8ff3206928b486346443ec148508f8c25f8e/buildscripts/gpuci/build.sh#L44-L48

then I think that will be sufficient.

As we're going to drop support for NumPy 1.21 (#9047), I think you might as well also remove the tests of 1.21 in the `azure-pipelines.yml` to save increasing the size of the test matrix

stuartarchibald (changes requested):
Thanks for the updates. Two actionable items:
1. Suggested changes on the patch.
3. The "Version support information" table in `docs/source/user/installing.rst` needs an update for 0.58.0 to note support for NumPy 1.25.

Other notes:
1. I've checked the NumPy 1.25 release notes and Numba doesn't implement any of the functions (or their use) that are (is) being deprecated.
2. The NumPy 1.25 release notes state that only `ndim-0` arrays are now treated as scalars, use of a single element array as a scalar is now deprecated. Am not sure of the impact on Numba yet but probably doesn't need covering in this PR.

esc:
I tried to run this through the Anaconda Build Farm and see the following across all supported platforms:

```
ERROR: test_eq_ne (numba.tests.test_npdatetime.TestTimedeltaArithmetic)
File "/opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/tests/test_npdatetime.py", line 414, in test_eq_ne
check(TD(1, 'Y'), TD(365, 'D'), False)
File "/opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/tests/test_npdatetime.py", line 397, in check
self.assertPreciseEqual(eq(a, b), expected_val)
TypeError: Cannot get a common metadata divisor for Numpy datetime metadata [Y] and [D] because they have incompatible nonlinear base time units.

FAIL: test_TestParfors (numba.tests.test_parfors.TestParforsRunner)
File "/opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/tests/test_parfors.py", line 98, in test_TestParfors
self.runner()
File "/opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/tests/test_parfors.py", line 84, in runner
self.subprocess_test_runner(test_module=themod,
File "/opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/tests/support.py", line 629, in subprocess_test_runner
AssertionError: 1 != 0 :
captured stdout:
captured stderr: ...............................s........./opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/core/typed_passes.py:334: NumbaPerformanceWarning: [1m
The keyword argument 'parallel=True' was specified but no transformation for parallel execution was possible.

To find out why, try turning on parallel diagnostics, see https://numba.readthedocs.io/en/stable/user/parallel.html#diagnostics for help.
[1m
File "opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/tests/test_parfors.py", line 1760:[0m
[1m        # test array.dtype transformation for datetime64
[1m        def test_impl(A):
[0m        [1m^[0m[0m
[0m
warnings.warn(errors.NumbaPerformanceWarning(msg,
./opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/core/typed_passes.py:334: NumbaPerformanceWarning: [1m
The keyword argument 'parallel=True' was specified but no transformation for parallel execution was possible.

To find out why, try turning on parallel diagnostics, see https://numba.readthedocs.io/en/stable/user/parallel.html#diagnostics for help.
[1m
File "opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/tests/test_parfors.py", line 2173:[0m
[1m        # This is a test for issue #6478.
[1m        def test_impl():
[0m        [1m^[0m[0m
[0m
warnings.warn(errors.NumbaPerformanceWarning(msg,
.E......................../opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/core/typed_passes.py:334: NumbaPerformanceWarning: [1m
The keyword argument 'parallel=True' was specified but no transformation for parallel execution was possible.

To find out why, try turning on parallel diagnostics, see https://numba.readthedocs.io/en/stable/user/parallel.html#diagnostics for help.
[1m
File "opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/tests/test_parfors.py", line 2163:[0m
[1m        # This is a test for issue #6478.
[1m        def test_impl():
[0m        [1m^[0m[0m
[0m
warnings.warn(errors.NumbaPerformanceWarning(msg,
..
ERROR: test_reduce (numba.tests.test_parfors.TestParfors)
File "/opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/tests/test_parfors.py", line 1427, in test_reduce
self.check(test_impl, A)
File "/opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/tests/test_parfors.py", line 288, in check
cfunc, cpfunc = self.compile_all(pyfunc, *args)
File "/opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/tests/test_parfors.py", line 188, in compile_all
cpfunc = self.compile_parallel(pyfunc, sig)
File "/opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/tests/test_parfors.py", line 176, in compile_parallel
return self._compile_this(func, sig, flags=self.pflags)
File "/opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/tests/test_parfors.py", line 173, in _compile_this
return compile_isolated(func, sig, flags=flags)
File "/opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/core/compiler.py", line 317, in compile_isolated
return compile_extra(typingctx, targetctx, func, args, return_type,
File "/opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/core/compiler.py", line 762, in compile_extra
File "/opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/core/compiler.py", line 460, in compile_extra
File "/opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/core/compiler.py", line 528, in _compile_bytecode
return self._compile_core()
File "/opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/core/compiler.py", line 507, in _compile_core
raise e
File "/opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/core/compiler.py", line 494, in _compile_core
pm.run(self.state)
File "/opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/core/compiler_machinery.py", line 368, in run
raise patched_exception
File "/opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/core/compiler_machinery.py", line 356, in run
File "/opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/core/compiler_lock.py", line 35, in _acquire_compile_lock
return func(*args, **kwargs)
File "/opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/core/compiler_machinery.py", line 311, in _runPass
File "/opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/core/compiler_machinery.py", line 273, in check
File "/opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/core/typed_passes.py", line 110, in run_pass
typemap, return_type, calltypes, errs = type_inference_stage(
File "/opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/core/typed_passes.py", line 88, in type_inference_stage
errs = infer.propagate(raise_errors=raise_errors)
File "/opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/core/typeinfer.py", line 1086, in propagate
raise errors[0]
[1m[1mUse of unsupported NumPy function 'numpy.amin' or unsupported use of the function.
[1m
File "opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/tests/test_parfors.py", line 1425:[0m
[1m        def test_impl(A):
[1m            min_val = np.amin(A)
[0m            [1m^[0m[0m
[0m
[0m[1mDuring: typing of get attribute at /opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/tests/test_parfors.py (1425)[0m
[1m
File "opt/conda/envs/testenv_a7c3db2b/lib/python3.10/site-packages/numba/tests/test_parfors.py", line 1425:[0m
[1m        def test_impl(A):
[1m            min_val = np.amin(A)
[0m            [1m^[0m[0m

```

apmasell:
For added fun, I cannot reproduce the parfors problems locally.

sklam:
RE https://github.com/numba/numba/pull/9011/commits/1a1c7c1bd5a6cc96cb8e6110480a3bf83d72a5eb 

I see that `numpy.amin` exists in numpy1.25 and did more digging. It turns out before numpy1.25 the `numpy.amin` is aliasing `np.min` but that's no longer true in numpy1.25.

```python
>>> import numpy as np
>>> np.min is np.amin
True
>>> np.__version__
'1.21.5'
```

```python
>>> import numpy as np
>>> np.min is np.amin
False
>>> np.__version__
'1.25.1'
```

You will need to do something like the following at https://github.com/numba/numba/blob/afb5246e5a512d9defd55d7326dfb27174ce0190/numba/np/arraymath.py#L491
```diff
 @overload(np.min)
+@overload(np.amin)
 @overload_method(types.Array, "min")
 def npy_min(a):
     if not isinstance(a, types.Array):
```

Same for `numpy.amax`. Probably want to check the numpy source diff to see if there are similiar changes to other function.

apmasell:
It's really unclear what's going on. Diffing NumPy, I see changes for:
- `amax` -> `max`
- `amin` -> `min`
- `around` -> `round`
- `product` -> `prod`
- `cumproduct` -> `cumprod`
- `sometrue` -> `any`
- `alltrue` -> `all`

With all of the original names marked as deprecated.

`around` is bound in both states. The rest are only bound to the future state.

apmasell:
Fixed and passing on the build farm.

numba/np/arraymath.py
```diff
@@ -489,6 +491,7 @@ def return_false(a):
 
 
 @overload(np.min)
+@overload(np.amin)
```
stuartarchibald, line 494: Two items:
1. In NumPy 1.25 `amin` and `min` cease to be aliases, as `amin` is actually declared as supported by Numba in the docs, the unit tests for `min` should also be applied to `amin` (same applies to the `amax`/`max` aliasing).
2. For the rest of the `@overload` additions in this module which have their aliases added, I don't think these are aliases by `id` i.e. the same function under a different name, they are just "equivalent functions", i.e. do the same thing, so they wouldn't have been supported by Numba as aliases. Essentially, for the function X and its potential alias Y, if  `X is Y`, returns `True` using NumPy 1.24 then it's a "true" alias and Numba needs support patching in for 1.25, if it's `False` then it does not.

numba/tests/test_fancy_indexing.py
```diff
@@ -406,7 +406,7 @@ def numba_get_item(array, idx):
         expected = get_item(arr, index)
         got = numba_get_item(arr, index)
         # Sanity check: In advanced indexing, the result is always a copy.
-        self.assertNotIn(expected.base, orig_base)
+        self.assertIsNot(expected.base, orig_base)
```
stuartarchibald, line 409: What prompted this change?

stuartarchibald (changes requested):
Thanks for the updates, there's a couple of comments in line to take a look at. I'm going to check the warnings from the deprecated aliases again just to make sure that they are occurring on execution not on attribute access. Thanks again!

numba/tests/test_fancy_indexing.py
apmasell: This triggered an error because other changes made the shape comparison logic different. Graham and I looked at the original PR, it was originally an is-not comparison though not using the unit test assertion, the review comment was ambiguous in that is said to use the method "in" the unittest system, and had it changed to not-in, which isn't the point of this comparison.

numba/np/arraymath.py
apmasell: 1. Any particular tests?
2. What is the action you want me to take?

stuartarchibald: 1. Suggest finding important cases involving `np.min` and `np.max`. Perhaps look in `test_array_reductions.py` as that's likely where they are used. It doesn't need to be exhaustive, just something to make sure that if anything "happens" to these currently untested aliases (`np.amax` and `np.amin`) it's picked up.
2. Suggest running the `X is Y` test for all the deprecated aliased functions and make sure that anything that is a true alias (function at same address) has an `@overload` declared in this patch, anything which is not a true alias needs the `@overload` removing from this patch.

apmasell: 1. Added
2. `X is Y` is `False` for all added overloads in the current patch.

numba/tests/test_fancy_indexing.py
stuartarchibald: Thanks for the explanation.

numba/np/arraymath.py
stuartarchibald: I've done what I hope is an exhaustive search for 2. The summary is as follows, if you observe the same I suggest updating this patch to reflect this.

The word `alias` herein means the test `X is Y` returns `True`.

From https://github.com/numpy/numpy/pull/23314:
* `product` - deprecated in 1.25, does not alias `prod` in 1.24 or 1.25.
* `cumproduct` - deprecated in 1.25, does not alias `cumprod` in 1.24 or 1.25.
* `sometrue` - deprecated in 1.25, does not alias `any` in 1.24 or 1.25.
* `alltrue` - deprecated in 1.25, does not alias `all` in 1.24 or 1.25.

From https://github.com/numpy/numpy/pull/23302:
* `round_` - deprecated in 1.25, aliases `round` in 1.24, does not alias `round` in 1.25.
* `around` - NOT deprecated in 1.25, does not alias `round` in 1.24 or 1.25.
* `amin` - NOT deprecated in 1.25, aliases `min` in 1.24, does not alias `min` in 1.25.
* `amax` - NOT deprecated in 1.25, aliases `max` in 1.24, does not alias `max` in 1.25.

Summary of anticipated impact on this PR/NumPy 1.25 support:
1. Numba does not need to add `@overload`s in the 1.25 support for `product`, `cumproduct`, `sometrue`, or `alltrue`, these are:
   a. not aliased functions as `X is Y` is `False` in 1.24 and 1.25.
   b. not declared as supported in Numba's documentation.
2. Numba does not need to add an `@overload` in the 1.25 support for `around` as it is:
   a. not an aliased function as `around is round` is `False` in 1.24 and 1.25.
   b. an `@overload` is already present in the code base and it's declared as supported in Numba's documentation.
3. Numba does need an `@overload` *and tests* adding in the 1.25 support for `round_` even thought it is deprecated as:
   a. this is an alias in 1.24 as `round_ is round` is `True`.
   b. this is not an alias in 1.25 as `round_ is round` is `False`.
   c. Numba declares `round_` as supported in its documentation.
4. Numba does need an `@overload` *and tests* adding in the 1.25 support for `amax` and `amin` as:
   a. these are aliased functions in 1.24 as `X is Y` is `True`.
   b. these are not aliased functions in 1.25 as `X is Y` is `False`.
   c. Numba declares `amax` and `amin` as supported in its documentation.

numba/cuda/ufuncs.py
stuartarchibald, line 1: Do the existing CUDA unit tests exercise the proposed additions to this module?

numba/tests/test_parfors.py
```diff
@@ -1421,9 +1422,14 @@ def test_impl(A):
...
-            min_val = np.amin(A)
-            return A - min_val
+        if numpy_version < (1, 25):
+            def test_impl(A):
+                min_val = np.amin(A)
+                return A - min_val
+        else:
+            def test_impl(A):
+                min_val = np.min(A)
+                return A - min_val
```
stuartarchibald, line 1432: What is the purpose of this change? Is it an artifact of fixing the test prior to discovering that `np.amin` and `np.min` do not alias in 1.25? If so it can probably be reverted?

sklam: This needs reverting. It's from 1a1c7c1bd5a6cc96cb8e6110480a3bf83d72a5eb and before new overloads were added

numba/np/arraymath.py
sklam: Thanks for the detail summary. It make sense.

numba/tests/test_parfors.py
apmasell: `np.min` is not available in NumPy 1.22

numba/cuda/ufuncs.py
apmasell: I believe so. They are hung off the existing ufunc testing infrastructure.

numba/np/arraymath.py
apmasell: I believe I've corrected this as requested.

numba/tests/test_array_methods.py
```diff
@@ -30,6 +30,9 @@ def np_around_unary(val):
 def np_round_array(arr, decimals, out):
     np.round(arr, decimals, out)
 
+def np_round__array(arr, decimals, out):
```
gmarkall, line 33: What is the double-underscore variant of this function and test for? It looks identical to the version above - am I missing something?

numba/np/ufunc_db.py
```diff
@@ -620,6 +620,10 @@ def _fill_ufunc_db(ufunc_db):
         'FF->?': npyfuncs.np_complex_gt_impl,
         'DD->?': npyfuncs.np_complex_gt_impl,
     }
+    if numpy_version >= (1, 25):
```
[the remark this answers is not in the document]
gmarkall: I'm not sure this is happening - there don't appear to be any tests generated that involve `qQ` or `Qq`:

```
$ python runtests.py numba.tests.test_ufuncs -l | egrep "(qQ|Qq)"
$
```

(i.e. no tests with signed / unsigned are generated).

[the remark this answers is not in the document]
gmarkall: Additionally, `BasicUFuncTest.basic_ufunc_test()` isn't set up to test binary functions with inputs of different types - it is set up to assume arguments are homogeneous in type:

https://github.com/numba/numba/blob/766f6d9b250ebf452c75590f753de0c003563541/numba/tests/test_ufuncs.py#L176-L184

(there is only one `input_type`)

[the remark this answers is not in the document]
gmarkall: My "egrep" check earlier was run erroneously because I didn't have NumPy 1.25 in my environment. The `TestLoopTypes` test classes were generated `qQ` and `Qq` variants of the tests. However, the `BasicUFuncTest` class didn't handle these combinations of types, so it was not testing on CUDA.

I tried experimenting with adding testing of these types to the `BasicUFuncTest` class as well - the commit https://github.com/gmarkall/numba/commit/865cfc6aea5aec24d97a0c5e961c4ac370a13580 on the branch https://github.com/gmarkall/numba/tree/numpy_1.25 implements this (and therefore tests these implementations with CUDA too) - feel free to pick it up / modify it if you'd like.

numba/tests/test_array_methods.py
apmasell: There's `round` and `round_` in NumPy, so, I'm trying to continue the pattern.

numba/np/ufunc_db.py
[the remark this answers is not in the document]
apmasell: Cherry-picked.

numba/tests/test_array_methods.py
gmarkall: The function calls `np.round` though, rather than `np.round_`.

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
@apmasell I hope you don't mind - I fixed the conflicts with the gpuCI setup and also fixed an issue with the config related to the Python version in conjunction with NumPy, so that this could be tested on gpuCI (rather than asking you to make changes and waiting).

numba/tests/test_array_methods.py
apmasell: Ooops.

apmasell:
No problem with the changes for gpuCI.

numba/tests/test_parfors.py
stuartarchibald: I think `min` is just an alias of `amin` in NumPy 1.22.x:
```pycon
>>> import numpy as np
>>> np.__version__
'1.22.3'
>>> np.min
<function amin at 0xsomeaddress>
```

I'm also of the view that this change needs reverting, `amin` is the function being tested originally and should continue to be tested invariant of how NumPy have implemented it.

numba/cuda/ufuncs.py
stuartarchibald: Fixed in 989e81d.

numba/np/arraymath.py
stuartarchibald: Thank you, I've taken a look and as of 0c086cf the diff looks like it covers the above details.

PR #9014, opened 2023-06-13
Change a bunch of var names created during lowering to start with a period.

DrTodd13:
Numba is supposed to use a convention that vars introduced during Numba passes start with $ and LLVM vars introduced during lowering start with a period.  There were a bunch of cases where vars introduced during lowering don't start with a period.  PyOMP relies on this convention.

### DIFF
--- a/numba/core/pythonapi.py
+++ b/numba/core/pythonapi.py
@@ -827,11 +827,11 @@ class PythonAPI(object):
     def set_iterate(self, set):
         builder = self.builder
 
-        hashptr = cgutils.alloca_once(builder, self.py_hash_t, name="hashptr")
-        keyptr = cgutils.alloca_once(builder, self.pyobj, name="keyptr")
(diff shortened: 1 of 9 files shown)


gmarkall:
gpuci run tests

gmarkall (changes requested):
Dots are not permitted in identifiers in the NVVM IR specification: https://docs.nvidia.com/cuda/nvvm-ir-spec/index.html#identifiers

I'm not sure I see a way around this other than continuing to use the non-dot versions of the identifiers.

DrTodd13:
gpuci run tests

gmarkall (approved):
This looks good to me now that the CUDA changes are undone and the tests passing again.

I think the other code changes look good but I'm not familiar with the general convention or whether there's any other implications, so I'd suggest another reviewer comment on that rather than me marking this RTM myself now.

PR #9020, opened 2023-06-15
Fix #9008: memory leak in dispatcher.

### LINKED ISSUE #9008: Memory leak when passing string literal as an argument to a jitted function
As title says, when passing a literal string as an argument to a jitted function, it leaks memory.  I'm using numba 0.56.4 and python 3.10.6

Here is a working example to reproduce it

```
#!/usr/bin/env python3
import numpy as np
from numba import njit,version_info
import psutil,os
print(version_info,end="\n\n")
pythonProcess = psutil.Process(os.getpid())
def printMem():
    print("mem: %.2fM"%(float(pythonProcess.memory_info().rss)/1e6))


@njit
def leakfunction(string):
  return string

for _ in range(10):
   for _ in range(100000):
       leakfunction("test")
   printMem()
```

which produces the output

```
version_info(major=0, minor=56, patch=4, short=(0, 56), full=(0, 56, 4), string='0.56.4', tuple=('0', '56', '4'), git_revision=None)

mem: 134.29M
mem: 135.94M
mem: 136.84M
mem: 137.65M
mem: 138.46M
mem: 139.22M
mem: 140.03M
mem: 140.84M
mem: 141.65M
mem: 142.46M
```

sklam:
Fixes #9008.

Dispatcher holding references to types in direct call to `typeof_pyval` outside of `compile_for_args` is causing memory leak.

### DIFF
--- a/numba/core/dispatcher.py
+++ b/numba/core/dispatcher.py
@@ -254,7 +254,7 @@ class _DispatcherBase(_dispatcher.Dispatcher):
         # but newer python uses a different name
         self.__code__ = self.func_code
         # a place to keep an active reference to the types of the active call
-        self._types_active_call = []
+        self._types_active_call = None
         # Default argument values match the py_func
         self.__defaults__ = py_func.__defaults__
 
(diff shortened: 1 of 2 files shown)


swap357:
Closing this as the fix for original issue was merged with https://github.com/numba/numba/pull/9757

PR #9033, opened 2023-06-22
No compilation in tests

apmasell:
Remove compilation from existing tests and then add a detector to prevent compilation in future tests.

### DIFF
--- a/numba/core/bad_gen.py
+++ b/numba/core/bad_gen.py
@@ -1,0 +1,75 @@
+import llvmlite.binding as ll
+import llvmlite.ir as llvmir
+import numba.core.cgutils as cgutils
+from numba.core.codegen import Codegen, CodeLibrary
+
+
+class BadCodeLibrary(CodeLibrary):
+    """
+    A code library that just explodes constantly, for testing purposes
+    """
+
+    def __init__(self, codegen: "BadCodegen", name):
+        super().__init__(codegen, name)
+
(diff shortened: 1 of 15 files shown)


gmarkall (changes requested):
I've just had a quick look at the implementation so far (not the test changes, will look at that in the next pass). There's a few questions in the diff. In addition:

- I think the intention here is to ensure that compilation only occurs during test execution, not during import or discovery - is that correct? (It's hard to understand exactly the intention from the title / description).
- When do we run the checks that this makes possible? Can they be run on CI, e.g. by creating a job that sets `NUMBA_JIT_BACKEND=bad` and then lists tests?
- Is the name `NUMBA_JIT_BACKEND` going to take any more values in future? I imagine it'll also be usable for switching between MCJIT and OrcJIT - if that's the case, I think the name is OK. But if this is only going to be used for toggling between normal codegen and codegen disabled, perhaps making it more of a "true / false"-type setting would be clearer about its use.
- I wonder if there's compilation in the CUDA test setup too - presently we don't run the tests in parallel because CUDA gets initialized before the fork, but maybe if we ensured there's no compilation we could also run those tests in parallel (if that's the sole thing causing CUDA to get initialized before the fork, I don't recall). I (or we) should be able to investigate that following the completion of this PR (I dont want to suggest creeping the scope of this PR).

Issue #9032, opened 2023-06-22
Numba+Cuda doesn't work for me on WSL 2 (but PyTorch+Cuda works fine)


jorlow:
Cuda works fine in PyTorch on my WSL 2 (Windows 11 + 22.04.2 LTS) install but it doesn't work for numba.

$ python -c "import torch; print(torch.cuda.is_available())"
`True`

$ python -c "from numba import cuda; print(cuda.is_available())"
`False`

$ python -c "import numba; print(numba.__version__)"
`0.58.0dev0+190.gc9cc06ba1`

I found https://github.com/numba/numba/issues/7104 while trying to troubleshoot and everything I'm seeing pretty much matches what Sahuta saw. I've pasted the output from the various diagnostic commands he was asked to run below.

$ python -c "from numba import cuda; cuda.detect()"
```
Traceback (most recent call last):
  File "/home/jorlow/mambaforge/envs/course22p2/lib/python3.10/site-packages/numba/cuda/cudadrv/driver.py", line 258, in ensure_initialized
    self.cuInit(0)
  File "/home/jorlow/mambaforge/envs/course22p2/lib/python3.10/site-packages/numba/cuda/cudadrv/driver.py", line 331, in safe_cuda_api_call
    self._check_ctypes_error(fname, retcode)
  File "/home/jorlow/mambaforge/envs/course22p2/lib/python3.10/site-packages/numba/cuda/cudadrv/driver.py", line 399, in _check_ctypes_error
    raise CudaAPIError(retcode, msg)
numba.cuda.cudadrv.driver.CudaAPIError: [100] Call to cuInit results in CUDA_ERROR_NO_DEVICE

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/jorlow/mambaforge/envs/course22p2/lib/python3.10/site-packages/numba/cuda/api.py", line 470, in detect
    print('Found %d CUDA devices' % len(devlist))
  File "/home/jorlow/mambaforge/envs/course22p2/lib/python3.10/site-packages/numba/cuda/cudadrv/devices.py", line 49, in __len__
    return len(self.lst)
  File "/home/jorlow/mambaforge/envs/course22p2/lib/python3.10/site-packages/numba/cuda/cudadrv/devices.py", line 26, in __getattr__
    numdev = driver.get_device_count()
[...]

PR #9034, opened 2023-06-23
CUDA libs test: Report the absolute path of the loaded libcuda.so on Linux, + other improvements

gmarkall:
This PR improves the CUDA library test function. The most important change is to report the absolute path of the loaded `libcuda.so` on Linux. 

This motivation is as follows: various driver-related issues have been reported by WSL2 users (see e.g. #9032, #7104, https://github.com/rapidsai/cudf/issues/13388, other questions on Gitter, Slacks, Discourse, etc.) and it is almost always due to a Linux (i.e. not- WSL2) driver being installed in a WSL2 system. Printing the absolute path gives the version number of the driver, which can be looked up to determine whether it was a version intended for "native" Linux, or the correct WSL2 driver.

There are also some minor changes to the tests and output to improve consistency and readability - see individual commit messages for details. 

With these changes, the output of the test function (identically reported by the Numba sysinfo tool, `numba -s`) is, for example:

```
$ python -c "from numba import cuda; cuda.cudadrv.libs.test()"
Finding driver from candidates:
	libcuda.so
	libcuda.so.1
	/usr/lib/libcuda.so
	/usr/lib/libcuda.so.1
	/usr/lib64/libcuda.so
	/usr/lib64/libcuda.so.1
Using loader <class 'ctypes.CDLL'>
	Trying to load driver...	ok
		Loaded from libcuda.so
	Mapped libcuda.so paths:
		/usr/lib/x86_64-linux-gnu/libcuda.so.530.30.02
Finding nvvm from Conda environment
	Located at /home/gmarkall/mambaforge/envs/numbadev/lib/libnvvm.so.4.0.0
	Trying to open library...	ok
Finding cudart from Conda environment
	Located at /home/gmarkall/mambaforge/envs/numbadev/lib/libcudart.so.11.8.89
	Trying to open library...	ok
Finding cudadevrt from Conda environment
	Located at /home/gmarkall/mambaforge/envs/numbadev/lib/libcudadevrt.a
	Checking library...	ok
Finding libdevice from Conda environment
	Located at /home/gmarkall/mambaforge/envs/numbadev/lib/libdevice.10.bc
	Checking library...	ok
```

I would suggest that each commit can be reviewed individually - the changes in each are independent, and commit messages contain further details.

### DIFF
--- a/docs/upcoming_changes/9034.cuda.rst
+++ b/docs/upcoming_changes/9034.cuda.rst
@@ -1,6 +1,6 @@
 Report absolute path of libcuda.so on Linux
 ===========================================
 
-`numba -s` now reported the absolute path to libcuda.so on Linux, to aid
+``numba -s`` now reports the absolute path to ``libcuda.so`` on Linux, to aid
 troubleshooting driver issues, particularly on WSL2 where a Linux driver can
 incorrectly be installed in the environment.
--- a/docs/upcoming_changes/9034.cuda.rst
+++ b/docs/upcoming_changes/9034.cuda.rst
@@ -1,0 +1,6 @@
+Report absolute path of libcuda.so on Linux
+===========================================
+
+`numba -s` now reported the absolute path to libcuda.so on Linux, to aid
+troubleshooting driver issues, particularly on WSL2 where a Linux driver can
+incorrectly be installed in the environment.
--- a/numba/cuda/cudadrv/libs.py
+++ b/numba/cuda/cudadrv/libs.py
@@ -98,6 +98,35 @@ def test():
         print(f'\tERROR: failed to open driver: {e}')
         failed = True
 
+    # Find the absolute location of the driver on Linux. Various driver-related
+    # issues have been reported by WSL2 users, and it is almost always due to a
+    # Linux (i.e. not- WSL2) driver being installed in a WSL2 system.
+    # Providing the absolute location of the driver indicates its version
+    # number in the soname (e.g. "libcuda.so.530.30.02"), which can be used to
+    # look up whether the driver was intended for "native" Linux.
+    if sys.platform == 'linux' and not failed:
+        pid = os.getpid()
+        mapsfile = f'/proc/{pid}/maps'
+        try:
+            with open(mapsfile) as f:
+                maps = f.read()
+        # It's difficult to predict all that might go wrong reading the maps
+        # file - in case various error conditions ensue (the file is not found,
+        # not readable, etc.) we use OSError to hopefully catch any of them.
+        except OSError:
+            # It's helpful to report that this went wrong to the user, but we
+            # don't set failed to True because this doesn't have any connection
+            # to actual CUDA functionality.
+            print(f'\tERROR: Could not open {mapsfile} to determine absolute '
+                  'path to libcuda.so')
+        else:
+            # In this case we could read the maps, so we can report the
+            # relevant ones to the user
+            locations = set(s for s in maps.split() if 'libcuda.so' in s)
+            print('\tMapped libcuda.so paths:')
+            for location in locations:
+                print(f'\t\t{location}')
+
     # Checks for dynamic libraries
     libs = 'nvvm cudart'.split()
     for lib in libs:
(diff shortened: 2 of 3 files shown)


gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

bdice (commented):
I appreciate this — I’ve helped field a lot of questions from users that boil down to this core problem with WSL2 driver installation. I believe this will help with diagnosis.

The changes look fine to me.

numba/cuda/cudadrv/libs.py
```diff
@@ -80,67 +79,97 @@ def _get_source_variable(lib, static=False):
...
+    # look up whether the driver was intended for "native" Linux.
+    if sys.platform == 'linux' and not failed:
+        pid = os.getpid()
+        mapsfile = f'/proc/{pid}/maps'
+        try:
+            with open(mapsfile) as f:
+                maps = f.read()
+        # It's difficult to predict all that might go wrong reading the maps
+        # file - in case various error conditions ensue (the file is not found,
+        # not readable, etc.) we use OSError to hopefully catch any of them.
```
stuartarchibald, line 116: Another option, if this approach does end up to be more problematic than anticipated, would perhaps be to use `dl_iterate_phdr`.

docs/upcoming_changes/9034.cuda.rst
```diff
@@ -0,0 +1,6 @@
+Report absolute path of libcuda.so on Linux
+===========================================
+
+`numba -s` now reported the absolute path to libcuda.so on Linux, to aid
```
stuartarchibald, line 4:
```suggestion
``numba -s`` now reports the absolute path to ``libcuda.so`` on Linux, to aid
```
think RST needs double backticks for code statements? (and line wrap)

numba/cuda/cudadrv/libs.py
```diff
@@ -80,67 +79,97 @@ def _get_source_variable(lib, static=False):
...
+        print('\tTrying to load driver', end='...')
         dll, path = load_driver(dlloader, candidates)
-        print(f'\tok, loaded from {path}')
+        print('\tok')
+        print(f'\t\tLoaded from {path}')
     except CudaSupportError as e:
         print(f'\tERROR: failed to open driver: {e}')
         failed = True
 
+    # Find the absolute location of the driver on Linux. Various driver-related
```
stuartarchibald, line 102: The absolute path may contain information that users might not want to share, however the point of this diagnostic is that the absolute path is reported. The `numba -s` tool has a large printed warning at the end of its output with regards to checking and redacting the output before sharing which I imagine covers this.

numba/cuda/cudadrv/libs.py
```diff
@@ -80,67 +79,97 @@ def _get_source_variable(lib, static=False):
...
+    # issues have been reported by WSL2 users, and it is almost always due to a
+    # Linux (i.e. not- WSL2) driver being installed in a WSL2 system.
+    # Providing the absolute location of the driver indicates its version
+    # number in the soname (e.g. "libcuda.so.530.30.02"), which can be used to
+    # look up whether the driver was intended for "native" Linux.
+    if sys.platform == 'linux' and not failed:
+        pid = os.getpid()
+        mapsfile = f'/proc/{pid}/maps'
```
stuartarchibald, line 110:
```suggestion
        mapsfile = os.path.join('proc', f'{pid}', 'maps')
```

stuartarchibald (commented):
Thanks for the patch, couple of minor things to resolve else looks good. I've also tested this manually on a system with CUDA system level drivers, worked as expected.

docs/upcoming_changes/9034.cuda.rst
gmarkall: Ah, I forgot this was RST! (despite it being in the name)

stuartarchibald (dismissed):

stuartarchibald (approved):
Thanks for fixing-up my previous suggestion via 98784c4, `proc` does indeed need accessing from the file system root. Patch re-approved!

gmarkall:
gpuci run tests

PR #9035, opened 2023-06-23
CUDA: Allow for debuginfo in nvdisasm output

Matt711:
I'm building some jupyter magic tooling as described in #8960.  I need to be able to map the SASS to original source lines. This PR adds debug info flag `-gi` to the `nvdisasm` command to get line information. For example output, see below:

```python
from numba import cuda
import numpy as np
@cuda.jit(lineinfo=True)
def gpu_add_one_kernel(x, out, N):
    i = cuda.grid(1)
    if i<N:
        out[i] = x[i] + 1

a = np.arange(4096,dtype=np.int32)
d_a = cuda.to_device(a)
d_out = cuda.device_array_like(d_a)
blocks_per_grid = 32*2
threads_per_block = 128
gpu_add_one_kernel[blocks_per_grid, threads_per_block](d_a, d_out, len(d_a))
cuda.synchronize()
print(next(iter(gpu_add_one_kernel.inspect_sass(debuginfo=True).items()))[1])
```

```asm
...
	//## File "/tmp/ipykernel_2373722/1083830147.py", line 3
        /*0000*/                   MOV R1, c[0x0][0x28] ;
	//## File "/tmp/ipykernel_2373722/1083830147.py", line 5 inlined at "/tmp/ipykernel_2373722/1083830147.py", line 3
	//## File "/tmp/ipykernel_2373722/1083830147.py", line 3
        /*0010*/                   S2R R0, SR_CTAID.X ;
        /*0020*/                   S2R R3, SR_TID.X ;
        /*0030*/                   IMAD R0, R0, c[0x0][0x0], R3 ;
	//## File "/tmp/ipykernel_2373722/1083830147.py", line 6 inlined at "/tmp/ipykernel_2373722/1083830147.py", line 3
	//## File "/tmp/ipykernel_2373722/1083830147.py", line 3
        /*0040*/                   ISETP.GE.U32.AND P0, PT, R0, c[0x0][0x1d0], PT ;
        /*0050*/                   SHF.R.S32.HI R4, RZ, 0x1f, R0 ;
        /*0060*/                   ISETP.GE.AND.EX P0, PT, R4, c[0x0][0x1d4], PT, P0 ;
        /*0070*/               @P0 EXIT ;
	//## File "/tmp/ipykernel_2373722/1083830147.py", line 7 inlined at "/tmp/ipykernel_2373722/1083830147.py", line 3
	//## File "/tmp/ipykernel_2373722/1083830147.py", line 3
        /*0080*/                   ISETP.GE.AND P0, PT, R0, RZ, PT ;
        /*0090*/                   ULDC.64 UR4, c[0x0][0x118] ;
        /*00a0*/                   SEL R3, RZ, c[0x0][0x188], P0 ;
        /*00b0*/                   SEL R5, RZ, c[0x0][0x18c], P0 ;
        /*00c0*/                   IADD3 R3, P1, R0, R3, RZ ;
        /*00d0*/                   IMAD.X R6, R4, 0x1, R5, P1 ;
        /*00e0*/                   LEA R2, P1, R3, c[0x0][0x180], 0x2 ;
        /*00f0*/                   LEA.HI.X R3, R3, c[0x0][0x184], R6, 0x2, P1 ;
        /*0100*/                   LDG.E R2, [R2.64] ;
        /*0110*/                   SEL R5, RZ, c[0x0][0x1c0], P0 ;
        /*0120*/                   SEL R7, RZ, c[0x0][0x1c4], P0 ;
        /*0130*/                   IADD3 R0, P0, R0, R5, RZ ;
        /*0140*/                   IMAD.X R5, R4, 0x1, R7, P0 ;
        /*0150*/                   LEA R4, P0, R0, c[0x0][0x1b8], 0x2 ;
        /*0160*/                   LEA.HI.X R5, R0, c[0x0][0x1bc], R5, 0x2, P0 ;
        /*0170*/                   IADD3 R7, R2, 0x1, RZ ;
        /*0180*/                   STG.E [R4.64], R7 ;
	//## File "/tmp/ipykernel_2373722/1083830147.py", line 3
        /*0190*/                   EXIT ;
...
```

### DIFF
--- a/numba/cuda/codegen.py
+++ b/numba/cuda/codegen.py
@@ -13,7 +13,7 @@ import tempfile
 CUDA_TRIPLE = 'nvptx64-nvidia-cuda'
 
 
-def disassemble_cubin(cubin):
+def disassemble_cubin(cubin, debuginfo=False):
     # nvdisasm only accepts input from a file, so we need to write out to a
     # temp file and clean up afterwards.
     fd = None
@@ -24,9 +24,14 @@ def disassemble_cubin(cubin):
             f.write(cubin)
 
         try:
-            cp = subprocess.run(['nvdisasm', fname], check=True,
-                                stdout=subprocess.PIPE,
-                                stderr=subprocess.PIPE)
+            if debuginfo:
+                cp = subprocess.run(['nvdisasm', '-gi', fname], check=True,
+                                    stdout=subprocess.PIPE,
+                                    stderr=subprocess.PIPE)
+            else:
+                cp = subprocess.run(['nvdisasm', fname], check=True,
+                                    stdout=subprocess.PIPE,
+                                    stderr=subprocess.PIPE)
         except FileNotFoundError as e:
             msg = ("nvdisasm is required for SASS inspection, and has not "
                    "been found.\n\nYou may need to install the CUDA "
@@ -198,8 +203,8 @@ class CUDACodeLibrary(serialize.ReduceMixin, CodeLibrary):
         except KeyError:
             raise KeyError(f'No linkerinfo for CC {cc}')
 
-    def get_sass(self, cc=None):
-        return disassemble_cubin(self.get_cubin(cc=cc))
+    def get_sass(self, cc=None, debuginfo=False):
+        return disassemble_cubin(self.get_cubin(cc=cc), debuginfo=debuginfo)
 
     def add_ir_module(self, mod):
         self._raise_if_finalized()
--- a/numba/cuda/codegen.py
+++ b/numba/cuda/codegen.py
@@ -13,7 +13,7 @@ import tempfile
 CUDA_TRIPLE = 'nvptx64-nvidia-cuda'
 
 
-def disassemble_cubin(cubin, debuginfo=False):
+def disassemble_cubin(cubin):
     # nvdisasm only accepts input from a file, so we need to write out to a
     # temp file and clean up afterwards.
     fd = None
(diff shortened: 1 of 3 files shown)


gmarkall:
gpuci run tests

numba/cuda/codegen.py
```diff
@@ -24,7 +24,7 @@ def disassemble_cubin(cubin):
             f.write(cubin)
 
         try:
-            cp = subprocess.run(['nvdisasm', fname], check=True,
+            cp = subprocess.run(['nvdisasm', '-gi', fname], check=True,
```
gmarkall, line 27: Could you just add a comment mentioning that `-gi` adds line info? Or maybe better yet, do `lineinfo_flag = '-gi'` on the previous line and use that in the list of arguments, so the code is self-explaining?

numba/cuda/tests/cudapy/test_inspect.py
```diff
@@ -98,6 +98,9 @@ def _test_inspect_sass(self, kernel, name, sass):
                 seen_function = True
         self.assertTrue(seen_function)
 
+        seen_lineinfo = 'line' in sass and 'inlined' in sass
```
gmarkall, line 101: I worry that this might be a little fragile - would checking for `//## File` in the output rather than `line` (which might have false positives from the kernel name) and `inlined` (which might not be found if a future toolkit makes a different inlining decision) be more robust? Or even maybe checking for a match of something like:

```
r'//## File ".*/test_inspect.py", line [0-9]'
```

and `self.assertRegex()`? I'm not good at regex off the top of my head so that might need fixing up, but the intention is to check the line from the beginning up until the line number, ignoring that the paths might vary

gmarkall:
gpuci run tests

numba/cuda/tests/cudapy/test_inspect.py
```diff
@@ -91,13 +92,17 @@ def foo(x, y):
         self.assertIn("foo", asmdict[float64, float64])
 
     def _test_inspect_sass(self, kernel, name, sass):
+        if not NVVM().is_nvvm70:
+            self.skipTest("lineinfo not generated for NVVM 3.4")
```
gmarkall, line 96: Whilst I'd normally mention that this also ends up skipping the tests that specific SASS instructions appear later on in the test, which should work even for NVVM 3.4, it doesn't matter in this case because we're going to drop support for all NVVM 3.4-based toolkits before the next release.

gmarkall (approved):
Many thanks for the improvement! This is looking good now!

I think this should have a CUDA buildfarm run in case something weird and completely different happens on Windows (you never know!), but if that's OK then I think this is good to merge.

gmarkall:
I've run this locally and it seems OK with my Windows setup. However, I still think a Buildfarm run would be good to make sure it's OK in that environment too.

gmarkall:
@esc Could this have a CUDA buildfarm run please?

esc:
`Build numba_smoketest_cuda_yaml_207 has started`

esc:
> `Build numba_smoketest_cuda_yaml_207 has started`

Has passed!

esc:
@gmarkall is this now RTM?

PR #9036, opened 2023-06-23
installing.rst documation fix

ellifteria:
I updated the link in `docs/source/user/installing.rs` in the section:

Installing from source
----------------------

from

```
git clone git://github.com/numba/numba.git
```

to 

```
git clone https://github.com/numba/numba.git
```


gmarkall:
The fix seems good, but there's something up with the history of your branch which is making this PR very weird - it looks like its history diverged from the Numba repository's `main` branch, so it appears as if it incorporates a lot of changes. I don't know what's gone wrong so I'd suggest the quickest way to resolve this might be to create a new branch from Numba's `main` branch, make a single commit with the change on it, and then open a new PR from that branch. What do you think of this plan?

gmarkall:
Many thanks! Closing in favour of #9039.

PR #9037, opened 2023-06-23
Recognize additional functions as being pure or not having side effects.

DrTodd13:
Adds just a few functions to be recognized as pure and not having side effects.

Also split dump_blocks into a dump routine for a single block so that Numba extensions can dump just a single block if it wishes.

### DIFF
--- a/numba/core/ir_utils.py
+++ b/numba/core/ir_utils.py
@@ -4,6 +4,7 @@
 #
 
 import numpy
+import math
 
 import types as pytypes
 import collections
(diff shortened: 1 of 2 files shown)


DrTodd13:
@stuartarchibald Maybe we could have a function that asserts that the function it wraps has no side effects.  We can put really common stuff in the default list but for uncommon stuff this could be a way to still get the same effect.

numba/core/ir_utils.py
```diff
@@ -1857,11 +1864,14 @@ def gen_np_call(func_as_str, func, lhs, args, typingctx, typemap, calltypes):
     np_assign = ir.Assign(np_call, lhs, loc)
     return [g_np_assign, attr_assign, np_assign]
 
+def dump_block(label, block):
+    print(label, ":")
+    for stmt in block.body:
+        print("    ", stmt)
+
 def dump_blocks(blocks):
```
stuartarchibald, line 1872: At present this debug routine is used solely when `NUMBA_DEBUG_ARRAY_OPT` is set, it continues to work as expected after the split.

stuartarchibald (approved):
Thanks for the patch. As perhaps hinted at in https://github.com/numba/numba/pull/9037#issuecomment-1605046483, a more broad approach may be needed in the long term.

PR #9039, opened 2023-06-26
Correct git clone link in installation instructions.

ellifteria:
I updated the link in `docs/source/user/installing.rs` in the section:

Installing from source
----------------------

from

```
git clone git://github.com/numba/numba.git
```

to 

```
git clone https://github.com/numba/numba.git
```

### DIFF
--- a/docs/source/user/installing.rst
+++ b/docs/source/user/installing.rst
@@ -110,7 +110,7 @@ If you are building Numba from source for other reasons, first follow the
 Once that is completed, you can download the latest Numba source code from
 `Github <https://github.com/numba/numba>`_::
 
-    $ git clone git://github.com/numba/numba.git
+    $ git clone https://github.com/numba/numba.git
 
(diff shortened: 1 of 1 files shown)


gmarkall (approved):
Many thanks for the fix! I've manually check that the old URL didn't work, and the new one does.

esc:
@ellifteria thank you for the PR. @gmarkall thank you for the review.

I was a bit surprised by this, since I expected the `git` protocol to just work. But it seems like `github` doesn't support this anymore, it now just seems to timeout.

```
 💣 zsh» time git clone git://github.com/numba/numba.git                                                                                                                                                                     
Cloning into 'numba'...
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
git clone git://github.com/numba/numba.git  0.01s user 0.02s system 0% cpu 16.123 total
```

PR #9040, opened 2023-06-27
Remove NVVM 3.4 and CTK 11.0 / 11.1 support

gmarkall:
This removes support for CUDA toolkits 11.0 and 11.1. A large side-effect of this is that there are no longer any supported toolkits that use NVVM 3.4, so all NVVM 3.4-related code and tests can be dropped.

I decided against removing support for Kepler (CC 3.5 and 3.7) devices for the 0.58 release, because keeping it doesn't seem to be introducing any additional burden, and as long as the user has an 11.x toolkit installed, their toolkit will support it. It might be just fine to keep supporting these for more than one release into the future, so I've changed the deprecation notice to avoid specifying when CC 3.5 and 3.7 support will be removed.

The CI configuration is updated to remove the 11.0 and 11.1 configurations. I needed one more gpuCI configuration to be able to test all supported NumPy versions with CUDA, so I've (somewhat arbitrarily) added 11.3 into the matrix.


gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

numba/cuda/cudadrv/nvvm.py
```diff
@@ -363,6 +354,7 @@ def get_log(self):
     (11, 8): ((3, 5), (9, 0)),
     (12, 0): ((5, 0), (9, 0)),
     (12, 1): ((5, 0), (9, 0)),
+    (12, 2): ((5, 0), (9, 0)),
```
stuartarchibald, line 357: This is adding 12.2 whereas this patch concerns moving the baseline to 11.2?

stuartarchibald (commented):
Thanks for the patch, this looks good on inspection. A local search of the code base for patterns that might match `3.4` or are related to `11.0/11.1` yields a couple of instances in the docs that might warrant adjustment:
https://github.com/numba/numba/blob/766f6d9b250ebf452c75590f753de0c003563541/docs/source/cuda/cooperative_groups.rst?plain=1#L108
https://github.com/numba/numba/blob/766f6d9b250ebf452c75590f753de0c003563541/docs/source/cuda/minor_version_compatibility.rst?plain=1#L10-L13

gmarkall:
> Thanks for the patch, this looks good on inspection. A local search of the code base for patterns that might match `3.4` or are related to `11.0/11.1` yields a couple of instances in the docs that might warrant adjustment:

Thanks - I think these should stay the same - the max blocks count (1152) may vary with CUDA driver and device, so it's not particularly invalid to state that example. The point was to show that a user running the example might get a different number with a different driver or device.

The second example is also relevant - CUDA driver 450 (CUDA 11.0) is supported because it supported later toolkits through MVC - it's just the toolkits we're dropping support for (as we require "a driver sufficient for the toolkit version").

numba/cuda/cudadrv/nvvm.py
gmarkall: Yes, since this is a one-liner to add with no changed support, I added this in.

stuartarchibald: OK, thanks, let's leave it as-is given support isn't changing.

stuartarchibald:
RE: https://github.com/numba/numba/pull/9040#issuecomment-1651841755

Thanks for considering these @gmarkall, am happy to leave them based on your assessment.

sklam:
BFID: `numba_smoketest_cuda_yaml_210`

stuartarchibald:
> BFID: `numba_smoketest_cuda_yaml_210`

This "passed", the CUDA 11.1 builds failed as would be expected as this PR removes support.

PR #9044, opened 2023-06-27
Modify test cases to not compile-on-list

apmasell:
Changes tests that inadvertently trigger the compiler when doing test discovery/listing.

Split off from #9033

### DIFF
--- a/numba/tests/chained_assign_usecases.py
+++ b/numba/tests/chained_assign_usecases.py
@@ -1,0 +1,69 @@
+from numba import jit
+import numpy as np
+
+
+@jit
+def inc(a):
+    for i in range(len(a)):
+        a[i] += 1
+    return a
+
+
+@jit
+def inc1(a):
+    a[0] += 1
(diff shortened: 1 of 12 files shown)


stuartarchibald:
This test class may prove useful in testing this patch:
```python
from unittest import TestCase
import sys
import subprocess

class TestNoCompile(TestCase):

    def test_no_compiling(self):

        code = """if 1:
        from unittest import mock
        with mock.patch.object(llvm.ExecutionEngine, 'finalize_object',
                               side_effect=RuntimeError("Compiling during listing")):
            import numba
            numba.test('-l')
        """
        cmd = [sys.executable, "-c", code]
        subprocess.check_call(cmd)
```
at present, it identifies that the `test_stencils.py` tests are accessing the LLVM compilation chain. If that test module is removed entirely, the test suite listing appears to not invoke compilation. Hope this helps?

apmasell:
That method will not be called when using OrcJIT.

PR #9048, opened 2023-06-28
Dynamically allocate parfor schedule.

### LINKED ISSUE #9029: Segfault when using assert with parallel=True
The following code triggers a segfault. Tested on Ubuntu 22.04 with Numba 0.56.4 and Numba 0.57.0.

```python
from numba import njit
import numpy as np

@njit(parallel=True)
def foo():
    n = 1000

    for _ in range(n):
        for _ in range(n):
            values = np.zeros(3)

            np.max(values)

            assert n >= 0

foo()
```

DrTodd13:
Resolves #9029 

Move from stack allocation of parfor schedules to dynamic allocation.  This will prevent stack overflow if there is a large loop with parfors in it.

### DIFF
--- a/numba/parfors/parfor_lowering.py
+++ b/numba/parfors/parfor_lowering.py
@@ -1668,7 +1668,9 @@ def call_parallel_gufunc(lowerer, cres, gu_signature, outer_sig, expr_args, expr
     intp_t = context.get_value_type(types.intp)
     uintp_t = context.get_value_type(types.uintp)
     intp_ptr_t = llvmlite.ir.PointerType(intp_t)
+    intp_ptr_ptr_t = llvmlite.ir.PointerType(intp_ptr_t)
     uintp_ptr_t = llvmlite.ir.PointerType(uintp_t)
+    uintp_ptr_ptr_t = llvmlite.ir.PointerType(uintp_ptr_t)
     zero = context.get_constant(types.uintp, 0)
     one = context.get_constant(types.uintp, 1)
     one_type = one.type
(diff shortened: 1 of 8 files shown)


esc:
As release manager, I have decided that this isn't mature enough to be ready for the upcoming 0.58 release (next week) and as such I am moving this into the next release.

sklam:
I confirmed that this patch fixes https://github.com/numba/numba/issues/9167.

sklam (commented):
Code changes look good.

Consideration on lifetime of the dynamic allocation:
- The call to gufunc is essentially straightline code.
- There's a guard for single-entry single-exit in parfor loop.
- Exceptions thrown within the parloop body is eaten/ignored.
- So, as long as the alloc/free are paired up there shouldn't be leaks.

numba/parfors/parfor_lowering.py
```diff
@@ -1747,11 +1757,27 @@ def load_range(v):
                                                   context.get_constant(types.uintp, num_dim),
                                                   dim_starts,
                                                   dim_stops])
+    # Set the chunksize to zero so that any nested calls get the default chunk size behavior.
     builder.call(set_chunksize, [zero])
 
+    # Each entry in the schedule is 2 times the number of dimensions long.
     multiplier = context.get_constant(types.uintp, num_dim * 2)
+    # Compute the total number of entries in the schedule.
     sched_size = builder.mul(num_divisions, multiplier)
```
sklam, line 1766: Is there a statically known maximum for `num_divisions`?

If yes, it is possible to use `alloca_once(max_sched_size)` to avoid the malloc/free.

sklam (commented):
Consideration for avoiding malloc/free.

numba/parfors/parfor_lowering.py
DrTodd13: With the chunking API, if you want, you can have each point in the iteration space be its own division of work and so the maximum divisions is bounded only by memory size.  The only way to be sure you'd never have to reallocate something larger is to allocate all of memory to the schedule.  This PR was designed for a possible underlying caching implementation so rather than allocate and deallocate each time, the deallocate routine can actually hold onto the memory and reuse it in the next allocation if it is no larger than the previous one.

stuartarchibald (approved):
Thanks for fixing this @DrTodd13. This seems like a reasonable approach given the ability to handle exceptions arising in parallel regions is not implemented. I agree with @sklam's analysis (thanks for reviewing @sklam) and the code looks good on inspection, the addition of commentary to the code generation in parfor lowering is appreciated. Thanks again!

PR #9056, opened 2023-07-04
adding weekly meeting notes script

esc:
Moving from the Gist at:

https://gist.github.com/sklam/e526a302eaefd405dbb2bae52d075012

To the `numba/maint` directory.

### DIFF
--- a/maint/github_weekly_meeting.py
+++ b/maint/github_weekly_meeting.py
@@ -1,9 +1,7 @@
 #! /usr/bin/env python
 
-import sys, os
-from datetime import datetime, timedelta
-from pprint import pprint
-from getpass import getpass
+import sys
+import os
 
 from github3 import login
 import github3
@@ -54,6 +52,7 @@ def fetch(orgname, reponame, last_num, gh):
         'max_iss_num': max_iss_num,
     }
 
+
 def display(data):
     print("## 1. New Issues")
     for line in reversed(data['opened_issues']):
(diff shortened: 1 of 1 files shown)


maint/github_weekly_meeting.py
```diff
@@ -0,0 +1,123 @@
...
+from datetime import datetime, timedelta
+from pprint import pprint
+from getpass import getpass
+
+from github3 import login
+import github3
+
+
+def fetch(target_repo, last_num, gh):
+    orgname, reponame = target_repo.split('/', 1)
```
apmasell, line 13: I think these should just come in as parameters since `target_repo` is never used

maint/github_weekly_meeting.py
```diff
@@ -0,0 +1,123 @@
...
+        merged = False
+        if issue.pull_request_urls:
+            # Is PR?
+            merged = bool(info['pull_request'].get("merged_at"))
+            where = {'opened': opened_prs, 'closed': closed_prs}
+        else:
+            # Is Issues
+            where = {'opened': opened_issues, 'closed': closed_issues}
+
+        fmt = '- [{}#{}]({}) - {}'
```
apmasell, line 41: I would replaced this with an f-string and a `{' - merged' if merged else ''}` for readability.

maint/github_weekly_meeting.py
esc: fixed in f681a847495e10b3e6f8e2eb71a2f233764ba298

maint/github_weekly_meeting.py
esc: fixed in f681a847495e10b3e6f8e2eb71a2f233764ba298

esc:
@apmasell I also fixed up all the flake8 issues in a194e345db6671bb217000e87d879041ae01338f

apmasell (approved):

PR #9057, opened 2023-07-04
CUDA: Remove specialize in `ForAll` object

Matt711:
I noticed in the following example that the return of `CUDADispatcher.signatures` was empty.  This PR removes `specialize` in `dispatcher.py`

### DIFF
--- a/numba/cuda/dispatcher.py
+++ b/numba/cuda/dispatcher.py
@@ -482,17 +482,12 @@ class ForAll(object):
         if self.ntasks == 0:
             return
 
-        if self.dispatcher.specialized:
-            specialized = self.dispatcher
-        else:
-            specialized = self.dispatcher.specialize(*args)
-        blockdim = self._compute_thread_per_block(specialized)
+        blockdim = self._compute_thread_per_block(args)
+        griddim = (self.ntasks + blockdim - 1) // blockdim
+        self.dispatcher[griddim, blockdim, self.stream, self.sharedmem](*args)
         griddim = (self.ntasks + blockdim - 1) // blockdim
 
-        return specialized[griddim, blockdim, self.stream,
-                           self.sharedmem](*args)
-
-    def _compute_thread_per_block(self, dispatcher):
+    def _compute_thread_per_block(self, args):
         tpb = self.thread_per_block
         # Prefer user-specified config
         if tpb != 0:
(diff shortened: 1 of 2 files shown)


gmarkall:
gpuci run tests

numba/cuda/dispatcher.py
```diff
@@ -482,27 +482,24 @@ def __call__(self, *args):
...
 
-        if self.dispatcher.specialized:
-            specialized = self.dispatcher
-        else:
-            specialized = self.dispatcher.specialize(*args)
-        blockdim = self._compute_thread_per_block(specialized)
+        blockdim = self._compute_thread_per_block(args)
+        griddim = (self.ntasks + blockdim - 1) // blockdim
+        self.dispatcher[griddim, blockdim, self.stream, self.sharedmem](*args)
         griddim = (self.ntasks + blockdim - 1) // blockdim
```
gmarkall, line 488: This seems superfluous now - if I'm reading the diff right it recomputes `griddim` just before returning?

numba/cuda/dispatcher.py
```diff
@@ -482,27 +482,24 @@ def __call__(self, *args):
...
         else:
             ctx = get_context()
-            # Dispatcher is specialized, so there's only one definition - get
-            # it so we can get the cufunc from the code library
-            kernel = next(iter(dispatcher.overloads.values()))
+            typingctx = self.dispatcher.typingctx
+            argtypes = tuple([typingctx.resolve_argument_type(a) for a in args])
+            kernel = self.dispatcher.overloads.get(argtypes)
+            if not kernel:
+                kernel = self.dispatcher.compile(argtypes)
```
gmarkall, line 502: It looks like `compile()` checks if there's already a compiled version, so maybe this can be simplified to:

```suggestion
            kernel = self.dispatcher.compile(argtypes)
```

gmarkall (commented):
Thanks for the PR! There are a couple of initial comments on the code.

gmarkall (changes requested):
Many thanks for having a go at removing `CUDADispatcher.specialize()`... Unfortunately I wasn't clear when we talked earlier, and I'm sorry about that. I really do want to get rid of the whole "specialized dispatchers" thing, but it's a public API so we can't just get rid of it - first we need to figure out if there's an impact on users and deprecate it for at least one release before removing it.

What I'd intended to convey was that you should only make the changes I'd already suggested on the diff, which continue with removing the use of `specialize()` from the `ForAll` class. Could you please back out the removal of `specialize()`, and only keep the removal of its use in `ForAll`?

gmarkall:
gpuci run tests

PR #9059, opened 2023-07-05
Make sure all parameter names match NumPy and CPython

### LINKED ISSUE #9053: Numpy Gamma distribution not working in Numba 0.57+
## Reporting a bug

- [ x] I have tried using the latest released version of Numba (most recent is

Vectorization of Numpy's gamma distribution no longer works. The below code works in Numba 0.56.4, but not in 0.57.0 or 0.57.1. 

```
from numba import float64, vectorize
from numpy.random import gamma
import numpy as np

@vectorize([float64(float64, float64)])
def vec_ig(shape, scale):
    ig = 1. / gamma(shape=shape, scale=1. / scale)
    return ig

a = np.array([0.1, 0.1], dtype=np.float64)
b = np.array([0.1, 0.1], dtype=np.float64)
res = vec_ig(a, b)
```

The error that I receive is:

```
Compilation is falling back to object mode WITHOUT looplifting enabled because Function "vec_ig" failed type inference due to: No implementation of function Function(<built-in method gamma of numpy.random.mtrand.RandomState object at 0x7f7ae5b65840>) found for signature:
 
 >>> gamma(shape=float64, scale=float64)
 
There are 6 candidate implementations:
  - Of which 6 did not match due to:
  Overload in function 'gammavariate_impl': File: numba/cpython/randomimpl.py: Line 774.
    With argument(s): '(shape=float64, scale=float64)':
     TypingError: missing a required argument: 'alpha'
  raised from /home/devin/.cache/pypoetry/virtualenvs/pybuc-opbxzPBX-py3.9/lib64/python3.9/site-packages/numba/core/typing/templates.py:784
```

Expected behavior:

Return an array of size 2 with random draws from the gamma distribution.

apmasell:
As titled.
Closes #9053

### DIFF
--- a/numba/cpython/randomimpl.py
+++ b/numba/cpython/randomimpl.py
@@ -198,30 +198,39 @@ def get_next_int(context, builder, state_ptr, nbits, is_numpy):
 
 
 @overload(random.seed)
-def seed_impl(seed):
-    if isinstance(seed, types.Integer):
-        return _seed_impl('py')
+def seed_impl(a):
+    if isinstance(a, types.Integer):
+        @intrinsic
+        def _impl(typingcontext, seed):
+            def codegen(context, builder, sig, args):
+                seed_value, = args
+                fnty = ir.FunctionType(ir.VoidType(), (rnd_state_ptr_t, int32_t))
+                fn = cgutils.get_or_insert_function(builder.function.module, fnty,
+                                                    'numba_rnd_init')
+                builder.call(fn, (get_state_ptr(context, builder, 'py'),
+                                  seed_value))
+                return context.get_constant(types.none, None)
+
+            return signature(types.void, types.uint32), codegen
+
+        return lambda a: _impl(a)
 
 
 @overload(np.random.seed)
 def seed_impl(seed):
     if isinstance(seed, types.Integer):
-        return _seed_impl('np')
-
-
-def _seed_impl(state_type):
-    @intrinsic
-    def _impl(typingcontext, seed):
-        def codegen(context, builder, sig, args):
-            seed_value, = args
-            fnty = ir.FunctionType(ir.VoidType(), (rnd_state_ptr_t, int32_t))
-            fn = cgutils.get_or_insert_function(builder.function.module, fnty,
-                                                'numba_rnd_init')
-            builder.call(fn, (get_state_ptr(context, builder, state_type),
-                              seed_value))
-            return context.get_constant(types.none, None)
-        return signature(types.void, types.uint32), codegen
-    return lambda seed: _impl(seed)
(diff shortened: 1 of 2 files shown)


numba/cpython/randomimpl.py
```diff
@@ -625,9 +634,9 @@ def np_uniform_impl0():
 
 
 @overload(random.uniform)
-def uniform_impl1(low):
-    if isinstance(low, (types.Float, types.Integer)):
-        return lambda low: random.uniform(low, 1.0)
+def uniform_impl1(a):
+    if isinstance(a, (types.Float, types.Integer)):
+        return lambda a: random.uniform(a, 1.0)
```
guilhermeleobas, line 639: `random.uniform` expects two arguments: `a` and `b`
https://docs.python.org/3/library/random.html#random.uniform

I guess this one is not really used/tested in Numba test suite.

guilhermeleobas (commented):
Thanks, @apmasell. I've checked all the implementations and the changes match with NumPy/CPython random module. There's only one implementation that I think could be removed, as it is not used.

Also, some of the changes exceed the 80 char limit, but I guess this is fine as flake8 doesn't run in this file.

numba/cpython/randomimpl.py
apmasell: The differing number of arguments are overloaded separately. `def uniform_impl2(a, b):` is below. None of these are called by name from the test suite.

apmasell (commented):
I believe all the changed lines are within the limit

numba/cpython/randomimpl.py
guilhermeleobas: Got it. But I think `random.uniform` takes two arguments: `a` and `b`. See:

```python
$ python
Python 3.10.12 | packaged by conda-forge | (main, Jun 23 2023, 22:41:52) [Clang 15.0.7 ] on darwin
>>> import random
>>> random.uniform(2)
TypeError: Random.uniform() missing 1 required positional argument: 'b'
```

guilhermeleobas:
> I believe all the changed lines are within the limit

Some lines go beyond the 80 char limit. See:

image

But I think that's fine as this file is not checked by flake8.

numba/cpython/randomimpl.py
apmasell: Okay. I can delete it, but that will go in a separate PR. It's out of scope for this change.

guilhermeleobas (approved):

stuartarchibald:
Are there any tests in this PR that demonstrate that the reported issue (#9053), and associated discovered issues, are fixed?

apmasell:
No.

PR #9067, opened 2023-07-09
Get LLVM lines per Python line

Matt711:
Needed as a part of #8960. For example:

```python
from numba import njit

@njit
def func(x,y):
    x+=x
    y+=y
    y+=x
    x+=y
    return x+y

func(1,2)
```
```python
>>> print(func.overloads[func.signatures[0]].fndesc.global_dict)
{'python_llvm_lines': OrderedDict([(3,
               OrderedDict([('entry', []), ('B0', [])])),
              (5,
               OrderedDict([('entry', []),
                            ('B0',
                             ['%".6" = add nsw i64 %"arg.x", %"arg.x"'])])),
              (6,
               OrderedDict([('entry', []),
                            ('B0',
                             ['%".7" = add nsw i64 %"arg.y", %"arg.y"'])])),
              (7,
               OrderedDict([('entry', []),
                            ('B0', ['%".8" = add nsw i64 %".7", %".6"'])])),
              (8,
               OrderedDict([('entry', []),
                            ('B0', ['%".9" = add nsw i64 %".6", %".8"'])])),
              (9,
               OrderedDict([('entry', []),
                            ('B0',
                             ['%".10" = add nsw i64 %".9", %".8"',
                              'store i64 %".10", i64* %"retptr"',
                              'ret i32 0'])]))]),
 'llvm_func_def': 'define i32 @"func"(i64* noalias nocapture %"retptr", {i8*, i32, i8*, i8*, i32}** noalias nocapture %"excinfo", i64 %"arg.x", i64 %"arg.y")'}
```


numba/core/lowering.py
```diff
@@ -69,6 +70,8 @@ def __init__(self, context, library, fndesc, func_ir, metadata=None):
                                       cgctx=context,
                                       directives_only=directives_only)
 
+        self.fndesc.global_dict = {"llvm_lines": OrderedDict()}
```
Matt711, line 73: What's the best way to optionally expose the llvm lines to a Dispatcher object?

Matt711: I'll add a test after this is resolved.

numba/core/lowering.py
```diff
@@ -593,6 +644,17 @@ def lower_inst(self, inst):
         else:
             raise NotImplementedError(type(inst))
 
+        after = self._get_llvm_module_blocks()
+        diff = self._diff_llvm(before, after)
+        print(line, diff)
```
gmarkall, line 649: I suspect this is causing your test failures.

PR #9072, opened 2023-07-12
Fix support for overloading dispatcher with non-compatible first-class functions

### LINKED ISSUE #9071: Cannot allow two signatures with `FunctionType`
After finding the case 5 in https://github.com/numba/numba/issues/6972#issuecomment-828302957,

I tried to specify two sigs for one caller, but numba will complain with 
```python
Unknown attribute 'sum' of type int64

File "~/numba/tests/test_multi_function_type.py", line 33:
def callee_array(c):
    return c.sum()
    ^

During: typing of get attribute at ~/numba/tests/test_multi_function_type.py (33)
```
My test script, on 0.56.3:
```python
import numpy as np
from numba import njit, types, typeof
from numba.types import int64, float64
from numba.tests.support import (
    MemoryLeakMixin, TestCase
)

import os

os.environ['NUMBA_CAPTURED_ERRORS'] = 'new_style'
os.environ['NUMBA_DEBUG_CACHE'] = '1'

nb_array = typeof(np.ones(2))
callee_int_type = types.FunctionType(int64(int64))
sig_int = int64(callee_int_type, int64)
callee_array_type = types.FunctionType(float64(nb_array))
sig_array = float64(callee_array_type, nb_array)

# @njit(sig_int, cache=True)
# @njit(sig_array, cache=True)
@njit([sig_array, sig_int], cache=True)
def caller(callee, a):
    return callee(a)


@njit(cache=True)
def callee_int(b):
    return b


@njit(cache=True)
def callee_array(c):
    return c.sum()


class TestMultiFunctionType(MemoryLeakMixin, TestCase):

    def test_base(self):
        b = 1
        c = np.ones(2)

        print(caller(callee_int, b))
        print(caller(callee_array, c))


if __name__ == "__main__":
    clz = TestMultiFunctionType()
    clz.test_base()

```
If only specify one sig, compilation and cache is fine. But if specify with two sigs, error happens.

sklam:
- Make it clear that `Dispatcher.get_compile_result` is raising ``NumbaError``.
- Fix casting from `Dispatcher` to `FunctionType` to handle the above error correctly. 

(Fixes #9071)


dlee992:
Ah, look good to me!
If the caller has multiple functiontype sigs, before this patch, numba only tries the first sig, and throws an error if dismatch. After this patch, will try all sigs. I think so.

dlee992:
I am not familiar with `Dispatcher` code, and I didn't see side effects with this patch. Hope merge this ASAP!

dlee992:
can this patch go into 0.58 milestone? @esc

dlee992 (approved):

gmarkall (approved):
This looks good to me - the code changes look safe and the test represents the issue well. I've merged `main` so that it can be ready to merge once CI has run.

PR #9074, opened 2023-07-13
Guilhermeleobas/trim zeros

guilhermeleobas:
This is a fork of https://github.com/numba/numba/pull/5543. I only merged Numba main branch and fixed git conflicts and added tests from NumPy.


numba/tests/test_np_functions.py
```diff
@@ -4965,6 +4969,128 @@ def test_cross2d_exceptions(self):
...
+            expected = pyfunc(arr)
+            got = cfunc(arr)
+            self.assertPreciseEqual(expected, got)
+
+        for arr, trim in explicit_trim():
+            expected = pyfunc(arr, trim)
+            got = cfunc(arr, trim)
+            self.assertPreciseEqual(expected, got)
+
+    def test_trim_zeros_NumPy(self):
```
gmarkall, line 5005: It's a tiny bit odd to switch into camel case for half of the test name:

```suggestion
    def test_trim_zeros_numpy(self):
```

docs/upcoming_changes/9074.new_feature.rst
```diff
@@ -0,0 +1,5 @@
+
+Add np.trim_zeros
+=================
+
+Pull request 9074 adds ``np.trim_zeros()`` support to Numba.
```
gmarkall, line 5: I think there was a recent discussion on the format of the towncrier snippets where the conclusion was along the lines that we should keep it brief by avoiding self-references - e.g.:

```suggestion
Support for ``np.trim_zeros()`` is added.
```

gmarkall (changes requested):
I think the code and tests are generally fine, except for accepting sequences and tuples then producing an array - the output type is preserved by NumPy, so we're deviating from its semantics here. We can demonstrate this with:

```python
from numba import njit
import numpy as np

@njit
def f():
    return np.trim_zeros([0, 1, 2, 3]).ndim

np.testing.assert_equal(f(), f.py_func())
```

which raises:

```
  File "/home/gmarkall/numbadev/issues/9074/repro.py", line 11, in <module>
    np.testing.assert_equal(f(), f.py_func())
  File "/home/gmarkall/numbadev/issues/9074/repro.py", line 7, in f
    return np.trim_zeros([0, 1, 2, 3]).ndim
AttributeError: 'list' object has no attribute 'ndim'
```

when calling the non-jitted version. For the purpose of this PR, I think we should only accept an array (especially since we only test with arrays). A workaround for users that want to use tuples or other sequences will be to manually convert to / from arrays, but I don't expect this to be a common use case.

numba/np/arraymath.py
```diff
@@ -4722,3 +4722,36 @@ def impl(a, b):
         return _cross2d_operation(a_, b_)
 
     return impl
+
+
+@overload(np.trim_zeros)
+def np_trim_zeros(a, trim='fb'):
+    if not isinstance(a, types.Array):
+        raise NumbaTypeError('The first argument must be an array-like')
```
gmarkall, line 4730: I think it should just be an array now, not array-like?

```suggestion
        raise NumbaTypeError('The first argument must be an array')
```

docs/source/reference/numpysupported.rst
```diff
@@ -538,6 +538,7 @@ The following top-level functions are supported:
 * :func:`numpy.tril` (second argument ``k`` must be an integer)
 * :func:`numpy.tril_indices` (all arguments must be integer)
 * :func:`numpy.tril_indices_from` (second argument ``k`` must be an integer)
+* :func:`numpy.trim_zeros` (ouput is always an array)
```
gmarkall, line 541:
```suggestion
* :func:`numpy.trim_zeros` (for NumPy array arguments only)
```

numba/np/arraymath.py
```diff
@@ -4722,3 +4722,36 @@ def impl(a, b):
...
+
+    if hasattr(a, 'ndim') and a.ndim > 1:
+        raise NumbaTypeError('array must be 1D')
```
gmarkall, line 4733: `a` should always have an `ndim` attr, if it's always an array:

```suggestion
    if a.ndim > 1:
        raise NumbaTypeError('array must be 1D')
```

numba/tests/test_np_functions.py
```diff
@@ -4965,6 +4969,128 @@ def test_cross2d_exceptions(self):
...
+            cfunc(np.array([[1, 2, 3], [4, 5, 6]]))
+        self.assertIn(
+            'array must be 1D',
+            str(raises.exception)
+        )
+
+        with self.assertRaises(TypingError) as raises:
+            cfunc(3)
+        self.assertIn(
+            'The first argument must be an array-like',
```
gmarkall, line 5076: (Noting that the test cases will need updating to not check for "array-like" but "array" in the message)

gmarkall (changes requested):
Many thanks for the updates - I think there's a little more clarity to add about only arrays being supported, with relevant sections noted on the diff.

guilhermeleobas:
Thanks, @gmarkall. I believe I have addressed all your comments.

gmarkall (approved):
Thanks for the fixes and catching the difference in the argument names - this now looks good!

guilhermeleobas:
Thanks for the reviews @gmarkall

PR #9077, opened 2023-07-15
Enable inlining of first-class function when it is statically known to be a dispatcher

sklam:
- change lower_cast of Dispatcher to FunctionType to support inlining.
- FunctionType now has a `.jit_addr` to store the numba-internal function address that follows the numba callconv.
   - as a side-effect, FunctionType that has non-NULL `.jit_addr` can raise exception.

TODO:
- ~[ ] https://github.com/numba/numba/issues/9206#issuecomment-1733756843 Avoid the cfunc wrapper would be good.~
    - I decided to not pursue this change in this PR. However, this PR has the foundation to help make the change. Mainly, it adds the use of the internal function symbol (the `.jit_addr` in the FunctionType struct).


numba/core/cpu.py
```diff
@@ -205,7 +205,7 @@ def create_cfunc_wrapper(self, library, fndesc, env, call_helper):
 
         status, out = self.call_conv.call_function(
             builder, wrapper_callee, fndesc.restype, fndesc.argtypes,
-            wrapfn.args, attrs=('noinline',))
+            wrapfn.args)
```
sklam, line 208: After discussion with @stuartarchibald, this will need to be undone or it will leads to many optimization issues.

The alternative is to make first-class function use the Numba function without the wrapper when possible.

dlee992: Yeah, I think so. making it always inline could lead to many issues. One potential drawback could increase the compilation time, if the body of `FunctionType` is huge. Or maybe it's a bug in inline-related passes...

numba/experimental/function_type.py
```diff
@@ -249,17 +285,48 @@ def lower_cast_function_type_to_function_type(
...
+    else:
+        # Use lower_get_wrapper_address() to get the cfunc
+        lower_get_wrapper_address
+        pyapi = context.get_python_api(builder)
+
+        gil_state = pyapi.gil_ensure()
+        addr = lower_get_wrapper_address(
+            context, builder, val, toty.signature,
+            failure_mode='return_exc')
+        sfunc.c_addr = pyapi.long_as_voidptr(addr)
```
stuartarchibald, line 328: Should the `sfunc.jit_addr` explicitly be set to `NULL` on this path?

stuartarchibald (commented):
Thanks for the patch @sklam. Great to see this improvement being made. I've tested this with PIXIE and it works amazingly well! I also tested this manually to see the impact on caching, it seems like this doesn't impact it _but_ I now wonder if a function sufficiently large to prevent inlining might cause a problem? What do you think?

I've left a few minor comments inline, once addressed should be good to merge. 

Thanks again!

numba/core/callconv.py
```diff
@@ -876,9 +876,11 @@ def call_function(self, builder, callee, resty, argtys, args,
...
                Equivalent iterable: ("noinline", "fast")
         """
-        # XXX better fix for callees that are not function values
-        #     (pointers to function; thus have no `.args` attribute)
-        retty = self._get_return_argument(callee.function_type).pointee
+        retty = self.get_return_type(resty).pointee
+        actual_retty = self._get_return_argument(callee.function_type).pointee
+        if retty != actual_retty:
+            m = f"Function type returns {actual_retty} but resty={retty}"
+            raise ValueError(m)
```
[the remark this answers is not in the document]
sklam: Yeah, it's an internal assumption for defensive programming. It cannot be triggered if everything is correct.

numba/experimental/function_type.py
sklam: It is zero-filled by default due to `create_struct_proxy()` which did it at: https://github.com/numba/numba/blob/ed19445d09694bdb82ee00e112c027a1685ee546/numba/core/cgutils.py#L130

numba/experimental/function_type.py
```diff
@@ -166,9 +172,24 @@ def _get_wrapper_address(func, sig):
...
+    """
+    if isinstance(func, Dispatcher):
+        cres = func.get_compile_result(sig)
+        jit_name = cres.fndesc.llvm_func_name
+        addr = cres.library.get_pointer_to_function(jit_name)
+    else:
+        addr = 0
+    if not isinstance(addr, int):
+        raise TypeError(
+            f'jit address must be integer, got {type(addr)} instance')
```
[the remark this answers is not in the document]
sklam: yes, defensive programming. can't be triggered if program is correct. it makes it easier to debug.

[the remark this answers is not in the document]
stuartarchibald: Thanks for confirming.

numba/experimental/function_type.py
stuartarchibald: ah yes, thanks, that makes sense.

numba/core/callconv.py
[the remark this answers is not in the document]
stuartarchibald: Thanks for confirming.

sklam:
> I also tested this manually to see the impact on caching, it seems like this doesn't impact it _but_ I now wonder if a function sufficiently large to prevent inlining might cause a problem? What do you think?

Caching should not affected because this patch allows first-class functions to be treated like regular jit functions. They are referenced by name with their implementation linked into the same LLVM module. Since caching operates on the LLVM module level, it works no matter inilning occurred.

stuartarchibald:
> > I also tested this manually to see the impact on caching, it seems like this doesn't impact it _but_ I now wonder if a function sufficiently large to prevent inlining might cause a problem? What do you think?
> 
> Caching should not affected because this patch allows first-class functions to be treated like regular jit functions. They are referenced by name with their implementation linked into the same LLVM module. Since caching operates on the LLVM module level, it works no matter inilning occurred.

Thanks for explaining this, it's v. apparent in hindsight. The first-class function is linked in as part of the executable, it's invariant of inlining, and caching will always work.

PR #9086, opened 2023-07-24
Support NVRTC using ctypes binding

gmarkall:
This adds support for NVRTC using the ctypes binding, which enables linking CUDA C / C++ sources, and`float16`, when either binding is in use.

The binding is modelled on the implementation of the NVVM binding, for consistency in Numba. The main API of this binding is a single function, `compile()`, which compiles a CUDA C / C++ source to PTX. Internal APIs provide a Pythonic interface to the underlying NVRTC C APIs accessed though ctypes via Numba's `open_cudalib()` function.

Since there is no CUDA Python binding for NVVM, I opted to always use the ctypes binding to NVRTC (like NVVM) rather than using one or the other and maintaining two kinds of bindings inconsistently with the NVVM binding.

This supersedes #8893, but is substantially different from it - there is no vendoring of pynvrtc, which I considered problematic because it vendored an existing library with extensive changes without necessarily covering the changes with tests. It also didn't appear to protect from race conditions in initialization, like the existing NVVM binding does. The test changes in that PR looked OK to me though, so I have incorporated them here under Michael Collison's authorship.

Note that this PR is failing with CUDA 11.0 because it binds APIs that didn't exist in that version - however, following the merge of #9040, that configuration will no longer be supported or tested, so that should not be an issue.

### DIFF
--- a/numba/cuda/api.py
+++ b/numba/cuda/api.py
@@ -88,9 +88,9 @@ def is_cuda_array(obj):
 def is_float16_supported():
     """Whether 16-bit floats are supported.
 
-    Returns True if float16 is supported, False otherwise.
+    float16 is always supported in current versions of Numba - returns True.
     """
-    return bool(config.CUDA_USE_NVIDIA_BINDING)
+    return True
 
 
 @require_context
--- a/numba/cuda/cudadrv/nvrtc.py
+++ b/numba/cuda/cudadrv/nvrtc.py
@@ -1,10 +1,10 @@
 from ctypes import byref, c_char, c_char_p, c_int, c_size_t, c_void_p, POINTER
+from enum import IntEnum
 from numba.core import config
 from numba.cuda.cudadrv.error import (NvrtcError, NvrtcCompilationError,
                                       NvrtcSupportError)
 
 import functools
-import sys
 import threading
 import warnings
 
@@ -14,23 +14,21 @@ nvrtc_program = c_void_p
 # Result code
 nvrtc_result = c_int
 
-RESULT_CODE_NAMES = """
-NVRTC_SUCCESS
-NVRTC_ERROR_OUT_OF_MEMORY
-NVRTC_ERROR_PROGRAM_CREATION_FAILURE
-NVRTC_ERROR_INVALID_INPUT
-NVRTC_ERROR_INVALID_PROGRAM
-NVRTC_ERROR_INVALID_OPTION
-NVRTC_ERROR_COMPILATION
-NVRTC_ERROR_BUILTIN_OPERATION_FAILURE
-NVRTC_ERROR_NO_NAME_EXPRESSIONS_AFTER_COMPILATION
-NVRTC_ERROR_NO_LOWERED_NAMES_BEFORE_COMPILATION
-NVRTC_ERROR_NAME_EXPRESSION_NOT_VALID
-NVRTC_ERROR_INTERNAL_ERROR
-""".split()
 
-for i, k in enumerate(RESULT_CODE_NAMES):
-    setattr(sys.modules[__name__], k, i)
+class NvrtcResult(IntEnum):
+    NVRTC_SUCCESS = 0
+    NVRTC_ERROR_OUT_OF_MEMORY = 1
+    NVRTC_ERROR_PROGRAM_CREATION_FAILURE = 2
+    NVRTC_ERROR_INVALID_INPUT = 3
+    NVRTC_ERROR_INVALID_PROGRAM = 4
+    NVRTC_ERROR_INVALID_OPTION = 5
+    NVRTC_ERROR_COMPILATION = 6
+    NVRTC_ERROR_BUILTIN_OPERATION_FAILURE = 7
+    NVRTC_ERROR_NO_NAME_EXPRESSIONS_AFTER_COMPILATION = 8
+    NVRTC_ERROR_NO_LOWERED_NAMES_BEFORE_COMPILATION = 9
+    NVRTC_ERROR_NAME_EXPRESSION_NOT_VALID = 10
+    NVRTC_ERROR_INTERNAL_ERROR = 11
+
 
 _nvrtc_lock = threading.Lock()
 
@@ -121,10 +119,13 @@ class NVRTC:
                     @functools.wraps(func)
                     def checked_call(*args, func=func, name=name):
                         error = func(*args)
-                        if error == NVRTC_ERROR_COMPILATION: # noqa: F821
+                        if error == NvrtcResult.NVRTC_ERROR_COMPILATION:
                             raise NvrtcCompilationError()
-                        elif error != NVRTC_SUCCESS: # noqa: F821
-                            error_name = RESULT_CODE_NAMES[error]
+                        elif error != NvrtcResult.NVRTC_SUCCESS:
+                            try:
+                                error_name = NvrtcResult(error).name
+                            except ValueError:
+                                error_name = f'Unknown nvrtc_result {error}'
                             msg = f'Failed to call {name}: {error_name}'
                             raise NvrtcError(msg)
 
--- a/numba/cuda/cudadrv/nvrtc.py
+++ b/numba/cuda/cudadrv/nvrtc.py
@@ -39,8 +39,8 @@ class NvrtcProgram:
     the class own an nvrtcProgram; when an instance is deleted, the underlying
     nvrtcProgram is destroyed using the appropriate NVRTC API.
     """
-    def __init__(self, driver, handle):
-        self._driver = driver
+    def __init__(self, nvrtc, handle):
+        self._nvrtc = nvrtc
         self._handle = handle
 
     @property
@@ -49,7 +49,7 @@ class NvrtcProgram:
 
     def __del__(self):
         if self._handle:
-            self._driver.destroy_program(self)
+            self._nvrtc.destroy_program(self)
 
 
 class NVRTC:
@@ -105,14 +105,14 @@ class NVRTC:
                 from numba.cuda.cudadrv.libs import open_cudalib
                 cls.__INSTANCE = inst = object.__new__(cls)
                 try:
-                    inst.driver = open_cudalib('nvrtc')
+                    lib = open_cudalib('nvrtc')
                 except OSError as e:
                     cls.__INSTANCE = None
                     raise NvrtcSupportError("NVRTC cannot be loaded") from e
 
                 # Find & populate functions
                 for name, proto in inst._PROTOTYPES.items():
-                    func = getattr(inst.driver, name)
+                    func = getattr(lib, name)
                     func.restype = proto[0]
                     func.argtypes = proto[1:]
 
@@ -216,8 +216,8 @@ def compile(src, name, cc):
     :return: The compiled PTX and compilation log
     :rtype: tuple
     """
-    driver = NVRTC()
-    program = driver.create_program(src, name)
+    nvrtc = NVRTC()
+    program = nvrtc.create_program(src, name)
 
     # Compilation options:
     # - Compile for the current device's compute capability.
@@ -230,10 +230,10 @@ def compile(src, name, cc):
     options = [arch, include, '-rdc', 'true']
 
     # Compile the program
-    compile_error = driver.compile_program(program, options)
+    compile_error = nvrtc.compile_program(program, options)
 
     # Get log from compilation
-    log = driver.get_compile_log(program)
+    log = nvrtc.get_compile_log(program)
 
     # If the compile failed, provide the log in an exception
     if compile_error:
@@ -245,5 +245,5 @@ def compile(src, name, cc):
         msg = (f"NVRTC log messages whilst compiling {name}:\n\n{log}")
         warnings.warn(msg)
 
-    ptx = driver.get_ptx(program)
+    ptx = nvrtc.get_ptx(program)
     return ptx, log
--- a/numba/cuda/cudadrv/nvrtc.py
+++ b/numba/cuda/cudadrv/nvrtc.py
@@ -1,0 +1,248 @@
+from ctypes import byref, c_char, c_char_p, c_int, c_size_t, c_void_p, POINTER
+from numba.core import config
+from numba.cuda.cudadrv.error import (NvrtcError, NvrtcCompilationError,
+                                      NvrtcSupportError)
+
+import functools
+import sys
+import threading
+import warnings
+
+# Opaque handle for compilation unit
+nvrtc_program = c_void_p
+
+# Result code
+nvrtc_result = c_int
+
+RESULT_CODE_NAMES = """
+NVRTC_SUCCESS
+NVRTC_ERROR_OUT_OF_MEMORY
+NVRTC_ERROR_PROGRAM_CREATION_FAILURE
+NVRTC_ERROR_INVALID_INPUT
+NVRTC_ERROR_INVALID_PROGRAM
+NVRTC_ERROR_INVALID_OPTION
+NVRTC_ERROR_COMPILATION
+NVRTC_ERROR_BUILTIN_OPERATION_FAILURE
+NVRTC_ERROR_NO_NAME_EXPRESSIONS_AFTER_COMPILATION
+NVRTC_ERROR_NO_LOWERED_NAMES_BEFORE_COMPILATION
+NVRTC_ERROR_NAME_EXPRESSION_NOT_VALID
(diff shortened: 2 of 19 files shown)


gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

numba/cuda/cudadrv/nvrtc.py
```diff
@@ -0,0 +1,248 @@
...
+        'nvrtcGetProgramLog': (nvrtc_result, nvrtc_program, c_char_p),
+    }
+
+    # Singleton reference
+    __INSTANCE = None
+
+    def __new__(cls):
+        with _nvrtc_lock:
+            if cls.__INSTANCE is None:
+                from numba.cuda.cudadrv.libs import open_cudalib
```
stuartarchibald, line 107: Is this local to prevent circular import?

numba/cuda/cudadrv/nvrtc.py
```diff
@@ -0,0 +1,248 @@
...
+                cls.__INSTANCE = inst = object.__new__(cls)
+                try:
+                    inst.driver = open_cudalib('nvrtc')
```
stuartarchibald, line 110: Is this a driver? Maybe it could be considered so?

numba/cuda/cudadrv/nvrtc.py
```diff
@@ -0,0 +1,248 @@
...
+                    func.restype = proto[0]
+                    func.argtypes = proto[1:]
+
+                    @functools.wraps(func)
+                    def checked_call(*args, func=func, name=name):
+                        error = func(*args)
+                        if error == NVRTC_ERROR_COMPILATION: # noqa: F821
+                            raise NvrtcCompilationError()
+                        elif error != NVRTC_SUCCESS: # noqa: F821
+                            error_name = RESULT_CODE_NAMES[error]
```
stuartarchibald, line 127: Is it worth guarding for an error code which is out of range?

numba/cuda/cudadrv/nvrtc.py
```diff
@@ -0,0 +1,248 @@
...
+    def create_program(self, src, name):
+        """
+        Create an NVRTC program with managed lifetime.
+        """
+        handle = nvrtc_program()
+        # The final three arguments are for passing the contents of headers -
+        # this is not supported, so there are 0 headers and the header names
+        # and contents are null.
+        self.nvrtcCreateProgram(byref(handle), src, name.encode(), 0,
+                                c_char_p(0), c_char_p(0))
```
stuartarchibald, line 153:
```suggestion
                               None, None)
```
`ctypes` convention is to use `None` as NULL pointer?

numba/cuda/cudadrv/nvrtc.py
```diff
@@ -0,0 +1,248 @@
...
+
+    def create_program(self, src, name):
+        """
+        Create an NVRTC program with managed lifetime.
+        """
+        handle = nvrtc_program()
+        # The final three arguments are for passing the contents of headers -
+        # this is not supported, so there are 0 headers and the header names
+        # and contents are null.
+        self.nvrtcCreateProgram(byref(handle), src, name.encode(), 0,
```
stuartarchibald, line 152: Does the name need to be `ascii` encoded?

stuartarchibald (commented):
Thanks for the patch, it's great to see this functionality now working out-of-the-box. Given the testing and work to enable NVRTC was done previously, this change is essentially augmenting the code base with an alternative implementation and existing testing etc will cover this. There's a few minor things to look at in the review but otherwise looks good. Thanks again!

numba/cuda/cudadrv/nvrtc.py
```diff
@@ -0,0 +1,248 @@
...
+NVRTC_ERROR_COMPILATION
+NVRTC_ERROR_BUILTIN_OPERATION_FAILURE
+NVRTC_ERROR_NO_NAME_EXPRESSIONS_AFTER_COMPILATION
+NVRTC_ERROR_NO_LOWERED_NAMES_BEFORE_COMPILATION
+NVRTC_ERROR_NAME_EXPRESSION_NOT_VALID
+NVRTC_ERROR_INTERNAL_ERROR
+""".split()
+
+for i, k in enumerate(RESULT_CODE_NAMES):
+    setattr(sys.modules[__name__], k, i)
```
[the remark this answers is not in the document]
gmarkall: The `IntEnum` suggestion sounds good. I did this because it closely follows the NVVM binding, but this is not a great pattern!

numba/cuda/tests/cudapy/test_operator.py
```diff
@@ -144,9 +145,13 @@ def test_truediv(self):
     def test_floordiv(self):
         self.operator_template(operator.floordiv)
 
-    @skip_unless_cuda_python('NVIDIA Binding needed for NVRTC')
     @skip_unless_cc_53
+    @skip_on_cudasim('NVVM not supported in the simulator')
     def test_fp16_binary(self):
+        if not NVVM().is_nvvm70:
+            self.skipTest('Skip due to incorrect float16 divide codegen '
+                          'in earlier versions of nvvm.')
```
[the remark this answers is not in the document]
gmarkall: Will just remove this now, given that the PR already doesn't work on all CUDA configurations (because of 11.0) anyway.

numba/cuda/tests/cudapy/test_intrinsics.py
```diff
@@ -652,18 +653,22 @@ def test_hmul_ptx(self):
         self.assertIn('mul.f16', ptx)
 
     @skip_unless_cc_53
-    @skip_unless_cuda_python('NVIDIA Binding needed for NVRTC')
+    @skip_on_cudasim('NVVM not supported in the simulator')
     def test_hdiv_scalar(self):
+        if not NVVM().is_nvvm70:
+            self.skipTest('Skip due to incorrect float16 divide codegen '
+                          'in earlier versions of nvvm.')
```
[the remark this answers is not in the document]
gmarkall: Likewise :-)

numba/cuda/cudadrv/nvrtc.py
gmarkall: Yeah, because it gets used by `driver.py` - NVVM doesn't have it locally because the driver interface doesn't use it.

numba/cuda/cudadrv/nvrtc.py
gmarkall: `driver` is a weird name that I used for consistency with the NVVM binding, but a different name would probably be better. Are you suggesting you'd prefer a "different but better" name than a consistent name? (happy either way)

numba/cuda/cudadrv/nvrtc.py
gmarkall: Yeah, good point. Future NVRTCs could extend beyond the list we know of.

numba/cuda/cudadrv/nvrtc.py
gmarkall: OK, will check this works and apply if so.

numba/cuda/cudadrv/nvrtc.py
gmarkall: I believe I got a ctypes type error without doing that, but I'll double-check.

numba/cuda/cudadrv/nvrtc.py
stuartarchibald: I was considering "different but better reflecting what it is" name, a comment could capture that it's essentially the same as `driver` in the NVVM binding?

gmarkall: I'll add the comment, that's a good point. I think overall it'd be nice to have a consistent tidy-up of both NVRTC and NVVM at some later point.

numba/cuda/cudadrv/nvrtc.py
stuartarchibald: Apologies, wasn't clear, I think `.encode()` defaults to UTF-8, so was wondering if that might cause a problem?

gmarkall:
gpuci run tests

numba/cuda/api.py
```diff
@@ -88,9 +88,9 @@ def is_cuda_array(obj):
 def is_float16_supported():
     """Whether 16-bit floats are supported.
 
-    Returns True if float16 is supported, False otherwise.
+    float16 is always supported in current versions of Numba - returns True.
```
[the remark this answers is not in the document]
gmarkall: Can we leave this as-is? It seems simpler to think about for a user reading the docstring rather than thinking about what version of Numba is currently in use.

numba/cuda/cudadrv/nvrtc.py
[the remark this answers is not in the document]
gmarkall: I've changed this to an `IntEnum` now.

numba/cuda/tests/cudapy/test_operator.py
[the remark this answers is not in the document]
gmarkall: Removed.

numba/cuda/tests/cudapy/test_intrinsics.py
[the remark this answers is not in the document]
gmarkall: Removed.

numba/cuda/cudadrv/nvrtc.py
gmarkall: I decided to rename things rather than explain the odd name, because that makes things nicer overall IMHO.

numba/cuda/cudadrv/nvrtc.py
gmarkall: Now passing `None` for null pointers.

gmarkall:
@stuartarchibald Many thanks for the review - all comments responded to / addressed. (Just waiting on CI, will fix up if any issues).

gmarkall:
gpuci run tests

gmarkall:
The fail was:

```
System info:
  File "/home/vsts/miniconda3/envs/azure_ci/bin/numba", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/home/vsts/work/1/s/bin/numba", line 8, in <module>
    main()
  File "/home/vsts/work/1/s/numba/misc/numba_entry.py", line 39, in main
    display_sysinfo()
  File "/home/vsts/work/1/s/numba/misc/numba_sysinfo.py", line 543, in display_sysinfo
    info = get_sysinfo()
  File "/home/vsts/work/1/s/numba/misc/numba_sysinfo.py", line 504, in get_sysinfo
    jsond = json.loads(conda_out.decode())
  File "/home/vsts/miniconda3/envs/azure_ci/lib/python3.9/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
  File "/home/vsts/miniconda3/envs/azure_ci/lib/python3.9/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/home/vsts/miniconda3/envs/azure_ci/lib/python3.9/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 2 column 6 (char 6)
```

Something was an issue with the conda output when running `numba -s`. I think this is unrelated so I'll re-run checks (the re-run button doesn't work for me in the Azure pipelines UI so I need to trigger them all with the command).

numba/cuda/api.py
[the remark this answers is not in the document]
stuartarchibald: I'm ok with that, thanks for considering it.

numba/cuda/tests/cudapy/test_operator.py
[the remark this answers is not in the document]
stuartarchibald: Thanks. Does the `@skip_on_cudasim` also need removing? There's no NVVM support needed now?

numba/cuda/tests/cudapy/test_intrinsics.py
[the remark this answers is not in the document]
stuartarchibald: Thanks. Does the `@skip_on_cudasim` also need removing? There's no NVVM support needed now?

numba/cuda/cudadrv/nvrtc.py
stuartarchibald: Thanks, the rename looks good.

numba/cuda/cudadrv/nvrtc.py
stuartarchibald: Great, thanks for checking and making the update.

numba/cuda/cudadrv/nvrtc.py
```diff
@@ -0,0 +1,254 @@
...
+                    @functools.wraps(func)
+                    def checked_call(*args, func=func, name=name):
+                        error = func(*args)
+                        if error == NvrtcResult.NVRTC_ERROR_COMPILATION:
+                            raise NvrtcCompilationError()
+                        elif error != NvrtcResult.NVRTC_SUCCESS:
+                            try:
+                                error_name = NvrtcResult(error).name
+                            except ValueError:
+                                error_name = f'Unknown nvrtc_result {error}'
```
stuartarchibald, line 128:
```suggestion
                                error_name = f'Unknown nvrtc_result (error code: {error}).'
```
Am wondering if the suggested change makes the error result/code more clear. It currently reads, e.g.:
```
numba.cuda.cudadrv.error.NvrtcError: Failed to call nvrtcCreateProgram: Unknown nvrtc_result 1000
```
with the proposed change it will read:
```
numba.cuda.cudadrv.error.NvrtcError: Failed to call nvrtcCreateProgram: Unknown nvrtc_result (error code: 1000).
```

numba/cuda/tests/cudapy/test_intrinsics.py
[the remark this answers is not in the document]
gmarkall: Removed.

numba/cuda/tests/cudapy/test_operator.py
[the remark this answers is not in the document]
gmarkall: Removed.

numba/cuda/cudadrv/nvrtc.py
gmarkall: Applied a similar change with line wrap.

gmarkall:
gpuci run tests

stuartarchibald:
Buildfarm ID: `numba_smoketest_cuda_yaml_211`.

stuartarchibald:
> Buildfarm ID: `numba_smoketest_cuda_yaml_211`.

This is failing for tests such as `numba.cuda.tests.cudapy.test_operator.TestOperatorModule.test_mixed_fp16_binary_arithmetic` with an approximate trace:
```
File "numba/cuda/cudadrv/nvrtc.py", line 247, in compile
   raise NvrtcError(msg)
... <skip>
error: cannot open source file "cuda_fp16.h" 1 catastrophic error detected in the compilation of "cpp_function_wrappers.cu".
Compilation terminated.
```
failure seems invariant of CUDA version. Perhaps there needs to be a check for headers? I assume that's where the problem lies?

gmarkall:
Thanks - I think this has always been a problem and we've never noticed because we've never had a configuration on the buildfarm that runs this check. If you don't have a full toolkit installed, this will happen, but I never noticed before. I will check if there is a conda package that includes these that we should require, but we should also guard against the header not being present somehow.

gmarkall:
gpuci run tests

gmarkall:
@esc Could this have another buildfarm run please?

gmarkall:
gpuci run tests

gmarkall:
Following @stuartarchibald 's approval, the changes to pass on the build farm are:

* The inclusion of the headers `cuda_fp16.{h,hpp}` and their installation at install time. Commit https://github.com/numba/numba/pull/9086/commits/ad931f81e17dd680516177fa6b22fde17f24c52e
* Addition of the license notes for these headers. Commit https://github.com/numba/numba/pull/9086/commits/937ca4060af6dd528a7af84507a189cd185213b3
* Addition of headers to the NVRTC include path. Commit https://github.com/numba/numba/pull/9086/commits/dc112f41a3409c3c209b7a151c53b0502235c14b

Could @esc or @sklam review these additional changes please?

esc (approved):
Vendoring of headers and their inclusion seems to work. Following @stuartarchibald "s approval, I add mine too!

PR #9087, opened 2023-07-25
Add trimseq from np.polynomial.polyutils and polyadd, polysub, polymul from np.polynomial.polynomial

KrisMinchev:
This PR improves on #4074 by implementing the functions `polyadd`, `polysub`, `polymul`. Note that since `np.poly*` functions are part of the old API as mentioned [here](https://numpy.org/doc/stable/reference/routines.polynomials.html), functions from the new `numpy.polynomial` package are implemented instead. The function `np.polynomial.polyutils.trimseq` is necessary for the implementations of the poly* functions, so it is included in the PR as well.

### DIFF
--- a/numba/np/polynomial.py
+++ b/numba/np/polynomial.py
@@ -98,11 +98,17 @@ def numpy_polyadd(c1, c2):
 
     if isinstance(c1, types.Number):
         s1 = str(as_dtype(c1))
+    elif isinstance(c1, types.Tuple):
+        t = [as_dtype(ty) for ty in c1.types]
+        s1 = str(np.result_type(*t))
     else:
         s1 = str(c1.dtype)
 
     if isinstance(c2, types.Number):
         s2 = str(as_dtype(c2))
+    elif isinstance(c2, types.Tuple):
+        t = [as_dtype(ty) for ty in c2.types]
+        s2 = str(np.result_type(*t))
     else:
         s2 = str(c2.dtype)
 
@@ -142,11 +148,17 @@ def numpy_polysub(c1, c2):
 
     if isinstance(c1, types.Number):
         s1 = str(as_dtype(c1))
+    elif isinstance(c1, types.Tuple):
+        t = [as_dtype(ty) for ty in c1.types]
+        s1 = str(np.result_type(*t))
     else:
         s1 = str(c1.dtype)
 
     if isinstance(c2, types.Number):
         s2 = str(as_dtype(c2))
+    elif isinstance(c2, types.Tuple):
+        t = [as_dtype(ty) for ty in c2.types]
+        s2 = str(np.result_type(*t))
     else:
         s2 = str(c2.dtype)
 
@@ -186,11 +198,17 @@ def numpy_polymul(c1, c2):
 
     if isinstance(c1, types.Number):
         s1 = str(as_dtype(c1))
+    elif isinstance(c1, types.Tuple):
+        t = [as_dtype(ty) for ty in c1.types]
+        s1 = str(np.result_type(*t))
     else:
         s1 = str(c1.dtype)
 
     if isinstance(c2, types.Number):
         s2 = str(as_dtype(c2))
+    elif isinstance(c2, types.Tuple):
+        t = [as_dtype(ty) for ty in c2.types]
+        s2 = str(np.result_type(*t))
     else:
         s2 = str(c2.dtype)
 
--- a/numba/np/polynomial.py
+++ b/numba/np/polynomial.py
@@ -65,6 +65,10 @@ def polyutils_trimseq(seq):
     if not type_can_asarray(seq):
         msg = 'The argument "seq" must be array-like'
         raise errors.TypingError(msg)
+    
+    if isinstance(seq, types.BaseTuple):
+        msg = 'Unsupported type %r for argument "seq"'
+        raise errors.TypingError(msg % (seq))
 
     if np.ndim(seq) > 1:
         msg = 'Coefficient array is not 1-d'
--- a/numba/np/polynomial.py
+++ b/numba/np/polynomial.py
@@ -4,12 +4,12 @@ Implementation of operations involving polynomials.
 
 
 import numpy as np
+from numpy.polynomial import polynomial as poly
 
 from numba import jit
-from numba.core import types
-from numba.core.extending import overload
+from numba.core import types, errors
+from numba.core.extending import overload, register_jitable
 from numba.np import numpy_support as np_support
-
 
 @overload(np.roots)
 def roots_impl(p):
@@ -57,3 +57,85 @@ def roots_impl(p):
             return roots
 
     return roots_impl
+
+
+@register_jitable
+def trimseq(seq):
+    """Remove small Poly series coefficients.
+
+    Parameters
+    ----------
+    seq : sequence
+        Sequence of Poly series coefficients. This routine fails for
+        empty sequences.
+
+    Returns
+    -------
+    series : sequence
+        Subsequence with trailing zeros removed. If the resulting sequence
+        would be empty, return the first element. The returned sequence may
+        or may not be a view.
+
+    Notes
+    -----
+    Do not lose the type info if the sequence contains unknown objects.
+
+    """
+    if len(seq) == 0:
+        return seq
+    else:
+        for i in range(len(seq) - 1, -1, -1):
+            if seq[i] != 0:
+                break
+        return seq[:i+1]
+
+
+@overload(poly.polyadd)
+def numpy_polyadd(a1, a2):
+    if not isinstance(a1, types.Array):
+        msg = 'The argument "a1" must be an array'
+        raise errors.TypingError(msg)
+    
+    if not isinstance(a2, types.Array):
+        msg = 'The argument "a2" must be an array'
+        raise errors.TypingError(msg)
+    
+    def impl(a1, a2):
+        arr1 = np.atleast_1d(a1).astype(np.float64)
+        arr2 = np.atleast_1d(a2).astype(np.float64)
+        diff = len(a2) - len(a1)
+        if diff > 0:
+            zr = np.zeros(diff) #, a1.dtype)
+            arr1 = np.concatenate((a1, zr))
+        if diff < 0:
+            zr = np.zeros(-diff) #, a2.dtype)
+            arr2 = np.concatenate((a2, zr))
+        val = arr1 + arr2
+        return trimseq(val)
+    
+    return impl
+
+@overload(poly.polysub)
+def numpy_polysub(a1, a2):
+    if not isinstance(a1, types.Array):
+        msg = 'The argument "a1" must be an array'
+        raise errors.TypingError(msg)
+    
+    if not isinstance(a2, types.Array):
+        msg = 'The argument "a2" must be an array'
+        raise errors.TypingError(msg)
+    
+    def impl(a1, a2):
+        arr1 = np.atleast_1d(a1).astype(np.float64)
+        arr2 = np.atleast_1d(a2).astype(np.float64)
+        diff = len(a2) - len(a1)
+        if diff > 0:
+            zr = np.zeros(diff) #, a1.dtype)
+            arr1 = np.concatenate((a1, zr))
+        if diff < 0:
+            zr = np.zeros(-diff) #, a2.dtype)
+            arr2 = np.concatenate((a2, zr))
+        val = arr1 - arr2
+        return trimseq(val)
+    
+    return impl
--- a/numba/np/polynomial.py
+++ b/numba/np/polynomial.py
@@ -60,37 +60,6 @@ def roots_impl(p):
     return roots_impl
 
 
-@register_jitable
-def _trimseq(seq):
-    """Remove small Poly series coefficients.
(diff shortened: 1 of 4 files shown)


sklam:
(did the close-reopen dance to trigger CI)

guilhermeleobas:
Hi Kristian, the functions you're implementing are part of the old polynomial API?

numba/np/polynomial.py
```diff
@@ -57,3 +58,116 @@ def roots_impl(p):
...
+
+@overload(poly.polyadd)
+def numpy_polyadd(c1, c2):
+    if not isinstance(c1, types.Array):
+        msg = 'The argument "c1" must be an array'
+        raise errors.TypingError(msg)
+    
+    if not isinstance(c2, types.Array):
+        msg = 'The argument "c2" must be an array'
+        raise errors.TypingError(msg)
```
guilhermeleobas, line 114: `c1` and `c2` must be array-like objects. See:

```python
>>> import numpy as np
>>> np.polynomial
>>> np.polynomial.polynomial.polyadd(1, 2)
array([3.])
```

In Numba there's a function called `type_can_asarray` that you can use to check if the type is convertible to an ndarray.

numba/np/polynomial.py
```diff
@@ -57,3 +58,116 @@ def roots_impl(p):
...
+
+@overload(poly.polysub)
+def numpy_polysub(c1, c2):
+    if not isinstance(c1, (types.Array, types.Integer)):
+        msg = 'The argument "c1" must be an array'
+        raise errors.TypingError(msg)
+    
+    if not isinstance(c2, (types.Array, types.Integer)):
+        msg = 'The argument "c2" must be an array'
+        raise errors.TypingError(msg)
```
guilhermeleobas, line 140: Same here

numba/np/polynomial.py
```diff
@@ -57,3 +58,116 @@ def roots_impl(p):
...
+
+@overload(poly.polymul)
+def numpy_polymul(c1, c2):
+    if not isinstance(c1, types.Array):
+        msg = 'The argument "c1" must be an array'
+        raise errors.TypingError(msg)
+    
+    if not isinstance(c2, types.Array):
+        msg = 'The argument "c2" must be an array'
+        raise errors.TypingError(msg)
```
guilhermeleobas, line 166: Same here

numba/np/polynomial.py
```diff
@@ -57,3 +58,116 @@ def roots_impl(p):
...
+    
+    # p is an array, x is a scalar
+    def impl(c1, c2):
+        val = np.convolve(c1, c2)
```
guilhermeleobas, line 170: `c1` and `c2` needs to be 1d arrays?

```python
>>> a = np.arange(10).reshape(5, 2)
>>> b = np.arange(10).reshape(5, 2)
>>> np.convolve(a, b)
  File "<__array_function__ internals>", line 180, in convolve
  File "/Users/guilhermeleobas/micromamba/envs/numba/lib/python3.10/site-packages/numpy/core/numeric.py", line 850, in convolve
    return multiarray.correlate(a, v[::-1], mode)
ValueError: object too deep for desired array
>>> a = np.arange(5)
>>> b = np.arange(5)
>>> np.convolve(a, b)
array([ 0,  0,  1,  4, 10, 20, 25, 24, 16])
```

numba/np/polynomial.py
```diff
@@ -57,3 +58,116 @@ def roots_impl(p):
             return roots
 
     return roots_impl
+
+
+@register_jitable
+def _trimseq(seq):
```
guilhermeleobas, line 64: No need for using a `@register_jitable` here. You can put the code inside the `pu.trimsec` overload and just call it inside other overloads.

numba/np/polynomial.py
```diff
@@ -57,3 +58,116 @@ def roots_impl(p):
...
+            if seq[i] != 0:
+                break
+        return seq[:i+1]
+
+
+@overload(pu.trimseq)
+def polyutils_trimseq(seq):
+    if not isinstance(seq, types.Array):
+        msg = 'The argument "seq" must be an array'
+        raise errors.TypingError(msg)
```
guilhermeleobas, line 98: `pu.trimseq` requires a 1d array?

```python
>>> np.polynomial.polyutils.trimseq(np.arange(4).reshape(2, 2))
  File "/Users/guilhermeleobas/micromamba/envs/numba/lib/python3.10/site-packages/numpy/polynomial/polyutils.py", line 75, in trimseq
    if seq[i] != 0:
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

guilhermeleobas (changes requested):
Hi Kristian, thanks for working on this topic. It's good to see `np.polynomial` support in Numba.

I gave an initial review. Let me know if you have any questions.

numba/np/polynomial.py
```diff
@@ -57,3 +58,150 @@ def roots_impl(p):
...
+        s1 = str(as_dtype(typeof(c1)))
+    else:
+        s1 = str(c1.dtype)
+
+    if isinstance(c2, types.Integer):
+        s2 = str(as_dtype(typeof(c2)))
+    else:
+        s2 = str(c2.dtype)
+
+    result_dtype = from_dtype(np.result_type(s1, s2, np.float64))
```
guilhermeleobas, line 109: The last argument in `np.result_type` will force the promotion to float64:

```
>>> a = np.arange(5, dtype=np.int32)
>>> b = np.arange(5, dtype=np.int16)
>>> np.polyadd(a, b)
array([0, 2, 4, 6, 8], dtype=int32)
```

numba/np/polynomial.py
```diff
@@ -57,3 +58,150 @@ def roots_impl(p):
...
+    if not type_can_asarray(c2):
+        msg = 'The argument "c2" must be array-like'
+        raise errors.TypingError(msg)
+
+    if np.ndim(c1) > 1 or np.ndim(c2) > 1:
+        msg = 'Coefficient array is not 1-d'
+        raise errors.NumbaValueError(msg)
+
+    if isinstance(c1, types.Integer):
+        s1 = str(as_dtype(typeof(c1)))
```
guilhermeleobas, line 100: Do you need to call `typeof` here?

numba/np/polynomial.py
```diff
@@ -57,3 +58,150 @@ def roots_impl(p):
...
+    result_dtype = from_dtype(np.result_type(s1, s2, np.float64))
+
+    def impl(c1, c2):
+        c1 = np.asarray(c1)
+        c2 = np.asarray(c2)
+        arr1 = np.atleast_1d(c1).astype(result_dtype)
+        arr2 = np.atleast_1d(c2).astype(result_dtype)
+        diff = len(arr2) - len(arr1)
+        if diff > 0:
+            zr = np.zeros(diff) # .astype(result_dtype)
```
guilhermeleobas, line 118:
```suggestion
            zr = np.zeros(diff)
```

numba/np/polynomial.py
```diff
@@ -57,3 +58,150 @@ def roots_impl(p):
...
+            arr1 = np.concatenate((arr1, zr))
+        if diff < 0:
+            zr = np.zeros(-diff) # .astype(result_dtype)
```
guilhermeleobas, line 121:
```suggestion
            zr = np.zeros(-diff)
```

numba/tests/test_polynomial.py
```diff
@@ -115,3 +131,175 @@ def check(a, **kwargs):
...
+            yield 3, (1, 2, 3)
+            #test different dtypes
+            yield np.array([1, 2, 3]), np.array([1.0, 2.0, 3.0])
+            yield np.array([1j, 2j, 3j]), np.array([1.0, 2.0, 3.0])
+            yield np.array([1, 2, 3]), np.array([1j, 2j, 3j])
+
+        for p1, p2 in inputs():
+            self.assertPreciseEqual(pyfunc(p1,p2), cfunc(p1,p2))
+
+    def test_polyadd_exception(self):
```
guilhermeleobas, line 183: The exception cases are the same for polyadd/polysub/polymul. Can you refactor this code to reduce code duplication? One alternative would be to have a function `_test_poly_exception(self, pyfunc)`:

```python
def _test_polyfunc_exception(self, pyfunc):
    ....

def test_polyadd_exception(self):
    self._test_polyfunc_exception(self, polyadd)

...
```

numba/tests/test_polynomial.py
```diff
@@ -115,3 +131,175 @@ def check(a, **kwargs):
...
+        
+        with self.assertRaises(TypingError) as e:
+            cfunc(np.array([1, 2, 3]), np.arange(10).reshape(5, 2))
+        self.assertIn('Coefficient array is not 1-d',
+                      str(e.exception))
+
+    def test_polysub_basic(self):
+        pyfunc = polysub
+        cfunc = njit(polysub)
+        def inputs():
```
guilhermeleobas, line 211: It seems that the inputs for polyadd/sub/mul are the same. Can you refactor this code to reduce code duplication? Also, are these inputs from NumPy?

guilhermeleobas (changes requested):
Hi @KrisMinchev, I gave an initial review. 

Code looks good, only some edge cases that need work.

numba/np/polynomial.py
KrisMinchev: Hi @guilhermeleobas, I believe you are using the old API. The new ``np.polynomial`` package promotes to float64:
```
>>> import numpy as np
>>> import numpy.polynomial.polynomial as poly
>>> a = np.arange(5, dtype=np.int32)
>>> b = np.arange(5, dtype=np.int16)
>>> c = poly.polyadd(a, b)
>>> c
array([0., 2., 4., 6., 8.])
>>> c.dtype
dtype('float64')
```

numba/np/polynomial.py
KrisMinchev: You are right, it is not necessary to call `typeof`. I will remove it accordingly.

numba/tests/test_polynomial.py
KrisMinchev: I have combined the code in `_test_polyarithm_basic()`. The only inputs taken from NumPy are the ones marked with the `# basic` comment. I have added the permalink accordingly.

numba/tests/test_polynomial.py
```diff
@@ -156,32 +156,38 @@ def test_trimseq_exception(self):
...
-        pyfunc = polyadd
-        cfunc = njit(polyadd)
+    def _test_polyarithm_basic(self, pyfunc, ignore_sign_on_zero = False):
+        # test suite containing tests for polyadd, polysub, polymul, polydiv
+        cfunc = njit(pyfunc)
         def inputs():
-            #basic
+            # basic, taken from https://github.com/numpy/numpy/blob/...
+            # 48a8277855849be094a5979c48d9f5f1778ee4de/numpy/polynomial/...
+            # tests/test_polynomial.py#L58-L123
```
guilhermeleobas, line 165: Keep the URL in a single line

numba/np/polynomial.py
```diff
@@ -57,3 +58,150 @@ def roots_impl(p):
...
+    result_dtype = from_dtype(np.result_type(s1, s2, np.float64))
+
+    def impl(c1, c2):
+        c1 = np.asarray(c1)
+        c2 = np.asarray(c2)
+        arr1 = np.atleast_1d(c1).astype(result_dtype)
+        arr2 = np.atleast_1d(c2).astype(result_dtype)
+        diff = len(arr2) - len(arr1)
+        if diff > 0:
+            zr = np.zeros(diff) #, a1.dtype)
```
guilhermeleobas, line 162:
```suggestion
            zr = np.zeros(diff)
```

numba/np/polynomial.py
```diff
@@ -57,3 +58,150 @@ def roots_impl(p):
...
+            arr1 = np.concatenate((arr1, zr))
+        if diff < 0:
+            zr = np.zeros(-diff) #, a2.dtype)
```
guilhermeleobas, line 165:
```suggestion
            zr = np.zeros(-diff)
```

guilhermeleobas (changes requested):

KrisMinchev:
Hi @guilhermeleobas, thank you for the suggestions. It is important to note that the function `pu.trimseq` cannot support tuple inputs, since the output length depends on the values of the input:
```python
>>> import numpy.polynomial.polyutils as pu
>>> pu.trimseq((1, 0, 0))
(1,)
>>> pu.trimseq((1, 0, 1))
(1, 0, 1)
```
This function is used at the end of all arithmetic functions(`polyadd`, `polysub`, etc.), but it is only applied to NumPy arrays so there is no issue there. Is it reasonable to keep `pu.trimseq` with its limited support, or should I remove it altogether?

KrisMinchev:
Thank you @guilhermeleobas, I have updated the behaviour of `pu.trimseq()` accordingly.

guilhermeleobas (changes requested):

guilhermeleobas (approved):
Current CI failure is not related to any change introduced in this PR.

esc:
@KrisMinchev thank you for the patch!

PR #9089, opened 2023-07-25
Fix segfault on passing `None` for args in PythonAPI.call

### LINKED ISSUE #5871: Bug on pyapi.call when args is None

hellozee:
Fixes #5871

Hi, was snooping around the codebase, trying to learn something (mostly playing with llvmlite). Never used numba (directly at least). So, hopefully this is what was expected, from what I understood after reading the discussion on the stale PR.


guilhermeleobas:
Hi @hellozee, CI is currently failing on your PR. Could you take a look?

numba/core/pythonapi.py
```diff
@@ -945,12 +945,14 @@ def call_method(self, callee, method, objargs=()):
...
         if args is None:
-            args = self.get_null_object()
+            args = self.tuple_new(0)
         if kws is None:
             kws = self.get_null_object()
         fnty = ir.FunctionType(self.pyobj, [self.pyobj] * 3)
         fn = self._get_function(fnty, name="PyObject_Call")
-        return self.builder.call(fn, (callee, args, kws))
+        result = self.builder.call(fn, (callee, args, kws))
+        self.decref(args)
```
gmarkall, line 954: (Reposting since my previous comment seemed to get lost)

I'm guessing here, but perhaps we should only do this decref if we had to set `args` to a new empty tuple?

numba/core/pythonapi.py
```diff
@@ -944,13 +944,17 @@ def call_method(self, callee, method, objargs=()):
         return self.builder.call(fn, args)
 
     def call(self, callee, args=None, kws=None):
+        args_was_none = args is None
         if args is None:
```
gmarkall, line 948: Since we only support Python 3.8+, I think this might be an appropriate place for a walrus to avoid repetition of the expression:

```suggestion
        if args_was_none := args is None:
```

numba/tests/test_pythonapi_emptyargs.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+@njit
+def foo(n: int):
+    d = Dense(N=n)
+    return d
+
+
+class PythonAPIEmptyArgs(unittest.TestCase):
+    def test_empty_args(self):
```
sklam, line 74: I think this test should go in the `numba/tests/test_pythonapi.py` file.

Also, the test is too involved for what it is testing. I'm suggesting the following as an alternative:


```python
from numba import njit
from numba.core.types import none, intp
from numba.core.extending import intrinsic
import unittest
from numba.tests.support import captured_stdout


class PythonAPIEmptyArgs(unittest.TestCase):
    def test_empty_args(self):
        def callme(**kwargs):
            print("callme", kwargs)

        @intrinsic
        def py_call(tyctx):
                pyapi = context.get_python_api(builder)
                gil = pyapi.gil_ensure()

                num = pyapi.long_from_long(context.get_constant(intp, 0xCAFE))
                kwds = pyapi.dict_pack({"key": num}.items())
                fn_print = pyapi.unserialize(pyapi.serialize_object(callme))
                # segfault: https://github.com/numba/numba/issues/5871
                res = pyapi.call(fn_print, None, kwds)

                pyapi.decref(res)
                pyapi.decref(fn_print)
                pyapi.decref(kwds)
                pyapi.decref(num)

                pyapi.gil_release(gil)
                return res

            return none(), codegen

        @njit
        def foo():
            py_call()

        with captured_stdout() as out:
            foo()
        d = {"key": 0xCAFE}
        expected = f"callme {d}\n"
        self.assertEqual(out.getvalue(), expected)


```

sklam (changes requested):
The changes in `call` looks good. My comments below are only for the tests, which I think should go in an existing file and can be shortened.

numba/tests/test_pythonapi.py
```diff
@@ -78,5 +79,45 @@ def test_PyBytes_AsStringAndSize(self):
...
+        def callme(**kwargs):
+            print("callme", kwargs)
+
+        @intrinsic
+        def py_call(tyctx):
+            def codegen(context, builder, sig, args):
+                pyapi = context.get_python_api(builder)
+                gil = pyapi.gil_ensure()
+
+                num = pyapi.long_from_long(
```
gmarkall, line 93: I'm guessing a bit, but maybe this should have been

```suggestion
                num = pyapi.long_from_longlong(
```

because I think `long` might be an `i32` on Windows, causing the current test failure.

hellozee: ahh yea, right, I did fix it before but here I forgot

gmarkall (approved):
Many thanks for your contribution and persisting through all the iterations! I think this now looks good.

gmarkall:
Oops, I forgot the changelog entry is missing - one moment...

gmarkall (dismissed):
Could you please add a changelog entry, `docs/upcoming_changes/9089.fix_pyapi_args_none.rst`? It should have a format similar to that of e.g. https://github.com/numba/numba/blob/e010b4cec45f45711aaea89d390a1af2122bc28a/docs/upcoming_changes/7262.bug_fix.rst?plain=1

gmarkall:
@hellozee Thanks for adding the changelog entry. I should have asked before, but failed to notice - could you please avoid doing anything that requires a force-push (i.e. a rebase or editing commits) as Github makes it extremely hard to tell what's changed between reviews when this is done - many thanks in advance!

@kc611 Do you know why towncrier failed here? It seems to think there are no new newsfragments but I'm not sure I understand why.

kc611:
I think for the fragments to be detected, they need to be named in a specific way. Towncrier checks if the second part of the name specified (after the PR number) matches any of the described types. In this case it does not.

gmarkall:
@kc611 Ah, I see - many thanks!

@hellozee I think the file should be named `docs/upcoming_changes/9089.bug_fix.pyapi_args_none.rst` instead.

hellozee:
> could you please avoid doing anything that requires a force-push (i.e. a rebase or editing commits) as Github makes it extremely hard to tell what's changed between reviews when this is done

My bad

sklam (approved):
LGTM

Issue #8564, opened 2022-11-01
Make Numba's `captured_errors="new_style"` by default.


stuartarchibald:
Following on from discussion at today's public Numba meeting (01 November, 2022). @guilhermeleobas suggested that it would be good to make Numba's "error capturing" use the "new style" by default, there was general agreement that this would be a good idea. The "new style" error capturing was implemented in Numba 0.55 via #7397 and has been used in Numba's CI systems ever since. The implications of the `captured_errors` style are discussed in the text accompanying #7397. Steps to make this change:

1. Issue deprecation warnings and create a deprecation cycle in relation to the use of the `old_style` error capturing behaviour.
2. Schedule removal and remove the `old_style` error capturing behaviour once the deprecation period ends.

esc:
Closing as #9090 has been merged

Seairth:
See my comment at https://github.com/numba/numba/issues/9091#issuecomment-1693344159.  It appears that deprecation notices have been added, but the default in config.py has _not_ been updated to be `new_style`.

PR #9090, opened 2023-07-25
Add deprecation notice for new_style error capturing.

sklam:
Addressing #8564.

### DIFF
--- a/docs/source/reference/deprecation.rst
+++ b/docs/source/reference/deprecation.rst
@@ -495,3 +495,44 @@ Schedule
 - In Numba 0.57: Support for CUDA toolkit 10.2 was removed.
 - In Numba 0.58: Support for CC < 5.0 and CUDA toolkits 11.0 and 11.1 will be
   removed.
+
+Deprecation of old-style ``NUMBA_CAPTURED_ERRORS``
+==================================================
+
+The old style of the ``NUMBA_CAPTURED_ERRORS`` environment variable is being 
+deprecated in Numba.
+
+Previously, this variable allowed controlling how Numba handles exceptions 
+during compilation that do not inherit from ``numba.core.errors.NumbaError``. 
+The default "old_style" behavior was to capture and wrap these errors, often 
+obscuring the original exception.
+
+The new "new_style" option treats non-``NumbaError`` exceptions as hard errors, 
+propagating them without capturing. This differentiates compilation errors from 
+unintended exceptions during compilation.
+
+The old style will eventually be removed in favor of the new behavior. Users 
+should migrate to setting ``NUMBA_CAPTURED_ERRORS='new_style'`` to opt-in to the 
+new exception handling. This will become the default in the future.
+
+Recommendations
+---------------
+
+- Projects that depends on Numba should set 
+  ``NUMBA_CAPTURED_ERRORS='new_style'`` for testing to find all places where 
+  non-``NumbaError`` exceptions are raised during compilation.
+- Replace any code that raises non-``NumbaError`` to indicate compilation error.
+  For instance, instead of raising ``TypeError``, raise 
+  ``numba.core.errors.NumbaTypeError``.
+
+
+Schedule
+--------
+
+- In Numba 0.58: support for ``NUMBA_CAPTURED_ERRORS='old_style'`` is 
+  deprecated.
+- In Numba 0.60: support for ``NUMBA_CAPTURED_ERRORS='new_style'`` becomes the
+  default.
+- In Numba 0.61: support for ``NUMBA_CAPTURED_ERRORS='old_style'`` will be 
+  removed.
+


sklam:
i’m still fine-tuning details for the deprecating old_style error. How “loud”/“annoying” should it be for 0.58?

For anyone NOT extending Numba, they probably never set NUMBA_CAPTURED_ERRORS thus I don’t think we should annoy them.  But, that also means it is hard to reach extension writers who are probably unknowingly relying on old_style.

Maybe we can make the old-style capturing logic warn at this release? (I have not-yet-pushed code for this)

gmarkall:
I think making it warn in this release (and calling it deprecated) is worth doing - since the documentation is unlikely to be noticed otherwise, only adding the documentation and not warnings in this release is almost effectively a "no-op" if many will not notice until 0.59.

docs/source/reference/deprecation.rst
```diff
@@ -495,3 +495,46 @@ Schedule
...
+unintended exceptions during compilation.
+
+The old style will eventually be removed in favor of the new behavior. Users 
+should migrate to setting ``NUMBA_CAPTURED_ERRORS='new_style'`` to opt-in to the 
+new exception handling. This will become the default in the future.
+
+Recommendations
+---------------
+
+- Projects that depends on Numba should set
```
gmarkall, line 521:
```suggestion
- Projects that depend on Numba should set 
```

docs/source/reference/deprecation.rst
```diff
@@ -495,3 +495,46 @@ Schedule
...
+
+Recommendations
+---------------
+
+- Projects that depends on Numba should set 
+  ``NUMBA_CAPTURED_ERRORS='new_style'`` for testing to find all places where 
+  non-``NumbaError`` exceptions are raised during compilation.
+- Replace any code that raises non-``NumbaError`` to indicate compilation error.
+  For instance, instead of raising ``TypeError``, raise 
+  ``numba.core.errors.NumbaTypeError``.
```
gmarkall, line 526:
```suggestion
- Modify any code that raises a non-``NumbaError`` to indicate a compilation
  error to raise a subclass of ``NumbaError`` instead. For example, instead of
  raising a ``TypeError``, raise a ``numba.core.errors.NumbaTypeError``.
```

docs/source/reference/deprecation.rst
```diff
@@ -495,3 +495,46 @@ Schedule
...
+--------
+
+- In Numba 0.58: support for ``NUMBA_CAPTURED_ERRORS='old_style'`` is 
+  deprecated.
+- In Numba 0.59: the use of ``NUMBA_CAPTURED_ERRORS='old_style'`` will raise 
+  deprecation warnings.
+- In Numba 0.60: support for ``NUMBA_CAPTURED_ERRORS='new_style'`` becomes the
+  default.
+- In Numba 0.61: support for ``NUMBA_CAPTURED_ERRORS='old_style'`` will be 
+  removed.
```
gmarkall, line 539: Given that the documentation in 0.58 might essentially be missed in a lot of cases and need the warning being raised before it starts being noticed, perhaps we can shorten this cycle by one release if we start raising warnings in 0.58, giving:

```suggestion
- In Numba 0.58: support for ``NUMBA_CAPTURED_ERRORS='old_style'`` is 
  deprecated and will raise deprecation warnings.
- In Numba 0.59: support for ``NUMBA_CAPTURED_ERRORS='new_style'`` becomes the
  default.
- In Numba 0.60: support for ``NUMBA_CAPTURED_ERRORS='old_style'`` will be 
  removed.
```

docs/upcoming_changes/9090.deprecation.rst
```diff
@@ -0,0 +1,6 @@
+Deprecation of old-style ``NUMBA_CAPTURED_ERRORS``
+==================================================
+
+Added deprecation schedule of ``NUMBA_CAPTURED_ERRORS=old_style``.
+``NUMBA_CAPTURED_ERRORS=new_style`` will become default in future releases.
+Details is documented at https://numba.readthedocs.io/en/stable/reference/deprecation.html#deprecation-of-old-style-numba-captured-errors
```
gmarkall, line 6: A minor nit (and perhaps a wrap that would generally occur):
```suggestion
Details are documented at
https://numba.readthedocs.io/en/stable/reference/deprecation.html#deprecation-of-old-style-numba-captured-errors
```

docs/upcoming_changes/9090.deprecation.rst
```diff
@@ -0,0 +1,6 @@
+Deprecation of old-style ``NUMBA_CAPTURED_ERRORS``
+==================================================
+
+Added deprecation schedule of ``NUMBA_CAPTURED_ERRORS=old_style``.
+``NUMBA_CAPTURED_ERRORS=new_style`` will become default in future releases.
```
gmarkall, line 5: A minor nit:
```suggestion
``NUMBA_CAPTURED_ERRORS=new_style`` will become the default in future releases.
```

gmarkall (changes requested):
The text of the deprecation notice and towncrier update generally look and render fine - I've made some additional comments on the diff.

docs/source/reference/deprecation.rst
sklam: When discussing this with @stuartarchibald, we talked about making 0.58 raise warnings when old_style capturing is used but not when the env-var is set to `old_style` which is the default.  Maybe in 0.59, we can differentiate between setting old_style explicitly vs leaving it unset and only add warnings if the env-var is set to `old_style`.

Also, we usually give downstream 6months notice before changing behaviors.

sklam: @gmarkall, i've pushed changes for this at https://github.com/numba/numba/pull/9090/commits/60b8027f90e63e2b9b591eb24627687b0df7c517

gmarkall (dismissed):
Looks good to me.

docs/source/reference/deprecation.rst
gmarkall: Understood - thanks for the clarifications.

esc:
post #9127 this has conflicts.

esc:
Merged `main` and resolved conflicts related to importing at: https://github.com/numba/numba/pull/9090/commits/58d6e9a07b6517236c350e20c43c3d905f4c1a27

esc:
@gmarkall this needs re-approval post merge of `main`, please.

gmarkall (approved):
Re-approving following the merge from `main`.

esc:
> Re-approving following the merge from `main`.

Excellent, thank you.

PR #9094, opened 2023-07-26
Add support for a 'max' level to NUMBA_OPT environment variable.

### LINKED ISSUE #8172: Feature Request: add ability to run -03 twice
This bug first occurs in version 0.54 (0.49-0.53 are OK). This results in a performance decrease of a factor of 4 in a bit optimized summation and a factor of 2 in a naive implementation.

**Example**

```python
import numpy as np
import numba as nb
import time


f = np.random.rand(32,33,2000,64)
b = np.random.rand(32,64)

@nb.njit(fastmath=True)
def test_numba_1(f,b):
    fnew = np.zeros((f.shape[1],f.shape[2]))
    
    for i in range(f.shape[0]):
        for j in range(f.shape[1]):
            for k in range(f.shape[2]):
                for l in range(f.shape[3]):
                    fnew[j,k] += f[i,j,k,l] * b[i,l]
    return fnew

@nb.njit(fastmath=True)
def test_numba_2(f,b):
    fnew = np.zeros((f.shape[1],f.shape[2])) 
    
    for i in range(f.shape[0]):
        for j in range(f.shape[1]):
            for k in range(f.shape[2]):
                acc=fnew[j,k]
                for l in range(f.shape[3]):
                    acc += f[i,j,k,l] * b[i,l]
                fnew[j,k]=acc
    return fnew

%timeit test_numba_1(f,b)
%timeit test_numba_2(f,b)

#'0.53.1' Last version without bug
#108 ms ± 1.35 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
#53.9 ms ± 809 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)

#'0.54.0' First version with bug
#216 ms ± 4.68 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
#200 ms ± 2.93 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

#'0.55.0'
#218 ms ± 5.36 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
#201 ms ± 1.73 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Another question would be if, a rewrite pass is possible in Numba to optimize `test_numba_1` automatically to `test_numba_2`. Clang shows exactly the same behavior as Numba before version 0.54, therefore this likely has to be fixed before generating LLVM-IR.

```c
//clang v11
//clang -shared -O3 -march=native -ffast-math -o summation.dll summation.c
__declspec(dllexport) void eq_nb_1(size_t s0,size_t s1,size_t s2,size_t s3,double f[][s1][s2][s3],double b[][s3],double fnew[][s2])
{
  for (size_t i=0;i<s0;i++){
    for (size_t j=0;j<s1;j++){
      for (size_t k=0;k<s2;k++){
         for (size_t l=0;l<s3;l++){
           fnew[j][k]+=f[i][j][k][l]*b[i][l];
         }
      }
    }
  }
}

__declspec(dllexport) void eq_nb_2(size_t s0,size_t s1,size_t s2,size_t s3,double f[][s1][s2][s3],double b[][s3],double fnew[][s2])
{
  for (size_t i=0;i<s0;i++){
    for (size_t j=0;j<s1;j++){
      for (size_t k=0;k<s2;k++){
         double acc=fnew[j][k];
         for (size_t l=0;l<s3;l++){
           acc+=f[i][j][k][l]*b[i][l];
         }
         fnew[j][k]=acc;
      }
    }
  }
}
```
```python
import numpy as np
import ctypes

lib = ctypes.cdll.LoadLibrary("summation.dll")

def eq_nb_1(f,b):
    fnew = np.zeros((f.shape[1],f.shape[2]))
    s_0=ctypes.c_size_t(f.shape[0])
    s_1=ctypes.c_size_t(f.shape[1])
    s_2=ctypes.c_size_t(f.shape[2])
    s_3=ctypes.c_size_t(f.shape[3])
    
    f_p   =f.ctypes
    b_p   =b.ctypes
    fnew_p=fnew.ctypes
    
    lib.eq_nb_1(s_0,s_1,s_2,s_3,f_p,b_p,fnew_p)
    return fnew

def eq_nb_2(f,b):
    fnew = np.zeros((f.shape[1],f.shape[2]))
    s_0=ctypes.c_size_t(f.shape[0])
    s_1=ctypes.c_size_t(f.shape[1])
    s_2=ctypes.c_size_t(f.shape[2])
    s_3=ctypes.c_size_t(f.shape[3])
    
    f_p   =f.ctypes
    b_p   =b.ctypes
    fnew_p=fnew.ctypes
    
    lib.eq_nb_2(s_0,s_1,s_2,s_3,f_p,b_p,fnew_p)
    return fnew

%timeit eq_nb_1(f,b)
%timeit eq_nb_2(f,b)
#100 ms ± 311 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
#50.5 ms ± 1.08 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

stuartarchibald:
This adds support for a new 'max' level to the NUMBA_OPT environment level. Prior to this values 0, 1, 2 and 3 were supported, these correspond approximately to the -O{level} flag available in many command line based compilation tools. Issue opt=3 both before and after the reference count pruning pass is run (Numba historically had this behaviour as default). This patch re-enables this behaviour and highlights the caveats in the docs (performance may not improve, compilation time may increase and there be no performance gain, but the opposite is also possible, the user can now choose!).

Closes #8172

### DIFF
--- a/docs/source/reference/envvars.rst
+++ b/docs/source/reference/envvars.rst
@@ -299,7 +299,16 @@ Compilation options
 
 .. envvar:: NUMBA_OPT
 
-   The optimization level; this option is passed straight to LLVM.
+   The optimization level; typically this option is passed straight to LLVM. It
+   may take one of the values ``0``, ``1``, ``2`` or ``3`` which correspond
+   approximately to the ``-O{value}`` flag found in many command line
+   compilation tools. The value ``max`` is also supported, this is Numba
+   specific, it has the effect of running with the optimization level set at
+   ``3`` both before and after a pass which in which reference count operation
+   pruning takes place. In some cases this may increase performance, in other
+   cases it may impede performance, the same can be said for compilation time.
+   This option is present to give users the opportunity to choose a value
+   suitable for their application.
 
    *Default value:* 3
 
--- a/numba/core/config.py
+++ b/numba/core/config.py
@@ -82,10 +82,11 @@ def _validate_captured_errors_style(style_str):
 
 class _OptLevel(int):
     """This class holds the "optimisation level" set in `NUMBA_OPT`. As this env
-    var can be an int or a string, but is almost always interpreted as an int.
-    This class subclasses int so as to get the common behaviour but stores the
-    actual value as a `_raw_value` member to make it available for cases where
-    accounting for the specific string supplied by the user is necessary."""
+    var can be an int or a string, but is almost always interpreted as an int,
+    this class subclasses int so as to get the common behaviour but stores the
+    actual value as a `_raw_value` member. The value "max" is a special case
+    and the property `is_opt_max` can be queried to find if the optimisation
+    level (supplied value at construction time) is "max"."""
 
     def __new__(cls, *args, **kwargs):
         assert len(args) == 1
@@ -96,6 +97,12 @@ class _OptLevel(int):
         # raw value is max or int
         new._raw_value = value if value == 'max' else _int_value
         return new
+
+    @property
+    def is_opt_max(self):
+        """Returns True if the the optimisation level is "max" False
+        otherwise."""
+        return self._raw_value == "max"
 
     def __repr__(self):
         if isinstance(self._raw_value, str):
--- a/numba/core/config.py
+++ b/numba/core/config.py
@@ -80,6 +80,42 @@ def _validate_captured_errors_style(style_str):
         return rendered_style
 
 
+class _OptLevel(int):
+    """This class holds the "optimisation level" set in `NUMBA_OPT`. As this env
+    var can be an int or a string, but is almost always interpreted as an int.
+    This class subclasses int so as to get the common behaviour but stores the
+    actual value as a `_raw_value` member to make it available for cases where
+    accounting for the specific string supplied by the user is necessary."""
+
+    def __new__(cls, *args, **kwargs):
+        assert len(args) == 1
+        (value,) = args
+        _int_value = 3 if value == 'max' else int(value)
+        # the int ctor is always called with an appropriate integer value
+        new = super().__new__(cls, _int_value, **kwargs)
+        # raw value is max or int
+        new._raw_value = value if value == 'max' else _int_value
+        return new
+
+    def __repr__(self):
+        if isinstance(self._raw_value, str):
+            arg = f"'{self._raw_value}'"
+        else:
+            arg = self._raw_value
+        return f"_OptLevel({arg})"
+
+
+def _process_opt_level(opt_level):
+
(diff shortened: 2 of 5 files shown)


docs/source/reference/envvars.rst
```diff
@@ -299,7 +299,16 @@ Compilation options
 
 .. envvar:: NUMBA_OPT
 
-   The optimization level; this option is passed straight to LLVM.
+   The optimization level; typically this option is passed straight to LLVM. It
+   may take one of the values ``0``, ``1``, ``2`` or ``3`` which correspond
+   approximately to the ``-O{value}`` flag found in many command line
+   compilation tools. The value ``max`` is also supported, this is Numba
+   specific, it has the effect of running with the optimization level set at
```
sklam, line 306: Can we phrase it so that what `max` does exactly may change but it currently does it the certain way? We don't need to promise that it will always run O3 twice. It may change as we discover other ways. The key thing is that its effect is to sacrifice compile time for better runtime.

numba/core/config.py
```diff
@@ -80,6 +80,42 @@ def _validate_captured_errors_style(style_str):
...
+    accounting for the specific string supplied by the user is necessary."""
+
+    def __new__(cls, *args, **kwargs):
+        assert len(args) == 1
+        (value,) = args
+        _int_value = 3 if value == 'max' else int(value)
+        # the int ctor is always called with an appropriate integer value
+        new = super().__new__(cls, _int_value, **kwargs)
+        # raw value is max or int
+        new._raw_value = value if value == 'max' else _int_value
```
sklam, line 97: This class can have a `is_max_opt` method/property so codegen doesn't need to access this attribute.

sklam (commented):
some suggestions

numba/core/config.py
stuartarchibald: Thanks, I'll make this change because it's preferable to not access the attribute directly, _but_, if in future for some reason we have to add numerous special cases, this may need revisiting as an approach.

docs/source/reference/envvars.rst
stuartarchibald: That's a good point, I'll rephrase prose that describe this effect to contain something more general.

stuartarchibald: Implemented in c91b102.

numba/core/config.py
stuartarchibald: Implemented in c91b102.

sklam (approved):

PR #9093, opened 2023-07-26
Updates the minimum supported NumPy to 1.22.

stuartarchibald:
This patch follows NEP-0029 and updates the minimum supported NumPy version to 1.22. The changes in this patch are mechanical deletions and code motion around version guards.

Closes #9047

PR #9099, opened 2023-07-27
Make all parameter names in `@overload`s match the API being overloaded.

### LINKED ISSUE #9053: Numpy Gamma distribution not working in Numba 0.57+

### LINKED ISSUE #9073: Call by named arg broken for some `@overload` definitions
Numba has an issue with some internal `@overload` implementations whereby the definition of the implementation contains different argument names to the function that's being overload. In general this isn't noticed, but occasionally a user might try to call a function with named arguments for what would normally be the positional arguments. Example:

This overloads `np.geomspace()` it has a signature in NumPy like `np.geomspace(start, stop, num=50, <more arguments>)`. If if an `@overload` that exhibits the problem is created like:
```python
from numba import njit
import numpy as np

# NOTE: the name of the first arg is incorrect, NumPy API has `start`,
# this has `WRONG_NAME`.
@overload(np.geomspace)
def ol_geomspace(WRONG_NAME, stop, num=50):
    def impl(WRONG_NAME, stop, num=50):
        print("compiled and ran!")
    return impl

# ------------------------------------------------------------------------------
# This works fine
@njit
def foo1():
    np.geomspace(1, 2)

foo1()


# ------------------------------------------------------------------------------
# This works fine
@njit
def foo2():
    np.geomspace(1, stop=2)

foo2()


# ------------------------------------------------------------------------------
# This does not work, the overload expects `WRONG_NAME` but gets `start`.
# The user is just using the NumPy API as expected.
@njit
def foo3():
    np.geomspace(start=1, stop=2)

foo3()
```

Unfortunately there's quite a few functions defined as `@overload`s in Numba with this problem.

Two questions:
1. Where are these incorrectly defined functions in the current code base, they need fixing?
2. Should Numba have some mechanism to prevent incorrect definition for future `@overload` implementations?

As to 1. applying this patch locally seems to list most of the offending functions:
```diff
diff --git a/numba/core/extending.py b/numba/core/extending.py
--- a/numba/core/extending.py
+++ b/numba/core/extending.py
@@ -1,5 +1,6 @@
 import os
 import uuid
+import inspect
 import weakref
 import collections
 import functools
@@ -127,6 +128,33 @@ def overload(func, jit_options={}, strict=True, inline='never',
     # this requires sorting out the circular imports first.
 
     def decorate(overload_func):
+        try:
+            func_sig = inspect.signature(func)
+        except ValueError:
+            pass
+        else:
+            ol_sig = inspect.signature(overload_func)
+            x = list(func_sig.parameters.keys())
+            y = list(ol_sig.parameters.keys())
+            for a, b in zip(x[:len(y)], y):
+                if a != b:
+                    p = func_sig.parameters[a]
+                    if p.kind == p.POSITIONAL_ONLY:
+                        # probably a builtin written in C
+                        continue
+                    elif '*' in str(p): # probably *args or similar
+                        continue
+                    else:
+                        if not func.__module__.startswith("numba"):
+                            print(f"{func} from module '{getattr(func, '__module__')}' has mismatched sig.")
+                            print(f"    - expected: {func_sig}")
+                            print(f"    -      got: {ol_sig}")
+                            lineno = inspect.getsourcelines(overload_func)[1]
+                            tmpsrcfile = inspect.getfile(overload_func)
+                            srcfile = tmpsrcfile.replace(numba.__path__[0], '')
+                            print(f"from {srcfile}:{lineno}")
+                            print()
+                            break
         template = make_overload_template(func, overload_func, opts, strict,
                                           inline, prefer_literal, **kwargs)
         infer(template)

```

stuartarchibald:
This is a continuation of #9059, with thanks to @apmasell for this patch, authorship is preserved.

What this PR is about: Some of the Numba ``@overload``s for functions in NumPy and Python's built-ins were written using parameter names that did not match those used in API they were overloading. The result of this being that calling a function with such a mismatch using the parameter names as key-word arguments at the call site would result in a compilation error. This has now been universally fixed throughout the code base and a unit test is running with a best-effort attempt to prevent reintroduction of similar mistakes in the future.

Review order suggestion:
* Look at the combined diff of the original 5bcbb3a, cf75079 (adding test) and 587b15b (fixing further issues and de-duplicating code) to see that #9053 is addressed.
* Look solely at 86bf5cf next, this patch adds a function that scans all `@overload` API use and validates the function parameters. If you are happy with what this function does and the unit test that uses it, then the subsequent patches are just fixing the things it highlights as somewhat "mechanical" changes, each patch is isolated to a single module as per the commit message title.
* Look at 1bd2c6d which is a change log update.

Fixes #9053
Closes #9073

### DIFF
--- a/docs/upcoming_changes/9099.bug_fix.rst
+++ b/docs/upcoming_changes/9099.bug_fix.rst
@@ -1,0 +1,69 @@
+Fix all ``@overload``s to use parameter names that match public APIs.
+=====================================================================
+
+Some of the Numba ``@overload``s for functions in NumPy and Python's built-ins
+were written using parameter names that did not match those used in API they
+were overloading. The result of this being that calling a function with such a
+mismatch using the parameter names as key-word arguments at the call site would
+result in a compilation error. This has now been universally fixed throughout
+the code base and a unit test is running with a best-effort attempt to prevent
+reintroduction of similar mistakes in the future. Fixed functions include:
+
+From Python built-ins:
+
+* ``complex``
+
+From the Python ``random`` module:
+
+* ``random.seed``
+* ``random.gauss``
+* ``random.normalvariate``
+* ``random.randrange``
+* ``random.randint``
+* ``random.uniform``
+* ``random.shuffle``
+
+From the ``numpy`` module:
+
+* ``numpy.argmin``
+* ``numpy.argmax``
+* ``numpy.array_equal``
+* ``numpy.average``
+* ``numpy.count_nonzero``
+* ``numpy.flip``
+* ``numpy.fliplr``
+* ``numpy.flipud``
+* ``numpy.iinfo``
+* ``numpy.isscalar``
+* ``numpy.imag``
+* ``numpy.real``
+* ``numpy.reshape``
+* ``numpy.rot90``
+* ``numpy.swapaxes``
+* ``numpy.union1d``
+* ``numpy.unique``
+
+From the ``numpy.linalg`` module:
+
+* ``numpy.linalg.norm``
+* ``numpy.linalg.cond``
+* ``numpy.linalg.matrix_rank``
+
+From the ``numpy.random`` module:
+
+* ``numpy.random.beta``
+* ``numpy.random.chisquare``
+* ``numpy.random.f``
+* ``numpy.random.gamma``
+* ``numpy.random.hypergeometric``
+* ``numpy.random.lognormal``
+* ``numpy.random.pareto``
+* ``numpy.random.randint``
+* ``numpy.random.random_sample``
+* ``numpy.random.ranf``
+* ``numpy.random.rayleigh``
+* ``numpy.random.sample``
+* ``numpy.random.shuffle``
+* ``numpy.random.standard_gamma``
+* ``numpy.random.triangular``
+* ``numpy.random.weibull``
--- a/docs/upcoming_changes/9059.bug_fix.rst
+++ b/docs/upcoming_changes/9059.bug_fix.rst
@@ -1,0 +1,7 @@
+
+Fix parameter names for ``random`` functions
+============================================
+
+This corrects a bunch of parameter names that did not match the names in
+``random`` and ``numpy.random`` so that using name parameters will work
+properly for these functions. Positional parameters worked correctly.
--- a/docs/upcoming_changes/9059.bug_fix.rst
+++ b/docs/upcoming_changes/9059.bug_fix.rst
@@ -1,7 +1,0 @@
-
-Fix parameter names for ``random`` functions
-============================================
-
-This corrects a bunch of parameter names that did not match the names in
-``random`` and ``numpy.random`` so that using name parameters will work
-properly for these functions. Positional parameters worked correctly.
--- a/numba/cpython/randomimpl.py
+++ b/numba/cpython/randomimpl.py
@@ -198,30 +198,39 @@ def get_next_int(context, builder, state_ptr, nbits, is_numpy):
 
 
 @overload(random.seed)
-def seed_impl(seed):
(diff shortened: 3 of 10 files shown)


stuartarchibald:
NOTE: The NumPy 1.21 tests are failing because:

The name of the first argument to `numpy.linalg.matrix_rank` changed between NumPy 1.21 (where it was `M`) and NumPy 1.22 (where it is `A`). This manifests as failing test (which was added in this PR, and is evidently working as it's catching this sort of thing):
```
numba.tests.test_extending.TestNumbaInternalOverloads.test_signatures_match_overloaded_api 
```
with exception:
```
ValueError: 
<function matrix_rank at 0x7f95fccbfa60> from module 'numpy.linalg' has mismatched sig.
    - expected: (M, tol=None, hermitian=False)
    -      got: (A, tol=None)
from /np/linalg.py:2495
```
As the base line NumPy version is being moved to 1.22 for Numba 0.58 (#9093), it's not worth fixing this, once that PR is merged the problem is resolved!

docs/upcoming_changes/9099.bug_fix.rst
```diff
@@ -0,0 +1,69 @@
+Fix all ``@overload``s to use parameter names that match public APIs.
+=====================================================================
+
+Some of the Numba ``@overload``s for functions in NumPy and Python's built-ins
```
sklam, line 4:
```suggestion
Some of the Numba ``@overload``\ s for functions in NumPy and Python's built-ins
```

sklam (commented):
I've read through the changes and it looks okay.

Side-notes:  too bad this needs the `register_jittable` wrapper hack

docs/upcoming_changes/9099.bug_fix.rst
```diff
@@ -0,0 +1,69 @@
+Fix all ``@overload``s to use parameter names that match public APIs.
```
sklam, line 1:
```suggestion
Fix all ``@overload``\ s to use parameter names that match public APIs.
```

sklam:
https://github.com/numba/numba/pull/9111 will fix the ReST syntax error at https://github.com/numba/numba/pull/9099#pullrequestreview-1555369076

esc:
@sklam can this be closed, since it has been abandoned?

sklam:
superseded by https://github.com/numba/numba/pull/9111

PR #9106, opened 2023-07-28
CUDA: Add overloads generated by specialization to the current dispatcher.

gmarkall:
Adding overloads to the current dispatcher when specializing saves recreating the overload for subsequent calls to the unspecialized dispatcher that would have used the overload.

This also has the side effects of making overloads available after a call to `ForAll()` on a dispatcher.

This provides an alternative to the implementation in #9057, which also aimed to make overloads available after a `ForAll()` call, but did it by changing the way `ForAll()` worked and potentially increasing latency for calls on specialized dispatchers (of which there are some use cases in cuDF).

Whilst looking into this I also noticed that all the CUDA dispatcher specialize tests worked in an "unexpected" way and didn't provide a good pattern to follow for this test added in this PR, so they are all fixed up in this PR too.


gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

Matt711:
The PR still works with my cuDF example for `numba-inspector`.

numba/cuda/tests/cudapy/test_dispatcher.py
```diff
@@ -92,20 +100,54 @@ def f(x, y):
...
+            pass
+
+        self.assertEqual(len(f.overloads), 0)
+
+        # 'F' order specialization
+        f_arg = np.zeros((2, 2), order='F')
+        f_arg_ty = typeof(f_arg)
+        f_f32a_f32a = f.specialize(f_arg, f_arg)
+        self.assertEqual(len(f.overloads), 1)
+        self.assertIs(f_f32a_f32a.overloads[f_arg_ty, f_arg_ty],
```
stuartarchibald, line 134:
```suggestion
        f_f32f_f32f = f.specialize(f_arg, f_arg)
        self.assertEqual(len(f.overloads), 1)
        self.assertIs(f_f32f_f32f.overloads[f_arg_ty, f_arg_ty],
```
suggest naming scheme consistency with `F`-ordered input.

numba/cuda/tests/cudapy/test_dispatcher.py
```diff
@@ -92,20 +100,54 @@ def f(x, y):
 
         self.assertEqual(len(f.specializations), 0)
 
-        # 'A' order specialization
-        f_f32a_f32a = f.specialize(float32[:], float32[:])
+        # 'F' order specialization
+        f_arg = np.zeros((2, 2), order='F')
+        f_f32a_f32a = f.specialize(f_arg, f_arg)
```
stuartarchibald, line 105:
```suggestion
        f_f32f_f32f = f.specialize(f_arg, f_arg)
```
suggestion to match ordering `F`.

stuartarchibald (commented):
Thanks for the patch, couple of minor suggestions RE variable naming else looks good.

gmarkall:
Still planning to finish this when I get a moment.

gmarkall:
Missed this due to being on PTO, going to reopen it to keep it relevant (to get round to one day).

gmarkall:
Created Issue https://github.com/NVIDIA/numba-cuda/issues/100 to avoid losing track of this work over there.

gmarkall:
Probably not worth keeping open over here.

PR #9108, opened 2023-07-28
Add noalias option to jit decorator.

DrTodd13:
Adds noalias option to jit decorator which the programmer can use to assert that none of the arguments to the function can share memory.  Ultimately, this will cause all the pointers in the memory models for the argument to have the noalias LLVM qualifier set when the Numba IR is lowered to LLVM.  This will enable more code to vectorize since the possibility of aliasing will frequently cause the LLVM vectorizer to give up.

Previously, parallel=True would try to do an analysis to see whether it can add the noalias flag to the arguments passed to the parallel region gufunc.  This PR improves that process and uses the aliasing information from the outer function better and so if noalias is given for the outer function it will often enable parallel=True to add the noalias flag to the outlined gufunc as well.


itamarst:
Any progress on this? Would love to see this make it in.

PR #9100, opened 2023-07-27
Add towncrier news snippets for PRs that are missing them.

stuartarchibald:
This adds news snippets for PRs that are scheduled for Numba 0.58 but were merged prior to towncrier being enabled and a news snippet being a requirement.

Closes #9070

PR #8995, opened 2023-05-31
Allow libraries that implement __array_ufunc__ to override DUFunc.__c…

jpivarski:
This PR addresses a long-standing issue in which array-like objects that define an `__array_ufunc__` method ([NEP-18](https://numpy.org/neps/nep-0013-ufunc-overrides.html)) can't be used with ufuncs created by `np.vectorize` unless a signature is provided. (I should have done this years ago!)

For instance, in Awkward Array,

```python
>>> import awkward as ak
>>> import numba as nb

>>> nb.vectorize([nb.float64(nb.float64)])(lambda x: x + 1)(ak.Array([[1.1, 2.2, 3.3], [], [4.4, 5.5]]))
<Array [[2.1, 3.2, 4.3], [], [5.4, 6.5]] type='3 * var * float64'>
```

but

```python
>>> nb.vectorize()(lambda x: x + 1)(ak.Array([[1.1, 2.2, 3.3], [], [4.4, 5.5]]))
<stdin>:1: NumbaWarning: 
Compilation is falling back to object mode WITHOUT looplifting enabled because Function "<lambda>" failed type inference due to: non-precise type pyobject
During: typing of argument at <stdin> (1)

File "<stdin>", line 1:
<source missing, REPL/exec in use?>

/home/jpivarski/mambaforge/lib/python3.9/site-packages/numba/core/object_mode_passes.py:151: NumbaWarning: Function "<lambda>" was compiled in object mode without forceobj=True.

File "<stdin>", line 1:
<source missing, REPL/exec in use?>

  warnings.warn(errors.NumbaWarning(warn_msg,
/home/jpivarski/mambaforge/lib/python3.9/site-packages/numba/core/object_mode_passes.py:161: NumbaDeprecationWarning: 
Fall-back from the nopython compilation path to the object mode compilation path has been detected. This is deprecated behaviour that will be removed in Numba 0.59.0.

For more information visit https://numba.readthedocs.io/en/stable/reference/deprecation.html#deprecation-of-object-mode-fall-back-behaviour-when-using-jit

File "<stdin>", line 1:
<source missing, REPL/exec in use?>

  warnings.warn(errors.NumbaDeprecationWarning(msg,
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/jpivarski/mambaforge/lib/python3.9/site-packages/numba/np/ufunc/dufunc.py", line 206, in _compile_for_args
    return self._compile_for_argtys(tuple(argtys))
[...]

PR #9112, opened 2023-07-31
Fixups for PR#9100

sklam:
Address review suggestions in https://github.com/numba/numba/pull/9100


gmarkall (approved):
This looks good - the docs for PRs #8974 and #8995 both render correctly, and the item for #8736 is removed.

PR #9113, opened 2023-08-01
Add support for np.diagflat

KrisMinchev:
This PR improves on #4074 by implementing the function `np.diagflat`.


guilhermeleobas (changes requested):
Hi Kristian, thanks for the addition. Only one small change and PR should be good to merge

guilhermeleobas (approved):

PR #9105, opened 2023-07-28
NumPy 1.25 support (PR #9011) continued

gmarkall:
This continues PR #9011, which has been closed as the original author is no longer working on it. The branch for this PR is the same as it was on #9011, with the `test_parfors.py` change reverted as requested in https://github.com/numba/numba/pull/9011/files#r1261153152.

PR #9114, opened 2023-08-01
update np min to 122

### LINKED ISSUE #9047: Drop support for Numpy 1.21
## Feature request

As per:

https://github.com/numba/numba/wiki/Minutes_2023_06_27

And:

https://numpy.org/neps/nep-0029-deprecation_policy.html

We have elected to drop support for Numpy 1.21 with Numba 0.58.

esc:
Adoption of #9093 

Fixes #9047.


esc:
Merged in `main` post merge of #9105

gmarkall:
gpuci run tests

gmarkall (dismissed):
Approved on the basis of #9093 already being approved and this is exactly the same plus a merge from `main`.

gmarkall:
Seems there's a problem with the docs build, so moved back to "Waiting on author".

gmarkall:
Ah, the problem is that `ufunc_db` no longer imports `numpy_version`.

numba/np/ufunc_db.py
```diff
@@ -49,7 +49,6 @@ def _fill_ufunc_db(ufunc_db):
     # module.
     from numba.np import npyfuncs
     from numba.cpython import cmathimpl, mathimpl, numbers
-    from numba.np.numpy_support import numpy_version
```
gmarkall, line 52: I think this needs to stay now that it is used to check for NumPy 1.25.

gmarkall:
gpuci run tests

numba/np/ufunc_db.py
esc: fixed in 1aef98a6bfba22d5c04615769ef03d1d9d4915ac

esc:
I ran this through the Build-Farm as `numba_smoketest_cpu_yaml_117` and it passed fine. 🍏

PR #9115, opened 2023-08-02
Add support for np.diag_indices() and np.diag_indices_from()

KrisMinchev:
This PR improves on #4074 by implementing the functions `np.diag_indices()` and `np.diag_indices_from()`. Note that the NumPy versions return tuples whereas the implementations presented here return NumPy arrays.

### DIFF
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -4532,7 +4532,7 @@ def np_diag_indices(n, ndim=2):
     if not isinstance(ndim, (int, types.Integer)):
         msg = 'The argument "ndim" must be an integer'
         raise errors.TypingError(msg)
-    
+
     # tup_init is used only to get the correct tuple type
     tup_init = (np.arange(1),) * 2
 
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -4544,13 +4544,13 @@ def np_diag_indices(n, ndim=2):
 
 @overload(np.diag_indices_from)
 def np_diag_indices_from(arr):
-    if not type_can_asarray(arr):
-        msg = 'The argument "arr" must be array-like'
+    if not isinstance(arr, types.Array):
+        msg = 'The argument "arr" must be an array'
         raise errors.TypingError(msg)
 
     def impl(arr):
         if not arr.ndim >= 2:
-            raise ValueError("input array must be at least 2-d")
+            raise ValueError("Input array must be at least 2-d")
         # For more than d=2, the strided formula is only valid for arrays with
         # all dimensions equal, so we check first.
         s = np.asarray(arr.shape)
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -4532,12 +4532,21 @@ def np_diag_indices(n, ndim=2):
     if not isinstance(ndim, (int, types.Integer)):
         msg = 'The argument "ndim" must be an integer'
         raise errors.TypingError(msg)
+    
+    # tup_init is used only to get the correct tuple type
+    tup_init = (np.arange(1),) * 2
+
+    from numba.cpython.unsafe.tuple import tuple_setitem
 
     def impl(n, ndim=2):
         res = np.arange(n * ndim)
         for i in range(n * ndim):
             res[i] = res[i] % n
-        return res.reshape((ndim, n))
+        x = res.reshape((ndim, n))
+        tup = tup_init
+        for i in range(ndim):
+            tup = tuple_setitem(tup, i, np.copy(x[i]))
+        return tup
 
     return impl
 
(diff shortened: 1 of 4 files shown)


numba/np/arrayobj.py
```diff
@@ -4523,6 +4523,45 @@ def impl(N, M=None, k=0, dtype=float):
...
+
+    if not isinstance(ndim, (int, types.Integer)):
+        msg = 'The argument "ndim" must be an integer'
+        raise errors.TypingError(msg)
+
+    def impl(n, ndim=2):
+        res = np.arange(n * ndim)
+        for i in range(n * ndim):
+            res[i] = res[i] % n
+        return res.reshape((ndim, n))
```
guilhermeleobas, line 4540: Numba has a function called `tuple_setitem` from `cpython/unsafe/tuple.py` where you can use to set items in a tuple at runtime.

```suggestion
    # tup_init is used only to get the correct tuple type
    tup_init = (np.arange(1),) * 2
    from numba.cpython.unsafe.tuple import tuple_setitem

    def impl(n, ndim=2):
        res = np.arange(n * ndim)
        for i in range(n * ndim):
            res[i] = res[i] % n
        x = res.reshape((ndim, n))

        tup = tup_init
        for i in range(n):
            tup = tuple_setitem(tup, i, x[i])
        return tup
```

numba/tests/test_np_functions.py
```diff
@@ -5547,6 +5559,103 @@ def test_nan_to_num_invalid_argument(self):
...
+                      str(raises.exception))
+
+    def test_diag_indices_from_basic(self):
+        pyfunc = diag_indices_from
+        cfunc = njit(pyfunc)
+
+        def inputs():
+            # Taken from https://github.com/numpy/numpy/blob/...
+            # 09ee7daa9113ecb503552913d76185d0ad3f0f31/numpy/lib/tests/...
+            # test_index_tricks.py#L513-L517
```
guilhermeleobas, line 5624: Keep URL in a single line

numba/tests/test_np_functions.py
```diff
@@ -5547,6 +5559,103 @@ def test_nan_to_num_invalid_argument(self):
...
+                      str(raises.exception))
+
+        with self.assertRaises(TypingError) as raises:
+            cfunc("abc")
+        self.assertIn('The argument "arr" must be an array',
+                      str(raises.exception))
+
+        # Based on tests from https://github.com/numpy/numpy/blob/...
+        # 09ee7daa9113ecb503552913d76185d0ad3f0f31/numpy/lib/tests/...
+        # test_index_tricks.py#L519-L527
```
guilhermeleobas, line 5648: Keep URL in a single line

guilhermeleobas (changes requested):
Hi @KrisMinchev, great work on adding these functions to Numba. 

> Note that the NumPy versions return tuples whereas the implementations presented here return NumPy arrays.

Numba tries to keep compatibility with NumPy as close as possible. I've left a suggestion on how you can return tuples instead of NumPy arrays. The idea is to create a tuple at compile-time with the same type as the tuple you will return at runtime, and use `tuple_setitem` to set each tuple item.

numba/np/arrayobj.py
KrisMinchev: Hi @guilhermeleobas, thank you for your suggestion. The only issue with this implementation is that the array elements that are being returned are immutable, which is different to what the NumPy function `np.diagflat` returns:
```python
import numpy as np
from numba import njit
@njit
def foodiagindices(n, ndim=2):
    res = np.diag_indices(n, ndim)
    return res
a=np.diag_indices(5)
print(a)
a[0][0] = 5
print(a)
c=foodiagindices(5)
print(c)
c[0][0] = 5
```
Returns

> (array([0, 1, 2, 3, 4]), array([0, 1, 2, 3, 4]))
(array([5, 1, 2, 3, 4]), array([5, 1, 2, 3, 4]))
(array([0, 1, 2, 3, 4]), array([0, 1, 2, 3, 4]))
  File "/home/kristian/Desktop/numba/a.py", line 13, in <module>
    c[0][0] = 5
ValueError: assignment destination is read-only

Another issue is that your proposed implementation implicitly assumes that `ndim` is equal to 2 at runtime by setting ` tup_init = (np.arange(1),) * 2`. Hence, the following code:
```python
import numpy as np
from numba import njit
@njit
def foodiagindices(n, ndim=2):
    res = np.diag_indices(n, ndim)
    return res
c = foodiagindices(5, 3)
```
throws a segmentation fault.

KrisMinchev:
The current implementation does not work as intended. It is hard to keep the behaviour of the NumPy function `np.diag_indices(n, ndim)` since the size of the tuple being is equal to ndim and hence cannot be determined at compile time. Since `np.diag_indices_from()` calls `np.diag_indices()`, the same issue is present there as well.

guilhermeleobas:
In today's public meeting ([notes](https://hackmd.io/UYPwMZaxSwCUA-5rAP0qJg)), it was decided to support `diag_indices` with `ndim` as a literal argument. 

Kristian, do you have any other question regarding supporting this function?

PR #9118, opened 2023-08-03
Add support for np.resize()

KrisMinchev:
This PR improves on #4074 by implementing the function `np.resize()`.

### DIFF
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -2149,6 +2149,73 @@ def np_reshape(a, shape):
     return np_reshape_impl
 
 
+@overload(np.resize)
+def numpy_resize(a, new_shape):
+
+    if not type_can_asarray(a):
+        msg = 'The argument "a" must be array-like'
+        raise errors.TypingError(msg)
+
+    if not ((isinstance(new_shape, types.UniTuple)
+             and
+             isinstance(new_shape.dtype, types.Integer))
+            or
+            isinstance(new_shape, types.Integer)):
+        msg = ('The argument "new_shape" must be an integer or '
+               'a tuple of integers')
+        raise errors.TypingError(msg)
+
+    if isinstance(new_shape, types.Integer):
+        def impl(a, new_shape):
+            a = np.asarray(a)
+            a = np.ravel(a)
+
+            if new_shape < 0:
+                msg = 'All elements of `new_shape` must be non-negative'
+                raise ValueError(msg)
+
+            new_size = new_shape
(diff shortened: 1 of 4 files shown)


numba/np/arrayobj.py
```diff
@@ -2149,6 +2149,73 @@ def np_reshape_impl(a, shape):
...
+    if not ((isinstance(new_shape, types.UniTuple)
+             and
+             isinstance(new_shape.dtype, types.Integer))
+            or
+            isinstance(new_shape, types.Integer)):
+        msg = ('The argument "new_shape" must be an integer or '
+               'a tuple of integers')
+        raise errors.TypingError(msg)
+
+    if isinstance(new_shape, types.Integer):
```
guilhermeleobas, line 2168: You can actually unify both implementations and move the isinstance check to inside `impl`:

```python
def impl(a, new_shape):
    ...
    if isinstance(new_shape, tuple):
        new_size = 1
        for dim_length in np.asarray(new_shape):
            new_size *= dim_length
            if dim_length < 0:
                msg = 'All elements of `new_shape` must be non-negative'
                raise ValueError(msg)
    else:
        if new_shape < 0:
            msg2 = 'All elements of `new_shape` must be non-negative'
            raise ValueError(msg2)
        new_size = new_shape

    if a.size == 0:
        return np.zeros(new_shape).astype(a.dtype)

    ....
```

The `isinstance` check will be eliminated at compile time and only the branch that is true is kept.

guilhermeleobas (changes requested):
Thanks for the addition, @KrisMinchev. I only have one request change.

KrisMinchev:
Hi @guilhermeleobas, thank you for the review. I have merged the implementations and moved the `isinstance` check.

guilhermeleobas (approved):
LGTM. Thanks, @KrisMinchev.

esc:
@KrisMinchev thank you for the patch!

mtsokol:
Hi @guilhermeleobas, is `np.resize()` missing from the "supported NumPy functions" section? I.e. Will `@njit` work with `np.resize`?

https://numba.pydata.org/numba-doc/dev/reference/numpysupported.html#functions

guilhermeleobas:
Hi @mtsokol, it seems that `np.resize()` is missing from the "supported NumPy functions" section. But it should work just fine in Numba.

esc:
@mtsokol pull-requests to update the documentation welcome, if anything is amiss ❤️

Issue #9097, opened 2023-07-27
Dispatch error


albop:
Under some conditions, the dispatch behavior of a function defined with overload seems unpredictible. Issue was initially raised here: https://github.com/EconForge/interpolation.py/issues/104. 
It can be replicated with the two attached files (run   `error.py` which calls some functions defined in `source.py`). Both are compressed in [bug.zip](https://github.com/numba/numba/files/12181883/bug.zip)

To summarize, error.py contains the following loop:

```python
import numpy as np
from source import eval_linear

grid = (np.linspace(0, 10, 11), np.linspace(0, 10, 11))
z1 = np.random.randn(11,11)
z1z2 = np.random.randn(11,11,2)
LINEAR = ((None,), (None,) * 2)  # a constant used for dispatching only

npoints = 300
eval_points = np.column_stack([np.linspace(0, 10, npoints)]*2)

for i in range(1000):

    z1_eval = eval_linear( grid,  z1,  eval_points, LINEAR )
    try:
        assert(z1_eval.shape == (npoints,))
    except Exception as e:
        print(f"Failed in iteration {i}")
        raise(e)

    z1z2_eval = eval_linear( grid,  z1z2, eval_points,  LINEAR)
    try:
        assert(z1z2_eval.shape == (npoints,2))
    except Exception as e:
        print(f"Failed in iteration {i}")
        raise(e)
```

In that loop, `eval_linear` is a nopython jitted function which returns a vector in the first call when z1 is a 2d array (values on the 2d grid to interpolate) but returns a matrix in the second call where z1z2 is a 3d array (vector -values on the 2d grid).

When iterating long enough I always get an error but not always for the same call, with the function returning the wrong type of output. For instance:

```
Failed in iteration 129
Traceback (most recent call last):
  File "/home/pablo/Econforge/interpolation.py/error_min.py", line 26, in <module>
    raise(e)
  File "/home/pablo/Econforge/interpolation.py/error_min.py", line 23, in <module>
    assert(z1z2_eval.shape == (npoints,2))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

The original definition of `eval_linear` comes from the interpolation.py library. I tried to reproduce a simplified version of it in the file `source.py`. The definition is somewhat convoluted (and certainly suboptimal), because:
- a function `eval_spline` does many different things based on the kind of (keyword) arguments it get. In the original library code gets generated. Here I included here just the two combinations needed for the example.
[...]

PR #9119, opened 2023-08-03
`memcpy` static buffer content into newly allocated buffer

guilhermeleobas:
Fixes issue #9097


stuartarchibald (commented):
Thanks for the patch, is it possible to add a test for this change please? Many Thanks!

stuartarchibald:
@guilhermeleobas this seems to be a reasonably reliable reproducer (I get a hit every time locally), it's based on the discussion on discourse RE long named tuple fields.
```python
from numba import jit, types, typeof

longname = 1024 * "a"

AType = namedtuple("AType", longname)
BType = namedtuple("BType", longname)

args = (AType(1), BType(1))

@jit(types.void(*(typeof(x) for x in args)))
def foo(arg1, arg2):
    pass

foo(*args)
```
If this reproduces for you, could this perhaps be a replacement for the unit test in this PR (which IIRC doesn't always reproduce?). Thanks!

guilhermeleobas:
Done! @stuartarchibald

stuartarchibald:
Thanks for adapting and adopting the suggestion @guilhermeleobas, the patch and test looks good to me, but as this now effectively contains my code, I'll ask @sklam for a final sign off. It's really great to get this fixed!

sklam (approved):
Changes make sense.

stuartarchibald:
Thanks @sklam!

PR #9120, opened 2023-08-04
Update conda-recipe for numba-rvsdg

sklam:
Base on #9012.
Depends on https://github.com/numba/numba-rvsdg/pull/91.

- https://github.com/numba/numba/pull/9120/commits/80eb15079f6e2b8f1fac69e9cf82febd4fca25d6 adds runtime constraints for numba-rvsdg.
- https://github.com/numba/numba/pull/9120/commits/1b19c223f2a5317015faf824feba9eee5c3f26f2 adds import guard and direct user to install numba-rvsdg when the rvsdg-frontend subpackage is used.

### DIFF
--- a/buildscripts/condarecipe.local/meta.yaml
+++ b/buildscripts/condarecipe.local/meta.yaml
@@ -63,6 +63,8 @@ requirements:
     - scipy >=1.0
     # CUDA Python 11.6 or later
     - cuda-python >=11.6
+    # numba-rvsdg
+    - numba-rvsdg ==0.0.2      # [py==311]
 
 test:
   requires:
--- a/buildscripts/condarecipe.local/meta.yaml
+++ b/buildscripts/condarecipe.local/meta.yaml
@@ -64,7 +64,7 @@ requirements:
     # CUDA Python 11.6 or later
     - cuda-python >=11.6
     # numba-rvsdg
-    - numba-rvsdg ==0.0.2      # [py==311]
+    - numba-rvsdg 0.0.*      # [py==311]
 
 test:
   requires:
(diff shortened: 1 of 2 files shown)


sklam:
Rebase and force pushed so we have a simpler history.

buildscripts/condarecipe.local/meta.yaml
```diff
@@ -63,6 +63,8 @@ requirements:
     - scipy >=1.0
     # CUDA Python 11.6 or later
     - cuda-python >=11.6
+    # numba-rvsdg
+    - numba-rvsdg 0.0.*      # [py==311]
```
sklam, line 67: I'd suggest we don't pin to any specific version so that we have rooms to fix stuff without needing to re-release.  Also, this is just a test run for the dependency. Only bleeding-edge developers will be using `numba-rvsdg` and they can manually pin versions if needed.

esc: we should pin to `0.0.5` once it's released via https://github.com/numba/numba-rvsdg/pull/105

esc (approved):
@kc611 and I pair reviewed this using the following procedure:

* build the package based on the recipe
* install package in test environment
* run NUMBA_USE_RVSDG_FRONTEND=1 python frontend-test.py to observe the guard being triggered
* install `-c numba numba-rvsdg`
* run  NUMBA_USE_RVSDG_FRONTEND=1 python frontend-test.py  and observe it succeed 

```
from numba import njit


@njit
def sum(a):
    c = 0
    for i in range(7):
        c += i
    return c


print(sum(7))
```

Issue #1269, opened 2015-06-25
Support for axis arguments on reduction functions


seibert:
Now that we support array expressions, I'm finding that I really want to be able to pass the `axis` argument to reduction functions like `np.sum()`, `np.mean()`, etc.  This raises the issue again of how to do dispatch based on argument values, but it is becoming much more critical now that I can so easily create arrays to reduce in nopython mode.

pitrou:
Well as long as we support only a couple signatures (e.g. `np.sum(a)` and `np.sum(a, axis=...)`), then it shouldn't be a problem. It's when there's some combinatorial explosion in argument signatures that our current approach doesn't work.

Another question: can specifying the axis change the return type?

seibert:
If you mean "can the dtype change if axis is specified?", I don't think so.  There are still the handful of reduction functions (like `np.mean`) that can return a different dtype, but that is independent of whether `axis` is set.

One aspect of `axis` that I didn't appreciate is that it can take a tuple of integers to sum multiple axes at once (starting with Numpy 1.7).  I don't think we need to support that in the first pass implementation of this feature since that usage is still pretty rare.

jorgehatccrma:
Is there a workaround to pass `axis` argument to a reduction function like `np.sum()`? I'm currently getting the following error:

```
...
Failed at nopython (nopython frontend)
Internal error at <numba.typeinfer.CallConstraint object at 0x127adfad0>:
--%<-----------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/numba/typeinfer.py", line 128, in propagate
    constraint(typeinfer)
  File "/usr/local/lib/python2.7/site-packages/numba/typeinfer.py", line 373, in __call__
    self.resolve(typeinfer, typevars, fnty)
  File "/usr/local/Cellar/python/2.7.12_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/contextlib.py", line 35, in __exit__
    self.gen.throw(type, value, traceback)
  File "/usr/local/lib/python2.7/site-packages/numba/errors.py", line 249, in new_error_context
    six.reraise(type(newerr), newerr, sys.exc_info()[2])
  File "/usr/local/lib/python2.7/site-packages/numba/errors.py", line 243, in new_error_context
    yield
  File "/usr/local/lib/python2.7/site-packages/numba/typeinfer.py", line 373, in __call__
    self.resolve(typeinfer, typevars, fnty)
[...]

PR #9123, opened 2023-08-07
Implement most ufunc attributes and ufunc.reduce

guilhermeleobas:
This PR exposes most ufunc attributes and implements `ufunc.reduce` for ufuncs created with `@vectorize`

- [x] ufunc.nin
- [x] ufunc.nout
- [x] ufunc.nargs
- [ ] ufunc.ntypes
- [ ] ufunc.types
- [x] ufunc.identity
- [x] ufunc.signature (always `None`)

Implementing ufunc.ntypes or ufunc.types is not possible as Numba needs to call the NumPy ufunc attribute at runtime. However, this is not possible as NumPy doesn't allow creating weak references to an ufunc, which is required by `types.BoundFunction`.

------------

## To-do:
- [ ] Check how Cython implements ufunc.reduce
- [ ] Raise an exception if ufunc doesn't have an identity.

------------

Reduce is implemented for the following set of arguments:

```python
def reduce(array, axis=0, dtype=None, initial=None):
    ...
```
keeping https://github.com/numba/numba/issues/1269#issuecomment-1404864085 as a reference here.

------------

Code I used to benchmark:

```python
from numba import vectorize, njit
import numpy as np
import time

@vectorize(identity=0)
def add(a, b):
    return a + b


@njit
def bar(a, axis):
    return add.reduce(a, axis=axis)


def measure_time(fn, *args):
    start = time.time()
    fn(*args)
    end = time.time()
    print(f'{fn.__name__}: {end - start}s')


N = 1_000_000_000
T = (10,) * int(np.log10(N))
a = np.arange(N).reshape(T)
axis = tuple(range(a.ndim))

# jit compile `bar`
bar(a, axis)

# measure time
measure_time(bar, a, axis)
measure_time(bar.py_func, a, axis)
```


guilhermeleobas:
@kc611, could you please take a look at this PR once you have a chance? I'd like to discuss possible ways to speedup ufunc.reduce

numba/np/ufunc/dufunc.py
```diff
@@ -231,6 +242,260 @@ def _compile_for_argtys(self, argtys, return_type=None):
...
+                # is (X, Y).
+                # Another way is to compute the result index using strides,
+                # which is faster than manipulating tuples.
+                view = r.ravel()
+                for idx, val in np.ndenumerate(array):
+                    flat_pos = compute_flat_idx(r.strides, r.itemsize, idx,
+                                                axis)
+                    lhs, rhs = view[flat_pos], val
+                    view[flat_pos] = ufunc(lhs, rhs)
+                return r
```
mattip, line 463: This works, but I think the re-computation of flat_pos from basics at each iteration is going to be a bit slow. I would think you could avoid `nd.enumerate` and have a loop that only increments the flat_pos.

I wonder how the non-reduce ufuncs do this indexing, since they too can accept an axis argument.

mattip (commented):
Cool. It would be nice if numba could do reduce ufuncs.

numba/np/ufunc/dufunc.py
```diff
@@ -231,6 +242,262 @@ def _compile_for_argtys(self, argtys, return_type=None):
...
+                        idx, e = i, tup[i]
+                return idx, e
+
+            def impl_1d(ufunc, array, axis=0, dtype=None, initial=None):
+                if init_none and id_none:
+                    r = array[0]
+                elif init_none:
+                    r = identity
+                else:
+                    r = initial
```
guilhermeleobas, line 417: Should this be the other way around? initial value first

numba/np/ufunc/dufunc.py
```diff
@@ -231,6 +242,262 @@ def _compile_for_argtys(self, argtys, return_type=None):
...
+                    for idx, _ in np.ndenumerate(r):
+                        # shape[0:axis] + 0 + shape[axis:]
+                        result_idx = tuple_slice_append(idx, axis, 0)
+                        r[idx] = array[result_idx]
+                elif initial is None and identity is not None:
+                    # Checking if identity is not none is redundant but required
+                    # compile this block
+                    r = np.full(shape, fill_value=identity, dtype=nb_dtype)
+                else:
+                    r = np.full(shape, fill_value=initial, dtype=nb_dtype)
```
guilhermeleobas, line 451: initial case should come first!?

numba/np/ufunc/dufunc.py
```diff
@@ -231,6 +242,262 @@ def _compile_for_argtys(self, argtys, return_type=None):
...
+                    lhs, rhs = view[flat_pos], val
+                    view[flat_pos] = ufunc(lhs, rhs)
+                return r
+
+            def impl_nd_axis_tuple(ufunc,
+                                   array,
+                                   axis=0,
+                                   dtype=None,
+                                   initial=None):
+                min_idx, min_elem = find_min(axis)
```
guilhermeleobas, line 472: Does NumPy selects the smallest index first?

numba/np/ufunc/dufunc.py
```diff
@@ -231,6 +242,241 @@ def _compile_for_argtys(self, argtys, return_type=None):
...
+            def impl(ufunc):
+                val = getattr(ufunc.key[0], attr)
+                return lambda ufunc: val
+            return impl
+
+        # ntypes/types needs "at" to be a BoundFunction rather than a Function
+        # But this fails as it cannot a weak reference to an ufunc due to NumPy
+        # not setting the "tp_weaklistoffset" field. See:
+        # https://github.com/numpy/numpy/blob/7fc72776b972bfbfdb909e4b15feb0308cf8adba/numpy/core/src/umath/ufunc_object.c#L6968-L6983  # noqa: E501
+
```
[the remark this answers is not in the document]
guilhermeleobas: When attempting to create a weak reference to an ufunc, CPython raises an error:
```python
>>> import weakref
>>> import numpy as np
>>> weakref.ref(np.minimum)
TypeError: cannot create weak reference to 'numpy.ufunc' object
```

numba/np/ufunc/dufunc.py
```diff
@@ -231,6 +242,262 @@ def _compile_for_argtys(self, argtys, return_type=None):
...
+                        s = tuple_setitem(s, j, val)
+                    else:
+                        e = tup[i]
+                        s = tuple_setitem(s, j, e)
+                        i += 1
+                    j += 1
+                return s
+
+            @intrinsic
+            def compute_flat_idx__(typingctx, strides, itemsize, idx, axis):
```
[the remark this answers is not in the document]
guilhermeleobas: `compute_flat_idx__` implements the same code as `compute_flat_idx` but it is just a bit switch table where each case is computed by hand beforehand. In short, it looks something like this:

```python
def compute_flat_idx(axis, ...):
    if axis == 0:
        # compute flat idx with axis == 0
    else axis == 1:
        # compute flat idx with axis == 1
    ...
```

numba/np/ufunc/dufunc.py
guilhermeleobas: Actually, no! Code seems fine like this

numba/tests/npyufunc/test_dufunc.py
```diff
@@ -85,6 +96,125 @@ def test_ufunc_props(self):
...
+                def impl():
+                    return duadd.{attr}
+            '''
+            l = {}
+            exec(textwrap.dedent(fn), {'duadd': duadd}, l)
+            return l['impl']
+
+        for attr, val in attributes.items():
+            cfunc = njit(get_attr_fn(attr))
+            self.assertEqual(val, cfunc(), f'attribute: {attr}')
```
kc611, line 124:
```suggestion
            self.assertEqual(val, cfunc(), f'Attribute differs from original: {attr}')
```

numba/tests/npyufunc/test_dufunc.py
```diff
@@ -85,6 +96,125 @@ def test_ufunc_props(self):
...
+                got = foo(array, axis, dtype, initial)
+                self.assertPreciseEqual(expected, got)
+
+    def _check_reduce_axis(self, ufunc, dtype, initial=None):
+
+        @njit
+        def foo(a, axis):
+            return ufunc.reduce(a, axis=axis, initial=initial)
+
+        exc_msg = (f"reduction operation '{ufunc.__name__}' is not "
```
kc611, line 161:
```suggestion
        exc_msg = (f"Reduction operation '{ufunc.__name__}' is not "
```

numba/np/ufunc/dufunc.py
```diff
@@ -231,6 +242,264 @@ def _compile_for_argtys(self, argtys, return_type=None):
...
+                raise errors.NumbaTypeError(msg)
+
+            axis_int = isinstance(axis, types.Integer)
+            axis_int_tuple = isinstance(axis, types.UniTuple) and \
+                isinstance(axis.dtype, types.Integer)
+            axis_tuple_size = len(axis) if axis_int_tuple else 0
+
+            if self.ufunc.identity is None and not (
+                    axis_int or (axis_int_tuple and axis_tuple_size == 1)):
+                msg = (f"reduction operation '{self.ufunc.__name__}' is not "
```
kc611, line 285:
```suggestion
                msg = (f"Reduction operation '{self.ufunc.__name__}' is not "
```

numba/tests/npyufunc/test_dufunc.py
```diff
@@ -85,6 +96,125 @@ def test_ufunc_props(self):
...
+                for axis in itertools.combinations(range(array.ndim), r=i):
+                    try:
+                        expected = foo.py_func(array, axis)
+                    except ValueError as e:
+                        self.assertEqual(e.args[0], exc_msg)
+                        with self.assertRaisesRegex(TypingError, exc_msg):
+                            got = foo(array, axis)
+                    else:
+                        got = foo(array, axis)
+                        self.assertPreciseEqual(expected, got)
```
kc611, line 178: Not sure if this is the correct way for error testing. We can just check error throwing once in a different function, the rest of the cases here can be simply all valid cases.

numba/np/ufunc/dufunc.py
```diff
@@ -231,6 +242,264 @@ def _compile_for_argtys(self, argtys, return_type=None):
...
+        @overload_method(at, 'reduce')
+        def ol_reduce(ufunc, array, axis=0, dtype=None, initial=None):
+            if not isinstance(array, types.Array):
+                msg = 'The first argument "array" must be array-like'
+                raise errors.NumbaTypeError(msg)
+
+            axis_int = isinstance(axis, types.Integer)
+            axis_int_tuple = isinstance(axis, types.UniTuple) and \
+                isinstance(axis.dtype, types.Integer)
+            axis_tuple_size = len(axis) if axis_int_tuple else 0
```
kc611, line 281: Might be worth adding corner cases like `axis=None` (reduce over all axes) and `axis=()` (reduce over nothing, but this had a different type signature i.e. `types.Tuple` with zero size).

numba/np/ufunc/dufunc.py
```diff
@@ -231,6 +242,264 @@ def _compile_for_argtys(self, argtys, return_type=None):
...
+                        lhs, rhs = view[flat_pos], val
+                        view[flat_pos] = ufunc(lhs, rhs)
+                return r
+
+            def impl_nd_axis_tuple(ufunc,
+                                   array,
+                                   axis=0,
+                                   dtype=None,
+                                   initial=None):
+                min_idx, min_elem = find_min(axis)
```
kc611, line 474: Will this affect non associative (with respect to axes) ufuncs vectorizations ?

numba/np/ufunc/dufunc.py
```diff
@@ -231,6 +242,264 @@ def _compile_for_argtys(self, argtys, return_type=None):
...
+                else:
+                    for idx, val in np.ndenumerate(array):
+                        if initial is None and identity is None and \
+                                idx[axis] == 0:
+                            continue
+                        flat_pos = compute_flat_idx(r.strides, r.itemsize,
+                                                    idx, axis)
+                        lhs, rhs = view[flat_pos], val
+                        view[flat_pos] = ufunc(lhs, rhs)
+                return r
```
kc611, line 467: The removal of use of `ndenumerate` here as a possible optimisation, should be opened as a separate issue.

kc611 (changes requested):

kc611:
@guilhermeleobas Thanks for the patch. This PR mostly looks good. Some nitpicks and questions above before we can mark this as RTM.

numba/tests/npyufunc/test_dufunc.py
guilhermeleobas: [NumPy](https://github.com/numpy/numpy/blob/cba186917f5b07a996c42886603ca808b857fa34/numpy/core/src/umath/reduction.c#L208-L209) uses "reduction" here

numba/np/ufunc/dufunc.py
guilhermeleobas: See above

numba/np/ufunc/dufunc.py
guilhermeleobas: Done

numba/tests/npyufunc/test_dufunc.py
guilhermeleobas: Problem is that some of the combinations generated by `itertools.combinations` are invalid, and it is not possible to know that beforehand

guilhermeleobas:
CI failure doesn't seem to be related to this PR.

kc611 (approved):
LGTM!

gmarkall:
gpuci run tests

gmarkall:
@kc611 Is this good to be RTM (assuming it passes the gpuCI run I just started)?

kc611:
I think it's good to go. (Unless someone objects)

gmarkall:
gpuCI failed one CUDA version for some kind of Conda error. I expected if this was going to break CUDA, it'd break for all configurations, so I'm happy to ignore that.

guilhermeleobas:
Thanks for the review, folks. I'll create the issues to track future work:
* Rewrite `ufunc.reduce` to not use `np.ndenumerate`
* Remove experimental feature warning in 2(?) Numba versions?
* Support `ufunc.ntypes` and `ufunc.types` attributes

Also, if it is ok, I'd like another commit that updates documentation and adds a test.

PR #9126, opened 2023-08-10
Add support for np.indices()

KrisMinchev:
This PR improves on #4074 by adding support for `np.indices()`. The current implementation supports the first argument only.


numba/np/arrayobj.py
```diff
@@ -4562,6 +4562,33 @@ def diag_impl(v, k=0):
         return diag_impl
 
 
+@overload(np.indices)
+def numpy_indices(dimensions):
+    if not isinstance(dimensions, types.Sequence):
+        msg = 'The argument "dimensions" must be a sequence of integers'
+        raise errors.TypingError(msg)
```
guilhermeleobas, line 4569: If you need to compute `len(dimensions)`, then, it cannot be an instance of `types.Sequence` as calling `np.indices([2, 3])` inside @jit will fail:

```python
...
  File "/Users/guilhermeleobas/git/numba/numba/core/typing/templates.py", line 614, in generic
    disp, new_args = self._get_impl(args, kws)
  File "/Users/guilhermeleobas/git/numba/numba/core/typing/templates.py", line 713, in _get_impl
    impl, args = self._build_impl(cache_key, args, kws)
  File "/Users/guilhermeleobas/git/numba/numba/core/typing/templates.py", line 786, in _build_impl
    ovf_result = self._overload_func(*args, **kws)
  File "/Users/guilhermeleobas/git/numba/numba/np/arrayobj.py", line 4575, in numpy_indices
    N = len(dimensions)
TypeError: object of type 'List' has no len()
```

guilhermeleobas (changes requested):
Thanks, @KrisMinchev. Only have one request, but otherwise PR looks good.

numba/np/arrayobj.py
KrisMinchev: Hi @guilhermeleobas, thank you for spotting this. The NumPy function supports all sequence inputs, but since the type of the output depends on the length of the input, the Numba version cannot support list inputs. I might change the function to only support `types.BaseTuple`, if that sounds good to you as well.

guilhermeleobas: Yes, unfortunately I think Numba can only support this function if the input is a tuple.

KrisMinchev:
Hi @guilhermeleobas, I have addressed your comment and I have changed the function to accept tuple inputs only.

guilhermeleobas (approved):

guilhermeleobas:
Hi @KrisMinchev, there are now conflicts in this PR. Could you address them before we merge it?

guilhermeleobas (approved):
LGTM. Thanks @KrisMinchev

karthikkurella (approved):

guilhermeleobas:
@sklam, could we merge this one?

PR #9138, opened 2023-08-16
Release Actions 0.58

esc:
As title

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,6 +23,7 @@ min_python_version = "3.8"
 max_python_version = "3.12"  # exclusive
 min_numpy_build_version = "1.11"
 min_numpy_run_version = "1.21"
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -23,6 +23,7 @@
 max_python_version = "3.12"  # exclusive
 min_numpy_build_version = "1.11"
 min_numpy_run_version = "1.21"
+max_numpy_run_version = "1.25"
```
kc611, line 26:
```suggestion
max_numpy_run_version = "1.26"
```

Since we're checking Numpy to be strictly less that max version

kc611 (changes requested):

setup.py
esc: done in fb03134

PR #9144, opened 2023-08-21
Fix error when literal is wrongly propagated in a PHI node

### LINKED ISSUE #9125: `isinstance` check produces wrong output
Mutating a variable inside an `isinstance` produces the wrong output

```python
from numba import njit

@njit
def foo(a):
    x = 0
    if isinstance(a, int):
        # renaming variable fixes the issue: "b = (a, a)"
        a = (a, a)

    for i in range(len(a)):
        x += i
    return x

a = 3
print(foo(a))  # 0
print(foo.py_func(a))  # 1
```

------------

*  Using a while loop fixes the issue
```python
@njit
def foo(a):
    x = 0
    if isinstance(a, int):
        a = (a, a)
    i = 0
    while i < len(a):
        x += i
        i += 1
    return x

a = 3
print(foo(a))  # 1
print(foo.py_func(a))  # 1
```

------------

* Renaming `b = (a, a)` fixes it as well

```python
@njit
def foo(a):
    x = 0
    if isinstance(a, int):
        b = (a, a)
    for i in range(len(b)):
        x += i
    return x

a = 3
print(foo(a))  # 1
print(foo.py_func(a))  # 1
```

guilhermeleobas:
Fix numba issue #9125 - Don't propagate literal if node is a PHI node.

Reproducer:

```python
from numba import njit

@njit
def foo(a):
    x = 0
    if isinstance(a, int):
        # renaming variable fixes the issue: "b = (a, a)"
        a = (a, a)

    for i in range(len(a)):
        x += i
    return x

a = 3
print(foo(a))  # 0
print(foo.py_func(a))  # 1
```

Closes #9125

### DIFF
--- a/numba/core/untyped_passes.py
+++ b/numba/core/untyped_passes.py
@@ -1577,8 +1577,10 @@ class PropagateLiterals(FunctionPass):
                 # Only propagate a PHI node if all arguments are the same
                 # constant
                 if isinstance(value, ir.Expr) and value.op == 'phi':
+                    # typemap will return None in case `inc.name` not in typemap
                     v = [typemap.get(inc.name) for inc in value.incoming_values]
-                    if any([v[0] != vi for vi in v]):
+                    # stop if the elements in `v` do not hold the same value
+                    if v[0] is not None and any([v[0] != vi for vi in v]):
                         continue
 
                 lit = typemap.get(target.name, None)
--- a/numba/core/untyped_passes.py
+++ b/numba/core/untyped_passes.py
@@ -1574,6 +1574,13 @@ class PropagateLiterals(FunctionPass):
                                    'due to a branch.')
                             raise errors.NumbaTypeError(msg, loc=assign.loc)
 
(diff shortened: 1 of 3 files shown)


numba/core/untyped_passes.py
```diff
@@ -1574,6 +1574,13 @@ def run_pass(self, state):
                                    'due to a branch.')
                             raise errors.NumbaTypeError(msg, loc=assign.loc)
 
+                # Only propagate a PHI node if all arguments are the same
+                # constant
+                if isinstance(value, ir.Expr) and value.op == 'phi':
+                    v = [typemap.get(inc.name) for inc in value.incoming_values]
```
stuartarchibald, line 1580: What happens if the `inc.name` isn't in the `typemap`?

stuartarchibald (commented):
Thanks for the patch, I think this will fix the issue described. There's a minor comment to resolve to do with a var name not being in the partial type-map, other than that looks good.

stuartarchibald (approved):

PR #9154, opened 2023-08-23
Add support for np.unwrap()

KrisMinchev:
This PR improves on #4074  by implementing the function `np.unwrap()`.

### DIFF
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -4454,7 +4454,11 @@ def np_asarray_chkfinite(a, dtype=None):
 
 
 @overload(np.unwrap)
-def numpy_unwrap(p, discont=None, period=6.283185307179586):
+def numpy_unwrap(p, discont=None, axis=-1, period=6.283185307179586):
+    if not isinstance(axis, (int, types.Integer)):
+        msg = 'The argument "axis" must be an integer'
+        raise TypingError(msg)
+
     if not type_can_asarray(p):
         msg = 'The argument "p" must be array-like'
         raise TypingError(msg)
@@ -4474,7 +4478,10 @@ def numpy_unwrap(p, discont=None, period=6.283185307179586):
     dtype = np.result_type(as_dtype(p.dtype), as_dtype(period))
     integer_input = np.issubdtype(dtype, np.integer)
 
-    def impl(p, discont=None, period=6.283185307179586):
+    def impl(p, discont=None, axis=-1, period=6.283185307179586):
+        if axis != -1:
+            msg = 'Value for argument "axis" is not supported'
+            raise NumbaNotImplementedError(msg)
         # Flatten to a 2D array, keeping axis -1
         p_init = np.asarray(p).astype(dtype)
         init_shape = p_init.shape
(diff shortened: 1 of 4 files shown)


numba/np/arraymath.py
```diff
@@ -4452,6 +4452,75 @@ def impl(a, dtype=None):
...
+
+    if not isinstance(discont, (types.Integer, types.Float, types.NoneType)):
+        msg = 'The argument "discont" must be a scalar'
+        raise TypingError(msg)
+
+    if not isinstance(period, (float, types.Number)):
+        msg = 'The argument "period" must be a scalar'
+        raise TypingError(msg)
+
+    axis = -1
```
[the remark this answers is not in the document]
KrisMinchev: Hi @guilhermeleobas, I think that if the user wants to set `axis = -1` manually, such a check will result in an exception, which might not be the best way to handle that case.

numba/tests/test_np_functions.py
```diff
@@ -441,6 +440,18 @@ def unwrap(p, discont=None, axis=-1, period=6.283185307179586):
     return np.unwrap(p, discont, axis, period=period)
 
 
+def unwrap1(p):
```
guilhermeleobas, line 443: Just out of curiosity, is there any reasoning for the function naming convention?
* unwrap1
* unwrap13
* unwrap123

numba/np/arraymath.py
```diff
@@ -4463,25 +4463,27 @@ def numpy_unwrap(p, discont=None, axis=-1, period=6.283185307179586):
         msg = 'The argument "p" must be array-like'
         raise TypingError(msg)
 
-    if not isinstance(discont, (types.Integer, types.Float, types.NoneType)):
+    if (not isinstance(discont, (types.Integer, types.Float, types.NoneType))
+            and (discont is not None)):
```
guilhermeleobas, line 4467: Use `cgutils.is_nonelike(discont)` to check if it is `None`

guilhermeleobas (changes requested):
Just one last request and it should be good to merge

numba/tests/test_np_functions.py
KrisMinchev: This naming convention helps me keep track of which of `unwrap`'s arguments it needs as input(only first, first and third, etc.).

numba/np/arraymath.py
```diff
@@ -4464,7 +4464,7 @@ def numpy_unwrap(p, discont=None, axis=-1, period=6.283185307179586):
         raise TypingError(msg)
 
     if (not isinstance(discont, (types.Integer, types.Float, types.NoneType))
-            and (discont is not None)):
+            and not cgutils.is_nonelike(discont)):
```
guilhermeleobas, line 4467: `cgutils.is_nonelike(discont)` already checks if discont is an instance of `types.NoneType`.

guilhermeleobas (approved):
LGTM. Thanks @KrisMinchev

KrisMinchev:
Hi @guilhermeleobas, thank you for your review. I have refactored the implementation of `np.unwrap` accordingly.

guilhermeleobas (approved):
Thanks @KrisMinchev

PR #9155, opened 2023-08-23
Remove unused test.cmd

sklam:
`test.cmd` is not referenced anywhere. Let's remove it.

The content of the file is inlined into [azure-windows.yml](https://github.com/numba/numba/blob/bd7ebcfd4b850208b627a3f75d4706000be36275/buildscripts/azure/azure-windows.yml) nowadays


stuartarchibald (approved):
Thanks for the patch. This does indeed appear unused.

PR #9159, opened 2023-08-24
Fix uncaught exception in `find_file()`

gmarkall:
(As discussed in the triage meeting on Tuesday)

`find_file()` assumed that all folders it was asked to search actually exist - this may not be the case, and is exposed now that we search more locations for CUDA packages. To prevent uncaught exceptions propagating outward (and thereby crashing `numba -s`, `cuda.cudadrv.libs.test()`, etc.), we catch `FileNotFoundError` when listing directories, and continue with the next dir when this occurs.

There didn't seem to be any tests of `findlib`, so I've added a new test file / class with this particular regression.

### DIFF
--- a/numba/misc/findlib.py
+++ b/numba/misc/findlib.py
@@ -53,7 +53,10 @@ def find_file(pat, libdir=None):
         libdirs = list(libdir)
     files = []
     for ldir in libdirs:
-        entries = os.listdir(ldir)
+        try:
+            entries = os.listdir(ldir)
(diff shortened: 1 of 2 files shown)


gmarkall:
I've added a release note skip, because I'm not sure this manifested in prior versions of Numba anyway, so there shouldn't be any notable change.

sklam (approved):

sklam:
Thanks for the patch!

PR #9161, opened 2023-08-24
allow casting `optional` type to numba concrete types

dlee992:
You can see the added test case.
Without this patch, the test will fail. With this patch, we allow this kind of casting from optional to concrete_type, e.g., `types.int64(optional_instance)`.

### DIFF
--- a/numba/core/typing/builtins.py
+++ b/numba/core/typing/builtins.py
@@ -825,7 +825,10 @@ class NumberClassAttribute(AttributeTemplate):
                 sig = fnty.get_call_type(self.context, (val, types.DType(ty)),
                                          {})
                 return sig.return_type
-            elif isinstance(val, (types.Number, types.Boolean, types.IntEnumMember)):
+            elif isinstance(val, (types.Number,
+                                  types.Boolean,
+                                  types.IntEnumMember,
+                                  types.Optional)):
                  # Scalar constructor, e.g. np.int32(42)
                  return ty
             elif isinstance(val, (types.NPDatetime, types.NPTimedelta)):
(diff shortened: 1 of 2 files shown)


dlee992:
I think this's a simple patch, tests are all passed. But how to fix the towncrier? @esc

dlee992:
Hi, @sklam . Perhaps this PR can go into 0.58.0rc2 or 0.58.0? I think you can give this a quick review, only change one line source code and add a test for it.

numba/core/typing/builtins.py
```diff
@@ -825,7 +825,10 @@ def typer(val):
                 sig = fnty.get_call_type(self.context, (val, types.DType(ty)),
                                          {})
                 return sig.return_type
-            elif isinstance(val, (types.Number, types.Boolean, types.IntEnumMember)):
+            elif isinstance(val, (types.Number,
+                                  types.Boolean,
+                                  types.IntEnumMember,
+                                  types.Optional)):
```
gmarkall, line 831: Does this permit something that is optional but not like a number?

dlee992: In current mainline, we have this, so I think we allow casting optional to any type in the lowering level?
https://github.com/numba/numba/blob/d70830f2262c2a911a3eb6e81781fe070a2e8891/numba/core/optional.py#L112

dlee992: This PR will let typeinfer also allow casting optional to any type in the typing level. but I could be wrong in some unknown aspect..

esc:
> > Perhaps this PR can go into 0.58.0rc2 or 0.58.0?
> 
> RC2 will be limited to regressions or critical security bugs. This patch will go towards 0.59.

Added to the milestone for 0.59

sklam:
I made an alternative PR for this at https://github.com/numba/numba/pull/9248
Closing.

PR #9168, opened 2023-08-28
fix the `get_template_info` method in `overload_method` template

### LINKED ISSUE #9162: `template.path` doesn't point to source code location in `overload_method`
when testing with `overload_method`, it turns out that the type template of `overload_method` has a wrong implementation in `get_source_code_info`, since the attribute `path` in it will point to `numba/core/typing/templates.py`, rather than the real source code file path which the `overload_method`-ed function is defined.

Need to give you a reproducer later..

And it's related to proceed the cache PR https://github.com/numba/numba/pull/8396, I think.

dlee992:
fix #9162


dlee992:
maybe also need to fix the same issue for `overload_attribute` and `bound_function`. not sure about these.

dlee992:
`_OverloadAttributeTemplate._resolve()` method directly returns `sig.return_type`. I feel this is different with `types.BoundFunction(MethodTemplate, typ)` in `_OverloadMethodTemplate._resolve()` method. Hard to apply the same fix, so ignore `overload_attribute` for now.

dlee992:
The new test case is copy-and-modify from #8396

dlee992:
Ready for review now! @sklam

dlee992:
Hi, @gmarkall . I think this's ready for review, and I already added a test for this.

dlee992:
If https://github.com/numba/numba/pull/8396 is included in 0.59rc, then this PR should also be included. That PR needs this one, I think.

gmarkall (approved):
I think this looks good and appears to resolve #9162 correctly. The `info` for the template before the fix is:

```
{'kind': 'overload',
 'name': '_OverloadMethodTemplate._resolve.<locals>.MethodTemplate.generic',
 'sig': '(_, args, kws)',
 'filename': 'numba/core/typing/templates.py',
 'lines': (1136, 1144),
 'docstring': None}
```

(everything but the docstring is wrong, and it's just a coincidence that it's correct :-) )

After the fix, it's:

```
{'kind': 'overload_method',
 'name': 'TestMiscIssues.test_issue_9162.<locals>.ol_aabbcc',
 'sig': '(self)',
 'filename': 'numba/tests/test_typeinfer.py',
 'lines': (712, 718),
 'docstring': None}
```

(all correct)

sklam:
I pushed a commit to resolve merge conflict

gmarkall (approved):
Re-approving following the merge (just waiting for CI to move back to RTM).

PR #9170, opened 2023-08-29
Fix an inappropriate test expression to remove a logical short circuit

munahaf:
In file: parfor.py, method: `compute_graph_info`, a logical expression uses the identity operator. A new object is created inside the identity check operation and then used for matching identity. Since this is a distinct, new object, it will not have identity and it will match with anything else. As a result, the identity check will have a logical short circuit and the program may have unintended behavior. In this case, the test expression will always evaluate to false. I suggested that the logical operation be done correctly. 

Sponsorship and Support:

### DIFF
--- a/numba/parfors/parfor.py
+++ b/numba/parfors/parfor.py
@@ -805,7 +805,7 @@ class ParforDiagnostics(object):
             val = a.get(x)
             if val is not None:
                 a[x] = val
-            elif val is []:
+            elif val == []:
                 not_roots.add(x) # debug only
             else:
                 a[x] = []


gmarkall (approved):
I looked at this earlier and the change makes sense to me. I presume it doesn't expose some other latent bug since the tests all pass. I think this would be OK to mark RTM as long as it looks OK to @DrTodd13.

DrTodd13 (approved):
Needs towncrier update but otherwise I think this should be okay.

gmarkall:
I added the `skip_release_notes` label to prevent the check running (this doesn't seem worth a whole changelog entry) but it ran anyway.

sarva-karthik (approved):

PR #1460, opened 2015-10-05
Refactor pycc API

pitrou:
This is a proposal for changing the `pycc` API. Currently you use some decorators together with an external script that you have to run on the command line. The proposed API looks more like the `cffi` "out-of-line" API (https://cffi.readthedocs.org/en/latest/cdef.html), that is:
- uses a dedicated instantiable class rather than a hidden global registry
- exposes configuration values (output directory, etc.) as properties or constructor arguments
- exposes a programmatic compilation API (just a `compile()` method)

This will let people build their toolchain however they want (they could for example build two different extensions from a single source file, or the converse). Also we could later build some distutils integration on top of that.

This PR is keeping the old API as deprecated. Honestly I don't know if anyone is using it, since it's not documented, and currently you can't do much in a pycc-compiled thing anyway.

By the way, even this new API should be kept experimental until we know clearly what we want to do with pycc.

PR #9172, opened 2023-08-29
Remove an undefined name

munahaf:
In file: compiler.py, the list named `__all__` contains an undefined name which can result in errors  when this module is imported. I removed the undefined name from the list. For more information regarding `__all__`, please read about [importing fields from a package](https://docs.python.org/3/tutorial/modules.html#importing-from-a-package).  

Sponsorship and Support:

### DIFF
--- a/numba/pycc/compiler.py
+++ b/numba/pycc/compiler.py
@@ -19,7 +19,7 @@ from numba.core import cgutils
 
 logger = logging.getLogger(__name__)
 
-__all__ = ['Compiler']
+__all__ = []
 
 NULL = ir.Constant(lt._void_star, None)
 ZERO = ir.Constant(lt._int32, 0)


munahaf:
This PR (https://github.com/numba/numba/pull/1452/files) shows that once `Compiler`  was defined. 

```
Compiler = CompilerPy3 if IS_PY3 else CompilerPy2
```

Later the import was deleted but the developer forgot to remove it from `__all__`.

sarva-karthik (approved):

numba/pycc/compiler.py
```diff
@@ -19,7 +19,7 @@
 
 logger = logging.getLogger(__name__)
 
-__all__ = ['Compiler']
+__all__ = []
```
kc611: During refactoring of the `pycc` API (see https://github.com/numba/numba/pull/1460) the name `Compiler` was changed to `ModuleCompiler`. So it can't be completely removed from the `__all__`, but instead should be renamed to `ModuleCompiler` as follows: 

```suggestion
__all__ = ['ModuleCompiler']
```

esc:
FWIW: here is a concrete reproducer:

```
In [1]: from numba.pycc.compiler import *
Cell In[1], line 1
----> 1 from numba.pycc.compiler import *

AttributeError: module 'numba.pycc.compiler' has no attribute 'Compiler'
```

esc:
@munahaf I have reviewed this PR, and I can confirm (with a merge of current `main`) that this PR fixes the issue:

```
In [1]: from numba.pycc.compiler import *

In [2]:
```

esc:
@munahaf can you rebase this onto current `main`? That should run the current CI tests and confirm that this PR doesn't break anything else. If that's fine, we can merge it.

numba/pycc/compiler.py
munahaf: ACK. Making the change.

PR #9184, opened 2023-09-06
Update changelog for 0.58.0rc2

sklam:
as titled


sklam:
just realize this should go into cherrypick PR as per release checklist

PR #9189, opened 2023-09-09
Align `searchsorted` behaviour with numpy

rjenc29:
This MR attempts to align `searchsorted` behaviour with numpy for a wider set of use cases, particularly where (i) the input array `a` is not properly sorted or (ii) the input array `a` is properly sorted but contains NaN and side='right'.

This PR was motivated by https://github.com/numba/numba/pull/9169 - the hypothesis being that if we could align `searchsorted` behaviour with numpy closely (including for weird inputs) then updating `digitize` should be trivial, since its implementation could in fact just hand off to `searchsorted` (in the same way that numpy does). 

(The aspiration to cope with improperly sorted inputs comes in part from the foreknowledge that `digitize` attempts to determine if the provided bins are monotonic increasing or decreasing before calling into `searchsorted`, and  can make seemingly incorrect inference in the presence of NaNs.)

I have added thousands of tests to assert behavioural similarity (happy to tone this down) and have run hours of randomised tests on GCP. 

It could be that there are some outstanding edge cases where the behaviour continues to be misaligned; the implementation choices I made are difficult to justify / rationalise, and there's an unfortunate / unexpected change in behaviour pre- and post- numpy 1.22, which appears to be an unintended consequence of some bugfix (as noted in the comments).

Interested in your thoughts.

### DIFF
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -3749,7 +3749,8 @@ def searchsorted(a, v, side='left'):
     if side_val == 'left':
         _impl = _searchsorted_left
     elif side_val == 'right':
-        # change in behaviour introduced by https://github.com/numpy/numpy/pull/21867
+        # change in behaviour introduced by
+        # https://github.com/numpy/numpy/pull/21867
         if numpy_version >= (1, 23):
             _impl = _searchsorted_right_np123_on
         else:
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -15,7 +15,7 @@ from numba.core import types, cgutils
 from numba.core.extending import overload, overload_method, register_jitable
 from numba.np.numpy_support import (as_dtype, type_can_asarray, type_is_scalar,
                                     numpy_version, is_nonelike,
-                                    check_is_integer)
+                                    check_is_integer, lt_complex)
 from numba.core.imputils import (lower_builtin, impl_ret_borrowed,
                                  impl_ret_new_ref, impl_ret_untracked)
 from numba.np.arrayobj import (make_array, load_item, store_item,
@@ -3675,6 +3675,11 @@ def np_bincount(a, weights=None, minlength=0):
 
 
 @register_jitable
+def less_than_or_equal(a, b):
+    return a <= b
+
+
+@register_jitable
 def custom_lt(a, b):
     if np.isnan(b):
         return not np.isnan(a)
@@ -3688,38 +3693,7 @@ def custom_le(a, b):
     return a <= b
 
 
-@register_jitable
-def custom_complex_lt(a, b):
-    if np.isnan(a.real):
-        if np.isnan(b.real):
-            if np.isnan(a.imag):
-                return False
-            else:
-                if np.isnan(b.imag):
-                    return True
-                else:
-                    return a.imag < b.imag
-        else:
-            return False
-
-    else:
-        if np.isnan(b.real):
-            return True
-        else:
-            if np.isnan(a.imag):
-                if np.isnan(b.imag):
-                    return a.real < b.real
-                else:
-                    return False
-            else:
-                if np.isnan(b.imag):
-                    return True
-                else:
-                    if a.real < b.real:
-                        return True
-                    elif a.real == b.real:
-                        return a.imag < b.imag
-                    return False
+custom_complex_lt = register_jitable(lt_complex)
 
 
 @register_jitable
@@ -3754,11 +3728,6 @@ def custom_complex_le(a, b):
                     elif a.real == b.real:
                         return a.imag <= b.imag
                     return False
-
-
-@register_jitable
-def less_than_or_equal(a, b):
-    return a <= b
 
 
 def _searchsorted(func_1, func_2):
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -3674,107 +3674,89 @@ def np_bincount(a, weights=None, minlength=0):
     return bincount_impl
 
 
+@register_jitable
+def custom_lt(a, b):
+    if np.isnan(b):
+        return not np.isnan(a)
+    return a < b
+
+
+@register_jitable
+def custom_le(a, b):
+    if np.isnan(b):
+        return True
+    return a <= b
+
+
 def _searchsorted(func):
-    def searchsorted_inner(a, v, v_last, lo, hi, n):
-        """Perform inner loop of searchsorted (i.e. a binary search).
+    def impl(a, v):
+        min_idx = 0
+        max_idx = len(a)
 
-        This is loosely based on the NumPy implementation in [1]_.
+        if v.size == 0:
+            return
 
-        Parameters
-        ----------
-        a: 1-D array_like
-            The input array.
-        v: array_like
-            The current value to insert into `a`.
-        v_last: array_like
-            The previous value inserted into `a`.
-        lo: int
-            The initial/previous "low" value of the binary search.
-        hi: int
-            The initial/previous "high" value of the binary search.
-        n: int
-            The length of `a`.
+        out = np.empty(v.size, np.intp)
 
+        last_key_val = v.flat[0]
 
-        .. [1] https://github.com/numpy/numpy/blob/809e8d26b03f549fd0b812a17b8a166bcd966889/numpy/core/src/npysort/binsearch.cpp#L173
-        """  # noqa: E501
-        if np.isnan(v):
-            # Find the first nan (i.e. the last from the end of a,
-            # since there shouldn't be many of them in practice)
-            for i in range(n, 0, -1):
-                if not np.isnan(a[i - 1]):
-                    return i
-            return 0
+        for i in range(v.size):
+            key_val = v.flat[i]
 
-        if v_last < v:
-            hi = n
-        else:
-            lo = 0
-            hi = hi + 1 if hi < n else n
-
-        while hi > lo:
-            mid = (lo + hi) >> 1
-            if func(a[mid], (v)):
-                # mid is too low => go up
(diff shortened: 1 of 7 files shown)


rjenc29:
An example of:

> ...input array `a` is properly sorted but contains NaN and side='right'.

...which behaves per numpy in this PR:

```python
import numpy as np
from numba import njit

a = np.array([42., 59., 62., 98., np.nan])
v = np.array([42., 59., 62., np.nan])

@njit
def foo(a, v):
    return np.searchsorted(a, v, 'right')

print(foo(a, v))
print(foo.py_func(a, v))
```

rjenc29:
Ah yes, @guilhermeleobas - I think it probably is (apologies, I raised the PR in such a way that I could not attach the appropriate label).

numba/np/arraymath.py
```diff
@@ -3674,107 +3674,101 @@ def bincount_impl(a, weights=None, minlength=0):
...
+                min_idx = 0
+                if max_idx < len(a):
+                    max_idx += 1
+                else:
+                    max_idx = len(a)
+
+            last_key_val = key_val
 
+            while min_idx < max_idx:
+                mid_idx = min_idx + ((max_idx - min_idx) >> 1)
```
guilhermeleobas, line 3718: Could add a comment here saying this is to avoid overflow

guilhermeleobas (changes requested):
Hi @rjenc29, great work. I only found one corner case which should be easy to address:

```python
import numpy as np
from numba import njit

a = np.arange(1)
v = np.arange(0)

@njit
def foo(a, v):
    return np.searchsorted(a, v, 'left')

print(foo(a, v))   # None
print(foo.py_func(a, v))  # []
```

But otherwise PR looks excelent.

numba/np/arraymath.py
rjenc29: done

rjenc29:
Hi @guilhermeleobas , thanks for flagging that edge case - I added a test for it and tweaked the implementation to suit & cleaned up the tests a little, in passing.

I briefly toyed with extending support to string inputs / branching the overload to avoid calling `np.isnan` for dtypes which are exact, but regretted these ideas quite quickly and figured maybe it would be best to pursue them in other PRs, if at all.

Anyway, I will stop fiddling with this PR for the time being - let me know what you think.

guilhermeleobas:
Hi @rjenc29, thank you for your contribution. I will review it again later this week.

numba/np/arraymath.py
```diff
@@ -3674,107 +3674,102 @@ def bincount_impl(a, weights=None, minlength=0):
...
-        loop_impl = _searchsorted_left
+        _impl = _searchsorted_left
     elif side_val == 'right':
-        loop_impl = _searchsorted_right
+        # change in behaviour introduced by
+        # https://github.com/numpy/numpy/pull/21867
+        if numpy_version >= (1, 23):
+            _impl = _searchsorted_right_np123_on
+        else:
+            _impl = _searchsorted_right_pre_np123
```
guilhermeleobas, line 3758: As of today, `side` argument needs to be a literal value known at compile-time. This could be changed to be a runtime value:

```python
    if isinstance(v, types.Array):
        def impl(a, v, side='left'):
            if side == 'left':
                return _searchsorted_left(a, v)
            elif side == 'right':
                return _searchsorted_right_np123_on(a, v)
            else:
                raise ValueError('error!')
```

But can be done in a future work as this goes beyond the scope of this PR.

numba/tests/test_np_functions.py
```diff
@@ -1327,6 +1327,67 @@ def nonconst_side(a, v, side='left'):
...
+        # `a` and `v` booleans
+        a = np.array([False, False, True, True])
+        v = np.array([False, True])
+        check(a, v)
+
+        # `v` is a (scalar) boolean
+        a = [1, 2, 3]
+        v = True
+        check(a, v)
+
```
guilhermeleobas, line 1390: Could you also include a test for complex types? The following is crashing on your PR:

```python
import numpy as np
from numba import njit

side = 'left'

@njit
def foo(a, v):
    return np.searchsorted(a, v, side='left')

a = np.zeros(9, dtype=np.complex64)
a.real += [0, 0, 1, 1, 0, 1, np.nan, np.nan, np.nan]
a.imag += [0, 1, 0, 1, np.nan, np.nan, 0, 1, np.nan]
v = a
print()
print(a, v, side)
print(foo(a, v))
print(a.searchsorted(v, side=side))
```

guilhermeleobas (changes requested):
Thanks @rjenc29, impressive work. I found another test case (complex types) that it is not working. Hope that should be easy to address. If not, let me know and we can postpone it to a future PR.

numba/tests/test_np_functions.py
rjenc29: Ah yes, I think that snippet will similarly raise in the current implementation / main - in that regard, it's not a behaviour which is changed or introduced in this PR.

I have a feeling that numba in general doesn't support the ordering of complex numbers - but I guess it would be possible to branch this implementation to do whatever numpy does (which would make it an exception, I _think_).

I guess the following could also be caught handled.

```python
import numpy as np
from numba import njit

@njit
def foo(a, v):
    return np.searchsorted(a, v)

a = np.array(['1', '2', '3'])
v = np.array(['2', '4'])

print(foo(a, v))
```

I'll have a look at these things when I get some time.

numba/np/arraymath.py
rjenc29: Good point - I tried to keep the changeset as small as possible in this PR; could be a good candidate change though.

rjenc29: I added a comment re this point

numba/tests/test_np_functions.py
rjenc29: @guilhermeleobas in the last few commits, I have extended support for complex dtypes and handled the example in the prior comment.

As this involved implementing a lex sort for complex numbers a la numpy, I plumbed it into `np.sort` (and `array.sort`) too since it was a very small incremental change. If this conflation of changes seems unwise, let me know and I will extract it into a different PR.

There are some residual edge cases not handled but they never have been and would require a lot more work, so I'd propose to cut the scope here and would kindly request a re-review when you have a mo.

rjenc29:
@guilhermeleobas - just to note: I think this PR is ready for a re-review when you have time.

guilhermeleobas:
Hi @rjenc29, I was out this past week. Will review your PR now

numba/np/arraymath.py
```diff
@@ -3674,107 +3674,177 @@ def bincount_impl(a, weights=None, minlength=0):
...
+def make_searchsorted_implementation(np_dtype, side):
+    assert side in VALID_SEARCHSORTED_SIDES
+
+    if np.issubdtype(np_dtype, np.floating):
+        _impl = bind_operators(
+            side,
+            less_than_float,
+            less_than_or_equal_float,
+            use_version_gate=True
+        )
```
guilhermeleobas, line 3800: Could we reduce code duplication here and leverage the use of `isinstance` to avoid having a `less_than_float` and `less_than_complex`? For instance, the `less_than_or_equal` function could be enhanced to support floats with the following addition:

```python
@register_jitable
def less_than_or_equal(a, b):
    if isinstance(b, float):
        if np.isnan(b):
            return True
    return a <= b
```

For the following to work, you need to update the `supported_var_ty` variable in `builtins.py` to add two new entries:

```python
    supported_var_ty = (types.Number, types.Bytes, types.RangeType,
                        types.DictType, types.LiteralStrKeyDict, types.List,
                        types.ListType, types.Tuple, types.UniTuple, types.Set,
                        types.Function, types.ClassType, types.UnicodeType,
                        types.ClassInstanceType, types.NoneType, types.Array,
                        types.Boolean, types.UnicodeCharSeq)  # <----
```

guilhermeleobas (changes requested):
Thanks again for your contributions @rjenc29. Let me know what you think about using `isinstance` instead of implementing a `less_than` comparison for float and complex. If you think that is too much work, I'm fine with the code you have at the moment.

numba/np/arraymath.py
rjenc29: Thanks very much for the review and feedback, @guilhermeleobas.

In my local branch, I have put something together which avoids separations like `less_than_float` and `less_than_complex`, but it has ended up looking a bit odd and might conceivably have performance / other implications. 

This probably means I've not interpreted your feedback correctly - nonetheless, I am happy to push it for review if that's useful.

rjenc29: Further to the above, I've pushed my branch - let me know what you think in due course.

guilhermeleobas:
@rjenc29, sorry for the delay. I was on PTO for the past few days. I'll do a proper review soon.

numba/np/arrayobj.py
```diff
@@ -6314,42 +6315,52 @@ def impl(ary, indices_or_sections):
...
         return func
 
 
+def lt_implementation(dtype):
+    if isinstance(dtype, types.Float):
+        return lt_floats
+    elif isinstance(dtype, types.Complex):
+        return lt_complex
+    else:
+        return default_lt
```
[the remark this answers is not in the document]
rjenc29: Hey, yep it's logically equivalent - I'm not sure, but I don't think I could use it here though as this function ends up being compiled and I think there's a check that the compilation target is a user function (i.e. not a built-in).

numba/tests/test_np_functions.py
guilhermeleobas: Thanks @rjenc29. Could you create an issue to track this limitation?

guilhermeleobas (approved):
Pull request looks good to me.

PR #9205, opened 2023-09-21
release notes and version support updates from release0.58 branch

esc:
As title


docs/upcoming_changes/9126.np_support.rst
```diff
@@ -1,4 +0,0 @@
-Added support for np.indices() function.
```
sklam, line 1: #9126 is in 0.59 milestone. It should not be removed on main.

esc: thank you for catching this, the snippet has been restored in 1754b8b9f824922ba4f6d0c8e2c8b1ab7a63cfaf

sklam (approved):

PR #9220, opened 2023-10-03
Support passing arbitrary flags to NVVM

gmarkall:
Maintaining a list of flags that we know about for NVVM and validating them is extra work to keep in sync with the CUDA documentation.

This PR removes this maintenance burden and allows a bit more flexibility for future changes by allowing arbitrary flags to be passed through, transforming them in the same manner as before but without a restriction on what the flag names are.

Some flags may not take a parameter - `-gen-lto` is an example of this. To support this case, options can be passed with a value of `None`. Since this expands slighty on the functionality, a new test cases is added with this particular flag, that validates it was passed correctly based on the NVVM output.

### DIFF
--- a/numba/cuda/cudadrv/nvvm.py
+++ b/numba/cuda/cudadrv/nvvm.py
@@ -231,72 +231,45 @@ class CompilationUnit(object):
         self.driver.check_error(err, 'Failed to add module')
 
     def compile(self, **options):
-        """Perform Compilation
+        """Perform Compilation.
 
-        The valid compiler options are
+        Compilation options are accepted as keyword arguments, with the
+        following considerations:
 
-         *   - -opt=
-         *     - 0 (disable optimizations)
-         *     - 3 (default, enable optimizations)
-         *   - -arch=
-         *     - compute_XX where XX is in (35, 37, 50, 52, 53, 60, 61, 62, 70,
-         *                                  72, 75, 80, 86, 89, 90).
-         *       The default is compute_52.
-         *   - -ftz=
-         *     - 0 (default, preserve denormal values, when performing
-         *          single-precision floating-point operations)
-         *     - 1 (flush denormal values to zero, when performing
-         *          single-precision floating-point operations)
-         *   - -prec-sqrt=
-         *     - 0 (use a faster approximation for single-precision
-         *          floating-point square root)
-         *     - 1 (default, use IEEE round-to-nearest mode for
-         *          single-precision floating-point square root)
-         *   - -prec-div=
-         *     - 0 (use a faster approximation for single-precision
-         *          floating-point division and reciprocals)
-         *     - 1 (default, use IEEE round-to-nearest mode for
-         *          single-precision floating-point division and reciprocals)
-         *   - -fma=
-         *     - 0 (disable FMA contraction)
-         *     - 1 (default, enable FMA contraction)
-         *
-         """
+        - Underscores (`_`_ in option names are converted to dashes (`-`), to
+          match NVVM's option name format.
+        - Options that take a value will be emitted in the form
+          "-<name>=<value>".
+        - Options which take no value (such as `-gen-lto`) should have a value
+          of `None` passed in and will be emitted in the form "-<name>".
 
-        # stringify options
-        opts = []
+        For documentation on NVVM compilation options, see the CUDA Toolkit
+        Documentation:
 
-        if 'opt' in options:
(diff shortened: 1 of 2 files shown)


gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
I've added the label to skip the release notes check because this is a refactoring that doesn't affect public APIs.

numba/cuda/tests/cudadrv/test_nvvm_driver.py
```diff
@@ -25,6 +25,22 @@ def test_nvvm_compile_simple(self):
         self.assertTrue('simple' in ptx)
         self.assertTrue('ave' in ptx)
 
+    def test_nvvm_compile_nullary_option(self):
```
stuartarchibald, line 28: Suggest adding a test for a "typo"ed option, like `nvvm.llvm_to_ptx(nvvmit, opt=3, this_is_not_an_option=1234)` and asserting that the error message produced is informative with respect to the mistake.

numba/cuda/cudadrv/nvvm.py
```diff
@@ -231,72 +231,45 @@ def lazy_add_module(self, buffer):
...
+        https://docs.nvidia.com/cuda/libnvvm-api/index.html#_CPPv418nvvmCompileProgram11nvvmProgramiPPKc
+        """
+
+        def stringify_option(k, v):
+            k = k.replace('_', '-')
+
+            if v is None:
+                return f'-{k}'
+
+            if isinstance(v, bool):
```
stuartarchibald, line 258: Want to mention this conversion convention in the docs/tests?

numba/cuda/tests/cudadrv/test_nvvm_driver.py
```diff
@@ -25,6 +25,22 @@ def test_nvvm_compile_simple(self):
...
+        # -gen-lto is not available prior to CUDA 11.5
+        if runtime.get_version() < (11, 5):
+            self.skipTest("-gen-lto unavailable in this toolkit version")
+
+        nvvmir = self.get_nvvmir()
+        ltoir = nvvm.llvm_to_ptx(nvvmir, opt=3, gen_lto=None, arch="compute_52")
+
+        # Verify we correctly passed the option by checking if we got LTOIR
+        # from NVVM (by looking for the expected magic number for LTOIR)
+        self.assertEqual(ltoir[:4], b'\xed\x43\x4e\x7f')
```
stuartarchibald, line 42: It would be good to get a reference for this sequence but it's understandable if it's not possible.

stuartarchibald (commented):
Thanks for the patch. This is a good improvement to reduce maintenance, I've left a couple of comments in-line once resolved should be good to merge. Thanks again.

numba/cuda/tests/cudadrv/test_nvvm_driver.py
gmarkall: I agree - there is no reference, unfortunately.

numba/cuda/tests/cudadrv/test_nvvm_driver.py
gmarkall: I've added a test for this.

numba/cuda/cudadrv/nvvm.py
```diff
@@ -231,72 +231,46 @@ def lazy_add_module(self, buffer):
...
-            raise NvvmError("unsupported option {0}".format(optstr))
-
-        c_opts = (c_char_p * len(opts))(*[c_char_p(x.encode('utf8'))
-                                          for x in opts])
+        """Perform Compilation.
+
+        Compilation options are accepted as keyword arguments, with the
+        following considerations:
+
+        - Underscores (`_`_ in option names are converted to dashes (`-`), to
```
stuartarchibald, line 239:
```suggestion
        - Underscores (`_`) in option names are converted to dashes (`-`), to
```

gmarkall:
gpuci run tests

stuartarchibald (approved):
Thanks for the patch and fixes, change set looks good on inspection.

gmarkall:
@stuartarchibald Many thanks for the review!

stuartarchibald:
> > @stuartarchibald Many thanks for the review!
> 
> No problem!
> 
> Buildfarm ID: `numba_smoketest_cuda_yaml_219`.

Passed.

PR #9140, opened 2023-08-17
Add support for Polynomial class

KrisMinchev:
This PR adds basic support for the [Polynomial](https://numpy.org/doc/stable/reference/generated/numpy.polynomial.polynomial.Polynomial.html) class from `numpy.polynomial.polynomial` package. The current implementation only supports the paramters `coef`, `domain` and `window`.

PR #9221, opened 2023-10-03
Add support for Chebyshev class from `np.polynomial.chebyshev`

KrisMinchev:
This PR adds basic support for the [Chebyshev](https://numpy.org/doc/stable/reference/generated/numpy.polynomial.chebyshev.Chebyshev.html) class from numpy.polynomial.chebyshev package. The current implementation only supports the parameters coef, domain and window. Implementation is similar to that in #9140.

### DIFF
--- a/numba/core/types/npytypes.py
+++ b/numba/core/types/npytypes.py
@@ -647,3 +647,15 @@ class PolynomialType(Type):
         # constructor, since the types of domain and window arguments depend on
         # that and we need that information when boxing
         self.n_args = n_args
+
+
+class ChebyshevType(Type):
+    def __init__(self, coef, domain=None, window=None, n_args=1):
+        super(ChebyshevType, self).__init__(name=f'ChebyshevType({coef}, {domain}, {domain}, {n_args})')
+        self.coef = coef
+        self.domain = domain
+        self.window = window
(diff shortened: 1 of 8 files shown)


stuartarchibald:
Thanks for the patch @KrisMinchev I've marked it for review.

numba/core/types/npytypes.py
```diff
@@ -647,3 +647,15 @@ def __init__(self, coef, domain=None, window=None, n_args=1):
         # constructor, since the types of domain and window arguments depend on
         # that and we need that information when boxing
         self.n_args = n_args
+
+
+class ChebyshevType(Type):
+    def __init__(self, coef, domain=None, window=None, n_args=1):
+        super(ChebyshevType, self).__init__(name=f'ChebyshevType({coef}, {domain}, {domain}, {n_args})')
```
guilhermeleobas, line 654: You can reuse most of the code written for Polynomial here. Create a base type, move all the code common to both types to the base type and inherit from base instead of `Type.`

guilhermeleobas (changes requested):
Hi Kristian, thanks again for your contributions to the Numba codebase. I have a request that will simplify the code for Chebyshev and one can reuse the code written before. Let me know your thoughts on this.

PR #9223, opened 2023-10-04
CUDA: Add support for compiling device functions with C ABI

gmarkall:
This makes it easier to use Numba as a compiler for user-defined Python functions, by more closely matching the prototype from other languages like C and C++.

A C calling convention is implemented using Numba's calling convention mechanisms - whilst this was possibly not technically required to support a C ABI for device functions, it should provide a base for further internal use of a C ABI where appropriate / possible (which will be more efficient than the Numba ABI).

Documentation of the ABI is added by adding a section on compilation to the user documentation - having some user docs for the compilation APIs was probably a good idea anyway.

### DIFF
--- a/docs/source/cuda/cuda_compilation.rst
+++ b/docs/source/cuda/cuda_compilation.rst
@@ -1,0 +1,107 @@
+
+.. _cuda_compilation:
+
+Compiling Python functions for use with other languages
+=======================================================
+
+Numba can compile Python code to PTX so that Python functions can be
+incorporated into CUDA code written in other languages (e.g. C/C++).  It is
+commonly used to support User-Defined Functions written in Python within the
+context of a library or application.
+
+The main API for compiling Python to PTX can be used without a GPU present, as
+it uses no driver functions and avoids initializing CUDA in the process. It is
+invoked through the following function:
+
+.. autofunction:: numba.cuda.compile_ptx
+   :noindex:
+
+If a device is available and PTX for the compute capability of the current
+device is required (for example when building a JIT compilation workflow using
+Numba), the ``compile_ptx_for_current_device`` function can be used:
+
+.. autofunction:: numba.cuda.compile_ptx_for_current_device
+   :noindex:
+
+
+Using the C ABI
+---------------
+
+Numba compiles functions with its ABI by default - this is as described in
+:ref:`device-function-abi`, without the ``extern "C"`` modifier. Calling Numba
+ABI device functions requires three issues to be addressed:
+
+- The name of the function will be mangled according to Numba's ABI rules -
+  these are based on the Itanium C++ ABI rules, but are extended beyond its
+  specifications.
+- The Python return value is expected to be stored into a pointer value passed
+  in the first argument.
+- The return value of the compiled function will contain a status code, instead
+  of the return value of the function. For use of Numba-compiled functions
+  outside of Numba, this can generally be ignored.
+
+A simple way to address all these issues is to compile device functions with the
+C ABI instead. This results in the following:
+
+- The name of the compiled device function in PTX will match the name of the
+  function in Python, so it is easy to determine. This is the function's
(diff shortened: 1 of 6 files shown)


gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

docs/source/cuda/cuda_compilation.rst
```diff
@@ -0,0 +1,107 @@
...
+.. code:: python
+
+   def add(x, y):
+       return x + y
+
+compiled for the Numba ABI using, for example:
+
+.. code:: python
+
+    ptx, resty = ptx, resty = cuda.compile_ptx(add, int32(int32, int32), device=True)
```
stuartarchibald, line 69:
```suggestion
    ptx, resty = cuda.compile_ptx(add, int32(int32, int32), device=True)
```

numba/cuda/target.py
```diff
@@ -366,3 +366,72 @@ def get_ufunc_info(self, ufunc_key):
...
+
+    def return_user_exc(self, builder, exc, exc_args=None, loc=None,
+                        func_name=None):
+        msg = "Python exceptions are unsupported in the CUDA C/C++ ABI"
+        raise NotImplementedError(msg)
+
+    def return_status_propagate(self, builder, status):
+        msg = "Return status is unsupported in the CUDA C/C++ ABI"
+        raise NotImplementedError(msg)
+        pass
```
stuartarchibald, line 397:
```suggestion
```

stuartarchibald (commented):
Thanks for working on this feature @gmarkall, it's great to see this working. I've given the implementation and docs a read through and they look good (just a couple of minor things to look at). Next thing I'd like to do is give this a try in practice. Thanks again!

gmarkall:
gpuci run tests

gmarkall:
@stuartarchibald Many thanks for the review!

gmarkall:
@stuartarchibald I just pushed https://github.com/numba/numba/pull/9223/commits/7fc5e030d9728e3ef838670e9bdbaefd0fddfea5 to address the concern raised out-of-band, that the function names for different variants will all clash if their names are all the same. The implementation allows the user to specify the name which I think will be more convenient / practical than forcing a specific convention - it doesn't require us to figure out a convention upfront, and gives users more flexibility that they might need.

As per our additional discussion, the option goes through a dict, `abi_info`, so that future ABIs that need different information can get it through this dict without reworking the compilation ABI again. With this method, we also control exactly what information goes to it, rather than passing any `**kwargs` to the ABI handler, which in my experience makes it confusing and difficult to reason about exactly what is accepted and passed on through the API.

numba/cuda/compiler.py
```diff
@@ -283,6 +283,10 @@ def compile_ptx(pyfunc, sig, debug=False, lineinfo=False, device=False,
                 ``"c"``. Note that the Numba ABI is not considered stable.
                 The C ABI is only supported for device functions at present.
     :type abi: str
+    :param abi_info: A dict of ABI-specific options. The ``"c"`` ABI supports
+                     one option, ``"abi_name"``, for providing the wrapper
+                     function's name. The ``"numba"`` ABI has no options.
```
stuartarchibald, line 288: Should the code assert that if the ABI in use is `"numba"` then the `abi_name` key in the `abi_info` dictionary is not set?

stuartarchibald (commented):
Thanks for the update. One query to resolve else looks good.

numba/cuda/compiler.py
gmarkall: I think it would be a strange precedent to set - what if another ABI adds another option - should all ABIs assert that they are only passed specifically the items they support? (My leaning is that they should not, simply because the burden of doing that guarding outweighs the potential small benefit of occasionally catching passing an irrelevant / misspelled key)

stuartarchibald (approved):

PR #9228, opened 2023-10-06
Fix #9222 - Don't replace `.` with `_` in func arg names in inline closures

### LINKED ISSUE #9222: Invalid closure handling in _create_function_from_code_obj
## Reporting a bug

Closure handling was updated as part of Python 3.11 upgrade to include [this line](https://github.com/numba/numba/blob/9f74128def1ca51deebfadecc663f0b80f641c57/numba/core/ir_utils.py#L1743) (`func_arg = func_arg.replace('.', '_')`), which replaces dots in function argument signature (to fix generated argument names I assume). However, this is invalid for default values that are float constants as demonstrated below (removed `InlineClosureLikes` to go through `MakeFunctionToJitFunction` code path which passes default values).

The commit that made this change ([here](https://github.com/numba/numba/commit/77c2672e22be8cb33e47f44884385fd75906881c)), mentions fixing `numba.tests.test_looplifting.TestLoopLiftingInAction.test_lift_listcomp_block0` but the test works with or without this change. Seems like the motivating problem may have been solved in the upstream calls and we can just remove the `func_arg = func_arg.replace('.', '_')` line. Is this correct?

```python
import numba
from numba.core.compiler import DefaultPassBuilder
from numba.core.untyped_passes import InlineClosureLikes

class TestPipeline(numba.core.compiler.CompilerBase):
    def define_pipelines(self):
        pm = DefaultPassBuilder.define_nopython_pipeline(self.state, "TestPipeline")
        for idx, (x, _) in enumerate(pm.passes):
            if x == InlineClosureLikes:
                break
        else:
            raise Exception("Could not find pass %s" % location)
        pm.passes.pop(idx)
        pm.finalize()
        return [pm]

@numba.njit(pipeline_class=TestPipeline)
def f(a):
    def g(x, y=1.1):
        return x+y
    return g(a)

# prints 15 instead of 5.1
print(f(4))
```

gmarkall:
This is suggested as a fix for Issue #9222 in its description, as it seems it may no longer be necessary.

Testing on CI for now.

### DIFF
--- a/docs/upcoming_changes/9222.bug_fix.rst
+++ b/docs/upcoming_changes/9222.bug_fix.rst
@@ -1,0 +1,5 @@
+Fixed handling of float default arguments in inline closures
+============================================================
+
+Float default arguments in inline closures would produce incorrect results since
+updates for Python 3.11 - these are now handled correctly again.
--- a/numba/core/ir_utils.py
+++ b/numba/core/ir_utils.py
@@ -1739,8 +1739,6 @@ def _create_function_from_code_obj(fcode, func_env, func_arg, func_clo, glbls):
     * func_clo - string for the closure args
     * glbls - the function globals
     """
-    # in py3.11, func_arg contains '.'
-    func_arg = func_arg.replace('.', '_')
     sanitized_co_name = fcode.co_name.replace('<', '_').replace('>', '_')
     func_text = (f"def closure():\n{func_env}\n"
                  f"\tdef {sanitized_co_name}({func_arg}):\n"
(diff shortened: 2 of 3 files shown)


gmarkall:
gpuci run tests

gmarkall:
Tests ran OK for me locally on Python 3.11:

```
Ran 11860 tests in 2518.960s

OK (skipped=576, expected failures=29)
```

gmarkall:
gpuci run tests

esc (approved):
LGTM pending smoketest

esc:
> LGTM pending smoketest

was green

Issue #7662, opened 2021-12-15
IntEnum as array index? No implementation of getitem found for signature


JapanYoshi:
## Reporting a bug

- [X] I have tried using the latest released version of Numba (most recent is
 visible in the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [X] I have included a self contained code sample to reproduce the problem.
  i.e. it's possible to run as 'python bug.py'.
[advent_2021_15.zip](https://github.com/numba/numba/files/7722820/advent_2021_15.zip)

I have an observably bad Python program that I tried to apply Numba to. Unfortunately, JIT fails with some arcane reason that could be related to trying to use an IntEnum as an array index.
```
Traceback (most recent call last):
  File "advent_2021_15.py", line 194, in <module>
    (width + height) * pages - 2, # heuristic
  File "C:\Users\Haley\AppData\Roaming\Python\Python37\site-packages\numba\core\dispatcher.py", line 482, in _compile_for_args
    error_rewrite(e, 'typing')
  File "C:\Users\Haley\AppData\Roaming\Python\Python37\site-packages\numba\core\dispatcher.py", line 423, in error_rewrite
    raise e.with_traceback(None)
numba.core.errors.TypingError: Failed in nopython mode pipeline (step: nopython frontend)
No implementation of function Function(<built-in function getitem>) found for signature:

 >>> getitem(reflected list(int32)<iv=None>, IntEnum<int32>(Route))

There are 22 candidate implementations:
      - Of which 22 did not match due to:
      Overload of function 'getitem': File: <numerous>: Line N/A.
        With argument(s): '(reflected list(int32)<iv=None>, IntEnum<int32>(Route))':
       No match.

During: typing of intrinsic-call at advent_2021_15.py (170)
During: typing of static-get-item at advent_2021_15.py (170)

File "advent_2021_15.py", line 170:
def push_route(route):
    <source elided>
 for i, r in enumerate(frontier):
  if r[Route.X] == route[Route.X] and r[Route.Y] == route[Route.Y]:
  ^
```

JapanYoshi:
I've begrudgingly removed the enum and I get this error:

`The use of a reflected list(reflected list(int32)<iv=None>)<iv=None> type, assigned to variable 'frontier' in globals, is not supported as globals are considered compile-time constants and there is no known way to compile a reflected list(reflected list(int32)<iv=None>)<iv=None> type as a constant.`

gmarkall:
[...]

PR #9230, opened 2023-10-09
allow using IntEnumMember as array index

dlee992:
related with https://github.com/numba/numba/issues/7662. but fix the similar issue for np.array.

**(edit by @gmarkall):** Issue #7662 seems to have disappeared, but from what we could work out with DDG searches, it seems to have been about the error `No implementation of function Function() found for signature: array(int32, 1d, C)[EnumType]`.


dlee992:
This test config `numba.numba (Linux py310_np123)` seems failed at installing python packages..

gmarkall (dismissed):
Many thanks for the PR - could you also add a check with an `IntEnum` where negative values are used in the Enum?

Also, this should have a release note adding (the `skip_release_notes` tag was erroneously added before) - create a new file called `docs/upcoming_chanages/9230.improvement.rst` following the format suggested in https://github.com/numba/numba/blob/main/docs/upcoming_changes/README.rst (you can also look at some of the other files in https://github.com/numba/numba/tree/main/docs/upcoming_changes for an idea).

dlee992:
Adding more tests and a simple doc as you suggested right now. @gmarkall

stuartarchibald:
@dlee992 thanks for the update. It looks like 45056af contains a lot of automated formatting changes which are "hiding" the proposed change, please could these be removed so that the review can continue? Many thanks.

gmarkall (commented):
I'm finding that I can't make the tests added here fail on `main` - @dlee992 is there a circumstance in which the tests fail on `main` for you? (I'm using Python 3.10)

dlee992:
Em, I met this issue on my customised 0.57.1. Perhaps another PR has already fixed this somehow?

numba/tests/test_indexing.py
```diff
@@ -521,6 +528,11 @@ def test_1d_integer_indexing(self, flags=enable_pyobj_flags):
         self.assertEqual(pyfunc(a, 9), cfunc(a, 9))
         self.assertEqual(pyfunc(a, -1), cfunc(a, -1))
 
+        # use IntEnumMember as index
+        self.assertEqual(pyfunc(a, 0), cfunc(a, EnumIndex.INDEX0))
```
gmarkall, line 532: These tests are testing with a function compiled for argument types `(arraytype, types.int32)` only - therefore it's not actually testing with an `IntEnumMember`. You need to compile a new version where the second argument is of the `IntEnumMember` type (e.g. `types.IntEnumMember(EnumIndex, types.int32)`) to test using an `IntEnumMember` as an array index.

gmarkall:
> Em, I met this issue on my customised 0.57.1. Perhaps another PR has already fixed this somehow?

I just realised what the problem with the test is - see the comment on the diff.

numba/tests/test_indexing.py
dlee992: rewrote the test cases. Thanks for ur suggestions!

dlee992:
Em, the error is `NameError: name 'compile_isolated' is not defined`. But it's imported at line 8.. Don't know why..
https://dev.azure.com/numba/numba/_build/results?buildId=16490&view=logs&j=7e7ebb60-d573-5e5d-b793-95b37f06385f&t=5ac25388-b2c0-5549-4759-b1f5889bad88&l=486

gmarkall:
It's because this was merge, which removed `compile_isolated`: https://github.com/numba/numba/pull/9330

dlee992:
> It's because this was merge, which removed compile_isolated: https://github.com/numba/numba/pull/9330
Got it. Work on this now.

dlee992:
This can be in review status again. gentle cc @gmarkall

esc:
> This can be in review status again. gentle cc @gmarkall

Thank you for the heads up, it's done.

neich:
Any chance to accept this PR? This would be really useful

Thanks!

dlee992:
@neich. Hi, Thanks for your notice.

I think this is added into milestone 0.61.0-rc1, it should be reviewed and merged when core devs begin to take actions for all issues and PRs in this release candidate.

Perhaps Oct or Nov this year, I guess. But you can use this patch locally as well, then don't have to wait for an official release.

dlee992:
I closed my other open PRs in the PR list.
For this one, since another user (neich) also wanna have this feature, perhaps someone can review it when available.

gmarkall:
I put this in a milestone with the aim of saving it from getting lost in future.

neich:
Any chance to accept this PR anytime soon ? Thanks

esc:
> Any chance to accept this PR anytime soon ? Thanks

@neich thank you for asking about this, two weeks have passed since your comment, so the answer to your question was probably a "no".

@gmarkall has added it to the 0.62.0 milestone so the current expectation is to get this reviewed within the next 3 to 6 months, but no firm commitment or guarantee can be made. Numba is a highly resource constrained project and the number of submitted PRs and issues vastly outnumbers the available resources for review and thus PRs may take years until they can be reviewed (this one was opened back in 2023). Thank you in advance for your kind understanding and patience.

dlee992:
I would appreciate it if someone could take over this PR and complete the 'changes requested' task, as I am currently unavailable.

neich:
@gmarkall let me know if I can contribute in some way

gmarkall:
@neich Do you have an example that reproduces the issue this PR fixes? The test cases added here still also seems to pass on `main`, so I don't think it's testing whatever changed. If you can provide an example I should be able to incorporate that into the test case and make this mergeable.

gmarkall:
(I've just merged `main` here so this PR is up to date and gets a recent CI run)

neich:
@gmarkall I've tried to write the simplest example. The PR allows to avoid to explicitly convert any IntEnum element to a indexing integer.

```
import numpy as np
import numba as nb
from enum import IntEnum, auto

class EnumIndex(IntEnum):
    INDEX0 = 0
    INDEX1 = auto()
    INDEX_NEG1 = -1

@nb.njit
def indexing_error(a):
    return a[EnumIndex.INDEX0]

@nb.njit
def indexing_ok(a):
    return a[np.intp(EnumIndex.INDEX0)]


a = np.arange(10)
print(indexing_ok(a))     # This will work correctly
print(indexing_error(a))  # This will raise an error
```

With the changes introduced in the PR, the test added tries to check if indexing directly with the IntEnum inside a numba function returns the same as indexing the array outside numba. Of course without using the explicit conversion. The example above should not raise an error with this PR.

To see a practical example, check this code and see the amount of verbosity that could be avoided with this PR:

https://github.com/neich/neuronumba/blob/f3281b76bac8e38474d6da3c140435827b118922/src/neuronumba/simulator/models/montbrio.py#L150

swap357 (commented):
I've reviewed this PR and it looks good to me. `get_array_index_type` is the right place for this. Without the patch, `test_1d_enum_indexing_npm` fails to type, and with the patch it passes.
```
>>> @jit
... def f(a):
...     return a[E.I0]
...     
>>> f.py_func(a)
np.int64(0)
>>> f(a)
  File "<python-input-11>", line 1, in <module>
    f(a)
    ~^^^
  File "/Users/swap357/Documents/dev/numba/numba/core/dispatcher.py", line 422, in _compile_for_args
    error_rewrite(e, 'typing')
    ~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "/Users/swap357/Documents/dev/numba/numba/core/dispatcher.py", line 363, in error_rewrite
    raise e.with_traceback(None)
numba.core.errors.TypingError: Failed in nopython mode pipeline(step: nopython frontend)
 
 >>> getitem(array(int64, 1d, C), IntEnum<int64>(E))

```

with patch -
```
>>> @jit
... def f(a):
...     return a[E.I0]
...     
>>> f.py_func(a)
np.int64(0)
>>> f(a)
0
```

I've also added test covering original issue report with [88a1fd](https://github.com/numba/numba/pull/9230/commits/a88a1fd0838f96ec7e996f2feac3e5ce738ba9f1)

docs/upcoming_changes/9230.improvement.rst
```diff
@@ -0,0 +1,4 @@
+Added support for using IntEnum as NumPy array's index.
+-------------------------------------------------------
+
+Added support for using IntEnum as NumPy array's index.
```
sklam: This is just repeating the headline. Can we have something more descriptive here?

sklam (commented):
One nitpick on the towncrier.

docs/upcoming_changes/9230.improvement.rst
sklam: Maybe just put an example here

swap357: addressed it with [f7b94bc](https://github.com/numba/numba/pull/9230/commits/f7b94bc981ff6f585d0b88134afdf75d5279ebee)

sklam (approved):

PR #9235, opened 2023-10-10
CUDA: Make `grid()` and `gridsize()` use 64-bit integers

### LINKED ISSUE #9229: cuda.grid(1) returns incorrect value due to signed integer overflow
`cuda.grid(1)` returns an incorrect value in the following code:

```py
import numpy as np
from numba import cuda


@cuda.jit
def bug(out):
    i = cuda.threadIdx.x + cuda.blockIdx.x * cuda.blockDim.x
    i2 = cuda.grid(1)
    if i != i2:
        # this shouldn't happen
        print(i, i2)


free_mem, _ = cuda.current_context().get_memory_info()
# max_num_parallel_iter is around 566281830 on my 4090
max_num_parallel_iter = free_mem // 4 - 128 * 1024 * 1024
dev_out = cuda.device_array(max_num_parallel_iter, dtype=np.float32)
bug[22121216, 256](dev_out)
```

One example of a case which is printed is when `i = 2147551327` and `i2 = -214741596`, which likely indicates a signed 32 bit integer overflow. The bug doesn't appear when setting `max_num_parallel_iter` to an integer constant, which must mean it's somehow related to the calculation from the memory info.

gmarkall:
It is possible to launch more than `2**31` threads, so `cuda.grid()` and `cuda.gridsize()` need to use 64 bit integers.

Fixes #9229.

### DIFF
--- a/numba/cuda/tests/cudapy/test_intrinsics.py
+++ b/numba/cuda/tests/cudapy/test_intrinsics.py
@@ -440,6 +440,29 @@ class TestCudaIntrinsic(CUDATestCase):
         compiled[nctaid, ntid](ary)
         self.assertEqual(ary[0], nctaid * ntid)
 
+    def test_issue_9229(self):
+        # Ensure that grid and grid size are correct - #9229 showed that they
+        # overflowed an int32.
+        @cuda.jit
+        def f(grid_error, gridsize_error):
+            i1 = cuda.grid(1)
+            i2 = cuda.blockIdx.x * cuda.blockDim.x + cuda.threadIdx.x
+            gs1 = cuda.gridsize(1)
+            gs2 = cuda.blockDim.x * cuda.gridDim.x
+            if i1 != i2:
+                grid_error[0] = 1
+            if gs1 != gs2:
+                gridsize_error[0] = 1
+
+        grid_error = np.zeros(1, dtype=np.uint64)
+        gridsize_error = np.zeros(1, dtype=np.uint64)
+
+        # A lerge enough grid for thread IDs to overflow an int32
+        f[22121216, 256](grid_error, gridsize_error)
+
+        self.assertEqual(grid_error[0], 0)
+        self.assertEqual(gridsize_error[0], 0)
+
     @skip_on_cudasim('Tests PTX emission')
     def test_selp(self):
         sig = (int64[:], int64, int64[:])
--- a/numba/cuda/intrinsics.py
+++ b/numba/cuda/intrinsics.py
@@ -15,9 +15,9 @@ from numba.cuda.extending import intrinsic
 def _type_grid_function(ndim):
     val = ndim.literal_value
     if val == 1:
-        restype = types.int32
+        restype = types.int64
     elif val in (2, 3):
-        restype = types.UniTuple(types.int32, val)
+        restype = types.UniTuple(types.int64, val)
     else:
         raise ValueError('argument can only be 1, 2, 3')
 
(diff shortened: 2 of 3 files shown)


gmarkall:
gpuci run tests

numba/cuda/tests/cudapy/test_intrinsics.py
```diff
@@ -440,6 +440,30 @@ def test_simple_gridsize1d(self):
...
+            gs2 = cuda.blockDim.x * cuda.gridDim.x
+            if i1 != i2:
+                grid_error[0] = 1
+            if gs1 != gs2:
+                gridsize_error[0] = 1
+
+        grid_error = np.zeros(1, dtype=np.uint64)
+        gridsize_error = np.zeros(1, dtype=np.uint64)
+
+        # A lerge enough grid for thread IDs to overflow an int32
```
stuartarchibald, line 461:
```suggestion
        # A large enough grid for thread IDs to overflow an int32
```

numba/cuda/tests/cudapy/test_intrinsics.py
```diff
@@ -440,6 +440,30 @@ def test_simple_gridsize1d(self):
...
+        f[22121216, 256](grid_error, gridsize_error)
```
stuartarchibald, line 462: Should the blocks per grid be more easily related to the size of an int32 given TPB is 256?

stuartarchibald (commented):
Thanks for the patch and for fixing this @gmarkall. Couple of minor comments to resolve else looks good.

numba/cuda/tests/cudapy/test_intrinsics.py
gmarkall: I felt the way I'd written it (taken in conjunction with the comment on the previous line) was sufficient - if you'd like a specific change could you please suggest what you have in mind?

gmarkall:
gpuci run tests

numba/cuda/tests/cudapy/test_intrinsics.py
```diff
@@ -440,6 +440,30 @@ def test_simple_gridsize1d(self):
...
+            gs2 = cuda.blockDim.x * cuda.gridDim.x
+            if i1 != i2:
+                grid_error[0] = 1
+            if gs1 != gs2:
+                gridsize_error[0] = 1
+
+        grid_error = np.zeros(1, dtype=np.uint64)
+        gridsize_error = np.zeros(1, dtype=np.uint64)
+
+        # A large enough grid for thread IDs to overflow an int32
```
gmarkall, line 461:
```suggestion
        # A large enough grid for thread IDs to overflow an int32
        # (22121216 * 256 = 5663031296, which is greater than 2 ** 32)
```

@stuartarchibald How about that above?

numba/cuda/tests/cudapy/test_intrinsics.py
stuartarchibald: I'd guess at `(2 ** 32 // 256) + 1` but a comment suggesting the source of the current number would be ok too (I couldn't work out why that specific number was there, but realise now it's from the reproducer in the ticket). Thanks!

gmarkall:
gpuci run tests

stuartarchibald (approved):
Thanks for the patch and fixes. As discussed OOB skipping a build farm run conditional on `gpuCI` passing as this patch is very specific about integer sizes.

gmarkall:
@stuartarchibald Looks like CI is OK.

stuartarchibald:
> @stuartarchibald Looks like CI is OK.

Yes, looks good, thanks!

Issue #9182, opened 2023-09-05
RecursionError in parfor prelowering


sklam:
## Reporting a bug

- [x] I have tried using the latest released version of Numba (most recent is
 visible in the change log (https://github.com/numba/numba/blob/main/CHANGE_LOG).
- [x] I have included a self contained code sample to reproduce the problem.
  i.e. it's possible to run as 'python bug.py'.

Original discourse thread: https://numba.discourse.group/t/maximum-recursion-depth-exceeded-while-using-prange/2107

Reproducer:

1. Get pyreason:
    ```
    git clone https://github.com/lab-v2/pyreason
    cd pyreason
    git checkout 80fac257a0981cd758927b8c0232f2c37451976d
    pip install .
    ```
2. Use Numba: b96eab17f817b59075ce84b4911842b5d9a59c9c

3. Run this script:
    ```python
    from pyreason.scripts.interpretation.interpretation_parallel import _ground_node_rule
    from numba.types import DictType, unicode_type, UniTuple, ListType, boolean
    from pyreason.scripts.numba_wrapper.numba_types.rule_type import RuleType
    from pyreason.scripts.numba_wrapper.numba_types.world_type import WorldType

    Rule = RuleType()
    World = WorldType()
    bool = boolean

    _ground_node_rule.compile(
        (Rule, DictType(unicode_type, World), DictType(UniTuple(unicode_type, 2), World), ListType(unicode_type), DictType(unicode_type, ListType(unicode_type)), DictType(unicode_type, ListType(unicode_type)), bool, bool, ListType(unicode_type))
    )
    ```

RecursionError is coming from https://github.com/numba/numba/blob/b96eab17f817b59075ce84b4911842b5d9a59c9c/numba/parfors/parfor.py#L3830-L3833 because there is a cycle in `defs`. Printing `var.name` in `lookup()` shows:

```
....
clause_var_1.5
clause_var_1.6
clause_var_1.7
clause_var_1.3
clause_var_1.4
clause_var_1.5
clause_var_1.6
clause_var_1.7
clause_var_1.3
clause_var_1.4
clause_var_1.5
clause_var_1.6
clause_var_1.7
clause_var_1.3
clause_var_1.4
clause_var_1.5
```

[...]

PR #9244, opened 2023-10-16
Fix Recursion error in parfor lookup

sklam:
Fixes: https://github.com/numba/numba/issues/9182#issuecomment-1708717510

### DIFF
--- a/docs/upcoming_changes/9244.bug_fix.rst
+++ b/docs/upcoming_changes/9244.bug_fix.rst
@@ -1,0 +1,26 @@
+Fixed ``RecursionError`` in ``prange`` 
+======================================
+
+A problem with certain loop patterns using ``prange`` leading to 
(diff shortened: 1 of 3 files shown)


DrTodd13 (approved):

dieda1x1:
I also run into this bug _without_ having any recursive function in my code. Saved my life: I have modified the lines proposed by @sklam and it works like charm. This is just a contribition to the question: is this bugfix relevant? Yes it is.

dieda1x1:
On another note, after like 350 times running the same parallel code (18 threads, i7, about 15 mins) I only once got a crash: 


  File "...\lib\site-packages\numba\parfors\parfor.py", line 3849, in <genexpr>
    in_vars = set(lookup(v, True).name for v in rhs.list_vars())
AttributeError: Failed in nopython mode pipeline (step: parfor prelowering)
'NoneType' object has no attribute 'name'

Issue #9204, opened 2023-09-19
InlineInlineables pass maybe duplicating work. Needs investigating


sklam:
@dlee992 mentioned in 9/19/2023 public meeting that `InlineInlineables` pass is observed to cause duplicated compilation of the same function.

dlee992:
I can describe more details: I have a call chain like, `A (not inlined) calls B (inlined)`, I saw several times when numba is handling `inline_inlinables` untyped pass for `A`, but each time, `A` has different argument types (most are same, but some are not the same). 

And in the `inline_inlinables` passes for `A` and `B`, `A` has exact numba types showed on the graph, `B` always only has args like this `(pyobject, pyobject)`.

Maybe somehow based on this description, we can make up a simple reproducer.

dlee992:
And a following question: if I have a callchain like this `A (not inlined) calls B(inlined), B calls C (not inlined)`, does this have some known drawbacks? I had this kind of code before, and somehow I felt it could be related to this issue.

sklam:
I have received a reproducer that highlighted the compile time problem with InlineInlineable from @Hardcode84 . Using `line_profiler`, I found that all the time were spending in `copy_ir` as @stuartarchibald predicted:

```
Total time: 2.19021 s
numba/numba/core/inline_closurecall.py
Function: inline_ir at line 360

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   360                                               @profile
   361                                               def inline_ir(self, caller_ir, block, i, callee_ir, callee_freevars,
   362                                                             arg_typs=None):
   363                                                   """ Inlines the callee_ir in the caller_ir at statement index i of block
   364                                                   `block`, callee_freevars are the free variables for the callee_ir. If
   365                                                   the callee_ir is derived from a function `func` then this is
   366                                                   `func.__code__.co_freevars`. If `arg_typs` is given and the InlineWorker
   367                                                   instance was initialized with a typemap and calltypes then they will be
   368                                                   appropriately updated based on the arg_typs.
   369                                                   """
   370                                           
   371                                                   # Always copy the callee IR, it gets mutated
   372         3          2.0      0.7      0.0          def copy_ir(the_ir):
   373                                                       kernel_copy = the_ir.copy()
   374                                                       kernel_copy.blocks = {}
   375                                                       for block_label, block in the_ir.blocks.items():
   376                                                           new_block = copy.deepcopy(the_ir.blocks[block_label])
[...]

PR #9245, opened 2023-10-16
Fix slow InlineInlineable

sklam:
Fixes https://github.com/numba/numba/issues/9204

Avoid deepcopying Loc in `Var.__deepcopy__.` [This fixes slow IR copying in `InlineInlineable`](https://github.com/numba/numba/issues/9204#issuecomment-1765096208).

In one case, time spend in `InlineInlineable` went from 1second to <300ms.

### DIFF
--- a/docs/upcoming_changes/9245.performance.rst
+++ b/docs/upcoming_changes/9245.performance.rst
@@ -2,4 +2,4 @@ Improvement to IR copying speed
 ===============================
 
 Improvements were made to the deepcopying of ``FunctionIR``. 
-In one case, ``InlineInlineable`` pass is 3x faster.
+In one case, the ``InlineInlineables`` pass is 3x faster.
--- a/docs/upcoming_changes/9245.performance.rst
+++ b/docs/upcoming_changes/9245.performance.rst
@@ -1,0 +1,5 @@
+Improvement to IR copying speed
+===============================
+
+Improvements were made to the deepcopying of ``FunctionIR``. 
+In one case, ``InlineInlineable`` pass is 3x faster.
--- a/numba/core/ir.py
+++ b/numba/core/ir.py
@@ -1128,6 +1128,11 @@ class Var(EqualityCheckMixin, AbstractRHS):
         """
         return self.versioned_names | {self.unversioned_name,}
 
+    def __deepcopy__(self, memo):
+        out = Var(copy.deepcopy(self.scope, memo), self.name, self.loc)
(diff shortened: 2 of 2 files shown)


stuartarchibald:
Thanks for the patch. Borrowing the `_gen_involved` function from `numba.tests.test_ir_inlining` and modifying it slightly demonstrates the issue and fix.
```python
from numba import njit
import numpy as np
import timeit

# this forces the inliner to recursively inline 5x then stop.
_count = 0
def cost_model(*args):
    global _count
    if _count < 5:
        _count += 1
        return True
    return False

# used in _gen_involved
_GLOBAL = 1234

def _gen_involved():
    _FREEVAR = 0xCAFE

    @njit(inline=cost_model)
        f = a + b
        a += _FREEVAR
        h = f + g
        i = 1j / d
        # For SSA, zero init, n and t
        n = 0
        t = 0
        if np.abs(i) > 0:
            k = h / i
            l = np.arange(1, c + 1)
            m = np.sqrt(l - g) + e * k
            if np.abs(m[0]) < 1:
                for o in range(a):
                    n += 0
                    if np.abs(n) < 3:
                        break
                n += m[2]
            p = g / l
            q = []
            for r in range(len(p)):
                q.append(p[r])
                if r > 4 + 1:
                    s = 123
                    t = 5
                    if s > 122 - c:
                        t += s
                t += q[0] + _GLOBAL

        fortran = int((f + o + r + t + r + a + n).real)
        if fortran < 0:
            return foo(fortran, fortran, fortran, fortran, fortran)
        else:
            fortran


    return foo

involved = _gen_involved()

@njit
def foo():
    involved(1, 1, 1, 1, 1)

print(timeit.timeit(lambda: foo.compile(()), number=10))
```
Locally, this patch makes the above compile about 2x faster.

docs/upcoming_changes/9245.performance.rst
```diff
@@ -0,0 +1,5 @@
+Improvement to IR copying speed
+===============================
+
+Improvements were made to the deepcopying of ``FunctionIR``. 
+In one case, ``InlineInlineable`` pass is 3x faster.
```
stuartarchibald, line 5:
```suggestion
In one case, the ``InlineInlineables`` pass is 3x faster.
```

stuartarchibald (commented):
Thanks for the patch. There's a typo to fix else looks good.

stuartarchibald (approved):

PR #9246, opened 2023-10-17
Support for Python 3.12

esc:
As title. PR is subject to rebase, please be mindful when basing any work of this.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -12,16 +12,16 @@ jobs:
     name: macOS
     vmImage: macos-11
     matrix:
-      py38_np122:
-        PYTHON: '3.8'
-        NUMPY: '1.22'
-        CONDA_ENV: 'azure_ci'
-        TEST_START_INDEX: 0
-      py311_np126:
+      py311_np125:
         PYTHON: '3.11'
-        NUMPY: '1.26'
+        NUMPY: '1.25'
         CONDA_ENV: 'azure_ci'
         TEST_THREADING: 'tbb'
+        TEST_START_INDEX: 0
+      py12_np126:
+        PYTHON: '3.12'
+        NUMPY: '1.26'
+        CONDA_ENV: 'azure_ci'
         TEST_START_INDEX: 1
 
 - template: buildscripts/azure/azure-linux-macos.yml
@@ -47,17 +47,7 @@ jobs:
         CONDA_ENV: azure_ci
         TEST_THREADING: workqueue
         TEST_START_INDEX: 4
-      py38_np123_svml:
-        PYTHON: '3.8'
-        NUMPY: '1.23'
-        CONDA_ENV: azure_ci
-        TEST_SVML: yes
-        TEST_START_INDEX: 5
-      py38_np124:
-        PYTHON: '3.8'
-        NUMPY: '1.24'
-        CONDA_ENV: azure_ci
-        TEST_START_INDEX: 6
+
       py39_np122_cov_doc:
         PYTHON: '3.9'
         NUMPY: '1.22'
@@ -66,58 +56,58 @@ jobs:
         RUN_FLAKE8: yes
         RUN_MYPY: yes
         BUILD_DOC: yes
-        TEST_START_INDEX: 7
+        TEST_START_INDEX: 5
       py39_np123_typeguard:
         PYTHON: '3.9'
         NUMPY: '1.23'
         CONDA_ENV: azure_ci
         RUN_TYPEGUARD: yes
-        TEST_START_INDEX: 8
+        TEST_START_INDEX: 6
       py39_np126:
         PYTHON: '3.9'
         NUMPY: '1.26'
         CONDA_ENV: azure_ci
-        TEST_START_INDEX: 9
+        TEST_START_INDEX: 7
       py39_np125:
         PYTHON: '3.9'
         NUMPY: '1.25'
         CONDA_ENV: azure_ci
-        TEST_START_INDEX: 10
+        TEST_START_INDEX: 8
       py310_np126:
         PYTHON: '3.10'
         NUMPY: '1.26'
         CONDA_ENV: azure_ci
-        TEST_START_INDEX: 11
+        TEST_START_INDEX: 9
       py310_np123:
         PYTHON: '3.10'
         NUMPY: '1.23'
         CONDA_ENV: azure_ci
-        TEST_START_INDEX: 12
+        TEST_START_INDEX: 10
       py310_np124:
         PYTHON: '3.10'
         NUMPY: '1.24'
         CONDA_ENV: azure_ci
-        TEST_START_INDEX: 13
+        TEST_START_INDEX: 11
       py310_np125:
         PYTHON: '3.10'
         NUMPY: '1.25'
         CONDA_ENV: azure_ci
-        TEST_START_INDEX: 14
+        TEST_START_INDEX: 12
       py311_np126:
         PYTHON: '3.11'
         NUMPY: '1.26'
         CONDA_ENV: azure_ci
-        TEST_START_INDEX: 15
+        TEST_START_INDEX: 13
       py311_np124:
         PYTHON: '3.11'
         NUMPY: '1.24'
         CONDA_ENV: azure_ci
-        TEST_START_INDEX: 16
+        TEST_START_INDEX: 14
       py311_np125:
         PYTHON: '3.11'
         NUMPY: '1.25'
         CONDA_ENV: azure_ci
-        TEST_START_INDEX: 17
+        TEST_START_INDEX: 15
       # RVSDG tests
       py311_np126_rvsdg:
         PYTHON: '3.11'
@@ -126,7 +116,19 @@ jobs:
         RUN_FLAKE8: yes
         CONDA_ENV: azure_ci
         TEST_RVSDG: yes
-        TEST_START_INDEX: 20   # Not used by test script
+        TEST_START_INDEX: 16   # Not used by test script
+
+      py311_np125_svml:
+        PYTHON: '3.11'
+        NUMPY: '1.25'
+        CONDA_ENV: azure_ci
+        TEST_SVML: yes
+        TEST_START_INDEX: 17
+      py312_np126:
+        PYTHON: '3.12'
+        NUMPY: '1.26'
+        CONDA_ENV: azure_ci
+        TEST_START_INDEX: 18
 
 - template: buildscripts/azure/azure-windows.yml
   parameters:
--- a/numba/_helperlib.c
+++ b/numba/_helperlib.c
@@ -876,9 +876,11 @@ static void traceback_add(const char *funcname, const char *filename, int lineno
 #if (PY_MAJOR_VERSION == 3) && (PY_MINOR_VERSION == 12) /* 3.12 */
 error:
     _PyErr_ChainExceptions1(exc);
-#elif (PY_MAJOR_VERSION == 3) && (PY_MINOR_VERSION <= 11) /* 3.11 and below */
+#elif (PY_MAJOR_VERSION == 3) && ((PY_MINOR_VERSION == 8) || (PY_MINOR_VERSION == 9) || (PY_MINOR_VERSION == 10) || (PY_MINOR_VERSION == 11)) /* 3.11 and below */
 error:
     _PyErr_ChainExceptions(exc, val, tb);
+#else
+#error "Python major version is not supported."
 #endif
 }
 
--- a/numba/_helperlib.c
+++ b/numba/_helperlib.c
@@ -873,8 +873,13 @@ static void traceback_add(const char *funcname, const char *filename, int lineno
     Py_DECREF(frame);
     return;
 
+#if (PY_MAJOR_VERSION == 3) && (PY_MINOR_VERSION == 12) /* 3.12 */
+error:
+    _PyErr_ChainExceptions1(exc);
+#elif (PY_MAJOR_VERSION == 3) && (PY_MINOR_VERSION <= 11) /* 3.11 and below */
 error:
     _PyErr_ChainExceptions(exc, val, tb);
+#endif
 }
 
 
--- a/numba/core/bytecode.py
+++ b/numba/core/bytecode.py
@@ -452,26 +452,14 @@ class ByteCodePy312(ByteCodePy311):
             been inlined. In this case we can skip the exception blocks
             entirely along with the dead exceptions that it points to.
             A pair of exception that sandwiches these exception will
-            also be clubbed into a single exception.
+            also be merged into a single exception.
         """
 
-        stack = []
+        def pop_and_merge_exceptions(entries: list, entry_to_remove: _ExceptionTableEntry):
+            lower_entry_idx = entries.index(entry_to_remove) - 1
+            upper_entry_idx = entries.index(entry_to_remove) + 1
 
-        def pop_and_club_exceptions(entries: list, end_entry):
-            start_entry = stack.pop()
-            start_offset = start_entry.start
-            target_offset = start_entry.target
-            end_offset = end_entry.end
-
-            to_be_removed_entries = [
-                e for e in entries if (e.start >= start_offset and
-                                       e.end <= end_offset and
-                                       e.target == target_offset)
-            ]
-
-            lower_entry_idx = entries.index(to_be_removed_entries[0]) - 1
-            upper_entry_idx = entries.index(to_be_removed_entries[-1]) + 1
-
+            # Merge the upper and lower exceptions if possible.
             if lower_entry_idx >= 0 and upper_entry_idx < len(entries):
                 lower_entry = entries[lower_entry_idx]
                 upper_entry = entries[upper_entry_idx]
@@ -484,25 +472,25 @@ class ByteCodePy312(ByteCodePy311):
                         upper_entry.lasti)
                     entries.remove(upper_entry)
 
-            dead_entries = [
-                e for e in entries if e.start == target_offset
-            ]
-
-            entries = [e for e in entries if e not in to_be_removed_entries]
-            entries = [e for e in entries if e not in dead_entries]
+            # Remove the exception entry.
+            entries.remove(entry_to_remove)
+            # Remove dead exceptions, if any, that the entry above may point to.
+            entries = [e for e in entries if not e.start == entry_to_remove.target]
             return entries
 
-        for offset, inst in self.table.items():
-            # Check start of pattern, three instructions.
-            # Work out the index of the insteruction.
-            index = self.ordered_offsets.index(offset)
-            # If there is a BUILD_{LIST, MAP, SET} instruction at this
-            # location.
-            curr_inst = self.table[self.ordered_offsets[index]]
-            if curr_inst.opname in ("BUILD_LIST", "BUILD_MAP", "BUILD_SET"):
-                # Check if there are actually two more instructions beyond
-                # this instruction in the table.
-                if len(self.ordered_offsets) <= index + 2:
+        work_remaining = True
+        while work_remaining:
+            # Temporarily set work_remaining to False, if we find a pattern
+            # then work is not complete, hence we set it again to True.
+            work_remaining = False
+            for entry in entries.copy():
+                # Check start of pattern, three instructions.
+                # Work out the index of the instruction.
+                index = self.ordered_offsets.index(entry.start)
+                # If there is a BUILD_{LIST, MAP, SET} instruction at this
+                # location.
+                curr_inst = self.table[self.ordered_offsets[index]]
+                if not curr_inst.opname in ("BUILD_LIST", "BUILD_MAP", "BUILD_SET"):
                     continue
                 # Check if the BUILD_{LIST, MAP, SET} instruction is followed
                 # by a SWAP(2).
@@ -514,41 +502,24 @@ class ByteCodePy312(ByteCodePy311):
                 if not next_inst.opname == "FOR_ITER":
                     continue
 
-                # This pattern should be a start of an exception block
-                entry = [e for e in entries if e.start == offset]
-                if not entry:
+                # Check end of pattern, two instructions.
+                # Check for the corresponding END_FOR, exception table end is
+                # non-inclusive, so subtract one.
+                index = self.ordered_offsets.index(entry.end)
+                curr_inst = self.table[self.ordered_offsets[index - 1]]
+                if not curr_inst.opname == "END_FOR":
                     continue
-                entry = entry[0]
-
-                # Append the exception entry to the stack
-                stack.append(entry)
-
-            # Check end of pattern, two instructions.
-            # The end can be in another entry of exception table but the
-            # target of exception will remain same. so we check for that.
-            # Check for the corresponding END_FOR, exception table end is
-            # non-inclusive, so subtract one.
-            if stack and curr_inst.opname == "END_FOR":
                 # END_FOR must be followed by SWAP(2)
-                next_inst = self.table[self.ordered_offsets[index + 1]]
+                next_inst = self.table[self.ordered_offsets[index]]
                 if not next_inst.opname == "SWAP" and next_inst.arg == 2:
                     continue
-
-                # This pattern should be a end of an exception block and
-                # must have same target as the top of the stack
-                # exception
-                entry = [e for e in entries
-                         if e.end == next_inst.offset
-                         and e.target == stack[-1].target]
-                if not entry:
-                    continue
-                entry = entry[0]
-                # All above conditions hold, pattern was found.
-                # Now using the stack, remove all the exceptions
-                # that are between the starting entry and the end
-                # entry with he given target
-                entries = pop_and_club_exceptions(entries, end_entry=entry)
-
+                # If all conditions are met that means this exception entry
+                # is for a list/dict/set comprehension and can be removed.
+                # Also if there exist exception entries above and below this
+                # entry pointing to the same target. those can be merged into
+                # a single bigger exception block.
+                entries = pop_and_merge_exceptions(entries, entry)
+                work_remaining = True
         return entries
 
 
--- a/numba/core/bytecode.py
+++ b/numba/core/bytecode.py
@@ -12,7 +12,7 @@ from numba.core.utils import PYVERSION
 if PYVERSION in ((3, 12), ):
     from opcode import _inline_cache_entries
     # Instruction/opcode length in bytes
-    INSTR_LEN
+    INSTR_LEN = 2
 elif PYVERSION in ((3, 8), (3, 9), (3, 10), (3, 11)):
     pass
 else:
--- a/numba/core/bytecode.py
+++ b/numba/core/bytecode.py
@@ -391,7 +391,7 @@ class ByteCodePy311(_ByteCode):
         """
         candidates = []
         for ent in self.exception_entries:
-            if ent.start <= offset <= ent.end:
+            if ent.start <= offset < ent.end:
                 candidates.append((ent.depth, ent))
         if candidates:
             ent = max(candidates)[1]
--- a/numba/core/bytecode.py
+++ b/numba/core/bytecode.py
@@ -368,22 +368,22 @@ def _fix_LOAD_GLOBAL_arg(arg):
 
 
 class ByteCodePy311(_ByteCode):
+
     def __init__(self, func_id):
         super().__init__(func_id)
-
-        def fixup_eh(ent):
-            from dis import _ExceptionTableEntry
-            # Patch up the exception table offset
-            # because we add a NOP in _patched_opargs
-            out = _ExceptionTableEntry(
-                start=ent.start + _FIXED_OFFSET, end=ent.end + _FIXED_OFFSET,
-                target=ent.target + _FIXED_OFFSET,
-                depth=ent.depth, lasti=ent.lasti,
-            )
-            return out
-
         entries = dis.Bytecode(func_id.code).exception_entries
-        self.exception_entries = tuple(map(fixup_eh, entries))
+        self.exception_entries = tuple(map(self.fixup_eh, entries))
+
+    @staticmethod
+    def fixup_eh(ent):
+        # Patch up the exception table offset
+        # because we add a NOP in _patched_opargs
+        out = dis._ExceptionTableEntry(
+            start=ent.start + _FIXED_OFFSET, end=ent.end + _FIXED_OFFSET,
+            target=ent.target + _FIXED_OFFSET,
+            depth=ent.depth, lasti=ent.lasti,
+        )
+        return out
 
     def find_exception_entry(self, offset):
         """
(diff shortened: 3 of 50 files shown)


numba/core/utils.py
```diff
@@ -751,3 +751,13 @@ def get_hashable_key(value):
...
+
+
+class threadsafe_cached_property(functools.cached_property):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._lock = threading.RLock()
+
+    def __get__(self, *args, **kwargs):
+        with self._lock:
+            return super().__get__(*args, **kwargs)
```
sklam, line 763: This change is needed due to py3.12 https://github.com/python/cpython/pull/101890 removal of the builtin lock. The reason for the removal being all instances of the class that use `cached_property` share the same lock. For Numba usecase, this `cached_property` class is mainly used for the compiler which is already under the compiler_lock, and the rare situation that some lazy properties are used outside of the compiler pipeline.

The changes here is simple and gets the job done.

esc:
This PR is now "Ready for Review". It has passed all technical requirements, that is to say, both our Anaconda internal CI system and the public CI system via Azure are as green as we would like them. There are certain limitations caused by missing packages and complete lack of ppc64le support. There are no dev builds as of yet, but developers should be able to build this branch easily from a source checkout and run it.

As for the reviews: this PR is essentially a patch stack and I will aim to keep it as such. That is to say, any review comments will not be added as new commits, but existing commits will be rebased to include the suggested modifications. As such this PR is subject to constant rebase. (Make sure you use `git fetch --all && git reset --hard @{u}` instead of `git pull` to update.) This is done for several reasons: firstly the stack is already quite large (>50 commits) and touches many different aspects of the Numba code-base. It's already a bit cumbersome to handle, and adding even more commits would only complicate matters further. Also, this way it is much easier to review, since each commit is largely orthogonal to the rest of the commits, so reviewers can review the PR one commit at a time and don't need to be overwhelmed by the complete diff of the patch stack. Lastly, when using either `git bisect` or `git blame` -- both programs will work better on a patch stack history rather than an incoherent "sausage" of commits.

TL;DR : this branch is raedy for review, please review one commit at a time, thank you.

PS: does anyone know anything about `flake8` on 3.12? I've not be able to get this to work, yet.

numba/core/bytecode.py
```diff
@@ -321,45 +361,199 @@ def get_used_globals(self):
...
+            ]
+
+            entries = [e for e in entries if e not in to_be_removed_entries]
+            entries = [e for e in entries if e not in dead_entries]
+            return entries
+
+        for offset, inst in self.table.items():
+            # Check start of pattern, three instructions.
+            # Work out the index of the insteruction.
+            index = self.ordered_offsets.index(offset)
```
sklam, line 496: performance nitpick

It's doing too many redundant `.index()` search here. Could have did this once after checking `inst.opname in in ("BUILD_LIST", "BUILD_MAP", "BUILD_SET")`.

numba/core/bytecode.py
```diff
@@ -321,45 +361,199 @@ def get_used_globals(self):
...
+                    continue
+                next_inst = self.table[self.ordered_offsets[index + 2]]
+                # Check if the SWAP is followed by a FOR_ITER
+                if not next_inst.opname == "FOR_ITER":
+                    continue
+
+                # This pattern should be a start of an exception block
+                entry = [e for e in entries if e.start == offset]
+                if not entry:
+                    continue
```
sklam, line 518: performance nitpick

If that's the case, we should just loop over all the exception-entries instead of every instructions in the loop. There must be a lot less exception-entries than instructions.

numba/core/bytecode.py
```diff
@@ -321,45 +361,199 @@ def get_used_globals(self):
...
+                e for e in entries if e.start == target_offset
+            ]
+
+            entries = [e for e in entries if e not in to_be_removed_entries]
+            entries = [e for e in entries if e not in dead_entries]
+            return entries
+
+        for offset, inst in self.table.items():
+            # Check start of pattern, three instructions.
+            # Work out the index of the insteruction.
```
sklam, line 495:
```suggestion
            # Work out the index of the instruction.
```

numba/core/bytecode.py
```diff
@@ -321,45 +361,199 @@ def get_used_globals(self):
...
+        self.exception_entries = tuple(entries)
+
+    @property
+    def ordered_offsets(self):
+        if not self._ordered_offsets:
+            # Get an ordered list of offsets.
+            self._ordered_offsets = [o for o in self.table]
+        return self._ordered_offsets
+
+    def remove_build_list_swap_pattern(self, entries):
```
sklam, line 439: Do we have a testcase for this? if not, i used the following to exercise the nested list-comp case:

```python
@njit
def foo(n):
    return [[i] + [j for j in range(sum([x for x in range(i)]))] for i in range(n)]

r = foo(10)
```

numba/core/bytecode.py
```diff
@@ -321,45 +361,199 @@ def get_used_globals(self):
...
+
+        # If this is a generator, we need to skip any exception table entries
+        # that point to the exception handler with the highest offset.
+        if func_id.is_generator:
+            # Get the exception handler with the highest offset.
+            max_exception_target = max([e.target for e in entries])
+            # Remove any exception table entries that point to that exception
+            # handler.
+            entries = [e for e in entries if e.target != max_exception_target]
+            #breakpoint()
```
sklam, line 428:
```suggestion
```

sklam (commented):
Submitting my review so far

numba/core/untyped_passes.py
```diff
@@ -1068,116 +1067,145 @@ def get_call_args(init_arg, want):
...
+                else:
+                    continue  # no getitem in this loop
+
+                ui = unroll_info(loop, literal_unroll_call, arg,
+                                 tuple_getitem)
+                literal_unroll_info.append(ui)
+            return literal_unroll_info
+
+        # 1. Collect info about the literal_unroll loops, ensure they are legal
+        literal_unroll_loops = find_unroll_loops(loops)
```
sklam, line 1199: My understanding of ee4cae1cc43613684c74c983e129e08e3d59c600 is that `literal_unroll_info` was previously the dictionary that has the collision. It is now changed into a list to prevent collision.  I agree that `test_22` should have been an unsupported case.

sklam (commented):
Remaining review up to 75db68181

numba/core/byteflow.py
```diff
@@ -854,6 +935,15 @@ def op_END_FINALLY(self, state, inst):
         blk = state.pop_block()
         state.reset_stack(blk['entry_stack'])
 
+    if PYVERSION in ((3, 12), ):
+        def op_END_FOR(self, state, inst):
+            _ = state.pop()
+            _ = state.pop()
```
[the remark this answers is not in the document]
esc: resolved in c33d907fc7821c467c0d7d49cf2e8830fac0d8de

numba/core/utils.py
stuartarchibald: Thanks for checking @sklam

numba/core/untyped_passes.py
stuartarchibald: @sklam this is correct. The algorithm essentially looked in the blocks reachable by a `literal_unroll` induced loop for _any_ getitem, without considering that a nested `literal_unroll` loop would necessarily have one loop with a getitem on its induction variable and another loop which with no getitem on its induction variable. The result would be that both loops would reference the same getitem, hence collision when used as a dictionary key. Making the case in `test_22` greatly simplifies things without material loss of functionality (there's another way to express the behaviour, i.e. `range(len(<container>))`).

numba/core/bytecode.py
```diff
@@ -154,6 +174,16 @@ def _unpack_opargs(code):
             for j in range(ARG_LEN):
                 arg |= code[i + j] << (8 * j)
             i += ARG_LEN
+            if PYVERSION in ((3, 12), ):
+                # Python 3.12 introduced cache slots. We need to account for
+                # cache slots when we determine the offset of the next opcode.
+                # The number of cache slots is specific to each opcode and can
+                # be looked up in the _inline_cache_entries dictionary.
+                i += _inline_cache_entries[op] * 2
```
stuartarchibald, line 182: Where does `2` come from? Should it be a named constant?

numba/core/bytecode.py
```diff
@@ -168,6 +198,16 @@ def _unpack_opargs(code):
         else:
             arg = None
             i += NO_ARG_LEN
+            if PYVERSION in ((3, 12), ):
+                # Python 3.12 introduced cache slots. We need to account for
+                # cache slots when we determine the offset of the next opcode.
+                # The number of cache slots is specific to each opcode and can
+                # be looked up in the _inline_cache_entries dictionary.
+                i += _inline_cache_entries[op] * 2
```
stuartarchibald, line 206: Same question about `2` being a named constant.

numba/core/bytecode.py
```diff
@@ -321,45 +361,199 @@ def get_used_globals(self):
...
+        self.exception_entries = tuple(map(self.fixup_eh, entries))
+
+    @staticmethod
+    def fixup_eh(ent):
+        # Patch up the exception table offset
+        # because we add a NOP in _patched_opargs
+        out = dis._ExceptionTableEntry(
+            start=ent.start + _FIXED_OFFSET, end=ent.end + _FIXED_OFFSET,
+            target=ent.target + _FIXED_OFFSET,
+            depth=ent.depth, lasti=ent.lasti,
```
stuartarchibald, line 386: Does `lasti` also need moving by `_FIXED_OFFSET`?

numba/_helperlib.c
```diff
@@ -872,8 +873,13 @@ static void traceback_add(const char *funcname, const char *filename, int lineno
     Py_DECREF(frame);
     return;
 
+#if (PY_MAJOR_VERSION == 3) && (PY_MINOR_VERSION == 12) /* 3.12 */
+error:
+    _PyErr_ChainExceptions1(exc);
+#elif (PY_MAJOR_VERSION == 3) && (PY_MINOR_VERSION <= 11) /* 3.11 and below */
```
stuartarchibald, line 879: This does not match the explicit declaration of supported versions pattern used elsewhere ?

numba/tests/test_looplifting.py
```diff
@@ -576,7 +577,8 @@ def foo(X):
 
         f = jit(forceobj=True)(foo)
         f(1)
-        self.assertEqual(len(f.overloads[f.signatures[0]].lifted), 1)
+        # Post 3.12 the second loop is also being lifted
+        self.assertEqual(len(f.overloads[f.signatures[0]].lifted), 2)
```
stuartarchibald, line 581: This change is "dead" by virtue of the skip?

numba/core/byteflow.py
```diff
@@ -1088,6 +1182,35 @@ def _dup_topx(self, state, inst, count):
...
+                state.append(inst, intrinsic="INTRINSIC_STOPITERATION_ERROR")
+                state.terminate()
+                return
+            if inst.arg == 5:  # UNARY_POSITIVE
+                val = state.pop()
+                res = state.make_temp()
+                state.append(inst, intrinsic="UNARY_POSITIVE",
+                             value=val, res=res)
+                state.push(res)
+                #self.op_UNARY_POSITIVE(state, inst)
```
stuartarchibald, line 1199: delete?

numba/core/byteflow.py
```diff
@@ -1088,6 +1182,35 @@ def _dup_topx(self, state, inst, count):
         for val in duped:
             state.push(val)
 
+    if PYVERSION in ((3, 12), ):
+        def op_CALL_INTRINSIC_1(self, state, inst):
+            # See https://github.com/python/cpython/blob/v3.12.0rc2/Include/
+            # internal/pycore_intrinsics.h#L3-L17C36
+            if inst.arg == 3:  # INTRINSIC_STOPITERATION_ERROR
+                state.append(inst, intrinsic="INTRINSIC_STOPITERATION_ERROR")
```
stuartarchibald, line 1190: Should the intrinsic be a string or an enum type given it's used again in `interpreter.py` in the corresponding opcode?

numba/core/bytecode.py
```diff
@@ -321,45 +361,199 @@ def get_used_globals(self):
...
+            This pattern indicates that a list/dict/set comprehension has
+            been inlined. In this case we can skip the exception blocks
+            entirely along with the dead exceptions that it points to.
+            A pair of exception that sandwiches these exception will
+            also be clubbed into a single exception.
+        """
+
+        stack = []
+
+        def pop_and_club_exceptions(entries: list, end_entry):
```
stuartarchibald, line 458:
```suggestion
        def pop_and_merge_exceptions(entries: list, end_entry):
```
Suggest doing s/club/merge/ throughout, merge is possibly a more common term for doing a combining operation?

numba/core/bytecode.py
```diff
@@ -321,45 +361,199 @@ def get_used_globals(self):
...
+            FOR_ITER
+            ...
+            END_FOR
+            SWAP(2)
+
+            This pattern indicates that a list/dict/set comprehension has
+            been inlined. In this case we can skip the exception blocks
+            entirely along with the dead exceptions that it points to.
+            A pair of exception that sandwiches these exception will
+            also be clubbed into a single exception.
```
stuartarchibald, line 453:
```suggestion
            also be merged into a single exception.
```

numba/core/interpreter.py
```diff
@@ -3164,3 +3296,31 @@ def op_LOAD_METHOD(self, *args, **kws):
...
+                return
+            raise NotImplementedError(intrinsic)
+    elif PYVERSION in ((3, 8), (3, 9), (3, 10), (3, 11)):
+        pass
+    else:
+        raise NotImplementedError(PYVERSION)
+
+
+if PYVERSION in ((3, 12), ):
+    class INTRINSIC_STOPITERATION_ERROR(AssertionError):
```
stuartarchibald, line 3321: Why does this extend from `AssertionError`? `StopIteration` MRO has `Exception` as its base?

numba/core/byteflow.py
```diff
@@ -460,6 +463,13 @@ def op_LOAD_CONST(self, state, inst):
 
     def op_LOAD_ATTR(self, state, inst):
         item = state.pop()
+        if PYVERSION in ((3, 12), ):
+            if inst.arg & 1:
+                state.push(state.make_null())
```
stuartarchibald, line 468: Does the case where `self` is involved need to be handled, could Numba extension code reach this?

numba/core/interpreter.py
```diff
@@ -1703,7 +1725,7 @@ def _insert_outgoing_phis(self):
                 val = self.get(varname)
             except ir.NotDefinedError:
                 # Hack to make sure exception variables are defined
-                assert PYVERSION == (3, 11), "unexpected missing definition"
+                assert PYVERSION >= (3, 11), "unexpected missing definition"
```
stuartarchibald, line 1728: Should this use the "explicit" pattern?

numba/core/byteflow.py
```diff
@@ -1088,6 +1182,35 @@ def _dup_topx(self, state, inst, count):
         for val in duped:
             state.push(val)
 
+    if PYVERSION in ((3, 12), ):
+        def op_CALL_INTRINSIC_1(self, state, inst):
+            # See https://github.com/python/cpython/blob/v3.12.0rc2/Include/
+            # internal/pycore_intrinsics.h#L3-L17C36
+            if inst.arg == 3:  # INTRINSIC_STOPITERATION_ERROR
```
stuartarchibald, line 1189: Would it be clearer to write this branching on the value of `inst.arg` as `if..elif...elif...else raise NotImplementedError(...)`? (and other similar uses).

stuartarchibald (commented):
Many thanks for all your work on support for Python 3.12 @esc and @kc611. I've reviewed the parts of this that I didn't influence/write directly, I think @sklam has covered those. Along with the inline review comments, some additional remarks.

* Suggest removing `hack` and `WIP` etc from comments and commit messages, this patch is nearing production.
* A number of commit messages contain the typo `byetflow`.
* e898c763 - suggest writing `fork` as `state.fork` so as to not confuse it with system `fork(2)` when searching through history later.
* c8ad5911 and 305266f might not be needed as subsequent patches fix up the same region in a different way. This needs looking at.
* A ticket needs adding for the issue with `LOAD_FAST_AND_CLEAR`.
* Does `LOAD_METHOD` exist in 3.12? `dis` suggests it's emitted as a `LOAD_ATTR` with a flag set in the argument.
* All Python types defined via C-structures appear to be missing the member `tp_watched`?

Thanks again!

numba/core/bytecode.py
esc: `2` is for byte length of an instruction/opcode/bytecode -- can make it a constant, yes

esc:
@stuartarchibald @sklam thank you for your reviews! I will hold off with the rebase until we have resolved all the review items in the queue. This is to ensure, that SHA1 IDs will remain constant throughout this phase of the review process. Once all the code review has been resolved, we will have a second round of reviewing of the commit messages and their order to fix the typos in the commit messages and the redundant commits mentioned.

numba/core/bytecode.py
esc: fixed in 9353b9a20d040d737381bdb42f22a40a9c9e46e0

numba/core/bytecode.py
esc: Fixed in 9353b9a20d040d737381bdb42f22a40a9c9e46e0

numba/core/bytecode.py
esc: Perhaps, I do wonder if this is even used by Numba. Perhaps there is a way to trigger a bug too?

numba/_helperlib.c
esc: Fixed in 7d848e1139383e9ed0f68bbc201f4008186c5cd3

numba/core/byteflow.py
esc: Fixed in aacc674beda0817792f2d9296009474c583e186d

numba/core/bytecode.py
esc: and in 101fcfb57f5b005932d7e04bd4eed33a728f0058

numba/core/bytecode.py
esc: and in 101fcfb57f5b005932d7e04bd4eed33a728f0058

esc:
@stuartarchibald 

> * All Python types defined via C-structures appear to be missing the member `tp_watched`?

This is now in: 28ab129310bb2e225297bd9d07179f909d681cd7

esc:
> * Does `LOAD_METHOD` exist in 3.12? `dis` suggests it's emitted as a `LOAD_ATTR` with a flag set in the argument.

Looks like it is a pseudo-instruction.

esc:
> > * Does `LOAD_METHOD` exist in 3.12? `dis` suggests it's emitted as a `LOAD_ATTR` with a flag set in the argument.
> 
> Looks like it is a pseudo-instruction.

Using the following diff, the 3.12 suite passed fine:

```
diff --git i/numba/core/byteflow.py w/numba/core/byteflow.py
index 6bde745a8e..8c0636fe71 100644
--- i/numba/core/byteflow.py
+++ w/numba/core/byteflow.py
@@ -1565,7 +1565,10 @@ class TraceRunner(object):
     # NOTE: Please see notes in `interpreter.py` surrounding the implementation
     # of LOAD_METHOD and CALL_METHOD.

-    if PYVERSION in ((3, 11), (3, 12)):
+    if PYVERSION in ((3, 12), ):
+        # LOAD_METHOD has become a pseudo-instruction in 3.12
+        pass
+    elif PYVERSION in ((3, 11), ):
         def op_LOAD_METHOD(self, state, inst):
             item = state.pop()
             extra = state.make_null()
```

esc:
> > > * Does `LOAD_METHOD` exist in 3.12? `dis` suggests it's emitted as a `LOAD_ATTR` with a flag set in the argument.
> > 
> > 
> > Looks like it is a pseudo-instruction.
> 
> Using the following diff, the 3.12 suite passed fine:
> 
> ```
> diff --git i/numba/core/byteflow.py w/numba/core/byteflow.py
> index 6bde745a8e..8c0636fe71 100644
> --- i/numba/core/byteflow.py
> +++ w/numba/core/byteflow.py
> @@ -1565,7 +1565,10 @@ class TraceRunner(object):
>      # NOTE: Please see notes in `interpreter.py` surrounding the implementation
>      # of LOAD_METHOD and CALL_METHOD.
> 
> -    if PYVERSION in ((3, 11), (3, 12)):
> +    if PYVERSION in ((3, 12), ):
> +        # LOAD_METHOD has become a pseudo-instruction in 3.12
> +        pass
> +    elif PYVERSION in ((3, 11), ):
>          def op_LOAD_METHOD(self, state, inst):
>              item = state.pop()
>              extra = state.make_null()
> ```

It is now: 8b556fe37f6ad04ef680dc36321f869452e0205e

azure-pipelines.yml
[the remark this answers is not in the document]
esc: The main problem here, was that, `ipykernel`, `gdb` and `gdb-pretty-print` -- but also NumPy < 1.26 were not available for 3.12 where needed, so the public CI currently does a "best effort" to test what it can. I think the current setup is a bit temporary and should be revised as when the necessary packages become available.

numba/core/bytecode.py
esc: After an OOB discussion with @stuartarchibald and @kc611 - we came to the conclusion, that Numba probably doesn't use `lasti` -- but that doesn't mean we shouldn't be consistent. The consensus was to simply add the `_FIXED_OFFSET`. 

There is also some more information here:

https://discuss.python.org/t/when-do-you-set-lasti-in-exception-tables/22021/6

esc: added via 907a60471496c81374d88ac4c91eb4e256f546c5

esc: I had to unresolve this conversation, as it turns out this change causes a significant number of failures across 3.11 and 3.12:

https://dev.azure.com/numba/numba/_build/results?buildId=16289&view=logs&jobId=0e486a9e-8d4d-585c-1131-b2203bf73f0e&j=0e486a9e-8d4d-585c-1131-b2203bf73f0e

This is largely related to exception handling and results in some issues with tracing the control-flow of the program. It may warrant some more investigation as to the root cause of this.

esc:
> * A ticket needs adding for the issue with `LOAD_FAST_AND_CLEAR`.

https://github.com/numba/numba/issues/9300

esc:
The following are now deferrd until the code review has been completed.

> * Suggest removing `hack` and `WIP` etc from comments and commit messages, this patch is nearing production.
> * A number of commit messages contain the typo `byetflow`.
> * [e898c76](https://github.com/numba/numba/commit/e898c763f91971f580340c33de37b48a4f102eb7) - suggest writing `fork` as `state.fork` so as to not confuse it with system `fork(2)` when searching through history later.
> * [c8ad591](https://github.com/numba/numba/commit/c8ad591106915aa38e47ec796ed46047dc532c70) and [305266f](https://github.com/numba/numba/commit/305266f9868898ae795cdd2a2e1bc824353f8180) might not be needed as subsequent patches fix up the same region in a different way. This needs looking at.

numba/core/byteflow.py
esc: Think this is fixed in: 5071096b731c1108ad5a906a71e77783a6924f27

numba/core/byteflow.py
esc: Yes: 5071096b731c1108ad5a906a71e77783a6924f27

numba/core/interpreter.py
esc: Yes, done in 0d3d988f9b94a81ea05fff9d5c82988a382d6ac2

numba/core/interpreter.py
```diff
@@ -2146,6 +2211,22 @@ def op_LOAD_FAST(self, inst, res):
...
+        op_LOAD_FAST_CHECK = op_LOAD_FAST
+
+        def op_LOAD_FAST_AND_CLEAR(self, inst, res):
+            const_none = ir.Const(None, loc=self.loc)
+            try:
+                srcname = self.code_locals[inst.arg]
+                self.store(value=self.get(srcname), name=res)
+            except NotDefinedError:
+                self.store(const_none, name=res)
+            self.store(const_none, name=srcname)
```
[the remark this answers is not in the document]
stuartarchibald: Fixing this is being deferred as it's likely to be quite involved, issue is tracked in #9300.

numba/core/bytecode.py
stuartarchibald: I'm reasonably sure the various comprehension tests cover this, e.g. https://github.com/numba/numba/blob/a4664180ddc91e4c8a37dd06324f4e229a76df91/numba/tests/test_comprehension.py#L314-L321

However, it would be good to add a couple of explicit tests for the most problematic cases, which were mostly to do with comprehensions inside try-blocks.

numba/core/bytecode.py
stuartarchibald: @esc this commented code needs removing, thanks.

numba/core/bytecode.py
```diff
@@ -321,45 +363,199 @@ def get_used_globals(self):
...
+        entries = [self.fixup_eh(e) for e in
+                   dis.Bytecode(func_id.code).exception_entries
+                   ]
+
+        # Remove exceptions, innermost ones first
+        # Can be done by using a stack
+        entries = self.remove_build_list_swap_pattern(entries)
+
+        # If this is a generator, we need to skip any exception table entries
+        # that point to the exception handler with the highest offset.
```
stuartarchibald, line 423: Is it always the highest offset (it seems to be in a few things I've tested)? Would it be a good idea to make sure that the `max_exception_target` is a `CALL_INTRINSIC_1(INTRINSIC_STOPITERATION_ERROR)`?

numba/core/interpreter.py
```diff
@@ -1492,6 +1509,11 @@ def init_first_block(self):
     def _iter_inst(self):
         for blkct, block in enumerate(self.cfa.iterliveblocks()):
             firstinst = self.bytecode[block.offset]
+            # If its an END_FOR instruction, the start location of block
+            # is set to start of the FOR loop, so take the location of
+            # next instruction
+            if firstinst.opname == 'END_FOR':
+                firstinst = self.bytecode[firstinst.next]
```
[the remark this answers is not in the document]
stuartarchibald: @esc this comment needs addressing. Thanks.

numba/core/interpreter.py
```diff
@@ -2273,6 +2360,13 @@ def op_SETUP_WITH(self, inst, contextmanager, exitfn=None):
         assert self.blocks[inst.offset] is self.current_block
         # Handle with
         exitpt = inst.next + inst.arg
+
+        # Python 3.12 hack for handling nested with blocks
+        if PYVERSION >= (3, 12):
+            if exitpt > self.last_active_offset:
+                # Use exception entries to figure out end of syntax block
+                exitpt = max([ex.end for ex in self.active_exception_entries
+                             if ex.target == exitpt])
```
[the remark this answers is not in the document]
stuartarchibald: @esc this comment needs addressing, thanks!

numba/core/interpreter.py
stuartarchibald: Further thought on this, should the `ir.StaticRaise` above just raise a `StopIteration`, i.e. this class isn't needed?

numba/core/bytecode.py
stuartarchibald: Thanks for capturing our OOB conversation @esc, as noted, no change needed to `lasti`.

gmarkall:
gpuci run tests

numba/core/bytecode.py
esc: Fixed in 0836b63048646a6d3d4ac91276468ea989d201a4

numba/core/bytecode.py
esc: This was fixed in 0836b63048646a6d3d4ac91276468ea989d201a4

numba/core/bytecode.py
esc: This was fixed in 0836b63048646a6d3d4ac91276468ea989d201a4

numba/core/bytecode.py
esc: Fixed in a6b849f95e1c8a9ad50041e8611f63922312c34a

numba/core/bytecode.py
esc: Fixed in 0836b63048646a6d3d4ac91276468ea989d201a4

numba/core/bytecode.py
esc: Fixed in 0836b63048646a6d3d4ac91276468ea989d201a4 -- no more use of `club` or similar vernacular.

numba/tests/test_looplifting.py
esc: Fixed in 50ba3b38d1fa7e67cd0bb13e6489d88a691a88f0

sklam: @esc, this should have been reverted back to `self.assertEqual(len(f.overloads[f.signatures[0]].lifted), 1)` instead of the line removed.

numba/core/interpreter.py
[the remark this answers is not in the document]
sklam: @esc, too much was removed.

The following lines should be kept:

https://github.com/esc/numba/commit/382a5316d2295436a3669fd81628ffc8973ddfd5#diff-bf52cc8cc7ef5a7f153acd1c35784daeba459c55aff21631e7d399b83e65ee5cL2381-L2387

sklam (commented):
Two new comment:

- https://github.com/numba/numba/pull/9246#discussion_r1406818641
- https://github.com/numba/numba/pull/9246#discussion_r1406824261. This will help with the CI failures.

esc:
Recent commits to fixup the breaking changes.

> * [Support for Python 3.12 #9246 (comment)](https://github.com/numba/numba/pull/9246#discussion_r1406818641)

b203d15e783a53a736ee53fed801493e69786f93

> * [Support for Python 3.12 #9246 (comment)](https://github.com/numba/numba/pull/9246#discussion_r1406824261). This will help with the CI failures.

6faa19f6b487e85aecfb153e15ec86f82bf79f7e

sklam (approved):

esc:
> The following are now deferrd until the code review has been completed.
> 
> > * Suggest removing `hack` and `WIP` etc from comments and commit messages, this patch is nearing production.
> > * A number of commit messages contain the typo `byetflow`.
> > * [e898c76](https://github.com/numba/numba/commit/e898c763f91971f580340c33de37b48a4f102eb7) - suggest writing `fork` as `state.fork` so as to not confuse it with system `fork(2)` when searching through history later.
> > * [c8ad591](https://github.com/numba/numba/commit/c8ad591106915aa38e47ec796ed46047dc532c70) and [305266f](https://github.com/numba/numba/commit/305266f9868898ae795cdd2a2e1bc824353f8180) might not be needed as subsequent patches fix up the same region in a different way. This needs looking at.

We decided against cleaning up the history for new,  since this was deemed to time intensive and error prone. So 3.12 support has been merged to `main` now and Numba will support 3.12 with the next release.

stuartarchibald:
Many thanks for all your efforts on this @esc and @kc611 , and for the review @sklam. Great to see this merged!

PR #9248, opened 2023-10-17
Fix automatic unwrapping of optional type

sklam:
Alternative to #9161.

The root problem of optional typed arguments not unpacking is due to lack of exception handling around the `typer`. With the fix, we don't need to specialize individual function to handle optional types.

### DIFF
--- a/numba/core/typing/templates.py
+++ b/numba/core/typing/templates.py
@@ -411,21 +411,29 @@ class CallableTemplate(FunctionTemplate):
             # problems
             raise TypingError(str(e)) from e
 
-        sig = typer(*args, **kws)
-
-        # Unpack optional type if no matching signature
-        if sig is None:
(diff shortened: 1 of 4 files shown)


guilhermeleobas (approved):
LGTM

karthikkurella:
LGTM

numba/core/typing/templates.py
```diff
@@ -411,21 +411,29 @@ def apply(self, args, kws):
...
-                return
+                if any(isinstance(x, types.Optional) for x in args):
+                    def unpack_opt(x):
+                        if isinstance(x, types.Optional):
+                            return x.type
+                        else:
+                            return x
+
+                    args = list(map(unpack_opt, args))
+                    sig = typer(*args, **kws)
```
stuartarchibald, line 432: Do the kwargs need unpacking too?

PR #9249, opened 2023-10-17
add support for checking dtypes equal

saulshanabrook:
This PR adds support for checking if two dtypes are equal.

I have code that I generate that looks like this:

```python
assert X.dtype == np.dtype(np.float64)
```

that I would like to compile with Numba. With this PR, it is able to work under no-python mode.

### DIFF
--- a/docs/upcoming_changes/9249.np_support.rst
+++ b/docs/upcoming_changes/9249.np_support.rst
@@ -1,0 +1,3 @@
+Adds support for checking if dtypes are equal.
+========================================================================================================================================================================
+Support is added for checking if two dtype objects are equal, for example `assert X.dtype == np.dtype(np.float64)`.
--- a/docs/upcoming_changes/9249.np_support.rst
+++ b/docs/upcoming_changes/9249.np_support.rst
@@ -1,3 +1,3 @@
 Adds support for checking if dtypes are equal.
-========================================================================================================================================================================
+=============================================
 Support is added for checking if two dtype objects are equal, for example `assert X.dtype == np.dtype(np.float64)`.
--- a/docs/upcoming_changes/9249.np_support.rst
+++ b/docs/upcoming_changes/9249.np_support.rst
@@ -1,3 +1,3 @@
 Adds support for checking if dtypes are equal.
-=============================================
+==============================================
 Support is added for checking if two dtype objects are equal, for example `assert X.dtype == np.dtype(np.float64)`.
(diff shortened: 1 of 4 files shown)


guilhermeleobas:
Thanks, @saulshanabrook. I've labeled this PR as ready for review.

numba/np/arrayobj.py
```diff
@@ -4150,6 +4150,14 @@ def iternext_numpy_nditer2(context, builder, sig, args, result):
     nditer.iternext_specific(context, builder, result)
 
 
+@lower_builtin(operator.eq, types.DType, types.DType)
+def dtype_eq_impl(context, builder, sig, args):
+    arg1, arg2 = sig.args
+    val = 1 if arg1 == arg2 else 0
+    res = ir.Constant(ir.IntType(1), val)
```
stuartarchibald, line 4157:
```suggestion
    res = ir.Constant(ir.IntType(1), int(arg1 == arg2))
```

docs/upcoming_changes/9249.np_support.rst
```diff
@@ -0,0 +1,3 @@
+Adds support for checking if dtypes are equal.
+========================================================================================================================================================================
```
stuartarchibald, line 2: I think this "underline" needs to be the same length as the title line?

stuartarchibald (commented):
Thanks for the patch. A couple of minor things to resolve else looks good.

Side-note: Ideally this sort of change would be written using the `numba.extending.overload` API but there's a large number of `operator.eq` implemented using the low-level extension API and it's not clear if mixing the APIs might cause problems to do with ordering/matching. Given this change is small and additive, continued use of the low-level API is acceptable.

docs/upcoming_changes/9249.np_support.rst
saulshanabrook, line 2:
```suggestion
Adds support for checking if dtypes are equal.
=============================================
```

saulshanabrook: That sounds right, fixed!

docs/upcoming_changes/9249.np_support.rst
```diff
@@ -0,0 +1,3 @@
+Adds support for checking if dtypes are equal.
+=============================================
```
stuartarchibald, line 2:
```suggestion
Adds support for checking if dtypes are equal.
==============================================
```
Think this needs one more `=` to cover the `.` on the end?

stuartarchibald (approved):

PR #9255, opened 2023-10-26
Fix SSA to consider variables whose use is not dominated by the definition

### LINKED ISSUE #9242: Broken SSA for conditionally defined var
```
@njit
def py_func(a):
    c = a > 0
    if c:
        d = a + 5

    return c and d > 0

py_func(10)
```
`NUMBA_DUMP_SSA=1` gives
```
label 0:
    a = arg(0, name=a)                       ['a']
    $const4.1 = const(int, 0)                ['$const4.1']
    c = a > $const4.1                        ['$const4.1', 'a', 'c']
    bool12 = global(bool: <class 'bool'>)    ['bool12']
    $12pred = call bool12(c, func=bool12, args=(Var(c, test_basic.py:929),), kws=(), vararg=None, varkwarg=None, target=None) ['$12pred', 'bool12', 'c']
    branch $12pred, 14, 22                   ['$12pred']
label 14:
    $const16.1 = const(int, 5)               ['$const16.1']
    d = a + $const16.1                       ['$const16.1', 'a', 'd']
    jump 22                                  []
label 22:
    bool24 = global(bool: <class 'bool'>)    ['bool24']
    $24pred = call bool24(c, func=bool24, args=(Var(c, test_basic.py:929),), kws=(), vararg=None, varkwarg=None, target=None) ['$24pred', 'bool24', 'c']
    $phi32.0 = c                             ['$phi32.0', 'c']
    branch $24pred, 26, 32                   ['$24pred']
label 26:
    $const28.1 = const(int, 0)               ['$const28.1']
    $30compare_op.2 = d > $const28.1         ['$30compare_op.2', '$const28.1', 'd']
    $phi32.0.1 = $30compare_op.2             ['$30compare_op.2', '$phi32.0.1']
    jump 32                                  []
label 32:
    $phi32.0.2 = phi(incoming_values=[Var($phi32.0.1, test_basic.py:933), Var($phi32.0, test_basic.py:933)], incoming_blocks=[26, 22]) ['$phi32.0', '$phi32.0.1', '$phi32.0.2']
    $32return_value.1 = cast(value=$phi32.0.2) ['$32return_value.1', '$phi32.0.2']
    return $32return_value.1                 ['$32return_value.1']
```
`d` is used in block 26 and defined in block 14, which doesn't dominate 26

sklam:
Fixes #9242 and #9256.

* SSA reconstruction is adjusted to also consider definitions that do not dominate all uses.
* The above changes unfortunately is not compatible with Parfors; therefore, its effect is undone when Phi is stripped.
* https://github.com/numba/numba/pull/9255/commits/80dd8df240a4b73aa3af64b1abe077897dfa23a1 is part of debugging the Parfors breakage due to the initial commit. Any user code that makes a new variable definition in a loop not in the pattern of reduction will trigger a problem in SROA-avoidance due to incompatible postcondition of use-def analysis on Parfor node.

### DIFF
--- a/numba/core/ssa.py
+++ b/numba/core/ssa.py
@@ -257,7 +257,7 @@ class _BaseHandler:
 class _GatherDefsHandler(_BaseHandler):
     """Find all defs and uses of variable in each block
 
-    ``states["label"] is a int; label of the current block
+    ``states["label"]`` is a int; label of the current block
     ``states["defs"]`` is a Mapping[str, List[Tuple[ir.Assign, int]]]
     ``states["uses"]`` is a Mapping[Set[int]]
     """
--- a/numba/core/ssa.py
+++ b/numba/core/ssa.py
@@ -258,7 +258,9 @@ class _GatherDefsHandler(_BaseHandler):
     """Find all defs and uses of variable in each block
 
     ``states["label"]`` is a int; label of the current block
-    ``states["defs"]`` is a Mapping[str, List[Tuple[ir.Assign, int]]]
+    ``states["defs"]`` is a Mapping[str, List[Tuple[ir.Assign, int]]]:
+        - a mapping of the name of the assignee variable to the assignment
+          IR node and the block label.
     ``states["uses"]`` is a Mapping[Set[int]]
     """
     def on_assign(self, states, assign):
--- a/numba/core/ssa.py
+++ b/numba/core/ssa.py
@@ -54,7 +54,7 @@ def _run_ssa(blocks):
     cfg = compute_cfg_from_blocks(blocks)
     df_plus = _iterated_domfronts(cfg)
     # Find SSA violators
-    violators = _find_defs_violators(blocks)
+    violators = _find_defs_violators(blocks, cfg)
     # Make cache for .list_vars()
     cache_list_vars = _CacheListVars()
 
@@ -142,7 +142,7 @@ def _get_scope(blocks):
     return first.scope
 
 
-def _find_defs_violators(blocks):
+def _find_defs_violators(blocks, cfg):
     """
     Returns
     -------
@@ -150,9 +150,22 @@ def _find_defs_violators(blocks):
         The SSA violators in a dictionary of variable names.
     """
     defs = defaultdict(list)
-    _run_block_analysis(blocks, defs, _GatherDefsHandler())
+    uses = defaultdict(set)
+    states = dict(defs=defs, uses=uses)
+    _run_block_analysis(blocks, states, _GatherDefsHandler())
     _logger.debug("defs %s", pformat(defs))
+    # Gather violators by number of definitions
     violators = {k for k, vs in defs.items() if len(vs) > 1}
+    # Gather violators by uses not dominated by the one def
+    doms = cfg.dominators()
+    for k, use_blocks in uses.items():
+        if k not in violators:
+            for label in use_blocks:
+                dom = doms[label]
+                def_labels = {label for _assign, label in defs[k] }
+                if not def_labels.intersection(dom):
+                    violators.add(k)
+                    break
     _logger.debug("SSA violators %s", pformat(violators))
     return violators
 
@@ -160,6 +173,7 @@ def _find_defs_violators(blocks):
 def _run_block_analysis(blocks, states, handler):
     for label, blk in blocks.items():
         _logger.debug("==== SSA block analysis pass on %s", label)
+        states['label'] = label
         for _ in _run_ssa_block_pass(states, blk, handler):
             pass
 
@@ -241,12 +255,26 @@ class _BaseHandler:
 
 
 class _GatherDefsHandler(_BaseHandler):
-    """Find all defs
+    """Find all defs and uses of variable in each block
 
-    ``states`` is a Mapping[str, List[ir.Assign]]
+    ``states["label"] is a int; label of the current block
+    ``states["defs"]`` is a Mapping[str, List[Tuple[ir.Assign, int]]]
+    ``states["uses"]`` is a Mapping[Set[int]]
     """
     def on_assign(self, states, assign):
-        states[assign.target.name].append(assign)
+        # keep track of assignment and the block
+        states["defs"][assign.target.name].append((assign, states["label"]))
+        # keep track of uses
+        for var in assign.list_vars():
+            k = var.name
+            if k != assign.target.name:
+                states["uses"][k].add(states["label"])
+
+    def on_other(self, states, stmt):
+        # keep track of uses
+        for var in stmt.list_vars():
+            k = var.name
+            states["uses"][k].add(states["label"])
 
 
 class UndefinedVariable:
--- a/numba/core/typed_passes.py
+++ b/numba/core/typed_passes.py
@@ -23,7 +23,8 @@ from numba.core.ir_utils import (raise_on_unsupported_feature, warn_deprecated,
                                  dead_code_elimination, simplify_CFG,
                                  get_definition,
                                  build_definitions, compute_cfg_from_blocks,
-                                 is_operator_or_getitem)
+                                 is_operator_or_getitem,
+                                 replace_vars)
 from numba.core import postproc
 from llvmlite import binding as llvm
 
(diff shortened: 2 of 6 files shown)


numba/core/typed_passes.py
```diff
@@ -881,6 +882,9 @@ def __init__(self):
     def run_pass(self, state):
         state.func_ir = self._strip_phi_nodes(state.func_ir)
         state.func_ir._definitions = build_definitions(state.func_ir.blocks)
+        self._simplify_conditionally_defined_variable(state.func_ir)
+        state.func_ir._definitions = build_definitions(state.func_ir.blocks)
```
sklam, line 886: TODO:

- [ ] maybe we can skip this if the function is not `parallel=True`

numba/core/typed_passes.py
```diff
@@ -956,3 +961,58 @@ def _strip_phi_nodes(self, func_ir):
...
+        """
+        any_block = next(iter(func_ir.blocks.values()))
+        scope = any_block.scope
+        defs = func_ir._definitions
+
+        # Find and simplify conditionally defined variables
+        suspects = set()
+        for varname, deflist in defs.items():
+            unique_defs = set(deflist)
+            atleast_one_undef = any(map(lambda x: x == ir.UNDEFINED,
```
stuartarchibald, line 989: Is this the same as `atleast_one_undef = ir.UNDEFINED in unique_defs`? Are there concerns about `ir.UNDEFINED` not being singleton under some condition?

sklam:
@stuartarchibald showed me this case that is producing a wrong result
```python
@njit(parallel=True)
def foo(x):
  c = x  >0
  set_num_threads(1)
  for i in prange(x):
    if c:
      d = x + 4
    c = i
  return c and d > 0
```

numba/tests/test_parfors.py
```diff
@@ -2311,6 +2311,17 @@ def test_impl(a, b):
         self.check(test_impl, 3.7, 4.3)
         self.assertEqual(countParfors(test_impl, (types.float64, types.float64)), 1)
 
+    def test_issue9256_lower_sroa_conflict(self):
```
stuartarchibald, line 2314: A variant on this test:
```python
from numba import njit, set_num_threads, prange
import numpy as np

def def_in_loop(x):
    c = x
    set_num_threads(1)
    for i in prange(x):
        if c:
            d = x + 4
    return c and d > 0

print(def_in_loop(4))
print(njit(def_in_loop)(4))
print(njit(parallel=True)(def_in_loop)(4))
```
produces:
```
True # python
1   # njit
0 # parfors
```
On `main` the same code sample hits the issue in #9256, so this patch appears to fix some of the issues in that it now compiles.

stuartarchibald (commented):
Thanks for the patch @sklam, this must have involved some complex debugging to produce a fix. I've given this an initial review, in part to just understand what the issues are and how they are being fixed. I think this patch fixes the issues linked to in the PR summary but note that a variant (in a review comment) is failing. There's also some minor issue with exception messages that's making CI fail. Thanks again!

numba/tests/test_parfors.py
sklam: The "undoing" code will need to be more involved to handle this case, which make 3 SSA variables for `d`:

```

             'd': [x + $const32.2],
             'd.1': [null(), Var(d.2, test_variant.py:10)],
             'd.2': [Var(d.1, test_variant.py:10), Var(d, test_variant.py:9)],
```

🤔

sklam: The exact code is producing an error on main: https://github.com/numba/numba/issues/9265

A variant of the test that works is added in [821cc2c](https://github.com/numba/numba/pull/9255/commits/821cc2cbbcf34d46e4c2cb51cb49de8e2e497f75).

sklam:
It's starting to affect `numba.tests.test_mixed_tuple_unroller.TestMixedTupleUnroll.{test_27 , test_28, test_29}`

numba/core/typed_passes.py
```diff
@@ -956,3 +961,56 @@ def _strip_phi_nodes(self, func_ir):
...
+            ...
+            ver1 = ver
+            ...
+            uses(ver1)
+
+        into:
+            # delete all assignments to ver1
+            uses(ver)
+
+        This is only needed for parfors.
```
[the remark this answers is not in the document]
sklam: done in a56dc4fca27e2cf3ef1f664ed9baa314081e7b33

numba/core/typed_passes.py
sklam: This part is now changed to https://github.com/numba/numba/pull/9255/files#diff-8da448c689273bcc37cdbc30f509dc9bc0a582921225308874101d173dd69eb3R994

stuartarchibald:
> It's starting to affect `numba.tests.test_mixed_tuple_unroller.TestMixedTupleUnroll.{test_27 , test_28, test_29}`

@sklam RE: ^. Does this PR depend on resolving #9269 ?

sklam:
> > It's starting to affect `numba.tests.test_mixed_tuple_unroller.TestMixedTupleUnroll.{test_27 , test_28, test_29}`
> 
> @sklam RE: ^. Does this PR depend on resolving #9269 ?

@stuartarchibald, No, i worked around the problem by: https://github.com/numba/numba/pull/9255/commits/c79f982d85e4d0cf566ffe4da1bc67a2ef02d064

stuartarchibald:
> > > It's starting to affect `numba.tests.test_mixed_tuple_unroller.TestMixedTupleUnroll.{test_27 , test_28, test_29}`
> > 
> > 
> > @sklam RE: ^. Does this PR depend on resolving #9269 ?
> 
> @stuartarchibald, No, i worked around the problem by: [c79f982](https://github.com/numba/numba/commit/c79f982d85e4d0cf566ffe4da1bc67a2ef02d064)

Thanks for confirming.

stuartarchibald (commented):
Thanks for the updates. There's a few minor things to look at inline. I've also tested this some more manually, this case seems to fail, it's basically the same as an existing test case but with a duplicated loop:
```python
@njit(parallel=True)
def def_in_loop(x):
    c = x
    set_num_threads(1)
    for _i in prange(x):
        if c:
            d = x + 4
    for _i in prange(x):
        if c:
            d = x + 4
    return c, d > 0

print(def_in_loop(10))
print(def_in_loop.py_func(10))
```
which gives:
```
(10, False)
(10, True)
```

numba/core/ssa.py
```diff
@@ -241,12 +255,26 @@ def on_other(self, states, stmt):
 
 
 class _GatherDefsHandler(_BaseHandler):
-    """Find all defs
+    """Find all defs and uses of variable in each block
 
-    ``states`` is a Mapping[str, List[ir.Assign]]
+    ``states["label"]`` is a int; label of the current block
+    ``states["defs"]`` is a Mapping[str, List[Tuple[ir.Assign, int]]]
```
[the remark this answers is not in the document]
sklam: Added more comment at 55028233b26eb9adf53015d9722fbf240ed423ba.

> does states need to be a dict, would a namedtuple work, looks like it's just a container of a consistent 3 things?

it can be a namedtuple with some refactoring. For now, everything `_BaseHandler` subclass has different requirements for what is in `states`.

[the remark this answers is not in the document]
stuartarchibald: ah ok, let's leave this as-is then, thanks for looking.

stuartarchibald:
> RE: [#9255 (review)](https://github.com/numba/numba/pull/9255#pullrequestreview-1755297593)
> 
> It's a pre-existing condition on `main`. My patch will not affect the `d` in this case because it has more than one definition. In fact, I verified that it is not a problem once SSA is turned off on `main`.

Many thanks for looking into this. As per OOB discussion, conclusion was that this patch is fixing the most it can with respect to the original objectives (tickets #9242 and #9256).

stuartarchibald (approved):

numba/parfors/parfor_lowering.py
```diff
@@ -52,6 +52,15 @@ def lower_inst(self, inst):
...
             super().lower_inst(inst)
 
+    @property
+    def _disable_sroa_like_opt(self):
+        """
+        Force disable this because Parfor use-defs is incompatible---it only
+        considers use-defs in blocks that must be executing.
+        See https://github.com/numba/numba/commit/017e2ff9db87fc34149b49dd5367ecbf0bb45268
+        """
+        return True
```
stuartarchibald, line 62: After thought on this. It's possible that this may increase compilation times, however it prevents miscompilation. It's better to work from a position of correctness towards performance...!

Issue #9257, opened 2023-10-28
DictType compiling hangs with NumPy unicode strings tuple key


Azmisov:
I'm trying to use a tuple of numpy unicode types, and compiling hangs and code doesn't run. A non-tuple numpy unicode key does work however.

```python
import numpy as np
from numba import int32, njit
from numba.types import UnicodeCharSeq, Tuple
from numba.typed import Dict

KeyType = Tuple((UnicodeCharSeq(6), UnicodeCharSeq(10)))

@njit
def test(rec) -> bool:
	d = Dict.empty(key_type=KeyType, value_type=int32)
	key = (rec["a"], rec["b"])
	return key in d
	
recs = np.array([("012345","0123456789")], dtype=[("a","<U6"),("b","<U10")])
test(recs[0])
```

This is with Python 3.11.5, Numba 0.58.1

sklam:
Thanks for the bug report. From the reproducer, I can see the compiler retrying to infer type on the same function indefinitely.

sklam:
There are 3 problems in Numba:

1) `UnicodeCharSeq` is castable in typing but not in lowering.

https://github.com/numba/numba/pull/4516/commits/7935c1a457ff3cc410c3591835b0992cf435ea69 removed the restriction for casting between `UnicodeCharSeq` of different sizes. 

2) cast in `typed_getitem()` should be a non-recoverable error.

The reproducer is failing to compile because of (1) at https://github.com/numba/numba/blob/452eb5755320506de96bed83021566d0093a3d3a/numba/cpython/tupleobj.py#L274-L275. If the failure is a hard error, the compiler will not retry indefinitely.

3) `BaseFunction.get_call_type()` should remember failed typing to avoid repeated fail attempt

The following diff allow the compiler the fail quickly.
```diff
diff --git a/numba/core/types/functions.py b/numba/core/types/functions.py
index e67a7f335..b760e8a4b 100644
--- a/numba/core/types/functions.py
+++ b/numba/core/types/functions.py
@@ -257,6 +257,7 @@ class BaseFunction(Callable):
         self._impl_keys = {}
         name = "%s(%s)" % (self.__class__.__name__, self.typing_key)
         self._depth = 0
+        self._failed = {}
         super(BaseFunction, self).__init__(name)

     @property
@@ -280,6 +281,9 @@ class BaseFunction(Callable):

     def get_call_type(self, context, args, kws):

[...]

PR #9259, opened 2023-10-31
Stop type inference from indefinite retry on failed function resolution

sklam:


sklam:
TODOs:

- [X] check https://github.com/numba/numba/issues/9375#issuecomment-1893955887 (edit by @gmarkall - it fixes it, see https://github.com/numba/numba/pull/9259#issuecomment-2962439967)
- [ ] (added by @gmarkall) check why `numba.tests.test_dispatcher.TestNoRetryFailedSignature.test_error_count` is not an unexpected success with this PR.

gmarkall:
@sklam I was curious about the status of this PR after seeing the xfail in `numba.tests.test_dispatcher.TestNoRetryFailedSignature.test_error_count` and looking into simplifying `_FunctionCache` in #10112, so I've merged `main` to see what happens on CI.

It looks like this was basically ready, in the sense of no known issues being present - is that correct?

gmarkall:
With this PR, the reproducer in #9257 fails quickly. It does still produce a huge traceback, but that is tangential to what is being addressed here.

I do wonder if this will speed up Numba compilation for other, successful use cases, given that fewer retries will be required in general.

gmarkall:
> check why `numba.tests.test_dispatcher.TestNoRetryFailedSignature.test_error_count` is not an unexpected success with this PR.

I suspect the test was not quite right. Without this PR, it fails with:

```
$ python -m numba.runtests numba.tests.test_dispatcher.TestNoRetryFailedSignature.test_error_count
F
FAIL: test_error_count (numba.tests.test_dispatcher.TestNoRetryFailedSignature.test_error_count)
  File "/home/gmarkall/numbadev/numba/numba/tests/test_dispatcher.py", line 1112, in test_error_count
    self.assertEqual(ct_bad, 1)
AssertionError: 40 != 1

Ran 1 test in 0.472s
```

With this PR, the failure is:

```
$ python -m numba.runtests numba.tests.test_dispatcher.TestNoRetryFailedSignature.test_error_count
F
FAIL: test_error_count (numba.tests.test_dispatcher.TestNoRetryFailedSignature.test_error_count)
  File "/home/gmarkall/numbadev/numba/numba/tests/test_dispatcher.py", line 1112, in test_error_count
    self.assertEqual(ct_bad, 1)
AssertionError: 2 != 1

Ran 1 test in 0.455s

FAILED (failures=1)
```

Which is 38 fewer retries! So, a definite improvement - I just need to check why there are two visits instead of one.

gmarkall:
>  I just need to check why there are two visits instead of one.

There were two calls, one for unliteral-ed arguments, and one for non-unliteral-ed arguments. I've pushed a fix to this branch.

docs/source/reference/envvars.rst
```diff
@@ -161,6 +161,15 @@ These variables influence what is printed out during compilation of
 
    If set to non-zero, print out debugging information about type inference.
 
+.. envvar:: NUMBA_DISABLE_TYPEINFER_FAIL_CACHE
```
gmarkall, line 164: I've just tested the env var - it works, and disabling the cache causes relevant test failures to occur as expected.

numba/core/typing/context.py
```diff
@@ -99,6 +119,66 @@ def match(self, py_func, args):
...
+        """Lookup resolution cache for the given function type and argument
+        types.
+        """
+        if not self._stack or config.DISABLE_TYPEINFER_FAIL_CACHE:
+            # if callstack is empty, bypass fail_cache
+            return _ResolveCache()
+
+        def normalize_dict(obj):
+            if isinstance(obj, dict):
+                return tuple(kws.items())
```
gmarkall: IIRC, the iteration order of a given dict is stable, but can two dicts with the same keys but different insertion orders have different iteration orders? Should we sort the items by their key here in case there are two calls `f(a=1, b=2)` and `f(b=2, a=1)` that have the same set of `kws` but not the same `items()`?

numba/core/typing/context.py
```diff
@@ -99,6 +119,66 @@ def match(self, py_func, args):
...
+        def hashable(obj):
+            try:
+                hash(obj)
+            except TypeError:
+                return False
+            else:
+                return True
+
+        key = func, args, normalize_dict(kws)
+        if not hashable(key):
```
gmarkall, line 144: A minor nit, but is there any reason not to use `isinstance(key, collections.abc.Hashable)` here?

docs/upcoming_changes/9259.bug_fix.rst
```diff
@@ -0,0 +1,20 @@
+Avoid type inference hangs
+==========================
```
gmarkall: I think this is what towncrier is complaining about:

```suggestion
--------------------------
```

numba/core/typing/context.py
sklam: `isinstance(obj, Hashable)` does not check elements in containers. It can accidentally allow types that have a `Hashable` interface but fail on `hash()`; e.g.

```python

In [3]: isinstance(([],), Hashable)
Out[3]: True

In [4]: hash(([],))
TypeError                                 Traceback (most recent call last)
Cell In[4], line 1
----> 1 hash(([],))

TypeError: unhashable type: 'list'
```

But maybe it's just need better normalization.

sklam:
> check why `numba.tests.test_dispatcher.TestNoRetryFailedSignature.test_error_count` is not an unexpected success with this PR.

 @gmarkall, this PR puts in the needed cache to prevent the repeated compilation. See notes in ddd40cd1e85d517844e79494c3689356780012f0.

numba/core/typing/context.py
sklam: I think we should sort

sklam:
@gmarkall, your changes look good. I have made the change to put sorted in `normalize_dict` https://github.com/numba/numba/pull/9259/commits/5a72dc0741d07add016e065182fe3765fce13ff2

gmarkall:
> > check why `numba.tests.test_dispatcher.TestNoRetryFailedSignature.test_error_count` is not an unexpected success with this PR.
> 
> @gmarkall, this PR puts in the needed cache to prevent the repeated compilation. See notes in [ddd40cd](https://github.com/numba/numba/commit/ddd40cd1e85d517844e79494c3689356780012f0).

I understand... However the issue was that the test was checking for one attempt to compile, when it should have checked for 2 (unliteral and literal) - this is why it was still failing and therefore "expected". However, it seems like we are aligned on the test as it is now anyway so I don't think any further change is needed.

gmarkall (approved):

PR #9261, opened 2023-11-01
Workaround for tempfile.TemporaryFile hanging indefinitely on Windows

### LINKED ISSUE #8755: ensure_cache_path() for Windows directory permission check may cause infinite loop
## Reporting a bug

This problem only occurs in Windows. To reproduce this bug, a simple `bug.py` could be written like this:
```
from numba import njit
@njit(cache=True)
def f():
    pass
```
Running this alone is totally fine. But if you toggle the privileges of the `__pycache__` folder, this code will hang _almost_ forever.

[ensure_cache_path()](https://github.com/numba/numba/blob/b89199e9323f3f801c35fa0f20524833c18f46c4/numba/core/caching.py#L106#L107)  in numba/core/caching.py uses `tempfile.TemporaryFile()` to check if writable. However there is a long-time bug of the Python built-in package [tempfile](https://github.com/python/cpython/issues/66305). This Python's own bug haven't been solved for ten years and confused many([example](https://stackoverflow.com/questions/55109076/python-tempfile-temporaryfile-hangs-on-windows-when-no-write-privilege)). An alternative way should be taken to ensure cache writability.

For my case, I installed Python and its packages in `%ProgramFiles%`. My program uses another package that depends on numba njit cache. If I don't choose to run as Administrator, loading that package will make me wait forever.

Huite:
This fixes #8755 for me, which is due to a nasty long-standing bug with tempfile on Windows in CPython: https://github.com/python/cpython/issues/66305

There is more use of tempfile within numba, but that does not seem to be causing problems. This likely warrants more research and better testing, but I don't think many maintainers have the time time currently.

Having said that, this minor change seems easily worth it it stops numba from hanging indefinitely on Windows.

### DIFF
--- a/numba/core/caching.py
+++ b/numba/core/caching.py
@@ -104,7 +104,11 @@ class _CacheLocator(metaclass=ABCMeta):
         path = self.get_cache_path()
         os.makedirs(path, exist_ok=True)
         # Ensure the directory is writable by trying to write a temporary file
-        tempfile.TemporaryFile(dir=path).close()
+        # Avoid using tempfile.TemporaryFile since it may hang indefinitely on Windows:
+        # https://github.com/python/cpython/issues/66305
+        filename = os.path.join(path, "cache_write_test.txt")
+        open(filename).close()
+        os.remove(filename)
 
     @abstractmethod
     def get_cache_path(self):


Huite:
Hi @gmarkall, there doesn't seem to be much point in proceeding. Given that CI seems to fail almost completely, I'm surprised that it did something for me locally. Most likely because I started with the unaltered code to begin with, had it hang and crash, and then I tried the change.

It felt like a bit of a long shot anyway!

I'll close this PR and make some notes in the linked issue.

PR #9274, opened 2023-11-08
CUDA: Add support for compilation to LTO-IR

gmarkall:
This adds support for compiling to LTO-IR, providing an alternative route to using PTX for linking code with non-Python source code with greater potential for optimization from being able to optimize at link-time across the whole body of source for different languages.

A summary of the changes:

- The first three commits are small refactors / tidy-ups of some paths that could be simplified now we don't have to deal with NVVM 3.4 anymore (556bb82677e24f3e7ded68e4c80503a054203e09, b6c3b85d7576a0ec2a0c4d0811277d4fd7f45b57, 6d8a22ce0e84156ed2066d53c89f434822b55191)
- The next commit clarifies the compilation behaviour with / without return types being specified, which was a bit of a hole in the documentation: 357d67d8969969b97be6e6d86f9151f7ed4ec686
- Then we add support for LTO-IR to to CUDA codegen, without exposing it externally: 8d6fd900e25b53ca5642a2c8b43a26da19ab17ac
- Then we add a public interface for LTO-IR code generation: ed99b9ec687639c3eef78783cec6965e16e36dfa
- Documentation is also added / updated: 0938da4a0d4d213ffca2ab2f81de47b1951c0a1b


gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

docs/source/cuda/cuda_compilation.rst
```diff
@@ -4,30 +4,42 @@
...
+The compilation API can be used without a GPU present, as it uses no driver
+functions and avoids initializing CUDA in the process. It is invoked through
+the following function:
 
-.. autofunction:: numba.cuda.compile_ptx
+.. autofunction:: numba.cuda.compile
    :noindex:
 
-If a device is available and PTX for the compute capability of the current
+If a device is available and code for the compute capability of the current
```
stuartarchibald, line 19: Is there a word other than `code` that could be used here as the output isn't "code" in the traditional sense? However, I'm not sure that there is a term that covers PTX and LTO-IR, seems like later on in the PR `compiled code` is used.

docs/source/cuda/cuda_compilation.rst
```diff
@@ -40,22 +52,28 @@ ABI device functions requires three issues to be addressed:
...
+A simple way to address all these issues is to compile device functions with
+the C ABI instead. This results in the following:
 
-- The name of the compiled device function in PTX can be controlled. By default
-  it will match the name of the function in Python, so it is easy to determine.
-  This is the function's ``__name__``, rather than ``__qualname__``, because
-  ``__qualname__`` encodes additional scoping information that would make the
-  function name hard to predict, and in a lot of cases, an illegal identifier
-  in C.
+- The name of the compiled device function in the generated code can be
```
stuartarchibald, line 58:
```suggestion
- The name of the compiled device function in the generated output can be
```

numba/cuda/codegen.py
```diff
@@ -134,40 +137,56 @@ def _get_ptxes(self, cc=None):
...
         if cubin:
             return cubin
 
         linker = driver.Linker.new(max_registers=self._max_registers, cc=cc)
 
-        ptxes = self._get_ptxes(cc=cc)
-        for ptx in ptxes:
+        if linker.lto:
+            ltoir = self.get_ltoir(cc=cc)
+            linker.add_ltoir(ltoir)
```
stuartarchibald, line 185: This is untested as a dependency on `pynvjitlink` would be needed to do such a test, on inspection only it seems ok.

numba/cuda/tests/cudapy/test_userexc.py
```diff
@@ -14,6 +14,12 @@ class MyError(Exception):
 
 class TestUserExc(CUDATestCase):
 
+    def setUp(self):
+        super().setUp()
+        # LTO optimizes away the exception status due to an oversight
+        # in the way we generate it (it is not added to the used list).
```
stuartarchibald, line 20: xref: #9526

numba/cuda/tests/cudapy/test_compiler.py
```diff
@@ -192,15 +193,37 @@ def test_c_abi_with_abi_name(self):
...
+        ltoir, resty = compile(f_module, int32(int32, int32), device=True,
+                               output="ltoir")
+
+        # There are no tools to interpret the LTOIR output, but we can check
+        # that we appear to have obtained an LTOIR file. This magic number is
+        # not documented, but is expected to remain consistent.
+        LTOIR_MAGIC = 0x7F4E43ED
+        header = int.from_bytes(ltoir[:4], byteorder='little')
+        self.assertEqual(header, LTOIR_MAGIC)
+
```
stuartarchibald, line 217: Should a test be added for setting the output type to something invalid, e.g. `output='invalid'` ?

stuartarchibald, line 217: Should a test be added for compiling to a kernel with LTO-IR as output? Or does this not make any sense as a combination?

numba/cuda/tests/cudapy/test_compiler.py
```diff
@@ -192,15 +193,37 @@ def test_c_abi_with_abi_name(self):
...
+        # The function name should match the Python function name, and its
+        # return value should be 32 bits
+        self.assertRegex(ptx, r"\.visible\s+\.func\s+\(\.param\s+\.b32\s+"
+                              r"func_retval0\)\s+f_module\(")
+
+    def test_compile_to_ltoir(self):
+        if runtime.get_version() < (11, 5):
+            self.skipTest("-gen-lto unavailable in this toolkit version")
+
+        ltoir, resty = compile(f_module, int32(int32, int32), device=True,
```
stuartarchibald, line 208: `resty` is unused?

stuartarchibald (commented):
Thanks for the patch @gmarkall, it's great to see this feature implemented. This has been through an OOB pair review between us already in which the feature, expectations and implementation were discussed. The review provided below is just catching a few small things in the resultant change set, the contents otherwise is good. Thanks again for working on this!

numba/cuda/tests/cudapy/test_compiler.py
gmarkall: It is, but I don't like anonymizing unused returned tuple values by assigning them to `_` or similar, because then it's not obvious what's being discarded if you don't know the API (`compile()`, in this case) very well.

numba/cuda/tests/cudapy/test_compiler.py
gmarkall: I think that's a good idea.

gmarkall: For now I'd omitted it because there's no practical way to make use of such an output - we'd need to provide a way to compile the kernel with a C ABI (or more C-like ABI) and to be able to know or enforce the name of the generated kernel. I do want to do this at some future point, and I think that point would be an appropriate time to add the test.

numba/cuda/tests/cudapy/test_compiler.py
stuartarchibald: I see, could you assert is has an expected value perhaps?

numba/cuda/tests/cudapy/test_compiler.py
stuartarchibald: OK, thanks for the explanation, I considered that right now it probably didn't make sense but thought I'd just check!

docs/source/cuda/cuda_compilation.rst
gmarkall: I'm not sure I agree that it's not "code" in the traditional sense - I think "code" is a very broad term that refers to any source or executable sequence of bits.

However, I think you're asking here that I disambiguate between the input code and the output code, and I think that's helpful to do - I think changing this to "compiled code" should do that.

docs/source/cuda/cuda_compilation.rst
gmarkall: I think changing this to "The name of the device function in the compiled code" is clearer still (IMHO :slightly_smiling_face:) and also consistent with changing to "compiled code" above, so I'm going to go with that.

docs/source/cuda/cuda_compilation.rst
stuartarchibald: Ah yes, perhaps input vs output is more of my concern opposed to the specific meaning of code in this context. Changing to "compiled code" for the output seems to be an agreeable change, thanks!

docs/source/cuda/cuda_compilation.rst
stuartarchibald: Agree, thanks!

gmarkall:
gpuci run tests

stuartarchibald (approved):

PR #9292, opened 2023-11-20
CUDA: Switch cooperative groups to use overloads

gmarkall:
This is a non-functional change that only moves the CGs API implementation up to the high-level extension API. Note that CGs were never defined with a data model, so it's a little confusing to see what's going on when the grid group is passed to the `sync()` method, but it's an `i64` that is passed around opaquely.

Type hints are added on the public API to help IDEs / autocompleting REPLs.

The CGs tests have accidentally been skipped for a long time - this PR also includes a fix that reinstates them.

### DIFF
--- a/numba/cuda/cg.py
+++ b/numba/cuda/cg.py
@@ -54,7 +54,7 @@ def _grid_group_sync(typingctx, group):
     return sig, codegen
 
 
-@overload_method(GridGroupClass, 'sync')
+@overload_method(GridGroupClass, 'sync', target='cuda')
 def _ol_grid_group_sync(group):
     if isinstance(group, GridGroupClass):
         def impl(group):
--- a/numba/cuda/cg.py
+++ b/numba/cuda/cg.py
@@ -18,7 +18,7 @@ def this_grid() -> GridGroup:
     return GridGroup()
 
 
-@intrinsic
+@intrinsic(target='cuda')
 def _this_grid(typingctx):
     sig = signature(grid_group)
 
@@ -40,7 +40,7 @@ def _ol_this_grid():
     return impl
 
 
-@intrinsic
+@intrinsic(target='cuda')
 def _grid_group_sync(typingctx, group):
     sig = signature(types.int32, group)
 
--- a/numba/cuda/cg.py
+++ b/numba/cuda/cg.py
@@ -1,0 +1,63 @@
+from numba.core import types
+from numba.core.extending import overload, overload_method
+from numba.core.typing import signature
+from numba.cuda import nvvmutils
+from numba.cuda.extending import intrinsic
+from numba.cuda.types import grid_group, GridGroup as GridGroupClass
+
+
+class GridGroup:
+    """A cooperative group representing the entire grid"""
+
+    def sync() -> None:
+        """Synchronize this grid group"""
+
+
+def this_grid() -> GridGroup:
+    """Get the current grid group."""
+    return GridGroup()
+
+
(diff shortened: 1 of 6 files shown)


gmarkall:
gpuci run tests

numba/cuda/cg.py
```diff
@@ -0,0 +1,63 @@
...
+            (*args, flags))
+
+    return sig, codegen
+
+
+@overload_method(GridGroupClass, 'sync', target='cuda')
+def _ol_grid_group_sync(group):
+    if isinstance(group, GridGroupClass):
+        def impl(group):
+            return _grid_group_sync(group)
```
gmarkall, line 61: This check is redundant.
```suggestion
    def impl(group):
        return _grid_group_sync(group)
```

numba/cuda/cg.py
```diff
@@ -0,0 +1,63 @@
...
+
+        return impl
```
gmarkall, line 63: This check is redundant.
```suggestion
    def impl(group):
        return _grid_group_sync(group)

    return impl
```

gmarkall:
gpuci run tests

numba/cuda/cg.py
```diff
@@ -0,0 +1,62 @@
...
+    def sync() -> None:
+        """Synchronize this grid group"""
+
+
+def this_grid() -> GridGroup:
+    """Get the current grid group."""
+    return GridGroup()
+
+
+@intrinsic
```
stuartarchibald, line 21:
```suggestion
@intrinsic(target='cuda')
```
suggest restricting this to the CUDA target.

numba/cuda/cg.py
```diff
@@ -0,0 +1,62 @@
...
+
+@overload(this_grid, target='cuda')
+def _ol_this_grid():
+    def impl():
+        return _this_grid()
+
+    return impl
+
+
+@intrinsic
```
stuartarchibald, line 43:
```suggestion
@intrinsic(target='cuda')
```
suggest restricting this to the CUDA target.

numba/cuda/cg.py
```diff
@@ -0,0 +1,62 @@
...
+class GridGroup:
+    """A cooperative group representing the entire grid"""
+
+    def sync() -> None:
+        """Synchronize this grid group"""
+
+
+def this_grid() -> GridGroup:
+    """Get the current grid group."""
+    return GridGroup()
```
stuartarchibald, line 18:
```suggestion
```
does this need a Python implementation?

stuartarchibald (commented):
Thanks for the patch @gmarkall, great to see more of the target-extension API used in practice. Couple of minor comments to resolve else looks good.

numba/cuda/cg.py
gmarkall: I thought it might be instructive for anyone using an IDE that "peeks" at the implementation.

gmarkall:
gpuci run tests

gmarkall:
@stuartarchibald Thanks for the review - I've restricted the intrinsics to the CUDA target - just kicking off gpuci.

numba/cuda/cg.py
stuartarchibald: Thanks for clarifying, am happy to leave as-is.

stuartarchibald (dismissed):

gmarkall:
gpuci run tests

stuartarchibald (approved):

Issue #9283, opened 2023-11-15
Identityless reduction fails when axis is a tuple


guilhermeleobas:
Reproducer:

```python
from numba import njit, vectorize
import numpy as np

@vectorize
def min(a, b):
    return np.minimum(a, b)

@njit
def foo(a):
    return min.reduce(a, axis=(0, 1))

a = np.arange(2*3*4, dtype=np.int64).reshape(2, 3, 4)
x = foo(a)
print(x)
```

PR #9295, opened 2023-11-21
Improve reorderable ufunc support

guilhermeleobas:
Improve reorderable ufunc support (#9283) and add NumPy ufunc.reduce tests.

### DIFF
--- a/numba/np/ufunc/dufunc.py
+++ b/numba/np/ufunc/dufunc.py
@@ -497,7 +497,6 @@ class DUFunc(serialize.ReduceMixin, _internal._DUFunc, UfuncBase):
                 msg = 'The first argument "array" must be array-like'
                 raise errors.NumbaTypeError(msg)
 
-            axis_tuple = isinstance(axis, types.BaseTuple)
             axis_int_tuple = isinstance(axis, types.UniTuple) and \
                 isinstance(axis.dtype, types.Integer)
             axis_empty_tuple = isinstance(axis, types.Tuple) and len(axis) == 0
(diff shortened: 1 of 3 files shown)


numba/np/ufunc/dufunc.py
```diff
@@ -496,17 +497,19 @@ def ol_reduce(ufunc, array, axis=0, dtype=None, initial=None):
...
-            if self.ufunc.identity is None and not (
-                    (axis_int_tuple and axis_tuple_size == 1) or
-                    axis_empty_tuple or axis_int or axis_none):
-                msg = (f"reduction operation '{self.ufunc.__name__}' is not "
+            identity_none = self.ufunc.identity is None
+            ufunc_name = self.ufunc.__name__
+
+            # In NumPy, a ufunc is reorderable if its identity type is **not**
+            # PyUfunc_None.
+            if not self.reorderable and axis_tuple and len(axis) > 1:
```
kc611, line 511: `UniTuple` does not inherit from `BaseTuple` so this might miss catching axes tuples that are of type other than `int` but of a uniform tuple nature ?

kc611 (commented):
LGTM! Small comment regarding the axes type. After addressing that the PR can be approved.

numba/np/ufunc/dufunc.py
guilhermeleobas: Fixed!

kc611 (approved):
LGTM!

guilhermeleobas:
Thanks for the review @kc611

PR #9297, opened 2023-11-22
Add axis support to np.take

guilhermeleobas:
Non-scientific script to measure time:

```python
import time
from numba import njit
import numpy as np

@njit
def foo(a, indices, axis):
    return np.take(a, indices, axis)


def measure_time(fn, *args):
    start = time.time()
    fn(*args)
    end = time.time()
    print(f'{fn.__name__}: {end - start}s')



# N = 100
N = 10 ** 9
shape = (10,) * int(np.log10(N))
a = np.arange(N).reshape(shape)
indices = (1, 0, 2)
axis=0
foo(a, indices, axis)
measure_time(foo.py_func, a, indices, axis)
measure_time(foo, a, indices, axis)
```


numba/np/arrayobj.py
```diff
@@ -4665,52 +4666,126 @@ def impl(v, k=0):
...
+def generate_getitem_setitem_with_axis(ndim, kind):
+    assert kind in ('getitem', 'setitem')
+
+    if kind == 'getitem':
+        fn = '''
+            @register_jitable
+            def _getitem(a, idx, axis):
+                if axis == 0:
+                    return a[idx, ...]
+        '''
```
guilhermeleobas, line 4678: This could be rewritten to:

```python
def _get_take_getitem(ndim):
    @register_jitable
    def _take_getitem(a, idx, axis):
        if axis == 0:
            return a[idx, ...]
        elif ndim >= 2 and axis == 1:
            return a[:, idx, ...]
        elif ndim >= 3 and axis == 2:
            return a[:, :, idx, ...]
        elif ndim >= 4 and axis == 3:
            return a[:, :, :, idx, ...]
        elif ndim >= 5 and axis == 4:
            return a[:, :, :, :, idx, ...]
        elif ndim >= 6 and axis == 2:
            return a[:,:,:,:,:, idx, ...]
        elif ndim >= 7 and axis == 2:
            return a[:,:,:,:,:,:, idx, ...]
        elif ndim >= 8 and axis == 2:
            return a[:,:,:,:,:,:,:, idx, ...]
        elif ndim >= 9 and axis == 2:
            return a[:, :, :, :, :, :, :, idx, ...]
        elif ndim >= 11 and axis == 10:
            ...
        elif ndim >= 12 and axis == 10:
        elif ndim >= 13 and axis == 10:
        elif ndim >= 14 and axis == 10:
        elif ndim >= 15 and axis == 10:
        elif ndim >= 16 and axis == 10:
        elif ndim >= 17 and axis == 10:
        elif ndim >= 18 and axis == 10:
        elif ndim >= 19 and axis == 10:
        elif ndim >= 20 and axis == 20:
        elif ndim >= 21 and axis == 20:
        elif ndim >= 22 and axis == 20:
        elif ndim >= 23 and axis == 20:
        elif ndim >= 24 and axis == 20:
        elif ndim >= 25 and axis == 20:
        elif ndim >= 26 and axis == 20:
        elif ndim >= 27 and axis == 20:
        elif ndim >= 28 and axis == 20:
        elif ndim >= 29 and axis == 20:
        elif ndim >= 30 and axis == 30:
        elif ndim >= 31 and axis == 31:
        elif ndim >= 32 and axis == 32:
    return _take_getitem
```

numba/np/arrayobj.py
```diff
@@ -4665,52 +4666,126 @@ def impl(v, k=0):
...
+
+        for i in range(1, ndim):
+            lst = (':',) * i
+            fn += f'''
+                elif axis == {i}:
+                    a[{", ".join(lst)}, idx, ...] = vals
+            '''
+
+    fn = textwrap.dedent(fn)
+    exec(fn, globals())
```
guilhermeleobas, line 4701: From an OOB meeting with Kaustubh, he suggested moving the `@register_jitable` to here.

numba/np/arrayobj.py
kc611: The string itself should not contain a `register_jitable`, the output function of this string should be jitted instead.

kc611 (changes requested):

numba/np/arrayobj.py
```diff
@@ -4665,52 +4666,125 @@ def impl(v, k=0):
...
+        if isinstance(a, types.Array) and isinstance(indices, types.Integer):
+            t = (0,) * (a.ndim - 1)
+
+            @register_jitable
+            def _squeeze(r):
+                # np.squeeze is too hard to implement in Numba
+                tup = tuple(t)
+                j = 0
+                for s in r.shape:
+                    if s != 1:
```
kc611, line 4752: The squeeze implementation needs to exclude the axis based on the axis provided not the value `1`:

For instance:
```python
import time
from numba import njit
import numpy as np

@njit
def foo(a, indices, axis):
    return np.take(a, indices, axis)


def measure_time(fn, *args):
    start = time.time()
    res = fn(*args)
    end = time.time()
    print(f'{fn.__name__}: {end - start}s')
    print(res.shape)
    print("ndim:", res.ndim)



shape = (10, 1, 11, 1, 12, 1, 13)
a = np.ones(shape)
indices = 0
axis=1
measure_time(foo.py_func, a, indices, axis)
# foo: 0.00011563301086425781s
# (10, 11, 1, 12, 1, 13)
# ndim: 6
measure_time(foo, a, indices, axis) # fails

shape = (10, 1, 11, 1, 12, 1, 13)
a = np.ones(shape)
indices = 0
axis = 2
measure_time(foo.py_func, a, indices, axis)
# foo: 6.4373016357421875e-06s
# (10, 1, 1, 12, 1, 13)
# ndim: 6
measure_time(foo, a, indices, axis) # fails

```

numba/np/arrayobj.py
```diff
@@ -4665,52 +4666,125 @@ def impl(v, k=0):
...
+                if axis < 0 or axis >= a.ndim:
+                    msg = (f"axis {axis} is out of bounds for array "
+                           f"of dimension {a.ndim}")
+                    raise ValueError(msg)
+
+                shape = tuple_setitem(a.shape, axis, len(indices))
+                out = np.empty(shape, dtype=a.dtype)
+                for i in range(len(indices)):
+                    y = _getitem(a, indices[i], axis)
+                    _setitem(out, i, axis, y)
```
kc611, line 4785: This is potentially very wasteful. If we're anyways generating the `getitem` and `setitem` specifically for this function we may want to combine them into a single function that copies the elements directly in the `out` resultant. Is that something you've tried out ?

kc611 (changes requested):

numba/np/arrayobj.py
guilhermeleobas: I've split getitem from setitem because the latter is used in https://github.com/numba/numba/pull/9313

numba/np/arrayobj.py
guilhermeleobas: Good catch. I'll fix this

guilhermeleobas:
CI failure is not related to this PR.

gmarkall:
I've given the CI a prod.

PR #9301, opened 2023-11-24
Release Notes 0.58.1 for main

esc:
As title


docs/source/user/installing.rst
```diff
@@ -5,7 +5,7 @@ Installation
 Compatibility
 -------------
 
-For software compatibility, please see the section on :ref:`version support
+For software compatability, please see the section on :ref:`version support
```
stuartarchibald, line 8: This being changed to an incorrect spelling?

stuartarchibald (approved):
Thanks for the patch, I've compared this to the `release0.58` branch and on visual inspection it appears that all the relevant changes added in association with the 0.58.1 release have been duplicated into this PR.

PR #9302, opened 2023-11-24
add missing backtick to example git tag command

esc:
As title


stuartarchibald (approved):
Thanks for the patch.

PR #9303, opened 2023-11-24
Add category to warning

kkokkoros:
### Background

For my project, I'm using [rembg](https://github.com/danielgatis/rembg) in order to remove background for images in a AWS lambda function. This library is using [PyMatting](https://github.com/pymatting/pymatting) to implement Alpha Matting which depends on numba. But, since [_multiprocessing.SemLock is not implemented when running on AWS Lambda](https://stackoverflow.com/questions/34005930/multiprocessing-semlock-is-not-implemented-when-running-on-aws-lambda), I get a warning every time the lambda function runs. This pollutes the function logs

### Workaround

Since

* I cannot control what Amazon offers
* I actually don't use Alpha Matting, so this warning is irrelevant for my use case

I would like to silence this specific warning using [warnings.simplefilter](https://docs.python.org/3/library/warnings.html#warnings.simplefilter)

### Solution
In order to support silencing this specific warning and not everything, I have added the category parameter to the `warnings.warn` call. Hopefully, this is the proper class for this warning, but if needed, another warning class can be added

### DIFF
--- a/numba/core/errors.py
+++ b/numba/core/errors.py
@@ -103,6 +103,12 @@ class NumbaDebugInfoWarning(NumbaWarning):
     Warning category for an issue with the emission of debug information.
     """
 
+
+class NumbaSystemWarning(NumbaWarning):
+    """
+    Warning category for an issue with the system configuration.
+    """
+
 # These are needed in the color formatting of errors setup
 
 
--- a/numba/np/ufunc/parallel.py
+++ b/numba/np/ufunc/parallel.py
@@ -321,7 +321,7 @@ def _set_init_process_lock():
             "this initialization sequence/module import is deferred to the "
             "user! ***\n"
         )
-        warnings.warn(msg % str(e), errors.NumbaParallelSafetyWarning)
+        warnings.warn(msg % str(e), errors.NumbaSystemWarning)
 
         _backend_init_process_lock = _nop()
 
--- a/numba/np/ufunc/parallel.py
+++ b/numba/np/ufunc/parallel.py
@@ -321,7 +321,7 @@ def _set_init_process_lock():
             "this initialization sequence/module import is deferred to the "
             "user! ***\n"
         )
-        warnings.warn(msg % str(e))
+        warnings.warn(msg % str(e), errors.NumbaParallelSafetyWarning)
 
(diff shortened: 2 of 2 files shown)


stuartarchibald:
Thanks for the patch, I'm not sure about the use of `NumbaParallelSafetyWarning`, its intended use is for warning about problems that might occur in a parallel region, e.g. race conditions. https://github.com/numba/numba/blob/a4664180ddc91e4c8a37dd06324f4e229a76df91/numba/core/errors.py#L62-L66

This does make me wonder if there is a more appropriate warning than just the standard `NumbaWarning` and I'm not sure there is. Use of this "base class" of warning would probably help a warnings filter, but the regex string match on the message would also need to be used (such a regex match might also work against the existing code base too?). Would changing the warning type to that still help with your use case @kkokkoros ?

kkokkoros:
Hello @stuartarchibald and thanks for the quick review!

>  Would changing the warning type to that still help with your use case @kkokkoros

Yes of course it would, I just couldn't think of any good name since I don't have much experience with numba internals. If you can propose a better name, I could add a new `NumbaWarning` subclass

stuartarchibald:
> Hello @stuartarchibald and thanks for the quick review!
> 
> > Would changing the warning type to that still help with your use case @kkokkoros
> 
> Yes of course it would, I just couldn't think of any good name since I don't have much experience with numba internals. If you can propose a better name, I could add a new `NumbaWarning` subclass

How about adding a class `NumbaSystemWarning` to `numba.core.errors` and then using that in the warning category? There's numerous warning classes defined in that module already for reference. Thanks!

stuartarchibald (approved):
Thanks for the patch and fixes. This looks good on inspection. It's quite hard to write tests for this particular piece of code (without a lot of effort that may not be proportionate to the gain), therefore I tested that the warning triggers by adding a `raise OSError` into the `try` here https://github.com/numba/numba/blob/58dd3cae5ca2dbaf6d583460a4829914fa41e208/numba/np/ufunc/parallel.py#L300-L323 and made sure that the warning was raised and it is now indeed a `NumbaSystemWarning`. Thanks for making this improvement @kkokkoros, it'll be available in the 0.59 release.

PR #9306, opened 2023-11-27
Charseq to charseq cast

### LINKED ISSUE #9305: Support Safe/Unsafe casts between `CharSeq` types.
## Overview

A char-sequence type should support conversion to another char-sequence type. It should be a safe conversion if the target's sequence count is larger or equal to the source's, otherwise it should be unsafe.

## Proposed implementation

- Add conversion rules to the [`CharSeq` type](https://github.com/numba/numba/blob/a4664180ddc91e4c8a37dd06324f4e229a76df91/numba/core/types/npytypes.py#L30)
- Implement `lower_cast` for charseq->charseq [here](https://github.com/numba/numba/blob/a4664180ddc91e4c8a37dd06324f4e229a76df91/numba/cpython/charseq.py#L220)

## Code Snippet

```

import numba as nb
import numpy as np

src = np.array(["this", "that"], dtype="S4")
dst = np.empty(2, dtype="S8")

@nb.njit
def assign_string_arr(src, dst):
    for i in range(len(src)):
        dst[i] = src[i]
    
assign_string_arr(src, dst)
dst
```

results in a typing error:
```
 
 >>> setitem(array([char x 8], 1d, C), int64, [char x 4])
 
    - Of which 16 did not match due to:
      With argument(s): '(array([char x 8], 1d, C), int64, [char x 4])':
     No match.
```

drewmoore27:
closes #9305 

Implements conversions/casts from CharSeq to CharSeq, to support assignment from numpy byte-string arrays. (e.g., np.array(["x", "y"], dtype="S")).

The  numpy.bytes_ type seems to be implemented with a signed int32 - `np.dtype(f"S{2**31}")` wraps around (this actually returns an un-instantiable `dtype('S-2147483648')`). Thus in the lowering code, we use an ir.IntType(32) for the count size.

The lowering code is mostly inspired by the `bytes_to_charseq` function above it on L169.

### DIFF
--- a/numba/core/types/npytypes.py
+++ b/numba/core/types/npytypes.py
@@ -31,6 +31,11 @@ class CharSeq(Type):
         if isinstance(other, Bytes):
             return Conversion.safe
 
+    def can_convert_to(self, typingctx, other):
+        if isinstance(other, CharSeq):
+            if other.count < self.count:
+                return Conversion.unsafe
+            return Conversion.safe
 
 class UnicodeCharSeq(Type):
     """
--- a/docs/upcoming_changes/9305.charseq_cast.rst
+++ b/docs/upcoming_changes/9305.charseq_cast.rst
@@ -1,0 +1,5 @@
+Implement casting from charseq to charseq
+=========================================
+
+Allows automatic casting from dtype "S<N>" to "S<M>" 
+if M is larger or equal to N.
--- a/numba/cpython/charseq.py
+++ b/numba/cpython/charseq.py
@@ -218,6 +218,35 @@ def charseq_to_bytes(context, builder, fromty, toty, val):
     cgutils.memcpy(builder, bstr.data, ptr, bstr.nitems)
     return bstr
 
+@lower_cast(types.CharSeq, types.CharSeq)
+def charseq_to_charseq(context, builder: ir.IRBuilder, fromty: types.CharSeq, toty: types.CharSeq, val: ir.values.values.Value):
+    char_t = ir.IntType(8)
+    count_t = ir.IntType(32)
+
+    src_ptr = cgutils.alloca_once_value(builder, value=val)
+    src = builder.bitcast(src_ptr, char_t.as_pointer())
+
+    lty = context.get_value_type(toty)
+    dst_ptr = cgutils.alloca_once(builder, lty)
+    dst = builder.bitcast(dst_ptr, char_t.as_pointer())
+
+    src_length = ir.Constant(count_t, fromty.count)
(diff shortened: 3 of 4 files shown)


numba/core/types/npytypes.py
```diff
@@ -31,6 +31,11 @@ def can_convert_from(self, typingctx, other):
         if isinstance(other, Bytes):
             return Conversion.safe
 
+    def can_convert_to(self, typingctx, other):
+        if isinstance(other, CharSeq):
+            if other.count < self.count:
+                return Conversion.unsafe
+            return Conversion.safe
```
sklam, line 38: Note for reviewer: 

- review why `UnicodeCharSeq` has a different cast logic: https://github.com/numba/numba/pull/9306/files#diff-98dd0dbec153c38b3afee6df1262d9e3fd3436dc477cc1d6b61771fb9b40bcfcL56-L59

guilhermeleobas: `UnicodeCharSeq` is lacking an if-case when `other` is also an `UnicodeCharSeq`. So, the following fails:

```python
from numba.core import types
from numba import njit
import numpy as np

src_typ = types.Array(types.UnicodeCharSeq(4), 1, "A")
dst_typ = types.Array(types.UnicodeCharSeq(5), 1, "A")
sig = types.int32(src_typ, dst_typ)

@njit(sig)
def assign_unicodecharseq_array(src, dst):
    for i in range(len(src)):
        dst[i] = src[i]
    return 0

x = np.array(["1234", "5678"], dtype="U4")
y = np.empty(2, dtype="U5")
assign_unicodecharseq_array(x, y)

# numba.core.errors.NumbaNotImplementedError: Failed in nopython mode pipeline (step: native lowering)
# Cannot cast [unichr x 4] to [unichr x 5]: %".191" = load [4 x i32], [4 x i32]* %".190"
# During: lowering "dst[i] = $22binary_subscr.4" at /Users/guilhermeleobas/git/numba/b.py (12)
```

guilhermeleobas (changes requested):
PR in general looks good. Only have a few comments that should be easy to address. Thanks @drewmoore27

guilhermeleobas:
@drewmoore27, I can take over this PR if needed. Just let me know.

drewmoore27:
Yes thanks! I have not been checking in hereSent from my iPhoneOn Mar 7, 2024, at 2:27 PM, Guilherme Leobas ***@***.***> wrote:﻿
@drewmoore27, I can take over this PR if needed. Just let me know.

PR #9309, opened 2023-11-29
Continue #9044, prevent compilation on the CPU target when listing tests.

stuartarchibald:
This PR continues on from #9044, with thanks to @apmasell for the original patch. It prevents compilation on the CPU target when listing tests. It adds to the original PR to prevent compilation through the use of the `@stencil` decorator and also adds a best effort unit test to ensure that test listing will not invoke CPU target compilation in future.

### DIFF
--- a/docs/upcoming_changes/9309.change.rst
+++ b/docs/upcoming_changes/9309.change.rst
@@ -1,0 +1,8 @@
+Make test listing not invoke CPU compilation.
+---------------------------------------------
+
+Numba's test listing command ``python -m numba.runtests -l`` has historically
+triggered CPU target compilation due to the way in which certain test functions
+were declared within the test suite. It has now been made such that the CPU
+target compiler is not invoked on test listing and a test is added to ensure
+that it remains the case.
--- a/numba/stencils/stencil.py
+++ b/numba/stencils/stencil.py
@@ -80,8 +80,6 @@ class StencilFunc(object):
         # stencils only supported for CPU context currently
         self._typingctx = registry.cpu_target.typing_context
         self._targetctx = registry.cpu_target.target_context
-        self._typingctx.refresh()
-        self._targetctx.refresh()
         self._install_type(self._typingctx)
         self.neighborhood = self.options.get("neighborhood")
         self._type_cache = {}
@@ -766,6 +764,7 @@ class StencilFunc(object):
         return new_func
 
     def __call__(self, *args, **kwargs):
+        self._typingctx.refresh()
         if (self.neighborhood is not None and
             len(self.neighborhood) != args[0].ndim):
             raise ValueError("{} dimensional neighborhood specified for {} "
--- a/numba/tests/chained_assign_usecases.py
+++ b/numba/tests/chained_assign_usecases.py
@@ -1,0 +1,69 @@
+from numba import jit
+import numpy as np
+
+
+@jit
+def inc(a):
+    for i in range(len(a)):
+        a[i] += 1
+    return a
+
+
+@jit
+def inc1(a):
+    a[0] += 1
+    return a[0]
+
+
+@jit
+def inc2(a):
+    a[0] += 1
+    return a[0], a[0] + 1
+
+
+def chain1(a):
+    x = y = z = inc(a)
+    return x + y + z
+
+
+def chain2(v):
+    a = np.zeros(2)
+    a[0] = x = a[1] = v
+    return a[0] + a[1] + (x / 2)
+
+
+def unpack1(x, y):
+    a, b = x, y
+    return a + b / 2
+
+
+def unpack2(x, y):
+    a, b = c, d = inc1(x), inc1(y)
+    return a + c / 2, b + d / 2
+
+
+def chain3(x, y):
+    a = (b, c) = (inc1(x), inc1(y))
+    (d, e) = f = (inc1(x), inc1(y))
+    return (a[0] + b / 2 + d + f[0]), (a[1] + c + e / 2 + f[1])
+
+
+def unpack3(x):
+    a, b = inc2(x)
+    return a + b / 2
+
+
+def unpack4(x):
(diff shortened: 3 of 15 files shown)


docs/upcoming_changes/9309.change.rst
```diff
@@ -0,0 +1,8 @@
+Make test listing not invoke CPU compilation.
+---------------------------------------------
```
stuartarchibald, line 2: @kc611 @esc Should these be `"` or does it make no difference?

esc: I think they should be either `"` or `=` or something else that is unused. We maybe need to revisit this. RST uses the convention, that you can use any character for underlining, but it must be consistent.

Looking at the last two releases:

https://github.com/numba/numba/tree/main/docs/source/release

https://raw.githubusercontent.com/numba/numba/main/docs/source/release/0.58.1-notes.rst

and

https://raw.githubusercontent.com/numba/numba/main/docs/source/release/0.58.0-notes.rst

0.58.0 used the following structure:

First level: `---` (hyphen)
Second level: `~~~` (tilde)
Third level: `""""` (quotes)

But 0.58.1 uses:

First level: `---` (hyphen)
Second level: `~~~` (tilde)
Third level: `====` (equals)

If we look at the upcoming changes in:

https://github.com/numba/numba/tree/main/docs/upcoming_changes

It's a mix of `=` and `"` unfortunately.

The ideal solution IMHO would be:

```
=========
Version xxx
=========

Highlights
=======

Feature X was implemented
----------------------------

Description of feature X

Bug Fixes
=======

Bug Y was fixed
----------------

Description of bug Y fixed

Pull-Requests
==========

* 1
* 2
* 3

Authors
======

* a
* b
* c

```

So, first level headings use double equals `=` (top and bottom, you can also uses overline and underline, but must be same length and same character). Second level heading uses underline `=` equals and third level heading to use `-` hyphens. Thus, the snippet titles would use `-` hyphen which, perhaps, is most natural for contributors. Only the docs about snippets, the PR-list generation script and the (soon to come) automatic checking action would need to be changed. Existing release notes can stay as is as each release note is in its own file.

stuartarchibald: Thanks for this explanation @esc, perhaps this should be captured in #9311 with view of enforcing it as part of the changes suggested in that task?

esc:
So I tried testing this in two ways. Patching the source code and writing a small script to list the tests. The patch for the source code is:

```
-from numba.testing._runtests import _main


 if __name__ == '__main__':
+    print("RUNNING")
     import sys
     # For parallel testing under Windows
-    from multiprocessing import freeze_support
-    freeze_support()
-    sys.exit(0 if _main(sys.argv) else 1)
+    #from multiprocessing import freeze_support
+    #freeze_support()
+    from unittest import mock
+    from llvmlite import binding as llvm
+    error = RuntimeError("Detected compilation during test listing")
+    with mock.patch.object(llvm.ExecutionEngine, 'finalize_object', side_effect=error):
+        from numba.testing._runtests import _main
+        sys.exit(0 if _main(sys.argv) else 1)
```

And the script is:

```
from unittest import mock

error = RuntimeError("Detected compilation during test listing")
with mock.patch.object(llvm.ExecutionEngine, 'finalize_object',
                       side_effect=error):
    import numba
    numba.test('-l')
```

Using these two approachs on `main` at 51fce0235e8b7ba1dc9ae41657c546a608b02847 however yields to different results. The patch doesn't trigger the error, but the script does. I'm still trying to work out why.

Using this `breakpoint()` patch:

```
diff --git i/numba/testing/__init__.py w/numba/testing/__init__.py
index 4eb89a92fc..851d252e63 100644
--- i/numba/testing/__init__.py
+++ w/numba/testing/__init__.py
@@ -51,6 +51,7 @@ def run_tests(argv=None, defaultTest=None, topleveldir=None,
         runner = xmlrunner.XMLTestRunner(output=xmloutput)
     else:
         runner = None
+    breakpoint()
     prog = NumbaTestProgram(argv=argv,
                             module=None,
                             defaultTest=defaultTest,
@@ -58,4 +59,5 @@ def run_tests(argv=None, defaultTest=None, topleveldir=None,
                             testRunner=runner, exit=False,
                             verbosity=verbosity,
                             nomultiproc=nomultiproc)
+    breakpoint()
     return prog.result
```

I can confirm that both approaches end up at the first breakpoint, before either listing the tests or crashing because of an exception. I can also confirm that the `finalize_object` function has indeed been replaced with a `MagicMock` object.

```
-> prog = NumbaTestProgram(argv=argv,
(Pdb++) from llvmlite import binding as llvm
(Pdb++) llvm.ExecutionEngine.finalize_object
<MagicMock name='finalize_object' id='4780673808'>
```

So the mock that is causing `llvmlite` to raise does appear to be active at this stage in both approaches.

I can confirm however, that this patch works when tested as advertised. Though I would like to confirm it disables compilation when used from the command line in the "standard" way.

docs/upcoming_changes/9309.change.rst
esc: https://github.com/numba/numba/issues/9311#issuecomment-1834341155

sklam (approved):
I tested it with the diff:

```diff
diff --git a/numba/core/codegen.py b/numba/core/codegen.py
index 28ab618a0..3775ea6da 100644
--- a/numba/core/codegen.py
+++ b/numba/core/codegen.py
@@ -1057,6 +1057,8 @@ class RuntimeLinker(object):
 def _proxy(old):
     @functools.wraps(old)
     def wrapper(self, *args, **kwargs):
+        if old not in {ll.ExecutionEngine.set_object_cache}:
+            raise NotImplementedError(old)
         return old(self._ee, *args, **kwargs)
     return wrapper

```
before and after this PR and can confirm that no compilation is triggered in LLVM when running `python -m numba.runtests -l`.

PR #9315, opened 2023-11-29
Fix support for LOAD_FAST_AND_CLEAR

### LINKED ISSUE #9300: [Python 3.12] `LOAD_FAST_AND_CLEAR`
Python 3.12 introduced the `LOAD_FAST_AND_CLEAR` opcode. This turned out to be more involved than anticipated and it has been decided it will need attention outside of #9246 . This ticket tracks that effort.

sklam:
Fix #9300

### DIFF
--- a/numba/cpython/builtins.py
+++ b/numba/cpython/builtins.py
@@ -1002,12 +1002,3 @@ def ol_str_generic(object=''):
         else:
             return repr(object)
     return impl
-
-
-@lower_cast(types.UndefVar, types.Any)
-def cast_undef_var_to_any(context, builder, fromty, toty, val):
-    """UndefVar can cast to anything.
-
-    This should only used be used in phi nodes.
-    """
-    return context.get_constant_null(toty)
--- a/numba/core/base.py
+++ b/numba/core/base.py
@@ -74,18 +74,6 @@ class OverloadSelector(object):
             same = list(takewhile(lambda x: genericity[x] == firstscore,
                                   ordered))
             if len(same) > 1:
-                # Resolve ambiguous cases in the pattern of:
-                #   (cast any -> T)
-                #   (cast U -> any)
-                # and chooses the single case that is (cast any -> T)
-                to_anys = [(ta, tb) for (ta, tb) in same if types.Any == tb]
-                from_anys = [(ta, tb) for (ta, tb) in same if types.Any == ta]
-                n_from_anys = len(from_anys)
-                n_to_anys = len(to_anys)
-                if n_from_anys + n_to_anys == len(same):
-                    if n_from_anys == 1 and n_to_anys >= 1:
-                        return from_anys[0]
-                # Report ambiguous
                 msg = ["{n} ambiguous signatures".format(n=len(same))]
                 for sig in same:
                     msg += ["{0} => {1}".format(sig, candidates[sig])]
@@ -704,7 +692,10 @@ class BaseContext(object):
         This implements implicit conversions as can happen due to the
         granularity of the Numba type system, or lax Python semantics.
         """
-        if fromty == toty or toty == types.Any:
+        if fromty is types._undef_var:
+            # Special case for undefined variable
+            return self.get_constant_null(toty)
+        elif fromty == toty or toty == types.Any:
             return val
         try:
             impl = self._casts.find((fromty, toty))


numba/core/types/__init__.py
```diff
@@ -28,6 +28,9 @@
 npy_rng = NumPyRandomGeneratorType('rng')
 npy_bitgen = NumPyRandomBitGeneratorType('bitgen')
 
+# _undef_var is used to represent undefined variables in the type system.
+_undef_var = UndefVar('_undef_var') #
```
stuartarchibald, line 32:
```suggestion
_undef_var = UndefVar('_undef_var')
```

numba/core/typeinfer.py
```diff
@@ -53,6 +53,11 @@ def __init__(self, context, var):
 
     def add_type(self, tp, loc):
         assert isinstance(tp, types.Type), type(tp)
+        # Special case for _undef_var.
+        # If the typevar is a _undef_var, use the incoming type directly.
```
stuartarchibald, line 57:
```suggestion
        # If the typevar is the _undef_var, use the incoming type directly.
```
Presumably there's only one `types._undef_var` as it's on the `types.__init__`?

numba/cpython/builtins.py
```diff
@@ -1003,3 +1003,11 @@ def impl(object=""):
             return repr(object)
     return impl
 
+
+@lower_cast(types.UndefVar, types.Any)
+def cast_undef_var_to_any(context, builder, fromty, toty, val):
+    """UndefVar cast to anything.
```
stuartarchibald, line 1009:
```suggestion
    """UndefVar can cast to anything.
```

numba/tests/test_typeinfer.py
```diff
@@ -707,6 +708,25 @@ def confuse_typer(x):
...
         self.assertTrue(all(vt == types.float64 for vt in return_vars.values()))
 
+    @unittest.skipUnless("LOAD_FAST_AND_CLEAR" in dis.opmap,
+                         "Requires LOAD_FAST_AND_CLEAR opcode")
+    def test_load_fast_and_clear(self):
+        @njit
+        def foo(a):
+            [x for x in (0,)]
+            if a:
+                # can't be a constant due to constant propagation issues.
```
stuartarchibald, line 718:
```suggestion
                # the test code cannot use a constant here due to constant
                # propagation issues.
```

numba/tests/test_typeinfer.py
```diff
@@ -707,6 +708,25 @@ def confuse_typer(x):
...
+                x = 3 + a
+            x += 10
+            return x
+
+        self.assertEqual(foo(True), foo.py_func(True))
+        # Interpreted version should raise an exception
+        with self.assertRaises(UnboundLocalError):
+            foo.py_func(False)
+        # Compiled version returns 10 as x is zero initialized.
+        self.assertEqual(foo(False), 10)
```
stuartarchibald, line 728: Note: This is indeed the case, it's unusual but valid.

numba/core/types/misc.py
```diff
@@ -33,6 +33,18 @@ def is_precise(self):
...
+    A type that is created by Expr.undef to represent an undefined variable.
+    This type can be promoted to any other type.
+    This is introduced to handle Python 3.12 LOAD_FAST_AND_CLEAR.
+    """
+
+    def can_convert_to(self, typingctx, other):
+        return Conversion.promote
+
+
+
```
stuartarchibald, line 47:
```suggestion
```
2 empty lines between class definitions.

numba/core/lowering.py
```diff
@@ -1413,6 +1413,11 @@ def lower_expr(self, resty, expr):
         elif expr.op == 'null':
             return self.context.get_constant_null(resty)
 
+        elif expr.op == 'undef':
+            # Numba do not raise UnboundLocalError for undefined variables.
```
stuartarchibald, line 1417:
```suggestion
            # Numba does not raise an UnboundLocalError for undefined variables.
```

stuartarchibald (commented):
Thanks for the patch, this seems to fix most of the issues associated with the `LOAD_FAST_AND_CLEAR` op code. In testing this, the following sample is a variant on a test from the CPython test suite, specifically: https://github.com/python/cpython/blob/d4a6229afe10d7e5a9a59bf9472f36d7698988db/Lib/test/test_listcomps.py#L368
```python
@njit
def foo():
    if False:
        x = 0
    [x for x in (1,)]
    return x

foo()
```
which results in a `TypeError: cannot convert native _undef_var to Python object`.

sklam:
RE: https://github.com/numba/numba/pull/9315#pullrequestreview-1755908955

addressed in https://github.com/numba/numba/pull/9315/commits/3f5ebbb5ba3857f79acf87f17ab6648ca592d13c

the type inferer will scan for use of `_undef_var` in call arguments and in return types.

sklam:
CI is failing due to:

- ambiguous cast between undef-to-any and optional cast
- objectmode not handling Expr.undef

esc (approved):
LGTM

PR #9319, opened 2023-11-30
Post py38 remove cleanup

esc:
Following #9310 I found two more bytecodes that were last seen with 3.7: `SETUP_LOOP` and `BREAK_LOOP`. These also appear in the documentation about the Numba architecture `docs/source/developer/architecture.rst`. However, updating the examples to show output generated under 3.11 appears to be quite rabitthole and will require some more work. The pull-request is opened as non-draft anyway, such that public CI can run and check if the code changes are good.


esc:
will punt this once, maybe I will make it.

PR #9325, opened 2023-12-01
Fix GPUCI

gmarkall:
This was broken by:

https://github.com/numba/numba/commit/57ad76320629ca0d40755c07f458091f1dc6c35d#diff-2b93d4ba7b2d38864ac7db14034ee95008ff895ad85ea31ff96d10f92b1ecc2bR2

It changed the base Python version in the container from 3.8 to 3.9. We don't use the base Python version, so it doesn't matter going forward if the base version is 3.8 - however, there is no 3.9 base version of the RAPIDS 21.12 container, so the base version can't be 3.9 unless we move to another image.


gmarkall:
gpuci run tests

esc (approved):
LGTM

esc:
@gmarkall thank you for the patch

PR #9336, opened 2023-12-06
Add TargetLibraryInfo pass to CPU LLVM pipeline.

stuartarchibald:
The TLI pass makes sure that the optimisations that take place during call simplification are appropriate for the target, without this the target is assumed to be Linux and code will be optimised to produce e.g. math symbols that do not exit on Windows.

This addresses #8898 in a more performant manner, it is almost a total revert of #8907 (original fix for #8898), only the test cases are kept.


sklam:
LGTM. Pending `numba_smoketest_cpu_yaml_184`

PR #8583, opened 2022-11-07
Skip tests that contribute to M1 RuntimeDyLd Assertion error 

sklam:
temporary workaround for #8567

PR #9337, opened 2023-12-06
Revert #8583 which skip tests due to M1 RuntimeDyLd Assertion error

sklam:
Revert https://github.com/numba/numba/pull/8583
Because the problem is fixed by https://github.com/numba/llvmlite/pull/1009


sklam:
All M1 test passed on the farm

gmarkall (approved):
Looks good to me. I note it mentioned in the private FPOC chat that the buildfarm has already passed.

PR #4677, opened 2019-10-09
Add support for np.setxor1d

synapticarbors:
This PR is adds `np.setxor1d`, which is currently unimplemented according to #4074 

One difference between the implementation here and numpy is that numpy handles non-1d arrays if the argument `assume_unique` is `False`, since the input arrays are run through `np.unique`. If not, then the numpy function fails with errors related to the dimensionality of the arrays at one of the two calls to `np.concatenate`. If the input arrays are of incompatible shape, then they fail at the first call, else they fail at the second when they are concatentated with `[True]`. See:

https://github.com/numpy/numpy/blob/v1.17.0/numpy/lib/arraysetops.py#L436-L476

I do not believe you can test the value of the boolean arg before returning the implementation in the `overload` outer function, you end up with a type unification problem that I can't seem to get around. So I opted to just insist that all inputs are 1d. Let me know if there is a workaround for this that I'm not seeing.

PR #9338, opened 2023-12-07
Add np.in1d, np.isin, np.setxor1d, np.setdiff1d, extend np.intersect1d.

jaredjeya:
* Adds support for the functions `np.in1d`, `np.isin`, `np.setxor1d`, and `np.setdiff1d`, all including optional argument `assume_unique` as well as `invert` for `np.in1d` and `np.isin`.

* Extends `np.intersect1d` to support the first optional argument (assume_unique).

Resolves #4677. Improves on #4074.

I have written unit tests for all the new functions and functionality, and applied flake8 to my code.

The only question I have is how to handle the optional argument `kind` in `np.in1d`: I've implemented the `kind="sort"` algorithm, but this is not the default behaviour of `numpy` (which picks an algorithm based on performance heuristics), so I've made it explicit that only `kind="sort"` is supported.


jaredjeya:
I'm not sure why those tests are failing all of a sudden - it seems to be at the `flake8` step? But the errors `flake8` is highlighting are unrelated to code I've modified and it didn't fail in the earlier test.

stuartarchibald:
@jaredjeya feel free to ignore the `flake8` failures until #9342 is merged later today (a new version of `flake8` was released and it picked up a couple of new issues). Thanks.

jaredjeya:
> @jaredjeya feel free to ignore the `flake8` failures until #9342 is merged later today (a new version of `flake8` was released and it picked up a couple of new issues). Thanks.

Now it’s been merged, how do I get the failed tests to run again? Is there anything I need to do to keep the new code up-to-date with the main branch? (Sorry, this is my first ever PR to a project in active development!)

gmarkall:
@jaredjeya Many thanks for the PR! I'm tentatively moving it to the 0.60 milestone - at the moment we're working on the 0.59 release, and the 0.60 release will be focused around supporting NumPy 2.0, so it may be some time before it's possible to review this PR.

numba/np/arraymath.py
```diff
@@ -4836,3 +4842,162 @@ def impl(filt, trim='fb'):
...
+
+@overload(np.setdiff1d)
+def jit_np_setdiff1d(ar1, ar2, assume_unique=False):
+    if not (type_can_asarray(ar1) or type_can_asarray(ar2)):
+        raise TypingError('setdiff1d: first two args must be array-like')
+    if not (isinstance(assume_unique, (types.Boolean, bool))):
+        raise TypingError('setdiff1d: Argument assume_unique must be boolean')
+
+    # https://github.com/numpy/numpy/blob/
+    # 03b62604eead0f7d279a5a4c094743eb29647368/numpy/lib/arraysetops.py#L940
```
guilhermeleobas, line 4889: Same here

numba/np/arraymath.py
```diff
@@ -4836,3 +4842,162 @@ def impl(filt, trim='fb'):
...
+    if not (isinstance(invert, (types.Boolean, bool))):
+        raise TypingError('in1d: Argument "invert" must be boolean')
+    if not (isinstance(kind, (types.UnicodeType, str))):
+        raise TypingError('in1d: Argument "kind" must be string')
+
+    # https://github.com/numpy/numpy/blob/
+    # 03b62604eead0f7d279a5a4c094743eb29647368/numpy/lib/arraysetops.py#L525
+    def np_in1d_impl(ar1, ar2, assume_unique=False, invert=False, kind="sort"):
+        if kind != "sort":
+            raise ValueError('Only kind="sort" is supported')
```
guilhermeleobas, line 4919: Since `kind="sort"` is the only accepted value for this keyword argument, move the check to be a compile-time one, rather than an exception raised at runtime.

numba/np/arraymath.py
```diff
@@ -4836,3 +4842,162 @@ def impl(filt, trim='fb'):
...
+        raise TypingError('in1d: first two args must be array-like')
+    if not (isinstance(assume_unique, (types.Boolean, bool))):
+        raise TypingError('in1d: Argument "assume_unique" must be boolean')
+    if not (isinstance(invert, (types.Boolean, bool))):
+        raise TypingError('in1d: Argument "invert" must be boolean')
+    if not (isinstance(kind, (types.UnicodeType, str))):
+        raise TypingError('in1d: Argument "kind" must be string')
+
+    # https://github.com/numpy/numpy/blob/
+    # 03b62604eead0f7d279a5a4c094743eb29647368/numpy/lib/arraysetops.py#L525
```
guilhermeleobas, line 4916: Same

numba/np/arraymath.py
```diff
@@ -4836,3 +4842,162 @@ def impl(filt, trim='fb'):
...
+        raise TypingError('in1d: Argument "invert" must be boolean')
+    if not (isinstance(kind, (types.UnicodeType, str))):
+        raise TypingError('in1d: Argument "kind" must be string')
+
+    # https://github.com/numpy/numpy/blob/
+    # 03b62604eead0f7d279a5a4c094743eb29647368/numpy/lib/arraysetops.py#L889
+    def np_isin_impl(element, test_elements, assume_unique=False, invert=False,
+                     kind="sort"):
+        if kind != "sort":
+            raise ValueError('Only kind="sort" is supported')
```
guilhermeleobas, line 4998: Move this check to run during typing

guilhermeleobas (commented):
Thanks, @jaredjeya. Code seems fine as it is almost a 1:1 port of what NumPy does. Just have some minor comments that should be easy to address.

numba/np/arraymath.py
jaredjeya: How would I do this? Using a check for ``Literal[str]("sort")``?

guilhermeleobas: Yes! But actually, let's keep it like this.

guilhermeleobas:
@jaredjeya, could you include the tests from NumPy for the set of functions added in this PR?
https://github.com/numpy/numpy/blob/b0371ef240560e78b651a5d7c9407ae3212a3d56/numpy/lib/tests/test_arraysetops.py#L15

jaredjeya:
So what I've done here is add the numpy test cases to the tests I'd already written, rather than writing new tests entirely. Also fixed that I'd been casting all the inputs to arrays before testing them, so I wasn't testing the functionality with lists.

Otherwise I also added a check to catch at compile time if someone calls these functions with incorrect literal values for ``kind``, but I don't enforce that it's a compile time constant. I also changed some of the error messages for readability and consistency.

jaredjeya:
I forgot to add the attribution for test_setops_manyways, which should read:

```
# https://github.com/numpy/numpy/blob/b0371ef240560e78b651a5d7c9407ae3212a3d56/numpy/lib/tests/test_arraysetops.py#L588 # noqa: E501
```

Since it's just a comment is there a way to add that in without re-running all the tests?

numba/np/arraymath.py
```diff
@@ -3589,9 +3589,10 @@ def jit_np_intersect1d(ar1, ar2, assume_unique=False):
     # https://github.com/numpy/numpy/blob/v1.19.0/numpy/lib
     # /arraysetops.py#L347-L441
     if not (type_can_asarray(ar1) or type_can_asarray(ar2)):
-        raise TypingError('intersect1d: first two args must be array-like')
+        raise TypingError('np.intersect1d: first two args must be array-like')
```
guilhermeleobas, line 3592: Although is it a common convention to import the NumPy package as `np`, it is not strictly true. I'd just keep the message as it was before, without the `np.` prefix.

numba/np/arraymath.py
```diff
@@ -4982,20 +4981,23 @@ def np_in1d_impl(ar1, ar2, assume_unique=False, invert=False, kind="sort"):
 def jit_np_isin(element, test_elements, assume_unique=False, invert=False,
                 kind="sort"):
```
guilhermeleobas, line 4982: The default value for `kind` should be `None`, not `"sort"`.

The following should work on your branch:
```python
import numpy as np
from numba import njit

@njit
def foo():
    return np.isin(3, 4, kind=None)

print(foo())
```

numba/np/arraymath.py
```diff
@@ -4904,28 +4905,28 @@ def np_setdiff1d_impl(ar1, ar2, assume_unique=False):
 @overload(np.in1d)
 def jit_np_in1d(ar1, ar2, assume_unique=False, invert=False, kind="sort"):
```
guilhermeleobas, line 4906: Same thing here, the default value for `kind` is `None`, not `"sort"`.

guilhermeleobas (changes requested):

numba/np/arraymath.py
jaredjeya: The reason I've done it like I did is because I didn't implement the numpy ``kind="table"`` method. (The output is equivalent, but I suppose the performance is higher in certain cases with the table method (I haven't tested)). When ``kind=None``, the numpy code will use the table method if certain conditions are met, otherwise the sorting method, so we don't support that either -- strictly, we only support ``kind="sort"``. So I thought ``np.isin(3, 4, kind=None)`` *should* fail.

If you think it should work I'm happy to change it, or otherwise we just say we don't support the "kind" argument and leave it at that. Or, I could even implement the "table" method, it isn't that much more complicated than the sort method, and then we fully support the "kind" argument.

guilhermeleobas: Okay. I think it is best to match as close as possible the NumPy signature but keep a check internally for `kind == "sort"`.

jaredjeya:
I decided to add some tests to make sure the correct errors result when nonsense inputs are supplied, including e.g. `kind="foo"`. However I can't for the life of me figure out how to get numba to actually compile these as literals.

This is the full type-checking code within the `@overload` function:

```python
if kind is not None:
    if (isinstance(kind, types.StringLiteral)
            and kind.literal_value != "sort"):
        raise NumbaValueError('isin: Only kind="sort" is supported')
    elif not isinstance(kind, (types.UnicodeType, str, types.NoneType)):
        raise TypingError('isin: Argument "kind" must be a string or None')
```

this is the code in the actual implementation,

```python
if kind is not None and kind != "sort":
    raise ValueError('isin: Only kind="sort" is supported')
```

and this is the testing code:

```python
@njit()
def table_isin(a, b):
    return np.isin(a, b, kind="table")
with self.assertRaises(NumbaValueError):
    table_isin(a, b)
```

I also tried `kind=literal("table")` and `kind=literally("table")`. But the code always raises a ValueError at runtime rather than a NumbaValueError at compilation time, and if I insert debug print statements, I see `repr(kind) == unicode_type` and `type(kind) == <class 'numba.core.types.misc.UnicodeType'>`.

What am I doing wrong? Other functions e.g. `np.searchsorted` use identical patterns in their tests.

jaredjeya:
Nevermind, I sorted it

numba/np/arraymath.py
```diff
@@ -4836,3 +4843,169 @@ def impl(filt, trim='fb'):
...
+            raise TypingError('in1d: Argument "kind" must be a string or None')
+        elif kindval != "sort":
+            raise NumbaValueError('in1d: Only kind="sort" is supported')
+
+    # https://github.com/numpy/numpy/blob/03b62604eead0f7d279a5a4c094743eb29647368/numpy/lib/arraysetops.py#L889 # noqa: E501
+    def np_isin_impl(element, test_elements, assume_unique=False, invert=False,
+                     kind=None):
+
+        if kind is not None and kind != "sort":
+            raise ValueError('isin: Only kind="sort" is supported')
```
guilhermeleobas, line 5005: If `kind=None`, NumPy will use `table` by default, which is not supported in this PR. Change the clause to something like this:

```suggestion
        if kind != "sort":
            raise ValueError('isin: Only kind="sort" is supported')
```

This will force a user to spell the `kind="sort"` keyword argument and not assume Numba supports the `table` option.

numba/np/arraymath.py
```diff
@@ -4836,3 +4843,169 @@ def impl(filt, trim='fb'):
...
+    if not (isinstance(invert, (types.Boolean, bool))):
+        raise TypingError('isin: Argument "invert" must be boolean')
+
+    if not (kind is None or isinstance(kind, types.NoneType)):
+        kindval = getattr(kind, "literal_value", kind)
+        if not isinstance(kindval,
+                          (types.UnicodeType, str, types.StringLiteral)):
+            raise TypingError('in1d: Argument "kind" must be a string or None')
+        elif kindval != "sort":
+            raise NumbaValueError('in1d: Only kind="sort" is supported')
```
guilhermeleobas, line 4998: That's fine, but we can keep only the `ValueError` to simplify the codebase.

numba/np/arraymath.py
```diff
@@ -4836,3 +4843,169 @@ def impl(filt, trim='fb'):
...
+    if not isinstance(invert, (types.Boolean, bool)):
+        raise TypingError('in1d: Argument "invert" must be boolean')
+
+    if not (kind is None or isinstance(kind, types.NoneType)):
+        kindval = getattr(kind, "literal_value", kind)
+        if not isinstance(kindval,
+                          (types.UnicodeType, str, types.StringLiteral)):
+            raise TypingError('in1d: Argument "kind" must be a string or None')
+        elif kindval != "sort":
+            raise NumbaValueError('in1d: Only kind="sort" is supported')
```
guilhermeleobas, line 4918: Same

numba/np/arraymath.py
```diff
@@ -4836,3 +4843,169 @@ def impl(filt, trim='fb'):
...
+
+    def np_in1d_impl(ar1, ar2, assume_unique=False, invert=False, kind=None):
+        # https://github.com/numpy/numpy/blob/03b62604eead0f7d279a5a4c094743eb29647368/numpy/lib/arraysetops.py#L525 # noqa: E501
+        if kind is not None and kind != "sort":
+            raise ValueError('in1d: Only kind="sort" is supported')
```
guilhermeleobas, line 4923: Same

guilhermeleobas (changes requested):

numba/np/arraymath.py
jaredjeya: > If kind=None, NumPy will use table by default, which is not supported in this PR.

Yes, that is why I didn't implement kind=None in the first place. I'll change it back.

guilhermeleobas: The function signature needs to be with `kind=None` as it is what NumPy specifies. But let's keep a check to only run the function if the user specifies `kind="sort"`.

Let me know if you have any questions regargind this.

numba/np/arraymath.py
jaredjeya: To be clear, I should get rid of this whole section?

jaredjeya: Actually I've reduced it to just,

```python
if not isinstance(kind, (types.UnicodeType, str, types.StringLiteral)):
    raise TypingError('in1d: Only kind="sort" is supported')
```

which makes sure kind=None doesn't make it past compilation. It seemed to be causing some type inference issues if I didn't have that.

jaredjeya:
@guilhermeleobas Tests seem to be failing now because `numpy < 1.24` doesn't support the `kind` keyword, and because we now force this to be included explicitly, the numpy and numba behaviours diverge. What do you suggest?

According to the [release notes](https://numpy.org/doc/stable/release/1.24.0-notes.html#faster-version-of-np-isin-and-np-in1d-for-integer-arrays), pre-1.24, the kind="sort" behaviour was the default. Newer versions automatically pick the correct algorithm.

It might not be too hard to implement kind="table", but it would take me a while as I'm quite busy atm. Otherwise we say we drop the keyword entirely and say we're only replicating pre-1.24 behaviour? The [documentation](https://numba.readthedocs.io/en/latest/extending/overloading-guide.html#implementing-overload-for-numpy-functions) suggests using a version gate but I don't know what that means in this context.

guilhermeleobas:
> According to the [release notes](https://numpy.org/doc/stable/release/1.24.0-notes.html#faster-version-of-np-isin-and-np-in1d-for-integer-arrays), pre-1.24, the kind="sort" behaviour was the default. Newer versions automatically pick the optimal algorithm.

I think we can have an if-else statement to support the kind argument if NumPy version if greater or equal 1.24:

```python
if numpy_version < (1, 24):
    @overload(np.isin)
    def jit_np_isin(element, test_elements, assume_unique=False, invert=False):
         ....

else:
    @overload(np.isin)
    def jit_np_isin(element, test_elements, assume_unique=False, invert=False,
                    kind=None):
        ...
```

You may have to update the tests as well to only pass kind="sort" if NumPy >= 1.24

> Otherwise we can drop the keyword entirely and say we're only replicating pre-1.24 behaviour?

That's also an option. Choose the one that is easiest for you to implement.

jaredjeya:
I decided it's simpler and less confusing to the end user (especially working across multiple versions) if we just specify that "kind" isn't supported, and that behaviour matches that of pre-1.24.

Unfortunately it looks like the build is failing due to changes to put in the new typing system - I'm not really sure what to do here?

guilhermeleobas (approved):
Sounds good to me. I'm happy with the current changes. Thanks for your effort in this PR.

esc:
Thank you for the patch.

PR #9342, opened 2023-12-08
Fix flake8 checks for v6.1.0

gmarkall:
To fix currently-failing CI checks.


stuartarchibald (approved):
Thanks for the patch!

PR #9346, opened 2023-12-08
Setting ``NUMBA_CAPTURED_ERRORS=old_style`` will now raise warnings.

sklam:
Additional changes in `errors.py` is to workaround issues with circular imports.

### DIFF
--- a/docs/upcoming_changes/9346.deprecation.rst
+++ b/docs/upcoming_changes/9346.deprecation.rst
@@ -5,4 +5,4 @@ As per deprecation schedule of old-style error-capturing, explicitly setting
 ``NUMBA_CAPTURED_ERRORS=old_style`` will raise deprecation warnings. 
 This release is the last to use "old_style" as the default.
 Details are documented at 
-https://numba.readthedocs.io/en/stable/reference/deprecation.html#deprecation-of-old-style-numba-captured-errors
+https://numba.readthedocs.io/en/0.58.1/reference/deprecation.html#deprecation-of-old-style-numba-captured-errors
--- a/docs/upcoming_changes/9346.deprecation.rst
+++ b/docs/upcoming_changes/9346.deprecation.rst
@@ -1,0 +1,8 @@
+Explicitly setting ``NUMBA_CAPTURED_ERRORS=old_style`` will raise deprecation warnings
+======================================================================================
+
+As per deprecation schedule of old-style error-capturing, explicitly setting 
+``NUMBA_CAPTURED_ERRORS=old_style`` will raise deprecation warnings. 
+This release is the last to use "old_style" as the default.
+Details are documented at 
+https://numba.readthedocs.io/en/stable/reference/deprecation.html#deprecation-of-old-style-numba-captured-errors
--- a/numba/core/config.py
+++ b/numba/core/config.py
@@ -4,6 +4,7 @@ import os
 import re
 import shutil
 import warnings
+import traceback
 
 # YAML needed to use file based Numba config
 try:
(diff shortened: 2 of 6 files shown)


numba/tests/test_errorhandling.py
```diff
@@ -469,10 +471,54 @@ def foo(x):
...
+        # if NUMBA_CAPTURED_ERRORS=old_style
+        env = {"NUMBA_CAPTURED_ERRORS": "old_style"}
+        _out, err = self._run_in_separate_process(code, env)
+        self.assertIn(_old_style_deprecation_msg, err.decode())
+
+        # Check that the deprecated message is NOT shown
+        # if NUMBA_CAPTURED_ERRORS is unset
+        env = {"NUMBA_CAPTURED_ERRORS": ""}
+        _out, err = self._run_in_separate_process(code, env)
+        # Check that the deprecated message is shown
```
kc611, line 511:
```suggestion
        # Check that the deprecated message is not shown
```

numba/tests/test_errorhandling.py
```diff
@@ -469,10 +471,54 @@ def foo(x):
...
+        self.assertNotIn("NumbaPendingDeprecationWarning", err.decode())
+
+        # Check that the deprecated message is NOT shown
+        # if NUMBA_CAPTURED_ERRORS=new_style
+        env = {"NUMBA_CAPTURED_ERRORS": "new_style"}
+        _out, err = self._run_in_separate_process(code, env)
+        # Check that the deprecated message is shown
```
kc611, line 518:
```suggestion
        # Check that the deprecated message is not shown
```

kc611 (commented):
Following nitpicks rest of it LGTM!

numba/tests/test_errorhandling.py
```diff
@@ -469,10 +471,54 @@ def foo(x):
...
+    def test_old_style_deprecation_on_import(self):
+        from numba.core.config import _old_style_deprecation_msg
+
+        code = """
+        import numba
+        """
+        # Check that the deprecated message is shown
+        # if NUMBA_CAPTURED_ERRORS=old_style
+        env = {"NUMBA_CAPTURED_ERRORS": "old_style"}
+        _out, err = self._run_in_separate_process(code, env)
```
stuartarchibald, line 504: Can `TestCase.run_test_in_subprocess` be used with the `PYTHONWARNINGS` env var to create the same effect as the above `_run_in_separate_process` with the `-Wall` option?

docs/upcoming_changes/9346.deprecation.rst
```diff
@@ -0,0 +1,8 @@
+Explicitly setting ``NUMBA_CAPTURED_ERRORS=old_style`` will raise deprecation warnings
+======================================================================================
+
+As per deprecation schedule of old-style error-capturing, explicitly setting 
+``NUMBA_CAPTURED_ERRORS=old_style`` will raise deprecation warnings. 
+This release is the last to use "old_style" as the default.
+Details are documented at 
+https://numba.readthedocs.io/en/stable/reference/deprecation.html#deprecation-of-old-style-numba-captured-errors
```
stuartarchibald, line 8: Should this reference a specific release which has this deprecation notice anchor, else this link from the release notes to the docs will eventually become invalid?

numba/core/config.py
```diff
@@ -193,10 +209,11 @@ def _readenv(name, ctor, default):
                 return default() if callable(default) else default
             try:
                 return ctor(value)
-            except Exception as e:
+            except Exception:
                 warnings.warn(f"Environment variable '{name}' is defined but "
                               f"its associated value '{value}' could not be "
-                              f"parsed.\nThe parse failed with exception: {e}.",
+                              "parsed.\nThe parse failed with exception:\n"
```
stuartarchibald, line 215: Should this be removed? The exception often explains what the issue is, for example: 
https://github.com/numba/numba/blob/e56e93e0d65170746befec5f92c903b1cc364b5f/numba/core/config.py#L39-L41

stuartarchibald (commented):
Thanks for the patch, couple of minor things to look at else looks good.

numba/tests/test_errorhandling.py
sklam: I chose to not use `TestCase.run_test_in_subprocess` because that will use `numba.runtests` which implicitly import numba. It end up making it very hard to debug if something goes wrong.

numba/core/config.py
sklam: Since this is an warning, the exception is never shown. Without the traceback, it's hard to tell what's wrong. I added this because I ended with some a weird bug at one point and just looking at the messages is not useful for debugging.

docs/upcoming_changes/9346.deprecation.rst
sklam: i can make it reference the last release

sklam: resolved in https://github.com/numba/numba/pull/9346/commits/b326582276620e74b48192422be184961363aa3e

numba/tests/test_errorhandling.py
stuartarchibald: Ah, ok, that makes sense, thanks for explaining.

docs/upcoming_changes/9346.deprecation.rst
stuartarchibald: Thanks.

stuartarchibald:
@kc611 @sklam thanks for the review and updates. I approve the single commit with SHA b326582 with reference to updates reflecting my comments above.

numba/core/config.py
stuartarchibald: OK, the error string is rendered into the message, but if it's not useful then removing it seems fine.

stuartarchibald (approved):
As this was approved by @kc611 at aa5347e, I approve from there through to b326582 in relation to resolving review comments. Thanks for patch @sklam and for reviewing @kc611!

PR #9347, opened 2023-12-08
Make Numba's deprecation warnings subclasses of the builtin ones.

sklam:
This will allow filter to be set to suppress all `DeprecationWarning`s including the one from Numba.

### DIFF
--- a/numba/tests/test_warnings.py
+++ b/numba/tests/test_warnings.py
@@ -6,7 +6,12 @@ import numpy as np
 
 import unittest
 from numba import jit
-from numba.core.errors import NumbaWarning, deprecated, NumbaDeprecationWarning
+from numba.core.errors import (
+    NumbaWarning,
+    deprecated,
+    NumbaDeprecationWarning,
+    NumbaPendingDeprecationWarning,
+)
 from numba.core import errors
 from numba.tests.support import ignore_internal_warnings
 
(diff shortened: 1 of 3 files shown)


gmarkall:
That test fail looked pretty odd, so I'm just re-running it.

numba/tests/test_warnings.py
```diff
@@ -234,6 +239,49 @@ def test_disable_performance_warnings(self):
...
+            warnings.simplefilter('ignore', category=PendingDeprecationWarning)
+            warnings.warn(DeprecationWarning("this is ignored"))
+            warnings.warn(PendingDeprecationWarning("this is ignored"))
+            warnings.warn(NumbaDeprecationWarning("this is ignored"))
+            warnings.warn(NumbaPendingDeprecationWarning("this is ignored"))
+            with self.assertRaises(NumbaWarning):
+                warnings.warn(NumbaWarning("this is not ignored"))
+
+    def test_filter_ignore_numba_deprecation_only(self):
+        # Make a filter that ignore Numba's deprecation warning but raises
```
stuartarchibald, line 257:
```suggestion
        # Make a filter that ignores Numba's deprecation warnings but raises on
```

numba/tests/test_warnings.py
```diff
@@ -234,6 +239,49 @@ def test_disable_performance_warnings(self):
...
+
+            with self.assertRaises(DeprecationWarning):
+                warnings.warn(DeprecationWarning("this is not ignored"))
+            with self.assertRaises(PendingDeprecationWarning):
+                warnings.warn(PendingDeprecationWarning("this is not ignored"))
+
+            warnings.warn(NumbaDeprecationWarning("this is ignored"))
+            warnings.warn(NumbaPendingDeprecationWarning("this is ignored"))
+
+            # now even Numba ones are raising
```
stuartarchibald, line 274:
```suggestion
            # now make it so that Numba deprecation warnings are raising
```

stuartarchibald (commented):
Thanks for making this change @sklam, this will help users with filtering Numba's deprecation and pending deprecation warnings via standard filters for these sorts of warnings. Patch looks good, just a couple of wording suggestions to look at. Thanks again!

stuartarchibald (approved):

PR #9352, opened 2023-12-11
Removing object mode fallback from `@jit`.

stuartarchibald:
This removes the long deprecated "object mode fallback" behaviour found in the `numba.jit` decorator (and other similar decorators). The removal attempts to be "minimal" in case there are unknown issues with making this change with view of making it easy to revert.


numba/core/decorators.py
```diff
@@ -143,34 +143,27 @@ def bar(x, y):
         raise DeprecationError(_msg_deprecated_signature_arg.format('argtypes'))
     if 'restype' in options:
         raise DeprecationError(_msg_deprecated_signature_arg.format('restype'))
-    if options.get('nopython', False) and forceobj:
+    nopython = options.get('nopython', None)
+    if nopython is True and forceobj:
```
sklam, line 147: Does it matter that `nopython` is exactly the `True` object? or does it just need to be truthy?

numba/core/decorators.py
```diff
@@ -143,34 +143,27 @@ def bar(x, y):
...
-                   "argument is currently False, but it will be changed to "
-                   f"True in Numba 0.59.0. See {url} for details.")
-        else:
-            msg = ("The keyword argument 'nopython=False' was supplied. From "
-                   "Numba 0.59.0 the default is being changed to True and use "
-                   "of 'nopython=False' will raise a warning as the "
-                   f"argument will have no effect. See {url} for "
-                   "details.")
-        warnings.warn(NumbaDeprecationWarning(msg), stacklevel=2)
+    if nopython is False:
```
sklam, line 155: same question about truthiness here.

numba/core/decorators.py
sklam:
```python
In [4]: @jit(nopython=1, forceobj=True)
   ...: def foo(): return

In [5]: foo()
```

sklam (commented):
Besides the CI failure on `test_globals_jit`, my only comment is below:

stuartarchibald:
> Besides the CI failure on `test_globals_jit`, my only comment is below:

Hopefully fixed in 774dc75.

numba/core/decorators.py
stuartarchibald: Assertion added to enforce `bool` in c3b8c33.

numba/core/decorators.py
stuartarchibald: Assertion added to enforce `bool` in c3b8c33.

stuartarchibald:
Note: the documentation/deprecation notices need updating to reflect this change.

sklam:
The CI failure is showing more test failures. Many seems to be resolvable by adding `forceobj=True`. Some looplifting tests may no longer be valid since (esp the "reject" ones since they are testing fallback behavior).

In addition, we might need to remove `Flags.enable_pyobject`.

gmarkall:
gpuci run tests

sklam (approved):

PR #9353, opened 2023-12-11
Remove `numba.generated_jit`

stuartarchibald:
As title, removes source code and paths that related to `generated_jit` as per the deprecation schedule. Also removes reference to the same from the documentation.


sklam:
Locally it's failing on 

```pytb
FAIL: test_error_count (numba.tests.test_dispatcher.TestNoRetryFailedSignature)
  File "/numba/numba/tests/test_dispatcher.py", line 1109, in test_error_count
    self.assertEqual(ct_bad, 1)
AssertionError: 40 != 1
```

stuartarchibald:
> Locally it's failing on
> 
> ```
> FAIL: test_error_count (numba.tests.test_dispatcher.TestNoRetryFailedSignature)
>   File "/numba/numba/tests/test_dispatcher.py", line 1109, in test_error_count
>     self.assertEqual(ct_bad, 1)
> AssertionError: 40 != 1
> ```

Ah, that reminds me, this was why I had a variant of #9259 as part of this patch originally.

numba/tests/test_dictimpl.py
```diff
@@ -606,7 +606,7 @@ def test_sizing(self):
         for i in range(1, 8):
             self.check_sizing(key_size=i, val_size=i, nmax=2**i)
 
-    def test_parametrized_types(self):
+    def test_parameterized_types(self):
```
sklam, line 609: but pytest use parametrized 😛

numba/tests/test_dispatcher.py
```diff
@@ -1052,6 +1052,9 @@ def foo(x):
 
         self.run_test(foo)
 
+    @unittest.skip(("Unconditionally skip. @overload does not have an error "
```
gmarkall, line 1055: Does this always fail? Perhaps it should be an `@xfail` so that once it starts passing, the fact it's started passing will be an error and reinstating the test won't be missed?

stuartarchibald: Good point, will adjust, thanks.

stuartarchibald: Done in ddd40cd.

sklam (approved):

PR #9330, opened 2023-12-04
Refactor and remove legacy APIs/testing internals.

stuartarchibald:
This PR removes the following (and does some minimal refactoring in the process):
* `numba.core.compiler.compile_isolated` this function was used a lot in Numba's test suite and exists from a time where the compiler was less isolated than it is now. The function itself also relies on "nested contexts" within the same target which has been discovered to cause numerous issues in practice.
* `numba.tests.support.TestCase::run_nullary_func` this method isn't really used very much in the test suite or needed any more, refactoring to remove it was trivial.
* `numba.tests.support.CompilationCache`, this was used in the test suite from a time when the Numba dispatchers were less well isolated and `compile_isolated` needed somewhere to cache results, this isn't needed any more since `compile_isolated` is removed.
* "nested context" as a concept from the `numba.core.registry.CPUTarget`, this was used a lot in conjunction with the now removed `compile_isolated`, it has a history of causing hard to debug problems.

Notes for the reviewer:
1. These patches are reasonably isolated to one file being changed per commit.
2. Patch b91050302 fixes a bug in `_typeof.cpp` that was hidden by `compile_isolated`, it is therefore less "easy" to review.
3. Patch 834ec6b53 also includes a minor refactor to `test_svml.py` to better use existing testing infrastructure to achieve isolation.

PR #9354, opened 2023-12-11
Alter the `test_print.py` test to match previous.

stuartarchibald:
This alters the "style" of the test to match something similar to what existed prior to the merge of #9330 (removal of `compile_isolated`) with view of "fixing" an NRT reference leak.

### DIFF
--- a/numba/tests/test_print.py
+++ b/numba/tests/test_print.py
@@ -44,7 +44,7 @@ class TestPrint(EnableNRTStatsMixin, TestCase):
         pyfunc = print_value
 
         def check_values(typ, values):
-            cfunc = njit((typ,))(pyfunc)
+            cfunc = njit((typ,))(pyfunc).overloads[(typ,)].entry_point
             for val in values:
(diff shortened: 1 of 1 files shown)


sklam:
This just need testing on the farm

sklam:
This test actually failed on the one that it should be fixing: https://dev.azure.com/numba/numba/_build/results?buildId=16486&view=logs&j=f6408ea9-ccc2-513a-5c09-13df1a476ff5&t=19427a42-3c67-579d-6c12-e1fdf54a08df&l=10786

sklam:
I reran AzureCI to see if it's randomly failing and sadly it is.

stuartarchibald:
> I reran AzureCI to see if it's randomly failing and sadly it is.

I wonder if there's now a few variations of some matching implementation present due to the shared context and a) there shouldn't be and b) because there is it's a dictionary order/hash seed problem.

stuartarchibald:
This did not have the intended effect. Closing.

Issue #9355, opened 2023-12-13
Intermittent NRT leak with `print(reflected list)`


stuartarchibald:
`main` branch as of e56e93e (and probably from the point of merging #9330) has an intermittent issue in the test:
```
numba.tests.test_print.TestPrint.test_print_values
```
specifically this part of the test:
https://github.com/numba/numba/blob/e56e93e0d65170746befec5f92c903b1cc364b5f/numba/tests/test_print.py#L73-L77

the issue is that the `assertNoNRTLeak()` fails with:
```
AssertionError 2 != 3 : number of data allocs != number of data frees
```

Current thoughts are that it is likely that adjusting the code base and tests to not use `compile_isolated` (#9330) has unveiled a latent bug either localised to `ir.Print` lowering or perhaps `box`/`unbox`ing of reflected lists. Another possibility is that it's to do with dictionary ordering/hash seed and what is in the context's caches, perhaps there's more than one definition/match for some part of the compilation path (and one of them is wrong/doesn't apply) and the first found wins, this might explain the intermittent nature of the problem.

PR #9356, opened 2023-12-13
Refactor print tests to avoid NRT leak issue.

stuartarchibald:
This refactors the basic value printing tests to avoid an intermittent issue with an NRT leak on printing reflected lists, see issue #9355 for details of the problem. The potentially leaking test is moved to a separate test and is skipped until the issue is resolved.


kc611 (approved):

stuartarchibald:
Thanks for the review @kc611 !

PR #9358, opened 2023-12-13
Remove note about OpenMP restriction in wheels.

stuartarchibald:
As title.


stuartarchibald:
The note this PR is removing seems out dated, the wheel builds report the OpenMP threading layer as supported.

sklam (approved):

PR #9359, opened 2023-12-13
Fix test_jit_module test against objmode fallback.

stuartarchibald:
This fixes a failing test, the use of the `forceobj` kwarg in the `@jit` decorator sets the `TargetOption` flag `forceobj`, and now also sets the `nopython` flag to `False`. The test is updated to reflect this.


sklam (approved):
LGTM.
Also manually confirmed that the test is passing with the fix.

PR #9362, opened 2023-12-14
Fix ``np.MachAr`` warning matching in test.

sklam:
Somehow #9347 changed the behavior of the filter.
The invisible characters from termcolor highlight now matters. But comparing the warning message before and after that PR does not reveal any visible difference.

Also added a `simplefilter("ignore")` to suppress any warning config from `PYTHONWARNINGS` in the python process.

### DIFF
--- a/numba/tests/test_np_functions.py
+++ b/numba/tests/test_np_functions.py
@@ -6239,8 +6239,9 @@ def foo():
         # typing part of the `np.MachAr` overload, which may already have been
         # executed for the given types and so an empty in memory cache is
         # needed.
-        msg = r'`np.MachAr` is deprecated \(NumPy 1.22\)'
+        msg = r'.*`np.MachAr` is deprecated \(NumPy 1.22\)'
         with warnings.catch_warnings(record=True) as w:
+            warnings.simplefilter('ignore')  # override warning behavior
             warnings.filterwarnings("always", message=msg,
                                     category=NumbaDeprecationWarning,)
             f = njit(lambda : np.MachAr().eps)


sklam:
To test this, make sure the environment has `colorama` and  `NUMBA_DISABLE_ERROR_MESSAGE_HIGHLIGHTING` is cleared.

Run: `NUMBA_DISABLE_ERROR_MESSAGE_HIGHLIGHTING= SUBPROC_TEST=1 python  runtests.py -k test_np_MachAr_deprecation_np122 -v`

sklam:
BFID: `numba_yaml_423`

stuartarchibald:
Thanks for fixing this @sklam. I debugged this a bit, AFAICT the actual strings in the message are the same, what I think is changing is what happens internally as part of the matching process, now the Numba warnings inherit from the standard library warnings the way the message strings are normalized internally is different:
https://github.com/python/cpython/blob/e0c6995b4fa9803a886fe1d88d6a9aec166a99e8/Python/_warnings.c#L657-L674

stuartarchibald (approved):
Thanks for the patch. I tested `main` manually against the reproducer in https://github.com/numba/numba/pull/9362#issuecomment-1855085927 and the test fails, application of this patch then makes it pass. The changes also make sense with respect to some debugging I did https://github.com/numba/numba/pull/9362#issuecomment-1855628388.

sklam:
CI passed and enough buildfarm config has passed (e.g just windows ones are taking a long time and we don't want to wait for it)

Issue #9322, opened 2023-12-01
Add `float("inf")` support


Azmisov:
`float("inf")` is probably the widest used variant for specifying infinity, so would be nice to support. Workaround is to use `np.inf` instead, so not super important.

w0rp:
I'll watch this issue. Pretty important to support the basic functionality of Python builtins through some means.

guilhermeleobas:
The issue is that you cannot create `float('inf')` inside the jitted function:

```python
from numba import njit

@njit
def fn():
    return float('inf')

print(fn())
```

PR #9378, opened 2023-12-23
Add `float(<string literal>)` ctor

guilhermeleobas:
Fixes: https://github.com/numba/numba/issues/9322

### DIFF
--- a/docs/upcoming_changes/9378.new_feature.rst
+++ b/docs/upcoming_changes/9378.new_feature.rst
@@ -1,4 +1,4 @@
-Add ``float('INF')`` and ``float('-INF')`` ctor
-===============================================
+Add ``float(string_literal)`` ctor
+==================================
 
-Support for ``float('INF')`` and ``float('-INF')`` is added.
+Support for ``float(string_literal)`` is added.
--- a/docs/upcoming_changes/9378.new_feature.rst
+++ b/docs/upcoming_changes/9378.new_feature.rst
@@ -1,4 +1,4 @@
-Add ``float(string_literal)`` ctor
+Add ``float(<string literal>)`` ctor
 ==================================
 
-Support for ``float(string_literal)`` is added.
+Support for ``float(<string literal>)`` is added.
--- a/docs/upcoming_changes/9378.new_feature.rst
+++ b/docs/upcoming_changes/9378.new_feature.rst
@@ -1,0 +1,4 @@
+Add ``float('INF')`` and ``float('-INF')`` ctor
+===============================================
+
+Support for ``float('INF')`` and ``float('-INF')`` is added.
--- a/docs/upcoming_changes/9378.new_feature.rst
+++ b/docs/upcoming_changes/9378.new_feature.rst
@@ -1,4 +1,4 @@
 Add ``float(<string literal>)`` ctor
-==================================
+====================================
 
 Support for ``float(<string literal>)`` is added.
--- a/numba/core/typing/builtins.py
+++ b/numba/core/typing/builtins.py
@@ -995,6 +995,12 @@ class Float(AbstractTemplate):
 
         [arg] = args
 
+        if isinstance(arg, types.StringLiteral):
+            if arg.literal_value.lower in ('inf', '-inf'):
+                msg = 'float(string) only supported for "inf" value'
+                raise errors.NumbaTypeError(msg)
+            return signature(types.float64, arg)
+
         if arg not in types.number_domain:
             raise errors.NumbaTypeError("float() only support for numbers")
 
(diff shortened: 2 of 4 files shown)


numba/tests/test_builtins.py
```diff
@@ -542,7 +542,7 @@ def test_float(self, flags=forceobj_flags):
             self.assertPreciseEqual(cfunc(x), pyfunc(x), prec='single')
 
         cfunc = jit((types.string,), **flags)(pyfunc)
```
gmarkall, line 544: Is this test using object mode to pass? A string literal seems to be required for your `float("inf")` implementation, so I'm having trouble seeing how else this could be working.

numba/cpython/builtins.py
```diff
@@ -298,6 +298,14 @@ def int_impl(context, builder, sig, args):
     return impl_ret_untracked(context, builder, sig.return_type, res)
 
 
+@lower_builtin(float, types.StringLiteral)
+def float_inf_impl(context, builder, sig, args):
+    [ty] = sig.args
+    assert ty.literal_value.lower() in ('inf', '-inf')
+    res = context.get_constant(sig.return_type, np.inf)
```
gmarkall, line 305: I guess this should depend on the sign of the input?

gmarkall (changes requested):
This seems not to work for negative infinity:

```python
from numba import njit

def neginf():
    return float("-inf")

pyresult = neginf()
nbresult = njit(neginf)()

assert pyresult == nbresult, f"{pyresult} != {nbresult}"
```

gives

```
$ python repro.py 
  File "/home/gmarkall/numbadev/issues/9378/repro.py", line 9, in <module>
    assert pyresult == nbresult, f"{pyresult} != {nbresult}"
           ^^^^^^^^^^^^^^^^^^^^
AssertionError: -inf != inf
```

numba/cpython/builtins.py
guilhermeleobas: Yes! My bad here. Will fix this in a bit.

numba/tests/test_builtins.py
guilhermeleobas: Hi Graham, I added a new test case with object mode disabled. The current implementation is specific for `float(+/- 'inf')` and if you think it is needed, I can work on arbitrary string -> float conversion.

numba/core/typing/builtins.py
```diff
@@ -995,6 +995,16 @@ def generic(self, args, kws):
 
         [arg] = args
 
+        if isinstance(arg, types.UnicodeType):
+            msg = 'argument must be a string literal'
```
gmarkall, line 999: The argument could be a number, it needn't be a string literal. Perhaps:

```suggestion
            msg = 'string argument must be a literal'
```

reflects the issue more accurately?

Also - I think this branch is untested.

gmarkall (changes requested):
Many thanks for the fixes - I think the behaviour is now correct. There's just a couple of small additional comments about the non-literal string error case on the diff.

guilhermeleobas:
CI failure do not seem to be related to this PR.

FAIL: test_unsigned_refusal_to_vectorize (numba.tests.test_parfors.TestParforsVectorizer.test_unsigned_refusal_to_vectorize)
This checks that if fastmath is set and the underlying hardware

gmarkall (dismissed):

stuartarchibald:
@guilhermeleobas any chance you could change the PR title to something that reflects the current patch please (i.e. it's `float(<string literal>)` that's supported now?!) so that when it's merged the merge commit message contains this updated message. Many thanks.

guilhermeleobas:
> @guilhermeleobas any chance you could change the PR title to something that reflects the current patch please (i.e. it's `float(<string literal>)` that's supported now?!) so that when it's merged the merge commit message contains this updated message. Many thanks.

Done!

stuartarchibald:
> > @guilhermeleobas any chance you could change the PR title to something that reflects the current patch please (i.e. it's `float(<string literal>)` that's supported now?!) so that when it's merged the merge commit message contains this updated message. Many thanks.
> 
> Done!

Great, thank you!

docs/upcoming_changes/9378.new_feature.rst
```diff
@@ -0,0 +1,4 @@
+Add ``float(<string literal>)`` ctor
+==================================
```
stuartarchibald, line 2:
```suggestion
Add ``float(<string literal>)`` ctor
====================================
```
think that this will fix the RST.

stuartarchibald (approved):
Following on from @gmarkall's approval of 177dc48, I approve 177dc48...62ede8f. Thanks for the patch @guilhermeleobas, thanks for reviewing @gmarkall.

PR #9390, opened 2024-01-12
Add .tobytes() method

guilhermeleobas:
This is a follow-up of https://github.com/numba/numba/pull/8663. I only rebased the PR, removed a part of the code that wasn't covered by tests and added a release note.


esc:
Thank you for patch and review.

PR #9397, opened 2024-01-18
Reorder block traversal order for correct hoisting.

### LINKED ISSUE #9379: njit(parallel = True)  gives wrong output
## njit with parallel = True gives the wrong output.

The function _groupby_apply_nb_parallel_dynamic_size partially replicates a dataframe.groupby().apply(func) using numba for speedup. Full minimal example code available below.

The function _groupby_apply_nb_parallel_dynamic_size returns wrong output when parallel is set to true.
I have also noticed that when parallel = True, the output of _groupby_apply_nb_parallel_dynamic_size (in the minimal example below) returns os.cpu_count() unique values, when it should return 1001 unique values. The function works correctly when setting Parallel = False. I have also managed to get the correct output by doing some modifications, such as: 
`    out_cols_size = 1 # Setting this to 1 statically also seems to solve the issue
`
Even though I don't know why this should fix the issue.

I can't see any data race conditions, thus why I think it is a Numba bug. 

OS: Ubuntu 22.04
Python 3.11.7
Numba version: '0.58.1', installed through pip
Numpy  '1.26.2'
Pandas : '2.1.4' (it's only used to generate the expected array, nothing else). The version doesn't matter

The below code has similar behavior as pandas.DataFrame.groupby

```
import os
import numpy as np
import numba
from numba.typed import List as NumbaList
import pandas as pd


PARALLEL = True # Set it to false so it works with no issues

@njit(cache=False)
def get_max_first_array_of_list_of_arrays(y_x: NumbaList, return_dtype=np.float64):
    """
    This function only uses the first element of y_x as a minimal example
    """
    y = y_x[0]
    ans = np.array([y.max()], dtype=return_dtype)
    return ans


@njit(parallel=PARALLEL, cache=False)
def _groupby_apply_nb_parallel_dynamic_size(
    start_indices: np.array,
    end_indices: np.array,
    # dataargs is a list of numpy 2d arrays
    dataargs: numba.typed.List,
    func: callable,
    funcargs,
    return_dtype=np.float64,
):
    """
    start_indices : 1d np.array with the start indices of each group
    end_indices : 1d np.array with the end indices of each group
    dataargs: List of np 2d arrays. These are arguments of func
    func: jitted function
    funcargs: other arguments to func.
    return_dtype : output array dtype
    
    """

    ngroups = len(start_indices)
    if ngroups == 0:
        raise Exception("empty number of grouped, (start_indices)")

    # First call to determinize the size of the output. Assumes the first group will have output size
    # equal to the output size of all other groups
    res = func(
        [data[start_indices[0] : end_indices[0], ...] for data in dataargs], *funcargs
    )
    out_cols_size = res.shape[0] # Setting this to 1 statically also seems to solve the issue
    result = np.empty(shape=(ngroups, out_cols_size), dtype=return_dtype)

    for k in prange(ngroups):
        start_idx = start_indices[k]
        end_idx = end_indices[k]
        res = func([data[start_idx:end_idx, ...] for data in dataargs], *funcargs)
        result[k, :] = res

    return result


if __name__ == "__main__":

    np.random.seed(42)
    size = 100_000
    qty_intervals = 1_000
    save_dir = os.path.join("random_experiments", "parallel_py")

    samples = np.sort(np.random.choice(range(size), qty_intervals, replace=False))
    start_indices = np.append(0, samples)
    end_indices = np.append(samples, size)

    x = np.zeros(shape = (size)) # This could be any np.array
    y = np.exp(np.cumsum(np.random.normal(0, 1e-2, size)))

    os.makedirs(save_dir, exist_ok=True)
    for i in range(10):
        max_vals = _groupby_apply_nb_parallel_dynamic_size(
            start_indices = start_indices,
            end_indices = end_indices,
            dataargs  = NumbaList([y, x]),
            func = get_max_first_array_of_list_of_arrays,
            funcargs=(),
            return_dtype=np.float64,
        )
        save_path = os.path.join(save_dir, f'output_{i}_.npy')
        np.save(save_path, max_vals)

    groups = np.zeros(size, dtype=np.int64)
    groups[start_indices] = 1
    groups = np.cumsum(groups)
    max_expected = pd.DataFrame(y).groupby(by = groups).max().values

    for i in range(10):
        save_path = os.path.join(save_dir, f'output_{i}_.npy')
        max_vals = np.load(save_path)
        equal =  np.isclose(max_expected, max_vals, equal_nan = True, rtol = 1e-6).all()
        if not equal:
            print(f"ERROR - parallel output {i} is wrong")

    print(f"Number of unique values : {len(np.unique(max_vals))}")
    print(f"Number of unique expected : {len(np.unique(max_expected))}")
```

DrTodd13:
Resolves #9379


DrTodd13:
The prior code was actually okay but requires the blocks to be processed in the right order.  The critical thing is that if there is a getattr on an object and then a call of that getattr, the getattr has to precede the call for the code that detects multiple writes to an object to detect this scenario as a second write.  My attempt at copying the offending code into a Numba test didn't work because the block order there was one that didn't manifest the problem.  Any test you could write now that used normal code I suppose you'd have no guarantee that the block order would continue to manifest the problem with any change later.

DrTodd13:
@sklam Do you think these test failures are related to the PR?  Seems unrelated to me.

netodw:
@sklam , would you be able to review this PR? The issue is still there.

thomasfrederikhoeck:
might be related to https://github.com/numba/numba/issues/9582 but I'm not sure to be honest.

sklam:
Confirmed that this fixes #9379 but it has also been fixed by #9532. 
This doesn't fix #9586.

DrTodd13:
> Confirmed that this fixes #9379 but it has also been fixed by #9532. This doesn't fix #9586.

So, 9532 fixes 9379 and 9586?  9397 only fixes 9379?

sklam:
> > Confirmed that this fixes #9379 but it has also been fixed by #9532. This doesn't fix #9586.
> 
> So, 9532 fixes 9379 and 9586? 9397 only fixes 9379?

#9532 fixes #9379
#9532 does NOT fix #9586.
#9586 as a PR is needed for https://github.com/numba/numba/issues/9581, https://github.com/numba/numba/issues/9490.

I am trying to understand what remaining cases are fixed by this PR uniquely. The original problem in https://github.com/numba/numba/issues/9379 seems to come from `build_tuple` hiding dependency of mutable object in code like:

```python
func([<listcomp>], *args)
```

where the `build_tuple` comes from handling of `*args`, which is translated to `func( ( [<listcomp>] ,) + args )`. I tested this by changing the original reproducer from

```python
        res = func([data[start_idx:end_idx, ...] for data in dataargs], *funcargs) 
```
to
```python
        res = func([data[start_idx:end_idx, ...] for data in dataargs]) 
```

and confirmed that the bug goes away.

I am unable to find a case that depends on the block traversal order. I believe such a case may appear from Numba internal changes; e.g. changes to inline-closure, changes to parfors details. 

I think this patch will prevent non-deterministic bug that may be introduced from randomized block insertion from other passes or changes to internal details. Since it seems unlikely for users to create such a case, and I am slightly concerned that the reordering may expose some unforeseen bugs, I suggest we merge it for 0.61 milestone.

sklam (dismissed):

sklam:
@DrTodd13, can you merge main to fix the merge conflict?

DrTodd13:
> @DrTodd13, can you merge main to fix the merge conflict?

Done.  @sklam

sklam (dismissed):

sklam (approved):

PR #9401, opened 2024-01-23
Add type annotations for @intrinsic

aneeshdurg:
Previously using the `intrinsic` decorator would cause typing issues when invoking the decorated function, e.g.:
```python
@intrinsic()
def myIntrinsic(typing_context, a, b):
   ...
   
myIntrinsic(a, b) # type checker complains that there's a missing parameter
```

This can be solved by annotating `intrinsic` with types that indicate that the decorated function does not take the typing context parameter.


aneeshdurg:
the failure in the docs build seems to be because python version before 3.10 did not have `typing.Concatenate`. This can by resolved by either taking `typing-extensions` as a dependency, or possibly moving the annotations into a stub file and making it conditional on the python version. Will try to test that out soon.

aneeshdurg:
I was able to resolve the CI issues by moving the annotations to a `.pyi` file. This allows importers of `numba.core.extending` to gain type information about `intrinsic`, but doesn't actually do any type enforcement within `extending.py`.

aneeshdurg:
@stuartarchibald now that this is passing tests can the review process proceed?

stuartarchibald:
> @stuartarchibald now that this is passing tests can the review process proceed?

@aneeshdurg yes, I've now queued it for review. Thanks!

aneeshdurg:
@stuartarchibald any updates on whether this can be reviewed?

stuartarchibald:
@aneeshdurg sure, I'll mention it at the maintainer meeting next week. Also, any chance you could resolve the branch conflicts please? Thanks.

sklam (approved):
Thanks for the contribution

aneeshdurg:
@sklam it seems like CI requires approval to run?

PR #9402, opened 2024-01-23
Doc updates for 0.59 final

sklam:
as titled.

### DIFF
--- a/docs/source/glossary.rst
+++ b/docs/source/glossary.rst
@@ -55,7 +55,7 @@ Glossary
       can be :term:`inferred <type inference>`.
 
       .. note:: Prior to Numba 0.59, ``@jit`` did not set ``nopython=True`` by
-                default and allowed automatically fall back to
+                default and allowed automatic fall back to
                 :term:`object mode`.
 
    Numba IR
--- a/docs/source/glossary.rst
+++ b/docs/source/glossary.rst
@@ -52,9 +52,11 @@ Glossary
       A Numba compilation mode that generates code that does not access the
       Python C API.  This compilation mode produces the highest performance
       code, but requires that the native types of all values in the function
-      can be :term:`inferred <type inference>`.  Unless otherwise instructed,
-      the ``@jit`` decorator will automatically fall back to :term:`object
-      mode` if nopython mode cannot be used.
+      can be :term:`inferred <type inference>`.
+
+      .. note:: Prior to Numba 0.59, ``@jit`` does not set ``nopython=True`` by
+                default and allows automatically fall back to
+                :term:`object mode`.
 
    Numba IR
    Numba intermediate representation
--- a/docs/source/glossary.rst
+++ b/docs/source/glossary.rst
@@ -54,8 +54,8 @@ Glossary
       code, but requires that the native types of all values in the function
       can be :term:`inferred <type inference>`.
 
-      .. note:: Prior to Numba 0.59, ``@jit`` does not set ``nopython=True`` by
-                default and allows automatically fall back to
+      .. note:: Prior to Numba 0.59, ``@jit`` did not set ``nopython=True`` by
+                default and allowed automatically fall back to
                 :term:`object mode`.
 
    Numba IR
--- a/docs/source/release/0.59.0-notes.rst
+++ b/docs/source/release/0.59.0-notes.rst
@@ -511,6 +511,7 @@ Authors:
 ~~~~~~~~
 
 * `apmasell <https://github.com/apmasell>`_
+* `dlee992 <https://github.com/dlee992>`_
 * `DrTodd13 <https://github.com/DrTodd13>`_
 * `esc <https://github.com/esc>`_
 * `gmarkall <https://github.com/gmarkall>`_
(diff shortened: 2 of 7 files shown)


sklam:
The Windows py312-np126 failure seems to be related to CPU changing from [`skylake-avx512`](https://dev.azure.com/numba/numba/_build/results?buildId=16576&view=logs&j=b9870dc3-ee10-5755-1295-228c227d996f&t=e6914cf2-5ab4-5411-f407-f7fa8bc1887e&l=21) to [`broadwell`](https://dev.azure.com/numba/numba/_build/results?buildId=16596&view=logs&j=54c7a26e-1406-55be-c666-794fefe6196f&t=9c5a4e35-ff74-5831-a047-9cbbd46fc80e&l=21)

sklam:
I also wonder if there are changes to the `unittest.skip`. The failing tests are all only skipping in the subprocess but not in the main process (these are all subprocess tests).

docs/source/release/0.59.0-notes.rst
```diff
@@ -456,7 +456,7 @@ Pull-Requests:
 * PR `#9149 <https://github.com/numba/numba/pull/9149>`_: update release checklist post 0.58.0rc1 (`esc <https://github.com/esc>`_)
 * PR `#9154 <https://github.com/numba/numba/pull/9154>`_: Add support for np.unwrap() (`KrisMinchev <https://github.com/KrisMinchev>`_)
 * PR `#9155 <https://github.com/numba/numba/pull/9155>`_: Remove unused test.cmd (`sklam <https://github.com/sklam>`_)
-* PR `#9168 <https://github.com/numba/numba/pull/9168>`_: fix the `get_template_info` method in `overload_method` template (`sklam <https://github.com/sklam>`_)
+* PR `#9168 <https://github.com/numba/numba/pull/9168>`_: fix the `get_template_info` method in `overload_method` template (`dlee992 <https://github.com/dlee992>`_ `sklam <https://github.com/sklam>`_)
```
stuartarchibald, line 459: The list of all the authors at the end of the release notes for 0.59.0 needs `dlee992` adding too now this entry is fixed?

docs/source/user/troubleshoot.rst
```diff
@@ -210,158 +210,16 @@ used to instruct the typing mechanism::
...
-    def f(a, b):
-    ^
-
-    3.0
+:term:`Object mode` provides little to no speedup compared to regular Python 
+interpretation, and its main point is to allow an internal optimization known as
+:term:`loop-lifting`: this optimization will allow to compile inner
+loops in :term:`nopython mode` regardless of what code surrounds those
+inner loops. The inner loops can still fallback to :term:`object mode` if it 
+uses types or operations that :term:`nopython mode` does not support.
```
stuartarchibald, line 222:
```suggestion
:term:`object mode` provides little to no speedup compared to regular Python 
interpretation, its main point is to allow an internal optimization known as
:term:`loop-lifting`: . This optimization will allow compilation of inner
loops in :term:`nopython mode` regardless of what code surrounds those
inner loops. The compilation of inner loops can still fallback to
:term:`object mode` if they use types or operations that
:term:`nopython mode` does not support.
```

stuartarchibald (commented):
Thanks for the patch. Couple of minor suggestions to look at. I also note that the `performance-tips.rst` doc probably  needs an update here:
https://github.com/numba/numba/blob/fed996d3177b64fabf62a526fd329851819dc2da/docs/source/user/performance-tips.rst?plain=1#L21-L33
Perhaps something like:
```
No Python mode
--------------
The default mode in which Numba's ``@jit`` decorator operates is
:term:`nopython mode`. This mode is most restrictive about what can be compiled,
but results in faster executable code.

.. note::
    Historically (prior to 0.59.0) the default compilation mode was a fall-back
    mode whereby the compiler would try to compile in :term:`nopython mode` and
    if it failed it would fall-back to :term:`object mode`. It is likely that
    you'll see ``@jit(nopython=True)``, or its alias ``@njit``, in use in
    code/documentation as this was the recommended best practice method to force
    use of :term:`nopython mode`. Since Numba 0.59.0 this is no long necessary
    as :term:`nopython mode` is the default mode for ``@jit``.
```
could replace it?

docs/source/release/0.59.0-notes.rst
sklam: good catch

docs/source/user/performance-tips.rst
```diff
@@ -66,6 +70,18 @@ the decorator the vectorized function is a couple of orders of magnitude faster.
...
 
+A Case for Object mode: LoopLifting
+-----------------------------------
+
+Some functions may be incompatible with the restrictive :term:`nopython mode` 
+but contain compatible loops. You can enable these functions to attempt nopython 
+mode on their loops by setting `@jit(forceobj=True)`. The incompatible code 
+segments will run in object mode.
+
+While using looplifting in object mode can provide some performance increase,
```
stuartarchibald, line 81:
```suggestion
Whilst using looplifting in object mode can provide some performance increase, 
```

stuartarchibald (commented):
Thanks for the update, a few minor typos/rendering things to look at else looks good. The rendered docs look largely good too (apart from the noted items). Thanks again!

docs/source/glossary.rst
```diff
@@ -52,9 +52,11 @@ Glossary
       A Numba compilation mode that generates code that does not access the
       Python C API.  This compilation mode produces the highest performance
       code, but requires that the native types of all values in the function
-      can be :term:`inferred <type inference>`.  Unless otherwise instructed,
-      the ``@jit`` decorator will automatically fall back to :term:`object
-      mode` if nopython mode cannot be used.
+      can be :term:`inferred <type inference>`.
+
+      .. note:: Prior to Numba 0.59, ``@jit`` did not set ``nopython=True`` by
+                default and allowed automatically fall back to
```
stuartarchibald, line 58:
```suggestion
                default and allowed automatic fall back to
```
apologies, I missed this one in the last review.

PR #9403, opened 2024-01-24
Fix test isolation for stateful configurations in the testsuite

sklam:
This fixes a SIGILL detected in https://github.com/conda-forge/numba-feedstock/pull/131 by adding needed isolation to test that changes the target machine.

Minimal reproducer is: 
```bash
python runtests.py numba.tests.test_vectorization.TestVectorization.test_nditer_loop numba.tests.test_withlifting.TestLiftObj.test_case02_mutate_array_ahead_of_ctx
```

### DIFF
--- a/numba/tests/test_caching.py
+++ b/numba/tests/test_caching.py
@@ -266,9 +266,11 @@ class DispatcherCacheUsecasesTest(BaseCacheTest):
             mod.self_test()
             """ % dict(tempdir=self.tempdir, modname=self.modname)
 
+        subp_env = os.environ.copy()
+        subp_env.update(envvars)
         popen = subprocess.Popen([sys.executable, "-c", code],
                                  stdout=subprocess.PIPE, stderr=subprocess.PIPE,
-                                 env=envvars)
+                                 env={**os.environ, **envvars})
         out, err = popen.communicate()
         if popen.returncode != 0:
             raise AssertionError(
--- a/numba/tests/test_caching.py
+++ b/numba/tests/test_caching.py
@@ -270,7 +270,7 @@ class DispatcherCacheUsecasesTest(BaseCacheTest):
         subp_env.update(envvars)
         popen = subprocess.Popen([sys.executable, "-c", code],
                                  stdout=subprocess.PIPE, stderr=subprocess.PIPE,
-                                 env={**os.environ, **envvars})
+                                 env=subp_env)
         out, err = popen.communicate()
         if popen.returncode != 0:
             raise AssertionError(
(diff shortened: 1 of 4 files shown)


sklam:
I have confirmed that it passes with conda-forge `build-locally.py` for a previously failing config.

stuartarchibald:
@sklam suggest changing the PR title to something like "Fix test isolation for stateful configurations" or something like that?

numba/tests/test_caching.py
```diff
@@ -267,8 +266,11 @@ def run_in_separate_process(self):
             mod.self_test()
             """ % dict(tempdir=self.tempdir, modname=self.modname)
 
+        subp_env = os.environ.copy()
+        subp_env.update(envvars)
         popen = subprocess.Popen([sys.executable, "-c", code],
-                                 stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+                                 stdout=subprocess.PIPE, stderr=subprocess.PIPE,
+                                 env={**os.environ, **envvars})
```
stuartarchibald, line 273:
```suggestion
                                 env=subp_env)
```
It seems like `subp_env` isn't used, should it go here? Or be deleted?

numba/tests/test_parfors.py
stuartarchibald, line 1: Note: l4604 of this file has `overrides.append(override_env_config(k, v))` which, by virtue of its use, is also creating skylake ISA specific machine code, however, all the tests in this test class are in subprocesses so it is "safe".

stuartarchibald (commented):
Thanks for the patch @sklam. I checked the test suite for uses of `override_env_config` where an overridden env var is likely to impact machine code generation, it seems like this patch isolates all remaining cases which weren't already covered. As noted OOB there's still a potential issue with the use of the `fast` flag, but this is better resolved through fixing #7168. There's one minor comment to resolve else looks good. Thanks again!

numba/tests/test_caching.py
sklam: oops, i wanted to switch to `subp_env` because i have to check which dictionary is overriding which in `{**os.environ, **envvars}` spelling.

numba/tests/test_parfors.py
sklam: Actually, it needed the `run_in_subprocess` for all test methods or else it fail: https://dev.azure.com/numba/numba/_build/results?buildId=16609&view=logs&j=7d262216-4420-5164-53df-7079f36c65da&t=9212e158-8292-5264-48bf-03dbdfc6e73f&l=712

stuartarchibald (approved):
Thanks for the update. This is approved conditional on CI passing once `main` is stable again.

stuartarchibald (approved):
Re-approve following merge of `main` into this branch.

PR #9404, opened 2024-01-26
Fix skipped test stderr change for Python 3.12.1.

stuartarchibald:
Python 3.12.1 fixed an issue with the exit code `unittest` reported for skipped tests and also changed the string sent to stderr. Numba was using this string as part of its testing and so was susceptible to this change, this patch fixes Numba's tests.

### DIFF
--- a/numba/tests/support.py
+++ b/numba/tests/support.py
@@ -588,7 +588,12 @@ class TestCase(unittest.TestCase):
         streams = (f'\ncaptured stdout: {status.stdout}\n'
                    f'captured stderr: {status.stderr}')
         self.assertEqual(status.returncode, 0, streams)
-        self.assertIn('OK', status.stderr)
+        # Python 3.12.1 report
+        no_tests_ran = "NO TESTS RAN"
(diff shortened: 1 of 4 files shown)


sklam (commented):
The new regex seems to be matching a specific text pattern that maybe be guaranteed to be stable but it's sufficient for the current 0.59 release.

Maybe, in the long run, we need to not rely on the stdout/stderr text matching.

stuartarchibald:
Looks like the `test_svml` tests are failing, `numba.tests.support` probably needs adjusting similarly to the rest of the changes in this patch. Will take a look.

stuartarchibald:
> The new regex seems to be matching a specific text pattern that maybe be guaranteed to be stable but it's sufficient for the current 0.59 release.
> 
> Maybe, in the long run, we need to not rely on the stdout/stderr text matching.

I agree, post 0.59.0 it would be good to see if just using the exit code of the subprocess is sufficient for Numba's needs.

sklam (approved):

PR #9405, opened 2024-01-26
Update for 0.59.0 final

sklam:
All included patches only affect documentation and the testsuite; thus do not require a new RC.

PR #9411, opened 2024-01-31
Doc updates for 0.59.0 final.

stuartarchibald:
Updates the version support table with the release date, also the release notes with the release date and the patches applied to fix issues found in 0.59.0rc1.


stuartarchibald:
The build is failing because the `release0.59` branch this is targeting does not yet contain #9405 (the fixes since the 0.59.0rc1 was released).

sklam (approved):

Issue #9414, opened 2024-01-31
Inconsistent support of log2


shaunc:
## Reporting a bug

The cpu platform supports `np.log2` but not `math.log2`. The gpu platform supports `math.log2` but not `np.log2`. This makes it impossible to maintain library routines using log2 that are called on both platforms.

- [x] I have tried using the latest released version of Numba (most recent is
  visible in the release notes
  (https://numba.readthedocs.io/en/stable/release-notes-overview.html).
- [x] I have included a self contained code sample to reproduce the problem.
  i.e. it's possible to run as 'python bug.py'.

```py
import math
import sys

import numba as nb  # type: ignore
import numpy as np
from numba import cuda  # type: ignore
from numpy import typing as npt  # type: ignore


@cuda.jit
def log2_math_kernel(x: npt.NDArray[np.float64]) -> None:
    if cuda.grid(1) == 0:
        x[0] = math.log2(x[0])


@cuda.jit
def log2_np_kernel(x: npt.NDArray[np.float64]) -> None:
    if cuda.grid(1) == 0:
        x[0] = np.log2(x[0])


@nb.njit
def log2_math_cpu(x: float) -> float:
    return math.log2(x)


@nb.njit
def log2_np_cpu(x: float) -> float:
    return np.log2(x)


x = np.ones(1, dtype=np.float64)
if __name__ == "__main__":
    if sys.argv[1] == "gm":
        log2_math_kernel[1, 1](x)
    elif sys.argv[1] == "gn":
        log2_np_kernel[1, 1](x)
    elif sys.argv[1] == "cm":
        log2_math_cpu(x[0])
    elif sys.argv[1] == "cn":
        log2_np_cpu(x[0])
```
When running, options "gm" and "cn" run, "gn" and "cm" do not compile.
I am using 0.58.1 on python3.11.3, ubuntu focal.

guilhermeleobas:
I've submitted https://github.com/numba/numba/pull/9416 which partially fixes this issue for the CPU case.

shaunc:
Thank you!

guilhermeleobas:
CUDA support for np.log2 is here:
https://github.com/numba/numba/pull/9417

PR #9416, opened 2024-02-02
Add math.log2 support

guilhermeleobas:
Partially fixes https://github.com/numba/numba/issues/9414

### DIFF
--- a/docs/upcoming_changes/9416.new_feature.rst
+++ b/docs/upcoming_changes/9416.new_feature.rst
@@ -1,0 +1,4 @@
+Support math.log2
+=================
+
+Support for ``math.log2`` is added.
--- a/docs/upcoming_changes/9416.new_feature.rst
+++ b/docs/upcoming_changes/9416.new_feature.rst
@@ -1,4 +1,4 @@
 Add support for ``math.log2``.
-==============================
+------------------------------
 
 Support for ``math.log2`` is added.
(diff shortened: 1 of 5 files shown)


docs/upcoming_changes/9416.new_feature.rst
```diff
@@ -0,0 +1,4 @@
+Support math.log2
+=================
```
stuartarchibald, line 2:
```suggestion
Add support for ``math.log2``.
==============================
```

numba/cpython/mathimpl.py
```diff
@@ -176,6 +176,7 @@ def float_impl(context, builder, sig, args):
 exp_impl = unary_math_intr(math.exp, 'llvm.exp')
 log_impl = unary_math_intr(math.log, 'llvm.log')
 log10_impl = unary_math_intr(math.log10, 'llvm.log10')
+log2_impl = unary_math_intr(math.log2, 'llvm.log2')
```
stuartarchibald, line 179: Technically this implementation should raise `ValueError('math domain error')` for input values that are zero or negative, however, what is present is no different to e.g. `log10` above which would suffer from the same issue.

stuartarchibald (commented):
Thanks for the patch. Please could you add that this is supported to the docs here:
https://github.com/numba/numba/blob/3edb458c69daeaaffbfc938145c367eda1cf043f/docs/source/reference/pysupported.rst?plain=1#L1076
one suggestion RE release note else looks good.

numba/cpython/mathimpl.py
guilhermeleobas: I can tackle this in a follow-up PR if needed.

stuartarchibald (approved):

Issue #9400, opened 2024-01-21
Some array methods much slower with non-C-order arrays


bmerry:
## Reporting a bug

- [x] I have tried using the latest released version of Numba (most recent is
  visible in the release notes
  (https://numba.readthedocs.io/en/stable/release-notes-overview.html).
- [x] I have included a self contained code sample to reproduce the problem.
  i.e. it's possible to run as 'python bug.py'.

While #7622 improved performance of ufuncs on non-C-order arrays, there are some array functions that operate per-element but are not ufuncs, and these could be improved to iterate according to the memory order. Two examples I came across while poking around the numba code are `np.clip` and `ndarray.astype`. Here are some timings from numba 0.59rc1 (code appears below):

```
clip, 1M                  np:        278 µs	numba:         97 µs
clip, 1K×1K, C            np:        271 µs	numba:        601 µs
clip, 1K×1K, F            np:        277 µs	numba:       1180 µs
clip, 500×500, A          np:         96 µs	numba:        253 µs
astype, 1M                np:        193 µs	numba:        176 µs
astype, 1K×1K, C          np:        205 µs	numba:        180 µs
astype, 1K×1K, F          np:        195 µs	numba:       1642 µs
astype, 500×500, A        np:         44 µs	numba:        123 µs
```
The "500×500, A" cases use a 500×500 slice from a 1000×1000 C-contiguous array: that means they're not C-contiguous, but they are row-major (with padding) and so C-order iteration should still be cache-friendly and it is not obvious why these cases perform significantly worse than the plain numpy version.

These also seem to exhibit different behaviour to each other: `clip` performance is much worse on a C-contiguous 2D (1K×1K) view compared to the original 1D (1M-element) array, while for `astype` these cases perform similarly.

```python
#!/usr/bin/env python3

import time

import numpy as np
from numba import njit

@njit
def clip(a):
    return np.clip(a, 0, 5)


@njit
def astype(a):
    return a.astype(np.uint64)


def timeit(passes, func, *args, **kwargs):
    func(*args, **kwargs)  # Warmup, compile JIT code
    start = time.perf_counter_ns()
    for i in range(passes):
        func(*args, **kwargs)
    stop = time.perf_counter_ns()
    elapsed = (stop - start) / passes
    return elapsed


def timeit_both(name, passes, func, *args, **kwargs):
    numba_time = timeit(passes, func, *args, **kwargs)
    np_time = timeit(passes, func.py_func, *args, **kwargs)
[...]

PR #9422, opened 2024-02-04
Implement ndarray.byteswap

bmerry:
It's implemented for numbers (including complex) and boolean.

On a 1D array it's about 7x faster than numpy. For higher dimensions it is slower, and it is much slower for Fortran-order arrays, for the same reasons as in #9400.

This is my first PR for number and I'm pretty new to LLVM as well, so I've got a few questions:

- I've kept the code smaller by unconditionally bit-casting to and from a suitably-sized integer type. If it's already an integer type (probably the common case) that's unnecessary LLVM IR instructions, but I'm assuming that's harmless.
- In numpy, `byteswap` on a scalar with `inplace=True` raises an exception, and I've mirrored that behaviour. Since numba is statically typed, an alternative is to simply not support that useless parameter, which would move the error to compile time (generally a good thing) but be less compatible with numpy. Any thoughts?
- Is there a better way to handle the array iteration? I looked into wrapping the `@intrinsic` scalar function into a ufunc and calling that (which uses a sensible iteration order for Fortran-order arrays), but ran into some cyclic import problems.

### DIFF
--- a/numba/cpython/numbers.py
+++ b/numba/cpython/numbers.py
@@ -13,7 +13,7 @@ from numba.core.imputils import (lower_builtin, lower_getattr,
                                     impl_ret_untracked)
 from numba.core import typing, types, utils, errors, cgutils, optional
 from numba.core.extending import intrinsic, overload_method
-from numba.cpython.unsafe.numbers import viewer
+from numba.cpython.unsafe.numbers import viewer, byteswap
 
 def _int_arith_flags(rettype):
     """
@@ -1202,6 +1202,17 @@ def number_item_impl(context, builder, sig, args):
     The no-op .item() method on booleans and numbers.
     """
     return args[0]
+
+
+@overload_method(types.Boolean, "byteswap")
+@overload_method(types.Number, "byteswap")
+def number_byteswap(a, inplace=False):
+    def number_byteswap_impl(a, inplace=False):
+        if inplace:
+            raise ValueError("cannot byteswap a scalar in-place")
+        return byteswap(a)
+
+    return number_byteswap_impl
 
 
 #------------------------------------------------------------------------------
--- a/numba/np/arrayobj.py
+++ b/numba/np/arrayobj.py
@@ -2739,6 +2739,16 @@ def array_view(context, builder, sig, args):
     return impl_ret_borrowed(context, builder, sig.return_type, res)
 
 
+@overload_method(types.Array, 'byteswap')
+def array_byteswap(arr, inplace=False):
+    def impl(arr, inplace=False):
+        out = arr if inplace else np.empty_like(arr)
+        for index in np.ndindex(arr.shape):
+            out[index] = arr[index].byteswap()
+        return out
+    return impl
+
+
 # ------------------------------------------------------------------------------
 # Array attributes
 
--- a/numba/tests/test_numbers.py
+++ b/numba/tests/test_numbers.py
@@ -88,3 +88,37 @@ class TestViewIntFloat(TestCase):
     def test_exceptions64(self):
         for pair in ((np.int32, np.int64), (np.int64, np.int32)):
             self.do_testing_exceptions(pair)
+
+
+@njit
+def swap(x):
+    return x.byteswap()
+
+
+class TestByteswap(TestCase):
+    """Test the 'byteswap' method on NumPy scalars."""
+
+    def do_test(self, x):
+        self.assertEqual(swap(x), x.byteswap())
+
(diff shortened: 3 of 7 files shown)


bmerry:
Also, is there anything that should be added/tested for CUDA? I've never tried out the CUDA support in numba.

stuartarchibald:
Thanks for submitting the PR @bmerry, it's been queued for review. RE: CUDA, there's nothing to be added for now, there is much less NumPy support on the CUDA target than on the CPU target, if the need arises it can go into a separate PR. Thanks again!

bmerry:
Is there anything else needed to get this reviewed, or is it just a case of too many PRs, not enough developers?

numba/tests/test_numbers.py
```diff
@@ -88,3 +88,37 @@ def test_exceptions32(self):
...
+        self.do_test(np.float32(123.4))
+        self.do_test(np.float64(-123.45))
+
+    def test_bool(self):
+        self.do_test(np.bool_(False))
+        self.do_test(np.bool_(True))
+
+    def test_complex(self):
+        self.do_test(np.complex64(123.4 - 567.8j))
+        self.do_test(np.complex128(123.4 - 567.8j))
```
kc611, line 124: This test fails with:

```
do_test(np.array(["abc"]))
```

Because there are no safeguards for types other than numerical types.

numba/np/arrayobj.py
```diff
@@ -2739,6 +2739,16 @@ def array_view(context, builder, sig, args):
     return impl_ret_borrowed(context, builder, sig.return_type, res)
 
 
+@overload_method(types.Array, 'byteswap')
+def array_byteswap(arr, inplace=False):
+    def impl(arr, inplace=False):
```
kc611, line 2887: A safeguard will need to be implemented here for cases when array elements are of types other than `types.Boolean` and `types.Integer`.

docs/upcoming_changes/9422.np_support.rst
```diff
@@ -0,0 +1,5 @@
+Added support for byteswap.
+===========================
+
+Support is added for the numpy ``byteswap`` method on numbers, booleans, and
```
kc611:
```suggestion
Added support for the NumPy ``byteswap`` method on Numbers, Booleans and
```

docs/upcoming_changes/9422.np_support.rst
```diff
@@ -0,0 +1,5 @@
+Added support for byteswap.
```
kc611:
```suggestion
Added support for ``.byteswap()`` method for numerical scalars and arrays.
```

numba/cpython/numbers.py
```diff
@@ -1204,6 +1204,17 @@ def number_item_impl(context, builder, sig, args):
     return args[0]
 
 
+@overload_method(types.Boolean, "byteswap")
+@overload_method(types.Number, "byteswap")
+def number_byteswap(a, inplace=False):
+    def number_byteswap_impl(a, inplace=False):
+        if inplace:
+            raise ValueError("cannot byteswap a scalar in-place")
```
kc611, line 1214:
```suggestion
            raise ValueError("Cannot byteswap a scalar in-place.")
```

bmerry: The original string is what numpy raises. To confirm, does numba prefer good grammar in exception messages over matching numpy behaviour?

bmerry:
> I took a look at this PR and the implementation itself looks good. A few nitpicks below:

Thanks @kc611. I'll try to take a look this week.

numba/tests/test_numbers.py
bmerry: I'll fix that and test it in test_array_methods.py.

FWIW the scalar case (i.e. `self.do_test(np.str_("abc"))`) raises the following:
```
Unknown attribute 'byteswap' of type unicode_type

File "numba/tests/test_numbers.py", line 95:
def swap(x):
    return x.byteswap()
    ^

During: typing of get attribute at /home/bmerry/src/numba/numba/tests/test_numbers.py (95)

File "numba/tests/test_numbers.py", line 95:
def swap(x):
    return x.byteswap()
    ^

During: Pass nopython_type_inference
```

bmerry:
> I had to merge main into my branch so that I could build against current versions of numpy (fortunately no merge conflicts). Will that be okay to push or should I try to remove that merge from my history for easier reviewing?

For now I've pushed so that the changes are available and won't be lost if my laptop dies.

PR #9428, opened 2024-02-08
Optimize `range()` to enable more auto-vectorization

sklam:
It is known that Numba's `range()` implementation is not optimal for LLVM loop optimizers. Even when one manually unroll loop for a constant loop bound, the way `range()` is implemented prohibits the loop-vectorizer to compute the loop bounds; thus failing to auto-vectorize in many cases. 

This patch focuses on adjusting the ind-var computation to avoid computing reminders and instead relies mainly on additions, multiplications and floordiv which are common in low-level address computations.

There is still a reminder computation only if user ask for `length_of_iterator(iter(range()))`, but it is a rare use-case. LLVM is able to optimize-away that reminder (and the storage of `range_iterator_type.count`)

Here's a notebook with a use-case I used to optimize `range()`: https://gist.github.com/sklam/6beddb2041580ceea4f25e0e496f50a0/420bc173da75996a307465f134c23bb3f4bb6562.

### DIFF
--- a/numba/core/datamodel/models.py
+++ b/numba/core/datamodel/models.py
@@ -1133,10 +1133,18 @@ class ZipType(StructModel):
 class RangeIteratorType(StructModel):
     def __init__(self, dmm, fe_type):
         int_type = fe_type.yield_type
-        members = [('iter', types.EphemeralPointer(int_type)),
-                   ('stop', int_type),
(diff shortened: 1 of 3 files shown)


gmarkall:
@sklam should this have been closed, or is it likely to move forward at some point?

Issue #9432, opened 2024-02-10
Bug: (regression) cooperative grid sync w/ cache=True - working in 0.58.1, not working in 0.59.0


shaunc:
## Reporting a bug

Ability to cache kernels using cooperative grid sync was added in https://github.com/numba/numba/pull/8895.

Unfortunately this seems to be broken again in 0.59.0, though is working in 0.58.1.

Test:
```py
from numba import cuda  # type: ignore


@cuda.jit(cache=True)  # type: ignore
def assign_edges_kernel() -> None:
    """
    Kernel which tests the assign_edges function.
    """
    gg = cuda.cg.this_grid()  # type: ignore
    gg.sync()


if __name__ == "__main__":
    assign_edges_kernel[1, 1]()  # type: ignore
```

Result in 0.58.1:
```
$ poetry add numba==0.58.1

Updating dependencies
Resolving dependencies... (0.4s)

Package operations: 0 installs, 2 updates, 0 removals

  • Downgrading llvmlite (0.42.0 -> 0.41.1)
  • Downgrading numba (0.59.0 -> 0.58.1)

Writing lock file
(python3.11) shauncutts@silk:~/.../mlops/ingolstadt$ poetry run python3.11 test.py
/home/shauncutts/.cache/pypoetry/virtualenvs/ingolstadt-EH7L9v-Y-py3.11/lib/python3.11/site-packages/numba/cuda/dispatcher.py:536: NumbaPerformanceWarning: Grid size 1 will likely result in GPU under-utilization due to low occupancy.
  warn(NumbaPerformanceWarning(msg))
  ```
  
  Result in 0.59.0:
  ```
  $ poetry add numba==0.59.0

Updating dependencies
Resolving dependencies... (0.4s)

Package operations: 0 installs, 2 updates, 0 removals

  • Updating llvmlite (0.41.1 -> 0.42.0)
  • Updating numba (0.58.1 -> 0.59.0)

Writing lock file
(python3.11) shauncutts@silk:~/.../mlops/ingolstadt$ poetry run python3.11 test.py
/home/shauncutts/.cache/pypoetry/virtualenvs/ingolstadt-EH7L9v-Y-py3.11/lib/python3.11/site-packages/numba/cuda/dispatcher.py:536: NumbaPerformanceWarning: Grid size 1 will likely result in GPU under-utilization due to low occupancy.
  warn(NumbaPerformanceWarning(msg))
Traceback (most recent call last):
[...]

PR #9444, opened 2024-02-19
[WIP] Empty commit to test on gpuCI

gmarkall:
For debugging / testing #9432, and anything else that happens to be wrong on `main` at present.


gmarkall:
gpuci run tests

PR #9446, opened 2024-02-20
Type system implementation #1 (Alternative): Added initial implementation for a new type system using redundancies.

kc611:
This is an alternative to #9408. 

It does the same changes except those are changes are being done in entirely new files to prevent any collision with old numba system.


numba/core/cpu.py
```diff
@@ -71,35 +71,36 @@ def load_additional_registries(self):
...
-        self.install_registry(printimpl.registry)
-        self.install_registry(randomimpl.registry)
-        self.install_registry(jitclassimpl.class_impl_registry)
-
-        # load 3rd party extensions
-        numba.core.entrypoints.init_all()
-
-        # fix for #8940
-        from numba.np.unsafe import ndarray # noqa F401
+        if numba.core.config.USE_LEGACY_TYPE_SYSTEM:
```
gmarkall, line 74: Note for anyone reading this: From discussions with @stuartarchibald - there is presently nothing imported when using the new type system, so that the new implementations can be built up correctly and imports / setup added here piece-by-piece until everything is implemented for the new type system.

(As opposed to starting with everythign in the new type system and trying to convert its old behaviour to new behaviour, which would be very hard to do thoroughly and correctly)

kc611:
Closing in favour of #9513

Issue #9441, opened 2024-02-17
Regression: searchsorted drastically slower in tight loops


GeorgWa:
It looks to me as if `np.searchsorted` shows drastically decreased performance in the `0.59.0` release compared to `0.58.1`.

The following example should replicate the bug:

```python
import numba as nb
import numpy as np
import time

n_samples = 10000

segment_len = np.random.randint(10, 10000, size=n_samples).astype(np.uint32)
stop_index = np.cumsum(segment_len, dtype=np.uint32)
start_index = np.concatenate(([0], stop_index[:-1]))

upper_limit = 1000.0
lower_limit = 0.0

intensities = []
for i in range(n_samples):
    intensities.append(
        sorted(
            np.random.uniform(lower_limit, upper_limit, size=segment_len[i]).astype(np.float32)
        )
    )

intensities = np.concatenate(intensities)
search_values = np.random.uniform(lower_limit, upper_limit, size=n_samples).astype(np.float32)

@nb.njit(parallel=False, fastmath=True)
def test_np_search(search_values, intensities, start_index, stop_index):
    y = 0
    for i in range(n_samples):
        for search_value in search_values:
            y += np.searchsorted(intensities[start_index[i]:stop_index[i]], search_value, "left")

    return y
```

```python
%%timeit
test_np_search(search_values, intensities, start_index, stop_index)
```

`0.58.1`:  2.9 s ± 4.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
`0.59.0`: 14.6 s ± 14.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

Using a naive implementation results in similar but slower times:
```Python
@nb.njit(parallel=False, fastmath=True)
def search_sorted_left(slice, value):
    left = 0
    right = len(slice)

    while left < right:
        mid = (left + right) // 2
        if slice[mid] < value:
            left = mid + 1
        else:
            right = mid
    return left
```
`0.58.1`: 4.44 s ± 3.23 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
`0.59.0`: 4.43 s ± 5.55 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

<details>
  <summary>Environment</summary>
  
  ```
# Name                    Version                   Build  Channel
appnope                   0.1.4              pyhd8ed1ab_0    conda-forge
asttokens                 2.4.1              pyhd8ed1ab_0    conda-forge
[...]

PR #9445, opened 2024-02-19
Fix #9427 np.searchsorted on datetime64

sklam:
Based on https://github.com/numba/numba/pull/9448.
Fixes #9427.

- Adds datetime64 tests for searchsorted
- Re-enables datetime64 support for searchsorted
- Fixes NAT mishandling

PR #9448, opened 2024-02-20
Fix `np.searchsorted` regression

guilhermeleobas:
Fixes https://github.com/numba/numba/issues/9441

Note for the reviewer: Previous code would allocate an array (`np.array([v])`) when `v` was a scalar. This allocation caused the code to be 4x slower when compared to previous implementation. This fixes this by only allocating an array when `v` is a container (list, tuple, ndarray)

```python
import numba as nb
import numpy as np
import time

n_samples = 10000

segment_len = np.array([10_000] * n_samples)
stop_index = np.cumsum(segment_len, dtype=np.int64)
start_index = np.concatenate(([0], stop_index[:-1]))

upper_limit = 1000.0
lower_limit = 0.0

intensities = []
for i in range(n_samples):
    intensities.append(
        sorted(
            np.random.uniform(lower_limit, upper_limit, size=segment_len[i]).astype(np.int64)
        )
    )

intensities = np.concatenate(intensities)
search_values = np.random.uniform(lower_limit, upper_limit, size=n_samples).astype(np.int64)

@nb.njit(parallel=False, fastmath=True)
def test_np_search(search_values, intensities, start_index, stop_index):
    y = 0
    for i in range(n_samples):
        for search_value in search_values:
            y += np.searchsorted(intensities[start_index[i]:stop_index[i]], search_value, "left")

    return y

test_np_search(search_values, intensities, start_index, stop_index)

start = time.time()
test_np_search(search_values, intensities, start_index, stop_index)
end = time.time()

print(end - start)
```

Numba 0.58: 3.0s
Numba 0.59: 12s
This patch: 3.3s

### DIFF
--- a/docs/upcoming_changes/9448.performance.rst
+++ b/docs/upcoming_changes/9448.performance.rst
@@ -1,0 +1,5 @@
+Improvement to ``np.searchsorted``
+----------------------------------
+
+Fixed a performance regression introduced in Numba 0.59 which made
+``np.searchsorted`` considerably slower.
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -3735,46 +3735,20 @@ def _less_than(a, b):
     return a < b
 
 
-def _searchsorted(func_1, func_2):
+def _searchsorted(cmp):
     # a facsimile of:
     # https://github.com/numpy/numpy/blob/4f84d719657eb455a35fcdf9e75b83eb1f97024a/numpy/core/src/npysort/binsearch.cpp#L61  # noqa: E501
 
-    def impl(a, v):
-        min_idx = 0
-        max_idx = len(a)
-
-        out = np.empty(v.size, np.intp)
-
-        last_key_val = v.flat[0]
-
-        for i in range(v.size):
-            key_val = v.flat[i]
-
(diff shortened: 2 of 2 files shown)


sklam:
NAT test that is failing on my branch (https://github.com/numba/numba/pull/9445) and before https://github.com/numba/numba/pull/9189 is:


```python
import numpy as np

from numba import njit

@njit
def foo(a, v):
    return np.searchsorted(a, v)

a = np.array([
    np.datetime64(1, 'ns'),
    np.datetime64('NAT'),
])

args = (a, a[-1])
# got
r = foo(*args)
print(r)
# expect
r = foo.py_func(*args)
print(r)
```

Outputs:

```
$ python chk_nat.py
0
1
```

My branch has a searchsorted datetime64 test at https://github.com/numba/numba/pull/9445/files#diff-20a798fb06097b4e960ba71473dd4aae54d2f87466fb460bbea151000b6076aaR1135 and it currently filtering out NAT.

sklam (approved):
I confirmed the performance fix. I have additional fixes for datetime64 and NAT handling at https://github.com/numba/numba/pull/9445 which is now based on this PR.

PR #9449, opened 2024-02-20
Remove deprecated CondaEnvironment@1

sklam:

PR #9450, opened 2024-02-20
Fix gpuci versions

gmarkall:
This fixes the current fail in `main` with CUDA 11.5 in gpuCI (which is down to a bad CI setup / configuration), and adds support for the CUDA Toolkit 12.3 supported compute capabilities.

Notes:

- gpuCI now passes with this fix, the Windows fails are due to the issue that will be fixed by #9449 
- I've marked it as "skip release notes" because it's only a fix for CI, plus a change to the CC matrix that makes no functional difference for users of CTK 12.3 at this time - Numba would have correctly inferred the supported range for 12.3 already anyway.


gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

sklam (approved):

PR #9454, opened 2024-02-21
Don't attempt to register overloads that aren't for this target in `BaseContext` and related fixes

gmarkall:
Fixes a cascading plethora of issues discovered as a result of looking into #9432. See individual commit messages for details of fixes - the pattern is generally that each fix exposed a new issue, resolved by a subsequent commit.

### DIFF
--- a/numba/cuda/tests/cudapy/test_overload.py
+++ b/numba/cuda/tests/cudapy/test_overload.py
@@ -1,8 +1,8 @@
 from numba import cuda, njit, types
 from numba.core.errors import TypingError
 from numba.core.extending import overload, overload_attribute
+from numba.core.typing.typeof import typeof
 from numba.cuda.testing import CUDATestCase, skip_on_cudasim, unittest
-from numba.tests.test_extending import mydummy_type, MyDummyType
 import numpy as np
 
 
@@ -297,6 +297,9 @@ class TestOverload(CUDATestCase):
         self.check_overload_cpu(kernel, expected)
 
     def test_overload_attribute_target(self):
+        MyDummy, MyDummyType = self.make_dummy_type()
+        mydummy_type = typeof(MyDummy())
+
         @overload_attribute(MyDummyType, 'cuda_only', target='cuda')
         def ov_dummy_cuda_attr(obj):
             def imp(obj):
@@ -308,7 +311,7 @@ class TestOverload(CUDATestCase):
         # CPU, and that an appropriate typing error is raised
         with self.assertRaisesRegex(TypingError,
                                     "Unknown attribute 'cuda_only'"):
-            @njit(types.void(mydummy_type))
+            @njit(types.int64(mydummy_type))
             def illegal_target_attr_use(x):
                 return x.cuda_only
 
--- a/numba/cuda/tests/cudapy/test_overload.py
+++ b/numba/cuda/tests/cudapy/test_overload.py
@@ -1,7 +1,8 @@
-from numba import cuda, njit
-from numba.core.extending import overload
+from numba import cuda, njit, types
+from numba.core.errors import TypingError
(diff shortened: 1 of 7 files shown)


gmarkall:
gpuci run tests

numba/cuda/tests/cudapy/test_overload.py
```diff
@@ -295,6 +296,29 @@ def kernel(x):
         expected = GENERIC_TARGET_OL_CALLS_TARGET_OL * GENERIC_TARGET_OL
         self.check_overload_cpu(kernel, expected)
 
+    def test_overload_attribute_target(self):
```
gmarkall, line 299: This should also have a test similar in `test_target_extension.py`

gmarkall:
gpuci run tests

gmarkall:
gpuci run tests

numba/cuda/tests/cudapy/test_overload.py
```diff
@@ -1,7 +1,8 @@
-from numba import cuda, njit
-from numba.core.extending import overload
+from numba import cuda, njit, types
+from numba.core.errors import TypingError
+from numba.core.extending import overload, overload_attribute
 from numba.cuda.testing import CUDATestCase, skip_on_cudasim, unittest
-
+from numba.tests.test_extending import mydummy_type, MyDummyType
```
stuartarchibald, line 5: Should the `make_dummy_type()` method on the `numba.tests.support.TestCase` be used to generate some new type so as to avoid cross test contamination?

numba/tests/test_target_extension.py
```diff
@@ -41,6 +42,7 @@
 from numba.core.compiler import CompilerBase, DefaultPassBuilder
 from numba.core.compiler_machinery import FunctionPass, register_pass
 from numba.core.typed_passes import PreLowerStripPhis
+from numba.tests.test_extending import mydummy_type, MyDummyType
```
stuartarchibald, line 45: Same comment as above RE used of `self.make_dummy_type()` within the test.

numba/tests/test_target_extension.py
```diff
@@ -317,6 +323,8 @@ def dispatcher_wrapper(self):
         if "nopython" in self._kwargs:
             topt["nopython"] = True
 
+        topt['target_backend'] = 'dpu'
```
stuartarchibald, line 326: This is going to be removed by #9539.

stuartarchibald (commented):
Thanks for fixing this @gmarkall. As we have debugged, discussed and worked on this logic through many OOB conversations I'm reasonably convinced that this fixes the issues with the use of `target=` in `@overload_method` etc and also a number of additional issues discovered in the debugging process. There's a few minor things to address, otherwise looks good.

numba/cuda/tests/cudapy/test_overload.py
gmarkall: Yeah, I think that would work, and it does look a bit weird to use bits from `numba.tests.test_extending` here.

numba/tests/test_target_extension.py
stuartarchibald: Yes, I am unconvinced it had any effect.

gmarkall: Thanks - I've now removed it.

numba/tests/test_target_extension.py
gmarkall: Changed.

numba/cuda/tests/cudapy/test_overload.py
gmarkall: Now changed.

gmarkall:
gpuci run tests

stuartarchibald (approved):

Issue #9427, opened 2024-02-08
`searchsorted` with dtype `datetime64` stopped working in `njit` mode going from 0.58.1 to 0.59


zansibal:
## Reporting a bug

- [x] I have tried using the latest released version of Numba (most recent is
  visible in the release notes
  (https://numba.readthedocs.io/en/stable/release-notes-overview.html).
- [x] I have included a self contained code sample to reproduce the problem.
  i.e. it's possible to run as 'python bug.py'.

Using `numpy.searchsorted()` with dtype `datetime64` does not work in `njit` mode in version 0.59. This works fine in version 0.58.1.

I guess this is related to the recent changes to the function. #9189 

Reproducible sample:
```
import numpy as np
from numba import njit

@njit
def foo(a, v):
    return np.searchsorted(a, v)

arr = np.asarray(
    ['2022-03-09 06:45:00', '2022-03-09 06:50:00', '2022-03-09 06:55:00'],
    dtype=np.dtype('datetime64[ns]')
)
print(foo(arr, arr[-1]))
```

Output:
```
---------------------------------------------------------------------------
TypingError                               Traceback (most recent call last)
Cell In[37], line 12
      6     return np.searchsorted(a, v)
      8 arr = np.asarray(
      9     ['2022-03-09 06:45:00', '2022-03-09 06:50:00', '2022-03-09 06:55:00'],
     10     dtype=np.dtype('datetime64[ns]')
     11 )
---> 12 print(foo(arr, arr[-1]))

File ~/.cache/pypoetry/virtualenvs/indefatigable-XRADjY79-py3.11/lib/python3.11/site-packages/numba/core/dispatcher.py:468, in _DispatcherBase._compile_for_args(self, *args, **kws)
    464         msg = (f"{str(e).rstrip()} \n\nThis error may have been caused "
    465                f"by the following argument(s):\n{args_str}\n")
    466         e.patch_message(msg)
--> 468     error_rewrite(e, 'typing')
    469 except errors.UnsupportedError as e:
    470     # Something unsupported is present in the user code, add help info
    471     error_rewrite(e, 'unsupported_error')

File ~/.cache/pypoetry/virtualenvs/indefatigable-XRADjY79-py3.11/lib/python3.11/site-packages/numba/core/dispatcher.py:409, in _DispatcherBase._compile_for_args.<locals>.error_rewrite(e, issue_type)
    407     raise e
    408 else:
--> 409     raise e.with_traceback(None)

[...]

PR #9455, opened 2024-02-21
Support datetime types in `isinstance()`

sklam:
Discovered this as part of debugging #9427, `isinstance()` should support datetime types for either/both arguments in `isinstance()`.


guilhermeleobas (approved):
Code looks good. Could you add a test for `np.searchsorted` to make sure https://github.com/numba/numba/issues/9427 was fixed?

PR #9456, opened 2024-02-21
Update release checklist

sklam:
Move the creation of GH release from "post-release" into "release" task

This is a task from 0.59.0 checklist.


esc (approved):
Attention to detail is much appreciated, thank you!

PR #9457, opened 2024-02-22
Remove unused recompilation for the same function and same signature

### LINKED ISSUE #9440: Overloaded function is compiled more than once with the same signature
Now, I think this should be a type template issue. The `OverloadMethodTemplate` doesn't respect the `kwargs` in the related `impl`. So, the dict `get` method also has this issue, if we use them and don't specify `default=a_variable` explicitly. And I also remember some overloaded methods for unicode type also use many `kwargs`.

======== original post =========

Test code:
```python
import os
os.environ["NUMBA_CHROME_TRACE"] = "relower.json"
os.environ["NUMBA_DUMP_OPTIMIZED"] = "1"

from numba import njit
from numba.core import types
from numba.typed import Dict


@njit
def foo():
    d = Dict.empty(key_type=types.int64, value_type=types.int64)
    d[1] = 1
    del d[1]

foo()
```
Then, I checked the JSON file in chrome, I found 2 overloaded functions are compiled twice with the same type signature.
I think this should be a compilation bug, or perhaps it's a feature or limitation in overload implementation?
From the common sense, the same signature of a jitted/overloaded function should be only compiled once, "no matter any situation" in the same process/run. But this case is against this long-term assumption...

1. `native_lowering [typeddict_empty.<locals>.impl]`
2. `native_lowering [impl_pop.<locals>.impl]`
image

If grep the LLVM IR header, 
```
-------------------------------OPTIMIZED DUMP nrt-------------------------------
-------------------OPTIMIZED DUMP impl_new_dict.<locals>.imp--------------------
------------------OPTIMIZED DUMP typeddict_empty.<locals>.impl------------------
--------------------OPTIMIZED DUMP ol_hasattr.<locals>.impl---------------------
-------------------OPTIMIZED DUMP ol_getattr_2.<locals>.impl--------------------
---------------------------OPTIMIZED DUMP _long_impl----------------------------
-------------------------OPTIMIZED DUMP process_return--------------------------
---------------------OPTIMIZED DUMP int_hash.<locals>.impl----------------------
-------------------OPTIMIZED DUMP ol_defer_hash.<locals>.impl-------------------
-------------------OPTIMIZED DUMP hash_overload.<locals>.impl-------------------
-------------------OPTIMIZED DUMP impl_setitem.<locals>.impl--------------------
---------------------OPTIMIZED DUMP impl_pop.<locals>.impl----------------------
---------------------OPTIMIZED DUMP impl_pop.<locals>.impl----------------------
-------------------OPTIMIZED DUMP impl_delitem.<locals>.impl--------------------
------------------OPTIMIZED DUMP typeddict_empty.<locals>.impl------------------
-------------------------------OPTIMIZED DUMP foo-------------------------------
```

For `impl_pop.<locals>.impl`, the first version (coming from the typer pass of `impl_delitem`) is compiled, but unused. The second version (coming from the lowering pass of `impl_delitem`) is actually recompiled and used by its caller.

When we use `overload` or `overload_method` intensively, our compilation time and consumed memory could be increased maybe 2x than it should be.

the cache_key somehow is wrong for this case:
https://github.com/numba/numba/blob/3edb458c69daeaaffbfc938145c367eda1cf043f/numba/core/typing/templates.py#L707
They have different `args`:
1. (DictType(int64, int64), int64)
2. (DictType(int64, int64), int64, omitted(default=None))

the root cause could come from the impl for `pop`: https://github.com/numba/numba/blob/3edb458c69daeaaffbfc938145c367eda1cf043f/numba/typed/dictobject.py#L814

The second arg types should be correct. Then the issue comes from the transition from typer pass of `impl_delitem` to the compilation of `impl_pop`.

I think this issue at least contains 2 problems:

1. `impl_pop` is compiled twice, and the first compilation result is useless;
2. the JSON trace graph doesn't exactly match with the actual execution, in the `args` and `return_type` fields.

even if I set `should_raise` to a const, `False`, it still compiles twice. This should be a fundamental bug. https://github.com/numba/numba/blob/3edb458c69daeaaffbfc938145c367eda1cf043f/numba/typed/dictobject.py#L811

### LINKED ISSUE #9462: overload_methoded function shouldn't set `no_cpython_wrapper` as `False`
During testing #9440 , I found another issue related.

```python
from os import environ

environ["NUMBA_CHROME_TRACE"] = "1.json"
environ["NUMBA_CAPTURED_ERRORS"] = "new_style"
environ["NUMBA_DUMP_OPTIMIZED"] = "1"

from numba import njit
from numba.core import types
from numba.typed import Dict

@njit
def goo(k, j):
    d = Dict.empty(types.int64, value_type=types.int64)
    d[1] = 1
    d[2] = 2
    d.pop(k)
    del d[j]

goo(1, 2)
```
We found impl_pop is compiled 3 times.
Based on #9440, we know 1st and 3rd have the different type signature, so it's recompiled.
More importantly in this issue, I notice 1st and 2nd have the same type signature, but different compilation flags for impl_caches, which results in the recompilation. The only obvious diff in their flags is `no_cpython_wrapper`. For this overloaded function, this flag should be `True`, but somehow and somewhere, it's overrided to `False` for the 1st case.
```
---------------------OPTIMIZED DUMP impl_pop.<locals>.impl----------------------
---------------------OPTIMIZED DUMP impl_pop.<locals>.impl----------------------
---------------------OPTIMIZED DUMP impl_pop.<locals>.impl----------------------
```

dlee992:
Fixes #9440 
Fixes #9462


dlee992:
Numba compilation still has at least 2 issues:

1. sometimes, if the return type of a jitted/overloaded function is Tuple, numba forcily doesn't cache this function into its cache dict...
2. different callers call the same overloaded function with the same type signature, the overloaded function still compiles twice, since they have different compilation flags, e.g., no_cpython_wrapper=True or False, and fields in flags are in different order..

Now, I only solved one issue (different from above 2): if a overloaded function is compiled in caller's typeinfer pass, then I can make sure it won't recompile in caller's native lowering pass...

dlee992:
An example to show why we need this PR:
```python
from os import environ

environ["NUMBA_CHROME_TRACE"] = "3.json"
environ["NUMBA_DUMP_LLVM"] = "1"

from numba import njit
from numba.core import types
from numba.typed import Dict


@njit
def goo(k, j):
    d0 = Dict.empty(types.int64, types.int64)
    d = Dict.empty(types.int64, value_type=types.int64)
    d1 = Dict.empty(key_type=types.int64, value_type=types.int64)
    d[1] = 1
    d[2] = 2
    d[3] = 3
    _ = d.get(k)
    d.pop(k)
    d.pop(j, None)
    del d[3]

@njit
def joo(k):
    d = Dict.empty(types.int64, value_type=types.int64)
    d[1] = 1
    return d.pop(k, None)

print(goo(1, 2))
print(joo(0))
```
without this patch, we compile these modules:
```
---------LLVM DUMP <function descriptor 'impl_new_dict.<locals>.imp$3'>---------
-------LLVM DUMP <function descriptor 'typeddict_empty.<locals>.impl$2'>--------
-------LLVM DUMP <function descriptor 'typeddict_empty.<locals>.impl$4'>--------
-------LLVM DUMP <function descriptor 'typeddict_empty.<locals>.impl$5'>--------
----------LLVM DUMP <function descriptor 'ol_hasattr.<locals>.impl$9'>----------
--------LLVM DUMP <function descriptor 'ol_getattr_2.<locals>.impl$10'>---------
--------------LLVM DUMP <function descriptor 'process_return$13'>---------------
----------------LLVM DUMP <function descriptor '_long_impl$14'>-----------------
----------LLVM DUMP <function descriptor 'int_hash.<locals>.impl$12'>-----------
--------LLVM DUMP <function descriptor 'ol_defer_hash.<locals>.impl$11'>--------
--------LLVM DUMP <function descriptor 'hash_overload.<locals>.impl$7'>---------
---------LLVM DUMP <function descriptor 'impl_setitem.<locals>.impl$6'>---------
----------LLVM DUMP <function descriptor 'impl_get.<locals>.impl$16'>-----------
----------LLVM DUMP <function descriptor 'impl_pop.<locals>.impl$17'>-----------
----------LLVM DUMP <function descriptor 'impl_pop.<locals>.impl$18'>-----------
----------LLVM DUMP <function descriptor 'impl_pop.<locals>.impl$20'>-----------
----------LLVM DUMP <function descriptor 'impl_pop.<locals>.impl$21'>-----------
--------LLVM DUMP <function descriptor 'impl_delitem.<locals>.impl$19'>---------
-------LLVM DUMP <function descriptor 'typeddict_empty.<locals>.impl$22'>-------
----------LLVM DUMP <function descriptor 'impl_get.<locals>.impl$23'>-----------
----------LLVM DUMP <function descriptor 'impl_pop.<locals>.impl$24'>-----------
--------------------LLVM DUMP <function descriptor 'goo$1'>---------------------
--------------------LLVM DUMP <function descriptor 'joo$25'>--------------------
```
with this patch, we compile these modules:
```
---------LLVM DUMP <function descriptor 'impl_new_dict.<locals>.imp$3'>---------
-------LLVM DUMP <function descriptor 'typeddict_empty.<locals>.impl$2'>--------
----------LLVM DUMP <function descriptor 'ol_hasattr.<locals>.impl$7'>----------
---------LLVM DUMP <function descriptor 'ol_getattr_2.<locals>.impl$8'>---------
--------------LLVM DUMP <function descriptor 'process_return$11'>---------------
----------------LLVM DUMP <function descriptor '_long_impl$12'>-----------------
----------LLVM DUMP <function descriptor 'int_hash.<locals>.impl$10'>-----------
--------LLVM DUMP <function descriptor 'ol_defer_hash.<locals>.impl$9'>---------
--------LLVM DUMP <function descriptor 'hash_overload.<locals>.impl$5'>---------
---------LLVM DUMP <function descriptor 'impl_setitem.<locals>.impl$4'>---------
----------LLVM DUMP <function descriptor 'impl_get.<locals>.impl$14'>-----------
----------LLVM DUMP <function descriptor 'impl_pop.<locals>.impl$15'>-----------
----------LLVM DUMP <function descriptor 'impl_pop.<locals>.impl$16'>-----------
--------LLVM DUMP <function descriptor 'impl_delitem.<locals>.impl$17'>---------
--------------------LLVM DUMP <function descriptor 'goo$1'>---------------------
--------------------LLVM DUMP <function descriptor 'joo$18'>--------------------
```
The diff is without it, our main will compile `typeddict_empty` 4 times and `impl_pop` 5 times. with this patch, it compiles `typeddict_empty` 1 time and `impl_pop` twice (for different `default` types, `Omitted(default=None)` and `types.none`)
cc @guilhermeleobas

numba/cpython/randomimpl.py
```diff
@@ -1952,7 +1952,7 @@ def getitem(a, a_i):
         raise TypeError("np.random.choice() first argument should be "
                         "int or array, got %s" % (a,))
 
-    if size in (None, types.none):
+    if size in (None, types.none) or isinstance(size, types.Omitted):
```
guilhermeleobas, line 1955: Not a proper review, but this ought to be `cgutils.is_nonelike(size)`?

dlee992: Yes, I also notice this API to check it, will refactor.

dlee992:
I didn't change this file, but CI reports the flake8 errors in it. Emm...
```
numba/tests/test_unicode.py:2670:27: E226 missing whitespace around arithmetic operator
```

dlee992:
@sklam I'm not sure if I choose the right way to fix them. Maybe the real fix should happen at the start of "funny" things, rather than at the end of them. I choose the latter way in this PR.

dlee992:
interesting failure: 
```python
  File "/home/vsts/work/1/s/numba/tests/test_compiler_flags.py", line 40, in test_fastmath_in_overload
    self.assertEqual(b, "Has fastmath")
AssertionError: 'No fastmath' != 'Has fastmath'
- No fastmath
? ^^
+ Has fastmath
? ^^^
```
It could inspire me how to fix `flags` more decently.

dlee992:
added a naive release doc and fix the failed test case.

dlee992:
this failure (numba.numba (Linux py310_np123)) is a CI infra issue, not code issue.

stuartarchibald:
Re-opening as this is waiting on additional review.

dlee992:
I don't have bandwidth to handle future change requests, so close it. 
In this PR, I basically give a clear POC to show where the bug is coming from but I didn't write it in a clear way that makes others hard to understand.
However, I do think it's a fundamental recompilation bug that we better fix it ASAP when available.

PR #9458, opened 2024-02-22
Add  `pinned` and `as_cuda_array` to cudasim mode

tornikeo:
I'm working with a small [cuda kernel project](https://github.com/tornikeo/cosine-similarity), and when testing my code with numba sim, I see that `cuda.pinned` has different arglist (single arg, instead of variadic *args) in these two modes for no good reason.

So, here's a small tweak to make it consistent.

### DIFF
--- a/numba/cuda/simulator/cudadrv/devicearray.py
+++ b/numba/cuda/simulator/cudadrv/devicearray.py
@@ -402,13 +402,7 @@ def auto_device(ary, stream=0, copy=True):
     return to_device(ary, stream, copy), True
 
 def as_cuda_array(obj, sync=True):
-    """Does nothing and returns the same input object.
-    
-    Primarily used as a simulated substitue of the real `as_cuda_array`
-    function for testing numba code, when GPU isn't available. 
-    
-    See :ref:`cuda array interface <cuda-array-interface>`.
-    """
+    "Do nothing, and return the same object"
     return obj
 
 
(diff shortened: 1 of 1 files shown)


gmarkall:
gpuci run tests

gmarkall:
Did this cause an error in your project? Or is this an aesthetic change only?

tornikeo:
@gmarkall  Not an error per se, but it forces me to monkeypatch the `cuda.pinned`. In a real run, I always pass several arrays in pinned. When I do testing, the pinned just errors out because of this. I'd expect a simulator to follow the real thing closely as possible (in syntax and everything).

tornikeo:
@gmarkall anything I should do for "Check Release Notes" failing?

gmarkall:
@tornikeo If you'd like a mention of your fix in the release notes, you could add a release note snippet RST as documented in: https://numba.readthedocs.io/en/stable/developer/contributing.html#release-notes - however, as this is a small change, I'm happy if you'd prefer me to just mark this to skip the release notes check, and then I don't think there's anything further you need to do - let me know what you'd prefer.

tornikeo:
@gmarkall  Ah, it's a tiny change. No need to pollute the release notes with this. Thanks! :+1:

tornikeo:
@gmarkall  somehow function `as_cuda_array` was missing from `numba/cuda/simulator/cudadrv/devicearray.py`. Ran into same issue when trying to test numba code on a gpuless machine. Added that function too.

PR #9459, opened 2024-02-22
Support pep695 type param syntax

### LINKED ISSUE #9443: `AttributeError` for inlined generic functions with py312 pep-695 type-parameter syntax
This regex doesn't match functions that use the [PEP 695](https://peps.python.org/pep-0695/) type parameters syntax:
https://github.com/numba/numba/blob/0.60.0dev0/numba/core/ir.py#L29

In my case, this resulted in:

```
.venv/lib/python3.12/site-packages/numba/core/ir.py:85: in _raw_function_name
    return self._defmatcher.match(defn.strip()).groups()[0]
E   AttributeError: 'NoneType' object has no attribute 'groups'
        defn       = 'def drink_cigar[N: int](X: MatrixF8[_1, N], b: VectorF8[_2]) -> VectorF8[N]:\n'
        self       = Loc(filename=/projects/clowns_are_trains/inflate_bouncy_castle.py, line=192, col=None)
```

This `drink_cigar()` function is wearing a `@njit(inline='always')`.
To give a bit more context, `drink_cigar` is passed as argument to some other (non-inlined) `@njit()` function, `paint_donkeys(drink_cigar, ...)` is this case.
Back when I was still using the archaic `TypeVars`, the code in question was doing what it was supposed to (without errors or warnings).

sklam:
Fix #9443. 

Patch simply adjusts regex to accept the syntax. 
This should be considered for 0.59.1 to unbreak support of Python 3.12 only syntax.

### DIFF
--- a/numba/core/ir.py
+++ b/numba/core/ir.py
@@ -26,7 +26,7 @@ class Loc(object):
     """Source location
 
     """
-    _defmatcher = re.compile(r'def\s+(\w+)\s*\[.*\]\s*\(.*')
+    _defmatcher = re.compile(r'def\s+(\w+)\s*[\[\(]')
 
     def __init__(self, filename, line, col=None, maybe_decorator=False):
         """ Arguments:
@@ -82,10 +82,11 @@ class Loc(object):
     def _raw_function_name(self):
         defn = self._find_definition()
         if defn:
-            return self._defmatcher.match(defn.strip()).groups()[0]
-        else:
-            # Probably exec(<string>) or REPL.
-            return None
+            m = self._defmatcher.match(defn.strip())
+            if m:
+                return m.groups()[0]
+        # Probably exec(<string>) or REPL.
+        return None
 
     def get_lines(self):
         if self.lines is None:
--- a/numba/core/ir.py
+++ b/numba/core/ir.py
@@ -26,7 +26,7 @@ class Loc(object):
     """Source location
 
     """
-    _defmatcher = re.compile(r'def\s+(\w+)\s*[\[\(]')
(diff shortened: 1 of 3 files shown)


numba/core/ir.py
```diff
@@ -26,7 +26,7 @@ class Loc(object):
     """Source location
 
     """
-    _defmatcher = re.compile(r'def\s+(\w+)\(.*')
+    _defmatcher = re.compile(r'def\s+(\w+)\s*\[.*\]\s*\(.*')
```
jorenham, line 29: Although I doubt that it would cause any issues, this regex currently also matches some invalid syntax, e.g.

```
def f[](a: T) -> T:
    assert a != 1
```

sklam: Your comment made me realize this regex can even be simpler. This regex is expected to run on valid source code. Its purpose is solely for extracting the function name. It doesn't even need to match the closing of the square bracket.

numba/core/ir.py
```diff
@@ -26,7 +26,7 @@ class Loc(object):
     """Source location
 
     """
-    _defmatcher = re.compile(r'def\s+(\w+)\(.*')
+    _defmatcher = re.compile(r'def\s+(\w+)\s*[\[\(]')
```
guilhermeleobas, line 29: `_raw_function_name` is the only function that uses this regex. If we only need the name, `r'def\s+(\w+)'` should be enough, no?

sklam: you're right.

guilhermeleobas (approved):

PR #9467, opened 2024-02-26
[DO NOT MERGE] New004

esc:
@kc611 @stuartarchibald -- here is the solution to the issue you are facing regarding preservation of git-history.

1) Make three branches, each does a `git mv` of `pythonapi.py` to `old_pythonapi.py`, `new_pythonapi.py` and `preservepythonapi.py` respectively.

2) Use an octo-merge to merge all three branches.

3) Make an additional commit that does `git mv preserve_pythonapi.py pythonapi.py` to restore the original file for editing. 

You can then edit all three files as needed and history will be retained through `git blame` without having to use `-C` and this should show up on GitHub.

Screenshot 2024-02-26 at 17 12 07

Hope this helps.


esc:
To be used in https://github.com/numba/numba/pull/9446 -- this does a double copy of a file whilst preserving the original and retaining the history visible through `git blame`.

esc:
closing now as purpose has been served

PR #9470, opened 2024-02-27
Add ability to link CUDA functions with in-memory PTX.

ed-o-saurus:
This commit adds the PTXCode class. It is a simple wrapper around a string. This allows the user to link CUDA functions with dynamically generated PTX code without having to write data to a file.

### DIFF
--- a/docs/source/cuda-reference/kernel.rst
+++ b/docs/source/cuda-reference/kernel.rst
@@ -10,6 +10,14 @@ be configured and launched:
 .. autofunction:: numba.cuda.jit
 
 
+.. _cuda-ptx-code:
+
+The ``cuda.PTXCode`` class is used to store in-memory PTX code to be linked
+with a CUDA dispatcher object
+
+.. autoclass:: numba.cuda.PTXCode
+
+
 Dispatcher objects
 ------------------
 
--- a/numba/cuda/codegen.py
+++ b/numba/cuda/codegen.py
@@ -13,6 +13,18 @@ import tempfile
 CUDA_TRIPLE = 'nvptx64-nvidia-cuda'
 
 
+class PTXCode :
+    """
+    Store in-memory PTX code to link with a compiled function.
+    """
+
+    def __init__(self, ptx) :
+        self._ptx = ptx
+
+    def __str__(self) :
+        return self._ptx
+
+
 def run_nvdisasm(cubin, flags):
     # nvdisasm only accepts input from a file, so we need to write out to a
     # temp file and clean up afterwards.
@@ -169,7 +181,10 @@ class CUDACodeLibrary(serialize.ReduceMixin, CodeLibrary):
         for ptx in ptxes:
             linker.add_ptx(ptx.encode())
         for path in self._linking_files:
-            linker.add_file_guess_ext(path)
+            if isinstance(path, PTXCode):
+                linker.add_ptx(path._ptx.encode(encoding='ascii'))
+            else :
+                linker.add_file_guess_ext(path)
         if self.needs_cudadevrt:
             linker.add_file_guess_ext(get_cudalib('cudadevrt', static=True))
 
--- a/numba/cuda/decorators.py
+++ b/numba/cuda/decorators.py
@@ -28,8 +28,8 @@ def jit(func_or_sig=None, device=False, inline=False, link=[], debug=None,
        .. note:: A kernel cannot have any return value.
     :param device: Indicates whether this is a device function.
     :type device: bool
-    :param link: A list of files containing PTX or CUDA C/C++ source to link
-       with the function
+    :param link: A list of files containing PTX or CUDA C/C++ source or
+      :class:`PTXCode <numba.cuda.PTXCode>` to link with the function.
     :type link: list
     :param debug: If True, check for exceptions thrown when executing the
        kernel. Since this degrades performance, this should only be used for
--- a/numba/cuda/device_init.py
+++ b/numba/cuda/device_init.py
@@ -22,6 +22,7 @@ from numba.cuda import initialize
 from .errors import KernelRuntimeError
 
 from .decorators import jit, declare_device
+from .codegen import PTXCode
 from .api import *
 from .api import _auto_device
 from .args import In, Out, InOut


gmarkall (commented):
Many thanks for the PR, @ed-o-saurus - this is an often-requested feature so I'm happy to see this PR!

There are a few thoughts I need to add on the design and scope of the changes, but I like the idea of using an object to hold the code whilst keeping strings being used for paths, maintaining backwards-compatibility.

I'll post a follow-up once I've had more time to get my thoughts down.

ed-o-saurus:
Is there an expected timeline for this feature to be included in Numba? The reason I ask is that I'm working on a project that uses it and I'd like to know if I should wait or implement a workaround in my project?

gmarkall:
@ed-o-saurus Apologies for the confusion here - are you using Linux and CUDA 12.0 or later? If so, the workaround is to add [`pynvjitlink`](https://github.com/rapidsai/pynvjitlink) as a dependency and call

```python
from pynvjitlink import patch
patch.patch_numba_linker()
```

then you can use `cuda.PTXSource` and the other classes (e.g. `cuda.CUSource`, etc.).

If you're on Windows, we have not yet implemented support for Windows but are working on it. If you're on Windows and need support right now, you could take the `LinkableCode` classes from pynvjitlink and adapt the patch to patch just the part of the Numba linker that handles them in your project.

ed-o-saurus:
@gmarkall Thank you for getting back to me. That meets my needs for now. 
However, I'm still unclear. Is the plan to incorporate the changes into `numba` eventually? That way it won't be necessary for users to use the `pynvjitlink` package.

gmarkall:
@ed-o-saurus Sorry I forgot to follow up earlier - the functionality is available in numba-cuda, but changes will not be backported to the built-in target - there is some explanation of this in https://numba.readthedocs.io/en/latest/cuda/overview.html#cuda-deprecation-status

If you install numba-cuda, you can execute something like the following, as long as pynvjitlink is enabled:

```python
from numba import cuda
import numpy as np


ptx = cuda.PTXSource("""
.version 6.5
.target sm_52
.address_size 64

.visible .func  (.param .b32 status) double_an_int32(
        .param .b64 retval,
        .param .b32 arg
)
{
        .reg .b32       %r<4>;
        .reg .b64       %retvalptr;

        ld.param.u64    %retvalptr, [retval];
        ld.param.u32    %r1, [arg];
        shl.b32         %r2, %r1, 1;
        st.u32          [%retvalptr], %r2;
        mov.u32         %r3, 0;
        st.param.b32    [status], %r3;
        ret;
}""".encode())

double_an_int32 = cuda.declare_device("double_an_int32",
                                      "int32(int32)")


@cuda.jit(link=[ptx])
def k(x):
    x[0] = double_an_int32(x[0])


x = np.ones(1, dtype=np.int32)
print(x[0])
k[1, 1](x)
print(x[0])
```

This should also be supported without pynvjitlink, but it will require the resolution of https://github.com/NVIDIA/numba-cuda/issues/101.

shaunc:
@gmarkall - I have a renewed need to use more advanced features of cooperative groups -- via cu code if necessary. In the forum thread - https://numba.discourse.group/t/pass-cooperative-group-to-external-routine-defined-in-cu-file/1742 - you tried to help me out, but came to the conclusion that it likely wasn't possible to pass numba-kernel-defined grid group to c.

However, what about replacing the kernel: kernel -> cu -> python device function, where the cu wrapper creates a grid group (possibly?) copies to heap, and passes back a void pointer? Then cu helper functions could always be called for the grid group operations.

For just the first step -- relevant to this thread -- can I actually call kernel -> cu -> py device function? Below is my sample code:

```py
"""
Test of ability to call back to a python device function from c device function.
"""

import os

import numba as nb  # type: ignore[import]
import numpy as np
from numba import cuda  # type: ignore[import]
from pynvjitlink import patch

patch.patch_numba_linker()


@cuda.jit(device=True, name="python_func")
def python_func(x):
    """
    Python device function.
    """
    return x * 2


@cuda.jit
def dummy_kernel(arr):
    """
    Dummy kernel to force compilation of python_func.
    """
    i = cuda.grid(1)  # type: ignore[assignment]
    if i < arr.size:
        arr[i] = python_func(arr[i])


# Compile the kernel to PTX
dummy_ptx = cuda.compile_ptx(dummy_kernel.py_func, (nb.float32[:],))[0]

CUDA_ARCH = "sm_80"  # ampere architecture
os.system(
    f"nvcc -lib -Xcompiler -fPIC -arch={CUDA_ARCH} "
    " --relocatable-device-code=true "
    "test_py_c_py.cu -o test_py_c_py.a"
)

# Declare C++ device functions
cpp_function = cuda.declare_device("cpp_function", "float32(float32)")


# Declare main kernel
@cuda.jit(link=["test_py_c_py.a", dummy_ptx.encode()])
def my_kernel(arr):
    i = cuda.grid(1)  # type: ignore[assignment]
    if i < arr.size:
        arr[i] = cpp_function(arr[i])  # type: ignore[no-untyped-call]


# Try to call the kernel
arr = cuda.to_device(np.array([1.0, 2.0, 3.0], dtype=np.float32))
my_kernel[1, 3](arr)
print(arr.copy_to_host())  # Output: [3.0, 5.0, 7.0]
```
with `test_py_c_py.cu` defined as:
```cu
extern "C" __device__ float python_func(float x);

__device__ float cpp_function(float x) {
    return python_func(x) + 1.0f;
}
```
However, on the kernel call, I get:
```
tests/cudautil/test_py_c_py.py:57: in <module>
    my_kernel[1, 3](arr)
/home/shauncutts/.cache/pypoetry/virtualenvs/ingolstadt-h73rb8xf-py3.11/lib/python3.11/site-packages/numba_cuda/numba/cuda/dispatcher.py:608: in __call__
/home/shauncutts/.cache/pypoetry/virtualenvs/ingolstadt-h73rb8xf-py3.11/lib/python3.11/site-packages/numba_cuda/numba/cuda/dispatcher.py:750: in call
/home/shauncutts/.cache/pypoetry/virtualenvs/ingolstadt-h73rb8xf-py3.11/lib/python3.11/site-packages/numba_cuda/numba/cuda/dispatcher.py:758: in _compile_for_args
/home/shauncutts/.cache/pypoetry/virtualenvs/ingolstadt-h73rb8xf-py3.11/lib/python3.11/site-packages/numba_cuda/numba/cuda/dispatcher.py:1005: in compile
    kernel.bind()
/home/shauncutts/.cache/pypoetry/virtualenvs/ingolstadt-h73rb8xf-py3.11/lib/python3.11/site-packages/numba_cuda/numba/cuda/dispatcher.py:257: in bind
    cufunc = self._codelibrary.get_cufunc()
/home/shauncutts/.cache/pypoetry/virtualenvs/ingolstadt-h73rb8xf-py3.11/lib/python3.11/site-packages/numba_cuda/numba/cuda/codegen.py:248: in get_cufunc
/home/shauncutts/.cache/pypoetry/virtualenvs/ingolstadt-h73rb8xf-py3.11/lib/python3.11/site-packages/numba_cuda/numba/cuda/codegen.py:227: in get_cubin
    self._link_all(linker, cc, ignore_nonlto=False)
/home/shauncutts/.cache/pypoetry/virtualenvs/ingolstadt-h73rb8xf-py3.11/lib/python3.11/site-packages/numba_cuda/numba/cuda/codegen.py:192: in _link_all
    linker.add_file_guess_ext(path, ignore_nonlto)
E   TypeError: PatchedLinker.add_file_guess_ext() takes 2 positional arguments but 3 were given
```

1) Have I gotten something confused here?
2) Is this approach overall going to work? (If plausible can start different issue or move discussion back to forum.)

PR #9474, opened 2024-03-01
(Do Merge) NumPy 2: Removed dead code conflicting with binary builds against NumPy 2

kc611:
As titled

### DIFF
--- a/numba/np/ufunc/_internal.c
+++ b/numba/np/ufunc/_internal.c
@@ -550,10 +550,10 @@ dufunc__add_loop(PyDUFuncObject * self, PyObject * args)
                                               arg_types_arr, &old_func) == 0) {
         /* TODO: Consider freeing any memory held by the old loop (somehow) */
         for (idx = 0; idx < ufunc->ntypes; idx++) {
-            if (ufunc->functions[idx] == (PyUFuncGenericFunction)loop_ptr) {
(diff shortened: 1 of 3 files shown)


numba/np/ufunc/_internal.c
```diff
@@ -545,16 +544,6 @@ dufunc__add_loop(PyDUFuncObject * self, PyObject * args)
...
-    } else if (PyUFunc_ReplaceLoopBySignature(ufunc,
-                                              (PyUFuncGenericFunction)loop_ptr,
-                                              arg_types_arr, &old_func) == 0) {
-        /* TODO: Consider freeing any memory held by the old loop (somehow) */
-        for (idx = 0; idx < ufunc->ntypes; idx++) {
-            if (ufunc->functions[idx] == (PyUFuncGenericFunction)loop_ptr) {
-                ufunc->data[idx] = data_ptr;
-                break;
-            }
-        }
```
sklam, line 557: I found a case that reaches this:

```python
from numba import vectorize, types


@vectorize
def foo(x):
    return x

foo.add(tuple([types.intp]))
foo.add(tuple([types.intp]))
```

I think it is valid for Numba to reject the second `.add`.

kc611:
This PR is to be continued at #9466 along wih all the related NumPy 2 changes

PR #9478, opened 2024-03-06
support assign a `tuple` to a row of a `recarray`

### LINKED ISSUE #9476: [feature request] assign a `tuple` to one row of `recarray`
test code:
```python
import numpy as np
from numba import njit


@njit
def foo(arr, name, age, height):
    items = (name, age, height)
    arr[1] = items
    return arr


dtype = [("Name", "U10"), ("Age", "i4"), ("Height", "f4")]
recarr = np.empty((2,), dtype=dtype)
recarr[0] = ("John", 25, 180.5)

print(recarr)
print(foo(recarr, "John", 25, 180.5))
```

np supports directly assign a tuple of given types to one row of recarray, but numba doesn't.
```
 
 >>> setitem(unaligned array(Record(Name[type=[unichr x 10];offset=0],Age[type=int64;offset=40],Height[type=float64;offset=48];56;False), 1d, C), Literal[int](1), Tuple(unicode_type, int64, float64))
```
If could, the user code can be only one-line, rather than multiple lines to assign each item.

kind of relating with #8471

dlee992:
A possible way to 
fix #9476


dlee992:
not sure why this failed. 
```
Found required file: docs/upcoming_changes/9478.improvement.rst

Checking naming convention.
Passed: Filename is valid

Checking file contents:

support assign a ``tuple`` to a row of a ``recarray``
--------------------------------------------------------

support assign a ``tuple`` to a row of a ``recarray``.
Error: Process completed with exit code 1.
```

dlee992:
oh, the reason is 
```
  File "/Users/dali/Code/open-numba/maint/towncrier_rst_validator.py", line 98, in <module>
    assert len(title) == len(underline), "Title and underline must be the same length." + \
AssertionError: Title and underline must be the same length. (Found Title: 53, Underline: 56)
```

dlee992:
Can this be marked with milestone of next release 0.60?

dlee992:
This one can enjoy a review! Thanks! @guilhermeleobas

PR #9481, opened 2024-03-07
Support return `int` rather than always `Optional` in `dct.get(key, default)`

### LINKED ISSUE #9473: want `int64` rather than `Optional(int64)` when `dict.get(key, default_int64)` and dict value type is `int64`
Coming from gitter discussion (https://matrix.to/#/!gFwSEpcuaweltInwDW:gitter.im/$GQbCOdu6zbw3baQZM2CSSAm2xfw2bvWkgf3MImzOPlk?via=gitter.im&via=matrix.org&via=tryp.io), a test case: 
```python
from os import environ

environ["NUMBA_CHROME_TRACE"] = "5.json"
# environ["NUMBA_DUMP_LLVM"] = "1"
environ["NUMBA_DEBUG_TYPEINFER"] = "1"

from numba import njit
from numba.core import types
from numba.typed import Dict


@njit
def goo():
    d = Dict.empty(types.int64, value_type=types.int64)
    d[0] = 0
    v = d.get(2, 0)
    return v

print(goo())
```
The root cause is the @intrinsic `_dict_lookup` will forcily generate `Optional` for the return value at 
```python
        out = context.make_optional_none(builder, td.value_type)
```
A potential fix is that let `_dict_lookup` knows the default_value type, and choose using `Optional` or not based on that.

dlee992:
A hacky/flaky way to
fix #9473 

This test case can pass:
```python
import numpy as np
from numba import njit
from numba.core import types
from numba.typed import Dict

@njit
def goo(arr):
    d = Dict.empty(types.int64, value_type=types.int64)
    d[0] = 0
    v1 = d.get(0, 0)
    arr[v1] = -1
    v2 = d.get(2, 9)
    arr[v2] = -2
    v3 = d.get(3)
    return arr, v3

arr = np.arange(10)
print(goo(arr))
```

### DIFF
--- a/docs/upcoming_changes/9481.improvement.rst
+++ b/docs/upcoming_changes/9481.improvement.rst
@@ -1,0 +1,4 @@
+Support return `int` rather than always `Optional` in `dct.get(key, default)`
+-----------------------------------------------------------------------------
+
+Supports return `int` rather than always `Optional` in `dct.get(key, default)`.
--- a/numba/tests/test_dicts.py
+++ b/numba/tests/test_dicts.py
@@ -229,5 +229,21 @@ class TestCompiledDict(TestCase):
             str(raises.exception),
         )
 
+    def test_dict_get_with_default_value(self):
+        # Test dict.get() with default value, then use the value as arr index
+        @njit
+        def goo(arr):
+            d = {0: 0, 1: 1}
+            d[0] = 0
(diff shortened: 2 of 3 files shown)


sklam:
We are assuming this PR is still in progress since it does not have a test yet.

dlee992:
Hi, test is added @sklam

Issue #9490, opened 2024-03-11
Non-det bug with prange code


sklam:
This issue is reported on discourse https://numba.discourse.group/t/nested-loop-fails-in-version-0-59/ and affects 0.59.

Behavior is non-deterministic on 0.59. Git bisect points to https://github.com/numba/numba/pull/9255 as first bad commit.
Running reproducer on 0.58 have always passed when I tested locally.

Given OP discover that the bug depends on Python versions, the current guess is that it is dependent on either the bytecode ordering or the behavior (iteration ordering) for containers.

This is a blocking issue for 0.59.1

loicdtx:
Really sorry to re-open this issue, but it appears that the bug has only been partially resolved. My suggestion to [simplify the specific test](https://github.com/numba/numba/pull/9491/commits/0538805d812536f8d496576e636dcc7e902ad282) for this bug is apparently bypassing some of the offending bits.

If I slightly modify the test and add a third condition, I get a segmentation fault.

Current test (passes in 0.59.1)

```python
@numba.jit(nopython=True, parallel=True)
def stable_fit(X, y, threshold=3):
    min_obs = int(X.shape[1] * 1.5)
    beta = np.zeros((X.shape[1], y.shape[1]), dtype=np.float64)
    residuals = np.full_like(y, np.nan)
    stable = np.empty((y.shape[1]))
    for idx in numba.prange(y.shape[1]):
        y_sub = y[:, idx]
        isna = np.isnan(y_sub)
        X_sub = X[~isna]
        y_sub = y_sub[~isna]
        is_stable = False

        # Run until minimum observations
        # or until stability is reached
        for jdx in range(len(y_sub), min_obs - 1, -2):
            # Timeseries gets reduced by two elements
            # each iteration
            y_ = y_sub[-jdx:]
            X_ = X_sub[-jdx:]
            beta_sub = np.linalg.solve(np.dot(X_.T, X_), np.dot(X_.T, y_))
            resid_sub = np.dot(X_, beta_sub) - y_
            # Check for stability
            rmse = np.sqrt(np.mean(resid_sub ** 2))
            first = np.fabs(resid_sub[0]) / rmse < threshold
            last = np.fabs(resid_sub[-1]) / rmse < threshold
            # Break if stability is reached
            is_stable = first & last
            if is_stable:
                break

        beta[:, idx] = beta_sub
        residuals[-jdx:, idx] = resid_sub
        stable[idx] = is_stable
[...]

PR #9491, opened 2024-03-11
Fix non-deterministic bug caused by unstableness in SSA 

sklam:
Fixes https://github.com/numba/numba/issues/9490

### DIFF
--- a/numba/core/utils.py
+++ b/numba/core/utils.py
@@ -418,7 +418,7 @@ class MutableSortedSet(MutableSet[T], _tp.Generic[T]):
         return len(self._values)
 
     def __iter__(self):
-        return iter(k for k in (self._values))
+        return iter(k for k in sorted(self._values))
 
     def __contains__(self, x: T) -> bool:
         return self._values.__contains__(x)
--- a/numba/core/utils.py
+++ b/numba/core/utils.py
@@ -385,6 +385,28 @@ def order_by_target_specificity(target, templates, fnkey=''):
 T = _tp.TypeVar('T')
 
 
+class OrderedSet(MutableSet[T]):
+
+    def __init__(self, iterable: _tp.Iterable[T] = ()):
+        # Just uses a dictionary under-the-hood to maintain insertion order.
+        self._data = dict.fromkeys(iterable, None)
+
+    def __contains__(self, key):
+        return key in self._data
+
+    def __iter__(self):
+        return iter(self._data)
+
+    def __len__(self):
+        return len(self._data)
+
+    def add(self, item):
+        self._data[item] = None
+
+    def discard(self, item):
+        self._data.pop(item, None)
+
+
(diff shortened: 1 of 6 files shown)


sklam:
https://github.com/numba/numba/pull/9491/commits/d848d30f88d6fe804bffcfb7184e6c1fb928420f is needed to make the subprocess-test not accidentally running without isolation. What can happen is that the `TestParforsRunner` runner launches each `TestCase` class in subprocess with `SUBPROC_TEST=1` then any test that needs further isolation by using `@TestCase.run_test_in_subprocess` will pickup the `SUBPROC_TEST`. The change makes `run_test_in_subprocess` set and expects in the runner that `SUBPROC_TEST` is equal to the test name. The `TestParforsRunner` will still just set `SUBPROC_TEST=1`.

(CC @stuartarchibald)

sklam:
`numba_smoketest_cpu_yaml_190`

sklam:
Reviewer notes from triage meeting:

- accidental stableness by sort?
- will removal of assignments impact liveness hack?

sklam:
> Reviewer notes from triage meeting:
> 
> * accidental stableness by sort?
> * will removal of assignments impact liveness hack?

https://github.com/numba/numba/pull/9491/commits/6b1de4e01bb708be47891d06bf48df2f2a160932 resolved these.

- `violators` is now based on insertion order which is in turn based on ordering of occurrence in bytecode.
- removal of self-assignments is not needed now. maybe i got it wrong initially because of non-det tests or that the use of insertion-order (above) helped.

sklam:
`numba_smoketest_cpu_yaml_191`

loicdtx:
In case that helps I applied some minor changes to the test case. See https://github.com/sklam/numba/pull/5

sklam:
@loicdtx, Does the changes affect the ability to reproduce the error? or does it reveal a different problem? I found the test very sensitive to changes. If the changes are purely for algorithm correctness in `stable_fit`, I'd prefer to not apply them so that we preserve the failing characteristic.

sklam:
I confirmed https://github.com/numba/numba/pull/9491/commits/0538805d812536f8d496576e636dcc7e902ad282 produces the same error with `PYTHONHASHSEED=1` without the fix.

numba/core/utils.py
```diff
@@ -396,7 +418,7 @@ def __len__(self):
         return len(self._values)
 
     def __iter__(self):
-        return iter(k for k in sorted(self._values))
+        return iter(k for k in (self._values))
```
sklam, line 421:
```suggestion
        return iter(k for k in sorted(self._values))
```
Shouldn't have removed the `sorted`. It was part of debugging and i  accidentally added to the patch. It turns out only the RVSDG frontend uses this so no tests were affected.

guilhermeleobas (approved):
Can confirm the proposed patch fixes the bug (for now? 😅).

It would be good to know one day why the ordering matters here.

PR #9495, opened 2024-03-13
Schedule guvectorize'd functions over dask.distributed

crusaderky:
Fix failure to pickle a dask array that embeds a guvectorized ufunc in its graph.

- Closes https://github.com/numba/numba/issues/4314
- Closes https://github.com/dask/distributed/issues/3450
- Closes https://github.com/dask/distributed/issues/7929


numba/tests/npyufunc/test_ufuncbuilding.py
```diff
@@ -317,6 +320,58 @@ def test_guvectorize_invalid_layout(self):
...
+        return self.array
+
+    def tolist(self):
+        return self.array.tolist()
+
+    def __array_ufunc__(self, ufunc, method, *args, **kwargs):
+        if method != "__call__":
+            return NotImplemented
+
+        return NEP13Array(ufunc(*[np.asarray(x) for x in args], **kwargs))
```
guilhermeleobas, line 338: Is this change needed for the PR to work?

guilhermeleobas (commented):
@crusaderky, thanks for your contribution. Do you have a test case using Dask where current code would fail?

numba/tests/npyufunc/test_ufuncbuilding.py
crusaderky: No, but I thought it would be easier to read if two very similar stub classes were declared next to each other.

crusaderky:
> @crusaderky, thanks for your contribution. Do you have a test case using Dask where current code would fail?

```python
import dask.array as da
import distributed
import numba

client = distributed.Client(processes=True, n_workers=1)

@numba.guvectorize(["f8,f8[:]"], "()->()")
def double(x, out):
    out[:] = x * 2

double(da.arange(5)).compute()
```

numba/tests/npyufunc/test_ufuncbuilding.py
crusaderky: No;  I felt that it would make it easier to understand the test ccode if these very two similar, and yet subtly different stub classes were presented right next to each other. I can revert if you prefer.

guilhermeleobas (dismissed):
Can confirm the changes fixes the issue.

guilhermeleobas:
@crusaderky, can you add a release notes file? https://numba.readthedocs.io/en/latest/developer/contributing.html#release-notes

crusaderky:
> @crusaderky, can you add a release notes file? https://numba.readthedocs.io/en/latest/developer/contributing.html#release-notes

Done

guilhermeleobas (approved):
Thanks, @crusaderky

Sorry for the delay on this last review.

Issue #9433, opened 2024-02-13
np.median raises AssertionError for empty arrays while numpy returns nan


leoschwarz:
## Reporting a bug

- [x] I have tried using the latest released version of Numba (most recent is
  visible in the release notes
  (https://numba.readthedocs.io/en/stable/release-notes-overview.html).
- [x] I have included a self contained code sample to reproduce the problem.
  i.e. it's possible to run as 'python bug.py'.

This is a very similar issue to #8451 but here it affects `numpy.median` and setting the error_model does not change the outcome.

```python
import unittest
import numpy as np
from numba import njit

def median_numpy(arr):
    return np.median(arr)

@njit
def median_numba(arr):
    return np.median(arr)

class TestCase(unittest.TestCase):
    def test_median_numpy(self):
        self.assertTrue(np.isnan(median_numpy(np.array([], dtype=float))))
        self.assertTrue(np.isnan(median_numpy(np.array([]))))

    def test_median_numba(self):
        self.assertTrue(np.isnan(median_numba(np.array([], dtype=float))))
        self.assertTrue(np.isnan(median_numba(np.array([]))))

if __name__ == "__main__":
    unittest.main()
```

While numpy returns `nan`, numba raises an AssertionError:

```
Failure
Traceback (most recent call last):
  File "/Users/leo/code/msi/code/tests/unit/misc/test_numpy_util.py", line 88, in test_median_numba
    self.assertTrue(np.isnan(NumpyUtil.median_numba(np.array([], dtype=float))))
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/leo/.miniconda3/envs/exp-2023-10/lib/python3.11/site-packages/numba/np/arraymath.py", line 1556, in _select_two
    assert high > low  # by construction
AssertionError
```

kc611:
Thanks for reporting this. I can reproduce this locally.

The error is possibly due to lack of a guards around the internal implementation for `np.median` within Numba. 

It originates from the following line:

https://github.com/numba/numba/blob/3edb458c69daeaaffbfc938145c367eda1cf043f/numba/np/arraymath.py#L1582

Where `n = shape of the flattened array = 0` which makes the logic think that variable `low` equals `0` and variable `high` equals `-1`.

A possible fix is to simply check for this edge case and return `nan` if no elements exist within the array. Marking this as a good first issue.

tanishac25:
[...]

PR #9498, opened 2024-03-14
median raises assertion error for an empty array

adityav1810:
This is my first time contributing to an opensource project. 
Solving issue #9433  
It would be great if someone could help me figure out where can i add some test cases to support this solution.

### DIFF
--- a/numba/np/arraymath.py
+++ b/numba/np/arraymath.py
@@ -1575,9 +1575,6 @@ def _median_inner(temp_arry, n):
     The main logic of the median() call.  *temp_arry* must be disposable,
     as this function will mutate it.
     """
-    # check if temp_arry is empty, return nan 
-    if n == 0 : 
-        return np.nan
     low = 0
     high = n - 1
     half = n >> 1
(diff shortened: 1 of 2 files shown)


numba/np/arraymath.py
```diff
@@ -1595,7 +1595,11 @@ def median_impl(a):
         # workspace anyway
         temp_arry = a.flatten()
         n = temp_arry.shape[0]
-        return _median_inner(temp_arry, n)
+        if n != 0: 
+            return _median_inner(temp_arry, n)
+        else:
+            raise ValueError('Cannot compute median of an empty array')
```
kc611, line 1601: I think we need follow NumPy behavior over here and return `np.nan` instead of raising a `ValueError`
```suggestion
            return np.nan
```

(`np.median(np.array([]))` outputs `np.nan` in NumPy)

kc611 (changes requested):
Hi @adityav1810, Thank you for your contribution to Numba. :smile: 

I think the tests you're looking for is in [numba/tests/test_array_reductions.py::TestArrayReductions::check_median_basic](https://github.com/numba/numba/blob/7135f8884ebc4c12d31012b185ac8e8355fcfcc4/numba/tests/test_array_reductions.py#L303), the actual test itself (on line 343 of same file) is quite confusing, so you'd want to simply add a empty array case in this particular function and run the `check` function on it. 

Also do remember to run `flake8` on the file that you change.

Feel free to ask questions if any.

adityav1810:
Hi @kc611 . Thanks for the help. I added the test and ran `flake8` as well. However, it seems like the tests fail. Anyidea what I am missing?

PR #9513, opened 2024-03-27
Type system implementation #1: Added initial implementation for a new type system using redundancies.

kc611:
This is an alternative to #9408. 

It does the same changes except those are changes are being done in entirely new files to prevent any collision with old numba system and also included NumPy 2 binary support.


kc611:
Closing in favour of #9662

PR #9520, opened 2024-04-01
give primitive types a faster key&value compare, copy, and zero methods in TypedDict

### LINKED ISSUE #9519: [Feature Request] `key_equal`, `copy_key`, `zero_key` in dict is slower than direct assignment if key type is primitive
Hi, I noticed an unoptimized situation that `key_equal`, `copy_key`, `zero_key` in dict are slower than direct assignment if key type is primitive. The root cause is if key_type doesn't contain meminfo, then `key_equal` will rollback to using `memcmp`, which is pretty slow compared to directly `this_key == an_integer`. Other two functions will rollback to using `memcpy`, which is also slow.

https://github.com/numba/numba/blob/a0605597430bb12c434dd116bc5eb84fb30513e0/numba/cext/dictobject.c#L448

https://github.com/numba/numba/blob/a0605597430bb12c434dd116bc5eb84fb30513e0/numba/cext/dictobject.c#L434

I think we can do more things in this intrinsic function, which includes generation of corresponding functions (i.e., `key_equal`, `copy_key`, `zero_key`) for primitive types. I have already tested this in an internal use-case, this optimization can boost performance at least 5%~10% if using numba typed.dict heavily (i.e., lots of dict lookup, insert operations)

https://github.com/numba/numba/blob/a0605597430bb12c434dd116bc5eb84fb30513e0/numba/typed/dictobject.py#L264

dlee992:
fixes #9519

### DIFF
--- a/numba/core/datamodel/models.py
+++ b/numba/core/datamodel/models.py
@@ -179,9 +179,8 @@ class OmittedArgDataModel(DataModel):
 @register_default(types.Boolean)
 @register_default(types.BooleanLiteral)
 class BooleanModel(DataModel):
-    _bit_type = ir.IntType(1)
+    be_type = _bit_type = ir.IntType(1)
     _byte_type = ir.IntType(8)
-    be_type = ir.IntType(1)
 
     def get_value_type(self):
         return self._bit_type
--- a/numba/core/datamodel/models.py
+++ b/numba/core/datamodel/models.py
@@ -181,6 +181,7 @@ class OmittedArgDataModel(DataModel):
 class BooleanModel(DataModel):
     _bit_type = ir.IntType(1)
     _byte_type = ir.IntType(8)
+    be_type = ir.IntType(1)
 
     def get_value_type(self):
         return self._bit_type
--- a/numba/typed/dictobject.py
+++ b/numba/typed/dictobject.py
@@ -310,7 +310,7 @@ def _dict_set_method_table(typingctx, dp, keyty, valty):
         dm_key = context.data_model_manager[keyty.instance_type]
         if dm_key.contains_nrt_meminfo():
             key_equal = _get_container_equal(
-                context, builder.module, dm_key, "dict_key")
+                context, builder.module, dm_key)
             key_incref, key_decref = _get_incref_decref(
                 context, builder.module, dm_key, "dict_key"
             )
@@ -334,7 +334,6 @@ def _dict_set_method_table(typingctx, dp, keyty, valty):
                 context,
                 builder.module,
                 dm_key,
-                "dict_key",
             )
             key_copy = _get_copy(
                 context,
--- a/numba/typed/dictobject.py
+++ b/numba/typed/dictobject.py
@@ -310,9 +310,9 @@ def _dict_set_method_table(typingctx, dp, keyty, valty):
         dm_key = context.data_model_manager[keyty.instance_type]
         if dm_key.contains_nrt_meminfo():
             key_equal = _get_container_equal(
-                context, builder.module, dm_key)
+                context, builder.module, dm_key, 'dict_key')
             key_incref, key_decref = _get_incref_decref(
-                context, builder.module, dm_key, "dict_key"
+                context, builder.module, dm_key, 'dict_key'
             )
             builder.store(
                 builder.bitcast(key_equal, key_equal_ptr.type.pointee),
@@ -363,7 +363,7 @@ def _dict_set_method_table(typingctx, dp, keyty, valty):
         dm_val = context.data_model_manager[valty.instance_type]
         if dm_val.contains_nrt_meminfo():
             val_incref, val_decref = _get_incref_decref(
-                context, builder.module, dm_val, "dict_value"
+                context, builder.module, dm_val, 'dict_value'
             )
             builder.store(
                 builder.bitcast(val_incref, value_incref_ptr.type.pointee),
--- a/numba/typed/dictobject.py
+++ b/numba/typed/dictobject.py
@@ -324,7 +324,8 @@ def _dict_set_method_table(typingctx, dp, keyty, valty):
                 builder.bitcast(key_decref, key_decref_ptr.type.pointee),
                 key_decref_ptr,
             )
-        else:
+
+        if isinstance(dm_key, models.PrimitiveModel):
             # dm_key is a primitive type, e.g., int64
             # generate key_equal, key_copy, key_zero
             key_equal = _get_primitive_equal(
@@ -371,7 +372,8 @@ def _dict_set_method_table(typingctx, dp, keyty, valty):
                 builder.bitcast(val_decref, value_decref_ptr.type.pointee),
                 value_decref_ptr,
             )
-        else:
+
+        if isinstance(dm_val, models.PrimitiveModel):
             # dm_val doesn't contain meminfo, is a primitive type, e.g., int64
             # generate value_copy, value_zero
             value_copy = _get_copy(
--- a/numba/typed/dictobject.py
+++ b/numba/typed/dictobject.py
@@ -36,7 +36,9 @@ from numba.core import typing
 from numba.typed.typedobjectutils import (_as_bytes, _cast, _nonoptional,
                                           _sentry_safe_cast_default,
                                           _get_incref_decref,
-                                          _get_equal, _container_get_data,)
+                                          _get_equal, _container_get_data,
+                                          _get_primitive_equal,
+                                          _get_copy, _get_zero,)
 
 ll_dict_type = cgutils.voidptr_t
 ll_dictiter_type = cgutils.voidptr_t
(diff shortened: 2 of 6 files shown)


numba/core/datamodel/models.py
```diff
@@ -181,6 +181,7 @@ def from_argument(self, builder, val):
 class BooleanModel(DataModel):
     _bit_type = ir.IntType(1)
     _byte_type = ir.IntType(8)
+    be_type = ir.IntType(1)
```
guilhermeleobas, line 184: Isn't `be_type` and `_bit_type` the same thing?

numba/typed/typedobjectutils.py
```diff
@@ -197,3 +197,78 @@ def build_wrapper(fn):
...
+def _get_primitive_equal(context, module, datamodel, element_type):
+    assert not datamodel.contains_nrt_meminfo()
+
+    data_ptr_ty = ir.IntType(8).as_pointer()
+    equal_fnty = ir.FunctionType(ir.IntType(32), [data_ptr_ty, data_ptr_ty])
+    equal_fn = cgutils.get_or_insert_function(
+        module,
+        equal_fnty,
+        name='.numba_{}.{}_equal'.format(
+            context.fndesc.mangled_name, element_type),
```
guilhermeleobas, line 211: NIT: you can use an f-string here

numba/typed/typedobjectutils.py
```diff
@@ -197,3 +197,78 @@ def build_wrapper(fn):
...
+    )
+    builder = ir.IRBuilder(equal_fn.append_basic_block())
```
guilhermeleobas, line 213: You're creating a new builder here, but you can just re-use the one from `_dict_set_method_table` codegen.

numba/typed/typedobjectutils.py
```diff
@@ -197,3 +197,78 @@ def build_wrapper(fn):
     builder.ret(context.get_constant(types.int32, -1))
 
     return equal_fn
+
+
+def _get_primitive_equal(context, module, datamodel, element_type):
```
guilhermeleobas, line 202: I'd change `element_type` to something more meaningful. It is not an element type but rather a suffix: `_dict_key`. In fact, there's only one place this function is called.

numba/typed/typedobjectutils.py
```diff
@@ -197,3 +197,78 @@ def build_wrapper(fn):
...
+        res = builder.icmp_signed('==', lhs, rhs)
+    else:
+        res = builder.icmp_unsigned('==', lhs, rhs)
+    with builder.if_else(res) as (then, orelse):
+        with then:
+            builder.ret(context.get_constant(types.int32, 1))
+        with orelse:
+            builder.ret(context.get_constant(types.int32, 0))
+
+    builder.ret(context.get_constant(types.int32, -1))
```
guilhermeleobas, line 228: You can return `res` instead, as it will be an integer holding either `0` or `1`.

guilhermeleobas (commented):
Thanks for your contribution, @dlee992. Ping me once the PR is ready and I can review it.

dlee992:
Thanks @guilhermeleobas !

I'm not familiar with IR Builder stuff, your comments are valuable for me! Will address them one-by-one.

dlee992:
Off topic: I often saw 
```
CondaHTTPError: HTTP 524 CONNECTION FAILED for url <[https://conda.anaconda.org/numba/label/dev/linux-64/llvmlite-0.43.0dev0-py310he1b5a44_12.tar.bz2>](https://conda.anaconda.org/numba/label/dev/linux-64/llvmlite-0.43.0dev0-py310he1b5a44_12.tar.bz2%3E)
``` 
this network error in CI recently, but normal users don't have permissions to rerun this single job. 

Perhaps someone can add retries=2 in CI config if detecting this kind of HTTP Error?

numba/typed/typedobjectutils.py
dlee992: In fact, I tried to directly `builder.ret(res)`, but I will get a compilation error, to complain I messed up sth with `i32`. 
Need to dig more.

numba/typed/typedobjectutils.py
dlee992: Sure, just the legacy code used this format, I just followed it..
Can change to f-format.

numba/typed/typedobjectutils.py
dlee992: Emm, I tried naively reusing the builder from `_dict_set_method_table`, the replace this `ir.IRBuilder(..)` with `builder.append_basic_block()`, then I got a lowering error.
Looks like I didn't point this builder to the first bb in the new function. Am I missing anything?

BTW, the creation way is also from the legacy code in this file.

dlee992:
I think this is ready for review. 
All failing jobs are `Before install` or `Build error` related to HTTP errors and llvmlite.
All other jobs are passed. 

Will add a release note soon and post some perf test and result.

numba/typed/typedobjectutils.py
dlee992: Oh, I realized if I change the return value type of `key_equal` in C declaration to `bool`, perhaps I don't need to add this branch. Here is a snippet of generated IR:
```llvm
; Function Attrs: mustprogress nofree norecurse nosync nounwind readonly willreturn
define i32 @.numba__ZN5numba5typed10dictobject13impl_new_dict12_3clocals_3e3impB3v15B42c8tJTIcFKzyF2ILShI4CrgQElcaMVJhEEUSJJgA_3dE16class_28int64_2916class_28int64_29x.dict_key_equal(i8* nocapture readonly %.1, i8* nocapture readonly %.2) #0 {
.4:
  %.5 = bitcast i8* %.1 to i64*
  %.6 = load i64, i64* %.5, align 8
  %.7 = bitcast i8* %.2 to i64*
  %.8 = load i64, i64* %.7, align 8
  %.9 = icmp eq i64 %.6, %.8
  %. = zext i1 %.9 to i32
  ret i32 %.
}
```

dlee992: But in the usage side, it wants to get 3 types of results as the comment pointed out, so I guess I can't directly return a `bool`. Here is the comment for `key_equal`:
```c
/* Returns -1 for error; 0 for not equal; 1 for equal */
```

dlee992:
I probably want to deprecate this PR, since we possibly have a better way to finish this goal.

guilhermeleobas:
Hi @dlee992, I can review this by the end of the week. Don't worry about the CI failures that are not directly related to your PR.

dlee992:
@guilhermeleobas, cool! All checks have passed this time.

A concern: as I mentioned in a previous comment `a more profound inherent limitation`, do you think PIXIE can overcome this limitation, given it would emit&inject LLVM IR of C/C++ code into numba generated LLVM module, which can bring more optimisation chance?

numba/core/datamodel/models.py
```diff
@@ -179,7 +179,7 @@ def from_argument(self, builder, val):
 @register_default(types.Boolean)
 @register_default(types.BooleanLiteral)
 class BooleanModel(DataModel):
-    _bit_type = ir.IntType(1)
+    be_type = _bit_type = ir.IntType(1)
```
guilhermeleobas, line 182:
```suggestion
    _bit_type = ir.IntType(1)
```

Couldn't find any place where `be_type` as actually used.

numba/typed/dictobject.py
```diff
@@ -262,46 +265,57 @@ def codegen(context, builder, sig, args):
...
+        key_incref_ptr = cgutils.gep_inbounds(builder, vtable, 0, 5)
+        key_decref_ptr = cgutils.gep_inbounds(builder, vtable, 0, 6)
+        value_incref_ptr = cgutils.gep_inbounds(builder, vtable, 0, 7)
+        value_decref_ptr = cgutils.gep_inbounds(builder, vtable, 0, 8)
 
         dm_key = context.data_model_manager[keyty.instance_type]
         if dm_key.contains_nrt_meminfo():
-            equal = _get_equal(context, builder.module, dm_key, 'dict_key')
+            key_equal = _get_container_equal(
+                context, builder.module, dm_key, "dict_key")
```
guilhermeleobas, line 313: Is this the only place that calls `_get_container_equal`? If so, the `"dict_key"` suffix can be moved to inside the function, as it won't change.

numba/core/datamodel/models.py
guilhermeleobas: Oh, it can be used inside one of the functions you're generating here.

numba/typed/typedobjectutils.py
```diff
@@ -197,3 +197,80 @@ def build_wrapper(fn):
...
+
+    with builder.if_else(res) as (then, orelse):
+        with then:
+            builder.ret(context.get_constant_generic(builder, types.int32, 1))
+        with orelse:
+            builder.ret(context.get_constant_generic(builder, types.int32, 0))
+
+    builder.ret(context.get_constant_generic(builder, types.int32, -1))
+
+    return equal_fn
```
guilhermeleobas, line 232: You can simplify the implementation like this:

```python
def _get_primitive_equal(context, module, datamodel, container_element_type):
    # construct equal_fn
    i8 = ir.IntType(8)
    i8p = i8.as_pointer()
    i32 = ir.IntType(32)

    equal_fnty = ir.FunctionType(i32, [i8p, i8p])
    equal_fn = cgutils.get_or_insert_function(
        module,
        equal_fnty,
        name=f".numba_{context.fndesc.mangled_name}.{container_element_type}_equal",
    )

    # populate the body of equal_fn
    builder = ir.IRBuilder(equal_fn.append_basic_block())
    lhs_ptr = builder.bitcast(equal_fn.args[0], datamodel.be_type.as_pointer())
    lhs = builder.load(lhs_ptr)
    rhs_ptr = builder.bitcast(equal_fn.args[1], datamodel.be_type.as_pointer())
    rhs = builder.load(rhs_ptr)

    if datamodel.fe_type.signed:
        res = builder.icmp_signed('==', lhs, rhs)
    else:
        res = builder.icmp_unsigned('==', lhs, rhs)
    builder.ret(builder.zext(res, i32))
    return equal_fn
```

numba/typed/typedobjectutils.py
```diff
@@ -197,3 +197,80 @@ def build_wrapper(fn):
     builder.ret(context.get_constant(types.int32, -1))
 
     return equal_fn
+
+
+def _get_primitive_equal(context, module, datamodel, func_name_suffix):
```
guilhermeleobas, line 202: If I understood correctly, this function would be equivalent to load two pointers and compare if their contents are equal? What is the input to those functions? Is the dictionary itself?

numba/typed/typedobjectutils.py
```diff
@@ -149,7 +149,7 @@ def _get_incref_decref(context, module, datamodel, container_element_type):
     return incref_fn, decref_fn
 
 
-def _get_equal(context, module, datamodel, container_element_type):
+def _get_container_equal(context, module, datamodel, func_name_suffix):
```
guilhermeleobas, line 152: You can leave the name as `container_element_type` as it is what it's used in the rest of this file.

guilhermeleobas (changes requested):
Hi @dlee992, your code seems to be ok, and it gives speedup in the example you provided. I have some comments and a couple of questions. Ping me once you're done and I can review it again.

numba/typed/typedobjectutils.py
dlee992: removed it, since it can be an internal detail.

numba/typed/typedobjectutils.py
dlee992: sure, simplified this function. especially use `zext` to avoid explicit if-branch.

numba/typed/dictobject.py
dlee992: yes, it's the only one usage. hide this argument into the function.

numba/typed/typedobjectutils.py
dlee992: here: 
```c
/* Returns -1 for error; 0 for not equal; 1 for equal */
static int
key_equal(NB_DictKeys *dk, const char *lhs, const char *rhs) {
    if ( dk->methods.key_equal ) {
        return dk->methods.key_equal(lhs, rhs);
    } else {
        return memcmp(lhs, rhs, dk->key_size) == 0;
    }
}
```
input arguments are `lhs` and `rhs` if we define a `key_equal` function in LLVM IR module.

dlee992:
not sure what changes, but the result of the mini-benchmark becomes 0.7 vs 1.4, speed up is 2x.

Can we put this into `0.60rc`?

gentle ping @guilhermeleobas , can review this again.

BTW, I noticed related `_key_equal`, `_key_copy` and `_val_copy` will appear in the LLVM IR module many times. Not sure if it's expected.

dlee992:
Looks like we can't put this into `0.60`. Then perhaps put it into `0.61rc`?

numba/typed/typedobjectutils.py
```diff
@@ -173,14 +173,18 @@ def build_wrapper(fn):
         context.call_conv.return_value(builder, intres)
 
     wrapfn = cgutils.get_or_insert_function(
-        module, wrapfnty, name='.numba_{}.{}_equal.wrap'.format(
-            context.fndesc.mangled_name, container_element_type))
+        module,
+        wrapfnty,
+        name=f".numba_{context.fndesc.mangled_name}.dict_key_equal.wrap",
+    )
```
guilhermeleobas, line 179: I was looking at https://github.com/numba/numba/pull/6402 and perhaps the `container_element_type` may be needed. Could you undo the changes related to removing the container_element_type? Just want to make sure we don't introduce any regression with this PR.

guilhermeleobas (changes requested):
Hi @dlee992, sorry for the delay as I was out for some days. The changes you proposed looks good to me, but want to make sure we don't introduce any regression fixed in https://github.com/numba/numba/pull/6402. Could you undo the changes related to `container_element_type`?

numba/typed/typedobjectutils.py
dlee992: Sure, addressed.

dlee992:
I believe this PR has been reviewed from you in a very quick speed! Compared to my other PRs. : )

guilhermeleobas:
> I believe this PR has been reviewed from you in a very quick speed! Compared to my other PRs. : )

If you need review in any other PR, feel free to ping me.

guilhermeleobas:
Since this touches part of the C code, it would be good if anyone else could review. Meanwhile, do you have the number of how fast these functions are? If so, could you also share the code you're using to benchmark these functions?

dlee992:
Here is a mini-benchmark (I guess you forgot this?) https://github.com/numba/numba/pull/9520#issuecomment-2033243981

guilhermeleobas:
> Here is a mini-benchmark (I guess you forgot this?) [#9520 (comment)](https://github.com/numba/numba/pull/9520#issuecomment-2033243981)

Oops, my bad. I'll try locally

guilhermeleobas (approved):

dlee992:
After 2nd thought, perhaps this way is too tricky to maintain? I won't mind if you don't wanna merge this into next release.

guilhermeleobas:
I'm ok with these changes but this needs a second review from one of the core devs.

PR #9521, opened 2024-04-02
Support binop reduction.

### LINKED ISSUE #9518: numba's bug with prange without warning
(Edited by maintainer to inline file attachment)

Here is a minimal code to demonstrate a bug I found.This code can run smoothly without error and warning,
but it gives an unreasonable result, with output number ‘err1’ equals ‘err2’.However, ‘err1’ and ‘err2’ should be of opposite sign. Bug code see the attachment

```python
import numba
import time
from numba import jit
from numba import prange
import numpy as np
test=np.arange(10000).reshape((100,100))

@jit(nopython=True,parallel=True)
def bug(test,pre,err1,err2):
	for k in range(0,100):
		pre=pre+test[k,0]
	for k in prange(0,100):
		err1=err1+test[k,0]
		err2=err2-test[k,0]
	return(err1,err2)

err1,err2=bug(test,0,0,0)
print(err1,err2)

```

I found there are two ways to overcome the bug.

1.not use prange
2.Delete the first loop at line 10 and line11
     But, still I’d like to know how the bug arises,in case of bugs appear without warning
 next time. By the way, Could you please send a copy of your responce to my email baiyunfei@iphy.ac.cn
Lookinng forward to your reply.
Yours sincerely

DrTodd13:
Resolves #9518.

### DIFF
--- a/numba/core/ir.py
+++ b/numba/core/ir.py
@@ -336,6 +336,9 @@ class Inst(EqualityCheckMixin, AbstractRHS):
         else:
             return []
 
+    def __repr__(self):
+        return self.__str__()
+
 
 class Stmt(Inst):
     """
--- a/numba/parfors/parfor.py
+++ b/numba/parfors/parfor.py
@@ -3803,12 +3803,22 @@ def get_reduction_init(nodes):
     # there could be multiple extra assignments after the reduce node
     # See: test_reduction_var_reuse
     acc_expr = list(filter(lambda x: isinstance(x.value, ir.Expr), nodes))[-1].value
-    require(isinstance(acc_expr, ir.Expr) and acc_expr.op=='inplace_binop')
-    if acc_expr.fn == operator.iadd or acc_expr.fn == operator.isub:
-        return 0, acc_expr.fn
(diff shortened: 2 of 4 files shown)


gmarkall:
As CI is green and this is non-draft, I'm assuming this is ready for review, but please let me know if it's otherwise.

sklam (commented):
One minor issue need addressing; otherwise, everything looks good.

numba/core/ir.py
```diff
@@ -336,6 +336,12 @@ def _rec_list_vars(self, val):
         else:
             return []
 
+    def __repr__(self):
+        return self.__str__()
+
+    def __str__(self):
+        return super().__repr__()
+
```
sklam, line 344: My [suggestion](https://github.com/numba/numba/pull/9521#discussion_r1559923460) is leading to other test failures since the string repr behavior changed. @DrTodd13, maybe it is easier just to remove the these changes all together, assuming it is just needed for debugging.

DrTodd13: Done. @sklam

sklam (approved):

PR #9508, opened 2024-03-20
Add compile-time coverage for compiled code

sklam:
Let `Lower` gather coverage during compilation and write to active `Coverage` object.

It will use "numba_compiled" context to mark the lines visited by the compiler:

Screenshot 2024-03-20 at 3 33 02 PM

PR #9523, opened 2024-04-04
Test Codecov

sklam:
Based on #9508 
Re-enable codecov reporting


sklam:
Now i understand, secrets are not provided to forks
Screenshot 2024-07-19 at 10 07 52 AM

sklam:
After triage discussion, we decided to not enable CodeCov for now given the need to share AzureCI secrets. We talked about potentially doing this on GHA instead.

